PR #3605, opened 2018-04-20
[React table] bug fix - Scroll bar that appears under frozen headers 

kongtiaowang:
This pull request removes the Scrollbar from the frozen header.
https://redmine.cbrain.mcgill.ca/issues/14227


ZainVirani:
Not sure if this is related but when I tested this on my sandbox (stats table - behavioural; breakdown by scan type), when I scrolled to the right the table would jump back to the left side once i reached a certain point.

kongtiaowang:
@cmadjar Please re-test it.

cmadjar:
@kongtiaowang I still get the problem of alignment on my sandbox when I decrease the size of the window, scroll down the list so that the row header becomes frozen and then move from left to right into the table... No idea why I see that on my sandbox and not you.

screen shot 2018-04-30 at 12 04 30 pm

llevitis:
On my sandbox all the columns and column headers were aligned properly.

zaliqarosli:
screenshot 2018-05-07 14 44 34

looks like it works at first but misalignment appears towards the very right columns

johnsaigle:
@zaliqarosli I was unable to reproduce this -- tested on both Firefox and Chrome.

zaliqarosli:
looks like regardless of the size of the window, when the table is at the top, showing the first results, the columns are aligned. but once you scroll down the table, and the headers become floating, it gets misaligned

headers for first few results: 
screenshot 2018-05-07 14 46 37
screenshot 2018-05-07 14 46 54

after scrolling dow, with floating headers:
screenshot 2018-05-07 14 46 41
screenshot 2018-05-07 14 46 57

zaliqarosli:
@kongtiaowang another clue! the misalignment happens only when the PSCID column becomes floating on the left of the table. otherwise, it's aligned and fine. i'm not sure what makes the PSCID column float though - it only happens sometimes for me

zaliqarosli:
update: the misalignment of frozen headers exists outside of this PR (https://redmine.cbrain.mcgill.ca/issues/14786) and is a bug that is to be fixed separately.

this PR should only be removing the extra scrollbar underneath the headers. however, in doing this, it appears that the header expands vertically to fill the space that is now created when removing the scrollbar. this, as a result, exacerbates the misalignment.

this distinction should hopefully help this PR to pass and get merged

PR #3606, opened 2018-04-20
[BrainBrowser] Allow viewing of NIfTI files into BrainBrowser

cmadjar:
This pull request allows opening NIfTI files into Brain Browser.

I am not entirely satisfied of the solution I propose but since I suck in JS, I figured I would send the PR and ask for advice from the experts :). 

I am especially not really thrilled of using `async=false` to get the filename of a file and check if the file is .mnc or .nii. I feel this could be improved but I don't know how... It works but it slows down the loading so if anyone knows a better solution, I am super receptive to improving it!!

@driusan I was not sure who to assign this to so I assigned it to you and you could maybe assign it to JS gurus :). Thanks!


modules/brainbrowser/js/brainbrowser.loris.js
```diff
@@ -750,15 +750,37 @@ $(function() {
...
-            }
+        var filename = null;
+        $.ajax({
+            url: loris.BaseURL + "/brainbrowser/ajax/getMincName.php",
+            data: 'minc_id=' + minc_ids_arr[i],
+            method: 'POST',
+            success: function (data) {
+              filename = data;
+            },
+            async: false
```
driusan: If all you're trying to do is get rid of the async, I think you could replace the `async: false` with a promise chain by using `.then` on the ajax call and putting the `if/else if` inside of a function inside of the `then`.

ie.

```
$.ajax({ ... filename url stuff, minus async}).then(function(filename) {
     if (filename.endsWith("mnc") { ... } else if { ... } 
     etc
});
```

Alternatively (and probably better..), the if the script at`ajax/minc.php?minc_id=...` can be modified to include a `Content-Type` header, you could use that to push the appropriate object to BrainBrowser's volumes based on the header in the reply and only have to do one request.

HenriRabalais: @driusan Is it not necessary to instantiate a promise object when using an ajax call?

driusan: @HenriRabalais jquery ajax calls are (allegedly, I haven't tried) promise compatible, so it shouldn't be necessary.

cmadjar:
@driusan sorry for the delay in replying. I was busy reading a book to understand your language :P. I mean, I was busy reading a book on JS, JQUERY and AJAX. Your answer made a lot more sense after that so I followed your wisdom ;). Thanks!

In terms of time performance, it takes as much time as with the current version of the BB module but now it can load both NIfTI and MINC files in BB, yeah!

cmadjar:
@driusan @xlecours @davidblader who wish to review and eventually approve this? 

I see @um4r12 tested the PR and added the label "passed manual tests". Thanks!

modules/brainbrowser/js/brainbrowser.loris.js
```diff
@@ -748,17 +749,52 @@ $(function() {
...
 
+    var request;
     for (i = 0; i < minc_ids_arr.length; i += 1) {
 
-        minc_volumes.push({
-            type: 'minc',
-	    header_url: "",
-	    raw_data_url: loris.BaseURL + "/brainbrowser/ajax/minc.php?minc_id=" + minc_ids_arr[i],
-            template: {
+      request = $.ajax({
```
davidblader: perhaps it's outside the scope of this PR but i think it would be a good idea to condense this into just 1 ajax call if possible
other than that, LGTM

modules/brainbrowser/ajax/getMincName.php
```diff
@@ -18,8 +18,12 @@
...
 $query = "select File from files where FileID = :MincID";
-$m     = $DB->pselectOne($query, array('MincID' => $_REQUEST['minc_id']));
-//echo json_encode(($m));
-$m = substr($m, strrpos($m, '/') + 1);
-echo ($m);
+$file  = $DB->pselectOne($query, array('MincID' => $_REQUEST['minc_id']));
+$file  = substr($file, strrpos($file, '/') + 1);
+
+// create a JSON object with the file information
+$result->filename = $file;
```
driusan: `$result` isn't defined. I'm fairly sure this will generate a warning/notice.

modules/brainbrowser/js/brainbrowser.loris.js
```diff
@@ -748,17 +749,52 @@ $(function() {
         minc_ids_arr = [minc_ids];
     }
 
+    var request;
```
driusan: should probably use `let` instead of `var` for any new code in general.

modules/brainbrowser/js/brainbrowser.loris.js
```diff
@@ -748,17 +749,52 @@ $(function() {
...
+              header_url: "",
+              raw_data_url: loris.BaseURL + "/brainbrowser/ajax/minc.php?minc_id=" + fileid,
+              template: {
                 element_id: "volume-ui-template4d",
                 viewer_insert_class: "volume-viewer-display"
-            }
-        });
+              }
+            });
+          } else if (filename.endsWith("nii")) {
```
driusan: I'd prefer this was derived from the `Content-Type` header of the response rather than the filename, but this should work.

modules/brainbrowser/js/brainbrowser.loris.js
```diff
@@ -748,17 +749,52 @@ $(function() {
...
+            minc_volumes.push({
+              type: 'nifti1',
+              nii_url: loris.BaseURL + "/brainbrowser/ajax/minc.php?minc_id=" + fileid,
```
driusan: should the `minc.php` script and `minc_id` parameters be renamed? And can this use the API rather than a custom ajax script?

modules/brainbrowser/js/brainbrowser.loris.js
cmadjar: Oups. I read about that after having sent the PR ;). I changed the occurrences in the code I inserted but `var` is used all over BrainBrowser's code. Should probably be taken care of in another PR? What do you think @driusan?

modules/brainbrowser/js/brainbrowser.loris.js
cmadjar: @driusan should be done with the latest commit

modules/brainbrowser/js/brainbrowser.loris.js
cmadjar: @driusan I can rename the parameters for now and work on getting the file from the API in another PR. Thanks!

modules/brainbrowser/ajax/minc.php
```diff
@@ -65,6 +65,12 @@
 }
 
 if (!empty($minc_file)) {
+    if (preg_match('/nii(\.gz)?/i', $minc_file)) {
+        header('Content-Type: application/nii');
+    } elseif (preg_match('/mnc/', $minc_file)) {
+        header('Content-Type: application/mnc');
+    }
+    header('Content-FileID: ' . $_REQUEST['minc_id']);
```
driusan: This header name isn't standard and needs an X- prefix.

modules/brainbrowser/ajax/minc.php
```diff
@@ -65,6 +65,12 @@
 }
 
 if (!empty($minc_file)) {
+    if (preg_match('/nii(\.gz)?/i', $minc_file)) {
+        header('Content-Type: application/nii');
+    } elseif (preg_match('/mnc/', $minc_file)) {
+        header('Content-Type: application/mnc');
```
driusan: These aren't standard registered mime types. They need an x- prefix.

modules/brainbrowser/ajax/minc.php
cmadjar: good to know. Should be fixed now. Thanks!

cmadjar:
@driusan I think I addressed your comments. Let me know if more changes need to be done before it can be merged. Thanks!

PR #3608, opened 2018-04-23
[BrainBrowser] Log error when $mincPath is not readable

johnsaigle:
This pull request handles a Fatal Error that occurred when a path in Config was not set; an error was thrown but never caught by its caller, resulting in a Fatal error. This was found in BrainBrowser where MincTools was not set.

At the same time, additional related error handling was implemented in BrainBrowser.

A file check for `$minc_file` was corrected to `$minc_path` and verbose error reporting was added to assist developers with debugging. 

The above problems are likely to occur when developers are using the Raisinbread DB -- paths to minc files exist in the DB but not in the filesytem. However, this is not immediately clear without error reporting.

### DIFF
--- a/modules/brainbrowser/ajax/minc.php
+++ b/modules/brainbrowser/ajax/minc.php
@@ -70,7 +70,10 @@ if (!is_readable($minc_path)) {
         header('HTTP/1.1 404 Not Found');
         exit();
     } else {
-        error_log("$minc_path was requested but is not readable. Possible permission error");
+        error_log(
+            "$minc_path was requested but is not readable. " .
+            'Possible permission error'
+        );
         header('HTTP/1.1 403 Forbidden');
         exit();
     }
--- a/modules/brainbrowser/ajax/minc.php
+++ b/modules/brainbrowser/ajax/minc.php
@@ -64,11 +64,18 @@ if (strpos($_REQUEST['minc_id'], 'l') !== false) {
     $minc_path = getMincLocation() . $minc_file;
 }
 
-if (!empty($minc_file)) {
-    readfile($minc_path);
+if (!is_readable($minc_path)) {
+    if (!file_exists($minc_path)) {
+        error_log("ERROR: $minc_path exists in the DB but not in the file system");
+        header('HTTP/1.1 404 Not Found');
+        exit();
+    } else {
+        error_log("$minc_path was requested but is not readable. Possible permission error");
+        header('HTTP/1.1 403 Forbidden');
+        exit();
+    }
 } else {
-    header("HTTP/1.1 404 Not Found");
-    exit();
+    readfile($minc_path);
 }
 
 
@@ -82,7 +89,11 @@ function getMincLocation()
 {
     $config    =& NDB_Config::singleton();
     $paths     = $config->getSetting('paths');
-    $minc_path = $paths['mincPath'];
+    $minc_path = $paths['mincPath'] ?? '';
+    if (empty($minc_path)) {
+        header('HTTP/1.1 500 Internal Server Error');
+        exit();
+    }
     return $minc_path;
 }
 
--- a/modules/brainbrowser/ajax/minc.php
+++ b/modules/brainbrowser/ajax/minc.php
@@ -66,6 +66,7 @@ if (strpos($_REQUEST['minc_id'], 'l') !== false) {
 
 if (!is_readable($minc_path)) {
     if (!file_exists($minc_path)) {
+        http_response_code(500);
         error_log("ERROR: $minc_path exists in the DB but not in the file system");
         exit();
     } else {
@@ -73,6 +74,7 @@ if (!is_readable($minc_path)) {
             "$minc_path was requested but is not readable. " .
             'Possible permission error'
         );
+        http_response_code(403);
         exit();
     }
 } else {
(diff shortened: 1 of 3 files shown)


modules/brainbrowser/ajax/minc.php
```diff
@@ -64,11 +64,19 @@
     $minc_path = getMincLocation() . $minc_file;
 }
 
-if (!empty($minc_file)) {
-    readfile($minc_path);
+if (!is_readable($minc_path)) {
+    if (!file_exists($minc_path)) {
+        error_log("ERROR: $minc_path exists in the DB but not in the file system");
+        exit();
```
xlecours: missing http response code

modules/brainbrowser/ajax/minc.php
```diff
@@ -64,11 +64,19 @@
...
+    } else {
+        error_log(
+            "$minc_path was requested but is not readable. " .
+            'Possible permission error'
+        );
+        exit();
```
xlecours: missing http response code

php/libraries/NDB_Config.class.inc
```diff
@@ -447,7 +447,13 @@ class NDB_Config
         // nothing in the config file, so get the value from the DB
         // This will throw a ConfigurationException if it does not
         // exist.
-        return $this->getSettingFromDB($name);
+        try {
+            return $this->getSettingFromDB($name);
+        } catch (ConfigurationException $e) {
```
xlecours: There might be places where that exception is expected. Returning null would be unexpected.
If not, then the comment above need to be changed

php/libraries/NDB_Config.class.inc
```diff
@@ -447,7 +447,13 @@ class NDB_Config
...
+            error_log($e);
+            http_response_code(500);
```
xlecours: I would let the calling class decide of that

php/libraries/NDB_Config.class.inc
```diff
@@ -447,7 +447,13 @@ class NDB_Config
         // nothing in the config file, so get the value from the DB
         // This will throw a ConfigurationException if it does not
         // exist.
-        return $this->getSettingFromDB($name);
+        try {
+            return $this->getSettingFromDB($name);
+        } catch (ConfigurationException $e) {
+            error_log($e);
```
xlecours: this will trigger an array tp string conversion. And I don't think it is useful to fill the log with that. The calling class should decide what to do.

modules/brainbrowser/php/MincEnv.php.inc
```diff
@@ -28,7 +28,16 @@
 require_once "NDB_Config.class.inc";
 
 $config    =& NDB_Config::singleton();
-$minc_path = $config->getSetting("MINCToolsPath");
+$minc_path = $config->getSetting("MINCToolsPath") ?? '';
```
xlecours: This should be surrounded by a try->catch

xlecours: And remove the null coalesce.

php/libraries/NDB_Config.class.inc
johnsaigle: Done, this no longer returns null

johnsaigle:
@xlecours thanks for reviewing. I've addressed your comments

php/libraries/NDB_Config.class.inc
xlecours: But is still catches the exception.

johnsaigle: Ah okay I misunderstood. Fixed now

johnsaigle:
@xlecours could you please re-review when you have the time?

zaliqarosli:
changes probability need to be retested @kongtiaowang

modules/brainbrowser/php/MincEnv.php.inc
```diff
@@ -27,8 +27,18 @@
 
 require_once "NDB_Config.class.inc";
 
-$config    =& NDB_Config::singleton();
-$minc_path = $config->getSetting("MINCToolsPath");
+$config =& NDB_Config::singleton();
+try {
+    $minc_path = $config->getSetting("MINCToolsPath") ?? '';
+} catch (ConfigurationException $e){
```
driusan: This try/catch shouldn't be here. getSetting throws an exception by design. A setting that's being used not being configured is an exceptional circumstance indicating the server is misconfigured, and shouldn't be handled unless there's a good reason that the config might not exist in normal circumstances.

modules/brainbrowser/ajax/minc.php
```diff
@@ -64,11 +64,21 @@
...
+    if (!file_exists($minc_path)) {
+        error_log("ERROR: $minc_path exists in the DB but not in the file system");
+        http_response_code(500);
+        exit();
+    } else {
+        error_log(
+            "$minc_path was requested but is not readable. " .
+            'Possible permission error'
+        );
+        http_response_code(403);
```
driusan: a 403 means that the user doesn't have permission to access the resource, not that the server couldn't retrieve it.

driusan: The quote type also change between lines for no good reason in the message above.

xlecours:
@johnsaigle Conflicts

PR #3164, opened 2017-10-02
[Imaging Uploader] Uploader improvements

davidblader:
PSCID and CandID fields are now drop-down select fields. Gets auto-populated once one is filled in. Switched order of fields to better match file naming convention.
Also check filename on front-end before file upload attempt

PR #3615, opened 2018-04-26
[Imaging Uploader] Adds validation checks in Upload tab and displays errors on form

victoriafoing:
### Brief summary of changes

This pull request adds validation checks in the Upload tab of Imaging Uploader.

Users can now only upload images if:

1. The candidate exists in the database
2. The PSCID is valid for the CandID
3. The Visit label is valid for the CandID
4. The candidate is active in the database
5. The session is active in the database
6. The file is of type .tgz or tar.gz or .zip. 
7. The file does not exceed 2M
8. The file name matches the PSCID_CandID_VisitLabel convention (once CandID, PSCID, and Visit Label are validated)

If there are validation errors, a swal window with "Submission error" will appear and the errors will be displayed under relevant form elements in the Upload form.

- Ensuring the user submits the correct values when uploading images
- Ensuring the user uploads a file with the proper extension, size, and name
- Not taking users to a new page when there is a validation error
- Clarifying errors and displaying error messages on the form so the user can easily fix them
- Fixing small bugs

### This resolves issue...

- Redmine Task #14006: https://redmine.cbrain.mcgill.ca/issues/14006
- Redmine Task #14355: https://redmine.cbrain.mcgill.ca/issues/14355
- Additional issues were discussed in the comment section of this PR
The target branch is minor.

### To test this change...

To test this, pick a candidate in the database.

Go through the list of validations above and try to upload a file.
See if the proper errors are returned on the form.

To test validation checks for CandID, PSCID, and Visit Label,
Set Phantom Scans to "No" and try to upload a .zip file for each scenario:

1. A CandID that does not exist in the 'candidate' database
2. A PSCID that is invalid for a specific CandID in the 'candidate' database
3. A Visit Label that is invalid for a specific CandID in the 'session' database
4. A candidate that has Active set to 'N' in the 'candidate' database
5. A session that has Active set to 'N' in the 'session' database

_e.g. SQL Commands
Setting candidate associated with CandID X as inactive
UPDATE candidate SET Active = 'N' WHERE CandID = *X*;_

_Setting session associated with CandID X and Visit Label Y as inactive
UPDATE session SET Active = 'N' WHERE CandID = *X* AND Visit_label = *'Y'*;_

To test validation checks for File to Upload,
Set Phantom Scans to "No" and try to upload a file for each scenario:

6. A file that is not of type .tgz or tar.gz or .zip. 
7. A file that exceeds 2M
8. A file that that does not match the PSCID_CandID_VisitLabel convention (once CandID, PSCID, and Visit Label are validated)

You should also test **a combination of errors** or **no errors**


kongtiaowang:
Run PHPCS to fix the errors and warnings.

modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -326,7 +326,32 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
...
+                        "SELECT COUNT(*) FROM session s
+                        JOIN candidate c ON (c.candid=s.candid)
+                        WHERE s.Visit_label = :vlabel
+                        AND c.pscid = :pid AND c.candid = :cid AND c.active = 'Y'",
+                        [
+                         ':cid'    => $candid,
+                         ':vlabel' => $visit_label,
+                         ':pid'    => $pscid,
+                        ]
+                    );
```
PapillonMcGill: Should also look for active session (s.active).

Instead of adding more query, could modify existing query to get individual attributes instead of count and then build proper message to display. (using left join)

If number of row = 0, candidate nonexistent
visitlabel is null, no visit 
c.active  = 'N', candidate not active
s.active = 'N' session not active

kongtiaowang:
Run "vendor/bin/phpcbf --standard=docs/LorisCS.xml php/libraries php/exceptions php/installer modules/imaging_uploader/ ". to fix the phpcs error.

modules/imaging_uploader/php/imaging_uploader.class.inc
nicolasbrossard: 👍 for Melanie's solution. Giving more details as to why the error occurred makes sense.

modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -242,6 +242,7 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
      *       .gz, .zip or .tgz
      *  - The fields   CandID, PSCID  and VisitLabel are filled out
      *  - The inputted PSCID,  CandID and VisitLabel exist in the database
+     *  - The inputted candidate is active in the database
```
nicolasbrossard: Nitpick: "inputted" sounds weird to me...Maybe change to "The <whatever> that was input...." ?

modules/imaging_uploader/php/imaging_uploader.class.inc
zaliqarosli: also agree

modules/imaging_uploader/php/imaging_uploader.class.inc
zaliqarosli: I'm not sure about "The that was input ..".

maybe "The input candidate" or "The candidate inputted"?

nicolasbrossard: @zaliqarosli @PapillonMcGill exactly

zaliqarosli: i approve "The candidate that was inputted" lol

modules/imaging_uploader/php/imaging_uploader.class.inc
HenriRabalais: Great, that would combine the query @victoriafoing made with the previous one starting on line 308 for an elegant solution.

modules/imaging_uploader/php/imaging_uploader.class.inc
driusan: The past tense of "input" is "input"

zaliqarosli: wow really?? amazing that i'm still learning new things about this language

nicolasbrossard:
Try experimenting with

```
SELECT c.candid, c2.candid, s.visit_label
FROM candidate c
LEFT JOIN candidate c2 
ON (c.candid=c2.candid AND c2.pscid = :pscid)
LEFT JOIN session s
ON (s.candid=c.candid and s.visit_label =:visit_label)
WHERE c.candid =:candid
```

and playing around with different values of :candid, :pscid and :visit_label. That should get you close to what you are looking for.

PapillonMcGill:
with current query in the PR, if the CandID ans PSCID do not match, there would be no row returned. If the candidate doesn't have the visit label, visit label value in the returned row will be null in both query.

@nicolasbrossard query will report a row if the CandID exist but is not associated with the PSCID. In that case, the c2.CandID attributes will be null. The danger is that the CandID is associated with more than one PSCID. In that case, multiple rows would be returned.

You could kept the query as is currently and change the first error message for "The candidate associated with CandID $candID and PSCID $PSCID does not ...."  or add more code and check the possibility of multiple rows

nicolasbrossard:
@PapillonMcGill a candidate ID can only be associated with one PSCID, so basically here's what can be returned  by the query I proposed :
- Nothing -> The candidate ID does not exist
- Multiple rows -> Should not happen: this means that the candidate ID is associated to more than one PSCID or that there is more than one visit in the session table with the given visit_label. 
- A candidate ID but a NULL PSCID -> Candidate ID exists but the PSCID does not match (visit label not important here as the mismatch is more critical to report)
- A candidate ID, a PSCID and a NULL visit label -> Candidate ID and PSCID match but the candidate has not been seen at the specified visit.
- A candidate ID, a PSCID and a visit label -> All is good!

Makes sense?

PapillonMcGill:
@nicolasbrossard If a candidate can only be associated with 1 PSCID, than what happend? we can't associate the same candidate with a different one (3rd bullet). PSCID can not be null in the DB, so the candidate as to be associated with a different PSCID. 
Do we still need the left join from candidate to candidate in your query? If one of CanID or PSCID from user input is not valid then the user doesn't exist. How could we tell if the user entered the wrong CandID vs wrong PSCID? 
For the list of check in your comment, yes, that should be the correct way and order to do it.

modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -304,27 +305,54 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
...
-                                " created in the database. Please refresh this".
-                                " page to return to the Imaging Uploader module";
+                if (sizeOf($candidate) == 0) {
+                    $errors[] = "The candidate associated with that CandID".
+                                " does not exist in the database.".
+                                " Please refresh this page to return to".
+                                " the Imaging Uploader module.";
+                } else {
+                    if (!$candidate['candid_with_pscid']) {
+                        $errors[] = "The CandID and the PSCID do not match.".
```
cmadjar: Would clarify this error to something like that:
`The PSCID provided in the upload for the candidate CandID do not match the one stored in the database. Please ...`

modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -304,27 +305,54 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
...
 
-                if ($result == 0) {
-                    $errors[] = "Make sure you enter the CandID: $candid,".
-                                " PSCID: $pscid and the corresponding".
-                                " Visit-label: $visit_label. Candidate and".
-                                " corresponding visit label must already be" .
-                                " created in the database. Please refresh this".
-                                " page to return to the Imaging Uploader module";
+                if (sizeOf($candidate) == 0) {
+                    $errors[] = "The candidate associated with that CandID".
```
cmadjar: in this case, it returns the following:
screen shot 2018-05-09 at 11 21 32 am

- I would tend to make the error logging consistent across the different cases maybe.
- Note sure the mri_file is needed in front of the error
- Would specify which CandID you are referring to in the error message (a bit like in the first row's error message but replacing DCCID by CandID)

modules/imaging_uploader/php/imaging_uploader.class.inc
cmadjar: Also, I would tend to display in the error message which PSCID and CandID they provided in the imaging uploader.

modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -304,27 +305,54 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
...
+                                " the Imaging Uploader module.";
+                } else {
+                    if (!$candidate['candid_with_pscid']) {
+                        $errors[] = "The CandID and the PSCID do not match.".
+                                    " Please refresh this page to return to".
+                                    " the Imaging Uploader module.";
+                    }
+                    if (!$candidate['visit_label']) {
+                        $errors[] = "The candidate associated with that CandID".
+                                    " and PSCID does not have that Visit Label.".
```
cmadjar: Would clarify this error to something like that:
`The provided Visit Label does not exist for the candidate associated with that CandID and PSCID. Please...`

Would also display here the provided Visit label, CandID, PSCID.

modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -304,27 +305,54 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
...
+                                    " Please refresh this page to return to".
+                                    " the Imaging Uploader module.";
+                    }
+                    if ($candidate['candidate_active'] == 'N') {
+                        $errors[] = "The candidate is not active in the database.".
```
cmadjar: Here I would add the PSCID and Candidate to know which candidate you are referring to.

modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -304,27 +305,54 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
...
+                                    " Please refresh this page to return to".
+                                    " the Imaging Uploader module.";
+                    }
+                    if ($candidate['session_active'] == 'N') {
+                        $errors[] = "The session is not active in the database.".
```
cmadjar: same here, I would add the Visit label and candidate CandID, PSCID

cmadjar:
@victoriafoing Needs rebase. One the rebase is done, is the PR ready for another round of review? Thanks!

modules/imaging_uploader/jsx/UploadForm.js
```diff
@@ -195,17 +195,15 @@ class UploadForm extends React.Component {
...
-        let responseText = err.responseText;
-        if (responseText.indexOf(errMessage) > -1) {
-          responseText = responseText.replace('history.back()', 'location.reload()');
-          document.open();
-          document.write(responseText);
-          document.close();
-        }
-        console.error(err);
-      }
+          console.log(err);
```
davidblader: use `console.error()`

victoriafoing:
After looking at the upload form in the Media module, I propose that we make the upload form in Imaging Uploader similar to that one. The upload form in the Media module provides dropdowns for PSCID and Visit Label that correspond with what is in the database. If we provide dropdowns in Imaging Uploader for CandID, PSCID, and Visit Label, we do not have to validate text entered by the user. All we would have to validate is that the File has the proper name and extension. Let me know if you agree and I'll proceed with this change.

cmadjar:
> After looking at the upload form in the Media module, I propose that we make the upload form in Imaging Uploader similar to that one. The upload form in the Media module provides dropdowns for PSCID and Visit Label that correspond with what is in the database. If we provide dropdowns in Imaging Uploader for CandID, PSCID, and Visit Label, we do not have to validate text entered by the user. All we would have to validate is that the File has the proper name and extension. Let me know if you agree and I'll proceed with this change.

I am uncertain if for the Imaging Uploader, we want the user to enter manually the PSCID, CandID and not populate it automatically. @MounaSafiHarab Do you remember? I feel we had this discussion in the past but my memory sucks...

davidblader:
> After looking at the upload form in the Media module, I propose that we make the upload form in Imaging Uploader similar to that one. The upload form in the Media module provides dropdowns for PSCID and Visit Label that correspond with what is in the database. If we provide dropdowns in Imaging Uploader for CandID, PSCID, and Visit Label, we do not have to validate text entered by the user. All we would have to validate is that the File has the proper name and extension. Let me know if you agree and I'll proceed with this change.
> I am uncertain if for the Imaging Uploader, we want the user to enter manually the PSCID, CandID and not populate it automatically. @MounaSafiHarab Do you remember? I feel we had this discussion in the past but my memory sucks...

@cmadjar 
it's something that i'm planning to implement in #3164

modules/imaging_uploader/jsx/UploadForm.js
```diff
@@ -195,17 +195,15 @@ class UploadForm extends React.Component {
...
+          swal({
+            title: "Errors with Field Values",
+            text: "Display errors here",
+            type: "error",
+            confirmButtonTexr: 'OK'
+          });
+          this.setState({uploadProgress: -1});
+       }.bind(this),
```
HenriRabalais: @davidblader should we be encouraging the use of arrow functions over the the binding of `this`?

davidblader: not sure. both ways are perfectly fine. i guess it would boil down to: what's more confusing, `.bind(this)` or `() => {...}`?
which, in my opinion, are probably about equally confusing for someone unfamiliar with JS, but i would have to go with `.bind(this)` since at least that is consistent with the rest of the code base, but i don't have a strong enough opinion one way or the other

modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -333,12 +358,15 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
...
-                        page to return to the Imaging Uploader module";
+            array_push($errors,
+                       "The file $temp_file is not of type".
+                       " .tgz, .tar.gz or .zip.");
         }
 
         if (count($errors) > 0) {
+            array_push($errors,
+                       "Please refresh this page to return to".
+                       " the Imaging Uploader module.");
```
HenriRabalais: If you would like to have this presented to the front in a way that conforms more with React, I suggest throwing a `400` error and then exiting the script with a json encoding of `$errors`. This will allow you to access all the errors in the error clause of the jquery ajax call in the jsx file.

HenriRabalais: e.g. 
`http_response_code(400);
header('Content-Type: application/json; charset=UTF-8);
exit(json_encorde(['errors'=>$errors]));`

victoriafoing:
This PR is ready for another round of review! Thanks @HenriRabalais for your help with React and json encoded errors

zaliqarosli:
PR still passes testing! you might still need to remove some unnecessary changes, particularly spacing for comments, before the PR gets approved

modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -299,34 +295,69 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
...
+                        array_push(
+                            $errors,
+                            "The CandID and the PSCID do not match."
+                        );
+                    }
+                    if (!$candidate['visit_label']) {
+                        array_push(
+                            $errors,
+                            "The candidate associated with that CandID".
+                            " and PSCID does not have that Visit Label."
```
zaliqarosli: this comment is mostly a reminder. only valid visit labels should be given as options in the form anyways. this should be addressed in another PR. @victoriafoing has a ticket already been made for it?

victoriafoing:
Please re-test this PR! 😄

modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -299,34 +295,79 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
...
+                        "</strong> does not exist in the database."
+                    );
+                } else {
+                    if (!$candidate['candid_with_pscid']) {
+                        array_push(
+                            $errors,
+                            "The PSCID <strong>" . $pscid . "</strong>" .
+                            " does not exist for the candidate" .
+                            " associated with CandID <strong>" . $candid .
+                            "</strong> in the database."
```
cmadjar: Here I would rephrase the error to:
"Invalid PSCID for CandID $candid. Check that you entered the correct PSCID and CandID."

modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -283,11 +279,11 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
             $pscid       = trim($args['values']['pSCID']);
             $visit_label = trim($args['values']['visitLabel']);
             if ((empty($candid)) || (empty($pscid)) || (empty($visit_label))) {
-                $errors[] ="Make sure the Cand ID, PSCID and Visit Label are ".
-                "filled out. Please refresh this page to return to the Imaging
-                Uploader module.";
+                array_push(
+                    $errors,
+                    "Make sure the Cand ID, PSCID and Visit Label are filled out."
```
cmadjar: "CandID" instead of "Cand ID"

ridz1208:
removing milstone. milstone is added when PR is merged

victoriafoing:
The errors are now displayed on the form. Please read the new testing instructions and re-test!

MounaSafiHarab:
@victoriafoing 

just a quick question 
is your branch up to date with the latest bug fixes on Imaging Uploader brought recently by @cmadjar  and @davidblader? if so, @nicolasbrossard  and I can test Imaging Uploader off of your branch...

MounaSafiHarab:
as pointed out by @davidblader 
the bug of the error messages not displayed to the user are NOT on the minor branch, so testing this for the major release will not solve the issue there (the bug is on major),
so I think the bug should be fixed on the major branch

victoriafoing:
Hi @MounaSafiHarab this PR is not intended to be a bug fix, it is adding more validations and changing the way the errors are displayed to the user.
In minor, when there are errors, they are displayed on a new page:
screen shot 2018-06-21 at 13 34 09

But in this PR, they are displayed on the form instead:
screen shot 2018-06-21 at 13 35 49

The bug on major seems to be a new issue.

victoriafoing:
Hi everyone, I'm closing this PR because it has been moved to Major: https://github.com/aces/Loris/pull/3761

PR #3616, opened 2018-04-27
Adding react and proptypes dependencies

xlecours:
This remove the react.min.js and react-with-addon files from the htdocs/js diretory and add those libraries to webpack.

This PR is a proof of work that should be build on or abandon.

The files in  htdocs/js/components have been deleted (almost all of them) because they should be include in the React App in import using an `import <Component> from 'jsx/<Component>'` command


smarty/templates/main.tpl
```diff
@@ -32,13 +32,7 @@
...
           $(document).ready(function() {
             {if $crumbs != "" && empty($error_message)}
-              var crumbs = {$crumbs|@json_encode},
-                      baseurl = "{$baseurl}",
-                      breadcrumbs = RBreadcrumbs({
-                        breadcrumbs: crumbs,
-                        baseURL: baseurl
-                      });
-              ReactDOM.render(breadcrumbs, document.getElementById("breadcrumbs"));
+              window.crumbs = {$crumbs|@json_encode};
```
xlecours: There should be a better way. 
those crumbs could be get with ajax maybe

smarty/templates/main.tpl
```diff
@@ -32,13 +32,7 @@
         <script type="text/javascript">
           $(document).ready(function() {
             {if $crumbs != "" && empty($error_message)}
-              var crumbs = {$crumbs|@json_encode},
-                      baseurl = "{$baseurl}",
-                      breadcrumbs = RBreadcrumbs({
-                        breadcrumbs: crumbs,
-                        baseURL: baseurl
-                      });
-              ReactDOM.render(breadcrumbs, document.getElementById("breadcrumbs"));
```
xlecours: Moved to the Breadcrumb component file

xlecours:
This was open for discussion.
Moved to https://github.com/xlecours/Loris/pull/4

PR #3578, opened 2018-04-10
[Help_editor] Fixes "Return to .." href location

zaliqarosli:
Fixes https://redmine.cbrain.mcgill.ca/issues/14111. Ideally would like to completely remove the "success" page and add a sweet alert but fixing this should be sufficient for now.

tpl_data['url'] comes from $_COOKIE["LastUrl"] as set by document.cookie

PR #3350, opened 2017-12-13
[Help Editor] Fix Help Editor module preview & help section rendering

zaliqarosli:
This pull request lets help section for all Loris pages and preview feature of Help Editor module to render Markdown content. It uses the Markdown and DisplayHelp React components to display and render the form content.

This PR lets modules take help content from the markdown files in their help directories, while instruments' help content is managed on the front-end, using the Help Editor module. This is up for discussion.

PR #3617, opened 2018-04-27
[Help_editor] Add ajax/swal on save and remove success page

zaliqarosli:
Replaces https://github.com/aces/Loris/pull/3578
Fixes https://redmine.cbrain.mcgill.ca/issues/14111

- [x] Remove the confirmation page
- [x] Use an ajax call to save the content
- [x] Add swal message on error/success
- [x] Replace use of cookies with Referer header

I could only figure out how to pass the $_GET variables to the js and ajax scripts via hidden input elements on the form smarty template. Please let me know if there is a better way to do this!! It feels a little hacky but I couldn't find an alternative

screenshot 2018-05-07 11 49 29

### DIFF
--- a/modules/help_editor/ajax/process.php
+++ b/modules/help_editor/ajax/process.php
@@ -1,0 +1,105 @@
+<?php
+/**
+ * This file contains code for editing help content.
+ * It holds the processing of adding/updating content
+ * for all modules.
+ *
+ * PHP Version 7
+ *
+ * @category Main
+ * @package  Loris
+ * @author   Zaliqa Rosli <zaliqa.rosli@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://github.com/aces/Loris
+ */
+
+namespace LORIS\help_editor;
+
+require_once 'Database.class.inc';
+require_once 'helpfile.class.inc';
+
+$user = \User::singleton();
+if (!$user->hasPermission('context_help')) {
+    header("HTTP/1.1 403 Forbidden");
+    exit(
+        "You do not have the correct permissions for this operation."
+    );
+}
+
+$DB = \Database::singleton();
+//Get the default values
+// Sanitize user input
+if (!empty($_POST['helpID'])
+    && !empty($_POST['title'])
+    && !empty($_POST['content'])
+) {
+    $helpID = $_POST['helpID'];
+    $help_file = HelpFile::factory($helpID);
+    // update the help file
+    $success = $help_file->update(
+        array(
+         'topic'   => $_POST['title'],
+         'content' => $_POST['content'],
+         'updated' => date(
+             'Y-m-d h:i:s',
+             time()
+         ),
+        )
+    );
+} else {
+    //content does not exist insert the help file
+    if (!empty($_POST['section'])
+        && $_POST['subsection'] != 'undefined'
+        && empty($_POST['parentID'])
+    ) { 
+        //create parent help section first
+        $parentID = HelpFile::insert(
+            array(
+             'hash'    => md5($_POST['section']),
+             'topic'   => "",
+             'content' => "Under construction",
+             'created' => date(
+                 'Y-m-d h:i:s',
+                 time()
+             ),  
+            )
+        );
+         // check errors
+    }    
+    if (!empty($_POST['section'])
+        && $_POST['subsection'] != 'undefined'
+        && !empty($_POST['parentID'])
+    ) { 
+
+        // insert the help file
+        $helpID = HelpFile::insert(
+            array(
+             'parentID' => $_POST['parentID'],
+             'hash'     => md5($_POST['subsection']),
+             'topic'    => $_POST['title'],
+             'content'  => $_POST['content'],
+             'created'  => date(
+                 'Y-m-d h:i:s',
+                 time()
+             ),  
+            )
+        );
+
+    } else if (!empty($_POST['section'])
+        && $_POST['subsection'] == 'undefined'
+    ) { 
+      //default case
+      $helpID = HelpFile::insert(
+      array(
+           'hash'    => md5($_POST['section']),
+           'topic'   => $_POST['title'],
+           'content' => $_POST['content'],
+           'created' => date(
+               'Y-m-d h:i:s',
+               time()
+           ),  
+          )
+      );
+    }
+}
+?>
--- a/modules/help_editor/ajax/process.php
+++ b/modules/help_editor/ajax/process.php
@@ -33,7 +33,7 @@ if (!empty($_POST['helpID'])
     && !empty($_POST['title'])
     && !empty($_POST['content'])
 ) {
-    $helpID = $_POST['helpID'];
+    $helpID    = $_POST['helpID'];
     $help_file = HelpFile::factory($helpID);
     // update the help file
     $success = $help_file->update(
@@ -51,7 +51,7 @@ if (!empty($_POST['helpID'])
     if (!empty($_POST['section'])
         && $_POST['subsection'] != 'undefined'
         && empty($_POST['parentID'])
-    ) { 
+    ) {
         //create parent help section first
         $parentID = HelpFile::insert(
             array(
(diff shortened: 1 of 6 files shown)


PapillonMcGill:
tested, work fine.

Found some character that are not utf8 in helpID 1, not sure if it as already been fixed in prod.

modules/help_editor/ajax/process.php
```diff
@@ -0,0 +1,105 @@
...
+ * @category Main
+ * @package  Loris
+ * @author   Zaliqa Rosli <zaliqa.rosli@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://github.com/aces/Loris
+ */
+
+namespace LORIS\help_editor;
+
+require_once 'Database.class.inc';
```
xlecours: This is done by AjaxHelper.php and can be removed.

modules/help_editor/ajax/process.php
```diff
@@ -0,0 +1,105 @@
...
+require_once 'helpfile.class.inc';
```
xlecours: This is done by AjaxHelper.php and can be removed.

modules/help_editor/ajax/process.php
```diff
@@ -0,0 +1,105 @@
...
+
+$user = \User::singleton();
+if (!$user->hasPermission('context_help')) {
+    header("HTTP/1.1 403 Forbidden");
+    exit(
+        "You do not have the correct permissions for this operation."
+    );
+}
+
+$DB = \Database::singleton();
```
xlecours: Please use `(NDB_Factory::singleton())->database()`

modules/help_editor/php/edit_help_content.class.inc
```diff
@@ -58,16 +58,22 @@ class Edit_Help_Content extends \NDB_Form
...
+        }
         if (!empty($_REQUEST['section'])) {
             $helpID = HelpFile::hashToID(md5($_REQUEST['section']));
         }
         if (!empty($_REQUEST['section'])
             && $_REQUEST['subsection'] != 'undefined'
         ) {
-            $helpID = HelpFile::hashToID(md5($_REQUEST['subsection']));
+            $helpID   = HelpFile::hashToID(md5($_REQUEST['subsection']));
+            $parentID = \LORIS\help_editor\HelpFile::hashToID(md5($_GET['section']));
```
xlecours: The namespace prefix is not required because this file is already in `\LORIS\help_editor\`

modules/help_editor/templates/form_edit_help_content.tpl
```diff
@@ -35,10 +28,13 @@
             <div class="col-xs-12"></div>
         {/foreach}
         <br>
-        <input class="btn btn-sm btn-primary col-sm-offset-3" name="fire_away" value="Save" type="submit" />
+        <input type="hidden" id="section" value="{$section}">
+        <input type="hidden" id="subsection" value="{$subsection}">
+        <input type="hidden" id="helpID" value="{$helpID}">
+        <input type="hidden" id="parentID" value="{$parentID}">
+        <input class="btn btn-sm btn-primary col-sm-offset-3" id="save-help" name="fire_away" value="Save" type="submit" />
         <input type="button" name="reset" value="Reset" class="btn btn-sm btn-primary" onclick="location.href='{$baseurl}/help_editor/edit_help_content/?section={$section}&subsection={$subsection}'" />
```
xlecours: This would need to be included in the javascript to prevent from doing a HTTP request (that would change the referer to this page instead of the previous)

zaliqarosli:
@PapillonMcGill i believe that was fixed in https://github.com/aces/Loris/pull/3350. if you remove the weird character's in your content and save it, it shouldn't reappear

modules/help_editor/templates/form_edit_help_content.tpl
zaliqarosli: i also need to change the swal confirm button text for when it returns to the Help Editor module, instead of the instrument/module the help content is for

modules/help_editor/js/help_editor_helper.js
```diff
@@ -35,7 +34,48 @@ $("input[name=preview]").click(function(e) {
...
+            section: section ? section : '',
+            subsection: subsection ? subsection : '',
+            parentID: parentID ? parentID : '',
+            helpID: helpID ? helpID : '',
+        },
+        success: function() {
+            swal({
+                title: "Content update successful!",
+                type: "success",
+                confirmButtonText: "Return to " + title,
```
johnsaigle: If I edit New Profile, for example, this confirmation button says "Return to New Profile" but takes me back to the Help Editor.

This should be changed to say e.g. "Back to Help Editor" or perform the redirect

modules/help_editor/js/help_editor_helper.js
```diff
@@ -35,7 +34,48 @@ $("input[name=preview]").click(function(e) {
...
+                closeOnConfirm: false,
+            },
+            function(){
+                location.href = document.referrer;
+            });
+        },
+        error: function(xhr, errStatus, errThrown) {
+            console.log("THIS IS THE STATUS:" + errStatus);
```
johnsaigle: I think this should be made more specific to this module or else removed 😛

modules/help_editor/templates/form_edit_help_content.tpl
johnsaigle: It would be nice here if the `helpID` GET parameter was preserved; right now when you click `Reset` a blank help editor page is loaded and you have to manually navigate back to the page you wanted to edit.

modules/help_editor/js/help_editor_helper.js
johnsaigle: Oops just saw you mentioned this to Xavier!

modules/help_editor/ajax/process.php
```diff
@@ -0,0 +1,105 @@
...
+//Get the default values
+// Sanitize user input
```
johnsaigle: This doesn't actually do any sanitizing as far as I can see! This likely isn't a problem because this is taken care of elsewhere in the code base, but this comment should be removed as it's misleading.

modules/help_editor/js/help_editor_helper.js
zaliqarosli: done

modules/help_editor/templates/form_edit_help_content.tpl
zaliqarosli: done

modules/help_editor/js/help_editor_helper.js
zaliqarosli: done!

zaliqarosli:
@xlecours @johnsaigle please re-review/test when you get a chance! :)

zaliqarosli:
@johnsaigle did you make any changes before hitting reset? reset pulls up content prior to change

ZainVirani:
@johnsaigle It worked for me. It doesn't reset it to a blank form, it resets it to the last saved value.

zaliqarosli:
@johnsaigle woohoo! thanks for testing :) @ZainVirani

modules/help_editor/js/help_editor_helper.js
```diff
@@ -35,7 +34,51 @@ $("input[name=preview]").click(function(e) {
...
+    $.ajax({
+        type: 'POST',
+        url: loris.BaseURL + '/help_editor/ajax/process.php',
+        data: {
+            title: title ? title : '',
+            content: content ? content : '',
+            section: section ? section : '',
+            subsection: subsection ? subsection : '',
+            parentID: parentID ? parentID : '',
+            helpID: helpID ? helpID : '',
```
driusan: not a blocker, but I think these would read better as `title: title || ''` (etc)

modules/help_editor/php/edit_help_content.class.inc
```diff
@@ -58,16 +58,22 @@ class Edit_Help_Content extends \NDB_Form
...
             && $_REQUEST['subsection'] != 'undefined'
         ) {
-            $helpID = HelpFile::hashToID(md5($_REQUEST['subsection']));
+            $helpID   = HelpFile::hashToID(md5($_REQUEST['subsection']));
+            $parentID = HelpFile::hashToID(md5($_GET['section']));
         }
         $this->tpl_data['section']    = $safeSection;
         $this->tpl_data['subsection'] = $safeSubsection;
+        $this->tpl_data['helpID']     = $helpID ? $helpID : '';
+        $this->tpl_data['parentID']   = $parentID ? $parentID : '';
```
driusan: Again, not a blocker, but these could use the elvis operator to make it a little cleaner (https://stackoverflow.com/questions/1993409/operator-the-elvis-operator-in-php)

PR #3618, opened 2018-04-30
[Survey][Instrument_List] Allow instruments from the battery to be set as surveys

ZainVirani:
This PR addresses one small part of the design change documented [here](https://forum.loris.ca/t/feature-proposal-enable-adding-surveys-from-instrument-list-in-candidate-timepoint/157).

Specifically it allows instruments in a battery for a timepoint that has been started to be set as a survey in the survey module (and optionally sent to a candidate/proband in an email).
In the instrument_list module, the link to the instrument will not appear for surveys *until* they have been set as surveys in the survey module (there is a tooltip explaining this to users who hover over the name of the survey in the instrument list).


ZainVirani:
This won't work right now, I need to make a slight change.

kongtiaowang:
Fixing the PHAN errors.

HenriRabalais:
@ZainVirani How should this be tested?

modules/instrument_list/php/instrument_list.class.inc
```diff
@@ -185,6 +207,10 @@ class Instrument_List extends \NDB_Menu_Filter
 
                 $Ins = "instruments";
 
+                if ($isSurvey == '1' && !$isSent) {
```
HenriRabalais: Should the value for `$isSurvey` be expected as a string?

modules/instrument_list/php/instrument_list.class.inc
```diff
@@ -167,6 +169,26 @@ class Instrument_List extends \NDB_Menu_Filter
...
+                    WHERE Test_name=:TN
+                    AND CommentID=:CID",
+                    array(
+                     'TN'  => $instrument['Test_name'],
+                     'CID' => $instrument['CommentID'],
+                    )
+                );
+
                 // print the sub group header row
                 if ($instrument['Sub_group'] != $prevSubgroup) {
```
HenriRabalais: given that `$instrument['Sub_group'] ` is set to `$prevSubgroup` I think you can add an extra equal sign (`!==`) for a type check as well.

HenriRabalais: Sorry, just realised this change was not made by you. Ignore that!

modules/instrument_list/php/instrument_list.class.inc
ZainVirani: You're right, it's a `tinyint(1)`

ZainVirani: It's `tinyint(1)` in the DB but `gettype($isSurvey)` returns string. I think this check is fine.

HenriRabalais:
Attempted a manual test, but my battery is not loaded due to the fact that I do not have instruments. I recommend that this be tested by someone else.

ZainVirani:
@christinerogers @me-evans I'm going to close this. It doesn't seem like any decision has been made and I haven't heard anything about this since May.

PR #3619, opened 2018-04-30
[Data Release] Adding error handling for missing default `user_uploads` directory

Jkat:
So that it doesn't silently fail upon improper setup

### DIFF
--- a/modules/data_release/ajax/FileUpload.php
+++ b/modules/data_release/ajax/FileUpload.php
@@ -29,6 +29,10 @@ if ($_POST['action'] == 'upload') {
     if (!file_exists(__DIR__ . "/../user_uploads/")) {
         header("HTTP/1.1 400 Bad Request");
         echo "File upload failed. Default user_uploads directory not found.";
+    } elseif (!is_writable(__DIR__ . "/../user_uploads/")) {
+        header("HTTP/1.1 400 Bad Request");
+        echo "File upload failed. Default user_uploads directory"
+        . " does not appear to be writeable.";
     } else {
         $target_path = $base_path . $fileName;
         if (move_uploaded_file($_FILES["file"]["tmp_name"], $target_path)) {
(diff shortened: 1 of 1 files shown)


Jkat:
Basically only new lines 29, 30 and 31 were added, and everything else in the else clause was already there before

ZainVirani:
Passed manual tests, but the errors could look a little nicer... ¯\_(ツ)_/¯

Jkat:
Yeah I thought about it but I was thinking to leave it up to the developer since they may want to customize(symlink, etc) this directory location as Data Releases may be rather large in some cases and/or take up a bit of space over time, so a separate file server may be used instead....maybe creating this directory in the installation process could be at least a decent middle ground..

johnsaigle:
@Jkat see #3620. I've added this directory during the install process.

modules/data_release/ajax/FileUpload.php
```diff
@@ -21,47 +21,55 @@
...
     $settings = $factory->settings();
 
     $baseURL = $settings->getBaseURL();
 
-    header("Location: {$baseURL}/data_release/?uploadSuccess=true");
+    if (!file_exists(__DIR__ . "/../user_uploads/")) {
+        header("HTTP/1.1 400 Bad Request");
+        echo "File upload failed. Default user_uploads directory not found.";
+    } elseif (!is_writable(__DIR__ . "/../user_uploads/")) {
+        header("HTTP/1.1 400 Bad Request");
```
driusan: These seem to be server configuration errors (500), not client errors (400).

They also seem like they're exceptional circumstances since they only happen on misconfigured servers. Is there any reason to not just throw an exception?

ZainVirani:
@Jkat I am unable to merge new changes to aces/minor because of a permission error on the `user_uploads` folder created by this PR 😄

johnsaigle:
> created by this PR

@ZainVirani This PR doesn't create a `user_uploads` folder. May be a problem on your side. 😄

PR #3622, opened 2018-05-01
Survey accounts empty instruments fix

intralizee:
This pull request `Fix Warning: Undefined variable: instrument in Utility.class.inc`.

See also: `Task #14375`

### DIFF
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -725,12 +725,16 @@ class Utility
      * This should return an array in a format suitable for addSelect() from
      * NDB_Page
      *
-     * @return associative array of test_names in a Test_Name => "Full Name"
+     * @return array $instruments associative array of test_names in a Test_Name => "Full Name"
      */
     static function getDirectInstruments()
     {
         $DB            =& Database::singleton();
+<<<<<<< HEAD
         $instruments   = array();
+=======
+        $instruments = [];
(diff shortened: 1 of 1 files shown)


php/libraries/Utility.class.inc
```diff
@@ -725,11 +725,12 @@ class Utility
      * This should return an array in a format suitable for addSelect() from
      * NDB_Page
      *
-     * @return associative array of test_names in a Test_Name => "Full Name"
+     * @return array $instruments associative array of test_names in a Test_Name => "Full Name"
```
driusan: I don't think the variable name is meaningful in the context of defining the return, it's a meaningless internal detail, but other than that +1

zaliqarosli:
linking this https://redmine.cbrain.mcgill.ca/issues/14375 for accessibility

intralizee:
Seems like a PR was merged that solves this bug and so closing it.

ridz1208:
This PR seems to only be moving a line downwards... approving anyways

driusan:
I don't see how this would fix any warning. It seems to just be moving a variable declaration from the top of the function to slightly further down?

intralizee:
@driusan If you checkout the bugfix branch you will notice the variable not initiated and I must have incorrectly rebased before which is why I just moved the variable so git would notice a file change.

PR #3623, opened 2018-05-01
[Media] removed html line breaks from markdown file

victoriafoing:
This pull request removes the html line breaks <br/> from the Setup markdown file in the media module.

### DIFF
--- a/modules/media/help/Setup.md
+++ b/modules/media/help/Setup.md
@@ -14,15 +14,20 @@ Ubuntu: $GROUP$ = www-data
 CentOS: $GROUP$ = apache
 ```
 
-To find the `user` of your web server, run:<br/> 
+To find the `user` of your web server, run:
+ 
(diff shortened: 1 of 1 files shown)


modules/media/help/Setup.md
```diff
@@ -14,15 +14,20 @@ Ubuntu: $GROUP$ = www-data
...
-To see if your web server's user or group owns the upload path, run:<br/> 
+To see if your web server's user or group owns the upload path, run:
+ 
 `ls -ld /data/uploads | awk '{ print "user:" $3 ", group:" $4 }'`
 
-If neither owns the folder, you should run the following two commands:<br/>
-`sudo chown <unix-user>:<web-server-group> /data/uploads`<br/>
+If neither owns the folder, you should run the following two commands:
+
+`sudo chown <unix-user>:<web-server-group> /data/uploads`
```
driusan: 1. Why is one of these sudo'd but not the other?
2. They can probably be grouped together into one ``` block.

PR #3624, opened 2018-05-01
[Instrument Manager] Sanitize instrument table name before using in exec

johnsaigle:
See also: #3612, Redmine 14377


modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -111,7 +111,7 @@ class Instrument_Manager extends \NDB_Menu_Filter
                     " -p" . escapeshellarg($db_config['quatPassword']).
                     " " . escapeshellarg($db_config['database']).
                     " < " . $this->path . "project/tables_sql/".
-                    $instrument->table.".sql"
+                    escapeshellarg($instrument->table).".sql"
```
driusan: the closing bracket is wrong, the argument being escaped includes the ".sql" part of the filename.

johnsaigle:
@driusan thanks for the comment. I made a fix accordingly.

PR #3625, opened 2018-05-01
[Instrument Manager] Fix warnings related to file access permissions

johnsaigle:
This pull request fixes a bunch of warnings that appear if instruments are not readable by the web user.

### DIFF
--- a/modules/instrument_manager/php/instrument_manager.class.inc
+++ b/modules/instrument_manager/php/instrument_manager.class.inc
@@ -275,8 +275,8 @@ class Instrument_Manager extends \NDB_Menu_Filter
                 return '? (File not found in filesystem)';
             }
         }
-        $fp       = fopen($filename, "r");
-        $db       = \Database::singleton();
+        $fp = fopen($filename, "r");
+        $db = \Database::singleton();
 
         while (($line = fgets($fp, 4096)) !== false) {
             $pieces = explode("{@}", $line);
@@ -367,8 +367,8 @@ class Instrument_Manager extends \NDB_Menu_Filter
                 return '? (File not found in filesystem)';
             }
         }
-        $fp       = fopen($filename, "r");
-        $db       = \Database::singleton();
+        $fp = fopen($filename, "r");
+        $db = \Database::singleton();
 
         while (($line = fgets($fp, 4096)) !== false) {
             $pieces       = explode("{@}", $line);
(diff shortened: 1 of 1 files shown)


modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -71,7 +71,8 @@ class Instrument_Manager extends \NDB_Menu_Filter
             $writable = true;
         }
         $this->tpl_data['writable'] = $writable;
-        if ($writable && $_POST['install'] == 'Install Instrument') {
+        $install = $_POST['install'] ?? '';
+        if ($writable && $install === 'Install Instrument') {
```
johnsaigle: This created PHP notice when not using POST.

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -267,8 +268,15 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
         $filename = $this->path . "project/instruments/$instname.linst";
-        $fp       = fopen($filename, "r");
-        $db       = \Database::singleton();
+        if (!is_readable($filename)) {
+            if (file_exists($filename)) {
+                return '? (Insufficient read permissions)';
+            } else {
+                return '? (File not found in filesystem)';
+            }
+        }
```
driusan: This change is very misleading. The "?" is for PHP instruments where it can't determine the status. "Insufficient read permission" should be an error, not an unknown status, and "File not found" is expected for non-LINST instruments. This makes it sound like an error, but it's the expected result for NDB_BVL_Instrument instruments.

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -352,8 +360,15 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
     {
         $filename = $this->path . "project/instruments/$instname.linst";
-        $fp       = fopen($filename, "r");
-        $db       = \Database::singleton();
+        if (!is_readable($filename)) {
+            if (file_exists($filename)) {
+                return '? (Insufficient read permissions)';
+            } else {
+                return '? (File not found in filesystem)';
+            }
```
driusan: Same

driusan:
#3635 was flagged as a dupe of this, and that one was merged instead since it doesn't make other unrelated changes.

PR #3628, opened 2018-05-02
Reports statistics demographic statistics fix

intralizee:
This pull request `Prevents console warning when going to Reports->Statistics and clicking on the Demographic Statistics when using a fresh installed version of Loris`. It `becomes fixed by checking if $params has been set and prevent undefined property.`.

See also: `Bug #14381`

### DIFF
--- a/modules/statistics/php/stats_demographic.class.inc
+++ b/modules/statistics/php/stats_demographic.class.inc
@@ -28,11 +28,8 @@ class Stats_Demographic extends \NDB_Form
 
     /**
      * The default params
-     *
-     * @var    array
-     * @access private
      */
-    var $params = null;
+    protected $params = null;
 
     /**
      * Checking user's permission
--- a/modules/statistics/php/stats_demographic.class.inc
+++ b/modules/statistics/php/stats_demographic.class.inc
@@ -25,6 +25,15 @@ namespace LORIS\statistics;
  */
 class Stats_Demographic extends \NDB_Form
 {
+
+    /**
+     * The default params
+     *
+     * @var    array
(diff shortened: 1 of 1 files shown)


driusan:
Why does Travis keep not starting on our PRs..

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -227,6 +227,11 @@ class Stats_Demographic extends \NDB_Form
         $this->tpl_data['Subprojects'] = $subprojects;
         $this->tpl_data['Visits']      = $Visits;
 
+        // PREVENT Undefined property Console Output
+        if (!isset($this->params)) {
+            $this->params = null;
```
driusan: Why not just define `params` as a class variable at the start of the class and instantiate it as null so that it's always defined?

intralizee: Way better solution, will do!

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -25,6 +25,15 @@ namespace LORIS\statistics;
...
 class Stats_Demographic extends \NDB_Form
 {
+
+    /**
+     * The default params
+     *
+     * @var    array
+     * @access private
+     */
+    var $params = null;
```
driusan: This is PHP4 style. The `@access private` should be removed, and use the language level private/protected/public keywords instead of `var` in the code rather than a comment.

(This should also probably be protected, not private.)

PapillonMcGill:
LWBTM (look way better to me :-) )

PR #3629, opened 2018-05-02
stats_behavioural.class.inc setup() function may not initiate  and pr…

intralizee:
…event undefined property from ->params

This pull request `fixes console warnings as the $currentProject may not be initiated inside the setup() function in the stats_behavioral.class.inc and added check if $this->params is set to prevent "undefined property" console output`.

See also: `Bug #14382`


johnsaigle:
@ZainVirani are you still working on this (from when I was working on this)? Do[es] your PR(s) address this?

ZainVirani:
@johnsaigle The PRs do not address this, thanks @intralizee

johnsaigle:
This needs `phpcs` as we talked about on Slack 👍

PR #3630, opened 2018-05-02
cleanup console output when CouchDB not configured

intralizee:
This pull request will prevent console output and when the CouchDB port not being a number or the host hasn't been configured. 

The user may not have configured the port after installing Loris. Fixed by preventing the error message when the function _getRelativeURL occurs inside the CouchDB.class.inc class.  A LorisException is thrown to show the error and no console warning is displayed.

See also: `Bug #14383`

### DIFF
--- a/php/libraries/CouchDB.class.inc
+++ b/php/libraries/CouchDB.class.inc
@@ -70,14 +70,7 @@ class SocketWrapper
      */
     function open()
     {
-        if (strpos($this->host, "%") !== false) { // %COUCH_HOSTNAME%
-            $this->socket = false;
-        }
-        if (!is_numeric($this->port)) { // %COUCH_PORT%
-            $this->socket = false;
-        } else {
-            $this->socket = fsockopen($this->host, $this->port);
-        }
+        $this->socket = fsockopen($this->host, $this->port);
         if ($this->socket === false) {
             throw new \LorisException(
                 "Could not connect to CouchDB server at $this->host:$this->port"
@@ -375,6 +368,8 @@ class CouchDB
      */
     function _getRelativeURL($url, $op = 'GET')
     {
+        $this->validHostAndPort();
+
         $handler = $this->SocketHandler;
         $handler->setHost($this->_host);
         $handler->setPort($this->_port);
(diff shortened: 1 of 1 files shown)


php/libraries/CouchDB.class.inc
```diff
@@ -70,7 +70,14 @@ class SocketWrapper
...
     {
-        $this->socket = fsockopen($this->host, $this->port);
+        if (strpos($this->host, "%") !== false) { // %COUCH_HOSTNAME%
+            $this->socket = false;
+        }
+        if (!is_numeric($this->port)) { // %COUCH_PORT%
+            $this->socket = false;
+        } else {
+            $this->socket = fsockopen($this->host, $this->port);
+        }
```
driusan: What are you trying to accomplish with this? Opening a connection to a CouchDB server that isn't configured is meaningless and this just pushes the error further downstream when someone tries to do the next operation on the socket.

intralizee: Just a console output cleanup. The throw new LorisException does the job by showing the error:
> The following errors occured while attempting to display this page:
>
> Could not connect to CouchDB server at %COUCH_HOSTNAME%:%COUCH_PORT%

A default untouched configuration (where the CouchDB hasn't been configured) will display the console warning: fsockopen() expects parameter 2 to be integer, string given on line 73. As well with the LorisException.

PapillonMcGill: Could not that be done just by setting proper error display in php?

driusan: Where is it trying to open the connection? If something is throwing an error out of the box, this should be checked at the call site before trying to open the connection, not in the CouchDB wrapper.

intralizee: Inside the [dataquery] module, php/dataquery.class.inc has a setup() function where on ( line 54 ) the queryView automatically runs the open() function of the SocketWrapper. 

The goal was to hide the extra console warning being displayed because the LorisException prints a message to the page about how the database hasn't been configured. 

I'm not certain if a check to see if the database isn't using the default installation values in the setup() of dataquery.class.inc is the best approach. How should I proceed with it?

intralizee:
Moved the hostname & port checks outside the wrapper as advised. The checks are now inside _getRelativeURL.

ridz1208:
@intralizee not sure about these changes. it seems a bit weak/hacky (checking for a %) maybe we could benefit from a `validHostname()` function ?? 

you might also want to update your description since it doesnt seem to be adding an @ anymore

intralizee:
Added the suggestions @ridz1208 proposed. Used built in php hostname valid check instead of checking for the % symbol and created the the function validHostAndPort().

kongtiaowang:
This PR also fixes the 500 error bug in latest 20.0 without CouchDB setting properly.

driusan:
I re-targetted this to 20.0 as per @kongtiaowang's comment. Can you rebase?

intralizee:
@driusan I did the rebase.

php/libraries/CouchDB.class.inc
```diff
@@ -735,5 +737,27 @@ class CouchDB
...
+     * @throws LorisException
+     *
+     * @return void
+     */
+    function validHostAndPort()
+    {
+        $config       = NDB_Config::singleton();
+        $couch_config = $config->getSetting('CouchDB');
+        if (!filter_var(gethostbyname($couch_config['hostname']), FILTER_VALIDATE_IP)
+            || !is_numeric($couch_config['port'])
```
driusan: what is port isn't set? is_numeric seems to be falsey in PHP. Maybe this should coalesce to the default CouchDB port?

PR #3603, opened 2018-04-18
[Doc Repo] Enhancements to readability and logging for edit/upload ajax file

johnsaigle:
This pull request touches up the doc repo to enhance its readability, error reporting, and performance.

I have self-reviewed to provide feedback on specific points. 

### To Test

* Everything should work the same as before. Try basic doc repo functions and make sure nothing is broken.

PR #3632, opened 2018-05-02
initiated variables to fix console log warnings

intralizee:
This pull request `If viewing Document Repository in Tools on Loris, the console will output warnings on a fresh install`. The `variables needed to be initiated in some cases`.

See also: `Bug #14386`

### DIFF
--- a/modules/document_repository/php/document_repository.class.inc
+++ b/modules/document_repository/php/document_repository.class.inc
@@ -78,10 +78,9 @@ class Document_Repository extends \NDB_Menu_Filter
                           'v.Data_dir',
                           'v.Date_uploaded',
                          );
-
-        $openAccordion = $_GET['openAccordion'];
+        $openAccordion = isset($_GET['openAccordion']) ?? false;
         $this->tpl_data['openAccordion'] = $openAccordion;
-        $filtered = $_GET['filtered'];
+        $filtered = isset($_GET['filtered']) ?? null;
         $this->tpl_data['filtered'] = $filtered;
         $this->query        = $query;
         $this->group_by     = '';
@@ -172,7 +171,13 @@ class Document_Repository extends \NDB_Menu_Filter
         $this->addBasicText('version', 'Version:');
 
         $this->addSelect('File_type', 'File type:', $fileTypes);
-        $this->addSelect('File_category', 'File category:', $fileCategories);
+        // Note: $fileCategories = $fileCategories ?? array()
+        // handles the case where $fileCategories hasn't been initiated.
+        $this->addSelect(
+            'File_category',
+            'File category:',
+            $fileCategories            = $fileCategories ?? array()
+        );
         $this->tpl_data['Sites']       = $list_of_sites;
         $this->tpl_data['Instruments'] = $list_of_instruments;
         $user   =& \User::singleton();
@@ -310,6 +315,7 @@ class Document_Repository extends \NDB_Menu_Filter
         $query    = "SELECT * FROM document_repository_categories".
                     " ORDER BY parent_id ASC";
         $dbresult = $DB->pselect($query, array());
+        $tree     = new \stdClass();
         foreach ($dbresult as $row) {
             $results[] =$row;
             $tree      = null;
@@ -340,6 +346,7 @@ class Document_Repository extends \NDB_Menu_Filter
         // as json and then decoding it into a php associative array
         $tmpIDCategories  = json_decode(json_encode($tree), true);
         $categoryOrdering = $this->parseTree($tmpIDCategories);
+        $fileCategories   = array();
 
         foreach ($categoryOrdering as $order) {
             foreach ($IDCategories as $key=>$value) {


modules/document_repository/php/document_repository.class.inc
```diff
@@ -78,10 +78,9 @@ class Document_Repository extends \NDB_Menu_Filter
                           'v.Data_dir',
                           'v.Date_uploaded',
                          );
-
-        $openAccordion = $_GET['openAccordion'];
+        $openAccordion = isset($_GET['openAccordion']) ? $_GET['openAccordion'] : '';
         $this->tpl_data['openAccordion'] = $openAccordion;
-        $filtered = $_GET['filtered'];
+        $filtered = isset($_GET['filtered']) ? $_GET['filtered'] : '';
```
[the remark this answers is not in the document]
intralizee: Awesome! I did not know and will definitely use.

modules/document_repository/php/document_repository.class.inc
```diff
@@ -166,6 +165,9 @@ class Document_Repository extends \NDB_Menu_Filter
             $fileTypes[$fileType['File_type']] = $fileType['File_type'];
         }
         $fileTypes = array_merge(array(null => 'Any') + $fileTypes);
+        if (!isset($fileCategories)) {
```
xlecours: I don't think `$fileCategories` is set in any scenario at this point. Could we pass an empty array to `$this->addSelect('File_category', 'File category:', $fileCategories);` on line 177 ?

modules/document_repository/php/document_repository.class.inc
```diff
@@ -310,6 +309,7 @@ class Document_Repository extends \NDB_Menu_Filter
         $query    = "SELECT * FROM document_repository_categories".
                     " ORDER BY parent_id ASC";
         $dbresult = $DB->pselect($query, array());
+        $tree     = array();
```
johnsaigle: See https://github.com/aces/Loris/pull/3656. It looks like this should be an object, not an array.

driusan: Yes, see line 326. Its properties are accessed as an object.

intralizee: Updated the $tree to be an object as in #3656

modules/document_repository/php/document_repository.class.inc
```diff
@@ -78,10 +78,9 @@ class Document_Repository extends \NDB_Menu_Filter
                           'v.Data_dir',
                           'v.Date_uploaded',
                          );
-
-        $openAccordion = $_GET['openAccordion'];
+        $openAccordion = isset($_GET['openAccordion']) ?? '';
```
driusan: This idiom doesn't make sense. 

isset returns a bool true or false, never null, so the `?? ''` null coalesce can never be reached, and even if it could be reached, it wouldn't really make sense for the possible values of the variable to be either true, false, or the empty string.

modules/document_repository/php/document_repository.class.inc
```diff
@@ -310,6 +309,7 @@ class Document_Repository extends \NDB_Menu_Filter
         $query    = "SELECT * FROM document_repository_categories".
                     " ORDER BY parent_id ASC";
         $dbresult = $DB->pselect($query, array());
+        $tree     = new \stdClass();
```
driusan: this is already done on major

modules/document_repository/php/document_repository.class.inc
```diff
@@ -172,7 +171,7 @@ class Document_Repository extends \NDB_Menu_Filter
         $this->addBasicText('version', 'Version:');
 
         $this->addSelect('File_type', 'File type:', $fileTypes);
-        $this->addSelect('File_category', 'File category:', $fileCategories);
+        $this->addSelect('File_category', 'File category:', array());
```
driusan: shouldn't this dropdown have options?

modules/document_repository/php/document_repository.class.inc
```diff
@@ -78,10 +78,9 @@ class Document_Repository extends \NDB_Menu_Filter
                           'v.Data_dir',
                           'v.Date_uploaded',
                          );
-
-        $openAccordion = $_GET['openAccordion'];
+        $openAccordion = isset($_GET['openAccordion']) ?? '';
         $this->tpl_data['openAccordion'] = $openAccordion;
-        $filtered = $_GET['filtered'];
+        $filtered = isset($_GET['filtered']) ?? '';
```
zaliqarosli: this is what @johnsaigle did on minor: https://github.com/aces/Loris/pull/3603/commits/cdb3c94c510b69302347413c29f67f61c2c3bef0#diff-9df499b8cbcfa8bd78cf13ca7924b888

modules/document_repository/php/document_repository.class.inc
```diff
@@ -78,9 +78,9 @@ class Document_Repository extends \NDB_Menu_Filter
                           'v.Data_dir',
                           'v.Date_uploaded',
                          );
-        $openAccordion = isset($_GET['openAccordion']) ?? '';
+        $openAccordion = isset($_GET['openAccordion']) ?? false;
```
johnsaigle: The correct idiom here would be 
`$openAccordion = $_GET['openAccordion'] ?? false;`
Same with the one below.

modules/document_repository/php/document_repository.class.inc
```diff
@@ -171,7 +171,7 @@ class Document_Repository extends \NDB_Menu_Filter
         $this->addBasicText('version', 'Version:');
 
         $this->addSelect('File_type', 'File type:', $fileTypes);
-        $this->addSelect('File_category', 'File category:', array());
+        $this->addSelect('File_category', 'File category:', $fileCategories = $fileCategories ?? array() );
```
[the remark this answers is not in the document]
intralizee: If $fileCategories hasn’t been initiated there will be an error.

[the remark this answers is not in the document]
johnsaigle: Could you add a comment explaining this in the code then? I find this line a little difficult to parse.

johnsaigle:
This PR is failing Travis due to PHPCS errors. Please run PHPCS on it.

modules/document_repository/php/document_repository.class.inc
```diff
@@ -78,10 +78,9 @@ class Document_Repository extends \NDB_Menu_Filter
                           'v.Data_dir',
                           'v.Date_uploaded',
                          );
-
-        $openAccordion = $_GET['openAccordion'];
+        $openAccordion = isset($_GET['openAccordion']) ?? false;
         $this->tpl_data['openAccordion'] = $openAccordion;
-        $filtered = $_GET['filtered'];
+        $filtered = isset($_GET['filtered']) ?? null;
```
zaliqarosli: i think this should be `$filtered = $_GET['filtered'] ?? null`

modules/document_repository/php/document_repository.class.inc
```diff
@@ -78,10 +78,9 @@ class Document_Repository extends \NDB_Menu_Filter
                           'v.Data_dir',
                           'v.Date_uploaded',
                          );
-
-        $openAccordion = $_GET['openAccordion'];
+        $openAccordion = isset($_GET['openAccordion']) ?? false;
```
zaliqarosli: and this request for change also still valid https://github.com/aces/Loris/pull/3632/files/f93b152ec814bd2b246164b752c3ea2b703733a3..7b414f0e7f5a60132c8579ac6fdf0d75b4eed867#r194892059

intralizee:
I reverted the changes @zaliqarosli requested because I noticed an error. Although it seems like the error has made its way into the bugfix branch not from this PR. Error is at Database.class.inc:775
Anyone else receiving this error on a clean install of Loris?

modules/document_repository/php/document_repository.class.inc
driusan: Any update on this? (And isn't that also true of the line above?)

intralizee:
https://github.com/aces/Loris/pull/3931

PR #3548, opened 2018-03-14
[User_accounts] Reject users cleanup

zaliqarosli:
This PR cleans up the code merged in https://github.com/aces/Loris/pull/2811 as requested in comments by @driusan at the bottom of the page. This is a new PR to replace https://github.com/aces/Loris/pull/3035 which is now closed because it was too messy.

Functionality should remain the same.

PR #3035, opened 2017-08-09
Reject users cleanup

leothomas:
Incorporated stylistic requests from @driusan on a PR that has been merged (#2811). Functionality remains the same.

PR #2811, opened 2017-05-17
[User accounts] Reject users

leothomas:
Functionality to reject pending user accounts (delete from users table) from front end. Reject_user button is only presented if the user is pending approval and has no password hash (to avoid rejecting active users by setting their approval status to pending) on the edit_user menu. Only people with user_account (admin) permissions have access to this function since the button is populated on edit_users page.

PR #3634, opened 2018-05-02
[User_accounts] Reject users cleanup 

zaliqarosli:
replacing #3548 because i somehow deleted the branch and closed the PR

moved PR to major branch

"This PR cleans up the code merged in #2811 as requested in comments by @driusan at the bottom of the page. This is a new PR to replace #3035 which is now closed because it was too messy.

Functionality should remain the same."

to test:
1. "Request Account" on login page
2. Check user in `users` table with Active='Y', Pending='Y'
2. Reject user on admin account via "Reject User" button in "User Accounts" module
3. Check user deleted from `users` table and there are no errors

### DIFF
--- a/modules/user_accounts/ajax/rejectUser.php
+++ b/modules/user_accounts/ajax/rejectUser.php
@@ -1,122 +1,63 @@
 <?php
 /**
- * Media uploader.
+ * User accounts
  *
- * Handles media upload and update actions received from a front-end ajax call
+ * Handles rejection of pending user accounts by a user that has permissions
  *
- * PHP Version 5
+ * Send DB query to remove user based on userID, checking that the
+ * user sending the request has admin permissions and the account
+ * can be removed (no password hash => has never had an activity on
+ * Loris, and is pending).
+ *
+ * PHP Version 7
  *
  * @category Loris
  * @package  User_Accounts
  * @author   Leo T. <lthomas.mcin@gmail.com>
+ *           Zaliqa Rosli <zaliqa.rosli@mcin.ca>
  * @license  Loris license
  * @link     https://github.com/aces/Loris
  */
 
-    define('NO_IDENTIFIER_SUPPLIED', 1);
-    define('USER_NOT_FOUND', 2);
-    define('INCORRECT_PERMISSION', 3);
-    define('ACCOUNT_ACTIVE', 4);
+namespace LORIS\user_accounts;
 
+// Make sure the permission is checked first to avoid malicious userID scans
+if (!(\User::singleton())->hasPermission('user_accounts')) {
+    header("HTTP/1.1 403 Forbidden");
+    header("Content-Type: text/plain");
+    exit(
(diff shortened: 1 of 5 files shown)


johnsaigle:
@zaliqarosli does this still `Need Work` or should we review/test?

zaliqarosli:
@johnsaigle still needs work - i need to add a heck for when the user is not an instance of the AnonymousUser class

modules/user_accounts/ajax/rejectUser.php
```diff
@@ -42,7 +42,7 @@
 $rejectee = \User::factory($_POST['identifier']);
 $username = $rejectee->getUsername();
 
-if (empty($username)) {
+if (empty($username) || $rejectee instanceof AnonymousUser) {
```
zaliqarosli: @driusan is this what you meant?

zaliqarosli:
@johnsaigle ready for review

johnsaigle:
@zaliqarosli The original PR was issued against `minor` (which feels more appropriate). Was there a reason that this was changed to `major`?

zaliqarosli:
@johnsaigle addition of the AnonymousUser class was made on major not minor. and changes made in this PR is big enough to deserve being in major as advised by @driusan

modules/user_accounts/ajax/rejectUser.php
```diff
@@ -1,122 +1,63 @@
...
-    }
+if (!Edit_User::canRejectAccount($rejectee)) {
+    header("HTTP/1.1 403 Forbidden");
+    header("Content-Type: text/plain");
+    exit(
+        "This account is active and cannot be rejected"
+    );
+}
 
+(\Database::singleton())->delete('users', array("UserID" => $username));
```
ZainVirani: can a not-yet-accepted user have any entries in `user_perm_rel`? if so, delete those too. if not, we good

zaliqarosli: https://github.com/aces/Loris/blob/9cefbcd13844cb44bc4c979eeb9af962676bdbf6/SQL/0000-00-01-Permission.sql#L59

yes, a user pending approval can have entries in `user_perm_rel` but the FK on that table has an on delete cascade clause to it so permissions are automatically deleted along with the entry is `users`

modules/user_accounts/ajax/rejectUser.php
```diff
@@ -1,122 +1,63 @@
...
-    // @codingStandardsIgnoreStart
-    function reject_user($identifier)
-    {
-    // @codingStandardsIgnoreEnd
-        $DB      = Database::singleton();
-        $config  = NDB_Config::singleton();
-        $baseURL = $config->getSetting('url');
-
-        $redirect = $baseURL . "/user_accounts/";
+if (empty($username) || $rejectee instanceof AnonymousUser) {
```
zaliqarosli: @driusan would i need \AnonymousUser here or is this fine?

PR #3640, opened 2018-05-04
[Core] Fix the build on the major branch

driusan:
phan rules were made stricter and fixed on major, but
regressions were introduced when minor (where some things
weren't fixed) was merged into major after releasing 19.1.

This fixes those regressions on major so that the build
will work again.

### DIFF
--- a/htdocs/installdb.php
+++ b/htdocs/installdb.php
@@ -51,6 +51,7 @@ require_once __DIR__ . "/../php/installer/Installer.class.inc";
 // smarty to use when it gets autoloaded.
 require_once __DIR__ . "/../php/installer/Database.class.inc";
 require_once __DIR__ . "/../php/installer/NDB_Config.class.inc";
+use \LORIS\Installer\Database as Database;
 
 $installer = new Installer("../project");
 
--- a/php/installer/Database.class.inc
+++ b/php/installer/Database.class.inc
@@ -14,6 +14,7 @@
  * @link     https://www.github.com/aces/Loris/
  */
 namespace LORIS\Installer;
+use \PDO;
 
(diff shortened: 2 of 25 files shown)


htdocs/installdb.php
```diff
@@ -51,6 +51,7 @@
 // smarty to use when it gets autoloaded.
 require_once __DIR__ . "/../php/installer/Database.class.inc";
 require_once __DIR__ . "/../php/installer/NDB_Config.class.inc";
+use \LORIS\Installer\Database as Database;
```
xlecours: How do this distinguish between Database (\Database) and Database (\LORIS\Installer\Database) ?

php/installer/Database.class.inc
```diff
@@ -13,6 +13,7 @@
  * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
  * @link     https://www.github.com/aces/Loris/
  */
+namespace LORIS\Installer;
```
xlecours: missing `use \PDO;`

driusan: done

htdocs/installdb.php
xlecours: But if this file is not in a namespace, then it is in the '\' by default. The same '\' as in \Database

driusan: the use alias will take precedence (and only the installer version is included, the installdb.php doesn't autoload)

driusan:
restarted because Travis seems to be stuck

driusan:
Merging this without Travis, because the build is failing without this anyways even if something does go wrong..

PR #3641, opened 2018-05-04
[Dashboard] fix links in incomplete forms

ridz1208:
This fixes the links from the dashboard to the incomplete forms page for each site. The `centerID` variable should be in the url but instead the `siteName` is beign passed

i.e.
current link: `http://localhost/statistics/statistics_site/?CenterID=The%20Institue`
should be: `http://localhost/statistics/statistics_site/?CenterID=1`


kongtiaowang:
LGTM

PR #3642, opened 2018-05-04
Phan incorrect param number

driusan:
…g number of parameters.

This fixes both the PhanTooFew and PhanTooMany error classes.

### DIFF
--- a/modules/server_processes_manager/php/serverprocesslauncher.class.inc
+++ b/modules/server_processes_manager/php/serverprocesslauncher.class.inc
@@ -37,17 +37,6 @@ class ServerProcessLauncher
     private $_databaseProvider;
 
     /**
-     * Builds a new ServerProcessesLauncher
-     *
-     * @param IDatabaseProvider $databaseProvider entity used to access the database
-     */
-    public function __construct(
-        IDatabaseProvider $databaseProvider
-    ) {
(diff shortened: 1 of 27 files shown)


modules/server_processes_manager/php/serverprocesslauncher.class.inc
```diff
@@ -36,17 +36,6 @@ class ServerProcessLauncher
...
-     * Builds a new ServerProcessesLauncher
-     *
-     * @param IDatabaseProvider $databaseProvider entity used to access the database
-     */
-    public function __construct(
-        IDatabaseProvider $databaseProvider
-    ) {
-        $this->setDatabaseProvider($databaseProvider);
-    }
-
```
driusan: @nicolasbrossard requested your review for this part.. as far as I can tell, the only location where a ServerProcessesLauncher is instantiated there's no parameter passed.

nicolasbrossard: Since `$databaseProvider` is initialized (lazily) in the getter method, your change is legit. Approved.

driusan:
Trying to trigger travis..

driusan:
@PapillonMcGill or @nicolasbrossard can you re-approve this? I finally got Travis to start on it, but it autodismissed your reviews because I rebased it while trying to get Travis running..

PR #3643, opened 2018-05-07
[candidate_list] Convert candidate_list module to use new data framework

driusan:
This adds candidate_list provisioners, similar to the DicomArchive.


driusan:
Closing this until the Reactification of candidate_list is done, since it's not going anywhere.

PR #3645, opened 2018-05-08
[Login] Remove obsolete captcha code

johnsaigle:
This pull request deletes a legacy file that was used to generate custom captcha challenges. This has been replaced in favour of Google re-captcha.


driusan:
This is already done in #3556, it's just that major hasn't had a release yet to make it back into the minor branch.

PR #3646, opened 2018-05-08
[Client] Add HTTPOnly header to session cookie

johnsaigle:
This pull request enables the usage of the [HTTP-Only](https://www.owasp.org/index.php/HttpOnly#What_is_HttpOnly.3F) and [Secure](https://stackoverflow.com/questions/13729749/how-does-cookie-secure-flag-work) modes for our session cookies, providing defense against certain hacking attempts.

### DIFF
--- a/php/libraries/NDB_Client.class.inc
+++ b/php/libraries/NDB_Client.class.inc
@@ -133,7 +133,10 @@ class NDB_Client
             . $config_additions
         );
         // start php session
-        session_start();
+        session_start([
+            'cookie_httponly' => true,
+            'cookie_secure' => true
+        ]);
 
         // if exiting, destroy the php session
         if (isset($_REQUEST['logout']) && $_REQUEST['logout']) {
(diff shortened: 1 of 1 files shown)


php/libraries/NDB_Client.class.inc
```diff
@@ -133,15 +133,25 @@ class NDB_Client
             . $config_additions
         );
         // start php session
```
kongtiaowang: ini_set('session.cookie_httponly', '1');
ini_set('session.use_strict_mode', '1');
session_start();

kongtiaowang: For the session.cookie_secure that failed in Docker test suits, we can't modify the apache server "Add following entry in httpd.conf
Header edit Set-Cookie ^(.*)$ $1;HttpOnly;Secure"

johnsaigle: We should modify e.g. Docker (probably Vagrant) to include these settings server-side. However for users who do not install users this way it's better to use the application code to guarantee security rather than rely on the server configuration.

PR #3647, opened 2018-05-08
SQL injection when creating instrument by linst form

intralizee:
This pull request `SQL injection possible by `superadmin` users via a malicious LINST form`. It `uses prepared statements, sanitizes the string for creating a table, checks if table already exists and adding logging with display message to user if error.`.

See also: `Bug #14390 and #8458`

### DIFF
--- a/modules/instrument_manager/php/instrument_manager.class.inc
+++ b/modules/instrument_manager/php/instrument_manager.class.inc
@@ -93,7 +93,9 @@ class Instrument_Manager extends \NDB_Menu_Filter
                     'php generate_tables_sql_and_testNames.php < '
                     . escapeshellarg($new_file)
                 );
-                if (!empty($error_message) && strpos($error_message, "Instrument") !== false) {
+                if (!empty($error_message)
+                    && strpos($error_message, "Instrument") !== false
+                ) {
                     // Instrument(s)
                     throw new \LorisException(
                         "Error uploading the Instrument(s) file:\n".
--- a/modules/instrument_manager/php/instrument_manager.class.inc
+++ b/modules/instrument_manager/php/instrument_manager.class.inc
@@ -89,21 +89,20 @@ class Instrument_Manager extends \NDB_Menu_Filter
                 array('testname' => $_FILES['install_file']['name'])
             );
             if ($exists <= 0) {
-                $db_config = $config->getSetting('database');
                 exec(
                     'php generate_tables_sql_and_testNames.php < '
-                    . escapeshellarg($new_file)
+                    . escapeshellarg($new_file),
+                    $error_message
                 );
-
-                $instrument = \NDB_BVL_Instrument::factory(
-                    $instname,
-                    '',
-                    ''
-                );
-                if (!isset($db_config['quatUser'], $db_config['quatPassword'])) {
+                if ($error_message) {
                     throw new \LorisException(
-                        "SQL needs to be sourced manually at /project/tables_sql/".
-                        $instrument->table.".sql"
+                        "Error uploading the Instrument(s) file!\n".
+                        "Details: $error_message"
+                    );
+                    // See the /project/tables_sql/{file} for debugging the error.
+                    error_log(
+                        "Error uploading the Instrument(s) file: "
+                        .$error_message
                     );
                 }
             }
--- a/modules/instrument_manager/php/instrument_manager.class.inc
+++ b/modules/instrument_manager/php/instrument_manager.class.inc
@@ -89,14 +89,12 @@ class Instrument_Manager extends \NDB_Menu_Filter
                 array('testname' => $_FILES['install_file']['name'])
             );
             if ($exists <= 0) {
-                exec(
+                $error_message = exec(
                     'php generate_tables_sql_and_testNames.php < '
-                    . escapeshellarg($new_file),
-                    $error_message
+                    . escapeshellarg($new_file)
                 );
-                $error_message = implode($error_message);
-                if (!empty($error_message)) {
-                    //echo $error_message;
+                if (!empty($error_message) && strpos($error_message, "Instrument") !== false) {
+                    // Instrument(s)
                     throw new \LorisException(
                         "Error uploading the Instrument(s) file:\n".
                         "$error_message"
--- a/modules/instrument_manager/php/instrument_manager.class.inc
+++ b/modules/instrument_manager/php/instrument_manager.class.inc
@@ -2,7 +2,7 @@
 /**
  * Instrument_manager
  *
- * PHP Version 5
+ * PHP Version 7
  *
  * @category Main
  * @package  Instrument_Manager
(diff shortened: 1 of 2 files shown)


modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -88,7 +88,13 @@ class Instrument_Manager extends \NDB_Menu_Filter
             );
             if ($exists <= 0) {
                 $db_config = $config->getSetting('database');
-                exec("php generate_tables_sql_and_testNames.php < $new_file");
+                exec(
+                    'php generate_tables_sql_and_testNames.php < '
+                    . escapeshellarg($new_file),
```
johnsaigle: This line has already been dealt with as of https://github.com/aces/Loris/pull/3612

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -88,7 +88,13 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
+                    $output,
+                    $return_var
+                );
+                print_r($output);
```
johnsaigle: I don't think we need to print the output.

tools/generate_tables_sql_and_testNames.php
```diff
@@ -23,96 +30,210 @@
...
-                $output.="`$bits[1]` $bits[0] default NULL,\n";
+        //generate specific column definitions for specific types of HTML elements
+        default:
+            if ($bits[1] == "") {
+                continue;
+            }
+            if ($bits[0]=="select") {
+                $bits[0] = enumizeOptions(
+                    $bits[3],
+                    isset($tablename)
```
johnsaigle: `??` idiom would be better here

tools/generate_tables_sql_and_testNames.php
```diff
@@ -23,96 +30,210 @@
...
 }
-
+/**
+ * The sanitizeSqlString function
+ *
+ * @param string $name to sanitize.
+ *
+ * @return string
+ */
+function sanitizeSqlString($name)
```
johnsaigle: I'm not sure how this sanitizes a string, exactly. Also, I think it's better not to reinvent the wheel here and instead used prepared statements -- at the very least it is probably better to use built in PHP functions to do this.

intralizee: Unable to bind "table names" or "column names" when using prepared statements. If attempted the column names will have single quotes. I'm using the sanitizeSqlString for when creating the table with the user supplied input for that part.

tools/generate_tables_sql_and_testNames.php
intralizee: Awesome fixed.

tools/generate_tables_sql_and_testNames.php
```diff
@@ -2,17 +2,24 @@
 <?php
 
 /**
- * generate_tables_sql.php takes the ip_output.txt file generated from
+ * Behavioural
+ *
+ * PHP Version 5
```
johnsaigle: Should be PHP 7 now that I got you to use `??` 😛

tools/generate_tables_sql_and_testNames.php
johnsaigle: I think I'd prefer in this case to use the custom `escape` function in the DB. It serves the same purpose as this function and follows the standard set by `mysql_real_escape_string` and so will be simpler to maintain going forward. https://github.com/aces/Loris/blob/minor/php/libraries/Database.class.inc#L1315

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -88,7 +88,10 @@ class Instrument_Manager extends \NDB_Menu_Filter
             );
             if ($exists <= 0) {
                 $db_config = $config->getSetting('database');
-                exec("php generate_tables_sql_and_testNames.php < $new_file");
+                exec(
+                    'php generate_tables_sql_and_testNames.php < '
+                    . escapeshellarg($new_file)
```
johnsaigle: This is going to conflict with the PR I linked earlier so I think you should rebase.

tools/generate_tables_sql_and_testNames.php
johnsaigle: I believe it should read `$tablename ?? null`. No `isset`

tools/generate_tables_sql_and_testNames.php
```diff
@@ -23,96 +30,210 @@
...
+    $output .= "PRIMARY KEY  (`CommentID`)\n) ENGINE=InnoDB DEFAULT CHARSET=utf8;\n";
+    $fp      = fopen($filename, "w");
+    fwrite($fp, $output);
+    fwrite(
+        $fp,
+        "REPLACE INTO test_names (Test_name, Full_name, Sub_group) "
+        ."VALUES ('$tablename', '$title', 1);\n"
+    );
+
+    foreach ($pages as $pageNo => $page) {
```
johnsaigle: A comment would be nice here explaining that this writes a record of the intended changes to a file but the file is not actually sourced directly (as was the previous behaviour)

modules/instrument_manager/php/instrument_manager.class.inc
intralizee: Did a rebase with adding escape function in the DB and added the comment suggested.

intralizee:
Added bug #8458 as just needed to adding logging and error message to user.

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -89,32 +89,24 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
+                    $error_message
                 );
-
-                $instrument = \NDB_BVL_Instrument::factory(
-                    $instname,
-                    '',
-                    ''
-                );
-                if (!isset($db_config['quatUser'], $db_config['quatPassword'])) {
+                $error_message = implode($error_message);
```
johnsaigle: This looks like array-string conversion and should probably be avoided.

This could probably be better written as
```php
$output = shell_exec(
    'php generate_tables_sql_and_testNames.php < '
    . escapeshellarg($new_file)
);
```

Note that both `exec` and `shell_exec` return all output from the external command. You should be sure that `generate_tables_sql_and_testNames.php` is silent on success if you are going to treat any non-empty string as an error message.

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,249 @@
...
+                    $bits[0] = "text";
+                } else if ($bits[0] == "text") {
+                    $bits[0] = "varchar(255)";
+                } else if ($bits[0] == "checkbox") {
+                    $bits[0] = "varchar(255)";
+                } else if ($bits[0] == "static") {
+                    $bits[0] = "varchar(255)";
+                } else {
+                    // Safety SQL injection prevention.
+                    $bits[0] = str_replace('`', '', $db->escape($bits[0]));
```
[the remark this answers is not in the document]
intralizee: I actually self tested without the str_replace and result with double backticks and when used with the sql prepared statements. The str_replace is necessary to not have ``example``.

PR #3649, opened 2018-05-08
[User] Remove lingering uses of mt_rand

johnsaigle:
`mt_rand` generates predictable values. There are lingering references to it in this function.

This PR switches everything over to using `random_int` once and for all.

See also https://github.com/aces/Loris/pull/3045 in which I should have done this in the first place... 😅

### DIFF
--- a/php/libraries/User.class.inc
+++ b/php/libraries/User.class.inc
@@ -398,13 +398,13 @@ class User extends UserPermissions
         if (strpbrk($password, $numbers) === false) {
             $password .= substr(
                 $numbers,
-                mt_rand(0, strlen($numbers)-1),
+                random_int(0, strlen($numbers)-1),
                 1
             );
         }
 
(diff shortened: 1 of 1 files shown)


php/libraries/User.class.inc
```diff
@@ -398,13 +398,13 @@ class User extends UserPermissions
         if (strpbrk($password, $numbers) === false) {
             $password .= substr(
                 $numbers,
-                mt_rand(0, strlen($numbers)-1),
+                random_int(0, strlen($numbers)-1),
```
xlecours: So this is using random_int to get a number from 0 to 9 that would be the random index of a character in the string "0123456789"? Could this whole block be replace by ```
$password .= random_int(0,9);
```

johnsaigle: Yes!

johnsaigle: Done

PR #3579, opened 2018-04-11
introduce the visit and visit_projet_subproject_rel tables for

PapillonMcGill:
normalizing visit to 3nf. Take current visits label from Visit_Windows
session table to populate visit table.

This pull request `add a visit table as well as visit_project_subproject_rel table to track which visit are included in project or subproject`. It also take current visit label in Visit_Windows and session tables to populate visit. 

The VisitTest.php files is a unit test that will be remove before merge. It require version 6 of PHPUnit and a 
`testConfig.inc` files with your DB credential (username, password, host and database variable)

require the php/libraries/Database.class.inc from https://github.com/aces/Loris/pull/3551.

Future PR will modify tables to replace visit label with the visit.VisitID attributes as foreign key and code accordingly.

PR #3652, opened 2018-05-09
added function to get visit label to instrument class

jesscall:
This pull request adds a function to get the visit label of a given timepoint. It will be useful in instances where certain instrument information is different (ex:headers) for different visits.

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1408,8 +1408,8 @@ class NDB_BVL_Instrument extends NDB_Page
         $sessionID = $this->getSessionID();
 
         if (!empty($sessionID)) {
-            $t          = Timepoint::singleton($sessionID);
-            $visitLabel = $t->getVisitLabel();
+            $timepoint  = \Timepoint::singleton($sessionID);
+            $visitLabel = $timepoint->getVisitLabel();
             return $visitLabel;
         }
         return '';
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1398,6 +1398,22 @@ class NDB_BVL_Instrument extends NDB_Page
         return false;
     }
 
+    /**
+     * Gets the VisitLabel for the timepoint to which this instrument pertains
+     *
+     * @return String for the VisitLabel
+     */
+    function getVisitLabel{
+        $sessionID = $this->getSessionID();
+
+        if(!empty($sessionID)){
+            $t = Timepoint::singleton($sessionID);
+            $visitLabel = $t->getVisitLabel();
+            return $visitLabel;
+        }
+        return false;
+    }
+
 
     /**
      * Determines what the data entry status flag should be set to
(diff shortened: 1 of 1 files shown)


php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1398,6 +1398,22 @@ class NDB_BVL_Instrument extends NDB_Page
         return false;
     }
 
+    /**
+     * Gets the VisitLabel for the timepoint to which this instrument pertains
+     *
+     * @return String for the VisitLabel
+     */
+    function getVisitLabel{
```
driusan: this isn't syntactically valid PHP. You need the argument list.

(You should also run phpcs on it or Travis is going to fail. You can just run `npm run lint:php` as a shortcut to run PHPCS and `php -l`.)

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1398,6 +1398,22 @@ class NDB_BVL_Instrument extends NDB_Page
...
+        $sessionID = $this->getSessionID();
+
+        if(!empty($sessionID)){
+            $t = Timepoint::singleton($sessionID);
+            $visitLabel = $t->getVisitLabel();
+            return $visitLabel;
+        }
+        return false;
```
driusan: false isn't a string

um4r12:
LGTM

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1398,6 +1398,23 @@ class NDB_BVL_Instrument extends NDB_Page
...
+     * Gets the VisitLabel for the timepoint to which this instrument pertains
+     *
+     * @return String for the VisitLabel
+     */
+    function getVisitLabel()
+    {
+        $sessionID = $this->getSessionID();
+
+        if (!empty($sessionID)) {
+            $t          = Timepoint::singleton($sessionID);
```
zaliqarosli: i am wondering if this should be namespaced? i.e. \Timepoint::singleton($sessionID)?

zaliqarosli: i am wondering if this should be namespaced? i.e. \Timepoint::singleton($sessionID)?

jesscall:
@kongtiaowang Can you help with this? not sure why travis is failing

kongtiaowang:
Travis is back!

driusan:
Thinking on this a little more, I'm a little opposed to this because I think it would be better to have something that returns a typed visit object rather than a string. Adding this to minor would encourage people to use this without the safety of type checking.

On the other hand, the "Visit" object is part of #3579 and I don't know what the timeline on that getting merged is.

driusan:
@xlecours any thoughts on the above comment?

PapillonMcGill:
@driusan @jesscall @xlecours 
the getVisitLabel in the TimePoint object already return a string, so this just extend on it. 
Of course it would be better to return an object as we will be able to use the getName() function of the object.

As for the [3579](https://github.com/aces/Loris/pull/3579/files), as you know my schedule is a little bit upside down. Can't promise anything now but hope soon. Worse case, we could modify all use of visit afterwards.

driusan:
There doesn't seem to be any movement on the other one, so I'll merge this, but we should deprecate this once we have an object to use instead.

PR #3656, opened 2018-05-10
[document_repository] Fix incorrectly initialized variable

driusan:
A variable in the document repository was initialized as null, but then treated as an object. This changes it to be initialized as an object.

Fixes the only PhanTypeExpectedObjectPropAccessButGotNull error.

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -16,7 +16,6 @@ return [
 	"scalar_implicit_cast" => false,
 	"ignore_undeclared_variables_in_global_scope" => false,
 	"suppress_issue_types" => [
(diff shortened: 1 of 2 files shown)


johnsaigle:
Hey @driusan, @intralizee addressed this problem in the PR referenced above. However, in that PR the variable was initialized as `array()`.

Just wanted to drop a note here so that these PRs can be made harmonious.

HenriRabalais:
@driusan should this PR be removed because of PR #3632 ? Or should this change be remove from the referenced PR?

driusan:
#3632 will likely be released before the next major release, so closing this.

PR #3657, opened 2018-05-10
[media] unathenticated viewing private data fix

intralizee:
This pull request `fixes private data from being viewed`. It `adds permission check for user accessing the getData in FileUpload.php`.

See also: `Bug #13847`

### DIFF
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -16,12 +16,27 @@
 if (isset($_GET['action'])) {
     $action = $_GET['action'];
     if ($action == "getData") {
-        echo json_encode(getUploadFields());
+        viewData();
     } else if ($action == "upload") {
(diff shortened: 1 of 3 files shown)


johnsaigle:
Hey @intralizee it looks like you have some unrelated commits on this branch. This will have to be resolved.

modules/media/ajax/FileUpload.php
```diff
@@ -16,14 +16,29 @@
...
         editFile();
     }
 }
 
+/**
+ * Handles the media view data process
+ *
+ * @return void
+ */
+function viewData()
```
johnsaigle: Could you move this function declaration to the bottom of the files with the other ones? It feels strange to have the functions in different places

intralizee: fixed

PR #3658, opened 2018-05-10
[InstallDB] Modifying the check to see whether http or https protocol is being used.

um4r12:
This pull request attempts to determine whether the http or https protocol is being used. It modifies the `getBaseURL()` function to check if `$_SERVER['HTTP_X_FORWARDED_PROTO']` is set to `https` rather than `$_SERVER['HTTPS']`.

### DIFF
--- a/php/installer/Installer.class.inc
+++ b/php/installer/Installer.class.inc
@@ -273,9 +273,13 @@ class Installer
         // This is apparently the most reliable way to figure out if
         // the requery is over http or https.. $_SERVER[REQUEST_SCHEME]
         // is not reliable.
-        $RequestScheme = isset($_SERVER['HTTP_X_FORWARDED_PROTO'])
-                            && $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https'
-                            ? 'https' : 'http';
+        $RequestScheme = 'http';
+        if ((isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] === 'on')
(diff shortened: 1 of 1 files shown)


php/installer/Installer.class.inc
```diff
@@ -273,7 +273,8 @@ class Installer
         // This is apparently the most reliable way to figure out if
         // the requery is over http or https.. $_SERVER[REQUEST_SCHEME]
         // is not reliable.
-        $RequestScheme = isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] === 'on'
+        $RequestScheme = isset($_SERVER['HTTP_X_FORWARDED_PROTO'])
```
johnsaigle: Perhaps it would be better to check for both?

driusan: It should definitely check for both. HTTP_X_FORWARDED_PROTO is only set if there's a proxy in between the server and PHP (and that proxy is setting the field).. it's not a reliable way to detect the protocol for servers that are directly served by a web server.

driusan: I don't know if PHP provides a way to get the server port, but even if it did it wouldn't be reliable since the server could be configured to run on any port for either http or https.

I think

```
$RequestScheme = 'http';
if (old way of checking || x-forwarded-proto) {
 $RequestSchema = 'https';
}
```

should handle most configurations.

driusan:
It looks good to me, but any word on @ZainVirani's problem? Has it been (better?) tested?

PR #3661, opened 2018-05-10
[LorisForm/Instruments] Format indentation/cleanup of smarty template

zaliqarosli:
fix https://redmine.cbrain.mcgill.ca/issues/11853

removed "lab" and duplicated "form-inline" class names - i don't think they actually do anything.

Before & after:
screenshot 2018-05-10 19 09 55screenshot 2018-05-10 19 09 32


zaliqarosli:
@ridz1208 please test with CCNA instruments!

victoriafoing:
Passed manual tests for various IBIS instruments.

PR #3662, opened 2018-05-11
[document_repository] Fixed multiple files with same name when uploading to document repository

intralizee:
This pull request `Allow files with duplicate name to exist and when user uploads a file to the document repository`. 

I made the mysql column "uuid" in the table "document_repository" for files to be stored in unique directories.

The uuid and file_name are used for the system directory structure when storing the files.
System Directory Structure Examples:

Note: {uuid} is unique per upload.
user_uploads/admin/text.png/{uuid}/text.png
user_uploads/admin/text.txt/{uuid}/text.txt

user_uploads/admin/text.png/{uuid}/text.png
user_uploads/admin/text.txt/{uuid}/text.txt

See also: `Bug #10847 and Bug #11243`

### DIFF
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -186,6 +186,16 @@ CREATE TABLE `psc` (
 
 INSERT INTO `psc` (Name, Alias, Study_site) VALUES ('Data Coordinating Center','DCC', 'Y');
 
+CREATE TABLE `language` (
+  `language_id` integer unsigned NOT NULL AUTO_INCREMENT,
+  `language_code` varchar(255) NOT NULL,
+  `language_label` varchar(255) NOT NULL,
+  PRIMARY KEY (`language_id`),
+  UNIQUE KEY (`language_code`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+INSERT INTO language (language_code, language_label) VALUES ('en-CA', 'English');
+
 CREATE TABLE `users` (
   `ID` int(10) unsigned NOT NULL auto_increment,
   `UserID` varchar(255) NOT NULL default '',
@@ -213,9 +223,11 @@ CREATE TABLE `users` (
   `Password_expiry` date NOT NULL default '1990-04-01',
   `Pending_approval` enum('Y','N') default 'Y',
   `Doc_Repo_Notifications` enum('Y','N') default 'N',
+  `language_preference` integer unsigned default NULL,
   PRIMARY KEY  (`ID`),
   UNIQUE KEY `Email` (`Email`),
-  UNIQUE KEY `UserID` (`UserID`)
+  UNIQUE KEY `UserID` (`UserID`),
+  CONSTRAINT `FK_users_2` FOREIGN KEY (`language_preference`) REFERENCES `language` (`language_id`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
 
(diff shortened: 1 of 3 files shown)


modules/document_repository/ajax/documentEditUpload.php
```diff
@@ -147,4 +170,71 @@
...
+/**
+ * Handles the version inspection process.
+ * Generates correct proceeding version in some cases.
+ *
+ * @param string $version the version to use for uploading file.
+ * @param array  $items   the array of existing files (name & version).
+ *
+ * @return string $version.
+ */
+function versionInspectionGenerator($version, $items)
```
ZainVirani: I don't know if I fully agree with the logic here. It seems like you'd be forcing `\d.\d.(\*|\d)`, which could also incorrectly default to `0.0.1` if there are already files in the user_uploads folder.

I feel like generated IDs are a way better way of handling unique file identifiers rather than version #s (however I get that this isn't ideal for file directories). I also don't think we should force version #s on user files, because they may get confused when a random 0.0.1 shows up.

modules/document_repository/ajax/documentDelete.php
```diff
@@ -49,6 +49,10 @@
     "Select Data_dir from document_repository where record_id =:identifier",
     array(':identifier' => $rid)
 );
+$version  = $DB->pselectOne(
+    "Select version from document_repository where record_id =:identifier",
```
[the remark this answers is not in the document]
intralizee: Awesome will fix!

modules/document_repository/ajax/documentDelete.php
```diff
@@ -67,4 +71,9 @@
     unlink($path);
 }
 
+$rm_directory = __DIR__ . "/../user_uploads/" . $userName . "/"
+                        . str_replace(".", "_", $version);
+
+rmdir($rm_directory);
```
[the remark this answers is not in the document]
intralizee: The rmdir function will only remove the directory if it's empty.

modules/document_repository/ajax/documentEditUpload.php
cmadjar: I feel it should be up to the user to enter the version of a file that they uploaded. If I want to upload a file called hello_toto.pdf and append the version 1.0 to it, it will not consider what I input in the upload file form and number it 0.0.1 which is not what I wanted and is confusing. I think you should use a different field than the version field of the document repository to version the uploaded file itself in case of duplication of files.

intralizee: @cmadjar That definitely makes more sense!

intralizee:
I updated the logic so version number is not needed. See original comment for details.

modules/document_repository/ajax/documentDelete.php
```diff
@@ -31,40 +36,58 @@
...
-    "Select uploaded_by from document_repository where record_id =:identifier",
+    'SELECT uploaded_by FROM document_repository WHERE record_id =:identifier',
     array(':identifier' => $rid)
 );
 $dataDir  = $DB->pselectOne(
-    "Select Data_dir from document_repository where record_id =:identifier",
+    'SELECT Data_dir FROM document_repository WHERE record_id =:identifier',
+    array(':identifier' => $rid)
+);
+$version  = $DB->pselectOne(
```
um4r12: Is `$version` being used anywhere else in the file? What is the purpose for querying the version?

SQL/0000-00-00-schema.sql
```diff
@@ -214,8 +214,7 @@ CREATE TABLE `users` (
   `Pending_approval` enum('Y','N') default 'Y',
   `Doc_Repo_Notifications` enum('Y','N') default 'N',
   PRIMARY KEY  (`ID`),
-  UNIQUE KEY `Email` (`Email`),
-  UNIQUE KEY `UserID` (`UserID`)
+  UNIQUE KEY `Email` (`Email`)
```
ridz1208: @intralizee the changes to this file all over the place I would ask to rebase this PR without any changes to the SQL schema (except necessary ones) and submit  the others in a clean PR

intralizee: Yah a mixup happened. I was on another PR than this one. After I pulled some unnecessary additions from minor when trying to recover the changes. Also with the single quotes you mentioned. :)

modules/document_repository/ajax/documentDelete.php
intralizee: Thanks! I forgot to remove from old code logic.

intralizee:
I'm closing this PR and moving the changes to minor branch.

PR #3666, opened 2018-05-14
[instrument_builder][instrument_manager] Date elements could be added with blank ranges which break data insertion

intralizee:
This pull request `fixes the frontend and backend possibilities of empty date data supplied by the user`. 

It `checks in both the react.question.js and generate_tables_sql_and_testNames.php if the dates supplied are empty`. 

**Note:** The extra code in generate_tables_sql_and_testNames.php is from https://github.com/aces/Loris/pull/3647. See line # 111 to 116 for the new addition of checking if dates are empty.

See also: `Bug #12174`


intralizee:
Moved to minor #3674

PR #3667, opened 2018-05-14
[New Profile] Fixed array index access

ZainVirani:
Code was attempting to access a php constant instead of using a string as array index value.

NOTE: when testing this you may find that new profile is broken due to a function being called on a string. This is a separate issue I will solve in another PR.


johnsaigle:
@ZainVirani if this change would make `new_profile` "more broken" until the related fix is made, you should consider adding the `Blocked` tag

ZainVirani:
@johnsaigle its not more broken, it's just as broken as it was before less one error 😃

PR #3668, opened 2018-05-14
[Site] ensure site is always an object

ZainVirani:


php/libraries/Site.class.inc
```diff
@@ -36,7 +36,9 @@ class Site
     static function &singleton($centerID)
     {
         static $siteList = array();
-        if (!isset($siteList[$centerID])) {
+        if (!isset($siteList[$centerID])
+            || gettype($siteList[$centerID]) != 'object'
+        ) {
```
driusan: I think this should be 2 checks rather than an or, so that it can throw a more explicit exception (and doesn't need to go through the effort of the try if it's not an object.)

ZainVirani: good points -- will make the change

driusan: You're right.. when I wrote the comment I was thinking that select wouldn't throw an exception (and if it did it should just be propagated) since you've validated the parameter, but it's possible that it's still a valid object for an invalid site and if it is it needs to be unset, so the catch still needs to be there.

ZainVirani:
Issue is actually a mis-assigned variable in the new_profile module

PR #3669, opened 2018-05-14
FilterForm bugfix

davidblader:
imaging_uploader isn't rendering on minor currently because of this error:
image

this checks that `value` is an array before checking its length


victoriafoing:
Fixes the loading issue in modules imaging_uploader and data_integrity_flag!

However, affects filter in other modules.
For example, in DICOM_archive when I type a value (e.g. 'L') into Patient ID, it filters the table below. 

l

But then when I erase the value, the table clears.

l erased

In the current version of minor, the table does not clear and returns to its unfiltered form.

jsx/FilterForm.js
```diff
@@ -127,7 +127,7 @@ class FilterForm extends React.Component {
     }
     if (key) {
       filter[key] = {};
-      if (value.length === 0) {
+      if (Array.isArray(value) && value.length === 0) {
```
victoriafoing: need to check that value is type string OR type object so change this to...
if ((typeof (value) === 'string' || typeof (value) === 'object') && value.length === 0)

victoriafoing: i think this fixes the filtering error we found above!

jsx/FilterForm.js
```diff
@@ -127,7 +127,7 @@ class FilterForm extends React.Component {
     }
     if (key) {
       filter[key] = {};
-      if (value.length === 0) {
+      if ((typeof (value) === 'string' || typeof (value) === 'object') && value.length === 0) {
```
ZainVirani: I don't think this will work either. I would replace 130-134 with
```
if (value) {     //all defined/non-null values must have some length, else empty string
  filter[key].value = Object.keys(value).length > 0 ? value : '';
} else {          //null and undefined handled here
  filter[key].value = '';
}
```

victoriafoing: Zain's edit works too

davidblader:
@victoriafoing if everything worked could you add the `Passed Manual Tests` label plz :cowboy_hat_face:

PR #3671, opened 2018-05-16
[Instrument Builder] Disable 'Load Instrument' button when text box is empty

victoriafoing:
This pull request disables the 'load instrument' button when the text box is empty.

This is a small UI/UX improvement addressing the following ticket: https://redmine.cbrain.mcgill.ca/issues/12002

disabled

enabled


modules/instrument_builder/jsx/react.instrument_builder.js
```diff
@@ -17,6 +17,7 @@ var LoadPane = React.createClass({
       // This is used to alert the user if the file was
       // loaded successfully or there was an error with
       // the loading.
+      disabled: true,
```
driusan: The comment above this is now wrong, the comment is for the alert, not the disabled attribute..

victoriafoing: Thanks. I moved it above the comments in my latest commit.

victoriafoing:
Let me know if you think it is unnecessary to set disabled as true in the chooseFile function since it is already set as true in the getInitialState function.
e.g.
this.setState({
      file: value,
      **disabled: true,**
      alert: ''
    });

colezweber:
Looks good on Firefox and Chrome

PR #3674, opened 2018-05-17
[instrument_manager][instrument_builder] SQL injection when creating instrument by linst form and date elements check

maltheism:
This pull request `fixes SQL injection by superadmin users via a malicious LINST form. It uses prepared statements, sanitizes the string for creating a table, checks if table already exists and adding logging with display message to user if error. Also fixes the frontend and backend possibilities of empty date data supplied by the user`.

See also: `Bug #14390, #8458 and #12174 `


modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -87,29 +87,24 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
-                    $instname,
-                    '',
-                    ''
+                $error_message = exec(
+                    'php generate_tables_sql_and_testNames.php < '
+                    . escapeshellarg($new_file)
                 );
-                if (!isset($db_config['quatUser'], $db_config['quatPassword'])) {
+                if (!empty($error_message)
+                    && strpos($error_message, "Instrument") !== false
```
ZainVirani: What is the purpose of this check? If the error doesn't include the word Instrument shouldn't it still be thrown?

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -87,29 +87,24 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
+                ) {
+                    // Instrument(s)
                     throw new \LorisException(
-                        "SQL needs to be sourced manually at /project/tables_sql/".
-                        $instrument->table.".sql"
+                        "Error uploading the Instrument(s) file:\n".
+                        "$error_message"
+                    );
+                    // See the /project/tables_sql/{file} for debugging the error.
+                    error_log(
```
ZainVirani: This error_log may not display if the above exception isn't handled

modules/instrument_manager/php/instrument_manager.class.inc
intralizee: The check was added so future php errors wouldn't suppressed by the LorisException. 
Example: someone is modifying the generate_tables_sql_and_testNames.php and writes some code resulting in a php error. The error message would get lumped into the LorisException without showing the php console warning.

modules/instrument_manager/php/instrument_manager.class.inc
intralizee: See above.

modules/instrument_builder/jsx/react.questions.js
```diff
@@ -616,10 +616,18 @@ var AddElement = React.createClass({
...
-        temp.dateOption = "End year append befor start year";
+        temp.dateOption = "Invalid start or end year";
+        this.setState({
+          error: temp
+        });
+        hasError = true;
+      } else if (minDate > maxDate) {
+        let temp = (this.state.error) ? this.state.error : {};
+
+        temp.dateOption = "End year append before start year";
```
davidblader: perhaps reword this to say "End date is earlier than start date" or "End date is before start date"

intralizee: I think it's better to use "End date is before start date" than the previous choice. Fixed.

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,234 @@
...
+    $pages = array();
+
+    $sql_query_statement  = "";
+    $sql_query_table_name = "";
+    $sql_query_columns    = array();
+
+    // Setup Database object.
+    $config    =& \NDB_Config::singleton();
+    $db_config = $config->getSetting('database');
+    $db        = \Database::singleton(
```
ZainVirani: as Muhammad mentioned before, this can just be `$db =& \Database::singleton();`

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,234 @@
...
     }
-    $output.="PRIMARY KEY  (`CommentID`)\n) ENGINE=InnoDB DEFAULT CHARSET=utf8;\n";
-    print "Filename: $filename\n";
-    $fp=fopen($filename, "w");
-    fwrite($fp, $output);
-    fwrite($fp, "REPLACE INTO test_names (Test_name, Full_name, Sub_group) VALUES ('$tablename', '$title', 1);\n");
-    foreach($pages as $pageNo => $page) {
-        fwrite($fp, "INSERT INTO instrument_subtests (Test_name, Subtest_name, Description) VALUES ('$tablename', '" . $tablename  . "_page" . ($pageNo+1) . "', '$page');\n");
 
+    // Execute the SQL tasks if no error_message.
```
ZainVirani: Why is anything being executed here at all?

johnsaigle: @ZainVirani not sure if I understand what you mean totally, but this is a rare example of a tool that is called by the front-end and commands are expected to be run. 
We could add an argument here but I imagine it would always be passed when called from the front-end

ZainVirani: this script can also be run from command line, and the dev may not want the queries to be run

intralizee: Hi @johnsaigle and @ZainVirani the latest commit adds the feature of not executing the SQL query if a secondary argument "false" is supplied. See the comment examples in the generate_tables_sql_and_testNames.php file for how it works.

johnsaigle:
@driusan Can we put this fix into the 20.0-release? It's a pretty serious security issue (though only possible by `superuser`s). 

If not, @intralizee can you instead issue this against bugfix? It belongs there more than in a minor release.

intralizee:
@johnsaigle @driusan  I think this PR wouldn't be an issue going into 20.0-release. Let me know if I should rebase it for 20.0-release.

intralizee:
Hey @ZainVirani, 

When viewing Loris, go to the Admin->Instrument Manager and try uploading some instruments. 
Some instrument files can be found in the Loris/docs/instruments directory. View the database to see if everything looks good.

Also you can test the php script without SQL execution by running: 
`php generate_tables_sql_and_testNames.php < ../instruments/[instrument_name].linst false`

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,239 @@
...
 require_once "../php/libraries/Database.class.inc";
 require_once "../php/libraries/NDB_Config.class.inc";
 require_once "../php/libraries/NDB_BVL_Instrument.class.inc";
 
-$data = stream_get_contents(STDIN);
+$execute_mysql_query = ($argv[1] ?? true) === true;
 
-$instruments=explode("{-@-}",trim($data));
+$data          = stream_get_contents($argv[0]);
+$instruments   = explode("{-@-}", trim($data));
```
ZainVirani: I don't fully understand what's happening here. Are we getting a linst file (delimited by {@}) with multiple instruments in it (each delimited by {-@-}) ?

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,239 @@
...
-//taken from php.net notes
-
 // require all relevant OO class libraries
 require_once __DIR__ . "/../vendor/autoload.php";
 require_once "../php/libraries/Database.class.inc";
 require_once "../php/libraries/NDB_Config.class.inc";
 require_once "../php/libraries/NDB_BVL_Instrument.class.inc";
 
-$data = stream_get_contents(STDIN);
+$execute_mysql_query = ($argv[1] ?? true) === true;
```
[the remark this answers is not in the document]
intralizee: I think the code should execute the SQL query when the argument is true.

johnsaigle:
@ZainVirani You may be a victim of the bugs in this PR https://github.com/aces/Loris/pull/3746. Do you have quatUser configured? Are `project/tables_sql` and `project/instruments` writable by the apache user?

tools/generate_tables_sql_and_testNames.php
```diff
@@ -29,6 +29,7 @@
 require_once "../php/libraries/Database.class.inc";
 require_once "../php/libraries/NDB_Config.class.inc";
 require_once "../php/libraries/NDB_BVL_Instrument.class.inc";
+require_once "./generic_includes.php";
```
johnsaigle: The above 3 libraries are included by `generic_includes` so no need to include them anymore.

tools/generate_tables_sql_and_testNames.php
```diff
@@ -29,6 +29,7 @@
...
 
 $execute_mysql_query = ($argv[1] ?? true) === true;
```
johnsaigle: Shouldn't this just be `$execute_mysql_query = $argv[1] ?? true`?

intralizee: I think not because it's passing in a string for the argument values.

intralizee:
Hey @ZainVirani can you reattempt the valid linst file with the latest commits.

tools/generate_tables_sql_and_testNames.php
johnsaigle: A value of `"true"` will be false in the logic you have here. Is that desired?

intralizee:
Okay, I fixed all the bugs encountered by @ZainVirani and it seems like it was broken badly but should be working perfectly now.

ZainVirani:
@intralizee I'll give it another look tmrw

johnsaigle:
@ZainVirani any word on whether this is working as expected?

johnsaigle:
I'll take over this PR.

driusan:
Can you close it and re-issue it in your name so that I don't accidentally re-add the "Needs Maintainer" label in a week or two?

PR #3675, opened 2018-05-17
Fix spelling of 'occured'

zaliqarosli:
addresses https://redmine.cbrain.mcgill.ca/issues/12680


modules/imaging_uploader/jsx/UploadForm.js
```diff
@@ -195,7 +195,7 @@ class UploadForm extends React.Component {
         });
       },
       error: function(err) {
-        const errMessage = "The following errors occured while " +
```
kongtiaowang: @zaliqarosli I think you need to compile this JSX to JS file, then upload it.

PR #3677, opened 2018-05-17
[Tools] Script to automate LORIS SQL patching

johnsaigle:
*This PR has been heavily refactored so a lot of the conversation is outdated.*

## Overview

The script will display to a user which patches they need to apply, in order. By passing the argument `--apply-patches` a user can have the update script apply SQL patch files to their database automatically. 

## Details
* `PHP_CLI_Helper.class.inc` provides some functions for nice output and can be re-used by other scripts. It is also used in #3981.

The function to retrieve patches is heavily dependent on the way we organize and name SQL files. If the directory structure changes, this functionality of the script is likely to break. 

## To Test

Run the script on your LORIS! Report anything weird or confusing.


tools/update.php
```diff
@@ -0,0 +1,699 @@
...
+    // @codingStandardsIgnoreStart
+    . "\tSee: https://github.com/aces/Loris/pulls?q=is%3Apr+label%3A%22Caveat+for+Existing+Projects%22+is%3Amerged"
+    // @codingStandardsIgnoreEnd
+    . PHP_EOL . PHP_EOL;
+sleep(3);
+main();
+
+/**
+ * Updates source code and dependencies for LORIS. Outputs missing patch files
+ * based on verion information
```
PapillonMcGill: should have an ended dot (**.**)

johnsaigle: Done!

tools/update.php
```diff
@@ -0,0 +1,705 @@
...
+ *
+ * @return bool Always true (for now). False if patches fail to be applied.
+ */
+function applyPatches($patches, $db_config, $report_only = true) : bool
+{
+    // Iterate over all patches and source them into MySQL
+    $A = $db_config['database'];
+    $u = $db_config['username'];
+    $p = $db_config['password'];
+    $h = $db_config['host'];
```
ZainVirani: descriptive variable names would be better

tools/update.php
```diff
@@ -0,0 +1,705 @@
...
+    if ($report_only === true) {
+        echo '[-] NOTE: Running in Report-Only mode. Patching commands will be '
+            . 'displayed but not executed.' . PHP_EOL;
+        sleep(1);
+    }
+    foreach ($patches as $patch) {
+        $cmd = "mysql -h $h -u $u -p -A $A < $patch";
+        if ($report_only === false) {
+            if (doExec($cmd) === false) {
+                break;
```
ZainVirani: so if I understand correctly, if a patch fails, doExec should spit out the error and this loop will break (and the script will terminate as this is the last action in main())

I think before breaking the rest of the commands should be output to the user along with a message indicating that the rest of the commands were not run. I figure if this didn't happen a user would just try to rerun this script which could make things worse.

tools/update.php
```diff
@@ -0,0 +1,705 @@
...
+    . ' as they are needed to e.g. update apt packages.' . PHP_EOL;
+echo 'You may wish to review code changes tagged with "Caveat For '
+    . 'Existing Projects" as they may include changes to your workflow. '
+    . PHP_EOL
+    // @codingStandardsIgnoreStart
+    . "\tSee: https://github.com/aces/Loris/pulls?q=is%3Apr+label%3A%22Caveat+for+Existing+Projects%22+is%3Amerged"
+    // @codingStandardsIgnoreEnd
+    . PHP_EOL . PHP_EOL;
+sleep(3);
+main();
```
ZainVirani: I think it would be best if the script required a `-confirm` to actually update, and without it would just spew out some help text (and the newest release version # so users can be sure they want to update), just so this script is never accidentally started

tools/update.php
johnsaigle: They are descriptive to the extent that they match the flags for `mysql` on the command line.

tools/update.php
johnsaigle: @ZainVirani that's a good point. I'll come back to this pending the conversation as to whether this script should even apply patche or if it should just print them.

tools/update.php
```diff
@@ -0,0 +1,705 @@
...
+    $db_config = $config->getSetting('database');
+    $db        = \Database::singleton(
+        $db_config['database'],
+        $db_config['username'],
+        $db_config['password'],
+        $db_config['host']
+    );
+    $paths     = $config->getSetting('paths');
+    $loris_root_dir = $paths['base'];
+    $backup_dir     = "/tmp/bkp-LORIS"; // TODO: should this be configurable?
```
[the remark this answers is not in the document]
johnsaigle: This is now configurable

tools/update.php
johnsaigle: There is now a confirm flag present.

tools/update.php
johnsaigle: @ZainVirani I have added this functionality. Thanks for the suggestion!

tools/update.php
```diff
@@ -0,0 +1,747 @@
...
+    if ($diff_major < 0) {
+        echo '[!] Your version of LORIS is ahead of the latest release. If you'
+        . ' are not a developer working on the bleeding edge of LORIS, then '
+        . 'something has gone very wrong. No SQL patches will be displayed.'
+        . PHP_EOL;
+        return array();
+    }
+    echo "[**] Latest version $version_to is ahead of installed $version_from "
+        . "by $diff_major MAJOR release(s), $diff_minor MINOR release(s)."
+        . PHP_EOL;
```
PapillonMcGill: This could end up as `Latest version 22.0 is ahead of installed 21.3 by 1 MAJOR release(s), -3 MINOR release(s).`

This negative minor can be quite confusing. I suggest to remove the minor part when major is > 0.

tools/update.php
```diff
@@ -0,0 +1,747 @@
...
+            . '.' . $to_versions[MINOR] . '/';
+        $patches       = array_merge($patches, glob($dev_patch_dir . '*.sql'));
+    }
+    return $patches;
+}
+
+/**
+ * Apply a list of SQL patch files using exec.  Currently this function is only
+ * called with $report_only = true so patches are never applied. This can be
+ * changed in the future if this functionality is required.
```
PapillonMcGill: This comment need to be updated

tools/update.php
```diff
@@ -0,0 +1,747 @@
...
+ *
+ * @param array      $patches       Patch files to be applied
+ * @param dictionary $db_config     DB config info. Needed for credentials
+ * @param bool       $apply_patches If true, MySQL commands will be run
+ *
+ * @return bool Always true (for now). False if patches fail to be applied.
```
PapillonMcGill: TBU

tools/update.php
```diff
@@ -0,0 +1,747 @@
...
+    if ($apply_patches === false) {
+        echo '[-] NOTE: Running in Report-Only mode. Patching commands will be '
+            . 'displayed but not executed.' . PHP_EOL;
+        sleep(1);
+    }
+    while (count($patches) > 0) {
+        $patch = array_shift($patches);
+        $cmd   = "mysql -h $h -u $u -p -A $A < $patch";
+        if ($apply_patches === true) {
+            if (doExec($cmd) === false) {
```
PapillonMcGill: Does it really apply the patch? No password is passed to the DoExec function and no where the password is passed as a prompt. So the user need to know the DB pwd (and enter it maybe multiple time).

johnsaigle: You're right, it would probably be best to import the settings from Config so that the user doesn't need to do this manually.

tools/update.php
```diff
@@ -0,0 +1,747 @@
...
+    // patches.
+    $start = $from_versions[MAJOR];
+    $end   = $start; // default case: there is a minor release difference
+    if ($diff_major > 0) {
+        $end = $to_versions[MAJOR];
+    }
+    $patch_dir           = $loris_root . 'SQL/Release_patches/';
+    $all_release_patches = glob($patch_dir . '*.sql');
+    $patches = [];
+    if ($diff_major > 0 || $diff_minor > 0) {
```
ridz1208: @johnsaigle most of the time projects will need to apply project specif SQL in between 2 releases to prepare the database, this cannot be supported in this script

tools/update.php
```diff
@@ -0,0 +1,747 @@
...
+                }
+            }
+        }
+    }
+    if (isDev()) {
+        echo "[+] Developer instance detected. Including developer patches..."
+            . PHP_EOL;
+
+        // Add all patches in Archive/$MAJOR.$MINOR. Everything other needed
+        // command will be in the Release patches which are been added above
```
ridz1208: the release patch is the concatenation of the patches in `Archive` why are you readding them here ?

tools/update.php
johnsaigle: >this cannot be supported in this script

What specifically can't be supported? I'm not sure I understand.

tools/update.php
johnsaigle: Right... I guess this should be `New_patches`. I remember at some point we were actually putting new patches in `Archive` so I based this on that.

I'll update this.

tools/update.php
johnsaigle: Ah nevermind I remember why I did it this way:

1.  The password is not passed to `doExec` because this function always prints out the command. If the password is passed this way then it will appear in plain text in the console which isn't desirable.
2) I think getting the user to enter the password manually is actually a good way to be _really_ sure that the user actually wants to make changes to the DB. While this can be done automatically it's probably good to have this inconvenience in the interest of data integrity.
3) While the user may need to enter the DB password manually, and perhaps multiple times, this is what they would have to do anyway without this script. So it's no worse than the current procedure.

tools/update.php
ridz1208: Assume I'm 2 releases behind, for each release I have PRE and POST custom SQL patches designed for my project (thats usually the case for any project upgrading). how can I run them using your script as follows:
1. `PRE_19.0.sql` -> project patches
2. `18.0_To_19.0.sql` -> loris patches
3. `POST_19.0.sql` -> project patches
4. `PRE_20.0.sql` -> project patches
5. `19.0_to_20.0.sql` -> loris patches
6. `POST_20.0.sql` -> project patches

tools/update.php
ridz1208: lets talk

tools/update.php
johnsaigle: I didn't know about this process for projects. 

If it's just one release you could do

1. Source `PRE` patch.
2. Run script.
3. Source `POST` patch.

I'm not sure how to handle the case where a project is more than 1 major release behind. If there is a standard location and naming convention for the files it would probably be possible to do these checks. Is that the case? Do you have any suggestions for how to incorporate this feature?

johnsaigle:
TODO: Add single-use patches to update process

tools/update.php
johnsaigle: As discussed in person, the script will be amended to not do SQL patching if the installed LORIS version is more than 1 major version behind the latest release, as it's difficult to automate and likely quite risky.

The script will display a message informing the user that the update process is complete in terms of src code, packages, etc. but that they will need to apply patches manually.

johnsaigle:
@ridz1208 @PapillonMcGill Please re-review when you can.

modules/login/ajax/verificationimage.php
```diff
@@ -0,0 +1,42 @@
+<?php
```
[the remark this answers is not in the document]
johnsaigle: Hm not sure. Something strange must have happened when I rebased.

johnsaigle:
@PapillonMcGill Thanks, I've incorporated your feedback.

johnsaigle:
@driusan Could you review this when you get a chance? It's been ready for final review for about a month

tools/update.php
```diff
@@ -0,0 +1,746 @@
...
+if (PHP_MAJOR_VERSION < $required_major_php
+    || PHP_MINOR_VERSION < $required_minor_php
+) {
+    die("[-] ERROR: {$argv[0]} and LORIS require PHP v$php_version or higher.");
+}
+
+/* Create db connection and get version info. */
+$config    = \NDB_Config::singleton();
+$db_config = $config->getSetting('database');
+$db        = \Database::singleton(
```
ridz1208: that might be problematic if the release patch requires admin credentials. (generally it shouldn't but all release patches have probably been run as admins to this day) just investigate

tools/update.php
```diff
@@ -0,0 +1,746 @@
...
+// Check that the backup directory argument is present and valid
+$backup_dir = $argv[1] ?? '';
+if (!is_dir($backup_dir)) {
+    echo '[-] ERROR: Argument suppled for backup directory is not a valid '
+        . 'directory.' . PHP_EOL;
+    die(usageString());
+}
+// Only apply SQL patches if user explicitly requests it
+$apply_patches = false;
+if (in_array('--apply-patches', $argv)) {
```
ridz1208: always better to use `true` as 3rd arg when using `in_array()`

see http://php.net/manual/en/function.in-array.php#106319

ridz1208: same for all below

tools/update.php
```diff
@@ -0,0 +1,746 @@
...
+// Update source code.
+echo '[***] Beginning LORIS update process.' . PHP_EOL;
+echo '[*] Release notes:' . PHP_EOL;
+echo $info->{'body'} . PHP_EOL . PHP_EOL;
+echo "[**] Updating LORIS source code "
+    . "($preupdate_version --> $release_version)" . PHP_EOL;
+if (updateSourceCode($loris_root_dir, $backup_dir)) {
+    echo '[+] LORIS source code files successfully updated.' . PHP_EOL;
+}
+
```
ridz1208: shouldn't it check the php version, apache version mysql version before starting ?

ridz1208: for example 20.0.0 requires apache 2.4 if the update runs without prompting for it, everything is gonna look fine but the website will not be functional

tools/update.php
```diff
@@ -0,0 +1,746 @@
...
+    $release_version
+);
+if ($patches) {
+    echo "[*] Patches to update:" . PHP_EOL;
+    foreach ($patches as $filename) {
+        echo "\t] " . $filename . PHP_EOL;
+    }
+    echo '[*] Applying SQL patches...' . PHP_EOL;
+    applyPatches($patches, $db_config, $apply_patches);
+}
```
ridz1208: shouldn't we add the `make` step from @driusan 's changes ?
https://github.com/aces/Loris/pull/3891

ridz1208:
Significant changes after manual testing.

out of curiosity @intralizee how did you test this ? was it just a general run or did you try to break it ?

tools/update.php
```diff
@@ -0,0 +1,746 @@
...
+            . PHP_EOL;
+    }
+
+    /* Following a discussion on this pull request, it was decided that patching
+     * SQL files automatically when a LORIS is more than one version behind
+     * should not be supported. This is to minimize errors that could impact
+     * data integrity.
+     * https://github.com/aces/Loris/pull/3677#discussion_r202814374
+     */
+    if ($diff_major > 1) {
```
ridz1208: now that I think of it it could also apply to multiple minor versions ? not sure

tools/update.php
johnsaigle: Well this is targeted to the minor release, not the major release. But otherwise yes.

tools/update.php
johnsaigle: It's still an extra command-line flag and the script prints out the release notes. Should we add more protections here?

tools/update.php
johnsaigle: Yes this is a good idea.

tools/update.php
johnsaigle: I think it depends on the privileges of the SQL user, e.g. whether they can do ALTER, CREATE etc. That'll depend in a large part on the administrator of the system and how they set up their databases. 

I think in the worst case MySQL will give an error about not having the right permissions and the script will continue onward which shouldnt cause issues as patch application is the last step.

ridz1208: it still places the DB is a half-update state and thats generally very difficult to recover front. A very Ideal way to tackle this would be to see what privileges are needed for a specific patch and verify if user have them.

the realistic way of doing this would be to add a notice before the run that says that all SQL patches are run as a regular user ORRR to use the quat credentials when/if they are available ?

tools/update.php
ridz1208: dont we still need to recompile even between minor releases @driusan ?

tools/update.php
ridz1208: maybe not, i might need to give it a run to understand fully. are all the possible flags defined anywhere ? any examples on how to run the script/ a man page

johnsaigle: I added an explicit warning and question-prompt so that a user will be sure of what they're doing.

tools/update.php
johnsaigle: The make file doesn't exist in the minor version, I mean

tools/update.php
johnsaigle: It has a usage message when you use the script wrong, but no there are no examples. Should I add some? Feels like that's the sort of thing that should maybe go into a README file

tools/update.php
johnsaigle: This has been added

tools/update.php
johnsaigle: I implemented the "very Ideal way" :smile_cat:

tools/update.php
ridz1208: ohh, that makes total sense

:man_facepalming:

tools/update.php
ridz1208: im very ideally impressed

tools/update.php
```diff
@@ -402,6 +460,92 @@ function applyPatches($patches, $db_config, $apply_patches = false) : bool
...
+
+/**
+ * Use bash commands to extract MySQL keywords from an sql script.
+ */
+function getKeywords(string $patch): array
+{
+    //Scan file for MySQL keywords. Exclude comments and parentheses using grep.
+    $cmd = "cut $patch -d ' ' -f 1 "
+        . "| sort -u "
+        . "| grep -v '\--' | grep -v ')'";
```
PapillonMcGill: also need to remove empty line that cut will generate on multi line instruction like CREATE TABLE.

Can use "uniq" to suppress multiple occurrences of an instruction after sort

tools/update.php
```diff
@@ -402,6 +460,92 @@ function applyPatches($patches, $db_config, $apply_patches = false) : bool
...
+                 */
+                shell_exec($cmd)
+            )
+        )
+    );
+}
+function getMissingPrivileges(string $patch, $db_config) {
+    $keywords = getKeywords($patch);
+    $privileges = getPrivileges($db_config);
+    $irrelevantWords = array(
```
PapillonMcGill: should also check for upper case as in_array is case sensitive

tools/update.php
johnsaigle: Empty lines are dealt with just below this by the `array_filter` call.

`sort -u` is pretty much equivalent to `sort | uniq`.

tools/update.php
johnsaigle: A few lines above this I used `array_map` to convert all `$keywords` to lower case which should take care of this.

johnsaigle:
@PapillonMcGill Thanks for reviewing again!

tools/update.php
```diff
@@ -0,0 +1,890 @@
...
+        $defaultAnswer = 'N';
+        writeQuestion('Apply patches anyway?', $answers);
+        $answer = readAnswer($answers, $defaultAnswer);
+        if ($answer != 'y') {
+            return array();
+        }
+    }
+
+
+    // For every major version released between the version that is installed
```
ridz1208: comment might need to be changed "for every" no longer applies as only 1 patch can be supported

tools/update.php
```diff
@@ -0,0 +1,890 @@
...
+
+/**
+ * Returns a list of SQL patch files to apply based on LORIS version
+ * differences.
+ *
+ * @param string $loris_root   The directory where LORIS is installed.
+ * @param string $version_from Pre-update LORIS version.
+ * @param string $version_to   Post-update LORIS version.
+ *
+ * @return array Patch files to be applied. Empty if up-to-date.
```
ridz1208: at this point I think this is just a single file

tools/PHP_CLI_Helper.class.inc
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @return void
+ */
+function writeQuestion($question, $answers) : void
+{
+    echo $question . ' (' . implode('/', $answers) . '): ' . PHP_EOL;
+}
+
+/**
+ * Gets user input from STDIN and checks if it matches a valye in
```
PapillonMcGill: s/valye/value

tools/PHP_CLI_Helper.class.inc
```diff
@@ -0,0 +1,103 @@
...
+ * function writeQuestion
+ *
+ * @param array $possibleAnswers Possible answers to a prompt
+ * @param array $defaultAnswer   Response if user entered invalid response
+ *
+ * @return string The user input representing their answer or the default answer
+ */
+function readAnswer($possibleAnswers, $defaultAnswer) : string
+{
+    $in     = fopen('php://stdin', 'rw+');
```
PapillonMcGill: 'r' should be enough for stdin

tools/update.php
```diff
@@ -49,12 +49,30 @@
...
+        "[-] ERROR: {$argv[0]} and LORIS require PHP v$required_php or higher."
+        . PHP_EOL
+    );
+}
+// PHP version required for LORIS.
+$apache_info = shell_exec('apache2 -v');
+/* Look for the string "$major.$minor" in info string. Also match on versions
+ * higher than minor version because we want AT LEAST that version.
+ */
+$pattern = "/$required_major_apache\.[$required_minor_apache-9]/";
```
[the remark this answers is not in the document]
PapillonMcGill: @johnsaigle still missing some change here!

tools/update.php
johnsaigle: Nope! For example if you're a developer and two patches have been released on the minor version but a new minor release patch hasn't been issued yet.

johnsaigle:
@ridz1208 @PapillonMcGill This is ready for review whenever you have the time.

tools/update.php
```diff
@@ -0,0 +1,787 @@
...
```
johnsaigle:
```suggestion
$pattern = "/$required_major_apache\.[$required_minor_apache-9]\.0-9]/";
```

tools/update.php
```diff
@@ -69,7 +69,7 @@
 /* Look for the string "$major.$minor" in info string. Also match on versions
  * higher than minor version because we want AT LEAST that version.
  */
-$pattern = "/$required_major_apache\.[$required_minor_apache-9]/";
+$pattern = "/$required_major_apache\.[$required_minor_apache-9]\.0-9]/";
 if (preg_match($pattern, $apache_info) === 0) {
```
PapillonMcGill: \.[0-9] would be better

HenriRabalais:
Looks good, but I think a lot of this code is out of my scope of expertise. 
@PapillonMcGill @xlecours @ridz1208 @driusan could you provide confirmation that this is good2go?

tools/update.php
PapillonMcGill: pinging @johnsaigle

driusan:
Are you going to work on this open PR or send new ones for the above? (If the later, should we close this?)

johnsaigle:
I think I'll convert #3981into the bootstrapping and use this one for the SQL patching stuff. Let's leave it open because there's an existing discussion on patching approaches here.

tools/update.php
```diff
@@ -0,0 +1,387 @@
...
+    // versions. The below loops will, in this case, match the minor release
+    // patches.
+    $start = $from_versions[MAJOR];
+    $end   = $start; // default case: there is a minor release difference
+    if ($diff_major > 0) {
+        $end = $to_versions[MAJOR];
+    }
+    $patch_dir           = $loris_root . 'SQL/Release_patches/';
+    $all_release_patches = glob($patch_dir . '*.sql');
+    $patches = [];
```
PapillonMcGill:
```suggestion
    $patches = array();
```

tools/update.php
```diff
@@ -0,0 +1,387 @@
...
+    return true;
+}
+
+function getPrivileges($db_config): array {
+    $A = $db_config['database'];
+    $u = $db_config['username'];
+    $h = $db_config['host'];
+    $p = $db_config['password'];
+    /* Run 'SHOW privileges' for this user and parse the output. We are passing
+     * the password value on bhelaf of the user here as the SHOW command does
```
PapillonMcGill:
```suggestion
     * the password value on behalf of the user here as the SHOW command does
```

driusan:
Regarding the "This PR has been heavily refactored so a lot of the conversation is outdated.", can you close it and send a new PR (when the refactoring is done) so that it's clear what's up to date and what's irrelevant?

ridz1208:
@johnsaigle Can you re-issue it when/if its done to get a clean history ?

closing for now

PR #3678, opened 2018-05-17
[Doc Repo] Fix path traversal security issue

johnsaigle:
See Redmine #14511.

Related to https://github.com/aces/Loris/pull/3603, which contains a better fix + more significant refactoring.

This PR provides the minimal changes necessary to fix the vulnerability.

### DIFF
--- a/modules/document_repository/ajax/documentEditUpload.php
+++ b/modules/document_repository/ajax/documentEditUpload.php
@@ -76,9 +76,7 @@ if ($userSingleton->hasPermission('document_repository_view')
             mkdir($base_path . $puser, 0777);
         }
 
-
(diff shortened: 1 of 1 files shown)


modules/document_repository/ajax/documentEditUpload.php
```diff
@@ -78,6 +78,7 @@
 
 
         $target_path = $base_path  . $fileBase;
+        error_log("Target path is: $target_path");
```
driusan: Left over debug statement.. this shouldn't be in the error log.

johnsaigle: Done!

PR #3679, opened 2018-05-18
[media][document_repository] Fix error with downloading files with names containing '+'

intralizee:
This pull request `is to allow users to download files containing "+" in them`.

It `solves the issue by encoding the file string to be download and so the $_GET['File'] will retrieve the correct string.`.

See also: `Bug #14468`


johnsaigle:
@ridz1208 I think you should be the one to test this since it was you who found the issue.

modules/document_repository/php/document_repository.class.inc
```diff
@@ -407,6 +407,16 @@ class Document_Repository extends \NDB_Menu_Filter
...
             $DisplayItem['File_name'] = utf8_encode($item['File_name']);
+            $DisplayItem['Data_dir']  = strtr(
+                rawurlencode($DisplayItem['Data_dir']),
+                array(
+                 '%21' => '!',
+                 '%2A' => '*',
+                 '%27' => "'",
+                 '%28' => '(',
+                 '%29' => ')',
+                )
```
PapillonMcGill: Just a question. Does this represent all the characters list that need to be encoded?
Are space and % sign accepted in files name?

intralizee: No,
rawurlencode — URL-encode according to RFC 3986 and strtr — Translate characters or replace substrings. The section you referenced of the array is to make it equivalent to the javascript function encodeURIComponent() in image uploading.

driusan: I'm a little confused about the relationship between this (which doesn't involve "+") and the PR title/description (which just says it's fixing "+"?

intralizee: @driusan The rawurlencode function handles the "+" character(s) conversion and many other characters needed to be converted as the function is designed for the problem the PR is designed to fix. The strtr function handles replacing what is in the array that @PapillonMcGill quoted for the output of the rawurlencode function and to be equivalent to how the javascript function handles conversion. The reason for it all is to make the javascript & php output the same.

driusan: Can you add that explanation to a code comment so that the reasoning isn't lost to someone reading it in the future?

PR #3680, opened 2018-05-21
[tools] Removed log warning for data dictionary builder

ridz1208:
This removes the following warning from the error log
`PHP Notice:  Undefined offset: 1 in /var/www/ccna/tools/data_dictionary_builder.php on line 172`

### DIFF
--- a/tools/data_dictionary_builder.php
+++ b/tools/data_dictionary_builder.php
@@ -169,7 +169,7 @@ foreach ($instruments AS $instrument) {
             //for HTML_QuickForm versions of standard HTML Form Elements...
         default:
             //continue; // jump straight to validity for debugging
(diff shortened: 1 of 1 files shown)


tools/data_dictionary_builder.php
```diff
@@ -169,7 +169,7 @@
             //for HTML_QuickForm versions of standard HTML Form Elements...
         default:
             //continue; // jump straight to validity for debugging
-            if (preg_match("/^Examiner/", $bits[1])) {
+            if (isset($bits[1]) && preg_match("/^Examiner/", $bits[1])) {
```
driusan: This might be cleaner using null coalesce on the bits instead of isset. ie. `preg_match("...", $bits[1] ?? "");`

driusan:
What's going on with Travis on this?

PR #3685, opened 2018-05-23
[datadict] Search keyword filter will now return rows for multiple words

intralizee:
This pull request `fixes filtering when searching with multiple words for keywords. Before no results were returned and now results are returned. Also fixes another bug where description column was never searched.`. 

Inside NDB_Menu_Filter.class.inc `I added a case for when the keyword variable has multiple words or text and used REGEXP for the mysql query`.

See also: `Bug #12910`

### DIFF
--- a/modules/datadict/php/datadict.class.inc
+++ b/modules/datadict/php/datadict.class.inc
@@ -56,7 +56,7 @@ class Datadict extends \NDB_Menu_Filter
                           'pt.sourceFrom',
                           'pt.Name',
                           'pt.sourceField',
-                          'coalesce(pto.description,pt.description)'.
+                          'COALESCE(pto.description,pt.description)'.
                           ' as description',
                          );
 
@@ -64,9 +64,9 @@ class Datadict extends \NDB_Menu_Filter
                                'pt.sourceFrom',
                                'pt.Description',
                               );
-        $this->query        = ' FROM parameter_type pt LEFT JOIN 
+        $this->query        = " FROM parameter_type pt LEFT JOIN 
                                 parameter_type_override pto USING (Name)
-                 		        WHERE pt.Queryable=1';
+                 		        WHERE pt.Queryable=1";
 
         $this->formToFilter = array(
                                'sourceFrom'  => 'pt.sourceFrom',
@@ -76,7 +76,7 @@ class Datadict extends \NDB_Menu_Filter
         $this->searchKeyword = array(
                                 'pt.Name',
                                 'pt.sourceField',
-                                'coalesce(pto.description,pt.description)',
+                                'COALESCE(pto.description,pt.description)',
                                );
         return true;
     }
@@ -271,16 +271,16 @@ class Datadict extends \NDB_Menu_Filter
                 || strtolower(substr($key, -10)) == 'categoryid'
                 || strtolower(substr($key, -6)) == 'gender'
             ) {
-                $query .= ' AND ' . $key . ' = \'' . $val . '\' ';
+                $query .= " AND $key = '$val' ";
             } else {
                 if ($val == 'empty') {
-                    $query .= ' AND COALESCE(pto.description,pt.description) = \'\'';
+                    $query .= " AND COALESCE(pto.description,pt.description) = ''";
                 } elseif ($val=='modified') {
-                    $query .= ' AND pto.name IS NOT NULL';
+                    $query .= " AND pto.name IS NOT NULL";
                 } elseif ($val=='unmodified') {
-                    $query .= ' AND pto.name IS NULL';
+                    $query .= " AND pto.name IS NULL";
                 } else {
-                    $query .= ' AND '.$key.' LIKE \'' . $val . '%\' ';
+                    $query .= " AND $key LIKE '$val%' ";
                 }
             }
         }
--- a/modules/datadict/php/datadict.class.inc
+++ b/modules/datadict/php/datadict.class.inc
@@ -76,7 +76,7 @@ class Datadict extends \NDB_Menu_Filter
         $this->searchKeyword = array(
                                 'pt.Name',
                                 'pt.sourceField',
-                                'pt.description',
+                                'coalesce(pto.description,pt.description)',
                                );
         return true;
     }
--- a/modules/datadict/php/datadict.class.inc
+++ b/modules/datadict/php/datadict.class.inc
@@ -27,28 +27,28 @@ class Datadict extends \NDB_Menu_Filter
 {
     var $AjaxModule = true;
     /**
-    * Overloading this method to allow access to site users (their own site
-    * only) and users w/ multisite privs
-    *
-    * @note   overloaded function
-    * @return bool
-    * @access private
-    */
+     * Overloading this method to allow access to site users (their own site
+     * only) and users w/ multisite privs
+     *
+     * @note   overloaded function
+     * @return bool
+     * @access private
+     */
     function _hasAccess()
     {
         // create user object
         $user =& \User::singleton();
 
         return ($user->hasPermission('data_dict_view') ||
-                $user->hasPermission('data_dict_edit'));
+            $user->hasPermission('data_dict_edit'));
     }
     /**
-    * Setup variables function
-    *
-    * @note   Setup variables function
-    * @return bool
-    * @access private
-    */
+     * Setup variables function
+     *
+     * @note   Setup variables function
+     * @return bool
+     * @access private
+     */
     function _setupVariables()
     {
         // set the class variables
@@ -57,7 +57,7 @@ class Datadict extends \NDB_Menu_Filter
                           'pt.Name',
                           'pt.sourceField',
                           'coalesce(pto.description,pt.description)'.
-                                                   ' as description',
+                          ' as description',
                          );
 
         $this->validFilters = array(
@@ -66,7 +66,7 @@ class Datadict extends \NDB_Menu_Filter
                               );
         $this->query        = " FROM parameter_type pt LEFT JOIN 
                                 parameter_type_override pto USING (Name)
-                 		WHERE pt.Queryable=1";
+                 		        WHERE pt.Queryable=1";
 
         $this->formToFilter = array(
                                'sourceFrom'  => 'pt.sourceFrom',
@@ -82,10 +82,10 @@ class Datadict extends \NDB_Menu_Filter
     }
 
     /**
-    * Set Filter Form
-    *
-    * @return none
-    */
+     * Set Filter Form
+     *
+     * @return none
+     */
     function setup()
     {
         parent::setup();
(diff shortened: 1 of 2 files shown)


modules/datadict/php/datadict.class.inc
```diff
@@ -76,7 +76,7 @@ class Datadict extends \NDB_Menu_Filter
         $this->searchKeyword = array(
                                 'pt.Name',
                                 'pt.sourceField',
-                                'pto.description',
+                                'pt.description',
```
driusan: This will make it so that it doesn't search the override (which should have priority over the default parameter_type table.)

kongtiaowang:
It will make another issue. For example, all the

modules/datadict/php/datadict.class.inc
intralizee: Thanks David for the help. Now it's set to 'coalesce(pto.description,pt.description)'.

php/libraries/NDB_Menu_Filter.class.inc
```diff
@@ -353,6 +353,16 @@ class NDB_Menu_Filter extends NDB_Menu
         if (isset($_REQUEST['keyword'])) {
             // It was included in the request, so save it.
             $key = $_REQUEST['keyword'];
+            // Check if multiple keyword search
+            if ($_REQUEST['test_name'] == 'datadict'
```
driusan: This shouldn't be in the base class. If datadict needs to do something special, it should override the function.

driusan: (same with all the instances below.)

intralizee: Awesome it makes total sense to do that! Will do.

victoriafoing:
I tested this PR and the Search keyword can now be longer than one word.
The search returns results even if the words following the first word are not in the table.
This can be convenient for users in case of misspellings.

search phrase

I would like to open a discussion about whether we should add a functionality that allows users to search for results that match the Search phrase exactly (i.e. by adding quotations around Search phrases)

intralizee:
Added the feature @victoriafoing suggested.

modules/datadict/php/datadict.class.inc
```diff
@@ -113,16 +113,149 @@ class Datadict extends \NDB_Menu_Filter
...
-    *
-    * @note   Get base query
-    * @return string
-    * @access private
-    */
+     * Builds the filter array to be used in this filter menu. The end result
+     * should be that all the NDB_Menu_Filter attributes required for the query
+     * to work are set up.
+     *
+     * Generally, it does the following:
```
davidblader: i wish everybody commented their code this well :+1:

davidblader:
image
looks like there is still some mix up when using quotes / doesn't do exact pattern matching when search value is put into quotes (see row 10)

modules/datadict/php/datadict.class.inc
```diff
@@ -113,16 +113,149 @@ class Datadict extends \NDB_Menu_Filter
...
+                }
+                $keywords = array_merge($keywords, $keywords_extra);
+            } else if (preg_match('/\\s/', $key)
+            ) {
+                // Check if multiple keyword search
+                $keywords = explode(' ', $key);
+                foreach ($keywords as &$k) {
+                    $k = '.*'.$k;
+                }
+                unset($k);
```
nicolasbrossard: Not sure if this is necessary since you are not using $k after this

modules/datadict/php/datadict.class.inc
```diff
@@ -113,16 +113,149 @@ class Datadict extends \NDB_Menu_Filter
...
+                    if (preg_match('\"[\w\s]+\"', $keyword)) {
+                        // Handle the quoted keyword.
+                        $keyword = '(?i)^' . $keyword . '$';
+                    } else if (preg_match('/\\s/', $keyword)) {
+                        // Handle multiple keywords.
+                        $split_keywords = explode(' ', $keyword);
+                        foreach ($split_keywords as &$split_keyword) {
+                            $split_keyword = '.*'.$split_keyword;
+                        }
+                        unset($split_keyword);
```
nicolasbrossard: Not sure this is necessary since you are not using this variable after this anyway

modules/datadict/php/datadict.class.inc
```diff
@@ -113,16 +113,149 @@ class Datadict extends \NDB_Menu_Filter
...
+                );
+            }
+            if (is_array($keywords)) {
+                // Quoted keyword(s) exist.
+                $keywords_extra = array();
+                foreach ($keywords as &$keyword) {
+                    if (preg_match('\"[\w\s]+\"', $keyword)) {
+                        // Handle the quoted keyword.
+                        $keyword = '(?i)^' . $keyword . '$';
+                    } else if (preg_match('/\\s/', $keyword)) {
```
nicolasbrossard: I think you are missing an `if` here to handle the case where `$keyword` is delimited by single quotes.

intralizee: You're right. 

I thought it would be best to not include single quotes. The code logic of splitting the segments into an array would likely break if a user had for example:
'I'm 27 years old'

What do you think is best?

modules/datadict/php/datadict.class.inc
nicolasbrossard: Fair enough.

modules/datadict/php/datadict.class.inc
nicolasbrossard: I think single and double quotes should be supported. You can either add an `if` clause here or modify the regex in your first `if` test so that it supports both quoting options.

intralizee: I'm not necessarily a regular expression expert. I noticed the problem @davidblader mentions is because the regexp I created needs fixing to include parenthesis with other characters. 

Currently the regexp code I wrote will take a search as example:
"exact search example" independent word searches "another exact search" ..

The result will do searches for exact two times and independent word searches.

The reason I'm leaning towards removing single quotes support is because of cases where single quotes are used multiple times and such as in the description column. I think it's unnecessary with added complexity but i'll take a look if I can figure out some magic with regex.

modules/datadict/php/datadict.class.inc
```diff
@@ -133,23 +271,223 @@ class Datadict extends \NDB_Menu_Filter
...
+                    $query .= ' AND COALESCE(pto.description,pt.description) = \'\'';
                 } elseif ($val=='modified') {
-                    $query .= " AND pto.name IS NOT NULL";
+                    $query .= ' AND pto.name IS NOT NULL';
                 } elseif ($val=='unmodified') {
-                    $query .= " AND pto.name IS NULL";
+                    $query .= ' AND pto.name IS NULL';
                 } else {
-                    $query .= " AND $key LIKE '$val%' ";
+                    $query .= ' AND $key LIKE \'' . $val . '%\' ';
```
ZainVirani: I believe `$key` will show up as `$key` in this string. Should be separated like $val.

ZainVirani: I believe $key will show up as $key in this string. Should be separated like $val.

modules/datadict/php/datadict.class.inc
```diff
@@ -57,16 +57,16 @@ class Datadict extends \NDB_Menu_Filter
                           'pt.Name',
                           'pt.sourceField',
                           'coalesce(pto.description,pt.description)'.
```
zaliqarosli: capitalize COALESCE

modules/datadict/php/datadict.class.inc
```diff
@@ -57,16 +57,16 @@ class Datadict extends \NDB_Menu_Filter
...
 
         $this->validFilters = array(
                                'pt.sourceFrom',
                                'pt.Description',
                               );
-        $this->query        = " FROM parameter_type pt LEFT JOIN 
+        $this->query        = ' FROM parameter_type pt LEFT JOIN 
                                 parameter_type_override pto USING (Name)
-                 		WHERE pt.Queryable=1";
+                 		        WHERE pt.Queryable=1';
```
zaliqarosli: i think we want to keep queries within " "

modules/datadict/php/datadict.class.inc
```diff
@@ -76,16 +76,16 @@ class Datadict extends \NDB_Menu_Filter
         $this->searchKeyword = array(
                                 'pt.Name',
                                 'pt.sourceField',
-                                'pto.description',
+                                'coalesce(pto.description,pt.description)',
```
zaliqarosli: capitalize coalesce

modules/datadict/php/datadict.class.inc
```diff
@@ -133,23 +271,223 @@ class Datadict extends \NDB_Menu_Filter
...
+                    $query .= ' AND COALESCE(pto.description,pt.description) = \'\'';
                 } elseif ($val=='modified') {
-                    $query .= " AND pto.name IS NOT NULL";
+                    $query .= ' AND pto.name IS NOT NULL';
                 } elseif ($val=='unmodified') {
-                    $query .= " AND pto.name IS NULL";
+                    $query .= ' AND pto.name IS NULL';
                 } else {
-                    $query .= " AND $key LIKE '$val%' ";
+                    $query .= ' AND '.$key.' LIKE \'' . $val . '%\' ';
```
zaliqarosli: if you keep queries within " ", you don't need these extra escaped ' s. there are also escape functions in the Database class for proper escaping

modules/datadict/php/datadict.class.inc
```diff
@@ -66,7 +66,7 @@ class Datadict extends \NDB_Menu_Filter
                               );
         $this->query        = " FROM parameter_type pt LEFT JOIN 
                                 parameter_type_override pto USING (Name)
-                 		WHERE pt.Queryable=1";
+                 		        WHERE pt.Queryable=1";
```
PapillonMcGill: lines not all indented equal.
Would also put the `LEFT JOIN` on the second line for clarity

modules/datadict/php/datadict.class.inc
```diff
@@ -113,16 +113,154 @@ class Datadict extends \NDB_Menu_Filter
...
+     * to work are set up.
+     *
+     * Generally, it does the following:
+     *   1. Reset filters (if applicable)
+     *   2. Set up the (new) filters
+     *   3. Set up the search keyword
+     *   4. Set the search order (if applicable)
+     *
+     * The end result should be that $this->filter and $this->having are setup
+     * properly so that the query can be constructed by $this->_getList()
```
PapillonMcGill: missing ending dot.

modules/datadict/php/datadict.class.inc
PapillonMcGill: I second that!

modules/datadict/php/datadict.class.inc
```diff
@@ -150,6 +288,206 @@ class Datadict extends \NDB_Menu_Filter
...
+                    if ($prepared_key != 'pending') {
+                        $qparams['v_'.$prepared_key] = $val;
+                    }
+                }
+            }
+        }
+
+        if (isset($this->searchKeyword)
+            && is_array($this->searchKeyword)
+            && count($this->searchKeyword) > 0
```
PapillonMcGill: These last 2 can be combine in `&& !empty($this->searchKeyword)`

modules/datadict/php/datadict.class.inc
```diff
@@ -150,6 +288,206 @@ class Datadict extends \NDB_Menu_Filter
...
+                'params' => $qparams,
+               );
+    }
+
+    /**
+     * Returns the number of pages that this menu has.
+     *
+     * @param string $query The query used to build the data in the table.
+     *
+     * @return integer The number of pages that the query returns
```
PapillonMcGill: missing ending dot

intralizee:
Reason for closing: New react changes has the filtering on the frontend and makes this PR obsolete.

PR #3688, opened 2018-05-24
[Database] Prevent password appearing in the logs by catchign PDO Exception

johnsaigle:
## This pull request prevents sensitive information from appearing in the error logs. 

It does this by handling the case in which a user has entered invalid MySQL configuration settings. Previously, the uncaught exception would be logged with all the supplied parameters, resulting in the password being logged.

### Example logs before

before

### Example logs after

after


php/libraries/Database.class.inc
```diff
@@ -165,11 +165,20 @@ class Database
...
+                $username,
+                $password
+            );
+        } catch (PDOException $e) {
+            error_log(
+                'Could not establish PDO connection using supplied '
+                .' credentials. Please correct the configuration settings in '
+                .' project/config.xml to valid MySQL credentials.'
+            );
+            exit(1);
```
driusan: The database class (or any class) should never call `exit`. This should throw a new PDOException with a different message rather than using error_log and calling exit..

ZainVirani: tested this, it works, will retest after this change^

johnsaigle:
@ZainVirani Please re-review when you can.

php/libraries/Database.class.inc
```diff
@@ -165,11 +165,20 @@ class Database
...
+        try {
+            $this->_PDO = new PDO(
+                "mysql:host=$host;dbname=$database;charset=UTF8",
+                $username,
+                $password
+            );
+        } catch (PDOException $e) {
+            error_log(
+                'Invalid credentials supplied in the config '
+                . 'file. Could not connect to database Please correct the MySQL'
```
zaliqarosli: full stop between database and Please please.

johnsaigle:
@zaliqarosli I get those errors too -- they arise because there is no DB connection so config settings cannot be loaded. They probably should be caught but not by this PR as this is meant only to address the logging of sensitive information.

php/libraries/Database.class.inc
```diff
@@ -165,11 +165,20 @@ class Database
...
-        );
+        try {
+            $this->_PDO = new PDO(
+                "mysql:host=$host;dbname=$database;charset=UTF8",
+                $username,
+                $password
+            );
+        } catch (PDOException $e) {
+            error_log(
+                'Invalid credentials supplied in the config file. Could not '
```
driusan: a PDOException doesn't mean that there were invalid credentials, it just means that it couldn't connect.

zaliqarosli:
im going to retest, temporarily removing passed manual test label

zaliqarosli:
this is good to go 👍

johnsaigle:
@xlecours The reason I chose to `error_log` and `return` was because **throwing an exception with a sanitized message still includes the stack trace** which displays the password. I explained this after Dave's initial review.

If I do not include the return statement, the very next line will also throw an exception which also displays the password....

In terms of altering the program flow, I'm not sure how this causes any real difference as without a valid DB connection... there's nothing you can do in LORIS anyway. Both before and after this patch the application stops working so there is no change in the experience of using LORIS.

As a result I think this fix is what makes the most sense.

Alternatively, if you know a way of throwing an exception without including the stack trace, please let me know and I'll alter this PR to do so. However, I haven't been able to find a way.

johnsaigle:
@xlecours thanks for changing to an approving review :smile:

zaliqarosli:
happy to retest after rebase

johnsaigle:
This issue has effectively been resolved by https://github.com/aces/Loris/pull/3746 and so I've closed this PR.

PR #3694, opened 2018-05-28
Added On Duplicate Update Function to Database Class

HenriRabalais:
This pull request adds two new methods to the database class: `insertOnDuplicateUpdate()` and `unsafeInsertOnDuplicateUpdate()`

These methods will attempt to `insert` a new row into the database given the query array passed as an argument. However, if one of the values to be inserted is a duplicate value for a unique key column, the query will be changed to an `update` statement for the row where the duplicate was found.

The `unsafe` version of the method will not escape insertion of HTML characters or JSON objects.

Reference:
https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html

This PR also includes a use-case in the media module, which consolidates a few lines of code using the new function.

How to test:
1. Go to `Media` module and click on the `Upload` tab.
2. Upload a file according the naming convention stated at the top of the form.
3. Confirm that the file was uploaded by finding the entry in the `Media` module filter.
5. Click the `edit` button, edit a field (e.g. data) and click `Update`.
6. Ensure that the file was properly edited by checking the entry in the `Media` module filter.

### DIFF
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -300,7 +300,7 @@ class Database
      *
      * This will attempt to insert a row into the database. If there is a duplicate
      * value for any unique key, the row where the duplicate is present will be
-     * updated. HTML from nay field in the row will *not* be automatically escaped.
+     * updated. HTML from any field in the row will *not* be automatically escaped.
      *
      * @param string $table the table into which to insert the row
      * @param array  $set   the values with which to fill the row
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -287,7 +287,7 @@ class Database
      * @param string $table the table into which to insert the row
      * @param array  $set   the values with which to fill the row
      *
-     * @return none
+     * @return void
      */
     public function insertOnDuplicateUpdate($table, $set)
     {
@@ -305,7 +305,7 @@ class Database
      * @param string $table the table into which to insert the row
      * @param array  $set   the values with which to fill the row
      *
-     * @return none
+     * @return void
      */
     public function unsafeInsertOnDuplicateUpdate($table, $set)
     {
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -295,7 +295,7 @@ class Database
     }
 
     /**
-     * Insert row into the database or  update on duplicate key without validating
+     * Insert row into the database or update on duplicate key without validating
      * input.
      *
      * This will attempt to insert a row into the database. If there is a duplicate
@@ -339,8 +339,8 @@ class Database
      *
      * @param string $table             the table into which to insert the row
      * @param array  $set               the values with which to fill the new row
-     * @param bool   $autoescape        determines whether the values to be set 
-                                        should automatically have the html escaped
+     * @param bool   $autoescape        determines whether the values to be set
+     *                                  should automatically have the html escaped
      * @param bool   $ignore            determines whether the insert throws and
      *                                  error or is discarded when value exists in DB
      * @param bool   $onDuplicateUpdate determines whether the row should be updated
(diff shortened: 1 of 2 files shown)


php/libraries/Database.class.inc
```diff
@@ -302,17 +337,30 @@ class Database
...
-     * @param array  $set        the values with which to fill the new row
-     * @param bool   $autoescape determines whether the values to be set should
-     *                           automatically have the html escaped
-     * @param bool   $ignore     determines whether the insert throws and error or
-     *                           is discarded when value exists in DB
+     * @param string $table             the table into which to insert the row
+     * @param array  $set               the values with which to fill the new row
+     * @param bool   $autoescape        determines whether the values to be set
+     *                                  should automatically have the html escaped
+     * @param bool   $ignore            determines whether the insert throws and
```
um4r12: small typo: `determines whether the insert throws and error` -> `determines whether the insert throws an error`.

um4r12: small typo: `determines whether the insert throws and error` -> `determines whether the insert throws an error`.

nicolasbrossard:
Let's say a call to your new method is made in these situations:

1. Record does not exist and is inserted.
2. Record exists and is updated.
3. Record exists but is not updated.

What will be returned by `getLastInsertID()` for each case? Might be worth adding this to the documentation somewhere.

HenriRabalais:
@nicolasbrossard 

Since getLastInsertID() only return the value of the last `insert`, the following would occur:

1. Returns ID of newly inserted row.
2. Does not return ID of updated row.
3. Does not return ID of updated row.

Perhaps it would be useful to have the `insertOnDuplicateKeyUpdate()` function return whether the a row was updated or inserted. 
@driusan what are your thoughts?

php/libraries/Database.class.inc
```diff
@@ -277,6 +277,41 @@ class Database
...
+     * Insert row into the database or update on duplicate key.
+     *
+     * This will attempt to insert a row into the database. If there is a duplicate
+     * value for any unique key, the row where the duplicate is present will be
+     * updated.
+     *
+     * @param string $table the table into which to insert the row
+     * @param array  $set   the values with which to fill the row
+     *
+     * @return none
```
driusan: void

driusan:
Does MySQL or the PDO provide any way to determine if the row was inserted or updated? I'm not sure that it's worth the effort, because I'm not sure what practical use cases it would be needed in, but if it's trivial to add, go ahead.

Assuming that there's no way (short of a second select query) to determine that, I think documenting it in the function comment is sufficient.

HenriRabalais:
Sounds good. Looks like there's ways to go about doing this, but it would required some combination of `ROW_COUNT` and `LAST_INSERT_ID`.

colezweber:
After function rename commit, I was able to edit the date of administration and changes appeared in the module!

victoriafoing:
Passed manual tests for me too.

driusan:
Looks good, but can you send it to 20.0-release branch so that it can be used after the next release instead of needing to wait for 20.1? I think I saw a PR to 20.0 that I wanted to recommend using this for, but then realized it's not merged there..

driusan:
Nevermind, I don't think it's appropriate where I thought in 20.0, so 20.1 is fine.

zaliqarosli:
@driusan hey dave, is this PR ready to go?

php/libraries/Database.class.inc
```diff
@@ -327,6 +375,13 @@ class Database
         $exec_params = array();
         $prepQ      .= $this->_implodeAsPrepared(",", $set, $exec_params);
 
+        if ($onDuplicateUpdate) {
+            $prepQ .= ' ON DUPLICATE KEY UPDATE ';
+            $prepQ .= $this->_implodeAsPrepared(",", $set, $exec_params);
```
zaliqarosli: from looking at line 382, should this line have a space after the comma in quotes? also, could we keep consistent with double and single quotes or is there a reason for the use of both here?

HenriRabalais: No, no space is necessary after the comma here. The `_implodeAsPrepared` function takes care of that. And I left it as a double quotes because the line is identical to line `380`.  If anything, in the context of this class, the single quotes on lines `379`, `381` and `382` look out of place. But I really don't mind either way tbh.

driusan:
I thought this was already merged..

PR #3697, opened 2018-05-30
[Media Module] Filtering candidates displayed on the drop down to the site(s) the user belongs to 

um4r12:
This pull request `attempts to restrict candidates displayed on the media module only to the sites that the user belongs to`. 

To test: 
1. Assign yourself to a set of sites on a non-superuser account. Ensure that the `Media files: Uploading/Downloading/Editing` is checked.
    - If the `access_all_profile` permission is checked, you should be able to see all candidates.
2. Go to the media module, and navigate to the `Upload` tab.
3. In the PSCID drop-down, you should only be able to see candidates that belong to the site(s) that you are assigned to.

See also: [Redmine Ticket 14561](https://redmine.cbrain.mcgill.ca/issues/14561)

### DIFF
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -33,7 +33,7 @@ if (isset($_GET['action'])) {
  */
 function editFile()
 {
-    $db   =& Database::singleton();
+    $db   =& \Database::singleton();
     $user =& User::singleton();
     if (!$user->hasPermission('media_write')) {
         header("HTTP/1.1 403 Forbidden");
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -188,36 +188,37 @@ function getUploadFields()
 {
 
     $db =& Database::singleton();
-
-    $instruments = $db->pselect(
-        "SELECT Test_name FROM test_names ORDER BY Test_name",
-        []
-    );
-    $candidates  = $db->pselect(
-        "SELECT CandID, PSCID FROM candidate ORDER BY PSCID",
-        []
-    );
-
-    $instrumentsList = toSelect($instruments, "Test_name", null);
-    $candidatesList  = toSelect($candidates, "PSCID", null);
-    $candIdList      = toSelect($candidates, "CandID", "PSCID");
-    $visitList       = Utility::getVisitList();
-    $siteList        = Utility::getSiteList(false);
+    $user =& User::singleton();
+    
+    $recordsQuery = "SELECT c.PSCID, c.CandID, s.Visit_Label, ".
+                    "f.Test_name ".
+                    "FROM candidate c ".
+				    "LEFT JOIN session s USING(CandID) ".
+        			"LEFT JOIN flag f ON (s.ID=f.SessionID) ";
+	
+    $siteIDs = $user->getCenterIDs();
+    if (!$user->hasPermission('access_all_profiles')) {
+        $recordsQuery .= "WHERE s.CenterID IN(".implode(",",$siteIDs).") ";
+    }
+    $recordsQuery .= "ORDER BY c.PSCID ASC";
+    $sessionRecords = $db->pselect($recordsQuery, []);
+		
+    $instrumentsList = toSelect($sessionRecords, "Test_name", null);
+    $candidatesList  = toSelect($sessionRecords, "PSCID", null);
+    $candIdList      = toSelect($sessionRecords, "CandID", "PSCID");
+    $visitList       = toSelect($sessionRecords, "Visit_Label", null);
+    error_log(print_r($visitList,1));
+    $siteList        = array();
+    foreach ($siteIDs as $siteID) {
+        $site = Site::singleton($siteID);
+        $siteList[$siteID] = $site->getCenterName();
+    }
     $languageList    = Utility::getLanguageList();
-
     // Build array of session data to be used in upload media dropdowns
     $sessionData    = [];
-    $sessionRecords = $db->pselect(
-        "SELECT c.PSCID, s.Visit_label, s.CenterID, f.Test_name " .
-        "FROM candidate c ".
-        "LEFT JOIN session s USING(CandID) ".
-        "LEFT JOIN flag f ON (s.ID=f.SessionID) ".
-        "ORDER BY c.PSCID ASC",
-        []
-    );
 
     foreach ($sessionRecords as $record) {
-
+        error_log(print_r($record,1));
         // Populate sites
         if (!isset($sessionData[$record["PSCID"]]['sites'])) {
             $sessionData[$record["PSCID"]]['sites'] = [];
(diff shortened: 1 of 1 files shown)


um4r12:
[[Media] Filtering candidates based on user sites](https://trello.com/c/KXjQtapl/6-media-filtering-candidates-based-on-user-sites)

modules/media/ajax/FileUpload.php
```diff
@@ -187,34 +187,36 @@ function uploadFile()
...
+    $recordsQuery = "SELECT c.PSCID, c.CandID, s.CenterID, s.Visit_label, ".
+                    "f.Test_name ".
+                    "FROM candidate c ".
+                    "LEFT JOIN session s USING(CandID) ".
+                    "LEFT JOIN flag f ON (s.ID=f.SessionID) ".
+                    "WHERE c.PSCID NOT LIKE 'scanner%' ".
+                    "AND f.Test_name NOT LIKE 'DDE_%' ";
+
+    $siteIDs = $user->getCenterIDs();
+    if (!$user->hasPermission('access_all_profiles')) {
```
PapillonMcGill: If $user as no permission, what happen? Do all candidates are listed? If yes, this could be exploited in some way!

um4r12: I'm not sure if I understand your question properly, but if $user does not have the 'access_all_profiles' permission, then they can only view candidates from the sites that $user belongs to.  Otherwise, they can view all the candidates.

modules/media/ajax/FileUpload.php
```diff
@@ -347,13 +349,15 @@ function toSelect($options, $item, $item2)
 {
     $selectOptions = [];
 
-    $optionsValue = $item;
+    $optionsVal = $item;
```
nicolasbrossard: This block could be written more concisely like this:

```php
$optionsVal = $item2 ?? $item;
```

modules/media/ajax/FileUpload.php
victoriafoing: If the user does not have permission for certain certain sites, the command $recordsQuery .= "AND s.CenterID IN(".implode(",", $siteIDs).") " eliminates records for those sites.

modules/media/ajax/FileUpload.php
ridz1208: @um4r12 why is this being changed ?

um4r12: @ridz1208 Before I had added the DDE filter in the query, the left join would produce some null records, and this resulted into empty select options for one of the fields. So I added an if statement to ensure that the select options were not empty, but with the current variable names, I was exceeding the 80 char limit with the indentation. Shortened it to meet that requirement.

um4r12: @nicolasbrossard done

modules/media/ajax/FileUpload.php
```diff
@@ -347,13 +349,12 @@ function toSelect($options, $item, $item2)
 {
     $selectOptions = [];
 
-    $optionsValue = $item;
-    if (isset($item2)) {
-        $optionsValue = $item2;
-    }
+    $optionsVal = $item2 ?? $item;
```
ridz1208: @um4r12 please update the php version at the top of the file to 7

ridz1208:
@nicolasbrossard please dismiss/re-review

modules/media/ajax/FileUpload.php
um4r12: @ridz1208 done.

um4r12:
@ridz1208 @kongtiaowang I don't know why Travis keeps failing. It's failing at this test:
`/home/travis/build/aces/Loris/modules/imaging_uploader/test/imaging_uploaderTest.php:178`.
I haven't made any changes to the imaging uploader tho. Any ideas?

driusan:
But it isn't failing?

kongtiaowang:
@um4r12 I re-start the travis again, It passed this time. Sometimes, too many PRs run in the travis at the same moment. Travis doesn't have enough processing time.

modules/media/ajax/FileUpload.php
```diff
@@ -187,34 +187,36 @@ function uploadFile()
...
+                    "f.Test_name ".
+                    "FROM candidate c ".
+                    "LEFT JOIN session s USING(CandID) ".
+                    "LEFT JOIN flag f ON (s.ID=f.SessionID) ".
+                    "WHERE c.PSCID NOT LIKE 'scanner%' ".
+                    "AND f.Test_name NOT LIKE 'DDE_%' ";
+
+    $siteIDs = $user->getCenterIDs();
+    if (!$user->hasPermission('access_all_profiles')) {
+        $recordsQuery .= "AND s.CenterID IN(".implode(",", $siteIDs).") ";
```
driusan: I think this is safe since $siteIDs comes from `$user->getCenterIDs();` above, but we usually use `FIND_IN_SET` so that we can use prepared statements.

For instance, see: https://github.com/aces/Loris/blob/aae05e7fbdd537f8535838377e80469b86ef562a/modules/data_team_helper/php/data_team_helper.class.inc#L247-L248 (I'm not sure if that's the best example, it's just the first one that showed up when searching on GitHub

ZainVirani: this is actually good to know. didn't realize `IN` was not usable in prepared statements

modules/media/ajax/FileUpload.php
```diff
@@ -187,34 +187,36 @@ function uploadFile()
...
-        []
-    );
-
-    $instrumentsList = toSelect($instruments, "Test_name", null);
-    $candidatesList  = toSelect($candidates, "PSCID", null);
-    $candIdList      = toSelect($candidates, "CandID", "PSCID");
-    $visitList       = Utility::getVisitList();
-    $siteList        = Utility::getSiteList(false);
-    $languageList    = Utility::getLanguageList();
+    $db   =& Database::singleton();
```
zaliqarosli: i think going forward we're using \NDB_Factory::singleton()->database() and namespacing

zaliqarosli:
i think this is still relevent? https://github.com/aces/Loris/pull/3697#pullrequestreview-129677696

um4r12:
@driusan @zaliqarosli  Made changes as suggested.

@ZainVirani Could you re-test and see that there aren't any erroneous php notices being displayed by trying to access from both a regular and admin user?

modules/media/ajax/FileUpload.php
um4r12: @driusan Made changes as suggested.

zaliqarosli:
@um4r12 could you update this discussion thread with update on steps moving forward for the auto-filling of the 'Site' field?

I will test and pass if everything nothing's changed from last week

zaliqarosli:
this ticket (https://redmine.cbrain.mcgill.ca/issues/11681) somewhat relevant. further discussion needed

zaliqarosli:
i will take over this PR

zaliqarosli:
replaced by #4084

PR #3698, opened 2018-06-01
[SQL] make computeDeepQC configurable

llevitis:
This PR adds `ComputeDeepQC` as a configurable option under the Imaging Pipelines section of the configuration module. An accompanying change will be added to https://github.com/aces/Loris-MRI/pull/306 such that a call to the DeepQC app and subsequent automatic QC prediction will only computed if `ComputeDeepQC` is set to yes.


MounaSafiHarab:
@driusan @cmadjar 

this is making it into 20.0 but the MRI side is not ready

driusan:
That's fine, I don't think there's any harm in it for now if it's going to be used relatively soon.

PR #3699, opened 2018-06-04
[SQL] Adding DeepQC to parameter_type table 

llevitis:
This PR adds `DeepQC` as a parameter type. In https://github.com/aces/Loris-MRI/pull/306, the most up-to date DeepQC probability is returned and used to populate or update the DeepQC value corresponding to a specific file.

### DIFF
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -1496,6 +1496,8 @@ INSERT INTO parameter_type (Name, Type, Description, RangeMin, RangeMax, SourceF
   ('Visit_label','varchar(255)','Visit_label',null,null,'visit_label','session',null,1,null),
   ('candidate_dob','date','Candidate_Dob',null,null,'DoB','candidate',null,1,null);
 
+INSERT INTO parameter_type (Name, Type) Values ("DeepQC", "double");
+
 CREATE TABLE `parameter_type_category` (
   `ParameterTypeCategoryID` int(11) unsigned NOT NULL auto_increment,
   `Name` varchar(255) default NULL,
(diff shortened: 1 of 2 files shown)


xlecours:
Hi @llevitis , do you have a PR where we can see the bigger picture? I guess this is for the DeepQC pipeline result reinsertion. Is there an insertion script, an API endpoint or an uploader that will populate the parameter tables?

llevitis:
> Hi @llevitis , do you have a PR where we can see the bigger picture? I guess this is for the DeepQC pipeline result reinsertion. Is there an insertion script, an API endpoint or an uploader that will populate the parameter tables?
Yeah! I link the LORIS-MRI in this PR's description. The flask app file is here: https://github.com/aces/DeepQC/blob/flask_app/flask_app/deepqc_flask.py.

SQL/0000-00-00-schema.sql
```diff
@@ -1496,6 +1496,8 @@ INSERT INTO parameter_type (Name, Type, Description, RangeMin, RangeMax, SourceF
   ('Visit_label','varchar(255)','Visit_label',null,null,'visit_label','session',null,1,null),
   ('candidate_dob','date','Candidate_Dob',null,null,'DoB','candidate',null,1,null);
 
+INSERT INTO parameter_type (Name, Type) Values ("DeepQC", "double");
```
MounaSafiHarab: Hi @llevitis 

I am not sure you even need this mysql patch.
On the MRI side, the utilities do this automatically for you...
am I not right @cmadjar?

cmadjar: @llevitis @MounaSafiHarab I agree with Mouna. This will automatically be done by the imaging pipeline. No need to create a patch here.

PR #3700, opened 2018-06-05
[document_repository] Fixed multiple files with same name when uploading to document repository

maltheism:
This pull request allow files with duplicate name to exist and when user uploads a file to the document repository.

I made the mysql column "uuid" in the table "document_repository" for files to be stored in unique directories.

The uuid and file_name are used for the system directory structure when storing the files.
System Directory Structure Examples:

Note: {uuid} is unique per upload.
user_uploads/admin/text.png/{uuid}/text.png
user_uploads/admin/text.txt/{uuid}/text.txt

user_uploads/admin/text.png/{uuid}/text.png
user_uploads/admin/text.txt/{uuid}/text.txt

See also: Bug #10847 and Bug #11243

### DIFF
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -864,6 +864,7 @@ CREATE TABLE `document_repository` (
   `EARLI` tinyint(1) DEFAULT '0',
   `hide_video` tinyint(1) DEFAULT '0',
   `File_category` int(3) unsigned DEFAULT NULL,
+  `UUID` varchar(36) DEFAULT '',
   PRIMARY KEY (`record_id`),
   KEY `fk_document_repository_1_idx` (`File_category`),
   CONSTRAINT `fk_document_repository_1` FOREIGN KEY (`File_category`) REFERENCES `document_repository_categories` (`id`) ON DELETE NO ACTION ON UPDATE NO ACTION
--- a/SQL/Archive/19.2/2018-06-05_Add_UUID_column_to_document_repository.sql
+++ b/SQL/Archive/19.2/2018-06-05_Add_UUID_column_to_document_repository.sql
@@ -1,0 +1,1 @@
+ALTER TABLE document_repository ADD `UUID` varchar(36) DEFAULT '';
--- a/modules/document_repository/ajax/documentDelete.php
+++ b/modules/document_repository/ajax/documentDelete.php
@@ -10,18 +10,23 @@
   * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
   * @link     https://github.com/aces/Loris
   */
+
 $user =& User::singleton();
+
 if (!$user->hasPermission('document_repository_delete')) {
     header("HTTP/1.1 403 Forbidden");
     exit;
 }
 
 set_include_path(get_include_path().":../../project/libraries:../../php/libraries:");
+
 require_once "NDB_Client.class.inc";
 require_once "NDB_Config.class.inc";
 require_once "Email.class.inc";
+
 $client = new NDB_Client();
 $client->initialize("../../project/config.xml");
+
 $factory = NDB_Factory::singleton();
 $baseURL = $factory->settings()->getBaseURL();
 
@@ -66,5 +71,19 @@ $path = __DIR__ . "/../user_uploads/$dataDir";
 if (file_exists($path)) {
     unlink($path);
 }
+
+// Cleanup empty directories
+set_error_handler(
+    function () {
+        // Silence the E_WARNING when files exist in the directory.
+    }
+);
+$rm_directory = __DIR__ . '/../user_uploads/'
+    . substr($dataDir, 0, strlen($dataDir)-(strlen($fileName)+1));
+rmdir($rm_directory);
+$rm_directory = __DIR__ . '/../user_uploads/'
+    . $userName . '/' . $fileName;
+rmdir($rm_directory);
+restore_error_handler();
 
 ?>
--- a/modules/document_repository/ajax/documentEditUpload.php
+++ b/modules/document_repository/ajax/documentEditUpload.php
@@ -25,15 +25,17 @@ set_include_path(
 require_once "NDB_Client.class.inc";
 require_once "NDB_Config.class.inc";
 require_once "Email.class.inc";
+
 $client = new NDB_Client();
 $client->initialize("../../project/config.xml");
+
 $factory = NDB_Factory::singleton();
 $baseURL = $factory->settings()->getBaseURL();
 
-$config = NDB_Config::singleton();
-
-// create Database object
-$DB =& Database::singleton();
+// Setup Database object.
+$config    =& \NDB_Config::singleton();
+$db_config = $config->getSetting('database');
+$db        =& \Database::singleton();
 
 $editNotifier = new NDB_Notifier(
     "document_repository",
@@ -58,26 +60,61 @@ if ($userSingleton->hasPermission('document_repository_view')
         $instrument = $_POST['instrument'] !== '' ? $_POST['instrument'] : null;
         $pscid      = $_POST['pscid']      !== '' ? $_POST['pscid'] : null;
         $visit      = $_POST['visit']      !== '' ? $_POST['visit'] : null;
-        $comments   = $_POST['comments']   !== '' ? $_POST['commnets'] : null;
+        $comments   = $_POST['comments']   !== '' ? $_POST['comments'] : null;
         $version    = $_POST['version']    !== '' ? $_POST['version'] : null;
+        $uuid       = uuid4();
 
         $fileSize = $_FILES["file"]["size"];
         $fileName = $_FILES["file"]["name"];
-        $fileType = end((explode(".", $fileName)));
+        $fileType = '';
+        // Handle retrieving the file type.
+        if (preg_match('/\./', $fileName)) {
+            $pos = strrpos($fileName, '.', -1);
+            if ($pos+1 != strlen($fileName)) {
+                $fileType = substr(
+                    $fileName,
+                    strrpos($fileName, '.', -1)+1
+                );
+            }
+        }
+        $sql_statement = $db->prepare(
+            'SELECT File_name, version FROM document_repository '
+            .'WHERE File_name=? AND uploaded_by=?'
+        );
+        $sql_statement->bindParam(1, $fileName, PDO::PARAM_STR);
+        $sql_statement->bindParam(2, $puser, PDO::PARAM_STR);
+        $sql_statement->execute();
+        $sql_result = $sql_statement->fetchAll(PDO::FETCH_ASSOC);
 
         // __DIR__ is the document_repository ajax directory
         // when this script is executing. Go up a level to the
         // document_repository module directory, and use a
         // user_uploads directory as a base for user uploads
-        $base_path = __DIR__ . "/../user_uploads/";
-        $fileBase  = $puser . "/" . $fileName;
+        $base_path = realpath(__DIR__ . '/..') . '/user_uploads/';
+        $fileBase  = $puser . '/'
+            . $fileName
+            . '/' . $uuid
+            . '/' . $fileName;
 
+        // Create user directory /base_path/user
         if (!file_exists($base_path . $puser)) {
-            mkdir($base_path . $puser, 0777);
+            mkdir($base_path . $puser, 0770);
         }
-
-
-        $target_path = $base_path  . $fileBase;
+        // Create filename directory /base_path/user/fileName
+        if (!file_exists($base_path . $puser . '/' . $fileName)) {
+            mkdir($base_path . $puser . '/' . $fileName, 0770);
+        }
+        // Create uuid directory /base_path/user/fileName/uuid
+        if (!file_exists(
+            $base_path . $puser . '/' . $fileName . '/' . $uuid
+        )
+        ) {
+            mkdir(
+                $base_path . $puser . '/' . $fileName . '/' . $uuid,
+                0770
+            );
+        }
+        $target_path = $base_path . $fileBase;
 
         if (move_uploaded_file($_FILES["file"]["tmp_name"], $target_path)) {
             $success = $DB->insert(
@@ -95,10 +132,10 @@ if ($userSingleton->hasPermission('document_repository_view')
                  'PSCID'         => $pscid,
                  'visitLabel'    => $visit,
                  'File_type'     => $fileType,
+                 'UUID'          => $uuid,
                 )
             );
-            $msg_data['newDocument']
-                = $baseURL . "/document_repository/";
+            $msg_data['newDocument'] = $baseURL . "/document_repository/";
             $msg_data['document']    = $fileName;
 
             $uploadNotifier->notify($msg_data);
@@ -110,7 +147,7 @@ if ($userSingleton->hasPermission('document_repository_view')
         } else {
             echo "There was an error uploading the file";
         }
-    } elseif ($action == 'edit') {
+    } else if ($action == 'edit') {
         $id         = $_POST['idEdit'];
         $category   = $_POST['categoryEdit'];
         $instrument = $_POST['instrumentEdit'];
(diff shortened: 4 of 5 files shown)


modules/document_repository/ajax/documentEditUpload.php
```diff
@@ -25,15 +25,22 @@
...
 $baseURL = $factory->settings()->getBaseURL();
 
-$config = NDB_Config::singleton();
-
-// create Database object
-$DB =& Database::singleton();
+// Setup Database object.
+$config    =& \NDB_Config::singleton();
+$db_config = $config->getSetting('database');
+$db        = \Database::singleton(
```
um4r12: Why is there a initialization to the db class with parameters? Can't it be done just with `$db =& \Database::singleton();`?
Tested it out on my instance, and I was getting an error, and when replaced with the above, it worked fine.

modules/document_repository/ajax/documentEditUpload.php
```diff
@@ -60,24 +67,63 @@
...
+        $sql_result = $sql_statement->fetchAll(PDO::FETCH_ASSOC);
 
         // __DIR__ is the document_repository ajax directory
         // when this script is executing. Go up a level to the
         // document_repository module directory, and use a
         // user_uploads directory as a base for user uploads
-        $base_path = __DIR__ . "/../user_uploads/";
-        $fileBase  = $puser . "/" . $fileName;
+        $base_path = __DIR__ . '/../user_uploads/';
+        $fileBase  = $puser . '/docs_repo/'
```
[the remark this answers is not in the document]
intralizee: The user_uploads is used elsewhere besides the docs_repo. So the extra nest is beneficial.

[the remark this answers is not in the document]
ZainVirani: I agree with @um4r12 
the user_uploads folder is used in other modules but the folder is a sub-directory of document_repository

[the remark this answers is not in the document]
johnsaigle: @intralizee There are two `user_upload` folders: one a subdir of doc_repo and the other a subdir of `date_release` (which was based on doc_repo)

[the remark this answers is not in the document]
johnsaigle: Also, every other module (that stores data) stores data under `/data/.../...`

kongtiaowang:
@intralizee If you update 0000-00-schema.sql , you need to update the testing data also.
Adding a "uuid" column for the document_repository table.

kongtiaowang:
@intralizee Please updating your new data here "test/RBdata.sql" with the new document_repository table.

modules/document_repository/ajax/documentEditUpload.php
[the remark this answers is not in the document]
intralizee: @johnsaigle @ZainVirani @um4r12 I'll get back to guys with a more concise response but one of the modules I tested before working on document_repository had uploaded a file or folder to the user_upload directory and resulted in a conflict when uploading a file from document_repository with the same name of the folder or directory previously created.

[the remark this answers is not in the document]
intralizee: Okay I must have been mistaken previously. Next commit removes the docs_repo as unnecessary.

modules/document_repository/ajax/documentEditUpload.php
```diff
@@ -58,26 +60,61 @@
...
+        // Create user directory /base_path/user
         if (!file_exists($base_path . $puser)) {
             mkdir($base_path . $puser, 0777);
         }
-
-
-        $target_path = $base_path  . $fileBase;
+        // Create filename directory /base_path/user/fileName
+        if (!file_exists($base_path . $puser . '/' . $fileName)) {
+            mkdir($base_path . $puser . '/' . $fileName, 0777);
```
johnsaigle: This would be better with a `0770` permission. Same goes for the folder below.

The discussion seems to have disappeared but @ridz1208 and I discussed upload permissions for the document repository in https://github.com/aces/Loris/pull/3603

modules/document_repository/ajax/documentEditUpload.php
```diff
@@ -147,4 +184,26 @@
...
+ *
+ * Source from comments:
+ * http://php.net/manual/en/function.com-create-guid.php
+ * Maybe move to Utilities class.
+ *
+ * @return string $version
+ */
+function uuid4()
+{
+    if (function_exists('com_create_guid') === true) {
```
johnsaigle: We don't support Windows so I think this is a potentially confusing comment.

modules/document_repository/ajax/documentEditUpload.php
```diff
@@ -147,4 +184,26 @@
...
+/**
+ * Create a UUID v4 string.
+ *
+ * Source from comments:
+ * http://php.net/manual/en/function.com-create-guid.php
+ * Maybe move to Utilities class.
+ *
+ * @return string $version
+ */
+function uuid4()
```
johnsaigle: We're trying to include [return type declarations](https://secure.php.net/manual/en/functions.returning-values.php#functions.returning-values.type-declaration) for new code. 

Could you add `String` as a return type here?

colezweber:
Hit a snag with “There was an error uploading the file” but it's probably an issue with my permissions RE: move_uploaded_file method as discussed. Will update.

SQL/0000-00-00-schema.sql
```diff
@@ -864,6 +864,7 @@ CREATE TABLE `document_repository` (
   `EARLI` tinyint(1) DEFAULT '0',
   `hide_video` tinyint(1) DEFAULT '0',
   `File_category` int(3) unsigned DEFAULT NULL,
+  `uuid` varchar(36) DEFAULT '',
```
zaliqarosli: @PapillonMcGill recently set up a standard for all new additions to our SQL database (https://github.com/aces/Loris/blob/19.0-dev/docs/SQLModelingStandard.md). I think uuid should maybe be UUID? or something a bit more descriptive? melanie, what do you think?

PapillonMcGill: @zaliqarosli @intralizee 
The standards say that `The acronym should only contain upper case letter.` So capitalized is the way to go. UUID might not be fully known by neuroimaging community (yet) but software developer within knows it well.

zaliqarosli: @intralizee i had to look up what UUID meant 😝 but i think its fine here

SQL/Archive/19.2/2018-06-05_Add_UUID_column_to_document_repository.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE document_repository ADD `UUID` varchar(36) DEFAULT '';
```
zaliqarosli: maybe this patch should go into just the Archive directory? we're now on the 20 release. also, patches usually get moved to the version folders after a release if on minor

modules/document_repository/ajax/documentEditUpload.php
```diff
@@ -58,26 +60,61 @@
...
         // when this script is executing. Go up a level to the
         // document_repository module directory, and use a
         // user_uploads directory as a base for user uploads
-        $base_path = __DIR__ . "/../user_uploads/";
-        $fileBase  = $puser . "/" . $fileName;
+        $base_path = realpath(__DIR__ . '/..') . '/user_uploads/';
+        $fileBase  = $puser . '/'
+            . $fileName
+            . '/' . $uuid
+            . '/' . $fileName;
```
PapillonMcGill: $filename is used twice. Is it supposed to be like that?

SQL/Archive/19.2/2018-06-05_Add_UUID_column_to_document_repository.sql
um4r12: While we are at this, when would you submit the patches into the `New_patches` dir?

modules/document_repository/ajax/documentEditUpload.php
um4r12: I believe the the first `$fileName` is referring to the directory, whereas the second one is referring to the actual file.
ex (from above). `user_uploads/admin/text.png/{uuid}/text.png`

modules/document_repository/ajax/documentEditUpload.php
```diff
@@ -58,26 +60,61 @@
...
         // __DIR__ is the document_repository ajax directory
         // when this script is executing. Go up a level to the
         // document_repository module directory, and use a
         // user_uploads directory as a base for user uploads
-        $base_path = __DIR__ . "/../user_uploads/";
-        $fileBase  = $puser . "/" . $fileName;
+        $base_path = realpath(__DIR__ . '/..') . '/user_uploads/';
+        $fileBase  = $puser . '/'
+            . $fileName
+            . '/' . $uuid
```
johnsaigle: Please put all the front-slashes at the front of the line or the back of the line for consistency.

modules/document_repository/ajax/documentDelete.php
```diff
@@ -67,4 +72,18 @@
     unlink($path);
 }
 
+// Cleanup empty directories
+set_error_handler(
+    function () {
+        // Silence the E_WARNING when files exist in the directory.
```
johnsaigle: I don't know if this is the best way to do this. Shouldn't we delete the files first in this case? I'm a little worried about this silencing useful warnings.

modules/document_repository/ajax/documentDelete.php
```diff
@@ -67,4 +72,18 @@
...
+    }
+);
+$rm_directory = __DIR__ . '/../user_uploads/'
+    . substr($dataDir, 0, strlen($dataDir)-(strlen($fileName)+1));
```
johnsaigle: Are you trying to chop off the filename from `$dataDir` here? If so, just use [`dirname()`](https://secure.php.net/manual/en/function.dirname.php);

modules/document_repository/ajax/documentEditUpload.php
```diff
@@ -58,26 +60,61 @@
...
-        $comments   = $_POST['comments']   !== '' ? $_POST['commnets'] : null;
+        $comments   = $_POST['comments']   !== '' ? $_POST['comments'] : null;
         $version    = $_POST['version']    !== '' ? $_POST['version'] : null;
+        $uuid       = uuid4();
 
         $fileSize = $_FILES["file"]["size"];
         $fileName = $_FILES["file"]["name"];
-        $fileType = end((explode(".", $fileName)));
+        $fileType = '';
+        // Handle retrieving the file type.
```
johnsaigle: This is better done using [`pathinfo()`](https://secure.php.net/manual/en/function.pathinfo.php)

It's also more accurate to say that you are grabbing the uploaded file's extension. Properly getting the file type would involve looking at the file signatures.

modules/document_repository/ajax/documentEditUpload.php
```diff
@@ -58,26 +60,61 @@
...
+        $sql_statement->execute();
+        $sql_result = $sql_statement->fetchAll(PDO::FETCH_ASSOC);
 
         // __DIR__ is the document_repository ajax directory
         // when this script is executing. Go up a level to the
         // document_repository module directory, and use a
         // user_uploads directory as a base for user uploads
-        $base_path = __DIR__ . "/../user_uploads/";
-        $fileBase  = $puser . "/" . $fileName;
+        $base_path = realpath(__DIR__ . '/..') . '/user_uploads/';
```
johnsaigle: Would you mind changing this to `$upload_path`? The "base path" in LORIS typically refers to the Config setting corresponding to the web root.

johnsaigle:
This PR is tagged as a Bug Fix, issued to minor, but in my opinion constitutes a new feature. @intralizee can you weigh in here as to where you think this PR should go?

SQL/Archive/19.2/2018-06-05_Add_UUID_column_to_document_repository.sql
zaliqarosli: @um4r12 when you are on a branch that has these new directories. i'm going to bring this up in tomorrow's meeting :)

zaliqarosli:
> Note: this code should not go live before bugfix is merged into minor during the next release cycle. The reason is that it does not contain the security fix issued in #3678.

$puser is a user-controllable value and its expanded usage in this PR allows for arbitrary file upload attacks.

I just wanted to leave a note here to the effect that these branches need to be synchronized.

should we maybe add a blocked label to this PR until the needed merging happens? @intralizee

driusan:
Sorry to be jumping into this so late but.. why are we even using UUIDs for this? A few bytes of random data encoded in some way (hex, base64, whatever) should be enough to fix the issue that this is trying to solve.

driusan:
(Another solution--probably more robust solution-- would be to use a content addressable naming convention and the backend and include a Content-Disposition header to associate it to a friendly filename when a user tries to download it from the frontend.)

johnsaigle:
@driusan I +1 the idea of using another approach. uuid feels like overkill. 

>content addressable naming convention

If I understand correctly, you mean something like storing the hash of the file contents mapped to a friendly name in the back-end, right? E.g.

---

#### The Database: 
| Friendly Name        |  Actual Filename
| ------------- |:-------------:|
`my_meme.png` | `bb1a5fd034191b49f094161e53e91f675430743b`

---

#### Frontend:

> Welcome to the Document Repository

* my_meme.png <[Download](url)>

---

#### Backend

* `/var/www/loris/modules/document_repository/user_uploads/john/bb1a5fd034191b49f094161e53e91f675430743b.png`

ZainVirani:
It seems like some discussion is still required on this PR, so I also want to bring the comments on the previous [[Closed] PR](https://github.com/aces/Loris/pull/3662) into the mix

modules/document_repository/ajax/documentEditUpload.php
intralizee: @PapillonMcGill um4r12 is correct and so it's supposed to be like that.

johnsaigle:
@ZainVirani (we talked this over i person but I wanted to reply here so others could see).

I don't like the counter solution because it makes it very easy for users to upload duplicates and waste space. The hashing method has a built-in ease of detecting duplicate files (if the hash is the same, it's a duplicate, reject the upload)

With hashing, the users can still upload files of the same name as their contents will differ. So long as they use the categorization feature, files of different types will be separated into the right trees so they won't get mixed up.

If users need multiple revisions of the same document and the same name, they can use data_release which has more robust versioning. I feel like the counter scheme by contrast leads to messiness.

intralizee:
Okay I feel like this PR needs to be discussed in person. I spent a decent amount of time planning how I would handle the problem solved by the PR before writing the code. I don't see my method of doing this as overkill and there are overlooked possibilities & benefits of multiple ways of storing files. One reason I don't see it as beneficial for not allowing multiple files (with same content & name) is when there are multiple sites and users with permissions factoring into the equation of reality. I don't believe this needs to be handled like google drive or dropbox but the edge cases are fun to think about solving.

driusan:
@johnsaigle yes, that's exactly what I meant by a content addressable naming convention.

@intralizee It's not the approach of using a random part of the path that I think is overkill, it's just the version 4 UUIDs (which comes with the overhead of having to explain what UUIDs are to anyone who looks at the code, ensure that the generation/formatting/etc is compliant with the standard, decide what to do if someone uses a different kind of UUID, etc) when any meaningless random number would do (and I doubt we need a whole 16 bytes to avoid collisions.)

driusan:
@ZainVirani Your approach of having a counter would result in race conditions, while @intralizee's random approach doesn't have that problem (if the randomness source has enough entropy.)

HenriRabalais:
Since changes need to be made to this PR, I am removing the `passed manual test` label.

modules/document_repository/ajax/documentEditUpload.php
```diff
@@ -58,26 +60,61 @@
...
+        ) {
+            mkdir(
+                $base_path . $puser . '/' . $fileName . '/' . $uuid,
+                0770
+            );
+        }
+        $target_path = $base_path . $fileBase;
 
         if (move_uploaded_file($_FILES["file"]["tmp_name"], $target_path)) {
             $success = $DB->insert(
```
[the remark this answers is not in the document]
Jkat: (and same with line 174 & 176)

johnsaigle:
Note that this will not be able to be merged when #3971 goes through. However we should still have this feature at some point.

maltheism:
I think I'll close this PR for now because of the PR involving the react rewrite. That PR has made notable changes and where this PR is obsolete. I'll see what can be done after that PR is merged into LORIS.

PR #3670, opened 2018-05-16
API now displays QC info of MINC images, and allow hyphens in MINC filenames: Redmine 14493

MounaSafiHarab:
The API was **not** 
1) displaying QC info for MINC images because it was trying to find the file Qc.php; however the file is called QC.php. So change this in the .htaccess file to be consistent with the codebase filenaming.
and
2) allowing MINC files to have hyphen in them (which might happen if the Loris-MRI project had a hyphen in it). So allow hyphens.

PR #3703, opened 2018-06-05
[API] fix typo in .htaccess for image qc

xlecours:
This fix a typo in .htaccess for image qc that prevent the endpoint from been reached.


MounaSafiHarab:
I was faster than you @xlecours 
#3670 :)

driusan:
Duplicate of fix already in bugfix, as @MounaSafiHarab pointed out..

PR #3705, opened 2018-06-07
[media] Remove unused code

ridz1208:
This removes the `validFilters` and `formToFilter` variables which seem to be unused when using react

@driusan are these actually not used ?


driusan:
If they're fully converted to react with live filtering, they're unused, otherwise they're used to map the form to the SQL generated to populate the table.

cmadjar:
@ridz1208 FYI failing our good friend Travis

ridz1208:
@cmadjar thx, re-running it now

PR #3706, opened 2018-06-11
[New Profile] Ensure Site object isn't overwritten

ZainVirani:
In the case of user defined PSCID, candidates were not able to be created due to a variable holding a reference to the Site object being overwritten.

To test this fix, set PSCID to `user` in your `config.xml`, and attempt to create a new candidate.

Clarification: Line 277 of the file changed assigns `$site` by reference, and then reassigns it on line 302. This causes the site object's value to change to `$value['psc']` which is potentially dangerous.

### DIFF
--- a/modules/new_profile/php/new_profile.class.inc
+++ b/modules/new_profile/php/new_profile.class.inc
@@ -299,10 +299,10 @@ class New_Profile extends \NDB_Form
         }
 
         // validate site entered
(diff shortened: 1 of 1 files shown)


colezweber:
I'm getting an error but it seems IBIS specific as it is citing the project libraries such as Candidate.class and LorisForm.class...

PR #3014, opened 2017-08-02
API v0.0.3; users at multiple sites now can create candidates and timepoints: Redmine 12850

MounaSafiHarab:
Users affiliated with multiple sites can now create candidates and timepoints (provided they specify the SiteName).

PR #3708, opened 2018-06-11
[API] Bug fix for visit creation for users at multiple sites

davidblader:
bug introduced in: #3014 
when putting an imploded comma separated array through a prepared select statement, i am guessing that only the first element of the array goes through when applied within a `WHERE = ...` clause.
i changed the ` CenterID WHERE = :cid` to `WHERE FIND_IN_SET(CenterID, :cid)` to allow for all center ID's to be considered


driusan:
Can you provide a better description of what the bug is/what it's fixing?

PR #3709, opened 2018-06-11
[Tools] updating date zeros removal script

llevitis:
This PR updates the `DB_date_zeros_removal.php` script. At present, the script doesn't check if not nullable fields are date times or timestamps, so this is modified.

### DIFF
--- a/tools/DB_date_zeros_removal.php
+++ b/tools/DB_date_zeros_removal.php
@@ -106,11 +106,18 @@ foreach ($field_names as $key=>$field)
             " SET ".$db->escape($field['COLUMN_NAME'])."=NULL".$autoUpdateSQL.
             " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00 00:00:00';\n";
     } else {
-	echo "COLUMN ".$field['COLUMN_NAME']." in TABLE ".$field['TABLE_NAME']." is NOT NULLABLE. ".
-	    "A date '1000-01-01' will be entered instead of '0000-00-00' values.\n"; 
-        $nonNullUpdates .= "UPDATE ".$db->escape($database['database']).".".$db->escape($field['TABLE_NAME']).
-            " SET ".$db->escape($field['COLUMN_NAME'])."='1000-01-01'".$autoUpdateSQL.
-            " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00';\n";
+        echo "COLUMN ".$field['COLUMN_NAME']." in TABLE ".$field['TABLE_NAME']." is NOT NULLABLE. ".
+            "A date '1000-01-01' will be entered instead of '0000-00-00' values.\n";
+
+        if ($field['DATA_TYPE'] == 'date') {
+            $nonNullUpdates .= "UPDATE ".$db->escape($database['database']).".".$db->escape($field['TABLE_NAME']).
+                " SET ".$db->escape($field['COLUMN_NAME'])."='1000-01-01'".$autoUpdateSQL.
+                " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00';\n";
+        } else if (($field['DATA_TYPE'] == 'datetime' || $field['DATA_TYPE'] == 'timestamp')) {
+            $nonNullUpdates .= "UPDATE ".$db->escape($database['database']).".".$db->escape($field['TABLE_NAME']).
+                " SET ".$db->escape($field['COLUMN_NAME'])."='1000-01-01 00:00:00'".$autoUpdateSQL.
+                " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00 00:00:00';\n";
+        }
     }
 
 }
--- a/tools/DB_date_zeros_removal.php
+++ b/tools/DB_date_zeros_removal.php
@@ -113,11 +113,15 @@ foreach ($field_names as $key=>$field)
             $nonNullUpdates .= "UPDATE ".$db->escape($database['database']).".".$db->escape($field['TABLE_NAME']).
                 " SET ".$db->escape($field['COLUMN_NAME'])."='1000-01-01'".$autoUpdateSQL.
                 " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00';\n";
-        } else if (($field['DATA_TYPE'] == 'datetime' || $field['DATA_TYPE'] == 'timestamp')) {
+        } else if ($field['DATA_TYPE'] == 'datetime') {
             $nonNullUpdates .= "UPDATE ".$db->escape($database['database']).".".$db->escape($field['TABLE_NAME']).
                 " SET ".$db->escape($field['COLUMN_NAME'])."='1000-01-01 00:00:00'".$autoUpdateSQL.
                 " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00 00:00:00';\n";
-        }
+        } else if ($field['DATA_TYPE'] == 'timestamp') { 
+            $nonNullUpdates .= "UPDATE ".$db->escape($database['database']).".".$db->escape($field['TABLE_NAME']).
+                " SET ".$db->escape($field['COLUMN_NAME'])."='1970-01-01 00:00:01'".$autoUpdateSQL.
+                " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00 00:00:00';\n";
+        }    
     }
 
 }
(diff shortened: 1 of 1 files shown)


tools/DB_date_zeros_removal.php
```diff
@@ -113,11 +113,15 @@
             $nonNullUpdates .= "UPDATE ".$db->escape($database['database']).".".$db->escape($field['TABLE_NAME']).
                 " SET ".$db->escape($field['COLUMN_NAME'])."='1000-01-01'".$autoUpdateSQL.
                 " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00';\n";
-        } else if (($field['DATA_TYPE'] == 'datetime' || $field['DATA_TYPE'] == 'timestamp')) {
+        } else if ($field['DATA_TYPE'] == 'datetime') {
             $nonNullUpdates .= "UPDATE ".$db->escape($database['database']).".".$db->escape($field['TABLE_NAME']).
                 " SET ".$db->escape($field['COLUMN_NAME'])."='1000-01-01 00:00:00'".$autoUpdateSQL.
                 " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00 00:00:00';\n";
```
zaliqarosli: should we change the 00:00:00 here as well to make it not zero?

tools/DB_date_zeros_removal.php
```diff
@@ -113,11 +113,15 @@
...
-        }
+        } else if ($field['DATA_TYPE'] == 'timestamp') { 
+            $nonNullUpdates .= "UPDATE ".$db->escape($database['database']).".".$db->escape($field['TABLE_NAME']).
+                " SET ".$db->escape($field['COLUMN_NAME'])."='1970-01-01 00:00:01'".$autoUpdateSQL.
```
zaliqarosli: how come we're setting a different date here to what's set above?

PapillonMcGill:
Does are server have the ALLOW_INVALID_DATES flag enabled? 
If yes, there is no need to set date to the first day of the appropriate range.

tools/DB_date_zeros_removal.php
PapillonMcGill: @zaliqarosli Datetime and Timestamp have different date range. Timestamp use the (32 bits) unix epoch range.

johnsaigle:
@PapillonMcGill I don't think we can make assumptions about server configurations generally -- even if everything at the MNI was standardized, other groups may have different settings.

tools/DB_date_zeros_removal.php
```diff
@@ -106,11 +106,22 @@
             " SET ".$db->escape($field['COLUMN_NAME'])."=NULL".$autoUpdateSQL.
             " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00 00:00:00';\n";
     } else {
-	echo "COLUMN ".$field['COLUMN_NAME']." in TABLE ".$field['TABLE_NAME']." is NOT NULLABLE. ".
-	    "A date '1000-01-01' will be entered instead of '0000-00-00' values.\n"; 
-        $nonNullUpdates .= "UPDATE ".$db->escape($database['database']).".".$db->escape($field['TABLE_NAME']).
-            " SET ".$db->escape($field['COLUMN_NAME'])."='1000-01-01'".$autoUpdateSQL.
-            " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00';\n";
+        echo "COLUMN ".$field['COLUMN_NAME']." in TABLE ".$field['TABLE_NAME']." is NOT NULLABLE. ".
+            "A date '1000-01-01' will be entered instead of '0000-00-00' values.\n";
```
jesscall: you should update this message as it's not accurate in the case of timestamp where you're changing it to 1970-01-01

HenriRabalais:
@llevitis Could you update this PR to outline how it should be tested?

tools/DB_date_zeros_removal.php
zaliqarosli: i agree. but what's the reason for the change from '1000-01-01' to '1970-01-01' in the first place?

llevitis: As per @PapillonMcGill's comment: "Datetime and Timestamp have different date range. Timestamp use the (32 bits) unix epoch range"

driusan:
Has this been tested by anyone yet?

tools/DB_date_zeros_removal.php
PapillonMcGill: should specify that the date can be either 1000-01-01 or 1970-01-01

tools/DB_date_zeros_removal.php
```diff
@@ -106,11 +106,25 @@
...
-            " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00';\n";
+        if ($field['DATA_TYPE'] == 'date') {
+            echo "COLUMN ".$field['COLUMN_NAME']." in TABLE ".$field['TABLE_NAME']." is NOT NULLABLE. ".
+                "A date '1000-01-01' will be entered instead of '0000-00-00' values.\n";
+            $nonNullUpdates .= "UPDATE ".$db->escape($database['database']).".".$db->escape($field['TABLE_NAME']).
+                " SET ".$db->escape($field['COLUMN_NAME'])."='1000-01-01'".$autoUpdateSQL.
+                " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00';\n";
+        } else if ($field['DATA_TYPE'] == 'datetime') {
+            echo "COLUMN ".$field['COLUMN_NAME']." in TABLE ".$field['TABLE_NAME']." is NOT NULLABLE. ".
+                "A date '1000-01-01' will be entered instead of '0000-00-00' values.\n";
```
sruthymathew123: @llevitis I think the message 'A date '1000-01-01' will be entered instead of '0000-00-00' values'  should be changed to something like this; 'A datetime '1000-01-01 00:00:00' will be entered instead of '0000-00-00' values.
Otherwise the update statement and the message doesn't match each other.

tools/DB_date_zeros_removal.php
```diff
@@ -106,11 +106,25 @@
...
+            $nonNullUpdates .= "UPDATE ".$db->escape($database['database']).".".$db->escape($field['TABLE_NAME']).
+                " SET ".$db->escape($field['COLUMN_NAME'])."='1000-01-01 00:00:00'".$autoUpdateSQL.
+                " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00 00:00:00';\n";
+        } else if ($field['DATA_TYPE'] == 'timestamp') {
+            echo "COLUMN ".$field['COLUMN_NAME']." in TABLE ".$field['TABLE_NAME']." is NOT NULLABLE. ".
+                "A date '1970-01-01' will be entered instead of '0000-00-00' values.\n";
```
sruthymathew123: @llevitis same here.

tools/DB_date_zeros_removal.php
sruthymathew123: @llevitis 
The new checks are working good; but the print messages for updating the data and the update statements(for date time & timestamp) are not matching each other.

llevitis: done!

tools/DB_date_zeros_removal.php
```diff
@@ -106,11 +106,25 @@
...
-            " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00';\n";
+        if ($field['DATA_TYPE'] == 'date') {
+            echo "COLUMN ".$field['COLUMN_NAME']." in TABLE ".$field['TABLE_NAME']." is NOT NULLABLE. ".
+                "A date '1000-01-01' will be entered instead of '0000-00-00' values.\n";
+            $nonNullUpdates .= "UPDATE ".$db->escape($database['database']).".".$db->escape($field['TABLE_NAME']).
+                " SET ".$db->escape($field['COLUMN_NAME'])."='1000-01-01'".$autoUpdateSQL.
+                " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00';\n";
+        } else if ($field['DATA_TYPE'] == 'datetime') {
+            echo "COLUMN ".$field['COLUMN_NAME']." in TABLE ".$field['TABLE_NAME']." is NOT NULLABLE. ".
+                "A datetime '1000-01-01' will be entered instead of '0000-00-00' values.\n";
```
sruthymathew123: @llevitis Is that datetime should exactly like '1000-01-01 00:00:00' ?

tools/DB_date_zeros_removal.php
```diff
@@ -106,11 +106,25 @@
...
+            $nonNullUpdates .= "UPDATE ".$db->escape($database['database']).".".$db->escape($field['TABLE_NAME']).
+                " SET ".$db->escape($field['COLUMN_NAME'])."='1000-01-01 00:00:00'".$autoUpdateSQL.
+                " WHERE CAST(".$db->escape($field['COLUMN_NAME'])." AS CHAR(20))='0000-00-00 00:00:00';\n";
+        } else if ($field['DATA_TYPE'] == 'timestamp') {
+            echo "COLUMN ".$field['COLUMN_NAME']." in TABLE ".$field['TABLE_NAME']." is NOT NULLABLE. ".
+                "A timestamp '1970-01-01' will be entered instead of '0000-00-00' values.\n";
```
sruthymathew123: @llevitis A timestamp '1970-01-01 00:00:01' will be.....

sruthymathew123:
Working on IBIS database as well.

driusan:
If you click on the yellow dot this has, in fact, passed Travis, the icon on GitHub just isn't updating.. so I'm merging this.

PR #3710, opened 2018-06-11
[Tools] updated script to split mri violations log commas

llevitis:
This PR updates the `normalize_protocol_split_row.php` to allow for the `mri_violations_log` table to be split by commas.

### DIFF
--- a/tools/normalize_protocol_split_rows.php
+++ b/tools/normalize_protocol_split_rows.php
@@ -50,7 +50,7 @@ function split_commas($table_name) {
     foreach ($rows as $row) {
         $num_commas = 0;
         foreach($row as $key => $value) {
-            if ($key == "ID") {
+            if ($key == "ID" || $key == "LogID") {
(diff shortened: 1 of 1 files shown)


tools/normalize_protocol_split_rows.php
```diff
@@ -63,12 +63,19 @@ function split_commas($table_name) {
...
+            }
+            $DB->delete($table_name, array("LogID" => $id));
+        } else {
+            foreach ($all_combinations as $combination) {
+                unset($combination["ID"]);
+                $DB->insert($table_name, $combination);
+            }
+            $DB->delete($table_name, array("ID" => $id));
         }
-        // remove the row for the original ID
```
ZainVirani: probably a good idea to preserve these comments

driusan:
I don't think mri violations log should be retroactively normalized. It's a log, and it wasn't normalized at the time that the entry was made.

PR #3713, opened 2018-06-12
[Create Timepoint] Sanitize CandID in href

johnsaigle:
Redmine #13790

While CandID comes from the database and is expected to be trusted, this PR explicitly sanitizes the candID variable before it is injected into the href element.

### DIFF
--- a/modules/create_timepoint/php/create_timepoint.class.inc
+++ b/modules/create_timepoint/php/create_timepoint.class.inc
@@ -81,11 +81,6 @@ class Create_Timepoint extends \NDB_Form
         );
 
         $this->tpl_data['success'] = true;
-        $this->tpl_data['candID']  = htmlentities(
-            $this->identifier,
-            ENT_QUOTES,
-            'UTF-8'
-        );
 
         // freeze it, just in case
         $this->form->freeze();
@@ -114,7 +109,7 @@ class Create_Timepoint extends \NDB_Form
         );
         $this->tpl_data['candID'] = $sanitizedIdentifier;
 
-        $this->addHidden('candID', $sanitizedIdentifier;
+        $this->addHidden('candID', $sanitizedIdentifier);
 
         $candidate   =& \Candidate::singleton($this->identifier);
         $subprojList = null;
--- a/modules/create_timepoint/php/create_timepoint.class.inc
+++ b/modules/create_timepoint/php/create_timepoint.class.inc
@@ -81,6 +81,11 @@ class Create_Timepoint extends \NDB_Form
         );
 
         $this->tpl_data['success'] = true;
+        $this->tpl_data['candID'] = htmlentities(
+            $this->identifier, 
+            ENT_QUOTES,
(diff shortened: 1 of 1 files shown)


johnsaigle:
@driusan I wanted to get your eyes on this in particular as you created the Redmine ticket.

johnsaigle:
Tagging this as security to match the Redmine ticket.

modules/create_timepoint/php/create_timepoint.class.inc
```diff
@@ -81,6 +81,11 @@ class Create_Timepoint extends \NDB_Form
         );
 
         $this->tpl_data['success'] = true;
+        $this->tpl_data['candID']  = htmlentities(
```
zaliqarosli: why are we doing this twice, once in process() and once in setup() ?

johnsaigle: You're right it only needs to happen in setup. I'll change this in a bit.

johnsaigle: Actually looking at this again it looks like we need to do it twice... In both functions an element is added to the front-end and it needs to be sanitized there. 

From what I know of LORIS it seems like some functionality is duplicated here and I'm not sure that `_process` should be creating HTML elements. It should probably be refactored but that's not the goal of this PR so the sanitization should occur in both place.

zaliqarosli: but i believe that you added the addition of the element to the front-end in process(). like 84 to add the identifier to tpl_data wasn't there before. what was the reason for adding it?

johnsaigle: Excellent point... Sorry I was working on another PR when looking at this and was distracted. I'll remove it.

johnsaigle:
@zaliqarosli I fixed it like you suggested.

HenriRabalais:
@johnsaigle how should this be tested?

johnsaigle:
@HenriRabalais If you look at the Redmine ticket there's some background there. Essentially Dave notes that we are protected "by accident" from XSS here due to some upstream quirk. This PR simply makes us extra sure. The functionality before and after this PR shouldn't change. However, as we restructure LORIS this existing accidental protection might fall away and this will stop us from having a security issue at that time.

In short, as long as this change doesn't break anything consider it to be passing.

HenriRabalais:
Created a timepoint and accessed it. Everything appears to be in order!

driusan:
I don't have any problems with this change, but it doesn't address the Redmine ticket that is referenced.

driusan:
Moving this to minor because nothing's currently broken, so it's not worth risking breaking something in a bugfix release.

johnsaigle:
@driusan From the ticket:

> I think we're safe entirely by accident, but **it still needs to be sanitized** and the JS should be updated.

This was just a quick fix to do sanitization until the module is converted to React.

PR #3717, opened 2018-06-13
[JSX] Added Re-usable React Loader Component

HenriRabalais:
This pull request will create a React Loader component which can be displayed on reactified pages, while content is being loaded to the page.

This is the most basic version of the loader and can be altered in the future to include text, or potentially the loris logo. 

I have included the implementation of the loader for the media module as a use-case.

loader


davidblader:
looks good. i'd recommend deprecating our current animated glyphicons entirely for this.
other modules / jsx files which could benefit from this:
image

HenriRabalais:
@davidblader sounds good. What would the the deprecation entail?

HenriRabalais:
@ZainVirani my commit to fix travis errors dismissed your review for some reason. If you could re-approve it that would be much appreciated!

davidblader:
> @davidblader sounds good. What would the the deprecation entail?

just replacing all the occurrences of [this](https://github.com/aces/Loris/pull/3717/files#diff-451e38c70bb2cc6a39b0928c87617bf6L62) with your component lol

htdocs/bootstrap/css/custom-css.css
```diff
@@ -453,3 +453,28 @@ a.btn.btn-primary {
 .sweet-alert p {
   word-wrap: break-word;
 }
+
+/* Loader customizations */
+
+.loader-container {
+    height: 80vh;
+    width: 100%;
```
driusan: why is one measured in vh but the other in percentage? I think the component would compose better if both were measured in percentages.

HenriRabalais: When expressing the height as a percentage, the loading circle would be positioned immediately below the breadcrumbs, no matter the percentage value. To resolve this it seemed like either needed to express the height as a fixed amount of pixels, or as a relative height of the viewport. I figured the latter would be more consistent across monitors.

loadernowww

driusan: why not add padding to the loader to ensure it doesn't hit up against the top? Measuring in vh means you can only use it for "page loading" and not for some sub-component of a page loading. I'd try adding some amount of padding measured in em (my gut tells me 3em would probably be fine...)

It also seems like you're trying to use flexbox just to center one item, which is overkill. You could use `margin-left: auto; margin-right: auto;` to do that going all the way back to browsers with CSS1 support.

(At that point, you might not even need the container at all, you could just make the loader a block component with centered block element with padding..)

htdocs/bootstrap/css/custom-css.css
```diff
@@ -453,3 +453,28 @@ a.btn.btn-primary {
...
+    align-items: center;
+    justify-content: center;
+}
+
+.loader {
+    border: 8px solid #FFFFFF; /* White */
+    border-top: 8px solid #3498db; /* Blue */
+    border-radius: 50%;
+    width: 120px;
+    height: 120px;
```
driusan: Think about what you'd have to do to get rid of the container for my comment above makes me think that maybe this should be something like:

```
width: 100%;
max-width: 120px;
```

But then you'd need some way to ensure height matches width or you'd get a weird oblong loader..

htdocs/bootstrap/css/custom-css.css
HenriRabalais: Sounds good! I'll give that a shot tomorrow.

HenriRabalais:
@driusan it seems that you can include css animations by using a facebook-provided add-on: https://reactjs.org/docs/animation.html

HenriRabalais:
I've altered the loader to make it more generalizable to sub-components. However, by removing the flexbox ,it is no long centered vertically, only horizontally. However, another container div can simply be added to wrap the `Loader` component where it is being implemented to force it to be centered vertically. There is now a prop option for setting the size of the spinning loader.

HenriRabalais:
With the latest commit, the Loader will be centred both vertically and horizontally with its parent div.

@driusan does this work? Or should I make some more adjustments?

driusan:
@HenriRabalais I don't know.. does it work? The code looks fine to me.

HenriRabalais:
@driusan it does work! I should clarify my question: does this have the functionality you're hoping to have out of a loader component? If so, I think we're good to go! 

@ZainVirani coudl you re-review/test?

HenriRabalais:
@ZainVirani @davidblader @zaliqarosli @driusan 
Please re-review!

PR #3719, opened 2018-06-14
Fix loading of datadict module

driusan:
Rename misnamed function. Probably the victim of a poorly resolved merge conflict.
See Redmine: #14714


driusan:
Fixed by e657f8e1bd8def861e8d338009fcde92e2246d50 (which I accidentally directly committed, so this is no longer necessary..)

PR #3720, opened 2018-06-14
Fix redirect on user accounts page

driusan:
This fixes the redirect to the user's page after
creating a new user.

It also fixes the identifier to be part of the URL,
rather than a get parameter and fixes the 6000 PHP warnings
you get when creating a new user.

See also: Redmine #14715


ZainVirani:
Didn't work 😞 @driusan

driusan:
Travis is failing due to phan errors unrelated to this PR, will fix them with another PR.

PR #3723, opened 2018-06-14
Remove reference to a function which doesn't exist in NDB_Page

driusan:
This removes the (only) reference to an undefined "PermissionDenied"
exception, and instead replaces it with a 403 response with the
words "Permission Denied" in the body.

See also: Redmine #14722

### DIFF
--- a/php/libraries/NDB_Page.class.inc
+++ b/php/libraries/NDB_Page.class.inc
@@ -620,7 +620,9 @@ class NDB_Page implements RequestHandlerInterface
     ) : ResponseInterface {
         $user = $request->getAttribute("user");
         if ($this->_hasAccess() !== true) {
-            throw new PermissionDenied();
+            return (new \LORIS\Middleware\PageDecorationMiddleware($user))
+                ->process($request, new \LORIS\Http\StringStream("Permission Denied"))
+                ->withStatus(403);
         }
 
         // The NDB_Page object itself is sometimes required by middleware,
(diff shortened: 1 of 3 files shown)


src/Http/StringStream.php
```diff
@@ -266,4 +270,20 @@ public function getMetadata($key=null)
         }
         return $metadata[$key];
     }
+
+    /**
+     * A StringStream can act as a handler, which resolves to a response with
+     * itself as a body.
+     *
+     * This is primarily to allow it to be used content by middleware.
```
johnsaigle: Don't need the word `content` here -- or the sentence needs to be reworked.

php/libraries/NDB_Page.class.inc
```diff
@@ -620,7 +620,11 @@ class NDB_Page implements RequestHandlerInterface
     ) : ResponseInterface {
         $user = $request->getAttribute("user");
         if ($this->_hasAccess() !== true) {
-            throw new PermissionDenied();
+            return (new \LORIS\Middleware\PageDecorationMiddleware($user))
+                ->process(
+                    $request,
+                    new \LORIS\Http\StringStream("Permission Denied")
```
cmadjar: Feels like the error message could be more indicative. Maybe something like:
"You do not have permission to view candidates that do not belong to your site" or something like that?

driusan: But this middleware is used for every page, regardless of the reason.

src/Http/StringStream.php
driusan: I think it was just missing the word "as".. is it better now?

PR #3724, opened 2018-06-14
[timepoint_list] Remove invalid buttons from timepoint_list page

driusan:
This removes the buttons that the test plan says are invalid on the timepoint_list page

See also: Redmine #14723

### DIFF
--- a/modules/timepoint_list/php/timepoint_list_controlpanel.class.inc
+++ b/modules/timepoint_list/php/timepoint_list_controlpanel.class.inc
@@ -64,13 +64,12 @@ Class TimePoint_List_ControlPanel extends \Candidate
         $this->tpl_data['candID'] = $this->getData('CandID');
 
         $this->tpl_data['isDataEntryPerson']
-            = $user->hasCenterPermission("data_entry", $cand_CenterID)
-            || $user->isUserDCC();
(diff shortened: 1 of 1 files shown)


modules/timepoint_list/php/timepoint_list_controlpanel.class.inc
```diff
@@ -64,13 +64,12 @@ Class TimePoint_List_ControlPanel extends \Candidate
         $this->tpl_data['candID'] = $this->getData('CandID');
 
         $this->tpl_data['isDataEntryPerson']
-            = $user->hasCenterPermission("data_entry", $cand_CenterID)
-            || $user->isUserDCC();
+            = $user->hasCenterPermission("data_entry", $cand_CenterID);
```
nicolasbrossard: Code looks good. Tagging as Passed Code Review. I don't see the relationship between the code modifications and the PR title though.

driusan: It was checking if the user has permission at all, but the test plan says it should only display it for candidates at the same site, so it needed to be changed to be center permission. The code in the javascript/template was already there, the bool was just flagged based on the wrong thing.

PR #3728, opened 2018-06-15
Imaging Browser 20.0 bugfix: Fix links to instruments

MounaSafiHarab:
The link to the instruments was following the 
baseurl/instrument_name/?commentID&sessonid&candid format
when it should now be:
baseurl/instruments/instrument_name/?commentID&sessonid&candid

I also added one comment in the README about the DICOM download option, identical to the one in the DICOM Archive module, in case of large .tar files.


MounaSafiHarab:
OK done.

I thought the readme changes were so straightforward that it was so much overhead for another PR.
Anyway, removed from here and created it in a new one here #3732

PR #3730, opened 2018-06-15
[conflict_resolver][minor] Upgraded module to use react & Modified jsx/Form.js ButtonElement

intralizee:
This pull request `removes smarty and upgrades module to use react`. 
It `removes the smarty templates and adds two new react files for conflicts unresolved & resolved`.

Also I added the ability for react ButtonElement to have the attributes "id & name" and be optional when creating a button. Reasoning is since a lot of tests have been written to use the WebDriverBy::ID('example') or by Name and for the goal to test clicking a button.

See (Feature #14841) on Redmine


modules/conflict_resolver/jsx/react.conflict_resolver.js
```diff
@@ -0,0 +1,262 @@
...
+ *	This file contains the React classes for conflict resolver
+ * 	module.
+ */
+
+/**
+ * This function is responsible for updating
+ * the UnresolvedConflictsPane state from the parent.
+ * @param filter
+ */
+function updateState(filter){
```
ZainVirani: why is this defined outside of the class?

modules/conflict_resolver/php/conflict_resolver.class.inc
```diff
@@ -340,49 +310,57 @@ class Conflict_Resolver extends \NDB_Menu_Filter_Form
...
         );
-        $this->addBasicText('Question', 'Search keyword:');
+        $this->addBasicText('question', 'Question:');
 
         // Project list, if applicable
-        if ($config->getSetting("useProjects")==="true") {
+        if ($config->getSetting('useProjects')==='true') {
             $list_of_projects = array(null => 'All');
             $projectList      = \Utility::getProjectList();
             foreach ($projectList as $key => $value) {
```
victoriafoing: In order for the Project filter to work, you need to make sure the key for is the same as the value in the $list_of_projects array.

modules/conflict_resolver/php/conflict_resolver.class.inc
```diff
@@ -340,49 +310,57 @@ class Conflict_Resolver extends \NDB_Menu_Filter_Form
...
                 $list_of_projects[$key] = $value;
             }
             $this->addSelect('Project', 'Project: ', $list_of_projects);
```
victoriafoing: $this->addSelect('project', 'Project: ', $list_of_projects);

make sure 'project' for $name is lower case

modules/conflict_resolver/php/resolved_conflicts.class.inc
```diff
@@ -197,40 +159,48 @@ class Resolved_Conflicts extends \NDB_Menu_Filter
...
-        $this->addBasicText('ResolutionTimestamp', 'Resolution Timestamp:');
+        $this->addBasicText('question', 'Question:');
+        $this->addBasicText('timestamp', 'Resolution Timestamp:');
 
         // Project list, if applicable
-        if ($config->getSetting("useProjects")==="true") {
+        if ($config->getSetting('useProjects')==='true') {
             $list_of_projects = array(null => 'All');
             $projectList      = \Utility::getProjectList();
             foreach ($projectList as $key => $value) {
```
victoriafoing: I can't comment on the necessary lines but make the same changes I requested in Unresolved Conflicts:

- Make sure the key is the same as the value in $list_of_projects (line 237)
$list_of_projects[$value] = $value;

- Make sure 'project' for $name is lower case (line 239)
$this->addSelect('project', 'Project: ', $list_of_projects);

modules/conflict_resolver/php/resolved_conflicts.class.inc
```diff
@@ -197,40 +159,48 @@ class Resolved_Conflicts extends \NDB_Menu_Filter
...
-             "size"      => 9,
-             "maxlength" => 7,
+             'size'      => 9,
+             'maxlength' => 7,
             )
         );
-        $this->addBasicText('Question', 'Search keyword:');
-        $this->addBasicText('ResolutionTimestamp', 'Resolution Timestamp:');
+        $this->addBasicText('question', 'Question:');
+        $this->addBasicText('timestamp', 'Resolution Timestamp:');
```
victoriafoing: Change to the following to fix Resolution Timestamp filter:
$this->addBasicText('resolutionTimestamp', 'Resolution Timestamp:');

johnsaigle:
JSLint seems to have a lot of problems with this PR and so it is failing Travis. 

Please run `npm run lint:javascript` locally and fix the errors reported.

driusan:
@intralizee conflicts

driusan:
(I mean the PR itself has conflicts.)

intralizee:
@driusan fixed conflicts.

modules/conflict_resolver/jsx/react.conflict_resolver.js
```diff
@@ -0,0 +1,259 @@
+import {Tabs, TabPane} from 'Tabs';
+import FilterForm from 'jsx/FilterForm';
+import formatColumn from './columnFormatter';
```
HenriRabalais: I believe that it was discussed in our reactification workshops/meetings that `formatColumn` should be a function within the index component for the filter/datatable page.

modules/conflict_resolver/jsx/react.conflict_resolver.js
```diff
@@ -0,0 +1,259 @@
...
+      Data: this.props.data
+    };
+    updateState = updateState.bind(this);
+  }
+
+  // Render the HTML
+  render() {
+    return (
+      <TabPane Title="" {...this.props}>
+        <form method="post" action={this.props.url.base + "/conflict_resolver/"} name="conflict_resolver" id="conflict_resolver">
```
HenriRabalais: These attributes could be split onto multiple lines for visual clarity.

modules/conflict_resolver/jsx/react.conflict_resolver.js
```diff
@@ -0,0 +1,259 @@
...
+      }.bind(this),
+      error: function(error) {
+        console.error(error);
+      }
+    });
+  }
+
+  updateFilter(filter) {
+    this.setState({filter});
+    updateState(filter);
```
HenriRabalais: Is this not redundant since what `updateState(filter)` does is update the state of the filter in the same manner that line `128` does?

modules/conflict_resolver/jsx/react.conflict_resolver.js
```diff
@@ -0,0 +1,259 @@
...
+    tabs.push(
+      <UnresolvedConflictsPane
+        TabId="UnresolvedConflicts"
+        ref="unresolvedConflictsPane"
+        key={1}
+        url={this.props.url}
+        data={this.state.Data}
+      />
+    );
+    var tabList = [
```
HenriRabalais: Can this be changed to `let`?

modules/conflict_resolver/jsx/react.resolved_conflicts.js
```diff
@@ -0,0 +1,250 @@
...
+ *	This is the React class for building the instrument
+ */
+class ResolvedConflictsPane extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      Data: this.props.data
+    };
+    updateState = updateState.bind(this);
```
HenriRabalais: Why is `updateState` being bound if the function is not used in this component?

jsx/Form.js
```diff
@@ -22,14 +21,41 @@ ButtonElement, LorisElement
  * Note that if both are passed `this.props.formElements` is displayed first.
  *
  */
-class FormElement extends React.Component {
-  constructor() {
-    super();
-    this.getFormElements = this.getFormElements.bind(this);
-    this.handleSubmit = this.handleSubmit.bind(this);
-  }
+var FormElement = React.createClass({
```
davidblader: it looks like you were based on an old branch, the Form components have changed a lot since then
adding Needs Rebase

modules/conflict_resolver/jsx/react.conflict_resolver.js
intralizee: @HenriRabalais The function "updateState" is being used to pass the filtered state between the two ConflictResolverApp and the UnresolvedConflictsPane components.

modules/conflict_resolver/jsx/react.resolved_conflicts.js
intralizee: @HenriRabalais See previous comment similar to this.

modules/conflict_resolver/jsx/columnFormatter.js
```diff
@@ -4,17 +4,17 @@
...
+ * @param {array} rowHeaders - array of table headers (column names)
+ * @return {*} a formatted table cell for a given column
  */
 function formatColumn(column, cell, rowData, rowHeaders) {
   if (loris.hiddenHeaders.indexOf(column) > -1) {
     return null;
   }
-
   // Create the mapping between rowHeaders and rowData in a row object.
   var row = {};
```
HenriRabalais: This should be set to `let`

HenriRabalais:
Just a little bug when testing:
- When selecting `All` for any of the filter dropdowns, the value `All` does not display in the filter.

modules/conflict_resolver/jsx/react.conflict_resolver.js
HenriRabalais: Hmm, I would normally say that this should be done through a call-back function from the parent component. But it seems in this case you don't have a parent `index` component. Ideally, both Filters should have a shared parent component, but perhaps that can be next reactification step when we lift the state of every module.

@davidblader @xlecours @ZainVirani @driusan  thoughts?

HenriRabalais: Shouldn't the filter be passed as a prop to `UnresolvedConflictsPane` after linue `158`?

intralizee:
@HenriRabalais Defaulting all the empty options to All is a great proposal. I'll add that change.

jsx/FilterForm.js
```diff
@@ -175,7 +175,6 @@ class FilterForm extends React.Component {
         formElements[fieldName].value = this.queryString[queryFieldName];
       }.bind(this));
     }
-
```
zaliqarosli: this change could be removed from the PR

jsx/Form.js
```diff
@@ -1284,6 +1284,7 @@ LinkElement.defaultProps = {
  * Button component
  * React wrapper for <button> element, typically used to submit forms
  */
+
```
zaliqarosli: this trivial change can also be removed

modules/conflict_resolver/jsx/react.conflict_resolver.js
```diff
@@ -0,0 +1,275 @@
...
+ */
+
+/**
+ * @doc function
+ * @name updateState
+ * @param {object} filter to update the state.
+ * @description This function is responsible for updating
+ * the UnresolvedConflictsPane state from the parent.
+ */
+var updateState = function(filter) {
```
zaliqarosli: use new syntax for defining functions?

zaliqarosli: and how come its defined outside the class?

modules/conflict_resolver/jsx/react.conflict_resolver.js
zaliqarosli: i agree with henri. we should be passing states through props

modules/conflict_resolver/php/conflict_resolver.class.inc
```diff
@@ -26,6 +26,7 @@ namespace LORIS\conflict_resolver;
  */
 class Conflict_Resolver extends \NDB_Menu_Filter_Form
 {
+    public $skipTemplate = true;
```
zaliqarosli: we can remove this line if the smarty template is deleted

modules/conflict_resolver/php/conflict_resolver.class.inc
```diff
@@ -78,11 +79,11 @@ class Conflict_Resolver extends \NDB_Menu_Filter_Form
                 if (!empty($row)) {
                     // insert into conflicts_resolved
                     $user1         = $DB->pselectOne(
-                        "SELECT UserID FROM flag WHERE CommentID=:CID",
+                        'SELECT UserID FROM flag WHERE CommentID=:CID',
```
zaliqarosli: double quotes here is fine

modules/conflict_resolver/php/conflict_resolver.class.inc
```diff
@@ -183,11 +183,14 @@ class Conflict_Resolver extends \NDB_Menu_Filter_Form
...
+        // Rest the filter because this is broken with react modules
+        //$this->_resetFilters();
+
         $config =& \NDB_Config::singleton();
 
         // set the class variables
-        $useProjects = $config->getSetting("useProjects");
-        if ($useProjects === "false") {
+        $useProjects = $config->getSetting('useProjects');
+        if ($useProjects === 'false') {
```
zaliqarosli: are we still using tpl_data if there's no smarty template?

modules/conflict_resolver/php/conflict_resolver.class.inc
intralizee: I get this error message below if I remove the $skipTemplate:
The following errors occurred while attempting to display this page:
Unable to load template 'file:menu_conflict_resolver.tpl'
If this error persists, please report a bug to your administrator .

modules/conflict_resolver/php/conflict_resolver.class.inc
intralizee: You're right, I removed the tpl_data with latest commit.

modules/conflict_resolver/jsx/react.conflict_resolver.js
intralizee: @HenriRabalais Please explain more, it currently it works and I'm not understanding the benefit.

modules/conflict_resolver/php/conflict_resolver.class.inc
intralizee: @zaliqarosli for your comment "i agree with henri. we should be passing states through props" here is the reason. I did it with an outside function because of the two components and from what I've read online you can only set the props once. I'm passing data between components and to update the state of another component.

modules/conflict_resolver/jsx/react.conflict_resolver.js
HenriRabalais: Simply put, `columnFormatter.js` is a file that contains a function that could be placed in this class instead of in it's own file. It's purely a matter of deprecated organisation which we are trying correct with new reactification PRs.

modules/conflict_resolver/jsx/react.conflict_resolver.js
HenriRabalais: When I made that comment three weeks ago, I tested the PR and managed to get the filters to work by passing the filter as a `prop`. I suggest going that route since it's best for us to be passing `props` down from a top level stateful controller.

@davidblader @driusan @davidblader perhaps you have some input on this?

jsx/Form.js
```diff
@@ -1299,6 +1299,8 @@ class ButtonElement extends React.Component {
       <div className="row form-group">
         <div className={this.props.columnSize}>
           <button
+            {...(this.props.id ? {id: this.props.id} : {})}
+            {...(this.props.name ? {name: this.props.name} : {})}
```
HenriRabalais: I'm unsure why the spread operator would be necessary here.

intralizee: @HenriRabalais The spread operator is handling both the key and the property (ex. id='btnExample). The id key or name key may not have values and so they shouldn't be included in rendering or it would be id='' & name='' inside the button component. I believe it's necessary, thoughts?

HenriRabalais: From what I can see, the ternary operator is accomplishing that goal by checking the values exist, and then setting the the key-value pair if they do, and an empty object otherwise. Where does the spread operator come into play here? The spread operator is reserved for iterables, such as an array. It doesn't seem like either of these ternary operations result in an expression that needs to iterated over.

modules/conflict_resolver/jsx/react.conflict_resolver.js
davidblader: not sure i understand @intralizee 's explanation

in the browser for a given tab, i see two components, `ConflictResolverApp` and either `UnresolvedConflictsPane` or `ResolvedConflictsPane`. updating the filter causes a `filter` object to be duplicated in the states of both `ConflictResolverApp` and `UnresolvedConflictsPane` / `ResolvedConflictsPane`

duplicating state like this is generally a bad idea, i would recommend following @HenriRabalais 's feedback

intralizee:
See PR #3907

PR #3733, opened 2018-06-18
[Candidate Parameters] 20.0 fix "Return to timepoint list" button

davidblader:
replace deprecated URL with updated timepoint list URL


driusan:
Merged because the Travis failure is due to #3735, not this PR and it was manually tested.

PR #3736, opened 2018-06-18
[Candidate Parameters, Participant Status] 20.0 -- fix NULL form entry saves

davidblader:
saving participant status without providing comments results in `null` being saved as a string
added checks around appends to FormData object to filter out nulls & undefineds


HenriRabalais:
Should this be done across all forms?

modules/candidate_parameters/jsx/ParticipantStatus.js
```diff
@@ -226,7 +226,11 @@ var ParticipantStatus = React.createClass(
       var self = this;
       var formData = new FormData();
       for (var key in myFormData) {
-        if (myFormData[key] !== "") {
+        if (myFormData.hasOwnProperty(key) &&
+          myFormData[key] !== "" &&
+          myFormData[key] !== null &&
+          myFormData[key] !== undefined
```
HenriRabalais: It may be better to simply use `!=` here due to javascript coercion. http://webreflection.blogspot.com/2010/10/javascript-coercion-demystified.html

I will look into this more though to confirm.

HenriRabalais: After reading through it appears that `myFormData[key] != null` would in fact be a more concise check.

ZainVirani:
@HenriRabalais 
> Should this be done across all forms?

Ideally, yes. I'm assuming that most forms already do something like this.

modules/candidate_parameters/jsx/ParticipantStatus.js
ZainVirani: I don't think myFormData will ever hold an object though, it would only hold values from the form, no?

davidblader: > After reading through it appears that myFormData[key] != null would in fact be a more concise check.
i would rather be explicit. i generally do not like using loose comparisons unless i have a good reason for doing so.

> I don't think myFormData will ever hold an object though, it would only hold values from the form, no?
i don't think i know what you're talking about here

ZainVirani: I don't think I know what I'm talking about either.

driusan:
This needs to be rebased

PR #3738, opened 2018-06-18
[install] can't load the installdb.php page  - fix typo

kongtiaowang:
This pull request fixes a typo, "," to " ."

PR #3739, opened 2018-06-18
[Installer] Fix error message typo

johnsaigle:
Same as #3738 but for the minor branch.


driusan:
This isn't necessary, it'll be pulled forward from 20.0 into the minor branch when it's released (which will be before there's any new releases based on the current minor branch..)

PR #3432, opened 2018-01-19
[Data Dictionary] Reactify module + searchKeyword functionality added to StaticDataTable

llevitis:
This PR reactifies the data dictionary module and adds searchKeyword functionality to StaticDataTable. The changes to StaticDataTable assume that the filter name of the keyword element will be 'keyword'. The multiselect functionality of the 'Instruments' relies on PR #3406.

PR #3741, opened 2018-06-18
Fix clear filters button of imaging uploader

cmadjar:
This pull request fixes the "Clear Filters" button that did not work on the imaging uploader (it was not mapped to any action).

See also: https://redmine.cbrain.mcgill.ca/issues/14721

Note: on the Redmine ticket, it is also mentioned than when clearing for example only CandID filter, row with empty CandID are not shown. This is fixed in a different PR: #3743


modules/imaging_uploader/jsx/ImagingUploader.js
```diff
@@ -75,13 +80,14 @@ class ImagingUploader extends React.Component {
                 Module="imaging_uploader"
                 name="imaging_filter"
                 id="imaging_filter"
+                ref="imagingUploaderFilter"
```
davidblader: string refs are being deprecated. use a callback function similar to [what was done here for data dictionary](https://github.com/aces/Loris/pull/3432/files#diff-b7cfdf269c10c4ab50f7710f6cfdaee7R81)

modules/imaging_uploader/jsx/ImagingUploader.js
```diff
@@ -49,6 +50,10 @@ class ImagingUploader extends React.Component {
     this.setState({filter});
   }
 
+  resetFilters() {
+    this.refs.imagingUploaderFilter.clearFilter();
```
davidblader: if using the ref callback method used below, you should remove the `.refs.` here

cmadjar:
@davidblader Thank you! This is fixed.

driusan:
Needs rebase.

cmadjar:
@driusan PR has been rebased. Thanks!

PR #3744, opened 2018-06-19
[Examiners] Add Case Sensitive for the table.

kongtiaowang:
This pull request Add Case Sensitive for the table.
https://redmine.cbrain.mcgill.ca/issues/14728

### DIFF
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -1157,7 +1157,7 @@ CREATE TABLE `family` (
 
 CREATE TABLE `examiners` (
   `examinerID` int(10) unsigned NOT NULL auto_increment,
-  `full_name` varchar(255) default NULL,
+  `full_name` varchar(255) CHARACTER SET utf8 COLLATE utf8_bin default NULL,
   `radiologist` tinyint(1) default 0 NOT NULL,
   `userID` int(10) unsigned DEFAULT NULL,
   PRIMARY KEY  (`examinerID`),
--- a/SQL/Archive/20.0/2018-06-19-ExaminersAddCaseSensitive.sql
+++ b/SQL/Archive/20.0/2018-06-19-ExaminersAddCaseSensitive.sql
@@ -1,0 +1,1 @@
+ALTER TABLE examiners MODIFY full_name varchar(255) CHARACTER SET utf8 COLLATE utf8_bin default NULL;
--- a/SQL/Release_patches/19.1_To_20.0_upgrade.sql
+++ b/SQL/Release_patches/19.1_To_20.0_upgrade.sql
@@ -1,0 +1,1 @@
+ALTER TABLE examiners MODIFY full_name varchar(255) CHARACTER SET utf8 COLLATE utf8_bin default NULL;


ZainVirani:
I'll test this in like 30 minutes

driusan:
I don't understand why this one column in one table needs to be case sensitive if that's not a standard elsewhere in LORIS

ZainVirani:
@driusan The issue is that this table is based on user Full Names instead of unique userIDs. So if I create 3 users with full names "Dave M" "dave m" and "DAVE M", and all of them are DCC examiners, only 1 entry will show up in this table ("dave m"). This PR only fixes the case sensitivity, but ideally this table would be based on userID (however this has its own issue, the add examiner section on the top right would need adjusting as well).

SQL/0000-00-00-schema.sql
```diff
@@ -1157,7 +1157,7 @@ CREATE TABLE `family` (
 
 CREATE TABLE `examiners` (
   `examinerID` int(10) unsigned NOT NULL auto_increment,
-  `full_name` varchar(255) default NULL,
+  `full_name` varchar(255) CHARACTER SET utf8 COLLATE utf8_bin default NULL,
```
PapillonMcGill: char set should be utf8mb4

driusan: If case sensitivity is the issue, I think utf8mb4 is case insensitive.

driusan:
@ZainVirani But isn't it a good thing that all of those would show up as one user? It would likely get confusing if "Dave M" and "dave m" were different users.

SQL/0000-00-00-schema.sql
PapillonMcGill: Some case sensitive collation now exist but maybe not for the version we use and maybe not with the proper ordering.

driusan:
Ping.

Still not seeing why this is required on a one column basis or why we would want to make names, in particular, case sensitive.

ZainVirani:
>But isn't it a good thing that all of those would show up as one user? It would likely get confusing if "Dave M" and "dave m" were different users.

What if they are different users? Keep in mind that we're talking about user full names here, not user names

driusan:
@ZainVirani I think assuming names are case sensitive and treating them as separate users opens a can of worms with regards to security, where someone can easily impersonate someone else by creating a user of the same name with a different case

ZainVirani:
@driusan that's exactly what the problem is now... again, if you as DaveM were to be listed as an examiner for site A, another user "davem" was an examiner at site B, davem would now be listed as an examiner at site A and B

ridz1208:
I see your point @ZainVirani but I agree with @driusan.

you are not mitigating anything by making it case sensitive you are just adding complications usually. the code will need to check for casing.. then on the front end you can have a Dave M, DAVE m, DaVe m, dave M....

yes it is a problem that we are reliying solely on full name but the issue is not resolved by looking at the casing of the name but rather by appending an ID that makes each entry unique. It's a valid ticket for an improvement not a bug, its like this by design (or lack thereof) but this PR is not a solution

zaliqarosli:
@ridz1208 @driusan are we saying we should close this PR? @ZainVirani

PR #3745, opened 2018-06-19
[Examiner] (20.0 Bugfix) Grant permission when user has examiner_multisite or examiner_view

johnsaigle:
This pull request addresses Redmine #14719:

>>> 

    The Examiner module README says:

    The Examiner module uses the following permissions. Any one of them
    is sufficient to have access to the module.

    examiner_view
    ...
    examiner_multisite

    However when a user has just one of either view or multisite, they cannot access the module.

User's can now access the Examiner module with either of these permissions and so it is in harmony with the Readme.

### To test

* On 20.0-release, try giving yourself only the `examiner_view` permission without multisite. You should not be able to view the module.

* On this branch, you should be able to view the module with only the `examiner_view` permission

### DIFF
--- a/modules/examiner/php/editexaminer.class.inc
+++ b/modules/examiner/php/editexaminer.class.inc
@@ -62,7 +62,7 @@ class EditExaminer extends \NDB_Form
                 $centerIDs[] = $row['centerID'];
             }
 
-            // Access is only granted if user sites are the same as examiner sites
+            // Access is only granted if user sites are a superset of examiner sites
             // considering that certification changes affect all sites
             $permitted = $user->hasPermission('examiner_view')
                 || $user->hasPermission('examiner_multisite');
--- a/modules/examiner/php/editexaminer.class.inc
+++ b/modules/examiner/php/editexaminer.class.inc
@@ -2,7 +2,7 @@
 /**
  * Examiner module: Form used to update examiner certification status
  *
- * PHP Version 5
+ * PHP Version 7
  *
  * @category Behavioural
  * @package  Loris
@@ -15,7 +15,7 @@ namespace LORIS\examiner;
 /**
  * Examiner module: Form used to update examiner certification status
  *
- * PHP Version 5
+ * PHP Version 7
  *
  * @category Behavioural
  * @package  Loris
--- a/modules/examiner/php/editexaminer.class.inc
+++ b/modules/examiner/php/editexaminer.class.inc
@@ -66,7 +66,7 @@ class EditExaminer extends \NDB_Form
             // considering that certification changes affect all sites
             $permitted   = $user->hasPermission('examiner_view')
                 || $user->hasPermission('examiner_multisite');
-            $sameCenters // lol phpcs
+            $sameCenters 
                 = empty(array_diff($centerIDs, $user->getData('CenterIDs')));
             return $permitted && $sameCenters;
         }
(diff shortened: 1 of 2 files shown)


modules/examiner/php/editexaminer.class.inc
```diff
@@ -70,8 +70,8 @@ class EditExaminer extends \NDB_Form
             // Access is only granted if user sites are a superset of examiner sites
             // considering that certification changes affect all sites
             return $user->hasPermission('examiner_view')
-                && (($user->hasPermission('examiner_multisite')
-                || empty(array_diff($centerIDs, $user->getData('CenterIDs')))));
+                || $user->hasPermission('examiner_multisite')
```
driusan: @MounaSafiHarab Is there any reason to have this permission at all now that users can be affiliated with multiple sites and not just use `$user->hasCenterPermission('examiner_view')`?

(That change is probably beyond the scope of 20.0.. but just in general?)

ZainVirani: I think this should be `(examiner_view || examiner_multisite) && empty(array_diff(...))` (ie require one of view or multisite), because if not this would return true even if the user didn't have examiner_view or multisite, as long as they belonged to all of the centers (unless this is what is intended).

johnsaigle: Yeah that makes sense. I'll change this line accordingly for now and @MounaSafiHarab can let us know if it's still necessary

modules/examiner/php/editexaminer.class.inc
```diff
@@ -69,9 +64,11 @@ class EditExaminer extends \NDB_Form
 
             // Access is only granted if user sites are a superset of examiner sites
             // considering that certification changes affect all sites
-            return $user->hasPermission('examiner_view')
-                && (($user->hasPermission('examiner_multisite')
-                || empty(array_diff($centerIDs, $user->getData('CenterIDs')))));
+            $permitted   = $user->hasPermission('examiner_view')
+                || $user->hasPermission('examiner_multisite');
+            $sameCenters // lol phpcs
```
driusan: please remove this comment

johnsaigle:
@driusan I have removed the comment. Please re-review when you have time.

modules/examiner/php/editexaminer.class.inc
MounaSafiHarab: @driusan @johnsaigle 
I am not sure if examiner multisite and user multisite is supposed to be identical?
is an examiner always a user?
I really don't know.

@ridz1208 you made this initial change, can you please comment?

ridz1208: @driusan @johnsaigle 

I personally feel that any `_multisite` permission is now antiquated (including `access_all_profiles`) which I brought up during the meeting after @MounaSafiHarab 's PR got merged... everyone seemed to disagree saying it's not the same as being affiliated with all the sites....

So I dont have an answer here

johnsaigle: @ridz1208 @MounaSafiHarab Thanks for weighing in here.

It sounds like we don't have a good reason to change this setting at this time so I feel like we shouldn't adjust it until we are sure that `_multisite` is deprecated.

johnsaigle:
@driusan Thanks for pointing that out. Fixed now.

PR #3746, opened 2018-06-19
[Instrument Manager] (20.0) Fix bug where LorisException should have been thrown when quatUser was not properly configured, improve error checking

johnsaigle:
## Background

This PR addresses two related issues in Redmine. Better documentation and refactoring have been done generally to clarify how instrument manager works.

### 14760

Correctly check if automatic instrument installation is possible by using `empty` rather than `isset`. Previously this condition would be true even with blank credentials in the config file.

### 14751

Dynamic feedback has been integrated to inform users of various differing server (mis)configurations.

Previously user's were informed when uploading had been "disabled by an administrator" (ie. vital project/ folders were not writable). In addition, users are now told that automatic installation is "disabled" when quatUser is not properly configured. 

More importantly, when an upload fails for any number of reasons, users are informed as to why and how to fix the problem. In cases where a back-end administrator is needed to remedy an issue, additional information will be entered in the error log.

## To Test

**On your branch:**

* Set quatUser or quatPassword to be blank values. Try installing an instrument. Nothing should happen (this is the bug).

**On my branch:**

* Go to instrument manager. With blank quat* values, a warning will be displayed. If you upload an instrument, more help will be displayed.


modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -44,76 +44,144 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
+            exec(
+                "mysql".
+                " -h" . escapeshellarg($db_config['host']).
+                " -u" . escapeshellarg($db_config['quatUser']).
+                " -p" . escapeshellarg($db_config['quatPassword']).
+                " " . escapeshellarg($db_config['database']).
+                " < " . $this->path . "project/tables_sql/".
+                escapeshellarg($instrument->table . '.sql'),
+                $output, // $output and $status are created automatically
+                $status  // by `exea`c and so need not be declared above.
```
PapillonMcGill: Look like there is a typo in the comment

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -44,76 +44,144 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
+            if ($status) {
+                $this->tpl_data['feedback'] = "The installation of "
+                    . "$instrument->table.sql failed. "
+                    . "Please contact your administrator.";
+                error_log(
+                    "The installation of $instrument->table.sql failed. "
+                    . "Either: the instrument table exists (but is not in the "
+                    . "test_names table), or "
+                    . "LORIS could not connect to the database using the "
+                    . "credentials supplied in project/config.xml."
```
PapillonMcGill: Could add the content of $output to the log.

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -386,4 +454,22 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
+     */
+    function isQuatUserConfigured() : bool
+    {
+        $db        = \Database::singleton();
+        $config    = \NDB_Config::singleton();
+        $db_config = $config->getSetting('database');
+
+        return !empty($db_config['quatUser'])
+            && !empty($db_config['quatPassword']);
+    }
```
PapillonMcGill: Should we ensure that the user/password combination is valid by actually opening a connection to the DB?

johnsaigle:
@PapillonMcGill I made the changes you requested. In order to check the DB connection I added a return variable to the DB class. It will now return true if a connection is made and false otherwise. This is the only was I could figure out to create a DB connection as LORIS would simply crash with invalid credentials previously (and so I could not test the validity of the quatUser settings).

This is a bit out of scope for the PR, though it shouldn't affect functionality for other parts of LORIS.

Please re-review when you have the time!

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -44,76 +44,145 @@ class Instrument_Manager extends \NDB_Menu_Filter
     {
         $config        = \NDB_Config::singleton();
         $this->path    = $config->getSetting("base");
-        $this->headers = array(
+        $this->headers = [
```
driusan: don't mix cosmetic changes with functional changes

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -44,76 +44,145 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
+                " -h" . escapeshellarg($db_config['host']).
+                " -u" . escapeshellarg($db_config['quatUser']).
+                " -p" . escapeshellarg($db_config['quatPassword']).
+                " " . escapeshellarg($db_config['database']).
+                " < " . $this->path . "project/tables_sql/".
+                escapeshellarg($instrument->table . '.sql'),
+                $output, // $output and $status are created automatically
+                $status  // by `exec` and so need not be declared above.
+            );
+            // In MSQL, an exit code of 0 is a success and 1 means failure
```
driusan: mysql.

And that's not a MySQL thing, that's a unix standard across all tools.

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -44,76 +44,145 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
+                $this->tpl_data['feedback'] = "The installation of "
+                    . "$instrument->table.sql failed. "
+                    . "Please contact your administrator.";
+                error_log(
+                    "The installation of $instrument->table.sql failed. "
+                    . "Either: the instrument table exists (but is not in the "
+                    . "test_names table), or "
+                    . "LORIS could not connect to the database using the "
+                    . "credentials supplied in project/config.xml. "
+                    . print_r($object)
```
driusan: there are many places that the config file can be other than the project directory (in fact, it can be anywhere since it can be set with an environment variable.)

php/libraries/Database.class.inc
```diff
@@ -139,11 +139,17 @@ class Database
      * @param boolean $trackChanges whether to use the trackChanges mechanism on
      *                              this connection (defaults to true)
      *
-     * @return void
+     * @return bool True if DB conneciton established. False otherwise.
```
driusan: connection

modules/instrument_manager/php/instrument_manager.class.inc
xlecours: Our convention is still array()

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -44,76 +44,145 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
                           'Instrument',
                           'Instrument_Type',
                           'Table_Installed',
                           'Table_Valid',
                           'Pages_Valid',
-                         );
+                         ];
         $this->query   = " FROM test_names t";
-        $this->columns = array(
+        $this->columns = [
```
xlecours: Our convention is still array()

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -44,76 +44,145 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
                           't.Test_name as Instrument',
                           "'x' as Instrument_Type",
                           "'x' as Table_Installed",
                           "'x' as Table_Valid",
                           "'x' as Pages_Valid",
-                         );
-        $this->validFilters = array('');
-        $this->formToFilter = array();
+                         ];
+        $this->validFilters = [''];
```
xlecours: Our convention is still array()

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -44,76 +44,145 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
+        $this->formToFilter = [];
```
xlecours: Our convention is still array()

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -44,76 +44,145 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
-        } else {
-            $writable = true;
+        // Check to see whether the quatUser is configured.
+        // If this user is not configured, the instrument cannot be
+        // automatically installed as the table will not be able to be
+        // created.
+        $quat_user_is_configured = $this->isQuatUserConfigured();
+        if (!$quat_user_is_configured) {
+            $this->tpl_data['can_install'] = false;
+            $this->tpl_data['feedback']    = "Instrument installation is not "
```
xlecours: how about saying that the quat_user is not configured?

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -44,76 +44,145 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
         }
+
+        // Ensure we can write to folders in project/. If not, the upload
+        // form will not be created on the front-end. Warnings will explain to
+        // to the user that installation is not possible.
+        $writable = is_writable($this->path . "project/instruments")
+            && is_writable($this->path . "project/tables_sql");
         $this->tpl_data['writable'] = $writable;
+        if (!$writable) {
+            $this->tpl_data['feedback'] = "Instrument installation has been "
```
xlecours: This is a assumption. The fact is that the directory is not writable for the webserver user.

php/libraries/Database.class.inc
```diff
@@ -165,11 +171,18 @@ class Database
...
+            $this->_PDO = new PDO(
+                "mysql:host=$host;dbname=$database;charset=UTF8",
+                $username,
+                $password
+            );
+        } catch (PDOException $e) {
+            // This exception will not be included in the error log as it will
+            // likely include credentials.
+            error_log("Bad credentials or couldn't connect to database host.");
+            return false;
```
xlecours: This is changing the behavior of the connect function.
It would be "better" to thrown a \DatabaseException with a sanitized message.

modules/instrument_manager/php/instrument_manager.class.inc
johnsaigle: I don't think that's a useful message for a front-end user.

modules/instrument_manager/php/instrument_manager.class.inc
johnsaigle: Yes but it is consistent with how this module worked before. I explain this in the PR description.

driusan: The previous error message in the smarty template was "-Instrument directory or tables_sql not writable.
-Automatic uploading of instruments has been disabled.", which doesn't make that assumption.

johnsaigle: Ah see I read that as "Automatic uploading of instruments has been disabled [by your administrator]" not disabled as a result of the misconfiguration. 

I'll correct the message here.

johnsaigle:
@driusan @xlecours Could you dismiss your reviews or approve the changes?

modules/instrument_manager/php/instrument_manager.class.inc
xlecours: @johnsaigle The instrument manager module requires the superuser privilege so it will not be 'any' frontend user that receive that message. I am not against the proposed message, I just think that it do not give any clue on how to solve the "probleme" when the situation,e.g. not having quat_user configured, is not intentional.

xlecours:
@johnsaigle 
> @driusan @xlecours Could you dismiss your reviews or approve the changes?
There is still changes to make

modules/instrument_manager/php/instrument_manager.class.inc
johnsaigle: Fair enough. I will change the error message.

johnsaigle: @xlecours Done (see below). Please re-review when possible.

johnsaigle:
@driusan re: your dismissed review comment, I believe I incorporated all of your feedback in the subsequent commits. If you took at their titles they should match roughly what you suggested. Please let me know if I missed something.

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -62,58 +62,130 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
+                    . "install the instrument.";
+                http_response_code(409);
+                return;
+            }
+
+            move_uploaded_file($_FILES['install_file']['tmp_name'], $file_path);
+            chmod($file_path, 0644);
+            // Scripts in tools/ often make relative imports, so we must change
+            // our effective directory in order to use them.
+            chdir($this->path . "/tools");
```
driusan: I don't think this is necessary. htdocs and tools are at the same relative depth.

johnsaigle: It's the same functionality [as before ](https://github.com/aces/Loris/pull/3746/files#diff-7d481f75a6241f8079ffa0079d2e93baL84); all I did was add a comment and reorder things.

driusan: Fair enough

PR #3747, opened 2018-06-19
[survey.php] fix permission check order  20.0 

kongtiaowang:
This pull request fix permission check order for "survey.php".
https://redmine.cbrain.mcgill.ca/issues/14718

### DIFF
--- a/php/libraries/UserPermissions.class.inc
+++ b/php/libraries/UserPermissions.class.inc
@@ -123,6 +123,9 @@ class UserPermissions
         ) {
             return true;
         }
+        if (!isset($this->permissions[$code])) {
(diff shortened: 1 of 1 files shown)


driusan:
This change doesn't make any sense. If the survey.php is depending on a config setting that doesn't exist, the solution is to make it use a valid config setting, not change LORIS so that it doesn't validate config settings for superusers

ZainVirani:
I agree with @driusan. I also don't see how this solves the problem in the redmine ticket unless the user is the superuser, the permissions are retrieved and assigned to `$this->permissions[]` in `function setPermissions()`.

kongtiaowang:
ok, I will close this PR, and the issue only happened in one VM, not all VMs.

PR #3748, opened 2018-06-20
[install] Bugfix 500 Internal Server Error 20.0-release

intralizee:
This pull request `fixes 500 Internal Server Error from happening after user submits admin mysql detail`.

The namespace references to the Database.class.inc in the php/installer directory was breaking the installer by some exceptions being thrown. Solution is to try catch the exceptions.

### DIFF
--- a/htdocs/test.php
+++ b/htdocs/test.php
@@ -1,7 +1,0 @@
-<?php
-/**
- * Created by IntelliJ IDEA.
- * User: intralizee
- * Date: 6/27/18
- * Time: 11:37 AM
- */
--- a/htdocs/test.php
+++ b/htdocs/test.php
@@ -1,0 +1,7 @@
+<?php
+/**
+ * Created by IntelliJ IDEA.
+ * User: intralizee
+ * Date: 6/27/18
+ * Time: 11:37 AM
+ */
--- a/php/installer/Installer.class.inc
+++ b/php/installer/Installer.class.inc
@@ -187,21 +187,12 @@ class Installer
         }
         $DB = Database::singleton();
         try {
-            if ($DB->connectAndCreate(
+            $DB->connectAndCreate(
                 $values['dbname'],
                 $values['dbadminuser'],
                 $values['dbadminpassword'],
                 $values['dbhost']
-            ) == false
-            ) {
-                if (!empty($DB->lastErr)) {
-                    $this->_lastErr = $DB->lastErr;
-                } else {
-                    $this->_lastErr = "Could not connect to MySQL server "
-                        . "and create database with credentials provided.";
-                }
-                return false;
-            }
+            );
         } catch (Exception $e) {
             if (!empty($DB->lastErr)) {
                 $this->_lastErr = $DB->lastErr;
@@ -226,22 +217,12 @@ class Installer
         $DB = Database::singleton();
 
         try {
-            if ($DB->connect(
+            $DB->connect(
                 $values['dbname'],
                 $values['dbadminuser'],
                 $values['dbadminpassword'],
                 $values['dbhost']
-            ) == false
-            ) {
-                if (!empty($DB->lastErr)) {
-                    $this->_lastErr = $DB->lastErr;
-                } else {
-                    $this->_lastErr = "Could not connect to MySQL server "
-                        ."with credentials provided.";
-                }
-                return false;
-
-            }
+            );
         } catch (Exception $e) {
             if (!empty($DB->lastErr)) {
                 $this->_lastErr = $DB->lastErr;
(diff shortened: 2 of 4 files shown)


kongtiaowang:
@intralizee 
I have tested it on Ubuntu and Centos, both get the same error "[Fri Jun 22 14:35:18.920845 2018] [php7:warn] [pid 23232] [client 132.216.237.16:49532] PHP Warning:  preg_match(): Compilation failed: nothing to repeat at offset 33 in /var/www/Loris/php/installer/Database.class.inc on line 76, referer: http://132.206.37.22/installdb.php
[Fri Jun 22 14:35:18.920916 2018] [php7:warn] [pid 23232] [client 132.216.237.16:49532] PHP Warning:  preg_match(): Compilation failed: nothing to repeat at offset 33 in /var/www/Loris/php/installer/Database.class.inc on line 76, referer: http://132.206.37.22/installdb.php
[Fri Jun 22 14:35:19.523620 2018] [php7:error] [pid 23232] [client 132.216.237.16:49532] PHP Fatal error:  Uncaught ConfigurationException: Config file config.xml does not exist in the defined locations. in /var/www/Loris/php/libraries/NDB_Config.class.inc:146\nStack trace:\n#0 /var/www/Loris/php/libraries/NDB_Config.class.inc(73): NDB_Config->configFilePath()\n#1 /var/www/Loris/php/libraries/NDB_Factory.class.inc(118): NDB_Config::singleton(NULL)\n#2 /var/www/Loris/php/libraries/NDB_Client.class.inc(54): NDB_Factory->config(NULL)\n#3 /var/www/Loris/htdocs/index.php(22): NDB_Client->initialize()\n#4 {main}\n  thrown in /var/www/Loris/php/libraries/NDB_Config.class.inc on line 146, referer: http://132.206.37.22/installdb.php"

intralizee:
@kongtiaowang have you been installing Loris with the use existing database option checked?
I only tested the PR with the **Use existing database** and with the checkbox for "Check this if the database already exists but does not have Loris installed yet" selected. That case works for me

I wasn't supplied with MySQL and a root user when I got my VM as well. I'll try installing mysql locally on my development laptop to attempt your install configuration which I'm assuming is when a database does not exist yet.

kongtiaowang:
@intralizee It usually provides to the new users to install new Loris, so without  "existing database" is necessary to test. I can give you an installing VM, It has everything you need.

htdocs/test.php
```diff
@@ -0,0 +1,7 @@
+<?php
```
johnsaigle: What's this file here? Why do we need this information in the codebase?

intralizee: oops that shouldn't be there.

intralizee:
@kongtiaowang the installer script should now work. 
I wasn't able test on the VM you supplied but I did test on my local machine with mysql and a root user.

php/installer/Installer.class.inc
```diff
@@ -185,21 +186,30 @@ class Installer
...
-        ) {
+        try {
+            if ($DB->connectAndCreate(
+                $values['dbname'],
+                $values['dbadminuser'],
+                $values['dbadminpassword'],
+                $values['dbhost']
+            ) == false
+            ) {
+                if (!empty($DB->lastErr)) {
```
johnsaigle: This looks like it does the same thing as the code below. Isn't the try...catch the only part that's needed here?

php/installer/Installer.class.inc
```diff
@@ -214,21 +224,32 @@ class Installer
...
-        ) {
+
+        try {
+            if ($DB->connect(
+                $values['dbname'],
+                $values['dbadminuser'],
+                $values['dbadminpassword'],
+                $values['dbhost']
+            ) == false
+            ) {
```
johnsaigle: Same here. The code in the `try` block seems redundant.

php/installer/Installer.class.inc
intralizee: You're right, will remove.

driusan:
@intralizee Looks fine to me, but has the latest iteration been tested by someone else?

intralizee:
@driusan I tested it on my local machine when creating a non-existing db. Also tested when using an already existing db for installation. I don't believe it's been tested by someone else. There was a conflict when using with Shen's VM that I never figured out and am unable to reproduce.

kongtiaowang:
@intralizee @driusan I am testing this, it works fine on Ubuntu, I will test it on Centos later.
If everything is good, I will add the passmanualtest.

intralizee:
Added a fix where an invalid password resulted in a 500 error. @um4r12 helped.

PR #3560, opened 2018-03-23
[Candidate_parameters, Core] Implement Normalizing Consent 2/2

zaliqarosli:
PR 2/2: Implementation of https://github.com/aces/Loris/pull/3532 once it gets merged

1. Moves configurations from `config.xml` to _Config_ module
2. Updates _Candidate Parameters_ ajax to use new SQL tables
3. Updates _CouchDB_ import script for _DQT_
4. Remove deprecated lines from DB schema
5. Update `RBdata.sql` to incorporate new history table

**to test:**
1. Run all update scripts from https://github.com/aces/Loris/pull/3532
2. Access Profile -> Candidate Parameters -> Consent status
- test _Consent Status_ tab renders only when _useConsent_ in Config is set to Yes
- input fields in _Consent Status_ tab and their functionality remains the same
- data gets entered into new `candidate_consent_rel` table, not deprecated columns in `participant_status`
3. Test DQT results for the `demographics` instrument and consent fields
4. Resource raisinbread data (if you are okay with losing data)

PR #3751, opened 2018-06-20
Merge major back into 20.0-release

driusan:
@ridz1208 pointed out in #3560 that it was merged into the wrong branch. Luckily, that's the only thing that's been merged into major, so we can just merge major back into 20.0-release to fix..


driusan:
I mostly sent this as a PR to make sure it didn't fail Travis.. and now it's passing non-manual tests

PR #3752, opened 2018-06-20
[Statistics] (20.0) Re-enable Behavioural Statistics Double Data Entry module

johnsaigle:
## Summary

This pull request addresses the issues raised in Redmine 14744. I believe many are related to some malfunctioning in the Raisinbread test database and are not an issue with the source code. Similarly, some of these issues seem to occur on the test VM but I am not able to reproduce them on my sandbox. 

However, major issues existed preventing DDE functionality from working at all.

## Details

> behavioural stats: Candidate links to instrument forms don’t work. Takes you to instrument "page", and not the candidate instrument form page (404: Page not found)

- [x] `/instruments/` was added to the Candidate links in the menu statistics site templates, restoring their functionality.

> behavioural stats: Don’t have access to either “Click here for breakdown per participant” links even though I have all the permissions 

- [x] I was able to replicate this bug for Double Data Entry but it was related to the below bug. I did not have this issue with the first link in the module (tried on both my VM and testing).

> All ‘View Details’ links for Double Data Entry break [getSetting(‘statistics’) always returns null]

-  [x] The cause was that `Statistics_DD_Site` was overriding its parent's parent's parent's `__construct` function resulting in none of the essential NDB_Page things being loaded (such as form, Module, Identifier, etc.)

> - imaging stats tab doesn't work

- [x] Occurs on test VM only

## Other bonus changes

There was also no reason for its parent (`Statistics_Site`) to extend `NDB_Form` so it now extends `NDB_Menu` as the page is more like a Menu than a Form. This is reflected in the other template files in this module being prefixed with `menu_` rather than `form_`.

I also fixed a redundancy in the function `notexlcluded` as well as several PHP warnings which occurred when `$key` was `false`.

`getCssDependencies` was removed as it was identical to its parent method and so does not need to be overridden. 

### On branch `20.0-release`

1. Go to the `Statistics` module. Click the `Behavioural Statistics Tab`

2. Scroll down until the Double Data Entry section. Click any `View Details` link.

3. Verify you get a 500 error.

4. Go back and, under DDE, "Click here for breakdown per participant". Another 500 should appear.

5. Go to the main module page and click Imaging Statistics tab. It should work. If not, tell me.

6. **If you are on an instance with real instruments** please check if the candidate links to instruments work. Let me know if they still break. If they work, the problem is Raisinbread and not LORIS.

### On my branch

- Do the first 4 steps again. The links should take you to the properly loaded module.


driusan:
@zaliqarosli Did you test this or just review it?

johnsaigle:
@zaliqarosli helped me make the changes in the first place -- we did most of this together.

Not sure if she tested it explicitly on her VM though

zaliqarosli:
just tested it! if i helped make these changes, should I not also be reviewing it?

johnsaigle:
@zaliqarosli I am removing the testing label since this PR is attempting to do more than it originally was when you tested. Specifically, someone with functioning instruments needs to test the bugs you found. See the (updated) PR description for more info.

PR #3754, opened 2018-06-20
[Core] New Pull Request template

johnsaigle:
This Pull Request aims to improve our workflow by encouraging developers to add more information to their new pull requests.

Importantly, it encourages developers to provide testing instructions. The hope is that this well prevent PRs becoming cursed when no one knows how to help the author.

Most importantly, this PR template includes an element of public shame if people fail to follow it.

This is inspired by the presentation by @ridz1208.

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,8 +1,6 @@
-Please add some information beneath all of these headers. Then, **delete this sentence**.
-
 ### Brief summary of changes
 
-This Pull Request...
+
 
 ### This resolves issue...
 
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -8,10 +8,6 @@
 
 - [] Github? #
 
-### This PR improves LORIS by...
-
-
-
 ### To test this change...
 
 - [] 
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,21 @@
-This pull request `[DESCRIBE WHAT THIS PULL REQUEST IS TRYING TO ACCOMPLISH HERE]`. It `[DESCRIBE HOW IT ACCOMPLISHES THE GOAL HERE]`.
+Please add some information beneath all of these headers. Then, **delete this sentence**.
 
-See also: `[PROVIDE REFERENCES TO OTHER TICKETS, PULL REQUESTS, OR MAILING LIST THREADS HERE. OPTIONAL.]`
+### Brief summary of changes
+
+This Pull Request...
+
+### This resolves issue...
+
+- [] Redmine? #
+
+- [] Github? #
+
+### This PR improves LORIS by...
+
+
+
+### To test this change...
+
+- [] 
+
+# If this sentence shows up it means I didn't read the instructions!


PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,21 @@
-This pull request `[DESCRIBE WHAT THIS PULL REQUEST IS TRYING TO ACCOMPLISH HERE]`. It `[DESCRIBE HOW IT ACCOMPLISHES THE GOAL HERE]`.
+Please add some information beneath all of these headers. Then, **delete this sentence**.
 
-See also: `[PROVIDE REFERENCES TO OTHER TICKETS, PULL REQUESTS, OR MAILING LIST THREADS HERE. OPTIONAL.]`
+### Brief summary of changes
+
+This Pull Request...
```
driusan: I think you should just delete this and leave the headers.

PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,21 @@
...
+
+### This resolves issue...
+
+- [ ] Redmine? #
+
+- [ ] Github? #
```
driusan: The question mark should go after the number sign

PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,21 @@
...
+
+### This PR improves LORIS by...
```
driusan: This is redundant, if the summary isn't describing how it's improving LORIS, it's not a very good summary

johnsaigle: I see your point I had imagined the Summary as being more of a description of code changes whereas this section could be a more high-level contextual view.

ie.

```
## Summary

This PR prevents an error where the subclass was overriding a parent constructor when ti should not have been doing that. The functionality of its constructor was moved to another method. I also deleted some redundant code.

## This improves LORIS by...

Restoring the double data entry "View Details" links in the behavioural statistics module.

```

This definitely wasn't clear in my headers and I can change them. I feel that having a code-perspective and a context-perspective is valuable though. What are your thoughts?

um4r12: I agree with dave, there is a bit of overlap between the two, and I can see people either trying to copy-paste the contents of one into the other or adding filler text, especially when a small change is made. If there was only one section, it would allow for some flexibility, as people could choose to go into a bit of depth or not, depending on the changes made.

PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,21 @@
...
+
+
+
+### To test this change...
+
+- [ ] 
+
+# If this sentence shows up it means I didn't read the instructions!
```
um4r12: Sorry, I'm a little bit confused. Should we be removing both the first and last line of the template?

johnsaigle: At this point, yes. It's a sneaky trap

PULL_REQUEST_TEMPLATE.md
driusan: I still don't think there should be two sections. The more sections there are to fill out, the less likely people are to do it. There should just be one good description.

driusan:
I switched the branch to minor on the web gui, but now this needs to be rebased

PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,19 @@
...
-See also: `[PROVIDE REFERENCES TO OTHER TICKETS, PULL REQUESTS, OR MAILING LIST THREADS HERE. OPTIONAL.]`
+
+
+### This resolves issue...
+
+- [] Redmine? #
+
+- [] Github? #
+
+### This PR improves LORIS by...
```
zaliqarosli: I personally don't feel that this 'improves LORIS by' section is needed because it could easily overlap content going into 'brief summary of changes' and 'this resolves issue' sections. it would be a bit redundant

PR #3762, opened 2018-06-22
[20.0][Create Timepoint] Actually check visit label in validation function

ZainVirani:
This PR addresses [Redmine #14782](https://redmine.cbrain.mcgill.ca/issues/14782).
Specifically, if a user selects a subproject and visit label, and then switches the subproject to one where its no longer valid and clicks 'Create Timepoint' quickly enough, they can create a timepoint with an invalid visit label (this issue would also be solved by reactifying this module, but we can't do that before 20.0 is released).

I added a check to `isValidVisitLabel()` in `TimePoint.class.inc` to see if the submitted Visit Label is assigned to the submitted subproject by checking the config setting (XML).

To test this PR follow the steps outlined in the redmine ticket.

### DIFF
--- a/php/libraries/TimePoint.class.inc
+++ b/php/libraries/TimePoint.class.inc
@@ -281,7 +281,9 @@ class TimePoint
                     $visitLabelSetting['labelSet']['item']
                 );
                 foreach ($items as $item) {
-                    if ($item['#'] == $visitLabel) $isValid = true;
+                    if ($item['#'] == $visitLabel) {
+                        $isValid = true;
+                    }
                 }
             }
         }
--- a/php/libraries/TimePoint.class.inc
+++ b/php/libraries/TimePoint.class.inc
@@ -272,6 +272,28 @@ class TimePoint
             );
         }
 
+        $config =& \NDB_Config::singleton();
+        $visitLabelSettings = $config->getSetting('visitLabel');
+        $isValid            = false;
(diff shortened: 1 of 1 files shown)


PapillonMcGill:
This could be done using the new Visit class. I believe it as been included in the major (20.0) version.  Next step for the Visit class, is to convert all the usage of VisitLabel (in any form) to use this class. 

@driusan could you confirm in which version Visit class is added?

php/libraries/TimePoint.class.inc
```diff
@@ -272,6 +272,27 @@ class TimePoint
...
+        $visitLabelSettings = $config->getSetting('visitLabel');
+        $isValid            = false;
+        foreach (\Utility::toArray($visitLabelSettings) as $visitLabelSetting) {
+            if ($visitLabelSetting['@']['subprojectID'] == $subprojectID) {
+                $items = \Utility::toArray(
+                    $visitLabelSetting['labelSet']['item']
+                );
+                foreach ($items as $item) {
+                    if ($item['#'] == $visitLabel) {
+                        $isValid = true;
```
davidblader: maybe add a break statement(s) once you know it's valid

ZainVirani: good point

ZainVirani: @davidblader done

ZainVirani:
@davidblader please re review

PR #3763, opened 2018-06-26
[document_repository] Different Files with Same Name Not Saved Properly

kongtiaowang:
This pull request adds a category name make a new path to avoid file overwritten.


johnsaigle:
I believe @intralizee is addressing this issue in https://github.com/aces/Loris/pull/3700 so this PR is a duplicate. I think we should go with the other one since many people have reviewed and tested it already.

driusan:
I agree with @johnsaigle, the other PR also seems a little more robust (what happens with this one if the same user uploads something twice to the same category? What happens if it's recategorized? Etc.)

PR #3545, opened 2018-03-12
[Statistics] Demographics cleanup

ZainVirani:
This PR includes the demographics and site changes from #3379 [Closed]
Multiple new PRs will break down the remaining functional changes from that PR for easier testing.

Note: The purpose of the original PR was to eliminate the barrage of warnings this module throws into the error_log. I ran a count and it cuts down exactly 2 of them / probably 100. The majority of these warnings come from lines 118-133 of `php/stats_demographic.class.inc` (Undefined index warnings). If anyone has any ideas on how to properly do this, please comment.

Tested on RaisinBread.

PR #3764, opened 2018-06-26
[Statistics] Demographic Cleanup (rebased)

ZainVirani:
Purpose of PR:
- To cleanup variables in the Demographics tab of the Stats module
- Remove some warnings

Changes made: formatting, variable names, variable declarations, making logic more clear and concise

To test: follow the test plan for the statistics module, ensure all the tables in Demographics display the same values as they would have without this PR

Old PR: #3545


modules/statistics/php/stats_demographic.class.inc
```diff
@@ -77,14 +78,14 @@ class Stats_Demographic extends \NDB_Form
...
-        $centres,
+        $centers,
         $data,
         $Subsection="",
         $disclamer='',
         $projectID=null
     ) {
-
-        $tpl_data = array();
+        $projectID = $projectID == '' ? null : $projectID;
```
johnsaigle: Small thing, but how do you feel about this instead?
`$projectID = empty($projectID) ? null : $projectID;`

I find it a bit easier on the brain.

ZainVirani: I'm all for it

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -77,14 +78,14 @@ class Stats_Demographic extends \NDB_Form
...
+        $tpl_data  = array();
```
johnsaigle: Since this is a cleanup PR, could you use short array syntax? ie. `$tpl_data = [];`.

I don't know if we have a policy on this exactly but this is a more modern way of doing PHP iirc.

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -98,7 +99,7 @@ class Stats_Demographic extends \NDB_Form
         $tpl_data['DropdownName']     = $dropdown_name;
         $tpl_data['DropdownOptions']  = $dropdown_opt;
         $tpl_data['DropdownSelected'] = $dropdown_selected;
-        $tpl_data['Centers']          = $centres;
+        $tpl_data['Centers']          = $centers ?? array();
```
johnsaigle: Same here, short syntax.

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -163,65 +163,73 @@ class Stats_Demographic extends \NDB_Form
...
-        $subprojects = \Utility::getSubprojectsForProject($currentProject);
-        foreach ($subprojects as $key=>$val) {
-            $subprojList .= $key.",";
-        }
-        $subprojList = substr($subprojList, 0, -1);
-        if (!empty($subprojList)) {
-            $suproject_query ="AND s.SubprojectID IN ($subprojList)";
-        } else {
-            $suproject_query ='';
+        $subprojList      = '';
```
johnsaigle: Since this is a cleanup PR, can you make these variables one consistent case? There are 4 different conventions in just these variables.

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -163,65 +163,73 @@ class Stats_Demographic extends \NDB_Form
...
+        $subproject_query = '';
+        $Visits           = \Utility::getExistingVisitLabels($demographicProject);
+        $project          = $demographicProject != '' ? null : $demographicProject;
+        $subprojects      = \Utility::getSubprojectsForProject($project);
+
+        if (!empty($subprojects)) {
+            foreach ($subprojects as $key=>$val) {
```
johnsaigle: If `$val` is not used, just do `foreach ($subprojects as $subproject) {`

In fact, this section may be better written as

> $subproject_list = implode(',', $subprojects);

That way you can also remove line 192 because it seems to be chopping off a comma that shouldn't be there anyway...

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -163,65 +163,73 @@ class Stats_Demographic extends \NDB_Form
...
+                $subprojList .= "$key,";
+            }
+            $subprojList = substr($subprojList, 0, -1);
+            if (!empty($subprojList)) {
```
johnsaigle: Redundant as `$subprojList` will only be non-empty when `$subprojects` is non-empty. The check on 188 is enough.

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -163,65 +163,73 @@ class Stats_Demographic extends \NDB_Form
...
 
         //SITES
-        if ($_REQUEST['DemographicSite'] ?? ''
-        ) {
-            $Param_Site          ='AND (c.CenterID IS NULL OR c.CenterID=:sid) ';
-            $this->params['sid'] = htmlspecialchars($_REQUEST['DemographicSite']);
-        } else {
-            $Param_Site = '';
+        $demographicSite = $_REQUEST['DemographicSite'] ?? '';
+        $demographicSite = htmlspecialchars($demographicSite);
```
johnsaigle: Move this below the `!empty` check

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -163,65 +163,73 @@ class Stats_Demographic extends \NDB_Form
         foreach (\Utility::getProjectList() as $key => $value) {
             $projects[$key] = $value;
         }
-        $currentProject = null;
-        if ($_REQUEST['DemographicProject'] ?? ''
-        ) {
-            $currentProject = htmlspecialchars($_REQUEST['DemographicProject']);
+
+        $demographicProject = $_REQUEST['DemographicProject'] ?? null;
```
johnsaigle: Here you make the variable `null` when not set and elsewhere you set it to the empty string. I think it would be better to choose just one unless the difference is significant. If that's the case, a comment explaining why would be good.

modules/statistics/php/stats_demographic.class.inc
johnsaigle: I see there are a lot of old `array()` style declarations below also. Would be nice to have these updated but it's not essential. I won't add more comments for each instance.

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -230,18 +238,19 @@ class Stats_Demographic extends \NDB_Form
...
-                          $Param_Project
-                          $Param_Site
-                      GROUP BY s.subprojectid",
+                    count(DISTINCT(c.PSCID)) as val
+            FROM candidate as c
+            LEFT JOIN session s ON (s.CandID=c.CandID)
+            WHERE c.CenterID <> '1'
+            AND c.Entity_type != 'Scanner'
+            AND c.Active = 'Y'
+            $Param_Project
```
johnsaigle: These are hard-coded strings above so should be safe from SQLi. 👍 A comment would be nice just to reassure anyone who reviews this code in the future.

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -374,60 +386,63 @@ class Stats_Demographic extends \NDB_Form
         );
         $instrument_list = \Utility::getAllInstruments();
         $this->tpl_data['all_instruments'] = $instrument_list;
-        if ($_REQUEST['DemographicInstrument'] ?? ''
-        ) {
+        $demographicInstrument = $_REQUEST['DemographicInstrument'] ?? '';
+        $demographicInstrument = htmlspecialchars($demographicInstrument);
```
johnsaigle: This can go below the empty check

modules/statistics/php/stats_demographic.class.inc
johnsaigle: @ZainVirani I'm wrong actually. Apparently old array syntax is our standard so leave this as they are.

modules/statistics/php/stats_demographic.class.inc
ZainVirani: Why? if its !empty it will be assigned to `$this->params['sid']`, so we would want it to be escaped before, no?

modules/statistics/php/stats_demographic.class.inc
ZainVirani: same question as above. its a !empty check

ZainVirani:
@johnsaigle made most of your changes, please check 😄

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -55,38 +56,38 @@ class Stats_Demographic extends \NDB_Form
...
+        $centers,
         $data,
-        $Subsection="",
+        $subsection="",
         $disclamer='',
         $projectID=null
     ) {
-
-        $tpl_data = array();
+        $projectID = empty($projectID) ? null : $projectID;
```
HenriRabalais: I'm pretty sure you could use the null coalesce operator here: 
`$projectID = $projectID ?? null;`

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -143,15 +144,14 @@ class Stats_Demographic extends \NDB_Form
     function setup()
     {
         parent::setup();
-
-        $DB          =& \Database::singleton();
+        $db          =& \Database::singleton();
```
HenriRabalais: If you're changing this line I think you may as well remove the ampersand `=&` -> `=`

modules/statistics/php/stats_demographic.class.inc
johnsaigle: Mostly because it's consistent with the order on line 171 and because there's no use calling the function on an empty string. 

It's not ultimately a huge issue and I won't block the PR just for this.

johnsaigle:
Looks good! Did not test.

modules/statistics/php/stats_demographic.class.inc
johnsaigle: @HenriRabalais I thought so too at first but it's more subtle than that: this line says "when `$projectID` is empty, instead make it null. Otherwise leave it alone."

Null coalesce would not make this change as it only kicks in when a variable is not set (different from when it IS set to empty).

johnsaigle: This line could probably benefit from a comment in the code actually.

modules/statistics/php/stats_demographic.class.inc
ZainVirani: OH i see now, sorry i thought you meant move it below that entire block, not inside the !empty check. I agree.

modules/statistics/php/stats_demographic.class.inc
HenriRabalais: Makes sense!

ZainVirani:
@HenriRabalais changes addressed. ty 4 review

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -55,38 +56,40 @@ class Stats_Demographic extends \NDB_Form
...
-        $dropdown_opt,
-        $dropdown_selected,
-        $centres,
+        $dropdownName,
+        $dropdownOpt,
+        $dropdownSelected,
+        $centers,
         $data,
-        $Subsection="",
+        $subsection="",
```
zaliqarosli: this is such a minor annoyance for me haha but could we stick with either single quotes or double quotes

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -55,38 +56,40 @@ class Stats_Demographic extends \NDB_Form
...
         $disclamer='',
         $projectID=null
     ) {
-
-        $tpl_data = array();
+        // This line replaces the empty $projectID string with null
+        // `$projectID ?? null` would retain empty strings.
```
zaliqarosli: i don't think this line is necessary. line 87 is good enough of an explanation of what you want to do, i think

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -163,203 +165,214 @@ class Stats_Demographic extends \NDB_Form
...
-        }
-        $subprojList = substr($subprojList, 0, -1);
-        if (!empty($subprojList)) {
-            $suproject_query ="AND s.SubprojectID IN ($subprojList)";
-        } else {
-            $suproject_query ='';
+        $subprojList     = '';
+        $subprojectQuery = '';
+        $visits          = \Utility::getExistingVisitLabels($demographicProject);
+        $project         = $demographicProject != '' ? null : $demographicProject;
```
zaliqarosli: what's this line doing?

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -95,10 +98,10 @@ class Stats_Demographic extends \NDB_Form
...
         $tpl_var = \Utility::getSubprojectsForProject($projectID);
         $tpl_data['Subprojects']      = $tpl_var;
-        $tpl_data['DropdownName']     = $dropdown_name;
-        $tpl_data['DropdownOptions']  = $dropdown_opt;
-        $tpl_data['DropdownSelected'] = $dropdown_selected;
-        $tpl_data['Centers']          = $centres;
+        $tpl_data['DropdownName']     = $dropdownName;
+        $tpl_data['DropdownOptions']  = $dropdownOpt;
+        $tpl_data['DropdownSelected'] = $dropdownSelected;
+        $tpl_data['Centers']          = $centers ?? array();
```
zaliqarosli: why does this need to be an empty array if not set?

zaliqarosli:
testing: filtering general demog stats table by projects doesn't work.

modules/statistics/php/stats_demographic.class.inc
johnsaigle: @zaliqarosli This is what that line is doing.

@ZainVirani This line definitely needs a comment. That's the 3rd person who has been tripped up by it.

modules/statistics/php/stats_demographic.class.inc
johnsaigle: @zaliqarosli This comment also applies to your comment below.

modules/statistics/php/stats_demographic.class.inc
johnsaigle: See `templates//form_stats_behavioural.tpl`. I think PHP will give a warning or notice if it tries to iterate over an unset or null value, whereas an empty array should be fine.

modules/statistics/php/stats_demographic.class.inc
zaliqarosli: @johnsaigle i think you misunderstanding me. I understand what the line does - it is explained with line 87. i just don't think the addition line 88 is needed. i think what @ZainVirani did here with the comments is more than enough

modules/statistics/php/stats_demographic.class.inc
zaliqarosli: @johnsaigle this isn't really the same as `$projectID = empty($projectID) ? null : $projectID` .. it's actually reading `$projectID = !empty($projectID) ? null : $projectID`which doesn't really make sense to me

modules/statistics/php/stats_demographic.class.inc
ZainVirani: yes will do

modules/statistics/php/stats_demographic.class.inc
ZainVirani: I think its perfect.

modules/statistics/php/stats_demographic.class.inc
ZainVirani: Good catch!!! Thanks

ZainVirani:
@zaliqarosli the bug you found should be fixed now! :)

zaliqarosli:
@ZainVirani the whole tab now breaks! 😬

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -163,203 +165,214 @@ class Stats_Demographic extends \NDB_Form
...
-            $suproject_query ='';
+        $subprojList     = '';
+        $subprojectQuery = '';
+        $visits          = \Utility::getExistingVisitLabels($demographicProject);
+        $project         = empty($demographicProject) ? null : $demographicProject;
+        $subprojects     = \Utility::getSubprojectsForProject($project);
+
+        if (!empty($subprojects)) {
+            $subprojList     = implode(',', $subprojects);
+            $subprojectQuery = "AND s.SubprojectID IN ($subprojList)";
```
zaliqarosli: the bug is coming from here. the query is coming out as `AND s.SubprojectID IN (Stale test,Fresh,Low Yeast,High Yeast)`. you want `AND s.SubprojectID IN ('Stale test', 'Fresh', 'Low Yeast', 'High Yeast')`

zaliqarosli:
general stats table works but Recruit breakdown by Sex breaks 😢

ZainVirani:
@zaliqarosli I'll look into it. Thank you :)

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -32,21 +32,22 @@ class Stats_Demographic extends \NDB_Form
...
+        $user = \User::singleton();
+        //TODO: Create a permission specific to statistics
         return $user->hasPermission('data_entry');
     }
     /**
      * InCenter function
      *
-     * @param string $centerID the value of centreID
-     * @param string $Centres  the value of Centres
+     * @param string $centerID the value of centerID
```
PapillonMcGill: Do we use US English or Canadian (UK) English in our spelling?
I mostly saw Canadian spelling elsewhere (Code, poster, ...)

ZainVirani: I'm not sure, I originally added this change when I was reimplementing @johnsaigle's original PR. @driusan, thoughts?

driusan: @samirdas used to insist on canadian spelling, but I don't think he's cared recently. I don't think "Center" has ever been spelled with Canadian spelling in LORIS, though.

Try bringing it up in a LORIS meeting.

PapillonMcGill: Putted it on the next Loris meeting agenda

PapillonMcGill: A quick grep in the code reveal 874 center vs 20 centre. Note that it include centered (at least once).
I think it is revealing. But still will bring it up at the meeting.

johnsaigle: My original goal was just to make the spelling consistent but I have no preference which one we use!

PapillonMcGill: I just want that we use a consistent language, not a mix of everything.
It make Loris look better (in my humble opinion).

ZainVirani: @PapillonMcGill I completely agree. After tomorrow's discussion I'll do a `:%s...` to whatever we decide.

PapillonMcGill: @ZainVirani Just be carefull, there are instance of string that contain `center` inside a word like `centered` or `docenter`. It might also be true with `centre`.

PapillonMcGill: removing all variable name, comments, ... I currently have 3 "center" and 1 "centre".

driusan:
@ZainVirani This PR is old enough (and big enough) that I'm afraid of Travis breaking since this was run on Travis before major was merged into minor (including the stricter phan settings.) Can you rebase this to make sure it has a run with the current Travis settings?

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -374,100 +387,104 @@ class Stats_Demographic extends \NDB_Form
...
-        ) {
-            $Subcategories = array(
-                              'Complete',
-                              'Incomplete',
-                             );
-            $result        = $DB->pselect(
+        $this->tpl_data['all_instruments'] = $instrumentList;
+        $demographicInstrument = $_REQUEST['DemographicInstrument'] ?? '';
+        if (!empty($demographicInstrument)) {
+            $demographicInstrument = htmlspecialchars($demographicInstrument);
```
zaliqarosli: @ZainVirani can you explain why are you are adding the !empty and htmlspecialchars? i'm trying to understand why the "big table" is breaking

ZainVirani: `!empty($demographicInstrument)` is the equivalent to `$_REQUEST['DemographicInstrument'] ?? ''`

This PR just rewrites code, it isn't supposed to change functionality. If I did something it's because the previous code did the same thing but in a messy way.

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -374,100 +387,104 @@ class Stats_Demographic extends \NDB_Form
...
+                        c.gender as Subcat,
+                        count(s.CandID) as val
+                FROM session s
+                JOIN candidate c ON (s.CandID=c.CandID)
                 WHERE s.active='Y' AND s.CenterID <> '1'
-                    AND (s.Current_stage IN ('Visit', 'Screening', 'Approval')
-                    $suproject_query)
-                    AND COALESCE(s.Screening,'') NOT IN ('Failure', 'Withdrawal')
-                    AND COALESCE(s.Visit,'') NOT IN ('Failure', 'Withdrawal')
+                AND s.Current_stage IN ('Visit', 'Screening', 'Approval')
```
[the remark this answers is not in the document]
PapillonMcGill: Maybe the $subproject_query part that is not included anymore?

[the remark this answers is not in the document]
ZainVirani: I'll look into it now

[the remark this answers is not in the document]
ZainVirani: @zaliqarosli I don't think the brackets make a difference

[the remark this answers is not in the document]
ZainVirani: @zaliqarosli @PapillonMcGill @cmadjar  fixed the query, but minor also returns all 0s on this table, so...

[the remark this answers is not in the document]
cmadjar: @ZainVirani mmm I checked this out on 20.0 and it was filled with numbers. Is it possible something was introduced to minor that made this break in minor then?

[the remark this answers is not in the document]
zaliqarosli: i have numbers now for both minor and this PR branch

PR #3765, opened 2018-06-26
[API] API Client class and Candidate Unit Tests

johnsaigle:
### Background
These classes were added to simplify the process of making a scripted series of requests to the LORIS API. 

When these classes are merged it will be easy to quickly create tests to interact with the API by leveraging their syntax. A proof-of-concept test is included here to demonstrate the usage of these classes.

### Brief technical summary

This pull request creates creates Client classes for doing requests with the LORIS API. A couple layers of wrappers are made around PHP's cURL library. These classes take care of the construction of cURL requests and session management. Implemented so far are GET and POST and the code is easily extensible to allow for PUT, PATCH etc. in the future.

All new code is PSR 7 compatible.

### This improves LORIS by...

- [x] Devs can easily interact with the API with very easy syntax and without knowing cURL

- [x] Power-users can leverage HttpClient to bootstrap API scripts

- [x] API tests can be written quickly

- [x] Lays potential groundwork for using the API internally within LORIS. `lorisGET` and `lorisPOST` can provide a readable way to interact with the API from the modules.

### Testing

This PR introduces a unit test for the Candidate endpoint in the API. If it passes Travis then the underlying classes added here must be working.

---

This PR was co-written by @xlecours.


kongtiaowang:
@xlecours @johnsaigle 
Could u figure out the username and password for the test?
I always get 401.

driusan:
Could only switch the branch to minor in the frontend, needs to be rebased properly to get rid of the extra commits. This is far too late to add this to 20.0

xlecours:
According to @kongtiaowang, this won't work with the built-in web server because it do not support multiple .htaccess files. We sill need to make the API a module

xlecours:
As @driusan pointed out, The Travis use the router.php file instead of htaccess. So we could change that file to add routes for the api. Or make it a module instead.

johnsaigle:
@xlecours Looks like it's working :smile:

src/Http/CurlClient.php
```diff
@@ -0,0 +1,207 @@
...
+     * by reference (===) is not possible.
+     *
+     * {@link https://github.com/php-fig/fig-standards/blob/master/accepted/
+     *                          PSR-7-http-message-meta.md#why-value-objects}
+     *
+     * @param RequestInterface $request Request object. See link above.
+     *
+     * @return ResponseInterface
+     */
+    public function sendRequest(RequestInterface $request): ResponseInterface
```
driusan: You seem to have copied this from PSR-18 without having included the "implements" line for the class.

src/Http/CurlClient.php
```diff
@@ -0,0 +1,207 @@
...
+        // Capture response instead of printing it.
+        curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+        // Connect via HTTP/2 if available. Some versions of PHP/cURL don't
+        // know about it so we are using the magic number 3 here.
+        // The usual alias is CURL_HTTP_VERSION_2_0 but Travis has an issue with
+        // it.
+        curl_setopt($ch, CURLOPT_HTTP_VERSION, 3);
+
+        // For VMs with expired ssl certificate change the value to false.
+        curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, true);
```
driusan: We don't currently depend on the curl PECL extension, so this will break on a lot of LORIS installs.. especially if it's not declared as required in the composer.json file.

test/integrationtests/api/ApiTestCase.php
```diff
@@ -0,0 +1,47 @@
...
+        $this->httpclient = new \LORIS\tests\api\HttpClient(
+            new Uri(
+                $this->factory->settings()->getBaseURL() . '/api/v0.0.3-dev/'
+            )
+        );
+    }
+
+    /**
+     * @test
+     * @doesNotPerformAssertions
```
driusan: This is a true statement.

htdocs/router.php
```diff
@@ -70,6 +70,12 @@
     $_REQUEST['subtest']   = $getParams[3];
 
     include_once __DIR__ . "/main.php";
+} elseif (preg_match(
+    '#^api/.*#',
+    $url
+)) {
+    $endpoint = explode('?', $url)[0];
+    include_once __DIR__ . '/' . $endpoint;
```
driusan: Missed the actual comment that I was here for in my review..

I couldn't get this to work. I'm still getting a 404 with the error "[Thu Sep 20 11:02:31 2018] ::1:65216 [404]: /api/v0.0.2/projects - Undefined error: 0" while trying to access `/api/v0.0.2/projects/` through the php built in web server

driusan: By changing the first if statement to:

```
f ($request != '/'
    && file_exists(__DIR__ . $request)
    && $request != "/acknowledgements/"
    && preg_match('/^api\//', $url) !== 1
) { 
```

I was at least able to reach this part of the router, but I'm still not sure how this is supposed to work since there isn't a 1-1 mapping between file system layout and URL endpoints (in particular, the capitalization and file extensions differ)

driusan: (and there's the parts that are dynamic in between like the IDs)

src/Http/CurlClient.php
johnsaigle: I think we can use apt instead of pecl, not sure if that matters. 

I can also include this line in the automatic update script -- if we plan to use it in the future then it could help ameliorate this issue.

I'm also hoping to use cURL for this PR https://github.com/aces/Loris/pull/3961 to interact with the Have I Been Pwned API.

Are you opposed to including curl as a requirement in future versions?

htdocs/router.php
johnsaigle: @xlecours Could you comment here? I believe you did this part of the code.

src/Http/CurlClient.php
driusan: It doesn't matter how it's installed, I just mean that it needs to declared in composer.json if it's depended on.

I've started working on #3970 (both because it's long overdue and because I need to be able to use the API for real under my dev environment..). With that, I don't think you need curl at all. You can get the ResponseInterface by simply calling the handler function directly on the router. You'd get a similar level of confidence but eliminate the network overhead, process forking overhead, and dependency on a network to run your test.

xlecours: We need something that can send an HTTP request. I need it for cbrain api calls. @driusan Would adding `ext-curl` to the required package be sufficient?

htdocs/router.php
xlecours: @driusan This PR is to test the current (v0.0.3-dev) api endpoints without any kind of url rewriting. `/api/v0.0.2/projects/` wont work with that. But `/api/v0.0.2/Projects.php?PrintProjects=true` will

src/Http/CurlClient.php
xlecours: I don't think we can claim to implement the whole interface. We are missing the Exception handling part; for the least.

xlecours: But yes, `sendRequest` is aiming at being compliant with PSR-18

htdocs/router.php
driusan: But there are no API endpoints without any kind of URL rewriting. The API is defined in terms of the URLs.

johnsaigle:
@xlecours @driusan From what I understand, this PR needs some refactoring but I don't want to throw it all away. Maybe we can schedule a meeting at some point to figure out how exactly we want to move forward here.

johnsaigle:
Blocked by #3968

driusan:
Closing this because a lot of the API is in the middle of being refactored with it being converted to a module, and once that's done there are better ways of doing this (either calling the handle() method directly with a request object, or by using the guzzle dependency which was recently added to LORIS.)

It's probably best to do this in a more modern way once the conversion is complete.

PR #3767, opened 2018-06-27
[schema] Fix the assumed reference keys in default schema

ridz1208:
This removes magical IDs and replaces them by subquery pointing to the same entry. This was causing a failure in installing loris on any database where AUTO_INCREMENT > 1


ridz1208:
@xlecours could you make sure I made all the correct associations between the integer values and the subqueries. the table they are referencing is right above the code changes

SQL/0000-00-00-schema.sql
```diff
@@ -1955,45 +1955,45 @@ CREATE TABLE `feedback_mri_predefined_comments` (
...
+  ((SELECT `CommentTypeID` FROM `feedback_mri_comment_types` WHERE `CommentName`='Overall'),'copy of prev data'),
+  ((SELECT `CommentTypeID` FROM `feedback_mri_comment_types` WHERE `CommentName`='Intensity artifact'),'checkerboard artifact'),
+  ((SELECT `CommentTypeID` FROM `feedback_mri_comment_types` WHERE `CommentName`='Intensity artifact'),'horizontal intensity striping (Venetian blind effect, DWI ONLY)'),
+  ((SELECT `CommentTypeID` FROM `feedback_mri_comment_types` WHERE `CommentName`='Intensity artifact'),'diagonal striping (NRRD artifact, DWI ONLY)'),
+  ((SELECT `CommentTypeID` FROM `feedback_mri_comment_types` WHERE `CommentName`='Intensity artifact'),'high intensity in direction of acquisition'),
+  ((SELECT `CommentTypeID` FROM `feedback_mri_comment_types` WHERE `CommentName`='Intensity artifact'),'signal loss (dark patches)'),
+  ((SELECT `CommentTypeID` FROM `feedback_mri_comment_types` WHERE `CommentName`='Overall'),'red artifact'),
+  ((SELECT `CommentTypeID` FROM `feedback_mri_comment_types` WHERE `CommentName`='Overall'),'green artifact'),
+  ((SELECT `CommentTypeID` FROM `feedback_mri_comment_types` WHERE `CommentName`='Overall'),'blue artifact'),
```
PapillonMcGill: Shouldn't the last 3 (red, green and blue) be 'Dominant Direction Artifact (DWI ONLY)'?

xlecours: I second that.

xlecours:
@ridz1208 See Melanie's comments

SQL/0000-00-00-schema.sql
driusan: Why? Do only DWI images have red/green/blue artifacts? @cmadjar ?

cmadjar: @PapillonMcGill is right, it should go under "Dominant Direction Artifact (DWI ONLY)"

As far as I know, only DWI can produce RGB maps with those artifacts.

ridz1208:
@PapillonMcGill  @xlecours good catch, fixed

driusan:
@ridz1208 does this need to be done on 20.0-release too?

ridz1208:
@driusan no, its only for new installs on clusters. not high risk enough

PR #3769, opened 2018-06-27
[instrument_builder] Fix error where the value for year was not exactly 4 digits

kongtiaowang:
This pull request adds a new function to check the year number should not be more than 4 digits numbers.
https://redmine.cbrain.mcgill.ca/issues/14781


modules/instrument_builder/jsx/react.questions.js
```diff
@@ -631,6 +631,15 @@ var AddElement = React.createClass({
         });
         hasError = true;
       }
+      if (minYear > 9999 || minYear < 1900 || maxYear > 9999 || maxYear < 1900) {
+        let temp = (this.state.error) ? this.state.error : {};
+
+        temp.dateOption = "The year should be 4 digits number";
```
johnsaigle: temp.dateOption = "The year must have exactly 4 digits.";

modules/instrument_builder/jsx/react.questions.js
```diff
@@ -631,6 +631,15 @@ var AddElement = React.createClass({
         });
         hasError = true;
       }
+      if (minYear > 9999 || minYear < 1900 || maxYear > 9999 || maxYear < 1900) {
```
PapillonMcGill: What happen if the min or max year entered is 1818 (between 1000 and 1900)? It as 4 digit but would still fail with a message saying it need 4 digit.

@driusan @ridz1208 
Would it be better to have a different message stating that year under 1900 are invalid?

driusan:
@kongtiaowang I changed this to minor as it seems to be adding new functionality, but now it needs to be rebased.

modules/instrument_builder/jsx/react.questions.js
driusan: Why don't we want to be supporting years < 1900? There are questions someone might want to ask where that would be a valid answer.

modules/instrument_builder/jsx/react.questions.js
```diff
@@ -631,6 +631,15 @@ var AddElement = React.createClass({
...
+        let temp = (this.state.error) ? this.state.error : {};
+
+        temp.dateOption = "The year must have exactly 4 digits.(between 1900 and 9999)";
```
miebeers: Please change the error message to "The year must have exactly 4 digits. Please choose an integer number between 1900 and 9999."

modules/instrument_builder/jsx/react.questions.js
PapillonMcGill: I don't expect people being 200 years old but should it be a different error message than date must be 4 chars long?  1818 is 4 chars long after all.

driusan: Years don't need to be ages. There are perfectly reasonable question to ask about years in a study that aren't directly related to the person's age ("When was the house that you were born in built?", "What year did your family migrate to Nunavut?", "When were your parents born?" (in a study relating to the elderly) etc)

PapillonMcGill: either change code check for `... minYear < 1900 ...` to `... minYear < 1000 ...` or text of the error message.

modules/instrument_builder/jsx/react.questions.js
```diff
@@ -631,6 +631,15 @@ var AddElement = React.createClass({
         });
         hasError = true;
       }
+      if (minYear > 9999 || minYear < 1900 || maxYear > 9999 || maxYear < 1900) {
+        let temp = (this.state.error) ? this.state.error : {};
+
+        temp.dateOption = "The year must have exactly 4 digits. Please choose an integer number between 1900 and 9999.";
```
zaliqarosli: could we add an 'e' after 'befor' on line 628? (https://github.com/kongtiaowang/Loris/blob/33ed933d2a665b4636c20242d033e410f5069ddd/modules/instrument_builder/jsx/react.questions.js#L628) i'm commenting that here because i can't comment on that line

PapillonMcGill: Excellent catch!

modules/instrument_builder/jsx/react.questions.js
ridz1208: I agree with dave, making the changes now

ridz1208:
@PapillonMcGill @ZainVirani 

made changes , please review and test accordingly

modules/instrument_builder/jsx/react.questions.js
```diff
@@ -625,16 +625,16 @@ var AddElement = React.createClass({
...
           error: temp
         });
         hasError = true;
       }
-      if (minYear > 9999 || minYear < 1900 || maxYear > 9999 || maxYear < 1900) {
+      if (minYear > 9999 || minYear < 1000 || maxYear > 9999 || maxYear < 1000) {
         let temp = (this.state.error) ? this.state.error : {};
 
-        temp.dateOption = "The year must have exactly 4 digits. Please choose an integer number between 1900 and 9999.";
+        temp.dateOption = "The year must have exactly 4 digits. Please choose an integer number between 1000 and 9999.";
```
PapillonMcGill: I would have prefer to keep the minimum year at 1900 and adjust the message like : 
"The year must have exactly 4 digits and be greater than 1900. Please choose an integer number between 1900 and 9999."

ridz1208: y 1900 ? we can clearly have values in the 1800's or 1700's no ?

ZainVirani: @ridz1208 the oldest living human being was born in 1901. technically you are correct but I don't see why we would allow it.

PapillonMcGill: @ridz1208 see comment from @driusan hidden somewhere in these comment.

zaliqarosli:
besides the discussion on what the min year should be, everything works fine!

modules/instrument_builder/jsx/react.questions.js
PapillonMcGill: You'd have to be over 118 years old to be born in 1900 which is more than the oldest (known) person on earth.  There is about zero chance that a date for a visit, other test or date of birth would be lower than that.

ridz1208: what about the date of your grand father's birth ? or the date your house was built? 

there is no boundary to questions that would come up in an instrument... its not only age

PapillonMcGill: @ridz1208 I only reported what  @driusan comments. For the records, my grandparent where born in early 20th century and I only know the year for 2 of them. Yes, date the house was build may be valid, specially on the other side of the pound.

PR #3771, opened 2018-06-28
[SQL] introduce Enum replacement normalization

PapillonMcGill:
This pull request `provide normalized way to replace ENUM attributes in SQL schema`. It `provide an option base table from witch other option tables can refer to in order to restrict the possible values of attributes`.

See also: `https://forum.loris.ca/t/sql-enum-db-type-replacement/140/6`


kongtiaowang:
You need to update RBdata.sql.

PapillonMcGill:
@kongtiaowang Thanks Shen for the comment. There is more than that, I used a reserved word for a table name. Could work with quote but I prefer to modify using no reserved name.

docs/SQLModelingStandard.md
```diff
@@ -112,16 +113,49 @@ infrastructure will be re-designed to follow this guideline.*
...
 
+### ENUM replacement
 
+- A **base lookup table** (`option`) exist to store all possible values for 
+  every enum requirement.
+- When an ENUM type field is needed, create a **second level** lookup table 
+  (i.e., `yes_no_option` for `yes` and `no` values) with a single attribute
+  referencing the **primary key** attribute of the `option` table. That second 
+  level table will only store the `OptionID` of the required values. In the 
+  table that require an ENUM type attribute, make a **foreing key** constraint
```
davidblader: typo in `foreign`

PapillonMcGill: Thanks, change done!

xlecours:
@PapillonMcGill This seems to be adding Language feature. I don't see where the enum normalization is happening

PapillonMcGill:
@xlecours The title might not be the best but as mentioned in the SQLModelignStandard.md files, using this, all occurrences of a value (i.e. `pass`) will have the same ID and all lookup  table would be of the same model. Thus if you need to change table because you had `pass` and `fail` as possible value and that lookup table is used at others places (thus you can't change possible values without affecting other portion of the code), you can just create a new lookup table with `pass`, `fail` and as example `not completed` and change the foreign key without have to worry about the ID not being the same. It follow the discussion on the forum.

xlecours:
@PapillonMcGill I am still not convince. I don't see any gain here compare to creating a yes_no_enum table with 2 columns (id, value) and make foreign keys reference to it. Before changing something in the yes_no_enum table, one would need to query information schema for possible table that reference it as well. 
With this PR's proposition, if I want to change pass/fail for Pass/Fail without affecting all other tables that might use that, I would need to query information schema for all table names that end with `_enum` then look in those tables for values referencing the ones I want to change. It feels like a complicated solution for a not so complex problem.

PapillonMcGill:
@xlecours This PR implement what was discussed in the forum (where you seemed to agree with). 
@driusan you where silent in the forum. 

I wanted to create a common way to work with lookup tables as (I don't remember who) someone said we need a global way to do it. We may go with one lookup table for every occurrence of the need for an enumeration.
 Maybe we need to go back to the forum in order to clear that matter first. So closing this PR for now.

driusan:
@PapillonMcGill Xavier's opinion on the forum seems to have been the same as his opinion here.

I was silent there (and here) because I didn't have anything to add on top of the discussion that was already happening.

PR #3775, opened 2018-06-28
[20.0][Dashboard] Fix discrepancy in data entry conflicts

davidblader:
The count in dashboard for conflicts is different from the count shown in the conflict resolver module
This makes the query responsible for retrieving the count closer to what's used in the conflict resolver module

### DIFF
--- a/modules/dashboard/php/dashboard.class.inc
+++ b/modules/dashboard/php/dashboard.class.inc
@@ -129,11 +129,11 @@ class Dashboard extends \NDB_Form
             if ($user->hasPermission('access_all_profiles')) {
                 $this->tpl_data['conflicts']      = $DB->pselectOne(
                     "SELECT COUNT(*) FROM conflicts_unresolved cu
-                    LEFT JOIN flag ON (cu.CommentId1=flag.CommentID) 
+                     LEFT JOIN flag ON (cu.CommentId1=flag.CommentID) 
                      LEFT JOIN session s ON (flag.SessionID=s.ID)
-			LEFT JOIN candidate c ON (s.CandID=c.CandID)
-                      WHERE s.CenterID <> 1
-			AND s.Active='Y' AND c.Active='Y'",
+			         LEFT JOIN candidate c ON (s.CandID=c.CandID)
+                     WHERE s.CenterID <> 1
+			         AND s.Active='Y' AND c.Active='Y'",
                     array()
                 );
                 $this->tpl_data['conflicts_site'] = 'Sites: all';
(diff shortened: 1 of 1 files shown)


driusan:
@me-evans Would you be able to test that these numbers are more reliable than the old query?

me-evans:
@driusan - sure, no problem.  Just need PR to be merged to an IBIS dev to be able to compare numbers with our prod.  

@sruthymathew123 - Would you mind merging this PR to your VM (if dev up to date so that total conflicts in the module matches total on production)?  Please let me know if yes, then I can access and complete testing requested by Dave. 

Thanks!

PapillonMcGill:
@driusan I check the queries and they do use the same tables and joins. The WHERE clause are build differently but the result should be pretty close.

davidblader:
image
this is apparently passing travis but without updating on GH itself

PapillonMcGill:
restarted Travis. migth have been a bug during data transmission from Travis to GH.

me-evans:
@driusan @davidblader  - tested PR on Sruthy's dev with IBIS numbers, and there is no difference in the conflict totals displayed on the dashboard, with or without PR.  What was supposed to change in terms of the criteria considered for the dashboard totals after merging the PR?
I've identified what the discrepancies are between dashboard & module, though:    

1. The total in the module is incorrect for one site, not the dashboard total.  The module counts/displays two conflicts for a candidate whose Active flag has been set to No, and whose data should no longer be anywhere on the front end.  The module should be changed LORIS-wide to exclude these candidates - made redmine ticket  [#14872](https://redmine.cbrain.mcgill.ca/issues/14872) for this.

2. The total conflicts displayed on the dashboard when user has "Across all sites" permission automatically excludes conflicts for DCC candidates, which do appear in the module.  I'm guessing this was by design as it makes sense to me that users/coordinators would not want to see DCC conflicts included in the total for the project in this "at a glance" section.  

_(NOTE: The only thing that's a bit strange is that a user that has DCC as their (or one of their) site affiliation(s) will only see DCC conflicts in the dashboard total if they do not have the "across all sites" permission. I don't think this needs to be changed because DCC staff who use the dashboard as a reference (ex. coordinator) would have the "across all sites" permission and wouldn't want to see DCC conflicts included in the total. However, it could potentially cause some confusion for a DCC user testing permissions/multi-site that isn't aware of this exclusion and flags the discrepancy with the module totals. Maybe adding a small note in the My Tasks section when the permission is selected to indicate DCC not included could address this?)_

driusan:
Thanks @me-evans. It looks like @sruthymathew123 already sent #3792 for the first point, and I guess there's nothing to do for the second point. Maybe we should document it in the test plan?

PR #3776, opened 2018-06-28
[20.0][Document Repository] Testplan modifications

um4r12:
Modifies the Testplan by updating the permission, fixing typos, updating test for deleting files. 

[Redmine ticket 14637](https://redmine.cbrain.mcgill.ca/issues/14637)

### DIFF
--- a/modules/document_repository/test/TestPlan
+++ b/modules/document_repository/test/TestPlan
@@ -22,7 +22,7 @@
 11. Add two files with the same name but with different contents in two different categories. Verify that, when 
    downloaded, their contents are different.
    [Manual Testing]
-12. Edit a file in the repository: check that “Last modified on” date is updated.
+12. Edit a file in the repository: check that “Date Uploaded” date is updated.
     [Manual Testing]
 13. Upload a file using another user and then update it. Check that “Last updated by” field is updated.
     [Manual Testing]
@@ -38,7 +38,7 @@
        - Addition of a category.
     Also check that the www address contained in the notification email is correct.
     [Manual Testing]
-18. Check that the following settings are in /etc/php5/apache2/php.ini:
+18. To upload large files, check that the following settings are in php.ini (default: /etc/php/{php_version}/apache2/php.ini):
       session.gc_maxlifetime 10800
       max_execution_time  10800
       upload_max_filesize 1020M
--- a/modules/document_repository/test/TestPlan
+++ b/modules/document_repository/test/TestPlan
@@ -1,44 +1,46 @@
-1. User has access to document repository iff he/she has permission “Access all documents” or is the super user.
+1. User has access to document repository if they have the "View and upload files in Document Repository” permission or is the super user.
    [Automation Testing]
 2. Create a category and a sub category.
    [Automation Testing]
 3. Check that the comments for a category are displayed properly. Also check that they can be edited/added.
    [Automation Testing]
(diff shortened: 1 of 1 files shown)


modules/document_repository/test/TestPlan
```diff
@@ -1,44 +1,46 @@
...
    [Manual Testing]
-8. Check that if a category contains special characters (e.g. space, dot or comma) it can be expanded.
+10. Check that if a category contains special characters (e.g. space, dot or comma) it can be expanded.
    [Manual Testing]
-9. Add two files with the same name but with different contents in two different categories. Verify that, when 
+11. Add two files with the same name but with different contents in two different categories. Verify that, when 
    downloaded, their contents are different.
    [Manual Testing]
-10. Edit a file in the repository: check that “Last modified on” date is updated.
+12. Edit a file in the repository: check that “Last modified on” date is updated.
```
um4r12: This column doesn't actually exist. Should this remain in the Testplan? Or should a PR be created to address this?

modules/document_repository/test/TestPlan
```diff
@@ -1,44 +1,46 @@
...
     [Manual Testing]
-11. Upload a file using another user and then update it. Check that “Last updated by” field is updated.
+13. Upload a file using another user and then update it. Check that “Last updated by” field is updated.
```
um4r12: This column doesn't actually exist. Should this remain in the Testplan? Or should a PR be created to address this?

modules/document_repository/test/TestPlan
davidblader: for some reason, the effective equivalent of this column is "Date Uploaded"
i think media uses a similar convention as well ¯\_(ツ)_/¯
for the time being i would just update the test plan to be consistent with the module...

modules/document_repository/test/TestPlan
```diff
@@ -1,44 +1,46 @@
...
     [Manual Testing]
-15. Check that if you have the 'Receive document repository notification' activated you will receive an email each 
+17. Check that if you have the 'Receive document repository notification' activated you will receive an email each 
     time one of the following event occurs:
        - Addition, deletion or modification of a file
        - Addition of a category.
     Also check that the www address contained in the notification email is correct.
     [Manual Testing]
-16. Check that the following settings are in /etc/php5/apache2/php.ini:
+18. Check that the following settings are in /etc/php5/apache2/php.ini:
```
davidblader: this should probably be changed to just be `php.ini`
nobody should be running php < 7

davidblader: also maybe this should be put towards the beginning of the test plan?

modules/document_repository/test/TestPlan
um4r12: Okay, I'll change it for now. In the future, we should consider adding that column or a better solution may be to rename `Data Uploaded` to `Last modified on`. I did check, if you make an edit to the file, the `Data Uploaded` column is modified. However, that isn't really accurate.

modules/document_repository/test/TestPlan
um4r12: For the most part, I don't think the document repository expects large files (mostly documents, I'd assume, as the name suggests), so I've added additional text at the beginning of the last point for uploading large files. 
Because I think its more of an optional change, leaving it at the end seems better. What do you think?

PR #3781, opened 2018-07-04
[Instruments] Rework breadcrumbs

xlecours:
This add timepoint details in breadcrumbs of instruments

Since our instruments are now treated as subtest in the instruments module, the breadcrumbs were only showing the instrument name. This adds Candidate list > Candidate Profile > Timepoint detail as crumbs for easy access the the aforementioned pages.

see: RM#14811

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2545,8 +2545,8 @@ class NDB_BVL_Instrument extends NDB_Page
      */
     public function getBreadcrumbs(): \LORIS\BreadcrumbTrail
     {
-        $visitlabel = $this->getVisitLabel()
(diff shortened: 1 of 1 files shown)


php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -932,7 +932,7 @@ class NDB_BVL_Instrument extends NDB_Page
                 )
             );
         } else {
-            $results = $db->pselect(
+            $timepoint->getCandID();$results = $db->pselect(
```
ZainVirani: Needs new line

ZainVirani: Also not sure why this line is required

xlecours:
@ZainVirani Can you rereviwe or dismiss?

PR #3784, opened 2018-07-04
cleanup

xlecours:
Removing functions that should not be there.

_checkDate is in Utility anyway.
_checkDateTaken is not used in our code base.


driusan:
Needs to be mentioned in release notes, I think some instruments may use them.

driusan:
Seems reasonable if it passes Travis

PR #3786, opened 2018-07-04
[Instrument Manager] Fix missing validation step when SQL tables do not match Linst files

johnsaigle:
## Background

The instrument manager does some validation in instruments installed in LORIS. It checks whether the back-end SQL schema matches that specified in a Loris Instrument (`linst`) file. Linst files are created via the instrument manager and installed by front-end users in the instrument manager or manually in the back-end.

## Issue

From Redmine:

>Here is the test case :
"Take a second sample .linst instrument file, load it via the Instrument Manager.
Then alter the linst file so that it doesn't exactly match the instrument table already loaded in mysql. (type, number of fields, fieldname spelling) Check that function CheckTable performs the appropriate checks (catches these discrepancies) on the instrument builder files and validates them properly - eg "Table Valid" column is valid when instrument fields match up and "Pages Valid" is valid when subpages are valid. [Manual Testing]"

> For a table to be valid, all the rows in the linst fuile must match a row in the table definition. So removing a row in the file wont make the validation fail.

Basically, if the back-end schema contains fields that do not exist in the linst file, flag this as an error in the validation step. Although outlined in the test plan, the Instrument Manager did not do this.

## This improves LORIS by

Implementing the validation that should have been there already based on the test plan.

I also added comments to the code.

## To Test

* Checkout this branch 

* Upload a linst file via the instrument manager (you can use docs/instruments/test_all_fields.linst)

* Verify that it uploads properly and its status is "Appears Valid" under the `Table Valid` column.

* Under `project/instruments/$YOUR_INST.linst`, delete one of the fields. 

* Return to instrument manager and check that the `Table Valid` status displays "Invalid: DB contains values not in linst"

## Redmine

* Bug #14814 filed by @xlecours

### DIFF
--- a/modules/instrument_manager/php/instrument_manager.class.inc
+++ b/modules/instrument_manager/php/instrument_manager.class.inc
@@ -338,15 +338,15 @@ class Instrument_Manager extends \NDB_Menu_Filter
      * checks whether all of the fields defined in the .linst file are present
      * in the SQL schema and vice-versa.  If the table is found to be consistent
      * it is considered valid and a String confirming this will be returned and
-     * eventually output in the front-end table.  Otherwise an error will be 
+     * eventually output in the front-end table.  Otherwise an error will be
      * returned and displayed to the user.
      *
+     * @param string $instname the value of instname
+     *
+     * @return bool
+     *
      * @see Example linst file: docs/instruments/test_all_fields.linst
-    *
-    * @param string $instname the value of instname
-    *
-    * @return bool
-    */
+     */
     function checkTable($instname)
     {
         $factory  = \NDB_Factory::singleton();
@@ -428,19 +428,19 @@ class Instrument_Manager extends \NDB_Menu_Filter
                 break;
             }
         }
-        // For every value in the `Field` column in the SQL schema, ensure a 
-        // matching value exists in the linst file.  This is done to check 
+        // For every value in the `Field` column in the SQL schema, ensure a
+        // matching value exists in the linst file.  This is done to check
         // whether a question has been deleted from the linst file but is still
         // present in the back-end.
         $sqlSelect = "SELECT COLUMN_NAME".
             " FROM information_schema.columns".
             " WHERE TABLE_SCHEMA=:dbname".
             " AND TABLE_NAME=:tablename";
-        $result      = $db->pselect(
+        $result    = $db->pselect(
             $sqlSelect,
             array(
-                'dbname'     => $factory->settings()->dbName(),
-                'tablename'  => $instname,
+             'dbname'    => $factory->settings()->dbName(),
+             'tablename' => $instname,
             )
         );
         // Get all fields present in DB but absent in linst file.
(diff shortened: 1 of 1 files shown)


modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -344,6 +356,36 @@ class Instrument_Manager extends \NDB_Menu_Filter
                 break;
             }
         }
+        // For every value in the `Field` column in the SQL schema, ensure a
+        // matching value exists in the linst file.  This is done to check
+        // whether a question has been deleted from the linst file but is still
+        // present in the back-end.
```
driusan: The point of checking the schema is to ensure that clicking "Save" doesn't break. There's no harm (and it's common) to not delete old columns so that data is still there if needed.

johnsaigle: @xlecours can you comment here? I issued this PR as a response to your Redmine ticket so perhaps you can provide more context.

xlecours:
@driusan @johnsaigle The testplan says: ""Table Valid" column is valid when instrument fields match up". 
I was not aware of the cultural background of that module. If this "feature" (been able to delete column) is supported, it should be documented. Or at least, the testplan needs to be amended. 

Question, could I delete a column then later accidentally use the same question name and corrupt my instrument?

driusan:
If you accidentally re-use the same question name than on the instrument manager side, it would claim that there's an mismatch, and on the data entry side, it would fail in the same way as any other way you might end up with a mismatch between the instrument and the database table. (ie. changing the options for an existing question.) (I can't remember if that update statement fails or silently gets converted to the empty string because MySQL..)

johnsaigle:
So what's the course of action here?

johnsaigle:
@xlecours I made the changes you requested. I'll wait for Dave to comment before proceeding with the TestPlan changes.

driusan:
@xlecours I'm not sure what you mean by "Now that we have a check for missing columns in the lint file that will display a message but break nothing, do you see a harm in that?".. we've always had that, haven't we?

xlecours:
@driusan The current validation check if all columns in the linst file exists in the database table. This PR adds a validation that __check if all columns in the database table exists in the linst file.__ That is what I meant by "missing columns in the lint file"

ridz1208:
@johnsaigle please resolve conflict

johnsaigle:
@ridz1208 Rebase done.

ridz1208:
@ZainVirani please test

ridz1208:
@johnsaigle @xlecours 

why forcing every column in the database to be in the LINST, we dont do it for php, why should we for LINST ??? sometimes its usefull to have additional columns "hidden" from the front end to store some data in that would still be queriable...

johnsaigle:
Just for the record I'm not sure about the engineering details behind this change. I just addressed what was reported in the ticket.

@xlecours will have to resolve the discussion here.

driusan:
I think this should be closed and the test plan should be updated to be more accurate. See @ridz1208's comment.

zaliqarosli:
Failed testing: Front-end not working.

error log gives `[Mon Aug 27 17:45:03.590289 2018] [:error] [pid 1783] [client 132.216.56.36:37950] PHP Parse error:  syntax error, unexpected end of file, expecting function (T_FUNCTION) in /var/www/loris/modules/instrument_manager/php/instrument_manager.class.inc on line 558
`

johnsaigle:
@zaliqarosli Thanks for testing. I'll hold off on patching this until the above conversation concludes

driusan:
Discussion on this hasn't happened in months. It doesn't seem to be going anywhere, so I'm closing it since at least some users have a use case for the current behaviour and it doesn't introduce any bugs.

driusan:
(I just did some calculations and "months" was the wrong plurality. But it's still been over a month.)

ridz1208:
@zaliqarosli keep this one in mind as you work on instruments. we can talk about it during development

PR #3787, opened 2018-07-04
API Dicom download for phatoms: add missing candID

MounaSafiHarab:
The API DICOM download was missing the CandID for the **phantom cases**, so add it in order for the query to get all its needed parameters, and therefore for the download to work.

To test, please try to download a DICOM from a phantom scan before and after this PR change.

Thanks to Mark Pratti for reporting this!

### DIFF
--- a/htdocs/api/v0.0.3-dev/candidates/visits/Dicoms.php
+++ b/htdocs/api/v0.0.3-dev/candidates/visits/Dicoms.php
@@ -92,8 +92,9 @@ class Dicoms extends \Loris\API\Candidates\Candidate\Visit
             ]
         );
         if ($cand_info['Entity_type'] == 'Scanner') {
-            $ID            = ":PVL";
-            $params['PVL'] = $this->VisitLabel;
+            $ID                 = ":PVL";
+            $params['PCandID']  = $this->CandID;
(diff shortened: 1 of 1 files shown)


driusan:
If memory serves the MySQL PDO driver (at least used to?) throw an exception/crash if the prepared statements parameters didn't match up with the array passed (size, order, etc), so needs to be tested..

MounaSafiHarab:
> While I won't block this PR for that, could you remove line #100 will we are here. I don't think $params['PPSCID'] is ever used

@xlecours 
isn't line 100 being used in building the `$ID` on the line just above for a real candidate (i.e. non phantoms), which in turn is then used in the query on line 118?
p.s: I did not invent those lines, they are recycled from Imaging Browser (albeit for a slightly different query here).

xlecours:
@MounaSafiHarab You are right about line 118. Scratch my comment :)

MounaSafiHarab:
@davidblader @cmadjar 
can anyone of you (or both of you) please tets this? since you both do have phantom data and have used the API before.
(@davidblader you need to copy any .tar for any phantom and test that it downloads now and it was not downloading before the change here).

PR #3790, opened 2018-07-09
[data_team_helper] fix filter for Site

kongtiaowang:
This pull request removes useless code and fixes filter for Site.
https://redmine.cbrain.mcgill.ca/issues/14807

### DIFF
--- a/modules/data_team_helper/php/data_team_helper.class.inc
+++ b/modules/data_team_helper/php/data_team_helper.class.inc
@@ -398,15 +398,8 @@ class Data_Team_Helper extends \NDB_Form
         }
         //filter for sites
         if ($site!='All Sites') {
-            if ($site == 'All User Sites') {
-                $user           = \User::singleton();
-                $siteID         = $user->getCenterIDs();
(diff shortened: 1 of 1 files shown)


modules/data_team_helper/php/data_team_helper.class.inc
```diff
@@ -398,15 +398,8 @@ class Data_Team_Helper extends \NDB_Form
...
         if ($site!='All Sites') {
-            if ($site == 'All User Sites') {
-                $user           = \User::singleton();
-                $siteID         = $user->getCenterIDs();
-                $where         .= " AND FIND_IN_SET(psc.CenterID, :CID)";
-                $qparams['CID'] = implode(',', $siteID);
-            } else {
                 $where         .= " AND s.CenterID = :CID";
                 $qparams['CID'] = $site;
-            }
```
sruthymathew123: @kongtiaowang I think this check is required to return User sites. Removing this will introduce permission issues.

kongtiaowang: @sruthymathew123 you are right.

PR #3791, opened 2018-07-09
[Instruments] 404 

xlecours:
This fix a bug where instruments could not be loaded (404 Not Found)
Introduce in #3768 , every main.php request were redirect to index.php that do not support instruments loading. 

In htdocs/.htacces the rule for instruments was redirecting to main.php with a [L|last] flag assuming there would be no more rewriting. reading http://httpd.apache.org/docs/current/rewrite/flags.html#flag_l there is a mention that while no more rule will apply for that url, the redirected url (main.php in this case) will still be run through the rewriting process.

Using the [END] flag will exit/stop the rewriting completely\

see: RM#14848

### DIFF
--- a/htdocs/.htaccess
+++ b/htdocs/.htaccess
@@ -6,8 +6,8 @@
 RewriteRule ^([a-zA-Z_-]+)/ajax/([a-zA-Z0-9_.-]+)$ AjaxHelper.php?Module=$1&script=$2 [QSA,L]
 
 # Instruments and reliability still use main.php, for now
-RewriteRule ^instruments/([a-zA-Z0-9_-]+)/$ main.php?test_name=$1 [QSA,L]
-RewriteRule ^instruments/([a-zA-Z0-9_-]+)/([a-zA-Z0-9_.-]+)/$ main.php?test_name=$1&subtest=$2 [QSA,L]
+RewriteRule ^instruments/([a-zA-Z0-9_-]+)/$ main.php?test_name=$1 [QSA,END]
+RewriteRule ^instruments/([a-zA-Z0-9_-]+)/([a-zA-Z0-9_.-]+)/$ main.php?test_name=$1&subtest=$2 [QSA,END]
 
 RewriteRule ^([a-zA-Z0-9_-]+)reliability([a-zA-Z0-9_-]*) main.php?test_name=$1reliability$2 [QSA,L]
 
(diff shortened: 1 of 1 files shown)


xlecours:
@davidblader Can you test this to make sure there is no conflicts with https://github.com/aces/Loris/pull/3768 ?
Thanks

xlecours:
@kongtiaowang Can we add an integration test for instrument loading process?

htdocs/.htaccess
```diff
@@ -6,8 +6,8 @@
 RewriteRule ^([a-zA-Z_-]+)/ajax/([a-zA-Z0-9_.-]+)$ AjaxHelper.php?Module=$1&script=$2 [QSA,L]
 
 # Instruments and reliability still use main.php, for now
-RewriteRule ^instruments/([a-zA-Z0-9_-]+)/$ main.php?test_name=$1 [QSA,L]
-RewriteRule ^instruments/([a-zA-Z0-9_-]+)/([a-zA-Z0-9_.-]+)/$ main.php?test_name=$1&subtest=$2 [QSA,L]
+RewriteRule ^instruments/([a-zA-Z0-9_-]+)/$ main.php?test_name=$1 [QSA,END]
+RewriteRule ^instruments/([a-zA-Z0-9_-]+)/([a-zA-Z0-9_.-]+)/$ main.php?test_name=$1&subtest=$2 [QSA,END]
 
 RewriteRule ^([a-zA-Z0-9_-]+)reliability([a-zA-Z0-9_-]*) main.php?test_name=$1reliability$2 [QSA,L]
```
davidblader: would it make sense to change `L` here as well?

xlecours: @davidblader I think END supersede L but I will check again

xlecours: @davidblader END is almighty
from doc:
> An alternative flag, [END], can be used to terminate not only the current round of rewrite processing but prevent any subsequent rewrite processing from occurring in per-directory (htaccess) context.

davidblader: right, i am just wondering if the same flags need to be set for reliability, since i don't have any way to test that
i think we should get an IBIS developer ( @sruthymathew123 , @llevitis , @suzanne-lee ) to make sure everything in reliability is working as well

driusan: Should probably be done for AjaxHelper too

driusan:
Added Add To Release notes because we'll have to document the minimum version of Apache, since now it needs to be recent enough to support the `[END]` tag.

PR #3792, opened 2018-07-09
[Conflict Resolver] Enabling Active Candidate and Active Session Check in Query

sruthymathew123:
[Redmine 14872](https://redmine.cbrain.mcgill.ca/issues/14872)

Conflict Resolver is pulling results of Inactive Candidates and Inactive Sessions. This leads to one of the reason for IBIS where our conflicts count in dashboard doesn't match with the count in conflict resolver.

### DIFF
--- a/modules/conflict_resolver/php/conflict_resolver.class.inc
+++ b/modules/conflict_resolver/php/conflict_resolver.class.inc
@@ -236,7 +236,7 @@ class Conflict_Resolver extends \NDB_Menu_Filter_Form
             LEFT JOIN session ON (flag.SessionID=session.ID)
             LEFT JOIN candidate ON (candidate.CandID=session.CandID)
             LEFT JOIN Project ON (candidate.ProjectID=Project.ProjectID )
-            WHERE 1=1 AND candidate.Active ='Y'";
+            WHERE candidate.Active ='Y'";
 
         $user = \User::singleton();
         if (!$user->hasPermission('access_all_profiles')) {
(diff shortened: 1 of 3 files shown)


modules/conflict_resolver/php/conflict_resolver.class.inc
```diff
@@ -236,7 +236,7 @@ class Conflict_Resolver extends \NDB_Menu_Filter_Form
             LEFT JOIN session ON (flag.SessionID=session.ID)
             LEFT JOIN candidate ON (candidate.CandID=session.CandID)
             LEFT JOIN Project ON (candidate.ProjectID=Project.ProjectID )
-            WHERE 1=1";
+            WHERE 1=1 AND candidate.Active ='Y'";
```
driusan: You can take out the 1=1. That's just there to ensure that if filters add "AND" clauses it doesn't become an SQL syntax error but is unnecessary if there's already a WHERE clause in the query.

modules/conflict_resolver/php/resolved_conflicts.class.inc
```diff
@@ -84,7 +84,7 @@ class Resolved_Conflicts extends \NDB_Menu_Filter
             LEFT JOIN session ON (flag.SessionID=session.ID)
             LEFT JOIN candidate ON (candidate.CandID=session.CandID)
             LEFT JOIN Project ON (candidate.ProjectID=Project.ProjectID )
-            WHERE 1=1";
+            WHERE 1=1 AND candidate.Active ='Y'";
```
driusan: Same here

sruthymathew123:
@driusan Done the requested changes.

xlecours:
@sruthymathew123 The dashboard also filter on session.Active = 'Y' 
should this query do it as well?

sruthymathew123:
@xlecours Good question. I think we have to consider that also.

kongtiaowang:
@sruthymathew123 
You have changed the query conditon,but the testing data doesn't update.
So put "$this->markTestSkipped('rewrite this part later')" into the failed cases.

PR #3796, opened 2018-07-10
[Help Editor] Reactifying module

um4r12:
This pull request `reactifies the help editor module`. The table should now automatically filter out rows based on the `Help Topic` and `Search Keyword`.

To test:
1. Navigate to the Help Editor.
2. Search for various words and see if the table is populated with correct rows.
3. Try using a search combination of both the `Help Topic` and `Search Keyword` fields.

Worked on with: @colezweber @jesscall @miebeers


kongtiaowang:
@um4r12 you have changed the UI. So the test function should be modified also.

modules/help_editor/jsx/help_editor.js
```diff
@@ -0,0 +1,135 @@
...
+      );
+    }
+
+    return (
+      <div>
+        <FilterForm
+          Module="help_editor"
+          name="help_filter"
+          id="help_filter"
+          ref="helpFilter"
```
davidblader: no string refs plz
see example: https://github.com/aces/Loris/pull/3807/files

modules/help_editor/jsx/help_editor.js
```diff
@@ -0,0 +1,135 @@
...
+window.onload = function() {
+  var dataURL = loris.BaseURL + "/help_editor/?format=json";
+  var helpEditor = (
+    <HelpEditor
+      Module="help_editor"
+      DataURL={dataURL}
+    />
+  );
+
+  // Create a wrapper div in which react component will be loaded
```
davidblader: any reason why to not just render directly into `lorisworkspace`?

modules/help_editor/templates/menu_help_editor.tpl
```diff
@@ -1,54 +0,0 @@
-<script type="text/javascript" src="{$baseurl}/js/filterControl.js"></script>
```
davidblader: you can remove the entire file. just add `public $skipTemplate = true` to the PHP class

webpack.config.js
```diff
@@ -46,7 +46,7 @@ var config = [{
     './modules/conflict_resolver/js/resolved_conflicts_columnFormatter.js': './modules/conflict_resolver/jsx/resolved_conflicts_columnFormatter.js',
     './modules/conflict_resolver/js/unresolved_columnFormatter.js': './modules/conflict_resolver/jsx/unresolved_columnFormatter.js',
     './modules/examiner/js/columnFormatter.js': './modules/examiner/jsx/columnFormatter.js',
-    './modules/help_editor/js/columnFormatter.js': './modules/help_editor/jsx/columnFormatter.js',
```
davidblader: you shouldn't remove the columnFormatter entry if you're keeping it in a separate file

HenriRabalais: I think it was discussed that the new standard is that columnFormatter should be a function within the parent component.

modules/help_editor/jsx/help_editor.js
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Help Editor Archive Page.
+ *
+ * Serves as an entry-point to the module, rendering the whole react
+ * component page on load.
+ *
+ * Renders Help Editor main page consisting of FilterTable and
+ * DataTable components.
+ *
+ * @author Alex Ilea
```
HenriRabalais: Since you made this file, I think you can author it ;). Or you put can put `Loris Team`.

modules/help_editor/jsx/help_editor.js
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * */
+class HelpEditor extends React.Component {
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isLoaded: false,
+      filter: {}
```
HenriRabalais: A state of `Data: {}` should be placed here, purely for documentation purposes. On that note, this may be a good time to change `Data` to `data` since it breaks lowerCamelCase convention.

modules/help_editor/jsx/help_editor.js
```diff
@@ -0,0 +1,135 @@
...
+   * Retrive data from the provided URL and save it in state
+   * Additionaly add hiddenHeaders to global loris vairable
+   * for easy access by columnFormatter.
+   */
+  fetchData() {
+    $.ajax(this.props.DataURL, {
+      method: "GET",
+      dataType: 'json',
+      success: function(data) {
+        loris.hiddenHeaders = data.hiddenHeaders ? data.hiddenHeaders : [];
```
HenriRabalais: I'm pretty sure @driusan had mentioned it at the react workshop, but I don't think there's any reason to be using the `loris` global variable here.

modules/help_editor/php/help_editor.class.inc
```diff
@@ -174,7 +131,7 @@ class Help_Editor extends \NDB_Menu_Filter
         return array_merge(
             $deps,
             array(
-             $baseURL . "/help_editor/js/help_editor_helper.js",
+             $baseURL . "/help_editor/js/help_editor.js",
              $baseURL . "/help_editor/js/columnFormatter.js",
```
HenriRabalais: Regardless of whether `columnFormatter` is in its own file or not, I don't think it needs to be included here because it's imported by your helpEditor.jsx.

modules/help_editor/jsx/help_editor.js
driusan: I had the same question during the React workshop.. I had looked into it afterwards and.. no, there was no reason. It works just as well when you render directly. I think people are just copying a module that does this for no reason and basing it on that.

modules/help_editor/jsx/help_editor.js
ZainVirani: what should it be replaced with @davidblader ?

davidblader: a callback ref similar to: https://github.com/aces/Loris/pull/3807/files

modules/help_editor/jsx/help_editor.js
HenriRabalais: Clarification: 
When things were starting to be reactified, hidden headers were set using the `loris` global variable. The same one that you use to get the baseURL (`loris.baseURL`). The reason for this is that the `columnFormatter` function was in a separate file and needed to access the `hiddenHeaders`. However, it's completely unnecessary that the `formatColumn` function be in a separate file. Therefore, on this line in particular, there's no need to set the value to the `loris` variable. I would suggest moving the `formatColumn` file into this function, and setting `hiddenHeaders` as a state.

zaliqarosli:
needs to pass travis `modules/help_editor/php/help_editor.class.inc:108 PhanParamSignatureMismatch Declaration of function toJSON() : \LORIS\help_editor\a should be compatible with function toJSON() : string defined in php/libraries/NDB_Menu_Filter.class.inc:778`

zaliqarosli:
testing: front-end still fails for me. i get an error for 'indexOf' involving `formatColumn`

modules/help_editor/jsx/help_editor.js
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * Modify behaviour of specified column cells in the Data Table component
+ * @param {string} column - column name
+ * @param {string} cell - cell content
+ * @param {arrray} rowData - array of cell contents for a specific row
+ * @param {arrray} rowHeaders - array of table headers (column names)
+ * @return {*} a formated table cell for a given column
+ */
+function formatColumn(column, cell, rowData, rowHeaders) {
+  if (loris.hiddenHeaders.indexOf(column) > -1) {
```
zaliqarosli: ahh found it. we can remove the loris hiddenheaders

modules/help_editor/jsx/help_editor.js
```diff
@@ -24,85 +25,122 @@ class HelpEditor extends React.Component {
...
     this.filter.clearFilter();
   }
 
+  /**
+   * Modify behaviour of specified column cells in the Data Table component
+   *
+   * @param {string} column - column name
+   * @param {string} cell - cell content
+   * @param {arrray} rowData - array of cell contents for a specific row
+   * @param {arrray} rowHeaders - array of table headers (column names)
```
driusan: Is this for pirate tables?

modules/help_editor/jsx/help_editor.js
```diff
@@ -0,0 +1,168 @@
+import FilterForm from 'FilterForm';
+import Loader from 'Loader';
+
+/**
+ * Help Editor Archive Page.
+ *
+ * Serves as an entry-point to the module, rendering the whole react
+ * component page on load.
+ *
+ * Renders Help Editor main page consisting of FilterTable and
```
zaliqarosli: comment update to 'FilterForm and StaticDataTable components'

modules/help_editor/jsx/help_editor.js
zaliqarosli: i am also wondering :p

modules/help_editor/jsx/help_editor.js
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+   * Modify behaviour of specified column cells in the Data Table component
+   *
+   * @param {string} column - column name
+   * @param {string} cell - cell content
+   * @param {arrray} rowData - array of cell contents for a specific row
+   * @param {arrray} rowHeaders - array of table headers (column names)
+   *
+   * @return {*} a formated table cell for a given column
```
zaliqarosli: formatted

modules/help_editor/jsx/help_editor.js
```diff
@@ -0,0 +1,168 @@
...
+    return <td>{cell}</td>;
+  }
+
+  render() {
+    // Waiting for async data to load
+    if (!this.state.isLoaded) {
+      return <Loader/>
+    }
+
+    //TODO: refs should be deprecated in future refactoring.
```
zaliqarosli: is this still relevant?

modules/help_editor/jsx/help_editor.js
```diff
@@ -0,0 +1,168 @@
...
+    const filterRef = function(f) {
```
zaliqarosli: this could maybe be put in the constructor as it doesn't need to be re-rendered?

zaliqarosli: it wouldn't have to be set as a state in order to be accessed within the render function. from this document (https://reactjs.org/docs/refs-and-the-dom.html#callback-refs), the ref call back method that you are using here is set within the constructor. i have also implemented in my PR (https://github.com/aces/Loris/pull/3924) on line (https://github.com/aces/Loris/blob/46bce3d252cfd3a153741f5a9e196794ee22fbdd/modules/examiner/jsx/examiner.js#L38). such a function doesn't belonging in the rendering of the DOM

modules/help_editor/jsx/help_editor.js
zaliqarosli: my understanding is that string refs are being deprecated, but callback refs are fine, which is what you're introducing here. so the todo seems to be resolved here to me. @davidblader what do you think?

davidblader: you are both correct. removing the clearFilter refs is a longterm goal since they are not really doing anything that would require a ref, but in the mean time the callback pattern should be used since the string format is prone to be dropped for any upcoming React release. 
i think the comment can be left as is

modules/help_editor/jsx/help_editor.js
HenriRabalais: Hmm, well based on React's own guidelines for how refs should be used (https://reactjs.org/docs/refs-and-the-dom.html#when-to-use-refs), we shouldn't be using callback refs in this case at all. I didn't introduce this line of code and I rather not move it around, so I think it's best to simply refactor the code now.

HenriRabalais:
@zaliqarosli @davidblader please re-review!

zaliqarosli:
on testing:

- mac : filtering works perfectly, but frozen column headers becomes sticky on scrolling 
screenshot 2018-10-15 13 23 08

- ubuntu: filtered results stay even if filters empty. clicking on 'Clear Filters' isn't updating the results for me
screenshot from 2018-10-15 13-21-22

HenriRabalais:
@zaliqarosli 
- I'm pretty sure the page shouldn't act differently across OS, but rather across browsers. Which one were you using for each test? 
- I'm pretty sure sticky frozen columns are a feature? Not sure. But regardless that configuration is likely from in the StaticDataTable component.
- I'm not sure why the filter isn't updating and the clearFilter button is not working for you, but I've tested it on both firefox and chrome and it is working for me.
- The auto-fill helper is the jurisdiction of the browser.

zaliqarosli:
@HenriRabalais can't replicate error, so it's probably fine! someone else should test as well

modules/help_editor/jsx/help_editor.js
```diff
@@ -0,0 +1,177 @@
...
+          <ButtonElement
+            label="Clear Filters"
+            type="reset"
+            onUserInput={this.resetFilters}
+          />
+        </FilterForm>
+        <StaticDataTable
+          Data={this.state.data.Data}
+          Headers={this.state.data.Headers}
+          hiddenHeaders={this.state.data.hiddenHeaders}
```
zaliqarosli: not on minor yet lolol

HenriRabalais:
This PR is now being maintained by @HenriRabalais #4006

PR #3797, opened 2018-07-10
[Instrument list] Populate the timepoint tpl_data for modules

xlecours:
This  fix the missing timepoint data in the candidate profile table of the instrument_list module

The ModuleRouteur is now in charge of adding candidate and timepoint objects to the request.

see: RM#

### DIFF
--- a/src/Router/ModuleRouter.php
+++ b/src/Router/ModuleRouter.php
@@ -82,6 +82,21 @@ class ModuleRouter extends PrefixRouter
      */
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
+        $gets = $request->getQueryParams();
+        if (!empty($gets['candID'])) {
+            $request = $request->withAttribute(
+                'Candidate',
+                (\NDB_Factory::singleton())->candidate($gets['candID'])
+            );
+        }
+
+        if (!empty($gets['sessionID'])) {
+            $request = $request->withAttribute(
+                'TimePoint',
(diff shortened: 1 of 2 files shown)


src/Router/ModuleRouter.php
```diff
@@ -82,6 +82,21 @@ public function __construct(\Module $module, string $moduledir)
...
+                (\NDB_Factory::singleton())->candidate($gets['candID'])
+            );
+        }
+
+        if (!empty($gets['sessionID'])) {
+            $request = $request->withAttribute(
+                'TimePoint',
+                \TimePoint::singleton($gets['sessionID'])
+            );
+        }
```
driusan: Does this belong in the module router? Shouldn't it be up to the individual pages/modules whether or not they want to populate these in the request that gets passed to the middleware?

xlecours: like that https://github.com/aces/Loris/pull/3799 ?

xlecours:
Modules need to add the atributes themselves for the candidate profile panel to show data

src/Router/ModuleRouter.php
driusan: Yes, I haven't tested it but that looks better to me. (I don't really think those "magic" panels should be in the page decoration middleware at all, it's a relic from main.php. At the minimum they should be a separate middleware that modules need to explicitly add.)

PR #3805, opened 2018-07-11
[dicom_archive] Fix site filtering

davidblader:
see https://redmine.cbrain.mcgill.ca/issues/14812#change-62314 for more details
Filtering by site in DICOM Archive wasn't doing exact match, so filtering for CenterID 5 would return CenterIDs 15, 25, 675, etc.

test filtering in modules with affected js files


jsx/FilterForm.js
```diff
@@ -132,7 +132,7 @@ class FilterForm extends React.Component {
       } else {         // null and undefined handled here
         filter[key].value = '';
       }
-      filter[key].exactMatch = (type === "SelectElement");
+      filter[key].exactMatch = (type === "SelectElement" || type === "select");
```
driusan: What's the difference between a "SelectElement" and a "select" type?

xlecours: Is there a situation where `type === "SelectElement"` ?

davidblader: not sure, i think this was originally developed to detect the React Form components but idk if that is currently functional anywhere.

davidblader: maybe? i couldn't find any personally but i also don't want to break anything :p

PR #3809, opened 2018-07-11
[Integration Test] How to setup integration test environment 

kongtiaowang:
This pull request adds a quick step for Loris developers to set up the integration test environment.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -92,7 +92,6 @@ You must specify a remote host for XDebug to connect to via the `XDEBUG_REMOTE_H
  sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
  sudo apt-get update
  sudo apt-get -y install docker-ce
- sudo apt-get install docker-engine
  sudo docker run hello-world
  sudo apt install docker-compose
  sudo docker-compose build
--- a/test/README.md
+++ b/test/README.md
@@ -85,3 +85,18 @@ You must specify a remote host for XDebug to connect to via the `XDEBUG_REMOTE_H
 - By default npm will output some irrelevant info when a script returns a non-zero error code, as described [here](https://github.com/npm/npm/issues/8821). To prevent this pass `-s` or `--silent` to `npm run`, e.g. `npm run -s tests:unit`.
 
 - Running the entire integration test suite with XDebug enabled sometimes results in a segmentation fault. This appears to be an issue with XDebug itself.
+
+## Quick install for Developer in Ubuntu
+```
+ curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+ sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+ sudo apt-get update
+ sudo apt-get -y install docker-ce
+ sudo apt-get install docker-engine
+ sudo docker run hello-world
+ sudo apt install docker-compose
+ sudo docker-compose build
+ sudo apt install npm
+ sudo composer install
+ sudo npm run tests:integration
+```
--- a/test/README.md
+++ b/test/README.md
@@ -86,16 +86,3 @@ You must specify a remote host for XDebug to connect to via the `XDEBUG_REMOTE_H
 
 - Running the entire integration test suite with XDebug enabled sometimes results in a segmentation fault. This appears to be an issue with XDebug itself.
 
-## Quick install for Developer in Ubuntu
-```
- curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
- sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
- sudo apt-get update
- sudo apt-get -y install docker-ce
- sudo docker run hello-world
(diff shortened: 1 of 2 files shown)


test/README.md
```diff
@@ -85,3 +85,18 @@ You must specify a remote host for XDebug to connect to via the `XDEBUG_REMOTE_H
 - By default npm will output some irrelevant info when a script returns a non-zero error code, as described [here](https://github.com/npm/npm/issues/8821). To prevent this pass `-s` or `--silent` to `npm run`, e.g. `npm run -s tests:unit`.
 
 - Running the entire integration test suite with XDebug enabled sometimes results in a segmentation fault. This appears to be an issue with XDebug itself.
+
+## Quick install for Developer in Ubuntu
+```
+ curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
```
PapillonMcGill: These step are not the same as in the `requirement` section at the top of this file.
Should have the same instruction at both place.

in the requirement section only install docker-compose while here, install docker-ce, docker-engine and docker-compose

test/README.md
```diff
@@ -85,3 +85,18 @@ You must specify a remote host for XDebug to connect to via the `XDEBUG_REMOTE_H
...
+ sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+ sudo apt-get update
+ sudo apt-get -y install docker-ce
+ sudo apt-get install docker-engine
```
johnsaigle: When I tried running this command, I got:

```
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package docker-engine is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  docker-ce
```

So can we remove this line and just install `docker-ce`?

test/README.md
PapillonMcGill: The requirement section should IMHO reflect the same requirement as what is installed in the quick install section. 
And both need to be up to date with the latest docker version.

test/README.md
```diff
@@ -86,17 +86,3 @@ You must specify a remote host for XDebug to connect to via the `XDEBUG_REMOTE_H
 
 - Running the entire integration test suite with XDebug enabled sometimes results in a segmentation fault. This appears to be an issue with XDebug itself.
 
-## Quick install for Developer in Ubuntu
```
PapillonMcGill: We should put a note for line 11 to use the latest version of docker-compose or at least a recent one
V 1.8.1 is old.

zaliqarosli: @PapillonMcGill has this been resolved? i see `sudo apt install docker-compose` in the code. does this satisfy your comment? please give approving review if yes

test/README.md
```diff
@@ -85,3 +85,4 @@ You must specify a remote host for XDebug to connect to via the `XDEBUG_REMOTE_H
 - By default npm will output some irrelevant info when a script returns a non-zero error code, as described [here](https://github.com/npm/npm/issues/8821). To prevent this pass `-s` or `--silent` to `npm run`, e.g. `npm run -s tests:unit`.
 
 - Running the entire integration test suite with XDebug enabled sometimes results in a segmentation fault. This appears to be an issue with XDebug itself.
+
```
zaliqarosli: @kongtiaowang do changes to this document need to be included in this PR at all now that you created a new separate document??

test/README.md
PapillonMcGill: Nope,
Near the top, the first instruction is to `curl` an (quite) older version of docker. 
It is only in the instruction for developer that instruction is to `sudo apt install docker-compose`

`curl` is ok but should not point to a old version. Should have text to download the latest version from docker site.

zaliqarosli: @PapillonMcGill could you suggest a change directly on here that @kongtiaowang could apply and commit?

PapillonMcGill: -Next, install Docker Compose: 
+Next, install Docker Compose, replacing <version> by the latest version of docker-compose available at https://github.com/docker/compose/releases.

curl -L https://github.com/docker/compose/releases/download/1.8.1/docker-compose-`uname -s`-`uname -m` > ~/docker-compose
+ curl -L https://github.com/docker/compose/releases/download/<version>/docker-compose-`uname -s`-`uname -m` > ~/docker-compose

zaliqarosli: @PapillonMcGill If you go to the 'Files changed' tab, and click on the line(s) you want changed as if to add a comment, you can use the 'Insert a suggestion' feature to make a suggested change that shen can then directly commit on GitHub. This will make changing the code faster and easier!

screenshot 2018-11-20 12 33 35

PapillonMcGill: @zaliqarosli not in this case as those lines haven't been modified nor are adjacent to modified lines by this PR.

zaliqarosli: @PapillonMcGill ahh i see, that's unfortunate :(

test/README.md
kongtiaowang: no, I will delete this.

kongtiaowang: no.

driusan:
Should this be worked into @ridz1208's wiki branch instead of here?

ridz1208:
@driusan given the vague timeline of my work. we can def merge this and i'll put a note to move it to the new location later on

PR #3810, opened 2018-07-11
Missing create language table in the SQL patch

cmadjar:
This pull request adds the missing `CREATE TABLE language` and `INSERT INTO language` from the 19.0_to_20.0 SQL patch. 

Without this, Media crashes completely as it cannot find the language table

See also: https://redmine.cbrain.mcgill.ca/issues/14897

PS: I added 'Add to Release Notes' and 'Document for Release' to remember talking about this new table when drafting the release notes


driusan:
What do we have to say about the table in the release notes?

cmadjar:
@driusan I don't know but since it is a new table, I thought you wanted to mention the new table. I'll let you decide. Do you know who created that new table?

zaliqarosli:
the language table was in the `19.0_To_19.1` upgrade patch  https://github.com/aces/Loris/blob/20.0-release/SQL/Release_patches/19.0_To_19.1_upgrade.sql

zaliqarosli:
putting 'Blocked' so this doesn't go through

cmadjar:
@zaliqarosli thank you for catching that!

@driusan @ridz1208 I will leave the PR open so that you remember to document that people needs to run the patches for 19.0 to 19.1 in case they upgrade straight from 19.0 to 20.0. You can close the PR once the documentation for the release has been updated

ridz1208:
@cmadjar it has always been the case that if your skipping a release you are more than a version behind you need to do each release independently.

not sure if we ever mention it but I will bring it up with dave

PR #3737, opened 2018-06-18
[RELEASE] Update main Readme for 20.0

johnsaigle:

PR #3292, opened 2017-11-17
Remove VERSION file and replace with a build step

driusan:
This removes the VERSION file and instead adds a build step to LORIS to autogenerate it.

The build step currently runs npm compile, composer install, and generates the VERSION file.

Ideally the currently committed compiled javascript files would be removed too, but something about our webpack config breaks if the files aren't already present, so that will need to be investigated separately.

PR #3814, opened 2018-07-12
[RELEASE] Version +

ridz1208:
This PR updates the version file and proposes release procedure for projects.

NOTE for all projects going from 19.0, release procedure must include 19.1 release first then 20.0

*suggested release procedure for NON-DEV*
```
# download the release
# run composer with -- no-dev option
composer install --no-dev
# source release SQL patch
cat SQL/Release_patches/19.1_To_20.0_upgrade.sql | mysql
# run consent php script
php tools/single_use/Normalise_Consent_Data.php
```
*suggested release procedure for DEV*
```
# git checkout latest code (20.0-release,master,bugfix)
# run composer
composer install
# if errors occur you are probably missing phan dependency
# phan requires php7.2  and ast

# installing ast (https://github.com/nikic/php-ast)
sudo apt install php-ast (tested by @johnsaigle )
sudo vi /etc/php/7.2/cli/php.ini

# try composer install again
composer install 

# source release SQL patch
cat SQL/Release_patches/19.1_To_20.0_upgrade.sql | mysql
# run consent php script
php tools/single_use/Normalise_Consent_Data.php
```
*originally `# add extension=ast.so to the CLI php.ini file (use php --ini to find the file to modify)` was included above but it seems to be specific to when pecl install is used instead of apt *

*Clean-up optional*
```
cat SQL/Cleanup_patches/delete_old_consent_tables.sql | mysql
```

additional steps for release: https://redmine.cbrain.mcgill.ca/projects/loris/wiki/LORIS_Release_Procedure


johnsaigle:
@ridz1208 @driusan I tested #3292 and it worked for me without an error. As it includes a handy-dandy Makefile I say we should fast-track it instead of manually touching the VERSION file. It would also simplify the update procedure.

johnsaigle:
@ridz1208 We're using PHP 7.2 officially so let's just tell users to do that instead of 7.1 for phan (in fact we need  >= 7.1.3 iirc)

PapillonMcGill:
More automation. less manual work

johnsaigle:
Finally, I don't know if this is the right place to discuss this but there's also the Update script (https://github.com/aces/Loris/pull/3677) to ease this process. It's slated currently for 20.1 but I think the release would be a logical time to test its operation -- especially since it has already `Passed Manual Tests`!

PapillonMcGill:
@ridz1208 `pecl` is out dated, composer can handle requirement tree better.

driusan:
composer can't install php extensions, it can only validate whether they're present or not.

ridz1208:
@johnsaigle @PapillonMcGill `pecl` is the recommended command by https://github.com/nikic/php-ast

PapillonMcGill:
@driusan Right, I hadn't realized it was for an extension and not a 3rd party package

PapillonMcGill:
@ridz1208 `pecl` can rapidly become a security nightmare. Use it only in the last resort.

ridz1208:
@johnsaigle made some changes, take a look !

I did keep the phan req to 7.1.3 (because that is the truth) but I also mentioned the 7.2 recommended.

johnsaigle:
> sudo apt install php-ast (UNTESTED)

I've tested it lol. It works. 

> add `extension=ast.so` to the CLI php.ini file (use php --ini to find the file to modify)

This step wasn't necessary for me when using `apt`. What does this do? I'm not sure it's needed.

> sudo vi /etc/php/7.2/cli/php.ini

It's a little confusing to specify `7.2` here when above you say `7.1.3`. I think we should just pick one to be consistent. 

My vote is PHP 7.2 because it's consistent with our documentation in the README. See also the review comment (hidden by default) in https://github.com/aces/Loris/pull/3737 where @driusan also mentions we should tell people to install 7.2, not 7.1

ridz1208:
@johnsaigle modified comment

johnsaigle:
> phan requires php2  and ast

@ridz1208 Can you fix this line?

ridz1208:
oops

ridz1208:
@johnsaigle I moved this info to the release draft, your comments are welcome on that one

PR #3782, opened 2018-07-04
[Examiner] fix menu doesn't exist with examiner_view permission.

kongtiaowang:
This pull request fixes the typo in the 0000-00-02-menu.sql
https://redmine.cbrain.mcgill.ca/issues/14847

PR #3817, opened 2018-07-12
[examiner] Fix permission condition for edit examiner

zaliqarosli:
addresses https://redmine.cbrain.mcgill.ca/issues/14900 and changes code from https://github.com/aces/Loris/pull/3745

Need to discuss how to replace error message "You do not have access to this page"

*note: if 'Examiner' doesn't appear in your menu when 'Add and certify examiners' only checked, run the patch from https://github.com/aces/Loris/pull/3782/*

**to test:**
1. on branch aces/20.0-release, give yourself 'Across all sites add and certify examiners' permission and notice that you can see all examiners from all sites, but don't have access to edit/certify them, even those belonging to the site you are also assigned to. this is only true if you are not affiliated with all sites.
2. on my branch, you are now able to edit/certify all examiners from all sites if have 'examiner_multisite' permission
3. if have 'examiner_view' permissions only, you should be able to see all examiners affiliated with your sites, but only able to edit ones who you share all sites with. e.g. if the examiner has Toronto and Montreal, but you only have Montreal, you will be able to see the examiner in the table, but not able to view the edit page once you click on their name
4. quickly check that nothing else changes

### DIFF
--- a/modules/examiner/php/editexaminer.class.inc
+++ b/modules/examiner/php/editexaminer.class.inc
@@ -64,11 +64,11 @@ class EditExaminer extends \NDB_Form
 
             // Access is only granted if user sites are a superset of examiner sites
             // considering that certification changes affect all sites
-            $permitted   = $user->hasPermission('examiner_view')
-                || $user->hasPermission('examiner_multisite');
-            $sameCenters
-                = empty(array_diff($centerIDs, $user->getData('CenterIDs')));
-            return $permitted && $sameCenters;
+            $permitted = $user->hasPermission('examiner_view')
+                && empty(array_diff($centerIDs, $user->getData('CenterIDs')));
+            $permittedAllSites
+                = $user->hasPermission('examiner_multisite');
+            return $permitted || $permittedAllSites;
         }
         return false;
     }
--- a/modules/examiner/php/editexaminer.class.inc
+++ b/modules/examiner/php/editexaminer.class.inc
@@ -67,7 +67,7 @@ class EditExaminer extends \NDB_Form
             $permitted = $user->hasPermission('examiner_view')
                 && empty(array_diff($centerIDs, $user->getData('CenterIDs')));
             $permittedAllSites
-                = $user->hasPermission('examiner_multisite');
+                       = $user->hasPermission('examiner_multisite');
             return $permitted || $permittedAllSites;
         }
         return false;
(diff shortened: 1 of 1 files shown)


zaliqarosli:
Need to discuss error message to replace generic "You do not have access to this page"

johnsaigle:
The code looks good! I'm having trouble testing on my VM. On both branches I always get sent to the "You do not have access to this page" when clicking on an examiner in the examiner table. 

I'll try to debug but for now it would be good if someone else could test the change!

ZainVirani:
@johnsaigle I am also testing.

Issues:
1. I have a user who has the examiner_view permission and cannot see the module in the menu under  'Clinical' (I can still access the module through the URL)
2. When I create new users and assign them the examiner_view_multisite permission, the first time they access the examiner module they themselves are added as an examiner (not sure if this is default functionality, it just seems confusing to me)
3. for point 3 of your test plan, there should be a more clear message as to why the user cannot access that page.

zaliqarosli:
@ZainVirani 
for your first point, you need to apply a patch that is mentioned in the PR description. it's not an issue on 20.0, you just need to apply the patch. for your third point, that is included in the 'Discussion Required' for this PR. and point two, i have no idea!

davidblader:
changes seem to be working for me. will hold off putting Passed Manual Tests until consensus is reached among other testers. 
as far as changing the standard 403 message goes, i agree it's a good idea to give a more descriptive message, the only problem is that there's no real way to do that without some sort of hack to the base classes / throwing the exception within the module itself (neither of which is a very good idea)

some ideas for workarounds:
1. Disable the Edit link in `columnFormatter.js` if the edit condition isn't met
2. Put a disclaimer somewhere around the menu filter detailing why a user may not be able to edit displayed examiners
3. Put a disclaimer in the help text
4. Create a read only version of the `editExaminer` page

i think 4. would probably be the best long term solution but for the interest of time i think 1-3 are fine

jesscall:
Tested it again and everything seems to be working smoothly!

johnsaigle:
@davidblader I agree with 1. being a good solution for now, perhaps also with some hover text saying (something like) "You do not have permission to edit this examiner."

zaliqarosli:
@davidblader thanks for your input! i agree about the problems surrounding changing the error message @johnsaigle i love this solution

driusan:
@zaliqarosli Check on Redmine if there's a ticket for 20.0 about it with the priority "Immediate". I think at this point we should just be pushing for fixing the things that prevent projects from updating after the release and whatever else gets in, gets in.

zaliqarosli:
@driusan if that meant "yes, we can push the need to change edit examiner workflow to after 20.0", then this PR is ready to go

zaliqarosli:
@jesscall could you please add the "Passed Manual Tests" label if this PR did indeed pass your testing. thanks

PR #3818, opened 2018-07-12
[User Accounts] Fix editor perms

xlecours:
This should prevent a user to grant permissions that he do not already own.

### DIFF
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -202,7 +202,11 @@ class Edit_User extends \NDB_Form
         $permIDs = array();
         // store the permission IDs
         if (!empty($values['permID'])) {
-            $permIDs = array_keys($values['permID']);
+            // An editor can only grant permission that he is granted of.
+            $permIDs = array_intersect(
+                array_keys($values['permID']),
+                $editor->getPermissionIDs()
+            );
         }
         unset($values['permID']);
         // store whether to send an email or not
@@ -487,6 +491,7 @@ class Edit_User extends \NDB_Form
                     }
                 }
             }
+
             if (!empty($permIDs)) {
                 $user->addPermissions($permIDs);
             }


modules/user_accounts/php/edit_user.class.inc
```diff
@@ -202,7 +202,11 @@ class Edit_User extends \NDB_Form
         $permIDs = array();
         // store the permission IDs
         if (!empty($values['permID'])) {
-            $permIDs = array_keys($values['permID']);
+            // An editor can only grant permission that he is granted of.
+            $permIDs = array_intersect(
+                array_keys($values['permID']),
+                $editor->getPermissionIDs()
+            );
```
driusan: What happens if you don't have a permission and edit a user who does have the permission to try and grant the a different perm?

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -202,7 +202,11 @@ class Edit_User extends \NDB_Form
         $permIDs = array();
         // store the permission IDs
         if (!empty($values['permID'])) {
-            $permIDs = array_keys($values['permID']);
+            // An editor can only grant permission that he is granted of.
```
ZainVirani: I would change this to "An editor can only grant permissions that they have already been granted"

modules/user_accounts/php/edit_user.class.inc
ZainVirani: It doesn't show up. It only shows the permissions you have been granted.

driusan: But does it delete that permission (which should have been unaffected) from the user?

johnsaigle:
I was able to reproduce the original bug on our test VM but was unable to do so on my personal account. As a result I don't think I'll be able to test this fix.

@ZainVirani if you are available today maybe we can try the fix on your VM?

modules/user_accounts/php/edit_user.class.inc
johnsaigle: @ZainVirani did you test this part? If not please remove Passed Manual Tests

ZainVirani: Oh I see, sorry I misunderstood. Will test now

um4r12:
I tried to reproduce this issue on my VM too, but it doesn't seem like it changes the permissions by adding the parameter `permID[1]=on` from a regular user.

If the regular user does not have access to the `User Account` module, it explicitly denies that request.
However, if the regular user does have access to the `User Account` module, although the request is made, no changes to the permissions are made unless the user providing that request is already able to give the highlander permission. So I can make that request to give highlander permission from the `admin` account, but not from a regular user. 

This probably sounds really confusing, so let me know if you need any clarification.

ZainVirani:
Just tested what @driusan mentioned above, and it does nullify all of the permissions the user cannot see.

johnsaigle:
@ridz1208 Should it be a separate PR then?

driusan:
How can the nullifying other permissions go back to 19.0.1 if the array_intersect is added in this PR?

ridz1208:
@driusan might be caused by something else ?? all I know is:
when you, as the editor, are missing permission X; if you modify someone else's permissions, they will automatically loose X.

@johnsaigle it depends how the two fixes are related, if they both require a revamp of the logic, they could go in the same PR. if they are from 2 different places in the code and can be doe in isolation, 2 PRs are def better.

So as long as we dont make 2 PRs that completely contradict each other... I prefer separation

johnsaigle:
@driusan @driusan It looks like [flush user permissions on every edit.](https://github.com/aces/Loris/pull/3818/files#diff-324943431518887444da99607e70a4e4L442) So if we use an intersection, only the permissions that both users have in common will still be attributed to the user and they will lose any that the editor does not have.

Feels like an easy enough fix.

johnsaigle:
@xlecours I am working on a fix now and have assigned myself to the PR.

johnsaigle:
@ZainVirani @ridz1208 Please test again with my updated code.

um4r12:
@johnsaigle Tested, and it seems to have fixed the issue where it would nullify the permissions. However, there are a bunch of notices and one warning that appears.

`[Mon Jul 23 15:17:04.551986 2018] [php7:warn] [pid 3242] [client 192.168.122.1:45504] PHP Warning:  in_array() expects parameter 2 to be array, null given in /var/www/loris/modules/user_accounts/php/edit_user.class.inc on line 222, referer: https://mkhan-dev.loris.ca/user_accounts/edit_user/?identifier=admin` occurred when I tried to remove and re-add permissions for the admin account from a non-admin account with the edit_user permission.
The notices that appear are mainly `PHP Notice:  Undefined index: name in /var/www/loris/php/libraries/LorisForm.class.inc on line 642, referer: https://mkhan-dev.loris.ca/user_accounts/edit_user/?identifier=muhammad.khan@mcin.ca` which weren't showing up prior to the change.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -199,10 +199,31 @@ class Edit_User extends \NDB_Form
...
+        $newPermissions    = array_keys($values['permID']);
+        $editorPermissions = $editor->getPermissionIDs();
+        $permIDs           = array();
         // store the permission IDs
         if (!empty($values['permID'])) {
-            $permIDs = array_keys($values['permID']);
+            // An editor can only grant permission that he is granted of.
+            $permIDs = array_intersect(
+                array_keys($values['permID']),
+                $editor->getPermissionIDs()
```
um4r12: use `$editorPermissions` if you have a var already retrieving the editor permissions. same thing with `$newPermissions`.

johnsaigle: Done.

johnsaigle:
@um4r12 Thanks for testing. I've incorporated your feedback and added a check for when the arrays could be null.. Could you test again when you have time?

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -199,10 +199,36 @@ class Edit_User extends \NDB_Form
...
         $current_permissionids = $user->getPermissionIDs();
-        $permIDs = array();
+        // All permissions are wiped on update.  $permIDs will contain ALL the
+        // user's permissions after an update.  Therefore its structure can be
+        // defined as:
+        // 1) The new permissions granted by the editor, AND
+        // 2) Certain old permissions that the editor is not permitted to alter
+        // In addition, old permissions should be removed if the editor has
+        // removed them and the editor has permissions to remove them.
+        $newPermissions    = array_keys($values['permID']);
```
ridz1208: `$values['permID']` might generate a notice it should be put in the if statement line 213 and should be instancited to empty array here

ridz1208:
For all those who are trying to check the logic

ridz1208:
@johnsaigle @xlecours just manually tested this for the follwing case
- user has permission that editor has selected -> permission is kept
- user has permission that editor does not have -> permiission is kept
- user does not have permission that editor selected -> premission is added
- user has permission that editor unselected -> permission is removed

all these cases passed with flying colors. the only case I have not tried is adding the permID[X]=on

@um4r12 @ZainVirani can one of you try it see what happens in the cases below:
- user had the permission X and permID[X]=on was added by editor
- user did not have the permission X and permID[X]=on was added by editor

um4r12:
I'm not sure if I tested this properly, however, here is what I did. 
From a non-admin account with a limited set of permissions, I attempted to add `&permID[1]=on` to the url to see if it would enable the superuser permission for another user(ex. testPermissionUser).

The non admin user is only able to see the set of permissions that they are also given, even though the testPermissionUser they are editing may have other permissions. Adding the `permID...` to the url nullifies all the permissions INCLUDING the ones that the non admin user doesn't have, and when I hit save, the testPermissionUser not only loses the set of permissions the non admin user has, but every other permission the non admin user shouldn't be able to edit.

Try this;
1. Backup your user_perm_rel table.
2. Create a new user with `User management` and `Across all sites create and edit users`
3. Login from that user.
4. Navigate to the `admin` user.
5. Append to the end of the url `&permID[1]=on` and press Enter.
6. Scroll to the bottom of the page, and hit `Save`.
7. Good fight admin (ie. Loses all permissions).

@ridz1208 If I try change the ID to an exisiting permission, so `&permID[X]=on`, which the user already has, and hit enter. Only that permission will be given, hitting `Save` will wipe out every other permission held by the user.
The case in which the editor doesn't have the permission, following the above procedure will wipe out all permissions.

ridz1208:
@um4r12 When the `&permID[1]=on` is appended to the url and hit enter, that clears all the permission checks on the front end only. hitting save after actually applies the changes... so this is basically the equivalent of manually unchecking all the entries.

for the issue where permissions not visible are also being removed, I fixed it in my last commit. it was an if statement that was not being entered when it should

please test again

um4r12:
@ridz1208 That seems to have fixed the issue where all other permissions are deleted.
Testing it again using the instructions provided by @johnsaigle:

user had the permission X and permID[X]=on was added by editor
- permission is kept

user did not have the permission X and permID[X]=on was added by editor
- if editor can grant that permission - permission is added
- if editor cannot grant that permisison - permission is not added

johnsaigle:
Applying the Passed Manual Tests based on @um4r12's feedback.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -199,11 +199,38 @@ class Edit_User extends \NDB_Form
...
+        $editorPermissions = $editorPermissions ?? array();
+        // Keep old permissions if the editor is not allowed to edit them.
+        foreach ($current_permissionids as $perm) {
+            if (!in_array($perm, $newPermissions)
+                && !in_array($perm, $editorPermissions)
+            ) {
+                $permIDs[] = $perm;
+            }
+        }
+        unset($values['permID']);
```
johnsaigle: twice?

johnsaigle:
@ridz1208 I moved around your empty array instantiation code. Let me know if it makes sense to you

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -231,7 +228,6 @@ class Edit_User extends \NDB_Form
             }
         }
         unset($values['permID']);
-        unset($values['permID']);
```
ridz1208: @johnsaigle hahaha I knew I pressed ctrl + D somewhere !!!

ridz1208:
@cmadjar makes a valid point

modules/user_accounts/test/TestPlan.md
```diff
@@ -59,11 +59,18 @@ When creating or editing a user: (subtest: `edit_user`)
     changes. 
 25. When editing an existing user: Clicking on the 'User Account' breadcrumb takes you to the User Account page
     without saving any changes to the user profile.
+    
+### Testing permissions:
+
+26. r
+1. r
+1. r
+1. r
```
cmadjar: @ridz1208 The Testing permission points look weird.

ridz1208: ha

ridz1208: ha

modules/user_accounts/test/TestPlan.md
```diff
@@ -59,18 +59,29 @@ When creating or editing a user: (subtest: `edit_user`)
     changes. 
 25. When editing an existing user: Clicking on the 'User Account' breadcrumb takes you to the User Account page
     without saving any changes to the user profile.
+    
+### Testing permissions:
+*Log in as a simple user (non-admin) and ensure the following. Define `editee` as the user being edited and `editor` as the logged in user making the changes.*
+
+26. Editor can only see and manipulate permissions for editee if and only if editor has the permissions in question
+1. When page is saved with any set of selected permissions, on reload all selected permissions are kept and all unselected permissions are removed.
+1. If editee has permissions that editor does not have, said permissions are not impacted by manipulating the editees's visible permissions (requires admin user login in a side window to be able to view all permissions for editee as editor will only be able to view permissions they own)
```
cmadjar: remove `s` from second iteration of word `editee` in the sentence: ..... impacted by manipulating the *editee's* visible permissions ...

modules/user_accounts/test/TestPlan.md
```diff
@@ -59,18 +59,29 @@ When creating or editing a user: (subtest: `edit_user`)
...
+29. See security testing below
+
+##### Security testing:
+ - Try manipulating the POST request from the browser and to add a permission to the editee. Make sure that the editee gets the permission if it is within the set of permissions of the editor Make sure the editee does not get the permission if it is not within the set of permissions of the editor. [See this PR](https://github.com/aces/Loris/pull/3818#issuecomment-408882440) for more details.
```
cmadjar: Try manipulating the POST request from the *browser to add* a permission to the editee. (removal of the `and` in the sentence)

Missing dot after ... within the set of permissions of the editor*.* ...

ridz1208:
@cmadjar changes made

driusan:
@ridz1208 It seems like that venn diagram would be significantly easier/more reliable to verify with a unit test than with manual testing.

ridz1208:
@driusan your are correct, you should def implement it

johnsaigle:
@ridz1208 Maybe save the diagram and attach it to a Redmine ticket for someone to eventually write it -- unless @driusan is feeling keen! 😄

ridz1208:
@johnsaigle done https://redmine.cbrain.mcgill.ca/issues/15070

modules/user_accounts/test/TestPlan.md
```diff
@@ -59,18 +59,29 @@ When creating or editing a user: (subtest: `edit_user`)
...
-34. Verify that the notifications are sent when the trigger event occurs. (NB the user triggering the event will not get the email, only other registered users are notified)
-35. Clicking on the 'User Account' breadcrumb takes you to the User Account page without saving any changes. If you do not have access to the user account module, the system should tell you so.
+1. Verify that all notification checkboxes match what the project has enabled in the `notification_modules_services_rel`. See WIKI for more details https://github.com/aces/Loris/wiki/Notification-system.
+1. Verify that checkboxes available respect the permission restrictions set in the `notification_modules_perm_rel`. See WIKI for more details https://github.com/aces/Loris/wiki/Notification-system.
+1. Verify that the notifications are sent when the trigger event occurs. (NB the user triggering the event will not get the email, only other registered users are notified)
+39. Clicking on the 'User Account' breadcrumb takes you to the User Account page without saving any changes. If you do not have access to the user account module, the system should tell you so.
```
driusan: Same here

modules/user_accounts/test/TestPlan.md
```diff
@@ -59,18 +59,29 @@ When creating or editing a user: (subtest: `edit_user`)
...
-28. Check that if password and confirmed password do not match you get an error.
-29. Check that saving fails if any field is left blank (except password).
-30. Check that if you do not enter an email address that is syntactically valid you get an error.
-31. Modify any field on the page and save, and go to the User Account page. Check that the modifications are
+30. Check that all users (even those with NO permissions) have access to the My Preferences page.
+1. Change the user’s password.  Check that the password rules are enforced.
+1. Check that if password and confirmed password do not match you get an error.
+1. Check that saving fails if any field is left blank (except password).
+1. Check that if you do not enter an email address that is syntactically valid you get an error.
+1. Modify any field on the page and save, and go to the User Account page. Check that the modifications are
```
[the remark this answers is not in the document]
ridz1208: numbers get auto generated in markdown and it avoids us having to manually change 40 numbers everytime we wanna add 1... same below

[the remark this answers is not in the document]
driusan: Numbers aren't "autogenerated" from anything, GitHub's rendering of markdown isn't "markdown". The numbers are wrong.

ridz1208:
@driusan done

PR #3819, opened 2018-07-12
[create_timepoint][20.0] Validation for VisitLabel accessing wrong variable therefore breaks on CCNA

jesscall:
This PR addresses the following Redmine issue: https://redmine.cbrain.mcgill.ca/issues/14913

When creating a timepoint in CCNA, the error message 'This visit label is not valid for this subproject.' is always displayed, even for VALID visitLabels.

After looking into this I realized that the validation function `isValidVisitLabel` in the Timepoint class accesses the wrong variable. It gets from config the front-end label of the visit item rather than the actual key used in the backend.
On Loris, it doesn't actually look like there's a problem because both the label and key are the same ('V1' === 'V1'), however in CCNA they are not the same ('Initial Assessment - Screening'  != 'Initial_Assessment_Screening'). 

Please refer to the redmine issue for an example of this, and feel free to come talk to me if it is still unclear.

### DIFF
--- a/php/libraries/TimePoint.class.inc
+++ b/php/libraries/TimePoint.class.inc
@@ -281,7 +281,7 @@ class TimePoint
                     $visitLabelSetting['labelSet']['item']
                 );
                 foreach ($items as $item) {
-                    if ($item['#'] == $visitLabel) {
+                    if ($item['@']['value'] == $visitLabel) {
                         $isValid = true;
                         break;
                     }


ZainVirani:
@zaliqarosli change your VL settings in your config.xml to be non-user generated

zaliqarosli:
hey @jesscall im trying to test, and not seeing a difference. could you provide me with steps to properly test please? i feel like im missing something

ZainVirani:
@zaliqarosli where you define VLs, change the value of one to be different for example instead of 'V1'->'V1' try 'V1'->'V_1' and see if it still works. @jesscall can confirm if this is right but im pretty sure it is

jesscall:
@zaliqarosli on LORIS you will not see a difference unless you make the changes Zain mentioned ^^

driusan:
This change makes me nervous since that weird array format depends on the config.xml, so I'm worried we might be breaking other projects that have a different config setup, but I can't pin my finger on how/why, so I'll merge this since it's passed manual tests and hope we get it in the next round of testing.

PR #3827, opened 2018-07-14
[Router] check in project modules as well

ridz1208:
This lets the code recognize project specific modules (modules not in the `loris/modules`) as loris modules as well allowing them to load instead of getting a 404

### DIFF
--- a/src/Router/BaseRouter.php
+++ b/src/Router/BaseRouter.php
@@ -82,7 +82,9 @@ class BaseRouter extends PrefixRouter implements RequestHandlerInterface
             $components = preg_split("/\/+?/", $path);
             $modulename = $components[0];
         }
-        if (is_dir($this->moduledir . "/" . $modulename)) {
+        if (is_dir($this->moduledir . "/" . $modulename)
+            || is_dir($this->projectdir . "/modules/" . $modulename)
+        ) {
             $uri    = $request->getURI();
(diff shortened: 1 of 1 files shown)


src/Router/BaseRouter.php
```diff
@@ -82,7 +82,9 @@ public function handle(ServerRequestInterface $request) : ResponseInterface
             $components = preg_split("/\/+?/", $path);
             $modulename = $components[0];
         }
-        if (is_dir($this->moduledir . "/" . $modulename)) {
+        if (is_dir($this->moduledir . "/" . $modulename)
+            || is_dir($this->projectdir . "/modules/" . $modulename)
+        ) {
```
driusan: I'd prefer to get rid of the moduledir/projectdir hack and instead having a configurable module search path, but for 20.0 this is probably fine.

driusan:
(Has this been tested?)

ridz1208:
@driusan not tested yet

@sruthymathew123 I think your the best person to test this one on IBIS:
this is necessary when the project has a module that does not exist in loris, i.e. a custom made module. basically testing without this PR the module wont load, with this PR everything is fine.

sruthymathew123:
@driusan @ridz1208 Tested on an IBIS custom module and it worked as expected.

Before

current_20 0

With this PR

rida_pr_fix

PR #3828, opened 2018-07-14
[SRC] add directory to PHPCS list

ridz1208:
This adds the new `src/` directory to the list of checks for the php code sniffer

### DIFF
--- a/src/Http/StringStream.php
+++ b/src/Http/StringStream.php
@@ -280,9 +280,9 @@ class StringStream implements \Psr\Http\Message\StreamInterface, RequestHandlerI
      * @param ServerRequestInterface $request The PSR15 Request being handled
      *
      * @return ResponseInterface A response whose body consists of this string
-
      */
-    public function handle(ServerRequestInterface $request) : ResponseInterface {
+    public function handle(ServerRequestInterface $request) : ResponseInterface
+    {
             return (new \Zend\Diactoros\Response())
                         ->withBody($this);
     }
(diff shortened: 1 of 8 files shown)


ridz1208:
@driusan was this omitted on purpose ? if not could you fill in the missing elements (mostly docs that would be easier for you to fill in)

driusan:
I think it was omitted because the plan was to require PSR2/12 + (line lengths + required doc comments) for the src directory rather than the PEAR based LorisCS.xml standard, but until we have a code sniff that enforces it LorisCS.xml is better than nothing, so I'll try and see if I have time to add these docs before I go on vacation.

driusan:
@ridz1208 I changed this to minor, but it's going to need a rebase now.

driusan:
How would changing the style guidelines break anything? (Other than potentially accidentally introducing bugs while updating them, but that would just be a bug fix..)

ridz1208:
@driusan its the "potentially accidentally introducing bugs" which could affect many modules or even a few edge case modules without being detected before getting to production... its risky if you ask me

and we have had cases where PHPCSing a module went bad !

driusan:
@ridz1208 That's why the plan was to do it progressively be introducing it in the src directory and not do it all at once.

ridz1208:
@driusan I see
that might be a better Idea. will close this and start making little pieces

driusan:
@ridz1208 I think we can keep this. If we start by enforcing LorisCS.xml, it shouldn't be a huge change to a new PHPCS config which enforces more modern standards. I think the only incompatible thing between the two standards that affects us is that the indentation of case statements is different between the two standards, and everything else can be reconciled by going with "enforce whichever has stricter standards"

PR #3829, opened 2018-07-16
[user accounts] Fixing users being always inserted into examiner table upon save

cmadjar:
This PR fixes the issue that whenever a user has the examiner_multisite permission and creates a new user, the new user will always be added to the examiner table even if no examiner options were set. 

This bug was found by @ZainVirani so assigning him to be a reviewer to this new PR.

### DIFF
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -311,19 +311,11 @@ class Edit_User extends \NDB_Form
                             WHERE full_name=:fullName",
                     array('fullName' => $values['Real_name'])
                 );
-//            } elseif (!empty($examinerID)
-//                && ((!empty($ex_radiologist)
-//                && !empty($ex_pending)
-//                && !empty($ex_curr_sites))
-//                || (empty($ex_radiologist)
-//                && empty($ex_pending)
-//                && empty($ex_curr_sites)))
-//            ) {
-                } elseif (!empty($examinerID)
-                  && ((!empty($ex_radiologist)
-                  && !empty($ex_pending)
-                  && !empty($ex_curr_sites))
-                  || empty($ex_curr_sites))
+            } elseif (!empty($examinerID)
+                && ((!empty($ex_radiologist)
+                && !empty($ex_pending)
+                && !empty($ex_curr_sites))
+                || empty($ex_curr_sites))
             ) {
                 // If examiner already exists in the examiner table AND
                 // radiologist, pending and current sites fields all set or
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -311,13 +311,19 @@ class Edit_User extends \NDB_Form
                             WHERE full_name=:fullName",
                     array('fullName' => $values['Real_name'])
                 );
-            } elseif (!empty($examinerID)
-                && ((!empty($ex_radiologist)
-                && !empty($ex_pending)
-                && !empty($ex_curr_sites))
-                || (empty($ex_radiologist)
-                && empty($ex_pending)
-                && empty($ex_curr_sites)))
+//            } elseif (!empty($examinerID)
+//                && ((!empty($ex_radiologist)
+//                && !empty($ex_pending)
+//                && !empty($ex_curr_sites))
+//                || (empty($ex_radiologist)
+//                && empty($ex_pending)
+//                && empty($ex_curr_sites)))
+//            ) {
+                } elseif (!empty($examinerID)
+                  && ((!empty($ex_radiologist)
+                  && !empty($ex_pending)
+                  && !empty($ex_curr_sites))
+                  || empty($ex_curr_sites))
             ) {
                 // If examiner already exists in the examiner table AND
                 // radiologist, pending and current sites fields all set or
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -254,6 +254,7 @@ class Edit_User extends \NDB_Form
         }
         unset($values['CenterIDs']);
         // END multi-site UPDATE
+
         // EXAMINER UPDATE
         if ($editor->hasPermission('examiner_multisite')) {
             // get all fields that are related to examiners
@@ -288,18 +289,14 @@ class Edit_User extends \NDB_Form
             );
 
             // START EXAMINER UPDATE
-            // If examiner not in table add him,
-            // otherwise update the radiologist field and get the current sites
             if (!empty($ex_radiologist)
                 && !empty($ex_pending)
                 && !empty($ex_curr_sites)
                 && empty($examinerID)
             ) {
+                // If examiner not in table and radiologist, pending and current
+                // sites fields set add the examiner to the examiner table
                 $ex_radiologist = $ex_radiologist === 'Y' ? 1 : 0;
-                $values         = \Utility::nullifyEmpty(
-                    $values,
-                    'examiner_radiologist'
-                );
                 $DB->insert(
                     'examiners',
                     array(
@@ -322,11 +319,10 @@ class Edit_User extends \NDB_Form
                 && empty($ex_pending)
                 && empty($ex_curr_sites)))
             ) {
+                // If examiner already exists in the examiner table AND
+                // radiologist, pending and current sites fields all set or
+                // unset update the examiner table with new values
                 $ex_radiologist = $ex_radiologist === 'Y' ? 1 : 0;
-                $values         = \Utility::nullifyEmpty(
-                    $values,
-                    'examiner_radiologist'
-                );
                 $DB->update(
                     'examiners',
                     array(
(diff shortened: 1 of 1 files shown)


johnsaigle:
@cmadjar This requires phpcs

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -293,102 +290,118 @@ class Edit_User extends \NDB_Form
...
+                      && ((!empty($ex_radiologist)
+                          && !empty($ex_pending)
+                          && !empty($ex_curr_sites)
+                         ) || (empty($ex_radiologist)
+                          && empty($ex_pending)
+                          && empty($ex_curr_sites)
+                         ))
+            ) {
+                print_r('other situation');
+                print_r($ex_radiologist);
```
xlecours: Was that for debuging?

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -293,102 +290,118 @@ class Edit_User extends \NDB_Form
...
+            } elseif (!empty($examinerID)
+                      && ((!empty($ex_radiologist)
+                          && !empty($ex_pending)
+                          && !empty($ex_curr_sites)
+                         ) || (empty($ex_radiologist)
+                          && empty($ex_pending)
+                          && empty($ex_curr_sites)
+                         ))
+            ) {
+                print_r('other situation');
```
xlecours: Was that for debuging?

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -293,102 +290,118 @@ class Edit_User extends \NDB_Form
...
             // otherwise update the radiologist field and get the current sites
-            $values = \Utility::nullifyEmpty($values, 'examiner_radiologist');
-            if (empty($examinerID)) {
-
+            if (!empty($ex_radiologist)
+                && !empty($ex_pending)
+                && !empty($ex_curr_sites)
+                && empty($examinerID)
+            ) {
+                print_r("radio, pending, curr site not empty + examiner empty");
```
xlecours: Was that for debuging?

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -293,102 +290,118 @@ class Edit_User extends \NDB_Form
...
                     array(
-                     'full_name'   => $values['Real_name'],
-                     'radiologist' => $ex_radiologist,
-                     'userID'      => $uid,
+                        'full_name'   => $values['Real_name'],
+                        'radiologist' => $ex_radiologist,
+                        'userID'      => $uid,
                     )
                 );
                 $examinerID = $examinerID = $DB->pselectOne(
```
um4r12: Is this supposed to be just `$examinerID = $DB->pselectOne(...)`, instead of `$examinerID = $examinerID = $DB->pselectOne(...)`

modules/user_accounts/php/edit_user.class.inc
cmadjar: oops. Yes. I will remove it. Sorry :S

cmadjar:
@xlecours removed the print_r statements.
@johnsaigle fixed the travis issues hopefully.

jesscall:
Created a user under <User Accounts> with no Examiner affiliations both before and after checking your branch out and it works as intended!

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -293,9 +290,16 @@ class Edit_User extends \NDB_Form
             // START EXAMINER UPDATE
             // If examiner not in table add him,
```
ZainVirani: unrelated but can we change this to "If examiner not in the table add them,"

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -293,9 +290,16 @@ class Edit_User extends \NDB_Form
...
             // otherwise update the radiologist field and get the current sites
-            $values = \Utility::nullifyEmpty($values, 'examiner_radiologist');
-            if (empty($examinerID)) {
-
+            if (!empty($ex_radiologist)
```
ZainVirani: A comment explaining the condition here is a good idea

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -304,14 +308,25 @@ class Edit_User extends \NDB_Form
...
                     "SELECT examinerID
-                 FROM examiners
-                 WHERE full_name=:fullName",
+                            FROM examiners
+                            WHERE full_name=:fullName",
                     array('fullName' => $values['Real_name'])
                 );
-            } else {
-
+            } elseif (!empty($examinerID)
```
ZainVirani: A comment explaining the condition logic here would be good

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -293,9 +290,16 @@ class Edit_User extends \NDB_Form
...
-            $values = \Utility::nullifyEmpty($values, 'examiner_radiologist');
-            if (empty($examinerID)) {
-
+            if (!empty($ex_radiologist)
+                && !empty($ex_pending)
+                && !empty($ex_curr_sites)
+                && empty($examinerID)
+            ) {
+                $ex_radiologist = $ex_radiologist === 'Y' ? 1 : 0;
+                $values         = \Utility::nullifyEmpty(
```
ZainVirani: This function call is useless here, we assigned `$values[examiner_radiologist]` to `$ex_radiologist` earlier and then checked for `!empty($ex_radiologist)`. The call should be made outside of this `if` statement.
I also don't see the point of making this call at all, the value is not used again and it's unset after this code.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -304,14 +308,25 @@ class Edit_User extends \NDB_Form
...
+            } elseif (!empty($examinerID)
+                && ((!empty($ex_radiologist)
+                && !empty($ex_pending)
+                && !empty($ex_curr_sites))
+                || (empty($ex_radiologist)
+                && empty($ex_pending)
+                && empty($ex_curr_sites)))
+            ) {
+                $ex_radiologist = $ex_radiologist === 'Y' ? 1 : 0;
+                $values         = \Utility::nullifyEmpty(
```
ZainVirani: See comment on line 299.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -304,14 +308,25 @@ class Edit_User extends \NDB_Form
                      'userID'      => $uid,
                     )
                 );
-                $examinerID = $examinerID = $DB->pselectOne(
+                $examinerID = $DB->pselectOne(
```
ZainVirani: I don't know why we need to extract the new examinerID here, it doesn't seem to be used.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -327,68 +342,68 @@ class Edit_User extends \NDB_Form
...
-                $result = $DB->pselectRow(
-                    "SELECT epr.centerID
-                     FROM examiners_psc_rel epr 
-                     WHERE epr.examinerID=:eid AND epr.centerID=:cid",
-                    array(
-                     "eid" => $examinerID,
-                     "cid" => $v,
-                    )
-                );
+                foreach ($ex_curr_sites as $k => $v) {
```
ZainVirani: you're missing a `}` before this line (and subsequently have an extra at the end), otherwise new examiners won't have any site affiliation. Previously this code was run regardless of the empty status of $examinerID, and now it is only run when it is set. Setting the examiner status of an existing user for the first time breaks the page as a result.

ZainVirani: This one is the one that sketches me out the most. I know that it does break when this condition is hit, but I didn't check if moving the `}` would fix it. Give it a shot and lmk @cmadjar.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -327,24 +338,23 @@ class Edit_User extends \NDB_Form
...
+                            FROM examiners_psc_rel epr
+                            WHERE examinerID=:eid",
                     array("eid" => $examinerID)
                 );
-
-                //get sites where user is already an examiner at for compare
-                foreach ($prev_sites as $row => $center) {
-                    array_push($ex_prev_sites, $center['centerID']);
-                }
+            }
```
ZainVirani: Sorry, I think this should be on line 350 instead of here. The original file had the next foreach loop inside of the else statement.

ZainVirani: Sorry, I think this should be on line 350 instead of here. The original file had the next foreach loop inside of the else statement.

modules/user_accounts/php/edit_user.class.inc
ZainVirani: @cmadjar I was wrong about this, it is used :+1:

modules/user_accounts/php/edit_user.class.inc
cmadjar: done

cmadjar:
@ZainVirani Good catch! I corrected the misplaced closing curly bracket. Please re-review at your convenience. Thanks a lot!

ridz1208:
@ZainVirani please re-review

@jesscall please re-test on ccna DB (make sure that when examiner sites are checked the examiner entry in the table contains the userID reference, make sure the removal of examiners also works)

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -304,14 +305,24 @@ class Edit_User extends \NDB_Form
...
-            } else {
-
+            } elseif (!empty($examinerID)
+                && ((!empty($ex_radiologist)
+                && !empty($ex_pending)
+                && !empty($ex_curr_sites))
+                || (empty($ex_radiologist)
+                && empty($ex_pending)
+                && empty($ex_curr_sites)))
+            ) {
```
ridz1208: fault in logic above, if examiner ID exists, there is no way the examiner pending and approval can be reset to null (the columns in SQL are not nullable, examiners can not be deleted and there is no code below to handle that case) thus the last segment of this statemnt should be removed.

```
|| (empty($ex_radiologist)
                && empty($ex_pending)
                && empty($ex_curr_sites))
```

The possible cases are:
- `!empty($examinerID) && !empty($ex_pending)  && !empty($ex_radiologist)) &&`
    -  `empty($ex_curr_sites)` -> examiner existed but we just removed all its sites
    - `||`
    - `!empty($ex_curr_sites)` -> examiner existed and we changed its sites
- `empty($examinerID) && !empty($ex_pending)  && !empty($ex_radiologist)) &&`
    - `!empty($ex_curr_sites)` -> examiner is new and we added its sites

ridz1208:
@ZainVirani @zaliqarosli wouldn't mind a double check on the logic, look at what the code is handling below to know what the if statement has to check for

modules/user_accounts/php/edit_user.class.inc
cmadjar: @ridz1208 I changed the logic as you suggested. FYI, the following case is impossible: `!empty($examinerID) && !empty($ex_pending) && !empty($ex_radiologist) && empty($ex_curr_sites)` due to data entry checks. If there are no sites for the examiner, then Radiologist and Pending fields must be emptied out in order for the form to save. That is why I originally added the second "or" with all empty fields. 
```
|| ((empty($ex_radiologist)
                && empty($ex_pending)
                && empty($ex_curr_sites)))
```
But indeed, in case the requirements for data entry change in the form, probably best to simplify the 'or' part of the logic to `|| empty($ex_curr_sites)`

Thanks!

um4r12: This makes sense to me as well. Just to clean it up, this is how it should look like:
```
if (!empty($ex_radiologist)
    && !empty($ex_pending)
    && !empty($ex_curr_sites)
    && empty($examinerID)
    ) {
    	// The case in where examiner is being added for the first time
    	// 
    	...
      }
else if (!empty($examinerID)
		 && !empty($ex_pending) 
		 && !empty($ex_radiologist)) 
		 && (empty($ex_curr_sites) || !empty($ex_curr_sites))
		) {

		 	// The case where the examiner already exists 
		 	// and has either all sites being deleted or the sites are being changed
		  }
```

um4r12: Shouldn't the else if statement be:
```
else if (!empty($examinerID)
		 && !empty($ex_pending) 
		 && !empty($ex_radiologist)) 
		 && (empty($ex_curr_sites) || !empty($ex_curr_sites))
		) {

		 	// The case where the examiner already exists 
		 	// and has either all sites being deleted or the sites are being changed
		  }
```

um4r12:
@cmadjar, addressing your most recent change, shouldn't the else if statement be:
```
else if (!empty($examinerID)
		 && !empty($ex_pending) 
		 && !empty($ex_radiologist)) 
		 && (empty($ex_curr_sites) || !empty($ex_curr_sites))
		) {

		 	// The case where the examiner already exists 
		 	// and has either all sites being deleted or the sites are being changed
		  }
```

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -304,14 +305,30 @@ class Edit_User extends \NDB_Form
...
+//                && !empty($ex_curr_sites))
+//                || (empty($ex_radiologist)
+//                && empty($ex_pending)
+//                && empty($ex_curr_sites)))
+//            ) {
+                } elseif (!empty($examinerID)
+                  && ((!empty($ex_radiologist)
+                  && !empty($ex_pending)
+                  && !empty($ex_curr_sites))
+                  || empty($ex_curr_sites))
```
um4r12: @cmadjar, addressing your most recent change, shouldn't the else if statement be:
```
else if (!empty($examinerID)
		 && !empty($ex_pending) 
		 && !empty($ex_radiologist)) 
		 && (empty($ex_curr_sites) || !empty($ex_curr_sites))
		) {

		 	// The case where the examiner already exists 
		 	// and has either all sites being deleted or the sites are being changed
		  }
```

cmadjar: Hi @um4r12 ,

Thanks for the sugesstion. Actually, I tried that and it does not work as it is impossible to be in the following condition `!empty($examinerID) && !empty($ex_pending) && !empty($ex_radiologist)) && empty($ex_curr_sites)` in the frontend. To remove all the sites from an examiner, you have to unselect radiologist and ex_pending. Currently, only lets you be in either of the following conditions:
- `!empty($examinerID) && !empty($ex_pending) && !empty($ex_radiologist)) && !empty($ex_curr_sites)` 
- `!empty($examinerID) && empty($ex_pending) && empty($ex_radiologist)) && empty($ex_curr_sites)` which I simplified in my last commit by `!empty($examinerID) && empty($ex_curr_sites)`

cmadjar:
@ridz1208 @um4r12 Please re-review. Thanks!

modules/user_accounts/php/edit_user.class.inc
ridz1208: you are correct but that means that there is a bigger fautl in logic in the field validation in the front end. Thats def not within the scope of this PR since its not a bugfix but rather a minor revamp.

This should do for the moment. I will try to test as soon as I can.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -304,14 +305,30 @@ class Edit_User extends \NDB_Form
...
                     "SELECT examinerID
-                 FROM examiners
-                 WHERE full_name=:fullName",
+                            FROM examiners
+                            WHERE full_name=:fullName",
                     array('fullName' => $values['Real_name'])
                 );
-            } else {
-
+//            } elseif (!empty($examinerID)
```
ridz1208: commented code should not be submitted

modules/user_accounts/php/edit_user.class.inc
ridz1208: @cmadjar is correct, its not ideal but is acceptable in the scope of this bugfix. a revap of the logic is def needed to prevent unsetting of all fields and to sync the validation checks to the processing.

ridz1208:
@cmadjar I hope you dont mind, I pushed directly to your branch to fix phpcs

PR #3830, opened 2018-07-16
[Dicom archive] Apply filters dicom achive to mri violations

xlecours:
This reuse the existing function that intercept the onClick of serieUID to send a postr request instead of a get.

This apply the filter in the mri_violation page instead of just populating the fields.

IBIS reaalllllly want this :)


xlecours:
I will remove the href attribute since it is not used anymore

PR #3832, opened 2018-07-16
Fixing Cannot Update Extra Parameters Added in Proband Information

sruthymathew123:
Extra parameters added in Proband Information Tab in Candidate Parameters is not working now; we cannot add/edit/update data to that field.
(For IBIS, we have Proband GUID: Ref on ibis-dev Proband Information tab)

This PR fixes that issue. The extra parameters added in Candidate Info tab is working as expected; so I tried to replicate the same behavior to the extra parameters in Proband Info also which solved the issue.


modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -149,6 +149,44 @@ function editProbandInfoFields($db, $user)
...
     $db->update('candidate', $updateValues, ['CandID' => $candID]);
+    foreach (array_keys($_POST) as $field) {
+        if (!empty($_POST[$field])) {
+            if (substr($field, 0, 4) === 'PTID') {
+                $ptid = substr($field, 4);
+
+                $updateValues = [
+                                 'ParameterTypeID' => $ptid,
+                                 'CandID'          => $candID,
+                                 'Value'           => $_POST[$field],
```
johnsaigle: I'm not very familiar with this module but $_POST[$field] should be sanitized here using [`htmlentities`](https://secure.php.net/manual/en/function.htmlentities.php) unless you know it's explicitly treated in a safe way in the front-end.

johnsaigle: Actually it should be done on all of the POST variables here...

PapillonMcGill: Even if sanitized in the front end by JS, you have to sanitized on the server side as a POST request can be forged.

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -149,6 +149,44 @@ function editProbandInfoFields($db, $user)
...
+                                 'InsertTime'      => time(),
+                                ];
+
+                $result = $db->pselectOne(
+                    'SELECT * from parameter_candidate
```
PapillonMcGill: SELECT * should not be used. If you just want to confirm that no entry exist for from parameter, you should select the primary key of that table (i.e.: SELECT CandID FROM ...)

modules/candidate_parameters/ajax/formHandler.php
johnsaigle: That's not what I meant exactly. I think what you're speaking of is technically _validation_ (sorry to be pedantic) and yes we need to do that both in the front- and back-ends.

Sanitization means treating something safely when it crosses the boundary of one layer of software to another. What I mean here is that someone could enter valid HTML in a POST request (forged or otherwise) which would be saved by the database. When that data leaves the database and is injected into the front-end, we must make sure that it is printed safely (ie. as text) rather than unsafely (ie. as mark up code that executes).

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -136,21 +136,22 @@ function editProbandInfoFields($db, $user)
         header("HTTP/1.1 403 Forbidden");
         exit;
     }
-
-    $candID = $_POST['candID'];
+    //Sanitizing the post data
+    $sanitize = array_map('htmlentities', $_POST);
```
ridz1208: @johnsaigle is this being done anywhere else in this php ?

johnsaigle: I haven't reviewed this module yet but I figured while we're adding new code we can do it correctly here at least.

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -136,21 +136,22 @@ function editProbandInfoFields($db, $user)
...
+    $candID   = $sanitize['candID'];
 
     // Process posted data
-    $gender = isset($_POST['ProbandGender']) ? $_POST['ProbandGender'] : null;
-    $dob    = isset($_POST['ProbandDoB']) ? $_POST['ProbandDoB'] : null;
+    $gender = isset($sanitize['ProbandGender']) ? $sanitize['ProbandGender'] : null;
```
johnsaigle: Could you write these lines instead as `$gender = sanitize['ProbandGender'] ?? null;`?

sruthymathew123: @johnsaigle done!

PR #3834, opened 2018-07-17
[Candidate] Sanitize candidate comments

johnsaigle:
## Background

Before injecting values from the DB into the front-end HTML, they must be sanitized to prevent cross-site scripting (XSS) attacks. To **sanitize on output** of the information - **instead of on _input_** to the DB - is considered best-practice in the context of XSS attack mitigations (as opposed to SQLi prevention, for example).

This PR addresses an issue where such an attack was possible.

## This improves LORIS by...

Patching a XSS attack vector.

## To test

- [ ] See me for information if you don't know how to test this one.

### DIFF
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -144,7 +144,7 @@ class Candidate
             $params = array_map(
                 function ($string) {
                     return htmlentities($string, ENT_QUOTES, 'UTF-8');
-                }, 
+                },
                 $params
             );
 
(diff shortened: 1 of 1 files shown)


php/libraries/Candidate.class.inc
```diff
@@ -139,6 +139,15 @@ class Candidate
                     $params[$row['Description']] = $row['Value'];
```
nicolasbrossard: Nitpick: instead of using the `array_map` call below, wouldn't it be simpler to just sanitize the values of `$params` right when they are created:

`$param[$row['Description']] = htmlentities($row['Value'], ENT_QUOTES, 'UTF-8');`

johnsaigle: Sure that seems simpler. I'll update my fix

johnsaigle: Done!

PR #3836, opened 2018-07-18
[bvl_feedback] Update test plan

zaliqarosli:
addresses https://redmine.cbrain.mcgill.ca/issues/14936.

updates test plan accordingly

### DIFF
--- a/modules/bvl_feedback/test/TestPlan.md
+++ b/modules/bvl_feedback/test/TestPlan.md
@@ -5,7 +5,7 @@
  * Instrument List
  * Any instrument
 [Automation Test]
-2. Remove 'bvl_feedback' permission. Behavioural feedback button should no longer appear.
+2. Remove 'bvl_feedback' permission ("Behavioural QC" on the front-end). Behavioural feedback button should no longer appear.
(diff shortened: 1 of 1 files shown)


modules/bvl_feedback/test/TestPlan.md
```diff
@@ -5,7 +5,7 @@
  * Instrument List
  * Any instrument
 [Automation Test]
-2. Remove 'bvl_feedback' permission. Behavioural feedback button should no longer appear.
+2. Remove 'bvl_feedback' permission ("Behavioural QC" on the front-end). Behavioural feedback button should no longer appear.
```
PapillonMcGill: Remove 'bvl_feedback' permission. Behavioural feedback button ("Behavioural QC") should no longer appear.

IMHO would be better. I still don't like parentheses and quotes at the same time do.

zaliqarosli: @PapillonMcGill I think the point of this PR if to make it clear that the 'bvl_feeback' permission is equivalent to the 'Behavioural QC' permission in the Configuration module

PapillonMcGill: Ok, the new text is better as is specify the location you will encounter the text.

PR #3419, opened 2018-01-17
[JSX] Add errorMessage and onUserBlur props to textbox element in Form.js

HenriRabalais:
This pull request allows the definition of props  "errorMessage" and "onUserBlur" for the TextboxElement. This allows the developer to validate a TextboxElement by triggering an onblur function that will then allow the parent component to provide a value for "errorMessage".

PR #3837, opened 2018-07-18
[JSX] Removed redundant hasError prop from SelectElement in Form.js

HenriRabalais:
Since the SelectElement has an `errorMessage` prop, this makes the `hasError` prop redundant. If an error message is passed to the SelectElement, this implies that the SelectElement has an error, therefore displaying the `errorMessage`.

Refer to the TextboxElement for a similar paradigm: #3419


driusan:
I understand for issues, but what does "Beginner Friendly" mean for a PR?

HenriRabalais:
@driusan I can't seem to find documentation for the new label, but I interpreted it as easy to review and test.

@davidblader makes sense, will do.

johnsaigle:
@driusan I agree that at first I was confused but I think it makes sense to invite the world into our review/test process if they want to join. Since this PR is just a few lines it seems easy to review and test so I think we could maybe use the `Beginner Friendly` label that way 😄

HenriRabalais:
This PR `needs work` before being reviewed and tested.

HenriRabalais:
Deprecated reactified modules need major refactoring before being able to implement this change.

PR #3843, opened 2018-07-23
[API] Fix loading instruments from API

ridz1208:
This fixes Redmine issue 14918 where an API request for instrument data fails with 500 error.

the error was due to a `NULL` value being passed where a string is expected in NDB_Page. the solution was to mimic the existing code above for the `page` variable and setting `commentId=''`

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -169,6 +169,8 @@ class NDB_BVL_Instrument extends NDB_Page
 
         // Ensure 'page' is a string.
         $page = $page ?? '';
+        // Ensure 'commentID' is a string.
+        $commentID = $commentID ?? '';
 
         // Make sure the instrument class has been included/required!
         $factory = NDB_Factory::singleton();


php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -169,6 +169,8 @@ class NDB_BVL_Instrument extends NDB_Page
 
         // Ensure 'page' is a string.
         $page = $page ?? '';
+        // Ensure 'commentID' is a string.
+        $commentID = $commentID ?? '';
```
johnsaigle: @ridz1208 I think this would be better written in the function signature, i.e.

```php
static function factory($instrument, $commentID = '', $page = '', $guarantee_exists = true)
```

I think I wrote the original line on 171 and realized it was silly but have been too busy to fix it. 😅 Let's address it now.

ridz1208: @johnsaigle done

zaliqarosli:
@ridz1208 please add in PR description how to test!

ridz1208:
@zaliqarosli I love it too but it doesnt work. will add description when I fix it

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -163,13 +163,14 @@ class NDB_BVL_Instrument extends NDB_Page
...
      * @access public
      */
-    static function factory($instrument, $commentID, $page, $guarantee_exists = true)
-    {
+    static function factory(
+        $instrument,
+        $commentID = '',
+        $page = '',
+        $guarantee_exists = true
+    ) {
```
HenriRabalais: While you're at it, should this parameters be typed?

ridz1208: valid point but not in a bugfix, risk of breaking silent failures. the additions above only get applied if entry is already null, they dont impose anything

um4r12:
I'm not sure if I'm testing it right, but using the following GET request `candidates/300011/V1/instruments/bmi` I still get a 500 error. In the log this is the error: 
```
[Fri Aug 03 11:09:27.475492 2018] [php7:error] [pid 6497] [client 192.168.122.1:43780] PHP Fatal error:  Uncaught TypeError: Argument 2 passed to NDB_Page::__construct() must be of the type string, null given, called in /var/www/loris/php/libraries/NDB_BVL_Instrument.class.inc on line 200 and defined in /var/www/loris/php/libraries/NDB_Page.class.inc:94\nStack trace:\n#0 /var/www/loris/php/libraries/NDB_BVL_Instrument.class.inc(200): NDB_Page->__construct(Object(Module), NULL, '300011MTL011112...', '300011MTL011112...', 'test_form')\n#1 /var/www/loris/htdocs/api/v0.0.2/candidates/InstrumentData.php(99): NDB_BVL_Instrument::factory('bmi', '300011MTL011112...', NULL, true)\n#2 /var/www/loris/htdocs/api/v0.0.2/candidates/InstrumentData.php(239): Loris\\API\\Candidates\\Candidate\\InstrumentData->__construct('GET', '300011', 'V1', 'bmi', false, false)\n#3 {main}\n  thrown in /var/www/loris/php/libraries/NDB_Page.class.inc on line 94```

ridz1208:
Do u have any overrides on ndb bvl instrument ? Did you tun composer install ? What version of php are you on ?

php/libraries/NDB_BVL_Instrument.class.inc
ridz1208: @johnsaigle I reverted back to this change for the following reason. If in the function call, the `null` argument is explicitely passed to the parameter `$commentID` and/or `$page` then even if you declare it as `$commentID = '', $page = ''`  php still sets it as `null` thus the NULL coalesce is required until we go over all the code calling this function and omit an expicit null or convert them to `''`

ridz1208:
@johnsaigle @zaliqarosli I had to revert the changes to the original commit, please re-review when possible

PR #3845, opened 2018-07-24
Amended the imaging_uploader module test plan.

nicolasbrossard:
Added a test case to validate the incremental search that would not work in certain cases (see [https://redmine.cbrain.mcgill.ca/issues/14925](url))

### DIFF
--- a/modules/imaging_uploader/test/TestPlan.md
+++ b/modules/imaging_uploader/test/TestPlan.md
@@ -9,7 +9,8 @@
 4. Ensure that the Browse tab allows searching based on CandID, PSCID, and/or Visit Label. Ensure that the Visit Label 
    is a dropdown menu containing all the possible visit labels (taken from the Visit_label column in the session table 
    for any Active candidate), and ensure that the CandID and PSCID automatically filter the list of entries in Imaging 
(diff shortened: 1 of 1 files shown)


modules/imaging_uploader/test/TestPlan.md
```diff
@@ -9,7 +9,8 @@
 4. Ensure that the Browse tab allows searching based on CandID, PSCID, and/or Visit Label. Ensure that the Visit Label 
    is a dropdown menu containing all the possible visit labels (taken from the Visit_label column in the session table 
    for any Active candidate), and ensure that the CandID and PSCID automatically filter the list of entries in Imaging 
-   Uploader table on the 'fly', as fields are being filled out. Ensure that 'Clear Filters' button works.  
+   Uploader table on the 'fly', as fields are being filled out. Make sure that this works no matter the page of the 
+   result table you are currently on. Ensure that 'Clear Filters' button works.
```
miebeers: Optional change to make it a little clearer:
"Make sure that this works no matter what page of the uploads table you are currently on" or "Make sure that this works no matter what page of results from the query to the uploads table you are currently on"

PR #3846, opened 2018-07-24
[Tools] Fix score_instrument.php case for scoring all instruments

davidblader:
the usage conditions for score_instrument.php didn't allow for `php score_instrument.php all` since it was failing `!in_array($argv[2], array('all', 'one'))`

this PR simply adds the condition that `$argv[2]` is actually set before exiting the script with the help screen and adds `php score_instrument.php all` as an acceptable use case

to test: run `php score_instrument.php all` and go grab a coffee since this script can take a while if you have a lot of candidates or instruments

### DIFF
--- a/tools/score_instrument.php
+++ b/tools/score_instrument.php
@@ -33,11 +33,15 @@ if (version_compare(phpversion(),'4.3.0','<'))
  * HELP SCREEN
  * display and stop processing if action=help
  */
-if (empty($argv[1]) || $argv[1] == 'help' || !in_array($argv[2], array('all','one')) || ($argv[2]=='one' && (empty($argv[3]) || empty($argv[4])))) {
+if (empty($argv[1]) || $argv[1] == 'help' ||
+    ($argv[2] && !in_array($argv[2], array('all','one'))) ||
+    ($argv[2]=='one' && (empty($argv[3]) || empty($argv[4])))
+    ) {
     fwrite(STDERR, "Usage: \n\n");
     fwrite(STDERR, "score_instrument.php help - displays this msg\n");
     fwrite(STDERR, "score_instrument.php <test_name> one <candID> <sessionID>\n");
     fwrite(STDERR, "score_instrument.php <test_name> all \n");
+    fwrite(STDERR, "score_instrument.php all \n");
     return;
 }
 


colezweber:
I'll test after impending commit

tools/score_instrument.php
```diff
@@ -51,74 +68,67 @@
...
+    $query = "
+        SELECT s.CandID, 
+            s.Visit_label, 
+            s.ID as SessionID, 
+            t.CommentID, 
+            c.PSCID
+        FROM candidate as c, 
+            session as s, 
+            flag as f, 
+            $test_name as t
```
PapillonMcGill: should not use a cross-join but an inner joint

FROM candidate c,
JOIN session s ON c.CandID = s,CandID
JOIN flag f ON s.ID = f.SessionID 
...

If some table might not have actual data, you can use outer join : LEFT JOIN table ON .... or RIGHT JOIN table ON ...

tools/score_instrument.php
```diff
@@ -145,41 +164,74 @@ function log_msg($message) {
...
 
         // print out candidate/session info
-        fwrite(STDERR, "Candidate: ".$record['CandID']."/".$record['Visit_label']."/".$record['SessionID'].":: (".$record['PSCID'].")\n");
-        //fwrite(STDERR, "Candidate: ".$instrument->_dob."/"."Test_name:".$test_name."/". $instrument->_pls3Age."/".$instrument->getDateOfAdministration().":: \n");
+        printOut("Candidate: ".$record['CandID']."/".$record['Visit_label']."/".$record['SessionID'].":: (".$record['PSCID'].")");
 
         // call the score function
-        $oldRecord = $db->pselectRow("SELECT * FROM $test_name WHERE CommentID=:cid", array('cid' => $record['CommentID']));
+        $oldRecord = $DB->pselectRow(
+            "SELECT * FROM $test_name WHERE CommentID=:cid",
```
PapillonMcGill: Should not select **\*** but specific field

tools/score_instrument.php
```diff
@@ -145,41 +164,74 @@ function log_msg($message) {
...
+            array('cid' => $record['CommentID'])
+        );
         $success = $instrument->score();
-        $newRecord = $db->pselectRow("SELECT * FROM $test_name WHERE CommentID=:cid", array('cid' => $record['CommentID']));
+        $newRecord = $DB->pselectRow(
+            "SELECT * FROM $test_name WHERE CommentID=:cid",
```
PapillonMcGill: same here

tools/score_instrument.php
```diff
@@ -145,41 +164,74 @@ function log_msg($message) {
...
+function logMessage($message) {
+    global $logfp;
+    if (!$logfp) {
+        //The log file could not be instanciated
+        //use print instead
+        printOut($message);
+    }
+    $now_array = getdate();
+    $now_string = sprintf("%04d-%02d-%02d %02d:%02d:%02d", $now_array['year'], $now_array['mon'], $now_array['mday'], $now_array['hours'], $now_array['minutes'], $now_array['seconds']);
+    fwrite($logfp, "[$now_string] $message\n");
```
PapillonMcGill: could use strftime(). Would be easier to read and can skip the $now_array

$now = strftime("%Y-%m-%d %H:%M:%S")

tools/score_instrument.php
ridz1208: that might chnage functionality and would then need to be sent to minor, the idea here is to maintain current functionality and just make it runnable

tools/score_instrument.php
ridz1208: functionality change

tools/score_instrument.php
ridz1208: talk to me about risk of unexpected behaviour, is sprintf worst than strftime ?

PapillonMcGill: Not sure about unexpected behaviour for each function but I can say that using strftime() is more specialised and have less case that sprintf(). Combine that the fact that you don't have to call getdate(), create an array for one time use,...

easy guest is that strftime() is not only faster but have less risk of unexpected behaviour.

tools/score_instrument.php
PapillonMcGill: Right on that one, had not seen that the table is a variable that depend on the instrument.

tools/score_instrument.php
driusan: It's already changing the functionality, making them LEFT JOINs with an ON clause is equivalent to cross joins with a WHERE clause (in fact, the WHERE clause is more likely to accidentally filter things out.)

PapillonMcGill: Inner join are more effective than cross joint with the same final result

tools/score_instrument.php
```diff
@@ -17,27 +17,48 @@
  * @package behavioural
  */
 
-//Ensure php version compatability
-//taken from php.net notes
-set_include_path(get_include_path().":../project/libraries:../php/libraries:");
-require_once __DIR__ . "/../vendor/autoload.php";
```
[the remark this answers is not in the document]
ridz1208: its in generic includes

ridz1208:
@PapillonMcGill @driusan Please re-review while keeping in mind that I will be making a more extensive PR to minor to clean-up the tool and put it in the correct subdirectory. this is just intended for immediate use on PROD environments

tools/score_instrument.php
PapillonMcGill: Should be easy to implement quickly (same for line 34-35). Understand that this PR is for a quick fix but it is not for extreme urgency.

ridz1208: it is for CCNA since we need to restore all instruments at once. but is okai we made an override incorporating this fix

@davidblader this PR can be closed now. Since you issued it I'll leave it to you.

PR #3847, opened 2018-07-25
[Battery Manager] New module for adding, activating, and deactivating entries in Test Battery

victoriafoing:
### Brief summary of changes

This PR creates a new module 'Battery Manager' which allows users to browse, add, activate, and deactivate entries in the `test_battery` table.

This module includes the following components:
- A **Browse** tab for searching and deactivating entries in the Test Battery
- An **Add** tab for adding and activating entries in the Test Battery
- AJAX calls for checking for duplicate entries, adding entries, deactivating/activating entries, and getting data for dropwdowns 
- SQL patches for permissions and LorisMenu
- Documentation (README, TestPlan, Help)

Allowing users (e.g. study administrators) to add to and update the `test_battery` table through a front-end interface.

### This resolves issue...

Of having to add to and update the `test_battery` table through the back-end.

### To test this change...

Check out the instructions in the Test Plan

**Note**: The Test Battery table has a lot of NULL values and uses FilterForm.js, which does not properly filter NULL values. Another PR will have to be made to update FilterForm.js.


modules/battery_manager/ajax/add_entry.php
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Retrieve values entered by user
+ *
+ * @return array
+ */
+function getFormData()
+{
+
+     $form_data = array(
+                   'Test_name'    => isset($_POST['instrument']) ? $_POST['instrument'] : null,
```
johnsaigle: These would be better written as 
```php
'Test_name' => $_POST['instrument'] ?? null;`
```


https://www.virendrachandak.com/techtalk/php-7-null-coalesce-operator/

modules/battery_manager/php/battery_manager.class.inc
```diff
@@ -0,0 +1,245 @@
...
+    /**
+     * Check user permissions
+     *
+     * @return bool
+     * @throws ConfigurationException
+     */
+    function _hasAccess()
+    {
+        //create user object
+        $user =& \User::singleton();
```
johnsaigle: I don't think you need the & here

modules/battery_manager/ajax/add_entry.php
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/**
+ * Return duplicate of entry if it exists and null otherwise
+ *
+ * @return Object or null
+ */
+function checkForDuplicate()
+{
+    $db =& Database::singleton();
```
johnsaigle: This should use the namespace syntax i.e. `$db = \Database::singleton()`;

modules/battery_manager/ajax/add_entry.php
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Retrieve values entered by user
+ *
+ * @return array
+ */
+function getFormData()
+{
+
+     $instrument = isset($_POST['instrument']) ? $_POST['instrument'] : null;
```
johnsaigle: Same here, please use the ?? operator.

modules/battery_manager/jsx/addEntry.js
```diff
@@ -0,0 +1,368 @@
...
+              value={this.state.formData.firstVisit}
+            />
+            <NumericElement
+              name="instrumentOrder"
+              label="Instrument Order"
+              onUserInput={this.setFormData}
+              ref="instrumentOrder"
+              required={false}
+              min="0"
+              max="127"
```
ZainVirani: this seems arbitrary

modules/battery_manager/README.md
```diff
@@ -0,0 +1,32 @@
+# Battery Manager
+
+## Purpose
+
+The Battery Manager module allows users to **add**, **activate**, and **deactivate** entries 
+in the Test Battery. 
+
+## Intended Users
+
+The Battery Manager module is used by study administrators
```
PapillonMcGill: Missing ending dot

modules/battery_manager/help/battery_manager.md
```diff
@@ -0,0 +1,33 @@
...
+button to reset all filters.
+
+Within the data table, results can be sorted in ascending or descending order by 
+clicking on any column header.
+
+## Adding an entry to the Test Battery
+
+Under the “Add” tab, users will be able to add a new entry to the Test Battery.
+Users must specify information about the entry by:
+- Selecting from a number of dropdown menus including Instrument, Stage, Subproject, Visit Label, and First Visit
```
PapillonMcGill: quotes

modules/battery_manager/ajax/add_entry.php
```diff
@@ -0,0 +1,137 @@
+<?php
+/**
+ * Battery Manager adder
```
ridz1208: `Battery Manager adder` -> `Battery Manager entry adder` ?

modules/battery_manager/ajax/add_entry.php
```diff
@@ -0,0 +1,137 @@
...
+    $i         = 0;
+    $connector = "WHERE ";
+    foreach ($form_data as $key => $value) {
+        if ($i > 0) {
+            $connector = "AND ";
+        }
+        if (isset($value)) {
+            $query .= $connector . $key . " = :" . $key . " ";
+        } else {
+            $query .= $connector . $key . " IS NULL ";
```
ridz1208: if a field is empty this would return false and yet they would be functionally equivalent, you should add a check for the empty case

modules/battery_manager/php/battery_manager.class.inc
```diff
@@ -0,0 +1,245 @@
...
+    public $AjaxModule         = true;
+    public $skipTemplate       = true;
+
+    /**
+     * Check user permissions
+     *
+     * @return bool
+     * @throws ConfigurationException
+     */
+    function _hasAccess()
```
PapillonMcGill: should include return type

modules/battery_manager/ajax/add_entry.php
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    // Select duplicate entry from Test Battery
+    $entry = $db->pselectRow(
+        $query,
+        $form_data
+    );
+
+    // Return JSON representation of duplicate entry if it exists, null otherwise
+    if ($entry) {
```
ridz1208: generally better to be more explicit, `if(is_null(...))` or `if(empty(...))`... there would be les chances of unexpected behaviour

modules/battery_manager/jsx/addEntry.js
```diff
@@ -0,0 +1,368 @@
...
+
+    let formObj = new FormData();
+    for (let key in formData) {
+      if (formData[key] !== "") {
+        formObj.append(key, formData[key]);
+      }
+    }
+     // Check if entry the user is trying to add already exists in the table
+     // Pass in callback function to call after check for duplicate
+    this.isDuplicate(formObj, this.giveOptions);
```
ridz1208: since this is on the front end only there is potential to hack it

modules/battery_manager/test/TestPlan.md
```diff
@@ -0,0 +1,85 @@
+## Battery Manager Module Test Plan
+
+### 📄 Overview
+
+Battery Manager module allows users to search for, add, activate, and deactivate entries in the
+Test Battery
```
PapillonMcGill: should add ending dots at the end of every phrases

modules/battery_manager/ajax/add_entry.php
```diff
@@ -0,0 +1,137 @@
...
+if (isset($_GET['action'])) {
+    $action = $_GET['action'];
+    if ($action == "checkForDuplicate") {
+        echo checkForDuplicate();
+    } else if ($action == "add") {
+        addEntry();
+    }
+}
+
+/**
```
ridz1208: this check seems to be done on the front-end which is okai but should also happen on the back end since the front end JS code can be manipulated by the user !

validation in general is minimally on the backend, max front end and back end but never FE alone

modules/battery_manager/test/TestPlan.md
```diff
@@ -0,0 +1,85 @@
...
+  1. Add view permission to the aforementioned user
+  2. Battery Manager module should be accessible and only present with **one** tab (Browse)
+  3. The **Deactivate** column should not be in the data table
+
+**Testing with edit permission** [Automation Testing]
+  1. Add edit permission
+  2. Battery Manager module should now have **two** tabs (Browse) and (Add)
+  3. The **Deactivate** should be in the data table
+  4. Clicking on Add tab should hide the data table and display a form with the following fields:
+     Instrument, Minimum age, Maximum age, Stage, Subproject, Visit Label, Site, First Visit,
```
PapillonMcGill: quotes

modules/battery_manager/test/TestPlan.md
```diff
@@ -0,0 +1,85 @@
+## Battery Manager Module Test Plan
+
+### 📄 Overview
```
PapillonMcGill: Should not have icon in the page.

modules/battery_manager/ajax/add_entry.php
```diff
@@ -0,0 +1,137 @@
...
+ * @license  Loris license
+ * @link     https://github.com/aces/Loris-Trunk
+ */
+
+
+// Determine which action is called
+if (isset($_GET['action'])) {
+    $action = $_GET['action'];
+    if ($action == "checkForDuplicate") {
+        echo checkForDuplicate();
```
ridz1208: curiousity, why does this line have an `echo` while line 23 doesnt ?

modules/battery_manager/ajax/add_entry.php
```diff
@@ -0,0 +1,137 @@
...
+        header("HTTP/1.1 403 Forbidden");
+        exit;
+    }
+
+    // Retrieve values entered by user
+    $form_data = getFormData();
+
+    // Make sure new entry is active
+    $form_data['Active'] = 'Y';
+
```
ridz1208: as mentioned above, this should double check that its not a duplicate entry, it can use the function above to do so

modules/battery_manager/ajax/add_entry.php
```diff
@@ -0,0 +1,137 @@
...
+                   'Test_name'    => $_POST['instrument'] ?? null,
+                   'AgeMinDays'   => $_POST['ageMinDays'] ?? null,
+                   'AgeMaxDays'   => $_POST['ageMaxDays'] ?? null,
+                   'Stage'        => $_POST['stage'] ?? null,
+                   'SubprojectID' => $_POST['subproject'] ?? null,
+                   'Visit_label'  => $_POST['visitLabel'] ?? null,
+                   'CenterID'     => $_POST['forSite'] ?? null,
+                   'firstVisit'   => $_POST['firstVisit'] ?? null,
+                   'instr_order'  => $_POST['instrumentOrder'] ?? null,
+                  );
```
ridz1208: i think all the entries above should be sanitized 

see https://github.com/aces/Loris/pull/3832/files#diff-a2650fffe7f2fdc8d8fd8f988fd27bc4R144

@johnsaigle please confirm

modules/battery_manager/ajax/get_form_data.php
```diff
@@ -0,0 +1,74 @@
...
+                  'Screening'     => 'Screening',
+                  'Visit'         => 'Visit',
+                  'Approval'      => 'Approval',
+                  'Subject'       => 'Subject',
+                  'Recycling Bin' => 'Recycling Bin',
+                 );
+    return $stageList;
+}
+
+/**
```
ridz1208: this might be an overkill but I LOVEE ITT

modules/battery_manager/ajax/update_entry.php
```diff
@@ -0,0 +1,58 @@
...
+        )
+    );
+
+    $new_entry = $DB->pselectRow(
+        " SELECT * FROM test_battery WHERE ID = :batteryID ",
+        array("batteryID" => $_POST["ID"])
+    );
+
+    if (empty($new_entry)) {
+        throw new Exception("Updated entry but could not fetch it");
```
ridz1208: did you mean to throw an exception or simply showerror as you did in add entry ??
eitherway a comment above explaining why you made that choice can be usefull

modules/battery_manager/ajax/update_entry.php
```diff
@@ -0,0 +1,58 @@
...
+function updateEntry($value)
+{
+
+    $DB = Database::singleton();
+
+    $DB->update(
+        "test_battery",
+        array("Active" => $value),
+        array(
+         "ID" => $_POST["ID"],
```
ridz1208: dont trust post

sanitize

modules/battery_manager/ajax/update_entry.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+}
+
+
+/**
+ * Update entry by changing Active status
+ *
+ * @param string $value which sets Active to Y or N
+ *
+ * @throws Exception
```
ridz1208: add more info -> when ? why ?

modules/battery_manager/ajax/update_entry.php
```diff
@@ -0,0 +1,58 @@
...
+
+    $DB->update(
+        "test_battery",
+        array("Active" => $value),
+        array(
+         "ID" => $_POST["ID"],
+        )
+    );
+
+    $new_entry = $DB->pselectRow(
```
ridz1208: I'm not sure why you are doing this? why are you reselcting the values ?

modules/battery_manager/php/battery_manager.class.inc
```diff
@@ -0,0 +1,245 @@
...
+        $this->addBasicText('maximumAge', 'Maximum age');
+        $this->addSelect('stage', 'Stage', $stages);
+        $this->addSelect('subproject', 'Subproject', $subprojects);
+        $this->addSelect('visitLabel', 'Visit Label', $visits);;
+        $this->addSelect('site', 'Site', $sites);
+        $this->addSelect('firstVisit', 'First Visit', $firstVisit);
+        $this->addBasicText('instrumentOrder', 'Instrument Order');
+        $this->addSelect('active', 'Active', $active);
+
+        $this->tpl_data["hasWritePermission"] = $this->hasWritePermission;
```
ridz1208: tpl_data???

modules/battery_manager/ajax/get_form_data.php
```diff
@@ -0,0 +1,74 @@
...
+                 'subprojects' => Utility::getSubprojectList(null),
+                 'visits'      => Utility::getVisitList(),
+                 'sites'       => Utility::getSiteList(false),
+                 'firstVisits' => getYesNoList(),
+                ];
+
+    return json_encode($formData);
+}
+
+/**
```
ridz1208: functions below seem to be duplications from the main class functions, use those instead, avoid codeduplications

modules/battery_manager/php/battery_manager.class.inc
```diff
@@ -0,0 +1,245 @@
...
+        // the base query
+        $query  = " FROM test_battery b
+                   LEFT JOIN test_names t ON b.Test_name = t.Test_name
+                   LEFT JOIN subproject s ON b.SubprojectID = s.SubprojectID
+                   LEFT JOIN psc p ON b.CenterID = p.CenterID ";
+        $query .= " WHERE 1=1 ";
+
+        // set the class variables
+        $this->columns
+            = [
```
ridz1208: consensus is to user `array(...)` if you found this in another class, the other class should be amended in a separate PR

modules/battery_manager/php/battery_manager.class.inc
```diff
@@ -0,0 +1,245 @@
...
+               'b.instr_order as instrumentOrder',
+               'b.Active as active',
+               'b.id',
+              ];
+
+        $this->query = $query;
+
+        $this->group_by = '';
+        $this->order_by = 'Full_name';
+        $this->headers  = [
```
ridz1208: `[...]` -> `array(...)`

modules/battery_manager/php/battery_manager.class.inc
```diff
@@ -0,0 +1,245 @@
...
+     * @return array of javascript to be inserted
+     */
+    function getCSSDependencies()
+    {
+        $factory = \NDB_Factory::singleton();
+        $baseURL = $factory->settings()->getBaseURL();
+        $deps    = parent::getCSSDependencies();
+        return array_merge(
+            $deps,
+            [$baseURL . "/battery_manager/css/battery_manager.css"]
```
ridz1208: `[...]` -> `array(...)`

ridz1208:
@davidblader please review JSX/JS code and see if its up to standards

modules/battery_manager/ajax/update_entry.php
johnsaigle: Technically we should be fine here because `update` does prepared statements and everything.

That said it is probably worth doing a simple check (in this case I think using `is_numeric`) and giving a 400 status code if it fails.

SQL/Archive/2018_07_23_battery_manager_permissions.sql
```diff
@@ -0,0 +1,14 @@
...
+INSERT INTO permissions (code, description, categoryID) VALUES ('battery_manager_view', 'View Battery Manager', 2);
+
+-- Add edit permission for battery manager
+INSERT INTO permissions (code, description, categoryID) VALUES ('battery_manager_edit', 'Add, activate, and deactivate entries in Test Battery', 2);
+
+-- Give view permission to admin
+INSERT INTO user_perm_rel (userID, permID) VALUES ((SELECT ID FROM users WHERE UserID = 'admin'), (SELECT permID FROM permissions WHERE code = 'battery_manager_view'));
+
+-- Give edit permission to admin
+INSERT INTO user_perm_rel (userID, permID) VALUES ((SELECT ID FROM users WHERE UserID = 'admin'), (SELECT permID FROM permissions WHERE code = 'battery_manager_edit'));
```
llevitis: There's no `UserID` equal to "admin" in the IBIS production `users` table, so this patch fails as a NULL value can't be inserted into the `userID` column of `user_perm_rel`.

johnsaigle:
@ridz1208 You'll need to find someone to adopt this PR as @victoriafoing is moving away!

modules/battery_manager/ajax/add_entry.php
```diff
@@ -0,0 +1,137 @@
...
+    $db =& \Database::singleton();
+
+    // Retrieve values entered by user
+    $form_data = getFormData();
+
+    // Build SQL query based on values entered by user
+    $query     = " SELECT * FROM test_battery ";
+    $i         = 0;
+    $connector = "WHERE ";
+    foreach ($form_data as $key => $value) {
```
davidblader: not sure about this. in my experience, an entry should be considered a duplicate if there is an existing entry based on `Test_name, Stage, SubprojectID, Visit_label`, some projects might also include `CenterID` as well. 
this would allow duplicates through variances in AgeMin/Max and instr_order

modules/battery_manager/jsx/addEntry.js
```diff
@@ -0,0 +1,368 @@
...
+   * Give options depending on duplicate entry:
+   * If there is an inactive duplicate entry, give option to activate it
+   * If there is an active duplicate entry, give no options
+   * If there is no duplicate entry, add entry
+   *
+   * @param {string} duplicateEntry returned by server
+   */
+  giveOptions(duplicateEntry) {
+        // if duplicate entry exists, convert to JSON
+    if (Object.keys(duplicateEntry).length > 0) {
```
davidblader: `duplicateEntry` is a string when you pass it to `Object.keys`
i think some other `isset`-esque check would be more appropriate

modules/battery_manager/jsx/addEntry.js
```diff
@@ -0,0 +1,368 @@
...
+      let duplicateEntryJSON = JSON.parse(duplicateEntry);
+          // create object with ID of duplicate entry
+      let entryID =	duplicateEntryJSON.ID;
+      var idObj = new FormData();
```
davidblader: maybe only define this within the anon function you have within the swal. also use `let`

modules/battery_manager/ajax/add_entry.php
```diff
@@ -44,7 +46,7 @@ function checkForDuplicate()
         if ($i > 0) {
             $connector = "AND ";
         }
-        if (isset($value)) {
+        if (isset($value) || $value !== null) {
```
johnsaigle: @victoriafoing I think @ridz1208 wanted you to check for an empty string `""` here instead of null.

johnsaigle:
```php
if (isset($value) && !empty($value)) {
```

modules/battery_manager/ajax/add_entry.php
```diff
@@ -58,11 +60,18 @@ function checkForDuplicate()
         $query,
         $form_data
     );
-
-    // Return JSON representation of duplicate entry if it exists, null otherwise
-    if ($entry) {
+    
+   
+   // Return JSON representation of duplicate entry if it exists, null otherwise
+    //if ($entry) {
```
johnsaigle: These should be deleted if no longer needed. :+1:

victoriafoing:
NOTE: I am leaving today but I have made the following changes to the module:
- Changed **Deactivate** column to **Change Status** column and added an **Activate** button in that column
- Added a **Edit Metadata** column with link "Edit" that users can click to edit the entry (similar to Media)

I will spend this week cleaning up my code and responding to the requested changes.
If the module needs more work, I will pass it on to someone else! :)

modules/battery_manager/ajax/add_entry.php
```diff
@@ -116,6 +161,15 @@ function getFormData()
                    'instr_order'  => $_POST['instrumentOrder'] ?? null,
                   );
 
+     /*if (isset($_POST['ID'])) {
+         $form_data['ID'] = $_POST['ID'];
+     }*/
+     
+
+     if (isset($_POST['active'])) {
+         $form_data['Active'] = $_POST['active'];
```
PapillonMcGill: need to be sanetized

modules/battery_manager/ajax/get_form_data.php
```diff
@@ -18,16 +18,53 @@
 if (isset($_GET['action'])) {
     $action = $_GET['action'];
     if ($action === "getFormData") {
-        echo getFormData();
+       if (isset($_GET['ID'])) {
+        $entryID = $_GET['ID'];
```
PapillonMcGill: also need to be sanitized

modules/battery_manager/ajax/get_form_data.php
```diff
@@ -18,16 +18,53 @@
...
+ $allData['batteryData'] = getEntryData($entryID);
+ //echo json_encode($allData);
+ return json_encode($allData);
+}
+
+function getEntryData($entryID)
+{
+ $db = \Database::singleton();
+
+ $query = " SELECT * FROM test_battery WHERE id = :id ";
```
PapillonMcGill: should specific the list of attributes instead of **\***

modules/battery_manager/test/TestPlan.md
victoriafoing: could you elaborate where I should add quotes?

modules/battery_manager/test/TestPlan.md
```diff
@@ -0,0 +1,85 @@
...
+
+Battery Manager module allows users to search for, add, activate, and deactivate entries in the
+Test Battery
+
+### 🔒 Permissions
```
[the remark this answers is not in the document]
victoriafoing: could you rephrase your comment? i'm not sure i understand

modules/battery_manager/test/TestPlan.md
PapillonMcGill: each field as they are label that are shown on the user interface.

modules/battery_manager/test/TestPlan.md
```diff
@@ -1,85 +1,83 @@
...
+     - Make sure that a success message appears and the page goes back to the Browse tab.
+     - Make sure the entry you just added is shown in data table.
 
 **Testing browse functionality**
-  1. After a couple of entries are added, make sure they are properly displayed in the data table
-  2. Make sure that information in the data table corresponds to the information in the `test_battery` table
-  3. Click on 👉  **column headers**  to make sure sorting functionality is working as expected (Ascending/Descending)
+  1. After a couple of entries are added, make sure they are properly displayed in the data table.
+  2. Make sure that information in the data table corresponds to the information in the `test_battery` table.
+  3. Click on 👉  **column headers**  to make sure sorting functionality is working as expected (Ascending/Descending).
```
PapillonMcGill: icon!

modules/battery_manager/ajax/add_or_edit_entry.php
```diff
@@ -0,0 +1,207 @@
...
+                  'AgeMinDays'   => $_POST['ageMinDays'] ?? null,
+                  'AgeMaxDays'   => $_POST['ageMaxDays'] ?? null,
+                  'Stage'        => $_POST['stage'] ?? null,
+                  'SubprojectID' => $_POST['subproject'] ?? null,
+                  'Visit_label'  => $_POST['visitLabel'] ?? null,
+                  'CenterID'     => $_POST['forSite'] ?? null,
+                  'firstVisit'   => $_POST['firstVisit'] ?? null,
+                  'instr_order'  => $_POST['instrumentOrder'] ?? null,
+                 );
+
```
PapillonMcGill: Data need to be sanitized to prevent SQL injection. Had been flagged before but still not resolved.

modules/battery_manager/help/battery_manager.md
```diff
@@ -0,0 +1,63 @@
+# Battery Manager
+
+The Battery Manager module serves as a front-end for browsing, adding, editing, and activating/deactivating entries in the Test Battery.
+
+## Searching for an entry in the Test Battery
+
+Under the `Browse` tab, use the `Selection Filters` to search for entries by fields such 
+as `Instrument`, `Minimum age (days)`, `Maximum age (days)`, `Stage`, `Subproject`, `Visit Label`,
+`Site`, `First Visit`, `Instrument Order`, and `Active`. As filters are selected, the data table below will
+dynamically update with relevant results. Click the “Clear Filters”
```
PapillonMcGill: suggest to put `Clear Filters` in bold instead than between quotes to be as other references to button on this page.

modules/battery_manager/php/battery_manager.class.inc
PapillonMcGill: Still not resolved  
``` function _hasAccess() :Bool```

same for other function

File: modules/battery_manager/php/battery_manager.class.inc [hunk 43]
```diff
@@ -0,0 +1,247 @@
...
+        $subprojects = $this->makeKeySameAsValue(\Utility::getSubprojectList(null));
+
+        $visits = \Utility::getVisitList();
+
+        $sites = $this->makeKeySameAsValue(\Utility::getSiteList(false));
+
+        $firstVisit = $this->getYesNoList();
+
+        $active = $this->getYesNoList();
+
```
PapillonMcGill: section hard to read. Suggest removing empty line and aligning equal signs

modules/battery_manager/php/edit.class.inc
```diff
@@ -0,0 +1,112 @@
...
+    public $skipTemplate = true;
+
+    /**
+     * Entry point for /battery_manager/edit/
+     * Checks if entry id is provided and otherwise redirects to
+     * /battery_manager/ page
+     *
+     * @return none
+     */
+    function setup()
```
PapillonMcGill: please add return type in function declaration

modules/battery_manager/ajax/add_or_edit_entry.php
```diff
@@ -0,0 +1,207 @@
...
+        editEntry();
+    }
+}
+
+/**
+ * Return duplicate of entry if it exists and null otherwise
+ *
+ * @return json object or null
+ */
+function checkForDuplicate()
```
PapillonMcGill: please include return type in function definition

modules/battery_manager/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Scope
+
+The Battery Manager module provides a tool for browsing, adding, editing, and activating/deactivating entries in the
+the Test Battery.
+
+### Browse
+
+The **browse** functionality is under the `Browse` tab.
+Users can browse entries in the Test Battery and filter the entries using the `Selection Filters`.
```
driusan: These paragraphs with line breaks after every sentence are weird and hard to read (also, some of them are too long. `fmt`.

modules/battery_manager/README.md
```diff
@@ -0,0 +1,53 @@
...
+The **edit** functionality can be accessed in the Menu Table under the `Browse` tab.
+Users can edit an entry by clicking on the corresponding `Edit` link in the `Edit Metadata` column.
+This will take them to a new page with a form that is originally populated by the values of the entry and that can be edited.
+Users cannot edit an entry if they have made no changes in the form or if they have made changes to the point where the entry becomes the same as another entry in the Test Battery table.
+
+### Activate/Deactivate
+
+The **activate/deactivate** functionalities can be accessed in the Menu Table under the `Browse` tab.
+
+In the `Change Status` column, users can directly activate or deactivate an entry by pressing the **Activate** or **Deactivate** button.
```
driusan: Is there any protection/check/confirmation/something to ensure they don't accidentally press them?

modules/battery_manager/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+In the `Edit Metadata` column, users can indirectly activate or deactivate an entry by navigating to the `Edit` form and changing the value of `Active`. Within an entry's `Edit` form, users can also indirectly activate or deactivate another entry if they edit an entry to the point where the entry becomes the same as another entry in the Test Battery but has a different `Active` value. In this case, the entry that was originally selected for editing will not be updated.
```
driusan: This functionality seems a little weird and I'm not sure of the reasoning behind it.

Also, is there history kept somewhere of the edits?

Given the below paragraph, maybe it would make more sense to implement edit as "Create new test battery and deactivate the current one." so that nothing is ever lost.

modules/battery_manager/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+The **activate** functionality is also in the `Add` tab. If users add an entry that already exists in the Test Battery table but has been deactivated, they will be given the option to activate the existing entry.
```
driusan: 👍

modules/battery_manager/ajax/add_or_edit_entry.php
```diff
@@ -0,0 +1,207 @@
+<?php
```
driusan: This will probably have to go into another pass of the module but:

1. Add and edit should be two different endpoints, since they're doing two different things
2. They should be implemented as handlers in the php directory, the ajax directory should be deprecated.

modules/battery_manager/ajax/add_or_edit_entry.php
```diff
@@ -0,0 +1,207 @@
...
+
+// Determine which action is called
+if (isset($_GET['action'])) {
+    $action = $_GET['action'];
+    if ($action == "checkForDuplicate") {
+        echo checkForDuplicate();
+    } else if ($action == "add") {
+        addEntry();
+    } else if ($action == "edit") {
+        editEntry();
```
driusan: Add and edit shouldn't be done via a get request, they should be done via a post or put

modules/battery_manager/ajax/add_or_edit_entry.php
driusan: also, the return type appears to be either null or string, not a json object.

modules/battery_manager/ajax/add_or_edit_entry.php
```diff
@@ -0,0 +1,207 @@
...
+                   instr_order FROM test_battery ";
+    $i         = 0;
+    $connector = "WHERE ";
+    // Iterate through values entered by user
+    foreach ($form_data as $key => $value) {
+        if ($i > 0) {
+            $connector = "AND ";
+        }
+        if (isset($value) || $value !== null) {
+            $query .= $connector . $key . " = :" . $key . " ";
```
driusan: This is dangerous and should be using a prepared statement.

modules/battery_manager/ajax/add_or_edit_entry.php
```diff
@@ -0,0 +1,207 @@
...
+ *
+ * @return void
+ */
+function showError($message)
+{
+    if (!isset($message)) {
+        $message = 'An unknown error occurred!';
+    }
+    header('HTTP/1.1 500 Internal Server Error');
+    header('Content-Type: application/json; charset=UTF-8');
```
driusan: This doesn't seem to be return json? (or doing anything with message?)

modules/battery_manager/ajax/add_or_edit_entry.php
driusan: It would be much more robust to prevent SQL injection via prepared statements

modules/battery_manager/ajax/change_active_status.php
```diff
@@ -0,0 +1,63 @@
...
+    $DB->update(
+        "test_battery",
+        array("Active" => $value),
+        array(
+         "ID" => $_POST["ID"],
+        )
+    );
+
+    $new_entry = $DB->pselectRow(
+        " SELECT * FROM test_battery WHERE ID = :batteryID AND Active = :active",
```
driusan: Is there a reason this can't be an explicit list of columns instead of `*`?

modules/battery_manager/help/battery_manager.md
```diff
@@ -0,0 +1,63 @@
+# Battery Manager
```
driusan: The line lengths make this really hard to read too.

modules/battery_manager/jsx/addEntry.js
```diff
@@ -0,0 +1,357 @@
...
+        self.setState({
+          error: 'An error occurred when loading the form!'
+        });
+      }
+    });
+  }
+
+  render() {
+    // Data loading error
+    if (this.state.error !== undefined) {
```
driusan: If `this.state.error` is the empty string, this will show up, which I don't think is intended.

modules/battery_manager/jsx/addEntry.js
```diff
@@ -0,0 +1,357 @@
...
+        <div className="alert alert-danger text-center">
+          <strong>
+            {this.state.error}
+          </strong>
+        </div>
+      );
+    }
+
+    // Waiting for data to load
+    if (!this.state.isLoaded) {
```
driusan: Can this use @HenriRabalais's new loader component?

modules/battery_manager/jsx/addEntry.js
```diff
@@ -0,0 +1,357 @@
...
+    this.setFormData = this.setFormData.bind(this);
+    this.handleAdd = this.handleAdd.bind(this);
+    this.isDuplicate = this.isDuplicate.bind(this);
+    this.giveOptions = this.giveOptions.bind(this);
+    this.activateEntry = this.activateEntry.bind(this);
+    this.addEntry = this.addEntry.bind(this);
+  }
+
+  componentDidMount() {
+    var self = this;
```
intralizee: let self = this;

modules/battery_manager/jsx/addEntry.js
```diff
@@ -0,0 +1,357 @@
...
+          Loading
+          <span
+            className="glyphicon glyphicon-refresh glyphicon-refresh-animate">
+          </span>
+        </button>
+      );
+    }
+
+    // Inform users about duplicate entries
+    var helpText = (
```
intralizee: const helpText = ...

modules/battery_manager/jsx/batteryManagerIndex.js
```diff
@@ -0,0 +1,260 @@
...
+   * Create Activate button in Change Status column if the entry is not Active
+   * @param {string} column - column name
+   * @param {string} cell - cell content
+   * @param {array} rowData - array of cell contents for a specific row
+   * @param {array} rowHeaders - array of table headers (column names)
+   * @return {*} a formatted table cell for a given column
+   */
+  formatColumn(column, cell, rowData, rowHeaders) {
+     // Create the mapping between rowHeaders and rowData in a row object.
+    var row = {};
```
intralizee: let row = {};

modules/battery_manager/jsx/batteryManagerIndex.js
```diff
@@ -0,0 +1,260 @@
...
+    }, this);
+
+     // create array of classes to be added to td tag
+    var classes = [];
+    classes = classes.join(" ");
+
+     // Create Change Status column with Active and Deactivate buttons
+    if (column === 'Change Status') {
+      let entryID = row['Change Status'];
+      var idObj = new FormData();
```
intralizee: let idObj = new FormData();

modules/battery_manager/jsx/batteryManagerIndex.js
```diff
@@ -0,0 +1,260 @@
...
+        return <td className={classes}><button onClick={() => {
+          this.activateEntry(idObj);
+        }}>Activate</button></td>;
+      }
+    }
+
+    // Create Edit Metadata column with Edit link that allows user to edit entry
+    if (column === 'Edit Metadata') {
+      let entryID = row['Edit Metadata'];
+      var editURL = loris.BaseURL + "/battery_manager/edit/?id=" + entryID;
```
intralizee: const editURL = ...

modules/battery_manager/jsx/addEntry.js
zaliqarosli: @driusan doesn't it need to be merged first before we can use it?

modules/battery_manager/php/edit.class.inc
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * @category Loris
+ * @package  Battery_Manager
+ * @author   Victoria Foing <victoria.foing@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+class Edit extends \NDB_Form
+{
+    public $skipTemplate = true;
```
zaliqarosli: do we need this line if we are not using smarty templates?

modules/battery_manager/php/edit.class.inc
```diff
@@ -0,0 +1,112 @@
...
+        );
+    }
+
+    /**
+     * Check user permissions
+     *
+     * @return bool
+     * @throws ConfigurationException
+     */
+    function _hasAccess()
```
zaliqarosli: not a big deal, but let's put hasAccess at the top of this class

modules/battery_manager/php/battery_manager.class.inc
```diff
@@ -0,0 +1,247 @@
...
+ * @package  Battery_Manager
+ * @author   Victoria Foing <victoria.foing@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+class Battery_Manager extends \NDB_Menu_Filter
+{
+    public $hasWritePermission = false;
+    public $AjaxModule         = true;
+    public $skipTemplate       = true;
```
zaliqarosli: not sure if this is needed is there are no smarty templates

modules/battery_manager/jsx/editEntryIndex.js
```diff
@@ -0,0 +1,30 @@
+/** global ReactDOM
+ *
+ * Battery Manager Edit Form
+ *
+ * Module component rendering Edit form
+ *
+ * @author Victoria Foing
+ *
+ */
+import BatteryManagerEditForm from './editEntry';
```
zaliqarosli: 1. it seems to me that there should really be only 1 index file that renders all the other components. 
2. if this index file only imports 1 component (editEntry), then I believe the call to render can just be appended to the end of the editEntry class

modules/battery_manager/php/edit.class.inc
driusan: the line is explicitly telling it not to use smarty templates.

modules/battery_manager/jsx/addEntry.js
driusan: good point.. I thought it was merged, ignore my comment until it is.

modules/battery_manager/php/edit.class.inc
zaliqarosli: but is it required if there are no smart templates in the module?

Jkat:
Hi, don't forget to add this module to the default `0000-00-02-Menus.sql` file

SQL/Archive/2018_07_23_battery_manager_permissions.sql
Jkat: The 'upgrading your loris' wiki section uses ID 1 for the 'admin' user. Until there's a better solution, maybe we could continue with this convention.

modules/battery_manager/ajax/add_or_edit_entry.php
Jkat: I concur

modules/battery_manager/ajax/get_form_data.php
Jkat: Haha, this should totally be converted into a Utility function or something

modules/battery_manager/php/battery_manager.class.inc
```diff
@@ -0,0 +1,247 @@
...
+     */
+    function getCSSDependencies()
+    {
+        $factory = \NDB_Factory::singleton();
+        $baseURL = $factory->settings()->getBaseURL();
+        $deps    = parent::getCSSDependencies();
+        return array_merge(
+            $deps,
+            array(
+             $baseURL . "/battery_manager/css/battery_manager.css",
```
Jkat: This part may not be needed if the .css file committed is empty

modules/battery_manager/php/edit.class.inc
```diff
@@ -0,0 +1,112 @@
...
+     */
+    function getCSSDependencies()
+    {
+        $factory = \NDB_Factory::singleton();
+        $baseURL = $factory->settings()->getBaseURL();
+        $deps    = parent::getCSSDependencies();
+        return array_merge(
+            $deps,
+            array(
+             $baseURL . "/battery_manager/css/battery_manager.css",
```
Jkat: This part may not be needed if the .css file committed is empty

modules/battery_manager/php/edit.class.inc
victoriafoing: i think its necessary because when i comment out this line, i get:
"The following errors occurred while attempting to display this page:
 Unable to load template 'file:menu_battery_manager.tpl'"

modules/battery_manager/ajax/add_or_edit_entry.php
victoriafoing: if i add exit($message) in the last line, the error message is passed to the front-end. however, i was told to avoid die/exit. is there a better to return the error?

modules/battery_manager/ajax/add_or_edit_entry.php
victoriafoing: to find a duplicate of an entry, we have to build an SQL query that looks at all the values of the entry. the for loop goes through all the values in the entry and determines if they have been filled out or not. if a value is filled out it adds " $key = :$key ". if a value is not filled out, it is NULL so the SQL query must contain " $key IS NULL " instead of " $key = :$key " and the $key must be removed from the parameter array.
for example, if I add an entry with the following values:
Test_name: bmi, AgeMinDays: 0, AgeMaxDays: 99999, Stage: Visit, and I do not fill out SubprojectID, Visit_label, CenterID, firstVisit, instr_order. 
The SQL query is "SELECT * FROM test_battery WHERE Test_name = :Test_name AND AgeMinDays =:AgeMinDays AND AgeMaxDays = :AgeMaxDays AND Stage = :Stage AND SubprojectID IS NULL AND Visit_label IS NULL AND CenterID is NULL AND firstVisit IS NULL AND instr_order IS NULL" with parameters ['Test_name' => 'bmi', 'AgeMinDays' => '0', 'AgeMaxDays' => '99999', 'Stage' => 'Visit']

modules/battery_manager/php/edit.class.inc
victoriafoing: is "@return void" correct?

victoriafoing: or should i change it to "@return bool" and add a "return true" at the end?

modules/battery_manager/php/battery_manager.class.inc
PapillonMcGill: @victoriafoing in the function declaration.

i.e. "function _hasAccess() : bool"
in the case the function accept parameters:
function name(string $param1, int $param2) : string
you can see more at http://php.net/manual/en/migration70.new-features.php and http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration

modules/battery_manager/ajax/add_entry.php
victoriafoing: because checkForDuplicate must return a JSON representation of the duplicate entry to the JS file (or null if there isn't one)

modules/battery_manager/README.md
PapillonMcGill: Not sure about the requirement but I would say permit edit to some field and still create + deactivate.

victoriafoing:
Hi everyone! Thanks for the feedback. I have tackled as much of it as I can.
This PR is now ready to be passed on to someone else.

Cheers,
Vicky :)

modules/battery_manager/README.md
```diff
@@ -2,16 +2,17 @@
 
 ## Purpose
 
-The Battery Manager module allows users to **browse**, **add**, **edit**, and **activate/deactivate** entries 
-in the Test Battery. 
+The Battery Manager module allows users to **browse**, **add**, **edit**, **activate**, **deactivate** entries
```
PapillonMcGill: should be ``` ... activate and deactivate ...```

modules/battery_manager/README.md
```diff
@@ -22,32 +23,52 @@ Users can browse entries in the Test Battery and filter the entries using the `S
...
+- Users cannot edit an entry if they have made no changes in the form.
+- Users cannot edit an entry if it becomes the same as another active entry in the Test Battery.
+- If the edited entry becomes the same as another deactivated entry in the Test Battery,
+  users will be given the option to activate the other entry and deactivate the original one.
 
-### Activate/Deactivate
+### Activate
 
-The **activate/deactivate** functionalities can be accessed in the Menu Table under the `Browse` tab.
+The **activate** functionality can be accessed in the `Browse` tab, `Edit` window, and the `Add` tab..
```
PapillonMcGill: 1 extra dot at the end

SQL/Archive/2018_07_23_battery_manager_permissions.sql
```diff
@@ -0,0 +1,14 @@
+-- Add Battery Manager to LorisMenu
```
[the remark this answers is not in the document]
HenriRabalais: `SQL/New_patches` doesn't seem to exist, should I put it there anyways?

SQL/Archive/2018_07_23_battery_manager_permissions.sql
```diff
@@ -0,0 +1,14 @@
...
+INSERT INTO LorisMenu (Parent, Label, Link, Visible, OrderNumber) VALUES (6, 'Battery Manager', 'battery_manager/', null, 7);
```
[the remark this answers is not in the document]
HenriRabalais: Could you specify exactly what should be selected for? Is it only the `Parent`?

e.g. (SELECT `Parent` From `LorisMenu` WHERE `Parent` = `Admin`)

SQL/Archive/2018_07_23_battery_manager_permissions.sql
```diff
@@ -0,0 +1,14 @@
...
+
+-- Add view permission for battery manager
+INSERT INTO permissions (code, description, categoryID) VALUES ('battery_manager_view', 'View Battery Manager', 2);
+
+-- Add edit permission for battery manager
+INSERT INTO permissions (code, description, categoryID) VALUES ('battery_manager_edit', 'Add, activate, and deactivate entries in Test Battery', 2);
+
+-- Give view permission to admin
+INSERT INTO user_perm_rel (userID, permID) VALUES ((SELECT ID FROM users WHERE UserID = 'admin'), (SELECT permID FROM permissions WHERE code = 'battery_manager_view'));
```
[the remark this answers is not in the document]
HenriRabalais: Could you give a quick example of how that would work in this context?

modules/battery_manager/help/battery_manager.md
```diff
@@ -0,0 +1,33 @@
...
```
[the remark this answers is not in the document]
HenriRabalais: Would love to but I'm afraid I have no idea. I will make a note and try and fill it in later.

modules/battery_manager/php/battery_manager.class.inc
HenriRabalais: The page will not load unless `$skipTemplate` is set to `true`

HenriRabalais:
Took over maintenance of PR from @victoriafoing and addressed reviews. Major JSX refactoring to come.

SQL/Archive/2018_07_23_battery_manager_permissions.sql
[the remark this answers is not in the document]
HenriRabalais: @ridz1208 Please answer.

SQL/Archive/2018_07_23_battery_manager_permissions.sql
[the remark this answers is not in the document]
HenriRabalais: @ridz1208 Please answer.

modules/battery_manager/ajax/get_form_data.php
HenriRabalais: @ridz1208 Please elaborate/be more specific about which functions.

modules/battery_manager/php/edit.class.inc
PapillonMcGill: setup() : void

SQL/New_patches/2018_07_23_battery_manager_permissions.sql
```diff
@@ -1,7 +1,8 @@
 -- Add Battery Manager to LorisMenu
-INSERT INTO   LorisMenu (Parent, Label, Link, Visible, OrderNumber)
+INSERT INTO LorisMenu (Parent, Label, Link, Visible, OrderNumber)
        VALUES (
-           (SELECT ID FROM LorisMenu WHERE Label = 'Admin', 'Battery Manager'),
+           (SELECT ID FROM (SELECT * FROM LorisMenu) as lm WHERE Label = 'Admin'),
```
PapillonMcGill: ... (SELECT ID, 'Battery Manager, null, 7 FROM LorisMenu where Label = 'Admin' );...

HenriRabalais: Instead of `SELECT *`?

PapillonMcGill: instead of lines 4-9

PapillonMcGill: Just saw your updated comment.

Right, you can't insert in the same table your selecting from.

PapillonMcGill: SELECT ID FROM (SELECT ID FROM LorisMenu WHERE Label = 'Admin') AS LM, ...

HenriRabalais: Sounds good!

PR #3463, opened 2018-01-30
Add AnonymousUser class to differentiate logged in from logged out users

driusan:
This should allow us to make better use of PHP's type system when dealing with users.

PR #3848, opened 2018-07-25
[survey module] Remove checks for permissions when entring direct data

ridz1208:
This implements a new function to be used solely to load survey instruments. this function is based on the regular `loadinstrument()` function in the caller but removes all unecessary calls pertaining to regular instrument loading whn logged in. Surveys is an anonymous module and thus should not be checking for permissions or loading panels...

### Why ?
The issue was raised during release testing (https://redmine.cbrain.mcgill.ca/issues/14889) and was due to this change (https://github.com/aces/Loris/pull/3463/files#diff-1947816db4abdcdd8722747580999a6dR52) which was preventing the load of permissions when it's a public module. previously all permissions were loaded by default because of the left join in the following query (https://github.com/aces/Loris/blob/minor/php/libraries/UserPermissions.class.inc#L81)

### DIFF
--- a/php/libraries/NDB_Caller.class.inc
+++ b/php/libraries/NDB_Caller.class.inc
@@ -251,7 +251,6 @@ class NDB_Caller
             $phpfile   = $base . "project/instruments/"
                                . "NDB_BVL_Instrument_$test_name.class.inc";
             $linstfile = $base."project/instruments/$test_name.linst";
-
             if ($this->existsAndRequire($phpfile)
                 || $test_name=='instrument_preview'
             ) {
@@ -542,6 +541,14 @@ class NDB_Caller
         );
 
         $this->instrument = $instrument;
+        $success = $instrument->save();
+        if ($redirectToOnSuccess !== null && $success !== false) {
+            header("Location: $redirectToOnSuccess");
+            return "";
+        }
+        // create an instrument status object
+        $status  = new NDB_BVL_InstrumentStatus;
+        $success = $status->select($commentID);
 
         $this->page = $instrument;
         $html       = $instrument->display();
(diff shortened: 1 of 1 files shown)


php/libraries/NDB_Caller.class.inc
```diff
@@ -485,6 +505,50 @@ class NDB_Caller
...
+    ) {
+        if ($page === 'finalpage') {
+            $instrument = NDB_BVL_Instrument::factory(
+                $instrumentName,
+                $commentID,
+                $page
+            );
+
+            $this->instrument = $instrument;
+            header("Location: $redirectToOnSuccess");
```
johnsaigle: What happens when you redirect someone to `null`?

php/libraries/NDB_Caller.class.inc
```diff
@@ -485,6 +505,50 @@ class NDB_Caller
...
+     *                                    successfully.
+     *
+     * @return string HTML of the page to render.
+     */
+    function loadInstrumentDirect(
+        $instrumentName,
+        $page,
+        $commentID,
+        $redirectToOnSuccess=null
+    ) {
```
johnsaigle: Can you add `: String` to indicate the return type?

johnsaigle: 🤷‍♂️ OK but it might help us catch errors.

sruthymathew123:
Tested on IBIS and everything works fine.

php/libraries/NDB_Caller.class.inc
ridz1208: we can do it in a followup PR to minor

PR #3851, opened 2018-07-26
modify the onclick function all range in the table data

PapillonMcGill:
This pull request `repair JS the use the proper link to jump to a page`. It `an onclick function was applied to all column of the table`.

See also: `redmine 14810`


ridz1208:
@PapillonMcGill you are modifying files that should not be in this PR, I have restarted travis twice because this PR is causing a failure in the conflict resolver (which it should not be affecting). if the PR fails this pass, please rebase and only submit the files you are modifying

ridz1208:
passing travis, good to go

PapillonMcGill:
@ridz1208 Should I rebase even if Travis pass?

ridz1208:
@PapillonMcGill no its okai

ridz1208:
@PapillonMcGill resolve conflicts please by recompiling JS files

PR #3853, opened 2018-07-26
[Imaging Browser] DICOM Tar download as original name

um4r12:
## Summary

This pull request will download the tar file generated by the imaging pipeline as the original name (with which it was uploaded with).

The original file name is queried and sent as a GET parameter to get_file.php, and the file is downloaded with that name.

Made for CAP, but as suggested by others, this maybe useful for LORIS.

## Testing Instructions
To test you will need an imaging sandbox.

1. Go to the DICOM Archive module to see a list of potential tarchives that are present in your DB.
2. Ensure that the file actually exists in your VM (under `/data/{projectName}/data/tarchive/{year}/{fileName.tar}`.
3. Select the 'View Images' under the `MRI Browser` column for a given candidate.
4. In the side panel, ensure that a `Download DICOM` button is present under the `Links` header.
5. Click `Download DICOM {#}` to download the tar file. It should automatically be downloaded as `PatientName_CandID_VL_%`, where the ending could be anything, but it will be the same name as the name it was initially uploaded with. To ensure, you could head to the `Imaging Uploader` module and verify that that is indeed the case for that candidate and visit.

## Questions
The query used to retrieve the `UploadLocation` does a join on the `mri_upload` table. Thus, this assumes every tar file processed by the pipeline will also exist in the `mri_upload` table. To the imaging experts: is this assumption valid? If the file only exists in the `tarchive` table, and not in the `mri_upload`, then the query result will be empty.

To the security gurus: How should I be sanitizing the `_GET` parameter, is that still necessary in this case, as it is only used in the php `header(...)` function? 

## References
See also: `[PR 293](https://github.com/aces/Loris-MRI/pull/293)`, previously closed as it was decided to implement modifying the name at the time of download rather than the imaging pipeline.

### DIFF
--- a/htdocs/mri/jiv/get_file.php
+++ b/htdocs/mri/jiv/get_file.php
@@ -153,7 +153,7 @@ case 'DICOMTAR':
     $FullPath         = $tarchivePath . '/' . $File;
     $MimeType         = 'application/x-tar';
     $DownloadFilename = basename($File);
-    $saveAs = $_GET['saveAs'];
+    $saveAs           = $_GET['saveAs'];
     break;
 default:
     $FullPath         = $DownloadPath . '/' . $File;
@@ -170,10 +170,10 @@ if (!file_exists($FullPath)) {
 
 header("Content-type: $MimeType");
 if (!empty($DownloadFilename)) {
-   
-   if ($FileExt === 'DICOMTAR' && !empty($saveAs)) {
+
+    if ($FileExt === 'DICOMTAR' && !empty($saveAs)) {
         header("Content-Disposition: attachment; filename=$saveAs");
-    
+
     } else {
 
         header("Content-Disposition: attachment; filename=$DownloadFilename");
--- a/htdocs/mri/jiv/get_file.php
+++ b/htdocs/mri/jiv/get_file.php
@@ -66,7 +66,6 @@ if ($imagePath === '/' || $DownloadPath === '/'
 // Now get the file and do file validation.
 // Resolve the filename before doing anything.
 $File = Utility::resolvePath($_GET['file']);
-
 // Extra sanity checks, just in case something went wrong with path resolution.
 // File validation
 if (strpos($File, ".") === false) {
@@ -154,6 +153,7 @@ case 'DICOMTAR':
     $FullPath         = $tarchivePath . '/' . $File;
     $MimeType         = 'application/x-tar';
     $DownloadFilename = basename($File);
+    $saveAs = $_GET['saveAs'];
     break;
 default:
     $FullPath         = $DownloadPath . '/' . $File;
@@ -170,8 +170,14 @@ if (!file_exists($FullPath)) {
 
 header("Content-type: $MimeType");
 if (!empty($DownloadFilename)) {
+   
+   if ($FileExt === 'DICOMTAR' && !empty($saveAs)) {
+        header("Content-Disposition: attachment; filename=$saveAs");
+    
+    } else {
 
-    header("Content-Disposition: attachment; filename=$DownloadFilename");
+        header("Content-Disposition: attachment; filename=$DownloadFilename");
+    }
 }
 $fp = fopen($FullPath, 'r');
 fpassthru($fp);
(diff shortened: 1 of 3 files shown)


ridz1208:
@um4r12 doesnt seem like a 20.0 PR to me. is there a testing ticket for this?  is it a bugfix ?

htdocs/mri/jiv/get_file.php
```diff
@@ -154,6 +153,7 @@
     $FullPath         = $tarchivePath . '/' . $File;
     $MimeType         = 'application/x-tar';
     $DownloadFilename = basename($File);
+    $saveAs           = $_GET['saveAs'];
```
PapillonMcGill: This parameter should be check for security issues before being use.

modules/imaging_browser/php/imaging_session_controlpanel.class.inc
```diff
@@ -105,18 +105,23 @@ class Imaging_Session_ControlPanel
...
             $params['PVL']     = $timePoint->getVisitLabel();
         }
+
         // To enable DICOM download: Add pair: tarchiveID, ArchiveLocation
         $qresult = $DB->pselect(
-            "SELECT TarchiveID, ArchiveLocation 
-            FROM tarchive 
-            WHERE PatientName LIKE $ID",
+            "SELECT t.TarchiveID, t.ArchiveLocation, 
+                SUBSTRING_INDEX(mu.UploadLocation, '/', -1) AS originalFileName,
```
cmadjar: I am not sure to understand why you are joining the mri_upload table. Also, doing that might break projects that are older than the mri_upload table (IBIS, PREVENT-AD...). For the renaming, I would simply append the patient name at the end of the archiveLocation. This way you have access to the patient name and can retraced from which archiveLocation this was downloaded if needed.

For example: `DCM_2017-04-10_ImagingUpload-11-33-4KNC3m.tar` could be renamed `DCM_2017-04-10_ImagingUpload-11-33-4KNC3m_MTL0123_456789_V01.tar`

cmadjar: Also, the query should be prepared

cmadjar: Also, if the `SessionID` has not been set by the imaging pipeline in the tarchive table, this query will return nothing and the download will download the DICOM archive with `ArchiveLocation` as the name. So `DCM_2017-04-10_ImagingUpload-11-33-4KNC3m.tar` would be downloaded as `DCM_2017-04-10_ImagingUpload-11-33-4KNC3m.tar`, no patient name info...

um4r12: The reason for joining the mri_upload table was so that we are able to retrieve the original filename and download the file with that name as that is what the users will be familiar with.
 It is not guaranteed that the original file name will be equivalent to the `PatientName` as you can add a suffix to the file that is being uploaded after the patient name (ex. `PatientName_anySuffix.zip`). 
If there are projects that do not have the mri_upload tables, then I agree we should just use the `ArchiveLocation` and append the PatientName. 

In which case will the `SessionID` not be set by the imaging pipeline? Do you suggest I remove the `t.sessionID=mu.sessionID` in the where clause?

cmadjar: It is possible that the sessionID is null in the tarchive table if no MINC files were inserted into the files table. I guess for the imaging browser, it does not matter since you are looking for inserted MINC files in that module. Might be important to remember when doing the download from the DICOM archive module.

cmadjar:
@johnsaigle I think this is still something that we want. @christinerogers can confirm. I can also ask during the imaging meeting tomorrow and let you know.

christinerogers:
Yes @johnsaigle @cmadjar This is very much still a desired feature.

johnsaigle:
Thanks for the info. I'll get to this ASAP

driusan:
I think @johnsaigle forgot to click the close button.

PR #3854, opened 2018-07-27
[Database] Change return type of pselectOne to null

ridz1208:
This is meant to prevent `pselectOne()` from returning an empty array when the result is not in the database. Given that on success the function is expecting a value, it is less error prone to return null when nothing is return instead of an empty array.

functions `pselectRow()` and `pselect()` already expect an array to be returned and thus can be left as they are.

This helps avoid countless errors and warnings of all sorts in the error log when the result of a `pselectOne` is processed without first doing something like
```
$result = $DB->pselectOne(SOMETHING)
if (!is_array($result)) {
    //do what I want to do
}
```

### DIFF
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -806,6 +806,8 @@ class Database
         if (is_array($result) && count($result)) {
             $valueArray = array_values($result);
             $result     = $valueArray[0];
+        } else {
+            return null;
         }
         return $result;
     }


ridz1208:
@driusan @xlecours how dangerous is this in a bugfix release ??

cmadjar:
I find this risky too for a bugfix. Call me paranoid but there are many uses of `pselectOne` in the code and I would tend to properly test each module after this change to make sure it does not break anything. If I remember correctly, massive testing does not happen during bugfix releases.

johnsaigle:
@ridz1208 As we discussed in person I think it would be better for this function to return an empty string rather than null. This was we can also use a return type declaration (which we eventually want for every function).

HenriRabalais:
@johnsaigle could you elaborate on the logic of why returning `''` would be better than returning `null`?

intralizee:
I've been searching the codebase for all the cases where pselectOne is used. 
So far it looks like devs have been writing:
`if (!empty($result)) {...}`
after using pselectOne and which will be the same result if null vs empty array.
Also the case where devs try to do `$result['Example']` will return '' when $result is null.

johnsaigle:
@HenriRabalais A classic source of ammo against PHP is that it allows you to do whatever you want in terms of typing... usually causing things to break. The source of the errors/warnings that @ridz1208 mentions above is treating a variable as one type instead of another. Specifically he was telling me about downstream DB insertion where some part of the processing gives a message like "Got array, was expecting string."

As if PHP 7, PHP now has [type hinting](https://secure.php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration) and [return type declarations](https://wiki.php.net/rfc/return_types) allowing us to strongly specify what types we are expecting. I've spoken to @driusan about this and he said that we should be using return type declarations everywhere (for now just with new code, but eventually the whole codebase will use this).

So all of this is to say that I think the right move here is to make this function ALWAYS return a string. This way PHP's error handling will agree with our documentation and everyone will (hopefully) know that they should always be treating the value as a string.

This adds consistency which I think is very helpful.

HenriRabalais:
@johnsaigle Oh I see. Yes, for `pselectOne()` returning a `string` does make sense. For some reason I thought this conversation was happening in the context of `pselect()`.

johnsaigle:
@Jkat I feel like you're right. It crossed my mind how to distinguish between empty strings and actual empty results but I didn't think it through enough.

The best case to fix the errors is probably to actually deal with them...

driusan:
I agree with @Jkat that this is dangerous to put in bugfix, and doesn't have any way to differentiate between "this value is explicitly null" and "the query returned no results." It should probably go to major. I also don't think we can type hint pselectOne because the type of the underlying data is unknown (it could be a date, or a varchar, or an integer database column.)

What if on major we made it throw an exception if nothing was returned? It might be overkill, but then at least we'd still be able to differentiate "no value" from "explicit null value"

johnsaigle:
@driusan That sounds good.

Should this PR be closed then?

johnsaigle:
@driusan I am doing some work now on adding return type declarations to the Database class and I've run into this issue again. `PDOStatement::fetchAll` itself returns a mixed type so we cannot enforce return types unless throwing an Exception (like you suggested) or otherwise converting the failure case to be an empty array ("The empty set" as the MySQL message describes).

driusan:
Is the method modified in #3880 used by pselectOne? If so, I agree that this can just be closed since it's superceded by that (@ridz1208 ?)

ridz1208:
APPROVE

PR #3856, opened 2018-07-30
[webpack][node_modules] Updated node_modules and webpack to 4.x.x

intralizee:
This pull request updates "node_modules" to latest versions. 

1) Specifically webpack is now 4.16.4 and I made the required adjustments to the webpack.config.js file. 
2) I removed babel-preset-es2015 which is deprecated and replaced with babel-preset-env. 
3) Also added webpack-cli which is now required.

4) Another edition to package.json:
> "lintfix:javascript": "eslint --fix --ext .js, htdocs jsx modules"

Developers can now execute: `npm run lintfix:javascript` to fix simple eslint errors automatically.

5) Added transform-object-rest-spread to babel. https://babeljs.io/docs/en/babel-plugin-transform-object-rest-spread/ Developers can now use the new js6 spread operator for more cases with react frontend development.
Examples:
const oldArray = [1,2,3];
const newArray = [...oldArray, 4, 5]; // now it's [1,2,3,4,5];
also:
const oldObject = {
    name: 'Alizee'
};
const newObject = {
   ..oldObject,
   gender: 'female'
}; // now it's {'name: Alizee', gender: 'female'}

I tested by running these command:
`npm run compile`
`npm run lint:javascript`
`npm run lintfix:javascript`
and everything seems to be working at first glance.


johnsaigle:
Removing myself because JavaScript is not my strong suit

intralizee:
Awesome Travis now reports:
`Unexpected var, use let or const instead`
and
'Strings must use singlequote' 

YaY! I prefer single quotes instead of having to press & hold shift key.

htdocs/js/instrument_form_control.js
```diff
@@ -1,26 +1,26 @@
...
   } else if (index === 0) {
     $('[name=' + name + ']').prop('disabled', false);
     $(this).parent().removeClass('has-warning');
-    $("#" + name).remove();
+    $('#' + name).remove();
   } else {
     $('[name=' + name + ']').prop('disabled', true);
     $(this).parent().addClass('has-warning');
-    $(this).after("<div class=\"col-xs-12 warning\" id=\"" + name + "\">Any entered data will not be saved</div>");
+    $(this).after('<div class="col-xs-12 warning" id="' + name + '">Any entered data will not be saved</div>');
```
um4r12: Is there a reason why double quotes are used within the html here and then single quotes (escaped) on other lines (for ex. line 38). I think we should just replace all double quotes and have escaped single quotes to be consistent. That shouldn't cause any problems, right?

intralizee: I actually didn't make that change because I used the new command: 
eslint --fix --ext .js, htdocs jsx modules
In javascript land it's usually done the way you're seeing it now after the eslint command.

modules/configuration/test/ConfigurationTest.php
```diff
@@ -65,10 +65,14 @@ public function tearDown()
...
     {
         $this->safeGet($this->url . "/configuration/");
-
-        $bodyText = $this->webDriver
-            ->findElement(WebDriverBy::cssSelector("body"))->getText();
-        $this->assertContains("Configuration", $bodyText);
+        $bodyText = $this->webDriver->findElement(
+            WebDriverBy::cssSelector("body")
+        )->getText();
+        $bodyText = $this->webDriver->executescript(
```
kongtiaowang: remove 71 to 74

modules/configuration/test/ConfigurationTest.php
```diff
@@ -65,10 +65,14 @@ public function tearDown()
...
+            "return document.querySelector".
+            "('#bc2 > a:nth-child(2) > div').textContent"
+        );
+        $this->assertRegexp("/Configuration/", $bodyText);
```
kongtiaowang: using "Please enter the various configuration variables" instead of "Configuration"

modules/document_repository/test/document_repositoryTest.php
```diff
@@ -127,7 +127,7 @@ function testDocumentRepositoryDoespageLoad()
         $bodyText = $this->webDriver->findElement(
             WebDriverBy::cssSelector("body")
         )->getText();
-        $this->assertContains("Document Repository", $bodyText);
+        $this->assertRegexp("/Document Repository/", $bodyText);
```
kongtiaowang: using "Upload File" instead of "Document Repository"

kongtiaowang: @intralizee using "Upload File" instead of "Document Repository"

kongtiaowang: using "Upload File" instead of "Document Repository"

webpack.config.js
```diff
@@ -50,52 +50,74 @@ var config = [{
...
+        sourceMap: true,
+      }),
+    ],
+  },
 }];
 
 // Support project overrides
 if (fs.existsSync('./project/webpack-project.config.js')) {
-  var projConfig = require('./project/webpack-project.config.js');
+  const projConfig = require('./project/webpack-project.config.js');
```
ridz1208: @intralizee does the webpack project need to be updated too??

webpack.config.js
```diff
@@ -1,12 +1,12 @@
...
+const path = require('path');
+const fs = require('fs');
 
-var config = [{
+const config = [{
   entry: {
     './htdocs/js/components/DynamicDataTable.js': './jsx/DynamicDataTable.js',
-    './htdocs/js/components/PaginationLinks.js': './jsx/PaginationLinks.js',
-    './htdocs/js/components/StaticDataTable.js': './jsx/StaticDataTable.js',
+    './htdocs/js/components/PaginationLinks.js': ['./jsx/PaginationLinks.js'],
```
ridz1208: why do these have square brackets but not the others ??

webpack.config.js
intralizee: I was actually confused about what to do there because I don't see a "webpack-project.config.js" file in my project directory. Currently I don't believe that line does or break anything.

webpack.config.js
intralizee: @ridz1208 The square brackets solves an error for the frontend when importing the StaticDataTable component. 

The error message "Module not found: Error: a dependency to an entry point is not allowed" will be thrown when running "npm run compile" without the brackets when importing the StaticDataTable into a component.

Supposedly the latest version of React solves the issue but we're still using an older version that requires this fix. Also to note the square brackets are usually used to compile an array of jsx files into one javascript file to serve to the client browser instead of multiple js files.

PR #3855, opened 2018-07-29
[tools] 20.0 tools and SQL moved to proper directory

ridz1208:
This moves 
- the `Normalize_protocol_split_rows.php` to the proper directory and updates its relative references
- the SQL output on `Normalize_Consent_Data.php` to the archive directory directly to avoid uncommitted code in the production environment's `Cleanup_patches` directory.

This also removes trialing white spaces from the end of the files

PR #3857, opened 2018-07-30
[tools/Normalise_Consent_Data.php] Modify insert statement

zaliqarosli:
while testing https://github.com/aces/Loris/pull/3855, I noticed that the script was throwing a DBException when inserting data that already exists in `candidate_consent_rel` from previous runs of the script (because SQL can't duplicate primary keys). 

screenshot from 2018-07-30 14-15-50

this script should be single use anyways so this PR adds an extra check to notify the user that they cannot import/replace existing data

**to test:**
1. On 20.0-branch:
  - run `php Normalise_Consent_Data.php` with data already in `candidate_consent_rel` table and notice the error in screenshot above
  - Call `DELETE FROM candidate_consent_rel`, run the script again and notice there are no issues
2. On my PR branch:
  - run the script again both with and without data in `candidate_consent_rel` table. Inserting the data in either case should not throw the error, but instead gives you the status of the import

### DIFF
--- a/tools/single_use/Normalise_Consent_Data.php
+++ b/tools/single_use/Normalise_Consent_Data.php
@@ -156,10 +156,19 @@ if (empty($configValue)) {
     array_push($errors, "useConsent's Config value missing from Config table.
                Run SQL/New_patches/2018-03-01_normalise_consent.sql");
 }
+// Check new consent tables exist
 if (!$db->tableExists('consent') || !$db->tableExists('candidate_consent_rel') || !$db->tableExists('candidate_consent_history')) {
     array_push($errors, "New, normalized consent tables do not exist.
                Run SQL/New_patches/2018-03-01_normalise_consent.sql");
 }
+// Check rel and history tables are empty
+$consentRelResult     = $db->pselect("SELECT * FROM candidate_consent_rel", array());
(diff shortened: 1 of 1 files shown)


zaliqarosli:
@ridz1208 please review

colezweber:
Running into errors testing this on IBIS but the same error is appearing when I checkout 20.0-release so its not an issue with your PR but rather the IBIS data tables/my config.

ridz1208:
@zaliqarosli as discussed, the changes made in #3694 should not concern your PR

tools/single_use/Normalise_Consent_Data.php
```diff
@@ -239,7 +239,25 @@
 
 // Populate candidate_consent_rel
 foreach ($dataArray as $consentValues) {
-    $db->insert('candidate_consent_rel', $consentValues);
+    $candidateID    = $consentValues['CandidateID'];
```
ridz1208: why not do this in the validation code above ??

zaliqarosli: 😑

zaliqarosli: because you require the `consent` table to be populated before doing this validation

ridz1208: why not check in the validation if ANY data is there
?

PR #3860, opened 2018-07-31
Fix duplication of MRI violations when multiple subprojects per candidate

cmadjar:
This pull request fixes a bug in the MRI violations module for which it displays the same violations more than once if a candidate has visits that are spread across multiple subprojects.

Example of MRI violations before the fix for a subject belonging to two different subprojects:
screen shot 2018-07-31 at 4 05 51 pm
Same candidate MRI violations after the fix from this PR:
screen shot 2018-07-31 at 4 07 35 pm

How to test this PR:
Pick one candidate that shows in the MRI violations module and that has multiple visits. If all these visits are all from the same subproject, manually edit one in the `session` table so that it has a different subprojectID than all other visits (could be NULL) and:
1) check that you can reproduce the bug on 20.0-release
2) check that the bug is fixed by this PR

See also: https://redmine.cbrain.mcgill.ca/issues/14960

### DIFF
--- a/modules/mri_violations/php/mri_violations.class.inc
+++ b/modules/mri_violations/php/mri_violations.class.inc
@@ -231,17 +231,19 @@ class Mri_Violations extends \NDB_Menu_Filter_Form
                 SeriesUID,
                 md5(concat_WS(':',minc_location,PatientName,SeriesUID,time_run))
                   as hash,
-                mri_protocol_violated_scans.ID as join_id,
+                mpvs.ID as join_id,
                 p.CenterID as Site,
                 violations_resolved.Resolved as Resolved
-            FROM mri_protocol_violated_scans
+            FROM mri_protocol_violated_scans AS mpvs
             LEFT JOIN violations_resolved
-            ON (violations_resolved.ExtID=mri_protocol_violated_scans.ID
+            ON (violations_resolved.ExtID=mpvs.ID
             AND violations_resolved.TypeTable='mri_protocol_violated_scans')
             LEFT JOIN candidate c
-            ON (mri_protocol_violated_scans.CandID = c.CandID)
+            ON (mpvs.CandID = c.CandID)
             LEFT JOIN session s
-            ON (mri_protocol_violated_scans.CandID = s.CandID)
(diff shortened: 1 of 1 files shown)


cmadjar:
FYI: to whomever is reviewing this, I added the "Needs Rebase" label but it should be OK once 20.0-release is merged into bug-fix.

modules/mri_violations/php/mri_violations.class.inc
```diff
@@ -294,18 +296,12 @@ class Mri_Violations extends \NDB_Menu_Filter_Form
                    )
                 ) as hash,
                 MRICandidateErrors.ID as join_id,
-                p.CenterID as Site,
+                null,
```
um4r12: @cmadjar Could you explain why lines 285-286 and 299 were changed to null?

um4r12:
Passed manual testing.

modules/mri_violations/php/mri_violations.class.inc
um4r12: Shouldn't a join return a `NULL` value anyways if the value doesn't exist? But that does make sense. Thanks for the clarification!

cmadjar:
@nicolasbrossard @um4r12 any chance you could review this PR?

modules/mri_violations/php/mri_violations.class.inc
driusan: What @um4r12 said.. wouldn't a join return a null value anyways?

cmadjar: I don't see the point of having a JOIN here even if it returns a NULL value. In this case, we don't know the PSCID or CandID or Visit label so why trying to join to the session or candidate table when you know you'll get a NULL value?

cmadjar: Well, for some errors you get a subprojectID and ProjectID with the assumption the PSCID is correct (but you have PSCID and CandID not matching so it might not be information for the correct candidate). 
Leigh agreed with me that it is better to have no information than information that could be wrong when a scan has not been properly associated with a session and candidate, hence the hardcoded NULL.

PR #3863, opened 2018-08-04
[JSX] recompiling files before release

ridz1208:
This recompiles all JSX files again


davidblader:
did a quick manual test of each affected module and everything seems good

PR #3864, opened 2018-08-06
[instrument_list] Remove superglobal access

driusan:
This refactors the fix in #3721 in order to remove references
to superglobals. Instead, the variables are instantiated on the
module level from the PSR object so that the page can assume
they exist.

### DIFF
--- a/modules/instrument_list/php/instrument_list.class.inc
+++ b/modules/instrument_list/php/instrument_list.class.inc
@@ -16,6 +16,7 @@
  * @package behavioural
  */
 namespace LORIS\instrument_list;
+
 use \Psr\Http\Message\ServerRequestInterface;
 use \Psr\Http\Server\RequestHandlerInterface;
 use \Psr\Http\Message\ResponseInterface;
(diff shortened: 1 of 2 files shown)


driusan:
I'm sending this to 20.0-release because that's the only place that #3721 exists, but it can go into a release after that.

johnsaigle:
@driusan I rebased for you

modules/instrument_list/php/instrument_list.class.inc
```diff
@@ -101,33 +125,13 @@ class Instrument_List extends \NDB_Menu_Filter
...
-        $gets = $request->getQueryParams();
-
-        $attribute = $request->getAttribute("CandID");
-        if ($attribute !== null) {
-            $this->candidate = \Candidate::singleton($attribute);
-        } else {
-            $this->candidate = \Candidate::singleton($gets['candID']);
+        $this->candidate = $request->getAttribute("Candidate");
+        $this->timePoint = $request->getAttribute("timePoint");
+        if ($this->candidate == null || $this->timePoint === null) {
```
PapillonMcGill: Is there a reason why for candidate `==` is used while `===` is used for timepoint? Would be better if both use `===`

modules/instrument_list/php/instrument_list.class.inc
```diff
@@ -47,6 +48,29 @@ class Instrument_List extends \NDB_Menu_Filter
...
+     *
+     * @return void
+     */
+    public function setTimePoint(\TimePoint $timepoint)
+    {
+        $this->timePoint = $timepoint;
+    }
+
+    /**
+     * Sets the candidate is being accessed on this page
```
zaliqarosli: Sets the candidate that is being accessed ..

modules/instrument_list/php/instrument_list.class.inc
zaliqarosli: i agree. this seems to be a typo

driusan:
@zaliqarosli / @PapillonMcGill should be fixed, if there was a reason I can't remember why one was using double equals Travis should catch it..

PR #3866, opened 2018-08-07
[Help Editor] Use safe update instead of unsafe update

johnsaigle:
### Brief summary of changes

The HelpFile class has been changed to use the safe update function from the database class instead of the explicitly unsafe version.

`unsafeupdate` is to be used ["with caution, only when you know you need to insert HTML and know that you can trust it"](https://github.com/aces/Loris/blob/20.0-release/php/libraries/Database.class.inc#L442). I don't believe that data submitted by users into the help editor can be considered trusted as this is a cross-site scripting vector.

This change fixes a security issue while preserving the markup functionality of the help editor. HTML will not directly render but Markdown will. I believe this is closer to the intended functionality of the help editor.

### This resolves issue...

No reported issues but this fix mitigates a confirmed cross-site scripting exploit.

### To test this change...

This fix should preserve Markdown functionality but direct HTML will no longer directly render; bold, italics, links and headers can be created using Markdown syntax.

Note that the current implementation of the help editor takes information from the database only for instruments, therefore **to test you must edit the help content for an instrument, not a module**.

- [x] Confirm XSS is not possible on the new branch. See me if you don't know how to do cross-site scripting. 

- [x] Edit the help content for an instrument, including Markdown features like bold, italics, headers, links, etc. All of these should work as expected.

### DIFF
--- a/modules/help_editor/php/helpfile.class.inc
+++ b/modules/help_editor/php/helpfile.class.inc
@@ -87,7 +87,7 @@ class HelpFile
         $DB =& \Database::singleton();
 
         // insert a help file
-        $success = $DB->unsafeinsert('help', $set);
+        $success = $DB->insert('help', $set);
         // return the help ID
         return $DB->lastInsertID;
     }
--- a/modules/help_editor/php/helpfile.class.inc
+++ b/modules/help_editor/php/helpfile.class.inc
@@ -106,7 +106,7 @@ class HelpFile
         $DB =& \Database::singleton();
 
         // update the help file
-        $success = $DB->unsafeupdate('help', $set, array('helpID' => $this->helpID));
+        $success = $DB->update('help', $set, array('helpID' => $this->helpID));
 
         return true;
     }


zaliqarosli:
im assigning myself to test and review this because i made this initial change from safe to unsafe in #3558 (https://github.com/aces/Loris/pull/3558/files#r176128444). i had reservations but the reason we made the change is to avoid escaped html stored in the DB from rendering as special chars on the front-end. i'll look into whether this is still the case

PapillonMcGill:
I do support that change. However, is it easy for "non-computer literate" to add some markdown in the text?
Do we want to offer a script to replace what is stored in the DB from HTML to markdown (Could be complicated)?

johnsaigle:
@cmadjar If you take a look at our [Markdown file](https://github.com/aces/Loris/blob/minor/jsx/Markdown.js) neither scratch nor lists are implemented. This should be true both before and after my changes.

zaliqarosli:
@PapillonMcGill I have spoken to meg evans as part of the UX movement to add a markdown text editor to the text area in help editor. a ticket has been made for this (https://redmine.cbrain.mcgill.ca/issues/14210)

zaliqarosli:
@johnsaigle i understand. im just saying that there might be another way that using the safe function changes the rendering of the html for which i will test! there might be a reason why we changed it in the first place (which i have forgotten) but also, there might be no issues!

modules/help_editor/php/helpfile.class.inc
```diff
@@ -106,7 +106,7 @@ class HelpFile
         $DB =& \Database::singleton();
 
         // update the help file
-        $success = $DB->unsafeupdate('help', $set, array('helpID' => $this->helpID));
+        $success = $DB->update('help', $set, array('helpID' => $this->helpID));
```
zaliqarosli: there is also an unsafeinsert on line 90 that we could change as part of this PR

johnsaigle: Nice catch! I'll change that now

zaliqarosli: so i think you're all good! i'm not sure why we changed it to unsafe insert/update in the first place

ZainVirani: @zaliqarosli out of curiousity, why is the insert function `static public` but the update function is not?

zaliqarosli: @ZainVirani i have no idea, i wasn't the original author

johnsaigle:
@zaliqarosli Done!

um4r12:
LGTM, tested changes; html is not rendered, and markdown works.

I don't know if this is worth noting, however, if you do "Preview" on the help_editor, html is rendered, and the XSS injection is still possible. 

screen shot 2018-08-13 at 2 25 48 pm

zaliqarosli:
@um4r12 i noticed this too but i think it's fine as the preview takes content directly from the text-area on the front-end and is never actually processed into the backend

johnsaigle:
@um4r12 Yeah I noticed that too. Self-xss isn't ideal but it's a pretty low-risk vulnerability as far as things go. 

@zaliqarosli is right that this doesn't affect other users.

driusan:
If I remember correctly it was using unsafe not to insert html, but because the escaping was escaping (and re-escaping) special characters like quotes on save, so if you edit something that had been saved the quotes would become the string "&quot" on re-save, then "&amp;quot", then "&amp;amp;quot", etc.

Has this been tested for that? (or am I misremembering? @zaliqarosli @me-evans ?)

johnsaigle:
@driusan If that ends up being the case we could maybe add a functionality to use htmlspecialchars with "no quotes" mode as an alternative.

zaliqarosli:
@driusan i've tested this and the issue with resaving escaped chars is not an issue because of this (https://github.com/aces/Loris/blob/24e633a7eda29ef5578830e82d88bdc0fd84ff7b/modules/help_editor/php/edit_help_content.class.inc#L90). escaped chars gets rendered as html in the form, and then get resaved as is, and then saved as escaped into the back-end with the changes made in this PR. 

this is also not an issue with the Markdown rendering because of this (https://github.com/aces/Loris/blob/24e633a7eda29ef5578830e82d88bdc0fd84ff7b/jsx/Markdown.js#L23).

if the issue that i'm saying doesn't exist still arises, it's because the content in the database needs to be cleaned up. i'm adding back the passed manual tests label

johnsaigle:
@zaliqarosli Thanks for that info!

driusan:
@zaliqarosli great, I just wanted to be sure before merging it.

PR #3867, opened 2018-08-07
[Smarty Errors] "Go to login page" --> "Go to main page"

johnsaigle:
When LORIS returns a 403, 404, or 500 error, a small template is shown to display that error. It currently says "Go to login page" when the URL actually points to baseURL. As a result this message is inaccurate when a user is already logged in as they'll be redirected to the dashboard (not logged out).


johnsaigle:
Note: issued to bugfix so the irrelevant commits will disappear when 20.0 is merged into bugfix.

PR #3869, opened 2018-08-09
[Core: NDB_Form] Cross-site Request Forgery (CSRF) protection proof-of-concept

johnsaigle:
### (Somewhat) Brief summary of changes

**This PR is a minimal proof-of-concept of how to implement [CSRF](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)) protection in LORIS.**

I chose to create a `setup()` function in `NDB_Form` that will add a hidden token to each form created. This provides something like a "middleware" which means that developers of modules will not need to know very much about the inner workings of the CSRF protections. They will need to manually add `__csrfToken` elements in Smarty. 

The token is set in a cookie and stored in the Database. NDB_Forms will be created using the Cookie value which will be checked against what the server has stored. For more information see @intralizee's link below. This approach allows for AJAX requests to submit the same cookie info as a custom header (not yet implemented).

The proof-of-concept takes places within the `user_accounts` module, specifically the _My Preferences_ page. This was done because it is not a heavily-trafficked part of LORIS and it is also not yet using ReactJS (which I am less familiar with than Smarty).

There should be no change to functionality except for the added token. If the token is not present or malformed, the POST request to change preference information will be rejected, displaying an error.

#### Design choices

I made certain decisions and assumptions after thinking about this problem for a while. **All of this is very much up for discussion**.

* After speaking with @driusan about how to do this and receiving feedback from @intralizee, this PR has been modified to use a cookie/database approach to storing the tokens.

* **The token expires after 1 day** -- this gives users the option to leave a page open for a reasonable length of time before the token would otherwise expire and cause them to have to start their form submission from the beginning.

### This resolves issue...

This is the first baby step toward mitigating an entire class of security vulnerabilities.

### Further work to be done

This is just a proof-of-concept so there is a lot left to do in terms of mitigating CSRF attacks.

#### Short term

- [ ] Refactor existing code utilizing `NDB_Form` to use CSRF tokens.

- [x] ~~Determine an equivalent process for injecting tokens into Reactified modules and forms not using `NDB_Form`~~. Tokens can be submitted via custom headers by AJAX. This is not yet implemented but is fairly straightforward to do.

- [ ] Decide on a way to delete old tokens. A cron script?

#### Medium term

* Identify an exhaustive list of state-changing requests in LORIS (to the extent this is feasible). 

* Use tokens for **every** state-changing request. We can start with especially sensitive operations first (logging in, file uploads, altering existing data...). Normally this will be POST requests but this is not always the case in LORIS. 

#### Long term

* Normalize LORIS such that all state-changing requests are done via POST (or PUT, etc. But not GET.)

* Create a middleware attaching CSRF tokens to all POST requests.

### To test this change...

- [ ] Checkout my branch. Go to the my preferences page. Look in the page source for `__csrfToken`. Check that there is a long random token attached.

- [ ] Alter some information on the preference page and Save. Everything should work smoothly.

- [ ] Resubmit the form but this time manually delete the CSRF token or alter its value. Verify that your request is rejected. Verify that no preference information changed.

- [ ] Login to a separate user account and try altering the token value to a token from another session. Verify that this fails.

- [ ] Alter the token `CreatedAt` date in MySQL to be older than a day. Verify that submitted the old token fails.


php/libraries/NDB_Form.class.inc
```diff
@@ -145,6 +217,49 @@ class NDB_Form extends NDB_Page
...
+
+            // This is the recommended way to get values from the JWT token
+            // according to the Firebase documentation.
+            $decodedArray = (array) $decoded;
+            $realUser     = $_SESSION['State']->getUsername();
+            error_log($realUser);
+            $authenticRequest = ($decodedArray['user'] === $realUser);
+
+            if (!$authenticRequest) {
+                error_log("ERROR: Bad CSRF token! Possible hacking attempt. $e");
```
PapillonMcGill: Why just log the possible hack attempt here when at line 241, the code dies after logging the error?

ZainVirani: I could be wrong but I think it's because line 239 will only catch the case where the JWT token is invalid, whereas 256 will handle the case of a valid (read: decodable) JWT token that, for whatever reason, had an invalid username.

johnsaigle: Yeah it's basically to distinguish a "bad" token (not well-formed and so not decode-able by the library" and an "evil" token (which can be decoded but has deliberately false information).

PapillonMcGill: Ok thanks!

ZainVirani:
@johnsaigle When I changed the token string I got the malformed JWT error, but when I deleted the element I was still able to save.

johnsaigle:
@ZainVirani that's by design for right now. If you look in the code comments I explain that the current code only checks if the token is valid -- but does NOT fail requests that don't have it. We can't do that right now or every NDB_Form on LORIS will break. 

When/if this PR is merged, we can start the process of attaching CSRF elements to the Smarty templates of every module and later doing the same for React and for everything that changes Database state.

intralizee:
I'm curious how we will get this to work when using react without having smarty to pass the CSRF token.

One way recommended found on [stackoverflow](https://stackoverflow.com/questions/20504846/why-is-it-common-to-put-csrf-prevention-tokens-in-cookies) is from the backend send the CSRF token from server to client in a cookie (Set-Cookie header) which can't be read or changed by a hacker and when POST happens we validate the token matches what should have been in the cookie.

johnsaigle:
@intralizee I'm not opposed to that approach and @driusan and I talked it over in our original conversation. I think you're right that it'll probably be the easiest method when it comes to using AJAX requests. 

I'll do some refactoring.

johnsaigle:
Removing Passed Manual Tests since I am going to change the functionality

johnsaigle:
Need to add the SQL patches to make this work.

johnsaigle:
@intralizee I have refactored this to go for a database- and cookie-based approach.

SQL/New_patches/2018_08_14_csrf_tokens.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE `csrf_tokens` (
+  `CsrfTokenID` int(11) NOT NULL AUTO_INCREMENT,
```
PapillonMcGill: should be unsigned int

SQL/New_patches/2018_08_14_csrf_tokens.sql
```diff
@@ -0,0 +1,7 @@
...
+  `Token` varchar(64) NOT NULL,
+  `CreatedAt` date NOT NULL,
+  `PhpSessionID` varchar(30) NOT NULL,
+  PRIMARY KEY (`CsrfTokenID`)
+) ENGINE=InnoDB AUTO_INCREMENT=24 DEFAULT CHARSET=latin1;
```
PapillonMcGill: suggest a primary number (at least an odd one) for auto_increment. This way, it wont try to reuse an already used number after circling around.


Any method to remove the old (>2 days) token from the table? In the long run, it will use lots of space.

johnsaigle: @PapillonMcGill Makes sense. I just dumped the table schema to create this. Don't know where 24 came from.

As for deleting the old tokens, I'm open to suggestions. Maybe a cron script to remove old cookies?

johnsaigle:
This PR is failing due to the `my_preferences` integration test... which is good! It means that the form is properly rejecting edits when no CSRF token is present.

I believe code needs to be added around here:
https://github.com/aces/Loris/blob/83fc245e65ac83cf5bce06b99994eac4d185fbb8/modules/user_accounts/test/user_accountsTest.php#L253

In order to grab the value `__csrfToken` element and submit it in the POST body of the request... however I have no experience with integration testing.

I'll try to figure out or meet with someone to learn how to do this. Maybe @nicolasbrossard @kongtiaowang or @driusan could shine some light on what to do here?

SQL/New_patches/2018_08_14_csrf_tokens.sql
PapillonMcGill: @johnsaigle I also suggest a cron job to remove all cookies older that 48 hour. This would keep the size of the table low enough and should not have impact (extremely low) on CPU time.

xlecours:
@johnsaigle , could we use a token that do note require a session state (i.e. JWT token?)
Also, you might have mentioned it in the comment but a middleware to verify Same Origin with standard headers would be a good first step I think.

johnsaigle:
@driusan I did some more research on this topic and it looks like PHP 7.3 will be released in late December with SameSite support for cookies. This flag totally prevents CSRF and would require one line of code to change on our part in `NDB_Client`. Chrome and Firefox both know about this flag and honour it.

I think it makes sense to close this PR as we're unlikely to universally deploy CSRF protection between now and then. Just as long as we make sure to use that flag when 7.3 drops. 

What do you think?

johnsaigle:
From what I understand, yes it will completely prevent CSRF. It prevents flagged cookies from being used from different domains and that's the CSRF attack vector. I will do more reading to confirm but I am confident.

You're right about the minimum PHP upgrade. 

I'm also happier to take the simpler route.

johnsaigle:
This should be replaced by #4083 and can be closed.

PR #3872, opened 2018-08-10
[POST RELEASE] PULL FORWARD OF 20.0.0 INTO MINOR

ridz1208:
Pull forward of all code changes from the **20.0.0** branch to the **minor** branch as per usual after any release.
*This PR should be merged without squashing*


driusan:
No they aren't. Pulling bugfix into minor would result in a diamond shaped graph, pulling 20.0 into each of them independently would result in a tree with two distinct branches.

ridz1208:
Yeah i guess codewise they are equivalent but history wise they arnt. I think weve been following more the diamond shape on loris so you might be right. Would you be able to make these updates today ?

driusan:
Replaced by #3876.. I'll do minor->major after that one goes in.

PR #3878, opened 2018-08-14
[Core: Database] Add type declarations to function signatures

johnsaigle:
### Brief summary of changes

Type declarations and return type declarations are added to all function signatures in the Database class. This follows a discussion about moving toward the universal use of PHP's typing features to prevent notices, warnings, and other nasty behaviour resulting from type coercion. 

I tried not to alter any functionality by following the return types already listed in the PHP Docs for each function:
* In some cases where the documentation conflicted with the actual return type, the documentation was amended and the functionality preserved. 
* Functions that were documented as returning void but had a return statement were changed to not return anything.
* ~~Default values were largely left as is even in cases where another might arguably be better (e.g. Strings set to `null` instead of the empty string)~~ 
    * This has been changed based on feedback by @driusan .

### Further work to be done

Some of the functions were not given return types (notably all `pselect*` functions) because they return mixed types. We need to refactor these functions to behave consistently.

### This resolves issue...

Nothing concrete but many potential issues arising from PHP's type coercion.

**Related PRs:** 
_A check means it's merged_
- [x] [[Core: Database] Add exception on fetchAll failure case ](https://github.com/aces/Loris/pull/3880)
- [x] [[Data Team Helper] Ensure test_name is always a string](#3909)
- [x] [[Core: Database] Add type check to HTMLEscapeArray](#3947)
- [x] [[Statistics] Change pselect calls to pass array()s, not null ](https://github.com/aces/Loris/pull/4003)
- [x] [[Statistics] Change *more* pselect calls to pass array()s, not null ](https://github.com/aces/Loris/pull/4009)
### To test this change...

**Important** if you get an error about a return type like `...must be an instance of void, none returned`, that means the version of PHP being used by your Apache is older than PHP 7.1. As of the last release we require >= 7.2 so be sure to correct this.

- [ ] The Database class should work as before. If it doesn't, it has been used improperly and the calling classes should be modified to use it as documented.

### DIFF
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -73,7 +73,7 @@ class Database
      *
      * @param string  $database     the database to select
      * @param string  $username     the username with which to log into
-     *                                      the database server
+     *                              the database server
      * @param string  $password     the password that matches the username
      * @param string  $host         the name of the database server
      * @param boolean $trackChanges boolean determining if changes should be
@@ -134,11 +134,11 @@ class Database
      *
      * @param string      $database     the database to select
      * @param string|null $username     the username with which to log into
-     *                              the database server
+     *                                  the database server
      * @param string|null $password     the password that matches the username
      * @param string|null $host         the name of the database server
      * @param boolean     $trackChanges whether to use the trackChanges
-     *                              mechanism on this connection
+     *                                  mechanism on this connection
      *
      * @return bool True if DB connection established. False otherwise.
      * @access public
@@ -910,7 +910,7 @@ class Database
      * Runs a query as a prepared statement and returns the value of the first
      * column of the first row.
      *
-     * FIXME This function does not have have a return type declaration because
+     * FIXME This function does not have a return type declaration because
      * the function pselectRow will sometimes return a string and sometimes
      * return an array. This behaviour should be made consistent (i.e. perhaps
      * returning an array of a single element) so that we can properly enforce
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -146,7 +146,7 @@ class Database
      * @access public
      */
     function connect(
-        string $database,
+        $database,
         string $username,
         string $password,
         string $host,
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
 /**
  * This file represents an SQL database abstraction layer for use in Loris
  *
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -147,9 +147,9 @@ class Database
      */
     function connect(
         $database,
-        string $username,
-        string $password,
-        string $host,
+        ?string $username,
+        ?string $password,
+        ?string $host,
         bool   $trackChanges = true
     ) : bool {
 
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -345,14 +345,14 @@ class Database
      * @param bool   $ignore     determines whether the insert throws and error or
      *                           is discarded when value exists in DB
      *
-     * @return void
+     * @return bool
      */
     private function _realinsert(
         string $table,
         array  $set,
         bool   $autoescape = true,
         bool   $ignore = false
-    ): void {
+    ): bool {
 
         if ($autoescape === true) {
             $set = $this->_HTMLEscapeArray($set);
@@ -370,7 +370,7 @@ class Database
 
         if (DEBUG) {
             fwrite(STDERR, $query."\n");
-            return;
+            return true;
         }
 
         // There is no where clause for an insert, so add a fake
@@ -526,7 +526,7 @@ class Database
 
         if (DEBUG) {
             fwrite(STDERR, $query.$where."\n");
-            return;
+            return true;
         }
 
         if (preg_match("/=NULL/", $where)) {
@@ -667,7 +667,7 @@ class Database
         }
 
         if (!empty($options['nofetch']) && $options['nofetch'] == true) {
-            return;
+            return array();
         }
 
         $rows = $prepared->fetchAll(PDO::FETCH_ASSOC);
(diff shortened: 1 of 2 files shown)


driusan:
(Added Needs Work because it's failing Travis, haven't looked into why.)

johnsaigle:
@driusan I believe it's failing in the cases in LORIS where we have baked-in type coercion. It's likely that several smaller PRs normalizing the use of function calls and return values from this function will need to be merged before this can be merged.

Perhaps `Blocked` is more appropriate than `Needs Work`?

driusan:
Sure, I changed the tag.

php/libraries/Database.class.inc
```diff
@@ -144,10 +147,10 @@ class Database
      */
     function connect(
         $database,
-        $username,
-        $password,
-        $host,
-        $trackChanges = true
+        ?string $username,
+        ?string $password,
+        ?string $host,
```
driusan: This one seems weird.. what does it mean to pass those as null?

johnsaigle: Honestly not totally sure. It requires further research. However when these are set to just strings (`null` not allowed) Travis throws up a ton of errors during the execution of unit tests. Something to the effect of "MockDatabase::connect() requires a string parameter for param X but null given".

I don't know enough about our automated testing to know what to do about that, so I allowed changed this function to allow for null. I'm very open to suggestions here.

driusan: @kongtiaowang Any thoughts on how to solve that?

php/libraries/Database.class.inc
```diff
@@ -67,27 +67,30 @@ class Database
...
     static function &singleton(
-        $database=null,
-        $username=null,
-        $password=null,
-        $host=null,
-        $trackChanges=true
+        string $database     = null,
+        string $username     = null,
+        string $password     = null,
+        string $host         = null,
```
driusan: This is going to fail, null isn't a sting.

johnsaigle: I ran phan locally and it didn't seem to have an issue.... which is weird cause you're right.

I'll change this and the if statement below.

driusan: That's.. problematic? Maybe it just checks it at the call site and it's never called without them being explicitly passed?

johnsaigle: Maybe? Not sure how the internals of phan work.

johnsaigle:
This depends on #3909 to pass Travis. Will rebase this once that is merged.

driusan:
Maybe this PR should also add `declare(strict_types=1);` to the file?

johnsaigle:
Yeah I had that earlier but think I removed it because something was going wrong and I thought it was the cause. I'll add it back.

driusan:
@johnsaigle #3909 is merged

php/libraries/Database.class.inc
```diff
@@ -866,14 +894,20 @@ class Database
 
     /**
      * Runs a query as a prepared statement and returns the value of the first
-     * column of the first row
+     * column of the first row.
+     *
+     * FIXME This function does not have have a return type declaration because
```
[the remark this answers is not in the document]
zaliqarosli: or "does not require"

php/libraries/Database.class.inc
```diff
@@ -67,34 +67,37 @@ class Database
...
      * Singleton design pattern implementation - creates the object
-     * and also connects to the database if necessary
+     * and also connects to the database if necessary.
+     * NOTE There is no return type declaration for this function as it returns
+     * the type `Database` which is not supported by PHPs built in return types.
      *
      * @param string  $database     the database to select
      * @param string  $username     the username with which to log into
-     *                              the database server
+     *                                      the database server
```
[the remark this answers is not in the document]
johnsaigle: Dunno phpcs is a nightmare somtimes.

php/libraries/Database.class.inc
```diff
@@ -67,34 +67,37 @@ class Database
...
      * @throws DatabaseException
      * @access public
      */
     static function &singleton(
-        $database=null,
-        $username=null,
-        $password=null,
-        $host=null,
-        $trackChanges=true
+        string $database     = '',
```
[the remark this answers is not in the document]
johnsaigle: Think so. Using `null` was actually less safe because this variable is meant to be a string. I changed the if statement just below it to check for `empty` instead of `null` and I believe that's the only place we need to worry about it.

php/libraries/Database.class.inc
```diff
@@ -328,10 +329,15 @@ class Database
      * @param bool   $ignore     determines whether the insert throws and error or
      *                           is discarded when value exists in DB
      *
-     * @return void
+     * @return bool
```
PapillonMcGill: This function always return true. Is it required?

johnsaigle: Hmm I think we could add a `return false;` after line 371 for completeness. But I would still like to have the proper return type here. Before, the function would return void and true in different cases. At least now we can expect one type.

PapillonMcGill: Ok, keep it to true in that case.

kongtiaowang:
screen shot 2018-10-15 at 11 15 10 am
screen shot 2018-10-15 at 11 14 51 am
It will help you to sove the Travis issue.

php/libraries/Database.class.inc
zaliqarosli: @johnsaigle other than resolving this ^ issue and travis, i think everything else looks good. will test in a sec

zaliqarosli: apache php7.2 upgrade not working for me so can't test at the moment

johnsaigle:
@driusan This PR is finally passing!

php/libraries/Database.class.inc
```diff
@@ -930,20 +964,27 @@ class Database
...
-     * @param array  $dataArray The array representing the WHERE condition to be
-     *                          glued together into a string
-     * @param array  $exec_vals The values which are being bound to the query. Used
-     *                          to generate the prepared variable name.
-     * @param string $prefix    A prefix to apply to prepared variable names.
-     *
-     * @return string A string that can be used to generate a prepared statement with
-     *                appropriate variable names generated for data binding.
+     * @param string     $glue      The glue used to combine parts of the
+     *                          dataArray (ie " AND ")
```
```suggestion
     *                          dataArray (ie " AND ").
```

php/libraries/Database.class.inc
```diff
@@ -930,20 +964,27 @@ class Database
...
+     * @param array      $dataArray The array representing the WHERE condition
+     *                          to be glued together into a string
```
```suggestion
     *                          to be glued together into a string.
```
To keep consistancy with other line of this doc block

php/libraries/Database.class.inc
```diff
@@ -930,20 +964,27 @@ class Database
...
+     * @return string A string that can be used to generate a prepared statement
+     *                with appropriate variable names generated for data
+     *                binding.
      */
     private function _implodeAsPrepared(
-        $glue, $dataArray, &$exec_vals = null, $prefix=''
-    ) {
+        string $glue,
+        array $dataArray,
+        array &$exec_vals = null,
```
driusan: This is nullable according to the doc bloc (and default value)

php/libraries/Database.class.inc
johnsaigle: IMO we should go ahead and merge this. I think the nullable strings thing is likely a quirk of PHPUnit. In any case if any of the params is null then the DB will error out anyway due to invalid credentials.

PR #3880, opened 2018-08-15
[Core: Database] Add exception on fetchAll failure case

johnsaigle:
### Brief summary of changes

In an effort to take advantage of PHP's type system, we should add type declarations to our code, especially in the core classes.

The [PDO documentation](https://secure.php.net/manual/en/pdostatement.fetchall.php) says that `fetchAll` returns the empty set when no data in the DB matches a given query and FALSE on failure. However we do not distinguish this in our codebase and so a FALSE value is allowed to bubble up through subsequent DB calls.

This makes it so that these functions must sometimes return the Bool type (in addition to String, Array, and Null) meaning we cannot use return type declarations.

Following a discussion in #3854, this PR adds an exception when the `fetchAll` class fails. This will allow us to remove Bool as one of the return types. 

This is likely to break some things (I found it by debugging a failing test in #3878) so it is issued to `major`.

### Related issues

- https://github.com/aces/Loris/pull/3854
- https://github.com/aces/Loris/pull/3878

### DIFF
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -635,10 +635,10 @@ class Database
         }
 
         $rows = $prepared->fetchAll(PDO::FETCH_ASSOC);
-        // The fetchAll function will return an empty set if no results are 
+        // The fetchAll function will return an empty set if no results are
         // found and FALSE on "failure". We don't expect this case so an
         // exception is thrown.
-        if ($rows === false) {
+        if (is_bool($rows)) {
             $err = $prepared->errorInfo();
             throw new DatabaseException($err[2], $prepared->queryString, $params);
         }
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -635,6 +635,13 @@ class Database
         }
 
         $rows = $prepared->fetchAll(PDO::FETCH_ASSOC);
+        // The fetchAll function will return an empty set if no results are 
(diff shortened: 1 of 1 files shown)


php/libraries/Database.class.inc
```diff
@@ -635,6 +635,13 @@ class Database
         }
 
         $rows = $prepared->fetchAll(PDO::FETCH_ASSOC);
+        // The fetchAll function will return an empty set if no results are
+        // found and FALSE on "failure". We don't expect this case so an
+        // exception is thrown.
+        if (is_bool($rows)) {
```
driusan: this isn't checking if it's false, it's checking if it's a boolean. It should be `=== false`

johnsaigle: I tried that at first but `phan` gave me an error about comparing an array to a boolean (in the case where it was NOT false).

I'll change this statement to include both and see how phan feels about it.

driusan: the docs pretty clearly say that the return type is `array|false`, so if phan is complaining about checking the false case it sounds like a bug in phan.

xlecours: How about `!is_array()` ?

johnsaigle:
@driusan @xlecours This is ready for re-review.

johnsaigle:
@xlecours Yeah a lot of the DB functions return many many different types which makes validation and use of them kinda tricky. This is one small step in the process....

PR #3514, opened 2018-02-27
Interface for classes that handle file uploads

nicolasbrossard:
This PR is the result of a request to break https://github.com/aces/Loris/pull/3476 in two. The first part (this PR) is a proposal for an interface that could be used by all classes handling files uploaded on a server. I'll leave https://github.com/aces/Loris/pull/3476 open for now but close it once we merge this.

PR #2953, opened 2017-07-17
[Security] Correct occurrences of (needle,haystack) to (haystack,needle) in strpos file validation

johnsaigle:
This pull request fixes several errors where code used to prevent path traversal was incorrectly using _needle-haystack_ syntax instead of _haystack-needle_ for the `strpos` function.

In order to prevent this occurring in the future, I strongly suggest that we merge #2918 and use the `resolvePath()` function for the prevention of path traversal. However, this PR addresses the immediate problem and should provide some hardening until the 2918 is merged and can be applied throughout the codebase.

The below changes were made based on executing `grep -nr "strpos(\".." .` in the LORIS root and modifying each of the results (excluding the `.vendor` results). 

Broken access control has been tested and confirmed on `GetStatic.php`, `GetJs.php` and `data_release`.

See also: 
* #2950 which addresses the same error in `GetStatic.php` and inspired this PR
* [strpos syntax](http://php.net/manual/en/function.strpos.php)
* #2918 for a more robust was of preventing path traversal

PR #2948, opened 2017-07-13
[Security!] Full file system access security vulnerability fix

Jkat:
how2codephp

PR #2876, opened 2017-06-14
[Security] Amend earlier download fix

johnsaigle:
This pull request restores the broken functionality merged in #2868. This fixes the access control issue while allowing valid files to download.

See also: Redmine 12586.

PR #3881, opened 2018-08-15
[Core, Media] Interface for safe file downloading

johnsaigle:
### Summary of changes

The idea is that **all** file downloading in LORIS will be as short as the new `GetFile` code with baked-in validation and security.

#### Details

There are many places in LORIS where file downloading has been implemented in an ad-hoc way. While functional, these implementations are coded with different methodologies and styles making it very time-consuming to test from a security perspective. At the same time, many of these implementations have been copied partially from each other and so share(d) security vulnerabilities.

This PR adds:

* The `AttachmentResponse` class suitable for doing straight-forward file downloads in the majority of cases
* An extension of the above class for use in the Doc Repo to demonstrate how to use it.
* A small refactor to the doc_repo file downloader as it has some quirks but is simple enough to illustrate the changes. 

### This resolves issue...

Consolidates common functionality in one place so that security and basic verification can be handled in a universal way.

Use an OO approach to allow individual classes to extend and different implement retrieval function to allow flexibility where needed.

Abstracts the following boilerplate code from file downloads:
* Checking permissions
* Checking file exists
* Checking file is a file
* Attach the appropriate headers (mime type, length, file attachment)
* **Ensuring no path traversal**

#### Related PRs:

Similar idea:
* https://github.com/aces/Loris/pull/3514

Related security issues 
* https://github.com/aces/Loris/pull/2953
* https://github.com/aces/Loris/pull/2948
* https://github.com/aces/Loris/pull/2876

### Further work

* Refactor existing modules to use (or extend) AttachmentResponse.

* Future modules requiring attachment downloads should also use AttachmentResponse

### To test this change...

- [ ] Ensure that you are able to download files from the doc_repo.
- [ ] Make sure the download dialog contains accurate information about the content length and MIME type


johnsaigle:
This PR is working and ready for code review even if it's failing Travis! 

I just need to do PHPCS and function documentation.

modules/document_repository/ajax/GetFile.php
```diff
@@ -13,48 +13,33 @@
...
+$downloadBasePath = $lorisRoot . 'modules/document_repository/user_uploads/';
+$downloader       = new DocRepoFileDownloader($downloadBasePath);
+
+// Format: username/filename.ext
+$partialPath = $_GET['File'];
+
+// Verify file exists in the database.
+if (!$downloader->isFileInDatabase($partialPath)) {
+    error_log("ERROR Requested file is not in the database");
+    http_response_code(400);
```
driusan: a file not being there should be a 404

php/libraries/FileDownloader.class.inc
```diff
@@ -0,0 +1,21 @@
...
+     * If the location of a file is stored from a database, the class
+     * implementing this function will need to specify how that is done.
+     */
+    function getFilenameFromDatabase(string $identifier);
+
+    /**
+     * We must be sure that a relative path is not being used maliciously to
+     * download a forbidden file.
+     */
+    function sanitizePath(string $fullPath);
```
driusan: I don't think this should be part of the interface, it should be part of the implementation.

php/libraries/GenericDownloader.class.inc
```diff
@@ -0,0 +1,155 @@
...
+ *  @author   John Saigle <john.saigle@mcgill.ca>
+ *  @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ *  @link     https://www.github.com/aces/Loris
+ */
+
+require_once '../php/libraries/FileDownloader.class.inc';
+/**
+ * This serves as a generic base class for other modules in LORIS to extend as
+ * needed. It implements the FileDownloader interface to provide core download
+ * functionality and security. It does not actually provide any functiaonlity
```
driusan: s/functiaonlity/functionality/

php/libraries/GenericDownloader.class.inc
```diff
@@ -0,0 +1,155 @@
...
+ * related to retrieiving additional file information from the database as in
```
driusan: s/retrieiving/retrieving

php/libraries/GenericDownloader.class.inc
```diff
@@ -0,0 +1,155 @@
...
+        if (!file_exists($fullPath)) {
+            error_log("ERROR: File requested for download does not exist.");
+            http_response_code(404);
+            return;
+        }
+
+        // Get mime type and do the file download.
+        $mimeType = mime_content_type($fullPath);
+        header("Content-type: $mimeType");
+        header("Content-Disposition: attachment; filename=$fileName");
```
driusan: There are the responsibility of the PSR7 handler. Maybe the interface should have a `GetContentType(string $filename) : string` (or similar) method to make it possible for a real middleware to inject the header.

php/libraries/GenericDownloader.class.inc
```diff
@@ -0,0 +1,155 @@
+<?php
+/**
+ * Class to provide "middleware" for file downloading functionality, taking care
```
driusan: This isn't middleware, it's just a class.

php/libraries/GenericDownloader.class.inc
```diff
@@ -0,0 +1,155 @@
...
+        // Validate that the path supplied by the user begins with the same base
+        // path as the proper directory of the file.
+        // If not, it is almost certainly a hacking attempt.
+        if ($realUserPath === false
+            || strpos($realUserPath, $this->realBasePath) !== 0
+        ) {
+            error_log("$realUserPath equals $this->realBasePath");
+            error_log("ERROR: Likely path traversal attempt in file download.");
+            http_response_code(400);
+            exit(1);
```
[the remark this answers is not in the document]
johnsaigle: Existing download implementations in LORIS exit which is why I did it here.

I agree with throwing an exception instead.

johnsaigle:
@driusan Thanks for all the feedback, it's all very helpful and I appreciate the perspective. I'll get to fixing this ASAP.

php/libraries/GenericDownloader.class.inc
```diff
@@ -0,0 +1,155 @@
...
+     * Downloads a file from the filesystem.
+     *
+     * @param string $userPath The full path supplied by the calling program.
+     *                          This path can be unsanitized (i.e. can accept
+     *                          raw request data) as sanitization is done
+     *                          automatically.
+     *
+     * @return none
+     */
+    public function downloadFile(string $userPath): void
```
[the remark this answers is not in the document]
johnsaigle: I don't think we actually have a separate file-access permission layer for LORIS front-end users in any case I can think of. Instead there is a combination of the module permissions and the UNIX permissions. 

Effectively if you are able to access the module then you are able to access the files in the module. Any variation in this is done in an ad-hoc way by the modules.

Alternatively, we could add a function to the interface where a class must implement LORIS-user-level permission checking per module. What are your thoughts?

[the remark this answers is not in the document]
johnsaigle: Agreed on the StreamInterface bit.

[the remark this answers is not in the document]
johnsaigle: Going to use this to work off of:
https://github.com/aces/Loris/blob/24e633a7eda29ef5578830e82d88bdc0fd84ff7b/src/Router/ModuleFileRouter.php#L84

Let me know if this makes sense

php/libraries/GenericDownloader.class.inc
```diff
@@ -0,0 +1,155 @@
...
+     * Constructor for GenericDownloader.
+     *
+     * @param string $basePath The full path to the directory that will serve
+     *                              the files for download.
+     *
+     * @return none
+     */
+    function __construct(string $basePath)
+    {
+        $this->realBasePath = realpath($basePath);
```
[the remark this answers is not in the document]
johnsaigle: I believe you're right but I was unable to find any place in the LORIS core where we create symlinks.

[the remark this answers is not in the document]
driusan: The symlinks wouldn't be created from LORIS, they would be created by someone (ie. a sysadmin) intentionally trying to point to a different location without changing LORIS (ie. a different mount point)

johnsaigle:
Still todo:

* Use middleware to determine the mime type.

* Make sure symlinks will work after path sanitization

php/libraries/GenericDownloader.class.inc
```diff
@@ -0,0 +1,156 @@
...
+    {
+        if (!is_file($fullPath)) {
+            $pathForPrinting = htmlentities($fullPath);
+            throw new LorisException(
+                "Cannot get mimetype of $pathForPrinting because it is not a "
+                . "file!"
+            );
+            return '';
+        }
+        return mime_content_type($fullPath);
```
PapillonMcGill: This should also use sanitized path. If not, it could be used to query and probe for existence of specific file

PapillonMcGill: still the line 126 is not using the sanitized path.

PapillonMcGill: Nope, it will raise an exception if the file doesn't exist and log an error if the file contain ".." but will no break in that case

johnsaigle: Hmm yeah you're right it works differently after I made the changes to allow for sym links. This needs more work

johnsaigle:
@PapillonMcGill I did a lot of refactoring and I believe I addressed your concerns.

johnsaigle:
@driusan I believe I've incorporated all the feedback you gave when I first issued this PR including:

* Simplifying the interface (I removed it)
* Preventing an issue where my method would prevent symlinks from working
* Using PSR stuff to determine MIME type

As to that last point, I'm not sure if I've really done best practices here because I am still learning how PSR/Diactoros works. I'd really appreciate some guidance here.

modules/document_repository/ajax/GetFile.php
```diff
@@ -13,48 +13,47 @@
...
+
+// Format: username/filename.ext
+
+$partialPath = $_GET['File'] ?? null;
+if (is_null($partialPath)) {
+    http_response_code(400);
+    echo "Bad request. A valid path must be specified.";
+} else {
+    $downloadBasePath = $lorisRoot . 'modules/document_repository/user_uploads/';
+    $fullPath = $downloadBasePath . $_GET['File'];
```
PapillonMcGill: $fullPath is never used.

driusan:
@johnsaigle sorry I missed your comment until now.. did you forget to push? It still appears to be using side-effects instead of a PSR7 response.

ridz1208:
@johnsaigle any further development on this one ?

johnsaigle:
@ridz1208 Yes I have resumed work on this today and have been doing a rewrite according to Dave's suggestions

johnsaigle:
@driusan I'm making an attempt to make the changes you wanted. I have used #3977 as a model for how to handle the request for file downloading as per the instructions in the Routing docs. Also I have chosen to switch the proof-of-concept to the `media` module as doc repo is being reactified right now. 

I am having trouble visiting this endpoint from the LORIS front-end. I get a 404 Loris Path error, e.g.

> File not found: https://jsaigle-dev.loris.ca/mediadownload?lorispath=media/mediadownload/

Curiously I also get this for the endpoint in #3977 when I try to visit it. Any ideas how I can get this response to be delivered? I'm stuck on this error being generated from the Module class.

johnsaigle:
Will resume development on this after #4163 gets merged. Otherwise it's too painful because every error gets swallowed by the 404 handling.

driusan:
#4163 was merged.

johnsaigle:
@driusan Yep, just been busy with other work and this is a slightly lower priority.

johnsaigle:
This will be rewritten to closely conform to the structure in #4181

johnsaigle:
To those assigning PRs (@ridz1208, @driusan): I'm going to hold off developing this class until #4181 is approved and merged. In my view these classes should conform to each other as closely as possible and I'm happier with the approach I took there than here. 

Feel free to close this PR if you're looking to organize the repo. We can re-open it when the other is merged. If you don't want to do that, rest assured that I have not forgotten about this PR and will be working on it as soon as the other one goes through.

driusan:
Superceded by #4181

PR #3882, opened 2018-08-16
[tools] Deprecate excelDump script

driusan:
This removes the excelDump.php from the tools directory and adds it to deprecated, since it's known to not work with large datasets and results in a warning from composer about using a deprecated library every time you install/update LORIS.

Since the excelDump was the only thing using the library, the dependency is also removed from LORIS's composer.json/lock file.


Jkat:
It is working fine for PreventAD (though we made specific work arounds for specific large datasets), but if no one else is using it then sure this can be deprecated in the main repo.

driusan:
What changes did you have to make "for specific large datasets" in PreventAD? In IBIS I had to port/rewrite it to Go to get acceptable performance.

cmadjar:
Not sure I would deprecate this since currently, it is the only tool that allows you to have data releases of your database. I would try to fix it instead so that it works on larger datasets. Just my two cents on the subject. :)

Jkat:
iirc I think I wrote custom queries and/or split the data across different files...Nothing fancy but it works, and it's probably hard for researchers to work with excessively large excel files anyway.

@cmadjar that is a good point.

I just did a quick search and maybe someone should look into https://github.com/PHPOffice/PhpSpreadsheet and if it is a feasible replacement, we can comfortable deprecate(or replace really) the current PHPExcel library

PapillonMcGill:
I used PHPOffice in a previous work, was easy to work with and offered all that I needed, but it might not be the case here.

driusan:
@cmadjar The excelDump script as it exists already raises a giant "deprecation" error because of a deprecated third party library dependency that we have no control over. This doesn't remove the script for people who want to use it, it just moves it to the deprecated folder. A complete rewrite is a different issue (but I don't think PHP is a very suitable language for working with things like large excel files anyways.)

PR #3883, opened 2018-08-16
Making lookupCenterNameUsing configuration field a drop down

cmadjar:
This pull request is making the 'lookupCenterNameUsing' configuration under the 'Imaging Pipeline section" of the config module a drop down instead of an open text field. The options of the drop down: PatientID, PatientName
screen shot 2018-08-16 at 5 08 12 pm

## How to test this PR:

1. check that this field is indeed an open text field on minor
2. pull the code, run the SQL patch and check that the field is now a drop down
3. make sure the saving is working properly and saving the proper value

See also: Redmine ticket https://redmine.cbrain.mcgill.ca/issues/15104

### DIFF
--- a/SQL/0000-00-03-ConfigTables.sql
+++ b/SQL/0000-00-03-ConfigTables.sql
@@ -17,7 +17,7 @@ CREATE TABLE `ConfigSettings` (
     `Description` varchar(255) DEFAULT NULL,
     `Visible` tinyint(1) DEFAULT '0',
     `AllowMultiple` tinyint(1) DEFAULT '0',
-    `DataType` ENUM('text', 'boolean', 'email', 'instrument', 'textarea', 'scan_type') DEFAULT NULL,
+    `DataType` ENUM('text', 'boolean', 'email', 'instrument', 'textarea', 'scan_type', 'lookup_center') DEFAULT NULL,
     `Parent` int(11) DEFAULT NULL,
     `Label` varchar(255) DEFAULT NULL,
     `OrderNumber` int(11) DEFAULT NULL,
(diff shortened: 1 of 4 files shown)


PapillonMcGill:
@cmadjar Could you tell me in which menu to look for that option. I am still not an expert with Loris. There is no mention in Redmine either.

Jkat:
Under 'Imaging Pipeline' in the Configuration module.

cmadjar:
@PapillonMcGill I updated the description of the PR to be a little more clearer. It is under the Imaging Pipeline section. I believe this config was added for the 20.0 release of LORIS.

SQL/0000-00-03-ConfigTables.sql
```diff
@@ -17,7 +17,7 @@ CREATE TABLE `ConfigSettings` (
     `Description` varchar(255) DEFAULT NULL,
     `Visible` tinyint(1) DEFAULT '0',
     `AllowMultiple` tinyint(1) DEFAULT '0',
-    `DataType` ENUM('text', 'boolean', 'email', 'instrument', 'textarea', 'scan_type') DEFAULT NULL,
+    `DataType` ENUM('text', 'boolean', 'email', 'instrument', 'textarea', 'scan_type', 'lookup_center') DEFAULT NULL,
```
driusan: I don't like the idea of needing to add a new specific entry to this table for every single dropdown. Maybe there should be an 'enum' type which looks into a normalized table for the labels to display in a dropdown?

cmadjar: @driusan agree with you. I did not like that idea either but this is the current behaviour of the config module. Maybe this could be tackle in a separate PR? Not sure it is in the scope of this PR. In which case, I can create a redmine ticket for that if none exist already. What do you think?

PR #3551, opened 2018-03-16
[Database] modify the _implodeWithKeys() function to allow IS NULL in query

PapillonMcGill:
This pull request `let IS NULL value in the WHERE clause`. It `convert null value in $param array to IS NULL in the SQL query`.

required for an upcoming PR

Step to test. 
create a test file to call implodeWithKeys directly with an associative array containing NULL as a value as no code currently in Loris do it and ensure that the query generated by the prepared statement contain and "IS NULL" for that value. For that, you will need to rename the _implodeWithKeys to implodeWithKeys and make it public.

PR #3887, opened 2018-08-21
[core Visit] Normalization of Visit 

PapillonMcGill:
### Brief summary of changes
Introduce the Visit class in order to normalize visits.

#### Related PRs:
new version of https://github.com/aces/Loris/pull/3579 

### Further work
Future PR needed to modify tables to replace visit label with the visit.VisitID attributes as foreign key and code accordingly.


php/libraries/VisitController.class.inc
```diff
@@ -0,0 +1,130 @@
+<?php
+/**
+ * This file contains the VisitController class.
+ *
+ * PHP Version 7
+ *
+ * @category Controler
```
johnsaigle: Should be `Controller`. Same for below.

test/unittests/VisitTest.php
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Visit object use in tests
+     *
+     * @var Visit
+     */
+    private $_visit;
+
+    protected function setUp()
+    {
+/*        parent::setUp();
```
johnsaigle: Commented code still needed?

PapillonMcGill: not really, different version of try. will clean up when every thing is fine with unit test

PapillonMcGill:
@driusan Phan is stubborn, in php/librairies/Visit.class.inc, line 48-49, when declaring string $visit = null, phan request that I indicated the null possibility. Change for ?string $visit = null, wasn't happy either "not a permitted replacement of the nullable param ... '?T' should be documented as 'T|null' or '?T' " so I tried string|null $visit = null, still complain " unexpected '|' " for line 48. 

Do we skip phan test for that part or have you a solution that Phan will be pleased with?

driusan:
I don't know what you're trying to accomplish so I don't know what solution is best, but, no, we don't skip phan for anything. phan reports things that are guaranteed to be wrong via static analysis. Once we've cleaned up a class of errors we don't let it sneak back in.

`?string` should work as a type hint in PHP, while sum types don't exist in PHP function signatures, only in doc blocks, so you can only use `string|null` in the `@param` part of the doc, not in the code. The first error sounds like it's complaining that the doc block and the function signature declared in PHP don't match to me, which is a legitimate complaint for phan to make.

PapillonMcGill:
@driusan guest you were right, forgot to align both doc block and declaration. Rule to never forget, if the error is not on the reported line, look the previous lines.

SQL/0000-00-00-schema.sql
```diff
@@ -2041,3 +2041,24 @@ CREATE TABLE `candidate_consent_history` (
...
+  `VisitID` int(10) unsigned NOT NULL auto_increment,
+  `VisitName` varchar(100) NOT NULL,
+  CONSTRAINT `visit_PK` PRIMARY KEY (`VisitID`),
+  CONSTRAINT `visit_name_UK` UNIQUE KEY (`VisitName`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE TABLE `visit_project_subproject_rel` (
+ `VisitID` int(10) unsigned NOT NULL,
+ `ProjectID` int(2) NOT NULL,
+  `SubProjectID` int(10) unsigned NOT NULL,
```
driusan: The spacing here is different than the other rows in the table. Also, `Subproject` has a lowercase P. (SAme with all the other references to it.)

php/libraries/VisitController.class.inc
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /**
+     * Retruns a list of configured Visits
+     *
+     * @return array The list of visits
+     */
+    public function getAllVisits(): array
+    {
+        return array_map(
```
driusan: I think this whole thing can be pselectCol now that it exists.

php/libraries/VisitController.class.inc
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Retruns a VisitID by name
+     *
+     * @param string $visit the name of the visit
+     *
+     * @return array the ID of the visit  //TODO array of Visit
+     */
+    public function getVisitsWithID(string $visit): array
+    {
+        return array_map(
```
driusan: pselectCol

php/libraries/VisitController.class.inc
```diff
@@ -0,0 +1,130 @@
...
+            )
+        );
+    }
+
+    /**
+     * Retruns a VisitID by name
+     *
+     * @param string $visit the name of the visit
+     *
+     * @return array the ID of the visit  //TODO array of Visit
```
driusan: The TODO doesn't seem valid. (also it should be @todo if it's in a php docblock.)

PapillonMcGill: left over from previous developement

SQL/0000-00-00-schema.sql
```diff
@@ -2041,3 +2041,24 @@ CREATE TABLE `candidate_consent_history` (
   `EntryStaff` varchar(255) DEFAULT NULL,
   CONSTRAINT `PK_candidate_consent_history` PRIMARY KEY (`CandidateConsentHistoryID`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
```
ridz1208: missing Drop tables at top of document. Keep in ind the order of creation and dropping is important !!!

SQL/0000-00-00-schema.sql
```diff
@@ -2041,3 +2041,24 @@ CREATE TABLE `candidate_consent_history` (
...
+CREATE TABLE `visit` (
+  `VisitID` int(10) unsigned NOT NULL auto_increment,
+  `VisitName` varchar(100) NOT NULL,
```
ridz1208: what name is that ? something like "V03" or "The 3rd visit" i.e. is this the new place for the visit_label or the front end label.

I think last discussion we said that we need both (one for presentation on the front end, the other to be used in file names and API ...)

php/libraries/Visit.class.inc
```diff
@@ -0,0 +1,64 @@
+<?php
+/**
+ * Describe a LORIS visit
+ *
+ * PHP Version 7
+ *
+ * @category Model
+ * @package  Visits
+ * @author   Xavier Lecours Boucher <xavier.lecoursboucher@mcgill.ca>
```
ridz1208: I dont think so

php/libraries/Visit.class.inc
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+namespace LORIS;
+
+/**
+ * Representation of a Visit in LORIS
+ *
+ * @category Model
+ * @package  Visits
+ * @author   Xavier Lecours Boucher <xavier.lecoursboucher@mcgill.ca>
```
ridz1208: I dont think so

php/libraries/Visit.class.inc
```diff
@@ -0,0 +1,64 @@
...
+     * @var int
+     */
+    protected $id;
+
+    /**
+     * The visit name (formely known as visit_label)
+     *
+     * @var string
+     */
+    protected $visit;
```
ridz1208: maybe call the variable `$name` ?? 

the `$visit` variable in the Visit class is not useful

php/libraries/VisitController.class.inc
```diff
@@ -0,0 +1,130 @@
...
+                 v.VisitName as "visit"
+                FROM
+                 visit v
+                ',
+                array()
+            )
+        );
+    }
+
+    /**
```
ridz1208: I'm having trouble understanding what this function does.

**with** keyword in the function name suggests that you are adding and ID to the visit object but instead you are querying from the database and retuning a whole new visit with only the ID and not the name... so thats a bit confusing. I'm not sure visits should be in plural, we do assume that the visit name is unique (maybe throw an exception if more than one result exist) and finally the parameter name $visit is not informative, $visitName would be much clearer

it seems like pselectOne is much cleaner than what is currently queried and if you follow my advice above, you can use pselectRow and return a visit oject with an ID and a Name because thats what the visit is, returning just the ID i na visit object having supplied the name is not ideal

php/libraries/VisitController.class.inc
```diff
@@ -0,0 +1,130 @@
...
+                FROM
+                 visit v
+                WHERE
+                 v.VisitName = :name
+                ',
+                array('name' => $visit)
+            )
+        );
+    }
+    /**
```
ridz1208: sismilar issues with the above function, first it reads "withProject" while project is never really supplied ? this function to me should be more like `getVisits` or maybe `getVisitsAssociation` or `getVisitsForAllProjects`

when i see **with** i'm expecting something like 
`$newVisit` = $oldVisit->withName("Rida")` where rida will be set as the $oldVisit's name and returned into a $newVisit object

SQL/0000-00-00-schema.sql
PapillonMcGill: Actually no, I was ask to remove the front end label from the table as it was not need at this moment.

I can add it back @driusan if requested.

php/libraries/Visit.class.inc
PapillonMcGill: Give back to Cesar what belong to Cesar. Xavier wrote this part

php/libraries/Visit.class.inc
PapillonMcGill: Same here.

php/libraries/VisitController.class.inc
PapillonMcGill: As the name of the fonction imply, it return a list of all the visit configure in the system as object of type visit.

driusan: I don't understand what you mean. Where do you see a "with" keyword?

SQL/0000-00-00-schema.sql
ridz1208: resolved

SQL/0000-00-00-schema.sql
ridz1208: i retract my comment then.

I understand that the current visitName is the the current visit_label

php/libraries/Visit.class.inc
ridz1208: sorry @xlecours

php/libraries/Visit.class.inc
ridz1208: sorry @xlecours again

php/libraries/VisitController.class.inc
ridz1208: `getVisitsWithID` WITHid

zaliqarosli:
this PR should have the `Blocked` label on it until https://github.com/aces/Loris/pull/3551 gets merged?

zaliqarosli:
@PapillonMcGill in that case, you should update the PR description with what's current

PapillonMcGill:
@ridz1208 @driusan Could you re-review?

SQL/0000-00-00-schema.sql
```diff
@@ -2041,3 +2043,24 @@ CREATE TABLE `candidate_consent_history` (
...
+  CONSTRAINT `visit_PK` PRIMARY KEY (`VisitID`),
+  CONSTRAINT `visit_name_UK` UNIQUE KEY (`VisitName`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE TABLE `visit_project_subproject_rel` (
+ `VisitID` int(10) unsigned NOT NULL,
+ `ProjectID` int(2) NOT NULL,
+ `SubprojectID` int(10) unsigned NOT NULL,
+  CONSTRAINT `visit_project_subprojet_rel_PK` PRIMARY KEY (`VisitID`, `ProjectID`, `SubprojectID`),
```
driusan: subproject is spelled wrong.

SQL/New_patches/2018-01-17_normalisation_visit.sql
```diff
@@ -0,0 +1,25 @@
...
+  `VisitName` varchar(100) NOT NULL,
+  CONSTRAINT `visit_PK` PRIMARY KEY (`VisitID`),
+  CONSTRAINT `visit_name_UK` UNIQUE KEY (`VisitName`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE TABLE `visit_project_subproject_rel` (
+  `VisitID` int(10) unsigned NOT NULL,
+  `ProjectID` int(2) NOT NULL,
+  `SubprojectID` int(10) unsigned NOT NULL,
+  CONSTRAINT visit_project_subprojet_rel_PK PRIMARY KEY (`VisitID`, `ProjectID`, `SubprojectID`),
```
driusan: Same here

PapillonMcGill: corrected, keep using the french spelling.

php/libraries/VisitController.class.inc
```diff
@@ -0,0 +1,133 @@
...
+     * Returns a list of configured Visits and their association with projects
+     * and subprojects
+     *
+     * @return array The list of visits, projectID and subprojectID
+     */
+    public function getVisitsAndProject(): array
```
ridz1208: Minor comment, how about `getVisitsProjectsSubprojects()` given that that's actually what it does and the `And` is not really required its kinda implioed...

otherwise its all good for me

ridz1208:
manual testing only on the patches and instanciation of the object

SQL/New_patches/2018-01-17_normalisation_visit.sql
```diff
@@ -0,0 +1,25 @@
+CREATE TABLE `visit` (
+  `VisitID` int(10) unsigned NOT NULL auto_increment,
+  `VisitName` varchar(100) NOT NULL,
+  CONSTRAINT `visit_PK` PRIMARY KEY (`VisitID`),
+  CONSTRAINT `visit_name_UK` UNIQUE KEY (`VisitName`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE TABLE `visit_project_subproject_rel` (
```
ridz1208: @PapillonMcGill getting a foreign key error on this table

ridz1208: nvm, engine issue on my end

PR #3888, opened 2018-08-21
[DQT] DQT Download files Message

intralizee:
This pull request provides a message to the user (**when no data is available to download**) by the user clicking on the "Download Data as Zip" button. Also removed some console output. Note: Plan to rewrite the module and this is just a quick fix for the current DQT.

See also: https://redmine.cbrain.mcgill.ca/issues/10453


johnsaigle:
@intralizee Can you edit the description for the PR? I think you are saying here that you have added an error message to the DQT but it's not clear how to test this change.

intralizee:
@johnsaigle Added more details on how to test.

modules/dataquery/jsx/react.app.js
```diff
@@ -103,7 +103,7 @@ var DataQueryApp = React.createClass({
         var promises = [];
         var that = this;
         for(var key in this.state.queryIDs){
-            console.log(this.state.queryIDs[key][0]);
+            //console.log(this.state.queryIDs[key][0]);
```
paiva: You are keeping the console log because this is just a quick fix and will be removed when the module is rewritten, right?

driusan:
@paiva Did you test it?

modules/dataquery/jsx/react.app.js
zaliqarosli: @intralizee or could we remove this comment? if not, is there a reason we want to keep it?

paiva:
@driusan Sorry, yes, I tested it. Works.

PR #3890, opened 2018-08-21
[Instruments] Add ability to load instrument JSON data from flag Data column.

driusan:
LORIS previously saved instrument data to a "Data" column
in the flag table as a test, this change gives users the
option to experiment with using it. Loading data from the
Data column means that the instrument SQL tables are no
longer required, but should be considered experimental and
not ready for production.

To test this: set the "$jsonData" variable in an instrument
to true, and try saving and loading and generally using the
instrument. There should be no noticeable difference. If
you're feeling lucky, drop the existing table and continue
trying to save and load data from the instrument.

Note that it might be easier to use a new instrument, since
scoring functions in PHP instruments are often intertwined
with database access.

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1445,13 +1445,11 @@ class NDB_BVL_Instrument extends NDB_Page
             return 'Complete';
         }
 
-        $db =& Database::singleton();
-
         $allData = $this->loadInstanceData($this);
         foreach ($this->_requiredElements as $field) {
             // this shouldn't be just empty() b/c 0 is a valid
             // value for some required fields
-            if (is_null($field) || $field === "") {
+            if (is_null($allData[$field]) || $allData[$field] === "") {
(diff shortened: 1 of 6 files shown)


php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2022,12 +2025,21 @@ class NDB_BVL_Instrument extends NDB_Page
      */
     static function loadInstanceData($instrumentInstance)
     {
-        $db       =& Database::singleton();
-        $defaults = $db->pselect(
-            "SELECT * FROM $instrumentInstance->table WHERE CommentID=:CID",
-            array('CID' => $instrumentInstance->getCommentID())
-        );
-        return $defaults[0];
+        $db = Database::singleton();
```
driusan: This part is the main bit of the change, the rest is just so that I could still load/save after dropping the table.

driusan:
For some reason I can't change this to "minor" on the GitHub frontend after rebasing it, so I'm just closing it and making a new one..

PR #3891, opened 2018-08-21
[JS] Remove all the compiled artifacts

driusan:
Now that there is a build step incorporated into installing
LORIS to generate the VERSION file, there's no longer any need
to directly track the compiled javascript files in git (which
is a recipe for out of date files and conflicts.)

This removes all of the compiled JSX artifacts in order to
make life easier for developers.


intralizee:
Hey @driusan if a developer wants to include a JS file that is not react but just some custom functions for utility. The js file will still go into the JS directory but we modify the .gitignore file?

driusan:
@intralizee You could either do that or just put it in the jsx folder and transpile it as a no-op since js is a subset of jsx.

driusan:
Added blocked because I don't want to risk conflicts with #3895 and I'd rather that one goes in first.

PR #3894, opened 2018-08-22
[Instruments] Add ability to load instrument JSON data from flag Data column.

driusan:
LORIS previously saved instrument data to a "Data" column
in the flag table as a test, this change gives users the
option to experiment with using it. Loading data from the
Data column means that the instrument SQL tables are no
longer required, but should be considered experimental and
not ready for production.

To test this: set the "$jsonData" variable in an instrument
to true, and try saving and loading and generally using the
instrument. There should be no noticeable difference. If
you're feeling lucky, drop the existing table and continue
trying to save and load data from the instrument.

Note that it might be easier to use a new instrument, since
scoring functions in PHP instruments are often intertwined
with database access.


php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2081,8 +2089,7 @@ class NDB_BVL_Instrument extends NDB_Page
         $CommentID = $this->getCommentID();
 
         $pscid = $db->pselectOne(
-            "SELECT c.PSCID FROM $this->testName i
```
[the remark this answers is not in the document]
driusan: it's not even required now, it's an unnecessary join on `$this->testName` to get a column out of the candidate table (which is linked to the flag table via the session, not the testName table.)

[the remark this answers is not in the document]
ZainVirani: gotcha :+1:

sruthymathew123:
@driusan I tsted this PR on a normal IBIS instrument by adding "$jsonData = true;" to that instrument. I dropped/renamed the table associated with the instrument and the instrument still loaded fine. But was unable to save the data.

sruthymathew123:
@driusan I removed the IBIS override on NDB_BVL_Instrument.class.inc

sruthymathew123:
@driusan Loading is totally fine. But when trying to save the instrument, I am getting this error.

saving_json

driusan:
@sruthymathew123 Can you request changes to this doesn't show up as approved?

ridz1208:
We should also have a script that confirms that all SQL data is in the JSON as part of this PR or a follow-up within the same release

ridz1208:
Tested on a very simple instrument, works flawlessly

Tested on a relatively complex instrument. there is an obvious issue with the scoring fields since they are saved seperately (in the case of CCNA at least) i.e. the scoring field values are saved using a `$DB->insert('table',$scores,$commenID)` call. I'm wondering if we can convert these to a `$this->_save($scores);` trivially and if so would that then work with this PR... 

until I can investigate that I cant set this to passed manual tests but it is promising

ridz1208:
@sruthymathew123 we need to talk about this one for IBIS

ridz1208:
blocked due to possible backwards compatibility issue

driusan:
@ridz1208 I'm fine with changing this to major, but what's backwards incompatible? Everything should be guarded by the "protected $jsonData = false;" default unless an instrument explicitly opts in.

driusan:
The decision on which branch this goes to has changed again, this should be rebased onto major..

driusan:
Rebased, but there were some conflicts, so it needs re-testing.

driusan:
After looking at this with @sruthymathew123, it's failing because getCandidateAge is doing a select statement on line 2071 which needs to intelligently look into either the table or the flag column based on whether it's a JSON instrument.

driusan:
`getFieldValue()` will also need to be updated.

sruthymathew123:
@driusan @ridz1208 Now I am able to load and save the instrument as JSON data.

ridz1208:
@driusan we still need the couchdb scripts in this PR

driusan:
@ridz1208 I think the CouchDB update should go into a future PR. This feature is still experimental and users need to opt-in in order to use it.

ridz1208:
@driusan my only issue with the CouchDB is that CCNA will not be able to migrate any instruments before the data can be exported...

HOWEVER, I'm not sure if it is the case here or not. If the data is still being saved into the SQL AND the JSON but loaded from the JSON only... then the old couchdb scripts would still work. if it is the case... then I dont mind it that much for now

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -746,11 +750,15 @@ class NDB_BVL_Instrument extends NDB_Page
             $values = Utility::nullifyEmpty($values, $key);
         }
         $db = \Database::singleton();
-        $db->update(
-            $this->table,
-            $values,
-            array('CommentID' => $this->getCommentID())
-        );
+        if ($this->jsonData !== true) {
```
ridz1208: I would suggest putting this in a `try{}catch` rather then bypassing completely until everything is JSON ready (ie CouchDB scripts, get/set dataEntryCompletionStatus, all all other getters in the class still referring to $this->testname)

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -746,11 +750,15 @@ class NDB_BVL_Instrument extends NDB_Page
...
+            // table may not exist, so don't try and update it.
+            $result = $db->update(
+                $this->table,
+                $values,
+                array('CommentID' => $this->getCommentID())
+            );
+        }
 
         // Extract the old data and merge it with what was submitted so that we
         // don't overwrite data from other pages.
```
ridz1208: this comment is for line 771

The current SQL column type is [TEXT](https://github.com/aces/Loris/blob/111bb660c9695cb4e5c6937323f677933ade0477/SQL/0000-00-00-schema.sql#L400)... its strange that a NULL text field is returning a "null" string. I would think it's an issue with saving where "null" is being submitted instead of NULL in the update

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1509,11 +1507,15 @@ class NDB_BVL_Instrument extends NDB_Page
 
         $db = \Database::singleton();
 
-        $db->update(
-            $this->table,
-            array('Data_entry_completion_status' => $status),
-            array('CommentID' => $this->getCommentID())
-        );
+        if ($this->jsonData !== true) {
+            // FIXME: Is this even required at all? If it is, we need to
```
ridz1208: this is definitely still needed it's the only way an instrument can be set to complete

tools/migrate_sql_to_json.php
```diff
@@ -0,0 +1,50 @@
+<?php
+/**
+ * Script exporting Update statements to remove 0000-00-00 values
```
ridz1208: that doesn't sound right

tools/migrate_sql_to_json.php
```diff
@@ -0,0 +1,50 @@
+<?php
```
ridz1208: please move into the `tools/single_use` directory

driusan: is it single use, though? It will be used multiple times for different instruments.

ridz1208: I mean it's not meant to be reused for the same instrument...

php/libraries/NDB_BVL_Instrument.class.inc
driusan: It is strange, but it's the behaviour I'm was getting and I don't see what alternative there is to having that check.

php/libraries/NDB_BVL_Instrument.class.inc
driusan: I considered that when I was writing it, but I don't think a try/catch block is a good idea. It forces a query that's known to fail to be run for people who don't want it and will hide issues/errors from places that are supposed to be changed, rather than complaining loudly so that issues are caught right away.

php/libraries/NDB_BVL_Instrument.class.inc
driusan: In that case, can we move it to the flag table?

ridz1208:
Discussion: how much of the instrument infrastructure needs to be converted to the new way (JSON) for this PR to be mergeable ?

 - couchdb imports
 - ...

tools/CouchDB_Import_Instruments.php
```diff
@@ -104,23 +104,18 @@ function generateDocumentSQL($instrument, $JSONData)
                         f.CommentID NOT LIKE 'DDE%' 
                         AND s.Active='Y' AND c.Active='Y'";
 
-        if ($JSONData) {
-            // the data is in the flag table, add the data column to the query
+        if ($tablename === "") {
```
ridz1208: Why depend on a string being empty or not when you have a more "explicit" boolean telling you what query you should run here? any gain ?

that said I agree with the changes below in the `UpdateCandidateDocs` functions.

driusan: one is directly derivable from the other, passing them both gives the possibility of them contradicting each other and I don't really find that the boolean is any clearer.. 
if there's no table to join get data from the flag table" seems reasonable to me.

I don't really feel strongly about it, I just don't see much reason to needlessly pass things around where it isn't necessary.

ridz1208:
there seems to be a typing issue see travis

driusan:
@ridz1208 the travis issues are fixed

php/libraries/NDB_BVL_Instrument.class.inc
ridz1208: This is the only thing blocking the PR at this point

sruthymathew123:
@driusan @ridz1208 Code changes looks all good to me. The instrument loaded and data got saved successfully. 
But when I tried to test the particular instrument instrument with CouchDB_Import_Instruments script; it took more time than I expected and suddenlyit crashed with the below attached error.
couch_db_crash

So I am unable to test the DQT import script. 

@ridz1208 Are you able to test the import script on CCNA?

driusan:
@sruthymathew123 It sounds like you don't have a CouchDB server running at the location that your config is pointing to when you tried running the script

sruthymathew123:
@driusan But before running this script I was able to access the DQT; and last day I successfully ran a DQT import script. So I don't know what exactly happened to make it unable to connect. Weird!

driusan:
Were you running the import script from the same location? `192.168.x.y` is an internal network address block.

sruthymathew123:
@driusan For me the script seems to be running as a loop with the data(data seems to be recurring and taking long time that expected) and it is then crashing with could not connect.

sruthymathew123:
@xlecours @xlecours  Can any of you test the DQT import instrument script?

driusan:
@sruthymathew123 you tagged @xlecours twice and said "can any of you".. did you mean to tag 2 different people?

sruthymathew123:
@driusan Oops! it was my bad. The other one i meant was @zaliqarosli

zaliqarosli:
on testing:

1. on removing the table and setting `$jsonData=true` for an instrument, the data loads from and saves to the data column of the flag table successfully but I am still shown this error on the front-end on clicking 'Save Data' 

screenshot 2019-02-25 13 47 16

2. `CouchDB_Import_Instruments.php` runs well for me with no errors but I am unable to test whether the data populates in the DQT because there are major front-end bugs in the module

driusan:
@zaliqarosli It looks like that error is from the code in the instrument's score function, not the LORIS code.

zaliqarosli:
@driusan you're right. in that case, i agree with rida's suggestion to have both the data column and tables for instruments and slowly phase out the sql tables because for both php and list instruments, all the scoring scripts will need to be updated. or is this not true and i'm missing something?

ridz1208:
@zaliqarosli your score functions should not be making direct database calls but instead be using the NDB_BVL_Instrument::loadinstancedata and save functions

driusan:
@zaliqarosli the instruments do not switch to json loading unless the flag inside the instrument is specifically switched to enable json loading.. so only the ones that have scoring which assume direct database access need to be updated, and they can be done one at a time.

zaliqarosli:
still needs rebase but passes testing

driusan:
@zaliqarosli rebased, might be worth quickly retesting one last time since there were conflicts

zaliqarosli:
@driusan if i set $jsonData back to false, it's not writing to the sql table if the entry doesn't exist

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -737,6 +739,8 @@ class NDB_BVL_Instrument extends NDB_Page
      */
     function _save(array $values): void
     {
+        $db = Database::singleton();
```
zaliqarosli: Database::singleton() called twice (on line 751 below)
```suggestion
```

zaliqarosli:
also .. `PHP Fatal error:  Uncaught TypeError: Return value of NDB_BVL_Instrument::loadInstanceData() must be of the type array, null returned in /var/www/loris/php/libraries/NDB_BVL_Instrument.class.inc:2070`

driusan:
> @driusan if i set $jsonData back to false, it's not writing to the sql table if the entry doesn't exist

Sorry, I don't understand what you mean..

driusan:
@zaliqarosli are you sure you're on the right branch? Line 2070 is the middle of a comment (and I don't see any way for the loadInstanceData function to return null on this branch..)

zaliqarosli:
@driusan this PR should still allow for instrument data to be loaded from and saved into the instrument's SQL table right? the functionality for $jsonData=true (using flag table data column) works, but for $jsonData=false (back to using sql table) it does not, the instrument breaks

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1469,22 +1473,15 @@ class NDB_BVL_Instrument extends NDB_Page
...
-            $query,
-            array('CID' => $this->getCommentID())
-        );
-        foreach ($dataFields as $field) {
+        $allData = $this->loadInstanceData($this);
+        foreach ($this->_requiredElements as $field) {
             // this shouldn't be just empty() b/c 0 is a valid
             // value for some required fields
-            if (is_null($field) || $field === "") {
+            if (is_null($allData[$field]) || $allData[$field] === "") {
```
zaliqarosli: i think there might be someone wrong here. `Data_entry_completion_status` gets set to Incomplete even though it should be complete / was complete before

driusan:
@ridz1208 Did you test both the case that @zaliqarosli was having trouble with and the case where it's enabled after your change?

ridz1208:
@driusan I was able to save/load from SQL with boolean set to false, save/load from Data column when boolean set to true.

I was able to export to CouchDB but as @zaliqarosli mentioned above, the DQT is unuseable on major thus i had to verify the data by directly accessing the backend of couchDB. I made a couple commits to get rid of PHP warnings as well for this script.

what is the "case where it's enabled after your change?" not sure what you mean by that

php/libraries/NDB_BVL_Instrument.class.inc
zaliqarosli: @ridz1208 did you have any issues with scoring when setting jsonData to true? when i enabled using the flag table, the value for `Data_entry_completion_status` column for that instrument always gets saved as "incomplete" and the code doesn't enter the instrument score function as a result

driusan:
I meant did you test with both jsonData true and false after adding the null coalesce.

ridz1208:
@driusan yes tested true and false

PR #3895, opened 2018-08-23
[react] Updated React to v16.2.4 (Latest) & with All Modules to es6 standard.

intralizee:
This PR updates Loris to using the latest & greatest version of React.
I had to scope through every module with a fine tooth comb and convert any react code to es6.

Reviewers should test all the modules made with react and to see if anything is broken.


driusan:
I think this would be slightly easier to do/review after #3891 (but still difficult because of the scope..)

If React requires upgrading to ES6, maybe this would be easier to split in 2? (one to update the to ES6, one after that to upgrade React..) Or is "upgrading React" after that point just changing 1 or 2 files?

intralizee:
@driusan Updating react is adding the react.development.js, react.production.js, react-dom.development.js, react-dom.production.js files to the htdocs/js/react directory in loris. Also making the changes to the NDB_Page.class.inc inside php/libraries/NDB_Page.class.inc to serve the latest react. When doing this everything will break unless updating the modules to es6 like I did. I don't believe #3891 will make it slightly easier to review but good news is either PR won't effect each other.
Also to note I found that MultiSelectDropdown.js was not being compiled beforehand and fixed that in this PR.

intralizee:
I accidentally hit the close button after adding my comment. 🙂

intralizee:
@kongtiaowang Hi Shen, travis is failing in the _testMytaskPanelAndLink() method at line 595 found in the modules/dashboard/test/DashboardTest.php. The lorisworkspace does appear to exist when visiting the dashbboard. Thoughts on what I should alter?

driusan:
@intralizee okay, but I think it should go in `vendor/js/` not `js`, we should be more careful about that than we have been in the past.

(Also, you can resolve the conflict by just deleting the file, it was removed from major.)

intralizee:
@driusan You're right i'll change it to vendor/js/ and thanks!

intralizee:
@driusan you're right i'll move it to vendor/js/ and thanks!

intralizee:
Hi @driusan & @kongtiaowang would you guys happen to know why travis is failing in the testBVLFeedbackLoads() function in the /modules/bvl_feedback/test/BvlFeedbackTest.php.

I've created a candidate using this PR and went to the url of the /candID/ with running the command:
`document.querySelector('#nav-right > li:nth-child(1) > a > span').click()`
...it opens the right side panel successfully.

Then I ran the command:
`document.querySelector('#bvl_feedback_menu > div.breadcrumb-panel > a').textContent`
and it returns the string: "Feedback for PSCID: DCC0000".

So right now I'm stumped on why travis is failing with it's test case.

driusan:
@intralizee No, but I'm having similar problems with the same test in #3910, so let's see what happens with #3913 ..

kongtiaowang:
@intralizee 
for the BVL_feedback I found this error on the page.
"The above error occurred in the <FeedbackPanelContent> component:
    in FeedbackPanelContent (created by FeedbackPanel)
    in div (created by AccordionPanel)
    in div (created by AccordionPanel)
    in div (created by AccordionPanel)
    in AccordionPanel (created by FeedbackPanel)
    in div (created by SliderPanel)
    in SliderPanel (created by FeedbackPanel)
    in FeedbackPanel

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://fb.me/react-error-boundaries to learn more about error boundaries." react-dom.development.js:14805:5

kongtiaowang:
@intralizee There is 
screen shot 2018-08-31 at 3 45 25 pm
another error on Data team helper.

intralizee:
Thanks @kongtiaowang for the help! 🙂

jsx/Breadcrumbs.js
```diff
@@ -10,11 +10,13 @@
  *
  */
 
+import React, {Component} from 'react';
```
HenriRabalais: @xlecours had mentioned that there might be a way to do this globally so that we are able to import `React`, `Component` and `PropTypes` to all `jsx` files. What are your thoughts Xavier?

intralizee: @HenriRabalais @xlecours Stackoverflow has an answer on how to do it found here:
https://stackoverflow.com/questions/30560883/es6-global-import
Yet, I'm not necessarily sure if it's good practice to globally import modules instead of just writing the imports necessary to the component. Also the way .eslintrc.json is configured in major will give errors if something isn't used in a component. Such as a component that doesn't use PropTypes. Thoughts?

kongtiaowang:
@intralizee 
All the ajax call should add a dataType:'json' in the react.behavioural_feedback_panel.js.
    $.ajax({
      type: 'GET',
      url: loris.BaseURL + '/bvl_feedback/ajax/get_thread_entry_data.php',
      dataType: 'json',
      data: {feedbackID: this.props.feedbackID},

.eslintrc.json
```diff
@@ -29,7 +29,7 @@
     "new-cap": ["error", {
       "capIsNewExceptions": ["DynamicTable"]
     }],
-    "no-console": ["error", {
+    "no-console": ["warn", {
```
driusan: What does this change do?

intralizee: https://eslint.org/docs/rules/no-console
The change makes "npm run compile" not output an error but just a warning when a developer is using console.log somewhere in the frontend. I think it's probably best but thoughts?

driusan: I think it should stay an error, warnings get ignored. We should in general try be trying to make static checks stricter, not weaker over time. A leftover `console.log` is likely a debugging statement that was left behind that we shouldn't be depending on review to catch if it can be caught by Travis proactively.

Fix this and I'll merge it (assuming it's still passing Travis), since no one's going to realistically be able to test/review all of LORIS for this PR and it's better that it goes in so we catch any bugs before the last minute..

jsx/Breadcrumbs.js
HenriRabalais: Sounds good to me! @xlecours ?

jsx/DynamicDataTable.js
```diff
@@ -1,88 +1,110 @@
...
       if (this.state.error !== undefined) {
         return <div className="alert alert-danger">
-                         <strong>
-                           {this.state.error}
-                         </strong>
-                       </div>;
+                 <strong>
+                   {this.state.error}
+                 </strong>
+               </div>;
```
HenriRabalais: Does this return statement not have to be wrapped in parentheses since it covers multiple lines of code?

jsx/Breadcrumbs.js
```diff
@@ -93,33 +96,37 @@ window.removeEventListener('resize', this.checkScreenSize);
...
       </div>
     );
   }
 }
+Breadcrumbs.propTypes = {
+  baseURL: PropTypes.string,
+  breadcrumbs: PropTypes.array,
+};
 
 let RBreadcrumbs = React.createFactory(Breadcrumbs);
```
HenriRabalais: This appears to be legacy syntax and should likely be removed throughout the code.

https://reactjs.org/docs/react-api.html#createfactory

jsx/MultiSelectDropdown.js
```diff
@@ -5,11 +5,24 @@
...
-  toggleCheckbox: function() {
+import React, {Component} from 'react';
+
+class SelectField extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+
+    };
```
HenriRabalais: I don't think we need a state declaration if this is a stateless component. Same thing for all the ones below.

modules/candidate_parameters/jsx/FamilyInfo.js
```diff
@@ -9,22 +13,28 @@ let FamilyInfo = React.createClass({
...
-  componentDidMount: function() {
+    this.setFormData = this.setFormData.bind(this);
+    this.onSubmit = this.onSubmit.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.showAlertMessage = this.showAlertMessage.bind(this);
+    this.deleteFamilyMember = this.deleteFamilyMember.bind(this);
+  }
+
+  componentDidMount() {
     let that = this;
```
HenriRabalais: If we are going to be using this sort of reference to this (although I'm not convinced that doing this is necessary), we should atleast retain internal consistency and decide between `self` or `that`.

davidblader: i think the `this`/`that`/`self` consistency issue should be left to a different PR

jsx/Breadcrumbs.js
```diff
@@ -10,11 +10,14 @@
  *
  */
 
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
```
zaliqarosli: why do we need to import PropTypes here? we've been using Proptypes without needing to import. are you removing the importing somewhere else? @intralizee

jsx/Breadcrumbs.js
zaliqarosli: i agree. i think we can use export instead?

jsx/DynamicDataTable.js
```diff
@@ -1,88 +1,110 @@
...
+    this.setState({
+      Headers: data.Headers,
+      Data: data.Data,
+      isLoaded: true,
+    });
+  }
+
+  ajaxFailure(data, errorCode, errorMsg) {
+    console.error(errorCode + ': ' + errorMsg);
+    this.setState({error: 'Error loading data'});
```
zaliqarosli: set state as 1 line or multiple lines like in xhrProgress() above?

jsx/DynamicDataTable.js
zaliqarosli: yes, let's wrap it!

jsx/FilterForm.js
```diff
@@ -5,7 +5,8 @@
  * @version 1.1.0
  *
  */
-
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
```
zaliqarosli: I understand that you are importing React because you've removed 'React.' but do PropTypes still also need to be imported?

jsx/PaginationLinks.js
```diff
@@ -69,25 +70,34 @@ let PaginationLinks = React.createClass({
...
-});
+      <ul className='pagination pagination-table'>
+          {pageLinks}
+      </ul>
+    );
+  }
+}
+PaginationLinks.propTypes = {
+  onChangePage: PropTypes.func,
+  Total: PropTypes.number.isRequired,
```
zaliqarosli: trailing comma here can be removed

jsx/PaginationLinks.js
```diff
@@ -69,25 +70,34 @@ let PaginationLinks = React.createClass({
...
+};
+PaginationLinks.defaultProps = {
+  RowsPerPage: 10,
+  Active: 1,
```
zaliqarosli: trailing comma here can be removed

jsx/PaginationLinks.js
```diff
@@ -1,29 +1,30 @@
 /* exported RPaginationLinks */
 
-let PaginationLinks = React.createClass({
-  mixins: [React.addons.PureRenderMixin],
```
zaliqarosli: what is this guy and did you get rid of it?

jsx/PaginationLinks.js
```diff
@@ -1,29 +1,30 @@
...
 
-      if (that.props.onChangePage) {
-        that.props.onChangePage(i);
+      if (this.props.onChangePage) {
+        this.props.onChangePage(i);
       }
-    };
-  },
-  render: function() {
+    }.bind(this);
```
zaliqarosli: i'm always unsure about this but if we bind this above in the constructor, do we need to bind it again at the end of the function? looks redundant to me but not sure

modules/candidate_parameters/jsx/FamilyInfo.js
```diff
@@ -9,22 +13,28 @@ let FamilyInfo = React.createClass({
...
     $.ajax(
-        this.props.dataURL,
+      this.props.dataURL,
```
zaliqarosli: could move ajax call out into a fetchData() function and then remove `let that = this;`

modules/dataquery/jsx/react.paginator.js
```diff
@@ -1,69 +1,73 @@
...
-            'RowsPerPage' : 10,
-            'Active' : 1
-        }
-    },
-    changePage: function(i) {
-        var that = this;
-        return function(evt) {
-            // Don't jump to the top of the page
-            evt.preventDefault();
+class PaginationLinks extends Component {
```
zaliqarosli: is this component the exact same as what is in `jsx/PaginationLinks.js`?

modules/dataquery/jsx/react.sidebar.js
```diff
@@ -1,66 +1,87 @@
...
-        return (
-                <div>
-                    <h2>{this.props.Name}</h2>
-                    {this.props.children}
-                </div>
-       );
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+
+class Sidebar extends Component {
```
zaliqarosli: im confused as to why all these React components live inside the data query module directory

jsx/PaginationLinks.js
driusan: PureRenderMixin? It used to be a mixin for react components where render is a pure function of state and props which made it more efficient, but I think it was deprecated a couple React releases ago.

jsx/PaginationLinks.js
davidblader: it looks like this is binding to the returned anon function, not `changePage`, so this usage is :ok_hand:

jsx/FilterForm.js
intralizee: @zaliqarosli Yes since it's no longer React.PropTyptes

jsx/PaginationLinks.js
intralizee: @zaliqarosli trailing comma is actually a requirement now with eslint changes.

jsx/PaginationLinks.js
intralizee: @zaliqarosli trailing comma is actually a requirement now with eslint changes.

jsx/PaginationLinks.js
intralizee: @zaliqarosli It's no longer needed with what @driusan said.

jsx/Breadcrumbs.js
intralizee: @zaliqarosli @HenriRabalais I'm having trouble with updating this one. Not sure what to replace it with because after the changes suggested it will not render the breadcrumbs. Thoughts?

jsx/Breadcrumbs.js
intralizee: @zaliqarosli No longer using React.PropTypes and which is why it's imported.

modules/dataquery/jsx/react.sidebar.js
intralizee: @zaliqarosli The data query module uses them and they seem to be custom. I think leaving cleaning up that to a different PR is best. Also the DQT is in status of being redesigned.

jsx/DynamicDataTable.js
intralizee: @zaliqarosli I'm not sure what you mean "set state as 1 line or multiple lines like in xhrProgress() above?" can you add to that?

modules/dataquery/jsx/react.paginator.js
intralizee: @zaliqarosli I think it's custom for the DQT

jsx/Breadcrumbs.js
intralizee: @zaliqarosli and @HenriRabalais It turns out the main.tpl is only capable of using the RBreadcrumbs without JSX. So I'll leave it for another PR that cleans up the main.tpl

HenriRabalais: How about for the other components?

Issue #2870, opened 2017-06-07
File uploaders should check filename with javascript, rather than checking on the backend after upload


gdevenyi:
I have a particularly slow connection, uploading a ZIP file with a typo in the name results in the error getting thrown after the upload rather than as a pre-submit check in Javascript.

johnsaigle:
Could you clarify what you mean by a typo? What was the intended filename and the inputted filename? Which error?

gdevenyi:
When I use the imaging uploader, I specify PSCID, CandID and VisitID which imposes a naming scheme on my zip file of:
```
<PSCID>_<CandID>_Initial_MRI.zip
```

Say I misspell "Initial" to be "Inital" as I did yesterday.

I select the file in my file browser, click upload, the file uploads, then LORIS errors out because the file is named wrong.

I think instead, it would make sense to use the inputted fields and filename to check the filename with javascript before upload. Reduces bandwidth use, etc.

johnsaigle:
You're right, that is annoying and inconsistent with other parts of the codebase. In some places, such as in the `media` uploader, we do have front-end alerts that do this sort of front-end validation. However, it has not yet been fully implemented across LORIS.

In future versions, this behaviour will be more consistent. 

Thanks for bringing this up.

PR #3896, opened 2018-08-23
[Imaging Uploader] Validate filename in front-end before attempting upload

davidblader:
### Brief summary of changes
Checks the filename for candidate uploads before attempting to place it onto the server

### This resolves issue...
attempting to upload incorrect filenames

- [ ] Redmine? #15109

- [ ] Github? #2870 

### To test this change...

- [ ] try uploading a file to imaging uploader where the filename does not match the PSCID_CANDID_VISIT-LABEL convention for a candidate upload. you should get an error swal


driusan:
@davidblader still failing Travis

modules/imaging_uploader/jsx/UploadForm.js
```diff
@@ -77,16 +77,42 @@ class UploadForm extends React.Component {
...
+      }
+      // Make sure file follows PSCID_CandID_VL[_*].zip|.tgz|.tar.gz format
+      const pcv = data.pSCID + '_' + data.candID + '_' + data.visitLabel;
+      const pcvu = pcv + '_';
+      const properName = new RegExp("^" + pcv + ".(zip|tgz|tar.gz)");
+      const properNameExt = new RegExp("^" + pcvu + ".*(.(zip|tgz|tar.gz))");
+      if (!fileName.match(properName) && !fileName.match(properNameExt)) {
+        swal({
+          title: "File improperly named!",
+          text: "File name must match " + pcv +
```
cmadjar: I would rephrase the error message to the following:
"File name and values in the PSCID, CandID and Visit label fields of the instrument do not match. Please verify that the information entered in the fields or the file name are correct."

I find it dangerous to suggest them that it is the file name that is not correct (it is possible that what is not correct is one of the field instead of the filename)

davidblader:
@cmadjar i fixed it :~)

PR #3898, opened 2018-08-23
[Core: Login, User Accounts] Allow certain users to access the system in a time window basis

racostas:
### Brief summary of changes

For the project we are working at CHU Sainte-Justine we realize we will need to set (for a given user) access to the system but only for a given time window (due to policies of data access and privacy). It's a specific requirement of our project that could be useful for others. 
The way it was implemented (following @ridz1208 recommendations. Very helpful!! Thanks!) should not break/interfere others functionalities of the system. 

The changes are all additions limited to the following files:

php/libraries/SinglePointLogin.class.inc
modules/user_accounts/templates/form_edit_user.tpl
modules/user_accounts/php/edit_user.class.inc

Also it's necessary to add two fields on the Users table:

SQL/New_patches/2018-08-22_user_activeFrom-activeTo.sql

### This resolves issue...

- [ ] Redmine? # -

- [ ] Github? # -

### To test this change...

1. To access the system with a user with permissions to add/modify another.
2. To modify the user access time windows filling the new fields 'Active from' and/or 'Active to' that should shows up between the fields 'Active' and 'Pending approval'. NULLs are allowed, which will mean that there is not restrictions.
3. According to the dates set the user modified should not be able to access the system (due to expired/no yet active user).

image

image2

All suggestions/modifications are very welcome! Thanks.
CSJ team.

### DIFF
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -783,6 +783,33 @@ class Edit_User extends \NDB_Form
 
         //------------------------------------------------------------
 
+	// active time windows
+	$this->addSelect('ActiveTW','Active Time Windows', array('Y' => 'YesTW', 'N' => 'NoTW'));
+
+	$dateOptions = array(
+				'language'	=>	'en',
+				'format'	=>	'YMd',
+				'addEmptyOption' =>	'true'
+			);
+
+	$dateAttributes = ['class' => 'form-control input-sm input-date'];
+
+
+	$this->addBasicDate(
+				'active_from',
+				'Active from',
+				$dateOptions,
+				$dateAttributes
+			);
+
+
+	$this->addBasicDate(
+				'active_to',
+				'Active to',
+				$dateOptions,
+				$dateAttributes
+			);
+
         // checking if the account can be rejected (no password hash
         // pending approval)
 
@@ -1086,6 +1113,18 @@ class Edit_User extends \NDB_Form
                 'Radiologist' and 'Pending Approval').";
             }
         }
+
+        //======================================
+	//	Validate Active time Windows (from - to)
+	//
+	//	Note: the only validation included for the moment
+	//	is that the time form which the user is active 
+	//	should be less or equal to the time to it will be active. 
+        //======================================
+
+	if(($values['active_to'] != NULL) && ($values['active_from'] > $values['active_to'])){
+		$errors['active_timeWindows'] = "Please notice that the <i>\"Active from\"</i> date should be lesser or equal to the <i>\"Active to\"</i> date.";
+	}
         return $errors;
     }
 
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -783,32 +783,29 @@ class Edit_User extends \NDB_Form
 
         //------------------------------------------------------------
 
-	// active time windows
-	$this->addSelect('ActiveTW','Active Time Windows', array('Y' => 'YesTW', 'N' => 'NoTW'));
+        // active time windows
 
-	$dateOptions = array(
-				'language'	=>	'en',
-				'format'	=>	'YMd',
-				'addEmptyOption' =>	'true'
-			);
+        $dateOptions = array(
+                        'language'       => 'en',
+                        'format'         => 'YMd',
+                        'addEmptyOption' => 'true',
+                       );
 
-	$dateAttributes = ['class' => 'form-control input-sm input-date'];
+        $dateAttributes = ['class' => 'form-control input-sm input-date'];
 
+        $this->addBasicDate(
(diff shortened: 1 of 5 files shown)


driusan:
@racostas This is failing Travis because it needs to have phpcs run on it before Travis runs the test suite. The test suite runs:

```
vendor/bin/phpcs --standard=docs/LorisCS.xml php/libraries php/exceptions php/installer;
vendor/bin/phpcs --standard=docs/LorisCS.xml --extensions=php/php htdocs;
vendor/bin/phpcs --standard=docs/LorisCS.xml --extensions=php/php,inc/php modules/;
```

from the root of the LORIS directory as a first step. (You can replace "phpcs" with "phpcbf" to fix the ones reported that have an `[x]` beside them, but the remaining ones like line length too line need to be fixed manually. You can see the log from Travis CI by clicking on the red x beside a commit id in GitHub)

driusan:
I don't have any feelings one way or another over to/until and I doubt it's possible to make it work without the alter table since the query needs to select the columns.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -1086,6 +1113,18 @@ class Edit_User extends \NDB_Form
...
+        //======================================
+	//	Validate Active time Windows (from - to)
+	//
+	//	Note: the only validation included for the moment
+	//	is that the time form which the user is active 
+	//	should be less or equal to the time to it will be active. 
+        //======================================
+
+	if(($values['active_to'] != NULL) && ($values['active_from'] > $values['active_to'])){
+		$errors['active_timeWindows'] = "Please notice that the <i>\"Active from\"</i> date should be lesser or equal to the <i>\"Active to\"</i> date.";
```
johnsaigle: We usually don't inject HTML directly into error messages like this as a matter of policy. If the italics are very important to you here, this should be done in the Smarty template instead.

cmadjar:
@johnsaigle I like your suggestion!

php/libraries/SinglePointLogin.class.inc
```diff
@@ -328,6 +331,26 @@ class SinglePointLogin
                     return false;
                 }
 
+		// check if the account is longer active
+		$date = new DateTime();
+		$currentDay = $date->getTimestamp();
```
johnsaigle: Technically this should be named something like `$currentTime` as it will incorporate the hour, minute, second, etc.

php/libraries/SinglePointLogin.class.inc
```diff
@@ -328,6 +331,26 @@ class SinglePointLogin
...
+			$this->_lastError = "Your account has expired."
+                        . " Please contact your project administrator to re-activate"
+                        . " this account.";
+                	return false;
+
+		}
+
+		// check if the account is active		
+		if(($row['active_from'] != NULL) && ($currentDay < strtotime($row['active_from']))){
+			$this->_lastError = "Your account is not active yet. According to our records it will be active from ".$row['active_from']."."
```
johnsaigle: You're already using double-quotes here so I think string interpolation is a bit cleaner than concatenating.

```php
$this->_lastError = "Your account is not active yet. According to our records it will be active from {$row['active_from']}."
```

racostas:
Hi @driusan , Thanks for the comments. They are very appreciated!

I did the phpcs and phpbf for the files I modified (php/libraries/SinglePointLogin.class.inc,  modules/user_accounts/templates/form_edit_user.tpl, modules/user_accounts/php/edit_user.class.inc). After that and some manual fix that was required all seems to be OK when I run phpcs again for this three files. What is strange now is that Travis is giving me format errors(specifically warnings) in files I have not touched at all? Maybe I’m missing something obvious??

The files I modified do not seem to be among the ones reported with issues now. But maybe because they did not got analyzed as it seems that Travis exceed the execution time. 

The final result is that Travis gives the test as failed.

screenshot_1

screenshot_2

Thanks in advance for the help you may provide us,
CNBP Team.

driusan:
@racostas The warnings are from the equivalent of phpcs for javascript and are harmless, they don't fail the build. Those warnings are printed on every pull request. The "E" lines are from the test suite and what matters, but since it's timing out before finishing you're not getting the summary of what they mean that's usually printed after it finishes, which makes it hard to say exactly what's wrong...

Since every test is failing and you're changing the login, I'd guess that something is broken with the login process under Travis causing every test to fail when it tries logging in. It could be because you didn't update the schema file, or because the test data needs to be updated to insert properly with the schema changes. @kongtiaowang might be able to help you get to the bottom of it.

kongtiaowang:
https://github.com/aces/Loris/pull/3916

racostas:
Hi @kongtiaowang. Travis test passed now thank to your insights. Great help! Thank you very much!

driusan:
@cmadjar Is your review still valid or has it been adequately addressed?

PR #3899, opened 2018-08-24
[Docs] Remove unused file

driusan:
I don't know why there's a main.css file in the docs directory.


driusan:
@cmadjar Dunno, the history says it was added 10 years ago in the initial commit. I think it's safe to say it hasn't been maintained since then.

PR #3901, opened 2018-08-24
[Core] Remove PT Sans and PT Sans Bold fonts

driusan:
This removes the PT Sans and PT Sans Bold fonts
from the htdocs directory (and the reference to
them in CSS.) They're only used in one place,
released under a different license from LORIS,
and only used for public modules. The overhead of
keeping them in LORIS is probably not worth the
benefit of using them over the generic sans-serif
font family.

From what I could find online (https://fonts.google.com/specimen/PT+Sans):

> PT Sans was developed for the project "Public Types of Russian Federation." 
> The second family of the project, PT Serif, is also available.
>
> The fonts are released with a libre license and can be freely redistributed: The 
> main aim of the project is to give possibility to the people of Russia to read and
> write in their native languages.

While certainly a noble goal, I don't think it's necessary to host the fonts in LORIS.


driusan:
@intralizee @zaliqarosli  Requesting your review on this not so much for the code but in case you know of anywhere on the frontend where we really need this font instead of a generic sans font.

PR #3906, opened 2018-08-27
[server_processes_manager] Removing hardcoded prod and environment file

cmadjar:
### Brief summary of changes

This PR request removes the reads the name of the MRI config file and environment file to be used by the server processes manager module directly from the Config tables (which are two new ConfigSettings in this PR).

### This resolves issue...

- Redmine: https://redmine.cbrain.mcgill.ca/issues/13727

### To test this change...

- [ ] run the SQL patch and make sure the default setting correspond to your instance (a.k.a. your MRI config file in `$LORIS-MRI/dicom-archive/.loris_mri/` directory is indeed called `prod` and your MRI environment file to source is indeed called `environment`). FYI, these two new config are in the `imaging pipeline` section of the config module
- [ ] make sure you set the config option `ImagingUploader Auto Launch` is set to `Yes` in the `Study` section of the config module
- [ ] upload a new scan via the imaging uploader and make sure the pipeline runs as expected.

# If this sentence shows up it means I didn't read the instructions!

I read it but just felt that leaving it ;).


SQL/0000-00-03-ConfigTables.sql
```diff
@@ -151,6 +151,7 @@ INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType,
...
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'excluded_series_description', 'Series description to be excluded from insertion into the database (typically localizers and scouts)', 1, 1, 'text', ID, 'Series description to exclude from imaging insertion', 18 FROM ConfigSettings WHERE Name="imaging_pipeline";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'ComputeDeepQC', 'Determines whether a call is made from LORIS-MRI to the DeepQC app for automatic QC prediction', 1, 0, 'boolean', ID, 'Compute automatic QC', 18 FROM ConfigSettings WHERE Name="imaging_pipeline";
+INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'MriConfigFile', 'Name of the MRI config file (stored in dicom-archive/.loris_mri/', 1, 0, 'text', ID, 'Name of the MRI config file', 19 FROM ConfigSettings WHERE Name="imaging_pipeline";
```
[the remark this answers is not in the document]
cmadjar: Indeed. I forgot to add it there. Thanks!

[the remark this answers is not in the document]
cmadjar: done!

modules/server_processes_manager/php/mriuploadserverprocess.class.inc
```diff
@@ -105,6 +99,26 @@ class MriUploadServerProcess extends AbstractServerProcess
             throw $exception;
         }
 
+        try {
+            $this->_prodFile = $config->getSetting('MriConfigFile');
```
driusan: There shouldn't never be a try/catch around getSetting. If the setting isn't present and the code depends on it, the server is misconfigured and it should die loudly so that the server can be fixed.

cmadjar: @driusan I simply reproduces the code for the other ConfigSettings (see line 98). Should I change this for all or just the ones I added. Let me know. Thanks!!

driusan: Those ones shouldn't have gotten in either.. it's probably best to fix them both so that people don't try copying it later.

modules/server_processes_manager/php/mriuploadserverprocess.class.inc
```diff
@@ -95,15 +89,9 @@ class MriUploadServerProcess extends AbstractServerProcess
...
-        } catch(\ConfigurationException $exception) {
-            error_log(
-                "Unable to construct MriUploadTask: 
-                 config setting MRICodePath undefined"
-            );
-            throw $exception;
-        }
+        $this->_mriCodePath     = $config->getSetting('MRICodePath');
+        $this->_prodFile        = $config->getSetting('MriConfigFile');
+        $this->_environmentFile = $config->getSetting('EnvironmentFile');
```
driusan: so much prettier..

cmadjar:
@driusan I removed the try/catch for the getConfigSettings. Thanks!

kongtiaowang:
It works well.

SQL/New_patches/2018-08-27_add_name_of_mri_config_file_to_config_module.sql
```diff
@@ -0,0 +1,35 @@
+--- INSERT THE 2 NEW CONFIGs INTO ConfigSettings TABLE
+INSERT INTO ConfigSettings
+  (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber)
+  SELECT
+    'MriConfigFile',
+    'Name of the MRI config file (stored in dicom-archive/.loris_mri/',
```
driusan: This is missing a closing bracket

cmadjar: Good catch! Fixed in the next commit. Thanks!

PR #3907, opened 2018-08-28
[conflict_resolver] Upgraded module to use react & Modified jsx/Form.js ButtonElement

maltheism:
### Brief summary of changes

This pull request removes smarty and upgrades module to use react.
It removes the smarty templates and adds two new react files for conflicts unresolved & resolved.

Also I added the ability for react ButtonElement to have the attributes "id & name" and be optional when creating a button. Reasoning is since a lot of tests have been written to use the WebDriverBy::ID('example') or by Name and for the goal to test clicking a button.

This PR is based from #3730 

### This resolves issue...

- [ ] Redmine? #14841

- [ ] Github? #

### To test this change...

- Checkout the PR and test the conflict resolver with unresolved & resolved conflict data.


modules/conflict_resolver/jsx/react.conflict_resolver.js
```diff
@@ -0,0 +1,271 @@
...
+          <FilterForm
+            Module="conflict_resolver"
+            name="conflict_resolver_filter"
+            id="conflict_resolver_filter"
+            ref="conflict_resolver_Filter"
+            columns={2}
+            formElements={this.state.Data.form}
+            onUpdate={this.updateFilter}
+            filter={this.state.filter}
+            emptyOption={false}
```
zaliqarosli: this emptyOption doesn't actually get passed right?

modules/conflict_resolver/jsx/react.conflict_resolver.js
```diff
@@ -0,0 +1,299 @@
...
+  formatColumn(column, cell, rowData, rowHeaders) {
+    if (loris.hiddenHeaders.indexOf(column) > -1) {
+      return null;
+    }
+    // Create the mapping between rowHeaders and rowData in a row object.
+    let row = {};
+
+    rowHeaders.forEach(function(header, index) {
+      row[header] = rowData[index];
+    }, this);
```
HenriRabalais: If I'm not mistaken, `this` doesn't need to be bound to the function.

modules/conflict_resolver/jsx/react.resolved_conflicts.js
```diff
@@ -0,0 +1,289 @@
...
+      error: function(error) {
+        console.error(error);
+      }
+    });
+  }
+
+  updateFilter(filter) {
+    this.setState({filter});
+    if (this.child !== undefined) {
+      this.child.updateFilterState(filter);
```
HenriRabalais: I still don't think this is the best way to go about updating filter data. A change in the filter form should trigger a state update in the parent component which is then passed down as props to the child component -- not forcing its child to update its own state. We should be attempting to centralise state at the highest point of the component hierarchy, and pass that state down as necessary to children via props.

jsx/Form.js
```diff
@@ -1304,6 +1304,8 @@ class ButtonElement extends React.Component {
       <div className="row form-group">
         <div className={this.props.columnSize}>
           <button
+            {...(this.props.id ? {id: this.props.id} : {})}
+            {...(this.props.name ? {name: this.props.name} : {})}
```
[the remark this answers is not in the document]
intralizee: @HenriRabalais this.props.id just contains what the developer passes to the component for the id value. It will not supply the html button element with the complete "id: 'ex_button'" in that case. The spread operator is used because I'm returning an object: {id: 'ex_button'} and I want both the 'key' & 'value' to be inserted into the html button. When there is nothing I don't want the id: '' to be placed in it.

modules/conflict_resolver/jsx/react.conflict_resolver.js
```diff
@@ -0,0 +1,299 @@
+import {Tabs, TabPane} from 'Tabs';
```
zaliqarosli: i think this got mixed in from your other PR! should be removed

zaliqarosli: oops sorry, *i* was on the wrong PR!! 😨

johnsaigle:
@maltheism Is this a PR you'll be able to pick up again?

maltheism:
@johnsaigle I remember this PR was working but there was some discussion about it should have been modified. I think i'll make a new branch and copy the code from here.

HenriRabalais:
Suggest closing since new PR has been made to replace this one (#4355)

maltheism:
Closing because see #4355 as replacement.

PR #3909, opened 2018-08-29
[Data Team Helper] Ensure test_name is always a string

johnsaigle:
### Brief summary of changes

Just ensure that `test_name` is a string. Shouldn't be too controversial.

This type coercion is causing issues with #3878 which is why it needs to be fixed.

### To test this change...

This shouldn't alter the behaviour of LORIS. I'm not really sure how `data_team_helper` works but it should work the same!


driusan:
Note to self: if you click things in the wrong order "Can only approve open pull requests "

PR #3910, opened 2018-08-29
[Database] Make pselectRow more exceptional

driusan:
This makes pselectRow error out in more invalid cases. In particular,
it throws an exception if called on a query that returns more than
one row (and as a result, also adds this behaviour to pselectOne) since
there's no way to know which row was intended. It also returns null,
not an empty array, if no rows are matched.

This improves LORIS by making latent errors more obvious
and exposing problems that might otherwise go unnoticed.

### DIFF
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -307,11 +307,11 @@ class Candidate
 
         $row = $db->pselectRow($query, $candidateIDs);
 
-        if ($row === null) {
-            return false;
+        if ($row !== null) {
+            return true;
         }
 
-        return true;
+        return false;
     }
 
     /**
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -307,7 +307,7 @@ class Candidate
 
         $row = $db->pselectRow($query, $candidateIDs);
 
-        if (!empty($row)) {
+        if ($row === null) {
             return true;
         }
 
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -307,11 +307,7 @@ class Candidate
 
         $row = $db->pselectRow($query, $candidateIDs);
 
-        if ($row !== null) {
-            return true;
-        }
-
-        return false;
+        return $row !== null;
     }
 
     /**
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -308,11 +308,10 @@ class Candidate
         $row = $db->pselectRow($query, $candidateIDs);
 
         if ($row === null) {
-            return true;
+            return false;
         }
 
-        return false;
-
+        return true;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


php/libraries/Candidate.class.inc
```diff
@@ -307,12 +307,11 @@ class Candidate
 
         $row = $db->pselectRow($query, $candidateIDs);
 
-        if (!empty($row)) {
-            return true;
+        if ($row === null) {
```
xlecours: How about `return !($row === null)` ?

php/libraries/Database.class.inc
```diff
@@ -742,16 +742,21 @@ class Database
...
+        if (count($rows) > 1) {
+            throw new \DomainException(
+                "Attempt to use pselectRow on a query that returns multiple rows"
+            );
+        }
         if (isset($rows[0])) {
             return $rows[0];
         }
-        return $rows;
+        return null;
```
xlecours: How about `return $rows[0] ?? null;` ?

php/libraries/NDB_BVL_Feedback.class.inc
```diff
@@ -369,7 +369,7 @@ class NDB_BVL_Feedback
         // new DB Object
         $db =& Database::singleton();
 
-        $query   = "SELECT Status FROM feedback_bvl_thread WHERE";
+        $query   = "SELECT MAX(Status) FROM feedback_bvl_thread WHERE";
```
xlecours: I would suggest to add ` 1=1 OR ` after the WHERE.

xlecours: There is no group by for the MAX function

driusan: I'm not certain if any of those options are right (or the one that I have), the BVL Feedback integration test is still failing, but I'm not able to get bvl feedback to work on any branch, so I'm not sure why this PR is the one causing it to fail..

PapillonMcGill:
Should not we check all 95 occurrences of pselectRow() to be sure it will not break somewhere?

driusan:
@PapillonMcGill Travis is passing in the general case, but whether or not `pselect*` breaks or not is data dependent, so I don't know how we would test other call sites in a way that gives more confidence than "our test suite passing."

PapillonMcGill:
@driusan Ok then. no more objection

php/libraries/Database.class.inc
```diff
@@ -742,16 +742,18 @@ class Database
...
      */
-    function pselectRow($query, $params)
+    function pselectRow(string $query, array $params) : ?array
     {
-        $rows = $this->pselect($query . " LIMIT 1", $params);
-        if (isset($rows[0])) {
-            return $rows[0];
+        $rows = $this->pselect($query . " LIMIT 2", $params);
+        if (count($rows) > 1) {
+            throw new \DomainException(
```
xlecours: Arguably UnexpectedValueException would apply.

PR #3911, opened 2018-08-29
[Examiner] Reactify Edit_Examiner page

zaliqarosli:
### Brief summary of changes
This PR reactifies the `edit_examiner` page of the Examiner module. The reactification of the index page will come in a separate PR as I will need to take over work that was done by @victoriafoing. For now, the two pages work fine together.

This is my first attempt at Reactification so please let me know if something could be done better. Thnx :)

### To test this change...

- [ ] Check out this branch
- [ ] Navigate to an examiner within the module
- [ ] Test functionality (everything should work the same, if not, better)
- [ ] Test display (everything should look the same minus the lengths of the date and comment fields as Form.js elements have a default label which, in this case, is not required)

**Note:** If you think this page could look better by some change, let me know. I tried to keep all the elements the same as before.


modules/examiner/ajax/editCertificate.php
```diff
@@ -0,0 +1,169 @@
+<?php
```
driusan: We shouldn't be adding new ajax/ scripts, we should be getting rid of them. I think this should be converted to an NDB_Page which overrides the process function to remove the middleware and uses the rest of the LORIS framework for hasAccess/setup/etc.

modules/examiner/jsx/editExaminer.js
```diff
@@ -0,0 +1,332 @@
...
+     * @param {string} formElement  name of the selected element
+     * @param {string} value        selected value for corresponding form element
+     *
+     */    
+    setFormData(formElement, value) {
+        let formData = this.state.formData;
+        let split = formElement.split("_");
+        let key = split[0];
+        let element = split[1];
+        formData[key][element] = value
```
zaliqarosli: missing ; at end of line

modules/examiner/jsx/editExaminer.js
```diff
@@ -0,0 +1,332 @@
...
+            success: data => {
+                this.setState({
+                    updateResult: "success"    
+                });
+                this.showAlertMessage();
+                this.fetchData();
+            },
+            error: err => {
+                if (err.responseText !== "") {
+                    var errorMessage = JSON.parse(err.responseText).message;
```
zaliqarosli: no var

modules/examiner/ajax/editCertificate.php
johnsaigle: > We shouldn't be adding new ajax/ scripts, we should be getting rid of them.

@driusan Could you clarify this? I didn't know we had this policy.

johnsaigle:
@zaliqarosli This is failing Travis due to phan errors.

zaliqarosli:
@johnsaigle for sure, this needs work

PR #3912, opened 2018-08-29
[MRI Preprocessing]: Adding tables to support BIDS dataset creation on the MRI side

MounaSafiHarab:
### Brief summary of changes

Add tables to the database to support the creation of a BIDS dataset for imaging data.
This is mostly to be used with PR #198 (https://github.com/aces/Loris-MRI/pull/198) on the MRI side.

Essentially 2 new tables are created, `bids_mri_scan_type_rel` and `bids_category`.
The former links to the `mri_scan_type` table and creates a link between the existing acquisition protocol IDs, and all the related derivatives needed from it to create a BIDS compliant dataset/filenames. The latter defines the different BIDS categories which structures the different MRI modalities in sub-directories.

More specifically, a detailed column description per table is as follows:
1) `MRIScanTypeID`: link to the mri_scan_type ID column (e.g. `44` for `t1` images in our default schema)
2) `BIDSCategory` which would be an `enum` of types `func`, `anat`, 'dwi`', `fmap`. But since `enum` is highly discouraged by LORIS standards, a new table called `BIDS_category` is created to house the different BIDS categories. The values inside this tables are used to create the appropriate folder in which to house the `nii` images. For example, a `t1` scan would be in a folder called `anat` because it fits in the anatomical category.
3) `BIDSScanTypeSubCategory`; this is mostly for fMRI. These `func` acquisitions need extra measures to define them (that also get stored in the accompanying JSON header file), so for a resting state fMRI, files should have a `task-rest` identifier, or what I define as the `BIDSScanTypeSubCategory`. A memory task fMRI would have a value of `task-memory` in the `BIDSScanTypeSubCategory` column. Other acquisitions can leave this field blank
4) `BIDSScanType` is the `mri_scan_type` Scan_type column, but named in a BIDS compliant manner. So `t1` becomes `T1w`, `dti` becomes `dwi`, etc... this is the column that would rename our schema's Scan_type to whatever BIDS wants it to be 
5) `BIDSMultiEcho` is a column that specifies which modalities are acquired in a multi-echo fashion. This is needed because those files need to be identified in BIDS with the string `echo-1`, `echo-2`, etc... in their filenames, so fill in this information in this column

**An example set of patches for CCNA mri_scan_types entries are added for now as example to illustrate the 5 points above, but will eventually be removed**

### This resolves issue...

- [ ] Redmine? #13438 (CCNA Redmine task though...). It is not an issue, but rather a new addition to LORIS-MRI to create a BIDS dataset of the existing `assembly/` MINC files.

### To test this change...

The MRI script in the referenced PR above needs to be run after the patches here have been applied.

- An example of how CCNA tables will look like after populating the proper BIDS table with expected values is shown in the following screenshot:
example_ccna_bidstables

- An example of how the BIDS files on the filesystem look like when running this script is shown in the following screenshot (this was generated using the argument `-dataset_name test` which is why the subdirectory housing the files is called `test`):
example_ccna_files

I added the `Blocked` tag until the MRI side/script is clear as well.

### DIFF
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -762,7 +762,7 @@ CREATE TABLE `mri_protocol_checks` (
 -- ********************************
 
 CREATE TABLE `bids_category` (
- `BIDSCategoryID` int(11) NOT NULL AUTO_INCREMENT,
+ `BIDSCategoryID` int(3) NOT NULL AUTO_INCREMENT,
  `ImagingCategory` varchar(10) NOT NULL UNIQUE,    
  PRIMARY KEY (`BIDSCategoryID`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -775,7 +775,7 @@ INSERT INTO `bids_category` (ImagingCategory) VALUES
 
 CREATE TABLE `bids_mri_scan_type_rel` (
   `MRIScanTypeID` int(10) unsigned NOT NULL,
-  `BIDSCategory`varchar(255) DEFAULT NULL,
+  `BIDSCategory`varchar(10) DEFAULT NULL,
   `BIDSScanTypeSubCategory`varchar(255) DEFAULT NULL,
   `BIDSScanType` varchar(255) DEFAULT NULL,
   `BIDSMultiEcho`varchar(255) DEFAULT NULL,
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -114,6 +114,8 @@ DROP TABLE IF EXISTS `tarchive`;
 
 DROP TABLE IF EXISTS bids_mri_scan_type_rel;
 DROP TABLE IF EXISTS bids_category;
+DROP TABLE IF EXISTS bids_scan_type;
+DROP TABLE IF EXISTS bids_scan_type_subcategory;
 
 DROP TABLE IF EXISTS `history`;
 DROP TABLE IF EXISTS `Visit_Windows`;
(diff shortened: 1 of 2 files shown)


SQL/New_patches/2018_08_27_BIDSScanTypeTable.sql
```diff
@@ -0,0 +1,45 @@
...
+INSERT INTO BIDS_mri_scan_type_rel (MRIScanTypeID, BIDSCategory, BIDSScanTypeSubCategory, BIDSScanType, BIDSMultiEcho) SELECT mst.ID, 'func', 'rest', 'bold', NULL FROM mri_scan_type mst WHERE mst.Scan_type = 'fMRI';
+INSERT INTO BIDS_mri_scan_type_rel (MRIScanTypeID, BIDSCategory, BIDSScanTypeSubCategory, BIDSScanType, BIDSMultiEcho) SELECT mst.ID, 'anat', NULL, 'FLAIR', NULL FROM mri_scan_type mst WHERE mst.Scan_type = 'flair';
+INSERT INTO BIDS_mri_scan_type_rel (MRIScanTypeID, BIDSCategory, BIDSScanTypeSubCategory, BIDSScanType, BIDSMultiEcho) SELECT mst.ID, 'anat', NULL, 'T1w', NULL FROM mri_scan_type mst WHERE mst.Scan_type = 't1';
+INSERT INTO BIDS_mri_scan_type_rel (MRIScanTypeID, BIDSCategory, BIDSScanTypeSubCategory, BIDSScanType, BIDSMultiEcho) SELECT mst.ID, 'anat', NULL, 'T2w', NULL FROM mri_scan_type mst WHERE mst.Scan_type = 't2';
+INSERT INTO BIDS_mri_scan_type_rel (MRIScanTypeID, BIDSCategory, BIDSScanTypeSubCategory, BIDSScanType, BIDSMultiEcho) SELECT mst.ID, 'dwi', NULL, 'dwi', NULL FROM mri_scan_type mst WHERE mst.Scan_type = 'dti';
+
+-- CCNA EXAMPLE (THOSE LINES WILL BE REMOVED)
```
MounaSafiHarab: will be removed; I know...
just adding them as an example for now in case whoever is testing needs this example!

SQL/0000-00-00-schema.sql
```diff
@@ -117,6 +117,13 @@ DROP TABLE IF EXISTS `mri_scanner`;
 DROP TABLE IF EXISTS `mri_processing_protocol`;
 DROP TABLE IF EXISTS `ImagingFileTypes`;
 
+DROP TABLE IF EXISTS `tarchive_files`;
+DROP TABLE IF EXISTS `tarchive_series`;
+DROP TABLE IF EXISTS `tarchive`;
```
cmadjar: why are the DROP TABLE statements for the tarchive tables in this PR?

SQL/0000-00-00-schema.sql
```diff
@@ -117,6 +117,13 @@ DROP TABLE IF EXISTS `mri_scanner`;
...
+
+DROP TABLE IF EXISTS BIDS_mri_scan_type_rel;
```
cmadjar: table names should be in loweracase according to the LORIS SQL standard (https://github.com/aces/Loris/blob/minor/docs/SQLModelingStandard.md)

SQL/0000-00-00-schema.sql
```diff
@@ -652,6 +659,46 @@ CREATE TABLE `mri_protocol_checks` (
   PRIMARY KEY (`ID`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
+
+-- ********************************
+-- BIDS tables
+-- ********************************
+
+CREATE TABLE `BIDS_category` (
+ `ImagingCategory` varchar(255) NOT NULL PRIMARY KEY
```
cmadjar: shouldn't there be an ID column in that new table? @PapillonMcGill

cmadjar: also varchar(225) seems a bit much for that column.

SQL/0000-00-00-schema.sql
```diff
@@ -754,6 +757,39 @@ CREATE TABLE `mri_protocol_checks` (
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
 
+-- ********************************
+-- BIDS tables
+-- ********************************
+
+CREATE TABLE `BIDS_category` (
```
cmadjar: table names should be all lowercase snake case according to the LORIS SQL standard (https://github.com/aces/Loris/blob/minor/docs/SQLModelingStandard.md)

SQL/New_patches/2018_08_27_BIDSScanTypeTable.sql
cmadjar: thanks for the examples. Very helpful!!

SQL/0000-00-00-schema.sql
MounaSafiHarab: I checked this document, and it says that exceptions are allowed for MRI acronyms, so I assumed BIDS is an allowed acronym, non?

SQL/0000-00-00-schema.sql
MounaSafiHarab: sorry a mistake copy/paste from lines 100-102 above

SQL/New_patches/2018_08_27_BIDSScanTypeTable.sql
```diff
@@ -0,0 +1,45 @@
...
+  `BIDSMultiEcho`varchar(255) DEFAULT NULL,
+  PRIMARY KEY  (`MRIScanTypeID`),
+  KEY `FK_BIDS_mri_scan_type_rel_1` (`MRIScanTypeID`),
+  CONSTRAINT `FK_BIDS_mri_scan_type_rel` FOREIGN KEY (`MRIScanTypeID`) REFERENCES `mri_scan_type` (`ID`) ON DELETE CASCADE ON UPDATE CASCADE,
+  CONSTRAINT `FK_BIDS_category` FOREIGN KEY (`BIDSCategory`) REFERENCES `BIDS_category`(`ImagingCategory`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+-- Default schema scan types; make some of them named in a BIDS compliant manner
+INSERT INTO BIDS_mri_scan_type_rel (MRIScanTypeID, BIDSCategory, BIDSScanTypeSubCategory, BIDSScanType, BIDSMultiEcho) SELECT mst.ID, 'func', 'rest', 'bold', NULL FROM mri_scan_type mst WHERE mst.Scan_type = 'fMRI';
```
MounaSafiHarab: comment from @cmadjar:
make this task-rest to be consistent with the MRI code

SQL/0000-00-00-schema.sql
MounaSafiHarab: done

SQL/0000-00-00-schema.sql
MounaSafiHarab: I changed it to lowercase in the last commit; so we have both versions, depending on what @PapillonMcGill decides

SQL/0000-00-00-schema.sql
MounaSafiHarab: I added an auto-increment ID column, again just in case @PapillonMcGill says I should (so we have both commits; one without and the last one with)

SQL/0000-00-00-schema.sql
MounaSafiHarab: changed it to `bids` instead in the last commit

SQL/New_patches/2018_08_27_BIDSScanTypeTable.sql
MounaSafiHarab: done

MounaSafiHarab:
I will remove now the example CCNA from the LORIS patch; but @davidblader if you ever need it back, you can check that PR and the earlier commits  would have the INSERT example for CCNA (or check the screenshot I attached to the top of this PR).

@cmadjar you can decide what to do with this patch and the corresponding MRI pull request, and when to remove the Blocked tag...
from my end, it is done!

SQL/0000-00-00-schema.sql
```diff
@@ -754,6 +757,40 @@ CREATE TABLE `mri_protocol_checks` (
...
+
+INSERT INTO `bids_category` (ImagingCategory) VALUES
+      ('anat'),
+      ('func'),
+      ('dwi'),
+      ('fmap');
+
+CREATE TABLE `bids_mri_scan_type_rel` (
+  `MRIScanTypeID` int(10) unsigned NOT NULL,
+  `BIDSCategory`varchar(10) DEFAULT NULL,
```
PapillonMcGill: This should be int(3) and link to the BIDSCategoryID not ImagingCategory. 

Sync update script below.

SQL/0000-00-00-schema.sql
```diff
@@ -754,6 +757,40 @@ CREATE TABLE `mri_protocol_checks` (
...
+  `BIDSScanTypeSubCategory`varchar(255) DEFAULT NULL,
+  `BIDSScanType` varchar(255) DEFAULT NULL,
+  `BIDSMultiEcho`varchar(255) DEFAULT NULL,
+  PRIMARY KEY  (`MRIScanTypeID`),
```
PapillonMcGill: This would create a 1:1 link with mri_scan_type. I.E.: only one BIDSCategory per MRIScanTypeID.  Is this the goal?

SQL/0000-00-00-schema.sql
```diff
@@ -754,6 +757,40 @@ CREATE TABLE `mri_protocol_checks` (
...
+INSERT INTO `bids_category` (ImagingCategory) VALUES
+      ('anat'),
+      ('func'),
+      ('dwi'),
+      ('fmap');
+
+CREATE TABLE `bids_mri_scan_type_rel` (
+  `MRIScanTypeID` int(10) unsigned NOT NULL,
+  `BIDSCategory`varchar(10) DEFAULT NULL,
+  `BIDSScanTypeSubCategory`varchar(255) DEFAULT NULL,
```
PapillonMcGill: look like this could be an enum type, so a look-up table might be appropriate.

SQL/0000-00-00-schema.sql
```diff
@@ -754,6 +757,40 @@ CREATE TABLE `mri_protocol_checks` (
...
+  `BIDSScanType` varchar(255) DEFAULT NULL,
```
PapillonMcGill: This should also be using a look-up table

SQL/0000-00-00-schema.sql
```diff
@@ -754,6 +757,40 @@ CREATE TABLE `mri_protocol_checks` (
...
+  `BIDSMultiEcho`varchar(255) DEFAULT NULL,
```
PapillonMcGill: Is the values options limited to a few predefined choice?  Yes => look-up table

SQL/0000-00-00-schema.sql
```diff
@@ -754,6 +759,98 @@ CREATE TABLE `mri_protocol_checks` (
...
+  `BIDSCategoryID`            int(3)           DEFAULT NULL,
+  `BIDSScanTypeSubCategoryID` int(3)           DEFAULT NULL,
+  `BIDSScanType`              int(3)           DEFAULT NULL,
+  `BIDSEchoNumber`            int(3)           DEFAULT NULL,
+  PRIMARY KEY  (`MRIScanTypeID`),
+  KEY `FK_bids_mri_scan_type_rel` (`MRIScanTypeID`),
+  CONSTRAINT `FK_bids_mri_scan_type_rel`     FOREIGN KEY (`MRIScanTypeID`)             REFERENCES `mri_scan_type` (`ID`) ON DELETE CASCADE ON UPDATE CASCADE,
+  CONSTRAINT `FK_bids_category`              FOREIGN KEY (`BIDSCategoryID`)            REFERENCES `bids_category`(`BIDSCategoryID`),
+  CONSTRAINT `FK_bids_scan_type_subcategory` FOREIGN KEY (`BIDSScanTypeSubCategoryID`) REFERENCES `bids_scan_type_subcategory` (`BIDSScanTypeSubCategoryID`),
+  CONSTRAINT `FK_bids_scan_type`             FOREIGN KEY (`BIDSScanTypeID`)            REFERENCES `bids_scan_type` (`BIDSScanTypeID`),
```
PapillonMcGill: need to remove the comma at the end of the line

cmadjar:
@PapillonMcGill Nice catch!!! I removed the comma at the end of the line. Thanks!

PapillonMcGill:
@cmadjar most of the work goes to Travis.

SQL/0000-00-00-schema.sql
```diff
@@ -754,6 +759,98 @@ CREATE TABLE `mri_protocol_checks` (
...
+  ('FLAIR'),
+  ('T1w'),
+  ('T2w'),
+  ('dwi');
+
+CREATE TABLE `bids_mri_scan_type_rel` (
+  `MRIScanTypeID`             int(10) UNSIGNED NOT NULL,
+  `BIDSCategoryID`            int(3)           DEFAULT NULL,
+  `BIDSScanTypeSubCategoryID` int(3)           DEFAULT NULL,
+  `BIDSScanType`              int(3)           DEFAULT NULL,
```
PapillonMcGill: BIDSScanType**ID**

SQL/0000-00-00-schema.sql
```diff
@@ -754,6 +759,98 @@ CREATE TABLE `mri_protocol_checks` (
...
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+-- Default schema scan types; make some of them named in a BIDS compliant manner
+INSERT INTO bids_mri_scan_type_rel
+  (MRIScanTypeID, BIDSCategoryID, BIDSScanTypeSubCategoryID, BIDSScanTypeID, BIDSEchoNumber)
+  VALUES
+  (
+    (SELECT ID FROM mri_scan_type WHERE Scan_type = 'flair'),
+    (SELECT BIDSCategoryID FROM bids_category WHERE BIDSCategory='anat'),
```
PapillonMcGill: /BIDSCategory/ImagingCategory

and in all other values sets

cmadjar: Good catch!! Should be fixed now.

cmadjar: @PapillonMcGill check the very last commit for those changes. 

I squashed all the other commits into one so I can see what changed in the tables into 2 commits instead of many for the MRI code changes.

SQL/0000-00-00-schema.sql
```diff
@@ -754,6 +759,98 @@ CREATE TABLE `mri_protocol_checks` (
...
+-- BIDS tables
+-- ********************************
+
+CREATE TABLE `bids_category` (
+ `BIDSCategoryID`   int(3)      NOT NULL AUTO_INCREMENT,
+ `BIDSCategoryName` varchar(10) NOT NULL UNIQUE,
+ PRIMARY KEY (`BIDSCategoryID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+INSERT INTO `bids_category` (ImagingCategory) VALUES
```
PapillonMcGill: Field not synch with create table (below too)

SQL/0000-00-00-schema.sql
```diff
@@ -754,6 +759,98 @@ CREATE TABLE `mri_protocol_checks` (
...
+  ('T1w'),
+  ('T2w'),
+  ('dwi');
+
+CREATE TABLE `bids_mri_scan_type_rel` (
+  `MRIScanTypeID`             int(10) UNSIGNED NOT NULL,
+  `BIDSCategoryID`            int(3)           DEFAULT NULL,
+  `BIDSScanTypeSubCategoryID` int(3)           DEFAULT NULL,
+  `BIDSScanTypeID`            int(3)           DEFAULT NULL,
+  `BIDSEchoNumber`            int(3)           DEFAULT NULL,
```
PapillonMcGill: all these IDs should be unsigned (don't forget the referred tables)  
except maybe BIDSEchoNumber. Can we have negative echo number?

Also, not sure exactly after reading past conversation but about BIDSEchoNumber, will the be a single value for each MRIScanTypeID or there might be multiple value for a MRIScanTypeID? It is important for the PK field list.

cmadjar:
@PapillonMcGill Thank you so much for the feedback. Should be fixed now. :)

cmadjar:
@PapillonMcGill I squashed all my commits to see all the changes that were made to the tables at once to update the LORIS-MRI code but it unfortunately dismissed your review. I did not change anything in the code, just squashed all the commits into one. Sorry :S

PapillonMcGill:
@cmadjar  waiting for Travis before I approve. Looks good now!

cmadjar:
@driusan Could you have this in your radar on Monday and merge it if all good for you?

PR #3913, opened 2018-08-30
Test

driusan:
There are multiple PRs where the bvl_feedback test is failing. This is to see if it's just a coincidence.


driusan:
@intralizee Seems the problem is with the individual branches that just happen to both be separately breaking the bvl feedback test

PR #3914, opened 2018-08-30
[Tools/Normalize_consent] Remove validation for consent status "no"

zaliqarosli:
### Brief summary of changes

Initially, we assumed that consent date would only be given if status=yes. Therefore, if status=no but there is a consent date, it is assumed that consent was withdrawn, thus, a withdrawn date is required. The script validates there being a withdrawn date for withdrawn consent. However, IBIS also takes consent date for status=no, breaking down our assumption on withdrawal. This PR removes that validation.

Edit: also adds check for empty consent status as found in IBIS data

### To test this change...

- [ ] @sruthymathew123 running the script on IBIS

### DIFF
--- a/tools/single_use/Normalize_Consent_Data.php
+++ b/tools/single_use/Normalize_Consent_Data.php
@@ -118,15 +118,6 @@ foreach ($consentList as $consentName=>$consentLabel) {
                        [CandID] => " . $candID . "
                        All 'yes' statuses must have a consent date.");
         }
-        // Check if consent status is no and consent date given, but withdrawal date is empty
-        if($status === "no" && !empty($date)) {
-            if(empty($withdrawal)) {
-                array_push($errors, "The date of withdrawal is missing for " . $consentName . ":
-                           [ID]     => " . $consentID . "
-                           [CandID] => " . $candID . "
-                           All 'no' statuses with given consent date must have a withdrawal date.");
-            }
-        }
         // Check for zero dates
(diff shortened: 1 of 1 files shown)


zaliqarosli:
@sruthymathew123 it might be that the CandidateID in the data isn't of type int(6)?

driusan:
@zaliqarosli but it's the Status column that it's complaining about

zaliqarosli:
@driusan oh oops, true 😬

sruthymathew123:
@zaliqarosli The New status column has ENUM ('yes', 'no') where the old consents have enum ('yes', 'no', not_answered'). But I didn't see any not_answered values in any consents .

zaliqarosli:
@sruthymathew123 yeah, the script should catch a "not_answered" status (https://github.com/aces/Loris/blob/bbae90e8c970e40d4833d005641800d8415dea83/tools/single_use/Normalize_Consent_Data.php#L101). are you able to track down which entry the error is for?

sruthymathew123:
@zaliqarosli @driusan 
So Finally sort out the issue; On IBIS, we have some bad data in our database. Even if the default value of consent is supposed to be NULL, some of the consents have blanks('') instead of NULL.

So during the transfer of blank consents to the new table, it throws data truncated error on Status fields which only accepts NULL/yes/no.

@driusan told to fix those bad data from IBIS side; so I will pass the manual tests.

sruthymathew123:
consent_ok

PR #3918, opened 2018-09-04
[JSX] Added Checkbox Element to Form.js

HenriRabalais:
A react component called `Checkbox` Element was added to Form.js. This can be used in any reactified form in LORIS.

Testing instructions coming soon...


jsx/Form.js
```diff
@@ -1285,6 +1285,95 @@ LinkElement.defaultProps = {
...
+      requiredHTML = <span className="text-danger">*</span>;
+    }
+
+    // Add error message
+    if (this.props.errorMessage) {
+      errorMessage = <span>{this.props.errorMessage}</span>;
+      elementClass = 'row form-group has-error';
+    }
+
+    if (this.props.label) {
```
HenriRabalais: This was introduced for potential in-line formatting, in the case where there was no label. However, it is not necessary and label can be made a mandatory field.

zaliqarosli: i agree with label not needing to be a mandatory field

jsx/Form.js
```diff
@@ -1378,6 +1467,9 @@ class LorisElement extends React.Component {
       case 'link':
         elementHtml = (<LinkElement {...elementProps} />);
         break;
+      case 'checkbox':
```
zaliqarosli: case 'advcheckbox'

jsx/Form.js
```diff
@@ -1285,6 +1285,95 @@ LinkElement.defaultProps = {
...
+    }
+
+    return (
+      <div className={elementClass}>
+        {label}
+        <div className={this.props.inputClass}>
+          <input
+            type="checkbox"
+            className="form-control"
+            style={{width: '5%'}}
```
zaliqarosli: we should remove this and make a class with css instead

jsx/Form.js
```diff
@@ -1285,6 +1285,95 @@ LinkElement.defaultProps = {
...
+      );
+    }
+
+    return (
+      <div className={elementClass}>
+        {label}
+        <div className={this.props.inputClass}>
+          <input
+            type="checkbox"
+            className="form-control"
```
zaliqarosli: remove form-control and add input-sm

zaliqarosli:
@driusan changed to an approving review!

PR #3922, opened 2018-09-05
[Data Dictionary] Fix multi-select filtering

llevitis:
### Brief summary of changes
This PR fixes multi-select filtering by making direct comparisons for every filter value (Object type element) instead of checking that the filter value is a substring of one of the column values. 

### This resolves issue...

- [x] https://redmine.cbrain.mcgill.ca/issues/15233

### To test this change...

- [x] @sruthymathew123, please try filtering for the EDI instrument in IBIS.


sruthymathew123:
Before this fix on selecting EDI instrument on IBIS data dictionary it was returning results of both ACE FamilyM**edi**calHistory  and edi :(

before_pr

With this PR, the issue seems to be resolved; so adding 'Passed manual tests' :)

after_liza_fix

tools/DB_date_zeros_removal.php
```diff
@@ -14,35 +14,28 @@
 set_include_path(get_include_path().":".__DIR__."/../project/libraries:".":".__DIR__."/../../php/libraries:");
 require_once __DIR__ . "/../vendor/autoload.php";
 //require_once "NDB_Config.class.inc";
-
```
davidblader: even though the changes in this file aren't really doing anything, you might want to `git checkout --` this file for the sake of best practices.
actual changes above are all good though

PR #3926, opened 2018-09-06
[Tools] Making sure the list of instruments to add is unique in assign_missing_instruments.php

Jkat:
This shouldn't happen in a properly configured LORIS instance but just in case your test_battery/overrides aren't ironed out, you may want to avoid adding duplicate instruments into your flag+instrument tables


kongtiaowang:
LGTM

PR #3927, opened 2018-09-06
[statistics] Fixed undefined index errors with Behavioural

intralizee:
### Brief summary of changes

Inside the Behavioural of the statistics module, undefined index errors were populated and I used isset to catch them.

### This resolves issue...

- [x] Redmine? #
https://redmine.cbrain.mcgill.ca/issues/15255

- [ ] Github? #

### To test this change...

- [ ] Go to Loris->statistics->Behavioural


tools/single_use/Normalize_Consent_Data.php
```diff
@@ -99,8 +99,8 @@
         $consentID  = $entry['ID'];
         $candID     = $entry['CandID'];
         // Check if consent status is not_answered
-        if($status === "not_answered") {
-            array_push($errors, "Deprecated consent status 'not_answered' found for " . $consentName . ":
+        if($status === "not_answered" || $status === "") {
```
driusan: doesn't look like this file is supposed to be part of this PR

modules/statistics/php/stats_behavioural.class.inc
```diff
@@ -160,19 +162,43 @@ class Stats_Behavioural extends \NDB_Form
...
+                $c['all']['total'] += $row['val'];
+            } else {
+                $c['all']['total'] = $row['val'];
+            }
             if ($row['Data_Entry'] == 'Complete') {
-                $c[$vl]['complete']   += $row['val'];
-                $c['all']['complete'] += $row['val'];
+                if (isset($c[$vl]['complete'])) {
+                    $c[$vl]['complete'] += $row['val'];
+                } else {
```
PapillonMcGill: IMHO would have been clearer with:
$c[param]['param2'] = $c[param]['param2'] ?? 0;
$c[param]['param2'] += $row['val'];   // or other math

driusan: I thought of suggesting that, but I think the `?? 0` will only suppress the undefined reference errors on the right hand side of the expression where the coalesce is used, there's still a problem with the right hand where it's 2 levels deep.

johnsaigle:
@intralizee Your tools script changes are showing up in this PR also. You've gotta sort out your branches.

modules/statistics/php/stats_behavioural.class.inc
PapillonMcGill: Yes, of course, the first level need to exist. I made that assumption.

PR #3929, opened 2018-09-06
[instrument_manager] Fix Warning: count(): Parameter must be an array or an object that implements Countable

intralizee:
### Brief summary of changes

Simple fix, check if $feedback has been set with isset in smarty and prevent the error:
Warning: count(): Parameter must be an array or an object that implements Countable

### This resolves issue...

- [ ] Redmine? #

- [ ] Github? #

### To test this change...

- [ ] Visit the instrument manager.


tools/single_use/Normalize_Consent_Data.php
```diff
@@ -99,8 +99,8 @@
         $consentID  = $entry['ID'];
         $candID     = $entry['CandID'];
         // Check if consent status is not_answered
-        if($status === "not_answered") {
-            array_push($errors, "Deprecated consent status 'not_answered' found for " . $consentName . ":
+        if($status === "not_answered" || $status === "") {
```
driusan: This doesn't look like it's in the right PR?

driusan:
@intralizee I think "undo" undid the wrong thing. It now *only* has the Normalize Consent changes and no changes to the instrument manager.

johnsaigle:
@intralizee This PR has one extra file.

intralizee:
@driusan @PapillonMcGill fixed, thanks!

tools/single_use/Normalize_Consent_Data.php
```diff
@@ -1,22 +1,22 @@
 <?php
 /**
- * This script is written for a one time use only to facilitate the normalization of the consent feature of candidate parameters. 
- * 
+ * This script is written for a one time use only to facilitate the normalization of the consent feature of candidate parameters.
```
zaliqarosli: this shouldn't be here

tools/single_use/Normalize_Consent_Data.php
johnsaigle: This still needs to be removed!

tools/single_use/Normalize_Consent_Data.php
```diff
@@ -99,8 +99,8 @@
         $consentID  = $entry['ID'];
         $candID     = $entry['CandID'];
         // Check if consent status is not_answered
-        if($status === "not_answered") {
-            array_push($errors, "Deprecated consent status 'not_answered' found for " . $consentName . ":
+        if($status === "not_answered" || $status === "") {
+            array_push($errors, "Deprecated consent status 'not_answered' or empty string found for " . $consentName . ":
```
johnsaigle: This still needs to be removed. 

@intralizee Please don't manually dismiss the reviews we've given on your PRs containing this file. The reviews will automatically be resolved when the file is actually removed.

intralizee: Hi @johnsaigle the problem is me setting the branch incorrectly.

driusan: @johnsaigle I'm not sure if that's true.. in my experience it's hit or miss whether reviews are autodismissed..

johnsaigle: @driusan Fair enough re the auto-dismissing but I still prefer to resolve my own reviews :man_dancing:

modules/instrument_manager/templates/menu_instrument_manager.tpl
```diff
@@ -17,7 +17,7 @@
     </div>
 </div>
 {/if}
-{if $feedback|count}
+{if isset($feedback) && $feedback|count}
```
johnsaigle: You can actually just remove the count part and check for `isset`. It does the same thing. Not sure why I originally put `count()` in here, maybe it used to be an array.

PR #3931, opened 2018-09-06
[document_repository] Fixed undefined variable and index errors.

intralizee:
### Brief summary of changes

Fixed undefined variable and index errors in the document_repository.

### This resolves issue...

- [x] Redmine? #
https://redmine.cbrain.mcgill.ca/issues/15259

- [ ] Github? #

### To test this change...

- [x] Visit the document repository on a fresh install of Loris.


modules/document_repository/php/document_repository.class.inc
```diff
@@ -167,6 +167,16 @@ class Document_Repository extends \NDB_Menu_Filter
         }
         $fileTypes = array_merge(array(null => 'Any') + $fileTypes);
 
+        $fileCategories = array();
+        $sqlSelect      = "SELECT DISTINCT File_category".
+            " FROM document_repository ORDER BY File_category";
+        $fileC          = $DB->pselect($sqlSelect, array());
```
driusan: I think this whole loop can just be replaced by `pselectCol`

modules/document_repository/php/document_repository.class.inc
```diff
@@ -79,9 +79,9 @@ class Document_Repository extends \NDB_Menu_Filter
                           'v.Date_uploaded',
                          );
 
-        $openAccordion = $_GET['openAccordion'];
+        $openAccordion = $_GET['openAccordion'] ?? '';
         $this->tpl_data['openAccordion'] = $openAccordion;
-        $filtered = $_GET['filtered'];
+        $filtered = $_GET['filtered'] ?? '';
```
johnsaigle: Both of these variables are used as booleans in the template, so this line and 82 should be `?? false;` instead.

driusan: In that case it should probably be using a ternary, not a null coalesce, since `$_GET['filtered']` is a string value.

modules/document_repository/php/document_repository.class.inc
```diff
@@ -167,6 +167,13 @@ class Document_Repository extends \NDB_Menu_Filter
         }
         $fileTypes = array_merge(array(null => 'Any') + $fileTypes);
 
+        $sqlSelect      = "SELECT DISTINCT File_category".
+            " FROM document_repository ORDER BY File_category";
+        $fileCategories = array();
+        try {
+            $fileCategories = $DB->pselectCol($sqlSelect, array());
+        } catch (\DatabaseException $e) {
```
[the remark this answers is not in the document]
driusan: And how could this throw a database exception anyways? It's a very simple hardcoded SQL statement, there's no way for the syntax of it to change dynamically.

modules/document_repository/php/document_repository.class.inc
```diff
@@ -79,9 +79,9 @@ class Document_Repository extends \NDB_Menu_Filter
                           'v.Date_uploaded',
                          );
 
-        $openAccordion = $_GET['openAccordion'];
+        $openAccordion = $_GET['openAccordion'] ?? '';
         $this->tpl_data['openAccordion'] = $openAccordion;
-        $filtered = $_GET['filtered'];
+        $filtered = isset($_GET['filtered']) ? true : false;
```
johnsaigle: This should be switched to `? false : true;` or else the file categories don't appear which changes the default behaviour of the module.

driusan: maybe it should just be `$filtered = !empty($_GET['filtered'])`?

modules/document_repository/php/document_repository.class.inc
```diff
@@ -79,9 +79,9 @@ class Document_Repository extends \NDB_Menu_Filter
                           'v.Date_uploaded',
                          );
 
-        $openAccordion = $_GET['openAccordion'];
+        $openAccordion = $_GET['openAccordion'] ?? '';
         $this->tpl_data['openAccordion'] = $openAccordion;
-        $filtered = $_GET['filtered'];
+        $filtered = !empty($_GET['filtered']) ? true : false;
```
johnsaigle: `$filtered = !empty($_GET['filtered']);` is correct. Don't need the ternary operator now.

tools/single_use/Normalize_Consent_Data.php
```diff
@@ -1,22 +1,22 @@
 <?php
 /**
- * This script is written for a one time use only to facilitate the normalization of the consent feature of candidate parameters. 
- * 
+ * This script is written for a one time use only to facilitate the normalization of the consent feature of candidate parameters.
```
zaliqarosli: not rebased properly?

intralizee: @zaliqarosli I think it's alright, accidentally modified it and undid the changes. Only difference now is my IDE cleaned some extra spaces.

johnsaigle:
@intralizee I think @zaliqarosli meant that the tools script shouldn't be there at all.

driusan:
Try `git checkout bugfix -- tools/single_use/Normalize_Consent_Data.php` to get the exact copy of the script without the changes.

intralizee:
@driusan I tried that just now with the update and that seems to be the result.

intralizee:
@zaliqarosli I don't think #3603 fixes what this PR is intended to fix. I just checkout'ed minor and see the errors that this PR fixes.

zaliqarosli:
@intralizee fair enough for $fileCategories but what about conflicts with lines 82 and 84?

intralizee:
@zaliqarosli It looks like line 84 in my PR is done correct from previous comments, while the other PR will not make it a boolean if undefined. I guess line 82 in my PR needs to be a boolean when undefined and not a string if the other PR is correct. I'll update line 82 to match the other PR. Good Catch!

modules/document_repository/php/document_repository.class.inc
```diff
@@ -79,9 +79,9 @@ class Document_Repository extends \NDB_Menu_Filter
                           'v.Date_uploaded',
                          );
 
-        $openAccordion = $_GET['openAccordion'];
+        $openAccordion = !empty($_GET['openAccordion']);
         $this->tpl_data['openAccordion'] = $openAccordion;
-        $filtered = $_GET['filtered'];
+        $filtered = !empty($_GET['filtered']);
```
johnsaigle: @intralizee @zaliqarosli @driusan Did more testing on this and this will always cause filtering by default (not desired). The smarty template only checks whether `tpl_data['filtered']` is set; it doesn't care whether it's true or false.

Two options here:
1. Revert to my change in #3603 where we set this variable to null if it is not present as a parameter
2. Change the template to pass along the boolean value of `$filtered` instead of just checking if it's set.

I prefer option 2.

intralizee: @johnsaigle Okay so there is no reason to check for isset since $filtered should always be either true or false. I modified the smarty template to remove isset. Thoughts?

johnsaigle: Looks good

PR #3933, opened 2018-09-06
Add instruments module to load instruments through new router

driusan:
This adds a new "instruments" module which takes the place of
what was previously done by NDB_Caller and main.php to load
instruments. The module does not do anything but delegate to
existing instruments that are in the project directory.

To test this: test a variety of instruments after this change,
in both LINST and PHP format. There should be no noticeable
difference.


driusan:
Closing this because it needs more work than I thought to resolve the incompatibility between `setup()` in NDB_Page and `setup($commentID, $pagename)` in NDB_BVL_Instrument (short circuiting out if commentID is null fixes instrument loading, but breaks the create table script..)

PR #3934, opened 2018-09-06
[tools] Move and Clean-up of MyISAM to INNODB scripts

ridz1208:
This PR moves the `Engine_Change_MyISAM_to_INNODB_schemaTables.php` and `Engine_Change_MyISAM_to_INNODB_allTables.php` to the new tools subdirectory and combines them to reduce code duplication. 

A new feature was also added to run the change on a single table in the DB 

This PR also cleans up the scripts and adds an automatic datacorruption validation check, it compares the values before and after the conversion of the tables and makes sure the data is the same.

A backup of the tables is still highly recommended before running either of these scripts

Testing instructions:
- make sure you have MyISAM tables in the DB (if not the script will throw an error and die)
- run the script using the `tosql` option with each of the possible actions `schemaTables`, `allTables` and/or a specific table name for any table that has a MyISAM engine. Check the output SQL after each run to make sure the SQL commands are correct. (don't run the SQL commands on the DB if you do not wish to reset the database before the next step)
- run the script using the `confirm` option with each of the possible actions (this will generate the SQL and run it so this will make changes to the DB for sure. I suggest starting by testing the single table action, then the schema then the all this way you can test all options without ever needing to reset the DB.
- add the Pass manual tests flag

Notice: possible memory issues if the table being converted is too large.


tools/generic_includes.php
```diff
@@ -23,15 +23,21 @@
...
+/**
+ * Helper function that writes a string supplied in the parameters to a filepath
+ * specified in the parameters using the indicated mode. This function should be
+ * used as a standard in the tools PHP scripts.
+ *
+ * @param string $filePath Desired path for the generated file (including extension)
+ * @param string $output   String to be written to the file specified in $filePath
+ * @param string $mode     Creation mode of the file
+ */
+function writeToFile(string $filePath, string $output, string $mode = "w")
```
driusan: I don't think this should be in the generic_includes file, it's specific to a couple scripts and if we ever want to modernize our code mixing function definitions with side-effects in the same file violates PSR-1 ("Basic Coding Standards").

"Files SHOULD either declare symbols (classes, functions, constants, etc.) or cause side-effects (e.g. generate output, change .ini settings, etc.) but SHOULD NOT do both."

tools/single_use/Engine_Change_MyISAM_to_INNODB_allTables.php
```diff
@@ -0,0 +1,157 @@
+<?php
+/**
+ * Script changing all table engines in the database to InnoDB
+ *
+ * PHP Version 7
+ *
+ * @category Main
+ * @package  Loris
+ * @author   Various <example@example.com>
```
driusan: This doesn't seem accurate

tools/single_use/Engine_Change_MyISAM_to_INNODB_allTables.php
```diff
@@ -0,0 +1,157 @@
+<?php
+/**
+ * Script changing all table engines in the database to InnoDB
```
driusan: I think it would be a good idea to have the ability to do one table at a time. Instrument tables are going to be a pain if it's all or nothing.

tools/single_use/Engine_Change_MyISAM_to_INNODB_allTables.php
ridz1208: anonymous ?

tools/generic_includes.php
ridz1208: can I create a tools helper class then ?

tools/single_use/Engine_Change_MyISAM_to_INNODB_allTables.php
ridz1208: one table at a time could be done with a single SQL query.. I can try adding that

tools/generic_includes.php
driusan: Sure.. (in fact, I think you need to because as far as I know you can't autoload functions in PHP.)

tools/single_use/Engine_Change_MyISAM_to_INNODB_allTables.php
driusan: @ridz1208 ?

ridz1208: ew no!

tools/generic_includes.php
ridz1208: it works though

tools/single_use/Engine_Change_MyISAM_to_INNODB.php
```diff
@@ -0,0 +1,212 @@
...
+    $output .= "ALTER TABLE `".$table."` ENGINE=INNODB;\n";
+}
+
+$output .="SET FOREIGN_KEY_CHECKS=@OLD_FOREIGN_KEY_CHECKS; \n";
+// END OUTPUT
+
+// START ACTION
+if ($printToSQL) {
+    // Generate the file
+    FileHelper::writeToFile($filePath, $output);
```
xlecours: I don't think you need a new helper for that. There is the SplFileObject if you want to use a OO method.
```
try {
    (new SplFileObject($filePath, "w"))->fwrite($output);
} catch (\RuntimeException $e) {
    // Most likely permission denied
    echo $e->getMessage();
}
```

ridz1208: good to know

ridz1208:
@xlecours made changes

Issue #3935, opened 2018-09-07
[Database] "feedback_bvl_type" and "feedback_bvl_thread" - Cannot add or update a child row: a foreign key constraint fails


intralizee:
I found issues attempting to "leave New profile level feedback" for a candidate. 

Steps to reproduce: 
1) Clean install of Loris
2) Import the test/RBdata.sql

Issues:
1) The mysql table "feedbback_bvl_thread" has a column named "Feedback_type" and where references are made to the table "feedbback_bvl_type" the column Feedback_type. I'm noticing we only have the values 3,4,5 in "Feedback_type" and it's missing 1 & 2. The test data from RBdata.sql inserts values of 1,2,3,4,5 in the Feedback_type column of the table feedback_bvl_thread.
2) When trying to leave a Feedback type of "Input Errors" the value of 1 for Feedback_type will be used but result in a  Warning: PDOStatement::execute(): SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`loris`.`feedback_bvl_thread`, CONSTRAINT `FK_feedback_bvl_thread_1` FOREIGN KEY (`Feedback_type`) REFERENCES `feedback_bvl_type` (`feedback_type`)) in php/libraries/Database.class.inc on line 358. The error is happening because the values of 1 & 2 are missing in the feedback_bvl_type table for the column Feedback_type with the Name & Description details.

Thoughts?

Jkat:
Our default `SQL/0000-00-00-schema.sql` has the following:
```
INSERT INTO `feedback_bvl_type` (Name, Description) VALUES
  ('Input','Input Errors'),
  ('Scoring','Scoring Errors');
```
which should give you IDs of 1 and 2, and when you import `test/RBdata.sql` it does 
```
DELETE FROM `feedback_bvl_type`;
INSERT INTO `feedback_bvl_type` (Name, Description) VALUES
  ('Input','Input Errors'),
  ('Scoring','Scoring Errors'),
  ('Other','Other');
```
which probably gave you IDs of 3, 4, and 5, instead of 1, 2, and 3, ideally.

`feedback_bvl_type` should instead be fixed to `truncate` before the insertion in `test/RBdata.sql`.
I'm not too familiar with this `test/RBdata.sql` but I'd bet that there may be other similar instances in there.

driusan:
Resolved by #3936

PR #3936, opened 2018-09-07
[RBdata.sql] Fixed feedback_bvl_type a foreign key constraint fails

intralizee:
### Brief summary of changes

Fixed an issue when using the RBdata.sql that inserts test data into Loris.
See the GitHub issue for more details.

### This resolves issue...

- [ ] Redmine? #

- [x] Github? #
https://github.com/aces/Loris/issues/3935

### To test this change...

- [ ] Install Loris, query the RBdata.sql and view the tables "feedback_bvl_type" and "feedback_bvlthread" for the Feedback_type reference to match.


test/RBdata.sql
```diff
@@ -2278,6 +2278,7 @@ UNLOCK TABLES;
 
 
 DELETE FROM `feedback_bvl_type`;
+TRUNCATE table feedback_bvl_type;
```
PapillonMcGill: no need to DELETE if using TRUNCATE.

Should use backtick around table name

kongtiaowang:
It worked.

PR #3937, opened 2018-09-07
[bvl_feedback] NDB_BVL_Feedback.class.inc fixed FeedbackID

intralizee:
### Brief summary of changes

This PR fixes the NDB_BVL_Feedback.class.inc and where the FeedbackID for a candidate feedback might've been wrong depending on the situation. See line number 720 in the php/libbraries/NDB_BVL_Feedbback.class.inc

### This resolves issue...

- [ ] Redmine? #

- [ ] Github? #

### To test this change...

- [ ] Leave feedback for a candidate.


php/libraries/NDB_BVL_Feedback.class.inc
```diff
@@ -713,6 +713,13 @@ class NDB_BVL_Feedback
             );
         }
 
+        $nextID = $DB->pselect('SELECT FeedbackID from feedback_bvl_thread ORDER BY FeedbackID DESC LIMIT 0, 1', null);
```
xlecours: `SELECT MAX(FeedbackID) +1 FROM feedback_bvl_threa` is, in my opinion, simpler to understand.

xlecours: And to get rid of the next if/else block:
`select IFNULL(MAX(FeedbackID), 0) + 1 as FeedbackID from feedback_bvl_thread;`

php/libraries/NDB_BVL_Feedback.class.inc
```diff
@@ -713,6 +713,13 @@ class NDB_BVL_Feedback
...
+        if ($nextID == null) {
+            $nextID = 0;
+        } else {
+            $nextID = $nextID[0]['FeedbackID'] + 1;
+        }
+        $setArray['FeedbackID'] = $nextID;
```
xlecours: But I don't understand why the autoincrement need to be bypassed here.

intralizee:
@xlecours You're right, this change is not needed.

PR #3938, opened 2018-09-07
[Media] Move getFormattedColumn method into the class that uses it

driusan:
This does some slight cleanup of media module by moving the column formatter into the
class that uses it, rather than using a naked function.

Also cleaned up by removing the pointless "loris.hiddenHeaders" global variable.


intralizee:
@driusan LGTM

modules/media/jsx/mediaIndex.js
```diff
@@ -56,6 +54,65 @@ class MediaIndex extends React.Component {
...
+    // If a column if set as hidden, don't display it
+    if (hiddenHeaders.indexOf(column) > -1) {
+      return null;
+    }
+
+    // Create the mapping between rowHeaders and rowData in a row object.
+    let row = {};
+    rowHeaders.forEach(function(header, index) {
+      row[header] = rowData[index];
+    }, this);
```
HenriRabalais: Not sure if we are moving towards an arrow function standard, but it may be useful to start writing functions like this as:
`rowHeaders.forEach((header, index) => row[header] = rowData[index]);`

driusan: Does it make sense to use an arrow function here? It's setting data, not being used as a lambda function.

I'll change it if the people who do more javascript than I do think it's the right way to go (@intralizee @davidblader @anyoneelsethatimmissing)

modules/media/jsx/mediaIndex.js
```diff
@@ -56,6 +54,65 @@ class MediaIndex extends React.Component {
...
+   * @param {string} cell - cell content
+   * @param {array} rowData - array of cell contents for a specific row
+   * @param {array} rowHeaders - array of table headers (column names)
+   * @return {*} a formated table cell for a given column
+   */
+  formatColumn(column, cell, rowData, rowHeaders) {
+    const hiddenHeaders = ['Cand ID', 'Session ID', 'Hide File', 'File Type'];
+
+    // If a column if set as hidden, don't display it
+    if (hiddenHeaders.indexOf(column) > -1) {
```
zaliqarosli: @um4r12 you can do this for your hiddenheaders in https://github.com/aces/Loris/pull/3796

modules/media/jsx/mediaIndex.js
driusan: I changed it since it's already 2-0 in favour of the arrow function..

driusan:
:(

zaliqarosli:
@driusan we are confused as to why that is. we manually changed your code back and the headers are still broken :(

zaliqarosli:
travis tells us this but we don't understand: `Arrow function should not return assignmentno-return-assign`

zaliqarosli:
@driusan actually, removing the loris.hiddenHeaders breaks how StatisDataTable renders. it was just falsely working before because your second commit didn't include the newly compiled js file

HenriRabalais:
Alright, so @zaliqarosli and I have been manually this PR, and we were having some trouble with the `hidden headers`. After some inquiry, we discovered the root of the bug that was occurring is that there are two instances of `loris.hiddenHeaders` on lines `348` and `350` of `staticDataTable.js`. This is causing a discrepancy wherein the data of hidden headers disappears, but the headers themselves remain. We suggest passing the `hiddenHeaders` array as a prop to the `StaticDataTable` component, and changing instances of `loris.hiddenHeaders` to `this.props.hiddenHeaders`. This may unintentionally affect other modules as well, which can either be addressed in this or another PR.

modules/media/jsx/mediaIndex.js
davidblader: after some consideration, i think conventional function syntax should be the way to go

zaliqarosli: @driusan use of arrow function here is fine - there's just a syntax error. there should be an extra ; before the last closing parenthesis i.e.

`rowHeaders.forEach((header, index) => {
          row[header] = rowData[index];
});`

but the closing semi-colon is missing here after the command

zaliqarosli: @HenriRabalais

kongtiaowang:
screen shot 2018-09-17 at 1 40 28 pm

HenriRabalais:
Blocked by #3955.

modules/media/jsx/mediaIndex.js
```diff
@@ -101,7 +156,7 @@ class MediaIndex extends React.Component {
             Data={this.state.Data.Data}
```
HenriRabalais: As of PR #3955, hiddenHeaders need to be passed as `props` to `StaticDataTable`.

modules/media/jsx/mediaIndex.js
```diff
@@ -56,6 +54,63 @@ class MediaIndex extends React.Component {
...
+  /**
+   * Modify behaviour of specified column cells in the Data Table component
+   * @param {string} column - column name
+   * @param {string} cell - cell content
+   * @param {array} rowData - array of cell contents for a specific row
+   * @param {array} rowHeaders - array of table headers (column names)
+   * @return {*} a formated table cell for a given column
+   */
+  formatColumn(column, cell, rowData, rowHeaders) {
+    const hiddenHeaders = ['Cand ID', 'Session ID', 'Hide File', 'File Type'];
```
HenriRabalais: As of PR #3955, hiddenHeaders should now be set and accessed as a state within this component.

zaliqarosli: correct PR is https://github.com/aces/Loris/pull/3967 and was issued to major. changes might not be relevant to current PR which is going to minor

modules/media/jsx/mediaIndex.js
driusan: Done

driusan:
@zaliqarosli As far as I can tell I've fixed this and rebased onto major, but I don't have the media module setup to test..

zaliqarosli:
@driusan you still have conflicts

driusan:
@zaliqarosli I don't know how that happened, but I rebased (again..)

zaliqarosli:
duplicate PR but on minor https://github.com/aces/Loris/pull/3957 - needs resolving

driusan:
Closed in favour of #3957

PR #3941, opened 2018-09-07
[New Feature] Adding the SQL patch for the EEG tables

cmadjar:
### Brief summary of changes

This PR includes the SQL patch to be run to create the EEG tables for the insertion scripts of the LORIS-MRI. Note that these tables will be used by the new Electrophysiology Browser module that will be in a separate PR.

This PR goes hand in hand with the following LORIS-MRI PR: https://github.com/aces/Loris-MRI/pull/337

For people that are visual, here is a screenshot summarizing the table structure for the EEG tables:

overall tables

### This resolves issue...

- [ ] Redmine? https://redmine.cbrain.mcgill.ca/issues/13603

### To test this change...

- Run the SQL patch and makes sure all statements work properly
- If you have an empty database, run the default schema with the new changes to make sure all statements work properly
- Feel free to make sure that the tables follow the LORIS SQL standard (crossing my fingers that I did not omit anything there!)

### DIFF
--- a/SQL/0000-00-05-ElectrophysiologyTables.sql
+++ b/SQL/0000-00-05-ElectrophysiologyTables.sql
@@ -1,0 +1,289 @@
+--
+-- Table structure for the electrophysiology component of LORIS
+--
+
+
+-- Create a physiological_modality table
+CREATE TABLE `physiological_modality` (
+  `PhysiologicalModalityID` INT(5) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `PhysiologicalModality`   VARCHAR(50)         NOT NULL UNIQUE,
+  PRIMARY KEY (`PhysiologicalModalityID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+-- Create a physiological_output_type table
+CREATE TABLE `physiological_output_type` (
+  `PhysiologicalOutputTypeID` INT(5)      UNSIGNED NOT NULL AUTO_INCREMENT,
+  `OutputTypeName`            VARCHAR(20)          NOT NULL UNIQUE,
+  `OutputTypeDescription`     VARCHAR(255),
+  PRIMARY KEY (`PhysiologicalOutputTypeID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+-- Create a physiological_file table
+-- Note that the field InsertedByUser refers to the Linux User that ran the script
+-- and not a LORIS user
+CREATE TABLE `physiological_file` (
+  `PhysiologicalFileID`       INT(10)    UNSIGNED  NOT NULL      AUTO_INCREMENT,
+  `PhysiologicalModalityID`   INT(5)     UNSIGNED  DEFAULT NULL,
+  `PhysiologicalOutputTypeID` INT(5)     UNSIGNED  NOT NULL,
+  `SessionID`                 INT(10)    UNSIGNED  NOT NULL,
+  `InsertTime`                TIMESTAMP            NOT NULL      DEFAULT CURRENT_TIMESTAMP,
+  `FileType`                  VARCHAR(12)          DEFAULT NULL,
+  `AcquisitionTime`           DATETIME             DEFAULT '1970-01-01 00:00:01',
+  `InsertedByUser`            VARCHAR(50)          NOT NULL,
+  `FilePath`                  VARCHAR(255)         NOT NULL,
+  PRIMARY KEY (`PhysiologicalFileID`),
+  CONSTRAINT `FK_session_ID`
+    FOREIGN KEY (`SessionID`)
+    REFERENCES `session` (`ID`),
+  CONSTRAINT `FK_ImagingFileTypes_type`
+    FOREIGN KEY (`FileType`)
+    REFERENCES `ImagingFileTypes` (`type`),
+  CONSTRAINT `FK_phys_modality_ModID`
+    FOREIGN KEY (`PhysiologicalModalityID`)
+    REFERENCES `physiological_modality` (`PhysiologicalModalityID`),
+  CONSTRAINT `FK_phys_output_type_TypeID`
+    FOREIGN KEY (`PhysiologicalOutputTypeID`)
+    REFERENCES `physiological_output_type` (`PhysiologicalOutputTypeID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+
+-- Create a physiological_parameter_file table that will store all JSON
+-- information that accompanies the BIDS physiological dataset
+CREATE TABLE `physiological_parameter_file` (
+  `PhysiologicalParameterFileID` INT(10) UNSIGNED NOT NULL  AUTO_INCREMENT,
+  `PhysiologicalFileID`          INT(10) UNSIGNED NOT NULL,
+  `ParameterTypeID`              INT(10) UNSIGNED NOT NULL,
+  `InsertTime`                   TIMESTAMP        NOT NULL  DEFAULT CURRENT_TIMESTAMP,
+  `Value`                        VARCHAR(255),
+  PRIMARY KEY (`PhysiologicalParameterFileID`),
+  CONSTRAINT `FK_phys_file_FileID`
+    FOREIGN KEY (`PhysiologicalFileID`)
+    REFERENCES `physiological_file` (`PhysiologicalFileID`)
+    ON DELETE CASCADE,
+  CONSTRAINT `FK_param_type_ParamTypeID`
+    FOREIGN KEY (`ParameterTypeID`)
+    REFERENCES `parameter_type` (`ParameterTypeID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+
+-- Create a physiological_status_type table
+CREATE TABLE `physiological_status_type` (
+  `PhysiologicalStatusTypeID` INT(5)      UNSIGNED NOT NULL AUTO_INCREMENT,
+  `ChannelStatus`             VARCHAR(10)          NOT NULL UNIQUE,
+  PRIMARY KEY (`PhysiologicalStatusTypeID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+
+-- Create a physiological_channel_type table
+CREATE TABLE `physiological_channel_type` (
+  `PhysiologicalChannelTypeID` INT(5)       UNSIGNED NOT NULL      AUTO_INCREMENT,
+  `ChannelTypeName`            VARCHAR(255)          NOT NULL      UNIQUE,
+  `ChannelDescription`         VARCHAR(255)          DEFAULT NULL,
+  PRIMARY KEY (`PhysiologicalChannelTypeID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+
+-- Create a physiological_channel table that will store all information from
+CREATE TABLE `physiological_channel` (
+  `PhysiologicalChannelID`     INT(10)    UNSIGNED NOT NULL      AUTO_INCREMENT,
+  `PhysiologicalFileID`        INT(10)    UNSIGNED NOT NULL,
+  `PhysiologicalChannelTypeID` INT(5)     UNSIGNED NOT NULL,
+  `PhysiologicalStatusTypeID`  INT(5)     UNSIGNED DEFAULT NULL,
+  `InsertTime`                 TIMESTAMP           NOT NULL      DEFAULT CURRENT_TIMESTAMP,
+  `Name`                       VARCHAR(50)         NOT NULL,
+  `Description`                VARCHAR(255)        DEFAULT NULL,
+  `SamplingFrequency`          INT(6)              DEFAULT NULL,
+  `LowCutoff`                  DECIMAL(8,3)        DEFAULT NULL,
+  `HighCutoff`                 DECIMAL(8,3)        DEFAULT NULL,
+  `ManualFlag`                 DECIMAL(5,4)        DEFAULT NULL,
+  `Notch`                      INT(6)              DEFAULT NULL,
+  `Reference`                  VARCHAR(255)        DEFAULT NULL,
+  `StatusDescription`          VARCHAR(255)        DEFAULT NULL,
+  `Unit`                       VARCHAR(255)        DEFAULT NULL,
+  `FilePath`                   VARCHAR(255)        DEFAULT NULL,
+  PRIMARY KEY (`PhysiologicalChannelID`),
+  CONSTRAINT `FK_phys_file_FileID_2`
+    FOREIGN KEY (`PhysiologicalFileID`)
+    REFERENCES `physiological_file` (`PhysiologicalFileID`)
+    ON DELETE CASCADE,
+  CONSTRAINT `FK_phys_channel_type_TypeID`
+    FOREIGN KEY (`PhysiologicalChannelTypeID`)
+    REFERENCES `physiological_channel_type` (`PhysiologicalChannelTypeID`),
+  CONSTRAINT `FK_phys_status_type_TypeID`
+    FOREIGN KEY (`PhysiologicalStatusTypeID`)
+    REFERENCES `physiological_status_type` (`PhysiologicalStatusTypeID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+-- Create physiololgical_electrode_type table
+CREATE TABLE `physiological_electrode_type` (
+  `PhysiologicalElectrodeTypeID` INT(5) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `ElectrodeType`                VARCHAR(50)     NOT NULL UNIQUE,
+  PRIMARY KEY (`PhysiologicalElectrodeTypeID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+-- Create physiological_electrode_material table
+CREATE TABLE `physiological_electrode_material` (
+  `PhysiologicalElectrodeMaterialID` INT(5) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `ElectrodeMaterial`                VARCHAR(50)     NOT NULL UNIQUE,
+  PRIMARY KEY (`PhysiologicalElectrodeMaterialID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+-- Create a physiological_electrode table
+CREATE TABLE `physiological_electrode` (
+  `PhysiologicalElectrodeID`          INT(10) UNSIGNED NOT NULL      AUTO_INCREMENT,
+  `PhysiologicalFileID`               INT(10) UNSIGNED NOT NULL,
+  `PhysiologicalElectrodeTypeID`      INT(5)  UNSIGNED DEFAULT NULL,
+  `PhysiologicalElectrodeMaterialID`  INT(5)  UNSIGNED DEFAULT NULL,
+  `InsertTime`                        TIMESTAMP        NOT NULL      DEFAULT CURRENT_TIMESTAMP,
+  `Name`                              VARCHAR(50)      NOT NULL,
+  `X`                                 DECIMAL(12,6)    NOT NULL,
+  `Y`                                 DECIMAL(12,6)    NOT NULL,
+  `Z`                                 DECIMAL(12,6)    NOT NULL,
+  `Impedance`                         INT(10)          DEFAULT NULL,
+  `FilePath`                          VARCHAR(255)     DEFAULT NULL,
+  PRIMARY KEY (`PhysiologicalElectrodeID`),
+  CONSTRAINT `FK_phys_file_FileID_3`
+    FOREIGN KEY (`PhysiologicalFileID`)
+    REFERENCES `physiological_file` (`PhysiologicalFileID`)
+    ON DELETE CASCADE,
+  CONSTRAINT `FK_phys_elec_type_ID`
+    FOREIGN KEY (`PhysiologicalElectrodeTypeID`)
+    REFERENCES `physiological_electrode_type` (`PhysiologicalElectrodeTypeID`),
+  CONSTRAINT `FK_phys_elec_material_ID`
+  FOREIGN KEY (`PhysiologicalElectrodeMaterialID`)
+  REFERENCES `physiological_electrode_material` (`PhysiologicalElectrodeMaterialID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+
+-- Create physiological_task_event table that will store all information
+-- regarding the task executed during the physiological recording
+CREATE TABLE `physiological_task_event` (
+  `PhysiologicalTaskEventID` INT(10) UNSIGNED NOT NULL      AUTO_INCREMENT,
+  `PhysiologicalFileID`      INT(10) UNSIGNED NOT NULL,
+  `InsertTime`               TIMESTAMP        NOT NULL      DEFAULT CURRENT_TIMESTAMP,
+  `Onset`                    DECIMAL(11,6)    NOT NULL,
+  `Duration`                 DECIMAL(11,6)    NOT NULL,
+  `EventCode`                INT(10)          DEFAULT NULL,
+  `EventValue`               INT(10)          DEFAULT NULL,
+  `EventSample`              INT(10)          DEFAULT NULL,
+  `EventType`                VARCHAR(50)      DEFAULT NULL,
+  `TrialType`                VARCHAR(255)     DEFAULT NULL,
+  `ResponseTime`             TIME             DEFAULT NULL,
+  `FilePath`                 VARCHAR(255)     DEFAULT NULL,
+  PRIMARY KEY (`PhysiologicalTaskEventID`),
+  CONSTRAINT `FK_phys_file_FileID_4`
+    FOREIGN KEY (`PhysiologicalFileID`)
+    REFERENCES `physiological_file` (`PhysiologicalFileID`)
+    ON DELETE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+
+-- Create physiological_archive which will store archives of all the files for
+-- Front-end download
+CREATE TABLE `physiological_archive` (
+  `PhysiologicalArchiveID`   INT(10) UNSIGNED NOT NULL   AUTO_INCREMENT,
+  `PhysiologicalFileID`      INT(10) UNSIGNED NOT NULL,
+  `InsertTime`               TIMESTAMP        NOT NULL   DEFAULT CURRENT_TIMESTAMP,
+  `Blake2bHash`              VARCHAR(128)     NOT NULL,
+  `FilePath`                 VARCHAR(255)     NOT NULL,
+  PRIMARY KEY (`PhysiologicalArchiveID`),
+  CONSTRAINT `FK_phys_file_FileID_5`
+    FOREIGN KEY (`PhysiologicalFileID`)
+    REFERENCES `physiological_file` (`PhysiologicalFileID`)
+    ON DELETE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+-- Insert into physiological_output_type
+INSERT INTO physiological_output_type
+  (`OutputTypeName`, `OutputTypeDescription`)
+  VALUES
+  ('raw',         'raw dataset'),
+  ('derivatives', 'derivative/processed dataset');
+
+-- Insert into physiological_channel_type
+INSERT INTO physiological_channel_type
+  (`ChannelTypeName`, `ChannelDescription`)
+  VALUES
+  ('EEG',              'ElectoEncephaloGram: EEG sensors'                    ),
+  ('VEOG',             'Vertical ElectroOculoGram (eyes)'                    ),
+  ('HEOG',             'Horizontal ElectOculoGram'                           ),
+  ('EOG',              'Generic EOG channel, if HEOG or VEOG information not available'),
+  ('ECG',              'ElectroCardioGram (heart)'                           ),
+  ('EMG',              'ElectroMyoGram (muscle)'                             ),
+  ('TRIG',             'System Triggers'                                     ),
+  ('REF',              'Reference electrode'                                 ),
+  ('GRD',              'Ground electrode'                                    ),
+  ('MISC',             'Miscellaneous'                                       ),
+  ('MEGMAG',           'MEG magnetometer'                                    ),
+  ('MEGGRADAXIAL',     'MEG axial gradiometer'                               ),
+  ('MEGGRADPLANAR',    'MEG planar gradiometer'                              ),
+  ('MEGGREFMAG',       'MEG reference magnetometer'                          ),
+  ('MEGREFGRADAXIAL',  'MEG reference axial gradiometer'                     ),
+  ('MEGREFGRADPLANAR', 'MEG reference planar gradiometer'                    ),
+  ('MEGOTHER',         'Any other type of MEG sensor'                        ),
+  ('ECOG',             'Electrode channel: electrocortigogram (intracranial)'),
+  ('SEEG',             'Electrode channel: stereo-electroencephalogram (intracranial)'),
+  ('DBS',              'Electrode channel: deep brain stimulation (intracranial)'),
+  ('AUDIO',            'Audio signal'                                        ),
+  ('PD',               'Photodiode'                                          ),
+  ('EYEGAZE',          'Eye Tracker gaze'                                    ),
+  ('PUPIL',            'Eye Tracker pupil diameter'                          ),
+  ('SYSCLOCK',         'System time showing elapsed time since trial started'),
+  ('ADC',              'Analog to Digital input'                             ),
+  ('DAC',              'Digital to Analog input'                             ),
+  ('HLU',              'Measure position of head and head coils'             ),
+  ('FITERR',           'Fit error signal from each head localization coil'   ),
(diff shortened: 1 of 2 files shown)


cmadjar:
@driusan @PapillonMcGill if you could review the SQL structure, that would be great.

Also, I did not include the new tables in the default schema. Should I add it to the default schema or create a 0000-00-05_EEG_table_structure? Let me know.

Thanks!

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+-- Create a physiological_file table
+CREATE TABLE `physiological_file` (
+  `PhysiologicalFileID`       INT(10)    UNSIGNED  NOT NULL      AUTO_INCREMENT,
+  `PhysiologicalModalityID`   TINYINT(3) UNSIGNED  DEFAULT NULL,
+  `Caveat`                    TINYINT(1) UNSIGNED  DEFAULT 0,
+  `PhysiologicalOutputTypeID` INT(5)     UNSIGNED  NOT NULL,
+  `SessionID`                 INT(10)    UNSIGNED  NOT NULL,
+  `InsertTime`                INT(10)    UNSIGNED  NOT NULL,
+  `FileType`                  VARCHAR(12)          DEFAULT NULL,
+  `AcquisitionTime`           VARCHAR(20)          DEFAULT NULL,
```
PapillonMcGill: Is this use to store a date and time? If yes, datetime format would be better

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+  `InsertedByUser`            VARCHAR(50)          NOT NULL,
```
PapillonMcGill: Is this a Loris user? if yes, UserID would be better

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+
+
+-- Create a physiological_file table
+CREATE TABLE `physiological_file` (
+  `PhysiologicalFileID`       INT(10)    UNSIGNED  NOT NULL      AUTO_INCREMENT,
+  `PhysiologicalModalityID`   TINYINT(3) UNSIGNED  DEFAULT NULL,
+  `Caveat`                    TINYINT(1) UNSIGNED  DEFAULT 0,
+  `PhysiologicalOutputTypeID` INT(5)     UNSIGNED  NOT NULL,
+  `SessionID`                 INT(10)    UNSIGNED  NOT NULL,
+  `InsertTime`                INT(10)    UNSIGNED  NOT NULL,
```
PapillonMcGill: Is this use to store a time in hour:minutes:second format? if yes, time type would be better.

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
+-- Create a physiological_modality table
+CREATE TABLE `physiological_modality` (
+  `PhysiologicalModalityID` TINYINT UNSIGNED NOT NULL AUTO_INCREMENT,
```
PapillonMcGill: should specify tinyint(3) to be as in physiological_file.

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+  `PhysiologicalModality`   VARCHAR(50)      NOT NULL,
+  PRIMARY KEY (`PhysiologicalModalityID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+-- Create a physiological_output_type table
+CREATE TABLE `physiological_output_type` (
+  `PhysiologicalOutputTypeID` INT(5) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `OutputType`                VARCHAR(20)     NOT NULL,
```
PapillonMcGill: OutputTypeName maybe?

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+
+
+
+-- Create a physiological_parameter_file table that will store all JSON
+-- information that accompanies the BIDS physiological dataset
+CREATE TABLE `physiological_parameter_file` (
+  `PhysiologicalParameterFileID` INT(10) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `PhysiologicalFileID`          INT(10) UNSIGNED NOT NULL,
+  `ParameterTypeID`              INT(10) UNSIGNED NOT NULL,
+  `InsertTime`                   INT(10) UNSIGNED NOT NULL,
```
PapillonMcGill: same comment as for physiological_file.

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+  `Value`                        VARCHAR(255),
```
PapillonMcGill: Could you be more precise for "value"?

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+  `ChannelStatus`             VARCHAR(10)         NOT NULL,
+  PRIMARY KEY (`PhysiologicalStatusTypeID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+
+-- Create a physiological_channel_type table
+CREATE TABLE `physiological_channel_type` (
+  `PhysiologicalChannelTypeID` TINYINT(3) UNSIGNED NOT NULL      AUTO_INCREMENT,
+  `ChannelType`                VARCHAR(255)        NOT NULL,
```
PapillonMcGill: ChannelTypeName?

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+  `PhysiologicalChannelTypeID` TINYINT(3) UNSIGNED NOT NULL,
+  `PhysiologicalStatusTypeID`  TINYINT(2) UNSIGNED DEFAULT NULL,
+  `Name`                       VARCHAR(50)         NOT NULL,
+  `Description`                VARCHAR(255)        DEFAULT NULL,
+  `SamplingFrequency`          INT(6)              DEFAULT NULL,
+  `LowCutoff`                  DECIMAL(8,3)        DEFAULT NULL,
+  `HighCutoff`                 DECIMAL(8,3)        DEFAULT NULL,
+  `ManualFlag`                 DECIMAL(5,4)        DEFAULT NULL,
+  `Notch`                      INT(6)              DEFAULT NULL,
+  `StatusDescription`          VARCHAR(255)        DEFAULT NULL,
```
PapillonMcGill: Should that field be in physiological_status_type instead?

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+  `Unit`                       VARCHAR(255)        DEFAULT NULL,
+  `File`                       VARCHAR(255)        DEFAULT NULL,
```
PapillonMcGill: How does this differ from the File field in physiological_file?

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+-- Create a physiological_electrode table
+CREATE TABLE `physiological_electrode` (
+  `PhysiologicalElectrodeID` INT(10) UNSIGNED NOT NULL      AUTO_INCREMENT,
+  `PhysiologicalFileID`      INT(10) UNSIGNED NOT NULL,
+  `Name`                     VARCHAR(50)      NOT NULL,
+  `X`                        DECIMAL(12,6)    NOT NULL,
+  `Y`                        DECIMAL(12,6)    NOT NULL,
+  `Z`                        DECIMAL(12,6)    NOT NULL,
+  `Type`                     VARCHAR(50)      NOT NULL,
+  `Material`                 VARCHAR(50)      NOT NULL,
```
PapillonMcGill: Should the Type and Material attributes be in a separate lookup table?

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+  `File`                     VARCHAR(255)     DEFAULT NULL,
```
PapillonMcGill: How does this differ from the File field in physiological_file?

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+-- Create physiological_task_event table that will store all information
+-- regarding the task executed during the physiological recording
+CREATE TABLE `physiological_task_event` (
+  `PhysiologicalTaskEventID` INT(10) UNSIGNED NOT NULL       AUTO_INCREMENT,
+  `PhysiologicalFileID`      INT(10) UNSIGNED NOT NULL,
+  `Onset`                    DECIMAL(11,6)    NOT NULL,
+  `Duration`                 DECIMAL(11,6)    NOT NULL,
+  `TrialType`                VARCHAR(255)     DEFAULT NULL,
+  `ResponseTime`             TIME             DEFAULT NULL,
+  `EventCode`                VARCHAR(255)     DEFAULT NULL,
```
PapillonMcGill: a 255 char length for code? I need more details

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+  `Sample`                   VARCHAR(255)     DEFAULT NULL,
+  `File`                     VARCHAR(255)     DEFAULT NULL,
```
PapillonMcGill: How does this differ from the File field in physiological_file?

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+
+-- Create physiological_archive which will store archives of all the files for
+-- Front-end download
+CREATE TABLE `physiological_archive` (
+  `PhysiologicalArchiveID`   INT(10) UNSIGNED NOT NULL       AUTO_INCREMENT,
+  `PhysiologicalFileID`      INT(10) UNSIGNED NOT NULL,
+  `InsertTime`               INT(10) UNSIGNED NOT NULL,
```
PapillonMcGill: datetime format maybe?

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+CREATE TABLE `physiological_modality` (
+  `PhysiologicalModalityID` TINYINT UNSIGNED NOT NULL AUTO_INCREMENT,
+  `PhysiologicalModality`   VARCHAR(50)      NOT NULL,
+  PRIMARY KEY (`PhysiologicalModalityID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+-- Create a physiological_output_type table
+CREATE TABLE `physiological_output_type` (
+  `PhysiologicalOutputTypeID` INT(5) UNSIGNED NOT NULL AUTO_INCREMENT,
```
driusan: There seems to be a lot of inconsistencies with the precision of the primary keys.. why are some int(5), some tinyint, some int(10), etc?

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
+-- Create a physiological_modality table
+CREATE TABLE `physiological_modality` (
+  `PhysiologicalModalityID` TINYINT UNSIGNED NOT NULL AUTO_INCREMENT,
+  `PhysiologicalModality`   VARCHAR(50)      NOT NULL,
```
driusan: Are these required to be unique?

SQL/New_patches/2018-09-07_EEG_tables.sql
driusan: Shouldn't this have an default of the current time on insert? (same with all the other InsertTime fields)

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+
+-- Create a physiological_file table
+CREATE TABLE `physiological_file` (
+  `PhysiologicalFileID`       INT(10)    UNSIGNED  NOT NULL      AUTO_INCREMENT,
+  `PhysiologicalModalityID`   TINYINT(3) UNSIGNED  DEFAULT NULL,
+  `Caveat`                    TINYINT(1) UNSIGNED  DEFAULT 0,
+  `PhysiologicalOutputTypeID` INT(5)     UNSIGNED  NOT NULL,
+  `SessionID`                 INT(10)    UNSIGNED  NOT NULL,
+  `InsertTime`                INT(10)    UNSIGNED  NOT NULL,
+  `FileType`                  VARCHAR(12)          DEFAULT NULL,
```
driusan: Should this be a foreign key to another table? Are these the same set of "file types" as the imaging browser, or different types?

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+  `AcquisitionTime`           VARCHAR(20)          DEFAULT NULL,
+  `InsertedByUser`            VARCHAR(50)          NOT NULL,
+  `File`                      VARCHAR(255)         NOT NULL,
```
driusan: Filename, maybe?

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+    REFERENCES `physiological_modality` (`PhysiologicalModalityID`),
+  CONSTRAINT `FK_physiological_output_type_PhysiologicalOutputTypeID`
+    FOREIGN KEY (`PhysiologicalOutputTypeID`)
+    REFERENCES `physiological_output_type` (`PhysiologicalOutputTypeID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+
+-- Create a physiological_parameter_file table that will store all JSON
+-- information that accompanies the BIDS physiological dataset
```
driusan: Is this only for JSON? Or is it parsed into a different structure? What's the format of this? Is there one or many per file? If it's just for storing JSON, it makes more sense to use a JSON column type somewhere than to emulate the parameter_file table.

SQL/New_patches/2018-09-07_EEG_tables.sql
driusan: Is 255 chars big enough for all values that might be parsed?

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+  `Description`                VARCHAR(255)        DEFAULT NULL,
+  `SamplingFrequency`          INT(6)              DEFAULT NULL,
+  `LowCutoff`                  DECIMAL(8,3)        DEFAULT NULL,
+  `HighCutoff`                 DECIMAL(8,3)        DEFAULT NULL,
+  `ManualFlag`                 DECIMAL(5,4)        DEFAULT NULL,
+  `Notch`                      INT(6)              DEFAULT NULL,
+  `StatusDescription`          VARCHAR(255)        DEFAULT NULL,
+  `Unit`                       VARCHAR(255)        DEFAULT NULL,
+  `File`                       VARCHAR(255)        DEFAULT NULL,
+  `SoftwareFilters`            VARCHAR(255)        DEFAULT NULL,
```
driusan: What's the format of this? By the name of it it sounds like it should be normalized into a different table.

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+
+-- Insert into physiological_status_type
+INSERT INTO physiological_status_type
+  (ChannelStatus)
+  VALUES
+  ('good'),
+  ('bad');
+
+-- Insert into ImagingFileTypes
+INSERT INTO ImagingFileTypes
```
driusan: Is this table shared with the imaging browser? If not the name is misleading (also, the capitalization style is different than all the other tables in this PR.)

SQL/New_patches/2018-09-07_EEG_tables.sql
cmadjar: oups. Looks like it is an integer. Not sure how big it can be though. Setting it to 10 for now

SQL/New_patches/2018-09-07_EEG_tables.sql
cmadjar: this one stores the BIDS events.tsv file path while physiological_file stores the path to the actual EEG recording.

SQL/New_patches/2018-09-07_EEG_tables.sql
cmadjar: see comment above on the same topic

SQL/New_patches/2018-09-07_EEG_tables.sql
cmadjar: Indeed, this is the same table than the one used for the MRI side. It did not make any sense to me to create an additional file type table just for EEG. The capitalization is due to the fact that this is what it is currently called. Renaming and normalization of that table should probably be independent of this PR.

SQL/New_patches/2018-09-07_EEG_tables.sql
cmadjar: mmm... Good point. The problem is that I have no idea what should go in those lookup tables as it is not specified in the BIDS spec. Maybe I could indeed have look up tables and insert into them each time there is something new for type and material with the insert scripts? @PapillonMcGill What do you think?

SQL/New_patches/2018-09-07_EEG_tables.sql
PapillonMcGill: Normally for PK, it is preferable no to limit to much the size on the attributes. People can add/remove/add a lot, specially during testing phase. If auto_increment is not reseted,  you migth risk reaching the limit.
If you plan to have about 20~25 values, should not be a problem. But my great friend Murphy tend to show up at the least expected occasion. How about INT(4), that would leave 65K possibilities.

SQL/New_patches/2018-09-07_EEG_tables.sql
PapillonMcGill: thanks!

SQL/New_patches/2018-09-07_EEG_tables.sql
PapillonMcGill: Can we ask an EEG expert or do a little search? I would like to know more. Specially if a single Type always have the same Material. It might also be very limited but new type show up once in a while when manufacturer introduce them. In that case a lookup table that simply add a missing value on the go

SQL/New_patches/2018-09-07_EEG_tables.sql
cmadjar: @PapillonMcGill Perfect! Will change it to INT(4) then

cmadjar: Done in the next commit!!

SQL/New_patches/2018-09-07_EEG_tables.sql
cmadjar: OK. I will contact the EEG people I know and let you know what they say. Thanks!

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -66,27 +66,27 @@ CREATE TABLE `physiological_parameter_file` (
...
 
 
 -- Create a physiological_channel table that will store all information from
 CREATE TABLE `physiological_channel` (
-  `PhysiologicalChannelID`     INT(10) UNSIGNED    NOT NULL      AUTO_INCREMENT,
-  `PhysiologicalFileID`        INT(10) UNSIGNED    NOT NULL,
+  `PhysiologicalChannelID`     INT(10)    UNSIGNED NOT NULL      AUTO_INCREMENT,
+  `PhysiologicalFileID`        INT(10)    UNSIGNED NOT NULL,
   `PhysiologicalChannelTypeID` TINYINT(3) UNSIGNED NOT NULL,
   `PhysiologicalStatusTypeID`  TINYINT(2) UNSIGNED DEFAULT NULL,
```
PapillonMcGill: These 2 need to be synchronized with referred tables

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,258 @@
...
+  `OutputTypeName`            VARCHAR(20)          NOT NULL UNIQUE,
+  `OutputTypeDescription`     VARCHAR(255),
+  PRIMARY KEY (`PhysiologicalOutputTypeID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+-- Create a physiological_file table
+CREATE TABLE `physiological_file` (
+  `PhysiologicalFileID`       INT(10)    UNSIGNED  NOT NULL      AUTO_INCREMENT,
+  `PhysiologicalModalityID`   TINYINT(3) UNSIGNED  DEFAULT NULL,
```
PapillonMcGill: need to be sync

cmadjar:
@PapillonMcGill Thank you!! Corrected the keys in the other table to be in sync with primary keys INT(5)

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -18,8 +18,8 @@ CREATE TABLE `physiological_output_type` (
 -- Create a physiological_file table
 CREATE TABLE `physiological_file` (
   `PhysiologicalFileID`       INT(10)    UNSIGNED  NOT NULL      AUTO_INCREMENT,
-  `PhysiologicalModalityID`   TINYINT(3) UNSIGNED  DEFAULT NULL,
-  `Caveat`                    TINYINT(1) UNSIGNED  DEFAULT 0,
+  `PhysiologicalModalityID`   INT(5)     UNSIGNED  DEFAULT NULL,
+  `Caveat`                    INT(5)     UNSIGNED  DEFAULT 0,
```
PapillonMcGill: Is this a yes/no field?

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,259 @@
...
+-- Create a physiological_file table
+CREATE TABLE `physiological_file` (
+  `PhysiologicalFileID`       INT(10)    UNSIGNED  NOT NULL      AUTO_INCREMENT,
+  `PhysiologicalModalityID`   INT(5)     UNSIGNED  DEFAULT NULL,
+  `Caveat`                    INT(5)     UNSIGNED  DEFAULT 0,
+  `PhysiologicalOutputTypeID` INT(5)     UNSIGNED  NOT NULL,
+  `SessionID`                 INT(10)    UNSIGNED  NOT NULL,
+  `InsertTime`                INT(10)    UNSIGNED  NOT NULL      DEFAULT CURRENT_TIMESTAMP,
+  `FileType`                  VARCHAR(12)          DEFAULT NULL,
+  `AcquisitionTime`           DATETIME             DEFAULT NULL,
```
PapillonMcGill: null datetime are not accepted. Please make a default value of '1970-01-01 00:00:01. See https://github.com/aces/Loris/pull/3709

SQL/New_patches/2018-09-07_EEG_tables.sql
cmadjar: Actually, this won't be used in the end. Thanks for bringing that up!! I did not see it anymore... Will remove it in the next commit.

SQL/New_patches/2018-09-07_EEG_tables.sql
cmadjar: @PapillonMcGill Done. Thank you!

PapillonMcGill:
waiting for more details on table physiological_electrode before future approving/future comments

SQL/New_patches/2018-09-07_EEG_tables.sql
cmadjar: @PapillonMcGill I had news regarding these data types from the EEG experts.

"There is no official names for type and materials because manufacturers can change as they see fit."

Based on that reply, I could either create two look up tables (one for type and one for material) and insert in them each time the insertion script encounter a new type (or new material) or leave it to VARCHAR. Let me know what you think.

I also asked them questions about new fields that are present in the BIDS structure and will add them in the next commit.

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,265 @@
+-- Create a physiological_modality table
+CREATE TABLE `physiological_modality` (
+  `PhysiologicalModalityID` INT(5) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `PhysiologicalModality`   VARCHAR(50)         NOT NULL UNIQUE,
+  PRIMARY KEY (`PhysiologicalModalityID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
driusan: in mysql "utf8" isn't. These needs to be "utf8mb4" in order to be utf8. ..

cmadjar: @driusan Thanks! Done.

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -0,0 +1,264 @@
...
+  `PhysiologicalChannelID`     INT(10)    UNSIGNED NOT NULL      AUTO_INCREMENT,
+  `PhysiologicalFileID`        INT(10)    UNSIGNED NOT NULL,
+  `PhysiologicalChannelTypeID` INT(5)     UNSIGNED NOT NULL,
+  `PhysiologicalStatusTypeID`  INT(5)     UNSIGNED DEFAULT NULL,
+  `InsertTime`                 TIMESTAMP           NOT NULL      DEFAULT CURRENT_TIMESTAMP,
+  `Name`                       VARCHAR(50)         NOT NULL,
+  `Description`                VARCHAR(255)        DEFAULT NULL,
+  `SamplingFrequency`          INT(6)              DEFAULT NULL,
+  `LowCutoff`                  DECIMAL(8,3)        DEFAULT NULL,
+  `HighCutoff`                 DECIMAL(8,3)        DEFAULT NULL,
```
cmadjar: @PapillonMcGill I have a user that has float values and the string "Inf" for that field in their channels.tsv file. His file has been created directly from a edf converter so we might end up with other users having that field populated to a number or "Inf".
Not sure what would be the best option here. Maybe convert the field to a varchar so we could store the float and "Inf" value?

What are your thoughts regarding this?

PapillonMcGill: After investigation on the subject and discussion with @cmadjar, recommendation is to use 99999.999 to indicate "inf" in the value. Code will need to support the bi-directional conversion.

SQL/New_patches/2018-09-07_EEG_tables.sql
PapillonMcGill: What type of value.

SQL/New_patches/2018-09-07_EEG_tables.sql
PapillonMcGill: As theses 2 parameters seem to be at manufacturer discretion and new type and material can appear as they introduce new products, they have to be open for additional value on the fly.
But they will probably be from a limited list, thus repeating values.

Having a reference table with a pre-defined list from which to take value as long as the code let the user add new values. 
This would ensure consistency in name and would facilitate adding to a list of predefined value without having to modify the code with each new addition.

SQL/New_patches/2018-09-07_EEG_tables.sql
cmadjar: This change is supported by the insertion script now. Currently, this value is not displayed in the Front End but we will keep this in mind. Thanks!

SQL/New_patches/2018-09-07_EEG_tables.sql
cmadjar: It could be anything (blake2b of the registered file, information stored in the side car JSON file etc...). Basically, this is the `Value` associated to the `ParameterTypeID` present in the `parameter_type` table where the description and name of the parameter is stored.

SQL/New_patches/2018-09-07_EEG_tables.sql
cmadjar: Sure! I can implement that. Will be doing that today.

SQL/New_patches/2018-09-07_EEG_tables.sql
PapillonMcGill: Ok then, a generic attribute name is justified.

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -113,26 +113,44 @@ CREATE TABLE `physiological_channel` (
     REFERENCES `physiological_status_type` (`PhysiologicalStatusTypeID`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
+-- Create physiololgical_electrode_type table
+CREATE TABLE `physiological_electrode_type` (
+  `PhysiologicalElectrodeTypeID` INT(5) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `ElectrodeType`                VARCHAR(50),
```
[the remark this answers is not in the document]
cmadjar: done!

SQL/New_patches/2018-09-07_EEG_tables.sql
```diff
@@ -113,26 +113,44 @@ CREATE TABLE `physiological_channel` (
...
+  PRIMARY KEY (`PhysiologicalElectrodeTypeID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
+-- Create physiological_electrode_material table
+CREATE TABLE `physiological_electrode_material` (
+  `PhysiologicalElectrodeMaterialID` INT(5) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `ElectrodeMaterial`                VARCHAR(50),
```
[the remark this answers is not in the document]
cmadjar: done

SQL/New_patches/2018-09-07_EEG_tables.sql
driusan: "longtext" is only needed to store gigabytes of a data. https://stackoverflow.com/questions/13932750/tinytext-text-mediumtext-and-longtext-maximum-storage-sizes#13932834

Let's leave it varchar(255) for now and if experience shows that to be too short we can convert it to text later..

SQL/New_patches/2018-09-07_EEG_tables.sql
driusan: seems to still be utf8, not utf8mb4? Did you forget to push?

SQL/New_patches/2018-09-07_EEG_tables.sql
driusan: Can you add a comment to that effect in the SQL file? It's not ideal, but I don't know where else to put it and it's a little confusing.

cmadjar: sure. Will do in the next commit!

SQL/New_patches/2018-09-07_EEG_tables.sql
cmadjar: sounds good. Thanks!

SQL/New_patches/2018-09-07_EEG_tables.sql
cmadjar: yeah, we have been experiencing with Mélanie and Tom regarding this. I did change everything to `utf8mb4` but since `physiological_file` is referencing to the `session` and `ImagingFileTypes` tables that are `utf8`, the foreign keys cannot be created and it fails creating that table and all tables that references it.
We also tried specifying the charset in the attribute and it failed too. We did not find any solution that seemed to work unless putting all table to utf8 like the rest of the schema.
We tried but in the end we had to revert back to utf8... :(

driusan: I think the right change is to fix the session and ImagingFileTypes tables, then, but okay, we'll leave it for now.

driusan:
@cmadjar Is the "Blocked" label still valid or can I merge this?

cmadjar:
@driusan It was missing the update to the schema as I was waiting for all the comments to be resolved to include it. As per your suggestion, I added a `0000-00-05-electrophysiologyTables.sql` file for the default schema. It is basically a copy of the file in the `New_patches` folder.

@thomasbeaudry  tested the patch with the import script on the LORIS-MRI side that goes with this PR. It works for his project as well as two other tests datasets that I have. It can probably be merged I guess given that it works for 3 different projects.

SQL/0000-00-05-ElectrophysiologyTables.sql
```diff
@@ -0,0 +1,289 @@
...
+  ('meg'),
+  ('ieeg');
+
+
+-- Insert into physiological_status_type
+INSERT INTO physiological_status_type
+  (ChannelStatus)
+  VALUES
+  ('good'),
+  ('bad');
```
christinerogers: null?

cmadjar: This is a lookup table so there is no need to NULL values. However, in the physiological_channel table (that uses this lookup table), NULL values are allowed. Hope this answers your question.

PR #3942, opened 2018-09-07
[Issue Tracker] Remove superglobal references and do slight cleanup of routing

driusan:
This does some very mild cleanup of the issue_tracker module to take advantage
of the LORIS router. It handles issues at the endpoint `/issue_tracker/issue/$ID`
rather than `/issue_tracker/issue/?issueID=ID`. In doing this, it also gets
rid of the references to the $_GET superglobal in the issue page code.
    
This should be extended to handle PUT and GET requests and negotiate content
type via the HTTP request headers, but as is serves as a very simple example
of how to do basic routing on a page.


ridz1208:
For future notice, this introduced a bug where when clicking to edit and existing issue the user gets redirected to create a new issue instead. a bugfix will be submitted for version 20.3

PR #3943, opened 2018-09-08
[create timepoint] Fix error in logic when projects are enabled

ridz1208:
### Brief summary of changes
Shifting calls a bit to rectify fault in logic. If projects are disabled all subrojects should be displayed by default when creating a timepoint (this was and still is working). When projects are enabled and the `project_rel` has data for all projects only the designated subprojects should be displayed in the dropdown when creating a timepoint (this was and still is working). When projects are enabled but the `project_rel` table IS NOT populated, no options should be displayed when creating a timepoint; this PR fixes that, before this PR the dropdown would default to all subprojects.

### To test this change...
- empty `project_rel` table
- enable `useProjects`
- go to create_timepoint
- the dropdown should be empty

### DIFF
--- a/modules/create_timepoint/php/create_timepoint.class.inc
+++ b/modules/create_timepoint/php/create_timepoint.class.inc
@@ -116,7 +116,7 @@ class Create_Timepoint extends \NDB_Form
         if ($config->getSetting('useProjects') === 'true') {
             // If projects enabled, loop through and add valid subprojects only
             $subprojList = $candidate->getValidSubprojects();
-            if (empty($allSubprojects)) {
+            if (empty($subprojList)) {
                 $this->form->errors['subprojectID'] = "No subprojects have been 
                 defined for the project this candidate is affiliated with. 
                 If you are an administrator, please use the Configuration module to 
--- a/modules/create_timepoint/php/create_timepoint.class.inc
+++ b/modules/create_timepoint/php/create_timepoint.class.inc
@@ -104,13 +104,24 @@ class Create_Timepoint extends \NDB_Form
 
         $candidate =& \Candidate::singleton($this->identifier);
         // List of all subprojects from config file
-        $allSubprojects  = \Utility::getSubprojectList();
+        $allSubprojects = \Utility::getSubprojectList();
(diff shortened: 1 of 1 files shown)


driusan:
@ridz1208 Wasn't there a decision a long time ago to remove the "EnableProjects" config and just require everyone to have at least one project (with a default name of "loris")?

sruthymathew123:
rida_fix

driusan:
@sruthymathew123 That screenshot looks like a bug, doesn't it? Shouldn't there be at least one subproject in order to create the timepoint?

sruthymathew123:
@driusan I followed the test conditions mentioned by Rida. He says when you emptied your proj_rel table and  useProjects is enabled; the Subproject dropdown in Create time point should be empty.
@rida Please correct me if I am wrong.

ridz1208:
@sruthymathew123 you are correct. @driusan it wont let you create the timepoint without a subproject. so in order to create a timepoint you will need to populate the project_rel table or disable use projects. thats the goal

xlecours:
@ridz1208 Could we add a message in case the drop down is empty? Something like : "in order to create a timepoint you will need to populate the project_rel table or disable use projects"

ridz1208:
@xlecours I dont mind adding it.

@driusan any oppositions/ better way of handling ?

modules/create_timepoint/php/create_timepoint.class.inc
```diff
@@ -102,27 +102,34 @@ class Create_Timepoint extends \NDB_Form
...
-        // List of all subprojects from config file
-
-        // Loop through the subprojects to get an id out and to create
-        // the subproject drop down.
-        $allSubprojects  = \Utility::getSubprojectList();
-        $sp_labelOptions = array(null => '');
-
-        foreach ($allSubprojects as $subprojectID => $title) {
-            if (! empty($subprojList)) {
+            if (empty($allSubprojects)) {
```
PapillonMcGill: Should not that be if (empty( $subprojList)) ?

ridz1208:
@driusan travis acting weird, the tests passed

PR #3944, opened 2018-09-10
[Server Processes Manager] Prevent OS command injection

johnsaigle:
### Brief summary of changes

The arguments being used in getShellCommand() are user-supplied and need to be checked. This PR validates and sanitizes the values that are eventually sent to the dangerous `exec` function.

A proof-of-concept exploit is available upon request.

### To test this change...

- [ ] Serves processes manager should work the same before and after.

Come see me for information if you want to see or learn about the exploit.

### DIFF
--- a/modules/server_processes_manager/php/mriuploadserverprocess.class.inc
+++ b/modules/server_processes_manager/php/mriuploadserverprocess.class.inc
@@ -139,7 +139,7 @@ class MriUploadServerProcess extends AbstractServerProcess
     {
         // Sanitize variables before using in exec statement.
         $env            = $this->_mriCodePath . '/' . self::ENVIRONMENT_BASENAME;
-        $env            = escapeshellarg($envFile);
+        $env            = escapeshellarg($env);
         $uploadScript   = $this->_mriCodePath . '/' . self::IMAGING_UPLOAD_FILE_PATH;
         $uploadScript   = escapeshellarg($uploadScript);
         $uploadId       = escapeshellarg($this->_mriUploadId);
--- a/modules/server_processes_manager/php/mriuploadserverprocess.class.inc
+++ b/modules/server_processes_manager/php/mriuploadserverprocess.class.inc
@@ -105,6 +105,18 @@ class MriUploadServerProcess extends AbstractServerProcess
             throw $exception;
         }
 
+        // Check that paths exist before using.
+        if (!file_exists($this->_mriCodePath)) {
+            throw new \LorisException(
+                "Source code for imaging upload could not be found!"
+            );
+        }
+        if (!is_dir($this->_mriCodePath)) {
+            throw new \ConfigurationException(
+                "Value for MRI code path is not a directory!"
+            );
+        }
+
         $this->_mriUploadId    = $mriUploadId;
         $this->_sourceLocation = $sourceLocation;
 
@@ -130,17 +142,6 @@ class MriUploadServerProcess extends AbstractServerProcess
      */
     public function getShellCommand()
     {
-        // Check that paths exist before using.
-        if (!is_dir($this->_mriCodePath)) {
-            throw new \ConfigurationException(
-                "Value for MRI code path is not a directory!"
-            );
-        }
-        if (!file_exists($this->_mriCodePath)) {
-            throw new \LorisException(
-                "Source code for imaging upload could not be found!"
-            );
-        }
         // Sanitize variables before using in exec statement.
         $codePath       = escapeshellarg($this->_mriCodePath);
         $uploadId       = escapeshellarg($this->_mriUploadId);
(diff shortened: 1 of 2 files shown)


modules/server_processes_manager/php/mriuploadserverprocess.class.inc
```diff
@@ -130,10 +130,25 @@ class MriUploadServerProcess extends AbstractServerProcess
...
-               . "{$this->_mriCodePath}/" . self::IMAGING_UPLOAD_FILE_PATH
-               . " -upload_id "    . escapeshellarg($this->_mriUploadId)
-               . " -profile prod " . escapeshellarg($this->_sourceLocation);
+        // Check that paths exist before using.
+        if (!is_dir($this->_mriCodePath)) {
+            throw new \ConfigurationException(
+                "Value for MRI code path is not a directory!"
+            );
+        }
+        if (!file_exists($this->_mriCodePath)) {
```
PapillonMcGill: Checking if the files exist after it as been check if it is a dir will never fail.
if it doesn't exist, it will fail previous test (is_dir).

either logic need to be change or ordering need to be changed

johnsaigle:
@xlecours The database isn't inherently a safe source of data, especially for Config settings that can be changed by a user in the front-end. An exploit is possible here.

I will re-order the file checking stuff :+1:

modules/server_processes_manager/php/mriuploadserverprocess.class.inc
```diff
@@ -105,6 +105,18 @@ class MriUploadServerProcess extends AbstractServerProcess
...
 
+        // Check that paths exist before using.
+        if (!file_exists($this->_mriCodePath)) {
+            throw new \LorisException(
+                "Source code for imaging upload could not be found!"
+            );
+        }
+        if (!is_dir($this->_mriCodePath)) {
+            throw new \ConfigurationException(
+                "Value for MRI code path is not a directory!"
```
driusan: Why is one of these a ConfigurationException but the other is a LorisException? In fact, I'm not sure why the file_exists one is even necessary. If it doesn't exist it's not a directory and the message is still valid.

johnsaigle: Sure, I'll remove the file check.

Jkat: Yea I was about to say file_exists checks for directory as well, but I'm not knowledgeable enough to know what we want to expect for `$this->_mriCodePath`, whether we want it as a directory or we allow a file..

johnsaigle: @Jkat Yeah I think the file_exists check was unnecessary and didn't make sense. I think I left it in from an earlier phase of the fix. `$this->_mriCodePath` should be the parent folder of the perl script that does image uploading.

johnsaigle:
@xlecours @PapillonMcGill The changes that I made for @driusan removed your approving reviews, but I believe this PR is still good to go? Thanks for reviewing!

modules/server_processes_manager/php/mriuploadserverprocess.class.inc
```diff
@@ -130,10 +137,14 @@ class MriUploadServerProcess extends AbstractServerProcess
      */
     public function getShellCommand()
     {
-        return "source {$this->_mriCodePath}/" . self::ENVIRONMENT_BASENAME . " ; "
-               . "{$this->_mriCodePath}/" . self::IMAGING_UPLOAD_FILE_PATH
-               . " -upload_id "    . escapeshellarg($this->_mriUploadId)
-               . " -profile prod " . escapeshellarg($this->_sourceLocation);
+        // Sanitize variables before using in exec statement.
+        $codePath       = escapeshellarg($this->_mriCodePath);
```
driusan: I think this is wrong. This variable seems to be used to construct another path (or two), and the whole argument(s) needs to be escaped after concatenation of the relevant parts, just just a subset of it.

modules/server_processes_manager/php/mriuploadserverprocess.class.inc
```diff
@@ -138,11 +138,14 @@ class MriUploadServerProcess extends AbstractServerProcess
     public function getShellCommand()
     {
         // Sanitize variables before using in exec statement.
-        $codePath       = escapeshellarg($this->_mriCodePath);
+        $env            = $this->_mriCodePath . '/' . self::ENVIRONMENT_BASENAME;
+        $env            = escapeshellarg($envFile);
```
PapillonMcGill: s/$envFile/$env

johnsaigle:
I'm not sure why this is failing Travis. it seems like the page isn't loading but I'm not having any problems on my VM. Could someone else help test this?

driusan:
@kongtiaowang can you give @johnsaigle a hand?

kongtiaowang:
There is no MRICodePath folder in the Travis_CI.
For passing the Travis_CI, it needs to add two lines 83- 84 [[      
- mkdir -p data/loris/bin/mri
- mysql LorisTest -e "UPDATE Config SET Value='data/loris/bin/mri' WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='MRICodePath')"       ]]  into travis.yml.

PR #3947, opened 2018-09-11
[Core: Database] Add type check to HTMLEscapeArray

johnsaigle:
### Brief summary of changes

In [the process](https://github.com/aces/Loris/pull/3878) of adding return types and type declarations to the
Database class, the LORIS integration tests failed when an int value was
passed to htmlspecialchars.

This PR checks that inputs to that function are of string type before
passing them as arguments.
A comment is added to this function to let developers know that
htmlspecialchars as called in the DB is not a magic bullet to preventing
JS injection.

### To test this change...

Functionality should remain the same.

### DIFF
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -315,7 +315,7 @@ class Database
         $retVal = array();
         foreach ($set as $key => $val) {
             // Only call htmlspecialchars on strings, or else an error occurs.
-            $retVal[$key] = is_string($val)? htmlspecialchars($val) : $val;
+            $retVal[$key] = is_string($val) ? htmlspecialchars($val) : $val;
         }
         return $retVal;
     }
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -297,30 +297,25 @@ class Database
     }
 
     /**
-     * Escapes the HTML from an array that is about to be inserted into the database
+     * Escape the HTML from an array that is about to be inserted into the
+     * database.
+     * NOTE this only provide safety when elements are inserted into the
+     * "innerHTML" of an element; if a value is instead injected
+     * into attribute such as 'name', 'value', 'class', etc.
+     * then Javascript injection is still possible.
+     *
+     * This function alone goes not guarantee safety.
      *
      * @param array $set The array to be inserted as a new row
      *
-     * @return A copy of $set with the HTML characters escaped
+     * @return array A copy of $set with the HTML characters escaped
      */
     private function _HTMLEscapeArray($set)
     {
         $retVal = array();
         foreach ($set as $key => $val) {
             // Only call htmlspecialchars on strings, or else an error occurs.
-            if (is_string($val)) {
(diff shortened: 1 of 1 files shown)


php/libraries/Database.class.inc
```diff
@@ -307,10 +307,23 @@ class Database
     {
         $retVal = array();
         foreach ($set as $key => $val) {
-            if ($val === 'Unrated') {
-                $retVal[$key] = '';
+            // Only call htmlspecialchars on strings, or else an error occurs.
+            if (is_string($val)) {
+                if ($val === 'Unrated') {
```
driusan: since this is going to major maybe it's time to get rid of this "Unrated" hack and see what breaks (I'm not sure what if was ever there for, but if anything I suspect the imaging browser..)

php/libraries/Database.class.inc
```diff
@@ -297,21 +297,25 @@ class Database
...
      * @param array $set The array to be inserted as a new row
      *
-     * @return A copy of $set with the HTML characters escaped
+     * @return array A copy of $set with the HTML characters escaped
      */
     private function _HTMLEscapeArray($set)
     {
         $retVal = array();
         foreach ($set as $key => $val) {
-            if ($val === 'Unrated') {
```
kongtiaowang: If remove this "unrated" check, it can't save the null value for QC status.

php/libraries/Database.class.inc
```diff
@@ -297,21 +297,25 @@ class Database
...
-                $retVal[$key] = '';
-            } else {
-                $retVal[$key] = is_null($val) ? null : htmlspecialchars($val);
-            }
+            // Only call htmlspecialchars on strings, or else an error occurs.
+            $retVal[$key] = is_string($val)? htmlspecialchars($val) : $val;
```
paiva: A minor thing, but would be nice to add a space between `is_string($val)` and `?`

php/libraries/Database.class.inc
johnsaigle: @driusan What do you want to do here?

driusan: I guess you can leave it for the sake of this PR, but we should make an issue to fix the imaging browser at the source.

johnsaigle:
@driusan @kongtiaowang The 'unrated' check has been added back.

johnsaigle:
@kongtiaowang Could you dismiss your old review if you're satisfied with the changes?

php/libraries/Database.class.inc
```diff
@@ -297,21 +297,33 @@ class Database
...
+     * then Javascript injection is still possible.
+     *
+     * This function alone does not guarantee safety.
      *
      * @param array $set The array to be inserted as a new row
      *
-     * @return A copy of $set with the HTML characters escaped
+     * @return array A copy of $set with the HTML characters escaped
      */
     private function _HTMLEscapeArray($set)
```
driusan: Shouldn't the return type be declared in the signature too?

johnsaigle: Yep!

PR #3949, opened 2018-09-11
[Core: Login] Log bad password attempts

johnsaigle:
_Note: It's best to add `?w=1` to the URL when reviewing the code as there are many spacing changes._

----

### Brief summary of changes

Adds a log entry when a user enters the wrong password. I imagine that this was intended to be included by the original author but was left out for some reason. 

Also did happy path refactoring in this function to make it more clear. The important line is [this](https://github.com/aces/Loris/compare/bugfix...johnsaigle:180911-LogBadPasswords?expand=1#diff-2c13360f65cfe03fa685449b9074176eR321).

Finally, changed the front-end error message to say "Invalid username" instead of "Invalid email" and updated the integration tests accordingly.

### To test this change...

- [ ] Keep an eye on the `user_login_history` table.
- [ ] On your branch, try to login with a bad password. Nothing should change
- [ ] Do the same on my branch. A new line should appear stating that a bad password was used.

### DIFF
--- a/php/libraries/SinglePointLogin.class.inc
+++ b/php/libraries/SinglePointLogin.class.inc
@@ -387,15 +387,14 @@ class SinglePointLogin
 
             $DB->insert('user_login_history', $setArray);
             return true;
-        } else {
-            /* When the password is wrong, log this detail for rate-limiting 
-             * purposes.
-             */
-            $this->insertFailedDetail(
-                "Bad password",
-                $setArray
-            );
         }
+        /* When the password is wrong, log this detail for rate-limiting 
+         * purposes.
+         */
+        $this->insertFailedDetail(
+            "Bad password",
+            $setArray
+        );
 
         // Fail by default with a generic error message
         $this->_lastError = "Incorrect username or password.";
--- a/php/libraries/SinglePointLogin.class.inc
+++ b/php/libraries/SinglePointLogin.class.inc
@@ -309,8 +309,8 @@ class SinglePointLogin
         $row   = $DB->pselectRow($query, array('username' => $username));
 
         if (intval($row['User_count'] ?? 0) !== 1) {
-            /* Display generic failure message to front-end so that an 
-             * attacker will have a harder time knowing the difference 
+            /* Display generic failure message to front-end so that an
+             * attacker will have a harder time knowing the difference
              * between a bad username and a bad password.
              */
             $this->_lastError = "Incorrect username or password.";
--- a/php/libraries/SinglePointLogin.class.inc
+++ b/php/libraries/SinglePointLogin.class.inc
@@ -388,7 +388,7 @@ class SinglePointLogin
             $DB->insert('user_login_history', $setArray);
             return true;
         }
-        /* When the password is wrong, log this detail for rate-limiting 
+        /* When the password is wrong, log this detail for rate-limiting
          * purposes.
          */
         $this->insertFailedDetail(
--- a/php/libraries/SinglePointLogin.class.inc
+++ b/php/libraries/SinglePointLogin.class.inc
@@ -316,17 +316,10 @@ class SinglePointLogin
             $this->_lastError = "Incorrect username or password.";
             return false;
         }
-        // validate passsword
+        // Validate passsword
         $oldhash = $row['Password_hash'];
-        if (!password_verify($password, $oldhash)) {
-            // Display generic failure message, as above.
-            $this->_lastError = "Incorrect username or password.";
-            $this->insertFailedDetail(
-                "Bad password",
-                $setArray
-            );
-            return false;
-        } else {
+        if (password_verify($password, $oldhash)) {
+            // Rehash according to latest standards if necessary.
             if (password_needs_rehash($oldhash, PASSWORD_DEFAULT)) {
                 $user = \User::factory($username);
                 $user->updatePassword($password);
@@ -345,7 +338,7 @@ class SinglePointLogin
                 return false;
             }
 
-            // check if the account is active
+            // Check if the account is active.
             if (($row['active_from'] != null)
                 && ($currentDay < strtotime($row['active_from']))
             ) {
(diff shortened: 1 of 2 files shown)


php/libraries/SinglePointLogin.class.inc
```diff
@@ -305,60 +305,70 @@ class SinglePointLogin
...
+                "Bad password",
+                $setArray
+            );
+            return false;
+        }
 
-                    return false;
-                }
+        if (password_needs_rehash($oldhash, PASSWORD_DEFAULT)) {
+            $user =& User::factory($username);
```
paiva: Should it be `\User::` or `User::`?

johnsaigle: It should probably be `\User::` but I just moved that line from elsewhere in the old code.

driusan: If the file's not in a namespace it doesn't really make a difference (but might be a good idea to include the `\` for consistency across files..)

um4r12:
Code LGTM. Passed manual testing as well.
I noticed that the `user_login_history` table contains a `Failcode` column. Should each of those errors (ex. `Bad password`, `user account not active`, `user account pending` etc.) be mapped to a `Failcode`?

johnsaigle:
@um4r12 `Failcode` is never actually used. There's a note somewhere in this class that someone wrote that we should implement the feature at a later date.

It's probably a good idea to do at some point but out of scope for this PR.

johnsaigle:
@driusan This has been rebased to minor.

php/libraries/SinglePointLogin.class.inc
```diff
@@ -308,86 +308,95 @@ class SinglePointLogin
...
         }
 
-        // bad usename or password
-        $this->_lastError = "Incorrect email or password";
-        return false;
+        // user is valid
+        $this->_username = $row['UserID'];
+
+        $DB->insert('user_login_history', $setArray);
+        return true;
```
driusan: This refactoring essentially makes the authentication default to "true" unless a blacklisted set of conditions is met rather than defaulting to "false" unless a whitelisted set of conditions are met. This seems dangerous.

johnsaigle: That's a good point. I changed my approach to fail by default and structure the logic more similarly to how it was previously.

php/libraries/SinglePointLogin.class.inc
```diff
@@ -308,85 +308,95 @@ class SinglePointLogin
...
+            /* Display generic failure message to front-end so that an
+             * attacker will have a harder time knowing the difference
+             * between a bad username and a bad password.
+             */
+            $this->_lastError = "Incorrect username or password.";
+            return false;
+        }
+        // validate passsword
+        $oldhash = $row['Password_hash'];
+        if (!password_verify($password, $oldhash)) {
```
driusan: This negation makes the logic difficult to follow (especially since every example I can find for password_needs_rehash basically says to use
```
if (password_verify) {
    if (password_needs_rehash) {
    ..
  ..
```
)

and it's now much more difficult to tell that that's what this is trying to do.

johnsaigle:
@driusan I addressed your points

php/libraries/SinglePointLogin.class.inc
```diff
@@ -245,7 +245,7 @@ class SinglePointLogin
      *                         If false, it won't provide any output
      *                         in those situation.
      *
-     * @return bool true if the username and password are valid
+     * @return bool False unless account is valid and password correct.
```
zaliqarosli: on line 240, comment for $password should be changed from username to password

php/libraries/SinglePointLogin.class.inc
```diff
@@ -308,85 +308,96 @@ class SinglePointLogin
...
-                        . " this account.";
-                    return false;
-
-                }
+        if (intval($row['User_count'] ?? 0) !== 1) {
+            /* Display generic failure message to front-end so that an
+             * attacker will have a harder time knowing the difference
+             * between a bad username and a bad password.
+             */
+            $this->_lastError = "Incorrect username or password.";
```
zaliqarosli: how come some cases have insertFailedDetail and some don't?

php/libraries/SinglePointLogin.class.inc
```diff
@@ -308,85 +308,96 @@ class SinglePointLogin
...
+        /* When the password is wrong, log this detail for rate-limiting
+         * purposes.
+         */
+        $this->insertFailedDetail(
+            "Bad password",
+            $setArray
+        );
+
+        // Fail by default with a generic error message
+        $this->_lastError = "Incorrect username or password.";
```
zaliqarosli: _lastError should already be set to this value from line 316 no?

php/libraries/SinglePointLogin.class.inc
```diff
@@ -308,85 +308,96 @@ class SinglePointLogin
...
-                    return false;
-
-                }
+        if (intval($row['User_count'] ?? 0) !== 1) {
+            /* Display generic failure message to front-end so that an
+             * attacker will have a harder time knowing the difference
+             * between a bad username and a bad password.
+             */
+            $this->_lastError = "Incorrect username or password.";
+            return false;
```
zaliqarosli: if this case returns false, can't it fail by default at the end of this function?

johnsaigle: It could but it's quicker if it fails out right away instead of going through all these other checks. If someone enters a bad password for a user account it doesn't make sense to go on and check whether that account is expired and tell them that info.

php/libraries/SinglePointLogin.class.inc
johnsaigle: I'm not sure. The below changes are mostly just formatting/whitespace stuff and I didn't write the original code. My goal here was just to do a small bug fix to log bad password attempts.

php/libraries/SinglePointLogin.class.inc
johnsaigle: In theory yes and I removed this in an earlier version of this PR. @driusan had me add it back so we have this set up to fail by default unless the user meets all of the above criteria.

I agree with you that the logic is confusing to read.

php/libraries/SinglePointLogin.class.inc
ridz1208: done

ridz1208:
@kongtiaowang please test again

Issue #3314, opened 2017-11-24
Password requirements don't follow NIST standards


gdevenyi:
The password setting requirements violate best practices as recommended by NIST:
https://pages.nist.gov/800-63-3/sp800-63b.html#appA

https://pages.nist.gov/800-63-3/sp800-63b.html#memsecret

gdevenyi:
> Verifiers SHOULD NOT impose other composition rules (e.g., requiring mixtures of different character types or prohibiting consecutively repeated characters) for memorized secrets. Verifiers SHOULD NOT require memorized secrets to be changed arbitrarily (e.g., periodically). However, verifiers SHALL force a change if there is evidence of compromise of the authenticator.

driusan:
Hi @gdevenyi,

I think it's important to take the NIST guidelines as a whole, and not selectively apply them. They don't simply say "don't have any requirements other than an 8 character limit." Selectively applying the parts of the best practices which contradict and relax previously published best practices is without simultaneously implementing the other parts which mitigate the risks is likely worse than using an outdated password requirement scheme.

For instance, this part of the document you linked to is important:

```
When processing requests to establish and change memorized secrets, verifiers SHALL compare the prospective secrets against a list that contains values known to be commonly-used, expected, or compromised. For example, the list MAY include, but is not limited to:

    Passwords obtained from previous breach corpuses.
    Dictionary words.
    Repetitive or sequential characters (e.g. ‘aaaaaa’, ‘1234abcd’).
    Context-specific words, such as the name of the service, the username, and derivatives thereof.

If the chosen secret is found in the list, the CSP or verifier SHALL advise the subscriber that they need to select a different secret, SHALL provide the reason for rejection, and SHALL require the subscriber to choose a different value.
```

As is this:

```
Verifiers SHALL implement a rate-limiting mechanism that effectively limits the number of failed authentication attempts that can be made on the subscriber’s account as described in Section 5.2.2.
```

That said, as a user I agree that the password requirements implemented by LORIS are dated and need to be modernized, but in order to remain secure I think it's import that it be implemented alongside the other updated best practices, and not simply by removing the password strength requirements currently in place as your ticket seems to be requesting.

driusan:
I went through the requirements and tried to enumerate them in an easier checklist.

The following MUST be done in order to be compliant with the best practices:

- [x] Passwords must be at least 8 characters long
- [x] Randomly generated tokens must be at least 6 characters long
[...]

PR #3950, opened 2018-09-11
[Core: Login] Add rate limiting/account lockout

johnsaigle:
### Brief summary of changes

Make it so that many bad passwords attempts result in a message telling the user they'll need to wait before trying to login again. This helps protect us in the case where an attacker would try many many logins to try to break in to LORIS.

More detailed explanations exist in the comments to the code.

Addresses rate limiting mentioned in #3314 

### To test this change...

**Note** until [[Core: Login] Log login attempts using bad passwords](#3949) is merged, the below won't work because we don't actually log bad passwords attempts right now...

- [x] Check out the branch and try to log in 10 times with a bad password. A message will appear

### DIFF
--- a/php/libraries/SinglePointLogin.class.inc
+++ b/php/libraries/SinglePointLogin.class.inc
@@ -319,7 +319,6 @@ class SinglePointLogin
         }
         /* Check whether a user's account is locked due to too many bad login
          * attempts before actually trying to authenticate their credentials.
-         * Doing so can help mititgate timing attacks against the server.
          */
         if ($this->accountLocked($_POST['username'])) {
             $this->_lastError = 'This account is currently suspended due '
--- a/php/libraries/SinglePointLogin.class.inc
+++ b/php/libraries/SinglePointLogin.class.inc
@@ -325,10 +325,6 @@ class SinglePointLogin
             $this->_lastError = 'This account is currently suspended due '
                 . 'to too many bad login attempts.';
             $this->insertFailedDetail('Account locked', $setArray);
-            error_log(
-                "The acccount for $username is locked out. Potential "
-                . "brute force login attempt"
-            );
             return false;
         }
         // Validate passsword
@@ -473,10 +469,11 @@ class SinglePointLogin
         $longLockoutThreshold  = 15;
 
         // SQL code to run.
-        $query  = "SELECT COUNT(loginhistoryID)" .
+        $query  = "SELECT COUNT(loginhistoryID) " .
             "FROM user_login_history " .
             "WHERE userID = :username " .
             "AND Success = 'N' " .
+            "AND IP_address = :ip " .
             "AND Login_timestamp > ( " .
             "SELECT IF( MAX(Login_timestamp) > DATE_SUB(NOW(), ".
             "INTERVAL :timeWindowInMinutes MINUTE), " .
@@ -490,6 +487,7 @@ class SinglePointLogin
         $params = array(
                    'username'            => $username,
                    'timeWindowInMinutes' => $shortWindowInMinutes,
+                   'ip' => $_SERVER['REMOTE_ADDR'],
                   );
 
         $shortFailCount = $this->_getFailedLoginCount($query, $params);
--- a/php/libraries/SinglePointLogin.class.inc
+++ b/php/libraries/SinglePointLogin.class.inc
@@ -469,11 +469,11 @@ class SinglePointLogin
         $longLockoutThreshold  = 15;
 
         // SQL code to run.
-        $query  = "SELECT COUNT(loginhistoryID) as loginattempts" .
+        $query  = "SELECT COUNT(loginhistoryID) as loginattempts " .
             "FROM user_login_history " .
             "WHERE userID = :username " .
-            "AND Success = 'N' " .
             "AND IP_address = :ip " .
+            "AND Success = 'N' " .
             "AND Login_timestamp > ( " .
             "SELECT IF( MAX(Login_timestamp) > DATE_SUB(NOW(), ".
             "INTERVAL :timeWindowInMinutes MINUTE), " .
@@ -484,6 +484,7 @@ class SinglePointLogin
             "WHERE userID = :username " .
             "AND Success = 'Y'" .
         ");";
+        error_log($_SERVER['REMOTE_ADDR']);
         $params = array(
                    'username'            => $username,
                    'timeWindowInMinutes' => $shortWindowInMinutes,
--- a/php/libraries/SinglePointLogin.class.inc
+++ b/php/libraries/SinglePointLogin.class.inc
@@ -468,7 +468,7 @@ class SinglePointLogin
         $longWindowInMinutes  = 60;
 
         // SQL code to run.
-        $query = "SELECT COUNT(loginhistoryID) as loginattempts " .
+        $query  = "SELECT COUNT(loginhistoryID) as loginattempts " .
             "FROM user_login_history " .
             "WHERE userID = :username " .
             "AND IP_address = :ip " .
--- a/php/libraries/SinglePointLogin.class.inc
+++ b/php/libraries/SinglePointLogin.class.inc
@@ -230,7 +230,8 @@ class SinglePointLogin
         $this->_username
             = isset($decodedArray['user'])
                 ? $decodedArray['user'] : 'Unknown';
-        return isset($decodedArray['user']);
+        return isset($decodedArray['user']) 
+            && !$this->accountLocked($decodedArray['user']);
     }
 
     /**
@@ -316,6 +317,16 @@ class SinglePointLogin
             $this->_lastError = "Incorrect username or password.";
             return false;
         }
+        /* Check whether a user's account is locked due to too many bad login
+         * attempts before actually trying to authenticate their credentials.
+         * Doing so can help mititgate timing attacks against the server.
+         */
+        if ($this->accountLocked($_POST['username'])){
+            $this->_lastError = 'This account is currently suspended due '
+                . 'to too many bad login attempts.';
+            $this->insertFailedDetail('Account locked', $setArray);
+            return false;
+        }
         // Validate passsword
         $oldhash = $row['Password_hash'];
         if (password_verify($password, $oldhash)) {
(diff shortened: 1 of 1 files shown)


johnsaigle:
Adding blocked until #3949 is merged into minor.

php/libraries/SinglePointLogin.class.inc
```diff
@@ -390,6 +404,97 @@ class SinglePointLogin
...
+            );
+            /* If a failed login attempt is made with a timestamp that is more
+             * recent than the time threshold, then it is considered in the
+             * decision to lockout a user. Timestamps older than this are
+             * considered too distant for concern.
+             */
+            if ($attemptTime > $timeThreshold) {
+                $badLoginsInsideTimeWindow++;
+            }
+        }
```
PapillonMcGill: replace these queries and code with:
```
SELECT COUNT(loginhistoryID) 
FROM user_login_history
WHERE userID = :username
AND Success = 'N'
AND Login_timestamp > (
    SELECT IF( MAX(Login_timestamp) > DATE_SUB(NOW(),  INTERVAL :timeWindowInMinutes MINUTE), 
                           MAX(Login_timestamp), DATE_SUB(NOW(), 
                           INTERVAL :timeWindowInMinutes MINUTE) 
                          )
    FROM user_login_history 
    WHERE userID = :username
    AND Success = 'Y'
);
```

johnsaigle:
@PapillonMcGill Thanks! I've made the changes you suggested.

php/libraries/SinglePointLogin.class.inc
```diff
@@ -463,37 +463,29 @@ class SinglePointLogin
     ): int {
         $factory = NDB_Factory::singleton();
         $db      = $factory->database();
-        $query   = "SELECT Login_timestamp from user_login_history " .
+        $query   = "SELECT COUNT(loginhistoryID)" .
```
PapillonMcGill: could use "SELECT COUNT(loginhistoryID) as 'count' ...  to simplify retrieval of data from the array
as well as pselectOne as you will alway get only 1 value

php/libraries/SinglePointLogin.class.inc
```diff
@@ -390,6 +404,89 @@ class SinglePointLogin
...
+            "SELECT IF( MAX(Login_timestamp) > DATE_SUB(NOW(), ".
+            "INTERVAL :timeWindowInMinutes MINUTE), " .
+                "MAX(Login_timestamp), DATE_SUB(NOW(), " .
+                "INTERVAL :timeWindowInMinutes MINUTE) " .
+            ") " .
+            "FROM user_login_history " .
+            "WHERE userID = :username " .
+            "AND Success = 'Y'" .
+        ");";
+        $result  = $db->pselect(
```
xlecours: When you make the same query with different parameters, it would be more efficient to use the same prepared query twice.

johnsaigle: Oops this shouldn't be here.

PapillonMcGill: @xlecours There is only 1 query here, with a nested but different query build-in.
@johnsaigle quey is fine

johnsaigle: @xlecours Could you follow-up here or else dismiss your review?

xlecours: The `countFailedLogins` function is called twice. I think this is exacly why it as been put in a function. You could have prepare the query and then call execute twice with the different parameters. 
```

File: php/libraries/SinglePointLogin.class.inc [hunk 4]
```diff
@@ -316,6 +317,20 @@ class SinglePointLogin
...
         }
+        /* Check whether a user's account is locked due to too many bad login
+         * attempts before actually trying to authenticate their credentials.
+         * Doing so can help mititgate timing attacks against the server.
+         */
+        if ($this->accountLocked($_POST['username'])) {
+            $this->_lastError = 'This account is currently suspended due '
+                . 'to too many bad login attempts.';
+            $this->insertFailedDetail('Account locked', $setArray);
+            error_log(
```
driusan: This isn't an error and doesn't belong in the error log

php/libraries/SinglePointLogin.class.inc
```diff
@@ -401,6 +444,82 @@ class SinglePointLogin
...
+        $longWindowInMinutes  = 60;
+        /* These represent Y and Y' above, but with more helpful names.
+         * These values represent the maximum umber of attempts a user may make
+         * to login before being locked out.
+         */
+        $shortLockoutThreshold = 10;
+        $longLockoutThreshold  = 15;
+
+        // SQL code to run.
+        $query  = "SELECT COUNT(loginhistoryID)" .
```
driusan: Without including IP address in this it introduces a trivial denial of service attack for any account.

php/libraries/SinglePointLogin.class.inc
johnsaigle: I'll remove it. In another context I think we should discuss creating a log for security-related events. Without detailed logging for high-value transactions it'll be difficult to do any kind of incident response in the future.

johnsaigle:
@driusan Can you re-review when you have the time?

php/libraries/SinglePointLogin.class.inc
driusan: This doesn't seem to be addressed

johnsaigle: @driusan @xlecours I updated the code after this comment. The difference can be seen between the commits on November 12 and 13. As I stated below, please let me know whether this version looks better to you than before. I'm not totally sure I understood the problem but I addressed it to my understanding.

php/libraries/SinglePointLogin.class.inc
```diff
@@ -401,6 +440,84 @@ class SinglePointLogin
...
+            "WHERE userID = :username " .
+            "AND Success = 'Y'" .
+        ");";
+        $params = array(
+                   'username'            => $username,
+                   'timeWindowInMinutes' => $shortWindowInMinutes,
+                   'ip'                  => $_SERVER['REMOTE_ADDR'],
+                  );
+
+        $shortFailCount = $this->_getFailedLoginCount($query, $params);
```
xlecours: What I meant was more like:
```
        $database = \NDB_Factory::singleton()->database();
        $statement = $database->prepare($query);
        $shortFailCount = intval($database->execute($statmnent, $params)[0]['COUNT(loginhistoryID)'];
        $params['timeWindowInMinutes'] = $longWindowInMinutes;
        $longFailCount =  intval($database->execute($statmnent, $params)[0]['COUNT(loginhistoryID)'];
```
also, using an alias like 'as loginattempts' in the query would make the result key 'COUNT(loginhistoryID)'  easier to see (subjective comment here)

xlecours:
> @xlecours I changed the function to according to your advice (I think). Let me know if you're happy with it.

Happiness is not triggered here but no big deal.

php/libraries/SinglePointLogin.class.inc
```diff
@@ -401,6 +440,93 @@ class SinglePointLogin
...
+     * time window specified in $params. Used to determine whether a user's
+     * login request should be denied as a security precaution.
+     *
+     * @param string $query  The DB query.
+     * @param array  $params Username and time window.
+     *
+     * @return int The number of failed login attempts that occurred during the
+     *              time window
+     */
+    private function _getFailedLoginCount(
```
xlecours: I don't think this is used anymore.

johnsaigle:
@xlecours Can you update your review?

paiva:
@johnsaigle Ready for review?

php/libraries/SinglePointLogin.class.inc
```diff
@@ -325,6 +336,34 @@ class SinglePointLogin
...
+                    . " reactivate this account.";
+                $this->insertFailedDetail(
+                    "user account not active",
+                    $setArray
+                );
+
+                return false;
+            }
+
+            // check if the account is longer active
```
PapillonMcGill:
```suggestion
            // check if the account is no longer active
```

php/libraries/SinglePointLogin.class.inc
```diff
@@ -401,6 +440,75 @@ class SinglePointLogin
...
+         * legitimate locked-out users to try again in a short time while also
+         * preventing attackers from simply trying another X login attempts
+         * after Y minutes.
+         */
+
+        // These represent X and X' above, but with more helpful names.
+        $shortWindowInMinutes = 15;
+        $longWindowInMinutes  = 60;
+        /* These represent Y and Y' above, but with more helpful names.
+         * These values represent the maximum umber of attempts a user may make
```
PapillonMcGill:
```suggestion
         * These values represent the maximum number of attempts a user may make
```

php/libraries/SinglePointLogin.class.inc
```diff
@@ -401,6 +440,75 @@ class SinglePointLogin
...
+     * @return bool
+     */
+    function accountLocked(string $username): bool
+    {
+        /* A user is locked out if they have made X attempts in the last Y
+         * minutes or X' attempts in the last Y' minutes.
+         * The reason for using two time windows is to provide some flexibility:
+         * legitimate locked-out users to try again in a short time while also
+         * preventing attackers from simply trying another X login attempts
+         * after Y minutes.
```
PapillonMcGill: The variable X, X', Y and Y' in the above text have opposite signification with the text below.
Above, X and X' represent attempts while
below, X and X' represent the time windows

johnsaigle:
@PapillonMcGill Thanks for the review, I updated your changes

php/libraries/SinglePointLogin.class.inc
```diff
@@ -316,6 +317,16 @@ class SinglePointLogin
             $this->_lastError = "Incorrect username or password.";
             return false;
         }
+        /* Check whether a user's account is locked due to too many bad login
+         * attempts before actually trying to authenticate their credentials.
+         * Doing so can help mititgate timing attacks against the server.
```
driusan: The last sentence is wrong. To mitigate timing attacks you need to ensure that all code paths are constant time and there's no branching/early return.

php/libraries/SinglePointLogin.class.inc
```diff
@@ -401,6 +439,75 @@ class SinglePointLogin
...
+            "SELECT IF( MAX(Login_timestamp) > DATE_SUB(NOW(), ".
+            "INTERVAL :timeWindowInMinutes MINUTE), " .
+                "MAX(Login_timestamp), DATE_SUB(NOW(), " .
+                "INTERVAL :timeWindowInMinutes MINUTE) " .
+            ") " .
+            "FROM user_login_history " .
+            "WHERE userID = :username " .
+            "AND Success = 'Y'" .
+        ");";
+        error_log($_SERVER['REMOTE_ADDR']);
```
xlecours: This can be removed

PR #3951, opened 2018-09-12
[Core] Fix getAssociativeSiteList in Utility.class.inc

Jkat:
### Brief summary of changes

Fixing the incorrectly parsed Data Coordinating Center site in `getAssociativeSiteList` in `Utilities.class.inc` for the past 6-7years

### This resolves issue...

- [ ] Redmine? #

- [ ] Github? #

### To test this change...

- [ ] Check that the dashboard does/doesn't include Data Coordinating Center data when function argument set to true/false

### DIFF
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -160,7 +160,7 @@ class Utility
             $query .= "WHERE Study_site='Y'";
         }
         if (!$DCC) {
-            $query .= " AND Name != 'Data Coordinating Center'";
+            $query .= " AND CenterID <> 1";
         }
 
         $result = $DB->pselect($query, array());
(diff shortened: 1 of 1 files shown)


php/libraries/Utility.class.inc
```diff
@@ -160,7 +160,7 @@ class Utility
             $query .= "WHERE Study_site='Y'";
         }
         if (!$DCC) {
-            $query .= " AND Name != 'DCC'";
+            $query .= " AND Name != 'Data Coordinating Center'";
```
driusan: The convention we adopted in LORIS a while ago (but obviously after this was written..) is `CenterID <> 1`

Jkat: Do you know if this 'convention' is documented somewhere

driusan: I'm not sure, it was discussed in a LORIS meeting a year or two ago, so it might be in the minutes. You're right though, it should be better documented somewhere.

PR #3952, opened 2018-09-12
[JSX/Form.js] Adding sortByValue for Selects and SearchableSelects

ridz1208:
## Brief summary of changes
This allows for the possibility to sort the options of a select dropdown (render with react) by the value for both SelectElements and SearchableDropdowns. i.e. you query from the DB and pass to the JS an array of the form `$examinerID =>$examinerName`, in PHP this array looks like 

```
array(
    5=>"Rida", 
    1=>"Dave", 
    4=>"Zal",
    2=>"Henri", 
    3=>"John", 
)
```
the object created in React will automatically sort the values as follows
```
{
   1:"Dave",
   2:"Henri",
   3:"Zal",
   4:"John",
   5:"Rida"
}
``` 
This allows you to pass in a bool `sortByValue` that would render the options sorted by name as follows:
```
{
   1:"Dave",
   2:"Henri",
   4:"John",
   5:"Rida",
   3:"Zal"
}
``` 
***PS:** To all react pros... if you have a better way of doing this, please share your wisdom with us lesser people.*

### To test this change...

- [ ] test a SelectElement and a SearchableDropdown element before making any changes
- [x] add `sortByValue:true` to the element and make sure the options in the dropdown get sorted properly
- [ ] make sure you can save the correct value in both the cases above
- [x] ensure that after saving and refreshing the page, the value is still correct
- [ ] try it in multiplle places in the code

# If this sentence shows up it means I didn't read the instructions!
YOU BET I DIDN'T

### DIFF
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -195,7 +195,7 @@ class SearchableDropdown extends Component {
   render() {
     let required = this.props.required ? 'required' : null;
     let disabled = this.props.disabled ? 'disabled' : null;
-    let sortByValue = this.props.sortByValue ? 'sortByValue' : null;
+    let sortByValue = this.props.sortByValue;
     let options = this.props.options;
     let strictMessage = 'Entry must be included in provided list of options.';
     let errorMessage = null;
@@ -233,8 +233,8 @@ class SearchableDropdown extends Component {
       }
     }
 
-    let newOptions ={};
-    let optionList =[];
+    let newOptions = {};
+    let optionList = [];
     if (sortByValue) {
       for (let key in options) {
         if (options.hasOwnProperty(key)) {
@@ -314,7 +314,7 @@ SearchableDropdown.defaultProps = {
   class: '',
   disabled: false,
   required: false,
-  sortByValue: false,
+  sortByValue: true,
   errorMessage: '',
   placeHolder: '',
   onUserInput: function() {
@@ -354,7 +354,7 @@ class SelectElement extends Component {
     let multiple = this.props.multiple ? 'multiple' : null;
     let required = this.props.required ? 'required' : null;
     let disabled = this.props.disabled ? 'disabled' : null;
-    let sortByValue = this.props.sortByValue ? 'sortByValue' : null;
+    let sortByValue = this.props.sortByValue;
     let options = this.props.options;
     let errorMessage = null;
     let emptyOptionHTML = null;
@@ -377,8 +377,8 @@ class SelectElement extends Component {
       elementClass = 'row form-group has-error';
     }
 
-    let newOptions ={};
-    let optionList =[];
+    let newOptions = {};
+    let optionList = [];
     if (sortByValue) {
       for (let key in options) {
         if (options.hasOwnProperty(key)) {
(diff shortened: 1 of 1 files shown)


kongtiaowang:
Using this "npm run lintfix:javascript"  to fix the error.

jsx/Form.js
```diff
@@ -416,6 +456,7 @@ SelectElement.defaultProps = {
   multiple: false,
   disabled: false,
   required: false,
+  sortByValue: false,
```
HenriRabalais: I'm thinking it may actually be useful to have this be default behaviour, thoughts?

ridz1208:
@davidblader @HenriRabalais better ?

jsx/Form.js
```diff
@@ -228,6 +229,27 @@ class SearchableDropdown extends React.Component {
...
 
+    let newOptions = {};
+    let optionList = [];
+    if (sortByValue) {
+      for (let key in options) {
+        if (options.hasOwnProperty(key)) {
+          newOptions[options[key]] = key;
+        }
+      }
+      optionList = Object.keys(newOptions).sort().map(function(option) {
```
davidblader: this looks good. my only concern is that putting this in the render function will become costly performance-wise, but since Form elements are stateless idk if putting it in a componentDidMount method would feasible / not sure about consequences of adding state to a Form component
wondering what @HenriRabalais thinks

ridz1208:
@HenriRabalais @davidblader any approvals ?

driusan:
Looks fine to me, but can you send it to major instead of minor since it (potentially) affects every single form in LORIS?

ridz1208:
@driusan you'r the one who suggested to make it the default 

> seems reasonable.. I think in most cases people expect it to be sorted and you can always manually turn it off in cases where you don't want that.

I'd rather remove it as a default and merge it in minor. I need a faster release on this one !! then I'll issue another PR to change the default into major.

deal ?

driusan:
@ridz1208 That was in response to the default value, not the branch. The fact that Form.js is changing means that it impacts any module that uses form.js regardless of the value and should theoretically be tested to release this (at least anything that gets regenerated when doing "npm run compile".

It's probably relatively minor and I'm willing to take a chance, but according to our policies of what goes into minor/major it should be going to major.

ridz1208:
yeah... okai

driusan:
@ridz1208 conflicts

ridz1208:
@zaliqarosli @HenriRabalais @davidblader I rebased this to major, please re-test and re-review

zaliqarosli:
@ridz1208 all React elements on major are broken right now .......

ridz1208:
@zaliqarosli did you npm runcompile or `make`

zaliqarosli:
@ridz1208 not sure what's wrong with my VM but this branch does not work at all for me :(

jsx/Form.js
```diff
@@ -195,6 +195,7 @@ class SearchableDropdown extends Component {
   render() {
     let required = this.props.required ? 'required' : null;
     let disabled = this.props.disabled ? 'disabled' : null;
+    let sortByValue = this.props.sortByValue;
```
zaliqarosli: surely all these changes in Form.js should be accompanied by changes made to compiled js files?

ridz1208:
No need since the make command does it

PR #3953, opened 2018-09-17
[Examiner] Fix warnings caused by absent variable

johnsaigle:
### Brief summary of changes
Warnings were created when `$certificationInstruments['test']` was not set. I'm not sure what this does exactly but this code change handles the case when it's not there. 

Also fixed a typo ~~and added a return type to the function according to its PHP doc.~~

### To test this change...

- [ ] Go to the edit section of the examiner module. Notice the warnings! On my branch there are no warnings.

### DIFF
--- a/modules/examiner/php/editexaminer.class.inc
+++ b/modules/examiner/php/editexaminer.class.inc
@@ -402,11 +402,11 @@ class EditExaminer extends \NDB_Form
 
     /**
      * Grabs the certification instruments from the config and creates an array.
-     * The instrument id id the key, and the instrument label is the value.
+     * The instrument id is the key, and the instrument label is the value.
      *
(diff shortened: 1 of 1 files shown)


modules/examiner/php/editexaminer.class.inc
```diff
@@ -402,11 +402,11 @@ class EditExaminer extends \NDB_Form
 
     /**
      * Grabs the certification instruments from the config and creates an array.
-     * The instrument id id the key, and the instrument label is the value.
+     * The instrument id is the key, and the instrument label is the value.
      *
      * @return array of instruments requiring certification
      */
-    function getCertificationInstruments()
+    function getCertificationInstruments(): array
```
driusan: The signature shouldn't change on the bugfix branch

PR #3954, opened 2018-09-17
[Meta] tabs r ew in NDB_Client.class.inc

Jkat:
### Brief summary of changes

Noticed someone likes tabs over spaces

### To test this change...

- [ ] Be OCD **and** agree that tabs are ew


driusan:
:(

PR #3955, opened 2018-09-17
[JSX] StaticDataTable - Replaced loris.hiddenHeaders with this.props.hiddenHeaders 

HenriRabalais:
### Brief summary of changes
The `loris.hiddenHeaders` variable has been replaced by `this.props.hiddenHeaders`. This is to allow the `formatColumns` function in reactified modules to live in the main index file for that module. States should be passed down as props, and not saved as global variables; hence the change.

The merging of this PR will allow those that reference this PR to be unblocked.


driusan:
Can you move this to major? It's an API change on the javascript side that's going to break any references to the old value.

HenriRabalais:
PR was moved to major #3967

PR #3958, opened 2018-09-17
[JSX] DataDictionary - Refactor and Clean-Up

HenriRabalais:
### Summary of changes
This PR refactors the DataDictionary jsx file to meet ES6 conventions and include the formatColumn function, rather than have it in a separate file.

### Testing
- Go to the DataDictionary Filter and Datatable and ensure that everything is functioning.


driusan:
That last commit title seems a little presumptuous.

HenriRabalais:
@davidblader @kongtiaowang @zaliqarosli @driusan please rereview/test!

modules/datadict/jsx/dataDictIndex.js
```diff
@@ -48,7 +48,7 @@ class DataDictIndex extends React.Component {
       success: data => {
         // FIXME: Remove the following line of code as soon as hiddenHEaders is
         // accepted as a prop by the StaticDataTable Component.
-        loris.hiddenHeaders = data.hiddenHeaders || [];
+        loris.hiddenHeaders = this.state.hiddenHeaders;
```
zaliqarosli: using data.hiddenHeaders here is fine because data from php has the hiddenHeaders data in it. we can remove this.state.hiddenHeaders though

zaliqarosli: (we should probs decide though where we want to specify hiddenheaders for consistency (i choose php if it already exists because some modules have conditions on php side on what the hiddenheaders could be)

HenriRabalais: In this case the php does not send this information, unless I'm missing something? There's no mention of hidden headers in the datadict php. Something is weird about this though, since in the previous version of this file on line `20` 'DescriptionStatus' is set as a hidden header, yet there is a selection filter for it? There doesn't even seem to be permissions associated with that field, so I'm going to go ahead and guess that's it's not supposed to be a hidden field at all.

HenriRabalais: Maybe @llevitis can shed some light here.

zaliqarosli:
testing still works after recent changes!

PR #3961, opened 2018-09-18
[Core] Add pwned passwords API to prevent common passwords and password reuse

johnsaigle:
### Brief summary of changes

Add the [pwned passwords API](https://haveibeenpwned.com/API/v2#PwnedPasswords) into the LORIS password logic. This is part of a refactoring effort to modernize the way LORIS does passwords.

In particular, this API prevents users from creating passwords that are known to be present in previous data leaks. After a password breach, hackers compose large lists of the password and try them in brute force attacks against the sites. 

After this PR users will be prevented from creating a password like `Spring2016!` which is a bad password even though it passes the (current) LORIS password strength requirements.

### Related to

#3314, #3949, #3950 

### To test this change...

- [ ] Checkout my branch.
- [ ] Run the new patch on your DB: `SQL/New_patches/2019-01-28_Add_Pwned_Password_ConfigSetting.sql`
- [ ] On your VM go to the Configuration menu and enable the "Use Pwned Password" check
- [ ] Try changing your password to `Spring2016!`. You should get an error message. 

As @zaliqarosli pointed out, curl is required so if you get an error do 
`sudo apt-get install php7.2-curl; sudo service apache2 restart`

### DIFF
--- a/SQL/New_patches/2019-01-28_Add_Pwned_Password_ConfigSetting.sql
+++ b/SQL/New_patches/2019-01-28_Add_Pwned_Password_ConfigSetting.sql
@@ -1,0 +1,31 @@
+-- Adds the option to toggle the usage of the Pwned Passwords
+-- API (https://haveibeenpwned.com/API/v2#PwnedPasswords) in a project. This
+-- should be enabled by default to allow for a higher level of security. This
+-- setting is added to allow projects to disable the API check in case of
+-- networking issues.
+INSERT INTO ConfigSettings
+  (
+    Name,
+    Description,
+    Visible,
+    AllowMultiple,
+    DataType,
+    Parent,
+    Label,
+    OrderNumber
+  )
+  SELECT
+    'usePwnedPasswordsAPI',
+    'Whether to query the Have I Been Pwned password API on password changes to prevent the usage of common and breached passwords',
+    1,
+    0,
+    'boolean',
+    ID,
+    'Enable "Pwned Password" check',
+    22
+  FROM
+    ConfigSettings
+  WHERE
+    Name="study";
+
+INSERT INTO Config (ConfigID, Value) SELECT ID, 'true' FROM ConfigSettings WHERE Name="usePwnedPasswordsAPI";
--- a/SQL/New_patches/2019-01-28_Add_Pwned_Password_ConfigSetting.sql
+++ b/SQL/New_patches/2019-01-28_Add_Pwned_Password_ConfigSetting.sql
@@ -1,6 +1,6 @@
 -- Adds the option to toggle the usage of the Pwned Passwords
 -- API (https://haveibeenpwned.com/API/v2#PwnedPasswords) in a project. This
--- is be enabled by default to allow for a higher level of security. This
+-- is enabled by default to allow for a higher level of security. This
 -- setting is added to allow projects to disable the API check in case of
 -- networking issues.
 INSERT INTO ConfigSettings
--- a/SQL/New_patches/2019-01-28_Add_Pwned_Password_ConfigSetting.sql
+++ b/SQL/New_patches/2019-01-28_Add_Pwned_Password_ConfigSetting.sql
@@ -1,6 +1,6 @@
 -- Adds the option to toggle the usage of the Pwned Passwords
 -- API (https://haveibeenpwned.com/API/v2#PwnedPasswords) in a project. This
--- should be enabled by default to allow for a higher level of security. This
+-- is be enabled by default to allow for a higher level of security. This
 -- setting is added to allow projects to disable the API check in case of
 -- networking issues.
 INSERT INTO ConfigSettings
--- a/modules/user_accounts/php/my_preferences.class.inc
+++ b/modules/user_accounts/php/my_preferences.class.inc
@@ -444,7 +444,6 @@ class My_Preferences extends \NDB_Form
             if (\User::pwnedPassword($values['Password_hash'])) {
                 $errors['Password_hash'] = 'The password you chose is too '
                     . 'common. Please choose a more unique password.';
-                return $errors;
             }
         }
 
--- a/modules/user_accounts/php/my_preferences.class.inc
+++ b/modules/user_accounts/php/my_preferences.class.inc
@@ -417,11 +417,6 @@ class My_Preferences extends \NDB_Form
         // if password is user-defined, and user wants to change password
         if (!empty($values['Password_hash'])) {
 
-            if (\User::pwnedPassword($values['Password_hash'])) {
-                $errors['Password_hash'] = 'The password you chose is too common. Please '
-                    . 'choose a more unique password.';
-                return $errors;
-            }
             // check password strength
             $isPasswordStrong = \User::isPasswordStrong(
                 $values['Password_hash'],
@@ -447,8 +442,8 @@ class My_Preferences extends \NDB_Form
                 }
             }
             if (\User::pwnedPassword($values['Password_hash'])) {
-                $errors['Password_hash'] = 'The password you chose is too common. Please '
-                    . 'choose a more unique password.';
+                $errors['Password_hash'] = 'The password you chose is too '
+                    . 'common. Please choose a more unique password.';
                 return $errors;
             }
         }
--- a/modules/user_accounts/php/my_preferences.class.inc
+++ b/modules/user_accounts/php/my_preferences.class.inc
@@ -416,6 +416,12 @@ class My_Preferences extends \NDB_Form
 
         // if password is user-defined, and user wants to change password
         if (!empty($values['Password_hash'])) {
+
+            if (\User::pwnedPassword($values['Password_hash'])) {
+                $errors['Password_hash'] = 'The password you chose is too common. Please '
+                    . 'choose a more unique password.';
+                return $errors;
+            }
             // check password strength
             $isPasswordStrong = \User::isPasswordStrong(
                 $values['Password_hash'],
@@ -439,6 +445,11 @@ class My_Preferences extends \NDB_Form
                     $errors['Password_hash'] = 'New and old passwords '
                         . 'are identical: please choose another one';
                 }
+            }
+            if (\User::pwnedPassword($values['Password_hash'])) {
+                $errors['Password_hash'] = 'The password you chose is too common. Please '
+                    . 'choose a more unique password.';
+                return $errors;
             }
         }
 
--- a/php/libraries/User.class.inc
+++ b/php/libraries/User.class.inc
@@ -531,6 +531,47 @@ class User extends UserPermissions
         $this->update($updateArray);
     }
 
+    function pwnedPassword(string $password): bool
+    {
+        // TODO Use the HttpClient in LORIS core to cURL more simply.
+        
+        /* The pwned passwords API searches a range of password hashes to 
+         * protect the actual SHA1 hash from being revealed.
+         * More here: <https://haveibeenpwned.com/API/v2#PwnedPasswords>
+         */
+        $hash = sha1($password);
+        $hashPrefix = substr(sha1($password), 0, 5);
+        $url = "https://api.pwnedpasswords.com/range/$hashPrefix";
+
+        /* Build curl */
+        $ch = curl_init();
+        curl_setopt($ch, CURLOPT_URL, $url);
+
+        // get response
+        curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+
+        $response = curl_exec($ch);
+        curl_close($ch);
+
+        $hashSuffixes = explode("\n", $response);
+        /* The response will be one long string consisting of suffixes matching
+         * our hash prefix, followed by a colon and the number of times the 
+         * password has occurred in data breaches.
+         * Here we iterate over all of the suffixes and check whether the prefix
+         * and the whole suffix line together contain $hash as a substring at 
+         * the beginning of the line (instead of preprocessing by removing
+         * the colon and count from the end of the $suffix line.
+         * Check the above link if you want more info.
+         */
+        foreach ($hashSuffixes as $suffix) {
+            $line = strtolower($hashPrefix . $suffix);
+            if ($hash == substr($line, 0, strlen($hash))) {
+                return true;
+            }
+        }
+        return false;
+    }
+
     /**
      * Determines if the user has a center
      *
--- a/php/libraries/User.class.inc
+++ b/php/libraries/User.class.inc
@@ -546,6 +546,10 @@ class User extends UserPermissions
         /* The pwned passwords API searches a range of password hashes to
          * protect the actual SHA1 hash from being revealed.
          * More here: <https://haveibeenpwned.com/API/v2#PwnedPasswords>
+         *
+         * NOTE your webserver will need to be configured to allow going 
+         * connections to the link below. If you encounter errors please
+         * verify your firewall settings.
          */
         $hash       = sha1($password);
         $hashPrefix = substr(sha1($password), 0, 5);
@@ -559,6 +563,15 @@ class User extends UserPermissions
         curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
 
         $response = curl_exec($ch);
+        // In case request fials for any reason
+        if (curl_error($ch)) {
+            error_log("ERROR: Problem connecting to pwned passwords API: "
+                . curl_error($ch));
+            /* Don't prevent password change if there is a cURL error. Using
+             * this API is nice but not critical to functioning.
+             */
+            return false;
+        }
         curl_close($ch);
 
         $hashSuffixes = explode("\n", $response);
--- a/php/libraries/User.class.inc
+++ b/php/libraries/User.class.inc
@@ -566,11 +566,10 @@ class User extends UserPermissions
         $hashPrefix = substr(sha1($password), 0, 5);
         // $url        = "https://api.pwnedpasswords.com/range/$hashPrefix";
 
-        $client   = Client::createWithConfig(
-            array('base_uri' => 'https://api.pwnedpasswords.com')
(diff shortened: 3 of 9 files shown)


php/libraries/User.class.inc
```diff
@@ -525,6 +525,55 @@ class User extends UserPermissions
...
+        /* The response will be one long string consisting of suffixes matching
+         * our hash prefix, followed by a colon and the number of times the
+         * password has occurred in data breaches.
+         * Here we iterate over all of the suffixes and check whether the prefix
+         * and the whole suffix line together contain $hash as a substring at
+         * the beginning of the line (instead of preprocessing by removing
+         * the colon and count from the end of the $suffix line.
+         * Check the above link if you want more info.
+         */
+        foreach ($hashSuffixes as $suffix) {
```
PapillonMcGill: What about a password like 'vDg62fF6$gN3&dsKe!dfw' that might have been pwn only once? If we remove all PWD that had been pwn, we will end up blocking a lot of them and eventually it will be difficult to find one that have not been pwn.
A quick extrapolation with 3 test would eliminate about 189M passwords, this would just increased in size over time.

Maybe combine the number of pwn and strength together

johnsaigle: I think the example you gave is a bit of an edge case, and someone with that password is likely to be using a password generator to create it. It's easy to click the button again and generate a new string of random characters.

Even eliminating 189M doesn't put a dent in the entropy of possible passwords. The end goal is to allow any password that is greater than 8 characters that consists of any printable ASCII characters and isn't silly. There's plenty of room to make a good password with that criteria.

php/libraries/User.class.inc
```diff
@@ -525,6 +525,55 @@ class User extends UserPermissions
...
+    {
+        // TODO Use the HttpClient in LORIS core to cURL more simply.
+
+        /* The pwned passwords API searches a range of password hashes to
+         * protect the actual SHA1 hash from being revealed.
+         * More here: <https://haveibeenpwned.com/API/v2#PwnedPasswords>
+         */
+        $hash       = sha1($password);
+        $hashPrefix = substr(sha1($password), 0, 5);
+        $url        = "https://api.pwnedpasswords.com/range/$hashPrefix";
```
PapillonMcGill: Need to add a note in the documentation that need access to outside site. Lots of network admin will prevent outside access from web server to prevent contact to a C&C.

Need to manage timeout in case network admin refuse to open firewall or site is down.
Caveat, some firewall require IP address and do not support FQDN.

PapillonMcGill: Sorry for double post to add in the comment, GitHub is quite slow to update a page with latest comment this morning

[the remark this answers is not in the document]
johnsaigle: I'll add a comment

johnsaigle:
@PapillonMcGill I added code comments to address the firewall case. I also added error handling in the code so that password changes can continue even if the API is not reachable.

PapillonMcGill:
@johnsaigle For the note in the documentation, I was thinking more in the Loris install notes (README.md).

Apart that, the code is fine

johnsaigle:
I'm trying to find an appropriate place in the documentation but I'm not sure where it should go. There doesn't seem to be a Wiki page really suited for it and it feels too detailed for the main README.md.

Do you have any thoughts on this?

PapillonMcGill:
in https://github.com/aces/Loris/wiki/Setup#installation-setup-and-configuration, 10. Opening firewall to ... and create a page for it.

As the wiki is not tied to any PR or release, I will approve this PR.

zaliqarosli:
from admin account, i did not get an error message when changing a user's password to `Spring2016!`. from user account, I get a 500 internal server error when attempting to save a new password

johnsaigle:
@zaliqarosli I wasn't able to reproduce those errors but I did realize that I hadn't added the pwned check to edit_user as I was testing in my_preferences.

Would you mind testing again? If it still doesn't work, could you show me your error logs?

zaliqarosli:
`PHP Fatal error:  Uncaught Error: Call to undefined function curl_init() in /var/www/loris/php/libraries/User.class.inc:553\nStack trace:\n#0 /var/www/loris/modules/user_accounts/php/edit_user.class.inc(1033): User::pwnedPassword('Spring2016!')\n#1 /var/www/loris/php/libraries/LorisForm.class.inc(1322): LORIS\\user_accounts\\Edit_User->_validateEditUser(Array)\n#2 /var/www/loris/php/libraries/NDB_Form.class.inc(132): LorisForm->validate()\n#3 /var/www/loris/php/libraries/NDB_Form.class.inc(190): NDB_Form->save()\n#4 /var/www/loris/modules/user_accounts/php/edit_user.class.inc(562): NDB_Form->handle(Object(Zend\\Diactoros\\ServerRequest))\n#5 /var/www/loris/src/Middleware/UserPageDecorationMiddleware.php(171): LORIS\\user_accounts\\Edit_User->handle(Object(Zend\\Diactoros\\ServerRequest))\n#6 /var/www/loris/src/Middleware/PageDecorationMiddleware.php(49): LORIS\\Middleware\\UserPageDecorationMiddleware->process(Object(Zend\\Diactoros\\ServerRequest), Object(LORIS\\user_accounts\\Edit_User))\n#7 /var/www/loris/php/libraries/NDB_Page.class.inc(650): LO in /var/www/loris/php/libraries/User.class.inc on line 553, referer: http://zrosli-dev.loris.ca/user_accounts/edit_user/admin
`

johnsaigle:
thanks @zaliqarosli! Curl is required. I added instructions to the description

johnsaigle:
@PapillonMcGill I think your approval got removed by my last couple commits. When this passes Travis, could you review quickly and approve again?

christinerogers:
Given `sudo apt-get install` - what is required for CentOS installs?

johnsaigle:
@christinerogers I think the equivalent would be using `yum` instead of `apt` but I don't have a CentOS server to test on. Also this dependency may be handled by `composer` in the future if we move to PSR18 as Dave mentioned.

johnsaigle:
Adding blocked because this probably won't be merged until we use a PSR18 implementation

PapillonMcGill:
@johnsaigle @christinerogers 
`yum install <package-name>`  will work for CentOS

nicolasbrossard:
@johnsaigle I like the idea of checking if a password was pwned. I do notice though that the logic for checking if a password change is legit (namely checking if the new password is "strong", different from the old password and not pwned) is duplicated in a couple of classes. We could encapsulate the algorithm in the `User` class and have a method like `isPasswordChangeValid(old, new)` that would execute the above mentioned checks. Maybe even have a `Password` class that will contain everything password-related? Up to you to see what is more appropriate.

johnsaigle:
@nicolasbrossard I agree with you that there is unnecessary duplication. I would be in favour of merging this PR and later on I could write a new PR that does exactly what you've said. Since this PR has been open for 4 months already, I am hesitant to make it more complex at this stage.

php/libraries/User.class.inc
```diff
@@ -525,6 +527,63 @@ class User extends UserPermissions
...
+        $hash       = sha1($password);
+        $hashPrefix = substr(sha1($password), 0, 5);
+        // $url        = "https://api.pwnedpasswords.com/range/$hashPrefix";
+
+        $adapter  = GuzzleAdapter::createWithConfig(
+            array('base_uri' => 'https://api.pwnedpasswords.com')
+        );
+        $request  = new Request('GET', "/range/$hashPrefix");
+        $response = $adapter->sendRequest($request);
+        print_r($response);
```
driusan: Leftover debug statement
```suggestion
```

php/libraries/User.class.inc
```diff
@@ -525,6 +527,63 @@ class User extends UserPermissions
...
+     * Use the Pwned Passwords API to determine whether a user is trying
+     * to use a password that has been revealed in public data breaches.
+     *
+     * @param string $password The password to check for pwnage.
+     *
+     * @return bool Whether the password is known to be pwned.
+     */
+    function pwnedPassword(string $password): bool
+    {
+        // TODO Use the HttpClient in LORIS core to cURL more simply.
```
driusan: I don't think this comment is valid anymore.

driusan:
I agree with @johnsaigle, we can make an issue about that instead of blocking this PR over it.

johnsaigle:
@driusan I prefer commenting out code with `#` because it's one less character and it won't get confused with real `/*` or `//` comments. :laughing:

nicolasbrossard:
@driusan @johnsaigle fine by me to do the modifications I requested later on. PR LGTM then.

johnsaigle:
@driusan @nicolasbrossard Could you give an approving review if satisfied with the code?

zaliqarosli:
tested user accounts as described in the PR description and it works! did not test single point login however. not sure if i should pass manual testing

johnsaigle:
@zaliqarosli An easy way to test the SinglePointLogin approach is to expire your current password in the database by setting `Password_expiry` to some date in the past. Thanks for testing!

zaliqarosli:
update: single point login allows pwned passwords to pass. this needs fixing

johnsaigle:
@zaliqarosli thanks for testing! Should be fixed now

PapillonMcGill:
LGTM after Travis validation.

zaliqarosli:
@johnsaigle not sure how to test the single point login if the password expiry case is covered my the login module. let me know!

johnsaigle:
@zaliqarosli I'm not actually sure how to trigger that case. I think a lot of this is left over from the old way we used to do logins and may not actually be reachable code... I think it should be tested and refactored as Nicolas said above. 

Since the pwnedPassword check does not break password changes either way, I think it's OK to merge this if your other tests worked well.

zaliqarosli:
tested all cases but single point login and they all work. passing manual test assuming that single point login is deprecated

php/libraries/User.class.inc
```diff
@@ -525,6 +527,60 @@ class User extends UserPermissions
...
+
+        /* We don't want password changes to fail in case there is a
+         * networking issue while making the request. If the API can't be
+         * reached for any reason, we will skip the pwned password check.
+         */
+        if (!$response->getStatusCode() === 200) {
+            return false;
+        }
+        $hashSuffixes = explode("\n", $response->getBody());
+        /* The response will be one long string consisting of suffixes matching
```
driusan:
```suggestion
        /*
         * The response will be one long string consisting of suffixes matching
```

php/libraries/User.class.inc
```diff
@@ -525,6 +527,60 @@ class User extends UserPermissions
...
+        $hashPrefix = substr(sha1($password), 0, 5);
+        // $url        = "https://api.pwnedpasswords.com/range/$hashPrefix";
+
+        $adapter  = GuzzleAdapter::createWithConfig(
+            array('base_uri' => 'https://api.pwnedpasswords.com')
+        );
+        $request  = new Request('GET', "/range/$hashPrefix");
+        $response = $adapter->sendRequest($request);
+
+        /* We don't want password changes to fail in case there is a
```
driusan:
```suggestion
        /*
         * We don't want password changes to fail in case there is a
```

php/libraries/User.class.inc
```diff
@@ -525,6 +527,60 @@ class User extends UserPermissions
...
+     * Use the Pwned Passwords API to determine whether a user is trying
+     * to use a password that has been revealed in public data breaches.
+     *
+     * @param string $password The password to check for pwnage.
+     *
+     * @return bool Whether the password is known to be pwned.
+     */
+    function pwnedPassword(string $password): bool
+    {
+        /* The pwned passwords API searches a range of password hashes to
```
driusan:
```suggestion
        /*
         * The pwned passwords API searches a range of password hashes to
```

php/libraries/User.class.inc
```diff
@@ -525,6 +527,60 @@ class User extends UserPermissions
...
 
+    /**
+     * Use the Pwned Passwords API to determine whether a user is trying
+     * to use a password that has been revealed in public data breaches.
+     *
+     * @param string $password The password to check for pwnage.
+     *
+     * @return bool Whether the password is known to be pwned.
+     */
+    function pwnedPassword(string $password): bool
```
driusan: I think this should start with a config option check to disable this so that LORIS isn't unuseable at places that either don't want or can't allow access to external sites.

modules/user_accounts/php/my_preferences.class.inc
```diff
@@ -429,6 +429,7 @@ class My_Preferences extends \NDB_Form
 
         // if password is user-defined, and user wants to change password
         if (!empty($values['Password_hash'])) {
+
```
driusan:
```suggestion
```

php/libraries/User.class.inc
johnsaigle: Line 562 makes sure that the password change will continue if for some reason the pwnedPassword API is unreachable, including networking configs or outage of the site. 

Given that LORIS should still be usable with this check, would you still prefer an explicit config option? If so, should that be something in `config.xml` or?

php/libraries/User.class.inc
johnsaigle: Is it worth adding a PHPCS rule for this? I didn't know this was a style thing we had.

php/libraries/User.class.inc
driusan: Without an explicit config check, it requires an attempt to contact the server to error out before it response to the user's request (which could take on the order of tens of seconds depending on the reason for the failure), even with that fallback. It's also possible that some users just won't want to use the external service, even though it's available.

I think it should go in the config tables, since we're trying to phase out config.xml

johnsaigle:
@driusan Can you update your review?

php/libraries/User.class.inc
```diff
@@ -11,6 +11,8 @@
  * @link     https://www.github.com/aces/Loris-Trunk/
  */
 
+use Http\Adapter\Guzzle6\Client as GuzzleAdapter;
+use GuzzleHttp\Psr7\Request;
```
driusan: This should be using PSR-18, not a specific hardcoded implementation of PSR-18.

(To do that you'll probably need to wrap the reference to Guzzle in a factory somewhere.)

SQL/New_patches/2019-01-28_Add_Pwned_Password_ConfigSetting.sql
```diff
@@ -1,6 +1,6 @@
 -- Adds the option to toggle the usage of the Pwned Passwords
 -- API (https://haveibeenpwned.com/API/v2#PwnedPasswords) in a project. This
--- should be enabled by default to allow for a higher level of security. This
+-- is be enabled by default to allow for a higher level of security. This
```
zaliqarosli: "be" to still be removed

php/libraries/User.class.inc
johnsaigle: @driusan I just sent a commit that I think does what you asked. Let me know if there's something to correct.

php/libraries/User.class.inc
```diff
@@ -531,6 +533,70 @@ class User extends UserPermissions
...
+    function pwnedPassword(string $password): bool
+    {
+
+        // If the project is configured to disable the Pwned Password check,
+        // always return false (meanning that the password has not been pwned).
+        if ('true' !== \NDB_Factory::singleton()->config()->getSetting(
+            'usePwnedPasswordsAPI'
+        )
+        ) {
+            error_log('Setting disabled!');
```
driusan: left over debug statement

src/Http/Client.php
```diff
@@ -0,0 +1,43 @@
...
+     * a wrapper for Guzzle at this time.
+     */
+    private function __construct() {
+    }
+
+    /**
+     * Guzzle's Client class is final and so cannot be extended directly. This
+     * function acts as a wrapper for this function and serves as a workaround.
+     */
+    public static function createWithConfig(array $config): GuzzleClient {
```
driusan: If a named constructor is required, this should still be returning a Psr\Http\Client\ClientInterface, not a specific implementation of one.

The "config" array is also undocumented here. It seems to only ever have one option passed, so I think it should be `public static function createWithBaseURI(string) : ClientInterface`

This whole class is also acting as a ClientFactory, not a Client.

johnsaigle:
Blocked by #4387

kchatpar:
Password reset emails are not sent to users because #4362 has not been merged to major yet.

johnsaigle:
@kchatpar My bad, my testing instructions weren't clear. Please run the SQL patch `SQL/New_patches/2019-01-28_Add_Pwned_Password_ConfigSetting.sql` in this PR and try again. I'll update the instructions in the PR description.

kchatpar:
PR passes the manual test with the password Spring2016! The password spring2016! does not trigger the error message. 

Consider, making a new password rule where one letter in the password must be capitalized.

johnsaigle:
Sorry @kchatpar it should do the opposite and I forgot one more testing step. You'll need to go to Admin > Configuration and set "Use Pwned Passwords" to True. 

We have some password rules but they're a little old-fashioned. I have another open PR to update those.

I'll update the testing instructions again.

kchatpar:
I set the "Use Pwned Passwords" to True and I still get the same result - Spring2016! is rejected but spring2016! is accepted.

johnsaigle:
@kchatpar That's OK for now. This PR addresses previously leaked passwords, not password complexity. That will be handled in #3962.

Thanks for testing!

johnsaigle:
Thanks again @zaliqarosli. 

For any new reviewers, password changes can be triggered via edit_user, my_preferences, or via an logging in with an expired password. There is another case that appears in `php/libraries/SinglePointLogin.class.inc` but we're not sure how to trigger that code path. However in all other circumstances the pwned password check works.

This last case will likely be cleaned up as login/password refactoring changes in PRs like #4395 and #4411.

PR #3962, opened 2018-09-18
[Core] Modernize password strength requirements

johnsaigle:
### Brief summary of changes

Use Zxcvbn library to estimate password strength instead of relying on
ad-hoc approach.
Move password validation logic into `User` class to reduce code
duplication.
Refactor validation code into simpler functions.

This differs from #3961 as it allows us to do away with the special symbols required in LORIS passwords. The [zxcvbn library](https://github.com/dropbox/zxcvbn) represents a more modern take on password strength and is designed and used by Dropbox. Demo here: https://github.com/dropbox/zxcvbn

### This is related to 

- [x] Github #3314 
- [x] Github #3961 

### To test this change...

- [x] Checkout the branch and try changing your password to `hunter2` and then to `hunter22`. You will not be able to and you will get different error messages for both.


php/libraries/SinglePointLogin.class.inc
```diff
@@ -99,42 +99,18 @@ class SinglePointLogin
...
-        )
-        ) {
-            $this->_lastError = 'The password is weak, or'
-               . ' the passwords do not match';
-        }
-
-        if (password_verify($_POST['password'], $data['Password_hash'])) {
-            $this->_lastError = 'You cannot keep the same password';
-        }
+        $user        =& User::factory($_POST['username']);
```
paiva: `\User`, for consistency

php/libraries/User.class.inc
```diff
@@ -411,89 +422,76 @@ class User extends UserPermissions
...
+                          'why'   => '',
+                         );
+        $minimumLength = 8;
+        if (strlen($newPassword) < $minimumLength) {
+            $result['why'] = 'Passphrase is too short.';
+            return $result;
+        }
+        $data = \User::singleton()->getData();
+        if (password_verify($newPassword, $data['Password_hash'])) {
+            $result['why'] = 'You cannot keep the same passphrase.';
```
PapillonMcGill: should be "... same password."

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -980,29 +980,11 @@ class Edit_User extends \NDB_Form
...
-            // check password strength
-            if (!$isPasswordStrong) {
-                $errors['Password_Group']
-                    = 'The password is weak, or the passwords do not match';
-            } else {
-                // New password must be different than current one
-                if (!$this->_passwordChanged($DB, $values['Password_hash'])) {
-                    $errors['Password_Group'] = 'New and old passwords '
-                        . 'are identical: please choose another one';
+            $newPassword = $values['Password_hash'];
```
PapillonMcGill: Do we want to check with latest 5 password or some numbering scheme like
myPa$$word1
myPa$$word2
myPa$$word3
...

php/libraries/User.class.inc
johnsaigle: No, there is no standard right now. I like passphrase but don't feel that strongly about it. I'll bring it up in the LORIS meeting.

modules/user_accounts/php/edit_user.class.inc
johnsaigle: We could do so. Usually this type of thing is done when there is a password expiration policy and people want to basically use the same password anyway. I can't remember exactly but I think our expiration window is fairly large so I don't think our users will be compelled to do this. I am planning another PR to remove password expiration altogether because it's not considered a good thing to have for exactly this reason.

Note also that this will require a database change as we only store the current `Password_hash` in the user table, not previous passwords.

php/libraries/SinglePointLogin.class.inc
```diff
@@ -99,42 +99,18 @@ class SinglePointLogin
...
+        $result = \User::validatePassword($newPassword);
+        if (!$result['valid']) {
+            $this->_lastError = $result['why'];
             return false;
         }
 
         // Reset passwordExpired flag
         $this->_passwordExpired = false;
-        $user->updatePassword($_POST['password']);
+        $user->updatePassword($newPassword);
```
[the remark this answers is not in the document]
johnsaigle: `$user->updatePassword()` uses `User::update()` which uses a prepared statement.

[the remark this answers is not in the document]
PapillonMcGill: perfect!

php/libraries/User.class.inc
PapillonMcGill: Then, just to keep the same word everywhere, I would use "password"

php/libraries/User.class.inc
```diff
@@ -435,21 +435,47 @@ class User extends UserPermissions
         /* This is an array of information related to the user. This is used by
          * Zxcvbn to check if a user is doing something foolish like making
          * their password 'johnnyloris'.
-         * We should add words as we think of them.
+         * $lorisKeywords comes from a list generated by the tool 'CeWL'
+         * https://tools.kali.org/password-attacks/cewl
+         * It represents words scraped from the loris.ca and mcin.ca websites
+         * and contains words that users are likely to include in their
+         * passwords.
```
driusan: I think it's a stretch to go from "this is a list of words used on the loris.ca or mcin.ca website" to "These are likely to be used by users of LORIS who have likely never been to either of those websites". Many of them are just references to projects which aren't likely to be used by anyone who isn't a member of that project, many of them are people's names (but not the logged in user), many of them are complex/scientific words used in texts of press releases which would probably be fairly secure to work into a password.

Just move "loris", "brain", and "mcgill" into a file and then you can include instructions somewhere on how to populate it with their own blacklist data source.

php/libraries/User.class.inc
```diff
@@ -435,21 +435,47 @@ class User extends UserPermissions
...
+            if ($fh !== false) {
+                while (($line = fgets($fh)) !== false) {
+                    if (strpos($line, '#') === 0) {
+                        continue;
+                    }
+                    array_push($keywords, trim($line));
+                }
+            }
+        } else {
+            error_log("Could not open keyword file at $keywordFile");
```
driusan: This is unnecessary

docs/LorisKeywords
```diff
@@ -0,0 +1,7 @@
+# This represents a list of keywords significant to LORIS. 
+# This file is used in our password logic; passwords containing a word from the 
+# below will be considered weak more likely to be rejected.
```
PapillonMcGill: ... from the *list* below ...

driusan:
```suggestion
# list below will be considered weak more likely to be rejected.
```

johnsaigle:
@PapillonMcGill @driusan Could you dismiss your reviews if you're satisfied with the changes?

php/libraries/User.class.inc
```diff
@@ -411,89 +422,106 @@ class User extends UserPermissions
...
+    /**
+     * Slurps contents of keyword file and returns a word list.
+     *
+     * @return array Keywords that shouldn't be used in passwords
+     */
+    private function _getLorisKeywords(): array
+    {
+        $keywords = array();
+        // The file is located two directories up from this file.
+        $keywordFile = __DIR__ . '/../../' . 'docs/LorisKeywords';
```
driusan: I don't think this is a very robust way of loading the file or a good place to put a config file that users might customize.

Can we move the file to `$PROJECT_DIR/password_blacklist.txt` (or something like that?)

driusan: (and put the default there on install)

johnsaigle: To be clear you mean something like "/var/www/loris/" and not the `project/` dir, right?

driusan: no, I mean it should be in the `project` directory since the project is going to customize it.

johnsaigle:
This PR has been made consistent with #4021 to reject the blacklisted passwords

johnsaigle:
@driusan When you have the time could you re-review and make sure the changes you requested have been resolved?

php/libraries/User.class.inc
johnsaigle: @ridz1208 I can reconfigure this PR to use a new table instead.

johnsaigle: @driusan Do you have a preference here? At this time the text file serves as both a password blacklist and as a list of plaintext words that should reduce a password's score (but not necessarily cause it to fail).

driusan: A table is probably more efficient for large blacklists

johnsaigle:
Adding Blocked until the discussion on #4021 is resolved. When a final DB schema is decided on there, this PR will be updated.

johnsaigle:
Re: dependency, should check whether this is available in Debian but it's not a deal breaker

johnsaigle:
TODO:
* Remove references to the blacklist text file
* Refactor to refer to the table generated by #4021 
* Use Guzzle for querying instead of curl

johnsaigle:
@driusan We discussed this in person but just so others are aware: yes this is blocked by #4021 or else the password blacklist check doesn't work.

johnsaigle:
Note: this will conflict with changes in https://github.com/aces/Loris/pull/4395 so this should probably be rebased onto that. In any case it's still Blocked...

php/libraries/User.class.inc
christinerogers: Would also recommend "password" for consistency.

johnsaigle:
#4021 has been closed and password blacklisting will be added as a feature at some point in the future. I will remove references to it in this PR so that things can proceed more quickly.

johnsaigle:
I'm going to close this PR in favour of adding a new `Password` class that validates all of these things in its constructor. I think it's a better approach than keeping this functionality within the `User` class.

Also the discussion around blacklisting has become convoluted and should be punted to another PR. It's a lower priority than making the passwords stronger.

PR #3963, opened 2018-09-19
[JSX] Add new FilterableDataTable component

driusan:
This adds a new FilterableDataTable component in order to make it easier to
create tables which have filters that dynamically restrict the content.

The filterable datatable component is like a DynamicDataTable, except it also
creates a filter based on a object passed to it as a prop. The data in the table
and the filters are dynamically linked.

This is intended to make the reactification of menu filters both easier and more consistent across LORIS.

### DIFF
--- a/jsx/FilterableDataTable.js
+++ b/jsx/FilterableDataTable.js
@@ -1,0 +1,36 @@
+import React, {Component} from 'react';
+
+import DynamicDataTable from 'jsx/DynamicDataTable';
+import FilterForm from 'jsx/FilterForm';
+
+class FilterableDataTable extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+        'filters': {},
+    };
+    this.updateFilter = this.updateFilter.bind(this);
+  }
+
+  updateFilter(filter) {
+      this.setState({'filters': filter});
+  }
+  render() {
+      return (<div>
+          <FilterForm
+            name={this.props.name + '_filter'}
+            id={this.props.name + '_filter_form'}
+            ref={this.props.name + 'Filter'}
+            columns={3}
+            formElements={this.props.FilterForm}
+            onUpdate={this.updateFilter}
+          />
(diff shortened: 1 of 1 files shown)


xlecours:
I would be nice to have a use of it. Dicom_archives?

driusan:
I was starting to use it to reactify candidate_list, but that turned into a bigger project than I expected so I separated this out.

ridz1208:
@davidblader @HenriRabalais take a look

jsx/FilterableDataTable.js
```diff
@@ -0,0 +1,37 @@
...
+    this.setState({filters: filter});
+  }
+
+  render() {
+    return (
+      <div>
+        <FilterForm
+          name={this.props.name + '_filter'}
+          id={this.props.name + '_filter_form'}
+          ref={this.props.name + 'Filter'}
```
zaliqarosli: we are trying to remove all string refs as its getting deprecated so should probably change this to a callback method

driusan: Is there an example somewhere of how to do this? I based this on (IIRC) the media module and just replaced the module name with a prop.

zaliqarosli: https://github.com/aces/Loris/pull/4019 and https://reactjs.org/docs/refs-and-the-dom.html#callback-refs

davidblader: better yet, don't use refs at all. the use of refs in the other modules was used due to a lack of an appropriate common parent component between the filterform and the data table. with this component it looks like the use of refs for filter clearing can be deprecated altogether

xlecours: I agree with @davidblader but it implies that the datatable should take data as props. Fortunately, the StaticDataTable does exacly that

HenriRabalais:
PR #4054 has absorbed this one. Suggest closing.

PR #3964, opened 2018-09-19
[Core]  Populate Instruments in Alphabetic Order

sruthymathew123:
### Brief Summary of Changes
This PR makes the Instruments in the Instrument dropdown in alphabetic order.
 
### To test this change...

- Any place where Utility::getDirectInstruments() or Utility::getAllInstruments() is called
(eg: Data Dictionary Instrument Dropdown, Survey Accounts Instrument Dropdown etc)

### DIFF
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -540,7 +540,7 @@ class Utility
         $Factory       = NDB_Factory::singleton();
         $DB            = $Factory->Database();
         $instruments_q = $DB->pselect(
-            "SELECT Test_name,Full_name FROM test_names",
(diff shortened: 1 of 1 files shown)


php/libraries/Utility.class.inc
```diff
@@ -540,7 +540,7 @@ class Utility
         $Factory       = NDB_Factory::singleton();
```
kongtiaowang: \NDB_Factory::singleton();
It will pass travis.

kongtiaowang:
@sruthymathew123 
You need to change the unit test as well.
test/unittests/api/v0.0.1/projects/Project_Test.php line 56
if($query == "SELECT Test_name,Full_name FROM test_names **ORDER BY Test_name**") {

PR #3966, opened 2018-09-19
[Core] Add hasAllPermissions and hasAnyPermission to facilitate privilege-checking

johnsaigle:
### Brief summary of changes

Now we can do

```php
return $user->hasAllPermissions(array('permission1', 'permission2', ...));
```
or
```php
return $user->hasAnyPermission(array('permission1', 'permission2', ...));
```

instead of 

```php
return $user->hasPermission('examiner_view')
            || $user->hasPermission('examiner_multisite')
            || ....;
```

Which will help to make it easier to review permissions.

### To test this change...

- [ ] Go to examiner and make sure the permissions still work as expected.

### DIFF
--- a/php/libraries/UserPermissions.class.inc
+++ b/php/libraries/UserPermissions.class.inc
@@ -113,7 +113,7 @@ class UserPermissions
      *
      * @return bool If the user has a permission.
      */
-    function hasPermission($code): bool
+    function hasPermission(string $code): bool
     {
         /* True if user superuser permission is set and user has it OR all
          * permissions are true.
@@ -130,7 +130,7 @@ class UserPermissions
      *
      * @return bool The setting of permission code $permission
      */
-    function permissionTrueAndValid(string $code)
+    function permissionTrueAndValid(string $code): bool
     {
         if (!isset($this->permissions[$code])) {
             throw new ConfigurationException("Invalid permission code $code");
@@ -148,7 +148,7 @@ class UserPermissions
      */
     function hasAllPermissions(array $permissions): bool
     {
-        if (!count($permissions)) {
+        if (count($permissions) === 0) {
             throw new LorisException(
                 "Cannot call hasAllPermissions with an empty array!"
             );
@@ -170,7 +170,7 @@ class UserPermissions
      */
     function hasAnyPermission(array $permissions): bool
     {
-        if (!count($permissions)) {
+        if (count($permissions) === 0) {
             throw new LorisException(
                 "Cannot call hasAnyPermission with an empty array!"
             );
--- a/php/libraries/UserPermissions.class.inc
+++ b/php/libraries/UserPermissions.class.inc
@@ -109,7 +109,7 @@ class UserPermissions
      * Determines if the user has the permission to access a specific module
      * or page.
      *
-     * @param string $code     The permission code
+     * @param string $code The permission code
      *
      * @return bool If the user has a permission.
      */
@@ -142,12 +142,17 @@ class UserPermissions
      * Used to test whether a user has ALL of an array of specified permissions.
      * Returns false on the first permission that the user doesn't have.
      *
-     * @param array $permissions Array of permissions codes.
+     * @param string[] $permissions Array of permissions codes.
      *
      * @return bool Whether the user has all the permissions in $permissions.
      */
     function hasAllPermissions(array $permissions): bool
     {
+        if (!count($permissions)) {
+            throw new LorisException(
+                "Cannot call hasAllPermissions with an empty array!"
+            );
+        }
         foreach ($permissions as $p) {
             if (!$this->permissionTrueAndValid($p)) {
                 return false;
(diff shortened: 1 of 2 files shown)


driusan:
I don't see the value in this. It adds confusion over whether `hasPermission` is an "any" or an "all" relationship. It can be moved to a less ambiguous function name, but in that case I don't see any benefit of making it variadic over just accepting an array parameter.

php/libraries/UserPermissions.class.inc
```diff
@@ -106,30 +106,63 @@ class UserPermissions
...
      *
      * @return bool
      */
-    function hasPermission($code)
+    function hasPermission($code, ...$codes): bool
+    {
+        /* True if $code is the only parameter passed and it is a valid
+         * permission code possessed by the user.
+         */
+        $hasSinglePermission = $this->permissionTrueAndValid($code)
```
paiva: Well written, it could have been a bit easier with an if statement, but the logic is correct

php/libraries/UserPermissions.class.inc
```diff
@@ -163,6 +156,24 @@ class UserPermissions
         return true;
     }
 
+    /**
+     * Used to test whether a user has ANY of an array of specified permissions.
+     * Returns true on the first permission that the user has.
```
driusan: The second sentence is unnecessary.

I think the behaviour of this is also wrong when passed an empty array. It should either be "true" (the empty set is a subset of any set) or an exception (it's probably a bug if called with an empty array)

php/libraries/UserPermissions.class.inc
```diff
@@ -163,6 +156,24 @@ class UserPermissions
...
+     *
+     * @param array $permissions Array of permissions codes.
```
driusan: probably better to be "string[]" than an untyped array.

php/libraries/UserPermissions.class.inc
johnsaigle: That's a good tip, thanks.

php/libraries/UserPermissions.class.inc
johnsaigle: Didn't know you could do that! I'll make the change.

php/libraries/UserPermissions.class.inc
```diff
@@ -106,30 +106,83 @@ class UserPermissions
...
+     * Determines if the user has the permission to access a specific module
+     * or page.
      *
      * @param string $code The permission code
      *
-     * @return bool
+     * @return bool If the user has a permission.
      */
-    function hasPermission($code)
+    function hasPermission(string $code): bool
```
driusan: since this is going to major and we're changing the type signature, maybe we should take the opportunity to explicitly declare this public too.

php/libraries/UserPermissions.class.inc
```diff
@@ -106,30 +106,83 @@ class UserPermissions
...
+
+    /**
+     * Check if a given permission code is present in a user's list of
+     * permissions. Also validates that the permission code is real.
+     *
+     * @param string $code Permission code to test.
+     *
+     * @return bool The setting of permission code $permission
+     */
+    function permissionTrueAndValid(string $code): bool
```
driusan: private

php/libraries/UserPermissions.class.inc
```diff
@@ -106,30 +106,83 @@ class UserPermissions
...
+
+    /**
+     * Used to test whether a user has ALL of an array of specified permissions.
+     * Returns false on the first permission that the user doesn't have.
+     *
+     * @param string[] $permissions Array of permissions codes.
+     *
+     * @return bool Whether the user has all the permissions in $permissions.
+     */
+    function hasAllPermissions(array $permissions): bool
```
driusan: public

php/libraries/UserPermissions.class.inc
```diff
@@ -106,30 +106,83 @@ class UserPermissions
...
-        ) {
-            return true;
+    /**
+     * Used to test whether a user has ANY of an array of specified permissions.
+     *
+     * @param string[] $permissions Array of permissions codes.
+     *
+     * @return bool Whether the user has any of the permissions in $permissions.
+     */
+    function hasAnyPermission(array $permissions): bool
```
driusan: public

driusan:
@johnsaigle I think this is fine now other than the fact that it's failing Travis.

php/libraries/UserPermissions.class.inc
```diff
@@ -106,30 +106,85 @@ class UserPermissions
...
-    function hasPermission($code)
+    public function hasPermission(string $code): bool
+    {
+        if (get_class() === 'UserPermissions') {
+            return $this->_permissionTrueAndValid['superuser']
+                || $this->_permissionTrueAndValid($code);
+        } else if (get_class() === 'User') {
+            return parent::_permissionTrueAndValid['superuser']
+                || parent::_permissionTrueAndValid($code);
+        }
```
driusan: This isn't how polymorphism is supposed to work..

PR #3969, opened 2018-09-20
[Core] Add Response object implementation

driusan:
A couple places in our code are tightly coupled to the Zend Diactoros
Response object. This weaks the coupling by adding a class in between
which can be changed to another PSR15 implementation if desired. It
also solves the problem wherein I can never remember how to spell
"diactoros".

### DIFF
--- a/php/libraries/NDB_Form.class.inc
+++ b/php/libraries/NDB_Form.class.inc
@@ -190,7 +190,7 @@ class NDB_Form extends NDB_Page
             $this->save();
         }
 
-        return (new \Zend\Diactoros\Response())
+        return (new \LORIS\Http\Response())
             ->withBody(new \LORIS\Http\StringStream($this->display() ?? ""));
     }
 }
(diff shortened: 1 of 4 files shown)


intralizee:
@driusan Looks like a great change. Since loris is configured to be https and never http. Should the code resemble this?

driusan:
I don't know where you got "Loris is never configured over http". LORIS is agnostic to the transport protocol used, whether it's over http, https, http2, or some other protocol. (In fact, it's run over http on travis.)

But regardless, this change doesn't make any reference to http or https?

intralizee:
@driusan I thought the setup process requires https to be configured for security. Was thinking maybe \LORIS\Http\Response() should be \LORIS\Https\Response() but I guess I'm mistaken.

driusan:
`LORIS\Http` is just the namespace. It's named that way to mirror the PSR7 namespace.

PR #3970, opened 2018-09-20
[API] Add new API module and begin transition

driusan:
This adds a new API module to replace the ad-hoc PHP
scripts and mod_rewrite rules which currently make up the
LORIS API. The module currently handles the login endpoint
only, but does so for all versions of the API rather than
having a separate handler for each.

The mod_rewrite rules are updated such that the login
endpoint is delegated to the module, while other endpoints
fall back on the older implementations to help the transition.

To test this try using the API login endpoint, as well as
an endpoint which is not handled by the module. Both should
login and non-login endpoints should still work.


modules/api/php/login.class.inc
```diff
@@ -0,0 +1,207 @@
...
+        if (!preg_match('/[^A-Za-z]/', $key)) {
+            return false;
+        }
+        // nothing but numbers
+        if (!preg_match('/[^0-9]/', $key)) {
+            return false;
+        }
+        // preg_match returns 1 on match, 0 on non-match
+        $CharTypes += preg_match('/[0-9]+/', $key);
+        $CharTypes += preg_match('/[A-Za-z]+/', $key);
```
PapillonMcGill: Could be split to force lower AND upper characters

driusan: This is directly copied from the existing function in `htdocs/api/v0.0.2/Login.php`, just moved to a new home.

driusan:
Merged in #3993

PR #3974, opened 2018-10-01
[data_release] Add permissions for upload and edit functionalities

ridz1208:
### Brief summary of changes
Currently in order to be able to upload and/or edit the access to files in the data release module one needs to have the `superuser` permission. Given that this permission gives access to so many other things, this PR separates these functionalities from `superuser` and adds 2 new permissions `data_release_upload` and `data_release_edit_permissions`.

### To test this change...

- [ ] source SQL patch
- [ ] follow the newly modified test plan
- [ ] make sure documentation was properly updated everywhere

### DIFF
--- a/SQL/0000-00-01-Permission.sql
+++ b/SQL/0000-00-01-Permission.sql
@@ -109,7 +109,9 @@ INSERT INTO `permissions` VALUES
     (45,'issue_tracker_reporter', 'Can add a new issue, edit own issue, comment on all', 2),
     (46,'issue_tracker_developer', 'Can re-assign issues, mark issues as closed, comment on all, edit issues.', 2),
     (47,'imaging_browser_phantom_allsites', 'Can access only phantom data from all sites in Imaging Browser', 2),
-    (48,'imaging_browser_phantom_ownsite', 'Can access only phantom data from own site in Imaging Browser', 2);
+    (48,'imaging_browser_phantom_ownsite', 'Can access only phantom data from own site in Imaging Browser', 2),
+    (49,'data_release_upload', 'Data Release: Upload file', 2),
+    (50,'data_release_edit_permissions', 'Data Release: Give user permission to view files', 2);
 
 
 INSERT INTO `user_perm_rel` (userID, permID)
--- a/SQL/0000-00-01-Permission.sql
+++ b/SQL/0000-00-01-Permission.sql
@@ -111,7 +111,7 @@ INSERT INTO `permissions` VALUES
     (47,'imaging_browser_phantom_allsites', 'Can access only phantom data from all sites in Imaging Browser', 2),
     (48,'imaging_browser_phantom_ownsite', 'Can access only phantom data from own site in Imaging Browser', 2),
     (49,'data_release_upload', 'Data Release: Upload file', 2),
-    (50,'data_release_edit_permissions', 'Data Release: Give user permission to view files', 2);
+    (50,'data_release_edit_file_access', 'Data Release: Grant other users view-file permissions', 2);
 
 
 INSERT INTO `user_perm_rel` (userID, permID)
--- a/SQL/New_patches/2018-10-01-Add_data_release_permissions.sql
+++ b/SQL/New_patches/2018-10-01-Add_data_release_permissions.sql
@@ -1,0 +1,6 @@
+INSERT INTO permissions (code,description,categoryID) VALUES ('data_release_upload','Data Release: Upload file',2), ('data_release_edit_permissions','Data Release: Give user permission to view files',2);
+
+INSERT IGNORE INTO user_perm_rel 
+SELECT upr.userID, p.permID FROM user_perm_rel upr JOIN permissions p WHERE upr.permID=1 AND p.code IN ('data_release_upload', 'data_release_edit_permissions');
+
+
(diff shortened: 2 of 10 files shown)


modules/data_release/templates/menu_data_release.tpl
```diff
@@ -1,4 +1,4 @@
-{if $superuser}
+{if $usecanupload}
```
johnsaigle: Typo

modules/data_release/php/data_release.class.inc
```diff
@@ -97,7 +89,8 @@ class Data_Release extends \NDB_Menu_Filter
             $data_release_ids[$row['id']] =$row['file_name'];
         }
 
-        $this->tpl_data['superuser']        = $superuser;
+        $this->tpl_data['usercanupload']    = $userCanUpload;
+        $this->tpl_data['usercanedit']      = $userCanEdit;
```
johnsaigle: Can you change this to be "user_can_*"? The underscores are easier to read.

modules/data_release/templates/menu_data_release.tpl
```diff
@@ -36,8 +36,8 @@
         </div>
     </div>
 </div>
-
-
+{/if}
+{if $usecanedit}
```
johnsaigle: typo again

modules/data_release/test/data_release_test_plan.md
```diff
@@ -1,30 +1,42 @@
...
+Make sure that a 'superuser' can see the 'Upload File' and 'Add 
+Permission' buttons at the top of the module, as well as the list of already 
+uploaded files
+
+## Sort table
  
- 5. Make sure that clicking on the table headers will sort the data present 
- in the data release table according to that column.
+Make sure that clicking on the table headers will sort the data present 
+ in the data release table according to that column\
```
johnsaigle: Should be period, not a backslash

SQL/0000-00-01-Permission.sql
```diff
@@ -109,7 +109,9 @@ INSERT INTO `permissions` VALUES
     (45,'issue_tracker_reporter', 'Can add a new issue, edit own issue, comment on all', 2),
     (46,'issue_tracker_developer', 'Can re-assign issues, mark issues as closed, comment on all, edit issues.', 2),
     (47,'imaging_browser_phantom_allsites', 'Can access only phantom data from all sites in Imaging Browser', 2),
-    (48,'imaging_browser_phantom_ownsite', 'Can access only phantom data from own site in Imaging Browser', 2);
+    (48,'imaging_browser_phantom_ownsite', 'Can access only phantom data from own site in Imaging Browser', 2),
+    (49,'data_release_upload', 'Data Release: Upload file', 2),
+    (50,'data_release_edit_permissions', 'Data Release: Give user permission to view files', 2);
```
johnsaigle: This also gives a user permissions to edit files, not just view

SQL/New_patches/2018-10-01-Add_data_release_permissions.sql
```diff
@@ -0,0 +1,6 @@
+INSERT INTO permissions (code,description,categoryID) VALUES ('data_release_upload','Data Release: Upload file',2), ('data_release_edit_permissions','Data Release: Give user permission to view files',2);
```
johnsaigle: Same here. Edit and view, not just view.

modules/data_release/php/data_release.class.inc
```diff
@@ -68,16 +68,8 @@ class Data_Release extends \NDB_Menu_Filter
...
-            array('UserID' => $user->getUsername())
-        );
-        $superuser = $DB->pselectOne(
-            "SELECT true FROM users WHERE ID IN "
-            . "(SELECT userid FROM user_perm_rel WHERE "
-            . "permid=1 AND userid=:UserID)",
-            array('UserID' => $user_ID)
-        );
+        $userCanUpload = $user->hasPermission("data_release_upload");
+        $userCanEdit   = $user->hasPermission("data_release_edit_permissions");
```
johnsaigle: Maybe the permission should be called just "data_release_edit"? We already know it's a permission.

ridz1208: Lol the "permission" is what u edit its not the editing permission. I.e. data_release module lets you control access to specific files which is the add permission functionality. I appendedn _pernission to the name because you might want a diff edit code for lets say changing the file name which would be edit_file_name

ridz1208: I could maybe rename it edit_access ?

SQL/0000-00-01-Permission.sql
ridz1208: no... again this is confusing. what this permission does is it allows the user that has it to GIVE ANOTHER user viewing permission for files (no edit) i.e. if user A has "edit" permission, user A can then EDIT the viewing permissions of user B on a file..


@johnsaigle if you get what I mean maybe you can formulate it better. As I suggested for the permission code, maybe replace with "Give user access to file" ?

SQL/New_patches/2018-10-01-Add_data_release_permissions.sql
ridz1208: see comment above

ridz1208:
@Jkat I dont see any reference to the permissions in addPermission.php

SQL/0000-00-01-Permission.sql
johnsaigle: How about something like "Grant other users view-file permissions"?

ridz1208:
I will look at you improvements PR maybe we can get that first

SQL/0000-00-01-Permission.sql
ridz1208: sounds good
what about the code field ?

modules/data_release/php/data_release.class.inc
ridz1208: @HenriRabalais @driusan you could also voice your opinion since you agree with @johnsaigle

ridz1208:
closing in favour of linked PR

Issue #3972, opened 2018-09-25
httpd cannot be accessed


Tsite2007:
When I start the httpd service , the error log say:
[Tue Sep 25 15:35:36 2018] [alert] [client 202.85.216.21] /var/www/loris/htdocs/.htaccess: RewriteRule: unknown flag 'END'.
so, the apache service   cannot be accessed

driusan:
What version of apache are you using? The END flag (and therefore LORIS) requires apache >= 2.4

johnsaigle:
I am closing this as the bug most likely resulted from the wrong Apache version and the linked PR has made it clear what version LORIS requires.

PR #3976, opened 2018-10-01
[Docs] Specify Apache 2.4 required in README

johnsaigle:
### Brief summary of changes

Specifiy that Apache 2.4 is required.

Heroku deploy is no longer "new" so that is removed.

### This addresses issue...

- [x] Github? #3972

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Please consult the [LORIS Wiki Setup Guide](https://github.com/aces/Loris/wiki/S
 # Prerequisites for Installation
 
  * Apache **2.4** or higher
- * MySQL 5.7
+ * MySQL >= 5.7 (or MariaDB >= 10.3) 
  * PHP <b>7.2</b> or higher
  * [Composer](https://getcomposer.org/)
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,6 @@ Please consult the [LORIS Wiki Setup Guide](https://github.com/aces/Loris/wiki/S
  * Apache **2.4** or higher
  * MySQL 5.7
  * PHP <b>7.2</b> or higher
- * Package manager (for LINUX distributions)
 
 <b>Important:</b>
  * Composer should be run with --no-dev option unless you are an active LORIS developer. 
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Please consult the [LORIS Wiki Setup Guide](https://github.com/aces/Loris/wiki/S
 
 # Prerequisites for Installation
 
- * LINUX (supported on Ubuntu 14+ and [CentOS 6.5](https://github.com/aces/Loris/blob/master/README.CentOS6.md))
+ * LINUX (supported on Ubuntu 14+ and [CentOS 7](https://github.com/aces/Loris/blob/master/README.CentOS6.md))
  * Apache **2.4** or higher
  * MySQL 5.7
  * PHP <b>7.2</b> or higher
--- a/README.md
+++ b/README.md
@@ -24,15 +24,12 @@ Please consult the [LORIS Wiki Setup Guide](https://github.com/aces/Loris/wiki/S
 
 # Prerequisites for Installation
 
- * LINUX (supported on Ubuntu 14+ and [CentOS 7](https://github.com/aces/Loris/blob/master/README.CentOS6.md))
  * Apache **2.4** or higher
  * MySQL 5.7
  * PHP <b>7.2</b> or higher
  * Package manager (for LINUX distributions)
- * Composer : should be run with --no-dev option
 
 <b>Important:</b>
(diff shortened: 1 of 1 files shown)


johnsaigle:
@driusan done!

README.md
```diff
@@ -25,9 +25,9 @@ Please consult the [LORIS Wiki Setup Guide](https://github.com/aces/Loris/wiki/S
 # Prerequisites for Installation
 
  * LINUX (supported on Ubuntu 14+ and [CentOS 6.5](https://github.com/aces/Loris/blob/master/README.CentOS6.md))
```
davidblader: might want to also specify CentOS 7 here since apache 2.4 doesn't work on < 7

johnsaigle: Thanks, didn't know!

driusan: As long as we're on the topic, maybe we should take out the line entirely. LORIS doesn't have any Linux-only dependencies.

README.md
```diff
@@ -24,15 +24,12 @@ Please consult the [LORIS Wiki Setup Guide](https://github.com/aces/Loris/wiki/S
...
 
- * LINUX (supported on Ubuntu 14+ and [CentOS 7](https://github.com/aces/Loris/blob/master/README.CentOS6.md))
  * Apache **2.4** or higher
  * MySQL 5.7
  * PHP <b>7.2</b> or higher
  * Package manager (for LINUX distributions)
- * Composer : should be run with --no-dev option
 
 <b>Important:</b>
- * If you are upgrading your LORIS, you'll also want to upgrade to both PHP 7.2 and MySQL 5.7.
```
driusan: This still seems worth having.

README.md
```diff
@@ -24,15 +24,12 @@ Please consult the [LORIS Wiki Setup Guide](https://github.com/aces/Loris/wiki/S
 
 # Prerequisites for Installation
 
- * LINUX (supported on Ubuntu 14+ and [CentOS 7](https://github.com/aces/Loris/blob/master/README.CentOS6.md))
  * Apache **2.4** or higher
  * MySQL 5.7
  * PHP <b>7.2</b> or higher
  * Package manager (for LINUX distributions)
- * Composer : should be run with --no-dev option
```
driusan: composer is still required. But maybe the line above it about needing a package manager should be removed while we're at it?

README.md
```diff
@@ -24,15 +24,12 @@ Please consult the [LORIS Wiki Setup Guide](https://github.com/aces/Loris/wiki/S
 
 # Prerequisites for Installation
 
- * LINUX (supported on Ubuntu 14+ and [CentOS 6.5](https://github.com/aces/Loris/blob/master/README.CentOS6.md))
- * Apache2
- * MySQL 5.7 
- * PHP <b>7.2</b>  
+ * Apache **2.4** or higher
+ * MySQL 5.7
```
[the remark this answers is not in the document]
driusan: I think @ridz1208 has tested it.

README.md
johnsaigle: I removed this line because it is basically a duplicate from line 36.

I'll remove the line above.

README.md
johnsaigle: You don't think it's redundant given that we have a bulleted list of prereqs above?

README.md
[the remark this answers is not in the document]
johnsaigle: Which version?

README.md
driusan: True, that makes sense.

driusan: But I don't think "Important" is a very good title anymore. "Note" might be more accurate. If you don't specify --no-dev, you just get more dependencies installed.

README.md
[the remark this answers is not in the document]
driusan: 10.3, I think. @ridz1208 ?

README.md
driusan: It would make more sense to me to keep it here and delete it below, because installing it is still a prerequisite.

README.md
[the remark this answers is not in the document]
driusan: @ridz1208 The MariaDB version is the only thing preventing this from being merged

README.md
```diff
@@ -24,15 +24,12 @@ Please consult the [LORIS Wiki Setup Guide](https://github.com/aces/Loris/wiki/S
...
 # Prerequisites for Installation
 
- * LINUX (supported on Ubuntu 14+ and [CentOS 6.5](https://github.com/aces/Loris/blob/master/README.CentOS6.md))
- * Apache2
- * MySQL 5.7 
- * PHP <b>7.2</b>  
- * Package manager (for LINUX distributions)
- * Composer : should be run with --no-dev option
+ * Apache **2.4** or higher
+ * MySQL 5.7
```
driusan:
```suggestion
 * MySQL >= 5.7 (or MariaDB >= 10.3) 
```

PR #3977, opened 2018-10-01
[candidate_list] Convert ajax script to real LORIS endpoint

driusan:
This converts the validateProfileIDs AJAX script from an ajax
dir hack into a real LORIS endpoint. It also makes the page
decoration middleware slightly smarter so that it only decorates
pages where the Content-Type header is either empty or "text/html"
so that things that are explicitly not HTML don't have the HTML
page decorations added.


modules/candidate_list/php/validateids.class.inc
```diff
@@ -0,0 +1,109 @@
+<?php
+/**
+* The candidate list menu
```
johnsaigle: This comment and the class comment need a little bit of love.

modules/candidate_list/php/validateids.class.inc
```diff
@@ -0,0 +1,109 @@
...
+*
+* PHP Version 5
```
johnsaigle: Should be version 7

modules/candidate_list/php/validateids.class.inc
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Implements the candidate_list menu
+ *
+* @category Main
+* @package  Candidate_List
+* @author   Loris Team <loris.mni@bic.mni.mcgill.ca>
+* @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+* @link     https://www.github.com/aces/Loris/
+ */
+class ValidateIDs extends \NDB_Page
```
ridz1208: design question:

is this really a `page` ?? what does it have in common with page components ? isnt it more like a GENERAL request than anything else (applies to all ajax scripts) especially that the first line says "skip template" kind of proving my point

the way I see it, ajax scripts in loris are either just a huge mess or one of 3 things:
- request for validation
- request for data (getData)
- request for saving data (fileUpload, addThis, editThat,....)

shouldn't we just have at least these 3 as interfaces and have newly created classes implement one or more of them ?

driusan: NDB_Page is the root of our endpoint/page type object hierarchy at the moment. There's nowhere higher for it to go.

ridz1208: thats too bad then i guess

driusan:
@johnsaigle Fixed the comments on this, I think

PR #3978, opened 2018-10-02
[Docker] Updating Docker MySQL version

Jkat:
### Brief summary of changes

We should match the MySQL versions we're using right

### To test this change...

- [ ]


kongtiaowang:
It works, ready to be merged.

PR #3981, opened 2018-10-03
[Tools] Ubuntu production environment "bootstrap" script

johnsaigle:
*This pull request has been heavily refactored so a lot of the discussion may not make sense.*

### Brief summary of changes

`bootstrap` verifies loris dependencies such as Apache/PHP versions and misc system tools and optionally installs them. This is meant to be run only on development environments as production environments will in the future have all of this handled by a packaged release.

`PHP_CLI_Helper` consists of helper functions to make it easier to write PHP scripts for the command line.

### Further work
* Make something equivalent for CentOS as it's another environment that our team uses for development.

### To test this change...

- [ ]  `tools/bootstrap` and let me know if there are any errors!


johnsaigle:
Adding `Needs Work` for now because I won't be at work until Tuesday but Rida wanted to see it and I feel others should too!

Makefile
```diff
@@ -1,9 +0,0 @@
-all:
-	composer install --no-dev
-	npm run compile
-	git describe --tags --always > VERSION
-
-dev:
```
johnsaigle: This functionality is preserved by `setup` and `bootstrap`; one can add a 'dev' argument upon invoking the script and have composer do the appropriate thing.

script/bootstrap
```diff
@@ -0,0 +1,202 @@
...
+        "apache2 -v | " . 
+        "head -n 1 | " .
+        "cut -d ' ' -f 3"
+    )
+);
+$apache_version = end($apache_parts);
+/* Look for the string "$major.$minor" in info string. Also match on versions
+ * higher than minor version because we want AT LEAST that version.
+ */
+$pattern = "/$required_major_apache\.[$required_minor_apache-9]/";
```
PapillonMcGill: 2.2.4 will pass the 2.4 minimun requirement

add \.[0-9]

script/bootstrap
```diff
@@ -0,0 +1,202 @@
...
+// Dependencies last updated for version: 20.0.1
+$apt_dependencies = [
+                       "wget",
+                       "zip",
+                       "unzip",
+                       "php-json",
+                       "npm",
+                       "software-properties-common",
+                       "php-ast",
+                       "php$required_php",
```
PapillonMcGill: depending on the distro, this might fail for the required version.

If the latest Ubuntu version directly support the required php version, it will simply be listed as "php". 
php -v is required

script/bootstrap
```diff
@@ -0,0 +1,202 @@
...
+
+// Run package managers and add dev flag if supplied to the script
+if (runPackageManagers(isset($argv[1]) && $argv[1] === 'dev')) {
+    echo '[**] Dependencies up-to-date.' . PHP_EOL;
+}
+
+/**
+ * Prints a list of missing apt packages and prompts user to install them.
+ *
+ * @param array $requirements Required packages determined missing earlier
```
PapillonMcGill: since other lines end with a dot, this line should also end with a dot
same of other doc block

script/bootstrap
```diff
@@ -0,0 +1,202 @@
...
+    } else {
+        $cmd = "sudo apt-get install ";
+    }
+    $cmd .= escapeshellarg($name);
+    return doExec($cmd);
+}
+
+/**
+ * Runs 3rd-party package managers using exec
+ *
```
PapillonMcGill: missing @param

script/bootstrap
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * Takes an array of packages to install using apt-get.  Uses exec to install
+ * or upgrade apt packages based on $upgrade_mode.
+ *
+ * @param array $packages     List of apt packages to install
+ * @param bool  $upgrade_mode If true, only upgrades packages. False to install
+ *
+ * @return bool true if all packages installed properly. False otherwise.
+ */
+function installAptPackages($packages, $upgrade_mode = false) : bool
```
PapillonMcGill: param type definition is missing (as well as in many others function)

script/bootstrap
johnsaigle: Yes this is Ubuntu-specific as written. I don't have a CentOS VM to test on at the moment but if you do I'd appreciate some feedback about what needs to be added!

Ubuntu typically does not support the PHP version we need... it seems slow to get mixed into Debian/Ubuntu which is why adding the external PPA is baked-in here. 

`php -v` won't be sufficient as often the CLI PHP version differs from the version used in Apache. We aren't able to configure this for people necessarily as the Apache configuration might not follow the standards we use internally.

script/bootstrap
johnsaigle: Added!

driusan:
Replacing a simple < 10 line Makefile with hundreds of lines of bespoke PHP scripts in order to adopt a third party's internal procedures is not a step forward.

If you want to simplify running the setup add some standard targets like `make clean`, `make test` or `make install`.

ridz1208:
hmmm, let's see:
 - first, the <10 line makefile is doing soooo much less than the hundreds of lines of bespoke php
 - `make` is a tool not a standard and we are in dying need of standards (especially some not defined by LORIS for LORIS) and these ones have some documentation and play ball with github... especially that they are described by [GITHUB Eng](https://githubengineering.com/scripts-to-rule-them-all/) and we use github for PRs, issues, projects, why not standards !!
 - Talking about standards, the intent of scripts to rule them all is to **standardize these scripts** in multi-project development and seems to me that LORIS, LORIS-MRI, CCNA, IBIS, BIOBANK, etc... do fit that picture pretty nicely. imagine having a **standard** way of setting them up or cheking for dependencies istead of ad hoc-ing our way during the installation. 
 - let's also not omit the fact that we can write these scripts in **whatever language we want** and we can have them do everything we need. no need to code `if` and `for` loops in bash !!! no we can use python, php what ever... that sounds pretty nice with an actual potential of being maintained
 - and how about just this... someone actually made the effort to code this and test it instead of just talking about it and dreaming.

the way I see it, if LORIS is WAY TOO ATTACHED to makefiles, then just create makefiles and have them called FROM the scripts to rule them all, i dont care as long as I dont have to explain to each and every new developer how to run a test, how to run phpcs, how to run phan, ...

a quick read https://www.jefframnani.com/project-build-protocol/

driusan:
I'm just going to copy and paste the blog post that you linked to here which applies to this repo:

```
Scenario One: One Project, Many Commands

In this scenario, you’re mostly working on one main app all day. Your company’s monolith, for example.

Let’s say you’re working in a Django project. To run the Django unit tests you run:

DJANGO_SETTINGS_MODULE=myapp.settings.local_test python manage.py test

To run the JavaScript front-end tests you run:

node_modules/.bin/gulp test

Now, I’m lazy. I don’t want to have to remember these two commands, their environment variables, or their arguments. What I normally do in this situation is write a simple Makefile.

test: test-py test-js

test-py:
    DJANGO_SETTINGS_MODULE=myapp.settings.local_test REUSE_DB=1 ./manage.py test

test-js:
    node_modules/.bin/gulp test

Now when I come back to this project, I only have to remember to run make
test, and all my tests get run.

(In my real Makefile I use .PHONY targets to tell Make that these names don’t correspond to actual files.)

For this example, I only created a test target, but the common targets I like to use are: build, test, run, and dist. These targets create a common protocol I can depend on. They also serve as a form of executable documentation for how to do one of these tasks.

Makefiles can get pretty hairy, pretty quickly, so the minute a task starts getting hard to read or maintain, I’ll write a separate script to perform the task, then just delegate to the external script within the Makefile.

Why Make? Because it’s ubiquitous and it’s easy to install.
```

At 9 lines which don't branch or loop, you can't even claim we're at the "pretty hair, pretty quickly" point that would necessitate delegating to an external script inside of the Makefile. The complexity in this PR is added *by* removing the Makefile and attempting to rewrite the functionality in another language that isn't well suited to the task of building software.

The other scenerio in the blog post (which is a little tortured to say applies to LORIS, although arguably does with Loris-MRI) doesn't provide any rational for abandoning that approach rather than just standardizing the makefile targets across the projects. Here are a list of standard Makefile targets used by GNU: https://www.gnu.org/prep/standards/html_node/Standard-Targets.html (not all apply to LORIS. I'm also surprised that "test" isn't included.) Adopting a real standard that's been standardized for building software for decades rather than a blog post standard would have a number of benefits such as making it trivial to build packages for any supported OS using their standard procedures.

script/bootstrap
PapillonMcGill: CentOS (RHEL) tend to be ever slower than Ubuntu to catch up with the latest and greatest version of a package. 
I just wanted to be sure the script will always work on Ubuntu in the eventuality that Ubuntu (almost) catch up with the latest and greatest version of PHP.

And you are right about which version of PHP Apache use

PapillonMcGill: @johnsaigle 
FYI 
`rpm -q {package-name}`  will tell you if a package is installed
`rpm -qa` will list all packages that are installed

CentOS 7+ also have `dnf search {package-name}` but only for those installed using DNF (yum).

ridz1208:
@driusan 

Call me cynic but I think 

`Makefiles can get pretty hairy, pretty quickly, so the minute a task starts getting hard to read or maintain, I’ll write a separate script to perform the task, then just delegate to the external script within the Makefile.`

applies to us rather well. 

I think that even though it is using `another language that isn't well suited to the task of building software` it might have more chances to be maintained by your average LORIS developer who probably hasn't touched a make file in the last 10 years. 

`At 9 lines which don't branch or loop, you can't even claim we're at the "pretty hair, pretty quickly"` well that's because the 9 lines are not doing much. I'd be curious to see what the makefiles would look like with the same functionality as the scripts to rule them all (checking for versions, maybe handling errors....)

`The other sceario in the blog post (which is a little tortured to say applies to LORIS, although arguably does with Loris-MRI)` in our effort to streamline project releases, I'm surprised you don't see the potential of using this for projects, I would loove to havea single lioner for every release `scripts/update`. And wait... I thought the plan was to extract all modules into individual repos... that sounds pretty hairy to me. 

As far as the standards are concerned, you wanna choose the make standards over the scripts to rule them all, by all means do, but no one has taken initiative to do that and I don't see anyone doing it... (test is included in the make check btw) like I said earlier we gonna appreciate the fact that the work has been done !!! I'm not saying to merge this blindly but it's at least worth a discussion and some testing.

I would like to see a comparison between the GNU way that's been used for "decades" vs the github way that might be used in the next decades ?? who knows maybe we'd actually be on to something futuristic there

johnsaigle:
TODO:

* Revert changes to `Makefile` and `travis.yml`
* Rename instances of `scripts` to `tools`
* Document changes, etc. "tools/bootstrap.php can be used to verify or install missing system requirements for Ubuntu. Used only for dev environments...."
* Delete `setup` in favour of `make clean` target with the same bash commands
* Add a `make test` target that runs `tools/test.php`. This script may need to be scrapped; will coordinate with @driusan and PR #4002

tools/PHP_CLI_Helper.class.inc
```diff
@@ -0,0 +1,103 @@
...
+ * @return bool representing if tool is installed
+ */
+function installed($tool) : bool
+{
+    // `which` returns empty if a tool is not installed.
+    // shell_exec captures this output.
+    if (shell_exec("which $tool")) {
+        return true;
+    }
+    // check installed pacakages with dpkg. Returns 0 if installed.
```
PapillonMcGill:
```suggestion
    // check installed packages with dpkg. Returns 0 if installed.
```

tools/PHP_CLI_Helper.class.inc
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @return void
+ */
+function writeQuestion($question, $answers) : void
+{
+    echo $question . ' (' . implode('/', $answers) . '): ' . PHP_EOL;
+}
+
+/**
+ * Gets user input from STDIN and checks if it matches a valye in
```
PapillonMcGill:
```suggestion
 * Gets user input from STDIN and checks if it matches a value in
```

tools/PHP_CLI_Helper.class.inc
```diff
@@ -0,0 +1,103 @@
...
+ * $possibleAnswers.  If not, the default answer is used.  Inteded to follow
```
PapillonMcGill:
```suggestion
 * $possibleAnswers.  If not, the default answer is used.  Intended to follow
```

tools/PHP_CLI_Helper.class.inc
```diff
@@ -0,0 +1,103 @@
...
+ * function writeQuestion
+ *
+ * @param array $possibleAnswers Possible answers to a prompt
+ * @param array $defaultAnswer   Response if user entered invalid response
+ *
+ * @return string The user input representing their answer or the default answer
+ */
+function readAnswer($possibleAnswers, $defaultAnswer) : string
+{
+    $in     = fopen('php://stdin', 'rw+');
```
PapillonMcGill:
```suggestion
    $in     = fopen('php://stdin', 'r');
```
You can't write to STDIN

tools/PHP_CLI_Helper.class.inc
```diff
@@ -0,0 +1,103 @@
+<?php
+
+/**
+ * Use bash to determine if certain unix tools are installed.  Which is used for
```
driusan: This isn't using bash, it's using the shell

tools/PHP_CLI_Helper.class.inc
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @param string $tool Name of tool
+ *
+ * @return bool representing if tool is installed
+ */
+function installed($tool) : bool
+{
+    // `which` returns empty if a tool is not installed.
+    // shell_exec captures this output.
+    if (shell_exec("which $tool")) {
```
driusan: I think this would be more reliable if it used exec and checked the shell return code, rather than trying to parse stdout.

tools/PHP_CLI_Helper.class.inc
driusan: Are you sure about that? dpkg(1) says:

```
EXIT STATUS         top

       0      The requested action was successfully performed.  Or a check
              or assertion command returned true.

       1      A check or assertion command returned false.

       2      Fatal or unrecoverable error due to invalid command-line
              usage, or interactions with the system, such as accesses to
              the database, memory allocations, etc.
```

(but doesn't really clarify what "a check or assertion command" is)

tools/PHP_CLI_Helper.class.inc
driusan:
```suggestion
 * Gets user input from STDIN and checks if it matches a value in
```

```suggestion
 * Gets user input from STDIN and checks if it matches an option in
```

tools/PHP_CLI_Helper.class.inc
driusan:
```suggestion
 * $possibleAnswers.  If not, the default answer is used.  Intended to follow
```

tools/PHP_CLI_Helper.class.inc
```diff
@@ -0,0 +1,103 @@
...
+    echo $question . ' (' . implode('/', $answers) . '): ' . PHP_EOL;
+}
+
+/**
+ * Gets user input from STDIN and checks if it matches a valye in
+ * $possibleAnswers.  If not, the default answer is used.  Inteded to follow
+ * function writeQuestion
+ *
+ * @param array $possibleAnswers Possible answers to a prompt
+ * @param array $defaultAnswer   Response if user entered invalid response
```
driusan: In most cases if a user responds with an invalid response, shouldn't the scripts be re-prompting for a valid answer?

tools/PHP_CLI_Helper.class.inc
```diff
@@ -0,0 +1,103 @@
...
+
+    if (!in_array($answer, $possibleAnswers, true)) {
+        return $defaultAnswer;
+    }
+
+    return $answer;
+}
+
+/**
+ * Prints and executes a bash command using exec.  Prints an error message on
```
driusan: Still not necessarily bash.

tools/PHP_CLI_Helper.class.inc
```diff
@@ -0,0 +1,103 @@
...
+ * failure (a 0 exit code in bash is a success. Anything else is considered an
+ * error here.
```
driusan:
```suggestion
 * error here.)
```

tools/PHP_CLI_Helper.class.inc
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @param string $cmd An executable bash command
```
driusan:
```suggestion
 * @param string $cmd An executable shell command
```

tools/PHP_CLI_Helper.class.inc
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @return bool True if bash command exits normally. False otherwise.
```
driusan:
```suggestion
 * @return bool True if command exits normally. False otherwise.
```

tools/PHP_CLI_Helper.class.inc
```diff
@@ -0,0 +1,103 @@
...
+ */
+function doExec($cmd) : bool
+{
+    echo "[+] Executing bash command `$cmd`... " . PHP_EOL;
```
driusan:
```suggestion
    echo "[+] Executing command `$cmd`... " . PHP_EOL;
```

tools/bootstrap.php
```diff
@@ -0,0 +1,214 @@
...
+ */
+$pattern = "/$required_major_apache\.[$required_minor_apache-9].[0-9]/";
+// When preg_match returns 0 it means no match was found.
+if (preg_match($pattern, $apache_version) === 0) {
+    $required_apache = "$required_major_apache.$required_minor_apache";
+    die(
+        "ERROR: LORIS requires Apache v$required_apache or higher."
+        . PHP_EOL
+        . "This must be done manually as it has possible security ramifications."
+        . PHP_EOL
```
driusan: The second sentence is unnecessary.

tools/bootstrap.php
```diff
@@ -0,0 +1,214 @@
+#!/usr/bin/env php
+# This script verifies a development installation of LORIS by ensuring that
+# the system has all of the required dependencies such as the correct PHP and
```
driusan: What if PHP isn't installed?

tools/PHP_CLI_Helper.class.inc
johnsaigle: Hmm well I mean the `-s` flag gives the status of a package and returns 1 when not found so I think it works in this case.

tools/bootstrap.php
johnsaigle: I'll add a check for it

johnsaigle:
Closing this PR and will re-open a fresh one.

tools/bootstrap.php
driusan: I mean the script is written in PHP, so how can it check that?

PR #3982, opened 2018-10-03
[General] Don't track compiled files or VERSION on Git

johnsaigle:
### Brief summary of changes

On major, due to the creation of the build step and also webpack (?) changes, git complains about a bunch of untracked files when you try to switch branches

This PR 

* Ignores all files in `modules/*/js/`
* adds VERSION to the main `.gitignore`
* sorts `.gitignore`


.gitignore
```diff
@@ -1,31 +1,31 @@
...
-modules/*/js/.module-cache/
+.gitmodules
+htdocs/js/components/*.map
 htdocs/js/components/.module-cache/
+htdocs/my_tests/
+htdocs/phpunit_codecoverage/
+.idea
+*.linst~
+logs/*.log
+modules/*/js/*
```
johnsaigle: This is the line that matters

.gitignore
```diff
@@ -1,31 +1,31 @@
...
 SQL/Archive/autogenerated/single_use/*.sql
-SQL/Archive/autogenerated/*.sql
\ No newline at end of file
+SQL/Archive/autogenerated/*.sql
+tools/ip_output.txt.back
+tools/logs/*.log
+tools/*.txt
+*.un~
+/vendor
+VERSION
```
johnsaigle: And this one

Jkat:
I don't know this webpack stuff but if we don't track VERSION how would we update it properly

modules/candidate_parameters/js/.gitignore
```diff
@@ -0,0 +1 @@
+index.js
```
driusan: This should be '*', there's no files there that aren't auto generated (it's already * on my computer, I think it just wasn't added because it ignored itself.)

.gitignore
```diff
@@ -1,31 +1,32 @@
...
-project-*
-*.un~
-*.linst~
-*.php~
-/vendor
-.idea
-tools/*.txt
-htdocs/phpunit_codecoverage/
-htdocs/my_tests/
+composer.lock
```
driusan: This should not be ignored

PR #3957, opened 2018-09-17
[JSX] Media - Refactor and Clean-Up

HenriRabalais:
### Summary of changes
This PR refactors the Media jsx file to meet ES6 conventions and include the formatColumns function, rather than have it in a separate file.

### Testing
- Go to the Media module and ensure that everything is functioning.

PR #3984, opened 2018-10-04

HenriRabalais:
### Summary of changes
Added a Modal Window Component to LORIS.

### Why?

### Features

### Testing
- Go to the Media module menu filter. Click the `Edit` button on any of the datatable rows.

mw


jsx/Modal.js
```diff
@@ -0,0 +1,137 @@
...
+          this.props.closeModal();
+        }
+      });
+    } else {
+      this.props.closeModal();
+    }
+  }
+
+  render() {
+    const headerStyle = {
```
davidblader: why not put this in with the CSS file?

davidblader: alternatively might be worth considering making these into default values and allowing for users to define their own CSS passed along via prop

jsx/Modal.js
```diff
@@ -0,0 +1,137 @@
...
+    const glyphStyle = {
+      marginLeft: 'auto',
+      cursor: 'pointer'
+    };
+
+    const bodyStyle = {
+      padding: 15
+    };
+
+    const children = () => this.props.show && this.props.children;
```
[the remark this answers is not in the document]
HenriRabalais: Which part? The arrow function or the conditional allocation of `this.props.children`?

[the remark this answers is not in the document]
driusan: I suspect he means the "conditional allocation". We should be able to generally assume that people looking at javascript can figure out an arrow function (moreso as time moves on and they become more standard.)

P.S. What does this line do?

jsx/Modal.js
HenriRabalais: I've sort of been struggling with where to put styling with regard to react. I ended up with a bit of a mixed bag in this case because there didn't seem to be an easy way to do the css animation in react without adding third-party dependencies. I like the idea of a default with prop though.

jsx/Modal.js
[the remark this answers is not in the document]
davidblader: using `&&` short circuiting to conditionally return righthand value. just a matter of preference tho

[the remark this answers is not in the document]
driusan: So it's basically `const children = this.props.children || this.props.show`?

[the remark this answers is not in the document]
HenriRabalais: @davidblader I've sort of gotten used to this shorthand because it's really useful for quick checks that don't necessitate a ternary operator. Much like `||`, the null coalesce equivalent of `??` in php, the `&&` for setting values or triggering functions uses the same boolean-check principle and has become quite common practice from what I've seen. Not only that, but the principle isn't javascript specific and works in php too! 

@driusan when the function `children` is called, it will check whether `this.props.show` is true. If so, it will return `this.props.children`, else it will return `null`.

jsx/Modal.js
HenriRabalais: On second thought, for the sake of consistency I'm not sure if anyone should be changing these values. Is there anything in particular you feel needs to be overrideable/customizable?

jsx/Modal.js
```diff
@@ -0,0 +1,137 @@
...
+        }
+      });
+    } else {
+      this.props.closeModal();
+    }
+  }
+
+  render() {
+    const headerStyle = {
+      display: 'flex',
```
zaliqarosli: i agree with @davidblader . if all these stylings are constant, why not put them in the css file? if the variables are only providing styling, it may as well be made into a class that you would add to each elements className attribute no? then the css styling would be where they belong, the css file

jsx/Modal.js
```diff
@@ -0,0 +1,137 @@
...
+ * =================================================
+ *
+ */
+class Modal extends React.Component {
+  constructor() {
+    super();
+    this.closeModal = this.closeModal.bind(this);
+  }
+
+  closeModal() {
```
zaliqarosli: i like this closeModal function!

jsx/Modal.js
```diff
@@ -0,0 +1,137 @@
+/**
+ * This file contains the React Component for a Modal Window.
+ *
+ * @author Henri Rabalais
+ * @version 1.1.0
+ *
+ */
+import swal from 'sweetalert2';
```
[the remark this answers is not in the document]
HenriRabalais: Yes, as per the SweetAlert2 Usage documentation: https://sweetalert2.github.io/#usage

jsx/Modal.js
HenriRabalais: There's definitely an argument to be made in either direction (ref for list of styling methods: https://codeburst.io/4-four-ways-to-style-react-components-ac6f323da822) and I think at the end of the day it comes down to context and personal preference. I'm easy either way in this case, but I think this merits a discussion of LORIS-documented standards for react.

zaliqarosli: i think we should definitely update and make more robust the LORIS react standards. in this case however, i think it really comes down to code organization. there's a reason why we have css files and css directories, and in this case, i think they should be utilized as intended. this is however, my own personal opinion

HenriRabalais: Yes, code organisation definitely comes into question here. Precisely for that reason, an argument could be made that this styling code is well suited to stay within this class. The styling here is not made to be used by any other component; it is exclusive to the modal window. Keeping the css inline also provides context to the components. CSS styling sheets have a tendency of becoming a bit of an unorganized styling graveyard where it becomes difficult to pinpoint what styling is associated with what html. Keeping the code here makes it easy to track and update. 

On the other hand, massive changes to CSS across loris could benefit from having everything in a well organised style sheet. Having a bunch of css styling nested in different components could make the task of changing many things at once difficult.

I see benefits and disadvantages to both, but neither one is 'right' or 'wrong'. It would be nice to have a meeting to discuss certain standards (such as whether LORIS uses inline styling or not) so that there can be a precedent from which to review future code. We can then add this to the CodingStandards.md in https://github.com/aces/Loris/tree/master/docs, or create an entirely new file, for reference.

jsx/Modal.js
davidblader: there may be cases where the CSS described here might look off in certain contexts or on certain displays. i wouldn't advocate for people to arbitrarily change the style but you never know when that extra flexibility might be needed

HenriRabalais: Hmm, okay. If we are moving the styling to the css style sheet, do you suggest I simply allow the possibility of passing a title className to to this component?

davidblader: yep that sounds good to me

jsx/Modal.js
```diff
@@ -0,0 +1,137 @@
...
+ * Modal Component.
+ * React wrapper for a Modal Window. Allows to dynamically toggle a Modal window.
+ *
+ * ================================================
+ * Usage:
+ * - Wrap the contents to be displayed by the Modal Window by the Modal Component.
+ * - Use the 'show' prop to toggle the Modal Component's presentation.
+ * - Use the 'title' prop to set a title for the Modal Component.
+ * - Use the 'onSubmit' prop set a submission function for the Modal's contents.
+ * - Use the 'throwWarning' prop to throw a warning upon closure of the Modal Window
```
zaliqarosli: looking to test. could you possibly add the usage for the closeModal prop? and could adding a default for that prop be useful?

HenriRabalais: For sure! I'll add it in there. Basically the closeModal props is the function that is triggered when the modal is closed. This should a be a function within the parent which changes a state so that the value passed as `show` becomes false. It's a required function, so I don't think it should have a default per se, but I could throw a console warning if it is not set. I had been planning on adding an implementation of the modal in loris somewhere so it would be easier to test. I can do that in my next commit.

zaliqarosli: what if there is no function that should be triggered i.e. if I click close and all i want is the modal to close and display the window behind it? I'm finding it hard to understand what this required closeModal function should be

HenriRabalais: If there is no function that is triggered, then the modal will not closed. The value of the prop `show` needs to go from `true` to `false` for the modal window is disappear. The function would look something like this.

```
closeModal() {
  let showModal = this.state.showModal;
  showModal = false;
  this.setState({showModal});
}
```

`this.state.showModal` would then be passed to `Modal` as the prop `show`.

modules/media/js/editFormIndex.js
```diff
@@ -1,2 +1,2 @@
...
...utable){var progressbar=$("#progressbar"),progresslabel=$("#progresslabel"),percent=Math.round(evt.loaded/evt.total*100);$(progressbar).width(percent+"%"),$(progresslabel).html(percent+"%"),progressbar.attr("aria-valuenow",percent)}},!1),xhr},success:function(data){$("#file-progress").addClass("hide"),self.setState({uploadResult:"success"}),self.showAlertMessage()},error:function(err){console.error(err),self.setState({uploadResult:"error"}),self.showAlertMessage()}})}},{key:"setFormData",value:function(formElement,value){var formData=this.state.formData;""===value?formData[formElement]=null:formData[formElement]=value,this.setState({formData:formData})}},{key:"showAlertMessage",value:function(){var self=this;if(null!==this.refs["alert-message"]){var alertMsg=this.refs["alert-message"];$(alertMsg).fadeTo(2e3,500).delay(3e3).slideUp(500,function(){self.setState({uploadResult:null})})}}}]),MediaEditForm}(React.Component);MediaEditForm.propTypes={DataURL:React.PropTypes.string.isRequired,action:React.PropTypes.string.isRequired};var RMediaEditForm=React.createFactory(MediaEditForm);window.MediaEditForm=MediaEditForm,window.RMediaEditForm=RMediaEditForm,exports.default=MediaEditForm}});
```
zaliqarosli: jsx/editFormIndex.js was deleted, so this should be too

modules/media/jsx/columnFormatter.js
```diff
@@ -1,72 +0,0 @@
-import Modal from 'Modal';
```
zaliqarosli: where was columnFormatter moved to?

HenriRabalais:
@zaliqarosli 
1. Yeah, that shouldn't be a problem.
2. That's one of the drawbacks of having the modal window in control of it's own closing and opening. The parent component can't tell it that the form has been submitted and therefore needs to close. A `promise` would need to be passed to the `onSubmit` which, if fulfilled, would trigger the closing of the modal window.

modules/media/js/editFormIndex.js
HenriRabalais: done!

johnsaigle:
@HenriRabalais Is there a specific module here that I can visit to test this PR? #3957 (from your post) has been merged but is this branch rebased to include it?

HenriRabalais:
Yeah, this should be tested on the media module. I will update testing instructions in the PR description.

jsx/Modal.js
```diff
@@ -0,0 +1,151 @@
...
+      open: false
+    };
+    this.onClose = this.onClose.bind(this);
+  }
+
+  onClose() {
+    if (this.props.throwWarning) {
+      swal({
+        title: 'Are You Sure?',
+        text: 'Leaving the form will result in the loss of any information' +
```
zaliqarosli: need a space between information and entered

modules/media/jsx/mediaIndex.js
```diff
@@ -127,8 +129,18 @@ class MediaIndex extends React.Component {
     }
 
     if (column === 'Edit Metadata') {
-      const editURL = loris.BaseURL + "/media/edit/?id=" + row['Edit Metadata'];
-      return <td className={classes}><a href={editURL}>Edit</a></td>;
+      let editButton = (
+        <Modal
+          title="Edit Media File"
+          trigger={<button>Edit</button>}
```
zaliqarosli: i think maybe the js file needs to be recompiled? the modal component here doesn't specify a throwWarning props but the form is still throwing the error modal window

johnsaigle:
@zaliqarosli Can you update your review here?

zaliqarosli:
@johnsaigle after some discussions, this PR still needs some work to be done on it

HenriRabalais:
@zaliqarosli done.

HenriRabalais:
Closed in favour of #4252

PR #3985, opened 2018-10-04
[Meta] Removing cccccc_pixel.gif

Jkat:
### Brief summary of changes

I don't like this fake gif

### To test this change...

- [ ] agree with me


driusan:
Can you switch this to major to ensure if there's anywhere referencing it it gets caught in a testing round?

Jkat:
I will make a new PR that deletes some other unused images

PR #3987, opened 2018-10-05
[Config / DICOM Archive / Imaging]Change default phantom regex config

davidblader:
### Brief summary of changes
This PR changes the default regex configurations to not contain forward slashes and replaces them with equivalent expressions. 

### This resolves issue...
LORIS-MRI would like to start using the configuration options made available in the Config module,
however the current default values are incompatible with Perl

### To test this change...

- [ ] If your dev environment uses the default regex values in Config, just run the patch. Then click some of the `View Details` links in Imaging Uploader to be taken to DICOM Archive and ensure that the PatientName & PatientID fields are still viewable for different types of scans (both patients and lego / living phantoms)

- [ ] If your dev environment has custom regexs for these options, your project may have to create an equivalent regex without using forward slashes

### DIFF
--- a/SQL/New_patches/2018-10-05_change_regex_config.sql
+++ b/SQL/New_patches/2018-10-05_change_regex_config.sql
@@ -1,19 +1,47 @@
+-- Convert all Perl regex that contain the case-insensitive modifier 'i'
 UPDATE Config
-SET Value = "."
+SET Value=CONCAT('(?i)', SUBSTRING(Value FROM 2 FOR LENGTH(Value)-3))
 WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='patientIDRegex')
-AND Value='/./i';
+AND Value LIKE '/%/i';
 
+-- Removes the leading and trailing forward slashes from every regex that do not contain any modifier
 UPDATE Config
-SET Value = "."
+SET Value=SUBSTRING(Value FROM 2 FOR LENGTH(Value)-2)
+WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='patientIDRegex')
+AND Value LIKE '/%/';
+
+-- Convert all Perl regex that contain the case-insensitive modifier 'i'
+UPDATE Config
+SET Value=CONCAT('(?i)', SUBSTRING(Value FROM 2 FOR LENGTH(Value)-3))
 WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='patientNameRegex')
-AND Value='/./i';
+AND Value LIKE '/%/i';
 
+-- Removes the leading and trailing forward slashes from every regex that do not contain any modifier
 UPDATE Config
-SET Value = "(?i)phantom"
+SET Value=SUBSTRING(Value FROM 2 FOR LENGTH(Value)-2)
+WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='patientNameRegex')
+AND Value LIKE '/%/';
+
+-- Convert all Perl regex that contain the case-insensitive modifier 'i'
+UPDATE Config
+SET Value=CONCAT('(?i)', SUBSTRING(Value FROM 2 FOR LENGTH(Value)-3))
 WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='LegoPhantomRegex')
-AND Value='/phantom/i';
+AND Value LIKE '/%/i';
 
+-- Removes the leading and trailing forward slashes from every regex that do not contain any modifier
 UPDATE Config
-SET Value = "(?i)phantom"
+SET Value=SUBSTRING(Value FROM 2 FOR LENGTH(Value)-2)
+WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='LegoPhantomRegex')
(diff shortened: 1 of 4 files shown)


modules/dicom_archive/php/dicomarchiveanonymizer.class.inc
```diff
@@ -53,13 +53,22 @@ class DICOMArchiveAnonymizer implements Mapper
         $newrow = json_decode($resource->toJSON(), true);
         $cid    = $resource->getCenterID();
 
-        if (!preg_match($config['patientNameRegex'], $newrow['Patient Name'])
-            && !preg_match($config['LegoPhantomRegex'], $newrow['Patient Name'])
-            && !preg_match($config['LivingPhantomRegex'], $newrow['Patient Name'])
+        if (!preg_match(
+            "/{$config['patientNameRegex']}/",
```
driusan: Bad things can happen here (and other places) if the config value has a backslash. It needs to be escaped.

davidblader: sure, but would you mind elaborating? if you escape backslashes then this would disallow for any reserved regex tokens to be used wouldn't it?

driusan: If someone enters a backslash anywhere in the config value then this will get directly interpolated based on the user data. ie if I enter the value `/data/` the value passed to preg_match will be `//data//' The matching part of the regex is then "//", with "data//" as the modifier, which wasn't the intended behaviour. I don't think a malicious user can do anything worse with this than they previously could (and since it comes from an admin in the config module is probably relatively safe), but this could generate a lot of false positive matches and/or PHP warnings.

SQL/New_patches/2018-10-05_change_regex_config.sql
```diff
@@ -0,0 +1,19 @@
+UPDATE Config
```
nicolasbrossard: I would issue:

```sql
-- Convert all Perl regex that contain the case-insensitive modifier 'i'
UPDATE Config
SET Value=CONCAT('(?i)', SUBSTRING(Value FROM 2 FOR LENGTH(Value)-3))
WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='patientIDRegex')
AND Value LIKE '/%/i';

-- Removes the leading and trailing forward slashes from every regex that do not contain any modifier
UPDATE Config
SET Value=SUBSTRING(Value FROM 2 FOR LENGTH(Value)-2)
WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='patientIDRegex')
AND Value LIKE '/%/';
```

This will update the default values _and_ all user defined regexes for config setting `patientIDRegex`.

SQL/New_patches/2018-10-05_change_regex_config.sql
```diff
@@ -0,0 +1,19 @@
...
+SET Value = "."
+WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='patientIDRegex')
+AND Value='/./i';
+
+UPDATE Config
```
nicolasbrossard: Same comment as the one for `patientIDRegex`.

SQL/New_patches/2018-10-05_change_regex_config.sql
```diff
@@ -0,0 +1,19 @@
...
+SET Value = "."
+WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='patientNameRegex')
+AND Value='/./i';
+
+UPDATE Config
```
nicolasbrossard: Same comment as the one for `patientIDRegex`.

SQL/New_patches/2018-10-05_change_regex_config.sql
```diff
@@ -0,0 +1,19 @@
...
+SET Value = "(?i)phantom"
+WHERE ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='LegoPhantomRegex')
+AND Value='/phantom/i';
+
+UPDATE Config
```
nicolasbrossard: Same comment as the one for `patientIDRegex`.

PapillonMcGill:
removed `Passed Manual Tests` as the SQL code change is intensive. Should be Ok but would need a retest.

If @nicolasbrossard is satisfied by the latest change, I approve this PR

modules/dicom_archive/php/viewdetails.class.inc
```diff
@@ -148,14 +148,36 @@ class ViewDetails extends \NDB_Form
...
+            '/\//',
+            '\/',
+            $dicomArchiveSettings['LivingPhantomRegex']
+        );
+        $pIDRegex        = preg_replace(
+            '/\//',
+            '\/',
+            $dicomArchiveSettings['patientIDRegex']
+        );
+
```
nicolasbrossard: The `preg_match` tests below do not seem to take into account whether the archive really *is* of phantom scans or not. I agree that that's the best we can do for projects that do not use the imaging uploader but for those that do, we can look at table `mri_upload` in order to know this. This way, we can perform the test that is appropriate to the type of archive considered.

PapillonMcGill:
@cmadjar could you retest?

PR #3924, opened 2018-09-05
[Examiner] Reactify index page

zaliqarosli:
### Brief summary of changes
Reactification of `Examiner` index page

~~This PR uses a new feature (Checkbox) in `Form.js` included in https://github.com/aces/Loris/pull/3918 which needs to be merged. I will add a `Blocked` label for the mean time. Once that PR is merged, I will rebase.~~

Edit: https://github.com/aces/Loris/pull/3918 merged, this PR has been rebased and is good to go!

### To test this change...

- [ ] Try filtering and adding an examiner!

PR #3990, opened 2018-10-09
[Examiner] Reactify Edit Examiner page

zaliqarosli:
### Brief summary of changes
This PR reactifies the `edit_examiner` page of the Examiner module (replaces https://github.com/aces/Loris/pull/3911). The reactification of the index page can be found here https://github.com/aces/Loris/pull/3924.

### To test this change...

- [ ] Check out this branch
- [ ] Navigate to an examiner within the module
- [ ] Test functionality (everything should work the same, if not, better)
- [ ] Test display (everything should look the same minus the lengths of the date and comment fields as Form.js elements have a default label which, in this case, is not required)


modules/examiner/jsx/editExaminer.js
```diff
@@ -0,0 +1,333 @@
...
+
+  /**
+   * Store the value of the element in this.state.formData
+   *
+   * @param {string} formElement  name of the form element
+   * @param {string} value        value of the form element
+   */
+  setFormData(formElement, value) {
+    let formData = this.state.formData;
+    let split = formElement.split("_");
```
davidblader: this seems weird. a comment explaining what key & element are would be nice

modules/examiner/jsx/editExaminer.js
```diff
@@ -0,0 +1,333 @@
...
+      },
+      error: error => {
+        console.error(error);
+        let message = error.responseText;
+        swal('Error!', message, 'error');
+      }
+    });
+  }
+
+  render() {
```
davidblader: this render function has a lot in it. i'd suggest getting some of the subcomponents from functions to make it a bit cleaner

modules/examiner/jsx/editExaminer.js
```diff
@@ -0,0 +1,355 @@
...
+
+  componentDidMount() {
+    this.fetchData();
+  }
+
+  /**
+   * Retrieve data from the provided URL and save it in state
+   */
+  fetchData() {
+    const self = this;
```
HenriRabalais: I don't believe there's a need for this line since this should be bound to the ajax call success clause via the arrow function.
```suggestion
```

modules/examiner/jsx/editExaminer.js
```diff
@@ -0,0 +1,355 @@
...
+    let formData = {};
```
HenriRabalais: Perhaps place this on line `62` before the if statement to keep it with relevant code.

modules/examiner/jsx/editExaminer.js
```diff
@@ -0,0 +1,355 @@
...
+   * @param {string} value        value of the form element
+   */
+  setFormData(formElement, value) {
+    let formData = this.state.formData;
+    // Form elements have format name="instrumentID_fieldName"
+    // We want to extract that information to be used as keys
+    // in the nested formData array
+    let splitNameOfElement = formElement.split("_");
+    let instrumentID = splitNameOfElement[0];
+    let fieldName = splitNameOfElement[1];
```
HenriRabalais: I'm pretty sure variables on lines 94-96 can be declared as `const`.

EDIT: After looking up `let` and `const` some more we really should be using `const` in more places - even for arrays that have values within them changed. I don't want to nit-pick about what should and shouldn't be a `const`, but consider reviewing the code and changing all variables that don't get reassigned to `const`

modules/examiner/jsx/editExaminer.js
```diff
@@ -0,0 +1,355 @@
...
+        inputElements.push(instrumentRow);
+      }
+    }
+    return (
+      <FormElement
+        Module="examiner"
+        name="edit_cert"
+        id="edit_cert_form"
+        onSubmit={this.handleSubmit}
+        ref="editCert"
```
HenriRabalais: string refs ☝️

modules/examiner/php/editcertification.class.inc
```diff
@@ -0,0 +1,277 @@
...
+    /**
+     * Checks if the user has access to the edit examiner form
+     *
+     * @return bool
+     */
+    function _hasAccess()
+    {
+        $user = \User::singleton();
+        return $user->hasPermission('examiner_view')
+            && $user->hasPermission('examiner_multisite');
```
johnsaigle: Should this maybe be an OR permission? I think `multisite` implies that you have `view`. It also means that a user has to be able to edit ALL sites, or none at all

modules/examiner/php/editcertification.class.inc
```diff
@@ -0,0 +1,277 @@
...
+
+        // Ensure it's a POST request.
+        if ($request->getMethod() != "POST") {
+            return (new \LORIS\Http\Response())
+                ->withHeader("Content-Type", "text/plain")
+                ->withStatus(405)
+                ->withHeader("Allow", "POST")
+                ->withBody(
+                    new \LORIS\Http\StringStream(
+                        "Only POST requests are allowed"
```
johnsaigle:
```suggestion
                        "Only POST requests are allowed."
```
Adding a period

zaliqarosli:
@johnsaigle @HenriRabalais i believe your approving reviews got dismissed by my last commit to satisfy travis. please re-approve if everything is still good, when you get a chance. thanks!

johnsaigle:
I'm not exactly sure how this module is typically meant to function as I've not really used it. 

I am getting a lot of Notices and Warnings from the PHP code though and unable to sort the columns. This was also true in the old version though.

If you know of any specifics about how to test it, let me know!

zaliqarosli:
needs to be moved to major and refactored to latest react template

ridz1208:
@zaliqarosli if it needs refactoring can this be closed and a new PR opened ?

PR #3997, opened 2018-10-11
Add document describing new routing

driusan:
This adds a document to the docs directory explicitly telling developers not to depend on the `module/*/ajax` directory and puts in place a (commented out, until more modules are updated) deprecation warnings in AjaxHelper.php.

The document explains why the deprecation is happening and points to an example PR of how to update an ajax script to the new architecture. It can be updated to be more helpful as people have more experience doing so or find common patterns, but for now this is better than nothing.

### DIFF
--- a/docs/Routing.md
+++ b/docs/Routing.md
@@ -1,0 +1,25 @@
+# Routing
+
+Since LORIS 20.0, LORIS has included a router and middleware based
+on (PSR15)[https://www.php-fig.org/psr/psr-15/] This has a number
+of benefits over the custom scripts in the htdocs directory used
+by previous LORIS releases such as reducing the dependency on Apache,
+removing the dependency on mod_rewrite, and making it possible for
+Modules to declare their own custom routes with complex path
+components or use the same access checking mechanisms as the rest
+of LORIS automatically.
+
+The AjaxHelper.php and associated module/*/ajax/*.php scripts are
+being deprecated as a result.
+
+In the new architecture, modules should implement an endpoint which
+handles the ServerRequestInterface passed to it in a RESTful manor.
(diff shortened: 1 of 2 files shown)


docs/Routing.md
```diff
@@ -0,0 +1,25 @@
...
+Since LORIS 20.0, LORIS has included a router and middleware based
+on (PSR15)[https://www.php-fig.org/psr/psr-15/] This has a number
+of benefits over the custom scripts in the htdocs directory used
+by previous LORIS releases such as reducing the dependency on Apache,
+removing the dependency on mod_rewrite, and making it possible for
+Modules to declare their own custom routes with complex path
+components or use the same access checking mechanisms as the rest
+of LORIS automatically.
+
+The AjaxHelper.php and associated module/*/ajax/*.php scripts are
```
PapillonMcGill: ajax need markup adjustment

driusan: what do you mean?

kongtiaowang:
There is a phpcs issue.

docs/Routing.md
```diff
@@ -0,0 +1,25 @@
...
+removing the dependency on mod_rewrite, and making it possible for
+modules to declare their own custom routes with complex path
+components or use the same access checking mechanisms as the rest
+of LORIS automatically.
+
+The AjaxHelper.php and associated `module/*/ajax/*.php` scripts are
+being deprecated as a result.
+
+In the new architecture, modules should implement an endpoint which
+handles the ServerRequestInterface passed to it in a RESTful manor.
```
johnsaigle: manor --> manner

driusan: thanks, fixed

PR #4000, opened 2018-10-15
[Issue Tracker] Improving README formatting and grammar

Jkat:
### Brief summary of changes

Improving README formatting and grammar

### To test this change...

- [ ] Check out the formatting and as a bonus, find the grammar mistake

### DIFF
--- a/modules/issue_tracker/README.md
+++ b/modules/issue_tracker/README.md
@@ -4,9 +4,9 @@
 The Issues Module allows users to track issues they have with data, or with their LORIS instance itself. A form with pre-defined fields is provided for users to submit issues, and a filter-form gives a sortable and filterable table view of issues viewable by the user.
 
 ## Permissions
-- `issue_tracker_reporter` permission allow adding an issue, editing issues created by the user, and comment on all issues for the site. 
-- `issue_tracker_developer` permission allow to do the same, as well as closing an issue or editing any field of a submitted issue for the site. 
-- `view_all_sites` permission allow user to view issues relevant to data from others sites. 
+- `issue_tracker_reporter` permission allows adding an issue, editing issues created by the user, and commenting on all issues for the site. 
+- `issue_tracker_developer` permission allows to do the same, as well as closing an issue or editing any field of a submitted issue for the site. 
+- `view_all_sites` permission allows a user to view issues relevant to data from other sites. 
 
 Most of the permissions are controlled in `IssueForm.js`, dependent on values returned in `editIssue.php`.
 
--- a/modules/issue_tracker/README.md
+++ b/modules/issue_tracker/README.md
@@ -1,10 +1,10 @@
-## Issue Tracker Readme
+# Issue Tracker
 
-#### Overview
+## Purpose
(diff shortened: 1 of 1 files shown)


modules/issue_tracker/README.md
```diff
@@ -1,10 +1,10 @@
...
+## Permissions
+Permissions in this module are complex and are not yet in their final iteration. Two new permissions were created, `issue_tracker_reporter` and `issue_tracker_developer`. The `view_all_sites` permission was also integrated to prevent users from viewing issues relevant to data they do not have permission to see. A reporter can add an issue, edit their own issue, and comment on all issues. A developer can do the same, plus they can close an issue and edit all fields of a submitted issue. Most of the permissions are controlled in `IssueForm.js`, dependent on values returned in `editIssue.php`.
```
PapillonMcGill: Not sure it is a good idea to say "... are complex and not yet in their final iteration". => Better not use it for now or don't open this can of worms might come to user mind.

So, the "view_all_sites" allow you to view what your are not supposed to? Maybe say that this permission is for users who need to see all issues even those related to data they can't see. But wait, if the text of the issue have include some of the data, that would equate to data leak?

Are there user who are developper? (I mean in a production environnement.)

Line about the code (js and PHP files) is irrelevant to end users

modules/issue_tracker/README.md
```diff
@@ -1,10 +1,10 @@
...
 
-#### Future Development
-Future development includes the ability to edit one’s own comments after submission, a pop-up or sliding view within other modules for easy issue reporting, further integration with the dashboard and with statistics, refinement of the form UI, a tabular sortable view of comments, the ability to click watching within the filter-form table and the inclusion of default users to notify in the config settings.
\ No newline at end of file
+## Future Development
+Future development includes the ability to edit one’s own comments after submission, a pop-up or sliding view within other modules for easy issue reporting, further integration with the dashboard and with statistics, refinement of the form UI, a tabular sortable view of comments, the ability to click watching within the filter-form table and the inclusion of default users to notify in the config settings.
```
PapillonMcGill: Might be a little speculative to talk about the future and is not relevant to the present version of Loris.

PapillonMcGill: in general

Jkat: Do you have a suggestion on where to put this section instead?

PapillonMcGill: I would just remove it. Update to this document could be done as each of these projected feature get included in the code.

Do we have a wishlist.md file somewhere?

modules/issue_tracker/README.md
PapillonMcGill:
```suggestion
Two new permissions were created, `issue_tracker_reporter` and `issue_tracker_developer`. The `view_all_sites` permission was also integrated allow *** WHAT TYPE OF USER EXACTLY FOR HERE?*** viewing issues relevant to data from others sites. A reporter can add an issue, edit their own issue, and comment on all issues. A developer can do the same, plus they can close an issue and edit all fields of a submitted issue. Most of the permissions are controlled in `IssueForm.js`, dependent on values returned in `editIssue.php`.
```

PapillonMcGill:
waiting for Travis

PR #4001, opened 2018-10-15
[Docs] Move templates to github folder

johnsaigle:
### Brief summary of changes

Move files to hidden location for tidiness.


driusan:
Should CONTRIBUTING.md be moved? It's not (really) GitHub specific and seems like it's better for it to be discoverable easily rather than hidden.

PR #4002, opened 2018-10-15
Add `make check` target

driusan:
This adds a "make check" target to perform basic unittests and
does some minor refactoring of the makefile

The target "make checkstatic" and "make unittests" are also
added to make it easier to quickly run a smoke test.

### DIFF
--- a/php/libraries/UserPermissions.class.inc
+++ b/php/libraries/UserPermissions.class.inc
@@ -51,7 +51,7 @@ class UserPermissions
     function select($username)
     {
         // create DB object
-        $DB =& Database::singleton();
+        $DB = \Database::singleton();
 
         // get the proper ID from the database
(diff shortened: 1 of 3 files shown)


driusan:
@PapillonMcGill Can you re-review?

php/libraries/UserPermissions.class.inc
```diff
@@ -51,7 +51,7 @@ class UserPermissions
     function select($username)
     {
         // create DB object
-        $DB =& Database::singleton();
+        $DB = \Database::singleton();
```
johnsaigle: Not relevant to this PR but also.... this is nice.

driusan: Yeah, it was throwing a database exception on that line and I thought that might be the cause..

PR #4005, opened 2018-10-16
[Battery Manager] New Module - HACKTOBERFEST 2018 🎃

HenriRabalais:
**This PR was originally made by @victoriafoing but is now being maintained by @HenriRabalais. I have made a new PR for the sake of clarity**

### Brief summary of changes

This PR creates a new module 'Battery Manager' which allows users to browse, activate/deactivate , edit, and add entries in the `test_battery` table.

This module includes the following components:
- A **Browse** tab for searching and deactivating entries in the Test Battery
- A **Change Status** column with buttons for directly activating and deactivating entries in the table
- A **Edit Metadata** column with a link that takes you to an **Edit** window for deactivating an entry and adding a new version of it to the table
- An **Add** tab for adding and activating entries in the Test Battery
- AJAX calls for checking for duplicate entries, adding entries, editing entries, deactivating/activating entries, and getting data for dropdowns 
- SQL patches for permissions and LorisMenu
- Documentation (README, TestPlan, Help)

Allowing users (e.g. study administrators) to add to and update the `test_battery` table through a front-end interface.

### This resolves issue...

Of having to add to and update the `test_battery` table through the back-end.

### To test this change...

Check out the instructions in the Test Plan

**Note**: The Test Battery table has a lot of NULL values and uses FilterForm.js, which does not properly filter NULL values. Another PR will have to be made to update FilterForm.js.


modules/battery_manager/ajax/change_active_status.php
```diff
@@ -0,0 +1,94 @@
...
+ * @author   Victoria Foing <victoria.foing@mcin.ca>
+ * @license  Loris license
+ * @link     https://www.github.com/aces/Loris
+ */
+
+// Determine whether entry should be activated or deactivated
+if (sanitize('action', 'get') !== null) {
+    $action = sanitize('action', 'get');
+    if ($action === "activate") {
+        echo changeStatus('Y');
```
davidblader: travis seems to be complaining about the use of `echo` here, seemingly due to the `@return` of the function doc of changeStatus, maybe change it to `@return string`?

modules/battery_manager/ajax/get_form_data.php
```diff
@@ -0,0 +1,146 @@
...
+        echo getAddFormData();
+    } else if ($form === "edit") {
+        echo getEditFormData();
+    }
+}
+
+/**
+ * Get form data for Add form
+ *
+ * @return json object
```
davidblader: nitpicky but technically is returning a string

modules/battery_manager/ajax/get_form_data.php
```diff
@@ -0,0 +1,146 @@
...
+{
+    $addFormData = getFormData();
+    return json_encode($addFormData);
+}
+
+/**
+ * Get form data for Edit form
+ * Add entry data in json form
+ *
+ * @return json object
```
davidblader: same as above

modules/battery_manager/jsx/addEntry.js
```diff
@@ -0,0 +1,352 @@
...
+        self.setState({
+          error: 'An error occurred when loading the form!'
+        });
+      }
+    });
+  }
+
+  render() {
+    // Data loading error
+    if (this.state.error !== undefined) {
```
davidblader: might be a better idea to just do `if (!this.state.error) {`

modules/battery_manager/jsx/batteryManagerIndex.js
```diff
@@ -0,0 +1,252 @@
...
+      }
+    });
+  }
+
+  updateFilter(filter) {
+    this.setState({filter});
+  }
+
+  resetFilters() {
+    this.refs.batteryManagerFilter.clearFilter();
```
davidblader: i'm not sure this is set anywhere

HenriRabalais:
@driusan 
Seems like this needs a major refactor before resurfacing for review. Taking @davidblader suggestions into consideration and closing PR.

modules/battery_manager/jsx/addEntry.js
HenriRabalais: or rather `if (this.state.error) {` i think

modules/battery_manager/jsx/batteryManagerIndex.js
HenriRabalais: okie doke

driusan:
@HenriRabalais The business logic of the scripts can likely be re-used, so it might not be as much work as it seems.

HenriRabalais:
@driusan No, it shouldn't be too bad. But I figure at this point I might as well introduce the dataframework and give the react an overhaul.

PR #4014, opened 2018-10-18
[Instrument Manager] Mitigate SQL injection vector

johnsaigle:
### Brief summary of changes

This is a copy of #3674 released under my name to make organization simpler.

### This resolves issue...

See earlier PR.

### To test this change...

- [ ] Contact me for instructions.


johnsaigle:
The fix mitigates the exploit I designed. This PR just needs a bit of cleanup to pass linters etc.

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,224 @@
...
-                $filename="../project/tables_sql/".$bits[1].".sql";
-                $output="CREATE TABLE `$bits[1]` (\n";
-                $output.="`CommentID` varchar(255) NOT NULL default '',
-                          `UserID` varchar(255) default NULL,
-                          `Examiner` varchar(255) default NULL,
-                          `Testdate` timestamp DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
-                          `Data_entry_completion_status` enum('Incomplete','Complete') NOT NULL default 'Incomplete',\n";
-            break;
+                $sql_query_table_name = str_replace('`', '', $db->escape($bits[1]));
+                $file_name = "../project/tables_sql/".$bits[1].".sql";
```
PapillonMcGill:
```suggestion
                $file_name = "../project/tables_sql/".$sql_query_table_name.".sql";
```
If backtick are present in $bits[1], this might pose problem with persons that are not bash expert.

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,224 @@
...
+                $output    = "CREATE TABLE `$bits[1]` (\n";
```
PapillonMcGill:
```suggestion
                $output    = "CREATE TABLE `$bits[1]` (\n";
```
same here, that would be 2 backtick back to back

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,224 @@
...
+                    }
+                } else {
+                    // Safety SQL injection prevention.
+                    $bits[0] = str_replace('`', '', $db->escape($bits[0]));
                 }
-
-                $bits[2]=htmlspecialchars($bits[2]);
-                $output.="`$bits[1]` $bits[0] default NULL,\n";
+                $bits[2]     = htmlspecialchars($bits[2]);
+                $output     .= "`$bits[1]` $bits[0] default NULL,\n";
```
PapillonMcGill: I would also use sanitized version here. You don't want to create SQL files that would pose a risk.

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,224 @@
...
+        } else {
+            $error_message = "Instrument(s) file uses existing name.";
+        }
+        $statement = $db->prepare(
+            'REPLACE INTO test_names (Test_name, Full_name, Sub_group)'
+            .' VALUES (?, ?, 1)'
+        );
+        $statement->bindParam(1, $sql_query_table_name, PDO::PARAM_STR);
+        $statement->bindParam(2, $title, PDO::PARAM_STR);
+        if ($execute_mysql_query) {
```
PapillonMcGill: I would move this test (and the one on line 193) to line 174 ( or maybe 163) as there is no need to create all these statement if they will never be executed

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,224 @@
...
+ * Creates a partial MySQL enum statement based on options extracted from
+ * a LINST file (which is in turn based on an HTML select field).
+ *
+ * @param array  $options the enum values.
+ * @param string $table   the table name used.
+ * @param string $name    the enum name used.
+ *
+ * @return string
+ */
+function enumizeOptions($options, $table, $name)
```
PapillonMcGill: The $table and $name parameter are not used anymore. I would remove them from the function def. (and adjust call to this function

PapillonMcGill: Not use in this function (enumizeOptions).

PapillonMcGill:
awaiting Travis approval

PapillonMcGill:
Travis passed even if showing not pass, ran it again

modules/instrument_builder/jsx/react.questions.js
```diff
@@ -622,10 +622,16 @@ var AddElement = React.createClass({
         hasError = true;
       }
 
-      if (minDate > maxDate && min !== '' && max !== '') {
+      if (min === '' || max === '') {
```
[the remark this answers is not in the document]
johnsaigle: Okay I'll remove it. This is from @intralizee's original PR so I'm not sure of the context.

johnsaigle:
@driusan To address your above comments:

>This appears to be completely removing the ability of the instrument manager to fully install instruments

A user is still able to install instruments via the front-end but it looks like @intralizee decided to offload this functionality to the `tools/` script instead.

>It used to read an ipoutput.txt 

I didn't know about this use case. How is this generated? I'll test this when I have it.

>or linst file from stdin and generate an SQL file based on it.
The script can still do this. @intralizee has changed the syntax to
```php
php generate_tables_and_test_names.php ../project/instruments/inst.linst
```
from
```php
php generate_tables_and_test_names.php < ../project/instruments/inst.linst
```
i.e. it takes an argument rather than input from STDIN. This isn't crucial and I can revert this if you want it the old way for compatibility.

> This doesn't fix SQL injection
It is currently possible to do SQLi on the instrument_manager and this PR does fix this issue. Please see the Redmine tickets in the original PR #3674. 

> It's valid to not have a min or max. [React dates]

One of these tickets (12174) also reports the blank date issue. If you don't want the front-end React fix could you suggest another way to fix this problem?

ridz1208:
@johnsaigle @driusan I like the 2 options in the script (we usually generate SQL and review it before we run it on the databse) it goes with all the confirm/tosql options in other scripts such as delete candidate.

I'm not a big fan of using "false" as the flag, I think we should stick with `confirm` or `tosql` explicitely with only a helper function displayed if neither are given. Thats my personal pref and it could be rectified in a later PR to not delay this one

Given the change of behaviour of the stdin vs parameter passing, this might require a caveat for existing projects flag

ridz1208:
given that te wiki might need to be altered I added the document at release

johnsaigle:
Note: #4142 may have an impact on this issue. The vector is still likely to exist in the new Reactified code as the issue is caused primarily by the `tools/` script. Further testing is needed.

johnsaigle:
@driusan Could you comment on the above points when you have a moment? We should really get this fixed.

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,213 @@
...
+ * from the /instrument_manager  module and installs an instrument for the table of each
+ * instrument it finds in the {instrument_name}.linst file.
+ * These sql files are output to the tables_sql/ subdirectory.
+ *
+ * Ex cmd with SQL execution:
+ * php generate_tables_sql_and_testNames.php ../instruments/[instrument_name].linst
+ *
+ * Ex cmd without SQL execution:
+ * php generate_tables_sql_and_testNames.php
+ * ../instruments/[instrument_name].linst false
```
johnsaigle: Change this to be consistent with other scripts:
* Should just print usage by default
* --to-sql flag
* --apply flag

johnsaigle:
@driusan I believe this is ready for re-review.

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,234 @@
...
+                $output    = "CREATE TABLE `$sql_query_table_name` (\n";
+                $output   .= "`CommentID` varchar(255) NOT NULL default '',\n"
+                    ."`UserID` varchar(255) default NULL,\n"
+                    ."`Examiner` varchar(255) default NULL,\n"
+                    ."`Testdate` timestamp DEFAULT CURRENT_TIMESTAMP ON "
+                    ."UPDATE CURRENT_TIMESTAMP,\n"
+                    ."`Data_entry_completion_status` "
+                    ."enum('Incomplete','Complete') "
+                    ."NOT NULL default 'Incomplete',\n";
+                $sql_query_statement = "CREATE TABLE `$sql_query_table_name` (\n";
```
ridz1208: why are we duplication `$output` in `$sql_query_statement` why not use one to generate the other instead

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,234 @@
...
+                    $bits[0] = "varchar(255)";
+                } else {
+                    // Sanitize $bits to mitigate SQL injection attacks.
+                    $bits[0] = str_replace('`', '', $db->escape($bits[0]));
                 }
-
-                $bits[2]=htmlspecialchars($bits[2]);
-                $output.="`$bits[1]` $bits[0] default NULL,\n";
+                $bits[2]     = htmlspecialchars($bits[2]);
+                $output     .= "`$sql_query_table_name` $bits[0] default NULL,\n";
```
ridz1208: bits[1] here is not table name

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,234 @@
...
+                    $fp,
+                    "INSERT INTO instrument_subtests "
+                    ."(Test_name, Subtest_name, Description)"
+                    ." VALUES ('$sql_query_table_name', '"
+                    .$sql_query_table_name."_page".($page_number+1)
+                    ."', '$page');\n"
+                );
+            }
+        }
+        // Writes the error message to the sql file.
```
ridz1208: but why ??

error message should just be printed and file discarded. I dont think the sql file should be used as a log

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,234 @@
...
+if (!is_array($instruments)) {
+    $error_message = "Instrument file could not be processed.";
+} else {
+    $pages = array();
+    $sql_query_statement  = "";
+    $sql_query_table_name = "";
+    $sql_query_columns    = array();
+    // Setup Database object.
+    $factory   = \NDB_Factory::singleton();
+    $db        = $factory->database();
```
ridz1208: generic includes should already give you a database

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,234 @@
...
-    $output.="PRIMARY KEY  (`CommentID`)\n) ENGINE=InnoDB DEFAULT CHARSET=utf8;\n";
-    print "Filename: $filename\n";
-    $fp=fopen($filename, "w");
-    fwrite($fp, $output);
-    fwrite($fp, "REPLACE INTO test_names (Test_name, Full_name, Sub_group) VALUES ('$tablename', '$title', 1);\n");
-    foreach($pages as $pageNo => $page) {
-        fwrite($fp, "INSERT INTO instrument_subtests (Test_name, Subtest_name, Description) VALUES ('$tablename', '" . $tablename  . "_page" . ($pageNo+1) . "', '$page');\n");
-
+    // Execute the SQL tasks if no error_message and CLI flag set.
+    if ($db->isConnected()
```
ridz1208: not sure why you are checking if connected, we never check for that

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,234 @@
...
+                $bits[2]     = htmlspecialchars($bits[2]);
+                $output     .= "`$sql_query_table_name` $bits[0] default NULL,\n";
+                $column_name = str_replace('`', '', $db->escape($bits[1]));
+                $sql_query_columns[] = "`$column_name` $bits[0] default NULL,\n";
+            }
         }
-
+        $output .= "PRIMARY KEY  (`CommentID`)\n) ENGINE=InnoDB"
+            ." DEFAULT CHARSET=utf8;\n";
+        $fp      = fopen($file_name, "w");
```
ridz1208: do we want to write the file if we are using the --apply ?

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,234 @@
...
-        $paramId="";
-        $bits=explode("{@}",trim($item));
-        if(preg_match("/Examiner[0-9]*/" , $bits[1])){
-            continue;
-        }
-        switch($bits[0]){
-            //generate the CREATE TABLE syntax
+require_once 'generic_includes.php';
+// Display usage message if flag parameter is absent or invalid.
+if (!isset($argv[2])
```
ridz1208: this is not consistent in any way, but I've been using `tosql` and `confirm` instead of `--to-sql` and `--apply` we might want to consider documenting a standard at some point and using it in all the tools

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,234 @@
...
+    echo $usage;
+};
+
+$execute_mysql_query = $argv[2] === '--apply';
+$data          = file_get_contents($argv[1]);
+$instruments   = explode("{-@-}", trim($data));
+$error_message = "";
+// Do validation
+if (!is_array($instruments)) {
+    $error_message = "Instrument file could not be processed.";
```
ridz1208: maybe add new lines if you are gonna print those in the echo statement below

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,234 @@
...
+                "PRIMARY KEY  (`CommentID`)\n) ENGINE=InnoDB "
+                ."DEFAULT CHARSET=utf8;\n";
+            $statement            = $db->prepare($sql_query_statement);
+            $statement->execute();
+        } else {
+            $error_message = "There is already an instrument installed with 
+                this name.";
+        }
+        $statement = $db->prepare(
+            'REPLACE INTO test_names (Test_name, Full_name, Sub_group)'
```
ridz1208: another SQL statement duplication. might wanna consider using the same statement appended to the output on line 145.

duplications add the risk of commands diverging in the futur

tools/generate_tables_sql_and_testNames.php
```diff
@@ -1,118 +1,234 @@
...
+            $statement    = $db->prepare(
+                'INSERT INTO instrument_subtests'
+                .' (Test_name, Subtest_name, Description)'
+                .' VALUES (?, ?, ?)'
+            );
+            $subtest_name = $sql_query_table_name . "_page" . ($page_number+1);
+            $statement->bindParam(1, $sql_query_table_name, PDO::PARAM_STR);
+            $statement->bindParam(2, $subtest_name, PDO::PARAM_STR);
+            $statement->bindParam(3, $page, PDO::PARAM_STR);
+            $statement->execute();
```
ridz1208: why are you not using a regular insert command ?

johnsaigle:
@ridz1208 Can't answer any of that! @maltheism made the changes to the `generate...` code in #3674.

tools/generate_tables_sql_and_testNames.php
johnsaigle: I can change these to `tosql` and `confirm`. No need for this to be different.

ridz1208:
@maltheism questions redirected to you

tools/generate_tables_sql_and_testNames.php
maltheism: Hi @ridz1208 it's been awhile since last working on #3674 but the generate_tables_sql_and_testNames.php executes from the instrument_manager.class.inc and as seen:
```
$error_message = exec(
                    'php generate_tables_sql_and_testNames.php '
                    . escapeshellarg($new_file)
                );
```
The $error_message inside the generate_tables_sql_and_testNames.php is returned to the $error_message in the code above. The $error_message will be a string containing only one line text and if there is an error. Otherwise the $error_message will be empty.
Then this code runs:
```
if (!empty($error_message)
                    && strpos($error_message, "Instrument") !== false
                ) {
                    // See the /project/tables_sql/{file} for debugging the error.
                    error_log(
                        "Error uploading the Instrument(s) file: "
                        .$error_message
                    );
                    // Instrument(s)
                    throw new \LorisException(
                        "Error uploading the Instrument(s) file:\n".
                        "$error_message"
                    );
                }
```
That shows the error to the user if an error occurred and in a nice format since last testing it manually.

tools/generate_tables_sql_and_testNames.php
maltheism: @ridz1208 line 145 is writing to a file that the statement that could potentially be executed or not because the generate_tables_sql_and_testNames.php may be used to execute the sql statement or just build one. Zain told me this was a functionality requirement the generate_tables_sql_and_testNames.php needed when I was fixing that PR from potential sql injection attacks. I don't believe it's a duplicate.

tools/generate_tables_sql_and_testNames.php
maltheism: @ridz1208 please see the comments above for more detail about the PR. The changes were to prevent sql injection and by using the PDO parameters with binding to a statement.

maltheism:
Hi @ridz1208 I answered the questions. Please let me know if you need more clarification. That PR referenced was awhile ago and was to prevent sql injection by a malicious instrument builder file.

tools/generate_tables_sql_and_testNames.php
ridz1208: this script is also used from the terminal

tools/generate_tables_sql_and_testNames.php
ridz1208: line 145:

```
"REPLACE INTO test_names (Test_name, Full_name, Sub_group) "
                ."VALUES ('$sql_query_table_name', '$title', 1);\n"
```

All im saying is that maybe we should not be writing SQL twice in this script. write it once and use it where you need it.

ridz1208:
@maltheism there are still a bunch of comments that were not answered.

@johnsaigle are you still the one who will maintain this one or should @maltheism take it back (just so I know who to discuss with)

ridz1208:
TBH I'm not seeing the benefit here or "re-using" the `generate_tables_sql_and_testNames.php` script given that every SQL query/variable in that script was duplicated and a giant `if` statement was added to accomodate the use by the instrument manager.

Seeing this I would honestly prefer leaving the script untouched and having the instrument manager use it's own functions to make these changes and not reusing this script at all. ACTUALLY it would be much better for this script to use functions defined in the instrument manager while your at it

tools/generate_tables_sql_and_testNames.php
maltheism: @ridz1208 Yah I was told by Zain and that's why he wanted the functionality of just generating the queries to a file and without executing them. I think he tested this code before he left and found some bugs that I fixed. Do you want something like this:
`$error_message = "\nInstrument file could not be processed.";`
So the terminal output has it on a new line? I've not written php code before that executes from the terminal and so I'm not necessary sure if \n does what I think it does with console.log in javascript.

tools/generate_tables_sql_and_testNames.php
maltheism: I used code I previously saw in LORIS when writing this and it was awhile ago. It should be removed.

tools/generate_tables_sql_and_testNames.php
maltheism: I used code I previously saw in LORIS when writing this and it was awhile ago. It should be updated.

tools/generate_tables_sql_and_testNames.php
maltheism: I think the reason was logging potential errors from users attempting to execute instrument files if my memory serves me correctly. The server will output an Loris exception to the user and someone could potentially look in the apache logs but I thought it would be easier to save the query that resulted in an error. It can be discarded if you think it should.

tools/generate_tables_sql_and_testNames.php
maltheism: I'm not really positive what the question is about but to build a sql query with preventing an injection it's done with individual variables that are assigned with checking as found here:
http://php.net/manual/en/book.pdo.php
The $output is just to hold what the query as a whole would be and saved to a file because that functionality was requested. I think if you look at my PR with the commit history, the functionality wasn't there and later it was included. Maybe that answers your question.

tools/generate_tables_sql_and_testNames.php
maltheism: I think it's because I found the extra variable for just holding the `REPLACE INTO test_names (Test_name, Full_name, Sub_group)` as bad practice because less variables sometimes is better. 
Specifically line 189 is:
```
$statement = $db->prepare(
            'REPLACE INTO test_names (Test_name, Full_name, Sub_group)'
            .' VALUES (?, ?, 1)'
        );
```
The above is doing a prepare and it only would be able to store `REPLACE INTO test_names (Test_name, Full_name, Sub_group)` as a string and similar to line 145 it would still need the:
`."VALUES ('$sql_query_table_name', '$title', 1);\n"` which seems taste to me. It can be put into a variable if you think it should.

maltheism:
Hey @ridz1208, I added comments to the other questions I missed. I think the instrument manager has always been using the `generate_tables_sql_and_testNames.php` script. I originally first touched it to add the sql injection protection by using sql prepare statements. Then Zain told me the functionality of the script was also used for the terminal and which is why it has the other things. It should still work for the terminal and I remember it working as intended when I was testing it.

tools/generate_tables_sql_and_testNames.php
maltheism: I haven't seen --apply before.

tools/generate_tables_sql_and_testNames.php
maltheism: I think the original code was using $bits[1] and it was working with the instrument files. What is it supposed to be?

maltheism:
@ridz1208 I also think you're right that it makes sense to have separate methods instead of this file for the instrument manager because this file can be confusing when looking at it. Also it's a pain to debug because the code is being executed and not ran similar to all of Loris.

johnsaigle:
> Seeing this I would honestly prefer leaving the script untouched and having the instrument manager use it's own functions to make these changes and not reusing this script at all. ACTUALLY it would be much better for this script to use functions defined in the instrument manager while your at it

Definitely, but keep in mind that this PR was intended to address a specific bug, namely a pretty serious security issue. The module has always used the tools/ script as far as I know and while this should be refactored it's a bit out of scope here...

...unless the approach taken here should instead be to refactor the module now. It's not a great idea to link command line scripts to the web interface in general because the security concerns result in all these hoops we have to jump through. Thoughts?

tools/generate_tables_sql_and_testNames.php
ridz1208: to my knowledge `insert()` already does that

tools/generate_tables_sql_and_testNames.php
ridz1208: I def dont like it no

tools/generate_tables_sql_and_testNames.php
ridz1208: i think bits[1] here is the name of the field

tools/generate_tables_sql_and_testNames.php
ridz1208: yeah.. im trying to clean these up, you can take a look at some of my tool refactoring PRs to get an idea

tools/generate_tables_sql_and_testNames.php
johnsaigle: I believe the injection vector was in the table name or column name which `insert()` doesn't handle.

See https://redmine.cbrain.mcgill.ca/issues/14390

driusan:
As I said in my October 28th review that was never addressed, this isn't "fixing a bug", it's completely changing what the script does. This PR is no more "not refactoring" than doing Rida's suggestion  would be.

ridz1208:
@johnsaigle I understand my only issue here is that the changes made to fix the issue are putting at risk a script that is used very often by projects. and in the current state, it is backwards incompatible.

I like releasing quick fixes before working on a longterm fix but this seems to have taken more time then just creating an ajax call and copying 90% of this tools file into the ajax php file...

the way I see it, we just added here a big if statement for `--apply` and `--to-sql` but we barely reused any of the code previously written, we just wraped it and added new code... why not just add a new file instead ?

johnsaigle:
Alright well it sounds like this approach should probably be abandoned in favour of moving this module away from using the tools/ script and copying some of its functions elsewhere instead. I'll close this PR and get started on something else when I've got the time.

PR #4021, opened 2018-10-22
[User Accounts] Fix bug where passwords were recorded in plaintext in history table

johnsaigle:
### Brief summary of changes

This PR fixes an issue where two updates were triggered when a user's password was changed via the user_accounts module (either by the "Edit User" page or the "My Preferences" page). The first update caused the user's password to be logged in plaintext in the history table. The second properly overwrites the plaintext version with its hashed equivalent.

A script has been provided to query the DB for these passwords and record their hashed equivalents in a file under `projects/password_blacklist.lst`. #3962 will check this list and reject potential passwords that match. This is to prevent users from reusing passwords that may have been revealed by this bug.

Finally, an SQL patch is to be run AFTER generating the blacklist which will delete the plaintext passwords from the `history` table and force a password reset for all users.

### To test this change...

- Checkout my branch and update your password via the front-end
- Run `select old,new from history where col='Password_hash';` Your new password should not appear in the results.
- Run `tools/single_use/generate_password_blacklist.php` and verify that it creates a blacklist table in the Database.
- Run the SQL patch and confirm that the passwords have been cleared from the `history` table
- Try logging in to LORIS and verify that you must reset your password.


johnsaigle:
@driusan This PR would probably be a good one to fast track if you have the time.

tools/generate_password_blacklist.php
```diff
@@ -0,0 +1,52 @@
...
+// update so we'll limit the query to that. Additionally we will filter out
+// results from the `new` data that are password hashes i.e. those that begin
+// with the $ character.
+$sql = "select new as password from history where col like '%hash%' AND type='U' AND new NOT LIKE '\$%';";
+$result = $DB->pselect($sql, array());
+
+$passwords = array();
+foreach($result as $key => $value) {
+    array_push($passwords, ($value['password']));
+}
```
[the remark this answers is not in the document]
johnsaigle: Doesn't work unfortunately as `$value` is actually an array with its own key-value pair. Classic annoyance caused by the Database class.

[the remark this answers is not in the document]
johnsaigle: Thanks!

SQL/New_patches/2018-10-22-ExpireAllPasswords.sql
```diff
@@ -0,0 +1,6 @@
+-- This forces a password reset for all users the next time they login. Before
+-- this patch, all passwords changed by the front-end were stored in plaintext.
+UPDATE users SET Password_expiry='1990-01-01';
+-- Remove old plaintext passwords. Also affects old history entries when the
+-- column was named `hash` instead of `Password_hash`
+DELETE FROM history WHERE col LIKE "%hash%" AND tbl='users';
```
driusan: I don't see why this uses `like '%hash%'` instead of `= 'password_hash'`

tools/generate_password_blacklist.php
```diff
@@ -0,0 +1,50 @@
...
+        )
+    ),
+    FILE_APPEND
+);
+
+/* Function exists because array_map doesn't support multiple arguments */
+function hash_password($password) {
+    return password_hash($password, PASSWORD_DEFAULT);
+}
+?>
```
driusan: PHP closing tags aren't required/shouldn't be included in most cases.

SQL/New_patches/2018-10-22-ExpireAllPasswords.sql
johnsaigle: IIRC I saw some instances of passwords (or their md5 hashes) being stored in the history table from when that column was named `md5hash` or whatever it was.

driusan: Then use an OR to cover both cases.. `%hash%` wouldn't catch it anyways (IIRC the column was called `Password_md5` back when it was an md5)

SQL/Cleanup_patches/2018-10-22-ExpireAllPasswords.sql
```diff
@@ -3,4 +3,4 @@
 UPDATE users SET Password_expiry='1990-01-01';
 -- Remove old plaintext passwords. Also affects old history entries when the
 -- column was named `hash` instead of `Password_hash`
-DELETE FROM history WHERE col LIKE "%hash%" AND tbl='users';
+DELETE FROM history WHERE col = "Password_hash" OR col = "Password_md5" AND tbl='users';
```
driusan: This is missing brackets around the OR clause. This will be evaluated as `"Password_hash" OR col = ("Password_md5" AND tbl='users');` as is. Also, `'` is SQL standard quote for strings, not `"`.

SQL/Cleanup_patches/2018-10-22-ExpireAllPasswords.sql
```diff
@@ -0,0 +1,6 @@
+-- This forces a password reset for all users the next time they login. Before
+-- this patch, all passwords changed by the front-end were stored in plaintext.
+UPDATE users SET Password_expiry='1990-01-01';
+-- Remove old plaintext passwords. Also affects old history entries when the
+-- column was named `hash` instead of `Password_hash`
+DELETE FROM history WHERE col = "Password_hash" OR col = "Password_md5" AND tbl='users';
```
driusan:
```suggestion
DELETE FROM history WHERE (col = 'Password_hash' OR col = 'Password_md5') AND tbl='users';
```

I keep forgetting this suggest features exists now

johnsaigle: Thanks!

tools/generate_password_blacklist.php
```diff
@@ -0,0 +1,49 @@
...
+// results from the `new` data that are password hashes i.e. those that begin
+// with the $ character.
+$sql = "select new as password from history where col like '%hash%' AND type='U' AND new NOT LIKE '\$%';";
+$result = $DB->pselectCol($sql, array());
+
+$passwords = array();
+$passwords = array_values($result);
+
+$base = \NDB_Config::singleton()->getSetting('paths')['base'];
+$dest = "$base/project/password_blacklist.lst";
```
driusan: this depends on the resolution to #3962

tools/generate_password_blacklist.php
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env php
```
ridz1208: dont need this

johnsaigle: Are you opposed to it? In general I like being able to run the script just using `./script`.

It also makes it so that it prints the below comments to the terminal when you run it.

ridz1208: no just curious why

johnsaigle: I dunno, just a small thing. Less to type.

PapillonMcGill: Like this it will direct the server to use the first occurrence of `php` in the $PATH of the user running the script.
which might be dangerous if someone have access to the server (well, in that case, it might be to late anyway).

not using it, but just running the script from command line with `php tools/generate_password_blacklist.php` will use the first occurrence of `php` in the $PATH of the user running the script. So the same thing.

This line is thus completely non necessary. It is best not to include it, this way if the script as the `execute` bit set for the user(or group) the person would have to run it with `php tools/generate_password_blacklist.php`  instead of just typing `tools/generate_password_blacklist.php`

johnsaigle:
Blocked by #4188

tools/generate_password_blacklist.php
```diff
@@ -0,0 +1,62 @@
...
+ * delete the password entries which prevents the blacklist from being
+ * generated.
+ *
+ * The plaintext passwords appeared due to a bug in LORIS. As these passwords
+ * have potentially been viewed by administrators or could be read in the event
+ * of a data breach, we have chosen to reject them in this way.
+ */
+<?php
+require_once 'generic_includes.php';
+sleep(5);
```
johnsaigle: This should wait for input instead of manually sleep.

johnsaigle:
Changes made:
* The user is now explicitly prompted to use the script
* Documentation is more clear
* Removed the 'expire all passwords' SQL patch as it made the process convoluted. Now the script will populate the table as well as expire all current passwords
* Added blacklist table to the default schema

tools/single_use/generate_password_blacklist.php
```diff
@@ -0,0 +1,127 @@
...
+ * @param array  $answers  List of acceptable answers to prompt
+ *
+ * @return void
+ */
+function writeQuestion($question, $answers) : void
+{
+    echo $question . ' (' . implode('/', $answers) . '): ' . PHP_EOL;
+}
+/**
+ * Gets user input from STDIN and checks if it matches a valye in
```
PapillonMcGill:
```suggestion
 * Gets user input from STDIN and checks if it matches a value in
```

SQL/0000-00-00-schema.sql
```diff
@@ -2043,3 +2043,13 @@ CREATE TABLE `candidate_consent_history` (
...
+
+
+-- A table to store the hashed value of passwords that were mistakenly recored
+-- in the `history` table. These passwords will not be allowed to be used as
+-- LORIS user passwords as they have potentially been compromised.
+CREATE TABLE `password_blacklist` (
+    `ID` integer unsigned NOT NULL AUTO_INCREMENT,
+    `password_hash` varchar(255) DEFAULT NULL,
+    PRIMARY KEY (`ID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
driusan: This still needs to be changed to minor so that the patch will be run.

johnsaigle:
@ridz1208 You said you were going to base this off of minor, right?

ridz1208:
@johnsaigle Done. sorry it got buried by other stuff

SQL/0000-00-00-schema.sql
```diff
@@ -2044,3 +2044,13 @@ CREATE TABLE `candidate_consent_history` (
   `EntryStaff` varchar(255) DEFAULT NULL,
   CONSTRAINT `PK_candidate_consent_history` PRIMARY KEY (`CandidateConsentHistoryID`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+
+-- A table to store the hashed value of passwords that were mistakenly recored
+-- in the `history` table. These passwords will not be allowed to be used as
```
driusan: I'm not going to block over it, but I don't think this comment is accurate. Passwords can be blacklisted for any reason you want.

SQL/0000-00-00-schema.sql
```diff
@@ -2044,3 +2044,13 @@ CREATE TABLE `candidate_consent_history` (
...
+-- LORIS user passwords as they have potentially been compromised.
+CREATE TABLE `password_blacklist` (
+    `ID` integer unsigned NOT NULL AUTO_INCREMENT,
+    `password_hash` varchar(255) DEFAULT NULL,
```
driusan: Is it a specific type of hash or can it be any hashing algorithm? (Presumably the former so that the blacklist can be compared against..) If it needs to be a specific hashing algorithm, the column should probably be named to reflect what it's expecting.

tools/single_use/generate_password_blacklist.php
```diff
@@ -0,0 +1,127 @@
...
+    'history', array(
+        'col' => 'Password_md5',
+        'tbl' => 'users'
+    )
+);
+
+// Function exists because array_map doesn't support multiple arguments.
+function hash_password($password): string
+{
+    return password_hash($password, PASSWORD_DEFAULT);
```
driusan: For the purposes of a blacklist, I don't see how this is going to work. (I also don't see where the blacklist is being compared against in the PR to check how it's supposed to be working..)

password_hash is not going to be reproducible and will change with each run, and the algorithm changes with different PHP versions by design. It's also intended to be slow, while you want the blacklist checking to be relatively fast.

johnsaigle:
@driusan As I explained in the description, #3962 handles the actual checking/verification of blacklisted passwords. This PR is intended primarily as a bugfix to invalidate the passwords already exposed. It also paves the way for #3962's blacklist-checking. 

It might be best to set-up a quick in-person discussion where I can explain how this all works.

SQL/0000-00-00-schema.sql
johnsaigle: As you noted below, the password algorithm can change so it's best not to hard-code a value here. PHP's default hashing algorithm at this time is bcrypt which stores an identifying prefix before the actual hash (`$2y`). `password_verify` takes care of identifying the algorithm, calculating the complexity, and checking that an input matches the hash.

tools/single_use/generate_password_blacklist.php
johnsaigle: The functionality is implemented in #3962.

>password_hash is not going to be reproducible and will change with each run
I don't know what you mean here. Password hashing should be reproducible and I'm using the built-in functions that we rely on to create passwords in the first place. 

> the algorithm changes with different PHP versions by design
We can use `password_verify` which should always correspond to `password_hash`.

> It's also intended to be slow, while you want the blacklist checking to be relatively fast
This PR is a single_use script so the speed shouldn't really be a factor.
If speed proves to be an issue in blacklist checking in the other PR, we can do a `password_hash` once on the input and do a string `===` check on blacklist values coming from the DB. This would be faster than doing `password_verify` in a for-each loop (current implementation).

johnsaigle:
@xlecours Would you mind taking a look at this PR and the other one to see if they make sense to you?

kchatpar:
@johnsaigle the path to tools/generate_password_blacklist.php is actually tools/single_use/generate_password_blacklist.php. You might want to update the PR summary.

johnsaigle:
@christinerogers 
> What happens if a (new) user tries to use a blacklisted password? Do they get an appropriate error message? (that password is invalid, please select another)

Nothing as of now because this is only laying the foundation for #3962 (also linked in PR description). This PR displays a specific error message for blacklisted passwords at this line: https://github.com/aces/Loris/pull/3962/files#diff-1947816db4abdcdd8722747580999a6dR594

I will update the readme.

christinerogers:
@johnsaigle  Feel free to push all my User Accounts suggestions to the #3962 PR - doesn't need to hold up merging this.

johnsaigle:
@kchatpar Can you add the `Passed Manual Tests` label? I thought we did when we were working together but it seems like it hasn't worked.

@driusan Can you respond to my January comment and update your review?

tools/single_use/generate_password_blacklist.php
driusan: We can't do a password hash on the input and a string comparison, we'll never be able to do that because by design password_hash will generate a different hash each time it's run. This isn't verifying passwords, this is checking if a password is blacklisted. Using password_hash is not scalable. In order to check whether or not a password is blacklisted, you need to run a (intentionally) slow and changing hashing algorithm against every single blacklisted password.

SQL/0000-00-00-schema.sql
driusan: I don't think password_verify is suitable for the purposes of looking up a blacklist.

johnsaigle:
@LeighMac I'm removing my assignment from this one as I am awaiting feedback from others so that ball isn't in my court here.

driusan:
@johnsaigle you've already gotten responses to all of your questions.

johnsaigle:
@driusan Sorry I missed them from Friday.

johnsaigle:
I'm making the decision to close this PR.

While I'm not convinced that my approach to blacklisting is flawed, it seems to be a sticky issue for people. Ultimately this part of the PR is sort of a bonus feature. The main point is to delete the existing passwords and to prevent them from being recorded going forward.

This is also holding up #3962 unnecessarily. At the time it felt like I may as well do blacklisting while modernizing the password strength but things have shifted.

I'll re-open this as a fresh PR with references to blacklisting removed. It can be added separately another time.

PR #4023, opened 2018-10-23
[Configuration] Path settings must be valid paths

johnsaigle:
### Brief summary of changes

I was recently able to develop an exploit where I could upload and execute scripts via manipulating config path settings. This PR ensures that these settings actually correspond to existing paths on the system before changing their values in the database.

### To Test
* Try changing one of the Configuration settings under `Paths` to an invalid path. You'll get a `400` error.

### DIFF
--- a/modules/configuration/ajax/process.php
+++ b/modules/configuration/ajax/process.php
@@ -32,73 +32,92 @@ $client->makeCommandLine();
 $client->initialize();
 $factory = \NDB_Factory::singleton();
 $DB      = $factory->database();
-$pathIDs = getPathIDs();
 foreach ($_POST as $key => $value) {
-    if (is_numeric($key)) { //update
+    if (is_numeric($key)) {
+        // When a $key is numeric, it means we are updating the entry in the
+        // Config table with ID == $key.
         if ($value == "") {
-            $DB->update('Config', array('Value' => null), array('ID' => $key));
+            // A blank value is the same as deleting.
+            $DB->delete('Config', array('ID' => $key));
         } else {
-            // if no duplicate value then do updating
-            if (noDuplicateInDropdown($key, $value)) {
-                if (in_array(intval($key), $pathIDs, true)) {
-                    if (!validPath($value)) {
-                        $err = 'Directory `'
-                            . htmlspecialchars($value, ENT_QUOTES)
-                            . '` is invalid';
-                        // Set response code and display error message.
-                        displayError(400, $err);
-                        return;
-                    }
-                }
-                $DB->update(
-                    'Config',
-                    array('Value' => $value),
-                    array('ID' => $key)
-                );
+            if (! noDuplicateInDropdown($key, $value)) {
+                // Don't alter the table if the same key was passed twice.
+                continue;
             }
-        }
-    } else { //add new or remove
-        $keySplit   = explode("-", $key);
-        $valueSplit = explode("-", $value);
-        if ($keySplit[0] == 'add') {
-            if ($value !== "") {
-                if (countDuplicate($keySplit[1], $value) == '0') {
-                    $DB->update(
-                        'Config',
-                        array('Value' => $value),
-                        array(
-                         'ConfigID' => $keySplit[1],
-                        )
-                    );
-                } else {
-                    header("HTTP/1.1 303 Duplicate value");
-                    exit();
+            // Get all the IDs in Config with a relation to an entry in the
+            // ConfigSettings table that has the web_path data type.
+            $pathIDs = getPathIDs('Config');
+            if (in_array($key, $pathIDs)) {
+                // If updating a path config setting, make sure that the
+                // value is a valid path. This helps to prevent malicious
+                // input.
+                if (!validPath($value)) {
+                    $err = 'Directory `'
+                        . htmlspecialchars($value, ENT_QUOTES)
+                        . '` is invalid';
+                    // Set response code and display error message.
+                    displayError(400, $err);
+                    return;
                 }
-                if (in_array(intval($keySplit[1]), $pathIDs, true)) {
-                    if (!validPath($value)) {
-                        $err = 'Directory `' . htmlspecialchars($value, ENT_QUOTES)
-                            . '` is invalid';
-                        // Set response code and display error message.
-                        displayError(400, $err);
-                        return;
-                    }
-                }
-                $DB->insert(
-                    'Config',
-                    array(
-                     'ConfigID' => $keySplit[1],
-                     'Value'    => $value,
-                    )
-                );
             }
-        } elseif ($valueSplit[0] == 'remove') {
+            // Update the config setting to the new value.
             $DB->update(
                 'Config',
-                array('Value' => null),
-                array('ID' => $valueSplit[1])
+                array('Value' => $value),
+                array('ID' => $key)
+            );
+        }
+    } else {
+        // This branch is executed when the key is prefixed with the string
+        // 'add' or 'remove' (i.e. the full key is not numeric and triggers this
+        // else branch).
+        // An example $key is 'add-38' which means "Add an entry in the Config
+        // table using foriegn key 38 from the ConfigSettings table." The
+        // number refers to ConfigSettings.ID which is different from
+        // Config.ID.
+        // This is different from the above is_numeric case; this makes use of
+        // Config.ID, not Config.ConfigID (which is a FK to ConfigSettings.ID).
+        // The Config table is the one that will be modified here.
+        $keySplit         = explode("-", $key); // e.g. 'add-10' or 'remove-49'
+        $action           = $keySplit[0];
+        $ConfigSettingsID = $keySplit[1];
+        if ($action == 'add') {
+            // This branch adds a new entry to the Config table.
+            if ($value === "") {
+                continue;
+            }
+            if (isDuplicate($ConfigSettingsID, $value)) {
+                header("HTTP/1.1 303 Duplicate value");
+                exit();
+            }
+            // Get all the IDs in ConfigSettings with the web_path data type.
+            $pathIDs = getPathIDs('ConfigSettings');
+            if (in_array($ConfigSettingsID, $pathIDs)) {
+                if (!validPath($value)) {
+                    $err = 'Directory `' . htmlspecialchars($value, ENT_QUOTES)
+                        . '` is invalid';
+                    // Set response code and display error message.
+                    displayError(400, $err);
+                    return;
+                }
+            }
+            // Add the new setting
+            $DB->insert(
+                'Config',
+                array(
+                 'ConfigID' => $ConfigSettingsID, // FK to ConfigSettings.
+                 'Value'    => $value,
+                )
+            );
+        } elseif ($action == 'remove') {
+            // Delete an entry from the Config table.
+            $DB->delete(
+                'Config',
+                array('ConfigID' => $ConfigSettingsID)
             );
         }
     }
+    unset($pathIDs);
 }
 /**
  * Check Duplicate value
@@ -106,20 +125,20 @@ foreach ($_POST as $key => $value) {
  * @param string $key   The value of the key
  * @param string $value The value of the value
  *
- * @return int $result
+ * @return bool $result
  */
-function countDuplicate($key,$value)
+function isDuplicate($key, $value): bool
 {
-    $factory   = \NDB_Factory::singleton();
-    $DB        = $factory->database();
-       $result = $DB->pselectOne(
-           "SELECT COUNT(*) FROM Config WHERE ConfigID =:ConfigID AND Value =:Value",
-           array(
-            'ConfigID' => $key,
-            'Value'    => $value,
-           )
-       );
-       return $result;
+    $factory = \NDB_Factory::singleton();
+    $DB      = $factory->database();
+    $result  = $DB->pselectOne(
+        "SELECT COUNT(*) FROM Config WHERE ConfigID =:ConfigID AND Value =:Value",
+        array(
+         'ConfigID' => $key,
+         'Value'    => $value,
+        )
+    );
+    return intval($result) > 0;
 }
 /**
  * Check dropdown list Duplicate value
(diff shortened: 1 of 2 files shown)


modules/configuration/ajax/process.php
```diff
@@ -28,12 +28,26 @@
...
+$DB      =& Database::singleton();
+$pathIDs = getPathIDs();
 foreach ($_POST as $key => $value) {
     if (is_numeric($key)) { //update
         if ($value == "") {
             $DB->delete('Config', array('ID' => $key));
         } else {
+            // Validate that a path exists before entering into database.
+            if (in_array(intval($key), $pathIDs, true)) {
+                if (!file_exists($value)) {
```
PapillonMcGill: Should there be a check to ensure it is not *outside* of Apache reach and it is ok to use (i.e.: "/bin/...")

modules/configuration/ajax/process.php
```diff
@@ -28,12 +28,30 @@
 $client->makeCommandLine();
 $client->initialize();
 
-$DB =& Database::singleton();
+$DB      =& Database::singleton();
```
paiva: `\Database`

johnsaigle:
@kongtiaowang You were right. I fixed the problem. Please test again if you have time.

modules/configuration/ajax/process.php
ridz1208: @johnsaigle

ridz1208:
@paiva can you please test this one ?

johnsaigle:
This might cause issues with the Logs path because I think it's deliberately set up as a partial path (i.e. tools/logs/)

I'm not sure where this is used exactly but I think I'll need to issue another PR to update those usages once this is merged.

modules/configuration/ajax/process.php
PapillonMcGill: @johnsaigle still missing in action

johnsaigle: @PapillonMcGill line 47.

PapillonMcGill: @johnsaigle right, seen it before but I must had some fog in the eyes this morning.

modules/configuration/ajax/process.php
```diff
@@ -59,7 +75,45 @@
...
+{
+    $ids = array();
+    $DB  = \Database::singleton();
+    /* Query adapated from:
+     * https://github.com/aces/Loris/wiki/Project-Customization
+     */
+    $query  = "SELECT c.ID FROM Config c "
+        . "LEFT JOIN ConfigSettings cs ON (c.ConfigID = cs.ID) "
+        . "JOIN ConfigSettings csp ON (cs.Parent = csp.ID) "
+        . "WHERE csp.Name = 'paths';";
```
driusan: I don't think just grabbing all config settings based on the name of the section is going to work. You don't know the intended use of the path (or even that it's a path and not some other type specified in the section), it could be a path for imaging scripts, there could be config options that aren't paths in the section of that name, or there could be config options that are paths which someone added to another section.

To do this right there needs to be (at least) 1 DataType added to ConfigSettings.DataType to indicate that a setting is intended to be a path, probably 2 (one for paths used be the web server one for arbitrary paths that don't get validated.)

PierreEMorin:
This code does what it is supposed to do. 

However, at the moment, default path values such as "/PATH/TO/EXTERNAL/LIBRARY/" or "/PATH/TO/Genomic-Data/" are not allowed anymore so the user is left with either:
    1. leave it blank
    2. put a valid but inadequate path (i.e. data/%project%/data for genomics) 
    3. add valid paths to the server that may be never used (i.e. a project not using the genomic module)

Should blank be allowed?  or should default values from the ConfigTables schema be allowed?

johnsaigle:
I guess ideally we would have a toggle switch on each of the modules saying whether a project will use them or not...

I think it makes sense to allow blank values in this case. The default params don't really make sense anyway and only some of the paths are configured to have one in the /NOT/REAL/PATH format. 

Blank paths will probably result in ConfigurationExceptions at some point in whatever modules use them. I don't know if this is the best solution but I think it's what we have available right now.

johnsaigle:
As discussed in the meeting, we are going to use blank paths as default and rely on the calling code to do validation.

This PR is now ready for review.

modules/configuration/ajax/process.php
```diff
@@ -18,22 +22,39 @@
...
 }
 
 $client = new NDB_Client();
-$client->makeCommandLine();
 $client->initialize();
 
-$DB =& Database::singleton();
+$DB      = \Database::singleton();
+$pathIDs = getPathIDs();
+print_r($pathIDs);
```
driusan: left over debug statement

driusan:
@johnsaigle I don't see any changes to the SQL to make paths blank as your comment suggests?

johnsaigle:
@driusan Sorry I was unclear above. You're right I haven't changed the SQL files yet. Will do so.

modules/configuration/js/configuration_helper.js
```diff
@@ -71,21 +71,24 @@ $(function () {
         e.preventDefault();
 
         var form = $(this).serialize();
```
maltheism: var form = $(this).serialize();
replace to:
`const form = $(this).serialize();`

modules/configuration/js/configuration_helper.js
```diff
@@ -71,21 +71,24 @@ $(function () {
...
+        $('.submit-area > label').remove();
 
         $.ajax({
             type: 'post',
             url: loris.BaseURL + '/configuration/ajax/process.php',
             data: form,
             success: function () {
-                var html = "<label>Submitted</label>";
-                $(html).hide().appendTo('.submit-area').fadeIn(500).delay(1000).fadeOut(500)
+                var html = "<label>Update successful</label>";
```
maltheism: `let html = "<label>Update successful</label>";`

modules/configuration/js/configuration_helper.js
```diff
@@ -71,21 +71,24 @@ $(function () {
...
+                $(html).hide().appendTo('.submit-area').fadeIn(500).delay(1000)
                 $('input[type="reset"]').attr('disabled','disabled');
             },
             error: function(xhr, desc, err) {
                 console.log(xhr);
                 console.log("Details: " + desc + "\nError:" + err);
+                var html = "<label>" + xhr.responseText + "</label>";
```
maltheism: `let html = "<label>" + xhr.responseText + "</label>";`

modules/configuration/js/configuration_helper.js
```diff
@@ -71,21 +71,24 @@ $(function () {
...
             data: form,
             success: function () {
-                var html = "<label>Submitted</label>";
-                $(html).hide().appendTo('.submit-area').fadeIn(500).delay(1000).fadeOut(500)
+                var html = "<label>Update successful</label>";
+                $(html).hide().appendTo('.submit-area').fadeIn(500).delay(1000)
                 $('input[type="reset"]').attr('disabled','disabled');
             },
             error: function(xhr, desc, err) {
                 console.log(xhr);
```
maltheism: We may not want to leave console.log in the frontend. It's similar to leaving print_r in php. There is console.error as well for errors. @HenriRabalais @zaliqarosli suggestions?

modules/configuration/ajax/process.php
```diff
@@ -59,7 +80,40 @@
...
+ * @param int    $code A valid HTTP Response Code.
+ * @param string $msg  Error text to display.
+ *
+ * @return void
+ */
+function displayError(int $code, string $msg): void
+{
+    // Display error message and return
+    http_response_code($code);
+    echo "ERROR: $msg";
```
maltheism: Hi @johnsaigle I'm not sure if it should be addressed in this PR but we're now using new \LORIS\Http\Response()) for the ajax.

johnsaigle: Yeah I think that depends on a larger refactoring of the module which isn't what this PR aims to do. it's a patch over the old way of doing things. Config eventually needs to be reworked.

modules/configuration/js/configuration_helper.js
zaliqarosli: i think console.error is fine!

zaliqarosli: (don't like console.log though) i think our linting rules allow error, but not log

modules/configuration/js/configuration_helper.js
johnsaigle: I'm not really looking to modernize the JS code in this PR, especially since it isn't Reactified right now. I think this change and your below suggestion about `console.log` can be addressed in another `Cleanup` PR. I'd like to keep this work focused on the security feature that this adds.

maltheism: @johnsaigle okay, I just saw one of the changes replacing var to let and thought to comment about this one as well. You can mark my suggestions as resolved.

modules/configuration/ajax/process.php
```diff
@@ -18,22 +22,38 @@
...
 
 $client = new NDB_Client();
-$client->makeCommandLine();
 $client->initialize();
 
-$DB =& Database::singleton();
+$DB      = \Database::singleton();
+$pathIDs = getPathIDs();
 foreach ($_POST as $key => $value) {
     if (is_numeric($key)) { //update
```
jesscall: When going from a blank field and entering an invalid path, update is successful. (see screenshot)

I think the issue is that validation of path only happens if the key is numeric. Though `configuration_helper.js` has a function which concatenates the prefix `add-` to the ID key (making it non-numeric)

Try printing the POST request to your error log to see the case of the add- prefix
EX:
`Array\n(\n    [16] => /var/www/loris/\n    [21] => tools/logs/\n    [add-22] => \n    `
.... `[add-65] => helloWorld/\n` ....


Screenshot from 2019-05-06 15-12-13

johnsaigle:
@jesscall Thanks for testing and providing all that helpful feedback. I think I fixed the issue with the latest commit. Would you be able to test again?

maltheism:
Hi @johnsaigle I tested by submitting an invalid path and the post appeared to had went through. Then I refreshed the page and received this error message:
```
Fatal error: Uncaught LorisNoSuchModuleException: No such module: configuration in /Users/alizee/Development/GitHub/McGill/Loris/php/libraries/Module.class.inc on line 63
--
```

johnsaigle:
@maltheism I can't reproduce this.

johnsaigle:
@driusan @ridz1208 If at all possible, it would be good to get this PR included into the 21.0 release. It fixes a pretty major vulnerability. I think it's in a state where it could be quickly tested and merged. 

Let me know if you think it's possible.

sruthymathew123:
@johnsaigle  I tested this PR and I found some issues. Sometimes the validation worked and sometimes it  didn't. 
Case1:I saved Genomic Data Path as empty and submitted the data. - Update Successful
update_success_1

Case2:I Refreshed the page and added Invalid Path to Genomic Data Path and submitted- Update Successful
success_2

Case 3: Without any refresh, I again gave an invalid path-  ERROR: Directory `/PATH/TO/Genomic-Data/sds` is invalid
not_success_3

johnsaigle:
Thanks for testing @sruthymathew123. I'll take a look

johnsaigle:
@sruthymathew123 I believe I've fixed the error. I also rebased this PR onto the 21.0 release branch.

Let me know if it works properly now.

sruthymathew123:
@johnsaigle Now it seems to be working as expected with the different test conditions.

johnsaigle:
Thanks for testing!

modules/configuration/ajax/process.php
```diff
@@ -3,7 +3,11 @@
  * This file is used by the Configuration module to update
  * or insert values into the Config table.
  *
- * PHP version 5
+ * FIXME This code should be refactored away from using the separate 'ajax'
+ * file model toward using a more robust Loris Module-based approached.
+ * This is dependent on the Reactifying of the Configuration module generally.
```
driusan: Not a blocker, but the last sentence is false.

modules/configuration/ajax/process.php
```diff
@@ -18,32 +22,39 @@
...
 $client->makeCommandLine();
 $client->initialize();
 $factory = \NDB_Factory::singleton();
 $DB      = $factory->database();
+$pathIDs = getPathIDs();
 foreach ($_POST as $key => $value) {
     if (is_numeric($key)) { //update
         if ($value == "") {
-            $DB->delete('Config', array('ID' => $key));
+            $DB->update('Config', array('Value' => null), array('ID' => $key));
```
driusan: Why is this changed?

modules/configuration/ajax/process.php
```diff
@@ -52,20 +63,40 @@
         if ($keySplit[0] == 'add') {
             if ($value !== "") {
                 if (countDuplicate($keySplit[1], $value) == '0') {
-                    $DB->insert(
+                    $DB->update(
```
driusan: This will not work when the value isn't in the table

modules/configuration/ajax/process.php
johnsaigle: The basic idea of my changes is to verify whether the incoming data is a `web_path` type in the DB and if so make sure that it's a valid path.

When debugging a problem another reviewer had, i realized that the script was actually deleting values when they are blank.

This is a problem because we decided in a meeting at some point that the paths should be left blank when they're not configured (rather than allowing for `/INVALID/PLACEHOLDER/PATHS` as a value).

I figured it would make sense to keep the IDs in the database and set them to null rather than delete them. It makes more sense to me intuitively to have a setting be blank/uninitialized than for it to vanish.

Do you see this causing problems?

driusan: Yes, when new config settings are added/unchanged from the default they do not have any entry in the Config table. The update statement will return 0 rows and report success but silently do nothing. As a result, no one will ever be able to set a config setting for the first time for any config that doesn't have a non-empty default. Not having a value in the table when the setting is blank is correct behaviour.

johnsaigle: > when new config settings are added/unchanged from the default they do not have any entry in the Config table

Aren't defaults stored in the Config table? ConfigSettings doesn't have any information about the value.

driusan:
@johnsaigle Why did you add "Discussion Required" to this PR?

johnsaigle:
@driusan I replied to your above comment and you didn't follow up.

I think that this PR makes sense the way I've done it. Maybe it makes the most sense for us to discuss it in person so I can walk you through my thinking.

johnsaigle:
@driusan I modified this PR to use insert/delete instead of updates, as requested. Please review when you have time.

modules/configuration/ajax/process.php
```diff
@@ -30,65 +32,112 @@
...
+        $keySplit         = explode("-", $key); // e.g. 'add-10' or 'remove-49'
+        $action           = $keySplit[0];
+        $ConfigSettingsID = $keySplit[1];
+        if ($action == 'add') {
+            // This branch adds a new entry to the Config table.
+            if ($value === "") {
+                continue;
+            }
+            if (isDuplicate($ConfigSettingsID, $value)) {
+                header("HTTP/1.1 303 Duplicate value");
```
driusan: HTTP/1.1 defines A 303 response code is "See Other", not "Duplicate Value"

modules/configuration/ajax/process.php
```diff
@@ -30,65 +32,112 @@
...
+                 'ConfigID' => $ConfigSettingsID, // FK to ConfigSettings.
+                 'Value'    => $value,
+                )
+            );
+        } elseif ($action == 'remove') {
+            // Delete an entry from the Config table.
+            $DB->delete(
+                'Config',
+                array('ConfigID' => $ConfigSettingsID)
+            );
```
driusan: else throw exception?

modules/configuration/ajax/process.php
```diff
@@ -30,65 +32,112 @@
...
+        // An example $key is 'add-38' which means "Add an entry in the Config
+        // table using foriegn key 38 from the ConfigSettings table." The
+        // number refers to ConfigSettings.ID which is different from
+        // Config.ID.
+        // This is different from the above is_numeric case; this makes use of
+        // Config.ID, not Config.ConfigID (which is a FK to ConfigSettings.ID).
+        // The Config table is the one that will be modified here.
+        $keySplit         = explode("-", $key); // e.g. 'add-10' or 'remove-49'
+        $action           = $keySplit[0];
+        $ConfigSettingsID = $keySplit[1];
```
driusan:
```suggestion
        $ConfigSettingsID = $keySplit[1];
        assert(count($keySplit) == 2);
```

modules/configuration/ajax/process.php
johnsaigle: OK. I didn't write this. Should we change this to Bad Request?

driusan: to be honest I'm not entirely sure what the "duplicate" check is trying to prevent and how it relates to settings which have AllowMultiple set.. but if anything,  yes, `400 Bad Request` is at least a valid HTTP response code.

johnsaigle: Yeah me neither.

johnsaigle:
@driusan I made your requested changes.

modules/configuration/ajax/process.php
johnsaigle: done

johnsaigle:
@driusan Tagging you again because this ideally would go into the next `minor` release as discussed in the LORIS meeting.

PR #4024, opened 2018-10-24
[Imaging Uploader] Prevent duplicate insertions

davidblader:
### Brief summary of changes
Imaging Uploader only checks if a file already exists on the front end. This makes it possible to insert duplicate scans which happens surprisingly often on CCNA. This change adds a backend check for an existing mri_upload entry that is not being intentionally overwritten / updated.

### This resolves issue...
mri_upload duplicates

### To test this change...
Open the Imaging Uploader Upload tab in two browser tabs. Fill out both simultaneously for the same scan, then submit one after the other. The first submission should go through and the other should throw an error.

### DIFF
--- a/modules/imaging_uploader/php/imaging_uploader.class.inc
+++ b/modules/imaging_uploader/php/imaging_uploader.class.inc
@@ -382,7 +382,9 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
             array('ul' => "%$file_name")
         );
 
-        if (!$updateFile && !empty($id)) {
+        $uploadPath = $config->getSetting('MRIUploadIncomingPath') . $file_name;
+
+        if (!$updateFile && !empty($id) && file_exists($uploadPath)) {
             array_push(
                 $errors['mri_file'],
                 "This file has already been uploaded! Please refresh your browser."
--- a/modules/imaging_uploader/php/imaging_uploader.class.inc
+++ b/modules/imaging_uploader/php/imaging_uploader.class.inc
@@ -369,6 +369,25 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
                 " .tgz, .tar.gz or .zip."
             );
         }
+        ///////////////////////////////////////////////////////////////////////
+        //////// Make sure there's no existing identical entry ////////////////
+        //////// for "new" scans //////////////////////////////////////////////
+        ///////////////////////////////////////////////////////////////////////
+        $updateFile = $args['values']['overwrite'] ? true : false;
+        $id       = $db->pselectOne(
+            "SELECT 
+                IF (InsertionComplete=0 AND 
+                (ISNULL(Inserting) OR Inserting=0), UploadID, NULL) as UploadID 
+                FROM mri_upload WHERE UploadLocation LIKE :ul",
+            array('ul' => "%$file_name")
+        );
+
+        if (!$updateFile && !empty($id)) {
+            array_push(
+              $errors['mri_file'],
+                "This file has already been uploaded! Please refresh your browser."
+            );
+        }
         // Loops through associative array of error messages
         // If there are error messages, array is passed to front end
         // If there are no error messages but the file name is wrong,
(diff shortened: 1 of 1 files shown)


modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -369,6 +369,25 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
...
+        //////// for "new" scans //////////////////////////////////////////////
+        ///////////////////////////////////////////////////////////////////////
+        $updateFile = $args['values']['overwrite'] ? true : false;
+        $id         = $db->pselectOne(
+            "SELECT 
+                IF (InsertionComplete=0 AND 
+                (ISNULL(Inserting) OR Inserting=0), UploadID, NULL) as UploadID 
+                FROM mri_upload WHERE UploadLocation LIKE :ul",
+            array('ul' => "%$file_name")
+        );
```
johnsaigle: It might be good to add a `file_exists` check here if you have access to the full path to the file. This only checks whether there's a record in the database.

modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -369,6 +369,25 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
...
+
+        if (!$updateFile && !empty($id)) {
+            array_push(
+                $errors['mri_file'],
+                "This file has already been uploaded! Please refresh your browser."
```
paiva: Should "refresh your browser" be part of the error message?

davidblader: i added it because this error is typically caught on the front end. i added the browser refresh suggestion in order to update the react component's state. not opposed to removing it though

johnsaigle: If you look around here there's some sort of auto-refresh redirect thing. I'm not sure if it's exactly what you want but if a refresh is necessary I think I like this better than asking the user to do it.

https://github.com/aces/Loris/blob/33de6999da1748f20ba23c17d77570052bd5a6e4/modules/issue_tracker/jsx/IssueForm.js#L330

modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -369,6 +369,27 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
                 " .tgz, .tar.gz or .zip."
             );
         }
+        ///////////////////////////////////////////////////////////////////////
+        //////// Make sure there's no existing identical entry ////////////////
+        //////// for "new" scans //////////////////////////////////////////////
+        ///////////////////////////////////////////////////////////////////////
+        $updateFile = $args['values']['overwrite'] ? true : false;
```
johnsaigle: I'm not sure this ternary operator is necessary. What's the value of 'overwrite'? Can you just pass it along to L387?

davidblader: i just copied it over from the `importFile($file, $args)` function. i think it only gets set when the user confirms they want the original upload overwritten

kongtiaowang:
pass test and ready to go.

nicolasbrossard:
I opened two tabs in Google Chrome and tried uploading the same file in both of these tabs: one of them reported an upload failure, saying that the file was already uploaded (as expected) :+1:

modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -369,6 +369,27 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
                 " .tgz, .tar.gz or .zip."
             );
         }
+        ///////////////////////////////////////////////////////////////////////
+        //////// Make sure there's no existing identical entry ////////////////
+        //////// for "new" scans //////////////////////////////////////////////
+        ///////////////////////////////////////////////////////////////////////
```
driusan: How long did it take you to type all those slashes?

PR #4025, opened 2018-10-24
[Smarty] Add noopener and noreferrer to external links

johnsaigle:
### Brief summary of changes
Without [noopener](https://mathiasbynens.github.io/rel-noopener/), an external page that is opened and is malicious or compromised gains control over the `window` of the opening page's DOM. 
Basically it allows a website partial control over the tab that a user navigated from.

Noreferrer is a workaround for the same for browsers that do not support noopener. It also scrubs the Referer header which is a tiny privacy boost.

Also checkout https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

### To test this change...

- [ ] Make sure external links still work, such as the study links in the footer.

### DIFF
--- a/smarty/templates/main.tpl
+++ b/smarty/templates/main.tpl
@@ -478,7 +478,7 @@
                     Powered by LORIS &copy; {$currentyear}. All rights reserved.
                 </div>
       		<div align="center" colspan="1">
-                    Created by <a href="http://mcin-cnim.ca/" target="_blank" rel="noopener noreferrer">
+                    Created by <a href="http://mcin-cnim.ca/" target="_blank">
                          MCIN
                     </a>
                 </div>
--- a/smarty/templates/main.tpl
+++ b/smarty/templates/main.tpl
@@ -466,7 +466,7 @@
                         </li>
                         {foreach from=$links item=link}
                                 <li>
-                                    <a href="{$link.url}" target="{$link.windowName}">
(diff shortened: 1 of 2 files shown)


paiva:
I think you should add these attributes in the login page too

smarty/templates/main.tpl
```diff
@@ -478,7 +478,7 @@
                     Powered by LORIS &copy; {$currentyear}. All rights reserved.
                 </div>
       		<div align="center" colspan="1">
-                    Created by <a href="http://mcin-cnim.ca/" target="_blank">
+                    Created by <a href="http://mcin-cnim.ca/" target="_blank" rel="noopener noreferrer">
```
driusan: These (in particular noreferrer) shouldn't be added to hosts which we know we control like mcin.ca or loris.ca

johnsaigle:
@paiva I believe the public_layout template covers the login page.

smarty/templates/main.tpl
paiva: Does this include `<a href="http://www.mni.mcgill.ca" target="_blank"> Montreal Neurological Institute and Hospital </a>`  as well ??

`

johnsaigle: Depends how much you trust McGill security. :)

PR #4027, opened 2018-10-24
[Core: User] Add type declarations to function signatures

johnsaigle:
### Brief summary of changes
Types are good!

Further PRs to follow if Travis fails.

### Related 
- [ ] #3878, #4020 

##### Bugs Found
* #4033

### To test this change...

- [ ] Everything should stay the same.

### DIFF
--- a/php/libraries/User.class.inc
+++ b/php/libraries/User.class.inc
@@ -36,12 +36,12 @@ class User extends UserPermissions
      *
      * Selects a user and its permissions
      *
-     * @param string $username Identifies the user
+     * @param string|null $username Identifies the user
      *
      * @return User
      * @access public
      */
-    public static function &factory(string $username): \User
+    public static function &factory(?string $username): \User
     {
         $obj = new User;
 
--- a/php/libraries/User.class.inc
+++ b/php/libraries/User.class.inc
@@ -113,7 +113,7 @@ class User extends UserPermissions
         // store user data in object property
         $row['examiner']  = $examiner_info;
         $row['CenterIDs'] = $user_cid;
-        $obj->_userInfo    = $row;
+        $obj->_userInfo   = $row;
         return $obj;
     }
 
--- a/php/libraries/User.class.inc
+++ b/php/libraries/User.class.inc
@@ -36,12 +36,12 @@ class User extends UserPermissions
      *
      * Selects a user and its permissions
      *
-     * @param string $username Identifies the user
+     * @param string|null $username Identifies the user
      *
      * @return User
      * @access public
      */
-    static function &factory(string $username): \User
+    static function &factory(?string $username): \User
     {
         $obj = new User;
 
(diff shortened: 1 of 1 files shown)


driusan:
Should this declare strict_types for the file?

php/libraries/User.class.inc
```diff
@@ -182,13 +182,13 @@ class User extends UserPermissions
...
      *
-     * @param string $var Name of variable to get
+     * @param string|null $var Name of variable to get
      *
      * @note   Call without any parameters to get the entire user data array
      * @return mixed
      * @access public
      */
-    function getData($var = null)
+    function getData(?string $var = null)
```
kongtiaowang: Why this function doesn't have a return type?

PapillonMcGill: @kongtiaowang 
Because it is *mixed* and the declaration do not support mixed return type (yet).

driusan:
@johnsaigle conflicts, needs rebase

php/libraries/User.class.inc
HenriRabalais: Do you need the `?` for type hinting function parameters? I was under the impression that was only for return type declarations.

johnsaigle: @HenriRabalais I'm not actually sure about that... I don't know whether the Factory should accept null as a parameter even though the PHPDOC says that it should

driusan: `?type` means "null or type" while `type` means "must be type and only type"

johnsaigle:
This will be rebased once #4033 is merged.

johnsaigle:
@PapillonMcGill LORIS won't work if the parameter to the factory function isn't nullable. I'm not sure why but that's why this line was changed in the latest commit. See discussion here: https://github.com/aces/Loris/pull/3878#discussion_r211684049

HenriRabalais:
@johnsaigle I don't the suggestion is the make the factory function non-nullable, but rather have it default to null if no value is passed:
i.e. `public static function &factory(?string $username=null): \User`

PapillonMcGill:
@johnsaigle I'm not asking to make them `not nullable` but to do like line 131 `?string $username = null`
just like Henry wrote in the previous comment.

php/libraries/User.class.inc
```diff
@@ -29,24 +29,24 @@ class User extends UserPermissions
      * @var    array
      * @access private
      */
-    var $userInfo = array();
+    private $_userInfo = array();
```
driusan: protected

php/libraries/User.class.inc
```diff
@@ -260,12 +262,12 @@ class User extends UserPermissions
      *
      * @return string
      */
-    function getSiteNames()
+    function getSiteNames(): string
```
driusan: getSiteName*s* is plural, shouldn't this be an array of strings, not a string?

johnsaigle: #4118 Addresses this change and fixes the calling code.

driusan:
@johnsaigle conflicts

PR #4028, opened 2018-10-24
[Media] Allow access to the module when a user has only 'media_write'

johnsaigle:
### Brief summary of changes
Previously when a user had the media_write permission but not the media_read permission they could not access the module. Although this is probably a misconfiguration, it makes sense that if someone can upload files that they should also be able to browse them.

Also added a note about whether FileDownload should be a read permission or a write permission.

### To test this change...

- [ ] Give a user the `media_write` permission but not `media_read`. You should be able to view the module.

### DIFF
--- a/modules/media/ajax/FileDownload.php
+++ b/modules/media/ajax/FileDownload.php
@@ -15,6 +15,8 @@
  */
 
 $user =& User::singleton();
+//NOTE Should this be 'media_read' instead? It seems that downloading files
+//should be a read permission, not write.
 if (!$user->hasPermission('media_write')) {
     header("HTTP/1.1 403 Forbidden");
     exit;
--- a/modules/media/php/media.class.inc
+++ b/modules/media/php/media.class.inc
@@ -51,8 +51,7 @@ class Media extends \NDB_Menu_Filter
         // Set permission to access user-hidden files, currently based on superuser
         $this->hasHidePermission = $user->hasPermission("superuser");
 
-        return $this->hasPermission('media_read') ||  $this->hasWritePermission;
-            || $user->hasPermission('media_write');
+        return $this->hasPermission('media_read') || $this->hasWritePermission;
     }
 
     /**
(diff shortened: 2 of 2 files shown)


modules/media/php/media.class.inc
```diff
@@ -51,7 +51,8 @@ class Media extends \NDB_Menu_Filter
         // Set permission to access user-hidden files, currently based on superuser
         $this->hasHidePermission = $user->hasPermission("superuser");
 
-        return $user->hasPermission('media_read');
+        return $user->hasPermission('media_read')
```
PapillonMcGill:
```suggestion
        return $this->hasPermission('media_read') ||  $this->hasWritePermission;
```
No need to call the function again and do all the calculation in double

modules/media/ajax/FileDownload.php
```diff
@@ -15,6 +15,8 @@
  */
 
 $user =& User::singleton();
```
paiva: Should be `\User` to stay consistent

johnsaigle: It's out of scope for this PR but I'll create an issue to address this generally.

johnsaigle: @paiva #4030

modules/media/php/media.class.inc
```diff
@@ -51,7 +51,7 @@ class Media extends \NDB_Menu_Filter
         // Set permission to access user-hidden files, currently based on superuser
         $this->hasHidePermission = $user->hasPermission("superuser");
 
-        return $user->hasPermission('media_read');
+        return $this->hasPermission('media_read') || $this->hasWritePermission;
```
driusan: $this->hasPermission doesn't appear to be defined

PapillonMcGill: should be $user->hasPermission('media_read');

probably from my copy/pasted suggestion

johnsaigle:
@driusan My mistake. Should be fixed now!

PR #4029, opened 2018-10-25
[NDB_Page] addBasicDate() use dateOptions by default

davidblader:
addBasicDate should use `$this->dateOptions` by default if applicable


kongtiaowang:
ready to go

PR #4031, opened 2018-10-29
[Imaging Browser] Simplify permission checking

johnsaigle:
### Brief summary of changes
The permissions string for imaging browser is pretty wacky. This PR takes advantage of a [new function](#3966 ) added on the major branch to hopefully make the permissions a little more readable.

### To test this change...

- [ ] The module should function as before.

### DIFF
--- a/modules/imaging_browser/php/viewsession.class.inc
+++ b/modules/imaging_browser/php/viewsession.class.inc
@@ -51,8 +51,6 @@ class ViewSession extends \NDB_Form
         /* User has access if they have an 'all site' permission or if they are
         * part of a study site and are permitted to view their own site.
          */
-        $timePoint =& \TimePoint::singleton($_REQUEST['sessionID']);
-
         return $user->hasAnyPermission(
             array(
                 'imaging_browser_view_allsites',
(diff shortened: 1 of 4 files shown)


modules/imaging_browser/php/viewsession.class.inc
```diff
@@ -48,18 +48,23 @@ class ViewSession extends \NDB_Form
...
+                'imaging_browser_view_allsites',
+                'imaging_browser_phantom_allsites',
+                'imaging_browser_phantom_ownsite'
+            )
+        )
+        || (
+            in_array(
+                $timePoint->getData('CenterID'),
+                $user->getData('CenterIDs')
+            ) && $user->hasPermission('imaging_browser_view_site')
```
driusan: I think hasCenterPermission was added after this code was written, so this can be simplified further by using that.

johnsaigle: Awesome, didn't know about that function. I'll include it.

ridz1208:
@davidblader please test

ridz1208:
@driusan please re-review

PR #4032, opened 2018-10-29
Template for Reactification of Index Files

HenriRabalais:
This Media file will serve as a template for the ongoing reactification of LORIS. 

### In this iteration of refactoring:
 1. I removed all uses of jQuery.
   - This was done by using `window.onload` as well as Javascript's `fetch()` API
 2. I moved Table Headers into React's jurisdiction by setting them as a state.
   - If table headers eventually become configurable, this may be something that is queried from a table, fetched and then set as state.

### Issues:
- Ideally, the filter form declaration will be under the jurisdiction of the jsx index file rather than the module php class. However, that will require significant refactoring to the Filter Form component itself, or some sort of wrapper component such as #3963. It will also require loading all the data necessary to populate the select or searchable dropdown filters.

Let me know if there is any more work that this PR should include, or if there are any suggestions for handling the Issues.


modules/media/jsx/mediaIndex.js
```diff
@@ -189,12 +190,12 @@ class MediaIndex extends React.Component {
   }
 }
 
-$(function() {
+window.onload = () => {
```
driusan:
```suggestion
window.addEventListener("load", () => {
```

modules/media/jsx/mediaIndex.js
```diff
@@ -189,12 +190,12 @@ class MediaIndex extends React.Component {
...
   const mediaIndex = (
     <div className="page-media">
-      <MediaIndex DataURL={`${loris.BaseURL}/media/?format=json`} />
+      <MediaIndex dataURL={`${loris.BaseURL}/media/?format=json`} />
     </div>
   );
 
   ReactDOM.render(mediaIndex, document.getElementById("lorisworkspace"));
-});
+};
```
driusan:
```suggestion
});
```

matches the change above.

modules/media/jsx/mediaIndex.js
```diff
@@ -12,7 +12,12 @@ class MediaIndex extends React.Component {
     this.state = {
       isLoaded: false,
       filter: {},
-      hiddenHeaders: ['Cand ID', 'Session ID', 'Hide File', 'File Type']
+      headers: {
```
driusan: These don't seem to change anywhere, maybe they should be a class property or a const instead of state?

HenriRabalais: The reason I'm setting them as a state is for the potential toggling of column visibility in the future. Based on our UX discussions it seems like column headers may become configurable and toggle-able in the future.

driusan: I don't know if that's a good idea.. it seems like it would teach people to misuse state instead of understanding it.

HenriRabalais: Hmm, okay. So i'll set both `headers.hidden` and `headers.all` to a const object?

johnsaigle:
Is this info documented somewhere permanent? There's probably value in creating a `docs/Reactifying-LORIS.md` with our ideal implementation model there.

modules/media/jsx/mediaIndex.js
HenriRabalais: @driusan what do you think?

HenriRabalais:
Closed in favour of #4054

PR #4033, opened 2018-10-29
[Core: User, My Preferences] Eliminate type juggling in password expiration functionality

johnsaigle:
### Brief summary of changes
After toggling the "Strict types" flag in the User class in another PR Travis gave the error `TypeError: mktime() expects parameter 5 to be integer, string given`.

This PR uses `strtotime()` to generate a password expiry date instead of `mktime()`. This way we don't have to juggle types. It also has the benefit of being more readable.

This function also standardizes the `$expiry` variable to be a string everywhere instead of a nullable string.

### Related
- [ ] Github? #4027

### To test this change...

- [ ] Make sure you can still change your password via My Preferences AND via Edit User.


johnsaigle:
@kongtiaowang Can you close and open this PR? Travis doesn't seem to be starting.

PR #4038, opened 2018-10-29
[Imaging Uploader] Simplify ajax scripts

johnsaigle:
### Brief summary of changes
Cleanup of ajax scripts used by the imaging uploader. 

Not sure whether any project uses read_log as it has had the wrong permission code for a long time, but it's spruced up now. We should probably discuss whether it's better to delete it.

### To Test
- [ ] These scripts should work the same as before.


modules/imaging_uploader/ajax/read_log.php
```diff
@@ -12,96 +12,71 @@
...
-/**
- * Create a temp file to avoid the original file to be truncated
- */
+// Create a temp file to avoid truncating the original file.
 $temp_data_source_file = $data_source_file . ".tmp";
 
-///file doesn't exist create it
+// Create new temp file for reading if it doesn't exist.
 if (file_exists($temp_data_source_file)) {
-    if (!file_exists($temp_data_source_file)) {
```
johnsaigle: lol

modules/imaging_uploader/ajax/read_log.php
```diff
@@ -12,96 +12,71 @@
...
-    ///////////////////////////////////////////////////////////////////////////
-    ////////////////find the file last changed/////////////////////////////////
-    ///////////////////////////////////////////////////////////////////////////
-    $files = glob($log_directory."/*MRI_upload*");
+    // Get the last file modified.
+    $files = glob($log_directory . "/*MRI_upload*");
     $files = array_combine($files, array_map("filemtime", $files));
     arsort($files);
-    $data_source_file = key($files);
-    if (!file_exists($data_source_file)) {
```
johnsaigle: Glob couldn't have grabbed it if it didn't exist

modules/imaging_uploader/ajax/getUploadSummary.php
```diff
@@ -85,12 +60,28 @@
...
+ * - uploadID must be present and be a positive integer
+ * - summary must be equal to 'true' or 'false' when present.
+ *
+ * @return bool Whether the above conditions apply.
+ */
+function validRequest(): bool
+{
+    if (empty($_POST['uploadId'])
+        || !is_numeric($_POST['uploadId'])
+        || ($_POST['summary'] !== 'true' && $_POST['summary'] !== 'false')
```
PapillonMcGill: doesn't check for uploadID <= 0

modules/imaging_uploader/ajax/read_log.php
```diff
@@ -12,96 +12,71 @@
...
-        $data = fread($file, filesize($temp_data_source_file));
-        ftruncate($file, 0);//truncate the file
-        fflush($file); //writes the buffered output to the file
-        flock($file, LOCK_UN);  //locks the file
-        echo $data;
-    } else {
-        echo "could not lock file";
-    }
+$contents = file_get_contents($temp_data_source_file);
+if (!$contents) {
```
PapillonMcGill: Doesn't remove the temp file. This will fill the disk rapidly with unneeded file
Is it still needed with this new algorithm?

johnsaigle: I don't know whether it was needed in the first place. Perhaps it would be better to remove the temp file entirely and simply read the original?

PapillonMcGill: Probably yes, I can't see a reason to make a copy and load the copy.
Old code was actually emptying the content of the files (without remove it)

PR #4041, opened 2018-10-30
[Core] Remove deprecated Message class

johnsaigle:
### Brief summary of changes
As discussed in LORIS meeting, this class doesn't seem to be used anywhere so let's get rid of it.

I ran `grep -rn "new Message("` on both the `php/libraries/` and `modules/` folder which showed no results.


johnsaigle:
@sruthymathew123 @ridz1208 Please let me know if your projects are using this class.

PR #4046, opened 2018-10-31
[Core: Utility]  Add type declarations to function signatures

johnsaigle:

This PR also fixes problematic calling classes in the statistics module.

### DIFF
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -38,17 +38,18 @@ class Utility
      * @param string $dob      date of birth (or first date), formatted YYYY-MM-DD
      * @param string $testdate date of testing (or second date), formatted YYYY-MM-DD
      *
-     * @return mixed an array with keys 'year', 'mon', 'day' containing the
-     *               computed age, or a PEAR::Error object
+     * @return array containing keys 'year', 'mon', 'day' containing the
+     *               computed age
+     * @throws LorisException
      */
-    static function calculateAge($dob, $testdate)
+    static function calculateAge($dob, $testdate): array
     {
         if (!preg_match(
             "/([0-9]{4})-?([0-9]{1,2})-?([0-9]{1,2})/",
             $dob,
             $matches
         )) {
-            throw new Exception(
+            throw new \LorisException(
                 "Argument 1 does not match expected date format (YYYY-MM-DD)"
             );
         }
@@ -63,7 +64,7 @@ class Utility
             $testdate,
             $matches
         )) {
-            throw new Exception(
+            throw new \LorisException(
                 "Argument 2 does not match expected date format (YYYY-MM-DD)"
             );
         }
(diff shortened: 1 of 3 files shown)


php/libraries/Utility.class.inc
```diff
@@ -38,17 +38,18 @@ class Utility
...
      * @param string $testdate date of testing (or second date), formatted YYYY-MM-DD
      *
-     * @return mixed an array with keys 'year', 'mon', 'day' containing the
-     *               computed age, or a PEAR::Error object
+     * @return array containing keys 'year', 'mon', 'day' containing the
+     *               computed age
+     * @throws LorisException
      */
-    static function calculateAge($dob, $testdate)
+    static function calculateAge($dob, $testdate): array
```
PapillonMcGill: ...(string $dbo, string $testdate)...

php/libraries/Utility.class.inc
```diff
@@ -117,9 +118,9 @@ class Utility
      *
      * @return array an associative array("center ID" => "site name")
      */
-    static function getSiteList($study_site = true)
+    static function getSiteList(bool $study_site = true): array
```
PapillonMcGill: bool|null in the doc block

php/libraries/Utility.class.inc
```diff
@@ -150,9 +151,11 @@ class Utility
      *         should be interpreted as an associative array since the keys
      *         refer to the centerID, not the array index.
      */
-    static function getAssociativeSiteList($study_site = true, $DCC = true)
-    {
-        $DB =& Database::singleton();
+    static function getAssociativeSiteList(
+        bool $study_site = true,
+        bool $DCC = true
+    ): array {
```
PapillonMcGill: same here

php/libraries/Utility.class.inc
```diff
@@ -425,14 +430,14 @@ class Utility
...
      *
-     * @param integer $age   age in days
-     * @param string  $stage study stage (screening or visit)
+     * @param integer     $age   age in days
+     * @param string|null $stage study stage (screening or visit)
      *
      * @return array list of instruments
      */
-    static function lookupBattery($age, $stage=null)
+    static function lookupBattery(int $age, string $stage = null): array
```
PapillonMcGill: **?** string $stage

php/libraries/Utility.class.inc
johnsaigle: As discussed in the other PR, this param isn't actually nullable.

php/libraries/Utility.class.inc
johnsaigle: same here. :)

php/libraries/Utility.class.inc
```diff
@@ -216,12 +219,12 @@ class Utility
...
      * Returns a list of study Subprojects
      *
-     * @param integer $projectID The project for which you would like
+     * @param integer|null $projectID The project for which you would like
      *                           to get the subprojects
      *
      * @return array       an associative array("SubprojectID" => "Subproject title")
      */
-    static function getSubprojectList($projectID = null)
+    static function getSubprojectList(int $projectID = null): array
```
PapillonMcGill: I know Travis is happy but should't it be `?int $projectID = null` (same for line 257)?

PR #4047, opened 2018-10-31
[Core: Candidate]  Add type declarations to function signatures

johnsaigle:

### DIFF
--- a/modules/new_profile/php/new_profile.class.inc
+++ b/modules/new_profile/php/new_profile.class.inc
@@ -2,7 +2,7 @@
 /**
  * New_profile
  *
- * PHP Version 5
+ * PHP Version 7
  *
  * @category Main
  * @package  Loris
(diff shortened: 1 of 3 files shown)


php/libraries/Candidate.class.inc
```diff
@@ -329,7 +333,7 @@ class Candidate
      *
      * @return int SessionID of candidate's nth visit, NULL if none found
      */
-    function getSessionID($visitNo)
+    function getSessionID(int $visitNo): int
```
PapillonMcGill: return can be null

php/libraries/Candidate.class.inc
```diff
@@ -844,17 +855,21 @@ class Candidate
...
-     * @param array  $idStructure The structure defined in the config file in the
+     * @param array       $idStructure The structure defined in the config file
+     *                          in the array format returned by NDB_Config.
      *                            array format returned by NDB_Config.
-     * @param string $siteAbbrev  The candidate being validate's site abbreviation,
-     *                            so that it can be inserted into relevant "site"
-     *                            parts of the regex generated.
+     * @param string|null $siteAbbrev  The candidate being validate's site
+     *                  abbreviation, so that it can be inserted into relevant
+     *                            "site" parts of the regex generated.
```
PapillonMcGill: should fix indentation

johnsaigle:
Blocked by #4066

php/libraries/Candidate.class.inc
```diff
@@ -596,7 +596,7 @@ class Candidate
      *               or NULL if none exist
      * @throws DatabaseException
      */
-    public function getValidSubprojects(): array
+    public function getValidSubprojects(): ?array
```
driusan: Since this is the major branch why not just update the code to take out the null case and return an empty array instead?

johnsaigle: Sure I can do that.

modules/new_profile/php/new_profile.class.inc
```diff
@@ -43,62 +43,42 @@ class New_Profile extends \NDB_Form
...
     }
     /**
-     * Processes the values and saves to database
+     * Processes form data and creates a new candidate.
      *
      * @param array $values form values
      *
      * @return void
      */
-    function _process($values)
```
johnsaigle: This function was causing a lot of bugs with regard to typing so I rewrote it to use strict types and a clearer style of operating.

PR #4048, opened 2018-10-31
[Core: SinglePointLogin] Add type declarations to function signatures

johnsaigle:
See #3878 for a better description.


php/libraries/SinglePointLogin.class.inc
```diff
@@ -248,8 +248,12 @@ class SinglePointLogin
      * @return bool true if the username and password are valid
      * @access public
      */
-    function passwordAuthenticate($username, $password, $redirect = true)
-    {
+    function passwordAuthenticate(
+        string $username,
+        string $password,
+        bool $redirect = true
```
PapillonMcGill: should indicate @param bool|null in doc block

PapillonMcGill: wright, when you give a default value (other than `null`), it is not `null`

PapillonMcGill: Objection lifted your honor.  But Travis is still objecting.

johnsaigle: I'm not sure that I understand. What I'm saying is that this variable should never be null so I don't think the suggestion to add `|null` is correct. It makes it seem that you are allowed to call the function like this:

```php
passwordAuthenticate('user', 'pass', null);
```
But this will give an error.

You are allowed to do this:
```php
passwordAuthenticate('user', 'pass');
```
but in this case `$redirect` is initially unset, not null. They're different in PHP. In this case the default argument will set it to a bool.

So there is no case where `$redirect` is allowed to be null, though it is optional.

PapillonMcGill: `bool $redirect = null` would have required 
`@param bool|null ....` and
`?bool $redirect = null`  

giving a default value of `true` make it never to be null

PapillonMcGill: About Travis, it is just a note that Travis failed (possibly not related, I haven't check)

johnsaigle:
For some reason this PR always times out on Travis although when I test it on my VM I have no trouble logging in. Any ideas?

PR #4049, opened 2018-10-31
[Core: TimePoint] Add type declarations to function signatures

johnsaigle:
See #3878 for a better description.

PR #4050, opened 2018-10-31
[Candidate, TimePoint] Refactor setData() to accept only array as argument

johnsaigle:
### Brief summary of changes

Previously the setData() function in both of these classes accepted arrays or two separate strings. This prevents being able to use return types in these classes.

These functions were refactored to accept key-value arrays as they are occasionally called with multiple values to update. Calling functions are changed to use this new method of passing arguments.

### Related

* #4049 
* #4047 

### To test this change...

- [ ] Modify data in for Candidates and TimePoints. Everything should work as before.

### DIFF
--- a/modules/new_profile/php/new_profile.class.inc
+++ b/modules/new_profile/php/new_profile.class.inc
@@ -70,7 +70,7 @@ class New_Profile extends \NDB_Form
         if ($config->getSetting('useProjects') == "true") {
             $candidate->setData(
                 array(
-                    'ProjectID' => $values['ProjectID']
+                 'ProjectID' => $values['ProjectID'],
                 )
             );
 
--- a/modules/new_profile/php/new_profile.class.inc
+++ b/modules/new_profile/php/new_profile.class.inc
@@ -63,41 +63,29 @@ class New_Profile extends \NDB_Form
         $centerIDs = \User::singleton()->getData('CenterIDs');
         if (count($centerIDs) > 1 || !isset($values['psc'])) {
             $values['psc'] = $centerIDs[0];
-
-            // get the candidate
-            $candidate =& \Candidate::singleton($candID);
-
-            if ($config->getSetting('useProjects') == "true") {
-                $candidate->setData(
-                    array(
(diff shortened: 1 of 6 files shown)


php/libraries/TimePoint.class.inc
```diff
@@ -330,29 +330,39 @@ class TimePoint
...
+     * @param array $newData Key-value pair(s) of the field(s) to change and its
+     *                              new value(s).
      *
-     * @return void (side-effect modifies session table.)
+     * @return bool true
+     *
+     * @throws DatabaseException
      */
-    function setData($var, $value = null)
+    public function setData(array $newData): bool
```
driusan: Is this called from anywhere that it can't be made protected?

johnsaigle:
@driusan I'm not sure but I'll change it to `protected` and see if phan etc. has any issues.

php/libraries/TimePoint.class.inc
johnsaigle: There are 3 instances in `next_stage` that use this function in a static context and are incompatible with `protected` according to phan.

modules/new_profile/php/new_profile.class.inc
```diff
@@ -90,7 +90,12 @@ class New_Profile extends \NDB_Form
         $candidate =& \Candidate::singleton($candID);
 
         if ($config->getSetting('useProjects') == "true") {
-            $candidate->setData('ProjectID', $values['ProjectID']);
+            $candidate->setData(
+                array(
+                 'ProjectID',
+                 $values['ProjectID'],
+                )
```
driusan: Why is this one not an associative array?

driusan:
@johnsaigle this has conflicts now

modules/new_profile/php/new_profile.class.inc
johnsaigle: You're right, it should be.

johnsaigle:
@driusan I've addressed your feedback and rebased. Please re-review when possible.

driusan:
@johnsaigle needs to be rebased now because of conflicts.

PR #4051, opened 2018-10-31
[Core: UserPermissions] Add type declarations to function signatures

johnsaigle:


php/libraries/UserPermissions.class.inc
```diff
@@ -217,7 +216,7 @@ class UserPermissions
         }
 
         // refresh the user permissions
-        $success = $this->setPermissions();
+        $this->setPermissions();
```
PapillonMcGill: variable $success it never used now.

johnsaigle:
I believe this PR is having errors that will be fixed by merging #4027. As a result I will add the `Blocked` tag for now and do further testing when that PR goes through.

PR #4007, opened 2018-10-16
UI Makeover <3

zaliqarosli:
### Brief summary of changes
As presented in a LORIS meeting a couple months back, here is the PR of the proposed UI/UX colour/elements changes. Since then, I have removed the big orange buttons but kept links as orange on hover.

### This PR includes changes to:
- [x] Menu bar
- [x] Filters + tables
- [x] Breadcrumbs
- [x] Buttons

### To test this change...
- Pick one or two of you favourite LORIS modules, and check out its new look!
- If something looks off, checkout `aces/minor` and compare the two to see if it's really an issue with this PR
- If you don't like something and have a suggestion to make it better, comment!

PR #4052, opened 2018-10-31
[CSS/Instrument_list] Install FontAwesome!

zaliqarosli:
### Brief summary of changes
Motivated by UI issues with the instrument panel from https://github.com/aces/Loris/pull/4007, I have installed Font Awesome as a dependency on LORIS and changed all icons on the instrument panel to fa icons.

### To test this change...

- [ ] Quickly check on minor branch what it currently looks like (instrument_list)
- [ ] Checkout this branch, run `npm install`
- [ ] Got to `instrument_list`, check that icons look cute, and the lists are rendered better
- [ ] Ensure that no functionality has changed

Before & after:
screenshot 2018-11-05 12 25 57screenshot 2018-11-05 12 25 27

### DIFF
--- a/htdocs/fontawesome
+++ b/htdocs/fontawesome
@@ -1,0 +1,1 @@
+./../node_modules/@fortawesome/fontawesome-free/
--- a/modules/instrument_list/php/instrument_list_controlpanel.class.inc
+++ b/modules/instrument_list/php/instrument_list_controlpanel.class.inc
@@ -406,7 +406,7 @@ class Instrument_List_ControlPanel extends \TimePoint
 
             // highlight the current option if it's the current status
             if ($currentStatus == $option) {
-                $this->tpl_data['status'][$i]['icon'] = 'selected';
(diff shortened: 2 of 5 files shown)


davidblader:
do we need an extra dependency for these icons? why not just use bootstrap glyphicons?

zaliqarosli:
@davidblader because we're trying to move away from bootstrap as being discussed in the UX SIG

htdocs/fontawesome
```diff
@@ -0,0 +1 @@
+./../node_modules/@fortawesome/fontawesome-free/
```
johnsaigle:
```suggestion
../node_modules/@fortawesome/fontawesome-free/
```
Do you need the first `./`? Seems redundant

zaliqarosli: no, probably not

zaliqarosli: @johnsaigle turns out we do need it

driusan:
Merging as discussed during today's LORIS meeting.

PR #4053, opened 2018-10-31
[Core: Project, ProjectDefault] Add type declarations to function signatures

johnsaigle:
See #3878 for more details. 

These two classes are done together so that their function signatures are compatible.

### DIFF
--- a/php/libraries/Project.class.inc
+++ b/php/libraries/Project.class.inc
@@ -3,7 +3,7 @@
  * This file contains a class which encapsulates the concept of a "Project"
  * in LORIS.
  *
- * PHP Version 5
+ * PHP Version 7
  *
  * @category Main
  * @package  Loris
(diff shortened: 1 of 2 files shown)


php/libraries/Project.class.inc
```diff
@@ -1,9 +1,9 @@
-<?php
+<?php declare(strict_types=1);
 /**
  * This file contains a class which encapsulates the concept of a "Project"
  * in LORIS.
  *
- * PHP Version 5
+ * PHP Version 7
```
paiva: Shouldn't it be 7.2?

PapillonMcGill: that is the minimum version of PHP require to run the code properly. Type declaration require php7.0.

kongtiaowang:
ready to go.

PR #4054, opened 2018-11-01
[Media] Reactification Template

HenriRabalais:
This Media file will serve as a template for the ongoing reactification of LORIS. 

### In this iteration of refactoring:
 1. I removed all uses of jQuery in the Media file
   - This was done by using `window.onload` as well as Javascript's `fetch()` API
 2. I moved Table Headers into React's jurisdiction by declaring them in the index file.
   - If table headers eventually become configurable, this may be something that is queried from a table, fetched and then set as state.
 3. I moved Filter Form Elements into React's jurisdiction by declaring them in the index file.
  - It is still necessary to fetch the necessary options for select elements from the serverside.
 4. Implemented @driusan's FilterableDataTable component which will handle filtering for all modules. 
 5. Refactored FilterForm component. This involved removing a lot of irrelevant code, such as query strings and all uses of jQuery.
 6. Altered StaticDataTable to take on the responsibility of dealing with not displaying hidden headers and data, as well as organizing the data in format that it useful for the `formatColumn()` function.

Let me know if there is any more work that this PR should include, or if there are any suggestions for handling the Issues.


jsx/FilterableDataTable.js
```diff
@@ -0,0 +1,52 @@
...
+      <div>
+        <FilterForm
+          name={this.props.name + '_filter'}
+          id={this.props.name + '_filter_form'}
+          columns={3}
+          filter={this.state.filter}
+          updateFilter={this.updateFilter}
+          clearFilter={this.clearFilter}
+        >
+          {this.props.formElements}
```
driusan: would it make sense for this to be `this.props.children` instead?

modules/media/jsx/mediaIndex.js
```diff
@@ -1,119 +1,174 @@
...
-    this.fetchData();
+    this.fetchData()
+      .then(() => this.setState({isLoaded: true}))
+      .catch((error) => console.error(error));
   }
 
   /**
-   * Retrieve data from the provided URL and save it in state
-   * Additionally add hiddenHeaders to global loris variable
+   * Retrive data from the provided URL and save it in state
```
driusan: Retrieve was right before

modules/media/jsx/mediaIndex.js
```diff
@@ -1,119 +1,174 @@
...
+   * Additionaly add hiddenHeaders to global loris vairable
```
driusan: Additionally was also right before

jsx/Form.js
```diff
@@ -780,7 +780,7 @@ class TextboxElement extends Component {
   }
 
   handleChange(e) {
-    this.props.onUserInput(this.props.name, e.target.value);
+    this.props.onUserInput(this.props.name, e.target.value, this.props.id, 'textbox');
```
HenriRabalais: These two changes in `Form.js` are what allow the `onElementUpdate` function within the `FilterForm` component to accept `id` and `type`. There is likely a way of doing this that doesn't necessitate changing Form.js, but I'm thinking that it may be useful to be able to access the `id` and `type` of components `onUserInput` loris-wide.

jsx/FilterForm.js
```diff
@@ -18,164 +18,49 @@ import Panel from 'Panel';
...
-        formElements[fieldName].onUserInput = this.onElementUpdate.bind(null, formElements[fieldName].type);
-        formElements[fieldName].value = this.queryString[queryFieldName];
-      }.bind(this));
-    }
+    const formChildren = React.Children.map(this.props.children, (child) => {
+      return React.cloneElement(
+        child,
+        {
+          ...child.props,
+          value: (this.props.filter[child.props.name]||{}).value,
```
davidblader: nitpick but could you put some spaces in `this.props.filter[child.props.name]||{}`

jsx/Form.js
```diff
@@ -328,7 +328,7 @@ class SelectElement extends Component {
       }
     }
 
-    this.props.onUserInput(this.props.name, value);
+    this.props.onUserInput(this.props.name, e.target.value, e.target.id, 'select');
```
davidblader: does using `e.target.value` negate the purpose of lines 322-329?

jsx/FilterForm.js
```diff
@@ -18,164 +18,49 @@ import Panel from 'Panel';
...
+   * @param {string} id
+   * @param {string} type
    */
-  onElementUpdate(type, fieldName, fieldValue) {
-    // Make sure key is of string type and value is of string or object type
-    // before sending them to querystring
-    if (typeof fieldName !== 'string' ||
-        (typeof fieldValue !== 'string' && typeof fieldValue !== 'object')) {
-      return;
+  onElementUpdate(name, value, id, type) {
```
davidblader: what's the point of passing `id`?

jsx/Form.js
davidblader: why does this use the `id` of the event target but `TextboxElement` uses the one passed by `props`?

modules/media/jsx/mediaIndex.js
HenriRabalais: lol

jsx/FilterableDataTable.js
HenriRabalais: It could be, I don't see why not.

jsx/Form.js
HenriRabalais: These are good questions.

jsx/FilterForm.js
HenriRabalais: For sure.

modules/media/jsx/mediaIndex.js
HenriRabalais: That's what I get when I don't copy and paste

modules/media/jsx/mediaIndex.js
```diff
@@ -1,119 +1,174 @@
...
-          <MediaUploadForm
-            DataURL={`${loris.BaseURL}/media/ajax/FileUpload.php?action=getData`}
-            action={`${loris.BaseURL}/media/ajax/FileUpload.php?action=upload`}
-            maxUploadSize={this.state.Data.maxUploadSize}
-          />
-        </TabPane>
-      );
-    }
+    const tabList = [{id: 'browse', label: 'Browse'}];
+    const uploadTab = () => {
```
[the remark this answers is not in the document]
HenriRabalais: If it beneficial when something is going to be conditionally rendered to put it as a function. That way the variable doesn't have to be declared outside the `if` statement like so:
```
    const uploadTab;                                               
      if (loris.userHasPermission('media_write')) {                             
        tabList.push({id: 'upload', label: 'Upload'});                          
        uploadTab = (                                                                
          <TabPane TabId={tabList[1].id}>                                       
            <MediaUploadForm                                                       
              DataURL={`${loris.BaseURL}/media/ajax/FileUpload.php?action=getData`}
              action={`${loris.BaseURL}/media/ajax/FileUpload.php?action=upload`}
              maxUploadSize={this.state.data.maxUploadSize}                     
            />                                                                  
          </TabPane>                                                            
        );                                                                      
      }                                                                         
    };      
```

jsx/Form.js
davidblader: makes sense

jsx/FilterableDataTable.js
HenriRabalais: I sent it as `this.props.children`

kongtiaowang:
When I test this PR all the react table have the issues.
react-dom.development.js:14805 The above error occurred in the <StaticDataTable> component:
    in StaticDataTable (created by DynamicDataTable)
    in DynamicDataTable

 Uncaught TypeError: Cannot read property 'indexOf' of undefined

That is why it failed in Travis.

kongtiaowang:
it is passing Travis.

jsx/Filter.js
```diff
@@ -0,0 +1,94 @@
...
+ *
+ */
+class Filter extends Component {
+  constructor(props) {
+    super(props);
+    this.onElementUpdate = this.onElementUpdate.bind(this);
+  }
+
+  /**
+   * Sets filter object and querystring to reflect values of input fields
```
zaliqarosli: leftover reference to something that was removed

modules/media/jsx/mediaIndex.js
```diff
@@ -1,121 +1,185 @@
...
+     {label: 'Uploaded By', display: true},
+     {label: 'Date Taken', display: true},
+     {label: 'Comments', display: true},
+     {label: 'Date Uploaded', display: true},
+     {label: 'File Type', display: false},
+     {label: 'Cand ID', display: false},
+     {label: 'Session ID', display: false},
+     {label: 'Hide File', display: false},
+     {label: 'Edit Metadata', display: true},
+    ];
```
HenriRabalais: `fields` and `headers` were placed as separate arrays because that's the way things had previously been setup. However, they don't need to be and can likely be placed in to a single array.

modules/media/jsx/mediaIndex.js
```diff
@@ -1,121 +1,185 @@
...
+  /**
+   * Modify behaviour of specified column cells in the Data Table component
+   *
+   * @param {string} column - column name
+   * @param {string} cell - cell content
+   * @param {object} row
+   *
+   * @return {*} a formated table cell for a given column
+   */
+  formatColumn(column, cell, row) {
```
xlecours: This is just a comment, an objective one: I prefer when functions have a single return statement.

modules/media/php/media.class.inc
```diff
@@ -108,31 +109,23 @@ class Media extends \NDB_Menu_Filter
...
+        $hiddenOptions = [];
         if ($this->hasHidePermission) {
-            // Unable to create checkbox using addCheckbox(), using addSelect for
-            // the time being
-            // select options
             $hiddenOptions = array(
-                              ""  => 'Both',
-                              "0" => 'Visible only',
-                              "1" => 'Hidden only',
+                              0 => 'Visible only',
```
xlecours: To display both, one needs to clear the filters?

HenriRabalais: Clear the filter or return it the blank value that already exists by default in the select dropdown.

jsx/DataTable.js
```diff
@@ -0,0 +1,579 @@
...
+
+    let hasFilters = (filterValuesCount !== 0);
+    if (filterMatchCount === 0 && hasFilters) {
+      return 0;
+    }
+
+    return (filterMatchCount === 0) ? tableData.length : filterMatchCount;
+  }
+
+  toCamelCase(str) {
```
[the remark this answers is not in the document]
HenriRabalais: As of now, yes. But I am trying to get rid of it in my next commit.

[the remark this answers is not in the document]
HenriRabalais: Done!

modules/media/jsx/mediaIndex.js
HenriRabalais: Okay, I will look into making that happen!

jsx/Filter.js
```diff
@@ -0,0 +1,113 @@
...
+  height: '100%',
+  title: 'Selection Filter',
+  clearFilter: function() {
+    console.warn('onUpdate() callback is not set!');
+  },
+};
+Filter.propTypes = {
+  filter: PropTypes.object.isRequired,
+  clearFilter: PropTypes.func.isRequired,
+  id: PropTypes.string.isRequired,
```
driusan: why is this required?

jsx/FilterableDataTable.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+}
+
+FilterableDataTable.defaultProps = {
+  columns: 3,
+};
+
+FilterableDataTable.propTypes = {
+  name: PropTypes.string.isRequired,
+  children: PropTypes.element.isRequired,
```
driusan: children is required but doesn't seem to be used anywhere in this code or passed to a sub-component

jsx/FilterableDataTable.js
```diff
@@ -0,0 +1,78 @@
...
+  data: PropTypes.object.isRequired,
+  headers: PropTypes.object.isRequired,
```
driusan: same problem as children

php/libraries/NDB_Menu_Filter.class.inc
```diff
@@ -764,8 +764,9 @@ class NDB_Menu_Filter extends NDB_Menu
         }
 
         return array(
-                'Headers' => $headers,
-                'Data'    => $MappedData,
+                'Headers'      => $headers,
+                'Data'         => $MappedData,
+                'fieldOptions' => $this->fieldOptions,
```
driusan: I'm still not crazy about this mixing of concerns for presentation and data, but this PR is still a step forward so I'm not going to hold it up over this..

HenriRabalais: I think in the next round of refactoring we can move to request the options in a separate call.

kongtiaowang:
@HenriRabalais 
Fix [cnv_browser.class.inc cpg_browser.class.inc snp_browser.class.inc] in the same way add [public $fieldOptions;]

PR #4055, opened 2018-11-02
[Core] Adding CandID Value Object

xlecours:
This adds a CandID value object with unit test and a simple usage in the RestAPI.

The goal is to start using objects instead of array or scalar values to migrate toward a dependencies injection pattern. This will facilitate unit tests and further developments.

To test this change send a http request to the /api/v0.0.3-dev/candidates/<your-favorite-candid>


src/StudyEntities/Candidate/CandID.php
```diff
@@ -0,0 +1,20 @@
...
+class CandID extends ValidatableIdentifier
+{
+    public function getType(): string
+    {
+        return 'CandID';
+    }
+
+    static protected function validate(string $value): bool
+    {
+        return preg_match('/[0-9]{6}/', $value) === 1;
```
paiva: Does this validates all conditions for CandIDs?  If you have 000001, it would be validated I think, but I am not sure if it is considered a valid CandID.

src/StudyEntities/Candidate/ValidatableIdentifier.php
```diff
@@ -0,0 +1,18 @@
...
+
+abstract class ValidatableIdentifier implements Identifier
+{
+    protected $value;
+
+    abstract static protected function validate(string $value): bool;
+
+    public function __construct(string $value)
+    {
+        if (!static::validate($value)) {
```
paiva: `\static` instead of `static`?

src/StudyEntities/Candidate/CandID.php
xlecours: @paiva A CandID is a string of 6 digits. This does not checks for unicity or other business logic related to state.

driusan: I think this is only intended to validate the statically analyzable / structure of the CandID. Things that require external dependencies (such as needing a database connection to check if it's valid) shouldn't be done as part of the model's validation.

src/StudyEntities/Candidate/ValidatableIdentifier.php
driusan: I could be mistaken, but I don't think language keywords are affected by namespacing.

src/StudyEntities/Candidate/ValidatableIdentifier.php
```diff
@@ -0,0 +1,18 @@
+<?php
```
driusan:
```suggestion
<?php declare(strict_types=1);
```

src/StudyEntities/Candidate/CandID.php
```diff
@@ -0,0 +1,20 @@
+<?php
```
driusan:
```suggestion
<?php declare(strict_types=1);
```

.phan/config.php
```diff
@@ -59,6 +59,7 @@
 	],
 	"exclude_analysis_directory_list" => [
 		"vendor",
+            "test/unittests/studyentities/",
```
driusan: why is this excluded?

src/StudyEntities/Candidate/CandID.php
```diff
@@ -0,0 +1,20 @@
...
+namespace LORIS\StudyEntities\Candidate;
+class CandID extends ValidatableIdentifier
+{
+    public function getType(): string
```
driusan: Comment here about how a CandID's type is always the string "CandID"

src/StudyEntities/Candidate/CandID.php
```diff
@@ -0,0 +1,20 @@
...
+    {
+        return 'CandID';
+    }
+
+    static protected function validate(string $value): bool
```
driusan: Comment here about what does/doesn't get validated (to address confusion like @paiva's question)

src/StudyEntities/Candidate/Identifier.php
```diff
@@ -0,0 +1,39 @@
+<?php
```
driusan:
```suggestion
<?php declare(strict_types=1);
```

src/StudyEntities/Candidate/ValidatableIdentifier.php
```diff
@@ -0,0 +1,18 @@
...
+namespace LORIS\StudyEntities\Candidate;
+
+abstract class ValidatableIdentifier implements Identifier
```
driusan: Comment here about what the contract for this class is.

src/StudyEntities/Candidate/ValidatableIdentifier.php
```diff
@@ -0,0 +1,18 @@
...
+{
+    protected $value;
+
+    abstract static protected function validate(string $value): bool;
```
driusan: comment here about what this is expected of validate, and maybe an explanation for the reasons of it being 'static validate(string)` and not `validate()`

test/unittests/studyentities/CandID_Test.php
```diff
@@ -0,0 +1,73 @@
...
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+class CandID_Test extends TestCase
+{
+    public function testConstructorEmptyValue(): void
+    {
+        $this->expectException(\ArgumentCountError::class);
+        $candid = new CandID();
+    }
```
driusan: Adding the strict typing that I suggested would make this test meaningless (and invalid?)

src/StudyEntities/Candidate/CandID.php
```diff
@@ -0,0 +1,64 @@
...
+ * @category StudyEntities
+ * @package  LORIS
+ * @author   Xavier Lecours Boucher <xavier.lecoursboucher@mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+namespace LORIS\StudyEntities\Candidate;
+
+/**
+ * A representation of a CandID object which is always a string of length 6
```
driusan:
```suggestion
 * A representation of a CandID object. A CandID is always a string of length 6.
```

I think this wording is slightly clearer.

src/StudyEntities/Candidate/CandID.php
```diff
@@ -0,0 +1,64 @@
...
+     * @return string Always 'CandID'
+     */
+    public function getType(): string
+    {
+        return 'CandID';
+    }
+
+    /**
+     * Validate that the value of the CandID is a string of length 6.
+     * This do not checks for unicity in the database or any other
```
driusan:
```suggestion
     * This does not check for uniqueness in the database or any other
```

src/StudyEntities/Candidate/Identifier.php
```diff
@@ -0,0 +1,39 @@
...
+ * @category Model
+ * @package  StudyEntities
+ * @author   Xavier Lecours Boucher <xavier.lecoursboucher@mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+interface Identifier
+{
+    /**
+     * Return the name of the indentifier type from this interface
```
driusan:
```suggestion
     * Return the name of the identifier type for this interface
```

src/StudyEntities/Candidate/ValidatableIdentifier.php
```diff
@@ -0,0 +1,66 @@
...
+     *
+     * @return bool True if the value format is valid
+     */
+    abstract protected function validate(string $value): bool;
+
+    /**
+     * Default constructor
+     *
+     * This is final to make sure that classes extending ValidatableIdentifier
+     * will call validate upon instanciation.
```
driusan:
```suggestion
     * will call validate upon instantiation.
```

I'm not convinced that it needs to be declared final, but the grammar in the comment should be fixed at any rate.

xlecours: For now, I think it is ok to assume that classes extending the ValidatableIdentifier should use its constructor. But I a willing to revisit that if need be.

xlecours:
`some grammar fixes as github suggestions...`
Thanks for that @driusan

src/StudyEntities/Candidate/Identifier.php
```diff
@@ -0,0 +1,39 @@
...
+ * @category Model
+ * @package  StudyEntities
+ * @author   Xavier Lecours Boucher <xavier.lecoursboucher@mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+namespace LORIS\StudyEntities\Candidate;
+
+/**
+ * PHP Version 7
```
driusan:
```suggestion
 * A Candidate\Identifier is something that it used to identify a candidate that
 * is part of a study.
 *
 * Instances of Identifier must have a type to distinguish what kind of identifier
 * it is (ie. CandID, PSCID) and must be serializable to a string.
```

xlecours:
> I don't have anywhere to add the comment, but is the composer.lock change supposed to be committed with this PR?

@driusan I removed the changes

driusan:
@ridz1208 can you review this so I can merge it?

src/StudyEntities/Candidate/CandID.php
```diff
@@ -0,0 +1,64 @@
...
+        return 'CandID';
+    }
+
+    /**
+     * Validate that the value of the CandID is a string of length 6.
+     * This does not check for uniqueness in the database or any other
+     * state related facts.
+     *
+     * This function is called by the contructor of ValidatableIdentifier
+     * to ensure that no CandID exists if it value is not valid.
```
johnsaigle:
```suggestion
     * to ensure that no CandID exists if its value is not valid.
```

test/unittests/studyentities/CandID_Test.php
```diff
@@ -0,0 +1,68 @@
...
+{
+    /*
+     * dataProviders for constructor invalid values
+     */
+    public function invalidValues(): array
+    {
+        return array(
+                ['A'],
+                [1],
+                [000000],
```
johnsaigle: I'm a little bit confused about the test cases. For example this case is invalid, but `'000000'` (string value of the same ID) is valid. This makes me think that CandIDs should be created via string representations of numbers, but below you use `123456` (not `'123456'`) to create a CandID.

I'm not sure what implications this has for CandIDs with leading 0s. PHP gets really weird about stuff like this, e.g. 

```php
'000000' == 0;
```
will return true
```php
'0123456' == 123456;
```
returns true, and
```php
'0123456' == 0123456
```
return false.

To avoid possible type issues I think it would be good to be really explicit about the types that are allowed here. Another thing we could do is enforce that a Candidate ID is at least `100000` which is the current method used in the `Candidate` class. Either way it should be documented. I'm sure similar issues will come up with PSCID/ExternalID.

johnsaigle: Otherwise this looks very good!

xlecours: @johnsaigle I don't mind adding a rule for `>= 100000`.
Regarding the test that calls the constructor with `123456`, the value is cast as a string in the function so there is no way, to my knowledge, to check that. I did some tests using 012345 and the value becomes the string "5349" which is the decimal value of octal 12345. It fails the preg_match but calling it with 0777777 would pass. Maybe Phan would cast that. do you know?

johnsaigle: I think treating CandID as always a string instead of an int is the way to go. We never actually do math with it so it doesn't need to be an int. The string representation won't remove leading `0`s or get cast to different numerical representations.

PR #4062, opened 2018-11-05
[PHPCS] Forbid closing tags in PHP-only files

johnsaigle:
### Brief summary of changes

No more `?>` at the end of files.

I believe this is an unofficial standard for our development process anyhow so we may as well use PHPCS to help us out.

This is only a suggestion so if there is a good reason not to include this please let me know.

Background on problems caused by PHP closing tags: https://stackoverflow.com/a/4499749


johnsaigle:
@driusan I think it's failing on API < 0.0.3 files. Should I change these or are they not to be touched?

Will change this to go to major.

driusan:
Closing tags shouldn't change the output/behaviour of the API, so they can be removed.

HenriRabalais:
Seems to have conflicts and needs a rebase.

htdocs/acknowledgements/acknowledgements.php
```diff
@@ -76,4 +76,4 @@
 
 echo "</html>";
 
-?>
+
```
HenriRabalais: Is there a reason to be keeping this extra space in all these files?

johnsaigle: I think the reason is that this is how PHPCBF corrects the old code.

driusan:
@johnsaigle conflicts here too

htdocs/acknowledgements/acknowledgements.php
driusan: The main reason to remove the closing tags is because then PHP isn't sensitive to whether there's extra space/lines at the end of the file or not, so it doesn't really matter. (with the closing tags it'll send any whitespace after the tag as a literal.)

zaliqarosli:
will approve after conflicts resolved

htdocs/feedback_mri_popup.php
```diff
@@ -161,3 +161,6 @@
 $smarty->display('feedback_mri_popup.tpl');
 
 ob_end_flush();
+
+exit;
```
zaliqarosli: should this be a part of this PR?

johnsaigle: No I don't think so. I think this happened as the result of a rebase or something. I'll fix it.

php/libraries/TimePoint.class.inc
```diff
@@ -939,4 +939,5 @@ class TimePoint
 
         return $candidate->getCandidateDoB();
     }
-}
+} // end class
```
driusan: not sure why this comment was added

driusan: (just a comment, not a blocker.. I'll merge once Travis passes..)

johnsaigle: I think a rebase error caused it to be added back

PR #4045, opened 2018-10-31
[Core: Site] Add type declarations to function signatures

johnsaigle:
### Description

See #3878 for a more detailed explanation.

PR #3971, opened 2018-09-21
[document_repository] Rewrite and Reactify 

kongtiaowang:
### To test 

It should be tested everything in the module based on the test plan.
todo: rewriting the test case.

newUI:
https://sketch.cloud/s/mOzgL

PR #4064, opened 2018-11-05
[PHPCS] Add type hinting to PHPCS rules

johnsaigle:
### Brief summary of changes
Require type hinting via PHPCS.

screen shot 2018-11-05 at 15 18 34

For a while now we have specified that all new code going into LORIS should have type hinting, including return types. Using a 3rd-party code standard we can enforce this using PHPCS. 

This will save [a](#4045) [lot](#4046 ) [of](#4047 ) [manual](#4048 ) [time](#4049 ) [doing](#3878 ) [PRs](#4027 ) and [code review](#3971 ) as **now PHPCBF can automatically generate type hints from the PHPDocs**.

Note that this library requires updating the PHPCS version we are using in composer. 

### To test this change...

- [ ] Checkout the branch, run `composer update` and run PHPCS on basically any file. It will offer to fix type hinting for you.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,12 +13,13 @@
         "zendframework/zend-diactoros": "^1.6"
     },
     "require-dev" : {
-        "squizlabs/php_codesniffer" : "2.5.*",
+        "squizlabs/php_codesniffer" : "3.3.2",
         "phpunit/phpunit" : "6.5.*",
         "phpunit/dbunit": "3.0.*",
         "facebook/webdriver" : "dev-master",
         "phpmd/phpmd": "^2.6",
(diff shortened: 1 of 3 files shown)


johnsaigle:
As discussed in the LORIS meeting, we will use Phan to check typing for now. See #4069.

This PR should be potentially revisited when the above Phan errors reported above are fixed. Phan can only require that calling locations of functions match the signatures and PHPDocs but it cannot enforce that every function have type hints. 

Because of this, this PR should be considered a more strict check than what Phan can do and so should be enforced at a later date when the above problems are solved.

PR #4066, opened 2018-11-05
[Next Stage, New Profile] Fix misc. errors and warnings that are blocking other PRs

johnsaigle:
### Brief summary of changes

Checks are addred to ensure that variables are defined well before being processed.

This PR addresses bugs that are causing problems in #4047 and #4049 and must be merged before they can progress.


modules/next_stage/php/next_stage.class.inc
```diff
@@ -185,7 +185,7 @@ class Next_Stage extends \NDB_Form
 
         // let user select subproject *only* if subprojectid is not > 0
         if (!($timePoint->getSubprojectID() > 0)) {
-            $list_of_objectives =& \Utility::getSubprojectList();
```
johnsaigle: This shouldn't be a reference.

PR #4069, opened 2018-11-06
[Phan] Add PhanUndeclaredTypeReturnType rule to static analysis config

johnsaigle:
### Brief summary of changes

This PR adds the PhanUndeclaredTypeReturnType rule which makes sure that the `return` docs are immediately followed by a type and not an english sentence.

I also made a bunch of small changes to other files because I had a newer version of phan installed at some point which caused other issues to appear.... so this PR will head some of those off by the time we actually update phan.

### DIFF
--- a/htdocs/api/v0.0.2/APIBase.php
+++ b/htdocs/api/v0.0.2/APIBase.php
@@ -143,9 +143,9 @@ abstract class APIBase
     /**
      * Calculate the ETag for this resource
      *
-     * @return string an ETag for this resource
+     * @return ?string an ETag for this resource
      */
-    abstract function calculateETag();
+    abstract function calculateETag(): ?string;
 
     /**
      * Handle a GET request
@@ -248,7 +248,7 @@ abstract class APIBase
      *
      * @param integer $code The program exit code
      *
-     * @return none but exits the running program
+     * @return void but exits the running program
      */
     function safeExit($code)
     {
--- a/htdocs/api/v0.0.2/APIBase.php
+++ b/htdocs/api/v0.0.2/APIBase.php
@@ -143,9 +143,9 @@ abstract class APIBase
     /**
      * Calculate the ETag for this resource
      *
-     * @return ?string an ETag for this resource
+     * @return string an ETag for this resource
      */
-    abstract function calculateETag(): ?string;
+    abstract function calculateETag();
 
     /**
      * Handle a GET request
@@ -248,7 +248,7 @@ abstract class APIBase
      *
      * @param integer $code The program exit code
      *
-     * @return void but exits the running program
+     * @return none but exits the running program
      */
     function safeExit($code)
     {
--- a/test/unittests/Loris_PHPUnit_Database_TestCase.php
+++ b/test/unittests/Loris_PHPUnit_Database_TestCase.php
@@ -89,7 +89,6 @@ abstract class Loris_PHPUnit_Database_TestCase extends TestCase
      *
      * @return \PHPUnit\DbUnit\Database\DefaultConnection
      */
-    /* @phan-suppress-next-line PhanUndeclaredTypeReturnType*/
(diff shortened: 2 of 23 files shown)


johnsaigle:
Many of the errors generated here should be solved by the PRs tracked by #4072

driusan:
What's the status of this? Would it be mergeable if you only added one of the checks instead of both?

johnsaigle:
The status is that it's plodding along. It depends a lot on the typing checks being merged. Both rules are causing issues. It's possible to split this into two PRs so that the changes are more bite-sized but neither could be merged now.

#4153 took care of some of them. https://github.com/aces/Loris/projects/11 will take care of more.

I am still rebasing this script periodically and monitoring the errors.

johnsaigle:
This PR is now only allowing `PhanUndeclaredTypeReturnType` through as opposed to `PhanTypeMismatchReturn`. The latter was generating a lot of false positives that I wasn't able to debug.

johnsaigle:
@driusan This is ready to be reviewed and merged assuming you're fine with the changes.

test/unittests/Loris_PHPUnit_Database_TestCase.php
```diff
@@ -89,6 +89,7 @@ protected function setUp()
      *
      * @return \PHPUnit_Extensions_Database_DB_DefaultDatabaseConnection
      */
+    /* @phan-suppress-next-line PhanUndeclaredTypeReturnType*/
```
driusan: Why is this here? It says above it the return type is PHPUnit_Extensions_Database_DB_DefaultDatabaseConnection

modules/genomic_browser/php/view_genomic_file.class.inc
```diff
@@ -146,7 +146,7 @@ class View_Genomic_File extends \NDB_Form
     * @param string      $type  type of the date
     * @param array       $array array containing the date
     *
-    * @return date if exists, if not an empty string
+    * @return int|string Date (Unix int timestamp) if exists, or an empty string
```
driusan: 🤦‍♂️

driusan: (I don't know why GitHub doesn't have a (:picard_facepalming: option))

test/unittests/Loris_PHPUnit_Database_TestCase.php
johnsaigle: I'm not sure, Phan seemed to have an issue with it at one stage. I'll go back and verify that this is the case.

johnsaigle: This issue is fixed in #4274 which is based on this PR. If you merge this one first then this line will be removed and the bug fixed.

htdocs/api/v0.0.2/APIBase.php
```diff
@@ -143,9 +143,9 @@ function handleETag()
     /**
      * Calculate the ETag for this resource
      *
-     * @return string an ETag for this resource
+     * @return ?string an ETag for this resource
      */
-    abstract function calculateETag();
+    abstract function calculateETag(): ?string;
```
driusan: I think this should always return a string (the empty string if there's no etag, rather than null), but as @xlecours says this is going to be dead code once the API PR gets merged and the api directory is ignored by phan, so we don't need to do anything about this.

johnsaigle:
@xlecours I'll just revert the changes I made to the API files instead of blocking this.

johnsaigle:
Blocked by #4387

driusan:
@johnsaigle the diff on github says that this literally changed 0 lines of code.. can this be closed or is there something wrong with github?

driusan:
This PR contains no changes according to the diff on GitHub (https://github.com/aces/Loris/pull/4069/files)

Issue #4037, opened 2018-10-29
PHP 7.3: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"?


johnsaigle:
When we move to PHP 7.3 for LORIS, we'll need to update ambiguous references to `continue` in the source code or else be inundated by warnings.

See https://wiki.php.net/rfc/continue_on_switch_deprecation for more info.

### TODO

Change ambiguous `continue`s to `break`s

driusan:
Have you confirmed that any of these exist? As far as I know we don't have very many switches, let alone ones in loops.

johnsaigle:
I came across it once in `NDB_BVL_Instrument` and I think also in the `generate_tables_and_test_names` script.

johnsaigle:
#4073 Fixes one occurrence of this issue in the NDB_BVL_Instrument class. The issue still exists in the script.

Ahmed-Badawy:
the issue still exists for me. 
![image](https://user-images.githubusercontent.com/5174871/53416516-57caaf00-39dc-11e9-97cc-40d54556458b.png)

johnsaigle:
@Ahmed-Badawy Just to be clear, are you running this on a LORIS codebase? It looks like you're just running on old version of PHPUnit here.

woasimtalokdar:
Warning: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in G:\xampp\htdocs\wpoldversion\wp-content\themes\Divi\includes\builder\functions.php on line 6230

johnsaigle:
@woasimtalokdar This appears to be an issue with WordPress, not our software.

zimkha:
please I have the same problem with graphlql that I want to integrate in my project laravel, how to make for this error, I specify that I am in envrironement php 7.1.26

johnsaigle:
I think this has been addressed at some point. #4099

It's not showing up anymore in `make checkstatic` reports, anyway.

PR #4073, opened 2018-11-07
[Core: Instrument, Instrument_LINST, instrument_preview] Add type declarations to function signatures

johnsaigle:
See #3878 for more info.

This is the first one of these PRs done using #4064 so things might get weird.

This PR includes multiple classes that are all required to have the same function signatures.

It also fixes one instance of the Warning describe in #4037

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -165,17 +165,11 @@ class NDB_BVL_Instrument extends NDB_Page
      */
     static function factory(
         string $instrument,
-        string $commentID,
-        string $page,
+        string $commentID = '',
+        string $page = '',
         bool $guarantee_exists = true
-    ): ?\Loris\Behavioural\NDB_BVL_Instrument {
-
+    ) {
         $class = "NDB_BVL_Instrument_$instrument";
-
-        // Ensure 'page' is a string.
-        $page = $page ?? '';
-        // Ensure 'commentID' is a string.
-        $commentID = $commentID ?? '';
 
         // Make sure the instrument class has been included/required!
         $factory = NDB_Factory::singleton();
@@ -2388,7 +2382,7 @@ class NDB_BVL_Instrument extends NDB_Page
      *
      * @param array $element   The element from the QuickForm array renderer
      * @param array $formArray The currently existing form that was converted
-     *                          to instrument_json.tpl format.
+     *                         to instrument_json.tpl format.
      *
      * @return void (but side-effect modified $formArray to populate the element
      *         with the appropriate format needed for JSON serialization.)
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -752,7 +752,7 @@ class NDB_BVL_Instrument extends NDB_Page
             $values = Utility::nullifyEmpty($values, $key);
         }
         $db     = \Database::singleton();
-        $result = $db->update(
+        $db->update(
             $this->table,
             $values,
             array('CommentID' => $this->getCommentID())
@@ -1695,10 +1695,10 @@ class NDB_BVL_Instrument extends NDB_Page
      * Wrapper to create an Hour/Minute field, with an accompanying status field.
      *
      * @param string      $field        Name given to the HTML QuickForm Element
-     *                             being created
+     *                                  being created
      * @param string      $label        The question text to display
      * @param array       $rules        Additional rules to apply to the element
-     *                             being added,{@} delimited
+     *                                  being added,{@} delimited
      * @param string|null $rule_message Message to display upon rule violation.
      *
      * @return void
@@ -2071,7 +2071,7 @@ class NDB_BVL_Instrument extends NDB_Page
      * field in the database.
      *
      * @param array $date (optional) The date that the age is to
-     *                          be calculated based on.
+     *                    be calculated based on.
      *
      * @return array The age of the candidate in an array containing years, months
      *         and days.
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -189,7 +189,7 @@ class NDB_BVL_Instrument extends NDB_Page
             && !file_exists($base."project/instruments/$instrument.linst")
             && $guarantee_exists==false
         ) {
-            return;
+            return null;
         }
         if (file_exists($base."project/instruments/$instrument.linst")
         ) {
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -168,7 +168,8 @@ class NDB_BVL_Instrument extends NDB_Page
         string $commentID,
         string $page,
         bool $guarantee_exists = true
-    ): \Loris\Behavioural\NDB_BVL_Instrument_LINST {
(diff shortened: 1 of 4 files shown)


php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -736,7 +741,7 @@ class NDB_BVL_Instrument extends NDB_Page
      *
      * @return void
      */
-    function _save($values)
+    function _save(array $values): void
```
PapillonMcGill: could remove the variable assignation on line 779.  unsafeUpdate return `void`

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1391,7 +1406,7 @@ class NDB_BVL_Instrument extends NDB_Page
      *
      * @return Integer The sessionID
      */
-    function getSessionID()
+    function getSessionID(): Integer
```
PapillonMcGill: integer or bool (false, line 1419)

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2207,7 +2227,7 @@ class NDB_BVL_Instrument extends NDB_Page
      *
      * @return string
      */
-    function getFeedbackPanel($candID, $sessionID)
```
PapillonMcGill: shouldn't these 2 parameters be `integer`?

php/libraries/NDB_BVL_Instrument.class.inc
johnsaigle: I am going to change this to return `-1` instead. SessionID in other classes should always be an int going forward.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -189,7 +189,7 @@ class NDB_BVL_Instrument extends NDB_Page
             && !file_exists($base."project/instruments/$instrument.linst")
             && $guarantee_exists==false
         ) {
-            return;
+            return null;
```
driusan: function signature isn't nullable

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1853,15 +1879,16 @@ class NDB_BVL_Instrument extends NDB_Page
      * Wrapper to create a field that only accepts a number, with an
      * accompanying status field.
      *
-     * @param string  $field   The database field in which the response
-     *                         will be stored
-     * @param string  $label   The question text to display
-     * @param unknown $options Does not appear to be used?
```
johnsaigle: Now removed.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2029,25 +2056,24 @@ class NDB_BVL_Instrument extends NDB_Page
...
+     * field in the database.
      *
-     * @param array $Date_taken (optional) The date that the age is to
+     * @param array $date (optional) The date that the age is to
      *                          be calculated based on.
      *
      * @return array The age of the candidate in an array containing years, months
      *         and days.
      */
-    function getCandidateAge($Date_taken = false)
```
johnsaigle: The bool case in this function was removed.

PapillonMcGill:
restarting Travis

kongtiaowang:
TypeError: Return value of NDB_BVL_Instrument::getCommentID() must be of the type string, null returned

php/libraries/NDB_BVL_Instrument.class.inc
johnsaigle: This function now does not have a return type as it must be able to return itself as well as child classes. The PHPDoc has been updated to show that it can return null in this case.

test/test_instrument/NDB_BVL_Instrument_testtest.class.inc
```diff
@@ -49,7 +49,7 @@ class NDB_BVL_Instrument_testtest extends NDB_BVL_Instrument
      * @return void
      * @access private
      */
-    function _setupForm()
+    function _setupForm(): void
```
PapillonMcGill: forgot function _main() at line 71

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -165,18 +165,12 @@ class NDB_BVL_Instrument extends NDB_Page
...
-    ): ?\Loris\Behavioural\NDB_BVL_Instrument {
-
+    ) {
         $class = "NDB_BVL_Instrument_$instrument";
 
-        // Ensure 'page' is a string.
-        $page = $page ?? '';
-        // Ensure 'commentID' is a string.
-        $commentID = $commentID ?? '';
-
```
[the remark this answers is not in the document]
PapillonMcGill: @ridz1208 ?

driusan:
This needs to be rebased.

php/libraries/NDB_BVL_Instrument_LINST.class.inc
```diff
@@ -630,14 +637,7 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
...
                     if ($addElements) {
-                        $options = array();
-                        if (!empty($pieces[3])) {
-                            $options['min'] = $pieces[3];
-                        }
-                        if (!empty($pieces[4])) {
-                            $options['max'] = $pieces[4];
-                        }
-                        $this->addNumericElement($pieces[1], $pieces[2], $options);
+                        $this->addNumericElement($pieces[1], $pieces[2]);
```
PapillonMcGill: Is this change Ok? No more `min` nor `max` in the element.

php/libraries/NDB_BVL_Instrument.class.inc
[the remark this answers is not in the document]
ridz1208: I am @ridz1208  and I approve this message

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1860,15 +1880,16 @@ class NDB_BVL_Instrument extends NDB_Page
...
+     * @param string $label The question text to display
      *
      * @return void
      */
-    function addNumericElement($field, $label, $options = null)
-    {
+    function addNumericElement(
+        string $field,
+        string $label
+    ): void {
```
johnsaigle: @driusan @PapillonMcGill re: your comments below, this function never uses the `$options` variable so I removed it from the function signature. Similarly I removed the calling code below because the values don't actually get used.

php/libraries/NDB_BVL_Instrument_LINST.class.inc
PapillonMcGill: @johnsaigle still not resolved. Please comment if this is the intent and why as it modify the functionality

johnsaigle: @PapillonMcGill As I commented above, the `$options` variable is never used in the `addNumericElements` function so I removed it from the function. I also modified the calling code here because we don't need to create an array that is never used.

PapillonMcGill: @johnsaigle Right, must still be under influence from the parties. Hadn't see that min and max are in $option.

php/libraries/NDB_BVL_Instrument.class.inc
ridz1208: @johnsaigle the parameter here should be a string not array

PR #4075, opened 2018-11-07
[Core: Candidate] Fix getFirstVisit method

davidblader:
### Brief summary of changes
ran into this while trying to run `assign_missing_instruments`
got:
`PHP Fatal error:  Uncaught DomainException: Attempt to use pselectRow on a query that returns multiple rows in /var/www/loris/php/libraries/Database.class.inc:764
Stack trace:
#0 /var/www/loris/php/libraries/Database.class.inc(892): Database->pselectRow('SELECT Visit_la...', Array)
#1 /var/www/loris/php/libraries/Candidate.class.inc(630): Database->pselectOne('SELECT Visit_la...', Array)
#2 /var/www/loris/tools/assign_missing_instruments.php(93): Candidate->getFirstVisit()
#3 /var/www/loris/tools/assign_missing_instruments.php(148): populateVisitLabel(Array, 'Initial_Assessm...')
`
this updates the query to ensure a singular result is returned

### To test this change...

- [ ] try running `assign_missing_instruments.php`, also try creating & starting a visit

### DIFF
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -409,7 +409,7 @@ class Candidate
     function getProjectTitle()
     {
         $ProjectList = Utility::getProjectList();
-        return $ProjectList[$this->getProjectID()];
+        return $ProjectList[$this->getProjectID()] ?? '';
     }
 
     /**
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -622,12 +622,13 @@ class Candidate
 
         $candID = $this->getCandID();
         $query  = "SELECT Visit_label FROM session 
-          WHERE CandID = :cid AND Date_visit IS NOT NULL 
-          ORDER BY Date_visit, ID ASC LIMIT 1";
+          LEFT JOIN candidate USING (CandID)
+          WHERE CandID = :cid AND VisitNo = 1
+          AND Entity_type <> 'Scanner'";
         $where  = array('cid' => $candID);
-        $result = $db->pselectCol($query, $where);
-
-        return (is_array($result) && count($result) > 0) ? array_shift($result) : '';
+        $result = $db->pselectOne($query, $where);
+        echo $result . "\n";
+        return $result ?? '';
     }
 
     /**
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -627,7 +627,6 @@ class Candidate
           AND Entity_type <> 'Scanner'";
         $where  = array('cid' => $candID);
         $result = $db->pselectOne($query, $where);
-        echo $result . "\n";
         return $result ?? '';
     }
 
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -627,7 +627,7 @@ class Candidate
         $where  = array('cid' => $candID);
         $result = $db->pselectCol($query, $where);
 
-        return is_array($result) ? array_shift($result) : '';
+        return (is_array($result) && count($result) > 0) ? array_shift($result) : '';
     }
 
     /**
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -624,30 +624,18 @@ class Candidate
         // Visits may occur on the same date, so ordering by Date_Visit
         // or selecting by MIN(Date_Visit) will not produce a unique result
         // as required by pselectOne
-        // This query creates a row_num variable and simply selects whichever
-        // one comes first
-        /*$query = "SELECT Visit_label FROM (
-                    SELECT Visit_label,
-                    @r:=@r + 1 AS row_num
-                    FROM session s
-                    JOIN candidate c ON (c.CandID = s.CandID)
-                    CROSS JOIN (SELECT @r:=0)t
-                    WHERE c.CandID=:cid AND Date_Visit IS NOT NULL
-                    ORDER BY Date_Visit
-                    ) t1
-                    where row_num <= 1";*/
-        print_r($candID . "\n");
+        // This selects the smallest session ID, in the event that there
+        // are multiple visits starting on the same day
         $query  = "SELECT Visit_label
              FROM session 
-             WHERE ID = 
-             (SELECT MIN(ID) FROM 
-              (SELECT ID FROM session
+             WHERE ID =  
+              (SELECT MIN(ID) FROM session
                 WHERE Date_Visit =
                   (SELECT MIN(Date_Visit) FROM session s 
                     JOIN candidate c ON s.CandID = c.CandID
                     WHERE c.CandID=:cid AND Date_Visit IS NOT NULL
-                  ) AS T1 
-              ) AS T2 
+                  ) 
+                AND CandID=:cid
              )";
         $where  = array('cid' => $candID);
         $vLabel = $db->pselectOne($query, $where);
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -625,7 +625,7 @@ class Candidate
         $query  = "SELECT Visit_label
             FROM session s JOIN candidate c ON (c.CandID = s.CandID)
             WHERE c.CandID=:cid AND Date_Visit IS NOT NULL
-            ORDER BY Date_Visit";
+            ORDER BY Date_Visit LIMIT 1";
         $where  = array('cid' => $candID);
         $vLabel = $db->pselectOne($query, $where, false);
 
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -621,24 +621,11 @@ class Candidate
         $db      = $factory->database();
 
         $candID = $this->getCandID();
-        // Visits may occur on the same date, so ordering by Date_Visit
-        // or selecting by MIN(Date_Visit) will not produce a unique result
-        // as required by pselectOne
-        // This selects the smallest session ID, in the event that there
-        // are multiple visits starting on the same day
-        $query  = "SELECT Visit_label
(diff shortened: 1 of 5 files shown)


johnsaigle:
I just did some testing of your error and I was able to reproduce it. It's definitely a problem.

I don't think I like the approach of hamstringing `pselectOne` in this way. The limit clause is there to tell developers that they're using the function correctly -- in this case it's doing exactly what it's supposed to do.

My opinion would be to attempt a rewrite of the query. Even if it becomes fairly unreadable, that's what code comments are for.

Another possible approach would be to change the signature of `getFirstVisit` to return an array instead of a string. This way it will typically return a one-element array. In the case where two results appear (i.e. when a candidate does two visits on the same day), the calling code can determine which or both of the visit labels is appropriate to use.

In any case the real issue here is in the Candidate class and not in the Database class , so I don't think we should change the Database.

christinerogers:
I'd agree, leaning towards fixing getFirstVisit() - including the function description (and maybe even name?) -- given we're now plausibly allowing for multiple first visits to be returned.

php/libraries/Candidate.class.inc
```diff
@@ -625,9 +625,9 @@ class Candidate
         $query  = "SELECT Visit_label
             FROM session s JOIN candidate c ON (c.CandID = s.CandID)
             WHERE c.CandID=:cid AND Date_Visit IS NOT NULL
-            ORDER BY Date_Visit";
+            ORDER BY Date_Visit LIMIT 1";
```
PapillonMcGill: do not need 'LIMIT 1' LIMIT clause is implemented in pselectRow (called from pselectOne)

php/libraries/Database.class.inc
```diff
@@ -749,17 +749,27 @@ class Database
...
+    function pselectRow(
+        string $query,
+        array $params,
+        bool $efficiencyLimit = true
+    ) : ?array {
+
+        if ($efficiencyLimit) {
+            $query .= " LIMIT 2";
+        }
+        $rows = $this->pselect($query, $params);
```
PapillonMcGill: pselectRow select only one row (LIMIT *1*) at the end of the query. Do not know why it was LIMIT 2. This might be the source cause of your problem.

php/libraries/Database.class.inc
```diff
@@ -882,14 +892,17 @@ class Database
...
+     * @param bool   $efficiencyLimit Determines whether to use LIMIT clause in
+     *                                pselectRow
      *
      * @return mixed The value returned by the query.
      */
-    function pselectOne($query, $params)
+    function pselectOne($query, $params, $efficiencyLimit = true)
     {
-        $result = $this->pselectRow($query, $params);
+        $result = $this->pselectRow($query, $params, $efficiencyLimit);
```
PapillonMcGill: should not be needed as pselectRow implement a hard coded limit of 1and you always want the most efficient query in you database.

php/libraries/Database.class.inc
johnsaigle: It's supposed to be 2 in order to trigger the error. If it returns 2, we say "you used this wrong".

PapillonMcGill: @johnsaigle  need to update my VM, it as `LIMIT 1`.  and no code for exception.

Thanks for the info

php/libraries/Candidate.class.inc
```diff
@@ -621,11 +621,22 @@ class Candidate
...
+             FROM session 
+             WHERE ID =  
+              (SELECT MIN(ID) FROM session
+                WHERE Date_Visit =
+                  (SELECT MIN(Date_Visit) FROM session 
+                    WHERE CandID=:cid 
+                    AND Date_Visit IS NOT NULL
+                  ) 
+                AND CandID=:cid
+             )";
```
driusan: I think it would be better to just not use the pselectOne helper and grab everything for the CandID, then find the earliest date in PHP than to have a subselect within a subselect within a query.

PapillonMcGill: faster to execute directly in MySQL than in PHP. Less data to transmit between the 2 and less memory used in Apache

xlecours: The LIMIT clause can help for that.
What about : `SELECT Visit_label FROM session WHERE CandID = 595267 AND Date_visit IS NOT NULL ORDER BY Date_visit, ID ASC LIMIT 1`

johnsaigle:
Adding Needs Work based on the discussion in Dave's review.

kongtiaowang:
Travis error needs to be fixed.

php/libraries/Candidate.class.inc
```diff
@@ -613,22 +613,19 @@ class Candidate
...
-            FROM session s JOIN candidate c ON (c.CandID = s.CandID)
-            WHERE c.CandID=:cid AND Date_Visit IS NOT NULL
-            ORDER BY Date_Visit";
+        $query  = "SELECT Visit_label FROM session 
+          WHERE CandID = :cid AND Date_visit IS NOT NULL 
+          ORDER BY Date_visit, ID ASC LIMIT 1";
         $where  = array('cid' => $candID);
-        $vLabel = $db->pselectOne($query, $where);
-
+        $vLabel = $db->pselectCol($query, $where)[0] ?? null;
```
kongtiaowang: pselectCol will return an Associative array, so [0] will be always null.
pselectOne will return the value of the first column of the first row.
https://github.com/kongtiaowang/Loris/pull/168/files

php/libraries/Candidate.class.inc
```diff
@@ -613,23 +613,20 @@ class Candidate
...
-            ORDER BY Date_Visit";
+        $query  = "SELECT Visit_label FROM session 
+          WHERE CandID = :cid AND Date_visit IS NOT NULL 
+          ORDER BY Date_visit, ID ASC LIMIT 1";
         $where  = array('cid' => $candID);
-        $vLabel = $db->pselectOne($query, $where);
-
-        return $vLabel;
+        $result = $db->pselectCol($query, $where);
+        return array_shift($result);
```
johnsaigle: This is failing Travis when the value of `$result` is null because `array_shift` can only operate on `arrays`. 

I think you just need to do this:

`return is_array($result)? array_shift($result) : null;`

driusan: I went to look up the documentation to see if `array_shift($result ?? []);` would work too and it says it "Returns the shifted value, or NULL if array is empty or is not an array. "... so given that the documentation (seems to be?) wrong about the not an array bit, it's probably safer to do your check than rely on the rest of the documentation being correct. (I think PHPCS might insist on a space before the `?` though)

johnsaigle: $result in this case is also `null` so the `??` condition wouldn't go off. It only works for values that are not set, iirc.

driusan: `??` is the null coalesce operator. It works for unset or null value (but not other cases that would be considered "empty" in PHP).

johnsaigle:
Looks like the unit tests needs to be updated.

zaliqarosli:
i'm trying to use this fix to the script and i'm still getting this type error 
`PHP Fatal error:  Uncaught TypeError: Return value of Candidate::getFirstVisit() must be of the type string, null returned in /var/www/loris/php/libraries/Candidate.class.inc:647
Stack trace:
#0 /var/www/loris/tools/assign_missing_instruments.php(93): Candidate->getFirstVisit()
#1 /var/www/loris/tools/assign_missing_instruments.php(148): populateVisitLabel(Array, 'V3')
#2 {main}
  thrown in /var/www/loris/php/libraries/Candidate.class.inc on line 647
`
i'm not quite sure why the empty string still isn't returning :confused:

zaliqarosli:
@davidblader i just figured out that if pselectCol returns an empty array, getFirstVisit returns null. you need an empty() check!

php/libraries/Candidate.class.inc
```diff
@@ -621,15 +621,13 @@ class Candidate
...
-            ORDER BY Date_Visit";
+        $query  = "SELECT Visit_label FROM session 
+          WHERE CandID = :cid AND Date_visit IS NOT NULL 
+          ORDER BY Date_visit, ID ASC LIMIT 1";
         $where  = array('cid' => $candID);
-        $vLabel = $db->pselectOne($query, $where);
+        $result = $db->pselectCol($query, $where);
 
-        return $vLabel;
+        return is_array($result) ? array_shift($result) : '';
```
```suggestion
        return (is_array($result) && count($result) > 0) ? array_shift($result) : '';
```
I think this should address the bug @zaliqarosli noted.

zaliqarosli:
you're also going to have to add `return $ProjectList[$this->getProjectID()] ?? '';` to line [412](https://github.com/aces/Loris/blob/c6451a84a2bd92f7f6ae09fc4724686b0efa09af/php/libraries/Candidate.class.inc#L412)

PapillonMcGill:
@zaliqarosli @davidblader 
line 4232+ of Travis report (PHP 7.2)
There was 1 failure:
1) CandidateTest::testGetFirstVisitReturnsFirstVisitLabel
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'V01'
+''
/app/test/unittests/CandidateTest.php:362

the test at line 362 of CandidateTest expected to have 'V01' as result but received and empty string.
Hope it help

zaliqarosli:
@davidblader i had a lot of candidates in my RaisinBread with multiple timepoints of VisitNo=1, but i'm passing manual test on this assuming that the real data on projects are clean. Might be nice to add a try catch block in `assign_missing_instruments.php` that catches the DomainException `Attempt to use pselectRow on a query that returns multiple rows` if the pselectOne call fails and throws a new error `Candidate $candID has two first visits where we expect one. Please clean the data` or something like that to the console. but i won't block this PR on that request

PR #4084, opened 2018-11-13
[MediaUpload] Restrict file upload to candidate who has visited user's site

zaliqarosli:
### Brief summary of changes

Users may only upload files to a candidate for a visit label at a site that the user belongs to.

This PR:
- replaces #3697, which I will be closing. this PR should include all important changes from the old PR
- cleans up #3697 by removing unnecessary changes
- removes $candIdList query and variable because it's not ever used

**Note: Media module should be refactored to have a real LORIS endpoint i.e. remove ajax. this is out of the scope of this PR. changes made here should only be sufficient to address the permission issue.**

### This resolves issue...
- [x] Redmine? https://redmine.cbrain.mcgill.ca/issues/11681

### To test this change...
1. On superuser/all sites account, go to Media module 'Upload' tab. 'PSCID' select element should have all candidates.
2. On non-superuser account,
- assign a set of (not all) sites
- uncheck `Across all sites access candidate profiles`
- check both `Media files` permissions. 'PSCID' select element should only have candidates that have a visit label at a site the user belongs to.

### DIFF
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -212,7 +212,7 @@ function getUploadFields()
         $candquery    .= " WHERE FIND_IN_SET(s.CenterID, :cid) ORDER BY PSCID";
         $qparam['cid'] = implode(",", $user->getCenterIDs());
     } else {
-        $candquery    .= " ORDER BY PSCID";
+        $candquery .= " ORDER BY PSCID";
     }
     $candidates = $db->pselect(
         $candquery,
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -199,43 +199,32 @@ function getUploadFields()
     $db   = \NDB_Factory::singleton()->database();
     $user = \User::singleton();
 
-    $instruments = $db->pselect(
-        "SELECT Test_name FROM test_names ORDER BY Test_name",
-        []
-    );
     // Select only candidates that have had visit at user's sites
-    $qparam    = array();
-    $candquery = "SELECT DISTINCT PSCID FROM candidate c
-                  LEFT JOIN session s USING (CandID)";
+    $qparam       = array();
+    $sessionQuery = "SELECT c.PSCID, s.Visit_label, s.CenterID, f.Test_name
+                      FROM candidate c
+                      LEFT JOIN session s USING (CandID)
+                      LEFT JOIN flag f ON (s.ID=f.SessionID)";
 
     if (!$user->hasPermission('access_all_profiles')) {
-        $candquery    .= " WHERE FIND_IN_SET(s.CenterID, :cid) ORDER BY PSCID";
+        $sessionQuery .= " WHERE FIND_IN_SET(s.CenterID, :cid) ORDER BY c.PSCID ASC";
         $qparam['cid'] = implode(",", $user->getCenterIDs());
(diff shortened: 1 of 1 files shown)


zaliqarosli:
@ridz1208 @PapillonMcGill @nicolasbrossard @driusan please review as reviewers of the previous PR

modules/media/ajax/FileUpload.php
```diff
@@ -196,20 +196,31 @@ function viewData()
...
-    $candidates  = $db->pselect(
-        "SELECT CandID, PSCID FROM candidate ORDER BY PSCID",
-        []
+    // Select only candidates that have had visit at user's sites
+    $qparam    = array();
+    $candquery = "SELECT DISTINCT PSCID FROM candidate c
+                  LEFT JOIN session s USING (CandID)";
+
+    if (!$user->hasPermission('access_all_profiles')) {
+        $candquery    .= " WHERE FIND_IN_SET(s.CenterID, :cid) ORDER BY PSCID";
```
PapillonMcGill:
```suggestion
        $candquery    .= " WHERE s.CenterID in (:cid) ORDER BY PSCID";
```
`FIND_IN_SET` in to be used in a `SELECT` clause as it return the index where the value was found. In a `WHERE` clause use `WHERE <attribute> in (<list, of, values>)`

zaliqarosli: i believe we're supposed to be using find_in_set in prepared statements as discussed in the old PR https://github.com/aces/Loris/pull/3697#discussion_r196096201

PapillonMcGill: Haven't found the discussion about that it #3697. FIND_IN_SET is less efficient than WHERE x IN (values). This also work with prepared statement as long as your values are in a comma separated string which can be acheved with implode()

zaliqarosli: changing my code to what you suggested doesn't give me the expected query. i think the issue is that our Database class pselect function doesn't append the params how you want when using 'IN'

PapillonMcGill: Database class should not cause any problem. Will approve this PR and make appropriate change in #4085

zaliqarosli:
@ridz1208 does the second commit make you happy?

modules/media/ajax/FileUpload.php
```diff
@@ -196,35 +196,35 @@ function viewData()
...
-    $sessionRecords = $db->pselect(
-        "SELECT c.PSCID, s.Visit_label, s.CenterID, f.Test_name " .
-        "FROM candidate c ".
-        "LEFT JOIN session s USING(CandID) ".
-        "LEFT JOIN flag f ON (s.ID=f.SessionID) ".
-        "ORDER BY c.PSCID ASC",
-        []
-    );
-
+    $sessionData = [];
```
johnsaigle:
```suggestion
    $sessionData = array();
```

zaliqarosli: this is a small preference change that provides no difference in overhead and is not worth committing and overriding the existing approval. the syntax here is absolutely fine

johnsaigle: It's a standard we have for declaring arrays in PHP in LORIS, not my preference.

That said I agree it's not worth blocking the PR which is why I commented instead of requesting changes.

PR #4085, opened 2018-11-13
[General] replace FIND_IN_SET inside WHERE clause with IN (<set>)

PapillonMcGill:
### Brief summary of changes
Replace occurence of `FIND_IN_SET()` function from `WHERE` clause with `WHERE <attribute> IN (<set>)

### This resolves issue...
for efficiency in query

### To test this change...

- use all pages where instance of the modified files are used
specially the imaging_browser modules

### DIFF
--- a/htdocs/api/v0.0.3-dev/Candidates.php
+++ b/htdocs/api/v0.0.3-dev/Candidates.php
@@ -132,7 +132,7 @@ class Candidates extends APIBase
                 "CenterID"
             );
             $allUserSiteNames = $this->DB->pselectColWithIndexKey(
-                "SELECT CenterID, Name FROM psc WHERE FIND_IN_SET(CenterID, :cid)",
+                "SELECT CenterID, Name FROM psc WHERE CenterID IN (:cid)",
                 array('cid' => implode(',', $centerIDs)),
                 "CenterID"
             );
--- a/modules/user_accounts/php/user_accounts.class.inc
+++ b/modules/user_accounts/php/user_accounts.class.inc
@@ -54,7 +54,7 @@ class User_Accounts extends \NDB_Menu_Filter
             WHERE 1=1";
         if (!$user->hasPermission('user_accounts_multisite')) {
             $site_arr = implode(",", $user->getCenterIDs());
-            $query   .= " AND FIND_IN_SET(user_psc_rel.CenterID,'$site_arr')";
+            $query   .= " AND user_psc_rel.CenterID IN ('$site_arr')";
         }
 
         // set the class variables
--- a/htdocs/api/v0.0.3-dev/candidates/Visit.php
+++ b/htdocs/api/v0.0.3-dev/candidates/Visit.php
@@ -201,7 +201,7 @@ class Visit extends \Loris\API\Candidates\Candidate
                 );
                 $allUserSiteNames = $this->DB->pselectColWithIndexKey(
                     "SELECT CenterID, Name FROM psc ".
-                    "WHERE FIND_IN_SET(CenterID, :cid)",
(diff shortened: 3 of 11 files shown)


driusan:
What version of PHP/PDO fixed the bug which mandated `FIND_IN_SET` (or expanding to n placeholders..)?

PapillonMcGill:
@driusan  Was there a bug in the first place that mandate using FIND_IN_SET?

xlecours:
This can't work all the time. For string substitution it will work but not for bind parameters.
The `IN` operator will receive a string representation of the array e.g `'1,2,3'` and will cast it into a number `1,2,3` is not number but will be converted as `1`
```
mysql> select CAST('1,2,3' AS UNSIGNED);
+---------------------------+
| CAST('1,2,3' AS UNSIGNED) |
+---------------------------+
|                         1 |
+---------------------------+
```
This will not throw an exception but will return different results than with the FIND_IN_SET.

PapillonMcGill:
@xlecours When an array of number is passed, is it from IDs extracted from the DB?

In this case, we don't have to pass them as prepared statement but just concatenate the string in the clause.

htdocs/api/v0.0.3-dev/Candidates.php
```diff
@@ -132,7 +132,7 @@ public function handlePOST()
                 "CenterID"
             );
             $allUserSiteNames = $this->DB->pselectColWithIndexKey(
-                "SELECT CenterID, Name FROM psc WHERE FIND_IN_SET(CenterID, :cid)",
+                "SELECT CenterID, Name FROM psc WHERE CenterID IN (:cid)",
```
xlecours: Won't work for multisite users

modules/user_accounts/php/user_accounts.class.inc
```diff
@@ -54,7 +54,7 @@ class User_Accounts extends \NDB_Menu_Filter
             WHERE 1=1";
         if (!$user->hasPermission('user_accounts_multisite')) {
             $site_arr = implode(",", $user->getCenterIDs());
-            $query   .= " AND FIND_IN_SET(user_psc_rel.CenterID,'$site_arr')";
+            $query   .= " AND user_psc_rel.CenterID IN ('$site_arr')";
```
xlecours: This will work because it's not a bind parameter

htdocs/api/v0.0.3-dev/Candidates.php
PapillonMcGill: Since the $centerIDs variable is coming from data in the DB, we could do 
```... WHERE CenterID IN ". implode(",", $centerIDs ...```
Any objection for that?

xlecours:
I do not mind using IN instead of FIND_IN_SET. My only concern is that it must work in a consistent way. As for efficiency, there is bigger gain elsewhere I am sure.

htdocs/api/v0.0.3-dev/Candidates.php
driusan: Using string concatenation to build a query is much less secure and error prone than using a bind parameter.

PapillonMcGill: @driusan error prone yes, we would have to check if data is INT of STRING. 
Less secure? When it is a list of INT that where extracted from the DB, risk of injection are nil unless someone can manipulate the code, and in that case, prepared statement can be manipulated anyway or the DB was tempered.

It might not be a 50% overall increase in speed, but the gain of efficiency is not negligible.

PapillonMcGill:
closing due to very limited number of rows in the affected tables. little gain to benefit with risk of cut&paste without proper check introducing security holes.

PR #4089, opened 2018-11-13
[Core: CouchDB] Add type declarations to function signatures

johnsaigle:
See #3878 for background.

### DIFF
--- a/php/libraries/CouchDB.class.inc
+++ b/php/libraries/CouchDB.class.inc
@@ -573,10 +573,10 @@ class CouchDB
      * @param string $id  The document name to PUT
      * @param array  $doc An array representing the JSON document
      *
-     * @return string The JSON string returned by the CouchDB server, or
+     * @return string|null The JSON string returned by the CouchDB server, or
      *                null if in bulk mode.
      */
-    function putDoc(string $id, array $doc): string
+    function putDoc(string $id, array $doc): ?string
     {
         $encoded = json_encode($doc);
 
@@ -677,14 +677,15 @@ class CouchDB
      * @param boolean $full_json If true will include the full json returned,
      *                           and not just the parsed rows.
      *
-     * @return array An array of the rows returned by the view
+     * @return array|string An array of the rows returned by the view, or a
+     *                          JSON string.
      */
     function queryView(
         string $designdoc,
         string $viewname,
         array $params,
         bool $full_json = false
-    ): array {
+    ) {
         $keys = null;
         if (isset($params['keys'])) {
 
--- a/php/libraries/CouchDB.class.inc
+++ b/php/libraries/CouchDB.class.inc
@@ -427,7 +427,8 @@ class CouchDB
      *                            having to parse it server-side if the request is
      *                            expected to be directly interpreted by javascript.
      *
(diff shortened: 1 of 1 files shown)


php/libraries/CouchDB.class.inc
```diff
@@ -725,7 +740,7 @@ class CouchDB
      *
      * @return string The string that was returned by CouchDB
      */
-    function commitBulkTransaction()
+    function commitBulkTransaction(): string
```
driusan: the line right below this is "return false"

johnsaigle: Thanks! I generated this with the PHPCBF tool and it just pulls the return type from the PHPDOC.

php/libraries/CouchDB.class.inc
```diff
@@ -488,8 +495,12 @@ class CouchDB
...
      *         CouchDB web server.
      */
-    function _postRelativeURL($url, $data, $op = 'POST', $printValue = false)
-    {
+    function _postRelativeURL(
+        string $url,
+        string $data,
+        string $op = 'POST',
+        bool $printValue = false
+    ) {
```
HenriRabalais: I the page content returned in the form of a string? It would be useful to have a return type here.

johnsaigle: That makes sense to me but I can't confirm since I don't have experience with this library. @ridz1208 Maybe you know?

driusan: It's a thin wrapper around _postURL. postURL is declared as     `function _postURL(
        string $url,
        array $data,
        array $headers = array(),
        bool $printValue = false
    ): string`

.. so yes, it should be a string. (But the signature of _postURL is also wrong, because while looking it up I noticed at one place it returns a null.)

php/libraries/CouchDB.class.inc
```diff
@@ -243,7 +243,7 @@ class CouchDB
         int    $port = 0,
         string $user = "",
         string $password = ""
-    ) {
+    ): CouchDB {
```
PapillonMcGill:
```suggestion
    ):  \CouchDB {
```

php/libraries/CouchDB.class.inc
```diff
@@ -564,7 +576,7 @@ class CouchDB
      * @return string The JSON string returned by the CouchDB server, or
      *                null if in bulk mode.
      */
-    function putDoc($id, $doc)
+    function putDoc(string $id, array $doc): string
```
PapillonMcGill:
```suggestion
    function putDoc(string $id, array $doc): ?string
```
see line 586. Also change the DocBlock

php/libraries/CouchDB.class.inc
```diff
@@ -667,8 +679,12 @@ class CouchDB
...
      * @return array An array of the rows returned by the view
      */
-    function queryView($designdoc, $viewname, $params, $full_json=false)
-    {
+    function queryView(
+        string $designdoc,
+        string $viewname,
+        array $params,
+        bool $full_json = false
+    ): array {
```
PapillonMcGill: can also return a string. See line 711
also need to adjust DocBlock

PR #4093, opened 2018-11-14
Wiki Revamp

ridz1208:
The goal of this PR is to bring to the branch what belongs on the branch. The GitHub WIKI does not allow for a proper review process or any other sort of control over it's content. There is also only a single instance of the wiki which means that there can not be documentation for each of our releases on there at a time unless we manually version every single page and with more than 70 pages currently on there the maintenance of that wiki for all releases would be near impossible.

This PR serves as a proof of concept to try and accomplish the following
- clean the content of the current wiki
- consolidate all the documentation that already sit on the branch with what was on the wiki
- ensure the documentation is up to date with our current codebase
- provide an organized way of presenting the information

This PR contains an untouched copy of the current wiki as well as a proposed directory structure for the NEW WIKI

Feel free to issue PRs to my branch if you would like to propose alternate directory structures. This structure was designed by looking at the documentation we currently have and attempting to compartementalize the information. I have already done some work towards adding content in each of the directories, I will be submitting separate PRs for those and linking it to this one.


johnsaigle:
I am going to list a few general points here for reference. A lot of people shared a lot of opinions yesterday and I want to make sure their perspectives are represented

### Benefits of tracking documentation with Git
* More eyes on the documentation means reduced chances of typos, things falling out of sync, etc
* Potentially reduces duplication -- once things are codified here we can start culling old Wiki, GDocs, Redmine, etc. (though this is likely a long way off)
* Docs will match the release version that someone is using

### Things to consider
* **We have to be diligent about keeping this up-to-date**. This means a lot of attention given to `Documentation` PRs and also quick branch merges (e.g. between bugfix and minor)
* **The documentation that goes in the repo should be for the general developer**. References to MCIN internal IT policies should not be included
* **The new documentation has to be easily navigable and highly visible**. However, we want to avoid a snake pit of intertwining links. 
* **We should not be documenting how to install our dependencies** in detail. It's a hopeless task that we won't be able to keep up with. As this is a self-hosted web application there will always need to be a technical person involved in the install process. Let's assume they can get a LAMP stack running (but of course we'll list the necessary versions clearly).
* **Let's document our packaging/containerization efforts**. We have Vagrant, Docker, Heroku, and a forth-coming Neurodebian package. Do we need all of this? What are the differences? Are they all functional? If we can't keep them up to date we shouldn't offer them. 


### Potential future work
* Use [read the docs](https://readthedocs.org/) to create a wiki-like thing that is synchronized with the markdown files in the repo. This is auto versioned and easier to navigate than Markdown files
* Eventually replace the contents of the Wiki with links to the appropriate files.

johnsaigle:
@christinerogers @ridz1208 @zaliqarosli (others?) I believe you all have notes on hurdles encountered during the install process -- either user confusion or specific bug reports for things like permissions, etc. If you could summarize these issues and make a post here I think we have a good opportunity here to smooth out the bumps in the install process.

zaliqarosli:
@johnsaigle im sorry i'm a bit confused. this PR looks to create the directory structure for the new 'wiki-on-the-branch'. where does documentation on the dev install process fit here? i think it should be issued in a separate PR as @ridz1208 mentioned in this PR's description

johnsaigle:
@zaliqarosli Yeah you're right it doesn't belong here.

driusan:
@johnsaigle we don't have Docker builds anymore. It was removed when I discovered that it didn't build anymore and had been that way for a long time.

christinerogers:
Nice, Rida. 
Just quickly re directory structure -- Has there already been discussion of matching to the Loris-MRI documentation directory structure? e.g. https://github.com/aces/Loris-MRI/tree/minor/docs (or matching Loris-MRI to this structure?)

driusan:
I don't think there's any reason for a "wiki" subdirectory rather than making the base just the top level "docs" directory. (Of course, we'd have to figure out where to put the things that are already there..)

johnsaigle:
@christinerogers I don't think it's been brought up yet but that's a good idea. 

As the LORIS docs are likely to be longer and more complex it probably makes more sense to figure out how to present that in whatever way we need to. 

Down the road we can see if it makes sense for LORIS-MRI to take the same shape.

ridz1208:
@christinerogers I took a quick look and as @johnsaigle said, LORIS will be much more complex

@driusan yeah, I dont like the wiki subdirectory but I think its useful while it's being built (helps to tell whats new and what was already there before). I can remove it if you disagree

ridz1208:
@johnsaigle Thanks for the visual representation of the directory structure and the detailed background.

if everyone (@driusan @xlecours @zaliqarosli @christinerogers) is satisfied with this structure I suggest that we merge this and start progressively moving/fixing docs

ridz1208:
@driusan would it be a superbad idea to make this release independent ?? (i.e. put it on it's own loris branch ??)

ridz1208:
@driusan I meant the revamp itself making it release independent and then merging it into the live branch once it's done (to avoid having parts of it in different releases)

johnsaigle:
I can't see the new `docs/` file on `major`. Did this actually get merged? I'm confused.

ridz1208:
@johnsaigle the PR got merged accidentally into the wrong branch
a new branch will be created and This PR will be issued there

driusan:
@johnsaigle It was auto-merged when #4115 was merged, and then (both) were reverted because it wasn't supposed to be merged into major, it was supposed to be merged into a different branch.

johnsaigle:
Thanks

PR #4094, opened 2018-11-14
[CORE] fix to version 20.1

PapillonMcGill:
### Brief summary of changes
Fix the version number in <LORIS_dir>/VERSION

### To test this change...

open the file and check for correct text


johnsaigle:
@driusan Do you think it's worth making a PR to the minor branch with the `Makefile`? Feels like we've solved this problem already.

PR #4095, opened 2018-11-14
Refactor changes made in Reactification Template PR #4054

HenriRabalais:
This PR refactors a change made in #4054. Rather than place the $fieldOptions variable in each individual class that extends NDB_Menu_Filter, this PR places the variable in NDB_Menu_Filter itself.


driusan:
@HenriRabalais Failing Travis

PR #4110, opened 2018-11-15
[JS/Config] Remove unnecessary eslint config from htdocs directory

driusan:
This removes an .eslint file that was in the htdocs directory for
some unknown reason.

It doesn't touch the main eslint file in the root directory.

(lintian was complaining about it while trying to build a debian package,
but besides that I don't know why it's there overriding the default.)

### DIFF
--- a/htdocs/js/.eslintrc.json
+++ b/htdocs/js/.eslintrc.json
@@ -1,5 +1,0 @@
-{
-  "parserOptions": {
-    "sourceType": "script"
-  }
-}


johnsaigle:
It looks like the file was being used by the linter and maybe had different rules. The two errors reported by Travis seem easy to fix though.

driusan:
Travis says it's necessary, so I'll just close this for now and we can revisit later.

PR #4118, opened 2018-11-16
[Core: User] getSiteNames now returns array instead of a string

johnsaigle:
### Brief summary of changes

The old function returned a string concatenated with semi-colons which is confusing and wasn't well documented. Now the function returns an array as you'd expect. I updated the calling code.

In updating the calling code I found that the function `mergeSettings()` in `NDB_Config` probably was never working correctly with multisites. `Utility::getCleanString()` would just remove the semicolons in the siteNames array, e.g. `dcc;ottawa;montreal;rome` --> `dccottawamontrealrome`. 

I don't think this was ever a valid array key which means the whole function is essentially pointless. The git blame shows that this `getCleanString` call hasn't been updated in 11 years so it's likely something we can remove.

If Travis doesn't complain I think this change should be safe.

### This resolves issue...

- [ ] Github? https://github.com/aces/Loris/pull/4027#discussion_r233911632

### To test this change...

Everything should work the same.

### DIFF
--- a/php/libraries/NDB_Config.class.inc
+++ b/php/libraries/NDB_Config.class.inc
@@ -39,13 +39,6 @@ class NDB_Config
     var $_settings = array();
 
     /**
-     * The merged array of study and site-specific settings
-     *
-     * @access private
-     */
-    var $_siteSettings = array();
-
-    /**
      * An optional override for the location of the config file.
      * (default is ../project/config.xml)
      *
--- a/php/libraries/NDB_Config.class.inc
+++ b/php/libraries/NDB_Config.class.inc
@@ -239,43 +239,6 @@ class NDB_Config
     }
 
     /**
-     * Attempts to determine the site of the user currently logged in and uses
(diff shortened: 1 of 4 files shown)


php/libraries/NDB_Config.class.inc
```diff
@@ -238,43 +238,6 @@ class NDB_Config
...
-        }
-
-        // make sure a user has logged in
-        $username = $_SESSION['State']->getUsername();
-        if (empty($username)) {
-            $this->_siteSettings =& $this->_settings;
-        } else {
-            $user     =& User::singleton($username);
-            $siteName = Utility::getCleanString($user->getSiteNames());
-            if (isset($this->_settings['sites'][$siteName])
```
johnsaigle: In my tests this condition was never true, meaning the function simply does line 273, 

```php
$this->_siteSettings =& $this->_settings;
```

Meaning that _siteSettings is always the same as _settings and therefore is unnecessary.

PR #4119, opened 2018-11-16
[Core: Utility, Menu Filter] Move functions from Utility to Menu Filter because they're not used elsewhere

johnsaigle:
### Brief summary of changes

`arrayToCsv` and `getCleanString` are only used by `NDB_Menu_Filter` and so they should go there instead of `Utility`.

This PR is based on #4118 as that PR removes the only occurrence of `getCleanString()` that occurs elsewhere.

### DIFF
--- a/php/libraries/NDB_Menu_Filter.class.inc
+++ b/php/libraries/NDB_Menu_Filter.class.inc
@@ -513,7 +513,7 @@ class NDB_Menu_Filter extends NDB_Menu
         // add filters to query
         if (is_array($this->filter) && count($this->filter) > 0) {
             foreach ($this->filter as $field => $val) {
-                $prepared_key = Utility::getCleanString($field);
+                $prepared_key = $this->getCleanString($field);
                 $query_piece  = $this->_addValidFilters($prepared_key, $field, $val);
                 if (!empty($query_piece)) {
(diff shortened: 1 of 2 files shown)


driusan:
Can you rebase this? #4118 is merged.

php/libraries/NDB_Menu_Filter.class.inc
```diff
@@ -734,7 +734,51 @@ class NDB_Menu_Filter extends NDB_Menu
...
+    /**
+    * Coverts array into CSV string
+    *
+    * @param array $array the array to be converted
+    *
+    * @return string CSV string of data
+    */
+    function arrayToCSV(array $array): string
+    {
+        $fp = fopen("php://temp", 'w+');
```
paiva: Is is necessary to do a try/catch or not necessary because it is on temp?

johnsaigle: I'm not sure I didn't write this code. This PR is just moving it from one place to another.

PapillonMcGill: Unless the server run out of memory, there should not be a problem. Even in that case, even a try/catch would not make much difference.

PR #4121, opened 2018-11-19
[SRC] Adding useProjects to studyParams

sruthymathew123:
### Brief summary of changes
I think this PR will help in the rectification process ongoing now as I found useful/needed in mine.

### This resolves issue...

- [ ] Redmine? #

- [ ] Github? #

### To test this change...

- [ ] Try fetching the value `loris.config('useProjects')` in any javascript file or simply
 try an alert for studyParams array in loris.js file. `alert(JSON.stringify(studyParams));`


driusan:
useProjects is deprecated and should always be true. We said a long time ago we would just force everyone to have at least 1 project named "loris" if they didn't want to set up projects, but we never got around to removing the code. I don't think we should be adding more references to it.

sruthymathew123:
Closing this PR according to above comments.

driusan:
@sruthymathew123 it's still open?

PR #4124, opened 2018-11-19
[Core: BVL_Feedback_Panel] Add type declarations to function signatures

johnsaigle:
See #3878 for background.


php/libraries/BVL_Feedback_Panel.class.inc
```diff
@@ -44,7 +44,7 @@ class BVL_Feedback_Panel
      * @param null $sessionID The session ID for a given visit. Optional.
      * @param null $commentID The comment ID for a given instrument. Optional.
      */
-    function __construct($candID, $sessionID=null, $commentID=null)
+    function __construct(int $candID, $sessionID=null, $commentID=null)
```
PapillonMcGill:
```suggestion
    function __construct(int $candID, ?int $sessionID=null, ?int $commentID=null) : void
```
 adjust doc_block accordingly

PR #4130, opened 2018-11-20
[Core: Utility, NDB_Caller] Move removeCommonAffix() to NDB_Caller

johnsaigle:
### Brief summary of changes

This function is used in only NDB_Caller so it belongs there.

### To test this change...

- [ ] Everything should work the same.

### DIFF
--- a/php/libraries/NDB_Caller.class.inc
+++ b/php/libraries/NDB_Caller.class.inc
@@ -93,7 +93,7 @@ class NDB_Caller
         ) {
             header("HTTP/1.1 303 See Other");
             $url  = $settings->getBaseURL();
-            $url .= Utility::removeCommonAffix($url, $_SERVER['REQUEST_URI']);
+            $url .= self::removeCommonAffix($url, $_SERVER['REQUEST_URI']);
             // sending user to the url that was requested
             header("Location: $url");
             return '';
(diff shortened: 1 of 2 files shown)


php/libraries/NDB_Caller.class.inc
```diff
@@ -94,7 +94,7 @@ class NDB_Caller
         ) {
             header("HTTP/1.1 303 See Other");
             $url  = $settings->getBaseURL();
-            $url .= Utility::removeCommonAffix($url, $_SERVER['REQUEST_URI']);
+            $url .= parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH);
```
driusan: These don't seem to do the same thing at all.

One removes any common prefix/suffix to make a path relative to the LORIS root's baseurl  and make it relative, the other just parses the absolute URL given by PHP.

driusan: But is this code even used at all? the login endpoint goes through the index.php now and NDB_Caller only exists for instruments (which don't have a module to route to yet) and surveys (which don't have a login)

johnsaigle: I'm not sure. You'd know better than I would. Should I try deleting this section of the code?

driusan: I think the only impact would be for a session timeout on an instrument page.

johnsaigle: What would happen in that case? A user would not be properly redirected to the login page? Let me know what approach you'd like to take here.

driusan: I'm not very familiar with this code, but I think what it's *trying* to do is redirect to the page that you were on when it notices a login attempt from a session timeout on a non-login page.

The approach I'd like to take is to eventually kill NDB_Caller entirely since our router does the same thing in a more robust way. Maybe we can just move removeCommonAffix into NDB_Caller since it's the only place that uses it and not put much effort into changing things that have the potential to break obscure code paths there

driusan:
@johnsaigle can't merge because of conflicts

php/libraries/Utility.class.inc
```diff
@@ -858,54 +858,25 @@ class Utility
...
+    {
+        if (!is_array($arr)) {
+            return false;
+        }
+        $keys = array_keys($arr);
+        if (count($keys) === 0) {
+            return false;
         }
-        return $prefixStr;
+        return isset($arr[0]);
```
PapillonMcGill: What about if an array is mixed as in

[0] => some value
['xyz'] => some other value

I'm not even sure this portion should be in this PR

johnsaigle: Yeah I'm not sure what happened here. I made many changes to Utility on different branches and I think this got added as a result of a mistake in rebasing. I'll remove this function.

driusan:
Not sure why this isn't finishing Travis, but it also has conflicts and needs to be rebased, which will trigger a new Travis run anyways..

driusan:
Still failing Travis

PR #4131, opened 2018-11-20
[Core: Utility] Rewrite candidateAge() function to use built-in functions

johnsaigle:
### Brief summary of changes

Instead of using a custom method for calculating differences in dates, this function uses PHP built-in functions.

### To test this change...

- [ ] Ensure that candidate ages do not change. I'm not actually sure how to do this but perhaps someone with more project experience can weigh in here.

### DIFF
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -44,52 +44,23 @@ class Utility
      */
     static function calculateAge(string $dob, string $testdate): array
     {
-        if (!preg_match(
-            "/([0-9]{4})-?([0-9]{1,2})-?([0-9]{1,2})/",
-            $dob,
-            $matches
-        )) {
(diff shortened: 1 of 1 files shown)


driusan:
This would affect the scoring of nearly every instrument in every project that would make their published results unreproducable since LORIS uses "1 month = 30 days". Age calculation methods can't be changed on a whim.

johnsaigle:
@driusan I'm going to issue a small PR including what you just said in a code comment. Could be helpful in case this comes up again in the future.

driusan:
I think it would be worthwhile to bring it up in a LORIS meeting first, to see if there's still strong objections to changing it to a more calendar-based system in a major release.

(There was last time it came up.)

johnsaigle:
Sounds good. I'll add it to the agenda for next week.

PR #4132, opened 2018-11-20
[Core: TimePoint] Make return value of getBVLQCType() nullable

johnsaigle:
### Brief summary of changes

#4049 introduced strict typing into TimePoint. While on my sandbox I found a case where getBVLQCType() needs to return null or else LORIS will crash with a TypeError.

### DIFF
--- a/php/libraries/TimePoint.class.inc
+++ b/php/libraries/TimePoint.class.inc
@@ -565,9 +565,9 @@ class TimePoint
     /**
      * Returns BVLQC exclusion type
      *
-     * @return string
+     * @return string|null
      */
-    function getBVLQCType(): string
+    function getBVLQCType(): ?string
     {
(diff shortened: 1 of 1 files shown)


driusan:
What is the case? Why can't the calling code be fixed?

johnsaigle:
The calling code is line 473 of `modules/instrument_list/php/instrument_list_controlpanel.class.inc` and there aren't any issues there. The issue is that this value is nullable in the database:

```MySQL
  `BVLQCType` enum('Visual','Hardcopy') DEFAULT NULL,
```

I am working off of a very old Raisinbread dataset and there are many cases where this value is NULL.
Maybe this isn't an issue in real environments where everything is set from the front-end, but at the moment this bug breaks `instrument_list` on `major` under the above conditions.

PR #4133, opened 2018-11-20
[Core: Utility] Remove asArray function

johnsaigle:
### Brief summary of changes

`asArray()` had a cleanup tag about deleting it and replacing its calling code with toArray instead since the functions are basically equivalent.

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -295,7 +295,7 @@ class NDB_BVL_Instrument extends NDB_Page
             foreach ($instrumentPermissions["instrument"] as $instrument) {
                 if ($instrument["Test_name"] == $this->testName) {
                     $instrumentListed = true;
-                    $instrumentPerms  = Utility::asArray($instrument["permission"]);
+                    $instrumentPerms  = Utility::toArray($instrument["permission"]);
                 }
             }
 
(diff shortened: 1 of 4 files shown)


kongtiaowang:
php lorisform_parser.php 
PHP Fatal error:  Uncaught TypeError: Return value of Utility::toArray() must be of the type array, string returned in /var/www/loris/php/libraries/Utility.class.inc:491

johnsaigle:
Upon further review, these functions actually have quite different behaviour.

* `toArray()` takes array|string. When given an associative array (i.e. one that does not have `0` as an array key`) it returns a new numeric array containing the associative_array as its first element. IN the case of numeric arrays or strings, they simply get returned to the calling code unchanged.

* `asArray()` converts scalar values to arrays, and returns arrays.

We can take two approaches here:

1) Update the function docs to make these distinctions clear. At the very least we need to remove the cleanup note about asArray()
2) Do a deeper refactor:
    * Change `toArray()` to `toNumericArray()` and ensure that only associative arrays are passed as parameters. Use TypeErrors or DeprecatedExceptions to prevent calling code from passing scalar values.

Thanks to @xlecours for doing testing on this. @xlecours @driusan Do you have a preference as to what approach to use here?

driusan:
I prefer simplifying if possible (so approach #2), personally.. but you can always start by improving the docs

johnsaigle:
Closing this PR in favour of #4151

PR #4137, opened 2018-11-21
[Core: Utility] Add code comment on calculateAge()

johnsaigle:
### Brief summary of changes

Pursuant to the conversation in #4131, this functionality of LORIS should be documented in the code. 

Whether to use 30 day months or more accurate date calculations will be discussed in an upcoming LORIS meeting. If it's decided that LORIS should use 30 day months, this PR should be merged. Otherwise #4131 should be re-opened and merged.

Adding Blocked and Discussion Required until we know the right approach to take based on the needs of researchers.

### DIFF
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -35,6 +35,10 @@ class Utility
     /**
      * Computes an age in years:months:days (assumes 30 days per months)
      *
+     * NOTE This function assumes 1 month = 30 days. Although this is not really
+     * accurate, it should not be changed as it impacts the reproducibility of
+     * the analysis of data in LORIS.
+     *
      * @param string $dob      date of birth (or first date), formatted YYYY-MM-DD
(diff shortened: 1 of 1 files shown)


xlecours:
Pandora's box

Jkat:
I think I brought this up many, many years ago during a LORIS meeting and it was decided to keep things the way they were albeit the non-accurate calculation. But I don't think it hurts to bring it up again briefly.

PapillonMcGill:
calculating number of months and days from a date goes well (predefined function) for a specific calendar date, but when it is question of a difference between 2 dates, ... doesn't work properly.

Using 30.5 days/month would be somewhat more accurate (with an unwanted fraction half of the time) but 30 days/month is pretty much what people expect. Sometime is it calculated as 28 days/month when it is in your disadvantage. Be sure to read the fine print.

johnsaigle:
As discussed in the meeting today this function will not be changed for now.

This PR should be merged to clarify this decision for any future devs that stumble across the code.

PR #4139, opened 2018-11-21
[Core: NDB Page and Children] Add return types to function signatures

johnsaigle:
See #3878 for background.

This PR covers NDB Page and its child classes. They must be updated together because their function signatures all need to match.

### DIFF
--- a/php/libraries/NDB_Menu.class.inc
+++ b/php/libraries/NDB_Menu.class.inc
@@ -66,7 +66,7 @@ class NDB_Menu extends NDB_Page
      *
      * @return object The new object of $menu type
      */
-    static function &factory($menu, $mode)
+    static function &factory(string $menu, string $mode): object
     {
         // get the name of the class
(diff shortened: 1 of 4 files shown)


johnsaigle:
This PR will be easier to deal with when #4073 is merged so I'm adding `Blocked` for the time being.

driusan:
Can we close this and reopen it after the next major release? NDB_Page type signature changes will also require the updating of every single instrument file in every project that upgrades, and we already forced people to do that a couple releases ago, so giving them a break might be a good idea.

johnsaigle:
@driusan Sure that makes sense. This also impacts other classes that are tightly coupled with NDB_Page, such as LorisForm I believe. I'll make a list of the libraries that will be excluded from the 21 release.

johnsaigle:
Closing for now. Should be re-opened after the LORIS 21 release.

PR #4140, opened 2018-11-22
[tools/Data Dictionary Builder] Cleanup and bugfixes

ridz1208:
### Brief summary of changes
Minor revamp of the tool script that:
- fixes issue where `parameter_type_category` table was never being cleansed from older values
- allows support for multiple instruments using the same database table (BUGFIX since LORIS is supposed to support that)
- improves readability slightly
- cleans up some notices
- moves it to new directory structure
- make tool runnable from any directory

### This resolves issue...

CCNA issue where the script fails when 2 instruments (one extends the other) use the same database table causing an INSERT failure due to duplication of the primary key in the `parameter_type` table

### To test this change...

- [ ] follow the general instructions to generate the `ip_output.txt` file (should be something like `cd /var/www/loris/tools/ && find ../project/instruments/NDB_BVL_Instrument*.class.inc | php lorisform_parser.php`)
- [ ] then try running the script from the `tools` directory directly and/or from any other directory
- [ ] script is expected to run to completion as described in the tool itself


ridz1208:
@sruthymathew123 @Jkat I think you should be aware of this one !!

tools/exporters/data_dictionary_builder.php
```diff
@@ -36,100 +36,102 @@
...
  * @category Behavioural
  * @package  Main
  * @author   Loris Team <loris.mni@bic.mni.mcgill.ca>
  * @license  Loris License
  * @link     https://github.com/aces/Loris
  */
+require_once __DIR__ . '/../generic_includes.php';
 
-set_include_path(get_include_path().":../project/libraries:../php/libraries:");
+$parameter_types = array();
```
ridz1208: fix notices by declaring variables.

tools/exporters/data_dictionary_builder.php
```diff
@@ -36,100 +36,102 @@
...
 
-require_once __DIR__ . "/../vendor/autoload.php";
-$configFile = "../project/config.xml";
-$client     = new NDB_Client();
-$client->makeCommandLine();
-$client->initialize($configFile);
-$DB = Database::singleton();
 
-// Set a user for the history table
-if (!isset($_SESSION['State'])) {
```
ridz1208: this is automatically done in the database class, no need for it here

tools/exporters/data_dictionary_builder.php
```diff
@@ -370,51 +373,6 @@ function enumizeOptions($options, $table, $name)
     return "enum($enum)";
 }
 
-/**
```
ridz1208: these functions now exist directly in the database class as `pselectColWithIndexKey` and `pselectCol`

tools/exporters/data_dictionary_builder.php
```diff
@@ -36,100 +36,102 @@
...
+
+        //delete all 'Instrument' entries from parameter_type
+        //RUN function call required because of the IN clause
+        $DB->run(
+            "DELETE FROM parameter_type
+        WHERE ParameterTypeID IN ($instrumentParameterTypeIDString)"
+        );
+    }
 
-if (!empty($instrumentParameterTypeCategoryIDString)) {
```
ridz1208: this variable `$instrumentParameterTypeCategoryIDString` was always empty since the string was never actually being generated... soooo bugfixed

tools/exporters/data_dictionary_builder.php
```diff
@@ -193,36 +196,41 @@
...
             //find values to insert
             $Name = $table . "_" . $bits[1];
 
             $query_params = array(
-                             'Name'            => $Name,
-                             'Type'            => $bits[0],
-                             'Description'     => $bits[2],
-                             'SourceField'     => $bits[1],
-                             'SourceFrom'      => $table,
-                             'CurrentGUITable' => 'quat_table_'
```
ridz1208: @driusan this was causing issues, is it being used anywhere ????

tools/exporters/data_dictionary_builder.php
```diff
@@ -36,100 +36,102 @@
...
+    //Get all types IDs related to instrument categories
+    $instrumentParameterTypeIDs = $DB->pselectCol(
         "SELECT ParameterTypeID
             FROM parameter_type_category_rel
             WHERE ParameterTypeCategoryID
-            IN (" . $prepIn['query'] . ")",
-        $DB,
-        $instrumentParameterTypeIDs,
-        $prepIn['array']
+            IN (" . $prepIn["query"] . ")",
```
ridz1208: @driusan should we be using find in set instead ??

SQL/Cleanup_patches/2018-11-20-remove_currentGUITable_field.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE parameter_type DROP COLUMN CurrentGUITable;
```
ridz1208: @driusan @xlecours any idea if this is still doing anything ??

ridz1208:
@kongtiaowang can you help with travis ?

tools/exporters/data_dictionary_builder.php
```diff
@@ -36,100 +36,102 @@
  * Usage:
  *    php data_dictionary_builder.php
  *
- * PHP Version 5
+ * PHP Version 5,7
```
PapillonMcGill: sure about 5, 7?  Should it be 7?

tools/exporters/data_dictionary_builder.php
```diff
@@ -36,100 +36,102 @@
...
-
 print "Cleared data from BVL instruments\n";
 
 print "Reading instruments\n";
 //Read the ip_output.txt staging file.
-$fp   =fopen("ip_output.txt", "r");
-$data =fread($fp, filesize("ip_output.txt"));
+$fp = fopen(__DIR__."/../ip_output.txt", "r");
+if (!$fp) {
+    print "There was an issue opening the ip_output file. Make sure permissions the
```
PapillonMcGill:
```suggestion
    print "There was an issue opening the ip_output file. Ensure the 
```

tools/exporters/data_dictionary_builder.php
```diff
@@ -138,33 +140,34 @@
...
-             $catId = $DB->lastInsertID;
-             $tblCount++;
+            $catId = $DB->lastInsertID;
+            $tblCount++;
             break;
 
         case "header":
             continue;
-            break;
+                break;
```
PapillonMcGill: Is `break` required after `continue`?

tools/exporters/data_dictionary_builder.php
```diff
@@ -138,33 +140,34 @@
...
         case "title":
-            $title  = $bits[1];
-            $error  = $DB->insert(
+            $title = $bits[1];
+            $error = $DB->insert(
                 "parameter_type_category",
                 array(
-                 'Name' => $title,
-                 'Type' => 'Instrument',
+                 "Name" => $title,
```
PapillonMcGill: are single or double quote the standard to follow in LORIS?

SQL/0000-00-00-schema.sql
```diff
@@ -1552,20 +1551,20 @@ CREATE TABLE `parameter_type` (
...
-  (3,'Intensity_artifact','text',NULL,NULL,NULL,NULL,'parameter_file',NULL,NULL,0,0),
-  (4,'Movement_artifacts_within_scan','text',NULL,NULL,NULL,NULL,'parameter_file',NULL,NULL,0,0),
-  (5,'Movement_artifacts_between_packets','text',NULL,NULL,NULL,NULL,'parameter_file',NULL,NULL,0,0),
-  (6,'Coverage','text',NULL,NULL,NULL,NULL,'parameter_file',NULL,NULL,0,0),
-  (7,'md5hash','varchar(255)','md5hash magically created by NeuroDB::File',NULL,NULL,'parameter_file.Value','parameter_file',NULL,'quat_table_1',1,0),
-  (8,'Color_Artifact','text',NULL,NULL,NULL,NULL,'parameter_file',NULL,NULL,0,0),
-  (9,'Entropy','text',NULL,NULL,NULL,NULL,'parameter_file',NULL,NULL,0,0);
-
-
-INSERT INTO parameter_type (Name, Type, Description, RangeMin, RangeMax, SourceField, SourceFrom, CurrentGUITable, Queryable, SourceCondition) VALUES
```
kongtiaowang: It is better to keep the values list.
(Name, Type, Description, RangeMin, RangeMax, SourceField, SourceFrom, CurrentGUITable, Queryable, SourceCondition)

tools/exporters/data_dictionary_builder.php
ridz1208: @driusan ??

tools/exporters/data_dictionary_builder.php
ridz1208: I dont think we have a standard but I was getting dizzy reading through this file I decided to just change them

tools/exporters/data_dictionary_builder.php
ridz1208: good question from what I have seen it seems like it does not no. I will remove it

tools/exporters/data_dictionary_builder.php
driusan: @ridz1208 ?

tools/exporters/data_dictionary_builder.php
driusan: no, both of them modify the control flow, the break will never be reachable.

ridz1208: on second thought this might be a bit more difficult. was the intent of `continue; break;` to stop exit the switch statement only or the entire `foreach ($items AS $item) {` (line 146) ?? because if thats the case, we would need to replace continue break by `continue 2;`

driusan: the break is never reachable. Removing it won't change any behaviour.

tools/exporters/data_dictionary_builder.php
driusan: the quat table part? Yeah, you can remove that..

SQL/0000-00-00-schema.sql
ridz1208: I am just removing one column from the entire table

PapillonMcGill: So?
@kongtiaowang  comment is valid and can prevent error in the future. You can also add the attributes list on line 1553 (copy&paste from line 2134 or RBdata.sql)  and also just not include the attributes that are always null. 

On line 1564, should have back-tick around table name.

tools/exporters/data_dictionary_builder.php
driusan: Double quotes unless single quotes makes it cleaner.

SQL/0000-00-00-schema.sql
driusan: I'm not sure what you guys are talking about.. it looks like he is keeping the explicit column list, he's just removing CurrentGUITable from it since the column was dropped from the table?

ridz1208: @PapillonMcGill I'm not removing the list... not sure what the issue is here

SQL/0000-00-00-schema.sql
```diff
@@ -1552,20 +1551,20 @@ CREATE TABLE `parameter_type` (
 
 
 INSERT INTO `parameter_type` VALUES
```
PapillonMcGill:
```suggestion
INSERT INTO `parameter_type` (ParameterTypeID, Name, Type, Description, RangeMin, RangeMax, SourceField, SourceFrom, SourceCondition, Queryable, IsFile) VALUES
```
with some lint for line lenght

SQL/0000-00-00-schema.sql
PapillonMcGill: @ridz1208  see comments for line 1553

SQL/0000-00-00-schema.sql
```diff
@@ -1552,20 +1551,20 @@ CREATE TABLE `parameter_type` (
...
+  (3,'Intensity_artifact','text',NULL,NULL,NULL,NULL,'parameter_file',NULL,0,0),
+  (4,'Movement_artifacts_within_scan','text',NULL,NULL,NULL,NULL,'parameter_file',NULL,0,0),
+  (5,'Movement_artifacts_between_packets','text',NULL,NULL,NULL,NULL,'parameter_file',NULL,0,0),
+  (6,'Coverage','text',NULL,NULL,NULL,NULL,'parameter_file',NULL,0,0),
+  (7,'md5hash','varchar(255)','md5hash magically created by NeuroDB::File',NULL,NULL,'parameter_file.Value','parameter_file',NULL,1,0),
+  (8,'Color_Artifact','text',NULL,NULL,NULL,NULL,'parameter_file',NULL,0,0),
+  (9,'Entropy','text',NULL,NULL,NULL,NULL,'parameter_file',NULL,0,0);
+
+
+INSERT INTO parameter_type (Name, Type, Description, RangeMin, RangeMax, SourceField, SourceFrom, Queryable, SourceCondition) VALUES
```
PapillonMcGill:
```suggestion
INSERT INTO `parameter_type` (Name, Type, Description, RangeMin, RangeMax, SourceField, SourceFrom, Queryable, SourceCondition) VALUES
```

SQL/0000-00-00-schema.sql
ridz1208: @PapillonMcGill your parameter count does not match the number of parameters in the INSERT statements.

an addition like that should be properly tested before submission or travis will most probably fail

SQL/0000-00-00-schema.sql
ridz1208: so just to be clear there is no issue here ? the issue is only on line 1553

SQL/0000-00-00-schema.sql
ridz1208: I will make the appropriate changes and add it to my PR

ridz1208:
@PapillonMcGill Done, please re-review

SQL/0000-00-00-schema.sql
PapillonMcGill: both line are ok now!

SQL/0000-00-00-schema.sql
PapillonMcGill: @ridz1208 sorry, Done the suggestion quickly while on a conf. call.

ridz1208:
@PapillonMcGill I dismissed your approval. please re-review when you get a chance

tools/exporters/data_dictionary_builder.php
Jkat: db class will get the current user vs here it's the owner of this script. this will break for any project that hasn't set up a matching db/system user.

tools/exporters/data_dictionary_builder.php
```diff
@@ -240,34 +247,33 @@
...
         continue;
     }
-    print "Inserting validity for $table\n";
-    // Insert validity
+
+    // INSTRUMENT VALIDITY
+    print "\tInserting validity for $table\n";
     $Name = $table . "_Validity";
 
-    $_type_enum = 'enum(\'Questionable\', \'Invalid\', \'Valid\')';
```
Jkat: 🤔

cmadjar: @Jkat can you clarify your comment?

tools/exporters/data_dictionary_builder.php
xlecours: We should use DELETE JOIN

SQL/Cleanup_patches/2018-11-20-remove_currentGUITable_field.sql
xlecours: no clue

tools/exporters/data_dictionary_builder.php
Jkat: I think I was very confused why this line was done this way instead of how it has been modified now on line 255: https://github.com/aces/Loris/pull/4140/files#diff-acb7e0105c9e180f46efe44050bf0e4dR255

SQL/Cleanup_patches/2018-11-20-remove_currentGUITable_field.sql
driusan: it's not

tools/exporters/data_dictionary_builder.php
ridz1208: will consider it

ridz1208:
@PapillonMcGill @xlecours Thx for the single query solution I LOVE IT
Now please re-review

@nicolasbrossard @sruthymathew123 @Jkat I would like someone from IBIS and PREVENTAD to test it just because it's definitely something you use.

we might need to talk before testing it because if the current tables you have are "dirty" this might throw an error. so come see me when you are ready to test

ridz1208:
@LeighMac In reviewer's hands now

tools/exporters/data_dictionary_builder.php
```diff
@@ -0,0 +1,361 @@
...
+// This data will be rebuilt below and IDs will be preserved when possible using
+// the data queried above
+$DB->run(
+    "DELETE parameter_type, 
+        parameter_type_category_rel, 
+        parameter_type_category
+    FROM parameter_type_category
+		JOIN parameter_type_category_rel USING(ParameterTypeCategoryID) 
+		JOIN parameter_type USING(ParameterTypeID) 
+    WHERE parameter_type_category.Type='Instrument';"
```
sruthymathew123: @ridz1208 Remove the `;`

sruthymathew123:
@ridz1208 So I just tried running this script
I tried running this script and it ended with below attached database error. 
As you suggested I manually run the the delete command in the script and tried again. But still the same error is throwing up. Did I miss anything?

rida_data_dictionary

tools/exporters/data_dictionary_builder.php
```diff
@@ -0,0 +1,361 @@
...
+$parameter_types = array();
+$instrumentParameterTypeCategoryIDs = array();
+$instrumentParameterTypeIDs         = array();
+
+
+$parameter_types = $DB->pselectColWithIndexKey(
+    "Select Name, ParameterTypeID from parameter_type",
+    array(),
+    "Name"
+);
```
ridz1208: old code:
```
getColumns(	
    "Select Name, ParameterTypeID from parameter_type",	
    $DB,	
    $parameter_types,	
    array()	
);
...
function getColumns($query, &$DB, &$result, $preparray)	
{	
    $TwoDArray = $DB->pselect($query, $preparray);	
    foreach ($TwoDArray as $containers=>$cells) {	
        $values = array_values($cells);	
        $result[$values[0]] = $values[1];	
    }	
    return $result;	
}
```

The function used here `pselectColWithIndexKey()` throws an error when it gets 2 rows with the same `Name`. The error is there to prevent data from being overwritten. Currently, all old projects will get an error at this stage because the `parameter_type` table was not well maintained.

when looking at the old code we see that it was doing the something similarto `pselectColWithIndexKey()` but it allows for a non-unique `name` value (it just overwrites the old value with the new one)... the problem is that this `$parameter_types` array result is used to prevent existing parameter type IDs from changing accross time and given that the old code allowed for duplicates on `Name` there is no guarantee that the "saved" parameter type ID is a valid one, it's just the last one in the list of fields with the same `Name`

All this to say, this error SHOULD be thrown but we might need to find a workaround for current projects to allow for some time to clean their data before they just get errors on this script

PapillonMcGill: @ridz1208 There is no garante that the ordering will the earliest to latest in your query.

Change to "Select Name, ParameterTypeID from parameter_type order by ParameterTypeID desc limit 1" will get you the latest entry in the table when multiple row have the same name.

ridz1208:
@sruthymathew123 could you test this again??

make sure to run the tool first and test that at the same time

tools/exporters/data_dictionary_builder.php
ridz1208: ordering is not important

PapillonMcGill: Using my query (leaving the `order by ParameterTypeID desc` part) would still be more efficient than you multi-lines php code. Just add the `limit 1` to your query.

sruthymathew123:
@ridz1208  My testing initially failed because of some 'BLANK/NULL'  'Name' entries in parameter_type table. This was very specific with my database table and I didn't see anything like that on production or staging database.  Anyway, if we can add  a case to detect/fix it, it will be really cool.

After deleting those unwanted blank entries the script worked as expected. :)

ridz1208:
@xlecours @PapillonMcGill unfortunately I had to remove the delete join, it's impossible to do the delete join apparently with foreign keys.

from MYSQL docs (https://dev.mysql.com/doc/refman/8.0/en/delete.html)
`If you use a multiple-table DELETE statement involving InnoDB tables for which there are foreign key constraints, the MySQL optimizer might process tables in an order that differs from that of their parent/child relationship. In this case, the statement fails and rolls back. Instead, you should delete from a single table and rely on the ON DELETE capabilities that InnoDB provides to cause the other tables to be modified accordingly.`

ridz1208:
@sruthymathew123 Before testing this, I'd like to make sure that the DELETE CASCADES are the right move here. I will let you know when it's ready to go

ridz1208:
failing because of unrelated issue

fix: https://github.com/aces/Loris/pull/4474/

ridz1208:
@xlecours agrees with the delete cascades and I only care about his opinion sooo... discussion no longer required.

@sruthymathew123 you cna go for testing

sruthymathew123:
@ridz1208 Tested again and all looks good to me.

PR #4141, opened 2018-11-22
[tools/Fix Candidate Age] Cleanup and bugfixes

ridz1208:
### Brief summary of changes
Minor revamp of the tool script that:
- make sure to use the table name used by the instrument rather than the value in the test_names table (BUGFIX)
- moved tool the the new directory structure
- improved readability

### This resolves issue...
CCNA issue where if an instrument is named differently than it's SQL table this tool fails miserably

### To test this change...

- [ ] simply run the tool and ensure proper functioning
- [ ] manually alter an age and run again, this value should be caught and corrected with the `confirm` option on
- [ ] change the value of the `$this->table` variable in any instrument (and change the SQL table name to match that) and ensure the script still fixes ages in this instrument.

### DIFF
--- a/tools/data_integrity/fix_candidate_age.php
+++ b/tools/data_integrity/fix_candidate_age.php
@@ -179,7 +179,7 @@ function showHelp()
 
     echo "When the 'check' option is used, the script only detects and reports 
     miscalculated and NULL ages. 
-    Using the 'confirm' option will apply the necessary corrections to the data.";
+    Using the 'confirm' option will apply the necessary corrections to the data.\n\n";
 
     die();
 }
--- a/tools/data_integrity/fix_candidate_age.php
+++ b/tools/data_integrity/fix_candidate_age.php
@@ -17,10 +17,7 @@
  * @license  Loris license
  * @link     https://www.github.com/aces/Loris-Trunk/
  */
-require_once 'generic_includes.php';
-
-$config   = NDB_Config::singleton();
-$database = $config->getSetting('database');
+require_once __DIR__ . '/../generic_includes.php';
 
 $instruments = Utility::getAllInstruments();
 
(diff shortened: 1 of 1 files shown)


tools/data_integrity/fix_candidate_age.php
```diff
@@ -29,50 +26,72 @@
...
+        echo "\t$inst does not use a `Date_taken` field and should be handled separately.\n";
         continue;
     }
 
     //Get instrument SQL table
     $DBInstTable = $DB->pselect(
-        "SELECT * 
-         FROM $inst i 
+        "SELECT i.CommentID, Date_taken, Candidate_Age
+         FROM $table i
```
sruthymathew123: @ridz1208 For IBIS, we have many survey instruments tables which do have the default examiner, Candidate_Age fields. So the script is failing on IBIS dataset.

ridz1208: can you add the fail message here or give me an example table that it fails on ?

cmadjar: @ridz1208 Might be good to have a way to check if there is a `Candidate_Age` field in the instrument. Looks like for PREVENT-AD, radiology_review does not have a `Candidate_Age` field making the script to crash with the following error and stopping updating `Candidate_Age` for other instruments:
```
PHP Warning:  PDOStatement::execute(): SQLSTATE[42S22]: Column not found: 1054 Unknown column 'Candidate_Age' in 'field list' in /var/www/loris/php/libraries/Database.class.inc on line 659
Could not execute SELECT i.CommentID, Date_taken, Candidate_Age
         FROM radiology_review i 
            LEFT JOIN flag f ON (i.commentID=f.CommentID)
            LEFT JOIN session s ON (f.SessionID=s.ID)
            LEFT JOIN candidate c ON (s.CandID=c.CandID)
         WHERE c.Active='Y' AND s.Active='Y'. Stack trace#0 /var/www/loris/php/libraries/Database.class.inc(694): Database->execute(Object(PDOStatement), Array)
#1 /var/www/loris/tools/data_integrity/fix_candidate_age.php(77): Database->pselect('SELECT i.Commen...', Array)
#2 {main}
PHP Fatal error:  Uncaught DatabaseException: Unknown column 'Candidate_Age' in 'field list' in /var/www/loris/php/libraries/Database.class.inc:662
Stack trace:
#0 /var/www/loris/php/libraries/Database.class.inc(694): Database->execute(Object(PDOStatement), Array)
#1 /var/www/loris/tools/data_integrity/fix_candidate_age.php(77): Database->pselect('SELECT i.Commen...', Array)
#2 {main}
  thrown in /var/www/loris/php/libraries/Database.class.inc on line 662
```
I would tend to first check that this field is available for the instrument, if not available print a warning, no `Candidate_Age` field for instrument XXX, if available, update the Age

tools/data_integrity/fix_candidate_age.php
```diff
@@ -17,10 +17,7 @@
  * @license  Loris license
  * @link     https://www.github.com/aces/Loris-Trunk/
  */
-require_once 'generic_includes.php';
-
-$config   = NDB_Config::singleton();
-$database = $config->getSetting('database');
+require_once __DIR__ . '/../generic_includes.php';
```
cmadjar: @ridz1208 Sorry, one last comment. It might be good to explain clearly what the confirm option does. I guessed it but feels it could be better documented.

tools/data_integrity/fix_candidate_age.php
ridz1208: @sruthymathew123 @cmadjar thats probably what I'm gonna do

tools/data_integrity/fix_candidate_age.php
```diff
@@ -133,3 +164,22 @@
...
+    echo "*** Fix Candidate Age ***\n\n";
+
+    echo "Usage: php fix_candidate_age.php [confirm | check | help | -h]\n";
+    echo "Example: php fix_candidate_age.php help\n";
+    echo "Example: php fix_candidate_age.php check\n";
+    echo "Example: php fix_candidate_age.php confirm\n\n";
+
+    echo "When the 'check' option is used, the script only detects and reports 
+    miscalculated and NULL ages. 
+    Using the 'confirm' option will apply the necessary corrections to the data.";
```
cmadjar: should probably add a `\n` at the end of the help section so that the prompt does not end up right to the last word of the help section.

sruthymathew123:
@ridz1208 Retested on IBIS data set and now it is all good.

tools/data_integrity/fix_candidate_age.php
cmadjar: Cool! Works now.

tools/data_integrity/fix_candidate_age.php
ridz1208: @cmadjar DONE

tools/data_integrity/fix_candidate_age.php
```diff
@@ -17,62 +17,95 @@
...
                 " Ignored for one of the following reasons:\n".
                 "  - The candidate is inactive.\n".
                 "  - The session is inactive.\n\n";
             continue;
         }
+
         if (!$instrument) {
             // instrument does not exist
-            echo $inst." for CommentID:$row[CommentID] could not be instantiated.\n";
+            echo "\t$inst for CommentID:$row[CommentID] could not be instantiated.\n";
```
PapillonMcGill: should it be
```suggestion
            echo "\t$inst for CommentID:$row['CommentID'] could not be instantiated.\n";
```
to properly display?

ridz1208: Because I'm still inside the string to evaluate. PHP evaluates `"$row[CommentID]"` the same as `$row["CommentID"]`

look for "Example #10 Simple syntax example" in here http://php.net/manual/en/language.types.string.php#language.types.string.parsing.simple

ridz1208:
@driusan ready to go

PR #4142, opened 2018-11-22
[Instrument manager] Reactify Menu Filters

xlecours:
### Brief summary of changes
Now use react for the menu filter.
Adding filters

Move the upload form in a different table.
Use ajax to upload the instrument.
Adds handler for POST requests in the php file.

### This resolves issue...
This module was slowly drifting into the abyss 

### To test this change...
Please run the test plan.


modules/instrument_manager/jsx/instrumentManagerIndex.js
```diff
@@ -0,0 +1,174 @@
...
+   * Modify behaviour of specified column cells in the Data Table component
+   *
+   * @param {string} column - column name
+   * @param {string} cell - cell content
+   * @param {object} row - row content indexed by column
+   *
+   * @return {*} a formated table cell for a given column
+   */
+  formatColumn(column, cell, row) {
+    // Set class to 'bg-danger' if file is hidden.
```
HenriRabalais: This line can be removed.

modules/instrument_manager/jsx/instrumentManagerIndex.js
```diff
@@ -0,0 +1,174 @@
...
+      return <Loader/>;
+    }
+
+    const fields = [
+      {label: 'Instrument', show: true, filter: {
+        name: 'instrument',
+        type: 'text',
+      }},
+      {label: 'Instrument Type', show: true, filter: {
+        name: 'instrument_type',
```
HenriRabalais: I believe we are going to go with a lowerCamelCase convention for the `name` index. This will help when a mapper is eventually made to properly associate the `data` and `fields` arrays.

modules/instrument_manager/jsx/instrumentManagerIndex.js
```diff
@@ -0,0 +1,174 @@
...
+      }},
+    ];
+
+    const tabs = [
+      {id: 'browse', label: 'Browse'},
+      {id: 'upload', label: 'Upload'},
+    ];
+
+    const feedback = () => {
+      if (!this.state.data.caninstall) {
```
HenriRabalais: Perhaps `canInstall` shuold be passed in a permission object? It just seems to odd to have it as part of 'data'. This may require some restructuring of how we handle permissions on the front end. Is there a way to access this information via `this.props.hasPermission`?

modules/instrument_manager/jsx/instrumentManagerIndex.js
```diff
@@ -0,0 +1,174 @@
...
+      let content = null;
+      if (this.state.data.writable) {
+        let url = loris.BaseURL.concat('/instrument_manager/?format=json');
+        content = (
+          <InstrumentUploadForm action={url}/>
+        );
+      } else {
+        content = (
+          <div className='alert alert-warning'>
+            Installation is not possiblegiven the current server configuration.
```
HenriRabalais:
```suggestion
            Installation is not possible given the current server configuration.
```

modules/instrument_manager/jsx/instrumentManagerIndex.js
```diff
@@ -0,0 +1,174 @@
...
+            File upload is still possible but instruments will need to be
+            installed manually
+          </div>
+        );
+      }
+    };
+
+    const uploadTab = () => {
+      let content = null;
+      if (this.state.data.writable) {
```
HenriRabalais: Same concern as above.

modules/instrument_manager/jsx/uploadForm.js
```diff
@@ -0,0 +1,97 @@
...
+    data.append('install_file', this.state.selectedFile);
+
+    fetch(this.props.action, {
+      method: 'POST',
+      credentials: 'same-origin',
+      body: data,
+    })
+    .then((resp) => {
+      if (resp.status == 201) {
+        swal({
```
HenriRabalais: We can now use the `swal2` dependency which uses promise syntax (e.g. `then`, `catch) which I find a lot more intuitive than the bootstrap swal. Perhaps it's not accessible on major yet though.

modules/instrument_manager/jsx/uploadForm.js
```diff
@@ -0,0 +1,97 @@
...
+      }
+    })
+    .catch((error) => {
+      this.setState({error: true});
+      console.error(error);
+    });
+  }
+
+  render() {
+    const disabled = () => this.state.selectedFile == null;
```
HenriRabalais: Can you use strict equality here? ie. `===`?

modules/instrument_manager/jsx/instrumentManagerIndex.js
driusan: I don't think so. Whether or not they can install is a function of the filesystem permissions/setup, not the user's permission.

xlecours: I partly agree with you. Yes, it is odd the it is part of the data but it is not a permission either. This tells if the project/instrument and project/table_sql directory can be written into.

modules/instrument_manager/jsx/instrumentManagerIndex.js
xlecours: Same,this tells if the project/instrument and project/table_sql directory can be written into.

modules/instrument_manager/jsx/uploadForm.js
xlecours: yes, the default is null so it wont be undefined.

modules/instrument_manager/jsx/instrumentManagerIndex.js
HenriRabalais: @driusan @xlecours Hmm, okay. Yeah, that makes. I guess as we are making standards for the objects we are passing to the front end (e.g. `data`, `fieldOptions`, etc.), we may want to group it into a `configuration` object? Rather than have stray configuration booleans in the `data` object. Not necessary for this PR, just something to think about.

modules/instrument_manager/jsx/uploadForm.js
xlecours: maybe in a second PR

driusan:
@xlecours failing Travis

PR #4143, opened 2018-11-22
Remove _getNumberPages

xlecours:
This is based on https://github.com/aces/Loris/pull/4142

### Brief summary of changes
Remove the function that counts. It was used for Pagination of menu filters. 
#4142  removes the last usage of it.

### This resolves issue...
Performance of doing the same query twice... in every menu filters

### To test this change...
I don't know, I can't see why this is still there.


kongtiaowang:
ready to go.

PR #4145, opened 2018-11-23
[travis] fix can't load config.xml and can't load ajax file for PHP built-in server

kongtiaowang:
### Brief summary of changes
This PR fix can't load config.xml and can't load ajax file for PHP built-in server in Travis and Docker. 

### To test this change...
Travis and docker can test it automatically.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,7 +84,8 @@ before_script:
     - cd ..
     - mysql LorisTest < project/tables_sql/testtest.sql
 env:
-    - LORIS_DB_CONFIG=fjdklsjfdjsaflkjkldsjfkladsjklfdfljdsfkljlkdsfjkljfdfklsdjflkalkfj
+    - TRAVIS_CI=true
+    - LORIS_DB_CONFIG=/test/config.xml
 
 script:
     - vendor/bin/phpunit --configuration test/phpunit.xml
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,7 +84,7 @@ before_script:
     - cd ..
     - mysql LorisTest < project/tables_sql/testtest.sql
 env:
-    - TRAVIS_CI=true LORIS_DB_CONFIG=/../test/config.xml
+    - LORIS_DB_CONFIG=/../test/config.xml
 
 script:
     - vendor/bin/phan && npm run lint:php && npm run lint:javascript && vendor/bin/phpunit --configuration test/phpunit.xml
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,8 +84,7 @@ before_script:
     - cd ..
     - mysql LorisTest < project/tables_sql/testtest.sql
 env:
-    - TRAVIS_CI=true
-    - LORIS_DB_CONFIG=/test/config.xml
+    - TRAVIS_CI=true LORIS_DB_CONFIG=/test/config.xml
 
 script:
     - vendor/bin/phpunit --configuration test/phpunit.xml
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,7 +84,7 @@ before_script:
     - cd ..
     - mysql LorisTest < project/tables_sql/testtest.sql
 env:
-    - LORIS_DB_CONFIG=test/config.xml
+    - LORIS_DB_CONFIG=../test/config.xml
 
 script:
     - vendor/bin/phpunit --configuration test/phpunit.xml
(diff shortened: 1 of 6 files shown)


php/libraries/NDB_Config.class.inc
```diff
@@ -72,10 +72,23 @@ class NDB_Config
...
+            $travis = getenv('TRAVIS');
+            if (!empty($env)) {
+                if (!empty($travis)) {
+                      $file  = dirname(__DIR__);
+                      $file .= $env;
+                } else {
+                    $file = $env;
+                }
+                 $config->load($file);
+            } else {
```
driusan: Why is this being added here? It should already be handled line ~125 (in configFilePath)

kongtiaowang: @driusan You are right, the first time call this function, it goes ~125. But when a module calls NDB_factory function, it will call NDB_Config.class.inc again, the second time it got an error. this is why I put the fix here.

kongtiaowang:
@driusan could u review again?

.travis.yml
```diff
@@ -84,8 +83,8 @@ before_script:
     - find ../project/instruments/NDB_BVL_Instrument_testtest.class.inc | php generate_tables_sql.php
     - cd ..
     - mysql LorisTest < project/tables_sql/testtest.sql
-#env:
- # - "LORIS_DB_CONFIG=test/config.xml"
+env:
+    - TRAVIS_CI=true LORIS_DB_CONFIG=/../test/config.xml
```
driusan: I think we can just use one of the existing ones rather than adding another environment variable

https://docs.travis-ci.com/user/environment-variables#default-environment-variables

htdocs/AjaxHelper.php
```diff
@@ -101,8 +101,9 @@
 }
 
 // Also check the module directory for PHP files
+// remove query string from url after '?' like: ajax/test.php?name=test
+$File     = explode("?", $File)[0];
```
driusan: Where is this coming up? File is defined as `$File   = $_GET['script'];`, on line 57, so the variable shouldn't have included it in the first place.

php/libraries/NDB_Config.class.inc
```diff
@@ -72,10 +72,23 @@ class NDB_Config
...
-            $config->load($configFile);
+            //  load config in test
+            $env    = getenv('LORIS_DB_CONFIG');
+            $travis = getenv('TRAVIS');
+            if (!empty($env)) {
+                if (!empty($travis)) {
+                      $file  = dirname(__DIR__);
+                      $file .= $env;
+                } else {
+                    $file = $env;
```
driusan: Instead of adding a hack into the code here for Travis, why doesn't the environment just get set properly/to the absolute path in the .travis.yml?

htdocs/AjaxHelper.php
kongtiaowang: remove query string from url after '?' like: ajax/test.php?name=test
$File = "ajax/test.php?name=test", it will not be found.
$File = "ajax/test.php" it can be found. (on line 106 )

php/libraries/NDB_Config.class.inc
kongtiaowang: This is also one option for docker setup.

.travis.yml
kongtiaowang: done

php/libraries/NDB_Config.class.inc
kongtiaowang: The absolute path in the .travis.yml depends on user's account.
/home/travis/build/### **kongtiaowang**/Loris/php/home/travis/build/aces/Loris/test/config.xml

kongtiaowang: @driusan The absolute path depends on user's account.
/home/travis/build/**kongtiaowang**/Loris/php
/home/travis/build/**aces**/Loris/php

driusan: `$TRAVIS_BUILD_DIR/php`?

kongtiaowang:
depends on https://github.com/aces/Loris/pull/4254.
I will rewire this PR.

kongtiaowang:
Docker test instead of Travis test. I will close this.

kongtiaowang:
using Docker test instead of Travis test. I will close this.

PR #4147, opened 2018-11-26
[Core:NDB_Config] fix parameter type

kongtiaowang:
### Brief summary of changes
Fix Loris can't load on local VM with a specific config setting. 
This function can't handle a string.

PHP Fatal error:  Uncaught TypeError: Argument 1 passed to NDB_Config::isNumericArray() must be of the type array, string given

### To test this change...
test loading your Loris on this branch without issue.


johnsaigle:
Which config setting caused this problem? Just wondering.

sruthymathew123:
Loris was not loading on my VM with my current config.xml file. This PR fix the issue.

PR #4151, opened 2018-11-27
[Core: Utility] Deprecate Utility::toArray()

johnsaigle:
### Brief summary of changes
 Following #4133 `toArray()` is being deprecated in favour of the more precise function `associativeToNumericArray()`. This function does the same thing as toArray() but only accepts and returns arrays. 

This change reduces confusion between toArray and asArray as well as allows us to be more precise with the way we're using types.

The calling code has been updated to use the new function and the old function will throw a deprecation exception in the next release.

### Related

- [ ] Github? #4133 #3878 #4046 

### To test this change...

- [ ] Try running the tools scripts and make sure they don't do anything weird.

### DIFF
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -422,16 +422,16 @@ class Utility
      * original functioning of toArray which was not type-safe. It is
      * primarily used for extracting configuration settings.
      *
-     * @param array $array An asssociative array.
+     * @param array $associative_array An asssociative array.
      *
      * @return array The new numeric array, or $array if it was already numeric.
      */
-    static function associativeToNumericArray(array $array): array
+    static function associativeToNumericArray(array $associative_array): array
     {
-        if (!array_key_exists(0, $array)) {
-            return array($var);
+        if (!array_key_exists(0, $associative_array)) {
+            return array($associative_array);
         }
-        return $array;
+        return $associative_array;
     }
 
     /**
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -417,6 +417,24 @@ class Utility
     }
 
     /**
+     * Converts an associative array to a numeric array consisting of the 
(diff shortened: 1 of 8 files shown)


php/libraries/Utility.class.inc
```diff
@@ -432,6 +432,24 @@ class Utility
...
+     * primarily used for extracting configuration settings.
+     *
+     * @param array $array An asssociative array.
+     *
+     * @return array The new numeric array, or $array if it was already numeric.
+     */
+    static function associativeToNumericArray(array $array): array
+    {
+        if (!array_key_exists(0, $array)) {
+            return array($var);
```
driusan: `$var` isn't defined

johnsaigle: thanks

php/libraries/Utility.class.inc
```diff
@@ -445,24 +463,31 @@ class Utility
      *
      * @return array If $var is an array, var, otherwise an array containing $var
      */
-    static function toArray($var): array
```
johnsaigle: This return type was actually incorrect as the function can return a string.

php/libraries/Utility.class.inc
```diff
@@ -429,24 +447,33 @@ class Utility
...
     {
+        // TODO Uncomment the exception code below after LORIS 21 is released.
+        //throw new DeprecatedException(
+        //    'Utility::toArray() is deprecated and should not be used. ' .
+        //    'Instead use Utility::associativeToNumericArray() for converting ' .
+        //    'arrays. Scalar values cannot be passed to this function.'
+        //);
+        error_log(
+            'Utility::toArray() will be deprecated in LORIS ' .
+            'version 21. Please use Utility::associativeToNumericArray() ' .
```
driusan: I merged this anyways, but in the future, I wouldn't specify the release number but just say "in a future release" to have a little more flexibility/just in case something comes up and it doesn't get done right away.

Issue #4070, opened 2018-11-06
Add new rules to Phan static analysis


johnsaigle:
#4069 causes Phan to report typing issues in LORIS. Changing the function signatures or calling functions to reduce these warnings is super helpful and pretty straightforward.

Any pull requests related to this topic should be issued to the `major` branch.

johnsaigle:
This is now a whole series of Phan-related PRs tracked by this project https://github.com/aces/Loris/projects/14

If anyone wants to work on this issue, look there.

johnsaigle:
Closing this in favour of the Project

PR #4153, opened 2018-11-28
Fix some mismatched return type errors flagged by phan

johnsaigle:
### Brief summary of changes

Fixes some mismatched return type declarations and PHPDocs discovered in #4069.

### This adresses issue...

- [ ] Github? #4070

### DIFF
--- a/src/Http/StringStream.php
+++ b/src/Http/StringStream.php
@@ -70,7 +70,11 @@ class StringStream implements \Psr\Http\Message\StreamInterface, RequestHandlerI
      */
     public function close()
     {
-        return fclose($this->stream);
+        if (fclose($this->stream) === false) {
+            throw new LorisException(
+                'Could not close stream!'
+            );
+        }
     }
 
     /**
--- a/htdocs/survey.php
+++ b/htdocs/survey.php
@@ -142,21 +142,22 @@ class DirectDataEntryMainPage
      *
      * @param integer $currentPage The current page number
      *
-     * @return The page which preceeded this one
+     * @return int The page which preceeded this one
      */
-    function getNextPageNum($currentPage)
+    function getNextPageNum($currentPage): int
     {
-        $DB = Database::singleton();
         if ($currentPage === null) {
             return 1;
         }
-        return $DB->pselectOne(
-            "SELECT Order_number FROM instrument_subtests
-            WHERE Test_name=:TN AND Order_number > :PN 
-            ORDER BY Order_number",
-            array(
-             'TN' => $this->TestName,
-             'PN' => $currentPage,
+        return intval(
+            \Database::singleton()->pselectOne(
+                "SELECT Order_number FROM instrument_subtests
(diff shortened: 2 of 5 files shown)


src/Http/StringStream.php
```diff
@@ -66,7 +66,7 @@ public function __toString()
     /**
      * Implements the PSR7 StreamInterface and closes the underlying stream.
      *
-     * @return void
+     * @return bool
```
driusan: The definition in PSR7 declares that this needs to return void.

https://www.php-fig.org/psr/psr-7/#34-psrhttpmessagestreaminterface

src/Middleware/MiddlewareChainerMixin.php
```diff
@@ -43,9 +43,11 @@
      *
      * @param MiddlewareChainer $next The middleware to append
      *
-     * @return MiddlewareChainer A new middleware queue with $next appended
+     * @return MiddlewareChainerMixin A new middleware queue with $next appended
      */
-    public function withMiddleware(MiddlewareChainer $next) :MiddlewareChainer
+    public function withMiddleware(
+        MiddlewareChainer $next
+    ): MiddlewareChainerMixin
```
driusan: This change is not correct. This trait is a helper for implementing the MiddlewareChainer interface, which returns a MiddlewareChainer, not a MiddlewareChainerMixin.

src/Http/StringStream.php
johnsaigle: Phan marks it as incorrect because it actually returns a bool. Should we change what the function returns to match PSR or suppress the phan warning?

src/Middleware/MiddlewareChainerMixin.php
johnsaigle: Phan seems to think it returns a Mixin. Should the warning be suppressed?

src/Http/StringStream.php
driusan: we should change what it returns to match the standard. (It should be pretty easy, check if it's false and throw an exception if it is.)

src/Middleware/MiddlewareChainerMixin.php
driusan: That's only because traits aren't allowed to declare that they implement an interface in PHP (even though in cases like this one it fully implements it), and PHP requires interfaces to be explicitly implemented with the `implements` keyword.

Is this the only thing preventing one of the currently ignored classes of warnings from being added to phan's config/Travis?

johnsaigle: I reverted my changes to the file

driusan: Until it's the last error and holding up the check being automated I don't think there any reason to suppress it, that would just make it more likely to be missed if something were to change that makes it fixable (such as a hypothetical future version of PHP that allowed traits to implement interfaces, or a bug fix to phan that does a better job dealing with this type of situation.)

PR #4157, opened 2018-11-28
[ImgBrowser, DicomArchive] Prepend PatientName to DICOM download filename

johnsaigle:
### Brief summary of changes

To make it more readable/recognizable to users, the DICOM tarball filename when downloaded will be prefixed with the PatientName, i.e. `PSCID_DCCID_VisitLabel_....tar`.
E.g. For participant PSCID _BBB0112_ from [demo.loris.ca](https://demo.loris.ca/963443), the downloaded DICOM tarball for _V01_ will be named `BBB0112_963443_V01_DCM_2009-05-27_...1156014.tar`

Previously, this tarball download matched the name created by the LORIS-MRI upload/insertion pipeline, starting with the prefix "DCM".

Note the actual DICOM tarball on the server's file system and listed in the DICOM Archive module will not be renamed by this change or prefixed with the PatientName; this affects only the copy of the file that is downloaded to the user's machine.

Taking over #3853.


johnsaigle:
@christinerogers @cmadjar Following the feedback on the original PR I used a simpler approach that just adds the patient name field to the end of the name of the file being downloaded. Let me know if I understood correctly and if not what is desired instead.

I've tried my best to test but my sandbox doesn't have a proper imaging set-up.

johnsaigle:
@cmadjar Thanks for testing. I'm having a little trouble with this because the links don't show up at all for me.. I think the imaging portion of my Raisinbread is broken. I'll look into this and see what I can do.

modules/imaging_browser/templates/imaging_session_controlpanel.tpl
```diff
@@ -37,7 +37,7 @@
         {/foreach}
         {foreach from=$subject.tarchiveIDLoc key=tarchive item=tarchiveLoc}
             <li><a href="{$baseurl}/dicom_archive/viewDetails/?tarchiveID={$tarchive}&backURL={$backURL|escape:"url"}">DICOM Archive {$tarchive}</a></li>
-            <li><a href="/mri/jiv/get_file.php?file={$tarchiveLoc}" class="btn btn-primary btn-small">
+            <li><a href="/mri/jiv/get_file.php?file={$tarchiveLoc['ArchiveLocation']}&patientName={$tarchiveLoc['patientName']}" class="btn btn-primary btn-small">
```
johnsaigle:
```suggestion
            <li><a href="/mri/jiv/get_file.php?file={$tarchiveLoc['ArchiveLocation']}&patientName={$tarchiveLoc['PatientName']}" class="btn btn-primary btn-small">
```

johnsaigle:
@cmadjar I think I fixed the problem. Please retest and let me know if it works.

christinerogers:
@johnsaigle Thanks for this and LGTM given Cecile's tests and given the filename of the download matches what Muhammad intended. 
To request a meta-change before merging : Copy from the original PR header/comments a summary of the change -- so that the exact filename formulation and application (only on the downloaded file) is clear from this PR header.

As a follow-up: this change should also be applied to the Dicom Archive module where the same tar can also be downloaded.  
@cmadjar I'll add an issue, and can raise this at the next imaging meeting if pre-discussion is warranted.

modules/imaging_browser/php/imaging_session_controlpanel.class.inc
```diff
@@ -83,6 +83,7 @@ class Imaging_Session_ControlPanel
                  'v_sessions_id' => $this->sessionID,
                 )
             );
+            print_r($qresult);
```
cmadjar: Hehe, looks like you left a print here :)

screen shot 2018-11-30 at 10 45 56 am

modules/imaging_browser/php/imaging_session_controlpanel.class.inc
```diff
@@ -105,23 +112,25 @@ class Imaging_Session_ControlPanel
...
-            "SELECT TarchiveID, ArchiveLocation 
-            FROM tarchive 
-            WHERE PatientName LIKE $ID",
+            'SELECT TarchiveID, ArchiveLocation, PatientName ' .
+            'FROM tarchive ' .
+            'WHERE PatientName LIKE ' . $ID,
             $params
         );
+        print_r($ID);
+        print_r($qresult);
```
cmadjar: 2 other prints that were left in your PR.

cmadjar:
@johnsaigle @christinerogers Good idea to have the PatientName before the Archive name!

johnsaigle:
@cmadjar @christinerogers  Thanks for the testing and feedback! I've incorporated your feedback. I've also added the same functionality to the `dicom_archive` module. The "Archive Location" column still displays the actual file location on the server (i.e. without the patient name pre-pended) for accuracy. however the downloaded file itself will include the patient name.

christinerogers:
Just tweaked the PR title because DICOM mention is important, and PatientName is a thing.

johnsaigle:
@christinerogers If you're satisfied with the changes would you mind changing your review?

johnsaigle:
@christinerogers I applied your change and also updated the PR description.

@cmadjar Your approving review was auto-dismissed, unfortunately.

Would one of you be able to give this the green checkmark so it can be merged?

christinerogers:
One follow-up : @johnsaigle could you update the TestPlan and the Readme for both modules?

PR #4162, opened 2018-11-28
[Core: Module] Make debugging more verbose

johnsaigle:
### Brief summary of changes

While working on #3881 I kept running into a problem where I would have a bug but all the error log would display was something like 
```
PHP Fatal error:  Uncaught NotFound: media in /var/www/loris/php/libraries/Module.class.inc:81
```
and a generic stack trace. As a result normal PHP errors got swallowed up by the NotFound exception making it very hard to determine what was actually going wrong.

This change logs the entire path that the Module class tries to resolve and also explains what's happening behind the scenes.

### DIFF
--- a/php/libraries/Module.class.inc
+++ b/php/libraries/Module.class.inc
@@ -77,7 +77,9 @@ class Module extends \LORIS\Router\PrefixRouter implements RequestHandlerInterfa
                     . ".class.inc";
                     if (!file_exists($fpath)) {
                         throw new \LorisException(
-                            "Could not resolve $fpath to a LORIS module."
+                            "Could not resolve $fpath to a LORIS module. " .
+                            "Note that PHP filenames within modules must be " .
+                            "entirely lowercase."
                         );
                     }
                     include $fpath;
--- a/php/libraries/Module.class.inc
+++ b/php/libraries/Module.class.inc
@@ -77,9 +77,8 @@ class Module extends \LORIS\Router\PrefixRouter implements RequestHandlerInterfa
                     . ".class.inc";
                     if (!file_exists($fpath)) {
                         throw new \NotFound(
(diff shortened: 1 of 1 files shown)


php/libraries/Module.class.inc
```diff
@@ -76,7 +76,11 @@ class Module extends \LORIS\Router\PrefixRouter implements RequestHandlerInterfa
                     . strtolower(substr($class, strlen("LORIS\\$name\\")))
                     . ".class.inc";
                     if (!file_exists($fpath)) {
-                        throw new \NotFound($name);
+                        throw new \LorisException(
+                            "Could not resolve $fpath to a LORIS module. " .
+                            "Note that PHP filenames within modules must be " .
+                            "entirely lowercase."
+                        );
```
driusan: This is both an assumption about what the problem is, and a much less specific exception than was previously thrown.

johnsaigle: In my example when a file was named `MediaFileDownloadValidator.class.inc` the error generated was `Uncaught NotFound: media` even though `media` itself was fine. I had to read the source code and change the original file to `mediafiledownloadvalidator` to move forward. The exception text didn't help at all and was very misleading.

Can you suggest something you'd prefer?

driusan: The problem is that it couldn't find the module. In *your* case that happened to be because of case sensitivity, what happens in all the other situations that can cause a file_exists check to fail?

johnsaigle: I can remove the part about modules needing to be lowercase. Are you fine with the rest of it?

johnsaigle: As discussed in person, will change this error to display both the name and the fpath

PR #4165, opened 2018-11-29
[DICOM archive] Ensure that settings necessary for anonymization exist

johnsaigle:
### Brief summary of changes

While hunting down a PHP Warning I discovered that `dicomarchiveanonymizer.class.inc` does not do any validation on the config settings it uses. This results in empty values being sent to `preg_match` with possible consequences where DICOMs are not properly anonymized. I did not actually discover a case where this is true but my dummy imaging data is very limited.

This PR adds some validation to `dicom_archive.class.inc`. It will now refuse to show DICOM data when the regex config values are not properly configured. 

### This resolves issue...

Should eliminate warnings like
`
PHP Warning:  preg_match(): Empty regular expression in /var/www/loris/modules/dicom_archive/php/dicomarchiveanonymizer.class.inc on line 62, referer: https://jsaigle-dev.loris.ca/dicom_archive/
`

### To test this change...

- [ ] On another branch, go to DICOM archive and check your error logs. See if you get the above warning.
* If so, check out my branch and you should see a new error message on the front-end as well as a ConfigurationException in the logs. 
   * Now change the Regex ConfigSettings (see source code) to valid regular expressions.
    * Navigate back to DICOM archive and ensure the page loads normally
* If you do not get a warning, delete one of the Regex Config Settings and follow the steps above.

### DIFF
--- a/modules/dicom_archive/php/dicom_archive.class.inc
+++ b/modules/dicom_archive/php/dicom_archive.class.inc
@@ -14,7 +14,6 @@
  * @link       https://www.github.com/aces/Loris-Trunk/
  */
 namespace LORIS\dicom_archive;
-use \Psr\Http\Server\RequestHandlerInterface;
 use \Psr\Http\Message\ServerRequestInterface;
 use \Psr\Http\Message\ResponseInterface;
 
--- a/modules/dicom_archive/php/dicom_archive.class.inc
+++ b/modules/dicom_archive/php/dicom_archive.class.inc
@@ -14,8 +14,6 @@
  * @link       https://www.github.com/aces/Loris-Trunk/
  */
 namespace LORIS\dicom_archive;
-use \Psr\Http\Message\ServerRequestInterface;
-use \Psr\Http\Message\ResponseInterface;
 
 /**
  * Provides the PHP code for the menu filter for the dicom archive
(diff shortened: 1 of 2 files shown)


modules/dicom_archive/php/dicom_archive.class.inc
```diff
@@ -106,6 +167,9 @@ class Dicom_Archive extends \NDB_Menu_Filter
      */
     function toJSON()
     {
+        /* Ensure that the server is well configured to prevent displaying
+         * data that is not properly anonmyized.
+         */
```
driusan: This comment seems to be out of place

modules/dicom_archive/php/dicom_archive.class.inc
```diff
@@ -98,6 +100,65 @@ class Dicom_Archive extends \NDB_Menu_Filter
...
+        /* Ensure that all required configuraton settings are properly set
+         * before continuing.
+         */
+        $config        = \NDB_Config::singleton()->getSetting("imaging_modules");
+        $configOptions = array(
+                          'patientNameRegex',
+                          'LegoPhantomRegex',
+                          'LivingPhantomRegex',
+                          'patientIDRegex',
+                         );
```
driusan: I don't see any reference to these in this file. I only see them referenced from the view session page, so I don't think this is the right place for this.

Unless I'm missing something, I think this should either:

1. Be moved to the viewsession page, where the configs are used and it just needs to add a !empty check or
2. Be checked by the Module class for the module, not the menu  page's class.

johnsaigle: Theyr'e used by `dicomarchiveanonymizer.class.inc` (I mentioned this in my description but maybe I wasn't clear). Would you prefer that the checks go there instead?

Alternatively I can make the comment above more clear about where those settings are used.

driusan: They should checked close to the thing that they're validating.

The DicomArchiveAnonymizer constructor seems reasonable, or in the module's Module class.

modules/dicom_archive/php/dicom_archive.class.inc
driusan: This comment is still here

PR #4166, opened 2018-11-29
 [Core: Middleware] Restore 403 page to its former beauty 

### LINKED ISSUE #3908: 403 page doesn't render in pretty template
### Previously

We had a pretty 403 page that rendered in the LORIS template (header, footer, logo etc.) and displayed the following message.

previously

### Now
Ugly!
now

This issue seems to have been introduced in 20.0.0.

johnsaigle:
### Brief summary of changes

Makes the 403 page pretty. See linked issue.

### This resolves issue...

Resolves #3908

### DIFF
--- a/src/Middleware/AuthMiddleware.php
+++ b/src/Middleware/AuthMiddleware.php
@@ -62,12 +62,16 @@ class AuthMiddleware implements MiddlewareInterface, MiddlewareChainer
             return $this->next->process($request, $handler);
         }
 
-        return new \LORIS\Middleware\PageDecorationMiddleware(
+        return (new \LORIS\Middleware\PageDecorationMiddleware(
             $request->getAttribute("user") ?? new \LORIS\AnonymousUser()
         ))->process(
             $request,
             new \LORIS\Router\NoopResponder(
-                new \LORIS\Http\Error($request, 403, "Permission denied")
+                new \LORIS\Http\Error(
+                    $request,
+                    403,
+                    "You do not have access to this page."
+                )
             )
         );
     }
--- a/src/Middleware/AuthMiddleware.php
+++ b/src/Middleware/AuthMiddleware.php
@@ -62,6 +62,13 @@ class AuthMiddleware implements MiddlewareInterface, MiddlewareChainer
             return $this->next->process($request, $handler);
         }
 
-        return (new \LORIS\Http\Error($request, 403, "Permission denied"));
+        return new \LORIS\Middleware\PageDecorationMiddleware(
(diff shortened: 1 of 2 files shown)


src/Middleware/AuthMiddleware.php
```diff
@@ -62,6 +62,17 @@ public function process(
...
+        ))->process(
+            $request,
+            new \LORIS\Router\NoopResponder(
+                new \LORIS\Http\Error(
+                    $request,
+                    403,
+                    "You do not have access to this page."
+                )
+            )
+        );
```
driusan: I don't think this is right. The middleware should already have been injected here:

https://github.com/aces/Loris/blob/minor/src/Router/BaseRouter.php#L136-L140

and the Auth shouldn't be injecting HTML to chains that didn't ask for it (such as API endpoints.)

johnsaigle: The code that you linked to seems to only cover 404 errors. 

Is the BaseRouter class a better place for this fix then? I can update this PR if so.

johnsaigle: Thanks for clarifying this. I think the logic is broken though. I believe this is why the 403 page isn't being decorated. https://github.com/aces/Loris/blob/15c0a2790843b089771e84fe8adf7ede2b53e750/php/libraries/Module.class.inc#L254-L264.

It seems like the `Module.class.inc` routing is only occurring for public modules (which is almost none of them, I believe) and this is why the raw, undecorated response from `AuthMiddleware.php` is being returned.

johnsaigle: Sorry, let me rephrase that actually. It's not that the logic is wrong necessarily, but that it isn't able to return a decorated response in the above code.

Maybe line 94 just needs to be changed to return a decorated response instead. However if you want the code to reach the Module routing then the logic will need to be changed.

driusan: Are you logged in when having this problem? The module routing should happen for both public and non-public modules.. but what seems to be happening is that on line 61 it's returning false if you're both not logged in and the module isn't public, and aborting calling the module at all.

(In which case this is the proper place to re-add the page decoration afterall..)

johnsaigle: You're right! Yes I was testing it from the perspective of a logged-out user. The module routing/decoration happens correctly for logged-in users without access to a given page. 

I'll modify line 94 to return a decorated response for logged-out users.

driusan: But that's not getting called because of line 61 here.. this is fine the way it is, I think.

PR #4112, opened 2018-11-15
[Issue Tracker] Reactify Menu Filter + Add DataFramework + Update FilterableDataTable

HenriRabalais:
This PR:
1. Reactifies the Issue Tracker Menu Filter
 - I ran into a lot of issues doing this. First of all, the `watching` filter necessitates the checkbox element which was only in minor, so I had to pull those changes. Secondly, to be able to properly convert `UserID` to a boolean, I had to implement to the dataframework to map the value. I also wanted to replace the unnecessary duplication of code for each tab of the menu filter, so instead I made the tabs pass filter objects to the FilterableDataTable Component.
2. Adds the Data FrameWork to Issue Tracker
- As mentioned above, I was lead to do this because of the issues I encountered when reactifying the menu filter. Since this needs to be done anyways, I figured I would go ahead and do it.
3. Updates FilterableDataTable
- Due to the need to have what are essentially 'filter presets', I added the ability to pass those to the Filterable DataTable. Also, since an action button was necessary to add issues, I allowed those to be passed as props as well. While doing this, I realized the configuration of the filter and datatable discussed in the UI meetings would be much more suitable for this sort of structure, and so I converted it to something that resembles the expected layout.
4. Cleans up the DataTable
- The DataTable was in need of some deep cleaning, and was giving me quite a few issues. I pulled up as much state as I could to the FilterableDatatable so that it could be in control of everything regarding the filter and datatable. Eventually it would be useful to have the filterableDatatable do the filtering itself, and that way the Datatable can simple display the data.

Let me know your thoughts.
issue tracker

PR #4167, opened 2018-12-01
[UserAccounts] Reactify Index

zaliqarosli:
### Brief summary of changes
- add ability to filter for multiple strings i.e. for a user that has multiple sites
- regular ol' reactification
- updates test plan and script accordingly

screenshot 2018-12-14 14 07 37

### To test this change...

- [ ] Follow the module's test plan included in this PR


kongtiaowang:
https://github.com/kongtiaowang/Loris/pull/157
Please use the new react test instead of the old one.

driusan:
Do you have a screenshot of what this CTA looks like?

zaliqarosli:
@driusan let me know what you think
screenshot 2018-12-03 15 54 27

modules/user_accounts/jsx/userAccountsIndex.js
```diff
@@ -0,0 +1,178 @@
...
+   *
+   * @return {object}
+   */
+  fetchData() {
+    return fetch(this.props.dataURL, {credentials: 'same-origin'})
+      .then((resp) => resp.json())
+      .then((data) => {
+        // Convert concatenated string of sites to array
+        for (let key in data.Data) {
+          if (data.Data.hasOwnProperty(key)) {
```
HenriRabalais: Since i believe `data.Data` is an array, you can do something like: 
```
data.Data = data.Data.map((row) => {
  row[0] = row[0].split('; '); // sites are the first element of the row array
  return row;
}
```
This avoids the need for the `hasOwnProperty` check, and is more in line with how javascript deals with arrays.

jsx/DataTable.js
```diff
@@ -309,12 +309,26 @@ class DataTable extends Component {
...
       searchKey = filterData.toLowerCase();
-      searchString = data.toLowerCase();
-
-      if (exactMatch) {
-        result = (searchString === searchKey);
-      } else {
-        result = (searchString.indexOf(searchKey) > -1);
+      switch (typeof data) {
+        case 'object':
+          // Handles the case where the data is an array
```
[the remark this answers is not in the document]
zaliqarosli: apparently in javascript, arrays are objects so the typeof operator returns "object" for arrays. maybe i should add this as a comment?

jsx/DataTable.js
```diff
@@ -309,12 +309,26 @@ class DataTable extends Component {
...
+          // and you want to search through it for
+          // the string you are filtering by
+          let searchArray = data.map((e) => e.toLowerCase());
+          if (exactMatch) {
+            result = searchArray.includes(searchKey);
+          } else {
+            result = searchArray.forEach((e) => (e.indexOf(searchKey) > -1));
```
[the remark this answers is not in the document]
zaliqarosli: mmm your suggestion didn't work. i essentially want to check `searchString.indexOf(searchKey) > -1` for each element of the searchArray. how would i do that?

zaliqarosli:
@HenriRabalais please rereview!

driusan:
I'm not sure I like the location of the Add User button because it adds yet another non-standardized location for action buttons. Can we discuss at the next UX meeting?

HenriRabalais:
My Issue Tracker PR adds the ability to pass `actions` to the `FilterableDataTable` that will be able to handle the placement of action buttons in a standardized way within the filter (#4112).

driusan:
Blocked because it requires a confirmed decision from the UX group about where/how to handle CTAs, what differentiates a card in the UX, how to pass them, etc. (Also depends on #4112)

zaliqarosli:
update: FilterableDataTable redesign implemented in this PR. the code in this PR includes code from https://github.com/aces/Loris/pull/4203. once #4203 has been merged, I will rebase this PR

screenshot 2018-12-14 14 07 37

kongtiaowang:
Please fix the conflict files and I will create a the new test case for your PR.

driusan:
This now has conflicts

driusan:
> see #4203. this PR includes code from #4203 until it gets merged, after which point this PR will be rebased

#4203 is merged.

kongtiaowang:
@zaliqarosli 
https://github.com/kongtiaowang/Loris/commit/40259af552386abed4c312aa6b9eab43a92c42f6

modules/user_accounts/jsx/userAccountsIndex.js
```diff
@@ -0,0 +1,174 @@
...
+        options: options.pendingApprovals,
+      }},
+    ];
+    const actions = [
+      {name: 'addUser', label: 'Add User', action: this.addUser},
+    ];
+
+    return (
+      <div id="userAccountsIndex">
+        <div id="userAccountsFilter">
```
HenriRabalais: What's the purpose of these `id`s?

modules/user_accounts/test/user_accountsTest.php
```diff
@@ -104,50 +103,114 @@ function testUserAccountsMyPreferencesDoespageLoad()
...
-        $this->_assertSearchBy("userID", "UnitTester", "UnitTester");
+        $this->safeGet($this->url . "/user_accounts/");
+        //testing data from RBdata.sql
+        $this-> _testFilter($this->_name, $this->_table, null, "UnitTester");
+        $this-> _testFilter($this->_site, $this->_table, "1 rows", "3");
     }
+    /**
+     * Testing filter funtion and clear button
+     *
+     * @param string $element The input element loaction
```
HenriRabalais: `loaction` -> `location`

modules/user_accounts/test/user_accountsTest.php
```diff
@@ -104,50 +103,114 @@ function testUserAccountsMyPreferencesDoespageLoad()
...
+        //      'UnitTester',
+        //      'Email',
+        //      'newemail@example.com'
+        //  );
+        //  $this->_verifyUserModification(
+        //      'user_accounts',
+        //      'UnitTester',
+        //      'Pending_approval',
+        //      'No'
+        //  );
```
HenriRabalais: Why are these tests commented out?

modules/user_accounts/jsx/userAccountsIndex.js
zaliqarosli: for description of the divs

modules/user_accounts/test/user_accountsTest.php
zaliqarosli: because @kongtiaowang amended it that way

modules/user_accounts/jsx/userAccountsIndex.js
HenriRabalais: Okay, but how are they being used? `FilterableDataTable` already receives a name prop which is used to set an `id` on the filter. This index file is already a `UserAccountsIndex` component, so why is there a need for a description? Are these `id`s used as targets for testing? We should only put `id`s that are required by the needs of our current application.

zaliqarosli: when inspecting elements, I find it useful to have ids that literally identify which divs are for what. in this case, that's what the ids are: to name the divs. they don't provide any function per say but I don't see why that's a problem.  the initial thinking was to use the ids as selectors in testing, but now that's no longer necessary. however, in the future, we might need to target those divs for some reason, and then the ids are already there to select

screenshot 2019-01-14 13 12 41

what reason is there for not allowing ids unless they are required for targets?

HenriRabalais: I'm not sure that ids are designed to name the divs. According to W3Schools they are used to specifically target elements for use by css stylesheets, javascript, etc. for manipulation. The resources I've looked into recommend that if they are not necessary, don't use them. If we aren't limiting what elements should have `id`s to those that necessitate them, then `id`s could be placed everywhere. If they are necessary in the future, they can be added then! 

https://css-tricks.com/the-difference-between-id-and-class/
https://stackoverflow.com/questions/42384788/should-we-always-specify-an-id-attribute-in-html
https://stackoverflow.com/questions/16153986/adding-id-attribute-to-all-html-elements-of-a-web-application

zaliqarosli:
@HenriRabalais IDs removed, please re-review

johnsaigle:
Same Gender/Sex thing. Can you rebase this first? I'll test after.

zaliqarosli:
@johnsaigle PR was already rebased today to include latest changes in major. have you applied the patch from Gender/Sex PR?

zaliqarosli:
@johnsaigle thank u :)

modules/user_accounts/jsx/userAccountsIndex.js
```diff
@@ -0,0 +1,174 @@
...
+        options: options.pendingApprovals,
+      }},
+    ];
+    const actions = [
+      {name: 'addUser', label: 'Add User', action: this.addUser},
+    ];
+
+    return (
+      <div>
+        <div>
```
driusan: What are these 2 divs for?

zaliqarosli: i guess nothing 😂😂i'll remove them

zaliqarosli:
@driusan i removed the useless divs! please re-review

jsx/DataTable.js
```diff
@@ -552,7 +568,7 @@ class DataTable extends Component {
     );
 
     return (
-      <div style={{margin: '14px'}}>
+      <div name={this.props.name} style={{margin: '14px'}}>
```
driusan: "name" in HTML is for attaching a name to submit data from an HTML form as. I'm not sure what it means to add it to a div.

zaliqarosli: i think i can get rid of this. good catch!

PR #4168, opened 2018-12-03
[deps] min phan version

Jkat:
Anyone know if this will break something? I think as our systems are getting newer, we're no longer installing low versions of ast which phan 0.12.x requires...also isn't phan 0.12.x a bit old?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "phpunit/dbunit": "3.0.*",
         "facebook/webdriver" : "dev-master",
         "phpmd/phpmd": "^2.6",
-        "phan/phan": "0.12.x"
+        "phan/phan": ">=0.12.0"
     },
     "scripts": {
       "pre-install-cmd": "mkdir -p project/libraries"


johnsaigle:
@ridz1208 Can you comment here? I think I remember you mentioning this being a problem earlier. Will this fix it?

johnsaigle:
This seems to have phan errors. I guess updating the version introduces more checks that we are not passing.

Also I think it's probably safe to update this to `1.0` since that's what `major` is using.

johnsaigle:
I don't think it's documented. It's not yet part of the standard LORIS install though I have amended this in the proposed (and probably cursed) [update script](https://github.com/aces/Loris/pull/3677) back in May.

My solution there was to install it using `apt` though I know @ridz1208 has used `pecl` with some difficult. Haven't looked into CentOS at this time. 

This in general is part of a [series](https://github.com/aces/Loris/pull/3981) of [related](https://github.com/aces/Loris/pull/3677) [efforts](https://github.com/aces/Loris/pull/4002)/larger discussion of installation/bootstrapping/updating. From what I understand this is to be discussed in roadmap meetings. Maybe @driusan or @ridz1208 could shed some light here but I believe these issues are still unresolved.

Jkat:
cool cool, and just a small note for context, i ran into this issue on CentOS 7

driusan:
the ast extension isn't required for LORIS, it's just required for development. It can't be installed by composer because it's a PHP extension, not a PHP package.

Jkat:
are we documenting the version for the ast dependency somewhere for developers? if im not mistaken developers are supposed to install the ast dependency externally(as in not part of the standard loris install process), which seems to require certain versions depending on what version of phan is being used.
(sorry wasn't being clear earlier)

johnsaigle:
@jkat I don't we are but we should be. Maybe in the process of this wiki revamp we can include post-install notes for developer instances.

driusan:
That's a good question.. @ridz1208 is the php-ast version documented anywhere in the wiki revamp.

ridz1208:
@driusan shouldnt it be documented somewhere in the readme ? since its a dependency (even if its for developers... it could be in a devs section)

Jkat:
(this PR can be closed btw once we decide on a documentation solution)

johnsaigle:
@ridz1208 Could you give a Wiki/documentation update in a LORIS meeting sometime soon?

driusan:
#4172 seems like a more appropriate place to have this discussion rather than keeping this PR open forever.

PR #4169, opened 2018-12-03
[Core] Add "npm install" to make's "dev" target

johnsaigle:
### Brief summary of changes

When `make`ing a dev environment it makes sense to install the node modules you need. Let me know if there's a reason not to do this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,5 +5,6 @@ all:
 
 dev:
 	composer install
+	npm install
 	npm run compile
 	git describe --tags --always > VERSION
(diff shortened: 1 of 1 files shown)


HenriRabalais:
Had to do some research on exactly how `npm install` should be used, and after doing so this seems logical to me!

johnsaigle:
@HenriRabalais I don't know much about it! I just know that it is supposed to install what's in `package.json` and I found that this wasn't happening for me even when running `make dev` when switching to major.

driusan:
Doesn't it also need to go into the all target?

PR #4170, opened 2018-12-03
[Core] Add babel-core as a dependency on minor branch

johnsaigle:
### Brief summary of changes

Should fix errors like `ERROR in Cannot find module 'babel-core'` when compiling React files on the `minor` branch. 

Issuing this to `minor` because I think we'll probably have a minor release soon and so this will bubble up into future bugfix releases.


zaliqarosli:
will approve after conflict resolved

ridz1208:
@johnsaigle the sweetalert version was changed recently

PR #4173, opened 2018-12-03
 [20.1] Creation or release SQL patch and archival of patches

ridz1208:


johnsaigle:
This seems to have a lot of unrelated changes included by accident.

ridz1208:
@johnsaigle it's because its based on the other PR that merged 20.1 into bugfix
once the other one is in this will be just the last commmit

PR #4174, opened 2018-12-03
[Core: BVL_Feedback] Fix error that makes LORIS crash when accessing candidate info on `major` branch

johnsaigle:
### Brief summary of changes

Merging #4124 caused some errors to appear in Travis and make it crash. This will make every PR to major fail and also means that no candidate info can be accessed.

This PR should fix those errors. For an example see this report: https://travis-ci.org/aces/Loris/jobs/462997840

### DIFF
--- a/php/libraries/BVL_Feedback_Panel.class.inc
+++ b/php/libraries/BVL_Feedback_Panel.class.inc
@@ -74,7 +74,7 @@ class BVL_Feedback_Panel
 
         $this->tpl_data['candID'] = $candidateID;
 
-        $candidateObject         = Candidate::singleton($candidateID);
+        $candidateObject         = Candidate::singleton(intval($candidateID)));
         $this->tpl_data['pscid'] = $candidateObject->getPSCID();
 
         $this->tpl_data['sessionID'] = $feedbackProfile["SessionID"] ?? null;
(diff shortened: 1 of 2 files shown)


johnsaigle:
Related to #4147. This was another type error that was only caught after merging and not during the code review phase. Not sure why.

johnsaigle:
@driusan just FYI I added the high-priority label because all candidate info is currently unavailable on `major`. On that branch go to access_profile and then click any candidate. LORIS crashes.

I get not wanting to overuse this label but I think it's probably justified here.

php/libraries/BVL_Feedback_Panel.class.inc
```diff
@@ -74,7 +74,7 @@ class BVL_Feedback_Panel
 
         $this->tpl_data['candID'] = $candidateID;
 
-        $candidateObject         = Candidate::singleton($candidateID);
+        $candidateObject         = Candidate::singleton(intval($candidateID)));
```
kongtiaowang: remove a ")"

driusan:
@johnsaigle I asked Rida to bring it up at the LORIS meeting, but people shouldn't be adding the high priority label themselves because it's intended to be used for roadmap planning.

johnsaigle:
@driusan Thanks for that clarification. Rida did bring it up and I have a better understanding of it now.

PR #4178, opened 2018-12-04
[Core: Utility] Fix type error in toArray

johnsaigle:
### Brief summary of changes

#4151 should be merged but this is causing problems until then.

### This resolves issue...

` PHP Fatal error:  Uncaught TypeError: Return value of Utility::toArray() must be of the type array, string returned in /var/www/loris/php/libraries/Utility.class.inc:437`

### DIFF
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -427,7 +427,7 @@ class Utility
      *
      * @param mixed $var The variable to be converted to an array.
      *
-     * @return array|string If $var is an array, $var, 
+     * @return array|string If $var is an array, $var,
      *                          otherwise an array containing $var
      */
     static function toArray($var)
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -427,9 +427,10 @@ class Utility
      *
      * @param mixed $var The variable to be converted to an array.
      *
-     * @return array If $var is an array, var, otherwise an array containing $var
+     * @return array|string If $var is an array, $var, 
+     *                          otherwise an array containing $var
      */
-    static function toArray($var): array
+    static function toArray($var)
     {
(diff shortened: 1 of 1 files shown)


php/libraries/Utility.class.inc
```diff
@@ -427,9 +427,10 @@ class Utility
      *
      * @param mixed $var The variable to be converted to an array.
      *
-     * @return array If $var is an array, var, otherwise an array containing $var
+     * @return array|string If $var is an array, $var,
+     *                          otherwise an array containing $var
      */
-    static function toArray($var): array
+    static function toArray($var)
```
PapillonMcGill: if $var is a string, the same string will be returned! 
```if (is_array($var) ...)``` will be false skipping to line 438 ```return $var```

johnsaigle: Yes, that's intentional.

PapillonMcGill: In that case, the docblock do not reflect the work done by this function.

johnsaigle: >      * Note: This does not change a string into an array with one string element

> @return array|string If $var is an array, $var,
     *                          otherwise an array containing $var

That's how I found out that this is intentional behaviour.

johnsaigle: Note that this PR is just a tiny bugfix and that #4151 addresses concerns about this confusing function.

zaliqarosli:
i don't understand why this PR is here. https://github.com/aces/Loris/pull/4151 is merged on the same branch as this PR and includes the code changes made here. the only difference is that the doc return type here is difference, but I don't see why that change wasn't just made in https://github.com/aces/Loris/pull/4151

johnsaigle:
@zaliqarosli #4151 was opened before this and I was trying to get this merged before hadn to fix the error. Anyway this change doesn't matter now, I'll close this PR

PR #4181, opened 2018-12-04
[Core] Create File Uploader class to standardize and validate file uploads in LORIS

johnsaigle:
### Brief summary of changes

This PR picks up on #3514 and introduces an attempt at creating a generic File Upload (Validator) class for use in LORIS file uploads. See attached issue for more details.

The newly Reactified `instrument_manager` is used as an example to demonstrate how this class should be used.

### Related

#4182.

### To test this change...

- [ ] Go to the instrument manager and try to upload a file. LINST files should succeed but files such as a PDF or JPG should fail.


johnsaigle:
### Some Design Choices

* I chose to validate against a whitelist of expected MIME types rather than using file extensions. 
    * File extensions are totally controllable by the user so validating against them in the back-end is not useful. 
    * MIME types are harder to lie about but require the developer of the module to know exactly what they are expecting

* The validator class has a property that stores the HTTP response code corresponding to what error occurred. This can be used to group Diactoros responses, as shown in the module

johnsaigle:
This needs documentation but is ready for testing.

php/libraries/File_Upload_Validator.class.inc
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * Class to validate file uploads to LORIS.
+ *
+ *  @category Main
+ *  @package  Behavioural
+ *  @author   John Saigle <john.saigle@mcin.ca>
+ *  @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ *  @link     https://www.github.com/aces/Loris
+ */
+class File_Upload_Validator
```
driusan: This should be FileUploadValidator

php/libraries/File_Upload_Validator.class.inc
```diff
@@ -0,0 +1,177 @@
...
+            throw new \LorisException(
+                'Upload directory is not a directory!'
+            );
+        }
+        if (! is_writable($this->uploadDirectory)) {
+            // Set __responseCode anyway in case some calling code decides to
+            // catch this exception.
+            $this->_responseCode = 403;
+            throw new \LorisException(
+                'Upload directory is not writable!'
```
driusan: That's a server error, not a client error.

php/libraries/File_Upload_Validator.class.inc
```diff
@@ -0,0 +1,177 @@
...
+     * @throws \LorisException
+     *
+     * @return void
+     */
+    protected function validate(): void
+    {
+        if (! is_dir($this->uploadDirectory)) {
+            $this->_responseCode = 500;
+            throw new \LorisException(
+                'Upload directory is not a directory!'
```
driusan: I don't think adding this restriction is a valid assumption, since often uploaders categorize things in some way (such as `dir/candid/filename` or `dir/uploaduser/filename`) and the directory may not exist.

php/libraries/File_Upload_Validator.class.inc
```diff
@@ -0,0 +1,177 @@
...
+        }
+        return;
+    }
+
+    /**
+     * Given an array of files, validate them and upload them to the server.
+     *
+     * @param array $uploadedFiles An array of files uploaded to the server.
+     *                              These files must implement the PSR-7
+     *                              UploadedFile interface.
```
driusan: Then the type of this should be UploadedFileInterface[] instead of having a type of any array and English prose to restrict the type.

php/libraries/File_Upload_Validator.class.inc
```diff
@@ -0,0 +1,177 @@
...
+     * @param int   $permissions   A UNIX permissions string (e.g. 644). All files
```
driusan: Unix permissions are octal (0644), not decimal. They're also not a string.

There should probably be some symbolic consts declared for common modes to make sure people don't make this mistake in code.

php/libraries/File_Upload_Validator.class.inc
```diff
@@ -0,0 +1,177 @@
...
+
+        foreach ($uploadedFiles as $uploadedFile) {
+            //Ensure that the $uploadedFile uses the PSR-7 UploadedFile
+            //interface
+            if (! $uploadedFile instanceof \Zend\Diactoros\UploadedFile) {
+
+                error_log(
+                    'File uploaded to File_Upload_Validator is not an ' .
+                    'instance of PSR-7 UploadedFIle. Aborting upload.'
+                );
```
driusan: This is an exception, not an error.

php/libraries/File_Upload_Validator.class.inc
```diff
@@ -0,0 +1,177 @@
...
+            /* basename() is used here to prevent path traversal characters
+             * from being used.
+             */
+            $targetPath = $this->uploadDirectory . basename(
+                $uploadedFile->getClientFilename()
+            );
+            // If file exists, set response code to 'Conflict'
+            if (file_exists($targetPath)) {
+                $this->_responseCode = 409;
+                return false;
```
driusan: How does someone use this if they're explicitly trying to overwrite something? (ie. in the imaging uploader when reuploading a file.)

php/libraries/File_Upload_Validator.class.inc
```diff
@@ -0,0 +1,177 @@
...
+                    $this->_responseCode = 400;
+                    return false;
+                }
+            }
+
+            // Try to move the file to $targetPath
+            try {
+                $uploadedFile->moveTo($targetPath);
+            } catch (\Exception $e) {
+                /* If an error occurs during moving, set response code to
```
driusan: An exception is not an error. An exception is an exception. I don't think this should be caught.

php/libraries/File_Upload_Validator.class.inc
```diff
@@ -0,0 +1,177 @@
...
+                 * 'Internal Server Error'.
+                 *
+                 * This covers file size checks as well as other errors that
+                 * can occur in the process of performing
+                 * `move_uploaded_file().
```
driusan: Whatever checks need to be done for error conditions should be done explicitly here.

php/libraries/File_Upload_Validator.class.inc
```diff
@@ -0,0 +1,177 @@
...
+                 *
+                 * See \Zend\Diactoros\UploadedFile for more information on
```
driusan: Any UploadedFileInterface can be passed to this function, this shouldn't be referring to one specific implementation. What if someone passed a different one?

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -30,24 +32,26 @@ use \LORIS\Http\EmptyStream;
...
                                   'not possible given the current server ' .
                                   'configuration. Please contact your ' .
                                   'administrator if you require this ' .
-                                  'functionality';
+                                  'functionality.';
 
     const FILE_ALREADY_EXISTS = 'This file has already been uploaded. ' .
                                   'Please contact your administrator to ' .
                                   'install the instrument';
+    const BAD_REQUEST         = 'Uploaded file is not of type LINST or is ' .
```
PapillonMcGill: Need ending dot at line 45.

php/libraries/FileUploadValidator.class.inc
```diff
@@ -0,0 +1,241 @@
...
+                'Upload directory is not writable'
+            );
+        }
+        return;
+    }
+
+    /**
+     * Given an array of files, validate them and upload them to the server.
+     *
+     * The permission paramaeter is a UNIX file permissions octet.
```
PapillonMcGill:
```suggestion
     * The permission parameter is a UNIX file permissions octet.
```

php/libraries/FileUploadValidator.class.inc
```diff
@@ -0,0 +1,241 @@
...
+                $uploadedFile->getClientFilename()
+            );
+            /* If file exists, set response code to 'Conflict' unless the
+             * calling code wants to overwrite the file.
+             */
+            if (file_exists($targetPath) && $overwrite === false) {
+                $this->_responseCode = 409;
+                return false;
+            }
+            // Validate theat the MIME type of the file is allowed
```
PapillonMcGill:
```suggestion
            // Validate that the MIME type of the file is allowed
```

ridz1208:
https://github.com/aces/Loris/pull/3514

php/libraries/FileUploadValidator.class.inc
```diff
@@ -0,0 +1,241 @@
...
+        bool $overwrite = false
+    ) : bool {
+        if (empty($uploadedFiles)) {
+            return false;
+        }
+
+        foreach ($uploadedFiles as $uploadedFile) {
+            //Ensure that the $uploadedFile uses the PSR-7 UploadedFile
+            //interface
+            if (! $uploadedFile instanceof \Zend\Diactoros\UploadedFile) {
```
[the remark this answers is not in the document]
johnsaigle: What type should it have?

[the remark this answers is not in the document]
driusan: `\Psr\Http\Message\UploadedFileInterface` (as in the change suggestion with the comment)

php/libraries/FileUploadValidator.class.inc
```diff
@@ -0,0 +1,241 @@
...
+     *
+     * @param UploadedFile[] $uploadedFiles An array of files uploaded to the
+     *                              server. These files must implement the PSR-7
+     *                              UploadedFile interface.
+     * @param int            $permissions   A UNIX permissions octet (e.g. 0644)
+     * @param bool           $overwrite     Whether existing files should be
+     *                                      overwritten.
+     *
+     * @return bool Whether the file upload was successful. Modifies
+     *                      $this->__responseCode as a side-effect.
```
johnsaigle:
```suggestion
     *                      $this->_responseCode as a side-effect.
```

php/libraries/FileUploadValidator.class.inc
```diff
@@ -0,0 +1,241 @@
...
+     *
+     * @param string $size A string representing an amount of memory.
+     *
+     * @return int An integer corresponding to the number of bytes in the
+     *              byte string
+     */
+    function parseSize(string $size): int
+    {
+        // Remove the non-unit characters from the size.
+        $unit = preg_replace('/[^bkmgtpezy]/i', '', $size);
```
xlecours: if this line was `$unit = preg_replace('/[^bkmgtpezy]/i', '', $size)[0] ?? '';` the `if` statement could be removed and always return `return round($size * pow(1024, stripos('bkmgtpezy', $unit)));`

php/libraries/FileUploadValidator.class.inc
```diff
@@ -0,0 +1,241 @@
...
+        $size = preg_replace('/[^0-9\.]/', '', $size);
+        if ($unit) {
+            /* Find the position of the unit in the ordered string which is
+             * the power of magnitude to multiply a kilobyte by.
+             *
+             * phan thinks that the arguments to stripos are in the wrong
+             * order but this is a false positive.
+             */
+            // @phan-suppress-next-line PhanParamSuspiciousOrder
+            return round($size * pow(1024, stripos('bkmgtpezy', $unit[0])));
```
xlecours: This is elegant.

php/libraries/FileUploadValidator.class.inc
```diff
@@ -0,0 +1,241 @@
...
+/**
+ * Class to validate file uploads to LORIS.
+ *
+ *  @category Main
+ *  @package  Behavioural
+ *  @author   John Saigle <john.saigle@mcin.ca>
+ *  @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ *  @link     https://www.github.com/aces/Loris
+ */
+class FileUploadValidator
```
xlecours: `FileUploader` would be enough. The Validator part is implicit in the constructor code.
I would not expect a `FileUploadValidator` to have a upload function. (This is debatable)

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -112,62 +116,81 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
-                );
-        }
         $uploaded_file = $request->getUploadedFiles()['install_file'];
         $filename      = $uploaded_file->getClientFilename();
         $instrument    = pathinfo($filename)['filename'];
+        // Do pre-processing before uploading files.
+        $err = '';
+        if (!$this->canWriteFiles()) {
+            $err    = self::CANNOT_WRITE_FILES;
+            $status = 503;
```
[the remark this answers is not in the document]
johnsaigle: This case shouldn't even come up really... I think I added this error handling a while ago but the module has changed a lot since then. The `projects/tables_sql/` and `projects/instruments` directories should always be writable by the Apache user. When https://github.com/aces/Loris/pull/4191 is merged it should take care of permissions checking and in that case this should really be a `500` error, or even a ConfigurationException.

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -112,62 +116,81 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
-                        )
-                    )
-                );
-        }
         $uploaded_file = $request->getUploadedFiles()['install_file'];
         $filename      = $uploaded_file->getClientFilename();
         $instrument    = pathinfo($filename)['filename'];
+        // Do pre-processing before uploading files.
+        $err = '';
+        if (!$this->canWriteFiles()) {
```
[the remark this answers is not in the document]
johnsaigle: No real reason, I guess I just wanted to group all of the error-checking code together.

php/libraries/FileUploadValidator.class.inc
```diff
@@ -0,0 +1,241 @@
...
+     * required to have differing permissions, this must be done
+     * manually in the calling code after this function has executed.
+     *
+     * If the $overwrite parameter is set to true, existing files will be
+     * overwritten.
+     *
+     * @param UploadedFile[] $uploadedFiles An array of files uploaded to the
+     *                              server. These files must implement the PSR-7
+     *                              UploadedFile interface.
+     * @param int            $permissions   A UNIX permissions octet (e.g. 0644)
```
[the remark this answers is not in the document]
johnsaigle: I think this and the below variables make more sense as parameters because you might want to upload multiple files having different permissions, I think? It makes a bit more sense to me to say "Object, create a file with these permissions" rather than use an instance variable to put the object into "0644" mode. 

What do you think?

johnsaigle:
Thanks for the review @xlecours. I addressed some of your points above. I'll also make the rest of the changes you mentioned.

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -112,62 +116,81 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
-             * Don't update the file if it already exists on the back-end.
-             * Instead, inform users that an administrator must install it on
-             * their behalf.
-             * This should only happen for users on a system where automatic
-             * installation is disabled (ie. has no quatUser), as the above
-             * error will return before this one.
-            */
+        // Attempt file upload if no errors encountered so far.
+        $success = false;
+        if (empty($err)) {
```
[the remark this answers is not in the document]
xlecours: ok

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -16,6 +16,8 @@ use \Psr\Http\Message\ServerRequestInterface;
 use \Psr\Http\Message\ResponseInterface;
 use \LORIS\Http\EmptyStream;
 
+require_once __DIR__ . '/../../../php/libraries/FileUploader.class.inc';
```
xlecours: This should be handled by composer-autoloading

ridz1208:
@xlecours could you test this one and ideally implement a unit tests for it ?

@johnsaigle you can participate in the creation testing of the unit tests.

johnsaigle:
@ridz1208 Would you prefer that unit tests get added in this PR or a separate one?

johnsaigle:
@xlecours @PapillonMcGill Could you add Passed Manual Tests if this PR worked for you?

PapillonMcGill:
@johnsaigle didn't tested it, just reviewed the code. 
Now as conflict!

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -141,25 +139,27 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
         }
-        $uploaded_file = $request->getUploadedFiles()['install_file'];
-        $filename      = $uploaded_file->getClientFilename();
-        $instrument    = pathinfo($filename)['filename'];
 
-        if ($this->instrumentExists($instrument)) {
-            return (new \Loris\Http\Response())
-                ->withStatus(409, 'Conflict')
+        // Do pre-processing before uploading files.
+        if (!$this->canWriteFiles()) {
```
driusan: This is still not reachable because line 133 already returns with the exact same if condition.

php/libraries/FilesUploadHandler.class.inc
```diff
@@ -0,0 +1,230 @@
...
+
+            // Validate that the MIME type of the file is allowed
+            if (!empty($this->_permittedMIMETypes)) {
+                if (!in_array(
+                    $file->getClientMediaType(),
+                    $this->_permittedMIMETypes,
+                    true
+                )) {
+                    return (new \LORIS\Http\Response())
+                        ->withStatus(400)
```
driusan: I think this should be a 403 forbidden

php/libraries/FilesUploadHandler.class.inc
```diff
@@ -0,0 +1,230 @@
...
+                                array('error' => 'Could not upload file')
+                            )
+                        )
+                    );
+            }
+            chmod($targetPath, $this->_permissions);
+        }
+
+        return (new \LORIS\Http\Response())
+                    ->withStatus(200);
```
driusan: needs a body, and if it's an empty body it should be a 204, not a 200.

php/libraries/FilesUploadHandler.class.inc
```diff
@@ -0,0 +1,230 @@
...
+     * @var int
+     */
+    private $_permissions = 0644;
+
+    /**
+     * Whether existing files should be overwritten.
+     *
+     * @var bool
+     */
+    private $_overwrite = false;
```
driusan: the above should all be protected, not private.

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -168,24 +155,38 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
+
+        if ($response->getStatusCode() !== 200) {
+            return $response;
+        }
 
         // Scripts in tools/ often make relative imports, so we must change
         // our effective directory in order to use them.
         chdir($this->path . "/tools");
         // Use tools/ script to generate an SQL patch file based on the
+        //
```
PapillonMcGill: This line is not needed

php/libraries/FilesUploadHandler.class.inc
```diff
@@ -0,0 +1,231 @@
...
+     * Given an HTTP request, upload all the files to the server in the
+     * same uploadDirectory.
+     *
+     * All files uploaded will get the same permissions.
+     *
+     * If the $overwrite parameter is set to true, existing files will be
+     * overwritten.
+     *
+     * If the files need different permissions on need to go in different
+     * directories, the file must be split in different request an passed to a
```
PapillonMcGill:
```suggestion
     * directories, the files must be split in different request an passed to a
```

johnsaigle:
Thanks @PapillonMcGill. Could you approve again? Committing your suggestion removed the review.

modules/instrument_manager/php/instrument_manager.class.inc
driusan: @johnsaigle please stop marking other people's comments as "Resolved" when they haven't been resolved.

zaliqarosli:
@johnsaigle i'm not sure if this is working as intended. when i "upload anything that is not a LINST file", i get either this
Screenshot 2019-05-06 15 55 42
 or this
Screenshot 2019-05-06 16 02 00

zaliqarosli:
i can also upload .tpl files and successfully installed files get added to my /project/instruments directory but doesn't appear in the browse tab table of the instrument manager module

johnsaigle:
@zaliqarosli 

That error is expected for the invalid file types. Seeing it with fresh eyes though I'll change the error message to "File type not allowed" because that's a more useful message.

Addressing the others points:
1. I think this is OK. `linst` isn't a real file format in the rest of the world so browsers interpret it as `application/octet-stream`. This is what I told the instrument manager to allow. I think `tpl` falls in the same category so this is OK.

2. Hm that's weird. Can you post the LINST file you're using? I'll debug.

3. Also shouldn't happen. I think this is happening because the script just checks if the instrument table exists,  not the file in `project/*`. I'll add a check. We used to have this but I think in all the commits and reordering this must have been erased.

Thanks for testing!

johnsaigle:
The latest commit should fix the above errors.

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -168,25 +155,40 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
         // structure of the uploaded .linst file.
+        error_log("Full path is `$fullpath`");
```
zaliqarosli: should we be leaving error logs?

modules/instrument_manager/php/instrument_manager.class.inc
zaliqarosli: this line is still not needed

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -168,25 +155,40 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
              * error will return before this one.
             */
-            return (new \Loris\Http\Response())
+            return (new \LORIS\Http\Response())
                 ->withStatus(409, 'Conflict')
                 ->withBody(
                     new \LORIS\Http\StringStream(
                         json_encode(
-                            array('error' => self::FILE_ALREADY_EXISTS)
+                            array('error' => self::ALREADY_INSTALLED)
```
zaliqarosli: i don't think you should remove the `FILE_ALREADY_EXISTS` case here. it's useful to know when the instrument has been uploaded but not installed

php/exceptions/UploadException.class.inc
```diff
@@ -0,0 +1,80 @@
+<?php declare(strict_types=1);
+/**
+ * This class is a wrapper that takes a upload error code
+ * an adds the approcpriate message to the exception.
```
zaliqarosli: appropriate

php/exceptions/UploadException.class.inc
```diff
@@ -0,0 +1,80 @@
+<?php declare(strict_types=1);
+/**
+ * This class is a wrapper that takes a upload error code
```
zaliqarosli: an upload

modules/instrument_manager/php/instrument_manager.class.inc
johnsaigle: This one is left over from testing. I'll delete it.

modules/instrument_manager/php/instrument_manager.class.inc
johnsaigle: I changed the code so that the FILE ALREADY EXISTS case now gets handled further down in the file. ALREADY INSTALLED is more appropriate here

johnsaigle:
@zaliqarosli Can you retest? I've rebased it onto aces/major. If it doesn't work can you let me know what issues you encounter?

modules/instrument_manager/php/instrument_manager.class.inc
zaliqarosli: then i think the comments above here need to be moved down to apply to the case for FILE_ALREADY_EXISTS, not for ALREADY_INSTALLED

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -162,19 +155,41 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
+        $uploader = (new \LORIS\FilesUploadHandler($targetdir))
+            ->withPermittedMIMETypes(
+                'application/octet-stream'
+            );
+
+        $response = $uploader->handle($request);
+
+        if (!in_array($response->getStatusCode(), array(200, 201), true)) {
+            if ($response->getStatusCode() == 409) {
+                return new \Loris\Http\Conflict(self::FILE_ALREADY_EXISTS);
```
zaliqarosli: i think this should be \LORIS\Http\Response\Conflict?

modules/instrument_manager/php/instrument_manager.class.inc
zaliqarosli: still needs to be deleted

php/exceptions/UploadException.class.inc
zaliqarosli: this change was reversed by the rebase i think. needs to be reapplied

php/exceptions/UploadException.class.inc
zaliqarosli: same here

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -190,8 +205,6 @@ class Instrument_Manager extends \NDB_Menu_Filter
                         json_encode(
                             array('message' => self::UPLOAD_NO_INSTALL)
                         )
-                    )
```
zaliqarosli: don't think these extra round brackets need to be removed

zaliqarosli: on testing, error_log `PHP Parse error:  syntax error, unexpected '}', expecting ',' or ')' in /var/www/loris/modules/instrument_manager/php/instrument_manager.class.inc on line 208`

johnsaigle: Hm okay I think this was a rebase error. i'll fix this.

johnsaigle:
@zaliqarosli I incorporated your feedback, fixed the syntax error, and things are working for me when I do local testing. Do you mind testing this again?

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -130,45 +128,53 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
+        $targetdir     = new \SplFileInfo($this->path . 'project/instruments/');
+        $fullpath      = $targetdir->getPathname() . '/' . $filename;
 
         if ($this->instrumentExists($instrument)) {
-            return new \LORIS\Http\Response\Conflict(
-                self::ALREADY_INSTALLED
-            );
+            // An instrument with this name already exists in the test_names
+            // table. Abort upload.
+            return new \Loris\Http\Conflict(self::ALREADY_INSTALLED);
```
zaliqarosli:
```suggestion
            return new \LORIS\Http\Response\Conflict(self::ALREADY_INSTALLED);
```

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -130,45 +128,53 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
+
+        if (!in_array($response->getStatusCode(), array(200, 201), true)) {
+            if ($response->getStatusCode() == 409) {
+                /* Don't update the file if it already exists on the back-end.
+                 * Instead, inform users that an administrator must install it
+                 * on their behalf.
+                 * This should only happen for users on a system where automatic
+                 * installation is disabled (ie. has no quatUser).
+                 */
+                return new \Loris\Http\Response\Conflict(self::FILE_ALREADY_EXISTS);
```
zaliqarosli:
```suggestion
                return new \LORIS\Http\Response\Conflict(self::FILE_ALREADY_EXISTS);
```

php/exceptions/UploadException.class.inc
zaliqarosli: easy typo fix here

php/exceptions/UploadException.class.inc
zaliqarosli: easy typo fix here

zaliqarosli:
passes testing but needs a couple more code changes + more to fix travis

johnsaigle:
I made all the suggestions you recommended in terms of typos etc. Thanks for testing again!

My suspicion is that this PR is ready to go. Travis is currently failing on all branches but once that's fixed this will probably pass.

modules/instrument_manager/php/instrument_manager.class.inc
zaliqarosli: everything looks good from my end but i wanted to follow up on other comments. @driusan i don't see/understand this problem. can you confirm this is in fact resolved?

johnsaigle:
@zaliqarosli Travis is passing and I made the typo changes. Are you comfortable approving this?

zaliqarosli:
@johnsaigle no im not, we'll need another quick round of testing i think given the rebase

modules/instrument_manager/php/instrument_manager.class.inc
zaliqarosli: @johnsaigle do you know anything about this?

driusan: looks like it was resolved. The line used to be unreachable, because there was another if above it checking the same thing that the function call was checking.

zaliqarosli:
on testing:

when i upload a PDF file, i get this error in the console log `Failed to load resource: the server responded with a status of 403 (Forbidden)` at `/instrument_manager/?format=json:1 `, the response lags for a couple second and then i get the swal error message.

when i upload a JPEG, swal error message pops up as expected, but i get `POST http://zrosli-dev.loris.ca/instrument_manager/?format=json 403 (Forbidden)` in console log at `uploadForm.js:27 ` which is error that can just be caught but not printed to the console

johnsaigle:
@zaliqarosli I removed two places in the JS code where there were console-logging statements so that should fix the first point. The reason for the lag is probably due to the filesize. I don't get any issue with small PDFs but larger files (such as a WAV file for a song) take a couple seconds to process before the error message occurs.

As to your second point I think that's just how Chrome displays errors? Like if you go to `google.ca/error` it'll give you the same thing but a `404`. Same with any non-existing page in LORIS. Firefox doesn't display error messages like that so I don't think the problem is in the JS code.

zaliqarosli:
@johnsaigle its actually an issue with the fetch call and its response in uploadForm.js . it so happens here that if there is an error, we're expecting and can catch it so it doesn't ever need to print to the console as an error. i can make a PR to your branch with the needed changes when VM goes back up

johnsaigle:
Sounds good, thanks!

zaliqarosli:
@johnsaigle i think you're right, i can't seem to be able to catch the error log although I think I should be able to! I don't want to spend too much time looking into this, so i'll approve the PR as is

johnsaigle:
Thank you x 1000 for testing this so many times

php/libraries/FilesUploadHandler.class.inc
```diff
@@ -0,0 +1,226 @@
...
+        return $new;
+    }
+
+    /**
+     * Given an HTTP request, upload all the files to the server in the
+     * same uploadDirectory.
+     *
+     * All files uploaded will get the same permissions.
+     *
+     * If the $overwrite parameter is set to true, existing files will be
```
driusan:
```suggestion
     * If the overwrite property is set to true, existing files will be
```

The function doesn't have an `$overwrite` parameter

php/libraries/FilesUploadHandler.class.inc
```diff
@@ -0,0 +1,226 @@
...
+     * overwritten.
+     *
+     * If the files need different permissions on need to go in different
```
driusan:
```suggestion
     * If the files need different permissions or need to go in different
```

Typo, I assume.

johnsaigle:
@driusan I applied your suggestions

johnsaigle:
@driusan Thoughts on merging this?

johnsaigle:
@zaliqarosli Do you mind approving again? Dave asked that i make changes to the documentation but the code is the same as when you last approved

php/exceptions/UploadException.class.inc
```diff
@@ -0,0 +1,80 @@
+<?php declare(strict_types=1);
+/**
+ * This class is a wrapper that takes an upload error code
+ * an adds the appropriate message to the exception.
```
xlecours: @cmadjar says: 
> missing d at the end of and

php/exceptions/UploadException.class.inc
```diff
@@ -0,0 +1,80 @@
...
+     * @see http://php.net/manual/en/features.file-upload.errors.php
+     */
+    public function __construct(int $code)
+    {
+        $message = $this->_codeToMessage($code);
+        parent::__construct($message, $code);
+    }
+
+    /**
+     * Mapper function to retreive the corresponding error message.
```
xlecours: @cmadjar says:

> typo: should be retrieve (not retreive)

php/exceptions/UploadException.class.inc
```diff
@@ -0,0 +1,80 @@
...
+     */
+    private function _codeToMessage(int $code) : string
+    {
+        switch ($code) {
+        case UPLOAD_ERR_INI_SIZE:
+            $message = 'The uploaded file exceeds the upload_max_filesize ' .
+                       'directive in php.ini';
+            break;
+        case UPLOAD_ERR_FORM_SIZE:
+            $message = 'The uploaded file exceeds the MAX_FILE_SIZE ' .
```
xlecours: @cmadjar says:

> not sure I completely understand this error. Where does it come from?

php/exceptions/UploadException.class.inc
```diff
@@ -0,0 +1,80 @@
...
+                       'directive that was specified in the HTML form';
+            break;
+        case UPLOAD_ERR_PARTIAL:
+            $message = 'The uploaded file was only partially uploaded';
+            break;
+        case UPLOAD_ERR_NO_FILE:
+            $message = 'No file was uploaded';
+            break;
+        case UPLOAD_ERR_NO_TMP_DIR:
+            $message = 'Missing a temporary folder';
```
xlecours: @cmadjar says:

> maybe replace with "Temporary folder not found"?

php/exceptions/UploadException.class.inc
```diff
@@ -0,0 +1,80 @@
...
+            break;
+        case UPLOAD_ERR_CANT_WRITE:
+            $message = 'Failed to write file to disk';
+            break;
+        case UPLOAD_ERR_EXTENSION:
+            $message = 'File upload stopped by extension';
```
xlecours: @cmadjar says:

> not sure I understand this error. Is it because the extension of the uploaded file is not zip, tar.gz, tar or tgz? If so, maybe rephrase to: "Wrong extension of the file to upload"? Or something like that?

php/libraries/FilesUploadHandler.class.inc
```diff
@@ -0,0 +1,226 @@
+<?php declare(strict_types=1);
+
+/**
+ * Class to upload files in a safe way.
+ *
+ * This class uploads files being uploaded to LORIS. It should
```
xlecours: @cmadjar says:

> I have a feeling you somehow got interrupted when re-writting the first sentence here ;). Maybe it was supposed to be: "This class handles files being uploaded to LORIS."?

php/libraries/FilesUploadHandler.class.inc
```diff
@@ -0,0 +1,226 @@
...
+     * Given an HTTP request, upload all the files to the server in the
+     * same uploadDirectory.
+     *
+     * All files uploaded will get the same permissions.
+     *
+     * If the overwrite property is set to true, existing files will be
+     * overwritten.
+     *
+     * If the files need different permissions or need to go in different
+     * directories, the file must be split in different request an passed to a
```
xlecours: @cmadjar says:

> missing d at the end of and (request and passed to a ...)

php/exceptions/UploadException.class.inc
johnsaigle: I think you wrote this code so I'm not sure.

php/exceptions/UploadException.class.inc
johnsaigle: Same here

php/exceptions/UploadException.class.inc
johnsaigle: Same here.

Also I think we should just go by MIME type. We can check extensions on the front-end if we want to but they don't really matter as long as the actual content is good.

php/exceptions/UploadException.class.inc
xlecours: Those are php constants described at https://www.php.net/manual/en/features.file-upload.errors.php . I will let you write the appropriate message if it needs to be clarified.

php/exceptions/UploadException.class.inc
xlecours: Same here

php/exceptions/UploadException.class.inc
johnsaigle: No that makes sense to me. 

MAX_FILE_SIZE is something you can put in HTML to limit how a file is uploaded. 

It's completely non-standard though and specific to PHP; I don't think anyone uses it.

php/exceptions/UploadException.class.inc
xlecours: Mime type or not, we need to handle this error code. 
I think I copy&pasted that https://www.php.net/manual/en/features.file-upload.errors.php#89374

xlecours: Also, reading the description, this is not related to the file extension but rather to a php extension in php.ini

johnsaigle: OK so then there are no changes to make?

xlecours: If @cmadjar asked, then we should probably clarify. It could be `File upload stopped by a php extension`

php/libraries/FilesUploadHandler.class.inc
```diff
@@ -0,0 +1,226 @@
...
+     * @var bool
+     */
+    protected $overwrite = false;
+
+    /**
+     * Create new instance of a File Uploader.
+     *
+     * @param \SplFileInfo $uploadDirectory The target upload directory
+     */
+    public function __construct(\SplFileInfo $uploadDirectory)
```
xlecours: Just a comment/suggestion
The constructor signature should/could be 
```
public function __construct(\SplFileInfo $uploadDirectory, array $permittedMIMETypes = [], int $permissions = 0644, bool $overwrite = false )
``` 
That way the default values for the class variables could be removed but you would need to set them in the constructor code block. I think it facilitates the instantiation for non-default values.
Feel free to change it or not (It's your PR.)

Issue #4184, opened 2018-12-06
[Imaging Browser] View Session page broken on major for projects not using EDC


cmadjar:
The view session page of the imaging browser returns a 500 error for projects not using EDC.

**Error in the apache log:**
`[Thu Dec 06 10:55:33.764466 2018] [php7:error] [pid 32502] [client 132.216.237.95:61312] PHP Fatal error:  Uncaught TypeError: Return value of Candidate::getCandidateEDC() must be of the type string, null returned in /var/www/loris/php/libraries/Candidate.class.inc:493\nStack trace:\n#0 /var/www/loris/modules/imaging_browser/php/viewsession.class.inc(706): Candidate->getCandidateEDC()\n#1 /var/www/loris/modules/imaging_browser/php/viewsession.class.inc(661): LORIS\\imaging_browser\\ViewSession->getSubjectData()\n#2 /var/www/loris/modules/imaging_browser/php/viewsession.class.inc(98): LORIS\\imaging_browser\\ViewSession->getHeaderTable()\n#3 /var/www/loris/php/libraries/NDB_Form.class.inc(188): LORIS\\imaging_browser\\ViewSession->setup()\n#4 /var/www/loris/src/Middleware/UserPageDecorationMiddleware.php(171): NDB_Form->handle(Object(Zend\\Diactoros\\ServerRequest))\n#5 /var/www/loris/src/Middleware/PageDecorationMiddleware.php(49): LORIS\\Middleware\\UserPageDecorationMiddleware->process(Object(Zend\\Diactoros\\ServerRequest), Object(LORIS\\imaging_browser\\ViewSession))\n#6 /var/www/loris/php/libraries/NDB_Page.class.inc(653): LORIS\\M in /var/www/loris/php/libraries/Candidate.class.inc on line 493, referer: https://cecile-dev.loris.ca/imaging_browser/`

johnsaigle:
@cmadjar I think the better solution here is to change the function signature of `getCandidateEDC()` to be nullable, e.g. return `?string` instead of `string`.

This bug was introduced in #4047. In adding return types to all of the function signatures I allowed `null` whenever my VM or Travis would fail. This one happened to sneak through. If this setting is likely to be null in some cases the function should allow it.

PR #4185, opened 2018-12-06
[Imaging Browser] Fix a bug in imaging browser view session page for projects without EDC

cmadjar:
### Brief summary of changes

The view session page is currently broken on major for projects not using EDC. This PR fixes the bug so that projects not using EDC can still load the view session page of the imaging browser module.

### This resolves issue...

- [x] Github? #4184 

### To test this change...

- [ ] make sure that your datasets does not include EDC (Config set to no and EDC=NULL in the candidate table for the candidate you are testing). Test the major branch and notice you get a 500 error
- [ ] pull the code from this branch and notice you can load the view session page.
- [ ] also test on a candidate with an EDC to make sure it works as well

### DIFF
--- a/modules/imaging_browser/php/viewsession.class.inc
+++ b/modules/imaging_browser/php/viewsession.class.inc
@@ -703,9 +703,7 @@ class ViewSession extends \NDB_Form
         $candidate            =& \Candidate::singleton($timePoint->getCandID());
         $subjectData['pscid'] = $candidate->getPSCID();
         $subjectData['dob']   = $candidate->getCandidateDoB();
-        if ($tpl_data['useEDC'] === 'true') {
-            $subjectData['edc'] = $candidate->getCandidateEDC();
-        }
+        $subjectData['edc']   = $candidate->getCandidateEDC();
         $subjectData['gender'] = $candidate->getCandidateGender();
 
         // This doesn't work.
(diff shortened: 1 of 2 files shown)


modules/imaging_browser/php/viewsession.class.inc
```diff
@@ -703,7 +703,9 @@ class ViewSession extends \NDB_Form
         $candidate            =& \Candidate::singleton($timePoint->getCandID());
         $subjectData['pscid'] = $candidate->getPSCID();
         $subjectData['dob']   = $candidate->getCandidateDoB();
-        $subjectData['edc']   = $candidate->getCandidateEDC();
+        if ($tpl_data['useEDC'] === 'true') {
```
kongtiaowang: $config = \NDB_Config::singleton();
        if ($config->getSetting('useEDC') === 'true') {
            $subjectData['edc'] = $candidate->getCandidateEDC();
        }

cmadjar:
@johnsaigle I integrated the comment you made on issue #4184. Thank you for the suggestion!!

modules/imaging_browser/php/viewsession.class.inc
cmadjar: @kongtiaowang Thank you for reviewing!!

I reverted the changes to `viewsesison.class.inc` back to what it was and made the changes in the `Candidate.class.inc` instead as suggested by @johnsaigle in issue #4184 .

Issue #4187, opened 2018-12-07
Add PSR18 implementation to LORIS


johnsaigle:
There is a need for a [PSR18 Http Client](https://www.php-fig.org/psr/psr-18/) implementation in LORIS. This will allow us to avoid rolling ad-hoc connections to outgoing servers. 

This will be useful in areas such as developing our API, creating "CBRAIN hooks", and leveraging external services such as "Have I Been Pwned?" when validating passwords.

### Related PRs
https://github.com/aces/Loris/pull/3962
https://github.com/aces/Loris/pull/3961
https://github.com/aces/Loris/pull/3765

driusan:
should have been autoclosed by #4188..

PR #4188, opened 2018-12-07
[Core] Add Guzzle dependency

johnsaigle:
### Brief summary of changes

[Guzzle](https://github.com/php-http/guzzle6-adapter) is an implementation of a PSR18 Http Client. It is used to make requests to external servers. As discussed in a recent LORIS meeting, this is required for many different functions we would like to include in LORIS. See ticket for examples.

Guzzle is chosen as it is the [most-downloaded](https://packagist.org/providers/psr/http-client-implementation) PSR18 implementation and likely to be well-maintained in the foreseeable future.

Merging this into major will allow for different features to be developed simultaneously. A good example usage will be a forthcoming update to the "Have I Been Pwned" API call in https://github.com/aces/Loris/pull/3961.

### This resolves issue...

https://github.com/aces/Loris/issues/4187

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
         "smarty/smarty" : "~3.1",
         "firebase/php-jwt" : "~3.0",
         "google/recaptcha": "~1.1",
-        "php-http/guzzle6-adapter": "*",
+        "php-http/guzzle6-adapter": ">=2.0.0",
         "psr/http-message": "~1.0",
         "psr/http-server-handler" : "*",
         "psr/http-server-middleware" : "*",
--- a/composer.json
+++ b/composer.json
@@ -6,6 +6,7 @@
         "smarty/smarty" : "~3.1",
         "firebase/php-jwt" : "~3.0",
         "google/recaptcha": "~1.1",
+        "php-http/guzzle6-adapter": "*",
         "psr/http-message": "~1.0",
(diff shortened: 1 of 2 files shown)


driusan:
You seem to have updated other dependencies in composer.lock (which presumably is the reason it's failing Travis.)

johnsaigle:
@driusan Thanks. I'll take a look.

johnsaigle:
@driusan I'm not really sure what to do here. All I did to create this PR is checkout `major`, add the dependency to composer.json, `composer install` and `composer update` and then committed the two files.

Is it possible the other dependencies are updated because they're required by Guzzle?

Also they don't seem to be related to Travis as far as I can tell.

driusan:
(some of them are probably added because guzzle depends on them, but that seems unlikely for things like smarty or recaptcha)

composer.json
```diff
@@ -6,6 +6,7 @@
         "smarty/smarty" : "~3.1",
         "firebase/php-jwt" : "~3.0",
         "google/recaptcha": "~1.1",
+        "php-http/guzzle6-adapter": "*",
```
PapillonMcGill: Could you put something like ">=2.0"  (or at least >= 1.0) to require at least version 2 instead of any version?
I would not like unstable version as permitted by *

Issue #4017, opened 2018-10-19
[Data Release, Instrument Manager] Module should not rely on superuser permission


johnsaigle:
These modules rely on superuser which acts as a short circuit to all other permissions. That means if a user wants to use these modules at all then they effectively have global access to all other parts of LORIS. These are administrative functions so it's likely that anyone using them will already have elevated permissions -- but there's still no good reason that these modules shouldn't have specific permissions.

### Pull Requests
- [x] Data Release #3062 
- [x] Instrument Manager #4191

ridz1208:
Data_release: https://github.com/aces/Loris/pull/3062

johnsaigle:
Done!

PR #4191, opened 2018-12-07
[Instrument Manager] Add specific permission to this module so that superuser isn't necessary

johnsaigle:
### Brief summary of changes

Add a permission particular to the Instrument Manager module. 

### This resolves issue...

- [x] Github? #4017 

### To test this change...

- [ ] Make sure the new permission shows up in the configuration module and that you can only see the module when you have it.

### DIFF
--- a/SQL/0000-00-01-Permission.sql
+++ b/SQL/0000-00-01-Permission.sql
@@ -110,6 +110,7 @@ INSERT INTO `permissions` VALUES
     (48,'imaging_browser_phantom_ownsite', 'Can access only phantom data from own site in Imaging Browser', 2),
     (49,'data_release_upload', 'Data Release: Upload file', 2),
     (50,'data_release_edit_file_access', 'Data Release: Grant other users view-file permissions', 2);
+    (51,'instrument_manager', 'Instrument Manager: View module and upload instruments', 2);
 
 
 INSERT INTO `user_perm_rel` (userID, permID)
--- a/SQL/0000-00-01-Permission.sql
+++ b/SQL/0000-00-01-Permission.sql
@@ -109,8 +109,9 @@ INSERT INTO `permissions` VALUES
     (47,'imaging_browser_phantom_allsites', 'Can access only phantom data from all sites in Imaging Browser', 2),
     (48,'imaging_browser_phantom_ownsite', 'Can access only phantom data from own site in Imaging Browser', 2),
     (49,'data_release_upload', 'Data Release: Upload file', 2),
-    (50,'data_release_edit_file_access', 'Data Release: Grant other users view-file permissions', 2);
-    (51,'instrument_manager', 'Instrument Manager: View module and upload instruments', 2);
+    (50,'data_release_edit_file_access', 'Data Release: Grant other users view-file permissions', 2),
+    (51,'instrument_manager_read', 'Instrument Manager: View module', 2),
+    (52,'instrument_manager_write', 'Instrument Manager: Install new instruments via file upload', 2);
 
 
 INSERT INTO `user_perm_rel` (userID, permID)
--- a/modules/instrument_manager/README.md
+++ b/modules/instrument_manager/README.md
@@ -32,8 +32,13 @@ The test battery is also managed separately directly from the database.
 
 ## Permissions
 
-The `instrument_manager` module requires the user to have the 
-`instrument_manager` permission. 
+Viewing the `instrument_manager` module requires the user to have one of the 
+following permissions:
+* `instrument_manager_read`
+* `instrument_manager_write`
+
+The write permission additionally allows a user to upload new LORIS 
+instrument. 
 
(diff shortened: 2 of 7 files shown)


SQL/0000-00-01-Permission.sql
```diff
@@ -107,7 +107,8 @@ INSERT INTO `permissions` VALUES
     (45,'issue_tracker_reporter', 'Can add a new issue, edit own issue, comment on all', 2),
     (46,'issue_tracker_developer', 'Can re-assign issues, mark issues as closed, comment on all, edit issues.', 2),
     (47,'imaging_browser_phantom_allsites', 'Can access only phantom data from all sites in Imaging Browser', 2),
-    (48,'imaging_browser_phantom_ownsite', 'Can access only phantom data from own site in Imaging Browser', 2);
+    (48,'imaging_browser_phantom_ownsite', 'Can access only phantom data from own site in Imaging Browser', 2),
+    (49,'instrument_manager', 'Instrument Manager: View module and upload instruments', 2);
```
driusan: I think there should be 2 permissions here.. one to see the list of instruments that are installed, and one to upload a new instrument.

johnsaigle: Done!

kongtiaowang:
A user can't find the menu in the drop-down menu with the permissions,

johnsaigle:
@kongtiaowang I'm not able to reproduce this.

driusan:
What's the discussion that's required on this?

johnsaigle:
@driusan Waiting on @kongtiaowang's feedback cause I can't reproduce this.

driusan:
I think to address his problem you just need to add the permissions for the menu items into LorisMenuPermissions in 0000-00-02-Menus.sql (and include a patch for existing projects)

driusan:
@kongtiaowang does the latest patch resolve your problem?

modules/instrument_manager/README.md
```diff
@@ -32,8 +32,13 @@ The test battery is also managed separately directly from the database.
...
 ## Permissions
 
-The `instrument_manager` module requires the LORIS `superuser`
-permission to be accessed.
+Viewing the `instrument_manager` module requires the user to have one of the 
+following permissions:
+* `instrument_manager_read`
+* `instrument_manager_write`
+
+The write permission additionally allows a user to upload new LORIS
```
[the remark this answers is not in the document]
johnsaigle: instrument is on another line due to `fmt`, I believe.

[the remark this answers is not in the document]
PapillonMcGill: I second Dave about removing "additionally"

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -112,7 +117,9 @@ class Instrument_Manager extends \NDB_Menu_Filter
      */
     protected function handlePOST(ServerRequestInterface $request): ResponseInterface
     {
-        if (!$this->canWriteFiles()) {
+        if (! ($this->canWriteFiles()
+            && \User::singleton()->hasPermission('instrument_manager_write'))
```
xlecours: I would put this check in a separated `if` statement because the error message won't be the same in the Forbidden response.
Also, split line expression in if statement always look weird. (But that is subjective)

modules/instrument_manager/README.md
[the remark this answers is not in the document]
driusan: I meant after removing the word "additionally" there's space on the line to move "instrument." up to the same line, not that it's wrong for it being on the same line now.. but there's no way to do a suggest changes with that kind of  multiline suggestion on GitHub as far as I know, so it still needs to be done in a manual commit.

modules/instrument_manager/php/instrument_manager.class.inc
driusan: I agree with @xlecours.. it should be a different response status based on why, so it needs to be 2 different if statements to do that.

ridz1208:
@driusan ready to go !

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -112,13 +117,26 @@ class Instrument_Manager extends \NDB_Menu_Filter
...
                         json_encode(
-                            array('error' => CANNOT_WRITE_FILES)
+                            array('error' => 'Forbidden')
+                        )
+                    )
+                );
+        }
+        // Ensure the server is properly configured.
+        if (! $this->canWriteFiles()) {
+            return (new \Zend\Diactoros\Response())
```
PapillonMcGill: I think this should be an http\response not Zend.
Would be good to change it in line 122 at the same time

driusan:
```suggestion
            return (new \Loris\Http\Response())
```

Melanie's comment as a suggestion

PR #4192, opened 2018-12-07
[MRI Violations] Add mri_protocol_violations_columnFormatter.js to gitignore

johnsaigle:
### Brief summary of changes

This compiled file should be ignored and it causes issues when changing branches after compiling.


driusan:
The column formatters are supposed to be moved into the React components themselves, but in the meantime until someone properly Reactifies mri violations this is fine..

PR #4193, opened 2018-12-11
[Core: Module] Fix errors in 500-reponse logic

johnsaigle:
### Brief summary of changes

Instrument List crashes on `bugfix` because the `$user` variable was not properly set. A TypeError is thrown by `responseStatus500()` as a result.

This isn't clear from the code changes but this should fix errors like:
`Fatal error: Uncaught TypeError: Argument 2 passed to Module::responseStatus500() must be an instance of User, null given, called in /home/travis/build/aces/Loris/php/libraries/Module.class.inc on line 309 and defined in /home/travis/build/aces/Loris/php/libraries/Module.class.inc:334`

### DIFF
--- a/modules/instrument_list/php/instrument_list.class.inc
+++ b/modules/instrument_list/php/instrument_list.class.inc
@@ -82,11 +82,11 @@ class Instrument_List extends \NDB_Menu_Filter
         $user      =& \User::singleton();
         $timePoint = $this->timePoint;
         if ($timePoint === null) {
-            throw new \LorisException("Page requires a timepoint to load");
+            throw new \NotFound("Page requires a timepoint to load");
         }
         $candidate = $this->candidate;
         if ($candidate === null) {
-            throw new \LorisException("Page requires a candidate to load");
+            throw new \NotFound("Page requires a candidate to load");
         }
 
         //These variable are only for checking the permissions.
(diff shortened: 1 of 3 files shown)


johnsaigle:
In fixing the above error, `instrument_list` stopped working because of the way it handles its routing.

A 500 error would appear on the front-end and this exception was in the error log:
`LorisException: Page requires a timepoint to load in /var/www/loris/modules/instrument_list/php/instrument_list.class.inc:85`

as the code throws a `LorisException` when TimePoint or Candidate are missing from the request. However, this Exception ended up being caught by `Module` after #4163 was merged. 

This code was changed to throw a `NotFound` exception instead so that the module routing code goes through the same branches as it did before the changes in #4163.

modules/instrument_list/php/instrument_list.class.inc
```diff
@@ -82,11 +82,11 @@ class Instrument_List extends \NDB_Menu_Filter
         $user      =& \User::singleton();
         $timePoint = $this->timePoint;
         if ($timePoint === null) {
-            throw new \LorisException("Page requires a timepoint to load");
```
johnsaigle: These were being caught by https://github.com/aces/Loris/pull/4163/files#diff-b9919f169806c905a8606fd31442044fR304, preventing `instrument_list` from loading.

driusan:
@kongtiaowang can you manually test this?

xlecours:
I think the `$user` parameter of the `responseStatus500` function could be retrieved from the request instead of being injected.
Also,  `responseStatus500`  should use the `AnonymousPageDecorationMiddleware` or `UserPageDecorationMiddleware` according to the class of user.
Arguably, the resulting response should be HTTP/1.1 400 Bad Request if the request is missing parameters.

driusan:
Merging despite @xlecours's comment since the instrument list page needs the bug fix, and the feedback could be done in another PR..

Issue #4072, opened 2018-11-07
Add type hinting and return types to all functions in library classes


johnsaigle:
We are making an effort to use typing everywhere we can in LORIS. This issue will track the status of which of the classes `php/libraries` have been covered.

### In Progress

* UserPerms #4051 
* Config #4127 
* NDB_BVL_Instrument #4073 
* NDB_BVL_InstrumentStatus and _ControlPanel #4126 
* NDB_Caller #4125 
* CouchDB #4089 
* User #4027 
* NDB_Factory #4123 

### TODO
* Everything else!

### Done -- 21.0 release
* Candidate #4047 
* Site #4045 
* Project, ProjectDefault #4053 
* SinglePointLogin #4048 
* Utility #4046 
* Client #4091 
* Settings #4090 
* Breadcrumbs* 
* Visit*
* Database #3878 
* TimePoint #4049 
* BVL_Feedback_Panel #4124 

### Postponed until LORIS 22
* LorisForm #4122 
* NDB_Page and child classes #4139 
    * NDB_Form
    * NDB_Menu_Filter
    * NDB_Menu_Filter_Form

driusan:
I think this would be easier to see the status of if it were a GitHub project instead of an issue. What's the current status?

johnsaigle:
It was current until you merged a few more this morning. I can make a project though instead

johnsaigle:
Closing in favour of project https://github.com/aces/Loris/projects/11

PR #4194, opened 2018-12-11
[Core: Notify] Add type declarations to function signatures

johnsaigle:
See https://github.com/aces/Loris/issues/4072


kongtiaowang:
It is ready to go!

PR #4195, opened 2018-12-11
[Core: State] Add type declarations to function signatures

johnsaigle:
### Brief summary of changes
 See https://github.com/aces/Loris/issues/4072

### DIFF
--- a/php/libraries/State.class.inc
+++ b/php/libraries/State.class.inc
@@ -50,11 +50,12 @@ class State
     /**
      * Sets the username
      *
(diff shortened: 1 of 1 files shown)


php/libraries/State.class.inc
```diff
@@ -103,7 +104,7 @@ class State
      *
      * @return String URL suitable for adding to a link of for a "Back" link
      */
-    function getLastURL()
+    function getLastURL(): string
```
PapillonMcGill: can also return null, see line 134

PR #4198, opened 2018-12-12
[Candidate Parameters] Add tab to log candidate identifiers in projects external to the LORIS study

ridz1208:
### Brief summary of changes
New tab in candidate parameters that logs the Identifier of a loris candidate from other studies/projects. this does not have any impact on the candidate anywhere else in LORIS. the only interest here is to log that the candidate is affiliated with other studies that the ongoing LORIS one.

### To test this change...

- [ ] make sure the tab loads in candidate parameters
- [ ] make sure you can add identifiers
- [ ] make sure you can delete values after adding them


SQL/0000-00-00-schema.sql
```diff
@@ -2044,3 +2049,20 @@ CREATE TABLE `candidate_consent_history` (
   `EntryStaff` varchar(255) DEFAULT NULL,
   CONSTRAINT `PK_candidate_consent_history` PRIMARY KEY (`CandidateConsentHistoryID`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE TABLE `project_external` (
+  `ProjectExternalID` int(2) NOT NULL AUTO_INCREMENT,
```
PapillonMcGill:
```suggestion
  `ProjectExternalID` int(4) NOT NULL AUTO_INCREMENT,
```
99 possibilities might not be enough for some studies

PapillonMcGill:
Could also use existing Project table adding an attributes external (y/n) or similar to distinguish. Would require modifying more existing code but would prevent a new tables with  almost identical attributes.

ridz1208:
@PapillonMcGill I considered that option, but all it would be doing is adding an attribute to every project row... when an external project and an internal one have almost nothing in common. external projects dont have recruitement targets, dont have subprojects, dont affect anything in loris really while loris projects are very important.

no queries would be querying both external and internal projects. and this way we avoid another "study site"-like confusion situation. keeping them completely separate makes it clearer that they are 2 different entites
I can also change `project` to be `external affiliations`... its completely different I just find the "project" is clearer

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -467,3 +471,65 @@ function editConsentStatusFields($db, $user)
...
+ *
+ * @throws DatabaseException
+ *
+ * @return void
+ */
+function editExternalIdentifierFields($db, $user)
+{
+    if (!$user->hasPermission('candidate_parameter_edit')) {
+        header("HTTP/1.1 403 Forbidden");
+        exit;
```
```suggestion
        return;
```
Dave gets mad if you use hard exits.

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -467,3 +471,65 @@ function editConsentStatusFields($db, $user)
...
+    }
+
+    $candID = $_POST['candID'];
+
+    // Process posted data
+    $study      = isset($_POST['ProjectID']) ?
```
johnsaigle: A little cleaner to use `??` instead.

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -467,3 +471,65 @@ function editConsentStatusFields($db, $user)
...
+ * Handles the deletion of an External Identifier
+ *
+ * @param Database $db   database object
+ * @param User     $user user object
+ *
+ * @throws DatabaseException
+ *
+ * @return void
+ */
+function deleteExternalIdentifier($db, $user)
```
```suggestion
function deleteExternalIdentifier($db, $user): void
```

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -467,3 +471,65 @@ function editConsentStatusFields($db, $user)
...
+{
+    if (!$user->hasPermission('candidate_parameter_edit')) {
+        header("HTTP/1.1 403 Forbidden");
+        exit;
```
```suggestion
        return;
```

modules/candidate_parameters/ajax/getData.php
```diff
@@ -470,3 +472,48 @@ function getConsentStatusHistory($pscid)
...
 }
+
+/**
+ * Handles the fetching of External Identifiers fields
+ *
+ * @throws DatabaseException
+ *
+ * @return array
+ */
+function getExternalIdentifierFields()
```
```suggestion
function getExternalIdentifierFields(): array
```

modules/candidate_parameters/ajax/getData.php
```diff
@@ -470,3 +472,48 @@ function getConsentStatusHistory($pscid)
...
+    $allStudiesUnformatted = $db->pselect(
+        "SELECT ProjectExternalID as ProjectID, Name FROM project_external",
+        array()
+    );
+    $allStudies            =array();
+    foreach ($allStudiesUnformatted as $k=>$row) {
+        $allStudies[$row['ProjectID']] =$row['Name'];
+    }
+
+    $result = [
```
johnsaigle: should be `array()` style

SQL/New_patches/2018-12-12-External_identifiers.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE `project_external` (
+  `ProjectExternalID` int(2) NOT NULL AUTO_INCREMENT,
+  `Name` varchar(255) DEFAULT NULL,
```
PapillonMcGill: should match the 0000-00-00-schema.sql files

SQL/0000-00-00-schema.sql
```diff
@@ -2044,3 +2049,20 @@ CREATE TABLE `candidate_consent_history` (
...
+
+CREATE TABLE `project_external` (
+  `ProjectExternalID` int(4) NOT NULL AUTO_INCREMENT,
+  `Name` varchar(255) UNIQUE DEFAULT NULL,
+  PRIMARY KEY (`ProjectExternalID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE TABLE `candidate_project_external_rel` (
+  `CandID` int(6) NOT NULL DEFAULT 0,
+  `ProjectExternalID` int(2) NOT NULL,
```
PapillonMcGill:
```suggestion
  `ProjectExternalID` int(4) NOT NULL,
```

SQL/New_patches/2018-12-12-External_identifiers.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE `project_external` (
+  `ProjectExternalID` int(2) NOT NULL AUTO_INCREMENT,
```
PapillonMcGill:
```suggestion
  `ProjectExternalID` int(4) NOT NULL AUTO_INCREMENT,
```

SQL/New_patches/2018-12-12-External_identifiers.sql
```diff
@@ -0,0 +1,16 @@
...
+  `Name` varchar(255) DEFAULT NULL,
+  PRIMARY KEY (`ProjectExternalID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE TABLE `candidate_project_external_rel` (
+  `CandID` int(6) NOT NULL DEFAULT 0,
+  `ProjectExternalID` int(2) NOT NULL,
```
PapillonMcGill:
```suggestion
  `ProjectExternalID` int(4) NOT NULL,
```

SQL/0000-00-00-schema.sql
```diff
@@ -2044,3 +2049,20 @@ CREATE TABLE `candidate_consent_history` (
...
   CONSTRAINT `PK_candidate_consent_history` PRIMARY KEY (`CandidateConsentHistoryID`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE TABLE `project_external` (
+  `ProjectExternalID` int(4) NOT NULL AUTO_INCREMENT,
+  `Name` varchar(255) UNIQUE DEFAULT NULL,
+  PRIMARY KEY (`ProjectExternalID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE TABLE `candidate_project_external_rel` (
```
driusan: I find the underscore in the table name makes this look a little like a trinary relationship table, but I guess there's nothing to do for that given our existing naming standards.

SQL/New_patches/2018-12-12-External_identifiers.sql
```diff
@@ -0,0 +1,16 @@
...
+CREATE TABLE `candidate_project_external_rel` (
+  `CandID` int(6) NOT NULL DEFAULT 0,
+  `ProjectExternalID` int(4) NOT NULL,
+  `ExtStudyID` varchar(255) NOT NULL DEFAULT '',
+  `DateUpdated` timestamp NOT NULL DEFAULT current_timestamp() ON UPDATE current_timestamp(),
+  PRIMARY KEY (`CandID`,`ProjectExternalID`),
+  KEY `FK_candidate_project_extid_rel_2` (`ProjectExternalID`),
+  CONSTRAINT `FK_candidate_project_external_rel_1` FOREIGN KEY (`CandID`) REFERENCES `candidate` (`CandID`),
+  CONSTRAINT `FK_candidate_project_external_rel_2` FOREIGN KEY (`ProjectExternalID`) REFERENCES `project_external` (`ProjectExternalID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
driusan: These should be utf8mb4, but to do that you'll probably need to shrink the size of the ExtStudyID column slightly.

modules/candidate_parameters/jsx/ExternalIdentifier.js
```diff
@@ -0,0 +1,350 @@
...
+            </strong>
+          </div>
+        );
+      }
+
+      return (
+        <button className="btn-info has-spinner">
+          Loading
+          <span
+            className="glyphicon glyphicon-refresh
```
driusan: GitHub highlights this as an error (even though it's not..) can you move the 2 glass names onto the same line so it doesn't get highlighted?

modules/candidate_parameters/jsx/ExternalIdentifier.js
```diff
@@ -0,0 +1,350 @@
...
+    formData.append('tab', this.props.tabName);
+    formData.append('candID', this.state.Data.candID);
+
+    existingStudies.push(existingStudy);
+
+    this.setState({
+      existingStudies: existingStudies
+    });
+
+    $.ajax({
```
driusan: are we moving towards using fetch instead and getting rid of the jQuery dependency? @HenriRabalais @davidblader @zaliqarosli @xlecours ?

I know that's what's happening on the menu filter PRs, and it's probably a good idea to make it universal?

davidblader: agree with using fetch and with removing jQuery dependency in general, however the fade animation in `showAlertMessage` might be tricky/somewhat time consuming to refactor w/o jQuery and i wouldn't necessarily block on that basis at this moment. using fetch is a good start though

driusan: I'm also undecided on if we should block over the ES6 thing.. it's a useful feature to block over style which might take a lot of effort to fix, but @zaliqarosli's also put a lot of effort into modernizing all of our existing javascript into ES6 and since it's a new class it would be good not to have regressions in that regard.

johnsaigle: IMO Reactification outranks ES standards since it's blocking so many other things such as using the data provisioners, project permissions, etc.

We can always make a ticket to update this code to ES6 standards. It's a good beginner/summer student task.

driusan: This isn't Reactification though, this is adding a new tab /feature to a module that's already written in React.

zaliqarosli: showAlertMessage can be replaces with swal which is very easy to implement, and looks very clean in ES6

zaliqarosli: also, any future reactifications should be coded in ES6 anyways

ridz1208: I'm not gonna go down that road. thx for the feedback but I'm most probably not gonna have time to refactor anything on the JS side. this PR will most probably be closed

if anyone ever decides to tackle the Candidate params module. keep this PR in mind

ridz1208:
@PapillonMcGill if you have got everything you need from this PR let me know. It seems that it is behind on the JS standards and will not be making it into loris so I will go ahead and close it.

PapillonMcGill:
@ridz1208 I took the diifs, closing this PR.
Thanks

PR #4203, opened 2018-12-14
[jsx] FilterableDataTable UI Redesign

zaliqarosli:
### Brief summary of changes

PR as a result of discussions in UX WG to address problem of adding 'Call To Action' buttons within the menu filter as found in https://github.com/aces/Loris/pull/4167

- Creates a new FieldsetElement component
- Changes Filter component to use FieldsetElement instead of FormElement. Any implementation of FormElement component should not be changed (i.e. Media Upload Form)
- Changes look of DataTable and insert call to action buttons within the table 'header'
- added title prop to Dicom Archive to render title in card panel header

screenshot 2018-12-14 13 24 38
screenshot 2018-12-14 13 43 20

### To test this change...

- [ ] Test UI and functionality of any module that uses FilterableDataTable e.g. Media, Dicom Archive and Instrument Manager.

-  UI should look like screenshots above, and functionality should remain the same. Report any bugs.

### DIFF
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -150,8 +150,7 @@ FormElement.defaultProps = {
  * React wrapper for <fieldset> element that is nested inside <FormElement></FormElement>,
  * and accepts child react components. A fieldset groups related elements in a form.
  *
- * The form elements can be passed by:
- * 1. Form components nested directly inside <FieldsetElement></FieldsetElement>
+ * The form elements can be passed by nesting Form components directly inside <FieldsetElement></FieldsetElement>
  *
  */
 class FieldsetElement extends Component {
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -148,7 +148,7 @@ FormElement.defaultProps = {
 /**
  * FieldsetElement Component.
  * React wrapper for <fieldset> element that is nested inside <FormElement></FormElement>,
- * and accepts children react components. A fieldset groups related elements in a form.
+ * and accepts child react components. A fieldset groups related elements in a form.
  *
  * The form elements can be passed by:
  * 1. Form components nested directly inside <FieldsetElement></FieldsetElement>
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -150,7 +150,7 @@ FormElement.defaultProps = {
  * React wrapper for <fieldset> element that is nested inside <FormElement></FormElement>,
  * and accepts child react components. A fieldset groups related elements in a form.
  *
- * The form elements can be passed by nesting Form components directly inside <FieldsetElement></FieldsetElement>
+ * The form elements can be passed by nesting Form components directly inside <FieldsetElement></FieldsetElement>.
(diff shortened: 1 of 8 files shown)


jsx/Form.js
```diff
@@ -145,6 +145,75 @@ FormElement.defaultProps = {
...
+    });
+    return formElementsHTML;
+  }
+
+  render() {
+    // Generate form elements
+    let formElements = this.getFormElements();
+
+    return (
+      <div style={{padding: '0px', margin: '0px'}}>
```
driusan: I don't see the point of this div. The fieldset is already a block level element.

(I'm not entirely sure that there's much benefit in having this component at all instead of just having the CSS and putting the <fieldset> tag directly into the jsx components where you want them..)

zaliqarosli: oops, you're right. i dont know why i thought that div was needed as a block level element.

fieldsets group related elements in a form, and so i see the benefit of having the ability to render multiple fieldsets in one FormElement

driusan: Right, I agree, I just don't see what benefit `<FieldsetElement>xxx</FieldsetElement>` adds over `<fieldset>xxx</fieldset>` in that regard.

kongtiaowang:
https://github.com/kongtiaowang/Loris/pull/163/files

jsx/Form.js
```diff
@@ -145,6 +145,73 @@ FormElement.defaultProps = {
   },
 };
 
+/**
+ * FieldsetElement Component.
+ * React wrapper for <fieldset> element that is nested inside <FormElement></FormElement>,
+ * and accepts children react components. A fieldset groups related elements in a form.
```
driusan:
```suggestion
 * and accepts child react components. A fieldset groups related elements in a form.
```

jsx/Form.js
```diff
@@ -145,6 +145,73 @@ FormElement.defaultProps = {
...
+ *
+ * The form elements can be passed by:
+ * 1. Form components nested directly inside <FieldsetElement></FieldsetElement>
```
driusan: Can't do a multiline suggest, but.. there's only one thing in this list, why not just make it a sentence?

jsx/Form.js
```diff
@@ -145,6 +145,72 @@ FormElement.defaultProps = {
   },
 };
 
+/**
+ * FieldsetElement Component.
+ * React wrapper for <fieldset> element that is nested inside <FormElement></FormElement>,
+ * and accepts child react components. A fieldset groups related elements in a form.
+ *
+ * The form elements can be passed by nesting Form components directly inside <FieldsetElement></FieldsetElement>
```
zaliqarosli:
```suggestion
  * The form elements can be passed by nesting Form components directly inside <FieldsetElement></FieldsetElement>.
```

driusan:
Merging this, but we should probably separately document somewhere how to add CTAs to a FilterableDataTable.

PR #4207, opened 2018-12-17
[Core] Bug fix for reenabling inactive user accounts trying to log back in

Jkat:
See https://github.com/aces/Loris/issues/4208 for more details.

### DIFF
--- a/php/libraries/SinglePointLogin.class.inc
+++ b/php/libraries/SinglePointLogin.class.inc
@@ -318,7 +318,8 @@ class SinglePointLogin
                 }
 
                 if ($row['Active'] == 'N'
-                    || $this->disabledDueToInactivity($username, 365)
+                    || ($this->disabledDueToInactivity($username, 365)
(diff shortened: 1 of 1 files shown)


php/libraries/SinglePointLogin.class.inc
```diff
@@ -318,7 +318,8 @@ class SinglePointLogin
                 }
 
                 if ($row['Active'] == 'N'
-                    || $this->disabledDueToInactivity($username, 365)
+                    || ($this->disabledDueToInactivity($username, 365)
+                        && $row['Active'] != 'Y')
```
PapillonMcGill: This will not work, even if you change Active to Y in the DB, the user will be shown the message below (line 324-235)

PapillonMcGill: I do believe that something to a larger scale, like adding a parameter in the DB `inactiveOverride` or change the process to disable the user after a period without login in.

PR #4216, opened 2018-12-20
[API] Adding /projects/$projectname

xlecours:
- Adding `/projects/$projectname` endpoint in the api module
- Moving endpoints in PSR-4 autoloading scope
- Adding logic in ETag middleware to not overwrite ETag header

### To test this change...
HTTP/1.1 GET api/v0.0.1/projects/`$projectname` 
HTTP/1.1 GET api/v0.0.2/projects/`$projectname` 
HTTP/1.1 GET api/v0.0.3-dev/projects/`$projectname` 

The response should match the specification found @ https://github.com/aces/Loris/blob/minor/docs/API/LorisRESTAPI_v0.0.3.md#21-single-project

### DIFF
--- a/htdocs/api/v0.0.1/.htaccess
+++ b/htdocs/api/v0.0.1/.htaccess
@@ -7,18 +7,20 @@ Options -Indexes
 RewriteCond %{ENV:REDIRECT_STATUS} 200
 RewriteRule ^ - [L]
 
-RewriteRule ^login Login.php?PrintLogin=true [L]
+# Login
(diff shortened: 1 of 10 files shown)


thomasbeaudry:
Hi Xavier, can you go one step further and add the subproject endpoint please?

xlecours:
> Hi Xavier, can you go one step further and add the subproject endpoint please?

Hi @thomasbeaudry, it is not in the scope of that pull request but I am rewriting most of the endpoints. Could you create a github issue where you could give me the response format and where we could define the need? I should be able to deliver that in v0.0.3-dev right away. Thanks

PR #4218, opened 2018-12-20
[CORE-Candidate] adding check to ensure no PSCID or ExternalID end with zero values

PapillonMcGill:
this replace all digits parts ending with 3 consecutive 0s by 001
or end with [a-z]0 by [a-z]1

### Brief summary of changes

after generating the tentative psicid (externalID), call a function to check if the string contain parts that equate to '0' value and replace them with a '1' value

side effect, there can be no more pscid in the for of 'abc1000' as it would be replaced 'abc1001'.

created this to support complicated pscid as at least one organization as IDs in the form of
xyz0001.1
xyz0001.a1

### This resolves issue...
issue identified by a group where the first candidate was 'xyz0000'

### To test this change...

- send string containing multiple consecutive '0's in it to the function and check that the last one is replaced by '1'

### DIFF
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -968,47 +968,47 @@ class Candidate
         return $option_array;
     }
 
-}
+    /**
+     * Ensure that generated IDs do not end with 0
+     * replace every consecutive digit pattern ending with 000 by 001
+     * or string ending with an alpha char followed by a 0
+     * e.g.: abc000 -> abc001
+     *       abc000r4 -> abc001r4
+     *       abc123d0 -> abc123d1
+     *
+     * @param string $string the string to correct
+     *
+     * @return string the string with offending 0 replaced by 1
+     */
+    function zeroSupress(string $string): string
+    {
+        $matches  = array();
+        $replaces = array();
+        $pattern  = "|0{3,}|";
 
-/**
- * Ensure that generated IDs do not end with 0
- * replace every consecutive digit pattern ending with 000 by 001
- * or string ending with an alpha char followed by a 0
- * e.g.: abc000 -> abc001
- *       abc000r4 -> abc001r4
- *       abc123d0 -> abc123d1
- *
- * @param string $string the string to correct
- *
- * @return string the string with offending 0 replaced by 1
- */
-function zeroSupress(string $string): string
-{
-    $matches  = array();
-    $replaces = array();
-    $pattern  = "|0{3,}|";
+        // find occurence of at least 3 consecutive 0
+        preg_match_all($pattern, $string, $matches);
+        $matches  = $matches[0];
+        $replaces = array();
+        foreach ($matches as $match) {
+            $replaces[] = str_pad("1", strlen($match), "0", STR_PAD_LEFT);
+        }
 
-    // find occurence of at least 3 consecutive 0
-    preg_match_all($pattern, $string, $matches);
-    $matches  = $matches[0];
-    $replaces = array();
-    foreach ($matches as $match) {
-        $replaces[] = str_pad("1", strlen($match), "0", STR_PAD_LEFT);
+        //replace the last 0 by 1
+        $pos = 0;
+        $len = 0;
+
+        for ($i = 0; $i < sizeof($matches); $i++) {
+            $pos    = strpos($string, $matches[$i], $pos+$len);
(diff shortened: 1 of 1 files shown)


driusan:
What's the point of this? Identifiers ending with 0 are perfectly valid identifiers.

PapillonMcGill:
@driusan 
zyx0010 is ok and will pass
xyz0100 will also pass
but some unit don't like 'xyz0000'  and ask to change them
so xyz000 will become xyz001 and xyz1000 will become xyz1001

ridz1208:
@PapillonMcGill @driusan I have also seen this request come up but singularly for patient 0s. ie. we should not prevent any other number combinations than 0000.

given that we dont really decide the exact length and format of PSCIDs we cant rely on the fact that there is a numeric element of 4 digits or less or more. I dont think we can draw the line at a minimum of 3 digits either so the implementation here I dont think is optimal.
the following are perfectly fine PSCIDs and yet they will be affected by the code here:
- AAA0001
- AAA1000
- AAA00000000001
- AAA00002000

The only problematic id (for the studies I have talked to are things like:
- AAA00 (which would not be caught here BTW)
- AAA00000000000
- AAA0000
- ...

I think what we should be doing is adding each digit of a numeric section and only increment by 1 if the sum is 0.  I have not dived in the code but I imagine that should be possible to implement directly in the gererator function.

Adding discussion required to establish if projects indeed agree with this change

PapillonMcGill:
just to clarify, the 3 consecutive 0s have to be the right most  part of the digit section so 

-     AAA0001  NOT affected 
-     AAA1000  modified to AAA1001
-     AAA00000000001 NOT affected
-     AAA00002000 modified to AAA000020001

Thing that might be problematic are like AAA00, this would slip trough with current implementation. Can modify to use @ridz1208 suggestion to some number up to 0. In that case, Would it be a problem if we have a number like  A3BF4H0J modified to A3BF4H1J?

Or change the minimum of 3 consecutive zero (right most) to a minimum of 2 would be better?

AAA00 would become AAA01
AAA0003000 would become AAA0003001
AAA0 would become AAA1 (currently implemented)
AA0A0 would become AA0A1

ridz1208:
@PapillonMcGill the config.xml file allows you to dictate which type of element you want in your PSCID. I think this rule should only be applied to numeric elements as is:

- if `A3BF4H0J` is defined in the config.xml as 6 alphanumerics + 1 numeric + 1 alpha then it **should be converted** to `A3BF4H1J` because the `0` evaluates to a SUM of 0 numeric element.

- if `A3BF4H0J` is defined in the config.xml as 8 alphanumerics then it **should NOT be converted** to `A3BF4H1J` because the `0` is part of a non-numeric sequence

***HOWEVER, by digging deeper into this right now, i see a min/max element in the numeric field. this issue might have been already resolved by having the study set the min=0001 or any other value of their choosing. investigate this before proceeding forward***

PapillonMcGill:
@ridz1208 
*min* and *max* parameter in config.xml (ExternalID) are never used. 
'CANDIDATE_MIN_PSCID' and 'CANDIDATE_MAX_PSCID' defined at the top of the php files are also never used.

Would you prefer that I use the value in the config.xml file or constant in the PHP file to implement a minimum and maximum value for PSCID?

ridz1208:
@PapillonMcGill 

my personnal opinion is
- no restrictions on alphanumeric
- min/max defined in config.xml
- externalID I dont care, I dont even know how thats used

@driusan any preferences here?

ridz1208:
@johnsaigle by that logic, we should start our counts in menu filters at 0 instead of 1... the reality of it is that no project (at least ones lead by non-developers) wants their first candidate to be 0000, normal people index at 1

johnsaigle:
I feel like a lot of this code could use a rewrite... I'm going to work on a pull request to revamp some of this.

driusan:
I think we should just fix the bug preventing min/max from working as advertised.

PapillonMcGill:
@driusan It is my intention to use min and max.
@johnsaigle we should coordinate as I will likely have to rework the initial string function.

johnsaigle:
@PapillonMcGill I was working on a rewrite yesterday so I'll open a PR with my changes and we can jointly work on it there.

PapillonMcGill:
blocking until #4241 is done

PR #4219, opened 2018-12-20
[API] root endpoints in src

xlecours:
This is moving the `/login` and `/projects/` in `src/Api/Endpoints` where all api endpoints belong.
The api module's module class is now doing the initial routing for the parents endpoints (`/login`, `/projects` and shortly `candidates`)


modules/api/php/module.class.inc
```diff
@@ -106,24 +108,43 @@ class Module extends \Module
...
+                    ->withStatus(404)
+                    ->withHeader('Content-Type', 'application/json')
+                    ->withBody(
+                        new \LORIS\Http\StringStream(
+                            json_encode(["error" => "not found"])
+                        )
+                    );
+        }
+
+        $response = $handler->handle($newrequest);
```
driusan:
```suggestion
        // This disregards the parent handler from the base module class, because most of the work
        // it does (adding page decoration middleware, delegating to a sub-page class in the Module, etc)
        // is not relevant for the API module.
        $response = $handler->handle($newrequest);
```

driusan:
Failing due to a couple line lengths in the comment I suggested:

```
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
 130 | WARNING | Line exceeds 85 characters; contains 98 characters
 131 | WARNING | Line exceeds 85 characters; contains 105 characters
```

(Sorry! GitHub's UI doesn't make it easy to see how long a line is when you're suggesting changes..)

PR #4220, opened 2018-12-20
[Api]  /projects/$project sub-endpoints

xlecours:
- [x] `/projects/$projectname/candidates`
- [x] `/projects/$projectname/images`
- [x] `/projects/$projectname/instruments`
- [x] `/projects/$projectname/instruments/$instrumentname`
- [x] `/projects/$projectname/visits`

based on/from https://github.com/aces/Loris/pull/4219

- [x] dataframework 
- [x] Add strict type


driusan:
@xlecours Can you rebase this? It conflicts since #4219 was merged.

src/Api/Endpoints/Projects/Images.php
```diff
@@ -0,0 +1,178 @@
...
+            return (new \LORIS\Http\Response())
+                    ->withStatus(400)
+                    ->withBody(
+                        new \LORIS\Http\StringStream($e->getMessage())
+                    );
+        }
+
+        if (!isset($this->responseCache[$projectname])) {
+
+            $rows = \NDB_Factory::singleton()->database()->pselect(
```
driusan: Maybe this refactoring would be a good time to update this to use a data provisioner?

src/Api/Endpoints/Projects/Instrument.php
```diff
@@ -0,0 +1,130 @@
...
+            try {
+                $instrument = \NDB_BVL_Instrument::factory(
+                    $instrumentname,
+                    null,
+                    null,
+                    true
+                );
+            } catch (\Exception $e) {
+                return (new \LORIS\Http\Response())
+                    ->withStatus(400)
```
driusan: shouldn't this be a 500?

driusan: or perhaps a 404 since the most likely way to get an exception is with a non-existent instrument (in which case the catch type should probably be more explicit.)

src/Api/Endpoints/Projects/Instruments.php
```diff
@@ -0,0 +1,165 @@
...
+            $meta = array('Project' => $projectname);
+
+            $ddeinstruments = \NDB_Factory::singleton()
+                ->config()
+                ->getSetting("DoubleDataEntryInstruments");
+
+            // FIXME: We should be able to instanciate an instrument and call
+            // ->isDDE() and ->getSubgroupName() instead of querying the
+            // persistence layer.
+            $result = \NDB_Factory::singleton()->database()->pselect(
```
driusan: feels like this could use a provisioner too so that there's less work when we start adding project permissions, but this is less immediately applicable than the candidate one above..

src/Api/Endpoints/Projects/Images.php
```diff
@@ -0,0 +1,178 @@
...
+    /**
+     * Create an array representation of this endpoint's reponse body
+     *
+     * @param ServerRequestInterface $request The incoming PSR7 request
+     *
+     * @return array
+     */
+    private function _toArray(ServerRequestInterface $request): array
+    {
+        $project     = $request->getAttribute('project');
```
driusan: why not inject this with the constructor instead? (Same with most of the other projects endpoints). It's more of a class attribute than a request attribute and that way you would have better type checking.

src/Api/Endpoints/Projects/Candidates.php
```diff
@@ -0,0 +1,127 @@
+<?php
```
driusan: Might be a good idea to use strict types in new classes from the start, rather than having to add them later?

src/Api/Endpoints/Projects/Images.php
```diff
@@ -0,0 +1,178 @@
...
+                   ON (s.CandID = c.CandID)
+                 LEFT JOIN psc p
+                   ON (s.CenterID = p.CenterID)
+                 LEFT JOIN files_qcstatus qc
+                   ON (f.FileID = qc.FileID)
+                 LEFT JOIN Project project
+                   ON (c.ProjectID = project.ProjectID)
+                 WHERE
+                   c.Active = "Y" AND
+                   project.Name = :v_projectname AND
```
PapillonMcGill: Should it check for session.Active = 'Y' ?

src/Api/Endpoints/Projects/Images.php
xlecours: While what you ask is/was out of scope for this PR, this first step in moving the /projects sub-tree in the module framework went faster than expected. I think I can spend additional time doing it right in this iteration.

modules/api/php/projectimagesrowprovisioner.class.inc
```diff
@@ -0,0 +1,92 @@
...
+             LEFT JOIN candidate c
+               ON (s.CandID = c.CandID)
+             LEFT JOIN psc p
+               ON (s.CenterID = p.CenterID)
+             LEFT JOIN files_qcstatus qc
+               ON (f.FileID = qc.FileID)
+             LEFT JOIN Project project
+               ON (c.ProjectID = project.ProjectID)
+             WHERE
+               c.Active = "Y" AND
```
driusan: I think @PapillonMcGill's comment about s.Active still applies but was lost because of moving the query around

src/Data/Filters/NoPhantoms.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+namespace LORIS\Data\Filters;
+
+use \Loris\Data\DataInstance;
+use \LORIS\Data\Filter;
+
+/**
+ * OnlyPhantoms filters out data for any images which is associated to a scanner
+ * For a DataInstance to be compatible with the NoPhantoms filter, it must
+ * implement a isPhantom  method which returns an boolean. The DataInstance will
```
PapillonMcGill:
```suggestion
 * implement a isPhantom  method which returns a boolean. The DataInstance will
```

src/Data/Filters/OnlyPhantoms.php
```diff
@@ -0,0 +1,49 @@
...
+ * @license    http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link       https://www.github.com/aces/Loris/
+ */
+namespace LORIS\Data\Filters;
+use \LORIS\Data\Filter;
+use \LORIS\Data\DataInstance;
+/**
+ * OnlyPhantoms filters out data for any images which is not associated to a scanner
+ * For a DataInstance to be compatible with the OnlyPhantoms filter, it must
+ * implement a isPhantom  method which returns an boolean. The DataInstance will
```
PapillonMcGill:
```suggestion
 * implement a isPhantom  method which returns a boolean. The DataInstance will
```

cmadjar:
@xlecours Actually, just realized that ProjectID is not set on my sandbox for candidates (meaning probably not set for PREVENT-AD as well). Are there other projects that may not have this column set?

PR #4231, opened 2019-01-03
[Instrument List] Prevent TypeErrors from crashing instrument list

johnsaigle:
### Brief summary of changes

When bad input is given to instrument list, LORIS will crash and the back-end will throw either a LorisException or TypeError depending on the type of input.

This PR handles these cases as a 500 response and displays an error to the user instead of crashing.

### To test this change...

- [ ] Go to instrument list and change one of the values to a negative number or to a string. LORIS will crash
- [ ] Do the same on my branch and you'll instead see an error message.

### DIFF
--- a/modules/instrument_list/php/module.class.inc
+++ b/modules/instrument_list/php/module.class.inc
@@ -96,7 +96,7 @@ class Module extends \Module
                 new \LORIS\Router\NoopResponder(
                     new \LORIS\Http\Error(
                         $request,
-                        500,
+                        404,
                         $message
                     )
                 )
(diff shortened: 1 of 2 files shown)


modules/instrument_list/php/module.class.inc
```diff
@@ -63,26 +63,46 @@ class Module extends \Module
...
+                . htmlspecialchars($gets['candID']) . ' and session '
+                . htmlspecialchars($gets['sessionID']) . '.';
+            return (new \LORIS\Middleware\PageDecorationMiddleware(
+                $user
+            ))->process(
+                $request,
+                new \LORIS\Router\NoopResponder(
+                    new \LORIS\Http\Error(
+                        $request,
+                        500,
```
driusan: If this exception is caused by not being able to find the candidate/session, that's a 404 (user trying to access a non-existent resource), not a 500 (server error). (But if there are other causes of the exception, then the error message is misleading..)

PR #3897, opened 2018-08-23
[Core] Replace the "Gender" field with "Sex" in the candidate table

driusan:
This replaces the word "Gender" with "Sex" throughout most of
LORIS (except old versions of the API which must remain stable
after being published).

To test: Most of LORIS needs to be tested, since the column
which was renamed is a common one used throughout many places
in LORIS.

PR #4234, opened 2019-01-07
[Core] Add return types to Module class

johnsaigle:
### Brief summary of changes
Add return types to the Module class

### DIFF
--- a/php/libraries/Module.class.inc
+++ b/php/libraries/Module.class.inc
@@ -154,8 +154,8 @@ class Module extends \LORIS\Router\PrefixRouter implements RequestHandlerInterfa
      * Returns true if this module should be publicly available to people who
      * are not logged in to LORIS.
      *
-     * @return bool (false: note that this is not currently possible because main.php
-     *         checks that the user is logged in before it calls NDB_Caller.)
+     * @return bool False. Subclasses must override this function in the case
+     *                  that they should be publicly available.
      */
     public function isPublicModule()
     {
--- a/php/libraries/Module.class.inc
+++ b/php/libraries/Module.class.inc
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
 /**
  * This file contains a class which encapsulates the concept of a "module"
  * in LORIS.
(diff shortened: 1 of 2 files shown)


php/libraries/Module.class.inc
```diff
@@ -150,7 +153,7 @@ class Module extends \LORIS\Router\PrefixRouter implements RequestHandlerInterfa
      * Returns true if this module should be publicly available to people who
      * are not logged in to LORIS.
      *
-     * @return bool (false: not that this is not currently possible because main.php
+     * @return bool (false: note that this is not currently possible because main.php
      *         checks that the user is logged in before it calls NDB_Caller.)
      */
     public function isPublicModule()
```
johnsaigle: I wasn't able to add a return type here without offending `phan`. Thsi signature must match the function in `modules/api/php/module.class.inc` but when both are delcared to return bool I get the following error.
```
modules/login/php/module.class.inc:36 PhanParamSignatureRealMismatchReturnType Declaration of function isPublicModule() should be compatible with function isPublicModule() : bool (method returning '' cannot override method returning 'bool') defined in php/libraries/Module.class.inc:159
```

I wasn't able to fix it.

driusan: That sounds like the login module class (and likely all the module's module classes) needs to be updated to have the same signature as the parent class.

php/libraries/Module.class.inc
```diff
@@ -150,7 +154,7 @@ class Module extends \LORIS\Router\PrefixRouter implements RequestHandlerInterfa
      * Returns true if this module should be publicly available to people who
      * are not logged in to LORIS.
      *
-     * @return bool (false: not that this is not currently possible because main.php
+     * @return bool (false: note that this is not currently possible because main.php
```
driusan: I know you're just fixing a typo, but I don't think this note is relevant anymore (main.php is only used for instrument loading on major right now.)

johnsaigle: Should I delete it?

driusan: Might as well since you're changing it anyways, otherwise we'll have to do it later

zaliqarosli:
@johnsaigle please rebase to include https://github.com/aces/Loris/pull/3897 for testing

PR #4236, opened 2019-01-07
[Core] Add type declarations to Email class

johnsaigle:


driusan:
Only failing because of the problem where 7.3 is failing everything to major right now, so I'm merging.

PR #4237, opened 2019-01-07
[Core] Add return type declarations to Notifier classes

johnsaigle:
### Brief summary of changes

Usual typing changes. Also tried to make the code a little smaller in places.

### DIFF
--- a/php/libraries/NDB_Notifier_Abstract.class.inc
+++ b/php/libraries/NDB_Notifier_Abstract.class.inc
@@ -506,7 +506,7 @@ abstract class NDB_Notifier_Abstract
                    "mn" => $module,
                    "ot" => $operation_type,
                   );
-        return \Database::singleton()->pselectOne($query, $params);
+        return intval(\Database::singleton()->pselectOne($query, $params));
     }
 
     /**
@@ -518,16 +518,12 @@ abstract class NDB_Notifier_Abstract
      */
     public static function getNotificationServiceID(string $service): int
     {
-        $DB = Database::singleton();
-
         $query  = "SELECT id
                   FROM notification_services
                   WHERE service=:serv
                     ";
         $params = array("serv" => $service);
-        $result = $DB->pselectOne($query, $params);
-
-        return $result;
+        return intval(\Database::singleton()->pselectOne($query, $params));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


driusan:
This is failing both 7.2 and 7.3, not just 7.3

php/libraries/NDB_Notifier_Abstract.class.inc
```diff
@@ -356,15 +350,12 @@ abstract class NDB_Notifier_Abstract
      *
      * @return string  Phone number of the user
      */
-    protected function getPhone($uid)
+    protected function getPhone(int $uid): string
```
PapillonMcGill: Is this true when there is no phone number for that user?

php/libraries/NDB_Notifier_Abstract.class.inc
```diff
@@ -478,13 +468,10 @@ abstract class NDB_Notifier_Abstract
      */
     public static function getNotificationModules()
```
PapillonMcGill:
```suggestion
    public static function getNotificationModules() :array
```

php/libraries/NDB_Notifier_Abstract.class.inc
```diff
@@ -194,10 +194,10 @@ abstract class NDB_Notifier_Abstract
      *
      * @return string   string in the form `First Last <email@domain.com>`
      */
-    private function _getLoggedInUser()
+    private function _getLoggedInUser(): string
```
zaliqarosli: hey @johnsaigle you declared return type to be string as taken from the doc @return but the function actually returns an array. this is causing bugs in modules. how should we fix it? @driusan

driusan: The easiest way to fix it is to change the return type in both to be array.. but I don't really see why it's returning an array and not the User object itself. (In fact, at that point the function is just returning User::singleton, so I'm not sure why this function is needed at all...)

zaliqarosli: good question. i dont know, i'm not familiar with this class

zaliqarosli: the same return type bug is happening with the function directly below this one as well

johnsaigle: Should be fixed in #4323.

PR #4241, opened 2019-01-08
[Core: Candidate] Improve ID generation (CandID, PSCID, ExternalID)

johnsaigle:
### Brief summary of changes

This is a rewrite of the ID-generation in the Candidate class. I found the old code to be a little confusing and contained a lot of repetition. This PR aims to improve the efficiency, organization and readability of the ID generation code by organizing it into classes.

## New stuff
* Min and Max values for IDs are now actually implemented
* Static prefixes can be used for PSCIDs (previously was hard-coded to the site)
* Handles edge-cases like all the IDs being used or none existing
* Creates IDs more efficiently (previously the code iterated over every ID in sequence, effectively brute-forcing IDs until a free one was found...)
* Lots and lots of commenting to explain the confusing structure of the config file

### To test this change...
- [ ] Ensure that you can create a Candidate and that the IDs generated for them make sense.
- [ ] Modify the values in `project/config.xml` to be different than your current settings. For example, if your `generation` value is 'sequential', change it to random (or vice-versa). Make a new candidate and ensure the IDs make sense.'

For an example of how these settings should look, check the default config.xml file in the docs/ directory.


johnsaigle:
Opened this PR so that @PapillonMcGill and I could coordinate our efforts on modifying this code. This PR is still a work in progress until the following items are complete:

#### TODO

- [ ] Use max and min length values correctly
- [ ] Change ExternalID code to use new ID generation functions
- [ ] Change CandID code to use new ID generation functions
- [ ] Document all functions clearly
- [ ] PHPCS, phan, ..., etc.

php/libraries/Candidate.class.inc
```diff
@@ -20,11 +20,9 @@ define('PSCID_NOT_UNIQUE', 3);
 define('PSCID_INVALID_STRUCTURE', 4);
 define('EDC_NOT_SPECIFIED', 5);
 
-// id ranges...
 define('CANDIDATE_MIN_CANDID', 100000);
 define('CANDIDATE_MAX_CANDID', 999999);
-define('CANDIDATE_MIN_PSCID', 1000);
-define('CANDIDATE_MAX_PSCID', 9999);
+define('CANDID_LENGTH', 6);
```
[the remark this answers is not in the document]
PapillonMcGill: right, I just saw more parameter defined in the code without thinking of the purpose of the parameter.

php/libraries/Candidate.class.inc
```diff
@@ -680,115 +675,185 @@ class Candidate
...
-        while ($db->pselectOne(
-            "SELECT count(CandID) FROM candidate WHERE CandID=:CaID",
-            array('CaID' => $candID)
-        ) > 0) {
-            // pick a new candidate id
-            $candID = random_int(CANDIDATE_MIN_CANDID, CANDIDATE_MAX_CANDID);
-        }
-
-        return $candID;
+        return intval(\Candidate::_generateID('CandID'));
```
driusan: I don't see the benefit of doing this for CandIDs. It's a non-configurable random number between 100000-999999 which has no other features or customizations. Trying to re-use the PSCID/ExternalID logic for CandIDs makes the more convoluted and the above code that's being deleted it pretty straight forward. (Maybe it could be improved by using a do/while loop instead of a while loop, but the logic of "generate a random number, check if it exists", repeat if necessary is hidden and is much harder to follow with this change.)

johnsaigle: I see what you're saying. I agree the old code isn't so bad. It did seem to be almost exactly the same as the `_generateID` code and it seemed valuable to use that function's approach of querying the DB just once instead of querying the DB each time a value is generated. However that might not be as much of an issue since these are randomly generated.

Do you feel like your concern could be addressed with a code comment? It could describe how the ID is generated and this would allow us to still follow the DRY rule.

johnsaigle:
@PapillonMcGill I added functionality here to allow for the min and max settings as discussed in the other PR. Would you be able to test that this code satisfies your issue from before?

php/libraries/Candidate.class.inc
```diff
@@ -680,115 +678,258 @@ class Candidate
...
-                throw new LorisException(
-                    "Well this is embarrassing, our automated PSCID generator "
-                    . "is about to enter an infinite loop."
+            break;
+        case 'random':
+            $id = \Candidate::_generateNewRandomID($alphabet, $length);
+            // If the random ID is already in use, increment it until we find
+            // an unused value.
+            if (in_array(
+                str_pad($id, $length, '0', STR_PAD_LEFT),
```
PapillonMcGill:
```suggestion
                str_pad($id, $length, $alphabet[0], STR_PAD_LEFT),
```
as well as in line 813 and 906
might be alpha characters only

johnsaigle: Thanks, I'll make this change.

I think I'll also make the alpha characters uppercase instead of lowercase in the range.

johnsaigle:
Adding Needs Work because of above review and because Travis fails if `length` is made to be a mandatory config setting. I'll update the code to make this optional and use a constant defined in the source code to use as the default length when one is not configured.

php/libraries/Candidate.class.inc
```diff
@@ -20,11 +20,10 @@ define('PSCID_NOT_UNIQUE', 3);
 define('PSCID_INVALID_STRUCTURE', 4);
 define('EDC_NOT_SPECIFIED', 5);
 
-// id ranges...
 define('CANDIDATE_MIN_CANDID', 100000);
 define('CANDIDATE_MAX_CANDID', 999999);
-define('CANDIDATE_MIN_PSCID', 1000);
-define('CANDIDATE_MAX_PSCID', 9999);
+define('CANDID_LENGTH', 6);
+define('PSCID_LENGTH', 3);
```
zaliqarosli: 3 or 4? i have both in my DB for some reason

xlecours:
@johnsaigle Was it broken?

PapillonMcGill:
@johnsaigle You might have to rebase as you now have change to modules/new_profile/test/new_profileTest.php. Might be the cause for the No database connection" reported by Travis

johnsaigle:
@xlecours The old method of generating IDs wasn't broken but I found it hard to follow. It also didn't actually use certain config settings or constants included in the file and this PR addresses those.

Regarding efficiency, if you read the old implementation closely, it would generate a sequential ID by starting at e.g. `0000`, querying the DB for `MON0000`, then trying `0001`, querying the DB for `MON0001`, and so on. I think my code is more efficient because it cuts down on network requests to the server and doesn't essentially brute-force IDs.

I'm not opposed to refactoring this into separate classes if that's considered better.

php/libraries/Candidate.class.inc
```diff
@@ -680,115 +679,272 @@ class Candidate
...
+     * @param string $kind   One of 'PSCID or 'ExternalID'
+     * @param string $prefix The prefix to an ID. Usually a Site abbreviation
+     * such as MON for Montreal but can be a static value.
+     *
+     * @return string A new ID with prefix $prefix.
+     */
+    static function _generateID(string $kind, string $prefix = '')
+    {
+        // Validation
+        if (!in_array($kind, array('PSCID', 'ExternalID', 'CandID'), true)) {
```
driusan:
```suggestion
        if (!in_array($kind, array('PSCID', 'ExternalID'), true)) {
```

php/libraries/Candidate.class.inc
```diff
@@ -680,115 +679,272 @@ class Candidate
...
-        while ($db->pselectOne(
-            "SELECT count(CandID) FROM candidate WHERE CandID=:CaID",
-            array('CaID' => $candID)
-        ) > 0) {
-            // pick a new candidate id
-            $candID = random_int(CANDIDATE_MIN_CANDID, CANDIDATE_MAX_CANDID);
-        }
-
-        return $candID;
+        $existingIDs = \Candidate::_getExistingIDs('CandID');
```
driusan: This introduces a race condition which can result in multiple simultaneous requests generating the same CandID.

johnsaigle: I added an exception in the calling code to go off when a newly-generated PSCID or CandID is not unique.

driusan: That doesn't address the race condition. 2 different requests independently get a list of IDs here, and then once one inserts something the list that the other one is verifying against is stale.

johnsaigle: Do you have a suggestion of how to approach this? I'm not sure how to deal with this either than querying the DB like in the exception case I added.

PapillonMcGill: @johnsaigle As CandID is the primary key, trying to insert a CandID that is already used will generate an error. You can do a try catch (or while loop) and if the error is "existing primary key", increment CandID and/or PSCID.

Not perfect as you might have to increment a few time (would be rare but if 10 persons are adding new candidates, it might append. Also, when increasing the CandID, you risk going over the max value and/or exhausting the bank of possible values.

Unless you want to use a lock table (not the candidate table!) or other lock mechanism.
@driusan  which method would you prefer?

johnsaigle:
@xlecours I have refactored the code into classes. please let me know what you think. I followed the structure that you used in #4055 so that these PRs will be simple to combine in the future.

johnsaigle:
Still TODO:
- [ ] PHPDocs
- [ ] More unit tests
- [ ] General cleanup

PapillonMcGill:
@xlecours Esoterics but cool.  Just need to handle when the value is already taken.

php/libraries/Candidate.class.inc
```diff
@@ -256,7 +253,7 @@ class Candidate
         }
 
         // Generate candid
-        $candID = Candidate::_generateCandID();
+        $candID = (new CandIDGenerator)->generate();
```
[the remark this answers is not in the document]
driusan: I don't think PHP insists on those parenthesis, but I agree that it makes it more readable.

johnsaigle:
I refactored to use @xlecours new CandID class. I also tried a different approach to handling the race condition. I understand the `esoteric` approach to using MySQL to generate the value but it feels more convenient to generate the CandID using the `CandIDGenerator` and `CandID` classes than creating it in-line in MySQL.

Maybe @driusan could weigh in here on whether this is best done PHP side or in the MySQL query._

johnsaigle:
Unit tests will be added to #4305 to keep this PR small.

johnsaigle:
```
1) NextStageTestIntegrationTest::testNextStageDoesNotPageLoadWithDifferentStudySite
PDO::exec(): SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`LorisTest`.`candidate`, CONSTRAINT `FK_candidate_1` FOREIGN KEY (`RegistrationCenterID`) REFERENCES `psc` (`CenterID`))
```

The `resetStudySite` integration test function seems to be having an issue I'm not able to fix. Is this related to changes with `RegistrationCenterID`? In any case I think that this error is causing the test failures.

PapillonMcGill:
@johnsaigle 
integrity constraint violation:  mean that you try to delete a row in a table but another row (from another table) reference to the first one. You need to delete the with in that is referencing the one you try do delete first then you'll be able to delete the intended row.

johnsaigle:
@PapillonMcGill I understand the error but not why it is arising in this context. I haven't changed the integration tests for that code.

PapillonMcGill:
@johnsaigle Do you remove the candidates that are created in your test?
It is probable that at some point a test try to remove a psc but can't as your test created some candidate for that same psc.

johnsaigle:
@PapillonMcGill No I reverted my changes to the test so they should work the same as before. I tried adding a line to delete the candidate created but still this error appeared

src/StudyEntities/Candidate/SiteIDGenerator.php
```diff
@@ -0,0 +1,256 @@
...
+                . 'Please correct your configuration file.'
+                . "Length: `{$this->length}`\n"
+                . "Alphabet: `" . implode($this->alphabet) . "`\n"
+                . "Min: `{$this->minValue}`\n"
+                . "Max: `{$this->maxValue}`\n"
+                . "Length: `{$this->prefix}`\n"
+                . "Generation: `{$this->generationMethod}`\n"
+            );
+        }
+        if ($this->kind !== 'PSCID' && $this->kind !== 'ExternalID') {
```
xlecours: I think we could skip this check. If a config.xml contains infos about an other kind of ID, either validate of getExistingIds will fail if it is not configured.

php/libraries/Candidate.class.inc
```diff
@@ -281,14 +266,32 @@ class Candidate
...
-            $externalID = Candidate::_generateExternalID();
+            $externalID = (new ExternalIDGenerator())->generate();
             $setArray['ExternalID'] = $externalID;
         }
 
-        $db->insert('candidate', $setArray);
+        // Generate a new CandID and insert it into the DB. In case of a race
+        // condition, keep trying new IDs until an unused one is found.
+        $generator = new CandIDGenerator();
+        do {
```
xlecours: This do|while can be delegated to the generate function of CandIDGenerator since it already have a dependency on Database for getExistingIDs.
That way, the Generators will have the responsibility of returning a valid Id.

src/StudyEntities/Candidate/IdentifierGenerator.php
```diff
@@ -0,0 +1,138 @@
...
+            break;
+        }
+
+        // Return padded value with appropriate prefix
+        return $this->prefix . $id;
+    }
+
+    protected function generateSequentialID(string $id = ''): string
+    {
+        $newID = $id ?? max($this->getExistingIDs());
```
xlecours: `$id` will always equal the empty string if not given an other value. The max($this->getExistingIDs()) will never be evaluated

johnsaigle:
@xlecours Thanks for your review! I've incorporated your feedback.

src/StudyEntities/Candidate/CandIDGenerator.php
```diff
@@ -35,6 +41,33 @@ public function generate(): CandID {
...
+     * database. This function will generate new IDs in the edge-case that a
+     * race condition occurs and causes an error upon insertion.
+     *
+     * @param array $setArray An associative array of column names and values.
+     *
+     * @return CandID The newly-created CandID object.
+     */
+    public function createIDAndInsertWithValues(array $setArray): CandID {
+        do {
+            $invalidID = false;
```
driusan: The scoping of this is weird since it's used in the looping condition and I wouldn't be surprised if it eventually became a php/phan/phpcs/something warning about an undefined variable.

johnsaigle: Alrighty I'll move it

src/StudyEntities/Candidate/SiteIDGenerator.php
```diff
@@ -0,0 +1,258 @@
...
+        }
+        return array_pop($seqAttributes);
+    }
+
+    /**
+     * Traverse the $idStructure array and collect all values that exist
+     * for $setting.
+     *
+     * @param array  $idStructure Settings concerning ID structure extracted
+     *                  from project/config.sml
```
PapillonMcGill:
```suggestion
     *                  from project/config.xml
```

php/libraries/Candidate.class.inc
```diff
@@ -281,14 +265,18 @@ class Candidate
...
 
         if ($config->getSetting('useExternalID') === 'true') {
-            $externalID = Candidate::_generateExternalID();
+            $externalID = (new ExternalIDGenerator())->generate();
             $setArray['ExternalID'] = $externalID;
         }
 
-        $db->insert('candidate', $setArray);
+        // Generate a new CandID and insert it into the database.
+        $candID = (new CandIDGenerator())
```
xlecours: I don't think it should be the responsibility of the ID generator to insert the whole candidate in the database. It should just return a valid, unused CandID. To circumvent the race condition, we could use LOCK tables;
```
        $db = $factory->database();
        $db->run('LOCK TABLES candidate WRITE');
        $candID = (new CandIDGenerator())->generate();
        $setArray['CandID'] = $candID;
        $db->insert('candidate', $setArray);
        $db->run('UNLOCK TABLES');
```

src/StudyEntities/Candidate/CandIDGenerator.php
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * Wrapper for \Database::insert() to add new Candidate information to the
+     * database. This function will generate new IDs in the edge-case that a
+     * race condition occurs and causes an error upon insertion.
+     *
+     * @param array $setArray An associative array of column names and values.
+     *
+     * @return CandID The newly-created CandID object.
+     */
+    public function createIDAndInsertWithValues(array $setArray): CandID {
```
xlecours: This is doing to much.
See comment in Candidate::createNew

php/libraries/Candidate.class.inc
johnsaigle: Thanks, this is very helpful.

driusan: I don't think we can (or should) lock the table. It requires permission that the default SQL account shouldn't have.

johnsaigle: @driusan What are your thoughts on how to approach the race condition issue?

driusan: I don't see why it needs to be any more complicated than @PapillonMcGill's suggestion of having a try/catch inside the loop, since the CandID is a primary key.

johnsaigle: That's how I did it before but got a conflicting review from @xlecours. Can we land on a final approach before I redo this code again?

driusan: where is the conflicting review from @xlecours? All I can find is him saying to delegate the loop into the generator, but there's a lot of comments so I might be missing one that's hidden somewhere..

xlecours: @driusan The conflict is from me saying to delegate the loop to the generator, because to get the exception one need to insert in the table; which should ne be done by the Generator in my opinion.

johnsaigle: Hmm I may be getting my thoguhts twisted on this. I'm just going to write some code and see if I understand what everybody wants.

driusan: If that makes it impossible, I don't see the problem with keeping the loop outside of the generator. You need to be able to handle the failure case, and you can't do that without inserting, whether it's done in the generator or in the place that calls it.

johnsaigle: I changed it so that the try-catch insertion loop is in the Candidate class but the CandIDGenerator does an initial check on its own to make sure the ID it returns is valid in the first place.

johnsaigle:
@xlecours @driusan Can you let me know whether the new changes work for you?

src/StudyEntities/Candidate/IdentifierGenerator.php
```diff
@@ -0,0 +1,144 @@
...
+            strval($this->alphabet[0]),
+            STR_PAD_LEFT
+        );
+    }
+
+    /**
+     * @throws \LorisException
+     */
+    protected function checkIDRangeFull() {
+        // Check that the number of existing IDs does not exceed the range of
```
xlecours: This would make a good function doc comment. phpcs would like that as well

xlecours:
This is a good step forward. 
Thanks @johnsaigle for that refactoring. It definitely makes the code less monolithic.

src/StudyEntities/Candidate/IdentifierGenerator.php
johnsaigle: There's no PHPCS in `src/`. It's anarchy. ;)

I will make the change though.

src/StudyEntities/Candidate/IdentifierGenerator.php
```diff
@@ -0,0 +1,144 @@
...
+        }
+        // Create the new ID by incrementing the value of the $id parameter OR
+        // by incrementing the highest existing ID.
+        $newID = !empty($id) ? $id : max($this->getExistingIDs());
+        // Increment newID until we find an unused value within the boundaries
+        // of $min and $max.
+        do {
+            // Make sure $newID is greater than or equal to minimum and less
+            // than or equal to maximum.
+            if (strcmp(strval($newID++), strval($this->minValue)) < 0) {
```
driusan: if statements generally shouldn't have side-effects.

Moving $newID++ out of the if statement would avoid the need to figure out the order of operations when reviewing this code.

src/StudyEntities/Candidate/IdentifierGenerator.php
```diff
@@ -0,0 +1,144 @@
...
+                // If newID is less than the minimum, increment again.
+                continue;
+            }
+            if (strcmp(strval($newID), strval($this->maxValue)) > 0) {
+                // If newID has exceeded the maximum, wrap back around to the
+                // minimum possible value. We know that there is a free ID
+                // available somewhere in the range of possible IDs or else
+                // _generateID above will have thrown an exception. This allows
```
driusan: What if it was filled in between the start of the loop and here in a different request?

I'm also not convinced that we should be re-using IDs if somehow one got deleted in the middle, I could foresee that causing a lot of ambiguity in the data.

src/StudyEntities/Candidate/SiteIDGenerator.php
```diff
@@ -0,0 +1,258 @@
...
+            );
+        }
+        if ($setting === 'alphabet') {
+            switch ($seqValue) {
+            case 'alpha':
+                return range('A', 'Z');
+            case 'numeric':
+                return range(0, 9);
+            case 'alphanumeric':
+                return array_merge(range(0, 9), range('A', 'Z'));
```
driusan: Is this supposed to be '0' and '9'? This is merging an array of strings with an array of integers, they aren't really comparable types. (Same with numeric above)

johnsaigle: It doesn't seem to cause issues in my testing but I'll change it to be safe.

src/StudyEntities/Candidate/IdentifierGenerator.php
johnsaigle: I can add another check for existing IDs in this loop.

I can throw an exception if the maximum ID is equal to the max value as defined by the range instead of looping around but this might add a weird edge case if the most-recent Candidate gets deleted. I'm not sure how to enforce that its ID won't get re-used as there will be no record of it in the DB.

driusan: They should never be deleted anyways, they should just be turned inactive. I think it's safest to just throw an exception if it's greater than the maximum saying you've run out of IDs.

ridz1208:
@johnsaigle @driusan How can i bump the priority of this without having to actually do any work on it ?

would begging work ?

johnsaigle:
@ridz1208 It's done as far as I'm concerned. Just waiting on Dave to approve and merge.

xlecours:
Travis was failing on PHP7.3 only at the `sudo apt-get update` step. I restarted the job.

PR #4243, opened 2019-01-08
[Media] Data FrameWork Implementation

HenriRabalais:
### Brief summary of changes
Implemented the Data Framework for the Media Module. 

### To Test
- Make sure that everything is loading properly in the datatable
- Make sure the filter is filtering data properly
- Make sure that when the permissions are not set to 'superuser', the hidden files are indeed hidden from the datatable.

### DIFF
--- a/modules/media/jsx/mediaIndex.js
+++ b/modules/media/jsx/mediaIndex.js
@@ -178,7 +178,7 @@ class MediaIndex extends Component {
             <MediaUploadForm
               DataURL={`${loris.BaseURL}/media/ajax/FileUpload.php?action=getData`}
               action={`${loris.BaseURL}/media/ajax/FileUpload.php?action=upload`}
-              maxUploadSize={this.state.fieldOptions.maxUploadSize}
+              maxUploadSize={options.maxUploadSize}
             />
           </TabPane>
         );
--- a/modules/media/jsx/mediaIndex.js
+++ b/modules/media/jsx/mediaIndex.js
@@ -15,6 +15,7 @@ class MediaIndex extends Component {
 
     this.state = {
       data: {},
+      fieldOptions: {},
       error: false,
       isLoaded: false,
     };
@@ -38,7 +39,7 @@ class MediaIndex extends Component {
   fetchData() {
     return fetch(this.props.dataURL, {credentials: 'same-origin'})
       .then((resp) => resp.json())
-      .then((data) => this.setState({data}))
+      .then((data) => this.setState({data: data.data, fieldOptions: data.fieldOptions}))
       .catch((error) => {
         this.setState({error: true});
         console.error(error);
@@ -79,6 +80,9 @@ class MediaIndex extends Component {
         result = <td className={style}><a href={sessionURL}>{cell}</a></td>;
       }
       break;
+    case 'Site':
+      result = <td className={style}>{this.state.fieldOptions.sites[cell]}</td>;
+      break;
     case 'Edit Metadata':
       const editButton = (
         <TriggerableModal title="Edit Media File" label="Edit">
@@ -113,7 +117,7 @@ class MediaIndex extends Component {
     * XXX: Currently, the order of these fields MUST match the order of the
     * queried columns in _setupVariables() in media.class.inc
     */
-    const options = this.state.data.fieldOptions;
+    const options = this.state.fieldOptions;
     const fields = [
       {label: 'File Name', show: true, filter: {
         name: 'fileName',
@@ -174,7 +178,7 @@ class MediaIndex extends Component {
             <MediaUploadForm
               DataURL={`${loris.BaseURL}/media/ajax/FileUpload.php?action=getData`}
               action={`${loris.BaseURL}/media/ajax/FileUpload.php?action=upload`}
-              maxUploadSize={this.state.data.maxUploadSize}
+              maxUploadSize={this.state.fieldOptions.maxUploadSize}
             />
           </TabPane>
         );
@@ -186,7 +190,7 @@ class MediaIndex extends Component {
         <TabPane TabId={tabs[0].id}>
           <FilterableDataTable
             name="media"
-            data={this.state.data.Data}
+            data={this.state.data}
             fields={fields}
             getFormattedCell={this.formatColumn}
           />
--- a/modules/media/php/hidefilefilter.class.inc
+++ b/modules/media/php/hidefilefilter.class.inc
@@ -37,7 +37,7 @@ class HideFileFilter implements \LORIS\Data\Filter
      */
     public function filter(\User $user, \Loris\Data\DataInstance $resource) : bool
     {
-        if ($resource->hideFile) {
+        if ($resource->getHideFile()) {
             return false;
         } else {
             return true;
--- a/modules/media/php/hidefilefilter.class.inc
+++ b/modules/media/php/hidefilefilter.class.inc
@@ -12,7 +12,6 @@
  * @link       https://www.github.com/aces/Loris/
  */
 namespace LORIS\media;
-define('SHOW', 0);
 
 /**
  * HideFilesFilter filters out data for any resource that has a value of 1
@@ -27,6 +26,8 @@ define('SHOW', 0);
  */
 class HideFileFilter implements \LORIS\Data\Filter
 {
+    const HIDE = 1;
+
     /**
      * Implements the HideFileFilter interface
      *
@@ -38,6 +39,6 @@ class HideFileFilter implements \LORIS\Data\Filter
      */
     public function filter(\User $user, \Loris\Data\DataInstance $resource) : bool
     {
-        return $resource->getHideFile() === SHOW;
+        return $resource->getHideFile() !== self::HIDE;
     }
 }
--- a/modules/media/php/hidefilefilter.class.inc
+++ b/modules/media/php/hidefilefilter.class.inc
@@ -12,10 +12,11 @@
  * @link       https://www.github.com/aces/Loris/
  */
 namespace LORIS\media;
+define('SHOW', 0);
 
 /**
- * HideFilesFilter filters out data for any resource that has a value of 'true'
- * for the attribute 'hide_file'
+ * HideFilesFilter filters out data for any resource that has a value of 1
+ * for the attribute 'hide_file'.
  *
  * @category   Data
  * @package    Main
@@ -33,14 +34,10 @@ class HideFileFilter implements \LORIS\Data\Filter
      *                                           filtered for.
      * @param \LORIS\Data\DataInstance $resource The data being filtered.
      *
-     * @return bool true if the file passes the filter and should not be hidden.
+     * @return bool Whether the file should be shown.
      */
     public function filter(\User $user, \Loris\Data\DataInstance $resource) : bool
     {
-        if ($resource->getHideFile()) {
-            return false;
-        } else {
-            return true;
-        }
+        return $resource->getHideFile() === SHOW;
     }
 }
--- a/modules/media/php/hidefilefilter.class.inc
+++ b/modules/media/php/hidefilefilter.class.inc
@@ -20,7 +20,7 @@ namespace LORIS\media;
  * @category   Data
  * @package    Main
  * @subpackage Data
- * @author     Dave MacFarlane <david.macfarlane2@mcgill.ca>
+ * @author     Henri Rabalais <henri.rabalais@mcin.ca>
  * @license    http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
  * @link       https://www.github.com/aces/Loris/
  */
--- a/modules/media/php/hidefilefilter.class.inc
+++ b/modules/media/php/hidefilefilter.class.inc
@@ -1,0 +1,46 @@
+<?php
+/**
+ * This file provides an implementation of the hide files filter.
+ *
+ * PHP Version 7
+ *
+ * @category   Data
+ * @package    Main
+ * @subpackage Data
+ * @author     Henri Rabalais <henri.rabalais@mcin.ca>
+ * @license    http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link       https://www.github.com/aces/Loris/
+ */
+namespace LORIS\media;
+
+/**
+ * HideFilesFilter filters out data for any resource that has a value of 'true'
+ * for the attribute 'hide_file'
+ *
+ * @category   Data
+ * @package    Main
+ * @subpackage Data
+ * @author     Dave MacFarlane <david.macfarlane2@mcgill.ca>
+ * @license    http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link       https://www.github.com/aces/Loris/
+ */
+class HideFileFilter implements \LORIS\Data\Filter
+{
+    /**
+     * Implements the HideFileFilter interface
+     *
+     * @param \User                    $user     The user that the data is being
+     *                                           filtered for.
+     * @param \LORIS\Data\DataInstance $resource The data being filtered.
+     *
+     * @return bool true if the file passes the filter and should not be hidden.
+     */
+    public function filter(\User $user, \Loris\Data\DataInstance $resource) : bool
+    {
+        if ($resource->hideFile) {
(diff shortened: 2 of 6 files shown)


johnsaigle:
Did some small cleanup, hope that's cool! Mostly just adding return types, style things, phpcs.

Module seems to work fine but generates this PHP Warning:

```
PHP Warning:  Invalid argument supplied for foreach() in /var/www/loris/php/libraries/NDB_Menu_Filter.class.inc on line 149, referer: https://jsaigle-dev.loris.ca/media/ 
```

modules/media/php/mediafileprovisioner.class.inc
```diff
@@ -0,0 +1,76 @@
...
+               			s.CenterID as site,
+               			m.uploaded_by,
+               			m.date_taken,
+               			substring(m.comments, 1, 50) as comments,
+               			m.date_uploaded,
+               			m.file_type,
+               			s.CandID as cand_id,
+               			s.ID as session_id,
+               			m.hide_file,
+               			m.id
```
driusan: What are you trying to do with this convoluted/slow sub-subselect query that you couldn't just do with a join?

HenriRabalais: This is the same as the previous query, I think the columns are just misaligned so it looks like several sub-queries. I'll try and fix the formatting!

modules/media/php/mediafileprovisioner.class.inc
```diff
@@ -0,0 +1,76 @@
...
+{
+    /**
+     * Create a MediaFileProvisioner, which gets files for the meida
+     * menu table.
+     */
+    function __construct()
+    {
+        parent::__construct(
+            "SELECT m.file_name,                                                   
+                    (SELECT PSCID from candidate WHERE CandID=s.CandID) as pscid,
```
PapillonMcGill: This sub query should not be needed as the table `candidate` is part of the join. 
```suggestion
                    c.PSCID as pscid,
```

modules/media/php/mediafileprovisioner.class.inc
```diff
@@ -0,0 +1,76 @@
...
+               			m.date_uploaded,
+               			m.file_type,
+               			s.CandID as cand_id,
+               			s.ID as session_id,
+               			m.hide_file,
+               			m.id
+             FROM media m
+             LEFT JOIN session s ON m.session_id = s.ID
+             LEFT JOIN candidate c ON c.CandID=s.CandID
+             LEFT JOIN language l USING (language_id)",
```
driusan:
```suggestion
             LEFT JOIN language l USING (language_id)
      WHERE s.Active='Y' and c.Active='Y'",
```

Suggest adding the standard active flags

driusan: All queries in LORIS pretty much always should have the Active flag added, since Active='N' is effectively "Deleted" and should never be included, regardless of the data model being used for the row. Moving it into an ActiveFilter just adds an extra layer for getting rows which will always be filtered out.

modules/media/jsx/mediaIndex.js
```diff
@@ -163,7 +167,7 @@ class MediaIndex extends Component {
             <MediaUploadForm
               DataURL={`${loris.BaseURL}/media/ajax/FileUpload.php?action=getData`}
               action={`${loris.BaseURL}/media/ajax/FileUpload.php?action=upload`}
-              maxUploadSize={this.state.data.maxUploadSize}
+              maxUploadSize={this.state.fieldOptions.maxUploadSize}
```
zaliqarosli: could this.state.fieldOptions just be the constant options defined above?

HenriRabalais: sure could! I'll change that in a second

zaliqarosli:
on testing:

- not all the data loads
aces/major vs this branch:
screenshot 2019-01-14 14 07 34
screenshot 2019-01-14 14 08 16

- the visit label link is broken (the sessionID is cut off)
- when on non-superuser account, i get the error `SyntaxError: Unexpected end of JSON input
    at mediaIndex.js:39` and I don't have access to the module at all. `An error occured while loading the page.` is displayed

HenriRabalais:
@zaliqarosli strange, I've tried all three of those conditions and I'm not getting any of the errors you mentioned.

johnsaigle:
@HenriRabalais Could you please rebase this onto major? I have applied the `GenderToSex` [SQL patch](https://github.com/aces/Loris/pull/3897) so am unable to test your code at this time.

In terms of the discussed issues, I was also unable to load the module when on a non-superuser account:

```
PHP Fatal error:  Uncaught Error: Call to a member function filter() on null in /var/www/loris/modules/media/php/media.class.inc:133\nStack trace:\n#0 /var/www/loris/modules/media/php/media.class.inc(154):
```

Can't comment on the rest yet because of the rebase issue.

HenriRabalais:
@johnsaigle rebase done

HenriRabalais:
@zaliqarosli
1. I'm not getting a difference between my number of rows on `addDataFrameWorkToMedia` and `major`. @johnsaigle could you test this and see what happens for you?
2. Instrument Lists seems to be broken across all of `major` since `Sex` was changed to `Gender`. Not sure if that should be addressed in this PR.
3. This was due to typo and should be fixed now!

zaliqarosli:
1. Does `PHP Warning:  Invalid argument supplied for foreach() in /var/www/loris/php/libraries/NDB_Menu_Filter.class.inc on line 149, referer: http://zrosli-dev.loris.ca/media/` mean anything for this bug I'm getting?
2. This is no longer an issue after your rebase and after Gender to Sex patch is applied
3. works as expected!

zaliqarosli:
@johnsaigle nope, I have different data with that entry for 'Telephone Screening Interview' (in screenshot) never showing up

johnsaigle:
It's not 29 total rows like in the other screenshot? The total for the top one is cut off

zaliqarosli:
@johnsaigle nope, it is 28 
screenshot 2019-01-14 16 57 57

johnsaigle:
Cool just making absolutely sure!

zaliqarosli:
@johnsaigle it can't be a coincidence that the row that is missing for me happens to be the only file uploaded by not the admin hmmmm 🤔

johnsaigle:
My issue is caused by #4260 which is out of scope for this PR. It sounds like @zaliqarosli is having the inverse issue where an admin cannot see their own files but non-privileged users are able to.

modules/media/php/mediafileprovisioner.class.inc
```diff
@@ -0,0 +1,77 @@
...
+               			substring(m.comments, 1, 50) as comments,
+               			m.date_uploaded,
+               			m.file_type,
+               			s.CandID as cand_id,
+               			s.ID as session_id,
+               			m.hide_file,
+               			m.id
+             FROM media m
+             LEFT JOIN session s ON m.session_id = s.ID
+             LEFT JOIN candidate c ON c.CandID=s.CandID
```
PapillonMcGill: `session` and `candidate` tables can use regular inner JOIN as a `media` must have a session_id  and a session must have a CandID.

kongtiaowang:
To pass Travis, depends on https://github.com/aces/Loris/pull/4261.

HenriRabalais:
@johnsaigle Not sure why this is labeled as blocked

@zaliqarosli I can't reproduce the discrepancy and I don't see any indication of what might be causing it.

PapillonMcGill:
@HenriRabalais 
line 3376+ of travis report (PHP7.2)
There was 1 error:
1) MediaTest::testBrowseFilter
System info: host: '4ae5e781343e', ip: '172.18.0.3', os.name: 'Linux', os.arch: 'amd64', os.version: '4.4.0-101-generic', java.version: '1.8.0_91'
/app/modules/media/test/MediaTest.php:176
/app/modules/media/test/MediaTest.php:90

This mean the test called at line 90 of modules/media/test/MediaTest.php fail because of some null value injected in some js code.  Ensure that the code you change do not push a null value somewhere unexpected
Hope it help

johnsaigle:
@HenriRabalais I added `Blocked` because of Shen's comment above. I'll remove it.

zaliqarosli:
@HenriRabalais i'm still missing that one entry. i'll investigate!

kongtiaowang:
@HenriRabalais 
try to use unchanged test file.

zaliqarosli:
@HenriRabalais found the issue! the candidate whose media file was missing was not an active candidate. the file was showing on major but this PR actually fixes it and hides it like it should! so that's fine. i'm still getting an error when on a non admin account though, `mediaIndex.js:45 SyntaxError: Unexpected end of JSON input
    at mediaIndex.js:41` with `An error occured while loading the page.` rendered on the front-end

HenriRabalais:
@zaliqarosli woohoo! Cool. I'll look into that error you're getting.

johnsaigle:
Needs a quick PHPCS for modules/media/php/mediafileprovisioner.class.inc

modules/media/php/mediafile.class.inc
```diff
@@ -0,0 +1,75 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```
Nice to have strict typing on new classes.

modules/media/php/mediafileprovisioner.class.inc
```diff
@@ -0,0 +1,76 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

johnsaigle:
Not sure why Travis is failing :(

PapillonMcGill:
@johnsaigle 
From line 3900+ of Travis:
```
There was 1 error:
1) MediaTest::testBrowseFilter
System info: host: 'f379ad542ecb', ip: '172.18.0.2', os.name: 'Linux', os.arch: 'amd64', os.version: '4.4.0-101-generic', java.version: '1.8.0_91'
/app/modules/media/test/MediaTest.php:176
/app/modules/media/test/MediaTest.php:90
```

So something at line 90 of MediaTest.php cause an exception that is unkown in the JS code. Likely a variable as a null value where it should not be. Check the test condition and value extracted in the test DB for this test.

zaliqarosli:
passes my testing too!

HenriRabalais:
@kongtiaowang Any idea why this isn't passing travis?

driusan:
```
modules/media/php/media.class.inc:161 PhanTypeVoidAssignment Cannot assign void return value
```

(There's also one of the tests that are failing with an error.)

kongtiaowang:
@HenriRabalais modules/media/test/MediaTest.php, revert your changes in this PR.

kongtiaowang:
modules/media/php/media.class.inc:161 PhanTypeVoidAssignment Cannot assign void return value

modules/media/php/media.class.inc
```diff
@@ -186,9 +150,21 @@ class Media extends \NDB_Menu_Filter
...
-        return json_encode($result);
+        $table = (new \LORIS\Data\Table())
+            ->withDataFrom($this->getDataProvisioner());
+        $arr   = array_map(
+            function ($row) {
+                return array_values($row);
+            },
+            json_decode($table->toJSON(\User::singleton()), true)
+        );
+        $options = $this->getOptions();
```
kongtiaowang: remove 161

modules/media/php/media.class.inc
```diff
@@ -186,9 +150,21 @@ class Media extends \NDB_Menu_Filter
...
+        return json_encode(
+            array(
+             'data'         => $arr,
+             'fieldOptions' => $options,
```
kongtiaowang: 'fieldOptions' => $this->getOptions(),

HenriRabalais:
@kongtiaowang reverted MediaTest and made suggested changes.

modules/media/php/hidefilefilter.class.inc
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @category   Data
+ * @package    Main
+ * @subpackage Data
+ * @author     Henri Rabalais <henri.rabalais@mcin.ca>
+ * @license    http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link       https://www.github.com/aces/Loris/
+ */
+namespace LORIS\media;
+define('SHOW', 0);
```
driusan: This shouldn't be here. We're not using PHP4. (Also defining "SHOW" as 0 is a really weird and confusion convention.

modules/media/php/media.class.inc
```diff
@@ -95,87 +86,60 @@ class Media extends \NDB_Menu_Filter
...
-                               'fileTypes'   => $fileTypeList,
-                               'visits'      => $visitList,
-                               'sites'       => $siteList,
-                               'instruments' => $instrumentList,
-                               'languages'   => $languageList,
-                               'hidden'      => $hiddenOptions,
-                              ];
-
-        return true;
+        return array(
```
driusan: the documentation for this says it returns void

modules/media/php/hidefilefilter.class.inc
HenriRabalais: As in it shouldn't be declared using `define`? or that we should have class constants at all?

driusan: it should
1. use a class `const`, not `define()` (or maybe just a namespaced const if it doesn't make sense on theclass)
2. use a value other than "0" as the definition (since 0 is traditionally more falsey than truey and it's confusing to have Show be 0.

HenriRabalais: I could set the `HIDE` to `0` and have the function `return $resource->getHideFile() !== HIDE;` if that makes more sense.

HenriRabalais: I could set the value of `HIDE` to `1` and have the function `return $resource->getHideFile() !== HIDE;` if that makes more sense.

johnsaigle: @driusan This was my change, so my bad. The idea of using `0` was so that it would match the back-end value for the `hile_file` column which, I agree, should probably be `show_file` which would make truthy/falsey-ness less confusing.

modules/media/php/media.class.inc
johnsaigle: Odd that it passes `phan` considering this... @HenriRabalais It may be a good idea to add a return type here and also to rebase to `major` if you haven't for a while

maltheism:
Hi @HenriRabalais I was manually testing your PR and the showMediaError in FileUpload.php line 121 isn't being returned to the user.
Screen Shot 2019-04-10 at 4 12 03 PM

HenriRabalais:
@maltheism Strange!

As far as I know, my code in no way touched the file uploader so this issue must also be present on major and doesn't concern this PR.

modules/media/jsx/mediaIndex.js
```diff
@@ -113,7 +117,7 @@ class MediaIndex extends Component {
     * XXX: Currently, the order of these fields MUST match the order of the
     * queried columns in _setupVariables() in media.class.inc
     */
-    const options = this.state.data.fieldOptions;
+    const options = this.state.fieldOptions;
```
maltheism: Hi @HenriRabalais I know this is just accessing the maxUploadSize but is it good practice to avoid possible state mutation by not grabbing something directly from the state?
Instead doing something like:
`const options = Object.assign({}, this.state.fieldOptions);`

driusan: What mutation are you referring to? This const seems to only ever be read.

HenriRabalais:
@maltheism Have you verified that this is an error with my PR and not with the major branch in general? The error you described does not seem pertinent to my code changes.

maltheism:
@HenriRabalais on major the error message does display.
error_on_major
Seems to be an issue with changes in this PR.

modules/media/jsx/mediaIndex.js
maltheism: @driusan just a concern for a potential problem in the future if new code is introduced. @HenriRabalais told me it's bad to directly grab a value from the state instead of copying it.

HenriRabalais:
@ridz1208 @driusan approved and passed manual testing.

PR #4077, opened 2018-11-07
Update API summary header and qc too grep more info

cmadjar:
### Brief summary of changes

**Updates to the API v.0.0.3 on the imaging side:**
- Addition of scanner information to the summary header of the API for a given filename (manufacturer, model, software version, serial number)
- Addition of the image Caveat flag with the reason for the Caveat flag in the file qc page of the API

**Fix to API v0.0.2 on the imaging side:**
- The selected were not queried at the right place. The selected has been part of the files_qcstatus for several LORIS releases so updated the API queries so we can grep the Selected flag (before it returned some hash, now it returns True or False)

### This resolves issue...

Did not take the time to create an issue for that. Needed for the phantom VM so I figured I would issue a PR to the LORIS repo at the same time :).

### To test this change...

You need imaging data to test this

- [ ] test $API/candidates/$CandID/$VisitLabel/images/$filename/headers (should see ScannerInfo now in addition to everything else that was already in the JSON)
- [ ] test $API/candidates/$CandID/$VisitLabel/images/$filename/qc (make sure the selected value is properly displayed with this PR and that Caveats is added to the JSON)

# If this sentence shows up it means I didn't read the instructions!

Yes I read it ;)

PR #4244, opened 2019-01-09
[Api] /candidates and sub-endpoints

### LINKED ISSUE #5131: [API] Create Candidate with Wrong Project returns 500
When creating a candidate via the api using a project that doesn't exists gives the end user a 500 Internal Error. This should instead return a 400 Bad Request error.

https://redmine.cbrain.mcgill.ca/issues/12951

xlecours:
This moves all the remaining endpoints from `htdocs/api/...` to `src/`

It adds classes in `php/libraries/` so **`$ composer dump-autoload`** need to be run.

Based on https://github.com/aces/Loris/pull/4220
 - [x] candidates(/*)$
 - [x] candidates/([0-9]+)(/*)$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)(/*)$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/instruments$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/instruments/([a-zA-Z0-9_.]+)$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/instruments/([a-zA-Z0-9_.]+)/dde$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/instruments/([a-zA-Z0-9_.]+)/flags$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/instruments/([a-zA-Z0-9_.]+)/dde/flags$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/images$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/qc/imaging$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/images/([a-zA-Z0-9_.-]+)/qc$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/images/([a-zA-Z0-9_.-]+)/format/raw$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/images/([a-zA-Z0-9_.-]+)/format/thumbnail$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/images/([a-zA-Z0-9_.-]+)/format/brainbrowser$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/images/([a-zA-Z0-9_.-]+)$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/images/([a-zA-Z0-9_.-]+)/headers$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/images/([a-zA-Z0-9_.-]+)/headers/full$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/images/([a-zA-Z0-9_.-]+)/headers/(.*)$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/dicoms$
 - [x] candidates/([0-9]+)/([a-zA-Z0-9_.]+)/dicoms/([a-zA-Z0-9_.-]+)$

todo:
 - [ ] Add unit tests
 - [x] Integrate changes from https://github.com/aces/Loris/pull/4077
 - [x] Review files doc comments
 - [x] Rebase after release

---

Fixes #5131


modules/api/php/candidatesrow.class.inc
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Returns the CenterID for this row, for filters such as
+     * \LORIS\Data\Filters\UserSiteMatch to match again.
+     *
+     * @return integer The CenterID
+     */
+    public function getCenterID(): int
+    {
+        return $this->DBRow['CenterID'];
```
johnsaigle:
```suggestion
        return intval($this->DBRow['CenterID']);
```

I'm pretty sure the results from the DB are always stings so if you are using strict return types you need to use `intval()`

modules/api/php/projectimagesrow.class.inc
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * Returns the CenterID for this row, for filters such as
+     * \LORIS\Data\Filters\UserSiteMatch to match again.
+     *
+     * @return integer The CenterID
+     */
+    public function getCenterID(): int
+    {
+        return $this->DBRow['CenterID'];
```
johnsaigle:
```suggestion
        return intval($this->DBRow['CenterID']);
```

modules/api/php/candidatesrow.class.inc
xlecours: good catch

modules/api/php/candidatesprovisioner.class.inc
```diff
@@ -0,0 +1,91 @@
...
+               c.EDC as EDC,
+               c.DoB as DoB,
+               c.Gender as Sex
+             FROM
+               candidate c
+             JOIN psc s
+               ON (s.CenterID = c.CenterID)
+             LEFT JOIN Project p
+               ON (c.ProjectID = p.ProjectID)
+             WHERE Active="Y"
```
PapillonMcGill:
```suggestion
             WHERE c.Active="Y"
```
prevent confusion, specially in future if tables changes or more are added

src/Api/Endpoints/Candidates.php
```diff
@@ -0,0 +1,234 @@
...
+     * @return ResponseInterface The outgoing PSR7 response
+     */
+    public function handle(ServerRequestInterface $request) : ResponseInterface
+    {
+        $user = $request->getAttribute('user');
+        if ($user instanceof \LORIS\AnonymousUser) {
+            return new \LORIS\Http\Response\Unauthorized();
+        }
+
+        if (!$this->hasAccess($user)) {
```
PapillonMcGill: would it revel that the user is valid compare to the previous error?

xlecours: The previous checks if the user is logged in.
The second check for that user's permissions

modules/api/php/visitinstrumentsrow.class.inc
```diff
@@ -0,0 +1,49 @@
+<?php declare(strict_types=1);
+/**
+ * This class implements a data Instance for visit instruments
```
PapillonMcGill:
```suggestion
 * This class implements a data Instance for visit instruments.
```

modules/api/php/visitinstrumentsrowprovisioner.class.inc
```diff
@@ -0,0 +1,71 @@
...
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+
+namespace LORIS\api;
+
+use \LORIS\Data\Provisioners\DBRowProvisioner;
+/**
+ * This class implements a data provisioner to get all possible rows
+ * for the projects/$projectname/instruments endpoint
```
PapillonMcGill:
```suggestion
 * for the projects/$projectname/instruments endpoint.
```

php/libraries/Image.class.inc
```diff
@@ -0,0 +1,256 @@
...
+ *  @category Imaging
+ *  @package  Main
+ *  @author   Xavier Lecours <xavier.lecours@mcin.ca>
+ *  @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ *  @link     https://www.github.com/aces/Loris/
+ */
+namespace LORIS;
+
+/**
+ * This class defines a Image to allows one to type hint function's parameters
```
PapillonMcGill:
```suggestion
 * This class defines a Image to allows one to type hint function's parameters.
```

php/libraries/Image.class.inc
```diff
@@ -0,0 +1,256 @@
...
+            $this->_filename     = basename($dbrow['filelocation']);
+            $this->_filelocation = $dbrow['filelocation'];
+            $this->_outputtype   = $dbrow['outputtype'];
+            $this->_acquisitionprotocol = $dbrow['acquisitionprotocol'];
+            $this->_filetype            = $dbrow['filetype'];
+        }
+    }
+
+    /**
+     * Get file headers from paramter file
```
PapillonMcGill:
```suggestion
     * Get file headers from paramter file.
```

php/libraries/TimePoint.class.inc
```diff
@@ -921,5 +921,39 @@ class TimePoint
         return $candidate->getCandidateDoB();
     }
 
-} // end class
+    /**
+     * Gets the Images of that session
```
PapillonMcGill:
```suggestion
     * Gets the Images of that session.
```

php/libraries/TimePoint.class.inc
```diff
@@ -921,5 +921,39 @@ class TimePoint
...
+        $provisioner = new \LORIS\api\VisitImagesRowProvisioner($this);
+        $traversable = (new \LORIS\Data\Table())
+            ->withDataFrom($provisioner)
+            ->getRows($user);
+
+        return iterator_to_array($traversable);
+    }
+
+    /**
+     * Gets the DicomTar of that session
```
PapillonMcGill:
```suggestion
     * Gets the DicomTar of that session.
```

src/Api/Endpoints/Candidate/Candidate.php
```diff
@@ -0,0 +1,146 @@
+<?php declare(strict_types=1);
+/**
+ * This implements the Candidate endpoint class under Candidates
```
PapillonMcGill:
```suggestion
 * This implements the Candidate endpoint class under Candidates.
```

src/Api/Endpoints/Candidate/Candidate.php
```diff
@@ -0,0 +1,146 @@
...
+    {
+        return array(
+                'v0.0.1',
+                'v0.0.2',
+                'v0.0.3-dev',
+               );
+    }
+
+    /**
+     * Handles a request that starts with /candidates/$candid
```
PapillonMcGill:
```suggestion
     * Handles a request that starts with /candidates/$candid.
```

src/Api/Endpoints/Candidate/Candidate.php
```diff
@@ -0,0 +1,146 @@
...
+        );
+
+        return $handler->process(
+            $newrequest,
+            $handler
+        );
+    }
+
+    /**
+     * Implements the ETagCalculator interface
```
PapillonMcGill:
```suggestion
     * Implements the ETagCalculator interface.
```

src/Api/Endpoints/Candidate/Visit/Dicom/Dicom.php
```diff
@@ -0,0 +1,170 @@
+<?php declare(strict_types=1);
+/**
+ * This implements a visit's dicom endpoint class
```
PapillonMcGill:
```suggestion
 * This implements a visit's dicom endpoint class.
```

src/Api/Endpoints/Candidate/Visit/Dicom/Dicom.php
```diff
@@ -0,0 +1,170 @@
...
+     *
+     * @return array a list of supported API versions.
+     */
+    protected function supportedVersions() : array
+    {
+        return array('v0.0.3-dev');
+    }
+
+    /**
+     * Handles a request to /candidates/$candid/$visit_label/dicoms/$tarname
```
PapillonMcGill:
```suggestion
     * Handles a request to /candidates/$candid/$visit_label/dicoms/$tarname.
```

src/Api/Endpoints/Candidate/Visit/Dicom/Dicom.php
```diff
@@ -0,0 +1,170 @@
...
+                );
+            }
+        }
+
+        // Delegate to sub-endpoints
+        return new \LORIS\Http\Response\NotFound();
+    }
+
+    /**
+     * Create an array representation of this endpoint's reponse body
```
PapillonMcGill:
```suggestion
     * Create an array representation of this endpoint's reponse body.
```

src/Api/Endpoints/Candidate/Visit/Dicom/Dicom.php
```diff
@@ -0,0 +1,170 @@
...
+                    new \LORIS\Http\StringStream(
+                        $file->fread($file->getSize())
+                    )
+                );
+        }
+        return $this->cache;
+    }
+
+    /**
+     * Implements the ETagCalculator interface
```
PapillonMcGill:
```suggestion
     * Implements the ETagCalculator interface.
```

src/Api/Endpoints/Candidate/Visit/Dicoms.php
```diff
@@ -0,0 +1,145 @@
+<?php declare(strict_types=1);
+/**
+ * This implements the visit's dicoms endpoint class
```
PapillonMcGill:
```suggestion
 * This implements the visit's dicoms endpoint class.
```

src/Api/Endpoints/Candidate/Visit/Dicoms.php
```diff
@@ -0,0 +1,145 @@
...
+     *
+     * @return array a list of supported API versions.
+     */
+    protected function supportedVersions() : array
+    {
+        return array('v0.0.3-dev');
+    }
+
+    /**
+     * Handles a request that starts with /candidates/$candid
```
PapillonMcGill:
```suggestion
     * Handles a request that starts with /candidates/$candid.
```

src/Api/Endpoints/Candidate/Visit/Dicoms.php
```diff
@@ -0,0 +1,145 @@
...
+        );
+
+        return $handler->process(
+            $newrequest,
+            $handler
+        );
+    }
+
+    /**
+     * Create an array representation of this endpoint's reponse body
```
PapillonMcGill:
```suggestion
     * Create an array representation of this endpoint's reponse body.
```

src/Api/Endpoints/Candidate/Visit/Dicoms.php
```diff
@@ -0,0 +1,145 @@
...
+                ...$dicomtars
+            ))->toArray();
+
+            $this->cache = new \LORIS\Http\Response\JsonResponse($view);
+        }
+        return $this->cache;
+    }
+
+    /**
+     * Implements the ETagCalculator interface
```
PapillonMcGill:
```suggestion
     * Implements the ETagCalculator interface.
```

src/Api/Endpoints/Candidate/Visit/Image/Image.php
```diff
@@ -0,0 +1,185 @@
+<?php declare(strict_types=1);
+/**
+ * This implements the visit's images endpoint class
```
PapillonMcGill:
```suggestion
 * This implements the visit's images endpoint class.
```

Let stop here .... for now.

php/libraries/DicomTar.class.inc
```diff
@@ -0,0 +1,485 @@
...
+ * This class defines a DicomTarSerieDTO which is an immutable representation of a
+ * Dicom series object.
+ *
+ *  @category Imaging
+ *  @package  Main
+ *  @author   Xavier Lecours <xavier.lecours@mcin.ca>
+ *  @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ *  @link     https://www.github.com/aces/Loris/
+ */
+class DicomSerieDTO
```
cmadjar: again with `seriesdescription` and `seriesnumber`. Seems clearer to use these terms and does not take much more place in the code

php/libraries/DicomTar.class.inc
```diff
@@ -0,0 +1,485 @@
...
+{
+    private $_id;
```
cmadjar: might be worth clarifying which ID it retruns. I get you are talking about the TarchiveSeriesID and not the tarchiveID but could be confusing to which ID you are referring to in this class.

php/libraries/DicomTar.class.inc
```diff
@@ -0,0 +1,485 @@
...
+        $this->_modality       = $modality;
+        $this->_seriesuid      = $seriesuid;
+    }
+
+    /**
+     * Accessor for id
+     *
+     * @return ?int
+     */
+    public function getId()
```
cmadjar: would tend to rename this function to `getTarchiveSeriesID`

php/libraries/DicomTar.class.inc
```diff
@@ -0,0 +1,485 @@
...
+    {
+        return $this->_id;
+    }
+
+    /**
+     * Accessor for description
+     *
+     * @return ?string
+     */
+    public function getDescription()
```
cmadjar: would tend to rename to `getSeriesDescription`

php/libraries/DicomTar.class.inc
```diff
@@ -0,0 +1,485 @@
...
+    {
+        return $this->_description;
+    }
+
+    /**
+     * Accessor for number
+     *
+     * @return ?int
+     */
+    public function getNumber()
```
cmadjar: would rename to `getSeriesNumber` for clarity

php/libraries/DicomTar.class.inc
```diff
@@ -0,0 +1,485 @@
...
+     *
+     * @return ?string
+     */
+    public function getRepetitiontime()
+    {
+        return $this->_repetitiontime;
+    }
+
+    /**
+     * Accessor for invertiontime
```
cmadjar: misspelled `inversiontime ` (replace first `t` by `s`)

php/libraries/Image.class.inc
```diff
@@ -0,0 +1,256 @@
+<?php declare(strict_types=1);
+/**
+ * File describe the Image class
```
cmadjar: File **that** describes the Image class

php/libraries/Image.class.inc
cmadjar: ... defines **an** Image to **allow** one ...

php/libraries/Image.class.inc
```diff
@@ -0,0 +1,256 @@
...
+    private $_fileid;
+
+    private $_filename;
+
+    private $_filelocation;
+
+    /**
+     * Constructor
+     *
+     * @param ?int $fileid This File unique id
```
cmadjar: would remove the term `unique` here.

php/libraries/Image.class.inc
```diff
@@ -0,0 +1,256 @@
...
+                     FROM
+                       files f 
+                     JOIN mri_scan_type mst
+                       ON (mst.ID=f.AcquisitionProtocolID)
+                     WHERE f.FileID = :v_fileid
+                    ',
+                    array('v_fileid' => $fileid)
+                );
+            if (empty($dbrow)) {
+                throw new \NotFound('There is no file with that id');
```
cmadjar: ... with that **FileID**

php/libraries/Image.class.inc
cmadjar: ...from ** the parameter_file table**

php/libraries/Image.class.inc
```diff
@@ -0,0 +1,256 @@
...
+            $this->_outputtype   = $dbrow['outputtype'];
+            $this->_acquisitionprotocol = $dbrow['acquisitionprotocol'];
+            $this->_filetype            = $dbrow['filetype'];
+        }
+    }
+
+    /**
+     * Get file headers from paramter file
+     *
+     * @param string $headername $the parameter name to reteive
```
cmadjar: ... $headername **the** parameter name to **retrieve** . ($ sign before `the` to remove and add a `r` in `retrieve`)

php/libraries/Image.class.inc
```diff
@@ -0,0 +1,256 @@
...
+                    array(
+                     'v_fileid'     => $this->_fileid,
+                     'v_headername' => $headername,
+                    )
+                );
+        return $dbrow['Value'] ?? '';
+    }
+
+    /**
+     * Creates a imutable object representation of this tarchive
```
cmadjar: Creates **an immutable** object representation of this **file** (or replace my `file` suggestion by `image` if you prefer)

php/libraries/Image.class.inc
```diff
@@ -0,0 +1,256 @@
...
+            $this->_filelocation,
+            $this->_outputtype,
+            $this->_acquisitionprotocol,
+            $this->_filetype
+        );
+    }
+}
+
+/**
+ * This class defines a ImageDTO which is an immutable representation of a
```
cmadjar: This class defines **an** ....

php/libraries/Image.class.inc
```diff
@@ -0,0 +1,256 @@
...
+     *
+     * @return int
+     */
+    public function getFileid(): int
+    {
+        return $this->_fileid;
+    }
+
+    /**
+     * METHODDESCRIPTION
```
cmadjar: missing proper description of the method here

php/libraries/Image.class.inc
```diff
@@ -0,0 +1,256 @@
...
+     * @access public
+     * @return RETURNTYPE RETURNDESCRIPTION
+     */
+    public function getFilename()
+    {
+        return $this->_filename;
+    }
+
+    /**
+     * METHODDESCRIPTION
```
cmadjar: missing proper description of the method here

php/libraries/Image.class.inc
```diff
@@ -0,0 +1,256 @@
...
+     * @access public
+     * @return RETURNTYPE RETURNDESCRIPTION
+     */
+    public function getFilelocation()
+    {
+        return $this->_filelocation;
+    }
+
+    /**
+     * METHODDESCRIPTION
```
cmadjar: missing proper description of the method here

php/libraries/Image.class.inc
```diff
@@ -0,0 +1,256 @@
...
+     * @access public
+     * @return RETURNTYPE RETURNDESCRIPTION
+     */
+    public function getOutputtype()
+    {
+        return $this->_outputtype;
+    }
+
+    /**
+     * METHODDESCRIPTION
```
cmadjar: missing proper description of the method here

php/libraries/Image.class.inc
```diff
@@ -0,0 +1,256 @@
...
+     * @access public
+     * @return RETURNTYPE RETURNDESCRIPTION
+     */
+    public function getAcquisitionprotocol()
+    {
+        return $this->_acquisitionprotocol;
+    }
+
+    /**
+     * METHODDESCRIPTION
```
cmadjar: missing proper description of the method here

php/libraries/TimePoint.class.inc
cmadjar: again, a given session could have multiple DicomTars so this should be reflected in the documentation

src/Api/Endpoints/Candidate/Visit/Dicom/Dicom.php
cmadjar: ... of this endpoint's **response** body

src/Api/Endpoints/Candidate/Visit/Dicoms.php
cmadjar: ... this endpont's **response** body

src/Api/Endpoints/Candidate/Visit/Image/Image.php
```diff
@@ -0,0 +1,185 @@
...
+                );
+            }
+        }
+
+        // Delegate to sub-endpoints
+        return new \LORIS\Http\Response\NotImplemented();
+    }
+
+    /**
+     * Create an array representation of this endpoint's reponse body
```
cmadjar: ... endpoint's **response** body

src/Api/Endpoints/Candidate/Visit/Instruments.php
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        $endpoint = new Instrument\Instrument($this->visit, $instrument);
+        $request  = $request->withAttribute('pathparts', $pathparts);
+
+        return $endpoint->process($request, $endpoint);
+    }
+
+    /**
+     * Create an array representation of this endpoint's reponse body
```
cmadjar: typo in **response** again :)

src/Api/Endpoints/Candidate/Visit/Qc.php
```diff
@@ -0,0 +1,207 @@
+<?php declare(strict_types=1);
+/**
```
cmadjar: Can you please make sure to integrate the changes from this PR that got merge for the imaging QC information?
https://github.com/aces/Loris/pull/4077

modules/api/php/candidatesprovisioner.class.inc
```diff
@@ -0,0 +1,91 @@
...
+            '
+             SELECT 
+               c.CandID as CandID,
+               p.Name as ProjectName,
+               c.PSCID as PSCID,
+               c.CenterID as CenterID,
+               s.Name as SiteName,
+               c.EDC as EDC,
+               c.DoB as DoB,
+               c.Gender as Sex
```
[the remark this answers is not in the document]
driusan: or a unit test?

modules/api/php/visitdicomrow.class.inc
```diff
@@ -0,0 +1,47 @@
...
+ * @package  Loris
+ * @author   Xavier Lecours Boucher <xavier.lecours@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+
+namespace LORIS\api;
+
+/**
+ * A VisitDicomRow is a single dicom archive
```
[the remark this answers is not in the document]
xlecours: @cmadjar Yes, and VisitDicomRow is on instance of a DICOM for a visit. Should I change the description for `A VisitDicomRow is a single dicom archive among that visit DICOM archives.` ?

modules/api/php/candidatesrow.class.inc
```diff
@@ -0,0 +1,70 @@
...
+    public function toJSON() : string
+    {
+        $obj = array(
+                'CandID'  => $this->DBRow['CandID'],
+                'Project' => $this->DBRow['ProjectName'],
+                'PSCID'   => $this->DBRow['PSCID'],
+                'Site'    => $this->DBRow['SiteName'],
+                'EDC'     => $this->DBRow['EDC'],
+                'DoB'     => $this->DBRow['DoB'],
+                'Gender'  => $this->DBRow['Sex'],
```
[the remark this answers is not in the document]
xlecours: The API specs still have Gender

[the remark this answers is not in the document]
driusan: v0.0.3-dev says Sex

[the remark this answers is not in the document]
xlecours: great!

modules/api/php/visitdicomrowprovisioner.class.inc
```diff
@@ -0,0 +1,75 @@
+<?php declare(strict_types=1);
+/**
+ * This file implements a data provisioner to get a single dicom
```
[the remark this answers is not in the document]
xlecours: It will return 2 VisitDicomsRow. I changed the description to say: `This file implements a data provisioner to get all dicoms from a visit.`

modules/api/php/visitdicomrowprovisioner.class.inc
```diff
@@ -0,0 +1,75 @@
...
+ * @package  Loris
+ * @author   Xavier Lecours Boucher <xavier.lecours@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+class VisitDicomRowProvisioner extends DBRowProvisioner
+{
+    /**
+     * Create a RowProvisioner, which gets rows for
+     * projects/$projectname/instruments endpoint.
```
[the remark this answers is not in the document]
xlecours: Changed to `Create a RowProvisioner, which gets rows for a visit's Dicoms.`

php/libraries/DicomTar.class.inc
```diff
@@ -0,0 +1,485 @@
...
+            return array();
+        }
+
+        $dbrow = \NDB_Factory::singleton()
+            ->database()
+            ->pselect(
+                '
+                 SELECT
+                   TarchiveSeriesID as id,
+                   SeriesDescription as description,
```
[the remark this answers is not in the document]
xlecours: This is internal to the class, I will keep it as is but the exposed part show Series...

modules/api/php/visitdicomrow.class.inc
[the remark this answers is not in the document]
cmadjar: I think your suggestion would clarify it. (also, DICOM instead of dicom ;)) Thanks!

modules/api/php/projectinstrumentsrow.class.inc
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Returns this row's shortname.
+     *
+     * @return string
+     */
+    public function getShortname() : string
+    {
+        return $this->DBRow['shortname'] ?? null;
```
driusan: null is not a string (same for most of these)

xlecours:
@cmadjar The PR adds classes in `php/libraries/` so **`$ composer dump-autoload`** need to be run.

cmadjar:
@xlecours Travis failure:
```
make: *** [checkstatic] Error 1
The command "make checkstatic" exited with 2.
```
Not sure what that means though...

php/libraries/Image.class.inc
```diff
@@ -0,0 +1,414 @@
...
+    /**
+     * Return a SPLFileInfo object based on this images's properties.
+     *
+     * @return \SplFileInfo
+     */
+    public function getFileInfo(): \SplFileInfo
+    {
+        $imagepath = \NDB_factory::singleton()
+            ->config()
+            ->getSetting('mincPath');
```
cmadjar: This setting does not exist anymore since 21.0. Please use "imagePath" instead

src/Api/Endpoints/Candidate/Visit/Dicom/Dicom.php
```diff
@@ -0,0 +1,164 @@
...
+                ->getSetting('tarchiveLibraryDir');
+
+            $fullpath = $tarchivepath . $dicom->getArchiveLocation();
+            $info     = new \SplFileInfo($fullpath);
+            if (!$info->isFile()) {
+                return new \LORIS\Http\Response\NotFound();
+            }
+
+            if (!$info->isReadable()) {
+                return new \LORIS\Http\Response\NotFound();
```
cmadjar: shouldn't it say "not readable" instead of not found here? Technically, the file is found but cannot be read due to wrong permissions. Would be helpful to send the proper error message when debugging why a file is not found.

cmadjar: Same comment to apply for when we download MINC files BTW.

xlecours: @cmadjar I don't want to give to much info about the existence of files. But if you think there should be a message for unreadable file, I can add that.

cmadjar: @xlecours Is there a reason why we don't want to give to much info about the existence of files? If that is the convention, no worries.

cmadjar:
PS: @xlecours failing travis :(

src/Api/Endpoints/Candidate/Visit/Dicom/Dicom.php
xlecours: @cmadjar I guess it's paranoia. @johnsaigle , any thoughts?

johnsaigle: I think it's fine in this case. The user is already logged in at this point so there's some level of trust there.

modules/imaging_browser/php/imaging_session_controlpanel.class.inc
```diff
@@ -159,16 +159,12 @@ class Imaging_Session_ControlPanel
...
-                    " MRICaveat FROM session WHERE ID=:sid";
-        $qcstatus = $DB->pselectRow(
-            $query,
-            array("sid" => $this->sessionID)
-        );
 
-        $subjectData['mriqcstatus']  = $qcstatus['MRIQCStatus'];
-        $subjectData['mriqcpending'] = $qcstatus['MRIQCPending'];
-        $subjectData['mricaveat']    = $qcstatus['MRICaveat'];
+        $qcstatus = $timePoint->getImagingQC();
```
xlecours: This is unrelated to the API but I wanted to show usage of the add `getImagingQC` function in TimePoint outside of the API.

php/libraries/ImageQcStatus.class.inc
```diff
@@ -0,0 +1,78 @@
...
+     * Constructor
+     *
+     * @param ?string $qcstatus Either "Pass" or "Fail"
+     * @param ?bool   $selected If selected
+     *
+     * @throws \DomainException When $qcstatus is not one of PASS or FAIL
+     */
+    public function __construct(?string $qcstatus, ?bool $selected)
+    {
+        if (!in_array($qcstatus, ['Pass', 'Fail', null], true)) {
```
xlecours: Not sure if we should do Database validations here. Maybe it's good to have it as a fail early.

xlecours:
@cmadjar @driusan This is ready for review and testing.

modules/api/php/candidatesprovisioner.class.inc
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Returns an instance of a CandidatesRow object for a given
+     * table row.
+     *
+     * @param array $row The database row from the LORIS Database class.
+     *
+     * @return \LORIS\Data\DataInstance
+     */
+    public function getInstance($row) : \LORIS\Data\DataInstance
```
johnsaigle:
```suggestion
    public function getInstance(array $row) : \LORIS\Data\DataInstance
```

modules/api/php/projectimagesrow.class.inc
```diff
@@ -33,7 +44,18 @@ class ProjectImagesRow implements \LORIS\Data\DataInstance
      */
     public function __construct(array $row)
     {
-        $this->DBRow = $row;
+        $this->_candid     = $row['Candidate'] ?? null;
```
johnsaigle: Could maybe use the CandID object here.

modules/api/php/projectimagesrow.class.inc
```diff
@@ -66,7 +96,7 @@ class ProjectImagesRow implements \LORIS\Data\DataInstance
      */
     public function getCenterID(): int
     {
-        return $this->DBRow['CenterID'];
+        return intval($this->_centerid);
```
johnsaigle: Just FYI `intval()` will return 0 when the input is null. This might not matter (and I'm not sure how you handle invalid values elsewhere) but this might be important info.

modules/api/php/projectinstrumentsrow.class.inc
```diff
@@ -0,0 +1,115 @@
...
+    {
+        $shortname      = $row['shortname'] ?? null;
+        $ddeinstruments = array_keys(\Utility::getAllDDEInstruments());
+
+        $this->_shortname    = $shortname;
+        $this->_fullname     = $row['fullname'] ?? null;
+        $this->_subgroupname = $row['subgroupname'] ?? null;
+        $this->_isDDE        = in_array(
+            $shortname,
+            $ddeinstruments
```
johnsaigle:
```suggestion
            $ddeinstruments,
            true
```

In general it's better for `in_array` to have the true parameter or else weird things can happen.

modules/api/php/projectinstrumentsrowprovisioner.class.inc
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Returns an instance of a ProjectInstrumentsRow object for a given
+     * table row.
+     *
+     * @param array $row The database row from the LORIS Database class.
+     *
+     * @return \LORIS\Data\DataInstance An instance representing this row.
+     */
+    public function getInstance($row) : \LORIS\Data\DataInstance
```
johnsaigle:
```suggestion
    public function getInstance(array $row) : \LORIS\Data\DataInstance
```

modules/api/php/visitdicomsrowprovisioner.class.inc
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Returns an instance of a DicomTar object for a given
+     * table row.
+     *
+     * @param array $row The database row from the LORIS Database class.
+     *
+     * @return \LORIS\Data\DataInstance An instance representing this row.
+     */
+    public function getInstance($row) : \LORIS\Data\DataInstance
```
johnsaigle:
```suggestion
    public function getInstance(array $row) : \LORIS\Data\DataInstance
```

I won't keep commenting this but it would be good to have wherever relevant.

modules/api/php/visitdicomsrowprovisioner.class.inc
```diff
@@ -0,0 +1,73 @@
...
+    {
+        return (new \LORIS\DicomTar(intval($row['tarchiveid'])))
```
johnsaigle: Again this good return 0 on a null input -- might matter, might not.

modules/api/test/login_Test.php
```diff
@@ -0,0 +1,92 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

php/libraries/InstrumentFlags.class.inc
```diff
@@ -0,0 +1,55 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

src/Api/Endpoints/Candidate/Candidate.php
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * Versions of the LORIS API which are supported by this
+     * endpoint.
+     *
+     * @return array a list of supported API versions.
+     */
+    protected function supportedVersions() : array
+    {
+        return array(
+                'v0.0.1',
```
johnsaigle: Is this still supported? I thought it got removed. I could very well be wrong though

src/Api/Endpoints/Candidate/Visit/Image/Qc.php
```diff
@@ -0,0 +1,207 @@
...
+        } catch (\DomainException $e) {
+            return new \LORIS\Http\Response\BadRequest(
+                $e->getMessage()
+            );
+        } catch (\Throwable $e) {
+            return  new \LORIS\Http\Response\BadRequest();
+        }
+
+        return (new \LORIS\Http\Response())
+            ->withStatus(204);
```
johnsaigle: Is it worth making a custom LORIS response for this? Not in this PR but as an issue?

src/Api/Endpoints/Candidate/Visit/Qc.php
```diff
@@ -0,0 +1,206 @@
...
+                           'Pending',
+                          );
+        $diff           = array_diff($requiredfields, array_keys($data));
+        if (!empty($diff)) {
+            return new \LORIS\Http\Response\BadRequest(
+                'Field(s) missing: ' . implode(', ', $diff)
+            );
+        }
+
+        $candid = $data['Meta']['CandID'] ?? null;
```
johnsaigle: Maybe use CandID object?

src/Api/Endpoints/Candidate/Visit/Visit.php
```diff
@@ -0,0 +1,285 @@
...
+            if ($timepoint->getCurrentStage() !== 'Not Started') {
+                return new \LORIS\Http\Response\Conflict(
+                    'This visit is already started'
+                );
+            }
+
+            $username = $user->getUsername();
+            $today    = date("Y-m-d");
+            // TODO :: Add a replace function in \Timepoint.class.inc
+            $timepoint->setData(array('CenterID' => $centerid));
```
johnsaigle: I think this can all be done with one function call to `setData()`

src/Api/Endpoints/Candidates.php
```diff
@@ -0,0 +1,234 @@
...
+        $candidate = \NDB_Factory::singleton()->candidate($candid);
+        $candidate->setData(
+            array('ProjectID' => $project->getId())
+        );
+
+        $body = (new \LORIS\Api\Views\Candidate($candidate))->toArray()['Meta'];
+        $link = '/'. $request->getURI()->getPath() . '/' . $candid;
+
+        return (new \LORIS\Http\Response\JsonResponse($body))
+            ->withStatus(201)
```
johnsaigle: I think there now exists a `\LORIS\Http\Response\Created`

src/Api/Endpoints/Projects.php
```diff
@@ -82,78 +81,71 @@ protected function supportedVersions() : array
      */
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
-        // FIXME: Validate permissions.
+        $user = $request->getAttribute('user');
+        if (!$this->_hasAccess($user)) {
+            return new \LORIS\Http\Response\Unauthorized();
```
johnsaigle: I think this should be Forbidden, not Unauthorized

src/Api/Endpoints/Projects.php
```diff
@@ -82,78 +81,71 @@ protected function supportedVersions() : array
...
         $request   = $request->withAttribute('pathparts', $pathparts);
 
         return $endpoint->process($request, $endpoint);
     }
 
     /**
-     * Returns an array of projects for this LORIS instance
-     * a format that can be JSON encoded to confirm to the
-     * API.
+     * Returns a loist of projects for this LORIS instance.
```
johnsaigle:
```suggestion
     * Returns a list of projects for this LORIS instance.
```

modules/api/php/candidatesprovisioner.class.inc
xlecours: Can we save this for an other PR?
`\LORIS\Data\Provisioners\DBRowProvisioner` has ` public abstract function getInstance($row) : DataInstance;` 

We should change all classes that extend `DBRowProvisioner` in the same PR.

modules/api/php/projectimagesrow.class.inc
xlecours: This is coming from the database and I don't think instantiating a CandID is useful here.

modules/api/php/projectimagesrow.class.inc
xlecours: I did not know. Thanks.
I this case, getCenterID must return an `int` so `0` is convenient for garbage data. Until someone creates a PSC with ID = 0

modules/api/php/projectinstrumentsrowprovisioner.class.inc
xlecours: Can we save this for an other PR?
\LORIS\Data\Provisioners\DBRowProvisioner has public abstract function getInstance($row) : DataInstance;

We should change all classes that extend DBRowProvisioner in the same PR.

modules/api/php/visitdicomsrowprovisioner.class.inc
xlecours: Can we save this for an other PR?
\LORIS\Data\Provisioners\DBRowProvisioner has public abstract function getInstance($row) : DataInstance;

We should change all classes that extend DBRowProvisioner in the same PR.

modules/api/php/visitdicomsrowprovisioner.class.inc
xlecours: tarchive.TarchiveID is an auto_increment column so this should be fine.
But if you have a suggestion to make this robust it would be welcome.

src/Api/Endpoints/Candidate/Candidate.php
xlecours: You are right, I will remove it.

src/Api/Endpoints/Candidate/Visit/Image/Qc.php
xlecours: Yes, 204 should have an empty body

src/Api/Endpoints/Candidate/Visit/Qc.php
xlecours: Yes, but not in this PR. TimePoint::getCandID still returns a integer.

src/Api/Endpoints/Candidate/Visit/Visit.php
xlecours: What can be done?

src/Api/Endpoints/Candidates.php
xlecours: nop

src/Api/Endpoints/Projects.php
xlecours: Debatable. In this case, `_hasAccess` return false when the user is not logged in. Forbidden response are usually sent when the user is authenticated but can't access the resource/service.

modules/api/php/projectimagesrowprovisioner.class.inc
```diff
@@ -81,6 +80,47 @@ class ProjectImagesRowProvisioner extends DBRowProvisioner
...
+                    new \LORIS\Data\Filters\UserSiteMatch()
+                );
+            } else {
+                $new = $new->filter(
+                    new \LORIS\Data\Filters\OnlyPhantoms()
+                );
+            }
+        }
+
+        // If a can't view phantoms from all sites, check if they can view phantoms
```
PapillonMcGill:
```suggestion
        // If a user can't view phantoms from all sites, check if they can view phantoms
```

modules/api/test/login_Test.php
```diff
@@ -0,0 +1,92 @@
...
+     */
+    public function setUp()
+    {
+        $this->_request = new ServerRequest();
+
+        $this->_authenticator = $this->createMock('\SinglePointLogin');
+    }
+
+    /**
+     * This checks if, given correct credentials, a HTTP 200 response
```
PapillonMcGill:
```suggestion
     * This checks if, given correct credentials, an HTTP 200 response
```

src/Api/Endpoints/Candidate/Visit/Qc.php
johnsaigle: This PR: https://github.com/aces/Loris/pull/4350

😄

src/Api/Endpoints/Candidate/Visit/Visit.php
johnsaigle: Instead of 
```php
$timepoint->setData(array('key' => 'value'));
$timepoint->setData(array('key2' => 'value2'));
$timepoint->setData(array('key3' => 'value2'));
...
```

Do 
```php
$timepoint->setData(
    array(
        'key' => 'value',
        'key2' => 'value2',
        'key3' => 'value3',
        ...
    )
);

Comment (johnsaigle) [reply src/Api/Endpoints/Candidate/Visit/Visit.php]: Less overhead with function calls and round-trips to the DB. Might not make a difference, but who knows

File: src/Api/Endpoints/Candidates.php (hunk 67 above)
Comment (johnsaigle) [reply src/Api/Endpoints/Candidates.php]: Oops I thought my PR had been merged...
https://github.com/aces/Loris/pull/4949

File: src/Api/Endpoints/Projects.php (hunk 68 above)
Comment (johnsaigle) [reply src/Api/Endpoints/Projects.php]: Got it, I didn't realize that this can the case for an unauthenticated user. You're correct then and 401 is appropriate

File: src/Api/Endpoints/Candidate/Visit/Image/Format/Raw.php [hunk 72]
```diff
@@ -0,0 +1,162 @@
...
+            error_log('file in database but not in file system');
+            return new \LORIS\Http\Response\NotFound();
+        }
+
+        if (!$info->isReadable()) {
+            error_log('file exist but is not readable by webserver');
+            return new \LORIS\Http\Response\NotFound();
+        }
+
+        if (!`which minctoraw`) {
```
cmadjar: I always get 'minctoraw not installed' here. Does it work for you?

I may be wrong but I doubt apache is sourcing the MINC tool kit `.sh` file therefore, `minctoraw` is never seen by apache and `which minctoraw` will never appear to be installed and therefore always return 'minctoraw not installed'. Am I missing something?

FYI, the MINC tools path is stored in `config.xml` under the `<MINCToolsPath>` tag. Path can be for example: `/opt/minc/`. The full path to minctoraw would then be `/opt/minc/bin/minctoraw` (a.k.a. `<MINCToolsPath>/bin/minctoraw`).

xlecours: I have the following lines in /etc/apache2/envvars
```
# Add minctoolkit to apache user environement 
export PATH=$PATH:/opt/minc/1.9.16/bin
```

xlecours: The line `which minctoraw` is from the old API code. I thought it was working at some point so I added the lines in envvars.

cmadjar: mmm. I added the line in `envvars`. Is it documented somewhere that I missed at the time of install that this need to be added to the `envvars` file of apache? Wouldn't it be better to use the config setting in the `config.xml` file? 

Looks like BrainBrowser is using the config <MINCToolsPath> so maybe it is best to do the same for the API? What do you think?

Also @driusan, @johnsaigle , is there a reason why this setting MINCToolsPath is in the `config.xml` file and not in the frontend? Maybe a security reason? Just wondering.

xlecours: @cmadjar I will change it to use de config.xml
And I second your proposition of moving that in the database.

johnsaigle: Yes it should be in the DB. Not sure of the origin of why it's in that file. https://github.com/aces/Loris/pull/4023

cmadjar: Maybe that path was forgotten when all the paths were moved to the Config module. I created an issue (#5019) for that. Should not be complicated to move it there. ;)

xlecours:
@cmadjar Ready for Re-re-review :)

cmadjar:
@xavier I retested your PR. Sorry again for the delay...

- is the `header/raw` endpoint returning an actual `Value` in the response for you? For me, the `Value` is empty (`{"Meta":{"CandID":400300,"Visit":"V1","File":"demo_400300_V1_t1_001.mnc","Header":"raw"},"Value":""}`)

- it looks like I can't download DICOM archives anymore from the API. It keeps telling me not found but it is in the filesystem

- your branch needs to be rebased unfortunately.

xlecours:
@cmadjar Regarding the last point, it's @ridz1208 fault but it was expected :)
Thanks. 

I will look into the 2 other points

xlecours:
@cmadjar Regarding point 1, sending a get request to `/candidates/$candid/$visit/images/$filename/headers/raw` is asking for the specific header called `raw`. I think you wish to get the `full` headers instead. The `full` keyword is a special case that supersede any specific header.

xlecours:
@cmadjar For point 2, does the filename appears in `/dicoms`

cmadjar:
@xlecours 

For point 1: ooopsie... Nevermind. I should have replaced `header` by `format` when testing. I wanted to test the download of the raw format but did not use the right endpoint 😞...

For point 2: the filename do appear in the `/dicoms` endpoint. Then trying this endpoint `https://cecile-dev.loris.ca/api/v0.0.3-dev/candidates/400300/V1/dicoms/DCM_2016-08-19_ImagingUpload-18-41-C4ZlTl.tar` I get the following:
Screen Shot 2019-09-20 at 12 29 31 PM

Not sure why...

src/Api/Endpoints/Candidate/Visit/Image/Format/Raw.php
```diff
@@ -0,0 +1,172 @@
...
+            );
+        }
+
+        $filename = $info->getFilename();
+        $fullpath = escapeshellarg($info->getPathname());
+
+        ob_start();
+        passthru("${mincpath}/bin/minctoraw -byte -unsigned -normalize $fullpath");
+        $content = ob_get_contents();
+        ob_end_clean();
```
driusan: This would ideally be done using some kind of (theoretical) ExecStream or ProgramStream or something so that it could be streamed as it's running instead of holding the entire contents of the output in memory as a string, but that's not a blocker since it's far from the only inefficient place in LORIS right now..

src/Api/Endpoints/Candidate/Visit/Image/Format/Thumbnail.php
```diff
@@ -0,0 +1,160 @@
...
+            return new \LORIS\Http\Response\NotFound();
+        }
+
+        $file     = $info->openFile('r');
+        $filename = $info->getFilename();
+
+        ob_start();
+        $file->fpassthru();
+        $content = ob_get_contents();
+        ob_end_clean();
```
driusan: Why doesn't this just use `\LORIS\Http\FileStream` to avoid needing to read the whole file into memory?

src/Api/Endpoints/Candidate/Visit/Image/Image.php
```diff
@@ -0,0 +1,215 @@
...
+            error_log('file exist but is not readable by webserver');
+            return new \LORIS\Http\Response\NotFound();
+        }
+
+        $file = $info->openFile('r');
+
+        ob_start();
+        $file->fpassthru();
+        $content = ob_get_contents();
+        ob_end_clean();
```
driusan: Same comment about FileStream

modules/api/php/candidatesprovisioner.class.inc
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /**
+     * Creates a new providionner with user specific filter.
+     *
+     * @param \User $user The requesting user
+     *
+     * @return CandidatesProvisioner
+     */
+    public function forUser(\User $user): CandidatesProvisioner
```
driusan: I'm not sure about this factory being the right way to do this, vs just unconditionally adding a filter and doing the permission check inside of the (new) filter, since the filter interface is already passed the User object and it seems to be a more appropriate place to handle the concern.

modules/api/php/candidatesprovisioner.class.inc
```diff
@@ -58,7 +58,7 @@ class CandidatesProvisioner extends \LORIS\Data\Provisioners\DBRowProvisioner
     }
 
     /**
-     * Creates a new providionner with user specific filter.
+     * Creates a new provisionner with user specific filter.
```
maltheism: Should it be provisionner or provisioner?

modules/api/php/visitdicomsrowprovisioner.class.inc
```diff
@@ -0,0 +1,73 @@
...
+            '
+             SELECT
+               t.TarchiveID as tarchiveid
+             FROM tarchive t
+             JOIN session s
+               ON (s.ID=t.SessionID)
+             JOIN candidate c
+               ON (s.CandID=c.CandID)
+             WHERE 
+               c.Active="Y" AND
```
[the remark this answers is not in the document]
xlecours: From doc:

> A string is a sequence of bytes or characters, enclosed within either single quote (') or double quote (") characters.

> If the ANSI_QUOTES SQL mode is enabled, string literals can be quoted only within single quotation marks because a string quoted within double quotation marks is interpreted as an identifier. 

I will make a pass to change them.

src/Api/Endpoints/Candidate/Visit/Image/Format/Raw.php
xlecours: @driusan Would you care drafting something? I would be glad to add it.

xlecours: Maybe `\Zend\Diactoros\CallbackStream.php`

modules/api/php/candidatesprovisioner.class.inc
xlecours: > doing the permission check inside of the (new) filter

 Would that mean to check permission for each row?

Originally, I wanted to pass the User in the constructor but I was shy to change `DBRowProvisioner` constructor signature. Having a provisioner know which filter to apply seemed like a good segregation of concerns as oppose to asking the user to add arbitrary filters. 

_On a side note_, I think that the filter function should be chnaged so that it adds expressions to the `WHERE` clause of the query. Databases are usually good at filtering rows, and it would reduce memory usage by reducing the result size.

driusan: > Would that mean to check permission for each row?

Yes, but the permissions are currently loaded by the User constructor. Checking permissions is an O(1) array lookup, not a database call.

> On a side note, I think that the filter function should be chnaged so that it adds expressions to the WHERE clause of the query. Databases are usually good at filtering rows, and it would reduce memory usage by reducing the result size.

It's unlikely to have a significant affect on memory usage since DBRowProvisioner uses an iterator which lazily converts to the respective DataInstance class on iteration. The memory overhead should be approximately the size of 1 row, plus a little overhead for garbage collection regardless of result set size. (It's the main reason it's more efficient than $DB->pselect, which returns an array that must be fully loaded into memory.) If memory has went up to higher than that, it's a regression that should be fixed.

modules/api/php/endpoints/candidate/candidate.class.inc
```diff
@@ -10,11 +10,11 @@
  * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
  * @link     https://github.com/aces/Loris
  */
-namespace LORIS\Api\Endpoints\Candidate;
+namespace LORIS\api\Endpoints\Candidate;
```
maltheism: Hey @xlecours wouldn't it make more sense if api was in all caps?
`namespace LORIS\API\Endpoints\Candidate`

xlecours: all our modules name are lower case. I agree that it's an acronyme like LORIS and should probably be all caps but I will stick with the module name convention.

xlecours:
@driusan Change requests adressed

cmadjar:
@xlecours needs rebase

@driusan could you re-review please?

PR #4252, opened 2019-01-11

HenriRabalais:
### Summary of changes
Added a Modal Window and Triggerable Modal Window Component to LORIS.

### Why?

### Operation
- Modal Window can be used as standalone component which is passed an `open` prop that dictates whether or not it is displayed.
- A Triggerable Modal Window can be used to extradite the responsibility of controlling the display of the modal window. A TriggerableModal will render a trigger button that, when clicked, will cause the Modal Window to be displayed.

### Features

### Testing
- Go to the Media module menu filter. Click the Edit button on any of the datatable rows.


HenriRabalais:
Having trouble with removing random committed files, so I'm closing this and opening a new one.

PR #4253, opened 2019-01-11

HenriRabalais:
### Summary of changes
Added a Modal Window and Triggerable Modal Window Component to LORIS.

### Why?

### Operation
-Modal Window can be used as standalone component which is passed an open prop that dictates whether or not it is displayed.
-A Triggerable Modal Window can be used to extradite the responsibility of controlling the display of the modal window. A TriggerableModal will render a trigger button that, when clicked, will cause the Modal Window to be displayed.

### Features

### Testing
- Go to the Media module menu filter. Click the Edit button on any of the datatable rows.


jsx/Form.js
```diff
@@ -1525,6 +1525,37 @@ ButtonElement.defaultProps = {
   },
 };
 
+/**
+  * Call To Action (CTA) component
+  * React wrapper for <button> element that is used for Call to Actions, usually
+  * outside the context of forms.
+  */
+ class CTA extends Component {
```
davidblader: is there any plan to deprecate the ButtonElement component in favor of this?

HenriRabalais: No, although the Button Element may end up using the CTA component. I'm not sure if this is the right file for the CTA, but we need some sort of button that doesn't conform to the `Form` formatting so that it can be used in table cells, above datatables, etc.

jsx/Modal.js
```diff
@@ -0,0 +1,162 @@
...
+    super();
+    this.handleClose = this.handleClose.bind(this);
+  }
+
+  handleClose() {
+    if (this.props.throwWarning) {
+      swal({
+        title: 'Are You Sure?',
+        text: 'Leaving the form will result in the loss of any information' +
+          'entered.',
```
zaliqarosli: a space is missing again between information and entered

jsx/Modal.js
```diff
@@ -0,0 +1,162 @@
...
+        type: 'warning',
+        showCancelButton: true,
+        confirmButtonText: 'Proceed',
+        cancelButtonText: 'Cancel',
+      }).then((result) => result.value && this.props.close());
+    } else {
+      this.props.close();
```
zaliqarosli: could we name this prop onClose() so it looks more like a function when the prop needs to be defined on calling the component?

jsx/Modal.js
```diff
@@ -0,0 +1,162 @@
...
+    }
+  }
+
+  render() {
+    const {open, children, onSubmit, title} = this.props;
```
zaliqarosli: is open essentially show?

jsx/Modal.js
```diff
@@ -0,0 +1,162 @@
...
+          </div>
+        </div>
+      </div>
+    );
+  }
+}
+
+Modal.propTypes = {
+  title: PropTypes.string,
+  trigger: PropTypes.element.isRequired,
```
zaliqarosli: i think trigger isn't here anymore but open and close can be added

zaliqarosli:
on testing, i'm getting errors in my console for the TriggerableModal label prop and that this.props.onClose is not a function

jsx/Modal.js
HenriRabalais: Yeah. If it makes more sense, I can set the `state` in `TriggerableModal` to be called `open`, and the `props` with `Modal` to be called `show`.

jsx/Modal.js
HenriRabalais: sure.

HenriRabalais:
@zaliqarosli should all be fixed!

jsx/Modal.js
```diff
@@ -150,8 +155,9 @@ class Modal extends Component {
 
 Modal.propTypes = {
   title: PropTypes.string,
-  trigger: PropTypes.element.isRequired,
   onSubmit: PropTypes.object,
+  onClose: PropTypes.func,
+  show: PropTypes.bool,
```
zaliqarosli: would show and onClose have to be isRequired?

zaliqarosli:
on testing, I was thinking we should call fetch after modifications to the edit form is submitted, and then i received this error `Warning: Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in MediaEditForm (created by DataTable)`. Is this maybe connected?

HenriRabalais:
@zaliqarosli I'm not entirely sure what you mean

HenriRabalais:
@zaliqarosli I've passed `fetchData` down to the `MediaEditForm` as prop so that upon succesful submission, the page data will be fetched again. Eventually, this function would ideally be passed to the `onSubmit` prop, but this will necessitate significant refactoring of the `MediaEditForm` component. As for the warning, I'm really not sure what is causing it and will look into it later this week.

HenriRabalais:
@zaliqarosli Addressed.

PR #4254, opened 2019-01-11
 Docker test in Travis with PHP 7.2 and 7.3

kongtiaowang:
Running Docker test in Travis.
Saving time
easy to test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,15 +11,9 @@ services:
 - docker
 
 install:
-- sudo apt-get update
-- sudo apt-get install -y libapparmor1
-- pecl install -f ast-0.1.6
-- sudo apt-get install npm
-- npm install
-- composer install
-- npm run compile
+- make dev-install
 before_script:
-- if [[ ${TRAVIS_PHP_VERSION:0:3} == "7.3" ]]; then sed -i 's/7.2/7.3/g' Dockerfile.test.php7; fi
+- sed -i 's/php7.2/php${TRAVIS_PHP_VERSION:0:3}/g' Dockerfile.test.php7
 script:
 - make checkstatic
 - docker-compose run -T --rm integration-tests vendor/bin/phpunit --configuration test/phpunit.xml
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,7 @@ language: php
 matrix:
   include:
   - php: "7.2"
-    env: PHP='7.2'
   - php: "7.3"
-    env: PHP='7.3'
 
 services:
 - docker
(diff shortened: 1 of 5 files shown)


johnsaigle:
Test speed cut in half.... pretty nice!

.travis.yml
```diff
@@ -1,99 +1,25 @@
...
-  - php: "7.3"
-    env: ALLOW_FAILURE='true'
-  allow_failures:
-  - php: "7.3"
-    env: ALLOW_FAILURE='true'
+# any language should do it
+language: php
 
-addons:
-    firefox: "44.0"
```
driusan: What browser/version does this test with now?

.travis.yml
```diff
@@ -1,99 +1,25 @@
...
-    # screen in the middle of our other tests
-    - DISPLAY=:99.0 java -jar selenium-server-standalone-2.52.0.jar  2>1 > /dev/null &
-
+- sudo apt-get update
+- sudo apt-get install -y libapparmor1
+- pecl install -f ast-0.1.6
+- sudo apt-get install npm
+- npm install
+- composer install
+- npm run compile
```
driusan: I think these 3 lines can just be replaced with `make dev`

.travis.yml
```diff
@@ -1,99 +1,25 @@
...
-    - mysql LorisTest -e "INSERT INTO test_battery (Test_name, AgeMinDays, AgeMaxDays, Active, Stage, SubprojectID, Visit_label, CenterID) VALUES ('testtest', '1', '99999', 'Y', 'V1', '1', '2', NULL)"
-    - cd tools
-    - find ../project/instruments/NDB_BVL_Instrument_testtest.class.inc | php lorisform_parser.php
-    - find ../project/instruments/NDB_BVL_Instrument_testtest.class.inc | php generate_tables_sql.php
-    - cd ..
-    - mysql LorisTest < project/tables_sql/testtest.sql
-#env:
- # - "LORIS_DB_CONFIG=test/config.xml"
-
+- if [[ ${TRAVIS_PHP_VERSION:0:3} == "7.3" ]]; then sed -i 's/7.2/7.3/g' Dockerfile.test.php7; fi
```
driusan: Can this be changed so that instead of hardcoding 7.3, it uses the environment variable in the sed command? If that's done, I don't think the if statement is required anymore, since replacing 7.2 with 7.2 doesn't hurt anything and we would still just need to maintain the PHP versions in the normal place in the .travis.yml instead of needing to remember to update this.

.travis.yml
kongtiaowang: firefox 47.0.1
I will update the version later, not in this PR.

.travis.yml
kongtiaowang: We can't use Travis variable in the sed command.

kongtiaowang: done.

driusan:
@kongtiaowang try "double quotes" instead of 'single quotes'

driusan:
🎉

PR #4255, opened 2019-01-14
[Do not merge] Project permissions

Jkat:
### Brief summary of changes

- [x] `user_project_rel` table (Discuss: do we want a 'default' project? otherwise can remove https://github.com/aces/Loris/blob/32786ce8cfbe553188988e273eafd450c4ae0a10/SQL/0000-00-00-schema.sql#L258)
- [x] new Config option "Use Project Permissions" 
- [x] un-reactified(?) modifications to User Accounts module to administer project access per user (multiselect element just like Site selector)
- [ ] php/libraries/Project.class.inc customizations to be be discussed - functions added to Project class are only to support Dashboard's recruitment counts
- [ ] example of a module using the permission e.g. candidate list.   CAP implemented user-project permission checks in each module to filter data

Many projects can be assigned to one user. 

( To do: )
- [ ] Add SQL patch  --> depends on outstanding question:  Should all users automatically get access to all existing projects?  That way no users lose access to data, but the onus would be on the admin to **remove** user access to projects they should not access.  
The conservative alternative is: don't assign any users to any projects, and make the admin manually **add** user access to projects (infeasible? could take a while.)  
Also -- Will/should there be a default Project (ProjectID=1) in every LORIS 
- [ ] Ensure Release Notes tells all admins to Manually assign Each users to Projects via User Accounts ASAP.  Big Caveat because this will lock all users out of all data without admin's manual intervention. 

(Table originally created by @jacobpenny for CAP project, and modelled on Mouna's `user_psc_rel` table)

### DIFF
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -253,7 +253,7 @@ CREATE TABLE `user_project_rel` (
   KEY `FK_user_project_rel_2` (`ProjectID`),
   CONSTRAINT `FK_user_project_rel_2` FOREIGN KEY (`ProjectID`) REFERENCES `Project` (`ProjectID`) ON DELETE CASCADE ON UPDATE CASCADE,
   CONSTRAINT `FK_user_project_rel_1` FOREIGN KEY (`UserID`) REFERENCES `users` (`ID`) ON DELETE CASCADE ON UPDATE CASCADE
-) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
 
 INSERT INTO user_project_rel (UserID, ProjectID) SELECT 1, ProjectID FROM Project;
 
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -246,6 +246,17 @@ CREATE TABLE `users` (
 INSERT INTO `users` (ID,UserID,Real_name,First_name,Last_name,Email,Privilege,PSCPI,DBAccess,Active,Pending_approval,Password_expiry)
 VALUES (1,'admin','Admin account','Admin','account','admin@example.com',0,'N','','Y','N','2016-03-30');
 
+CREATE TABLE `user_project_rel` (
+  `UserID` int(10) unsigned NOT NULL,
+  `ProjectID` int(2) NOT NULL,
+  PRIMARY KEY  (`UserID`,`ProjectID`),
+  KEY `FK_user_project_rel_2` (`ProjectID`),
+  CONSTRAINT `FK_user_project_rel_2` FOREIGN KEY (`ProjectID`) REFERENCES `Project` (`ProjectID`) ON DELETE CASCADE ON UPDATE CASCADE,
+  CONSTRAINT `FK_user_project_rel_1` FOREIGN KEY (`UserID`) REFERENCES `users` (`ID`) ON DELETE CASCADE ON UPDATE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+INSERT INTO user_project_rel (UserID, ProjectID) SELECT 1, ProjectID FROM Project;
+
 CREATE TABLE `user_psc_rel` (
   `UserID` int(10) unsigned NOT NULL,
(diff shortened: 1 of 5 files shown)


SQL/0000-00-00-schema.sql
```diff
@@ -246,6 +246,17 @@ CREATE TABLE `users` (
...
 VALUES (1,'admin','Admin account','Admin','account','admin@example.com',0,'N','','Y','N','2016-03-30');
 
+CREATE TABLE `user_project_rel` (
+  `UserID` int(10) unsigned NOT NULL,
+  `ProjectID` int(2) NOT NULL,
+  PRIMARY KEY  (`UserID`,`ProjectID`),
+  KEY `FK_user_project_rel_2` (`ProjectID`),
+  CONSTRAINT `FK_user_project_rel_2` FOREIGN KEY (`ProjectID`) REFERENCES `Project` (`ProjectID`) ON DELETE CASCADE ON UPDATE CASCADE,
+  CONSTRAINT `FK_user_project_rel_1` FOREIGN KEY (`UserID`) REFERENCES `users` (`ID`) ON DELETE CASCADE ON UPDATE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
driusan:
```suggestion
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```

SQL/0000-00-00-schema.sql
```diff
@@ -246,6 +246,17 @@ CREATE TABLE `users` (
...
+
+INSERT INTO user_project_rel (UserID, ProjectID) SELECT 1, ProjectID FROM Project;
```
driusan: Why does User 1 get all projects but no other users get any projects by default?

Jkat: https://github.com/aces/Loris/pull/4255/files#diff-49c955f1af7f1ad35cdd7a9aceb590d1R270
^ Related: Why does User 1 get all centers but no other users get any centers by default?

Possible answer: there's only User 1 upon sourcing the schema, and we can't know who are the additional users, and also this line will probably do nothing(at least so far) until we come up with a possible default project to insert...

driusan: Ah.. I was thinking this was the patch for existing projects, not the default schema. It makes sense for the default schema.

christinerogers:
To do : 
[ ] Add patch

SQL/0000-00-03-ConfigTables.sql
```diff
@@ -69,6 +69,7 @@ INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType,
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'ImagingUploaderAutoLaunch', "Allows running the ImagingUpload pre-processing scripts", 1, 0, 'boolean', ID, 'ImagingUploader Auto Launch',23 FROM ConfigSettings WHERE Name="study";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'citation_policy', 'Citation Policy for Acknowledgements module', 1, 0, 'textarea', ID, 'Citation Policy', 24 FROM ConfigSettings WHERE Name="study";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'CSPAdditionalHeaders', 'Extensions to the Content-security policy allow only for self-hosted content', 1, 0, 'text', ID, 'Content-Security Extensions', 25 FROM ConfigSettings WHERE Name="study";
+INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'useProjectPermissions', "Enable project level permissions", 1, 0, 'boolean', ID, 'Use project-level permissions', 26 FROM ConfigSettings WHERE Name="study";
```
xlecours: Could we not use an additional config?

cmadjar:
@ridz1208 Should this PR be closed since you are revamping the projects?

ridz1208:
@cmadjar not yet, once I cover it in one of my PRs i'll close it

PR #4256, opened 2019-01-14
[API] Documentation for DICOM upload and processing using the API

bopucky:
### Documentation for DICOM upload and processing using the API.

Documenting the proposed API endpoints for enabling DICOM file upload and to trigger the Imaging pipeline to process the file.

### This resolves issue...

Currently no known associated issue

### To test this change...

- Read and review


docs/API/LorisRESTAPI_v0.0.3.md
```diff
@@ -814,3 +814,49 @@ files, and a `.tar.gz` of the raw DICOM data as acquired during the candidate
...
 Only `GET` is currently supported.
+
+### 5.3 Upload DICOM file
+```
+PUT /candidates/$CandID/$VisitLabel/dicoms/$Filename/$IsPhantom
+```
+
+Upload a DICOM file that is compressed as a `.zip`,`.tar.gz` or `.tgz`.
+$CandID must be the full name of the file and its extension.
+$IsPhantom must be `'Y'` or `'y'` if it's a phantom file, else `'N'` or `'n'`
```
christinerogers: Can the single quotes be removed? 
(the pipeline input file doesn't require single quotes, not sure why this would)
i.e. Leave the ticks, changing:  `'Y'` to `Y` for all instances

xlecours: Could the isPhantom be derived from the entity_type of the candidate? Usually, when the candidate entity_type is `scanner` the images are phantoms. @cmadjar right?

docs/API/LorisRESTAPI_v0.0.3.md
```diff
@@ -814,3 +814,49 @@ files, and a `.tar.gz` of the raw DICOM data as acquired during the candidate
...
+Upload a DICOM file that is compressed as a `.zip`,`.tar.gz` or `.tgz`.
+$CandID must be the full name of the file and its extension.
+$IsPhantom must be `'Y'` or `'y'` if it's a phantom file, else `'N'` or `'n'` 
+if not.
+
+Returns JSON data upon success having the form:
+
+```js
+{
+    "success":"Uploaded",
```
xlecours: mri_upload_id would be enough

docs/API/LorisRESTAPI_v0.0.3.md
```diff
@@ -814,3 +814,49 @@ files, and a `.tar.gz` of the raw DICOM data as acquired during the candidate
 scanning session, and as retrieved from `/candidates/$CandID/$Visit/dicoms`.
 
 Only `GET` is currently supported.
+
+### 5.3 Upload DICOM file
+```
+PUT /candidates/$CandID/$VisitLabel/dicoms/$Filename/$IsPhantom
+```
+
+Upload a DICOM file that is compressed as a `.zip`,`.tar.gz` or `.tgz`.
```
christinerogers: Suggest changing "a DICOM file" to "a tarred set of DICOM files" 
Can/Should it be a directory?

docs/API/LorisRESTAPI_v0.0.3.md
```diff
@@ -814,3 +814,49 @@ files, and a `.tar.gz` of the raw DICOM data as acquired during the candidate
...
+```
+
+mri_upload_id is the Upload ID for the file.
+
+### 5.4 Process DICOM file
+```
+POST /candidates/$CandID/$VisitLabel/dicoms/$Filename/$IsPhantom
+```
+
+Launch processing of an uploaded DICOM file that has not been unprocessed.
```
christinerogers: Change "unprocessed" -> "processed" ?
Change "DICOM file" to "DICOM fileset"

docs/API/LorisRESTAPI_v0.0.3.md
```diff
@@ -814,3 +814,49 @@ files, and a `.tar.gz` of the raw DICOM data as acquired during the candidate
...
+```js
+{
+    "success":"Uploaded",
+    "mri_upload_id":"3"
+}
+```
+
+mri_upload_id is the Upload ID for the file.
+
+### 5.4 Process DICOM file
```
christinerogers: "file" -> "files" or even "DICOM fileset"

docs/API/LorisRESTAPI_v0.0.3.md
```diff
@@ -814,3 +814,49 @@ files, and a `.tar.gz` of the raw DICOM data as acquired during the candidate
...
+Returns JSON data upon success having the form:
+
+```js
+{
+    "success":"Uploaded",
+    "mri_upload_id":"3"
+}
+```
+
+mri_upload_id is the Upload ID for the file.
```
christinerogers: "file" -> fileset

docs/API/LorisRESTAPI_v0.0.3.md
```diff
@@ -814,3 +814,49 @@ files, and a `.tar.gz` of the raw DICOM data as acquired during the candidate
...
+
+### 5.4 Process DICOM file
+```
+POST /candidates/$CandID/$VisitLabel/dicoms/$Filename/$IsPhantom
```
xlecours: I suggest `POST /candidates/$CandID/$VisitLabel/dicoms/$Filename/processes` with a payload like 
```js
{
  "process_type": "mri_upload",
  "filename" : $filename
}
```
with a response like 
```js
{
  "process_id": 123,
  "status": "started|running|complete|failed",
  "message": "..."
}
```
Then, `GET /candidates/$CandID/$VisitLabel/dicoms/$Filename/processes/123`
would return 
```js
{
  "process_id": 123,
  "status": "started|running|complete|failed",
  "message": "..."
}
```

docs/API/LorisRESTAPI_v0.0.3.md
```diff
@@ -814,3 +814,49 @@ files, and a `.tar.gz` of the raw DICOM data as acquired during the candidate
 scanning session, and as retrieved from `/candidates/$CandID/$Visit/dicoms`.
 
 Only `GET` is currently supported.
+
+### 5.3 Upload DICOM file
+```
+PUT /candidates/$CandID/$VisitLabel/dicoms/$Filename/$IsPhantom
```
driusan: I don't think $IsPhantom belongs in the URL, it would make more sense to add it to a request header and have something like

```
PUT /candidates/$CandID/$VisitLabel/dicoms/$Filename`
X-Is-Phantom: 1
```

or perhaps as a GET parameter

```
PUT /candidates/$CandID/$VisitLabel/dicoms/$Filename?phantom=1
```

But moreover, where is the file expected to live after being uploaded? If it's always going to be retrieved at 
` /candidates/$CandID/$VisitLabel/dicoms/$Filename` after being uploaded, this makes sense (but this section of the documentation should be combined with section 5.2, I think) but if the filename is derived from the header/patient data it would make more sense to have this a `POST` request to `/candidates/$CandID/$VisitLabel/dicoms` since the filename is discarded anyways.

docs/API/LorisRESTAPI_v0.0.3.md
driusan: maybe "status": "uploaded" to leave a little flexibility for other responses?

docs/API/LorisRESTAPI_v0.0.3.md
```diff
@@ -814,3 +814,49 @@ files, and a `.tar.gz` of the raw DICOM data as acquired during the candidate
...
+```
+
+Upload a DICOM file that is compressed as a `.zip`,`.tar.gz` or `.tgz`.
+$CandID must be the full name of the file and its extension.
```
driusan: I assume this is a typo and you meant '$Filename' not "$CandID"?

docs/API/LorisRESTAPI_v0.0.3.md
driusan: This should also describe where the data goes after it's `PUT` to the server. Does it end up in the dicom archive? Or the imaging browser? Somewhere else?

docs/API/LorisRESTAPI_v0.0.3.md
```diff
@@ -814,3 +814,49 @@ files, and a `.tar.gz` of the raw DICOM data as acquired during the candidate
...
+
+Launch processing of an uploaded DICOM file that has not been unprocessed.
+
+The JSON must be of the form:
+
+```js
+{
+    "Filename":$Filename, 
+    "mri_upload_id":$mri_upload_id, 
+    "IsPhantom":$IsPhantom,
```
driusan:
```suggestion
    "IsPhantom": bool, 
```

docs/API/LorisRESTAPI_v0.0.3.md
```diff
@@ -814,3 +814,49 @@ files, and a `.tar.gz` of the raw DICOM data as acquired during the candidate
...
+    "CandID":$CandID,
+    "VisitLabel":$VisitLabel
+}
+```
+
+Returns JSON data with a status if processing has been launched in the form:
+
+```js
+{"success":"Process launched"}
+```
```
driusan: It should also return a response code of "202 Accepted" (and be documented as such..)

docs/API/LorisRESTAPI_v0.0.3.md
driusan: (Also, I think @xlecours's comment below is right and you can probably just derive whether or not it's a phantom from the CandID)

docs/API/LorisRESTAPI_v0.0.3.md
cmadjar: indeed, given that the CandID is in the URL, you technically could derive whether the scan is a phantom by looking at the "Entity_type" of the candidate (would be "Scanner" for phantom datasets or "Human" for actual candidates.

At least, this is the current behaviour. Hope this helps!

driusan: Is the value of 'Entity_type' 'Scanner' for human phantoms?

docs/API/LorisRESTAPI_v0.0.3.md
bopucky: The file is supposed to live in the location that is configured in Loris as given by 
MRI-Upload Directory which is specified Configuration->"Paths"

bopucky: That would be good to explicitly state in the documentation also, so that one knows where to look for uploaded files.

docs/API/LorisRESTAPI_v0.0.3.md
bopucky: A typo indeed!

docs/API/LorisRESTAPI_v0.0.3.md
bopucky: Yes indeed. I will add that. 

There are supposed to be two steps (based on what i thought was the existing Loris way of doing thngs)

1. Upload the file. It will be automatically deposited/saved in the location specified in Configuration->Paths->MRI-Upload Directory.

This will also insert the file details into the mri_upload table and return the Upload ID in the HTTP response.

If "ImagingUploader Auto Launch" is enabled, then the Imaging pipeline is automatically triggered. Otherwise nothing else will happen.

2. Trigger the Imaging pipeline (explicitly) by giving the Upload Id as well as the other details to process the files.

docs/API/LorisRESTAPI_v0.0.3.md
bopucky: Yes changing it to bool makes more sense. 

I initially had it as Y/y or N/n since the function isValid() in imaging_uploader.class.inc validates for Y/y or N/n

docs/API/LorisRESTAPI_v0.0.3.md
bopucky: Add it it as a request header is best I think so that the uploading the file happens only with the PUT request. 

So I will add that to the doc (along with an example) as well as update the current code to check for the X-Is-Phantom header.

docs/API/LorisRESTAPI_v0.0.3.md
bopucky: Great! I will make the changes.

The logic to get the status of the process will be added (since there i hadn't put in yet)

docs/API/LorisRESTAPI_v0.0.3.md
bopucky: Thanks Christine. I will make the changes

docs/API/LorisRESTAPI_v0.0.3.md
bopucky: :-) nice catch

docs/API/LorisRESTAPI_v0.0.3.md
bopucky: Regarding where the file is supposed to live, it will live in the location that is configured in Loris as given by MRI-Upload Directory which is specified Configuration->"Paths"

However I will state that explicitly in the documentation also so that one knows where to look for uploaded files.

Also adding IsPhantom as a request header in the PUT is best. This way uploading the DICOM fileset still happens only with the PUT request. Sticking IsPhantom in the url was feeling super kludgy but I couldn't think of something better at the time :-)

I will make the changes to the doc (along with giving an example) as well as update the current code to check for the X-Is-Phantom header.

docs/API/LorisRESTAPI_v0.0.3.md
bopucky: Will change. Thanks.

docs/API/LorisRESTAPI_v0.0.3.md
bopucky: Yes, I will remove the single quotes.  

Based on one of Dave's comments I will switch Y to 1 and N to 0 as well since $IsPhantom is now going to expect a boolean.

docs/API/LorisRESTAPI_v0.0.3.md
bopucky: Yes, will change.

driusan:
Replaced by #4262.

In the future, you can just incorporate feedback by adding commits to the branch and pushing them to GitHub. GitHub automatically updates the PR, you don't need to send a new PR with the changes.

PR #4257, opened 2019-01-14
[Candidate] - Fixed Sex Column Bug

HenriRabalais:
### Brief summary of changes
The Candidate class was still selecting for `Sex` even the column has been changed to `Gender`. This PR resolves this bug.


driusan:
Thanks.. must have been incorrectly resolved one of the many times I rebased the original PR.

driusan:
Wait, no, this is backwards. It was changed from 'Gender' to 'Sex', not the other way around. Why is this being changed back?

PR #4264, opened 2019-01-16
[Core: FeedbackMRI] Add return types to function signatures

johnsaigle:


php/libraries/FeedbackMRI.class.inc
```diff
@@ -57,34 +57,13 @@ class FeedbackMRI
         }
     }
 
-    /**
-     * Clears all predefined comments from the object
-     * (used prior to re-setting the list of predefined comments)
-     *
-     * @return bool true if operation succeeded
-     */
-    function clearPredefinedComments()
```
johnsaigle: this function isn't used anywhere.

php/libraries/FeedbackMRI.class.inc
```diff
@@ -57,34 +57,13 @@ class FeedbackMRI
...
-    {
-        // start the query
-        $query = "DELETE FROM feedback_mri_comments WHERE ".$this->_buildWhere();
-
-        // limit to predefined comments
-        $query .= "AND PredefinedCommentID IS NOT NULL";
-
-        // run the query
-        $success = mysql_query($query);
```
johnsaigle: This function has not existed in PHP since before 7.0.0 and LORIS has never complained so... I think it's safe to get rid of this function altogether.

php/libraries/FeedbackMRI.class.inc
```diff
@@ -405,7 +384,7 @@ class FeedbackMRI
      *
      * @return bool true if operation succeeded
      */
-    function _addComment($comment='', $commentTypeID='', $predefinedCommentID='')
+    function _addComment(string $comment='', int $commentTypeID='', int $predefinedCommentID=''): bool
```
driusan: the defaults assigned to the latter 2 aren't ints.

PR #4266, opened 2019-01-16
[Testing] Add PhanTypeMismatchForeach rule to static analysis

johnsaigle:
This rule makes sure that when you pass a variable to foreach that the variable is expected to be something iterable. We had a few cases where parameters that were declared to be strings were actually arrays.

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -28,7 +28,7 @@ return [
         "PhanTypeArraySuspiciousNullable",
 		"PhanUndeclaredClassMethod",
 		"PhanUndeclaredConstant",
-		"PhanTypeMismatchForeach",
+		"PhanTypeMismatchDefault",
 		"PhanTypeMismatchArgument",
 		"PhanTypeMismatchArgumentInternal",
 		"PhanTypeMismatchReturn",
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -28,7 +28,6 @@ return [
         "PhanTypeArraySuspiciousNullable",
 		"PhanUndeclaredClassMethod",
 		"PhanUndeclaredConstant",
-		"PhanTypeMismatchDefault",
 		"PhanTypeMismatchArgument",
 		"PhanTypeMismatchArgumentInternal",
 		"PhanTypeMismatchReturn",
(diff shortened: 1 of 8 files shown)


htdocs/feedback_mri_popup.php
```diff
@@ -47,6 +47,9 @@
     $comments->setPredefinedComments($_POST['savecomments']['predefined']);
 
     // save all textual comments but only if there is an entry [sebas]
+    // The foreach below is operating on an array even though phan doesn't
```
zaliqarosli: this seems hacky? why would phan think it was not an array if it was and if the rule is supposed to check exactly that?

zaliqarosli: would defining an empty array variable, and then setting it to `$_POST['savecomments']['text']` bypass this?

johnsaigle: I'll give it a shot!

johnsaigle: No 😞

zaliqarosli: :(

driusan: We should always be using the most up-to-date version of phan (or just about any other library) that doesn't cause breakages in Travis..

With regard to this, I don't like the suppression either. Why can't you use `\Utility::asArray` to ensure foreach is working with an array?

johnsaigle: I'll give it a try

johnsaigle: Passes phan! Not really sure how to test it though

driusan: Click on image level feedback in the imaging browser and make sure things still work/save in the popup.

johnsaigle: Found some type errors while testing. Fixed in #4285. Will block this until that's merged.

johnsaigle:
@zaliqarosli @driusan Everything seems to work so this is ready for re-review!

.phan/config.php
```diff
@@ -30,7 +30,7 @@
 		"PhanUndeclaredClassMethod",
 		"PhanUndeclaredTypeParameter",
 		"PhanUndeclaredConstant",
-		"PhanTypeMismatchForeach",
+		"PhanTypeMismatchDefault",
```
driusan: This looks like it's just trading one ignored type for another. Why is MismatchDefault re-added?

johnsaigle: Oops bad rebase. I'll re-delete it.

htdocs/feedback_mri_popup.php
zaliqarosli: the comment on these lines can be removed now no?

Issue #4267, opened 2019-01-18
Typo in variable name in Statistics module


ridz1208:
class: `modules/statistics/php/stats_demographic.class.inc`

line 212, variable `$paramSite` is never used. should be `$param_site` instead

related to #4058

PR #4273, opened 2019-01-21
[Testing] Add PhanUndeclaredVariable rule to static analysis

johnsaigle:
### Brief summary of changes

Don't allow LORIS to proceed when using undeclared variables.

This PR contains a hybrid approach to fixing the issues. In some cases simple initializations were all that were needed. In others, I had to dig through old commits to find out what the variables should use.

There are some cases where I deleted old code when I couldn't imagine how it could work when the variable was not defined. This is what happened with the `bvl_feedback` ajax scripts. I assumed that this code simply isn't used generally and so only exists vestigially in LORIS. Please correct me if I've made a mistake.

### Related

Resolves https://github.com/aces/Loris/issues/4267

### To Test
On major, delete the phan suppression line `PhanUndeclaredVariable` from .phan/config.php. Run phan and you should get many errors.

Checkout my branch and run phan. The errors should be gone.


modules/bvl_feedback/ajax/get_bvl_feedback_summary.php
```diff
@@ -1,23 +0,0 @@
...
- * @license  GPLv3 <http://www.gnu.org/licenses/gpl-3.0.en.html>
- * @link     https://www.github.com/aces/Loris-Trunk/
- */
-
-header("content-type:application/json");
-ini_set('default_charset', 'utf-8');
-
-require "bvl_panel_ajax.php";
-
-$feedbackThreadSummary = $feedbackThread->getSummaryOfThreads();
```
johnsaigle: `$feedbackThread` is undefined in all of these files. Given that they rely on it being populated, I think it makes sense to remove them.

modules/document_repository/php/document_repository.class.inc
```diff
@@ -316,10 +316,15 @@ class Document_Repository extends \NDB_Menu_Filter
         foreach ($dbresult as $row) {
             $results[] =$row;
             $tree      = null;
+            // FIXME the for-loop below is pretty confusing and is not 
+            // documented. This should be cleaned up and the phan suppressions
+            // removed.
             foreach ($results as $result) {
+                //@phan-suppress-next-line PhanUndeclaredVariable
                 $thisref = &$refs->{$result['id']};
```
johnsaigle: black magic PHP reference juggling i just don't know

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -190,17 +195,17 @@ class Stats_Demographic extends \NDB_Form
         }
 
         //SUBPROJECTS
-        $suproject_query = '';
-        // Only search for subprojects if $currentProject is a positive int.
```
johnsaigle: The name of this variable was updated in Zain's statstics refactoring PR.

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -190,17 +195,17 @@ class Stats_Demographic extends \NDB_Form
...
-        if (is_numeric($currentProject) && intval($currentProject) >= 0) {
+        $subprojectQuery = '';
```
johnsaigle: The camelCase changes in this file represent the actual names of the variables, not an aesthetic choice. The snake_case variable names are incorrect.

modules/training/ajax/markQuiz.php
```diff
@@ -139,8 +139,6 @@ function process($values)
         // the identifier, training passes it as a value
         if (!empty($values['examiner'])) {
             $examinerID = $values['examiner'];
-        } else {
-            $examinerID = $this->identifier;
```
johnsaigle: `$this` does not exist as the `ajax` script is not an object.

php/libraries/Utility.class.inc
```diff
@@ -587,54 +587,6 @@ class Utility
...
-     * @param string $inst The test_name of the instrument for which you would
-     *                     like to know the score columns
-     *
-     * @return array (non-associative) of scored columns for the instrument given
-     */
-    static function getScoreColsForInstrument(string $inst): array
-    {
-        include_once "NDB_BVL_Instrument.class.inc";
-        $i = NDB_BVL_Instrument::factory($inst, '', '', true);
-        if (isset($i->scoreLabels)) {
```
johnsaigle: The `NDB_BVL_Instrument` object does not have the class variables `scoreLabels`, `scoreColumns` or `scores` so this function and the one above were removed.

php/libraries/Utility.class.inc
```diff
@@ -587,54 +587,6 @@ class Utility
         return $instruments;
     }
 
-    /**
-     * Get a list of all instruments which have scoring implemented.
-     * Used in statistics module.
-     *
-     * @return array non-associative array of test_names
-     */
-    static function getScoredInstruments(): array
```
johnsaigle: This function is never used in LORIS.

php/libraries/Utility.class.inc
driusan: specific instrument instances in projects have them set.

php/libraries/Utility.class.inc
driusan: it used to be used by the statistics module (and DQT?) in order to get a list of instruments and there columns that should be displayed for scatterplots. I don't think it's used anymore.

modules/document_repository/php/document_repository.class.inc
PapillonMcGill: Might be $ref was used before but some code modification removed it's declaration or other strange thing that append when someone modify some existing core.

driusan: @PapillonMcGill no, as far as I know this code was written by someone before we had much of a code review process in place and was just always poorly written..

php/libraries/Utility.class.inc
driusan: (note: the above comment doesn't mean you can't remove it.. I think it was for the same place as getScoredInstruments which doesn't exist anymore.)

johnsaigle: Good to know

modules/document_repository/php/document_repository.class.inc
johnsaigle: @PapillonMcGill Yes I tried selectively removing parts of the could or declaring `$ref` and `$thisref` explicitly and it broke the tree structure.

PapillonMcGill: I mean remove from line 313 to 335

driusan:
I sent https://github.com/johnsaigle/Loris/pull/13 to fix the suppressions.

johnsaigle:
@driusan Are you asking me to rewrite the doc repo code? The suppressions are added by this PR because otherwise phan will fail.

driusan:
The phan errors from the doc repo code are fixed by https://github.com/johnsaigle/Loris/pull/13, you just need to merge it into the branch (and rebase now that it seems to have a conflict with stats_demographic.class.inc)

johnsaigle:
Just looked into the Genomics file. It looks like the variables I commented out never existed in this repo (I went back to the first commit).

I think that there's no harm in merging my approach here. The module looks like it needs deeper refactoring.

driusan:
@ridz1208 is going to find someone to test the bvl_feedback frontend

johnsaigle:
@driusan I reverted my changes to the genomic file and addressed the problem in another PR https://github.com/aces/Loris/pull/4330. As a result this PR is blocked until that one is merged.

Could you update your review?

driusan:
@ridz1208 what's the status of finding someone to manually test BVL Feedback with this change?

sruthymathew123:
@johnsaigle The bvl feedback module is not working as expected. I am not able to save a visit level feedback; and lots of error in the backend.
error_bvl

johnsaigle:
I discovered that the issue with the ajax/ scripts is due to the way phan handles global variables. I have implemented the fix recommended by the developers of phan which is to toggle on the option `ignore_undeclared_variables_in_global_scope`.

sruthymathew123:
BVL feedback is no more impacted with this PR. All other changes looks good to me and test condition satisfied.

PR #4274, opened 2019-01-21
[Testing] Add PhanUndeclaredTypeParameter rule to static analysis

johnsaigle:
Makes sure that the parameters in PHPDocs have real types associated.

Based on #4069 because there some issues are flagged by both rules.


php/libraries/FeedbackMRI.class.inc
```diff
@@ -58,6 +58,28 @@ class FeedbackMRI
...
+
+        // run the query
+        $success = mysql_query($query);
+
+        // return success
+        return $success;
+    }
+
+    /**
+>>>>>>> 9e8834ab4... Fix PhanUndeclaredTypeReturnType errors
```
driusan: Pretty sure this isn't supposed to be here

test/unittests/Loris_PHPUnit_Database_TestCase.php
```diff
@@ -89,6 +89,7 @@ protected function setUp()
      *
      * @return \PHPUnit_Extensions_Database_DB_DefaultDatabaseConnection
      */
+    /* @phan-suppress-next-line PhanUndeclaredTypeReturnType*/
```
driusan: Why isn't it declared as PHPUnit_Extensions_Database_DB_DefaultDatabaseConnection?

htdocs/api/v0.0.2/Candidates.php
```diff
@@ -218,7 +218,7 @@ protected function verifyField($data, $field, $values)
      * @param string $gender   Gender of the candidate to be created
      * @param string $PSCID    PSCID of the candidate to be created
      *
-     * @return none
+     * @return \Candidate
```
driusan: I think these API changes are going to cause conflicts with #4244 and I'd prefer that one goes in first.

Maybe it's best to just temporarily add "htdocs/api" to the "exclude_analysis_directory_list" in the phan config since it's going to disappear as soon as @xlecours is done converting it to a module anyways.

johnsaigle: That's a good idea, will do

test/unittests/Loris_PHPUnit_Database_TestCase.php
johnsaigle: I remember that phan just didn't seem to know about it... will re-test and try again.

driusan: if phan doesn't know about it, that means the PHP doesn't either and phan is flagging a bug

htdocs/api/v0.0.2/Candidates.php
johnsaigle: @driusan I'll make a separate PR to add `htdocs/` to the exclude_analysis_directory_list because I think this is likely to come up in the other Phan PRs. It'll be easier to merge that one first and rebase the others.

driusan: 'htdocs' shouldn't be excluded, only 'htdocs/api'. Other php files in the 'htdocs' directory should still be analyzed.

modules/imaging_uploader/php/file_decompress.class.inc
```diff
@@ -137,7 +137,7 @@ class File_Decompress
      * extracting it to the destination folder and
      * then closing the zip-archive handler
      *
-     * @return booean
```
driusan: I never knew you existed, but I'll miss you, "booean"

driusan:
Needs to pass Travis before I can merge it, but it looks good now

php/libraries/FeedbackMRI.class.inc
```diff
@@ -57,6 +57,27 @@ class FeedbackMRI
...
+    function clearPredefinedComments()
+    {
+        // start the query
+        $query = "DELETE FROM feedback_mri_comments WHERE ".$this->_buildWhere();
+
+        // limit to predefined comments
+        $query .= "AND PredefinedCommentID IS NOT NULL";
+
+        // run the query
+        $success = mysql_query($query);
```
driusan: Didn't you remove this in a previous PR since mysql_query hasn't existed since PHP 7? Why is it getting added here?

modules/data_team_helper/php/data_team_helper.class.inc
```diff
@@ -790,7 +790,7 @@ class Data_Team_Helper extends \NDB_Form
      *
      * @param string $visit_label the value of visit label
      *
-     * @return   unknown
+     * @return   array
      * @throws   \DatabaseException
      * @internal param unknown_type $visit
      */
```
PapillonMcGill: Missed getJSDependencies() and getCSSDependencies()  on line 831 and 850

modules/imaging_browser/php/viewsession.class.inc
```diff
@@ -351,8 +351,8 @@ class ViewSession extends \NDB_Form
     /**
      * Gets the date according to the type
      *
-     * @param MRIFile $file file object
-     * @param string  $type type of the date
+     * @param \MRIFile $file file object
+     * @param string   $type type of the date
      *
      * @return string The date if exists, if not an empty string
```
PapillonMcGill:
```suggestion
     * @return int|string Date (Unix int timestamp) if exists, or an empty string
```

driusan:
@PapillonMcGill's comments can be done in another PR.. since this is both passing Travis and removing some rules from the list of ignored phan rule types, it's a step forward and I'm going to merge it.

PR #4275, opened 2019-01-22
[API] Adding json responses classes

xlecours:
This brings current development from https://github.com/aces/Loris/pull/4244 to minor so it can be share/use earlier then when the other PR will be merged.

### Brief summary of changes
- Allow to generate PSR-7 responses in one line instead of repeating the same 7 lines each time.
- Login endpoint now use the new response classes
- Etag are added only if the header do not exists. (This resolves issue where etag from subendpoints were overwritten by parents endpoint.)

Todo:
- [ ] ~~unit tests --> https://github.com/xlecours/Loris/pull/5 (discussion required)~~ will be sent in an other PR

### DIFF
--- a/src/Api/Endpoints/Login.php
+++ b/src/Api/Endpoints/Login.php
@@ -88,7 +88,7 @@ class Login extends Endpoint
 
             if ($user === null || $password === null) {
                 return new \LORIS\Http\Response\BadRequest(
-                    'missing username or password'
+                    'Missing username or password'
                 );
             }
 
@@ -102,7 +102,7 @@ class Login extends Endpoint
                     );
                 } else {
                     return new \LORIS\Http\Response\InternalServerError(
-                        'unacceptable JWT key'
+                        'Unacceptable JWT key'
                     );
                 }
             }
--- a/src/Api/Endpoints/Login.php
+++ b/src/Api/Endpoints/Login.php
@@ -41,13 +41,11 @@ class Login extends Endpoint
     /**
      * Return which methods are supported by this endpoint.
      *
-     * Login can only be POSTed to
-     *
      * @return array supported HTTP methods
      */
     protected function allowedMethods() : array
     {
-        return ['POST'];
+        return array('POST');
     }
 
     /**
@@ -61,11 +59,11 @@ class Login extends Endpoint
      */
     protected function supportedVersions() : array
     {
-        return [
-                "v0.0.1",
-                "v0.0.2",
-                "v0.0.3-dev",
-               ];
+        return array(
+                'v0.0.1',
+                'v0.0.2',
+                'v0.0.3-dev',
+               );
     }
 
     /**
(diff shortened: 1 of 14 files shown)


src/Api/Endpoints/Login.php
```diff
@@ -77,56 +75,49 @@ protected function supportedVersions() : array
...
-            return (new \LORIS\Http\Response())
-                ->withBody(
-                    new \LORIS\Http\StringStream(
-                        json_encode(
-                            array('error' => 'Missing username or password')
-                        )
-                    )
-                )
-                ->withStatus(400);
+        if (count($request->getAttribute('pathparts')) !== 1) {
```
davidblader: just out of curiosity, what is this checking exactly?

src/Api/Endpoints/Login.php
```diff
@@ -77,56 +75,49 @@ protected function supportedVersions() : array
...
+        switch ($request->getMethod()) {
+        case 'POST':
+            $requestdata = json_decode((string) $request->getBody(), true);
+
+            $user     = $requestdata['username'] ?? null;
+            $password = $requestdata['password'] ?? null;
+
+            if ($user === null || $password === null) {
+                return new \LORIS\Http\Response\BadRequest(
+                    'missing username or password'
```
davidblader: nitpick but might be best to maintain capitalization of original err. message

src/Api/Endpoints/Login.php
```diff
@@ -77,56 +75,49 @@ protected function supportedVersions() : array
...
+
+            if ($login->passwordAuthenticate($user, $password, false)) {
+                $token = $this->getEncodedToken($user);
+                if (!empty($token)) {
+                    return new \LORIS\Http\Response\JsonResponse(
+                        array('token' => $token)
+                    );
+                } else {
+                    return new \LORIS\Http\Response\InternalServerError(
+                        'unacceptable JWT key'
```
davidblader: same comment re: capitalization

src/Api/Endpoints/Login.php
xlecours: @davidblader The idea is to check if someone entered `/login/xyz/...`. In that case, it is safe to say that it should be a NotFound response because there is nothing under `/login`
The operator could be `> 1` instead.

src/Api/Endpoints/Login.php
xlecours: adressed

src/Api/Endpoints/Login.php
xlecours: addressed

xlecours:
Last commit was to comply to https://tools.ietf.org/html/rfc7232#section-4.1
`A 304 response cannot contain a message-body; it is always terminated by the first empty line after the header fields.`

src/Http/Response/BadRequest.php
```diff
@@ -0,0 +1,44 @@
...
+ * A LORIS Http Response is an implementation of the PSR15 ResponseInterface
+ * to use in LORIS specific for 400 Bad Request.
+ *
+ * @category PSR15
+ * @package  Http
+ * @author   Xavier Lecours Boucher <xavier.lecours@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+class BadRequest extends JsonResponse
```
driusan: Why are BadRequests always JsonResponses? I think it would make more sense for a BadRequest to be a type of response with an error message, and a middleware can convert that message to the appropriate format based on the 'Content-Type' header of the response. (I'm undecided if that should be part of the PageDecorationMiddleware or a different one, but in either case it needs to be done carefully wrt. the ETag middleware to ensure that different content types get different etags..)

src/Http/Response/BadRequest.php
```diff
@@ -0,0 +1,44 @@
+<?php declare(strict_types=1);
+/**
+ * File contains the PSR15 ResponseInterface implementation for
```
driusan:
```suggestion
 * File contains the PSR7 ResponseInterface implementation for
```

ResponseInterface is part of PSR7, not 15.

src/Http/Response/BadRequest.php
xlecours: When the need arise the change should not be too hard to make.
If it is only for html, the template file could use the json_decode(response->getBody())['error'] as a error message value.

src/Http/Response/BadRequest.php
driusan: (I only requested changes on the one file, but the comments for all of the files have this problem.)

xlecours:
@driusan Do you want me to add the middleware or can we punt that to when it will be needed?

xlecours:
@driusan this is in your court. Thanks

driusan:
@xlecours sorry, I missed in in the middle of all the other notifications on the repo.. I'm still uncomfortable with error responses being inherently JSON, and not being semantically errors which are converted to JSON based on the content type negotiated, but since this is only used for the API right now I guess we can punt it for a bit and revisit later.

PR #4276, opened 2019-01-22
[Testing] Add PhanTypeVoidAssignment to static analysis

johnsaigle:
This rule prevents the variables being assigned to functions that are declared to return void, which makes sense.

### DIFF
--- a/modules/statistics/php/statistics_mri_site.class.inc
+++ b/modules/statistics/php/statistics_mri_site.class.inc
@@ -209,7 +209,7 @@ class Statistics_Mri_Site extends Statistics_Site
 
         $data = array();
         foreach ($this->issues as $issue) {
-            // FIXME The function _completeCount() seems ot be unfinished and
+            // FIXME The function _completeCount() seems to be unfinished and
             // always returns null.
             $complete_count = $this->_completeCount($centerID, $projectID, $issue);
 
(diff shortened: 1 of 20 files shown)


php/libraries/NDB_Form.class.inc
```diff
@@ -54,7 +54,7 @@ class NDB_Form extends NDB_Page
      * @param string $page       Identifies the page to show
      * @param string $identifier Variables passed to form
      *
-     * @return void
+     * @return object
```
driusan:
```suggestion
     * @return \NDB_Form
```

I think the factory for an NDB_Form should always return some type of NDB_Form, not a generic object.

modules/statistics/php/statistics_mri_site.class.inc
```diff
@@ -73,9 +73,9 @@ class Statistics_Mri_Site extends Statistics_Site
      * @param string $projectID the value of projectID
      * @param string $issue     the value of issue
      *
-     * @return void
+     * @return ?string
      */
-    function _completeCount($centerID, $projectID, $issue)
+    function _completeCount($centerID, $projectID, $issue): ?string
```
PapillonMcGill: line 92   _getResults()  add **:array**

PapillonMcGill: And stressed that same point in another PR :-)

driusan:
#4283 is merged.

modules/statistics/php/statistics_mri_site.class.inc
```diff
@@ -209,6 +209,8 @@ class Statistics_Mri_Site extends Statistics_Site
 
         $data = array();
         foreach ($this->issues as $issue) {
+            // FIXME The function _completeCount() seems ot be unfinished and
```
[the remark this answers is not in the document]
driusan: @johnsaigle

PR #4277, opened 2019-01-22
[Testing] Add PhanTypeMismatchDimFetch to static analysis

johnsaigle:
This rule makes sure that when you access an array that the key you're using makes sense. E.g. if you are accessing an element from an array using a non-numeric key, but the array is actually declared to be a string, this will flag that case.

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -19,7 +19,6 @@ return [
         "PhanTypeInvalidDimOffset",
 		"PhanUndeclaredMethod",
 		"PhanUndeclaredVariableDim",
-        "PhanTypeMismatchDimFetch",
 		"PhanUndeclaredClassMethod",
 		"PhanTypeMismatchArgument",
 		"PhanTypeMismatchReturn",
--- a/modules/document_repository/php/document_repository.class.inc
+++ b/modules/document_repository/php/document_repository.class.inc
@@ -111,7 +111,7 @@ class Document_Repository extends \NDB_Menu_Filter_Form
     /**
      * Handler of parsing category
      *
-     * @param string $value the value
+     * @param array $value the value
      *
      * @return array
      */
--- a/modules/document_repository/php/document_repository.class.inc
+++ b/modules/document_repository/php/document_repository.class.inc
@@ -111,7 +111,7 @@ class Document_Repository extends \NDB_Menu_Filter_Form
     /**
      * Handler of parsing category
      *
-     * @param array $value the value
+     * @param string $value the value
      *
      * @return array
      */
--- a/modules/next_stage/php/next_stage.class.inc
+++ b/modules/next_stage/php/next_stage.class.inc
@@ -215,7 +215,7 @@ class Next_Stage extends \NDB_Form
     /**
      * Validate function
      *
-     * @param string $values the value
+     * @param array $values the value
      *
      * @return array|bool
      */
--- a/php/libraries/File_Upload.class.inc
+++ b/php/libraries/File_Upload.class.inc
@@ -490,7 +490,7 @@ class File_Upload
      * @param string $dir  The target directory for the file.
      * @param string $name The proposed name for the file.
      *
-     * @return array New filename
+     * @return string New filename
      * @see    File_Upload::moveFile()
      */
     function nameToIncremental(string $dir, string $name): string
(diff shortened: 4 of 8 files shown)


php/libraries/Utility.class.inc
```diff
@@ -874,11 +874,11 @@ class Utility
      * This is used by create timepoint and start timepoint pages,
      * so it probably shouldn't be in the instrument class.
      *
-     * @param string $dateElement The date element in QuickForm array format.
+     * @param array|string $dateElement The date element in QuickForm array format.
```
```suggestion
     * @param array $dateElement The date element in QuickForm array format.
```

modules/next_stage/php/next_stage.class.inc
```diff
@@ -215,7 +215,7 @@ class Next_Stage extends \NDB_Form
     /**
      * Validate function
      *
-     * @param string $values the value
+     * @param array $values the value
```
PapillonMcGill: Can also return 'true' (see line 281)

php/libraries/File_Upload.class.inc
```diff
@@ -489,7 +489,7 @@ class File_Upload
      * @param string $dir  The target directory for the file.
      * @param int    $name The proposed name for the file.
      *
-     * @return string New filename
+     * @return array New filename
```
PapillonMcGill:
```suggestion
     * @return string New filename
```

.phan/config.php
```diff
@@ -24,7 +24,7 @@
 		"PhanUndeclaredMethod",
 		"PhanUndeclaredVariable",
 		"PhanUndeclaredVariableDim",
-        "PhanTypeMismatchDimFetch",
+        "PhanUndeclaredTypeReturnType",
```
driusan: Bad rebase again?

php/libraries/Utility.class.inc
```diff
@@ -874,11 +874,11 @@ class Utility
...
      * so it probably shouldn't be in the instrument class.
      *
-     * @param string $dateElement The date element in QuickForm array format.
+     * @param array $dateElement The date element in QuickForm array format.
      *
-     * @return True if the date is invalid, false if it's valid.
+     * @return bool True if the date is invalid, false if it's valid.
      */
-    static function _checkDate(string $dateElement): bool
+    static function _checkDate($dateElement): bool
```
driusan: why doesn't the signature change to an array, only the doc?

johnsaigle:
See #4320 for information on the checkDate function. It should be deleted. This PR should probably be blocked until that one is merged or else this will just create a conflict.

php/libraries/Utility.class.inc
```diff
@@ -870,6 +870,36 @@ class Utility
...
+        ) {
+            // otherwise, if any of the three elements are empty, return
+            // false b/c date entry has been attempted
+            return false;
+        }
+        return checkdate($dateElement['M'], $dateElement['d'], $dateElement['Y']);
+    }
+
+    /**
+>>>>>>> b479d3bbc... Add PhanTypeMismatchDimFetch
```
driusan: Left over conflict marker

johnsaigle:
Travis build failed on 7.3 with 
System info: host: 'f19bd295f6ce', ip: '172.18.0.3', os.name: 'Linux', os.arch: 'amd64', os.version: '4.4.0-101-generic', java.version: '1.8.0_91'

Tagging #4394 as related. Restarting Travis now.

modules/document_repository/php/document_repository.class.inc
```diff
@@ -233,7 +233,7 @@ class Document_Repository extends \NDB_Menu_Filter
     /**
      * ParseCategory function
      *
-     * @param string $value the value
+     * @param array $value the value
      *
      * @return array
```
PapillonMcGill: It look like the return value is a string
Plus, the DB query is not sanitized. Should be change to pselect (in another PR).

php/libraries/Utility.class.inc
```diff
@@ -824,6 +824,35 @@ class Utility
         return $test_names;
     }
 
+    /**
```
driusan: Why is this PR adding this function?

Other than that it looks good, given that it's just changing comments and still passing Travis..

driusan:
Actually can you put this on minor before they diverge too much?

php/libraries/Utility.class.inc
johnsaigle: Rebase error I guess. I'll update it to remove this and also change the base branch.

PR #4294, opened 2019-02-04
[Help Editor] Remove compiled Javascript file

johnsaigle:
This file is leftover from module Reactification and causes problems when switching branches after running `make dev`

### DIFF
--- a/modules/help_editor/js/.gitignore
+++ b/modules/help_editor/js/.gitignore
@@ -1,1 +1,2 @@
 columnFormatter.js
+help_editor.js


HenriRabalais:
Should the file simply be removed instead of adding it to git ignore?

ridz1208:
@HenriRabalais both are necessary, the file already does not exist on the branch, but adding it to the git ignore prevents git from even tracking it on dev VMs

ridz1208:
@johnsaigle by the same logic shouldn't the media module have a git ignore file  for the entire `js/*` directory ?

johnsaigle:
@ridz1208 Yeah all new Reactified modules shouldn't be tracking the compiled JS files. This file just kept preventing me from switching branches so I was more aware of it.

PR #4302, opened 2019-02-04
[candidate list] Reactification

cmadjar:
### Description

This PR reactifies the candidate list menu filter.

**1. Display according to permissions**

- Screen shot of the reactified menu filter for users that have 'access all profiles' permission:
screen shot 2019-02-04 at 6 39 40 pm

- Screen shot of the reactified menu filter for users that do not have 'access all profiles' permission:
screen shot 2019-02-04 at 6 39 55 pm

**2. Menu filter Basic/Advanced options**

- Display of Basic filter options:
screen shot 2019-02-04 at 6 50 07 pm

- Display of Advanced filter options:
screen shot 2019-02-04 at 6 50 20 pm

### This resolves

[x] see reactification project (not sure how to link cards in this PR)

### To test this

- [ ] make sure that the filters are all working
- [ ] make sure the toggle button for Advanced or Basic filters works as expected show more or less options
- [ ] if a user has 'access all profile' permission, make sure that the PSCID field becomes a link and that the 'Open Profile' button does not show
- [ ] if a user does not have 'access all profile' permission, make sure that the PSCID field is not a link and that the 'Open Profile' button shows up


driusan:
@cmadjar it's failing travis

kongtiaowang:
@cmadjar It looks so nice. There is only one minor issue. If set not use project in configuration, the project and subproject filters will not work, these filters should be disappeared in the frontend.

modules/candidate_list/jsx/candidateListIndex.js
```diff
@@ -0,0 +1,363 @@
...
+    }
+    if (column === 'Scan Done' && cell === 'Y') {
+      return (
+        <td className="scanDoneLink">
+        <a href="#"
+          onClick={loris.loadFilteredMenuClickHandler('imaging_browser/',
+          {pscid: row['PSCID']})}
+        >
+          {cell}
+        </a>
```
HenriRabalais: <a> needs to be indented!

modules/candidate_list/jsx/candidateListIndex.js
```diff
@@ -0,0 +1,363 @@
...
+          'label': 'EDC',
+          'show': true,
+          'filter': {
+            name: 'edc',
+            type: 'date',
+            hide: this.state.hideFilter,
+          },
+        }
+      );
+    }
```
HenriRabalais: Mutation of the array (i.e. `fields.push`) should generally be avoided in functional programming. But I don't think it's a big deal here.

modules/candidate_list/jsx/candidateListIndex.js
```diff
@@ -0,0 +1,363 @@
...
+      const newFilterState = this.state.hideFilter === true ? false : true;
+      let newLabel = '';
+      if (this.state.buttonLabel === 'Show Advanced Filters') {
+        newLabel = 'Hide Advanced Filters';
+      } else {
+        newLabel = 'Show Advanced Filters';
+      }
+      this.setState({hideFilter: newFilterState});
+      this.setState({buttonLabel: newLabel});
+    };
```
HenriRabalais: I would write this as a function outside of the render function because it doesn't need to be rendered on every mount. It can also be condensed into:

```
toggleFilters() {
  const hideFilter = !this.state.hideFilter;
  const buttonLabel = hideFilter ? 'Show Advanced Filters' : 'Hide Advanced Filters';
  this.setState({hideFilter, buttonLabel});
}
```

modules/candidate_list/jsx/candidateListIndex.js
```diff
@@ -0,0 +1,363 @@
...
+
+window.addEventListener('load', () => {
+  ReactDOM.render(
+  <CandidateListIndex
+  dataURL={`${loris.BaseURL}/candidate_list/?format=json`}
+  hasPermission={loris.userHasPermission}
+  baseURL={loris.BaseURL}
+  />,
+  document.getElementById('lorisworkspace')
+);
```
HenriRabalais: There's some indentation alignment that needs to be done here.

modules/candidate_list/jsx/openProfileForm.js
```diff
@@ -89,34 +91,32 @@ class AccessProfilePanel extends Component {
...
-                 />
-             </FormElement>
-        </Panel>
-        </div>
-       );
+    return (
+      <FormElement
+        name='openprofile'
+        onSubmit={this.validateAndSubmit}
+        onUserInput={this.validateAndSubmit}>
```
HenriRabalais: I'm not sure if `FormElement` takes a prop called `onUserInput`. Is there are reason for having it here? I also noticed you have the button element at the bottom of this form also triggering `this.validateAndSubmit`. I think believe only one of these is necessary.

modules/candidate_list/php/candidate_list.class.inc
```diff
@@ -210,125 +135,45 @@ class Candidate_List extends \NDB_Menu_Filter
         \Module::factory('candidate_parameters');
 
         // create user object
-         $user    =& \User::singleton();
-          $config =& \NDB_Config::singleton();
-        // PSC
+        $user   =& \User::singleton();
+        $config =& \NDB_Config::singleton();
```
HenriRabalais: I don't believe you need the `&`s here anymore.

kongtiaowang:
The Sex filter doesn't work, it should show "Female" and "Male", not "0" and "1".

modules/candidate_list/jsx/candidateListIndex.js
cmadjar: Actually, for some reason I cannot move it out of the render function. I get many errors if I do. At least I simplified it in the next commit.

cmadjar:
> @cmadjar It looks so nice. There is only one minor issue. If set not use project in configuration, the project and subproject filters will not work, these filters should be disappeared in the frontend.

@kongtiaowang Great catch! This should be taken care of in the next commit. Thanks!

modules/candidate_list/jsx/candidateListIndex.js
cmadjar: ha! I had an epiphany last night and realized why I got errors when moving the function out of the render() function. I forgot to bind the function in the constructor, duh... 

Anyway, all done!

kongtiaowang:
All the filters don't work anymore.

cmadjar:
> All the filters don't work anymore.
@kongtiaowang oupsie.. This is fixed now! Please re-review. Thanks!

webpack.config.js
```diff
@@ -18,9 +18,9 @@ const config = [{
     './modules/configuration/js/SubprojectRelations.js': './modules/configuration/jsx/SubprojectRelations.js',
     './modules/bvl_feedback/js/react.behavioural_feedback_panel.js': './modules/bvl_feedback/jsx/react.behavioural_feedback_panel.js',
     './modules/data_team_helper/js/behavioural_qc_module.js': './modules/data_team_helper/jsx/behavioural_qc_module.js',
-    './modules/candidate_list/js/AccessProfilePanel.js': './modules/candidate_list/jsx/AccessProfilePanel.js',
-    './modules/candidate_list/js/columnFormatter.js': './modules/candidate_list/jsx/columnFormatter.js',
+    './modules/candidate_list/js/AccessProfilePanel.js': './modules/candidate_list/jsx/openProfileForm.js',
```
HenriRabalais: Should this be written `'./modules/candidate_list/js/openProfileForm.js': './modules/candidate_list/jsx/openProfileForm.js',` instead?

modules/candidate_list/jsx/candidateListIndex.js
HenriRabalais: You could either reassign the value of fields by doing something like: 
```
let fields = [...fields,
        {
           'label': 'EDC',
           'show': true,
           'filter': {
             name: 'edc',
             type: 'date',
             hide: this.state.hideFilter,
           },
         }
]
```

Or the alternative would be to use a `map` or `reduce` function. I honestly think in this case, after taking a look at it, using `fields.push` is just fine!

cmadjar: good to know. Thanks for the suggestion!

webpack.config.js
cmadjar: Oupsie... This is fixed now. Nice catch!

maltheism:
Adding priority high tag since we would like to get this merged into major for biobank.

driusan:
@kongtiaowang you said above that the filters weren't working.. is there an update? Does it work with the latest changes?

kongtiaowang:
@driusan It is already updated.

modules/candidate_list/jsx/candidateListIndex.js
```diff
@@ -0,0 +1,363 @@
...
+import Modal from 'Modal';
+
+import OpenProfileForm from './openProfileForm';
+
+/**
+ * Candidate List
+ *
+ * Main module component rendering the candidate list and open profile form
+ *
+ * @author Dave McFarlane
```
driusan: I don't know who this guy is or how he became an author.

cmadjar: technically, you started the work and created that file . I just polished it so did not feel like adding me as an author. ;) I can remove your name if you prefer and make it LORIS team generic (is there such a thing in the LORIS code? A LORIS team generic author by default?).

driusan: Can you just fix the spelling of it?

cmadjar: @driusan so sorry for the typo. It is fixed now.

driusan:
:tada: :100:

PR #4308, opened 2019-02-06
[Phan] Add PhanTypeArraySuspiciousNullable rule

johnsaigle:
### Brief summary of changes

Adds PhanTypeArraySuspiciousNullable rule to our analysis. This rule flags code where array elements are accessed in places where the arrays can be null. As a result this PR adds a lot of `is_null` checks to DB results and throws a lot of Exceptions in case they happen.

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -22,7 +22,6 @@ return [
 		"PhanUndeclaredVariable",
 		"PhanUndeclaredVariableDim",
         "PhanTypeMismatchDimFetch",
-        "PhanTypeArraySuspiciousNullable",
 		"PhanUndeclaredClassMethod",
 		"PhanTypeMismatchArgument",
 		"PhanTypeMismatchArgumentInternal",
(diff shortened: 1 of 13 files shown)


php/libraries/Candidate.class.inc
```diff
@@ -595,6 +595,7 @@ class Candidate
         $result = $db->pselectRow($query, array('CaID' => $candID));
 
         //return the next visit No
+        //@phan-suppress-next-line PhanTypeArraySuspiciousNullable
```
driusan: 👎

johnsaigle: I added this here because the function is deprecated. Is it worth adding other changes to a function that people shouldn't be using anyway?

driusan:
@cmadjar Why did you assign this to me? I already reviewed it

cmadjar:
@driusan It looks like John replied to your comment. Also I was with Leigh when we did the assignments and I believe she mentioned it would be in your court (not sure).

LeighMac:
My bad, should not have given it to you @driusan  - should be w/  @johnsaigle .

php/libraries/Candidate.class.inc
johnsaigle: I've opened #4363 to remove the function since it's been marked as deprecated for 5+ years according to git blame. I'll mark this PR as blocked assuming that PR will go through.

driusan:
This is still failing Travis.

PR #4309, opened 2019-02-07
[WIKI] Study Parameters - Visits

ridz1208:
### Brief summary of changes

How to add/configure Visits
___

*This and all upcoming PRs for the WIKI will contain one or a couple files at most. this will hopefully prevent PRs from staying open for very long periods.*

*This and all upcoming PRs are open to discussions and suggestions. Most of the content has been copied over from the existing WIKI with sometimes minor, major or total changes.*

*Feel free to make suggestions and/or push directly to the branch !!

### DIFF
--- a/docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Timepoints.md
+++ b/docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Timepoints.md
@@ -1,0 +1,52 @@
+# Timepoints
+
+## Overview
+**Timepoints** are a list of possible instances in time where a candidate can visit or contact a site with the purpose of undergoing any form of data collection.
+
+**Timepoint** and **Visit** are terms used interchangeably in LORIS (the preferred term is **Timepoint**). **Timepoints** are defined in the `Visit_Windows` and `visit` tables of the database as well as the `<visitLabel>` tags in the `config.xml`.
+
+## Adding Timepoint Options
+
+### Front End
+ _not yet available_
+
+### SQL
+
+The `Visit_Windows` table lists the timepoints in the study and adds some study-specific age ranges defined for each timepoint. This table must be populated with visit labels for the proper functioning of LORIS. To populate the table with visit labels insert study-specific information as follows:
+
+```sql
+INSERT INTO Visit_Windows (Visit_label,  WindowMinDays, WindowMaxDays, OptimumMinDays, OptimumMaxDays, WindowMidpointDays) VALUES ('V1', 0, 100, 40, 60, 50);
+```
+
+If age is not a critical factor in study timepoint scheduling, define `Min` and `Max` values as `NULL`.
+
+> In event where the timepoints are already defined in the `config.xml` (see *Interaction With Loris* section below), run `tools/single_use/populate_visit_windows.php` to automatically import the visit labels into the `Visit_Windows` table.
+
+The `visit` table lists all the timepoints. This table must be populated with visit labels for the proper functioning of LORIS. To populate with timepoints insert information as follows:
+
+```sql
+INSERT INTO visit (VisitName) VALUES ('V1');
+```
+
+### API
+ _not yet available. See [API documentation](../../../API/) for latest additions_
+ 
+## Interaction With LORIS
+
+### Subprojects
+ **Timepoints** should be assigned to subprojects in order to be able to create timepoints for candidates. This association should be defined in the `config.xml` file of the `%LORIS_ROOT%/project/` directory as follows:
+ 
+ ```xml
+ <visitLabel subprojectID="1">
+   <labelSet>
+     <item value="V1">V1 label description</item>   
+     <item value="V2">V2 label description</item>   
+   </labelSet>
+ </visitLabel>
+ <visitLabel subprojectID="2">
+   <labelSet>
+     <item value="V1">V1 label description</item>   
+     <item value="V3">V3 label description</item>   
+   </labelSet>
+ </visitLabel>
+ ```
--- a/docs/wiki/99 - Developers/SQL Dictionary.md
+++ b/docs/wiki/99 - Developers/SQL Dictionary.md
@@ -9,17 +9,41 @@
 	|     Field    |                    Description                   |                                             Notes                                            |
 	|:------------:|:------------------------------------------------:|:--------------------------------------------------------------------------------------------:|
 	| `CenterID`   | Identifier of the site                           | Avoid setting this field explicitly when inserting data, it auto increments.                 |
-	| `Name`       | Full name of the site                          |                                                                                              |
-	| `PSCArea`    |                                                  | Deprecated                                                                                              |
-	| `Address`    | Civic address of the site                      |                                                                                              |
-	| `City`       | City where the site is located                 |                                                                                              |
-	| `StateID`    | Identifier of the state where the site is located  |                                                                                              |
-	| `ZIP`        | ZIP code of the site                           |                                                                                              |
+	| `Name`       | Full name of the site                            |                                                                                              |
+	| `PSCArea`    |                                                  | Deprecated                                                                                   |
+	| `Address`    | Civic address of the site                        |                                                                                              |
+	| `City`       | City where the site is located                   |                                                                                              |
+	| `StateID`    | Identifier of the state where the site is located|                                                                                              |
+	| `ZIP`        | ZIP code of the site                             |                                                                                              |
 	| `Phone1`     | Phone number 1                                   |                                                                                              |
 	| `Phone2`     | Phone number 2                                   |                                                                                              |
 	| `Contact1`   | Contact person 1                                 |                                                                                              |
 	| `Contact2`   | Contact person 2                                 |                                                                                              |
-	| `Alias`      | Shortname of the site. Limited to 3 characters | This field affects the alias section of the PSCID of candidates                              |
+	| `Alias`      | Shortname of the site. Limited to 3 characters   | This field affects the alias section of the PSCID of candidates                              |
 	| `MRI_alias`  | MRI specific shortname. Limited to 4 characters  |                                                                                              |
-	| `Account`    |                                                  | Deprecated                                                                                              |
-	| `Study_site` | Does this site recruit candidates?             | This field will affect the appearance of the center in dropdowns in several modules of LORIS |
+	| `Account`    |                                                  | Deprecated                                                                                   |
+	| `Study_site` | Does this site recruit candidates?               | This field will affect the appearance of the center in dropdowns in several modules of LORIS |
+
+- Table: `visit`
+
+ 	*This table stores the list of timepoints configured for the study.*
+ 	
+ 	|     Field    |          Description          |                                    Notes                                    |
+	|:------------:|:------------------------------------------------:|:--------------------------------------------------------:|
+	| `VisitID`   | Identifier of the Timepoint    | Avoid setting this field explicitly when inserting data, it auto increments.|
+	| `VisitName` | Back end name of the Timepoint | This name is used in file and imaging uploads                               |
+
+- Table: `Visit_Windows`
+
+ 	*This table associates timepoints with optimal age ranges for quality control monitoring.*
+ 	
+ 	|        Field         |                                        Description                                        |                         Notes                          |
+	|:--------------------:|:-----------------------------------------------------------------------------------------:|:------------------------------------------------------:|
+	| `ID `                | Identifier of the entry                                                                                   | Avoid setting this field explicitly when inserting data, it auto increments.           |
+	| `Visit_label `       | Name of the visit                                                                                         |  |
+	| `WindowMinDays `     | Candidate's minimum age in days for visit to be flagged as _within **permitted** parameters of the study_ | The only effect of this field is a YES/NO flag showing up on the instrument_list page|
+	| `WindowMaxDays `     | Candidate's maximum age in days for visit to be flagged as _within **permitted** parameters of the study_ | The only effect of this field is a YES/NO flag showing up on the instrument_list page|
+	| `OptimumMinDays `    | Candidate's minimum age in days for visit to be flagged as _within **optimal** parameters of the study_   | The only effect of this field is a YES/NO flag showing up on the instrument_list page|
+	| `OptimumMaxDays `    | Candidate's maximum age in days for visit to be flagged as _within **optimal** parameters of the study_   | The only effect of this field is a YES/NO flag showing up on the instrument_list page|
+	| `WindowMidpointDays `| Candidate's ideal age in days for a visit                                                                 |  |
+	
--- a/docs/wiki/99 - Developers/SQL Dictionary.md
+++ b/docs/wiki/99 - Developers/SQL Dictionary.md
@@ -28,10 +28,10 @@
 
  	*This table stores the list of timepoints configured for the study.*
  	
- 	|     Field    |          Description          |                                    Notes                                    |
-	|:------------:|:------------------------------------------------:|:--------------------------------------------------------:|
-	| `VisitID`   | Identifier of the Timepoint    | Avoid setting this field explicitly when inserting data, it auto increments.|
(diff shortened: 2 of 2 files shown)


docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
```diff
@@ -0,0 +1,44 @@
...
+
+### SQL (Recommended)
+
+This table must be populated with visit labels - the Imaging Pipeline critically depends on this.  To populate with visit labels, run tools/populate_visit_windows.php - or manually insert study-specific information:
+
+```sql
+INSERT INTO Visit_Windows (Visit_label,  WindowMinDays, WindowMaxDays, OptimumMinDays, OptimumMaxDays, WindowMidpointDays) VALUES ('V1', '0', '100', '40', '60', '50');
+```
+
+If age is a not critical factor in study visit scheduling, define Min value as 0, and Max value as 2147483647 (this is the maximum integer size).
```
davidblader: which `Min` value(s)?

davidblader: also might be best to just say "maximum integer size" rather than provide a specific value. column type could change or size could be variable across different SQLs

ridz1208: right on all counts

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
```diff
@@ -0,0 +1,44 @@
...
+**Visit** and **Timepoint** are terms used interchangeably in LORIS. **Visits** are defined in the `Visit_Windows` table of the database.
+
+## Adding Options
+
+### Front End
+ _not yet available_
+
+### SQL (Recommended)
+
+This table must be populated with visit labels - the Imaging Pipeline critically depends on this.  To populate with visit labels, run tools/populate_visit_windows.php - or manually insert study-specific information:
```
driusan: I think this is a little misleading.. the populate_visit_windows was a one-time script to populate it based on the old config.xml settings. It can't pull the values out of nowhere, and most new people reading this shouldn't have to run this (it should have been in the release notes of whichever release included it.)

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
```diff
@@ -0,0 +1,44 @@
...
+
+```sql
+INSERT INTO Visit_Windows (Visit_label,  WindowMinDays, WindowMaxDays, OptimumMinDays, OptimumMaxDays, WindowMidpointDays) VALUES ('V1', '0', '100', '40', '60', '50');
```
driusan: This document should probably describe what the Visit_Windows is used for a little more before telling people to populate it.

It's used to flag the Candidate Age as an error/warning when an instrument is administered, with "Optimum" determining when the study protocol says it's supposed to be done, and "Window" being when it becomes an error.

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
driusan: I just realized it says to configure the config.xml below.. in that case, this should say that it requires the config.xml be set up (as below) before running the script (and probably be moved to after that part of the document)

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
```diff
@@ -0,0 +1,44 @@
...
+The **Visit** designates the instance in time where a candidate has visited or contacted a site and underwent any sort of data collection.
+
+**Visit** and **Timepoint** are terms used interchangeably in LORIS. **Visits** are defined in the `Visit_Windows` table of the database.
+
+## Adding Options
+
+### Front End
+ _not yet available_
+
+### SQL (Recommended)
```
cmadjar: this may be nitpicky but it seems weird to put (Recommended) here when this is actually the only option at the moment to add visit options ;)

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
```diff
@@ -0,0 +1,44 @@
+# Visits
+
+## Overview
+The **Visit** designates the instance in time where a candidate has visited or contacted a site and underwent any sort of data collection.
+
+**Visit** and **Timepoint** are terms used interchangeably in LORIS. **Visits** are defined in the `Visit_Windows` table of the database.
+
+## Adding Options
```
cmadjar: I find this title confusing. I am guessing that you mean to talk about the possible ways to add a Visit or Timepoint type in LORIS?

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
cmadjar: I agree with @driusan. Would be best to mention how to populate `config.xml` first and then ask them to run the script `tools/populate_visit_windows.php`.

Also, there should be back tick around `tools/populate_visit_windows.php`

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
cmadjar: couldn't those field be simply set to `NULL`? Then it is clear that there is no restriction on the age?

Also, it might be good to explain what each of those fields stand for or are supposed to represent as I am not sure what is the difference between  `WindowMinDays ` and `OptimumMinDays ` for example. I would expect that someone new to LORIS would not know either.

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
PapillonMcGill:
```suggestion
This table must be populated with visit labels - the Imaging Pipeline critically depends on this.  To populate the table with visit labels, run tools/populate_visit_windows.php - or manually insert study-specific information:
```

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
```diff
@@ -0,0 +1,44 @@
+# Visits
+
+## Overview
+The **Visit** designates the instance in time where a candidate has visited or contacted a site and underwent any sort of data collection.
+
+**Visit** and **Timepoint** are terms used interchangeably in LORIS. **Visits** are defined in the `Visit_Windows` table of the database.
```
me-evans: Is there any chance we could change all mentions of **visit** to **timepoint** for consistency?  Is there any particular reason to be using both terms (other than we switched at some point and didn't change all) ?

Some projects (ex. one of CAP's data contributors) are creating **timepoints** just for surveys, which makes **visit** misleading and potentially confusing.  And there may be other different milestones used to classify a data collection; **timepoints** is more flexible.

_(Side Note: Perhaps the practice of creating separate timepoints for surveys could be resolved in a different way, like frontend labels/separations of instrument type within **behavioural forms** as discussed @ridz1208 ?  Although they may need to be able to set the windows according to a different **timepoint date**... TBD)_

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
me-evans: agreed - although not as concise, **Options for Adding Timepoints** would be much clearer

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
me-evans: Agreed the definition/function of windows needs to be clarified before we ask users to set them.  As discussed with @ridz1208 this week, there have been several conversations about this in the last year-ish, there's always confusion.  

See attached photos of what user sees on frontend/relevant notes.  I need help determining exactly what these windows do (someone to check the code) so I can write clear explanation to insert in this Wiki & propose frontend GUI changes to make this feature clear to users (i.e. label tweaks for clarity, better display of warning when in instrument_list, etc.). 

config_windows
windows_instrument_view_test-loris

Also, if I am correct in assuming how feature is meant to work, it appears to be broken on test. loris.

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
ridz1208: will set them all to null here 

As for the specific explanation of the fields, same as for other PRs, I'm not comfortable adding them in here as they will most probably not be maintained. either a separate document or a front end would have that

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
ridz1208: i will remove it completely I dont know why we would first populate config.xml and then not use it lol

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
ridz1208: @cmadjar what I mean by "Adding options" is: 
how to add a site, add a subproject, add... whatever. So adding site options, adding visit options... If you can think of other titles let me know, I'll change it for all the other docs

I would like to have the same title in all documents (ie not have add Visits, add subprojects, ...)
alternatives:
 - Adding to Database
 - How to Add?
 - Creating New
 - ????

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
ridz1208: I understand but I wanna make it consistent with the other docs that do have more then one option...

ridz1208:
@cmadjar @driusan I refactored a bit, please re-review

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Visits.md
PapillonMcGill: @ridz1208  Still need to add 'the table' in the last sentence.

ridz1208: sorry, should be done now

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Timepoints.md
```diff
@@ -0,0 +1,52 @@
...
+INSERT INTO visit (VisitName) VALUES ('V1');
+```
+
+### API
+ _not yet available. See [API documentation](../../../API/) for latest additions_
+ 
+## Interaction With LORIS
+
+### Subprojects
+ **Timepoints** Should be assigned to subprojects in order to be able to create timepoints for candidates. This association should be defined in the `config.xml` file of the `%LORIS_ROOT%/project/` directory as follows
```
cmadjar: 'should' instead of 'Should' (replace upper case S by lower case s)

cmadjar: also, add `:` after follows for consistency

docs/wiki/01 - STUDY PARAMETERS SETUP/01 - Study Variables/05 - Timepoints.md
```diff
@@ -0,0 +1,52 @@
...
+
+### SQL
+
+The `Visit_Windows` table lists the timepoints in the study and adds some study-specific age ranges defined for each timepoint. This table must be populated with visit labels for the proper functioning of LORIS. To populate the table with visit labels insert study-specific information as follows:
+
+```sql
+INSERT INTO Visit_Windows (Visit_label,  WindowMinDays, WindowMaxDays, OptimumMinDays, OptimumMaxDays, WindowMidpointDays) VALUES ('V1', 0, 100, 40, 60, 50);
+```
+
+If age is not a critical factor in study timepoint scheduling, define `Min` and `Max` values as `NULL`.
```
cmadjar: lol. Here they should be set to `NULL` and in the battery table they should be set to `0`. Very consistent ;).

docs/wiki/99 - Developers/SQL Dictionary.md
```diff
@@ -9,17 +9,41 @@
...
+	| `Study_site` | Does this site recruit candidates?               | This field will affect the appearance of the center in dropdowns in several modules of LORIS |
+
+- Table: `visit`
+
+ 	*This table stores the list of timepoints configured for the study.*
+ 	
+ 	|     Field    |          Description          |                                    Notes                                    |
+	|:------------:|:------------------------------------------------:|:--------------------------------------------------------:|
+	| `VisitID`   | Identifier of the Timepoint    | Avoid setting this field explicitly when inserting data, it auto increments.|
+	| `VisitName` | Back end name of the Timepoint | This name is used in file and imaging uploads                               |
```
cmadjar: is it the same as the Visit label?

docs/wiki/99 - Developers/SQL Dictionary.md
```diff
@@ -9,17 +9,41 @@
...
+	| `Visit_label `       | Name of the visit                                                                                         |  |
+	| `WindowMinDays `     | Candidate's minimum age in days for visit to be flagged as _within **permitted** parameters of the study_ | The only effect of this field is a YES/NO flag showing up on the instrument_list page|
+	| `WindowMaxDays `     | Candidate's maximum age in days for visit to be flagged as _within **permitted** parameters of the study_ | The only effect of this field is a YES/NO flag showing up on the instrument_list page|
+	| `OptimumMinDays `    | Candidate's minimum age in days for visit to be flagged as _within **optimal** parameters of the study_   | The only effect of this field is a YES/NO flag showing up on the instrument_list page|
+	| `OptimumMaxDays `    | Candidate's maximum age in days for visit to be flagged as _within **optimal** parameters of the study_   | The only effect of this field is a YES/NO flag showing up on the instrument_list page|
```
cmadjar: I don't understand, what is the difference between `WindowMinDays` and `OptimumMinDays`? According to the description of both, they do the same thing?

Same question for Max days.

docs/wiki/99 - Developers/SQL Dictionary.md
ridz1208: yes new nomenclature I could add that in the notes

docs/wiki/99 - Developers/SQL Dictionary.md
ridz1208: yes, the only difference is one gets a "optimal" flag, the other "permitted" flag

@driusan how wrong am i ?

driusan: "Permitted" means it needs to be in that range (ie. if it's not it shouldn't be in the study). "Optimal" means it should be in that range (and if it's not, there's a warning on the instrument.) They're used for the Window Difference on an instrument, with things outside "Permitted" being highlighted in red as an error.

ridz1208: @driusan but as far as the software is concerned, they are just flags, there is no real effect of being within permitted or optimal right ?

johnsaigle:
Should go to minor after #5305 merged

PR #4312, opened 2019-02-07
PUSH FORWARD OF THE 20.2.0 RELEASE TO MINOR

ridz1208:
This PR should not be squashed


driusan:
Shouldn't this be based on bugfix, not based on 20.2.0?

ridz1208:
@driusan yeah that would save us a merge commit

ill close this

PR #4313, opened 2019-02-07
PUSH FORWARD OF THE 20.2.0 RELEASE TO MAJOR

ridz1208:
This PR should not be squashed


driusan:
Same comment as the other.. shouldn't this be based on minor?

ridz1208:
@driusan yeah that would save us a merge commit

ill close this

Issue #4197, opened 2018-12-12
nifti files are stored under mincpath


driusan:
I noticed while reviewing #4096 that .nii and .nii.gz files are both using the mincPath configuration setting for the location to store them.

This makes the configuration name confusing and should probably be cleaned up.

cmadjar:
@driusan Looking at the Config paths, there is a lot of redundancy and I think this could be cleaned up: `mincPath` = `data` = `imagePath`

- `imagePath` used by `get_file.php`, tool `generate_project_statistics_csv.php`, `php/installer/Installer.class.inc`, `mri_violations.class.inc`, API 0.0.2 and 0.0.3-dev (`candidates/visits/images/format/Thumbnail.php`) and `imaging_install.sh`
- `data` does not appear to be used anywhere in the LORIS code...
- `mincPath` used by `php/installer/Installer.class.inc`, brainbrowser `image.php`, `get_file.php`, API 0.0.2 and 0.0.3-dev (`candidates/visits/images/Image.php`) and `imaging_install.sh`

Will bring it up at the next Imaging meeting so we can decide as a group what to do with it but we should probably just use the `imagePath` one.

cmadjar:
Discussed during the LORIS-MRI meeting of January 18th. Will keep only the `dataDirBasepath` field in the Config module

PR #4314, opened 2019-02-08
Cleanup duplicated data config paths

cmadjar:
### Brief summary of changes

This PR cleans up all the different paths that are referring to the same path (a.k.a. `/data/%PROJECT%/data/`. This gets rid of the Config settings `mincPath` and `data` and updates the code accordingly.

The code changes on the LORIS-MRI front can be found in https://github.com/aces/Loris-MRI/pull/399

### This resolves issue...

- [x] Github #4197 and LORIS-MRI https://github.com/aces/Loris-MRI/issues/367

### To test this change...

- [ ] run the SQL patch
- [ ] make sure the imaging modules work properly and are able to see the images
- [ ] test the API v0.0.2 and v0.0.3-dev to make sure the images can be found and read

### Caveat for existing projects

Make sure that your Config setting `dataDirBasepath` and `imagePath` are configured to the proper data path (typically `/data/%PROJECT%/data`). Note that `dataDirBasepath` is the path used by the LORIS-MRI pipelines and `imagePath` is the path where the imaging data can be read by the web server.

### DIFF
--- a/SQL/0000-00-03-ConfigTables.sql
+++ b/SQL/0000-00-03-ConfigTables.sql
@@ -74,9 +74,7 @@ INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType,
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, Label, OrderNumber) VALUES ('paths', 'Specify directories where LORIS-related files are stored or created. Take care when editing these fields as changing them incorrectly can cause certain modules to lose functionality.', 1, 0, 'Paths', 2);
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'imagePath', 'Path to images for display in Imaging Browser (e.g. /data/$project/data/) ', 1, 0, 'text', ID, 'Images', 9 FROM ConfigSettings WHERE Name="paths";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'base', 'The base filesystem path where LORIS is installed', 1, 0, 'text', ID, 'Base', 1 FROM ConfigSettings WHERE Name="paths";
-INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'data', 'Path to main imaging data directory (e.g. /data/$project/data/) ', 1, 0, 'text', ID, 'Imaging data', 5 FROM ConfigSettings WHERE Name="paths";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'extLibs', 'Path to external libraries', 1, 0, 'text', ID, 'External libraries', 3 FROM ConfigSettings WHERE Name="paths";
-INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'mincPath', 'Path to MINC files (e.g. /data/$project/data/)', 1, 0, 'text', ID, 'MINC files', 8 FROM ConfigSettings WHERE Name="paths";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'DownloadPath', 'Where files are downloaded', 1, 0, 'text', ID, 'Downloads', 4 FROM ConfigSettings WHERE Name="paths";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'log', 'Path to logs (relative path starting from /var/www/$projectname)', 1, 0, 'text', ID, 'Logs', 2 FROM ConfigSettings WHERE Name="paths";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'MRICodePath', 'Path to directory where Loris-MRI (git) code is installed', 1, 0, 'text', ID, 'LORIS-MRI code', 6 FROM ConfigSettings WHERE Name="paths";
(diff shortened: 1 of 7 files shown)


SQL/0000-00-03-ConfigTables.sql
```diff
@@ -133,7 +130,7 @@ INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType,
 
 -- Loris-MRI/Imaging Pipeline options from the $profile (commonly "prod") file
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, Label, OrderNumber) VALUES ('imaging_pipeline', 'Imaging Pipeline settings', 1, 0, 'Imaging Pipeline', 12);
-INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'dataDirBasepath', 'Base Path to the data directory of Loris-MRI', 1, 0, 'text', ID, 'Loris-MRI Data Directory', 1 FROM ConfigSettings WHERE Name="imaging_pipeline";
+INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'dataDirBasepath', 'Base Path to the data ' || 'directory of Loris-MRI', 1, 0, 'text', ID, 'Loris-MRI Data Directory', 1 FROM ConfigSettings WHERE Name="Paths";
```
PapillonMcGill: @cmadjar Are you sure about the `... data ' || 'directory ...`?  This will result in a logical comparison between 2 strings.

SQL/New_patches/2019-02-08_cleanup_duplicated_data_path_in_Config.sql
```diff
@@ -0,0 +1,21 @@
...
+
+-- Remove entries for mincPath, data and imagePath from the ConfigSettings table
+DELETE
+FROM ConfigSettings
+WHERE Name IN ('mincPath', 'data', 'imagePath');
+
+-- Update the parent of dataDirBasePath to be in the Path section instead of the
+-- imaging pipeline section of the Config module
+UPDATE ConfigSettings
+SET Parent=(SELECT ID FROM (SELECT * FROM ConfigSettings) AS X WHERE Name='Paths')
```
PapillonMcGill:
```suggestion
SET Parent=(SELECT ID FROM (SELECT ID FROM ConfigSettings) AS X WHERE Name='Paths')
```

SQL/0000-00-03-ConfigTables.sql
driusan: The description of this should specify whether this is relative to the imaging server or the web server (and the type will need to be changed once #4023 is merged)

johnsaigle:
Travis is failing because of exception handling from this PR https://github.com/aces/Loris/pull/4165/files. If the settings there have been deleted, make sure to remove them from the checks. That should satisfy Travis. :)

SQL/0000-00-03-ConfigTables.sql
cmadjar: oups. The editor added this because the line was too big and I forgot to remove the extra ' || ' it added when separating the line in multiple. Sorry...

cmadjar:
@PapillonMcGill Thanks! I fixed your comments

SQL/0000-00-03-ConfigTables.sql
cmadjar: At the last imaging meeting, we discussed we would only keep one setting so technically this one would work for both. We can bring it up again at the next imaging meeting. I'll add it to the agenda.
Thanks!

cmadjar:
@johnsaigle I am confused. I have not touched the config settings mentioned in https://github.com/aces/Loris/pull/4165/files so there should not be an issue. I only removed a few path config settings and cannot grep in the code for the old config settings so I am not sure where it gets it from. 
Travis started again so I cannot look for the error message. :S. Will check again later.

johnsaigle:
@cmadjar Hm yeah I see your point. That function will catch any ConfigurationException that occurs so maybe there is something else going on?

SQL/0000-00-03-ConfigTables.sql
driusan: But it can't work for both imaging pipeline and web server, the paths are different if they're mounted on different places on the different servers

cmadjar: @driusan in the next commit I reintegrated the imagePath for the apache server and kept the dataDirBasepath for the LORIS-MRI server. Please re-review

driusan:
I don't see how this is going to work when the web server and imaging pipeline server don't have the same data dir mounted in the same location.

Don't we need 2 settings, one relative to each?

cmadjar:
@driusan I did the change you requested  and there is indeed two paths remaining: one for the webserver and one for the imaging pipeline. 

I forgot to update the description of the PR and apparently to add again the `imagePath` into `SQL/0000-00-03-ConfigTables.sql` . Should be fixed now. Please re-review. Thanks!

driusan:
@cmadjar sorry, I'm still confused.. I'm just looking at the diff in `SQL/0000-00-03-ConfigTables.sql` and I only see one path being added for 'dataDirBasepath'. What's the other one?

SQL/0000-00-03-ConfigTables.sql
```diff
@@ -230,7 +226,7 @@ INSERT INTO Config (ConfigID, Value) SELECT ID, "Produced by LorisDB" FROM Confi
 INSERT INTO Config (ConfigID, Value) SELECT ID, "S3cret" FROM ConfigSettings WHERE Name="JWTKey";
 
 
-INSERT INTO Config (ConfigID, Value) SELECT ID, "/PATH/TO/DATA/location" FROM ConfigSettings cs WHERE cs.Name="Loris-MRI Data Directory";
+INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/%PROJECTNAME%/data/" FROM ConfigSettings cs WHERE cs.Name="dataDirBasepath";
```
cmadjar: @driusan in this line, I modified the default path of dataDirBasepath but I did not add it (the line in red above is for the same config setting)

SQL/0000-00-03-ConfigTables.sql
```diff
@@ -187,9 +185,7 @@ INSERT INTO Config (ConfigID, Value) SELECT ID, "YMd" FROM ConfigSettings WHERE
 
 INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/%PROJECTNAME%/data/" FROM ConfigSettings WHERE Name="imagePath";
```
cmadjar: imagePath is in this INSERT statement. I did not need to modify the default path in it which is why it does not show up in the diff, but the INSERT statement is still here :)

cmadjar:
@driusan @PapillonMcGill Could you approve the PR if you are happy with the changes please? Thank you :)

cmadjar:
@driusan @PapillonMcGill I rebased the branch to major. Please re-review at your convenience :)

PR #4317, opened 2019-02-08
[JSX] Refactor SearchableDropdown to not use document.querySelector

davidblader:
### Brief summary of changes
this refactors the SearchableDropdown to avoid reading directly from the DOM by relying on an additional prop value

### This resolves issue...
CCNA was having fatal JS errors on some forms due to reliance on document.querySelector

### To test this change...

- [ ] Make sure media's uploadForm still works, specifically for Site input


davidblader:
@driusan @ridz1208  if possible i'd like to rebase this to be on bugfix, if that's OK / if there's a bugfix release planned before a minor release

jsx/Form.js
```diff
@@ -168,6 +167,7 @@ class SearchableDropdown extends React.Component {
     if (!this.props.strictSearch && value === undefined) {
       value = e.target.value;
     }
+    this.props.onUserInput(this.props.name + 'Temp', e.target.value);
```
driusan: I don't understand what this "Temp" thing is all about in this PR.

driusan:
If it's a bug fix, I don't see any issue with it going to bugfix..

jsx/Form.js
davidblader: there needs to be an intermediary "temporary" value for the user's input to be stored in as they're typing. couldn't think of a better naming convention. open to any suggestions

PR #4318, opened 2019-02-08
[JSX] Refactor SearchableDropdown to not use document.querySelector

davidblader:
### Brief summary of changes

this refactors the SearchableDropdown to avoid reading directly from the DOM by adding the user's live text input to the component's state
### This resolves issue...

CCNA was having fatal JS errors on some forms due to reliance on document.querySelector

### To test this change...

- [ ] Make sure media's uploadForm still works, specifically for Site input


jsx/Form.js
```diff
@@ -178,16 +178,12 @@ class SearchableDropdown extends React.Component {
       let options = this.props.options;
       if (Object.values(options).indexOf(value) === -1) {
         // empty string out both the hidden value as well as the input text
-        document.querySelector(`input[name="${this.props.name + '_input'}"]`).value = '';
+        this.props.onUserInput(this.props.name + 'Temp', '');
```
davidblader: > I don't understand what this "Temp" thing is all about in this PR.
- dave

davidblader: there needs to be an intermediary "temporary" value for the user's input to be stored in as they're typing. couldn't think of a better naming convention. open to any suggestions

driusan: I still don't really understand.. why can't it just use state?

davidblader: i think it's simpler to keep it stateless and to just rely on the user component's state, since this tends to be the norm currently for our form components.
this way also fixes another issue we were having on CCNA: in media, after a successful upload `this.state.formData` should clear out, but because the text of the input was not kept in `formData`, the text inside the Site input would remain.

HenriRabalais: I agree that Form.js components should generally remain stateless, but maybe this should be an exception? I'm not really sure that the user component's state should be holding on to this temp value. Is the implication that the temp value will be held in the parent components `formData` (or equivalent) object? It also seems like this solution necessitates the parent component to pass both a value and temp value down as a prop. If the parent component doesn't need this information, I would opt for it to be held in state in this component.

davidblader: > I agree that Form.js components should generally remain stateless, but maybe this should be an exception? I'm not really sure that the user component's state should be holding on to this temp value. Is the implication that the temp value will be held in the parent components `formData` (or equivalent) object? It also seems like this solution necessitates the parent component to pass both a value and temp value down as a prop. If the parent component doesn't need this information, I would opt for it to be held in state in this component.

you have a good point. the only benefit of keeping it in the parent component's state would be, for example, in media's upload form, upon successful upload formData gets cleared out, but the text inside the input would remain. i'll tinker around with keeping it in `SearchableDropdown`'s state



> @davidblader @driusan @HenriRabalais Why not just set a default value for the this.props.name? You could set the default value to be 'Temp' and not need to add the + 'Temp' which looks ugly.

can you elaborate? `this.props.name` is meant to be the key in the formData object that gets passed to the server. i'm not sure we want to mess around with that.

driusan: I added the "Needs Work" tag until this is sorted out.. (it's not the most appropriate tag, but I couldn't think of a better one..)

maltheism: @davidblader Oh I see now. I guess another prop could be made to replace the string and just set the defaultProps for the new prop to Temp. I'm not sure now.

ridz1208:
@HenriRabalais @driusan @maltheism please re-review/test asap

driusan:
Looks fine to me now if the other people who reviewed it agree.

maltheism:
Hi @davidblader @ridz1208 @driusan  I pulled the PR and I'm using a new database that's using the RBdata.sql and to generate sites for testing these changes with the SearchableDropdown. The only module in Loris using the SearchableDropdown appears to be the Clinical->Media module for uploading. 

Here are the results:
Chrome is working great!
Firefox not so great. The user has to double click on the input for sites to have all the sites show. Otherwise they click the box and nothing shows but it does work once you type a letter of one of the sites.
Safari doesn't work at all. I know it's not supported by Loris but thought to let you know. I use safari and typically everything works and so I was confused at first.

davidblader:
> Safari doesn't work at all. I know it's not supported by Loris but thought to let you know. I use safari and typically everything works and so I was confused at first.

yeah, safari doesn't support `<datalist>` unfortunately :disappointed:

driusan:
@HenriRabalais

PR #4320, opened 2019-02-11
[Core] Remove checkDate() function from Utility

johnsaigle:
### Brief summary of changes

Remove the `_checkDate()` function from Utility and delete instances of calling code.

#### Justification
* The function was declared to accept only strings and generated no bugs -- therefore it was never used in the `array` context
* The `empty()` checks at the beginning of the function always return true on a string input, therefore the function would always short circuit and not validate dates properly
* The calling code pass strings, never arrays in "QuickForm array format"
* The calling code in `next_stage` and `new_profile` are generated using JavaScript date picker elements and so there should never be an occasion where a date is manually entered by a user and require this kind of validation.

### DIFF
--- a/modules/new_profile/php/new_profile.class.inc
+++ b/modules/new_profile/php/new_profile.class.inc
@@ -239,10 +239,6 @@ class New_Profile extends \NDB_Form
         if ($config->getSetting('useEDC') == "true") {
             if (empty($values['edc1'])) {
                 $errors['edc1'] = 'Expected Date of Confinement field is required';
-            } else if (strlen(implode($values['edc1'])) > 2
-                && !\Utility::_checkDate($values['edc1'])
-            ) {
-                $errors['edc1'] = 'EDC is not a valid date';
             }
 
             if (empty($values['edc2'])) {
@@ -250,7 +246,7 @@ class New_Profile extends \NDB_Form
             }
 
             if ($values['edc1'] != $values['edc2']) {
-                $errors['edc1'] = 'Estimated Due date fields must match.';
+                $errors['edc1'] = 'Estimated due date fields must match.';
             }
         }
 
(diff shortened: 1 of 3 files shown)


maltheism:
@johnsaigle lgtm, My only question is there really no point in validating the date string on the backend?

johnsaigle:
@maltheism I could replace the calls to our built-in function with PHP's `checkdate()` if that's better.

My point is that we haven't really been doing date checking if this function was broken anyway.

maltheism:
@johnsaigle Yah I see your point and maybe checkdate() should be added in another PR.

johnsaigle:
@driusan Would you be able to merge this one when you get a chance? It's ready to go and is blocking another PR.

driusan:
I don't think the assumptions made in this PR are correct.

Utility::_checkdate is used by instruments, and where it is used LorisForm converts the format of the POSTed data from a string to an array of the format that was used by QuickForm. `<input>` elements also fall back to "text" type if the type (such as date) isn't supported, and frontend validation isn't sufficient for validating user input. 

It can not be replaced by PHP's `checkdate` since they do different things.

@sruthymathew123 @ridz1208 can you check if any instrument's date handling is broken with this PR on instruments you have that might depend on this function?

sruthymathew123:
@driusan I couldn't find any IBIS instruments using this function right now.

ridz1208:
@driusan @johnsaigle I looked through project code and no instance of this function remain.

HOWEVER should'nt we be fixing this rather then removing it ?

> The calling code in `next_stage` and `new_profile` are generated using JavaScript date picker elements and so there should never be an occasion where a date is manually entered by a user and require this kind of validation.

It's still good practice to validate back-end even if we assume front-end is already doing the job... plus given the variability in support of these things across browsers (even the one we don't support because users use them anyways...) and the history of issues we have had with dates in the past (2015-2017) I dont see harm in having a date checker before submission of any date.

That said it doesn't have to be this function we could def make something cleaner and more configurable later one

johnsaigle:
@ridz1208 Thanks for checking.

Yeah it's probably good to check dates in some way but I think we'd be better served using `checkdate()` or `DateTime` objects rather than this function.

driusan:
If I was mistaken about it being required for project's instruments, I agree that there's no point in keeping the function around.

PR #4323, opened 2019-02-11
[Core] Fix TypeError in Notifier

johnsaigle:
### Brief summary of changes

A previous PR introduced a TypeError issue by using the PHPDoc return type as two functions' return type when the functions were not annotated properly. 

This PR removes one function altogether as it acted as a wrapper for \User which can be directly accessed. The other was changed to use the array type instead of the string type.

### DIFF
--- a/php/libraries/NDB_Notifier.class.inc
+++ b/php/libraries/NDB_Notifier.class.inc
@@ -86,16 +86,16 @@ class NDB_Notifier extends NDB_Notifier_Abstract
     {
         foreach ($this->notified['email_text'] as $name=>$uid) {
             $notifierID = $this->notifier->getData('ID');
-            $fromEmail  = $this->notifier->getData('Email');
+            $senderEmail  = $this->notifier->getData('Email');
             $this->tpl_data['notified_user'] = $name;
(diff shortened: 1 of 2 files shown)


php/libraries/NDB_Notifier.class.inc
```diff
@@ -85,20 +85,21 @@ class NDB_Notifier extends NDB_Notifier_Abstract
...
+            $this->tpl_data['notifier_user'] = "$notifierID <$fromEmail>";
+            $toEmail = $this->getEmail($uid);
 
-            if ($this->notifier['ID'] === $uid) {
+            if ($notifierID === $uid) {
                 // TODO send confirmation email instead
             } else {
                 Email::send(
-                    $email,
+                    $toEmail,
```
maltheism: Everything looks great! My only suggestion is maybe replacing $toEmail with $replyEmail but that's just my preference and $toEmail works.

maltheism: or maybe even $replyEmailAddress

johnsaigle: It's actually the other way around. The toEmail is the recipient and the fromEmail is the sender. I'll change the variable names though. You're right, they could be better.

PR #3355, opened 2017-12-14
[Instruments][Docs] JSON Instrument Spec Sheet

ZainVirani:
I've added comments at points I feel further discussion is needed. They will be removed as they are addressed.

The file location is up for discussion as well, I just added details on top of the existing JSON spec sheet in /docs.

For elements from the original spec sheet that have not been implemented, I left Dave's original writeup but marked them as not implemented. I can either implement them if needed or remove them entirely before merge.

PR #4324, opened 2019-02-12
[Instruments][Docs] Z-JSON Instrument Proposal

zaliqarosli:
### Brief summary of changes

**THIS PR IS NO LONGER OPEN FOR REVIEW. IT IS REPLACED BY A COMMENTABLE [GOOGLE DOC](https://drive.google.com/open?id=1IqHCBM4yW5fpPqZeKvjhtyS59H6ZIll8_op-DTWwwUo)**

This PR includes the specifications for a new JSON Instrument structure, along with a template, an example (bmi instrument), and a doc for the LORIS Logic Parser syntax.

replaces https://github.com/aces/Loris/pull/3355


docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+While instrument `fields` have corresponding `data` associated with it upon form submission, a `helper` is rendered on the front-end as an integral element that however does not collect data. There are
+5 types of helpers: the static element, and 4 other layout-specific formatting elements, group, row, table, and section.
+
+## 4.1: Helper Types
+
+### 4.1.1: Static
+
+A `helper` of type "static" represents headings, sub-headings, labels, or informational text that is required in
+an instrument. These can be rendered on the front-end as a header, static, or link element.
+Static helpers are some text displayed to provide information and without an accompanying data input.
```
driusan: I don't understand why these are separated in the JSON object. It seems to primarily be for grouping them for human comprehension, but they can be separated in the document into different sections without separating them in the object and they have the same structure and requirements (ie. global uniqueness, references in setup, etc..)

Any technical reasons to separate them require parsing the object anyways, so whether they're "helper" or "data" types can easily be inferred from the type attribute.

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+Static helpers do not have an `options` key.
+
+```js
+{
+    "schema" : {
+        ...
+        "helpers" : {
+            "nameOfHelper" : {
+                "type": "static",
```
driusan: What does a "static" type mean?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+                    ...
+                }
+        ...
+}
+```
+
+### 4.1.2: Group
+
+A helper of type "group" denotes a horizontal collection of elements which should be displayed
+together and is separated by a delimiter.
```
driusan: Do multiple rows in a group have the same spacing? (ie. like a table.)

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+
+They often have rules which work together and may be interdependent. Groups have the following form:
```
driusan: "interdependent" how?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+```
+
+`options.order`: Required array of strings.
+                 An array of the names of `fields` or `helpers`, in order, that are to be formatted
+                 within this group. It may contain any field, but only helpers of the type "static".
+ 
+### 4.1.3: Row
+
+A helper of type "row" denotes a row in a table. If rows are not in a table, they should be rendered
+as a "group" helper.
```
driusan: In that case, why have "group" helpers and not just define groups as rows that aren't embedded in a table? (or vice versa)?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+                    ...
+                }
+        ...
+}
+```
+`options.colSize` : Required int.
+                    The number of columns in the table. 
+
+`options.rowSize` : Required int.
+                    The number of rows in the table.
```
driusan: Why is this required?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+                    "showAsFooter": boolean,
+                },
+                "rules": {
+                    ...
+                }
+        ...
+}
+```
+`options.colSize` : Required int.
+                    The number of columns in the table.
```
driusan: Can't this be inferred?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+                    The number of columns in the table. 
+
+`options.rowSize` : Required int.
+                    The number of rows in the table.
+
+`options.order`: Required array of strings.
+                 An array of the names of `row` helpers, in order, that are to be formatted
+                 within this table.
+
+The size of this array must equal `options.rowSize`.
```
driusan: Why not just infer rowSize from the size of this array?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+
+At each index (or column) of every "row" element, the `helper` or `field` that is
+referenced inside that index has to have its `description` match. This description
+will be rendered as the column header if `options.showColHeaders` is true.
```
driusan: Is there going to be some kind of validator to validate rules like this?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+`options.showDesc`: Boolean. Default is true.
+                    True or false that the table's `description` is displayed.
+
+`options.showAsFooter`: Boolean. Default is false.
+                        True or false that the description is displayed as a footer, instead of a
+                        header.
+
+### 4.1.5: Section
+
+A helper of type "section" is a vertical grouping of elements whose `rules` can affect its children's rules.
```
driusan: I'm not sure what you mean by "affect its children's rules".. affect them how?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+}
+```
+
+`options.order`: Required array of strings. An array of the names of `fields` or `helpers`, in order, that are to be
+                 formatted within this group. It may contain any field, but only helpers of the type "static".
+
+`options.showDesc`: Boolean. Default is true. True or false that the section's `description` is displayed as the header for
+                    section of elements.
+
+An implementation could choose to set this key to false, and display an independent "static" helper as a header instead.
```
driusan: Why would/wouldn't an implementation do that? It seems like there's value in requiring it so that more semantic meaning can be extracted from the string

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+ 
+# 5.0: UI Properties
+
+The UI component of the Z-JSON Instrument contains information on how we render our `fields` and `helpers` objects
+on the front-end. It is composed of JSON objects whose keys match those of the field or helper object it's describing.
+This component should only make decisions on which HTML or HTML Form element those objects get rendered as,
```
driusan: What if it's not rendered to HTML?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+```js
+{
+    "ui": {
+        "nameOfField" : {
+            "type" : "select"
+        }
+    }
+}
+```
+
```
driusan: I think all of the above need descriptions of what they mean.. I'd also call "select"s "dropdown"s, because that's less specific to HTML.

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+            "type" : "textarea",
+            "options": {
+                "placeholder": string
+            }
+        }
+    }
+}
+```
+
+`options.placeholder`: String. Text that is rendered inside the text element as a placeholder before user input.
```
driusan: I don't see any reason to separate text and textarea, and not just make them options for the same type.

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+            "options": {
+                "level": int
+            }
+        }
+    }
+}
+```
+
+`options.level`: Int. The number indicating the level of the html heading element to be rendered i.e. h1, the highest
+and most prominent, and h6, the lowest and reserved for subheadings.
```
driusan: Why link the levels specifically HTML? Are they required to be consecutive? I think it would be better to require the level to be semantic and just say it's recommended that it not go more than 6 levels deep to make it easier to map to html.

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+
+```js
+{
+    "ui": {
+        "nameOfHelper" : {
+            "type" : "link",
+        }
+    }
+}
+```
```
driusan: Where is the URL for the link type defined?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+
+```js
+{
+    "ui": {
+        "nameOfHelper" : {
+            "type" : "static"
+        }
+    }
+}
+```
```
driusan: Still not sure what a "static element" means in this spec..

It would probably be a good idea to include an "image" type as well.

docs/instruments/ZJSONInstrument_template.json
```diff
@@ -0,0 +1,82 @@
...
+      "instrumentVersion": "",
+      "lastUpdateOn": "",
+      "longName": {
+          "en-CA": "",
+          "fr-CA": ""
+      },
+      "multilingual": true,
+      "previousVersion": "",
+      "schemaVersion": "",
+      "shortName": ""
```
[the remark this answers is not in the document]
driusan: The description in the spec

> The unique name and identifier of the instrument, also the test name in the
database.

implies that the shortName is intended for backend purposes.. so I'm not sure that it should be multilingual.

Maybe it would be a good idea to explicitly name it something like "backendName"

driusan:
Suggestion: put this in a Google Doc and use that for collabouratively editing it until it's more stable, since GitHub isn't a great way to review/suggest changes for large documents.

docs/instruments/ZJSONInstrument_v0.0.1.md
xlecours: And probably File as well ?

driusan: I think files would be field types and not UI types, but that's a good idea..

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+- [1.0: Introduction](#10-introduction)
+- [2.0: Schema Overview](#20-schema-overview)
+- [3.0: Field Elements](#30-field-elements)
+- [4.0: Helper Elements](#40-helper-elements)
+- [5.0: UI Properties](#50-ui-properties)
+
+# 1.0: Introduction
+
+The Z-JSON Instrument is comprised of 3 components: the core schema, an accompanying UI configuration, and the
+instrument data.
```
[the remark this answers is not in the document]
ridz1208: I think it's a good idea however to mention that the data is handled seperately. like dont erase all traces of data, explicitely mention that its separate

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+        "meta" : {
+            "createdBy": string,
+            "createdOn": string,
+            "createdWith": string,
+            "defaultLanguage": string,
+            "derivedFrom": string,
+            "importedFrom": string,
+            "includeMetaDataFields": boolean,
+            "instrumentVersion": string,
+            "lastUpdateOn": string,
```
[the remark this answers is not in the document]
ridz1208: i think it's usefull for the instrument builder, that was the idea the builder would always update that field

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+instrument, and the creation of a new version does not suffice.
+                    
+This level of modification does not include format transcription i.e. linst to
+JSON, unless the content of the instrument has also been modified. The
+`meta.importedFrom` key should be used in the case of conversions.
+
+`meta.importedFrom`: String.
+
+The resource instrument that this Z-JSON Instrument was converted from, given
+that the imported content is preserved, and only its format is modified.
```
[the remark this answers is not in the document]
ridz1208: if you convert a php or JSON instrument into Z-JSON. you refer to it here I think. if you later it, you keep this null and refer to the previous version in derived from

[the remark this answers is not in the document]
driusan: How about something like:

```
The source that this instrument was converted from, if it was imported from another format and is otherwise unmodified. After any modifications to an imported instrument are made, this field should be set to the empty string.
```

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+                      ...
+                ]
+            }
+        ]
+    }
+}
+```
+
+`setup.name`: String.
+              The name and ID of the page object.
```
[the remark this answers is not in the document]
ridz1208: i think it's what is in the instrument_subtest table we have now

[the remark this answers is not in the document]
driusan: To be honest, I've never understood why we had names there either.

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+```
+
+The value of these keys are string expressions of logical formulas. These formulas, when calculated, evaluate to true
+or false on the front-end. The formulas are parsed and calculated using the Evaluator function of the LORIS Logic
+Parser, [LParse](https://gitlab.com/zainvirani/LParse). The currently supported logic operations are available in the LParse [README.md](https://gitlab.com/zainvirani/LParse/blob/master/README.md), as well as
+instructions on how to customize functions. 
+
+`rules.requireIf`: String. True or false that the field requires data input, given the condition in the string expression.
+
+`rules.displayIf`: String. True or false that the field is displayed in the instrument, given the condition in the string
```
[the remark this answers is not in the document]
ridz1208: ouuuh I'm curious too now

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+expression.
+
+`rules.disableIf`: String. True or false that the field is disabled i.e. not editable nor submittable, given the condition in
+the string expression.
+
+## 3.2: Field types
+
+There are 8 different field types: enum, string, int, decimal, date, time, boolean, and score. Each type
```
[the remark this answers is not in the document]
ridz1208: I'd say numeric not number but agree

@driusan any ideas to replace enum ? list ? set ?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+                           If true, a data input of "false" cannot be submitted, and a 'not_answered' value cannot be given.
+
+`options.readonly`: Boolean.
+                    True or false that the field is read only i.e. not editable on the front-end but
+                    submittable on save. This key is different to `rules.disableIf` in that it is not
+                    dependent on a condition, and a 'disabled' field is not submittable.
+
+### 3.2.6: Score
+
+A score field is a placeholder whose values are based on the other fields' data. A score field is not
```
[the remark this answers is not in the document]
ridz1208: a field has a value in the data array. a helper does not.

docs/instruments/ZJSONInstrument_v0.0.1.md
ridz1208: label, heading subheading ... anything text that does not get submitted

docs/instruments/ZJSONInstrument_v0.0.1.md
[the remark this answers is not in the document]
driusan: Every type has specific requirements that need to be validated/enforced in how they affect other things, I still don't see why that requirement requires a different structure for the format.

docs/instruments/ZJSONInstrument_v0.0.1.md
driusan: I would call it something like "statictext", then.

docs/instruments/ZJSONInstrument_v0.0.1.md
ridz1208: I might be able to answer that one. Due to a recent bug we found in CCNA I think its a good idea to distinguish rows which should depend on the headers of the table they belong in  and groups which are just floating elements grouped together.

the bug: because in the current implementation you define a group of headers and add them to the page and then subsequently define groups of input fields (rows in the table) and add them to the page some  `n` number of times. During a routine bugfix, 2 elements of the hearder (assume header2 and header 3) group were switched (assume header2 went to column 3 and vice versa) and that was not caught during code review, which means that since that was released users had been entering values for header 3 in column 2 and values for header 2 in column 3..... naturally this was a mess to clean up after and we thought it would be a good idea to make tables more strict (ie not adding rows as independent groups). If i remember correctly the idea was to define the header of the column of a field in each field of a row and thus in case the headers dont line up, an error would be thrown

ridz1208: see line 772

docs/instruments/ZJSONInstrument_v0.0.1.md
ridz1208: we define an other ui object

ridz1208: I think this object is meant to be the mapping from field to HTML object... did you have something else in mind ?

docs/instruments/ZJSONInstrument_v0.0.1.md
ridz1208: this goes back to the comment on line 830

for me this was where we map input types to HTML types

docs/instruments/ZJSONInstrument_v0.0.1.md
driusan: mobile app, desktop app, some theoretically successor to HTML.. if it's only for mapping to HTML, why not just use HTML directly?

docs/instruments/ZJSONInstrument_v0.0.1.md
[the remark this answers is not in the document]
driusan: why is it useful for the instrument builder? And if it depends on it, what happens if someone edits a file outside of the instrument builder, and then tries to edit it with the instrument builder?

docs/instruments/ZJSONInstrument_v0.0.1.md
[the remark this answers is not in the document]
driusan: I don't have any problems with enum being a type.. "list" seems like a UI element that displays enumerations, and "sets" are unordered.

docs/instruments/ZJSONInstrument_v0.0.1.md
driusan: There's also an argument to be made that separating them are more intuitive for people to understand, I was just wondering if any thought had went into combining them since it would simplify implementations.

docs/instruments/ZJSONInstrument_v0.0.1.md
driusan: I think this should be mapping field types to UI concepts, not to HTML types. There are other ways than HTML to build a UI, different versions of HTML, and there are multiple ways even in HTML to build the same UI concept.

docs/instruments/ZJSONInstrument_template.json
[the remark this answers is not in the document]
zaliqarosli: yes, it should be the ID so a unique string. what about "uniqueName" or "testName" as that's what we call it in LORIS anyways? i don't know if i like "backendName".

docs/instruments/ZJSONInstrument_template.json
```diff
@@ -0,0 +1,82 @@
...
+    },
+    "fields": {
+      "nameOfField" : {
+        "type": "",
+        "description": {
+            "en-CA": "",
+            "fr-CA": ""
+        },
+        "options": {
+
```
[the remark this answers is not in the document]
zaliqarosli: as discussed offline, this is empty for the purpose of a template but should be filled according to the options of whatever type of field/helper it is

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
+# Z-JSON Instrument Specification
+
+This documentation describes and explains the Z-JSON Instrument. It is supported by a template and an example,
+both of which can be found in the same directory.
+
+The Z-JSON Instrument is the JSON representation of a LORIS instrument. Describing an instrument as a JSON
```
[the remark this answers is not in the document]
zaliqarosli: thanks for the review dave! 🎉 ill make changes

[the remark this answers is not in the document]
zaliqarosli: to your point above, this spec is LORIS specific as it talks about the Loris logic parser, the loris metadata fields for instruments, and the loris api and new instrument builder. at what point are we making it too generalizable? at the moment, the only use case is for LORIS and it is the only place Z-JSON is going to be implemented, for now. should generalizing the spec come after in a future version?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+object will allow the instrument to be language-independent, human readable, and shareable across different
+platforms. This schema will become an important part of the LORIS API and the New Instrument Builder, and was built
```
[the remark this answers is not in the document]
zaliqarosli: i just copied that from the api instrument format doc, thinking it was important. but you're right!

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+to be compatible with both the [ReproNim JSON-LD Form Schema](https://github.com/ReproNim/schema-standardization)
```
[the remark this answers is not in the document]
zaliqarosli: interoperable?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+    "schema" : {
+        ...
+    },
+    "ui" : {
+        ...
+    }
+}
+```
+The core schema describes the instrument's data structure and format, while the UI component decides certain
+front-end properties. Both the schema and the UI component are required in order to have a fully loaded and
```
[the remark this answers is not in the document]
zaliqarosli: that the instrument is loaded and rendered on the front-end with no errors and with all the components of the instrument

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+            "defaultLanguage": string,
+            "derivedFrom": string,
+            "importedFrom": string,
+            "includeMetaDataFields": boolean,
+            "instrumentVersion": string,
+            "lastUpdateOn": string,
+            "longName": {
+                ...
+            },
+            "multilingual": boolean,
```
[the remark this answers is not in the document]
zaliqarosli: fair enough

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+metadata fields, such as 'Candidate Age', or 'Examiner', which
+calculates the candidate's age, and selects the examiner
+administering the instrument, respectively.  
+
+`meta.instrumentVersion`: Required string.
+
+The version of the instrument being described.
+
+It follows some convention as determined by the author of the instrument,
+starting with some representation of the number 1 and incrementing itself alongside any changes made to the `schema` object. It is self-contained and
```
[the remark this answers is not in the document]
zaliqarosli: do versions not have to increment? i put down 'representation' of the number 1, and not the number itself. so e.g. v0.0.1 would be a representation of 1. I thought the point of versioning is that you can see how far from 1 that version is and see the lineage through numbers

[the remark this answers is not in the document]
zaliqarosli: i think an implementation would need to set a standard for their version string. otherwise, could we set a standard? not sure if this would be compatible with existing versions of current instruments

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+
+`meta.previousVersion`: String.
+
+The previous version of the Z-JSON Instrument, if significant changes
+were made to justify creating this new version of the instrument.
+
+This may be redundant metadata but it may also be helpful in showing the
+lineage of existing instruments. If changes to an instrument is so
+significant that they no longer share the same lineage, `meta.derivedFrom`
+would be used to track changes.
```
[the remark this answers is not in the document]
zaliqarosli: instrument A.1 was modified to a new version instrument A.2. In this case, `previousVersion': "instrumentA.1".

instrument A.2 was modified SO much that it is now no longer the same instrument, and has actually become instrument B.1. In this case, instrument B.1 is the first of its kind and does not have a previous version but is instead `derivedFrom`: "instrumentA.2"

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+
+This may be redundant metadata but it may also be helpful in showing the
+lineage of existing instruments. If changes to an instrument is so
+significant that they no longer share the same lineage, `meta.derivedFrom`
+would be used to track changes. 
+
+`meta.schemaVersion`: Required string.
+
+The version of this Z-JSON Instrument Specification document that the
+instrument complies with. It follows the format of v*SpecVersion*-dev i.e. `v0.0.1-dev` if following this
```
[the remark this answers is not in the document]
zaliqarosli: it could be defined as semver, thats not a bad idea. i'll need some help defining the versions because i'm not totally confident in my knowledge of this

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+current version.
+
+`meta.shortName`: Required string.
+
+The unique name and identifier of the instrument, also the test name in the
+database.
+
+The name of the Z-JSON Instrument JSON file will follow the format *shortName_instrumentVersion*.
```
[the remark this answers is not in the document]
zaliqarosli: :( can we not hope?? That would be so nice. It could be enforced if the instrument is saved and output from the instrument builder no?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+    }
+}
+```
+
+`setup.name`: String.
+              The name and ID of the page object.
+
+`setup.typ`: Required string with value "page".
+
+`setup.description`: Required string. The label or title of the page that will be rendered on the front-end as a link to the
```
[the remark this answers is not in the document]
zaliqarosli: so this is my intention: if `multilingual`: false, then all `description` keys have value strings that are the description text. if `multilingual`: true, then `description` keys have value object with language tags as keys and description strings as values.

JSON-LD only needs language tags if there are multiple languages. otherwise, the string suffices and you define a "default language" in the context of the JSON-LD to compensate. I'm trying to imitate this so its easier to parse.

I guess I should define this earlier so it's clear that `description` can be either a string or an object depending on `multilingual`. what do you think about keeping the multilingual key in the meta component after all?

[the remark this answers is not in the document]
driusan: I still don't like it. Types shouldn't be context-sensitive.

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+
+```js
+{
+    "data" : {
+        "nameOfField" : `schema.fields.nameOfField.type`
+    }
+}
+```
+
+`description`: Required string. The human readable description of this field, provided in every available language. It
```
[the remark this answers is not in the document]
zaliqarosli: the language code or the description?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+                    "displayIf": string, /* LORIS Logic Parser formula */
+                    "disableIf": string, /* LORIS Logic Parser formula */
+                }
+        ...
+}
+```
+
+The value of these keys are string expressions of logical formulas. These formulas, when calculated, evaluate to true
+or false on the front-end. The formulas are parsed and calculated using the Evaluator function of the LORIS Logic
+Parser, [LParse](https://gitlab.com/zainvirani/LParse). The currently supported logic operations are available in the LParse [README.md](https://gitlab.com/zainvirani/LParse/blob/master/README.md), as well as
```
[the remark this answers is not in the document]
zaliqarosli: should i add it to this same PR?

docs/instruments/ZJSONInstrument_v0.0.1.md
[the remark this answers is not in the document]
driusan: The description encoding of the multilingual values.. I think JSON might be defined as being utf-8, but it doesn't hurt to be explicit.

docs/instruments/ZJSONInstrument_v0.0.1.md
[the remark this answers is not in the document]
zaliqarosli: a required field must always be displayed. if it's not displayed, then it's not required. however a required field can become disabled, and thus, no longer required. does this make sense?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+instructions on how to customize functions. 
+
+`rules.requireIf`: String. True or false that the field requires data input, given the condition in the string expression.
+
+`rules.displayIf`: String. True or false that the field is displayed in the instrument, given the condition in the string
+expression.
+
+`rules.disableIf`: String. True or false that the field is disabled i.e. not editable nor submittable, given the condition in
```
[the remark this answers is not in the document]
zaliqarosli: answered above ^

docs/instruments/ZJSONInstrument_v0.0.1.md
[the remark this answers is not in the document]
zaliqarosli: int, decimal and date, time are grouped further below in the spec into numeric and 'Date and Time'. but they are different data types. if we have a field of type numeric, how else can we validate the data if we don't specify that the data is of type int or decimal? i think it makes sense to group them together, but not coalesce them into the same field type. 

from my research, the datetime input element is about to be unsupported. so if we do want to input a datetime, it would have to be two different fields of date and time. thoughts?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,1028 @@
...
+
+`options.requireResponse`: Boolean. Default is true.
+                           True or false that an input is required. This key is different to
+                           `rules.requireIf` in that it is a true boolean and not dependent on a condition - it is always either true or false.
+                           If true, an implementation should automatically add a 'not_answered' value, e.g. rendered as a checkbox, to allow
+                           explicit non-answering while still requiring some input value.
+
+### 3.2.4: Date and Time
+
+A date field takes a data type of form "YYYY-MM-DD". A time field takes a data type of form "HH:MM". The general format is as follows:
```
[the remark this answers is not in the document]
zaliqarosli: i for some reason thought that datetime was being deprecated but i see that it exists so i will add this implementation in!

docs/instruments/ZJSONInstrument_v0.0.1.md
zaliqarosli: to ensure that the `order` array has the same number of elements as `rowSize` and that none are missing

docs/instruments/ZJSONInstrument_v0.0.1.md
zaliqarosli: this is also for validating the elements of the table

docs/instruments/ZJSONInstrument_v0.0.1.md
zaliqarosli: for validation

docs/instruments/ZJSONInstrument_v0.0.1.md
zaliqarosli: yes, I would like to create a JSON schema of the spec that can be used to validate all Z-JSONs http://json-schema.org/. just as JSON-LD also has its schema and validator.

docs/instruments/ZJSONInstrument_v0.0.1.md
zaliqarosli: if a section's `requireIf` evaluates to true, then all the helper/fields in the section will also be required a.k.a their `requireIf` becomes true

docs/instruments/ZJSONInstrument_v0.0.1.md
zaliqarosli: the `description` is required. But what if an implementation doesn't want to show it on the front-end?

docs/instruments/ZJSONInstrument_v0.0.1.md
zaliqarosli: i see, i guess it doesn't necessarily have to be on the browser

docs/instruments/ZJSONInstrument_v0.0.1.md
zaliqarosli: file will be for second phase implementation. i like implementation a ui image type but will need to think how that helper element would be defined

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -0,0 +1,970 @@
...
+The resource instrument that the Z-JSON Instrument derived from as a result of significant content modification such that the instrument is no longer the same                    
+instrument, and the creation of a new version does not suffice. What is considered significant here is not defined in this specification, but is instead up to the author to decide.
+                    
+This level of modification does not include format transcription i.e. linst to
+JSON, unless the content of the instrument has also been modified. The
+`meta.importedFrom` key should be used in the case of conversions.
+
+`meta.importedFrom`: String.
+
+The resource instrument that this Z-JSON Instrument was converted from, if it was imported from another format and is otherwise unmodified. If any modifications to the content of the instrument are madeafter conversion, this value should be empty and `meta.derivedFrom` used instead.
```
PapillonMcGill:
```suggestion
The resource instrument that this Z-JSON Instrument was converted from, if it was imported from another format and is otherwise unmodified. If any modifications to the content of the instrument are made after conversion, this value should be empty and `meta.derivedFrom` used instead.
```

docs/instruments/ZJSONInstrument_v0.0.1.md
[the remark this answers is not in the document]
driusan: ..but what does interoperable mean in this context? How do they interoperate?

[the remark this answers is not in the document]
zaliqarosli: it means that we built this schema with the intention to easily mappable to both JSON-LD and NDAR via converters that will also be created to be included in the package. how are you looking for me to word this on the spec? @driusan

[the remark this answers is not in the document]
driusan: Hm.. I'm not sure about wording, but "compatible" doesn't really say much in this context.

How about the wording that you just said instead of "compatible"? "...was built with the intention to be easily mappable to both ReproNim's JSON-LD schema and NDAR's schema via converters"?

docs/instruments/ZJSONInstrument_v0.0.1.md
```diff
@@ -551,6 +534,14 @@ represented on the front-end by an input element. Its value is instead calculate
...
 }
 ```
+`options.scoringFormula`: Required string. An expression of a logical formula to be calculated by LORIS LParse. The formula is evaluated to true or false if a boolean score, or a numeric if a number score. The result is dependent and defined by the value of this key.
+
+`options.readonly`: Required boolean.
+                    True or false that the field is read only i.e. not editable on the front-end but
+                    submittable on save. This key is different to `rules.disableIf` in that it is not
+                    dependent on a condition, and a 'disabled' field is not submittable.
+
+`options.hideInSurvey`: Required boolean. True or false that this score field will be hidden and not displayed when the instrument is in survey mode.
```
sruthymathew123: @zaliqarosli I think in some cases, we need to hide other fields also in Survey mode apart from score fields.

docs/instruments/ZJSONInstrument_v0.0.1.md
zaliqarosli: i think we should keep 'select' because there's also multiselect elements which doesn't make sense as 'multidropdown'

docs/instruments/bmi.json
```diff
@@ -0,0 +1,532 @@
...
+        "options": {
+          "minValue": "",
+          "maxValue": "",
+          "requireResponse": true,
+          "readonly": false,
+          "hideInSurvey": false,
+          "showDesc": false
+
+        },
+        "rules": {
```
christinerogers: Rules for these 5 fields  (and also their `_status` fields?) : 
* height_feet
* height_inches
* weight_lbs
* height_cms
* weight_kgs

are found in the *.rules file for this Linst : 
https://github.com/aces/Loris/blob/minor/docs/instruments/bmi.rules

docs/instruments/bmi.json
```diff
@@ -0,0 +1,532 @@
...
+        },
+        "options": {
+          "order": [
+            "height_cms",
+            "height_cms_status"
+          ],
+          "showDesc": true
+        },
+        "rules": {
+          "requireIf": "",
```
christinerogers: How do "requireIf" rules on Groups interact with "requireIf" rules on elements?   Does one override the other?

docs/instruments/bmi.json
```diff
@@ -0,0 +1,532 @@
...
+    "weight_kgs": {
+      "type": "number"
+    },
+    "weight_kgs_status": {
+      "type": "select",
+      "options": {
+        "showEmptyOption" : true
+      }
+    },
+    "bmi": {
```
christinerogers: If this is the Score field, the scoring calculation is found in the *.score file (for a given linst) : 

https://github.com/aces/Loris/blob/minor/docs/instruments/bmi.score

docs/instruments/bmi.json
```diff
@@ -0,0 +1,532 @@
+{
+  "schema": {
+    "meta": {
+      "createdBy": "Zaliqa Rosli",
+      "createdOn": "2019-02-19",
+      "createdWith": "",
+      "defaultLanguage": "en-CA",
+      "derivedFrom": "",
+      "importedFrom": "bmi.linst",
```
christinerogers: ..., bmi.rules, bmi.score

docs/instruments/LORISLogicParser.md
```diff
@@ -0,0 +1,143 @@
...
+- [PHP Use](#php-use)
+- [Development](#development)
+- [Syntax](#syntax)
+
+The parser provides a human readable syntax for front end equation building, real-time calculations, and back end validation. 
+
+The syntax can be used to determine whether instrument fields are shown, required, or hidden in survey mode.
+It can also be used to perform real-time scoring calculations while referencing other fields or subject meta-data.
+
+Fun features: infinite variable nesting, easily managed evaluation, human readable syntax (similar to Excel), AST (abstract syntax tree) evaluation, fraction factorials, variance and standard deviation, date calculations, boolean operations, type checking and much more.
```
driusan:
```
[..] human readable syntax (similar to Excel) [...]
```

which is it?

docs/instruments/LORISLogicParser.md
```diff
@@ -0,0 +1,143 @@
...
+### PHP Grammar Specifics
+Tokens are defined in `lexer.php`. Precedence and assertions are defined by the order of parsing functions in `parser.php`. The grammar is defined in `lexer.php`. Similarly to JS, an array defining `tag`, `op`, and `args` is output to `evaluator.php`.
+
+### Evaluator Changes
+To add new types of operations, add a case to the switch statement in `js/Evaluator.js` and `php/evaluator.php`.
+
+### Unit Testing
+JS tests can be added to `Loris/test/js-tests/Parser.test.js`. Run tests with `npm run tests:unit:js:watch`.
+
+PHP tests can be added to `Loris/test/unittests/ParserTest.php`. Run tests with `Loris/vendor/bin/phpunit --configuration phpunit.xml --testsuite 'PHPParserTest'`.
```
driusan: Most of these sections don't belong here, since this is just here to define the language for the Z-JSON format to reference. The above descriptions of how to change things shouldn't be in the spec documentation.

docs/instruments/LORISLogicParser.md
```diff
@@ -0,0 +1,143 @@
...
+### Numerical Operations
+| Operation 	| Syntax 	| Notes                                                             	|
+|-----------	|--------	|--------------------------------------------------------------------   |
+| add       	| a + b  	|                                                                   	|
+| subtract  	| a - b  	|                                                                    	|
+| negate    	| - a    	|                                                                    	|
+| multiply  	| a * b  	|                                                                     	|
+| divide    	| a / b  	| cannot divide by 0                                                	|
+| exponent  	| a ^ b  	|                                                                     	|
+| percentage  	| a %   	| divides the value of a by 100                                         |
```
driusan: why does a '%' suffix divide a value by 100?

docs/instruments/LORISLogicParser.md
```diff
@@ -0,0 +1,143 @@
...
+|-----------	|--------------	|----------------------------------------------------------------------------------	|
+| if        	| if(cond,x,y) 	| If the condition 'cond' evaluates to true, x is returned. If not, y is returned. 	|
+
+### Functions
+| Operation          	| Syntax               	| Notes                                                                                	|
+|--------------------	|----------------------	|--------------------------------------------------------------------------------------	|
+| not a number       	| isNaN(a)             	| returns true if a is not a number                                                    	|
+| modulo             	| mod(a, b)            	| returns the remainder of a / b                                                       	|
+| round              	| round(a, b)          	| rounds a to b decimal places; note all rounding functions support trailing 0s        	|
+| round up           	| roundup(a, b)        	| rounds a up to b decimal places                                                      	|
```
driusan: "ceil" would be more consistent with most other languages (and "floor" instead of "rounddown")

docs/instruments/LORISLogicParser.md
```diff
@@ -0,0 +1,143 @@
...
+| median             	| median(a,b,c,d...)   	| sorts its arguments in ascending order and returns the median                        	|
+| sum                	| sum(a,b,c,d...)      	| returns the sum of its arguments                                                     	|
+| product            	| product(a,b,c,d...)  	| returns the product of its arguments                                                 	|
+| variance           	| variance(a,b,c,d...) 	| returns the population variance of its arguments                                     	|
+| standard deviation 	| stdev(a,b,c,d...)    	| returns the population standard deviation (square root of variance) of its arguments 	|
+
+### Date Operations
+| Operation       	| Syntax                                        	|
+|-----------------	|-----------------------------------------------	|
+| date difference 	| datediff(date1, date2, units, signed)         	|
```
driusan: what type is the return value of a date difference?

docs/instruments/bmi.json
zaliqarosli: thanks, this implementation is still not complete so that's why there are things missing

docs/instruments/ZJSONInstrument_template.json
```diff
@@ -0,0 +1,82 @@
+{
+  "schema": {
+    "meta": {
+      "createdBy": "",
+      "createdOn": "",
```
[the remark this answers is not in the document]
zaliqarosli: @christinerogers the names of these fields are based on the PAV ontology (https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4177195/) to be used in the Repronim JSON-LD schema (https://github.com/ReproNim/schema-standardization/blob/master/contexts/generic.jsonld) . I can prepend "date" onto the LORIS version no problemo

docs/instruments/bmi.json
zaliqarosli: I don't think adding these are necessary. the instrument itself is the linst file. the use of *.rules is not currently reliable and the *.score contains calculations that accompany the instrument. if "importedFrom" points to a .linst instrument, it's up to the source instrument to show whether or not it's accompanied by a *.score or *.rules file i.e. they should be in the same directory as noted on the linst documentation (https://github.com/aces/Loris/wiki/Behavioural-Database)

docs/instruments/LORISLogicParser.md
zaliqarosli: are you suggesting that Excel is not human readable syntax? lol

zaliqarosli:
Let's hold off on reviewing until this discussion get's moved to a google doc. I will keep this PR open for now to address existing reviews

docs/instruments/LORISLogicParser.md
zaliqarosli: @driusan im not too sure. this might be an implementation from redcap. @christinerogers do you know anything from CAP?

zaliqarosli: hold on, this makes sense . if you wanted to e.g. evaluate 5% of 20, you would write (5%*20) which would calculate to (5/100*20). whats the issue?

docs/instruments/LORISLogicParser.md
zaliqarosli: in JS, a number. in PHP, a float. essentially both a decimal

docs/instruments/bmi.json
zaliqarosli: a field's rule would be checked first. if the rule is violated, the group's rule would be checked if it exists

docs/instruments/bmi.json
zaliqarosli: @christinerogers please review the scores and rules added in the last commit

docs/instruments/LORISLogicParser.md
christinerogers: Hey @zaliqarosli this operator isn't from Redcap, I'm 99% sure Zain added it as part of a greater set of functions/operators he thought we might someday use in LORIS instruments.

docs/instruments/bmi.json
```diff
@@ -0,0 +1,554 @@
...
+          "disableIf": "" 
+        }
+      },
+      "bmi" : {
+        "type": "score",
+        "description": {
+            "en-CA": "Your BMI"
+        },
+        "options": {
+          "scoringFormula": "if([unit_classification]='metric', round([weight_kgs]/([height_cms]/100)^2, 2), round(([weight_lbs]*703)/([height_feet]*12+[height_inches])^2, 2))",
```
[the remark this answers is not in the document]
zaliqarosli: should follow BODMAS (BEDMAS is canada?) where exponents are evaluated before divisions but I'm only assuming this standard is implemented

docs/instruments/bmi.json
```diff
@@ -303,7 +303,7 @@
             "en-CA": "Your BMI Classification"
         },
         "options": {
-          "scoringFormula": "if([bmi] < 18.5, 'Underweight', if([bmi]>=18.5 and [bmi]<=22.9, 'Normal Range, if([bmi]>=23.0 and [bmi]<=24.9, 'Overweight - At Risk', if([bmi]>=25.0 and [bmi]<=29.9, 'Overweight - Moderately Obese', 'Overweight - Severely Obese')))))",
+          "scoringFormula": "if([bmi] < 18.5, 'Underweight', if([bmi]>=18.5 and [bmi]<=22.9, 'Normal Rangei', if([bmi]>=23.0 and [bmi]<=24.9, 'Overweight - At Risk', if([bmi]>=25.0 and [bmi]<=29.9, 'Overweight - Moderately Obese', 'Overweight - Severely Obese'))))",
```
christinerogers: typo @zaliqarosli   "Rangei" -> "Range"

christinerogers:
Thanks for the changes @zaliqarosli  
I found one new typo and didn't re-review the Rules exhaustively across every field, but I know we'll come back to those and this is a great step forward with the example BMI instrument already.

zaliqarosli:
Google Doc: https://drive.google.com/open?id=1IqHCBM4yW5fpPqZeKvjhtyS59H6ZIll8_op-DTWwwUo

docs/instruments/ZJSONInstrument_v0.0.1.md
zaliqarosli: hey @driusan i am moving towards consolidating the fields and helpers into one "elements" object. do you still stand by your points above before I move forward? thanks

driusan:
PR was replaced by Google Doc (see updated description for link.)

PR #4325, opened 2019-02-12
[cleanup] package.json, compile errors and [update] webpack.config.js & .babelrc

maltheism:
# [cleanup] package.json, compile errors and [update] webpack.config.js & .babelrc

### Brief summary of changes

The following files are notable: 
package.json,
webpack.config.js, 
babelrc, 
js files (breaking webpack compile because of format rules now being used)

I modified the mininum of what I believe is necessary for this PR. This PR should be discussed because in my opinion it (slightly) improves with modernization of how JS & CSS files are served to the “user” browser client.

#### Notable changes found in webpack.config.js. 

Modern web apps will bundle the JS & CSS files into one single file for the CSS and JS. Benefit can be decreasing server requests with caching and nullifying reason to have separate JS files. The bundle strategy arguably should make the backend simpler when the client just needs the same (CSS & JS) bundled file(s).

Since Loris has many individual modules (serving the JS files separately), the process of making all the JS or CS files be one bundle is cumbersome. So I specifically split the JS files, I believe to be utilities into a bundle and from the other JS files are still being served individually to the client browser. Some CSS files are not included in the packaged bundle because it needs to be discussed as well.

### This resolves issue...

- [ ] Redmine? #

- [ ] Github? #

### To test this change...

- [x] npm install, npm run compile


maltheism:
I added @HenriRabalais and @zaliqarosli as reviewers to just let them know because of their expertise in frontend.

HenriRabalais:
I think there may be some issues with this PR, it seems that it's adding over 100k lines of code.

johnsaigle:
@maltheism Can you provide more detail as to why these changes are beneficial, how they impact LORIS, etc. ? As @HenriRabalais a PR of this size is extremely difficult to review. I don't know much about JS but it seems to include everything from small formatting changes to large architectural cahnges about how we're structuring code.

Also for the modules I think we're explicitly not tracking files in `modules/*/js/*` as there's no need to track the compiled JS when it can be built on the server using `make`. If I'm mistaken here and the files you added are not compiled JS but something else, please let me know.

In general it's difficult to evaluate this approach without more info about pros and cons.

maltheism:
Hi @HenriRabalais and @johnsaigle the files changed are actually needed after updating the package.json to the latest modules. The rules found from the .eslintrc.json became necessary for all the files to be changed for the command "nom run compile" to execute without errors.
I tried to make it clear in the initial description a lot should be ignored. The main scope of this PR is to discuss if people think we should go the bundle.js or bundle.css approach that is gained from the now modified version of webpack.config.js. Basically the webpack.config.js is what should be examined and discussed. Everything else is just for people to execute npm run compile on their own VMs if they pull the PR.

zaliqarosli:
would be nice to maybe bring this up in the loris meeting. not sure if I have much to say right now in this discussion

johnsaigle:
@maltheism Can you add this as a discussion point in next week's LORIS meeting?

maltheism:
@johnsaigle Okay.

maltheism:
PR was discussed today at LORIS Meeting. Closing for now will reopen or make new PR with less changes.

PR #4329, opened 2019-02-12
[Phan] Add PhanTypeMismatchDimAssignment to static analysis rules

johnsaigle:
### Brief summary of changes

Add another phan rule to our checks. This rule makes sure that when array elements are accessed that they have the right type (e.g. numeric arrays should have numeric keys).

### DIFF
--- a/modules/imaging_browser/php/mrinavigation.class.inc
+++ b/modules/imaging_browser/php/mrinavigation.class.inc
@@ -27,6 +27,9 @@ namespace LORIS\imaging_browser;
 
 class MRINavigation
 {
+
+    var $urlParams = array();
+
     /**
     *  Gets the session
     *
--- a/modules/imaging_browser/php/mrinavigation.class.inc
+++ b/modules/imaging_browser/php/mrinavigation.class.inc
@@ -28,7 +28,7 @@ namespace LORIS\imaging_browser;
 class MRINavigation
 {
 
-    var $urlParams = array();
+    public $urlParams = array();
 
     /**
     *  Gets the session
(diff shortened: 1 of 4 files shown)


modules/imaging_browser/php/mrinavigation.class.inc
```diff
@@ -27,6 +27,9 @@ namespace LORIS\imaging_browser;
 
 class MRINavigation
 {
+
+    var $urlParams = array();
```
PapillonMcGill: This is not the only class variable. Should we define all class variable here?

johnsaigle: Maybe, I was just adding the minimum changes needed here to include the Phan rule and fix the errors.

driusan:
```suggestion
    public $urlParams = array();
```

Since this is new and we don't need to support ancient versions of PHP, we should be using visibility modifiers instead of "var" (I didn't look into if this should be more strict than public, but the suggested change suggests public since that's equivalent to var..)

modules/issue_tracker/php/issue_tracker_controlpanel.class.inc
```diff
@@ -115,12 +115,15 @@ class Issue_Tracker_ControlPanel
 
 class IssueNavigation
 {
+
+    var $urlParams = array();
```
driusan:
```suggestion
    public $urlParams = array();
```

same as above

johnsaigle:
@driusan Thanks for the review. I didn't know `var` was considered bad practice.

PR #4330, opened 2019-02-12
[Genomics] Quick cleanup

johnsaigle:
### Brief summary of changes

Some brief TLC to this code. There were some PHP Notices generated and I noticed certain issues pop up when trying to implement phan rules so I figured I'd tie up some loose threads here.

### DIFF
--- a/modules/genomic_browser/php/genomic_file.class.inc
+++ b/modules/genomic_browser/php/genomic_file.class.inc
@@ -2,7 +2,7 @@
 /**
 * Contains the Genomic_File class used for LORIS Genomic data
 *
-* PHP Version &
+* PHP Version 7
 *
 * @category LORIS_Module
 * @package  Genomics
@@ -30,8 +30,10 @@ class Genomic_File
     * Construct a GenomicFile
     *
     * @param integer $fileID The FileID to be loaded
+    *
+    * @return void
     */
-    function __construct($fileID)
+    function __construct(int $fileID)
     {
         $db     =& \Database::singleton();
         $params = array('FID' => $fileID);
@@ -59,13 +61,14 @@ class Genomic_File
     }
 
     /**
-    * Gets a parameter for this file from the appropriate place
+    * Gets a parameter for this file from the appropriate place.
     *
-    * @param string $parameterName The parameter to be retrieved
+    * @param string $parameterName The parameter to be retrieved.
     *
-    * @return mixed The value of the parameter
+    * @return ?string The value of the requested parameter. All values will be
+    * represented as strings even if they are e.g. ints or dates.
     */
-    function getParameter($parameterName)
+    public function getParameter(string $parameterName): ?string
     {
         if (isset($this->QCData[$parameterName])) {
             return $this->QCData[$parameterName];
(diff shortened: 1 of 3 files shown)


modules/genomic_browser/php/view_genomic_file.class.inc
```diff
@@ -57,10 +57,8 @@ class View_Genomic_File extends \NDB_Form
     {
         parent::setup();
 
-        $this->DB = \Database::singleton();
-
-        $this->genomic_file_id = $_REQUEST['GenomicFileID'];
-        $this->candID          = $_REQUEST['candID'];
+        $this->genomic_file_id = $_REQUEST['GenomicFileID'] ?? null;
+        $this->candID          = $_REQUEST['candID'] ?? null;
```
johnsaigle: This variable was causing undefined notices.

modules/genomic_browser/php/view_genomic_file.class.inc
```diff
@@ -103,7 +101,7 @@ class View_Genomic_File extends \NDB_Form
             // load $file array
             $file = array(
                      'FileID'           => $fileRow['GenomicFileID'],
-                     'Filename'         => $paramFileName,
+                     'Filename'         => $paramFilename,
```
johnsaigle: This is the correct name of the variable (see line 95)

modules/genomic_browser/php/genomic_file.class.inc
```diff
@@ -78,18 +81,4 @@ class Genomic_File
             return null;
         }
     }
-
-    /**
-    * Gets the AnalysisProtocol for this file
-    *
-    * @return String from the file properties
-    */
-    function getAnalysisProtocol()
```
johnsaigle: This variable can be retrieved using `getParameter` like all the others. No need for a special function.

modules/genomic_browser/php/view_genomic_file.class.inc
```diff
@@ -120,16 +118,7 @@ class View_Genomic_File extends \NDB_Form
                      'AnalysisModality' => $FileObj->getParameter(
                          'AnalysisModality'
                      ),
-                     'AnalysisDate'     => $this->_getDate(
-                         $FileObj,
-                         'Date_taken',
-                         $analysisDate
```
johnsaigle: $analysisDate and $sampleDate don't exist.

modules/genomic_browser/php/view_genomic_file.class.inc
```diff
@@ -120,16 +118,7 @@ class View_Genomic_File extends \NDB_Form
...
-                     ),
-                     'SampleDate'       => $this->_getDate(
-                         $FileObj,
-                         'Date_sample',
```
johnsaigle: This column does not exist in the DB.

modules/genomic_browser/php/view_genomic_file.class.inc
```diff
@@ -139,25 +128,5 @@ class View_Genomic_File extends \NDB_Form
...
-    * @param array        $array array containing the date
-    *
-    * @return int|string Date (Unix int timestamp) if exists, or an empty string
-    */
-    function _getDate($file, $type, $array)
-    {
-        $pattern  = "/(\d{4})-?(\d{2})-?(\d{2})/";
-        $fileType = $file->getParameter($type);
-        if (preg_match($pattern, $fileType, $array)) {
-            return (mktime(12, 0, 0, $array[2], $array[3], $array[1]));
```
johnsaigle: Since `$array` depended on the two variables above that do not exist, `mktime()` is effectively being called with (12,0,0,0,0,0) as parameters which generates the output `943981200` (Nov. 30 1999).

Instead, the `AnalysisDate` variable above is retrieved using the getParameter function above like the other variables do.

PR #4331, opened 2019-02-12
 [Phan] Add rule PhanTypeMismatchArgumentInternal 

johnsaigle:
### Brief summary of changes

This rule makes sure that when we use PHP built-in functions that the arguments we pass to them are of the correct type.

~~Blocked by https://github.com/aces/Loris/pull/4330 and https://github.com/aces/Loris/pull/4320.~~

### DIFF
--- a/modules/dataquery/ajax/retrieveCategoryDocs.php
+++ b/modules/dataquery/ajax/retrieveCategoryDocs.php
@@ -10,7 +10,9 @@
  * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
  * @link     https://www.github.com/aces/Loris/
  */
-ini_set("max_input_vars", 10000);
+
+
+ini_set("max_input_vars", '10000');
 $user =& User::singleton();
 if (!$user->hasPermission('dataquery_view')) {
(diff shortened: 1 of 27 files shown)


modules/dataquery/ajax/retrieveCategoryDocs.php
```diff
@@ -10,7 +10,9 @@
  * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
  * @link     https://www.github.com/aces/Loris/
  */
-ini_set("max_input_vars", 10000);
+
+
```
PapillonMcGill: @johnsaigle If you leave 2 empty line here, you should do the same thing in modules/dataquery/ajax/saveQuery.php

johnsaigle: Okay I changed the other file to have new lines but I really don't think something like this is worth blocking the PR for.

driusan: I don't know of any coding standard that suggests two blank lines, this'll just make it require another PR later to fix this. There's no reason to not fix it now.

Issue #4210, opened 2018-12-19
Add stronger typing 


driusan:
@johnsaigle has been adding PHP type hints to all of our classes based on the doc block (Status: https://github.com/aces/Loris/projects/11), but how well the compiler can find errors is dependent on how granular our type system is, and right now the vast majority of places are using "string", "array", or (sometimes) "int".

We should implement classes/interfaces for some commonly used types so that PHP/phan can better determine when the wrong type of thing is accidentally passed and yell at us.

#4055 adds a class for CandIDs, and my gut feeling tells me SessionIDs and CommentIDs are probably low hanging fruit. (In most cases what the code that's using them probably really needs is a Candidate or Session object, not an ID, but it would be easier to start by typing the IDs than it would be to completely rewrite the code that's using them to use a data model.)

johnsaigle:
Create [a project](https://github.com/aces/Loris/projects/15) to track the steps involved in this issue since it'll need to be broken into many separate tasks.

driusan:
This is probably too vague to be an issue, and it should just be something we look out for in code review

PR #4333, opened 2019-02-13
[Core] Add SessionID value class

johnsaigle:
### Brief summary of changes

Creates a SessionID class for use in validating types in LORIS.

### Related

* #4210 
* #4055

### DIFF
--- a/php/libraries/Identifier.php
+++ b/php/libraries/Identifier.php
@@ -1,0 +1,42 @@
+<?php declare(strict_types=1);
+/**
+ * PHP Version 7
+ *
+ * @category Model
+ * @package  Behavioural
+ * @author   Xavier Lecours Boucher <xavier.lecoursboucher@mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+
+/**
+ * An Identifier is something that it used to identify an element of a study,
+ * such as a Candidate, Session, etc.
+ *
+ * Instances of Identifier must have a type to distinguish what kind of identifier
+ * it is (e.g. CandID, SessionID) and must be serializable to a string.
+ *
+ * @category Model
+ * @package  Behavioural
+ * @author   Xavier Lecours Boucher <xavier.lecoursboucher@mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+interface Identifier
+{
+    /**
+     * Return the name of the identifier type for this interface
+     * implementation.
+     *
+     * @return string The identifier type
+     */
+    public function getType(): string;
+
+    /**
+     * Serializes the identifier's value to a string
+     *
+     * @return string The identifier type
+     */
+    public function __toString(): string;
+}
+
--- a/php/libraries/SessionID.php
+++ b/php/libraries/SessionID.php
@@ -5,20 +5,19 @@
  *
  * PHP Version 7
  *
- * @category StudyEntities
+ * @category Main
  * @package  LORIS
  * @author   Xavier Lecours Boucher <xavier.lecoursboucher@mcgill.ca>
  * @author   John Saigle <john.saigle@mcin.ca>
  * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
  * @link     https://www.github.com/aces/Loris/
  */
-namespace LORIS\StudyEntities\Candidate;
 
 /**
  * A representation of a SessionID object. A SessionID is always an integer with a
  * maximum of 10 digits (as definied in the database `session` table.
  *
- * @category StudyEntities
+ * @category Main
  * @package  LORIS
  * @author   Xavier Lecours Boucher <xavier.lecoursboucher@mcgill.ca>
  * @author   John Saigle <john.saigle@mcin.ca>
@@ -40,7 +39,7 @@ class SessionID extends ValidatableIdentifier
     /**
      * Validate that the value of the SessionID is a positive integer that does
      * not exceed 10 digits in length.
-     * This does not check for uniqueness in the database or any other 
+     * This does not check for uniqueness in the database or any other
      * state-related facts.
      *
      * This function is called by the contructor of ValidatableIdentifier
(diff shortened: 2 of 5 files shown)


src/StudyEntities/Candidate/SessionID.php
```diff
@@ -0,0 +1,68 @@
...
+ * maximum of 10 digits (as definied in the database `session` table.
+ *
+ * @category StudyEntities
+ * @package  LORIS
+ * @author   Xavier Lecours Boucher <xavier.lecoursboucher@mcgill.ca>
+ * @author   John Saigle <john.saigle@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+class SessionID extends ValidatableIdentifier
```
driusan: This is in  `namespace LORIS\StudyEntities\Candidate;` A SessionID is not a validatable candidate identifier.

src/StudyEntities/Candidate/SessionID.php
```diff
@@ -0,0 +1,68 @@
...
+     */
+    protected function validate(int $value): bool
+    {
+        return ($value > 0) && (strlen((string) $value <= 10));
+    }
+
+    /**
+     * Generates a string representation of this SessionID.
+     *
+     * @return string The strin representation of the SessionID int value.
```
driusan:
```suggestion
     * @return string The string representation of the SessionID int value.
```

src/StudyEntities/Candidate/SessionID.php
johnsaigle: @driusan I have no preference about where the files are stored but I would like to use the ValidatableIdentifier class. It seems generic enough to be used for classes beyond CandID|PSCID|ExternalID.

How do you feel about me moving Identifier and ValidatableIdentifier to php/libraries and updating the `StudyEntities\Candidate` classes accordingly? SessionID can also be moved to libraries as you're right it doesn't belong in the Candidate folder.

driusan: I think it would be a good idea to keep the StudyEntities namespace self-contained enough that it has no coupling to LORIS and could eventually be moved into a different repo and just composer required by LORIS (whether or not that ever happens, it should be possible to do it since a bunch of models for scientific study purposes is useful outside of LORIS) and I think it's a good idea to cleanly separate whether an identifier is a Candidate identifier, or an identifier for something else in the class hierarchy.

How about just making a copy of ValidatableIdentifier in php/libraries?

driusan: @xlecours @ridz1208

johnsaigle: Sure I can make a copy. Happy to go with the flow~

johnsaigle: The files are now in the libraries/ folder

php/libraries/SessionID.php
```diff
@@ -0,0 +1,69 @@
...
+ * @link     https://www.github.com/aces/Loris/
+ */
+
+/**
+ * A representation of a SessionID object. A SessionID is always an integer with a
+ * maximum of 10 digits (as definied in the database `session` table.
+ *
+ * @category Main
+ * @package  LORIS
+ * @author   Xavier Lecours Boucher <xavier.lecoursboucher@mcgill.ca>
```
xlecours: My name can be removed :)

xlecours:
I am lukewarm about this proposal. While I am strongly for specific typing in function parameters (and I think this is the purpose of this PR), I think this deviate from its goal in the sens that instead of passing a Timepoint object, one could just pass a SessionID casted int or string. 
Most function that receive a sessionid should receive a Timepoint and call `$timepoint->getSessionID()` if they need its id. Otherwise, there will be a static call for Timepoint::singelton($session_id) in the called function (And that is... beyond bad).

johnsaigle:
@xlecours I see your point. I was mostly going off of the "low-hanging fruit" in https://github.com/aces/Loris/issues/4210.

At the same time this validation might be nice. For example if we make a CommentID class it will be useful to have a SessionID class because a CommentID is composed of `$candID . $PSCID . $sessionID. ...` and we can leverage each class to do its own validation when we validate the CommentID. There may be other useful examples also.

@driusan What do you think?

driusan:
I think I agree with @xlecours in terms of how it should things should be modeled in the StudyEntities namespace, but also that strong, specific types are valuable as function parameters/return arguments.. that's why I suggest just adding them into php/libraries until proper models for everything are holistically designed to go into StudyEntities.

xlecours:
@driusan @johnsaigle Sold. This is valuable as an intermediate state and allow for value-object validation.

johnsaigle:
@driusan Can you update your review?

php/libraries/SessionID.php
```diff
@@ -0,0 +1,67 @@
...
+ * @category Main
+ * @package  LORIS
+ * @author   John Saigle <john.saigle@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+
+/**
+ * A representation of a SessionID object. A SessionID is always an integer with a
+ * maximum of 10 digits (as definied in the database `session` table.
```
driusan: Missing the closing bracket

php/libraries/ValidatableIdentifier.php
```diff
@@ -0,0 +1,65 @@
...
+ * @author   Xavier Lecours Boucher <xavier.lecours@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+abstract class ValidatableIdentifier implements Identifier
+{
+    /**
+     * The Identifier's value
+     *
+     * @var int|string The identifier value to be validated
```
driusan: It looks like this can only be a string since it's set by the constructor, which takes a string (which means that __toString can be moved up a level from SessionID to ValidatableIdentifier)

johnsaigle: You're right, I left this in by mistake from an earlier idea.

php/libraries/Identifier.php
```diff
@@ -0,0 +1,42 @@
...
+     * Return the name of the identifier type for this interface
+     * implementation.
+     *
+     * @return string The identifier type
+     */
+    public function getType(): string;
+
+    /**
+     * Serializes the identifier's value to a string
+     *
```
PapillonMcGill:
```suggestion
     *ggg
```

PR #4334, opened 2019-02-13
Update Phan to newest version

johnsaigle:
### Brief summary of changes

Our previous version of Phan was 1.0.1 and the current is 1.2.3 so I figured it was time to update. Since this is a dev-only tool updating shouldn't have any adverse effects on projects. 

Since new features have been introduced I have added the new rules to our list of ignored rules. They can be tracked and amended using https://github.com/aces/Loris/projects/14.

The PR also deletes a function in FeedbackMRI.class.inc that was [previously removed](https://github.com/aces/Loris/pull/4264/files#r248417342) and mistakenly re-added to the codebase.

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -31,6 +31,10 @@ return [
 		"PhanTypeMismatchProperty",
 		"PhanNonClassMethodCall",
 		"PhanTypeArraySuspicious",
+        "PhanTypeExpectedObjectPropAccess",
+        "PhanTypeSuspiciousStringExpression",
+        "PhanContinueTargetingSwitch",
+        "PhanTypeVoidAssignment"
 	],
 	"analyzed_file_extensions" => ["php", "inc"],
 	"directory_list" => [
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -31,10 +31,8 @@ return [
 		"PhanTypeMismatchProperty",
 		"PhanNonClassMethodCall",
 		"PhanTypeArraySuspicious",
-        "PhanTypeExpectedObjectPropAccess",
         "PhanTypeSuspiciousStringExpression",
         "PhanContinueTargetingSwitch",
-        "PhanTypeVoidAssignment"
 	],
 	"analyzed_file_extensions" => ["php", "inc"],
 	"directory_list" => [
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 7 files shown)


composer.lock
```diff
@@ -862,16 +862,16 @@
         },
         {
             "name": "composer/xdebug-handler",
-            "version": "1.1.0",
+            "version": "1.3.2",
```
johnsaigle: This and the other libraries updated below are required by `phan`.

.phan/config.php
```diff
@@ -31,6 +31,10 @@
 		"PhanTypeMismatchProperty",
 		"PhanNonClassMethodCall",
 		"PhanTypeArraySuspicious",
+        "PhanTypeExpectedObjectPropAccess",
+        "PhanTypeSuspiciousStringExpression",
+        "PhanContinueTargetingSwitch",
+        "PhanTypeVoidAssignment"
```
driusan: Weren't PhanTypeExpectedObjectPropAccess and PhanTypeVoidAssignment already fixed?

johnsaigle: Oops yes, but I saw new errors being generated. Not sure if that's due to a bad merge into major or because Phan got better at detecting errors. I'll remove these rules again and also issue fixes for the errors flagged.

driusan: Could also be due to a bug fix in phan where it's catching things that it didn't previously

driusan: If it's catching new things we can still upgrade it and then fix them in separate PRs

johnsaigle: Nah they're two quick one-liners. May as well do it here.

modules/login/php/passwordreset.class.inc
```diff
@@ -71,7 +71,7 @@ class PasswordReset extends \NDB_Form
                   $password = \User::newPassword();
                   // reset the password in the database
                   // expire password so user must change it upon login
-                  $success = $user->updatePassword($password, '1999-01-01');
+                  $user->updatePassword($password, '1999-01-01');
```
johnsaigle: Fixes a PhanTypeVoidAssignment error. updatePassword() returns void.

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -228,7 +228,7 @@ class Instrument_Manager extends \NDB_Menu_Filter
         // An exit code of 0 is a success and 1 means failure
         if ($status) {
             error_log(
-                "The installation of $instrument->table.sql failed. "
+                "The installation of $instrument.sql failed. "
```
johnsaigle: $instrument is a string, not an object.

PR #4335, opened 2019-02-13
[Coding Standards] Enforce PSR2 standards on files in src/

johnsaigle:
### Brief summary of changes

The `src/` directory and its children are intended to meet the PSR2 coding standards. This PR lays the foundation for achieving this by:
* Adding a new coding standards XML file implementing PSR2 checks on only the src/ directory
* Adding a line in the test script to be added to the analysis process
* Fixing all existing problems

EDIT: This should now fix all the problems and can be incorporated into testing.


docs/SrcCS.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+<ruleset name="Loris Standard">
+    <description>The coding standard for the external src files</description>
+
+    <!-- Include the PEAR standard -->
+    <rule ref="PSR2"/>
```
driusan: I think we still want all the other things that the LorisCS.xml adds on top of PEAR to be enforced for the "src" directory.. it would be better to just copy the file and change the PEAR ref to "PSR2"

johnsaigle: I added them below

src/Api/Endpoints/Login.php
```diff
@@ -34,7 +34,7 @@ class Login extends Endpoint
      *
      * @return boolean true if access is permitted
      */
-    function _hasAccess(\User $user) : bool
+    private function hasAccess(\User $user) : bool
```
driusan: shouldn't this be public?

johnsaigle: Probably? I was just going based off of the underscore.

driusan: PEAR standards used `_` for both protected and private.. based on the underscore, it should be protected (if that still works.. but I don't see why you'd want to prevent another class from checking user access)

johnsaigle: Alright I'll change it to protected

src/Data/Filters/UserSiteMatch.php
```diff
@@ -51,7 +51,7 @@ public function filter(\User $user, \Loris\Data\DataInstance $resource) : bool
                 }
             }
             return false;
-        } else if (method_exists($resource, 'getCenterID')) {
+        } elseif (method_exists($resource, 'getCenterID')) {
```
driusan: I didn't realize using the stupid "elseif" keyword was part of PSR2, but okay :(

src/Http/Response.php
```diff
@@ -22,8 +22,6 @@
  *
  * It is intended to reduce our coupling to any particular PSR15 implementation.
  */
-class Response 
-    extends \Zend\Diactoros\Response 
-    implements \Psr\Http\Message\ResponseInterface
+class Response extends \Zend\Diactoros\Response implements \Psr\Http\Message\ResponseInterface
```
driusan:
```suggestion
class Response extends \Zend\Diactoros\Response implements
    \Psr\Http\Message\ResponseInterface
```

This line is too long and hurts my eyes, which is easy to fix implements may be split across multiple lines

src/Data/Filters/UserSiteMatch.php
johnsaigle: Yeah :(

src/Api/Endpoints/Project/Instrument.php
```diff
@@ -113,7 +113,7 @@ public function handle(ServerRequestInterface $request) : ResponseInterface
      *
      * @return array
      */
-    private function _toArray(ServerRequestInterface $request)
+    private function toArray(ServerRequestInterface $request)
```
driusan: @xlecours can you verify that this won't break anything? I suspect that maybe this one should be protected too, but it's your code

src/Api/Endpoints/Projects.php
```diff
@@ -38,7 +39,7 @@ class Projects extends Endpoint implements \LORIS\Middleware\ETagCalculator
      *
      * @return boolean true if access is permitted
      */
-    function _hasAccess(\User $user) : bool
+    private function hasAccess(\User $user) : bool
```
driusan: @xlecours will this cause problems?

src/StudyEntities/Candidate/ValidatableIdentifier.php
```diff
@@ -55,12 +55,11 @@ abstract class ValidatableIdentifier implements Identifier
      *
      * @throws \DomainException When the value is not valid
      */
-    public final function __construct(string $value)
+    final public function __construct(string $value)
```
driusan: Why is this declared final? What if a subclass needs to change it?

test/run-php-linter.sh
```diff
@@ -18,4 +18,11 @@ vendor/bin/phpcs --standard=docs/LorisCS.xml --extensions=php/php htdocs || exit
 # Run PHPCS on all modules
 vendor/bin/phpcs --standard=docs/LorisCS.xml --extensions=php/php,inc/php modules/ || exit $?;
 
+# Run PHPCS on src/ directory using a different ruleset
+# FIXME Currently the files in the src/ directory do not meet the standards of
+# PSR2. Until they are cleaned up we cannot include this test in our static
+# analysis without breaking Travis. This is not a priority issue but at some
+# point the issues should be remedied and the following line uncommented.
+vendor/bin/phpcs --standard=docs/SrcCS.xml --extensions=php/php,inc/php src/ || exit $?;
```
driusan: the inc/php is only required for the php/libraries directory, in src they should all have the php extension.

src/StudyEntities/Candidate/ValidatableIdentifier.php
johnsaigle: IIRC it's so that every other identifier is forced to call its validate function.

driusan: Whether it's a good idea or not, I don't see why enforcing PSR2 would mandate that, though.. it seems like more than a stylistic change.

johnsaigle: It was final before -- this just changes the order of the keywords.

driusan: Oh, nevermind then. GitHub is only highlighting one of the words, so I didn't catch that.

src/Api/Endpoints/Project/Instrument.php
xlecours: @driusan  `private` is correct

src/Api/Endpoints/Projects.php
xlecours: Nop, see https://github.com/aces/Loris/pull/4971

src/Router/BaseRouter.php
```diff
@@ -106,37 +107,38 @@ public function handle(ServerRequestInterface $request) : ResponseInterface
             $path    = $uri->getPath();
             $baseurl = $uri->withPath("/")->withQuery("");
             switch (count($components)) {
-            case 1:
-                $request = $request
-                    ->withAttribute("baseurl", rtrim($baseurl->__toString(), '/'))
+                case 1:
+                    $request = $request
+                    ->withAttribute("baseurl", $baseurl->__toString())
```
xlecours: @johnsaigle Did you test before removing `rtrim` ?

johnsaigle: I don't remember since I made this PR 6 months ago 🤷‍♂ 

I'll add it back.

src/Router/BaseRouter.php
```diff
@@ -106,37 +107,38 @@ public function handle(ServerRequestInterface $request) : ResponseInterface
             $path    = $uri->getPath();
             $baseurl = $uri->withPath("/")->withQuery("");
             switch (count($components)) {
-            case 1:
-                $request = $request
-                    ->withAttribute("baseurl", rtrim($baseurl->__toString(), '/'))
+                case 1:
+                    $request = $request
+                    ->withAttribute("baseurl", rtrim($baseurl->__toString()))
```
xlecours: it should be `rtrim($baseurl->__toString(), '/')`

PR #4337, opened 2019-02-14
[phan] Fix error about continue targetting switch

driusan:
Fix errors in instrument manager about a continue statement inside
of a switch statement. These are changed to a break, since that's
how the continue behaves in that context anyways. "continue 2" or
"break" are equivalent in this context, since the switch in question
is at the end of a loop, so breaking out of the switch will already
continue to the next iteration.

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -32,7 +32,6 @@ return [
 		"PhanNonClassMethodCall",
 		"PhanTypeArraySuspicious",
         "PhanTypeSuspiciousStringExpression",
-        "PhanContinueTargetingSwitch",
 	],
 	"analyzed_file_extensions" => ["php", "inc"],
 	"directory_list" => [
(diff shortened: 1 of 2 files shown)


driusan:
(this also fixes some other PHP warnings from the linting step, where it was detecting the same error.)

.phan/config.php
```diff
@@ -32,7 +32,6 @@
 		"PhanNonClassMethodCall",
 		"PhanTypeArraySuspicious",
         "PhanTypeSuspiciousStringExpression",
-        "PhanContinueTargetingSwitch",
 	],
```
maltheism: Everything looks correct from what I gather. 
My only question is if "PhanContinueTargetingSwitch" should be removed. It seems like a good way to warn future mistakes of using continue instead of break.

driusan: It's hard to tell because GitHub doesn't show the full context, but this is a list of ignored types that it's being removed from (ie. to make it warn of mistakes in the future), not a list of enforced types.

maltheism: @driusan Okay then everything looks good to me.

PR #4339, opened 2019-02-18
[composer] ext-json is missing in composer.json

maltheism:
### Brief summary of changes

I added the ext-json to composer.json and so the "ext-json is missing in composer.json" warning is not shown. The extension is used with JavaScript Object Notation and specifically when we use json_encode in Loris. I'm not sure if it's necessary because the method does work already without needing to require the extension in composer but it does get rid of the warning in my IDE that is setup to watch composer.json.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,8 @@
         "psr/http-message": "~1.0",
         "psr/http-server-handler" : "*",
         "psr/http-server-middleware" : "*",
-        "zendframework/zend-diactoros": "^1.6"
+        "zendframework/zend-diactoros": "^1.6",
+        "ext-json": "*"
     },
     "require-dev" : {
         "squizlabs/php_codesniffer" : "2.5.*",


composer.json
```diff
@@ -9,7 +9,8 @@
         "psr/http-message": "~1.0",
         "psr/http-server-handler" : "*",
         "psr/http-server-middleware" : "*",
-        "zendframework/zend-diactoros": "^1.6"
+        "zendframework/zend-diactoros": "^1.6",
+        "ext-json": "*"
```
driusan: Does this break anything on non-Debian based systems?

(Iast I checked the JSON functions are a part of the PHP standard library, but removed from Debian and put into a separate package because the license for JSON is "non-free")

maltheism: @driusan I'm not sure. What distro would you recommend me test on to find out?

johnsaigle: Probably CentOS since it's what we use for prod servers.

driusan: doesn't even have to be a Linux distro, a local Mac copy of php would be fine.. but CentOS is a good idea too.

maltheism: @driusan Well it doesn't seem to be an issue on macOS. I'll try testing on CentOS when I can get it setup.

maltheism: @driusan composer install does succeed on centos with the ext-json included in composer.json.

PR #4340, opened 2019-02-18
[libraries/TimePoint.class.inc] change return type to array from string

maltheism:
### Brief summary of changes

The return type for the function getBVLQCExclusionStatusList should be an array.

### DIFF
--- a/php/libraries/TimePoint.class.inc
+++ b/php/libraries/TimePoint.class.inc
@@ -597,9 +597,9 @@ class TimePoint
     /**
      * Returns the list of BVLQC exclusion statuses
      *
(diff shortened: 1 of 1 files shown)


driusan:
I'll let Travis be the judge of that..

maltheism:
> I'll let Travis be the judge of that..

@driusan I don't even thing Loris uses this function after searching the codebase.

driusan:
I wouldn't be surprised if it didn't

johnsaigle:
Yeah I mean if Travis didn't catch it the first time around I don't know that it'll catch it this time.

PR #4341, opened 2019-02-18
[CouchDB] $port should be an int not a string

maltheism:
### Brief summary of changes

The $port should be an int but $couchConfig returns a string. I added the intval function when using the $couchConfig function to retrieve the port from config.xml. I also updated the comment for the constructor to reflect an int for $port and not a string.

### DIFF
--- a/php/libraries/CouchDB.class.inc
+++ b/php/libraries/CouchDB.class.inc
@@ -199,14 +199,14 @@ class CouchDB
      *
      * @param string $database The couchDB database name
      * @param string $host     The couchDB host name
-     * @param string $port     The couchDB port number
+     * @param int    $port     The couchDB port number
      * @param string $user     The couchDB user name
      * @param string $password The couchDB password
      */
     private function __construct(
         string $database = '',
         string $host = '',
-        string $port = '',
+        int $port = null,
         string $user = '',
         string $password = ''
     ) {
@@ -260,7 +260,7 @@ class CouchDB
 
                 $database = $couchConfig['dbName'];
                 $host     = $couchConfig['hostname'];
-                $port     = $couchConfig['port'];
+                $port     = intval($couchConfig['port']);
                 $user     = $couchConfig['admin'];
                 $password = $couchConfig['adminpass'];
                 error_log(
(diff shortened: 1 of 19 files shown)


php/libraries/CouchDB.class.inc
```diff
@@ -199,14 +199,14 @@ class CouchDB
      *
      * @param string $database The couchDB database name
      * @param string $host     The couchDB host name
-     * @param string $port     The couchDB port number
+     * @param int    $port     The couchDB port number
```
johnsaigle:
```suggestion
     * @param ?int    $port     The couchDB port number
```

php/libraries/CouchDB.class.inc
```diff
@@ -199,14 +199,14 @@ class CouchDB
...
      * @param string $user     The couchDB user name
      * @param string $password The couchDB password
      */
     private function __construct(
         string $database = '',
         string $host = '',
-        string $port = '',
+        int $port = null,
```
johnsaigle:
```suggestion
        ?int $port = null,
```

php/libraries/CouchDB.class.inc
```diff
@@ -199,14 +199,14 @@ class CouchDB
...
-     * @param string $port     The couchDB port number
+     * @param ?int   $port     The couchDB port number
      * @param string $user     The couchDB user name
      * @param string $password The couchDB password
      */
     private function __construct(
         string $database = '',
         string $host = '',
-        string $port = '',
+        ?int   $port = null,
```
driusan: What does it mean to have a null port? I think this should be non-nullable and have a default of 5984 (CouchDB's default port)

ridz1208:
@driusan @maltheism this should have gone to bugfix

driusan:
@ridz1208 bugfix doesn't have any type hinting, does it?

ridz1208:
@driusan it does for this
thats why we are having these issues on releases like 20.1.2 and 20.2.0.

Not sure however why CCNA is not having that problem but others are (not sure if there is a global PHP setting for Types-strict or not...

driusan:
@ridz1208 okay, then it should have went to bugfix.. it should be a simple fix to backport.

PR #4342, opened 2019-02-18
[libraries/Project.class.inc] Allow null type hinting for setID

maltheism:
### Brief summary of changes

The php/libraries/ProjectDefault.class.inc will set the projectId to null in the constructor. The type hinting was not allowing it to be null and expecting an Int. So I added the ability for it to be assigned null.

### DIFF
--- a/php/libraries/Project.class.inc
+++ b/php/libraries/Project.class.inc
@@ -158,7 +158,7 @@ class Project
     /**
      * Set this project's id
      *
-     * @param integer $projectId This project's id
+     * @param int|null $projectId This project's id
      *
      * @return void
      */
--- a/php/libraries/Project.class.inc
+++ b/php/libraries/Project.class.inc
@@ -25,8 +25,8 @@ class Project
 {
     private static $_instances = array();
 
-    private $_projectId;
-    private $_projectName;
+    protected $_projectId;
+    protected $_projectName;
     private $_recruitmentTarget;
 
     /**
@@ -158,11 +158,11 @@ class Project
     /**
      * Set this project's id
      *
-     * @param int|null $projectId This project's id
+     * @param integer $projectId This project's id
      *
      * @return void
      */
-    public function setId(int $projectId = null): void
+    public function setId(int $projectId): void
     {
         $this->_projectId = $projectId;
     }
--- a/php/libraries/Project.class.inc
+++ b/php/libraries/Project.class.inc
@@ -162,7 +162,7 @@ class Project
      *
      * @return void
      */
-    public function setId(int $projectId): void
+    public function setId(int $projectId = null): void
     {
         $this->_projectId = $projectId;
     }
--- a/php/libraries/Project.class.inc
+++ b/php/libraries/Project.class.inc
@@ -25,8 +25,8 @@ class Project
 {
     private static $_instances = array();
 
-    protected $_projectId;
-    protected $_projectName;
+    private $_projectId;
+    private $_projectName;
     private $_recruitmentTarget;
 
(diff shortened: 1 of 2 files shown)


driusan:
I don't think I like this, it would be better to fix the constructor. But I'll let @xlecours decide if this is a good idea or not, since he wrote the class.

xlecours:
I agree with @driusan , The ProjectDefault constructor should use:
```
$this->_projectId = null;
$this->_projectName = 'loris';
```

xlecours:
But the variables are private to the superclass... they will need to be set as protected I think.

driusan:
@xlecours can we use an int like "-1" or something for the default in ProjectDefault? Or maybe even a hardcoded id of "1", since ProjectDefault should only be used if useProjects is false.

php/libraries/Project.class.inc
```diff
@@ -158,11 +158,11 @@ class Project
     /**
      * Set this project's id
      *
-     * @param int|null $projectId This project's id
+     * @param integer $projectId This project's id
```
driusan: int, not integer

xlecours:
@driusan It could produce unwanted results. If there is no ID, there should be no ID. return `getId(): ?int` makes more sense to me.

xlecours:
```
FILE: /home/travis/build/aces/Loris/php/libraries/Project.class.inc
 28 | ERROR | Protected member variable "_projectId" must not be
    |       | prefixed with an underscore
 29 | ERROR | Protected member variable "_projectName" must not be
    |       | prefixed with an underscore
FILE: ...e/travis/build/aces/Loris/php/libraries/ProjectDefault.class.inc
 33 | ERROR | [x] Equals sign not aligned with surrounding
    |       |     assignments; expected 3 spaces but found 1 space
```

driusan:
@xlecours `getId(): ?int` would fix the bug too, but then we'd have to specifically handle null every place that calls `getId()`. If useProjects is false then I don't see why it's a problem to systematically give it an arbitrary id. It's still an Id, it's just not in an id in relation to a database table.

xlecours:
@driusan Let's find out.
I hope that the check will be made before inserting new canddiates with projectid = 1 or -1 because the database would not complain.

maltheism:
@driusan @xlecours I'm trying the getId(): ?int fix. Maybe if it fails travis I should try the original commit but fix the params in the comments and see if it's okay?

PapillonMcGill:
@xlecours If you change the ProjectID attribute in the DB for `unsigned int`, then the DB will complain with -1.

xlecours:
@maltheism In ProjectDefault.class.inc if the contructor use `$this->_projectId = null` instead of `setId`, would it solve the issue?

xlecours:
@PapillonMcGill The id is meant to stay in the logic layer and never go down to persistance. The ProjectDefault class is a stub for studies that do not use Project.

maltheism:
@xlecours _projectId is private and so should I set it to public?

php/libraries/Project.class.inc
```diff
@@ -158,11 +158,11 @@ class Project
     /**
      * Set this project's id
      *
-     * @param integer $projectId This project's id
+     * @param integer|null $projectId This project's id
      *
      * @return void
      */
-    public function setId(int $projectId): void
+    public function setId(int $projectId = null): void
```
xlecours:
```suggestion
    public function setId(?int $projectId): void
```

maltheism:
Closing because ProjectDefault has been deleted. @xlecours thanks!

PR #4343, opened 2019-02-18
[Core] Fixing a Bug in ConflictDetector class

sruthymathew123:
### This PR fixes a typo and a bug introduced with this PR #4279 .
Noticed while testing another PR.

### This resolves issue...

- [ ] Redmine? #

- [ ] Github? #


driusan:
Alas, the Id is no longer Commendable :(

PR #4345, opened 2019-02-18
Add new phan rule PhanTypeSuspiciousStringExpression

johnsaigle:
This rule checks whether value is being converted to a string when it doesn't make sense to do that, such as for booleans.

### DIFF
--- a/modules/dashboard/php/dashboard.class.inc
+++ b/modules/dashboard/php/dashboard.class.inc
@@ -258,7 +258,7 @@ class Dashboard extends \NDB_Form
 
         // Behavioural Feedback Items
         if ($user->hasPermission('bvl_feedback')) {
-            $bvl_feedback = $DB->pselect(
+            $bvl_feedback      = $DB->pselect(
                 "SELECT fbt.Name, fbe.Testdate, fbe.Comment, fbth.FieldName, 
                 fbth.CommentID, fbth.SessionID, fbth.CandID, fbth.Feedback_level
                  FROM feedback_bvl_entry fbe 
--- a/modules/dashboard/php/dashboard.class.inc
+++ b/modules/dashboard/php/dashboard.class.inc
@@ -268,6 +268,7 @@ class Dashboard extends \NDB_Form
                  ORDER BY Testdate DESC LIMIT 4",
                 array()
             );
+            $frontend_feedback = array();
             foreach ($bvl_feedback as $row) {
                 if ($row['Testdate'] > $last_login) {
                     $row['new'] = 1;
@@ -293,9 +294,10 @@ class Dashboard extends \NDB_Form
                             . $row['CommentID'];
                     }
                 }
+                $frontend_feedback[] = $row;
             }
             $this->tpl_data['bvl_feedback_notifications']
-                = $bvl_feedback;
+                = $frontend_feedback;
         }
     }
 
--- a/modules/dashboard/php/dashboard.class.inc
+++ b/modules/dashboard/php/dashboard.class.inc
@@ -268,29 +268,29 @@ class Dashboard extends \NDB_Form
                  ORDER BY Testdate DESC LIMIT 4",
                 array()
             );
-            for ($i=0; $i < sizeof($bvl_feedback); $i++) {
-                if ($bvl_feedback[$i]['Testdate'] > $last_login) {
-                    $bvl_feedback[$i]['new'] = 1;
+            foreach ($bvl_feedback as $row) {
+                if ($row['Testdate'] > $last_login) {
+                    $row['new'] = 1;
                 } else {
-                    $bvl_feedback[$i]['new'] = 0;
+                    $row['new'] = 0;
                 }
-                if ($bvl_feedback[$i]['Feedback_level'] === 'profile') {
-                    $bvl_feedback[$i]['URL'] = '/' . $bvl_feedback[$i]['CandID'];
-                } else if ($bvl_feedback[$i]['Feedback_level'] === 'visit') {
-                    $bvl_feedback[$i]['URL'] = '/instrument_list/?candID='
-                        . $bvl_feedback[$i]['CandID'] . '&sessionID='
-                        . $bvl_feedback[$i]['SessionID'];
-                } else if ($bvl_feedback[$i]['Feedback_level'] === 'instrument') {
+                if ($row['Feedback_level'] === 'profile') {
+                    $row['URL'] = '/' . $row['CandID'];
+                } else if ($row['Feedback_level'] === 'visit') {
+                    $row['URL'] = '/instrument_list/?candID='
+                        . $row['CandID'] . '&sessionID='
+                        . $row['SessionID'];
+                } else if ($row['Feedback_level'] === 'instrument') {
                     $instrument = $DB->pselectOne(
                         "SELECT Test_name from flag WHERE CommentID=:cid",
-                        array('cid' => $bvl_feedback[$i]['CommentID'])
+                        array('cid' => $row['CommentID'])
                     );
                     if (!empty($instrument)) {
-                        $bvl_feedback[$i]['URL'] = '/instruments/'
+                        $row['URL'] = '/instruments/'
                             . $instrument . '/?candID='
-                            . $bvl_feedback[$i]['CandID'] . '&sessionID='
-                            . $bvl_feedback[$i]['SessionID'] . '&commentID='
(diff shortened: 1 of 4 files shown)


johnsaigle:
I'm not sure why Travis is having a problem with this one. It's working fine on my VM.

PapillonMcGill:
@johnsaigle 
line 3893+
CandidateListTestIntegrationTest::testFilterByPscid

Multiple test fail with similar document.querySelector(...) is null.

This is likely caused by some JS code getting null values instead of expected parameter in unit test. Ensure that the function noted just before (CandidateListTestIntegrationTest::testFilterByPscid in this case) pass an appropriate value for unit testing.

php/libraries/NDB_Page.class.inc
```diff
@@ -766,12 +766,12 @@ class NDB_Page implements RequestHandlerInterface
                   $baseurl . '/js/polyfills.js',
                     // $baseurl . '/js/react/react-with-addons' . $min . '.js',
                     // $baseurl . '/js/react/react-dom' . $min . '.js',
-                  $baseurl . ($min === '') ?
+                  $baseurl . ($min === '' ?
                       '/vendor/js/react/react.development.js' :
-                      '/vendor/js/react/react.production.js',
-                  $baseurl . ($min === '') ?
+                      '/vendor/js/react/react.production.js'),
```
kongtiaowang: '/vendor/js/react/react.production'.$min.'.js'), instead of '/vendor/js/react/react.production.js'),

php/libraries/NDB_Page.class.inc
```diff
@@ -766,12 +766,12 @@ class NDB_Page implements RequestHandlerInterface
...
+                  $baseurl . ($min === '' ?
                       '/vendor/js/react/react-dom.development.js' :
-                      '/vendor/js/react/react-dom.production.js',
+                      '/vendor/js/react/react-dom.production.js'),
```
kongtiaowang: '/vendor/js/react/react-dom.production'.$min.'.js'), instead of '/vendor/js/react/react-dom.production.js'),

php/libraries/Database.class.inc
```diff
@@ -108,7 +108,7 @@ class Database
             }
         } else {
             // name the connection
-            $connection = md5($database.$host.$username.$password.$trackChanges);
+            $connection = md5($database.$host.$username.$password);
```
driusan: why are you removing trackchanges from here? I think if it's removed singleton will return the same object for trackchanges true and false to the same db.

johnsaigle: phan doesn't like the idea of concatenating a boolean value to a string.

We could change it so that the value appended to the string ends up being explicitly `1` or `0` depending on the value of `$trackChanges` though

driusan: It's just constructing a string to take an md5 sum of, so maybe we can use string interpolation instead of concatenation to make phan happy?

johnsaigle: Just tried `$connection = md5("$database$host$username$password$trackChanges");` and the same warning was raised by phan.

> php/libraries/Database.class.inc:111 PhanTypeSuspiciousStringExpression Suspicious type bool of a variable or expression used to build a string. (Expected type to be able to cast to a string)

driusan: I guess building the string to pass to md5 and conditionally adding some string based on the bool instead of inlining is the only choice, then..

driusan:
still failing Travis

johnsaigle:
Thanks for reviewing @kchatpar. I confirmed the error and the latest commit should fix it. Can you verify?

driusan:
Travis did not approve.

kchatpar:
@johnsaigle I re-tested, the first two direct to an instrument. The last two re-direct to the dashboard.

> On the dashboard on the frontend, when clicking on one of the items in the Behavioral Feedback Notifications panel it doesn't direct to the instrument within the timepoint, it redirects to the main dashboard.

johnsaigle:
Yes I saw your previous comment. Glad the first 2 are working now at least. I'll look into why the others aren't.

johnsaigle:
@kchatpar From my testing it seems that the same issue exists on `major` and so the issue is perhaps with Raisinbread and not with this PR. 

Can you confirm that the same issue exists on `major` and if so, approve this PR?

kchatpar:
LGTM

johnsaigle:
Thanks @kchatpar !

php/libraries/NDB_Page.class.inc
```diff
@@ -776,12 +776,12 @@ class NDB_Page implements RequestHandlerInterface
                   $baseurl . '/js/polyfills.js',
                     // $baseurl . '/js/react/react-with-addons' . $min . '.js',
                     // $baseurl . '/js/react/react-dom' . $min . '.js',
-                  $baseurl . ($min === '') ?
+                  $baseurl . ($min === '' ?
                       '/vendor/js/react/react.development.js' :
-                      '/vendor/js/react/react.production.js',
-                  $baseurl . ($min === '') ?
+                      '/vendor/js/react/react.production' . $min . '.js'),
```
ridz1208: why is this change done ?

johnsaigle: I think there was some warning about the order of operations in PHP. Like it evaluates the `.` before the `?` so it tries to concatenate a bool to a string which phan doesn't like.

If you look at the context of the change it basically just adds `$min` or an empty string. But if `$min` IS an empty string than you can just add its value cause it comes out to be the same thing.

php/libraries/NDB_Page.class.inc
```diff
@@ -776,12 +776,12 @@ class NDB_Page implements RequestHandlerInterface
...
+                  $baseurl . ($min === '' ?
                       '/vendor/js/react/react-dom.development.js' :
-                      '/vendor/js/react/react-dom.production.js',
+                      '/vendor/js/react/react-dom.production' . $min . '.js'),
```
johnsaigle:
```suggestion
                      '/vendor/js/react/react-dom.production.js'),
```

modules/dashboard/php/dashboard.class.inc
```diff
@@ -268,34 +268,36 @@ class Dashboard extends \NDB_Form
...
-                        . $bvl_feedback[$i]['CandID'] . '&sessionID='
-                        . $bvl_feedback[$i]['SessionID'];
-                } else if ($bvl_feedback[$i]['Feedback_level'] === 'instrument') {
+                if ($row['Feedback_level'] === 'profile') {
+                    $row['URL'] = '/' . $row['CandID'];
+                } else if ($row['Feedback_level'] === 'visit') {
+                    $row['URL'] = '/instrument_list/?candID='
+                        . $row['CandID'] . '&sessionID='
+                        . $row['SessionID'];
+                } else if ($row['Feedback_level'] === 'instrument') {
```
driusan: These all look like they're changing behaviour significantly.. why is `$bvl_feedback[$i]` changed to `$row`?

johnsaigle: I think Phan gave some sort of error about getting an `int` as an array key when it was expecting another type. I can't remember exactly.

I think the behaviour should be the same though.

modules/dashboard/php/dashboard.class.inc
```diff
@@ -268,34 +268,36 @@ class Dashboard extends \NDB_Form
                  ORDER BY Testdate DESC LIMIT 4",
                 array()
             );
-            for ($i=0; $i < sizeof($bvl_feedback); $i++) {
-                if ($bvl_feedback[$i]['Testdate'] > $last_login) {
-                    $bvl_feedback[$i]['new'] = 1;
+            $frontend_feedback = array();
+            foreach ($bvl_feedback as $row) {
```
driusan:
```suggestion
            foreach ($bvl_feedback as &$row) {
```

The PHP documentation (https://www.php.net/manual/en/control-structures.foreach.php) implies that foreach passes by value and `&` is needed to make it by reference. Since this is mutating `$row`, it should be a reference to ensure the behaviour remains the same.

johnsaigle: `$bvl_feedback` is just a DB fetch that gets added to the front-end. `$frontend_feedback` keeps track of what gets modified in the for loop and then that array gets passed to the tpl data.

Why do you want to use references here? Not that it wouldn't work, but it's really unorthodox for a pretty routine LORIS thing.

PR #4347, opened 2019-02-18
Brainbrowser not working after fixing paths in violation tables

cmadjar:
### Brief summary of changes

The paths stored in the 3 violation tables (`MRICandidateErrors`, `mri_violations_log` and `mri_protocol_violated_scans`) were wrong and a tool script has been provided on the [LORIS-MRI](https://github.com/aces/Loris-MRI/pull/403) side to correct the paths to the MINC files so that they are valid paths.

Due to the bad paths being present in those tables, brainbrowser was hacked on the LORIS side to be able to display the images. Now that the paths are corrected, this hack can be removed which also simplifies the code.

This PR goes with https://github.com/aces/Loris-MRI/pull/403 on the LORIS-MRI side where more details can be found regarding the bad paths stored in the violation tables.

### This resolves issue...

- [x] Github https://github.com/aces/Loris-MRI/issues/401

### Caveat for existing projects

Make sure to run the tool script called `cleanup_paths_of_violation_tables.pl` on the LORIS-MRI side in order to update the MINC location in the violation tables.

### To test this change...

- [ ] make sure to run the tool provided on the LORIS-MRI side to update the violation tables and then try clicking on an image link in the MRI violation module to see if the image gets displayed

### DIFF
--- a/modules/brainbrowser/ajax/image.php
+++ b/modules/brainbrowser/ajax/image.php
@@ -37,17 +37,7 @@ if (strpos($_REQUEST['file_id'], 'l') !== false) {
 
     if (!empty($query)) {
         $image_file = $DB->pselectOne($query, array('LogID' => $id));
-        $file       = implode('/', array_slice(explode('/', $image_file), -2));
-
-        if (strpos($image_file, 'assembly') !== false) {
-            if (strpos($image_file, 'assembly') === 0) {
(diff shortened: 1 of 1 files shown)


cmadjar:
@ridz1208 As we discussed, I added the flag Critical to release so that this PR makes it to 20.3 along with the MRI one linked in the description. Thanks!

PR #4349, opened 2019-02-19
Update PR template

johnsaigle:
Added some tweaks to the existing PR template.

Goals:
* Less deleting of text
* More focus on testing instructions than to linked tickets
* Eliminate checkbox confusion that make PRs appear unfinished when browsing the PR tab

Let me know what you think :rainbow:


driusan:
I think we should just delete the file, but at least this is removing more than it's adding.

PR #4350, opened 2019-02-19
[Core] Use CandID class in LORIS

### LINKED ISSUE #4295: Refactor LORIS to use the new CandID class
Now that we have the CandID type in LORIS, we should be changing occurrences of CandID using `int` or `string` types to always use the new `\CandID` class.

Related: #4210, #4055, #4141

johnsaigle:
Use the CandID class for built-in verification throughout LORIS. Also cleans up code that was made obsolete by #4241.

Resolves #4295


driusan:
In the diff this seems to be doing much more than the description says, is it based on the wrong branch?

johnsaigle:
@driusan Nope these are all necessary changes to get the CandID to work. The namespacing changes are necessary for older code that don't know about the `src/` directory. In fact I think there are many more changes to make.

modules/bvl_feedback/ajax/react_get_bvl_threads.php
```diff
@@ -2,46 +2,43 @@
...
+ * PHP version 7
  *
  * @category Behavioural
  * @package  Main
  * @author   Evan McIlroy <evanmcilroy@gmail.com>
  * @license  GPLv3 <http://www.gnu.org/licenses/gpl-3.0.en.html>
  * @link     https://www.github.com/aces/Loris-Trunk/
  */
-header("content-type:application/json");
-ini_set('default_charset', 'utf-8');
```
driusan: This seems to be changing behaviour outside of what's required for CandID

modules/bvl_feedback/ajax/react_get_bvl_threads.php
```diff
@@ -2,46 +2,43 @@
...
+namespace LORIS\bvl_feedback;
+use \LORIS\StudyEntities\Candidate\CandID;
 
-set_include_path(
-    __DIR__ . "/../../../project/libraries:" .
-    __DIR__ . "/../../../php/libraries:"
-);
```
driusan: This too

modules/bvl_feedback/ajax/react_get_bvl_threads.php
johnsaigle: Yeah okay I can revert these.

modules/bvl_feedback/ajax/react_get_bvl_threads.php
```diff
@@ -2,46 +2,48 @@
...
  * @license  GPLv3 <http://www.gnu.org/licenses/gpl-3.0.en.html>
  * @link     https://www.github.com/aces/Loris-Trunk/
  */
+namespace LORIS\bvl_feedback;
+use \LORIS\StudyEntities\Candidate\CandID;
 header("content-type:application/json");
 ini_set('default_charset', 'utf-8');
 
 set_include_path(
-    __DIR__ . "/../../../project/libraries:" .
```
driusan: `__DIR__` doesn't have a trailing slash (https://secure.php.net/manual/en/language.constants.predefined.php), so this is wrong

modules/bvl_feedback/ajax/react_get_bvl_threads.php
```diff
@@ -2,46 +2,48 @@
...
-    __DIR__ . "/../../../php/libraries:"
+    __DIR__ . "../../../project/libraries:" .
```
johnsaigle:
```suggestion
    __DIR__ . "/../../../project/libraries:" .
```

modules/bvl_feedback/ajax/react_get_bvl_threads.php
```diff
@@ -2,46 +2,48 @@
...
  */
+namespace LORIS\bvl_feedback;
+use \LORIS\StudyEntities\Candidate\CandID;
 header("content-type:application/json");
 ini_set('default_charset', 'utf-8');
 
 set_include_path(
     __DIR__ . "/../../../project/libraries:" .
-    __DIR__ . "/../../../php/libraries:"
+    __DIR__ . "../../../php/libraries:"
```
johnsaigle:
```suggestion
    __DIR__ . "/../../../php/libraries:"
```

modules/bvl_feedback/ajax/bvl_panel_ajax.php
```diff
@@ -15,25 +15,28 @@
...
 
-$user     =& User::singleton();
-$username = $user->getUsername();
+namespace LORIS\bvl_feedback;
+use \LORIS\StudyEntities\Candidate\CandID;
+
+$username = \User::singleton()->getUsername();
+$candID   = new CandID($_POST['candID']);
 
 if (isset($_POST['candID']) && (empty($_POST['sessionID']))) {
```
maltheism: @johnsaigle The code is checking here if $_POST['candID'] is set. Although above you use it for making a new CandID and without checking if it's set.

johnsaigle: @maltheism I believe this is resolved now. Can you re-review?

maltheism:
Blocking this PR until #4416 is merged.

modules/bvl_feedback/ajax/bvl_panel_ajax.php
```diff
@@ -15,25 +15,30 @@
...
-$user     =& User::singleton();
-$username = $user->getUsername();
+namespace LORIS\bvl_feedback;
+use \LORIS\StudyEntities\Candidate\CandID;
+
+$username = \User::singleton()->getUsername();
 
 if (isset($_POST['candID']) && (empty($_POST['sessionID']))) {
-    $feedbackThread =& NDB_BVL_Feedback::Singleton($username, $_POST['candID']);
+    $candID         = new CandID($_POST['candID']);
```
PapillonMcGill: `$_POST['candID']` not sanitized.
Did some code disappeared in a rebase?

johnsaigle: The CandID class automatically throws an exception when the string input is not numeric and is not exactly 6 characters in length. I don't believe it's possible to create a malicious JS injection attack using numbers so this input can be considered safe.

PapillonMcGill: In that case, I will clear my opposition and approve this PR.

php/libraries/NDB_BVL_Feedback.class.inc
```diff
@@ -63,15 +64,15 @@ class NDB_BVL_Feedback
...
+     * @param ?CandID $candID    The 6 digit CandID
      * @param integer $sessionID The session ID with feedback
      * @param string  $commentID The commentID receiving feedback.
      *
      * @return NDB_BVL_Feedback object
      */
     static function &singleton(
         $username,
-        $candID=null,
+        ?CandID $candID=null,
```
driusan: Why did this become nullable?

johnsaigle: The default param was `null` so I just kept it the same but added a type hint. I'm not sure whether there is a valid case for CandID to be null or not.

driusan: @ridz1208 do you know if there are any valid cases for a feedback thread to have a null candid? I can't think of any, but I'm not that familiar with it.

kchatpar:
@johnsaigle Could you add some testing instructions to the PR description?

modules/bvl_feedback/ajax/react_get_bvl_threads.php
```diff
@@ -19,32 +21,34 @@
...
+    $candID         = new CandID($_POST['candID']);
     $feedbackThread =&
-         NDB_BVL_Feedback::Singleton(
-             $username,
-             $_POST['candID'],
-             $_POST['sessionID']
-         );
+        \NDB_BVL_Feedback::Singleton(
+            $username,
+            $candID,
```
kchatpar: Why isn't $username and $candID using $_Post to get it's information?

johnsaigle: `$candID` is populated by `$_POST` on lines 29 and 34.

In general it's not a good idea (at least by our team) to access PHP superglobals such as `$_POST` directly. We're trying to move away from it.

johnsaigle:
@kchatpar I'm not really able to provide more detailed testing instructions. This is a kind of architectural change that covers many different modules in many ways.... I'm not sure I could provide an exhaustive list of where to test.

IMO If the integration tests are passing then it means things are working well. Maybe @driusan could weigh in here to say whether this is sufficient.

driusan:
I guess the only way to test it is to manually test each of the modules that are touched by it, but since it's the major branch tests passing and a couple of them manually being tested is probably sufficient.

johnsaigle:
@kchatpar Maybe following the TestPlans for `instrument_list` and `timepoint_list` is the best way to test then.

driusan:
and bvl_feedback

php/libraries/NDB_BVL_Feedback.class.inc
driusan: @ridz1208 do you know if there are any valid cases for a feedback thread to have a null candid? I can't think of any, but I'm not that familiar with it.

ridz1208: logically no, the highest place you can put a feedback is at the candidate level... so no

php/libraries/Candidate.class.inc
```diff
@@ -42,40 +43,40 @@ class Candidate
...
      *              object
      * @throws DatabaseException | LorisException
      */
-    public function select(int $candID): void
+    public function select(CandID $candID): void
     {
         $factory = NDB_Factory::singleton();
         $config  = $factory->config();
 
-        $CandArray = array('Candidate' => $candID);
```
ridz1208: @driusan is this the best way to get the `value` variable of the CanID object ? we dont want to have a `getValue()` function ?

modules/timepoint_list/php/timepoint_list.class.inc
```diff
@@ -84,7 +85,7 @@ class Timepoint_List extends \NDB_Menu
         ServerRequestInterface $request,
         RequestHandlerInterface $handler
     ) : ResponseInterface {
-        $attribute = $request->getAttribute("CandID");
+        $attribute = new CandID($request->getAttribute("CandID"));
```
ridz1208: shouldn't line 94 do the same thing ?? `new CandID()...`

src/Middleware/UserPageDecorationMiddleware.php
```diff
@@ -58,6 +59,7 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
             $candID = $get['candID'];
```
ridz1208: is CandID here of type candID ??

ridz1208:
@johnsaigle I briefly looked at the changes but I quickly realized that not all instances of calls to `Candidate::singleton()` were modified to not throw a type error.

- APIs 0.0.2 and 0.0.3-dev classes still contain some calls (@xlecours should those even be changed ?)
- tools
   - assign_missing_instruments.php
   - fix_timepoint_date_problems.php
- libraries
   - NDB_Factory.class.inc
- modules
   - candidate_parameters/ajax

driusan:
@johnsaigle not sure how much have changed since you removed "Needs Rebase" 1 hour ago, but it still has conflicts (again?) according to GitHub

ridz1208:
@johnsaigle I saw you made all the necessary changes, the only missing ones are the API and I'm not sure how these should be handled, I'll leave it to @xlecours and @driusan to comment on those.

As far as review goes, at first glance it looks good. I'l try to do a round of testing to see how it behaves. I'll approve the code after diving a bit more into the dtails

php/libraries/Candidate.class.inc
```diff
@@ -42,40 +43,40 @@ class Candidate
...
      *
-     * @param integer $candID A 6 digit Loris CandID that matches
+     * @param CandID $candID A 6 digit Loris CandID that matches
      *                        a candidate in the candidate table
      *
      * @return Candidate the Candidate object
      * @static
      */
-    static public function &singleton(int $candID)
+    static public function &singleton(CandID $candID)
```
xlecours: to add return type `Candidate`

johnsaigle:
@xlecours I'm happy to undo my changes to the API rather than interfere with the work you've done.

xlecours:
@johnsaigle I can see conflicts where the API is making calls to php/libraries classes that their function signature have changed. But it should not be a big deal.

modules/candidate_list/php/validateids.class.inc
```diff
@@ -84,32 +85,29 @@ class ValidateIDs extends \NDB_Page
...
-            }
-            // Return "0" or "1" based on whether it exists or not.
-            // We always return a 200 error code and not a 404, because
-            // even if the CandID doesn't exist or match, the validateids
-            // endpoint still exists and has a valid response.
-            $exists = \Candidate::candidateExists(
-                $gets['CandID'],
-                $gets['PSCID']
-            ) ? "1" : "0";
+        if (!isset($gets['CandID'], $gets['PSCID'])) {
```
ridz1208: nice !! i didnt know that was possible

modules/instrument_list/php/module.class.inc
```diff
@@ -89,6 +91,7 @@ class Module extends \Module
             $message = 'Could not find candidate information for candidate '
                 . htmlspecialchars($gets['candID']) . ' and session '
                 . htmlspecialchars($gets['sessionID']) . '.';
+            error_log($message . ' ' . $e->getMessage());
```
ridz1208: is this a debugging statement or do you mean to have it permanently in the code ?

modules/timepoint_list/php/timepoint_list.class.inc
```diff
@@ -84,14 +85,11 @@ class Timepoint_List extends \NDB_Menu
...
-            // FIXME: Update the code to add the singleton here, not the identifier.
-            $this->candID = $attribute;
-        } else {
-            $gets         = $request->getQueryParams();
-            $this->candID = $gets['candID'];
-        }
+        // FIXME: Update the code to add the singleton here, not the identifier.
+        $this->candID = new CandID(
+            $request->getAttribute('CandID')
+            ?? $request->getAttribute('candID')
```
ridz1208: shouldn't this be `?? $request->getQueryParams()['candID']` ??

php/libraries/Candidate.class.inc
```diff
@@ -203,8 +203,8 @@ class Candidate
         ?string $edc,
         string $sex,
         ?string $PSCID = null,
-        int $projectID = null
-    ): int {
+        ?int $projectID = null
```
ridz1208: Now that use projects is mandatory, this should not have a `?` however if it is causing you issues, fixing it would be outside of the scope of your PR and should be reported as an issue on github

php/libraries/NDB_BVL_Feedback.class.inc
ridz1208: I'm okai with this given that it maintains status quo however if should be removed in the future

php/libraries/NDB_BVL_Feedback.class.inc
```diff
@@ -112,19 +120,21 @@ class NDB_BVL_Feedback
...
+     * @param string  $username  The name of the currently logged in user.
      *                          (why is this here?)
-     * @param string $candID    The CandID we're getting a feedback thread for.
-     *                          (optional)
-     * @param string $sessionID The sessionID we're getting a feedback thread for.
-     *                          (optional)
-     * @param string $commentID The commentID we're getting a feedback thread for.
-     *                          (optional)
+     * @param ?CandID $candID    The CandID we're getting a feedback thread for.
+     * @param ?string $sessionID The sessionID we're getting a feedback thread for.
```
ridz1208: doc doesnt match actual parameters (missing `?`)

test/unittests/CandidateTest.php
```diff
@@ -269,161 +254,6 @@ public function testGetDataReturnsAllInformationIfGivenNull()
         );
     }
 
-    /**
```
ridz1208: why are all these tests deleted ?

ridz1208:
@johnsaigle just a few comments/questions to help me understand more the PR

modules/timepoint_list/php/timepoint_list.class.inc
driusan: or better: fix the place where the capitalization is wrong?

modules/candidate_list/php/validateids.class.inc
johnsaigle: Yes it's much nicer 😄

modules/instrument_list/php/module.class.inc
johnsaigle: I can't remember... It can probably be deleted.

johnsaigle:
@ridz1208 I believe I applied all of your suggestions, including restoring the deleted unit tests. Could you re-review?

modules/timepoint_list/php/timepoint_list.class.inc
```diff
@@ -84,14 +85,8 @@ class Timepoint_List extends \NDB_Menu
...
-        $attribute = $request->getAttribute("CandID");
-        if ($attribute !== null) {
-            // FIXME: Update the code to add the singleton here, not the identifier.
-            $this->candID = $attribute;
-        } else {
-            $gets         = $request->getQueryParams();
-            $this->candID = $gets['candID'];
-        }
+        // FIXME: Update the code to add the singleton here, not the identifier.
+        $this->candID = new CandID($request->getAttribute('CandID'));
```
ridz1208: @johnsaigle this is my last point of concern, if you are removing the if/else, whats handling the logic now ?

johnsaigle: I'm not sure... I don't know where the old lowercase `candID` (vs `CandID`) was being sent and was relying on the integration tests to catch it. But there wasn't any error. I'm wondering if the if-else was a hack to manage a typo from somewhere else which has since been fixed.

driusan: IIRC it was to handle both `$URL/$candid/` and `$URL/main.php?test_name=timepoint_list&candID=$candid`, but it was a while ago so I might be misremembering..

johnsaigle: Ok thanks. Are these paths still a concern?

driusan: If there is I think we should fix the URL to point to the `$URL/$candid` one.. if that's working (and I'm not misremembering why it was needed), I don't think it should be a concern.

ridz1208: given that its going to major which means it will undergo a full round of testing... im fine with it

ridz1208:
@johnsaigle I had approved this one right ?

johnsaigle:
@ridz1208 Yes I think so.

PR #4358, opened 2019-02-22

maltheism:
### Brief summary of changes

**Note:** I changed how the visit label is handled. It's now done on the frontend and so there is no delay for the user. 

### This resolves issue...

- [x] Redmine? #16008

### To test this change...


maltheism:
I'm updating this PR to move the ajax code to the create_timepoint php directory. Please wait on reviewing.

HenriRabalais:
Waiting on PR to be update to review.

(putting this here because I was assigned this PR as a review priority)

maltheism:
@HenriRabalais The updated changes to use the ServerRequestInterface have been added. So the PR can be reviewed.

modules/create_timepoint/jsx/create_timepoint.js
```diff
@@ -0,0 +1,355 @@
...
+        disabled={false}
+        required={true}
+      />
+    ) : '';
+    // Include visit select.
+    const visit = this.state.form.display.visit ? (
+      <SelectElement
+        id={'visit'}
+        name={'visit'}
+        ref={'visit'}
```
HenriRabalais: String refs are deprecated.

modules/create_timepoint/jsx/create_timepoint.js
```diff
@@ -0,0 +1,355 @@
...
+        disabled={false}
+        required={true}
+      />
+    ) : '';
+    // Include psc select.
+    const psc = this.state.form.display.psc ? (
+      <SelectElement
+        id={'psc'}
+        name={'psc'}
+        ref={'psc'}
```
HenriRabalais: String refs are deprecated.

modules/create_timepoint/jsx/create_timepoint.js
```diff
@@ -0,0 +1,355 @@
...
+      return <Loader/>;
+    }
+    // Include form errors.
+    const errors = this.state.errors;
+    // Include subproject select.
+    const subproject = this.state.form.display.subproject ? (
+      <SelectElement
+        id={'subproject'}
+        name={'subproject'}
+        ref={'subproject'}
```
HenriRabalais: String refs are deprecated.

modules/create_timepoint/jsx/create_timepoint.js
```diff
@@ -0,0 +1,355 @@
...
+          // Populate the form errors.
+          if (data.errors && data.errors.length > 0) {
+            this.setState({errors: data.errors});
+          }
+          // Populate the select options for subproject.
+          if (data.subproject) {
+            this.state.form.options.subproject = data.subproject;
+            this.state.form.value.subproject = Object.keys(data.subproject)[0];
+            this.state.form.display.subproject = true;
+            this.setState(this.state);
```
HenriRabalais: `124`-`126` Direct mutation of state.

modules/create_timepoint/jsx/create_timepoint.js
```diff
@@ -0,0 +1,355 @@
...
+    if (!this.state.success) {
+      return (
+        <div>
+          <div>
+            <h3>Create Time Point</h3> <br/>
+            {errors}
+            <FormElement
+              name={'timepointInfo'}
+              fileUpload={false}
+              ref={'form'}
```
HenriRabalais: String refs are deprecated.

modules/create_timepoint/jsx/create_timepoint.js
```diff
@@ -0,0 +1,355 @@
...
+
+  // Create a wrapper div in which React component will be loaded
+  const browserDOM = document.createElement('div');
+  browserDOM.id = 'page_create_timepoint';
+
+  // Append wrapper div to page content
+  const rootDOM = document.getElementById('lorisworkspace');
+  rootDOM.appendChild(browserDOM);
+
+  ReactDOM.render(createTimepoint, document.getElementById('page_create_timepoint'));
```
HenriRabalais: I'm not sure if all this wrapping is necessary. I think the app can be loaded directly into `lorisworkspace`.

This can be written as:
```
  ReactDom.render(
    <CreateTimepoint
      dataURL={`${loris.BaseURL}/create_timepoint/AjaxTimepoint`}
    />,
    document.getElementById('lorisworkspace')
});
```

modules/create_timepoint/jsx/create_timepoint.js
```diff
@@ -0,0 +1,355 @@
...
+  handleSubmit(e) {
+    const send = this.urlSearchParams({
+      command: 'create',
+      candID: this.state.url.params.candID,
+      identifier: this.state.url.params.identifier,
+      subproject: this.state.form.value.subproject,
+      psc: this.state.form.value.psc,
+      visit: this.state.form.value.visit,
+    });
+    const url = this.props.DataURL + '/create_timepoint/AjaxTimepoint';
```
HenriRabalais: indexes should be lowerCamelCase (i.e. `this.props.dataURL`).

modules/create_timepoint/jsx/create_timepoint.js
```diff
@@ -0,0 +1,355 @@
...
+   * Retrieve data from the provided URL and save it in state.
+   */
+  fetchInitializerData() {
+    const send = this.urlSearchParams({
+      command: 'initialize',
+      candID: this.state.url.params.candID,
+      identifier: this.state.url.params.identifier,
+      subprojectID: this.state.form.subproject,
+    });
+    const url = this.props.DataURL + '/create_timepoint/AjaxTimepoint';
```
HenriRabalais: indexes should be lowerCamelCase (i.e. `this.props.dataURL`).

modules/create_timepoint/jsx/create_timepoint.js
```diff
@@ -0,0 +1,355 @@
...
+  DataURL: PropTypes.string,
+};
+
+/**
+ * Render create_timepoint on page load.
+ */
+window.onload = function() {
+  const createTimepoint = (
+    <CreateTimepoint
+      Module='create_timepoint'
```
HenriRabalais: I don't think this prop is necessary.

johnsaigle:
This is failing Travis's integration tests. Make sure to double check that the module is working locally. If so, it's possible that the integration tests needs to be updated to match the new code.

modules/create_timepoint/jsx/create_timepoint.js
maltheism: @HenriRabalais I found that Module property included in previous modules using react. I guess it doesn't really serve a purpose.

maltheism:
@johnsaigle The module works on my VM. I believe it's the integration tests.

modules/create_timepoint/jsx/create_timepoint.js
maltheism: Thanks @HenriRabalais that's way better!

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,292 @@
...
+                    new \LORIS\Http\StringStream(
+                        'Only POST requests are allowed'
+                    )
+                );
+        }
+
+        /**
+         * Parse POST request body.
+         */
+        $_POST = $request->getParsedBody();
```
johnsaigle: This goes against the purpose of using the PSR7 implementations. You'd be better off accessing these vairables using `$request->getAttribute()`.

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,292 @@
...
+        $user       =& \User::singleton();
+        $identifier = isset($_POST['identifier']) ? $_POST['identifier'] : '';
+        $candidate  =& \Candidate::singleton($identifier);
+        if (!$user->hasPermission('data_entry')
+            && !(in_array(
+                $candidate->getData('RegistrationCenterID'),
+                $user->getData('CenterIDs')
+            ))
+        ) {
+            header('HTTP/1.1 403 Forbidden');
```
johnsaigle: Should return using a Loris\Http\Response, not `header()` + `exit`

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: Hey @johnsaigle I messaged you on slack about this and maybe you haven't seen it yet. I'm confused by your comment because the getParsedBody() is from the the ServerRequestInterface and I think is part of the PSR7 interface. The getParsedBody() returns an array of what is sent in the POST javascript fetch method and pertaining the the "body" key.

driusan: The `$request->getParsedBody` is fine, I think it's the `$_POST` he has a problem with since it's a special PHP superglobal name. I think it should be fine (and less confusing) if you just give the variable a different name.

maltheism: Thank you @driusan, that makes sense. I just copied the old code from the previous commits and which is why the variable is named $_POST. I'll change it to something else.

johnsaigle:
Failing due to phan issues:
```
modules/create_timepoint/php/ajaxtimepoint.class.inc:227 PhanTypeArraySuspicious Suspicious array access to ?mixed
modules/create_timepoint/php/ajaxtimepoint.class.inc:232 PhanTypeArraySuspicious Suspicious array access to ?mixed
```

maltheism:
@johnsaigle do you happen to know why I'm getting that error because I'm using the Utility::toArray and the code is the same for on major in the create_timepoint.class.inc.

johnsaigle:
Hmm not sure. `PhanTypeArraySuspicious` was only just added and sometimes Travis doesn't catch everything. 

Maybe try wrapping both levels of the array in `toArray`? My _suspicion_ is that the issue is happening because of the nested array.

maltheism:
@johnsaigle Okay the PhanTypeArraySuspicious seems to be resolved.

maltheism:
@kongtiaowang can you help with the travis test cases.

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,302 @@
...
+
+        /**
+         * Parse POST request body.
+         */
+        $postResponse = $request->getParsedBody();
+
+        /**
+         * User permission verification.
+         */
+        $user       =& \User::singleton();
```
xlecours: `$request->getAttribute('user')` would be better. It is added in BaseRouter.php

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,302 @@
...
+                    new \LORIS\Http\StringStream(
+                        'Only POST requests are allowed'
+                    )
+                );
+        }
+
+        /**
+         * Parse POST request body.
+         */
+        $postResponse = $request->getParsedBody();
```
xlecours: `postResponse` name do not fit it's content. `requestBody` or `requestParams` maybe?

kongtiaowang:
When a candidate only has one site, it can't choose a site for this candidate to Create Time Point.
screen shot 2019-03-01 at 3 43 25 pm

maltheism:
Hi @kongtiaowang, I fixed that error and thanks for finding it!

maltheism:
Okay new changes added.

modules/create_timepoint/jsx/create_timepoint.js
```diff
@@ -0,0 +1,352 @@
...
+  urlSearchParams(json) {
+    return Object.keys(json).map((key) => {
+      return encodeURIComponent(key) + '=' + encodeURIComponent(json[key]);
+    }).join('&');
+  }
+  /**
+   * Retrieve data from the provided URL and save it in state.
+   */
+  fetchInitializerData() {
+    const state = Object.assign({}, this.state);
```
zaliqarosli: i dont think you need to define the state const here because you do it again down below in the if statements

modules/create_timepoint/jsx/create_timepoint.js
```diff
@@ -0,0 +1,352 @@
...
+    Object.keys(values).forEach(function(key) {
+      errors.push(
+        <div className='col-xs-12 col-sm-12 col-md-12'>
+          <label className='error form-group'>
+            {values[key]}
+          </label>
+        </div>
+      );
+    });
+    this.setState({errors: errors});
```
zaliqarosli: what data type is this.state.errors? it seems weird to me that it's being set here as an array of jsx. they seem to be strings elsewhere in loris

modules/create_timepoint/jsx/create_timepoint.js
```diff
@@ -0,0 +1,352 @@
...
+        name={'subproject'}
+        label={'Subproject'}
+        value={this.state.form.value.subproject}
+        options={this.state.form.options.subproject}
+        onUserInput={this.setForm}
+        emptyOption={false}
+        disabled={false}
+        required={true}
+      />
+    ) : '';
```
zaliqarosli: would it maybe be better to set as null where the empty string is?

webpack.config.js
```diff
@@ -16,6 +16,7 @@ const config = [{
     './modules/issue_tracker/js/index.js': './modules/issue_tracker/jsx/index.js',
     './modules/candidate_parameters/js/CandidateParameters.js': './modules/candidate_parameters/jsx/CandidateParameters.js',
     './modules/configuration/js/SubprojectRelations.js': './modules/configuration/jsx/SubprojectRelations.js',
+    './modules/create_timepoint/js/create_timepoint.js': './modules/create_timepoint/jsx/create_timepoint.js',
```
zaliqarosli: i suggest renaming the file to `createTimepointIndex.js` for no reason other than consistency :)

modules/create_timepoint/php/create_timepoint.class.inc
```diff
@@ -22,27 +22,20 @@ namespace LORIS\create_timepoint;
...
+    protected $candidate;
 
     /**
-     * Check user permissions
+     * Determine whether the user has permission to view this page
      *
      * @param \User $user The user whose access is being checked
      *
      * @return bool
+     * @throws \LorisException
```
zaliqarosli: is the LorisException thrown from the hasPermission function?

modules/create_timepoint/jsx/create_timepoint.js
maltheism: @zaliqarosli I've never used null in this situation here. I've seen empty string used elsewhere online when looking at react code and so I sort of think empty string is preferred. Thoughts?

webpack.config.js
maltheism: Yah I saw the other day we have a template for react that @HenriRabalais sent me in another PR. Thanks @zaliqarosli I'll change it.

modules/create_timepoint/php/create_timepoint.class.inc
maltheism: @zaliqarosli my IDE suggests it and I just allowed it to be autogenerated in the comment because otherwise my IDE shows ugly squiggle colours around the comment. Do you want me to remove it? I believe you're right and it's from the hasPermission function.

modules/create_timepoint/jsx/create_timepoint.js
```diff
@@ -0,0 +1,352 @@
...
+
+    if (!this.state.success) {
+      return (
+        <div>
+          <div>
+            <h3>Create Time Point</h3> <br/>
+            {errors}
+            <FormElement
```
zaliqarosli: to keep consistent with recent UX group UI decisions, I would suggest nesting the <FormElement /> with a <Panel /> element with a title prop "Create Time Point" and remove the h3 tag

modules/create_timepoint/php/create_timepoint.class.inc
zaliqarosli: I'm not sure. I just don't see a "throw LorisException" in the code so I was wondering if it actually does throw a LorisException but i'm sure your IDE is trustworthy haha

modules/create_timepoint/jsx/create_timepoint.js
zaliqarosli: apparently we should be using false, null, or undefined https://stackoverflow.com/questions/30097091/correct-way-to-define-an-empty-dom-element-in-react

modules/create_timepoint/jsx/create_timepoint.js
maltheism: I'm using the state const to access for building the json that's send to the server. I think its alight.

modules/create_timepoint/jsx/create_timepoint.js
maltheism: The errors should be just a string from the server.

modules/create_timepoint/jsx/create_timepoint.js
zaliqarosli: oh i see where you use it just below

modules/create_timepoint/jsx/create_timepoint.js
zaliqarosli: isn't it an array of jsx here however?

maltheism: @zaliqarosli Oh yes you're correct. Should I rename errors to something else. I guess multiple errors could be returned and were stacked on top of each other. Should I name errors to something else?

modules/create_timepoint/js/.gitignore
```diff
@@ -0,0 +1 @@
+create_timepoint.js
```
zaliqarosli: needs updating. i think doing *.js as suggested by @driusan in a previous PR is a good idea

modules/create_timepoint/jsx/create_timepoint.js
zaliqarosli: the issue here is that this.state.errors is defined as a boolean (false), and then set as a string from 'data.errors' (line 123), and then set as an array of jsx in populateErrors() (line 202). i think this.state.errors should either be a boolean (true or false that there is an error) or a string (the error message itself), and any rendering of the error is instead done outside of the state. does this make sense?

modules/create_timepoint/js/.gitignore
zaliqarosli: needs updating. i think doing *.js as suggested by @driusan in a previous PR is a good idea

zaliqarosli:
on testing: the select elements should have an empty option so that a default option isn't populated before user input

maltheism:
Hi @zaliqarosli, I made the changes.

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,309 @@
...
+    public $skipTemplate = true;
+
+    /**
+     * This function will return a json file with DCCID and PSCID
+     *
+     * @param ServerRequestInterface $request The incoming PSR7 request
+     *
+     * @return ResponseInterface The outgoing PSR7 response
+     * @throws \DatabaseException
+     * @throws \DeprecatedException
```
johnsaigle: Does it?

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,309 @@
...
+            for this study. If you are an administrator, please use the 
+            Configuration module to add new subprojects.';
+        }
+        // List of valid subprojects for a given project
+        if ($config->getSetting('useProjects') === 'true') {
+            // If projects enabled, loop through and add valid subprojects only
+            $subprojList = '';
+            try {
+                $subprojList = $candidate->getValidSubprojects();
+            } catch (\DatabaseException $e) {
```
johnsaigle: Can you explain why you're silencing the exception here?

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,309 @@
...
+        if (empty($values['psc'])) {
+            unset($values['psc']);
+        }
+
+        return $values;
+    }
+
+    /**
+     * Validate the post data,
+     * from user user form selection.
```
johnsaigle: This should be cleaned up.

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: @johnsaigle My IDE believes it's possible because the functions inside the handle(){...} can throw those exceptions. I personally don't believe it will ever happen. The reason I add those @throws tags is because my IDE shows an ugly squiggly line around the function comment if I don't add them.

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,309 @@
...
+     *
+     * @param array $values the form values.
+     *
+     * @return array
+     * @throws \LorisException
```
johnsaigle: This doesn't throw a LorisException

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: @johnsaigle the DatabaseException if thrown would pollute the return JSON object and break the frontend (I believe). The line after checks if subprojList is empty and handles the error message that is displayed to the user.

modules/create_timepoint/php/create_timepoint.class.inc
```diff
@@ -59,203 +52,16 @@ class Create_Timepoint extends \NDB_Form
...
     /**
      * Generate a breadcrumb trail for this page.
      *
      * @return \LORIS\BreadcrumbTrail
      */
     public function getBreadcrumbs(): \LORIS\BreadcrumbTrail
     {
-        $candid = $this->candidate->getCandID();
-        $pscid  = $this->candidate->getPSCID();
+        $candidate =& \Candidate::singleton($this->identifier);
```
johnsaigle:
```suggestion
        $candidate = \Candidate::singleton($this->identifier);
```

This is the preferred style

modules/create_timepoint/test/create_timepointTest.php
```diff
@@ -51,113 +60,104 @@ function tearDown()
...
      *
      * @param string $canID      ID of candidate
-     * @param string $subproject text of subproject
-     * @param string $visitlabel text of visit label
+     * @param string $subproject index of subproject
+     * @param string $visitlabel index of visit label
      *
      * @return void
      */
     private function _createTimepoint($canID, $subproject, $visitlabel)
```
johnsaigle: Probably a good oppotunity to change this to `$candID`, not `$canID`.

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: @johnsaigle `\TimePoint::isValidVisitLabel($candid, $subprojectID, $visitLabel);` can possibly throw the error.

modules/create_timepoint/php/ajaxtimepoint.class.inc
johnsaigle: This is maybe a philosophical point about how Exceptions should be handled and annotated... we should discuss it as a team.

modules/create_timepoint/php/ajaxtimepoint.class.inc
johnsaigle: It should go inside the `catch` then, I think.

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,309 @@
...
+        $visit_options      = array();
+        $visitLabelSettings = $config->getSetting('visitLabel');
+        foreach (
+            \Utility::toArray($visitLabelSettings) as $visitLabel
+        ) {
+            if (!empty($values['subproject'])) {
+                if (isset($visitLabel['generation'])
+                    && $visitLabel['generation'] !== 'sequence'
+                ) {
+                    throw new \DeprecatedException(
```
johnsaigle: This has been Deprecated for a while. We should make it so that it's not possible to do this from the front-end and then delete this exception.

modules/create_timepoint/php/ajaxtimepoint.class.inc
johnsaigle: But you catch that function call so it can't actually throw this exception

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: @johnsaigle I'll bring it up at the next meeting. I hope these @throw tags can be added because it cleans up my IDE warnings.

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: @johnsaigle I'm not actually sure what visitLabel['generation'] is used for doing. Should I just delete this code?

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: @johnsaigle yes but I believe the point of the @throws are for making the process of debugging a problem simpler. Since you can navigate from method to method and find where a problem is because an Exception is usually what breaks functionality. Specifically if we're logging things correctly and @throws help in navigating the code as well.

modules/create_timepoint/php/ajaxtimepoint.class.inc
johnsaigle: On the other hand you could configure the settings in your IDE :D

Either way we should figure this out as a team. It might end up going this way where we document every exception a function could throw including the functions it calls. However that's not how things are now.

modules/create_timepoint/php/ajaxtimepoint.class.inc
johnsaigle: Yeah but it doesn't actually throw the Exception lol. It feels very misleading to me.

I understand it can help you navigate in your IDE but I just use Vim and several others do also so your example doesn't really provide that much utility to several of us. 

I think this is part of the discussion to be had in the LORIS meeting that we mentioned above, don't worry about this until a decision is made there.

modules/create_timepoint/php/ajaxtimepoint.class.inc
johnsaigle: I think so, as long as it doesn't break anything.

modules/create_timepoint/php/ajaxtimepoint.class.inc
driusan: I also think you should configure your IDE.

maltheism: @johnsaigle @driusan I believe my IDE is correct to display these @throws warnings. I can look into disabling it for LORIS specifically but it's beneficial from my viewpoint to know the method will possibly throw an exception and even if the function inside is responsible; no matter how far down the chain of functions it goes.

maltheism: @johnsaigle @driusan I also know nobody seems to care because everybody is coding with vim.

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: Hey @johnsaigle, I looked at the method getValidSubprojects() and it no longer throws the DatabaseException. So I removed the catch in the code and the @throw tag from the method comment tag as well.

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: @johnsaigle I understand and I don't necessarily use the @throw tags myself and when debugging where an exception is thrown. Typically I use breakpoints to see what is causing the code to break. The only point I would voice at the meeting in favour of @throw tags in functions that use functions that possibly throw and exception is how I think it's best practice because it's helpful in the sense of knowing wherever the starting point is where an possible exception can be thrown. To me it's similar to the php functions type hinting in the sense that it's optional but useful in a small way.

maltheism: @johnsaigle also I think the comments in general for functions in LORIS need to be cleaned up. I've noticed some core functions with comments including @throw exceptions and it's no longer the case anymore. I'll make a PR in the future to clean some of it up.

modules/create_timepoint/jsx/create_timepoint.js
maltheism: Hey @zaliqarosli is it alright if I make the state.errors be null and it becomes the divs with error messages to be shown when errors are returned from the server. Please see new code changes. I think it's simpler this way and I don't know if rendering of the error would be better if done outside of the state.

zaliqarosli: i think its still best practice to put in state the most minimal amount of data needed to represent your UI's state, leaving all render props to be in the render function. this way, the rendering of the error is only defined once, and only the error data gets re-rendered on change. because of this, i would argue that rendering of the error is definitely better done inside the render function (it is what it's made for), and not as data passed from the state object

maltheism: @zaliqarosli I'm a little confused. The error(s) from the server are stored in the state and so it will re-render the change of the errors to display in the render function. The error(s) can change depending on the user submitting the form submission and the depending on the error(s)-case in question. I'm using the state to trigger when things are to be displayed as well. I guess since I'm storing possibly multiple divs (depending on single error or multiple) is what's upsetting. That's sort of the reason why it ended up the way it did. Are you suggesting I place the function in render and have it be triggered by an array that's stored in state?

zaliqarosli: im not sure i fully understand but what I am trying to say is that (referring to lines 196-204) we shouldn't be storing jsx elements in state. if you are trying to display multiple errors, then the errors prop should be an array of string errors. how this array of errors is rendered should be decided inside a render function. so perhaps the div of errors should be inserted on line 305. 

in other words, you are storing divs of error labels in state, and then rendering that state data. what you should be doing is storing an array of string errors in state, and then rendering divs of labels of that state data.

maltheism: @zaliqarosli I added the change to remove jsx from state.

maltheism:
@kongtiaowang could you help with the travis case: CreateTimepointTestIntegrationTest::testCreateTimepointSuccessLink
I think the problem is because there are now more divs than before.

maltheism:
See #4415 because I messed up the rebase on this PR.

PR #4362, opened 2019-02-26
[Login] Fix issue where password reset emails were not sent

### LINKED ISSUE #4360: Password reset via email is impossible in LORIS 20.x
## Summary
As reported in the loris-dev email list, password reset is not possible in this version. This is because when a user is not logged in they are configured to be a new AnonymousUser which short-circuits the ability to retrieve a user's email from the Database.

### To confirm
* Try resetting your password via email. You will not receive an email.
* The error log will report `PHP Notice:  Undefined index: Email in /var/www/loris/php/libraries/User.class.inc on line 209`

### Problematic lines

https://github.com/aces/Loris/blob/8c571758ab82a1409e8b8375e0945e7b7a2592c6/php/libraries/User.class.inc#L44-L55

https://github.com/aces/Loris/blob/8c571758ab82a1409e8b8375e0945e7b7a2592c6/modules/login/php/passwordreset.class.inc#L60-L74

johnsaigle:
### Brief summary of changes

This PR changes the module to use the `factory` function instead of the `singleton` function from the `User` class. Singleton is having issues as it returns an AnonymousUser instance when no logged-in session is active. This prevents the retrieval of the Email value for users attempting to reset their passwords.

This fix should solve the issue but the logic of this functionality should be reworked down the line.

### This resolves issue...

Resolves #4360 

### To test this change...

* Trigger a password reset email by using the "Forgot Password?" link on the front-page of LORIS. Make sure you can reset your password.

### DIFF
--- a/modules/login/php/passwordreset.class.inc
+++ b/modules/login/php/passwordreset.class.inc
@@ -65,7 +65,7 @@ class PasswordReset extends \NDB_Form
 
         // create the user object
         $username = $values['username'];
-        $user     =& \User::singleton($username);
+        $user     = \User::factory($username);
 
         $email = $user->getData('Email');
         // check that it is a valid user


driusan:
@kongtiaowang can you test this?

PR #4364, opened 2019-02-26
[new_profile] Prevent duplication when button is clicked repetitively

ridz1208:
### Brief summary of changes
Added safeguard for disabling button after first click while the request is being processed in the backend. this was copied from the `next_stage` change done a while back (see related PR)

https://github.com/aces/Loris/pull/2358 

### This resolves issue...
Issue reported by collaborators at the CRU

### To test this change...

- [ ] before checking out the branch, try creating a candidate, fill out all the information on the form and click multiple times consecutively on the submission button. `n` new candidates will be created
- [ ] after checking out this branch. as soon as you click the first time the button should be disabled and no additional candidates created

### DIFF
--- a/modules/new_profile/templates/form_new_profile.tpl
+++ b/modules/new_profile/templates/form_new_profile.tpl
@@ -69,7 +69,7 @@
     {/if}
 
 	<div class="form-group col-sm-12">
-		<div class="col-sm-12"><input class="btn btn-primary col-sm-offset-2 col-sm-2" name="fire_away" value="Create" type="submit" /></div>
+		<div class="col-sm-12"><input class="btn btn-primary col-sm-offset-2 col-sm-2" name="fire_away" value="Create" type="submit" onClick="this.form.submit(); this.disabled=true;"/></div>
 	</div>
 </table>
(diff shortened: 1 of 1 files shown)


ridz1208:
@zaliqarosli you seemed so interested, could you give it a try when you get a chance ?

johnsaigle:
Is this only an issue for tpl forms? Is it worth trying to make something generic (or at least adding this to other buttons also)?

ridz1208:
@johnsaigle maybe worth just making this change on tpl forms until we move them to react

ridz1208:
@driusan MERGE MEEEE

driusan:
Seems hacky, but okay, if it works and it's just a stop gap until things are converted to React.

thomasbeaudry:
> Seems hacky, but okay, if it works and it's just a stop gap until things are converted to React.

I'm surprised you think this is hacky, it's a pretty common practice for this issue.

driusan:
The non-hacky way to do it is a CSRF token (ie. view source on this page, the form has a hidden "authenticity_token", not javascript to disable the form.)

PR #4366, opened 2019-02-27
[Core - UserPermissions] hasAnyPermission does not include "superuser" 

kongtiaowang:
Add superuser into the hasAnyPermission function.

How to test.
Only give a user "superuser" permission and load the examiner module.

### DIFF
--- a/php/libraries/UserPermissions.class.inc
+++ b/php/libraries/UserPermissions.class.inc
@@ -169,7 +169,9 @@ class UserPermissions
             );
         }
         foreach ($permissions as $p) {
-            if ($this->_permissionTrueAndValid($p)) {
+            if ($this->_permissionTrueAndValid($p)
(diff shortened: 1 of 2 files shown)


php/libraries/UserPermissions.class.inc
```diff
@@ -169,7 +169,9 @@ class UserPermissions
             );
         }
         foreach ($permissions as $p) {
-            if ($this->_permissionTrueAndValid($p)) {
+            if ($this->_permissionTrueAndValid($p)
```
driusan: Why not just change this to hasPermission where it's already handled? I can't imagine any situation where you'd want hasAnyPermission and hasPermission to have different boundary permissions that would justify bypassing hasPermission.

kongtiaowang: nice

kongtiaowang:
@driusan Could u review it?

PR #4372, opened 2019-02-28
[Login: Request Account] Fixing 500 Error that arises due to email discrepancy 

jesscall:
### Brief summary of changes
Hey y'all! 

Charlie and I came across a weird use case that results in a 500 error on the request account page. 
Normally, if the email provided is not unique, a success message is still supposed to be displayed for security reasons. 
However, the current code only checks if the email exists under the 'Email' column in users table. In the case where the UserID and Email are different (as in, 2 different email addresses), this leads to a database duplication error. This PR addresses that problem by checking whether the email provided exists in either of the columns mentioned. 

### To test this change...

- [ ] Duplicate the error by heading to the 'User Accounts' page and changing the email associated with your account. If you are reluctant to do so, you can also run this query and find accounts that already have the email discrepancy.
`SELECT Email, UserID FROM users WHERE Email <> UserID;`
- [ ] First test when not on this branch. Request a new account using the email from column `UserID`. You should get a HTTP 500 error.
- [ ] Now checkout this branch and try requesting account again. You should be redirected to "thank you" message.

### DIFF
--- a/modules/login/php/requestaccount.class.inc
+++ b/modules/login/php/requestaccount.class.inc
@@ -160,7 +160,7 @@ class RequestAccount extends \NDB_Form
 
         // check email address' uniqueness
         $result = $DB->pselectOne(
-            "SELECT COUNT(*) FROM users WHERE Email = :VEmail",
+            "SELECT COUNT(*) FROM users WHERE Email OR UserID=:VEmail",
             array('VEmail' => $from)
         );
 
--- a/modules/login/php/requestaccount.class.inc
+++ b/modules/login/php/requestaccount.class.inc
@@ -160,7 +160,7 @@ class RequestAccount extends \NDB_Form
 
         // check email address' uniqueness
         $result = $DB->pselectOne(
-            "SELECT COUNT(*) FROM users WHERE Email OR UserID=:VEmail",
+            "SELECT COUNT(*) FROM users WHERE Email:VEmail OR UserID=:VEmail",
             array('VEmail' => $from)
         );
 
--- a/modules/login/php/requestaccount.class.inc
+++ b/modules/login/php/requestaccount.class.inc
@@ -160,7 +160,7 @@ class RequestAccount extends \NDB_Form
 
         // check email address' uniqueness
         $result = $DB->pselectOne(
-            "SELECT COUNT(*) FROM users WHERE Email:VEmail OR UserID=:VEmail",
+            "SELECT COUNT(*) FROM users WHERE Email=:VEmail OR UserID=:VEmail",
             array('VEmail' => $from)
         );
 


ridz1208:
@driusan we really need to modernize that whole logic !!

modules/login/php/requestaccount.class.inc
```diff
@@ -160,7 +160,7 @@ class RequestAccount extends \NDB_Form
 
         // check email address' uniqueness
         $result = $DB->pselectOne(
-            "SELECT COUNT(*) FROM users WHERE Email = :VEmail",
+            "SELECT COUNT(*) FROM users WHERE Email OR UserID=:VEmail",
```
driusan: I don't think this is valid SQL.

maltheism:
@johnsaigle @ridz1208  does the "UserID" change if the user updates their email to something else? If the UserID changes as well, then this PR should be good. Otherwise this logic will break something.

ridz1208:
@maltheism no it does not can you explain your logic ?

maltheism:
@ridz1208 I'm thinking of an edge case where someone changes their email to something else and the UserID column won't update to the new email changed. Thus resulting in a potential error if a new account is made again but with the previous email that matches the other account UserID.

Also I'm now under the impression that the UserID column may be unnecessary in LORIS because it's kind of useless. Since the Email column could be the substitution being its unique and only if no other account is allowed to share the same email address.

What are your thoughts?

johnsaigle:
@maltheism I believe `UserID` is a username. No reason why a username can't be an email address if that's what people want. I don't believe we do any validation to prevent that username != email. But it's needed as a field to store a short name/display name in LORIS.

johnsaigle:
I'm thinking this through a little more and here is my understanding of how this is supposed to work.
* When an account is requested, a user is asked to provide their Email (along with some other fields)
* By default, a `UserID` is set to be equal to the `Email` value. `UserID` is a username that is displayed on the front-end

I think there's an issue with this PR (or at least its premises) in assuming that these values must always be equal. They are optionally equal, and by default it seems like they are, but if my account email is `john@loris.ca` my username can be simply `john` or anything else. If you assume they're equal then that prevents people from changing their associated mailing address or username without changing both. 

It seems like this is the example that happened here: someone must have changes their email address within LORIS but is probably trying to login with their username, which is in turn set to their old email address. Otherwise I don't understand the case where someone would request an account with an email address that is already present, either in UserID or Email, in the database.

It's a bit tricky to solve this because our users are apparently trained to use their emails as usernames. This is probably why [this PR](https://github.com/aces/Loris/pull/3042) came up and why the LORIS UI uses Email and Username interchangeably.

maltheism:
Hey @johnsaigle your second comment is what I was trying express but also with the rare possible edge case. I think the standard of convention (not in LORIS but elsewhere) is not having the UserID as an email address if the database has a column for email. Otherwise the UserID is typically an email address that's used for the user logging in.

ridz1208:
@johnsaigle @maltheism 
I think you both have valid points here but let's not forget that this is a bug that is live on production and this is a bugfix designed to avoid a 500 error and not a revamp of the logic.

Consider this scenario (because we all for it) User X has the following:
`UserID = A@email.com`
`Email = B@email.com`

Scenario 1: Now X forgot that he has a loris account already and goes in to the request account and requests a new account using `B@email.com`. X gets a "success" message for security reasons but no account is created because that email address already exists.

Scenario2: (most likely to happen) X goes in and tries to request an account using `A@email.com`. X currently gets a 500 because the code checks for an email duplication only when we allow emails in both `userid` and `email`...

I can only see 3 solutions here:
 1. when a user requests an account, do not prepopulate the `UserID` column, leave it empty and only fill in when the request is being approved.
 2. (this PR) prevent the creation of an account if the email is either a duplicate of an existing email OR a duplicate of a userID
3,. ask for a userID when requesting an account.

In all 3 cases above we will always display a success message either way on the request account and this is exactly what this PR is doing anyways

in cases 1 and 3 a new account would be created (which should not be because that user should not be creating a new account everytime they forget they already have one)

in case 2 the user will not be given a new account and the user will probably reach out to the loris team who will remind them that they indeed have an account already that they probably used 100 000 times already...

johnsaigle:
@ridz1208 Fair enough. This approach (2) works for now and I also think an ticket for approach (1) should be created.

maltheism:
@ridz1208 @johnsaigle I think this PR should be merged because it fixes the problem on live production.

driusan:
I'm not very comfortable with this approach given the above discussion, but I'll merge it if the latest change has been manually tested.

maltheism:
@driusan I've manually tested the change and it satisfies the intent of checking if the UserID column contains the email address being used when the user is requesting an account. I'm uneasy about this PR because I think the logic of the UserID being assigned as an email to be bad design. Specifically when the user can change their email to be something else and the UserID doesn't update and will always have an old email address.

PR #4373, opened 2019-02-28
[ImagingUploader] Revert file upload element name

zaliqarosli:
### Brief summary of changes

Somehow the name got changed and now there's a bug. I'm changing back the name to match the prop key

To test:
- on major, in 'Imaging Uploader' module, under 'Upload' tab, try to add a file to upload. the file element field stays empty and you can't click submit.
- on this branch, there are no issues

### DIFF
--- a/modules/imaging_uploader/jsx/UploadForm.js
+++ b/modules/imaging_uploader/jsx/UploadForm.js
@@ -385,7 +385,7 @@ class UploadForm extends Component {
               value={this.state.formData.visitLabel}
             />
             <FileElement
-              name='mri_file'
+              name='mriFile'
               label='File to Upload'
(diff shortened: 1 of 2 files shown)


maltheism:
Hi @zaliqarosli I searched LORIS to find one other place existing in the codebase using the "mri_file" name. I think it's unrelated but maybe it should be changed as well? Inside the modules/imaging_uploader/php/imaging_uploader.class.inc at line # 389. Where the array_push of the $errors['mri_file']

zaliqarosli:
@maltheism that's unrelated and doesn't break anything. the issue isn't how the variables are named but more that it was defined as one thing in React, but then used as another. the index of the errors array can be called mri_file no problem

PR #4374, opened 2019-02-28
[Database] Add NOQUOTES as an HTML escaping option for JSON

### LINKED ISSUE #4371: Create function to safely insert JSON objects to Database
In order to work with JSON objects in the DB, the `unsafe*` family of functions within the Database class must be used as all other functions go through `_HTMLEscapeArray`. This causes corruptions in the data as the function converts the double-quotes necessary to the integrity of the JSON structure.

A new family of CRUD functions should be created to deal with this use case. They should follow the structure of the other functions, including auto escaping, but instead of using `htmlspecialchars()` (as is done in `_HTMLEscapeArray` they should use `html_entities` with the `ENT_NOQUOTES` flag (see [docs](https://secure.php.net/manual/en/function.htmlentities.php)). This should restore some of the security functionality provided by existing functions while also not causing problems for JSON data.

johnsaigle:
### Brief summary of changes
Modify the unsafe* family of functions to allow for a NOQUOTES option
when inserting data. This provides a middle ground between escaping all
HTML characters and none. This is important for working with JSON in the
database as escaping the quotation marks seems to make the data
difficult to work with.

Data that is stored in this way will need to be decoded using the html_entity_decode() function when it is retrieved.

### This resolves issue...

- Resolves #4371 

### To test this change...

- [ ] Ensure that LORIS doesn't break when you check out this PR.
- [ ] When working with JSON, try using the unsafe* functions with a value of `1` as the 3rd parameter. Make sure that your data comes out fine (after decoding with html_entity_decode()).

### DIFF
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -2,7 +2,7 @@
 /**
  * This file represents an SQL database abstraction layer for use in Loris
  *
- * PHP Version 5
+ * PHP Version 7
  *
  * @category Main
  * @package  Loris
@@ -11,6 +11,9 @@
  * @link     https://www.github.com/aces/Loris-Trunk/
  */
 define("DEBUG", false);
+define('ESCAPE_ALL', 0);
+define('ESCAPE_NOQUOTES', 1);
+define('ESCAPE_NONE', 2);
 
 /**
  * Database abstraction layer presents a unified interface to database
@@ -28,6 +31,16 @@ define("DEBUG", false);
  */
 class Database
 {
+    /**
+     * Constants determining whether HTML characters will be escaped before
+     * being inserted into the Database.
+     *
+     * @access private
+     */
+    const ESCAPE_ALL      = 0;
+    const ESCAPE_NOQUOTES = 1;
+    const ESCAPE_NONE     = 2;
+
     /**
      * The database handle, created by the connect() method
      *
@@ -260,7 +273,7 @@ class Database
      */
     public function insert(string $table, array $set): void
     {
-        $this->_realinsert($table, $set, true);
+        $this->_realinsert($table, $set, ESCAPE_ALL);
     }
 
     /**
@@ -270,14 +283,20 @@ class Database
      * automatically escaped. This should only be called when we know the source of
      * the input is trustworthy and must contain HTML.
      *
-     * @param string $table the table into which to insert the row
-     * @param array  $set   the values with which to fill the new row
+     * @param string $table      the table into which to insert the row
+     * @param array  $set        the values with which to fill the new row
+     * @param int    $escapeMode Which HTML characters to escape. A value of 1
+     *                           will escape all characters but quotation marks
+     *                           and a value of 2 will not escape any characters.
      *
      * @return void
      */
-    public function unsafeinsert(string $table, array $set): void
-    {
-        $this->_realinsert($table, $set, false);
+    public function unsafeinsert(
+        string $table,
+        array $set,
+        int $escapeMode = ESCAPE_NONE
+    ): void {
+        $this->_realinsert($table, $set, $escapeMode);
     }
 
     /**
@@ -294,7 +313,7 @@ class Database
     public function insertIgnore(string $table, array $set): void
     {
         $ignore = true;
-        $this->_realinsert($table, $set, true, $ignore);
+        $this->_realinsert($table, $set, ESCAPE_ALL, $ignore);
     }
 
     /**
@@ -311,7 +330,7 @@ class Database
      */
     public function insertOnDuplicateUpdate($table, $set)
     {
-        return $this->_realinsert($table, $set, true, false, true);
+        return $this->_realinsert($table, $set, ESCAPE_ALL, false, true);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


johnsaigle:
@HenriRabalais Would you be able to test this using the biobank module? I don't have anywhere in the codebase where I'm able to work with JSON data.

ridz1208:
@johnsaigle it could also be tested with the instruments saving JSON data in the flag.data column

php/libraries/Database.class.inc
```diff
@@ -270,14 +283,20 @@ class Database
      * automatically escaped. This should only be called when we know the source of
      * the input is trustworthy and must contain HTML.
      *
-     * @param string $table the table into which to insert the row
-     * @param array  $set   the values with which to fill the new row
+     * @param string $table      the table into which to insert the row
+     * @param array  $set        the values with which to fill the new row
+     * @param int    $escapeMode Which HTML characters to escape. A value of 1
+     *                           will escape all characters but quotation marks
+     *                           and a value of 2 will not escape any characters.
```
PapillonMcGill: should have ending dot on every param def or none for consistency

php/libraries/Database.class.inc
```diff
@@ -322,14 +341,20 @@ class Database
      * value for any unique key, the row where the duplicate is present will be
      * updated. HTML from any field in the row will *not* be automatically escaped.
      *
-     * @param string $table the table into which to insert the row
-     * @param array  $set   the values with which to fill the row
+     * @param string $table      the table into which to insert the row
+     * @param array  $set        the values with which to fill the row
+     * @param int    $escapeMode Which HTML characters to escape. A value of 1
+     *                           will escape all characters but quotation marks
+     *                           and a value of 2 will not escape any characters.
```
PapillonMcGill: same as before for ending dots

php/libraries/Database.class.inc
```diff
@@ -342,13 +367,18 @@ class Database
      *
      * This function alone does not guarantee safety.
      *
-     * @param array $set The array to be inserted as a new row
+     * @param array $set  The array to be inserted as a new row
```
```suggestion
     * @param array $set  The array to be inserted as a new row.
```

php/libraries/Database.class.inc
```diff
@@ -504,46 +545,47 @@ class Database
...
      *
-     * @param string $table   the table into which to insert the row
-     * @param array  $set     the values with which to fill the new row
-     * @param array  $i_where the selection filter, joined as a boolean and
+     * @param string $table      the table into which to insert the row
+     * @param array  $set        the values with which to fill the new row
+     * @param array  $i_where    the selection filter, joined as a boolean and
+     * @param int    $escapeMode Which HTML characters to escape. A value of 1
+     *                           will escape all characters but quotation marks
+     *                           and a value of 2 will not escape any characters.
```
PapillonMcGill: same as before for ending dots.

php/libraries/Database.class.inc
```diff
@@ -28,6 +31,16 @@ define("DEBUG", false);
...
 {
+    /**
+     * Constants determining whether HTML characters will be escaped before
+     * being inserted into the Database.
+     *
+     * @access private
+     */
+    const ESCAPE_ALL      = 0;
+    const ESCAPE_NOQUOTES = 1;
+    const ESCAPE_NONE     = 2;
```
PapillonMcGill: Are these const needed here?
all are already DEFINEd and no use of $this->CONST is made.

php/libraries/Database.class.inc
```diff
@@ -11,6 +11,9 @@
  * @link     https://www.github.com/aces/Loris-Trunk/
  */
 define("DEBUG", false);
+define('ESCAPE_ALL', 0);
+define('ESCAPE_NOQUOTES', 1);
+define('ESCAPE_NONE', 2);
```
driusan: I don't see why you include both define's nd consts below. The const should be enough.

php/libraries/Database.class.inc
```diff
@@ -28,6 +31,16 @@ define("DEBUG", false);
  */
 class Database
 {
+    /**
+     * Constants determining whether HTML characters will be escaped before
+     * being inserted into the Database.
+     *
+     * @access private
```
driusan: These are not private. They're used as arguments to public methods.

php/libraries/Database.class.inc
johnsaigle: Yep, they are left over from a previous approach. Will delete.

php/libraries/Database.class.inc
johnsaigle: You're right, this is left over from a previous commit.

johnsaigle:
@ridz1208 I don't think I have instrument data to test this. Would you be able to test for me or provide some data/instructions I can use?

php/libraries/Database.class.inc
PapillonMcGill: and capital `T`

johnsaigle:
Blocked by #4387

ridz1208:
@johnsaigle for the manual testing could you come and see me ? i have a few questions

php/libraries/Database.class.inc
```diff
@@ -2,15 +2,18 @@
...
  * @package  Loris
  * @author   Alex Zijdenbos <zijdenbos@example.com>
  * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
  * @link     https://www.github.com/aces/Loris-Trunk/
  */
-define("DEBUG", false);
+define('DEBUG', false);
+define('ESCAPE_ALL', 0);
+define('ESCAPE_NOQUOTES', 1);
+define('ESCAPE_NONE', 2);
```
driusan: These should all be class constants.

johnsaigle:
@driusan I think @ridz1208 found that this code has an issue where some data was double-escaped and so it might not be working well. I didn't have data to test on so I can't confirm that this issue does or doesn't exist at this point, but I trust his testing.

This merge should probably be reverted. I'll add `Needs Work`. It also shouldn't be merged until it has `Passed Manual Tests`

driusan:
@johnsaigle I knew I should have requested that unit tests be added... I sent a PR to revert the merge, can you review it?

PR #4375, opened 2019-03-01
[user_accounts] my_preferences fix when ['perm'] is undefined

maltheism:
### Brief summary of changes

The user_accounts module has the $services['perm'] inside my_preferences.class.inc and where it could be undefined & polluting the logs with Notice: Undefined index: perm. I wrote an isset check for the bugfix.

### To test this change...

- [x] Visit http://(your_loris)/user_accounts/my_preferences/


maltheism:
Closing for now. I believe I didn't run the 0000-00-01-Permission.sql and encountered this error because of it.

PR #4377, opened 2019-03-01
[Instrument_builder] Build & Load were broken in react.

maltheism:
### Brief summary of changes

This PR fixes the Build & Load of instrument_builder.

### To test this change...

- [x] Pull the PR and test on your VM.

### DIFF
--- a/modules/instrument_builder/jsx/react.instrument_builder.js
+++ b/modules/instrument_builder/jsx/react.instrument_builder.js
@@ -65,35 +65,33 @@ class LoadPane extends Component {
   }
   // Render the HTML
   render() {
-    let alert = '';
+    let alert = {
+      message: '',
+      details: '',
+      display: 'none',
+    };
     // Set up declared alerts, if there is any.
     switch (this.state.alert) {
       case 'success':
-        alert = (
-          <div className='alert alert-success alert-dismissible' role='alert'>
-            <button type='button' className='close' onClick={this.resetAlert}><span aria-hidden='true'>&times;</span></button>
-            <strong>Success!</strong> Instrument Loaded
-          </div>
-        );
+        alert = {
+          message: 'Success!',
+          details: 'Instrument Loaded',
+          display: 'block',
+        };
         break;
       case 'typeError':
-        alert = (
-          <div className='alert alert-danger alert-dismissible' role='alert'>
-            <button type='button' className='close' onClick={this.resetAlert}><span aria-hidden='true'>&times;</span></button>
-            <strong>Error!</strong> Wrong file format
-          </div>
-        );
+        alert = {
+          message: 'Error!',
+          details: 'Wrong file format',
+          display: 'block',
+        };
         break;
       case 'duplicateEntry':
-        alert = (
-          <div className='alert alert-danger alert-dismissible' role='alert'>
-            <button type='button' className='close' onClick={this.resetAlert}>
-              <span aria-hidden='true'>&times;</span>
-            </button>
-            <strong>Error!</strong><br/>
-              {this.state.alertMessage}
-          </div>
-        );
+        alert = {
+          message: 'Error!',
+          details: this.state.alertMessage,
+          display: 'block',
+        };
         break;
       default:
         break;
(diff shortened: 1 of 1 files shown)


modules/instrument_builder/jsx/react.instrument_builder.js
```diff
@@ -325,7 +329,7 @@ class DisplayElements extends Component {
         );
       }
       return row;
-    });
+    }.bind(this));
```
HenriRabalais: `this` shouldn't have to be bound to the render function.

modules/instrument_builder/jsx/react.instrument_builder.js
```diff
@@ -65,43 +65,47 @@ class LoadPane extends Component {
...
+        };
         break;
       default:
         break;
     }
     return (
       <TabPane Title='Load Instrument' {...this.props}>
         <div className='col-sm-6 col-xs-12'>
-          {alert}
+          <div id='load_alert' style={{display: alert.display}} className='alert alert-danger alert-dismissible' role='alert'>
```
zaliqarosli: alert-danger is set even when the alert is successful?

modules/instrument_builder/jsx/react.instrument_builder.js
zaliqarosli: i agree with @HenriRabalais. i don't think this is the way to fix the bug

zaliqarosli:
need to import prop types into the react file. add `import PropTypes from 'prop-types'`

modules/instrument_builder/jsx/react.instrument_builder.js
maltheism: Nice catch @zaliqarosli 🙂I added a fix so it will show alert-success when success.

modules/instrument_builder/jsx/react.instrument_builder.js
maltheism: You and @HenriRabalais are correct! 
I also found a blog post about it: https://ryanfunduk.com/articles/never-bind-in-render/
New changes remove it from the render function.

maltheism: See the comment from @zaliqarosli and I removed it from the render function.

maltheism:
I updated the code with change requests from @zaliqarosli and @HenriRabalais. Not trying to rewrite the module but just get it back to the state of functioning again because BioBank is going to be using Major.

modules/instrument_builder/jsx/react.instrument_builder.js
```diff
@@ -348,6 +368,11 @@ class DisplayElements extends Component {
     );
   }
 }
+DisplayElements.propTypes = {
+  draggable: PropTypes.bool,
```
zaliqarosli: i think you're missing this.props.elements and this.props.updateElement

modules/instrument_builder/jsx/react.instrument_builder.js
```diff
@@ -325,13 +340,18 @@ class DisplayElements extends Component {
         );
       }
       return row;
-    });
+    }.bind(this));
```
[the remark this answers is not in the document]
maltheism: @zaliqarosli I think it's necessary to have the bind there. The reason being the state is accessed inside a function inside the tableRows() function.

modules/instrument_builder/jsx/react.instrument_builder.js
```diff
@@ -65,43 +66,50 @@ class LoadPane extends Component {
...
-            </button>
-            <strong>Error!</strong><br/>
-              {this.state.alertMessage}
-          </div>
-        );
+        alert = {
+          message: 'Error!',
+          details: this.state.alertMessage,
+          display: 'block',
+          class: 'alert alert-success alert-dismissible',
```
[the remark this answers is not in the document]
maltheism: Oops, thanks for this find lol.

modules/instrument_builder/jsx/react.instrument_builder.js
```diff
@@ -65,43 +66,50 @@ class LoadPane extends Component {
...
-          <div className='alert alert-success alert-dismissible' role='alert'>
-            <button type='button' className='close' onClick={this.resetAlert}><span aria-hidden='true'>&times;</span></button>
-            <strong>Success!</strong> Instrument Loaded
-          </div>
-        );
+        alert = {
+          message: 'Success!',
+          details: 'Instrument Loaded',
+          display: 'block',
+          class: 'alert alert-danger alert-dismissible',
```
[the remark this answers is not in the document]
maltheism: Oops, thanks for this find lol.

PR #4378, opened 2019-03-04
[Test] unittest for AnonymousUser.class.inc

kongtiaowang:
This is the easiest test case demo for a class without dependency. (For summer student)
https://github.com/aces/Loris/pull/4383
Todo: 
Breadcrumb.class.inc
BreadcrumbTrail.class.inc
Identifier.class.inc
LorisForm.class.inc

### DIFF
--- a/test/unittests/AnonymousUserTest.php
+++ b/test/unittests/AnonymousUserTest.php
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
 /**
  * AnonymousUser class tests
  *
--- a/test/unittests/AnonymousUserTest.php
+++ b/test/unittests/AnonymousUserTest.php
@@ -1,0 +1,57 @@
+<?php
+/**
+ * AnonymousUser class tests
+ *
+ * PHP Version 7
+ *
+ * @category Tests
+ * @package  Test
+ * @author   Shen Wang <wangshen.mcin@gmail.com>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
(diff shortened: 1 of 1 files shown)


test/unittests/AnonymousUserTest.php
```diff
@@ -0,0 +1,57 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

test/unittests/AnonymousUserTest.php
```diff
@@ -0,0 +1,57 @@
...
+        $this->anonymousUser = new AnonymousUser();
+
+    }
+
+    /**
+     * Tests getSiteNames function,it will return an empty array.
+     *
+     * @return void
+     */
+    public function testGetSiteNames()
```
johnsaigle:
```suggestion
    public function testGetSiteNames(): void
```

test/unittests/AnonymousUserTest.php
```diff
@@ -0,0 +1,57 @@
...
+    protected function setUp()
+    {
+        parent::setUp();
+
+        $this->anonymousUser = new AnonymousUser();
+
+    }
+
+    /**
+     * Tests getSiteNames function,it will return an empty array.
```
johnsaigle:
```suggestion
     * Tests getSiteNames function which should return an empty array.
```

johnsaigle: The way this was written before makes it sound like _this_ function will return an empty array.

PR #4379, opened 2019-03-05
add ignore files

zaliqarosli:
### Brief summary of changes
add files to git ignore so compiled js files don't show up when trying to change branches

### DIFF
--- a/modules/conflict_resolver/js/.gitignore
+++ b/modules/conflict_resolver/js/.gitignore
@@ -1,0 +1,2 @@
+resolved_conflicts_columnFormatter.js
+unresolved_columnFormatter.js
--- a/modules/conflict_resolver/js/.gitignore
+++ b/modules/conflict_resolver/js/.gitignore
@@ -1,2 +1,1 @@
-resolved_conflicts_columnFormatter.js
-unresolved_columnFormatter.js
+*.js
--- a/modules/dicom_archive/js/.gitignore
+++ b/modules/dicom_archive/js/.gitignore
@@ -1,0 +1,1 @@
+dicom_archive.js
--- a/modules/dicom_archive/js/.gitignore
+++ b/modules/dicom_archive/js/.gitignore
@@ -1,1 +1,1 @@
-dicom_archive.js
+*.js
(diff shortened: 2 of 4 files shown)


zaliqarosli:
@ridz1208

maltheism:
Hi @zaliqarosli, I always see the below in my `git status` after running `npm install`:
```
modules/conflict_resolver/js/
modules/data_integrity_flag/js/
modules/dicom_archive/js/
modules/examiner/js/
modules/imaging_browser/js/imagingBrowserIndex.js
modules/instrument_manager/js/
modules/issue_tracker/js/
modules/media/js/
modules/quality_control/js/
modules/server_processes_manager/js/
modules/user_accounts/js/userAccountsIndex.js
```
Is there a way to add these to ignore as well.

zaliqarosli:
@maltheism hey, are they tracked or untracked? the ones I added in this PR are tracked which makes it anyone for when you want to switch branches. but i think they ones you provided above are untracked right? so they should be somewhat "ignored" for branch switches

maltheism:
@zaliqarosli They're untracked but I wish they were not there because I cannot use `git add *` instead of having to manually type everything to git add.

zaliqarosli:
@maltheism i think once everything is reactified, and we remove pure js, we can gitignore `modules/*/js/*.js`. but for now, i think the only way is to add each directory to the main .gitignore file, which might be too specific for a base file? not too sure. i wonder what @ridz1208 and @driusan think

modules/conflict_resolver/js/.gitignore
```diff
@@ -0,0 +1,2 @@
+resolved_conflicts_columnFormatter.js
+unresolved_columnFormatter.js
```
driusan: Shouldn't all of these be `*.js`?

driusan:
@zaliqarosli There's a few modules which have (tracked, non-jsx) javascript directly in the js folder, so that pattern doesn't work for now.

zaliqarosli:
@driusan i'm confused. so i should revert this back to the first commit?

maltheism:
@zaliqarosli, I think your pattern works for the modules you're using it for but not for the other modules that have js files (that are not compiled from weback into the js directory).

driusan:
no, I think it looks good now.. I was referring to your suggestion of adding a pattern for `modules/*/js/*.js`

PR #4386, opened 2019-03-08
[src/Middleware] AnonymousPageDecorationMiddleware.php Post response bugfix

maltheism:
### Brief summary of changes

Yesterday I was having issues with attempting to receive a JSON object from the server (after a post request from the client). The client was receiving data from the server but html content was included as well. Making my JSON object not be a JSON object and the javascript error would fail inside the fetch function.

One of my PRs is working on making the login form react. Well it turns out this bug only happens when the user isn't logged in. This bugfix is needed for when the user is not logged in and for the server to respond to a post request with data that's correctly formatted (such as a JSON object).

Note: I made a PR for both major & bugfix branch because I'm in the processing of making the login react on major.

### This resolves issue...

- [ ] Redmine? #

- [ ] Github? #

### To test this change...

- [x] Trust me it fixes the problem above. Otherwise make a post request without being logged in and attempt to receive a json response from the server. You would be debugging for days without this fix!

### DIFF
--- a/src/Middleware/AnonymousPageDecorationMiddleware.php
+++ b/src/Middleware/AnonymousPageDecorationMiddleware.php
@@ -52,7 +52,18 @@ class AnonymousPageDecorationMiddleware implements MiddlewareInterface {
             );
         }
 
+        // Handle needs to be called before form action, because handle potentially
+        // calls setup which modifies the $page->FormAction value (ie in the imaging
+        // browser)
         $undecorated = $handler->handle($request);
+        $contenttype = $undecorated->getHeaderLine("Content-Type");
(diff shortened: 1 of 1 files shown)


maltheism:
See #4385

driusan:
I thought when I had tried to do this myself a long time ago (because you're right, it should only be for text/html content-types) it was breaking a lot of pages in LORIS.. but maybe I forgot the "" check? Can someone test LORIS on this branch to verify that it's not breaking other things?

cmadjar:
@maltheism is this ready for review?

cmadjar:
Oops. I guess not since it needs to be rebased on major. Sorry

maltheism:
Closing for new PR on major #5237

PR #4387, opened 2019-03-11
[Meta] Fix Travis failure on major

johnsaigle:
### Brief summary of changes

#4273 was just merged and flags an issue in the Installer.class.inc. Since this was merged to major, every major build will fail until this PR goes through. Not sure why it wasn't caught in the original PR.

Discovered in PR #4374 which fails with this message:
> php/installer/Installer.class.inc:127 PhanUndeclaredVariable Variable $PHP_RELEASE_VERSION is undeclared

### DIFF
--- a/php/installer/Installer.class.inc
+++ b/php/installer/Installer.class.inc
@@ -123,8 +123,8 @@ class Installer
     {
         // Check PHP version is supported.
         if (PHP_MAJOR_VERSION < 7) {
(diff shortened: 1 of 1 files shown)


driusan:
@johnsaigle this failed Travis

driusan:
nm, might've just been a transient failure with one of the builds. I'll restart and see what happens..

PR #4389, opened 2019-03-12
[Meta] Prevent verbose "No syntax errors detected in..." from appearing in test output

johnsaigle:
### Brief summary of changes

Make Travis/linting less verbose by hiding messages telling us everything is fine. Only STDOUT is affected so error messages will still appear.

#### To test
* Deliberately make a syntax error in a php file in LORIS. Run `make checkstatic` and it should fail and flag the syntax error
* Do the same on my branch. The same should happen but without output from `php -l`


driusan:
I don't see much point in this, but okay.

PR #3006, opened 2017-08-01
[Core] Added module override support to webpack

ridz1208:
This pull request does what the title says it does.

To test:
- create the project config file (see below for example)
- add specific jsx to compile in the same manner they are added in the loris webpack config
- run `npm run compile` make sure both the loris and project files are compiled properly

PR #4391, opened 2019-03-12
[webpack] simplification of the project webpack file

ridz1208:
### Brief summary of changes
This is necessary to keep supporting projects adding their own modules to webpack from the `webpack-project.config.js`

this functionality was added in https://github.com/aces/Loris/pull/3006

Everything was simplified here the only requirement is to create a `webpack-project.config.js` with the following content

```
const entry = {
    './project/modules/candidate_parameters/js/index.js': './project/modules/candidate_parameters/jsx/index.js',
    './project/modules/media/js/mediaIndex.js': './project/modules/media/jsx/mediaIndex.js',
};

module.exports = entry;
```

This will simply concatenate the lines to the `webpack.config.js` loris file and all the jsx files will be then transpiled.

Also, added the `project/*` to the ignore list for eslint since project files do not need to be linted.

### Testing
- simply override a module with JSX files
- make changes in the overridden JSX
- without this PR, the changes would never take effect because the JSX files in the overridden module are never compiled
- with this PR and the properly configured  `webpack-project.config.js` file. the changes will be transpiled into JS

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -126,7 +126,7 @@ const config = [{
 // Support project overrides
 if (fs.existsSync('./project/webpack-project.config.js')) {
   const projConfig = require('./project/webpack-project.config.js');
-  config.push(projConfig);
+  config[0].entry = Object.assign(config[0].entry, projConfig);
 }
 
 module.exports = config;
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,7 @@
 node_modules/*
 modules/*/js/*.js
 htdocs/js/components/*
+project/*
 
 # Ignore until ESLint is run
 modules/dataquery/
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,7 +1,6 @@
 node_modules/*
 modules/*/js/*.js
 htdocs/js/components/*
-project/*
 
 # Ignore until ESLint is run
 modules/dataquery/


webpack.config.js
```diff
@@ -126,7 +126,7 @@ const config = [{
 // Support project overrides
 if (fs.existsSync('./project/webpack-project.config.js')) {
   const projConfig = require('./project/webpack-project.config.js');
-  config.push(projConfig);
+  config[0].entry = Object.assign(config[0].entry, projConfig);
```
maltheism: This line of code is slick compared to the previous.

driusan:
I don't see why project javascript files shouldn't be linted. Good code practice is good code practice.

ridz1208:
@driusan we dont enforce phpcs, it's on a private repo, collaboration is usually limited to 1 or 2 people.

yes good code practices are recommended but that's not in loris' control in a way

driusan:
@ridz1208 if you really want to ignore eslint in the project directory, why can't you just add a .eslintignore in your project directory?

ridz1208:
@driusan done

johnsaigle:
Failing Travis dude to 

See #4394.

ridz1208:
@johnsaigle thx

ridz1208:
@maltheism @driusan  please re-review

PR #4392, opened 2019-03-13
[Cleanup] Remove many magic numbers from the codebase

johnsaigle:
### Brief summary of changes

Replaced [magic numbers](https://en.wikipedia.org/wiki/Magic_number_%28programming%29#Unnamed_numerical_constants) in LORIS with documented constants where it was
easy to do so. These were flagged by the tool PHP Magic Number Detector.

Also updated some documentation and formatting.

### DIFF
--- a/src/Api/Endpoints/Login.php
+++ b/src/Api/Endpoints/Login.php
@@ -198,7 +198,7 @@ class Login extends Endpoint
     {
         // Note: this code adapted from User::isPasswordStrong
         $CharTypes = 0;
-        if (strlen($key) < MIN_JWT_KEY_LENGTH) {
+        if (strlen($key) < self::MIN_JWT_KEY_LENGTH) {
             return false;
         }
         // nothing but letters
@@ -213,7 +213,7 @@ class Login extends Endpoint
         $CharTypes += preg_match('/[0-9]+/', $key);
         $CharTypes += preg_match('/[A-Za-z]+/', $key);
         $CharTypes += preg_match('/[!\\\$\^@#%&\*\(\)]+/', $key);
-        if ($CharTypes < MIN_NUM_CHARSETS) {
+        if ($CharTypes < self::MIN_NUM_CHARSETS) {
             return false;
         }
 
--- a/src/Api/Endpoints/Login.php
+++ b/src/Api/Endpoints/Login.php
@@ -7,7 +7,7 @@
  * @category API
  * @package  Loris
  * @author   Dave MacFarlane <dave.macfarlane@mcin.ca>
- * @license  Loris license
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
  * @link     https://github.com/aces/Loris
  */
 namespace LORIS\Api\Endpoints;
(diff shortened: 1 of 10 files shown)


src/Api/Endpoints/Login.php
```diff
@@ -22,11 +22,21 @@
  * @category API
  * @package  Loris
  * @author   Dave MacFarlane <dave.macfarlane@mcin.ca>
- * @license  Loris license
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
  * @link     https://github.com/aces/Loris
  */
 class Login extends Endpoint
 {
+    /* @var int A supplied JWT key must have at least this length or it will
```
driusan: this is missing a newline after the first `/*` (same with below)

src/Api/Endpoints/Login.php
```diff
@@ -22,11 +22,21 @@
...
+     * be rejected for being too weak.
+     */
+    protected const MIN_JWT_KEY_LENGTH = 20;
```
driusan: missing a newline to separate the comments for the header below from the property above it

src/Api/Endpoints/Login.php
```diff
@@ -204,7 +213,7 @@ static function isKeyStrong($key)
         $CharTypes += preg_match('/[0-9]+/', $key);
         $CharTypes += preg_match('/[A-Za-z]+/', $key);
         $CharTypes += preg_match('/[!\\\$\^@#%&\*\(\)]+/', $key);
-        if ($CharTypes < 3) {
+        if ($CharTypes < self::MIN_NUM_CHARSETS) {
```
driusan: I think a name like "MIN_COMPLEXITY_CHARSETS" would be more descriptive

src/StudyEntities/Candidate/CandID.php
```diff
@@ -24,6 +24,11 @@
  */
 class CandID extends ValidatableIdentifier
 {
+    /* @var int The minimum allowed value for valid CandIDs. Origin unclear but
```
driusan: the comment formatting is incorrect here too

johnsaigle:
@driusan I think I made the changes you wanted. I altered the var annotations as you suggested but to be fair I don't think we actually have a standard for var annotations, definitely not one that's enforced by PHPCS. Maybe we should find a sniff for this.

src/Api/Endpoints/Login.php
johnsaigle: Is this addressed with the newest changes? Not sure if I'm understanding you correctly.

driusan: No.

I meant it's written

```
/**
 * blah blah blah
 */
 protected const foo;
 /*
  * blah blah blah
  */
 protected const bar;
```

Instead of:
```
/**
 * blah blah blah
 */
 protected const foo;

 /*
  * blah blah blah
  */
 protected const bar;
```

The newline separating the two is missing.

johnsaigle: Ok, thanks for clarifying

PR #4393, opened 2019-03-13
[Meta] Update phpcs to 2.9.2

### LINKED ISSUE #4099: PHPCS version is old, causing clutter with PHP 7.3
The PHP 7.3 Travis CI build is generating warnings like this:

>PHP Warning:  "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in /home/travis/build/aces/Loris/vendor/squizlabs/php_codesniffer/CodeSniffer/File.php on line 1754

Newer versions of PHPCS have patched this and it would be nice to update to not run into these warnings all the time.

johnsaigle:
### Brief summary of changes

Our PHPCS was a few versions behind so I figured it was time to update. Versions 3.0 - 3.3 are incompatible with PHP 7.3.* and there is [a bug](https://github.com/squizlabs/PHP_CodeSniffer/issues/2442) in version 3.4 that causes unresolvable errors with certain parts of our code base. As a result the best case for us right now is the last stable version of version 2.

This PR also contains coding style changes generates by phpcbf according to the new phpcs version.

#### Related

Resolves #4099

### DIFF
--- a/htdocs/router.php
+++ b/htdocs/router.php
@@ -27,7 +27,8 @@ if ($request != '/'
 if (preg_match(
     '#^([a-zA-Z_-]+)/ajax/([a-zA-Z0-9_.-/]+)$#',
     $url
-)) {
+)
+) {
     // RewriteRule
     //      ^([a-zA-Z_-]+)/ajax/([a-zA-Z0-9_.-]+)$
     //      /AjaxHelper.php?Module=$1&script=$2 [QSA]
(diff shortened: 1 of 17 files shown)


htdocs/router.php
```diff
@@ -27,7 +27,8 @@
 if (preg_match(
     '#^([a-zA-Z_-]+)/ajax/([a-zA-Z0-9_.-/]+)$#',
     $url
-)) {
+)
```
maltheism: This new coding change is ugly.

johnsaigle: I think it's a bit messy either way... the regex could be in it's own variable.

maltheism: I'm happy to be in javascript land and not requiring this change because it just seems like waste of lines.

johnsaigle: Likewise -- when working with it I feel like Javascript code is a waste of lines :D

johnsaigle: For real though I think this change indicates bad code smells/style. It does look ugly but multi-line if statements are generally pretty ugly anyway.

PR #4395, opened 2019-03-14
[login] upgraded module to use react and bugfixes for major

maltheism:
### Brief summary of changes

Upgraded module to use react. I had to fix some things in LORIS to get this all to work and there are multiple changes in some JSX files as well.

What was updated to react:
Login, password expired, request account, and reset password.

Summary:
Turns out the Content-Security policy was broken for google captcha and I made a fix to include the iframe that recaptcha uses. Major was also broken inside the php/libraries/Email.class.inc and I made some fixes. That had to do with the password reset.

Note:
This PR includes the bugfix PR #4386 and so this PR can function.

### This resolves issue...

- [ ] Redmine? [#16050](https://redmine.cbrain.mcgill.ca/issues/16050)

### To test this change...

- [x] Please test on your VM.


maltheism:
@ridz1208 I don't have write access to this repository and for resolving the conflict with module/login/php/passwordreset.class.inc

johnsaigle:
@maltheism Resolving conflicts with the GitHub front-end is extremely buggy so I don't think any of us have this permission. 

You're better off rebasing onto major and resolving this issue locally.

modules/login/php/ajaxlogin.class.inc
```diff
@@ -0,0 +1,405 @@
...
+    /**
+     * Reset password.
+     *
+     * @param array $values the user data received.
+     *
+     * @return array
+     */
+    function resetPassword(array $values) : array
+    {
+        // Always claim success to phishers.
```
johnsaigle:
```suggestion
        // Always claim success to avoid providing account info to phishers.
```
This comment always bugged me a bit

maltheism:
@johnsaigle the conflict is that I deleted the passwordreset.class.inc since it's no longer necessary.

modules/login/php/ajaxlogin.class.inc
```diff
@@ -0,0 +1,405 @@
...
+
+        // Create the user object.
+        $username = $values['username'];
+        $user     =& \User::singleton($username);
+
+        // Check if valid user.
+        $email = '';
+        try {
+            $email = $user->getData('Email');
+            if (empty($email) || empty($user)) {
```
johnsaigle: The \User::singleton() will never return a value that equates to `empty` so this should be changed.

modules/login/php/ajaxlogin.class.inc
```diff
@@ -0,0 +1,405 @@
...
+                $values['error'] = 'invalid';
```
johnsaigle: Is this displayed to the user? We should return a more descriptive error than "invalid".

modules/login/php/ajaxlogin.class.inc
```diff
@@ -0,0 +1,405 @@
...
+                return $values;
+            }
+        } catch (\Exception $exception) {
```
johnsaigle: Maybe a more specific Exception, even just \LorisException, would be better.

modules/login/php/ajaxlogin.class.inc
```diff
@@ -0,0 +1,405 @@
...
+                $msg_data['url']      = $config->getSetting('url');
+                $msg_data['realname'] = $user->getData('Real_name');
+                $msg_data['password'] = $password;
+                \Email::send($email, 'lost_password.tpl', $msg_data);
+            } else {
+                $values['error'] = 'invalid';
+                error_log(
+                    $_SERVER['REMOTE_ADDR']
+                    . '(' . $_SERVER['HTTP_X_FORWARDED_FOR'] . ')'
+                    . " attempted to reset password for $username"
```
johnsaigle: Mix of different quotation marks.

modules/login/php/ajaxlogin.class.inc
```diff
@@ -0,0 +1,405 @@
...
+                );
+            }
+        } catch(\Exception $e) {
+            error_log('Unhandled exception in reset_password.');
```
johnsaigle: This isn't the best... without including the exception message this will essentially just hide whatever exception occurred and provide no information to help debug.

modules/login/php/ajaxlogin.class.inc
```diff
@@ -0,0 +1,405 @@
...
+     *
+     * @param array $values the user data received.
+     *
+     * @return array
+     * @throws \DatabaseException
+     */
+    function requestAccount(array $values) : array
+    {
+        // Verify reCAPTCHA
+        if (isset($values['g-recaptcha-response']) && isset($reCAPTCHAPrivate)) {
```
johnsaigle: `$reCAPTCHAPrivate` doesn't appear to be defined in this scope.

modules/login/php/ajaxlogin.class.inc
maltheism: @johnsaigle It's not displayed to the user. I reused all of the php code for the login with a few changes of my own. I'm likely not doing anything when the $values['error'] is invalid because I haven't tested a case to fire that yet.

modules/login/php/ajaxlogin.class.inc
```diff
@@ -0,0 +1,405 @@
...
+                }
+            }
+        }
+        // Show success message even if email already exists for security reasons
+        $values['success'] = true;
+        return $values;
+    }
+
+    /**
+     * Login account.
```
johnsaigle: This should be more descriptive. Something like "Processes login data submitting via form, attempts authentication and verifies that a password has not expired.

modules/login/php/ajaxlogin.class.inc
maltheism: I'll change it to email couldn't be sent because I believe that's where the error would be from.

johnsaigle:
@maltheism I think what I said still applies. You should be able to sort this out locally. As a rule we don't resolve conflicts via the GitHub interface afaik.

modules/login/php/ajaxlogin.class.inc
maltheism: I'm going to make a separate PR about this but reCAPTCHAPrivate isn't used anymore by Google. Since they have their own authentication for domains with only needing a public key.

johnsaigle: So none of this code works anyway? Might as well delete it now then.

modules/login/php/ajaxlogin.class.inc
johnsaigle: A more specific error might be better though and I think you should log `$e->getMessage()`. `Exception` will catch _any_ exception that occurs in this whole call chain so this could hide problems by accident.

modules/login/php/ajaxlogin.class.inc
johnsaigle: In that case you should provide more detailed testing instructions. At minimum make sure that the `TestPlan.md` is up-to-date with your new changes and that the module works according to what it says.

modules/login/php/ajaxlogin.class.inc
maltheism: @johnsaigle I tested the frontend with adding a public key of my own and the reCaptcha does display on LORIS on this PR. I haven't debugged the backend but I assume it never has worked and nobody has ever used reCaptcha. Since the Content-Security was never allowing reCaptcha in the first place because iframe-src would have broke it and I had to add that rule.

maltheism:
@johnsaigle I've used GitHub's resolve before and it was okay. I'll try doing it how you suggest.

modules/login/php/ajaxlogin.class.inc
johnsaigle: Yeah I'm not surprised it never worked or hasn't for a long time. Definitely was the case for me but I never looked into it closely.

So you're saying that this code is more old i.e. not how Google does things anymore BUT it does work now, right?

modules/login/php/ajaxlogin.class.inc
maltheism: @johnsaigle I guess it should be a is_null check for $user.

modules/login/php/ajaxlogin.class.inc
maltheism: @johnsaigle the frontend now shows reCaptcha but I didn't debug through the backend to see if it matters or not.

modules/login/php/ajaxlogin.class.inc
maltheism: @johnsaigle I'm hoping to see what travis complains about.

modules/login/php/ajaxlogin.class.inc
johnsaigle: It also can't return null if you check the signature in `php/libraries/User.class.inc`. It either returns a User object (singleton) with all the correct data or an AnonymousUser object.

modules/login/php/ajaxlogin.class.inc
johnsaigle: That's fine Travis can definitely help you figure out if we can remove some manual testing steps. The TestPlan should still be up-to-date with how the module is written though.

In fact it looks like this module got passed over and has its own format for testing, which isn't good.

I'm going to tag this with `Needs Documentation`. Before this gets merged you should update consolidate the testing instructions into one document, remove typos, and make sure it accurately reflects the new state of the module. It'll help everyone with manual testing also.

modules/login/php/ajaxlogin.class.inc
maltheism: @johnsaigle okay thanks i'll remove it.

modules/login/php/ajaxlogin.class.inc
johnsaigle: Yeah it would be a good idea to test if the reCaptcha actually works rather than if it just shows up.

johnsaigle:
@maltheism I don't think it's a bad feature in every single scenario but it can cause a lot of issues with a codebase like ours that has so many moving pieces. In any case @driusan isn't gonna let you use it so try the other way.

maltheism:
Okay, travis is failing because new react code needs to reflect travis test cases for login. New commit might address the problem.

driusan:
added Needs Work because this is failing Travis.

sruthymathew123:
@maltheism As I was assigned to test this PR, I am adding my comments on this PR
1) Login -Works well as expected!
2) Request Account -> Account Created Successfully!
3) Password Expired-> NOT WORKING AS EXPECTED!
 I tried to login to the newly created account; the login is frozen with no error messages on front end.   By default the password is expired; but it is not redirecting me to update the password.
login_error
4) Reset Password - Update Password worked successfully!

maltheism:
Thanks @sruthymathew123 for testing. I'll take a look at the password Expired error. I tested it before submitting this PR and I thought it was working. Maybe your manual test was executed differently for the expiry case. The console.log screenshot helps a lot. I'll just search to where /login/password-expiry  still is in the code because the logic is not using that anymore.

maltheism:
Hey @sruthymathew123, could you try again with an account that has an expired password. 
I re-read your steps for the expiry case and I think you may have attempted with a new user that requires the admin to change the pending_approval to No. LORIS doesn't allow a new account to login or even view anything (including a notification) when the account is pending approval (I think). I've tested the password expiry case by changing the database column "Pending_approval" for table users to 1990-09-18. That triggers the expiry page when the user attempts to login.

jsx/Form.js
```diff
@@ -909,15 +915,21 @@ class TextboxElement extends Component {
...
-        <div className="col-sm-9">
+      );
+    }
+    return (
+      <div className={elementClass}>
+        {label}
+        <div className={this.props.class}>
           <input
-            type="text"
+            type={this.props.type}
```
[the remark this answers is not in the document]
maltheism: @zaliqarosli when the type=password.

jsx/Form.js
```diff
@@ -1334,13 +1353,20 @@ class StaticElement extends Component {
...
-        <div className="col-sm-9">
-          <p className="form-control-static">{this.props.text}</p>
+      );
+    }
+
+    return (
+      <div className="row form-group">
+        {label}
+        <div className={this.props.class}>
+          <p className={this.props.textClass}>{this.props.text}</p>
```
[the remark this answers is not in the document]
maltheism: @zaliqarosli if you take a look at login or checkout my PR, you will notice why I added the changes to the jsx/Form.js. As an example the styles & classes for the input fields in the form needed to be modified to display the login fields as perfectly as before. Otherwise the labels being removed (or even kept) would look ugly.

jsx/Form.js
```diff
@@ -1428,25 +1458,34 @@ class CheckboxElement extends React.Component {
     // Add error message
     if (this.props.errorMessage) {
       errorMessage = <span>{this.props.errorMessage}</span>;
-      elementClass = 'checkbox-inline col-sm-offset-3 has-error';
+      elementClass = this.props.class === 'checkbox-inline col-sm-offset-3'
+        ? 'checkbox-inline col-sm-offset-3 has-error'
```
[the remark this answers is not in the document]
maltheism: @zaliqarosli it totally makes sense to do that. Thanks i'll make that change.

jsx/Form.js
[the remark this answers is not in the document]
zaliqarosli: could we maybe add a boolean isPassword prop that changes the type to password if true, and otherwise, text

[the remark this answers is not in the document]
zaliqarosli: orr would be cool to have a separate password input react component

jsx/Form.js
```diff
@@ -1428,25 +1458,34 @@ class CheckboxElement extends React.Component {
...
-            checked={this.props.value}
-            required={required}
-            disabled={disabled}
-            onChange={this.handleChange}
-          />
-          {errorMessage}
-          {this.props.label}
-          {requiredHTML}
-        </label>
+        <div className={'col-sm-12'}>
```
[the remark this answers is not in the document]
maltheism: @zaliqarosli the checkbox component was a disaster trying to use with this login component. The alignment specifically with the other form elements was off and the style wasn't working well with the react Form component; specifically when labels are not used. I'm unhappy with it right now from seeing the above screenshot and I agree with your above comment the border should be removed. It looks better right now for the login module. I've not used examiner before and it looks awful from your screenshot.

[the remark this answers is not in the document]
zaliqarosli: a checkbox shouldn't be rendered without a label. and this is what it looks like on major for examiner: 
Screenshot 2019-03-18 12 42 37 which I think works. the alignment doesn't bother me that much but if we wanted it to alignment the checkbox with the other elements, we could change the "padding-left" attribute of the ".checkbox-inline" bootstrap css. other than that, i think the checkbox renders fine on major. what other issues are you having the rendering?

jsx/Form.js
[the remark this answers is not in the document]
maltheism: @zaliqarosli also I made the default be "text" and I don't really think it's bad to have this option. I don't think we should design the react component in this case to where it restricts the developer from modifying the html element value.

jsx/Form.js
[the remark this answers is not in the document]
maltheism: @zaliqarosli you're misunderstanding my original comment. I mean when the other form components are not using a label. Please see the screen shot:
example1
Such as the First name, Last name, Email address, (the site options),...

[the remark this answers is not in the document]
maltheism: @zaliqarosli I also think the above screenshot looks really nice compared to if all the form components were displaying labels.

jsx/Form.js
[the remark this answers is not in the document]
zaliqarosli: true! as long as the default is text, then i think we're fine

jsx/Form.js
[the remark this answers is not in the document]
zaliqarosli: i think in reactifying, we should only really be reactifying and leaving any style changes for another PR / in consultation with the UX group

modules/login/jsx/passwordExpiry.js
```diff
@@ -0,0 +1,185 @@
...
+          />
+        </FormElement>
+        <a onClick={() => window.location.href = window.location.origin}
+           style={{cursor: 'pointer'}}>Back to login page</a>
+      </div>
+    ) : (
+      <div className={'success-message'}>
+        <h1>Thank you!</h1>
+        <p>Your request for an account has been received successfully.</p>
+        <a onClick={() => window.location.href = window.location.origin}
```
sruthymathew123: @maltheism I think we should also add "Please contact your project administrator to activate this account" to avoid confusions to the user.

jsx/Form.js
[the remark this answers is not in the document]
HenriRabalais: Just my two cents, but I would think that if the input field is taking on an entirely new function it should probably deserve it's on component called `PasswordElement`. Perhaps Passwords and Textboxes are close enough that that isn't necessary though. Just something to consider. 

I also agree that if we are to keep passwords as `TexboxElement`, thhen a boolean prop of `password` would be better than passing any type you want to the `TextboxElement`

jsx/Form.js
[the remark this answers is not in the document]
maltheism: @zaliqarosli The current react components for the form, will produce a really ugly login page without the added changes that are just giving the components more properties to access classes which in my opinion should be an option by default.

sruthymathew123:
@maltheism Password Expired is working now. As people are suggesting changes; will retest after making all the requested changes.

jsx/Form.js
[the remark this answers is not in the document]
maltheism: @HenriRabalais @zaliqarosli I think making a new component for a password input is unnecessary and kind of ridiculous if I'm playing devil's advocate. I'll try to express why I think components should be designed without restricting the developer from accessing or changing the original html values in simple cases such as type=password. We shouldn't add properties with "new names" that need to be found & learned by developers and to do things of common knowledge in the html world. Most developers with html knowledge know how to change an input type to not display the text when it's a password being inputed.
I'm only writing this because I think the components right now are restricting access to classes and it shouldn't be done this way.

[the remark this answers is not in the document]
HenriRabalais: Given the current structure of `Form.js` I don't think creating an entirely new component seems odd. Actually, it seems to me that a different input types are the distinguishing characteristic of defining a new component (select, button, date, etc.) in `Form.js`. I don't know everything that inputs of `type=password` are capable of, but if there are many options for them they could be expanded upon within a `PasswordElement` component.

The concern specifically with the change to `type={this.props.type}` is that someone could easily pass the value 'date' as a type prop. This shouldn't be allowed. There's a DateElement which provides other options specific to dates that should be used. A password prop flag would be able to handle this password configuration is a clearer and more controlled way. 

I may be wrong, but I think restricting access to classes and html functionality is a bit of the point. By doing this all forms across loris will be standardised. If components become more lenient, there will be less internal coherence within loris. I know that can be frustrating (I've been struggling a lot with Form.js because of Biobank), but I think the solution is to introduce new features (e.g. inline form, label-less form) rather than making everything customisable.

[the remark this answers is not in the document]
maltheism: @HenriRabalais Idk I just find it limiting and too much of encapsulation. I agree a developer could go and make a date input but I don't necessarily think this PR with the password is comparable to that situation. Now if the password input was to be heavily stylized with it's own feature of frontend verification for matching LORIS password rules.. then I would find an individual password component as a valid reason for existence because the logic would be associated with it and inside the component.
The classes changes were actually needed to preserve the Login style of LORIS. Please look at the login page and even test the PR. The indentation with labels looked awful on the Login Page, before I added the modifications to not require labels and which shouldn't break the other forms. I'm not sure if you're suggesting we have a new component for without label? I'm alright if that is to be done but I think the changes I made aren't bad.

jsx/Form.js
```diff
@@ -1428,25 +1458,34 @@ class CheckboxElement extends React.Component {
     // Add error message
     if (this.props.errorMessage) {
       errorMessage = <span>{this.props.errorMessage}</span>;
-      elementClass = 'checkbox-inline col-sm-offset-3 has-error';
+      elementClass = this.props.class === 'checkbox-inline' + ' ' + this.props.offset
```
zaliqarosli:
```suggestion
      elementClass = this.props.class + ' ' + this.props.offset + ' has-error';
```

jsx/Form.js
```diff
@@ -1428,25 +1458,34 @@ class CheckboxElement extends React.Component {
...
+        ? 'checkbox-inline col-sm-offset-3 has-error'
```
zaliqarosli:
```suggestion

```

jsx/Form.js
```diff
@@ -1428,25 +1458,34 @@ class CheckboxElement extends React.Component {
...
+        : 'checkbox-inline col-sm-offset-2 has-error';
```
zaliqarosli:
```suggestion

```

jsx/Form.js
```diff
@@ -1428,25 +1458,33 @@ class CheckboxElement extends React.Component {
     // Add error message
     if (this.props.errorMessage) {
       errorMessage = <span>{this.props.errorMessage}</span>;
-      elementClass = 'checkbox-inline col-sm-offset-3 has-error';
+      elementClass = this.props.class = 'checkbox-inline'
```
zaliqarosli: this still doesn't look right. i think it should actually just be `elementClass = elementClass + ' has-error'` no?

maltheism: I messed up there pretty bad. Thanks @zaliqarosli !

maltheism:
@zaliqarosli I updated the style to remove the border around checkboxes. It also should now look good on examiner and login. Please verify for me.

maltheism:
@kongtiaowang I'm having trouble getting the testLoginSuccess to succeed.

kongtiaowang:
@maltheism Monday, I will debug this.

modules/login/jsx/passwordExpiry.js
zaliqarosli: @maltheism this still needs to be implemented?

zaliqarosli:
on testing:

1. buttons seem to be misaligned on 3 pages. this is also making the 'Login to LORIS' panel not line up with the other panel.

Screenshot 2019-03-25 16 44 04

2. In 'Request Account' panel, Sites select element should have 'Choose your site:' as the empty option. Otherwise, the first option is set as default.

3. In 'Request Account', checkboxes line up with misaligned button, but not the other elements

4. 'Add Examiner' still broken for me

Screenshot 2019-03-25 16 51 52

johnsaigle:
Adding Needs Work given the above review and the Travis failure

kongtiaowang:
It works on Chrome, but not works on Firefox. 
When the page reloads,  lose the session data.

modules/login/php/ajaxlogin.class.inc
```diff
@@ -0,0 +1,407 @@
...
+    {
+
+        $login   = new SinglePointLogin();
+        $success = $login->passwordAuthenticate(
+            $values['username'],
+            $values['password'],
+            false
+        );
+        if ($success) {
+            // login success.
```
kongtiaowang: From here, a user already has the log in the permission, why not redirect the page here? Is it necessary to pass the login info to  REACT front-end to check again?

kongtiaowang:
I will make a pull request to update the testing browser to the latest version. (also update selenium and test cases, it needs some time, it is not easy.) This PR is going to fail on the lower version browser.

maltheism:
Hi @zaliqarosli, I made some changes and I'm not necessarily happy about the fix as I'll explain further. This is regarding the issues you commented [here](https://github.com/aces/Loris/pull/4395#issuecomment-476372974). It seems the button class: "btn" adds margin and making the button be not aligned. I added a style to the button to override it with margin set to zero. It seems that class is from bootstrap and I think we still need to eventually remove bootstrap or update it. I believe the checkboxes should be fixed appropriately. Let me know if you have any suggestions because one of the fixes does work but is sort of hackish.

johnsaigle:
This still `Needs Rebase` because of the conflicting files...

zaliqarosli:
@maltheism still some issues, let's talk about it offline tomorrow (Monday)

modules/login/php/passwordexpiry.class.inc
```diff
@@ -1,158 +0,0 @@
-<?php
```
zaliqarosli: is this correct? to delete the 3 php classes?

maltheism: the module no longer uses them.

jsx/Panel.js
```diff
@@ -46,21 +46,23 @@ class Panel extends Component {
...
+        data-toggle={this.props.collapsing ? 'collapse' : ''}
         data-target={'#' + this.props.id}
-        style={{cursor: 'pointer'}}
+        style={this.props.collapsing ?
+          {cursor: 'pointer'} : {cursor: 'default'}
+        }
       >
-        {this.props.title}
-        <span className={glyphClass}></span>
+        <h3 className={'panel-title'}>{this.props.title}</h3>
```
zaliqarosli: this added `<h3>` might be why the glyphicon gets bumped down? not too sure but regardless, this style change looks ugly on the other modules. could we keep it to just `{this.props.title}`?

jsx/Panel.js
```diff
@@ -46,21 +46,23 @@ class Panel extends Component {
...
+        {this.props.collapsing ? <span className={glyphClass}/> : ''}
```
zaliqarosli: rendering ' ' if collapsing is false makes me feel bad :( could we put all this into a `const` and render that const instead?

jsx/Panel.js
zaliqarosli: i do see why you need this added header tag for the login module though. could we pass another prop called "bold" or something?

modules/login/jsx/passwordExpiry.js
zaliqarosli: is my above comment still relevant?

modules/login/jsx/requestAccount.js
```diff
@@ -0,0 +1,243 @@
...
+            type={'text'}
+            placeholder={'Email address'}
+          />
+          <SelectElement
+            name={'site'}
+            options={this.props.data.site}
+            value={this.state.form.value.site}
+            onUserInput={this.setForm}
+            class={'col-sm-12'}
+            emptyOption={false}
```
zaliqarosli: any way to add 'Choose your site:' as the "empty" option?

modules/login/php/passwordexpiry.class.inc
```diff
@@ -1,158 +0,0 @@
...
-
-    /**
-     * _process updates the password of the user and then redirects to the
-     * dashboard.
-     *
-     * @param array $values The values that were submitted to the page.
-     *
-     * @return void
-     */
-    function _process($values)
```
zaliqarosli: surely we need this process function to update expire passwords? or did you move this somewhere else?

maltheism: @zaliqarosli that functionality was moved to the ajaxlogin.class.inc

jsx/Form.js
```diff
@@ -1497,9 +1538,11 @@ class ButtonElement extends Component {
       <div className="row form-group">
         <div className={this.props.columnSize}>
           <button
+            {...(this.props.id ? {id: this.props.id} : {})}
```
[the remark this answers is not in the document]
maltheism: @zaliqarosli It adds id: '' to the button if one is supplied and otherwise nothing is added. You can test it.

jsx/Form.js
```diff
@@ -1497,9 +1538,11 @@ class ButtonElement extends Component {
...
             name={this.props.name}
             type={this.props.type}
             className={this.props.buttonClass}
+            {...(this.props.style ? {style: this.props.style} : {})}
```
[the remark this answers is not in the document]
maltheism: Similar to the comment above but for adding a style. I'm assuming there's no right way of doing it. I thought this way I came up with is slick.

jsx/Panel.js
```diff
@@ -46,21 +46,23 @@ class Panel extends Component {
     // Add panel header, if title is set
     const panelHeading = this.props.title ? (
       <div
-        className="panel-heading"
+        className='panel-heading'
         onClick={this.toggleCollapsed}
-        data-toggle="collapse"
+        data-toggle={this.props.collapsing ? 'collapse' : ''}
```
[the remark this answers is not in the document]
maltheism: @zaliqarosli Hmm I should be able to fix this by setting the correct default property.

jsx/Panel.js
maltheism: @zaliqarosli awesome fixed.

jsx/Panel.js
maltheism: @zaliqarosli done and made it null if false

modules/login/jsx/requestAccount.js
maltheism: @zaliqarosli I added some code to resolve this problem.

jsx/Form.js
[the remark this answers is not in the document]
HenriRabalais: I think it may be the way that's written that is confusing. The following might be a bit more straight forward.
`style={this.props.style ? this.props.style : {}}`
`style={this.props.style && this.props.style}`

 I'm also not sure if you have to do the check in the first place. I believe if you pass a value of `null` to a prop, it does not get passed to the child.

jsx/Form.js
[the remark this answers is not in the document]
HenriRabalais: I'm pretty sure you can simply do the following:
`id={this.props.id}`
OR 
`id={this.props.id || undefined}`
etc.

I think the use of a spread operator here is a bit confusing on unnecessary

jsx/Form.js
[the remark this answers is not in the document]
maltheism: @HenriRabalais Okay see latest changes. Previously I was worried if not supplied it would be added but react handles that.

jsx/Form.js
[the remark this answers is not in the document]
maltheism: @HenriRabalais Okay see latest changes. Previously I was worried if not supplied it would be added but react handles that.

johnsaigle:
This is failing Travis. I don't think it's able to login successfully. The integration tests probably need updating.

maltheism:
Hey @johnsaigle I think @kongtiaowang mentioned he has to fix something with Selenium or Travis to get this PR to work. Login should be able to work with this PR because it's been manually tested. I'm waiting on the fix. I think this PR fixes some bugs as well as now the module is react.

johnsaigle:
@kongtiaowang Do you remember whether the problem Alizée mentioned is addressed?

johnsaigle:
Adding the Security label because this PR addresses an issue with recaptcha

maltheism:
Messed up rebase (see #5225).

PR #4397, opened 2019-03-15
[Instrument] Fix instrument typehinting

ridz1208:
### Brief summary of changes
a null value should be allowed here for all the `field{@}=={@}NEVER_REQUIRED` cases of xin rules.

By default any field is required by xin rules. in order to make fields not required, a rule is added to them that translates into `this field is only required if this field is equal to NEVER_REQUIRED` this bypasses the mandatory nature of xin rules on instrument fields.

If if a field have NEVER REQUIRED on it then it will probably not be filled in and thus this value here is then null which should be okai. without this PR all instruments are broken

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1339,7 +1339,7 @@ class NDB_BVL_Instrument extends NDB_Page
      * @return boolean true if element should be required, false otherwise.
      */
     function XINRunRuleFunction(
-        string $controller,
+        ?string $controller,
         array $values,
         string $operator
     ): bool {
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1248,7 +1248,7 @@ class NDB_BVL_Instrument extends NDB_Page
      *
      * **** DONT TOUCH THIS, EVER. ****
      *
-     * @param ?string $elname   The name of the element being validated
+     * @param string $elname   The name of the element being validated
      * @param array   $elements Array of values submitted on current page.
      * @param array   $rules    Array of rules that have been registered.
      *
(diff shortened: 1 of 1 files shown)


php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1331,15 +1331,15 @@ class NDB_BVL_Instrument extends NDB_Page
     /**
      * Run a XIN rule on an individual element
      *
-     * @param string $controller The element which affects the validity of this
+     * @param ?string $controller The element which affects the validity of this
```
driusan: I'm confused about how this works with "NEVER_REQUIRED" as the PR describes. Wouldn't this change simply make instruments always claim that the rule is broken due to lines 1355-1356?

maltheism: @ridz1208 @driusan my interpretation of this change is at line 1303 will result in the $ElementResult having the value of false because the operator case would be ==. Otherwise without this PR the function XINRunRuleFunction doesn't except a $controller being null. Am I missing something?

driusan: But doesn't it having a value of false mean it'll be flagged as being required when it's explicitly flagged as "NEVER_REQUIRED"? (The whole logic of this function is confusing.)

PapillonMcGill: @ridz1208 Could you clarify the summary of changes?

maltheism: @driusan it's confusing to me. I'm under the impression the The XINRunRuleFunction function comment states: `@return boolean true if element should be required, false otherwise.` and I'm currently thinking it does return false for a null $controller because I'm assuming $value isn't null because the function cannot take a null value with the type hinting. I'm also under the impression that the case is always '==' for NEVER_REQUIRED and never anything else to throw the LorisException.

driusan: We can probably merge this since it just brings it (closer) to the state it was in before the type hinting was added, but I think it would be a good idea to write some unit tests for this function.

ridz1208: The way I understand it (and I barely do) is:

if a field is optional, to allow it to be empty we add a rule that saays `only require fieldX when fieldX==NEVER_REQUIRED` that is a statementthat males it always not required unless it is actually `NEVER_REQUIRED` in which case there is no error because it was filled in and it was required... fun stuff. There is no special case for the string `NEVER_REQUIRED` it could be `RIDA_IS_THE_BEST` and this would still work

So... yeah it works and without this change it will break for anyone that uses this logic to bypass xin rules

PR #4402, opened 2019-03-18
[User Accounts] Fix bug where passwords were recorded in plaintext in history table

johnsaigle:
### Brief summary of changes

Fixes a bug that was logging plaintext passwords. Also deletes the logged passwords and expires all active passwords in order to trigger a reset.

This is a new version of #4021 that removes the blacklisting functionality because it was controversial.


modules/user_accounts/php/edit_user.class.inc
```diff
@@ -454,12 +455,10 @@ class Edit_User extends \NDB_Form
...
-        // prepend two random characters
-        if (isset($set['Password_hash'])) {
-            // Update CouchDB. Must do before password is salted/hashed.
-            $expiry = isset($values['Password_expiry'])
-                ? $values['Password_expiry'] : null;
-            $user->updatePassword($set['Password_hash'], $expiry);
+        // Now set the password. Note that this field is named incorrectly
+        // and represents a plaintext password, not a hash.
+        if (isset($values['Password_hash'])) {
+            $user->updatePassword($values['Password_hash']);
```
driusan: what happened to Password_expiry? (Same with below.)

tools/single_use/remove_logged_passwords.php
```diff
@@ -0,0 +1,47 @@
...
+require_once __DIR__ . "/../generic_includes.php";
+$info = <<<INFO
+This script deletes occurrences of plaintext passwords that were stored in the
+`history` table.
+
+This script will also expire all current passwords for LORIS users. 
+
+The plaintext passwords appeared due to a bug in LORIS. As these passwords
+have potentially been viewed by administrators or could be read in the event
+of a data breach, we have chosen to reject them in this way.
```
driusan: The "expire all passwords" should be an option (or disabling it should be an option, at least) toggled by a flag to the script.

johnsaigle: I'm reluctant to hamstring this fix. Since the passwords are available in plaintext in the back-end I think they should be considered unfit to use in the future. This was the original idea behind implementing a blacklist. While the fix no longer includes the blacklisting functionality, I feel that forcing a password reset is a minimum requirement to properly address this issue.

driusan: Letting projects decide their own policy rather than forcing yours on them isn't "hamstringing the fix". The only people who potentially could read from the history table/audit trail (which isn't exposed anywhere) can already trivially change a user's password* if they wanted to access their account. (In fact, given that they have to already have backend access, they could even change the password, login as the user, then update the password hash back to the original in the user table to do it without a trace, with or without this fix.) Expiring the password in LORIS does not add any extra security, as the risk is password re-use on other sites that you don't have any ability to expire passwords on. Choosing to do so or not should be up to the projects using LORIS to decide.

* There are also scripts and frontends where an admin could reset a user's password to a known value that are even less restrictive than update access.

johnsaigle: This, I feel, is the responsible path for dealing with the issue of an internal password exposure:
* Delete the logged plaintext passwords
* Fix the code so that this doesn't happen again
* Notify the users that their password have been revealed in the database
* Reset the logged passwords
* Strongly recommend that the users reset passwords on other sites where they use the same password (if any -- but it's probable)

I think the level of risk is greater than you're making it out to be. You are describing what a current project administrator can do but this is not the only threat scenario. These passwords would also have been visible to people that no longer work at the facilities using LORIS who no longer have any other means of accessing the software. There is also the real possibility that the database could have been hacked into... even now we have a known SQL injection vector so it's not an imagined risk.

As you said a primary issue here is password reuse and of course we can't control what users do on other sites. But we can encourage best practices within the LORIS domain by triggering a password reset along with an explanation of what occurred. This puts the potential of exposure in the user's mind and if they are made to change their LORIS password because it has potentially been revealed, they may also reset their passwords elsewhere if they are prudent. So we can't reset passwords elsewhere but we can suggest that users do so of their own choice. A quieter patch without a reset makes that moment of reflection less likely which means a greater risk to the people involved.

We should respond to this bug in the best way possible and a password reset + explanation is the best way. As always a project administrator is able to edit or ignore the patch; nobody is being forced to do anything.

modules/user_accounts/php/edit_user.class.inc
johnsaigle: There is no `Password_expiry` in the HTML of these pages. I believe they were copy-pasted from elsewhere once upon a time. Think that's also why there are references to CouchDB.

tools/single_use/remove_logged_passwords.php
driusan: There is no proof of internal password exposure. There is a theoretical possibility of it because it was logged in the most restrictive place which is never intentionally exposed anywhere. If you have any evidence of someone having illegally accessed it from there with the intent of stealing user's passwords, you should be filling a criminal report, not a pull request.

Even though it's probably impossible to tell if someone *did* access it, your script precludes the possibility of notifying users who *may* have had someone read it, since it deletes entries in the history table without having logged who they were first and just blindly expires everyone's password.

If you don't want to add a flag to to allow projects to decide their own policies on passwords expiry without having to modify code you should (in fact, it's probably a better idea than having a flag now that I think about it) be printing a report of "UserID Email Active Y/N (Date(s) Of Change)" before deleting them so that they can be contacted. After having a query that does that, those are the only users who are at any risk and should have their password expired.

johnsaigle:
@driusan I've updated this script based on your feedback and our discussion yesterday,

Example output after the changes:

> This script deletes occurrences of plaintext passwords that were stored in the
`history` table.
The plaintext passwords appeared due to a bug in LORIS. As these passwords
have potentially been viewed by administrators or could be read in the event
of a data breach, we have chosen to delete them from the history table.
Additionally, users with potentially exposed passwords will be forced to
change their passwords.
The following users' passwords may have been exposed: 
	Username: admin
	Email: john.saigle+sandbox@mcin.ca
	Date of Password Change: 2019-05-08 12:38:19
	Active? Y
	Username: passboy
	Email: john.saigle+passwordtest@mcin.ca
	Date of Password Change: 2019-05-22 11:46:05
	Active? Y
These users should be contacted and informed of the potential password exposure and encouraged to change their passwords in other software if their LORIS password is also used elsewhere.

There is more spacing out of the blocks in the actual execution but it breaks GitHub formatting.

modules/user_accounts/php/my_preferences.class.inc
```diff
@@ -192,25 +192,23 @@ class My_Preferences extends \NDB_Form
...
         // update the user
         $success = $user->update($set);
 
-        // prepend two random characters
-        if (isset($set['Password_hash'])) {
-            // Update CouchDB. Must do before password is salted/hashed.
-            $expiry = isset($values['Password_expiry'])
-                ? $values['Password_expiry'] : null;
-            $user->updatePassword($set['Password_hash'], $expiry);
+        // Now set the password. Note that this field is named incorrectly
```
driusan:
```suggestion
        // Now set the password. Note that this field is named incorrectly at this point in the code
```

to remove ambiguity (the database field stores a hash, it's just the UI form name that's not named correctly at this point)

driusan:
@ridz1208 any thoughts on if this should be rebased onto 21.0-release (which is likely to be released before the next minor release)?

ridz1208:
@johnsaigle 
21 !!!!!

driusan:
@johnsaigle approved other than needing to be rebased onto the release branch and the fact that it's failing travis because of line length with the comment clarification..

johnsaigle:
Not sure why I'm getting a conflict. I didn't modify that file in the latest commits...

driusan:
What conflict? It says the branch has no conflicts to me..

PR #4180, opened 2018-12-04
[JSX] Survey Accounts - Reactify Menu Filter

HenriRabalais:
### Brief summary of changes
Reactifies Survey Account Menu Filter!

### To test this change...
Make sure that all the datatable filters are functioning.

PR #4403, opened 2019-03-19
[Survey Module] Fix broken survey URL

zaliqarosli:
### Brief summary of changes
Fixes some typehinting done in previous PRs on this branch

### This resolves issue...

- [ ] Broken link in reactification PR #4180

### To test this change...

- [ ] on major, go to Survey Accounts module, click on any link in the URL column of the table. Link breaks
- [ ] check out this PR branch and links work


zaliqarosli:
@johnsaigle please review. I didn't end up changing any of the variable names in case in breaks other things. thanks!

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -158,22 +158,22 @@ class NDB_BVL_Instrument extends NDB_Page
...
+     *                                      instrument is missing.
      *
      * @return object|null the new object of $instrument type
      * @access public
      */
     static function factory(
         string $instrument,
         string $commentID = '',
-        string $page = '',
+        ?string $page = null,
```
driusan: I don't think it makes sense to make this nullable, it just leads to confusion between the empty string and null. Why not just use the null coalesce operator with '' at the call site to fix the caller?

johnsaigle: Hmm yeah that's probably a better solution.

zaliqarosli:
@johnsaigle @driusan please re-review. thanks!

johnsaigle:
The new changes work for me!

zaliqarosli:
@johnsaigle thank you again lol

Issue #4394, opened 2019-03-13
TestPscIDLink() in candidate_list integration test is causing transitory failures on Travis


johnsaigle:
Haven't been able to figure out why but this test occasionally fails on Travis. I'll link back to this issue in future PRs when it causes build issues.

Curiously, this seems to have come up before... see ["Remove flaky test in candidate_list module"](https://github.com/aces/Loris/pull/3552) 

https://github.com/aces/Loris/blob/6c59ee0a00c473498a22bf527e7cbe9e71dae37f/modules/candidate_list/test/candidate_listTest.php#L288

johnsaigle:
This doesn't seem to be happening anymore so I'll close this.

PR #4404, opened 2019-03-19
[PHP Mess Detector] Add rules forbidding leftover debug statements and unused variables

johnsaigle:
### Brief summary of changes

Adds the modules directory to PHPMD to mark unused variables as well as
leftover debug statements from developers. This will prevent these
issues from being merged into the codebase in the future by causing Travis to fail
when they are detected. These issues will also show up when you run
`make checkstatic` locally.

Cleans up many unused variables in the modules/ directory, fixing some
bugs in the process (and hopefully nto creating new ones).

Also adds the PHPMD "Design Rules" with the exception of ExitExpression and
CouplingBetweenObjects which caused no errors.

### Reviewing
The deletions in this PR are the removal of variables that aren't used anyway. In some places I've optimized by rewriting code to be more readable and smaller.

You can validate the changes by checking out `major`, modifying `test/run-php-linter.sh` as I have here and running it. You will gaze upon a wall of errors.

On my branch you should not get these errors.


modules/login/php/requestaccount.class.inc
```diff
@@ -61,14 +60,6 @@ class RequestAccount extends \NDB_Form
         $this->addCheckbox("examiner", "Examiner role", array());
         $this->addCheckbox("radiologist", "Radiologist", array());
 
-        // Get reCATPCHA keys
-        $reCAPTCHAPrivate = $config->getSetting('reCAPTCHAPrivate');
```
johnsaigle: I deleted references to reCAPTCHA here as a temporary measure because many of the variables relating to its error reporting were marked as unused. @maltheism [discovered](https://github.com/aces/Loris/pull/4395#discussion_r265660761) that this feature hasn't been working in LORIS for some time now anyway so I don't think there's any harm done here. A separate issue/PR will be created by Alizee to address the future of reCAPTCHA in LORIS.

php/libraries/Database.class.inc
```diff
@@ -1463,5 +1463,20 @@ class Database
             );
         }
     }
+
+    /**
+     * Converts the output of a PDO array of errors into a printable string.
+     *
+     * @return string Printable representation of the error message.
+     */
+    public function createPDOErrorString(): string
```
johnsaigle: The creation of this function allows us to avoid using `print_r` in the Database class. This in turn allows us to flag all future inclusions of `print_r` and `var_dump` in the codebase as errors that are leftover from debugging.

test/run-php-linter.sh
```diff
@@ -18,4 +18,4 @@ vendor/bin/phpcs --standard=docs/LorisCS.xml --extensions=php/php htdocs || exit
 # Run PHPCS on all modules
 vendor/bin/phpcs --standard=docs/LorisCS.xml --extensions=php/php,inc/php modules/ || exit $?;
 
-vendor/bin/phpmd php/libraries text docs/LorisPHPMD.xml || exit $?;
+vendor/bin/phpmd php/,modules/ text docs/LorisPHPMD.xml || exit $?;
```
johnsaigle: This change makes it so that all sub-folders of `php/` as well as the `modules/` folder will be checked against our PHP Mess Detector ruleset. Previously only `php/libraries` was checked.

johnsaigle:
Failed (in part) due to 

Tagging #4394 as related

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -197,11 +197,8 @@ class Stats_Demographic extends \NDB_Form
         $subprojectQuery = '';
         // Only search for subprojects if $demographicProject is a positive int.
         if (is_numeric($demographicProject) && intval($demographicProject) >= 0) {
-            $subprojList ="";
             $subprojects = \Utility::getSubprojectsForProject($demographicProject);
-            foreach ($subprojects as $key=>$val) {
-                $subprojList .= $key.",";
-            }
+            $subprojList = implode(',', array_keys($subprojects));
             $subprojList = substr($subprojList, 0, -1);
```
PapillonMcGill: Should remove this line too, it was to remove an ending coma

modules/statistics/php/stats_mri.class.inc
```diff
@@ -243,12 +243,9 @@ class Stats_MRI extends \NDB_Form
...
 
         //SUBPROJECTS
-        $subprojList ="";
         $subprojects = \Utility::getSubprojectsForProject($currentProject);
-        foreach ($subprojects as $key=>$val) {
-            $subprojList .= $key.",";
-        }
-        $subprojList = substr($subprojList, 0, -1);
+        $subprojList = implode(',', array_keys($subprojects));
+        $subprojList = substr($subprojList, 0, -1); // why???
```
PapillonMcGill: delete this line, same as above

johnsaigle:
Thanks @PapillonMcGill. I removed the lines you recommended

docs/LorisPHPMD.xml
```diff
@@ -16,4 +16,9 @@
 			about them not being used -->
 		<exclude name="UnusedFormalParameter" />
 	</rule>
+
+    <rule ref="rulesets/design.xml">
+        <exclude name="CouplingBetweenObjects" />
+        <exclude name="ExitExpression" />
```
driusan: Going from the list of design rules at https://phpmd.org/rules/design.html, I think the only two we should be enforcing are "ExitExpression", "DevelopmentCodeFragment", and "EvalExpression" since they're things that people are already looking for in code review and it saves times to have Travis do them instead.

I don't think the others should be enforced, because they'll just cause people to be frustrated when they can't figure out Travis errors because of some arbitrary tool developer's opinion.

docs/LorisPHPMD.xml
```diff
@@ -16,4 +16,9 @@
 			about them not being used -->
 		<exclude name="UnusedFormalParameter" />
 	</rule>
+
+    <rule ref="rulesets/design.xml">
```
driusan: This change doesn't seem to be doing the same thing as the PR description and code changes say. The unused variables rules are in a different ruleset from what I can tell, not the design one.

docs/LorisPHPMD.xml
johnsaigle: Okay, I'll add the others to the excluded list. For now I think it would be best to continue to exclude `ExitExpression` because it caused a bunch of errors. These errors can be solved in another pull request to keep things simple.

docs/LorisPHPMD.xml
johnsaigle: > **Adds the modules directory to PHPMD** to mark unused variables as well as
leftover debug statements from developers...

The Unused variable rules are already active, but the `modules/` directory was excluded from analysis.

> **Also adds the PHPMD "Design Rules"** with the exception of ExitExpression and
CouplingBetweenObjects...

docs/LorisPHPMD.xml
driusan: Okay

test/run-php-linter.sh
driusan: what about the `src` directory?

modules/statistics/php/stats_behavioural.class.inc
```diff
@@ -119,11 +119,8 @@ class Stats_Behavioural extends \NDB_Form
         $suproject_query = '';
         // Only search for subprojects if $currentProject is a positive int.
         if (is_numeric($currentProject) && intval($currentProject) >= 0) {
-            $subprojList ="";
             $subprojects = \Utility::getSubprojectsForProject($currentProject);
-            foreach ($subprojects as $key=>$val) {
-                $subprojList .= $key.",";
-            }
+            $subprojList = implode(',', array_keys($subprojects));
             $subprojList = substr($subprojList, 0, -1);
```
maltheism: Hey @johnsaigle Is this line supposed to be removed?
`$subprojList = substr($subprojList, 0, -1);`
reason I ask is because you removed it similar in the modules/statistics/php/stats_demographic.class.inc

johnsaigle: Yep, thanks!

test/run-php-linter.sh
driusan: @johnsaigle what about the `src` directory?

johnsaigle: Oops I missed this. I can add the `src` directory.

johnsaigle:
I think this PR may not be worth developing, depending on how discussion goes in #4475

johnsaigle:
@cmadjar I saw you just tagged me to rebase and fix this PR but the discussion in #4475 needs to be resolved first.

cmadjar:
@johnsaigle ooops. Good point! Did not see that.

johnsaigle:
Reopening this PR because PHPMD has new maintainers and the unfixable bug from before may now be fixable.

johnsaigle:
Suppress Warnings will be needed to fix the issue in this PR. This requires PHPMD 2.7 which is added by #5350 so I'm adding the Blocked tag here.

driusan:
@johnsaigle conflicts

johnsaigle:
@driusan yes, I put the Needs Rebase tag on there to mark that. Thanks for merging the PHPMD update; now I can resume work on this.

johnsaigle:
This needs the PHPMD update that currently exists on the release branch.

modules/issue_tracker/php/my_issue_tracker.class.inc
```diff
@@ -0,0 +1,380 @@
+<?php
```
PapillonMcGill: Should these 2 issue_tracker/ files be part of this PR or left over from a rebase?

johnsaigle: they're leftover. I'll undo the changes

modules/behavioural_qc/php/behavioural_qc.class.inc
```diff
@@ -763,9 +767,7 @@ class Behavioural_QC extends \NDB_Form
      */
     function getTestNameusingMappedName($full_name, $subprojectID = null): string
     {
-        $DB     =& \Database::singleton();
-        $values = array();
-
+        $DB        = \Database::singleton();
```
ridz1208: since you are changing that line already, may use the factory ?

modules/api/php/views/visit/dicoms.class.inc
```diff
@@ -54,7 +54,10 @@ class Dicoms
         return array(
             'Tarname'    => $dicom->getTarname(),
             'SeriesInfo' => array_map(
-                'self::_formatSeries',
+                array(
+                    'self',
+                    '_formatSeries'
+                ),
```
ridz1208: I'm not sure this makes sense

ridz1208: nvm, TIL

modules/configuration/php/project.class.inc
```diff
@@ -53,7 +53,7 @@ class Project extends \NDB_Form
         $projectList    = \Utility::getProjectList();
         $subprojectList = \Utility::getSubprojectList();
         $projects       = array();
-        foreach ($projectList as $ProjectID => $Name) {
+        foreach ($projectList as $ProjectID) {
```
ridz1208: I don't think these are equivalent, after your change the value is actually being inserted in the `$ProjectID` variable, not the key, before your change the key was assigned to that variable

modules/configuration/php/subproject.class.inc
```diff
@@ -51,9 +51,9 @@ class Subproject extends \NDB_Form
 
         $subprojectList = \Utility::getSubprojectList();
         $subprojects    = array();
-        foreach ($subprojectList as $subprojectID => $title) {
+        foreach ($subprojectList as $subprojectID) {
```
ridz1208: same comment as above, this change switches from key assignment to value assignment to the same variable

modules/conflict_resolver/php/conflict_resolver.class.inc
```diff
@@ -255,8 +255,7 @@ class Conflict_Resolver extends \NDB_Menu_Filter_Form
     {
         parent::setup();
         // Create user object
-        $user   =& \User::singleton();
-        $config =& \NDB_Config::singleton();
+        $user = \User::singleton();
```
ridz1208: maybe use factory instead ?

modules/dicom_archive/php/dicom_archive.class.inc
```diff
@@ -71,12 +71,8 @@ class Dicom_Archive extends \DataFrameworkMenu
...
     {
         // get the list of sites available for the user
-        $list_of_sites = \Utility::getSiteList();
-        $site_options  = array();
-
-        foreach ($list_of_sites as $id => $name) {
-            $site_options[$name] = $name;
-        }
+        $site_names   = array_values(\Utility::getSiteList());
+        $site_options = array_combine($site_names, $site_names);
```
ridz1208: this works, i just find it more confusing. I think removing the `$id` would have had the same effect.

not a breaking change though

modules/document_repository/php/doctree.class.inc
```diff
@@ -145,30 +145,29 @@ class DocTree extends \NDB_Page
     {
         $result           = array();
         $allsubcategories = array();
-        $user = \User::singleton();
-        $db   = \Database::singleton();
+        $db = \Database::singleton();
```
ridz1208: factory ?

modules/document_repository/php/doctree.class.inc
```diff
@@ -99,7 +99,7 @@ class DocTree extends \NDB_Page
         if ($is_first_time) {
             $arr = [];
         }
-        foreach ($data as $key => $val) {
+        foreach (array_values($data) as $val) {
```
ridz1208: I don't think you need the array_values call.

modules/document_repository/php/doctree.class.inc
```diff
@@ -145,30 +145,29 @@ class DocTree extends \NDB_Page
     {
         $result           = array();
         $allsubcategories = array();
-        $user = \User::singleton();
-        $db   = \Database::singleton();
+        $db = \Database::singleton();
 
         $tree = $this->getTreeData();
-        $pids = $this->getAllNodesID($tree, $id);
+        $pids = array_values($this->getAllNodesID($tree, $id));
```
ridz1208: dont think array_values is needed here either

modules/document_repository/php/document_repository.class.inc
```diff
@@ -142,28 +142,6 @@ class Document_Repository extends \NDB_Menu_Filter_Form
              "id"   => $id,
          );
     }
-    /**
```
ridz1208: that took me way too long to process. thx for removing it :+1:

modules/examiner/php/editexaminer.class.inc
```diff
@@ -58,7 +55,7 @@ class EditExaminer extends \NDB_Form
                 array('EID' => $this->identifier)
             );
             $centerIDs =array();
-            foreach ($cids as $k=>$row) {
+            foreach (array_values($cids) as $row) {
```
ridz1208: same as above, dont think array_values is needed

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -239,7 +239,8 @@ class Instrument_Manager extends \NDB_Menu_Filter
                 . "Either: the instrument table exists (but is not in the "
                 . "test_names table), or "
                 . "LORIS could not connect to the database using the "
-                . "credentials supplied in the config file."
+                . "credentials supplied in the config file. "
+                . "Exec output: `" . implode("\n", $output) . "`"
```
ridz1208: is there a risk the password could be printed out here ?

PapillonMcGill: It could be seen in the running process list while it is running. (top, ps, ...)

modules/media/ajax/FileUpload.php
```diff
@@ -358,7 +358,7 @@ function toSelect($options, $item, $item2)
         $optionsValue = $item2;
     }
 
-    foreach ($options as $key => $value) {
```
ridz1208: this is a change of behaviour, the value would be stored in the`$key` variable

modules/next_stage/php/next_stage.class.inc
```diff
@@ -221,12 +221,9 @@ class Next_Stage extends \NDB_Form
...
     function _validate($values)
     {
-        $config =& \NDB_Config::singleton();
+        $config = \NDB_Config::singleton();
 
-        $timePoint =& \TimePoint::singleton($this->identifier);
-        $candID    = $timePoint->getCandID();
-
-        $candidate =& \Candidate::singleton($candID);
+        $timePoint = \TimePoint::singleton($this->identifier);
```
ridz1208: factory ?

modules/statistics/php/statistics_dd_site.class.inc
```diff
@@ -72,7 +72,7 @@ class Statistics_DD_Site extends statistics_site
     function _setInstrumentName()
     {
         $this->instruments = \Utility::getAllInstruments();
-        foreach ($this->instruments as $k=>$v) {
+        foreach ($this->instruments as $k) {
```
ridz1208: change of behaviour

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -1200,8 +1200,8 @@ class Edit_User extends \NDB_Form
         //======================================
         if ($editor->hasPermission('examiner_multisite')) {
             $matched = false;
-            foreach ($values as $k => $v) {
-                if (preg_match("/^ex_[0-9]+$/", $k)) {
+            foreach ($values as $v) {
+                if (preg_match("/^ex_[0-9]+$/", $v)) {
```
ridz1208: I think it's the key we should be matching on here, not the value. the value is only `on` or NULL

ridz1208:
mostly looks good, couple errors in travis that need to be fixed.

Most of my comments were 
- to use the factory instead of singleton...
- you use a lot `array_values()` but unless I'm mistaken the default of a foreach loop is to loop trough the values anyway.
- in some instances you are looping through values but assigneing them to the key variable. these might cause unexpected behaviour ! I think those are the most important changes to amend

johnsaigle:
> you use a lot array_values() but unless I'm mistaken the default of a foreach loop is to loop trough the values anyway.

Yes, but the key ends up being unused if that array is associative. The key-value changes are because of a PHPMD rule forbidding unused variables.

For example if you write:

```php
foreach ($array as $key => $value) {
    $square = $value * $value;
}
```

The value $k is technically declared but not used.

The goal of using `array_values()` is to avoid that error being flagged. I think it's a good change for the most part. For example, a lot of the time we end up using key-value for loops to parse arrays that are returned by `pselect*` functions. However, most of the time those queries would be better written as using `pselectCol` if what we actually care about are only the array values.

johnsaigle:
I'll change the singleton declarations and double-check that the loop behaviour isn't changed.

ridz1208:
@johnsaigle for the pselect -> pselectcol or pselectcolwithindex, I absolutely agree.

but for the "using too much array_values" what I meant was that

```
foreach (array_values($array) as $value) {}
```

I think is exactly the same as

```
foreach ($array as $value) {}
```

I'm not saying dont remove the $key variable, by all means please do. All i was saying is I dont think you need to use `array_values()` because by default the foreach loop uses the array values when you dont assign a key

johnsaigle:
> by default the foreach loop uses the array values when you dont assign a key

I didn't know this 😅 Thanks for the tip. I'll update the code.

modules/dicom_archive/php/dicom_archive.class.inc
johnsaigle: I'll change the comment to make it more clear

modules/document_repository/php/document_repository.class.inc
johnsaigle: No problem, I love deleting code

modules/instrument_manager/php/instrument_manager.class.inc
johnsaigle: It prints only the output of exec, in this case a MYSQL error code. It doesn't print the output of the command.

It might be possible to use `top` etc. to view it but this has been true since at least 2012 when the exec code above was written.

johnsaigle:
@ridz1208 This is ready for re-review

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -386,7 +386,7 @@ class Edit_User extends \NDB_Form
                 );
 
                 //get sites where user is already an examiner at for compare
-                foreach ($prev_sites as $row => $center) {
+                foreach (array_values($prev_sites) as $center) {
```
ridz1208: unnecessary array_values()

modules/dashboard/test/DashboardTest.php
```diff
@@ -599,7 +599,6 @@ public function testDocumentRepository()
         $this->safeGet($this->url . '/dashboard/');
         $bodyText = $this->webDriver->getPageSource();
         $this->assertContains("test.jpg", $bodyText);
-        $this->assertContains("NEW", $bodyText);
```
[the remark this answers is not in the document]
johnsaigle: I was getting issues with Travis tbh. I don't know if this is a good test.

modules/genomic_browser/php/cnv_browser.class.inc
```diff
@@ -327,7 +326,7 @@ class CNV_Browser extends \NDB_Menu_Filter
         );
 
         $platform_options = array('' => 'Any');
-        foreach ($platform_results as $row => $result) {
+        foreach (array_values($platform_results) as $result) {
```
johnsaigle:
```suggestion
        foreach ($platform_results as $result) {
```

modules/help_editor/php/helpfile.class.inc
```diff
@@ -98,15 +98,15 @@ class HelpFile
...
+     * @return true
      */
     function update($set)
     {
-        // create DB object
-        $DB = \Database::singleton();
-
         // update the help file
-        $DB->update('help', $set, array('helpID' => $this->helpID));
+        \Database::singleton()
```
johnsaigle:
```suggestion
        \NDB_Factory::singleton()->database()
```

modules/timepoint_list/php/module.class.inc
```diff
@@ -40,12 +40,19 @@ class Module extends \Module
...
+        // FIXME This should be reworked to not silently handle the try-catch
         try {
             $resp = parent::handle($request);
             if ($resp->getStatusCode() != 404) {
                   return $resp;
             }
         } catch (\NotFound $e) {
+            /**
+             * This will suppress EmptyCatchBlock
+             * warnings in this method
```
driusan: This warning type should just not be enforced by the phpmd config, it's a useless warning to have. Empty catch blocks are a perfectly valid thing to have.

modules/timepoint_list/php/timepoint_list.class.inc
```diff
@@ -277,9 +276,7 @@ class Timepoint_List extends \NDB_Menu
      */
     function getJSDependencies()
     {
-        $factory = \NDB_Factory::singleton();
-        $baseURL = $factory->settings()->getBaseURL();
-        $deps    = parent::getJSDependencies();
+        $deps = parent::getJSDependencies();
```
driusan: just delete the function, the whole thing is a no-op.

php/libraries/Database.class.inc
```diff
@@ -1446,5 +1446,20 @@ class Database
...
```
driusan: This shouldn't be public.

src/Data/ProvisionerInstance.php
```diff
@@ -97,6 +97,12 @@ public function map(Mapper $map) : Provisioner
     /**
      * Implements the execute function for the Provisioner interface.
      *
+     * This will suppress UnusedLocalVariable
+     * warnings in this method. PHPMD has some issues with anonymous
+     * functions, apparently.
+     *
+     * @SuppressWarnings(PHPMD.UnusedLocalVariable)
+     *
```
driusan: If PHPMD isn't working, remove the rule from the config, don't suppress it.

src/Data/Provisioners/DBRowProvisioner.php
```diff
@@ -72,6 +72,12 @@ public function __construct(string $query, array $params)
      * the cursor used by the PDO does not conflict with multiple calls
      * to the function, and to make the code more deterministic.
      *
+     * This will suppress UnusedLocalVariable
+     * warnings in this method. PHPMD has some issues with anonymous
+     * functions, apparently.
+     *
+     * @SuppressWarnings(PHPMD.UnusedLocalVariable)
+     *
```
driusan: same

modules/timepoint_list/php/module.class.inc
johnsaigle: I'll disable this in the PHPMD config. For reference, here's the explanation.

> Usually empty try-catch is a bad idea because you are silently swallowing an error condition and then continuing execution. Occasionally this may be the right thing to do, but often it's a sign that a developer saw an exception, didn't know what to do about it, and so used an empty catch to silence the problem.

driusan: There's a lot of assumptions in that explanation.

johnsaigle:
@driusan I believe I addressed all your points. Could you take another look?

modules/behavioural_qc/php/behavioural_qc.class.inc
```diff
@@ -489,13 +488,16 @@ class Behavioural_QC extends \NDB_Form
         $site=null,
         $project=null
     ) {
-        $config         =& \NDB_Config::singleton();
+        $config = \NDB_Factory::singleton()->config();
+        $DB     = \NDB_Factory::singleton()->database();
         $ddeInstruments = $config->getSetting('DoubleDataEntryInstruments');
-        for ($i=0; $i<count($ddeInstruments); ++$i) {
+
+        // Why is the first instrument skipped?
```
PapillonMcGill: Could we remove this comment as it should start at $ddeInstruments[0]?

driusan: oops, sorry, I clicked merge before seeing your comment. I'll send a PR to do that and make another change which I think shouldn't have been in this that I didn't want to block over.

PapillonMcGill: should have requested change.

driusan: Sent #5936

PR #4407, opened 2019-03-20
Make getFullName and getSubtestList abstract

### LINKED ISSUE #4405: Instrument creation / installation workflow is broken
In order to install a NDB_BVL_Instrument style instrument the following steps are required:
1. Use lorisform_parser to generate the ip_output.txt
2. Use generate_tables_sql_and_testNames with ip_output.txt to generate the SQL
3. Source the SQL file to create the instrument table and populate the test names table.

However, lorisform_parser seems to call getFullName(), which does a select on test_names to get the testname, resulting in step 1 being broken because of a circular dependency between steps 1 and 3.

driusan:
The NDB_BVL_Instrument class has two functions 
which select from the database to get metadata
about the instrument: getFullName and getSubtestList.

Neither of these are pieces of metadata that should
be coming from the database, since they're information
about the instrument itself, not data.

This makes the functions abstract, forcing instruments
to implement the functions themselves and return appropriate
values.

A trait is provided to provide the legacy behaviour for
existing instruments.

Fixes #4405.

### DIFF
--- a/php/libraries/LegacyInstrumentTrait.class.inc
+++ b/php/libraries/LegacyInstrumentTrait.class.inc
@@ -12,7 +12,6 @@
  * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
  * @link     https://www.github.com/aces/Loris/
  */
-namespace LORIS;
 
 /**
  * The LegacyInstrumentTrait provides implementations of
--- a/php/libraries/LegacyInstrumentTrait.class.inc
+++ b/php/libraries/LegacyInstrumentTrait.class.inc
@@ -1,0 +1,71 @@
+<?php
+/**
+ * This file contains the a trait to provide legacy behaviour
+ * (database lookup) for abstract functions getFullNames and
+ * getSubtestList.
+ *
+ * PHP Version 7
+ *
+ * @category Main
+ * @package  Behavioural
+ * @author   Loris team <info-loris.mni@mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+namespace LORIS;
+
+/**
+ * The LegacyInstrumentTrait provides implementations of
+ * abstract functions in NDB_BVL_Instrument which previously
+ * had implementations provided. It can be used as a mixin
+ * for existing instruments as a stop-gap until the existing
+ * instruments are updated.
+ *
+ * @category Main
+ * @package  Behavioural
+ * @author   Loris team <info-loris.mni@mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+trait LegacyInstrumentTrait
+{
+    /**
+     * Return the full, human readable name for the
+     * current instrument.
+     *
+     * @return string the full name of the instrument
+     */
+    public function getFullName(): string
+    {
+        $db = \Database::singleton();
+
+        $success = $db->pselectOne(
+            "SELECT Full_name FROM test_names WHERE Test_name=:TN",
+            array('TN' => $this->testName)
+        );
+        return $success;
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


php/libraries/LegacyInstrumentTrait.class.inc
```diff
@@ -0,0 +1,71 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

php/libraries/LegacyInstrumentTrait.class.inc
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * This file contains the a trait to provide legacy behaviour
```
johnsaigle:
```suggestion
 * This file contains a trait to provide legacy behaviour
```

driusan:
@kongtiaowang The error on this isn't very useful. It says:

```
Fatal error: Class Mock_NDB_BVL_Instrument_6e474f8f contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (NDB_BVL_Instrument::getSubtestList) in /app/vendor/phpunit/phpunit-mock-objects/src/Generator.php(264) : eval()'d code on line 1
```

Is there any way to figure out which test the error is coming from?

kongtiaowang:
@driusan 
PHP Fatal error:  Trait 'LORIS\LegacyInstrumentTrait' not found in /var/www/loris/php/libraries/NDB_BVL_Instrument_LINST.class.inc on line 31

php/libraries/NDB_BVL_Instrument_LINST.class.inc
```diff
@@ -23,6 +23,12 @@ namespace Loris\Behavioural;
  */
 class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
 {
+    /**
+     * FIXME: LINST instruments should parse the full name
+     * and the subtest names from the linst file, not select
+     * them from the database.
+     */
+    use \LORIS\LegacyInstrumentTrait;
```
johnsaigle: I think maybe you don't need this namespacing because it's in `php/libraries` not in `src/`?

driusan: It's not necessary, but I thought it would be useful.. maybe it's conflicting because the same namespace is used for the `src` directory, I'll remove it to see

driusan:
@kongtiaowang traits should be autoloaded just like classes from what I can tell, and the error seems to be coming from the mocks, not from the LINST class according to the error message. I removed the namespacing, so I don't think it's caused by a namespace conflict here..

kongtiaowang:
@driusan In this test/unittests/NDB_BVL_Instrument_ToJSON_Test.php, line 45 and 367
        $this->i = $this->getMockBuilder("\NDB_BVL_Instrument") .... 
"getMockBuilder" can't create a mock for an abstract class and abstract method.

kongtiaowang:
@driusan 
demo for abstract class and method:
abstract class AbstractClass
{
    public function concreteMethod()
    {
        return $this->abstractMethod();
    }

    public abstract function abstractMethod();
}

class AbstractClassTest extends TestCase
{
    public function testConcreteMethod()
    {
        $stub = $this->getMockForAbstractClass(AbstractClass::class);

        $stub->expects($this->any())
             ->method('abstractMethod')
             ->will($this->returnValue(true));

        $this->assertTrue($stub->concreteMethod());
    }
}

driusan:
thanks @kongtiaowang!

ridz1208:
@driusan I would just like to put it in writing that I disagree with this

driusan:
@ridz1208 What alternative solution do you have to break the cyclic dependency between the functions requiring the tables to be populated in order to run the scripts which populate the tables?

I'm open to other solutions (moreso for getFullName which is needed in various queries than for getSubtestList, which I see no reason to have in the database at all as it's only used to make unnecessary queries in places that already have an Instrument object..)

ridz1208:
@driusan agreed for getSubtestList.

For getFullName why not ask for it on the front end or command line at the time where you are running the tool ?

PR #4409, opened 2019-03-20
[cleanup] php/libraries cleanup of comments, unnecessary variables,..

maltheism:
### Brief summary of changes

General code cleanup inside php/libraries. Such as incorrect return types documented, spelling mistakes in the comments, unnecessary @throws or not included when an exception is thrown for the function and cleanup of unused variables.

Note: I configured my IDE to only check for @throws in the functions themselves.

### To test this change...

- [x] make checkstatic


php/libraries/ExternalIDGenerator.php
```diff
@@ -27,9 +27,7 @@ class ExternalIDGenerator extends SiteIDGenerator
     /**
      * Generates ExternalIDs.
      *
-     * @param ?string $prefix The site prefix to prepend to the ID value.
-     *
-     * @return void
+     * @param string|null $prefix The site prefix to prepend to the ID value.
```
johnsaigle: `?string` is a valid annotation, just FYI

php/libraries/NDB_BVL_Battery.class.inc
```diff
@@ -104,11 +104,12 @@ class NDB_BVL_Battery
      * @param integer $center_ID    The center of the candidate having a battery
      *                              created
      * @param boolean $firstVisit   Whether this is a first visit battery.
-     *                              true means ONLY first visit
-     *                              false means NOT first visit
-     *                              null can be any visit.
+     *                              true means ONLY first visit false
+     *                              means NOT first visit null can be any
```
johnsaigle: This param should probably be `boolean|null` then, even though that feels horrible.

php/libraries/ExternalIDGenerator.php
maltheism: @johnsaigle thanks, you want me to revert those? 
The string|null is more descriptive to me from being someone who isn't a php expert. Personally I had to google the ?string when we introduced type hinting to know what was going on. Although I can understand people would prefer the shorthand annotation. I can revert if you want.

php/libraries/SiteIDGenerator.php
```diff
@@ -219,19 +221,19 @@ private function _getIDSetting(
...
-     *                  from project/config.sml
+     *                            from project/config.sml
      * @param string $setting     The name of the variable for which we want the
-     *                  value.
-     *
-     * @throws \ConfigurationException
+     *                            value.
      *
-     * @return ?string
+     * @return string|null ?string
```
johnsaigle:
```suggestion
     * @return string|null
```

php/libraries/ExternalIDGenerator.php
johnsaigle: No it's not worth the effort of going back and changing it. I like it in the sense that it matches type hinting for params and return types but it is initially confusing. Just wanted to let you know that it's not incorrect.

php/libraries/NDB_BVL_Battery.class.inc
maltheism: @johnsaigle Null is equivalent to false for a boolean and so I don't think that's correct.

johnsaigle:
Related to #4404 as it also removes unused variables

php/libraries/NDB_BVL_Battery.class.inc
johnsaigle: I don;t think that's correct. `true` `false` and `null` are all different things in PHP. The function comment here makes it sound like 3 different code paths are taken depending on the value of this param.

johnsaigle: `null` is "falsey" but `null === false` is not a correct statement.

maltheism: @johnsaigle nice I'll add the boolean|null.

johnsaigle: Also maybe add a FIXME comment that this code should be refactored. This is a bad code smell

php/libraries/NDB_BVL_Battery.class.inc
```diff
@@ -107,6 +107,7 @@ class NDB_BVL_Battery
      *                              true means ONLY first visit false
      *                              means NOT first visit null can be any
      *                              visit.
+     * FIXME: Someone needs to refactor this code. Bad code smell!
```
johnsaigle: It would be good to say why. Something like "types cannot be enforced because the bool param abuses the fact that PHP allows setting this value to `null` resulting in 3 code paths from a boolean parameter which should only allow for 2...." etc.

maltheism: @johnsaigle added.

php/libraries/Database.class.inc
```diff
@@ -688,12 +696,15 @@ class Database
      * @param array        $params   The values to bind to the statement
      *                               while executing it
      * @param array        $options  A list of key=>value pairs.
-     *                               - nofetch : to prevent the fetchAll
-     *                                 function to be excuted. Useful for
-     *                                 insert or update prepared statements.
+     *                               - nofetch : to prevent the
+     *                               fetchAll function to be
+     *                               excuted. Useful for insert
```
```suggestion
     *                               executed. Useful for inserting
```

php/libraries/Database.class.inc
```diff
@@ -688,12 +696,15 @@ class Database
...
+     *                               or update prepared
```
```suggestion
     *                               or updating prepared
```

php/libraries/NDB_BVL_Battery.class.inc
```diff
@@ -97,18 +97,23 @@ class NDB_BVL_Battery
...
+     * @param string       $stage        Either 'visit' or 'screening'
+     * @param string       $visit_label  The visit label to create a battery for.
+     * @param integer      $center_ID    The center of the candidate having a battery
      *                              created
-     * @param boolean $firstVisit   Whether this is a first visit battery.
-     *                              true means ONLY first visit
-     *                              false means NOT first visit
-     *                              null can be any visit.
+     * @param boolean|null $firstVisit   Whether this is a first visit battery.
+     *                              true means ONLY first visit false
```
PapillonMcGill:
```suggestion
     *                              true means ONLY first visit
```

php/libraries/NDB_BVL_Battery.class.inc
```diff
@@ -97,18 +97,23 @@ class NDB_BVL_Battery
...
+     *                              means NOT first visit null can be any
```
PapillonMcGill:
```suggestion
     *                              false means NOT first visit
```

php/libraries/NDB_BVL_Battery.class.inc
```diff
@@ -97,18 +97,23 @@ class NDB_BVL_Battery
...
+     *                              visit.
```
PapillonMcGill:
```suggestion
     *                              null can be any visit.
```

will be clearer with one line per possible value. Could also add '- ' in front.

php/libraries/NDB_Caller.class.inc
```diff
@@ -570,7 +571,7 @@ class NDB_Caller
         string $prefixStr
     ): string {
         $lastFound = '';
-        // Start from the end of $suffixStr and work backwords, to make sure
+        // Start from the end of $suffixStr and work backwards, to make sure
```
PapillonMcGill:
```suggestion
        // Start from the end of $suffixStr and work backwards, to ensure
```

php/libraries/NDB_Notifier.class.inc
```diff
@@ -29,10 +29,7 @@ class NDB_Notifier extends NDB_Notifier_Abstract
      * @param array  $extra_tpl_data additional variables to be sent to the tpl
      * @param bool   $asAdmin        when true, sends notification as if
      *                               triggered by administrator not user. Emails
```
PapillonMcGill:
```suggestion
     *                               triggered by an administrator, not the user. Emails
```
will have to check line length

php/libraries/SessionID.php
```diff
@@ -40,7 +40,7 @@ public function getType(): string
      * This does not check for uniqueness in the database or any other
      * state-related facts.
      *
-     * This function is called by the contructor of ValidatableIdentifier
+     * This function is called by the constructor of ValidatableIdentifier
      * to ensure that no SessionID exists if its value is not valid.
```
PapillonMcGill: These 2 lines of comments seem to not match this function.

php/libraries/NDB_BVL_Instrument_LINST.class.inc
```diff
@@ -322,6 +322,7 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
      *                            registered.
      *
      * @return boolean true if the element passes, false if it fails a rule.
+     * @throws \LorisException
```
PapillonMcGill: All others `@throws` do not have a `\`. Should all be the same
@driusan Do we need to put a `\` in front on the exception?

johnsaigle:
Needs PHPCS

php/libraries/LorisForm.class.inc
```diff
@@ -119,7 +120,7 @@ class LorisForm
      *                        get added. Currently only the "class"
      *                        attribute gets added.
      *
-     * @return &array A reference to the array that was added to $this->form
+     * @return array A reference to the array that was added to $this->form
```
driusan: These changes are all wrong. The function(s) are returning references to arrays, not array values.

php/libraries/LorisForm.class.inc
```diff
@@ -1067,7 +1070,7 @@ class LorisForm
 
     /**
      * Generates the HTML to add to the page when rendered for a date
-     * element with the year formate only
+     * element with the year formatted only
```
PapillonMcGill:
```suggestion
     * element with only the year formatted.
```

php/libraries/NDB_BVL_Instrument_LINST.class.inc
driusan: `\` is required if you're in a namespace, optional if you're in the root namespace..

johnsaigle:
@driusan I think that this PR will be covered by currently open Phan and PHPCS PRs. The rules added fix the same errors and also do the extra step of making sure they don't happen again.

It might be best to close this (at least for now) as it's unmaintained.

maltheism:
I'm closing and will reissue new PR to replace this because of rebase failure.

PR #4410, opened 2019-03-20
Sync package-lock.json with package.json

### LINKED ISSUE #4408: package-lock.json: To commit or not to commit?
This file should either be committed explicitly to the repo or explicitly included in our `.gitignore`. Right now it floats around between branches and causes warnings when switching.

I'm not sure on the best practices here because I'm not really involved with JavaScript, but a decision should be made.

johnsaigle:
Brings `package-lock.json` in sync with `package.json`. At some point the latter was changed without the former being committed to the repo.

Resolves #4408

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,10 +20,10 @@
       "dev": true,
       "requires": {
         "@babel/types": "7.0.0-beta.44",
-        "jsesc": "^2.5.1",
-        "lodash": "^4.2.0",
-        "source-map": "^0.5.0",
-        "trim-right": "^1.0.1"
+        "jsesc": "2.5.2",
+        "lodash": "4.17.11",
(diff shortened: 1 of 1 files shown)


zaliqarosli:
not sure why mine and john's git diffs are different :/ could someone else run `make dev` and see what their git diff it?

maltheism:
@zaliqarosli is right. I'm guessing my package-lock.json is similar to hers. @johnsaigle your package-lock.json has updated packages and I'm not sure why.

johnsaigle:
The latest commit makes more sense than my initial one but I can't really comment on the accuracy of all of it

driusan:
I have no idea how to review this. I guess I just assume it's good based on the reviews?

PR #4411, opened 2019-03-20
[Core] Create Password class and refactor password updates

johnsaigle:
_By request, this PR has been updated to rework how password are changed in LORIS. This is to demonstrate the benefits of the class and how it is to be used._

### Brief summary of changes
This PR adds a Password class to LORIS.
* Creates a concept of a password in LORIS. 
    * Valid passwords are created based on strings and must meet or exceed a minimum length and be sufficiently complex as determined by the [Zxcvbn](https://github.com/dropbox/zxcvbn) library.
    * There are no character requirements on passwords; they must be at least 8 characters long and sufficiently complex. This means either mixing many character sets or using longer passwords.
    * The Password class checks only whether a Password is valid and well-formed; it does not do any state-based checks or interact with the Database.
* Adds unit tests for the Password class.
* Removes password responsibility from the User class
    * Deletes `User::isPasswordStrong()`, `User::newPassword()`, `\User::pwnedPassword` and `User::updatePassword()`
    * Password structure constraints are now handled by `Password`
    * State-based password operations are handled by `Utility` via `Utility::updatePassword()` and `Utility::passwordChanged`
         * The `updatePassword()` function accepts a User object, Password object, and optional DateTime used for password expiry. This is used to change a specific user's password.
         * The `passwordChanged()` function accepts a `User` object and a _plaintext_ password and uses `password_verify()` to check whether the proposed password is different from the DB password
* Reduces code duplication across the `login` and `user_accounts` modules by leveraging the new functionalities of `Utility` and `Password`
* Adds a new function `Utility::randomString()` that accepts an int `length` and string `keyspace` that allows LORIS to generate cryptographically secure pseudorandom strings
    * Calling code in LORIS previously using `User::newPassword()` now use `Utility::randomString()`

### Testing

Ensure that you are able to reset a user's password in each of the following ways:
1. By the Forgot Password form
    * This method should force you to reset the password after logging in
2. Via My Preferences
3. Via Edit User in User Accounts, specifying a manual password
4. Via Edit User in User Accounts, using a system generated password
    * This method should force you to reset the password after logging in

### Related

A Password class was requested to be created in #3961 

Removing password functionality from the User class was suggested in #4978 

This replaces some of the functionally from #3962. 

This change allows LORIS to satisfy many more NIST password requirements as detailed in #3314 

Stronger types are desirable #4210


php/libraries/Password.class.inc
```diff
@@ -0,0 +1,135 @@
...
+     */
+    private $_hash;
+
+    /**
+     * Ensures that the value submitted as a potential new password is
+     * long enough.
+     *
+     * @param string $value The proposed password value.
+     *
+     * @return bool Whether the submitted value is valid.
```
PapillonMcGill:
```suggestion
     * @return bool Whether the submitted value meet minimal length requirement.
```

php/libraries/Password.class.inc
```diff
@@ -0,0 +1,135 @@
...
+        return strlen($value) >= self::MIN_PASSWORD_LENGTH;
+    }
+
+    /**
+     * Ensures that the value submitted as a potential new password is
+     * well-formed.
+     *
+     * @param string $value The proposed password value.
+     *
+     * @return bool Whether the submitted value is valid.
```
PapillonMcGill:
```suggestion
     * @return bool Whether the submitted value meet minimum complexity requirement.
```

johnsaigle:
@PapillonMcGill There is a companion library for this on the JS side so yes we could add a Reactified version of this where a password strength meter appears as someone types a password. Its value should match what's on the back-end.

PapillonMcGill:
@johnsaigle Thanks for the confirmation of JS side code.

johnsaigle:
@nicolasbrossard Would you be able to review? You mentioned that you wanted a PR like this in #3961

php/libraries/Password.class.inc
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * Ensures that the value submitted as a potential new password is
+     * well-formed.
+     *
+     * @param string $value The proposed password value.
+     *
+     * @return bool Whether the submitted value meets the minimum score required
+     *              by Zxcvbn as defined by MIN_COMPLEXITY_SCORE.
```
PapillonMcGill:
```suggestion
     *              FROM Zxcvbn as defined by MIN_COMPLEXITY_SCORE.
```
As current, it look like zxcvbn require a specific score while it is in fact the (calculator) provide of the score.

kongtiaowang:
#4474
After this PR, you need to rebase your PR to pass Travis.

johnsaigle:
I'm going to add Passed Manual Tests because this class has unit tests and there is no actual code instantiating this class at the moment so I don't think it can cause problems.

PapillonMcGill:
@xlecours I mean that we should implement in JS code to check if a new password is valid, for change password screen, not when someone try to login.

johnsaigle:
@xlecours I disagree. I'm not sure if there is much benefit to having an 'invalid' password object that can exist in the code base. I think it would be preferable to have something like:

```php
try {
    $newPassword = new \Password($userInput);
    \Database::singleton()->update('users', array ('Password_hash', $newPassword));
    return \LORIS\Http\Response\OK();
catch (\DomainException) {
    return \LORIS\Http\Response\BadRequest();
}
```

As far as login/other places in the code, that could potentially be handled by another PR. I created this as a design idea. It doesn't actually get used anywhere in the code right now.

For the strength requirements, I know that the JWT/API functionality also makes use of the isPasswordStrong() function of the User class.

I think one option would be to create a JWTToken class that extends Password and overrides `checkSufficientComplexity` and has a different length value. That way it can follow the same structure but doesn't need to use Zxcvbn.

xlecours:
@PapillonMcGill Ok, but this PR only contains php code so my comment was about php.
@johnsaigle If it's only for password changes then this would work. But it will need to be adjusted, **IF**, we change the `passwordAuthenticate`, `pwnedPassword`, ... , functions signature to take a "UnvalidatedPassword".

Regarding JWTToken, they are not passwords or subclasses of passwords.

johnsaigle:
@xlecours I think `pwnedPassword` should actually be refactored and included into this class at some point because it also only matters on password updates.

johnsaigle:
`passwordAuthenticate` can probably just stay as a string. Combining front-end validation with the `Password` class (on update) will probably take care of any UI and validation requirements.

A password without any validation is just a string anyway.

johnsaigle:
@xlecours can you update your review? or let me know if there's something you still want to discuss.

driusan:
I don't think pwnedPassword should be refactored to include in this class since the API now requires a monthly fee to a third party, it shouldn't be that deeply embedded.

johnsaigle:
> the API now requires a monthly fee to a third party

source?

driusan:
https://www.troyhunt.com/authentication-and-the-have-i-been-pwned-api/#there-s-a-us-3-50-per-month-fee-to-use-the-api

johnsaigle:
> Edit: Just to be crystal clear, this doesn't impact Pwned Passwords. Cloudflare picks up pretty much all the costs for running that so the service is still freely accessible.

He offers multiple APIs; the one being restricted by the monthly fee is the one that has to do with email addresses.

The one we're using is not affected from what I can tell.

driusan:
Ah, good.. that seems like something he should have been less ambiguous about in the blog. What's the source for that thread?

johnsaigle:
It's the comment section at the bottom of that blog post you linked.

driusan:
Huh... apparently if you turn on javascript, there are comments..

johnsaigle:
Hahaha yeah NoScript is cool and all but it definitely makes everything harder

php/libraries/Password.class.inc
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Validates the $value param by examining its length and complexity.
+     *
+     * @param string $value The proposed password plaintext.
+     *
+     * @return void
+     *
+     * @throws \DomainException When password is too short or weak.
+     */
+    public function validate($value): void
```
xlecours: This should be private since it's only used inside it's own class. Unless a password would need to validate an other value once instanciated.

php/libraries/Password.class.inc
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Ensures that the value submitted as a potential new password is
+     * well-formed.
+     *
+     * @param string $value The proposed password value.
+     *
+     * @return bool Whether the submitted value meets the minimum score defined
+     *              above and given by Zxcvbn.
+     */
+    protected function hasSufficientComplexity(string $value): bool
```
xlecours: This should probably be private since it's only used inside this class.
But I wont debate school of thought if you object.

php/libraries/Password.class.inc
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Ensures that the value submitted as a potential new password is
+     * long enough.
+     *
+     * @param string $value The proposed password value.
+     *
+     * @return bool Whether the submitted value has length of at least
+     *              MIN_PASSWORD_LENGTH;
+     */
+    protected function hasSufficientLength(string $value): bool
```
xlecours: This should probably be private since it's only used inside this class.
But I wont debate school of thought if you object.

php/libraries/Password.class.inc
```diff
@@ -0,0 +1,137 @@
...
+     */
+    private const MIN_COMPLEXITY_SCORE = 3;
+
+    /**
+     * The minimum length of a valid password. This value does not need to be
+     * too long if MIN_COMPLEXITY_SCORE is high enough.
+     *
+     * @var int
+     */
+    public const MIN_PASSWORD_LENGTH = 8;
```
xlecours: private, like MIN_COMPLEXITY_SCORE

kongtiaowang:
PHPCS

xlecours:
The code seems fine. 
I don't think we should merge it until it is being used though. It is adding a class and a dependency with no gain (yet).

johnsaigle:
@xlecours I only added the class in this PR just like was done in #4055 and #4333. It's also targeted to major so we have a lot of time to use it and the dependency won't come into affect until probably 6-12 months from now.

I'm happy to add an example if necessary but I don't really see why this PR is different than those other ones and why it needs to be used first.

driusan:
I don't see how it's possible to review this without knowing how it's planned to be used. When do you need to pass a representation of a password that can only contain valid passwords?

The only time I can think of you need to have a password instantiated from plain text is before validating a new password while changing it (in which case an exception is inappropriate to use as control flow, since a user providing bad input is not exceptional behaviour.) The only other time you would even have a plain text password to construct the object with is when logging in, in which case throwing an exception from the constructor is also inappropriate (what happens to people who already passwords that don't meet these requirements? Can they log in? Can they change their password?)

johnsaigle:
OK I will also include password/login refactoring here.

php/libraries/Utility.class.inc
```diff
@@ -917,5 +919,100 @@ class Utility
...
+     * Generate a random string, using a cryptographically secure
+     * pseudorandom number generator (random_int)
+     *
+     * @param int     $length   How many characters do we want?
+     * @param ?string $keyspace A string of all possible characters
+     *                         to select from. Will default to the alphanum set.
+     *
+     * @return string
+     */
+    static function randomString(
```
johnsaigle: Note that this can also be used for survey_accounts URLs in the future.

test/unittests/PasswordTest.php
```diff
@@ -0,0 +1,87 @@
...
+     * dataProviders for constructor invalid values
+     */
+    public function invalidValues(): array
+    {
+        return array(
+            // Should fail for not meeting length requirements
+            [implode('', range(1, Password::getMinimumPasswordLength() - 1))],
+            // Should fail for being an alphabet subset
+            ['abcdefghi'],
+            // Should fail based on dictionary attack ('hunter2' is meme)
```
johnsaigle:
```suggestion
            // Should fail based on dictionary attack ('hunter2' is a meme)
```

test/unittests/PasswordTest.php
```diff
@@ -0,0 +1,87 @@
...
+            ['hunter22'],
+            // Should fail for simplicty
```
johnsaigle:
```suggestion
            // Should fail for simplicity
```

test/unittests/PasswordTest.php
```diff
@@ -0,0 +1,87 @@
...
+            ['8chars!!'],
+            // Should fail based on using a common name
+            ['johnnyloris'],
+            // Should fail for obvious reasons
+            ['password'],
+            // Should fail for using a recent year
+            ['Spring2016!'],
+            // Should fail for using just a few simple English words
+            ['i am cool'],
+            // Should fail for common L33T substituions
```
johnsaigle:
```suggestion
            // Should fail for common L33T substitutions
```

johnsaigle:
@xlecours @driusan I've changed this PR to include a refactor of LORIS password updating using the new Password class. Please review and let me know your thoughts.

php/libraries/Utility.class.inc
```diff
@@ -933,5 +935,100 @@ class Utility
...
+     *
+     * @param \User     $user     The plain text password to be hashed and saved.
+     * @param \Password $password The plain text password to be hashed and saved.
+     * @param ?DateTime $expiry   When the password will expire. Can be set to a
+     *                            date in the past in order to create a one-time
+     *                            password.
+     *
+     * @return void
+     */
+    static function updatePassword(
```
xlecours: I think this belongs to the User class.

php/libraries/Utility.class.inc
```diff
@@ -933,5 +935,100 @@ class Utility
...
+     * Determines if the supplied password parameter is different than a User's
+     * current password (the one stored in the database).
+     *
+     * @param \User  $user              User whose password may have changed
+     * @param string $plaintextPassword The new desired password string entered
+     *                                  by the user.
+     *
+     * @return bool  true if the password has changed, false otherwise.
+     */
+    static function passwordChanged(
```
xlecours: The `!=` operator would replace that function.
something like `$user->getPassword() != (new \Password($value))`

php/libraries/Utility.class.inc
johnsaigle: See PR description and Dave's linked PR for why this was moved from the User class

php/libraries/Password.class.inc
```diff
@@ -0,0 +1,219 @@
...
+     *
+     * @param string $password The password to check for pwnage.
+     *
+     * @return bool Whether the password is known to be pwned.
+     */
+    function pwned(string $password): bool
+    {
+
+        // If the project is configured to disable the Pwned Password check,
+        // always return false (meanning that the password has not been pwned).
```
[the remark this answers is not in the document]
johnsaigle: That makes more sense

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -454,12 +454,27 @@ class Edit_User extends \NDB_Form
             $user    = \User::factory($this->identifier);
             $success = $user->update($set);
         }
-
         // Now set the password. Note that this field is named incorrectly
         // and represents a plaintext password, not a hash.
-        if (isset($values['Password_hash'])) {
-            $user->updatePassword($values['Password_hash']);
+        if (isset($values['Password_hash'], $values['Password_expiry'])) {
```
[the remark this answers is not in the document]
johnsaigle: The third param for updatePassword must be a `?\DateTime` so `$input_password_expiry` needs to be a DateTime object. However, doing `new DateTime(null);` gives today's date instead of throwing any kind of error.

php/libraries/Utility.class.inc
johnsaigle: In the PR description I linked another PR's of Dave's where he suggested removing all the password functionality from the User class in order to more accurately represent a user.

By the `!=` operator are you talking about the old `save()` function and the `User` class? I was always confused by what that did and where the value came from.

Also, even with those functions there was a `passwordChanged` function needed in both the Edit User and My Preferences code. 

I think it makes more sense to do away with it to simplify the User class as well as to reduce code reusage by replacing `save()` and the `passwordChanged` functions with this one.

johnsaigle:
@xlecours I also couldn't really figure out what `save()` was meant to do (or at least what it was doing differently) and found that it wasn't really necessary after doing other refactoring. 

In general I wanted to reuse (and not duplicate) code as much as possible and also try to remove password-related tasks from User. Removing `save()` helps with this.

php/libraries/Utility.class.inc
xlecours: @johnsaigle It is not related to the save function.

the `!=` operator could be used to compare two `Password` objects. 
```
$userpassword = $user->getPassword();
$newpassword = new \Password($value);
if ($userpassword != $newpassword) {
    echo "Those are different password.";
} else {
    echo "Those are the same password.";
}
```

php/libraries/Utility.class.inc
xlecours: From what I can see, @driusan is looking for a model/data-transfer-object(dto) representation of a User. Our current User class is a controller, to deal with CRUD stuff related to a user as well as a model for a user. By striping all business related logic, what remain should be close to a model. but we need to find a place to put all the logic. Moving it to Utility is not a sustainable solution.

php/libraries/Utility.class.inc
johnsaigle: The `__toString()` function of the `Password` class is the hash of the password, not the plaintext password. I chose to do this so that there's no risk of the plaintext getting exposed anywhere.

Also there isn't a `getPassword()` function.

php/libraries/Utility.class.inc
johnsaigle: I agree that Utility probably isn't the best solution long-term but I'm curious whether the functions should stay in User for now or not.

@driusan do you have an opinion?

php/libraries/Utility.class.inc
xlecours: @johnsaigle keeping the hash is great and do not prevent from comparing 2 password using `==`. 
But what will prevent it is the random salt used by password_hash.

johnsaigle:
`updatePassword` and `passwordChanged` have been moved back to User from Utility.

The tests are working now so this is ready for review.

johnsaigle:
@ridz1208 @driusan Thoughts on merging this now that it's approved and passing travis?

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -454,12 +454,25 @@ class Edit_User extends \NDB_Form
...
+            // If both hash and expiry are set, use both values when resetting
+            // the password. This occurs when a new password is generated
+            // by the system.
+            $user->updatePassword(
+                new \Password($values['Password_hash']),
+                new \DateTime($values['Password_expiry'])
+            );
+        } else if (isset($values['Password_hash'])) {
+            // If just the Password_hash value is present, no not specify a
```
driusan:
```suggestion
            // If just the Password_hash value is present, do not specify a
```

driusan:
Now that it's passing Travis..."This branch has conflicts that must be resolved" ...

johnsaigle:
Thanks @driusan. I rebased so hopefully everything is good to go now.

php/libraries/SinglePointLogin.class.inc
```diff
@@ -312,8 +261,28 @@ class SinglePointLogin
...
+                    $newHash = new \Password($password);
+                    \User::updatePassword(
+                        $newHash
+                    );
+                } catch (\InvalidArgumentException $e) {
+                    // Edge-case: if a user has a password based on old LORIS
+                    // password standards and it needs rehashing, it will
+                    // generate this exception when used above.
+                    // Rather than force a password change, we will manually
+                    // update the password here without using the Password class
```
driusan: I think it would make more sense to just always do it the way that works in all situations, but I'm not going to block over it either.

PR #4412, opened 2019-03-20
Make session CenterID and Visit_label not nullable

driusan:
These should always be populated, so this updates the schema
to enforce that they can not be null.

### DIFF
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -302,9 +302,9 @@ CREATE TABLE `candidate` (
 CREATE TABLE `session` (
   `ID` int(10) unsigned NOT NULL AUTO_INCREMENT,
   `CandID` int(6) NOT NULL DEFAULT '0',
-  `CenterID` integer unsigned DEFAULT NULL,
+  `CenterID` integer unsigned NOT NULL,
   `VisitNo` smallint(5) unsigned DEFAULT NULL,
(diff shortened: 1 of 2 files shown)


johnsaigle:
Seems like an integration test needs to be updated.

maltheism:
@driusan I believe travis is failing because of the type hinting changes in major and where the bugfix is implemented in #4358 PR.

driusan:
After some investigation, #4414 is blocking this.

maltheism:
@driusan #4358 fixes that if you don't have me on mute 😆 .

driusan:
I haven't tested #4358, but skimming the code it doesn't look like #4358 fixes it, it looks like it just uses a null coalesce to suppress the warning about it here: https://github.com/aces/Loris/pull/4358/files#diff-49cf0767b74741ffcf9dce30cdeead3fR145 .. it's still passing an invalid value to an argument that shouldn't be nullable, and the TimePoint::createNew signature is still wrong.

driusan:
After looking into it a bit with @maltheism #4358 does prevent it from getting inserted as null (in a way, it always gets inserted as Center 1 because of an already existing bug in `TimePoint::createNew` on the major branch when passed a null), but I'm still going to refactor the TimePoint code.

johnsaigle:
Is this still Blocked now that #4416 has been merged?

PR #4413, opened 2019-03-20
[SQL] Add instrument hash for Z-JSON instruments

zaliqarosli:
### Brief summary of changes

Instruments Tables (3)

`TestJsonID` will be a hash generated by the content in `JsonData` using SHA 256 (32-byte). Having a unique hash allows for version tracking and data integrity. If the JsonData is modified in the back end, an error will throw when being rendered on the front-end if hash and content don't match

`instrument_subtests` should be dropped later as Z-JSON contains page information.

### DIFF
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -225,16 +225,32 @@ CREATE TABLE `test_subgroups` (
 
 INSERT INTO test_subgroups (Subgroup_name) VALUES ('Instruments'),('Imaging');
 
+CREATE TABLE `instrument_schema` (
+  `InstrumentSchemaID` int(10) unsigned NOT NULL auto_increment,
+  `PreviousVersion` int(10) unsigned default NULL,
+  `UsersID` int(10) unsigned NOT NULL,
+  `SchemaHash` char(64) NOT NULL,
+  `DateUpdated` datetime NOT NULL DEFAULT NOW(),
+  `SchemaURI` TEXT NOT NULL,
+  `SchemaJSON` TEXT NOT NULL,
+  CONSTRAINT `PK_instrument_schema` PRIMARY KEY (`InstrumentSchemaID`),
+  CONSTRAINT `UK_instrument_schema_SchemaHash` UNIQUE KEY (`SchemaHash`),
+  CONSTRAINT `FK_instrument_Schema_PreviousVersion` FOREIGN KEY (`PreviousVersion`) REFERENCES `instrument_schema` (`InstrumentSchemaID`) ON DELETE RESTRICT ON UPDATE RESTRICT,
+  CONSTRAINT `FK_instrument_schema_UsersID` FOREIGN KEY (`UsersID`) REFERENCES `users` (`ID`) ON DELETE RESTRICT ON UPDATE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
 CREATE TABLE `test_names` (
   `ID` int(10) unsigned NOT NULL auto_increment,
   `Test_name` varchar(255) default NULL,
   `Full_name` varchar(255) default NULL,
   `Sub_group` int(11) unsigned default NULL,
+  `InstrumentSchemaID` int(10) unsigned default NULL,
   `IsDirectEntry` boolean default NULL,
   PRIMARY KEY  (`ID`),
   UNIQUE KEY `Test_name` (`Test_name`),
   KEY `FK_test_names_1` (`Sub_group`),
-  CONSTRAINT `FK_test_names_1` FOREIGN KEY (`Sub_group`) REFERENCES `test_subgroups` (`ID`)
+  CONSTRAINT `FK_test_names_1` FOREIGN KEY (`Sub_group`) REFERENCES `test_subgroups` (`ID`),
+  CONSTRAINT `FK_test_names_InstrumentSchemaID` FOREIGN KEY (`InstrumentSchemaID`) REFERENCES `instrument_schema` (`InstrumentSchemaID`) ON DELETE RESTRICT ON UPDATE RESTRICT
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
 CREATE TABLE `instrument_subtests` (
@@ -262,6 +278,7 @@ CREATE TABLE `flag` (
   `UserID` varchar(255) default NULL,
   `Testdate` timestamp NOT NULL default CURRENT_TIMESTAMP on update CURRENT_TIMESTAMP,
   `Data` TEXT default NULL,
+  `InstrumentSchemaID` int(10) unsigned default NULL,
   PRIMARY KEY  (`CommentID`),
   KEY `Status` (`Flag_status`),
   KEY `flag_ID` (`ID`),
@@ -273,7 +290,8 @@ CREATE TABLE `flag` (
   KEY `flag_Administration` (`Administration`),
   KEY `flag_UserID` (`UserID`),
   CONSTRAINT `FK_flag_1` FOREIGN KEY (`SessionID`) REFERENCES `session` (`ID`) ON DELETE CASCADE ON UPDATE CASCADE,
-  CONSTRAINT `FK_flag_2` FOREIGN KEY (`Test_name`) REFERENCES `test_names` (`Test_name`)
+  CONSTRAINT `FK_flag_2` FOREIGN KEY (`Test_name`) REFERENCES `test_names` (`Test_name`),
+  CONSTRAINT `FK_flag_InstrumentSchemaID` FOREIGN KEY (`InstrumentSchemaID`) REFERENCES `instrument_schema` (`InstrumentSchemaID`) ON DELETE RESTRICT ON UPDATE RESTRICT
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
 CREATE TABLE `history` (
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -244,13 +244,13 @@ CREATE TABLE `test_names` (
   `Test_name` varchar(255) default NULL,
   `Full_name` varchar(255) default NULL,
   `Sub_group` int(11) unsigned default NULL,
-  `InstrumentSchemaID` int(10) unsigned default NULL,
+  `LatestInstrumentSchemaID` int(10) unsigned UNIQUE default NULL,
   `IsDirectEntry` boolean default NULL,
   PRIMARY KEY  (`ID`),
   UNIQUE KEY `Test_name` (`Test_name`),
   KEY `FK_test_names_1` (`Sub_group`),
   CONSTRAINT `FK_test_names_1` FOREIGN KEY (`Sub_group`) REFERENCES `test_subgroups` (`ID`),
-  CONSTRAINT `FK_test_names_InstrumentSchemaID` FOREIGN KEY (`InstrumentSchemaID`) REFERENCES `instrument_schema` (`InstrumentSchemaID`) ON DELETE RESTRICT ON UPDATE RESTRICT
+  CONSTRAINT `FK_test_names_LatestInstrumentSchemaID` FOREIGN KEY (`LatestInstrumentSchemaID`) REFERENCES `instrument_schema` (`InstrumentSchemaID`) ON DELETE RESTRICT ON UPDATE RESTRICT
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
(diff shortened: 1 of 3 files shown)


johnsaigle:
Not sure why Travis is failing. Running your patch didn't cause any errors for me.

zaliqarosli:
if someone could help me avoid circular foreign key references, please let me know

zaliqarosli:
@johnsaigle patch works, schema doesn't because of FK referencing a table that doesn't exist yet. thanks for looking

SQL/0000-00-00-schema.sql
```diff
@@ -368,6 +370,24 @@ CREATE TABLE `test_names` (
   CONSTRAINT `FK_test_names_1` FOREIGN KEY (`Sub_group`) REFERENCES `test_subgroups` (`ID`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
+CREATE TABLE `test_json` (
+  `TestJsonID` varchar(64) NOT NULL,
+  `JsonData` TEXT NOT NULL,
```
ridz1208: shouldnt this be `JSONSchema` ?

SQL/0000-00-00-schema.sql
```diff
@@ -368,6 +370,24 @@ CREATE TABLE `test_names` (
   CONSTRAINT `FK_test_names_1` FOREIGN KEY (`Sub_group`) REFERENCES `test_subgroups` (`ID`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
+CREATE TABLE `test_json` (
```
ridz1208: @driusan should we get rid of `test` for instruments and just call it... `instrument_JSON`

SQL/0000-00-00-schema.sql
```diff
@@ -368,6 +370,24 @@ CREATE TABLE `test_names` (
...
+  `TestJsonID` varchar(64) NOT NULL,
+  `JsonData` TEXT NOT NULL,
+  CONSTRAINT `PK_test_json` PRIMARY KEY (`TestJsonID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE TABLE `test_names_json_rel` (
```
ridz1208: given our current convention this table name does not follow the `table1_table2_rel` name

SQL/0000-00-00-schema.sql
```diff
@@ -368,6 +370,24 @@ CREATE TABLE `test_names` (
...
+  `TestNamesID` int(10) unsigned NOT NULL,
+  `TestJsonID` varchar(64) NOT NULL,
+  `DateUpdated` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `UsersID` int(10) unsigned NOT NULL,
+  `Current` BIT NOT NULL,
```
ridz1208: just to make it clearer maybe `currentlyActive` is more dsciptive

ridz1208:
@zaliqarosli Looking at this... I'm thinking we could include the fields of both these tables in one single table. we dont necessarily need the `rel` since we only have a one to many relationship between test_names and test_json.

unless I'm missing something ?

PapillonMcGill:
@zaliqarosli 
No circular reference in this schema. Usage of Rel table is Ok.
I would change the name of Test_Name_Json_Rel to something else like maybe Test_Name_Detail and reverse the FK relation with the Test_Json table.
Maybe we can have a quick talk about the purpose of these table.
@ridz1208  Having a `text` field that can grow quite large in a table by itself with a 1:1 mapping is Ok.

zaliqarosli:
> I would change the name of Test_Name_Json_Rel to something else like maybe Test_Name_Detail and reverse the FK relation with the Test_Json table

I don't think I agree with this. Maybe I can explain how and why the tables are meant to be related offline

SQL/0000-00-00-schema.sql
zaliqarosli: sure

SQL/0000-00-00-schema.sql
zaliqarosli: i had thought about calling it instrument_JSON but seeing as it relates to test_names and belongs in the same group of tables, I wanted to keep uniform the naming

SQL/0000-00-00-schema.sql
zaliqarosli: no it does not .. but I think it's an acceptable short form because `test_names_test_json_rel` seems redundant so i removed the repeat of `test`. this can easily be changed to properly conform to the convention

SQL/0000-00-00-schema.sql
zaliqarosli: yes sir

zaliqarosli: @ridz1208 what about just `Active`?

zaliqarosli:
is this what you guys are thinking? @PapillonMcGill @ridz1208 
Instruments Tables (4)

PapillonMcGill:
This is better.
What size (in bytes or KB) a typical value would be for the `JsonSchema` attributes?

zaliqarosli:
@PapillonMcGill we need the one-to-many relation between test_name and instrument_json though. removing TestNameID will lose this relation

zaliqarosli:
maybe this makes most sense @PapillonMcGill 

Instruments Tables (5)

zaliqarosli:
updated proposal: 
Screenshot 2019-05-17 16 01 44

zaliqarosli:
@PapillonMcGill i changed utf8 to utf8mb4 and it fails travis. do you happen to know why?

PapillonMcGill:
@zaliqarosli It might be the `FK_test_names_InstrumentSchemaID` constraint name that fail with utf8mb4.  Constraint name are limited to 64 chars but with utf8mb4, per experience, it is somewhere around 21 (or 27). To know for sure, I suggest running the 0000-00-00schema.sql script on your VM and see what is the failure code.  Travis only report failure in that file but not the exact failure code.

zaliqarosli:
@PapillonMcGill i checked again and travis does indeed tell you the error code! `ERROR 1071 (42000) at line 373: Specified key was too long; max key length is 767 bytes` so i guess it is the length of the FK name

SQL/New_patches/2019-03-20_add_instrument_hash.sql
```diff
@@ -0,0 +1,20 @@
...
+  `PreviousVersion` int(10) unsigned default NULL,
+  `UsersID` int(10) unsigned NOT NULL,
+  `SchemaHash` char(64) NOT NULL,
+  `DateUpdated` datetime NOT NULL DEFAULT NOW(),
+  `SchemaJson` TEXT NOT NULL,
+  CONSTRAINT `PK_instrument_schema` PRIMARY KEY (`InstrumentSchemaID`),
+  CONSTRAINT `UK_instrument_schema_SchemaHash` UNIQUE KEY (`SchemaHash`),
+  CONSTRAINT `FK_instrument_Schema_PreviousVersion` FOREIGN KEY (`PreviousVersion`) REFERENCES `instrument_schema` (`InstrumentSchemaID`) ON DELETE RESTRICT ON UPDATE RESTRICT,
+  CONSTRAINT `FK_instrument_schema_UsersID` FOREIGN KEY (`UsersID`) REFERENCES `users` (`ID`) ON DELETE RESTRICT ON UPDATE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```
kongtiaowang: utf8

SQL/0000-00-00-schema.sql
```diff
@@ -404,8 +421,9 @@ CREATE TABLE `flag` (
   KEY `flag_Administration` (`Administration`),
   KEY `flag_UserID` (`UserID`),
   CONSTRAINT `FK_flag_1` FOREIGN KEY (`SessionID`) REFERENCES `session` (`ID`) ON DELETE CASCADE ON UPDATE CASCADE,
-  CONSTRAINT `FK_flag_2` FOREIGN KEY (`Test_name`) REFERENCES `test_names` (`Test_name`)
-) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+  CONSTRAINT `FK_flag_2` FOREIGN KEY (`Test_name`) REFERENCES `test_names` (`Test_name`),
+  CONSTRAINT `FK_flag_InstrumentSchemaID` FOREIGN KEY (`InstrumentSchemaID`) REFERENCES `instrument_schema` (`InstrumentSchemaID`) ON DELETE RESTRICT ON UPDATE RESTRICT
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```
kongtiaowang: use utf8 instead of utf8mb4.

SQL/New_patches/2019-03-20_add_instrument_hash.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE `instrument_schema` (
```
ridz1208: I think we should add the actual schema as well... the case being that we cannot stop all data-entry in the event that the server is offline, unreachable or corrupted. plus it would avoid an external call for each instrument load/save/build...

zaliqarosli: sure, that's shouldn't be a problem as long as we store only the final, complete, and validated schema

SQL/New_patches/2019-03-20_add_instrument_hash.sql
```diff
@@ -0,0 +1,20 @@
...
+  `DateUpdated` datetime NOT NULL DEFAULT NOW(),
+  `SchemaURI` TEXT NOT NULL,
+  CONSTRAINT `PK_instrument_schema` PRIMARY KEY (`InstrumentSchemaID`),
+  CONSTRAINT `UK_instrument_schema_SchemaHash` UNIQUE KEY (`SchemaHash`),
+  CONSTRAINT `FK_instrument_Schema_PreviousVersion` FOREIGN KEY (`PreviousVersion`) REFERENCES `instrument_schema` (`InstrumentSchemaID`) ON DELETE RESTRICT ON UPDATE RESTRICT,
+  CONSTRAINT `FK_instrument_schema_UsersID` FOREIGN KEY (`UsersID`) REFERENCES `users` (`ID`) ON DELETE RESTRICT ON UPDATE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+ALTER TABLE `test_names`
+  ADD COLUMN `InstrumentSchemaID` int(10) unsigned default NULL,
```
PapillonMcGill:
```suggestion
  ADD COLUMN `InstrumentSchemaID` int(10) unsigned default NULL after Sub_group,
```
to keep the same ordering.

SQL/New_patches/2019-03-20_add_instrument_hash.sql
```diff
@@ -0,0 +1,20 @@
...
+  ADD CONSTRAINT `FK_test_names_InstrumentSchemaID` FOREIGN KEY (`InstrumentSchemaID`) REFERENCES `instrument_schema` (`InstrumentSchemaID`) ON DELETE RESTRICT ON UPDATE RESTRICT;
+
+ALTER TABLE `flag`
+  ADD COLUMN `InstrumentSchemaID` int(10) unsigned default NULL,
```
PapillonMcGill:
```suggestion
  ADD COLUMN `InstrumentSchemaID` int(10) unsigned default NULL after Data,
```

ridz1208:
@zaliqarosli the changes are good, I just want to put this here, it might be up for discussion later on.

While a user is building an instrument, does the partial instrument get saved in this table, or does it only get inserted into the table once it's published to the central instrument repository  cause technically thats when the validation is done. can an instrument be saved in an "invalid" form ? can it even exist in an invalid form since they are generated by the instrument builder ?


anyways food for thought

SQL/0000-00-00-schema.sql
```diff
@@ -225,16 +225,32 @@ CREATE TABLE `test_subgroups` (
 
 INSERT INTO test_subgroups (Subgroup_name) VALUES ('Instruments'),('Imaging');
 
+CREATE TABLE `instrument_schema` (
+  `InstrumentSchemaID` int(10) unsigned NOT NULL auto_increment,
+  `PreviousVersion` int(10) unsigned default NULL,
+  `UsersID` int(10) unsigned NOT NULL,
+  `SchemaHash` char(64) NOT NULL,
```
driusan: What hash is this hashed with? It should have the hashing algorithm in the name so that it can be interpreted.

driusan:
This went in a different direction and Z-JSON is being replaced by a JSON-LD based format, so closing this old PR..

PR #4415, opened 2019-03-21

maltheism:
### Brief summary of changes

### This resolves issue...

- [x] Redmine? #16008

### To test this change...


modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,283 @@
...
+        /**
+         * User permission verification.
+         */
+        $user       = $request->getAttribute('user');
+        $identifier = $requestBody['identifier'] ?? '';
+        $candidate  =& \Candidate::singleton($identifier);
+        if (!$user->hasPermission('data_entry')
+            && !(in_array(
+                $candidate->getData('RegistrationCenterID'),
+                $user->getData('CenterIDs')
```
johnsaigle: Please add `true` as a third parameter to `in_array`

modules/create_timepoint/php/create_timepoint.class.inc
```diff
@@ -22,31 +22,24 @@ namespace LORIS\create_timepoint;
...
      *
      * @param \User $user The user whose access is being checked
      *
      * @return bool
+     * @throws \LorisException
      */
     function _hasAccess(\User $user) : bool
     {
-        $candidate =& \Candidate::singleton($this->identifier);
+        $candidate = \Candidate::singleton($this->identifier);
```
johnsaigle: Just since you're editing this file anyway can you also add the `true` param to in_array below?

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,283 @@
+<?php
```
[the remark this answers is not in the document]
maltheism: @johnsaigle You mean `<?php declare(strict_types=1);` I believe and thanks i'll add!

johnsaigle:
Needs work on the integration tests

maltheism:
Blocking this PR until #4416 is merged.

modules/create_timepoint/php/create_timepoint.class.inc
maltheism: @johnsaigle when I add true I'm unable to access the page when using my admin account. Any idea why? I'll try debugging in the meantime.

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -115,11 +119,12 @@ class AjaxTimepoint extends \NDB_Page
     {
         // Verify cand ID.
         $values['identifier'] = isset($values['identifier']) ?
-            htmlentities(
+            intval(htmlentities(
```
johnsaigle: You can just use `intval()` here instead of wrapping `htmlentities` because it'll only ever give you an integer.

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -0,0 +1,356 @@
+import React from 'react';
+import Panel from 'Panel';
+import Loader from 'Loader';
+import PropTypes from 'prop-types';
```
zaliqarosli: i would group react and proptypes together, then panel and loader together to make it clear what's a react package and what's a loris component

php/libraries/Candidate.class.inc
```diff
@@ -630,43 +630,17 @@ class Candidate
...
         return intval($result);
     }
 
-    /**
-     * Returns list of consents and their respective statuses for this candidate
-     *
-     * @return array List of consents and their associated values for this candidate
-     *               The keys of the arrays are the IDs of the consents
-     */
-    public function getConsents(): array
```
zaliqarosli: why is the candidate class being touched and why is this entire function getting deleted? I'm pretty sure it's being used somewhere

zaliqarosli:
on testing: 

Screenshot 2019-03-25 17 11 34

error messages show up on first render, but should only be enabled on submit and validation

php/libraries/Candidate.class.inc
maltheism: @zaliqarosli thanks, not sure how that happened.

maltheism:
Fixed the above, removing needs work tag.

johnsaigle:
Integration tests still appear to need work before this can be merged.

zaliqarosli:
error: on clicking 'Create Time Point' button 

Screenshot 2019-04-01 13 52 15

maltheism:
Thanks @zaliqarosli, can you test again. I fixed that error I believe.

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -0,0 +1,366 @@
...
+  }
+
+  /**
+   * Handle form submission
+   * @param {object} e - Form submission event
+   */
+  handleSubmit(e) {
+    const state = Object.assign({}, this.state);
+    state.form.options.required = {
+      subproject: true,
```
zaliqarosli: i think this should be done in render() and not handleSubmit() because a post request is still being sent here even if fields are empty.
Screenshot 2019-04-01 14 32 20

I think a better solution to requiring fields is to set their required prop to true but turn off the error message. the html form can then throw the required error on the browser itself, i.e. ..

Screenshot 2019-04-01 14 43 14

maltheism: Okay i'll fix this tomorrow, have to go but note for me from @zaliqarosli is to set the values to null and not empty string.

maltheism:
@zaliqarosli I applied the correct fix as you suggested. The form uses the default browser functionality for missing fields now.

zaliqarosli:
@maltheism looks good! on major, only the valid subprojects show up as options. on this branch, all the subprojects show up. any chance you could make these options dynamically available like the time point options?

maltheism:
Hi @zaliqarosli, I added the update to show valid subprojects. Please let me know if something is wrong with it.

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -122,9 +122,13 @@ class CreateTimepoint extends React.Component {
...
       .then(
         (data) => {
+          console.log(data);
           // Populate the form errors.
-          if (data.errors && data.errors.length > 0) {
+          if (data.errors) {
+            // data for the form errors.
             this.setState({errors: data.errors});
+            this.setState({isLoaded: true});
+            return;
```
zaliqarosli: why these changes?

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -122,9 +122,13 @@ class CreateTimepoint extends React.Component {
     ).then((response) => response.json())
       .then(
         (data) => {
+          console.log(data);
```
zaliqarosli: console.error is better

modules/create_timepoint/jsx/createTimepointIndex.js
zaliqarosli: nevermind, just realized these were removed after

modules/create_timepoint/jsx/createTimepointIndex.js
zaliqarosli: nevermind, just realized these were removed after

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -122,13 +123,12 @@ class CreateTimepoint extends React.Component {
...
-          console.log(data);
           // Populate the form errors.
           if (data.errors) {
-            // data for the form errors.
-            this.setState({errors: data.errors});
-            this.setState({isLoaded: true});
-            return;
+            const state = Object.assign({}, this.state);
+            state.errors = data.errors;
+            state.form.disabled = true;
```
zaliqarosli: could you explain what you're doing here with disabling the button? I can't trigger this case on the front-end

maltheism: Thanks @zaliqarosli, that's code that isn't needed anymore since your help with the unfilled form errors.

jsx/Form.js
```diff
@@ -1496,6 +1496,7 @@ class ButtonElement extends Component {
             type={this.props.type}
             className={this.props.buttonClass}
             onClick={this.handleClick}
+            disabled={this.props.disable}
```
zaliqarosli: while i do like this feature, i don't think it should be introduced in this PR

modules/create_timepoint/js/.gitignore
```diff
@@ -0,0 +1 @@
+*.js
```
zaliqarosli: i think we're standardizing the .gitignore to be in the module directory with js/* in it

zaliqarosli: reference https://github.com/aces/Loris/pull/4426

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,300 @@
...
+         * Parse POST request body.
+         */
+        $requestBody = $request->getParsedBody();
+
+        /**
+         * User permission verification.
+         */
+        $user = $request->getAttribute('user');
+        if (!isset($requestBody['identifier'])) {
+            throw new LorisException(
```
johnsaigle: I think this could be a `404` or `400` instead.

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,300 @@
...
+                    = array_filter($labelOptions);
+            }
+        }
+        $values['visit'] = $visit_options;
+        if (!empty($visit_options)) {
+            // List of sites for the user.
+            $user = \User::singleton();
+            $DB   = \Database::singleton();
+            $user_list_of_sites = $user->getData('CenterIDs');
+            $num_sites          = count($user->getData('CenterIDs'));
```
johnsaigle:
```suggestion
            $num_sites          = count($user_list_of_sites);
```

maltheism:
Thanks @zaliqarosli and @johnsaigle. I made added the suggestions.

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    /**
+     * Sanitize the $postResponse data.
+     *
+     * @param array $values that have been sanitized.
+     *
+     * @return array
+     */
+    function sanitizer(array $values) : array
```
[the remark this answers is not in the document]
maltheism: @johnsaigle I'm a little confused how security has dibs on the word sanitizer. I typically use that word when writing a method that confirms if the receiving data is actually valid; which the function her is checking if anything is set and or make it an int.

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,300 @@
...
+                    $values['visit'],
+                    $site
+                );
+                $response['status'] = 'success';
+            } else {
+                $response['status'] = 'error';
+                $response['errors'] = $errors;
+            }
+        } else {
+            $response['status'] = 'error';
```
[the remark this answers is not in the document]
maltheism: @johnsaigle the error message is on line 166 and is returned to the frontend in $response['errors'].

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: @johnsaigle thanks will fix.

modules/create_timepoint/php/ajaxtimepoint.class.inc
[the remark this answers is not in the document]
johnsaigle: This function was originally doing HTML-escaping of data that would be printed to the front-end. This is done in order to prevent from JavaScript injection attacks. That was the meaning behind the `$sanitizedIdentifier` in the old code. It's not that security has dibs but that you've changed what this function is doing from previous commits.

The discussion about the HTML-escaping not being necessary anymore seems to have been lost with all the commits, but that's the context I'm coming at this from.

It's misleading to me to still have this function given that we do use the word sanitize to mean security-checks in an informal context.

Even without that though, you don't need this function. You're already using `identifer` without any check on line 78. Also both your `validate()` and `initializeSetup` do different ways of validating `psc` so you are duplicating your validation code a few times within this file. 

So given that it's potentially misleading and also not really used, I think it makes sense to delete this function.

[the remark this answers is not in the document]
maltheism: @johnsaigle Thanks for the in-depth writeup. Yah when I look at the code it seems weird that I didn't place the function call of validating the response data before the permissions check in any case you're right that the function seems pointless. I'll just make some tweaks to storing the identifier as an int when I do the permissions check and delete the function. Hopefully nothing breaks.

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -0,0 +1,362 @@
...
+            const state = Object.assign({}, this.state);
+            state.storage.visit = data.visit;
+            this.setState(state);
+            // Handle visit selection.
+            this.handleVisitLabel();
+          }
+          // Display form to user.
+          this.setState({isLoaded: true});
+        }).catch((error) => {
+      this.populateErrors({message: 'Server error.'});
```
zaliqarosli: i think you can remove the call to this function now

maltheism: @zaliqarosli I think it's good practice to have an error to display to the user by the rare chance the server responds with an error. Even if that case should only arise if the server or database goes down at the very moment a post is being sent. Thoughts?

maltheism: @zaliqarosli, I made a change that should resolved it.

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -118,44 +118,50 @@ class CreateTimepoint extends React.Component {
...
+          this.setState(state);
+          // Handle visit selection.
+          this.handleVisitLabel();
+        }
+        // Display form to user.
+        this.setState({isLoaded: true});
+      } else {
+        response.text().then((message) => {
+          const state = Object.assign({}, this.state);
+          state.errors = [{error: message}];
```
zaliqarosli: should this just be `state.errors = message`?

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -118,44 +118,50 @@ class CreateTimepoint extends React.Component {
...
+          this.setState(state);
+        });
+      }
         }).catch((error) => {
```
zaliqarosli: i think this line's indentation needs to be moved to the left a few

modules/create_timepoint/jsx/createTimepointIndex.js
zaliqarosli: will the message get rendered if it is only stored into state.error here?

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -118,44 +118,50 @@ class CreateTimepoint extends React.Component {
...
-      this.populateErrors({message: 'Server error.'});
+          console.log(error);
```
zaliqarosli: console.error is better!

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -118,50 +118,46 @@ class CreateTimepoint extends React.Component {
...
+            state.storage.visit = data.visit;
+            this.setState(state);
+            // Handle visit selection.
+            this.handleVisitLabel();
+          }
+          // Display form to user.
+          this.setState({isLoaded: true});
         }).catch((error) => {
-          console.log(error);
+      const state = Object.assign({}, this.state);
```
zaliqarosli: ahh! im confused. you changed it back!

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -118,50 +118,46 @@ class CreateTimepoint extends React.Component {
...
+      state.errors = [{message: error}];
```
zaliqarosli: this is still not right i don't think. should be `state.errors = error`. or error.message even? i don't remember what the error object looks like

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -118,50 +118,46 @@ class CreateTimepoint extends React.Component {
...
+      this.setState(state);
```
zaliqarosli: the error is going into state but does it get rendered? should at least have console.error

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -118,50 +118,46 @@ class CreateTimepoint extends React.Component {
...
       this.setState({isLoaded: true});
```
zaliqarosli: you're setting state twice here back to back, could do it only once

modules/create_timepoint/jsx/createTimepointIndex.js
maltheism: @zaliqarosli state.errors is an array for holding errors by the format of a json object and the rendering handles it as an array of json objects.

modules/create_timepoint/jsx/createTimepointIndex.js
maltheism: @zaliqarosli The indentation is actually correct how it is. The .then().then().. makes this unpleasant on the eyes but it's actually correct.

modules/create_timepoint/jsx/createTimepointIndex.js
maltheism: Yah I examined it for a bit and think it's better this way.

modules/create_timepoint/jsx/createTimepointIndex.js
maltheism: @zaliqarosli see the above comment and we can chat offline if you want.

modules/create_timepoint/jsx/createTimepointIndex.js
maltheism: The error will be displayed to the user. See line 262 and 313. It's how all the errors are shown to the user.

kongtiaowang:
#4474
After this PR, you need to rebase your PR to pass Travis.

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -0,0 +1,364 @@
...
+                {visit}
+                <ButtonElement
+                  label={'Create Time Point'}
+                  type={'submit'}
+                  name={'fire_away'}
+                />
+              </FormElement>
+            </Panel>
+          </div>
+        </div>
```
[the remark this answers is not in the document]
maltheism: @HenriRabalais thanks, I fixed it but seems to not be showing as outdated on Github?

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -0,0 +1,364 @@
...
+    } else {
+      return (
+        <div>
+          <div>
+            <h3>New time point successfully registered.</h3>
+            <a href={'/' + this.state.url.params.candID}>
+              Click here to continue.
+            </a>
+          </div>
+        </div>
```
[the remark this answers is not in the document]
maltheism: @HenriRabalais thanks, I fixed it but seems to not be showing as outdated on Github?

modules/create_timepoint/jsx/createTimepointIndex.js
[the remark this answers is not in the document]
HenriRabalais: It disappeared for me!

maltheism:
Hi @kongtiaowang, today I had to resolve conflicting files that were merged and now some of the tests are failing in travis. I need help because after manually testing the module everything seems to be working.

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,290 @@
...
+
+        // All subprojects from config file (error).
+        if (empty($allSubprojects)) {
+            $errors['subprojectID'] = 'No subprojects have been defined 
+            for this study. If you are an administrator, please use the 
+            Configuration module to add new subprojects.';
+        }
+
+        // List of valid subprojects for a given project
+        if ($config->getSetting('useProjects') === 'true') {
```
kongtiaowang: we don't use 'useProjects' anymore.

kongtiaowang: https://github.com/aces/Loris/pull/4496/

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,284 @@
...
+                    $labelOptions[$item['@']['value']] = $item['#'];
+                }
+                $visit_options[$visitLabel['@']['subprojectID']]
+                    = array_filter($labelOptions);
+            }
+        }
+        $values['visit'] = $visit_options;
+        if (!empty($visit_options)) {
+            // List of sites for the user.
+            $user = \User::singleton();
```
xlecours: This whole block could be replaced by:
```
$values['psc'] = array_intersect(\Utility::getSiteList(), \User::singleton()->getSiteNames());
```

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,284 @@
...
+
+    /**
+     * Initialize setup, the extra values for the
+     * create timepoint form. (psc & errors)
+     *
+     * @param array $values the user data received.
+     *
+     * @return array
+     */
+    function initializeSetup(array $values) : array
```
xlecours: suggestion :
`private function _handleGET(ServerRequestInterface $request): ResponseInteface`

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,284 @@
...
+     * Processes the values and saves to database
+     *
+     * @param array $values the user data received.
+     *
+     * @return array
+     */
+    function processRequest(array $values) : array
+    {
+        if (isset($values['command']) && $values['command'] == 'initialize') {
+            $response = $this->initializeSetup($values);
```
xlecours: This should not be a POST. If the goal is to get values from the server then GET should be used.

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,284 @@
...
+
+        // List of valid subprojects for a given project
+        $subprojList = $candidate->getValidSubprojects();
+        if (empty($subprojList)) {
+            $this->form->errors['subprojectID'] = "No subprojects have been 
+            defined for the project this candidate is affiliated with. 
+            If you are an administrator, please use the Configuration module to 
+            add new subprojects and associate them with projects.";
+        }
+        $tmp = array();
```
xlecours: `$values['subproject'] = array_intersect_key(\Utility::getSubprojectList(), $candidate->getValidSubprojects());`

maltheism:
> Here are some change requests.
> General comment: I think that `GET ...create_timepoint/AjaxTimepoint` should be used to access initial values for the drop down.

Hey @xlecours The reason I'm using a POST is because the method contains data that's being sent (ex. 'initialize' or 'create') for retrieval and GET contains no data with the request.

xlecours:
> > Here are some change requests.
> > General comment: I think that `GET ...create_timepoint/AjaxTimepoint` should be used to access initial values for the drop down.
> 
> Hey @xlecours The reason I'm using a POST is because the method contains data that's being sent (ex. 'initialize' or 'create') for retrieval and GET contains no data with the request.

@maltheism But you could assume that a GET request is for 'initialize' and POST for 'create'.

maltheism:
> > > Here are some change requests.
> > > General comment: I think that `GET ...create_timepoint/AjaxTimepoint` should be used to access initial values for the drop down.
> > 
> > 
> > Hey @xlecours The reason I'm using a POST is because the method contains data that's being sent (ex. 'initialize' or 'create') for retrieval and GET contains no data with the request.
> 
> @maltheism But you could assume that a GET request is for 'initialize' and POST for 'create'.

@xlecours Okay, I'll refactor it to do that once I fix the PR. Since it seems to be broken as of today.

maltheism:
Hey @xlecours, I added the changes you suggested. 🙂

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,285 @@
...
+            case 'Forbidden':
+                return new \LORIS\Http\Response\Forbidden(
+                    'Forbidden'
+                );
+            default:
+                break;
+            }
+        }
+
+        if (isset($values['command']) && $values['command'] == 'create') {
```
xlecours: the command parameter is not necessary. A POST request to this endpoint means that a timepoint should be created with the given data.

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,285 @@
...
+
+    /**
+     * Permission verification.
+     *
+     * @param ServerRequestInterface $request The incoming PSR7 request.
+     * @param array                  $values  The form values.
+     *
+     * @return array
+     */
+    private function verifyPermissions(
```
xlecours: By convention, there should be `private function _hasAccess(\User $user): bool` that only check for permission.
Then, in the handlePOST, check for mandatory fields and return a BadRequest when one is missing.
Then, instantiate the candidate and compare its RegistrationCenterId to the user's sites. 

In it's current state, the `verifyPermissions` function is doing all this and have a side effect of modifying the values

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,285 @@
...
+        }
+
+        if (isset($values['command']) && $values['command'] == 'create') {
+            $errors = $this->validate($values);
+            // convert site entered to a \Site object
+            $site = null;
+            $user = \User::singleton();
+            $user_list_of_sites = $user->getCenterIDs();
+            $num_sites          = count($user_list_of_sites);
+            if ($num_sites == 1) {
```
xlecours: Could you make the React app always send the Site even when the user as only one? Maybe just don't display the fields when it's the case.
We would then be able to avoid this "auto-select" feature in the backend.

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,285 @@
...
+                return new \LORIS\Http\Response\Forbidden(
+                    'Forbidden'
+                );
+            default:
+                break;
+            }
+        }
+
+        if (isset($values['command']) && $values['command'] == 'create') {
+            $errors = $this->validate($values);
```
xlecours: if there is an error, `return new \LORIS\Http\Response\BadRequest` right away. 
Doing this prevent from having to do special `if` statement like at line#193

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,285 @@
...
+                );
+                $response['status'] = 'success';
+            } else {
+                $response['status'] = 'error';
+                $response['errors'] = $errors;
+            }
+        } else {
+            $response['status'] = 'error';
+        }
+        return new \LORIS\Http\Response\JsonResponse(
```
xlecours: (Comment/ not a formal change request) :: It would be nice if this was a 201 with the location header equal to baseurl + /candid so that the React app just need to send `window.lcoation($url)` instead of the success page with the link to click.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -0,0 +1,359 @@
...
+  /**
+   * Retrieve data from the provided URL and save it in state.
+   */
+  fetchInitializerData() {
+    const state = Object.assign({}, this.state);
+    let url = new URL(this.props.dataURL);
+    const params = {
+      candID: state.url.params.candID,
+      identifier: state.url.params.identifier,
+      subprojectID: state.form.subproject,
```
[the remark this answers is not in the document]
maltheism: @xlecours I think you're right. I cannot remember why I included it for the request.

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: @xlecours I think the error message displayed on the frontend might be more helpful than the BadRequest?

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: @xlecours this function is using the code before I modified the module for react. The $candidate is made from the $values['identifier] and that's used in the hasPermission('data_entry') check. Do you want me to move the validating the identifier sent from the fronted out of this function?

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: @xlecours you want me to skip the "New time point successfully registered" and just redirect to the /candID link automatically? I can do that by just checking with the candID gets updated.

modules/create_timepoint/php/ajaxtimepoint.class.inc
xlecours: @maltheism yes, I would like that.

modules/create_timepoint/php/ajaxtimepoint.class.inc
xlecours: @maltheism I the POST request body, there should be a site.

modules/create_timepoint/php/ajaxtimepoint.class.inc
xlecours: you can add any message to a BadResquest response.

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: @xlecours you're right. I'm still confused on what you want me to do because won't I still need to use $site = \Site::singleton($user_list_of_sites[0]); for retrieving the site object that's used for creating the timepoint?

xlecours: `$site = \Site::singleton(intval($values['psc']));` would be enough. Because `$values` will always have a `psc` otherwise a BadRequest would have been sent earlier.

the only scenario where this would fail is if the user is only part of a study site. The array intersect at line#136. in the handleGET, would not send a psc to the front end because Utility::getSiteList would remove the DCC sites from the list. Which would create a weird behavior with `$site = \Site::singleton($user_list_of_sites[0]);` (I would register the session as part of the study_site)

xlecours:
Regarding Travis,
There were 2 errors:
```
1) CreateTimepointTestIntegrationTest::testCreateTimepoint
UnknownServerException: input is null
System info: host: '274ed7ad209e', ip: '172.18.0.2', os.name: 'Linux', os.arch: 'amd64', os.version: '4.15.0-1028-gcp', java.version: '1.8.0_91'
/app/modules/create_timepoint/test/create_timepointTest.php:104
/app/modules/create_timepoint/test/create_timepointTest.php:79
```
It is a Javascript error from line 104 of create_timepointTest.php. The querySelector returns nothing for `subproject, select`. Try changing the value of `static $subProject` to `'select[name="subproject"]'`. It can be tested in the browser console `document.querySelector('select[name="subproject"]')`

```
2) CreateTimepointTestIntegrationTest::testCreateTimepointErrorEmptySubproject
NoSuchElementException: Unable to locate element: {"method":"name","selector":"fire_away"}
System info: host: '274ed7ad209e', ip: '172.18.0.2', os.name: 'Linux', os.arch: 'amd64', os.version: '4.15.0-1028-gcp', java.version: '1.8.0_91'
/app/vendor/facebook/webdriver/lib/WebDriverExceptions.php:67
/app/vendor/facebook/webdriver/lib/remote/RemoteWebDriver.php:133
/app/modules/create_timepoint/test/create_timepointTest.php:133
```
Since it's now a React app, the button is rendered after the page load. The current WebdriverBy call at line#133 should be change to execute script.
```
document.querySelector('button[name="fire_away"]').click();
```

maltheism:
Closing because of failed rebase. See #5238

PR #4416, opened 2019-03-21
 Refactor TimePoint::createNew

### LINKED ISSUE #4414: Users at single site create invalid sessions
When a user is affiliated with a single site, the timepoint creation is broken and the session has no CenterID set.

This was broken by #4049, which incorrectly set the type hints for the visit label and centre to nullable strings in the `TimePoint::createNew` function (presumably because that was the default value prior to type hinting, even though the doc comment was correct and it shouldn't be valid to have a session with no site or visit label.) When a user is affiliated with multiple sites, the new visit page prompts them to select which site the session is for, and the field is populated based on that. When a user is only affiliated with a single site, it does not get set at all and ends up being passed as null by the create_timepoint page.

The API also appears to not pass the center to createNew (even though it validates it, it doesn't pass it to create new.)  (The create timepoint page also doesn't check for the case where a user has no site affiliations, while the API does.)

The nullable visit label code in createNew() appears to be written with the intend of allowing the possibility of having visits created consecutively (V1, V2, V3, etc) by passing a "null" value as the visit label. This is probably valid, but should also probably be validated by the caller checking the config settings before calling createNew, not by createNew assuming that's the case based on a null parameter.

driusan:
This refactors the TimePoint::createNew function so that it:

1. Uses classes instead of integers/strings for its parameters
2. Does not directly access data from SQL
3. Requires the site to be passed explicitly.

It also fixes a bug where V0.0.2 of the API was not passing the CenterID to 
createNew and verifies that a user has at least one site before allowing them
to create timepoints in create_timepoint.

Fixes #4414.

### DIFF
--- a/htdocs/api/v0.0.2/candidates/Visit.php
+++ b/htdocs/api/v0.0.2/candidates/Visit.php
@@ -196,7 +196,12 @@ class Visit extends \Loris\API\Candidates\Candidate
                 $this->safeExit(0);
             }
             // need to extract subprojectID
-            $this->createNew($this->CandID, $subprojectID, $this->VisitLabel, $centerID);
+            $this->createNew(
+                $this->CandID,
+                $subprojectID,
+                $this->VisitLabel,
+                $centerID
+            );
             $this->header("HTTP/1.1 201 Created");
         }
     }
@@ -211,6 +216,7 @@ class Visit extends \Loris\API\Candidates\Candidate
      * @param integer $subprojectID The subproject for the new visit
      * @param string  $VL           The visit label of the visit to
      *                              be created
+     * @param integer $CID          The CenterID for this timepoint
      *
      * @return void
      */
--- a/htdocs/api/v0.0.2/candidates/Visit.php
+++ b/htdocs/api/v0.0.2/candidates/Visit.php
@@ -196,7 +196,7 @@ class Visit extends \Loris\API\Candidates\Candidate
                 $this->safeExit(0);
             }
             // need to extract subprojectID
-            $this->createNew($this->CandID, $subprojectID, $this->VisitLabel);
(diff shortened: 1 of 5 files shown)


php/libraries/TimePoint.class.inc
```diff
@@ -188,21 +188,22 @@ class TimePoint
...
      *
-     * @param integer     $candID         6 digit CandID having a timepoint created.
+     * @param \Candidate  $candidate      The Candidate to create this timepoint for
      * @param integer     $SubprojectID   The subprojectID of the new timepoint
      * @param string|null $nextVisitLabel The visit label of the timepoint
-     *                                    being created
-     * @param string|null $psc            The centre of the timepoint
+     *                                    being created. If null, consecutive V1, V2,
+     *                                    etc will be created.
+     * @param \Site       $psc            The centre of the timepoint
```
johnsaigle:
```suggestion
     * @param \Site       $psc            The center of the timepoint
```

I think we decided to change this a while ago.

htdocs/api/v0.0.2/candidates/Visit.php
```diff
@@ -196,7 +196,12 @@ public function handlePUT()
                 $this->safeExit(0);
             }
             // need to extract subprojectID
-            $this->createNew($this->CandID, $subprojectID, $this->VisitLabel);
+            $this->createNew(
+                $this->CandID,
```
johnsaigle: Might be nice to use the CandID class here. Unless you want to wait for #4350

driusan: This was just the minimal change to update the API call. It should be passing the \Candidate class itself and match the static function that it's delegating to, but this is dead code once #4244 gets merged anyways.

PR #4417, opened 2019-03-21
[Tools] Prototype of selective data extraction and importing to open datasets

johnsaigle:
### Brief summary of changes

Prototype for data extraction and importing. Must have a mapping CSV file in order for this script to work. This has the format:

> OLDPSCID, NEWPSCID, NEWCANDID

#### To Test
* Run `php tools/dataExtractor.php candidate <column>`.
    * Only the `candidate` table is supported for now
* This script will generate a csv file `candidate_$column_dataExtract_output.csv`
* Run `php tools/dataImporter.php <mapping.csv> <data.csv>`
* Verify that the UPDATE statements printed by the second script correspond to the data in the source database but with the PSCID changed to the new PSCID value.

More details are in the code comments.


tools/dataImporter.php
```diff
@@ -0,0 +1,255 @@
...
+const SET_COLUMN = 0;
+const SET_NEWVALUE = 1;
+const WHERE_COLUMN = 0;
+const WHERE_OLDVALUE = 1;
+
+require_once "generic_includes.php";
+
+$usage = <<<USAGE
+Usage: php {$argv[0]} <mapping.csv> <data.csv>
+<mapping.csv>   A CSV file containg columns with NEWPSCID,
```
Jkat: OLDPSCID, NEWPSCID, NEWCANDID

tools/dataImporter.php
```diff
@@ -0,0 +1,255 @@
...
+    $oldPSCIDsInDataFile[] = $row[$dataHeaders[OLD_PSCID]];
+}
+// array_intersect preserves keys but we don't want them.
+$sharedCandidates = array_values(
+    array_intersect($oldPSCIDsInMappingFile, $oldPSCIDsInDataFile)
+);
+
+// Build a one-to-one mapping of old to new PSCIDs with the old IDs as the key.
+$PSCIDMapping = array();
+foreach($mappingRows as $row) {
```
Jkat: How about iterating over `$sharedCandidates` instead? (haven't thought about whether this approach would be more extensible or not though)

johnsaigle: Yeah that probably makes more sense

Jkat: You could probably iterate over `$sharedCandidates` on line 147(https://github.com/aces/Loris/pull/4417/files#diff-4b03c0429757434e7d7c0d9634690cceR147) too and have less checks and directly access the data array with the help of `$PSCIDMapping`. 
But yeah both of these are just bonus enhancements. This is a nice prototype! Good job.

johnsaigle: Actually I think this way works more sense just given the way the data is organized. The function I'm using to put the CSV into an array creates an array like this:
```
[0] => Array (
     [PSCID] => MON123,
     [COLUMN] => some_field
)
...
```
So there's no way to directly look up on PSCID (using `$sharedCandidates`).

The issue comes with line 152 -- you can't directly grab a field like that given the array structure. You'd have to search through `$dataRows` every time which makes the loop O(n^2) whereas the preprocessing above line 147 makes the foreach loop O(n)... I think

Not sure if that makes sense, I am a little braindead right now

PR #4418, opened 2019-03-21
Add new phan rule PhanNonClassMethodCall

johnsaigle:
### Brief summary of changes

Phan will now make sure that we are only calling methods on entities that it believes are objects.

Not too many bugs to fix with this one.

### DIFF
--- a/php/libraries/NDB_Factory.class.inc
+++ b/php/libraries/NDB_Factory.class.inc
@@ -197,15 +197,18 @@ class NDB_Factory
         }
         $config = $this->config();
         $dbc    = $config->getSetting('database');
-        if (!is_null($db_ref)) {
-            $db_ref->connect(
-                $dbc['database'],
-                $dbc['username'],
-                $dbc['password'],
-                $dbc['host'],
-                true
+        if ($db_ref === null) {
+            throw new \ConfigurationException(
+                'Could not configure database for LORIS'
             );
         }
+        $db_ref->connect(
+            $dbc['database'],
+            $dbc['username'],
+            $dbc['password'],
+            $dbc['host'],
+            true
+        );
         return $db_ref;
     }
 
--- a/php/libraries/NDB_Factory.class.inc
+++ b/php/libraries/NDB_Factory.class.inc
@@ -181,7 +181,6 @@ class NDB_Factory
      */
     function database(): \Database
     {
-        $db_ref = null;
         if ($this->Testing) {
             $db_ref = &self::$testdb;
             if ($db_ref !== null) {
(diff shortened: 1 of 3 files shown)


driusan:
@johnsaigle can you upgrade this one to minor since minor and major are so close now?

johnsaigle:
@driusan Done

php/libraries/NDB_Factory.class.inc
```diff
@@ -194,17 +193,19 @@ class NDB_Factory
...
-        );
+        if (!is_null($db_ref)) {
+            $db_ref->connect(
+                $dbc['database'],
+                $dbc['username'],
+                $dbc['password'],
+                $dbc['host'],
+                true
+            );
+        }
```
driusan: I don't understand this. Given the function signature, I only see 2 possibilities:

1. $db_ref is never null (and determined as such by static analysis), in which case the if statement is meaningless.
2. $db_ref can be null in some code paths (and is determined as such by static analysis), in which case the return line below should get flagged as being incompatible with the function signature (which isn't nullable)

johnsaigle: I don't understand it either but it keeps coming up. It's somehow an issue existing in the code already but doesn't get flagged by `phan`.

A similar thing happened in my other PR here:
https://github.com/aces/Loris/pull/4914/files#diff-f89f7046f643b2862edaf5cce173dc17R111

Phan always thinks this is null.

I think it have some sort of troubles working with references that are set to `null` initially and later kind of juggled around in this way.

I tried updating phan to the latest version locally but the issue still comes up so it's not a bug (or at least not one they're aware of or disagree with).

I'm open to suggestions, either a different refactoring approach (like in the PR I linked) or to adding a suppression here.

Let me know what you want to do.

driusan: Would phan be happy if you refactored it to

```
if ($db_ref === null) {
     throw Exception
}
$db_ref->connect();
return $db_ref;
```
?

johnsaigle: I think so. I'll change it.

php/libraries/NDB_Factory.class.inc
```diff
@@ -194,10 +193,15 @@ class NDB_Factory
             if ($db_ref !== null) {
                 return $db_ref;
             }
-                self::$db = Database::singleton();
+            self::$db = Database::singleton();
         }
         $config = $this->config();
         $dbc    = $config->getSetting('database');
+        if ($db_ref === null) {
```
zaliqarosli: looks to me here that the else case (db_ref != null) will already exit on either line 187 or 194 so the happy path below the exception throw will never be reached?

johnsaigle: Yeah Dave and I discussed this above but the discussion was collapsed. Phan seems to have an issue interpreting references the way this class is using them.

zaliqarosli: yeah i saw the conversation and collapsed it because it was resolved but i didn't think this was connected. the code is a bit confusing to me

johnsaigle: Yeah it's confusing. I'll add a comment. This should be fine/not break things though.

PR #4419, opened 2019-03-21
[Server Processes Manager] Fix LORIS crashing when mriCodePath is invalid

johnsaigle:
### Brief summary of changes

Fixes a white-screen-of-death when `mriCodePath` is not set and the user tries to go to Server Processes Manager

Also adds a new integration test to check for this case.

### To test this change...

- [x] Delete the `mriCodePath` setting in the Congifuration module
- [x] On `major`, Server Processes Manager will crash
- [x] On my branch you will get an error message instead


johnsaigle:
Travis failed due to 

Another case of this issue #4394

PR #4425, opened 2019-03-22
[Utility] Add isPositiveInteger function (+ unit tests)

johnsaigle:
### Brief summary of changes

When checking data coming from e.g. the database or a form, there are some weird subtleties with using `intval()`. It'll return `0` on an input of `null` which makes it hard to differentiate between inputs that are `null` and `0` and those that are unset. It also has non-intuitive behaviour for array and object inputs, see: https://secure.php.net/manual/en/function.intval.php

This PR adds a simple function to check whether a given value is a positive integer as well as unit tests to prove that it works.

### DIFF
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -963,7 +963,7 @@ class Utility
      *
      * @return bool Whether $value is a positive int.
      */
-    public static function isPositiveInteger($value): bool
+    public static function valueIsPositiveInteger($value): bool
     {
 
         // Reject arrays and objects
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -952,5 +952,28 @@ class Utility
         }
         return implode('', $pieces);
     }
+
+     /**
+     * Takes an argument and determine whether it is a positive integer
+     * i.e. > 0.
+     *
+     * @param mixed $value The value to check. Intentionally not type hinted as
+     * this function is meant to assist in cases where the type is unknown, e.g.
+     * when a parameter is sent via a form.
+     */
+    public static function isPositiveInteger($value): bool {
+        // First clause checks that only digits are present
+        // Second clause ensures a positive value
+        // We can't use intval() alone because it has odd behaviour when given
+        // arrays and objects.
+        // @link https://secure.php.net/manual/en/function.intval.php
+        
+        // Reject arrays and objects
+        if (!is_scalar($value)) {
+            return false;
+        }
+
+        return ctype_digit(strval($value)) && intval($value) > 0;
+    }
 }
 
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -963,8 +963,9 @@ class Utility
      *
      * @return bool Whether $value is a positive int.
      */
-    public static function isPositiveInteger($value): bool {
-        
+    public static function isPositiveInteger($value): bool
+    {
+
         // Reject arrays and objects
         if (!is_scalar($value)) {
             return false;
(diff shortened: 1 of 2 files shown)


php/libraries/Utility.class.inc
```diff
@@ -916,5 +916,28 @@ class Utility
...
+
+        // Reject arrays and objects
+        if (!is_scalar($value)) {
+            return false;
+        }
+
+        // First clause checks that only digits are present
+        // Second clause ensures a positive value
+        return ctype_digit(strval($value)) && intval($value) > 0;
+    }
```
ridz1208: @johnsaigle isnt it a bit confusing if you pass a string and it returns true ?

i.e. `'5'` is a positive integer but not an integer since its a string... 

you see what I mean ? might be confusing with `is_int ( mixed $var ) : bool`

johnsaigle: Yeah I do see what you mean. Maybe I could change the name to like `stringIsPositiveInteger` and have it only take a string as input?

ridz1208: maybe or something like `containsPositiveInteger()` making it clear its not it's type ?

johnsaigle: Yeah but `contains` makes me think that if a positive integer exists anywhere in the input that function should return true. Technically `abc100def` contains a positive integer...

PapillonMcGill: maybe `valueIsPositiveInterger` or `equalPositiveInteger`?

I Approve pending the proper name for this function.

johnsaigle: I like `valueIsPositiveInteger`

johnsaigle:
I've updated the function name. @ridz1208 @PapillonMcGill could you please review when you have time?

johnsaigle:
@driusan Thoughts on merging this?

test/unittests/UtilityTest.php
```diff
@@ -831,3 +823,54 @@ public function testGetSourcefieldsWithAllThreeParameters()
     }
 }
 
+=======
+
+=======
+>>>>>>> 81d586a44... Fix file and class doc
```
johnsaigle:
```suggestion
>>>>>>> 81d586a44... Fix file and class doc
```

test/unittests/UtilityTest.php
```diff
@@ -831,3 +823,54 @@ public function testGetSourcefieldsWithAllThreeParameters()
     }
 }
 
+=======
+
+=======
```
johnsaigle:
```suggestion
```

test/unittests/UtilityTest.php
```diff
@@ -831,3 +823,54 @@ public function testGetSourcefieldsWithAllThreeParameters()
     }
 }
 
+=======
```
johnsaigle:
```suggestion
```

test/unittests/UtilityTest.php
johnsaigle:
```suggestion
```

test/unittests/UtilityTest.php
```diff
@@ -829,5 +821,51 @@ public function testGetSourcefieldsWithAllThreeParameters()
...
+                [-1],
+                [0],
+                [3.14],
+                ['abcdefg'],
+                ['-1'],
+                ['-98.6'],
+                ['0'],
+                [array()],
+                [array(1)],
+                [new stdClass()]
```
maltheism: Maybe null should be added?

johnsaigle: Sure, I'll do that

johnsaigle:
@ridz1208 @HenriRabalais Could you re-approve now that the code has been rebased?

PR #4426, opened 2019-03-23
Standardize module .gitignore files

ridz1208:
### Brief summary of changes
Given that we are heading towards self-contained modules I think we should be using a central `.gitignore` file per module. currently these files are added in multiple different directories such as `training_docs/`, `user_uploads/` and `js/`. I find that adding them to the module root makes them easier to find and standardized accross all modules

This PR does 3 things:
 - [x] move all `.gitignore` files to module root
 - [x] add new compiled JS files to list of ignored files (to avoid untracked files on production environments)
 - [x] remove joker `*.js` and replace with explicitly ignore files to avoid errors

### To test this change...

without this PR: 
- [ ] clean your git repository from all untracked files
- [ ] checkout aces/major
- [ ] run `make dev`
- [ ] you should see several "untracked" newly compiled js files

with this PR:
- [ ] clean repository again
- [ ] checkout this PR
- [ ] run `make dev`
- [ ] No compiled JS files should appear as untracked


ridz1208:
@PapillonMcGill @johnsaigle Thx for reviewing, I made an additional change that removed your reviews.

I didnt like the `js/*` and `js/*.js` because they might be untracking files that should be tracked so i made them all explicit.

If you can re-review and maybe give it a test run as well would be great

ridz1208:
@driusan reverted all jocker removals

PR #4430, opened 2019-03-28
[Tool/Innodb] Fix help documentation and change confirm to apply

ridz1208:
### Brief summary of changes
part 1: The instructions refer to the wrong php file name
part 2: I would like to standardize some aspects of how we write tools and expected behaviour. one of them is offering the possibility to generate SQL code vs applying changes directly in the database. the `confirm` option was changed to `apply` here to make it clearer that the changes will take effect directly


PapillonMcGill:
@ridz1208 Does that mean that all others scripts will have to change from `confirm` to `apply`?

ridz1208:
@PapillonMcGill yes, I'm starting with the scripts that have benn "cleaned up" and will be making a document with "recommendations" for tools

PR #4432, opened 2019-03-28
[User_accouts] bugfix - can't create a new user

kongtiaowang:
Remove the password check, new user the password always be null.

To test: create a new user.


johnsaigle:
Failing Travis due to phan error:
> modules/user_accounts/php/edit_user.class.inc:1155 PhanTypeArraySuspiciousNullable Suspicious array access to nullable ?array

PR #1965, opened 2016-07-02
Initial Debian packaging draft

mih:
This is the result of the sprint at OHBM2016. This can be used to built a package, but it is incomplete and untested.

I was not sure which branch to make the PR against (we also talked about this).

PR #4434, opened 2019-03-28
Make tools/ scripts executable

johnsaigle:
### Brief summary of changes

This script adds executable permissions to all the tools/ scripts.

### This resolves issue...

This is an issue created by `lintian` as reflected in the output in #1965 and #3154. We need to add execute permissions to all scripts starting with `#!` in order to create a LORIS package.


johnsaigle:
Failed due to #4394

driusan:
Technically, I don't think lintian *needs* to pass in order to build a package, it just needs to pass in order to have it accepted into Debian.

PR #4435, opened 2019-04-01
[php/libraries/Project.class.inc] fix type hinting error by cast int

maltheism:
### Brief summary of changes

This fixes a bug from type hinting on major the LORIS API:
https://{loris-instance}/api/v0.0.3-dev/projects/{project_name}/images

### To test this change...

- [x] Execute the above API

### DIFF
--- a/php/libraries/Project.class.inc
+++ b/php/libraries/Project.class.inc
@@ -25,6 +25,10 @@ class Project
 {
     private static $_instances = array();
 
+    /**
+     * The id of the project in database.
+     * @var int
+     */
     private $_projectId;
     private $_projectName;
     private $_recruitmentTarget;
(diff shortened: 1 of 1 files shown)


php/libraries/Project.class.inc
```diff
@@ -75,7 +75,7 @@ class Project
 
                 $project = new Project();
 
-                $project->_projectId         = $projectData['projectId'];
+                $project->_projectId         = (int) $projectData['projectId'];
```
driusan: Can you add a `@var int` annotation at the definition of `$project->_projectId` so that phan can detect if there are any other usages where the the type conflicts?

maltheism: @driusan thanks and added.

johnsaigle:
Needs PHPCS

driusan:
This is passing Travis if you click through, I don't know why GitHub isn't updating the icon.

PR #4436, opened 2019-04-01
Revert "[Database] Add NOQUOTES as an HTML escaping option for JSON"

driusan:
Reverts aces/Loris#4374.


ridz1208:
@driusan @johnsaigle  reasons ?

driusan:
@ridz1208 because @johnsaigle said you found bugs with it and requested it be reverted. See comment on #4374

PR #4438, opened 2019-04-01
[React/jsx] Render action buttons when data table is empty

zaliqarosli:
### Brief summary of changes
At the moment, if the data table is empty, the action buttons aren't rendered. This PR renders action buttons, i.e. to add an entry to the table, when no result found.

before: 
55336947-c8496d80-546b-11e9-81a5-01663b035302

after:
Screenshot 2019-04-01 16 03 23

## To test:

1. choose newly reactified module with action buttons e.g. 'examiner'. manipulate the data so that the data table is empty. you will see that there is no way to add an examiner.
2. on this branch, you can add an entry to the empty table.

### DIFF
--- a/htdocs/bootstrap/css/custom-css.css
+++ b/htdocs/bootstrap/css/custom-css.css
@@ -340,7 +340,7 @@ div.navbar div.container div.navbar-brand {
     color: #064785;
     background-color: #E4EBF2;
     border-color: #E4EBF2;
-    margin: 0px 14px;
+    margin: 0 14px;
 }
 
 /*
--- a/htdocs/bootstrap/css/custom-css.css
+++ b/htdocs/bootstrap/css/custom-css.css
@@ -340,6 +340,7 @@ div.navbar div.container div.navbar-brand {
     color: #064785;
     background-color: #E4EBF2;
     border-color: #E4EBF2;
+    margin: 0px 14px;
 }
 
 /*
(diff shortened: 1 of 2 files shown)


maltheism:
Hi @zaliqarosli, I haven't been able to replicate the Acknowledgements module screenshot of yours with clearing out everything in the acknowledgements table of my database. I think the Acknowledgements module uses the jsx/StaticDataTable.js

htdocs/bootstrap/css/custom-css.css
```diff
@@ -340,6 +340,7 @@ div.navbar div.container div.navbar-brand {
     color: #064785;
     background-color: #E4EBF2;
     border-color: #E4EBF2;
+    margin: 0px 14px;
```
maltheism: Inside CCS-land you don't need "px" or whatever measurment when the value is 0.
You can use this instead `margin: 0 14px;` and it just makes it a little cleaner. Unless you really write your own CSS a lot, you won't be bothered by it.

zaliqarosli:
```suggestion
    margin: 0 14px;
```

zaliqarosli:
@maltheism i agree but i dont think that belongs in this PR. I'll create a github issue to add doc comment for the renderActions() function. please approve if nothing else if holding this PR back!

zaliqarosli:
@maltheism github issue: https://github.com/aces/Loris/issues/4457

PR #4447, opened 2019-04-04
[major] Updated major node_modules in package.json

maltheism:
### Brief summary of changes

This PR fixes the "found 50 **vulnerabilities**" after running npm install. 
- Major is now using the **latest packages** in package.json.
- Updated what should fall under "**dependencies**" and "**devDependencies**" in package.json.
- This PR required "**npm run lintfix:javascript**" to satisfy the .eslintrc.json upgrade of the eslint-config-google package (I believe).

### To test this change...

- [x] npm install & npm run compile 

Note: **All files changed were necessary and done by lintfix:javascript**.
Reviewers can look at the webpack.config.js, package.json, and .babelrc (all other files were edited by npm run lintfix:javascript and nothing I could do about that so ignore them but they're necessary in fixing npm run compile).

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,9 @@
 {
   "presets": [
-    "react",
-    "env"
+      "@babel/preset-env",
+      "@babel/preset-react"
   ],
-  "plugins": ["transform-object-rest-spread"]
+  "plugins": [
+      "@babel/plugin-proposal-object-rest-spread"
+  ]
 }
--- a/jsx/Breadcrumbs.js
+++ b/jsx/Breadcrumbs.js
@@ -77,19 +77,19 @@ class Breadcrumbs extends Component {
       const url = baseURL + element.query;
       if (i < this.props.breadcrumbs.length - this.state.displayCount) {
         dropdown.push(
-          <li key={'drop_' + i}>
-            <a href={url}>
-              {element.text}
-            </a>
-          </li>
+            <li key={'drop_' + i}>
+              <a href={url}>
+                {element.text}
+              </a>
+            </li>
         );
       } else {
         breadcrumbs.push(
-          <a key={'crumb_' + i} href={url} className='btn btn-primary'>
-            <div>
-              {element.text}
-            </div>
-          </a>
+            <a key={'crumb_' + i} href={url} className='btn btn-primary'>
+              <div>
+                {element.text}
+              </div>
+            </a>
         );
       }
     }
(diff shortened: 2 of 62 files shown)


johnsaigle:
This seems pretty hard to review... do you think you could issue one PR with the ESLint changes and later a second one with the node updates?

In general it's best to separate formatting changes from changes with consequence.

maltheism:
@johnsaigle I don't think it's that hard to review. Reviewers just need to look in the `webpack.config.js`, `package.json` and `.babelrc` because those are the only files I modified. Everything else is from running npm run lintfix:javascript
The reason I had to run lintfix:javascript is because npm run compile was literally broken with output of errors after running it. I'm thinking Travis won't complete if I do what you're suggesting because npm run compile work. Thoughts?

maltheism:
.babelrc
https://github.com/aces/Loris/pull/4447/files#diff-e56633f72ecc521128b3db6586074d2c

package.json
https://github.com/aces/Loris/pull/4447/files?file-filters%5B%5D=.json#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

johnsaigle:
`npm run compile` doesn't seem to be broken on `major` -- though I agree we should fix the vulnerabilities. This PR does seem to cause Travis to fail though. 

If it's only one package causing the problem I think a simpler approach would be to start fresh and use ESLint on only the files the `eslint-config-google` package flags, if that is indeed the problem. 

Alternatively if there are no vulnerabilities in that package, we don't need to upgrade it to the latest version just yet.

johnsaigle:
Running `npm audit` shows that updating `js-yaml`, `debug`, and `webpack` should suffice.

maltheism:
@johnsaigle after upgrading the packages the problem resulted. Example running `npm run compile` would result in errors after the packages were updated. I ran `npm run lintfix:javascript` to resolve that problem. I'm not referring to running `npm run compile` on major without the updates to package.json.

maltheism:
Okay, so `npm run lintfix:javascript` doesn't target or fix the htdocs directory in loris. It seems that's why Travis is complaining and broken. The exact errors would be happening for everywhere else if I didn't run `npm run lintfix:javascript`.

driusan:
@maltheism can you split this into 2 commits? One that does the changes you've done, and one that only runs the autofixer. I think that would address the problem of it being difficult to review and make it easier to be sure that you didn't accidentally change anything.

maltheism:
@driusan okay

maltheism:
See #4449

PR #4449, opened 2019-04-08
[major] Updated major node_modules in package.json

maltheism:
### Brief summary of changes

This PR fixes the "found 50 **vulnerabilities**" after running npm install. 
- Major is now using the **latest packages** in package.json.
- Updated what should fall under "**dependencies**" and "**devDependencies**" in package.json.
- This PR required "**npm run lintfix:javascript**" to satisfy the .eslintrc.json upgrade of the eslint-config-google package (I believe).

### To test this change...

- [x] npm install & npm run compile 

Note: **All files changed were necessary and done by lintfix:javascript**.
Reviewers can look at the webpack.config.js, package.json, and .babelrc (all other files were edited by npm run lintfix:javascript and nothing I could do about that so ignore them but they're necessary in fixing npm run compile).

Please see #4447 
- First commit updated package.json, package-lock.json, .babelrc
- Second commit ran the command: npm run lintfix:javascript
- Third commit ran the command: eslint --fix --ext .js, htdocs/js/
- Fourth commit corrected indentation of webpack.config.js


johnsaigle:
Adding Needs Rebase because of the file conflict

maltheism:
Thanks @johnsaigle I fixed the conflict.

johnsaigle:
Another conflict has come up

package.json
```diff
@@ -7,28 +7,27 @@
...
-    "webpack-cli": "3.1.0"
+    "@babel/cli": "^7.4.3",
+    "@babel/core": "^7.4.3",
+    "@babel/preset-env": "^7.4.3",
+    "@babel/preset-react": "^7.0.0",
+    "@babel/plugin-proposal-object-rest-spread": "^7.4.3",
+    "babel-eslint": "^10.0.1",
+    "babel-loader": "^8.0.5",
+    "eslint": "^5.16.0",
+    "eslint-config-google": "^0.12.0",
```
johnsaigle: Just did some testing and it looks like you can update everything without causing any syntax errors if you leave out `eslint-config-google` and keep it at version 0.9.1

I think it's best for organization purposes to update the security-related packages in this PR and have another cleanup/style change PR separately. It goes with our convention of trying to separate `Cleanup` tasks from `Bug Fix` tasks.

It makes the review process and git history cleaner. Also gives more of an opportunity for JS people to decide whether to follow Google's indentation standards or keep things as they have been.

For reference, these are the only changes made between `eslint-config-google` between versions 0.9.1 and 0.12 https://github.com/google/eslint-config-google/compare/0.9.1...master

maltheism: @johnsaigle I'm a little confused why this PR is difficult to review. The first commit does the changes. All the other commits were just updating the syntax with the automatic eslint command. This is following the suggestion of @driusan 

I think we're supposed to be following the eslint-config-google standard as it's always been included in the configuration file but for some reason it wasn't being in-forced like now.

I could make another PR with removing eslint-config-google and update the modules (that's also the point of this PR). Then make another PR that updates esling-config-google and apply the syntax changes. I just think it would be easy if people just reviewed the first commit because all the other commits afterwards are just running the automatic eslint fixes. I'm not doing anything manually that would break anything.

zaliqarosli:
smaller, newer PR https://github.com/aces/Loris/pull/4968

PR #4454, opened 2019-04-10
[Issue_tracker]  bug fix for 20.3-release | can't edit an issue

kongtiaowang:
### Brief summary of changes
Give an issueID on url.

### This resolves issue...
https://redmine.cbrain.mcgill.ca/issues/16241

### To test this change...
edit an issue


kongtiaowang:
@ridz1208 
This PR is for the 20.3-release

modules/issue_tracker/jsx/columnFormatter.js
```diff
@@ -26,7 +26,7 @@ function formatColumn(column, cell, rowData, rowHeaders) {
   if (column === 'Title') {
     let cellLinks = [];
     cellLinks.push(
-      <a href={loris.BaseURL + '/issue_tracker/issue/' + row['Issue ID']}>
+      <a href={loris.BaseURL + '/issue_tracker/issue/?issueID=' + row['Issue ID']}>
```
driusan: This is not correct. The ID should be coming from the handle method parsing the URL at https://github.com/aces/Loris/blob/minor/modules/issue_tracker/php/issue.class.inc#L144

kongtiaowang: @driusan If use RESTful API in this case, the "QueryString.get()" will get a null value. 
Major branch is also using "?issueID=" on URL. I think if you need this to be fixed, it should on Major branch.

driusan: in that case the major branch is wrong and also needs to be fixed. The ID should be coming from the URL, not the query string

ridz1208: @driusan will look into the `handle` function and try to solve the issue

ridz1208:
https://github.com/aces/Loris/pull/3942

modules/issue_tracker/jsx/columnFormatter.js
kongtiaowang: @driusan @ridz1208 could u review again?

PR #4456, opened 2019-04-10
[timepoint_list] Breadcrumb link fix for Candidate profile

maltheism:
### Brief summary of changes

Fixes the breadcrumb containing an extra forward slash. Ex. http://localhost//162517/

### To test this change...

- [x] Click on candidate and then click on the Candidate Profile breadcrumb.

https://github.com/aces/Loris/issues/4723

### DIFF
--- a/modules/timepoint_list/php/timepoint_list.class.inc
+++ b/modules/timepoint_list/php/timepoint_list.class.inc
@@ -309,8 +309,6 @@ class Timepoint_List extends \NDB_Menu
         $candid = $this->tpl_data['candID'];
         $pscid  = $this->tpl_data['PSCID'];
 
-        $link = ltrim($_SERVER['REQUEST_URI'], '/');
-
         return new \LORIS\BreadcrumbTrail(
             new \LORIS\Breadcrumb(
                 'Access Profile',
@@ -318,7 +316,7 @@ class Timepoint_List extends \NDB_Menu
             ),
             new \LORIS\Breadcrumb(
                 "Candidate Profile $candid / $pscid",
-                "/$candid"
+                "$candid"
             )
         );
     }
--- a/modules/timepoint_list/php/timepoint_list.class.inc
+++ b/modules/timepoint_list/php/timepoint_list.class.inc
@@ -316,7 +316,7 @@ class Timepoint_List extends \NDB_Menu
             ),
             new \LORIS\Breadcrumb(
                 "Candidate Profile $candid / $pscid",
-                "$candid"
+                $this->candID
             )
         );
     }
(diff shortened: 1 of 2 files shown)


xlecours:
@maltheism This would solve 1 instance of the problem. Maybe the React Breadcrumb could be refactor to cleanup the url.

maltheism:
@xlecours the bug seems to only occur for me when going to the Candidate Profile with breadcrumbs. I could write some javascript code that checks if the URL string contains "//" and alter it to just have one slash but that seems weird when the problem is from the backend.

modules/timepoint_list/php/timepoint_list.class.inc
```diff
@@ -309,16 +309,14 @@ class Timepoint_List extends \NDB_Menu
...
-
         return new \LORIS\BreadcrumbTrail(
             new \LORIS\Breadcrumb(
                 'Access Profile',
                 '/candidate_list'
             ),
             new \LORIS\Breadcrumb(
                 "Candidate Profile $candid / $pscid",
-                "/$candid"
+                "$candid"
```
zaliqarosli: i see this issue in timepoint list but i don't think this is the solution. I wonder if $candid has a `\` in it when it shouldn't? the fix might be in there. could maybe use `$this->candID` directly.

zaliqarosli: add: i am proposing this because this issue doesn't seem to be appearing anywhere else and   the code for other modules include a `/` in the candid input for `new \LORIS\Breadcrumb`

maltheism: @zaliqarosli okay.

modules/timepoint_list/php/timepoint_list.class.inc
```diff
@@ -309,16 +309,14 @@ class Timepoint_List extends \NDB_Menu
...
-
         return new \LORIS\BreadcrumbTrail(
             new \LORIS\Breadcrumb(
                 'Access Profile',
                 '/candidate_list'
             ),
             new \LORIS\Breadcrumb(
                 "Candidate Profile $candid / $pscid",
-                "/$candid"
+                $this->candID
```
zaliqarosli: this isn't the solution. you're still essentially only removing the `/` here

maltheism:
Hi @xlecours @zaliqarosli it seems to actually be a frontend issue like @xlecours suggested. Trying to pinpoint the proper solution still but at least I'm now less confused than before.

maltheism:
Okay it seems like the problem was that the loris.BaseURL has a slash at the end when it's not a react module. The main.tpl renders the Breadcrumbs and that's where I originally found out by testing the examiner module (thank you @zaliqarosli) and then testing the timepoint list.

ridz1208:
@driusan review ?

PR #4459, opened 2019-04-11
Separate DROP TABLES from CREATES in schema

ridz1208:
### Brief summary of changes
Proposition for everyone. instead of trying to manage the drop tables in a specific order in 5 different files. I put them all in here and fixed the order.

Commit message:
Separate DROP TABLEs from CREATE TABLEs in the 0000-00-08 schema files to allow for dumping and resourcing the entire schema without errors and without FOREIGN_KEY_CHECKS disabled. The order is key and and it is only possible to get the right order by removing the tables from the files


ridz1208:
@driusan @johnsaigle thats the only solution I have without using foreign key check disabling.

feel free to close if you dont like

PapillonMcGill:
@ridz1208 Why do this? What problem do you have exactly?

ridz1208:
@PapillonMcGill as I explained in person, This is to allow the drop tables to run successfully in the correct order. since the 00 schema is usually sourced first, some tables created in the 01,02,03,04,05 reference tables from 00

when time comes to drop the tables, the DROP TABLEs from 05 schema should be run before the 00 DROP TABLEs, the reverse order of creation. the only way of doing that was to separate the DROP TABLEs from the CREATE TABLEs

ridz1208:
@johnsaigle can you give it a try since you'r the one who found the issue in the first place ?

SQL/9999-99-99-drop_tables.sql
```diff
@@ -0,0 +1,163 @@
...
+DROP TABLE IF EXISTS `issues`;
+DROP TABLE IF EXISTS `issues_categories`;
+
+DROP TABLE IF EXISTS `media`;
+DROP TABLE IF EXISTS `server_processes`;
+DROP TABLE IF EXISTS `StatisticsTabs`;
+DROP TABLE IF EXISTS `user_login_history`;
+DROP TABLE IF EXISTS `user_account_history`;
+DROP TABLE IF EXISTS `data_integrity_flag`;
+DROP TABLE IF EXISTS `certification_training_quiz_answers`;
```
driusan: The training ones are going to conflict with #4535

ridz1208: @driusan merge this one first :)

PR #4463, opened 2019-04-16
[Instrument_builder] Fix page min-height for list

zaliqarosli:
### Brief summary of changes

Fixes page cutoff on instrument builder and any other page where there's not enough content to make the page bigger than the screen.

### This resolves issue...

- As reported by Yang Ding

### To test this change...

- [ ] Test on Firefox

### DIFF
--- a/htdocs/main.css
+++ b/htdocs/main.css
@@ -31,7 +31,7 @@ html, body {
 }
 
 #page {
-    min-height: 100%;
+    min-height: 100vh;
     position: relative;
     padding-top: 50px;
 }


zaliqarosli:
@dyt811

htdocs/main.css
```diff
@@ -31,7 +31,7 @@ html, body {
 }
 
 #page {
-    min-height: 100%;
+    min-height: -webkit-fill-available;
```
driusan: have you tried adding a min-height on the html/body or setting the min-height in vh instead of adding a webkit-only hack?

zaliqarosli: yeah i only just realized that webkit isn't supported on lets say, firefox

zaliqarosli: what's vh?

zaliqarosli:
```suggestion
    min-height: 100vh;
```

zaliqarosli: okay got it! the issue isn't html/body, both are at 100% height. it's the "page" div that gets cut because of the footer underneath it

driusan: percentage is relative to the parent. The previous CSS is 100% of the parent height, and I think the problem is that the parent isn't 100% of the size of the window (because its height is 100%, not its min-height)

zaliqarosli: it goes 
```
<html>
  <body>
    <div id="wrap">
      <div id="page">
```

body has height 100%, wrap has min-height 100% (no height because otherwise footer would hover over everything), and page has min-height 100%. i agree with you that page's height should become 100% but it doesn't for some reason

zaliqarosli: @driusan setting min-height to 100 vh seems to be the best option. is this still not a good idea?

driusan: 100vh is fine, as far as I know most browsers support css3 units.

johnsaigle:
Can you provide some screenshots or instructions about what to look for? Happy to test this but I'm not sure how.

dyt811:
> Can you provide some screenshots or instructions about what to look for? Happy to test this but I'm not sure how.

Sure thing. Here: 
[YD] Instrument Builder Minor UI Issue: 

- Instrument builder question creation dropdown clipped instead of flowing smoothly?
- Video of issue replication: https://drive.google.com/open?id=1eeuDGrQ26G7Up7od_ivPlwZYbr2awdvz  
- Notice the two scrolling bars on the right side of the video

driusan:
@zaliqarosli can you move this to the release branch?

zaliqarosli:
@driusan done!

PR #4467, opened 2019-04-18
Add publication module SQL to the schema files

ridz1208:
### Brief summary of changes

The SQL for publication module was omitted accidentally
+ changed a `utf8mb4` to `utf8` for consistency


driusan:
Is the converting to utf8 from utf8mb4 required? Our standards say to use mb4.

ridz1208:
1. all other tables are currently `utf8` they should all be `utf8mb4` in that case
2. Most production environments will need to be upgraded to 5.7 for this patch to work, but this is a minor release and that would be risky (but mostly reason 1)

driusan:
There are conflicts on this now (because of the patch archiving, I think).

ridz1208:
@driusan conflict resolved

PR #4476, opened 2019-04-25
[Travis] on minor fix node

kongtiaowang:
Fixing node to pass Travis.


ridz1208:
@driusan shouldn't we be removing the `dist:precise` so it defaults to Travis' default ??

https://github.com/aces/Loris/pull/4476/files#diff-354f30a63fb0907d4ad57269548329e3R102

kongtiaowang:
replace by https://github.com/aces/Loris/pull/4477

PR #4484, opened 2019-04-26
Sorted and Added Composer.lock and package-lock.json to .gitignore

ridz1208:
This adds 2 entries to the git ignore `Composer.lock` and `package-lock.json` both of which are generated by running `make dev` on the branch


driusan:
composer.lock and package-lock.json shouldn't be ignored. They're required to be tracked for builds to be reproducible..

ridz1208:
@driusan its strange then why on major composer.lock is appearing in my untracked list not uncommited changes

driusan:
dunno, but the solution should be to commit it. https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control

PR #4486, opened 2019-04-29
Issue tracker fix merged into wrong branch

ridz1208:
Fix bug where an issue could not be edited in issue tracker, they would create a new issue instead.

https://github.com/aces/Loris/pull/4454


ridz1208:
pending merge of https://github.com/aces/Loris/pull/4481

PR #4487, opened 2019-04-29
[Travis] phan and RB data fix

kongtiaowang:
### Brief summary of changes
fix phan error and schema

PR #4488, opened 2019-04-29
[Raisinbread] Fix/Update dataset

ridz1208:
- Apply latest SQL patches on the dataset
- add some configurations
- enable useproject
- added migration record to track(manually) which patches have been run on the dataset

Added phan fixes to pass travis
https://github.com/aces/Loris/pull/4487


raisinbread/migration
```diff
@@ -0,0 +1,32 @@
+#20.2
```
driusan: what's this file?

ridz1208: > added migration record to track(manually) which patches have been run on the dataset

ridz1208: because there are PRs open in parallel that are not based on a version that already has raisinbread, some things ar ebeing merged with SQL patches without being run on RB which means at any time, I dont know what has been exectuted and what hasnt

PR #2956, opened 2017-07-18
Set smarty's autoescape template variable option

driusan:
This sets the smarty escape_html option to true by default in the
Smarty_hook class, as well as adding a helper to change the value.

Instruments currently disable autoescaping, because a number of instruments
use hacks for layout which would break if this was changed without notice.

The override for instruments should likely be changed in LORIS 18.0, but
otherwise this should help minimize the risk of XSS attacks in LORIS modules.

PR #2921, opened 2017-07-03
[Security] Fix Reflected XSS vulnerability on all NDB_Pages 

johnsaigle:
## This PR sanitizes data in LorisForm, boosting security and removing the security burden from module developers

Previously, it was trivial to inject Javascript into nearly every page as malicious data could be injected via GET requests and served without being sanitized. This makes users vulnerable to CSRF, XSS, and phishing attacks among others.

By sanitizing values in LorisForm, we prevent many of these vulnerabilities. 

I suggest that this fix in combination with #2919, #2826 warrant another release on 17.0.

PR #4491, opened 2019-04-30
[Authentication] Fixing password encoding / decoding error

### LINKED ISSUE #4490: Password requirements
Hi,

Using password `z2XtUOHc[&Rw^cK` in the new password prompt on https://openpreventad.loris.ca results in being logged off and not being able to log in again using this password or the old one. Trying from the same environment with a simpler password works.

Jkat:
Resolves #4490 

See for more info: https://www.php.net/manual/en/function.htmlspecialchars-decode.php

### DIFF
--- a/php/libraries/User.class.inc
+++ b/php/libraries/User.class.inc
@@ -508,6 +508,7 @@ class User extends UserPermissions
      */
     function updatePassword($password, $expiry = null)
     {
+        $password = htmlspecialchars_decode($password);
         if (is_null($expiry)) {
             $expiry = date(
                 'Y-m-d',


php/libraries/User.class.inc
```diff
@@ -508,6 +508,7 @@ class User extends UserPermissions
      */
     function updatePassword($password, $expiry = null)
     {
+        $password = htmlspecialchars_decode($password);
```
driusan: I think this should be done before the call to updatePassword. What's encoding it? A different code path might not and would introduce a similar bug because it doesn't match the caller's expectations.

johnsaigle: Yeah the password shouldn't get encoded this way. We should track down the code that's calling htmlspecialchars and remove it (or replace it with base64 which would be more appropriate).

johnsaigle:
tldr: Justin's fix is correct and we should merge it. All password-change code paths are escaped by LorisForm and we shouldn't alter that. `tools/resetpassword.php` bypasses the User class so there is no issue there either.

---

Alrighty I've tracked down this issue and it turns out to be a problem with technical debt, essentially.

The issue arises with code introduced in [this PR](https://github.com/aces/Loris/pull/2921) which is still present in LorisForm. This code calls `htmlspecialchars()` on everything POSTed by a user via a LorisForm before doing any further processing. This kick in before any `_process()` function in a module. The goal of this is to provide a way to protect against JavaScript injection as far upstream as possible.

We're seeing a ramification of that here -- passwords containing html entities (including `&` but also e.g. `<`, `>`) get escaped. They get escaped _using the & symbol_: `testing1&` becomes `testing1&amp;` which passes our password standards because it still contains `&` (lol). When trying to log in, a user enters the original version which no longer matches the hash and creates and account lockout.

Looking back on that PR with two more years of perspective, it definitely uses a broad brush to tackle that security issue. However, I believe we still need it. In many cases it's the only way that JS injection is prevented in many modules, afaik. This may have changed as things have become Reactified but until that process is complete and smarty is removed, I think it would be dangerous to delete the above line. 

This is why it goes back to technical debt... the only reason that the code is added in LorisForm is because [we are unable to use smarty's auto escaping](https://github.com/aces/Loris/pull/2956) or LORIS will break. 

This issue should be revisited once Smarty is removed from LORIS.

PR #4492, opened 2019-04-30
[SQL] Rename project_rel table and add constraints

ridz1208:
### Brief summary of changes
This renames the `project_rel` table to `project_subproject_rel` to respect the [SQLmodeling standards](https://github.com/aces/Loris/blob/minor/docs/SQLModelingStandard.md)

This also adds a front-end control to associate subprojects to projects from the project configuration page. (this should simplify testing)

other changes in this PR
 - add foreign key constraints to the `Project` and `subproject` tables for integrity
 - change fields to match their respective foreign keys


php/libraries/ProjectDefault.class.inc
```diff
@@ -85,7 +85,7 @@ class ProjectDefault extends Project
     }
 
     /**
-     * Get that project subprojects as defined in the project_rel table
+     * Get that project subprojects as defined in the project_subproject_rel table
```
xlecours: This is not accurate. Project default gets all subprojects because it is meant to be used when there is no project

xlecours: Otherwise, it seems fine

ridz1208: fixed

ridz1208:
blocked by https://github.com/aces/Loris/pull/4493

ridz1208:
DO NOT MERGE. its on the wrong base for now

SQL/New_patches/2019-04-30-project-rel-rename.sql
```diff
@@ -0,0 +1,9 @@
+RENAME TABLE project_rel TO project_subproject_rel;
+
+-- Change fields to math their respective reference keys
+ALTER TABLE project_subproject_rel CHANGE `SubprojectID` `SubprojectID` int(10) unsigned NOT NULL;
+ALTER TABLE project_subproject_rel CHANGE `ProjectID` `ProjectID` int(2) NOT NULL;
```
PapillonMcGill: This line seam to be not needed. `ProjectID` attribute is the same as before.

johnsaigle:

modules/configuration/ajax/updateProject.php
```diff
@@ -52,6 +54,18 @@
...
+
+$db->delete('project_subproject_rel', array('ProjectID' => $projectID));
+foreach ($_POST['SubprojectIDs'] as $sid) {
+    $db->insertIgnore(
+        'project_subproject_rel',
+        array(
+         'ProjectID'    => $projectID,
+         'SubprojectID' => $sid,
+        )
+    );
```
PapillonMcGill: What is the purpose of this?
Are you deleting all association for a project before re-inserting them all?

ridz1208: yes instead of comparing before and after the submission

PapillonMcGill: Then all `ProjectSubprojectRelID` will change with every update. 
This is not good.

ridz1208: good point that rel column was added after this PR was created, I'll have to change the logic

ridz1208: @PapillonMcGill fixed

PR #4493, opened 2019-04-30
[Visit] Removed redundancy with project_rel table and standardize tables

ridz1208:
### Brief summary of changes
This PR mainly removes the `ProjectID` column in the new visit table. the project is redundant with the `project_rel` table and given the future direction of the Projects in LORIS, its easier to reinforce that relationship now rather then later.

other changes here:
 - update to reflect latest [standards](https://github.com/aces/Loris/blob/minor/docs/SQLModelingStandard.md)
 - updated RB, tests and controller


ridz1208:
@PapillonMcGill Sorry, I know I instructed you to have the project in there but since then we had multiple discussions on reinforcing the projects role in loris.

ma bad

SQL/0000-00-00-schema.sql
```diff
@@ -2140,20 +2140,17 @@ CREATE TABLE `candidate_consent_history` (
 CREATE TABLE `visit` (
   `VisitID` int(10) unsigned NOT NULL auto_increment,
   `VisitName` varchar(100) NOT NULL,
-  CONSTRAINT `visit_PK` PRIMARY KEY (`VisitID`),
-  CONSTRAINT `visit_name_UK` UNIQUE KEY (`VisitName`)
+  CONSTRAINT `PK_visit` PRIMARY KEY (`VisitID`),
+  CONSTRAINT `UK_visit_name` UNIQUE KEY (`VisitName`)
```
driusan: what if the visit isn't in the UK?

ridz1208: Im not sure if i should laugh or answer seriously

PapillonMcGill:
approved after Travis agree!

ridz1208:
@PapillonMcGill thx for reviewing. I will be making one more structural change that will actually be a half way between what I have now and what you had before.

Dismissing the review for now

ridz1208:
@PapillonMcGill please review now

test/unittests/VisitTest.php
```diff
@@ -76,112 +76,28 @@ protected function setUp()
...
-                                       2,
-                                       4,
-                                      ),
-                                      array(
-                                       $v6,
-                                       3,
-                                       3,
-                                      ),
-                                     );
+            array($v1, 1, 1),
```
johnsaigle: lol the old code here makes me want to die. Excellent change

ridz1208:
@driusan I dont think anything is using this yet so I'm not sure if anything is testable other than the SQL patch

@sruthymathew123 @nicolasbrossard could you try running the SQL patch on IBIS and see if you get any errors?

SQL/New_patches/2018-01-17-normalisation_visit.sql
```diff
@@ -1,25 +1,27 @@
...
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
 
 CREATE TABLE `visit_project_subproject_rel` (
+  `VisitProjectSubprojectID` int(10) unsigned NOT NULL AUTO_INCREMENT,
   `VisitID` int(10) unsigned NOT NULL,
-  `ProjectID` int(2) NOT NULL,
-  `SubprojectID` int(10) unsigned NOT NULL,
-  CONSTRAINT visit_project_subproject_rel_PK PRIMARY KEY (`VisitID`, `ProjectID`, `SubprojectID`),
-  CONSTRAINT visit_project_subproject_rel_VisitID_visit_VisitID_FK FOREIGN KEY (`VisitID`) 
+  `ProjectSubprojectID` int(10) unsigned NOT NULL,
```
driusan: I find the name of this column confusing. Can we call it something like a "ProjectSubprojectPair"?

ridz1208: I dont like that
`ProjectSubprojectTuple` ??

driusan: I'm fine with "Tuple" too.

ridz1208: Made the change and added an amendment to the SQL modeling standards since that breaks the rules

https://github.com/aces/Loris/pull/4493/files#diff-190c4f51875cb53f0ebafdc6ae74b336R65

SQL/New_patches/2018-01-17-normalisation_visit.sql
```diff
@@ -1,25 +1,27 @@
+ALTER TABLE project_rel DROP PRIMARY KEY, ADD COLUMN `ProjectSubprojectTupleID` int(10) unsigned NOT NULL PRIMARY KEY AUTO_INCREMENT FIRST;
```
PapillonMcGill: @ridz1208 Have you tested this line?
Had to split in 2 alter table lines in the past (1-drop PK, 2- add column with PK)

docs/SQLModelingStandard.md
```diff
@@ -62,6 +62,8 @@ infrastructure will be re-designed to follow this guideline.*
   primary key should be `MyTableID`).
   - *i.e. the name of the primary identifier of the `country` table should be 
   `CountryID`.*
+> If a relational table requires a primary key ID (instead of the usual composite
```
PapillonMcGill: I don't see the need for this.  <TableNameID> should be enough. 
This modification should be in a separate PR.
It should also be below the next point if you want to complement it. You now have 2 contradicting points (Line 65 and 67)

SQL/New_patches/2018-01-17-normalisation_visit.sql
ridz1208: I tested it on mysql 5.7

I actually could not split it because mysql refuses to drop a primary key unless you specify another one in the same statement

docs/SQLModelingStandard.md
ridz1208: Without this statement, the name of my primary key does not respect this document. it should be `ProjectSubprojectRelID` but I have `ProjectSubprojectTupleID`

And I dont see the contradiction with line 67, this line says that rel tables specifically should name their keys as such. line 67 says that most tables should have an autoincrement primary key

PapillonMcGill: Sorry, contradict lines 59-64.

Does change my comment about should be in a separate PR. If we create exceptions in the standard every time someone doesn't like the standard, why have standards in the first place?

ridz1208: I see this less as a contradiction and more as a specification, line 59 applies to all tables EXCEPT rel tables.

And not sure I would say this is trying to bypass the standards, I see this as amending the standards to include a usecase we had not thought of before... so more like complementing it and standards are not set in stone they should be updated if need be.

As far as the suggestion itself initially made by @driusan I kind of agree with the fact that `ProjectSubprojectTupleID` is clearer than `ProjectSubprojectID` or `ProjectSubprojectRelID` so the amendment here would be actually improving our standards.

What do you think ??

PapillonMcGill: I don't see any 'EXCEPT' in the standard, so 'ProjectSubprojectRelID' is in line with the current standard.  If you want to modify/complement/improve the standard, it should be on a separate PR where people can study and comment appropriately.

ridz1208: So in this PR I add a new column called `ProjectSubprojectRelID` wait for that to get merged and then open another one changing it to `tuple` ?

PapillonMcGill: create a PR to modify the standard, once approved (if approved as nobody can predict the future, I surely can't) then use whatever the standard say.

ridz1208: ORRRRR just call it rel

I pick option number 2

ridz1208:
@PapillonMcGill I made the changes, please re-review carefully. due to the number of times changes were made it got messy

SQL/0000-00-00-schema.sql
```diff
@@ -24,9 +24,11 @@ INSERT INTO subproject (title, useEDC, WindowDifference) VALUES
   ('Experimental', false, 'optimal');
 
 CREATE TABLE `project_rel` (
+  `ProjectSubprojectRelID` int(10) unsigned NOT NULL AUTO_INCREMENT,
```
PapillonMcGill: @ridz1208 Shouldn't it be ProjectRelID?
unless you also want to rename the table to 'project_subproject_rel'.
don't forget to adjust PK def if needed and subsequent code.

SQL/New_patches/2018-01-17-normalisation_visit.sql
```diff
@@ -1,25 +1,27 @@
+ALTER TABLE project_rel DROP PRIMARY KEY, ADD COLUMN `ProjectSubprojectRelID` int(10) unsigned NOT NULL PRIMARY KEY AUTO_INCREMENT FIRST;
```
PapillonMcGill: some version of MySQL will complain.
```suggestion
ALTER TABLE project_rel DROP PRIMARY KEY;
ALTER TABLE project_rek ADD COLUMN `ProjectSubprojectRelID` int(10) unsigned NOT NULL PRIMARY KEY AUTO_INCREMENT FIRST;
```

SQL/0000-00-00-schema.sql
ridz1208: it's already done in another PR. I just didnt wanna do the changes twice or mix the code

SQL/New_patches/2018-01-17-normalisation_visit.sql
ridz1208: your suggestion gives me an error on mariadb.

driusan: What's the error? The deleted line (trying to drop and add in the same command) looks weird to me.

ridz1208:
```
Query execution failed

Reason:
SQL Error [1075] [42000]: (conn:228) Incorrect table definition; there can be only one auto column and it must be defined as a key
Query is : ALTER TABLE project_rel DROP PRIMARY KEY
```

SQL/0000-00-00-schema.sql
PapillonMcGill: Ok. reason accepted

SQL/New_patches/2018-01-17-normalisation_visit.sql
PapillonMcGill: The error is not with the DROP statement, it is in the order of the parameter of the ADD COLUMN.  
`PRIMARY KEY` as to be after the `AUTO_INCREMENT`.

I haven't check well the ordering in your instruction.

ridz1208: but if you see the error message. I get the error when I run the `ALTER TABLE project_rel DROP PRIMARY KEY` query

PapillonMcGill: Where you trying the entire drop and add in one statement?

Could you post the exact query and result here?

on my VM:
```
mysql> ALTER TABLE project_rel DROP PRIMARY KEY;
Query OK, 6 rows affected (0.46 sec)
Records: 6  Duplicates: 0  Warnings: 0
mysql> ALTER TABLE project_rel ADD COLUMN `ProjectSubprojectRelID` int(10) unsigned NOT NULL  AUTO_INCREMENT PRIMARY KEY FIRST;
Query OK, 0 rows affected (0.10 sec)
Records: 0  Duplicates: 0  Warnings: 0
```

ridz1208:
```
mysql> ALTER TABLE project_rel DROP PRIMARY KEY;
SQL Error [1075] [42000]: (conn:228) Incorrect table definition; there can be only one auto column and it must be defined as a key
```

PapillonMcGill: Can you post the current definition of you table?
```
SHOW CREATE TABLE project_rel
```

ridz1208: NEVERMINDDDDD MA BAD

PR #4496, opened 2019-05-03
Kill useProjects

ridz1208:
### Brief summary of changes
This gets rid of the `useProjects` configuration setting and all code sections referring to it. Projects will be required for all new studies. For existing studies, if the projects were not used and the `Project` table is empty a default project `loris` will be added and associated with all existing subprojects and all candidates already in the database.

Testing these changes requires extensive work and should be done mostly during the release testing phase. 

What should be tested just for this PR:
 - test that the SQL patch works as intended
   - if the `Project` table is populated, nothing should happen, nothing should be added.
   - if the `Project` table is empty, a new project should be added by default with `Name='loris'`, the `project_rel` table should be autopopulated with the `loris` project associated with every subproject fro mthe `subprojects` table and every candidate with `ProjectID=NULL` should automatically be reassigned to `loris` project.

**NOTE FOR PROJECTS:** Project admins should make sure all candidates in the database are affiliated with a project after running SQL patches. The default `loris` project will be added if none existed before and associated with all candidates. However if some projects already existed in the table, the `loris` default will not be added and thus candidates not affiliated with a project will need to be updated manually.


test/unittests/CandidateTest.php
```diff
@@ -492,6 +488,10 @@ private function _setUpTestDoublesForSelectCandidate()
             ->method('pselectRow')
             ->willReturn($this->_candidateInfo);
 
+        $this->_dbMock->expects($this->at(0))
+            ->method('pselect')
+            ->willReturn();
```
kongtiaowang: forget return _listOfProjects

ridz1208: OOOPS lol

kongtiaowang:
Screen Shot 2019-05-07 at 2 32 50 PM
The subproject is empty, even though we already setup subproject.

PR #4498, opened 2019-05-10
[Core/Instrument] added function loadInstanceDataFromArray

jesscall:
### Brief summary of changes
Added a new function that filters the instance data and returns only columns specified in the array that is passed. This can be useful in the case where we only want a subset of the data for an instrument as opposed to the full table. 

### To test this change...
- [ ] Open any instrument and test by passing to the function:  (a) The instrument instance, (b) An array of fields you want to select. 
            `$this->loadInstanceDataFromArray($this, $arrayofFields)`
- [ ] Print the results or replace a query

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2048,6 +2048,31 @@ class NDB_BVL_Instrument extends NDB_Page
     }
 
     /**
+     * Gets specified fields of data from an instrument out of the database and
+     * returns it as an array.
+     *
+     * @param NDB_BVL_Instrument $instrumentInstance The object whose data is
+     *                                               to be retrieved.
+     * @param array $fieldsArray Contains the fields of interest
+     *
+     * @return array containing the data for the fields specified in $fieldsArray
+     */
+    static function loadInstanceDataFromArray($instrumentInstance, $fieldsArray)
+    {
+        // Load instance data
+        $instanceData = NDB_BVL_Instrument::loadInstanceData($instrumentInstance);
+
+        // Filter results
+        $filteredData = array();
+        foreach ($fieldsArray as $key) {
+            if (array_key_exists($key, $instanceData)) {
+                $filteredData[$key] = $instanceData[$key];
+            }
+        }
+        return $filteredData;
+    }
+
+    /**
      * Calculates the candidate's age at the time of this instrument.
      * If Date_taken was based in, it will calculate the age as of
      * the date passed as a parameter (this is used, for instance, in
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2053,7 +2053,7 @@ class NDB_BVL_Instrument extends NDB_Page
      *
      * @param NDB_BVL_Instrument $instrumentInstance The object whose data is
      *                                               to be retrieved.
-     * @param array $fieldsArray Contains the fields of interest
+     * @param array              $fieldsArray        Contains the fields of interest
      *
      * @return array containing the data for the fields specified in $fieldsArray
      */


php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2047,6 +2047,31 @@ class NDB_BVL_Instrument extends NDB_Page
...
+     * Gets specified fields of data from an instrument out of the database and
+     * returns it as an array.
+     *
+     * @param NDB_BVL_Instrument $instrumentInstance The object whose data is
+     *                                               to be retrieved.
+     * @param array $fieldsArray Contains the fields of interest
+     *
+     * @return array containing the data for the fields specified in $fieldsArray
+     */
+    static function loadInstanceDataFromArray($instrumentInstance, $fieldsArray)
```
driusan: I don't see what you're gaining by this. You're loading all the data anyways, and then just discarding it.

jesscall: it would be to reduce duplication of code when a only subset of fields are needed especially in scoring functions.

jesscall: Also since the cost of a select is the same whether we fetch all fields or not, calling loadInstanceData also reduces the redundancy of having to check whether the instrument is using jsonData or not.

driusan: The cost of a SELECT isn't the same whether you fetch all fields or not. See https://blog.jooq.org/2017/03/08/many-sql-performance-problems-stem-from-unnecessary-mandatory-work/ .. I still don't see how this reduces code duplication. In what circumstances would this reduce duplication as opposed to just ignoring the fields you don't care about?

(re: the question about major/minor, the release branch is pushed into both of them now, so it's no longer relevant..)

ridz1208: @driusan in almost all our instruments that have scoring, we fetch the data saved in the table and use the data to run the score, except the score fields and the data are in the same table !! what we used to do is to select the specific fields necessary for scoring (ie `SELECT A,B,C,D ... FROM instruement`) and then run a for loop that does A + B + C + D.


In order tostart using the JsonData we cant run direct query databases anymore, instead we would need to load the instance data and either run through and unset the score fields and then calculate score OR in the same loop doing the scoring we add a `if(this field is not a scoring field) { do something}`

in either of the cases above that would be code that we would have to add to each and every scoring instrument. that function allows us to maintain status quo in a way where the code currently expecting a fixed array of fields would not have to be altered... the query would just be done using the new php function instead of a database query

whatcha think ??

driusan: I still don't understand.. whether you're using `loadInstanceData` or you're using `SELECT A, B, C, D` you're still doing `$score = $row['A'] + $row['B'] + $row['C'] + $row['D'];`. `$row` containing other items that aren't used isn't going to affect that.

ridz1208: when 50* fields contribute to the score and its the same in 20 other instruments, we dont do `$score = $row['A'] + $row['B'] + ...`, we do `SELECT A, B, C, D` and then `foreach ($rows as row) { sum this and that}`.

1. because that the way it's currently implemented and so changing that at this point is way more costly then just putting a wrapper around `loadInstanceData`
2. because we reuse the same array multiple times in different ways for different scoring algorithms... so might as well get the entire array just once and loop through

ridz1208: @driusan final thoughts ??

Either way we would copy the changes into the CCNA instrument extension classes if LORIS doesnt need it... so no waste or anything

driusan: I've never heard of anyone else doing SELECT A, B, C and then foreaching on the columns of the row (in most scoring algorithms I've seen it's more complex than that anyways since there are usually a couple columns that are reverse-coded or have a lookup or something.)

Do you know of any other projects doing that? If so, I'm not strictly opposed to it, it just seems unnecessary to me based on my experience.

jesscall: I agree with Rida, will most likely copy this over to CCNA instrument extensions if LORIS doesn't want it. Otherwise a huge refactor of scoring will be needed and I'm not willing to go  down that road...

PR #4499, opened 2019-05-10
[dataquery] Removed last instance of getJS.php

ridz1208:
### Brief summary of changes
GetJS.php no longer exists and throws a 404 error on current production environments

### To test this change...

- [ ] when attempting to download data fron the DQT as zip, you get a 404 without this PR


ridz1208:
@johnsaigle could you review ?

modules/dataquery/jsx/react.tabs.js
```diff
@@ -213,7 +213,7 @@ var ViewDataTabPane = React.createClass({
         window.dataBlobs = [];
 
         if(FileList.length < 100 || confirm("You are trying to download more than 100 files. This may be slow or crash your web browser.\n\nYou may want to consider splitting your query into more, smaller queries by defining more restrictive filters.\n\nPress OK to continue with attempting to download current files or cancel to abort." )) {
-            saveworker = new Worker(loris.BaseURL + '/GetJS.php?Module=dataquery&file=workers/savezip.js');
+            saveworker = new Worker(loris.BaseURL + '/dataquery/js/workers/savezip.js');
```
driusan: why is the compiled file above (can't comment on it) still part of this branch? Shouldn't it be generated by make?

ridz1208: This is going to bugfix on version 20.3.0, I'd rather have it in v20.3.1 given the difficulty for projects to go to 21 and the need for this to use the DQT

driusan:
This is a simple and obvious enough bug fix that I think we just need someone to test it, not review it..

johnsaigle:
@ridz1208 Code changes look fine but I don't really have any experience with DQT. Actually I don't even have it set up on my sandbox.

May be better to have someone else test this for you. Sorry about that.

PR #4507, opened 2019-05-13
[CSS] .btn class CSS misaligned

zaliqarosli:
any buttton using the .btn class is misaligned to the right. this PR is a quick fix for 21.0 before we revamp all of LORIS's css into flexbox anyways

### DIFF
--- a/htdocs/bootstrap/css/custom-css.css
+++ b/htdocs/bootstrap/css/custom-css.css
@@ -449,7 +449,7 @@ div.navbar div.container div.navbar-brand {
 
 .btn {
     font-size: 14px;
-    margin: 10px 5px;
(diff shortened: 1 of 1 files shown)


htdocs/bootstrap/css/custom-css.css
```diff
@@ -449,7 +449,7 @@ div.navbar div.container div.navbar-brand {
 
 .btn {
     font-size: 14px;
-    margin: 10px 5px;
+    margin: 10px 5px 10px 0;
```
zaliqarosli:
```suggestion
    margin-right: 5px;
```

zaliqarosli:
@johnsaigle this should fix your issue https://github.com/aces/Loris/issues/4557

Issue #4519, opened 2019-05-13
[acknowledgements] Start Date and End Date share same input when typing


maltheism:
The Start Date & End Date field apparently share the same input when user types.
https://github.com/aces/Loris/projects/19#card-21438166

zaliqarosli:
https://github.com/aces/Loris/pull/4528

jftunteng:
@maltheism  closing this issue since the PR is merged. Feel free to reopen if this is still an issue.

Issue #4509, opened 2019-05-13
[acknowledgements] End Date can be earlier than Start Date


maltheism:
When adding an acknowledgement the End Date can be earlier than Start Date. 
https://github.com/aces/Loris/projects/19#card-21438166

zaliqarosli:
https://github.com/aces/Loris/pull/4528

ridz1208:
@maltheism closing the issue since the PR has been merged. please reopen if issue is still occuring

PR #4528, opened 2019-05-13
[Acknowledgement] fix end date filtering

zaliqarosli:
### Brief summary of changes

- fixes issue https://github.com/aces/Loris/issues/4519
- fixes issue https://github.com/aces/Loris/issues/4509
- removes swal warning on exiting 'add acknowledgements' form if form is empty

### DIFF
--- a/modules/acknowledgements/jsx/acknowledgementsIndex.js
+++ b/modules/acknowledgements/jsx/acknowledgementsIndex.js
@@ -215,7 +215,7 @@ class AcknowledgementsIndex extends Component {
         title='Add Acknowledgement'
         onClose={this.closeModalForm}
         show={this.state.showModal}
-        throwWarning={(Object.keys(this.state.formData).length !== 0) && true}
+        throwWarning={(Object.keys(this.state.formData).length !== 0)}
       >
         <FormElement
           Module='acknowledgements'
--- a/modules/acknowledgements/jsx/acknowledgementsIndex.js
+++ b/modules/acknowledgements/jsx/acknowledgementsIndex.js
@@ -140,7 +140,6 @@ class AcknowledgementsIndex extends Component {
     .then((resp) => {
       if (resp.ok && resp.status === 200) {
         swal('Success!', 'Acknowledgement added.', 'success').then((result) => {
-          this.setState({formData: {}});
           if (result.value) {
             this.closeModalForm();
             this.fetchData();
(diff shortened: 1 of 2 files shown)


zaliqarosli:
@maltheism please review

modules/acknowledgements/jsx/acknowledgementsIndex.js
```diff
@@ -213,7 +215,7 @@ class AcknowledgementsIndex extends Component {
         title='Add Acknowledgement'
         onClose={this.closeModalForm}
         show={this.state.showModal}
-        throwWarning={true}
+        throwWarning={(Object.keys(this.state.formData).length !== 0) && true}
```
kongtiaowang: Why still keep the [&& true ] at this line?

modules/acknowledgements/jsx/acknowledgementsIndex.js
```diff
@@ -274,7 +276,7 @@ class AcknowledgementsIndex extends Component {
             name='addStartDate'
             label='Start date'
             value={this.state.formData.addStartDate}
-            maxYear={this.state.data.maxYear}
+            maxYear={this.state.formData.addEndDate || this.state.data.maxYear}
```
kongtiaowang: The maxYear comes from configuration settings.  This fix can set the year greater than maxYear.
the same issue for minYear.

modules/acknowledgements/jsx/acknowledgementsIndex.js
zaliqarosli: true haha i'll fix this!

zaliqarosli:
```suggestion
        throwWarning={(Object.keys(this.state.formData).length !== 0)}
```

modules/acknowledgements/jsx/acknowledgementsIndex.js
zaliqarosli: i'm not sure i understand what you mean. what this fix does is set the max year to be whatever you input into the 'end date' field (so that start date can't be after end date) if it exists. if you haven't put anything into the 'end date' field yet, the max year would be the config max year. does this not make sense for this form?

maltheism:
The fix looks good to me.

modules/acknowledgements/jsx/acknowledgementsIndex.js
maltheism: @kongtiaowang I think you're thinking about the new_profile module and this fix by @zaliqarosli is just to make sure the end_date is not before the start_date of the acknowledgements module.

zaliqarosli:
@maltheism thanks for reviewing! could you also add the passed manual test label if it does please? thank you :)

driusan:
@kongtiaowang are your requested changes still relevant?

Issue #4542, opened 2019-05-13
[data integrity flag] can't filter for user other than admin


zaliqarosli:

zaliqarosli:
https://github.com/aces/Loris/issues/4542

jesscall:
Instrument does not appear in table 
![image](https://user-images.githubusercontent.com/34260251/58193356-cb552300-7c90-11e9-93dc-370a5d7a95a8.png)

zaliqarosli:
@jesscall even after running ‘make’?

zaliqarosli:
PR merged

Issue #4541, opened 2019-05-13
[Data integrity flag] - link location with pre-set filters


zaliqarosli:
Number 4. of test plan - click on the instrument link takes you to data team helper with correct visit label filtered, but no instrument

zaliqarosli:
fix https://github.com/aces/Loris/pull/4548

zaliqarosli:
PR merged

PR #4548, opened 2019-05-13
[Data Integrity Flag] Add users option in filtering + fix link to datateamhelper

zaliqarosli:
fixes https://github.com/aces/Loris/issues/4542

### DIFF
--- a/modules/data_integrity_flag/php/data_integrity_flag.class.inc
+++ b/modules/data_integrity_flag/php/data_integrity_flag.class.inc
@@ -111,6 +111,8 @@ class Data_Integrity_Flag extends \NDB_Menu_Filter
      */
     function _setupVariables()
     {
+        $db = \NDB_Factory::singleton()->database();
+
         // set the class variables
         $this->columns = [
                           'dataflag_id',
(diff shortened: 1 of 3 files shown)


modules/data_integrity_flag/php/data_integrity_flag.class.inc
```diff
@@ -127,6 +129,17 @@ class Data_Integrity_Flag extends \NDB_Menu_Filter
...
+
+        $users         = array();
+        $selectedUsers = $db->pselect(
+            "SELECT u.UserID, u.Real_name FROM data_integrity_flag dif " .
+            "INNER JOIN users u ON (dif.dataflag_userid=u.UserID)",
+            array()
+        );
+        foreach ($selectedUsers as $row) {
+            $users[$row['UserID']] = $row['Real_name'];
+        }
```
driusan: seems like you're trying to do a `pselectWithIndexKey`?

ridz1208:
#4541

modules/data_integrity_flag/php/data_integrity_flag.class.inc
zaliqarosli: @driusan the problem with pselectWithIndexKey in this case is that i need to query with a 'SELECT DISTINCT'. do you prefer i do this or just turn the filter field from a select to a textbox?

PR #4566, opened 2019-05-14
fix syntax to drop foreign key in script

PapillonMcGill:
### To test this change...

run the New_patches/2019-04-26-Delete_cascades_parameter_type_rel.sql script on 20.0.0 database


driusan:
Is this MySQL version specific? How come it wasn't caught the first time/when it was written?

PapillonMcGill:
@driusan I tried to find the original PR, wasn't successful (didn't search for long). The equivalent code is not in the Release_Patches script. So I suspect that the review missed that and those 2 lines where never executed.

driusan:
Even without review, I don't think the person who wrote the patch could have made a typo that's that far off from the correct syntax unless it was database/version specific.

PapillonMcGill:
drop constraint exist in mariaDB latest
not it mysql 8.0

ridz1208:
I made these patches. they work on my VM

ridz1208:
https://mariadb.com/kb/en/library/alter-table/#drop-constraint

Issue #4533, opened 2019-05-13
[Candidate List] Cannot access the scan from candidate profile


charliehenrib:
When test point 11 in the test plan for the candidate list, the scan link is only going to imaging browser, not the specific scan itself.

coreyAbout:
Yes, I spot this one as well.

HenriRabalais:
The following PR should resolve this issue: #4590

Issue #4536, opened 2019-05-13
[JSX] The Filter Form is not adding query params to url. 


HenriRabalais:
When changing the values in the filter form, this is not reflected in the url query params. As a consequence, the filter form cannot be auto-loaded from another page redirect.

HenriRabalais:
PR Sent #4590

PR #4590, opened 2019-05-14
[JSX] Enabled Filter Query Params

HenriRabalais:
## Summary
Changes to the Filter are now reflected in the URL as query params. This allows the filter to be auto-loaded upon page load for all reactified menu filters. 

## Issue
#4536
#4533

## Testing
1. Go to a module that has a reactified menu Filter.
2. Change some of the values in the filter.
3. Make sure those changes are reflected as URL query params.
4. Refresh the page. Make sure that the query param values in the URL are populated in the filter.
5. Copy the URL to another browser window and make sure the query param values in the URL are populated in the filter.
6. Press 'Clear Filter' and make sure that the URL query params are also cleared.
7. Repeat 1-6 with another module.

### DIFF
--- a/jsx/Filter.js
+++ b/jsx/Filter.js
@@ -37,7 +37,6 @@ class Filter extends Component {
    * @param {string} type - type of the form element
    */
   onFieldUpdate(name, value, id, type) {
-    console.warn(value);
     const searchParams = new URLSearchParams(location.search);
     const filter = JSON.parse(JSON.stringify(this.props.filter));
     const exactMatch = type === 'textbox' ? false : true;
--- a/jsx/Filter.js
+++ b/jsx/Filter.js
@@ -17,6 +17,13 @@ class Filter extends Component {
     this.renderFilterFields = this.renderFilterFields.bind(this);
   }
 
+  componentDidMount() {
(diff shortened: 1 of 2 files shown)


johnsaigle:
This addresses https://github.com/aces/Loris/issues/4536?

HenriRabalais:
@johnsaigle Yes!

HenriRabalais:
@johnsaigle Nevermind! I think this addresses the issue #4536 entirely.

jesscall:
This is also supposed to resolve #4533.

HenriRabalais:
The URL query params can not currently handle multiselect filters. This needs to be addressed before the PR is merged.

HenriRabalais:
This PR is ready to be reviewed and tested!

jsx/Filter.js
```diff
@@ -37,15 +37,21 @@ class Filter extends Component {
    * @param {string} type - type of the form element
    */
   onFieldUpdate(name, value, id, type) {
+    console.warn(value);
```
driusan: Why would this be a warning?

HenriRabalais: Left over from debugging. I'll take it out.

Issue #4555, opened 2019-05-13
[statistics] link does not work under behavioural statistics tab


coreyAbout:
[TestPlan](https://github.com/aces/Loris/blob/minor/modules/statistics/test/TestPlan.md#behavioural-statistics) 
`6. Returning to the behavioural statistics page, check that the "Click here for breakdown per participant" link below the Data Entry Statistics table works.`
`11. Returning to the behavioral statistics tab, check that the "Click here for breakdown per participant for" link below the Double Data Entry Statistics table works.`

Problem:
"Click here for breakdown per participant" link does not work for both tables.

johnsaigle:
@ridz1208 Any chance you can post a stack trace here?

ridz1208:
```
[Tue May 14 15:52:03.562359 2019] [php7:error] [pid 28062] [client 132.216.56.93:34880] PHP Fatal error:  Uncaught TypeError: Argument 2 passed to User::hasCenterPermission() must be of the type integer, string given, called in /var/www/loris/modules/statistics/php/statistics_site.class.inc on line 47 and defined in /var/www/loris/php/libraries/User.class.inc:623\nStack trace:\n#0 /var/www/loris/modules/statistics/php/statistics_site.class.inc(47): User->hasCenterPermission('data_entry', '')\n#1 /var/www/loris/php/libraries/Module.class.inc(280): LORIS\\statistics\\Statistics_Site->_hasAccess(Object(User))\n#2 /var/www/loris/src/Middleware/ResponseGenerator.php(49): Module->handle(Object(Zend\\Diactoros\\ServerRequest))\n#3 /var/www/loris/src/Middleware/AuthMiddleware.php(62): LORIS\\Middleware\\ResponseGenerator->process(Object(Zend\\Diactoros\\ServerRequest), Object(LORIS\\statistics\\Module))\n#4 /var/www/loris/src/Router/ModuleRouter.php(97): LORIS\\Middleware\\AuthMiddleware->process(Object(Zend\\Diactoros\\ServerRequest), Object(LORIS\\statistics\\Module))\n#5 /var/www/loris/src/Router/BaseRouter.php(99): LORIS\\Router\\ModuleRoute in /var/www/loris/php/libraries/User.class.inc on line 623, referer: https://test.loris.ca/statistics/

```

johnsaigle:
This is addressed in the above PR

zaliqarosli:
PR merged

PR #4591, opened 2019-05-14
[Statistics] Fix type error causing "breakdown by participant" link to give 500 error

johnsaigle:
### Brief summary of changes

The `$centerID` variable was being passed as a string when it should've been an int. This caused a TypeError to be thrown.

Resolves https://github.com/aces/Loris/issues/4555

### DIFF
--- a/modules/statistics/php/statistics_site.class.inc
+++ b/modules/statistics/php/statistics_site.class.inc
@@ -53,7 +53,7 @@ class Statistics_Site extends \NDB_Menu
         if (!empty($_REQUEST['CenterID'])) {
             $hasCenterPermission = $user->hasCenterPermission(
                 'data_entry',
-                intval($centerID)
+                intval($_REQUEST['CenterID'])
             );
         }
 
--- a/modules/statistics/php/statistics_site.class.inc
+++ b/modules/statistics/php/statistics_site.class.inc
@@ -43,15 +43,19 @@ class Statistics_Site extends \NDB_Menu
         //TODO: Create a permission specific to statistics
         $hasAccessToAllProfiles = $user->hasAllPermissions(
             array(
-                'access_all_profiles',
-                'data_entry'
+             'access_all_profiles',
+             'data_entry',
             )
         );
         // If a CenterID is passed in the request, check if the user has the
         // data_entry permission at the site/center specified by CenterID.
-        $hasCenterPermission = !empty($_REQUEST['CenterID'] ?? '') ?
-            $user->hasCenterPermission('data_entry', intval($centerID))
-            : false;
+        $hasCenterPermission = false;
+        if (!empty($_REQUEST['CenterID'] ?? '')) {
+            $hasCenterPermission = $user->hasCenterPermission(
+                'data_entry',
+                intval($centerID)
+            );
+        }
 
         return $hasAccessToAllProfiles || $hasCenterPermission;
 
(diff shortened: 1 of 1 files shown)


modules/statistics/php/statistics_site.class.inc
```diff
@@ -40,12 +40,20 @@ class Statistics_Site extends \NDB_Menu
      */
     function _hasAccess(\User $user) : bool
     {
-        $centerID = htmlspecialchars($_REQUEST['CenterID']);
```
johnsaigle: Sanitization not necessary here because the value is not passed to the front-end.

modules/statistics/php/statistics_site.class.inc
```diff
@@ -40,12 +40,20 @@ class Statistics_Site extends \NDB_Menu
...
-            $accessAllProfiles;
+        $hasAccessToAllProfiles = $user->hasAllPermissions(
+            array(
+                'access_all_profiles',
+                'data_entry'
+            )
+        );
+        // If a CenterID is passed in the request, check if the user has the
+        // data_entry permission at the site/center specified by CenterID.
+        $hasCenterPermission = !empty($_REQUEST['CenterID'] ?? '') ?
```
johnsaigle: Using `$_REQUEST['CenterID'] ?? ''` fixes a PHP Notice when CenterID was not passed.

driusan: There's no reason to use a ternary operator here instead of a simple if statement, it obfuscates the logic for no gain.

modules/statistics/php/statistics_site.class.inc
```diff
@@ -40,12 +40,24 @@ class Statistics_Site extends \NDB_Menu
...
+        $hasAccessToAllProfiles = $user->hasAllPermissions(
+            array(
+             'access_all_profiles',
+             'data_entry',
+            )
+        );
+        // If a CenterID is passed in the request, check if the user has the
+        // data_entry permission at the site/center specified by CenterID.
+        $hasCenterPermission = false;
+        if (!empty($_REQUEST['CenterID'] ?? '')) {
```
driusan: I don't think the null coalesce is necessary here.. empty already handles null (and not set)

johnsaigle: It does but PHP will complain anyway (give a PHP Notice) if you try to access `$_REQUEST['CenterID']`when 'CenterID' isn't passed

driusan: I'm pretty sure that's only true outside of an `empty()` call. `empty` has the same properties as null coalesce about silencing unset warnings.

(I just double checked:

```
$ php --version
PHP 7.3.4-2 (cli) (built: Apr 13 2019 19:05:48) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.3.4, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.3.4-2, Copyright (c) 1999-2018, by Zend Technologies
$ php -a
Interactive mode enabled

php > ini_set('error_reporting', E_ALL | E_STRICT);
php > echo $foo['bar'];
PHP Notice:  Undefined variable: foo in php shell code on line 1
php > if (empty($foo['bar'])) { echo 'true'; } else { echo 'false'; }
true
php > ^D
```

)

johnsaigle: Alright I'll update this

kongtiaowang:
@johnsaigle 
For this Visit unit test, It should be rewritten later. You can mark it as skip to pass Travis.

driusan:
@johnsaigle Travis

kongtiaowang:
modules/statistics/php/statistics_site.class.inc:56 PhanUndeclaredVariable Variable $centerID is undeclared
[debug] Restarted process exited 1
make: *** [checkstatic] Error 1

driusan:
still looks like it's failing

johnsaigle:
@driusan It's failing cause of the testPSCIDlink test. I'll rebase and it should be fine after

Issue #4353, opened 2019-02-20
[Reliability] module Class missing


PapillonMcGill:
Blank page on the screen and following error in log:

[Wed Feb 20 09:07:08.559584 2019] [php7:error] [pid 3084] [client 192.168.122.1:36190] PHP Fatal error:  Uncaught LorisModuleMissingException: reliability is missing Module class in /var/www/Loris/php/libraries/Module.class.inc:75\nStack trace:\n#0 /var/www/Loris/src/Router/BaseRouter.php(90): Module::factory('reliability')\n#1 /var/www/Loris/src/Middleware/ResponseGenerator.php(49): LORIS\\Router\\BaseRouter->handle(Object(Zend\\Diactoros\\ServerRequest))\n#2 /var/www/Loris/src/Middleware/ContentLength.php(51): LORIS\\Middleware\\ResponseGenerator->process(Object(Zend\\Diactoros\\ServerRequest), Object(LORIS\\Router\\BaseRouter))\n#3 /var/www/Loris/htdocs/index.php(47): LORIS\\Middleware\\ContentLength->process(Object(Zend\\Diactoros\\ServerRequest), Object(LORIS\\Router\\BaseRouter))\n#4 {main}\n  thrown in /var/www/Loris/php/libraries/Module.class.inc on line 75, referer: https://cbigr-dev.loris.ca/candidate_list/

TO REPRODUCE
install latest major (2019-02-20),
login and click on Clinical -> Reliability menu

driusan:
Reliability was deleted in #3886 ?

PapillonMcGill:
Then we should remove it from the menu

driusan:
But #3886 also removed it from the menu..

PapillonMcGill:
It just remove the creation of the menu for new install but there is not removal from existing instance.
reliability table and entry in LorisMenu are still present.

@driusan do you want me to create a PR to remove them?

johnsaigle:
@PapillonMcGill @driusan Was Reliability ever removed from the menu in a SQL patch? This should be done for the 21.0.0 release if not.

PapillonMcGill:
@johnsaigle  Nope, wasn't removed in major when I created this issues.

PR #4596, opened 2019-05-15
Remove Menu link for reliability

ridz1208:
### Brief summary of changes
Module was deleted

### This resolves issue...

- [ ] Github? #4353

### DIFF
--- a/SQL/Cleanup_patches/2019-05-15-Reliability_link.sql
+++ b/SQL/Cleanup_patches/2019-05-15-Reliability_link.sql
@@ -1,0 +1,2 @@
+DELETE FROM LorisMenu WHERE Link='reliability/';
+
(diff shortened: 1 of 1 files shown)


SQL/Cleanup_patches/2019-05-15-Reliability_link.sql
```diff
@@ -0,0 +1,2 @@
+DELETE FROM LorisMenu WHERE Link='reliability/';
```
driusan: Should this one be done based on the label? If anyone has Reliability in their menu it's likely to be old enough to be an old test_name= style link.

ridz1208: @driusan I put both in case someone had renamed the menu item

Issue #4595, opened 2019-05-15
[SQL] inexistant instruction in MySQL


PapillonMcGill:
in script 2019-05-03-deprecate_useprojects.sql, there is a clause `WHERE NOT EXISTS` that is non-existant in MySQL

driusan:
This doesn't make sense. `WHERE (NOT) EXISTS` is completely standard SQL going at least as far back as SQL92. The oldest MySQL documentation that I can find on their website supports it: https://dev.mysql.com/doc/refman/5.5/en/exists-and-not-exists-subqueries.html as well as the newest version: https://dev.mysql.com/doc/refman/8.0/en/exists-and-not-exists-subqueries.html and

PapillonMcGill:
@driusan Have you tried it with an INSERT?  it is Ok with SELECT, but fail with INSERT (at least in my VM MySQL 5.5)

driusan:
Subqueries and WHERE clauses syntax do not change between SELECTs and INSERTs. If you're trying it and it's failing, there's some reason other than the `WHERE NOT EXISTS`. We shouldn't be changing patches from ANSI standard SQL92 to MySQL-only language extensions (INSERT IGNORE) because of invalid assumptions.

ridz1208:
@PapillonMcGill can you post the error message here ?

PapillonMcGill:
@ridz1208 mysql> INSERT INTO Project (Name) SELECT 'loris' WHERE NOT EXISTS (SELECT * FROM Project);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'WHERE NOT EXISTS (SELECT * FROM Project)' at line 1

@driusan 
from MySQL doc https://dev.mysql.com/doc/refman/5.5/en/exists-and-not-exists-subqueries.html:
SELECT column1 FROM t1 WHERE EXISTS (SELECT * FROM t2);

from Mariadb  https://mariadb.com/kb/en/library/subqueries-and-exists/ :1234: 
Syntax

SELECT ... WHERE EXISTS <Table subquery>

In this case, it is not available with an INSERT statement

driusan:
[...]

PR #4597, opened 2019-05-15
Fix INSERT ... WHERE EXIST

PapillonMcGill:
### Brief summary of changes
select from DUAL table as  we can't use WHERE clause in a SELECT without FROM table

### This resolves issue...
Resolves https://github.com/aces/Loris/issues/4595

### To test this change...

run the SQL/New_patches/2019-05-03-deprecate_useprojects.sql script on a version 20 database

### DIFF
--- a/SQL/New_patches/2019-05-03-deprecate_useprojects.sql
+++ b/SQL/New_patches/2019-05-03-deprecate_useprojects.sql
@@ -6,10 +6,10 @@ SELECT 'loris'
 WHERE NOT EXISTS (SELECT * FROM Project);
 
 -- associate all subprojects to the loris project by default if project table was empty
-INSERT INTO project_rel (ProjectID,SubprojectID)
-SELECT ProjectID, SubprojectID 
-FROM Project JOIN subproject 
-WHERE Project.Name='loris';
+INSERT IGNORE INTO Project (ProjectID, Name)
+  (SELECT * FROM
+     (SELECT min(ProjectID), 'loris' FROM Project AS q1)
+  AS q2);
 
 -- if the loris project was added, set all candidates to that default project
 UPDATE candidate SET ProjectID=(SELECT ProjectID FROM Project WHERE Name='loris') WHERE ProjectID IS NULL AND Entity_type='Human';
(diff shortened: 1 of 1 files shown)


SQL/New_patches/2019-05-03-deprecate_useprojects.sql
```diff
@@ -6,10 +6,10 @@ SELECT 'loris'
 WHERE NOT EXISTS (SELECT * FROM Project);
 
 -- associate all subprojects to the loris project by default if project table was empty
-INSERT INTO project_rel (ProjectID,SubprojectID)
-SELECT ProjectID, SubprojectID 
-FROM Project JOIN subproject 
-WHERE Project.Name='loris';
+INSERT IGNORE INTO Project (ProjectID, Name)
```
ridz1208: @PapillonMcGill you replaced the wrong query

PapillonMcGill: fixed!

PapillonMcGill:
@ridz1208  Could you re-review

SQL/New_patches/2019-05-03-deprecate_useprojects.sql
```diff
@@ -2,8 +2,8 @@ DELETE cs, c FROM ConfigSettings cs JOIN Config c ON c.ConfigID=cs.ID WHERE cs.N
 
 -- if Project table is empty, add default 'loris' project
 INSERT INTO Project (Name)
-SELECT 'loris'
-WHERE NOT EXISTS (SELECT * FROM Project);
+(SELECT 'loris' FROM DUAL
```
ridz1208: @PapillonMcGill what is DUAL ?

driusan: it's a fake table with exactly one row (it originates from Oracle, but the SQL standard doesn't really have any useable equivalent for cases like this and MySQL and MariaDB seem to both support it)

ridz1208: lol... learning stuff everyday

Issue #4502, opened 2019-05-13
[Login] request-account Undefined index: radiologist


maltheism:
The bug occurs on a fresh install of loris version 21.0.0.
Steps to reproduce: Login->request-account->complete form with checkbox examiner role and submit the request account.
https://github.com/aces/Loris/projects/19#card-21296846

Output:

```
Notice: Undefined index: radiologist in /Users/alizee/Development/GitHub/McGill/Loris/modules/login/php/requestaccount.class.inc on line 175
--


1 | 0.0012 | 407504 | {main}( ) | .../index.php:0
2 | 0.0069 | 700200 | LORIS\Middleware\ContentLength->process( ) | .../index.php:47
3 | 0.0069 | 700200 | LORIS\Middleware\ResponseGenerator->process( ) | .../ContentLength.php:51
4 | 0.0069 | 700200 | LORIS\Router\BaseRouter->handle( ) | .../ResponseGenerator.php:49
5 | 0.0078 | 712736 | LORIS\Router\ModuleRouter->handle( ) | .../BaseRouter.php:99
6 | 0.0078 | 712736 | LORIS\login\Module->handle( ) | .../ModuleRouter.php:100
7 | 0.0125 | 816184 | LORIS\login\RequestAccount->process( ) | .../Module.class.inc:341
8 | 0.0125 | 817576 | LORIS\Middleware\PageDecorationMiddleware->process( ) | .../NDB_Page.class.inc:653
9 | 0.0151 | 820784 | LORIS\Middleware\AnonymousPageDecorationMiddleware->process( ) | .../PageDecorationMiddleware.php:38
10 | 0.0162 | 823920 | LORIS\login\RequestAccount->handle( ) | .../AnonymousPageDecorationMiddleware.php:55
11 | 0.0172 | 895496 | LORIS\login\RequestAccount->save( ) | .../NDB_Form.class.inc:190
12 | 0.0240 | 895776 | LorisForm->process( ) | .../NDB_Form.class.inc:133
13 | 0.0241 | 896952 | LORIS\login\RequestAccount->_save( ) | .../LorisForm.class.inc:1439
14 | 0.0241 | 896952 | LORIS\login\RequestAccount->_process( ) | .../NDB_Form.class.inc:162
```

PR #4598, opened 2019-05-15
[login] Fix login Error log notice for examiner/radiologist

ridz1208:
### Brief summary of changes
Checkboxes are not submitted when they are not checked and thus they are not set in the `$_REQUEST` array(). added `isset()` checks to avoid warnings

### This resolves issue...

- [ ] Github? #4502

### To test this change...

- [ ] check if notices are generated when boxes are checked/uncheccked and the form is submitted

### DIFF
--- a/modules/login/php/requestaccount.class.inc
+++ b/modules/login/php/requestaccount.class.inc
@@ -170,9 +170,11 @@ class RequestAccount extends \NDB_Form
                 )
             );
 
-            if ($_REQUEST['examiner'] == 'on') {
+            if (isset($_REQUEST['examiner']) && $_REQUEST['examiner'] == 'on') {
                 $rad = 0;
-                if ($_REQUEST['radiologist'] == 'on') {
+                if (isset($_REQUEST['radiologist'])
+                    && $_REQUEST['radiologist'] == 'on'
+                ) {
                     $rad = 1;
                 }
                 //insert in DB as inactive until account approved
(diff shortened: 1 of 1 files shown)


modules/login/php/requestaccount.class.inc
```diff
@@ -170,9 +170,11 @@ class RequestAccount extends \NDB_Form
                 )
             );
 
-            if ($_REQUEST['examiner'] == 'on') {
+            if (isset($_REQUEST['examiner']) && $_REQUEST['examiner'] == 'on') {
```
driusan:
```suggestion
            if (($_REQUEST['examiner'] ?? 'off') === 'on') {
```

Probably cleaner to use null coalesce to coalesce onto a default value. (same with below, but that one's spread across multiple lines so I can't recommend changes..)

ridz1208: I dont like it with null coalesce inside the if

if you want I can define them in a variable above

driusan: sure, use a variable. Either is cleaner than isset, I think.

driusan:
looks good other than Travis.

johnsaigle:
> FILE: ...avis/build/aces/Loris/modules/login/php/requestaccount.class.inc
 172 | ERROR | [x] Whitespace found at end of line
 173 | ERROR | [x] Equals sign not aligned with surrounding
     |       |     assignments; expected 4 spaces but found 1 space
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY

ridz1208:
@johnsaigle 3rd time is the charm

Issue #4611, opened 2019-05-15
[Acknowledgements] Warning: Failed prop type: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.


maltheism:
https://github.com/aces/Loris/projects/19#card-21438166
The FilteredDataTable needs to be fixed for changing Maximum rows to display for the table.

Error:
`Warning: Failed prop type: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.`

zaliqarosli:
PR merged

PR #4612, opened 2019-05-15
[acknowledgements] [FilteredDataTable] onChange handler for Maximum rows per page

maltheism:
### Brief summary of changes

Fixes the console react error for the FilteredDataTable Maximum rows to display for the table.

No longer will this error output:
`Warning: Failed prop type: You provided a valueprop to a form field without anonChangehandler. This will render a read-only field. If the field should be mutable usedefaultValue. Otherwise, set either onChangeorreadOnly.`

### This resolves issue...

- [ ] Redmine? #

- [x] Github? #4611

### To test this change...

- [x] Go to the acknowledgements module or any module that uses the FilteredDataTable

### DIFF
--- a/jsx/DataTable.js
+++ b/jsx/DataTable.js
@@ -467,7 +467,7 @@ class DataTable extends Component {
     let rowsPerPageDropdown = (
       <select
         className="input-sm perPage"
-        onChange={this.props.updatePageRows}
+        onChange={this.updatePageRows}
         value={this.state.page.rows}
       >
         <option>20</option>


zaliqarosli:
hey @maltheism i can't reproduce this error. how are you getting it and on which browser? i am changing the maximum rows per page option but i get no error in my console log

zaliqarosli:
this branch is changing the UI for some reason :/ putting blocked so i can investigate

zaliqarosli:
figured it out.. this PR isn't fixing what's on test.loris.ca but instead fixes errors merged into `21.0-release` from https://github.com/aces/Loris/pull/4112 i think

zaliqarosli:
issue https://github.com/aces/Loris/issues/4626 issued

Issue #4559, opened 2019-05-13
Data Dictionary untestable


ridz1208:
the data dictionary builder was not run on the test vm

jftunteng:
@ridz1208 closing this issue since the PR is merged. Feel free to reopen if this is still an issue or if it reoccurs.

Issue #4537, opened 2019-05-13
[Data Release] Missing data_release_edit_file_access permission


Jkat:
Possibly missing from permissions table and/or `admin` user, therefore unable to assign to regular user

ridz1208:
resolved

PR #4614, opened 2019-05-15
Update RB to 21

ridz1208:
### Brief summary of changes
- Fix instruments for 21
- Build the datadictionnary
- Source unsourced Patches

### This resolves issue...

- [ ] Github? 
  - #4537
  - #4559
  - #4630


raisinbread/RB_files/RB_LorisMenu.sql
```diff
@@ -12,7 +12,6 @@ INSERT INTO `LorisMenu` (`ID`, `Parent`, `Label`, `Link`, `Visible`, `OrderNumbe
 INSERT INTO `LorisMenu` (`ID`, `Parent`, `Label`, `Link`, `Visible`, `OrderNumber`) VALUES (9,2,'Reliability','reliability/',NULL,1);
 INSERT INTO `LorisMenu` (`ID`, `Parent`, `Label`, `Link`, `Visible`, `OrderNumber`) VALUES (10,2,'Conflict Resolver','conflict_resolver/',NULL,2);
 INSERT INTO `LorisMenu` (`ID`, `Parent`, `Label`, `Link`, `Visible`, `OrderNumber`) VALUES (11,2,'Examiner','examiner/',NULL,3);
-INSERT INTO `LorisMenu` (`ID`, `Parent`, `Label`, `Link`, `Visible`, `OrderNumber`) VALUES (12,2,'Training','training/',NULL,4);
 INSERT INTO `LorisMenu` (`ID`, `Parent`, `Label`, `Link`, `Visible`, `OrderNumber`) VALUES (13,2,'Media','media/',NULL,5);
 INSERT INTO `LorisMenu` (`ID`, `Parent`, `Label`, `Link`, `Visible`, `OrderNumber`) VALUES (14,3,'DICOM Archive','dicom_archive/',NULL,1);
```
johnsaigle: The `ID` value now skips from 11 to 13.

ridz1208: thats fine

raisinbread/RB_files/RB_parameter_type.sql
```diff
@@ -12,5 +12,188 @@ INSERT INTO `parameter_type` (`ParameterTypeID`, `Name`, `Type`, `Description`,
 INSERT INTO `parameter_type` (`ParameterTypeID`, `Name`, `Type`, `Description`, `RangeMin`, `RangeMax`, `SourceField`, `SourceFrom`, `SourceCondition`, `Queryable`, `IsFile`) VALUES (10,'candidate_label','text','Identifier_of_candidate',NULL,NULL,'PSCID','candidate',NULL,1,0);
```
driusan: I can't comment on the history file above because the diff is too large.. but is adding the whole history table to RB really necessary if it wasn't there before?

johnsaigle: I was also wondering about this. I don't think we need it.

ridz1208: I usually dont commit the history changes
it happened accidentally here since it got modified by the data_dictionary_builder

I dont see the issue with it, but i can remove it if need be

raisinbread/RB_files/RB_ConfigSettings.sql
```diff
@@ -96,5 +96,6 @@ INSERT INTO `ConfigSettings` (`ID`, `Name`, `Description`, `Visible`, `AllowMult
 INSERT INTO `ConfigSettings` (`ID`, `Name`, `Description`, `Visible`, `AllowMultiple`, `DataType`, `Parent`, `Label`, `OrderNumber`) VALUES (97,'publication_uploads','Path to uploaded publications',1,0,'text',26,'Publications',11);
 INSERT INTO `ConfigSettings` (`ID`, `Name`, `Description`, `Visible`, `AllowMultiple`, `DataType`, `Parent`, `Label`, `OrderNumber`) VALUES (98,'publication_deletions','Path to deleted publications',1,0,'text',26,'Deleted Publications',11);
 INSERT INTO `ConfigSettings` (`ID`, `Name`, `Description`, `Visible`, `AllowMultiple`, `DataType`, `Parent`, `Label`, `OrderNumber`) VALUES (99,'usePwnedPasswordsAPI','Whether to query the Have I Been Pwned password API on password changes to prevent the usage of common and breached passwords',1,0,'boolean',1,'Enable \"Pwned Password\" check',22);
+INSERT INTO `ConfigSettings` (`ID`, `Name`, `Description`, `Visible`, `AllowMultiple`, `DataType`, `Parent`, `Label`, `OrderNumber`) VALUES (100,'EnvironmentFile','Name of the environment file that need to be sourced for the imaging pipeline',1,0,'text',69,'Name of the environment file',20);
```
christinerogers: This ideally should not be ConfigSetting: ID=`96` in RaisinBread. It should be right after the `MRIConfigFile` setting.  We'll have a hard time checking later if all the proper config settings are in place if they're out of order with the default schema.

raisinbread/RB_files/RB_LorisMenu.sql
driusan: Down with non-prime numbers!

raisinbread/RB_files/RB_ConfigSettings.sql
ridz1208: RB data is handled like any project. patches are run on it and the data is submitted I never handle the data manually and if tomorrow someone decides to reorder the config settings in the 0000 files then they will go out of sync with any project out there.

If you would like to make a patch to make sure they are in sync then it can go in the New patches directory and can be run on all projects at release... I dont mind that.

driusan: I don't feel strongly about this one way or another but I think this change request is the only point stopping this PR right now.. @christinerogers can you respond to @ridz1208 or dismiss the review?

christinerogers: The issue is that the 20.2 upgrade patch was incorrectly run on RB, which caused this to happen. 
That's why we should fix it, totally understanding that other projects out there will have variance anyway from the default schema. 
I'm not in a position to make a patch so... will dismiss.

PR #4621, opened 2019-05-15
[TimePoint] Allow getApprovalStatus function to return null

### LINKED ISSUE #4585: Timepoint_list typing error when trying to access module
Only happens on a subset of candidates i.e. DCC090 after having completed all the instruments in V1

> [Tue May 14 15:17:56.435449 2019] [php7:error] [pid 28062] [client 132.216.56.251:53587] PHP Fatal error:  Uncaught TypeError: Return value of TimePoint::getApprovalStatus() must be of the type string, null returned in /var/www/loris/php/libraries/TimePoint.class.inc:495\nStack trace:\n#0 /var/www/loris/php/libraries/TimePoint.class.inc(618): TimePoint->getApprovalStatus()\n#1 /var/www/loris/modules/timepoint_list/php/timepoint_list.class.inc(146): TimePoint->getCurrentStatus()\n#2 /var/www/loris/php/libraries/NDB_Page.class.inc(668): LORIS\\timepoint_list\\Timepoint_List->setup()\n#3 /var/www/loris/src/Middleware/UserPageDecorationMiddleware.php(171): NDB_Page->handle(Object(Zend\\Diactoros\\ServerRequest))\n#4 /var/www/loris/src/Middleware/PageDecorationMiddleware.php(49): LORIS\\Middleware\\UserPageDecorationMiddleware->process(Object(Zend\\Diactoros\\ServerRequest), Object(LORIS\\timepoint_list\\Timepoint_List))\n#5 /var/www/loris/php/libraries/NDB_Page.class.inc(653): LORIS\\Middleware\\PageDecorationMiddleware->process(Object(Zend\\Diactoros\\ServerRequest), Object(LORIS\\timepoint_list\\Timepoint_List))\n#6 /var/www/loris/modules/ in /var/www/loris/php/libraries/TimePoint.class.inc on line 495, referer: https://test.loris.ca//candidate_list

johnsaigle:
### Brief summary of changes

This variable can be null in certain cases. The `sessions` table in MySQL which stores this value has a `NULL` value as the default, so it follows that this value can be `null` within PHP.

Resolves #4585

### DIFF
--- a/php/libraries/TimePoint.class.inc
+++ b/php/libraries/TimePoint.class.inc
@@ -490,9 +490,9 @@ class TimePoint
     /**
      * Return the approval status of this timepoint.
      *
-     * @return string 'In Progress', 'Pass' or 'Failure'
+     * @return string|null 'In Progress', 'Pass' or 'Failure'
      */
-    function getApprovalStatus(): string
+    function getApprovalStatus(): ?string
     {
         return $this->_timePointInfo["Approval"];
     }


johnsaigle:
@ridz1208 Could you test whether this issue fixes the issue you reported? Or if you could provide testing instructions that would help a lot.

ridz1208:
@johnsaigle it does work but it is not the only error occurring.

Testing is difficult to describe because it only happens in specific circumstances. I think a way to reproduce is to comkplete all the different instruments in a timepoint and try to set it to send to dcc

Issue #4626, opened 2019-05-15
Issue Tracker PR breaks 21.0-release


zaliqarosli:
#4112 changed things that it shouldn't have on `21.0-release`, causing errors that don't currently exist on test.loris.ca

1. Fix DataTable UI @zaliqarosli https://github.com/aces/Loris/pull/4629 - merged
2. Fix pagination functionality @HenriRabalais https://github.com/aces/Loris/pull/4632 - merged
3. Fix maximum number of rows option error @maltheism https://github.com/aces/Loris/pull/4612 - merged

HenriRabalais:
Pagination Bug Fix PR has been merged #4632

zaliqarosli:
all PRs merged

PR #4629, opened 2019-05-15
[jsx/DataTable] Revert UI changes made in issuetracker PR

zaliqarosli:
fixes 1 part of https://github.com/aces/Loris/issues/4626

### DIFF
--- a/jsx/DataTable.js
+++ b/jsx/DataTable.js
@@ -486,7 +486,7 @@ class DataTable extends Component {
     }
 
     let header = this.props.hide.rowsPerPage === true ? '' : (
-      <div className="table-header panel-heading">
+      <div className="table-header">
         <div className="row">
           <div style={{
             display: 'flex',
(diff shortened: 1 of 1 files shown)


driusan:
I don't understand this issue or the fix enough to write a commit message for this. Why is this reversion required? Why does the download as CSV feature need to be removed to fix.. something?

zaliqarosli:
the IssueTracker PR had some old DataTable UI code in it that got merged. this PR just replaces that old UI with the new UI that was merged in a previous PR. soo the IssueTracker PR reverted UI changes. this PR reintroduces new UI that got removed

zaliqarosli:
i.e. we had decided to remove the header and footer panels for the datatable

driusan:
So why does the download table as CSV button get removed in the diff for this PR?

zaliqarosli:
so that its only above the table with the rest of the "calls to actions". this change was made in line with talks in UX group regarding call to actions and their placement

Issue #3948, opened 2018-09-11
Filtering breaks while *not* on first page of table


zaliqarosli:
You are scrolling through the data table and then decide to filter while on the 2nd or 3rd page of the data. The filtering doesn't work. It only works while you're on the first page of the data table.

<img width="1278" alt="screenshot 2018-09-11 11 57 48" src="https://user-images.githubusercontent.com/17415878/45372153-43276580-b5ba-11e8-877a-f72a24a00b3d.png">
<img width="1280" alt="screenshot 2018-09-11 11 57 53" src="https://user-images.githubusercontent.com/17415878/45372166-49b5dd00-b5ba-11e8-8de2-1551e3b76006.png">

driusan:
Is this across all modules or only the media module?

zaliqarosli:
@driusan i believe all reactified menu filters

HenriRabalais:
PR sent #4632

zaliqarosli:
PR merged

PR #4632, opened 2019-05-15
[JSX] Fixed Datatable Pagination Bugs

HenriRabalais:
## Summary
This PR Kills 2 bugs with one stone.
1. Pagination works again.
2. When the user is not on the first page of the datatable and alters the number of rows in the datatable via the filter such that the new number of rows is less than the previous number of rows, the user is returned to the first page of the datatable.

## Issues
1. #4626
2. #3948

## Testing
1. Go to your favourite reactified module.
2. Make sure pagination works.
3. Change one of the filter values
4. Make sure that you are returned to the first page of the datatable.
5. Repeat steps 1-4 for another reactified module.

### DIFF
--- a/jsx/DataTable.js
+++ b/jsx/DataTable.js
@@ -21,6 +21,7 @@ class DataTable extends Component {
       },
     };
 
+    this.changePage = this.changePage.bind(this);
     this.setSortColumn = this.setSortColumn.bind(this);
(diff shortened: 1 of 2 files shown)


driusan:
In case number 2, wouldn't it make more sense to go to the last page of the results instead of the first?

PR #4633, opened 2019-05-15
[JSX] Fixed Bug That Revealed Hidden Headers

HenriRabalais:
## Summary
Made a slight change so that Hidden Headers remain hidden in non-reactified modules.

## Issues
#4634

## Testing
1. Go to a non-reactified module (e.g. MRI Violations, etc.) and make sure there are no extra columns at the end of the datatable with empty rows.
2. Test another non-reactified module.

### DIFF
--- a/jsx/StaticDataTable.js
+++ b/jsx/StaticDataTable.js
@@ -356,8 +356,8 @@ class StaticDataTable extends Component {
     ];
 
     for (let i = 0; i < this.props.Headers.length; i += 1) {
-      if (typeof this.props.hiddenHeaders === 'undefined' ||
-        this.props.hiddenHeaders.indexOf(this.props.Headers[i]) === -1) {
+      if (typeof loris.hiddenHeaders === 'undefined' ||
+        loris.hiddenHeaders.indexOf(this.props.Headers[i]) === -1) {
         let colIndex = i + 1;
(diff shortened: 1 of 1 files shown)


driusan:
@HenriRabalais failing Travis.

kongtiaowang:
This PR fixed the hidden header issue, But it also made media table out of boundary. 
Screen Shot 2019-05-16 at 11 42 04 AM
Screen Shot 2019-05-16 at 11 43 14 AM

HenriRabalais:
@kongtiaowang That is very strange! I check the other reactified menu filters and it does the same thing when shrinking the width of the screen. I really have no idea why this is happening since none of the reactified modules are using the `StaticDataTable.js` component. I'm a bit perplexed.

In my opinion, I would issue another bug. This should all be resolved once we reactify all the menu filters and can delete the `StaticDataTable` component.

maltheism:
@HenriRabalais I just tested this PR on Mri Violations and the solution is working for my installation of LORIS. I even tested 21.0-release and I'm not getting an overflow but the headers are indeed visible on 21.0-release contrary to this PR.

driusan:
I think it's more important to hide the hidden headers than it is to fix the overflow issue, so I'm going to merge this and create a ticket (for the release) about the "out of bounds".

Issue #4511, opened 2019-05-13
[Login] password-reset submit without filling out form


maltheism:
The password reset form can be submitted without a username or email.
https://github.com/aces/Loris/projects/19#card-21296846

zaliqarosli:
PR https://github.com/aces/Loris/pull/4638

driusan:
Fixed by the above PR

PR #4638, opened 2019-05-16
[Login] Add required to password-reset field

zaliqarosli:
fixes https://github.com/aces/Loris/issues/4511


kchatpar:
Looks good to me.

Issue #4581, opened 2019-05-14
[instrument] Trying to set instrument to 'Complete' throws 500 error


leighevans:
Trying to set any instrument to 'Complete' results in a 500 error

ridz1208:
```
[Tue May 14 15:31:28.863179 2019] [php7:error] [pid 28062] [client 132.216.56.251:53699] PHP Fatal error:  Uncaught TypeError: Argument 3 passed to NDB_BVL_Instrument::factory() must be of the type string, null given, called in /var/www/loris/php/libraries/ConflictDetector.class.inc on line 62 and defined in /var/www/loris/php/libraries/NDB_BVL_Instrument.class.inc:173\nStack trace:\n#0 /var/www/loris/php/libraries/ConflictDetector.class.inc(62): NDB_BVL_Instrument::factory('mri_parameter_f...', '587630DCC090105...', NULL)\n#1 /var/www/loris/php/libraries/NDB_BVL_InstrumentStatus.class.inc(209): ConflictDetector::detectConflictsForCommentIds('mri_parameter_f...', '587630DCC090105...', 'DDE_587630DCC09...')\n#2 /var/www/loris/php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc(166): NDB_BVL_InstrumentStatus->setDataEntryStatus('Complete')\n#3 /var/www/loris/php/libraries/NDB_BVL_Instrument.class.inc(2238): NDB_BVL_InstrumentStatus_ControlPanel->save()\n#4 /var/www/loris/php/libraries/NDB_Caller.class.inc(417): NDB_BVL_Instrument->getControlPanel()\n#5 /var/www/loris/php/libraries/NDB_Caller.class.inc(250): NDB_C in /var/www/loris/php/libraries/NDB_BVL_Instrument.class.inc on line 173, referer: https://test.loris.ca/instruments/mri_parameter_form/?commentID=587630DCC0901053141524238782&sessionID=1053&candID=587630

```

zaliqarosli:
@ridz1208 @leighevans was MRI parameter form working before? because it's not now

zaliqarosli:
PR fix https://github.com/aces/Loris/pull/4640

driusan:
PR merged

PR #4640, opened 2019-05-16
[Instrument] Change null to empty string in ConflictDetector

zaliqarosli:
fixes https://github.com/aces/Loris/issues/4581


zaliqarosli:
@maltheism could you please add the passed manual test tag as well please? thanks for reviewing

Issue #4553, opened 2019-05-13
[Dicom Archive] Clicking on 'View Details' Returns Error.


HenriRabalais:
Item 5 of the Test Plan:
`Click on 'View Details' under Metadata to view the Details of MRI Scan [Manual Testing]`

When clicking on 'View Details' of some of the rows of the datatable, it leads to a `500` error. This possibly occurs when there is no imaging data for that specific row. Regardles, it should not be returning a `500` error with a blank page. The user should be warned when there are no details to be viewed, or the 'View Details' button should not be accessible.

ridz1208:
@nicolasbrossard please add the Pr sent label when you fix an issue

thx

zaliqarosli:
PR merged

PR #4643, opened 2019-05-16
DICOM archive details does not work when more than one upload is associated to the archive.

nicolasbrossard:
### Brief summary of changes

When more than one MRI upload is associated to a given archive, its details cannot be displayed in the DICOM archive: an HTTP 500 error is generated (a `pselectRow` statement returns more than one row).

This PR also fixes the breadcrumbs on the DICOM archive details page: clicking on the 'details' breadcrumb if your are already on the DICOM archive details page takes you to a DICOM archive details page where all the lines in the displayed table are empty.

### This resolves issue...

- [ ] Github? #4553

### To test this change...

Access the DICOM archive and try to view the details of an archive associated to multiple MRI uploads. On the test VM, one such archive has TarchiveID=2.

### DIFF
--- a/modules/dicom_archive/php/viewdetails.class.inc
+++ b/modules/dicom_archive/php/viewdetails.class.inc
@@ -25,7 +25,10 @@ namespace LORIS\dicom_archive;
  */
 class ViewDetails extends \NDB_Form
 {
-   var $tarchiveID;
+    /**
+     * ID of the tarchive for which the details should be displayed.
+     */
+    var $tarchiveID;
 
     /**
      * Keeps array of protocols from mri_protocol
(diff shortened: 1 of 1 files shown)


modules/dicom_archive/php/viewdetails.class.inc
```diff
@@ -25,6 +25,11 @@ namespace LORIS\dicom_archive;
  */
 class ViewDetails extends \NDB_Form
 {
+    /**
+     * ID of the tarchive for which the details should be displayed.
+     */
+    var $tarchiveID;
```
driusan:
```suggestion
    public $tarchiveID;
```

driusan: @nicolasbrossard please fix this so it can be merged

Issue #4523, opened 2019-05-13
[user_accounts] bug with viewing and certifying examiners within own user's site


jesscall:
Permission `examiner_view` should give a user the ability to view `Examiner Status` and certify examiners within their site (on user accounts edit page). This currently is not the case. 

However when the user has `examiner_multisite` permission, they are able to view / edit all users' examiner status.

PR #4644, opened 2019-05-16
[user_accounts] fix permissions for viewing / certifying examiners from user accounts edit page

jesscall:
### Brief summary of changes
Examiner status fields in user_accounts edit form were only show for users with `examiner_multisite` permission. 
This PR makes it so users with `examiner_view` permissions can view / certify users from their own site.

Recap on the permissions:
`examiner_multisite`: add / certify examiners across all sites
`examiner_view` : add / certify examiner at users' own site

### This resolves issue...
- [ ] Github? #4523 

### To test this change...

- [ ] Play around with different permissions and site affiliations. 
- [ ] When you have `examiner_multisite` you should see all fields related to examiner and all center aliases displayed. 
- [ ] When you have `examiner_view` you should ONLY see the examiner fields and center aliases that you are affiliated with IF: 1) you share a site with the user you are editing, 2) you are creating a new user.

### DIFF
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -278,7 +278,7 @@ class Edit_User extends \NDB_Form
         // END multi-site UPDATE
 
         // EXAMINER UPDATE
-        if ($editor->hasPermission('examiner_multisite')) {
+        if ($this->showExaminerFields($editor)) {
             // get all fields that are related to examiners
             $ex_curr_sites  = array();
             $ex_prev_sites  = array();
@@ -716,10 +716,16 @@ class Edit_User extends \NDB_Form
             array('multiple' => 'multiple')
         );
 
-        if ($editor->hasPermission('examiner_multisite')) {
+        // Should be able to view examiner fields in the following cases:
+        // (1) Editor has 'examiner_multisite' permission.
+        // (2) Editor has 'examiner_view' permission
+        //      AND  either:
+        //          i) The user is from same site as editor,
+        //         OR
+        //          ii) The editor is creating a new user.
+        if ($this->showExaminerFields($editor)) {
             //get site aliases
-            $DB      =& \Database::singleton();
-            $aliases = $DB->pselect("SELECT CenterID, Alias FROM psc ", array());
+            $aliases = $this->getEditorSiteAliases($editor);
 
             foreach ($aliases as $row) {
                 $groupA[] = $this->createCheckbox(
@@ -1087,7 +1093,7 @@ class Edit_User extends \NDB_Form
         //======================================
         //        Validate Examiner Status
         //======================================
-        if ($editor->hasPermission('examiner_multisite')) {
+        if ($this->showExaminerFields($editor)) {
             $matched = false;
             foreach ($values as $k => $v) {
                 if (preg_match("/^ex_[0-9]+$/", $k)) {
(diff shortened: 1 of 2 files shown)


jesscall:
TODO: change Test Plan (will do tomorrow)

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -1218,6 +1224,70 @@ class Edit_User extends \NDB_Form
...
+    function showExaminerFields($editor)
+    {
+        $siteMatch = false;
+        if (!$this->isCreatingNewUser()) {
+            $user        = \User::factory($this->identifier);
+            $userSites   = $user->getCenterIDs();
+            $editorSites = $editor->getCenterIDs();
+            foreach ($userSites as $site) {
+                if (in_array($site, $editorSites)) {
+                    $siteMatch = true;
```
maltheism: Hi @jesscall, shouldn't you add a `break;` after $siteMatch = true; and so you break out of the foreach for the performance benefit of not having to continue looping when you found the match?

jesscall: good point! I've made the suggested change

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -1218,6 +1224,71 @@ class Edit_User extends \NDB_Form
...
+     *              AND  either:
+     *                  i) The user is from same site as editor,
+     *                      OR
+     *                  ii) The editor is creating a new user.
+     *
+     * @param \User $editor User editing the form
+     *
+     * @return bool  - whether or not to display examiner fields
+     */
+    function showExaminerFields($editor)
```
driusan: I find this method name confusing because it sounds like an action, not a question. Maybe something like `shouldShowExaminerFields` would be better (though I'm not entirely happy with that either..)

Also, since this is the major branch can you make sure you add the return type for PHP instead of just the comment?

jesscall: I renamed the function to something a little clearer.

For the return type, none of the other functions in the class have return type indicated. Was i supposed base off of master rather than 21.0-release?

modules/user_accounts/test/TestPlan.md
```diff
@@ -49,7 +49,8 @@ When creating or editing a user: (subtest: edit_user)
 16. Check that selecting sites for the "Examiner At:" Section and saving, adds user X to the Examiner list (and in examiners table).
 17. Check that de-selecting sites from the "Examiner At:" section and saving, does NOT delete X from the Examiner table but rather sets them as inactive for that site.
 18. Check that setting the radiologist to Yes/No changes the values for X for all sites
-19. Check that by removing the "examiner_multisite" permission, the user(editor) can not modify X's Examiner Status.
+19. Check that by removing the "examiner_view" and "examiner_multisite" permission, the user(editor) cannot modify X's Examiner Status.
+20. Check that by removing the "examiner_multisite" permission, the user(editor) can only modify X's Examiner Status if the editor and user X are affiliated with a common site.
 20. Setting the Pending Approval for user X prevents user X from logging in until his/her account is approved.
```
ridz1208: renumbering

ridz1208:
@driusan please rereview

modules/user_accounts/test/TestPlan.md
jesscall: okai done

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -1218,6 +1224,71 @@ class Edit_User extends \NDB_Form
...
+    /**
+     * This function returns a boolean representing whether or not
+     * to display examiner fields to a user that is editing an account.
+     * Examiner fields SHOULD be displayed in the following cases:
+     *      (1) Editor has 'examiner_multisite' permission
+     *          (which allows to add / certify examiners across all sites).
+     *      (2) Editor has 'examiner_view' permission
+     *          (which allows them to add / certify examiner at their own site)
+     *              AND  either:
+     *                  i) The user is from same site as editor,
```
ridz1208: please add explicitly that:

The editor does not need to have ALL the sites of the editee (generally thats how it works in other places of loris) because the editor will not see on the front end any sites that he/she does not belong to

driusan:
@jesscall I added the needs work tag because of Rida's comments, but also this has conflicts now.

jesscall:
Spoke to Rida and Dave about this. The changes are not actually critical to release as they are adding new logic / feature to the module which was not present before. 
Future changes to this PR will include making the examiner sites a multiselect rather than checkboxes + reflecting the logic of conditional display of sites

ridz1208:
I dont forsee this happening any time soon and it could be reassigned to a more junior developer, if Jess agrees I would say lets close the PR and issue a new one when it's ready

Issue #4580, opened 2019-05-14
[bvl_feedback] feedback will only save after toggling 'Feedback Type'


leighevans:
Currently when trying to submit new feedback (at any level. i.e. profile, visit, instrument), the field "Feedback Type" must be toggled in order to save successfully. The Feedback Type field is set to "Input Errors" by default, but trying to press Save Data without toggling it first will not register anything (even if the type of feedback you are saving is an input error, you must toggle to a different option and re-select it in order to save).

driusan:
Fixed by #4580

PR #4645, opened 2019-05-17
[bvl_feedback] feedback will only save after toggling 'Feedback Type'

kongtiaowang:
to resolve https://github.com/aces/Loris/issues/4580


jesscall:
lgtm

Issue #4586, opened 2019-05-14
[bvl_feedback] Amend test plan


leighevans:
There is an error in the test plan for bvl feedback. "**Threads should be the thread number**" accidentally contains markdown header formatting and should be fixed to be regular text. 
![Screen Shot 2019-05-14 at 3 50 57 PM](https://user-images.githubusercontent.com/32552953/57727748-43539580-7660-11e9-8313-d50228b50451.png)

driusan:
Fixed by above PR

PR #4646, opened 2019-05-17
[bvl_feedback] Amend test plan - format

kongtiaowang:
To resolve the test plan format.
https://github.com/aces/Loris/issues/4586


modules/bvl_feedback/test/TestPlan.md
```diff
@@ -17,7 +17,7 @@
     * QC class should be what page the feedback was submitted on (i.e. profile, instrument)
     * Instrument should be populated if QC class is instrument
     * Visit should be populated if QC class is not profile
-    * # Threads should be the thread number
+    * Threads should be the thread number
```
driusan: Can you try `* "# Threads" should be the thread number`? I think it's labeled "# Threads" in the frontend, not "Threads", but I'm not sure how to otherwise resolve this formatting problem..

Issue #4647, opened 2019-05-17
[Install] Formatting error in frontend of the LORIS installation


AlexandraLivadas:
When installing LORIS for the first time, you are asked to open "<loris-url>/installdb.php" in your browser.
At the bottom of the page, there are three checkboxes that are badly formatted, making them unclear.
Here is a screenshot of the issue:
<img width="1127" alt="Screen Shot 2019-05-17 at 12 12 26 PM" src="https://user-images.githubusercontent.com/35467361/57941479-10e2ac00-789d-11e9-8a96-8e43997cfd14.png">

PR #4649, opened 2019-05-17
[install] Fixed checkbox formatting in installdb.php frontend

AlexandraLivadas:
### Brief summary of changes
I made changes to the [Loris/smarty/templates/install.tpl](https://github.com/aces/Loris/blob/minor/smarty/templates/install.tpl) file to fix the formatting of the checkboxes in the loris-url/installdb.php frontend.

Here is the original formatting:
Screen Shot 2019-05-17 at 12 12 26 PM

And the fixed formatting:
Screen Shot 2019-05-17 at 12 32 57 PM

### This resolves issue...
- [ ] Github #4647 

### To test this change...

Since Loris is already installed, you will need to move the config.xml file out of the project folder so that when you check out <loris-url>/installdb.php, it does not say "Loris already installed".

1. Move config.xml from loris/project to another folder (ex. just the loris folder)
2. Put loris-url/installdb.php in browser (ex. for me, it is alivadas-dev.loris.ca/installdb.php)
3. Check that the changes were made.

### DIFF
--- a/smarty/templates/install.tpl
+++ b/smarty/templates/install.tpl
@@ -174,33 +174,27 @@
             </fieldset>
             <fieldset>
                 <legend>Options</legend>
-            <div>
-                <div class="col-md-9">
-                    <label for="use_existing_database">Use existing database:</label>
-                </div>
-                <div class="col-md-9">
-                    <input type="checkbox" id="use_existing_database" name="use_existing_database" value="on" {($use_existing_database == 'on') ? 'checked' : ''}/>
-                                        Check this if the database already exists but does not have Loris installed yet
-                </div>
+            <div class="col-md-12">
+                <label for="use_existing_database">Use existing database:</label>
             </div>
-            <div>
-                <div class="col-md-10">
-                    <label for="use_existing_tables">Use existing tables:</label>
-                </div>
-                <div class="col-md-10">
-                    <input type="checkbox" id="use_existing_tables" name="use_existing_tables" value="on" {($use_existing_tables == 'on') ? 'checked' : ''}/>
-                                        Check this if you already installed Loris at the specified database
-                </div>
+            <div class="col-md-12">
+                <input type="checkbox" id="use_existing_database" name="use_existing_database" value="on" {($use_existing_database == 'on') ? 'checked' : ''}/>
+                                    Check this if the database already exists but does not have Loris installed yet
             </div>
-            <div>
-                <div class="col-md-10">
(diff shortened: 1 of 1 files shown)


smarty/templates/install.tpl
```diff
@@ -1,4 +1,4 @@
-<html>
```
kongtiaowang: missing "<"

smarty/templates/install.tpl
```diff
@@ -1,4 +1,4 @@
...
+html>
```
kongtiaowang: missing "<"

AlexandraLivadas:
@christinerogers I just checked it with different browser widths/heights and it seems to be responding well and is remaining well formatted.

kchatpar:
@AlexandraLivadas could you make the files in the description hyperlinks.

smarty/templates/install.tpl
```diff
@@ -175,7 +175,7 @@
             <fieldset>
                 <legend>Options</legend>
             <div>
-                <div class="col-md-2">
+                <div class="col-md-9">
```
driusan: Isn't bootstrap based on a 12 column grid?

This changes it to 9+9 = 18 columns. (Granted, 9+2=11 also wasn't right, but it's closer..)

AlexandraLivadas:
@driusan You are correct, but no matter what combination I format the columns in (2-10, 3-9, etc), it still displays strangely. 
For example, here is the format when all the columns have "col-md-2" for the label and "col-md-10" for the checkbox:
Screen Shot 2019-05-22 at 2 00 14 PM

I noticed something strange too where if I made an extra column with "col-md-2" at the end of each line and shortened the checkbox column to "col-md-8", the formatting looked fine (so the columns would be 2-8-2 rather than 2-10). However, this is obviously not good code and just a weird fix. Here is a picture, with the two extra columns tagged Test1 and Test2:
Screen Shot 2019-05-22 at 2 06 57 PM

I think the best solution for now would be to make each piece of text its own row with only one column of length 12. This is what it would look like:
Screen Shot 2019-05-22 at 2 09 19 PM

I have implemented the last solution (making everything its own line) and will push my changes here shortly!

driusan:
@AlexandraLivadas given the 2-8-2 method seems to work, you could also add a 2 column cell with content such as "&nbsp;" instead of printed text. Another thing you could try is ensuring the `<div>` wrappers have a "row" class (I don't know enough about bootstrap to know how likely that is to be the source of the problem, but it doesn't hurt to try.)

Issue #4540, opened 2019-05-13
[Candidate List] Advanced filters are not expanded on refresh when an advanced filter is set


coreyAbout:
[TestPlan](https://github.com/aces/Loris/blob/minor/modules/candidate_list/test/TestPlan.md) 6. Verify advanced filters are expanded on page load when an advanced filter is set, and collapsed otherwise.

Problem: When one advanced filter is selected, the advanced filter section is still collapsed when refresh the page or returning to the page from another page.

kongtiaowang:
@coreyAbout We need to update this test plan after this module reactified.

kongtiaowang:
https://github.com/aces/Loris/pull/4636

HenriRabalais:
PR Sent to resolve this issue by introducing a query param that tracks the advanced filter.
#4650

zaliqarosli:
@coreyAbout @kongtiaowang @HenriRabalais i can't figure it out .. is this issue fixed now?

kongtiaowang:
@zaliqarosli It has been fixed.

PR #4650, opened 2019-05-17
[Candidate List] Enabled Advanced Filter Toggle Via Query Params

HenriRabalais:
## Summary
Toggling the 'Show Advanced Filter' Button will append a query param to the URL that will propagate the state of the toggled filter when the loading the page with that URL.

## Issue
#4540

## Testing
1. Go to Candidate List module.
2. Press the 'Show Advanced Filter' Button.
3. Make sure a query param in the url is added called `hide` with value `false`
4. Refresh the page.
5. Make sure the advanced filters are still revealed.
6. Repeat steps 2-5 except by pressing the `Hide Advanced Filter` Button.

### DIFF
--- a/jsx/Filter.js
+++ b/jsx/Filter.js
@@ -17,6 +17,13 @@ class Filter extends Component {
     this.renderFilterFields = this.renderFilterFields.bind(this);
   }
 
+  componentDidMount() {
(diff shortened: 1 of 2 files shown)


maltheism:
Passed manual testing

Issue #4608, opened 2019-05-15
[instrument_builder] Clicking "Edit" on a Header crashes the system


atspahr:
This only works if you save an instrument and then load it back in to re-edit it.

![image](https://user-images.githubusercontent.com/48061979/57785516-e6a7b780-76ff-11e9-9e9d-66e4fab19446.png)

atspahr:
This also happens when trying to edit "Scored Field"

![image](https://user-images.githubusercontent.com/48061979/57785709-40a87d00-7700-11e9-814a-962fac342922.png)

atspahr:
This also happens when trying to edit "Label"

![image](https://user-images.githubusercontent.com/48061979/57785805-69307700-7700-11e9-9c6e-f2dfd28c2a32.png)

atspahr:
Same thing happens on Microsoft Edge browser

![image](https://user-images.githubusercontent.com/48061979/57786660-e6102080-7701-11e9-91e7-ab137b0d27ff.png)

driusan:
Resolved by #4652

PR #4652, opened 2019-05-17
[instrument_builder] Clicking "Edit" on a Header crashes the system

kongtiaowang:
Resolve : https://github.com/aces/Loris/issues/4608

### DIFF
--- a/modules/instrument_builder/jsx/react.questions.js
+++ b/modules/instrument_builder/jsx/react.questions.js
@@ -2,7 +2,7 @@
  NumericOptions, ListElements, AddElement */
 
 /* global Instrument */
-
+z
 /**
  * This file contains the React classes for instrument builder
  * module. It is used to add and edit questions in the instrument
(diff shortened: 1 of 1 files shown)


modules/instrument_builder/jsx/react.questions.js
```diff
@@ -608,7 +608,7 @@ class AddElement extends Component {
     if (this.props !== undefined && this.props.element) {
       // Editing an element, set to elements state
       this.state = {
-        Options: Instrument.clone(this.props.element.Options),
+        Options: Instrument.clone(this.props.element.Options === undefined ? {} : this.props.element.Options ),
         Description: Instrument.clone(this.props.element.Description),
         Name: Instrument.clone(this.props.element.Name),
```
maltheism: Hi @kongtiaowang line number 613 needs to be changed to:
`Name: Instrument.clone(this.props.element.Name === undefined ? '' : this.props.element.Name),`
and it will fix the undefined error.

driusan: I'm far from a javascript expert.. but wouldn't that be the same as (and more idiomatic as) `this.props.Element.Name || ''`?

maltheism: @driusan that works as well and can be used instead. 
I'm so fond of ? .. : ..; that I never think to write anything else.

modules/instrument_builder/jsx/react.questions.js
```diff
@@ -2,7 +2,7 @@
  NumericOptions, ListElements, AddElement */
 
 /* global Instrument */
-
+z
```
maltheism: The 'z' needs to be removed

Issue #4138, opened 2018-11-21
Fix JS lint warnings


johnsaigle:
There are many JavaScript files within the code base that generate warnings in our automated testing procedure. Most of them have to do with lines being too long or functions missing documentation.

Fixing these warnings would be a good way for someone new to get familiar with the code base as well as to learn our code review procedure. 

### TODO
* On a development instance of LORIS run `npm run lint:javascript`
* Issue a pull request that eliminates warnings

driusan:
Note: you don't need to fix *all* of the warnings, fixing some of them would still be a step forward.

johnsaigle:
@driusan This one has been untapped for almost a year, should we leave it up?

If so, we can add the `Bug` label so we don't have to fight stale bot.

driusan:
Sure. I think we should definitely keep the issue, because all those warnings are really annoying and should be fixed.

laemtl:
Replaced by #6750

PR #4653, opened 2019-05-17
Fixed the 'no-warning-comments' ESLint warnings

AlexandraLivadas:
### Brief summary of changes
I made changes to some of the .js files so that when running "npm run lint:javascript" there would be less warning from eslint. 

### This resolves issue...

- [ ] Github? #4138 

### To test this change...

- [ ] Run "npm run lint:javascript"

### DIFF
--- a/modules/candidate_parameters/jsx/ConsentStatus.js
+++ b/modules/candidate_parameters/jsx/ConsentStatus.js
@@ -190,7 +190,7 @@ class ConsentStatus extends React.Component {
 
   render() {
     // If error occurs, return a message.
-    // XXX: Replace this with a UI component for 500 errors.
+    // Replace this with a UI component for 500 errors.
     if (this.state.error) {
       return <h3>An error occured while loading the page.</h3>;
     }
--- a/modules/datadict/jsx/dataDictIndex.js
+++ b/modules/datadict/jsx/dataDictIndex.js
@@ -74,7 +74,7 @@ class DataDictIndex extends React.Component {
     this.setState({filter});
   }
 
-  // TODO: deprecate clearing filters via refs in future refactoring.
+  // Deprecate clearing filters via refs in future refactoring.
   /**
    * Reset the filter elements with textInput refs to empty values
    */
--- a/modules/help_editor/jsx/help_editor.js
+++ b/modules/help_editor/jsx/help_editor.js
@@ -25,7 +25,7 @@ class HelpEditor extends React.Component {
       data: {}
     };
 
-    // TODO: refs should be deprecated in future refactoring.
+    // Refs should be deprecated in future refactoring.
     /**
      * Set filter to the element's ref for filtering
      */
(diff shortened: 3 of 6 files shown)


AlexandraLivadas:
@johnsaigle So far, I have only resolved the "no-warning-comment" warnings. I was unsure about our coding standards when it comes to line length and some of the issues were asking for jsdoc comments and I did not know enough about the code to write good comments.

@zaliqarosli @alisterdev @llevitisI @HenriRabalais @davidblader @cmadjar @xlecours @leothomas @jacobpenny @driusan : I made comment changes to some files that you authored. Please let me know if this is okay or if I should rollback some of the changes!

davidblader:
lgtm

zaliqarosli:
the 'TODO', 'FIXME' are important because they tell us that these are things we need to do / lines of code we need to fix. 'XXX' is also a way to highlight todo/fix. removing them will remove the context of these comments. these comments are here mostly for developers and less so actual documentation for the functions. i would maybe turn off 'no-warning-comments' than removing the comments? @johnsaigle i am curious what you think

maltheism:
I think it would be best to leave the comments and even the warnings in regard to TODO, FIXME, XXX because it helps developers in remembering where to go for cleanup and fixes. That's kind of useful warnings compared to the countless warnings for function comments that haven't been generated when running npm run compile. 

One of these days I would like to go through the javascript files and make a comment with the necessary params to get rid of all those warnings for function comments. Then we will get one or two warnings that are actual reminders that we need to write a comment for a function and quickly do it instead of just ignoring the hundreds of warnings in front of us.

Anyway that's my opinion but I'm in favour of whatever people agree with.

johnsaigle:
Yeah I agree with the others. I like the TODO/FIXMEs. They even have special syntax highlighting in certain dev environments so that they'll jump out to the author.

In terms of line length our ideal is 80 characters or less; this is strictly enforced in PHP via `phpcs` though only marked as a warning in `eslint`. 

Also, a few of the people you tagged don't work here anymore just as an FYI :) (Alex/alisterdev, davidblader, jacobpenny, and leothomas).

johnsaigle:
And yes I agree with @zaliqarosli that `no-warning-comments` should be disabled instead.

AlexandraLivadas:
I looked at the package.json file and it looks like we are using the Google configuration. I took a look at their latest configuration online and it seems like they do not allow these "warning comments". I think that in order for us to get rid of those warnings, we would need to write our own ESLint configuration.

Otherwise, for now, I can undo my changes on the comments and get started on the line length warnings!

driusan:
I agree with the others that it's better to leave the TODO/FIXME/XXX part of the comments for context.

You're right that the javascript eslint config was based on the Google coding standards when it was setup, but I don't think there's any reason we can't set up our own config and diverge from them.

johnsaigle:
Hey @AlexandraLivadas I'm going to close this PR just to keep things tidy. I definitely think it would be helpful to make the changes to the eslint file or adjust line lengths though! If you have time to do that, that would be great. No worries if you're focusing on test stuff instead.

Issue #4512, opened 2019-05-13
[Login] request-account post is sent when form not complete


maltheism:
POST is sent when the request-account form hasn't been filled out.
https://github.com/aces/Loris/projects/19#card-21296846

kchatpar:
@maltheism We can't reproduce the error. Could you explain how where you're seeing the post request? @zaliqarosli

maltheism:
Hi @kchatpar when visiting http://localhost/login/request-account/ and clicking the "Request Account" button. My apache access.log contains this:
`::1 - - [17/May/2019:10:09:47 -0400] "POST /login/request-account/ HTTP/1.1" 200 5938`

zaliqarosli:
@maltheism i'm not sure but could this be because the `_validate()` function in `RequestAccount` class needs to be called? it looks to me that the form doesn't actually submit because the validate function returns errors

zaliqarosli:
we could make the fields required inputs but then the browser takes over the front-end validation and will provide its own errors instead of our own custom ones. what should we do?

maltheism:
Hi @zaliqarosli I thought it was not good design to allow a post from the frontend when nothing is included in any of the fields. I think adding "required" to the input tag does make the form require something from the user and chrome likely handles the notification that the field needs a value. Sorry about the late response.

zaliqarosli:
PR fix https://github.com/aces/Loris/pull/4659

driusan:
Fixed by the above

PR #4659, opened 2019-05-21
[login/RequestAccount] Add required to form fields

zaliqarosli:
fixes https://github.com/aces/Loris/issues/4512

### DIFF
--- a/modules/login/php/requestaccount.class.inc
+++ b/modules/login/php/requestaccount.class.inc
@@ -38,17 +38,26 @@ class RequestAccount extends \NDB_Form
         $this->addBasicText(
             "firstname",
             "First name",
-            array("placeholder" => "First Name")
+            array(
+              "placeholder" => "First Name",
+              "required"    => true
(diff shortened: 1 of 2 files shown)


maltheism:
Hey @zaliqarosli awesome! 
I just tested this PR and there is one error preventing these changes from working. Basically in the php/libraries/LorisForm.class.inc on line number 790 needs to be changed to $el['required'] and similar for line number 793.

maltheism:
I just searched all of LORIS to see if "requiredMsg" should be replaced anywhere else and just inside the LorisForm.class.inc contains this array key. All the locations besides the two line numbers I already listed above are 850, 1217, 1784, 1849. If these are updated to 'required' nothing should break.

zaliqarosli:
@maltheism the issue is only a php notice for an undefined index so ive added an isset which removes the notice. i dont think any other changes is required

kchatpar:
Looks good to me.

PR #4667, opened 2019-05-22
[instrument_list] can not load for next stage

kongtiaowang:
Changing the return type.
To test: load instrument_list on test.loris.ca (has been fixed on testing VM).

### DIFF
--- a/php/libraries/TimePoint.class.inc
+++ b/php/libraries/TimePoint.class.inc
@@ -642,7 +642,7 @@ class TimePoint
      *
      * @return string study stage title
      */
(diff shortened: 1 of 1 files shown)


driusan:
I don't understand what it would mean for this to return null. The "next stage" has an enumeration of possible values, why can't we provide a value for whichever case this is trying to solve?

johnsaigle:
@driusan Both values are defined in the database to have default null so it's possible that this case can occur.

driusan:
Fair enough. I guess I can merge it once it passes Travis.

Issue #4545, opened 2019-05-13
[statistics]  Imaging Statistics not working


coreyAbout:
[TestPlan](https://github.com/aces/Loris/blob/minor/modules/statistics/test/TestPlan.md) 
`2. Try clicking on all the tabs. Do they appear to be working?`

Problem:
Click one `Imaging Statistics` tab, nothing happens.

johnsaigle:
Could you include the output of your error log?

My suspicion is that the server you're testing on is missing some configuration settings

ridz1208:
```[Mon May 13 15:08:09.758368 2019] [php7:error] [pid 17239] [client 70.81.209.39:55182] PHP Fatal error:  Uncaught DatabaseException: Unknown column 'DTI_Scan_Done' in 'field list' in /var/www/loris/php/libraries/Database.class.inc:707\nStack trace:\n#0 /var/www/loris/php/libraries/Database.class.inc(739): Database->execute(Object(PDOStatement), Array)\n#1 /var/www/loris/modules/statistics/php/stats_mri.class.inc(384): Database->pselect('SELECT s.Subpro...', Array)\n#2 /var/www/loris/php/libraries/NDB_Form.class.inc(188): LORIS\\statistics\\Stats_MRI->setup()\n#3 /var/www/loris/src/Middleware/UserPageDecorationMiddleware.php(171): NDB_Form->handle(Object(Zend\\Diactoros\\ServerRequest))\n#4 /var/www/loris/src/Middleware/PageDecorationMiddleware.php(49): LORIS\\Middleware\\UserPageDecorationMiddleware->process(Object(Zend\\Diactoros\\ServerRequest), Object(LORIS\\statistics\\Stats_MRI))\n#5 /var/www/loris/php/libraries/NDB_Page.class.inc(653): LORIS\\Middleware\\PageDecorationMiddleware->process(Object(Zend\\Diactoros\\ServerRequest), Object(LORIS\\statistics\\Stats_MRI))\n#6 /var/www/loris/php/libraries/Module.class.inc(341): NDB_Page-> in /var/www/loris/php/libraries/Database.class.inc on line 707, referer: https://test.loris.ca/statistics/
```

ridz1208:
the mri_parameter_form does not contain all the necessary columns for each scan_type

jftunteng:
@ridz1208  closing this since the PR is merged.

PR #4672, opened 2019-05-22
[statistics] fix imaging tab loading when Scan_Done is not in mri_parameter_form

ridz1208:
### Brief summary of changes
The imaging tab of statistics depends on the scan types defined in the `mri_scan_type` table. each scan type is crossmatched with a column in the `mri_parameter_form` table. if that column does not exist the SQL query is now bypassed and an error message is displayed stating that the column does not exist

> to review use the hide white spaces changes option

### This resolves issue...

- [ ] Github? #4545

### To test this change...

- [ ] make sure the imaging tab of statistics loads even if the mri_parameter_form does not have all the necessary columns

### DIFF
--- a/modules/statistics/php/stats_mri.class.inc
+++ b/modules/statistics/php/stats_mri.class.inc
@@ -70,6 +70,7 @@ class Stats_MRI extends \NDB_Form
      * @param string  $Subsection        the value of Subsection=''
      * @param string  $disclamer         the value of disclamer=''
      * @param ?string $projectID         the value of projectID is null
+     * @param bool    $scanDoneExists    does the column exist in the databse
      *
      * @return string
      */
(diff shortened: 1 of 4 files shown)


modules/statistics/php/stats_mri.class.inc
```diff
@@ -334,18 +341,6 @@ class Stats_MRI extends \NDB_Form
         } else {
             $MRI_Type = $scan_types[$_REQUEST['mri_type']];
         }
-        //TODO IBIS
```
ridz1208: @sruthymathew123 you should know about these for IBIS

sruthymathew123: @ridz1208 Thanks Rida. But for some reason IBIS is no more using imaging statistics.

modules/statistics/php/stats_mri.class.inc
```diff
@@ -70,6 +70,7 @@ class Stats_MRI extends \NDB_Form
      * @param string  $Subsection        the value of Subsection=''
      * @param string  $disclamer         the value of disclamer=''
      * @param ?string $projectID         the value of projectID is null
+     * @param bool    $scanDoneExists    does the column exist in the databse
```
johnsaigle:
```suggestion
     * @param bool    $scanDoneExists    whether the "Scan done" column exists in the database
```

johnsaigle:
Will approve after the above documentation change

ridz1208:
@johnsaigle done

ridz1208:
@johnsaigle approve meeee

PR #4679, opened 2019-05-22
[data_integrity_flag] fix validation on non-required element before db insert

### LINKED ISSUE #4673: [data_integrity_flag] data is not being submitted in update form
When submitting an update, the `data_integrity_flag` table doesn't get updated and changes are not reflected in the front-end browse table.

jesscall:
Resolves #4673 

To test:
Follow last step of test plan

### DIFF
--- a/modules/data_integrity_flag/php/data_integrity_flag.class.inc
+++ b/modules/data_integrity_flag/php/data_integrity_flag.class.inc
@@ -52,7 +52,6 @@ class Data_Integrity_Flag extends \NDB_Menu_Filter
         if ($_SERVER['REQUEST_METHOD'] === "POST") {
             $stdin = file_get_contents('php://input');
             $req   = json_decode($stdin, true);
-            error_log(print_r($req, true));
             $this->_insertDataFlag($req);
         }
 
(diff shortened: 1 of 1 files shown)


modules/data_integrity_flag/php/data_integrity_flag.class.inc
```diff
@@ -73,13 +73,15 @@ class Data_Integrity_Flag extends \NDB_Menu_Filter
...
             || !isset($req['visitLabel'])
-            || !isset($req['comment'])
             || !isset($req['date'])
             || !isset($req['flagStatus'])
         ) {
             return "Error";
         }
         $req = \Utility::asArray($req);
+        // comment may not be set since it is not a required field
+        $comment = (isset($req['comment']) && !empty($req['comment'])) ?
```
johnsaigle:
```suggestion
        $comment = !empty($req['comment']) ?
```

@driusan just tuaght me that `empty()` will actually check for a variable being set all on its own, so I don't think you need it here.

driusan: @jesscall ping

Issue #4584, opened 2019-05-14
[instrument_list] Cannot select BVL QC Type or BVL QC Status


leighevans:
Instrument data entry and DDE is complete with no conflicts and Send to DCC is not available (even with Send to DCC permission and visit set to Pass), and clicking on any box for BVL QC (e.g. Visual, Hardcopy) does not register nor does clicking Complete for BVL QC Status as a result. 

![Screen Shot 2019-05-14 at 3 40 26 PM](https://user-images.githubusercontent.com/32552953/57727195-d5f33500-765e-11e9-9b6a-58407eaa39ba.png)

zaliqarosli:
@ridz1208 i can't reproduce this bug because all the instruments (except bmi) break so we can't bring the timepoint to the approval stage

zaliqarosli:
https://github.com/aces/Loris/issues/4641

zaliqarosli:
i fixed the BVLQCType issue (you can now click on it and it submits) @leighevans if you would like to test https://github.com/aces/Loris/pull/4683

zaliqarosli:
PR for BVLQCType issue merged. PR for Send to DCC issue still needs to be made by @kchatpar

ridz1208:
I guess its 50% done, repoened.

@kchatpar please issue PR

sruthymathew123:
@zaliqa I am trying to catch up the things to work on this issue. But, I cannot reproduce the issue (Send to DCC) on my VM or on test.loris.ca anymore. Is that fixed already?
![send_to_DCC](https://user-images.githubusercontent.com/23702452/59118273-b7125680-891d-11e9-8755-faf796fa8be9.png)

zaliqarosli:
@sruthymathew123 i don't think it was fixed yet. @kchatpar can you verify? if 'Send to DCC' is available and working for you, i'm happy to close this issue. let me know!

sruthymathew123:
@kchatpar Can you verify if this issue(Send to DCC not available) is still existing?

sruthymathew123:
@driusan @kchatpar @zaliqarosli  I don't know for some reason 'Send to DCC' is not clickable on **test.loris.ca** even if instrument data entry and DDE is complete with no conflicts and visit set to Pass.

But on my VM with IBIS data;  I couldn't recreate this issue. So I doubt it might be a RB data issue or test VM specific issue.  If anyone can test this on their own VM, that will help to move forward with this issue.

cmadjar:
[...]

PR #4683, opened 2019-05-22
[instrument_list] Fix BVLQCType on ControlPanel

zaliqarosli:
fixes 1 part of https://github.com/aces/Loris/issues/4584

### DIFF
--- a/modules/instrument_list/php/instrument_list_controlpanel.class.inc
+++ b/modules/instrument_list/php/instrument_list_controlpanel.class.inc
@@ -294,7 +294,7 @@ class Instrument_List_ControlPanel extends \TimePoint
                 )
             );
             return;
-        } else if (isset($_REQUEST['setBVLQCStatus'])) && empty($_REQUEST['setBVLQCStatus'])) {
+        } else if (isset($_REQUEST['setBVLQCStatus']) && empty($_REQUEST['setBVLQCStatus'])) {
             $this->setData(
                 array('BVLQCStatus' => null)
             );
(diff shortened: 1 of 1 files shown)


zaliqarosli:
@kchatpar please review and test! :)

modules/instrument_list/php/instrument_list_controlpanel.class.inc
```diff
@@ -294,7 +294,9 @@ class Instrument_List_ControlPanel extends \TimePoint
                 )
             );
             return;
-        } else {
+        } else if (isset($_REQUEST['setBVLQCStatus'])
```
ridz1208: to avoid this kind of confusion in the future could you add one line comment to each of these stating why you added these checks ?

something along the lines of:
`// In order to nullify the value for setBVLQCStatus the front end appends it to the request with an empty string as value`

modules/instrument_list/php/instrument_list_controlpanel.class.inc
```diff
@@ -294,7 +294,12 @@ class Instrument_List_ControlPanel extends \TimePoint
...
+            && empty($_REQUEST['setBVLQCStatus'])
```
driusan: empty doesn't throw an error if the variable isn't set. The isset is redundant.

modules/instrument_list/php/instrument_list_controlpanel.class.inc
zaliqarosli: @ridz1208 please resolve if you're happy

zaliqarosli: i changed it again .. its a bit pedantic but now clear

PR #4684, opened 2019-05-23
Correct column type of ValidMin ValidMax columns of mri_protocol_checks

cmadjar:
### Brief summary of changes

When creating the columns `ValidMin` and `ValidMax` of `mri_protocol_checks`, their type was set to `INT(4)` while it should have been `DECIMAL(10,4)`. This PR corrects the default schema and creates a patch that will need to be run for 21.0.0 release of LORIS.

### This resolves issue...

- bug found while testing 21.0 and migrating the MRI `range` columns to `min` and `max` on a dataset with decimals.

### To test this change...

- [ ]


cmadjar:
@driusan @ridz1208 this PR will need to be part of 21.0. Not sure how to tag it so I put the priority high label. Feel free to modify the label if incorrect. Thanks!

driusan:
they're new columns that aren't used yet, so I'm just going to merge this..

Issue #4583, opened 2019-05-14
[Issue Tracker] Unable to filter by the "Date Created" filter


Jkat:
I'm not sure there's currently any proper implementation of a date filter in any module though.

HenriRabalais:
@zaliqarosli 
@ridz1208 and @driusan are assigning people to testing bugs.

zaliqarosli:
@HenriRabalais i know! i was just letting you know which other issues were open for the issue tracker in case you hadn't seen them and wanted to fix them too!

ridz1208:
@maltheism can you take care of this one ?

zaliqarosli:
this is because column header and row data are mismatched

<img width="561" alt="Screenshot 2019-05-28 19 00 48" src="https://user-images.githubusercontent.com/17415878/58518461-4b8df380-817d-11e9-90ce-cc6dc145ce9e.png">

driusan:
@zaliqarosli's header issue is also reported as #4762

zaliqarosli:
closing this issue so https://github.com/aces/Loris/issues/4762 can deal with header misalignment

PR #4685, opened 2019-05-23
[Issue Tracker] fixed filter by the "Date Created" filter

maltheism:
### Brief summary of changes

Allows not exact matches for 'date' elements and for it to work when there is a timestamp concatenated with the date string.

### This resolves issue...

- [ ] Redmine? #

- [x] Github? #4583

### To test this change...

- [x] Test filter by date created in the issue tracker.


modules/user_accounts/php/edit_user.class.inc
```diff
@@ -454,7 +454,6 @@ class Edit_User extends \NDB_Form
             $user    = \User::factory($this->identifier);
             $success = $user->update($set);
         }
```
maltheism: Made this change because travis was failing because of the extra line.

modules/user_accounts/php/my_preferences.class.inc
```diff
@@ -191,8 +191,8 @@ class My_Preferences extends \NDB_Form
         // update the user
         $success = $user->update($set);
 
-        // Now set the password. Note that this field is named incorrectly at this point in the code
-        // and represents a plaintext password, not a hash.
+        // Now set the password. Note that this field is named incorrectly
```
maltheism: Made this change because travis was failing from line length exceeding 85 characters.

modules/user_accounts/php/edit_user.class.inc
driusan: #4688

zaliqarosli:
i believe this issue is actually because the columns and headers are not matched! 'Date created' doesn't technically exist because the data is actually under 'Visit label' column. if you put the date in there, it will filter correctly lol

Screenshot 2019-05-28 18 57 23

so unfortunately, this PR is not the solution

zaliqarosli:
more info for mismatch

Screenshot 2019-05-28 19 00 48

jsx/Filter.js
```diff
@@ -39,7 +39,8 @@ class Filter extends Component {
   onFieldUpdate(name, value, id, type) {
     const searchParams = new URLSearchParams(location.search);
     const filter = JSON.parse(JSON.stringify(this.props.filter));
-    const exactMatch = type === 'textbox' ? false : true;
+    let exactMatch = type === 'textbox' ? false : true;
```
zaliqarosli: i do think this functionality is needed though. but it should be done like this:

```suggestion
    let exactMatch = (type === 'textbox' || type === 'date') ? false : true;
```

zaliqarosli: and line 980 of `Form.js` changed to `this.props.onUserInput(this.props.name, e.target.value, e.target.id, 'date');`

maltheism: Thanks @zaliqarosli, I didn't think to make it one line like that. I made it a little shorter by doing: `const exactMatch = (!(type === 'textbox' || type === 'date'));`

zaliqarosli:
@maltheism nice! i can approve and pass this PR as long as the table header misalignment issue is being fixed elsewhere?

maltheism:
Hey @zaliqarosli, I think that's fixed elsewhere.

zaliqarosli:
@maltheism which PR and who fixed it?

Issue #4567, opened 2019-05-14
[new profile] no error thrown when user generated PSCID does not respect expected format


ridz1208:
If the configuration for PSCID generation in the config.xml file is set to `user`, and you try to input an incorrect PSCId, no error is returned to the user, a 500 error is thrown in the backend only

ridz1208:
Several issues are happening here 
 1. The validation function from before the reactification was removed ! so there is currently only front end validation and that should not be the case (see #4088)
 2. There is no PSCID validation for when the `user` generation method is used
 3. errors returned from the backend when `user` PSCID generation is used appear as 500 errors in the console and are not returned in an alert message on the front end

Issues 1 and 3 above are necessary, Issue 2 is ideal but not critical for the release

ridz1208:
Also, the test plan should probably be amended to include an item to test the backend validation independently from the front end

driusan:
Added the Critical tag because of the lack of backend validation

zaliqarosli:
new PR https://github.com/aces/Loris/pull/4834

cmadjar:
Fixed by the above PR so closing the issue

PR #4693, opened 2019-05-24
[new profile] error thrown when user generated PSCID does not respect expected format

maltheism:
### Brief summary of changes

I added the previous backend checks and they should fix the issue.

### This resolves issue...

- [ ] Redmine? #

- [x] Github? #4567

### To test this change...

- [x] Test the new_profile module.

### DIFF
--- a/modules/new_profile/jsx/NewProfileIndex.js
+++ b/modules/new_profile/jsx/NewProfileIndex.js
@@ -19,6 +19,7 @@ class NewProfileIndex extends React.Component {
       isLoaded: false,
       isCreated: false,
       errMessage: '',
+      formError: null,
     };
     this.handleSubmit = this.handleSubmit.bind(this);
     this.setFormData = this.setFormData.bind(this);
@@ -69,7 +70,7 @@ class NewProfileIndex extends React.Component {
   /**
    * Handles form submission
    *
-   * @param {event} e - Form submition event
+   * @param {event} e - Form submission event
    */
   handleSubmit(e) {
     e.preventDefault();
@@ -95,8 +96,15 @@ class NewProfileIndex extends React.Component {
         })
       .then((resp) => resp.json())
       .then((data) => {
-        this.setState({newData: data});
-        this.setState({isCreated: true});
+        console.log(data);
+        if (data.error) {
+          console.log('error:' + data.error);
+          this.setState({formError: data.error});
+        } else {
+          this.setState({formError: null});
+          this.setState({newData: data});
+          this.setState({isCreated: true});
+        }
        });
     }
   }
@@ -130,6 +138,11 @@ class NewProfileIndex extends React.Component {
     let pscid = null;
     let minYear = this.state.configData.startYear-this.state.configData.ageMax;
     let maxYear = this.state.configData.endYear-this.state.configData.ageMin;
+    let error = this.state.formError ? (
+      <div>
+        <h4 style={{color: 'red'}}>{this.state.formError}</h4>
+      </div>
+    ) : null;
     project =
       <SelectElement
         name = "project"
@@ -173,7 +186,7 @@ class NewProfileIndex extends React.Component {
         />;
     }
     if (!this.state.isCreated) {
-      profile =
+      profile = (
         <FormElement
           name = "newProfileForm"
           onSubmit = {this.handleSubmit}
@@ -219,7 +232,8 @@ class NewProfileIndex extends React.Component {
           {pscid}
           {project}
           <ButtonElement label = "Create" id = "button"/>
-        </FormElement>;
+        </FormElement>
+      );
     } else {
       profile =
         <div>
@@ -228,7 +242,11 @@ class NewProfileIndex extends React.Component {
           <p><a href = "/new_profile/" > Recruit another candidate </a></p>
         </div>;
     }
-    return (<Panel title="Create a new profile">{profile}</Panel>);
+    return (
+      <Panel title="Create a new profile">
+        {error}
+        {profile}
+      </Panel>);
   }
 }
 window.addEventListener(
--- a/modules/new_profile/jsx/NewProfileIndex.js
+++ b/modules/new_profile/jsx/NewProfileIndex.js
@@ -1,5 +1,7 @@
 import Panel from 'Panel';
 import Loader from 'Loader';
+import Swal from 'sweetalert2';
+
 /**
  * New Profile Form
  *
@@ -79,6 +81,12 @@ class NewProfileIndex extends React.Component {
       this.setState({
         errMessage: 'Date of Birth or EDC fields must match',
       });
+      Swal.fire({
+        title: 'User Error',
+        text: 'Date of Birth or EDC fields must match',
+        type: 'info',
+        confirmButtonText: 'Okay!',
+      });
     } else {
       this.setState({formError: null});
       let formData = this.state.formData;
@@ -98,6 +106,12 @@ class NewProfileIndex extends React.Component {
         .then((data) => {
           if (data.error) {
             this.setState({formError: data.error});
+            Swal.fire({
+              title: 'User Error',
+              text: data.error,
+              type: 'info',
+              confirmButtonText: 'Okay!',
+            });
           } else {
             this.setState({
               formError: null,
--- a/modules/new_profile/jsx/NewProfileIndex.js
+++ b/modules/new_profile/jsx/NewProfileIndex.js
@@ -74,39 +74,32 @@ class NewProfileIndex extends React.Component {
    */
   handleSubmit(e) {
     e.preventDefault();
-    const match = this.validateMatchDate();
-    if (!match) {
-      this.setState({
-        errMessage: 'Date of Birth or EDC fields must match',
-        isCreated: false,
-      });
-    } else {
-      let formData = this.state.formData;
-      let formObject = new FormData();
-      for (let key in formData) {
-        if (formData[key] !== '') {
-          formObject.append(key, formData[key]);
-        }
+    this.setState({formError: null});
+    let formData = this.state.formData;
+    let formObject = new FormData();
+    for (let key in formData) {
+      if (formData[key] !== '') {
+        formObject.append(key, formData[key]);
       }
-      fetch(this.props.submitURL, {
-        method: 'POST',
-        cache: 'no-cache',
-        credentials: 'same-origin',
-        body: formObject,
-        })
-      .then((resp) => resp.json())
-      .then((data) => {
-        if (data.error) {
-          this.setState({formError: data.error});
-        } else {
-          this.setState({
-            formError: null,
-            newData: data,
-            isCreated: true,
-          });
-        }
-       });
     }
+    fetch(this.props.submitURL, {
+      method: 'POST',
+      cache: 'no-cache',
+      credentials: 'same-origin',
+      body: formObject,
+      })
+    .then((resp) => resp.json())
+    .then((data) => {
+      if (data.error) {
+        this.setState({formError: data.error});
+      } else {
+        this.setState({
+          formError: null,
+          newData: data,
+          isCreated: true,
+        });
+      }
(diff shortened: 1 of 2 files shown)


kongtiaowang:
PHPCS

ridz1208:
@zaliqarosli please test/review

modules/new_profile/php/profile.class.inc
```diff
@@ -209,17 +209,17 @@ class Profile extends \NDB_Page
                 $site->getSiteAlias()
             )
             ) {
-                return array('error' => 'PSCID does not match the required structure.');
+                return ['error' => 'PSCID does not match the required structure.'];
```
zaliqarosli: i think we prefer array over []

modules/new_profile/jsx/NewProfileIndex.js
```diff
@@ -95,8 +96,15 @@ class NewProfileIndex extends React.Component {
         })
       .then((resp) => resp.json())
       .then((data) => {
-        this.setState({newData: data});
-        this.setState({isCreated: true});
+        console.log(data);
```
zaliqarosli: no console.logs please

modules/new_profile/jsx/NewProfileIndex.js
```diff
@@ -95,8 +96,15 @@ class NewProfileIndex extends React.Component {
...
+        if (data.error) {
+          console.log('error:' + data.error);
+          this.setState({formError: data.error});
+        } else {
+          this.setState({formError: null});
```
zaliqarosli: this set state can be called once for all three state updates

modules/new_profile/php/profile.class.inc
```diff
@@ -120,8 +135,126 @@ class Profile extends \NDB_Page
...
+                $centerIDs = $user->getData('CenterIDs');
+                $centerID  = $centerIDs[0];
+                $site      =& \Site::singleton($centerID);
+            } else {
+                // user has multiple sites, so validate
+                // PSCID against the Site selected.
+                $site =& \Site::singleton($values['site']);
+            }
+            if (empty($values['PSCID'])) {
+                return array('error' => 'PSCID must be specified.');
```
zaliqarosli: on testing, i keep getting the error 'PSCID must be specified.' even though `formData.pscid` is not empty. the value must not be appending to the request some how

Screenshot 2019-05-28 19 48 08

modules/new_profile/php/profile.class.inc
maltheism: Cannot use array here because of phpcs.

modules/new_profile/php/profile.class.inc
maltheism: Hey @zaliqarosli, can you test again?
I think the new commit should fix it. The backend logic was using all caps for PSCID and when it should've been lowercase to reflect the data from the frontend form.

modules/new_profile/php/profile.class.inc
johnsaigle:
```
$msg = 'PSCID does not match the required structure.';
return array('error' => $msg);
```

modules/new_profile/php/profile.class.inc
```diff
@@ -120,8 +135,126 @@ class Profile extends \NDB_Page
...
+     * Processes the values and saves to database
+     *
+     * @param array $values the user data received..
+     *
+     * @return array
+     */
+    function processValidity($values)
+    {
+        // Validate (dateTaken, dateTakenConfirm, sex) not set.
+        if (!isset($values['dateTaken'])
```
johnsaigle: isset accepts multiple arguments:

`if (!isset($values['dateTaken'], $values['dateTakenConfirm'], $values['sex']))...`

modules/new_profile/php/profile.class.inc
```diff
@@ -120,8 +135,126 @@ class Profile extends \NDB_Page
...
+        if (strcasecmp($values['sex'], 'Male') !== 0
+            && strcasecmp($values['sex'], 'Female') !== 0
+            && strcasecmp($values['sex'], 'Other') !== 0
+            && strcasecmp($values['sex'], 'Unknown') !== 0
+        ) {
+            return array('error' => 'Invalid sex option.');
+        }
+        // Validate PSCID.
+        $pscidSettings = $config->getSetting('PSCID');
+        $user          =& \User::singleton();
```
johnsaigle:
```suggestion
        $user          = \User::singleton();
```

modules/new_profile/php/profile.class.inc
```diff
@@ -120,8 +135,126 @@ class Profile extends \NDB_Page
...
+        if ($pscidSettings['generation'] == 'user') {
+            // Setup Database object.
+            $db = \Database::singleton();
+            if (empty($values['site'])) { // user is in only one site.
+                $centerIDs = $user->getData('CenterIDs');
+                $centerID  = $centerIDs[0];
+                $site      =& \Site::singleton($centerID);
```
johnsaigle:
```suggestion
                $site      = \Site::singleton($centerID);
```

modules/new_profile/php/profile.class.inc
```diff
@@ -120,8 +135,126 @@ class Profile extends \NDB_Page
...
+            } else {
+                // user has multiple sites, so validate
+                // PSCID against the Site selected.
+                $site =& \Site::singleton($values['site']);
```
johnsaigle:
```suggestion
                $site = \Site::singleton($values['site']);
```

modules/new_profile/php/profile.class.inc
```diff
@@ -120,8 +135,126 @@ class Profile extends \NDB_Page
                 ->withStatus(405)
                 ->withHeader("Allow", "GET,POST");
         }
+    }
 
+    /**
+     * Processes the values and saves to database
```
johnsaigle: This function comment isn't accurate. Can you update it? Also the params and return values could use more detail.

johnsaigle:
@maltheism Can you also include changes to the Test Plan as Rida requested in the linked issue?

modules/new_profile/php/profile.class.inc
```diff
@@ -120,8 +135,126 @@ class Profile extends \NDB_Page
...
+     *
+     * @param array $values the user data received..
+     *
+     * @return array
+     */
+    function processValidity($values)
```
johnsaigle:
```suggestion
    function processValidity(array $values): array
```

modules/new_profile/php/profile.class.inc
maltheism: @johnsaigle okay two lines instead :)

modules/new_profile/php/profile.class.inc
maltheism: Hey @johnsaigle, how do you do this with not having phpcs complain about the line being too long? I've tried line breaks but then it complains about needing a boolean and other stuff. Could you try it and see if there is a way to satisfy phpcs.

johnsaigle: You can just leave it if PHPCS is having a big issue with it.

maltheism:
I think the module still reflects the test plan. @johnsaigle what should be changed about the Test Plan?

modules/new_profile/php/profile.class.inc
driusan: you could also split the newlines on comma (unless PHPCS complains about that.)

maltheism:
Hey @zaliqarosli, new commit should fix the above error. Please let me know after testing.

johnsaigle:
> Also, the test plan should probably be amended to include an item to test the backend validation independently from the front end

from the linked issue

modules/new_profile/jsx/NewProfileIndex.js
```diff
@@ -74,39 +74,32 @@ class NewProfileIndex extends React.Component {
    */
   handleSubmit(e) {
     e.preventDefault();
-    const match = this.validateMatchDate();
-    if (!match) {
-      this.setState({
-        errMessage: 'Date of Birth or EDC fields must match',
```
zaliqarosli: you just removed the errors. where did they go?

modules/new_profile/php/profile.class.inc
```diff
@@ -120,8 +135,128 @@ class Profile extends \NDB_Page
...
+        if ($config->getSetting('useEDC') == 'true'
+            && isset($values['edcDateTaken'])
+            && isset($values['edcDateTakenConfirm'])
+        ) {
+            if (strlen($values['edcDateTaken']) > 2
+                && !$this->_checkDate($values['edcDateTaken'])
+            ) {
+                return array('error' => 'EDC is not a valid date.');
+            } else if ($values['edcDateTaken'] != $values['edcDateTakenConfirm']) {
+                return array('error' => 'Estimated Due date fields must match.');
```
zaliqarosli: this should say "Expected Date of Confinement fields must match."

EDC does not stand for "estimated due date"

modules/new_profile/php/profile.class.inc
```diff
@@ -120,8 +135,128 @@ class Profile extends \NDB_Page
...
+            return array('error' => 'Incomplete candidate fields.');
+        }
+        // Validate dob.
+        if (strlen($values['dateTaken']) > 2
+            && !$this->_checkDate($values['dateTakenConfirm'])
+        ) {
+            return array('error' => 'Birth date is not a valid date.');
+        } else if ($values['dateTaken'] != $values['dateTakenConfirm']) {
+            // Validate (dob1 & dob2) are not equal.
+            return array('error' => 'Birth date confirmation mismatch.');
```
zaliqarosli: this should also say "Date of Birth fields must match." for clarity!

modules/new_profile/php/profile.class.inc
```diff
@@ -120,8 +135,128 @@ class Profile extends \NDB_Page
...
+     * @return array the verified parsed POST response.
+     */
+    function processValidity(array $values): array
+    {
+        // Validate (dateTaken, dateTakenConfirm, sex) not set.
+        if (!isset($values['dateTaken'])
+            || !isset($values['dateTakenConfirm'])
+            || !isset($values['sex'])
+        ) {
+            return array('error' => 'Incomplete candidate fields.');
```
zaliqarosli: this is only checking for if dob and sex fields are set but all the fields are required on the front-end so shouldn't they also all be required in the back-end?

zaliqarosli: except for edc because that depends on configuration

johnsaigle: And it would be good to add an EDC check here too if LORIS is configured that way.

modules/new_profile/jsx/NewProfileIndex.js
zaliqarosli: i see now that you're opting for the php errors instead of defining it in react which is fine! i'm wondering though if @ridz1208 cares about the validation to also be done on the front-end.

if not, you can remove the validateMatchData() method as it's not being used!

modules/new_profile/php/profile.class.inc
zaliqarosli: true!

modules/new_profile/jsx/NewProfileIndex.js
ridz1208: I do, validation can definitely be done on the front end !!  But it must also exist on the backend in case the front end is... incapacitated

maltheism:
Hey @ridz1208, I added swal errors for you.

kongtiaowang:
PHPCS

maltheism:
Thanks @kongtiaowang, I believe the PHPCS should be fixed by the latest commit.

driusan:
@zaliqarosli can you retest/review?

zaliqarosli:
new PR https://github.com/aces/Loris/pull/4834 that uses the NDB_Form back-end validation

driusan:
Superceded by #4834

PR #4695, opened 2019-05-27
[Dicom Archive + Candidate List] Skip test causing transitory failures

johnsaigle:
### Brief summary of changes

This test seems to fail sometimes without a clear cause, forcing us to restart Travis. This PR skips it for now and marks it as needing work.


driusan:
Can you comment out the TestPSCIDLink one from candidate list? I see that fail without cause more often than this one.

PR #4699, opened 2019-05-27
[login] check for X-FORWARDED-FOR before REMOTE_ADDR

ridz1208:
### Brief summary of changes
the `user_login_history` table of the database has been logging the IP address of the proxy instead of the real IP of the user. to prevent that, we are checking for an IP address set in the `X-FORWARDED-FOR` variable and only if that is NULL we are fetching the `REMOTE_ADDR` and setting it.

### DIFF
--- a/php/libraries/SinglePointLogin.class.inc
+++ b/php/libraries/SinglePointLogin.class.inc
@@ -239,7 +239,8 @@ class SinglePointLogin
                      'Failcode'        => null,
                      'Login_timestamp' => date('Y-m-d H:i:s'),
                      'IP_address'      => $_SERVER['HTTP_X_FORWARDED_FOR']
-                         ?? $_SERVER['REMOTE_ADDR'],
+                         ?? $_SERVER['REMOTE_ADDR']
+                         ?? 'Unknown',
                      'Page_requested'  => isset($_SERVER['REQUEST_URI'])
                                               ? $_SERVER['REQUEST_URI']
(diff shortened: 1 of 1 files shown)


driusan:
@jkat can you review/test that this fixes your problem with the IP address in the history?

php/libraries/SinglePointLogin.class.inc
```diff
@@ -238,9 +238,8 @@ class SinglePointLogin
                      'Success'         => 'Y',
                      'Failcode'        => null,
                      'Login_timestamp' => date('Y-m-d H:i:s'),
-                     'IP_address'      => isset($_SERVER['REMOTE_ADDR'])
-                                              ? $_SERVER['REMOTE_ADDR']
-                                              : 'Unknown',
+                     'IP_address'      => $_SERVER['HTTP_X_FORWARDED_FOR']
+                         ?? $_SERVER['REMOTE_ADDR'],
```
driusan:
```suggestion
                         ?? $_SERVER['REMOTE_ADDR'],
                         ?? 'Unknown',
```

Add fallback on unknown back.

Jkat:
```suggestion
                         ?? $_SERVER['REMOTE_ADDR']
                         ?? 'Unknown',
```

Extra comma in syntax

ridz1208:
@driusan @Jkat done

PR #4697, opened 2019-05-27
Merge bugfix branch into release branch

driusan:
This pulls the latest bugfix changes into the 21.0-release branch.

Note that there were conflicts in php/libraries/User.class.inc and modules/dataquery/jsx/reacts.tabs.js. Extra care should be taken reviewing file changes from those files.

PR #4700, opened 2019-05-27
Push bugfix into 21.0.0

ridz1208:
This pulls the latest bugfix changes into the 21.0-release branch.

Note that there were conflicts in php/libraries/User.class.inc and modules/dataquery/jsx/reacts.tabs.js. Extra care should be taken reviewing file changes from those files.

### DIFF
--- a/modules/issue_tracker/jsx/index.js
+++ b/modules/issue_tracker/jsx/index.js
@@ -4,12 +4,12 @@ import IssueForm from './IssueForm';
  * Render IssueForm on page load
  */
 $(function() {
-  const args = QueryString.get();
+  var id = location.href.split("/issue/")[1];
   const issueTracker = (
     <div className="page-issue-tracker">
       <IssueForm
         Module="issue_tracker"
-        DataURL={`${loris.BaseURL}/issue_tracker/ajax/EditIssue.php?action=getData&issueID=${args.issueID}`}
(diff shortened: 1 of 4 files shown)


driusan:
approvals were on #4697, so I'm just merging this once it passes.

kongtiaowang:
/home/travis/build/aces/Loris/modules/issue_tracker/jsx/index.js
       7:3   error    Unexpected var, use let or const instead       no-var
       7:32  error    Strings must use singlequote                   quotes

modules/issue_tracker/jsx/index.js
```diff
@@ -4,12 +4,12 @@ import IssueForm from './IssueForm';
  * Render IssueForm on page load
  */
 $(function() {
-  const args = QueryString.get();
+  var id = location.href.split("/issue/")[1];
```
kongtiaowang: const instead fo var

ridz1208: @HenriRabalais does that looks like it makes sense to you Mr issue tracker guy ?

maltheism: @ridz1208 var isn't supposed to be used anymore. The two options are either const or let and with const being what should be used in this case like @kongtiaowang suggested.

Issue #4704, opened 2019-05-28
[Publication] Email filter incorrectly fails on addresses containing '+'


johnsaigle:
### Problem

The email `john.saigle+test@mcin.ca` is valid but the front-end validation marks it as invalid.

![bademailfilter](https://user-images.githubusercontent.com/4022790/58495604-ee2a8000-8145-11e9-8f99-9982a5271701.png)

### To reproduce
(User must have "Publication - Propose a project" permission)
Complete step 5 of the test plan:
> 5. Fill out all the fields in the form and try to submit. Make sure you have
a corresponding directory to the directory specified in Config if you are 
attempting a file upload. Try submitting without filling in required fields.

Add a collaborator during the above step. When prompted to add their email, enter an email like the example above.

PR #4730, opened 2019-05-29
[publication] 21.0 release publication email verify fix

maltheism:
### Brief summary of changes

Incorporated the frontend browser check for validation of email by changing type to email.

### This resolves issue...

- [ ] Redmine? #

- [x] Github? #4704

### To test this change...

- [x] Test the publication module with an email containing the '+' symbol.


modules/publication/jsx/uploadForm.js
```diff
@@ -27,12 +27,8 @@ class PublicationUploadForm extends React.Component {
...
   validateEmail(field, email) {
     let formErrors = this.state.formErrors;
-
-    // don't supply error if email is blank
-    if (email === '' || email === null || email === undefined) {
-      delete formErrors[field];
-      // if email is invalid, set error, else nullify error
-    } else if (/^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/.test(email) === false) {
+    const emailDiv = document.getElementById(field);
+    if (!emailDiv.checkValidity()) {
```
driusan: this doesn't appear to be defined?

jsx/Form.js
```diff
@@ -1365,6 +1365,57 @@ StaticElement.defaultProps = {
...
+ * Email element component.
+ * Used to displays plain/formatted email input as part of a form
+ *
+ * To pass a formatted text, you need to wrap it in a single parent element.
+ * Example usage:
+ *
+ * ```
+ * let myText = (<span>This is my <b>text</b></span>);
+ * <EmailElement
+ *    text={myText}
```
driusan: why not use this.props.children? That would make this cleaner, I think

modules/publication/jsx/viewProject.js
```diff
@@ -257,10 +247,11 @@ class ViewProject extends React.Component {
           label="Lead Investigator"
           text={this.state.formData.leadInvestigator}
         />
-        <StaticElement
-         name="leadInvestigatorEmail"
-         label="Lead Investigator Email"
-         text={this.state.formData.leadInvestigatorEmail}
+        <EmailElement
+          id={'leadInvestigatorEmail'}
+          name="leadInvestigatorEmail"
```
zaliqarosli: please keep consistent with double/single quotes and i don't think the value of id needs curly brackets too

jsx/Form.js
```diff
@@ -1365,6 +1365,57 @@ StaticElement.defaultProps = {
...
+    super(props);
+  }
+  render() {
+    return (
+      <div className="row form-group">
+        <label className="col-sm-3 control-label">
+          {this.props.label}
+        </label>
+        <div className="col-sm-9">
+          <input className="form-control-static" type={'email'} id={this.props.id}>
```
zaliqarosli: i think just 'email' without the curlies is sufficient no?

jsx/Form.js
```diff
@@ -1365,6 +1365,57 @@ StaticElement.defaultProps = {
...
+            {this.props.text}
+          </input>
+        </div>
+      </div>
+    );
+  }
+}
+
+EmailElement.propTypes = {
+  label: PropTypes.string,
```
zaliqarosli: you can add name prop

modules/publication/jsx/projectFields.js
```diff
@@ -39,7 +39,7 @@ class EmailElement extends React.Component {
         </label>
         <div className="col-sm-7">
           <input
-            type="text"
+            type="email"
```
zaliqarosli: i'm a bit confused here. do you also want to remove this EmailElement component out of this jsx file?

jsx/Form.js
```diff
@@ -1365,6 +1365,57 @@ StaticElement.defaultProps = {
...
 
+/**
+ * Email element component.
+ * Used to displays plain/formatted email input as part of a form
+ *
+ * To pass a formatted text, you need to wrap it in a single parent element.
+ * Example usage:
+ *
+ * ```
+ * let myText = (<span>This is my <b>text</b></span>);
```
zaliqarosli: im curious .. whats the reason behind allowing formatted text? and what is this text attribute? is it just the value of the field? i am wondering why it's not going into a 'value' attribute like the others

jsx/Form.js
maltheism: @zaliqarosli My IDE does that by default, I don't think it matters but removed.

modules/publication/jsx/projectFields.js
maltheism: @zaliqarosli I think I left this because the component has different column than the other component.

modules/publication/jsx/uploadForm.js
maltheism: @driusan I think I modified the previous code to be similar but the element emailDiv should be defined from the 'field' parameter that's passed into validateEmail.

driusan: I mean the function `checkValidity` isn't defined.

modules/publication/jsx/viewProject.js
```diff
@@ -257,10 +247,11 @@ class ViewProject extends React.Component {
           label="Lead Investigator"
           text={this.state.formData.leadInvestigator}
         />
-        <StaticElement
-         name="leadInvestigatorEmail"
-         label="Lead Investigator Email"
-         text={this.state.formData.leadInvestigatorEmail}
+        <EmailElement
```
zaliqarosli: it looks to me that this EmailElement is the one from jsx/Form.js which is an input element `<input type='email'/>` except this use of it is replacing a StaticElement (a display) and not an input element?

jsx/Form.js
```diff
@@ -1365,6 +1365,48 @@ StaticElement.defaultProps = {
   text: null,
 };
 
+/**
+ * Email element component.
+ * Used to displays plain/formatted email input as part of a form.
```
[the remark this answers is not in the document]
maltheism: @zaliqarosli I'll redefine the comment here. Also to answer your other comment. I left the module alone without using this component that was made with this PR because of the css is different than a typical form. I didn't feel this PR was targeting that but just the bug of solving correct email verification.

modules/publication/jsx/viewProject.js
maltheism: @zaliqarosli I'm confused how you phrased this question. The reason for the EmailElement is similar to why there was a request for a unique password component in the react login PR.

modules/publication/jsx/uploadForm.js
maltheism: @driusan See https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/checkValidity

modules/publication/jsx/viewProject.js
zaliqarosli: is the password component you're mentioning an input element or a display element?

jsx/Form.js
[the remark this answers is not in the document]
zaliqarosli: > I left the module alone without using this component that was made with this PR because of the css is different than a typical form.

I don't understand why a component is made in this PR but then is not used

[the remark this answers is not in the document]
zaliqarosli: perhaps this PR includes more code changes than is needed to target the actual bugfix? I'm not quite sure what I'm reviewing here. it looks to me that there are 2 different EmailElement components with the same name, i'm unsure which one is being used, and when it is being used, its to replace a StaticElement (a display) and not an input element 😕

modules/publication/jsx/viewProject.js
maltheism: Hey @zaliqarosli, the PR for upgrading the Login module to react is what i'm referring to and when I'm referencing the new password component in that PR (you suggested it). Instead of modifying one of the components to add a property that would change the text to be type "password" it was suggested to just make another component. I thought similar for creating an email component because of the type being email.

zaliqarosli: okay great. I agree that we should be making separate components. Can you point me to where in the code in this PR that you are using this new EmailComponent that you added to jsx/Form.js?

maltheism: I'll delete the component then. As your above comment mentions it confuses the review process.

jsx/Form.js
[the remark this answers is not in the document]
maltheism: Hi @zaliqarosli, I'll delete the component then. As your comment mentions it confuses the review process.

modules/publication/jsx/viewProject.js
zaliqarosli: no need to delete it. if it's not being used in this PR, then I think it's worth it for you to open up a separate PR for this new EmailElement component and include a use case there

maltheism: Hi @zaliqarosli, I removed the component from this PR.

zaliqarosli: this StaticElement should stay a static element right?

modules/publication/jsx/projectFields.js
```diff
@@ -403,11 +411,15 @@ class ProjectFormFields extends React.Component {
...
-        <StaticElement
-          text={voiHelp}
-        />
+        <div className="row form-group">
+          <label className="col-sm-3 control-label">
+          </label>
+          <div className="col-sm-9">
+            <div className="form-control-static">{voiHelp}</div>
+          </div>
+        </div>
```
zaliqarosli: what's the reason for this change? it looks to me that the html here is the same as is in the StaticElement render function

modules/publication/jsx/uploadForm.js
```diff
@@ -195,14 +191,18 @@ class PublicationUploadForm extends React.Component {
...
+            Propose a new project
+          </h3>,
+          <TextboxElement
+            name="title"
+            label="Title"
+            onUserInput={this.setFormData}
+            required={true}
+            value={this.state.formData.title}
+          />,
+        </div>,
```
zaliqarosli: createElements is an array of jsx. i don't think you need the div that you added here and it's also incorrect syntax right? you're nesting elements of an array within div tags.

modules/publication/jsx/viewProject.js
```diff
@@ -25,16 +25,6 @@ class ViewProject extends React.Component {
 
   validateEmail(field, email) {
     let formErrors = this.state.formErrors;
-
-    // don't supply error if email is blank
```
zaliqarosli: what has this validation been replaced with?

modules/publication/jsx/uploadForm.js
maltheism: @zaliqarosli The extra div is fixing the error "Warning: Each child in an array or iterator should have a unique "key" prop." from displaying in the console. I'll change the formatting to be createElement = () instead of an array. Currently we don't have a way to supply a key to TextboxElement or I would find the extra div unnecessary

modules/publication/jsx/projectFields.js
maltheism: @zaliqarosli There is a slight change if you compare the render of StaticElement to my change. Turns out this change solved the console error: "Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>." but maybe I should modify the StaticElement and get rid of the use of <p>. What do you think?

zaliqarosli: could you try changing to <div> tags in `voiHelp` to <span>? I think this would be the best solution

modules/publication/jsx/viewProject.js
zaliqarosli: hey @maltheism, what's your solution for this? the validation is still missing

modules/publication/jsx/projectFields.js
maltheism: @zaliqarosli sure why not

modules/publication/jsx/viewProject.js
maltheism: Hey @zaliqarosli I'm not sure why the validateEmail is showing here on GitHub. I've removed the validateEmail function here if you checkout the PR and look at the code. The validation for the email is being done in the uploadForm.js

zaliqarosli: I know that you've removed the validateEmail function here. my question is why? shouldn't we keep the validation here?

it looks to me that validateEmail prop is to be supplied to the `<ProjectFormFields />` component of which there are two - one in `viewProject.js` and one in `uploadForm.js`. it looks to me that validateEmail function needs to be defined in both` viewProject.js` and `uploadForm.js`.

You are removing the validateEmail function from `viewProject.js`. If it looks to you that `viewProject.js` doesn't need the validateEmail function, please explain to me why. And explain to me why the `<ProjectFormFields />` component in this same class doesn't need the validateEmail prop (which you are also removing)

maltheism: @zaliqarosli Thanks for all the help!

maltheism:
Updated PR to just use type=email for the validation because the browser handles it when type is email.

modules/publication/jsx/viewProject.js
```diff
@@ -258,9 +242,10 @@ class ViewProject extends React.Component {
           text={this.state.formData.leadInvestigator}
         />
         <StaticElement
-         name="leadInvestigatorEmail"
-         label="Lead Investigator Email"
-         text={this.state.formData.leadInvestigatorEmail}
+          id='leadInvestigatorEmail'
```
zaliqarosli: i think you can get ride of this id too

maltheism: thanks @zaliqarosli it should be fixed now.

Issue #4742, opened 2019-05-30
[User account] Creating new user with a user name containing a space character gives a 500 error


nicolasbrossard:
As the title says. The user name should be validated and an error displayed on the front-end instead of the HTTP 500 error. The funny thing is, the user was successfully created anyway!

PR #4744, opened 2019-05-30
[User account] Fixes user name containing a space

maltheism:
### Brief summary of changes

Solves the 500 error for username with spaces.

### This resolves issue...

- [ ] Redmine? #

- [x] Github? #4742

### To test this change...

- [x] Make user containing a space and see issue details for more information.

### DIFF
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -80,7 +80,7 @@ class Edit_User extends \NDB_Form
         $defaults = array();
 
         if (!$this->isCreatingNewUser()) {
-            $user = \User::factory($this->identifier);
+            $user = \User::factory(urldecode($this->identifier));
             // get the user defaults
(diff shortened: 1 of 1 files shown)


nicolasbrossard:
@driusan do we want to allow user names that contain spaces? Are there any special characters that are not allowed?

johnsaigle:
It might be good to limit it to alphanumeric characters for simplicity.

driusan:
I don't know if we want to allow spaces. It seems like a security problem that could let people impersonate other people if not handled carefully.

nicolasbrossard:
@maltheism can you please update your PR so that user names with non-alphanumeric characters are not allowed? I'll review/test once you are done

ridz1208:
I'm with @driusan on this one, spaces in usernames are not a good idea.

nicolasbrossard:
Has been replaced by #4888

PR #4757, opened 2019-06-03
[core] 2019 06 03 fix id generation

PapillonMcGill:
### Brief summary of changes
Sequential ID generation fail when creating a minimum value ID.  Generate a PHP fatal error.

### To test this change...
Create a new site,  add a minimum value parameter in config.xml file `"min"=1` in sequential generation for PSCID and create a first candidate for that new site.


PapillonMcGill:
need to be against 21.0-release

PR #4758, opened 2019-06-03
[core] 2019 06 03 fix id generation

PapillonMcGill:
### Brief summary of changes
fix creation of first ID in sequential mode with min value. 

### This resolves issue...
Currently create a php fatal error

### To test this change...
- create a new test site
- set PSCID generation to sequential with min value `"min"=1`
- create a first candidate for this site PSCID should be `<sitealias>0001` if `"minLength"=4` as been set.

### DIFF
--- a/php/libraries/IdentifierGenerator.php
+++ b/php/libraries/IdentifierGenerator.php
@@ -86,7 +86,7 @@ abstract class IdentifierGenerator
         if (count($this->getExistingIDs()) < 1) {
             return str_pad(
                 strval($this->minValue),
-                $this->length,
+                is_null($this->length) ? 4 : $this->length,
                 "0",
                 STR_PAD_LEFT
             );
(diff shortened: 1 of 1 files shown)


php/libraries/IdentifierGenerator.php
```diff
@@ -84,7 +84,12 @@ protected function generateSequentialID(string $id = ''): string
     {
         // If this is the first ID ever created, return the minimum value.
         if (count($this->getExistingIDs()) < 1) {
-            return $this->minValue;
+            return str_pad(
+                strval($this->minValue),
+                $this->length,
```
driusan: `$this->length` doesn't seem to get set anywhere in this class, can you add a check to make sure it's an integer and throw an exception before this?

PapillonMcGill: @driusan It is set in the extended class. I have tested this PR for the CRU, It is there that I discovered the bug.

driusan: It being set in a derived class doesn't mean it's set in every derived class. Can you please add a check to throw a reasonable exception if it's not set?

PR #4759, opened 2019-06-03
[Login page] Fix notice about undefined index REDIRECT_URL

driusan:
Fix warning printing on login page about 'Undefined index REDIRECT_URL'
when run under the built in PHP web server. REDIRECT_URL is only available
under apache.


zaliqarosli:
@kongtiaowang i can't seem to get this environment working for me to test / replicate error. is this something you could test for me shen?

driusan:
This was fixed by #4761 but not autoclosed because the merge was squashed.

PR #4763, opened 2019-06-03
[Issue Tracker] Fix incorrect links to individual issues

driusan:
There was a regression introduced in #4112 where the issue tracker
links reverted to getting the ID from a query parameter instead
of the URL. This was done inconsistently, and as a result some
places loaded URLs such as "...&issueID=?issueID=1". This fixes
the links to parse the ID from the URL, so that issues can now
be loaded with the php development web server.


mohadesz:
@driusan how should I test this change?

driusan:
@mohadesz no.. go to the issue tracker, click on a link to an issue, see if the issue loads. For me it didn't, and this fixed it.

PR #4765, opened 2019-06-03
[Data Provisioner] Throw more explicit exception when CenterID is null

driusan:
When a CenterID is null, the code falls through to the exception, but
the error message is misleading. The resource had a getCenterID, it
just returned null. Add a more explicit error message.


kongtiaowang:
@driusan you are right, it is the RB_data issue. There is one piece of data without any site.

Issue #4722, opened 2019-05-28
[imaging_browser] click on QC comments on image returns blank page


coreyAbout:
TestPlan Item B14. `Verify Visit-level QC controls and comments can be deleted/unchecked/emptied and saved`

Problem:
Click on QC comments on image returns blank page
e.g., ROM184_400184_V3: loris_400184_V3_t1_001.mnc
`This page isn’t working test.loris.ca is currently unable to handle this request.
HTTP ERROR 500`

driusan:
Are you sure this isn't a problem with test.loris.ca? Can you try on another sandbox?

coreyAbout:
> Are you sure this isn't a problem with test.loris.ca? Can you try on another sandbox?

Haven't tried yet. Currently developing the instance for Douglas, switching to raisinbread involves switching the code base, the database tables, and all the kinds of imaging files. Need to switch back and forth, haven't figured out best way to do it now.

Can you try that on your sandbox?

johnsaigle:
Addressed by #4734

zaliqarosli:
@johnsaigle just to double-check, this issue can be closed right?

cmadjar:
Now the QC comments show the page but the check boxes are missing. This is still an issue to be fixed. I'll add the "Critical to release" Tag as not being able to do QC seems like an important bug and remove the PR sent tag since this still needs fixing

cmadjar:
I removed the tag critical to release since it was a bug in the raisin bread dataset and not the code. I sent a PR to modify the RB dataset on the LORIS code to fix the issue.

zaliqarosli:
thanks @cmadjar !

driusan:
fixed by above PR

PR #4782, opened 2019-06-05
Fix RB feedback MRI links between MRI feedback tables

cmadjar:
### Brief summary of changes

During testing, the QC comments did not show correctly (could not see the check boxes). In this PR, the ID for the predefined comments and comment types on the MRI side are harmonize so that they link to the `feedback_mri_comments` table

### This resolves issue...

- [x] Github #4722 which ended up being a raisin bread issue

### To test this change...

- [ ]


cmadjar:
@ridz1208 These changes will need to be pulled on demo as this is broken as well. I will update the table on the demo site so that the bug is fixed for the conference.

Issue #4713, opened 2019-05-28
[imaging_browser] Filter "Pending new" Blank option not working


coreyAbout:
TestPlan Item A3. `Test that all Filters work. Upon first loading, Site filter should be set to 'All User Sites' (with the dropdown menu populated with the sites the logged on user is affiliated with) IFF user does not have permission imaging_browser_view_allsites nor imaging_browser_phantom_allsites, and is set to 'All' otherwise.`

Problem:
When click on the blank option of `Pending new`, which should includes all the options, but it functions the same as selection the `new` option.

driusan:
Since it's a multiselect, I think the solution here is to 1. Fix #4794 and 2. Remove the blank option.

I'm not sure what it means to choose explicitly "blank" in a multiselect filter.

cmadjar:
This got fixed by the referenced PR and merged so closing the ticket

PR #4795, opened 2019-06-06
[JSX] Improve handling of empty multilselects

### LINKED ISSUE #4794: [JSX] Multiselect filters show 0 results if unselected
If you select a multiselect filter, and then unselect it in a menu, the result is that there are 0 rows displayed, not the expected results of the number of rows before having selected anything.

driusan:
This fixes 2 bugs relating to the handling of how "nothing" is handled in multiselects.

1. The empty option is removed, because it has undefined semantics and the way that it works for single selects (ie. act as no filter) is better expressed by simply unselecting everything
2. A bug regarding unselecting everything from a multiselect is fixed. Before this, it was filtering out all results, instead of resetting the filter to not filter anything.

Fixes #4794, #4713

### DIFF
--- a/jsx/Filter.js
+++ b/jsx/Filter.js
@@ -69,7 +69,7 @@ class Filter extends Component {
           element = <SelectElement key={filter.name} options={filter.options}/>;
           break;
         case 'multiselect':
-          element = <SelectElement key={filter.name} options={filter.options} multiple={true}/>;
+          element = <SelectElement key={filter.name} options={filter.options} multiple={true} emptyOption={false}/>;
           break;
         case 'date':
           element = <DateElement key={filter.name}/>;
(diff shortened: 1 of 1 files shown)


driusan:
@cmadjar I had a quick look and the datadict module doesn't seem to be using the FilterableDataTable, which is why it's not fixed by this. I think updating the datadict module to use the new format is out of scope for this.

jsx/Filter.js
```diff
@@ -40,7 +40,7 @@ class Filter extends Component {
     const searchParams = new URLSearchParams(location.search);
     const filter = JSON.parse(JSON.stringify(this.props.filter));
     const exactMatch = (!(type === 'textbox' || type === 'date'));
-    if (value === null || value === '') {
+    if (value === null || value === '' || (value.constructor === Array && value.length === 0)) {
```
driusan: @zaliqarosli can you review this part of it too to make sure it makes sense?

PR #4692, opened 2019-05-23
[Imaging Browser] Sequence type column

kongtiaowang:
to resolve: 
https://github.com/aces/Loris/issues/4635

PR #4797, opened 2019-06-06
[Imaging Browser] Hide Sequence Type in data table

driusan:
The imaging browser data table had a "Sequence Type" column
added by #4692.

This was unintentional, as the data is only there for the
filter. The table does not properly format/display it and
will likely not scale to large protocols.

This sets the column back to hidden, but the filter should
still work.


cmadjar:
Note: this resolves #4713

Issue #4784, opened 2019-06-05
[imaging_browser] Drop down QC MRI comments not saving


cmadjar:
Test Plan section D. MRI-QC: 
23. Comments save, checkboxes save, dropdown values save

=> After testing the QC comment box on my sandbox I noticed that the drop down in the MRI QC pop up window does not save. Everything else saves.

cmadjar:
@driusan I put the Critical to release tag here as saving the comments seem like something quite important. Feel free to change it if you don't agree.

cmadjar:
Note to self. This is indeed an issue with 21.0 code as this is working on demo with the same dataset and 20.3.

driusan:
@cmadjar seems like a critical bug to me too.

driusan:
Digging into it a little, it looks like they're saving in the backend, they're just not loading/being displayed

cmadjar:
indeed, it got saved in the backend. Your PR fixed the frontend. Thanks!

driusan:
Fixed by above PR

PR #4801, opened 2019-06-06
[Imaging Browser] Fix loading of feedback popup for non-QC users

### LINKED ISSUE #4785: [imaging_browser] QC info not viewable if do not have permission to edit the QC
Test Plan section D. MRI-QC:
22. Viewable by all, editable IFF permission imaging_browser_qc

=> When trying to view the QC information with a user that do not have permission imaging_browser_qc, we get a 403 error.

This error got inserted here: https://github.com/aces/Loris/commit/16bb75f7cd813915d43bb61c9bd08329c8b7203c#diff-44bcdd7c78b4cad3c9529f51aa8598f9

driusan:
This should fix item 22 of the imaging browser test plan -- the
page should be viewable by all, regardless of imaging_browser_qc
permission.

Fixes #4785

### DIFF
--- a/htdocs/feedback_mri_popup.php
+++ b/htdocs/feedback_mri_popup.php
@@ -32,7 +32,8 @@ if (!$user->hasPermission('imaging_browser_qc')
     return;
 }
 
-$tpl_data['has_permission'] = true;
+$tpl_data['has_permission'] = $user->hasPermission('imaging_browser_qc');
+
 // instantiate feedback mri object
 $comments = new FeedbackMRI($_REQUEST['fileID'] ?? '', $_REQUEST['sessionID'] ?? '');
 
--- a/htdocs/feedback_mri_popup.php
+++ b/htdocs/feedback_mri_popup.php
@@ -18,8 +18,6 @@ if ($client->initialize() == false) {
     return;
 }
 
-require_once "FeedbackMRI.class.inc";
-
 // create DB object
 $DB = \Database::singleton();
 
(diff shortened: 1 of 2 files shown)


driusan:
@cmadjar can you test this? I don't have QC data and can't create any because of #4784, but the page now loads for me without the permission.

htdocs/feedback_mri_popup.php
```diff
@@ -18,16 +18,16 @@
     return;
 }
 
-require_once "FeedbackMRI.class.inc";
```
driusan: This is in php/libraries. It's loaded by the autoloader on line 13

htdocs/router.php
```diff
@@ -10,13 +10,14 @@
...
-
-
 $url     = ltrim($_SERVER['REQUEST_URI'], "/");
 $request = $_SERVER['REQUEST_URI'];
 
 if ($request != '/'
-    && file_exists(__DIR__ . $request)
+    && (
+        file_exists(__DIR__ . $request)
+        || file_exists(__DIR__ . "/" . $urlpath)
```
driusan: this should probably just be the $urlpath line, but I didn't want to risk breaking anything else this late in the release cycle, so I just kept both.

driusan:
@cmadjar that looks like a regression that was introduced at the same time.. I think it should be fixed by the latest commit.

driusan:
Not sure why this is failing travis, I just merged #4761 into it to see if that fixes it (locally for me nothing works without it, and this branch works with php -S with it..)

driusan:
Seems to be passing Travis with FixAjaxRouter merged in, so adding blocked tag for now.

Issue #4779, opened 2019-06-05
[Document repository] Downloading documents don't work


zaliqarosli:
i believe @johnsaigle wants to fix this

driusan:
Merged

PR #4804, opened 2019-06-06
[document_repository] fix - can't download pdf and image files.

kongtiaowang:
to solve https://github.com/aces/Loris/issues/4779
to test: download an image file or a pdf file.


modules/document_repository/php/files.class.inc
```diff
@@ -144,6 +139,25 @@ class Files extends \NDB_Page
...
+    /**
+     * Handles the document download process
+     *
+     * @param string $file the file name
+     *
+     * @return void
+     */
+    function download(string $file): void
+    {
+        header('Content-Description: File Transfer');
```
driusan: This is a big step backwards.. we shouldn't be taking something that's already returning a PSR7 response and regressing to using side-effects and exit. If there's a problem with the withBody() part of the response, that should be fixed

modules/document_repository/php/files.class.inc
```diff
@@ -119,13 +119,13 @@ class Files extends \NDB_Page
...
             if (!is_numeric($record)) {
                 $file = __DIR__ . "/../user_uploads/$name/$record";
+                readfile($file);
                 return (new \LORIS\Http\Response())
-                ->withHeader(
-                    'Content-Disposition',
-                    'attachment; filename=' . basename($file)
-                )
-                ->withHeader('Content-Transfer-Encoding', 'Binary')
-                ->withBody(new \LORIS\Http\StringStream(readfile($file)));
```
driusan: I'm pretty sure Zend Diactoros has a stream type for dealing with files (probably https://github.com/zendframework/zend-diactoros/blob/master/src/Stream.php), why not make a LORIS wrapper around similar to https://github.com/aces/Loris/blob/minor/src/Http/Response.php ?

src/Http/FileStream.php
```diff
@@ -0,0 +1,268 @@
...
+ * A FileStream provides a simple wrapper over a file to convert
+ * it to a \Psr\Http\Message\StreamInterface.
+ *
+ * @category PSR7
+ * @package  Http
+ * @author   Shen Wang <shen.wang2@mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+class FileStream implements \Psr\Http\Message\StreamInterface
```
xlecours: This should only extend `Zend\Diactoros\Stream` instead of copying its content.
See: https://github.com/aces/Loris/blob/minor/src/Http/Response.php for example

xlecours:
What I meant was:

```
class FileStream extends \Zend\Diactoros\Stream implements \Psr\Http\Message\StreamInterface
{
}
```

with nothing inside the class

xlecours:
tested. It pass

Issue #4781, opened 2019-06-05
[MRI violations] Scan appears three times in 'resolved' tab


nicolasbrossard:
1. I access the MRI violations module and search for scans for which the type of problem is 'Could not identify scan type'.
2. I select a resolution type (any) and click 'Save'
3. I browse the 'Resolved' tab of the MRI and I see the scan I edited in step 2 repeated three times.

cmadjar:
The same kind of changes than this PR https://github.com/aces/Loris/pull/3860/files should probably be applied on the resolved tab

cmadjar:
PR merged

PR #4805, opened 2019-06-06
[mri_violation ] Fix resolved violated scans duplication

cmadjar:
This pull request fixes a bug in the MRI violations module for which it displays the same violations more than once if a candidate has visits that are spread across multiple subprojects.

How to test this PR:
Pick one candidate that shows in the MRI violations module and that has multiple visits. If all these visits are all from the same subproject, manually edit one in the `session` table so that it has a different subprojectID than all other visits (could be NULL) and:
1) check that you can reproduce the bug on 20.0-release
2) check that the bug is fixed by this PR

See also: https://github.com/aces/Loris/issues/4781

(Note: a similar fix had been sent for the unresolved tab of the module, see https://github.com/aces/Loris/pull/3860)

### DIFF
--- a/modules/mri_violations/php/resolved_violations.class.inc
+++ b/modules/mri_violations/php/resolved_violations.class.inc
@@ -205,17 +205,19 @@ class Resolved_Violations extends \NDB_Menu_Filter_Form
                               time_run
                    )
                 ) as hash,
-                mri_protocol_violated_scans.ID as join_id,                
+                mpvs.ID as join_id,                
                 p.CenterID as Site,
                 violations_resolved.Resolved as Resolved
-            FROM mri_protocol_violated_scans 
+            FROM mri_protocol_violated_scans AS mpvs
             LEFT JOIN violations_resolved
-            ON (violations_resolved.ExtID=mri_protocol_violated_scans.ID 
+            ON (violations_resolved.ExtID=mpvs.ID 
             AND violations_resolved.TypeTable='mri_protocol_violated_scans')
             LEFT JOIN candidate c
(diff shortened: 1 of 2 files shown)


modules/mri_violations/php/resolved_violations.class.inc
```diff
@@ -205,17 +205,19 @@ class Resolved_Violations extends \NDB_Menu_Filter_Form
...
             LEFT JOIN violations_resolved
-            ON (violations_resolved.ExtID=mri_protocol_violated_scans.ID 
+            ON (violations_resolved.ExtID=mpvs.ID 
             AND violations_resolved.TypeTable='mri_protocol_violated_scans')
             LEFT JOIN candidate c
-            ON (mri_protocol_violated_scans.CandID = c.CandID)
+            ON (mpvs.CandID = c.CandID)
             LEFT JOIN session s
-            ON (mri_protocol_violated_scans.CandID = s.CandID)
+            ON (SUBSTRING_INDEX(mpvs.PatientName,'_',-1) = s.Visit_label
```
nicolasbrossard: I am not sure it's safe to assume that the patient name will always contain the visit label. If we change this convention some day this SQL statement might cease to work...Why not join with `tarchive` and then with `session` since both tables are linked via a session ID?

driusan: Is this still relevant, given your "approved" review?

nicolasbrossard: @driusan Not anymore

nicolasbrossard:
Tested on my sandbox with some violations that were resolved. I see duplicates without your fix and I don't see them if I have your fix. :+1:

PR #4822, opened 2019-06-07
[Issue Tracker] Fix watching dropdown when editing issue

### LINKED ISSUE #4582: [Issue Tracker] Unable to update "Watching?" field after creating issue
I tried setting "Watching?" from "Yes" to "No" and it still remains as "Yes"

driusan:
Fix changing the "Watching?" dropdown when editing an issue. There
were 2 problems:
1. The code was checking if a statement returned an array to determine
   if there was data. Comparing to null is more reliable.
2. The Watching? insert statement was done before Other watchers. This
   was unreliable, since Other Watchers starts by deleting all watches.

Fixing both of these issues resolves both the saving and display of
the Watching? dropdown when editing an issue.

Fixes #4582

### DIFF
--- a/modules/issue_tracker/ajax/EditIssue.php
+++ b/modules/issue_tracker/ajax/EditIssue.php
@@ -124,23 +124,6 @@ function editIssue()
         emailUser($issueID, $issueValues['assignee']);
     }
 
-    // Adding editor to the watching table unless they don't want to be added.
-    if ($_POST['watching'] == 'Yes') {
-        $nowWatching = array(
-                        'userID'  => $user->getData('UserID'),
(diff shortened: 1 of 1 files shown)


kongtiaowang:
modules/issue_tracker/ajax/EditIssue.php:146 PhanTypeArraySuspiciousNullable Suspicious array access to nullable ?array
modules/issue_tracker/ajax/EditIssue.php:152 PhanTypeArraySuspiciousNullable Suspicious array access to nullable ?array

driusan:
@cmadjar I rebased and after fixing some improperly resolved conflicts, I'm not able to reproduce.. it's switching fine for me.

PR #4827, opened 2019-06-10
[Media] Fix HTTP response codes when uploading files

driusan:
All types of errors were being returned to the user as
internal server error. This was clearly incorrect in
some situations where the error was caused by user error.

Update the showMediaError to take a status code to go
with it, and at the same time improved some error messages
(such as permission denied, which didn't include any
message at all before this change.)

### DIFF
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -36,7 +36,7 @@ function editFile()
     $db   =& Database::singleton();
     $user =& User::singleton();
     if (!$user->hasPermission('media_write')) {
-        header("HTTP/1.1 403 Forbidden");
+        showMediaError("Permission Denied", 403);
         exit;
     }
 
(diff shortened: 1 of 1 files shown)


modules/media/ajax/FileUpload.php
```diff
@@ -82,21 +82,23 @@ function uploadFile()
...
     }
 
     // Validate media path and destination folder
     $mediaPath = $config->getSetting('mediaPath');
 
     if (!isset($mediaPath)) {
-        showMediaError("Error! Media path is not set in Loris Settings!");
-        exit;
+        showMediaError(
+            "Media path not set in Loris settings! "
```
cmadjar: I know. I tiny request but it always bugs me when LORIS is not all CAP ;)

driusan: Fixed.. (I didn't write that sentence, I just changed the whitespace of the line and took out the "Error!")

cmadjar: Thanks!

Issue #4825, opened 2019-06-10
[tools] install script - directory missing


PapillonMcGill:
- try to change group ownership to apache (www-data) on 
     `modules/document_repository/user_uploads` and 
     `modules/data_release/user_uploads`   but none exist.    Both on Ubuntu and CentOS

cmadjar:
Closing this issue since the PR was merged

Issue #4824, opened 2019-06-10
[tools] install script for CentOS - multiple failures.


PapillonMcGill:
- Need to change text to install composer on CentOS as installing as instructed in the script is not working.
    `sudo php composer-setup.php --install-dir=/usr/local/bin --filename=composer`  will work.

- chgrp for ../project dir, sudo not working (not permitted as on Ubuntu) (lines 169,183)

- directory modules/document_repository/user_uploads and modules/data_release/user_uploads are non-existant (lines 156-166).  Need to be created.  chown will fail on CentOS (see previous)

- can't create apache config for Loris (write permission missing)

PapillonMcGill:
Sent a PR to solve message on installing composer https://github.com/aces/Loris/pull/4828
Waiting for future discussion on how to fix permissions

cmadjar:
@PapillonMcGill Since we discussed this during the meeting, would you mind updating the ticket and maybe creating a PR with the outcome of the discussion. If I am not mistaken, it was:
- do a sudo test at the beginning of the script and exit if user has no sudo permission
- document what needs to be done in order to be able to install LORIS on CentOS.

Thanks!

PapillonMcGill:
@cmadjar Checking for sudo permission would require to actually try to install (every instructions requiring privileges permission). 
@driusan is to discuss with Pam.
Documentation exist in the wiki. Suggest moving in the docs section after 21.0

cmadjar:
@driusan did you talk to Pam about this? Is there anything to be done for this ticket or can it be closed in the end?

johnsaigle:
@maltheism @PapillonMcGill Is this still a problem after the recent updates to the install script and CentOS documentation?

PapillonMcGill:
@johnsaigle Yes, the install script still try to use sudo on RedHat based distro. Even in 22.0-release. Although the READMECentOS7.md file describe the step that need to be done and that the user doing it (lorisadmin) should have `sudo` permission, by default the install.sh script will fail. It is even not mentionned which sudo permission the user need. 
Remember by default, no single user have sudo permission, and each of these permissions (mdkir, chown, install composer, ...) have to be granted one by one.

I do believe that for RedHat based distro, the install.sh script should be run by a user with `root` access and that the script adjust the files ownership and permission accordingly after creation.

I need to confirm but I suspect that starting with RedHat 8, it is possible to designate a user to have full sudo acces during (OS) install. But it is not an obligation and less secure IMHO.

[...]

PR #4828, opened 2019-06-10
[tools] fix message to install composer,

PapillonMcGill:
### Brief summary of changes
- change the message on how to install composer if not already done
- create user_uploads dir if are not existing before changing group ownership

### This resolves issue...
https://github.com/aces/Loris/issues/4825
https://github.com/aces/Loris/issues/4824 (first item only)

### To test this change...
for the message, run the script on a server where composer is not installed (or not accessible), check that new instruction are showne.

For second part run the script on a new installation of Loris (in a new directory i.e. /tmp or ~/) and check that the 2 directories are created and have group set to www-data (Ubuntu)

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -84,7 +84,9 @@ else
     echo ""
     echo "PHP Composer does not appear to be installed. Please install it before running this script."
     echo ""
-    echo "(e.g. curl -sS https://getcomposer.org/installer | php)"
+    echo "(e.g. wget https://getcomposer.org/installer"
+    echo "php installer --install-dir=/usr/local/bin --filename=composer)"
+    echo "while having root permission)";
     exit 2;
 fi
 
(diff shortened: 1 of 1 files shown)


tools/install.sh
```diff
@@ -84,7 +84,9 @@ else
     echo ""
     echo "PHP Composer does not appear to be installed. Please install it before running this script."
     echo ""
-    echo "(e.g. curl -sS https://getcomposer.org/installer | php)"
+    echo "(e.g. wget https://getcomposer.org/installer"
+    echo "php installer --install-dir=/usr/local/bin --filename=composer)"
```
driusan: Is there a reason this can't still use the pipe? You're only going to install it once, I don't see much reason to save thescript.

PapillonMcGill: not really, just have to be sure the pass the argument to install as installing with default value result to composer being installed in the current directory with another filename. Will work on that.

PR #4829, opened 2019-06-10
[Media] Ensure date of administration is not in the future

### LINKED ISSUE #4573: [media] date taken in the future
when uploading a media file, I can specify a date in the future and it will work without notice.
Same apply when editing a  previously uploaded file, it is possible to change the date taken to a future date

driusan:
Add check to verify that the date of administration is not
in the future, if provided.

Fixes #4573.

### DIFF
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -49,8 +49,11 @@ function editFile()
         showMediaError("Media ID $idMediaFile not found", 404);
     }
 
+    $dateTaken = $req['dateTaken'];
+    checkDateTaken($dateTaken);
+
     $updateValues = [
-                     'date_taken' => $req['dateTaken'],
+                     'date_taken' => $dateTaken,
                      'comments'   => $req['comments'],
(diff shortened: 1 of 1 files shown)


driusan:
Blocked by #4827

driusan:
@cmadjar done

driusan:
@cmadjar I think that was fixed in a different PR, I rebased it so it should be working now if you re-test.

modules/media/ajax/FileUpload.php
```diff
@@ -390,3 +396,27 @@ function getFilesList()
...
+{
+    if (!empty($dateTaken)) {
+        $date = date_create_from_format("Y-m-d", $dateTaken);
+        if ($date === false) {
+            showMediaError("Invalid date: $dateTaken", 400);
+        }
+
+        $now  = new DateTime();
+        $diff = date_diff($date, $now)->format("%a");
+        if ($diff > 0) {
```
jesscall: @driusan Shouldn't this be `if ($diff < 0)` ??
`date_diff` does `$now - $date` which should technically be positive if the `$date` is in the past...

PR #4830, opened 2019-06-10
[Media] Fix errors for users with no media_write permission

### LINKED ISSUE #4737: [Media] Edit Button Still Available with No Write Permission
When a user does not have write permission, the 'Edit' button still appears. You can even open the form and click submit. Then message that is received is 'Upload Error'. It should more clear to the user that they do not have the proper permission to edit. This should likely be done by removing the 'Edit' button entirely.

driusan:
This fixes 2 problems when a user does not have write permission:
1. The reason the update failed is not given to the user.
2. The user should not have seen the button in the first place.

Fixes #4737.

### DIFF
--- a/modules/media/jsx/editForm.js
+++ b/modules/media/jsx/editForm.js
@@ -209,7 +209,8 @@ class MediaEditForm extends Component {
         this.props.fetchData();
       },
       error: function(err) {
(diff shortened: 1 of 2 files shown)


driusan:
@cmadjar done

Issue #4770, opened 2019-06-05
[configuration] wrong error message


PapillonMcGill:
in the project section of configuration, when entering a specifying a value that is not numerical for target, either for a new project or modifying an existing one, the error message is:
` Failed to save, same name already exist!` and disappear fast,  in about 1 second.

in that case, text should be ` Failed to save, value not integer` and should also stay a few more seconds to let user time to read.

Also entering float value do not generate error and only the integer part is saved in the DB.

driusan:
Fixed by above PR

PR #4832, opened 2019-06-10
[Config] Fix error message in the project page of the configuration module

cmadjar:
### Brief summary of changes

This PR customizes the error message shown in the configuration module under the project section when someone tries to save a target recruitment that is not a number. Now proper error message is displayed to the user telling them that the recruitment target must be a number.

### This resolves issue...

- [x] Github? #4770

### To test this change...

- [ ] Try to save a project with a string instead of a number in the target recruitment box, a proper error message should appear

### DIFF
--- a/modules/configuration/js/project.js
+++ b/modules/configuration/js/project.js
@@ -18,9 +18,15 @@ $(document).ready(function() {
         }
         
         var errorClosure = function(i, form) {
-            return function() {
-                $(form.find(".saveStatus")).text("Failed to save, same name already exist!").css({ 'color': 'red'}).fadeIn(500).delay(1000).fadeOut(500);
+          if (isNaN(recruitmentTarget)) {
+            return function () {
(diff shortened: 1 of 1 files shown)


cmadjar:
@PapillonMcGill since you logged the issue for the configuration module, could you test/review this PR please? Thanks!!

modules/configuration/js/project.js
```diff
@@ -18,9 +18,15 @@ $(document).ready(function() {
         }
         
         var errorClosure = function(i, form) {
-            return function() {
-                $(form.find(".saveStatus")).text("Failed to save, same name already exist!").css({ 'color': 'red'}).fadeIn(500).delay(1000).fadeOut(500);
+          if (isNaN(recruitmentTarget)) {
```
driusan: Can you create an issue (not for 21.0) saying we should have an integer or numeric config type? I think it makes more sense than hardcoding special cases for some fields.

cmadjar: https://github.com/aces/Loris/issues/4833

PR #4729, opened 2019-05-29
[new_profile] Fix 6 digit years & month functionality

zaliqarosli:
fixes https://github.com/aces/Loris/issues/4728 and https://github.com/aces/Loris/issues/4564

to test:
- [ ] on test.loris.ca or 21.0-release branch, type in the year by keyboard into any date element. you will be able to enter a 6 digit year
- [ ] on this branch, you can only enter 4 digits on Chrome. The max attribute is broken on Firefox
- [ ] repeat the above with dob format in Configurations set to 'Ym' or 'YM'

caveat: Firefox allows 5/6 digit years because the "max" attribute breaks on that browser

PR #4088, opened 2018-11-13
[New Profile] Reactify

kongtiaowang:
### Brief summary of changes

Reactify the new profile module.
This PR fixes a PHP warning in the NDB_Menu_Filter.class.inc.

### To test this change...

test everything with the test plan.

PR #4834, opened 2019-06-10
[New_profile] Use NDB_Form validation and process

zaliqarosli:
### Brief summary of changes

This PR restructures the new_profile class to use the validation and processing function of NDB_Form, resulting in the following:

1. Fixes https://github.com/aces/Loris/issues/4567
2. Replaces https://github.com/aces/Loris/pull/4693
3. Fixes the last error reported by Shen in https://github.com/aces/Loris/pull/4729 by calling validate() in LorisForm via NDB_Form `$this->save()`
4. Restores all of back-end validation recently omitted in https://github.com/aces/Loris/pull/4088

#### JSX changes

- change error messaging to swals
- add minYear/maxYear where needed

#### PHP changes

- new_profile: change to NDB_Form so we can use the validate/process of that class
- profile: delete this class because everything here was moved to new_profile
- NDB_Form: add toArray and toJSON functionality so we can fetch data from ?format=json when request='GET'
- NDB_Page: move defining $fieldOptions two levels up to this class and add ?format=json functionality


modules/new_profile/php/new_profile.class.inc
```diff
@@ -44,19 +53,278 @@ class New_Profile extends \NDB_Page
...
+    function _process($values): void
+    {
+        // Change empty strings to null before passing as arguments.
+        foreach ($values as $key => $value) {
+            if (empty($value)) {
+                $values[$key] = null;
+            }
+        }
+        /* When a user has only one site, or when a centerID has not been
+         * passed, choose the first one in the User's centerID list.
```
driusan: I think this should only be true if the user has 1 site. If it has more, it should be an error to not pass one.

modules/new_profile/php/new_profile.class.inc
```diff
@@ -44,19 +53,278 @@ class New_Profile extends \NDB_Page
...
+     *
+     * @param ServerRequestInterface $request The incoming PSR7 request
+     *
+     * @return ResponseInterface The outgoing PSR7 response
+     */
+    public function handle(ServerRequestInterface $request) : ResponseInterface
+    {
+        $this->setup();
+        if ($request->getMethod() === "POST") {
+            $this->save();
```
driusan: Since you're overriding handle, did you test this both with and without valid permissions?

modules/new_profile/php/new_profile.class.inc
```diff
@@ -44,19 +53,278 @@ class New_Profile extends \NDB_Page
...
+        }
+
+        // Validate site entered
+        if (empty($values['site'])) {
+            $errors['site'] = 'Site is required.';
+        } else {
+            $psc = (int)$values['site'];
+            $user_list_of_sites = $user->getData('CenterIDs');
+            $num_sites          = count($user_list_of_sites);
+            if ($num_sites > 1 && !$user->hasCenter($psc)) {
```
zaliqarosli: > I think this should only be true if the user has 1 site. If it has more, it should be an error to not pass one.

@driusan yes, you're right. the validation is done here so i'll remove the case in `_process()`

modules/new_profile/php/new_profile.class.inc
zaliqarosli: yep, you can't access without data_entry permissions

zaliqarosli:
@driusan @kongtiaowang RB data missing 'useProjects' config setting. was this changed to something else or moved somewhere?

zaliqarosli:
@cmadjar thanks for testing! that error was reported in https://github.com/aces/Loris/pull/4729 which I will fix there once this PR is merged and i can rebase

cmadjar:
@zaliqarosli Perfect! So I'll approve this PR since the only issue I found is to be resolved in a different PR.

Issue #4733, opened 2019-05-29
[Issue Tracker] Page loading error when user has issue_tracker_reporter permissions


AlexandraLivadas:
When a user has issue_tracker reporter permission (adding issues, editing their issues, commenting on issues), the issue tracker page is available in the "Tools" drop-down menu but it does not load. 

Here is what appears: 
<img width="877" alt="Screen Shot 2019-05-29 at 2 07 07 PM" src="https://user-images.githubusercontent.com/35467361/58580487-3c5d8300-821b-11e9-8275-bb33a16c2e7c.png">

This is the console error: 
<img width="730" alt="Screen Shot 2019-05-29 at 2 07 29 PM" src="https://user-images.githubusercontent.com/35467361/58580499-42ebfa80-821b-11e9-98d0-3205303dd87c.png">

driusan:
I think this is a problem with the data on that VM. Are there issues with null CenterIDs?

kongtiaowang:
This is a testing data issue, will update the RB data on test.loris.ca

zaliqarosli:
also getting this issue, but 'Issue Tracker' module doesn't even appear in the 'Tools' dropdown menu for me

driusan:
@zaliqarosli if you can reproduce can you look into fixing this?

zaliqarosli:
@driusan if i can get around to it, yes. for now, i'll leave this unassigned if that's alright

zaliqarosli:
https://github.com/aces/Loris/pull/4835

PR #4835, opened 2019-06-10
[Issue tracker] Fix permissions

zaliqarosli:
### Brief summary of changes

fixes https://github.com/aces/Loris/issues/4733 and other back-end error that popped up. if you still get an error, make sure that all your entries in `issues` table have a valid 'centerID' that is not 'NULL'. if you still get an error in your error log after that, please report

### This resolves issue...
- [ ] Github? # https://github.com/aces/Loris/issues/4733


zaliqarosli:
@driusan @kongtiaowang travis is failing because the data set has NULLs as CenterID .. should i amend raisin bread with this PR?

modules/issue_tracker/jsx/IssueForm.js
```diff
@@ -338,7 +338,7 @@ class IssueForm extends Component {
         console.error(err);
         this.setState({submissionResult: 'error'});
         let msgType = 'error';
-        let message = err.responseJSON.message || 'Failed to submit issue :(';
+        let message = err.responseText.message || 'Failed to submit issue :(';
```
driusan: I don't think this is correct.. most of the ajax calls return a JSON  object with a "message" key. If there's any that don't, it would probably make more sense to update those

zaliqarosli:
travis fail: while permissions do work on my own VM, travis and raisinbread are having trouble accessing the module for some reason

modules/issue_tracker/test/Issue_TrackerTest.php
```diff
@@ -100,7 +100,7 @@ function testIssueTrackerDoespageLoad()
      */
```
kongtiaowang: line :63 
         $this->DB->insert(
             "issues",
             array(
              'issueID'  => '999999',
              'title'    => 'Test Issue',
              'status'   => 'new',
              'priority' => 'low',
              'reporter' => 'TestUser',
              'centerID' => '55',     // add centerID here
              )

Issue #4777, opened 2019-06-05
[Document repository] New subcategory doesn't appear


zaliqarosli:
Newly created subcategory does not appear in parent category directory on front-end as long as the parent category does not contain files

driusan:
PR merged

Issue #4780, opened 2019-06-05
[Document repository] Newly uploaded file does not appear until refresh


zaliqarosli:
Newly uploaded file does not appear until refresh

PR #4838, opened 2019-06-11
[document_repository] show sub-category in empty table

kongtiaowang:
to solve: https://github.com/aces/Loris/issues/4777

### DIFF
--- a/jsx/DataTable.js
+++ b/jsx/DataTable.js
@@ -331,7 +331,7 @@ class DataTable extends Component {
   }
 
   render() {
-    if (this.props.data === null || this.props.data.length === 0) {
+    if ((this.props.data === null || this.props.data.length === 0) && this.props.nullTableShow !== 'show' ) {
       return (
         <div>
           <div className="row">
--- a/jsx/DataTable.js
+++ b/jsx/DataTable.js
@@ -331,7 +331,7 @@ class DataTable extends Component {
   }
 
   render() {
-    if ((this.props.data === null || this.props.data.length === 0) && this.props.nullTableShow !== 'show' ) {
+    if ((this.props.data === null || this.props.data.length === 0) && !this.props.nullTableShow) {
       return (
         <div>
           <div className="row">
@@ -601,6 +601,7 @@ DataTable.propTypes = {
   onSort: PropTypes.func,
   actions: PropTypes.object,
   hide: PropTypes.object,
+  nullTableShow: PropTypes.bool,
 };
 DataTable.defaultProps = {
   headers: [],
@@ -612,6 +613,7 @@ DataTable.defaultProps = {
     downloadCSV: false,
     defaultColumn: false,
   },
+  nullTableShow: false,
 };
 
 export default DataTable;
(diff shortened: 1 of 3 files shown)


kongtiaowang:
@ZainVirani 
the folder list is in the table header, so "no data found" will not render the table.

jsx/DataTable.js
```diff
@@ -331,7 +331,7 @@ class DataTable extends Component {
   }
 
   render() {
-    if (this.props.data === null || this.props.data.length === 0) {
+    if ((this.props.data === null || this.props.data.length === 0) && this.props.nullTableShow !== 'show' ) {
```
coreyAbout: It seems that the `if` statement will only be executed when `this.props.data === null ` && this.props.data.length !== 0:
because: `this.props.data.length === 0` traced back to to line 198 `const treeTable = (this.state.tableData.length === 0)?`  in `modules/document_repository/jsx/docIndex.js`, at that same place, the `nullTableShow="show"` is defined. and it is passed down to here.
This means, if `this.props.data.length === 0`, then `this.props.nullTableShow === 'show'`

kongtiaowang: @coreyAbout 
Before If the data is empty(this.props.data === null or this.props.data.length === 0), this will render a null table. In this particular case. When data is empty, we don't want the render a null table, that is why I put a props variable here, and this props variable will not bother the other modules with react table. "this.props.nullTableShow !== 'show'" always is true, if you don't give a value to nullTableShow. [This means, if this.props.data.length === 0, then this.props.nullTableShow === 'show'] that is what I extractly want. It means don't render this part, jump out this if condition.

zaliqarosli: @kongtiaowang are you still going to add the change that we talked about on tuesday that if this.props.childrenNode is not empty then make this.props.nullTableShow = 'show'?

coreyAbout: @kongtiaowang 
Thanks for the explanation! I think we may leave a little comment in the code. I have no other problems. Cheers!

coreyAbout: A comment may be something like (you choose the right word): 
this.props.nullTableShow !== 'show' is for: do not render this if statement when there is null data for doc_repository, where nullTableShow is set to show and passed down here.

kongtiaowang: @zaliqarosli I tried, it is will cause the React Datatable failed. The Datatable only accept the data one time. If you change the table or anything in this table it will crash. That is why I used two tables (Datatable and nullDatatable) to render this module.

zaliqarosli:
@coreyAbout please rereview!

kongtiaowang:
@coreyAbout It should work, did u run the "sudo make dev"? you have to re-compile these react component firsts. Please let me know, it still not work for you.

zaliqarosli:
@coreyAbout also make sure to refresh because what you're seeing might be due to this issue https://github.com/aces/Loris/issues/4780 and changes to fix that issue haven't been pulled into this branch

driusan:
@coreyAbout any update on this?

coreyAbout:
@kongtiaowang @zaliqarosli @driusan 
I re-tested it after use "make dev", it is good!

Issue #4776, opened 2019-06-05
[Document repository] Upload formdata gets erased if error


zaliqarosli:
When error thrown on trying to upload duplicate file name, form data gets erased and it shouldn't

driusan:
Fixed by above

PR #4839, opened 2019-06-11
[Document repository] fix Upload formdata gets erased if error

kongtiaowang:
to solve: https://github.com/aces/Loris/issues/4776
to test: upload a duplicate file.


driusan:
I tested this, and it doesn't seem to fix the bug for me. The form data is still disappearing.

kongtiaowang:
@davidblader Could u test it again?

Issue #4851, opened 2019-06-12
[issue_tracker] Control Panel Navigation empty


zaliqarosli:
should have 'Back to List', 'Next', and 'Previous' to navigate through issues. currently nothing renders

zaliqarosli:
update: completely remove navigation control panel in Issue Tracker

driusan:
PR that completely removes the side panel merged

PR #4855, opened 2019-06-12
[issue_tracker] Remove navigation control panel

zaliqarosli:
fixes https://github.com/aces/Loris/issues/4851 and does some cleanup


driusan:
I tested this and

1. The next and previous buttons are reversed
2. Even though it was there before, I don't really see much point in the sidebar. When is someone going to want to just go through all issues sequentially?

zaliqarosli:
@driusan 
1. the buttons are not actually reversed .. it so happens that the issueIDs descend as you go through the data table
2. i can't answer this question as i have no experience using this module - im only fixing the bug. this may be a design question for projects.

zaliqarosli:
https://github.com/aces/Loris/pull/4890 fixes the order in which the data table is sorted

driusan:
Screen Shot 2562-06-17 at 16 39 52

This is how it loads by default.. when I click on "1" in the first row the url contains Issue ID "1" and the only option is "Previous".

zaliqarosli:
@driusan #4890 fixes this

driusan:
Needs to be retested with latest changes

driusan:
Tested it.. it's not quite as pretty with the form being full page width, but it doesn't look really bad either, so I'll call it passed.

Issue #4866, opened 2019-06-13
[imaging uploader] Data table incorrect in the browse tab


cmadjar:
The data table displayed in the browse page of the imaging browser is incorrect:
- There is a PatientName column with no data attached to it (should probably be hidden)
- The "View Details" link for the DICOM archive is not a link and the tarchive ID is displayed in that column instead of the View details link that should be there

PR #4869, opened 2019-06-13
[Imaging Uploader] Fix browse data table

cmadjar:
### Brief summary of changes

The data table of the browser page of the imaging uploader showed several flaws:
- a PatientName name column was present in the header of the table but not the content (was not properly hidden)
- The Tarchive Info column showed the tarchiveID instead of the "View details" link to the DICOM archive module for that tarchive ID
- The Number of MINCs created column leads to the ViewSession page of the imaging browser since a given upload will always be linked to only one SessionID. 
- The header row has prettier strings for the eyes

### This resolves issue...

- [x] Github #4866

### To test this change...

Check that the improvements mentioned above works better in this PR than on the current 21.0 branch.

### DIFF
--- a/modules/imaging_uploader/jsx/ImagingUploader.js
+++ b/modules/imaging_uploader/jsx/ImagingUploader.js
@@ -10,7 +10,7 @@ import UploadForm from './UploadForm';
 class ImagingUploader extends Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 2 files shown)


nicolasbrossard:
Tested on my VM and everything looks sharp! Tagging as Passed Manual Test. :+1:

Issue #4846, opened 2019-06-11
[timepoint_list] "Imaging Scan Done 'yes'" link not lead to one specific participant


coreyAbout:
TestPlan Item `4. Datatable content:
Imaging Scan Done: If 'yes', ensure links point to correct place (imaging_browser for that candidate and visit_label)`

Problem:
If yes, links point to the datasets for all participants, instead of that one participant.
e.g., MTL007 V3

Issue #4845, opened 2019-06-11
[timepoint_list] "View Imaging datasets" button not lead to the right participant


coreyAbout:
TestPlan Item `3. Button links: 
Ensure the "View Imaging datasets" button points to correct place. (imaging_browser module for that candidate)`

Problem:
Click on it points to datasets for all participant, instead of for that candidate only.

driusan:
Fixed by above

PR #4871, opened 2019-06-14
[timepoint list] Fix links to the imaging browser

cmadjar:
### Brief summary of changes

This fixes the broken links to the imaging browser from the timepoint list module (filters were not being set or link to the wrong page)

### This resolves issue...

- [x] Github #4846 and #4845


modules/timepoint_list/templates/menu_timepoint_list.tpl
```diff
@@ -60,7 +60,7 @@
     <!-- </div> -->
     <!-- <div class="col-xs-4"> -->
         {$actions}
-           <button class="btn btn-primary timepoint_imaging_datasets" data-pscid="{$PSCID}">
+           <button class="btn btn-primary timepoint_imaging_datasets" onclick="self.location.href=('{$baseurl}/imaging_browser/?DCCID={$candID}')">
```
driusan: If this is a link, why can't it be an `<a>` tag (perhaps styled as a button)? "onclick" only works with a mouse or something that can click. `<a href>` works with screen readers, tablets, keyboards, etc.

cmadjar: @driusan done :)

PR #4873, opened 2019-06-14
[Document repository] fix Newly uploaded file does not appear until refresh

kongtiaowang:
to solve:https://github.com/aces/Loris/issues/4780

### DIFF
--- a/modules/document_repository/jsx/uploadForm.js
+++ b/modules/document_repository/jsx/uploadForm.js
@@ -169,7 +169,6 @@ class DocUploadForm extends React.Component {
       console.log(data);
       if (data == 'uploaded successfully') {
       swal('Upload Successful!', '', 'success');
-      this.fetchData();
       this.props.refreshPage();
       } else {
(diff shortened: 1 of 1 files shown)


modules/document_repository/jsx/uploadForm.js
```diff
@@ -171,6 +170,7 @@ class DocUploadForm extends React.Component {
       if (data == 'uploaded successfully') {
       swal('Upload Successful!', '', 'success');
       this.fetchData();
+      this.props.refreshPage();
```
driusan: 1. What's the point of fetching the data and then immediately refreshing the page? Fetching the data to discard it seems unnecessary in that case.
2. What happens to the "success" swal right above? How long do users have to see that message?

kongtiaowang: 1. You are right, I will remove fetchData.
2. Click the OK button, then close this message. 
Screen Shot 2019-06-18 at 3 58 33 PM

PR #4877, opened 2019-06-14
[instrument_list] Fixed raisinbread null register by field

kchatpar:
### Brief summary of changes
The column RegisteredBy in the session table in raisinbread had null entries which was not allowing time points to be sent to DCC. The Null entries are now 'admin' and the button send To DCC should work.

### This resolves issue...

- [ ] Github? #4584

### To test this change...

- Check out my branch and verify that the RegisteredBy Column in the session table doesn't have any null values.
- Verify that send to DCC works after all Data Entry, Double Data Entry, and conflicts are resolved.


kongtiaowang:
@kchatpar 
modules/dashboard/test/DashboardTest.php line 443, change "4" to "10". This new data makes Loris have 10 new scans, not 4 anymore.

PR #4888, opened 2019-06-17
Do not allow whitespace characters in user names.

nicolasbrossard:
### Brief summary of changes

Spaces are not allowed anymore in user names. This PR implements this new restriction.

### This resolves issue...

- [ ] Github? #4742

### To test this change...

- [ ] Try creating a new user with a username that contains a whitespace character and ensure you cannot.
- [ ] Create a new user and set the user name to be the user's email address. Enter an email address that contains a whitespace (e.g an email with an embedded comment that contains a space, as in crusty(the clown)@simsons.com) and try clicking 'Save'. Ensure that you get an error message.

This replaces #4744, which is now closed.

### DIFF
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -978,11 +978,11 @@ class Edit_User extends \NDB_Form
 
                 // Check that user name does not contain a whitespace character
                 if (preg_match('/\s/', $effectiveUID)) {
-                    // Note: email adresses can contain comments which themselves can
-                    // contain spaces
+                    // Note: email addresses can contain comments which themselves
+                    // can contain spaces
                     if ($values['NA_UserID'] == 'on') {
                         $errors['UserID_Group']
-                            = 'You cannot have the user name match an email adress'
+                            = 'You cannot have the user name match an email address'
                             . ' that contains a whitespace character';
                     } else {
                         $errors['UserID_Group']
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -973,7 +973,21 @@ class Edit_User extends \NDB_Form
 
                 if (strlen($effectiveUID) > 255) {
                     $errors['UserID_Group']
-                        = 'The user name must not exsceed 255 characters';
+                        = 'The user name must not exceed 255 characters';
+                }
+
+                // Check that user name does not contain a whitespace character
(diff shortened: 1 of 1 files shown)


modules/user_accounts/php/edit_user.class.inc
```diff
@@ -973,7 +973,21 @@ class Edit_User extends \NDB_Form
...
                 if (strlen($effectiveUID) > 255) {
                     $errors['UserID_Group']
-                        = 'The user name must not exsceed 255 characters';
+                        = 'The user name must not exceed 255 characters';
+                }
+
+                // Check that user name does not contain a whitespace character
+                if (preg_match('/\s/', $effectiveUID)) {
+                    // Note: email adresses can contain comments which themselves can
+                    // contain spaces
```
driusan: That's true for mail clients software, but is it true for LORIS?

The name part of `"John Smith" <john.smith@example.com>` is constructed, not user supplied. (Assuming that's what you're referring to..)

nicolasbrossard: You can enter an email address like `me(yeah really me)@test.com` on the create new user page in LORIS, and that is a valid email address with a user-supplied comment. My PR prevents the user name from being set to the email address in such a (rare/extreme) case.

driusan: TIL. (That's an insane thing for the spec to allow.)

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -973,7 +973,21 @@ class Edit_User extends \NDB_Form
...
+                    if ($values['NA_UserID'] == 'on') {
+                        $errors['UserID_Group']
+                            = 'You cannot have the user name match an email adress'
```
johnsaigle:
```suggestion
                            = 'You cannot have the user name match an email address'
```

Small typo.

Also I'm a little bit confused by this part of the code. What is this checking for?

nicolasbrossard: @johnsaigle I am trying to display an error message that is coherent with the way that the user name is set (i.e whether it was typed in the user name text field or it the 'Make user name match email address' checkbox was used).

driusan:
@nicolasbrossard this is failing Travis

nicolasbrossard:
@driusan It fails for PHP 7.2 but passes for PHP 7.3. Travis "bug"?

driusan:
you're right, looks transient.. I just restarted it

Issue #4778, opened 2019-06-05
[Document repository] Comments do not appear for categories


zaliqarosli:
they used to appear as tooltips but don't now

PR #4889, opened 2019-06-17
[document_repository] add tooltip 

kongtiaowang:
to solve : #4778
When you hover the category list, it will show the tooltip


modules/document_repository/jsx/childTree.js
```diff
@@ -42,17 +70,21 @@ class ChildTree extends Component {
...
+            <div
+              style={headerRow}
               key={index}
               className="pointer"
-              style={spanStyle}
             >
-              <i className="fa fa-folder"></i>
-              <span style={spanStyle}>
+              <i className="fa fa-folder" style={iconStyle}></i>
+              <div style={spanStyle} id="tooltip">
```
driusan: `id`s need to be unique across the entire page. I think this component is used multiple times on the same page, so a hardcoded "tooltip" won't work.

nicolasbrossard: This PR does not work on my sandbox (no tooltip shown...)

kongtiaowang:
Screen Shot 2019-06-18 at 3 25 25 PM

kongtiaowang:
@nicolasbrossard Could u test it again? it only shows when the comment is not empty.

johnsaigle:
Comments for categories do not appear for me.

Issue #4887, opened 2019-06-17
[Candidate List] Open Profile Gets Stuck Validating


suzanne-lee:
Open profile works fine when the DCCID and PSCID match, but sometimes gets stuck validating when inputs are invalid 

<img width="730" alt="Screenshot 2019-06-17 14 15 15" src="https://user-images.githubusercontent.com/29237731/59626834-e63d7a80-910a-11e9-8b9b-b1b687e1ca64.png">
<img width="722" alt="Screenshot 2019-06-17 14 14 55" src="https://user-images.githubusercontent.com/29237731/59626836-e63d7a80-910a-11e9-925a-af8fc4333c4d.png">

driusan:
Is there any response in the network tab when it gets stuck?

suzanne-lee:
@driusan 500 error

ridz1208:
I ran into this, but its not consistent....

suzanne-lee:
PR sent #4893

driusan:
Fixed by above PR

PR #4893, opened 2019-06-18
[candidate_list] Check If DCCID is an Integer String

suzanne-lee:
### Brief summary of changes
Open Profile gets stuck validating when you try to input a DCCID that contains alphabet characters. This PR adds a validation to check if DCCID is an integer.

### This resolves issue...

- [ ] Github? #4887

### To test this change...

- [ ] Uncheck the "Across all sites access candidate profiles" permission to be able to view the Open Profile button
- [ ] Ensure invalid IDs and invalid ID combinations result in an alert/error message


sruthymathew123:
success_1

PR #4894, opened 2019-06-18
[Couch importers] Issue in escaping

ridz1208:
### Brief summary of changes

the escaping was happening mid field name. the resulting would look like this

```
cc`1`.Status
```

instead of 

```
`cc1`.Status
```

### DIFF
--- a/tools/CouchDB_Import_Instruments.php
+++ b/tools/CouchDB_Import_Instruments.php
@@ -116,7 +116,7 @@ class CouchDBInstrumentImporter
 
         // add the SQL table to the query
         $extraSelect = ", i.* ";
-        $extraJoin = "JOIN ".$this->SQLDB->escape($tablename)." i ON (i.CommentID=f.CommentID) ";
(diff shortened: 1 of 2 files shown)


tools/CouchDB_Import_Instruments.php
```diff
@@ -116,7 +116,7 @@ function generateDocumentSQL(string $tablename) : string
 
         // add the SQL table to the query
         $extraSelect = ", i.* ";
-        $extraJoin = "JOIN $tablename i ON (i.CommentID=f.CommentID) ";
+        $extraJoin = "JOIN ".$this->SQLDB->escape($tablename)." i ON (i.CommentID=f.CommentID) ";
```
driusan:
```suggestion
        $extraJoin = "JOIN " . $this->SQLDB->escape($tablename) . " i ON (i.CommentID=f.CommentID) ";
```

Maybe it's just the way GitHub is highlighting it, but this looks weird (it looks like the dots are inside the string surrounding the table name, even though they're just the PHP concatenation operator outside of quotes.)

ridz1208:
@driusan incorporated the changes

Issue #4857, opened 2019-06-12
[Document Repository] Files are not clearly separated from categories; visual improvements should be made


driusan:
I uploaded a test document into a category named "test2". After creating a category named "Another Test", the document moved into the "Another Test" category in the frontend without having modified it.

driusan:
@zaliqarosli can you look into this one?

zaliqarosli:
@driusan i can't reproduce this error! can you give me more information if you have any?

driusan:
I just started with a completely empty document repository and follow the above steps. This is what the frontend looks like:

![Screen Shot 2562-06-19 at 14 05 36](https://user-images.githubusercontent.com/498329/59789157-62ff5e80-929b-11e9-9ca5-1d84ff2f5dfd.png)

While this is what document_repository_categories and document_repository tables look like:

```
mysql> select * from document_repository_categories\G
*************************** 1. row ***************************
           id: 1
category_name: test2
    parent_id: 0
     comments: NULL
*************************** 2. row ***************************
           id: 2
category_name: Test
    parent_id: 1
     comments: xxxxxxx
*************************** 3. row ***************************
           id: 3
category_name: Another Test
    parent_id: 0
     comments: This is a test with a very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very long description
3 rows in set (0.00 sec)

mysql> select * from document_repository\G
*************************** 1. row ***************************
    record_id: 2
        PSCID: NULL
   Instrument: NULL
   visitLabel: NULL
   Date_taken: NULL
Date_uploaded: 2019-06-12 15:31:46
     Data_dir: SEA0002_V24_audiology.jpg
    File_name: SEA0002_V24_audiology.jpg
    File_type: jpg
      version: NULL
    File_size: 100015
[...]

PR #4896, opened 2019-06-19
[document_repository] category and site can edit

kongtiaowang:
to solve: use has edit permission can edit the file category and site.


driusan:
I don't know if this fixes any other issues, but the document from #4857 is still showing up in the wrong category for me.

kongtiaowang:
it will add in major.

PR #4897, opened 2019-06-19
[DQT] Fix crash when CouchDB values not set in project/config.xml

johnsaigle:
### Brief summary of changes
If your `project/config.xml` file does not contain entries for `CouchDB` values, LORIS crashes with a blank page. 

This PR prints a simple error message and logs a more informative error instead of crashing.

### To test this change...

- [ ] Modify your `project/config.xml` so that it does not contain all of the settings necessary for a CouchDB connection, i.e. delete any one of `port`, `hostname`, `admin`, `adminpass` or `dbName`
- [ ] Navigate to the DQT. The page should be blank and return a 500 error. 
- [ ] On this branch, an error message will appear instead. Details will be shown in the error log.

### DIFF
--- a/modules/dataquery/php/dataquery.class.inc
+++ b/modules/dataquery/php/dataquery.class.inc
@@ -59,17 +59,16 @@ class Dataquery extends \NDB_Form
             $couchConfig['adminpass']
         );
         if (!$couchConfigValid) {
-            throw new \ConfigException(
-                'Missing CouchDB configuration settings. Cannot load ' .
-                'Data Query Tool.'
-            );
             error_log(
                 'Data Query Tool cannot be loaded due to missing ' .
                 'configuration settings. Ensure that `dbName`, `hostname`, ' .
                 '`port`, `admin` and `adminpass` values are properly ' .
                 'configured in project/config.xml'
             );
-            return;
+            throw new \ConfigException(
+                'Missing CouchDB configuration settings. Cannot load ' .
+                'Data Query Tool.'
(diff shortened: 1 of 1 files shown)


modules/dataquery/php/dataquery.class.inc
```diff
@@ -51,7 +51,25 @@ class Dataquery extends \NDB_Form
...
+        $couchConfigValid = isset(
+            $couchConfig['dbName'],
+            $couchConfig['hostname'],
+            $couchConfig['port'],
+            $couchConfig['admin'],
+            $couchConfig['adminpass']
+        );
+        if (!$couchConfigValid) {
+            echo 'Missing CouchDB configuration settings. Cannot load ' .
+                'Data Query Tool.';
```
driusan: There is no reason to echo inside of the setup of a class, this will just break the returned result from handle().

johnsaigle: OK, I changed it to just throw a ConfigException. This still breaks the page but it now at least displays 500 instead of a blank page.

The module is odd -- it doesn't use Smarty exactly but also doesn't seem to be properly Reactified/using PSR stuff. I'm not exactly sure how to display an error on the front-end in the right way.

driusan: Now that you've changed it to throw an exception, lines below the throw in the if statement are unreachable.

johnsaigle:
@driusan Could you re-review?

PR #4904, opened 2019-06-19
[Travis] Fix phan errors on the release branch

driusan:
Fix phan errors that somehow crept into the release branch.


driusan:
#4903 beat me to it.

Issue #4905, opened 2019-06-20
[DQT] runQuery shows to info tab content after the query is run.


xlecours:
When results should appear in the `View Data` tab, the content of the `Info` tab is present

driusan:
PR merged

PR #4910, opened 2019-06-20
[DQT] Fix info tab showing instead of query results

xlecours:
### Brief summary of changes
Using the Active property of TabPane to "select" the Active tab

Bonus: fix Warning: Each child in an array or iterator should have a unique "key" prop in the render method of `StatsVisualizationTabPane`

### This resolves issue...

- [ ] Github? https://github.com/aces/Loris/issues/4905

### DIFF
--- a/modules/dataquery/jsx/react.tabs.js
+++ b/modules/dataquery/jsx/react.tabs.js
@@ -682,7 +682,7 @@ class StatsVisualizationTabPane extends Component {
         </table>
       );
 
-      let content = (
+      content = (
         <div>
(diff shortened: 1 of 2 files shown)


modules/dataquery/jsx/react.tabs.js
```diff
@@ -679,7 +682,7 @@ class StatsVisualizationTabPane extends Component {
         </table>
       );
 
-      let content = (
```
xlecours: This was causing the Stats tab to display content from the parent block (line 634)

PR #4913, opened 2019-06-26
[Release] SQL archiving and release patch + composer update

ridz1208:
### Brief summary of changes
Preparation for release 21

concatenation bash command:
```
for n in SQL/New_patches/*.sql; do echo $n; echo -e "\nSELECT 'Running: $n';\n" >> SQL/Release_patches/20.3_To_21.0_upgrade.sql; cat $n >> SQL/Release_patches/20.3_To_21.0_upgrade.sql; done;

```


SQL/Release_patches/20.3_To_21.0_upgrade.sql
```diff
@@ -0,0 +1,325 @@
...
+    'boolean',
+    ID,
+    'Enable "Pwned Password" check',
+    22
+  FROM
+    ConfigSettings
+  WHERE
+    Name="study";
+
+INSERT INTO Config (ConfigID, Value) SELECT ID, 'true' FROM ConfigSettings WHERE Name="usePwnedPasswordsAPI";
```
xlecours: This should be set to false. I don't think that all instances should send HTTP request to a server in Netherlands at each password changes.

xlecours:
Travis is mad; `modules/issue_tracker/test/Issue_TrackerTest.php::testClearFormIssueTracker` is all commented

SQL/Release_patches/20.3_To_21.0_upgrade.sql
ridz1208: hmm... I guess it was approved by someone.

I track that and see if it was previously discussed

driusan: It was discussed.. @xlecours is right, it's probably just that the schema was reviewed/changed and the patch was missed.

driusan: I was mistaken, the discussion was about whether to include a config setting, not what the default should be.

PR #4433, opened 2019-03-28
[Test] PHPUNIT 6.5 to PHPUNIT 8 | php-ast 0.1.6 to php-ast 1.0.1|

kongtiaowang:
Update PHPUNIT Version
Get rid of dbunit. This package is abandoned and no longer maintained.
The @expectedException, @expectedExceptionCode, @expectedExceptionMessage, and @expectedExceptionMessageRegExp annotations are deprecated. 
They will be removed in PHPUnit 9.  

To test: Travis

Issue #4284, opened 2019-01-28
Add return types to Database class "pselect*" functions


johnsaigle:
#3878 did not add return types to `pselect*` functions because they can return array, null, and sometimes false (I believe). Wrapper functions in other classes and modules that call `pselect` functions are unable to have return types for the same reason.

It would be good to standardize and document the types that these functions can return and eventually to add appropriate return types.

driusan:
They shouldn't return false, they should all be either `?array` or `?string`, I think.

johnsaigle:
Agreed I think some of them have strange error handling though.

This issue is somewhat addressed by https://github.com/aces/Loris/pull/6039 because it makes the array formats more clear. When that is merged we can take a look at making the functions more consistent.

johnsaigle:
I think for the most part they return `?array` now. Only `pselectOne()` needs a return type. Adding it causes phan errors for downstream functions that use it though, so it will require a slightly larger cleanup PR.

laemtl:
It seems to have been addressed.

PR #4914, opened 2019-06-27
Add new phan rule PhanMismatchReturn

johnsaigle:
### Brief summary of changes

This PR adds the new phan rule PhanMismatchReturn which flags code that is declared to return a type but in reality returns a different type. e.g.

```php
/**
* @return int
*/
function returnInt() {
    return 'hello';
}
```

Turns out we do this all over the place in LORIS so this PR is pretty big.

A lot of phan errors came up related to https://github.com/aces/Loris/pull/4308 so this PR also fixes a lot of those.

There are also minor optimizations/rewrites of functions that were needlessly complicated. In pretty much every case the replacements should be logically equivalent and hopefully more maintainable.

### DIFF
--- a/htdocs/survey.php
+++ b/htdocs/survey.php
@@ -369,7 +369,7 @@ class DirectDataEntryMainPage
             $this->tpl_data['complete']  = true;
 
             $this->updateStatus('Complete');
-            $Responses      = $DB->update(
+            $DB->update(
                 $this->TestName,
                 array(
                  'Date_taken' => date('Y-m-d'),
@@ -378,7 +378,7 @@ class DirectDataEntryMainPage
                  'CommentID' => $this->CommentID,
                 )
             );
-            $Responses_flag = $DB->update(
+            $DB->update(
                 'flag',
                 array(
                  'Data_entry'     => 'Complete',
--- a/modules/create_timepoint/php/create_timepoint.class.inc
+++ b/modules/create_timepoint/php/create_timepoint.class.inc
@@ -208,7 +208,9 @@ class Create_Timepoint extends \NDB_Form
                         WHERE CenterID =:cid",
                         array('cid' => $siteID)
                     );
-                    $psc_labelOptions[$siteID] = $center['Name'];
+                    if (!is_null($center['Name'])) {
+                        $psc_labelOptions[$siteID] = $center['Name'];
+                    }
                 }
             }
             $this->addSelect('psc', 'Site', $psc_labelOptions);
--- a/modules/create_timepoint/php/create_timepoint.class.inc
+++ b/modules/create_timepoint/php/create_timepoint.class.inc
@@ -208,7 +208,7 @@ class Create_Timepoint extends \NDB_Form
                         WHERE CenterID =:cid",
                         array('cid' => $siteID)
                     );
-                    if (!is_null($center['Name'])) {
+                    if (!is_null($center)) {
                         $psc_labelOptions[$siteID] = $center['Name'];
                     }
                 }
--- a/modules/data_release/ajax/FileUpload.php
+++ b/modules/data_release/ajax/FileUpload.php
@@ -43,7 +43,7 @@ if ($_POST['action'] == 'upload'
     } else {
         $target_path = $base_path . $fileName;
         if (move_uploaded_file($_FILES["file"]["tmp_name"], $target_path)) {
-            $success = $DB->insert(
+            $DB->insert(
                 'data_release',
                 array(
                  'file_name'   => $fileName,
@@ -68,7 +68,7 @@ if ($_POST['action'] == 'upload'
                  'upload_date' => $upload_date,
                 )
             );
-            $success = $DB->insert(
+            $DB->insert(
                 'data_release_permissions',
                 array(
                  'userid'          => $user_ID,
--- a/modules/server_processes_manager/php/serverprocesslauncher.class.inc
+++ b/modules/server_processes_manager/php/serverprocesslauncher.class.inc
@@ -70,7 +70,7 @@ class ServerProcessLauncher
      *
      * @param AbstractServerProcess $process the process that was just started
      *
-     * @return void
+     * @return string
      * @throws \InvalidArgumentException if the process is null or if either the
      *                                  PID or the process start time is null.
      */
@@ -128,7 +128,7 @@ class ServerProcessLauncher
      *
      * @param AbstractServerProcess $process process to launch.
      *
-     * @return int ID of the process saved in the database.
+     * @return ?int ID of the process saved in the database.
      *
      * @throws \LorisException if the process cannot be launched successfully.
      */
@@ -152,7 +152,7 @@ class ServerProcessLauncher
             );
         }
 
-        return $this->_saveProcess($process);
+        return intval($this->_saveProcess($process));
     }
 
     /**
--- a/modules/server_processes_manager/php/serverprocesslauncher.class.inc
+++ b/modules/server_processes_manager/php/serverprocesslauncher.class.inc
@@ -161,7 +161,7 @@ class ServerProcessLauncher
      * @param int    $mriUploadId    ID of the MRI upload in the mri_upload table.
      * @param string $sourceLocation location of the MRI file
      *
-     * @return int ID (in the database) of the launched process or null
+     * @return ?int ID (in the database) of the launched process or null
      *             if the process could not be run
      */
     public function mriUpload($mriUploadId, $sourceLocation)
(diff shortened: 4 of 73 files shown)


htdocs/survey.php
```diff
@@ -372,7 +372,7 @@ function display()
             $this->tpl_data['complete']  = true;
 
             $this->updateStatus('Complete');
-            $Responses      = $DB->update(
+            $DB->update(
```
johnsaigle: unused, same with below.

htdocs/survey.php
```diff
@@ -372,7 +372,7 @@ function display()
             $this->tpl_data['complete']  = true;
 
             $this->updateStatus('Complete');
-            $Responses      = $DB->update(
```
johnsaigle: unused. same with below.

modules/candidate_list/php/candidate_list.class.inc
```diff
@@ -152,8 +152,6 @@ class Candidate_List extends \NDB_Menu_Filter
                                'participantstatus' => $participant_status_options,
                                'useedc'            => $config->getSetting("useEDC"),
                               ];
-
-        return true;
```
johnsaigle: When you see this line deleted in this PR in other files, it typically means that the function should actually return void.

modules/create_timepoint/php/create_timepoint.class.inc
```diff
@@ -208,7 +208,9 @@ class Create_Timepoint extends \NDB_Form
                         WHERE CenterID =:cid",
                         array('cid' => $siteID)
                     );
-                    $psc_labelOptions[$siteID] = $center['Name'];
+                    if (!is_null($center)) {
```
johnsaigle: Another theme in this PR is verifying that DB results are not `null`. Since [the `pselect*` family does not have any return types](https://github.com/aces/Loris/issues/4284) it is possible for them to return `null` instead of the expected `array` of data. This causes phan to get upset when we try to access an array key without first ensuring that the DB result is not null.

modules/data_release/ajax/FileUpload.php
```diff
@@ -43,7 +43,7 @@
     } else {
         $target_path = $base_path . $fileName;
         if (move_uploaded_file($_FILES["file"]["tmp_name"], $target_path)) {
-            $success = $DB->insert(
```
johnsaigle: insert/update always return `void` so they should not be assigned to a value.

driusan:
This should go to major, not the 21.0-release branch at this point.

johnsaigle:
Has 21 been merged back into major? I'll mark this as Needs Rebase for now but I think changing the base branch right now will make all the changes meaningless.

johnsaigle:
```
modules/statistics/php/stats_behavioural.class.inc:162 PhanTypeMismatchDimAssignment When appending to a value of type 'All Projects'|mixed|null|string, found an array access index of type mixed, but expected the index to be of type int
modules/statistics/php/stats_behavioural.class.inc:228 PhanTypeMismatchDimAssignment When appending to a value of type 'All Projects'|mixed|null|string, found an array access index of type mixed, but expected the index to be of type int
php/libraries/NDB_Factory.class.inc:112 PhanTypeMismatchReturn Returning type null but config() is declared to return \NDB_Config
php/libraries/NDB_Factory.class.inc:147 PhanTypeMismatchReturn Returning type null but user() is declared to return \User
php/libraries/NDB_Factory.class.inc:204 PhanTypeMismatchReturn Returning type null but database() is declared to return \Database
src/Middleware/MiddlewareChainerMixin.php:46 PhanTypeMismatchDeclaredReturn Doc-block of withMiddleware contains declared return type \LORIS\Middleware\MiddlewareChainerMixin which is incompatible with the return type \LORIS\Middleware\MiddlewareChainer declared in the signature
src/Middleware/MiddlewareChainerMixin.php:62 PhanTypeMismatchReturn Returning type \LORIS\Middleware\MiddlewareChainerMixin but withMiddleware() is declared to return \LORIS\Middleware\MiddlewareChainer
test/unittests/NDB_BVL_FeedbackTest.php:73 PhanTypeMismatchReturn Returning type \PHPUnit_Extensions_Database_DataSet_CompositeDataSet but getDataSet() is declared to return \PHPUnit\DbUnit\DataSet\IDataSet
```

These are the remaining `phan` errors after this rule is applied. I'm not exactly sure how to solve these ones so any help would be appreciated.

modules/server_processes_manager/php/abstractserverprocess.class.inc
```diff
@@ -47,79 +47,79 @@ abstract class AbstractServerProcess
     /**
      * ID of the process when recorded in the database.
      *
-     * @var int.
+     * @var int
```
johnsaigle: Newer versions of phan don't like the `.` after the type

johnsaigle:
The only remaining phan issue (i.e. what is causing Travis to fail) is caused by a reference to DbUnit which is removed by #4433. 

Until it's merged, this PR will be marked as Blocked.

johnsaigle:
```
php/libraries/NDB_Factory.class.inc:112 PhanUndeclaredFunction Call to undeclared function \MockNDB_Config()
php/libraries/NDB_Factory.class.inc:147 PhanUndeclaredFunction Call to undeclared function \MockUser()
```

@driusan These are my remaining phan errors. Phan doesn't seem to understand what we're doing were with PHPUnit.

I tried using the PHPUnit extensions for phan -- no change.

I also tried using Mock::generate(CLASS) but that doesn't work with our tests.

Any thoughts?

driusan:
Are you sure those line numbers are right? In NDB_Factory for those lines I see:

line 112: `return self::$config`
line 147: `*/`

The "MockUser" and "MockNDB_Config" references near those lines both have the `new` keyword before them, so I would expect phan to be smart enough to at least realize they aren't function calls.

johnsaigle:
Yes, sorry I reverted wrong. My issue is this:

> php/libraries/NDB_Factory.class.inc:122 PhanTypeMismatchReturn Returning type \MockNDB_Config but config() is declared to return \NDB_Config

If I add `\MockNDB_Config` to the `@return` it gives an error about this class not existing, e.g.

> php/libraries/NDB_Factory.class.inc:107 PhanTypeMismatchDeclaredReturn Doc-block of config contains declared return type \MockNDB_Config which is incompatible with the return type \NDB_Config declared in the signature
php/libraries/NDB_Factory.class.inc:109 PhanUndeclaredTypeReturnType Return type of config() is undeclared type \MockNDB_Config

driusan:
I don't think we'll ever be able to fix that since from what I can tell PHPUnit creates the mock classes using eval. https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/MockObject/MockClass.php#L42 .. the mock objects are dynamically generated code, they're not very amenable to static analysis.

Since PHP type hinting doesn't support sum types while doc comments do, our options are either:
1. Tell phan to ignore those 2 lines and add a comment explaining why, since we know that it's an instance of NDB_Config even if phan doesn't.
2. Remove the type hinting from the signature but leave it in the docblock, so that phan can still analyze the types but doesn't flag the mismatch as an error.

Short of refactoring the code, I think this is one of the rare cases where I'd prefer 1, but I have nothing against 2 either.

johnsaigle:
I'd prefer approach 1) too. I'll add the suppression and a note explaining it

johnsaigle:
I'm getting similar issues with:

>php/libraries/NDB_Factory.class.inc:112 PhanTypeMismatchReturn Returning type null but config() is declared to return \NDB_Config
3926php/libraries/NDB_Factory.class.inc:157 PhanTypeMismatchReturn Returning type null but user() is declared to return \User

I'm going to also add a suppression for these.

modules/server_processes_manager/php/serverprocesslauncher.class.inc
```diff
@@ -70,7 +70,7 @@ class ServerProcessLauncher
      *
      * @param AbstractServerProcess $process the process that was just started
      *
-     * @return void
+     * @return string
```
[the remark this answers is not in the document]
johnsaigle: Agreed, however this type change should probably be done in the Database class. The function `lastInsertID` should be changed to return an int instead of a string.

src/Api/Endpoints/Project/Images.php
```diff
@@ -99,7 +99,7 @@ public function handle(ServerRequestInterface $request) : ResponseInterface
      *
      * @param ServerRequestInterface $request The incoming PSR7 request
      *
-     * @return array
+     * @return array|\Psr\Http\Message\ResponseInterface
```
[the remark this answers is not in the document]
johnsaigle: Phan wants this because of the Exception case that returns a 400 response.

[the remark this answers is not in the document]
driusan: Okay.. it's still a code smell, but it's fine for this PR since it's just documenting what's already there

PR #4915, opened 2019-07-02
[candidate_parameters] Add 'Date of Birth' tab

jesscall:
### Brief summary of changes
This PR adds a new tab to candidate parameters which can be used to amend a candidate's date of birth. 

This change is being applied as a result of the DoD meeting which took place on June 25, 2019. To follow is a PR which will add `Date of Death`.

**_[NOTE]:_** I am basing this PR on `21.0-release` branch for now, but will change it to `minor` after release happens.

### To test this change...

- [ ] Navigate to candidate parameters. Make sure that you can only update date of birth if you have the permission `candidate_dob_edit`. Make sure any update to the date actually saves.

### DIFF
--- a/SQL/New_patches/2019_07_02_Add_Edit_DoB_Permissions.sql
+++ b/SQL/New_patches/2019_07_02_Add_Edit_DoB_Permissions.sql
@@ -1,0 +1,5 @@
+INSERT INTO permissions (code, description, categoryID) VALUES
+    ("candidate_dob_edit","Edit dates of birth",2);
+    
+INSERT INTO user_perm_rel VALUES
+    (1, (SELECT permID FROM permissions WHERE code='candidate_dob_edit'));
--- a/modules/candidate_parameters/ajax/formHandler.php
+++ b/modules/candidate_parameters/ajax/formHandler.php
@@ -55,6 +55,10 @@ case 'consentStatus':
     editConsentStatusFields($db, $user);
     break;
 
+case 'candidateDOB':
+    editCandidateDOB($db, $user);
+    break;
+
 default:
     header("HTTP/1.1 404 Not Found");
     exit;
@@ -523,5 +527,34 @@ function editConsentStatusFields($db, $user)
             }
             $db->insert('candidate_consent_history', $updateHistory);
         }
+    }
+}
+
+/**
+ * Handles the updating of candidate's date of birth.
+ *
+ * @param Database $db   database object
+ * @param User     $user user object
+ *
+ * @throws DatabaseException
+ *
+ * @return void
+ */
+function editCandidateDOB(\Database $db, \User $user): void
+{
+    $candID       = $_POST['candID'];
+    $dob          = $_POST['dob'];
+    $strippedDate = null;
+    if (!empty($dob)) {
+        $config    = \NDB_Config::singleton();
+        $dobFormat = $config->getSetting('dobFormat');
+        if ($dobFormat === 'YM') {
+            $strippedDate = date("Y-m", strtotime($dob))."-01";
+        }
+        $db->update(
+            'candidate',
+            array('DoB' => $strippedDate ?? $dob),
+            array('CandID' => $candID)
+        );
     }
 }
--- a/modules/candidate_parameters/ajax/formHandler.php
+++ b/modules/candidate_parameters/ajax/formHandler.php
@@ -542,7 +542,7 @@ function editConsentStatusFields($db, $user)
  */
 function editCandidateDOB(\Database $db, \User $user): void
 {
-    $candID       = $_POST['candID'];
+    $candID       = new CandID($_POST['candID']);
     $dob          = $_POST['dob'];
     $strippedDate = null;
     if (!empty($dob)) {
@@ -554,7 +554,7 @@ function editCandidateDOB(\Database $db, \User $user): void
         $db->update(
             'candidate',
             array('DoB' => $strippedDate ?? $dob),
-            array('CandID' => $candID)
+            array('CandID' => $candID->__toString())
         );
     }
 }
--- a/modules/candidate_parameters/ajax/getData.php
+++ b/modules/candidate_parameters/ajax/getData.php
@@ -500,7 +500,6 @@ function getConsentStatusHistory($pscid)
 /**
  * Handles the fetching of candidate's date of birth.
  *
- *
  * @return array
  */
 function getDOBFields(): array
@@ -515,9 +514,9 @@ function getDOBFields(): array
     $pscid         = $candidateData['PSCID'] ?? null;
     $dob           = $candidateData['DoB'] ?? null;
     $result        = [
-                      'pscid'  => $pscid,
-                      'candID' => $candID->__toString(),
-                      'dob'    => $dob,
-                     ];
+        'pscid'  => $pscid,
+        'candID' => $candID->__toString(),
+        'dob'    => $dob,
+    ];
     return $result;
 }
--- a/modules/candidate_parameters/ajax/getData.php
+++ b/modules/candidate_parameters/ajax/getData.php
@@ -500,7 +500,6 @@ function getConsentStatusHistory($pscid)
 /**
  * Handles the fetching of candidate's date of birth.
  *
- * @throws DatabaseException
  *
  * @return array
  */
--- a/modules/candidate_parameters/ajax/getData.php
+++ b/modules/candidate_parameters/ajax/getData.php
@@ -48,6 +48,9 @@ case 'participantStatus':
 case 'consentStatus':
     echo json_encode(getConsentStatusFields());
     exit;
+case 'candidateDOB':
+    echo json_encode(getDOBFields());
+    exit;
 default:
     header("HTTP/1.1 404 Not Found");
     exit;
@@ -492,4 +495,30 @@ function getConsentStatusHistory($pscid)
           $formattedHistory[$key] = $consentHistory;
     }
     return $formattedHistory;
+}
+
+/**
+ * Handles the fetching of candidate's date of birth.
+ *
+ * @throws DatabaseException
+ *
+ * @return array
+ */
+function getDOBFields(): array
+{
+    $candID = new CandID($_GET['candID']);
+    $db     = \Database::singleton();
+    // Get PSCID
+    $candidateData = $db->pselectRow(
+        'SELECT PSCID,DoB FROM candidate where CandID =:candid',
+        array('candid' => $candID->__toString())
+    );
+    $pscid         = $candidateData['PSCID'] ?? null;
+    $dob           = $candidateData['DoB'] ?? null;
+    $result        = [
+                      'pscid'  => $pscid,
+                      'candID' => $candID->__toString(),
+                      'dob'    => $dob,
+                     ];
+    return $result;
 }
--- a/modules/candidate_parameters/jsx/CandidateDOB.js
+++ b/modules/candidate_parameters/jsx/CandidateDOB.js
@@ -1,0 +1,157 @@
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+import Loader from 'Loader';
+import swal from 'sweetalert2';
+
+class CandidateDOB extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      data: {},
+      formData: {
+        dob: null,
+      },
+      error: false,
+      isLoaded: false,
+    };
+
+    this.fetchData = this.fetchData.bind(this);
+    this.setFormData = this.setFormData.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+  }
+
+  componentDidMount() {
+    this.fetchData()
+    .then(() => this.setState({isLoaded: true}));
+  }
+
+  fetchData() {
+    return fetch(this.props.dataURL, {credentials: 'same-origin'})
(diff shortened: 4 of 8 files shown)


modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -527,3 +529,31 @@ function editConsentStatusFields($db, $user)
...
+ * @param Database $db   database object
+ * @param User     $user user object
+ *
+ * @throws DatabaseException
+ *
+ * @return void
+ */
+function editCandidateDOB($db, $user)
+{
+    if (!$user->hasPermission('candidate_parameter_edit')
```
```suggestion
    if (!$user->hasAnyPermission(array('candidate_parameter_edit', 'candidate_dob_edit')))
```

If you want to, you can use this function to check multiple permissions.

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -527,3 +529,31 @@ function editConsentStatusFields($db, $user)
...
+ * Handles the updating of candidate's date of birth.
+ *
+ * @param Database $db   database object
+ * @param User     $user user object
+ *
+ * @throws DatabaseException
+ *
+ * @return void
+ */
+function editCandidateDOB($db, $user)
```
johnsaigle: Do you need to pass these as params? I think you could just use the singletons instead.

If you do need them, it would be good to add types:
```suggestion
function editCandidateDOB(\Database $db, \User $user): void
```

modules/candidate_parameters/ajax/getData.php
```diff
@@ -470,3 +472,29 @@ function getConsentStatusHistory($pscid)
...
 }
+
+/**
+ * Handles the fetching of candidate's date of birth.
+ *
+ * @throws DatabaseException
+ *
+ * @return array
+ */
+function getDOBFields()
```
```suggestion
function getDOBFields(): array
```

modules/candidate_parameters/ajax/getData.php
```diff
@@ -470,3 +472,29 @@ function getConsentStatusHistory($pscid)
...
+{
+    $candID = $_GET['candID'];
+    $db     = \Database::singleton();
+    // Get PSCID
+    $candidateData = $db->pselectRow(
+        'SELECT * FROM candidate where CandID =:candid',
```
```suggestion
        'SELECT PSCID,DoB FROM candidate where CandID =:candid',
```

Might as well only ask for what you need.

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -527,3 +529,29 @@ function editConsentStatusFields($db, $user)
...
+ * @param Database $db   database object
+ * @param User     $user user object
+ *
+ * @throws DatabaseException
+ *
+ * @return void
+ */
+function editCandidateDOB(\Database $db, \User $user): void
+{
+    if (!$user->hasAnyPermission('candidate_parameter_edit', 'candidate_dob_edit')) {
```
```suggestion
    if (!$user->hasAnyPermission(array('candidate_parameter_edit', 'candidate_dob_edit'))) {
```
This function actually takes an array, not strings, as an argument. This should fix your Travis/phan error.

jesscall: Okay thanks!

SQL/New_patches/2019_07_02_Add_Edit_DoB_Permissions.sql
```diff
@@ -0,0 +1,5 @@
+INSERT IGNORE INTO permissions (code, description, categoryID) VALUES
```
coreyAbout: why use "ignore" here?

jesscall: will amend this.

modules/candidate_parameters/jsx/CandidateDOB.js
```diff
@@ -0,0 +1,223 @@
...
+        alertMessage = 'Update Successful!';
+      } else if (this.state.updateResult === 'error') {
+        let errorMessage = this.state.errorMessage;
+        alertClass = 'alert alert-danger text-center';
+        alertMessage = errorMessage ? errorMessage : 'Failed to update!';
+      }
+    }
+
+     return (
+      <div className='row'>
```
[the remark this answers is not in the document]
jesscall: @coreyAbout not sure what you mean by this. On my end, the form looks the same as other two tabs. 
Could you provide a screenshot?

jesscall:
Hey @coreyAbout! Thanks for the feedback.
To address your points:

1. Option B has been the standard way of changing DOB. Projects put in a request and someone on our team would run the `fix_timepoint_date_problems.php` script or run a patch. Since we will soon be supporting date of death as well, it was decided to add a tab for date of birth in cand params so that 1. the burden could be taken off of the developers and 2. to remove the DOB feature from the timepoints date script since it is already a convoluted script and it is questionable to have both non-timepoint dates and timepoint dates in the same script.
In either case A or B, the `fix _candidate_age` script needs to be run after making the change. The script updates the calculated date in all instruments, etc. I can not attest how other projects have done this, but in CCNA the script is run automatically on daily basis through a cronjob.

2. The permission should appear in user accounts after running the patch in this PR. Currently the patch is only attributing the permission to the admin account and so if you are logged into your personal account you wouldn't be able to see permissions that you don't already have. 

3. You are right about this. The code currently checks whether the user has either `candidate_dob_edit` or `candidate_parameter_edit` permission. Perhaps it is not necessary to have both. 
However, I could see value in changing the logic here to having both of those permissions required. Example is a project who doesn't want data entry staff to be able to change DOBs but wants them to be able to enter Consent Status.

modules/candidate_parameters/jsx/CandidateDOB.js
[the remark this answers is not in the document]
coreyAbout: I was trying to reproduce it today, but when I run `make dev`, a bunch of syntax errors appears, such as `
    ERROR in ./modules/candidate_parameters/jsx/CandidateDOB.js
    Module Error (from ./node_modules/eslint-loader/index.js):
    
    /var/www/loris/modules/candidate_parameters/jsx/CandidateDOB.js
        1:1   error    Unexpected var, use let or const instead        no-var
       10:22  error    Missing trailing comma                          comma-dangle
       14:7   error    Unexpected var, use let or const instead        no-var
       20:13  error    Unexpected var, use let or const instead        no-var
       30:33  error    Missing trailing comma                          comma-dangle`

Also, this branch is not based on the newest 21.0 release, 20.3-21.0 sql patch was not under Release_patches?

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -527,3 +529,31 @@ function editConsentStatusFields($db, $user)
...
+    if (!$user->hasPermission('candidate_parameter_edit')
+        && !$user->hasPermission('candidate_dob_edit')
+    ) {
+        header("HTTP/1.1 403 Forbidden");
+        exit;
+    }
+    $candID       = $_POST['candID'];
+    $dob          = $_POST['dob'];
+    $strippedDate = date("Y-m", strtotime($dob))."-01";
+    $db->update(
```
PapillonMcGill: if $dob is not a valid date, the DB update will fail

modules/candidate_parameters/jsx/CandidateDOB.js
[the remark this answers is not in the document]
coreyAbout: First tab & second tab
Screenshot from 2019-07-22 16:42:19
Screenshot from 2019-07-22 16:42:19

Third tab birth date:
Screenshot from 2019-07-22 16:42:52

The third tab spread over all the way to the right, and aligned to center, which are different with the other two.

SQL/New_patches/2019_07_02_Add_Edit_DoB_Permissions.sql
```diff
@@ -0,0 +1,5 @@
+INSERT INTO permissions (code, description, categoryID) VALUES
+    ("candidate_dob_edit","Edit dates of birth",2);
+
+INSERT INTO user_perm_rel VALUES
+    ((SELECT ID from users WHERE UserID='admin'),(SELECT permID FROM permissions WHERE code='candidate_dob_edit'));
```
coreyAbout: userID = 'admin' not correct, in my case, I used 'lorisadmin' as userID. In other people's case they may used a different userID. Consider use something not varies across people.

ridz1208: @driusan correct me if I'm wrong but `admin` exists in the default install and is always referred to when creating new permissions no ?

should we rely on `ID=1` instead ? that doesnt seem like a better idea

johnsaigle: `admin` does exist in the default install afaik. 

In any case I don't think we need this line. The `admin` equivalent will probably have the superuser permission. Other admin-level users can probably be granted this permission on case-by-case basis.

driusan: admin exists in the default install but may have been renamed or deleted on various installs, or installs that happened before that default was created.

But doesn't admin have highlander anyways?

coreyAbout: Installing the Database - 2 of 2
[2. Fill in \<loris-admin-username\>, \<loris-admin-passwrd\>; e.g. admin](https://github.com/aces/Loris/wiki/Hosting-the-Database-Myself)
The installing instruction used "e.g. admin", people may randomly choose their user names. Not sure whether that is something to consider.

sruthymathew123:
I think @jesscall has to address the issues reported by Melanie and Corey.

modules/candidate_parameters/jsx/CandidateDOB.js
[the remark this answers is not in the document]
jesscall: @coreyAbout Should be fixed now!

SQL/New_patches/2019_07_02_Add_Edit_DoB_Permissions.sql
jesscall: alright, removed-

modules/candidate_parameters/jsx/CandidateDOB.js
```diff
@@ -0,0 +1,235 @@
...
+          />
+          <StaticElement
+            label='DCCID'
+            text={this.state.Data.candID}
+          />
+          <StaticElement
+            label='Disclaimer:'
+            text='Any changes to the date of birth requires an
+            administrator to run the fix_candidate_age script.'
+          />
```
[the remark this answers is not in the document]
jesscall: This is what it looks like now. Let me know if it's okay

Screen Shot 2019-10-05 at 4 09 50 PM

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -517,3 +521,32 @@ function editConsentStatusFields($db, $user)
...
+ * @param Database $db   database object
+ * @param User     $user user object
+ *
+ * @throws DatabaseException
+ *
+ * @return void
+ */
+function editCandidateDOB(\Database $db, \User $user): void
+{
+    $candID       = $_POST['candID'];
```
```suggestion
    $candID       = new CandID($_POST['candID']);
```

johnsaigle:
One small change, after that I'll approve

jesscall:
Whoever reviews this PR, can you double check the permissions ? (candidate_dob_edit)

modules/candidate_parameters/ajax/getData.php
```diff
@@ -493,3 +496,29 @@ function getConsentStatusHistory($pscid)
     }
     return $formattedHistory;
 }
+
+/**
+ * Handles the fetching of candidate's date of birth.
+ *
+ * @throws DatabaseException
```
```suggestion
```

PapillonMcGill:
@jesscall Although it pass the manual test, I found some annoying points:
- there is a disclaimers that an admin is required to run the "fix_candidate_age" script. There is no such script and the disclaimer is distracting for nothing.
- I tried to only change the month (or year) directly in the date fields and it changed other values to default (i.e.:  yyyy-02-dd). Using the graphic calendar is Ok. Firefox 68.4.1esr on Linux

ridz1208:
@PapillonMcGill 
for your first point https://github.com/aces/Loris/blob/master/tools/data_integrity/fix_candidate_age.php

for the second, @jesscall could you investigate ?

PapillonMcGill:
would be good to complete the disclaimer saying that the fix_candidate_age.php script would need to be run if some instruments as already been entered for that candidate.

jesscall:
@PapillonMcGill I don't experience the same issue with changing parts of the date. I am using Chrome. Perhaps this is only reproducible on Firefox

jesscall:
@PapillonMcGill 
You are right, this is the case for all date fields on Firefox. This should be fixed in a separate PR. 
Please create a github ticket for it.

PapillonMcGill:
@jesscall You want me to approved this even if the bug is not fix?

ridz1208:
@PapillonMcGill if the issue is not related to this PR specifically and the PR does what it should please approve/pass manual tests

PapillonMcGill:
@ridz1208 just wanted confirmation on next step.

modules/candidate_parameters/jsx/CandidateDOB.js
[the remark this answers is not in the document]
driusan: I'm not going to request the changes now, but after @nicolasbrossard's presentation on the server processes manager today it occurs to me that this would be a good place to use the SPM to launch the fix_candidate_age script asynchronously.

PR #4917, opened 2019-07-03
Add new phan rule PhanUndeclaredVariableDim to static analysis

johnsaigle:
Adding this rule makes it so that developers must explicitly declare arrays they are using before referencing keys used in those arrays.

e.g.

```php
function foo() {
     $arr['key'] = 'value';
     return $arr;
}
```

must now be

```php
function foo() {
     $arr = array('key' => 'value');
     // or: $arr = array(); $arr['key'] = 'value';
     return $arr;
}
```

This PR also fixes existing code where this occurs (and does some light cleanup).

### DIFF
--- a/modules/configuration/php/configuration.class.inc
+++ b/modules/configuration/php/configuration.class.inc
@@ -2,7 +2,7 @@
 /**
  * The configuration module is used to manage the configuration of Loris.
  *
- * PHP version 5
+ * PHP version 7
  *
  * @category Behavioural
  * @package  Main
@@ -14,7 +14,7 @@
 namespace LORIS\configuration;
 
 /**
- * Admin_Config Class
+ * Configuration Class
  *
  * This class is to configure the system settings
  *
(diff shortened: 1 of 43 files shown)


modules/configuration/php/configuration.class.inc
```diff
@@ -47,22 +47,22 @@ class Configuration extends \NDB_Form
...
             $val = $type['Scan_type'];
-            $scan_types[$val] =$val;
+            $scan_types[$val] = $val;
         }
-        $scan_types = array_merge(array('' => ''), $scan_types);
 
-        $instruments = \Utility::getAllInstruments();
-        $instruments = array_merge(array('' => ''), $instruments);
+        $instruments     = \Utility::getAllInstruments();
+        $instruments[''] = '';
```
PapillonMcGill: ??

modules/create_timepoint/php/create_timepoint.class.inc
```diff
@@ -178,8 +178,8 @@ class Create_Timepoint extends \NDB_Form
                         "longer supported in Loris."
                     );
                 }
-                $labelOptions[''] = null;
-                $items            = \Utility::associativeToNumericArray(
+                $labelOptions = array('' => null);
```
PapillonMcGill: same here??

modules/configuration/php/configuration.class.inc
johnsaigle: If you look at the previous line 65, that's what it was doing, creating an array element with both the key and value set to an empty string.

I rewrote it just to avoid the function call and make it more clear.

PapillonMcGill: Yes, I do understand that but do you know why declare such an empty key-value pair?

johnsaigle: I don't know why but I didn't want to alter the functionality.

PapillonMcGill: Sound good.
Should check why tpl_data['instruments'] need some empty entry. Possibly some poor underlying design.

PR #4918, opened 2019-07-03
[Data Framework] Update candidate list to use data framework

driusan:
This updates the candidate_list module to use the LORIS data framework. It does so by adding a new abstract DataFrameworkMenu class, which can be extended in order to remove some of the boilerplate of converting menu filters to the data framework.

In order to use the DataFrameworkMenu class, an implementation must supply:

1. a getFieldOptions function which returns a list of fieldOptions to include in the array for the frontend
2. a getBaseDataProvisioner function which returns a data provisioner
3.  a useSiteFilter function which returns whether or not the data provisioner should have the UserSiteFilter match applied.

In return, the class takes care of using that dataprovisioner, adding default LORIS data filters, and serializing to JSON in the format expected by react menu filters.

### DIFF
--- a/modules/candidate_list/php/candidatelistrow.class.inc
+++ b/modules/candidate_list/php/candidatelistrow.class.inc
@@ -1,0 +1,66 @@
+<?php
+/**
+ * This class implements a data Instance which represents a single
+ * row in the candidate list menu table.
+ *
+ * PHP Version 7
+ *
+ * @category   Behavioural
+ * @package    Main
+ * @subpackage Imaging
+ * @author     Dave MacFarlane <david.macfarlane2@mcgill.ca>
+ * @license    http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link       https://www.github.com/aces/Loris/
+ */
+
+namespace LORIS\candidate_list;
+
+/**
+ * A CandidateListRow represents a row in the Candidate List menu table.
+ *
+ * @category   Behavioural
+ * @package    Main
+ * @subpackage Imaging
+ * @author     Dave MacFarlane <david.macfarlane2@mcgill.ca>
+ * @license    http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link       https://www.github.com/aces/Loris/
+ */
+class CandidateListRow implements \LORIS\Data\DataInstance
+{
+    protected $DBRow;
+    protected $CenterID;
+
+    /**
+     * Create a new CandidateListRow
+     *
+     * @param array   $row The row (in the same format as \Database::pselectRow
+     *                     returns
+     * @param integer $cid The centerID affiliated with this row.
+     */
+    public function __construct(array $row, $cid)
+    {
+        $this->DBRow    = $row;
+        $this->CenterID = $cid;
+    }
+
+    /**
+     * Implements \LORIS\Data\DataInstance interface for this row.
+     *
+     * @return string the row data.
+     */
+    public function toJSON() : string
+    {
+        return json_encode($this->DBRow);
+    }
+
+    /**
+     * Returns the CenterID for this row, for filters such as
+     * \LORIS\Data\Filters\UserSiteMatch to match again.
+     *
+     * @return integer The CenterID
+     */
+    public function getCenterID()
+    {
+        return $this->CenterID;
+    }
+}
--- a/modules/candidate_list/php/candidatelistrow.class.inc
+++ b/modules/candidate_list/php/candidatelistrow.class.inc
@@ -59,7 +59,7 @@ class CandidateListRow implements \LORIS\Data\DataInstance
      *
      * @return integer The CenterID
      */
-    public function getCenterID()
+    public function getCenterID(): int
     {
         return $this->CenterID;
     }
--- a/modules/candidate_list/php/candidatelistrow.class.inc
+++ b/modules/candidate_list/php/candidatelistrow.class.inc
@@ -37,7 +37,7 @@ class CandidateListRow implements \LORIS\Data\DataInstance
      *                     returns
      * @param integer $cid The centerID affiliated with this row.
      */
-    public function __construct(array $row, $cid)
+    public function __construct(array $row, int $cid)
     {
         $this->DBRow    = $row;
         $this->CenterID = $cid;
--- a/modules/candidate_list/php/candidatelistrow.class.inc
+++ b/modules/candidate_list/php/candidatelistrow.class.inc
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
 /**
  * This class implements a data Instance which represents a single
  * row in the candidate list menu table.
--- a/modules/candidate_list/php/candidatelistrowprovisioner.class.inc
+++ b/modules/candidate_list/php/candidatelistrowprovisioner.class.inc
@@ -86,7 +86,7 @@ clASs CandidateListRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisio
      *
      * @return \LORIS\Data\DataInstance An instance representing this row.
      */
-    public function getInstance($row) : \LORIS\Data\DataInstance
+    public function getInstance(array $row) : \LORIS\Data\DataInstance
     {
         // XXX This should probably be updated to allow users
         // to access candidates who have any session at one of the
--- a/modules/candidate_list/php/candidatelistrowprovisioner.class.inc
+++ b/modules/candidate_list/php/candidatelistrowprovisioner.class.inc
@@ -93,7 +93,7 @@ clASs CandidateListRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisio
         // user's sites, but for now this maintains the previous
         // behaviour of requiring the registration site to match
         // one of the user's sites.
-        $cid = $row['RegistrationCenterID'];
+        $cid = (int )$row['RegistrationCenterID'];
         unset($row['RegistrationCenterID']);
         return new CandidateListRow($row, $cid);
     }
--- a/modules/candidate_list/php/candidatelistrowprovisioner.class.inc
+++ b/modules/candidate_list/php/candidatelistrowprovisioner.class.inc
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
 /**
  * This file implements a data provisioner to get all possible rows
  * for the candidate list menu page.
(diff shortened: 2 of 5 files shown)


johnsaigle:
```
FILE: ...modules/candidate_list/php/candidatelistrowprovisioner.class.inc
 34 | ERROR | [x] Whitespace found at end of line
 39 | ERROR | [x] Equals sign not aligned with surrounding
    |       |     assignments; expected 3 spaces but found 1 space
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY
FILE: ...d/aces/Loris/modules/candidate_list/php/candidate_list.class.inc
FOUND 9 ERRORS AFFECTING 9 LINES
  23 | ERROR | [x] Opening brace of a class must be on the line after
     |       |     the definition
  24 | ERROR | [ ] Missing function doc comment
  57 | ERROR | [x] Opening brace should be on a new line
 108 | ERROR | [x] Array key not aligned correctly; expected 16
     |       |     spaces but found 12
 109 | ERROR | [x] Array key not aligned correctly; expected 16
     |       |     spaces but found 12
 110 | ERROR | [x] Array key not aligned correctly; expected 16
     |       |     spaces but found 12
 111 | ERROR | [x] Array key not aligned correctly; expected 16
     |       |     spaces but found 12
 112 | ERROR | [x] Array key not aligned correctly; expected 16
     |       |     spaces but found 12
 113 | ERROR | [x] Closing parenthesis not aligned correctly;
     |       |     expected 15 space(s) but found 8
PHPCBF CAN FIX THE 8 MARKED SNIFF VIOLATIONS AUTOMATICALLY
```

johnsaigle:
There seems to be a phan error in the File Upload class that's making Travis fail 🤷‍♂

driusan:
@johnsaigle I still can't figure out why this PR in particular would cause that phan error, but I added code to "fix" it (throw an exception in the invalid case that it claims to be detecting.)

php/libraries/DataFrameworkMenu.class.inc
```diff
@@ -0,0 +1,135 @@
...
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris-Trunk/
+ */
+
+/**
+ * A DataFrameworkMenu is a type of menu (filterable data table)
+ * where the data comes from the LORIS data framework, and the
+ * formatting comes from React.
+ *
+ * "Menu" type data tables usually have one column which is clickable
```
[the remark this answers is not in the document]
driusan: I don't know if it's relevant here, but it's a fact that isn't explicitly documented anywhere else to the best of my knowledge and gives context to why "Menu" is in the name.

php/libraries/File_Upload.class.inc
```diff
@@ -349,7 +349,17 @@ class File_Upload
             return false;
         }
         //Test each method
-        foreach ($this->fileHandlers[$field]['methods'] AS $method) {
+        $methods = $this->fileHAndlers[$field]['methods'];
+        if (is_iterable($methods)) {
+            // Phan thinks methods is a string and fails below without
```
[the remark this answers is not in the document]
driusan: I don't think that would work, because it's not a class variable being iterated over, it's an item inside of the class variable that's being iterated over.

php/libraries/DataFrameworkMenu.class.inc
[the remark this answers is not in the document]
xlecours: In take case I would suggest: `"Menu" are html pages where data relevant to a specific domain is presented bulk, in a table, and usually accompanied with filters. This class toJson function returns an object containing the "Data" from the dataframwork and the fieldoptions required to build the filters dropdown. It is sometime used as a landing page for LORIS's modules.`
or something like that.

modules/candidate_list/php/candidatelistrowprovisioner.class.inc
```diff
@@ -0,0 +1,100 @@
...
+    public function getInstance($row) : \LORIS\Data\DataInstance
+    {
+        // XXX This should probably be updated to allow users
+        // to access candidates who have any session at one of the
+        // user's sites, but for now this maintains the previous
+        // behaviour of requiring the registration site to match
+        // one of the user's sites.
+        $cid = $row['RegistrationCenterID'];
+        unset($row['RegistrationCenterID']);
+        return new CandidateLIstRow($row, $cid);
```
xlecours: CandidateLIstRow => CandidateL<b>i</b>stRow

modules/candidate_list/php/candidatelistrow.class.inc
```diff
@@ -0,0 +1,66 @@
...
+        return json_encode($this->DBRow);
+    }
+
+    /**
+     * Returns the CenterID for this row, for filters such as
+     * \LORIS\Data\Filters\UserSiteMatch to match again.
+     *
+     * @return integer The CenterID
+     */
+    public function getCenterID()
```
johnsaigle:
```suggestion
    public function getCenterID(): int
```

modules/candidate_list/php/candidatelistrow.class.inc
```diff
@@ -0,0 +1,66 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

modules/candidate_list/php/candidatelistrowprovisioner.class.inc
```diff
@@ -0,0 +1,100 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

php/libraries/DataFrameworkMenu.class.inc
```diff
@@ -0,0 +1,135 @@
+<?php declare(strict_types=1);
+
+/**
+ * This class implements a DataframeWorkMenu, a specific type of
```
johnsaigle:
```suggestion
 * This class implements a DataFrameworkMenu, a specific type of
```

php/libraries/DataFrameworkMenu.class.inc
```diff
@@ -0,0 +1,135 @@
...
+ * menu filter where the front end is generated from React and the
+ * data comes from a \LORIS\Data\Provisioner.
+ *
+ * PHP Version 5
```
johnsaigle:
```suggestion
 * PHP Version 7
```

modules/candidate_list/php/candidatelistrowprovisioner.class.inc
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * Returns an instance of a CandidateListRow object for a given
+     * table row.
+     *
+     * @param array $row The database row from the LORIS Database class.
+     *
+     * @return \LORIS\Data\DataInstance An instance representing this row.
+     */
+    public function getInstance($row) : \LORIS\Data\DataInstance
```
johnsaigle:
```suggestion
    public function getInstance(array $row) : \LORIS\Data\DataInstance
```

driusan: I had to revert this, because it causes a signature incompatibility with the base class and violates the Liskov substitution principle (which also gets flagged as a phan error)

php/libraries/DataFrameworkMenu.class.inc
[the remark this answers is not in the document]
driusan: I'm not sure I see much of a difference, but I've reworded it be more similar to your suggestion. Is that better?

johnsaigle:
```
1) CandidateListTestIntegrationTest::testDataEntryAndOpenProfile
System info: host: '691e950d2d1a', ip: '172.18.0.2', os.name: 'Linux', os.arch: 'amd64', os.version: '4.4.0-101-generic', java.version: '1.8.0_91'
/app/modules/candidate_list/test/candidate_listTest.php:250
--
```

driusan:
@johnsaigle the tests were fine, the page wasn't loading because of the type signature/strict typing changes. It's passing now that those are resolved.

driusan:
Adding Blocked so that I don't accidentally merge because @xlecour's comments aren't showing up as "Request Changes" and I can't Request Changes on my own PR to ensure that I don't accidentally merge.

driusan:
@xlecours 

for point 1: I don't see what that would solve.

for point 2: The "getDataProvisionerWithFilters" function is intended to apply all base LORIS filters and add a single point where we can update them in the future if new ones are added. The fact that UserSiteMatch throws an exception if there's no getCenterID(s) function means that it can't get added to some pages, but we still want it to be added consistently across LORIS where applicable. Changing it to `getApplicableFilters(): Filter[]` would require every filter to know which filters are valid and which are not. I'd rather just have a "getDataProvisioner" function (which is how I originally had it) than that and get rid of the Base/withFilters distinction.

The reason I made it an abstract method instead of a flag is so that filters need to explicitly declare whether or not site filters are applicable, and can't forget to change a property and accidentally end up with an incorrect default.

There's also something like 7 moduleX_view_allsites permissions (or variations thereof) which will need to disable the access_all_profiles check despite having centers associated just because they use a different permission name, but that's a different story.

Hopefully these types of problems will be avoided for new filters on different non-site attributes as we improve the granularity of the permissions, which is why the sites are given special treatment in this PR.

modules/candidate_list/php/candidatelistrowprovisioner.class.inc
```diff
@@ -0,0 +1,100 @@
...
+ * PHP Version 7
+ *
+ * @category   Behavioural
+ * @package    Main
+ * @subpackage Imaging
+ * @author     Dave MacFarlane <david.macfarlane2@mcgill.ca>
+ * @license    http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link       https://www.github.com/aces/Loris/
+ */
+clASs CandidateListRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisioner
```
xlecours: `clASs` -> `class`

driusan: oops.. fixed. (I had done a `s/as/AS/` to fix the capitalization of the SQL keyword and that substring slipped in..)

driusan:
@xlecours can you re-review?

christinerogers:
Adding extra labels due to removal of Keyword search (if I read Dave's PR right) 
* Document at Release : Filters change -> check/update Help text, Module Readme
* Add to Release notes : because Keyword search is super important for the Data Dictionary above all modules (searching for any "age" fields, e.g.)

PR #4922, opened 2019-07-04
[Tools] Fix Raisinbread exporter location

### LINKED ISSUE #4878: [Tools] Fix raisinbread database dump export location
In line 5 of the file [DB_dump_table_data.php](https://github.com/aces/Loris/blob/21.0-release/tools/exporters/DB_dump_table_data.php) the export location should be: BASE_PATH_TO_LORIS/raisinbread/RB_files/RB_TABLE_NAME.sql.

johnsaigle:
### Brief summary of changes

Resolves #4878 by changing the export location to be built using the base URL instead of using a relative path.

Also adds some cleanup.

### DIFF
--- a/tools/exporters/DB_dump_table_data.php
+++ b/tools/exporters/DB_dump_table_data.php
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
 /**
  * This script generates data-only dumps for all tables in the currently active database.
  * Each table in the database gets dumped into a single file in the following location :
@@ -15,7 +15,7 @@
  * allowing to dump data using only the database name. The functionality to input
  * credentials interactively should be added.
  *
- * PHP Version
+ * PHP Version 7
  *
  * @category Main
  * @package  Loris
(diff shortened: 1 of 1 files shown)


tools/exporters/DB_dump_table_data.php
```diff
@@ -56,7 +56,8 @@
 
 // Loop through all tables to generate insert statements for each.
 foreach ($tableNames as $tableName) {
-    $filename = __DIR__ . "/../../raisinbread/RB_files/RB_$tableName.sql";
+    $filename = \NDB_Factory::singleton()->settings()->getBaseURL()
```
ridz1208: @johnsaigle doesnt baseurl return something like "abou-haider-dev.loris.ca" ?? thats not a path

johnsaigle: You're right, my bad. It should be project root or whatever the setting is called

ridz1208: i think thats base

PR #4924, opened 2019-07-04
[RB] small change in instructions

ridz1208:
### Brief summary of changes
adding helpful troubleshooting tips


raisinbread/README.md
```diff
@@ -111,6 +110,19 @@ RB_parameter_file.sql can become too big due to the complete header being dumped
...
 ```
+
+### Troubleshooting
+
+If you are having issues sourcing RB using the single command above, you can try to 
+use the following comands sequentially. These commands echo the name of the SQL 
+script before running it which helps to identify exactly what SQL statement is failing.
+
+```
+for n in /var/www/loris/SQL/0000-*.sql; do echo $n; cat $n | mysql || break; done;
```
CamilleBeau: These commands assume installation in this particular directory

ridz1208:
@CamilleBeau changes done

PR #4925, opened 2019-07-04
[Clean up] deletion of the full MINC header entry from parameter_file as not used anymore

cmadjar:
### Brief summary of changes

Until now, the whole header of a MINC file was dumped into a row of parameter_file, taking a lot of space in the database and the SQL dump created for that table, while not being used at all.

During the imaging meeting of July 4th, a decision has been made to no longer store that row to gain some space in the database, hence, removing it from the imaging pipeline code.

This is linked to https://github.com/aces/Loris-MRI/pull/453 PR on the LORIS-MRI side.

Note: the API does not look specifically for that header but dumps all the values present in the `parameter_file` table so no changes were needed.

### This resolves issue...

- [x] Github https://github.com/aces/Loris-MRI/issues/388

### To test this change...

- [ ] Make sure the SQL patch run correctly

### DIFF
--- a/SQL/New_patches/2019_07_04_remove_header_row_from_parameter_file_and_convert_back_to_Value_field_to_text.sql
+++ b/SQL/New_patches/2019_07_04_remove_header_row_from_parameter_file_and_convert_back_to_Value_field_to_text.sql
@@ -1,0 +1,3 @@
+DELETE FROM parameter_file WHERE ParameterTypeID=(SELECT ParameterTypeID FROM parameter_type WHERE Name='header' AND SourceFrom='parameter_file');
+DELETE FROM parameter_type WHERE Name='header' AND SourceFrom='parameter_file';
+ALTER TABLE parameter_file MODIFY Value TEXT;


cmadjar:
@ridz1208 I will need to update the RB documentation to remove the query to delete that `header` field when creating a RB dataset with new imaging data. Cannot do it now since 21.0 was not merged into minor...

driusan:
This seems like a pretty major change to me since there might be users depending on that data being in the database.

cmadjar:
It feels like we are going in circle with this one. Didn't we say that this specific header was completely useless and not queryable anyway at the last imaging meeting and that we should remove it to save some space in the `parameter_file` table?

driusan:
Yes, we did. I just mean in terms of which branch it should go to.

driusan:
(and I don't think we said it was *completely* useless, I'm pretty sure it's been used for various cron jobs in ibis over the years, it's just not nearly as useful as its cost..)

cmadjar:
OK. I doubt this is so much of a big change though but if you prefer moving it to major, no big deal. Can do that. Just let me know.

cmadjar:
@jesscall I had to rebase the branch to major and it dismissed your review. Would you mind reviewing it again? Thanks!!

SQL/New_patches/2019_07_04_remove_header_row_from_parameter_file_and_convert_back_to_Value_field_to_text.sql
```diff
@@ -0,0 +1,3 @@
+DELETE FROM parameter_file WHERE ParameterTypeID=(SELECT ParameterTypeID FROM parameter_type WHERE Name='header' AND SourceFrom='parameter_file');
+DELETE FROM parameter_type WHERE Name='header' AND SourceFrom='parameter_file';
+ALTER TABLE parameter_file MODIFY Value TEXT;
```
PapillonMcGill: @cmadjar I don't know the kind of value that are there but any value longer that 64KB will be trunct to the 1st 64KB.

If there should not be any value longer than that after the delete, I approve this PR

cmadjar: @PapillonMcGill the insertion script is filtering any value longer than 1000 characters so moving it back to text is fine. The only exception was the 'header' which was crazy long and that we are getting rid of in this PR.

PapillonMcGill: Thanks for the clarification.

PR #4927, opened 2019-07-04
[SQL] Missing statement from default schema

ridz1208:
### Brief summary of changes
Adding missing statemnt into the default schema for `loris` default project.

### DIFF
--- a/SQL/Release_patches/20.3_To_21.0_upgrade.sql
+++ b/SQL/Release_patches/20.3_To_21.0_upgrade.sql
@@ -281,7 +281,7 @@ FROM Project JOIN subproject
 WHERE Project.Name='loris';
 
 -- if the loris project was added, set all candidates to that default project
-UPDATE candidate SET ProjectID=(SELECT ProjectID FROM Project WHERE Name='loris') WHERE ProjectID IS NULL AND Entity_type='Human';
+UPDATE candidate SET ProjectID=(SELECT ProjectID FROM Project WHERE Name='loris') WHERE ProjectID IS NULL;
 
 SELECT 'Running: SQL/New_patches/2019-05-15-LorisMenuPermissions_QC_and_datarelease.sql';
 
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -9,6 +9,8 @@ CREATE TABLE `Project` (
     PRIMARY KEY (`ProjectID`)
 ) ENGINE = InnoDB  DEFAULT CHARSET=utf8;
 
+INSERT INTO `Project` (Name) VALUES ('loris');
+
 CREATE TABLE `subproject` (
     `SubprojectID` int(10) unsigned NOT NULL auto_increment,
     `title` varchar(255) NOT NULL,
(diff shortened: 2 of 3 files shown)


ridz1208:
@xlecours could you test/review?

SQL/Release_patches/20.3_To_21.0_upgrade.sql
```diff
@@ -281,7 +281,7 @@ FROM Project JOIN subproject
 WHERE Project.Name='loris';
 
 -- if the loris project was added, set all candidates to that default project
-UPDATE candidate SET ProjectID=(SELECT ProjectID FROM Project WHERE Name='loris') WHERE ProjectID IS NULL AND Entity_type='Human';
+UPDATE candidate SET ProjectID=(SELECT ProjectID FROM Project WHERE Name='loris') WHERE ProjectID IS NULL;
```
driusan: The default project named "loris" is added in this patch for 21.0.. it won't have existed in 20.3 and is only created above for projects that didn't previously use projects.

I'm not really sure what the right way to fix this is. Options are:
1. Always create I project named "loris" rather than only if there are no current projects
2. Make the default MIN(Project.ID) (or MAX) instead of Name="loris"
3. Write a script instead of an SQL patch to handle both cases before running the schema upgrade patch.

3 is probably the most robust, but I'm not sure how much extra work we want to be adding to writing/testing at this point in the release process

xlecours: I don't think there is any harm in creating a loris project for everyone.

ridz1208: Not sure i agree. Last time we discussed we said the "loris" default would be there unless its replaced by the study managers. The patch does exactly that, if there are no projects it adds one. If there already are projects it assumes the study managers will use one of these for the different candidates sessions and fantoms since they determined these projects themselves.

I feel like the current state is the closest to what we had previously discussed. We can add release statement describing that or a php script to just check the state of the database but i dont see it more complexe than that

driusan: The problem isn't that the patch does that for projects that don't have any projects set up, the problem is that the update statement trying to make sure there's no null ProjectID will fail (or at least, not update any rows) for projects that *do* already have some projects set up (unless they happen to have one named "loris")

ridz1208: @driusan in the current state of the upgrade patch no errors would be generated regardless of the state of your project table.

the following behaviour is to be expected.

 - if the project table is empty
  1. a new project "loris will be added"
  2. all candidates will be assigned to the new project

- if the project table already has projects:
 1. no new project is added
 2. candidates with no project affiliation will remain as they are (makes logical sens since if the projects table was populated, the useprojects was probably on and candidates SHOULD already belong to projects)

what I can do is add a php script that runs after the SQL to remind projects to add a project ID to their candidates with no projects

driusan: In that case I don't think we need a PHP script, we can just mention in the release notes that moving forward, all candidates should have projects and they should verify the state of their database

PR #4928, opened 2019-07-04
[Test] Add phpstan tool to test suite and fix detected errors.

johnsaigle:
### Brief summary of changes

I found a cool static analysis tool called [PHP Stan](https://github.com/phpstan/phpstan) which is like `phan` but much faster and flags additional issues. 

This PR cleans up some of the errors it flags.

~~I'm going to add another commit that adds phpstan as a dependency but it requires PHPUnit >= 7 to work properly. This should be solved by #4433 so I'm marking this PR as blocked for now.~~

### DIFF
--- a/php/libraries/BVL_Feedback_Panel.class.inc
+++ b/php/libraries/BVL_Feedback_Panel.class.inc
@@ -35,7 +35,7 @@ class BVL_Feedback_Panel
      *
      * @private NDB_BVL_Feedback
      */
-    protected $_feedbackThread;
+    protected $feedbackThread;
 
     /**
      * An array containing data that will be passed to the Smarty template.
@@ -58,7 +58,7 @@ class BVL_Feedback_Panel
         ?string $commentID=null
     ) {
 
-        $this->_feedbackThread
+        $this->feedbackThread
             = \NDB_BVL_Feedback::singleton(
                 \User::singleton()->getUsername(),
                 $candID,
@@ -77,7 +77,7 @@ class BVL_Feedback_Panel
      */
     function display(): string
     {
-        $feedbackProfile = $this->_feedbackThread->_feedbackCandidateProfileInfo;
+        $feedbackProfile = $this->feedbackThread->_feedbackCandidateProfileInfo;
         $candidateID     = new CandID($feedbackProfile["CandID"]);
 
         $this->tpl_data['candID'] = (string) $candidateID;
@@ -87,16 +87,16 @@ class BVL_Feedback_Panel
 
         $this->tpl_data['sessionID'] = $feedbackProfile["SessionID"] ?? null;
 
-        $feedbackObject = $this->_feedbackThread->_feedbackObjectInfo;
+        $feedbackObject = $this->feedbackThread->_feedbackObjectInfo;
         $this->tpl_data['commentID'] = $feedbackObject["CommentID"] ?? null;
 
-        $this->tpl_data['thread_list']    = $this->_feedbackThread->getThreadList();
-        $this->tpl_data['feedback_level'] = $this->_feedbackThread->_feedbackLevel;
+        $this->tpl_data['thread_list']    = $this->feedbackThread->getThreadList();
+        $this->tpl_data['feedback_level'] = $this->feedbackThread->_feedbackLevel;
         $this->tpl_data['feedback_types']
             = NDB_BVL_Feedback::getFeedbackTypes();
         $this->tpl_data['feedbacktypes']  = NDB_BVL_Feedback::getFeedbackTypes();
 
-        $summary = $this->_feedbackThread->getSummaryOfThreads();
+        $summary = $this->feedbackThread->getSummaryOfThreads();
         $this->tpl_data['thread_summary_headers'] = json_encode($summary);
 
         $field_names = Utility::getSourcefields($_REQUEST['test_name'] ?? '');
--- a/php/libraries/BVL_Feedback_Panel.class.inc
+++ b/php/libraries/BVL_Feedback_Panel.class.inc
@@ -33,7 +33,7 @@ class BVL_Feedback_Panel
     /**
      * The feedback thread that this panel is for
      *
-     * @private NDB_BVL_Feedback
+     * @protected NDB_BVL_Feedback
      */
     protected $feedbackThread;
 
--- a/php/libraries/BVL_Feedback_Panel.class.inc
+++ b/php/libraries/BVL_Feedback_Panel.class.inc
@@ -35,12 +35,12 @@ class BVL_Feedback_Panel
      *
      * @private NDB_BVL_Feedback
      */
-    private $_feedbackThread;
+    protected $_feedbackThread;
 
     /**
      * An array containing data that will be passed to the Smarty template.
      *
-     * @var array
+     * @public array
      */
     public $tpl_data;
 
--- a/php/libraries/BVL_Feedback_Panel.class.inc
+++ b/php/libraries/BVL_Feedback_Panel.class.inc
@@ -33,16 +33,16 @@ class BVL_Feedback_Panel
     /**
      * The feedback thread that this panel is for
      *
-     * @var NDB_BVL_Feedback
+     * @private NDB_BVL_Feedback
      */
-    var $_feedbackThread;
+    private $_feedbackThread;
 
     /**
      * An array containing data that will be passed to the Smarty template.
      *
      * @var array
      */
-    var $tpl_data;
+    public $tpl_data;
 
     /**
      * Creates the feedback thread for the given combination of candID,
--- a/php/libraries/BVL_Feedback_Panel.class.inc
+++ b/php/libraries/BVL_Feedback_Panel.class.inc
@@ -38,6 +38,13 @@ class BVL_Feedback_Panel
     var $_feedbackThread;
 
     /**
+     * An array containing data that will be passed to the Smarty template.
+     *
+     * @var array
+     */
+    var $tpl_data;
+
+    /**
      * Creates the feedback thread for the given combination of candID,
      * sessionID and commentID.
      *
@@ -101,7 +108,7 @@ class BVL_Feedback_Panel
 
         $this->tpl_data['FieldNames'] = json_encode($fields);
 
-        $smarty = new Smarty_neurodb("bvl_feedback_panel");
+        $smarty = new Smarty_NeuroDB("bvl_feedback_panel");
         $smarty->assign($this->tpl_data);
         $html = $smarty->fetch("bvl_feedback_panel.tpl");
         return $html;
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -59,6 +59,13 @@ class Database
     var $_preparedStoreHistory;
 
     /**
+     * The name of the database.
+     *
+     * @access public
+     */
+    var $_databaseName = '';
+
+    /**
      * Singleton design pattern implementation - creates the object
      * and also connects to the database if necessary.
      * NOTE There is no return type declaration for this function as it returns
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -59,7 +59,7 @@ class Database
     var $_preparedStoreHistory;
 
     /**
-     * The name of the database.
+     * The name of the database. Used internally for printing log messages.
      *
      * @access private
      */
@@ -1469,4 +1469,3 @@ class Database
         );
     }
 }
-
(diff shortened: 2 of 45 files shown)


johnsaigle:
Forgot to do `make checkstatic`. There are some phan errors I'll fix here.

xlecours:
Nice catches!
I see that you choose `var` instead of a visibility keyword. Is it recommended?

johnsaigle:
@xlecours No I don't think the tool cares. As far as I know `var` is the same as `public` in PHP. I can change it if `public` is preferred.

xlecours:
I usually prefer explicit statements. But in this case, the visibility would need to be set to public, because there might be calls like `$that->_variablename`, and ideally they should be set to protected or private (in a utopian futur). 

[TL;DR] leave them `var`

driusan:
"public" is better than "var" because it's explicit. The "var" keyword shouldn't be used outside of legacy code.

johnsaigle:
There are some phan issues that exist on the `21.0-release` branch.. not sure how they got there.

Many of these are/will be addressed by https://github.com/aces/Loris/pull/4914 so I'm going to mark this PR as Blocked until that one is merged.

johnsaigle:
This PR now includes phpstan in the test suite. I was able to run it at max level and fix all detected errors! Pretty sweet.

johnsaigle:
Travis is failing because PHP 7.2 is too old to run the dependencies for this version.

Since this is unlikely to go into LORIS 22 anyway, I think I'll punt this until 7.3 is required

johnsaigle:
This PR requires PHP 7.3 so ti is Blocked by #5723.

johnsaigle:
@driusan @xlecours This is ready for review.

php/libraries/NDB_Form.class.inc
```diff
@@ -206,10 +206,9 @@ class NDB_Form extends NDB_Page
     }
 
     /**
-     * Converts the results of this form to a JSON format to be retrieved
-     * with ?format=json
+     * {@inheritDoc}
```
ridz1208: what does this do ?

php/libraries/NDB_Menu.class.inc
```diff
@@ -115,11 +113,22 @@ class NDB_Menu extends NDB_Page
...
+        $smarty = new Smarty_NeuroDB($this->Module->getName());
         $smarty->assign('mode', $this->mode);
         $smarty->assign($this->getTemplateData());
         $html = $smarty->fetch("menu_$this->menu.tpl");
         return $html;
     }
+
+    /**
+     * {@inheritDoc}
+     * TODO
```
ridz1208: TODO !!

php/libraries/NDB_Page.class.inc
```diff
@@ -721,6 +722,7 @@ class NDB_Page implements RequestHandlerInterface
     {
         switch($this->format) {
         case 'json':
+            // NOTE this function doesn't actually exist but seems to work?!
```
ridz1208: if it works, why add the function ?

php/libraries/NDB_Form.class.inc
driusan: `{@inheritDoc}` is a convention in PHP recognized by phpcs/phpdoc/etc so that you don't need to copy the doc from the parent class.

php/libraries/NDB_Page.class.inc
johnsaigle: I don't remember the nuances of this 🤷‍♂ I'll delete it

php/libraries/BVL_Feedback_Panel.class.inc
```diff
@@ -33,9 +33,16 @@ class BVL_Feedback_Panel
     /**
      * The feedback thread that this panel is for
      *
-     * @var NDB_BVL_Feedback
+     * @private NDB_BVL_Feedback
      */
-    var $_feedbackThread;
+    private $_feedbackThread;
```
driusan:
```suggestion
    protected $_feedbackThread;
```

protected is more appropriate here, any (theoretical) class that derives from this would need access to it.

php/libraries/BVL_Feedback_Panel.class.inc
```diff
@@ -33,9 +33,16 @@ class BVL_Feedback_Panel
...
+
+    /**
+     * An array containing data that will be passed to the Smarty template.
+     *
+     * @var array
```
driusan:
```suggestion
     * @public array
```

consistency

php/libraries/Database.class.inc
```diff
@@ -58,6 +58,13 @@ class Database
 
     var $_preparedStoreHistory;
 
+    /**
+     * The name of the database.
```
driusan:
```suggestion
     * The name of the database. Used internally for printing log messages.
```

php/libraries/LorisForm.class.inc
```diff
@@ -49,6 +49,8 @@ class LorisForm
         'email',
         'maxlength',
     );
+    public $validOperators;
```
driusan: I don't understand this one. It looks like it's never set in the code and I don't see why it's public.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2544,17 +2550,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
-            $formArray = $this->form->toElementArray();
-        } else {
-            $smarty   = new Smarty_neurodb;
-            $renderer = new HTML_QuickForm_Renderer_Array($smarty);
-
-            $this->form->accept($renderer);
-
-            $formArray = $renderer->toArray();
-        }
+        $formArray = $this->form->toElementArray();
```
driusan: What about the other part of the else statement? What is this trying to solve? I think this change of behaviour is likely to break things with different types of instruments

php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
```diff
@@ -26,7 +26,10 @@ use LORIS\StudyEntities\Candidate\CandID;
  */
 class NDB_BVL_InstrumentStatus_ControlPanel extends NDB_BVL_InstrumentStatus
 {
-
+    public $tpl_data;
+    public $validityEnabled;
+    public $validityRequired;
+    private $_instrument;
```
driusan: protected

php/libraries/LorisForm.class.inc
johnsaigle: It's used by `function validateRequireIf`... sort of. If you take a look at that function it looks like it needs a bit of refactoring. Should I do that in this PR or no?

php/libraries/NDB_BVL_Instrument.class.inc
johnsaigle: I believe there was an issue with `HTML_QuickForm_Renderer_Array` because it doesn't exist in the code. 

I made this PR 6 months ago so I'm not sure about the details.

php/libraries/LorisForm.class.inc
driusan: yeah, but it's only used in this context:

```
            throw new LorisException(
                "operator is not a valid operator from the set "
                . join(', ', $this->validOperators)
            );
```

and it's never set.. when I grep the code for 'validOperators' it's also never set externally even though it's public. I guess you're right that it needs refactoring and the reference to it seems to be a bug, but fixing it properly is out of the scope of this PR.

johnsaigle: Do you object to adding this variable so that we can satisfy phpstan for now and leave refactoring for the future?

php/libraries/NDB_BVL_Instrument.class.inc
driusan: I just did a little investigation of the impact of losing this else case and I think it means the API would be broken for the endpoint which gets the instrument in a JSON format for projects that had old instruments that couldn't be updated to LorisForm and were still using QuickForm.

There was only 1 such instrument that I could find when I searched the IBIS repo, and I'm not aware of anyone/thing depending on that endpoint of the API, so I guess it should be fine.

driusan: @xlecours or @ridz1208 do you want to chime in or know of anything using the `/projects/$proj/instruments/$instrumentname` endpoint?

php/libraries/LorisForm.class.inc
driusan: No objections, I think it's fine to satisfy phpstan.

php/libraries/NDB_BVL_Instrument.class.inc
johnsaigle: We can add a note in the CHANGELOG about this if it ends up being a breaking change.

johnsaigle:
@driusan I've incorporated your suggestions. I added a note in the CHANGELOG about QuickForm complications and added the `Caveat...` label.

Could you re-review?

php/libraries/BVL_Feedback_Panel.class.inc
```diff
@@ -33,9 +33,16 @@ class BVL_Feedback_Panel
     /**
      * The feedback thread that this panel is for
      *
-     * @var NDB_BVL_Feedback
+     * @private NDB_BVL_Feedback
      */
-    var $_feedbackThread;
+    protected $feedbackThread;
```
driusan: The doc (`@private`) and the code (`protected`) here still don't seem to agree.

php/libraries/Module.class.inc
```diff
@@ -228,7 +228,8 @@ abstract class Module extends \LORIS\Router\PrefixRouter
      *
      * @param string $page The name of the page within the module to load.
      *
-     * @return a NDB_Page subclass representing this page.
+     * @return NDB_Page|NDB_Form|NDB_Menu|NDB_Menu_Filter|NDB_Menu_Filter_Form
```
driusan:
```suggestion
     * @return NDB_Page
```

All of the things listed are subclasses of NDB_Page. If phpstan was complaining, it's probably because of the "a".

php/libraries/SessionID.php
```diff
@@ -50,9 +50,9 @@ public function getType(): string
      */
     protected function validate(string $value): bool
     {
-        return (intval($value) > 0)
-            && (strlen($value) <= self::SESSIONID_MAX_LENGTH)
-            && (is_integer($value));
+        return intval($value) > 0
+            && strlen($value) <= self::SESSIONID_MAX_LENGTH
+            && ctype_digit($value);
```
driusan: what was wrong with is_integer?

test/phpstan-loris.neon
```diff
@@ -0,0 +1,10 @@
+parameters:
+    autoload_directories:
+        - %currentWorkingDirectory%/php/
+        - %currentWorkingDirectory%/src/
+        - %currentWorkingDirectory%/modules/
+        - %currentWorkingDirectory%/test/integrationtests/
+        - %currentWorkingDirectory%/test/unittests/
+    excludes_analyse:
+        - %currentWorkingDirectory%/modules/*/ajax/*
+        - %currentWorkingDirectory%/modules/*/test/*
```
driusan: I don't see why these are excluded.

johnsaigle: Just provisionally. The PR is long enough

php/libraries/Module.class.inc
johnsaigle: I'll try

test/phpstan-loris.neon
driusan: fair enough

php/libraries/SessionID.php
johnsaigle: > is_int() returns true if the argument is an integer type, ctype_digit() takes a string argument and returns true if all the characters in the string are digits.

The value is passed as a string so `is_integer()` always returns false

https://stackoverflow.com/questions/236406/is-there-a-difference-between-is-int-and-ctype-digit

php/libraries/BVL_Feedback_Panel.class.inc
```diff
@@ -33,9 +33,16 @@ class BVL_Feedback_Panel
     /**
      * The feedback thread that this panel is for
      *
-     * @var NDB_BVL_Feedback
+     * @private NDB_BVL_Feedback
```
johnsaigle:
```suggestion
     * @protected NDB_BVL_Feedback
```

php/libraries/BVL_Feedback_Panel.class.inc
johnsaigle: Thanks, just fixed it above

PR #4929, opened 2019-07-04
[candidate_parameters] Add DoD Feature

jesscall:
### Brief summary of changes
This PR adds a new tab to candidate parameters which can be used to record a candidate's date of death. The PR also adds support for this new feature in instrument class. If the date of administration of an instrument is after the date of death, the candidate's age at death will be displayed rather than age at time of administration. 

This change is being applied as a result of the DoD meeting which took place on June 25, 2019.

[NOTE]: I am basing this PR on 21.0-release branch for now, but will change it to minor after release happens.

### To test this change...
- [ ]  Navigate to candidate parameters. Make sure that you can only update date of death if you have the permission `candidate_dod_edit`. Make sure any update to the date actually saves.
- [ ] Fill out any instrument of your choice. Set `Date of Administration` to a later date than the DoD. Upon refresh, the age displayed should correspond to the candidate's age at death (in months).


johnsaigle:
It looks like the changes might be causing some issues with the integration tests on Travis.

kongtiaowang:
I have restarted Travis to debug this issue.

SQL/New_patches/2019_07_04_add_DoD_feature.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE candidate
```
kongtiaowang: 0000-00-00-schema.sql file also needs these changes.
SQL/0000-00-01-Permission.sql

modules/candidate_parameters/jsx/CandidateDOD.js
```diff
@@ -0,0 +1,156 @@
...
+          />
+          <StaticElement
+            label='DCCID'
+            text={this.state.data.candID}
+          />
+          <StaticElement
+            label='Disclaimer:'
+            text='Any changes to the date of death requires an
+            administrator to run the fix_candidate_age script.'
+            // class='form-control-static text-danger bg-danger col-sm-10'
```
jesscall: blocked by #4915

jesscall:
Something to consider: should we have a config option for `DOD Format`?

johnsaigle:
@jesscall Would it need to be different from the one I added in #5004 ?

jesscall:
@johnsaigle I believe the cases are different. 

In the case of DOB Format, when a date is entered we strip the day and store '01' in the DB so that candidates cannot be identified if ever these dates are queried and released. The idea would be the same for DOD. Your PR is addressing a display preference, which I think is different.

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -534,8 +534,8 @@ function editConsentStatusFields($db, $user)
  */
 function editCandidateDOD(\Database $db, \User $user): void
 {
-    $candID       = $_POST['candID'];
-    $dod          = $_POST['dod'];
+    $candID = $_POST['candID'];
+    $dod    = $_POST['dod'];
```
johnsaigle: It may be good to use `DateTime` here to validate whether this is a proper date string and throw an exception if it's not.

modules/candidate_parameters/ajax/getData.php
```diff
@@ -490,3 +493,29 @@ function getConsentStatusHistory($pscid)
...
+/**
+ * Handles the fetching of candidate's date of death.
+ *
+ * @throws DatabaseException
+ *
+ * @return array
+ */
+function getDODFields(): array
+{
+    $candID = $_GET['candID'];
```
```suggestion
    $candID = new CandID($_GET['candID']);
```

Just to make sure that the input is valid. You may get a security issue here if you later return user-supplied data to the front-end. Passing it through this class will make sure that the user supplied a valid CandID first.

You'll probably want to wrap this in a try-catch with a \DomainException and return an error if encountered.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2095,16 +2110,18 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+                      'Date_taken' => $this->getFieldValue('Date_taken'),
+                      'DoB'        => $this->getDoB(),
+                      'DoD'        => $this->getDoD(),
+                     );
+        $dateTaken = empty($date) ? $dates['Date_taken'] : $date;
 
         $age = array();
-        if (empty($date)) {
-            $age = Utility::calculateAge($dates['DoB'], $dates['Date_taken']);
+        if (empty($dates['DoD']) || $dates['DoD'] > $dateTaken) {
```
johnsaigle: This date comparison should probably be done with `DateTime` objects. Depending on the format of the date string you may get weird results.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2130,6 +2147,55 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+        $db        = \Database::singleton();
+        $CommentID = $this->getCommentID();
+        $dod       = $db->pselectOne(
+            "SELECT c.DoD FROM flag f
+            JOIN session s ON (f.SessionID=s.ID)
+            JOIN candidate c USING (CandID)
+            WHERE f.CommentID=:CID",
+            array('CID' => $CommentID)
+        );
+        return $dod;
```
```suggestion
        return new DateTime($dod);
```

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2130,6 +2147,55 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         return $dob;
     }
 
+    /**
+     * Get the candidates DOD.
+     *
+     * @return mixed string or array The candidate's DOD.
+     */
+    function getDoD()
```
```suggestion
    function getDoD(): DateTime
```

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2130,6 +2147,55 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         return $dob;
     }
 
+    /**
+     * Get the candidates DOD.
+     *
+     * @return mixed string or array The candidate's DOD.
```
```suggestion
     * @return DateTime The candidate's DOD.
```

johnsaigle:
Made some requests largely to do with #5003.

php/libraries/NDB_BVL_Instrument.class.inc
jesscall: This is tricky because `calculateAge` expects strings and not `DateTime` objects

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2130,6 +2147,55 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         return $dob;
     }
 
+    /**
+     * Get the candidates DOD.
+     *
+     * @return ?string The candidate's DOD.
+     */
+    function getDoD(): ?string
```
driusan: Since this is a date that should really never change, why not return a ?\DateTimeImmutable?

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -847,10 +847,25 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         $this->addBasicDate('Date_taken', 'Date of Administration', $dateOptions);
 
         if (strrpos($this->testName, '_proband') === false) {
-            $this->addScoreColumn(
-                'Candidate_Age',
-                'Candidate Age (Months)'
-            );
+            $dates = array(
+                      'Date_taken' => $this->getFieldValue('Date_taken'),
```
ridz1208: this breaks lorisform parser

raisinbread/RB_files/RB_Config.sql
```diff
@@ -33,8 +33,8 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (33,37,'/PATH/TO/Genomic
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (34,38,'/data/uploads/');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (35,40,'main.css');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (36,41,'25');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (37,44,'localhost');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (38,45,'http://localhost');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (37,44,'jcallegaro-dev.loris.ca');
```
maltheism: localhost or jcallegaro-dev.loris.ca

raisinbread/RB_files/RB_Config.sql
```diff
@@ -33,8 +33,8 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (33,37,'/PATH/TO/Genomic
...
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (38,45,'http://jcallegaro-dev.loris.ca');
```
maltheism: localhost or jcallegaro-dev.loris.ca

raisinbread/RB_files/RB_Config.sql
```diff
@@ -54,7 +54,7 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (54,72,'yourname@example
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (55,73,'/data/demo/bin/mri/dicom-archive/get_dicom_info.pl');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (56,74,'1');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (57,75,'1');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (58,76,'dcm2mnc');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (58,76,'dcm2mnc-centos-6p9-T2star');
```
maltheism: `-centos-6p9-T2star` should be here?

modules/candidate_parameters/jsx/CandidateDOD.js
```diff
@@ -0,0 +1,183 @@
...
```
maltheism: commented out code

raisinbread/RB_files/RB_Config.sql
jesscall: oops, ill fix these

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -525,3 +527,42 @@ function editConsentStatusFields($db, $user)
...
+
+    if (!$dod) {
+        throw new \LorisException('Date not valid.');
+    }
+
+    if (!empty($dod)) {
+        $config    = \NDB_Config::singleton();
+        $dodFormat = $config->getSetting('dodFormat');
+        if ($dodFormat === 'YM') {
+            $strippedDate = $dod->format('Y-m') . "-01";
```
```suggestion
            $strippedDate = $dod->format('Y-m-01');
```

I think that this is possible, just for future reference.

modules/candidate_parameters/ajax/getData.php
```diff
@@ -493,3 +494,28 @@ function getConsentStatusHistory($pscid)
     }
     return $formattedHistory;
 }
+
+/**
+ * Handles the fetching of candidate's date of death.
+ *
+ * @throws DatabaseException
```
```suggestion
```

php/libraries/NDB_BVL_Instrument.class.inc
johnsaigle: I agree that's my favourite solution but not necessary.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2161,6 +2181,55 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+     * Calculates the time since candidate's death.
+     * If $date is passed, this function  will calculate the age as of
+     * the value of that argument. Otherwise the date will be calculated
+     * based on the value of the Date_taken field in the database.
+     *
+     * @param string $date (optional) The date that the age is to
+     *                     be calculated based on.
+     *
+     * @return array The time since candidate's death in an array containing
+     *         years, months and days.
```
johnsaigle: There's a bit of complexity here with line 2223 and it depends what you want.

That Utility function assumes that a month always has 30 days. If you're looking for a precise "time since death" then this function will give you the wrong answer.

tools/CouchDB_Import_Demographics.php
```diff
@@ -16,6 +16,10 @@ class CouchDBDemographicsImporter {
             'Description' => 'Date of Birth',
             'Type' => 'varchar(255)'
         ),
+        'DoD' => array(
+            'Description' => 'Date of Death',
+            'Type' => 'varchar(255)'
```
johnsaigle: This has the Date type in your patch, not varchar(255). It feels like these should match.

php/libraries/NDB_BVL_Instrument.class.inc
jesscall: This is a valid point, though is the candidate age / time since death every really accurate in the case that the DOB used has format 'Ym' ? (This is the case for CCNA)

Should we be more accurate in this case?

johnsaigle: I'm not sure. I don't know enough about the science part of it. I think it depends what the Date of Death is being used for. I know calculateAge isn't supposed to be touched because it may impact reproducibility if we start calculating the age differently.

Maybe you could bring it up in the LORIS meeting just to be safe?

jesscall: After our discussion, I looked into the CCNA post-mortem instruments / use case and realized that this feature is never used... I may get rid of this, unless there are strong feelings about keeping it ?

johnsaigle: Sure, if there's no use for it.

tools/CouchDB_Import_Demographics.php
jesscall: I was following what was done for Date of Birth (above), should I change it ?

johnsaigle: That makes sense. @ridz1208 can you comment here? I don't know enough about CouchDB to know what's correct here.

modules/candidate_parameters/jsx/CandidateDOD.js
jesscall: uncommenting - but should be noted that this only works if the changes in  #4915 are accepted and merged 
specifically this part:
https://github.com/aces/Loris/pull/4915/files#diff-bb24eb9a9e1227db8dd174deb9712093R1445

tools/CouchDB_Import_Demographics.php
ridz1208: yes date is better

you can change the dob one too

modules/candidate_parameters/jsx/CandidateDOD.js
johnsaigle: Should this PR be marked as Blocked then?

maltheism:
@jesscall I'm getting the error seen in the picture when adding a DOD. Seems like it does get added when I refresh the page to see if it's still there.
error

modules/candidate_parameters/ajax/getData.php
```diff
@@ -51,11 +51,12 @@
 case 'candidateDOB':
     echo json_encode(getDOBFields());
     exit;
+case 'candidateDOD':
+    echo json_encode(getDODFields());
+    exit;
 default:
     header("HTTP/1.1 404 Not Found");
-    exit;
```
driusan: Why is this exit removed?

maltheism: @driusan I believe a rebase mistake. There was an earlier one that caused a bug.

jesscall: oops, yup it was a rebase mistake. 
I'll issue another PR to add it back

driusan: you should probably re-rebase it before you make the fix..

SQL/New_patches/2019_07_04_add_DoD_feature.sql
```diff
@@ -0,0 +1,16 @@
...
+INSERT IGNORE INTO permissions (code, description, categoryID) VALUES
+    ('candidate_dod_edit', "Edit dates of death", 2);
+
+INSERT IGNORE INTO user_perm_rel (userID, permID) VALUES
+    (1, (SELECT permID FROM permissions where code='candidate_dod_edit'));
+
+INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber)
+    SELECT 'dodFormat', 'Format of the Date of Death', 1, 0, 'text', ID, 'DOD Format', 10 FROM ConfigSettings WHERE Name="study";
+
+INSERT INTO Config (ConfigID, Value) SELECT ID, 'YMd'  FROM ConfigSettings WHERE Name="dodFormat";
```
driusan: These config settings look like they're added in the patch and raisinbread but missing from the default schema.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -165,6 +165,17 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+    /**
+     * True if the instrument is administered after the candidate's death.
+     * This is used to determine which candidate age should be displayed as
+     * part of metadata fields. (Either Candidate Age or Candidate Age at Death).
+     * To be called from within individual instrument forms.
+     *
+     * @var    bool
+     * @access private
+     */
+    var $postMortem = false;
```
driusan: there should be a real access modifier not a `var` keyword in new code.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2166,6 +2186,25 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         return $dob;
     }
 
+    /**
+     * Get the candidates DOD.
+     *
+     * @return ?string The candidate's DOD.
+     */
+    function getDoD(): ?string
```
driusan: why is this a string and not a datetime?

jesscall: two reasons:
- The helper is used mainly in `getCandidateAge` function which calls `Utility::calculateAge`. This utility expects both date parameters to be strings.
- The similar helper `getDOB` for dates of birth also returns string.

johnsaigle: I agree that it should be DateTime ideally (and mentioned this in an earlier review) but given that this PR was opened in July I'm in favour with leaving it for a future refactoring effort.

php/libraries/NDB_BVL_Instrument_LINST.class.inc
```diff
@@ -79,10 +79,17 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
         $this->addBasicDate('Date_taken', 'Date of Administration', $dateOptions);
 
         if (strrpos($this->testName, '_proband') === false) {
-            $this->addScoreColumn(
-                'Candidate_Age',
-                'Candidate Age (Months)'
-            );
+            if ($this->postMortem) {
```
christinerogers, line 82: Just found a bug in LINST -- Missing negation here, should be
` if ( ! $this->postMortem) { `

Done Correctly: [ Instrument class line 89](https://github.com/aces/Loris/blob/fef5b306936e7b464dcac49e4c5965198333c1a1/php/libraries/NDB_BVL_Instrument.class.inc#L859)
incorrectly: [ LINST line 77 ](https://github.com/aces/Loris/blob/c2e956a80c6fc1d3c9b70f94bf05e3f60539f2f8/php/libraries/NDB_BVL_Instrument_LINST.class.inc#L77)

PR #4861, opened 2019-06-12
[Test] unittests for Utility.class.inc

AlexandraLivadas:
### Brief summary of changes
I have started writing unit tests for the Utility class on this branch. 

### This is related to:
#4378 

### To test this change...
Run "npm run tests:unit -- --filter Utility" or check out Travis

PR #4936, opened 2019-07-10
[Test] Adding to unittests for Candidate.class.inc

AlexandraLivadas:
## Brief summary of changes
I am adding more unit tests for [Candidate.class.inc](https://github.com/aces/Loris/blob/major/php/libraries/Candidate.class.inc) in the [CandidateTest.php](https://github.com/aces/Loris/blob/major/test/unittests/CandidateTest.php) file. I am also making changes and adding to previous tests.

#### Testing instructions 

Run "npm run tests:unit -- --filter Candidate" or check out Travis

**Note:** If you run "npm run tests:unit -- --filter Candidate" it will also run some tests from the LORIS\StudyEntitites\Candidate\Test\CandID_Test file!


test/unittests/CandidateTest.php
```diff
@@ -244,6 +311,271 @@ public function testSetDataWithValueSucceeds()
...
+            $this->_candidateInfo['Sex'],
+            $this->_candidate->getCandidateSex()
+        );
+    }
+
+    /**
+     * Test getCandidateEthnicity returns the correct ethnicity of the candidate
+     *
+     * @note   This function is deprecated so it was causing a PHAN error. 
+               It is marked test incomplete for this reason
```
PapillonMcGill:
```suggestion
 *              It is marked test incomplete for this reason
```

test/unittests/CandidateTest.php
```diff
@@ -467,6 +858,101 @@ public function testValidatePSCID()
...
+     * @covers Candidate::getConsents
+     * @return void
+     */
+    public function testGetConsents()
+    {
+        $this->_setUpTestDoublesForSelectCandidate();
+        $this->_candidate->select(969664);
+ 
+        $result = array(
+                      array('ConsentID' => 1,
```
PapillonMcGill: `=>` should all be aligned

test/unittests/CandidateTest.php
```diff
@@ -467,6 +858,101 @@ public function testValidatePSCID()
...
+     * @return void
+     */
+    public function testGetParticipantStatusSubOptions()
+    {
+        $this->_setUpTestDoublesForSelectCandidate();
+        $this->_candidate->select(969664);
+
+        $result = array(
+                      array('ID' => 1,
+                            'Description' => 'description1'));
```
PapillonMcGill: same here

test/unittests/CandidateTest.php
```diff
@@ -11,6 +11,71 @@
  * @link     https://www.github.com/aces/Loris/
  */
 use PHPUnit\Framework\TestCase;
+
+/**
+ * Fake Candidate class that has exact copies of methods in the
+ * Candidate class to test certain SQL queries easily
```
PapillonMcGill: Can you change the code so calling `$DB           =& Database::singleton();` would use the fake DB? i.e. using different access parametre than in the config.xml file to connect to the DB?

This would prevent you code from being obsolete in case the actual libraries get updated/modified

AlexandraLivadas:
Hello! So in the latest commit, I got rid of the FakeCandidate class that I was using to test two of the methods because @PapillonMcGill pointed out another way to test them without having to create a fake class with fake methods. Let me know if this looks good to everyone! I'm also thinking of applying the same logic to the Utility tests in PR #4861 since the FakeUtility class is very very long.

test/unittests/CandidateTest.php
```diff
@@ -211,7 +237,8 @@ public function testSetDataWithArraySucceeds()
 
     /**
      * Test setData method sets data when passing variable name and value
-     *
+     * TODO Same test as above?
```
AlexandraLivadas: This test seems to be the same as the one above it. Could someone take a look?

driusan: Looks the same to me too

driusan: @kongtiaowang ?

test/unittests/CandidateTest.php
```diff
@@ -244,6 +271,271 @@ public function testSetDataWithValueSucceeds()
...
+     *         It is marked test incomplete for this reason
+     *         however it is fully implemented and functional
+     * @covers Candidate::getCandidateEthnicity
+     * @return void
+     */
+    public function testGetCandidateEthnicity()
+    {
+        $this->markTestIncomplete(
+            "This test is marked as incomplete because the function is deprecated"
+        );
```
driusan: I think you should just delete it, then (also GitHub is highlighting the commented out code as an error because of the `/**`

test/unittests/CandidateTest.php
```diff
@@ -244,6 +271,271 @@ public function testSetDataWithValueSucceeds()
...
+    public function testLastRecordChangeBy()
+    {
+        $this->_setUpTestDoublesForSelectCandidate();
+        $this->_candidate->select($this->_candidateInfo['CandID']);
+
+        $this->assertEquals(
+            $this->_candidateInfo['UserID'],
+            $this->_candidate->lastRecordChangeBy()
+        );
+    }
```
driusan: shouldn't this change something and assert that the UserID changed too?

AlexandraLivadas: As of now, the setData() method doesn't seem to change the UserID when something else is changed. So if I were to set the 'Sex' or something to a new value, the UserID would not change. I'm not sure where or how the UserID is changed once a change is made by a user, I'm assuming somehow in the frontend?

test/unittests/CandidateTest.php
kongtiaowang: It is the same!

johnsaigle:
@AlexandraLivadas There was an issue with Travis that has since been resolved. You should do `git fetch aces` and then `git rebase -i aces/major` and finally force push the changes from your local branch to this one. That should allow Travis to properly test your code.

test/unittests/CandidateTest.php
driusan: That's a good question.. @ridz1208 @xlecours do either of you know how the UserID gets set in the candidate table?

ridz1208: do we even have a way to change any of the values in the candidate from the front end ??

and if so which chnages should update the userId and which should not ? cause if I remember correctly some candidate parameters are stored in the candidate table. 

My guess is the usedId is only updated on creation

johnsaigle: Afaik, there isn't a way to change it from the front-end

AlexandraLivadas: I think the UserID is supposed to be updated whenever that user makes a change. But in the Candidate.class.inc file, I'm not seeing anything that performs that action. There's also no setUserID or anything.

test/unittests/CandidateTest.php
driusan: @AlexandraLivadas are you planning on removing this duplicate?

AlexandraLivadas:
@johnsaigle @kchatpar @ridz1208 Let me know what you think!

test/unittests/CandidateTest.php
```diff
@@ -29,14 +29,14 @@ class CandidateTest extends TestCase
      */
     private $_candidateInfo
         = array(
-           'RegistrationCenterID'     => '2',
+           'RegistrationCenterID'     => 2,
```
driusan: If this supposed to simulate the data returned by the database? If so, I think it should remain a string. The database returns everything as a string (or null) regardless of the MySQL type.

driusan:
@AlexandraLivadas I just skimmed it, but the duplicate test seems to still be there, and I noticed the potential problem of the mock array not matching the format the $DB->pselect would return it.

PR #4940, opened 2019-07-15
[Config] Add settingEnabled() helper function

johnsaigle:
## Brief summary of changes

Adds a new helper function `settingEnabled` to determine whether a boolean-style config setting is enabled or not.

This is designed to add some syntactic sugar and  hide some of the implementation details of our Config library from calling code. 

The calling code only cares whether a setting is enabled. Developers shouldn't need to concern  themselves with doing a strict string type comparison on config settings.

e.g. this code:

```php
if ($config->getSetting('useEDC') === 'true') {
    ...
}
```

can now be written as

```php
if ($config->settingEnabled('useEDC')) {
    ...
}
```

I feel like this is an improvement because it's a bit cleaner. I think it removes some potential confusion also as it's not immediately clear why LORIS code is doing a string check for something that looks like a bool.

### DIFF
--- a/php/libraries/NDB_Config.class.inc
+++ b/php/libraries/NDB_Config.class.inc
@@ -415,6 +415,20 @@ class NDB_Config
     }
 
     /**
+     * Helper function to determine whether a setting with a boolean value is
+     * enabled, i.e. equal to "true".
+     * Since settings are returned as strings, this must be done as a string
+     * comparison.
+     *
+     * @param string $name The name of the setting to retrieve. This setting
+     *                      should have a boolean value in the config file or DB.
+     * @return bool True if the setting is equal to "true".
+     */
+    function settingEnabled(string $name) {
+        return $this->getSetting($name) === 'true';
+    }
+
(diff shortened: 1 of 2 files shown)


php/libraries/NDB_Config.class.inc
```diff
@@ -414,6 +414,22 @@ class NDB_Config
...
+     * comparison.
+     *
+     * @param string $name The name of the setting to retrieve. This setting
+     *                      should have a boolean value in the config file or DB.
+     *
+     * @return bool True if the setting is equal to "true".
+     */
+    function settingEnabled(string $name): bool
+    {
+        return $this->getSetting($name) === 'true';
```
driusan:
```suggestion
        $val = $this->getSetting($name);
        return $val === 'true' || $val === '1';
```

If adding a helper, might as well make it robust. This should also be used somewhere and not just added for the sake of being added.

johnsaigle:
I'll add an example.

PR #4943, opened 2019-07-15
update php version in docker

kongtiaowang:

### DIFF
--- a/Dockerfile.test.php7
+++ b/Dockerfile.test.php7
@@ -1,7 +1,7 @@
 FROM php:7.2.2-fpm
 
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


Dockerfile.test.php7
```diff
@@ -1,7 +1,7 @@
-FROM php:7.2
+FROM php:7.2.2-fpm
 
 RUN apt-get update && \
-    apt-get install -y mysql-client
+    apt-get install -y mysql-client zlib1g-dev
```
johnsaigle: Why does bugfix have this extra line compared to minor and major?

driusan:
Replaced by #4955

PR #4944, opened 2019-07-15
update php version in docker

kongtiaowang:
Update PHP version from php7.2 to php7.2-fpm to pass Travis.


driusan:
Replaced by #4955

PR #4945, opened 2019-07-15
[Core] Update password expiry logic to no longer expire after 6 months

### LINKED ISSUE #4458: Remove 6 month password expiry from LORIS
### Background

Time-based password expiration is not considered a best practice for password management. A password with enough entropy doesn't "go bad". Also most people in practice end up making seasonal passwords or incrementing their old passwords when a time-based expiration is imposed.

We still would like the ability to reset passwords. Currently this is done by manually setting the `Password_expiry` column in the `candidate` table to a date some time in the past.

### TODO

* Change Password_expiry from a date type to a bool type (and perhaps the name should be something like `Password_expired`)
* Update calling code to use this new value properly

johnsaigle:
## Brief summary of changes
This PR removes the `Password_expiry` date column and adds a `PasswordExpired` boolean column.

Passwords will no longer be required to reset after 6 months as this is not actually a security benefit (see discussion in linked issue) and because most projects were subverting this by setting passwords to have an infinite expiry time anyway.

Calling code and relevant tables are updated.

1. Follow the Test plan at `modules/login/test/Reset_Password_Test_Plan.md`

* Resolves #4458

### DIFF
--- a/modules/login/php/passwordreset.class.inc
+++ b/modules/login/php/passwordreset.class.inc
@@ -71,7 +71,7 @@ class PasswordReset extends \NDB_Form
                   $password = \User::newPassword();
                   // reset the password in the database
                   // expire password so user must change it upon login
-                  $user->updatePassword($password, '1999-01-01');
+                  $user->updatePassword($password);
                   // send the user an email
                   $msg_data['study']    = $config->getSetting('title');
                   $msg_data['url']      = $config->getSetting('url');


PapillonMcGill:
@johnsaigle I believe that some files are missing!

johnsaigle:
hmmm yep probably something to do with rebasing. I'll look into it

johnsaigle:
I think I lost the work for this one 🤷‍♂ 😞

PR #4946, opened 2019-07-15
update php version in docker

kongtiaowang:
Update PHP version from php7.2 to php7.2-fpm to pass Travis.

### DIFF
--- a/Dockerfile.test.php7
+++ b/Dockerfile.test.php7
@@ -1,7 +1,7 @@
-FROM php:7.2
+FROM php:7.2.2-fpm
 
 RUN apt-get update && \
-    apt-get install -y mysql-client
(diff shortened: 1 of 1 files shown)


Dockerfile.test.php7
```diff
@@ -1,7 +1,7 @@
-FROM php:7.2
+FROM php:7.2.2-fpm
```
driusan: The description of this PR doesn't really describe what this is changing and why. Why do we need the -fpm variant of php? Why do we need such a specific version?

kongtiaowang: https://github.com/aces/Loris/pull/4944/files
7 lines, no different.

driusan:
Replaced by #4955

PR #4949, opened 2019-07-16
[Core] Add '200 OK' and '201 Created' responses to src/

johnsaigle:
## Brief summary of changes
Builds on the work of @xlecours's PR https://github.com/aces/Loris/pull/4275 to add the OK and Created responses with our src/ directory.

This will allow us to simplify the writing of PSR-compliant code.

### DIFF
--- a/src/Http/Response/Created.php
+++ b/src/Http/Response/Created.php
@@ -34,16 +34,20 @@ class Created extends JsonResponse
     /**
      * Create a Json response specific to 201 Created
      *
-     * @param string $location The created resource location
-     * @param array  $body     The response content
+     * @param array $body Data to convert to JSON.
+     * @param array $headers Array of headers to use at initialization.
      *
      * @return void
      */
-    public function __construct(string $location, array $body = [])
+    public function __construct(array $body = [], array $headers = [])
     {
-        $headers = array();
-        if (isset($location)) {
-            $headers['Location'] = $location;
+        if (empty($body)) {
+            if (!isset($headers['Location'])) {
+                throw new \LorisException(
+                    'Created response must return either a non-empty body or ' .
+                    'a Location header for redirection.'
+                );
+            }
         }
 
         parent::__construct($body, 201, $headers);
--- a/src/Http/Response/Created.php
+++ b/src/Http/Response/Created.php
@@ -32,11 +32,17 @@ use \LORIS\Http\Response\JsonResponse;
 class Created extends JsonResponse
 {
     /**
-     * Create a Json response containing the 200 OK Response
+     * Create a Json response specific to 201 Created
+     *
+     * @param array $body The response content
+     * @param string $location The created resource location
+     *
+     * @return void
      */
-    public function __construct()
+    public function __construct(array $body = [], string $location = '/')
     {
-        parent::__construct(new EmptyStream(), 201);
+        $headers = array('Location' => $location);
+        parent::__construct($body, 201, $headers);
     }
 }
 
--- a/src/Http/Response/Created.php
+++ b/src/Http/Response/Created.php
@@ -1,0 +1,42 @@
+<?php declare(strict_types=1);
+/**
+ * File contains the PSR7 ResponseInterface implementation for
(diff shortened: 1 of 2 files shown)


xlecours:
@johnsaigle It would be nice to add have a place in the code that uses those classes.  
P.S.: I do not need authorship on those files; but thanks anyways.

src/Http/Response/Created.php
```diff
@@ -0,0 +1,42 @@
...
+ * @category PSR15
+ * @package  Http
+ * @author   Xavier Lecours Boucher <xavier.lecours@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+class Created extends JsonResponse
+{
+    /**
+     * Create a Json response containing the 200 OK Response
```
PapillonMcGill:
```suggestion
     * Create a Json response containing the 201 Created Response
```

johnsaigle:
@xlecours I want to use them in https://github.com/aces/Loris/pull/4181 and will update that code when both are merged.

src/Http/Response/Created.php
```diff
@@ -0,0 +1,42 @@
...
+ * @author   Xavier Lecours Boucher <xavier.lecours@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+class Created extends JsonResponse
+{
+    /**
+     * Create a Json response containing the 201 Created Response
+     */
+    public function __construct()
```
xlecours:
```
/**
     * Create a Json response specific to 201 Created
     *
     * @param string $location The created ressource location
     */
    public function __construct(array $body = [], string $location)
    {
        $headers = array('Location' => $location);
        parent::__construct($body, 201, $headers);
    }
```
I think we should facilitate the addition of the `Location` header. Also, the body is not necessarily empty. 

reference : https://tools.ietf.org/html/rfc7231#section-6.3.2

src/Http/Response/OK.php
```diff
@@ -0,0 +1,45 @@
...
+ * @link     https://www.github.com/aces/Loris/
+ */
+class OK extends JsonResponse
+{
+    /**
+     * Create a Json response containing the 200 OK Response
+     *
+     * @param string $msg The response text
+     */
+    public function __construct(string $msg = 'OK')
```
xlecours:
```
     * @param array $data Data to convert to JSON.
     * @param array $headers Array of headers to use at initialization.
     */
    public function __construct(
        array $data,
        array $headers = []
    ) {
        parent::__construct($body, 200, $headers);
    }
```
Most 200 Response body will contain a body that is not just a message.

src/Http/Response/Created.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    /**
+     * Create a Json response specific to 201 Created
+     *
+     * @param array  $body     The response content
+     * @param string $location The created resource location
+     *
+     * @return void
+     */
+    public function __construct(array $body = [], string $location = '/')
```
xlecours: Please remove the default value of `$location`.
The location header must be used when the url of the new resource is different than the url used to create the resource. (e.g.: POST /candidates => 201 Created 'location: /candidates/$candid' )

xlecours: @johnsaigle I see that you need to put the location parameter first because I ask that the default value be removed but it creates a counter-intuitive usage (imo).
Why not simply stick with the same constructor than the OK with `array $body = [], array $headers = []` byut add a check to make sure that if the body is empty, that there is a location header?

driusan: I don't feel strongly about this one way or another, but I don't really find "location, body" counter-intuitive..

johnsaigle: @driusan It's not counter-intuitive in itself but all the other response classes have `body` first.

@xlecours I'll update the code with your suggestion.

PR #4963, opened 2019-07-17
[Testing] Update Phan to latest version

johnsaigle:
## Brief summary of changes

Our Phan version is aging. Since it's a static analysis dev tool, we should keep it up-to-date to benefit from bug fixes and new features.

In particular, this version of `phan` supports the `--redundant-condition-detection` which is unsupported by our current version.

I recommend that this PR gets merged into `minor` and `major` as well as bugfix ASAP so that the same static analysis conditions are run against every branch.

Takes over some functionality of #4433

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "phpunit/phpunit" : "6.5.*",
         "facebook/webdriver" : "dev-master",
         "phpmd/phpmd": "^2.6",
-        "phan/phan": "~2.2.5"
+        "phan/phan": ">2.2.4"
     },
     "scripts": {
       "pre-install-cmd": "mkdir -p project/libraries"
--- a/composer.json
+++ b/composer.json
@@ -18,8 +18,7 @@
         "phpunit/phpunit" : "6.5.*",
         "facebook/webdriver" : "dev-master",
         "phpmd/phpmd": "^2.6",
-        "phan/phan": "~2.2.5",
-        "microsoft/tolerant-php-parser": "~0.0.18"
+        "phan/phan": "~2.2.5"
     },
     "scripts": {
       "pre-install-cmd": "mkdir -p project/libraries"
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,8 @@
         "phpunit/phpunit" : "6.5.*",
         "facebook/webdriver" : "dev-master",
         "phpmd/phpmd": "^2.6",
-        "phan/phan": ">1.2.2"
+        "phan/phan": "~2.2.5",
+        "microsoft/tolerant-php-parser": "~0.0.18"
     },
     "scripts": {
       "pre-install-cmd": "mkdir -p project/libraries"
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ services:
 install:
 - sudo apt-get update
 - sudo apt-get install -y libapparmor1
-- pecl install -f ast-0.1.6
+- pecl install -f ast-1.0.1
 - sudo apt-get install npm
 - make dev
 
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "8498ac61c0fc01551b0ada55adb75620",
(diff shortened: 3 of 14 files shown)


composer.json
```diff
@@ -19,7 +19,8 @@
         "phpunit/dbunit": "3.0.*",
         "facebook/webdriver" : "dev-master",
         "phpmd/phpmd": "^2.6",
-        "phan/phan": ">1.2.2"
+        "phan/phan": "~2.2.5",
+        "microsoft/tolerant-php-parser": "~0.0.18"
```
johnsaigle: This is a require dependency of phan.

driusan: then it shouldn't be added here, dependencies are resolved recursively.

johnsaigle: I see what you're saying. I tried editing `composer.json` and then running `composer update phan/phan` so that other packages didn't update. I got a conflict with the microsoft package so I added this line here.

Is there a better way?

driusan: I just tried changing the phan version to `>2.2` locally and composer didn't have any conflicts.

```
➜  Loris git:(d17aa5d5d) ✗ composer update phan/phan
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 1 update, 0 removals
  - Updating phan/phan (1.3.1 => 2.2.3): Downloading (100%)         
Package phpunit/dbunit is abandoned, you should avoid using it. No replacement was suggested.
Package phpunit/phpunit-mock-objects is abandoned, you should avoid using it. No replacement was suggested.
Writing lock file
Generating autoload files
```

johnsaigle: It looks like you're installing version 2.2.3 where my update gets us to 2.2.5

Does the same thing happen if you install the newest version that way?

driusan: works fine so long as I update both dependencies:

```
$ composer update phan/phan microsoft/tolerant-php-parser
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 2 updates, 0 removals
  - Updating microsoft/tolerant-php-parser (v0.0.17 => v0.0.18): Downloading (100%)         
  - Updating phan/phan (2.2.3 => 2.2.5): Downloading (100%)         
Package phpunit/dbunit is abandoned, you should avoid using it. No replacement was suggested.
Package phpunit/phpunit-mock-objects is abandoned, you should avoid using it. No replacement was suggested.
Writing lock file
Generating autoload files
```

johnsaigle: OK I'll redo this

johnsaigle: Thanks for reviewing and testing so quickly.

composer.json
```diff
@@ -19,7 +19,7 @@
         "phpunit/dbunit": "3.0.*",
         "facebook/webdriver" : "dev-master",
         "phpmd/phpmd": "^2.6",
-        "phan/phan": ">1.2.2"
+        "phan/phan": "~2.2.5"
```
driusan: This is unnecessarily specific.. the composer.lock file pins the specific version for reproducibility (until the next time composer update is run).. I think it should stay a `>` operation.

johnsaigle:
```suggestion
        "phan/phan": ">2.2.4"
```

johnsaigle: Done

johnsaigle:
@driusan I've changed the base to minor. Don't we want it on all branches though?

driusan:
major: yes, it'll get pushed forward next time we push minor into major (and shouldn't be difficult because the branches haven't diverged from 21.0), bugfix, no, not yet, because we don't want it to block fixing bugs in the 21.0.1+ release.

johnsaigle:
Blocked by #4433 as phan requires a newer version of php-ast

driusan:
@johnsaigle can the 1 line in .travis.yml not just be updated here? That PR has conflicts anyways, and the other changes in it seem to be unrelated to what's needed for this one.

johnsaigle:
@driusan Done

composer.json
driusan: This seems to have gotten reverted somehow

racostas:
It seems that taking off lines 179 and 180 from **modules/new_profile/php/new_profile.class.inc**  makes appear the comboBox **Site** in **new_profile** module (when having only one site associate to the user) with no options to choose from. At the same time this is a mandatory field.

image

image

PR #4964, opened 2019-07-17
[Testing] Cleanup phan file and make explicit the settings we are using

johnsaigle:
## Brief summary of changes

* Explicitly declare severity level (unchanged from before)
* ~~Explicitly declare targeted version of PHP version~~ This could cause a problem with Travis
* Update link from the etsy repo to phan
* Remove a duplicate, conflicting setting
* ~~Remove backward compatibility check because we don't need it~~ Dave wants this to stay
* Change tabs to spaces

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -37,18 +37,18 @@ return [
     "suppress_issue_types" => [
         "PhanTypeInvalidDimOffset",
         "PhanTypePossiblyInvalidDimOffset",
-		"PhanUndeclaredMethod",
+        "PhanUndeclaredMethod",
         "PhanTypeMismatchDimFetch",
-		"PhanTypeMismatchReturn",
-		"PhanUndeclaredClassMethod",
-		"PhanTypeMismatchArgument",
-		"PhanTypeMismatchProperty",
+        "PhanTypeMismatchReturn",
+        "PhanUndeclaredClassMethod",
+        "PhanTypeMismatchArgument",
+        "PhanTypeMismatchProperty",
         "PhanTypeArraySuspiciousNullable",
-<<<<<<< HEAD
         "PhanPossiblyUndeclaredVariable",
         "PhanNoopNew",
-		"PhanNonClassMethodCall",
-        "PhanTypeSuspiciousStringExpression"
+        "PhanNonClassMethodCall",
+        "PhanUndeclaredVariableDim",
+        "PhanTypeSuspiciousStringExpression",
 	],
 	"analyzed_file_extensions" => ["php", "inc"],
 	"directory_list" => [
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -77,16 +77,6 @@ return [
         // because Phan disables xdebug by default.
         "xdebug"     => "vendor/phan/phan/.phan/internal_stubs/xdebug.phan_php",
     ],
-
-    // The line below is required to prevent PhanUndeclaredVariable problems in
-    // the bvl_feedback ajax scripts. They all require() a file in this directory
-    // that declares the $feedbackThread variable. Phan doesn't know about this.
-    //
-    // Toggling the below rule to true is the suggested fix by the phan devs.
-    // see https://github.com/phan/phan/issues/1650.
-    //
-    // When the module is refactored, this line should be deleted.
-    'ignore_undeclared_variables_in_global_scope' => true,
     'plugins' => [
         //'UnreachableCodePlugin',
     ]
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -14,7 +14,7 @@ return [
     // suspicious code.
     "minimum_severity" => 5,
     // TODO: Update this value to the minimum required by LORIS.
-    "target_php_version" => "7.2",
+    "target_php_version" => "7.3",
     // FIXME: allow_missing_properties should be false, but there's
     // too many other things to fix first.
     "allow_missing_properties" => true,
@@ -77,7 +77,7 @@ return [
         // because Phan disables xdebug by default.
         "xdebug"     => "vendor/phan/phan/.phan/internal_stubs/xdebug.phan_php",
     ],
-    'plugins' => [
-        //'UnreachableCodePlugin',
+    "plugins" => [
+        "UnreachableCodePlugin",
     ]
 ];
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -48,9 +48,6 @@ return [
         "PhanPossiblyUndeclaredVariable",
         "PhanNoopNew",
         "PhanNonClassMethodCall",
-        "PhanUndeclaredVariableDim",
-        "PhanTypeSuspiciousStringExpression",
-        "PhanNonClassMethodCall",
         "PhanTypeSuspiciousStringExpression"
     ],
     "analyzed_file_extensions" => ["php", "inc"],
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -30,6 +30,9 @@ return [
 	"ignore_undeclared_variables_in_global_scope" => true,
     // FIXME: We should add this.
     "dead_code_detection" => false,
+    // FIXME: We should add this. Note that dead_code_detection also covers
+    // unused_variable_detection. It might be a good idea to enable this first.
+    "unused_variable_detection" => false,
 	"suppress_issue_types" => [
         "PhanTypeInvalidDimOffset",
         "PhanTypePossiblyInvalidDimOffset",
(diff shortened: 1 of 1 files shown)


.phan/config.php
```diff
@@ -1,20 +1,33 @@
 <?php
 
 return [
-	"backward_compatibility_checks" => true,
```
johnsaigle: We're firmly on PHP 7 now so I don't think we need to check this.

driusan: I believe this enables backwards compatibility checks against target_php_version, not against PHP7.

.phan/config.php
```diff
@@ -1,20 +1,38 @@
...
+    // Toggling the below rule to true is the suggested fix by the phan devs.
+    // see https://github.com/phan/phan/issues/1650.
+    //
+    // When the module is refactored, this line should be deleted.
+	"ignore_undeclared_variables_in_global_scope" => true,
+    // FIXME: We should add this.
+    "dead_code_detection" => false,
+    // FIXME: We should add this. Note that dead_code_detection also covers
+    // unused_variable_detection. It might be a good idea to enable this first.
+    "unused_variable_detection" => false,
```
driusan: Are these finding things? phpmd already did/does these things, so we should have 0 results found.

johnsaigle: It's finding many things... I enabled just the `unused_variable_detection` rule and got 247 results after running phan.

PHPMD is also only configured to run on `php/libraries/`. 72 of the above results occurred in that folder so PHPMD is definitely missing something OR phan is giving a ton of false positives. Taking a look at the results though they seem accurate.

.phan/config.php
johnsaigle: The references I can find seem to specify PHP 5 --> PHP 7 changes. I don't think it has to do with target_php_version

https://github.com/phan/phan/wiki/Tutorial-for-Analyzing-a-Large-Sloppy-Code-Base#slowly-ramp-up-the-strength-of-the-analysis

https://github.com/phan/phan/wiki/Phan-Config-Settings#backward_compatibility_checks

.phan/config.php
johnsaigle: (Some of the above errors are already fixed by my PRs that add new Phan rules.)

.phan/config.php
PapillonMcGill: from the 2nd link:
```
Backwards Compatibility Checking. This is slow and expensive, but you should consider running it before upgrading your version of PHP to a new version that has backward compatibility breaks
```

Reading this, it is now specified for php 7 or 5.  We might have php8 in the future when thing break with php7.  Even if this possibility is low.  
But as @johnsaigle mentionned, all the code is done for php7 and the cost of this is high. WIll help with time to check.

driusan: Even if it's "high cost", high is at most on the order of seconds for phan, in a CI run that takes over 20 minutes to run the integration tests. I think we should still keep it.

xlecours:
@johnsaigle could you fix indentation?

johnsaigle:
what's wrong with the indentation?

xlecours:
It's a mix of tabs and spaces

PapillonMcGill:
@johnsaigle @xlecours ask you to replace tab with spaces

johnsaigle:
that's no good! will do.

johnsaigle:
I got pattern not found `^I` when running your command. Are you sure the formatting is bad? It may just be the way GitHub is displaying the diff.

.phan/config.php
```diff
@@ -1,63 +1,79 @@
...
+        // This is excluded because the API is in the process of being converted
+        // to a proper LORIS module. Everything in this folder will be deleted
+        // so we're not going to spend time cleaning it up.
+        "htdocs/api/",
+    ],
+    "autoload_internal_extension_signatures" => [
         // Xdebug stubs are bundled with Phan 0.10.1+/0.8.9+ for usage,
         // because Phan disables xdebug by default.
-        'xdebug'     => 'vendor/phan/phan/.phan/internal_stubs/xdebug.phan_php',
+        "xdebug"     => "vendor/phan/phan/.phan/internal_stubs/xdebug.phan_php",
```
PapillonMcGill:
```suggestion
        "xdebug" => "vendor/phan/phan/.phan/internal_stubs/xdebug.phan_php",
```

driusan:
`^I` means ctrl-I (or tab), not caret I. You may need to use `ctrl-m ctrl-I`

johnsaigle:
Yep I know it means Ctrl, I needed to be in a specific Vim mode. Xavier and I cleared it up via Slack.

The file is fixed now.

xlecours:
@johnsaigle the file is still all messed up.

johnsaigle:
@xlecours Hmmm I thought I had fixed it. How is it now?

xlecours:
> @xlecours Hmmm I thought I had fixed it. How is it now?

Almost there. Line#4 is still offending.

.phan/config.php
```diff
@@ -1,54 +1,25 @@
...
-		"vendor",
-		"htdocs/api/",
-	],
-    'autoload_internal_extension_signatures' => [
-        // Xdebug stubs are bundled with Phan 0.10.1+/0.8.9+ for usage,
-        // because Phan disables xdebug by default.
-        'xdebug'     => 'vendor/phan/phan/.phan/internal_stubs/xdebug.phan_php',
-    ],
-
+"backward_compatibility_checks" => true,
```
xlecours:
```suggestion
    "backward_compatibility_checks" => true,
```

johnsaigle:
@xlecours thanks!

xlecours:
@johnsaigle Still need to be rebased.

johnsaigle:
😭

johnsaigle:
I'll fix it

johnsaigle:
@xlecours Let me know if my fix works.

xlecours:
> > @xlecours Hmmm I thought I had fixed it. How is it now?
> 
> Almost there. Line#4 is still offending.

bis

johnsaigle:
How's that?

johnsaigle:
Many of the Travis errors are caused by phan analyzing the `htdocs/api` directory, which is deleted on 22.0-release. When that branch is merged into `master` this shouldn't be a problem.

.phan/config.php
```diff
@@ -1,70 +1,87 @@
...
+    // This is false on the assumption that that's a typo.
+    // It doesn't seem to have any effect on LORIS's codebase anyways.
+    "quick_mode" => false,
+    "analyze_signature_compatibility" => true,
+    // The default severity level for phan is 5. After removing all the
+    // suppressed rules, we should consider reducing this value to detect more
+    // suspicious code.
+    "minimum_severity" => 5,
+    // TODO: Update this value to the minimum required by LORIS.
+    "target_php_version" => "7.3",
```
driusan: This should not be explicitly set, or Travis won't be testing the right version.

.phan/config.php
```diff
@@ -1,70 +1,87 @@
...
         "PhanTypeArraySuspiciousNullable",
-        "PhanNoopNew"
-	],
-	"analyzed_file_extensions" => ["php", "inc"],
-	"directory_list" => [
-		/* This doesn't include php/installer, because there's
-		   (intentionally) classes in the installer namespace
+        "PhanNoopNew",
+        "PhanNonClassMethodCall",
+        "PhanTypeSuspiciousStringExpression"
```
driusan: You seem to have unnecessarily added some new exception types, probably from a bad rebase..

.phan/config.php
```diff
@@ -1,70 +1,87 @@
...
+    ],
+    "analyzed_file_extensions" => ["php", "inc"],
+    "directory_list" => [
+        // This doesn't include php/installer, because there's
+        //  (intentionally) classes in the installer namespace
```
driusan: Unless I'm misreading this diff these lines are now duplicated

ridz1208:
@driusan are all your comments addressed ?

@johnsaigle conflict

driusan:
No

johnsaigle:
@driusan I think I've now incorporated all of your suggestions

.phan/config.php
```diff
@@ -1,66 +1,74 @@
...
         "PhanTypePossiblyInvalidDimOffset",
-		"PhanUndeclaredMethod",
+        "PhanUndeclaredMethod",
         "PhanTypeMismatchDimFetch",
-		"PhanTypeMismatchReturn",
-		"PhanUndeclaredClassMethod",
-		"PhanTypeMismatchArgument",
-		"PhanTypeMismatchProperty",
+        "PhanTypeMismatchReturn",
+        "PhanUndeclaredVariableDim",
```
driusan: When I look at this with `?w=1`, this seems to be accidentally sneaking PhanUndeclaredVariableDim into the list of ignores.

PR #4965, opened 2019-07-17
[Testing] Add a progress bar when phan is run from Makefile

johnsaigle:
Screen Shot 2019-07-17 at 12 45 38

Adds a visual indicator of phan's progress when running `make checkstatic`.


driusan:
This violates best practices, is counter to the UNIX philosophy, noisy, and makes the code less robust for scripts. If you want a progress bar, run phan manually.

PR #4966, opened 2019-07-17
[Doc] Update modules/README.md

driusan:
This updates the document explaining the structure of a directory in the
modules/ directory, both to provide more detail and remove outdated
information such as telling people to use the ajax/ directory.

### DIFF
--- a/modules/README.md
+++ b/modules/README.md
@@ -4,6 +4,9 @@ Modules should be entirely self-contained into a subdirectory with the following
 
     ModuleName/
 
+        README.md     - This file should contain technical specifications for any module in LORIS.
+                        See any module (ie. modules/candidate_list/README.md) for a template.
+
         css/          - Any necessary new CSS classes/files should go here
 
         help/         - Contains a markdown file of help content for each page in this module.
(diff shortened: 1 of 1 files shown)


modules/README.md
```diff
@@ -4,20 +4,45 @@ Modules should be entirely self-contained into a subdirectory with the following
 
     ModuleName/
 
-        ajax/         - This directory contains any AJAX helper script used by the module. They should
-                        be invoked by using the AjaxHelper.php script in Loris
-
         css/          - Any necessary new CSS classes/files should go here
 
-        help/         - Contains a markdown file of help content for each page in this module
+        help/         - Contains a markdown file of help content for each page in this module.
```
johnsaigle: It might be good to specify that quirk about the help texts... I think it's that if the help file exists in the DB then the one from the help editor won't be shown? Or it could be the other way around.

modules/README.md
```diff
@@ -4,20 +4,45 @@ Modules should be entirely self-contained into a subdirectory with the following
 
     ModuleName/
 
-        ajax/         - This directory contains any AJAX helper script used by the module. They should
```
johnsaigle: It might be good to keep this with a deprecated warning, similar to what you added for the `templates/` folder below. Many module still rely on this directory.

driusan: No one reading the document who doesn't know what it's already doing there should be pointed to it, though. The templates directory is a bit of a "softer" deprecation since it's not holding us back as much while the ajax directory adds an incredible amount of technical debt. Regardless, I added something to mention that it exists and is deprecated.

modules/README.md
driusan: ok

modules/README.md
```diff
@@ -4,20 +4,52 @@ Modules should be entirely self-contained into a subdirectory with the following
...
-                        be invoked by using the AjaxHelper.php script in Loris
+        README.md     - This file should contain technical specifications for any module in LORIS.
+                        See any module (ie. modules/candidate_list/README.md) for a template.
```
PapillonMcGill: I would have preferred a moc template that to refer to 'any'.  Do we have a guarantee that 'all' README.md file in 'all' module conform?

driusan: If they don't they should be updated

Jkat:
Late but LGTM - looked* good to me 👍

PR #4967, opened 2019-07-17
[Core] Remove unreachable code in LORIS using phan

johnsaigle:
## Brief summary of changes

* Adds a phan plugin to create a rule to detect unreachable code
* Removes unreachable code


modules/document_repository/php/files.class.inc
```diff
@@ -88,7 +88,6 @@ class Files extends \NDB_Page
                     )
                 );
             }
-            break;
```
driusan: These ones are scary.. do we have any other static analysis rules to warn about unexpected fallthroughs if the if statements change?

johnsaigle: Not that I know of... I think it'd be tricky because PHP doesn't make you do `break` at the end and we take advantage of empty fallthrough cases quite a bit, like in LINST parsing.

I could add a suppression for the unreachable code rule here, or we could decide not to enable the setting. What do you prefer?

johnsaigle:
```suggestion
             //NOTE: If the above exception is removed the break below MUST be added back.
             //break;
```
Or something like this?

driusan: I think this is one of the very few cases where I'd be fine with a suppression and a comment explaining why.

Another possibility that would make it less scary is to remove the elses, so that they're of the form

```
if foo {
       return $x;
}
return $y;
```

and there's always a return the line above the case, so that it's obvious that all code paths don't fall through regardless of changes to the above logic.

modules/genomic_browser/test/genomic_browserTest.php
```diff
@@ -163,58 +163,58 @@ function testGenomicBrowserFilterEachTab()
         $this->markTestSkipped("Skipping long test");
         return;
```
PapillonMcGill: Could also just remove the `return;` as code past `markTestSkipped()` is not executed.
Well, maybe Travis would flag it anyway.

johnsaigle:
@driusan Can you re-review when you have a minute?

modules/server_processes_manager/test/server_processes_managerTest.php
```diff
@@ -134,14 +134,19 @@ function testFilters()
...
+        //this->_testFilter("/server_processes_manager/", "pid", "317", "1 rows");
+        //this->_testFilter(
+        //"/server_processes_manager/",
+        //"type",
+        //"mri_upload",
+        //"51"
+        //);
+        //this->_testFilter(
+        //   "/server_processes_manager/",
+        //   "userid",
```
driusan: Why is this code all commented out? If it's deleted, it should be deleted (same with the other test above)

johnsaigle: Works for me, just thought it might have been left in as a reference.

PR #4969, opened 2019-07-17
[tools/raisinbread exporter] Use config.xml credentials

ridz1208:
## Brief summary of changes

The previous iteration of the script was dependent on the existence of a mysql configuration file (`.my.cnf` for example). this instead makes it depend on the quat values in the configuration file

### DIFF
--- a/tools/exporters/DB_dump_table_data.php
+++ b/tools/exporters/DB_dump_table_data.php
@@ -35,6 +35,18 @@ $tableNames = $DB->pselectCol("
                       WHERE TABLE_SCHEMA =:dbn",
     array("dbn"=>$databaseInfo['database'])
 );
+
+$dbUser = $databaseInfo["quatUser"];
+$dbPassword = $databaseInfo["quatPassword"];
+$dbHost = $databaseInfo["host"];
+
+if (empty($dbUser) || empty($dbPassword) || empty($dbHost)) {
+    echo "\n\n Some database credentials are missing, please make sure administrator 
+    credentials (quatUser, quatPassword) and a host value are available in your 
+    configuration file. \n\n";
+    die();
+}
+
 /*
  * Definitions of the flags used in the command below (ORDERING is IMPORTANT):
  * --complete-insert -> Use complete INSERT statements that include column names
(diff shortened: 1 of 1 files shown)


tools/exporters/DB_dump_table_data.php
```diff
@@ -35,6 +35,18 @@
                       WHERE TABLE_SCHEMA =:dbn",
     array("dbn"=>$databaseInfo['database'])
 );
+
+$dbUser = $databaseInfo["quatUser"];
+$dbPassword = $databaseInfo["quatPassword"];
+$dbHost = $databaseInfo["host"];
+
+if (empty($dbUser) || empty($dbPassword) || empty($dbHost)) {
+    echo "\n\n Some database credentials are missing, please make sure administrator
```
PapillonMcGill:
```suggestion
    echo "\n\n Some database credentials are missing, please ensure administrator 
```

ridz1208:
@cmadjar when you tested did you try removing your `.my.cnf` file to make sure the command is not defaulting to the values in that file ?

cmadjar:
@ridz1208 I don't remember which DB I used to test it. I just retested it and used a different DB than the one using `.my.cnf` and it dumped the correct DB (not the one using `.my.cnf`).

I also get the following warning from mysqldump which tells me it is using the config.xml DB config and not the `my.cnf`:
```
mysqldump: [Warning] Using a password on the command line interface can be insecure.
```

ridz1208:
yeah thats because its not propmting for password.

@driusan you think it's okai ?

ridz1208:
@driusan your review would be great here

tools/exporters/DB_dump_table_data.php
```diff
@@ -54,11 +66,13 @@
  *                      values to accommodate for timezone differences.
  */
 
+
 // Loop through all tables to generate insert statements for each.
 foreach ($tableNames as $tableName) {
     $paths = \NDB_Config::singleton()->getSetting('paths');
     $filename = $paths['base'] . "/raisinbread/RB_files/RB_$tableName.sql";
-    exec('mysqldump '.$databaseInfo['database'].' '.
+    exec('mysqldump -u '.$dbUser.' -p'.$dbPassword.' -h '.$dbHost.' '.
```
driusan: These variables need to be shell escaped. You can use https://www.php.net/manual/en/function.escapeshellarg.php

Regarding the warning: MySQL flags it as a warning because anyone running "ps" on the host will be able to see the password. To properly fix it securely, you'd probably have to use proc_open instead of exec so you can write to the user input side of the pipe from php, but that would make things significantly more complicated.

ridz1208: changes done

PR #4973, opened 2019-07-22
[Tools] Add script for changing config settings

driusan:
This adds a new script to change the config settings
in the database without SQL access. This may be is useful
when the Config module page is unavailable due to bad
paths or otherwise when providing user support, where
it can be easier to just tell the user what to type
copy/paste in a terminal than describe where to find
a setting on the frontend.

1.

*

### DIFF
--- a/tools/setconfig.php
+++ b/tools/setconfig.php
@@ -36,12 +36,21 @@ if (count($args) < MIN_NUMBER_OF_ARGS) {
 $setting = $args[1];
 $value   = $args[2];
 
-$id= $DB->pselectOne(
-    "SELECT ID FROM ConfigSettings WHERE Name=:config",
+$config = $DB->pselectRow(
+    "SELECT ID, AllowMultiple FROM ConfigSettings WHERE Name=:config",
     array('config' => $setting)
 );
-if (!is_numeric($id) || $id <= 0) {
+
+if (count($config) === 0) {
     die("Invalid config name");
+}
+
+$multiple = $config['AllowMultiple'];
+if ($multiple == '1') {
+    die(
+        "Script only valid on single value settings. " .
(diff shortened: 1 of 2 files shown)


tools/setconfig.php
```diff
@@ -0,0 +1,57 @@
...
+
+$id= $DB->pselectOne(
+    "SELECT ID FROM ConfigSettings WHERE Name=:config",
+    array('config' => $setting)
+);
+if (!is_numeric($id) || $id <= 0) {
+    die("Invalid config name");
+}
+
+// Determine whether to insert or update
```
ridz1208: you also have to check if its a config of type "multiple"

driusan:
@ridz1208 it's not clear what the behaviour should be for "multiple" settings (add? update? delete?) so I just made it die with an error message.. as far as I know none of the config settings that make the frontend inaccessible are multiple settings.

tools/setconfig.php
johnsaigle: I think this is addressed now?

tools/setconfig.php
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/php
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

johnsaigle:
Also a small thing but the script gives this error when an invalid config setting is supplied:

php tools/setconfig.php notReal true

> PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /var/www/loris/tools/setconfig.php on line 44

I think the DB function returns `null` not `array()` when no value is found so `count()` throws a warning.

driusan:
@johnsaigle thanks, I missed that while refactoring the select for checking the multiple setting. I've also fixed the warning.

PR #4974, opened 2019-07-22
[Database] Remove unused function

### LINKED ISSUE #4972: Database->getAffected()  returns 0
While trying to write a script that uses it, I found that calling $DB->getAffected() after an update statement always returned 0, even when there were rows that were updated.

We should either add tests to ensure that it's working or remove the function, since at the moment the results are not correct and it can't be relied upon.

driusan:
The getAffected() function was unused. Upon trying
to use it, I discovered that it was also always
returning "0" after an update regardless of how many
rows were actually updated. This removes the function
and related code so that no one accidentally tries to use
a function which silently returns an incorrect result.

Resolves #4972.

1.

*

### DIFF
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -45,13 +45,6 @@ class Database
     var $_HistoryPDO = null;
 
     /**
-     * Integer number of rows affected by the last replace/update/delete/insert
-     *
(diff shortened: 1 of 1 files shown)


driusan:
@xlecours done

PR #4976, opened 2019-07-23
[Timepoint/Create Timepoint] Add timepoint language feature

jesscall:
## Brief summary of changes
This PR adds language feature to timepoints. 

1. Run patches in this PR + make sure you have changes that were merged yesterday from #4492
2. Create a new timepoint and choose a language.
3. You can add french test names  to `test_names_fr` so the instrument list is translated according to the language of the session.


SQL/0000-00-00-schema.sql
```diff
@@ -195,6 +197,7 @@ CREATE TABLE `session` (
   `MRIQCFirstChangeTime` datetime DEFAULT NULL,
   `MRIQCLastChangeTime` datetime DEFAULT NULL,
   `MRICaveat` enum('true','false') NOT NULL DEFAULT 'false',
+  `Language` enum('EN', 'FR') NOT NULL DEFAULT 'EN',
```
xlecours: This should be `language_code` referencing a `language` table row.

SQL/0000-00-00-schema.sql
```diff
@@ -237,6 +240,16 @@ CREATE TABLE `test_names` (
   CONSTRAINT `FK_test_names_1` FOREIGN KEY (`Sub_group`) REFERENCES `test_subgroups` (`ID`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
+CREATE TABLE `test_names_fr` (
```
xlecours: this could be `test_name_dictionary` if `Full_name_fr` was replace by `alternative_name` and a `language_code` column was added referencing a `language` table row.
It could support any languages.

johnsaigle:
```
FILE: /home/travis/build/aces/Loris/php/libraries/TimePoint.class.inc
 188 | ERROR | Doc comment for parameter "$lang" missing
```

PHPCS issue

php/libraries/TimePoint.class.inc
```diff
@@ -194,6 +194,7 @@ class TimePoint
      *                                    being created. If null, consecutive V1, V2,
      *                                    etc will be created.
      * @param \Site       $psc            The center of the timepoint
+     * @param string      $lang           The language code
```
xlecours: `string` -> `?string`

SQL/New_patches/2019_07_05_Add_Language_feature.sql
```diff
@@ -1,12 +1,14 @@
-CREATE TABLE `test_names_fr` (
+CREATE TABLE `test_names_dictionary` (
   `ID` int(11) NOT NULL AUTO_INCREMENT,
   `Test_name` varchar(255) DEFAULT NULL,
-  `Full_name_fr` varchar(255) NOT NULL,
+  `Full_name_alternative` varchar(255) NOT NULL,
+  `language_code` varchar(255) NOT NULL,
```
driusan: I think this should be a foreign key to a languages table, not a varchar, so that there's still an enumeration of supported languages by the loris instance (especially for things like dropdowns where you need to select the language.)

xlecours: it is a foreign key to the language table.

driusan: oops, missed that..

php/libraries/TimePoint.class.inc
ridz1208: are you suggesting that it should be optional?

I think not null is a better idea. we can get the front end to default to a language when only one is available but this part I think should require a lang

xlecours: It was not a suggestion. The function signature has `?string $lang` so the doc comment should be the same.
It could be `string $lang` but the function signature would need to be changed.

ridz1208: oh okai.

@jesscall had you considered making lang not optional ?

jesscall: oh oops, I'll change the function signature. I think lang should be required.

ridz1208: I think so too, make sure that all calls to this function are including a language

driusan: You could also make it non-nullable but provide a default value for the argument.

modules/instrument_list/php/instrument_list.class.inc
```diff
@@ -214,9 +214,28 @@ class Instrument_List extends \NDB_Menu_Filter
...
+                    )
+                );
+                $fullName    = (!empty($altName)) ? $altName
+                    : $instrument['Full_name'];
+
                 $Ins = "instruments";
 
-                $ins = $instrument['Full_name'];
+                $ins = $fullName;
                 $this->tpl_data[$Ins][$x][$i]['fullName'] = $ins;
```
maltheism: Just a suggestion buy why not just have line 239 be:
`$this->tpl_data[$Ins][$x][$i]['fullName'] = $fullName;`
There are a few other lines in this class similar and where it seems like assigning a variable to $ins is pointless because the next few lines has it assigned to something else.

johnsaigle: It may have to do with our max line length of 80 chars... just my suspicion.

maltheism: @johnsaigle I didn't think of that and I guess then it will still be two lines.

maltheism:
@jesscall I was assigned to review this PR. I'm just leaving a comment to get a notification when the conflicts are resolved.

php/libraries/TimePoint.class.inc
```diff
@@ -208,7 +209,8 @@ class TimePoint
         int $SubprojectID,
         ?string $nextVisitLabel,
         \Site $psc,
-        \Project $project
+        \Project $project,
+        int $lang = 1
```
xlecours: This will break https://github.com/aces/Loris/blob/master/modules/api/php/endpoints/candidate/visit/visit.class.inc#L273

jesscall: @xlecours even with the default?

xlecours: @jesscall good point. resolved :)

ridz1208: Nothing garantees you that language id `1` exists !!

xlecours: But en-CA will. 


> 
> I specially like the fact that you do not use `language_id` which is meaningless anyway.

jesscall: So the reason I went away from using language_code is as follows:
1) The utility getLanguageList returns an array with `[language_id] => language_label`
2) When preferred language was implemented for users, language ID is also used. 

It just seems we should keep that convention going?

ridz1208: kill IDs

jesscall:
Just having some trouble with integration tests... travis is saying it cannot find element psc.
Maybe someone with more experience has some insight on how to fix this?

ridz1208:
@johnsaigle is my review even meaningful if @xlecours already approved it??

i think not :bowing_man:

SQL/0000-00-00-schema.sql
```diff
@@ -67,7 +67,9 @@ CREATE TABLE `language` (
   UNIQUE KEY (`language_code`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
-INSERT INTO language (language_code, language_label) VALUES ('en-CA', 'English');
+INSERT INTO language (language_code, language_label) VALUES
+    ('en-CA', 'English'),
```
driusan: It probably makes sense to only provide `en-US` or `en-CA` as the default language and not `fr-CA`, since we don't support French (canadian or other variant) anywhere in LORIS

SQL/New_patches/2019_07_05_Add_Language_feature.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE `session`
+    ADD COLUMN `languageID` integer unsigned DEFAULT NULL,
+    ADD CONSTRAINT `FK_session_4` FOREIGN KEY (`languageID`) REFERENCES `language` (`language_id`);
+
+INSERT INTO language (language_code, language_label) VALUES
+    ('fr-CA', 'French');
```
driusan: The patch doesn't match the default schema change.

modules/create_timepoint/php/create_timepoint.class.inc
```diff
@@ -202,10 +203,23 @@ class Create_Timepoint extends \NDB_Form
...
+                $this->addSelect(
+                    'languageID',
+                    'Language',
+                    $languages
+                );
+                $this->addRule(
+                    'languageID',
+                    'Language is required',
+                    'required'
+                );
```
driusan: I don't think it makes sense to do this unless `count($languages)` is `> 1`.. it doesn't make sense to require language be a required user input on unilingual projects.

ridz1208: I think this ties back to https://github.com/aces/Loris/issues/5380

ridz1208: if that is feasible then we can just leave it as it is

driusan: I don't think sites and languages are comparable in that regard. Languages are a server configuration that aren't tied to the data, not having them defined should be equivalent to be a server configuration saying "this project doesn't use languages" and is just confusing to end users.. sites are variable-number-of-associations within a single loris instance.

SQL/New_patches/2019_07_05_Add_Language_feature.sql
jesscall: en-CA is already part of the schema and presumably already in the db - is that wrong? Shoudl en-CA appear in the patch as well?

driusan: Oh, I see. I misread the default schema change. I don't see why this is adding fr-CA.

jesscall: should i remove it>

modules/create_timepoint/php/create_timepoint.class.inc
ridz1208: same applies to site, a loris instance with a single site shouldnt worry about seeing the site... it's a given that any data entered is at that site...

language in this context is tied to the data !! it's the language of collection its a parameter like any other

SQL/New_patches/2019_07_05_Add_Language_feature.sql
ridz1208: yes

modules/create_timepoint/php/create_timepoint.class.inc
driusan: If you mean "a project with only 1 psc in the table" I agree, the ticket you referenced is about dealing with users that have multiple vs 1 site the same in a study that has multiple sites.

jesscall: I leave language as required but make it default to the study language if count = 1?

driusan: No, language should not be required or displayed if there is only 1 language configured.

xlecours: I agree with @driusan 
We should always aim at minimizing user clicks count. If there is only one language, the frontend should hide the select element. Then, having it optional or required is a matter of: do we want to rely on the default value of the column in the database.
If the field is optional, createTimepoint should default to the column default.

ridz1208: @driusan I mean absolutely any drop-down with a single option. regardless of what it is...

I don't like hiding stuff for several reasons:
1. when things are hidden they don't get tested during manual testing rounds because testers omit or forget to change the necessary configurations to enable/disable a hidden field, as demonstrated with the latest bug that Tomas caught (#6008) and so many before 
2. It's also confusing for users, and user manuals where we constantly have to mention "if you have more than one site affiliation, make sure to choose a site..... If you have a multi lingual study make sure to choose a language" and even at that we get "bug reports" with title, CANNOT CHOOSE SITE OF....
3. it's inconsistent across modules and it's inconsistent with instruments
4. makes the code ugly and bug prone weather it's in the php code, the templates or in the react.

Solution, either implementing #5380 ! I think it's the concern of the view to manage "clicks" I think it should just auto select the first element when there is only one AND it is required because there is also the case where a selection is not required and then you have to display the field anyways

driusan: Can you name a single site anywhere on the internet that forces you to choose your language with only 1 option when submitting a form?

ridz1208: yes

Screenshot from 2020-02-06 10-50-45

only master branch is available, it's preselected BUT THEY DON'T HIDE THE DROPDOWN !!

driusan: since when are branches languages?

ridz1208: @driusan 
22b79a640bc5089497a6262901b7274c

driusan: It's not a question of "should we display dropdowns with 1 option". That's already resolved: we display the same form to the users on the LORIS instance regardless of multi-site affiliations in order to have all users have the same experience and avoid confusion. It's a question of "should we display dropdowns with 1 option *when there is no possibility of any users on that LORIS instance ever having multiple options in that dropdown*".

In your GitHub screenshot, someone can create a branch on that repo tomorrow (in fact, if they wanted to do it through the GitHub UI they'd have to go through that dropdown) and there are other repos on GitHub which have multiple branches.

ridz1208: and loris admins can add a language too... anyways, this is going too far. I made my point

driusan: An admin adding a language would be analogous to an admin going to the configuration module and toggling the "useFeatureX"  config setting. If that were to happen, the dropdown should appear for all users.

jesscall: Okay so both options are valid. 

I think there isn't any harm in having the dropdown visible for everyone, with a default selected when only one language is available. It gives users a unified experience and also signals that there is the possibility of adding more languages (in case users or admins were not aware). 

Another reason why I am for having it displayed is because currently the integration testing we have does not cover the case where there are multiple languages (nor the case of multiple sites / projects, for that matter). This leads to the neglect of this feature during testing since the dropdown would never be displayed when using current instances of raisinbread. 

@driusan what do you think?

driusan: No, if there is only 1 language configured for the LORIS instance there should not be a dropdown displayed.

jesscall: ok, implemented. Please review

driusan:
@jesscall this has raisinbread conflicts

jesscall:
@driusan conflicts resolved

ridz1208:
🎉

PR #4977, opened 2019-07-23
[Data Dict] Update Data Dictionary module to use Data Framework

driusan:
This updates the datadict module to use the data framework, using
the DataFrameworkMenu class infrastructure that was put in place
in #4918.

This necessitated updating the datadict from an older style of
our React components (where the QuickForm was serialized to
javascript server side) to the a newer form (where most of
the display concerns are handled from React, except for the
dynamic parts which must be sent by implementing getFieldOptions).

Unfortunately, the current React side filtering doesn't support
"Keyword" search in the same manner that was used, so it's
replaced by more consistent/traditional filters for now.

1.

*

### DIFF
--- a/modules/datadict/jsx/dataDictIndex.js
+++ b/modules/datadict/jsx/dataDictIndex.js
@@ -115,10 +115,10 @@ class DataDictIndex extends Component {
             },
         },
         {
-            label: 'Field',
+            label: 'Source Field',
             show: true,
             filter: {
-                name: 'Field',
+                name: 'Source Field',
                 type: 'text',
             },
         },
--- a/modules/datadict/jsx/dataDictIndex.js
+++ b/modules/datadict/jsx/dataDictIndex.js
@@ -77,7 +77,7 @@ class DataDictIndex extends Component {
         <td
           contentEditable="true"
           className="description"
-          onBlur={updateDict(rowData[1])}>
+          onBlur={updateDict(rowData.Name)}>
             {cell}
         </td>
       );
--- a/modules/datadict/jsx/dataDictIndex.js
+++ b/modules/datadict/jsx/dataDictIndex.js
@@ -1,5 +1,7 @@
-import FilterForm from 'FilterForm';
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
 import Loader from 'Loader';
+import FilterableDataTable from 'FilterableDataTable';
 
 /**
  * Data Dictionary Page.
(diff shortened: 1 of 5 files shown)


johnsaigle:
Travis is having some issues with the integration tests.

modules/datadict/php/datadict.class.inc
```diff
@@ -43,148 +40,37 @@ class Datadict extends \NDB_Menu_Filter
...
-                    $query .= " AND pto.name IS NOT NULL";
-                } elseif ($val=='unmodified') {
-                    $query .= " AND pto.name IS NULL";
-                } else {
-                    $query .= " AND $key LIKE '$val%' ";
-                }
-            }
-        }
-        return $query;
+        return array('sourceFrom' => $instruments);
```
johnsaigle: Could simplify this to a one-line if you wanted. No need to declare a variable.

driusan:
Just follow the test plan, nothing should change other than the keyword filter being replaced by normal filters

kongtiaowang:
PHP Fatal error:  Uncaught NotFound: Could not load module `datadict`: file `/var/www/loris/modules/datadict/php/datadictrowprovisioner.class.inc` does not exist in /var/www/loris/php/libraries/Module.class.inc:89

driusan:
@kongtiaowang oops, thanks.. I forgot to commit a couple files.

modules/datadict/test/datadictTest.php
```diff
@@ -88,7 +88,7 @@ function testDatadictDoespageLoad()
```
kongtiaowang: line 39 :'Source Field' => 'Field' 
The front-end UI have changed.

modules/datadict/jsx/dataDictIndex.js
```diff
@@ -14,71 +16,38 @@ import Loader from 'Loader';
...
-      method: 'GET',
-      dataType: 'json',
-      success: (data) => {
-        this.setState({
-          data: data,
-          isLoaded: true,
+    return fetch(this.props.dataURL, {credentials: 'same-origin'})
+        .then((resp) => resp.json())
+        .then((data) => this.setState({data}))
+        .catch((error) => {
```
xlecours: On a side note: This will catch error thrown by fetch, not HTTP 4xx/5xx. But conveniently, since `DataFrameworkMenu` extends `NDB_Page`, when an error happen (4xx/5xx) an html page is return and the `resp.json()` throws a TypeError because the html can't be parsed.

driusan: That doesn't sound like a reliable side-effect to depend on.

driusan:
@cmadjar I believe your problem is fixed now (plus another one that came up after rebasing)

PR #4978, opened 2019-07-23
[Design] Add interface for User objects

driusan:
This adds an interface to describe a User and essential
properties that something which implements User must have.

I started with the User class, removed anything related
to password/permissions (which is a different concern from
describing a User), and removed the functions that were
variations of getSites(), then added Active and Pending
(which are in the User Accounts menu filter.) This is
what remained.

Note that the User class doesn't currently implement this
interface.

1.

*

### DIFF
--- a/php/libraries/IUser.class.inc
+++ b/php/libraries/IUser.class.inc
@@ -53,12 +53,16 @@ interface IUser
      * Returns true if the user is Active (ie. has been
      * approved and was not explicitly deactivated for
      * other reasons.)
+     *
+     * @return bool
      */
     public function isActive() : bool;
 
     /**
      * Returns true if the user has requested an account
      * which has not yet been reviewed/approved.
+     *
+     * @return bool
      */
     public function pendingApproval() : bool;
 }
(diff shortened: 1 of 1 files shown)


johnsaigle:
Looks good to me. 👍 I agree that permissions and passwords shouldn't be tightly coupled to the rest of the class. 

Excited to move forward with things like this.

php/libraries/IUser.class.inc
```diff
@@ -0,0 +1,65 @@
...
+     * Returns an array of sites a user belongs to.
+     *
+     * @return Site[]
+     */
+    public function getSites(): array;
+
+    /**
+     * Returns true if the user is Active (ie. has been
+     * approved and was not explicitly deactivated for
+     * other reasons.)
```
johnsaigle:
```suggestion
     * other reasons.)
     *
     * @return bool
```

php/libraries/IUser.class.inc
```diff
@@ -0,0 +1,65 @@
...
+     */
+    public function isActive() : bool;
+
+    /**
+     * Returns true if the user has requested an account
+     * which has not yet been reviewed/approved.
```
johnsaigle:
```suggestion
     * which has not yet been reviewed/approved.
     *
     * @return bool
```

johnsaigle:
@driusan Can this be re-opened and merged?

PR #4979, opened 2019-07-23
[Test] unittests for User.class.inc

AlexandraLivadas:
## Brief summary of changes

Unit tests for the [User.class.inc](https://github.com/aces/Loris/blob/major/php/libraries/User.class.inc) class.

Run `npm run tests:unit -- --filter User` or check out Travis!


php/libraries/User.class.inc
```diff
@@ -495,6 +495,9 @@ class User extends UserPermissions
 
     /**
      * Check if user belongs to DCC
+     * FIXME 'CenterID' is no longer used.
+     *       This should check that the 'CenterIDs' array cointains a 1
```
johnsaigle: Why not implement this? 😄

test/unittests/UserTest.php
```diff
@@ -0,0 +1,526 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare (strict_types=1);
```

test/unittests/UserTest.php
```diff
@@ -0,0 +1,526 @@
...
+/**
+ * Unit tests for User class
+ *
+ * PHP Version 5
```
johnsaigle:
```suggestion
 * PHP Version 7
```

test/unittests/UserTest.php
```diff
@@ -0,0 +1,526 @@
...
+ * @category Tests
+ * @package  Main
+ * @author   Alexandra Livadas <alexandra.livadas@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+require_once __DIR__ . '/../../vendor/autoload.php';
+use PHPUnit\Framework\TestCase;
+/**
+ * Unit test for User class
```
johnsaigle:
```suggestion
 * Unit tests for the User class
```

test/unittests/UserTest.php
```diff
@@ -0,0 +1,526 @@
...
+                'Password_expiry'        => '2020-07-16',
+                'Pending_approval'       => 'Y',
+                'Doc_Repo_Notifications' => 'Y',
+                'language_preference'    => 2,
+                'active_from'            => '2017-07-16',
+                'active_to'              => '2020-07-16'
+          );
+    /**
+     * The userInfo table that should result from calling the factory function
+     * TODO There should be a way to write this that takes up less space!
```
johnsaigle: You could set `$_userInfoComplete` equal to `$_userInfo` and then just push the new elements to the end:

```php
$_userInfoComplete['Sites']                  => 'psc_test;psc_test2';
...
```

test/unittests/UserTest.php
```diff
@@ -0,0 +1,526 @@
...
+     *
+     * @return void
+     * @covers User::singleton
+     * @covers User::factory
+     * @covers User::getData
+     */
+    public function testFactoryRetrievesUserInfo()
+    {
+        $this->_setUpTestDoublesForFactoryUser();
+        $this->_user = \User::factory($this->_username);
```
johnsaigle: Is it possible to do this once in `setUp()` instead of including it in every unit test?

test/unittests/UserTest.php
```diff
@@ -0,0 +1,526 @@
...
+        $this->assertEquals(
+            $this->_userInfoComplete['CenterIDs'],
+            $this->_user->getCenterIDs()
+        );
+    }
+
+    /**
+     * Test that getLanguagePreference returns the correct integer from the user
+     * TODO getLanguagePreference is returning a string rather than an int.
```
johnsaigle: The function signature in the User class looks like getLanguagePreference() can only return `int`. What error are you getting with this test?

test/unittests/UserTest.php
```diff
@@ -0,0 +1,526 @@
...
+     */
+    private $_username;
+    /**
+     * Stores user information
+     *
+     * @var array
+     */
+    private $_userInfo
+        = array('ID'                     => 1,
+                'UserID'                 => '968775',
```
johnsaigle: It's not necessary but consider using constants for these values instead of "magic numbers". It'll make this code a little easier to maintain if the values change, for whatever reason.

php/libraries/User.class.inc
AlexandraLivadas: I just wanted to comment first to see if anyone had any reasons for me not to. Will do now!

test/unittests/UserTest.php
AlexandraLivadas: The error is that it is returning a string rather than an int. It's strange because in the "users" table, the language_preferences row is supposed to be an int, but when you use either getData('language_preference') or getLanguagePreference(), it returns a string. I think when the user object gets populated from the database (in the factory() method), it saves it as a string rather than an int!

AlexandraLivadas: I can edit the factory() method so that it saves 'language_preference' as an int, or I can change the getLanguagePreference method so that it expects a string. Which do you think is best @johnsaigle ?

test/unittests/UserTest.php
AlexandraLivadas: I don't think so because the factory() method essentially just fills the user object with information from the database. If you take a look at the [Candidate unit tests](https://github.com/aces/Loris/blob/minor/test/unittests/CandidateTest.php) (including the ones written not by me), it's similar to how the select() method is used in every unit test to populate the Candidate object with information.

test/unittests/UserTest.php
driusan: Anything that's selected from the database comes back as a string unless explicitly cast. That's just the way the database driver works in PHP.

AlexandraLivadas: So should I just change getLanguagePreference so that it expects a string?

driusan: I'm not really familiar with getLanguagePreference, but if the parameter is really an integer encoded as a string I think it would be better to cast it to an int before calling than to use a string representation of a number.

johnsaigle: I think you can update the function itself to be `return intval(...)` and that should fix it.

AlexandraLivadas: OK so in the factory method, I can just cast it to an int when it's setting up the user object.

johnsaigle: Yeah that should work.

AlexandraLivadas:
I added a big TODO comment on the isPasswordStrong method in the User class. There was already a FIXME there because the method is kind of confusing. If someone could take a look at what I wrote, I can either make an issue for that or fix the method myself in a future commit! Let me know!

test/unittests/UserTest.php
```diff
@@ -12,6 +12,43 @@
  */
 require_once __DIR__ . '/../../vendor/autoload.php';
 use PHPUnit\Framework\TestCase;
+
+/**
+ * Fake User class that has an exact copy of the User::hasLoggedIn method
```
PapillonMcGill: This should not use a fake version of the  class. It should use the actual User class.

kongtiaowang: The user class has too many dependencies in the original class. Unless rewriting the user class for unit testing.

AlexandraLivadas: If you look at the [method](https://github.com/aces/Loris/blob/53712c1c20b799d9194c3d2781206c17b7a3ab0f/php/libraries/User.class.inc#L638), the factory and database are actually declared within this $count definition, which makes it really hard to test without this Fake class. I think this is related to issue #4989 , but this method creates its database very oddly!

AlexandraLivadas: I think I can fix this by changing how the database is declared. Again, this is related to issue #4989. I will make that change now and see if it works!

test/unittests/UserTest.php
```diff
@@ -0,0 +1,598 @@
...
+    private $_configMap = array();
+    /**
+     * This method is called before each test is executed.
+     *
+     * @return void
+     */
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->_factory = NDB_Factory::singleton();
```
kongtiaowang: \NDB_Factory::singleton();

test/unittests/UserTest.php
```diff
@@ -0,0 +1,598 @@
...
+        $this->_factory->reset();
+        $this->_factory->setTesting(false);
+        $this->_configMock = $this->_factory->Config(CONFIG_XML);
+        $database     = $this->_configMock->getSetting('database');
+        $this->_dbMock     = Database::singleton(
```
kongtiaowang: \Database::singleton(

test/unittests/UserTest.php
```diff
@@ -0,0 +1,598 @@
...
+     *         into a new user object.
+     * @return void
+     * @covers User::insert
+     */
+    public function testInsert()
+    {
+        $newUserInfo = $this->_userInfo;
+        $newUserInfo['ID'] = 2;
+        $newUserInfo['UserID'] = '968776';
+        $this->assertTrue(User::insert($newUserInfo));
```
kongtiaowang: \User::insert($newUserInfo)

test/unittests/UserTest.php
```diff
@@ -0,0 +1,598 @@
...
+
+    /**
+     * Test that newPassword generates a password of the correct length
+     *
+     * @return void
+     * @covers User::newPassword
+     */
+    public function testNewPassword()
+    {
+        $newPassword = User::newPassword(6);
```
kongtiaowang: \User::newPassword(6);

php/libraries/User.class.inc
```diff
@@ -639,12 +649,14 @@ class User extends UserPermissions
      */
     public function hasLoggedIn(): bool
     {
-        $count = (\NDB_Factory::singleton()->database())->pselectOne(
+        $DB = \Database::singleton();
```
PapillonMcGill: should keep NDB_Factory

johnsaigle: Why?

PR #4985, opened 2019-07-24
[Testing] Remove DB Unit package

johnsaigle:
## Brief summary of changes

This PR takes over some of the responsibility of #4433. 
It is concerned only with removing the DB Unit package which is blocking PR #4914.

I'm not sure if this will actually worked but Shen says we don't actually use this package and made moves to remove it. 

1. Travis should succeed. No manual testing necessary.

* Covers some of the functionality of #4433 
* Unblocks #4914 (and therefore #4928) once merged.


johnsaigle:
Seems fine 🤷‍♂

PR #4992, opened 2019-07-31
[Node] Update insecure npm packages

johnsaigle:
## Brief summary of changes
When running `make` on various branches I noticed that several packages we use are marked with high security vulnerabilities. I used `npm audit fix` to address them.

All I did:

`git fetch aces`
`git checkout aces/minor`
`make dev`
`npm audit fix`

and then committed the package-lock.json file.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1009,6 +1009,11 @@
             "ms": "2.0.0"
           }
         },
+        "lodash": {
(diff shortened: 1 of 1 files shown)


driusan:
Should probably go the the major branch by policy, but it seems relatively low-risk and the next major release is probably a ways away, so I'll just merge it

Issue #4998, opened 2019-08-05
[Media] Bug with date of administration: error message says date is in future even when it isn't


jesscall:
<img width="1123" alt="Screen Shot 2019-08-05 at 2 11 57 PM" src="https://user-images.githubusercontent.com/34260251/62485811-c54ce980-b78b-11e9-81c9-b0b32ba341be.png">

Using date 2016-10-28 as an example, get error message saying that date is cannot be in the future.

PR #4999, opened 2019-08-05
[Media Uploader] Bugfix date of administration field

jesscall:
## Brief summary of changes
This PR fixes issue reported in #4998. When entering a date in the past and attempting to upload a file, you get an error saying that the date cannot be in the future. 

date_diff is used to find the difference between the date entered and now. 
`$now - $date` returns a positive number if `$date` is in the past, therefore the PR simply switches the comparison operator. 

1. Try uploading a media file with a date in the past, should work fine.
2. Try uploading a media file with a date in the future, should get the expected message that date cannot be in the future.

### DIFF
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -414,7 +414,7 @@ function checkDateTaken($dateTaken)
         }
 
         $now  = new DateTime();
-        $diff = date_diff($date, $now)->format("%R%a");
+        $diff = intval(date_diff($date, $now)->format("%R%a"));
         if ($diff < 0) {
             showMediaError("Date of administration cannot be in the future", 400);
         }
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -416,7 +416,7 @@ function checkDateTaken($dateTaken)
         $now  = new DateTime();
         $diff = date_diff($date, $now)->format("%R%a");
         if ($diff < 0) {
-            showMediaError("Date of administration can not be in the future", 400);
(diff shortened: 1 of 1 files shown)


driusan:
This is going to the wrong branch. 21.0 is already out.

Jkat:
Hmm it seems `date_diff` always returns a positive number, unless `%r` is specified in the format.

modules/media/ajax/FileUpload.php
```diff
@@ -414,8 +414,8 @@ function checkDateTaken($dateTaken)
         }
 
         $now  = new DateTime();
-        $diff = date_diff($date, $now)->format("%a");
-        if ($diff > 0) {
+        $diff = date_diff($date, $now)->format("%R%a");
+        if ($diff < 0) {
             showMediaError("Date of administration can not be in the future", 400);
```
johnsaigle:
```suggestion
            showMediaError("Date of administration cannot be in the future", 400);
```

modules/media/ajax/FileUpload.php
```diff
@@ -414,9 +414,9 @@ function checkDateTaken($dateTaken)
         }
 
         $now  = new DateTime();
-        $diff = date_diff($date, $now)->format("%a");
-        if ($diff > 0) {
-            showMediaError("Date of administration can not be in the future", 400);
+        $diff = date_diff($date, $now)->format("%R%a");
```
ridz1208: @jesscall the uppercase `R` will return either `+5` or `-5` both being strings and should not be compared with an int as done below in the if statement

use `intval` to convert to integer

https://www.php.net/manual/en/function.intval.php

PR #5000, opened 2019-08-05
[candidate_parameter] Consent Status data validation bugfix

jesscall:
## Brief summary of changes

This PR fixes the logic for data validation on consent status tab. As it stands, if no status stays no or record existed as NULL the validation checks that :
1. consent date **is not** empty
2. withdrawal date **is** empty

However, if Consent Status at one point changes from 'YES' to 'NO' ==> user must fill in date of withdrawal. Therefore, consents that = 'NO' and have non-empty dates of withdrawal are considered non-valid (Displays error message).

1. Change consent from 'yes' to 'no'. Fill in date of withdrawal & submit.
2. Change some other consent or date or simply re-submit the same form. 
3. Make sure you don't get a data validation error.

### DIFF
--- a/modules/candidate_parameters/ajax/formHandler.php
+++ b/modules/candidate_parameters/ajax/formHandler.php
@@ -482,7 +482,7 @@ function editConsentStatusFields($db, $user)
                     // consent date and empty withdrawal date still required
                     // old withdrawal may be filled in
                 if (($oldStatus === null || $oldStatus === 'no') && !empty($date)
-                    && (empty($withdrawal) || !empty($oldWithdrawal))
+                    && ((empty($oldWithdrawal) && empty($withdrawal)) || (!empty($oldWithdrawal) && !empty($withdrawal)))
                 ) {
                     $validated = true;
                 } else if ($oldStatus === 'yes' && !empty($date)
--- a/modules/candidate_parameters/ajax/formHandler.php
+++ b/modules/candidate_parameters/ajax/formHandler.php
@@ -482,7 +482,8 @@ function editConsentStatusFields($db, $user)
                     // consent date and empty withdrawal date still required
                     // old withdrawal may be filled in
                 if (($oldStatus === null || $oldStatus === 'no') && !empty($date)
-                    && ((empty($oldWithdrawal) && empty($withdrawal)) || (!empty($oldWithdrawal) && !empty($withdrawal)))
+                    && ((empty($oldWithdrawal) && empty($withdrawal)) 
+                       || (!empty($oldWithdrawal) && !empty($withdrawal)))
                 ) {
                     $validated = true;
                 } else if ($oldStatus === 'yes' && !empty($date)
--- a/modules/candidate_parameters/ajax/formHandler.php
+++ b/modules/candidate_parameters/ajax/formHandler.php
@@ -450,9 +450,10 @@ function editConsentStatusFields($db, $user)
                          ];
 
         // Validate data
-        $recordExists = array_key_exists($consentID, $candidateConsent);
-        $oldStatus    = $candidateConsent[$consentID]['Status'] ?? null;
-        $validated    = false;
+        $recordExists  = array_key_exists($consentID, $candidateConsent);
+        $oldStatus     = $candidateConsent[$consentID]['Status'] ?? null;
+        $oldWithdrawal = $candidateConsent[$consentID]['DateWithdrawn'] ?? null;
+        $validated     = false;
 
         switch ($status) {
         case 'yes':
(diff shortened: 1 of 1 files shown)


sruthymathew123:
@jesscall Jessica, I am trying to recollect the existing behavior of the consent status; correct me if I am wrong.  Currently(as per the existing test plan), if a consent is updated from and Yes to No, withdrawal date needs to be filled and an error is thrown otherwise. I couldn't find any before or after changes with testing this PR. Can you explain me why this change is really needed. I mean the exact case.

zaliqarosli:
hey @jesscall i don't fully understand from reading the description what the bug actually is. could you say again which logic is not correct?

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -477,12 +477,25 @@ function editConsentStatusFields($db, $user)
...
+                        )
+                    );
+                    // If consent was 'yes' at some point, withdrawal date is !empty
+                    // if it was never 'yes', withdrawal date should be empty
+                    // Either is valid.
+                    if ((!empty($withdrawal) && $countYesStatusHistory > 0)
+                        || (empty($withdrawal) && $countYesStatusHistory === 0)
+                    ) {
+                        $validated = true;
+                    }
```
zaliqarosli: i believe the case you're covering above is already covered below in the else if case? its for when the 'oldStatus' = 'yes'. i'm not sure if i'm fully understanding but counting how many times in the past the consent has been 'yes' doesn't say anything about whether you're currently changing it from 'yes' to 'no'. am i missing something?

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -477,12 +477,25 @@ function editConsentStatusFields($db, $user)
...
+                    $countYesStatusHistory = $db->pselectOne(
+                        "SELECT COUNT(*) FROM candidate_consent_history
+                        WHERE PSCID=:pscid AND ConsentName=:cn AND Status='yes'",
+                        array(
+                         'pscid' => $pscid,
+                         'cn'    => $consentName,
+                        )
+                    );
+                    // If consent was 'yes' at some point, withdrawal date is !empty
+                    // if it was never 'yes', withdrawal date should be empty
```
zaliqarosli: the case for this comment is also covered on line 471-479

jesscall:
@sruthymathew123 @zaliqarosli 

Okay so say we have Consent A and B


User X comes in and sets **consent A** = 'Yes' + date of consent = YYYYMMDD.
A few weeks later, user X removes **consent A**, setting it to 'No' and also submitting a _**date of withdrawal.**_ 

Throughout the study, user X is now entering data for **consent B.** 
They come in and enter a **consent for B** and upon submitting the form receive a data validation error (even though all entries are correct).

All data on the page gets re-submitted. The error in logic comes from Consent A going through the following code:

```
if (($oldStatus === null || $oldStatus === 'no') && !empty($date)
                    && empty($withdrawal)
                ) {
                    $validated = true;
``` 
If you remember correctly, Consent A has consent = 'No' + has a date of withdrawal (not empty!). It therefore never enters the first condition. Instead it defaults to the following:

```
 } else {
                    http_response_code(400);
                    echo('Data failed validation. Resolve errors and try again.');
                    return;
                }
```

The validation error therefore occurs when submitting the form again AFTER consent has gone from 'yes' -->> 'no'


Hope this is clear? Let me know if you still don't understand.

modules/candidate_parameters/ajax/formHandler.php
jesscall: whether date of withdrawal is set depends on whether consent was ever 'yes' in the past. 
The case you are mentioning in the else if checks whether the consent has just NOW changed from 'yes'

zaliqarosli: you should only be putting in a date of withdrawal if you're currently withdrawing a yes consent. is this not true for you?

zaliqarosli:
@jesscall thanks for the explanation! i understand now. i'll look into it given the context!

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -477,12 +477,25 @@ function editConsentStatusFields($db, $user)
...
                     return;
                 }
-            } else { // If no status stays no or record existed as NULL,
-                     // consent date and empty withdrawal date still required
+            } else { // If no status stays no or record existed as NULL
                 if (($oldStatus === null || $oldStatus === 'no') && !empty($date)
-                    && empty($withdrawal)
                 ) {
-                    $validated = true;
+                    $countYesStatusHistory = $db->pselectOne(
```
zaliqarosli: okay so good catch @jesscall . i think the best way to do is to, instead of querying again, define $oldWithdrawal like how $oldStatus was by using candidate->getConsents() and check the new $withdrawal against those

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -477,10 +478,11 @@ function editConsentStatusFields($db, $user)
...
                     return;
                 }
-            } else { // If no status stays no or record existed as NULL,
-                     // consent date and empty withdrawal date still required
+            } else { // If no status stays no or record existed as NULL
+                    // consent date and empty withdrawal date still required
+                    // old withdrawal may be filled in
                 if (($oldStatus === null || $oldStatus === 'no') && !empty($date)
-                    && empty($withdrawal)
+                    && (empty($withdrawal) || !empty($oldWithdrawal))
```
zaliqarosli: i think this allows you to delete an existing withdrawn date without having changed the status

jesscall: @zaliqarosli do you think they should be able to edit the withdrawal date, or should they not be able to make any changes at all ?

zaliqarosli: i think we should allow them to edit date

ridz1208: I dont think theres a better way of doing this simply from the oldstatus info since after 2 reloads we cant possibly know if it was a `yes -> no` change, `no -> no` or `null -> no` 2 reloads ago.

I think there are 2 choices.

1. digging in history to find if the status was ever set to `yes` and requiring the withdrawal date
2. allowing them to remove the withdrawal date after reloading the page...

I think given the amount of changes required for something that simple, option 2 is acceptable, with a note in the code saying this behaviour is expected although  not idea

jesscall: @ridz1208 I had originally done this with option 1 implemented. Should I just revert to that ? Either option is okay with me.

ridz1208: I'd have to see how you were doing it, which commit is it ?

jesscall: The first one

ridz1208:
@jesscall in this case your first commit had it right I think but please add the `inval()` around the result of pselectone since you are comparing it with `===` to an INT at `$countYesStatusHistory === 0`

@zaliqarosli do you agree ?

PR #5004, opened 2019-08-06
[Core] Add config option for projects to specify how they want dates to be displayed

johnsaigle:
## Brief summary of changes

Following today's meeting, this adds a config option to allow projects to specify how they want dates to be displayed in LORIS.

This PR only adds two quick examples of using this Config setting; it is by no means comprehensive.

Also, this will only take place in "PHP contexts" and not within JS date elements as these are OS- and locale-dependent.

## Testing

1. On aces/major go to the issue tracker module and look at the `Last Updated` column.
2. Checkout my branch and run the SQL patch.
3. Go to the configuration module and change the date format to something like `ymd`.
4. Reload the issue tracker module. The date format in the `Last Updated` column should match the format you supplied in step 3.


SQL/0000-00-03-ConfigTables.sql
```diff
@@ -252,3 +253,4 @@ INSERT INTO Config (ConfigID, Value) SELECT ID, 't1'    FROM ConfigSettings WHER
 INSERT INTO Config (ConfigID, Value) SELECT ID, 't2'    FROM ConfigSettings WHERE Name="modalities_to_deface";
 INSERT INTO Config (ConfigID, Value) SELECT ID, 'pd'    FROM ConfigSettings WHERE Name="modalities_to_deface";
 INSERT INTO Config (ConfigID, Value) SELECT ID, 'false'  FROM ConfigSettings WHERE Name="usePwnedPasswordsAPI";
+INSERT INTO Config (ConfigID, Value) SELECT ID, 'Y-m-d'  FROM ConfigSettings WHERE Name="dateDisplayFormat";
```
johnsaigle: The default value here is the ISO date standard.

johnsaigle:
Need to add Raisinbread changes.

johnsaigle:
I'm getting the error:

> 1) NextStageTestIntegrationTest::testNextStageDoespageLoad
Failed asserting that 'Fatal error: Uncaught ConfigurationException: Config setting dateDisplayFormat does not exist in database in /app/php/libraries/NDB_Config.class.inc:302

I don't know why it says the config setting doesn't exist... the patch looks right to me and worked on my VM. Am I doing something wrong that makes it so Travis can't see this config setting?

ridz1208:
@johnsaigle it's becaude the RB changes have not been done. the RB data replaces the defualt configuration to mimic a real project sourcing the patches

johnsaigle:
@ridz1208 Thanks for clarifying. I'll fix this

raisinbread/RB_files/RB_Config.sql
```diff
@@ -33,8 +33,8 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (33,37,'/PATH/TO/Genomic
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (34,38,'/data/uploads/');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (35,40,'main.css');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (36,41,'25');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (37,44,'localhost');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (38,45,'http://localhost');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (37,44,'https://jsaigle-dev.loris.ca');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (38,45,'https://jsaigle-dev.loris.ca');
```
driusan: This isn't right

johnsaigle: Hmmm I think this may be a side-effect of the RB exporter script. Maybe worth opening a ticket for it

raisinbread/RB_files/RB_Config.sql
```diff
@@ -33,8 +33,8 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (33,37,'/PATH/TO/Genomic
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (34,38,'/data/uploads/');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (35,40,'main.css');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (36,41,'25');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (37,44,'localhost');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (38,45,'http://localhost');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (37,44,'https://jsaigle-dev.loris.ca');
```
johnsaigle:
```suggestion
INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (37,44,'localhost');
```

raisinbread/RB_files/RB_Config.sql
johnsaigle:
```suggestion
INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (38,45,'http://localhost');
```

johnsaigle: @driusan I reverted these changes.

SQL/0000-00-03-ConfigTables.sql
```diff
@@ -60,6 +60,7 @@ INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType,
...
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'CSPAdditionalHeaders', 'Extensions to the Content-security policy allow only for self-hosted content', 1, 0, 'text', ID, 'Content-Security Extensions', 25 FROM ConfigSettings WHERE Name="study";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'usePwnedPasswordsAPI', 'Whether to query the Have I Been Pwned password API on password changes to prevent the usage of common and breached passwords', 1, 0, 'boolean', ID, 'Enable "Pwned Password" check', 26 FROM ConfigSettings WHERE Name="study";
+INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'dateDisplayFormat', 'The date format to use throughout LORIS for displaying date information. Note that this may not be reflected in JavaScript elements as they are browser- and locale-dependent.', 1, 0, 'text', ID, 'Date display format', 27 FROM ConfigSettings WHERE Name="study";
```
zaliqarosli: i'm not sure if saying 'this may not be reflected in JavaScript element' is useful for the user who sees this description. Maybe saying something like 'this configuration will affect date displays throughout LORIS - formats for date inputs are browser and locale dependent.' and then also maybe add that this config setting is different to 'DOB Format' config option and say how

raisinbread/RB_files/RB_ConfigSettings.sql
```diff
@@ -97,5 +97,6 @@ INSERT INTO `ConfigSettings` (`ID`, `Name`, `Description`, `Visible`, `AllowMult
...
 INSERT INTO `ConfigSettings` (`ID`, `Name`, `Description`, `Visible`, `AllowMultiple`, `DataType`, `Parent`, `Label`, `OrderNumber`) VALUES (99,'usePwnedPasswordsAPI','Whether to query the Have I Been Pwned password API on password changes to prevent the usage of common and breached passwords',1,0,'boolean',1,'Enable \"Pwned Password\" check',22);
 INSERT INTO `ConfigSettings` (`ID`, `Name`, `Description`, `Visible`, `AllowMultiple`, `DataType`, `Parent`, `Label`, `OrderNumber`) VALUES (100,'EnvironmentFile','Name of the environment file that need to be sourced for the imaging pipeline',1,0,'text',69,'Name of the environment file',20);
+INSERT INTO `ConfigSettings` (`ID`, `Name`, `Description`, `Visible`, `AllowMultiple`, `DataType`, `Parent`, `Label`, `OrderNumber`) VALUES (101,'dateDisplayFormat','The date format to use throughout LORIS for displaying date information. Note that this may not be reflected in JavaScript elements as they are browser- and locale-dependent.',1,0,'text',1,'Date display format',27);
```
zaliqarosli: this needs to be updated to be identical with the other SQL files above

raisinbread/RB_files/RB_Config.sql
```diff
@@ -91,5 +91,7 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (94,99,'true');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (95,49,'1000');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (96,100,'environment');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (97,70,'/data-raisinbread/');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (98,65,'true');
```
zaliqarosli: What config is this for?

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -780,13 +780,11 @@ class Edit_User extends \NDB_Form
...
 
-        //------------------------------------------------------------
-
         // active time windows
-
+        $dateFormat  = $config->getSetting('dateDisplayFormat');
         $dateOptions = array(
                         'language'       => 'en',
-                        'format'         => 'YMd',
+                        'format'         => $dateFormat,
```
zaliqarosli: i believe doing this just changes what format gets appended to the date element in the LorisForm. I'm not sure if this does much in terms of changing the UI. regardless, you're making these changes for date elements (input i believe - i tracked addBasicDate function to this https://github.com/aces/Loris/blob/c75433f8dabd31b32bf5e69e659c744ec99af96e/php/libraries/LorisForm.class.inc#L270 - see addDate's doc comment) whereas this PR should change format for date displays

zaliqarosli: i guess a good example of a date display to play with would be the dates in the data table of media, candidate_list or issue_tracker

zaliqarosli: orrr you can maybe convert the format display on the js side using a number of functions in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date

raisinbread/RB_files/RB_Config.sql
johnsaigle: I think this was generated by the Raisinbread export script so I'm not sure exactly

johnsaigle: Oh I think this might be for usePwnedPasswords actually. Looks like Raisinbread is missing it.

johnsaigle: Not sure. I'll delete it.

raisinbread/RB_files/RB_Config.sql
```diff
@@ -33,9 +33,9 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (33,37,'/PATH/TO/Genomic
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (34,38,'/data/uploads/');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (35,40,'main.css');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (36,41,'25');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (37,44,'localhost');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (38,45,'http://localhost');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (39,48,'This database provides an on-line mechanism to store both imaging and behavioral data collected from various locations. Within this framework, there are several tools that will make this process as efficient and simple as possible. For more detailed information regarding any aspect of the database, please click on the Help icon at the top right. Otherwise, feel free to contact us at the DCC. We strive to make data collection almost fun.');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (37,44,'https://jsaigle-dev.loris.ca');
```
johnsaigle:
```suggestion
INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (37,44,'localhost');
```

raisinbread/RB_files/RB_Config.sql
```diff
@@ -33,9 +33,9 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (33,37,'/PATH/TO/Genomic
...
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (38,45,'https://jsaigle-dev.loris.ca');
```
johnsaigle:
```suggestion
INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (38,45,'localhost);
```

raisinbread/RB_files/RB_Config.sql
```diff
@@ -34,8 +34,8 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (34,38,'/data/uploads/')
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (35,40,'main.css');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (36,41,'25');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (37,44,'localhost');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (38,45,'http://localhost');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (39,48,'This database provides an on-line mechanism to store both imaging and behavioral data collected from various locations. Within this framework, there are several tools that will make this process as efficient and simple as possible. For more detailed information regarding any aspect of the database, please click on the Help icon at the top right. Otherwise, feel free to contact us at the DCC. We strive to make data collection almost fun.');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (38,45,'localhost);
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (39,48,NULL);
```
johnsaigle:
```suggestion
INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (39,48,'This database provides an on-line mechanism to store both imaging and behavioral data collected from various locations. Within this framework, there are several tools that will make this process as efficient and simple as possible. For more detailed information regarding any aspect of the database, please click on the Help icon at the top right. Otherwise, feel free to contact us at the DCC. We strive to make data collection almost fun.');
```

raisinbread/RB_files/RB_Config.sql
```diff
@@ -34,7 +34,7 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (34,38,'/data/uploads/')
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (35,40,'main.css');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (36,41,'25');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (37,44,'localhost');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (38,45,'http://localhost');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (38,45,'localhost);
```
johnsaigle:
```suggestion
INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (38,45,'http://localhost');
```

johnsaigle:
@zaliqarosli You're right, the examples I used didn't actually make sense.

I went into issue tracker like you suggested and changed the dates that way. Rather than pass an unformatted date into JS and format it there, I created a function to convert the date format in PHP before it gets passed into the front-end. 

I chose to do this because the config setting is a PHP date format such as `'Y-m-d'` and there's no native way to convert that format into something that JS understands.

I updated the PR description with testing instructions.

raisinbread/RB_files/RB_Config.sql
```diff
@@ -91,5 +91,6 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (94,99,'true');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (95,49,'1000');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (96,100,'environment');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (97,70,'/data-raisinbread/');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (99,101,'Y-m-d');
```
zaliqarosli: im wondering why its ID 99 and not 98

johnsaigle:
```suggestion
INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (98,101,'Y-m-d');
```

I'm just gonna change it manually.

johnsaigle:
@zaliqarosli Responding to your points:

1. How come? I chose the format cause it's the ISO standard. I don't know how often the hour, minute, second are needed. Maybe for something like biospecimen entry it will be important? But in general I think dates down to the day are sufficient. I think your point 3 actually supports this.

2. Fair - I could add a validation section in the config class to make sure that it's a valid date format by trying to instantiate a new DateTime and catching an Exception when it fails. I think a dropdown isn't a good option just because there are so many configurations of dates... it's probably equal to the permutations of all the date symbols and all the possible delimiters.

3. In that case, `H:i:s` become 0:

```php
php > $dt = new DateTime('2019-10-02');
php > echo $dt->format('Y-m-d H:i:s');
2019-10-02 00:00:00
```

zaliqarosli:
@johnsaigle 

Because if the data was in `Y-m-d H:i:s` before, i.e. on major branch, setting the default to only `Y-m-d` makes you lose data. this is the case with your chosen example in issue tracker.

there are lots of "datetime" or "timestamp" types in our sql tables and I don't think we should be erasing that data as a default. I think providing too much information, e.g. `00:00:00` time, as the default is better than letting it appear that time doesn't exist at all. people can choose to opt out of the time, but may not necessarily know they can opt in, especially if the config doesn't show you all the possible options

johnsaigle:
Fair enough, I'll change it to include `H:i:s`

zaliqarosli:
@johnsaigle this still "needs work"? I'm ready to re-review

johnsaigle:
@zaliqarosli I wanted to add this feature based on one of your reviews:

> Fair - I could add a validation section in the config class to make sure that it's a valid date format by trying to instantiate a new DateTime and catching an Exception when it fails.

Haven't gotten around to it yet.

zaliqarosli:
ahh i see. yeah no worries, take your time! was just wondering

zaliqarosli:
@johnsaigle hahah what the heck! that's pretty hilarious. Yeah maybe added it in the document would be good. Not sure where is the most suitable place however. Is there a Config wiki explaining all the config options on the front-end?

johnsaigle:
No I don't think there is a document explaining all the config settings... I could start a PR writing one for the new Wiki though. It seems like an important thing to have anyway.

driusan:
A list of config settings doesn't belong in the wiki since they change with (nearly) every release.. it needs to be tracked by git (which I guess is maybe what you mean by "new wiki")

johnsaigle:
Yes exactly. I think a page could be create don the "new Wiki", i.e. on ReadTheDocs and tracked according to every release.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -868,13 +868,50 @@ class Edit_User extends \NDB_Form
...
+            );
+
+            // active time windows
+            $dateFormat  = $config->getSetting('dateDisplayFormat');
+            $dateOptions = array(
+                'language'       => 'en',
+                'format'         => $dateFormat,
+                'addEmptyOption' => 'true',
+            );
+
```
zaliqarosli: something doesn't seem right with all of these changes. @johnsaigle could you verify and/or explain if this is intentional?

johnsaigle: I think I made a mistake while rebasing. I'll fix it.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -868,13 +868,26 @@ class Edit_User extends \NDB_Form
...
+            $this->addSelect('Active', 'Active', array('Y' => 'Yes', 'N' => 'No'));
+            $this->addSelect(
+                'Pending_approval',
+                'Pending approval',
+                array(
+                    'Y' => 'Yes',
+                    'N' => 'No',
+                )
+            );
+
```
zaliqarosli: result of rebase i think. needs to be removed

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -871,12 +871,13 @@ class Edit_User extends \NDB_Form
                 )
             );
 
-            $dateOptions = array(
+            // active time windows
+            $dateFormat     = $config->getSetting('dateDisplayFormat');
+            $dateOptions    = array(
                 'language'       => 'en',
-                'format'         => 'YMd',
+                'format'         => $dateFormat,
```
zaliqarosli: i believe this date option is for a date input element. this PR wants to format date displays i.e. date shown as string

raisinbread/RB_files/RB_Config.sql
```diff
@@ -92,9 +92,12 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (95,49,'1000');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (96,100,'environment');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (97,70,'/data-raisinbread/');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (98,93,'V1');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (99,101,'');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (102,19,'false');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (99,65,'true');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (100,19,'false');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (101,65,'true');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (102,101,'Y-m-d H:i:s');
```
zaliqarosli: these changes are unintentionally affecting other configs

raisinbread/RB_files/RB_Config.sql
```diff
@@ -92,9 +92,12 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (95,49,'1000');
...
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (103,102,'/data/document_repository_uploads/');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (104,103,'/data/data_release_uploads/');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (105,104,'Y-m-d H:i:s');
```
zaliqarosli: you have this config duplicated on line 98

johnsaigle: Thanks, I'll redo the RB export

ridz1208:
@johnsaigle from what I can see, in RB you are adding the config setting but not the actual config value, which means the default will be used in the issue tracker which is ATOM

is that your intent ?

johnsaigle:
No I think I need to fix that. I was having some issues with raisinbread exporting + rebasing

ridz1208:
@johnsaigle let me know if you need help with it

johnsaigle:
Feel free to do it. I think you should be abel to push to my branch. Let me know if there's an issue

ridz1208:
@driusan all urs

PR #5013, opened 2019-08-08
[Core] Forbid running the install script as root

### LINKED ISSUE #5008: [install] `sudo ./install.sh` will fail, but won't tell you
Since #4109 is new in Release 21 : 
If the install script is called w sudo (per the current wiki install guides, links below) - `project/config.xml` is not owned by lorisadmin --
The install script gives a success message, but Loris itself will be incompletely(non-functionally) set up, and the user can't proceed.  The manual troubleshooting is a little complicated. 

It's occurring because #4109 introduced a new script called by install.sh which is run basically as root (when the install script is sudo'd by lorisadmin) not as lorisadmin 

Graceful fix: 
- [ ] Fix tools/create-project.sh so that it sets the permissions appropriately in case of sudo 

less graceful: 
- [ ] Check/Abort/Warn install.sh if run with sudo.  -> update:  PR #5008

Also to do: (in either case) remove the `sudo` from the install.sh instruction in :
- [x] Update the [Brief Install guide here](https://github.com/aces/Loris/wiki/Installing-Loris-in-Brief#running-the-install-script)
- [x] Update the [Detailed install guide here](https://github.com/aces/Loris/wiki/Installing-Loris-in-Depth#running-the-install-script)

johnsaigle:
## Brief summary of changes

Using the install script with root permissions causes issues because it will create files and folders inaccessible to the `lorisadmin` user. The script executes just fine without it so we may as well prevent less experienced developers from getting stuck.

In general we shouldn't require root level permissions to do anything because that opens all kinds of cans of worms from a usability and security perspective. 

1. Within the `tools/` directory, run `sudo bash install.sh`. The script should exit with an error message
2. Without sudo, make sure the same command starts the script normally.

* Resolves #5008

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -13,6 +13,11 @@ if ! test -t 0 -a -t 1 -a -t 2 ; then
     exit 2
 fi
 
+if [ $(id -u) = 0 ]; then
+   echo "This script must not be run as root"
+   exit 1
+fi
+
 # Create logs directory, if needed.
 mkdir -p logs
 


driusan:
This is a duplicate of #5010 (which is already merged)

christinerogers:
Thanks Dave! 

* Resolves #5008 

Would have avoided the duplicate PRs.

PR #5022, opened 2019-08-13
[Imaging Browser - Data Framework] Make Imaging Browser menu filter use the DataFrameworkMenu

cmadjar:
## Brief summary of changes

This updates the imaging browser module to use the DataFrameworkMenu class.

Follow the test plan of the module to make sure there was no change in behaviour


modules/imaging_browser/php/imaging_browser.class.inc
```diff
@@ -284,117 +98,84 @@ END";
...
         }
 
-        $outputTypes = $DB->pselect(
-            "SELECT DISTINCT OutputType AS outputType 
-            FROM files WHERE FileType='mnc' AND OutputType!='native'",
+        // get list of scan types that should be present in the data table
+        $config             = \NDB_Config::singleton();
+        $toTable_scan_types = $config->getSetting('tblScanTypes');
+        $all_scan_types_2d  = $DB->pselect(
+            "SELECT ID, Scan_type FROM mri_scan_type mri",
```
cmadjar: I have a feeling this is not right. It works though. @driusan is that OK to still have a query and use `$DB` in here?

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
```diff
@@ -0,0 +1,275 @@
...
+        $DB     = \Database::singleton();
+
+
+        // ===========================================================
+        // Get the different scan types to be displayed in data table
+        // ===========================================================
+
+        // Grep all the different scan types present in mri_scan_type
+        $toTable_scan_types = $config->getSetting('tblScanTypes');
+        $all_scan_types_2d  = $DB->pselect(
```
cmadjar: Same comment here. Is that OK to still use $DB here to get the list of available scan types? @driusan?

modules/imaging_browser/php/imaging_browser.class.inc
```diff
@@ -56,214 +56,28 @@ class Imaging_Browser extends \NDB_Menu_Filter
...
-             's.ID as sessionID',
-             'GROUP_CONCAT(DISTINCT modality.Scan_type) as sequenceType',
-             "$PendingNewquery as pending",
-            ),
-            $modalities_subquery_as
-        );
-
-        $this->order_by = 'c.PSCID, s.Visit_label';
-        $this->group_by = 's.ID';
+        return true;
```
driusan: hard to see what this is commenting on because of the way github is displaying it, but: setting useSiteFilter to true is going to mean that the access_all_profile permission applies to the imaging browser. Isn't that a change in behaviour/incorrect, because it uses differently named all site filters?

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
```diff
@@ -0,0 +1,275 @@
...
+            "SELECT ID, Scan_type FROM mri_scan_type mri",
+            array()
+        );
```
driusan: It seems like it would make more sense to have the list of scan types injected as a parameter to the constructor, rather than querying the database directly here for testability..

@xlecours @ridz1208 (maybe @AlexandraLivadas) thoughts?

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
```diff
@@ -0,0 +1,275 @@
...
+            ) ";
+        }
+
+        
+        // ===================================================
+        // Determine which data should be returned depending
+        // on user permission
+        // ===================================================
+
+        if (!$user->hasPermission('imaging_browser_view_allsites')) {
```
driusan: I think these should be done with filters on the base provisioner similar to the UserSiteMatch filter, not by concatenating changes to the SQL

modules/imaging_browser/php/imaging_browser.class.inc
driusan: You need to get it from somewhere, but I think it would be a better idea to encapsulate it in some kind of `getScanTypes()` method (not sure if the proper place for that would be here or in the Utility class, but it would make the code easier to follow)

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
driusan: see my other comment.. I think it would be a better idea to inject the list via the constructor. You're directly instantiating this class with the `new` keyword where it's used (unlike, say, the classes that extend NDB_Page which are instantiated via reflection..) so you can change the signature however you want.

modules/imaging_browser/php/imaging_browser.class.inc
cmadjar: true. Did not think of that. I'll set it to false instead...

modules/imaging_browser/php/imaging_browser.class.inc
cmadjar: OK. I will add it to the Utility class then as it might be useful for other modules down the road.

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
cmadjar: haha. That was another one I wanted to ask you but totally forgot about it. I guess my suspicions that it was not right were correct ;).

cmadjar: should be done in the last commit. @driusan was it what you expected?

Note: that breaks the candidate list as it expects that the new SiteFilters is called in the data provisioner of the candidate list. I have a feeling I missed something cause it is weird to have to change all the modules every time we have a new permission to use for filters.

modules/imaging_browser/php/imaging_browser.class.inc
```diff
@@ -56,214 +56,50 @@ class Imaging_Browser extends \NDB_Menu_Filter
...
-            ),
-            $modalities_subquery_as
-        );
+    /**
+     * Tells the base class that this page's provisioner can support
+     * the UserSiteMatch filter.
+     *
+     * @return bool always true
+     */
+    public function useImagingBrowserSiteFilter() : bool
```
driusan: This doesn't seem to be used anywhere

modules/imaging_browser/php/imaging_browser.class.inc
```diff
@@ -56,214 +56,50 @@ class Imaging_Browser extends \NDB_Menu_Filter
...
 
-        $this->order_by = 'c.PSCID, s.Visit_label';
-        $this->group_by = 's.ID';
+    /**
+     * Tells the base class that this page's provisioner can support
+     * the UserSiteMatch filter.
+     *
+     * @return bool always true
+     */
+    public function useImagingBrowserPhantomSiteFilter() : bool
```
driusan: This doesn't seem like it's used anywhere either.

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
driusan: I don't see any changes to the candidate list in the last commit, so I'm assuming you didn't push it.. but if I can guess what the problems are:

1. You should just overload the `getDataProvisionerWithFilters` instead of doing it in the base class
2. It might make sense to have the base class refactored to have a function which returns a list of types of per-site permissions instead of a boolean, but that might require quite a big refactoring

modules/imaging_browser/php/imaging_browser.class.inc
cmadjar: sorry... Forgot to commit DataFrameworkMenu file in the last commit. The next commit will have those changes.

modules/imaging_browser/php/imaging_browser.class.inc
cmadjar: same. Forgot to commit the DataFrameworkMenu file...

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
cmadjar: did suggestion 1. in the last commit

modules/imaging_browser/php/imaging_browser.class.inc
```diff
@@ -284,117 +98,97 @@ END";
...
-    function toArray()
+    public function getDataProvisionerWithFilters(): \LORIS\Data\Provisioner
     {
-        $data  = parent::toArray();
-        $index = array_search('SessionID', $data['Headers']);
-        if ($index !== false) {
-            $_SESSION['State']->setProperty(
-                'mriSessionsListed',
-                array_column($data['Data'], $index)
+
```
driusan: It's getting closer but:

1. I think it should start with `$provisioner = parent::getDataProvisionerWithFilters` instead of duplicating the code from it, so that other filters added to the base (ie. projects eventually) automatically get applied
2. I don't think the logic for the if statements is correct. It looks like the way it's written unless you have *both* permissions anything not at your site gets filtered out.. but I'm not entirely sure what the logic is supposed to be for the permissions intersection either (since the logic in the code was always confusing.)
3. You may need to create new filter for the imaging browser instead of using the UserSiteMatch, because the filter you're trying to use is dependent on both CenterID *and* Entity Type at the same time, not just CenterID.

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
cmadjar: done.

modules/imaging_browser/php/imaging_browser.class.inc
cmadjar: 1. oh right. Fixed in the next commit
2. oops... Yeah.. I spent a fair amount of time thinking about this logic. I think I finally got it! Check the next commit
3. some of the filters were already created by Xavier so I used them in the code. I just had to create an additional filter for users that have permission view_ownsite + phantom_allsites. For everything else, I just needed to call the right filter.

Thank you for your help and guidance!!

modules/imaging_browser/php/imaging_browser.class.inc
```diff
@@ -284,117 +98,119 @@ END";
...
-                'mriSessionsListed',
-                array_column($data['Data'], $index)
-            );
+
+        $provisioner = parent::getDataProvisionerWithFilters();
+
+        $user = \User::singleton();
+
+        // UserSiteMatch for imaging browser specific permissions
+        if ($user->hasPermission("imaging_browser_view_allsites") == false) {
```
driusan: I think this might be easier to read/review if you used "&&" to structure it more like a truth table:

ie
```
if($user->hasPermission(A) && $user->hasPermission(B)) {
    // description of what that means
} else if($user->hasPermission(A) && !user->hasPermission(B)) {
    // description of what that means
} else if(!$user->hasPermission(A) && user->hasPermission(B)) {
    // description of what that means
} else {
   // has neither, description of what that means
}
```

xlecours: Here is the logic I came up with for `ProjectImagesRowProvisioner`. 
```
        // If a user can't view images from all sites check if they can view images
        // from its own sites. If not, limit images to phantoms.
        if (!$user->hasPermission('imaging_browser_view_allsites')) {
            if ($user->hasPermission('imaging_browser_view_site')) {
                $new = $new->filter(
                    new \LORIS\Data\Filters\UserSiteMatch()
                );
            } else {
                $new = $new->filter(
                    new \LORIS\Data\Filters\OnlyPhantoms()
                );
            }
        }

        // If a user can't view phantoms from all sites, check if they can view
        // phantoms from its own sites. If not, filter out phantoms.
        if (!$user->hasPermission('imaging_browser_phantom_allsites')) {
            if ($user->hasPermission('imaging_browser_phantom_ownsite')) {
                $new = $new->filter(
                    new \LORIS\Data\Filters\UserSiteMatch()
                );
            } else {
                $new = $new->filter(
                    new \LORIS\Data\Filters\NoPhantoms()
                );
            }
        }
        return $new;
```
I your case, replace `$new` by `$provisioner`.
I hope it helps.

There is still a "weak" point where a user that do not have `imaging_browser_view_allsites` but have `imaging_browser_view_site` will get only images and phantom from his own sites regarless of if he have `imaging_browser_phantom_allsites`. That is because `UserSiteMatch` applies to every rows, not just on images or phantom. This is yet an other argument to have a specific menu filter for phantoms.

modules/imaging_browser/php/imaging_browser.class.inc
```diff
@@ -284,117 +98,119 @@ END";
...
+            if ($user->hasPermission("imaging_browser_view_site") == false) {
+                // user only has access to phantom scans from all sites
```
xlecours: there should be a check to see if the user have `imaging_browser_phantom_ownsite`

modules/imaging_browser/php/imaging_browser.class.inc
```diff
@@ -284,117 +98,119 @@ END";
...
+                } else {
+                    // user has access to own site scans only
+                    $provisioner = $provisioner->filter(
+                        new \LORIS\Data\Filters\UserSiteMatch()
+                    );
+                    $provisioner = $provisioner->filter(
+                        new \LORIS\Data\Filters\NoPhantoms()
+                    );
+                }
+            }
```
xlecours: missing else statement for when the user do have `imaging_browser_view_allsites` but might need to filter out phantoms.

modules/imaging_browser/php/imaging_browser.class.inc
cmadjar: right. Good catch. Will add it in the next commit.

I totally agree that there should be two different menu filters for phantoms and candidates. Would be much easier to maintain. But I guess it is behind the scope of this PR though...

modules/imaging_browser/php/imaging_browser.class.inc
```diff
@@ -284,117 +98,119 @@ END";
...
+                if ($user->hasPermission('imaging_browser_phantom_allsites')) {
+                    // user has access to own site scans and all sites phantom
+                    // TODO
+                    $provisioner = $provisioner->filter(
+                        new \LORIS\Data\Filters\AllSitesPhantomsOwnSitesScans()
+                    );
+                } else if ($user->hasPermission('imaging_browser_phantom_ownsite')) {
+                    // user has access to own site scans and own site phantom
+                    $provisioner = $provisioner->filter(
+                        new \LORIS\Data\Filters\UserSiteMatch()
```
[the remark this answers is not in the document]
cmadjar: actually, for this one, user has access to his own sites candidate scans and phantom scans.

(imaging_browser_view_site == true and imaging_browser_phantom_ownsites == true in this if statement)

However, the filter you are suggesting will be added by adding an else statement to the first `if` for all sites candidate scans and phantom restriction...

That is intense...

modules/imaging_browser/php/imaging_browser.class.inc
cmadjar: indeed. Will add it shortly!

cmadjar:
@xlecours I think I implemented all your suggestions in the last commit. Those permissions are making me dizzy...

modules/imaging_browser/php/imaging_browser.class.inc
driusan: If you think that's a good idea to split it and make things easier, maybe it would be a good idea to do that first and block this until it's done?

cmadjar: I think it would be easier to merge this one and then work on splitting this it into two menu filters. The work to convert the imaging browser module to the data framework is done. 

All that would be left is splitting this into two menu filters for next round. Seems a lot easier to first merge this and then work on the split.

modules/imaging_browser/php/imaging_browser.class.inc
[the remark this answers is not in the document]
xlecours: You are right. Sorry :)

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
```diff
@@ -0,0 +1,235 @@
...
+        // =================================================
+        // Final query
+        // =================================================
+
+        parent::__construct(
+            "SELECT 
+              p.Name as Site,
+              c.PSCID as PSCID,
+              c.CandID as DCCID,
+              (SELECT Name
```
xlecours: use LEFT JOIN on project instead of a subquery.

php/libraries/Utility.class.inc
```diff
@@ -918,5 +918,31 @@ class Utility
         }
         return $languages;
     }
+
+    /**
+     * Get the list of MRI scan type available in the database
```
xlecours: I am not sure what this function is supposed to do. 
should it return all the scan types from mri_scan_type or all the scan_types where the is at least one file with that scan_type? 
If it's only used in the imaging_browser to limit the number on queries programmatically generated then it should probably be removed. Maintaining this function and the array intersect adds overhead for something that should probably be handle by the frontend anyway.

cmadjar: It was first in the provisionner row file but @driusan made a comment about it (https://github.com/aces/Loris/pull/5022#discussion_r313568765) and I moved it the Utility.class.inc. I think he mentioned that this function could also be used by the API? Honestly I don't remember exactly.

@xlecours @driusan could you discuss? I don't have any opinion on what would be best here and will perform the change you suggest after agreement :).

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
```diff
@@ -0,0 +1,235 @@
...
+            "SELECT 
+              p.Name as Site,
+              c.PSCID as PSCID,
+              c.CandID as DCCID,
+              (SELECT Name 
+                FROM Project 
+                WHERE ProjectID=c.ProjectID) AS project,
+              s.visit_label as Visit_Label,
+              $PendingFailSubquery as Visit_QC_Status,
+              MIN(md.AcquisitionDate) as First_Acquisition,
```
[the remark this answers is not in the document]
cmadjar: done in the next commit

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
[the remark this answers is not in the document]
cmadjar: done in the next commit

php/libraries/Utility.class.inc
driusan: Did you link to the right comment? The one you linked to doesn't say anything about the Utility class, it just says the scan types should be passed in via a constructor argument

xlecours: @cmadjar @driusan, I guess it is fine having it in the Utility. I was missing the place where it's also used in the imaging_browser.class.inc

xlecours:
@cmadjar , following @driusan review on #4244 I did some changes in the Phantom and Scans filter/ing . Now the filters are always applied but they have more responsibilities. They look at the user's permissions. Feel free to grab the logic from there (or not).

cmadjar:
@xlecours I was a bit too drastic when I deleted the project subquery and the visit label got deleted too, hence the bug you found... Oooops. Should be fixed now

cmadjar:
Waiting for https://github.com/aces/Loris/pull/4244/commits/c5ba29dfc74dc6b3f953eb0c25c68b838256635f to get merged so can pull the changes for the permissions and adapt the same logic for phantom as the one used for the API

johnsaigle:
#4244 has been merged so I'm removing Blocked

cmadjar:
Added back the blocked as it is now blocked by PR #5326. Sorry, should have updated it last week...

cmadjar:
@xlecours I integrated the logic for the permission that you worked on for the API. If you want to take a last look?

cmadjar:
@xlecours would you have time today to re-review this PR? This one is important for the 22.0.0 release of LORIS. Thank you!

driusan:
@cmadjar It's failing Travis

cmadjar:
@driusan it's passing Travis now

@johnsaigle Minor will be pulled to major once this is merged (with the project PR as well I think)

cmadjar:
@driusan weird... I swear I saw the green checkmark but now it is back to running... Go figure.

cmadjar:
@driusan yeah! It's green again!

driusan:
@cmadjar okay, but it also has 387 files changed.. I think you messed up a rebase

php/libraries/Utility.class.inc
```diff
@@ -915,5 +915,31 @@ class Utility
...
+            "SELECT ID, Scan_type 
+             FROM mri_scan_type mri
+                JOIN files f ON (f.AcquisitionProtocolID=mri.ID)",
+            array()
+        );
+
+        $scan_types = array();
+        foreach ($scan_types_DB as $scan_type) {
+            $scan_types[$scan_type['ID']] = $scan_type['Scan_type'];
+        }
```
johnsaigle:
```php 
 foreach ($scan_types_DB as $id => $scanType) {
     $scan_types[$id] = $scanType;
 }
```

A little nicer. Not a blocking change

src/Data/Filters/AllSitesPhantomsOwnSitesScans.php
```diff
@@ -0,0 +1,58 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

src/Data/Filters/AllSitesScansOwnSitesPhantoms.php
```diff
@@ -0,0 +1,59 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
```diff
@@ -0,0 +1,234 @@
...
+{
+    /**
+     * Create an ImagingBrowserRowProvisioner, which gets rows for
+     * the imaging browser menu table.
+     */
+    function __construct()
+    {
+        $config = \NDB_Config::singleton();
+        $user   = \User::singleton();
+        $DB     = \Database::singleton();
```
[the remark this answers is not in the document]
cmadjar: oooops. Removed in the next commit

PR #5026, opened 2019-08-14
[Test] Update Makefile unittests target to run the 'dockerized' script

johnsaigle:
## Brief summary of changes

We want the unit tests to run inside our docker environment so that everything works nicely. 

This changes the target of the Makefile to run the `npm` unit test script which in turn runs the "dockerized unit tests" script.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ checkstatic: phpdev
 	vendor/bin/phan
 
 unittests: phpdev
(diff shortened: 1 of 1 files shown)


driusan:
No we don't. There's no reason to add a dependency to docker locally.

johnsaigle:
I'm not really confident that the unit tests work outside of Docker anymore. Are you able to get them to work locally running `make unittests` as it is now?

PR #5051, opened 2019-08-21
Enable bash strict checking on all shell scripts

gdevenyi:
## Brief summary of changes
- standardize bash shebang
- enable strict checking for all scripts

These scripts need to be used in their expected operation to see if this exposes any bugs

### DIFF
--- a/test/dockerized-integration-tests.sh
+++ b/test/dockerized-integration-tests.sh
@@ -1,5 +1,5 @@
-#!/bin/bash
-set -e
+#!/usr/bin/env bash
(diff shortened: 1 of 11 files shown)


johnsaigle:
Thanks for issuing this! It's a big help and some of our bash scripts definitely need some love. 

I think many of the scripts modified here will be tested automatically by Travis (everything to do with unit and integration tests).

We'll need to manually test:
* install.sh
* create-project.sh
* tools/package_files.sh

And perhaps others, but definitely at least these ones.

PR #5058, opened 2019-08-21
Add shebang to executable scripts and files non-executable if they are not a script

johnsaigle:
## Brief summary of changes

Adds PHP shebangs to files that are executable. This is required by the Debian standards and we'd like to make a package one day ideally. This is a precursor to that.

There are also a few files in the repository that were executable but should not have been. This PR fixes those.

This also allows us to run our scripts using `./script.php` instead of `php script.php` which is nice.

Related: #4434 and https://github.com/aces/Loris/issues/3154


gdevenyi:
Should check shell scripts for +x as well

johnsaigle:
Yep `lintian` flags those as well. There was one (deprecated) shell script that did not have the shebang necessary and it's updated here.

PR #5065, opened 2019-08-22
[jsx/ButtonElement] Bring classname out to a prop with default

zaliqarosli:
## Brief summary of changes

All this PR does is essentially pull out the classname "row form-group" into a default value for a prop called className. This change does not affect any existing ButtonElements out there, but does allow any future ones to have a custom class name for its wrapper div.

En plus, I have added some missing props to the ButtonElement.propTypes object, and added an empty string default to this.props.name which didn't have one previously.

#### Testing instructions

1. Run `make` on this branch and test a couple reactified buttons in Loris. Make sure they are rendered from the ButtonElement in Form.js. There should be no changes whatsoever to these existing buttons.

### DIFF
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -1552,7 +1552,7 @@ class ButtonElement extends Component {
 
   render() {
     return (
-      <div className="row form-group">
+      <div className={this.props.className}>
         <div className={this.props.columnSize}>
           <button
             name={this.props.name}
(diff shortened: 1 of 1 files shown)


jsx/Form.js
```diff
@@ -1552,7 +1552,7 @@ class ButtonElement extends Component {
 
   render() {
     return (
-      <div className="row form-group">
+      <div className={this.props.className}>
```
driusan: What happens if the "row" and "form-group" are missing from the div? Is that something that might sometimes be desirable, or would it completely break the rendering (in which case, maybe it should be passing in additional classes that it merges with them instead of an exhaustive list of class names)

maltheism: Hey @driusan, the defaultProps will populate the "row form-group" as default if className isn't supplied. 
```
ButtonElement.defaultProps = {
  className: 'row form-group
}
```
The change made by @zaliqarosli has the ability to then add custom css class for when we don't want to use "row form-group" and a few cases around LORIS this is useful.

driusan: I know what the change she did it, I'm asking if that's the right change.

maltheism: @driusan I think it is the right change.

zaliqarosli: @driusan removing `row form-group` removes right, left, and bottom margins from the button. so those classes are desirable when you want to line up the button element with the rest of the elements in <FormElement />. However, there are times when you don't want the margins because you want to fit the ButtonElement in some where that is not a row in the form. Having said that though, I no longer need this case. And realized that we have the CTA (call to action) element that doesn't have formatting. so i think i'll close this PR

PR #5118, opened 2019-08-26
[Candidate Parameters] fix permission check

xlecours:
The requesting user must have candidate_parameter_edit, not just being authenticated.

Also fix `\Candidate::singleton((int) $candID);` call with a string. It wad generating an error : 
`PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Candidate::singleton() must be of the type integer, string given`

### DIFF
--- a/modules/candidate_parameters/ajax/formHandler.php
+++ b/modules/candidate_parameters/ajax/formHandler.php
@@ -13,36 +13,49 @@
  * @link     https://github.com/aces/Loris-Trunk
  */
 
-$user =& \User::singleton();
+$user = \User::singleton();
 if (!$user->hasPermission('candidate_parameter_edit')) {
     header("HTTP/1.1 403 Forbidden");
     exit;
 }
 
-if (isset($_POST['tab'] ?? '')) {
-    $tab = $_POST['tab'];
-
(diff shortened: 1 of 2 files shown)


modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -12,31 +12,50 @@
...
-    }
+
+$user = \User::singleton();
+if (!$user->hasPermission('candidate_parameter_edit')) {
+    header("HTTP/1.1 403 Forbidden");
+    exit;
 }
 
+$tab = $_POST['tab'] ?? '';
+if ($tab === '') {
```
maltheism: Hey @xlecours, why not do something similar to what you're doing in modules/candidate_parameters/ajax/getData.php? I think you could save 4 lines by removing the Bad Request check and just leave it for the default case in the switch statement. Otherwise shouldn't the Bad Request be in getData.php?

xlecours: @maltheism yes, good catch. I will add this to getData because there is a slight difference between missing a field (bad request) and having a value mapping to an nonexistent function (not found).

PR #5160, opened 2019-08-27
Enable shellcheck on all shell scripts

gdevenyi:
## Brief summary of changes

Enable shellcheck on all scripts inside loris

Find things like #5159 before they bite.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ checkstatic: phpdev
 	npm run lint:php
 	npm run lint:javascript
 	vendor/bin/phan
+	bash -c 'shopt -s globstar nullglob; GLOBIGNORE="vendor*:node_modules*"; shellcheck **/*.{sh,ksh,bash}'
 
 unittests: phpdev
 	vendor/bin/phpunit --configuration test/phpunit.xml


gdevenyi:
Followup, this is likely to find bugs.

PapillonMcGill:
shellcheck not installed neither on my dev VM nor my desktop. This would require to have it installed previously. 

also checking if bash is present while calling bash is a little bit redundant.

PapillonMcGill:
Travis failing due to an empty line before the #! in a vendor file :-(

gdevenyi:
I'll look into moving this further up in the command list so that those vendor files don't exist yet.

gdevenyi:
Weird, travis picked up the wrong commit. Can someone with power kill/restart the build?

PapillonMcGill:
@gdevenyi Done!

PapillonMcGill:
@gdevenyi ran it in my VM and it is effectively picking a lot of bugs (warning and error). I can take care of some errors but could you take care of solving this one:
```
In test/run-php-linter.sh line 6:
for i in `ls php/libraries/*.class.inc modules/*/php/* modules/*/ajax/* htdocs/*.php htdocs/*/*.php`;
         ^-- SC2045: Iterating over ls output is fragile. Use globs.
         ^-- SC2006: Use $(..) instead of legacy `..`.
```

gdevenyi:
Travis is weird, it still didn't get the current version of the code.

Fix for above is quick:
```
for i in php/libraries/*.class.inc modules/*/php/* modules/*/ajax/* htdocs/*.php htdocs/*/*.php;
```

gdevenyi:
In addition, its probably a good idea to do a recursive glob to find all the files:
```
shopt -s globstar
for i in **/*class.inc **/*php;
```

PapillonMcGill:
> ```
> shopt -s globstar
> for i in **/*class.inc **/*php;
> ```
is likely to pick files in vendor/ directory.

gdevenyi:
Weirdly, this still ran the old .travis.yml file.

gdevenyi:
Now ignoring vendor/ and node_modules/*

PapillonMcGill:
adding block as it require https://github.com/aces/Loris/pull/5204 to pass Travis

johnsaigle:
@gdevenyi Would you be able to add this to the `checkstatic` target in the `Make file`? Instead of directly in `.travis.yml`?

As you can see Travis will run `make checkstatic` so the command will still be a part of our CI. This will also give the benefit the options of developers being able to easily run the command locally.

.travis.yml
```diff
@@ -24,5 +24,6 @@ before_script:
 script:
 # Unable to initialize database, use the following code to debug. - docker-compose up --build
 # docker-compose up --build
+- bash -c 'shopt -s globstar nullglob; GLOBIGNORE="vendor*:node_modules*"; shellcheck **/*.{sh,ksh,bash}'
```
johnsaigle: @gdevenyi Would you be able to add this to the `checkstatic` target in the `Make file`? Instead of directly in `.travis.yml`?

As you can see on the line below, Travis will run `make checkstatic` so the command will still be a part of our CI. This will also give the benefit the options of developers being able to easily run the command locally.

cmadjar: @gdevenyi Would it be possible for you to address what @johnsaigle raise?

gdevenyi: Missed this request, done (hopefully edited the Makefile properly since I can't really test that command locally)

johnsaigle:
Hey @gdevenyi I created a new pull request based on these commits #6053. I took an approach whereby we can introduce one script at a time to be analyzed by shellcheck. It seems like nobody has the time to tackle all of these issues at once.

gdevenyi:
Close in favour of #6053

PR #5171, opened 2019-08-28
Add issue templates

driusan:
Added issue templates based on slight modifications of the defaults suggested by GitHub.

(This is identical to #5170, which GitHub automatically created from the UI, but this one is properly coming from my fork instead of polluting the branches on aces/Loris)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -24,7 +24,7 @@ A clear and concise description of what the expected behaviour was.
 A clear and concise description of what you expected to happen.
 
 **Browser Environment (please complete the following information):**
-- OS: [e.g. Plan9, HURD, etc]
+- OS: [e.g. Plan9, HURD, Oberon, etc]
 - Browser [e.g. chrome, safari]
 - Version [e.g. 22]
 
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,37 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: Bug Fix
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is
+
+**To Reproduce**
+Steps to reproduce the behavior (attach screenshots if applicable):
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**What happened?**
+A clear and concise description of what the expected behaviour was.
+
+**What did you expect to happen?**
+A clear and concise description of what you expected to happen.
+
+**Browser Environment (please complete the following information):**
+- OS: [e.g. Plan9, HURD, etc]
+- Browser [e.g. chrome, safari]
+- Version [e.g. 22]
+
+**Server Environment (if known):**
+- LORIS Version: [e.g. 21.0.0, 16.04]
+- Linux distribution and Version: [e.g. Ubuntu 16.04, CentOS 7]
+- MySQL/MariaDB Version: [e.g. MySQL 5.7, MariaDB 10.3]
+
+**Additional context**
+Add any other context about the problem here.
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,7 +2,7 @@
 name: Bug report
 about: Create a report to help us improve
 title: ''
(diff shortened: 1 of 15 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,37 @@
...
+4. See error
+
+**What happened?**
+A clear and concise description of what the behaviour was. 
+
+**What did you expect to happen?**
+A clear and concise description of what you expected to happen.
+
+**Browser Environment (please complete the following information):**
+- OS: [e.g. Plan9, HURD, etc]
```
PapillonMcGill: Not sure that the usual user would understand the example!
Maybe Linux, Ubuntu, Windows7, Windows10 would be more effective

xlecours: Oberon would be a great addition

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,37 @@
...
+- Browser [e.g. chrome, safari]
```
johnsaigle:
```suggestion
- Browser [e.g. chrome, firefox] (note that only these two browsers are supported)
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,37 @@
...
+- Version [e.g. 22]
+
+**Server Environment (if known):**
+- LORIS Version: [e.g. 21.0.0]
+- Linux distribution and Version: [e.g. Ubuntu 16.04, CentOS 7]
```
johnsaigle:
```suggestion
- Linux distribution and Version: [e.g. Ubuntu 18.04, CentOS 7]
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,37 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: Bug Fix
+assignees: ''
+
+---
+
```
johnsaigle: It might be nice to have something like

- [ ] I have searched for this bug on the Issues page and am convinced this is not a duplicate.

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: Feature
```
johnsaigle:
```suggestion
labels: Proposal
```

Or maybe it should be both?

johnsaigle:
@christinerogers 

> The "what happened" question - I can't think of an example where it's meaningfully distinct from the previous 2 questions. Thoughts?

I see it like this:

Describe the bug
> I tried to upload a file in document repository but it crashed and I couldn't do the thing I wanted

To reproduce
> 1. Go to document repository 2. Click Upload

What happened?
> I got an error message that said "..."

What did you expect to happen?
> To go to the file upload form


Maybe we could add a clarification message to the description under What happened? to say something like "If you got an error message, please paste it here"

.github/ISSUE_TEMPLATE/feature_request.md
driusan: This was just the default

.github/ISSUE_TEMPLATE/bug_report.md
driusan: This is a template, not a checklist

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: Bug Fix
```
maltheism: Does the template automatically fill the labels for the issue? I'm wondering because we have no label named "Bug Fix" but we do have the "Bug" label.

driusan: That's a good question. I don't know.

maltheism: Hey @driusan, apparently it does now as found [here](https://github.community/t5/How-to-use-Git-and-GitHub/Apply-labels-automatically-at-issue-creation/td-p/10719).
So maybe we should remove the "Fix" from "Bug Fix" for it to work.

driusan:
A couple people approved this and there doesn't seem to be any opposition to the idea at the moment, so I'm going to merge this and, if we don't like it we can revert later.

PR #5195, opened 2019-09-03
[install] add make into install.sh - make easy to install Loris.

kongtiaowang:
Users don't need to run make command.


kongtiaowang:
Users need to run make manually.

christinerogers:
note: this was closed because production and dev environment need different `make` arguments -- requires an instruction for admins rather than automation in the install script

Issue #5050, opened 2019-08-20
[install] add basic mkdir checks to install.sh 


christinerogers:
Add a check to `tools/install.sh` so that if the roles (sudo/lorisadmin) are wrong, script will abort instead of giving a success message. 
e.g. if `mkdir logs` failed, abort the script and tell the user.

PR #5204, opened 2019-09-05
[script] fix bugs indentified by shellcheck

PapillonMcGill:
## Brief summary of changes
fix 3 bugs indentified by running shellcheck (thanks @gdevenyi)

for tools/install.sh, run as if doing a new install
for run-php-linter.sh, run `npm run lint:php`
for integration.sh,  run as if doing integration test (`test/integration.sh`)

required by https://github.com/aces/Loris/pull/5160


test/run-php-linter.sh
```diff
@@ -3,7 +3,7 @@ set -euo pipefail
 
 # Run PHP -l on everything to ensure there's no syntax
 # errors.
-for i in `ls php/libraries/*.class.inc modules/*/php/* modules/*/ajax/* htdocs/*.php htdocs/*/*.php`;
+for i in php/libraries/*.class.inc modules/*/php/* modules/*/ajax/* htdocs/*.php htdocs/*/*.php;
 do
   php -l $i >/dev/null || exit $?;
```
gdevenyi: With "set -euo pipefail" there is no need for || exit $?, as any error in the script will trigger an exit with error from the script.

test/integration.sh
```diff
@@ -52,7 +52,7 @@ sed -i \
 export LORIS_DB_CONFIG=$(pwd)/config.xml
 
 #start PHP's built in webserver
-php -S localhost:8000 -t ../htdocs ../htdocs/router.php 2>&1 > /dev/null &
+php -S localhost:8000 -t ../htdocs ../htdocs/router.php /dev/null 2>&1 > &
```
driusan: What are you trying to do with this change?

PapillonMcGill: put the redirection of stderr to stdout in the right position. It came as an error when Travis run on. Oh wait, I have the and '>' misplaced.

PapillonMcGill:
@ridz1208 Could you approve so @driusan can merge?

tools/install.sh
```diff
@@ -160,19 +160,19 @@ os_distro=$(hostnamectl |awk -F: '/Operating System:/{print $2}'|cut -f2 -d ' ')
 debian=("Debian" "Ubuntu")
 redhat=("Red" "CentOS" "Fedora" "Oracle") 
 
-if [[ " ${debian[*]} " =~ " $os_distro " ]]; then
-    mkdir -p ../modules/document_repository/user_uploads
```
johnsaigle: Why are these `-p` flags removed?

PapillonMcGill: Some mistake during rebase probably.
Will add them back

PapillonMcGill: Where probably removed by the rebase. 
They are not needed as the individual modules repositories already exist and it just create 1 level of extra directories.

johnsaigle: They are needed because `user_uploads/` is not tracked by git. #5245 #5050 were issued to add `-p`.

Please add them back.

PapillonMcGill: '-p' is required when you want to create multiple level of subdir in one instruction. In this case, both '../modules/document_repository/' and '../modules/data_release/' already exist as they are managed by Git. Thus, only one level of subdir need to be create 'user_uploads', '-p' is not required

PapillonMcGill:
@ridz1208 ping

tools/install.sh
```diff
@@ -160,7 +160,7 @@ os_distro=$(hostnamectl |awk -F: '/Operating System:/{print $2}'|cut -f2 -d ' ')
 debian=("Debian" "Ubuntu")
 redhat=("Red" "CentOS" "Fedora" "Oracle") 
 
-if [[ " ${debian[*]} " =~ " $os_distro " ]]; then
+if [[ " ${debian[*]} " == *" $os_distro "* ]]; then
```
driusan: These changes are far less readable and understandable. What are you trying to solve with them?

PapillonMcGill: because shellcheck (see #5160) fail without these change.

johnsaigle: @driusan The shellcheck tool is giving an error about the regex operator `=~` being used with quotation marks. The quotation marks mean that the RHS will match literally rather than as a regex, e.g. that if the string contains `^` this will be treated literally rather than as a metacharacter.

Instead of treating this expression as a regex, this change does a more basic substring check. The wildcards are necessary or it doesn't work.

Another solution is to unquote the variable but still use the regex operator, e.g.:

```bash
if [[ " ${debian[*]} " =~ $os_distro ]]; then
```

However since we're not actually using a regex here, the substring comparison is more appropriate.

johnsaigle: And for the record it seems to be a fairly common bash idiom from what I'm finding on Google, though I agree with you it's a bit hard to parse.

PR #5203, opened 2019-09-04
[install README CentOS and Ubuntu] 21 install updates, including make - bugfix

kongtiaowang:
Add this [Run make (for production instances) or make dev (for development sandboxes). ] into readme

to help solve issue #5197
related to #5198

This PR updates the CentOS install readme for the LORIS 21 release which is out already

Resolves #3946

Issue #5197, opened 2019-09-03
[install] tell users to run make in Readme, update wiki


christinerogers:
per LORIS meeting discussion today: 
To install LORIS 21+, `make` needs to be run -- one way for dev installs, another way for production installs. 
To do: 
- [x] update the main Readme.md (Ubuntu) and the CentOS Readme on `bugfix` (#5203) 
- [x] Update wiki install in depth : add make command(s) at the end 
- [x] Update wiki install in brief : add/mention make command(s) at the end 

(#5195 added and closed by shen)

christinerogers:
#5203 merged.  closing this.

PR #5205, opened 2019-09-05
[install.sh] Cleanup for 21: dependency checks, logic, user messages

kongtiaowang:
Major cleanup for the core install script, on the `bugfix` branch
remove `composer install` check and tell users to do `make` next
It will help users to know what to do next.

to help solve issue #5197
related to PR #5203


tools/install.sh
```diff
@@ -249,7 +251,7 @@ while true; do
                 < ../docs/config/apache2-site | sudo tee /etc/httpd/conf.d/$projectname.conf > /dev/null
 
             sudo service httpd restart
-            echo "You may need to manually uncomment the load rewrite module line of your conf."
+            echo "You may need to manually uncomment the load rewrite module line of your apache conf."
```
PapillonMcGill:
```suggestion
            echo "You may need to manually uncomment the load rewrite module line of your apache configuration."
```

christinerogers: thanks @PapillonMcGill  I committed these changes.

tools/install.sh
```diff
@@ -262,4 +264,9 @@ else
     exit 1
 fi
 
-echo "Installation complete."
+echo "The installation of LORIS software components is now complete."
+echo "(For future troubleshooting assistance, you may wish to copy and save the output from this script.)"
+echo "Next steps: "
+echo "- Run 'make' (or 'make dev') from inside your $RootDir folder."
+echo "- Verify/enable your apache configuration and restart apache"
+echo "- Open Chrome (or Firefox) to <loris-url>/installdb.php to continue installing the database."
```
```suggestion
echo "- Navigate  to <loris-url>/installdb.php using a supported web browser (Chrome or Firefox) to continue installing the database."
```

driusan: @kongtiaowang ping

tools/install.sh
```diff
@@ -1,10 +1,12 @@
...
-# This will:
-#   1. Set up the LORIS DB schema
-#   2. Log the installation in the logs directory
-# This will only install the database components and LORIS config file.
+# This script will:
+#   1. verify composer is installed to manage dependencies
+#   2. Set up the directory structure with appropriate permissions
+#   3. Log the installation in the logs directory
+#   4. Configure apache (optional)
+# This will only install the software components and directory structure.
```
christinerogers: This line should be moved above the points, and changed to say `This script verifies key libraries and sets up the directory structure, as the first part of the LORIS installation.

tools/install.sh
```diff
@@ -1,10 +1,12 @@
...
 
 #
-# This will:
-#   1. Set up the LORIS DB schema
-#   2. Log the installation in the logs directory
-# This will only install the database components and LORIS config file.
+# This script will:
+#   1. verify composer is installed to manage dependencies
+#   2. Set up the directory structure with appropriate permissions
+#   3. Log the installation in the logs directory
```
christinerogers: To do: change "the installation" to "this script's output"

tools/install.sh
```diff
@@ -1,10 +1,12 @@
 #!/bin/bash
 
 #
-# This will:
-#   1. Set up the LORIS DB schema
-#   2. Log the installation in the logs directory
-# This will only install the database components and LORIS config file.
+# This script will:
+#   1. verify composer is installed to manage dependencies
```
christinerogers: This needs an edit given that `make` (run after this script) is the step that *installs* composer

tools/install.sh
```diff
@@ -262,4 +264,9 @@ else
     exit 1
 fi
 
-echo "Installation complete."
+echo "The installation of LORIS software components is now complete."
```
christinerogers: needs to be changed since the installation of software components happens *after* during `make`

kongtiaowang:
@christinerogers  Thanks for adding more accurate details and explanation.

christinerogers:
> @christinerogers Thanks for adding more accurate details and explanation.

Ok i'll make the commits directly, and dismiss my review.  
The script will need re-testing after.

tools/install.sh
christinerogers: done : see my [commit 475ca8](https://github.com/aces/Loris/pull/5205/commits/475ca805073aba4296b34ac0698d3e16e4a5d3fa)

tools/install.sh
christinerogers: - [x] done : see my [commit 475ca8](https://github.com/aces/Loris/pull/5205/commits/475ca805073aba4296b34ac0698d3e16e4a5d3fa)

tools/install.sh
christinerogers: - [x] done : see my [commit 475ca8](https://github.com/aces/Loris/pull/5205/commits/475ca805073aba4296b34ac0698d3e16e4a5d3fa)

tools/install.sh
christinerogers: - [x] done : see my [commit 475ca8](https://github.com/aces/Loris/pull/5205/commits/475ca805073aba4296b34ac0698d3e16e4a5d3fa)

tools/install.sh
```diff
@@ -192,18 +181,14 @@ if [ -d logs ]; then
...
 cd tools
 
-
 if [[ " ${debian[*]} " =~ " $os_distro " ]]; then
 echo "Ubuntu distribution detected."
     # for CentOS, the log directory is called httpd
     logdirectory=/var/log/apache2
     while true; do
-        read -p "Would you like to automatically create/install apache config files? (Works for Ubuntu 14.04 or later default Apache installations) [yn] " yn
+        read -p "Would you like to automatically set up your apache configuration files? [yn] " yn
```
christinerogers: this is where the projectname should get asked, i think.   I can make an edit but it will be a big one. @kongtiaowang

tools/install.sh
```diff
@@ -72,38 +78,21 @@ else
...
-Please ensure you have the following information ready (if applicable):
-
-  1) Your project directory name.
-     (Will be used to modify the paths for Imaging data in the generated
-     config.xml file for LORIS, and may also be used to automatically
-     create/install apache config files.) If unsure, a default like "LORIS"
-     should be acceptable.
+This install script will ask you to provide your project name. 
+If unsure, we suggest using "LORIS"
+(This will be used to define Imaging data filepaths automatically in a config.xml file for LORIS, and (optionally) for the apache configuration file.)
```
christinerogers: I'm pretty sure this statement could be changed to "This will be used IFF you want this script to configure your apache"

tools/install.sh
kongtiaowang: @christinerogers You are right.

tools/install.sh
```diff
@@ -72,38 +78,21 @@ else
...
-    echo "PHP Composer appears to be installed."
-    composer_scr="./composer install --no-dev"
-else
-    echo ""
-    echo "PHP Composer does not appear to be installed. Please install it before running this script."
-    echo ""
-    echo "(e.g. wget https://getcomposer.org/installer"
-    echo "php installer --install-dir=/usr/local/bin --filename=composer)"
-    echo "while having root permission)";
+PHP_VERSION=$(php -v | head -n 1 | cut -d " " -f 2 | cut -f1-2 -d".")
```
johnsaigle: This may be a weak way to test the PHP version given that the CLI version of PHP can differ from the version Apache is configured to run with.

Not blocking this PR but I'm going to create an issue so that someone creates a more robust test in the future.

johnsaigle:
@christinerogers Sorry for catching this late in the process but we probably should re-add the composer check deleted by your commit.

While users do not need to run `composer` directly, it is called by `make`. There's still value in having the check because a user won't be able to proceed without it.

christinerogers:
@johnsaigle I just restored the composer check, omitting lines that used to set up the later execution of `composer install` in this script.  Could you re-review? 
Thanks for catching this...

tools/install.sh
christinerogers: any updates on `$projectname` cleanup, @kongtiaowang ?

christinerogers:
@PapillonMcGill could you re-review ?

tools/install.sh
```diff
@@ -180,7 +137,12 @@ elif [[ " ${redhat[*]} " =~ " $os_distro " ]]; then
     sudo chgrp apache ../project
     sudo chmod 770 ../project
 else
-    echo "$os_distro Linux distribution detected. We currently do not support this. Please manually chown/chgrp to the web server user in: the user_uploads directory in ../modules/data_release/ and ../modules/document_repository/, as well as ../smarty/templates_c/"
+    echo "$os_distro Linux distribution detected. We currently do not support this. "
+    echo "Please manually run chown/chgrp to the web server user, the following subdirectories: "
```
PapillonMcGill:
```suggestion
    echo "Please manually change ownership and permission to let the web server read and write in  the following subdirectories: "
```

tools/install.sh
```diff
@@ -83,8 +83,7 @@ echo ""
 cat <<QUESTIONS
 This install script will ask you to provide your project name. 
 If unsure, we suggest using "LORIS"
-(This will be used to define Imaging data filepaths automatically in a config.xml file for LORIS, and (optionally) for the apache configuration file.) 
-
+(This will be used IFF you want this script to configure your apache)
```
driusan: I wouldn't expect most users running a script to know the meaning of the word "iff".

christinerogers: Hi @driusan - that line was removed yesterday. Could you Update and re-review?  
(Or I can dismiss for you if you like)

christinerogers:
@driusan : Please re-review, this the key PR for 21.0.* install 

cc @ridz1208 
As discussed previously, 21.0.3 bugfix release can be released asap after. 

I can contribute to the 21.0.3 release notes as needed.

tools/install.sh
```diff
@@ -1,10 +1,10 @@
...
 
 #
-# This will:
-#   1. Set up the LORIS DB schema
-#   2. Log the installation in the logs directory
-# This will only install the database components and LORIS config file.
+# This script will:
+#   1. Install the directory structure with appropriate permissions
+#   2. Configure apache (optional)
+#   3. Log these steps in the logs/ directory (if writable)
```
```suggestion
#   3. Log these steps in the logs/ directory
```
This script creates the `logs/` directory so I don't see why it wouldn't be writable.

tools/install.sh
```diff
@@ -28,11 +28,11 @@ exec 1>$LOGPIPE 2>&1
 CWD=`pwd`
 RootDir=`dirname $CWD`
 
-
 echo "LORIS Installation Script starting at $START"
 
 if [ ! -w $LOGDIR ] ; then
```
johnsaigle: As I noted above I'm not sure why the `tools/logs/` directory wouldn't be writable since it's created by the install script. However, we can punt this to another issue so that this PR isn't delayed anymore.

tools/install.sh
```diff
@@ -238,10 +206,21 @@ echo "CentOS distribution detected."
 # for CentOS, the log directory is called httpd
 logdirectory=/var/log/httpd
 while true; do
-    read -p "Would you like to automatically create/install apache config files? (In development for CentOS 6.5) [yn] " yn
+    read -p "Would you like to automatically set up your apache configuration files? (In development for CentOS) [yn] " yn
```
johnsaigle: > (In development for CentOS)

Is this true? Does this work or not? It would be good to have an issue specifically for this.

PapillonMcGill: Not sure if someone is developing it but Ubuntu and CentOS have different basic config using different file structure. So change applied by the script might be overridden by default configuration due to order of appearance.

johnsaigle: OK well it sounds like further testing should be done to make sure that Apache can be automatically configured by the install script for CentOS. I'll make an issue for it.

PR #5208, opened 2019-09-06
[jsx/Form.js] Make rendering label for SelectElement and TextboxElement optional

zaliqarosli:
## Brief summary of changes

This PRs allows the ability to render SelectElements and TextboxElements without its label. If a label prop is provided, the label will render and take up 3 spaces in the row, with the input element taking up the rest of the 9 spaces. This is also the case if the label prop is an empty string. If a label prop is not provided at all, the default is that the label won't be rendered at all and the input element takes up all 12 spaces of the row.

This change is useful for modules such as login that still need to be reactified and have input elements without labels:

Screenshot 2019-09-06 13 52 44

This PR pulls out some code from https://github.com/aces/Loris/pull/4395

1. This change shouldn't affect any existing SelectElements and TextboxElements in loris. Browse different modules, and see if any of them are missing labels or have weird formatting. Make sure you run `make` before testing

### DIFF
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -938,7 +938,7 @@ class TextboxElement extends Component {
           {requiredHTML}
         </label>
       );
-      inputclass = 'col-sm-9';
+      inputClass = 'col-sm-9';
     }
 
     return (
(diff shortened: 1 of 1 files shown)


HenriRabalais:
I like this! I think this will also generally be useful for Mobile compatibility and making forms more compact!

johnsaigle:
@HenriRabalais Do you wanna give this an approval + Passed Manual Tests then?

Issue #5061, opened 2019-08-21
Data release reactification


cmadjar:
WIP

cmadjar:
PR still in review phase but getting closer than ever to be merged :).

PR #5224, opened 2019-09-13
[Data release] Reactification of the module

### LINKED ISSUE #5843: [Data Release] File upload fails silently for large files
Back-end log:

>[Fri Dec 06 16:10:48.196409 2019] [php7:warn] [pid 28727] [client 192.168.122.1:60390] PHP Warning:  POST Content-Length of 56173016 bytes exceeds the limit of 8388608 bytes in Unknown on line 0, referer: https://jsaigle-dev.loris.ca/data_release/
[Fri Dec 06 16:10:48.274852 2019] [php7:notice] [pid 28727] [client 192.168.122.1:60390] PHP Notice:  Undefined index: action in /var/www/loris/modules/data_release/ajax/FileUpload.php on line 18, referer: https://jsaigle-dev.loris.ca/data_release/
[Fri Dec 06 16:10:48.724921 2019] [php7:notice] [pid 24309] [client 192.168.122.1:60392] PHP Notice:  Trying to get property 'page' of non-object in /var/www/loris/src/Middleware/UserPageDecorationMiddleware.php on line 92, referer: https://jsaigle-dev.loris.ca/data_release/ajax/FileUpload.php

The module should be updated with a proper failure message if the uploaded file exceeds the max file size allowed by the server. 

To replicate, try uploading a large file.

cmadjar:
### Description

This PR includes the code changes needed to reactify the data release module. 

### This resolves

- Resolves https://github.com/aces/Loris/issues/5061
- Resolves #5843 

### To test this

- follow the test plan. The functionality of the module should not have been changed.


cmadjar:
@johnsaigle @kongtiaowang I am looking at the code and I don't see why Travis complains about an unexpected closing bracket on line 91. I looked both in VI and PHPSTORM and the bracket should be there to close the pselect statement. Am I missing something obvious?

cmadjar:
Nevermind. I guess it was the comma before the closing bracket...

modules/data_release/ajax/AddPermission.php
```diff
@@ -20,9 +20,16 @@
 $baseURL  = $settings->getBaseURL();
 $DB       = $factory->database();
 
-if (isset($_POST['action']) && $_POST['action'] == 'addpermission'
+if (isset($_GET['action']) && $_GET['action'] == 'addpermission'
```
johnsaigle: Do you think you could rewrite this in the "happy path" format?

Basically it should fail early if `'action'` is not set and from there proceed knowing that the value is good (maybe with a switch-case instead of if-else).

modules/data_release/ajax/AddPermission.php
```diff
@@ -183,9 +196,52 @@
...
     header("HTTP/1.1 400 Bad Request");
     echo "There was an error adding permissions";
 }
 
+/**
+ * Gets the data for the manage permission modal window.
+ *
+ * @return void
+ */
+function getManagePermissionsData()
```
johnsaigle:
```suggestion
function getManagePermissionsData(): void
```

modules/data_release/ajax/FileUpload.php
```diff
@@ -15,11 +15,11 @@
 $DB   = \Database::singleton();
 $user = \User::singleton();
 
-if ($_POST['action'] == 'upload'
+if ($_GET['action'] == 'upload'
```
johnsaigle: Same here, maybe fail early if `'action'` param is bad.

modules/data_release/ajax/FileUpload.php
```diff
@@ -15,11 +15,11 @@
...
     && $user->hasPermission("data_release_upload")
 ) {
     $fileName    = $_FILES["file"]["name"];
-    $version     = $_POST['version'];
+    $version     = $_POST['version'] ? $_POST['version'] : null;
```
johnsaigle:
```suggestion
    $version     = $_POST['version'] ?? null;
```

johnsaigle: Could you make this check more explicit, e.g. if it's supposed to be not empty, or an int, etc.?

This statement will for example change a value of `0` to null. Maybe this is correct behaviour but it's hard to know.

modules/data_release/ajax/FileUpload.php
```diff
@@ -51,23 +52,29 @@
                  'upload_date' => $upload_date,
                 )
             );
-
+            // grep the ID of the user who uploaded the file
```
johnsaigle: Why "grep"? It's just a table select right?

modules/data_release/php/data_release.class.inc
```diff
@@ -51,70 +54,87 @@ class Data_Release extends \NDB_Menu_Filter
...
-            "SELECT ID, UserID FROM users",
-            array(),
-            "ID"
-        );
+    /**
+     * Create the form for the data release menu page
+     *
+     * @return void
+     **/
+    function setup(): void
```
johnsaigle:
```suggestion
    function setup()
```

I love return types... but this will throw an error since the function signature for `NDB_Menu_Filter::setup()` does not have a return type.

modules/data_release/php/data_release.class.inc
```diff
@@ -51,70 +54,87 @@ class Data_Release extends \NDB_Menu_Filter
...
+    {
+        parent::setup();
+
+        // Get list of users in the database
+        $list_of_users = $this->getUsersList();
+        $users_options = array();
+        foreach ($list_of_users as $id => $name) {
+            $users_options[$id] = $name;
```
johnsaigle: Isn't this equivalent to `$list_of_users`?

modules/data_release/ajax/AddPermission.php
cmadjar: should be done in the next commit.

cmadjar:
@johnsaigle Thank you for your review. I integrated your feedback. Feel free to re-review at your convenience. Thank you!!

modules/data_release/ajax/FileUpload.php
johnsaigle: Yep that's good

modules/data_release/ajax/AddPermission.php
```diff
@@ -20,9 +20,27 @@
 $baseURL  = $settings->getBaseURL();
 $DB       = $factory->database();
 
-if (isset($_POST['action']) && $_POST['action'] == 'addpermission'
-    && $user->hasPermission('data_release_edit_file_access')
-) {
+
+if (!isset($_GET['action'])) {
+    header('');
```
johnsaigle:
```suggestion
    http_response_code(400);
```

This should probably return a Bad Request if there is no action.

modules/data_release/ajax/AddPermission.php
```diff
@@ -20,9 +20,27 @@
...
+
+if (!$user->hasPermission('data_release_edit_file_access')) {
+    header("HTTP/1.1 403 Forbidden");
+    exit;
+}
+
+
+if ($_GET['action'] == 'addpermission') {
+
+    if (!isset($_POST) || empty($_POST)) {
```
johnsaigle: It should be either a GET or a POST right? I don't think both of these arrays will have data at the same time.

modules/data_release/ajax/FileUpload.php
```diff
@@ -15,11 +15,30 @@
...
+        . " does not appear to be writeable."
+    );
+    header("HTTP/1.1 500 Internal Server Error");
+    exit;
+}
+
+if ($_GET['action'] == 'upload') {
     $fileName    = $_FILES["file"]["name"];
-    $version     = $_POST['version'];
+    $version     = $_POST['version'] ? !empty($_POST['version']) : null;
```
johnsaigle: I don't think this is correct. The value to the right of the `?` will be `true` or `false` (the `bool` return value of `empty()`. `$version` is probably not supposed to be a bool.

modules/data_release/ajax/AddPermission.php
cmadjar: With Ajax you can send data to a PHP script via GET and POST in the same request. 

From: https://coursesweb.net/ajax/get-post

Since this is an ajax request, I guess it is fine. (it is definitely working when I test it I think)

modules/data_release/ajax/FileUpload.php
```diff
@@ -28,59 +47,68 @@
...
+        // add permission to the user for the uploaded data_release file
+        $DB->insert(
+           'data_release_permissions',
+            array(
+             'userid'          => $user_ID,
+             'data_release_id' => $ID,
+            )
+        );
+    }
+    header("Location: {$baseURL}/data_release/?uploadSuccess=true");
```
johnsaigle: It could be good to add a `201 Created` header here.

modules/data_release/ajax/FileUpload.php
cmadjar: oooops. I did not put the function empty in the right place...

modules/data_release/ajax/AddPermission.php
johnsaigle: Hmmm okay. I don't think this is good practice but it can be handled in another request.

I believe the correct way to use GET is that it should not modify anything on the server where POST will. It's better to keep these methods separate instead of mixing them.

modules/data_release/ajax/FileUpload.php
cmadjar: added in the next commit

modules/data_release/ajax/FileUpload.php
cmadjar: done in the last commit

modules/data_release/ajax/FileUpload.php
```diff
@@ -28,59 +47,69 @@
...
+    header("201 Created");
```
johnsaigle:
```suggestion
    header("HTTP/1.1 201 Created");
```

Alternatively, `http_response_code(201)`

modules/data_release/ajax/AddPermission.php
johnsaigle: That makes sense. Just funny to mix them.

modules/data_release/php/data_release.class.inc
```diff
@@ -51,70 +54,83 @@ class Data_Release extends \NDB_Menu_Filter
...
-                $data_release_ids[$fileID] = $fileName . " - " . $versionName;
-            }
-            $data_release_versions[$versionName] = $versionName;
+        // Get list of data release versions and filenames
+        $versionedFiles = $this->getVersionedFiles();
 
+        $filenames_options = array();
+        $versions_options  = array();
+
+        foreach ($versionedFiles as $versionName => $versionFiles) {
```
johnsaigle: Not a blocking change but would you mind changing these variable names? I know you didn't write them in the first place but it looks like it's iterating over the same array in both loops because the names are odd.

cmadjar: how about I rename `$versionFiles` to `$filesArray`? Would that be less confusing?

I could also rename `$versionedFiles` to `$filesOrganizedPerDataReleaseVersions`?

Any better idea?

johnsaigle: `$vFiles` is used instead of `$versionedFiles` elsewhere in this module so maybe that?

I don't really mind as long as the words look less similar. Thinking of last week's clean code ADM :)

cmadjar: Done! Thank you!

jesscall:
Did a quick manual test, looks good!

modules/data_release/php/data_release.class.inc
```diff
@@ -51,70 +54,83 @@ class Data_Release extends \NDB_Menu_Filter
...
      *
-     * @param \Database $DB Database to use for querying
-     *
      * @return array Associative array of $versions=>(files associate with version)
      */
-    function getVersionedFiles(\Database $DB)
+    function getVersionedFiles()
     {
+        $Factory = \NDB_Factory::singleton();
+        $DB      = $Factory->Database();
```
driusan: This looks like a step backwards

modules/data_release/php/data_release.class.inc
```diff
@@ -51,70 +54,83 @@ class Data_Release extends \NDB_Menu_Filter
...
+                              ];
+    }
+
+
+    /**
+     * Greps the list of users available in the users database table.
+     *
+     * @return array Array of user ID and user name
+     */
+    function getUsersList()
```
driusan: This should also take a \Database as a parameter instead of using NDB_Factory

modules/data_release/jsx/uploadFileForm.js
```diff
@@ -0,0 +1,218 @@
...
+    let formData = this.state.formData;
+    let formObj = new FormData();
+    for (let key in formData) {
+      if (formData[key] !== '') {
+        formObj.append(key, formData[key]);
+      }
+    }
+
+    // ajax call to upload the file
+    $.ajax({
```
driusan: Can this use the fetch API instead of jQuery?

modules/data_release/jsx/uploadFileForm.js
```diff
@@ -0,0 +1,218 @@
...
+    };
+
+    this.updateFormElement = this.updateFormElement.bind(this);
+    this.validateAndSubmit = this.validateAndSubmit.bind(this);
+    this.uploadFile = this.uploadFile.bind(this);
+  }
+
+  componentDidMount() {
+    let self = this;
+    $.ajax(this.props.DataURL, {
```
driusan: fetch

modules/data_release/jsx/managePermissionsForm.js
```diff
@@ -0,0 +1,198 @@
...
+    let myFormData = JSON.parse(JSON.stringify(this.state.formData));
+    let formObj = new FormData();
+    for (let key in myFormData) {
+      if (myFormData[key] !== '') {
+        formObj.append(key, JSON.stringify(myFormData[key]));
+      }
+    }
+
+    // ajax call to update the permission
+    $.ajax({
```
driusan: fetch api

modules/data_release/jsx/addPermissionForm.js
```diff
@@ -0,0 +1,224 @@
...
+      }
+    }
+
+    // validate the form data
+    if (!this.isValidFormData()) {
+      return;
+    }
+
+    // ajax call to update the permission
+    $.ajax({
```
driusan: fetch api

modules/data_release/php/data_release.class.inc
cmadjar: done

modules/data_release/php/data_release.class.inc
cmadjar: oh noooo. I prefer moving forward too. Changed it to \Database::singleton() (guessing this is the forward move?)

modules/data_release/jsx/uploadFileForm.js
cmadjar: Sure! Is that enough balls? LOL

image

modules/data_release/jsx/uploadFileForm.js
cmadjar: Is there an example somewhere that use the fetch API? Not sure how to do that...

modules/data_release/php/data_release.class.inc
johnsaigle: I think Dave meant that this function should take a Database as a parameter rather than using the factory or singleton.

johnsaigle: I'm not sure why this is better though, haha

modules/data_release/jsx/uploadFileForm.js
driusan: There are other places in this PR that you used it.. ie. modules/data_release/jsx/addPermissionForm.js line 35.

For POST requests, I know @HenriRabalais looked up how to do it a while ago but I can't remember in which module.. @HenriRabalais do you have an example to point @cmadjar to?

cmadjar:
@johnsaigle lol! I am doing the rebase. You are too fast ;)

johnsaigle:
Travis is failing with PHPCS errors even though its build status displays as pending.

modules/data_release/jsx/managePermissionsForm.js
```diff
@@ -0,0 +1,198 @@
...
+
+  /**
+   * Retrieve data from the provided URL and save it in state
+   *
+   * @return {boolean}
+   */
+  fetchData() {
+    let self = this;
+
+    return $.ajax(this.props.DataURL, {
```
HenriRabalais: fetch api

modules/data_release/jsx/uploadFileForm.js
HenriRabalais: @cmadjar  The media module is using fetch api https://github.com/aces/Loris/blob/master/modules/media/jsx/mediaIndex.js

cmadjar: @HenriRabalais Thank you! I will look into it. :)

modules/data_release/jsx/uploadFileForm.js
```diff
@@ -0,0 +1,196 @@
...
+    let formData = this.state.formData;
+    let formObj = new FormData();
+    for (let key in formData) {
+      if (formData[key] !== '') {
+        formObj.append(key, formData[key]);
+      }
+    }
+
+    // fetch API to upload the file
+    fetch(this.props.action, {
```
cmadjar: @HenriRabalais @driusan I still have an issue with that fetch for the file upload.

The file name is correctly passed to the PHP side, however, the rest of the $_FILES['file'] object sent to the PHP side is filled with '\n'... Including the 'tmp_name' key which is then used by the PHP side to move the file to the user_uploads directory of the module.

Any idea what I am doing wrong? I googled a lot but could not find the solution on my own. Your help with this would be greatly appreciated :). Thanks!

cmadjar: PS: all the other ajax requests have been replaced by `fetch` BTW. This is the only one giving me trouble...

HenriRabalais: @cmadjar perhaps i can come take a look at it offline next week

cmadjar: @HenriRabalais As you probably guessed by now, I was away ;). If you are back this week, maybe we could take a look at it together tomorrow (Tuesday)?

Thank you and happy new year!!

modules/data_release/jsx/uploadFileForm.js
```diff
@@ -0,0 +1,215 @@
...
+      hasError.Filename = true;
+      this.setState({errorMessage, hasError});
+      return;
+    }
+
+    // Check that the size of the file is not bigger than the allowed size
+    let fileSize = formData.file ? Math.round((formData.file.size/1024)) : null;
+    const maxSizeAllowed = this.state.data.maxUploadSize;
+    if (parseInt(fileSize, 10) > parseInt(maxSizeAllowed, 10)*1024) {
+      let msg = 'File size exceeds the maximumn allowed (' + maxSizeAllowed + ')';
```
johnsaigle:
```suggestion
      let msg = 'File size exceeds the maximum allowed (' + maxSizeAllowed + ')';
```

modules/data_release/jsx/uploadFileForm.js
```diff
@@ -0,0 +1,215 @@
...
+        text: 'A file with this name already exists!\n Would you like to override existing file?',
+        type: 'warning',
+        showCancelButton: true,
+        confirmButtonText: 'Yes, I am sure!',
+        cancelButtonText: 'No, cancel it!',
+      }, function(isConfirm) {
+        if (isConfirm) {
+          this.uploadFile();
+        } else {
+          swal('Cancelled', 'Your imaginary file is safe :)', 'error');
```
johnsaigle:
```suggestion
          swal('Cancelled', 'Your file is safe :)', 'error');
```

driusan: But are my imaginary files in danger now? :(

johnsaigle: Not if you have imaginary backups

cmadjar:
@HenriRabalais @johnsaigle I believe this PR is ready for another review now that it finally passes Travis :). Thank you!

modules/data_release/ajax/FileUpload.php
```diff
@@ -17,68 +17,100 @@
...
+    header("HTTP/1.1 500 Internal Server Error");
+    exit;
+} elseif (!is_writable(__DIR__ . "/../user_uploads/")) {
+    error_log(
+        "File upload failed. Default user_uploads directory"
+        . " does not appear to be writeable."
+    );
+    header("HTTP/1.1 500 Internal Server Error");
+    exit;
+}
```
johnsaigle: I think this can be deleted now that #5815 is merged. The default upload path is `modules/data_release/user_uploads` which will be checked below

modules/data_release/test/TestPlan.md
```diff
@@ -13,8 +15,8 @@ Ensure that regular users can see the data release module without the
  
 ## Permissions
 
-1. Give the user the `data_release_edit_file_access` *(Data Release: Give user 
-permission to view files)* permission.
+1. Give the user the `data_release_edit_file_access` *(Data Release: Grant other
```
johnsaigle: I think the language here should match the README and give the proper name of the permissions.

I think it's `data_release_permissions` which is on a per-file basis? But I'm not sure. I'm not very familiar with the module.

cmadjar: actually, this is the correct permission name. `data_release_permissions` is the table where the user permissions for data releases are stored.

cmadjar:
@johnsaigle I integrated your comments. Thanks!

modules/data_release/php/data_release.class.inc
```diff
@@ -26,7 +26,11 @@ namespace LORIS\data_release;
 
 class Data_Release extends \NDB_Menu_Filter
 {
+    public $AjaxModule   = true;
+    public $skipTemplate = true;
+
     /**
+<<<<<<< HEAD
```
johnsaigle:
```suggestion
```

johnsaigle: Leftover from rebase/merge

cmadjar:
@johnsaigle Thank you for testing!!

1. I fixed the sweet alert window so that it displays the appropriate error message if there are any for the upload file form.

2. a user needs at minimum to have the `data_release_view` permission in order to access the module so I modified the test plan accordingly.

3. add permission and manage permission submission are fixed

Finally:
> it would be good if the upload file dialog closed after a successful upload. Currently the user has to click "OK" on the sweet alert and also manually close the Upload File modal behind it.

Agreed... Can't seem to figure out how to do that though. @HenriRabalais any idea?

modules/data_release/php/data_release.class.inc
cmadjar: oooops ;)

cmadjar:
@johnsaigle good catch! I fixed the link bug in the data table for users other than superusers. Hopefully I fix Travis too...

Regarding the Browse button, I am not able to reproduce the issue you are having. I tested on both Chrome and Firefox and it works on my sandbox. Can you retry maybe?

cmadjar:
@johnsaigle @kongtiaowang Travis keeps failing but I don't understand why. Any idea?

modules/data_release/ajax/AddPermission.php
```diff
@@ -182,10 +204,51 @@
...
 
+    $DB     = \Database::singleton();
+    $users  = $data_release->getUsersList($DB);
+    $vFiles = $data_release->getVersionedFiles($DB);
+    $prePermissions = $data_release->getUserVersionPermissions($vFiles, $DB);
+
+    $data_table_values = [];
+    $index = 0;
+    foreach ($users as $userid => $username) {
+        foreach ($vFiles as $versionName => $versionFiles) {
```
johnsaigle: Travis is having an issue with these two lines.

>/home/travis/build/aces/Loris/modules/data_release/ajax/AddPermission.php:237	Avoid unused local variables such as '$userid'.
/home/travis/build/aces/Loris/modules/data_release/ajax/AddPermission.php:238	Avoid unused local variables such as '$versionFiles'.

Since you're not using `$userid`, you can change the first loop to be:
`foreach ($users as $username) {`
PHP iterates over the values of an array by default.

For the next one you're not using `$versionFiles`. In this case you can just iterate over the keys of the array:
`foreach (array_keys($vFiles) as $versionName) {`

cmadjar: @johnsaigle Thank you! Resolved in the last commit.
I am a bit confused though because I thought Travis failed when running npm which is for JS file, no?
Anyway, thanks for your help!

johnsaigle: `npm` is usually associate with JS but we have it configured to run other scripts for us.

If you look in `package.json` you'll see references to things like `npm run tests:unit` which in turn actually point to PHP or bash scripts. 

I'm not exactly sure why we do this but... that's how it is.

modules/data_release/jsx/uploadFileForm.js
```diff
@@ -0,0 +1,224 @@
...
+    };
+
+    if (!formData.file) {
+      errorMessage.Filename = 'You must select a file to upload';
+      hasError.Filename = true;
+      this.setState({errorMessage, hasError});
+      return;
+    }
+
+    // Check that the size of the file is not bigger than the allowed size
```
cmadjar: @johnsaigle re-#5843: yes, that PR adds the check for file size for the data release module.

cmadjar:
@johnsaigle I just tried a few times what you described and it works on my sandbox. Is it possible you granted the permission to the file to the wrong user?

modules/data_release/test/TestPlan.md
```diff
@@ -1,10 +1,12 @@
 # Data Release Module Test Plan
```
ridz1208: valuable additions to this file:

 - granting permission to aversion from the Add permission button is not covered
 - try to add permission to both a file and a version, an error should be thrown.
 - in the permissions section, item 10 should come right before 8. otherwise it gets confusing that an unchecked box does not actually revoke permissions
- add test for download CSV
- another fun usecase is to test multiple version "names". We had a bug a couple of versions ago where version names with a `.` in them were breaking the module (more specifically managing permissions)

cmadjar:
@ridz1208 Thank you for testing the PR and your review. All your points should be taken care of. Please re-review at your convenience. Thanks!

@johnsaigle Finally figured out how to get the swal window to close the modal window, yeah!! Also, maybe the bug you found when testing multiple users was linked to the one Rida reported. This has been fixed so maybe retest?

johnsaigle:
@cmadjar I'll re-test this this morning.

If you don't mind, could you rebase this on `aces/master`? At this point there is a breaking change between this PR and `master` so I have to rebuild raisinbread each time I test the PR.

Also there have been a few new static analysis tools added so it would be good to make sure this code meets those standards. Otherwise we may get some issues on the master branch.

modules/data_release/jsx/uploadFileForm.js
```diff
@@ -0,0 +1,230 @@
...
+      formData[formElement] = value;
+    }
+
+    this.setState({formData: formData});
+  }
+
+  /**
+   * Validate and submit the upload
+   */
+  validateAndSubmit() {
```
johnsaigle: It would be good to upgrade this function (or the uploadFile function) to handle the error cases in `ajax/FileUpload.php`.

For example I tried uploading a file when the directory didn't exist and I got a 500 error/blank page.

Instead it would be good to have the ajax script send error messages here to be displayed by the SWAL.

cmadjar: actually, this is happening in `uploadFile` function. The only thing I don't get is why the error message do not show in the swal window when the file is overwritten but it shows for a new file not already uploaded...

cmadjar:
@johnsaigle I rebased to aces/master. Not sure what you meant by "Also there have been a few new static analysis tools added so it would be good to make sure this code meets those standards. Otherwise we may get some issues on the master branch." but I guess Travis will let me know, lol ;).

> Improvement to error handling needed (see comment below)
See my reply to your comment. I have no idea why the messages don't show up when overwritting a file but do show up if the file did not exist in the DB. Probably some JS magic I don't get... @HenriRabalais @driusan any idea?

> The Manage Permissions button opens a modal with an empty table (no entires for either low-privileged User or for Admin).
I could not replicate that issue. Here is what I get when clicking on the "Manage Permissions" button.
Screen Shot 2020-02-10 at 2 01 50 PM

> UI overflow on file upload modal for long filenames (#5842)
Looks like somehow it was fixed in aces/master as I could not replicate the issue. Here is the screenshot I have for a very long filename.
Screen Shot 2020-02-10 at 2 03 07 PM

modules/data_release/jsx/uploadFileForm.js
cmadjar: @johnsaigle I fixed it. Don't ask me the details on how I fixed it, I have no idea, lol!! At least now the message shows in the swal window:
Screen Shot 2020-02-10 at 2 20 29 PM


@HenriRabalais @driusan forget my earlier comment on this PR. Looks like I figured out how to fix this somehow...

cmadjar:
@johnsaigle I guess this is ready for another round of review/testing. Thank you!!!

cmadjar:
> The README says
> 
> Note: At the moment, the only way to remove a user's permission to a specific
> data release is via the backend, by manually deleting rows in the
> data_release_permissions MySQL table.
> 
> Is this still true? I thought the Manage Permissions allows you to do this.

This is still true, the only way to remove a user's permission to a specific data release file is via the backend. You can only remove version-level permissions from the frontend. 

> Another permissions question -- a user seems to be able to grant themselves permission to any file or release they want if they have data_release_edit_file_access. Maybe this is the point of the permission but it makes it so that they can do whatever they want.

This is the intended behaviour so far.

PS: one little note to keep in mind, this PR was sent to reactify the data release module, not change the functionality unless it was necessary for the reactification. 

Thank you for your review!! :)

cmadjar:
@driusan This PR is ready for you. Thanks! :)

ridz1208:
@cmadjar passes my manual tests too.

@driusan all yours

modules/data_release/ajax/AddPermission.php
```diff
@@ -15,14 +15,32 @@
...
+if (!$user->hasPermission('data_release_edit_file_access')) {
+    header("HTTP/1.1 403 Forbidden");
+    exit;
+}
+
+
+if ($_GET['action'] == 'addpermission') {
+
+    if (!isset($_POST) || empty($_POST)) {
+        $message = "No data were sent in the POST request";
```
driusan:
```suggestion
        $message = "No data was sent in the POST request";
```

modules/data_release/ajax/AddPermission.php
```diff
@@ -61,16 +79,21 @@
...
+} elseif ($_GET['action'] == 'managepermissions') {
     // It is important here to not truncate all current permissions and rebuild
     // them but instead to only make modifications to specifically altered
     // permissions from the front end by the user. Otherwise, if a user has
     // access to SOME files for a version, the checkbox would not be checked and
     // on update the user will loose access to ALL files for that release.
 
-    // Get permission list before any changes from main class
+    if (!isset($_POST) || empty($_POST)) {
+        $message = "No data were sent in the POST request";
```
driusan:
```suggestion
        $message = "No data was sent in the POST request";
```

modules/data_release/ajax/FileUpload.php
```diff
@@ -17,33 +17,32 @@
...
-
     if (!file_exists($path)) {
-        error_log(
-            "ERROR: File upload failed. Upload"
-            . " directory not found."
-        );
-        header("HTTP/1.1 500 Internal Server Error");
+        $msg = "File upload failed. Upload directory not found.";
+        error_log('ERROR: ' . $msg);
+        header("HTTP/1.1 404 $msg");
```
driusan: The string part of a standard HTTP response code shouldn't be customized, it should still be "404 Not Found".

modules/data_release/ajax/FileUpload.php
```diff
@@ -17,33 +17,32 @@
...
     } elseif (!is_writable($path)) {
-        error_log(
-            "File upload failed. Upload directory"
-            . " does not appear to be writeable."
-        );
-        header("HTTP/1.1 500 Internal Server Error");
+        $msg = "File upload failed. Upload directory is not writeable.";
+        error_log('ERROR: ' . $msg);
+        header("HTTP/1.1 404 $msg");
```
driusan: same. The old 500 Internal Server Error also seems to be a more accurate response.

modules/data_release/ajax/AddPermission.php
cmadjar: updated in the last commit

modules/data_release/ajax/AddPermission.php
cmadjar: updated in the last commit

modules/data_release/ajax/FileUpload.php
cmadjar: updated in the last commit

modules/data_release/ajax/FileUpload.php
cmadjar: updated in the last commit

cmadjar:
@driusan Ready for you!

PR #5231, opened 2019-09-18
[Media] Use DataFrameworkMenu class

johnsaigle:
## Brief summary of changes

Update the Media module to use the new DataFrameworkMenu class.

## Testing

Should work the same as before.

### DIFF
--- a/modules/media/php/media.class.inc
+++ b/modules/media/php/media.class.inc
@@ -131,7 +131,15 @@ class Media extends \DataFrameworkMenu
      */
     public function getBaseDataProvisioner(): \LORIS\Data\Provisioner
     {
-        return new MediaFileProvisioner();
+        $provisioner = parent::getDataProvisionerWithFilters();
+
+        if (!\NDB_Factory::singleton()->user()->hasPermission('superuser')) {
+            $provisioner = $provisioner->filter(
+                new HideFileFilter()
+            );
+        }
+
+        return $provisioner;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


modules/media/php/media.class.inc
```diff
@@ -118,12 +124,22 @@ class Media extends \NDB_Menu_Filter
...
+
     /**
      * Gets the data source for this menu filter.
      *
      * @return \LORIS\Data\Provisioner
      */
-    function getDataProvisioner() : \LORIS\Data\Provisioner
+    public function getDataProvisionerWithFilters() : \LORIS\Data\Provisioner
     {
         $provisioner = new MediaFileProvisioner();
```
xlecours:
```
        $provisioner = parent::getDataProvisionerWithFilters();
        
        if (!$user->hasPermission("superuser")) {
            $provisioner = $provisioner->filter(
                new HideFileFilter()
            );
        }
        
        return $provisioner;
```

modules/media/php/media.class.inc
```diff
@@ -151,7 +167,7 @@ class Media extends \NDB_Menu_Filter
     function toJSON() : string
```
xlecours: This function is inherited now. It can be removed

modules/media/php/media.class.inc
xlecours: @johnsaigle This suggestion was applied to the wrong place. 
`getBaseDataProvisioner` should have 1 line `return  new MediaFileProvisioner();`

And `getDataProvisionerWithFilters` body should be :
```
        $provisioner = parent::getDataProvisionerWithFilters();
        
        if (!\NDB_Factory::singleton()->user()->hasPermission("superuser")) {
            $provisioner = $provisioner->filter(
                new HideFileFilter()
            );
        }
        
        return $provisioner;
```

The `parent::getDataProvisionerWithFilters();` will take care of 'access_all_profiles'.

johnsaigle: Thanks, I also just realized this after looking at the Travis logs.

modules/media/php/media.class.inc
```diff
@@ -118,52 +124,32 @@ class Media extends \NDB_Menu_Filter
...
-        $user = \User::singleton();
+        $provisioner = parent::getDataProvisionerWithFilters();
 
-        if ($user->hasPermission("access_all_profiles") == false) {
-            $provisioner = $provisioner->filter(
-                new \LORIS\Data\Filters\UserSiteMatch()
-            );
-        }
-        if (!$user->hasPermission("superuser")) {
+        if (!\NDB_Factory::singleton()->user()->hasPermission('superuser')) {
```
driusan: since User is already passed to `HideFileFilter->filter` and the filter is only used in this module, I think it would be cleaner to move the permission check there and unconditionally add the filter.

(cc @xlecours since this is basically the same comment I had on one of your PRs and that discussion isn't resolved yet.)

johnsaigle:
Let me know if these are the changes you had in mind @driusan.

driusan:
@xlecours can you update your review (if applicable)?

PR #5233, opened 2019-09-18
[timepoint_list][bugfix] 'View Candidate Info' button wasn't redirecting

jesscall:
## Brief summary of changes
When user has the permissions:
"View candidate parameters" but NOT "Edit candidate parameters", the button is supposed to read "View candidate info" (which it does), but it is not responsive. Clicking on it does not load the candidate parameters page.

Changed the tag in template so that the button in question actually redirects.

1. Assign yourself the appropriate permissions and see if the link redirects to candidate parameters


driusan:
@ridz1208 can you test/approve?

xlecours:
@ridz1208 , @jesscall What about removing one of the button and chnage the remaining one label's to `Candidate Info` ? Effectively, it's the candidate_parameters page that will decide if the info can be edited. right?

ridz1208:
@jesscall your solution is def valid but to avoid that same issue in the future, lets move forward with this change.

PR #5237, opened 2019-09-23
[src/Middleware] AnonymousPageDecorationMiddleware.php Post response bugfix

maltheism:
Brief summary of changes

Yesterday I was having issues with attempting to receive a JSON object from the server (after a post request from the client). The client was receiving data from the server but html content was included as well. Making my JSON object not be a JSON object and the javascript error would fail inside the fetch function.

One of my PRs is working on making the login form react. Well it turns out this bug only happens when the user isn't logged in. This bugfix is needed for when the user is not logged in and for the server to respond to a post request with data that's correctly formatted (such as a JSON object).

Note: I made a PR for both major & bugfix branch because I'm in the processing of making the login react on major.

This resolves issue...

  Redmine? #

  Github? #

To test this change...

  Trust me it fixes the problem above. Otherwise make a post request without being logged in and attempt to receive a json response from the server. You would be debugging for days without this fix!

### DIFF
--- a/src/Middleware/AnonymousPageDecorationMiddleware.php
+++ b/src/Middleware/AnonymousPageDecorationMiddleware.php
@@ -56,7 +56,18 @@ class AnonymousPageDecorationMiddleware implements MiddlewareInterface
                                    );
         }
 
+        // Handle needs to be called before form action, because handle potentially
+        // calls setup which modifies the $page->FormAction value (ie in the imaging
+        // browser)
         $undecorated = $handler->handle($request);
+        $contenttype = $undecorated->getHeaderLine("Content-Type");
+        if ($contenttype != "" && strpos($contenttype, "text/html") === false) {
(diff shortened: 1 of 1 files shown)


maltheism:
See #5237 4386

johnsaigle:
Shouldn't this return a 401 response instead?

maltheism:
@johnsaigle I don't understand what you mean. My react login PR requires this change and has it in the PR. LORIS currently cannot receive json from POST or GET requests if not logged in. This PR fixes that problem.

driusan:
@johnsaigle why would it return a 401 if the handle method of the page didn't add a Content-Type header?

The problem when I tried only decorating pages that have a content-type of text/html when originally writing the Middleware was that 90% of the pages don't add an explicit content-type because it wasn't required before. I suspect this that's still the case because no one's made an effort to fix that.

This should be fine because it only returns the undecorated if there *is* a content-type *and* it doesn't contain text/html.

PR #5140, opened 2019-08-26
[Conflict Resolver] Reactify + DataFramework

xlecours:
The frontend and the backend follows the same structure; There is a Conflict_resolver React app that renders a `<FilterableDatatable>` for unresolved or resolved conflict according to the selected tab.
This React app is provided by the conflict_resolver.class.inc php file along with all the CSS and other JS. The rendered `<FilterableDatatable>` will the make a fetch (GET) call to either conflict_resolver/unresolved or conflict_resolver/resolved that correspond to 2 RequestHandlers with the same name. Each handler will then process the request and return the data and the filterOptions.

The conflict_resolver/unresolved endpoint also support POST request with conflict_id and correct_answer to resolve a conflict. As long as the page is not refreshed, it is possible to fix a conflict multiple time; each time changing the values in the "wrong" instrument, recalculate score, inserting/replacing the row in conflicts_resolved and removing the row in conflicts_unresolved if not already removed.

A UNIQUE constraint is added on ConflictID in conflicts_resolved table to allow REPLACE sql statement to work when a conflict is resolved then corrected.

**Still missing**:
- [x]  Feedback to user when a conflict is fixed. "Saved" tooltip or label which fades after 1 second or something like that.
- [x] Integration test needs to be adapted

**Bonus**: 
This provide a new database provisioner, `src/Data/Provisioners/DBObjectProvisioner.php`, that returns a Traversable of objects of a given class by using PDOStatement::setFetchMode( PDO::FETCH_CLASS). Basically, each rows are transformed from an associative array to an object with column names as properties. (see unresolveddto.class.inc and resolveddto.class.inc)

This also adds a open-api spec schema.yml file in the `static` directory of the module to view the endpoitns in swagger-ui.

PR #5238, opened 2019-09-23

maltheism:
### Brief summary of changes

### This resolves issue...

- [x] Redmine? #16008

### To test this change...

Please see #5140


modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,280 @@
...
+    }
+    /**
+     * Permission verification.
+     *
+     * @param ServerRequestInterface $request The incoming PSR7 request.
+     * @param array                  $values  The form values.
+     *
+     * @return array
+     */
+    private function _verifyPermissions(
```
johnsaigle: Why not use `_hasAccess`?

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,280 @@
...
+     */
+    private function _handleGET(ServerRequestInterface $request) : ResponseInterface
+    {
+        // Parse GET query params.
+        $values = $request->getQueryParams();
+        // Verify identifier from Post.
+        $values['identifier'] = isset($values['identifier']) ?
+            intval($values['identifier']) : null;
+        // User permission verification.
+        $values = $this->_verifyPermissions($request, $values);
```
johnsaigle: I think it makes more sense to create private class variables rather than pass `$values` all over the place in this class.

modules/create_timepoint/php/ajaxtimepoint.class.inc
```diff
@@ -0,0 +1,280 @@
...
+        }
+        $values['subproject'] = array_intersect_key(
+            \Utility::getSubprojectList(),
+            $candidate->getValidSubprojects()
+        );
+        // Retrieve visit labels.
+        $visit_options      = array();
+        $visitLabelSettings = $config->getSetting('visitLabel');
+        foreach (
+            \Utility::toArray($visitLabelSettings) as $visitLabel
```
johnsaigle: `Utility::toArray()` is deprecated and shouldn't be used. You can use `associativeToNumericArray` instead.

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: Hey @johnsaigle I'm making a change to include the _hasAccess but I'm still needing the _verifyPermissions function. The new endpoints process for handling the requests was introduced by @xlecours and has the _hasAccess called before the 'get' or 'post' request data can be processed. The permissions needs to check the "candidate" for RegistrationCenterID & CenterID;. The candidate is made with the needed identifier sent from the client.

modules/create_timepoint/php/ajaxtimepoint.class.inc
maltheism: @johnsaigle I kind of prefer passing the $values here. I'm sort of against class variables here because I find them less intuitive to keep track of the process and when the function returns the new set of values.

modules/create_timepoint/.gitignore
```diff
@@ -0,0 +1 @@
+js/*
```
kongtiaowang: js/* means all js file should be ignored. 
How about this modules/create_timepoint/js/createTimepointIndex.js?

modules/create_timepoint/php/endpoints/timepoint.class.inc
```diff
@@ -0,0 +1,339 @@
+<?php declare(strict_types=1);
```
kongtiaowang: Is it necessary to add sub-folder endpoints under php? If you insist on using this, you need to modify test/run-php-linter.sh to pass Travis. (add the endpoints into line 6 to make the checkstatic works)

johnsaigle: > you need to modify test/run-php-linter.sh to pass Travis

Why does this make Travis fail? The command looks like it should check every PHP file under `modules/`

xlecours: see: https://github.com/aces/Loris/pull/5249

maltheism: @xlecours should that be merged into major and not minor?

modules/create_timepoint/.gitignore
maltheism: @driusan is the conventional standard removing all .js files in the js directory? Otherwise should I remove just the js file in .gitignore as @kongtiaowang suggests.

driusan: the convention is generally "if there are some pre-React .js files already tracked there, ignore individual files, otherwise ignore all files in the js directory"

maltheism:
See #5269

PR #5240, opened 2019-09-23
[Utility] Deprecate toArray() function

johnsaigle:
## Brief summary of changes

This function should've been properly deprecated in LORIS v21 but was missed during the release process.

See #4151


johnsaigle:
@driusan This is now going to major.

driusan:
Any idea why Travis doesn't seem to want to run on it?

johnsaigle:
@driusan This is passing Travis even though it appears as pending on GitHub.

PR #5245, opened 2019-09-24
[Install] Add -p to mkdir so install doesn't fail

### LINKED ISSUE #5244: [tools/install.sh] not working like before
I've been using the tools/install.sh to generate the config.xml inside the project directory. 

Changes have been made to install.sh and now the script will stop before making the config.xml file unlike before. 

Basically if a "user_uploads" directory already exists and this can be in the document_repository or the data_release module. I would prefer the script to just skip the create directory commands if the directories already exists. Otherwise I have to manually delete the directories before I can run the script again and unlike before.

Also the script is reporting:
> mkdir: cannot create directory ‘../modules/document_repository/user_uploads’: File exists
Note: user_uploads is not a file.

johnsaigle:
## Brief summary of changes

Adds `-p` flag to `mkdir` so that the install script won't fail when `user_uploads/` already exists (for whatever reason) in doc_repo or data_release.

Fixes #5244.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -161,8 +161,8 @@ debian=("Debian" "Ubuntu")
 redhat=("Red" "CentOS" "Fedora" "Oracle") 
 
 if [[ " ${debian[*]} " =~ " $os_distro " ]]; then
-    mkdir ../modules/document_repository/user_uploads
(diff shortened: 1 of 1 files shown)


johnsaigle:
@maltheism Please let me know whether this resolves your issue (#5244).

PR #5246, opened 2019-09-25
Add ghooks: forces "make checkstatic" before git push and "make dev" after git merge

johnsaigle:
## Brief summary of changes

Adds a new developer dependency `ghooks` that forces certain actions to be run when certain `git` events are triggered.

Two hooks are added here with the following results:

* When using `git push`, `make checkstatic` must exit successfully or the push will fail.
* When using `git merge` or `git pull`, `make dev` will be run automatically.

This should help development by:
* forcing developers to fix `phan`, `phpcs`, and `eslint` errors locally rather than waiting for Travis (and often another developer) to tell them that there is an issue with their code.
* Preventing confusion when a developer switches branches and forgets to recompile their JavaScript code

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "eslint-config-google": "0.9.1",
     "eslint-loader": "2.1.0",
     "eslint-plugin-react": "7.11.1",
+    "ghooks": "^2.0.4",
     "prop-types": "latest",
     "react": "^16.4.2",
     "react-addons-create-fragment": "^15.6.2",
@@ -51,5 +52,11 @@
   "bugs": {
     "url": "https://github.com/aces/Loris/issues"
   },
-  "homepage": "https://github.com/aces/Loris"
+  "homepage": "https://github.com/aces/Loris",
+  "config": {
+      "ghooks": {
+          "pre-push": "make checkstatic",
+          "post-merge": "make dev"
+      }
+  }
 }


ridz1208:
I like the idea but for the git merge/git pull, I'm worried it might be a bit time consuming... the make dev takes too long

johnsaigle:
@ridz1208 I think it's important to do this every time you merge/pull as-is. Otherwise you won't see JavaScript changes and may be dealing with the wrong dependencies. 

To be clear this doesn't trigger on `checkout`. It's only on `merge` or `pull`.

ridz1208:
@johnsaigle I understand, I guess the only scenario I'm thinking about is the following.

I work off of 2 workstations regularly so I'm contantly merging changes from my remotes, and it happens often that I merge to just fix something in PHP and/or SQL that has no effect otherwise... (given that I rarely deal with the front end JSX stuff)

I'm not gonna block this, I think it has alot of value but its just a consideration

johnsaigle:
@ridz1208 We could add another `make` target that just updates PHP and JS dependencies but doesn't trigger webpack. 

I agree webpack takes a long time.

driusan:
I don't think shoehorning this into some javascript package that messes with your .git directory (I assume, I don't know how else it would work..) is the right way to do this. It would be better to include standard git commit hook scripts that developers could opt in to.

johnsaigle:
I don't have any preference whether it happens via JS or via regular git hooks but I think that (for the `make checkstatic` check at least) it might be good to make it mandatory.

maltheism:
@johnsaigle I really like the change for this functionality:
> When using git push, make checkstatic must exit successfully or the push will fail.

I'm undecided on how I feel about:
> When using git merge or git pull, make dev will be run automatically.

Having "make dev" be needed on git pull seems like it could slow me down unnecessarily. Anyway just my feedback.

johnsaigle:
@maltheism Thanks for the feedback. I'll likely close this PR in order to add standard git hooks, as Dave suggested.

From there developers can opt-in to the functionality they like. After it's merged I might give a small presentation on how to set them up.

PR #5248, opened 2019-09-25
[media] Use study's start/end date as min/max values for Date of Administration field

jesscall:
## Brief summary of changes
When uploading a form to media module, you cannot enter a date greater than 2017-12-31. The date of administration field is currently hard-coded to have a minYear of '2000' and a maxYear of '2017'. This PR aims to change this. The min and max year are now determined by the study's start and end year (taken from configuration).

1. Try uploading a file with a date of administration that is within the bounds of the study's start and end year.

### DIFF
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -207,6 +207,7 @@ function getUploadFields()
 
     $db   = \NDB_Factory::singleton()->database();
     $user = \User::singleton();
+    $config = \NDB_Config::singleton();
 
     // Select only candidates that have had visit at user's sites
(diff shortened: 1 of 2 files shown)


modules/media/ajax/FileUpload.php
```diff
@@ -207,6 +207,7 @@ function getUploadFields()
 
     $db   = \NDB_Factory::singleton()->database();
     $user = \User::singleton();
+    $config = \NDB_Config::singleton();
```
maltheism: The equal sign of $config is not aligned with the previous variables. `make checkstatic`

PR #5249, opened 2019-09-26
Fix php linter

xlecours:
## Brief summary of changes
With the introduction of php subdirectories in modules, run-php-linter fails. This uses `find` instead of `ls` to create the file list.

create a subdirectory in a `module/?/php`. 
add a `.php` or `.class.inc` file in the directory.
test with a syntax error and without.
`./test/run-php-linter.sh` should exit if an error is found.

### DIFF
--- a/test/run-php-linter.sh
+++ b/test/run-php-linter.sh
@@ -3,7 +3,7 @@ set -euo pipefail
 
 # Run PHP -l on everything to ensure there's no syntax
 # errors.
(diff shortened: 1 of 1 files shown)


driusan:
I tried:

1. Running the new command and checking the exit status, got an exit code of 0
2. Adding noise to a file in php and got a non-0 exit code
3. Adding a new PHP file in a subdirectory of a module which just contains noise and got a non-0 exit code.

I think this is good.

PR #4962, opened 2019-07-17
[SQL] Removed mri_acquisition_dates table from the database and the LORIS code

cmadjar:
## Brief summary of changes

This PR removes the `mri_acquisition_dates` table from the database as this table does not necessarily store the correct acquisition date. Instead, replaced all the queries to look for the minimal date present in the `parameter_file` table which is more robust than what was stored in `mri_acquisition_dates`.

Note: the decision to remove this table was taken during an imaging meeting in 2017 and confirmed at an other imaging meeting in 2019. 

Link to the MRI PR: https://github.com/aces/Loris-MRI/pull/463

#### Testing instructions (if applicable)

1. Run the SQL patch `SQL/New_patches/2019-07-17_remove_mri_acquisition_dates_table.sql`
2. Make sure the dashboard is loading and plotting the correct graph for the MRI acquisitions
3. Make sure the imaging browser is loading and that the First Acquisition date properly populated

#### Links to related tickets (GitHub, Redmine, ...)

* https://redmine.cbrain.mcgill.ca/issues/6040

PR #5260, opened 2019-10-02
[Tools] Add script to delete and refresh raisinbread data

### LINKED ISSUE #5255: Create tool to refresh Raisinbread data
The tool should flush current database data and then run the commands in the `raisinbread/README.md`:

First:

```bash
cat raisinbread/instruments/instrument_sql/9999-99-99-drop_instrument_tables.sql \
    SQL/9999-99-99-drop_tables.sql | mysql
```

and then:

```bash
cat SQL/0000-00-00-schema.sql \
    SQL/0000-00-01-Permission.sql \
    SQL/0000-00-02-Menus.sql \
    SQL/0000-00-03-ConfigTables.sql \
    SQL/0000-00-04-Help.sql \
    SQL/0000-00-05-ElectrophysiologyTables.sql \
    raisinbread/instruments/instrument_sql/aosi.sql \
    raisinbread/instruments/instrument_sql/bmi.sql \
    raisinbread/instruments/instrument_sql/medical_history.sql \
    raisinbread/instruments/instrument_sql/mri_parameter_form.sql \
    raisinbread/instruments/instrument_sql/radiology_review.sql \
    raisinbread/RB_files/*.sql | mysql
```

Finally, the user should be prompted for the `url` and `host` Config settings in case they need to be changed from the default `localhost`.

This could also chain with `tools/resetpassword.php` so that the `admin` password must be changed from the default.

johnsaigle:
## Brief summary of changes

This script will dump your raisinbread database and then re-import the data. It automates the steps in [raisinbread/README.md](https://github.com/aces/Loris/tree/minor/raisinbread).

* It will also restore your `url`, `base`, `host` settings so that LORIS won't break if you're not hosting it on `localhost`.

* It will ask you to change the front-end `admin` password so you can login to LORIS without using the public Raisinbread password.

* I've added `make testdata` as a quick way to refresh your database.

* This can also be used upon switching between branches so that you never get an issue where your database is broken because of sync issues.

~~As a precaution I decided to force the user of this script to manually type in the database name of the database they want to effectively delete. This should help prevent data loss. Feel free to suggest a different security mechanism.~~

UPDATE I incorporated Rida's suggestion to force the script to exit if the `sandbox` flag is not set in config.xml.

1. Checkout the branch.
2. run `make testdata` OR go to `tools/` and run `php raisinbread_refresh.php`
3. Verify that you have fresh Raisinbread data and that your front-end still works.

* Resolves #5255.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,5 +34,4 @@ unittests: phpdev
 check: checkstatic unittests
 
 testdata:
-	cd tools/ && php raisinbread_refresh.php
-	cd ..
+	php tools/raisinbread_refresh.php
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: clean dev all check checkstatic unittests test phpdev javascript
+.PHONY: clean dev all check checkstatic unittests test phpdev javascript raisinbread
 
 all: VERSION javascript
 	composer install --no-dev
@@ -32,3 +32,7 @@ unittests: phpdev
 
 # Perform all tests that don't require an install.
 check: checkstatic unittests
+
+raisinbread:
+	cd tools/ && php raisinbread_refresh.php
+	cd ..
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: clean dev all check checkstatic unittests test phpdev javascript raisinbread
+.PHONY: clean dev all check checkstatic unittests test phpdev javascript testdata
 
 all: VERSION javascript
 	composer install --no-dev
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,6 @@ unittests: phpdev
 # Perform all tests that don't require an install.
 check: checkstatic unittests
 
-raisinbread:
+testdata:
 	cd tools/ && php raisinbread_refresh.php
 	cd ..
--- a/tools/raisinbread_refresh.php
+++ b/tools/raisinbread_refresh.php
@@ -8,12 +8,12 @@
  * The `sandbox` flag in config.xml is checked for this very reason. The script
  * will abort if it is not set to 1.
  *
- * After dropping the tables, the script will source the Raisinbread test data 
+ * After dropping the tables, the script will source the Raisinbread test data
  * using the commands found in raisinbread/README.md.
  *
- * The script also restores the url, base, and host config settings to their 
- * pre-drop values. This prevents an issue where a developer will need to 
- * manually change these values in a MySQL shell when they are not hosting a 
+ * The script also restores the url, base, and host config settings to their
+ * pre-drop values. This prevents an issue where a developer will need to
+ * manually change these values in a MySQL shell when they are not hosting a
  * LORIS on localhost.
  *
  * Finally, the script runs tools/resetpassword.php so that the default admin
@@ -22,7 +22,7 @@
  * In order to prevent accidental data loss, the script prompts the user to
  * manually type the exact name of the database which will be affected.
  *
- * In normal cases, the database connection will be established via the 
+ * In normal cases, the database connection will be established via the
  * credentials in the config file. However, if the database tables have already
  * been deleted then this will not work. In this case it is still possible to
  * import Raisinbread data if the user has properly set up a MySQL configuration
@@ -64,13 +64,13 @@ try {
         );
     }
     $dbname = $dbInfo['database'];
-    $host = $dbInfo['host'];
+    $host   = $dbInfo['host'];
     // The admin user and password should be configured to a user with DROP
     // and CREATE permissions for the database.
     $username = $dbInfo['adminUser'];
     $password = $dbInfo['adminPassword'];
 
-    $urlConfigSetting = $config->getSetting('url');
+    $urlConfigSetting  = $config->getSetting('url');
     $baseConfigSetting = $config->getSetting('base');
     $hostConfigSetting = $config->getSetting('host');
 
@@ -78,7 +78,7 @@ try {
     Please type the database name `$dbname` to confirm you wish to drop tables
     and import test data: 
 CONFIRMATION;
-    
+
     $input = trim(fgets(STDIN));
     if ($input !== $dbname) {
         die('Input did not match database name. Exiting.');
@@ -108,7 +108,7 @@ mysql -A $dbname
 CMD;
 
 // Test whether a connection to MySQL is possible via a MySQL config file.
-// If not, read DB information from the config file. This method is not as 
+// If not, read DB information from the config file. This method is not as
 // preferable because it generates MySQL warnings due to the password being
 // supplied via a command-line argument.
 exec($mysqlCommand . ' -e "show tables;" 2>&1 1>/dev/null', $output, $status);
@@ -119,7 +119,7 @@ CMD;
 }
 
 // Drop tables
-echo PHP_EOL .'[*] Dropping LORIS tables....' . PHP_EOL;
+echo printHeader('Dropping LORIS tables.');
 dropTables();
 
 // Print the names of remaining tables, if any. Some tables may remain if they
@@ -144,27 +144,21 @@ if (count($tables) > 0) {
 }
 
 // Source LORIS core tabes
-echo <<<INFO
-[*] Creating LORIS core tables.... 
+printHeader('Creating LORIS core tables...');
 
-INFO;
-$coreTables = glob("../SQL/0000-*.sql"); 
+$coreTables = glob("../SQL/0000-*.sql");
 array_walk($coreTables, 'runPatch');
 
 // Create instrument tables
-echo <<<INFO
-[*] Creating instrument tables.... 
+printHeader('Creating instrument tables...');
 
-INFO;
-$rbInstrumentTables = glob("../raisinbread/instruments/instrument_sql/*.sql"); 
+$rbInstrumentTables = glob("../raisinbread/instruments/instrument_sql/*.sql");
 array_walk($rbInstrumentTables, 'runPatch');
 
 // Import Raisinbread data
-echo <<<INFO
-[*] Importing Raisinbread data....
+printHeader('Importing Raisinbread data...');
+$rbData = glob("../raisinbread/RB_files/*.sql");
 
-INFO;
-$rbData= glob("../raisinbread/RB_files/*.sql");
 array_walk($rbData, 'runPatch');
 
 // Restore config settings if they were successfully found before.
@@ -180,9 +174,11 @@ if (isset($hostConfigSetting)) {
 
 // Trigger a password reset because the password for `admin` in the Raisinbread
 // database is public.
-echo "[*] Changing the admin password..." . PHP_EOL;
+printHeader('Changing the admin password...');
 echo 'Please choose a new password for the admin user:' . PHP_EOL;
 runCommand('php resetpassword.php admin');
+
+printHeader('Test data successfully installed.');
 
 
 // END SCRIPT
@@ -190,21 +186,24 @@ runCommand('php resetpassword.php admin');
 
 /**
  * Drops core LORIS tables as well as Raisinbread instrument tables
- * 
+ *
  * @return void
  */
 function dropTables(): void
 {
     runPatch(
-        '../raisinbread/instruments/instrument_sql/9999-99-99-drop_instrument_tables.sql'
-        );
+        '../raisinbread/instruments/instrument_sql/'
+        . '9999-99-99-drop_instrument_tables.sql'
+    );
     runPatch('../SQL/9999-99-99-drop_tables.sql');
 }
 
 /**
- * Wrapper for runCommand() that pipes the content of an SQL file to an CLI 
+ * Wrapper for runCommand() that pipes the content of an SQL file to an CLI
  * instance of MySQL.
- * 
+ *
+ * @param string $file The name of the file to run.
+ *
  * @return void
  */
 function runPatch(string $file): void
@@ -241,16 +240,16 @@ function runCommand(string $command): void
 /**
  * Update a config setting in LORIS to $value.
  *
- * @param string $name Name of config setting in ConfigSettings table.
+ * @param string $name  Name of config setting in ConfigSettings table.
  * @param string $value Value to be changed in the Config table.
  *
  * @return void
  *
  * @throws \DatabaseException
  */
-function restoreConfigSetting(string $name, string $value): void 
+function restoreConfigSetting(string $name, string $value): void
 {
-    echo "[*] Restoring config settings..." . PHP_EOL;
+    printHeader('Restoring config settings...');
     echo "Restoring config setting `$name` to value `$value`"
         . PHP_EOL . PHP_EOL;
     try {
@@ -269,14 +268,23 @@ function restoreConfigSetting(string $name, string $value): void
     }
 }
 
-function dropRemainingTables(array $tables) {
+/**
+ * Drops tables that exist in the database after DROP tables scripts have been
+ * executed.
+ *
+ * @param string[] $tables The name of tables in the database.
+ *
+ * @return void
+ */
+function dropRemainingTables(array $tables): void
+{
     global $mysqlCommand;
     $commands = ["SET FOREIGN_KEY_CHECKS = 0;"];
     foreach ($tables as $table) {
         $commands[] = "DROP TABLE IF EXISTS $table;";
     }
     $commands[] = "SET FOREIGN_KEY_CHECKS = 1;";
-    $script = <<<SCRIPT
+    $script     = <<<SCRIPT
 $mysqlCommand -e '%s'
 SCRIPT;
     exec(sprintf($script, implode("\n", $commands)));
@@ -289,7 +297,8 @@ SCRIPT;
  *
  * @return void
  */
-function printBulletPoint(string $line): void {
+function printBulletPoint(string $line): void
+{
     echo "\t* $line\n";
 }
 
@@ -300,7 +309,8 @@ function printBulletPoint(string $line): void {
  *
  * @return void
  */
-function printHeader(string $line) {
-    // Takes 'input', prints '[*] input...' in green text.
-    fwrite(STDOUT, "\e[32m[*] $line... \e[0m\n");
+function printHeader(string $line): void
+{
+    // Takes 'input', prints '[*] input' in green text.
+    fwrite(STDOUT, PHP_EOL . "\e[32m[*] $line\e[0m" . PHP_EOL);
 }
--- a/tools/raisinbread_refresh.php
+++ b/tools/raisinbread_refresh.php
@@ -44,24 +44,22 @@ them with new test data.
 
 
 INFO;
-
 echo $info;
 
 echo "\e[0;31m*** Executing this script will result in the LOSS OF DATA! ***\e[0m\n";
 
 $cwd = getcwd();
-if (substr_compare($cwd, 'tools', mb_strlen($cwd) - mb_strlen('tools')) !== 0) {
-    die('Please run this script from the tools/ directory.' . PHP_EOL);
-}
 
 try {
     $dbInfo = $config->getSettingFromXML('database');
 
     if (! $config->getSetting('dev')['sandbox']) {
-        throw new \LorisException(
+        fwrite(
+            STDERR,
             "Config file indicates that this is not a sandbox. Aborting to " .
             "prevent accidental data loss." . PHP_EOL
         );
+        exit(1);
     }
     $dbname = $dbInfo['database'];
     $host   = $dbInfo['host'];
@@ -96,8 +94,6 @@ CONFIRMATION;
     }
     echo PHP_EOL . 'Please enter the name of your database:' . PHP_EOL;
     $dbname = trim(fgets(STDIN));
-} catch (\LorisException $e) {
-    die($e->getMessage());
 } catch (Exception $e) {
     die("Could not load project/config.xml");
 }
@@ -146,18 +142,20 @@ if (count($tables) > 0) {
 // Source LORIS core tabes
 printHeader('Creating LORIS core tables...');
 
-$coreTables = glob("../SQL/0000-*.sql");
+$coreTables = glob(__DIR__ . "/../SQL/0000-*.sql");
 array_walk($coreTables, 'runPatch');
 
 // Create instrument tables
 printHeader('Creating instrument tables...');
 
-$rbInstrumentTables = glob("../raisinbread/instruments/instrument_sql/*.sql");
+$rbInstrumentTables = glob(
+    __DIR__ . "/../raisinbread/instruments/instrument_sql/*.sql"
+);
 array_walk($rbInstrumentTables, 'runPatch');
 
 // Import Raisinbread data
 printHeader('Importing Raisinbread data...');
-$rbData = glob("../raisinbread/RB_files/*.sql");
+$rbData = glob(__DIR__ . "/../raisinbread/RB_files/*.sql");
 
 array_walk($rbData, 'runPatch');
 
@@ -176,13 +174,13 @@ if (isset($hostConfigSetting)) {
 // database is public.
 printHeader('Changing the admin password...');
 echo 'Please choose a new password for the admin user:' . PHP_EOL;
-runCommand('php resetpassword.php admin');
+flush();
+// Invoke the script `tools/resetpassword.php` for user 'admin'.
+$cmd = implode(' ', array('php', __DIR__ . '/resetpassword.php', 'admin'));
+exec($cmd);
 
 printHeader('Test data successfully installed.');
-
-
 // END SCRIPT
-
 
 /**
  * Drops core LORIS tables as well as Raisinbread instrument tables
@@ -192,10 +190,10 @@ printHeader('Test data successfully installed.');
 function dropTables(): void
 {
     runPatch(
-        '../raisinbread/instruments/instrument_sql/'
+        __DIR__ . '/../raisinbread/instruments/instrument_sql/'
         . '9999-99-99-drop_instrument_tables.sql'
     );
-    runPatch('../SQL/9999-99-99-drop_tables.sql');
+    runPatch(__DIR__ . '/../SQL/9999-99-99-drop_tables.sql');
 }
 
 /**
@@ -284,9 +282,7 @@ function dropRemainingTables(array $tables): void
         $commands[] = "DROP TABLE IF EXISTS $table;";
     }
     $commands[] = "SET FOREIGN_KEY_CHECKS = 1;";
-    $script     = <<<SCRIPT
-$mysqlCommand -e '%s'
-SCRIPT;
+    $script     = "$mysqlCommand -e '%s'";
     exec(sprintf($script, implode("\n", $commands)));
 }
 
--- a/tools/raisinbread_refresh.php
+++ b/tools/raisinbread_refresh.php
@@ -33,83 +33,107 @@
 $info = <<<INFO
 This script is used by LORIS developers to empty their databases and replace
 them with new test data.
-DO NOT RUN THIS SCRIPT ON PRODUCTION.
 
 
 INFO;
 
 echo $info;
 
+$cwd = getcwd();
+if (substr_compare($cwd, 'tools', strlen($cwd) - strlen('tools')) !== 0) {
+    die('Please run this script from the tools/ directory.' . PHP_EOL);
+}
+
 try {
     require_once 'generic_includes.php';
 
     $dbInfo = $config->getSettingFromXML('database');
+
+    if (! $config->getSetting('dev')['sandbox']) {
+        throw new \LorisException(
+            "Config file indicates that this is not a sandbox. Aborting to " .
+            "prevent accidental data loss."
+        );
+    }
     $dbname = $dbInfo['database'];
     $host = $dbInfo['host'];
     $username = $dbInfo['username'];
+    $password = $dbInfo['password'];
 
     $urlConfigSetting = $config->getSetting('url');
     $baseConfigSetting = $config->getSetting('base');
     $hostConfigSetting = $config->getSetting('host');
 
-    echo 'Please enter the name of your database to continue.' . PHP_EOL;
 } catch (\DatabaseException $e) {
     echo 'Could not connect to the database in the Config file.' .
-        'It\'s possible that the LORIS tables have been dropped already.' .
+        'It\'s possible that the LORIS tables have already been dropped.' .
         PHP_EOL;
+    echo 'Continue attempting to install Raisinbread database? (yN)' . PHP_EOL;
+    $input = trim(fgets(STDIN));
+
+    if (strtolower($input) !== 'y') {
+        die;
+    }
+    echo PHP_EOL . 'Please enter the name of your database:' . PHP_EOL;
+    $dbname = trim(fgets(STDIN));
+} catch (\LorisException $e) {
+    die($e->getMessage());
 } catch (Exception $e) {
     die("Could not load project/config.xml");
 }
 
-$input = trim(fgets(STDIN));
 
+echo <<<CONFIRMATION
+Please re-type the database name `$dbname` to confirm you wish to drop tables
+and import test data: 
+CONFIRMATION;
+
+$input = trim(fgets(STDIN));
 if ($input !== $dbname) {
-    die(
-        "Input did not match the name of the database in the config.xml file "
-        . "`$dbname`. Exiting as a precaution to prevent data loss."
-        . PHP_EOL
-    );
+    die('Input did not match database name. Exiting.');
 }
 
 // Create a connection to mysql via bash. All credentials are supplied via 
 // command line arguments except for the password which must be entered manually.
-$mysqlCommand = "mysql -A \"$dbname\" -u \"$username\" -h \"$host\" -p ";
+$mysqlCommand = "mysql -A \"$dbname\"";
 
 // Drop tables
 echo PHP_EOL .'Dropping LORIS tables....' . PHP_EOL;
-echo 'You will be prompted for your database password.' . PHP_EOL;
 $dropCommand = <<<DROP
-cat raisinbread/instruments/instrument_sql/9999-99-99-drop_instrument_tables.sql \
-    SQL/9999-99-99-drop_tables.sql | $mysqlCommand
+cat ../raisinbread/instruments/instrument_sql/9999-99-99-drop_instrument_tables.sql \
+    ../SQL/9999-99-99-drop_tables.sql | $mysqlCommand
 DROP;
 
 runCommand($dropCommand);
 
+
 // Source LORIS core tabes
 echo <<<INFO
-Creating LORIS core tables.... '
-You will be prompted for your database password.'
+Creating LORIS core tables.... 
 
 INFO;
 $createTablesCommand = <<<CMD
-for n in SQL/0000-*.sql; do echo \$n; ${mysqlCommand} < \$n || break; done;
+for n in ../SQL/0000-*.sql; do echo \$n; ${mysqlCommand} < \$n || break; done;
 CMD;
 runCommand($createTablesCommand);
 
 // Create instrument tables
-echo PHP_EOL .'Creating instrument tables.... ' .
-    'This may take some time.' . PHP_EOL;
-echo 'You will be prompted for your database password.' . PHP_EOL;
+echo <<<INFO
+Creating instrument tables.... 
+
+INFO;
 $createInstrumentTablesCommand = <<<CMD
-for n in raisinbread/instruments/instrument_sql/*.sql; do echo \$n; ${mysqlCommand} < \$n || break; done;
+for n in ../raisinbread/instruments/instrument_sql/*.sql; do echo \$n; ${mysqlCommand} < \$n || break; done;
 CMD;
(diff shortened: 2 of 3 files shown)


tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,138 @@
...
+
+    $dbInfo = $config->getSettingFromXML('database');
+    $dbname = $dbInfo['database'];
+    $host = $dbInfo['host'];
+    $username = $dbInfo['username'];
+
+    $urlConfigSetting = $config->getSetting('url');
+    $hostConfigSetting = $config->getSetting('host');
+
+    echo 'Please enter the name of your database to continue.' . PHP_EOL;
```
ridz1208: I think your TRY block is too wide here, it feels like you are adding it only for lines 50-51.

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,138 @@
...
+    SQL/0000-00-05-ElectrophysiologyTables.sql \
+    raisinbread/instruments/instrument_sql/aosi.sql \
+    raisinbread/instruments/instrument_sql/bmi.sql \
+    raisinbread/instruments/instrument_sql/medical_history.sql \
+    raisinbread/instruments/instrument_sql/mri_parameter_form.sql \
+    raisinbread/instruments/instrument_sql/radiology_review.sql \
+    raisinbread/RB_files/*.sql | mysql -A "$dbname" -u "$username" -h "$host" -p
+IMPORT;
+exec($importCommand, $output, $exitStatus);
+if ($exitStatus) {
```
ridz1208: Something that would be very useful here is to report which script caused the error. since everything is CATed together, the error you report below will only say something like 

`error on line 123456 foreign key reference...`

not useful when debugging. Since you are creating a script to do the work, you dont need to use the single command to source everything. you could divide them into sourcing schema, instrument schema, and data (and even in the data the individual files). and keep track at which file ur at so in case of failure you report the specific file that caused it

take a look at the troubleshooting section of the raisinbread readme for more details
https://github.com/aces/Loris/blob/minor/raisinbread/README.md#troubleshooting

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,138 @@
...
+    echo 'An error occurred when trying to import tables' . PHP_EOL;
+    print_r($output);
+    die;
+}
+
+
+if (isset($urlConfigSetting, $hostConfigSetting)) {
```
ridz1208: another important variable to restore I think is the `base`. that would be a great addition

Also the password for the admin user might be valuable

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,138 @@
...
+ * @package  Loris
+ * @author   John Saigle <john.saigle@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris-Trunk/
+ */
+
+$info = <<<INFO
+This script is used by LORIS developers to empty their databases and replace
+them with new test data.
+DO NOT RUN THIS SCRIPT ON PRODUCTION.
```
ridz1208: you should check if the sandbox flag is enabled and abort if it's not

christinerogers:
Fantastic.  Given that branch-switching with RaisinBread is a learning curve for new developers, log files would be an asset for troubleshooting.
@johnsaigle Would you consider logging when, where and especially on what this script was (last) run?

tools/raisinbread_refresh.php
johnsaigle: That's what I did at first but `getSettingFromXML()`will crash with a PHP Fatal Error if the right tables don't exist. That happened to me when I had a different error after dropping but before importing that crashed the script. 

Might be an edge-case but I figure there's no harm making it wide.

ridz1208: oh i see.. it's fine then

tools/raisinbread_refresh.php
johnsaigle: The `admin` password can be reset from the back-end using `tools/resetpassword.php`. Since this script is only meant to be run on dev instances I think triggering a dev reset is harmless and probably a good idea.

tools/raisinbread_refresh.php
johnsaigle: I improved the exception handling so that things are more fine-grained.

tools/raisinbread_refresh.php
johnsaigle: Done, thanks!

It turns out that `cat`-ing the files together makes everything WAY slower, like several minutes on my VM.

The for loop executes in a few seconds though. 

We should consider just recommending the for loop method in the README since it also gives error handling in addition to being faster.

tools/raisinbread_refresh.php
johnsaigle: Done!

johnsaigle:
@christinerogers I agree logging could be helpful. However, since this is written in PHP the logging process is more complicated than for example `install.sh` (which uses some bash pipe tricks to automatically copy its output to a new file).

This could easily be accomplished by running the script into `tee` though, e.g.

`php raisinbread_refresh.php | tee logs/raisinbread_refresh-$(date "+%Y-%m-%dT%H:%M:%S").log`

johnsaigle:
@ridz1208 Could you please re-review when you have time?

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,222 @@
...
+
+function runPatch(string $file): void
+{
+    global $mysqlCommand;
+    runCommand(
+        "cat $file | $mysqlCommand"
+    );
+}
+
+function sqlFileFilter(string $file): bool
```
ridz1208: what does this function do ?

ridz1208:
@johnsaigle code looks good. I wanna see how it behaves in some specific usecases. I will be taking out for a spin later.

👍

johnsaigle:
thank you!

tools/raisinbread_refresh.php
johnsaigle: It's leftover from an earlier approach I think. I'll remove it.

johnsaigle:
There's an unfortunate hiccup here because of a table that got removed on major.

If you have a `minor` Raisinbread and want to get a `major` one, the script will fail because it won't be able to DROP tables properly.

This is because a table `mri_acquisition_dates` got removed in #4962. Therefore when dropping the tables using the `major` version of the drop tables script, this table will not be dropped because it doesn't exist anymore:

https://github.com/aces/Loris/pull/4962/files#diff-73bc3032de06554055c4ea2f07de82e8L140

So it's foreign key doesn't get removed, meaning the tables don't drop properly, meaning you can't properly switch from `major` to `minor` Raisinbread unless you manually drop this table first.

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,216 @@
...
+} catch (\LorisException $e) {
+    die($e->getMessage());
+} catch (Exception $e) {
+    die("Could not load project/config.xml");
+}
+
+
+echo <<<CONFIRMATION
+Please re-type the database name `$dbname` to confirm you wish to drop tables
+and import test data:
```
PapillonMcGill: Could you add in CAPITALS letters something like
```
*** WARNING this will result in the LOST OF DATA *** 
           Not to be run on production server!
```

If you can add a few bell and whistle and flashing lights too, that would be good. People have the bad habit of not reading :-( .

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,216 @@
...
+// Import Raisinbread data
+echo <<<INFO
+[*] Importing Raisinbread data....
+
+INFO;
+$rbData= glob("../raisinbread/RB_files/*.sql");
+array_walk($rbData, 'runPatch');
+
+// Restore config settings if they were successfully found before.
+if (isset($urlConfigSetting)) {
```
PapillonMcGill: scope of $urlConfigSetting different than the try-catch block.

johnsaigle: I don't think try-catch has a different scope actually. It's not the same as declaring a new variable in a for-loop or a new function anyway.

The code runs as written and will restore the config setting.

tools/raisinbread_refresh.php
driusan: "LOSS" not "LOST"

PapillonMcGill: Can also use color Red on black background:
echo "\e[0;31;40m*** WARNING this will result in the LOST OF DATA ***\e[0m\n";

johnsaigle:
Updated the script to have a scary warning!

https://giphy.com/gifs/thesimpsons-the-simpsons-3x13-l2Je1EmFsrvSF504w

PapillonMcGill:
@johnsaigle Tried the script.
All mysql command ended in `Access denied for user ...`. The script never ask me for the password and look like it didn't use the one in the config file (`... using password: NO`).
It also used lorisadmin and not the user configured.  Should also use quatUser for dropping table as regular user in missing this permission.

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,218 @@
...
+CONFIRMATION;
+
+$input = trim(fgets(STDIN));
+if ($input !== $dbname) {
+    die('Input did not match database name. Exiting.');
+}
+
+// Create a connection to mysql via bash. All credentials are supplied via 
+// command line arguments except for the password which must be entered manually.
+$mysqlCommand = "mysql -A \"$dbname\"";
```
cmadjar: here, it would be nice so that the mysql user, host and password could be read from the `quatuser` of the config.xml file instead of having to set `.my.conf` for it to work

johnsaigle: It's possible to connect via the config file instead already. I was using the main user instead of the quatUser though. I'll update this.

johnsaigle:
@cmadjar @PapillonMcGill I've updated the script to use the quatUser.

PapillonMcGill:
The command to load the script into the DB are still using unix (lorisadmin) not the user from config.xml.
Plus, I don't want to enter the pwd 100 time.

johnsaigle:
You shouldn't have to. The script tries to load from a MySQL config file and falls back to the credentials in the configuration file. It also pulls the password from that file.

Have you tested the script? if so could you paste the output?

PapillonMcGill:
Yes, I tried to test the script. During that test, all calls to mysql where done with lorisadmin (Unix) user which of course doesn't exist in the DB and failed. None of the call used the userID/PWD combination in config.xml

johnsaigle:
@cmadjar @PapillonMcGill I properly added the ability to authenticate via the config file in the latest commit. Unfortunately this has the side-effect of generating many mysql warnings, but it's now possible.

johnsaigle:
I'm cancelling Travis manually to facilitate 22 testing. This needs to be restarted later.

johnsaigle:
Rebased. Build cancelled manually to facilitate release testing. Travis needs to be restarted.

PapillonMcGill:
Just 1 thing, If you can find a way to clear `STDIN` buffer before reading the new admin password (or checking for empty string) that would be great.

I had it enter while the script was running and the cursor was in that shell and at the end the script used the empty string for new password.

johnsaigle:
This script uses the `resetpassword.php` script in `tools/` to reset the password. Maybe a separate issue can be opened to add validation to that script? Though it seems to be designed without validation on purpose.

ridz1208:
@johnsaigle another addition to the script:

After the script drops all tables, it should check if any tables are left over in the database. I think it should at least list the remaining tables or if you wanna get fancy you could ask the user if they want to continue or stop the script.

Usecase: the RB refresh script is particularly useful when switching branches/versions. Given that between different versions some extra tables might be created or renamed or deleted, if the refresher script is run it might encounter a table created in a different version and thus not existing in the 9999 deletion scripts....

Not sure if it's clear but I can clarify in person

johnsaigle:
Screen Shot 2019-12-09 at 12 35 44

@ridz1208 I incorporated your feedback. The script now outputs a colourful warning and prints the tables remaining in the database after the drop command executes.

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,264 @@
...
+exec(
+    "$mysqlCommand -e 'SHOW TABLES;'",
+    $tables,
+    $status
+);
+array_shift($tables); // remove the column name
+if (count($tables) > 0) {
+    echo "\e[33mWARNING: Untracked tables still exist in the database:\e[0m\n";
+    array_walk($tables, 'printBulletPoint');
+}
```
PapillonMcGill: @johnsaigle I would ask the user if he want to drop the tables automatically and delete the tables if yes.
Optimally, in that case, the script should do a while loop to drop the tables until table count == 0 as some `drop table` might fail due to foreign key and require the referring table to be dropped first.

maltheism: @johnsaigle @PapillonMcGill would `SET FOREIGN_KEY_CHECKS = 0;` solve the issue of failing as a result of a foreign key. I've had to run that query before deleting tables in some situations.

PapillonMcGill: Yes it will solve for FK dependencies blocking drop instruction but not for files not in 9999 script.

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,286 @@
...
+        echo "Couldn't config setting. " .
+            "This may need to be manually if your LORIS is not hosted at " .
+            "localhost." .
+            PHP_EOL;
+    }
+}
+
+function dropRemainingTables(array $tables) {
+    global $mysqlCommand;
+    $commands = ["SET FOREIGN_KEY_CHECKS = 0;"];
```
PapillonMcGill: is this a new way to add to an array? Same for L270

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,286 @@
...
+    foreach ($tables as $table) {
+        $commands[] = "DROP TABLE IF EXISTS $table;";
+    }
+    $commands = ["SET FOREIGN_KEY_CHECKS = 1;"];
```
johnsaigle:
```suggestion
    $commands[] = "SET FOREIGN_KEY_CHECKS = 1;";
```

tools/raisinbread_refresh.php
johnsaigle: This initializes an array with one element. Line 270 had a typo -- thanks for drawing attention to it. I've fixed it now.

Makefile
```diff
@@ -32,3 +32,7 @@ unittests: phpdev
 
 # Perform all tests that don't require an install.
 check: checkstatic unittests
+
+raisinbread:
```
johnsaigle:
```suggestion
testdata:
```

Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: clean dev all check checkstatic unittests test phpdev javascript
+.PHONY: clean dev all check checkstatic unittests test phpdev javascript raisinbread
```
johnsaigle:
```suggestion
.PHONY: clean dev all check checkstatic unittests test phpdev javascript testdata
```

johnsaigle:
@driusan @ridz1208 Please re review when you have some time.

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,286 @@
...
+// supplied via a command-line argument.
+exec($mysqlCommand . ' -e "show tables;" 2>&1 1>/dev/null', $output, $status);
+if ($status != 0) {
+    $mysqlCommand = <<<CMD
+mysql -A "$dbname" -u "$username" -h "$host" -p$password
+CMD;
+}
+
+// Drop tables
+echo PHP_EOL .'Dropping LORIS tables....' . PHP_EOL;
```
cmadjar: The script outputs the following:
```
Please type the database name `loris` to confirm you wish to drop tables
and import test data: loris

Dropping LORIS tables....
...
[*] Creating LORIS core tables.... 
...
[*] Creating instrument tables.... 
...
[*] Importing Raisinbread data....
...
Restoring config setting `url` to value `http://localhost:8080`

Restoring config setting `base` to value `/var/www/loris/`

Restoring config setting `host` to value `localhost`

Please choose a new password for the admin user:
Running command: `php resetpassword.php admin`
```
However, my OCD would be very happy if headers of the different sections were consistant :): (for example:)
```
Please type the database name `loris` to confirm you wish to drop tables
and import test data: loris

[*] Dropping LORIS tables....
...

[*] Creating LORIS core tables.... 
...

[*] Creating instrument tables.... 
...

[*] Importing Raisinbread data....
...

[*] Restoring config settings...
Restoring config setting `url` to value `http://localhost:8080`
Restoring config setting `base` to value `/var/www/loris/`
Restoring config setting `host` to value `localhost`

[*] Changing the admin password...
Running command: `php resetpassword.php admin`
```
In short: always have a new line before a section, have each sections identified with `[*]` before it, no empty lines between the different restoring config setting messages, adding a section name for the change to the admin password.

Does that make sense?

johnsaigle: Yep, I'll do that

johnsaigle:
@cmadjar done! I also made the headers green so they really stand out.

tools/raisinbread_refresh.php
```diff
@@ -95,7 +95,7 @@
 
 
 echo <<<CONFIRMATION
-Please type the database name `$dbname` to confirm you wish to drop tables
+Please type the database name `$dbname` to confirm you wish to drop 3ables
```
driusan: This change looks like a step backwards.

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,286 @@
...
+            "prevent accidental data loss."
+        );
+    }
+    $dbname = $dbInfo['database'];
+    $host = $dbInfo['host'];
+    // The "quat" user and password should be configured to a user with DROP
+    // and CREATE permissions for the database.
+    $username = $dbInfo['quatUser'];
+    $password = $dbInfo['quatPassword'];
+
```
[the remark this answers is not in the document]
johnsaigle: These should be `adminUser`/`adminPassword` after a recent PR. I'll update them.

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,286 @@
...
+function dropRemainingTables(array $tables) {
+    global $mysqlCommand;
+    $commands = ["SET FOREIGN_KEY_CHECKS = 0;"];
+    foreach ($tables as $table) {
+        $commands[] = "DROP TABLE IF EXISTS $table;";
+    }
+    $commands[] = "SET FOREIGN_KEY_CHECKS = 1;";
+    $script = <<<SCRIPT
+$mysqlCommand -e '%s'
+SCRIPT;
```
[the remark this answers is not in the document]
johnsaigle: I think as of PHP 7.3 it's customizable so I'll fix this

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,286 @@
...
+// deleted by the above script. Issues with e.g. foreign keys can be a source
+// of error and so it is useful to know what tables remain.
+exec(
+    "$mysqlCommand -e 'SHOW TABLES;'",
+    $tables,
+    $status
+);
+array_shift($tables); // remove the column name
+if (count($tables) > 0) {
+    echo "\e[33mWARNING: Untracked tables still exist in the database:\e[0m\n";
```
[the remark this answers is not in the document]
johnsaigle: I have other code that works as a tools wrapper but it was included in a PR that was never merged.

I'll open a new PR for it because I agree it would be helpful.

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,286 @@
...
+    die("Could not load project/config.xml");
+}
+
+
+echo <<<CONFIRMATION
+Please type the database name `$dbname` to confirm you wish to drop tables
+and import test data: 
+CONFIRMATION;
+
+$input = trim(fgets(STDIN));
```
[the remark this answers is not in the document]
johnsaigle: I agree it's kinda weird but I left it there as an extra safety check so that data doesn't get deleted. I'll change it though.

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,286 @@
...
+
+INFO;
+
+echo $info;
+
+echo "\e[0;31m*** Executing this script will result in the LOSS OF DATA! ***\e[0m\n";
+
+$cwd = getcwd();
+if (substr_compare($cwd, 'tools', mb_strlen($cwd) - mb_strlen('tools')) !== 0) {
+    die('Please run this script from the tools/ directory.' . PHP_EOL);
```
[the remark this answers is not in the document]
johnsaigle: Most of our tools work like this. It allows for easier use of relative paths.

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,286 @@
...
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris-Trunk/
+ */
+
+$info = <<<INFO
+This script is used by LORIS developers to DELETE DATA and replace
+them with new test data.
+
+
+INFO;
```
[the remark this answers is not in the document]
johnsaigle: I prefer heredoc syntax because it's easier to read a paragraph like that and you don't have to worry about matching quotation marks, adding `.` append characters and managing line length.

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,286 @@
...
+        global $DB;
+        $DB->run(
+            "UPDATE Config c 
+            SET c.Value='$value'
+            WHERE c.ConfigID
+            IN (SELECT cs.ID FROM ConfigSettings cs WHERE cs.Name = '$name')"
+        );
+    } catch (\DatabaseException $e) {
+        echo "Couldn't config setting. " .
+            "This may need to be manually if your LORIS is not hosted at " .
```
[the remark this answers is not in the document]
johnsaigle: It's a catch for if the `host`, `base`, etc. doesn't work for whatever reason

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,290 @@
...
```
johnsaigle:
```suggestion
            "This may need to be set manually if your LORIS is not hosted at " .
```

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,286 @@
...
+    restoreConfigSetting('base', $baseConfigSetting);
+}
+if (isset($hostConfigSetting)) {
+    restoreConfigSetting('host', $hostConfigSetting);
+}
+
+// Trigger a password reset because the password for `admin` in the Raisinbread
+// database is public.
+echo 'Please choose a new password for the admin user:' . PHP_EOL;
+runCommand('php resetpassword.php admin');
```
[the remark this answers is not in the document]
johnsaigle: There's a prompt on line 175. I don't understand why we need another one

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,290 @@
...
+$input = trim(fgets(STDIN));
+if ($input !== $dbname) {
+    die('Input did not match database name. Exiting.');
+}
+
+$mysqlCommand = <<<CMD
+mysql -A $dbname
+CMD;
+
+// Test whether a connection to MySQL is possible via a config file.
```
johnsaigle:
```suggestion
// Test whether a connection to MySQL is possible via a MySQL config file.
```

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,290 @@
...
+
+try {
+    require_once 'generic_includes.php';
+
+    $dbInfo = $config->getSettingFromXML('database');
+
+    if (! $config->getSetting('dev')['sandbox']) {
+        throw new \LorisException(
+            "Config file indicates that this is not a sandbox. Aborting to " .
+            "prevent accidental data loss."
```
johnsaigle:
```suggestion
            "prevent accidental data loss." . PHP_EOL
```

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,286 @@
...
+// Restore config settings if they were successfully found before.
+if (isset($urlConfigSetting)) {
+    restoreConfigSetting('url', $urlConfigSetting);
+}
+if (isset($baseConfigSetting)) {
+    restoreConfigSetting('base', $baseConfigSetting);
+}
+if (isset($hostConfigSetting)) {
+    restoreConfigSetting('host', $hostConfigSetting);
+}
```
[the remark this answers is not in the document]
johnsaigle: Screen Shot 2020-01-15 at 14 25 26


I'm not able to reproduce this.

tools/raisinbread_refresh.php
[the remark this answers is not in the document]
johnsaigle: Screen Shot 2020-01-15 at 14 24 36

tools/raisinbread_refresh.php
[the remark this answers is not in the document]
ridz1208: most of the old/uncleaned ones do. as for relative paths you can just prepend the `__DIR__` variable and it will turn relative paths into absolute ones ish.

I think it's better to do that because when tools are being used in cronjobs we dont need to jump in and out of directories through the cron script.

tools/raisinbread_refresh.php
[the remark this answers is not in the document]
ridz1208: because the `Running command:...` displays after the `Please choose...`

I waited a good 5 min before realising that i had to enter the password now.

If these 2 lines can be inversed or the `Please choose ...` can be removed entirely and have the resetpassword script promt itself for the new password would make more sens

tools/raisinbread_refresh.php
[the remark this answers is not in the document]
ridz1208: mmm the error message doesnt seem to be relevant to the issue itself we can look at that one in more details

tools/raisinbread_refresh.php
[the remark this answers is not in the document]
ridz1208: yes. but thats not what I was talking about.

If you intentionally remove the adminuser and admin password but you have a mysql config file
there should not be a notice saying they are not defined because you don't really need them

ridz1208:
@cmadjar I dismissed your old review. please review again.

ridz1208:
@johnsaigle did you by any chance address the issue reported in my 6th test case (see my last review comment)?

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,316 @@
...
+ * After dropping the tables, the script will source the Raisinbread test data
+ * using the commands found in raisinbread/README.md.
+ *
+ * The script also restores the url, base, and host config settings to their
+ * pre-drop values. This prevents an issue where a developer will need to
+ * manually change these values in a MySQL shell when they are not hosting a
+ * LORIS on localhost.
+ *
+ * Finally, the script runs tools/resetpassword.php so that the default admin
+ * password with the Raisinbread data set is not used.
```
driusan: I don't understand this part.. does this mean every time someone runs this script the password on raisinbread changes?

tools/raisinbread_refresh.php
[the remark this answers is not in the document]
driusan: PHP supports multiline strings. You could just use `"`. (I don't care in this particular instance, just FYI.)

tools/raisinbread_refresh.php
ridz1208: thtas the thing. yes if the users table is then submitted back to the repo... might be problematic

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,316 @@
...
+    echo 'Continue attempting to install Raisinbread database? (y/N)' . PHP_EOL;
+    $input = trim(fgets(STDIN));
+
+    if (mb_strtolower($input) !== 'y') {
+        die;
+    }
+    echo PHP_EOL . 'Please enter the name of your database:' . PHP_EOL;
+    $dbname = trim(fgets(STDIN));
+} catch (\LorisException $e) {
+    die($e->getMessage());
```
driusan: The only reason to catch this seems to be the unnecessary throw above which makes the control flow convoluted.

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,316 @@
...
+}
+
+// Drop tables
+echo printHeader('Dropping LORIS tables.');
+dropTables();
+
+// Print the names of remaining tables, if any. Some tables may remain if they
+// have been created at a different time during development and are not
+// deleted by the above script. Issues with e.g. foreign keys can be a source
+// of error and so it is useful to know what tables remain.
```
driusan: I'm reviewing the script from the top-down so maybe I'll be elucidated later, but I don't really understand this comment. What's the difference between a "LORIS table" and a "remaining table"? A different time during development of what?

tools/raisinbread_refresh.php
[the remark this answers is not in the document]
driusan: I had no problem with the first one, but this is just abusing heredocs. It's not even a multiline string and it has no special characters in it.

tools/raisinbread_refresh.php
PapillonMcGill: @driusan Instruments table. These are not deleted by the 9999-99-99 script and prevent come LORIS table to be deleted due to foreign keys.

ridz1208: @PapillonMcGill not exactly. There is a 9999 deletion script for instruments

The "remaining tables" are basically tables that were created at a later stage of development. 
Imagine PR 1 is open and still being reviewed and PR 5 adds a new table `BLA`, I come and test PR 5 so I create the `BLA` table on my database. then I try to test PR 1, but this tool will not delete `BLA` because the 9999 script on PR 1 does not contain a drop statement for `BLA`. Now imagine `BLA` has a foreign key reference on candidate, then not only will `BLA` still be there but it will cause this tool to fail entirely when deleting the candidate table.

So the tool will need to be flexible to be able to work on branches at different points in time

PapillonMcGill: @ridz1208 Then why when I tried it, Instruments table where not deleted and prevented some LORIS tables to be deleted due to foreign keys?

ridz1208: @PapillonMcGill line 194 is the deletion for instruments

and if you have your own it's not technically the RB dataset... so yes this would help you, but that's not why it is necessary and finally there are usually not foreign key constraints between instruments and the rest of core loris tables.

PapillonMcGill: @ridz1208 It didn't ran or it tried but failed for any reason without any error message when I tried to test this PR.

tools/raisinbread_refresh.php
johnsaigle: When you source Raisinbread, your admin user will have its password set to the public password contained in the Raisinbread dataset. This comment explains that the script will run `php resetpassword.php admin` and force you to change your password for your dev environment to a new, private password.

Yes there are potential complications when submitting changes to Raisinbread, but only if the `users` table is updated which is the one containing the password. In this case, the author of the PR will need to change the RB file manually back to the public RB admin password. This is similar to the process whent he Config table is updated; the author needs to change e.g. `host` back to localhost in their PR.

tools/raisinbread_refresh.php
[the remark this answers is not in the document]
johnsaigle: OK, I'll change this.

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,316 @@
...
+    echo 'Could not connect to the database in the Config file.' .
+        'It\'s possible that the LORIS tables have already been dropped.' .
+        PHP_EOL;
+    echo 'Continue attempting to install Raisinbread database? (y/N)' . PHP_EOL;
+    $input = trim(fgets(STDIN));
+
+    if (mb_strtolower($input) !== 'y') {
+        die;
+    }
+    echo PHP_EOL . 'Please enter the name of your database:' . PHP_EOL;
```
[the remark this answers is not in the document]
johnsaigle: If you abort the script halfway through or some error occurs, it's possible that a DB exception can occur if LORIS tables have already been dropped. 

This prompt is a way to recover execution of the script when this happens.

tools/raisinbread_refresh.php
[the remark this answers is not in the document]
johnsaigle: OK I'll change it

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,286 @@
...
+if (isset($hostConfigSetting)) {
+    restoreConfigSetting('host', $hostConfigSetting);
+}
+
+// Trigger a password reset because the password for `admin` in the Raisinbread
+// database is public.
+echo 'Please choose a new password for the admin user:' . PHP_EOL;
+runCommand('php resetpassword.php admin');
+
+
```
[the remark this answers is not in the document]
johnsaigle: Done!

tools/raisinbread_refresh.php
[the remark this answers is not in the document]
johnsaigle: I'll change it

tools/raisinbread_refresh.php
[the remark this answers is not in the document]
johnsaigle: It's a fallback in case the credentials in the config file don't work.

This branch sets up the ability to connect using the command `mysql -A $dbname` below. This can happen if you have a MySQL configuration file set up. E.g. I have a user and password for MySQL configured in my home directory. If the DB info from the `project/config.xml` file doesn't work, this script prompts for the name of your LORIS test DB and tries to connect using your MySQL config instead.

tools/raisinbread_refresh.php
johnsaigle: Yeah Rida's explanation is correct. This part of the code is for two cases:

1. You've installed a table from another PR that isn't installed on the current branch (and as a corollary is not included in the deletion scripts). E.g. if I have `biobank` tables in my DB, this makes sure they get deleted.
2. You have non-standard tables that you've created. These might have FKs to core tables. This will cause the script to fail when it tries to delete core tables.

This part of the code should print and optionally delete all tables in your DB that aren't part of Raisinbread on the current branch.

tools/raisinbread_refresh.php
ridz1208: nothing prevents us from handling these fields in the dump script to make sure they are reset before the db is dumped !! that could be a nice addition beyond the scope of this PR

johnsaigle: Yeah I agree with you. I think it's a modification to the exporter script, not this one.

johnsaigle:
> did you by any chance address the issue reported in my 6th test case (see my last review comment)?

No I don't think I did. Thanks for the reminder.

johnsaigle:
@ridz1208 @driusan I changed the DB connection code quite a lot so that's it's more clear and also provides more output. Please test/review again and let me know what you think.

tools/raisinbread_refresh.php
PapillonMcGill: @ridz1208 A good example of why you have extra table and foreign key, is when testing PRs. Test a new PR that add new tables (and data using foreign keys to existing data) then switch to another PR and use this script to reset your DB. You have some tables that are still unknown to the 9999-99-99 script in the later PR.

ridz1208: lol @PapillonMcGill thats litterally my explanation from 7 days ago

https://github.com/aces/Loris/pull/5260#discussion_r368546801

PapillonMcGill: @ridz1208 from what I remembered, you seam to find that as not needed.

ridz1208: I dont beleive I ever found it not needed no. thus the `need` key work in my comment
```
So the tool will need to be flexible to be able to work on branches at different points in time
```

johnsaigle: Either way that feature exists so we should be good!

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,358 @@
...
+tables and import test data: 
+CONFIRMATION;
+
+$input = trim(fgets(STDIN));
+if ($input !== $dbname) {
+    die(printWarning('Input did not match database name. Exiting.'));
+}
+
+$mysqlCommand = <<<CMD
+mysql -A $dbname
```
driusan: heredocs are appropriate for formatting multi-line strings which are getting directly displayed to the user. This does not meet that criteria. This should just be a normal string.

Also, the command arguments should be properly shell escaped.

tools/raisinbread_refresh.php
```diff
@@ -0,0 +1,358 @@
...
+                "Could not connect to database. One of the settings: "
+                . "`host`, `database`, `username` or `password` "
+                . " are missing from your project's configuration file."
+            )
+        );
+    }
+
+    // Try connecting by supplying parameters on command line.
+    $mysqlCommand = <<<CMD
+mysql -A "$dbname" -u "$username" -h "$host" -p$password
```
driusan: Also not really a reasonable use of the style of string and not properly escaping the arguments.

tools/raisinbread_refresh.php
johnsaigle: Alright I'll change these.

driusan:
Conflicts

johnsaigle:
@driusan This should be good to go after Travis passes

christinerogers:
The Readme for Raisinbread includes how to load raisinbread.  This script should definitely be mentioned there.   Will make an issue to address for 23.

PR #5264, opened 2019-10-02
Fix PHP linter

johnsaigle:
## Brief summary of changes

This PR is identical to #5249 but sent to `major`. This fix is needed because it is causing failures on Reactification PRs (by @maltheism) written with new PHP subdirectories.


driusan:
can you just send a PR to push minor into major? We might as well get everything at once..

johnsaigle:
Adding Blocked as @ridz1208 asked me to wait until after the next bugfix release so that we can push that into minor first then then minor into bugfix.

Issue #5063, opened 2019-08-21
LORIS should start requiring PHP 7.3 on minor + bugfix (and 7.4 on major when it's out)


johnsaigle:
PHP 7.4 comes out at the end of November 2019. At the same time, PHP 7.2 will no longer have active support. 

https://www.php.net/supported-versions.php

On the major branch we should start using PHP 7.3 (and discontinue 7.2 testing) as well as enable PHP 7.4 testing on all branches once it's out.

Importantly, 7.4 will allow us to use strong types in PHP class variables.

Requiring all LORIS instances to use >= 7.3 also means they'll have guaranteed CSRF protection (#4083)

johnsaigle:
Requiring 7.4 for LORIS will also provide us with performance and security enhancements in newer versions of MySQL as discussed in #5603

PR #5265, opened 2019-10-02
Add PHP 7.4 (snapshot build) to TravisCI tests

johnsaigle:
## Brief summary of changes

PHP 7.4 will be released in 2 months, at the end of November. It would be nice to move LORIS toward this as a minimum version requirement for LORIS 22. #5063 

At the same time PHP 7.2 will stop receiving active support. https://www.php.net/supported-versions.php

Until its release, Travis offers a "snapshot" build which should be equivalent to the rc1 or rc2 of 7.4. This can be updated after the official release, but for now this gives us a chance to see how LORIS is doing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ services:
 install:
 - sudo apt-get update
 - sudo apt-get install -y libapparmor1
-- sudo apt install -y php-ast
+- pecl install -f ast-1.0.1
 - sudo apt-get install npm
 - make dev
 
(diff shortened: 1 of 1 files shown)


johnsaigle:
7.4 seems to be failing due to some issues with PECL and php-ast. As a result I'm marking this as Blocked until #5266 or #4433 is merged.

driusan:
since it's not released yet can you add it to the allowed failures list in .travis.yml?

PR #5270, opened 2019-10-03
[RBData/Candidate] Remove multiple first visits in session

zaliqarosli:
## Brief summary of changes

Follows  up on suggestion made in https://github.com/aces/Loris/pull/4075 to catch error  

```
Attempt to use pselectRow on a query that returns multiple rows
```

thrown in `Database.class.inc`, when calling function Candidate->getFirstVisit()  in `assign_missing_instruments.php`, when multiple first visits are returned, and throw new LorisException with more meaningful message within the function getFirstVisit().

1. On minor branch, try `php assign_missing_instruments.php` and see if you get the error:

```
PHP Fatal error:  Uncaught DomainException: Attempt to use pselectRow on a query that returns multiple rows in /var/www/loris/php/libraries/Database.class.inc:818
```

If you do, switch to this PR's branch, and run `php assign_missing_instruments.php` again. You should instead get something similar to:

```
PHP Fatal error:  Uncaught LorisException: Candidate 300299 has multiple first visits recorded where only 1 is expected. in /var/www/loris/php/libraries/Candidate.class.inc:622
```

2. If you don't have a candidate in your database that has multiple first visits (you don't get the above errors), mock one by duplicating values in 'VisitNo' column in `session` table for a candidate, and repeat step 1.

https://github.com/aces/Loris/pull/4075


ridz1208:
@zaliqarosli it's a good catch but if you are going to throw an exception, you have to submit a php tool to automatically renumber the visits. otherwise projects would get random sporadic failures on production.

you can test out that tool on raisinbread

ridz1208:
@zaliqarosli the alternative would be to call the PR, "clarify error message" because this does not seem to be "Remov[ing] multiple first visits in session"

zaliqarosli:
@ridz1208 i had initially removed the multiple first visits in raisinbread as a part of this PR but omitted the change and forgot to rename the PR accordingly. I would like to still change it in raisinbread but a previous PR to fix up the imaging data on major had actually done the reverse of this - change 2nd visits to first visits. I had made this PR originally on minor so thought it would be too much work to rebase those DB changes onto major, plus noticing that what i thought was "bad data" might have actually been done intentionally to flag things in testing

ridz1208:
@zaliqarosli I see.

Now the PR will be going to a signle development branch so that should not be an issue anymore. I would recommend that you add a script for existing projects and the changes for the RB files. that would be the ideal scenario

zaliqarosli:
@ridz1208 check this out

tools/single_use/Cleanup_multiple_firstVisits.php
```diff
@@ -0,0 +1,65 @@
...
+                  GROUP BY CandID 
+                  HAVING COUNT(CandID) > 1";
+$candIDs      = $db->pselect($query_candID, array());
+
+if (empty($candIDs)) {
+    echo "No multiple first visits.\n";
+} else {
+    foreach ($candIDs as $entry) {
+        // Order session IDs by date of visit
+        echo "Multiple first visits detected for $entry[CandID]\n";
```
zaliqarosli:
```suggestion
        echo "Multiple first visits detected for CandID $entry[CandID]\n";
```

can i be annoying and add this? besides that, its all good

zaliqarosli:
approve and tested! (can't approve my own PR)

PR #5272, opened 2019-10-04
added isset() check in if statement

PapillonMcGill:
## Brief summary of changes
add a check to ensure array  element exist before comparing value

Without, script assign_missing_instruments.php raised a PHP notice.

1.
Can run php tools/assign_missing_instruments.php as it use this part of the code.


zaliqarosli:
i suggest closing this PR because 'options' index should always be set. if it's not set, it's a data issue

PR #5280, opened 2019-10-04
[jsx/Form.js] new slider component! 

maltheism:
## Brief summary of changes

- I created a "slider" component for the form in LORIS.
- The component is named: `<SliderElement/>`
- Also it will resize well with the browser in the form!

**Note:** I kept the design simple for now.

Preview:
example_component_slider

* https://github.com/aces/instrument-builder/issues/56
* https://github.com/aces/instrument-builder/issues/14

1. Checkout PR.
2. Add to a form.
3. Test away!

Example replace modules/new_profile/jsx/NewProfileIndex.js with:
```
import Panel from 'Panel';
import Loader from 'Loader';
import swal from 'sweetalert2';

/**
 * New Profile Form
 *
 * Create a new profile form
 *
 * @author  Shen Wang
 * @version 1.0.0
 * */
class NewProfileIndex extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      configData: {},
      formData: {},
      newData: {},
      isLoaded: false,
      isCreated: false,
      error: false,
      sliderValue: 0,
    };
    this.handleSubmit = this.handleSubmit.bind(this);
    this.setFormData = this.setFormData.bind(this);
    this.fetchData = this.fetchData.bind(this);
    this.handleSlider = this.handleSlider.bind(this);
  }

  componentDidMount() {
    this.fetchData()
      .then(() => this.setState({isLoaded: true}));
  }

  /**
   * Retrieve data from the provided URL and save it in state
   *
   * @return {object}
   */
  fetchData() {
    return fetch(this.props.dataURL,
      {credentials: 'same-origin'})
      .then((resp) => resp.json())
      .then((data) => this.setState({configData: data.fieldOptions}))
      .catch((error) => {
        this.setState({error: true});
      });
  }
  /**
   * It checks the date of birth and Expected Date of Confinement,
   * the date fields must match.
   * If match, this function will return true.
   *
   * @return {boolean}
   */
  validateMatchDate() {
    let validate = false;
    const formData = this.state.formData;
    if (formData.dobDate !== formData.dobDateConfirm) {
      swal('Error!', 'Date of Birth fields must match', 'error');
    } else if (this.state.configData['edc'] === 'true' &&
         (formData.edcDate !== formData.edcDateConfirm)
    ) {
      swal('Error!', 'EDC fields must match', 'error');
    } else {
      validate = true;
    }
    return validate;
  }

  /**
   * Handles form submission
   *
   * @param {event} e - Form submission event
   */
  handleSubmit(e) {
    e.preventDefault();
    const match = this.validateMatchDate();
    if (!match) {
      this.setState({
        isCreated: false,
      });
    } else {
      let formData = this.state.formData;
      let formObject = new FormData();
      for (let key in formData) {
        if (formData[key] !== '') {
          formObject.append(key, formData[key]);
        }
      }
      formObject.append('fire_away', 'New Candidate');

      fetch(this.props.submitURL, {
        method: 'POST',
        cache: 'no-cache',
        credentials: 'same-origin',
        body: formObject,
        })
      .then((resp) => {
        if (resp.ok && resp.status === 201) {
          resp.json().then((data) => {
            this.setState({newData: data});
            this.setState({isCreated: true});
          });
        } else {
          resp.json().then((message) => {
            swal('Error!', message, 'error');
          });
        }
      })
      .catch((error) => {
        console.error(error);
      });
    }
  }

  /**
   * Set the form data based on state values of child elements/components
   *
   * @param {string} formElement - name of the selected element
   * @param {string} value - selected value for corresponding form element
   */
  setFormData(formElement, value) {
    let formData = Object.assign({}, this.state.formData);
    formData[formElement] = value;

    this.setState({formData: formData});
  }

  handleSlider(formElement, value) {
    const state = Object.assign({}, this.state);
    state.sliderValue = value;
    this.setState(state);
  }

  render() {
    // If error occurs, return a message.
    if (this.state.error) {
      return <h3>An error occured while loading the page.</h3>;
    }

    // Waiting for async data to load
    if (!this.state.isLoaded) {
      return <Loader/>;
    }
    let profile = null;
    let edc = null;
    let pscid = null;
    let site = null;
    let minYear = this.state.configData.minYear;
    let maxYear = this.state.configData.maxYear;
    let dateFormat = this.state.configData.dobFormat;

    if (this.state.configData['edc'] === 'true') {
      edc =
        <div>
          <DateElement
            name = "edcDate"
            label = "Expected Date of Confinement"
            minYear = {minYear}
            maxYear = {maxYear}
            dateFormat = {dateFormat}
            onUserInput = {this.setFormData}
            value = {this.state.formData.edcDate}
            required = {true}
          />
          <DateElement
            name = "edcDateConfirm"
            label = "Confirm EDC"
            minYear = {minYear}
            maxYear = {maxYear}
            dateFormat = {dateFormat}
            onUserInput = {this.setFormData}
            value = {this.state.formData.edcDateConfirm}
            required = {true}
          />
        </div>;
    }
    if (this.state.configData['pscidSet'] === 'true') {
      pscid =
        <TextboxElement
          name = "pscid"
          label = "PSCID"
          onUserInput = {this.setFormData}
          value = {this.state.formData.pscid}
          required = {true}
        />;
    }
    if (this.state.configData['site'] !== null) {
      site =
        <SelectElement
          name = "site"
          label = "Site"
          options = {this.state.configData.site}
          onUserInput = {this.setFormData}
          value = {this.state.formData.site}
          required = {true}
        />;
    }
    if (!this.state.isCreated) {
      profile = (
        <FormElement
          name = "newProfileForm"
          onSubmit = {this.handleSubmit}
        >
          <DateElement
            name = "dobDate"
            label = "Date of Birth"
            minYear = {minYear}
            maxYear = {maxYear}
            dateFormat = {dateFormat}
            onUserInput = {this.setFormData}
            value = {this.state.formData.dobDate}
            required = {true}
          />
          <DateElement
            name = "dobDateConfirm"
            label = "Date of Birth Confirm"
            minYear = {minYear}
            maxYear = {maxYear}
            dateFormat = {dateFormat}
            onUserInput = {this.setFormData}
            value = {this.state.formData.dobDateConfirm}
            required = {true}
          />
          {edc}
          <SelectElement
            name = "sex"
            label = "Sex"
            options = {this.state.configData.sex}
            onUserInput = {this.setFormData}
            value = {this.state.formData.sex}
            required = {true}
          />
          {site}
          {pscid}
          <SelectElement
            name = "project"
            label = "Project"
            options = {this.state.configData.project}
            onUserInput = {this.setFormData}
            value = {this.state.formData.project}
            required = {true}
          />
          <SliderElement
            name={'mySlider'}
            label={'Slider example'}
            id='mySlider'
            value={this.state.sliderValue}
            valueMin={0}
            valueMax={10}
            required={true}
            onUserInput={this.handleSlider}
          />
          <ButtonElement
            name = "fire_away"
            label = "Create"
            id = "button"
            type = "submit"
          />
        </FormElement>
      );
    } else {
      profile = (
        <div>
          <p>New candidate created. DCCID: {this.state.newData.candID} PSCID: {this.state.newData.pscid} </p>
          <p><a href = {'/' + this.state.newData.candID}> Access this candidate </a></p>
          <p><a href = "/new_profile/" > Recruit another candidate </a></p>
        </div>
      );
    }
    return (<Panel title="Create a new profile">{profile}</Panel>);
  }
}
  ReactDOM.render(
    <NewProfileIndex
      dataURL = {`${loris.BaseURL}/new_profile/?format=json`}
      submitURL = {`${loris.BaseURL}/new_profile/`}
      hasPermission = {loris.userHasPermission}
    />,
    document.getElementById('lorisworkspace')
  );
});

```

### DIFF
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -1792,6 +1792,28 @@ class SliderElement extends React.Component {
             width: '100%',
           }}>
             <div style={{
+              flexGrow: 1000,
+              display: 'flex',
+              flexDirection: 'column',
+              flexBasis: '100%',
+              maxWidth: this.props.maxWidth,
+              marginRight: '5px',
+              flex: 2,
+            }}>
+              <input
+                type='range'
+                name={this.props.name}
+                id={this.props.id}
+                value={this.props.value}
+                min={this.props.minValue}
+                max={this.props.maxValue}
+                required={required}
+                disabled={disabled}
+                onChange={this.handleChange}
+                style={{width: '100%'}}
+              />
+            </div>
+            <div style={{
               display: 'flex',
               flexDirection: 'column',
               flexBasis: '100%',
@@ -1813,26 +1835,6 @@ class SliderElement extends React.Component {
                 }}
               />
             </div>
-            <div style={{
-              flexGrow: 1000,
-              display: 'flex',
-              flexDirection: 'column',
-              flexBasis: '100%',
-              flex: 2,
-            }}>
-              <input
-                type='range'
-                name={this.props.name}
-                id={this.props.id}
-                value={this.props.value}
-                min={this.props.minValue}
-                max={this.props.maxValue}
-                required={required}
-                disabled={disabled}
-                onChange={this.handleChange}
-                style={{width: '100%'}}
-              />
-            </div>
           </div>
         </div>
       </div>
@@ -1846,6 +1848,7 @@ SliderElement.propTypes = {
   value: PropTypes.number.isRequired,
   minValue: PropTypes.number.isRequired,
   maxValue: PropTypes.number.isRequired,
+  maxWidth: PropTypes.string.isRequired,
   disabled: PropTypes.bool,
   required: PropTypes.bool,
   errorMessage: PropTypes.string,
@@ -1859,6 +1862,7 @@ SliderElement.defaultProps = {
   value: 0,
   minValue: 0,
   maxValue: 100,
+  maxWidth: 'auto',
   disabled: false,
   required: false,
   errorMessage: '',
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -1741,8 +1741,8 @@ class LorisElement extends Component {
  * React wrapper for a <input type='range'> element.
  */
 class SliderElement extends React.Component {
-  constructor() {
-    super();
+  constructor(props) {
+    super(props);
     this.handleChange = this.handleChange.bind(this);
   }
 
@@ -1750,12 +1750,12 @@ class SliderElement extends React.Component {
     // Handles empty, min & max cases.
     const inputValue = e.target.value
       ? parseFloat(e.target.value)
-      : this.props.valueMin;
-    let value = inputValue > this.props.valueMax
-      ? this.props.valueMax
+      : this.props.minValue;
+    let value = inputValue > this.props.maxValue
+      ? this.props.maxValue
       : inputValue;
-    value = value < this.props.valueMin
-      ? this.props.valueMin
+    value = value < this.props.minValue
+      ? this.props.minValue
       : value;
     this.props.onUserInput(this.props.name, value);
   }
@@ -1802,8 +1802,8 @@ class SliderElement extends React.Component {
                 type='number'
                 name={'input_' + this.props.name}
                 value={this.props.value}
-                min={this.props.valueMin}
-                max={this.props.valueMax}
+                min={this.props.minValue}
+                max={this.props.maxValue}
                 required={required}
                 disabled={disabled}
                 onChange={this.handleChange}
@@ -1825,8 +1825,8 @@ class SliderElement extends React.Component {
                 name={this.props.name}
                 id={this.props.id}
                 value={this.props.value}
-                min={this.props.valueMin}
-                max={this.props.valueMax}
+                min={this.props.minValue}
+                max={this.props.maxValue}
                 required={required}
                 disabled={disabled}
                 onChange={this.handleChange}
@@ -1844,8 +1844,8 @@ SliderElement.propTypes = {
   name: PropTypes.string.isRequired,
   label: PropTypes.string.isRequired,
   value: PropTypes.number.isRequired,
-  valueMin: PropTypes.number.isRequired,
-  valueMax: PropTypes.number.isRequired,
+  minValue: PropTypes.number.isRequired,
+  maxValue: PropTypes.number.isRequired,
   disabled: PropTypes.bool,
   required: PropTypes.bool,
   errorMessage: PropTypes.string,
@@ -1857,8 +1857,8 @@ SliderElement.defaultProps = {
   name: null,
   label: null,
   value: 0,
-  valueMin: 0,
-  valueMax: 100,
+  minValue: 0,
+  maxValue: 100,
   disabled: false,
   required: false,
   errorMessage: '',
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -1848,7 +1848,7 @@ SliderElement.propTypes = {
   value: PropTypes.number.isRequired,
   minValue: PropTypes.number.isRequired,
   maxValue: PropTypes.number.isRequired,
-  maxWidth: PropTypes.string.isRequired,
+  maxWidth: PropTypes.string,
   disabled: PropTypes.bool,
   required: PropTypes.bool,
   errorMessage: PropTypes.string,
@@ -1857,11 +1857,6 @@ SliderElement.propTypes = {
 };
 SliderElement.defaultProps = {
   id: null,
-  name: null,
-  label: null,
-  value: 0,
-  minValue: 0,
-  maxValue: 100,
   maxWidth: 'auto',
   disabled: false,
   required: false,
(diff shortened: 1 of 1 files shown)


jsx/Form.js
```diff
@@ -1736,6 +1736,138 @@ class LorisElement extends Component {
...
+            width: '100%',
+          }}>
+            <div style={{
+              display: 'flex',
+              flexDirection: 'column',
+              flexBasis: '100%',
+              maxWidth: '50px',
+              flex: 1,
+            }}>
+              <input
```
zaliqarosli: could we have the value box come after the slider? format/styling wise as well, i think it would be nice to have the slider element take half the space when big (elements can layer when the browser size is small), and float the input box to the left

maltheism: Hey @zaliqarosli, can you expand on why it would be better to have the input on the right. I think it looks more aesthetically pleasing on the left compared to the right. Also when I go with my mouse to type into an input box. I find myself moving the mouse nearest to the left of the input and not the right. No big deal for me if you desire the right though. I just find myself always going to the far left of the input field when clicking on it.

Lastly can you explain more about "slider element take half the space when big (elements can layer when the browser size is small), and float the input box to the left" as I'm not certain when looking at the component when resizing the browser.

I'll include screenshots of current implementation. My thoughts right now are the current implementation saves vertical space when having the input next to the slider and even on the smallest resize of chrome there is a lot to work with for the slider. I would think that's better as typically vertical space is valuable. Interested to read your thoughts.

example size

zaliqarosli: for your first question, the way the form is set up makes the user interact with the form and its elements from top to bottom, and left to right. so it makes sense that you would automatically go to the left of the field before going to the right. because the range input is the only really important part of this element (the slider should be the main focus of a SliderElement), it should come first. the user should interact with it first before choosing to use the direct number input if they so desire. if we're encouraging users to use the number input first and ignore the slider, than there's no point in adding this new element, and we could use the NumericElement component instead. the value box is a display of the value at the very least, but is also additionally an input option

zaliqarosli: for the second point, when the browser is at full size, i just think the slider is way bigger than is necessary. so, at full size (left example in screenshot), i think the slider can take up half the space, and when not at full size (right example in screenshot), the slider can take up the whole space.

i also think there should be some margin/padding between the value box and the slider. they are very close to each other right now. they should have the same spacing as what's to the left and right of all those input fields.

maltheism: @zaliqarosli I think it's hard to know for certain since we're kind of postulating on the left being preferred. I'll move it to the right because I don't use sliders much to know what makes sense and I never use them to be certain about this situation.

jsx/Form.js
```diff
@@ -1736,6 +1736,138 @@ class LorisElement extends Component {
...
+      </div>
+    );
+  }
+}
+SliderElement.propTypes = {
+  id: PropTypes.string,
+  name: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
+  value: PropTypes.number.isRequired,
+  valueMin: PropTypes.number.isRequired,
```
zaliqarosli: could we change to minValue and maxValue? or even just min and max to keep consistent with props of other elements

maltheism: @zaliqarosli I originally had min & max. I thought it would be better since valueMin & valueMax are more coherent to the min & max are being used for value. It doesn't bother me but I'm curious what you think with the previous said. I know you use vim but as a developer with an IDE if you start typing "value" all the other prop types with value show up and helps in aiding what else. In any case I don't really mind.

jsx/Form.js
zaliqarosli: I would appreciate it if you could just trust me on that first point.

for the second point, it is true that all the other fields don't need to be as big as they are. if you are not willing to change the size of the slider, then maybe it would be best to give the person implementing the component the option to set the size via a prop.

seeing as the first use case of the slider is the new instrument builder, it would be nice if I could use this component the way I need to use it. thanks for your help!

jsx/Form.js
zaliqarosli: if you don't mind, i would prefer minValue and maxValue please. thank you!

jsx/Form.js
maltheism: Hey @zaliqarosli I trust you!!  I wasn't trying to dismiss your point about the slider can be too large in width. I'm just trying to clarify everything before modifying the css. I'm working from home today because of a cold with an appointment I have to be at and otherwise I would just clarify everything in person. I'll try adding a prop that manages the max width because then the component can be used in other cases without looking giant.

zaliqarosli: thank you!

maltheism:
Hi @zaliqarosli, I added the changes requested and it's ready for review again. Let me know if I missed something or anything comes to mind.

jsx/Form.js
```diff
@@ -1736,6 +1736,142 @@ class LorisElement extends Component {
...
+  maxWidth: PropTypes.string.isRequired,
+  disabled: PropTypes.bool,
+  required: PropTypes.bool,
+  errorMessage: PropTypes.string,
+  elementClass: PropTypes.string,
+  onUserInput: PropTypes.func,
+};
+SliderElement.defaultProps = {
+  id: null,
+  name: null,
```
zaliqarosli: if a prop is required, i don't think we should provide it a default value

Issue #5268, opened 2019-10-03
PSCID generation not supporting `static` anymore


PapillonMcGill:
new code for generating PSCID do not include option for `static`
```
         <PSCID>                                                                 
              <!-- PSCID (Study Center ID) generation method possible options: sequential/random/user -->
              <generation>sequential</generation>                                 
              <structure>                                                         
                    <seq type="static">STATIC_TEXT</seq>
...
```

johnsaigle:
https://github.com/aces/Loris/blob/1c3949246482a44dec96382941004398b7cb8e39/php/libraries/SiteIDGenerator.php#L200-L221

Can you provide a more detailed bug report?

PR #5281, opened 2019-10-05
[Core] Fix static PSCID generation

ridz1208:
## Brief summary of changes
The `static` PSCID generation was not working anymore because the `$prefix` was never null since the site abbreviation always exist. the logic of choosing `siteAbbrev` vs `static` was moved in this PR to the if/else statement of the `prefix case`

1. modify your config.xml to use a static option.
2. check that the static string gets prepended
3. change the config.xml to use the siteAbbrev
4. check that the correct site Alias gets prepended

* PR changing the logic https://github.com/aces/Loris/pull/4241
* Resolves https://github.com/aces/Loris/issues/5268

### DIFF
--- a/php/libraries/SiteIDGenerator.php
+++ b/php/libraries/SiteIDGenerator.php
@@ -37,7 +37,7 @@ class SiteIDGenerator extends IdentifierGenerator
      *
      * @return void
      */
-    public function __construct(?string $prefix = null)
+    public function __construct(?string $siteAbbrevPrefix = null)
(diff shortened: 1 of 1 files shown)


johnsaigle:
At some point we should make sure to write unit tests to cover each of the PSCID generation methods.

johnsaigle:
@ridz1208 Dave isn't here today and this issue seems to be a problem so I think you should go ahead and merge it.

PR #5284, opened 2019-10-06
[Raisinbread] update migration log major

ridz1208:


driusan:
I'm just going to assume that these have, in fact, been applied to RB.

PR #5285, opened 2019-10-07
Add project to PSCID generation 

ridz1208:
## Brief summary of changes
Added option to generate PSCIDs using the project alias instead of the site alias. this is needed by several projects where permissions will be more project oriented then site.

to test:
- in your config.xml change between `static`, `siteAbbrev` and `projectAbbrev` make sure the correct PSCID gets generated

### DIFF
--- a/SQL/New_patches/2019-10-05-Add_alias_to_projects.sql
+++ b/SQL/New_patches/2019-10-05-Add_alias_to_projects.sql
@@ -1,1 +1,3 @@
-ALTER TABLE `Project` ADD COLUMN `Alias` char(4) NOT NULL AFTER `Name`;
+ALTER TABLE `Project` ADD COLUMN `Alias` char(4) DEFAULT NULL AFTER `Name`;
+UPDATE Project SET Alias=UPPER(LEFT(Name,4)) WHERE Alias IS NULL;
+ALTER TABLE `Project` CHANGE COLUMN `Alias` `Alias` char(4) NOT NULL;
--- a/SQL/New_patches/2019-10-05-Add_alias_to_projects.sql
+++ b/SQL/New_patches/2019-10-05-Add_alias_to_projects.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `Project` ADD COLUMN `Alias` char(4) NOT NULL AFTER `Name`;
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -272,7 +272,10 @@ class Candidate
         );
 
         if ($config->getSetting('useExternalID') === 'true') {
-            $externalID = (new \ExternalIDGenerator())->generate();
+            $externalID = (new \ExternalIDGenerator(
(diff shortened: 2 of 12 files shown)


php/libraries/Candidate.class.inc
```diff
@@ -191,7 +191,7 @@ class Candidate
      *                                      configured to be true.
      * @param string      $sex         sex, either 'Male' or 'Female'
      * @param string|null $PSCID       PSCID specified by the user, if available
-     * @param int|null    $projectID   The project ID, if available
+     * @param int         $projectID   The project ID, if available
```
johnsaigle: It might be nice to change this signature to actually pass a Project object instead of an ID here, if possible. We'd benefit from built-in validation this way.

Up to you though - I'm not sure how complicated this would be.

php/libraries/Project.class.inc
```diff
@@ -186,6 +224,34 @@ class Project
...
+     */
+    public function setAlias(string $projectAlias): void
+    {
+        if (strlen($projectAlias) > 4 ) {
+            throw new ConfigurationException(
+                "The length of the project's alias can not be longer than ".
+                "4 characters."
+            );
+        }
+        $this->_projectAlias = $projectAlias;
```
johnsaigle: This should also update the Database value, I think.

php/libraries/Utility.class.inc
```diff
@@ -306,7 +307,8 @@ class Utility
      */
     static function structureToPCRE(
         array $structure,
-        ?string $siteAbbrev = null
+        ?string $siteAbbrev = null,
```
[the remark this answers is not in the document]
johnsaigle: It also has the benefit of being backwards-compatible.

SQL/New_patches/2019-10-05-Add_alias_to_projects.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `Project` ADD COLUMN `Alias` char(4) NOT NULL;
```
PapillonMcGill:
```suggestion
ALTER TABLE `Project` ADD COLUMN `Alias` char(4) NOT NULL AFTER `Name`;
```
to keep ordering

php/libraries/Project.class.inc
ridz1208: i made it similar to the name and id...

johnsaigle: Fair enough, I didn't look at the rest of the class.

php/libraries/Utility.class.inc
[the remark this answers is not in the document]
ridz1208: @johnsaigle I'd like to keep both finally. I would like to be able to one day generate Ids with more than one prefix.

I know it's not the case right now but I think users should be able to generate PROJ-SITE-0000 IDs and that would not be possible with passing around a single `prefix` value.

ridz1208:
Documentation will be done in a different PR to avoid another rebase

PR #5286, opened 2019-10-07
[tools/install.sh] fix for other operating systems

### LINKED ISSUE #5262: [tools/install.sh] broken on macOS
Latest changes to the install.sh script on minor has broken the script on macOS.

Output after running the script:

> ./install.sh: line 156: hostnamectl: command not found

maltheism:
## Brief summary of changes

Allows the shell script to continue if the command "hostnamectl" is not found.
Note: Let me know if the else statement is unnecessary.

1. Test on macOS
2. Test on Ubuntu or centOS

* Resolves #5262

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -155,10 +155,15 @@ chmod 770 ../smarty/templates_c
 
 # Changing group to 'www-data' or 'apache' to give permission to create directories in Document Repository module
 # Detecting distribution
-os_distro=$(hostnamectl |awk -F: '/Operating System:/{print $2}'|cut -f2 -d ' ')
+if ! os_distro=$(hostnamectl |awk -F: '/Operating System:/{print $2}'|cut -f2 -d ' ')  2> /dev/null
+then
+  os_distro="other"
+else
+  os_distro=$(hostnamectl |awk -F: '/Operating System:/{print $2}'|cut -f2 -d ' ')
+fi
 
 debian=("Debian" "Ubuntu")
-redhat=("Red" "CentOS" "Fedora" "Oracle") 
+redhat=("Red" "CentOS" "Fedora" "Oracle")
 
 if [[ " ${debian[*]} " =~ " $os_distro " ]]; then
     mkdir -p ../modules/document_repository/user_uploads
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -155,7 +155,7 @@ chmod 770 ../smarty/templates_c
 
 # Changing group to 'www-data' or 'apache' to give permission to create directories in Document Repository module
 # Detecting distribution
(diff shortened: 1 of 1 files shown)


tools/install.sh
```diff
@@ -155,10 +155,15 @@ chmod 770 ../smarty/templates_c
 
 # Changing group to 'www-data' or 'apache' to give permission to create directories in Document Repository module
 # Detecting distribution
-os_distro=$(hostnamectl |awk -F: '/Operating System:/{print $2}'|cut -f2 -d ' ')
+if ! os_distro=$(hostnamectl |awk -F: '/Operating System:/{print $2}'|cut -f2 -d ' ')  2> /dev/null
```
johnsaigle: I think a better way to do this is:

```bash
command -v hostnamectl > /dev/null
if [[ "$?" -eq "0" ]]; then
     os_distro=$(hostnamectl |awk -F: '/Operating System:/{print $2}'|cut -f2 -d ' ')
else
     os_distro='other'
fi
```

`command -v hostnamectl` checks for the existence of the program `hostnamectl`.

If it exists, the return code of this statement will be `0`; otherwise, it will be `1`.

The variable `$?` captures the exit code of the last command run.

Then the if statement checks whether `hostnamectl` was found. If so, we are on some kind of Linux that we support and can parse the output of hostnamectl. Otherwise we default to other.

`command` is standard and exists on macOS so this should cover your issue.

johnsaigle:
Also I think this might be better going to `bugfix`.

tools/install.sh
maltheism: Hey @johnsaigle I tested your code just now and nothing gets outputted after that line for the rest of the installation script and contrary to my fix.

johnsaigle: That's really weird but I also confirmed that on macOS. `command -v ...` works fine on Ubuntu but fails silently in the same script on macOS. No idea why, couldn't find an answer via Google.

```suggestion
if ! os_distro=$(hostnamectl 2>/dev/null)
```

Given that my solution doesn't work, please change the line to this instead. It's a bit shorter and properly suppresses the error message (`2>/dev/null` should be inside the brackets).

christinerogers:
Hi @maltheism  Please also see the major install script cleanup for the upcoming 21 bugfix release - 
https://github.com/aces/Loris/pull/5205

johnsaigle:
@christinerogers They differ a bit since the one you've linked is a documentation change whereas this one is a functional change/bugfix.

johnsaigle:
Adding needs work because this should be sent to bugfix instead of minor.

maltheism:
@johnsaigle your recent change does work!

johnsaigle:
@maltheism Can you close this since you opened a new PR going to bugfix?

maltheism:
Closing for bugfix PR.

PR #5287, opened 2019-10-07
[Exceptions] Added New BadRequest and Forbidden Exceptions

HenriRabalais:
## Summary
Added new BadRequest and Forbidden Exceptions for various uses when validating requests.

I am currently using these in the Biobank module and need them to be on major so that the module can be based off of major

## BadRequest
Can be thrown when the user submitted data is invalid or does not meet certain requirements.

## Forbidden 
Can be used when an action shouldn't be taken by the user based on their permissions.

### DIFF
--- a/php/exceptions/BadRequest.class.inc
+++ b/php/exceptions/BadRequest.class.inc
@@ -1,0 +1,28 @@
+<?php
+/**
+ * This file contains the BadRequest exception type.
+ *
+ * PHP Version 7
+ *
+ * @category Errors
+ * @package  Loris
+ * @author   Loris Team <loris-dev@bic.mni.mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://github.com/aces/Loris
+ */
+
+/**
+ * The NotFound exception is thrown when a resource can not
+ * be found. It maps to a 400 error.
+ *
+ * @category Errors
+ * @package  Loris
+ * @author   Loris Team <loris-dev@bic.mni.mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://github.com/aces/Loris
+ */
+
+class BadRequest extends LorisException
+{
+}
+
--- a/php/exceptions/BadRequest.class.inc
+++ b/php/exceptions/BadRequest.class.inc
@@ -13,7 +13,7 @@
 
 /**
  * The BadRequest exception is thrown when the user's input is invalid or insufficient.
- * be found. It maps to a 400 error.
+ * It maps to a 400 error.
  *
  * @category Errors
  * @package  Loris
--- a/php/exceptions/BadRequest.class.inc
+++ b/php/exceptions/BadRequest.class.inc
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
 /**
  * This file contains the BadRequest exception type.
  *
--- a/php/exceptions/BadRequest.class.inc
+++ b/php/exceptions/BadRequest.class.inc
@@ -12,7 +12,7 @@
  */
 
 /**
- * The NotFound exception is thrown when a resource can not
+ * The BadRequest exception is thrown when the user's input is invalid or insufficient.
  * be found. It maps to a 400 error.
  *
  * @category Errors
--- a/php/exceptions/BadRequest.class.inc
+++ b/php/exceptions/BadRequest.class.inc
@@ -12,7 +12,8 @@
  */
 
 /**
- * The BadRequest exception is thrown when the user's input is invalid or insufficient.
+ * The BadRequest exception is thrown when the user's input is invalid or
+ * insufficient.
  * It maps to a 400 error.
  *
  * @category Errors
--- a/php/exceptions/BadRequest.class.inc
+++ b/php/exceptions/BadRequest.class.inc
@@ -12,9 +12,10 @@
  */
 
 /**
- * The BadRequest exception is thrown when the user's input is invalid or
- * insufficient.
- * It maps to a 400 error.
+ * The BadRequest exception is thrown when the server cannot or will not
+ * process the request due to something that is perceived to be a client error
+ * (e.g., malformed request syntax, invalid request message framing, or
+ * deceptive request routing). It maps to a 400 error.
  *
  * @category Errors
  * @package  Loris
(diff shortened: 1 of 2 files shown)


php/exceptions/Forbidden.class.inc
```diff
@@ -0,0 +1,28 @@
...
+ * @category Errors
+ * @package  Loris
+ * @author   Loris Team <loris-dev@bic.mni.mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://github.com/aces/Loris
+ */
+
+/**
+ * The NotFound exception is thrown when a resource can not
+ * be found. It maps to a 404 error.
```
johnsaigle:
```suggestion
 * be loaded due to insufficient permissions. It maps to a 403 error.
```

php/exceptions/Forbidden.class.inc
```diff
@@ -0,0 +1,28 @@
+<?php
+/**
+ * This file contains the NotFound exception type.
```
johnsaigle:
```suggestion
 * This file contains the Forbidden exception type.
```

php/exceptions/Forbidden.class.inc
```diff
@@ -0,0 +1,28 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

php/exceptions/Forbidden.class.inc
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @category Errors
+ * @package  Loris
+ * @author   Loris Team <loris-dev@bic.mni.mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://github.com/aces/Loris
+ */
+
+/**
+ * The NotFound exception is thrown when a resource can not
```
johnsaigle:
```suggestion
 * The Forbidden exception is thrown when a resource can not
```

php/exceptions/BadRequest.class.inc
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @category Errors
+ * @package  Loris
+ * @author   Loris Team <loris-dev@bic.mni.mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://github.com/aces/Loris
+ */
+
+/**
+ * The NotFound exception is thrown when a resource can not
```
johnsaigle:
```suggestion
 * The BadRequest exception is thrown when the user's input is invalid or insufficient.
```

php/exceptions/BadRequest.class.inc
```diff
@@ -0,0 +1,28 @@
...
+ * be found. It maps to a 400 error.
```
johnsaigle:
```suggestion
 * It maps to a 400 error.
```

php/exceptions/BadRequest.class.inc
```diff
@@ -0,0 +1,28 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

HenriRabalais:
@johnsaigle thanks! fixed!

php/exceptions/BadRequest.class.inc
```diff
@@ -0,0 +1,29 @@
...
+ * @category Errors
+ * @package  Loris
+ * @author   Loris Team <loris-dev@bic.mni.mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://github.com/aces/Loris
+ */
+
+/**
+ * The BadRequest exception is thrown when the user's input is invalid or
+ * insufficient.
```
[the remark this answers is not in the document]
HenriRabalais: I think I meant that the data was missing information that was required. Permission issues would trigger a `Forbidden` exception. Bad Request is to be used exclusively for data submission.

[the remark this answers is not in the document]
PapillonMcGill: Could you add some text to complement insufficient?

HenriRabalais:
@PapillonMcGill please take another look!

xlecours:
> It would make more sense to use middleware to map known exceptions to their HTTP counterpart and include typed exceptions that are meaningful in and of themselves. 

This seems to be in sync with what Zend propose in https://docs.zendframework.com/zend-httphandlerrunner/runner/#server-request-error-response-generator

xlecours:
@HenriRabalais I which context are you using those exceptions?

HenriRabalais:
@xlecours I am using a BadRequest exception when the user-provide data does not meet the required validation and I am using a Forbidden exception when the user does not having permission for the given request.

xlecours:
> @xlecours I am using a BadRequest exception when the user-provide data does not meet the required validation and I am using a Forbidden exception when the user does not having permission for the given request.

@HenriRabalais I am asking because if you are using RequestHandlerInterface implementations then those two condition could be check early in the handle function; then return a ResponseInterface implementation. 

If you really can't return a ResponseInterface then you can always add the exception classes under your module namespace.

driusan:
Would a `\DomainException` be applicable for the case you wanted a `BadRequest` exception for?

HenriRabalais:
@xlecours I am using the RequestHandlerInterface implementations but the validation checks are nested quite deeply within functions and it seems unruly to sent them back up the function chain. Maybe I haven't organized this properly, but within the handle function I am trying to resolve the request and then catch any exceptions that are thrown and at that point returning a ResponseInterface depending on the type of exception thrown.

php/exceptions/BadRequest.class.inc
```diff
@@ -0,0 +1,30 @@
...
+ * deceptive request routing). It maps to a 400 error.
+ *
+ * @category Errors
+ * @package  Loris
+ * @author   Loris Team <loris-dev@bic.mni.mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://github.com/aces/Loris
+ */
+
+class BadRequest extends LorisException
```
driusan: So are you going to remove this class from the PR in favour of the standard PHP exception?

HenriRabalais: Yes

HenriRabalais:
Will re-open when addressed, or issue a new PR with desired changes.

PR #5288, opened 2019-10-07
[HTTP] Added new JSON response for OK (200) server response

HenriRabalais:
## Summary
I added a new JSON response for when the server returns a 200. This is useful for any successful server request and can be accompanied by data in the form of an array which is returned to the client.


johnsaigle:
I think this is a duplicate of https://github.com/aces/Loris/pull/4949

HenriRabalais:
@johnsaigle Ah, perfect!

PR #5290, opened 2019-10-07
[Database] Update database class

HenriRabalais:
## Summary
1. Created a check to make sure that both the `ignore` and `onDuplicateInsert` flags are not used simultaneously in the `_realInsert` function.
2. Removed the unsetting of the unique key when using the pselectWithIndexKey function. I believe there are a lot of cases where still having the index key within the array and be able to access it's value via a key is incredibly useful.

### DIFF
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -803,7 +803,6 @@ class Database
 
             // If you get here, we just need to build the new array
             $filteredResult[$row[$uniqueKey]] = $row;
-            //unset($filteredResult[$row[$uniqueKey]][$uniqueKey]);
         }
 
         return $filteredResult;
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -383,7 +383,13 @@ class Database
         bool   $autoescape = true,
         bool   $ignore = false,
         bool   $onDuplicateUpdate = false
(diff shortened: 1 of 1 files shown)


php/libraries/Database.class.inc
```diff
@@ -797,7 +803,6 @@ class Database
 
             // If you get here, we just need to build the new array
             $filteredResult[$row[$uniqueKey]] = $row;
-            unset($filteredResult[$row[$uniqueKey]][$uniqueKey]);
```
maltheism: What was the reason the unset was being used here in the first place? I'm just wondering if its done for any particular reason.

HenriRabalais: @ridz1208?

ridz1208: removing this line is fine !

basically when I added this function, the result of the query will be indexed by the key but the key would not be in the result itself.

removing it would still be indexed by key but the key would also still be in the array result

PapillonMcGill: Would this cause any problem for existing use of the `pselectWithIndexKey` function? Like having a foreach assuming that the key is NOT included.

HenriRabalais: I went through each use of the function in the code, and to the best of my knowledge it will not cause any problems.

PR #5299, opened 2019-10-08
[Data Tables] Improve efficiency of DataTable filtering for large data tables

driusan:
This improves the efficiency of the DataTable JSX component when dealing with
large data tables. When counting the number of rows which match the filters
for the count in the header/footer, it returns an array of indexes which matched
the filter rather than returning a count. This prevents the need to needlessly
duplicate the work of tracking which rows match for the data table.

The indexed array is then sorted, and we work with that sorted list of indexes
instead of repeatedly going over this.props.data after the first time the matching
rows are calculated.

I suspect that somewhere in the deleted code there was an O(n^2) iteration over
the data that was avoided by re-writing and only directly working with the filtered
data after computing it, but I didn't track down where that was happening to it's
possible there was some other cause.

### DIFF
--- a/jsx/DataTable.js
+++ b/jsx/DataTable.js
@@ -28,8 +28,8 @@ class DataTable extends Component {
     this.updatePageNumber = this.updatePageNumber.bind(this);
     this.updatePageRows = this.updatePageRows.bind(this);
     this.downloadCSV = this.downloadCSV.bind(this);
-    this.countFilteredRows = this.countFilteredRows.bind(this);
-    this.getSortedRows = this.getSortedRows.bind(this);
+    this.getFilteredRows = this.getFilteredRows.bind(this);
+    this.sortRows = this.sortRows.bind(this);
     this.hasFilterKeyword = this.hasFilterKeyword.bind(this);
     this.renderActions = this.renderActions.bind(this);
   }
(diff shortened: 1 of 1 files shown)


johnsaigle:
Maybe this should go to bugfix instead of major?

driusan:
1. It's a core jsx file, changing it should require testing every data table in LORIS. I've only tested the candidate list with a large amount of data (and even then, only minimally)
2. It's not a bug fix, it's a performance improvement.

ridz1208:
@driusan 

1. bufixes are bugfixes... regardless of their size of impact. if we release something that doesn't work as advertised... we need to fix it.
2. Not sure the projects will see it like that when trying to navigate the conflict resolver and it takes 45 seconds at each keystroke.

Let's discuss this before we make a decision

jsx/DataTable.js
```diff
@@ -380,100 +390,63 @@ class DataTable extends Component {
...
-        // Get custom cell formatting if available
-        if (this.props.getFormattedCell) {
-          if (this.props.fields[j].show === false) {
-            data = null;
-          } else {
-            // create mapping between rowHeaders and rowData in a row Object
-            const row = {};
-            this.props.fields.forEach((field, k) => {
-              row[field.label] = this.props.data[index[i].RowIdx][k];
-            });
```
HenriRabalais: I think the removal of this may be causing some problems. This section of the code removes data when it is not meant to be shown and creates a row indexed by column header, which is what `getFormattedCell` needs to format the columns.

PR #5300, opened 2019-10-08
[Script] Add script to generate fake candidates

driusan:
This adds a script used to generate a large number of candidates
for testing #5299. The candidates are not realistic or even necessarily
valid, but only intended to see how the candidate_list data table
holds up with a large amount of data  (I tested with up to 20,000 candidates.)

This needs a lot of work to be made robust enough to be useable
for general purpose testing and is unlikely to be maintained, so
is added directly to the "deprecated" directory, but I'm sending it
to LORIS at @maltheism's request.

### DIFF
--- a/tools/deprecated/create_candidates.php
+++ b/tools/deprecated/create_candidates.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * This script inserts 5000 candidates. The CandIDs are consecutive instead
+ * of random, the PSCID is dccXXXX where XXXX is an incrementing integer, and
+ * the rest of the data is hardcoded.
+ *
+ * This is intended to stress test the candidate_list module with large numbers
+ * of candidates and is not intended to produce realistic data.
+ *
+ * PHP Version 7
+ *
+ * @category Main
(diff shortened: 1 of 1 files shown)


maltheism:
Hi @driusan, I tried running the script by typing "php tools/deprecated/create_candidates.php" and I'm getting the below output.
`PHP Warning:  require_once(/Users/alizee/Development/GitHub/McGill/Loris/tools/deprecated/../vendor/autoload.php): failed to open stream: No such file or directory in /Users/alizee/Development/GitHub/McGill/Loris/tools/deprecated/create_candidates.php on line 23
`
How do I fix that?

driusan:
Move it out of deprecated.

maltheism:
I tested it and it works. The script generated 104999 candidates for me. @driusan why have it in the deprecated directory?

driusan:
> This needs a lot of work to be made robust enough to be useable
for general purpose testing and is unlikely to be maintained, so
is added directly to the "deprecated" directory, but I'm sending it
to LORIS at @maltheism's request.

PR #5301, opened 2019-10-09
Move Response objects to explicit JSON namespace

driusan:
Move ambiguously named HTTP response classes which only handle JSON responses to an explicit JSON sub-directory/namespace.

This should avoid confusion by making the fact that they're only useful for JSON more explicit.


johnsaigle:
Any opposition to someone creating regular, concise responses in the future? I really liked the cleanliness of these classes for returning responses.

PR #5304, opened 2019-10-09
PUSH OF v21.0.2 INTO MAJOR

ridz1208:
DO NO SQUASH

conflicts:
	both modified:   .phan/config.php
	both modified:   composer.lock
	both modified:   modules/candidate_parameters/ajax/formHandler.php
	both modified:   modules/candidate_parameters/ajax/getData.php
	both modified:   modules/document_repository/php/files.class.inc
	both modified:   modules/timepoint_list/php/timepoint_list.class.inc
	both modified:   package-lock.json
	both modified:   php/libraries/File_Upload.class.inc
	both modified:   php/libraries/User.class.inc
	both modified:   php/libraries/Utility.class.inc
	both modified:   raisinbread/RB_files/RB_session.sql
	both modified:   raisinbread/migration.md
	both modified:   src/Api/Endpoints/Projects.php


johnsaigle:
Hmmm this is having a lot of issues with `phan` because of PhanTypeArraySuspiciousNullable problems. 

My thinking is that several of the problems were fixed on `major` (because #4308 went into `major` and that one addresses PhanTypeArraySuspiciousNullable) but new ones have now been introduced by new code going into `minor` and `bugfix`...

Once you've fixed the conflicts I can help with the `phan` part if you want.

In the future I'll try to

php/libraries/User.class.inc
```diff
@@ -149,22 +154,17 @@ class User extends UserPermissions
...
      *
      * @param array $set The array formatted for use in a Database call
      *
-     * @return bool True
+     * @return void
      * @access public
      * @static
      */
-    public function insert(array $set): bool
+    public static function insert(array $set): void
```
kongtiaowang: bool or void?

johnsaigle:
I'm readding the phan rule `PhanTypeArraySuspiciousNullable` for this PR. It's causing a ton of errors and isn't worth holding up the release schedule.

The majority of them come from the user of `pselect()` because it has a mixed return type. When code tries to access array values from its results, phan complains because `pselect` could in theory return null.

In another PR soon we can add an array return type to `pselect` and re-add the `PhanTypeArraySuspiciousNullable` rule in one go. This will be much easier to manage and review than doing so here.

PR #5306, opened 2019-10-09
[imaging uploader] Automatically populate PSCID, CandID & visit label based on the filename upon upload

### LINKED ISSUE #2909: Imaging Uploader: Why so much manual naming?
There seems to be three manual naming schemes that need to be all filled in properly before the Imaging Uploader will work properly.

1. The zip/tar.gz filename
2. The web form
3. The PatientName field in the DICOMs

It occurs to me that one of these could be used to properly populate all the fields, reducing the amount of manual work.

Personally, I'd prefer the web form (and DICOM data) being populated from the name of the zip/tar.gz file.

cmadjar:
## Brief summary of changes

Modified the behaviour of the upload tab of the imaging uploader module. PSCID, CandID and Visit Label fields are now automatically populated based on the name of the selected file to be uploaded.

1. Follow the test plan of the module

* [x] Resolves #2909

#### Caveats for existing projects

Note that there has been a change of behaviour for the upload tab of the imaging uploader module. The PSCID, CandID and Visit Label fields are now populated automatically based on the name of the file to be uploaded when a scan is not a phantom scan.

### DIFF
--- a/modules/imaging_uploader/jsx/UploadForm.js
+++ b/modules/imaging_uploader/jsx/UploadForm.js
@@ -27,7 +27,6 @@ class UploadForm extends Component {
     };
 
     this.onFormChange = this.onFormChange.bind(this);
-    this.getDisabledStatus = this.getDisabledStatus.bind(this);
     this.submitForm = this.submitForm.bind(this);
     this.uploadFile = this.uploadFile.bind(this);
   }
@@ -71,17 +70,6 @@ class UploadForm extends Component {
       form: form,
       formData: formData,
     });
-  }
-
-  /*
-   Returns false if Phantom Scans is set to No, and true otherwise
-   Result disables the element that calls the function
-   */
-  getDisabledStatus(phantomScans) {
-    if (phantomScans === 'N') {
-      return false;
-    }
-    return true;
   }
 
   submitForm() {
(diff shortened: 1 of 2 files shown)


ridz1208:
@cmadjar I always thought these serve as "double data entry" if we are enabling this for imaging uploader, I kow a few user that would kill for it to be added to media as well !!

cmadjar:
@ridz1208 Actually, this served as a triple data entry. The patient name had to be correct in the filename, in the header of the DICOM files and in the text fields for CandID, PSCID, Visit label of the upload form. This removes one layer of verification. So, the double data entry is done by comparing the filename with the patient name header in the DICOM file (performed by the imaging insertion script).

I would not do that for media as I doubt the information is present in the file itself too.

Does that make sense?

maltheism:
Hi @cmadjar, I just manual tested the PR and the CandID, PSCID, Visit label don't seem to be populated in the table after uploading the file when Phantom Scan is set to No. Otherwise the CandID, PSCID, Visit Label are populated in the table when the Phantom Scan is set to Yes. Although I still have to type the CandID, PSCID, with selecting the Visit Label on the frontend when the Phantom Scan is set to Yes.

cmadjar:
@maltheism weird. Did you run `make dev` after checking out the branch?

maltheism:
I apologize @cmadjar because it's now working. I just needed to clear my browser cache.

Issue #4172, opened 2018-12-03
Create developer install TestPlan.md document


johnsaigle:
We should create a document defining steps to take to verify the install of a  _developer_ instance of LORIS. This document would be followed by a developer during our release testing to determine that it is easy and straight-forward to install a dev instance. 

This could include steps to take during post-install, such as 
* "Bootstrapping": installing additional dependencies for the dev tools
* Getting a working copy of test data (i.e. the "Raisinbread" data set)
* ...

---

Some other things: 
* This document would need to be thoroughly vetted each release to be up-to-date with our dependencies, etc. This would help us toward our goal of reducing the need for ad-hoc Wiki hotfixes, which are a necessary evil while there are gaps in our documentation.

* The document should be as general as possible and should not contain information pertaining to our internal IT procedures.

* There should also be a companion document for Vagrant installs, if specific information is needed for that context.

* There is existing information available to help with this task: the "LORIS 101" document, existing Wiki pages, as well as institutional knowledge not written down but known by members of the team. There are likely some common bugs as well that can be fixed as we codify this process.

johnsaigle:
#4168 is relevant to this ticket. `phan` has a dependency on the PHP `ast` extension. This is required only for developer instances of LORIS and so is not installed by default. It should be done automatically on dev installs or at the very least explicitly documented for developer installs.

johnsaigle:
https://github.com/aces/Loris/issues/3313 is related. Need to make sure Vagrant install works effortlessly.

johnsaigle:
@ridz1208 Following up on the discussion in #4168 would you be able to do a small talk to the group about the status of the Wiki? What branch it's on, tentative schedule, how to contribute, etc?

johnsaigle:
#4365 is an attempt at automating some of the post-install steps and satisfying (Ubuntu) system dependencies as LORIS moves forward.

johnsaigle:
@kchatpar and I discovered more issues for Ubuntu-based developer installs today.

* Fresh Ubuntu lack the following tools by default:
    * make
    * npm

* The version of `node` shipped with Debian (and therefore Ubuntu) is very out of date to the point where it does not work with webpack. To remedy this we updated node using the following commands
```
curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
sudo apt-get install -y nodejs
```

These steps will likely be important for the bootstrap script, the Vagrant file, as well as the Wiki.

johnsaigle:
[...]

PR #5310, opened 2019-10-09
Very simple script to install system dependencies for development instances of LORIS

johnsaigle:
## Brief summary of changes

Getting a developer instance of LORIS working is still harder than it should be.

This is my latest attempt at trying to automate the process a bit. This is a simplified, less robust `bash` rewrite of #4365.

1. On an Ubuntu system, run `make devdependencies`
2. Try running this the next time you need to install LORIS and see whether it helps you.

* Another attempt at the same thing #4365
* Fixes some issues in #4172

### DIFF
--- a/tools/install-dev-dependencies.sh
+++ b/tools/install-dev-dependencies.sh
@@ -33,17 +33,17 @@ do
    fi
 done
 
-# Last updated for LORIS version: 21.0.2
+# Last updated for LORIS version: 22.0.0
 export PHP='7.2'
 declare -a dependencies=(
+"make"
 "curl"
+"git"
 "wget"
 "zip"
 "unzip"
 "php-json"
-"make"
 "software-properties-common"
-"php-ast"
 "php$PHP"
 "php$PHP-mysql"
 "php$PHP-xml"
@@ -60,10 +60,13 @@ declare -a dependencies=(
 echo "Updating apt package directory..."
 sudo apt update > /dev/null
 
-## Install all dependencies using sudo
+# Install all dependencies using sudo
 for i in "${dependencies[@]}"
 do
    echo "Installing dependency '$i' using apt...."
    sudo apt install -f -y "$i"
 done
 
+# Install php-ast using pecl as the Ubuntu version is too old.
+# This is required for phan to operate.
+sudo pecl install -f ast-1.0.3
--- a/tools/install-dev-dependencies.sh
+++ b/tools/install-dev-dependencies.sh
@@ -43,7 +43,6 @@ declare -a dependencies=(
 "zip"
 "unzip"
 "software-properties-common"
-"php$PHP"
 "php$PHP-mysql"
 "php$PHP-xml"
 "php$PHP-json"
--- a/tools/install-dev-dependencies.sh
+++ b/tools/install-dev-dependencies.sh
@@ -42,7 +42,6 @@ declare -a dependencies=(
 "wget"
 "zip"
 "unzip"
-"php-json"
 "software-properties-common"
 "php$PHP"
 "php$PHP-mysql"
--- a/tools/install_dev_dependencies.sh
+++ b/tools/install_dev_dependencies.sh
@@ -53,8 +53,10 @@ declare -a dependencies=(
 "libapache2-mod-php$PHP"
 )
 
-echo "Adding 'ppa:ondrej/php' (needed for newer versions of PHP)...."
-sudo add-apt-repository ppa:ondrej/php
+## XXX If you're unable to install the above PHP packages, uncomment the
+# following lines.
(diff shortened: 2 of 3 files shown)


tools/install_dev_dependencies.sh
```diff
@@ -0,0 +1,69 @@
...
+debian=("Debian" "Ubuntu")
+if [[ ! " ${debian[*]} " =~ " $os_distro " ]]; then
+    echo "Only Debian and Ubuntu are supported by $0."
+    exit 0;
+fi
+
+# apache2 is not included here because a developer may use a different server
+# config, such as the builtin PHP web server.
+declare -a prereqs=(
+"php"
```
PapillonMcGill: The script should be able to install php. It should be installed by dependecies for some of the packages in the second part (php$PHP-...).

It should not be a case of failure.

johnsaigle: I think that's a bit tricky because the version of PHP used by the command line isn't necessarily the same as the one Apache2 is using. I think this part should be left to the person setting up the system, similar to how it is done now.

PapillonMcGill: Yes, and in the second part, even if you install php7.2-xyz package, Apache might use php7.0-xyz package. 
To know which version Apache is using, you need to get to call phpinfo() (from the web server) and parse the result.
Maybe we could add a page with this function accessible only from localhost. That would required more apache config which could cause problem in multi-site hosting.

johnsaigle: The point of the PR is just to install system dependencies that so far don't get automatically installed. It's not intended to configure apache but to check that the right packages are installed.

PapillonMcGill: So why fail when PHP is not installed?

johnsaigle: These are prerequisites for LORIS that we are not configuring for people, but just documenting:

> Apache 2.4 or higher
MySQL >= 5.7 (or MariaDB >= 10.3)
PHP 7.2 or higher
Composer
NodeJS 8.0 or higher

We also require these libraries, but they are not well-documented:

> "curl"
"wget"
"zip"
"unzip"
"php-json"
"make"
"software-properties-common"
"php-ast"
"php$PHP"
"php$PHP-mysql"
"php$PHP-xml"
"php$PHP-json"
"php$PHP-mbstring"
"php$PHP-gd"
"libapache2-mod-php$PHP"

I'm looking to install the second list. They are easy to install and can be done by a simple bash script.

If we get into trying to automatically configure the entire LAMP stack for developers than we are *much* better off working toward some kind of containerization.

I'm in favour of making the LORIS install as easy as possible but frankly this is my third complete rewrite of trying to make the developer install easier. It seems that there isn't the time or will within the team to review and test a more complex/robust developer set-up. That's fine, that's the reality of it, but the problem remains that many people are tripping over these developer dependencies.

I'm keeping this script simple because I want it to have a chance of getting merged and fixing this problem that I've been trying to remedy for a year and a half. 

If you'd like to work on something more robust please go ahead and I'll be the first to review it.

tools/install-dev-dependencies.sh
```diff
@@ -0,0 +1,72 @@
...
+# Last updated for LORIS version: 22.0.0
+export PHP='7.2'
+declare -a dependencies=(
+"make"
+"curl"
+"git"
+"wget"
+"zip"
+"unzip"
+"php-json"
```
PapillonMcGill: php-json and on line 50 php$php-json for 7.2 version.

tools/install-dev-dependencies.sh
```diff
@@ -0,0 +1,72 @@
...
+"software-properties-common"
+"php$PHP"
```
PapillonMcGill: should already be installed at this point as it is a requisite.

ridz1208:
@johnsaigle should be updated to 7.3 now i guess

ridz1208:
@johnsaigle TRAVIS does not like this one

tools/install-dev-dependencies.sh
```diff
@@ -0,0 +1,70 @@
...
+       echo "$i is required for development but cannot be installed by this script."
+       echo "Please install $i and run this script again"
+       exit 1
+   else
+       echo "Requirement $i appears to be satisfied"
+   fi
+done
+
+# Last updated for LORIS version: 22.0.0
+export PHP='7.2'
```
PapillonMcGill: next major will require PHP 7.3. Should we require this now as it is to be merge in `master`?

johnsaigle:
Going to close this because I think the better approach is the Configuration Check script. #5965  That way we can flag what's not installed but we don't have to worry about installing on behalf of a user which is proving to be complicated. 

May revisit this later.

PR #5317, opened 2019-10-15
[Core: Database] Add array return type to pselect

johnsaigle:
## Brief summary of changes

`Database->pselect()` now always returns an array, thus removing the need to check if its result is an array before proceeding.

This should be safe as it is just a wrapper for the `execute()` function which already returns an array.

I'm not sure if this has implications for overrides or instruments so for now I'm adding the Caveat label.

## Related

#5318, #4284


driusan:
I don't see how this would have any implications for overrides or instruments.. as you say, it's already always returning an array, and no one should be overriding the database class..

johnsaigle:
Fair enough!

Issue #5311, opened 2019-10-09
installdb.php breaks when providing wrong credentials


zaliqarosli:
installdb.php returns 500 when you provide wrong credentials. it should just throw error warning and allow you to complete the process until submission succeeds

johnsaigle:
On which branch?

zaliqarosli:
21.0-release

PR #5320, opened 2019-10-16
[Install] Fix Database connection edge-case that resulted in 500 error

johnsaigle:
## Brief summary of changes

There was a slight issue with the logic during the install process that swallowed a DB exeption in favour of returning false; however, the return value of this function was never checked so the code kept trying to function without a DB connection. This could give a 500 error in some cases, making the install look messy and hard to debug.

This PR simply checks the return status of the `connect` function and throws a DB exception if it is false.

## To test

* Try to install a fresh LORIS. Provide incorrect DB information and use these installation settings:

Screen Shot 2019-10-16 at 11 09 53

You will get a blank page (500 error).

* On this branch, do the same thing. You will get a front-end error message with MySQL debug information.

## Related

* #5311

### DIFF
--- a/php/installer/Installer.class.inc
+++ b/php/installer/Installer.class.inc
@@ -297,12 +297,16 @@ class Installer
     {
         $DB = Database::singleton();
         try {
-            $DB->connect(
+            $success = $DB->connect(
                 $values['dbname'],
                 $values['dbadminuser'],
                 $values['dbadminpassword'],
                 $values['dbhost']
             );
+
+            if (! $success) {
+                throw new \DatabaseException('Could not connect to database');
+            }
         } catch (\Exception $e) {
             if (!empty($DB->lastErr)) {
                 $this->_lastErr = $DB->lastErr;


johnsaigle:
@zaliqarosli Tagging you on this because you reported #5311. Is this the problem you were getting or was it something else?

zaliqarosli:
@johnsaigle yes, wrong credentials return blank page (500 error) but regardless of what installation options you select

php/installer/Installer.class.inc
```diff
@@ -297,12 +297,16 @@ class Installer
...
+            $success = $DB->connect(
                 $values['dbname'],
                 $values['dbadminuser'],
                 $values['dbadminpassword'],
                 $values['dbhost']
             );
+
+            if (! $success) {
+                throw new \DatabaseException('Could not connect to database');
+            }
```
zaliqarosli: looking at the code, it looks to me like we should be editing what's in the catch block below. looks to me like line 308 here is redundant to lines 314-315. maybe the solution here is to do something with `$this->_lastErr` e.g. displaying it on the front-end and/or throw the DatabaseException there

zaliqarosli: ahhh i see

zaliqarosli:
will test when setting up the QPN stg environment this week!

johnsaigle:
thanks!

Issue #5297, opened 2019-10-08
On The Subject of Sudo -- a brief essay discussing UNIX permissions for LORIS


johnsaigle:
There has been a lot of discussion across various pull requests on the subject of `sudo` in LORIS, specifically during the install process.

Since the conversation is getting a little lost within the PR threads and somewhat hidden as new commits get added, I thought it would be best to summarize what's going on here.

### Sudo and why to avoid it

* This command originally stood for "superuser do"; effectively, anything that is executed with sudo is the same as running it as `root`.
* This gives a user with sudo privileges, _or a process running with these privileges_, **full control** to do whatever they want on a system
* For this reason it's considered bad to use `sudo`/`root` unless absolutely necessary because you can accidentally break things and subvert *nix systems' warnings not to take risky action.
    * It also violates the [security principle of least-privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege) 
        * > "every module (such as a process, a user, or a program, depending on the subject) must be able to access only the information and resources that are necessary for its legitimate purpose"

### LORIS and Sudo

Currently `sudo` is used somewhat in order to install LORIS. Usage falls into one of a few categories:

1. Changing file permissions (e.g. `chmod` to change smarty folders to be writeable/readable by the Apache user)
2. Configuring Apache (e.g. `a2enmod` to enable an Apache module)
3. Installing a package (e.g. `sudo apt install make`)

#### Current approach

Right now our documentation will tell you to do the following: 

- As `root` create `lorisadmin` then give `lorisadmin` sudo privileges
- Run install.sh as `lorisadmin`
- As `lorisadmin`, use `sudo` to do any task of 1-3 above
- Also, the script will fail when run as root

The last step was added recently because running as  `root` would create files that only `root` could access, breaking them from being read by `lorisadmin` or Apache.
On CentOS, even running the install script with sudo will create subdir with `root` ownership, thus only accessible by `root`

#### My proposed approach

I believe our requirements are the following:
* No files should be owned by root
* Root is needed during the install only
* All files should be readable/writable by `lorisadmin`
* Some files should be selectively readable/writable by the Apache user
    * Currently: `user_uploads/` under document_repository and data_release; `project/`; `logs/` (unused); and `smarty/templates_c`.

### So here's what we need to do:

- [ ] Do not give `lorisadmin` sudo.

[...]

PR #5323, opened 2019-10-16
[Core/Install] Improve UNIX permission usage

### LINKED ISSUE #5057: Change to UNIX file permissions on LORIS install
I'm exploring creating a Debian package for LORIS and its error-checking tool reports that we are using "non-standard permissions" (according to their standard at least). 

These include permissions such as `775` and `664`. Typically Linux permissions will be `755` and `644`.

From what I can tell, this is because we've proceeded with installation by asking users to create a `lorisadmin` user and then our install script changes the group to `www-data`/`apache2`. This effectively allows the web-server user to do everything that the owner can do. (This is indicated by the fact that the first two permission digits match.)

A better approach to file permissions would be to have `lorisadmin` as both the owner and the group and add the web user (e.g. `apache2`) to the `lorisadmin` group. 

This would allow us to use standard `755` and `644` permissions and also eliminate the need to change file permissions on certain directories during the install script.

See also: https://github.com/johnsaigle/loris-debian-package/issues/1

johnsaigle:
## Brief summary of changes

This PR improves the way UNIX permissions are used in LORIS.

#### Old permissions

* `lorisadmin` owns all files
* `lorisadmin` has sudo
* `www-data`/`apache2` selectively owns and is the group for web-writable files
    * e.g. `drwxrwxrwx 2 www-data    www-data 4096 Oct  9 16:30 user_uploads`
*  Uses the same permissions as `lorisadmin`
* Typical permission codes: `775`, for directories, `664` for files 
* **Result**: On certain files and folders, Apache gets to do everything `lorisadmin` can do

#### New permissions
* `lorisadmin` owns all files
* `lorisadmin` has sudo
* `lorisadmin` is the group for all files
    * e.g. `drwxrwxrwx 2 lorisadmin    lorisadmin 4096 Oct  9 16:30 user_uploads`
* `www-data`/`apache2` is part of the `lorisadmin` group
* Typical permission codes: `755` for directories, `644` for files.
* **Result**: Certain group permissions on files can be used to limit `www-data`/`apache2` access.

### Why?

More details in #5297. In short:

1. **Simplifies install** - no need to selectively `chgrp` to apache2 or www-data user
2. **Prerequisite for creating a Debian package** - the `lintian` tool causes LORIS to fail because of odd permission codes like `775` and `664`. See #5057 
3. **Easier maintenance** - everything is now `lorisadmin.lorisadmin`. Permissions can be tracked using GitHub rather than changed manually.
4. **This is how UNIX permissions are usually used**.
5. Brings us closer to the principle of least-privilege.

### Testing instructions

1. Test the install according to the new instructions. Ensure that LORIS still works.

### Links to related tickets (GitHub, Redmine, ...)

* Resolves #5057 
* Related #5297

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,6 @@ Consult the [LORIS Wiki](https://github.com/aces/Loris/wiki/Setup) page on this 
 
 ```bash
 # Create lorisadmin user and group
-sudo groupadd lorisadmin
 # Give lorisadmin `sudo` permission. This is required for the install process
 # in order to automatically generate Apache configuration files.
 # Sudo privileges can be revoked once the install is completed.
--- a/README.md
+++ b/README.md
@@ -51,8 +51,7 @@ Consult the [LORIS Wiki](https://github.com/aces/Loris/wiki/Setup) page on this 
 # in order to automatically generate Apache configuration files.
 # Sudo privileges can be revoked once the install is completed.
     sudo useradd -U -m -G sudo -s /bin/bash lorisadmin
-# Add apache and lorisadmin to the lorisadmin group
-    sudo groupadd -a -G lorisadmin lorisadmin
+# Add apache to the lorisadmin group
     sudo groupadd -a -G lorisadmin www-data
 # Set the password for the lorisadmin account
     sudo passwd lorisadmin
--- a/README.md
+++ b/README.md
@@ -46,18 +46,18 @@ Consult the [LORIS Wiki](https://github.com/aces/Loris/wiki/Setup) page on this 
 
 ```bash
 # Create lorisadmin user and group
-    sudo groupadd lorisadmin
+sudo groupadd lorisadmin
 # Give lorisadmin `sudo` permission. This is required for the install process
 # in order to automatically generate Apache configuration files.
 # Sudo privileges can be revoked once the install is completed.
-    sudo useradd -U -m -G sudo -s /bin/bash lorisadmin
+sudo useradd -U -m -G sudo -s /bin/bash lorisadmin
 # Add apache to the lorisadmin group
-    sudo groupadd -a -G lorisadmin www-data
+sudo groupadd -a -G lorisadmin www-data
 # Set the password for the lorisadmin account
-    sudo passwd lorisadmin
-    sudo mkdir -m 755 -p /var/www/$projectname
-    sudo chown lorisadmin.lorisadmin /var/www/$projectname
-    su - lorisadmin
+sudo passwd lorisadmin
+sudo mkdir -m 755 -p /var/www/$projectname
+sudo chown lorisadmin.lorisadmin /var/www/$projectname
+su - lorisadmin
 ```
 
  <i>$projectname ⇾ "loris" or one-word project name</i>
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Consult the [LORIS Wiki](https://github.com/aces/Loris/wiki/Setup) page on this 
 # Sudo privileges can be revoked once the install is completed.
 sudo useradd -U -m -G sudo -s /bin/bash lorisadmin
 # Add apache to the lorisadmin group
-sudo groupadd -a -G lorisadmin www-data
+sudo usermod -a -G lorisadmin www-data
 # Set the password for the lorisadmin account
 sudo passwd lorisadmin
 sudo mkdir -m 755 -p /var/www/$projectname
--- a/README.md
+++ b/README.md
@@ -42,20 +42,24 @@ These dependencies are subject to change so be sure to verify your version of My
 
 Consult the [LORIS Wiki](https://github.com/aces/Loris/wiki/Setup) page on this [Install process](https://github.com/aces/Loris/wiki/Installing-Loris) for more information.
 
-1. Set up LINUX user lorisadmin, with `sudo` privilege, and create LORIS base directory:
+1. Set up LINUX user and group lorisadmin and create LORIS base directory:
 
 ```bash
-sudo useradd -U -m -G sudo -s /bin/bash lorisadmin
-sudo passwd lorisadmin
-su - lorisadmin
+# Create lorisadmin user and group
+    sudo groupadd lorisadmin
+# Give lorisadmin `sudo` permission. This is required for the install process
+# in order to automatically generate Apache configuration files.
+# Sudo privileges can be revoked once the install is completed.
+    sudo useradd -U -m -G sudo -s /bin/bash lorisadmin
+# Add apache and lorisadmin to the lorisadmin group
+    sudo groupadd -a -G lorisadmin lorisadmin
+    sudo groupadd -a -G lorisadmin www-data
+# Set the password for the lorisadmin account
+    sudo passwd lorisadmin
+    sudo mkdir -m 755 -p /var/www/$projectname
+    sudo chown lorisadmin.lorisadmin /var/www/$projectname
+    su - lorisadmin
 ```
-
- <b>Important: All steps from this point forward must be executed by lorisadmin user</b>
-
- ```bash
- sudo mkdir -m 775 -p /var/www/$projectname
- sudo chown lorisadmin.lorisadmin /var/www/$projectname
- ```
 
  <i>$projectname ⇾ "loris" or one-word project name</i>
 
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -138,7 +138,6 @@ elif [[ " ${redhat[*]} " == *" $os_distro "* ]]; then
     sudo chown apache.apache ../smarty/templates_c
     # Make Apache the group for project directory, so that the web based install
     # can write the config.xml file.
-    sudo chgrp apache ../project
     sudo chmod 770 ../project
 else
     echo "$os_distro Linux distribution detected. We currently do not support this. "
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,9 @@ changes in the following format: PR #1234***
 
 #### Core
 - Menus are now maintained by modules and no longer in the SQL database (PR #5839)
-- Very old instrument relying on QuickForm may have issues due to code changes (PR #4928)
+- Very old instruments relying on QuickForm may have issues due to code changes (PR #4928)
+- Unix user permissions have been updated which may affect access to files. New
+documentation for file permissions has been added to the README.md file (PR #5323)
 
 #### Modules 
 
--- a/tools/create-project.sh
+++ b/tools/create-project.sh
@@ -3,6 +3,6 @@
 set -euo pipefail
 
 for d in data libraries instruments templates tables_sql modules; do
-    mkdir -p "${1}/${d}"
+    mkdir -p -m 750 "${1}/${d}"
 done
 
--- a/tools/create-project.sh
+++ b/tools/create-project.sh
@@ -3,6 +3,6 @@
 set -euo pipefail
 
 for d in data libraries instruments templates tables_sql modules; do
-    mkdir -p -m 750 "${1}/${d}"
+    mkdir -p "${1}/${d}"
 done
 


johnsaigle:
If this is merged, the installation readmes (CentOS, "Install in depth", macOS) should be updated. 

Since at this time there are many open PRs dealing with these same files, it will be easier to make these changes once those PRs go through.

tools/install.sh
```diff
@@ -161,41 +161,25 @@ debian=("Debian" "Ubuntu")
...
     # Make Apache the group for project directory, so that the web based install
     # can write the config.xml file.
-    sudo chgrp www-data ../project
-    sudo chmod 770 ../project
+    chmod 770 ../project
 elif [[ " ${redhat[*]} " =~ " $os_distro " ]]; then
-    mkdir -p ../modules/document_repository/user_uploads
-    mkdir -p ../modules/data_release/user_uploads
+    mkdir -p -m 770 ../modules/document_repository/user_uploads
+    mkdir -p -m 770 ../modules/data_release/user_uploads
```
driusan: This doesn't seem to do what the PR description says (not this line, the lines below it where things are still being chown'd to apache.apache)

gdevenyi:
I strongly suggest symbolic rather than numeric permissions to make things easier to read/interpret at a glance. https://en.wikipedia.org/wiki/Chmod#Symbolic_modes

driusan:
I disagree.. when talking about absolute permissions, numeric ones are far more common and, as a result, easier to read

tools/install.sh
johnsaigle: You're right. I'll remove them

PapillonMcGill:
- If linst file uploaded by a user get 644 permission, LORIS will not be able to edit (read only).
Is this desired?

- Files (linst or otherwise) created by LORIS (apache) would get www-data:www-data ownership and won't be readable by lorisadmin (without sudo).
This would be a security improvement but is it desired? (Do lorisadmin user need access to these files?)

johnsaigle:
> If linst file uploaded by a user get 644 permission, LORIS will not be able to edit (read only).
Is this desired?

The `6` permission at the beginning means that Apache will be able to read and write the file. As far as I know the execute permission doesn't need to be set for editing LINST files in the instrument builder.

> Files (linst or otherwise) created by LORIS (apache) would get www-data:www-data ownership and won't be readable by lorisadmin (without sudo).
This would be a security improvement but is it desired? (Do lorisadmin user need access to these files?)

I think it's alright. I can't really see a scenario where you'd need to do this and wouldn't have sudo. Maybe on some production servers? But in this case it makes sense to have an extra measure against deleting user's files. 

It's more of an administration question since files uploaded to LORIS shouldn't be under `/var/www/loris/` anyway, but rather some other folder like `/data/` that is outside the code base. Whoever is administering this on the back-end (dev or sysadmin) should have the powers to delete or read a file if they really need to anyway.

PapillonMcGill:
@johnsaigle The 6 is for the owner of the files (user=>lorisadmin), not for the group. Apache would have group permission (4) and not be able to modify the file.

As for the 2nd point, entirely approve. and yes, data should not be with the code. I am just asking if the behavior would cause problem to some process (like shell script).

johnsaigle:
I believe www-data will be the owner, as you said here:
> Files (linst or otherwise) created by LORIS (apache) would get www-data:www-data ownership and won't be readable by lorisadmin (without sudo).

README.md
```diff
@@ -34,20 +34,23 @@ These dependencies are subject to change so be sure to verify your version of My
...
+
+```bash
+# Create lorisadmin user and group
+sudo groupadd lorisadmin
+# Give lorisadmin `sudo` permission. This is required for the install process
+# in order to automatically generate Apache configuration files.
+# Sudo privileges can be revoked once the install is completed.
+sudo useradd -U -m -G sudo -s /bin/bash lorisadmin
+# Add apache to the lorisadmin group
+sudo groupadd -a -G lorisadmin www-data
```
PapillonMcGill:
```suggestion
sudo usermod -a -G lorisadmin www-data
```

README.md
```diff
@@ -34,20 +34,23 @@ These dependencies are subject to change so be sure to verify your version of My
...
-
-    ```
-    sudo mkdir -m 775 -p /var/www/$projectname
-    sudo chown lorisadmin.lorisadmin /var/www/$projectname
-    ```
+1. Set up LINUX user and group lorisadmin and create LORIS base directory:
+
+```bash
+# Create lorisadmin user and group
+sudo groupadd lorisadmin
```
PapillonMcGill: not needed as it will be created by the `useradd`

tools/install.sh
PapillonMcGill: Even if `group` would have `RWX` on the directories, files created by Apache would have ownership of www-data:www-data and Lorisadmin would not be able to modify/erase them and possibly read them.
If you want that, it would require to set

README.md
johnsaigle:
```suggestion
```

ridz1208:
@johnsaigle Before I comment on the code, i have a few questions

- If we use `lorisadmin` as a group including the apache user, wouldn't it get confusing (especially internally) for devs ? we already use lorisadmin right now just for the user, shouldn't we use a different name for the group? like webgroup or something that would encompass lorisadmin and apache ? I can just see people altering the permissions anyways thinking that it got misconfigured or wtv...
- when you say "permissions are going to be tracked on github" aren't they already tracked on github ?
- shouldn't lorisadmin not have sudo post-installation step ?

ridz1208:
@johnsaigle to make this easier
https://github.com/aces/Loris/pull/5815

PapillonMcGill:
```
Moreover, www-data should never be part of lorisadmin, which is already a sudo user, way too dangerous.
```
@lingma Do not confuse the user `lorisadmin` with the group `lorisadmin`.

johnsaigle:
> Moreover, www-data should never be part of lorisadmin, which is already a sudo user, way too dangerous.

As @PapillonMcGill mentioned above, the group called `lorisadmin` will not have the same permissions as the user `lorisadmin`. `www-data` will be limited to group permissions of `lorisadmin`. That's why changing e.g. `755` to `775` make a difference - it restricts the access that the web server has to those files.

lingma:
Even if the www-data is only lorisadmin group user, the apache2 www-data will have access to all lorisadmin user data, it can read all the history data of access, any user data but no one notices this since no one ever thinks that www-data is there to visit. I don't think the whole idea is acceptable. Sorry I did mix the idea of group and user, but the whole idea is too strange to be acceptable to me.

johnsaigle:
> the apache2 www-data will have access to all lorisadmin user data, it can read all the history data of access, any user data but no one notices this since no one ever thinks that www-data is there to visit

Can you explain this point a little more? I don't think it's true that members of UNIX groups are able to access the data of a user sharing the same name as a group.

I think a source of confusion may be the name. Imagine instead that we created a group called `lorisgroup` and added both `lorisadmin` and `www-data` to that group and kept the same permissions that are used in this pull request. Does thinking about it that way make more sense?

lingma:
The www-data user was there to make sure that apache could not access any system file. If www-data is part of a group which is also a system sudo user, which will leave a door open there for something to happen. For example, www-data could now read lorisadmin data, which, sometimes not all files will be written as 644, some file might be written a 664.
www-data is there to limit access, but now, with the patch, www-data will get more permission in fact, this is against the initial design of web server security.

johnsaigle:
> If www-data is part of a group which is also a system sudo user

The group `lorisadmin` will not have sudo privileges. Only the user called `lorisadmin` will have sudo privileges. Therefore www-data will never have sudo access.

> not all files will be written as 644, some file might be written a 664

As I mentioned in the PR description, we will be changing files to have 644/755 permissions so that www-data is further restricted.

lingma:
what I said in the last post is "How will you limit the files in the lorisadmin user home directory? And the files which could be shared with a normal registered user?" www-data is meant to be a non existing user.

lingma:
The principle  of security: to limit access privileges.
The modification: to give www-data more privileges via a super user's group, except "775 to 755" part as I initially mentioned.

That is not acceptable in my eyes. Yes, this could work if you pay attention to what lorisadmin does or you know what you have done. But this could be a trap to a normal user

lingma:
"we will be changing files to have 644/755 permissions"  This part is necessary as a patch, which should not be part of this improvement, which should be a mandatory corrections to the initial versions, if running under correct users.

johnsaigle:
What I'm understanding from what I'm saying is that you're worried about the case where someone is able to read files from the home directory of `lorisadmin`. 

This is a possibility given that this would add `www-data` to the `lorisadmin` group. If a bug exists that allows www-data to read arbitrary files on the server, then data could be extracted from the lorisadmin's home directory.

We have a couple approaches to dealing with this:
1. We can use a new group instead called e.g. `lorisapp` with `lorisadmin` and `www-data` as members. Since this is distinct from the default group `lorisadmin`, then `www-data` will not have access to files in `lorisadmin`'s home directory.
2. `lorisadmin` can change the permissions on its home directory to forbid read access from groups.

For the record, `lorisadmin` does not have any `sudo` access on production installs of LORIS hosted at the MNI. Only staging and development servers are configured this way. Also, lorisadmin shouldn't be storing any sensitive information in their home directories anyway. But in any case we may as well prevent the possibility using either of the approaches above.

lingma:
> 1. We can use a new group instead called e.g. `lorisapp` with `lorisadmin` and `www-data` as members. Since this is distinct from the default group `lorisadmin`, then `www-data` will not have access to files in `lorisadmin`'s home directory.

if `www-data` is in the same group as `lorisadmin`, then `www-data` could access `lorisadmin` group data. I did not read any part who stating that `lorisadmin` won't have `sudo` acess.
The idea behind the current security modification design is not general, I don't see why we should make this modification. This could be a per application setting if you know what you are doing.

johnsaigle:
> if www-data is in the same group as lorisadmin, then www-data could access lorisadmin group data.

They would both be able to access whatever the group can access. In my example with `lorisapp`, `www-data` wouldn't be able to access e.g. `lorisadmin`'s home directory, which was your concern if I understood correctly.

> I did not read any part of documentation which states that lorisadmin won't have sudo acess.

This isn't in the documentation on the repository but this is how servers are set-up at the MNI. We should probably add this to the documentation on the repository. 

> This could be a per application setting if you know what you are doing.

That's what I'm trying to move toward here. E.g. if you installed OwnCloud on the same server you could create a user and group called `owncloud` and add `www-data` to the `owncloud` group. This way both applications will be readable by Apache but `lorisadmin` has no access to `owncloud`.

lingma:
> OwnCloud

Then apache will have access to owncloud user information, which is not correct as I stated early, the will increase `www-data` to some authorized user's file. Of course, there is not any problem unless someone else knows and tries. Still, I don't see the point of doing this.

> They would both be able to access whatever the group can access. In my example with `lorisapp`, `www-data` wouldn't be able to access e.g. `lorisadmin`'s home directory, which was your concern if I understood correctly.

if Apache `www-data` is in for example `lorisapp`, then `lorisadmin` needs to be in the same group, so whatever `lorisadmin` could see will be visible to `www-data`, so the visibility of `www-data` is then increased.

johnsaigle:
> if Apache www-data is in for example lorisapp, then lorisadmin needs to be in the same group, whatever lorisadmin could see will be visible to www-data, so the visibility of www-data is then increased.

The visibility/access will be whatever files `lorisapp` can access. `www-data` does not get access to the files of `lorisadmin`.

lingma:
> access

If that is the case, what is the point of add a group `lorisapp`? Adding a group is to share something, otherwise it is just a new name, no advantage to add complexity, as a result there is no necessity to do anything

johnsaigle:
I laid out what I feel the advantages are in the original post as well as the linked issues.

lingma:
> The visibility/access will be whatever files `lorisapp` can access. `www-data` does not get access to the files of `lorisadmin`.

In fact, we are just turning in round here. `www-data` will not have access to the files of `lorisadmin` even if it is in the group named `lorisadmin`. The important here is not the name.

If `www-data` will have access to the files with the same group permissions as `lorisadmin`, not the `lorisadmin`'s own files. Then we are returning to the initial point of our discussion. `www-data` gets the increased visibility.

If `www-data` will NOT have access to the files with the same group permissions as `lorisadmin`, and no permission to `lorisadmin` 's file, otherwise we will return to the above case with group permission, then it is just a new name as `www-data` group or `lorisadmin` whatever. No point at all.

johnsaigle:
I'll zoom out for a second and explain the problems this is trying to fix.

My goal is to have normal LORIS development and production installs be able to operate without requiring `sudo` for normal usage. (Let's not include things like using `apt` or restarting system services for now as these are always privileged operations. This includes the first install of LORIS.)

Right now, our files and folders are set up sub-optimally.
For example, we have a folder `smarty/templates_c/` which contains compiled files which help format the front-end of LORIS. 

Right now, this is configured to have www-data as the owner and the group:
`drwxrwx--- 2 www-data   www-data   4096 Nov 27 17:18 templates_c`

www-data needs access to those files in order to render the page. However, sometimes we'll want to clean this folder. `lorisadmin` currently must run with `sudo` to do this because it does not own these files. This isn't ideal and requires that `lorisadmin` force changes to the files. 

Instead, I think that `lorisadmin` should be the owner of all the files in the project, which would give us the following:

`drwxrwx--- 2 lorisadmin   www-data   4096 Nov 27 17:18 templates_c`

This works and solves the above problem. However for this folder (and others like it), we need the install script to manually change the group of this folder to be owned by `www-data`. This has to be done differently for each operating system, e.g. CentOS instead uses `apache`.

My proposal is to use set every file to the same owner and group, `lorisadmin` (or `lorisapp`, whatever), bringing us to this:

`drwxrwx--- 2 lorisadmin   lorisadmin   4096 Nov 27 17:18 templates_c`

So long as the server user is a member of the group `lorisadmin` then everything is good. We haven't given the server any additional access to files, just changed the name (as you said). This approach though means that every file in the project has the same owner and group permissions and we don't have to do any manual configuration or run into hiccups where people with less experience with UNIX use sudo to solve the problem (or else get stuck).

My feeling is that this will simplify the development process.

johnsaigle:
>"we will be changing files to have 644/755 permissions" This part is necessary as a patch, which should not be part of this improvement, which should be a mandatory corrections to the initial versions, if running under correct users.

And yes I agree this should be done either way. However part of the problem is that things aren't running as the correct users, at least not according to the principle of least privilege, which makes people think they need sudo, which in turn makes things even worse.

lingma:
I understood the initial thought about simplifying the development to only one user, what concerns me is that this comes at the cost of security (in some level, which could be minimal to more serious). In fact, even if you using this method, the files in the `templates_c` will have owner as `www-data`, unless you set the default web server user as `lorisadmin`, in the later case, I will feel that we are creating some problem. As a result, I think the current solution won't solve any problem except that the github files will look better.

johnsaigle:
> the files in the templates_c will have owner as www-data

If a user has write permissions on a folder then that user can delete any files in that folder.

> cost of security

I addressed the example you gave about the web server user accessing `lorisadmin`'s home folder and offered two solutions to deal with your point. What specific security concerns do you have?

lingma:
It is true that we can delete a file in this way, but usually I don't like to answer yes for each of those files. I don't see much benefit from this solution, which is against the principle I listed early, as a result this is not what I like to know that group files are visible. A hacker could always exploit this kind of possibilities, of course we could put nothing there to prevent. If you concern about development experience, you can set web server user as `lorisadmin` or `www-data`, and all files as `lorisadmin` or `www-data`, in this case, there will be no trouble at all to delete a file or folder.

This solution could work of course in some way.

lingma:
One more problem that I could think of, the apache group name could change, so you will need user to know what name to add to the group lorisadmin, if not done correctly, then there will be a lot of questions concerning why the system doesn't work.

johnsaigle:
Blocked by #5815

johnsaigle:
Well @lingma I think we'll have to agree to disagree here. 

I am confident that this change simplifies the development experience and I feel that I've responded to your security concern to the best of my ability. I respect that you disagree and I'm not sure what i can do to change your mind beyond what I've already said.

I am happy to leave it to another reviewer to weigh in here and hopefully we'll reach a group consensus with some more input.

lingma:
Loris as a software product has nothing to do with group permission, this is unnecessary and creating possible problems. Just change all read only files to 644, folders to 755. This will not create any problem in development. I think it will be perfect to put all those web server writable folders some where and tell users what to do with them with a guide.

Of course your solution works. I tested without issue, but the idea itself is much out of my taste.

johnsaigle:
Adding Needs Documentation as a note that I should update the CHANGELOG

driusan:
wouldn't "Add To Release Notes" be a better reminder?

ridz1208:
I ain't touching this one, i think it's getting enough TLC

README.md
PapillonMcGill: lost in rebased?

HenriRabalais:
The amount of labels on this PR is impressive.
labels

PR #4469, opened 2019-04-19
[WIKI] Study Parameter - Identifiers

ridz1208:
### Brief summary of changes

How to add, configure the ID generation for the study

___

*This and all upcoming PRs for the WIKI will contain one or a couple files at most. this will hopefully prevent PRs from staying open for very long periods.*

*This and all upcoming PRs are open to discussions and suggestions. Most of the content has been copied over from the existing WIKI with sometimes minor, major or total changes.*

*Feel free to make suggestions and/or push directly to the branch !!*

PR #5328, opened 2019-10-16
[Core] Move PSCID generation config settings from config file to Configuration module

### LINKED ISSUE #5292: Move CandID and PSCID generation options from config.xml to the Configuration module

johnsaigle:
## Brief summary of changes

Moves settings related to generated PSCIDs and ExternalIDs to the Config module instead of the config.xml table.

This should simplify the code and make it easier for users to configure LORIS without the need for a developer.

#### Details

* Adds new patch for new config settings
* Updates SiteIDGenerator to grab config settings from DB instead of config.xml
* Refactors some of the Config module to be more modern/clean
* Adds function `Utility::toDropdownArray` to simplify the creation of arrays like:

```php
array(
'' => '',
'option1' => 'option1',
'option2' => 'option2',
...
);
```

1. Checkout the branch and run the new patch
2. Go to the configuration module and verify that you can change settings related to ID generation under the Study heading
3. Create new candidates using various config settings and ensure that the values make sense given your settings. E.g. IDs created with the `numeric` alphabet should only contain numbers.

* Resolves #5292


johnsaigle:
Still TODO:

- [ ] Add support for minValue and maxValue
- [ ] Config settings should be validated upon submission
- [ ] Remove support for `user` generation method?
    * This is broken on 21 due to a TypeError anyway
    * Will ask in LORIS meeting if anyone uses it.
- [ ] Delete old `_getIDSetting()` code in SiteIDGenerator.php

johnsaigle:
Wiki file added in #4469 will need to be updated when these two changes are on the same branch.

modules/configuration/php/configuration.class.inc
```diff
@@ -69,10 +69,47 @@ class Configuration extends \NDB_Form
...
+        $this->tpl_data['config']          = $this->_getConfigSettingTree();
+        $this->tpl_data['sandbox']         = \NDB_Factory::singleton()
+            ->config()
+            ->getSetting("sandbox");
+
+        // Build dropdown menus
+        $this->tpl_data['instruments']   = \Utility::toDropdownArray(
+            \Utility::getAllInstruments()
+        );
+        $this->tpl_data['scan_types']    = \Utility::toDropdownArray(
```
ridz1208: I dont think this is equivalent to what was there before. the previous array used the scan ID from the database as key, this will use sequential keys. you can use the pselectColWithIndexKey and modify your `toDropdownArray` to accept associative arrays.

php/libraries/SiteIDGenerator.php
```diff
@@ -44,54 +44,88 @@ class SiteIDGenerator extends IdentifierGenerator
...
         $this->siteAlias    = $siteAlias;
         $this->projectAlias = $projectAlias;
-        $this->prefix       = $this->_getIDSetting('prefix');
+        // Determine prefix to be used for the ID.
+        switch ($config->getSetting('idPrefixMethod')) {
+        case 'static':
+            $this->prefix = $config->getSetting('idStaticPrefix');
+            break;
+        case 'alias':
+            $this->prefix = $siteAbbrevPrefix;
```
ridz1208: i think this was not modified to account for projects

modules/configuration/php/configuration.class.inc
```diff
@@ -69,10 +69,47 @@ class Configuration extends \NDB_Form
...
+        $this->tpl_data['generationMethod'] = \Utility::toDropdownArray(
+            array(
+             'sequential',
+             'random',
+            )
+        );
+
+        $this->tpl_data['idPrefix'] = \Utility::toDropdownArray(
+            array(
+             'alias',
```
ridz1208: where do you define if you want your alias to be the project or the site alias ?

php/libraries/SiteIDGenerator.php
johnsaigle: Yeah I think this PR predates that change. It'll need to be updated

johnsaigle:
After discussing with Rida, I'm closing this PR for now. We should create a clear concept of what a PSCID can be and create an interface and class for this. At that time, we can decide whether it makes sense to configure the values from the config file or in the Configuration module. #5926

PR #5329, opened 2019-10-16
[new_profile] can't create a new profile on major

kongtiaowang:
It needs to return a candID as a string, not an object.

### DIFF
--- a/modules/new_profile/php/new_profile.class.inc
+++ b/modules/new_profile/php/new_profile.class.inc
@@ -91,7 +91,7 @@ class New_Profile extends \NDB_Form
         /* Update front-end data. Use passed PSCID when present, otherwise
          * retrieve it from the newly-created candidate.
          */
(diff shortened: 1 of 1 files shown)


modules/new_profile/php/new_profile.class.inc
```diff
@@ -91,7 +91,7 @@ class New_Profile extends \NDB_Form
         /* Update front-end data. Use passed PSCID when present, otherwise
          * retrieve it from the newly-created candidate.
          */
-        $this->_candID = $candID;
+        $this->_candID = \Candidate::singleton($candID)->getCandID();
```
johnsaigle:
```suggestion
        $this->_candID = (string) $candID;
```

kongtiaowang: PHP can't swith object as a string.

driusan:
@johnsaigle did you review this?

johnsaigle:
I'll do a deeper review on Monday.

Issue #5331, opened 2019-10-16
PSCID generation method "user" is broken (bugfix)


johnsaigle:
Right now it gives a TypeError in the latest release (#5332)

Fixing the TypeError produces another error:

<img width="1369" alt="Screen Shot 2019-10-16 at 17 53 48" src="https://user-images.githubusercontent.com/4022790/66962013-ef4d4280-f03d-11e9-8b19-d6734b6a3981.png">

But there is no PSCID field in the form to provide... Normally it seems that this PSCID value would be populated based on a value of a config setting but it does not get set anywhere.

It may have been removed in a Reactification PR; right now there doesn't seem to be any place where this field can appear.

We should fix this or remove this ID generation method.

## To reproduce

Change your PSCID generation method in config.xml to `user` (right now it's probably `sequential` or `random`).

Try to create a new user.

johnsaigle:
It's working now? I'm not sure what I was doing wrong before.

PR #5332, opened 2019-10-16
[New Profile] Fix crash when creating candidate with "user" generation method

johnsaigle:
## Brief summary of changes

A project with the `user` ID generation method is not able to create new candidates. A TypeError occurs because Site must take an int, not a string:

>PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Site::singleton() must be of the type int, string given, called in /var/www/loris/modules/new_profile/php/new_profile.class.inc on line 280

~~This PR fixes the type error.... but candidate creation does still not seem to be possible. See #5331. We should discuss whether this should be fixed or removed.~~

~~In any case, this PR at least stops LORIS from crashing so it's a small step forward.~~

1. See #5331.

### DIFF
--- a/modules/new_profile/php/new_profile.class.inc
+++ b/modules/new_profile/php/new_profile.class.inc
@@ -282,7 +282,7 @@ class New_Profile extends \NDB_Form
             } else {
                 // user has multiple sites,
                 // so validate PSCID against the Site selected
-                $site =& \Site::singleton($values['site']);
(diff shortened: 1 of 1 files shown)


xlecours:
@johnsaigle I thought the API was handling the candidate creation well. I sent this https://github.com/aces/Loris/pull/5336 to make the new_profile module use the api.

johnsaigle:
@xlecours Should your PR replace this one then?

PR #5336, opened 2019-10-17
[new_profile] Using api endpoint

xlecours:
This remove code duplication by sending the candidate creation request to the /api/v0.0.2/candidates endpoint. 

To test: 
Use the `new_profile` module to create candidates with different config options (PSCID generation: user/random/sequential, useEDC)

### DIFF
--- a/modules/new_profile/jsx/NewProfileIndex.js
+++ b/modules/new_profile/jsx/NewProfileIndex.js
@@ -74,43 +74,44 @@ class NewProfileIndex extends React.Component {
    */
   handleSubmit(e) {
     e.preventDefault();
-    const match = this.validateMatchDate();
-    if (!match) {
-      this.setState({
-        isCreated: false,
-      });
-    } else {
-      let formData = this.state.formData;
-      let formObject = new FormData();
-      for (let key in formData) {
-        if (formData[key] !== '') {
-          formObject.append(key, formData[key]);
-        }
-      }
-      formObject.append('fire_away', 'New Candidate');
-
-      fetch(this.props.submitURL, {
-        method: 'POST',
-        cache: 'no-cache',
-        credentials: 'same-origin',
-        body: formObject,
-        })
-      .then((resp) => {
-        if (resp.ok && resp.status === 201) {
-          resp.json().then((data) => {
-            this.setState({newData: data});
-            this.setState({isCreated: true});
-          });
-        } else {
-          resp.json().then((message) => {
-            swal('Error!', message, 'error');
-          });
-        }
-      })
-      .catch((error) => {
-        console.error(error);
-      });
+    if (!this.validateMatchDate()) {
+      console.error('something wong');
+      return;
     }
+
+    const formdata = this.state.formData;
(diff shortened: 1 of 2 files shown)


johnsaigle:
Integration test failure:

>There was 1 error:
> 1) NewProfileTestIntegrationTest::testNewProfilePageLoads
System info: host: '2ab33cd83ae7', ip: '172.18.0.3', os.name: 'Linux', os.arch: 'amd64', os.version: '4.15.0-1028-gcp', java.version: '1.8.0_91'
/app/modules/new_profile/test/new_profileTest.php:60
--

ridz1208:
@xlecours I do like this approach. I have a question though. I we decide to add a new field tomorrow for new_profile... i.e. RecruitmentDiagnosis

Do we add that new field to the v0.0.2 API? or do we have to add it to the 0.0.3-dev and then wait for a release of the API?

modules/new_profile/jsx/NewProfileIndex.js
```diff
@@ -74,43 +74,44 @@ class NewProfileIndex extends React.Component {
...
-      });
-    } else {
-      let formData = this.state.formData;
-      let formObject = new FormData();
-      for (let key in formData) {
-        if (formData[key] !== '') {
-          formObject.append(key, formData[key]);
-        }
+    if (!this.validateMatchDate()) {
+      console.error('something wong');
```
maltheism: Maybe output of "date validation error" would be better?

modules/new_profile/jsx/NewProfileIndex.js
```diff
@@ -246,20 +247,26 @@ class NewProfileIndex extends React.Component {
...
+      </div>
+    );
   }
 }
 window.addEventListener('load', () => {
   ReactDOM.render(
     <NewProfileIndex
       dataURL = {`${loris.BaseURL}/new_profile/?format=json`}
-      submitURL = {`${loris.BaseURL}/new_profile/`}
+      submitURL = {`${loris.BaseURL}/api/v0.0.2/candidates`}
```
maltheism: The loris.BaseURL could be replaced with window.location.origin and I'm unsure why we use the BaseURL instead.

xlecours: I think it was not supported by one of the browser. (IE)

johnsaigle: Since we don't support IE we could probably go with `window.location.origin` going forward

driusan: BaseURL is used throughout LORIS because `window.location.origin` will break on anything that isn't hosted under the root of the domain name (as far as I know)

xlecours:
> Do we add that new field to the v0.0.2 API? or do we have to add it to the 0.0.3-dev and then wait for a release of the API?
If it's not a required field, it won't break backward compatibility with 0.0.2. Currently, both version are handled by the same code.

johnsaigle:
@xlecours This needs a rebase

PR #5345, opened 2019-10-21
[DQT] Fix 500 error caused by bad Exception name

johnsaigle:
## Brief summary of changes

"ConfigException" doesn't exist and causing LORIS to crash when an exception is thrown.

1. Using invalid CouchDB credentials, go the DQT module. LORIS will crash.


johnsaigle:
This PR is set to go to bugfix but this problem also exists on minor and major.

PR #5353, opened 2019-10-22
[User Accounts] Hide UI option allowing users to edit their own approval status

### LINKED ISSUE #4743: [User account] I am able to edit my own 'Pending Approval' status
As the title says. This does not really make sense...You should only be able to edit another user's 'Pending Approval' or 'Active' status IMHO.

johnsaigle:
## Brief summary of changes

Fix #4743 where a user could accidentally lock themselves out but editing their own approval status.

1. Checkout this branch, go to user accounts.
2. You will no longer see the `Pending Approval` fields when editing your own accounts

### DIFF
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -823,9 +823,9 @@ class Edit_User extends \NDB_Form
                     "examiner_pending",
                     "Pending Approval: ",
                     array(
-                     null => "",
-                     'Y'  => 'Yes',
-                     'N'  => 'No',
+                        null => "",
+                        'Y'  => 'Yes',
+                        'N'  => 'No',
                     )
                 );
             }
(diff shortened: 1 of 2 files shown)


johnsaigle:
I think I'll have to update the integration tests

ridz1208:
@johnsaigle i made an attempt to fix the failing test, however I have a question.

why not also hide the active field ? this way you cant lock urself out by setting it to no.

johnsaigle:
@ridz1208 I updated this PR to also hide the active status as well as the active window. 

I'm not really clear on why you updated the integration test the way you did. Can you explain the changes?

ridz1208:
@johnsaigle The logged in user for testingg is UnitTester and the changes are being tested on the user itself. so obviously it's failing when it gets to testing the pending approval dropdown.

I tried adding an other user and editing that one instead but I'm getting a weird duplicated value error when there should not be any other 999991 user from what I can tell.

@kongtiaowang is looking into that particular error

kongtiaowang:
I will reopen this to debug Travis.

Issue #5360, opened 2019-10-23
[21.0] [access_profile] Can't open any candidate profile without "across all sites access candidate profiles"


leighevans:
Unable to access any candidate profile, despite having all site affiliations, without "Across all sites access candidate profiles". The PSCID is un-clickable and you cannot open the candidate's profile to see their timepoints/do data entry etc. without this permission. 

![67125451-cd30fd00-f1c2-11e9-91be-01be6334549e](https://user-images.githubusercontent.com/32552953/67409888-35982980-f589-11e9-93c8-cd55aec8fa5f.png)

driusan:
This is a behaviour change, not a bug, but Samir approved the change of behaviour.

Issue #5359, opened 2019-10-23
[21.0] [access_profile] "Visit Label" in Selection Filter no longer a dropdown


leighevans:
it's now a textbox instead of a dropdown which could be an issue for users since the visits have backend naming conventions
![Screen Shot 2019-10-22 at 1 53 54 PM](https://user-images.githubusercontent.com/32552953/67409626-d0dccf00-f588-11e9-9ed6-95d2f589ee6c.png)

driusan:
merged

PR #5362, opened 2019-10-23
[Candidate List] change visitLabel filter to dropdown, fix link issue

jesscall:
## Brief summary of changes
This PR does two things:
1. It changes the visit label filter from a textbox to a dropdown. 
2. It addresses the issue reported in #5360 in which if a user does not have `across all sites` permission, links to the candidate profiles do not appear. However, the list of candidates returned in the table is already based on their permissions and only displays the candidates that they SHOULD have access to. 

1. Try filtering by visit label 
2. Remove the `across all site` permission and see if you are able to access profiles. 

* #5360
* #5359

### DIFF
--- a/jsx/Filter.js
+++ b/jsx/Filter.js
@@ -39,7 +39,7 @@ class Filter extends Component {
   onFieldUpdate(name, value, id, type) {
     const searchParams = new URLSearchParams(location.search);
     const filter = JSON.parse(JSON.stringify(this.props.filter));
-    const exactMatch = (!(type === 'textbox' || type === 'date'));
+    const exactMatch = (!(type === 'textbox' || type === 'date' || type === 'select'));
     if (value === null || value === '' || (value.constructor === Array && value.length === 0)) {
       delete filter[name];
       searchParams.delete(name);
--- a/jsx/FilterForm.js
+++ b/jsx/FilterForm.js
@@ -132,7 +132,6 @@ class FilterForm extends Component {
       } else { // null and undefined handled here
         filter[key].value = '';
       }
-      filter[key].exactMatch = (type === 'SelectElement' || type === 'select');
     }
     if (filter && key && value === '') {
       delete filter[key];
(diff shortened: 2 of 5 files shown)


johnsaigle:
It looks like Travis is having an issue with the integration tests. Maybe they're expecting a textbox to exist?

The issues are arising from these lines:
```
/app/modules/candidate_list/test/candidate_listTest.php:206

/app/modules/candidate_list/test/candidate_listTest.php:172
```

jesscall:
Not sure how to fix Travis in this case.. 
I though it might be related to the number of records indicated on line 172 of `/app/modules/candidate_list/test/candidate_listTest.php` not matching the result set.

However changing that to the correct number of 'V1' visits in RB's session table doesn't remedy the situation...

Does anyone have some insight on debugging Travis / understanding the test case ?

kongtiaowang:
@jesscall 
It fixed the visitLabel filter as dropdown, but you are using the react filter component which only searches exactly dropdown data. It means only can filter the "V1" as visit label, not " V1 and V2".
Some candidate has more than one visit label. So if you want to fix the dropdown filter, fix the component first.

jesscall:
So I edited the filter component. In my opinion, a `select` should not be filtered by exact match since it would never match candidates who have more than one visit.

jsx/Filter.js
```diff
@@ -39,7 +39,7 @@ class Filter extends Component {
   onFieldUpdate(name, value, id, type) {
     const searchParams = new URLSearchParams(location.search);
     const filter = JSON.parse(JSON.stringify(this.props.filter));
-    const exactMatch = (!(type === 'textbox' || type === 'date'));
+    const exactMatch = (!(type === 'textbox' || type === 'date' || type === 'select'));
```
driusan: Why is this PR removing the exactMatch criteria for selects? If you select a dropdown with id `1` it shouldn't match id `11`

jesscall: Right now, each field in the data table is a string. Let's look at an example:

**X** has visits: `V1`, `V2`, `V3`
**Y** has visit: `V2`

If you select `V2` in the dropdown, you'd expect that both **X** and **Y** are shown in the filtered data table. However, that is not the case since at the moment, it is looking for an exact match. Thus, the only result you will see is a row for **Y**.

I agree with the case you've given above, but I also don't think we should be returning a subset of results. Any suggestions?

Issue #5358, opened 2019-10-23
Deprecate Class::getData('variable') idioms in favour of Class::getVariable() functions


johnsaigle:
Many LORIS classes are structured in a weird way.

E.g. The TimePoint class has information stored in a private array:

```php
private var $_timePointInfo = array();
```

This array gets populates when the object gets instantiated. It contains numerous variables about the TimePoint such as its center ID, subprojectID, etc.

There exists a function

```php
public function getData(?string $key = null);
```

Which can be used to get the value of these variables. At the same time, there exist individual functions to do this:

```php
public function getCenterID();
```
```php
public function getSubprojectID;
```

---

## So what?

* There's no point having a private array of variables that can be accessed publicly. They're not really private in this case
* Using `TimePoint::getData()` prevents type hinting since this function needs to return `?mixed` as it is a catch-all for all data the object uses
* Creating an array of what are essentially class variables defeats the whole point of having a class. You might as well use arrays everywhere because this basically reduces the class to being an anonymous struct with no typing
* Without class variables, we will be unable to take advantage of class variable typing in PHP 7.4
* All of these points are likely to make things more error-prone as scope isn't respected and types can't be used.

## Solution

Refactor TimePoint and other classes that are designed this way to have EITHER

1. Public class variables OR
2. Private class variables with traditional getters/setters

Additionally, new code going into LORIS should use something like `TimePoint::getCenterID()` rather than `TimePoint::getData('centerID')` as this will allow us to make use of type hinting on functions of the latter type as well as smooth the transition to better classes.

johnsaigle:
It may also be a good idea to put a DeprecationException in `Class::getData()` functions informing developers to use more specific functions instead.

driusan:
>    There's no point having a private array of variables that can be accessed publicly. They're not really private in this case

That's not really true. It prevents setting and allows the possibility of modifying the data in some way before returning it.

>    Using TimePoint::getData() prevents type hinting since this function needs to return ?mixed as it is a catch-all for all data the object uses

I think the easiest solution is simply to make it protected and use it internally in the getters, which can do their own casting. 

>    Creating an array of what are essentially class variables defeats the whole point of having a class. You might as well use arrays everywhere because this basically reduces the class to being an anonymous struct with no typing

Since it's private and people need to use the getter, this is just an implementation detail. I suspect it's done this way so that it can be easily retrieved from the database and put into the class.. (as an aside: arrays and objects are just different syntax on the same internal structure in PHP.)

[...]

PR #5366, opened 2019-10-23
[Candidate] Create class variables and clean up the code

johnsaigle:
## Brief summary of changes

* Adds class variables to represent Candidate data
* Remove `_candidateInfo` array that was previously storing all candidate information
* Group all getter methods together in the class
* Add var annotations to new class variables + suggestions for further work
* Deprecate `getData()` function in favour of more specific getter methods
* Remove references to deprecated "Ethnicity" data
* Small readability improvements

In general, this allows us to better take advantage of the class, use types, and improves the readability of the code.

1. The class should work the same as before. Calling code using `getData()` will need to be updated.

* #5358

### DIFF
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -166,7 +166,8 @@ QUERY;
 
         if (!is_array($row) || sizeof($row) == 0) {
             throw new \LorisException(
-                "Could not select Candidate data from the database (DCCID: $candID)",
+                "Could not select Candidate data from the database: "
+                . "(CandID: $candID)",
                 CANDIDATE_INVALID
             );
         }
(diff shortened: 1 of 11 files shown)


php/libraries/Candidate.class.inc
```diff
@@ -35,9 +35,74 @@ define('EDC_NOT_SPECIFIED', 5);
  */
 class Candidate
 {
-    var $candidateInfo;
```
johnsaigle: Since `$candidateInfo` is public, effectively all the candidate data before was public before and could have been changed using 

```php
Candidate->candidateInfo['field'] = $value;
```

I'm not sure if this is done in practice, but I've left these variables as being public in scope.

If this is not desirable, I can change them to be private. I'm not sure which variables (if any) should have setter functions in this case.

johnsaigle:
Alright, I'll close this and give that a shot.

PR #5370, opened 2019-10-23
[user_account] can't add or edit a user on major

### LINKED ISSUE #5368: [major] can't add new user.

kongtiaowang:
Admin user can't add or edit a user.

Resolves #5368

### DIFF
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -1046,6 +1046,7 @@ class Edit_User extends \NDB_Form
         ) {
             try {
                 // New password must be different than current one
+                $password = new \Password($values['Password_hash']);
                 $passwordChanged = \User::factory($this->identifier)
                     ->passwordChanged($values['Password_hash']);
                 if (! $passwordChanged) {
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -1048,7 +1048,7 @@ class Edit_User extends \NDB_Form
                 $password = new \Password($values['Password_hash']);
                 // New password must be different than current one
                 $passwordChanged = \User::factory($this->identifier)
-                    ->passwordChanged($password);
+                    ->passwordChanged((string)$password);
                 if (! $passwordChanged) {
                     $errors['Password_Group'] = 'New and old passwords '
                         . 'are identical: please choose another one';
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -1048,7 +1048,7 @@ class Edit_User extends \NDB_Form
                 $password = new \Password($values['Password_hash']);
                 // New password must be different than current one
                 $passwordChanged = \User::factory($this->identifier)
-                    ->passwordChanged($_POST['password']);
+                    ->passwordChanged($password);
                 if (! $passwordChanged) {
                     $errors['Password_Group'] = 'New and old passwords '
                         . 'are identical: please choose another one';
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -1045,10 +1045,9 @@ class Edit_User extends \NDB_Form
             && (!empty($values['Password_hash']) || !empty($values['__Confirm']))
         ) {
             try {
-                $password = new \Password($values['Password_hash']);
                 // New password must be different than current one
                 $passwordChanged = \User::factory($this->identifier)
-                    ->passwordChanged((string)$password);
+                    ->passwordChanged($values['Password_hash']);
                 if (! $passwordChanged) {
                     $errors['Password_Group'] = 'New and old passwords '
                         . 'are identical: please choose another one';
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -1046,7 +1046,7 @@ class Edit_User extends \NDB_Form
         ) {
             try {
                 // New password must be different than current one
-                $password = new \Password($values['Password_hash']);
+                $password        = new \Password($values['Password_hash']);
                 $passwordChanged = \User::factory($this->identifier)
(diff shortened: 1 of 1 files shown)


modules/user_accounts/php/edit_user.class.inc
```diff
@@ -1048,7 +1048,7 @@ class Edit_User extends \NDB_Form
                 $password = new \Password($values['Password_hash']);
                 // New password must be different than current one
                 $passwordChanged = \User::factory($this->identifier)
-                    ->passwordChanged($_POST['password']);
+                    ->passwordChanged($password);
```
johnsaigle:
```suggestion
                    ->passwordChanged($_POST['Password_hash']);
```

I believe this should fix your problem @kongtiaowang

kongtiaowang: @johnsaigle This is not correct. In this function only take care values 
function _process($values).

johnsaigle: This fixed an issue for me when I tried to change a user's password on `major`.

It would be helpful if you described what error you're getting.

johnsaigle:
https://github.com/aces/Loris/blob/a8612ff63289b3dc412241328557a0bca372914e/php/libraries/Password.class.inc#L210-L218

toString returns the hashed value of the password, not the plaintext password.

https://github.com/aces/Loris/blob/a8612ff63289b3dc412241328557a0bca372914e/php/libraries/User.class.inc#L474-L503

The toString() function needs the plaintext password in order to use [password_verify()](https://www.php.net/manual/en/function.password-verify.php). If the hash is supplied instead of the plaintext, then `passwordChanged()`will always return true even when the actual password is the same because there will never be a match.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -1048,7 +1048,7 @@ class Edit_User extends \NDB_Form
                 $password = new \Password($values['Password_hash']);
                 // New password must be different than current one
                 $passwordChanged = \User::factory($this->identifier)
-                    ->passwordChanged($_POST['password']);
+                    ->passwordChanged((string)$password);
```
johnsaigle:
```suggestion
                    ->passwordChanged($_POST['Password_hash']);
```
See discussion for details

kongtiaowang: $password = new \Password($values['Password_hash']); useless??

johnsaigle: > useless??

No it will throw an InvalidArgumentException when the password is too weak or not long enough:
https://github.com/aces/Loris/blob/a8612ff63289b3dc412241328557a0bca372914e/php/libraries/Password.class.inc#L102

Validate gets called automatically when a Password object is created.

The try-catch beginning on 1047 in `edit_user.php` converts this into a front-end error message if the Password is too weak.

kongtiaowang: Why $_POST['Password_hash']? why not $values['Password_hash'] I can't find any $_POST used in this whole file.

johnsaigle: `$values['Password_hash']` should work too.

modules/user_accounts/php/edit_user.class.inc
kongtiaowang: [Wed Oct 23 14:18:34.013500 2019] [php7:error] [pid 7294] [client 192.168.122.1:53884] PHP Fatal error:  Uncaught TypeError: Argument 1 passed to User::passwordChanged() must be of the type string, null given, called in /var/www/Loris/modules/user_accounts/php/edit_user.class.inc on line 1051 and defined in /var/www/Loris/php/libraries/User.class.inc:483\nStack trace:\n#0 /var/www/Loris/modules/user_accounts/php/edit_user.class.inc(1051): User->passwordChanged(NULL)\n#1 /var/www/Loris/php/libraries/LorisForm.class.inc(1324): LORIS\\user_accounts\\Edit_User->_validateEditUser(Array)\n#2 /var/www/Loris/php/libraries/NDB_Form.class.inc(132): LorisForm->validate()\n#3 /var/www/Loris/php/libraries/NDB_Form.class.inc(192): NDB_Form->save()\n#4 /var/www/Loris/modules/user_accounts/php/edit_user.class.inc(579): NDB_Form->handle(Object(Zend\\Diactoros\\ServerRequest))\n#5 /var/www/Loris/src/Middleware/UserPageDecorationMiddleware.php(187): LORIS\\user_accounts\\Edit_User->handle(Object(Zend\\Diactoros\\ServerRequest))\n#6 /var/www/Loris/src/Middleware/PageDecorationMiddleware.php(53): LORIS\\Middleware\\UserPageDecorationMiddleware->proce in /var/www/Loris/php/libraries/User.class.inc on line 483, referer: https://test-loris-dev.loris.ca/user_accounts/edit_user/

johnsaigle: Yeah that's what I got when I tried to test. `$_POST['Password_hash']` or `$values['Password_hash']` should fix that.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -1045,10 +1045,9 @@ class Edit_User extends \NDB_Form
             && (!empty($values['Password_hash']) || !empty($values['__Confirm']))
         ) {
             try {
-                $password = new \Password($values['Password_hash']);
```
johnsaigle: This line is necessary. please add it back or else the password won't get validated.

kongtiaowang: already back

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -1045,10 +1045,10 @@ class Edit_User extends \NDB_Form
...
                 // New password must be different than current one
+                $password        = new \Password($values['Password_hash']);
                 $passwordChanged = \User::factory($this->identifier)
-                    ->passwordChanged($_POST['password']);
+                    ->passwordChanged($values['Password_hash']);
```
driusan: Whether or not this fixes the bug, removing a superglobal reference is always a step forward..

PR #5374, opened 2019-10-24
[user_accounts] fix react table url with undefined param

kongtiaowang:
Click the username URL link gets 500 error.
Oct-25-2019
Test: click the username link in the react table to test.


driusan:
This should be fixed by #5381 instead

kongtiaowang:
Not the same issue. It hadn't been fixed by #5381.

driusan:
did you run make before testing #5381 and make sure your old javascript wasn't cached? It's fixing the issue when I test it.

Issue #5364, opened 2019-10-23
[21.0] [DQT] Multiple buttons not working


leighevans:
- "Remove All" fields button not working
- In order to successfully save a query, the query must be named with no spaces (but the example is "My Query" with a space so this is misleading. If a space is included there is no error message, but the query simply doesn't save

driusan:
Merged

PR #5379, opened 2019-10-24
[dataquery] Fixed broken link for 'deleteAll' and fixed save error on query names that contain spaces

jesscall:
### Brief summary of changes
This PR addresses the issues reported in #5364 
1. The `Delete All` link in DQT was broken and not actually deselecting the fields. 
2. You couldn't save a query if it had a space in the name. No error was displayed, it just wouldn't save. The code now replaces any space by underscores. 

1. Head over to DQT. 
2. Add some fields to your query. 
3. Attempt to deselect those fields using the `Delete All` button. 
4. Try saving a query that has spaces in the name. When you refresh, your query should appear. 

* #5364
* aces/CCNA#3342

### DIFF
--- a/modules/dataquery/jsx/react.app.js
+++ b/modules/dataquery/jsx/react.app.js
@@ -173,7 +173,7 @@ class DataQueryApp extends Component {
       for (let i = 0; i < this.state.queryIDs[key].length; i += 1) {
         let curRequest;
         curRequest = Promise.resolve(
-          $.ajax(loris.BaseURL + '/AjaxHelper.php?Module=dataquery&script=GetDoc.php&DocID=' + this.state.queryIDs[key][i]), {
+          $.ajax(loris.BaseURL + '/AjaxHelper.php?Module=dataquery&script=GetDoc.php&DocID=' + encodeURIComponent(this.state.queryIDs[key][i])), {
             data: {
               DocID: this.state.queryIDs[key][i]
             },
@@ -233,12 +233,11 @@ class DataQueryApp extends Component {
     // Used to save the current query
 
     let filter = this.saveFilterGroup(this.state.filter);
-    let queryName = name.replace(/ /g,'_');
 
     $.post(loris.BaseURL + '/AjaxHelper.php?Module=dataquery&script=saveQuery.php', {
       Fields: this.state.selectedFields,
       Filters: filter,
-      QueryName: queryName,
+      QueryName: name,
       SharedQuery: shared,
       OverwriteQuery: override
     }, (data) => {
--- a/modules/dataquery/jsx/react.app.js
+++ b/modules/dataquery/jsx/react.app.js
@@ -233,11 +233,12 @@ class DataQueryApp extends Component {
     // Used to save the current query
 
     let filter = this.saveFilterGroup(this.state.filter);
+    let queryName = name.replace(/ /g,'_');
 
     $.post(loris.BaseURL + '/AjaxHelper.php?Module=dataquery&script=saveQuery.php', {
       Fields: this.state.selectedFields,
       Filters: filter,
-      QueryName: name,
+      QueryName: queryName,
       SharedQuery: shared,
       OverwriteQuery: override
     }, (data) => {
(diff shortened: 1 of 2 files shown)


modules/dataquery/jsx/react.app.js
```diff
@@ -233,11 +233,12 @@ class DataQueryApp extends Component {
     // Used to save the current query
 
     let filter = this.saveFilterGroup(this.state.filter);
+    let queryName = name.replace(/ /g,'_');
```
driusan: This seems to be bandaging over the problem without fixing the root cause. Why can't it contain a space? Is there somewhere that needs to be urlencode/decoded that it isn't happening? If so, this would still break with other characters that have special meaning (+, &, etc)

@nicolasbrossard you had something regarding `+` and saved queries on IBIS, right? Is it related?

maltheism: @driusan, @jesscall, @nicolasbrossard it seems to be used here:
https://github.com/aces/Loris/blob/e7e999dc8f768ae51e630616bd302c5360fc0662/modules/dataquery/ajax/saveQuery.php#L36

and spaces break it.

maltheism: Maybe the frontend fix should be removed and the fix could be done here?

jesscall: @driusan @maltheism 
I've been looking into this and I think the problem runs deeper than the line you linked. 

When saving to the couchDB server with a encoded query name, we start to run into problems with actually retrieving the documents. 

For example I used the query name `Clinical DTT`.
This then gets attributed an id of `jessica.callegaro@mail.mcgill.ca_Clinical+DTT`

However, when attempting to retrieve this document from couchDB, for some reason the `+` is not understood. 

```
curl -X GET http://ccna_dev_user:XXXX@couchhost-dev.acelab.ca:5984/ccna_dev_dqt/"jessica.callegaro@mail.mcgill.ca_Clinical+DTT"

{"error":"not_found","reason":"missing"}
```

Instead of using `+`, its corresponding encoding  needs to be used for retrieval (`%2B`) It looks something like this:

```
curl -X GET http://ccna_dev_user:XXXX@couchhost-dev.acelab.ca:5984/ccna_dev_dqt/"jessica.callegaro@mail.mcgill.ca_Clinical%2BDTT"

{"_id":"jessica.callegaro@mail.mcgill.ca_Clinical+DTT","_rev":"1-d97f5f2aeb78ed68654beb3e299ab134","Meta":{"DocType":"SavedQuery","user":"jessica.callegaro@mail.mcgill.ca","name":"Clinical DTT"},"Fields":{ ... }
```

So once we arrive here https://github.com/aces/Loris/blob/e7e999dc8f768ae51e630616bd302c5360fc0662/modules/dataquery/ajax/GetDoc.php#L33 in the code, it looks something like this:

`$_REQUEST['DocID']` = `jessica.callegaro@mail.mcgill.ca_Clinical DTT`
followed by...
`$docID (encoded)`  = `jessica.callegaro%40mail.mcgill.ca_Clinical+DTT` <---- this will always return undefined!

It's kind of as if we need to double encode the space to actually be able to retrieve a document that contains spaces? idk i spent a lot of time being confused by this, so i hope this explanation made sense? I'd love to hear your opinions though

nicolasbrossard: @driusan @maltheism @jesscall I had a similar problem in IBIS: one of our users had saved a query that contained a `+` and the DQT could not retrieve it. I don't think you have to do anything special when saving a query. When you retrieve a query though, you have to call `encodeURIComponent` on the query ID. Look for this ajax call in `modules/dataquery/jsx/react.app.js`:

```javascript
$.ajax(loris.BaseURL + "/AjaxHelper.php?Module=dataquery&script=GetDoc.php&DocID=" + that.state.queryIDs[key][i])
```

It needs to be replaced by:

```javascript
$.ajax(loris.BaseURL + "/AjaxHelper.php?Module=dataquery&script=GetDoc.php&DocID=" + encodeURIComponent(that.state.queryIDs[key][i]))
```

This fixed the issue we had in IBIS.

PR #5381, opened 2019-10-24
Fixing bugs introduced when optimizing DataTable

HenriRabalais:
When optimising the Datatable, #5299 introduced some bugs in the file. This PR removes those bugs.

1. It maps the headers to values to allow the `this.props.getFormattedCell` to function properly.
2. It pushes `rowData` to `filterData` so that when downloading the CSV, only the filtered data is passed.
3. Checks for cases where `data=null` because `null.toLowerCase()` cannot be executed.

There are still issues with pagination, which could be fixed in either this or another PR.

### DIFF
--- a/jsx/DataTable.js
+++ b/jsx/DataTable.js
@@ -424,7 +424,7 @@ class DataTable extends Component {
 
             let row = {};
             this.props.fields
-              .forEach((field, k) => {row[field.label] = rowData[k]});
+              .forEach((field, k) => row[field.label] = rowData[k]);
 
             // Get custom cell formatting if available
             if (this.props.getFormattedCell) {
--- a/jsx/DataTable.js
+++ b/jsx/DataTable.js
@@ -287,7 +287,7 @@ class DataTable extends Component {
           }
           break;
         default:
-            searchString = data.toLowerCase();
+            searchString = data ? data.toString().toLowerCase() : '';
             if (exactMatch) {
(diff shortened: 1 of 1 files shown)


jsx/DataTable.js
```diff
@@ -421,13 +422,16 @@ class DataTable extends Component {
             let celldata = rowData[j];
             let cell = null;
 
-
+            const row = {};
+            this.props.fields.forEach((field, k) => {
+              row[field.label] = rowData[k];
+            });
```
driusan: Since this is adding a new loop inside of a loop, did you verify that this doesn't affect the optimization? Also, "const" doesn't feel right considering you then go on to modify it.. (I know it's just the reference that's const, not the data, at the reference, but it still doesn't feel right..)

driusan:
I tested the above PRs with this branch. Will wait for confirmation about no major performance regressions and someone else to review before merging..

jsx/DataTable.js
HenriRabalais: I've tried it on Biobank with 20k+ rows and it doesn't seem to be a problem!

HenriRabalais: I can also change line 425 to `let` if you still want

driusan: Doesn't really matter to me.. its valid javascript in either case, it just feels semantically weird to me.

PR #5385, opened 2019-10-28
[ReadTheDocs] Fix build failure by adding explicit dependencies

johnsaigle:
## Brief summary of changes

Old versions of mkdocs (used by ReadTheDocs) have incompatibilities with Python 3.7. https://github.com/readthedocs/readthedocs.org/issues/5332

This PR adds explicit settings for our project so that ReadTheDocs will run with Python 3.7 on a relatively new version of mkdocs. This should fix our current [build issue](https://readthedocs.org/projects/acesloris/builds/9873748/).


johnsaigle:
Exactly. It's done on the ReadTheDocs site though so pip/python isn't a dependency for LORIS or anything like that.

PR #5388, opened 2019-10-28
Add configuration file for the Stale Bot tool

johnsaigle:
## Brief summary of changes

Adds a configuration for [stale bot](https://github.com/probot/stale), a tool that marks inactive issues and PRs and eventually closes them if no action is taken on them.

If used, it should assist with triage and keep us on our toes about what may be slipping through the cracks.

I used the default config file here, modified slightly to match the labels we use.

In order for this to work, someone with Admin privileges must [enable the Stale bot](https://github.com/apps/stale/installations/new). Until then this PR will remain in Draft state.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -5,7 +5,7 @@ daysUntilStale: 60
 
 # Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
 # Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
-daysUntilClose: 7
+daysUntilClose: false
 
 # Only issues or pull requests with all of these labels are checked if stale. Defaults to `[]` (disabled)
 onlyLabels: []
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -37,8 +37,9 @@ markComment: >
   for your contributions.
 
 # Comment to post when removing the stale label.
-# unmarkComment: >
-#   Your comment here.
+unmarkComment: >
+  The Stale label is being removed automatically because some activity has occurred or because the
+  developers have decided that this pull request is important and should not continue to be overlooked.
 
 # Comment to post when closing a stale Issue or Pull Request.
 # closeComment: >
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -12,8 +12,6 @@ onlyLabels: []
 
 # Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
 exemptLabels:
-  - pinned
-  - security
   - "Bug Fix"
   - "Security"
   - "Critical to release"
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -18,6 +18,7 @@ exemptLabels:
   - "Security"
   - "Critical to release"
   - "Blocked"
+  - "Blocking"
 
 # Set to true to ignore issues in a project (defaults to false)
 exemptProjects: false
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,63 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
+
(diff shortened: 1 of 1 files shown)


johnsaigle:
Bringing this up in tomorrow's LORIS meeting to see if it's desirable.

johnsaigle:
@driusan I'm assigning this to you because you'll need to enable Stale bot before this goes into LORIS.

https://probot.github.io/apps/stale/

.github/stale.yml
```diff
@@ -0,0 +1,63 @@
...
+# Only issues or pull requests with all of these labels are checked if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - pinned
+  - security
+  - "Bug Fix"
+  - "Security"
+  - "Critical to release"
```
PapillonMcGill: What about PR that are `block` by another PR and the latest is active?

johnsaigle: Well we can't specify logic like "is the latest active?"

I don't feel much one way or the other about adding Blocked here. Blocked is more of a statement about the workflow of GitHub then it is about the contents of the code in the PR. The other labels listed here reflect types of code changes we absolutely don't want to miss.

PapillonMcGill: Could we had `"blockec"`?

johnsaigle: sure, I don't mind

johnsaigle:
```suggestion
  - "Critical to release"
  - "Blocked"
```

driusan: Unresolving this because Blocked doesn't seem to be included.

johnsaigle: Thanks for reviewing. I had included it at some point (pretty sure anyway). Maybe have been lost somehow in my rebasing.

```suggestion
  - "Critical to release"
  - "Blocked"
```

driusan: Can you also add "Blocking" so that things that are blocked don't get blocked forever if the blocker goes stale?

.github/stale.yml
```diff
@@ -0,0 +1,64 @@
...
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
+
+# Only issues or pull requests with all of these labels are checked if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - pinned
+  - security
```
driusan: I don't think we have the above 2 labels

.github/stale.yml
```diff
@@ -0,0 +1,64 @@
...
+  - "Bug Fix"
+  - "Security"
+  - "Critical to release"
+  - "Blocked"
```
johnsaigle:
```suggestion
  - "Blocked"
  - "Blocking"
```

.github/stale.yml
```diff
@@ -0,0 +1,64 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
```
driusan:
```suggestion
daysUntilClose: 30
```

7 days is too short. The stale PR can easily be missed for a week if someone is on vacation.

johnsaigle: I think we decided in the meeting that we wouldn't auto close any issues or PRs actually. At least initially.
```suggestion
daysUntilClose: false
```

driusan: I don't have any recollection of that. I don't see any reason for this tool without autoclosing.

.github/stale.yml
```diff
@@ -0,0 +1,63 @@
...
+exemptProjects: false
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: false
+
+# Set to true to ignore issues with an assignee (defaults to false)
+exemptAssignees: false
+
+# Label to use when marking as stale
+staleLabel: wontfix
```
johnsaigle:
```suggestion
staleLabel: "Stale"
```

I've added a stale label

johnsaigle:
@driusan Bumping this. Should be good to go

PR #5390, opened 2019-10-28
[Core] Add value objects for EntityType and Sex

johnsaigle:
## Brief summary of changes

Adds Value Object classes to help with validation for these two fields in the `candidate` table.

This allows us to use more specific types when processing candidate information as right now we can only validate these values on an ad-hoc basis as they are treated as generic string arrays.

### DIFF
--- a/modules/dicom_archive/php/dicomarchiverow.class.inc
+++ b/modules/dicom_archive/php/dicomarchiverow.class.inc
@@ -59,7 +59,7 @@ class DICOMArchiveRow implements \LORIS\Data\DataInstance
      *
      * @return integer The CenterID
      */
-    public function getCenterID(): int
+    public function getCenterID()
     {
         return $this->CenterID;
     }
--- a/modules/dicom_archive/php/dicomarchiverow.class.inc
+++ b/modules/dicom_archive/php/dicomarchiverow.class.inc
@@ -59,7 +59,7 @@ class DICOMArchiveRow implements \LORIS\Data\DataInstance
      *
      * @return integer The CenterID
      */
-    public function getCenterID()
+    public function getCenterID(): int
     {
         return $this->CenterID;
     }
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -98,8 +98,17 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
         }
 
         // store user data in object property
-        foreach ($row AS $key=>$value) {
-            $this->candidateInfo[$key] = $value;
+        foreach ($row as $key=>$value) {
+            switch ($key){
+            case 'Sex':
+                $this->candidateInfo[$key] = new Sex($value);
+                break;
+            case 'EntityType':
+                $this->candidateInfo[$key] = new EntityType($value);
+                break;
+            default:
+                $this->candidateInfo[$key] = $value;
+            }
         }
 
         $this->candidateInfo['ProjectTitle'] = $this->getProjectTitle();
@@ -505,7 +514,7 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
      */
     function getCandidateSex(): ?Sex
     {
-        return new Sex($this->candidateInfo["Sex"]);
+        return $this->candidateInfo["Sex"];
     }
 
     /**
@@ -515,7 +524,7 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
      */
     function getEntityType(): ?EntityType
     {
-        return new EntityType($this->candidateInfo["EntityType"]);
+        return $this->candidateInfo["EntityType"];
     }
 
     /**
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -509,7 +509,7 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
     }
 
     /**
-     * Return the candidate's sex
+     * Return the candidate's entity type.
      *
      * @return ?EntityType
      */
@@ -913,4 +913,3 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
         return '/^'.$regex.'$/i';
     }
 }
-
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -7,6 +7,10 @@
  */
 
 use LORIS\StudyEntities\Candidate\CandID;
+use LORIS\StudyEntities\Candidate\EntityType;
+use LORIS\StudyEntities\Candidate\Sex;
+
+/* Define constants used by the file */
 // error codes
 define('CANDIDATE_INVALID', 1);
 define('PSCID_NOT_SPECIFIED', 2);
@@ -182,7 +186,7 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
      *                                           written as YYYY-MM-DD. Value should
      *                                           be null unless the useEDC setting
      *                                           is configured to be true.
-     * @param string      $sex                   sex, 'Male', 'Female', or 'Other'
+     * @param Sex         $sex                   An instance of the Sex class.
      * @param string|null $PSCID                 PSCID specified by the user,
      *                                           if available
      * @param int|null    $registrationProjectID The project ID, if available
@@ -195,7 +199,7 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
         int $centerID,
         string $dateOfBirth,
         ?string $edc,
-        string $sex,
+        Sex $sex,
         ?string $PSCID = null,
         int $registrationProjectID = null
     ): CandID {
@@ -497,12 +501,21 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
     /**
      * Return the candidate's sex
      *
-     * @return string|null Male|Female|Other String describing the sex
-     *                     and null if scanner
+     * @return ?Sex
      */
-    function getCandidateSex(): ?string
+    function getCandidateSex(): ?Sex
     {
-        return $this->candidateInfo["Sex"];
+        return new Sex($this->candidateInfo["Sex"]);
+    }
+
+    /**
+     * Return the candidate's sex
+     *
+     * @return ?EntityType
+     */
+    function getEntityType(): ?EntityType
+    {
+        return new EntityType($this->candidateInfo["EntityType"]);
     }
 
     /**
--- a/php/libraries/EntityType.class.inc
+++ b/php/libraries/EntityType.class.inc
@@ -1,0 +1,71 @@
+<?php declare(strict_types=1);
+/**
+ * This defines the entity type for a Candidate. It is a Value Object used
+ * to assist with validation.
+ *
+ * PHP Version 7
+ *
+ * @category Main
+ * @package  LORIS
+ * @author   John Saigle <john.saigle@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+
+/**
+ * This defines the entity type for a Candidate. It is a Value Object used
+ * to assist with validation.
+ *
+ * @category Main
+ * @package  LORIS
+ * @author   John Saigle <john.saigle@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+class EntityType
+{
+    /* @var string */
+    $value = '';
+
+    private const VALID_VALUES = array('Human', 'Scanner');
+
+    /**
+     * Creates a new EntityType. Calls validate() immediately.
+     *
+     * @param string $value The value of this EntityType.
+     *
+     * @throws \DomainException When the value is not valid
+     */
+    final public function __construct(string $value)
+    {
+        if (!$this->validate($value)) {
+            throw new \DomainException(
+                'The value is not valid. Must be one of: '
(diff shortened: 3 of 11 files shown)


johnsaigle:
@xlecours @driusan Requesting your review as this is related to #5369.

I can see these sorts of classes adding a lot of boilerplate as they're created so let me know if you'd prefer that these functions be abstracted to a parent class.

php/libraries/EntityType.class.inc
```diff
@@ -0,0 +1,71 @@
...
+ * @category Main
+ * @package  LORIS
+ * @author   John Saigle <john.saigle@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+class EntityType
+{
+    /* @var string */
+    $value = '';
```
driusan: This is missing an access modifier.. I didn't think it was even valid without either an access modifier or the `var` keyword.

php/libraries/EntityType.class.inc
```diff
@@ -0,0 +1,71 @@
...
+ * to assist with validation.
+ *
+ * PHP Version 7
+ *
+ * @category Main
+ * @package  LORIS
+ * @author   John Saigle <john.saigle@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
```
driusan: I think this belongs in namespace LORIS\StudyEntities\Candidate; (and in the appropriate filesystem location for it..)

php/libraries/Sex.class.inc
```diff
@@ -0,0 +1,71 @@
...
+ * @category Main
+ * @package  LORIS
+ * @author   John Saigle <john.saigle@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+class Sex
+{
+    /* @var string */
+    $value = '';
```
driusan: missing access modifier

php/libraries/Sex.class.inc
```diff
@@ -0,0 +1,71 @@
+<?php declare(strict_types=1);
```
driusan: I'm a little less convinced that this one belongs in the Candidate namespace, but it probably does.

php/libraries/EntityType.class.inc
johnsaigle: OK, I'll move it.

php/libraries/EntityType.class.inc
xlecours: And there is non need for a default value.

src/StudyEntities/Candidate/EntityType.class.inc
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Ensures that EntityType is well-formed.
+     *
+     * @param string $value The value to be validated
+     *
+     * @return bool True if the value format is valid
+     */
+    protected function validate(string $value): bool
```
xlecours: private. There is no way a subclass can get its overridden method to be called by the final constructor of the super class.

src/StudyEntities/Candidate/Sex.class.inc
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Ensures that the value is well-formed.
+     *
+     * @param string $value The value to be validated
+     *
+     * @return bool True if the value format is valid
+     */
+    protected function validate(string $value): bool
```
xlecours: private. There is no way a subclass can get its overridden method to be called by the final constructor of the super class.

php/libraries/Candidate.class.inc
```diff
@@ -490,12 +493,22 @@ class Candidate
...
-     * @return string|null Male|Female String describing the sex
+     * @return ?Sex
+     */
+    function getCandidateSex(): ?Sex
+    {
+        return new Sex($this->candidateInfo["Sex"]);
+    }
+
+    /**
+     * Return the candidate's sex
```
maltheism: @johnsaigle Should `Return the candidate's sex` be changed?

```suggestion
     * Return the candidate's entity type.
```

src/StudyEntities/Candidate/Sex.php
```diff
@@ -0,0 +1,75 @@
...
+ * @author   John Saigle <john.saigle@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+class Sex
+{
+    /* @var string */
+    public $value;
+
+    private const VALID_VALUES = array(
```
maltheism: @johnsaigle I think LORIS has `Other` is a possibility for one of the valid values. I remember someone mentioned this in the Loris slack channel.

johnsaigle: That sort of rings a bell, but I can't find that option in the codebase. The database lists an enum of `Male`, `Female` so I went with that.

maltheism: @johnsaigle found what I vaguely remember here https://test-loris-dev.loris.ca/dicom_archive/?site=Data+Coordinating+Center&sex=O

johnsaigle: I opened an issue for it and added an item to the meeting agenda because these two should match but I'm not sure which is correct for projects.

johnsaigle:
Added 'Other' as an option for Sex.

Added Blocked label so the appropriate SQL PR gets merged first.

src/StudyEntities/Candidate/Sex.php
johnsaigle: @maltheism I've added 'Other' as an option

php/libraries/Candidate.class.inc
```diff
@@ -504,12 +507,21 @@ class Candidate
     /**
      * Return the candidate's sex
      *
-     * @return string|null Male|Female|Other String describing the sex
-     *                     and null if scanner
+     * @return ?Sex
```
zaliqarosli:
```suggestion
     * @return Sex|null Sex value object
```

php/libraries/Candidate.class.inc
```diff
@@ -504,12 +507,21 @@ class Candidate
...
+     */
+    function getCandidateSex(): ?Sex
+    {
+        return new Sex($this->candidateInfo["Sex"]);
+    }
+
+    /**
+     * Return the candidate's entity type.
+     *
+     * @return ?EntityType
```
zaliqarosli:
```suggestion
     * @return EntityType|null EntityType value object
```

php/libraries/Candidate.class.inc
driusan: I disagree. `?` for nullable is standard PHP, anyone who doesn't know it will have to learn it pretty quickly.

zaliqarosli: sure. i don't really have a strong stance on this

johnsaigle:
@zaliqarosli Thanks. I made the comparison case-insensitive. Can you see if it works now?

src/StudyEntities/Candidate/EntityType.php
```diff
@@ -0,0 +1,75 @@
+<?php declare(strict_types=1);
+/**
+ * This defines the entity type for a Candidate. It is a Value Object used
```
driusan: The first part of this sentence is confusing. It sounds like it's saying it's defining a candidate class.

src/StudyEntities/Candidate/EntityType.php
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * Ensures that EntityType is well-formed.
+     *
+     * @param string $value The value to be validated
+     *
+     * @return bool True if the value format is valid
+     */
+    private function validate(string $value): bool
```
driusan: I think this should be public static. Nothing about the validation depends on the class state, there's no reason you shouldn't be able to validate a string without instantiating it.

src/StudyEntities/Candidate/Sex.php
```diff
@@ -0,0 +1,76 @@
...
+                                 );
+
+    /**
+     * Calls validate() immediately.
+     *
+     * @param string $value The value of this EntityType.
+     *
+     * @throws \DomainException When the value is not valid
+     */
+    final public function __construct(string $value)
```
driusan: There's no reason for this to be final.

src/StudyEntities/Candidate/Sex.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * Ensures that the value is well-formed.
+     *
+     * @param string $value The value to be validated
+     *
+     * @return bool True if the value format is valid
+     */
+    private function validate(string $value): bool
```
driusan: same comment as above re: public static.

modules/dicom_archive/php/dicomarchiverow.class.inc
```diff
@@ -59,7 +59,7 @@ class DICOMArchiveRow implements \LORIS\Data\DataInstance
      *
      * @return integer The CenterID
      */
-    public function getCenterID()
+    public function getCenterID(): int
```
driusan: I don't see what this change has to do with the PR.

src/StudyEntities/Candidate/EntityType.php
johnsaigle: The parent class declares it as protected so I don't think it's possible to change this to public without changing other child classes of ValidatableIdentfier.

src/StudyEntities/Candidate/Sex.php
johnsaigle: The parent class (written by @xlecours) defines it that way.

src/StudyEntities/Candidate/EntityType.php
johnsaigle:
```suggestion
 * This defines the entity type property for a Candidate. It is a Value Object used
```

src/StudyEntities/Candidate/Sex.php
johnsaigle: Same response as above re: parent class

modules/dicom_archive/php/dicomarchiverow.class.inc
johnsaigle:
```suggestion
    public function getCenterID()
```

php/libraries/Candidate.class.inc
```diff
@@ -504,12 +507,21 @@ class Candidate
     /**
      * Return the candidate's sex
      *
-     * @return string|null Male|Female|Other String describing the sex
-     *                     and null if scanner
+     * @return ?Sex
+     */
+    function getCandidateSex(): ?Sex
+    {
+        return new Sex($this->candidateInfo["Sex"]);
```
driusan: I don't see why this should return a new one every time, and not create one in the constructor and return the same one.

php/libraries/Candidate.class.inc
```diff
@@ -504,12 +507,21 @@ class Candidate
...
+    /**
+     * Return the candidate's entity type.
+     *
+     * @return ?EntityType
      */
-    function getCandidateSex(): ?string
+    function getEntityType(): ?EntityType
     {
-        return $this->candidateInfo["Sex"];
+        return new EntityType($this->candidateInfo["EntityType"]);
```
driusan: same

src/StudyEntities/Candidate/EntityType.php
driusan: There is no parent class. This class is the root of it's own hierarchy.

src/StudyEntities/Candidate/Sex.php
driusan: This also doesn't have a parent class, and if it *did* have a parent class with a final constructor that would make this code illegal because `final` means "can't be overloaded".

src/StudyEntities/Candidate/Sex.php
driusan: Response also doesn't seems valid here.

src/StudyEntities/Candidate/EntityType.php
johnsaigle: Oops I thought i extended ValidatableIdentfier.

Do we want to move away from using the design @xlecours used in that class? It seems very good for value objects imo

src/StudyEntities/Candidate/Sex.php
johnsaigle: You're right I was mistaken

src/StudyEntities/Candidate/EntityType.php
driusan: You shouldn't be able to instantiate an invalid object, but that doesn't mean you should need an object in order to validate an input.

driusan:
This is failing Travis and has conflicts.

ridz1208:
@johnsaigle travis still failing

johnsaigle:
😢

johnsaigle:
This is failing a candidate_list filter:

```
241         $this->_filterTest($
242             self::$sex,$
243             self::$display,$
244             self::$clearFilter,$
245             "Female",$
246             '225'$
247         );$
```
I'm not sure how to test this. I assume this is expecting 225 Female candidate but my Raisinbread VM has 334. In fact none of the hard-coded values in the integration tests match what's on my VM so I'm unsure how to debug.

johnsaigle:
@kongtiaowang do you have any ideas?

cmadjar:
@pierre-p-s can you fix Travis for this PR? Thanks.

pierre-p-s:
Replaced by this PR: #6855

PR #5393, opened 2019-10-29
[ReadTheDocs] Fix bad rendering

johnsaigle:
## Brief summary of changes

**After this Read The Docs should actually work --> https://acesloris.readthedocs.io/en/latest/**

Right now the docs are building properly but the home page renders as the default RTD landing page so it looks like it's broken because nothing shows up.

e.g. if you visit the link below it looks like nothing works, but if you visit https://acesloris.readthedocs.io/en/latest/API/LorisRESTAPI_v0.0.3/ you can see that the docs are building fine except for the home page.

This makes the home page (docs/index.md) explicit so everything should render nicely. (I tested this on my local copy.)

This PR also provides a functional Edit link which leads readers back to the repo.

The key `pages` is reverted to `nav` because this is actually more appropriate for newer versions of mkdocs, it turns out.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,3 +1,3 @@
 # THIS DOCUMENTATION IS UNDER CONSTRUCTION
 
-Please visit https://github.com/aces/Loris/wiki for support.
+Please visit the [GitHub Wiki](https://github.com/aces/Loris/wiki) for support.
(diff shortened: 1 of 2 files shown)


johnsaigle:
This seems to be fine actually... it's working

PR #5396, opened 2019-10-29
[PHPCS] Make type hinting mandatory

### LINKED ISSUE #4271: add return types to data_release
Related to: https://github.com/aces/Loris/pull/4270/files/101d2d13b26d2e7f4787325692d240242ae9781f#diff-0b0d599fe0612ffeea1aa7ce6185c9c3

This class has return types on some functions and not others, should be made consistent

### LINKED ISSUE #4282: Module function isPublicModule() should return a bool
The function signature in the Module class for `isPublicModule` should include a boolean return type (`public function isPublicModule() : bool`), which wasn't done in #4234.

It should be fairly easy to update it, but updating it will require changing every Module subclass in each `modules/$modulaname/php` directory.

### LINKED ISSUE #4284: Add return types to Database class "pselect*" functions

johnsaigle:
## Brief summary of changes

PHPCS now makes type hinting mandatory for function parameters and return types. You can add them automatically based on PHPDocs using `phpcbf`. `mixed` annotations don't result in any changes.

* This is a revamp of #4064 
* Resolves #4282 
* Resolves #4271 
* Resolves #4284

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ changes in the following format: PR #1234***
 
 #### Core
 - Menus are now maintained by modules and no longer in the SQL database (PR #5839)
+- Unix user permissions have been updated which may affect access to files. New
+documentation for file permissions has been added to the README.md file (PR #5323)
 
 #### Modules 
 ##### module1
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Consult the [LORIS Wiki](https://github.com/aces/Loris/wiki/Setup) page on this 
 # Sudo privileges can be revoked once the install is completed.
 sudo useradd -U -m -G sudo -s /bin/bash lorisadmin
 # Add apache to the lorisadmin group
-sudo groupadd -a -G lorisadmin www-data
+sudo usermod -a -G lorisadmin www-data
 # Set the password for the lorisadmin account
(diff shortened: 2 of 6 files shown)


johnsaigle:
Adding needs work because this is unlikely to pass Travis for another few rounds.

johnsaigle:
This will be rebased to the "Dev" branch once our new branching system takes effect.

johnsaigle:
This is now incorporating changes from 22 but there are many phan errors related to this to fix.

ridz1208:
@johnsaigle wouldn't it be easier to tackle this in smaller chunks ?

johnsaigle:
@ridz1208 Maybe but not after I've put all this time into it 🤣 

I think PRs adding PHPCS and Phan rules are always pretty big but we can sort it out. We're not releasing anytime soon so even if there are some type errors we can get them fixed.

ridz1208:
@johnsaigle I'm more concerned about travis

johnsaigle:
Let me handle Travis 😈

PR #5009, opened 2019-08-07
[User accounts] Add project option to user accounts

ridz1208:
## Brief summary of changes
Add the functionality to affiliate users to specific projects.

At this stage affiliation with a project does not affect any of the data access, that will be added later on.

Changes in this PR were made to very closely resemble the CenterID/User interaction for consistency

#### Testing instructions (if applicable)

1. Try creating new users with projects
2. try adding/removing projects from existing users

This PR depends on the changes in #4919 for the changes in the Project table SQL. Here is the diff https://github.com/ridz1208/Loris/compare/2019_07_01_project_to_registration_project...ridz1208:2019-08_05_add_user_project

PR #5398, opened 2019-10-29
[Core] Create UserProjectMatch filter

ridz1208:
## Brief summary of changes
Creation of the userProjectMatch filter for the data framework and implementing it on the candidate_list and media modules.

With this filter you should now be able to restrict the data in  the data-table by both sites and projects.

1. assign a specific set of projects to your user and make sure the user is only able to access candidates and media files from those specific projects.

### DIFF
--- a/modules/candidate_list/test/candidate_listTest.php
+++ b/modules/candidate_list/test/candidate_listTest.php
@@ -170,10 +170,10 @@ class CandidateListTestIntegrationTest extends LorisIntegrationTestWithCandidate
         $this-> _testFilter(self::$DCCID, "1 rows", '300001');
         $this-> _testFilter(self::$DCCID, "0 rows", 'test');
         $this-> _testFilter(self::$visitLabel, "362", 'V1');
-        $this-> _testFilter(self::$visitLabel, "261", 'V2');
-        $this-> _testFilter(self::$site, "8 rows", '1');
+        $this-> _testFilter(self::$visitLabel, "223", 'V2');
+        $this-> _testFilter(self::$site, "6 rows", '1');
         $this-> _testFilter(self::$site, "168", '2');
-        $this-> _testFilter(self::$entityType, "8 rows", '1');
+        $this-> _testFilter(self::$entityType, "438", '1');
 
         // test advanced filter - sex
         // Switch to Advanced mode
@@ -182,9 +182,9 @@ class CandidateListTestIntegrationTest extends LorisIntegrationTestWithCandidate
                "return document.querySelector('$btn').click()"
            );
            //female
-           $this-> _testFilter(self::$sex, "20 rows displayed of 334", '1');
+           $this-> _testFilter(self::$sex, "20 rows displayed of 228", '1');
            // male
-           $this-> _testFilter(self::$sex, "20 rows displayed of 328", '2');
+           $this-> _testFilter(self::$sex, "20 rows displayed of 210", '2');
 
     }
     /**
--- a/modules/candidate_list/test/candidate_listTest.php
+++ b/modules/candidate_list/test/candidate_listTest.php
@@ -169,7 +169,7 @@ class CandidateListTestIntegrationTest extends LorisIntegrationTestWithCandidate
         $this-> _testFilter(self::$PSCID, "1 rows", 'MTL001');
         $this-> _testFilter(self::$DCCID, "1 rows", '300001');
         $this-> _testFilter(self::$DCCID, "0 rows", 'test');
-        $this-> _testFilter(self::$visitLabel, "374", 'V1');
+        $this-> _testFilter(self::$visitLabel, "362", 'V1');
         $this-> _testFilter(self::$visitLabel, "261", 'V2');
         $this-> _testFilter(self::$site, "8 rows", '1');
         $this-> _testFilter(self::$site, "168", '2');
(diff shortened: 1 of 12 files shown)


ridz1208:
blocked by https://github.com/aces/Loris/pull/5009

once its merged this PR should be rebased and tested

driusan:
#5009 is now merged

modules/media/php/media.class.inc
```diff
@@ -47,7 +47,8 @@ class Media extends \DataFrameworkMenu
     }
 
     /**
-     * {@inheritDoc}
+     * Tells the base class that this page's provisioner can support
+     * the UserProjectMatch filter.
```
driusan: This comment is wrong.. it's on top of the useSiteFilter function

modules/media/php/mediafile.class.inc
```diff
@@ -61,6 +61,17 @@ class MediaFile implements \LORIS\Data\DataInstance
         return $this->DBRow['centerId'];
     }
 
+    /**
+     * Returns the ProjectID for this instance, for filters such as
+     * \LORIS\Data\Filters\UserProjectMatch to match against.
+     *
+     * @return integer The CenterID
```
johnsaigle:
```suggestion
     * @return integer The ProjectID
```

src/Data/Filters/UserProjectMatch.php
```diff
@@ -0,0 +1,68 @@
...
+                if ($user->hasProject($project)) {
+                    return true;
+                }
+            }
+            return false;
+        } elseif (method_exists($resource, 'getProjectID')) {
+            $resourceProject = $resource->getProjectID();
+            if (!is_null($resourceProject)) {
+                return $user->hasProject($resourceProject);
+            }
```
johnsaigle: getProjectID() can't return null so I don't think you need this check.

src/Data/Filters/UserProjectMatch.php
```diff
@@ -0,0 +1,68 @@
...
+            // We don't know if the resource thought a null ProjectID
+            // should mean "no one can access it" or "anyone can access
+            // it", so throw an exception.
+            throw new \LorisException("getProjectID on resource returned null");
```
johnsaigle: Not needed as getProjectID can't return null

src/Data/Filters/UserProjectMatch.php
driusan: All that's been verified by the method_exists check is that there's a function named `getProjectID` on the resource

driusan: I mean the code hasn't verified anything other than a method exists with that name and the signature can be anything, so the null check is required.

To know anything for sure about the signature, you'd need an interface and an `instanceof` check instead of a method_exists check.

modules/media/php/media.class.inc
```diff
@@ -47,7 +47,8 @@ class Media extends \DataFrameworkMenu
     }
 
     /**
-     * {@inheritDoc}
+     * Tells the base class that this page's provisioner can support
+     * the UserSitetMatch filter.
```
driusan:
```suggestion
     * the UserSiteMatch filter.
```

johnsaigle:
>There was 1 failure:
1) CandidateListTestIntegrationTest::testFilters
Failed asserting that '20 rows displayed of 362. (Maximum rows per page: 20501001000500010000)' contains "374".
/app/modules/candidate_list/test/candidate_listTest.php:226
/app/modules/candidate_list/test/candidate_listTest.php:172
--

johnsaigle:
> There was 1 failure:
1) CandidateListTestIntegrationTest::testFilters
Failed asserting that '20 rows displayed of 165. (Maximum rows per page: 20501001000500010000)' contains "168".
/app/modules/candidate_list/test/candidate_listTest.php:226
/app/modules/candidate_list/test/candidate_listTest.php:175
--

ridz1208:
@johnsaigle I KNOWWW ITS THE 1000th time travis fails for the same test !!!

johnsaigle:
Hahaha okay sorry I won't update the PR anymore

modules/candidate_list/test/candidate_listTest.php
```diff
@@ -182,9 +182,9 @@ function testFilters()
                "return document.querySelector('$btn').click()"
            );
            //female
-           $this-> _testFilter(self::$sex, "20 rows displayed of 334", '1');
+           $this-> _testFilter(self::$sex, "20 rows displayed of 228", '1');
            // male
-           $this-> _testFilter(self::$sex, "20 rows displayed of 328", '2');
+           $this-> _testFilter(self::$sex, "20 rows displayed of 210", '2');
```
johnsaigle:
```suggestion
           $this-> _testFilter(self::$sex, "20 rows displayed of ", '2');
```

I don't think the maximum number of candidates matters for this test. It might help.

ridz1208: i think the intent of the test is to check the "correct" number of candidates get returned....
ur solution bypasses that BUT I'M VERY TEMPTED TO DO IT

johnsaigle: Is 210 the number of Male candidates when the Project filter is applied when you test locally? I think that's what it's trying to do

ridz1208: because it seems that travis tests the PR on the current RB dataset on the major branch, with the different PRs being merged by the time I rebase the PR to determine the numbers, another PR gets merged and modifies the raisinbread data and changes the number.

This test is particularly annoying, if any of the asserts fail, the remaining ones dont get run so if they will also fail they fail only after the first assert gets fixed and the PR is submitted to travis again. so if all the asserts are gonna fail, you'll need to run it 10 times through travis... gets what happens meanwhile, another PR gets merged and the numbers change again

PR #5400, opened 2019-10-30
[Candidate List] change visitLabel filter to dropdown, fix link issue

jesscall:
Re-issue of #5362 

### Brief summary of changes
This PR does 3 things:
1. It changes the visit label filter from a textbox to a dropdown. 
2. It allows for proper filtering of subproject and visit label. Using exactMatch within the array.
3. It addresses the issue reported in #5360 in which if a user does not have `across all sites` permission, links to the candidate profiles do not appear. However, the list of candidates returned in the table is already based on their permissions and only displays the candidates that they SHOULD have access to. 

1. Try filtering by visit label and / or subbproject
2. Remove the `across all site` permission and see if you are able to access profiles. 

* #5360
* #5359

### DIFF
--- a/modules/candidate_list/jsx/candidateListIndex.js
+++ b/modules/candidate_list/jsx/candidateListIndex.js
@@ -92,7 +92,7 @@ class CandidateListIndex extends Component {
    * @return {*} a formated table cell for a given column
    */
   formatColumn(column, cell, row) {
-    if (column === 'PSCID' && this.props.hasPermission('access_all_profiles')) {
+    if (column === 'PSCID') {
       let url = this.props.baseURL + '/' + row['DCCID'] + '/';
       return <td><a href ={url}>{cell}</a></td>;
     }
@@ -155,7 +155,8 @@ class CandidateListIndex extends Component {
         show: false,
         filter: {
           name: 'visitLabel',
-          type: 'text',
+          type: 'select',
+          options: options.visitlabel,
         },
       },
       {
--- a/modules/candidate_list/jsx/candidateListIndex.js
+++ b/modules/candidate_list/jsx/candidateListIndex.js
@@ -64,7 +64,17 @@ class CandidateListIndex extends Component {
   fetchData() {
     return fetch(this.props.dataURL, {credentials: 'same-origin'})
       .then((resp) => resp.json())
-      .then((data) => this.setState({data}))
+      .then((data) => {
+        // Convert concatenated string of subproject and visit labels to array
+        data.Data = data.Data.map((row) => {
+          // Visit label
+          row[2] = (row[2]) ? row[2].split(',') : null;
+          // Subproject
+          row[4] = (row[4]) ? row[4].split(',') : null;
+          return row;
+        });
+        this.setState({data});
+      })
       .catch((error) => {
         this.setState({error: true});
         console.error(error);
@@ -112,6 +122,13 @@ class CandidateListIndex extends Component {
           <a href={url}>{cell}</a></td>
       );
     }
+
+    if (column === 'Visit Label' || column === 'Subproject') {
+      // If user has multiple visits or subprojects, join array into string
+      let result = (cell) ? <td>{cell.join(',')}</td> : null;
+      return result;
+    }
+
     return <td>{cell}</td>;
   }
 
@@ -131,7 +148,6 @@ class CandidateListIndex extends Component {
      * XXX: Currently, the order of these fields MUST match the order of the
      * queried columns in _setupVariables() in candidate_list.class.inc
      */
-    // FIXME: if candidate in more than one subproject, filter does not work
     const options = this.state.data.fieldOptions;
     const fields = [
       {
--- a/modules/candidate_list/jsx/candidateListIndex.js
+++ b/modules/candidate_list/jsx/candidateListIndex.js
@@ -125,7 +125,7 @@ class CandidateListIndex extends Component {
 
     if (column === 'Subproject') {
       // If user has multiple subprojects, join array into string
-      let result = (cell) ? <td>{cell.join(',')}</td> : null;
+      let result = (cell) ? <td>{cell.join(', ')}</td> : null;
       return result;
     }
 
--- a/modules/candidate_list/jsx/candidateListIndex.js
+++ b/modules/candidate_list/jsx/candidateListIndex.js
@@ -123,8 +123,8 @@ class CandidateListIndex extends Component {
       );
     }
 
-    if (column === 'Visit Label' || column === 'Subproject') {
-      // If user has multiple visits or subprojects, join array into string
+    if (column === 'Subproject') {
+      // If user has multiple subprojects, join array into string
       let result = (cell) ? <td>{cell.join(',')}</td> : null;
       return result;
     }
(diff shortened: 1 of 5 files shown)


jesscall:
See @zaliqarosli 's PR #4167 which allows for filtering through arrays

driusan:
@jesscall can you

1. update the PR description, I don't think it's accurate anymore
2. verify that this is going to the right branch?

jesscall:
@driusan This PR should be going to `21.0-release` I'm just waiting for confirmation from @ridz1208 that he has pushed changes from v21.0.3 tag onto the release branch

modules/candidate_list/jsx/candidateListIndex.js
```diff
@@ -64,7 +64,17 @@ class CandidateListIndex extends Component {
   fetchData() {
     return fetch(this.props.dataURL, {credentials: 'same-origin'})
       .then((resp) => resp.json())
-      .then((data) => this.setState({data}))
+      .then((data) => {
+        // Convert concatenated string of subproject and visit labels to array
+        data.Data = data.Data.map((row) => {
```
zaliqarosli: could you do all this in `php/candidate_list.class.inc`? manipulate result of `\Utility::getVisitList()` so you get an associative array i.e. 'V1'=>'V1', 'V2'=>'V2'

i personally don't think this should be handled on the client side

modules/candidate_list/jsx/candidateListIndex.js
```diff
@@ -92,7 +102,7 @@ class CandidateListIndex extends Component {
    * @return {*} a formated table cell for a given column
    */
   formatColumn(column, cell, row) {
-    if (column === 'PSCID' && this.props.hasPermission('access_all_profiles')) {
+    if (column === 'PSCID') {
```
zaliqarosli: way to sneak this one in jess haha should this change be part of this PR? 😛

zaliqarosli: way to sneak this one in jess haha should this change be part of this PR? 😛

modules/candidate_list/jsx/candidateListIndex.js
```diff
@@ -64,7 +64,17 @@ class CandidateListIndex extends Component {
   fetchData() {
     return fetch(this.props.dataURL, {credentials: 'same-origin'})
       .then((resp) => resp.json())
-      .then((data) => this.setState({data}))
+      .then((data) => {
+        // Convert concatenated string of subproject and visit labels to array
```
zaliqarosli: could you do all this in php/candidate_list.class.inc? manipulate result of \Utility::getVisitList() so you get an associative array i.e. 'V1'=>'V1', 'V2'=>'V2'

i personally don't think this should be handled on the client side

modules/candidate_list/jsx/candidateListIndex.js
```diff
@@ -112,6 +122,13 @@ class CandidateListIndex extends Component {
           <a href={url}>{cell}</a></td>
       );
     }
+
+    if (column === 'Visit Label' || column === 'Subproject') {
```
zaliqarosli: is subproject already getting send as array?

modules/candidate_list/jsx/candidateListIndex.js
```diff
@@ -155,7 +171,8 @@ class CandidateListIndex extends Component {
         show: false,
```
zaliqarosli: visit label is currently set to not 'show' in the data table so what you added in the format column for column === 'Visit Label' won't get implement

modules/candidate_list/jsx/candidateListIndex.js
zaliqarosli: ohh nevermind, i just read that you did indeed mean to include this in the PR

modules/candidate_list/jsx/candidateListIndex.js
jesscall: I followed your implementation from `user_accounts` (#4167). 
I can make the requested change though. What do you think?

modules/candidate_list/jsx/candidateListIndex.js
jesscall: see line 73 - I'm doing the same as for visit label

modules/candidate_list/jsx/candidateListIndex.js
jesscall: oh true! good point

modules/candidate_list/jsx/candidateListIndex.js
zaliqarosli: ohh i see. i thought the array would have directly sent over but because we json'd it, i think it's still necessary here to convert it back to an array. i think this makes sense

modules/candidate_list/jsx/candidateListIndex.js
```diff
@@ -112,6 +122,13 @@ class CandidateListIndex extends Component {
           <a href={url}>{cell}</a></td>
       );
     }
+
+    if (column === 'Subproject') {
+      // If user has multiple subprojects, join array into string
+      let result = (cell) ? <td>{cell.join(',')}</td> : null;
```
zaliqarosli:
```suggestion
      let result = (cell) ? <td>{cell.join(', ')}</td> : null;
```

could we add spacing so it looks nicer? i'll approve after. testing it now

ridz1208:
@jesscall I hate to say it but... i told you so!!

each test will fail in each run

zaliqarosli:
can i ask why this is going to 21.0-release? will this PR be included in 22.0-release or major/master down the line?

ridz1208:
@zaliqarosli it was a bug found on 21 by the ccna testers. it will be included in 22 as soon as its merged

modules/candidate_list/test/candidate_listTest.php
```diff
@@ -26,12 +26,11 @@ class CandidateListTestIntegrationTest extends LorisIntegrationTestWithCandidate
     //filter location
     static $PSCID          = ".col-xs-12:nth-child(2) > .row .form-control";
     static $DCCID          = ".col-xs-12:nth-child(3) > .row .form-control";
-    static $visitLabel     = ".col-xs-12:nth-child(4) .form-control";
-    static $site           = ".col-xs-12:nth-child(5) .form-control, select";
-    static $entityType     = ".col-xs-12:nth-child(7) .form-control, select";
-    static $sex            = "#candidateList_filter > div > div > fieldset >".
-                             " div:nth-child(11) > div > div > select";
-    static $project        = ".col-xs-12:nth-child(15) .form-control, select";
+    static $visitLabel     = "select[name=\"visitLabel\"]";
```
driusan: you're not interpolating anything, why don't you just change the quoting style?

ie `static $visitLabel = 'select[name="visitLabel"]';`

ridz1208: because i copy pasted and my IDE did the escaping

jesscall:
@driusan can you re-review?

modules/candidate_list/jsx/candidateListIndex.js
driusan: I talked to Samir yesterday and he approved this change.

zaliqarosli:
has this been added to 22.0-release yet?

ridz1208:
@zaliqarosli yes

PR #5401, opened 2019-10-30
Bugfix display echo time repetition time

cmadjar:
## Brief summary of changes

The echo time in the header info section always displayed 0 for values below 1s.

Example: echo_time = 0.030 => the (int) before its value makes the echo_time becomes 0 then we multiply it by 1000 = 0
However, the value displayed in the front end should be 30ms not 0ms...

That simple bug fix fixes it.

1. Look at the echo time value on aces/bugfix (it will display 0 if echo_time < 1, 1 if 1 < echo_time < 2 ...)
2. Checkout this branch and you will see that the actual echo_time is displayed in ms

### DIFF
--- a/modules/imaging_browser/php/viewsession.class.inc
+++ b/modules/imaging_browser/php/viewsession.class.inc
@@ -221,11 +221,11 @@ class ViewSession extends \NDB_Form
             $SeriesNumber        = $FileObj->getParameter('series_number');
             $SeriesUID           = $FileObj->getParameter('series_instance_uid');
             $EchoTime            = number_format(
-                (int)$FileObj->getParameter('echo_time')*1000,
+                (float)$FileObj->getParameter('echo_time') * 1000,
                 2
             );
             $RepetitionTime      = number_format(
-                (int)$FileObj->getParameter('repetition_time') * 1000,
+                (float)$FileObj->getParameter('repetition_time') * 1000,
                 2
             );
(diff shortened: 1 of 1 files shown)


ridz1208:
is this a bug on 22 or 21 ?

cmadjar:
It is broken on 21.0 and 22.0 :). So maybe send it to 21.0 and then merge it to 22.0?

ridz1208:
please rebase on the 21.0-release branch for now

cmadjar:
@ridz1208 Done :)

modules/imaging_browser/php/viewsession.class.inc
```diff
@@ -221,11 +221,11 @@ class ViewSession extends \NDB_Form
             $SeriesNumber        = $FileObj->getParameter('series_number');
             $SeriesUID           = $FileObj->getParameter('series_instance_uid');
             $EchoTime            = number_format(
-                (int)$FileObj->getParameter('echo_time')*1000,
+                (float)$FileObj->getParameter('echo_time') * 1000,
```
driusan: This looks really familiar.. didn't we already do this?

cmadjar: the PR was sent to bug fix but got closed when bugfix had to be removed so I had to rebase it and send another PR so that it actually makes it to the code base ;)

PR #5426, opened 2019-10-30
[User Accounts] Fix password bugs: "confirm password" must match; password cannot be email

### LINKED ISSUE #5416: [user_accounts] Can set password and email to be the same
One of the password rules "The password and the email address must not be the same" isn't enforced; can set email as password with no issue

### LINKED ISSUE #5417: [user_accounts] Issues with password validation
- Passwords don't have to match and will save successfully
- When editing or creating a user's password, don't have to enter anything into Confirm Password

johnsaigle:
## Brief summary of changes

Password and Confirm Password on the front-end don't need to match on major. This check was removed by mistake during Password refactoring.

1. For both Edit User and My Preferences, try creating a new password and enter something different in the Confirm field. You should get an error message.

* Resolves #5417 
* Resolves #5416

### DIFF
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -2,7 +2,7 @@
 /**
  * The user account page
  *
- * PHP Version 5,7
+ * PHP Version 7
  *
  * @category Main
  * @package  User_Account
@@ -25,6 +25,10 @@ use \Psr\Http\Message\ResponseInterface;
  */
 class Edit_User extends \NDB_Form
 {
+    private const PASSWORD_ERROR_NO_MATCH  = 'The passwords do not match.';
+    private const PASSWORD_ERROR_NO_CHANGE = 'New and old passwords are ' .
+                                        'identical: please choose another one';
+
     /**
      * Determines whether this form is in edit or create mode.
      *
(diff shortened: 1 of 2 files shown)


modules/user_accounts/php/edit_user.class.inc
```diff
@@ -1081,19 +1088,25 @@ class Edit_User extends \NDB_Form
                     = 'Please specify password or click Generate new password';
             }
         }
+        // Ensure that the password and confirm password fields match.
+        // TODO This validation should be done on the front-end instead.
```
PapillonMcGill:
```suggestion
        // TODO This validation should also be done on the front-end.
```
Never rely to front-end check!

modules/user_accounts/php/my_preferences.class.inc
```diff
@@ -386,20 +390,32 @@ class My_Preferences extends \NDB_Form
         $DB     = \Database::singleton();
         $errors = array();
 
+        // NOTE The 'Password_hash' key actually represents a plaintext password
+        $plaintext = $values['Password_hash'];
+
+        // Ensure that the password and confirm password fields match.
+        // TODO This validation should be done on the front-end instead.
```
PapillonMcGill:
```suggestion
        // TODO This validation should also be done on the front-end.
```

driusan:
This should go to the release branch.. we want it fixed for 22.0, not 23.0, don't we?

johnsaigle:
I've rebased it to master.

modules/user_accounts/php/edit_user.class.inc
johnsaigle: I don't think it's really an issue in this case. It's just a double-check so that someone doesn't make a typo while entering their password.

driusan:
Also fixes #5439

johnsaigle:
I tested this locally and it worked but it wasn't manually tested. Hopefully any potential errors will come up during the next testing round

PR #5444, opened 2019-10-31
[imaging_browser] fixed query for the imaging browser

cmadjar:
## Brief summary of changes

This fixes the query for the imaging browser by taking the ProjectID from the session table instead of the candidate table.

#### Testing instructions

1. Test that loading imaging browser on 22.0 fails
2. Test that after running the SQL patches it works again

#### Special note: 
The following patch (`SQL/New_patches//2019-07-01_add_projects_to_sessions.sql`) should be modified so that the `ProjectID` is populated correctly in the `session` table when studies have `scanner` candidates that are not associated to projects.
The update statement in that file should have been:
```
UPDATE session s JOIN candidate c ON s.CandID=c.CandID SET ProjectID=c.RegistrationProjectID WHERE PSCID != 'scanner';
```
@driusan @ridz1208 What should I do about this one? Should I create a new patch or just modify the one with the update since this is from a patch for that release? Let me know.

### DIFF
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -178,7 +178,7 @@ CREATE TABLE `session` (
   `ID` int(10) unsigned NOT NULL AUTO_INCREMENT,
   `CandID` int(6) NOT NULL DEFAULT '0',
   `CenterID` integer unsigned NOT NULL,
-  `ProjectID` int(10) unsigned NOT NULL,
+  `ProjectID` int(10) unsigned DEFAULT NULL,
   `VisitNo` smallint(5) unsigned DEFAULT NULL,
   `Visit_label` varchar(255) NOT NULL,
   `SubprojectID` int(10) unsigned DEFAULT NULL,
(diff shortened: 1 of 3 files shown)


ridz1208:
@cmadjar modify the same patch
once its merged I'll make the release patch

driusan:
Yeah, modify the patch since it hasn't been released.

cmadjar:
Patch updated.

Note: modified the column RegistrationProjectID to have default NULL as well since this is how it is defined in the default schema and NULL is needed for the scanner candidates for that field

zaliqarosli:
@driusan @kongtiaowang @cmadjar has this been pulled on the testing vm yet?

kongtiaowang:
@zaliqarosli it has been updated on testing VM.

kongtiaowang:
This PR failed on testing VM.

zaliqarosli:
:(

zaliqarosli:
@kongtiaowang you ran the patch?

kongtiaowang:
@zaliqarosli I ran the 0000-00-00-schema.sql.

cmadjar:
@zaliqarosli @kongtiaowang should be fixed now that #5528 has been merged in 22.0-release :). It was a separate bug that I did not catch cause I still had the `mri_acquisition_dates` table that was deleted from the database as useless.

Issue #5472, opened 2019-10-31
[Install] Fix error reporting when schema sourcing fails. 


ridz1208:
When running the installdb.php step of the install and a patch from the 0000-** series fails, no error is being reported to the front end, only a warning in the error log.

cmadjar:
@maltheism we assigned you the install issues for LORIS Monday. This is one of them. If you could look at them, that would be awesome!

We know that it will be 8 issues but I have a feeling they would all be related somehow and since you played with the install, you seemed like the best person to tackle them. Let us know if you have any issues.

maltheism:
@cmadjar @ridz1208 I'm not getting any output in my error log file when running the installdb.php on the 22.0-release branch.

driusan:
@maltheism the error in the SQL file was fixed, so there's currently no error.. but if you modify something in the SQL schema so that it's not valid and try installdb.php you'll see that the error handling is broken

maltheism:
@driusan thanks for the explanation. I'll investigate that!

PR #5497, opened 2019-11-01
[installdb.php] Fixes sourceSchema not reporting error

maltheism:
## Brief summary of changes

The _sourceSchemaFile function wasn't logging potential errors when running the installdb.php in the browser.

Note: Let me know if you think the current solution could be done differently.

1. Rerun the installdb.php before completing and the error should be logged.

* https://github.com/aces/Loris/issues/5472


ridz1208:
@maltheism I think the idea was to also get an error on the front end saying the database was not sourced correctly.

there should be something on the UI

maltheism:
thanks @johnsaigle for the help I'll use that

maltheism:
The error now shows on the frontend with @johnsaigle's addition.
example

Issue #5439, opened 2019-10-31
[user_accounts] Can set new password to be the same as current password


leighevans:
Test plan: 
15. Check that when editing a user account it is not possible to set the password to its actual value (i.e. it needs to change).

There no currently no validation on this; you can set and save a new password to be the same as the current one

johnsaigle:
This was fixed in https://github.com/aces/Loris/pull/5370 but the PR got merged into `major`... not sure what the test VM is based off of right now.

johnsaigle:
Also @leighevans did this occur in Edit User or My Preferences?

driusan:
According to @johnsaigle this was fixed by #5426

johnsaigle:
I gave @driusan bad information. This issue is actually addressed by https://github.com/aces/Loris/pull/5514/files

driusan:
According to @johnsaigle this was fixed by #5514

PR #5514, opened 2019-11-04
[Testing] Add unit tests for User::passwordChanged

johnsaigle:
## Brief summary of changes

Adds unit tests to the passwordChanged function. 

### Testing

While this adds a unit test, the front-end should still be tested. 
* Item 15 in the [test plan](https://github.com/aces/Loris/blob/major/modules/user_accounts/test/TestPlan.md).

* #5439 reported that a passwordChanged() was not returning a false value, i.e. a user was able to keep the same password. I'm adding this function to prevent this kind of issue going forward.


driusan:
@johnsaigle conflicts

Issue #5467, opened 2019-10-31
[BVL Feedback] Issues with display and closing of threads after hitting 'Close' button


SantiagoTG:
Once a thread has been initiated, the "Open/Close" button will display as open, and a new thread will appear for the visit. If you "Close" the thread, it will display as closed, while not actually being closed. 
Additionally, if you try to change the "closed" thread to "Open" again, it will continue to display as closed. Thus, closing a thread does not really close it, instead it only changes the display of the "Open/Close" button. 
Refreshing the page or opening a new thread will then change all threads displayed to "Open". The "Open threads summary" will always display all threads as open.

cmadjar:
@xlecours can you please look into this one? Thanks!

driusan:
Fixed by above PR

PR #5522, opened 2019-11-04
Fix 5467 bvl feedback thread status update

xlecours:
FIXES https://github.com/aces/Loris/issues/5467

It fixes the status update and adds various error message and HTTP status codes.
While the error code are still not handled by the React app, the update now works.

It is now using `prepare` and `execute` because it needs to know how many rows got updated. Specially when zero row have been updated. This information can be retrieved from the PDO statement (returned by `prepare`) with `rowCount()`

### DIFF
--- a/modules/bvl_feedback/ajax/close_bvl_feedback_thread.php
+++ b/modules/bvl_feedback/ajax/close_bvl_feedback_thread.php
@@ -21,7 +21,7 @@ try {
     $candid = new \LORIS\StudyEntities\Candidate\CandID($_POST['candID'] ?? '');
 } catch (\DomainException $e) {
     header("HTTP/1.1 400 Bad Request");
-    header("Content-Type: aaplication/json");
+    header("Content-Type: application/json");
     print json_encode(array('error' => 'invalid candID'));
     exit;
 }
--- a/modules/bvl_feedback/ajax/close_bvl_feedback_thread.php
+++ b/modules/bvl_feedback/ajax/close_bvl_feedback_thread.php
@@ -17,11 +17,47 @@ require "bvl_panel_ajax.php";
 $user     =& User::singleton();
 $username = $user->getUsername();
 
-if (isset($_POST['feedbackID']) && isset($_POST['candID'])) {
-    $feedbackThread =& NDB_BVL_Feedback::Singleton($username, $_POST['candID']);
-    $feedbackThread->closeThread($_POST['feedbackID']);
+try {
+    $candid = new \LORIS\StudyEntities\Candidate\CandID($_POST['candID'] ?? '');
+} catch (\DomainException $e) {
+    header("HTTP/1.1 400 Bad Request");
+    header("Content-Type: aaplication/json");
+    print json_encode(array('error' => 'invalid candID'));
+    exit;
 }
 
-exit();
+if (!isset($_POST['feedbackID'])) {
+    header("HTTP/1.1 400 Bad Request");
+    header("Content-Type: aaplication/json");
+    print json_encode(array('error' => 'Missing FeedbackID'));
+    exit;
+}
+
+$feedbackid = $_POST['feedbackID'];
+
+try {
+    $feedbackThread   =& NDB_BVL_Feedback::Singleton($username, $candid);
+    $closethreadcount = $feedbackThread->closeThread($feedbackid);
+} catch (\Exception $e) {
+    error_log($e->getMessage());
+    header("HTTP/1.1 404 Not Found");
+    header("Content-Type: aaplication/json");
+    print json_encode(
+        array('error' => 'The requested feedback thread can`t be found')
+    );
+    exit;
+}
+
+if ($closethreadcount === 0) {
+    header("HTTP/1.1 500 Internal Server Error");
+    header("Content-Type: aaplication/json");
+    print json_encode(
+        array('error' => 'No feedback thread updated')
+    );
+    exit;
+}
+
+header("HTTP/1.1 204 No Content");
+exit;
 
 
--- a/modules/bvl_feedback/ajax/close_bvl_feedback_thread.php
+++ b/modules/bvl_feedback/ajax/close_bvl_feedback_thread.php
@@ -39,7 +39,7 @@ try {
     $feedbackThread =& NDB_BVL_Feedback::Singleton($username, $candid);
 
     // FIXME This allows to close any thread as long as the feedbackid
-    //       exists. It do not have to be related to this feedbackthreah.
+    //       exists. It does not have to be related to this feedbackthread.
     $closethreadcount = $feedbackThread->closeThread((int) $feedbackid);
 } catch (\Exception $e) {
     error_log($e->getMessage());
--- a/modules/bvl_feedback/ajax/close_bvl_feedback_thread.php
+++ b/modules/bvl_feedback/ajax/close_bvl_feedback_thread.php
@@ -28,7 +28,7 @@ try {
 
 if (!isset($_POST['feedbackID'])) {
     header("HTTP/1.1 400 Bad Request");
-    header("Content-Type: aaplication/json");
+    header("Content-Type: application/json");
     print json_encode(array('error' => 'Missing FeedbackID'));
     exit;
 }
@@ -36,8 +36,11 @@ if (!isset($_POST['feedbackID'])) {
 $feedbackid = $_POST['feedbackID'];
 
 try {
-    $feedbackThread   =& NDB_BVL_Feedback::Singleton($username, $candid);
-    $closethreadcount = $feedbackThread->closeThread($feedbackid);
+    $feedbackThread =& NDB_BVL_Feedback::Singleton($username, $candid);
+
+    // FIXME This allows to close any thread as long as the feedbackid
+    //       exists. It do not have to be related to this feedbackthreah.
+    $closethreadcount = $feedbackThread->closeThread((int) $feedbackid);
 } catch (\Exception $e) {
     error_log($e->getMessage());
     header("HTTP/1.1 404 Not Found");
--- a/modules/bvl_feedback/ajax/close_bvl_feedback_thread.php
+++ b/modules/bvl_feedback/ajax/close_bvl_feedback_thread.php
@@ -41,7 +41,7 @@ try {
 } catch (\Exception $e) {
     error_log($e->getMessage());
     header("HTTP/1.1 404 Not Found");
-    header("Content-Type: aaplication/json");
+    header("Content-Type: application/json");
     print json_encode(
         array('error' => 'The requested feedback thread can`t be found')
     );
@@ -50,7 +50,7 @@ try {
 
 if ($closethreadcount === 0) {
     header("HTTP/1.1 500 Internal Server Error");
-    header("Content-Type: aaplication/json");
+    header("Content-Type: application/json");
     print json_encode(
         array('error' => 'No feedback thread updated')
     );
--- a/modules/bvl_feedback/ajax/open_bvl_feedback_thread.php
+++ b/modules/bvl_feedback/ajax/open_bvl_feedback_thread.php
@@ -27,7 +27,7 @@ try {
 
 if ($openedthreadcount === 0) {
     header("HTTP/1.1 500 Internal Server Error");
-    header("Content-Type: aaplication/json");
+    header("Content-Type: application/json");
     print json_encode(
         array('error' => 'No feedback thread updated')
     );
--- a/modules/bvl_feedback/ajax/open_bvl_feedback_thread.php
+++ b/modules/bvl_feedback/ajax/open_bvl_feedback_thread.php
@@ -13,8 +13,27 @@
 ini_set('default_charset', 'utf-8');
 require_once "bvl_panel_ajax.php";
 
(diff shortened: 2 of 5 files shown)


modules/bvl_feedback/ajax/close_bvl_feedback_thread.php
```diff
@@ -17,11 +17,47 @@
...
 $username = $user->getUsername();
 
-if (isset($_POST['feedbackID']) && isset($_POST['candID'])) {
-    $feedbackThread =& NDB_BVL_Feedback::Singleton($username, $_POST['candID']);
-    $feedbackThread->closeThread($_POST['feedbackID']);
+try {
+    $candid = new \LORIS\StudyEntities\Candidate\CandID($_POST['candID'] ?? '');
+} catch (\DomainException $e) {
+    header("HTTP/1.1 400 Bad Request");
+    header("Content-Type: aaplication/json");
```
driusan:
```suggestion
    header("Content-Type: application/json");
```

modules/bvl_feedback/ajax/open_bvl_feedback_thread.php
```diff
@@ -13,8 +13,27 @@
...
 require_once "bvl_panel_ajax.php";
 
-if (isset($_POST['feedbackID']) && isset($_POST['candID'])) {
-    $feedbackThread->openThread($_POST['feedbackID']);
+try {
+    $openedthreadcount = $feedbackThread->openThread($_POST['feedbackID']);
+} catch (\Exception $e) {
+    error_log($e->getMessage());
+    header("HTTP/1.1 404 Not Found");
+    header("Content-Type: aaplication/json");
```
driusan:
```suggestion
    header("Content-Type: application/json");
```

modules/bvl_feedback/ajax/close_bvl_feedback_thread.php
```diff
@@ -17,11 +17,47 @@
...
+
+$feedbackid = $_POST['feedbackID'];
+
+try {
+    $feedbackThread   =& NDB_BVL_Feedback::Singleton($username, $candid);
+    $closethreadcount = $feedbackThread->closeThread($feedbackid);
+} catch (\Exception $e) {
+    error_log($e->getMessage());
+    header("HTTP/1.1 404 Not Found");
+    header("Content-Type: aaplication/json");
```
johnsaigle:
```suggestion
    header("Content-Type: application/json");
```

modules/bvl_feedback/ajax/close_bvl_feedback_thread.php
```diff
@@ -17,11 +17,47 @@
...
+    print json_encode(
+        array('error' => 'The requested feedback thread can`t be found')
+    );
+    exit;
+}
+
+if ($closethreadcount === 0) {
+    header("HTTP/1.1 500 Internal Server Error");
+    header("Content-Type: aaplication/json");
```
johnsaigle:
```suggestion
    header("Content-Type: application/json");
```

modules/bvl_feedback/ajax/open_bvl_feedback_thread.php
```diff
@@ -13,8 +13,27 @@
 ini_set('default_charset', 'utf-8');
 require_once "bvl_panel_ajax.php";
 
-if (isset($_POST['feedbackID']) && isset($_POST['candID'])) {
-    $feedbackThread->openThread($_POST['feedbackID']);
+try {
+    $openedthreadcount = $feedbackThread->openThread($_POST['feedbackID']);
```
johnsaigle: It would be good to have an isset check for feedbackID, as you did in the other file

modules/bvl_feedback/ajax/open_bvl_feedback_thread.php
```diff
@@ -13,8 +13,27 @@
...
+    print json_encode(
+        array('error' => 'The requested feedback thread can`t be found')
+    );
+    exit;
 }
 
-exit();
+if ($openedthreadcount === 0) {
+    header("HTTP/1.1 500 Internal Server Error");
+    header("Content-Type: aaplication/json");
```
johnsaigle:
```suggestion
    header("Content-Type: application/json");
```

php/libraries/NDB_BVL_Feedback.class.inc
```diff
@@ -941,58 +941,52 @@ class NDB_BVL_Feedback
...
         } else {
             // at least one of the above has to be true
             throw new Exception(
                 "Error: No threads selected were selected to be closed"
             );
         }
-
         // close the threads - update feedback_bvl_thread table
-        $db->update('feedback_bvl_thread', $setArray, $whereArray);
+        $stmt = $db->prepare(
```
johnsaigle: Why not use `update` here instead? Just wondering.

php/libraries/NDB_BVL_Feedback.class.inc
```diff
@@ -1002,30 +996,32 @@ class NDB_BVL_Feedback
      *
      * @author Evan McIlroy <evanmcilroy@gmail.com>
      *
-     * @return void
+     * @return int The number of updated threads
      */
     function openThread($feedbackID)
     {
+        if (empty($feedbackID)) {
```
johnsaigle: `int`s can't be empty. Might be nice to use type hints here instead.

php/libraries/NDB_BVL_Feedback.class.inc
```diff
@@ -1002,30 +996,32 @@ class NDB_BVL_Feedback
...
+            throw new Exception(
```
johnsaigle:
```suggestion
            throw new InvalidArgumentException(
```

Might be nice to be more specific if you leave this here and don't use type hints

php/libraries/NDB_BVL_Feedback.class.inc
xlecours: @johnsaigle because I need the `PDOStatement` object to call `rowCounts` on.

php/libraries/NDB_BVL_Feedback.class.inc
driusan: 0 is empty for an int.

johnsaigle: Oops

johnsaigle: I still think type hints would be nice.

Alternatively
```suggestion
        if (! \Utility::valueIsPositiveInteger($feedbackID))) {
```

or

```suggestion
        if ($feedbackID < 1) {
```

I think these are nicer because it makes it clear we're dealing with a number. I won't block the PR over it though

modules/bvl_feedback/ajax/close_bvl_feedback_thread.php
```diff
@@ -17,11 +17,50 @@
...
+    exit;
+}
+
+$feedbackid = $_POST['feedbackID'];
+
+try {
+    $feedbackThread =& NDB_BVL_Feedback::Singleton($username, $candid);
+
+    // FIXME This allows to close any thread as long as the feedbackid
+    //       exists. It do not have to be related to this feedbackthreah.
```
johnsaigle:
```suggestion
    //       exists. It does not have to be related to this feedbackthread.
```

modules/bvl_feedback/ajax/open_bvl_feedback_thread.php
```diff
@@ -13,8 +13,31 @@
...
+    header("HTTP/1.1 400 Bad Request");
+    header("Content-Type: application/json");
+    print json_encode(
+        array('error' => 'feedbackId missing or invalid')
+    );
+    exit;
 }
 
-exit();
+// This is realy powerfull; it allows to reopen any feedbackthread as long as the
```
johnsaigle:
```suggestion
// This is really powerful; it allows to reopen any feedbackthread as long as the
```

modules/bvl_feedback/ajax/open_bvl_feedback_thread.php
```diff
@@ -13,8 +13,31 @@
 ini_set('default_charset', 'utf-8');
 require_once "bvl_panel_ajax.php";
 
-if (isset($_POST['feedbackID']) && isset($_POST['candID'])) {
-    $feedbackThread->openThread($_POST['feedbackID']);
+$feedbackid = $_POST['feedbackID'] ?? '';
+
+if (empty($feedbackid) || !is_numeric($feedbackID)) {
```
johnsaigle:
```suggestion
if (! \Utility::valueIsPositiveInteger) {
```

This will also make sure that it's not a decimal. Not a big deal though

php/libraries/NDB_BVL_Feedback.class.inc
```diff
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
```
johnsaigle: ❤️

xlecours: It might break things though...

xlecours: My Faith is in Travis.

php/libraries/NDB_BVL_Feedback.class.inc
```diff
@@ -941,91 +941,81 @@ class NDB_BVL_Feedback
...
-        );
+        $whereStr = 'Public = \'Y\' AND Active = \'Y\' ';
         if (!empty($feedbackID)) {
             // only one thread to be closed
-            $whereArray = array_merge(
-                $whereArray,
-                array("FeedbackID" => $feedbackID)
-            );
+            $whereStr .= 'AND FeedbackID = :v_feedbackid ';
+            $params    = array('v_feedbackid' => $feedbackID);
```
johnsaigle: What happens if `$feedbackID` is null?

xlecours: I think it is meant to close more thant one thread.  i.e.: All threads related to either the commentid, the sessionid of the candid.

php/libraries/NDB_BVL_Feedback.class.inc
driusan: Your faith might break things, though.

php/libraries/NDB_BVL_Feedback.class.inc
johnsaigle: It might be good to note that in the function description. As you said, it's quite powerful.

xlecours: But I can't find any place where it is called without the feedbackid.

johnsaigle: Maybe it should just be `int` then with no default param? Might as well eliminate that potential for bugs if it's not used anyway.

php/libraries/NDB_BVL_Feedback.class.inc
xlecours: Well, after a quick test, instrument_list is broken
```
PHP Fatal error:  Uncaught TypeError: Argument 3 passed to NDB_BVL_Feedback::setup() must be of the type string or null, int given, called in /var/www/loris_1/php/libraries/NDB_BVL_Feedback.class.inc on line 102
```
I will remove the strict_types

johnsaigle: 😿

php/libraries/NDB_BVL_Feedback.class.inc
```diff
@@ -941,91 +941,81 @@ class NDB_BVL_Feedback
      *
      * @param integer $feedbackID The ID of the feedback thread being closed
      *
-     * @return void
-     *  FIXME: This shares a lot of code from activateThread, it should be
-     *         refactored
-     * FIXME: This needs to properly check for sites.
+     * @return int The number of updated threads
```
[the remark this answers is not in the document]
xlecours: PDOStatement::rowCount always return a number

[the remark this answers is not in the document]
johnsaigle: You're right, I commented on the wrong line. 942, the param line, needs a null annotation

php/libraries/NDB_BVL_Feedback.class.inc
xlecours: It won't be done in this PR.

php/libraries/NDB_BVL_Feedback.class.inc
```diff
@@ -941,91 +941,81 @@ class NDB_BVL_Feedback
      *
      * @param integer $feedbackID The ID of the feedback thread being closed
```
johnsaigle: This line needs `?int`

php/libraries/NDB_BVL_Feedback.class.inc
```diff
@@ -939,93 +939,83 @@ class NDB_BVL_Feedback
...
-    function openThread($feedbackID)
+    function openThread(int $feedbackID): int
     {
         $db =& Database::singleton();
 
-        $setArray = array(
-            "Status" => 'opened',
-            "UserID" => $this->_username,
-        );
+        $whereStr = 'Public = \'Y\' AND Active = \'Y\' ';
```
driusan: I would just use a different type of quotation mark, but okay..

Issue #5506, opened 2019-11-04
[Configuration] multiple entry fields broken


ridz1208:
> 5. Go through each of the fields in the configuration module that have the 
> "Add field" button are configuration values that allow multiple entries. For each of 
> these fields:
>    - try clicking the "Add field" button to see that this adds a new field to enter 
>    data. Enter data and press save at the bottom of the page. Refresh the page, check
>    that the data saved
>    - try deleting a field with the 'X' button. Press save at the bottom of the page. 
>    Refresh the page and check that the field was in fact deleted

for both deletion of an entry or addition. I get in the console:

```
jquery-1.11.0.min.js:4 POST https://test-loris-dev.loris.ca/configuration/ajax/process.php 400 (Bad Request)
```

driusan:
This should be resolved by the above PR, according to the people who tested and reviewed it.

PR #5524, opened 2019-11-04
[Configuration] multiple entry fields broken - bug fix

kongtiaowang:
multiple entry fields broken
https://github.com/aces/Loris/issues/5506

### DIFF
--- a/modules/configuration/ajax/process.php
+++ b/modules/configuration/ajax/process.php
@@ -77,10 +77,12 @@ foreach ($_POST as $key => $value) {
         // This is different from the above is_numeric case; this makes use of
         // Config.ID, not Config.ConfigID (which is a FK to ConfigSettings.ID).
         // The Config table is the one that will be modified here.
-        $keySplit         = explode("-", $key); // e.g. 'add-10' or 'remove-49'
+        $keySplit         = explode("-", $key); // e.g. 'add-17-1' or 'remove'
         $action           = $keySplit[0];
-        $ConfigSettingsID = $keySplit[1];
-        assert(count($keySplit) == 2);
+        $ConfigSettingsID = $keySplit[1];        
+        $valueSplit       = explode("-", $value); // e.g. "remove-74"
+        $removeID         = $valueSplit[1]; 
+        //assert(count($keySplit) == 2);
         if ($action == 'add') {
             // This branch adds a new entry to the Config table.
             if ($value === "") {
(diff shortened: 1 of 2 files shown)


ridz1208:
@kongtiaowang the issue is still there when you click on the red X next to the field

kongtiaowang:
@ridz1208 Could u test it again?

modules/configuration/ajax/process.php
```diff
@@ -77,18 +77,20 @@
         // This is different from the above is_numeric case; this makes use of
         // Config.ID, not Config.ConfigID (which is a FK to ConfigSettings.ID).
         // The Config table is the one that will be modified here.
-        $keySplit         = explode("-", $key); // e.g. 'add-10' or 'remove-49'
+        $keySplit         = explode("-", $key); // e.g. 'add-17-1' or 'remove'
```
ridz1208: the example on this line doesnt seem to match the behaviour below.

I think it should be `add-17' or 'remove-1'`

ridz1208:
works fine, one comment is a bit confusing. should be fixed

modules/configuration/js/configuration_helper.js
```diff
@@ -81,13 +81,12 @@ $(function () {
             data: form,
             success: function () {
                 var html = "<label>Submitted</label>";
-                $(html).hide().appendTo('.submit-area').fadeIn(500).delay(1000).fadeOut(500)
-                $('input[type="reset"]').attr('disabled','disabled');
+                $(html).hide().appendTo('.submit-area').fadeIn(500).delay(1000).fadeOut(500);
+                location.reload();
```
driusan: This shouldn't reload the page, the data is maintained by ajax.

modules/configuration/ajax/process.php
ridz1208: nevermind, all good

ridz1208:
@kongtiaowang please address dave's comment and I will approve after

modules/configuration/js/configuration_helper.js
kongtiaowang: when user click the remove (it will call this ajax) or add instruments, the user also changes the other data at the same time, it will break the reset button. That is why I put reload function.

driusan: That sounds like it's just bandaging over a much deeper problem, but okay, for the interest of getting this release done that's fine

Issue #5473, opened 2019-10-31
[brainbrowser] No permissions for access restriction


ridz1208:
if you are given a link like the one below, you can access the scans regardless of your permissions

[LINK](https://test-loris-dev.loris.ca/brainbrowser/?minc_id=[134,135]&overlay=true) for example scans

The readme.md file says that the imaging browser's permissions control access to brainbrowser, it's not the case

PR #5530, opened 2019-11-04
[brainbrowser] Add permissions to view the module

cmadjar:
#### Brief summary of changes

In the test plan, we mention that only users with imaging browser permissions can view the scans in BB. Added the check for user permission before loading the module.

1. Try out the following URL on RB dataset <your_VM>/brainbrowser/?minc_id=[134] with a user that do not have imaging browser permission. This should lead you to a "No permission" with a redirect link to the dashboard.

* #5473

### DIFF
--- a/modules/brainbrowser/php/brainbrowser.class.inc
+++ b/modules/brainbrowser/php/brainbrowser.class.inc
@@ -37,16 +37,16 @@ class Brainbrowser extends \NDB_Page
         * part of a study site and are permitted to view their own site.
          */
         return $user->hasAnyPermission(
-                array(
-                    'imaging_browser_view_allsites',
-                    'imaging_browser_phantom_allsites',
-                    'imaging_browser_phantom_ownsite',
-                )
+            array(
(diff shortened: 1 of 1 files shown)


kongtiaowang:
PHPCS

modules/brainbrowser/php/brainbrowser.class.inc
```diff
@@ -24,6 +24,31 @@ namespace LORIS\brainbrowser;
  */
 class Brainbrowser extends \NDB_Page
 {
+    /**
+     * Determine whether the user has permission to view this page
+     *
+     * @param \User $user The user whose access is being checked
+     *
+     * @return bool whether the user hass access
```
johnsaigle:
```suggestion
     * @return bool whether the user has access
```

johnsaigle: 🐍

cmadjar: lol! Thanks! Fixed in the next commit.

Issue #5526, opened 2019-11-04
[imaging_uploader] Forbidden file types can be uploaded (e.g. jpg, txt)


kongtiaowang:
According to the test plan, imaging_uploader only can upload .tgz and .gz files.

driusan:
Fixed by above

PR #5536, opened 2019-11-05
[imaging_uploader] fix forbidden file types that could be uploaded

cmadjar:
## Brief summary of changes

This fixes the issue of any file type being able to be uploaded when the 'Phantom Scans' was set to 'Yes'. 

1. make sure you cannot upload file types different than `.zip`, `.tgz` or `.tar.gz` both when 'Phantom Scans' is set to 'Yes' and 'No'

* #5526

### DIFF
--- a/modules/imaging_uploader/jsx/UploadForm.js
+++ b/modules/imaging_uploader/jsx/UploadForm.js
@@ -82,7 +82,7 @@ class UploadForm extends Component {
 
     const fileName = data.mriFile.name;
     // Make sure file is of type .zip|.tgz|.tar.gz format
-    const properExt = new RegExp('\.(zip|tgz|tar.gz)$');
+    const properExt = new RegExp('\.(zip|tgz|tar\.gz)$');
     if (!fileName.match(properExt)) {
       swal({
         title: 'Invalid extension for the uploaded file!',
--- a/modules/imaging_uploader/jsx/UploadForm.js
+++ b/modules/imaging_uploader/jsx/UploadForm.js
@@ -82,7 +82,7 @@ class UploadForm extends Component {
 
     const fileName = data.mriFile.name;
     // Make sure file is of type .zip|.tgz|.tar.gz format
-    const properExt = new RegExp('.(zip|tgz|tar.gz)$');
+    const properExt = new RegExp('\.(zip|tgz|tar.gz)$');
     if (!fileName.match(properExt)) {
       swal({
         title: 'Invalid extension for the uploaded file!',
@@ -92,7 +92,7 @@ class UploadForm extends Component {
       });
 
       let errorMessage = {
-        mriFile: 'The file ' + fileName + ' is not of type .tgz, .tar.gz or .zip.',
+        mriFile: 'The file ' + fileName + ' must be of type .tgz, .tar.gz or .zip.',
         candID: undefined,
         pSCID: undefined,
         visitLabel: undefined,
--- a/modules/imaging_uploader/jsx/UploadForm.js
+++ b/modules/imaging_uploader/jsx/UploadForm.js
@@ -81,6 +81,34 @@ class UploadForm extends Component {
     }
 
     const fileName = data.mriFile.name;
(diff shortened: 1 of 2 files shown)


cmadjar:
@kongtiaowang could you review this PR? It should fix the issue you reported :).

johnsaigle:
@cmadjar If possible could you add back-end validation here too? This should give a 400 response when the files are submitted to the back-end.

A person will still be able to upload other file types by submitting a raw POST request and ignoring the front-end.

cmadjar:
@johnsaigle the check of the file type is already done in imaging_upload.class.inc on line 381 (when the file gets validated). The `isValid` function is called by the `_saveFile` function which is called by the `_process` function so I guess everything is fine, no?

gdevenyi:
Beyond checking filenames, I think validation should be done by testing the file magic numbers before doing any further processing, either calling "file" and parsing the output, or perhaps there's a module for PHP.

cmadjar:
@gdevenyi it looks like this is already done after having checked the filename. Line 397 of `imaging_uploader.class.inc` calls the function `isCompressed` on the file which is checking the mime_content_type of the file.

It does not look like this is checked on the JS side though. Not sure if this is an issue...

johnsaigle:
@cmadjar My understanding from the ticket was that a file upload of forbidden types successfully occurred. That would mean that the check on the back-end is not called or is broken.

I haven't done any testing for this issue though.

cmadjar:
Ha! Thank you @johnsaigle for the reminder!! Indeed, on the PHP side, the checking was also done only on the non-phantom scans... Moved the file type check out of the `if (isPhantom=='N')` statement since it should apply to all uploaded files :).

johnsaigle:
I tested the front-end checks but not the back-end.

modules/imaging_uploader/jsx/UploadForm.js
```diff
@@ -81,6 +81,34 @@ class UploadForm extends Component {
     }
 
     const fileName = data.mriFile.name;
+    // Make sure file is of type .zip|.tgz|.tar.gz format
+    const properExt = new RegExp('.(zip|tgz|tar.gz)$');
```
driusan: The "."s in this regex aren't escaped, "." means "any character" in a regular expression not a literal "."

johnsaigle: Is this fixed by a backslash?

```suggestion
    const properExt = new RegExp('\.(zip|tgz|tar.gz)$');
```

I'm not sure good with JS

modules/imaging_uploader/jsx/UploadForm.js
```diff
@@ -81,6 +81,34 @@ class UploadForm extends Component {
...
+    if (!fileName.match(properExt)) {
+      swal({
+        title: 'Invalid extension for the uploaded file!',
+        text: 'Filename extension does not match .zip, .tgz or .tar.gz ',
+        type: 'error',
+        confirmButtonText: 'OK',
+      });
+
+      let errorMessage = {
+        mriFile: 'The file ' + fileName + ' is not of type .tgz, .tar.gz or .zip.',
```
johnsaigle:
```suggestion
        mriFile: 'The file ' + fileName + ' must be of type .tgz, .tar.gz or .zip.',
```

Not a big deal but I think this is a bit more instructive

modules/imaging_uploader/jsx/UploadForm.js
driusan: yes, but you're missing the one in tar.gz and I think it needs to be retested

driusan: @cmadjar

cmadjar: right. I saw that and meant to correct it but I guess I was distracted ;). Done in the last commit

driusan:
@johnsaigle can you retest just in case there's some interaction with string and regex escaping?

johnsaigle:
front-end check is working

PR #5538, opened 2019-11-05
[User Accounts] Add integration tests for password change logic

### LINKED ISSUE #5416: [user_accounts] Can set password and email to be the same

johnsaigle:
## Brief summary of changes

Integration tests now verify:
* "Password" and "Confirm password" must match
* Password must not match email
* Password must be changed (i.e. submitting the same password as you had before gives an error)

This is done for both "My Preferences" and "Edit User"

This PR also does some refactoring of the file to make it easier to add to going forward.

### Documentation changed

* User Accounts test plan

## Related

* Resolves #5416

### DIFF
--- a/modules/user_accounts/test/TestPlan.md
+++ b/modules/user_accounts/test/TestPlan.md
@@ -28,14 +28,14 @@ When creating or editing a user: (subtest: edit_user)
       - First name,
       - Last name,
       - Email.
-10. If password and confirmed password do not match, an error should be displayed.
+10. If password and confirmed password do not match, an error should be displayed. [Automated]
 11. Email fields containing submitted with invalid formats should generate an error. 
      
 * Ensure the confirm email text field is not displayed on the edit user page (and only on the create new user page).
 
 12. For an existing active user, edit the user's account and click 'Generate new password' and check 'Send email'.
-    Save. Check that an email is sent to the user with the new password. Check that the password rules are enforced 
-    for this new password. Check that after logging in, the user is immediately asked to update his/her password.
+    Save. Check that an email is sent to the user with the new password. 
+    Check that after logging in, the user is immediately asked to update his/her password.
 13. Check that if creating a new user an email is sent to him/her (requires email server). Also check that when a new
     user is logging in for the first time he/she is asked to change his/her password.
     1. Check that when creating a new user, leading and trailing spaces in the username are stripped.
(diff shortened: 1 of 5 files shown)


johnsaigle:
I'll resume work on this after the other user accounts changes from #5514 and #5426 have been merged

modules/user_accounts/test/TestPlan.md
```diff
@@ -28,22 +28,22 @@ When creating or editing a user: (subtest: edit_user)
...
       - Last name,
       - Email.
-10. If password and confirmed password do not match, an error should be displayed.
+10. If password and confirmed password do not match, an error should be displayed. [Automated]
 11. Email fields containing submitted with invalid formats should generate an error. 
      
 * Ensure the confirm email text field is not displayed on the edit user page (and only on the create new user page).
 
 12. For an existing active user, edit the user's account and click 'Generate new password' and check 'Send email'.
-    Save. Check that an email is sent to the user with the new password. Check that the password rules are enforced
```
johnsaigle: "Checking the password rules" is more complex since we're using Zxcvbn instead of requiring special characters. I don't feel that this needs to be done manually by user's since this is covered now by a Password value object as well as unit tests for that file.

php/libraries/User.class.inc
```diff
@@ -492,8 +492,12 @@ class User extends UserPermissions
         \Password $password,
         ?\DateTime $expiry = null
     ): void {
+        if (password_verify($this->userInfo['Email'], (string) $password)) {
```
johnsaigle: This should've been added in https://github.com/aces/Loris/pull/5538 but was left out somehow.

johnsaigle:
Added Critical flag because it solves an issue that was marked as critical

modules/user_accounts/test/user_accountsTest.php
```diff
@@ -332,7 +332,7 @@ function testAddNewUser()
 
     /**
      * Modifies a field on either the user account or my preferences page
-     * and checks that the modification was recorded in the database.
+     * and checks that the modification was updated on the front-end.
```
driusan: github is highlighting the below `'user_accounts/my_preferences'`, line as a whitespace error.

johnsaigle:
@driusan This is good to go

PR #5534, opened 2019-11-05
Improve instrument builder test plan

coreyAbout:
## Brief summary of changes
Group validate process together to improve the test work flow

#### Testing instructions (if applicable)

1. Read the TestPlan item 1 to item 8.

#### Links to related tickets (GitHub, Redmine, ...)
fixed:
#5494 
#5493 
#5488 
#5484 
*

PR #5540, opened 2019-11-05
[Issue Tracker] update test plan to delete items not relevant

### LINKED ISSUE #5495: [issue_tracker] Test Plan needs to be updated
Some tests cannot be completed based on previous changes made to the module

coreyAbout:
## Brief summary of changes

To fix the test plan indicating that some the function was not available now

Check the test plan that the following items are removed:
Issue Tracker Create New Issue [Manual Testing]
7. Navigation bar should lead you back to list

Issue Tracker Edit Existing Issue [Manual Testing]
7. Check that the navigation bar bring you to the next issue in your previously filtered list of issues.

fixes #5495
*

### DIFF
--- a/modules/issue_tracker/test/issue_tracker_test_plan.md
+++ b/modules/issue_tracker/test/issue_tracker_test_plan.md
@@ -14,8 +14,7 @@ Issue Tracker Create New Issue [Manual Testing]
 4. Submit invalid and valid PSCID and visit label pairs. Error messages should respond accordingly. Not that you cannot submit PSCIDs from other sites unless you have access all profiles permission
 5. Submit just a visit label - this should give an error message.
 6. Check that all values are propagated and saved correctly.
-7. Navigation bar should lead you back to list [not available now].
-8. Check that watching logging is working - turn it off and on for your current user, and for other watchers on the issue
+7. Check that watching logging is working - turn it off and on for your current user, and for other watchers on the issue
 
 Issue Tracker Edit Existing Issue [Manual Testing]
 1. User can access the page if they have (developer or reporter permission) and (they have either access_all_profiles or are a member of the site of the issue or the site has no issue).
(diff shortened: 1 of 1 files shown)


johnsaigle:
I think it makes sense to delete these lines from the test plan rather than say that they are not available now. Whether or not a feature exists or is planned for development is relevant from the standpoint of someone testing.

coreyAbout:
I am not sure about that. Keep the items there would work as a reminder for developers when writing the test plan after the function is available. Otherwise, it may be easily forgotten, or sliped through. For testing purpose, testers know to skip it when the not available label is there. This should be the less harmful way. But anyway, whoever is in the position to make a final decision, I will happen to implement it.

driusan:
@johnsaigle is right. Test plans aren't the place to document a backlog of work..

johnsaigle:
We can make issues for those features so that there is a record that they should be implemented.

johnsaigle:
Would you mind adding a new issue card for these features @coreyAbout ?

coreyAbout:
Done as you wish @johnsaigle @driusan . 
Note that in another pull request I made, #5534, I was told to add not available label, but keep the item there.

coreyAbout:
@christinerogers contributed in discussion about this one as well.

christinerogers:
why is this a backlog / feature to be added?  The navigation panel was removed for 21.0 in this PR :
* #4855 `[issue_tracker] Remove navigation control panel`

We clearly wanted it removed.  Has this changed?

johnsaigle:
You're right, we can close this issue actually since it relates to an old feature.

When reviewing an older version of this PR there were items saying "not yet implemented" or something like that so I thought they were feature requests, not deleted features.

I'll close the issue that was opened.

driusan:
If it was intentionally removed then that's even more reason to delete the line and not annotate it as "[not available now]."

Issue #5259, opened 2019-10-02
Add bullet points to React Markdown rendering in LORIS


johnsaigle:
@knoll-alex 
> Does all of markdown work within LORIS?

No, at this time only a subset of Markdown works within LORIS:

https://github.com/aces/Loris/blob/3d6d2a060e349b34093146134e13caff440747e3/jsx/Markdown.js#L7-L12

So it looks like we don't support formatted bullet points at this time.

I'll create a new issue proposing that we add bullet points. I think it would be useful.

_Originally posted by @johnsaigle in https://github.com/aces/Loris/pull/5258#issuecomment-537563146_

maltheism:
I'll try to implement this feature.

knoll-alex:
Hello! Wondering what sort of timeline this might be on? I'm going to be updating all the helptext over the next couple weeks and this feature would be super helpful. Thanks.

johnsaigle:
Hi @knoll-alex! @maltheism will have to answer ultimately but development could potentially be delayed since we are in the middle of release testing. That means that new features will probably get pushed until the next release -- but with luck perhaps we can sneak this in.

maltheism:
@knoll-alex the issue might get held back as @johnsaigle mentioned it being release testing. I'll try to work on the bullet feature when time permits. I suggest you write Markdown with bullets when needed and the PRs can be merged in after this feature has been implemented.

johnsaigle:
I agree; go ahead and format it with bullet points for now and we'll make sure they work later.

johnsaigle:
@knoll-alex A pull request has been opened to add this feature. It will likely be released in LORIS version 23... so not for (likely at least a few months) unfortunately.

https://github.com/aces/Loris/pull/5542

PR #5542, opened 2019-11-05
[jsx/Markdown.js] Markdown feature bullets

maltheism:
## Brief summary of changes

Adds Markdown "bullets" which is the equivalent to unordered html list. 

* First implementation of the bullets to the unordered html list (feature). I couldn't think of a way to do this with regex.
* I'm not happy with the first attempt because it only works with single list. Next implementation need to make the function recursive for lists within a list. Although I thought it would be good to send a PR for now. 

1. Make a markdown file inside help for one of the modules with bullets.
2. View on the frontend.

* https://github.com/aces/Loris/issues/5259


ridz1208:
if this is not a bugfix please base it on major for now as it will only be part of the next release

maltheism:
I messed up the rebase @ridz1208 :(
See https://github.com/aces/Loris/pull/5615 for the new PR on major

christinerogers:
@ridz1208  This would be really nice to have in all the Documentation PRs going in on the 22 release, since there are big updates this week going on in the Help text (cc @knoll-alex).  
We've had to remove bullets in some cases - the option to keep them in now would be better.

PR #5543, opened 2019-11-05
Fix repetitive messages in review page issue 5116

### LINKED ISSUE #5116: [Survey Module] Fix Repetitive messages in Review page
The text 'Unanswered questions are highlighted in red' is repeated (once in a paragraph and again in a single line underneath
Screenshot from 2017-06-05 10_36_45

VictorVermeulen:
Fixes #5116

### DIFF
--- a/smarty/templates/directentry.tpl
+++ b/smarty/templates/directentry.tpl
@@ -65,7 +65,6 @@
 {$workspace}
 <h1 align="center">Review Page</h1>
 <p>Please review the data entered below. Unanswered questions are highlighted in red. Please use the back button to go access the previous pages and enter any data that is missing.</p>
-<p>Unanswered questions are highlighted in red.</p>
 {if $review}
 <hr>
 <br><br>


johnsaigle:
Thanks for submitting @VictorVermeulen! The changes look good to me.

VictorVermeulen:
Thanks for your patience! Made so many mistakes for my first pull request but I did learn a lot of new things. Maybe make it clearer to new people that you should create a branch from the default branch and not from the master branch? I thought the master branch was always the default and it took me a long time to figure this out since apparently it did not work because of that.

Also this was in the Survey Module according to the file, but I learned that looking in the Survey folder did not work for this. This was a good learning experience that module doesn't mean a specific folder to find the file. Is this for other projects also like this that you just have to search yourself where to make a change?

Those two points made this very time consuming for me, so maybe it can help other people in the future.

johnsaigle:
Thanks for the feedback.
You've caught us at an interesting time as we're actually changing how our standards for branching at the moment. After our next major release we will be reworking how contributors should base their PRs. At that time we'll update our CONTRIBUTING file with instructions.

To your second point: generally when something is marked as "Survey Module" you would be correct in looking for it under that folder in `modules/`.

Surveys are a bit of an exceptional case because: 
1) the code there is very old, and
2) Some portions of this module must be publicly accessible to users without a user account (i.e. no authentication)

Those reasons are why this file is in a weird place. Sorry for the confusion -- we're working on it!

In the future you should be able to find files within the proper `module/` folder. 

If you're unsure of which file a change should be made in, please comment on the issue and someone should be able to provide that info for you.

driusan:
Merged.. thanks @VictorVermeulen !

Issue #5544, opened 2019-11-05
[Examiner] Edit Examiner Page Broken if Certification Instruments Have Similar Names


sruthymathew123:
Edit Examiner Page Broken if Certification Instruments Have Similar Names

driusan:
Fixed by above

Issue #5423, opened 2019-10-30
[Examiner] Saving the New Examiner certification details giving 500 error 


sruthymathew123:
Testing Platform: [https://test-loris-dev.loris.ca/](https://test-loris-dev.loris.ca/)

On adding and saving certification details on a newly created examiner is giving a 500 error page.
Cases throwing Error
1) When trying to save with no certification data (500 error)
2) When trying to save data with one instrument details (500 error)

driusan:
Fixed by above

PR #5545, opened 2019-11-05
[Examiner] Fixing Errors while saving the form

sruthymathew123:
## Brief summary of changes
This fixes 

- https://github.com/aces/Loris/issues/5423 (I think this bug is caused as an impact of #4914. The phan rule addition changed the behavior of the examiner module a little bit. So I am replacing the Loris Exception with some` if `cases in the code.
- https://github.com/aces/Loris/issues/5544

1) Try to create a new examiner and edit the examiner details by just clicking 'Save' with no data or with just one instrument details filled. It will throw 500 error. Check out to  this branch and try testing with the same case and see if the issue is resolved

2)  Add Certification instruments with similar naming (like BMI, BMI_version_2 etc)  and try to load the edit examiner page. It will throw 500 error.  Check out to this branch and try testing with the same case and see if the issue is resolved

*

### DIFF
--- a/modules/examiner/php/editexaminer.class.inc
+++ b/modules/examiner/php/editexaminer.class.inc
@@ -201,17 +201,12 @@ class EditExaminer extends \NDB_Form
                         'TID' => $testID,
                     )
                 );
-                if (is_null($oldVals)) {
-                    throw new \LorisException(
-                        'No old certification values found for specified '
-                        . 'examiner and test IDs.'
-                    );
+                if (!empty($oldVals)) {
+                    // since there is an ORDER BY in pselect
+                    // $oldVals[0] corresponds to the latest date
+                    $oldVal  = $oldVals[0]['new'] ?? null;
+                    $oldDate = $oldVals[0]['new_date'] ?? null;
                 }
-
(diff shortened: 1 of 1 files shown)


johnsaigle:
I tested the first point and it seems to work (i.e. no 500 error generated). 

I guess my VM isn't configured properly to be able to test the second point because there are no Certification Instruments enabled. I'm not totally sure how to set this up because there are no examples in the config files and I'm not sure what is compatible with Raisinbread. I opened an issue for this, #5567.

sruthymathew123:
@johnsaigle Thank you for testing the first part.  I caught the second issue while testing on my VM with IBIS data. We have instruments with similar naming and it failed as the testname checking with `LIKE CONCAT('%', :testName, '%')",`

driusan:
@nicolasbrossard you probably have certification set up, can you test the second part of this?

nicolasbrossard:
I tested this on my VM (without Sruthy's fix) and I indeed get the 500 error. Page displays fine after I pull Sruthy's branch. 

Sruthy: 1    Bug: 0 .

:+1:

PR #5555, opened 2019-11-05
[New Profile] Fix TypeError when creating candidate with "user" generation method

johnsaigle:
## Summary

Fixes an issue when creating a new profile. This was originally sent to bugfix but the PR got auto-closed.

## Testing

Using the "user" generation method, try creating a new profile. 

You'll get a freeze on the front-end and a TypeError on the back-end. 

This should fix it.

### DIFF
--- a/modules/new_profile/php/new_profile.class.inc
+++ b/modules/new_profile/php/new_profile.class.inc
@@ -286,7 +286,7 @@ class New_Profile extends \NDB_Form
             } else {
                 // user has multiple sites,
                 // so validate PSCID against the Site selected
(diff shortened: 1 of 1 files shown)


johnsaigle:
@HenriRabalais Requesting your review because you're assigned to this module. It may not be possible to test on the test VM since it's not possible to change the generation method without accessing the config file.

ridz1208:
@johnsaigle does that mean it's also a bug on 21 ?

if so you could send it to 21.0-release

driusan:
declare(strict_type=1) was added to Site in 22, not 21.

Issue #5539, opened 2019-11-05
[Projects] add alias field to configuration page


ridz1208:
Alias was just added to projects for PSCID generation but it was forgotten in the configuration front end page.

johnsaigle:
Assigning you cause you said you'd take this

driusan:
Fixed by above

Issue #5432, opened 2019-10-30
[Testing] Remove magic numbers in integration and unit tests in favor of class constants


johnsaigle:
Many integration tests are getting stuck when data in Raisinbread changes. Part of this has to do with magic numbers.

This issue would be mitigated somewhat by changing [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants) into class constants so they can be updated in one place.

ridz1208:
can you link to examples ? and @cmadjar why is this 22.0.0 testing ?

cmadjar:
no idea why I did that... I removed the 22.0.0 testing label...

johnsaigle:
@ridz1208 This came to mind when you were trying to get a PR relating to projects to pass travis. In that case a NUM_CANDIDATES class constant would've been more useful than a magic number in the text.

ridz1208:
@johnsaigle thats what I thought.

In that particular case, yes constants would make it clearer but what would also be useful is for the different asserts to be split into different tests so they can all fail at once and be fixed at once rather then one at a time.

johnsaigle:
This ticket is probably too general to be useful. I think this should be more of a guiding principle that we incorporate during code review.

PR #5556, opened 2019-11-05
[Configuration] Added new Alias field to the front end

ridz1208:
## Brief summary of changes
https://github.com/aces/Loris/issues/5539

### DIFF
--- a/modules/configuration/ajax/updateProject.php
+++ b/modules/configuration/ajax/updateProject.php
@@ -25,8 +25,8 @@ $client->initialize();
 $factory = NDB_Factory::singleton();
 $db      = $factory->database();
 
-$ProjectList = Utility::getProjectList();
-$projectName = $_POST['Name'] ?? '';
+$ProjectList   = Utility::getProjectList();
+$projectName   = $_POST['Name'] ?? '';
 $projectAlias  = $_POST['Alias'] ?? '';
 $recTarget     = empty($_POST['recruitmentTarget'])
     ? null : $_POST['recruitmentTarget'];
--- a/modules/configuration/ajax/updateProject.php
+++ b/modules/configuration/ajax/updateProject.php
@@ -27,7 +27,6 @@ $db      = $factory->database();
 
 $ProjectList = Utility::getProjectList();
 $projectName = $_POST['Name'] ?? '';
-// FIXME This field should be added to the front-end.
 $projectAlias  = $_POST['Alias'] ?? '';
 $recTarget     = empty($_POST['recruitmentTarget'])
     ? null : $_POST['recruitmentTarget'];
--- a/modules/configuration/ajax/updateProject.php
+++ b/modules/configuration/ajax/updateProject.php
@@ -28,7 +28,7 @@ $db      = $factory->database();
 $ProjectList = Utility::getProjectList();
 $projectName = $_POST['Name'] ?? '';
 // FIXME This field should be added to the front-end.
-$projectAlias  = '';
+$projectAlias  = $_POST['Alias'] ?? '';
 $recTarget     = empty($_POST['recruitmentTarget'])
     ? null : $_POST['recruitmentTarget'];
 $projectID     = $_POST['ProjectID'] ?? null;
@@ -50,6 +50,7 @@ if ($projectID == 'new') {
     // Update Project fields
     $project = \Project::getProjectFromID($projectID);
     $project->updateName($projectName);
+    $project->updateAlias($projectAlias);
     $project->updateRecruitmentTarget($recTarget);
 
 }
--- a/php/libraries/Project.class.inc
+++ b/php/libraries/Project.class.inc
@@ -231,7 +231,7 @@ class Project
      *
      * @return void
      */
-    public function updateName(?string $projectName = null): void
+    public function updateName(?string $projectName = ''): void
     {
         if (!empty($projectName)) {
             $this->setName($projectName);
@@ -274,7 +274,7 @@ class Project
      *
      * @return void
      */
-    public function updateAlias(?string $projectAlias = null): void
+    public function updateAlias(?string $projectAlias = ''): void
     {
         if (!empty($projectAlias)) {
             $this->setAlias($projectAlias);
--- a/php/libraries/Project.class.inc
+++ b/php/libraries/Project.class.inc
@@ -227,13 +227,16 @@ class Project
     /**
      * Set this project's name
      *
-     * @param string $projectName This project's name
+     * @param ?string $projectName The new name. Uses class property when null.
      *
      * @return void
      */
-    public function updateName(string $projectName): void
+    public function updateName(?string $projectName = null): void
     {
-        $this->_projectName = $projectName;
+        if (!empty($projectName)) {
+            $this->setName($projectName);
+        }
+        $this->_updateField("Name", $this->_projectName);
     }
 
     /**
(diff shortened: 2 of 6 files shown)


modules/configuration/ajax/updateProject.php
```diff
@@ -28,7 +28,7 @@
 $ProjectList = Utility::getProjectList();
 $projectName = $_POST['Name'] ?? '';
 // FIXME This field should be added to the front-end.
```
johnsaigle: this line should be deleted now that you've added this feature

modules/configuration/ajax/updateProject.php
```diff
@@ -50,6 +50,7 @@
     // Update Project fields
     $project = \Project::getProjectFromID($projectID);
     $project->updateName($projectName);
+    $project->updateAlias($projectAlias);
```
johnsaigle: It would be good to add a back-end check to the front-end check that Alias is > 4.

An InvalidArgumentException should be thrown by the Project class when Alias > 4 characters.

Currently this is done only in `setAlias()`, but not in the constructor or in `updateAlias()`.

I'd suggest adding a `_validateAlias(): void` function that throws the Exception based on the `_projectAlias` variable and then calling that function in set, update, and the constructor.

johnsaigle: Come to think of it this would be a good use case for an Alias value object that can also be used by Site. 

We can make a separate issue for this and punt to the next release if you prefer.

ridz1208: Shouldnt we be then using the set functions and then updateProject, instead of separate update functions for each value ?

ridz1208: and for the alias object i would punt it because we dont have the same criteria for a psc alias, a mri alias and a project alias, we would need to standardise first

johnsaigle: Created #5565 for adding a value object.

johnsaigle: >Shouldnt we be then using the set functions and then updateProject, instead of separate update functions for each value ?

That would work too, just depends on how we want the class to work. I don't mind either way

modules/configuration/ajax/updateProject.php
johnsaigle: I'll approve after you fix this; we can save the back-end check for later

modules/configuration/ajax/updateProject.php
ridz1208: I'll add the validation still but we'll rethink the logic later

ridz1208: I made it pass through set which allowed me to remove the `??`

php/libraries/Project.class.inc
```diff
@@ -227,13 +227,16 @@ class Project
     /**
      * Set this project's name
      *
-     * @param string $projectName This project's name
+     * @param ?string $projectName The new name. Uses class property when null.
      *
      * @return void
      */
-    public function updateName(string $projectName): void
+    public function updateName(?string $projectName = null): void
```
johnsaigle: Why did you add nullable here? Do we want the project name to be null?

php/libraries/Project.class.inc
```diff
@@ -265,33 +268,18 @@ class Project
...
-     * Update the Alias field in the databse for this Project.
+     * Update the Alias field in the database for this Project.
      *
-     * @param ?string $alias The new alias. Uses class property when null.
+     * @param ?string $projectAlias The new alias. Uses class property when null.
      *
      * @return void
      */
-    public function updateAlias(?string $alias = null): void
+    public function updateAlias(?string $projectAlias = null): void
```
johnsaigle: Alias is non-nullable in the DB. Maybe we should keep this a string?
```suggestion
    public function updateAlias(string $projectAlias = ''): void
```

php/libraries/Project.class.inc
```diff
@@ -341,10 +329,28 @@ class Project
      */
     public function updateRecruitmentTarget(?int $target = null): void
     {
-        $this->_updateField(
-            "recruitmentTarget",
-            $target ?? $this->_recruitmentTarget
-        );
+        if (isset($target)) {
+            $this->setRecruitmentTarget($target);
+        }
```
johnsaigle: I think this value will always be set, either to the value passed or `null`.

test/unittests/NDB_ConfigTest.php
```diff
@@ -246,11 +246,13 @@ public function testGetProjectSettings()
         $info   = array(
                    'ProjectID'         => '999',
                    'Name'              => 'test',
+                   'Alias'             => 'TST',
```
johnsaigle: Class constant would be nicer

php/libraries/Project.class.inc
ridz1208: it defaults to the objects _projectName which is required by the singleton. the reason it is nullable here is if you do something like
```
setName("BLA");
updateName();
```
that would persist the name to BLA

php/libraries/Project.class.inc
ridz1208: I had left that as is from your PR and adapted the other update functions to behave similarly.

since one validation is already being done in the set function, I would say we can add the non null check to the set function and the create new function maybe ?

php/libraries/Project.class.inc
ridz1208: this should be doing the same thing as ur null coalesce, i dont think the behaviour changes.

it defaults to the classes' _recruitmentTarget if null is passed and if the classes in null it sets it to null i guess

test/unittests/NDB_ConfigTest.php
ridz1208: just for that value ?

php/libraries/Project.class.inc
johnsaigle: Why not use an empty string instead of null?

ridz1208: not sure i understand, whats the advantage of an empty string ?

php/libraries/Project.class.inc
johnsaigle: I think it makes sense to type hint as `string` so that it can't be null. Then we can use empty strings where we're using `null` and leave the empty() checks.

The types in the class should agree with the types in the DB or else we're inviting bugs.

php/libraries/Project.class.inc
johnsaigle: Null coalesce handles `null` and not set. `isset()` only handles not set.

johnsaigle: When PHP 7.4 comes out we can use typed class properties so this won't be an issue. 😄

ridz1208: https://www.php.net/manual/en/function.isset.php

isset() only returns false when a value is NULL

johnsaigle: My bad. Don't know why I thought that wasn't the case but I definitely have for a long time.

Thanks

test/unittests/NDB_ConfigTest.php
johnsaigle: Just if you're gonna add something new a constant would be cool. not a bit deal though, just relating to our discussion on #5432

php/libraries/Project.class.inc
johnsaigle: In terms of static analysis it's a bit better to have functions that deal strictly with strings if they're meant to be strings instead of `string|null`. If something can be achieved with a stricter type it's usually helpful.

I would flip it around and say what is the advantage of `null` when we can just use an empty string?

ridz1208: i see you point for the static analyses.

there are 2 things here I wanna make sure not to confuse

 - the nullable `?` is because given the current logic of the class, that would allow people the set the value and then choose the persist it by calling the object function with no arguments. i would hate seeing anywhere is the code `updateName(null)` or `updateName('')`
- the `= null` part, is just because i prefer doing null checks then empty string checks when possible. in this specific case it doesnt make a difference since the empty function returns true to both. So i can change it to `''`

johnsaigle: Yep that sounds good to me. Like I said I thing `string $projectName = ''` is a bit better to `?string $projectName = null` but in this case specifically they're logically equivalent and either works.

php/libraries/Project.class.inc
ridz1208: because PHP doesnt make sens

$x=null
isset($x) -> flase

:confused:

johnsaigle:
@ridz1208 I did a PHPCBF for you

ridz1208:
@johnsaigle thanks for the phpcbf.

I made the change null to string (phpdoc didnt need to be corrected since the type was string already). I left the `?` for the reasons described above. Does that work ?

johnsaigle:
> the nullable ? is because given the current logic of the class, that would allow people the set the value and then choose the persist it by calling the object function with no arguments. i would hate seeing anywhere is the code updateName(null) or updateName('')

You can still do `updateName()`with no arguments as long as there is a default argument (in this case `''` the empty string). You don't need `?`.

ridz1208:
@johnsaigle thats absolutely true lool

Can u please remove them ?

Issue #5551, opened 2019-11-05
[Data Dict] Test Plan Needs Updating


HenriRabalais:
Test plan is erroneous - the data_dictionary_builder.php script can be found in `tools/exporters/` rather than `tools/`

driusan:
Fixed by above

PR #5558, opened 2019-11-06
[datadict] Test plan improved readability and fixed script location

ridz1208:
## Brief summary of changes
https://github.com/aces/Loris/issues/5551

mostly white space changes except the path in item 11

### DIFF
--- a/modules/datadict/test/TestPlan.md
+++ b/modules/datadict/test/TestPlan.md
@@ -1,29 +1,39 @@
 # Data Dictionary - Test Plan  
 
-1. Check that you have access to the Data Dictionary if the user has one or more of these permissions: 
-   data_dict, data_dict_edit or superuser.[Automation Testing]
-2. Check that the instruments combo box contains all the instrument names.[Automation Testing]
-3. Perform various searches and validate the results. Use any combination of the available search criteria. Make sure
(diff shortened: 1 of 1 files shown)


ridz1208:
@johnsaigle I wouldnt because we should still test everything that is automated anyways
automated tests often dont cover edge cases, dont change permissions, dont use sing sites vs multiple sites, single projects vs multi projects, tests could be corrupted if they were merged without proper review....

this is our chance to not only test the code but also make sure to test the tests

johnsaigle:
> dont change permissions

I think almost all of them do this actually. But I see your point.

ridz1208:
https://github.com/aces/Loris/issues/5560

thats a good example...

johnsaigle:
Thanks for the example. I guess I had assumed that when there was an automated test for something then we could assume that that part is covered. And that was why some thing were marked manual and other things as automated in the test plan.

Is the example in that issue marked as automated?

ridz1208:
in that case nothing is marked. but examples like media where there is a test with no permissions and one with only `media_read` permission but non with `media_write` permission come to mind.

HenriRabalais:
I was making the same assumption as @johnsaigle

PR #5571, opened 2019-11-06
[Test] Fix Docker scripts crashing when variables not set

### LINKED ISSUE #5515: [Testing] Strict bash flags cause docker scripts to fail locally
`test/dockerized-unit-tests.sh` checks if the `$DEBUG` variable is given a certain value to determine how it should run.

This doesn't work after #5051  because the script will fail when this variable is not set.

johnsaigle:
## Brief summary of changes

A recent PR causes strict checking of variables in bash. We are doing some sneaky environment variable tricks relying on variables sometimes being set so this made our scripts crash.

They should now work as before.

1. Running `npm run tests:unit` right now should cause an 'unbound variable error'
2. You should be able to run `npm run tests:unit` locally without that error (though the tests may fail for other reasons)

* Resolves #5515

### DIFF
--- a/Dockerfile.test.php7.debug
+++ b/Dockerfile.test.php7.debug
@@ -1,9 +1,9 @@
-FROM php:7.0
+FROM php:7.2
 
 RUN apt-get update && \
-    apt-get install -y mysql-client zlib1g-dev
+    apt-get install -y mariadb-client zlib1g-dev
 
-RUN yes | pecl install xdebug-2.4.1
+RUN yes | pecl install xdebug-2.8.0
 RUN echo "zend_extension=$(find /usr/local/lib/php/extensions/ -name xdebug.so)" > /usr/local/etc/php/conf.d/xdebug.ini
 RUN echo "xdebug.remote_enable=on" >> /usr/local/etc/php/conf.d/xdebug.ini
 RUN echo "xdebug.remote_autostart=on" >> /usr/local/etc/php/conf.d/xdebug.ini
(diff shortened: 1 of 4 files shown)


test/dockerized-integration-tests.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-if [ "$DEBUG" == "true" ]
+if [ -v DEBUG ];
```
driusan: What does this do? I can't find any documentation of "[" or "test" having  "-v" option anywhere.

johnsaigle: I just found it from searching how to check whether a variable is set in bash. Here's an example from Stack Exchange

https://unix.stackexchange.com/a/212192

johnsaigle: I tested it manually and the scripts work for me now where before they would crash.

Dockerfile.test.php7.debug
```diff
@@ -1,9 +1,9 @@
-FROM php:7.0
+FROM php:7.2
```
johnsaigle: Running this file (`npm run tests:unit:debug`) fails without these changes.

driusan: This made me nervous until I checked the .travis.yml and you're right, it's doing

`- sed -i "s/7.2/$TRAVIS_PHP_VERSION/g" Dockerfile.test.php7` (7.2, not 7.0).. so how is this working now?

Issue #5449, opened 2019-10-31
[document_repository] files from all directories shown without filtering


3design:
task
9. Check that expand/collapse works for non-empty categories. Check that when performing a search, the categories containing the resulting files (and only those) are shown in a list format [Manual Testing]

expected results:
when in a directory, only files in that directory are shown in the list

actual results: 
all files from all sub directories are shown in the list

comments: For example I have the following files in these directories
root> file1.txt
root> file2.txt
root>level1> file3.txt
root>level1>level1a> file4.txt

I expect to see
root>
file1.txt
file2.txt

I actually see
root>
file1.txt
file2.txt
file3.txt
file4.txt

when in level 1 I see
root>level1
file3.txt
file4.txt

zaliqarosli:
same issue as https://github.com/aces/Loris/issues/4857

driusan:
Fixed by @zaliqarosli's PR!

PR #5572, opened 2019-11-06
[DocRepo] Clean up directory content and let filter without folders

zaliqarosli:
## Brief summary of changes

1. Remove extraneous files that actually belong to subdirectories
2. Add global filtering of files only, without directory structure
3. Rename a couple states and functions to be more meaningful

On `22.0-release`, verify:

1.  Files in subcategories are recursively shown in parent categories (read description of this issue https://github.com/aces/Loris/issues/5449 for more clarity)
2. Clicking on 'Global Selection Filter' just takes you to the root directory

On this branch, test:

1. Files exist in their rightful categories and their categories alone.
2. Clicking on 'Filter globally' removes the directory structure and lists all files so that you can filter for all. The directory tree display ( _Root > ...._ ) disappears as well.
3. Unchecking 'Filter globally' will return table to normal folder structure with directory tree display. Filtering whilst in this state will only filter for files within the directory you're in.

* Fixes https://github.com/aces/Loris/issues/5449
* Fixes https://github.com/aces/Loris/issues/4857 (duplicate)

### DIFF
--- a/modules/document_repository/jsx/docIndex.js
+++ b/modules/document_repository/jsx/docIndex.js
@@ -18,15 +18,16 @@ class DocIndex extends React.Component {
       tableData: [],
       childrenNode: [],
       parentNode: [],
-      checked: false,
+      global: false,
     };
     // Bind component instance to custom methods
-    this.handleCheck = this.handleCheck.bind(this);
+    this.handleGlobal = this.handleGlobal.bind(this);
     this.fetchData = this.fetchData.bind(this);
     this.formatColumn = this.formatColumn.bind(this);
     this.newCategoryState = this.newCategoryState.bind(this);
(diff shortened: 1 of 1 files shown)


zaliqarosli:
@jesscall if you're interested!

jesscall:
This is the greatest thing ever

driusan:
Is there a reason the "Filter globally" checkbox is outside of the filters section?

zaliqarosli:
Because it’s not rendered within the FilterableDataTable component. It’s rendered in the Doc Repo index file. I think this makes sense because it’s not a feature needed in other uses of FilterableDataTable. It’s specific to doc repo alone

driusan:
It looks a little out of place to me, but I'm not going to block over it.

PR #5573, opened 2019-11-07
First visit column of table test_battery not taken into account.

### LINKED ISSUE #5569: 'First Visit' setting in table test_battery not taken into account
On the testing VM, instrument `medical_history` is set to be part of visit V06 iff it's not the first visit. I created visit V06 for a candidate that had no visit so far and `medical_history` was added to the list of instruments for that (first) visit.

Found during testing of 22.0.

nicolasbrossard:
When creating a visit, the computation of the list of instruments that should be administered is incorrect: the `first_visit` column of table `test_battery` is not handled properly by the algorithm. 

This PR fixes #5569.

### DIFF
--- a/php/libraries/NDB_BVL_Battery.class.inc
+++ b/php/libraries/NDB_BVL_Battery.class.inc
@@ -495,7 +495,7 @@ class NDB_BVL_Battery
             $query_firstVisit = "
                 SELECT DISTINCT Test_name
                 FROM test_battery
-                WHERE firstVisit=:FV AND SubprojectID=:subID";
+                WHERE firstVisit=:FV AND (SubprojectID IS NULL OR SubprojectID=:subID)";
             $where            = array(
                 'FV'    => 'Y',
                 'subID' => $SubprojectID,
@@ -511,7 +511,7 @@ class NDB_BVL_Battery
             $query_notfirstVisit = "
                 SELECT DISTINCT Test_name
                 FROM test_battery
-                WHERE firstVisit=:FV AND SubprojectID=:subID";
+                WHERE firstVisit=:FV AND (SubprojectID IS NULL OR SubprojectID=:subID)";
             $where = array(
                 'FV'    => 'N',
                 'subID' => $SubprojectID,
--- a/php/libraries/NDB_BVL_Battery.class.inc
+++ b/php/libraries/NDB_BVL_Battery.class.inc
@@ -444,7 +444,11 @@ class NDB_BVL_Battery
                         (b.AgeMinDays<= :vAge AND b.AgeMaxDays >= :vAge)
                     )
                     AND b.Active='Y'
-                    AND (SubprojectID=:SubprojID OR SubprojectID IS NULL)";
+                    AND (
+                         SubprojectID=:SubprojID
+                            OR
+                         SubprojectID IS NULL
+                    )";
 
         if (!is_null($visit_label)) {
             $VisitBatteryExists = $DB->pselectOne(
@@ -495,7 +499,8 @@ class NDB_BVL_Battery
             $query_firstVisit = "
                 SELECT DISTINCT Test_name
                 FROM test_battery
-                WHERE firstVisit=:FV AND (SubprojectID IS NULL OR SubprojectID=:subID)";
+                WHERE firstVisit=:FV
+                AND (SubprojectID IS NULL OR SubprojectID=:subID)";
             $where            = array(
                 'FV'    => 'Y',
                 'subID' => $SubprojectID,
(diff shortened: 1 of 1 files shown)


driusan:
@nicolasbrossard this is failing Travis

php/libraries/NDB_BVL_Battery.class.inc
```diff
@@ -495,7 +499,8 @@ class NDB_BVL_Battery
             $query_firstVisit = "
                 SELECT DISTINCT Test_name
                 FROM test_battery
-                WHERE firstVisit=:FV AND SubprojectID=:subID";
+                WHERE firstVisit=:FV
+                AND (SubprojectID IS NULL OR SubprojectID=:subID)";
```
sruthymathew123: @nicolasbrossard @driusan I am getting confused with these changes; Every valid test battery entry is supposed to have a valid subproject associated with that; right? Then why do we need to check `AND (SubprojectID IS NULL` Please correct me if I am wrong. 

Also, the first visit case is working as expected on my VM (22.0 Release branch) without any issues. But I don't know if it is a test VM specific issue.

nicolasbrossard: @sruthymathew123 @driusan I do find the `first_visit` thing a bit confusing myself...The way I see it is that if you specify something in the `first_visit` column without specifying something in the subproject column then is_first_visit/is_not_first_visit constraint applies to all visits/candidates/sub-projects. Whereas if you specify something in both `first_visit` and `subproject` then the constraint applies only to the targeted subproject. 

That's how I tested this.

sruthymathew123: @driusan @nicolasbrossard  I personally approve this logic and pass manual tests if `First visit `**should be considered** even if SubprojectID is not defined in the test battery.  But for IBIS I think, most of the valid test battery entries demand a valid SubprojectID. So @driusan leaving to your court for the final approval.

driusan: @ridz1208 you're the most recent person to have worked out the logic.. what does the documentation/help text say on this?

From my perspectively, if the column is nullable, it makes sense that null should be valid.

ridz1208: my notes on first visit

 - FirstVisit:
   - If `firstVisit` is set to `Y`, the test battery instance is applied
   only if it is the first visit.
   - If `firstVisit` is to `N`, the test battery instance is applied only if it
   *not* the first visit.
   - If `firstVisit` is set to null, the test battery instance is applied to any
   visit.
   - **NOTE:** *The `firstVisit` flag can allow to bypass all other rules in
   some instances. In these instances, the `stage`, `CenterID`, `Visit_label`,
   `AgeMinDays` and `AgeMaxDays` will not affect in any way the administration
   of the instrument. Only the `subprojectID` value will impact if the
   instrument gets administered or not; the `subprojectID` value must match
   the timepoint's in order for the instrument to be administered in these
   instances.*

@sruthymathew123 if this is the current behaviour, it's better not to change it to avoid unexpected changes on projects

driusan: I don't understand what the note means WRT null. Should this be merged or closed?

ridz1208: if I understand @sruthymathew123 comment correctly, this PR should be closed because it would be altering the intended way that first visits work

driusan: I don't see how this can break things for existing projects, in order for the behaviour to change this would mean that it was intentionally setup with *both* FirstVisit not null *and* SubprojectID null intentionally, which is a no-op according to the documentation you quoted.

ridz1208: but if any project had that in their database, it never took effect but would start taking effect when this is merged. so unless you run a script checking that no entry has that combination and if some do, notify the user that these will now take effect... it would be considered as unexpected behaviour

and in that case the label "Feature" should be added and I dont see why it's critical to release

driusan: @nicolasbrossard thoughts?

nicolasbrossard: @driusan @ridz1208 @sruthymathew123 I am for merging this. If you read the LORIS setup/configuration wiki, there's no mention that you have to set the SubprojectID column **and** the first_visit column otherwise that line is ignored, so technically a line like this is valid and should be taken into account. In other words, the current issue is a bug.

As for existing projects having these no-op lines: I doubt these would go unnoticed for long and someone would eventually point out that the "first visit" criterium does not work properly.

ridz1208: I guess it boils down to what do we want out of the first visit flag. I'm only reporting how it is currently working and how it has been for as long as I could check (19.0 to 22?).

We get to decide now if we alter the behaviour in favour of the documentation or if we alter the documentation in favour of the behaviour.

maybe we could even have an "ideal" solution ?? ie if no subproject is set treat it as "all" if one is set apply it only for that subproject and so on and so forth for site and visit and age...

nicolasbrossard: Your ideal solution makes a lot of sense to me: all the criteria should be ANDed together and if you have NULL for any criterium, that criterium is ignored when computing the test battery.

driusan:
This was discussed in the LORIS meeting. Since it's a change of behaviour, we're not going to do it for the 22.0 release.

driusan:
This to be fixed when test battery rewritten as per Nov 20, 2019 comment

Issue #5363, opened 2019-10-23
[21.0] [media] Form not resetting after upload


leighevans:
After uploading a file, the fields are reset and blank, but the Site options are still constrained for the previous upload. In theory this issue is resolved by simply entering the PSCID, which updates the Site list to the proper sites, but just an FYI that if you don't follow the data input in order, this is what happens.

Screenshot shows the site dropdown with options that pertain to the previous upload, rather than displaying all sites for a new upload. 

@HenriRabalais EDIT: The same behaviour that occurs for sites also occurs for visit labels and instruments.

![Screen Shot 2019-10-23 at 11 46 18 AM](https://user-images.githubusercontent.com/32552953/67410950-dc30fa00-f58a-11e9-8cf8-28577313d3ab.png)

johnsaigle:
Duplicate of #5055

jesscall:
@johnsaigle I believe #5055 has to do with the `Browse` page not updating. This issue refers to the upload form's options for site, visit and instrument not resetting

johnsaigle:
Got it, thanks for clarifying.

cmadjar:
@HenriRabalais We assigned you the media module bug fixes for Monday. While you are at it, maybe you can fix this one too? Thanks!

HenriRabalais:
@cmadjar sounds good! should be an easy fix!

HenriRabalais:
After looking into this further, there is the same issue for visit label and instruments, so I might as well tackle all the similar issues with this one PR.

HenriRabalais:
PR Sent #5577

driusan:
Does #5577 replace this or complement this?

driusan:
5577 is merged

PR #5577, opened 2019-11-07
[Media] Resolving Issues from Testing

HenriRabalais:
Addresses Issues From:
#5363 - Prevents Users from selecting a site, visit label, or instrument unless a pscid is selected.

This also removes the Site Field from the upload form and edit form in media. Site is redundant since visits are already associated with a site. This removes two unnecessary clicks for the user when uploading a file, and doesn't alter any data going to the database since the site was never saved anyways.

### DIFF
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -113,8 +113,7 @@ function uploadFile()
     $language   = isset($_POST['language']) ? $_POST['language'] : null;
 
     // If required fields are not set, show an error
-    if (!isset($_FILES) || !isset($pscid) || !isset($visit) ||
-        !isset($instrument) || !isset($language)) {
+    if (!isset($_FILES, $pscid, $visit, $instrument, $language)) {
         showMediaError("Please fill in all required fields!", 400);
         return;
     }
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -108,13 +108,13 @@ function uploadFile()
     $pscid      = isset($_POST['pscid']) ? $_POST['pscid'] : null;
     $visit      = isset($_POST['visitLabel']) ? $_POST['visitLabel'] : null;
     $instrument = isset($_POST['instrument']) ? $_POST['instrument'] : null;
-    $site       = isset($_POST['forSite']) ? $_POST['forSite'] : null;
     $dateTaken  = isset($_POST['dateTaken']) ? $_POST['dateTaken'] : null;
     $comments   = isset($_POST['comments']) ? $_POST['comments'] : null;
     $language   = isset($_POST['language']) ? $_POST['language'] : null;
 
     // If required fields are not set, show an error
-    if (!isset($_FILES) || !isset($pscid) || !isset($visit) || !isset($site)) {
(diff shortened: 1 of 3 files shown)


HenriRabalais:
@maltheism thanks for the testing! I'm blocking this though based on a recent discussion with Rida. We are actually going to remove the 'Site' because it is unnecessary. This seems like the right PR to make that change.

HenriRabalais:
@maltheism unblocked!

modules/media/ajax/FileUpload.php
```diff
@@ -209,7 +208,7 @@ function getUploadFields()
     $user   = \User::singleton();
     $config = \NDB_Config::singleton();
 
-    // Select only candidates that have had visit at user's sites
+    // Select only candidates that have had visit at user's ites
```
maltheism: ites or sites

HenriRabalais: oops lol

modules/media/ajax/FileUpload.php
```diff
@@ -108,13 +108,13 @@ function uploadFile()
...
     $visit      = isset($_POST['visitLabel']) ? $_POST['visitLabel'] : null;
     $instrument = isset($_POST['instrument']) ? $_POST['instrument'] : null;
-    $site       = isset($_POST['forSite']) ? $_POST['forSite'] : null;
     $dateTaken  = isset($_POST['dateTaken']) ? $_POST['dateTaken'] : null;
     $comments   = isset($_POST['comments']) ? $_POST['comments'] : null;
     $language   = isset($_POST['language']) ? $_POST['language'] : null;
 
     // If required fields are not set, show an error
-    if (!isset($_FILES) || !isset($pscid) || !isset($visit) || !isset($site)) {
+    if (!isset($_FILES) || !isset($pscid) || !isset($visit) ||
```
johnsaigle: Just FYI you can pass multiple argument to isset

```php
if (!isset($arg1, $arg2, ...))
```

HenriRabalais: love it

HenriRabalais: @johnsaigle looks like that should fix the phpcs errors as well

HenriRabalais:
@johnsaigle resolved issue

ridz1208:
blocked by discussion on individual issues

ridz1208:
@HenriRabalais 

could you add more details on how you solved this : 

> #5363 - Prevents Users from selecting a site, visit label, or instrument unless a pscid is selected.

Cause from what I can tell simply removing the site selection should be sufficient to solve the error reported by Leigh

driusan:
Scratch what I said about the site being for uploading content for other sites.. IBIS uses a different module for reliability videos, so inferring the site in this one should be fine. (The other points stand.)

HenriRabalais:
@ridz1208 if you scroll down and read the comments on the issues (#5363) I mention that I discovered that the same issue that involves sites also occurs for both visit label and instruments. Should I update the issue or make a new one to reflect those similar problems?

HenriRabalais:
@driusan @ridz1208 I have removed language and instrument as required fields from the form. Language as a required field can be addressed in another PR if necessary.

PR #5365, opened 2019-10-23
[issue_tracker] Fixes An error occurred when loading the form!

maltheism:
## Brief summary of changes

Two small solutions for fixing the "An error occurred when loading the form!" frontend message.

#### Testing instructions (if applicable)

1. Pull PR
2. Test on Rasinbread data.
3. Click an issue from the table column "Issue ID" and no error with PR changes.

PR #5580, opened 2019-11-08
[issue_tracker] Cleanup PHP Notice caused by unset variable

maltheism:
## Brief summary of changes

Solves error log pollution when issueComment is null.

1. Checkout PR
2. View issue when no comment

#5365

### DIFF
--- a/modules/imaging_browser/php/viewsession.class.inc
+++ b/modules/imaging_browser/php/viewsession.class.inc
@@ -221,7 +221,7 @@ class ViewSession extends \NDB_Form
             $SeriesNumber        = $FileObj->getParameter('series_number');
             $SeriesUID           = $FileObj->getParameter('series_instance_uid');
             $EchoTime            = number_format(
-                (int)$FileObj->getParameter('echo_time') * 1000,
+                (int)$FileObj->getParameter('echo_time')*1000,
                 2
             );
             $RepetitionTime      = number_format(
--- a/modules/imaging_browser/php/viewsession.class.inc
+++ b/modules/imaging_browser/php/viewsession.class.inc
@@ -221,11 +221,11 @@ class ViewSession extends \NDB_Form
             $SeriesNumber        = $FileObj->getParameter('series_number');
             $SeriesUID           = $FileObj->getParameter('series_instance_uid');
             $EchoTime            = number_format(
-                (int)$FileObj->getParameter('echo_time')*1000,
+                (float)$FileObj->getParameter('echo_time') * 1000,
                 2
             );
             $RepetitionTime      = number_format(
-                (int)$FileObj->getParameter('repetition_time') * 1000,
+                (float)$FileObj->getParameter('repetition_time') * 1000,
                 2
             );
             $SliceThickness      = number_format(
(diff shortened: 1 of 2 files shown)


maltheism:
@ridz1208 should I rebase this to 22.0-release?

johnsaigle:
@maltheism I don't think anything else will be issued to 21.0-release. From what I understand we should be using 22.0-release for that release, `master` for bug fixes post-22.0, and `major` for breaking changes (23 release).

So yes I think it would be good to rebase this to `22.0-release` or `master`.

I'm also going to remove the `Critical` label -- not because the fix isn't important, but I think it's up to Dave, Rida, Samir, etc. to determine what is and isn't critical, not us.

modules/imaging_browser/php/viewsession.class.inc
```diff
@@ -221,11 +221,11 @@ class ViewSession extends \NDB_Form
             $SeriesNumber        = $FileObj->getParameter('series_number');
             $SeriesUID           = $FileObj->getParameter('series_instance_uid');
             $EchoTime            = number_format(
-                (float)$FileObj->getParameter('echo_time') * 1000,
+                (int)$FileObj->getParameter('echo_time') * 1000,
```
johnsaigle: I think this is a step backwards from https://github.com/aces/Loris/pull/5401/files

What is the reason for this change?

maltheism:
@johnsaigle everyone seems to be adding the critical to release tag to their PRs. I thought this one would qualify since the module breaks if no comment exists for the issue.

johnsaigle:
I think they're only adding them when fixing a bug marked as Critical.

johnsaigle:
Can you expand the testing instructions? I'm not able to reproduce the bug reported in the issue (though I'm also not sure what to expect).

I'm testing with fresh Raisinbread data and clicking a link in Issue ID column. It's taking me to the issue and I'm not seeing any issue or anything appear in the error log.

maltheism:
@johnsaigle "View issue when no comment" meaning you need to view an issue that has no comment associated with it.

johnsaigle:
I understand that. I'm trying a few issues in the Raisinbread dataset and as far as I can tell none of them have comments. There doesn't seem to be an issue.

maltheism:
@johnsaigle you could make a new issue without a comment and try clicking on that issue if rasinbread doesn't have any issues without comments.

johnsaigle:
I tried that and there was still no issue.

johnsaigle:
We resolved this in person -- I'm not able to replicate the bug originally reported but this PR still prevents a PHP Notice so it's helpful to have.

It can probably go to `master` if not `22.0-release`

maltheism:
Updated description of error just now. The undefined seems to throw an error message on my setup of LORIS but mainly it pollutes the error log for other VMs.

christinerogers:
@maltheism was there any visible front-end side effect?  (and/or was there an issue ticket?)

@3design is re-testing today -- what should he look for?

maltheism:
@christinerogers just error log from the backend.

PR #5584, opened 2019-11-08
[acknowledgements] fix bug where users with edit permission only could not access the module

cmadjar:
## Brief summary of changes

Users with only `acknowledgements_edit` permission could not access the acknowledgements module. This fixes this issue.


cmadjar:
@Jkat Could you review this PR? Should be pretty straightforward ;). Thanks!

Issue #5581, opened 2019-11-08
[Media] Edit metadata window not loading


SantiagoTG:
When you try to open the edit metadata window, the window gets stuck loading, and thus can't edit data.

![Screen Shot 2019-11-08 at 11 44 08 AM](https://user-images.githubusercontent.com/54332302/68494981-4e593e00-021d-11ea-954d-6b53fba3b464.png)

ridz1208:
@SantiagoTG good catch. I'm adding this here for more details
![Screenshot from 2019-11-08 13-14-35](https://user-images.githubusercontent.com/15268287/68500482-cded0a00-0229-11ea-8f8b-67e527ed574f.png)

HenriRabalais:
PR Sent: #5585

HenriRabalais:
Resolved

Issue #5459, opened 2019-10-31
[Media] Language option in 'Edit' page


SantiagoTG:
When trying to edit data using the Edit button under the "Edit Metadata" tab, there is no language option, and so language cannot be edited.

cmadjar:
@HenriRabalais Can you look into this?

HenriRabalais:
@cmadjar There seems to be other issues preventing me from resolving this - the media edit form won't load at the moment. It seems like there's issues with ID value being passed to the edit form. I will look at resolving that issue first, and then tackling this one.

HenriRabalais:
PR Sent: #5585

HenriRabalais:
Resolved

PR #5585, opened 2019-11-08
[Media] Added Project to Columns; Added Language to Edit File Form

HenriRabalais:
This PR Addresses Two Issues: 

#5581
I fixed the media datatable by adding Project to the column list and filter. Projects had been added to the query in MediaFileProvisioner, but was not added to the fields list which populates the Datatable. This was causing an issue where the values were not displayed right, and the edit file form was not accessible.

#5459
There was no field for 'Language' in the edit media form, and therefore there was no way to edit language. Language has now been added as a field and is editable.

### DIFF
--- a/modules/media/jsx/editForm.js
+++ b/modules/media/jsx/editForm.js
@@ -32,6 +32,7 @@ class MediaEditForm extends Component {
 
   componentDidMount() {
     let self = this;
+    console.log(this.props.DataURL);
     $.ajax(this.props.DataURL, {
       dataType: 'json',
       success: function(data) {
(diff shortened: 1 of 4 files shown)


modules/media/jsx/editForm.js
```diff
@@ -32,6 +32,7 @@ class MediaEditForm extends Component {
 
   componentDidMount() {
     let self = this;
+    console.log(this.props.DataURL);
```
maltheism: You forgot to remove console.log

HenriRabalais: Good catch!

PR #5591, opened 2019-11-11
[Statistics / Behavioural Statistics] Fix: Users with permissions different from 'access_all_profiles' were not able to see the 'breakdown per participant' page.

racostas:
### Description of the issue

Users with permissions different from 'access_all_profiles' were not able to see the 'breakdown per participant' page. An access restricted page were shown instead.

### Brief summary of changes
The "_hasAccess" function was modified to allow access the page in the case not only one CenterID is passed on the URL

The function "_checkCriteria" was modified for only retrieve data from the sites(centers) the user have access to. 

**Note:** In future developments the capacity of setting permissions per site per user could be desirable.  A note in rapport was included in the codeset. In this case the proposed function "_hasAccess" should be updated accordingly .

A user with permission 'data_entry' should be now able to access the 'breakdown per participant' page for the sites it have access to.

### DIFF
--- a/modules/statistics/php/statistics_site.class.inc
+++ b/modules/statistics/php/statistics_site.class.inc
@@ -89,10 +89,11 @@ class Statistics_Site extends \NDB_Menu
             $this->query_criteria   .= " AND s.CenterID =:cid ";
             $this->query_vars['cid'] = $centerID;
         } else {
-            if ($user->hasPermission('access_all_profiles')) {
+            $currentUser = \User::singleton();
+            if ($currentUser->hasPermission('access_all_profiles')) {
                 $list_of_sites = \Utility::getSiteList();
             } else {
-                $list_of_sites = \User::getStudySites();
+                $list_of_sites = $currentUser->getStudySites();
             }
             $sitesString = implode(",", array_keys($list_of_sites));
 
--- a/modules/statistics/php/statistics_site.class.inc
+++ b/modules/statistics/php/statistics_site.class.inc
@@ -55,6 +55,20 @@ class Statistics_Site extends \NDB_Menu
                 'data_entry',
                 intval($_REQUEST['CenterID'])
             );
+        } else {
+
+            // If no CenterID is passed in the request, check if the user has the
+            // data_entry permission in all the sites/centers it have access to.
+            // TODO: There are no means of set permissions per site
+            // for a given user right now: (e.g.) The user X can have
+            // the permission data_entry on site Y but not on site Z.
+
+            $hasCenterPermission = true;
+            foreach ($user->getStudySites() as $key => $value) {
+                if (!$user->hasCenterPermission('data_entry', intval($key))) {
+                    $hasCenterPermission = false;
+                }
+            }
         }
 
         return $hasAccessToAllProfiles || $hasCenterPermission;
@@ -66,15 +80,27 @@ class Statistics_Site extends \NDB_Menu
      *
      * @param string $centerID  the value of centerID
      * @param string $projectID the value of projectID
+     * @param \User  $user      The user whose access is being checked
      *
      * @return void
      */
-    function _checkCriteria($centerID, $projectID)
+    function _checkCriteria($centerID, $projectID, \User $user)
     {
         if (!empty($centerID)) {
             $this->query_criteria   .= " AND s.CenterID =:cid ";
             $this->query_vars['cid'] = $centerID;
+        } else {
+            if ($user->hasPermission('access_all_profiles')) {
+                $list_of_sites = \Utility::getSiteList();
+            } else {
+                $list_of_sites = $user->getStudySites();
+            }
+            $sitesString = implode(",", array_keys($list_of_sites));
+
+            $this->query_criteria    .= " AND s.CenterID IN (:cids) ";
+            $this->query_vars['cids'] = $sitesString;
         }
+
         if (!empty($projectID)) {
             $this->query_criteria   .= " AND s.ProjectID =:pid ";
             $this->query_vars['pid'] = $projectID;
--- a/modules/statistics/php/statistics_site.class.inc
+++ b/modules/statistics/php/statistics_site.class.inc
@@ -134,7 +134,7 @@ class Statistics_Site extends \NDB_Menu
                     }
                 }
             }
-            $sitesString = implode(",", array_keys($list_of_sites));
+            $sitesString = implode(",", array_keys($list_of_permitted_sites));
 
             $this->query_criteria    .= " AND s.CenterID IN (:cids) ";
             $this->query_vars['cids'] = $sitesString;
--- a/modules/statistics/php/statistics_site.class.inc
+++ b/modules/statistics/php/statistics_site.class.inc
@@ -80,11 +80,10 @@ class Statistics_Site extends \NDB_Menu
      *
      * @param string $centerID  the value of centerID
      * @param string $projectID the value of projectID
-     * @param \User  $user      The user whose access is being checked
      *
      * @return void
      */
-    function _checkCriteria($centerID, $projectID, \User $user)
+    function _checkCriteria($centerID, $projectID)
     {
         if (!empty($centerID)) {
             $this->query_criteria   .= " AND s.CenterID =:cid ";
@@ -93,7 +92,7 @@ class Statistics_Site extends \NDB_Menu
             if ($user->hasPermission('access_all_profiles')) {
                 $list_of_sites = \Utility::getSiteList();
             } else {
-                $list_of_sites = $user->getStudySites();
+                $list_of_sites = \User::getStudySites();
             }
             $sitesString = implode(",", array_keys($list_of_sites));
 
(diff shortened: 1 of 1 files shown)


johnsaigle:
Hi @racostas. Thanks for tackling this issue.

The `phan` tool in Travis is reporting some errors with your code. Basically `_checkCriteria` is a function in the parent class and the signatures for both classes must match (e.g. have the same number and types of parameters.

You'll need to refactor your solution a little bit so that this constraint is satisfied.

Here's a copy of the errors reported by Travis. You can test locally by running `make checkstatic`.

> modules/statistics/php/statistics_dd_site.class.inc:40 PhanParamSignatureMismatch Declaration of function _checkCriteria(string $centerID, string $projectID) : void should be compatible with function _checkCriteria(string $centerID, string $projectID, \User $user) : void defined in modules/statistics/php/statistics_site.class.inc:87
modules/statistics/php/statistics_dd_site.class.inc:40 PhanParamSignatureRealMismatchTooFewParameters Declaration of function _checkCriteria($centerID, $projectID) should be compatible with function _checkCriteria($centerID, $projectID, \User $user) (the method override accepts 2 parameter(s), but the overridden method can accept 3) defined in modules/statistics/php/statistics_site.class.inc:87
modules/statistics/php/statistics_mri_site.class.inc:39 PhanParamSignatureMismatch Declaration of function _checkCriteria(string $centerID, string $projectID) : void should be compatible with function _checkCriteria(string $centerID, string $projectID, \User $user) : void defined in modules/statistics/php/statistics_site.class.inc:87
modules/statistics/php/statistics_mri_site.class.inc:39 PhanParamSignatureRealMismatchTooFewParameters Declaration of function _checkCriteria($centerID, $projectID) should be compatible with function _checkCriteria($centerID, $projectID, \User $user) (the method override accepts 2 parameter(s), but the overridden method can accept 3) defined in modules/statistics/php/statistics_site.class.inc:87
modules/statistics/php/statistics_site.class.inc:143 PhanParamTooFew Call with 2 arg(s) to \LORIS\statistics\statistics_site::_checkCriteria(string $centerID, string $projectID, \User $user) which requires 3 arg(s) defined at modules/statistics/php/statistics_site.class.inc:87
modules/statistics/php/statistics_site.class.inc:173 PhanParamTooFew Call with 2 arg(s) to \LORIS\statistics\statistics_site::_checkCriteria(string $centerID, string $projectID, \User $user) which requires 3 arg(s) defined at modules/statistics/php/statistics_site.class.inc:87

modules/statistics/php/statistics_site.class.inc
```diff
@@ -55,6 +55,20 @@ class Statistics_Site extends \NDB_Menu
...
+            // data_entry permission in all the sites/centers it have access to.
+            // TODO: There are no means of set permissions per site
+            // for a given user right now: (e.g.) The user X can have
+            // the permission data_entry on site Y but not on site Z.
+
+            $hasCenterPermission = true;
+            foreach ($user->getStudySites() as $key => $value) {
+                if (!$user->hasCenterPermission('data_entry', intval($key))) {
+                    $hasCenterPermission = false;
+                }
```
ridz1208: I think it's better here to use the `getCenterIDs()` from the user class.

Also, I'm not sure this does what you intend it to do. It's true currently you can not set per-site permissions so I'm not sure why we are using `hasCenterPermission()` but in the even this feature gets added. A user in this case will be denied access to the entire page if they dont have `data_entry` on a single one of their sites. 

I think a better way of doing this check is to return true if the user has `data_entry` on any of their sites, not all of their sites...

racostas: Hi @ridz1208,

"Also, I'm not sure this does what you intend it to do. It's true currently you can not set per-site permissions so I'm not sure why we are using hasCenterPermission() but in the even this feature gets added. A user in this case will be denied access to the entire page if they dont have data_entry on a single one of their sites" **... hahaha... yes, in what I was thinking...**

I my opinion the long term robust solution will be to not allow access if the centerIDs aren't specified during the GET query. This will take a bit more of coding and a slightly modification on the link that calls this page. We could pass an array of centerIDs instead.

CenterIDs

For the **short term solution** I would propose the following:

Let's say we leave the` hasCenterPermission()` function with the name it has now as **the semantic** (I didn't developed this one but I think) is the one we're searching for: **to check if a user have permissions for a given site.** Even if its not the way it is implemented right now... we could embed the logic inside after, when the permission framework ready. Also not sure across how many pages this function is used as it seems to be a core one.

During the filtering, in `_checkCriteria` we could incorporate the logic to only get the sites in which the user have the **'data_entry'** permission using the function `hasCenterPermission()`, **which for the moment will do not what is intended but will don't break it neither.**  

Then, once we implemented the correct logic inside `hasCenterPermission()` the system should be compatible and effectively filtering.

Regarding `getCentersIDs()` and `getStudySites()` the main difference I found so far is that `getStudySites()` is checking if the center is flagged as a study one? ... I'm not familiar with this two functions or the business logic they encapsulated... any light in this sense is more than welcome. 

Thanks,
Rolando (CHU Sainte-Justine)

ridz1208: assuming we keep `hasCenterPermission`, there is no reason to block access to a user who has `data_entry` on only one of their `n` sites...

I think the only check here should be, does the user have data_entry on any of their sites ? if yes, allow access and filter the data below (already in this PR); if no, then block access.

I'm reading my proposition from above

> In this particular case, what I would do is remove the if statement on line 53 simply do a check if the user has data_entry allow access to the page, otherwise block access. the filtering below will make sure user does not have access to the incorrect data

johnsaigle:
Travis failing due to PHPCS errors:
>  50 | ERROR   | [x] Whitespace found at end of line
  52 | ERROR   | [x] Whitespace found at end of line
  61 | ERROR   | [x] Whitespace found at end of line
  70 | ERROR   | [x] Whitespace found at end of line
 100 | ERROR   | [x] Whitespace found at end of line
 109 | ERROR   | [x] Whitespace found at end of line
 116 | ERROR   | [ ] You must use "/**" style comments for a function comment
 124 | ERROR   | [x] Whitespace found at end of line
 129 | WARNING | [ ] Line exceeds 85 characters; contains 93 characters

and also due to unstable DICOM test (fixed by #5719):
>1) DicomArchiveTestIntegrationTest::testdicomArchiveFilterClearBtn
UnexpectedJavascriptException: TypeError: input is null

PapillonMcGill:
> 128 | WARNING | Line exceeds 85 characters; contains 93 characters
almost there!

racostas:
Thanks for the comments @PapillonMcGill ! Are very appreciated!!

I have two questions regarding this in particular:

Are "Warnings" required for Travis not to fail?

This is the code line that gives the warning:
`if ($currentUser->hasCenterPermission('data_entry', intval($centerID))) {`

I think the indentation is also counting as part of the characters Travis is seeing as let's say "net weight" is 74 characters.

I could split it into two lines without issues but not sure will gain readability or lost it?

Thanks,
Rolando (CHU Sainte-Justine)

johnsaigle:
Hey @racostas, yes right now Travis is configured to fail on PHPCS warnings. 

Normally we'd write that line like this:

```php
if ($currentUser->hasCenterPermission(
    'data_entry', 
    intval($centerID))
) {
```

it's not the prettiest but it keeps things from getting too long.

racostas:
"Help!, I don't understand Travis!" ... now I'm getting warnings in files I didn't touch in this PR.

johnsaigle:
> modules/statistics/php/statistics_site.class.inc:137 PhanUndeclaredVariable Variable $list_of_sites is undeclared

This is the error I'm seeing, which is in the file you're changing.

racostas:
image

johnsaigle:
These are eslint errors and those are allowed by Travis; they won't cause your build to fail. PHPCS formatting errors are different.

The error I pasted above is further down in the build log and is caused by a third tool called `phan`. That's the issue that's causing the failure and needs to be fixed.

racostas:
**line 156:**  `... $this->instruments = \Utility::getAllInstruments();...`
**line 265:** `... foreach ($this->instruments as $instrument=>$label) { ...`

As the page is built going from the instruments to the computation an user with permission in one of the sites is been able to see **"The Name"** of instruments it doesn't have in the center it belongs to. This is not let's say "extremely bad" as the data for those will be not computed but maybe in the future we want to restrict only to show those instruments in the center/centers the user have access rights.

Something like:

`$intrumentsToShow = $this->instruments->getInCenterInstruments($user->getCenters());` **--- TO BE DISCUSSED as other design approach could be possible to. NOTE: `getInCenterInstruments()` is not an existing function at the moment this is being written**

racostas:
Hi LORIS team,

I found a bug during manual testing in the solution I give here in this PR. Currently working in a fix. 

The issue seems to be related with the way I'm using the function **pselect** from the class **Database**.

Please add the "Need Work" label and don't release it until fix it.

Thanks,
Rolando Acosta (CHU Sainte-Justine)

PapillonMcGill:
Hello @racostas This PR has already been merged. You will need to open a new PR to fix this one.

racostas:
The following issue is addressed on #5950 

> Hi LORIS team,
> 
> I found a bug during manual testing in the solution I give here in this PR. Currently working in a fix.
> 
> The issue seems to be related with the way I'm using the function **pselect** from the class **Database**.
> 
> Please add the "Need Work" label and don't release it until fix it.
> 
> Thanks,
> Rolando Acosta (CHU Sainte-Justine)

PR #5596, opened 2019-11-11
[RB] Add SNPs to fix orphan fields from SNP_candidate_rel

ridz1208:
## Brief summary of changes
SNP data somehow got partially deleted from RB. the SNP_candidate_rel data was still there but the SNP table was empty


ridz1208:
@xlecours all RB_files are independent inserts, it's easier for github to show diffs like that.

when i tried single statements inserts , any change to any row shows up as if the entire file changes, or the entire group of insters chnaged when SQLdump decides to group them together.

ridz1208:
this file is generated by the `tools/exporters/DB_dump_table_data.php` script, so any change should be done there and all RB files regenerated

xlecours:
Ok, you are the expert.
It's not a big deal for a file of that size.

ridz1208:
@driusan this doesnt need manual testing, travis does the RB loading

PR #5603, opened 2019-11-11
[Install] System requirements state MySQL >= 5.7 but in fact MySQL 8 will run into trouble

lingma:
System requirements statement is not precise.

## Brief summary of changes

MySQL 5.7 or 8 both runs fine with the program, except that some configuration should be modified on MySQL 8 to make the system work.

1. Install MySQL, then run the installation script, the installation will fail at the step 6. In fact, the note I added could go to the step 6 either, without much difference.

*

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,11 +33,6 @@ Deploy and log in with username *admin* and the password that's set up during de
 
 These dependencies are subject to change so be sure to verify your version of MySQL and PHP when updating LORIS. Installing some dependencies may require `sudo` privileges.
 
-Note: the default password plugin of MySQL 8 is caching_sha2_password instead of mysql_native_password. The most PHP versions don't support the caching_sha2_password, you can work around the problem by modifying the 
-default_authentication_plugin=mysql_native_password in the section [mysqld] of my.cnf, or modify that of one particular database user with a command similar to the following:
-alter user '<user>'@'localhost' identified with mysql_native_password by '<password>';
-This solution is temporary, pending the program upgrading to support caching_sha2_password.
-
 ### Install Steps
 
 Consult the [LORIS Wiki](https://github.com/aces/Loris/wiki/Setup) page on this [Install process](https://github.com/aces/Loris/wiki/Installing-Loris) for more information.
@@ -87,6 +82,8 @@ If your apache configuration was not completed by the Install script, run the fo
     sudo a2ensite $projectname
     sudo service apache2 reload
 6. Go to http://localhost/installdb.php and follow the instructions to finalize LORIS installation, then restart apache.
+
+If you use MySQL 8, please read [this link](https://www.php.net/manual/en/mysqli.requirements.php) and also [this](https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password-compatible-connectors).
 
 7. Follow the [Setup Guide in the LORIS Wiki](https://github.com/aces/Loris/wiki/Setup) to complete your post-installation setup and configuration, and for more documentation.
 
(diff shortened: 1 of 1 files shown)


johnsaigle:
This seems to be an issue with PHP and MySQL compatibility rather than with LORIS. So I don't think this is really an issue with our documentation per se.

How would you feel about a small amendment to step 6, something like "If using MySQL v8.00 or above, please see [this link](https://www.php.net/manual/en/mysqli.requirements.php)."

I'd prefer this approach because:
* this is not really our issue, I think it's preferable to not devote so much space to it on our README (since it effectively doubles as a landing page for people checking out the software)
* linking to the official PHP support page means that that link should stay up-to-date over time, reducing our maintenance burden

lingma:
I am OK with the suggestion, I feel it is important to mention something somewhere, I don't know where. Also I found that the php documentation is not up to date either, 
I tried mysql_xdevapi, which doesn't seem working. Maybe add an additional link:
https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password-compatible-connectors

lingma:
I verified, loris using PDO, which has nothing to do with mysql_xdevapi, that was why it doesn't work for me.

lingma:
The new authentication method should be used with MySQL 8, Loris uses the old library based on PDO of PHP 5, this need an improvement later to support MySQL 8 instead of downgrading authentication method. I tried to figure out a method some time gao, but not yet found. Since this is not urgent, I put it aside for now.

johnsaigle:
> The new authentication method should be used with MySQL 8, Loris uses the old library based on PDO of PHP 5, this need an improvement later to support MySQL 8 instead of downgrading authentication method

Can you explain what you mean by this? I'm not familiar enough with PDO to know why it would be considered old or out of date.

lingma:
PHP: the PDO_MySQL and ext/mysqli extensions do not support caching_sha2_password. In addition, when used with PHP versions before 7.1.16 and PHP 7.2 before 7.2.4, they fail to connect with default_authentication_plugin=caching_sha2_password even if caching_sha2_password is not used. (1) 

Loris uses php/libraries/Database.class.inc to connect to database, which states clearly PHP Version 5 at the version declaration lines.

The caching_sha2_password and sha256_password authentication plugins provide more secure password encryption than the mysql_native_password plugin, and caching_sha2_password provides better performance than sha256_password. Due to these superior security and performance characteristics of caching_sha2_password, it is as of MySQL 8.0 the preferred authentication plugin, and is also the default authentication plugin rather than mysql_native_password. (1)

We should start thinking to change PDO to a better solution to improve security. I tried to find a solution, but I consider that we need first make a ticket if necessary, then see when to do it.

source:

1). https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password-compatible-connectors

johnsaigle:
> Loris uses php/libraries/Database.class.inc to connect to database, which states clearly PHP Version 5 at the version declaration lines.

For the record this is out of date and should say PHP version 7, not 5. All of LORIS requires at least PHP 7.2 to run but some of the documentation is still out of date, unfortunately.

As for the rest, I'm definitely in favour of finding a way to improve MySQL security if we can do it. We would need to find a way that is backward-compatible as well because at this time I do think it's our intention to support older MySQL versions. For example, most of our internal servers as far as I know are on MySQL 5.7.x (for better or for worse).

Ideally PHP will support this soon; this would be a lot better than rewriting the guts of the Database class.

johnsaigle:
https://bugs.php.net/bug.php?id=76243

At the bottom of this conversation it says that this caching support will be added in PHP 7.4 which will be released later this month.

For now, it's probably best to add this note to the README as you've done here. Ideally we'll [require 7.4 in LORIS](#5063) in one or two releases because we'd also like to take advantages of other features added in that release. At that time we'll get the security + performance boost for free.

PR #5614, opened 2019-11-12
[User Accounts] Update password strength rules documentation

johnsaigle:
## Brief summary of changes

Update user accounts module to properly display the new password strength guidelines. These match the ones on the password expiry form in the login module.

### DIFF
--- a/modules/user_accounts/templates/form_edit_user.tpl
+++ b/modules/user_accounts/templates/form_edit_user.tpl
@@ -36,13 +36,15 @@
     {/if}
 
     <div class="panel panel-default">
-      <div class="panel-body">
+  <div class="panel-body">
        <h3>Password Rules</h3>
-       <ul>
-          <li>The password must be at least 8 characters long</li>
-          <li>The password must contain at least 1 letter, 1 number and 1 character from   !@#$%^&amp;*()</li>
-          <li>The password and the user name must not be the same</li>
-          <li>The password and the email address must not be the same</li>
+      <ul>
+        <li>The password must be at least 8 characters long.</li>
+        <li>The password and the user name must not be the same.</li>
+        <li>The password and the email address must not be the same.</li>
+      </ul>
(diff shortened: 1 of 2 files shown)


modules/user_accounts/templates/form_edit_user.tpl
```diff
@@ -36,13 +36,15 @@
...
+  <div class="panel-body">
        <h3>Password Rules</h3>
-       <ul>
-          <li>The password must be at least 8 characters long</li>
-          <li>The password must contain at least 1 letter, 1 number and 1 character from   !@#$%^&amp;*()</li>
-          <li>The password and the user name must not be the same</li>
-          <li>The password and the email address must not be the same</li>
+      <ul>
+        <li>The password must be at least 8 characters long.</li>
+        <li>The password and the user name must not be the same.</li>
```
maltheism: Optional suggestion would be "The password and the username or email address must not be the same." instead of repeating similar again.

modules/user_accounts/templates/form_my_preferences.tpl
```diff
@@ -1,12 +1,14 @@
...
-        <li>The password and the email address must not be the same</li>
-    </ul>
+      <ul>
+        <li>The password must be at least 8 characters long.</li>
+        <li>The password and the user name must not be the same.</li>
+        <li>The password and the email address must not be the same.</li>
+      </ul>
+        <p><b>Please choose a unique password.</b></p>
+        <p>We suggest using a password manager to generate one for you.</p>
+      </ul>
```
driusan: Your tags aren't matched up.. you're closing 2 `<ul>` tags but only opening one. Styling should also be done with CSS instead of presentational tags (ie. `<b>` shouldn't be used..)

johnsaigle: Alrighty, I just copied from the login module so that should maybe be fixed at some point

PR #5615, opened 2019-11-12
[jsx/Markdown.js] Markdown feature bullets

maltheism:
## Brief summary of changes

Adds Markdown "bullets" which is the equivalent to unordered html list. 

* First implementation of the bullets to the unordered html list (feature). I couldn't think of a way to do this with regex.
* I'm not happy with the first attempt because it only works with single list. Next implementation need to make the function recursive for lists within a list. Although I thought it would be good to send a PR for now. 

1. Make a markdown file inside help for one of the modules with bullets.
2. View on the frontend.

* https://github.com/aces/Loris/issues/5259

### DIFF
--- a/jsx/Markdown.js
+++ b/jsx/Markdown.js
@@ -29,9 +29,39 @@ class Markdown extends Component {
       .replace(/&quot;/g, '"');
   }
 
+  bulletsHandler(lines) {
+    let bulletFlag = false;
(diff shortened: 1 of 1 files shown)


maltheism:
Closing in favour of PR https://github.com/aces/Loris/pull/5881

PR #5616, opened 2019-11-12
[instrument_builder] Blank line fix

### LINKED ISSUE #4604: [instrument_builder] Formatting Blank Line is not inserting
Adding a Blank Line is not inserting

### LINKED ISSUE #5486: [instrument builder] Add "Blank Line" does not work
Test plan Item 1
1.j Add Blank Line

In instrument builder, nothing changes when you add a "Blank Line" field. Keeping adding field, there is no evidence showing the blank line.

If you download the linst file, there is no blank line in the file.

maltheism:
## Brief summary of changes

See #https://github.com/aces/Loris/issues/5486

1. Checkout PR.
2. Replicate the https://github.com/aces/Loris/issues/5486 gif that @coreyAbout made.

* https://github.com/aces/Loris/issues/5486
* Resolves #4604
* Resolves #5486

### DIFF
--- a/modules/instrument_builder/js/instrument_builder.instrument.js
+++ b/modules/instrument_builder/js/instrument_builder.instrument.js
@@ -90,7 +90,7 @@ var Instrument = {
             for (element of elements[i].Elements) {
                 switch (element.Type) {
                     case "line":
-                        content += '<br>\n';
+                        content += 'static{@}{@}<br>\n';
                     case "select":
                         if (element.Options.AllowMultiple) {
                             content += "selectmultiple{@}"
--- a/modules/instrument_builder/js/instrument_builder.instrument.js
+++ b/modules/instrument_builder/js/instrument_builder.instrument.js
@@ -88,10 +88,8 @@ var Instrument = {
                 content += "page{@}{@}" + elements[i].Description + "\n";
             }
             for (element of elements[i].Elements) {
+              console.log(element.Type);
                 switch (element.Type) {
-                    case "line":
-                      content += 'line{@}{@}' + element.Description + "\n";
-                        break;
                     case "select":
                         if (element.Options.AllowMultiple) {
                             content += "selectmultiple{@}"
--- a/modules/instrument_builder/js/instrument_builder.instrument.js
+++ b/modules/instrument_builder/js/instrument_builder.instrument.js
@@ -89,6 +89,8 @@ var Instrument = {
             }
             for (element of elements[i].Elements) {
                 switch (element.Type) {
+                    case "line":
+                        content += '<br>\n';
                     case "select":
                         if (element.Options.AllowMultiple) {
                             content += "selectmultiple{@}"
--- a/modules/instrument_builder/js/instrument_builder.instrument.js
+++ b/modules/instrument_builder/js/instrument_builder.instrument.js
@@ -90,7 +90,7 @@ var Instrument = {
             for (element of elements[i].Elements) {
                 switch (element.Type) {
                     case "line":
-                        content += 'static{@}{@}<br>\n';
+                      content += 'line{@}{@}' + element.Description + "\n";
                         break;
                     case "select":
                         if (element.Options.AllowMultiple) {
@@ -298,6 +298,14 @@ var Instrument = {
                                 value: "Header"
                             };
                             break;
+                      case "line":
+                          tempElement.Type = 'line';
+                          tempElement.Description = pieces[2];
+                          tempElement.selected = {
+                            id: 'line',
+                            value: 'Blank Line'
+                          };
+                          break;
                         default:
                             break;
                     }
--- a/modules/instrument_builder/js/instrument_builder.instrument.js
+++ b/modules/instrument_builder/js/instrument_builder.instrument.js
@@ -91,6 +91,7 @@ var Instrument = {
                 switch (element.Type) {
                     case "line":
                         content += 'static{@}{@}<br>\n';
+                        break;
                     case "select":
                         if (element.Options.AllowMultiple) {
                             content += "selectmultiple{@}"
--- a/modules/instrument_builder/js/instrument_builder.instrument.js
+++ b/modules/instrument_builder/js/instrument_builder.instrument.js
@@ -90,6 +90,10 @@ var Instrument = {
             for (element of elements[i].Elements) {
               console.log(element.Type);
                 switch (element.Type) {
+                    case "line":
+                      console.log('test');
+                      content += 'line{@}{@}' + element.Description + "\n";
+                      break;
                     case "select":
                         if (element.Options.AllowMultiple) {
                             content += "selectmultiple{@}"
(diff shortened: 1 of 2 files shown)


coreyAbout:
Switched to your branch, run make dev, and not able to login: error:
`This page isn’t working cobra-lab-dev.loris.ca is currently unable to handle this request.
HTTP ERROR 500`

modules/instrument_builder/jsx/react.questions.js
```diff
@@ -502,6 +502,10 @@ class ListElements extends Component {
     let textSize = 'small';
     // Set the options for the desired type
     switch (newId) {
+      case 'line':
+        newState.Options = {};
+        newState.Name = '';
+        newState.Description = '';
```
driusan: There was a "blank line" option, why does this now need a linst spec change to work?

maltheism:
@coreyAbout I'm not sure why you're getting that error. I think it's unrelated to the PR because only a few lines of the code are changed in the instrument_builder of this PR.

modules/instrument_builder/jsx/react.questions.js
maltheism: @driusan when debugging the bug I encountered that the previous state of Options, Name, Description were getting carried over when creating a 'break line' and setting them to default values would allow the 'break line' to work.

driusan: So does this change anything in the saved file? Also, does this case need a `break`?

maltheism: @driusan forgot to test that. I'll check and report back.

maltheism: @driusan seems to not being adding to .linst file. I'll debug some more and fix. As of now just the frontend error is fixed for rendering the break lines.

driusan: What about the `break` statement part? Right now this is falling through to the textarea/textbox cases (though I doubt that's the cause of it not saving properly..)

maltheism: @driusan nice catch on the missing break. That part of the code makes the Blank Line render on the frontend of the instrument builder. Otherwise before the code change, it wasn't rendering the Blank Line to the frontend. The new code I just added fixes the save to a .linst file.

modules/instrument_builder/js/instrument_builder.instrument.js
```diff
@@ -89,6 +89,8 @@ var Instrument = {
             }
             for (element of elements[i].Elements) {
                 switch (element.Type) {
+                    case "line":
+                        content += '<br>\n';
```
driusan: The format for static content (including br tags) needs to be like the format of "label", including the static type and (empty) fieldname. I don't think this will load properly in LORIS as written. It needs to be

`static{@}{@}<br>\n`

modules/instrument_builder/jsx/react.questions.js
christinerogers: nice catches.  Do we also need a change to an existing linst spec file, @driusan ?  if so, let's tag update or tag as `needs documentation`

christinerogers: Can you confirm @maltheism @coreyAbout  - 
Does this reset (wipe) the last-used Options array?  
It should keep the last options for the next (e.g. dropdown) question added, if at all possible. 
The intended behaviour is that when the user adds `dropdown1:options{a, b, c}` then `blank line` then `dropdown2` -- the options will already be loaded with `{a, b, c}`

driusan:
@kongtiaowang can you test this (including making sure the instrument saved can be loaded by LORIS and renders the blank line)?

maltheism:
@christinerogers I just tested the PR with dropdowns and everything seems to work.

kongtiaowang:
load: function (file, callback) is missing the "blank line" element.

maltheism:
@kongtiaowang please review again and thanks for the help!

modules/instrument_builder/jsx/react.questions.js
```diff
@@ -37,6 +37,8 @@ class LorisElement extends Component {
       case 'label':
         elementHtml = <p>{element.Description}</p>;
         break;
+      case 'line':
+        elementHtml = <div></div>;
```
kongtiaowang: missing break, so it has a tiny render issue.

maltheism: @kongtiaowang thanks, I added it.

modules/instrument_builder/js/instrument_builder.instrument.js
```diff
@@ -89,6 +89,9 @@ var Instrument = {
             }
             for (element of elements[i].Elements) {
                 switch (element.Type) {
+                    case "line":
```
driusan: there is no "line" type saved in a linst file. Why did you add this type? It's always been a static label with a <br> content.

maltheism: @driusan I removed it. I've tested the code and that part is unnecessary. I added it because I thought it was necessary from reviewing @kongtiaowang's past PR on this issue.

kongtiaowang:
@driusan @maltheism 
After the change, when the user load this linst file again, It will show the label tag with title <br> on the page instead of a blank line.

maltheism:
@driusan PR is ready for merge I think.

christinerogers:
@coreyAbout  Could you re-test Monday?  this looks very ready to merge. 
Thanks @maltheism  and @kongtiaowang !

coreyAbout:
OK @christinerogers

coreyAbout:
Looks like more work need to be done. blank line not showing up after instrument install.
@christinerogers @driusan 

Screenshot from 2019-11-25 13_58_53
Screenshot from 2019-11-25 13_56_25

maltheism:
@coreyAbout I think that issue isn't in the instrument_builder

coreyAbout:
@maltheism you might be right. Wait for @driusan to indicate whether to put a new issue or not.

maltheism:
@driusan can this PR get merged. I think the issue @coreyAbout mentioned should be solved in another PR because the issue is outside the instrument_builder module code.

driusan:
Is the feature working in the frontend for previously built instruments that have a blank line?

maltheism:
@driusan I'm not sure if any previous built instruments had the blank line added to the linst file. Since the code never seemed to have had the functionality for adding it to the linst files.

driusan:
There are many in the IBIS repo if you search for `"static{@}{@}<br />"` (with quotes).. if you don't have access to the repo, maybe @sruthymathew123 or @nicolasbrossard can get you access to one.

nicolasbrossard:
@maltheism Would you like me to send you a linst and/or class file containing an example of this?

maltheism:
@nicolasbrossard yes please

maltheism:
@nicolasbrossard @driusan The old linst files don't seem to work with this PR. The result ends up being a label with <br> inside it.

zaliqarosli:
@maltheism any chance you can move this to 23.0-release branch? if that's okay with @driusan . this PR is needed for me to pass my testing of the instrument builder

driusan:
This PR isn't going in to the 23 release. We aren't changing the LINST spec.

zaliqarosli:
@driusan even if it's a bug fix for failed testing?

driusan:
this is not a bug fix, this is a complete change of behaviour.

driusan:
Replaced by #6859

PR #5619, opened 2019-11-12
[Core/Acknowledgments] Create generic way to check module permissions using integration tests

johnsaigle:
#### Issue

We've had many issues reported lately where module `X` doesn't load when a user has only the `X_edit` permission without the `X_view` permission. 

In almost every case, it should be possible to at least view a module when you have any of its permissions.

This is trivial to check with integration tests, but many modules are lacking permission checking or they are out of date with respect to new permissions added to a module.

#### Solution
This PR aims to address both of these by creating a generic way to check module permissions using integration tests by adding new methods to the parent class that can validate permissions in a standard way.

All module integration tests going forward can use this idiom to simply and concisely validate that a module loads IFF the user has one of the permission codes for that module.

This should also reduce the size of individual integration tests by eliminating boilerplate.

The LorisIntegrationTest class now also conforms to our PHPCS config.

* This PR builds on the work of @cmadjar's PR #5584. A new permission code was added here but the integration tests didn't properly test the permissions. This is fixed by this PR.

### DIFF
--- a/modules/acknowledgements/php/acknowledgements.class.inc
+++ b/modules/acknowledgements/php/acknowledgements.class.inc
@@ -28,7 +28,11 @@ namespace LORIS\acknowledgements;
 
 class Acknowledgements extends \NDB_Menu_Filter_Form
 {
-    public $skipTemplate = true;
+    public const PERMISSIONS = array(
+        'acknowledgements_view',
+        'acknowledgements_edit'
+    );
+    public $skipTemplate     = true;
 
     /**
      * Determine whether user should be granted access.
@@ -39,8 +43,7 @@ class Acknowledgements extends \NDB_Menu_Filter_Form
      */
     function _hasAccess(\User $user) : bool
     {
-        return ($user->hasPermission('acknowledgements_view')
-                ||$user->hasPermission('acknowledgements_edit'));
+        return $user->hasAnyPermission(self::PERMISSIONS);
     }
 
     /**
--- a/modules/acknowledgements/php/acknowledgements.class.inc
+++ b/modules/acknowledgements/php/acknowledgements.class.inc
@@ -28,11 +28,7 @@ namespace LORIS\acknowledgements;
 
 class Acknowledgements extends \NDB_Menu_Filter_Form
 {
-    public const PERMISSIONS = array(
-        'acknowledgements_view',
-        'acknowledgements_edit'
-    );
-    public $skipTemplate     = true;
+    public $skipTemplate = true;
 
     /**
      * Determine whether user should be granted access.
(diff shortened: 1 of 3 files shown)


test/integrationtests/LorisIntegrationTest.class.inc
```diff
@@ -110,11 +110,14 @@ abstract class LorisIntegrationTest extends TestCase
         $user->updatePassword($password);
 
         $dev          = $this->config->getSettingFromXML('dev');
-        $selenium_url = isset($dev['selenium_url']) ? $dev['selenium_url'] : 'http://localhost:4444/wd/hub';
+        $selenium_url = $dev['selenium_url']) 'http://localhost:4444/wd/hub';
```
maltheism: line 113 looks to be missing ??

johnsaigle: Thanks, weird that that didn't get caught by `make checkstatic`.

modules/acknowledgements/php/acknowledgements.class.inc
```diff
@@ -28,7 +28,11 @@ namespace LORIS\acknowledgements;
 
 class Acknowledgements extends \NDB_Menu_Filter_Form
 {
-    public $skipTemplate = true;
+    public const PERMISSIONS = array(
```
driusan: This class should not be changed if you're just adding tests. This const is useless.

johnsaigle:
@driusan Can you re-review? This PR is holding up others which have now grown stale.

test/integrationtests/LorisIntegrationTest.class.inc
```diff
@@ -556,5 +573,87 @@ abstract class LorisIntegrationTest extends TestCase
...
+            $this->_verifyPermission(
+                $successMessage,
+                $this->_loadWithPermission($path, $code)
+            );
+        }
+    }
+
+    /**
+     * Asserts that a message appears in the HTML body of the page.
+     * Rests the permissions checked to cleanup before further tests.
```
PapillonMcGill:
```suggestion
     * Resets the permissions checked to cleanup before further tests.
```

test/integrationtests/LorisIntegrationTest.class.inc
```diff
@@ -556,5 +573,87 @@ abstract class LorisIntegrationTest extends TestCase
...
+     * @return void
+     */
+    private function _verifyPermission(string $needle, string $haystack): void
+    {
+        $this->assertContains($needle, $haystack);
+        $this->resetPermissions();
+    }
+
+    /**
+     * Helper function to set up and load a page with a given permisison code.
```
PapillonMcGill:
```suggestion
     * Helper function to set up and load a page with a given permission code.
```

johnsaigle:
@driusan Travis has passed. Thanks for reviewing.

Issue #5469, opened 2019-10-31
[Brainbrowser] Minor UI improvement on Overlay Panel


ridz1208:
The overlay control panel does not have a border like the remainder of the panels on the same page

see examples
![Screenshot from 2019-10-31 16-58-36](https://user-images.githubusercontent.com/15268287/67985747-e93f9000-fbff-11e9-837f-163e5a459f9d.png)
![Screenshot from 2019-10-31 16-58-26](https://user-images.githubusercontent.com/15268287/67985751-eb095380-fbff-11e9-8913-2ad09a89438a.png)

driusan:
PR merged

Issue #5474, opened 2019-10-31
[brainbrowser] Loading slowness and lack of clarity


ridz1208:
When brainbrowser is loading multiple volumes, after the first volume loads, the second takes a bit of time (sometimes longer 15-20 seconds, sometimes faster 2-4 seconds)

Also There is no clear sign that a second volume is loading (no `LOADING...` message loading under the first volume)

example [link](https://test-loris-dev.loris.ca/brainbrowser/?minc_id=[134,135]&overlay=true)

cmadjar:
This one is very weird. Sometimes it loads the 2 volumes + overlay, sometimes it loads only the first one... 
When the 2 volumes + overlay are displayed, the `LOADING...` message appears for all volumes, one after the other. It is only when only 1 volume out of the 2 is displayed in the page. No idea why the other done does not get displayed. 

Try hitting refresh on the link attached and you will see what I mean. Sometimes, it loads everything, sometimes it loads only the first volume...

No idea why this is doing that or how to debug this... :(

driusan:
PR merged to fix this

PR #5625, opened 2019-11-13
[brainbrowser] Fix overlay control panel display

cmadjar:
## Brief summary of changes

This fixes the UI around the Overlay control panel display of BB.

Before:
Screen Shot 2019-11-13 at 10 11 48 AM

Now:
Screen Shot 2019-11-13 at 10 12 36 AM

1. Load BB with the overlay and check that the overlay control panel has the same border as the other control panels

* #5469

### DIFF
--- a/modules/brainbrowser/templates/form_brainbrowser.tpl
+++ b/modules/brainbrowser/templates/form_brainbrowser.tpl
@@ -11,41 +11,43 @@
         <div class="filename"></div>
         <h3 class="panel-title filename">Overlay</h3>
       </div>
-    </div>
+      <div class="panel-body">
+        <div class="volume-viewer-controls volume-controls .coords">
+          <div class="coords">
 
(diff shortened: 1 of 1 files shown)


johnsaigle:
Hey I'd like to test this but I don't have much experience with BB.

Here's what I have in my VM (on 22.0-release) when looking at an MRI file.

Where do I find the overlay element you've edited?

Screen Shot 2019-11-13 at 10 46 20

cmadjar:
@johnsaigle to get the overlay you will need to select 2 files using the checkboxes next to the filenames in the imaging browser view session page and then click on the button "3D + overlay" in the side panel in blue. This will start BB on the 2 files and add the overlay of the 2 files at the bottom. You might need to refresh a few times BB if you only see one file displayed. There is a weird bug that I can't figure out how to fix (#5474)

PR #5627, opened 2019-11-13
Add PHP minimum version badge to README

johnsaigle:
Screen Shot 2020-01-13 at 16 53 16

The badge version info gets pulled from our master branch on Travis so it's auto-updating.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 <img src="./htdocs/images/LORIS_logo.svg" width="35%">
 
 [![Build Status](https://travis-ci.org/aces/Loris.svg?branch=master)](https://travis-ci.org/aces/Loris) [![Documentation Status](https://readthedocs.org/projects/acesloris/badge/?version=latest)](https://acesloris.readthedocs.io/en/latest/?badge=latest)
+[![Minimum PHP Version](https://img.shields.io/travis/php-v/aces/loris/22.0-release?color=787CB5)](https://php.net/)
 
 # LORIS Neuroimaging Platform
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,4 +1,8 @@
-# [![Build Status](https://travis-ci.org/aces/Loris.svg?branch=master)](https://travis-ci.org/aces/Loris) LORIS Neuroimaging Platform 
+[![Build Status](https://travis-ci.org/aces/Loris.svg?branch=master)](https://travis-ci.org/aces/Loris) 
+[![Minimum PHP Version](https://img.shields.io/travis/php-v/aces/loris/22.0-release?color=787CB5)](https://php.net/)
```
PapillonMcGill: should this point to branch master just like status? If pointing to 22.0, it might have to be updated for each (major) release of LORIS.

driusan:
I don't see what purpose this badge servers or what it has to do with the 22.0 release.

johnsaigle:
I think it's helpful to communicate which versions of PHP we support.

It doesn't need to go into this release.

driusan:
The README already says what version of PHP is required

johnsaigle:
You're right. 

I think one day down the line (with ReadTheDocs) it would be good to have the install section be something like this instead of what it is now:

> ## Installation
Please visit our documentation at www.acesloris.readthedocs.org/... for install instructions for [Ubuntu](url) and [CentOS](url).

Especially given that our install process is pretty extensive.

Anyway I think this looks nice and draws the eye which is handy because this is basically the landing page for the software.

PR #5635, opened 2019-11-13
Travis parallelization (performance boost)

johnsaigle:
## Brief summary of changes

This PR uses caching + parallelization to reduce Travis testing time:

Screen Shot 2020-01-15 at 12 13 14

Travis will now also show an error status on GitHub as soon as any one of the jobs fails. However, other builds will still complete.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,20 +28,10 @@ jobs:
         - stage: "Tests"
           name: "Static Analysis"
           script: make checkstatic
-          php: "7.2"
           php: "7.3"
-          php: "7.4snapshot"
         - script: test/dockerized-unit-tests.sh
           name: "Unit Tests"
-          php: "7.2"
           php: "7.3"
-          php: "7.4snapshot"
         - script: test/dockerized-integration-tests.sh
           name: "Integration Tests"
-          php: "7.2"
           php: "7.3"
-          php: "7.4snapshot"
-  allow_failures:
-    - php: "7.4snapshot"
-  fast_finish: true
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,14 @@ sudo: required
 
 language: php
 
+matrix:
+  include:
+  - php: "7.2"
+  - php: "7.3"
+  - php: "7.4snapshot"
+  allow_failures:
+  - php: "7.4snapshot"
+
 services:
 - docker
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,3 +28,8 @@ install:
 - make dev
 
 script: "npm run tests:$TEST_SUITE"
+
+# This should improve build time by caching composer files.
+cache:
+  directories:
+    - $HOME/.composer/cache/files
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,14 @@ sudo: required
 language: php
 
 php: 
-- "7.2"
-- "7.3"
-- "nightly"
+- 7.2
+- 7.3
+- nightly
+
+matrix:
+    fast_finish: true
+    allow_failures:
+        - php: nightly
 
 services:
 - docker
@@ -23,20 +28,10 @@ before_script:
 jobs:
     include:
         - stage: "Tests"
+          php: 7.3
           name: "Static Analysis"
           script: make checkstatic
-          php: 
-            - "7.3"
-            - "nightly"
         - script: test/dockerized-unit-tests.sh
           name: "Unit Tests"
-          php: 
-            - "7.3"
-            - "nightly"
         - script: test/dockerized-integration-tests.sh
           name: "Integration Tests"
-          php: 
-            - "7.3"
-            - "nightly"
-    allow_failures:
-          php: "nightly"
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,10 @@ php:
   - 7.3
   - "nightly"
 
+matrix:
+    - allow_failures:
+        - php: "nightly"
+
 env:
   - TEST_SUITE=static
   - TEST_SUITE=unit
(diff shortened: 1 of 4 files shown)


maltheism:
@johnsaigle GitHub isn't showing the travis status anymore for this PR. I don't know why 🤨

johnsaigle:
@maltheism Yeah I broke things 🤷‍♂Still a WIP

johnsaigle:
Also once this is on the same branch as the webpack caching, we should get even better performance out of the integration tests. #5386

Makefile
```diff
@@ -1,14 +1,14 @@
 .PHONY: clean dev all check checkstatic unittests test phpdev javascript
 
 all: VERSION javascript
-	composer install --no-dev
+	composer install --no-dev --prefer-dist
```
johnsaigle: This can prevent caching errors on Travis. It'll try to download entire GitHub repositories unless this flag is set.

.travis.yml
```diff
@@ -1,31 +1,35 @@
...
+  - TEST_SUITE=integration
 
 services:
 - docker
 
 install:
 - sudo apt-get update
 - sudo apt-get install -y libapparmor1
-- pecl install -f ast-1.0.1
+- pecl install -f ast-1.0.3
```
johnsaigle: This fixes PHP 7.4 build errors.

Dockerfile.test.php7.debug
```diff
@@ -1,9 +1,9 @@
-FROM php:7.0
+FROM php:7.2
```
johnsaigle: This file was broken.

test/dockerized-integration-tests.sh
```diff
@@ -1,12 +1,11 @@
...
-if [ "$DEBUG" == "true" ]
+if [ -v DEBUG ];
 then
     CONTAINER=integration-tests-debug
 else
     CONTAINER=integration-tests
 fi
 
-docker-compose run -T --rm ${CONTAINER} vendor/bin/phpunit --configuration test/phpunit.xml --testsuite LorisCoreIntegrationTests $*
-docker-compose run -T --rm ${CONTAINER} vendor/bin/phpunit --configuration test/phpunit.xml --testsuite LorisModuleIntegrationTests $*
```
johnsaigle: local integration rests were run twice I guess? 😂

maltheism:
It would be interesting to know how much time in comparison will be saved with all these fixes compared to before. @johnsaigle this PR is exciting. 🙂

johnsaigle:
I think it'll be helpful! I just realized that my initial screenshot (since deleted) may not have been showing the integration test time properly... I thought it was too good to be true...

Definitely this should help our workflow out though. PRs should be "Pending" for nearly as long.

ridz1208:
I like this, will look more into it

driusan:
Why is this going to the 22.0 release branch? I don't see what this has to do with the release

johnsaigle:
I don't mind which branch it goes to. On the other hand if we merge it it should reduce the time PRS spend in the "Pending" status which should accelerate the code review process.

Let me know if you'd prefer me to rebase it.

johnsaigle:
It's kind of a "Meta" thing anyway. it shouldn't impact the release one way or the other.

johnsaigle:
@driusan I have rebased it to master.

Also Travis will run up to **200** jobs for both public and private repositories (see light blue box on this page: https://docs.travis-ci.com/user/build-matrix/).

Sometimes Travis will throttle builds depending on the global load on their servers, e.g. when both EU and NA are active in the evening then Travis runs more slowly for everybody.

Nevertheless we should be able to run far more than 3 on average. While creating this PR I've definitely seen 5 running at the same time.

We can also discuss taking advantage of stages during testing. An example would be to run all of the static checks for each version and proceeding with unit tests and integration tests IFF the static checks pass. This would greatly speed up the CI report.

driusan:
Screenshot of Travis-ci.com

(IIRC it used to be 5, so it's possible we're grandfathered in, but we certainly don't get 200 concurrent jobs.)

johnsaigle:
Hm maybe the page I linked means you can set up 200 total jobs but not that they run concurrently? Still it's misleading.

johnsaigle:
The failure is caused by existing issues in our unit tests and integration tests. These are resolved by #5571.

johnsaigle:
#5571 is now merged into the release branch. This PR should pass Travis when the release code gets merged into `major`.

johnsaigle:
@ridz1208 @driusan This is ready for re-review when you have time.

PapillonMcGill:
LGTM

johnsaigle:
@ridz1208 @driusan ping

Makefile
driusan: What "caching errors"? We shouldn't be changing our build process for a dubious optimization.

johnsaigle: Sorry for the imprecise comment there.

This is what the flag does

>--prefer-dist: Reverse of --prefer-source, Composer will install from dist if possible. **This can speed up installs substantially on build servers** and other use cases where you typically do not run updates of the vendors. It is also a way to circumvent problems with git if you do not have a proper setup.
 https://getcomposer.org/doc/03-cli.md (emphasis mine)

I saw it recommended somewhere that this option is good to set because it can reduce the Travis start up time.

Makefile
```diff
@@ -1,14 +1,14 @@
...
 
 # If anything changes, re-generate the VERSION file
 VERSION: .
 	tools/gen-version.sh
 
 phpdev:
-	composer install
+	composer install --prefer-dist
```
driusan, line 11:
```suggestion
	composer install
```

Makefile
driusan, line 4:
```suggestion
	composer install --no-dev
```

driusan: dist is already the default for stable versions of packages and there isn't sufficient justification to override the composer defaults in this PR.

PR #5642, opened 2019-11-13
[Password] Fix account lockout when special character present in password

### LINKED ISSUE #5630: [user_accounts] Cannot include "&" in passwords
New CCNA developer (Pierre) discovered a bug in user accounts where if an ampersand is included in a user's password, they will not be able to log in. Reproduced on test-loris-dev.loris.ca. Can edit a user's password to include an ampersand and save without issue, but user cannot log in thereafter and it says that their password is incorrect.

johnsaigle:
## Brief summary of changes

Special characters can now be used in passwords. I forgot that they get automatically encoded upon submission to the codebase so passwords were never decoded before being hashed.

I added some unit tests to make sure everything is working fine.

1. Checkout the branch, change your password to something containing a special character such as `&`, make sure you can login with that password.

* Resolves #5630 
* Extensive background reading about the history of why encoding happens https://github.com/aces/Loris/pull/4491#issuecomment-490567790

### DIFF
--- a/php/libraries/Password.class.inc
+++ b/php/libraries/Password.class.inc
@@ -200,6 +200,11 @@ class Password
      */
     public final function __construct(string $value)
     {
+        // If a password contains a special character such as '&' or '<' then
+        // its possible that it has been encoded during transmission to the
+        // class.
+        $value = htmlspecialchars_decode($value);
+
         // Ensure proposed value is well-formed.
         $this->_validate($value);
         // Don't store the value in the object; instead use the hashed version
--- a/test/unittests/PasswordTest.php
+++ b/test/unittests/PasswordTest.php
@@ -120,4 +120,42 @@ class PasswordTest extends TestCase
             password_verify(self::VALID_PASSWORD, (string) $password)
         );
     }
+
+    /**
+     * dataProvider for testAcceptsSpecialChars
+     *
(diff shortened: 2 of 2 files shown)


test/unittests/PasswordTest.php
```diff
@@ -120,4 +120,42 @@ public function testToString(): void
...
+     *
+     * Provides strings containing special characters.
+     *
+     * @see https://www.php.net/manual/en/function.htmlspecialchars.php
+     */
+    public function specialValues(): array
+    {
+        return array(
+            // test single quote
+            ["password' AND 1=1; --"],
```
johnsaigle: this should be OR but it doesn't actually affect anything

php/libraries/Password.class.inc
```diff
@@ -200,6 +200,11 @@ class Password
      */
     public final function __construct(string $value)
     {
+        // If a password contains a special character such as '&' or '<' then
+        // its possible that it has been encoded during transmission to the
+        // class.
+        $value = htmlspecialchars_decode($value);
```
driusan: This is mixing concerns.. the Password class shouldn't know anything about HTML. The right place to do this is the caller before the constructor is called.

johnsaigle:
Closing this. The bug will be addressed by updating the calling code

PR #5644, opened 2019-11-13
[Install] Fix crash due to unbound variable access

### LINKED ISSUE #5640: [install] after pushing 21 into 22. "unbound variable" in install.sh
[install] after merging 21 into 22. "unbound variable" in install.sh

johnsaigle:
## Brief summary of changes

Fixes a bug where the script would crash due to `$projectname` being accessed before it is initialized.

1. Run the install script. 

* Fixes #5640

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -176,6 +176,7 @@ echo "Ubuntu distribution detected."
         echo $yn | tee -a $LOGFILE > /dev/null
         case $yn in
             [Yy]* )
+                export projectname=""
                 while [ "$projectname" == "" ]; do
                         read -p "Please enter your Project name (if unsure, use LORIS) : " projectname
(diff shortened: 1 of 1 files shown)


tools/install.sh
```diff
@@ -176,6 +176,7 @@ echo "Ubuntu distribution detected."
         echo $yn | tee -a $LOGFILE > /dev/null
         case $yn in
             [Yy]* )
+                export projectname=""
```
driusan: Why is this exported?

johnsaigle: I was getting an error without putting that keyword in front. I'm not that familiar with bash in this regard.

Do you know whether `declare` would work better?

driusan: if you were getting an error without it it's fine, I'll merge it. It just looks out of place.

PR #5647, opened 2019-11-14
[install.sh] fix unbound variable

kongtiaowang:
Fix unbound variable in the install.sh file.
https://github.com/aces/Loris/issues/5640


kongtiaowang:
duplicate

PR #5648, opened 2019-11-14
Fix account lockout when special character present in password

### LINKED ISSUE #5630: [user_accounts] Cannot include "&" in passwords

johnsaigle:
_Replaces #5642_

## Brief summary of changes

Special characters can now be used in passwords. I forgot that they get automatically encoded upon submission to the codebase so passwords were never decoded before being hashed.

1. Checkout the branch, change your password to something containing a special character such as `&`, make sure you can login with that password.

* Resolves #5630 
* Extensive background reading about the history of why encoding happens https://github.com/aces/Loris/pull/4491#issuecomment-490567790

### DIFF
--- a/modules/login/php/passwordexpiry.class.inc
+++ b/modules/login/php/passwordexpiry.class.inc
@@ -88,7 +88,7 @@ class PasswordExpiry extends \NDB_Form
     {
         $this->_user = \User::factory($this->_username);
         $data        = $this->_user->getData();
-        $plaintext = htmlspecialchars_decode($values['password']);
+        $plaintext   = htmlspecialchars_decode($values['password']);
 
         try {
             $password = new \Password($plaintext);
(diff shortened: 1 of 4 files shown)


ridz1208:
Code looks good. I guess the only thing confusing me is, why didn't this get done when #4491 got tested ?

is the only difference that #4491 targeted update password while this is targeting new passwords ?
`

johnsaigle:
Well since that PR all the Password stuff got rewritten. I guess I removed the decoding when I was doing that.

The difference in this approach is that now the modules are responsible for decoding the input instead of the libraries, and that probably makes more sense anyway.

maltheism:
@johnsaigle my only thought from reviewing the code is a question of coding pattern. Would it make more sense to have the htmlspecialchars_decode inside the functions dealing with the password; and Instead of having to remember to use htmlspecialchars_decode whenever the password functions are being used. I know the password functions will likely never be used anywhere else but I thought it might look better. Thoughts?

Issue #5576, opened 2019-11-07
[docs][GoogleSoD]  Update Help text : Phase 2


christinerogers:
For Google Season of Docs // Alex Knoll 
> Followup to #5241 : Update Help text (phase 1) 

- [x] Help text : updates to existing markdown file found in LORIS codebase under `modules/_module_name_/help/`

Companion docs:
- [x] _in progress_ Help text Style guide: to be committed as a Markdown - for devs 
- [x] _in progress_ Workflow User guide: in a googledoc (link shared with CR)

* Your draft / screenshots should be done looking at the 22 release -> https://test-loris-dev.loris.ca/
(credential to login is in the 22 release Project : [See this card](https://github.com/aces/Loris/projects/19#card-21299625) )
* To make your changes: create a new branch based on the `major` branch (re-synced on your fork to get recent updates) 
* it's ok to pile a few modules into a single PR, but try to limit commits to one module at a time (good practice for easy reverts)
* Post screenshots here of the Help text when done, for easy review. 

**Updates, Nov12-19**
* Use the `22.0-release` branch as your base (not `major`) for this week
* The following modules may have old behaviour (e.g. `Show Data`) because [not reactified yet.](https://github.com/aces/Loris/projects/8)  They are: `genomic browser`, `mri_violations`, `data_team_helper` aka (newly renamed) `behavioural_QC`, and `data_release`
* Subpages may have been neglected due to Bug #5709 (Nov.19)

Inputs: 
* old Loris video
* current help text 
* 2014 edition Basic User Guide (abbreviated walkthrough) -- see my email Nov.7
* Module spec (Readme markdown)  e.g. in the codebase: modules/new_profile/README.md

christinerogers:
Modules under the [modules/ directory](https://github.com/aces/Loris/tree/22.0-release/modules) not listed in the front-end Menus: 
* candidate_parameters
* Timepoint_list
* create_timepoint
* BrainBrowser - excepted : doesn't really count & no need for review/updating

Subpages / other Help content : 
- dicom_archive/help/viewDetails.md
- User Preferences 
- Genomic Browser subtabs? 
exceptions: 
- MRI Protocol violations - never had any, made a separate issue: #5714 
- Imaging Browser: View Session page covered by #5690

christinerogers:
👍  Closing this ticket : All PRs sent and merged/approved. 💯 
* Only 1 left to merge #5746 
What is not done will be spun off into a separate Issue. 

[...]

PR #5653, opened 2019-11-14
[Publications] updated module help text

knoll-alex:
New copy for Publications module help text

#### Testing instructions

1. Review copy
2. See screenshot for rendering

#### Links to related tickets

* https://github.com/aces/Loris/issues/5576

Screenshot 2019-11-14 14 14 28


knoll-alex:
Note - in the menu at the top of LORIS, this module is labelled Publications, but labelled as Publication (no s) everywhere else

christinerogers:
@Jkat  -- could you take 30 secs to review this update to the Publication module Help text (as the module expert) ? 
Please [review this markdown](https://github.com/aces/Loris/blob/001c58a8994aed212895014f43f75870eef63314/modules/publication/help/publication.md)  (not the out of date screenshot above)

modules/publication/help/publication.md
```diff
@@ -1,11 +1,12 @@
...
+In the **Browse** tab, use the selection filters to search for a project of interest. 
 
+You can submit a new proposal in the **Propose a Project** tab, and upload specific files to associate with it. 
 
-Navigate to the "__Propose a Project__" tab for proposing research projects based on the data required. You may upload publication media which may then be reviewed by study PIs.
+Pay special attention to user permissions when submitting proposals. 
+Note that PIs can designate other users to edit and later update the project or publication information. Only users affiliated with the proposal can delete publications. 
 
-You should specify which specific variables of interest are relevant to your research so that the reviewers can determine if they should grant access to the data.
+You should also make sure to specify which variables of interest are relevant to your research, so reviewers can properly determine if they should grant access to the data.
```
PapillonMcGill:
```suggestion
You should also ensure to specify which variables of interest are relevant to your research, so reviewers can properly determine if they should grant access to the data.
```

christinerogers: Thanks @PapillonMcGill - we're shifting the style away from `ensure` to make it more friendly / easily readable.

PapillonMcGill: @christinerogers In my opinion, it look more professional and less sloppy
I don't think going back to make sure in a good move

knoll-alex: What about removing that altogether? Instead: "It's important to specify which variables of interest are relevant to your research..."

christinerogers: I think "Important to specify" is better than "make sure" -- though I know "make sure" has already made it into a lot of the Help texts already merged.

christinerogers:
> The writing looks good to me. However, I think @LeighMac would be in a better position to validate this text since this module is being actively used by IBIS (and not by CCNA unfortunately even though it was originally requested by CCNA)

Thanks @mohadesz  - please leave an Approval review if you're comfortable with that.  thanks for also tagging Leigh.

christinerogers:
2 approvals - can be merged @driusan

PR #5674, opened 2019-11-18
[script] fix bugs indentified by shellcheck

PapillonMcGill:
## Brief summary of changes
rebased #5204 on master
fix 3 bugs indentified by running shellcheck (thanks @gdevenyi)

for tools/install.sh, run as if doing a new install
for run-php-linter.sh, run npm run lint:php
for integration.sh, run as if doing integration test (test/integration.sh)

required by #5160


PapillonMcGill:
@johnsaigle @ridz1208 ping!

tools/install.sh
```diff
@@ -128,13 +128,13 @@ fi
 debian=("Debian" "Ubuntu")
 redhat=("Red" "CentOS" "Fedora" "Oracle")
 
-if [[ " ${debian[*]} " =~ " $os_distro " ]]; then
+if [[ " ${debian[*]} " == *" $os_distro "* ]]; then
```
driusan: I still find this much less clear and don't understand the reason for it.

PapillonMcGill: it solve an error reported by change proposed in #5160.
This change using regex to match exact string inside an array.

johnsaigle: @driusan I commented in the original PR that this seems to be a pretty common way of doing things in bash.

driusan: What is the "error" that it's trying to solve?

PapillonMcGill: from https://travis-ci.org/aces/Loris/jobs/599793149?utm_medium=notification&utm_source=github_status
```
In tools/install.sh line 163:

if [[ " ${debian[*]} " =~ " $os_distro " ]]; then

                          ^------------^ SC2076: Don't quote right-hand side of =~, it'll match literally rather than as a regex.

```
without quote, it was not working. Don't remember the problem.

johnsaigle: From memory:

Basically `=~` is a regex operator and using quotation marks causes it to match literally. So if we wanted "\? $os_distro " the escape character wouldn't work.

Since this isn't a regex anyway, we use `==` instead. For this new operator to match anywhere in the string, the `*`s are needed.

driusan: I still don't like this and think it's confusing and should have a better solution, but since I don't know bash well enough to suggest one, I'll begrudingly merge this..

gdevenyi: From the bash manual in "[[ expression ]]"
> When the == and != operators are used, the string to the right of the operator is considered a pattern and matched according to the rules described below under Pattern Matching. 

From pattern matching:

> Any character that appears in a pattern, other than the special pattern characters described below, matches itself. The NUL character may not occur in a pattern. A backslash escapes the following character; the escaping backslash is discarded when matching. The special pattern characters must be quoted if they are to be matched literally.
> 
> The special pattern characters have the following meanings:
> 
> *
> Matches any string, including the null string.

So that test checks for the contents of ``$os_distro``, surrounded by spaced on either side (explicitly) and anything else on either side.

My guess is perhaps the original test was meant to be a regex test of "contains this string". Personally I think that this one is more explicit and clear.

PR #5677, opened 2019-11-18
[brainbrowser] fix multiple volumes display bug

cmadjar:
## Brief summary of changes

When trying to load more than one volume in BB via the imaging browser, only the first volume got displayed. When we refreshed the page, then the second volume would be displayed and so on.

This fixes that bug and now all volumes are being displayed the first time the page is loaded.

1. From the imaging browser page, select two volumes and click on either the "3D Only" or "3D + Overlay" button and verify that both volumes are being loaded. Note: the volumes might be slow to load given the size of the data being loaded in the browser...

* #5474 fixes the lack of clarity part of the issue (the slowness cannot really be fixed due to the size of the data being sent I guess...)

### DIFF
--- a/modules/brainbrowser/js/brainbrowser.loris.js
+++ b/modules/brainbrowser/js/brainbrowser.loris.js
@@ -749,10 +749,10 @@ $(function() {
         minc_ids_arr = [minc_ids];
     }
 
-    let request;
+    let requests = [];
     for (i = 0; i < minc_ids_arr.length; i += 1) {
 
-      request = $.ajax({
+      let single_request = $.ajax({
         url: loris.BaseURL + "/brainbrowser/ajax/image.php",
         data: 'file_id=' + minc_ids_arr[i],
         method: 'GET',
@@ -796,6 +796,7 @@ $(function() {
           }
         }
       });
+      requests.push(single_request);
     }
 
     if (getQueryVariable("overlay") === "true") {
@@ -816,7 +817,7 @@ $(function() {
       $("#panel-size").change();
     }
 
-    //////////////////////////////
+    /////////////////////////////(
     // Load the default color map.
     //////////////////////////////
     viewer.loadDefaultColorMapFromURL(color_map_config.url, color_map_config.cursor_color);
@@ -836,7 +837,7 @@ $(function() {
     /////////////////////
     // Load the volumes.
     /////////////////////
-    request.then(function(filename) {
+    $.when.apply($, requests).then(function() {
       bboptions.volumes = minc_volumes;
       viewer.render();                // start the rendering
       viewer.loadVolumes(bboptions);  // load the volumes
--- a/modules/brainbrowser/js/brainbrowser.loris.js
+++ b/modules/brainbrowser/js/brainbrowser.loris.js
@@ -2,14 +2,14 @@
 function getQueryVariable(variable) {
     "use strict";
     var query = window.location.search.substring(1),
-      vars = query.split("&"),
-      i,
-      pair;
+        vars = query.split("&"),
+        i,
+        pair;
     for (i = 0; i < vars.length; i += 1) {
-      pair = vars[i].split("=");
-      if (pair[0] === variable) {
-        return unescape(pair[1]);
-      }
+        pair = vars[i].split("=");
+        if (pair[0] === variable) {
+            return unescape(pair[1]);
+        }
     }
 }
 
@@ -553,9 +553,9 @@ $(function() {
                $('#filename-additional-info-'+vol_id).slideToggle("fast");
                var arrow = $(this).siblings('.arrow');
                if (arrow.hasClass('glyphicon-chevron-down')) {
-                     arrow.removeClass('glyphicon-chevron-down').addClass('glyphicon-chevron-up');
+                   arrow.removeClass('glyphicon-chevron-down').addClass('glyphicon-chevron-up');
                } else {
-                     arrow.removeClass('glyphicon-chevron-up').addClass('glyphicon-chevron-down');
+                   arrow.removeClass('glyphicon-chevron-up').addClass('glyphicon-chevron-down');
                }
 
        });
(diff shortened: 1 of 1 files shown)


maltheism:
Hi @cmadjar, I tried to test and I'm not 100% sure on how to do it.
I'm getting an error in my browser console "Error: Sorry, this does not look like a NetCDF file." with the brainbrowser page stucking on "Loading" but I think that's non-related to the PR. I selected "demo_676061_V5_t2_001_t2-defaced_001.mnc" & "demo_676061_V5_fMRI_001.mnc" from http://localhost/imaging_browser/viewSession/?sessionID=2150&backURL=/imaging_browser/ and I'm just using rasinbread data. I don't believe I have those mnc files. Anyway I can mimic the files or maybe I'm doing something wrong.

cmadjar:
@maltheism to have brain browser working, you need to have the MINC tools installed and you need to set the path to the MINC toolkit in the `config.xml` file.

The latest MINC toolkit  versions can be downloaded here: https://bic-mni.github.io/#notes-about-versions

maltheism:
thanks @cmadjar for the setup help of the brainbrowser. Everything is working! @driusan

modules/brainbrowser/js/brainbrowser.loris.js
```diff
@@ -836,7 +837,7 @@ $(function() {
     /////////////////////
     // Load the volumes.
     /////////////////////
-    request.then(function(filename) {
+    $.when.apply($, requests).then(function() {
```
HenriRabalais: What exactly does `$.when.apply` do? Is it waiting for all the `requests` to resolve and then triggering the function? 

Eventually I think it would be preferable to switch to using JavaScript promises to handle this sort of thing, but I think that would involve major refactoring of this file.

maltheism: @HenriRabalais [$.when.apply](https://stackoverflow.com/questions/14777031/what-does-when-apply-somearray-do) you're correct. Maybe update to forEach with promises can be done after release?

driusan: Why would it require major refactoring? It seems to me that this is just 1 (new) request chain which should be stand alone relative to the rest of the JS in the file.

HenriRabalais: @driusan Sorry, I meant the switch from jquery to javascript would probably take a lot of refactoring. This change could be done (relatively?) easily I think. Just wondering if it's better to keep it with jQuery if that's what the rest of the file is already using.

driusan: My personal thoughts are that it's worth using a real promise chain here since it's part of the javascript standard and doesn't interfere with anything else, rather than getting more-and-more dependent on jQuery.

modules/brainbrowser/js/brainbrowser.loris.js
```diff
@@ -740,71 +740,79 @@ $(function() {
...
+            resolve();
+          },
+          error: function (jqXHR, status, errorThrown) {
+            console.log('error');
+            if (errorThrown === "Not Found") {
+              let msg = "ERROR: file not found.";
+              $("#loading").html(msg);
+              console.error(msg);
+            }
+            resolve();
```
driusan: shouldn't this reject the promise if there's an error?

modules/brainbrowser/js/brainbrowser.loris.js
```diff
@@ -836,7 +844,7 @@ $(function() {
     /////////////////////
     // Load the volumes.
     /////////////////////
-    request.then(function(filename) {
+    Promise.all([...promises]).then(() => {
```
driusan: Is `[...promises]` a js idiom I'm not familiar with? Why not just `promises`?

maltheism: @driusan the `...` is the spread operator. See [here about it](https://dev.to/sagar/three-dots---in-javascript-26ci).
Insead of doing something like this:
```
Promise.all([Promise1, Promise2, Promise3])
 .then(result) => {
   console.log(result)
 })
 .catch(error => console.log(`Error in promises ${error}`))
```
I thought to use the spread operator. I'm not sure if it will work with just promises.

driusan: I know what the spread operator does, but you're spreading out the results of an array and then putting the result back into a new array.

maltheism: @driusan lol you're right.

modules/brainbrowser/js/brainbrowser.loris.js
maltheism: @driusan I asked @HenriRabalais and we thought it would be best to not do that because it would result in the other brains not loading. We tjpigjt it would be best to work on the error handling in another PR when the jquery ajax is removed to be fetch.

modules/brainbrowser/js/brainbrowser.loris.js
cmadjar: @driusan @maltheism fixed now

modules/brainbrowser/js/brainbrowser.loris.js
driusan: @cmadjar thoughts on this from a user perspective? If there's an error with one should the others still show?

cmadjar: I don't really have a strong opinion on this. I see pros and cons on both sides...

driusan: Okay, then once the whitespace is fixed and it's passing Travis I'll merge it as-is and it can be revisited later.

cmadjar:
@driusan My OCD is struggling very hard on your space request cause the spaces changes made by @maltheism actually fixes the alignments and makes the code clearer to read... Is that absolutely necessary to remove the space changes?

driusan:
PRs should do one, atomic thing. Formatting changes should be in a different PR.

PR #5679, opened 2019-11-18
[integration test] prepare to rewrite the integration test for react element

kongtiaowang:
Make it easy to write a react test.
It is difficult to finish all the tasks in one PR. 
This PR only contains all the react test functions.


htdocs/router.php
```diff
@@ -16,8 +16,7 @@
 $request = $_SERVER['REQUEST_URI'];
 
 if ($request != '/'
-    && (    file_exists(__DIR__ . $request)
-    || file_exists(__DIR__ . "/" . $urlpath))
+    && file_exists(__DIR__ . $request)
```
driusan: Why are you changing the routing here?

php/libraries/NDB_Config.class.inc
```diff
@@ -92,6 +92,12 @@ class NDB_Config
      */
     function load(string $configFile = "../project/config.xml"): void
     {
+        // in test mode,load configFile from Docker DB
+        $env = getenv('LORIS_DB_CONFIG');
+        if (!empty($env)) {
+            $configFile = $env;
+        }
+
```
driusan: This is going to override the parameter passed as an argument. The determination of where the config file is done in singleton() (and this also misses a bunch of cases handled there), which should already be taking care of this. Why is this change being added here?

kongtiaowang: This load function will import the Database from /project/config.xml. In the Docker test, this file should be changed read from LORIS_DB_CONFIG. ( docker-compose.yml environment: - LORIS_DB_CONFIG=/app/test/config.xml)

driusan: The Docker test shouldn't be calling load directly, it should be going through the factory, and it's the caller's responsibility to have called singleton with the correct location the first time.

test/integrationtests/LorisIntegrationTest.class.inc
```diff
@@ -556,5 +559,184 @@ abstract class LorisIntegrationTest extends TestCase
...
+            } catch (Exception $e) {
+         //       echo 'Caught exception: ',  $e->getMessage(), "\n";
+                        $err = $e;
+                        $attempts++;
+                        sleep(1);
+                        continue;
+            }
+        } while($attempts < self::NUM_OF_ATTEMPTS);
+        if ($err && $attempts > 4) {
+           echo 'Caught exception: ',  $err->getMessage(), "\n";
```
driusan: Shouldn't this fail the test with the error instead of just printing the message to stdout?

test/integrationtests/LorisIntegrationTest.class.inc
```diff
@@ -556,5 +559,184 @@ abstract class LorisIntegrationTest extends TestCase
...
+         //       echo 'Caught exception: ',  $e->getMessage(), "\n";
+                        $err = $e;
+                        $attempts++;
+                        sleep(1);
+                        continue;
+            }
+            break;
+        } while($attempts < self::NUM_OF_ATTEMPTS);
+        if ($err && $attempts > 4) {
+           echo 'Caught exception: ',  $err->getMessage(), "\n";
```
driusan: same

test/integrationtests/LorisIntegrationTest.class.inc
```diff
@@ -556,5 +559,184 @@ abstract class LorisIntegrationTest extends TestCase
...
+           //     echo 'Caught exception: ',  $e->getMessage(), "\n";
+                        $err = $e;
+                        $attempts++;
+                        sleep(1);
+                        continue;
+            }
+              break;
+        } while($attempts < self::NUM_OF_ATTEMPTS);
+        if ($err && $attempts > 4) {
+           echo 'Caught exception: ',  $err->getMessage(), "\n";
```
driusan: same

test/integrationtests/LorisIntegrationTest.class.inc
```diff
@@ -556,5 +559,184 @@ abstract class LorisIntegrationTest extends TestCase
...
+         //       echo 'Caught exception: ',  $e->getMessage(), "\n";
+                        $err = $e;
+                        $attempts++;
+                        sleep(1);
+                        continue;
+            }
+            break;
+        } while($attempts < self::NUM_OF_ATTEMPTS);
+        if ($err && $attempts > 4) {
+           echo 'Caught exception: ',  $err->getMessage(), "\n";
```
driusan: same

test/integrationtests/LorisIntegrationTest.class.inc
```diff
@@ -556,5 +559,184 @@ abstract class LorisIntegrationTest extends TestCase
...
+             //   echo 'Caught exception: ',  $e->getMessage(), "\n";
+                        $err = $e;
+                        $attempts++;
+                        sleep(1);
+                        continue;
+            }
+            break;
+        } while($attempts < self::NUM_OF_ATTEMPTS);
+        if ($err && $attempts > 4) {
+           echo 'Caught exception: ',  $err->getMessage(), "\n";
```
driusan: same

test/integrationtests/LorisIntegrationTest.class.inc
```diff
@@ -556,5 +559,184 @@ abstract class LorisIntegrationTest extends TestCase
...
+        do {
+            try {
+                return $this->webDriver->executescript(
+                    "return document.querySelector('$ui').textContent"
+                );
+            } catch (Exception $e) {
+         //       echo 'Caught exception: ',  $e->getMessage(), "\n";
+                        $err = $e;
+                        $attempts++;
+                        sleep(1);
```
driusan: doesn't selenium have a "wait for element" that would be more effective than trying and sleeping repeatly?

kongtiaowang:
replaced by https://github.com/aces/Loris/pull/5964.

PR #5680, opened 2019-11-18
[Login] Fix issue where password could not be reset

johnsaigle:
## Brief summary of changes

This file was not updated when the function `passwordChanged` was renamed to `isPasswordDifferent` which caused a PHP Fatal Error when trying to reset an expired password.

1. On `22.0-release`, Manually expire a user's password.
2. Try logging in to the front-end.
3. Enter a new password
4. You should get a 403 error.
5. Do the same on my branch and everything should work normally.


johnsaigle:
@ridz1208 @driusan This is likely critical but I'll leave it to you to add the label if you agree.

PR #5686, opened 2019-11-19
Script to create VERSION file without git

### LINKED ISSUE #5682: [Install] Is Git a dependency for production installs?
This error was reported on the dev mailing list.
```
root@ubuntu-s-1vcpu-2gb-tor1-01:/var/www/loris# make
git describe --tags --always > VERSION
fatal: not a git repository (or any of the parent directories): .git
Makefile:8: recipe for target 'VERSION' failed
make: *** [VERSION] Error 128
```

Even for non-developer installs, git seems to be required in order to generate the VERSION file or else the make command will fail.

Does this mean git is a dependency for LORIS?

johnsaigle:
## Brief summary of changes

When a user runs `make` without git being installed, the script will crash because git was required to create the VERSION file.

This adds a small script to fix that problem.

1. Checkout the branch. Delete your VERSION file if it exists.
2. Run `make VERSION`.
3. The VERSION file should contain information from git.
4. Run `mv .git/ .gitbkp/` to simulate a system without git.
5. Run `make VERSION`
6. The VERSION file should contain the string 'Unknown'
7. Run `mv .gitbkp/ .git/` to restore git.

* Resolves #5682

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ all: VERSION javascript
 
 # If anything changes, re-generate the VERSION file
 VERSION: .
-	git describe --tags --always > VERSION
+	tools/create-version-file.sh
 
 phpdev:
 	composer install
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ all: VERSION javascript
 
 # If anything changes, re-generate the VERSION file
 VERSION: .
-	tools/create-version-file.sh
+	tools/gen-version.sh
 
 phpdev:
 	composer install
--- a/tools/create-version-file.sh
+++ b/tools/create-version-file.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+# Creates the VERSION file in the LORIS root folder.
+set -euo pipefail
+
+# The name of the version file
+VERSION='VERSION'
+
+if [ -d '.git/' ]; then
+    git describe --tags --always > $VERSION
+else
+    echo 'Unknown' > $VERSION
+fi
--- a/tools/gen-version.sh
+++ b/tools/gen-version.sh
@@ -5,7 +5,9 @@ set -euo pipefail
 # The name of the version file
 VERSION='VERSION'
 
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -5,7 +5,7 @@ all: VERSION javascript
 
 # If anything changes, re-generate the VERSION file
 VERSION: .
-	git describe --tags --always > VERSION
+	tools/create-version-file.sh
```
driusan: this should be called something like `gen-version.sh` since it also gets invoked if the version changes.

tools/create-version-file.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+# Creates the VERSION file in the LORIS root folder.
+set -euo pipefail
+
+# The name of the version file
+VERSION='VERSION'
+
+if [ -d '.git/' ]; then
```
driusan: this should also check if the `git` command is installed

johnsaigle:
@driusan I incorporated your changes. please re-review.

tools/gen-version.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+# Creates the VERSION file in the LORIS root folder.
+set -euo pipefail
+
+# The name of the version file
+VERSION='VERSION'
+
+GIT_INSTALLED=$()
```
driusan: it looks like you started to do things one way, and then changed your mind and did something different.. this doesn't seem to be used

johnsaigle: oops sorry, you're right.
```suggestion
```

tools/gen-version.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+# Creates the VERSION file in the LORIS root folder.
+set -euo pipefail
+
+# The name of the version file
+VERSION='VERSION'
+
```
johnsaigle:
```suggestion
```

PR #5699, opened 2019-11-19
[Core] Fix TypeError/500 error when Certification settings not well-formatted

### LINKED ISSUE #5698: [Core] TypeError causes 500 response when installing an instrument via Instrument Manager
>PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Utility::associativeToNumericArray() must be of the type array, string given, called in /var/www/loris/php/libraries/NDB_BVL_Instrument.class.inc on line 912

Also there are issues in the module with bad filepath to the page decorators:
> PHP Fatal error:  Uncaught Error: Class 'LORIS\\Http\\Response\\Conflict' not found in /var/www/loris/modules/instrument_manager/php/instrument_manager.class.inc:174

This should be JSONResponse, not response

johnsaigle:
# Brief summary of changes

Throws a configuration exception if the data in the config file is not formatted the way it's supposed to be.

Also fixes a namespacing issue in the Instrument Manager that was triggered in the same call stack. This caused a PHP Fatal Error.

1. Enable `EnableCertification` in your config.xml file.
2. Leave the rest of the tags empty.
3. Try to upload the `test_all_fields.linst` instrument to the instrument manager. (`test/instruments/test_all_fields.linst`)
4. It won't work
5. Do the same on this PR.

* Resolves #5698

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -29,7 +29,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
 {
     protected const ERROR_CERTIFICATION_BAD_FORMATTING
         = 'Certification settings are not formatted correctly in the '
-        . 'config file. Cannot continue';
+        . 'configuration file. Cannot continue.';
     /**
      * Determines whether this instrument should load data in
      * JSON format from the "Data" column of the flag table, or
@@ -180,6 +180,8 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      *
      * @return object|null the new object of $instrument type
      * @access public
+     * @throws NotFound
+     * @throws Exception
      */
     static function factory(
         string $instrument,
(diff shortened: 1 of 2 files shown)


driusan:
@johnsaigle this is failing Travis

johnsaigle:
> DicomArchiveTestIntegrationTest::testdicomArchiveFilterClearBtn
UnexpectedJavascriptException: TypeError: input is null

I've seen this a few times on other PRs. I think it's a transitory fialure since this is only failing on 7.3 and not 7.2. We should keep an eye on it. 

I'll restart travis

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -264,8 +269,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         // check that user has access
         if ($access == false) {
             throw new Exception(
-                "You do not have access to this page.",
-                403
+                "You do not have access to this page."
```
driusan: The numeric code should not be removed from the exception or it'll break the HTTP return code.

johnsaigle: This is a generic Exception so I figured this was a mistake because the codes are arbitrary. I don't know of any place in LORIS that translates the int `403` into a proper forbidden response. If these is a place though let me know.

driusan: main.php and survey.php both do according to `git grep 403`

johnsaigle: Thanks, hadn't seen that before. I'll add it back.

johnsaigle:
Failed again due to DICOM Archive unstable test. Restarting Travis.

Issue #5700, opened 2019-11-20
[instrument_list] "Delete instrument data" button leads to 500 error


leighevans:
Clicking the "Delete instrument data" button in the top lefthand corner of the page of an instrument leads to a 500 error.
Also what is this new button supposed to do (with two ambiguous checkboxes next to it)? Is it permission based? It's not in the test plan so not sure how it's supposed to function...

![Screen Shot 2019-11-20 at 11 05 12 AM](https://user-images.githubusercontent.com/32552953/69255506-ee9c5480-0b85-11ea-973c-29449aac9b0d.png)

driusan:
Should be fixed by #5712

PR #5712, opened 2019-11-20
[instrument_list] fix "Delete instrument data" 500 Error

kongtiaowang:
https://github.com/aces/Loris/issues/5700
test: click "Delete instrument data" button and two checkboxes. 
Then check the form of this instrument.

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2203,8 +2203,6 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             array('CID' => $this->commentID),
             array('nofetch' => true)
         );
-        // after update database, it needs to refresh the front-end page
-        header("Refresh:0");
     }
 
     /**
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2707,14 +2707,9 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
-        if ($request->getMethod() === "POST" && !isset($_POST['ClearInstrument'])) {
+        $req = $request->getParsedBody();
+        if ($request->getMethod() === "POST" && !isset($req['ClearInstrument'])) {
             $this->save();
-        }
-        if (isset($_POST['ClearInstrument'])
-            && isset($_POST['deleteconfirm'])
-            && isset($_POST['deleteconfirm2'])
-        ) {
-            $this->ClearInstrument();
         }
         return (new \Zend\Diactoros\Response())
             ->withBody(new \LORIS\Http\StringStream($this->display() ?? ""));
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2203,6 +2203,8 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             array('CID' => $this->commentID),
             array('nofetch' => true)
         );
+        // after update database, it needs to refresh the front-end page
+        header("Refresh:0");
     }
 
     /**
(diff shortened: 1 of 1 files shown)


ridz1208:
@maltheism did you test it too ?

maltheism:
@ridz1208 I just tested the PR. I'm a little confused after looking at what @leighevans commented in the issue. I think maybe one of the checkboxes should be removed or both?  This PR didn't address that but could be removed from inside the smarty/templates/instrumentstatus_controlpanel.tpl on line 7 & 8.

ridz1208:
the main concern right now is the functionality and the 500 error. 

if you tested and they were fixed please add the manual test label

maltheism:
@ridz1208 it does work now without 500 error.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2707,10 +2707,15 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
-        if ($request->getMethod() === "POST") {
+        if ($request->getMethod() === "POST" && !isset($_POST['ClearInstrument'])) {
             $this->save();
         }
-
+        if (isset($_POST['ClearInstrument'])
+            && isset($_POST['deleteconfirm'])
+            && isset($_POST['deleteconfirm2'])
+        ) {
+            $this->ClearInstrument();
```
ridz1208: This function is being called from the NDB_BVL_InstrumentStatus_ControlPanel.class.inc

I dont think this is the correct fix. the problem sounds like an order of operations issue

maltheism: @ridz1208 @kongtiaowang  the `$this->ClearInstrument();` is also called in NDB_BVL_Instrument.class.inc on line 2717 if that's helpful.

kongtiaowang: @ridz1208 @maltheism Why I changed the NDB_BVL_Instrument.class.int not NDB_BVL_InstrumentStatus_ControlPanel.class.inc ? Reason 1: They all use the same ClearInstrument function in NDB_BVL_Instrument.class.inc on line 2717. Reason 2: When the clear button clicks in ControlPanel, we need the instrument page change, not the ControlPanel. If we call this function in ControlPanel, it will not change instrument page's front-end. We have to refresh the whole page again to see the change.

maltheism: @ridz1208 what do you think about @kongtiaowang's comment? Seems to make sense to me. Although it seems weird to have the same function inside of the save method.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2707,10 +2707,15 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
-        if ($request->getMethod() === "POST") {
+        if ($request->getMethod() === "POST" && !isset($_POST['ClearInstrument'])) {
             $this->save();
         }
-
+        if (isset($_POST['ClearInstrument'])
+            && isset($_POST['deleteconfirm'])
```
driusan: I agree with @ridz1208 that this is already getting called elsewhere and the solution shouldn't be to duplicate it enough.. the isset from the first `if` statement should be sufficient to fix the 500 bug.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2707,10 +2707,15 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
-        if ($request->getMethod() === "POST") {
+        if ($request->getMethod() === "POST" && !isset($_POST['ClearInstrument'])) {
```
driusan: values being checked should come from the `$request` ServerRequestInterface argument to the function, not the `$_POST` superglobal.

driusan:
@kongtiaowang any updates on this?

kongtiaowang:
@driusan Could u review again?

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2203,6 +2203,8 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             array('CID' => $this->commentID),
             array('nofetch' => true)
         );
+        // after update database, it needs to refresh the front-end page
+        header("Refresh:0");
```
driusan: this line shouldn't have been added

driusan:
@ridz1208 can you review?

PR #5713, opened 2019-11-20
[Core] Fix help text not loading for subpages

### LINKED ISSUE #5709: Subtab/page Help text is not displaying
Help text markdown files for subpages is not displaying in the front end. These files are separate markdowns in the codebase. 

* Imaging Browser : View Session  
* My preferences show the User accounts Help (not ok!) at http://demo.loris.ca/user_accounts/my_preferences/#

_To confirm_: likely also in (but not sure if reactification is relevant?)
* MRI Violations 
* DICOM archive
* more modules with sub-pages... 

Similar to: https://github.com/aces/Loris/issues/5676

johnsaigle:
## Brief summary of changes

The `subtest` field was not being populated by the Middleware. As a result, the parameter was never passed to the back-end in order to load subpage help text. Instead the code would always fall back to loading the main help text for the module.

Visit a module with a subpage. An example would be going to DICOM Archive and clicking "View Details". There should be specific help text for that module that is different from the help text for the main module.

* Resolves #5709

### DIFF
--- a/modules/imaging_browser/templates/form_viewSession.tpl
+++ b/modules/imaging_browser/templates/form_viewSession.tpl
@@ -1,5 +1,4 @@
 <!-- Main table -->
-{* Olga: show3DViewer not tested *}
 {if $show3DViewer}
 {*<td nowrap="nowrap">the first opening td already opened in main.tpl *}<input type="button" name="button" value="3D Viewer" class="button" id = "dccid" name = "dccid" style = "background-color: #816e91" onclick="window.open('BrainBrowser/display.html?sessionID={$subject.sessionID}')" /></td>
 
--- a/src/Middleware/UserPageDecorationMiddleware.php
+++ b/src/Middleware/UserPageDecorationMiddleware.php
@@ -47,7 +47,7 @@ class UserPageDecorationMiddleware implements MiddlewareInterface
         ob_start();
         // Set the page template variables
         // $user is set by the page base router
-        $user = $request->getAttribute("user");
+        $user     = $request->getAttribute("user");
         $tpl_data = array(
                      'test_name' => $this->PageName,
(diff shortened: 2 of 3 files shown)


modules/imaging_browser/templates/form_viewSession.tpl
```diff
@@ -1,5 +1,4 @@
 <!-- Main table -->
-{* Olga: show3DViewer not tested *}
```
christinerogers: 👍

src/Middleware/UserPageDecorationMiddleware.php
```diff
@@ -47,9 +47,11 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
         ob_start();
         // Set the page template variables
         // $user is set by the page base router
-        $user     = $request->getAttribute("user");
+        $user = $request->getAttribute("user");
+        //print_r($request);
```
christinerogers: delete the `// print_r...` ?

src/Middleware/UserPageDecorationMiddleware.php
```diff
@@ -50,6 +50,7 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
         $user     = $request->getAttribute("user");
         $tpl_data = array(
                      'test_name' => $this->PageName,
+                     'subtest'   => $request->getAttribute('subtest') ?? '',
```
driusan: What is this doing? It looks like it's getting overwritten by the other line below in this class.. is "subtest" or "pageclass" the right way to get it?

johnsaigle: Yeah you're right, this may have been the result of two different approaches. I believe that "pageclass" is correct. I'll remove this line and test.

Issue #5651, opened 2019-11-14
Fix Phan errors introduced in PHP 7.4


johnsaigle:
>htdocs/postdeploy.php:45 PhanTypeMismatchArgumentInternalProbablyReal Argument 2 ($algo) is null but \password_hash() takes int
modules/configuration/php/configuration.class.inc:142 PhanCompatibleDimAlternativeSyntax Array and string offset access syntax with curly braces is deprecated in PHP 7.4. Use square brackets instead. Seen for $setting{'ID'}
php/installer/Installer.class.inc:554 PhanTypeMismatchArgumentInternalProbablyReal Argument 2 ($algo) is null but \password_hash() takes int
php/libraries/Password.class.inc:207 PhanTypeMismatchArgumentInternalProbablyReal Argument 2 ($algo) is null but \password_hash() takes int
php/libraries/SinglePointLogin.class.inc:314 PhanTypeMismatchArgumentInternalProbablyReal Argument 2 ($algo) is null but \password_needs_rehash() takes int
php/libraries/SinglePointLogin.class.inc:328 PhanTypeMismatchArgumentInternalProbablyReal Argument 2 ($algo) is null but \password_hash() takes int

driusan:
phan seems to be wrong about the `password_hash`. In all the cases, the second argument being passed on those lines is the `PASSWORD_DEFAULT` constant.

johnsaigle:
Fair enough. We should re-evaluate these bugs after TravisCI updates to PHP 7.4 instead of the snapshot.

Also phan should be updated in case this is a bug on their side. I have a PR doing this here: https://github.com/aces/Loris/pull/5762

Once these two things are updated we can see whether the error still appears and go from there.

xlecours:
Maybe we should explicitely use `PASSWORD_BCRYPT` instead of `PASSWORD_DEFAULT`

> It is worth noting that over time this constant can (and likely will) change. Therefore you should be aware that the length of the resulting hash can change. Therefore, if you use PASSWORD_DEFAULT you should store the resulting hash in a way that can store more than 60 characters (255 is the recommended width). 

from https://www.php.net/manual/en/password.constants.php

johnsaigle:
https://github.com/aces/Loris/blob/4155ea37030eecc913701b1aebd81e0798916dbb/SQL/0000-00-00-schema.sql#L95

If we're going to require a specific password algorithm i think we should do more research into possible advantages/disadvantages there are to the different algorithms.

[...]

Issue #5356, opened 2019-10-22
Add phpstan to static analysis test suite


johnsaigle:
[PHPStan](https://github.com/phpstan/phpstan) is another cool static analysis tool. It found a bunch of issues in #4928 and it would be good to add it to our suite.

johnsaigle:
Blocked until PHPUnit is upgraded in #5351  or #4433 as version >= 7 is required.

PR #5723, opened 2019-11-21
[Core] Raise minimum PHP requirement to 7.3

### LINKED ISSUE #5063: LORIS should start requiring PHP 7.3 on minor + bugfix (and 7.4 on major when it's out)

johnsaigle:
## Brief summary of changes

PHP 7.2 is no longer receiving active support as of Nov. 30th, 2019. PHP 7.4 is being released at about the same time.

Typically we only support the last two releases of PHP. This PR brings the minimum requirement up to 7.3 from 7.2.

Detailed changes between these versions can be found here: https://www.php.net/manual/en/migration73.php

I'm adding the Security label because this upgrade will enable all LORIS instances to be fully protected from CSRF attacks via the SameSite cookie.

* Resolves #5063

### DIFF
--- a/modules/create_timepoint/php/create_timepoint.class.inc
+++ b/modules/create_timepoint/php/create_timepoint.class.inc
@@ -85,8 +85,7 @@ class Create_Timepoint extends \NDB_Form
         $user_list_of_projects = $user->getProjectIDs();
         if (count($user_list_of_projects) == 1) {
             //if there is only one project, autoselect first project from array of 1
-            //TODO: change this to array_key_first() when support is only PHP 7.3+
-            $project = \Project::getProjectFromID(array_pop($user_list_of_projects));
+            $project = array_key_first($user_list_of_projects);
         } else if (count($user_list_of_projects) > 1) {
             $project_id = intval($values['project']);
             $project    = \Project::getProjectFromID($project_id);
--- a/modules/create_timepoint/php/create_timepoint.class.inc
+++ b/modules/create_timepoint/php/create_timepoint.class.inc
@@ -85,12 +85,11 @@ class Create_Timepoint extends \NDB_Form
         $user_list_of_projects = $user->getProjectIDs();
         if (count($user_list_of_projects) == 1) {
             //if there is only one project, autoselect first project from array of 1
-            $project_id = $user_list_of_projects[0];
+            $project = \Project::getProjectFromID(array_pop($user_list_of_projects));
         } else if (count($user_list_of_projects) > 1) {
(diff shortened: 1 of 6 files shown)


PapillonMcGill:
Will also require to update README.CentOS7.md files (currently in 22.0-release).

johnsaigle:
PHP 7.3 is required to use phpstan, so this is blocking #5356.

johnsaigle:
PHP 7.4 is out so this PR can go ahead.

It's possible that Travis doesn't yet support this in the build environments though.

https://www.php.net/releases/7_4_0.php

johnsaigle:
Adding Blocked because TravisCI doesn't yet support testing against the PHP 7.4 release.

johnsaigle:
Blocked by issue #5651

driusan:
This is no longer blocked, phan was updated.

johnsaigle:
@driusan Please review as this is no longer blocked by phan and is also holding up other PRs.

modules/create_timepoint/php/create_timepoint.class.inc
```diff
@@ -85,13 +85,13 @@ class Create_Timepoint extends \NDB_Form
...
             //if there is only one project, autoselect first project from array of 1
-            //TODO: change this to array_key_first() when support is only PHP 7.3+
-            $project = \Project::getProjectFromID(array_pop($user_list_of_projects));
+            $project_id = $user_list_of_projects[0];
         } else if (count($user_list_of_projects) > 1) {
             $project_id = intval($values['project']);
-            $project    = \Project::getProjectFromID($project_id);
         }
 
+        $project = \Project::getProjectFromID($project_id);
```
driusan: This looks like a change in behaviour? Why is this required for 7.3?

driusan: other than that if it's passing Travis, it looks like changes are minimal so it's good enough for me.

johnsaigle: It's not required. It was an enhancement but I agree it's not very helpful here. I'll revert this file.

PR #5726, opened 2019-11-21
[Core] Escape shell argument in Instrument class

johnsaigle:
## Brief summary of changes

This argument should be escaped as it may be possible for a front-end user to supply malicious data via the test name.


driusan:
can you send this to the release branch?

johnsaigle:
@driusan done

Issue #5736, opened 2019-11-22
[Instruments/instrument_list] Instrument set to in progress as soon as it is accessed


ridz1208:
When you access the instrument from the instrument list, it is automatically set as "In Progress" this is normal when it is the first time the instrument is opened.

Issue: the same behaviour is happening when the instrument is already set to complete. As soon as you access it it goes back to in progress

christinerogers:
@ridz1208 are you sure the first point is really a bug? 
This first part is normal behaviour I think, based on my recollection over too many years of LORIS -- (unless it was changed recently?)
> When you access the instrument from the instrument list, it is automatically set as "In Progress" this is normal when it is the first time the instrument is opened.


However on the second point: IMO definitely a bug.  : 

> Issue: the same behaviour is happening when the instrument is already set to complete. As soon as you access it it goes back to in progress


cc: @maltheism, who already submitted a fix #5737

maltheism:
@christinerogers I think the second part is a bug because a user could click on the instrument to view details and navigate back without realizing the instrument is now set to in progress. @ridz1208 thoughts?

christinerogers:
got it -- sorry i missed the "this is normal" part. 
only the second part is the bug report, cool. 
i think i read your fix PR title too early before reading Rida's details.  thanks!

driusan:
merged

PR #5737, opened 2019-11-22
[Instruments/instrument_list] Fix bug where Instrument set to in progress as soon as it is accessed

maltheism:
## Brief summary of changes

Solves the issue #https://github.com/aces/Loris/issues/5736 

1. Access an instrument and check if status is switched.

* https://github.com/aces/Loris/issues/5736

### DIFF
--- a/php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
+++ b/php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
@@ -105,7 +105,7 @@ class NDB_BVL_InstrumentStatus_ControlPanel extends NDB_BVL_InstrumentStatus
 
         if (empty($this->subtest)) {
             $hasAccess = $this->_hasAccess();
-            if ($hasAccess) {
+            if ($hasAccess && $this->getDataEntryStatus() == null) {
                 $this->setDataEntryStatus(
                     $this->_status['Data_entry']
                     ?? 'In Progress'
--- a/php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
+++ b/php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
@@ -105,10 +105,9 @@ class NDB_BVL_InstrumentStatus_ControlPanel extends NDB_BVL_InstrumentStatus
 
         if (empty($this->subtest)) {
             $hasAccess = $this->_hasAccess();
+
             if ($hasAccess && $this->getDataEntryStatus() == null) {
-                $this->setDataEntryStatus(
-                    'In Progress'
-                );
+                $this->setDataEntryStatus('In Progress');
             }
 
             // generate the administration flag buttons
--- a/php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
+++ b/php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
@@ -105,9 +105,6 @@ class NDB_BVL_InstrumentStatus_ControlPanel extends NDB_BVL_InstrumentStatus
 
         if (empty($this->subtest)) {
             $hasAccess = $this->_hasAccess();
-            if ($this->_status === null && !empty($this->_commentID)) {
-                $this->select($this->_commentID);
-            }
             if ($hasAccess) {
                 $this->setDataEntryStatus(
                     $this->_status['Data_entry']
--- a/php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
+++ b/php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
@@ -104,11 +104,9 @@ class NDB_BVL_InstrumentStatus_ControlPanel extends NDB_BVL_InstrumentStatus
         }
 
         if (empty($this->subtest)) {
-            $hasAccess = $this->_hasAccess();
-
-            if ($this->getDataEntryStatus() == null && $hasAccess) {
-                $this->setDataEntryStatus('In Progress');
-            }
+            $status = new \NDB_BVL_InstrumentStatus;
+            $status->select($this->tpl_data['commentID']);
+            $this->setDataEntryStatus($status->_status['Data_entry'] ?? 'In Progress');
 
             // generate the administration flag buttons
             $this->tpl_data['access']['administration']
--- a/php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
+++ b/php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
@@ -104,12 +104,15 @@ class NDB_BVL_InstrumentStatus_ControlPanel extends NDB_BVL_InstrumentStatus
         }
 
         if (empty($this->subtest)) {
-            $status = new \NDB_BVL_InstrumentStatus;
+            $hasAccess = $this->_hasAccess();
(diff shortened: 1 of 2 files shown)


php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
```diff
@@ -104,11 +104,12 @@ class NDB_BVL_InstrumentStatus_ControlPanel extends NDB_BVL_InstrumentStatus
...
-
-            if ($this->getDataEntryStatus() == null && $hasAccess) {
-                $this->setDataEntryStatus('In Progress');
-            }
+            $status = new \NDB_BVL_InstrumentStatus;
+            $status->select($this->tpl_data['commentID']);
+            $this->setDataEntryStatus(
+                $status->_status['Data_entry']
+                ?? 'In Progress'
+            );
```
ridz1208: why was the access check removed ?

maltheism: @ridz1208 the hasAccess check didn't make sense when looking at the previous code and I thought to remove it because it seems unnecessary. Should I add it back?

ridz1208: I think the reason it's there is if the instrument has specific permissions and someone without these permissions clicks on it, it would not change the instrument to in progress

maltheism: @ridz1208 well now the code shouldn't change the instrument to "in progress" when viewing the instrument. So maybe it's unnecessary now?

ridz1208: which is always the case on first access.

so with your code, if I dont have access to instrument X, but i click on it in the instrument list, it will still set it to in progress if it's the first time the instrument is accessed. so it's still an issue

maltheism: @ridz1208 I added it back.

php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
```diff
@@ -105,9 +105,13 @@ class NDB_BVL_InstrumentStatus_ControlPanel extends NDB_BVL_InstrumentStatus
 
         if (empty($this->subtest)) {
             $hasAccess = $this->_hasAccess();
-
-            if ($this->getDataEntryStatus() == null && $hasAccess) {
-                $this->setDataEntryStatus('In Progress');
+            $status    = new \NDB_BVL_InstrumentStatus;
```
driusan: This looks like it's just peppering over a bigger problem.. why is `getDataEntryStatus() == null` not sufficient? Why do we need a `\NDB_BVL_InstrumentStatus` object that's separate from `$this->_status` to get the Data_entry property on it?

maltheism: @driusan getDataEntryStatus() returns null and even if there is a data entry status. It only returns the status when setDataEntryStatus has been called. I'm confused by it myself.

driusan: please try just adding

```
        if ($this->_status === null && !empty($this->_commentID)) {
            $this->select($this->_commentID);
        }
```

at the top of getDataEntryStatus and see if that fixes it. It looks like that used to be done by NDB_Caller which doesn't exist anymore

maltheism: @driusan that works. See newest commit with it added.

php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
```diff
@@ -105,9 +105,14 @@ class NDB_BVL_InstrumentStatus_ControlPanel extends NDB_BVL_InstrumentStatus
 
         if (empty($this->subtest)) {
             $hasAccess = $this->_hasAccess();
-
-            if ($this->getDataEntryStatus() == null && $hasAccess) {
-                $this->setDataEntryStatus('In Progress');
+            if ($this->_status === null && !empty($this->_commentID)) {
+                $this->select($this->_commentID);
+            }
```
driusan: 1. This should be in getDataEntryStatus so that anything that calls the function gets a populated _status property.
2. What happened to the `getDataEntryStatus() === null` part of the if statement?

maltheism: @driusan I moved the code with latest commit. My assumption is the `getDataEntryStatus() === null` isn't necessary since I'm doing:
```
$this->setDataEntryStatus(
                    $this->_status['Data_entry']
                    ?? 'In Progress'
                );
```
and that sets the status to "In Progress" if this->_status['Data_entry'] is null (for keeping the original functionality if first time accessing instrument).

php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
```diff
@@ -105,9 +105,11 @@ class NDB_BVL_InstrumentStatus_ControlPanel extends NDB_BVL_InstrumentStatus
 
         if (empty($this->subtest)) {
             $hasAccess = $this->_hasAccess();
-
-            if ($this->getDataEntryStatus() == null && $hasAccess) {
-                $this->setDataEntryStatus('In Progress');
+            if ($hasAccess) {
```
driusan: this still needs the `$this->getDataEntryStatus() == null`, because otherwise 1. the select isn't getting called and 2. it's updating the database unnecessarily.

maltheism: @driusan that makes sense. I added it again.

driusan:
@ridz1208 since you filed the original ticket can you manually test the latest iteration of this?

php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
```diff
@@ -105,9 +105,11 @@ class NDB_BVL_InstrumentStatus_ControlPanel extends NDB_BVL_InstrumentStatus
 
         if (empty($this->subtest)) {
             $hasAccess = $this->_hasAccess();
-
-            if ($this->getDataEntryStatus() == null && $hasAccess) {
-                $this->setDataEntryStatus('In Progress');
+            if ($hasAccess && $this->getDataEntryStatus() == null) {
+                $this->setDataEntryStatus(
+                    $this->_status['Data_entry']
+                    ?? 'In Progress'
```
ridz1208: you probably dont need the null coalesce anymore

php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
```diff
@@ -105,9 +105,10 @@ class NDB_BVL_InstrumentStatus_ControlPanel extends NDB_BVL_InstrumentStatus
 
         if (empty($this->subtest)) {
             $hasAccess = $this->_hasAccess();
-
-            if ($this->getDataEntryStatus() == null && $hasAccess) {
-                $this->setDataEntryStatus('In Progress');
+            if ($hasAccess && $this->getDataEntryStatus() == null) {
```
driusan: other than stylistic differences this is identical to the previous code

Issue #5708, opened 2019-11-20
[Data Release] Can't remove permissions under Manage Permissions


Jkat:
Unchecking permissions and saving doesn't do anything

ridz1208:
@Jkat could you give steps to reproduce ?
i just tried it out and it seems to be working

Jkat:
I uncheck one of the checked boxes under Manage Permissions for a user and it doesn't seem to save this change. Looks like unchecking both(/all) boxes works...

ridz1208:
@Jkat nice catch !! 

it was a edge case when the version file is 0

driusan:
PR merged

PR #5752, opened 2019-11-25
[Data release] Edge case where version 0 always shows up checked

ridz1208:
## Brief summary of changes

Fixes https://github.com/aces/Loris/issues/5708

Issue: when the version of the file is set to `0` the `in_array()` call will always return true because of cross-type checking. this eliminates the cross-type checking by adding the strict argument


maltheism:
Travis failing for PHP 7.2 but passing for 7.3. Restarting travis for PHP 7.2

Issue #5733, opened 2019-11-22
[New Profile] Limit date of birth to present date.


SantiagoTG:
When creating a new profile, the date of birth should not exceed today's date. However, I had no issue creating a profile whose date of birth is far ahead in the future. On the lower range, the current threshold is set to restrict if the date is before to year 1917.

zaliqarosli:
this is not necessarily a bug, but a configuration that needs to be set

driusan:
Doesn't seem to be an open issue based on @zaliqarosli's comment?

PR #5758, opened 2019-11-25
[new_profile] Limit date of birth to present date and made DOB not required with EDC

maltheism:
## Brief summary of changes

* I made a change inside the DateElement and where the **current** "month & day" will be the limit for max date (month & day) . 
* I changed the NewProfile DateElement to give the current year for the maxYear props.
* Fixed a bug as well, when an empty string was used for the DoB it could result in an infinite loop in php/libraries/Candidate.class.inc. 

1. edc can be in the future and so leave it as the config.
2. birthday should just be the current date.
3. birthday should not be required if using edc.

1. Checkout PR.
2. Go to New Profile module.
3. Attempt to set a date into the future (you can't).

* https://github.com/aces/Loris/issues/5733


jsx/Form.js
```diff
@@ -1056,11 +1056,17 @@ class DateElement extends Component {
...
+    const currentMonth = ('0' + (currentDate.getMonth() + 1)).slice(-2);
+
+    console.log(maxYear);
+
     // Handle date format
     let format = this.props.dateFormat;
     let inputType = 'date';
     let minFullDate = minYear + '-01-01';
-    let maxFullDate = maxYear + '-12-31';
+    let maxFullDate = maxYear + '-' + currentMonth + '-' + currentDay;
```
zaliqarosli: this line makes sense to me. the same change could be done for line 1073 as well

modules/new_profile/jsx/NewProfileIndex.js
```diff
@@ -199,7 +199,7 @@ class NewProfileIndex extends React.Component {
             name = "dobDate"
             label = "Date of Birth"
             minYear = {minYear}
-            maxYear = {maxYear}
+            maxYear = {(new Date()).getFullYear()}
```
zaliqarosli: what if max year is before today though? this maxYear is a meaningful number calculated on the php side using config values (https://github.com/aces/Loris/blob/0910c7bd784e63ab9e0643eb9c0ab52b65627753/modules/new_profile/php/new_profile.class.inc#L169). it doesn't make sense to me project-wise to replace it with today's date.

what's your reasoning for this? i notice that it's going to `22.0-release` branch, is this a bugfix for an issue?

maltheism: @zaliqarosli see the linked related ticket? The issue is to resolve the issue of being able to set a date into the future and doesn't make sense for date of birth.

zaliqarosli: i guess that the restriction on maxYear being today would be if the calculated maxYear from config values is in the future

jsx/Form.js
maltheism: @zaliqarosli nice catch, I'll update it.

modules/new_profile/jsx/NewProfileIndex.js
maltheism: @zaliqarosli yah the maxYear isn't being the current year from the backend. So I thought to add it here on the frontend. thoughts?

zaliqarosli: what we can do is redefine maxYear on line 143 as

```
let today = (new Date()).getFullYear();
let maxYear = (this.state.configData.maxYear > today) ? today : this.state.configData.maxYear
```

maltheism: @zaliqarosli adding it and the current month and current day should still be used. correct? since it solves not having beyond the current date as an option when selecting the date for birthday.

jsx/Form.js
zaliqarosli: actually, thinking about it again, i don't think this makes sense either. let's say the maxYear is 2000 as provided by the prop. then today's month being November shouldn't stop us from being able to enter a date that has December as its month.

but then let's say the maxYear is 2019, then I understand the need to limit the month to November. I think because of these two cases, you would need to add a check for whether maxYear is before today's year or equal to or greater than today's year

modules/new_profile/jsx/NewProfileIndex.js
zaliqarosli: ^ see the comment above that i just posted haha we're thinking the same thing

maltheism: @zaliqarosli quick question, do you know what the edcDate and edcDateConform is for? Since I'm wondering if that should be able to be set to a greater year than the current date or if it should be similar to the birthday.

zaliqarosli: ^ discussed in person

johnsaigle:
Is discussion still required here? I think the discussion was resolved in last week's LORIS meeting.

johnsaigle:
Still one phan error:

> modules/new_profile/php/new_profile.class.inc:263 PhanTypeMismatchArgumentInternal Argument 1 ($string) is null but \strlen() takes string

php/libraries/Candidate.class.inc
```diff
@@ -595,8 +605,8 @@ class Candidate
         $subprojList = array();
 
         $projID  = $this->getProjectID();
-        $query   = "SELECT SubprojectID 
-                    FROM project_subproject_rel 
+        $query   = "SELECT SubprojectID
```
kongtiaowang: need all spaces back, otherwise, it will be like this: "SELECT SubprojectID\nFROM project_subproject_rel"

johnsaigle:
@maltheism This PR looks close to being merged. I'm trying to test it now but I'm getting errors because of references to `LorisMenu` on this branch. Would you mind rebasing the code with the latest changes on master? It'll make testing much easier. I'll try testing again afterward.

maltheism:
@johnsaigle I just pushed the rebase of this PR onto aces/master.

johnsaigle:
The PR works if you use the date picker pop-up: the user is unable to select a date past the current date.

However, it's still possible to enter a date in the future by typing it in manually. Typing in a value of `2020-04-04` allowed be to create a new candidate without any error message.

Could you add a restriction on the text value of the fields to prevent submission when a user enters a date this way?

maltheism:
@johnsaigle nice find. I'll investigate and see what I can do.

maltheism:
@johnsaigle I'm wanting to ask of your opinion. The input type date in html seems to not have validation built in the browser and contrary to min & max for the date GUI the browser handles. I'm thinking if I were to implement something for the regular typing field when user types instead of clicking the calendar, well there are two ways of going about it and I'm unsure about it.

One way is to have the validation done outside the component (in the module code) and when the form is submitted an error is shown if triggered. I don't like this because then each module has to take care of it.

Second way would be validation in the component but has a typing timer countdown and when the user stops typing it assumes that's the date and throws an error if it doesn't fit between the min or max. Thoughts?

johnsaigle:
Hmm I'm not really sure. Maybe we can include that as a separate issue - let's get this PR merged as it is now and improve upon it later.

jsx/Form.js
```diff
@@ -1062,15 +1062,19 @@ class DateElement extends Component {
       maxYear = '9999';
     }
 
+    const currentDate = new Date();
+    const currentDay = ('0' + currentDate.getDate()).slice(-2);
+    const currentMonth = ('0' + (currentDate.getMonth() + 1)).slice(-2);
```
driusan, line 1067: What are you trying to do with this '0' hack? What happens in November or December? Or day of the month >= 10?

maltheism: @driusan it keeps a '0' if single digit response for the day or month. So September would be 09 instead of 9. Similar for the day of the month. If the day is 7 the currentDay will be 07. If the day is 11 the currentDay will be 11.

ridz1208:
@maltheism can you update testing instructions for other features in this PR (dob not required with EDC for example)

jsx/Form.js
ridz1208: @maltheism i think it would work but could you please add a comment line in the code describing that it's needed because getmonth and getdate return values need to be padded before saving

modules/new_profile/jsx/NewProfileIndex.js
```diff
@@ -146,17 +146,21 @@ class NewProfileIndex extends React.Component {
     let pscid = null;
     let site = null;
     let minYear = this.state.configData.minYear;
-    let maxYear = this.state.configData.maxYear;
+    let today = (new Date()).getFullYear();
```
ridz1208, line 149: `today` should probably called `thisYear` instead

modules/new_profile/jsx/NewProfileIndex.js
```diff
@@ -146,17 +146,21 @@ class NewProfileIndex extends React.Component {
...
+    let maxYear = (this.state.configData.maxYear > today)
```
ridz1208, line 150: this should probably be called `dobMaxYear` since it's not the maxyear of EDC

modules/new_profile/php/new_profile.class.inc
```diff
@@ -205,13 +207,11 @@ class New_Profile extends \NDB_Form
...
-        }
-        if (empty($values['dobDateConfirm'])) {
-            $errors['dobDateConfirm'] = 'Confirm Date of Birth field is required.';
-        }
-        if ($values['dobDate'] != $values['dobDateConfirm']) {
+        // Validate DOB fields
+        if (isset($values['dobDate'])
+            && isset($values['dobDateConfirm'])
+            && $values['dobDate'] != $values['dobDateConfirm']
+        ) {
```
ridz1208, line 214: you are missing a validation for when DOB is missing and EDC is not on (I know it get handled in react but malicious users can modify the JS validation) server validation is still required

modules/new_profile/php/new_profile.class.inc
```diff
@@ -205,13 +207,11 @@ class New_Profile extends \NDB_Form
...
-        if (empty($values['dobDate'])) {
-            $errors['dobDate'] = 'Date of Birth field is required.';
-        }
-        if (empty($values['dobDateConfirm'])) {
-            $errors['dobDateConfirm'] = 'Confirm Date of Birth field is required.';
-        }
-        if ($values['dobDate'] != $values['dobDateConfirm']) {
+        // Validate DOB fields
+        if (isset($values['dobDate'])
+            && isset($values['dobDateConfirm'])
```
ridz1208, line 212: FYI you can do `isset($values['dobDate'],$values['dobDateConfirm'])`

isset() accepts multiple parameters

ridz1208:
@driusan this is changing functionality, making DOB not required when EDC is on, it's intentional but I'm wondering where in loris this could cause errors... calculate age in instruments probably... anywhere else ?

modules/new_profile/php/new_profile.class.inc
maltheism: @ridz1208 validation happens on line 227 and I don't believe I wrote that code.

ridz1208:
Added 
- **blocked** for PR #5922 that contains part of these changes
- **discussion required** for potential other necessary changes due to DOB not being required anymore

otherwise changes look good

jsx/Form.js
driusan: I think it should be written in a more straightforward fashion.

ridz1208:
@maltheism I think you can add the changes for getcandidateage in this PR if you want to. In order to make the changes complete
otherwise I'm worried this would fail testing

driusan:
@ridz1208 since it's going to the main branch I think it's fine to catch things that are missed after it's merged as long as we fix the things that we can think of.

jsx/Form.js
maltheism: @driusan what do you think of the latest change? I found that the attaching a zero can be replaced with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

ridz1208:
@laemtl could you give this a test please ?

maltheism:
thanks @laemtl!

modules/new_profile/php/new_profile.class.inc
```diff
@@ -238,6 +239,9 @@ class New_Profile extends \NDB_Form
                 $errors['edcDate'] = 'Expected Date of Confinement is not a '
                     . 'valid date.';
             }
+        } else if (empty($values['dobDate'])) {
+            $errors['dobDate'] = "Date of Birth cannot be empty
+            if Expected Date isn't configured.";
```
driusan, line 244: I don't think the "if" part should be in this error message. Users don't know anything about the study's LORIS configuration, so it's confusing and unnecessary.

maltheism: @driusan fixed in latest commit

ridz1208:
@driusan please re review

modules/new_profile/php/new_profile.class.inc
```diff
@@ -64,7 +64,9 @@ class New_Profile extends \NDB_Form
         /* Create the candidate and retrieve its ID.
          * Use form values when present, otherwise default to null.
          */
-        $dob = (new \DateTime($values['dobDate']))->format('Y-m-d');
+        $dob = isset($values['dobDate'])
```
driusan, line 67: use an if statement, not a ternary

modules/new_profile/php/new_profile.class.inc
```diff
@@ -64,7 +64,9 @@ class New_Profile extends \NDB_Form
...
+            ? (new \DateTime($values['dobDate']))->format('Y-m-d')
+            : null;
```
ridz1208, line 69:
```suggestion
         $dob = null;
         if (isset($values['dobDate'])) {
             $dob = (new \DateTime($values['dobDate']))->format('Y-m-d')
         }
```

ridz1208:
@driusan made the changes, waiting for travis

maltheism:
@ridz1208 you had extra spaces so I removed.

ridz1208:
@maltheism thank you

Issue #5701, opened 2019-11-20
[instrument_list] Can't select "Send to DCC" even with initial + DDE complete and no conflicts


leighevans:
"Send to DCC" option not available even once initial and DDE are set to complete within a visit (and there are no conflicts). Permission to Send to DCC not the issue either (ensured I have it). 
Screenshots attached:

**Send to DCC option not available even with complete data for initial and DDE:**
![Screen Shot 2019-11-20 at 11 12 17 AM](https://user-images.githubusercontent.com/32552953/69256639-cada0e00-0b87-11ea-9e30-c54fa40c7f6c.png)

**No conflicts exist for this candidate preventing Send to DCC option:**
![Screen Shot 2019-11-20 at 11 12 57 AM](https://user-images.githubusercontent.com/32552953/69256683-d9c0c080-0b87-11ea-98b4-65afa0a697b8.png)

ridz1208:
this was not reproducible on other visits, turns out the issue was because of the RB data. the submitted column in the `session` table was null in some places making the code think it has been submitted.

will issue a PR to resolve

driusan:
PR merged

PR #5721, opened 2019-11-21
[RB] Updated migration file

ridz1208:
## Brief summary of changes

`2019-07-04-remove_header_row_from_parameter_file_and_convert_back_to_Value_field_to_text.sql` had not been applied.

PR #5717, opened 2019-11-21
[SQL] Release patches archiving and concatenation

ridz1208:
## Brief summary of changes
Preparation for release 22.0

Concatenation of the patches was done using the following commands
```
mkdir SQL/Archive/22.0
git mv SQL/New_patches/*.sql SQL/Archive/22.0/
for n in SQL/Archive/22.0/*.sql; do echo $n; echo -e "\nSELECT 'Running: $n';\n" >> SQL/Release_patches/21.0_To_22.0_upgrade.sql; cat $n >> SQL/Release_patches/21.0_To_22.0_upgrade.sql; done;```

PR #5760, opened 2019-11-25
[instrument_list] Change check for submitted and populate RB

ridz1208:
## Brief summary of changes
Addresses: https://github.com/aces/Loris/issues/5701 in 3 different ways

1. The RB data had `NULL` values under the `Submitted` column. however the default in Timepoint class is set to `N` [See Timpoint class](https://github.com/aces/Loris/blob/22.0-release/php/libraries/TimePoint.class.inc#L248). This PR adds the `N` value wherever `Submitted=NULL`
2. The check for submitted in the instrument_list class was done through a double negation instead of simply using the `isSubmitted()` function.
3. The default in SQL for `Submitted` is NULL and thus the database does not enforce the rules in the timepoint class. This PR adds the SQL part. 

1. try to reproduce error in the linked issue by setting the Submitted column to null for a session and see if the data can be submitted.


SQL/New_patches/2019-11-25-Default_value_for_session_submitted.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE session SET Submitted='N' WHERE Submitted IS NULL;
+ALTER TABLE session CHANGE Submitted Submitted ENUM('Y','N') DEFAULT 'N' NOT NULL;
```
driusan: Since this is going to the 22.0-release branch and #5717 is merged can you update the concatenated patch and put this straight into archived?

ridz1208: do we want to merge it though ?

hasnt been tested and not really critical ? might be changing functionality

ridz1208: also could you merge https://github.com/aces/Loris/pull/5721

driusan: not merging it would mean it can't go into a bugfix since it requires an SQL patch, so I'd err on the side of merging it because making it not-nullable with a default makes sense to me.

driusan: #5721 is already merged

PR #5762, opened 2019-11-25
Update phan to latest version

johnsaigle:


johnsaigle:
We may as well wait until the next version of phan is released to do this because we need it for #5651

driusan:
superceded by #5891

Issue #5643, opened 2019-11-13
[Install] Script fails when not run in `tools/` directory


johnsaigle:
It's documented, but it would be good to have the script exit early and say something like "please run this script from the `tools` directory"

PR #5764, opened 2019-11-25
[install.sh] Script fails when not run in `tools/` directory

maltheism:
## Brief summary of changes

Make the script fail when ran outside of the tools directory.

1. Attempt to run the install.sh from outside the tools directory.

* https://github.com/aces/Loris/issues/5643

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -9,6 +9,12 @@
 
 set -euo pipefail
 
+# Script must be run from tools directory.
+if [[ "$PWD" != *'/tools'* ]]; then
+    echo "Please run this script from the tools directory."
+    exit 2
+fi
+
 # Must be run interactively.
 if ! test -t 0 -a -t 1 -a -t 2 ; then
     echo "This installation program should be run interactively."
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -10,7 +10,7 @@
 set -euo pipefail
 
 # Script must be run from tools directory.
-if [[ ${PWD:(-6)} != *'/tools'* ]]; then
+if [[ "$PWD" != *'/tools' ]]; then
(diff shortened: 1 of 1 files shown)


tools/install.sh
```diff
@@ -9,6 +9,12 @@
 
 set -euo pipefail
 
+# Script must be run from tools directory.
+if [[ "$PWD" != *'/tools'* ]]; then
+    echo "Please run this script from the tools directory."
+    exit 2
```
```suggestion
    exit 1
```

The error codes have a specific meaning for bash and I don't think 2 matches what's happening here.

maltheism:
@PapillonMcGill @johnsaigle what if they run it from a directory that's named tools but is not the tools directory of Loris?

johnsaigle:
I think that @PapillonMcGill's suggestion is good enough. If someone manually copies the install script into an unrelated folder also called tools well.... that's on them.

maltheism:
@johnsaigle @PapillonMcGill how about `if [[ "$PWD" != *'oris/tools'* ]]; then` because I'm having trouble finding a way to get the last part of the string in bash. I doubt there will be a case when a user has oris/tools in the path without being in the tools directory.

tools/install.sh
```diff
@@ -9,6 +9,12 @@
 
 set -euo pipefail
 
+# Script must be run from tools directory.
+if [[ ${PWD:(-6)} != *'/tools'* ]]; then
```
```suggestion
if [[ "$PWD" != *'/tools' ]]; then
```

Can you see if this works?

maltheism: @johnsaigle I just added a commit that works for getting the last six characters of the string for comparison.

johnsaigle: I know, I'd prefer not to use magic numbers though.

Issue #5705, opened 2019-11-20
[imaging_browser] Empty dropdown for "Overall" in QC Comments


leighevans:
Might be a non-essential cosmetic issue, but the dropdown menu for "Overall" at the bottom of the QC Comments window is empty. On CCNA at least (currently 20.1.2), there is no dropdown for Overall, only checkboxes (screenshot attached), so maybe a dropdown accidentally got added when there shouldn't be one?

LORIS 22.0
![Screen Shot 2019-11-20 at 1 51 21 PM](https://user-images.githubusercontent.com/32552953/69268378-0206ea00-0b9d-11ea-9624-455ce9432b9c.png)

CCNA on 20.1.2
![Screen Shot 2019-11-20 at 1 54 13 PM](https://user-images.githubusercontent.com/32552953/69268670-996c3d00-0b9d-11ea-92ad-8cc0879ec029.png)

cmadjar:
PR merged.

PR #5767, opened 2019-11-25
[feedback_mri_comments] Removed empty Overall drop down

cmadjar:
## Brief summary of changes

There should not be any drop down under the Overall MRI QC comments section like so:
image

On 22.0, there is an empty drop down for the overall section that should not be there:
image

This PR simply removes the empty drop down from the overall section of the MRI QC comments pop up page.

1. Check that no drop down are displayed next to the Overall header in the MRI QC pop up
2. Check that the other drop downs are correctly populated and can be saved

* #5705


cmadjar:
@jesscall Is this something you would have time to test by any chance?

Issue #5778, opened 2019-11-26
[Instrument Manager] Change instrument installation to require super user permission


johnsaigle:
Installing instruments is highly sensitive as it is a feature that effectively allows a user to run arbitrary SQL statements against the database via uploaded LINST files.

LISNT will be deprecated in LORIS version 23 according to the current Roadmap. Until then, this module should require `superuser` so that only the most-trusted users have access to this feature. This is the consensus reached by the team in order to retain this features which facilitates instrument installation while also mitigating the risk.

This ticket replaces #4399.

PR #5781, opened 2019-11-26
[Instrument Manager] Require 'superuser' permission for instrument upload

johnsaigle:
## Brief summary of changes

This feature should be restricted to superusers, as discussed in the LORIS meeting.

I did not remove the `instrument_manager_write` permission from the codebase as we may want to use it in the future when JSON instruments are ready.

Background: #5778

### DIFF
--- a/modules/instrument_manager/README.md
+++ b/modules/instrument_manager/README.md
@@ -37,7 +37,7 @@ following permissions:
 * `instrument_manager_read`
 * `instrument_manager_write`
 
-The write permission allows a user to upload new LORIS instrument. 
+Currently, only users with the `superuser` permission can install instruments.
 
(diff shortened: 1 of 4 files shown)


modules/instrument_manager/README.md
```diff
@@ -51,6 +51,3 @@ patch respectively.)
 In order to automatically source the SQL patch and fully configure
 LINST instruments, the LORIS `quatUser` and `quatPassword` configuration
 must be set to a user which has the MySQL `CREATE TABLE` permission.
-(The name `quatUser` and `quatPassword` is an anachronism and should
```
johnsaigle: Created issue #5782. The module README is not a place to track issues.

PR #5783, opened 2019-11-26
[Core] Add Other to possible values for candidate Sex

### LINKED ISSUE #5759: [SQL] `candidate.sex` column type does not match `tarchive.PatientSex` column
The candidate.sex column is an enum with `Male` and `Female` as possible values.

The tarchive.PatientSex column is a varchar(255). Raisinbread contains values `M`, `F`, and `O`.

The second case implies an "Other" value for sex which may have impact on study ontologies. 

These columns should be made consistent.

johnsaigle:
## Brief summary of changes

Adds a value of `Other` to candidate Sex value.

* Resolves #5759


maltheism:
@johnsaigle looks good. My only question is how are the M,F,O corresponding with Male, Female, and Other. https://test-loris-dev.loris.ca/dicom_archive/?site=Data+Coordinating+Center&sex=O

johnsaigle:
In the meeting Cecile mentioned that those values come from DICOM headers and are not standardized. The field must remain a varchar and not an enum. Essentially that data doesn't correspond with LORIS's concept of a Candidate and shouldn't be updated.

johnsaigle:
Blocking #5390

PR #5789, opened 2019-11-26
Add tool to remove condescending language from LORIS documentation

johnsaigle:
## Brief summary of changes

[How to remove condescending language from documentation](https://dev.to/meeshkan/how-to-remove-condescending-language-from-documentation-4a5p)

This post outlines why condescending documentation is a bad idea and recommends the `alex` tool to fix it.

This tool should be helpful as we go forward in creating our documentation on ReadTheDocs.

This PR:

* Adds the `alex` tool to devDependencies
* Creates configuration files for this tool
* Fixes detected occurrences
* Adds `alex` to our CI pipeline to prevent future issues.

### DIFF
--- a/.alexrc.yml
+++ b/.alexrc.yml
@@ -1,0 +1,3 @@
+allow:
+    - invalid
+    - special
--- a/.alexrc.yml
+++ b/.alexrc.yml
@@ -1,3 +1,4 @@
 allow:
     - invalid
     - special
+    - executed
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,8 +77,8 @@ using the command `make checkstatic` and fix any resulting errors. Otherwise,
 would affect the data or custom code of a study - document this in your pull 
 request description and tag it with **Caveat for Existing Projects**. 
 This helps us to document our release notes.
-* In addition to the above guidelines, we use [Alex JS](https://github.com/get-alex/alex) 
-to help us create documentation that is friendly and accessible.
+* You can run the command `make checklanguage` to ensure that documentation
+you've contributed contains language that is friendly and accessible.
 
 If you're unsure about any of the above, feel free to ask us for
 clarification via the mailing list.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,6 +77,8 @@ using the command `make checkstatic` and fix any resulting errors. Otherwise,
 would affect the data or custom code of a study - document this in your pull 
 request description and tag it with **Caveat for Existing Projects**. 
 This helps us to document our release notes.
+* In addition to the above guidelines, we use [Alex JS](https://github.com/get-alex/alex) 
+to help us create documentation that is friendly and accessible.
 
 If you're unsure about any of the above, feel free to ask us for
 clarification via the mailing list.
--- a/Makefile
+++ b/Makefile
@@ -22,11 +22,17 @@ clean:
 	rm -rf vendor
 
 # Perform static analysis checks
+#
+# The 'alex' tool scans documentation for condescending language.
+# Arguments:
+#     --quiet Shows only warnings and errors
+#     --why   Explains why something is problematic
+#     --diff  On Travis, only scans files changed
 checkstatic: phpdev
 	npm run lint:php
 	npm run lint:javascript
 	vendor/bin/phan
-	npx alex
+	npx alex --quiet --why --diff
 
 unittests: phpdev
 	vendor/bin/phpunit --configuration test/phpunit.xml
--- a/Makefile
+++ b/Makefile
@@ -22,16 +22,17 @@ clean:
 	rm -rf vendor
 
 # Perform static analysis checks
-#
+checkstatic: phpdev
+	npm run lint:php
+	npm run lint:javascript
+	vendor/bin/phan
+
 # The 'alex' tool scans documentation for condescending language.
 # Arguments:
 #     --quiet Shows only warnings and errors
 #     --why   Explains why something is problematic
 #     --diff  On Travis, only scans files changed
-checkstatic: phpdev
-	npm run lint:php
-	npm run lint:javascript
-	vendor/bin/phan
+make checklanguage:
 	npx alex --quiet --why --diff
 
 unittests: phpdev
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ checkstatic: phpdev
 	npm run lint:php
 	npm run lint:javascript
 	vendor/bin/phan
+	npx alex
 
 unittests: phpdev
 	vendor/bin/phpunit --configuration test/phpunit.xml
--- a/README.Mac.md
+++ b/README.Mac.md
@@ -193,7 +193,7 @@ b. Create the file `ext-xdebug.ini` with this command:
 sudo touch /usr/local/etc/php/<your_php_version>/conf.d/ext-xdebug.ini
 ```
 
-c. Type the following into the `ext-xdebug.ini` file you just created (Note: replace the `zend_extension` path with the correct one on your machine):
+c. Type the following into the `ext-xdebug.ini` file you created (Note: replace the `zend_extension` path with the correct one on your machine):
 ```
 [xdebug]
 zend_extension="/usr/local/Cellar/php@7.2/7.2.14/pecl/20170718/xdebug.so"
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 
 # LORIS Neuroimaging Platform
 
+<!--alex ignore easy-->
 LORIS (Longitudinal Online Research and Imaging System) is a self-hosted web application that provides data- and project-management for neuroimaging research. LORIS makes it easy to manage large datasets including behavioural, clinical, neuroimaging and genetic data acquired over time or at different sites.
 
 * Try the LORIS demo instance at https://demo.loris.ca.
--- a/README.md
+++ b/README.md
@@ -5,7 +5,6 @@
 
 # LORIS Neuroimaging Platform
 
-<!--alex ignore easy-->
 LORIS (Longitudinal Online Research and Imaging System) is a self-hosted web application that provides data- and project-management for neuroimaging research. LORIS makes it easy to manage large datasets including behavioural, clinical, neuroimaging and genetic data acquired over time or at different sites.
 
 * Try the LORIS demo instance at https://demo.loris.ca.
--- a/docs/API/LorisRESTAPI.md
+++ b/docs/API/LorisRESTAPI.md
@@ -33,8 +33,8 @@ DELETE is not supported on any resource defined in this API.
 # 1.1 Authentication
 
 If a user is logged in to Loris and can be authenticated using the standard session mechanism,
-no further authentication is required. Requests will be evaluated as requests from that user,
-so that standard Loris modules can simply use the API.
+no further authentication is required. Requests will be evaluated as requests from that user
+so that standard Loris modules are able to use the API.
 
 If a user is not logged in to Loris (for instance, in a third party app or a CORS application),
 they can be authenticated using [JSON Web Tokens](https://jwt.io).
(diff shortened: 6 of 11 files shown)


.alexrc.yml
```diff
@@ -0,0 +1,4 @@
+allow:
```
johnsaigle: These terms are potentially insensitive when used to refer to people, but are helpful in the context of discussing code, e.g.

"invalid data"
"special characters"
"executed a command"

README.md
```diff
@@ -1,5 +1,6 @@
 # [![Build Status](https://travis-ci.org/aces/Loris.svg?branch=master)](https://travis-ci.org/aces/Loris) LORIS Neuroimaging Platform 
 
+<!--alex ignore easy-->
```
johnsaigle: The word 'easy' is not helpful to include in code instructions because different tasks are more difficult depending on your background.

However, it's a nice "sales" line in the context of this paragraph

johnsaigle:
I think the warnings do cause build failures actually. I imagine this is configurable. 

I'm OK with having it cause build failures because it's quick to fix the errors and I don't think anyone will see them or fix them otherwise (e.g. the JS lint warnings).

PapillonMcGill:
Whaaatt? There are JS lint warning not resolved!     You have a good point here.

I guess it wouldn't be difficult to had an exception in case some word are flag even if correct in the context.

johnsaigle:
Yes I agree. And if a word causes problems but we agree it's fine in context then we can add it to the exclude list there.

README.md
driusan: If we need to start hacking up our documents to add machine annotation for this tool, I don't think it should be run automatically as part of Travis and it adds another dependency (both for developers and Travis). We should just be cognizant of the writing style during code review and someone can periodically run the tool to make sure nothing slipped through.

The link that you provided even says that it's intended to flag things for human review, not be used as an absolute rule.

> As tempting as it is, you don't want to remove every instance that is flagged by alex or your own search. Because, sometimes, these terms can be present out of necessity or as an attempt to be more welcoming.

(That said, the few changes identified in this PR look good to me.)

johnsaigle: Fair enough, I see your point.

Maybe we could assign someone to run the tool during our release process?

johnsaigle:
@driusan I added a commit that creates a separate `make` target for the alex script which also removes it from Travis. I also removed the annotation from README as it's no longer needed to pass Travis.

README.md
driusan: Makes sense to me, as long as we remember

christinerogers: that's always precisely the issue.   Do we have a new checklist for the release process, that we actually use?

docs/SQLModelingStandard.md
```diff
@@ -76,7 +76,7 @@ infrastructure will be re-designed to follow this guideline.*
   be named like in `OriginCountryID` and `DestinationCountryID`.*
 - Both **primary key and foreign key** fields should end with **`ID`**
   in upper case.  
-- **Date** fields should not just be named “Date”. A
+- **Date** fields should not be named “Date”. A
```
christinerogers: this use of `just` and the one above are not condescending.  their synonyms here are "only" and the support the logical construct.  I don't agree with stripping them here.

docs/SQLModelingStandard.md
```diff
@@ -26,7 +26,7 @@ infrastructure will be re-designed to follow this guideline.*
   of the words in the name should reveal the concern of the table in a 
   broad-to-specific manner. 
   - *i.e. A table containing the consent information of a candidate would be 
-    named `candidate_consent`, candidate being the broader concern and just 
+    named `candidate_consent`, candidate being the broader concern and
```
christinerogers: not the same `just` as in `simply` -- this one means `only`.  
if it is to be removed, it should be replaced by a comma to retain the clarity that it contributed to the sentence structure

README.md
johnsaigle: The GitHub project functions as a kind of distributed checklist, I guess.

Anyway I've added a `make` target for this tool (not included as a part of Travis though) so that it's quick to run the check anyway. Still a step forward.

docs/SQLModelingStandard.md
johnsaigle: @christinerogers I reworded the sentence. let me know if that works for you

johnsaigle:
@christinerogers I agree with your points. I added some commits _just_ now commits to reword the sentences to be more clear while avoiding the word `just`. Not that the word is bad in the context but I would like to be able to use the tool without ignoring all cases of the word 'just' or manually adding exceptions to these documents.

docs/SQLModelingStandard.md
```diff
@@ -76,9 +76,8 @@ infrastructure will be re-designed to follow this guideline.*
   be named like in `OriginCountryID` and `DestinationCountryID`.*
 - Both **primary key and foreign key** fields should end with **`ID`**
   in upper case.  
-- **Date** fields should not be named “Date”. A
-  **qualifier** like “BoughtDate” or “DateAccepted” should be added as
-  required.  
+- **Date** fields must have **qualifers** (i.e. they should not be named “Date”). A
```
PapillonMcGill: I like it, and it is not in the negative tone :+1:

docs/SQLModelingStandard.md
christinerogers: sure, and thanks

docs/SQLModelingStandard.md
christinerogers: typo: `qualifer`

docs/API/LorisRESTAPI.md
```diff
@@ -33,8 +33,8 @@ DELETE is not supported on any resource defined in this API.
 # 1.1 Authentication
 
 If a user is logged in to Loris and can be authenticated using the standard session mechanism,
-no further authentication is required. Requests will be evaluated as requests from that user,
-so that standard Loris modules can simply use the API.
+no further authentication is required. Requests will be evaluated as requests from that user
+so that standard Loris modules are able to use the API.
```
christinerogers: @johnsaigle   I'm a little amused here -- is it really condescending to use "simply" here?  it describes a more simple access solution for the modules, not an instruction to users which could have negative connotations.

README.Mac.md
```diff
@@ -193,7 +193,7 @@ b. Create the file `ext-xdebug.ini` with this command:
 sudo touch /usr/local/etc/php/<your_php_version>/conf.d/ext-xdebug.ini
 ```
 
-c. Type the following into the `ext-xdebug.ini` file you just created (Note: replace the `zend_extension` path with the correct one on your machine):
+c. Type the following into the `ext-xdebug.ini` file you created (Note: replace the `zend_extension` path with the correct one on your machine):
```
christinerogers: Again -- this is not the use of `just` we're looking to expunge...

docs/SQLModelingStandard.md
```diff
@@ -76,9 +76,8 @@ infrastructure will be re-designed to follow this guideline.*
   be named like in `OriginCountryID` and `DestinationCountryID`.*
 - Both **primary key and foreign key** fields should end with **`ID`**
   in upper case.  
-- **Date** fields should not just be named “Date”. A
-  **qualifier** like “BoughtDate” or “DateAccepted” should be added as
-  required.  
+- **Date** fields must have **qualifers** (i.e. they should not be named “Date”). A
```
christinerogers: typo fix: 
```suggestion
- **Date** fields must have **qualifiers** (i.e. they should not be named “Date”). A
```

johnsaigle:
> Add this link to our Contributing.md in a sentence like "Please note we try to avoid using certain language in our codebase, for details see this list", accompanied by:

I'd be open to a "We use a tool called Alex to help us create accessible and friendly documentation. For more details please view their GitHub repository at [link]"

> a hotlink to the actual list of flagged words, for clarity (e.g. should...)

It uses various lists, including ones that contain e.g. racist slurs so I don't think we should do this. A link to the tool is sufficient imo.

> modules can simply connect via the API

Is this really so simple though? I have 3 years of experience here and I'm not 100% confident I could explain to someone how to do this. I don't think "simply" adds anything to this sentence and functions at best as a buzzword and at worst as a source of frustration. 

I agree with you it's not insensitive but the motivation of the tool aren't limited to words that are disrespectful. it's more subtle as well as more broad than that use case.

Rewording this kind of sentence would be an improvement for this reason. 

> Yes we can find substitutes but why not just flag for validation in the commit process instead of expurgation?

The idea here is to be able to run the tool and scan for problematic use of language. This tool will generate a report based on everything it thinks is an issue. In some cases these will be false positives, as you've noted and I agree with what you're saying. 

At the same time, I'd prefer reword these few instances so that the report contains all signal and no noise when we use it down the line. If the tool is to be useful it shouldn't contain a lot of noise. An example is the ESLint warnings that have been in the repo for years now but nobody cleans up because there's no incentive to do so.

README.Mac.md
johnsaigle: 3 approaches in this case:

1. Reword the sentence 
```suggestion
c. Type the following into the `ext-xdebug.ini` file create the above command (Note: replace the `zend_extension` path with the correct one on your machine):
```

2. Add an exception like `<!--alex ignore just-->`, but Dave asked me not to do this.

3. Let this be flagged each time the tool is run, generating noise in the report that we have no intention of fixing.

My favourite solution is the 1st.

johnsaigle:
@driusan @christinerogers Could you please update your reviews?

docs/HelpStyleGuide.md
```diff
@@ -66,3 +66,6 @@ This means using more direct language and tone to get information across to the
 * Break up large blocks of text. It's more visually pleasing and less intimidating.
 
 For more information, including research studies conducted, on the use and value of Plain Language, please see [this article](https://www.nngroup.com/articles/plain-language-experts/) and feel free to browse the rest of this website for further helpful details, tips, and tricks. 
+
+In addition to the above guidelines, we use [Alex JS](https://github.com/get-alex/alex)$
+to help us create documentation that is friendly and accessible.
```
driusan: I don't think this advertisement for a third party tool belongs here.

Makefile
```diff
@@ -27,6 +27,14 @@ checkstatic: phpdev
 	npm run lint:javascript
 	vendor/bin/phan
 
+# The 'alex' tool scans documentation for condescending language.
+# Arguments:
+#     --quiet Shows only warnings and errors
+#     --why   Explains why something is problematic
+#     --diff  On Travis, only scans files changed
+make checklanguage:
+	npx alex --quiet --why --diff
```
driusan: There should probably be a note in the CONTRIBUTING.md suggesting that this can be run to check the language, otherwise it's unlikely to be run by anyone.

docs/HelpStyleGuide.md
johnsaigle: @christinerogers asked that it be mentioned somewhere in the documentation. It felt most relevant to include it in a document outlining our documentation style.

I don't think it's an advertisement so much as just a fact about the project, similar to saying "we use phan to make sure our code doesn't suck"

christinerogers: "and if you want to know why we might change particualr language you use, here's the info... "

Makefile
PapillonMcGill: are you sure about  `npx alex ...`?
```

Comment (johnsaigle) [reply Makefile]: Yes, I'm sure. It was a little confused at first but if you test it that's the correct command.

File: docs/HelpStyleGuide.md (hunk 9 above)
Comment (johnsaigle) [reply docs/HelpStyleGuide.md]: I feel that providing the link to the tool implies indicates that that's the place to find info.

File: Makefile (hunk 10 above)
Comment (PapillonMcGill) [reply Makefile]: np_x_?

Comment (maltheism) [reply Makefile]: @PapillonMcGill https://www.npmjs.com/package/npx

Comment (johnsaigle) [reply Makefile]: Yes, np**x** is correct.

Comment (johnsaigle) [reply Makefile]: Done

Comment (PapillonMcGill) [reply Makefile]: Ok, just wanted to be sure it was not a type.

File: CONTRIBUTING.md [hunk 11]
```diff
@@ -76,6 +76,8 @@ using the command `make checkstatic` and fix any resulting errors. Otherwise,
 would affect the data or custom code of a study - document this in your pull 
 request description and tag it with **Caveat for Existing Projects**. 
 This helps us to document our release notes.
+* In addition to the above guidelines, we use [Alex JS](https://github.com/get-alex/alex)
```
driusan: This still shouldn't be advertising external tools. The HelpStyleGuide mention of it is fine, but for CONTRIBUTING.md the contributor should just need to know that they can run `make checklanguage` to verify the documentation.

johnsaigle: Done

PR #5795, opened 2019-11-27
[Core] Remove unused configuration settings

### LINKED ISSUE #3879: Update Config module help text with link to Issue Tracker
From Redmine ticket filed by @christinerogers 

> Extremely easy pull request to fix a Config setting description(hover/tooltip) that doesn't make sense anymore :

>Current `issue_tracker_url` Config setting description:

    "The **new** bug/issue tracker url"

>This was relevant ~17.1 when this PR was created, and `mantis_url` setting still existed, and the Issue Tracker was new.

This should be replaced with a link to the `Issue Tracker` module.

### LINKED ISSUE #4100: Remove unused 'log' config setting
The 'log' config setting isn't used anywhere but the `Message` and `Log` class which are now removed from the major branch (#3999, #4041).

We should remove traces of it from LORIS. Places to scrub:
* Configuration module
* DB Schema (probably 0000-00-03)
* ?

The actual folder that this setting points to, `tools/logs/`, is still used in instrument scoring and should not be touched.

### LINKED ISSUE #5706: Remove outdated configuration settings
* Issue tracker URL -- it should always be `/issue_tracker/` no need to customize
* File Group for uploads -- is this used?
* Table rows per page -- overridden by module code, I believe
* Logs -- not used by any module #4100 
* CSS -- A FIXME in the middleware says that it's not actually supported

johnsaigle:
## Brief summary of changes

We have some configuration settings that are leftover from LORIS's younger days. This PR cleans them up.

* Resolves #4100 
* Resolves #5706 
* Resolves #3879

### DIFF
--- a/SQL/New_patches/2019-11-27-RemoveOldConfigSettings.sql
+++ b/SQL/New_patches/2019-11-27-RemoveOldConfigSettings.sql
@@ -1,0 +1,4 @@
+DELETE FROM ConfigSettings WHERE Name='Logs';
+DELETE FROM ConfigSettings WHERE Name='css';
+DELETE FROM ConfigSettings WHERE Name='issue_tracker_url';
+DELETE FROM ConfigSettings WHERE Name='FileGroup';
--- a/php/libraries/File_Upload.class.inc
+++ b/php/libraries/File_Upload.class.inc
@@ -271,7 +271,6 @@ class File_Upload
             );
             return false;
         } else if (!file_exists($dest_dir) && $this->createDirectories) {
-            $config = NDB_Config::singleton();
             mkdir($dest_dir, 0775);
         } else if (!file_exists($dest_dir) && $this->createDirectories) {
             $this->setError(
--- a/php/libraries/File_Upload.class.inc
+++ b/php/libraries/File_Upload.class.inc
@@ -273,10 +273,6 @@ class File_Upload
         } else if (!file_exists($dest_dir) && $this->createDirectories) {
             $config = NDB_Config::singleton();
             mkdir($dest_dir, 0775);
-            $grp = $config->getSetting('FileGroup');
-            if (!empty($grp)) {
-                chgrp($dest_dir, $grp);
-            }
         } else if (!file_exists($dest_dir) && $this->createDirectories) {
             $this->setError(
                 $this->fieldName,
--- a/SQL/0000-00-03-ConfigTables.sql
+++ b/SQL/0000-00-03-ConfigTables.sql
@@ -67,7 +67,6 @@ INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType,
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'base', 'The base filesystem path where LORIS is installed', 1, 0, 'text', ID, 'Base', 1 FROM ConfigSettings WHERE Name="paths";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'extLibs', 'Path to external libraries', 1, 0, 'text', ID, 'External libraries', 3 FROM ConfigSettings WHERE Name="paths";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'DownloadPath', 'Where files are downloaded', 1, 0, 'text', ID, 'Downloads', 4 FROM ConfigSettings WHERE Name="paths";
-INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'log', 'Path to logs (relative path starting from /var/www/$projectname)', 1, 0, 'text', ID, 'Logs', 2 FROM ConfigSettings WHERE Name="paths";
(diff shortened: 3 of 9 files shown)


christinerogers:
Per my question in #5706 -- 
> Re removing the Issue tracker setting -- Where/when was this feature discussed/decided as no longer relevant?
> The intention was to let this be the default, but allow links to outside issue trackers.
(linked) from the Imaging Browser : View Session page.

johnsaigle:
@christinerogers I commented on the issue. Maybe we can discuss this point in tomorrow's meeting?

johnsaigle:
@driusan @ridz1208 Explicitly requesting review from the two of you since you said the issue tracker point would be discussed in a roadmap meeting at some point. let me know which was the decision falls and I'll update this PR accordingly.

johnsaigle:
@driusan @ridz1208 Bump since two months have passed

SQL/New_patches/2019-11-27-RemoveOldConfigSettings.sql
```diff
@@ -0,0 +1,4 @@
+DELETE FROM ConfigSettings WHERE Name='Logs';
+DELETE FROM ConfigSettings WHERE Name='css';
+DELETE FROM ConfigSettings WHERE Name='issue_tracker_url';
+DELETE FROM ConfigSettings WHERE Name='FileGroup';
```
driusan: This should go into cleanup patches, and wasn't there some controversy about `issue_tracker_url`?

SQL/New_patches/2019-11-27-RemoveOldConfigSettings.sql
```diff
@@ -0,0 +1,4 @@
+DELETE FROM ConfigSettings WHERE Name='Logs';
+DELETE FROM ConfigSettings WHERE Name='css';
```
driusan: This should not be removed. Your code changes to remove this are making it far more complicated. There's now multiple sources scattered across the code base referencing the file instead of a single source of truth.

SQL/New_patches/2019-11-27-RemoveOldConfigSettings.sql
driusan: This (FileGroup) appears to not be unused, you're just deleting the code that uses it and removing the feature in this PR.

johnsaigle: From what I recall there was meant to be a roadmap meeting where you and Rida (and Samir?) were going to discuss what to do about it.

johnsaigle: Where is it used?

driusan: In the 5 lines deleted in  php/libraries/File_Upload.class.inc  in this PR.

SQL/New_patches/2019-11-27-RemoveOldConfigSettings.sql
johnsaigle: > CSS -- A FIXME in the middleware says that it's not actually supported

I wrote this in the original issue a long time ago. Do you remember this? I feel like you're probably the author.

I don't mind keeping it if it's used.

SQL/New_patches/2019-11-27-RemoveOldConfigSettings.sql
johnsaigle: OK good point. I haven't really looked at this since November. I'll revert this.

SQL/New_patches/2019-11-27-RemoveOldConfigSettings.sql
driusan: I don't mind the feature of changing the filename/the config setting being removed (and it's possible that it doesn't work if it's changed, I haven't checked..), it's the fact that this PR adds multiple scattered constants to replace what was previously a single well-defined canonical value that I think is a step backwards and I'd rather have a useless config than the maintenance nightmare of keeping them in sync.

I don't currently see any FIXME referencing it in the Middleware directory.

johnsaigle: OK this part may be outdated. I'll revert these changes.

SQL/New_patches/2019-11-27-RemoveOldConfigSettings.sql
driusan: Sorry, I don't think that ever happened. I'll try and remember to bring it up next time.

johnsaigle: Thanks!

SQL/New_patches/2019-11-27-RemoveOldConfigSettings.sql
johnsaigle: Found it:

https://github.com/aces/Loris/blob/6190c99098bd2b40892b8f72bfd85606ed381264/src/Middleware/UserPageDecorationMiddleware.php#L131-L135

php/libraries/File_Upload.class.inc
```diff
@@ -271,12 +271,7 @@ class File_Upload
             );
             return false;
         } else if (!file_exists($dest_dir) && $this->createDirectories) {
-            $config = NDB_Config::singleton();
             mkdir($dest_dir, 0775);
-            $grp = $config->getSetting('FileGroup');
```
ridz1208: it seems like you'r deleting the config and the code that uses it. why is this being removed ?

driusan:
After going through the review process, these config settings are all used, so I think this PR should be closed.

PR #5796, opened 2019-11-27
[Utility] Deprecate getExistingVisitLabels()

### LINKED ISSUE #5056: Multiple get Visit Label functions in Utility
There are several function in Utility class that returns a list of Visit labels, we need to clean up the code and use one function.

From redmine https://redmine.cbrain.mcgill.ca/issues/4291

johnsaigle:
## Brief summary of changes

There were two functions for getting visit labels, `getExistingVisitLabels()` and `getVisitList()`. 

This PR replaces the contents of `getVisitList()` with that of `getExistingVisitLabels()`. That's because the former has a better name (more concise + consistent with other functions in the class), but the latter has more robust code, including the option to filter by project.

This function is covered by unit tests so that should suffice.

* resolves #5056
* Similar cleanup done in #5780

### DIFF
--- a/modules/survey_accounts/php/addsurvey.class.inc
+++ b/modules/survey_accounts/php/addsurvey.class.inc
@@ -61,26 +61,6 @@ class AddSurvey extends \NDB_Form
     {
         $db = \Database::singleton();
 
-        // Check that all required fields are present.
-        if (!empty($values['Email']) ) {
-            if (!filter_var($values['Email'], FILTER_VALIDATE_EMAIL) ) {
-                return array('Email' => 'The email address is not valid.');
-            }
-        }
-        if (!empty($values['Email']) && !empty($values['Email2'])) {
-            if ($values['Email'] != $values['Email2']) {
-                return array('Email2' => 'The email addresses do not match.');
-            }
-        }
-
-        if (!empty($values['Email']) && empty($values['Email2'])) {
-            return array('Email2' => 'Please confirm the email address.');
-        }
-
-        if (empty($values['Test_name'])) {
-            return array('Test_name' => 'Please choose an instrument.');
-        }
-
         $numCandidates = $db->pselectOne(
             "SELECT COUNT(*) FROM candidate 
             WHERE PSCID=:v_PSCID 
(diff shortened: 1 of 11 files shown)


php/libraries/Utility.class.inc
```diff
@@ -747,41 +764,14 @@ class Utility
      * @return array of study visits in the format array('VL' => 'VL')
      *         where VL is the visit label
      *
-     * @note       THIS FUNCTION USES STRING CONCATENATION TO GENERATE AN
-     *             SQL STATEMENT.
-     *             IT IS VULNERABLE TO AN SQL INJECTION ATTACK.
```
johnsaigle: I removed this comment because I do not think it is vulnerable:

* Only fixed strings are use to extend the query. 
* pselect is used which should still prepare the only part of the query that can be supplied by a user
* The `int` type hint will throw an error on anything that isn't numeric or null and I don't think it's possible to create any kind of SQL injection attack with integers only.

modules/survey_accounts/php/addsurvey.class.inc
```diff
@@ -61,6 +61,26 @@ class AddSurvey extends \NDB_Form
     {
         $db = \Database::singleton();
 
+        // Check that all required fields are present.
```
PapillonMcGill: @johnsaigle Are you sure change to this file and  `modules/survey_accounts/test/survey_accountsTest.php` goes into this PR? Look like it should be in another PR.

php/libraries/Utility.class.inc
driusan: The comment was likely written before it was switched to pselect..

(Even if it wasn't, changing the function to just throw an exception unconditionally also makes it non-vulnerable..)

johnsaigle: Fair enough, I figured it was probably a historical thing. I figured somebody would ask about it so I wanted to lay out why this change is safe.

johnsaigle:
@PapillonMcGill @driusan I removed the changes to survey accounts.

driusan:
This is both failing Travis and needs rebase.

php/libraries/Utility.class.inc
```diff
@@ -742,42 +758,14 @@ class Utility
...
-            $ExtraProject_Criteria ORDER BY Visit_label";
-        $visitlabels    = $db->pselect($query, $qparams);
-        $visits         = array();
-        $numVisitLabels = count($visitlabels);
-        for ($i = 0; $i < $numVisitLabels; $i++) {
-            $visits[$visitlabels[$i]['Visit_label']]
-                = ($visitlabels[$i]['Visit_label']);
-        }
-        return $visits;
+        throw new \DeprecatedException(
```
HenriRabalais, line 765: If there isn't any part of the code using this function, do we even need to keep it here to display this error message?

johnsaigle: I think we do for the case where a project is using it in their custom code.

Issue #5794, opened 2019-11-27
[Issue Tracker] Raisinbread issues 


xlecours:
The issues that are closed can't be changed back to an other status. The POST request to `https://test-loris-dev.loris.ca/issue_tracker/ajax/EditIssue.php?action=edit` is responded with a 500.
The php error log on the test server should have more info about the cause of the error.

ridz1208:
why is this tagged as an RB issue not a bug ?

xlecours:
Because it's working on my VM without any extra care given to the database updates

ridz1208:
@xlecours can you detail how to reproduce ?
which issue did you click on ?

xlecours:
1. Go to https://test-loris-dev.loris.ca/issue_tracker/issue/14 as a user with both `Can add a new issue, edit own issue, comment on all` and `Can re-assign issues, mark issues as closed, comment on all, edit issues.` and with Montreal as its only site.
Change the the issue status to something else than Closed. 
Nothing happen; network log shows a 500 for the POST request.

xlecours:
More details:
```
PHP Warning:  PDOStatement::execute(): SQLSTATE[42S22]: Column not found: 1054 Unknown column 'CenterID' in 'where clause' in php/libraries/Database.class.inc on line 699, referer: https://test-loris-dev.loris.ca/issue_tracker/issue/14
```

PR #5719, opened 2019-11-21
[DICOM Archive] Skip unstable integration test

johnsaigle:
## Brief summary of changes

This test is unstable and causing failures on PRs that should otherwise pass.

#### Links to related tickets (GitHub, Redmine, ...)

* This test caused failures in unrelated PRs, including #5717, #5699, #5634, #5690, #5591 and others referenced below

PR #5798, opened 2019-11-27
Fix issue tracker only pscid provided

xlecours:
fixes https://github.com/aces/Loris/issues/5794


johnsaigle:
Transitory failure caused by #5719. Restarting

Issue #5801, opened 2019-11-27
[Candidate parameters] Cannot save consent to study status


nicolasbrossard:
On the test VM, I accessed a candidate for which consent information was never provided. I select 'Yes' to consent to the Study and enter a valid date. When I click `Update`, I get an error popup that says `Update failed`. I see in the server's error log a coding bug:

```
[Wed Nov 27 14:01:18.748132 2019] [php7:error] [pid 24309] [client 192.168.122.1:33176] PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Candidate::singleton() must be an instance of LORIS\\StudyEntities\\Candidate\\CandID, string given, called in /var/www/Loris/modules/candidate_parameters/ajax/formHandler.php on line 401 and defined in /var/www/Loris/php/libraries/Candidate.class.inc:52\nStack trace:\n#0 /var/www/Loris/modules/candidate_parameters/ajax/formHandler.php(401): Candidate::singleton('300001')\n#1 /var/www/Loris/modules/candidate_parameters/ajax/formHandler.php(52): editConsentStatusFields(Object(Database), Object(User))\n#2 /var/www/Loris/htdocs/AjaxHelper.php(111): require('/var/www/Loris/...')\n#3 {main}\n  thrown in /var/www/Loris/php/libraries/Candidate.class.inc on line 52, referer: https://test-loris-dev.loris.ca/candidate_parameters/?candID=300001&identifier=300001
```

nicolasbrossard:
Fixed by #5802

PR #5802, opened 2019-11-27
Fixed bug that prevented any consent status from being saved.

nicolasbrossard:
## Brief summary of changes

When accessing the candidate parameter page, there was a bug preventing you from saving any consent change you perform. This PR fixes that.

#5801

### DIFF
--- a/modules/candidate_parameters/ajax/formHandler.php
+++ b/modules/candidate_parameters/ajax/formHandler.php
@@ -13,6 +13,9 @@
  * @link     https://github.com/aces/Loris-Trunk
  */
 
+use \LORIS\StudyEntities\Candidate\CandID;
+
+
 $user = \User::singleton();
 if (!$user->hasPermission('candidate_parameter_edit')) {
     header("HTTP/1.1 403 Forbidden");
@@ -396,7 +399,7 @@ function editConsentStatusFields($db, $user)
     // Get CandID
     $candIDParam = $_POST['candID'];
     $candID      = (isset($candIDParam) && $candIDParam !== 'null') ?
-        $candIDParam : null;
+        new CandID($candIDParam) : null;
 
     $candidate   = \Candidate::singleton($candID);
     $currentUser = \User::singleton();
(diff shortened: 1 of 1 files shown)


modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -396,7 +399,7 @@ function editConsentStatusFields($db, $user)
     // Get CandID
     $candIDParam = $_POST['candID'];
     $candID      = (isset($candIDParam) && $candIDParam !== 'null') ?
-        $candIDParam : null;
+        new CandID($candIDParam) : null;
 
     $candidate   = \Candidate::singleton($candID);
```
johnsaigle: Would you be able to modify this file to return a `400` response when `$_POST['candID']` is not set?

It's not possible to pass `null` into `Candidate::singleton()`. I think it will generate a TypeError in that case.

It may be an edge-case but it would be nice to handle it properly.

nicolasbrossard: @johnsaigle Makes sense. Done!

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -395,9 +398,11 @@ function editConsentStatusFields($db, $user)
 {
     // Get CandID
     $candIDParam = $_POST['candID'];
-    $candID      = (isset($candIDParam) && $candIDParam !== 'null') ?
-        $candIDParam : null;
-
+    if (!isset($candIDParam) || $candIDParam === 'null') {
+        http_response_code(404);
```
```suggestion
        http_response_code(400);
```
close! should be "Bad Request", not "Not Found"

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -395,9 +398,11 @@ function editConsentStatusFields($db, $user)
...
+        die(json_encode(["error" => "Appointment does not exist."]));
```
```suggestion
        die(json_encode(["error" => "You must supply a CandID"]));
```

modules/candidate_parameters/ajax/formHandler.php
nicolasbrossard: Done

modules/candidate_parameters/ajax/formHandler.php
nicolasbrossard: Done

PR #5803, opened 2019-11-27
[Core] Add Alias value object

### LINKED ISSUE #5565: Create value object for Alias
Use `src/StudyEntities/Candidate/ValidatableIdentifier.php` as a template

#### Class properties

protected const MAX_LENGTH;
protected const MIN_LENGTH;
protected string $value;

#### Functions

`private function _validate(): void();` verifies:
* $value is alphabetical (no numbers or special characters)
* length of $value is between MAX_LENGTH and MIN_LENGTH

`public function __toString(): string;`
Same content as template.

`private final __constuct;`
Same content as template.

### Create child classes `SiteAlias` and `ProjectAlias`

Should extend `Alias` changing only the length values if necessary.

It's possible that these classes won't actually need to change the value but it would be good to decouple them right away just in case.

johnsaigle:
## Brief summary of changes

Adds a value object for an "Alias", either a Site or Project alias. This provides some built-in validation.

* Resolves #5565

### DIFF
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -14,6 +14,7 @@
 
 use LORIS\StudyEntities\Candidate\CandID;
 use LORIS\StudyEntities\Candidate\SiteAlias;
+use LORIS\StudyEntities\Candidate\ProjectAlias;
 /* Define constants used by the file */
 // error codes
 define('CANDIDATE_INVALID', 1);
@@ -683,22 +684,23 @@ class Candidate
 
     /**
      * Validate's that a PSCID matches the format defined by project
-     * config
+     * config.
      *
-     * @param string $PSCID         The PSCID to validate
-     * @param SiteAlias $siteAbbrev    The site of the candidate
-     *                              being validated to check
-     *                              against appropriate parts
-     *                              of PSCID definition
-     * @param ProjectAlias $projectAbbrev The project of the candidate
-     *                              being validated to check against appropriate
-     *                              parts of PSCID definition
+     * @param string        $PSCID         The PSCID to validate.
+     * @param ?SiteAlias    $siteAbbrev    The site of the candidate
+     *                                     being validated to check
+     *                                     against appropriate parts
+     *                                     of PSCID definition>
+     * @param ?ProjectAlias $projectAbbrev The project of the candidate
+     *                                     being validated to check against
+     *                                     appropriate parts of PSCID
+     *                                     definition.
      *
-     * @return boolean True if PSCID is valid, false otherwise
+     * @return bool True if PSCID is valid, false otherwise.
      */
     static public function validatePSCID(
         string $PSCID,
-        ?SiteAlias $siteAbbrev = null
+        ?SiteAlias $siteAbbrev = null,
         ?ProjectAlias $projectAbbrev = null
     ): bool {
 
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -13,6 +13,7 @@
  */
 
 use LORIS\StudyEntities\Candidate\CandID;
+use LORIS\StudyEntities\Candidate\SiteAlias;
 /* Define constants used by the file */
 // error codes
 define('CANDIDATE_INVALID', 1);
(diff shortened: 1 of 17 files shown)


johnsaigle:
Adding needs work because this should have a unit test.

php/libraries/Candidate.class.inc
```diff
@@ -682,23 +684,24 @@ class Candidate
...
-     *                              parts of PSCID definition
-     *
-     * @return boolean True if PSCID is valid, false otherwise
+     * config.
+     *
+     * @param string        $PSCID         The PSCID to validate.
+     * @param ?SiteAlias    $siteAbbrev    The site of the candidate
+     *                                     being validated to check
+     *                                     against appropriate parts
+     *                                     of PSCID definition>
```
PapillonMcGill:
```suggestion
     *                                     of PSCID definition.
```

php/libraries/ExternalIDGenerator.php
```diff
@@ -27,15 +29,15 @@ class ExternalIDGenerator extends SiteIDGenerator
     /**
      * Generates ExternalIDs.
      *
-     * @param string $siteAlias    To be appended to the ID value. Usually an
-     *                             abbreviation for the name of a site.
-     * @param string $projectAlias To be appended to the ID value. Usually an
-     *                             abbreviation for the name of a project.
+     * @param ?SiteAlias    $siteAlias    The site prefix to prepend to the ID value.
+     * @param ?ProjectAlias $projectAlias To be appended to the ID value.
```
PapillonMcGill:
```suggestion
     * @param ?ProjectAlias $projectAlias The project prefix to prepend to the ID value.
```
To be consistent (see php/libraries/PSCIDGenerator.php lines 33-36).

php/libraries/Utility.class.inc
```diff
@@ -957,6 +959,28 @@ class Utility
         return implode('', $pieces);
     }
 
+    /**
+     * Generate a random alphanumeric string.
```
```suggestion
     * Generate a random alphabetical string.
```

php/libraries/Project.class.inc
```diff
@@ -25,6 +26,13 @@ class Project
 {
     private static $_instances = array();
 
+    /**
+     * A short name for the Project.
+     *
+     * @var ProjectAlias
+     */
+    var $alias;
```
driusan: since this is new code it should use an access modifier, not `var`

src/StudyEntities/Candidate/Alias.php
```diff
@@ -0,0 +1,69 @@
...
+     * @var int
+     */
+    public const MIN_LENGTH = 3;
+
+    /*
+     * The minimum length of the identifier.
+     *
+     * @var int
+     */
+    public const MAX_LENGTH = 3;
```
driusan: Why are all types of aliases always exactly 3 characters long?

php/libraries/Site.class.inc
```diff
@@ -25,6 +26,20 @@ class Site
...
+     * @var SiteAlias
+     */
+    var $alias;
+
+    /**
+     * A short name for the MRI.
+     *
+     * @var MRIAlias
+     */
+    var $MRIAlias;
```
driusan: do we actually use/need this? The difference between a "Site Alias" and "MRI alias" has always been confusing and it doesn't seem like you're updating anything to use it in the PR.

johnsaigle: I'm not sure. I saw references to it and thought I'd include a value object for it.

src/StudyEntities/Candidate/Alias.php
johnsaigle: I wanted to use constants and just picked a number. Is there a better range?

php/libraries/Candidate.class.inc
PapillonMcGill: need to be redone.

php/libraries/ExternalIDGenerator.php
PapillonMcGill: same here

php/libraries/Utility.class.inc
PapillonMcGill: same

src/StudyEntities/Candidate/Alias.php
driusan: Why is there a range that needs to be enforced for aliases?

johnsaigle: I guess there doesn't need to be one enforced

Issue #5800, opened 2019-11-27
[new_profile] Pressing create multiple time creates multiple candidates "silently"


ridz1208:
This issue is on 21.0 as well and should be resolved there

ridz1208:
the changes in this PR were undone at some point 
https://github.com/aces/Loris/pull/4364

PR #5808, opened 2019-11-28
[new_profile] Pressing create multiple time creates multiple candidates "silently"

maltheism:
## Brief summary of changes

Solves the bug where multiple candidates were made from the frontend form submission.

**Note this implementation is similar to how it was done before.** Another way would be to create a disabled props for the button component.

1. Checkout PR.
2. Go to New Profile and press the submit button as much as you want.

* https://github.com/aces/Loris/issues/5800

### DIFF
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -1555,6 +1555,7 @@ class ButtonElement extends Component {
       <div className="row form-group">
         <div className={this.props.columnSize}>
           <button
+            id={this.props.id}
             name={this.props.name}
             type={this.props.type}
             className={this.props.buttonClass}
@@ -1569,6 +1570,7 @@ class ButtonElement extends Component {
 }
 
 ButtonElement.propTypes = {
+  id: PropTypes.string,
   name: PropTypes.string,
   label: PropTypes.string,
   type: PropTypes.string,
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -1580,6 +1580,7 @@ ButtonElement.propTypes = {
 ButtonElement.defaultProps = {
   label: 'Submit',
   type: 'submit',
+  disabled: null,
   buttonClass: 'btn btn-primary',
   columnSize: 'col-sm-9 col-sm-offset-3',
   onUserInput: function() {
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -1555,11 +1555,11 @@ class ButtonElement extends Component {
       <div className="row form-group">
         <div className={this.props.columnSize}>
           <button
-            id={this.props.id}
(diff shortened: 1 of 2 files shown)


maltheism:
Hi @zaliqarosli, I made the changes to make the code better as suggested by passing disable through props.

driusan:
@maltheism can you send this to 21.0-release? It's affecting some production instances and needs to go into a bugfix release before 22.0

maltheism:
@driusan branch was rebased to 21.0-release

driusan:
@zaliqarosli re-review?

modules/new_profile/jsx/NewProfileIndex.js
```diff
@@ -89,6 +90,9 @@ class NewProfileIndex extends React.Component {
       }
       formObject.append('fire_away', 'New Candidate');
 
+      // disable form from resubmission.
+      this.setState({disable: true});
```
HenriRabalais: I would call this something a bit more descriptive like `disableSubmit` or something to that affect. Perhaps `submitDisabled` because it is written as more a state (true/false) whereas `disable` sounds like an action (e.g. function).

HenriRabalais: Also, I think if the submission fails, the button should be re-enabled so that the user can attempt another submission.

maltheism: Thanks @HenriRabalais, I added the suggested changes.

modules/new_profile/jsx/NewProfileIndex.js
```diff
@@ -103,6 +107,8 @@ class NewProfileIndex extends React.Component {
           });
         } else {
           resp.json().then((message) => {
+            // enable form from resubmission.
```
HenriRabalais: might as well fix the typo here while i  test it manually!

maltheism: @HenriRabalais I fixed it. thanks!

jsx/Form.js
```diff
@@ -1572,6 +1573,7 @@ ButtonElement.propTypes = {
   name: PropTypes.string,
   label: PropTypes.string,
   type: PropTypes.string,
+  disabled: PropTypes.bool,
```
zaliqarosli: should we add default false below if not prop not provided?

maltheism: @zaliqarosli I think if no property is supplied there won't be a disabled added to the button.

maltheism:
@HenriRabalais @zaliqarosli @driusan everything was updated for suggestions & modifications.

PR #5809, opened 2019-11-28
[login] TRIM inputs before inserting into users or examiners

ridz1208:
## Brief summary of changes
another issue on 21

user_accounts and examiner modules trim inputs before submission. request account does not do any trimming so entries like first/last name and emails get submitted with empty spaces around them.

This is particularly problematic for examiners where an examiner created with the incorrect `Real_name` value will cause a duplication of the examiner the first time the form is saved on user accounts because the name then gets trimmed and restored in it's correct format in the examiner table

result -> 500 error due to pselectrow returning more than one row


christinerogers:
Does it need sanitizing too?

Issue #4989, opened 2019-07-30
Update singleton() declarations to use the Factory class


johnsaigle:
Occurrences like:
```php
$DB =& \Database::singleton();
```

should be changed to:
```php
$DB = \NDB_Factory::singleton()->database();
```

Going through the Factory class allows for more robust testing down the line.

AlexandraLivadas:
Here's a text file of all the files that use it and with what frequency: 

[dbSingletonListWithFrequency.pdf](https://github.com/aces/Loris/files/3448327/dbSingletonListWithFrequency.pdf)

Which of these do you think would be good to start with @johnsaigle @driusan ? I can start with the Utility class if you'd like because I already have unit tests written for it.

driusan:
I think you should start with finalizing your currently existing PRs and trying to get them merged.. after that, I don't think it matters much, either whichever one you think is easiest, or whichever one the current singleton() behaviour is blocking you the most from writing tests that you want to write.

The classes that are probably highest value to have tests for in that list are User, Candidate, TimePoint, Utility, and NDB_Menu_Filter (in no particular order.)

AlexandraLivadas:
Ok. Should I finalize the PRs that have tests written differently because of the Database::singleton() problem though?

johnsaigle:
If it simplifies the testing process for you, I think it would be fine to e.g. change all the singleton problems in Config in the same PR where you're adding tests for Config

driusan:
It's up to you if you want to go back and change those PRs to use the factory or keep them as-is, if the current tests are already working.

AlexandraLivadas:
The tests are working but there would be less clutter if I were to change to the factory way. I can do it for my Utility PR and see how it goes? I'll ask @kongtiaowang as well!

AlexandraLivadas:
Occurences like this: 

`$user = User::singleton();` 

should also be changed, to this:

`$user = $factory->user();`

This also makes unit testing easier. 

@johnsaigle Should I make a new issue for this?

johnsaigle:
Yep that'd be great!

johnsaigle:
I think this is another one that we can add to the "Code Review Checklist" document rather than keep as an open-ended issue.

PR #5815, opened 2019-11-29
[data_release/document_repository] Make upload directory configurable

### LINKED ISSUE #5062: Document Repository and Data Release should not store uploaded files in the codebase
Right now both modules create a custom `user_uploads/` directory within the source code for user uploads. I don't think there is a good reason for this (but correct me if you know why this was done).

Other modules have a config setting like `$module_upload_path` which is customizable. This is preferable to having a hard-coded path, especially one that mixes user data with source code.

ridz1208:
## Brief summary of changes

This gives the option to the users to configure where the default upload directory of data_release and document_repository modules will be. this also removes the burden on LORIS to ensure these locations are readable/writeable and follows the standard established by the other modules uploading data to loris.

## testing
- before checking out the branch, make sure you are able to upload a file to doc repo and data release. If it is not the case, you should resolve the issue as it is not due to this change.
- check out this branch and ensure you are still able to upload and download files to the 2 modules (the sql patch for existing projects is designed to be backwards compatible)
- change the upload paths from the configuration module and make sure the new paths exist on your machine and are apache writable
- try uploading downloading other documents to both modules to ensure that the functionality is unaffected.

# Important note for release

Projects should be encouraged to move their files out of the old directories and into new directories somewhere else on their data mount/drive.

### Related

* Resolves #5062

### DIFF
--- a/SQL/0000-00-03-ConfigTables.sql
+++ b/SQL/0000-00-03-ConfigTables.sql
@@ -75,8 +75,8 @@ INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType,
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'publication_uploads', 'Path to uploaded publications', 1, 0, 'web_path', ID, 'Publications', 10 FROM ConfigSettings WHERE Name="paths";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'publication_deletions', 'Path to deleted publications', 1, 0, 'web_path', ID, 'Deleted Publications', 11 FROM ConfigSettings WHERE Name="paths";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'MINCToolsPath', 'Path to the MINC tools', 1, 0, 'web_path', ID, 'Path to the MINC tools', 12 FROM ConfigSettings WHERE Name="paths";
-INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'documentRepositoryPath', 'Path to uploaded document repository files', 1, 0, 'text', ID, 'Document Repository Upload Path', 13 FROM ConfigSettings WHERE Name="paths";
-INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'dataReleasePath', 'Path to uploaded data release files', 1, 0, 'text', ID, 'Data release Upload Path', 14 FROM ConfigSettings WHERE Name="paths";
+INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'documentRepositoryPath', 'Path to uploaded document repository files', 1, 0, 'web_path', ID, 'Document Repository Upload Path', 13 FROM ConfigSettings WHERE Name="paths";
+INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'dataReleasePath', 'Path to uploaded data release files', 1, 0, 'web_path', ID, 'Data release Upload Path', 14 FROM ConfigSettings WHERE Name="paths";
 
 
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, Label, OrderNumber) VALUES ('gui', 'Settings related to the overall display of LORIS', 1, 0, 'GUI', 3);
@@ -192,8 +192,6 @@ INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/uploads/" FROM ConfigSett
 INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/publication_uploads/" FROM ConfigSettings WHERE Name="publication_uploads";
 INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/publication_uploads/to_be_deleted/" FROM ConfigSettings WHERE Name="publication_deletions";
 INSERT INTO Config (ConfigID, Value) SELECT ID, "%MINCToolsPath%" FROM ConfigSettings WHERE Name="MINCToolsPath";
-INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/document_repository_uploads/" FROM ConfigSettings WHERE Name="documentRepositoryPath";
-INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/data_release_uploads/" FROM ConfigSettings WHERE Name="dataReleasePath";
 
 
 INSERT INTO Config (ConfigID, Value) SELECT ID, "main.css" FROM ConfigSettings WHERE Name="css";
--- a/SQL/0000-00-03-ConfigTables.sql
+++ b/SQL/0000-00-03-ConfigTables.sql
@@ -75,6 +75,8 @@ INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType,
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'publication_uploads', 'Path to uploaded publications', 1, 0, 'web_path', ID, 'Publications', 10 FROM ConfigSettings WHERE Name="paths";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'publication_deletions', 'Path to deleted publications', 1, 0, 'web_path', ID, 'Deleted Publications', 11 FROM ConfigSettings WHERE Name="paths";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'MINCToolsPath', 'Path to the MINC tools', 1, 0, 'web_path', ID, 'Path to the MINC tools', 12 FROM ConfigSettings WHERE Name="paths";
+INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'documentRepositoryPath', 'Path to uploaded document repository files', 1, 0, 'text', ID, 'Document Repository Upload Path', 13 FROM ConfigSettings WHERE Name="paths";
+INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'dataReleasePath', 'Path to uploaded data release files', 1, 0, 'text', ID, 'Data release Upload Path', 14 FROM ConfigSettings WHERE Name="paths";
 
 
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, Label, OrderNumber) VALUES ('gui', 'Settings related to the overall display of LORIS', 1, 0, 'GUI', 3);
@@ -190,6 +192,8 @@ INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/uploads/" FROM ConfigSett
 INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/publication_uploads/" FROM ConfigSettings WHERE Name="publication_uploads";
 INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/publication_uploads/to_be_deleted/" FROM ConfigSettings WHERE Name="publication_deletions";
 INSERT INTO Config (ConfigID, Value) SELECT ID, "%MINCToolsPath%" FROM ConfigSettings WHERE Name="MINCToolsPath";
+INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/document_repository_uploads/" FROM ConfigSettings WHERE Name="documentRepositoryPath";
+INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/data_release_uploads/" FROM ConfigSettings WHERE Name="dataReleasePath";
 
 
 INSERT INTO Config (ConfigID, Value) SELECT ID, "main.css" FROM ConfigSettings WHERE Name="css";
--- a/modules/data_release/ajax/FileUpload.php
+++ b/modules/data_release/ajax/FileUpload.php
@@ -37,7 +37,7 @@ if ($_POST['action'] == 'upload'
         header("HTTP/1.1 500 Internal Server Error");
     } elseif (!is_writable($path)) {
         error_log(
-            "File upload failed. Default upload directory"
+            "File upload failed. Upload directory"
             . " does not appear to be writeable."
         );
         header("HTTP/1.1 500 Internal Server Error");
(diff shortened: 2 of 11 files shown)


raisinbread/RB_files/RB_Config.sql
```diff
@@ -94,5 +94,7 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (97,70,'/data-raisinbrea
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (98,93,'V1');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (99,101,'');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (102,19,'false');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (103,102,'/var/www/loris/modules/document_repository/user_uploads/');
```
johnsaigle: Maybe we can change the default to be outside of the module now that the install script doesn't create it.

PapillonMcGill:
@ridz1208 Could we do the same thing with `project/tables_sql/` and `project/instruments/`?

This way, in no situation, Apache would require writing inside the loris directory.

modules/data_release/ajax/FileUpload.php
```diff
@@ -12,36 +12,37 @@
  * @link     https://github.com/aces/Loris
  */
 
-$DB   = \Database::singleton();
-$user = \User::singleton();
+$DB     = \Database::singleton();
+$user   = \User::singleton();
```
johnsaigle: Not a big deal but could you update these to use the Factory like the line below?

ridz1208: i could... and i wanted to... I just decided to avoid "unrelated changes". I'm gonna do it though

ridz1208:
@PapillonMcGill The 2 directories you mentioned are much more complicated. Projects track and commit the data in these directories generally with git. it would be very tricky directory to decide which files get read from where if a project is using both LINST and tracked php instruments. same goes for tables_sql, some generated patches are commited to the repo

modules/data_release/ajax/FileUpload.php
johnsaigle: Seems related to me! 😄

johnsaigle: It's nice to leave code cleaner than you found it, especially when it's an [open issue](#4989 ). I think you're safe here

ridz1208: sold man stop selling

tools/install.sh
```diff
@@ -152,8 +144,6 @@ elif [[ " ${redhat[*]} " =~ " $os_distro " ]]; then
 else
     echo "$os_distro Linux distribution detected. We currently do not support this. "
     echo "Please manually change subdirectory ownership and permissions to ensure the web server can read *and write* in the following: "
-    echo "../modules/data_release/user_uploads "
-    echo "../modules/document_repository/user_uploads "
```
johnsaigle: Line 120 in this file should be removed

> # Changing group to 'www-data' or 'apache' to give permission to create directories in Document Repository module

SQL/0000-00-03-ConfigTables.sql
```diff
@@ -75,6 +75,8 @@ INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType,
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'publication_uploads', 'Path to uploaded publications', 1, 0, 'web_path', ID, 'Publications', 10 FROM ConfigSettings WHERE Name="paths";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'publication_deletions', 'Path to deleted publications', 1, 0, 'web_path', ID, 'Deleted Publications', 11 FROM ConfigSettings WHERE Name="paths";
 INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'MINCToolsPath', 'Path to the MINC tools', 1, 0, 'web_path', ID, 'Path to the MINC tools', 12 FROM ConfigSettings WHERE Name="paths";
+INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'documentRepositoryPath', 'Path to uploaded document repository files', 1, 0, 'text', ID, 'Document Repository Upload Path', 13 FROM ConfigSettings WHERE Name="paths";
```
driusan: Shouldn't these be of type `web_path`, not text, so that the config module verifies they exist?

SQL/0000-00-03-ConfigTables.sql
```diff
@@ -190,6 +192,8 @@ INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/uploads/" FROM ConfigSett
 INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/publication_uploads/" FROM ConfigSettings WHERE Name="publication_uploads";
 INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/publication_uploads/to_be_deleted/" FROM ConfigSettings WHERE Name="publication_deletions";
 INSERT INTO Config (ConfigID, Value) SELECT ID, "%MINCToolsPath%" FROM ConfigSettings WHERE Name="MINCToolsPath";
+INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/document_repository_uploads/" FROM ConfigSettings WHERE Name="documentRepositoryPath";
+INSERT INTO Config (ConfigID, Value) SELECT ID, "/data/data_release_uploads/" FROM ConfigSettings WHERE Name="dataReleasePath";
```
driusan: Why don't we leave the default empty and update the code to have reasonable defaults if the value is null?

SQL/0000-00-03-ConfigTables.sql
johnsaigle: Yes, absolutely. That's a great catch.

SQL/0000-00-03-ConfigTables.sql
johnsaigle: Do you have an idea about what those constants would be? Where would be a reasonable place on a linux server for these files to go? Maybe somewhere under `/srv/`?

driusan: I don't know what constants you're talking about.. I'm suggesting leaving them null.

SQL/0000-00-03-ConfigTables.sql
ridz1208: ouuh i forgot about that

SQL/0000-00-03-ConfigTables.sql
ridz1208: @driusan but you suggested to update the code to have reasonable defaults. what do you mean by that ?

driusan: "defaults" was the wrong word.. I meant it should behave reasonably if the value is unset (ie. provide an error message pointing to the config setting if it's required or disable the feature if it's an optional thing similarly to how to the instrument_manager disables uploading if it doesn't have write access.)

ridz1208: i see yeah that makes sens
should  I do that in this PR or... we wait for the centralised uploader ?

not sure

driusan: What's the affect on this PR if the type is changed to `web_path` and the defaults are invalid paths?

ridz1208: I have no idea, but for existing projects my patch will maintain the current directories until the project people switch it to another directory voluntarily.

My guess is when someone tries uploading something they will get a 500

johnsaigle: `web_path` only comes into affect when processing a POST Request to the config module. If the file path isn't set in the uploader, the module code will hopefully throw an error but these modules each do it in their own ad-hoc way.

The only bug that might occur is if someone tries to change a path in Config without changing the upload paths. At that point they'll need to correct every path before the submission succeeds.

ridz1208: Is null a valid webpath ?

maltheism:
@ridz1208 what about the genomic_browser module?

ridz1208:
@maltheism what about it ?

maltheism:
@ridz1208 you can upload a file in that module and it complains about a directory if one doesn't exist. Seems similar to what this PR is trying to get rid of.

ridz1208:
@maltheism no, genomics seems to have an already configurable path

SQL/0000-00-03-ConfigTables.sql
johnsaigle: https://github.com/aces/Loris/blob/4e3ee740fc4cb143f770ed4a4060377e327734ac/modules/configuration/ajax/process.php#L236-L242


So I don't think null works. In a meeting back when I was making this change we decided that an empty string will be used for default values. Discussion begins here: https://github.com/aces/Loris/pull/4023#issuecomment-464803516

Basically the calling code is responsible for doing something with the empty string when it finds it.

driusan: There should never be an empty string in the config module. If it's empty, the row should be removed from the table. If there was a regression on the value being properly normalized it should be fixed.

johnsaigle: I'm sorry I wasn't being precise in my wording. You're right i don't think a blank entry will exist in the table so I'm incorrect about the defaults.

When a blank string is submitted for a field in the Config module, it has the effect of deleting that setting row from the table. 

So the behaviour in the code is like this:

If value is valid path --> write new entry to table
If value is empty string --> delete entry from table
If value is not empty and not valid --> reject input because invalid web path supplied.

driusan: that's the expected behaviour, which means that it should go through the 'delete from table' logic and not the validPath code that you quoted above. It's not a "valid path" but the config module shouldn't reject it either.

johnsaigle: Yeah sorry my mind is still waking up. 😅 NULL should probably work as a default value.

ridz1208: okai so if i understand correctly;
 - in RB set the paths to "ideal" values `/data/...`
 - in 0000 conf files do not add an entry into the Config table and let projects set a validpath
 - in the patch for existing projects, maintain status quo and discuss in the release the recommendation of moving the directories.


agreed ?

johnsaigle: Yes, though we'll need to provide some detail about also moving the files from one destination to another, not just changing the path.

driusan: I don't know what the status quo is so I can't comment on the last point, but that sounds right.

johnsaigle: e.g. if there's `modules/document_repository/user_uploads/admin/meme.jpg` then the `admin/` folder will need to be moved to the new location or else it will be inaccessible.

ridz1208: yes for sure but that would not be an issue in the current state because existing projects will have their path point to the old path.

it should only be a recommendation on our end

ridz1208:
@johnsaigle @driusan please review

modules/data_release/ajax/FileUpload.php
```diff
@@ -12,36 +12,37 @@
...
     $settings = $factory->settings();
 
     $baseURL = $settings->getBaseURL();
+    $path    = $config->getSetting('dataReleasePath');
 
-    if (!file_exists(__DIR__ . "/../user_uploads/")) {
+    if (!file_exists($path)) {
         error_log(
-            "ERROR: File upload failed. Default user_uploads"
+            "ERROR: File upload failed. Default upload"
```
johnsaigle:
```suggestion
            "ERROR: File upload failed. Upload"
```

I think this error should be more general because it doesn't really matter whether the directory is default or not if it isn't found.

modules/data_release/ajax/FileUpload.php
```diff
@@ -12,36 +12,37 @@
...
             . " directory not found."
         );
         header("HTTP/1.1 500 Internal Server Error");
-    } elseif (!is_writable(__DIR__ . "/../user_uploads/")) {
+    } elseif (!is_writable($path)) {
         error_log(
-            "File upload failed. Default user_uploads directory"
+            "File upload failed. Default upload directory"
```
johnsaigle:
```suggestion
            "File upload failed. Upload directory"
```
Same here

modules/data_release/ajax/FileUpload.php
ridz1208: right

ridz1208:
@driusan @johnsaigle back to you

modules/document_repository/php/files.class.inc
```diff
@@ -205,7 +209,7 @@ class Files extends \NDB_Page
             $Notifier->notify($msg_data);
         }
 
-        $path = __DIR__ . "/../user_uploads/$dataDir";
+        $path = $path.$dataDir;
```
driusan: I don't think it's safe to assume that path is defined with a `/` at the end of it (same for the other places the filename is concatenated to it.) It should explicitly add it (and throw an exception or otherwise die beforehand it path is empty for security reasons.)

ridz1208: I just followed the assumption made in other uploading forms. If i make changes to this one do i make it to the others too ?

driusan: you mean other forms in this module or other forms in LORIS?

ridz1208: Loris. Like media for example

driusan: no, just fix these ones.

johnsaigle: @ridz1208 If you don't mind it might be useful to do this using a Utility function. 

e.g.

`\Utility::AppendFrontSlash(string $path)` that adds a front slash to the end of a string unless one already exists. Then we can easily use it throughout LORIS.

ridz1208: I was actually gonna look for the function that validates paths on config module. We could reuse that code to validate before upload too no ?

johnsaigle: Hmm we could temporarily I suppose. A better solution would be to rework modules to use the new File Uploader class because that has all of that validation built in. But that's a more substantial change.

driusan: I think it would be a good idea to more consistently handle this with something like `os.path.join` (Python) or `filepath.Join` (Go), but that's beyond the scope of this PR.

johnsaigle: Yeah I think I'm agreeing with you just doing the PHP equivalent. I think this works (adapted from some WordPress code):

```php
    return rtrim( $string, '/\\' ) . '/';
```

We could wrap it in a Utility function just so that we have a standard way of doing this. not necessary though.

ridz1208: Lets talk after i have questions

driusan: As far as I know PHP doesn't have a built in equivalent. I think we should write one with appropriate unit tests, but I don't think this PR is the right place to do it.

ridz1208:
@driusan @johnsaigle is that sufficient ?

modules/document_repository/php/files.class.inc
```diff
@@ -368,7 +373,7 @@ class Files extends \NDB_Page
         $fileSize     = $uploadedFile->getSize();
         $fileName     = $uploadedFile->getClientFileName();
         $fileType     = pathinfo($fileName, PATHINFO_EXTENSION);
-        $uploadPath   = "$base/modules/document_repository/user_uploads/$name/";
+        $uploadPath   = $path.$name."/";
```
johnsaigle: I think this should be using the new Utility function too

PR #5816, opened 2019-12-02
[Survey Accounts] Make integration test more robust

johnsaigle:
## Brief summary of changes

In another PR I encountered an issue with these integration tests. 

Basically it's required to submit an instrument and a visit label in this form but the test was not actually checking for this.

* First discovered in #5796. This PR blocks that one.

### DIFF
--- a/modules/survey_accounts/test/survey_accountsTest.php
+++ b/modules/survey_accounts/test/survey_accountsTest.php
@@ -215,7 +215,34 @@ class Survey_AccountsTestIntegrationTest extends LorisIntegrationTest
      */
     function testSurveyAccountsAddSurvey()
     {
-        //Visit does not exist for given candidate.
+        $visitLabel        = 'V1';
+        $invalidVisitLabel = '';
+        $instrument        = 'bmi';
+
+        // Ensure that an instrument must be supplied.
+        $this->safeFindElement(
+            WebDriverBy::Name("CandID")
+        )->sendKeys("999999");
+        $this->safeFindElement(
+            WebDriverBy::Name("PSCID")
+        )->sendKeys("8889");
+        $this->safeFindElement(
+            WebDriverBy::Name("VL")
+        )->sendKeys($visitLabel);
+        $this->safeFindElement(
+            WebDriverBy::Name("fire_away")
+        )->click();
+        $bodyText =  $this->safeFindElement(
+            WebDriverBy::cssSelector(".error")
+        )->getText();
+        $this->assertContains(
+            "Visit $invalidVisitLabel does not exist for given candidate",
+            $bodyText
+        );
+
+        // Ensure that an error is thrown when an invalid visit label is
+        // supplied. This tests that the visit label is valid as well as
+        // not-empty.
         $this->safeGet($this->url . "/survey_accounts/");
         $btn = self::$add;
         $this->safeClick(WebDriverBy::cssSelector($btn));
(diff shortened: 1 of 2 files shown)


johnsaigle:
I have a suspicion that my issues with this PR may be addressed by https://github.com/aces/Loris/pull/6644

AlexandraLivadas:
@johnsaigle When I checked the branch out and ran integration tests, I got this error:
`ConfigurationException: Config setting url does not exist in database`

This was resolved in #6563. If you rebase with master, that should be resolved.

modules/survey_accounts/test/survey_accountsTest.php
```diff
@@ -208,14 +208,77 @@ function testSurveyAccountsWithoutPermission()
...
+            WebDriverBy::Name("Test_name")
+        )->sendKeys($instrument);
+        $this->safeFindElement(
+            WebDriverBy::Name("fire_away")
+        )->click();
+        $bodyText =  $this->safeFindElement(
+            WebDriverBy::cssSelector(".error")
+        )->getText();
+        $this->assertContains(
+            "Visit $invalidVisitLabel does not exist for given candidate",
```
AlexandraLivadas, line 275: Changing this to:`"Visit $visitLabel does not exist for given candidate`might fix the integration test that is failing.

johnsaigle, line 275:
```suggestion
            "Visit $visitLabel does not exist for given candidate",
```

johnsaigle: Nice catch. I think this was leftover from an earlier thing I tried.

modules/survey_accounts/test/survey_accountsTest.php
```diff
@@ -208,14 +208,77 @@ function testSurveyAccountsWithoutPermission()
...
+            WebDriverBy::Name("Test_name")
+        )->sendKeys($instrument);
+        $this->safeFindElement(
+            WebDriverBy::Name("fire_away")
+        )->click();
+        $bodyText =  $this->safeFindElement(
+            WebDriverBy::cssSelector(".error")
+        )->getText();
+        $this->assertContains(
+            "Visit $visitLabel does not exist for given candidate",
```
AlexandraLivadas, line 275: Travis is failing at this assertion. Here is the error: 
`Failed asserting that 'Visit V2 does not exist for given candidate' contains "Visit V1 does not exist for given candidate".`

I'm unsure why `V2` is being returned since at line 264 you can see that the variable submitted is `$visitLabel`, which is equal to `V1`. 

I can look into this a bit more this afternoon @johnsaigle

johnsaigle: That would be awesome

modules/survey_accounts/test/survey_accountsTest.php
```diff
@@ -208,14 +208,77 @@ function testSurveyAccountsWithoutPermission()
...
+        );
+        $this->safeFindElement(
+            WebDriverBy::Name("CandID")
+        )->sendKeys("999999");
+        $this->safeFindElement(
+            WebDriverBy::Name("PSCID")
+        )->sendKeys("8889");
+        $this->safeFindElement(
+            WebDriverBy::Name("VL")
+        )->sendKeys($visitLabel);
```
AlexandraLivadas, line 264: Tested on this branch on my own VM:
```suggestion
        $visitField = $this->safeFindElement(
            WebDriverBy::Name("VL"));
        $visitOption = new WebDriverSelect($visitField);
        $visitOption->selectByValue($visitLabel);
```
When you add a survey, the visit label element is a drop-down menu, so I think the original way that this code is written wasn't properly selecting a drop-down menu option. Changing it to this solved it on my VM. I suggest making this change throughout this test whenever you have to set the visit label.

**Note:** If you implement this change, you need to add this line at the top of the file: 
`use Facebook\WebDriver\WebDriverSelect;
`

modules/survey_accounts/test/survey_accountsTest.php
```diff
@@ -225,14 +288,52 @@ function testSurveyAccountsAddSurvey()
...
             WebDriverBy::Name("fire_away")
         )->click();
         $bodyText =  $this->safeFindElement(
             WebDriverBy::cssSelector(".error")
         )->getText();
         $this->assertContains(
-            "Visit V1 does not exist for given candidate",
+            "You must specify a valid Visit Label.",
             $bodyText
         );
```
AlexandraLivadas, line 338: This is the exact same code as 255-277, so this is also causing an integration test failure. I suggest either removing this section or changing the error message back to `"Visit V1 does not exist for given candidate"`

modules/survey_accounts/test/survey_accountsTest.php
```diff
@@ -225,14 +288,52 @@ function testSurveyAccountsAddSurvey()
         $this->safeFindElement(
             WebDriverBy::Name("PSCID")
         )->sendKeys("8889");
+        $this->safeFindElement(
+            WebDriverBy::Name("VL")
+        )->sendKeys($invalidVisitLabel);
```
AlexandraLivadas, line 293: This section of the test also causes an integration test error. Since the visit label selection is a drop-down menu with the options `V1-V6`, trying to pass an empty string here causes this error when the test is run:
`Facebook\WebDriver\Exception\NoSuchElementException: Cannot locate option with value: ' '`

I suggest that this section of the test be removed.

ridz1208:
@AlexandraLivadas we have assigned you to maintain this PR (fix it when needed and incorporate reviews). If you are having trouble working directly on John's branch feel free to close this one and issue a new PR from your fork

@kongtiaowang please review and test when Alex is done fixing

modules/survey_accounts/test/survey_accountsTest.php
```diff
@@ -183,7 +184,7 @@ public function tearDown()
      */
     function testSurveyAccountsDoespageLoad()
     {
-        $this->setupPermissions(["survey_accounts_view"]);
+        $this->setupPermissions(["user_accounts"]);
```
kongtiaowang, line 187: survey_accounts_view not user_accounts

driusan:
@AlexandraLivadas if your review is no longer valid (I assume it's not since you added commits..) can you dismiss it?

PR #5820, opened 2019-12-03
[DEV] Gitignore package lock

xlecours:
## Brief summary of changes
Removing a daily annoyance by not tracking npm generated file `package-lock.json`
Travis is using `make` and ignore the package-lock anyways.


johnsaigle:
Transitory failure caused by #5719

maltheism:
I always do `git restore package-lock.json` but we could just commit the file like @johnsaigle commented.

driusan:
I believe a lot of the changes to this file come from different versions or configurations of `npm` changing between `https` and `http`. Is there a setting in the non-lock we could use to explicitly tell it to use `https` and avoid the false positives?

maltheism:
@driusan 

> Is there a setting in the non-lock we could use to explicitly tell it to use https and avoid the false positives?

not that I'm aware of but I was reading [here](https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/43) and it seems to be a problem with the registry of npm.

Apparently there is a way to have npm replace all the http to https in package-lock.json if they occur. Such as adding the command  `"postshrinkwrap": "replace --silent 'http://' 'https://' ./package-lock.json",` to package.json. Credits to found [here](https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/44). I'm sure if we made npm run compile execute that command after it's done.. we wouldn't run into the package-lock.json issue nearly as often.

PapillonMcGill:
@johnsaigle This might be a goal but it is technically very hard to achieve. Even using make, I quite very often have to revert that single files in order to change branches or pull latest commits even if I use make. I second to remove this files from tracking. 
We could add it back later if there is a way to solve this problem.

johnsaigle:
> if there is a way to solve this problem

I think we should try the fix I suggested (tightening up the dependencies) before giving up.

I've also done a fix similar to the one @maltheism suggested when I was moving from npm 6 (Ubuntu's default) to npm 8 (our required version) and it did fix the http/https issue.

PR #5826, opened 2019-12-04
[API] Add README for api module

driusan:
The API module is missing a README, because it didn't exist
as a module when it lived under htdocs and none was written
when it was migrated to a "real" LORIS module. This takes
a first pass at writing one to officially define its purpose.

### DIFF
--- a/modules/api/README.md
+++ b/modules/api/README.md
@@ -1,0 +1,38 @@
+# API
+
+## Purpose
+
+The api module is intended to provide a stable, versioned REST interface
+for interacting with the core LORIS functionality without requiring backend
+SQL access.
+
+## Intended Users
+
+The module has 2 intended users:
+1. Users (or creators) of LORIS data that are external to the LORIS core
+   software that must extract or import data from LORIS.
+2. Internal LORIS modules
+
+## Scope
+
+The API only provides information about concepts that are cross-module
+or data that is of general concern to users. 
+
+NOT in scope:
+
+- Data and functionality that is specific to a single module, and not of general
+  nature.
+- A graphical frontend
+
+## Permissions
+
+????
+
+## Configurations
+
+useEDC - This configuration setting determines whether the EDC is returned
+  or required for creating/retrieving candidates with the API.
+
+## Interactions with LORIS
+
+???
--- a/modules/api/README.md
+++ b/modules/api/README.md
@@ -11,7 +11,7 @@ SQL access.
 The module has 2 intended users:
 1. Users (or creators) of LORIS data that are external to the LORIS core
    software that must extract or import data from LORIS.
-2. Internal LORIS modules
+2. Internal LORIS modules.
 
 ## Scope
 
@@ -22,7 +22,7 @@ NOT in scope:
 
 - Data and functionality that is specific to a single module, and not of general
   nature.
-- A graphical frontend
+- A graphical frontend for interacting with the API.
 
 ## Permissions
 
--- a/modules/api/README.md
+++ b/modules/api/README.md
@@ -26,9 +26,17 @@ NOT in scope:
 
 ## Permissions
 
+The `api` module is a public module so that the `login` endpoint can be accessed,
+but otherwise all requests require a token (retrieved from the `login` endpoint)
+or the user accessing the API to be logged in to LORIS.
+
+The following permissions affect accessing data from the API:
 ????
 
 ## Configurations
+
+The following configuration settings affect the format of data returned from
+the API:
 
 useEDC - This configuration setting determines whether the EDC is returned
   or required for creating/retrieving candidates with the API.
--- a/modules/api/README.md
+++ b/modules/api/README.md
@@ -4,7 +4,7 @@
 
 The api module is intended to provide a stable, versioned REST interface
 for interacting with the core LORIS functionality without requiring backend
-SQL access.
+database access.
 
 ## Intended Users
 
--- a/modules/api/README.md
+++ b/modules/api/README.md
@@ -9,7 +9,7 @@ database access.
 ## Intended Users
 
 The module has 2 intended use cases:
-1. Users (or creators) of LORIS data that are external to the LORIS core
+1. Interactions with LORIS data coming from outside the LORIS core
    software that must extract or import data from LORIS.
 2. Internal LORIS modules which retrieve data from the API.
 
--- a/modules/api/README.md
+++ b/modules/api/README.md
@@ -8,10 +8,10 @@ SQL access.
 
 ## Intended Users
 
-The module has 2 intended users:
+The module has 2 intended use cases:
 1. Users (or creators) of LORIS data that are external to the LORIS core
    software that must extract or import data from LORIS.
-2. Internal LORIS modules.
+2. Internal LORIS modules which retrieve data from the API.
 
 ## Scope
 
@@ -43,4 +43,8 @@ useEDC - This configuration setting determines whether the EDC is returned
 
 ## Interactions with LORIS
 
-???
+- A token retrieved from the login endpoint bypasses the normal cookie based
+  authentication of LORIS when provided in as "Authorization" bearer token. See
+  API documentation for details.
+
+- ??? (describe other interactions here)
--- a/modules/api/README.md
+++ b/modules/api/README.md
@@ -10,7 +10,7 @@ database access.
 
 The module has 2 intended use cases:
 1. Interactions with LORIS data coming from outside the LORIS core
-   software that must extract or import data from LORIS.
+   software that extract or import data from LORIS.
 2. Internal LORIS modules which retrieve data from the API.
 
 ## Scope
--- a/modules/api/README.md
+++ b/modules/api/README.md
@@ -31,7 +31,8 @@ but otherwise all requests require a token (retrieved from the `login` endpoint)
 or the user accessing the API to be logged in to LORIS.
 
 The following permissions affect accessing data from the API:
-????
+- The `data_entry` permission is required to access the candidates portion of
+  the API.
 
 ## Configurations
 
@@ -46,5 +47,3 @@ useEDC - This configuration setting determines whether the EDC is returned
 - A token retrieved from the login endpoint bypasses the normal cookie based
   authentication of LORIS when provided in as "Authorization" bearer token. See
   API documentation for details.
-
-- ??? (describe other interactions here)
--- a/modules/api/README.md
+++ b/modules/api/README.md
@@ -2,7 +2,7 @@
 
 ## Purpose
 
-The api module is intended to provide a stable, versioned REST interface
+The API module is intended to provide a stable, versioned REST interface
 for interacting with the core LORIS functionality without requiring backend
 database access.
 


modules/api/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+NOT in scope:
+
+- Data and functionality that is specific to a single module, and not of general
+  nature.
+- A graphical frontend
+
+## Permissions
+
+????
```
driusan: I could use some help with this part from anyone who knows the current state of it.

modules/api/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Configurations
+
+useEDC - This configuration setting determines whether the EDC is returned
+  or required for creating/retrieving candidates with the API.
+
+## Interactions with LORIS
+
+???
```
driusan: Same here

modules/api/README.md
maltheism: @driusan maybe it should be stated that a login session needs to be  retrieved before accessing other API requests? @xlecours

modules/api/README.md
maltheism: @driusan we could list the current API interactions with LORIS.
Such as candidates, login, projects, and update this when new API gets added. thoughts? @xlecours

modules/api/README.md
```diff
@@ -0,0 +1,38 @@
...
+The api module is intended to provide a stable, versioned REST interface
+for interacting with the core LORIS functionality without requiring backend
+SQL access.
+
+## Intended Users
+
+The module has 2 intended users:
+1. Users (or creators) of LORIS data that are external to the LORIS core
+   software that must extract or import data from LORIS.
+2. Internal LORIS modules
```
maltheism: Point two is missing period.

modules/api/README.md
```diff
@@ -0,0 +1,38 @@
...
+## Scope
+
+The API only provides information about concepts that are cross-module
+or data that is of general concern to users. 
+
+NOT in scope:
+
+- Data and functionality that is specific to a single module, and not of general
+  nature.
+- A graphical frontend
```
maltheism: Missing period if consistency.

modules/api/README.md
driusan: Yes, that's what I was thinking.. I just don't know the current extent of which modules affect where the data is coming from.

modules/api/README.md
christinerogers: This should be under Interactions with Loris. or can you clarify why modules are users?

modules/api/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Permissions
+
+????
+
+## Configurations
+
+useEDC - This configuration setting determines whether the EDC is returned
+  or required for creating/retrieving candidates with the API.
+
```
christinerogers: Projects... this is under the Config module, so should get mentioned, non?

modules/api/README.md
driusan: I was thinking I needed help with permissions which affect what data can be accessed, but that's a good point, I'll add that too.

modules/api/README.md
christinerogers: i think modules are not "users" but should go here, as a general mention.

modules/api/README.md
driusan: They don't affect the API that I could tell from grepping the code for `getSetting`, the `useEDC` config variable changes the format of what's returned or required by certain endpoints (as described in the API documentation, which should probably be moved under this module in a separate pull request.)

modules/api/README.md
driusan: The API module is different than other modules in that it's intended to be an API, as such I think modules are (or, at least, should be) users of the API.

modules/api/README.md
driusan: I've added a sentence to clarify a little

modules/api/README.md
```diff
@@ -0,0 +1,50 @@
...
+- A graphical frontend for interacting with the API.
+
+## Permissions
+
+The `api` module is a public module so that the `login` endpoint can be accessed,
+but otherwise all requests require a token (retrieved from the `login` endpoint)
+or the user accessing the API to be logged in to LORIS.
+
+The following permissions affect accessing data from the API:
+????
```
cmadjar: Imaging permissions affect how you can access the imaging data via the API (will be the same for electrophysiology data). Don't know about the rest. I guess there is site and project permission? @xlecours ?

modules/api/README.md
```diff
@@ -0,0 +1,50 @@
+# API
+
+## Purpose
+
+The api module is intended to provide a stable, versioned REST interface
```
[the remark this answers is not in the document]
driusan: I don't understand.. it's already written api?

modules/api/README.md
```diff
@@ -0,0 +1,50 @@
...
+for interacting with the core LORIS functionality without requiring backend
+SQL access.
+
+## Intended Users
+
+The module has 2 intended use cases:
+1. Users (or creators) of LORIS data that are external to the LORIS core
+   software that must extract or import data from LORIS.
+2. Internal LORIS modules which retrieve data from the API.
```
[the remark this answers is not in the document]
driusan: I think the imaging pipeline falls under the first use case

modules/api/README.md
christinerogers: for consistency, as @cmadjar pointed out
```suggestion
The API module is intended to provide a stable, versioned REST interface
```

modules/api/README.md
```diff
@@ -0,0 +1,50 @@
+# API
+
+## Purpose
+
+The api module is intended to provide a stable, versioned REST interface
+for interacting with the core LORIS functionality without requiring backend
+SQL access.
```
christinerogers:
```suggestion
database access.
```

modules/api/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Intended Users
+
+The module has 2 intended use cases:
+1. Users (or creators) of LORIS data that are external to the LORIS core
```
christinerogers: To capture the abstraction a bit better, I'm suggesting: 
```suggestion
1. Interactions with LORIS data coming from outside the LORIS core
```

modules/api/README.md
```diff
@@ -0,0 +1,50 @@
...
+   software that must extract or import data from LORIS.
```
christinerogers:
```suggestion
   software that extract or import data from LORIS.
```

modules/api/README.md
```diff
@@ -0,0 +1,50 @@
...
+2. Internal LORIS modules which retrieve data from the API.
+
+## Scope
+
+The API only provides information about concepts that are cross-module
```
christinerogers: concepts?  how about instead: entity/object?

modules/api/README.md
```diff
@@ -0,0 +1,50 @@
...
+or data that is of general concern to users.
```
christinerogers: what does `general concern` actually mean-- that the super-technical metadata is not available?  
It would be nice to state this more clearly as "all/most data visible in the web front-end is accessible via API"

modules/api/README.md
```diff
@@ -0,0 +1,50 @@
...
+2. Internal LORIS modules which retrieve data from the API.
+
+## Scope
+
+The API only provides information about concepts that are cross-module
+or data that is of general concern to users. 
+
+NOT in scope:
+
+- Data and functionality that is specific to a single module, and not of general
```
christinerogers: Does this include QC?  Dicoms?

modules/api/README.md
```diff
@@ -0,0 +1,50 @@
...
+  nature.
+- A graphical frontend for interacting with the API.
+
+## Permissions
+
+The `api` module is a public module so that the `login` endpoint can be accessed,
+but otherwise all requests require a token (retrieved from the `login` endpoint)
+or the user accessing the API to be logged in to LORIS.
```
christinerogers:
```suggestion
for the user accessing the API to be logged in to LORIS.
```

modules/api/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+The following permissions affect accessing data from the API:
+????
+
+## Configurations
+
+The following configuration settings affect the format of data returned from
```
christinerogers:
```suggestion
The following Configuration module settings affect the format of data returned from
```

modules/api/README.md
```diff
@@ -0,0 +1,50 @@
...
+the API:
+
+useEDC - This configuration setting determines whether the EDC is returned
```
christinerogers:
```suggestion
* useEDC : determines whether the EDC is returned (Estimated Date of Confinement, or birth Due Date)
```

modules/api/README.md
```diff
@@ -0,0 +1,50 @@
...
+  or required for creating/retrieving candidates with the API.
+
+## Interactions with LORIS
+
+- A token retrieved from the login endpoint bypasses the normal cookie based
```
christinerogers:
```suggestion
- A token retrieved from the login endpoint bypasses the normal cookie-based
```

modules/api/README.md
```diff
@@ -0,0 +1,50 @@
...
+  authentication of LORIS when provided in as "Authorization" bearer token. See
```
christinerogers:
```suggestion
  authentication of LORIS when provided in an "Authorization" bearer token. See
```

modules/api/README.md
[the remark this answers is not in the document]
PapillonMcGill: possibly that in line 1, `API` is used while in line 5, it is `api`. Have a constant way to write it across the files (every files in LORIS).
As it is an abbreviation, it should be `API`.

modules/api/README.md
PapillonMcGill: Yes, concepts is vague. Would also suggest `entity or object` instead of using a slash for readability.

modules/api/README.md
```diff
@@ -0,0 +1,50 @@
...
+The following permissions affect accessing data from the API:
+????
+
+## Configurations
+
+The following configuration settings affect the format of data returned from
+the API:
+
+useEDC - This configuration setting determines whether the EDC is returned
+  or required for creating/retrieving candidates with the API.
```
PapillonMcGill:
```suggestion
  or required for creating or retrieving candidates with the API.
```
as per latest document style format.

modules/api/README.md
PapillonMcGill: Could add that the user need to be logged in via `the web interface` as the concept of being `logged in` can be vague.

driusan: "or" was correct. You either need a login token from the API without being logged in *or* you need to be logged in to loris as a regular user.

modules/api/README.md
driusan: It means it's a cross module concern that multiple people would be interested in, not something added because a single module needs it. It's a little vague because it's not very well defined what should be in the API and what should be in a module at the moment

modules/api/README.md
driusan: That doesn't seem to be the case. I did a grep for "hasPermission" and the only call to the function in the API looks like the data_entry for the candidates endpoint.

modules/api/README.md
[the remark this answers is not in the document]
cmadjar: hahaha, I meant either mark `api` in code format or if not in code format, write it in capital letter however, not a blocker.

modules/api/README.md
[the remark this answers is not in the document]
cmadjar: indeed. Not a blocker for this PR to be merged :)

modules/api/README.md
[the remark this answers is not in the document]
driusan: I had written it lowercase because I thought all the README's started "The xxx module" where xxx is the lower case directory name, but I just checked a couple randomly and it seems to be very inconsistent, so I'll capitalize it since it looks better.

PR #5830, opened 2019-12-04
[Timepoint List] Add README for timepoint_list module

driusan:
Add README for timepoint_list module.

This is the last module in the GitHub project https://github.com/aces/Loris/projects/6 that does not have an open PR.

### DIFF
--- a/modules/timepoint_list/README.md
+++ b/modules/timepoint_list/README.md
@@ -1,0 +1,46 @@
+# Timepoint List
+
+## Purpose
+
+The `timepoint_list` module serves as an entry-point for data entry
+about a candidate. It also serves as a place for linking to other
+modules that provide more specific information and actions for a
+candidate's data.
+
+It is also known as the "Candidate Profile" page.
+
+## Intended Users
+
+The `timepoint_list` module is primarily intended for data entry
+staff who are accessing a candidate.
+
+## Scope
+
+The `timepoint_list` module provides a mechanism to link to
+information about candidates. 
+
+NOT in scope:
+
+The `timepoint_list` module does *not* provide any data entry
+functionality or mechanisms to configure study timepoints,
+only access timepoints of specific candidates.
+
+## Permissions
+
+The `timepoint_list` module requires the user accessing it to
+have access to at least 1  timepoint of the candidate being
+accessed. This can either occur by having the `access_all_profiles`
+permission, being a member of the same site as the candidate's
+RegistrationCenterID, or having access to a CenterID in common
+with one of the candidate's timepoints.
+
+## Configurations
+
+None
+
+## Interactions with LORIS
+
+- The `timepoint_list` links to the `candidate_parameters` module,
+  `create_timepoint` module, `imaging_browser` module, and `instrument_list`
+  module to provide more specific data entry/browsing functionality
+  about the candidate being accessed.


modules/timepoint_list/README.md
```diff
@@ -0,0 +1,48 @@
+# Timepoint List
+
+## Purpose
+
+The `timepoint_list` module serves as an entry-point for data entry
+about a candidate. It serves as a dashboard for a candidate linking
+to other modules that provide specific functionality about a candidate
```
maltheism: @driusan Missing period at end of sentence.

driusan: thanks, fixed

modules/timepoint_list/README.md
```diff
@@ -0,0 +1,48 @@
+# Timepoint List
+
+## Purpose
+
+The `timepoint_list` module serves as an entry-point for data entry
```
christinerogers: * add: It is also known as the Candidate Profile page. 
* Don't say "dashboard".  `There is only one Dashboard.`

modules/timepoint_list/README.md
```diff
@@ -0,0 +1,48 @@
...
+about a candidate. It serves as a dashboard for a candidate linking
+to other modules that provide specific functionality about a candidate.
```
christinerogers: how about... `more detailed information and actions for this candidate's data `

modules/timepoint_list/README.md
```diff
@@ -0,0 +1,48 @@
...
+## Intended Users
+
+The `timepoint_list` module is primarily intended for data entry
+staff who are accessing a candidate.
+
+## Scope
+
+The `timepoint_list` module provides a mechanism to link to
+information about candidates. 
+
```
christinerogers:
```suggestion
In particular, it provides access to longitudinally-organized behavioural data, and links to Imaging data. It is also the only access point for Candidate Parameters, where Consent is entered as well as other information relevant to the candidate across timepoints. 
```

modules/timepoint_list/README.md
```diff
@@ -0,0 +1,48 @@
...
+NOT in scope:
+
+The `timepoint_list` module does *not* provide any data entry
+functionality or mechanisms to configure study timepoints,
+only access timepoints of specific candidates.
```
christinerogers: And Imaging data.  and  Candidate Info. 

Also should be mentioned, under scope or (technical details) interactions: 
* Basic candidate profile details are visible at top, including DoB. 
* Viewing at a glance the status of each timepoint -- including stage and status. 
* Behavioural Feedback 
* Bvl Exclusion -- > this should be mentioned under Interactions with Loris

modules/timepoint_list/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Permissions
+
+The `timepoint_list` module requires the user accessing it to
+have access to at least 1  timepoint of the candidate being
+accessed. This can either occur by having the `access_all_profiles`
+permission, being a member of the same site as the candidate's
+RegistrationCenterID, or having access to a CenterID in common
+with one of the candidate's timepoints.
+
```
christinerogers: useEDC.  -> will show in top bar of basic candidate profile

modules/timepoint_list/README.md
driusan: I don't like either the word "specific" or "detailed" but I left it as "specific" because it's slightly more accurate and took the rest of your suggestion.

modules/timepoint_list/README.md
```diff
@@ -0,0 +1,48 @@
...
+## Configurations
+
+None
+
+#### Database Configurations
+
+None
+
```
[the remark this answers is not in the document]
driusan: no, I just missed it when I copied the template from the imaging_browser README.. good catch.

modules/timepoint_list/README.md
christinerogers: @driusan are you planning to address this?

modules/timepoint_list/README.md
christinerogers: sure, works for me.

modules/timepoint_list/README.md
driusan: I addressed the first part in the paragraph below, and I haven't decide on the second. The sentence is using "dashboard" in the English language definition of the word and doesn't claim to be the dashboard module. There's no reason that there can't be multiple different kinds of dashboard to provide overviews of different features/entities. Calling it a "dashboard" here gives more flexibility to house features that aren't related to listing timepoints (which it already does.)

christinerogers: my point is -- it confuses the terminology we're using in Loris, not that the word is incorrect.

(yes, we could call many things dashboards, like the Stats module or the bvl/Img QC modules, but why confuse it?)

This is the Candidate Profile page as well as the Timepoint List, there's no need to add "dashboard" which is an Actual Thing elsewhere. -- let's wait til the candidate module (replacing this) gets deployed before we call this a dashboard.  
Presumably there will be a lot more dashboard-like info visible in the new page, like: how many outstanding forms? what's the consent status?  how many QCd images?  any biospecimens? etc...

modules/timepoint_list/README.md
driusan: I don't see why it would go into either.

modules/timepoint_list/README.md
driusan: I don't see why people would think those things fall under the scope of this module. The timepoint related ones are mentioned because of the module name.

modules/timepoint_list/README.md
driusan: I checked the code and there's no reference to useEDC in the module.

PapillonMcGill:
@christinerogers ping

Issue #5380, opened 2019-10-24
[new_profile] If user assigned to 1 site, unhide site field and auto-populate default site


zaliqarosli:
currently, if the user is assigned to only 1 site, new profile will not show the site field. on the php side, the value of site is assigned by default the user's first and only site.

ideally, the site field will always show and just auto-populate the one and only available site so users don't have to select it themselves.

driusan:
Why?

zaliqarosli:
@driusan because it looks like there's a bug when the site field is missing!

cmadjar:
@ridz1208 could you look into this one please? Thank you!

ridz1208:
@cmadjar this is by design right now, it's how Mouna intended it to be. we would have to decide if thats the way people prefer it before changing the functionality...

I agree it looks like a bug but it isn't one unless we all agree on it

cmadjar:
@ridz1208 sounds good. Can you add this to the agenda of the LORIS meeting?

ridz1208:
@cmadjar done

zaliqarosli:
@ridz1208 do you remember the decision made post discussion on this? i can go ahead and make the changes if it is good to be implemented

ridz1208:
@zaliqarosli the decision was to go ahead with this !

zaliqarosli:
> Is there any way to do this automatically for ANY dropdown that only has one option ? (maybe we can discuss that)

@ridz1208 this PR does what you mentioned above

PR #5831, opened 2019-12-04
[create_timepoint] upgraded module to react

maltheism:
## Brief summary of changes

1. Checkout PR
2. Go through the Create Timepoint process.

* Redmine #16008


modules/create_timepoint/php/create_timepoint.class.inc
```diff
@@ -48,273 +35,28 @@ class Create_Timepoint extends \NDB_Form
     function _hasAccess(\User $user) : bool
     {
         $candidate = \Candidate::singleton(new CandID($this->identifier));
-
         // check user permissions
         return (
             $user->hasPermission('data_entry') &&
             (in_array(
```
johnsaigle: Instead you can use

`$user->hasCenter($candidate->getCenterID())`

modules/create_timepoint/php/endpoints/timepoint.class.inc
```diff
@@ -0,0 +1,335 @@
...
+            //TODO: change this to array_key_first() when support is only PHP 7.3+
+            $project = \Project::singleton(array_pop($projects));
+        } else if (count($projects) > 1) {
+            $project_id = intval($values['psc']);
+            $project    = \Project::singleton($projects[$project_id]);
+        }
+        assert(isset($site) && $site !== null);
+        $errors   = $values['errors'];
+        $response = array();
+        if (!$errors && $site) {
```
johnsaigle: Can you make the check on `$site` more precise? I don't know whether a Site object in PHP has a truthy value. Also if a new object does have a True value then this part of the if statement will always evaluate to true so it's not adding anything to have it.

modules/create_timepoint/php/endpoints/timepoint.class.inc
```diff
@@ -0,0 +1,335 @@
...
+            ))
+        ) {
+            $values['error'] = 'Forbidden';
+            return $values;
+        }
+        return $values;
+    }
+    /**
+     * Returns true if the user has permission to access
+     * the Genomic Browser module
```
johnsaigle:
```suggestion
     * the Create Timepoint module
```

modules/create_timepoint/php/endpoints/timepoint.class.inc
```diff
@@ -0,0 +1,335 @@
...
+        $user      = $request->getAttribute('user');
+        $candidate = \Candidate::singleton(
+            new CandID($values['candID'])
+        );
+        if (!$user->hasPermission('data_entry')
+            && !(in_array(
+                intval($candidate->getData('RegistrationCenterID')),
+                $user->getData('CenterIDs'),
+                true
+            ))
```
johnsaigle:
```php
&& !$user->hasCenter($candidate->getCenterID())
```

modules/create_timepoint/php/endpoints/timepoint.class.inc
maltheism: @johnsaigle I'm not quite sure whats best to do here. The code is originally from the create_timepoint.class.inc _process(values) function on master. My understanding is the assert doesn't make sense here because the site object should always either be created successfully or an error would be thrown in the singleton function of the class. Should I remove the assert?

johnsaigle: Yes I think the assert can be removed and this line can be changed to check only for `$errors`

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -0,0 +1,335 @@
...
+        options={this.state.form.options.visit}
+        onUserInput={this.setForm}
+        emptyOption={true}
+        disabled={false}
+        required={this.state.form.options.required.visit}
+      />
+    ) : null;
+
+    return (
+      <div>
```
HenriRabalais: `<div>`'s are unnecessary here

kongtiaowang:
testCreateTimepoint function should be skipped in this PR.
This new feature, after creating a timepoint will redirect to the candidate profile page instead of the "New time point successfully registered" page. I will rewrite the new test method to match this new change.

maltheism:
Thanks @HenriRabalais and @kongtiaowang for the help. 🙂

maltheism:
Latest commit added the language feature from @jesscall .

modules/create_timepoint/php/create_timepoint.class.inc
```diff
@@ -17,13 +17,12 @@ use \LORIS\StudyEntities\Candidate\CandID;
...
  * @package  Create_Timepoint
- * @author   Loris Team <loris.mni@bic.mni.mcgill.ca>
+ * @author   Alizée Wickenheiser <alizee.wickenheiser@mcin.ca>
  * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
  * @link     https://github.com/aces/Loris-Trunk
  */
-
 class Create_Timepoint extends \NDB_Form
 {
     /**
```
zaliqarosli, line 28:
```suggestion
    public $skipTemplate  = true;
    /**
```

zaliqarosli: you'll can remove the templates directory from this module. in order to do that without errors, you'll need to add the line above ^

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -0,0 +1,358 @@
...
+          class={'form-group col-sm-12'}
+          onSubmit={this.handleSubmit}
+        >
+          <StaticElement
+            label={'DCCID'}
+            text={this.state.data.dccid}
+          />
+          {subproject}
+          {psc}
+          {visit}
```
zaliqarosli, line 329: on master (screenshot below), the create timepoint form has a Project field as well that doesn't appear here. not sure if this is intentional but the PR description made no mention of removing it

Screen Shot 2020-07-07 at 6 31 46 PM

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -0,0 +1,358 @@
...
+    fetch(
+      url, {
+        method: 'POST',
+        credentials: 'same-origin',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        body: send,
+      }).then((response) => {
+      return response.json();
```
zaliqarosli, line 207: i think it would be nice here to add a success swal if (response.ok) or (response.status === 200). if not a swal, at least something to replace the current confirmation 'New time point successfully registered':

Screen Shot 2020-07-07 at 6 31 56 PM

i also think it would be better to check if response.ok first, then else, call response.json() and check if there errors. this way, you are checking against the response actually being ok, rather than whether the response returns an error in the json.

```
    .then((response) => {
      if (response.ok || response.status === 200) {
        swal.fire('Success!', 'New time point successfully registered.', 'success').then((result) => {
          if (result.value) {
              // whatever you need to do here i.e. refresh / replace
          }
        });
      } else {
        response.json().then((data) => {
          if (data.errors) {
             // do what you want to do here
          }
        });
      }
    })
    .catch((error) => {
      console.error(error);
    });
```

modules/create_timepoint/php/timepoint.class.inc
```diff
@@ -0,0 +1,344 @@
...
+                $site,
+                $project,
+                intval($values['languages']) ?? null,
+            );
+            $response['status'] = 'success';
+        } else {
+            $response['status'] = 'error';
+            $response['errors'] = $errors;
+        }
+        return new \LORIS\Http\Response\JsonResponse(
```
zaliqarosli, line 218: might be good to return https://github.com/aces/Loris/blob/master/src/Http/Response/JSON/OK.php if status is success or the 201 Created.php so the response is more robust

modules/create_timepoint/jsx/createTimepointIndex.js
maltheism: thanks @zaliqarosli, I'm not sure what happened to it. I'll fix.

modules/create_timepoint/jsx/createTimepointIndex.js
maltheism: @zaliqarosli great idea. I'll add it.

maltheism: @zaliqarosli I'm assuming but I think if the `return response.json()` were to throw an error the following code would run:
```
.catch((error) => {
        console.error('Error! ' + error);
      });
```
I can change it in any case.

modules/create_timepoint/php/timepoint.class.inc
maltheism: @zaliqarosli I'm a little confused what you mean by that? The browser should be receiving 201 if it successfully returns the json response.

zaliqarosli: ahh then you shouldnt be returning a 201 json response with 'errors'. if there are errors, it shouldn't be a 201

maltheism: @zaliqarosli maybe I should rename errors to something else? The old form logic would show a custom error message depending on specific submission mistakes. I'm actually not sure if they can ever be shown anymore with the frontend change of select. It has been awhile since I've written the code. I'll see if I can just remove it or if it makes sense to have it.

maltheism:
@zaliqarosli Latest commit adds the project select that was missing. It's kind of odd because project is never used for making a timepoint but just to see if the user can make a timepoint.

modules/create_timepoint/php/timepoint.class.inc
maltheism: @zaliqarosli i'll rename errors to validation

maltheism: actually I'm not happy with that name

maltheism: maybe "formValidation" is better

maltheism: @zaliqarosli I guess the confusion came from how I had the functionality of form messages be assigned to errors. I renamed to message when formMessage exist. The messages are not errors and instead are these examples:
You must be affiliated with a site to create a timepoint.
Site must be selected from the available dropdown.
A language must be selected.

maltheism:
@zaliqarosli I still haven't added the swal confirmation. Please wait until I can get that added before reviewing.

maltheism:
@zaliqarosli latest commits added requested swal.

modules/create_timepoint/jsx/createTimepointIndex.js
zaliqarosli: hey @maltheism , sorry i'm only just seeing this comment now. you would still access the custom error messages using `data.errors` in the else block i.e.

```
        response.json().then((data) => {
          if (data.errors) {
             // access custom errors messages here with data.errors
             // do what you want to do here
          }
        });
```

i'm not sure what you mean by 

> That would be the same as response.ok or response.status being an error

response.ok and response.status === 200 is by definition a response without any errors. the catch error block you pasted above would be called i think only if there was a 500 or a bug with the component. as long as you don't have any JS or PHP errors, the `.then((response) => {` block would be entered and THEN, you have to see if the response is OK or maybe a 409 or whatever.

response.json() would never throw an response error because what response.json() is is just some string containing the custom error message that you then use to display. but receiving the response.json() error string isn't a response error itself, it actually is a 200 OK "I have received this string"

i found this little bit from https://reactjs.org/docs/faq-ajax.html helpful. let me know if you want to discuss further on slack!

Screen Shot 2020-07-20 at 11 34 25 AM

modules/create_timepoint/php/timepoint.class.inc
zaliqarosli: hey @maltheism , seems like from what i gathered on the internet, that you could either return  a 200 with json errors, or return a 406 or 422. seeing as we don't have those response classes yet, i think it's okay for us to return a 200/201, and I think your fetch api implementation in the JS class is fine!

xlecours:
@maltheism Just a question: why don't you use a PUT request to  `/candidates/$candid/$visit_label` ?

maltheism:
@zaliqarosli see latest commits and thanks so much for your help!
@xlecours I didn't see your latest comment until now. I'll work on replacing the POST with a PUT.

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -0,0 +1,390 @@
...
+            .then(() => {
+              window.location.replace(
+                `${this.props.baseURL}/${this.state.url.params.candID}`
+              );
+            });
+        }
+      })
+      .catch((error) => {
+        console.error('Error! ' + error);
+      });
```
zaliqarosli, line 240:
```suggestion
      }).then((response) => {
        if (response.ok) {
          swal.fire('Success!', 'Time Point created.', 'success')
            .then(() => {
              window.location.replace(
                `${this.props.baseURL}/${this.state.url.params.candID}`
              );
            });
        } else {
          response.json().then((data) => {
            if (data.error) {
              // display conflicts on form.
              this.setState({messages: JSON.parse(data.error)});
            }
          });
        }
      })
      .catch((error) => {
        console.error('Error! ' + error);
      });
```

modules/create_timepoint/php/timepoint.class.inc
```diff
@@ -0,0 +1,369 @@
...
+            \Candidate::singleton(new CandID($values['candID'])),
+            intval($values['subproject']),
+            $values['visit'] ?? null,
+            $site,
+            $project,
+            intval($values['languages']) ?? null
+        );
+        return new \LORIS\Http\Response\JsonResponse(
+            $response
+        );
```
zaliqarosli, line 230: I don't think the $response exists anymore which your latest refactoring

```suggestion
        return new \LORIS\Http\Response\JsonResponse\Created();
```

modules/create_timepoint/php/timepoint.class.inc
```diff
@@ -0,0 +1,369 @@
...
+                    return new \LORIS\Http\Response\JSON\Forbidden(
+                        'Forbidden'
+                    );
+                default:
+                    break;
+            }
+        }
+
+        // response to send to frontend.
+        $response = [];
```
zaliqarosli, line 190:
```suggestion

```

maltheism:
Thanks @zaliqarosli!

ridz1208:
@zaliqarosli can you re-review

xlecours:
> @xlecours I didn't see your latest comment until now. I'll work on replacing the POST with a PUT

What I meant was that the code in `modules/create_timepoint/php/timepoint.class.inc` should be identical to the one in `modules/api/php/endpoints/candidate/visit/visit.class.inc`

maltheism:
Hi @xlecours, I'm unsure if that's a good idea. I've added the edge cases while rebasing this PR with all the constant changes from others' PRs while reactifying this module. I don't see the language code in the `visit.class.inc`. Thoughts? I'm thinking it would be best to get this merged if I'm right that the visit.class.inc API is missing a few things that could be added. Then I could make a new PR to change the module to use it? I'm willing to do whatever you think is best.

zaliqarosli (dismissed):
The test plan will need to be updated to match the new UI/work flow.

Additionally, item 8 of the test plan fails. I am able to create a time point if i don't have the data entry permission. 

item 4:

> Ensure that if the logged in user has only 1 site affiliation, the site dropdown is not displayed on the page and the form can be submitted without any errors.

also fails. The site does appear. However, there had been discussions to show the one and only site but to auto-populate this (https://github.com/aces/Loris/issues/5380). this change can either be targeted in this PR or a separate one.

maltheism:
@zaliqarosli I asked @ridz1208 and he believes that not only data_entry people need a timepoint. Maybe just testplan needs updating? I think other changes should be left for another PR since this one is already big.

zaliqarosli:
@maltheism i think the test plan in this PR should pass for whatever code is in this PR. if there are discrepancies, i think it should be targeted here. if not, then the intended workflow on the test plan should be followed in this PR with updates targeted in another but i feel like that's more work. it shouldn't be too much to update the test plan here. you'll only need to modify a sentence or two

zaliqarosli:
@maltheism ahh i see! i feel like you need data_entry in order to create a timepoint? creating a timepoint seems like data entry to me but i'm not sure what historically its been for loris

maltheism:
Notes for after meeting discussion. It appears data_entry should be the necessary and verified the RegistrationCenterID with user CenterIDs? @zaliqarosli @ridz1208 @driusan

Latest commit solves that issue. I think the testplan doesn't need updated anymore.

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -0,0 +1,390 @@
...
+    this.fetchInitializerData()
+      .then(() => this.setState({isLoaded: true}));
+  }
+
+  /**
+   * Retrieve params from the browser URL and save it in state.
+   */
+  collectParams() {
+    const url = new URL(window.location.href);
+    const state = Object.assign({}, this.state);
```
HenriRabalais, line 95: I think we should be using `JSON.parse(JSON.stringify()) throughout`

modules/create_timepoint/test/create_timepointTest.php
```diff
@@ -73,11 +73,17 @@ function testCreateTimepointDoespageLoad()
...
+            'Skipping tests until Travis and React get along better.'
+        );
+        // $this->_createTimepoint('900000', 'Stale', 'V1');
+        // $bodyText = $this->webDriver->findElement(
+        // WebDriverBy::cssSelector("body")
+        // )->getText();
+        // $this->assertContains(
+        //    "New time point successfully registered",
+        //    $bodyText
+        // );
```
HenriRabalais, line 86: Commented out code

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -0,0 +1,390 @@
...
+          state.form.options.languages = data.languages;
+          state.form.value.languages = null;
+          state.form.display.languages = true;
+        }
+        this.setState(state);
+      }).catch((error) => {
+        const state = Object.assign({}, this.state);
+        state.messages = [{message: error}];
+        this.setState(state);
+      });
```
HenriRabalais, line 161: I'm a little confused as to what's going on here. It seems like a lot of if statements for something we have control over sending from the backend. Is this just to load options? It seems like the `value` is always set to `null` and the `display` is always set to `true`, so would there be a way to write this more efficiently?

maltheism: @HenriRabalais the if statements are necessary I believe because I think the module changes depending on configurations.

maltheism:
Hi @HenriRabalais can you review

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -0,0 +1,399 @@
...
+   */
+  async collectParams() {
+    const url = new URL(window.location.href);
+    const state = JSON.parse(JSON.stringify(this.state));
+    state.url.params = {
+      candID: url.searchParams.get('candID'),
+      identifier: url.searchParams.get('identifier'),
+    };
+    state.data.dccid = state.url.params.candID;
+    await this.setState(state);
```
HenriRabalais, line 103: I don't think `this.setState` returns a promise, so I'm not sure this works like this.

HenriRabalais: I think you would either want to do:
`return new Promise((resolve) => this.setState(state, resolve()))`

or if you want to use async/await it would be something like described here:
https://medium.com/front-end-weekly/async-await-with-react-lifecycle-methods-802e7760d802

maltheism: @HenriRabalais thanks. Please see latest commit.

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -0,0 +1,399 @@
...
+   */
+  async collectParams() {
+    const url = new URL(window.location.href);
+    const state = JSON.parse(JSON.stringify(this.state));
+    state.url.params = {
+      candID: url.searchParams.get('candID'),
+      identifier: url.searchParams.get('identifier'),
+    };
+    state.data.dccid = state.url.params.candID;
+    await new Promise((resolve) => this.setState(state, resolve));
```
HenriRabalais, line 103: I think this should work, but I'm not entirely sure this is set up the way async/await is design to be used. 

Because `collectParams` is declared as an async function, I think the idea is that when it is called in `componentDidMount()` is should be called like this:
```
      await this.collectionParams();
      this.fetchInitializerData().then(() => {
        this.setState({isLoaded: true});
      });
```

The promise at the end of `collectParams`, does not need to be awaited, but simple returned upon resolution: 
`return new Promise((resolve) => this.setState(state, resolve()));` (note `resolve()` instead of `resolve`)
This will then indicate that the call of `collectionParams` in the componentDidMount function can move forward to the next line.

Let me know if this doesn't make sense.

HenriRabalais, line 103: I think this should work... although resolve need to be called as a function, i.e. `resolve()`

HenriRabalais: I'm not sure if this is really making full use of `async/await` though, and I think it could be written like this with the same effect:
```
  collectParams() {
    const url = new URL(window.location.href);
    const state = JSON.parse(JSON.stringify(this.state));
    state.url.params = {
      candID: url.searchParams.get('candID'),
      identifier: url.searchParams.get('identifier'),
    };
    state.data.dccid = state.url.params.candID;
    return new Promise((resolve) => this.setState(state, resolve()));
```

maltheism: @HenriRabalais I'm a little confused as I don't think the `resolve()` is necessary instead of `resolve`. 

I think calling `resolve` with no parenthesis is just calling the method but not the value of that method and calling `resolve()` with parenthesis is calling the value of that method.

A few of the answers found here: https://stackoverflow.com/questions/43370176/using-async-setstate don't use the parentheses either.

I can change it to `resolve()` if you want but waiting on what you think about the above.

HenriRabalais (approved):

modules/create_timepoint/jsx/createTimepointIndex.js
HenriRabalais: Sorry, my mistake, resolve doesn't need to be called here! this.setState will take care of calling the function upon state update.

zaliqarosli:
hi @maltheism can you point me to the commit or bit of code that you changed to fix the access permission issue?

modules/create_timepoint/php/create_timepoint.class.inc
```diff
@@ -34,290 +33,27 @@ class Create_Timepoint extends \NDB_Form
...
     {
-        $candidate = \Candidate::singleton(new CandID($this->identifier));
+        $this->candidate = \Candidate::singleton(new CandID($this->identifier));
 
         // check user permissions
         return (
             $user->hasPermission('data_entry') &&
             (in_array(
-                $candidate->getData('RegistrationCenterID'),
+                $this->candidate->getData('RegistrationCenterID'),
```
zaliqarosli, line 50: i can't seem to be able to do a multi-line suggestion but I think the in_array() check from RegistrationcenterID and centerIDs can be replaced with `$candidate->isAccessibleBy($user)` as was done in this [PR](https://github.com/aces/Loris/pull/6093) for timepoint_list

zaliqarosli (dismissed):

modules/create_timepoint/php/create_timepoint.class.inc
maltheism: Thanks @zaliqarosli, I didn't know about that method. Also it can replace the permission check for loading the module as well. Cool

zaliqarosli: yeah it's pretty neat! could you still point to me to what you did to make the 'data_entry' permission actually work?

maltheism: @zaliqarosli oops, checkout the `_verifyCreatePermissions` function inside the timepoint.class.inc on line 221.

modules/create_timepoint/php/timepoint.class.inc
```diff
@@ -0,0 +1,344 @@
...
+    private function _verifyCreatePermissions(
+        ServerRequestInterface $request,
+        string $identifier
+    ) : bool {
+        $user      = $request->getAttribute('user');
+        $candidate = \Candidate::singleton(new CandID($identifier));
+        return (
+            $user->hasPermission('data_entry') &&
+            $candidate->getData('RegistrationCenterID')
+        );
```
zaliqarosli, line 230: should this maybe be the same check as the access function in create_timepoint.class.inc ? i think we said that as long as the user has access to the timepoint_list page and has data_entry permissions, they should be able to createa  time point. is this right @driusan @ridz1208?

zaliqarosli (changes requested):
if i don't have data_entry permissions and cannot access the candidate's create timepoint (directly via the URL, not by clicking the button because it's not available on the front-end), i get this page 

Screen Shot 2020-09-21 at 11 41 13 AM

when I should be getting this page

Screen Shot 2020-09-21 at 11 43 27 AM

modules/create_timepoint/php/timepoint.class.inc
maltheism: @zaliqarosli I pasted the wrong line on 229 from your fix.
It used to be:
```
return (
            $user->hasPermission('data_entry') &&
            (in_array(
                $candidate->getData('RegistrationCenterID'),
                $user->getData('CenterIDs')
            )
            )
        );
```
should be similar to what you proposed. I'll fix in latest commit.

maltheism: @zaliqarosli ready

modules/create_timepoint/php/timepoint.class.inc
```diff
@@ -0,0 +1,344 @@
...
+     * Returns true if the user has permission to access
+     * the Create Timepoint module
+     *
+     * @param \User $user The user whose access is being checked
+     *
+     * @return bool true if user has permission
+     */
+    function _hasAccess(\User $user) : bool
+    {
+        return $user->hasPermission('data_entry');
```
zaliqarosli, line 315: should this maybe also have `$this->candidate->isAccessibleBy($user)`?

maltheism: @zaliqarosli I asked @xlecours (i think last week) about how to access `candidate` in this function because the identifier needed to make the `candidate` isn't accessible until the `handle` function gets called after the _hasAccess function. I believe he said the _hasAccess function should just be checking if the module is accessible for the user and not checking whether the user can carry out what the requests are meant for doing.

zaliqarosli: okay! i think this is fine. i'll do a final test!

modules/create_timepoint/php/timepoint.class.inc
zaliqarosli: hey @maltheism sorry for suggesting this change! turns out `$this->candidate->isAccessibleBy($user)` isn't exactly the same as what was there before. before, only if you were affiliated with the candidate's registration centre were you able to create the timepoint. using the isAccessibleBy method allows you to create a timepoint if you aren't affiliated with the candidate's registration centre. could you please change it back? sorry about that

maltheism: @zaliqarosli done reverted changes.

zaliqarosli (commented):
you'll need to update test plan item #7 to suit the new swal UI/workflow

maltheism:
@zaliqarosli updated.

zaliqarosli (dismissed):
okay this should be my last changes requested! thanks for bearing with me! the 3rd and 4th points of number 4. in the test plan should be

```
   - Ensure that if the logged in user has only 1 site affiliation, that site is
   auto-selected in the site dropdown and the form can be submitted without any errors.
   - Ensure that if the logged in user has only 1 project affiliation, that project 
   is auto-selected in the project dropdown and the form can be submitted without any 
   errors.
```

this is currently the functionality for the site field. the project field however doesn't show. i think it should match what happens for the site field

maltheism:
@zaliqarosli updated. thanks!

modules/create_timepoint/php/timepoint.class.inc
```diff
@@ -0,0 +1,348 @@
...
+        $user = \User::singleton();
+        $user_list_of_projects = $user->getProjectIDs();
+        if (count($user_list_of_projects) > 1) {
+            $projectOptions = [];
+            foreach ($user_list_of_projects as $projectID) {
+                $project = \Project::getProjectFromID($projectID);
+                $projectOptions[$projectID] = $project->getName();
+            }
+            $values['project'] = $projectOptions;
+        }
```
zaliqarosli, line 137: the project isn't showing up . i think you should remove the > 1 check so that the project just gets listed, even though there's only 1, similar to 'list sites'

maltheism: @zaliqarosli updated thanks

zaliqarosli (approved):
okay i think it's ready!

modules/create_timepoint/php/timepoint.class.inc
```diff
@@ -0,0 +1,346 @@
...
+     * an incoming HTTP request.
+     *
+     * @param ServerRequestInterface $request The incoming PSR7 request.
+     *
+     * @return string The value to use for the ETag header.
+     */
+    public function ETag(ServerRequestInterface $request): string
+    {
+        if ($request->getMethod() === 'POST') {
+            return 'Etag not calculated on POST request';
```
driusan, line 330: This will set an ETag of "Etag not calculated on POST request". It should be an empty string to avoid adding an ETag.

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -0,0 +1,399 @@
...
+          }
+        });
+      }
+    }).catch((error) => {
+      console.error('Error! ' + error);
+    });
+  }
+
+  /**
+   * @return {DOMRect}
```
driusan, line 253: I don't know what a "DOMRect" is and this is missing a comment.

driusan (dismissed):

modules/create_timepoint/php/timepoint.class.inc
maltheism: thanks @driusan, see latest commits

modules/create_timepoint/jsx/createTimepointIndex.js
maltheism: thanks @driusan, see latest commits

driusan (approved):
pending travis

PR #5833, opened 2019-12-05
[Menu] Do not show Data Query Tool menu item with data_dict_view permission

driusan:
The default schema incorrectly added a LorisMenuPermissions item for the
Data Query Tool if the user had the data_dict_view permission in the
default schema. This removes that line, as the correct permission
(according to the code) is the dataquery_view permission (which is
separately already included in the default schema. See line 209.)


christinerogers:
why not include a patch? too tricky to remove?

ridz1208:
This could be sent out to 21 as well. I have a different PR for 21 that will also need a patch
https://github.com/aces/Loris/pull/5838

ridz1208:
@driusan could you at least send a patch for this for master ?

PR #5837, opened 2019-12-05
Remove main.php file from LORIS

### LINKED ISSUE #5710: Remove main.php file
Instrument loading (now done by the instruments module) and the API (now done by the API module) were the last 2 stragglers to be converted to LORIS modules. As of LORIS 22.0, the main.php file should no longer be required (and the `.htaccess` file makes no reference to it) and it can be removed.

This should have probably been done early for 22.0, but removing it implies that a full testing round should be done, so for the next release we should get rid of main.php.

driusan:
All routing now goes through index.php and there are no more entry points that depend on main.php, as of LORIS 22.0. This removes the file and a straggling reference to it in the helpHandler.js (the help itself does not depend on main.php).

Resolves #5710.


htdocs/js/helpHandler.js
```diff
@@ -93,7 +93,7 @@ $(document).ajaxError(function(event, jqxhr, settings, thrownError) {
         };
         $.ajax({
           type: 'post',
-          url: loris.BaseURL + '/main.php',
+          url: loris.BaseURL + '/',
```
maltheism: @driusan I believe the `+ '/'` is unnecessary. Tested your PR without it and couldn't find any issue.

driusan: Okay, I removed it.. to be honest I'm not sure that any of this part of the code is necessary, I don't know what it's trying to do.

maltheism: @driusan seems to be used for logging in. I think it will be unnecessary when the react login pr is merged.

PR #5838, opened 2019-12-05
[SQL] Remove permissions for data_release in LorisMenu

ridz1208:
## Brief summary of changes
The permissions removed in this PR are not enforced at the module level, they are solely on the LorisMenu and thus not very secure. Also, they force users to get eith upload permission or edit access permission to just be able to view the module. Users should be able to view the module without any of these permissions.

This change might need to be adapted to 22 before it is released to 22.0.1


johnsaigle:
Does the module still show up in the menu without these permissions?

johnsaigle:
This PR seems to conflict with Dave's changes to the Menu permissions. if we merge his approach, this PR won't be needed.

ridz1208:
@johnsaigle yes it always shows up for everyone, which is expected

In 22 we added the data_release_view permission

SQL/New_patches/2019-12-05-Remove_lorisMenuPermissions_for_data_release.sql
```diff
@@ -0,0 +1,3 @@
+DELETE FROM LorisMenuPermissions WHERE PermID=(SELECT PermID FROM permissions WHERE code='data_release_upload');
+DELETE FROM LorisMenuPermissions WHERE PermID=(SELECT PermID FROM permissions WHERE code='data_release_edit_file_access');
```
driusan: This should go to the master branch, we don't have SQL patches with bugfix releases because it's far too complicated to keep track of what needs to be done for upgrades that way.

ridz1208: ... i dont like that we can t issue bugfixes with patches... bugs can be in the database

johnsaigle: @ridz1208 Is it possible to modify the code causing this bug on bugfix and issue the SQL changes as more of a "cleanup" on master?

ridz1208: there is no code really... it's the patch and the schema changes. its just SQL.

i think we just need a documented procedure on how to handle this

ridz1208:
@driusan I moved it to clean up patches is that enough ?

PR #5839, opened 2019-12-06
[Core] Add modules table and make modules responsible for maintaining LORIS menu

driusan:
There was no previously centralized list of modules in the LORIS database, but the module descriptor was lazily loaded when trying to access a module. As a result features such as the menu required a centralized configuration (at one point in config.xml, now in the `LorisMenu` table), modules could not be deactivated (other than by deleting them) and the menus table had to be manually maintained (by our release patches, with no front end to maintain it) including ensuring the permissions were synchronized with the code.

This adds a `modules` table which lists modules that are available (and still needs to be maintained by our release patches currently.) This allows us to add a `getMenuItems` function to the \Module class, which modules can use the register which menu items should be added to LORIS menu and can be managed by the module itself without SQL access. It also ensures that the menu items are directly tied to the user's permissions by using the same `hasAccess` function that the router uses to validate, rather than the approximation of the `LorisMenuPermissions` table.

Modules must now implement a `getLongName` function to return the human friendly name for the module to be used by default in menus (but can also be used other places that a human readable version of the module name is needed in the future.)

## Testing

1. Run the patch `SQL/New_patches/2019-12-05-AddModuleTable.sql`
2. Ensure that the menu links should all go to the same places
3. You can also change the Active flag for a module or two and they should be removed from the menu.. but since the Modules table should have every module in it in the patch provided, and there's no way to currently change the active flag, that's not something that users can do right now.


modules/document_repository/php/module.class.inc
```diff
@@ -26,4 +26,38 @@ namespace LORIS\document_repository;
...
+     * @param \User $user The user whose access is being checked.
+     *
+     * @return bool whether access is granted
+     */
+    public function hasAccess(\User $user) : bool
+    {
+        if (parent::hasAccess($user) == false) {
+            return false;
+        }
+        return $user->hasPermission('document_repository_view');
```
johnsaigle:
```suggestion
        return $user->hasAnyPermission(
            ['document_repository_view', 'document_repository_delete']
        );
```

php/libraries/Module.class.inc
```diff
@@ -142,15 +165,32 @@ class Module extends \LORIS\Router\PrefixRouter implements RequestHandlerInterfa
...
      */
     public function getName() : string
     {
         return $this->name;
     }
 
+    /**
+     * Return the long name of this module. The long name is intended
+     * to be the friendly, human-readable name and can not be automatically
+     * derived. It may contain spaces or, capitalization. The long name
```
johnsaigle:
```suggestion
     * derived. It may contain spaces or capitalization. The long name
```

php/libraries/NDB_Form.class.inc
```diff
@@ -65,15 +65,10 @@ class NDB_Form extends NDB_Page
...
-        try {
-            if (isset($_REQUEST['test_name'])) {
-                $module = \Module::factory($_REQUEST['test_name']);
-            } else {
-                $module = \Module::factory($name);
-            }
-        } catch(\LorisModuleMissingException $e) {
-            $module = new Module($name, "");
+        if (isset($_REQUEST['test_name'])) {
+            $module = \Module::factory($_REQUEST['test_name']);
```
johnsaigle: I think these lines could be condensed using `??`

src/GUI/MenuItem.php
```diff
@@ -0,0 +1,53 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

src/Middleware/PageDecorationMiddleware.php
```diff
@@ -42,13 +42,15 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
             )->process($request, $handler);
         }
 
+        $DB = \Database::singleton();
```
johnsaigle: Could you change this to use the Factory declaration instead?

modules/document_repository/php/module.class.inc
driusan: I modeled the hasAccess functions for modules that were missing it on the `LorisMenuPermissions` values in `SQL\0000-00-02-Menus.sql`.. this isn't the current behaviour.

php/libraries/NDB_Form.class.inc
driusan: I didn't write these lines (well maybe I did, but not in this PR).. this is just removing the try/catch block because `\Module` can't be instantiated since it's abstract..

src/Middleware/PageDecorationMiddleware.php
driusan: done

php/libraries/Module.class.inc
driusan: I don't think this is valid grammar.. there are only 2 things in the "or", not a comma separated list of things.

driusan:
What does RB need? I looked for all the instances of `0000-00-02-Menus.sql` and replaced it with `0000-00-02-Modules.sql`

php/libraries/Module.class.inc
johnsaigle: Yeah I think we're agreeing? Your comment as the comma.

modules/document_repository/php/module.class.inc
johnsaigle: It matches the hasAccess function for the module though. I think the intended behaviour is that someone with delete permissions should be able to view -- how else would they delete?

If it wasn't in the SQL table and showing up in the menu then I think that was an oversight. No harm fixing it here, imo.

php/libraries/Module.class.inc
driusan: oh, oops, I read the diff backwards.

driusan:
This seems to be getting stuck on "waiting for mysqld" in Travis and I'm not sure why..

modules/document_repository/php/module.class.inc
driusan: This *is* the `hasAccess` function for the module.. but I just checked and you're right, it doesn't match any of the other pages, so I guess we can fix it here.

johnsaigle:
I checked out this branch and ran `npm run tests:unit` and got this error:

There was 1 failure:

1) NDB_PageTest::testConstruct
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'test_module'
+''

/app/test/unittests/NDB_PageTest.php:73

I'm not sure that this is what is confusing Travis but it seems like it will need to be changed anyway.

SQL/0000-00-02-Modules.sql
```diff
@@ -0,0 +1,45 @@
+CREATE TABLE `modules` (
+  `Name` varchar(255) NOT NULL,
```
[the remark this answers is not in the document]
driusan: since there's nothing referencing it at the moment, I defined the Name as the primary key. It can be revisited if we end up with foreign keys to the table later.

christinerogers:
> Modules must now implement a `getLongName` function to return the human friendly name for the module to be used by default in menus (but can also be used other places that a human readable version of the module name is needed in the future.)

This is great!  I'll add an Issue so that we replace the Breadcrumb-displayname of the module with this value too.  

adding "needs documentation" label because the setup docs that this is now modules are managed.

driusan:
@christinerogers what setup docs are you talking about? As far as I know this should not affect anything about the setup.

christinerogers:
> @christinerogers what setup docs are you talking about? As far as I know this should not affect anything about the setup.

Sure, I can elaborate. We often get the question from users setting up LORIS: How do I activate/hide modules?
This should be added to @ridz1208 's new setup guide -- old docs for this include: 
*  [this old page](https://github.com/aces/Loris/wiki/LORIS-Modules) to be improved upon
* defining the long name when adding a module e.g. in: https://github.com/aces/Loris/wiki/How-to-make-a-LORIS-module

src/Middleware/UserPageDecorationMiddleware.php
```diff
@@ -52,11 +54,51 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
+        // Coincidentally, this almost-alphabetic sorting results in the exact same
+        // order that LORIS previously had.
+        $adminmenu = $menu['Admin'];
+        unset($menu['Admin']);
+        ksort($menu);
+        $menu['Admin'] = $adminmenu;
+
+        // We unconditionally sort within each section because the within-menu
+        // order has never been stable in LORIS, and sorting adds some consistency.
+        foreach ($menu as $cat => $val) {
```
kongtiaowang: $val = is_null($val) ? array() : $val; 
It will fix the first failure.
PHP warning will be printed out in Travis. 
PHP Warning:  usort() expects parameter 1 to be array, null given in /var/www/Loris/src/Middleware/UserPageDecorationMiddleware.php on line 93

driusan:
None of those documents are in the code or can be updated by the PR, so I'm removing the label.

src/Middleware/UserPageDecorationMiddleware.php
driusan: thanks, fixed

driusan:
@johnsaigle @PapillonMcGill this had to have a bunch of changes to get Travis to pass, can you re-review?

src/GUI/MenuItem.php
```diff
@@ -0,0 +1,53 @@
+<?php declare(strict_types=1);
```
PapillonMcGill: Should this declare be on line 2?

Not a blocking thing.
If you change, you could also add ending dots at line 13 and 14.

driusan: I just grepped through the code for "strict_types" and it seems to be on the same line as `<?php` in all the other files in our code.

PapillonMcGill: Yes, I saw that in a few others files I reviewed this morning. Wasn't really sure about that.

johnsaigle:
Could you provide testing instructions? Even if it's just as simple as "make sure the menu links still work."

driusan:
Everything should work as before with no user noticeable changes. So, I guess:

1. To test this: completely follow every module's test plan, @johnsaigle.

johnsaigle:
Yikes. I'll just review the code then and place my faith in Travis.

driusan:
If you're only interested in the menu portion:

1. The menu links should all go to the same places 
2. You can also change the Active flag for a module or two and they should be removed from the menu.. but since the Modules table should have every module in it in the patch provided, and there's no way to currently change the active flag, that's not something that users can do right now.

modules/acknowledgements/php/module.class.inc
```diff
@@ -24,4 +24,43 @@ namespace LORIS\acknowledgements;
...
+    {
+        if (parent::hasAccess($user) == false) {
+            return false;
+        }
+        return $user->hasAnyPermission(
+            [
+                'acknowledgements_view',
+                'acknowledgements_edit'
+            ]
+        );
```
johnsaigle: This may be better written as 

```php
return parent::hasAcess($user)
    && $user->hasAnyPermission(
    [
        'acknowledgements_view',
        'acknowledgements_edit'
    ]
);

File: modules/behavioural_qc/php/module.class.inc [hunk 10]
```diff
@@ -23,4 +23,38 @@ namespace LORIS\behavioural_qc;
...
+     * @param \User $user The user whose access is being checked.
+     *
+     * @return bool whether access is granted
+     */
+    public function hasAccess(\User $user) : bool
+    {
+        if (parent::hasAccess($user) == false) {
+            return false;
+        }
+        return $user->hasPermission('quality_control');
```
johnsaigle: Same here, an `&&` statement is a little nicer imo

modules/candidate_list/php/module.class.inc
```diff
@@ -26,4 +26,38 @@ namespace LORIS\candidate_list;
...
+    /**
+     * {@inheritDoc}
+     *
+     * @param \User $user The user whose access is being checked.
+     *
+     * @return bool whether access is granted
+     */
+    public function hasAccess(\User $user) : bool
+    {
+        if (parent::hasAccess($user) == false) {
```
johnsaigle: And here. I won't keep commenting about it though

php/libraries/NullModule.class.inc
```diff
@@ -0,0 +1,52 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

modules/issue_tracker/php/issuerow.class.inc
```diff
@@ -29,18 +29,23 @@ class IssueRow implements \LORIS\Data\DataInstance
...
     /**
      * Create a new IssueRow
      *
-     * @param array   $row The row (in the same format as \Database::pselectRow
-     *                     returns
-     * @param integer $cid The centerID affiliated with this row.
+     * @param array   $row    The row (in the same format as \Database::pselectRow
+     *                        returns.
+     * @param integer $cid    The centerID affiliated with this row.
+     * @param \Module $module The module that this issue is for.
```
johnsaigle: This is a cool feature. It would be good to update the TestPlan to provide some instructions to make sure it's working.

driusan: As far as I know this doesn't change the feature other than using the abstract function to get the name

modules/acknowledgements/php/module.class.inc
driusan: I somewhat disagree about which one is easier to read, but I don't feel strongly about it so I can change it..

johnsaigle: It's not a big deal

driusan: I updated them.

johnsaigle:
I think this PR should include changes to the CHANGELOG file based on the labels you've added

driusan:
Done

CHANGELOG.md
```diff
@@ -8,9 +8,11 @@ core section.***
 - ***When possible please provide the number of the pull request(s) containing the 
 changes in the following format: PR #1234***
 
-### VERSION
+### LORIS 23.0 (Released: ??)
+
 
 #### Core
+- Menus are now maintained by modules and no longer in the SQL database
```
johnsaigle:
```suggestion
- Menus are now maintained by modules and no longer in the SQL database (PR #5839)
```

driusan:
Two approvals! Probably a record speed for a blocking change of this size! \o/

PR #5840, opened 2019-12-06
Add CHANGELOG file

ridz1208:
## Brief summary of changes
As discussed during post-mortem, adding the changelog file for tracking all changes on Pr at a time.

I'm open to any suggestions on the format/location/name/... of the document.

The faster we get this in the faster people can start contributing to it

NB: we should expect several conflicts in this file as PRs modify it independently... that might cause an issue


maltheism:
@ridz1208 maybe "release_notes" is a bad name for the intended purpose? I feel like the name "release_notes" should be for the official release notes summary. Maybe I'm wrong but I feel like something like tracking_changes would be better for the intended purpose.

maltheism:
@ridz1208 I really like how Node.js keeps track of changes by the format `node/doc/changelogs/CHANGELOG_V13.md` and see [here](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V13.md#13.3.0).

johnsaigle:
Agreed that I prefer the name `CHANGELOG.md`, as it's a file I've come across in other open source projects.

I'd also recommend putting it in the main folder, similar to README. I think it should be very easy to find when visiting the repo.

`docs/` itself should also ideally trend toward being the home for ReadTheDocs and nothing else, imo.

driusan:
I also prefer `CHANGELOG.md` (or `NEWS.md`) which seem to be more standard.

christinerogers:
I agree -- let's aim for a Change Log.  And out of that, we'll formulate the Release Notes.  

Rida would you consider also updating this PR title?   Change log is more in line with what was discussed in the meeting.

johnsaigle:
My preference is `NEWS.md` actually. I don't think we need to catalog each and every change which is more what a "change log" is in my mind. I think this file should be including significant changes that users would need to know -- similar to how our "add to release notes" label functions.

ridz1208:
@maltheism @driusan @christinerogers @johnsaigle 

As I said in the description I dont mind renaming/ changing whatever. you just need to let me know to what.

For `/CHANGELOG.md` hit :rocket: 
For `/NEWS.md` hit :tada:

ridz1208:
all of @driusan's  votes were cancelled cause he can't follow the rules.

CHANGELOG it is

ridz1208:
@maltheism @driusan @christinerogers @johnsaigle

done

CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
+## CHANGELOG
+
+***This document logs all changes to the core code or individual modules which cause a
```
driusan: This paragraph doesn't really belong in the file. It's superfluous information for users who are reading it.

CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
...
+core section.
+
+- When possible please provide the number of the pull request(s) containing the 
+changes in the following format: PR #1234
+
+
+### Core
+
+
+### Module
```
driusan:
```suggestion
### Modules 
```

CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
...
+
+- When adding content to this document, make sure to create a section for each module 
+if the changes only impact a single module and that section does not already exist in
+the document. When changes affect the entire software, make sure to add them in the 
+core section.
+
+- When possible please provide the number of the pull request(s) containing the 
+changes in the following format: PR #1234
+
+
```
driusan: There should be a version number header here, since it's a changelog it will track more than 1 release.

CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
...
+### Core
+
+
+### Module 
+#### module1
```
driusan: There must already be some things that have been merged worth adding for a start? Otherwise the things that have been already merged into master will likely be missed when doing the release.

CHANGELOG.md
ridz1208: the problem with that is people submitting content to the version number but their PRs only being merged for another version... it's the same issue with preselecting milestones and stuff

i think we should put the version number before release

CHANGELOG.md
ridz1208: yes but i will add them after merging this in a separate PR because we might have a discussion on each element and I want people submitting PRs to start contributing asap

CHANGELOG.md
driusan: They add it to the top (whatever the next version is going to be) and then we change the heading to the version number before the release

ridz1208:
@driusan please re review

driusan:
@ridz1208 My previous comment doesn't seem to be addressed (re: version number)

ridz1208:
@driusan oops

done now

johnsaigle:
Should we update the CONTRIBUTING file asking people to add to the changelog?

driusan:
I think for now we can ask people when reviewing if the change seems major enough to warrant it. We don't want every single PR adding something there.

PR #5841, opened 2019-12-06
[DQT Importer] Issue where script breaks if scantype has a '-' character

ridz1208:
## Brief summary of changes

adding backticks to make sure no errors are thrown if these are dashes in the scan type name

### DIFF
--- a/tools/CouchDB_MRI_Importer.php
+++ b/tools/CouchDB_MRI_Importer.php
@@ -99,12 +99,12 @@ class CouchDBMRIImporter
             $Query .= ", (SELECT f.File FROM files f LEFT JOIN mri_scan_type msc
               ON (msc.ID= f.AcquisitionProtocolID)
               WHERE f.SessionID=s.ID AND msc.Scan_type='$scantype' LIMIT 1)
-                    as Selected_$scantype, (SELECT fqc.QCStatus
+                    as `Selected_$scantype`, (SELECT fqc.QCStatus
                     FROM files f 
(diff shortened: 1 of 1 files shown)


tools/CouchDB_MRI_Importer.php
```diff
@@ -99,12 +99,12 @@ function _generateCandidatesQuery($ScanTypes)
             $Query .= ", (SELECT f.File FROM files f LEFT JOIN mri_scan_type msc
               ON (msc.ID= f.AcquisitionProtocolID)
               WHERE f.SessionID=s.ID AND msc.Scan_type='$scantype' LIMIT 1)
-                    as Selected_$scantype, (SELECT fqc.QCStatus
+                    as `Selected_$scantype`, (SELECT fqc.QCStatus
```
maltheism: Hey @ridz1208 Backticks aren't part of standard ANSI SQL. What if we move away from mysql?

driusan: How are you suggesting the field name be quoted?

maltheism: @driusan I'm not sure whats best here. I was reading [stackoverflow](https://stackoverflow.com/questions/261455/using-backticks-around-field-names) about the topic of backticks and one person suggests to enable MySQL's ANSI mode for using double quotes instead but I'm not sure.

johnsaigle: I think there are other examples of backticks in the codebase. If we decide to support something other than MySQL/MariaDB then there will be many more changes to make. I think this is fine.

PR #5853, opened 2019-12-13
Update phan to latest release

driusan:
Update phan to 2.4.5. This fixes most of the issues in #5651.


driusan:
The last issue was fixed in the dev branch of phan, so it might be best to just wait for 2.4.6

Issue #5774, opened 2019-11-26
[Survey module] Formatting of error message 


nicolasbrossard:
When I trying to add a survey that had already been sent, I got this error message:

Instrument medical_history already exists for given candidate for visit V1

The full test name ('Medical history') should be used in this error message instead of the test ID ('medical_history').

PR #5854, opened 2019-12-13
[New Profile] new installation won't have default PSC

lingma:
## Brief summary of changes

A fresh Loris installation won't be able to add a new profile because the "Site" field won't display anything, at the same time the field "Site" is required. I don't see any place on the interface to be able to enter a Site other than database access. On the other hand, if you click "Access Profile" page, the system will show a default installed site "Data Coordinating Center" available in the search criteria zone. The information on the Site is confusing and inconsistent.

1. Fresh install a Loris site, go to "New Profile" menu, there will be nothing displayed in the "Site" zone without the patch, will show "Data Coordinating Center" with the patch applied.

* Nothing, I just found out when trying to work on the Issue #5774


modules/new_profile/php/new_profile.class.inc
```diff
@@ -173,7 +173,7 @@ class New_Profile extends \NDB_Form
         $user_list_of_sites = $user->getData('CenterIDs');
         $num_sites          = count($user_list_of_sites);
         $psc_labelOptions   = array();
-        if ($num_sites > 1) {
+        if ($num_sites > 0) {
```
PapillonMcGill: I believe this is a wanted feature to ignore `DCC`.
Would be best to modify installation procedure to add a site. And also a page under Config menu to edit sites.

@driusan @ridz1208

lingma: If this is a wanted feature to ignore DCC, then we should ignore DCC on the `Access Profile` page's drop down list, otherwise, the information is inconsistent.

johnsaigle:
I think this behaviour is intentional... sort of.

Part of the post-install process is the Define Study Sites. I think after this is done then the site list should have valid values on the front-end and we should be good to go.
https://github.com/aces/Loris/wiki/Project-Customization#4-define-study-sites

I think a better approach here would be to have this module present an error saying something like "Study Sites have not been configured for this project so new subjects cannot be added." This can be triggered when `$num_sites` is less than 2.

lingma:
I noticed early that the previous versions of Loris 20 and 21 will set the DSC to the default one if there is only 1 DSC present. I will come up with more information on this subject.

modules/new_profile/php/new_profile.class.inc
ridz1208: this is not designed to ignore the DCC this is made intentionally to hide the dropdown when the user is only affiliated to 1 site.

If user is affiliated to 1 site, that site is supposed to be selected by default and thus no need to display it on the front end. we agreed to undo that behaviour in an existing issue https://github.com/aces/Loris/issues/5380

The new behaviour should display the site list even if only 1 exists

lingma:
Following ridz1208's comment, I seems that the fix is acceptable for the time being if there is no other solution. Adding a page to manage Site is not what I could decide. It looks strange with inconsistent information, and this behavior comes only from the version 22.

ridz1208:
@lingma changing >1 to >0 is not the proper fix.

if this behaviour is problematic, the entire if statements should be removed from all functions like `process`, `validate`, ...

lingma:
OK. I have further cleaned the code. I did not want to touch too much the initial code.

modules/new_profile/php/new_profile.class.inc
```diff
@@ -310,14 +306,6 @@ class New_Profile extends \NDB_Form
             }
         }
 
-        // Validate site entered
```
ridz1208: you cant entirely delete the site validation

modules/new_profile/php/new_profile.class.inc
```diff
@@ -170,21 +170,17 @@ class New_Profile extends \NDB_Form
...
-        if ($num_sites > 1) {
-            foreach ($user_list_of_sites as $key => $siteID) {
-                $center = $DB->pselectRow(
-                    "SELECT CenterID as ID, Name FROM psc WHERE CenterID =:cid",
-                    array('cid' => $siteID)
-                );
-                if (!is_null($center)) {
-                    $psc_labelOptions[$siteID] = $center['Name'];
-                }
+        $user_sites = $user->getData('CenterIDs');
```
ridz1208: since you are renaming the variables you could go with camelCase to be more consistent

modules/new_profile/php/new_profile.class.inc
```diff
@@ -170,21 +170,17 @@ class New_Profile extends \NDB_Form
         $maxYear   = (isset($endYear, $ageMin)) ? $endYear - $ageMin : null;
 
         // Get sites for the select dropdown
-        $user_list_of_sites = $user->getData('CenterIDs');
-        $num_sites          = count($user_list_of_sites);
-        $psc_labelOptions   = array();
-        if ($num_sites > 1) {
```
ridz1208: since you are removing this you will need to modify the JSX file to remove the if site is not null -> display

modules/new_profile/php/new_profile.class.inc
johnsaigle: I think we need the date of birth stuff too.

modules/new_profile/php/new_profile.class.inc
lingma: > since you are removing this you will need to modify the JSX file to remove the if site is not null -> display

I believe this part is not necessary. The site could be null if the DSC table has nothing.

ridz1208: but then there is no use of hiding it since it's required.

modules/new_profile/php/new_profile.class.inc
```diff
@@ -207,127 +203,9 @@ class New_Profile extends \NDB_Form
...
-    }
-
-    /**
-     * Validate function
-     *
-     * @param array $values the value
-     *
-     * @return array $errors
-     */
-    function _validateNewCandidate($values): array
```
ridz1208: why is the validation function being deleted ?

lingma: Oh, I wrongly understood your request early.

modules/new_profile/php/new_profile.class.inc
lingma: > you cant 
I read `cant` as `can`, in fact, this check is unnecessary.

modules/new_profile/php/new_profile.class.inc
lingma: reverted

modules/new_profile/php/new_profile.class.inc
```diff
@@ -170,21 +161,17 @@ class New_Profile extends \NDB_Form
...
-                $center = $DB->pselectRow(
-                    "SELECT CenterID as ID, Name FROM psc WHERE CenterID =:cid",
-                    array('cid' => $siteID)
-                );
-                if (!is_null($center)) {
-                    $psc_labelOptions[$siteID] = $center['Name'];
-                }
+        $userSites = $user->getData('CenterIDs');
+        $sites     = array();
+        foreach ($userSites as $key => $siteID) {
```
PapillonMcGill: instead of multiple call to the DB, you could do a query with `WHERE CenterID IN (...)` and provide a comma separated list of values. $userSites is obtained from the code which get it from the DB.

modules/new_profile/php/new_profile.class.inc
```diff
@@ -229,13 +216,14 @@ class New_Profile extends \NDB_Form
         $user         = \User::singleton();
         $db           = \Database::singleton();
 
-        // Validate DOB fields
+        // Validate DOB fields, these checks are redundant
```
PapillonMcGill:
```suggestion
        // Validate DOB fields
```

PR #5855, opened 2019-12-13
[Survey Accounts] Fix formatting of error message

lingma:
## Brief summary of changes

Use the designed function to get the full name of the test name.

1. Go to Admin / Survey module, and try to send a candidate's existing Instrument. a translated message should appear.

* Issue #5774

### DIFF
--- a/modules/survey_accounts/php/addsurvey.class.inc
+++ b/modules/survey_accounts/php/addsurvey.class.inc
@@ -112,13 +112,12 @@ class AddSurvey extends \NDB_Form
             if ($values['Test_name'] == $instrument['Test_name']) {
                 $test_name = $db->pselectRow(
                     "SELECT Full_name FROM `test_names` WHERE Test_name = :test_name",
-                        array(
-                            'test_name' => $values['Test_name']
-                        )
-                    );
-
+                    array(
+                        'test_name' => $values['Test_name']
+                    )
+                );
                 return array(
-                    'Test_name' => "Instrument ". 
+                    'Test_name' => "Instrument ".
                                        ($test_name? $test_name['Full_name']: $values['Test_name']).
                                        $reminder. $values['VL'],
                 );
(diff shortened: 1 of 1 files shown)


modules/survey_accounts/php/addsurvey.class.inc
```diff
@@ -110,9 +110,16 @@ class AddSurvey extends \NDB_Form
         $reminder        = " already exists for given candidate for visit ";
         foreach ($instrument_list as $instrument) {
             if ($values['Test_name'] == $instrument['Test_name']) {
+                $test_name = $db->pselectRow(
```
PapillonMcGill: Should use `pselectOne`, this way, you'll have a single value instead of an array.
```suggestion
                $test_name = $db->pselectOne(
```

modules/survey_accounts/php/addsurvey.class.inc
```diff
@@ -110,9 +110,16 @@ class AddSurvey extends \NDB_Form
...
+                    "SELECT Full_name FROM `test_names` WHERE Test_name=:test_name",
+                    array(
+                        'test_name' => $values['Test_name']
+                    )
+                );
                 return array(
-                    'Test_name' => "Instrument ". $values['Test_name'].
-                                       $reminder. $values['VL'],
+                    'Test_name' => "Instrument ".
+                        ($test_name? $test_name['Full_name']:$values['Test_name']).
```
PapillonMcGill:
```suggestion
                        ($test_name? $test_name:$values['Test_name']).
```

lingma:
> `getFullName()` on the NDB_BVL_Instrument class.

I believe the `getFullName()` on the NDB_BVL_Instrument class is not suitable for this case, unless we modify a lot the module. Moreover we need to support the case with `LegacyInstrumentTrait`

driusan:
I don't understand what you mean by "modify a lot of modules". There's one place that you're modifying, but you're getting the information from the wrong place. `getFullName()` is already defined for every instrument. The contract of the method is that it returns the exact information you're looking for--the full name of the instrument. It may or may not access the database to do that for any given instrument. The table that you're accessing is not a reliable source for the information.

lingma:
You are right, I have fixed the problem, but I forgot to check indentation. Right now the new commit on indentation fix is not showing on this page, don't know whether because there is too many commits. I will check back later on this PR.

PR #5858, opened 2019-12-16
Document composer version

### LINKED ISSUE #5845: Document required composer version
We should be explicit about what version of composer is required for LORIS< similar to how we specify nodejs >= v8.0.0.

Running `composer -v` gives me version `Composer version 1.7.2 2018-08-16 16:57:12` on my VM. This seems to work fine.

johnsaigle:
## Brief summary of changes

Adds an explicit minimum version for composer based on discussion in #5845. 

There was some concern on the mailing list that an issue during the install was caused by an old composer version. It makes sense to be a little more specific about it in our documentation.

* Resolves #5845

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Deploy and log in with username *admin* and the password that's set up during de
  * Apache **2.4** or higher
  * MySQL >= 5.7 (or MariaDB >= 10.3) 
  * PHP <b>7.2</b> or higher
- * [Composer](https://getcomposer.org/)
+ * [Composer](https://getcomposer.org/) <b>1.4.2</b> or higher
  * NodeJS <b>8.0</b> or higher
  * NPM
  * make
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,7 +29,7 @@ Deploy and log in with username *admin* and the password that's set up during de
  * Apache **2.4** or higher
  * MySQL >= 5.7 (or MariaDB >= 10.3) 
  * PHP <b>7.2</b> or higher
- * [Composer](https://getcomposer.org/)
+ * [Composer](https://getcomposer.org/) <b>1.4.2</b> or higher
```
driusan: how did you arrive at this version? It seems oddly specific for a minimum version.

johnsaigle: I made an issue (linked above) and Tom said this version works for him. I figured we'd use the oldest one reported.

driusan: can we just say > 1.4? 1.4.2 makes it sound like we have some known bug with 1.4.1

johnsaigle:
```suggestion
 * [Composer](https://getcomposer.org/) <b>1.4</b> or higher
```

PR #5847, opened 2019-12-09
[Dashboard/Recruitment -- view site breakdown] Restricting users to see statistics only from sites they have access to.

racostas:
### Description of the issue
In **dashboard-> Recruitment-> View site breakdown** users were able to see statistics from sites they do not belong/have access

### Testing instructions

Now each user should be able to see the **view site breakdown** only for the sites it belong/have access

### Notes

This is a fix for 22.0.0, major design changes as creating a permission dedicated to statistics still pending.

**Notes relative to the code-set:**

**Note 1:** using the function `getSiteList() `pending to by refactored to act as `getAssociativeSiteList()` now shows stats from DCC for user having 'access_all_profiles' permission .

**Note 2:**  for the `/User `class we don’t have currently a similar function to `getSiteList()` but could be useful. In place, two of function are used in the current PR: `getCenterIDs()` and after creating the centrer from this ID then `getCenterName()`

PR #5861, opened 2019-12-17
[Dashboard/Study Progression -- View recruitment per site] Restricting users to see statistics only from sites they have access to.

racostas:
### Brief summary of changes

In **dashboard-> Study Progression-> View recruitment per site** users were able to see statistics from sites they do not belong/have access

### Testing instructions

Now each user should be able to see the **view recruitment per site** only for the sites it belong/have access

### Notes
This is a fix for 22.0.0, major design changes as creating a permission dedicated to statistics still pending.

### Notes relative to the code-set:

The structure:
```
$recruitmentStartDate = $DB->pselectOne(
    "SELECT MIN(Date_registered)
     FROM candidate
     WHERE RegistrationCenterID IN (:Sites)",
    array(*'Sites' => implode(",", array_keys($list_of_sites))*)
);
```
Doesn't seem to be working for me. **Only the first element of the array** `$list_of_sites` is been passed to the IN SQL clause. Not sure if I'm using correctly this way. For the moment I write it as:

```
$recruitmentStartDate = $DB->pselectOne(
    "SELECT MIN(Date_registered)
     FROM candidate
     WHERE RegistrationCenterID IN (" . $sitesString . ")",
    array()
);
```

(Please see additional related code notes on https://github.com/aces/Loris/pull/5847)

### Links to related tickets (GitHub, Redmine, ...)

https://github.com/aces/Loris/pull/5847

### DIFF
--- a/modules/dashboard/ajax/get_recruitment_line_data.php
+++ b/modules/dashboard/ajax/get_recruitment_line_data.php
@@ -15,22 +15,40 @@
 header("content-type:application/json");
 ini_set('default_charset', 'utf-8');
 
-$DB = Database::singleton();
+$DB            = Database::singleton();
+$currentUser   = \User::singleton();
+$site          = array();
+$list_of_sites = array();
+
+//TODO: Create a permission specific to statistics
+if ($currentUser->hasPermission('access_all_profiles')) {
+    $list_of_sites = \Utility::getSiteList();
+} else {
+    $site_id_arr = $currentUser->getCenterIDs();
+    foreach ($site_id_arr as $key => $val) {
+        $site[$key]          = &Site::singleton($val);
+        $list_of_sites[$val] = $site[$key]->getCenterName();
+    }
+}
+
+$sitesString = implode(",", array_keys($list_of_sites));
 
 $recruitmentData      = array();
 $recruitmentStartDate = $DB->pselectOne(
-    "SELECT MIN(Date_registered) FROM candidate",
+    "SELECT MIN(Date_registered)
+     FROM candidate
+     WHERE RegistrationCenterID IN (" . $sitesString . ")",
     array()
 );
 $recruitmentEndDate   = $DB->pselectOne(
-    "SELECT MAX(Date_registered) FROM candidate",
+    "SELECT MAX(Date_registered) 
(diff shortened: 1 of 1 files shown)


johnsaigle:
>     array(*'Sites' => implode(",", array_keys($list_of_sites))*)

I'm a bit confused about the `*`s in the example. Are these meant for Markdown emphasis or are they actually in the code?

racostas:
Hi @johnsaigle. Thanks. No, sorry, the `*s` was trying to put that part of the text in bold here in github. My bad. 

In the actual code is without it, just: `array('Sites' => implode(",", array_keys($list_of_sites)))`. 

As mentioned before, **it is**  able to retrieve the first element in the array but **only the the first**. So the behaviour I'm seeing is that the query does `IN ('FIRST_ELEMENT', 'ignore', 'the', 'rest').` It does not break but just ignore the rest of the elements acting more like an `"="`.

Cheers,
Rolando.

johnsaigle:
Thanks for the feedback.

I'm not sure why the code is behaving that way. I think it would be good to dig into why that's happening because it's not usually a good idea to take strings and put them directly into SQL statements. It's likely safe in this case but it would be better to use prepared statements if possible.

I don't have time to look into it right now but I can look at it soon and try to find a solution if no one else weighs in first.

modules/dashboard/ajax/get_recruitment_line_data.php
```diff
@@ -15,23 +15,41 @@
...
+$DB            = Database::singleton();
+$currentUser   = \User::singleton();
+$site          = array();
+$list_of_sites = array();
+
+//TODO: Create a permission specific to statistics
+if ($currentUser->hasPermission('access_all_profiles')) {
+    $list_of_sites = \Utility::getSiteList();
+} else {
+    $site_id_arr = $currentUser->getCenterIDs();
```
PapillonMcGill: Need to remove `DCC` center (ID=1) to follow current implementation

racostas:
I'm aborting this PR in favour of the great major redesign proposed on #5896. Nice!

racostas:
Deprecated by #5917

Issue #4995, opened 2019-08-01
Post-release 21 Documentation updates


christinerogers:
(non-exhaustive list, in progress) 
Needs cross-checking with new [Readmes from Wiki-revamp](https://github.com/aces/Loris/pull/4093)

### Things moved under docs/ RaisinBread -> broken links 

*  public instruments removed / moved to Raisinbread
- [ ]  bmi.linst which used to be [under docs/instruments/](https://github.com/aces/Loris/tree/20.3-release/docs/instruments)
- [ ] MRI Parameter form which is required(?) for the upcoming QC Browser (can't load without)

- [ ] requires updating the Behavioural Setup Guide... probably other pages too.
* readmes : e.g. https://github.com/aces/Loris/blob/master/LORIS_react.README.md
- [x] #5392 broken links in Contributing.md, 
- [x] several wiki pages
- [ ] other places in the Repo

### Install process Readme(s): 

- [x] #5197 Makefile needs to be run for fresh installs. not composer (? TBC) : add to Readmes, wiki
- [x] #5017 NodeJS version matters (PR sent for main readme only)
- [x] #5018  CentOS 21.0.0 specific updates (including NodeJS)

### Reliability module removed :  Remove all mentions #4596
- [x] Code: [statistics/js/form_stats_reliability.js](https://github.com/aces/Loris/blob/21.0-release/modules/statistics/js/form_stats_reliability.js) still exists (technically this is Stats module code cleanup, not Reliability module 
- [x] config.xml #5433

### Training module removed  - Documentation to be updated: 
- [x] [Examiner module readme](https://github.com/aces/Loris/tree/21.0-release/modules/examiner) : remove mentions of Training module

### Reactified modules: 

- [x] #5410 Remove [mentions of "Show data"  button](https://github.com/aces/Loris/search?q=%22Show+data%22&unscoped_q=%22Show+data%22) if no longer relevant 
e.g. [15 markdown files](https://github.com/aces/Loris/search?l=Markdown&q=%22Show+data%22) files including: 
- [x] Imaging Browser Help text
- [ ] (very minor) [5 Test plans to be updated](https://github.com/aces/Loris/search?q=%22Show+data%22+test_plan&unscoped_q=%22Show+data%22+test_plan) 

### Tools scripts moved to a different directory
- [ ] Wiki / wiki-readmes to be updated

### CenterID -> Registration Center ID  #4078

[...]

PR #5874, opened 2019-12-17
Change remaining instances of "gender" to "sex"

johnsaigle:
Documentation update requested in https://github.com/aces/Loris/issues/4995

### DIFF
--- a/docs/API/LorisRESTAPI.md
+++ b/docs/API/LorisRESTAPI.md
@@ -279,7 +279,7 @@ PSCID is only required if the generation type in the Loris config is set to
 The candidate will be created at the site of the user using the API's site.
 A response code of 201 Created will be returned on success, 409 Conflict if
 the PSCID already exists, and a 400 Bad Request if any data provided is invalid
-(PSCID format, date format, gender something other than Male|Female|Other, invalid project
+(PSCID format, date format, sex set to something other than Male|Female|Other, invalid project
 name, etc). A successful POST request will return the CandID for the newly
 created candidate in a JSON object of the form: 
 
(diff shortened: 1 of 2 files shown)


docs/API/LorisRESTAPI.md
```diff
@@ -279,7 +279,7 @@ PSCID is only required if the generation type in the Loris config is set to
 The candidate will be created at the site of the user using the API's site.
 A response code of 201 Created will be returned on success, 409 Conflict if
 the PSCID already exists, and a 400 Bad Request if any data provided is invalid
-(PSCID format, date format, gender something other than Male|Female|Other, invalid project
+(PSCID format, date format, sex set to something other than Male|Female|Other, invalid project
```
driusan: This change doesn't make sense. The JSON object (which can't be changed because it's a version of the API that's already been released as stable) is "Gender", not "Sex")

johnsaigle: OK, I didn't know. i'll get rid of this change.

christinerogers:
hmmm @driusan to be clear,  this gender -> sex change in release 21 wasn't made in the API?  
(we should add an issue)

driusan:
It was made to v0.0.3 of the API. It was explicitly not done or older, released versions of the API because modifying the API would be a different API.

docs/API/LorisRESTAPI.md
johnsaigle:
```suggestion
(PSCID format, date format, gender something other than Male|Female|Other, invalid project
```

johnsaigle:
@ridz1208 Do you think you could merge this one?

PR #5879, opened 2019-12-18
[Core] Update package-lock file

johnsaigle:
## Brief summary of changes

This should prevent issues for developers when switching between branches.

Steps taken:
1. Checkout aces/master
2. `make dev`
3. `git add package-lock.json`

* Alternative to #5820

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1828,7 +1828,7 @@
     },
     "buffer": {
       "version": "4.9.1",
-      "resolved": "http://registry.npmjs.org/buffer/-/buffer-4.9.1.tgz",
+      "resolved": "https://registry.npmjs.org/buffer/-/buffer-4.9.1.tgz",
       "integrity": "sha1-bRu2AbB6TvztlwlBMgkwJ8lbwpg=",
       "dev": true,
       "requires": {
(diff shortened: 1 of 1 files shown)


maltheism:
Hey @johnsaigle I think part of the issue mentioned in #5820 might be when developers install the npm modules and the npm registry for some reason returns `http` instead of `https` for the sources of the packages. We might still run into the issue of having the package-lock in the git status but this PR will definitely make it less frequently than now.

johnsaigle:
I'm still pretty sure that's being caused by people using older versions of npm. We're supposed to be using >= 8.0 so I'd be curious to know if those issues are being caused by default npm in Ubuntu (v. 6)

maltheism:
@johnsaigle this [comment](https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/31) makes me think it's not the version of npm.

johnsaigle:
Fair enough. Maybe the solution is that cache cleaning process you mentioned in the other PR?

maltheism:
@johnsaigle I'm not sure/ I think this PR should be merged and if the issue comes up again we will know.

PR #5785, opened 2019-11-26
Rename quatUser & quatPassword to something more descriptive

maltheism:
## Brief summary of changes

The quatUser is now dbUser and the quatPassword is now dbPassword.

#### Links to related tickets (GitHub, Redmine, ...)

* Resolves https://github.com/aces/Loris/issues/5782

PR #5883, opened 2019-12-19
upgrade loris database with release_patch

kongtiaowang:
## Brief summary of changes
Add readme to explain how to run release schemas.
Add a script to help the user to upgrade.
This script only runs release_patch schemas and shows the error messages during the upgrading.
It will create an error log file under a log folder automatically if user can not update.  
This script will fail if the database structure already had been changed by users before upgrading.
It will backup the dababase first automaticlly, then run the upgrade file in the database, if there is any error, it will restroe your database.
This script is only for the users to upgrade the Loris database.
php loris_db_upgrade.php (version No.)


lingma:
First of all, line 27 of the script misses a semicolon, the script won't run without correction.
As an upgrade tool, the script doesn't look safe and useful enough to be acceptable. We could not assume that the customer authorizes the program to go ahead with the login credential taken from the config file.

The script will break out with my test run saying that "`PHP Warning:  PDO::exec(): SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'issue_tracker_url' for key 'Name' in loris_db_upgrade.php on line 47`", which is not an error from the tool itself, but if the tool won't know how to handle this kind of situation, I don't see the usability of this tool. As a result, the tool needs more work done if the direction taken is good.

README.md
```diff
@@ -92,6 +92,22 @@ If you use MySQL 8, please read [this link](https://www.php.net/manual/en/mysqli
 
 7. Follow the [Setup Guide in the LORIS Wiki](https://github.com/aces/Loris/wiki/Setup) to complete your post-installation setup and configuration, and for more documentation.
 
+### Upgrade Step
+
+Backing Up the Database Before Upgrading the Database Schemas.
+If you have installed previous version Loris, you can upgrade Loris to the latest in the following ways:
```
PapillonMcGill:
```suggestion
If you have installed a previous version Loris, you can upgrade Loris to the latest in the following ways:
```

README.md
```diff
@@ -92,6 +92,22 @@ If you use MySQL 8, please read [this link](https://www.php.net/manual/en/mysqli
 
 7. Follow the [Setup Guide in the LORIS Wiki](https://github.com/aces/Loris/wiki/Setup) to complete your post-installation setup and configuration, and for more documentation.
 
+### Upgrade Step
+
+Backing Up the Database Before Upgrading the Database Schemas.
```
PapillonMcGill:
```suggestion
It is recommended to do a backup of the Database before upgrading the database schemas.
```

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,51 @@
...
+$path    = __DIR__ . "/../SQL/Release_patches";
+
+$files = array_diff(scandir($path), array('.', '..'));
+
+echo "This script will upgrade the loris database to the lastest.\n";
+echo "Backing Up the Database Before Upgrading the Database.\n"
+
+$v = readline("\n Enter current version of Loris database (ex:21.0): ");
+  
+if (!file_exists('../project/config.xml')) {
```
PapillonMcGill: This test should be done before asking the current version of the database (line 29).

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,51 @@
...
+    die("config.xml file doesn't exist. Aborting update.");
```
PapillonMcGill: suggest printing the message with `print` or `fwrite(STDERROR, ...)` and exit() with a number other than 0.

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,51 @@
...
+
+    if (substr($file, 0, 5)>=$v) {
+     print_r($file);
+      $a = file_get_contents(__DIR__ . "/../SQL/Release_patches/".$file);
+      $sql .= $a;
+
+    }
+}
+    try {
+       $DB->_PDO->exec($sql);
```
PapillonMcGill: schema upgrade will fail as any modification to the schema (ALTER TABLE, DROP TABLE, CREATE TABLE, ...) require the `quatuser`.
Only the `INSERT, UPDATE, DELETE` modification would succeed. 
@johnsaigle created some PR to facilitate the upgrade process, maybe you could improve these.

johnsaigle: I have a tool to update the Raisinbread database but it's not safe to use for upgrading production. It deletes all data. As a result I don't think it's useful in this context.

kongtiaowang:
@lingma This script or readme file is an option for the users. It helps users to know where the update schema files are and how to update them.  In your case, it helps you find error "Integrity constraint violation: 1062 Duplicate entry 'issue_tracker_url' for key 'Name'".  It will not force the users to modify the database with real data or structure that has been changed. 
All the files under the tools folder, we already assumed that the customer authorizes the program to go ahead with the login credential taken from the config file.

tools/loris_db_upgrade.php
PapillonMcGill: @johnsaigle I though that you had started work to update a production environment.

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,52 @@
...
+ * @package behavioural
+ */
+require_once __DIR__ . "/../vendor/autoload.php";
+require_once __DIR__ . "/generic_includes.php";
+ set_include_path(get_include_path().":../project/libraries:../php/libraries:");
+require_once "../php/libraries/Database.class.inc";
+require_once "../php/libraries/NDB_Config.class.inc";
+$path    = __DIR__ . "/../SQL/Release_patches";
+
+$files = array_diff(scandir($path), array('.', '..'));
```
johnsaigle:
```suggestion
$files = glob("*.sql")
```

I think this is a bit more straightforward

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @category Install
+ * @package  Loris
+ * @author   Shen Wang <shen.wang2@mcgill.ca>
+ * @license  Loris license
+ * @link     https://www.github.com/aces/Loris-Trunk/
+ * @package behavioural
+ */
+require_once __DIR__ . "/../vendor/autoload.php";
+require_once __DIR__ . "/generic_includes.php";
```
johnsaigle: I think all of the `require_once` statements and the include path statement can be deleted because generic_includes covers this.

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,52 @@
...
+ set_include_path(get_include_path().":../project/libraries:../php/libraries:");
+require_once "../php/libraries/Database.class.inc";
+require_once "../php/libraries/NDB_Config.class.inc";
+$path    = __DIR__ . "/../SQL/Release_patches";
+
+$files = array_diff(scandir($path), array('.', '..'));
+
+echo "This script will upgrade the loris database to the lastest.\n";
+echo "Backing Up the Database Before Upgrading the Database.\n";
```
PapillonMcGill: This line would let user think the the script is doing a DB backup. 
```suggestion
```

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/php
+<?php
+
+/**
+ * This script will upgrade the loris database to the lastest. 
+ * Usage: php loris_db_upgrade.php [-D]
```
johnsaigle:
```suggestion
 * Usage: php loris_db_upgrade.php
```

The `-D` flag doesn't seem to be used in the script.

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/php
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * This script will upgrade the loris database to the lastest.
```
johnsaigle:
```suggestion
 * This script will upgrade the loris database to a specified release version.
```

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,52 @@
...
+ * @license  Loris license
+ * @link     https://www.github.com/aces/Loris-Trunk/
+ * @package behavioural
+ */
+require_once __DIR__ . "/../vendor/autoload.php";
+require_once __DIR__ . "/generic_includes.php";
+ set_include_path(get_include_path().":../project/libraries:../php/libraries:");
+require_once "../php/libraries/Database.class.inc";
+require_once "../php/libraries/NDB_Config.class.inc";
+$path    = __DIR__ . "/../SQL/Release_patches";
```
johnsaigle: It would be good to have this script exit if not being run in the `tools` directory since it won't work otherwise.

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,52 @@
...
+
+echo "This script will upgrade the loris database to the lastest.\n";
+echo "Backing Up the Database Before Upgrading the Database.\n";
+echo "User must be granted all necessary privileges (ALTER TABLE, INSERT, UPDATE, DELETE ...) on Loris database.\n";
+
+if (!file_exists('../project/config.xml')) {
+    exit("config.xml file doesn't exist. Aborting update.");
+}
+
+$v = readline("\n Enter current version of Loris database (ex:21.0): ");
```
johnsaigle: It might be a little nicer to do this with a command-line argument instead.

E.g.

`php loris_db_upgrade.php 22.0`

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,52 @@
...
+  
+$DB = \Database::singleton();
+$sql = null;
+foreach($files as $file){
+
+    if (substr($file, 0, 5)>=$v) {
```
johnsaigle: Where is the number 5 coming from? Can you rewrite this to use a string constant instead of magic numbers?

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,52 @@
...
+ set_include_path(get_include_path().":../project/libraries:../php/libraries:");
+require_once "../php/libraries/Database.class.inc";
+require_once "../php/libraries/NDB_Config.class.inc";
+$path    = __DIR__ . "/../SQL/Release_patches";
+
+$files = array_diff(scandir($path), array('.', '..'));
+
+echo "This script will upgrade the loris database to the lastest.\n";
+echo "Backing Up the Database Before Upgrading the Database.\n";
+echo "User must be granted all necessary privileges (ALTER TABLE, INSERT, UPDATE, DELETE ...) on Loris database.\n";
```
PapillonMcGill: The script use the regular user from config.xml which do not have permission to modify the schema.
Instead to echoing this file, should manually open a connection to the DB using quatuser.
see https://github.com/aces/Loris/pull/5260

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,52 @@
...
+      $sql .= $a;
+
+    }
+}
+    try {
+       $DB->_PDO->exec($sql);
+    } catch (PDOException $e){
+       echo $e->getMessage();
+    }
+echo "Upgrade completed successfully!";
```
johnsaigle: It looks like this message will be printed even then a PDOException occurs. That's probably not the right behaviour.

tools/loris_db_upgrade.php
johnsaigle: I don't think so. Most of the work I've done recently is in streamlining developer instances of LORIS. I'd be happy if something like this PR goes ahead though.

README.md
```diff
@@ -92,6 +92,22 @@ If you use MySQL 8, please read [this link](https://www.php.net/manual/en/mysqli
...
 7. Follow the [Setup Guide in the LORIS Wiki](https://github.com/aces/Loris/wiki/Setup) to complete your post-installation setup and configuration, and for more documentation.
 
+### Upgrade Step
+
+It is recommended to do a backup of the Database before upgrading the database schemas.
+If you have installed a previous version Loris, you can upgrade Loris to the latest in the following ways:
+
+1. Using Git command checkout to the latest version of Loris. 
+ ```bash
+ git checkout aces/22.0-release
```
christinerogers: this is on the master branch.  this number in this readme will need to be updated right before every release.  Can we find an easier way to write/maintain this?

README.md
```diff
@@ -92,6 +92,22 @@ If you use MySQL 8, please read [this link](https://www.php.net/manual/en/mysqli
...
+ ```
+2.-- Manual upgrade(optional): Run upgrade.sql files in your database in the /var/www/$projectname/SQL/Release_patches/ directory.
+  -- Automatic upgrade(optional): Run loris_db_upgrade.php in /var/www/$projectname/tools/ directory.
+ ```bash
+ cd /var/www/$projectname/tools/
+ php loris_db_upgrade.php
```
christinerogers: which database permissions are required to run the script?  are those of the `config.xml` credential (`lorisdbuser`) sufficient?

tools/loris_db_upgrade.php
christinerogers: Agree @johnsaigle  that it should probably target a milestone/branch name, and this should be more explicit in terms of input/output

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,52 @@
...
+ * This script will upgrade the loris database to the lastest. 
+ * Usage: php loris_db_upgrade.php [-D]
+ *
+ * PHP Version 7
+ *
+ * @category Install
+ * @package  Loris
+ * @author   Shen Wang <shen.wang2@mcgill.ca>
+ * @license  Loris license
+ * @link     https://www.github.com/aces/Loris-Trunk/
```
christinerogers:
```suggestion
 * @link     https://www.github.com/aces/Loris/
```

remove old repo name

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,52 @@
...
+ * @package behavioural
```
christinerogers:
```suggestion
 * @package core
```
if core is a package?

tools/loris_db_upgrade.php
christinerogers: I very very much agree this should go in the commandline.  👍

tools/loris_db_upgrade.php
christinerogers: agree

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,52 @@
...
+require_once __DIR__ . "/../vendor/autoload.php";
+require_once __DIR__ . "/generic_includes.php";
+ set_include_path(get_include_path().":../project/libraries:../php/libraries:");
+require_once "../php/libraries/Database.class.inc";
+require_once "../php/libraries/NDB_Config.class.inc";
+$path    = __DIR__ . "/../SQL/Release_patches";
+
+$files = array_diff(scandir($path), array('.', '..'));
+
+echo "This script will upgrade the loris database to the lastest.\n";
```
christinerogers:
```suggestion
echo "This script will upgrade the LORIS database to the latest.\n";
```
Please also add: the version number it is upgrading to.  Important for output capture/logging.

tools/loris_db_upgrade.php
christinerogers: Please move this comment to the top (first line of output) and change it to: 
```suggestion
echo "Please ensure your have backed up a copy of your database before upgrading.\n";
```

tools/loris_db_upgrade.php
christinerogers: the quatuser was for the old DQT and shouldn't even have delete permission.  Would be good to rename/fix the credential...

PapillonMcGill:
@christinerogers `quatuser` has DELETE permission, at least in every instance I manage. `lorisdbuser` as insufficient permissions

tools/loris_db_upgrade.php
kongtiaowang: In this way, I can get file names with the path. I have to split the $files again.

README.md
kongtiaowang: I will add it later.

README.md
johnsaigle: There's also no guarantee the remote will be named `aces`.

README.md
johnsaigle: The script should use `adminUser` and `adminPassword` ([formerly known as](#5785) "quatUser").

johnsaigle:
@PapillonMcGill @christinerogers N.B. quatUser was renamed to `adminUser` in #5785

tools/loris_db_upgrade.php
kongtiaowang: **14.10**_To_14.12_upgrade.sql.  
**14.12**_To_15.04_upgrade.sql. 0-5 means the version.

tools/loris_db_upgrade.php
kongtiaowang: I only need the file names, in your way, I can get file names and path. I have to split the array again.

christinerogers:
> @PapillonMcGill @christinerogers N.B. quatUser was renamed to `adminUser` in #5785

Thanks @johnsaigle - i'll ping you a side question.

tools/loris_db_upgrade.php
johnsaigle: What about e.g. 19.0 to 19.1?

tools/loris_db_upgrade.php
```diff
@@ -14,38 +14,82 @@
...
 
 if (!file_exists('../project/config.xml')) {
     exit("config.xml file doesn't exist. Aborting update.");
 }
 
-$v = readline("\n Enter current version of Loris database (ex:21.0): ");
-  
-$DB = \Database::singleton();
+$v = $argv[1];
+if (floatval($v) < 14.0 || floatval($v) > 99.0 )
```
johnsaigle: This isn't really correct. `14.19` and `56.0000000000000000000001` are invalid.

I think it would be better to read this from the filenames themselves.

You can split the filename on the `_` character and use only those exact numbers as valid values.

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * @category Install
+ * @package  Loris
+ * @author   Shen Wang <shen.wang2@mcgill.ca>
+ * @license  Loris license
+ * @link     https://www.github.com/aces/Loris/
+ * @package core
+ */
+require_once __DIR__ . "/generic_includes.php";
+ set_include_path(get_include_path().":../project/libraries:../php/libraries:");
```
johnsaigle:
```suggestion
```

tools/loris_db_upgrade.php
johnsaigle: This hasn't been resolved yet

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,95 @@
...
+  echo "Upgrade completed successfully!";
+}
+
+function write_log($log_msg)
+{
+    $log_filename = "log";
+    if (!file_exists($log_filename))
+    {
+        // create directory/folder uploads.
+        mkdir($log_filename, 0777, true);
```
johnsaigle:
```suggestion
        mkdir($log_filename, 0640, true);
```
This file should allow read and write ownership by the owner but nobody else should be able to modify it.

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,95 @@
...
+       write_log("Error No.  => " .$e['errno']);
+       write_log("SQL State  => " .$e['sqlstate']);
+       write_log("Error      => " .$e['error']);
+   }
+  echo "Check error log under log directory!";
+} else {
+  echo "Upgrade completed successfully!";
+}
+
+function write_log($log_msg)
```
johnsaigle:
```suggestion
function write_log(string $log_msg): void
```

tools/loris_db_upgrade.php
johnsaigle: You append the path in 43 though so you do need the path.

If you have the path you can always use `basename()` to get only the name.

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,95 @@
...
+   exit;
+ }
+
+
+$sql = null;
+$query = null;
+foreach($files as $file){
+
+    if (substr($file, 0, strpos($file,'_'))>=$v) {
+     print_r($file);
```
johnsaigle: It might be good to write this to the log also

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,126 @@
...
+ * @license  Loris license
+ * @link     https://www.github.com/aces/Loris/
+ * @package core
+ */
+require_once __DIR__ . "/generic_includes.php";
+
+// check the running path if not in tools, it will quit
+$currentDir = __DIR__;
+$dir = explode("/",$currentDir);
+if ( end($dir) !== "tools") {
```
johnsaigle: Instead you can delete these lines and replace with

```php
if (basename(__DIR__) !== 'tools') {
```

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,124 @@
...
+       write_log($q);
+       write_log($error);
+       write_log("----------------------------------------");
+    } else {
+           echo "OK! \n";
+      }
+
+}
+
+if (!$mysqli->error) {
```
PapillonMcGill: Will display "Upgrade completed successfully" as long as the last query executed properly even if multiple failure occurred before.

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,124 @@
...
+   echo "Upgrade completed successfully!";
+}
+
+
+function write_log(string $log_msg): void
+{
+    $log_filename = "log";
```
PapillonMcGill: This would create a "log" directory under "</path/to/loris/tools/" as long as Apache as write permission. 
- Apache might not have write permission
- It is bad practice to let Apache write inside the PHP tree.

Suggest using other location (/var/log/Apache2/ on Ubuntu or /var/log/httpd/ on CentOS)

lingma: > This would create a "log" directory under "</path/to/loris/tools/" as long as Apache as write permission.

This has nothing to do with Apache, the executor might not have any permission on Apache folder. Anyway it is a good practice to put the log other than the `tool` folder.

PapillonMcGill: @lingma Correct, I forgot that this is a script in tools and run manually.

johnsaigle: I think there's already a tools/log script used by other scripts, e.g. `install.sh`.

Also this is a back-end tool. Apache is not involved here.

README.md
```diff
@@ -92,6 +92,22 @@ If you use MySQL 8, please read [this link](https://www.php.net/manual/en/mysqli
 
 7. Follow the [Setup Guide in the LORIS Wiki](https://github.com/aces/Loris/wiki/Setup) to complete your post-installation setup and configuration, and for more documentation.
 
+### Upgrade Step
+
+It is recommended to do a backup of the Database before upgrading the database schemas.
+If you have installed a previous version Loris, you can upgrade Loris to the latest in the following ways:
```
PapillonMcGill:
```suggestion
If you have installed a previous version of LORIS, you can upgrade LORIS to the latest in the following ways:
```

README.md
```diff
@@ -92,6 +92,22 @@ If you use MySQL 8, please read [this link](https://www.php.net/manual/en/mysqli
...
+
+1. Using Git command checkout to the latest version of Loris.
```
PapillonMcGill:
```suggestion
1. Using Git command checkout the latest version of Loris.
```

README.md
```diff
@@ -92,6 +92,22 @@ If you use MySQL 8, please read [this link](https://www.php.net/manual/en/mysqli
...
+ ```bash
+ git checkout <loris-repo>/<loris-branch>
+ ```
+2.-- Manual upgrade(optional): Run upgrade.sql files in your database in the /var/www/$projectname/SQL/Release_patches/ directory.
```
PapillonMcGill:
```suggestion
2.-- Manual upgrade(optional): Run upgrade.sql files in `/var/www/$projectname/SQL/Release_patches/` directory in your database.
```

README.md
```diff
@@ -92,6 +92,22 @@ If you use MySQL 8, please read [this link](https://www.php.net/manual/en/mysqli
...
+  -- Automatic upgrade(optional): Run Ex.(php loris_db_upgrade.php 21.0) in /var/www/$projectname/tools/ directory.
```
PapillonMcGill:
```suggestion
  -- Automatic upgrade(optional): Run the upgrade script in `/var/www/$projectname/tools/` directory. Ex. `php loris_db_upgrade.php 21.0`
```

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/php
+<?php declare(strict_types=1);
+
+/**
+ * This script will upgrade the loris database to a specified release version.
```
PapillonMcGill:
```suggestion
 * This script will upgrade the LORIS database to a specified release version.
```

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,130 @@
...
+ {
+   echo "Please input an available version! \n";
+   echo "Pick one version from the version list below! \n";
+   foreach ($versionList as $list){
+      print_r($list."\n");
+   }
+   exit;
+ }
+
+// read sql line by line from sql files.
```
PapillonMcGill:
```suggestion
// read sql file needed for update.  
```
file_get_contents() read the entire file in one shoot, not line by line

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,130 @@
...
+   }
+   exit;
+ }
+
+// read sql line by line from sql files.  
+$sql = null;
+$query = null;
+foreach($files as $file){
+
+    if (substr($file, 0, strpos($file,'_'))>=$v) {
```
PapillonMcGill:
```suggestion
    if (substr($file, 0, strpos($file,'_'))<=$v) {
```

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,130 @@
...
+
+function write_log(string $log_msg): void
+{
+    $log_filename = "log";
+    if (!file_exists($log_filename))
+    {
+        // create directory/folder uploads.
+        mkdir($log_filename, 0640, true);
+    }
+    $log_file_data = $log_filename.'/log_' . date('d-M-Y') . '.log';
```
PapillonMcGill:
```suggestion
    $log_dir = "log";
    if (!file_exists($log_dir))
    {
        // create directory/folder uploads.
        mkdir($log_dir, 0640, true);
    }
    $log_file_data = $log_dir.'/log_' . date('d-M-Y') . '.log';
```
would be a more appropriate variable name.

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,130 @@
...
+        // if you don't add `FILE_APPEND`, the file will be erased each time you add a log
```
PapillonMcGill:
```suggestion
        // append the message at the end of the log file.
```

tools/loris_db_upgrade.php
kongtiaowang: @PapillonMcGill " >= "  It means all the version of patch files greater than the current version should be run.
Ex.(21.0) $file is the version fo the patch file, $v (20.0) is the current version.

PapillonMcGill: Ok, I though it should run script less than requested version.

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,130 @@
...
+//get sql file path
+$path    = __DIR__ . "/../SQL/Release_patches";
+
+// file names from the Release_patches
+$files = array_diff(scandir($path), array('.', '..'));
+
+echo "This script will upgrade the LORIS database to the latest release version.\n";
+
+if (!file_exists('../project/config.xml')) {
+    exit("config.xml file doesn't exist. Aborting update.");
```
PapillonMcGill:
```suggestion
    print("config.xml file doesn't exist. Aborting update.");
    exit(1);
```
This can help in future automation

README.md
```diff
@@ -92,6 +92,22 @@ If you use MySQL 8, please read [this link](https://www.php.net/manual/en/mysqli
 
 7. Follow the [Setup Guide in the LORIS Wiki](https://github.com/aces/Loris/wiki/Setup) to complete your post-installation setup and configuration, and for more documentation.
 
+### Upgrade Step
```
christinerogers, line 95: I think this should be a one line mention of this step/process/script -- and point to the rest of this content under the markdown docs/ (what Rida calls the "new wiki)

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,131 @@
...
+
+    }
+}
+
+$lines = explode(PHP_EOL, $sql);
+foreach ($lines as $line) {
+  // remove comment line from sql file
+   if (substr($line, 0, 2) !== '--') {
+     $query .= $line." ";
+   }
```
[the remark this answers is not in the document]
kongtiaowang: // remove comment line from sql file
   only "  -- commmnet   " will be removed from sql file.
  "select * from psc; -- comment" will keep it.

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,131 @@
...
+
+}
+
+$factory      = \NDB_Factory::singleton();
+$db_config    = $factory->config()->getSetting('database');
+$mysqli = new mysqli($db_config['host'], $db_config['adminUser'], $db_config['adminPassword'], $db_config['database']);
```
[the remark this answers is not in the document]
kongtiaowang: I need to run all the sql files line by line. Our database class doesn't have closeCursor function and in this case also need to set PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.  MYSQLI is more convenient for me.

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,131 @@
...
+   }
+   exit;
+ }
+
+// read sql file needed for update.  
+$sql = null;
+$query = null;
+foreach($files as $file){
+
+    if (substr($file, 0, strpos($file,'_'))>=$v) {
```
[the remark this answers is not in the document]
kongtiaowang: done

tools/loris_db_upgrade.php
```diff
@@ -0,0 +1,137 @@
...
+$path    = __DIR__ . "/../SQL/Release_patches";
+
+// file names from the Release_patches
+$files = array_diff(scandir($path), array('.', '..'));
+print_r($files);
+echo "This script will upgrade the LORIS database to the latest release version.\n";
+
+
+$v = $argv[1];
+$versionList = array();
```
lingma, line 35: I think the modified rules prefer `[]` now

kongtiaowang:
Because our release process is more complicated than I thought. This PR is not good enought, so I will close this PR. Maybe I will create a new one to replace it.

PR #5884, opened 2019-12-20
[Lorisform] label vs option vs attribute function misuses/wrong signatures

lingma:
## Brief summary of changes

The initial third parameters of the form elements were set as the initial fourth parameters somehow in the program, this is clearly a bug. Often options are mixed with attributes, or options are passed to label. This will create confusion for programmer. Unfortunately since the same problem existed in the previous versions, migrations will become difficult once the patch has been applied. The system will show warnings if the patch would not be applied.

[**Important**] Possible impact:
Some old instruments will stop displaying correctly once being migrated, the verification required besides some function signatures changes, including `setup`, `score`. 

The old format could be working in the following way (_I just pass the label at the wrong place to make the system work, not sure whether there are other peoples who were coding in this way._):
`$this->form->createElement('static', null, null, "label");`

The new format:
`$this->form->createElement('static', null, "label");`

The instrument migration fix could be easy with a full text replacement, with verification.

The removal of the variable `$closer` is not too much related with the main task.

Note: I noticed that the [wiki](https://github.com/aces/Loris/wiki/Instrument-Groups) manual uses the format:
`$group[] =& $this->form->createElement("static", null, null, "<label>");`

1.

N/A

### DIFF
--- a/php/libraries/LorisForm.class.inc
+++ b/php/libraries/LorisForm.class.inc
@@ -1959,11 +1959,11 @@ class LorisForm
     ) {
         switch($type) {
         case 'text':
-            return $this->createText($elname, $label, $options);
+            return $this->createText($elname, $label, $attribs);
         case 'select':
             return $this->createSelect($elname, $label, $options, $attribs);
         case 'submit':
-            return $this->createSubmit($elname, $label, $options);
+            return $this->createSubmit($elname, $label, $attribs);
         case 'static':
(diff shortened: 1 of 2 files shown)


lingma:
doc modification and removal of the variable **`$closer`**

lingma:
This PR has been recreated and combined with PR #5886

PR #5887, opened 2019-12-24
[Dashboard] Fix study progression performance

driusan:
For long running studies the "study progression" widget on the
dashboard was doing an excessive number of queries on the database
(one per site per month of the study having been running) which
was causing the dashboard to take an excessive amount of time to
load.

This replaces it with a single query using a GROUP BY, and then
massages the data in PHP to the same format.

To test this compare the dashboard/ajax/get_scan_line_data.php results
before and after this change. The JSON returned should be identical, but the
performance should be faster (if there's a large enough amount of data
that the before was noticeably slow to load.).

### DIFF
--- a/modules/dashboard/ajax/get_scan_line_data.php
+++ b/modules/dashboard/ajax/get_scan_line_data.php
@@ -40,79 +40,58 @@ $scanEndDate        = $DB->pselectOne(
      WHERE pt.Name='acquisition_date'",
     array()
 );
-$scanData['labels']
-    = createLineChartLabels($scanStartDate, $scanEndDate);
+
+// Run a query to get all the data. Order matters to ensure that the labels are calculated
+// in the correct order.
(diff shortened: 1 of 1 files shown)


driusan:
(Note: get_recruitment_line_data probably should get a similar treatment, but the query on that one isn't quite as expensive so it takes seconds instead of minutes to run on my sandbox.)

modules/dashboard/ajax/get_scan_line_data.php
```diff
@@ -25,94 +25,102 @@
...
     );
 }
 
-print json_encode($scanData);
-
-return 0;
-
 /**
- * Create chart labels (dates)
+ * Massages the data passed into the format expected by C3.
```
PapillonMcGill: What is the meaning of `C3` ?

driusan: It's the graphing library used by the dashboard

PapillonMcGill: Ok thanks, Would it be possible to add `... by the C3 library` in that case?
I will approve now as it is really minor.

driusan: okay, changed.

PR #5888, opened 2019-12-30
[timepoint, instrument] pageclass obj null warning

lingma:
## Brief summary of changes

Each page loading of timepoint list, instrument list and instrument will trigger a warning "PHP Notice:  Trying to get property 'page' of non-object in UserPageDecorationMiddleware.php on line 94", which is not pleasant to see when you are working with them every day. The problem comes from the Loris logic to process first, if not having gone through then to fail back to timepoint/instrument. I don't want to modify the logic though I assume that two times' processing might not be necessary, but it is too soon to say not. As a result, I just add a check to avoid the PHP warning.

1. Load a timepoint list, instrument list and instrument page, then check web server log. The fix will avoid the warning.


src/Middleware/UserPageDecorationMiddleware.php
```diff
@@ -89,7 +89,10 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
         // I don't think anyone uses this. It's not really supported
         $tpl_data['css'] = $this->Config->getSetting('css');
 
-        $tpl_data['subtest'] = $request->getAttribute("pageclass")->page;
+        // Some page might call without pageclass
+        if (!is_null($request->getAttribute("pageclass"))) {
+            $tpl_data['subtest'] = $request->getAttribute("pageclass")->page;
+        }
```
driusan: This doesn't feel like it's solving at the right level.. why are some requests getting here without pageclass defined?

lingma: > why are some requests getting here without pageclass defined?
Because in the `module.class` or the timepoint_list, the logic first try to load the page without being going through the middleware layers, if that logic fails to get a page, then the system tries to load the correct page. It is too strange to see that kind of logic, so every page will being rendered two times because the first won't go through. I do not understand why initially we wanted to do that, so I just let it be like what I mentioned early.

driusan:
@lingma This needs to be rebased, it's including all sorts of commits that shouldn't be there.

lingma:
This PR will be closed, I might try to test whether I can remove those commits from rebase.

PR #5889, opened 2020-01-05
[Dashboard/Study Progression -- View scans per site] Restricting users to see statistics only from sites they have access to.

racostas:
### Brief summary of changes

In **dashboard-> Study Progression-> View scans per site** users were able to see statistics from sites they do not belong/have access

#### Testing instructions

Now each user should be able to see the **view scans per site** only for the sites it belong/have access

#### Notes

This is a fix for 22.0.0, major design changes as creating a permission dedicated to statistics still pending.

* Resolves #5847

### DIFF
--- a/modules/dashboard/ajax/get_scan_line_data.php
+++ b/modules/dashboard/ajax/get_scan_line_data.php
@@ -23,7 +23,21 @@ $client = new NDB_Client();
 $client->makeCommandLine();
 $client->initialize();
 
-$DB = Database::singleton();
+$DB            = Database::singleton();
+$currentUser   = \User::singleton();
+$site          = array();
+$list_of_sites = array();
+
+//TODO: Create a permission specific to statistics
+if ($currentUser->hasPermission('access_all_profiles')) {
+    $list_of_sites = \Utility::getSiteList();
+} else {
+    $site_id_arr = $currentUser->getCenterIDs();
+    foreach ($site_id_arr as $key => $val) {
+        $site[$key]          = &Site::singleton($val);
+        $list_of_sites[$val] = $site[$key]->getCenterName();
+    }
+}
 
 $scanData           = array();
 $scanStartDate      = $DB->pselectOne(
@@ -42,7 +56,6 @@ $scanEndDate        = $DB->pselectOne(
 );
 $scanData['labels']
     = createLineChartLabels($scanStartDate, $scanEndDate);
-$list_of_sites      = Utility::getAssociativeSiteList(true, false);
 foreach ($list_of_sites as $siteID => $siteName) {
     $scanData['datasets'][] = array(
         "name" => $siteName,


modules/dashboard/ajax/get_scan_line_data.php
```diff
@@ -23,7 +23,21 @@
...
+$DB            = Database::singleton();
+$currentUser   = \User::singleton();
+$site          = array();
+$list_of_sites = array();
+
+//TODO: Create a permission specific to statistics
+if ($currentUser->hasPermission('access_all_profiles')) {
+    $list_of_sites = \Utility::getSiteList();
+} else {
+    $site_id_arr = $currentUser->getCenterIDs();
```
PapillonMcGill: need to remove `DCC` center (ID = 1) to respect current functionality.
This code would be a good security improvement!

racostas: Hi @PapillonMcGill , thanks! 

Here what I propose in this case. We could rewrite **when the user has access to all profiles** as: `$list_of_sites = \Utility::getSiteList(true, false)` Already discussed on a LORIS meeting this functions will deprecate `Utility::getAssociativeSiteList(true, false)`.

**Two notes about this:** I see this already implemented on the master branch so before testing this code then we should bring this change to v22 or the place we decide to include this PR. Please note that the implementation I find in the master branch will break function in the case we call `Utility::getAssociativeSiteList(false, any_other).`

For **the case the user doesn't have the 'access_all_profiles'** permission the DCC should no be included at least explicitly specified during the user creation/edition (and hardy discouraged in the documentation if I remember well). In this case, that the user was explicitly linked to the DCC, my feeling is to allow him to see the statistics for the DCC as it was explicitly linked to it... but open to discussion.

Let me know your thoughts,
Thanks,
Rolando.

racostas: I'm aborting this PR in favour of the great major redesign proposed on #5896. Nice!

racostas:
I'm aborting this PR in favour of the great major redesign proposed on #5896. Nice!

modules/dashboard/ajax/get_scan_line_data.php
driusan: Won't refactoring to add this permission restriction still be relevant after #5896?

I'm planning on separating out the statistics/charts/XXX endpoints to a different PR to simplify the dashboard part, so once that's done you can just base on that to avoid conflicts

racostas: Hi @driusan, 

Definitely we should incorporate this logic. 

I'm not sure if as an external collaborator I'm able to use labels. I would tag this PR as "Blocked" or "Pending" in the case we don't incorporate this logic right away during the redesign. In this case I will be glad to resume work after but if its incorporated right away as part of the redesign I'm OK also.

Very nice work with the redesign!

Cheers,
Rolando (CHU Sainte-Justine)

driusan: You probably can't, so I added the "Blocked" tag for you

driusan:
@racostas I believe you can start working on these PRs again now that #5900 is in.

racostas:
Deprecated by #5917

PR #5890, opened 2020-01-06
Use module long name for Breadcrumb

### LINKED ISSUE #5856: (core) update Breadcrumbs to display long Module name
Breadcrumbs also should be updated to get and display this LongName for each module. 2 reasons: 
* to match the module names in the Menus (since #5839)
* currently some breadcrumb module names are wonky, e.g. capitalization in `Mri violations` http://demo.loris.ca/mri_violations/

Made possible by Dave's recent PR #5839 to add human-friendly Module names : 
> Modules must now implement a `getLongName` function to return the human friendly name for the module to be used by default in menus (but can also be used other places that a human readable version of the module name is needed in the future.)

_Originally posted by @christinerogers in https://github.com/aces/Loris/pull/5839#issuecomment-566076128_

After this is done (make a new ticket when ready):  
* All screenshots in documentation should probably be updated, if the long module name != short module name.

driusan:
This updates the default breadcrumb for a page to
use the module's getLongName for the first level of
the breadcrumb, rather than the hack where it replaces
'_' with space in the URL.

Fixes #5856.

### DIFF
--- a/modules/mri_violations/test/mri_violationsTest.php
+++ b/modules/mri_violations/test/mri_violationsTest.php
@@ -29,7 +29,9 @@ class MriViolationsTestIntegrationTest extends LorisIntegrationTest
      * UI elements and locations
      * breadcrumb - 'MRI Violated Scans'
      */
-    private $_loadingUI = array('MRI Violated Scans' => '#bc2 > a:nth-child(2) > div');
+    private $_loadingUI = array(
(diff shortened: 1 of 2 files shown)


driusan:
Blocked by #5839

driusan:
Looks like it's failing because the Mri Violations breadcrumb became more useful.. I'll wait until the blocking PR is in before rebasing/fixing it.

PR #5891, opened 2020-01-06
Update phan to the newest version.

### LINKED ISSUE #5651: Fix Phan errors introduced in PHP 7.4

driusan:
This updates phan to the latest version. Replaces #5762 (which is updating to an older version of phan that still has problems with PHP 7.4.)

Fixes #5651.

### DIFF
--- a/modules/instrument_list/php/instrument_list.class.inc
+++ b/modules/instrument_list/php/instrument_list.class.inc
@@ -199,6 +199,7 @@ class Instrument_List extends \NDB_Menu_Filter
             $x            = -1;
             $prevSubgroup = null;
 
+            $i = 0;
             foreach ($listOfInstruments as $instrument) {
                 // print the sub group header row
                 if ($instrument['Sub_group'] != $prevSubgroup) {
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -17,6 +17,7 @@ return [
 	"ignore_undeclared_variables_in_global_scope" => false,
 	"suppress_issue_types" => [
         "PhanTypeInvalidDimOffset",
+        "PhanTypePossiblyInvalidDimOffset",
 		"PhanUndeclaredMethod",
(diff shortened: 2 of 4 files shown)


modules/instrument_list/php/instrument_list.class.inc
```diff
@@ -199,6 +199,7 @@ class Instrument_List extends \NDB_Menu_Filter
             $x            = -1;
             $prevSubgroup = null;
 
+            $i = 0;
```
maltheism: I don't think `$i` is being used for anything in this function.

driusan: it's to prevent phan from flagging line 284 (`$i++`) as an error, because otherwise `$i` is initialized inside of an if statement and phan thinks it might be null

PR #5892, opened 2020-01-06
[new_profile/create_timepoint] code improvement

lingma:
## Brief summary of changes

This has been changed to a code improvement since I noticed that the problem has been fixed from frontend. The initial modification and intention from the PR #5854 is no longer useful but the code improvement could be kept.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,6 @@ clean:
 
 # Perform static analysis checks
 checkstatic: phpdev
-	npm run lint:shell
 	npm run lint:php
 	vendor/bin/phan
 	npm run lint:javascript


modules/new_profile/php/new_profile.class.inc
```diff
@@ -170,21 +170,27 @@ class New_Profile extends \NDB_Form
...
-        if ($num_sites > 1) {
-            foreach ($user_list_of_sites as $siteID) {
-                $center = $DB->pselectRow(
-                    "SELECT CenterID as ID, Name FROM psc WHERE CenterID =:cid",
-                    array('cid' => $siteID)
-                );
-                if (!is_null($center)) {
-                    $psc_labelOptions[$siteID] = $center['Name'];
-                }
+        $userSites = $user->getData('CenterIDs');
```
johnsaigle, line 173: I think it would be useful to include this code as a new function in the `User` class because there are probably other places in the code where we will need to do this.

lingma: OK, I put it in the `Utility`

php/libraries/Utility.class.inc
```diff
@@ -150,6 +151,39 @@ class Utility
...
     }
 
+    /**
+     * Returns a list of sites by site IDs from the database
+     *
+     * @param array $siteIDs a list of the study center IDs
+     *
+     * @return array an associative array("center ID" => "site name")
+     */
+    static function getSiteListByIDs(array $siteIDs): array
```
ridz1208, line 161: I would rename the function to `getSiteNameListFromIDs()` because you are not getting sites but site names.

php/libraries/Utility.class.inc
```diff
@@ -150,6 +151,39 @@ class Utility
...
+        $sites  = array();
+        $params = array();
+        $values = array();
+        foreach ($siteIDs as $key => $siteID) {
+            $para          = "cid$key";
+            $params[]      = ":$para";
+            $values[$para] = $siteID;
+        }
+
+        $centerLists = $DB->pselect(
```
ridz1208, line 175: I'm not a big fan of switching between `site` and `center` just choose one and use it for variable names

php/libraries/Utility.class.inc
lingma: `getSiteListByIDs` is to be consistent with `getSiteList`

php/libraries/Utility.class.inc
lingma: I ignore why initially that center vs site was named in that way, but I changed center to site.

modules/create_timepoint/php/create_timepoint.class.inc
```diff
@@ -223,17 +222,8 @@ class Create_Timepoint extends \NDB_Form
...
-                        $center = $DB->pselectRow(
-                            "SELECT CenterID as ID, Name FROM psc 
-                        WHERE CenterID =:cid",
-                            ['cid' => $siteID]
-                        );
-                        if (!is_null($center)) {
-                            $psc_labelOptions[$siteID] = $center['Name'];
-                        }
-                    }
+                    $psc_labelOptions = [null => ''] +
```
johnsaigle, line 225: I don't think you can use `+` to glue arrays together.

lingma: why not?

johnsaigle: as far as I know PHP doesn't work like that.

```php
php > $arr1 = [1,2,3];
php > $arr2 = [4,5,6];
php > print_r($arr1 + $arr2);
Array
(
    [0] => 1
    [1] => 2
    [2] => 3
)
```

lingma: Oh, this is different. You are trying one dimensional array.

php/libraries/Utility.class.inc
ridz1208: `getsitelist` is incorrectly named because we were less rigorous before, since it returns names not sites I think it should be properly named

php/libraries/Utility.class.inc
```diff
@@ -150,6 +151,39 @@ class Utility
...
+        $sites = $DB->pselect(
+            "SELECT CenterID as ID, Name FROM psc WHERE CenterID in (" .
+            implode(', ', $params) . ")",
+            $values
+        );
+
+        $list = array();
+        foreach ($sites as $site) {
+            $list[$site['ID']] = $site['Name'];
+        }
```
ridz1208, line 183: this should be using `pselectColWithIndexKey()` from the database instead

php/libraries/Utility.class.inc
ridz1208, line 161: @driusan shouldnt a function like this exist in the site class rather than utility ?

lingma:
Sorry, the wrong branch was pushed. please revert

ridz1208:
I can't revert you need to repush the correct branch I think

lingma:
oh, the change was not merged, that is good.

PR #5894, opened 2020-01-07
Replace zend-diactoros with laminas-diactoros

driusan:
Fixes the error "Package zendframework/zend-diactoros is
abandoned, you should avoid using it. Use laminas/laminas-diactoros instead."
while running composer install on LORIS.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         "psr/http-message": "~1.0",
         "psr/http-server-handler" : "*",
         "psr/http-server-middleware" : "*",
-        "zendframework/zend-diactoros": "~1.8.4",
+        "laminas/laminas-diactoros" : "~1.8.4",
         "ext-json": "*",
         "bjeavons/zxcvbn-php": "^0.4"
     },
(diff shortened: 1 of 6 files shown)


johnsaigle:
The error doesn't occur for me when running `make checkstatic` on `aces/master`. Can you provide detailed steps to replicate?

I did:
```bash
git fetch aces
git checkout aces/master
make dev
make checkstatic
```

and there were no issues.

driusan:
@johnsaigle ?

driusan:
I don't have detailed steps to replicate, I just started getting the error when I followed those same steps (at the `make dev` part) this morning.

driusan:
Maybe you can replicate by deleting your `~/.composer/cache` directory?

johnsaigle:
```bash
git fetch aces
git checkout aces/master
rm -rf ~/composer/cache/
make dev
```

No issue appeared.

driusan:
I'm not getting the error anymore either, but I still think we should move forward with this since, as you say, it's definitely abandoned (or at least renamed.)

PR #5896, opened 2020-01-07
Refactor dashboard to move widgets into modules.

driusan:
This updates the dashboard in order to move responsibility for dashboard widgets into the modules themselves. It adds the concept of a "widget" to LORIS as a way to provide loosely coupled cross-module dependencies. Modules may now register "widgets" of different types to be used by other modules. As a proof-of-concept, this architecture can be the basis of the upcoming candidates dashboard.

There are 2 types of widgets defined in this PR: dashboard widgets, which add a panel to the dashboard, and usertasks widgets, which add a task to the "My Tasks" panel on the dashboard.

All existing widgets are moved to the most appropriate module that currently exists. The dashboard only maintains ownership of the "Welcome" widget and the "My Tasks" widget itself.

PR #5897, opened 2020-01-07
[Core] Add renderTemplate helper to Module class

driusan:
This adds a "renderTemplate" function to the module class, in order
to render a template from a module's template directory, given its \Module
descriptor.

This is a helper extracted from #5896 which is mostly self-contained
enough to stand on its own.


driusan:
This isn't really big enough to make a significant change in the size of the other PR and no one seems interested in reviewing it on its own, so I'll just close this since it's already incorporated in the other PR.

PR #5898, opened 2020-01-07
Remove stale bot exemptions because we're not using its auto-close feature

johnsaigle:
Stale bot was configured incorrectly. My bad.


driusan:
I could have sworn we had Bugfix but not Bug? But since you're changing it anyways, can you add " Priority: High" to the list of exemptions?

driusan:
Or perhaps we shouldn't have any exemptions at all, since it's not set to close?

johnsaigle:
Yeah I agree with removing them. If we do enable auto-closing, then we can add these exemptions back.

PR #5899, opened 2020-01-07
[Core] Add getLastLogin helper to user class

driusan:
This moves the query to get the last user login from the
dashboard code to the User class, where it more accurately
belongs.

(This was extracted from #5896 where it was necessary because
multiple modules use the last login time to determine if something
is new.)

### DIFF
--- a/modules/dashboard/php/dashboard.class.inc
+++ b/modules/dashboard/php/dashboard.class.inc
@@ -40,7 +40,7 @@ class Dashboard extends \NDB_Form
         $site   = $user->getSiteNames();
 
         $userID     = $user->getUsername();
-        $last_login = $user->getLastLogin();
+        $last_login = $user->getLastLogin($DB);
 
         $siteID = $user->getCenterIDs();
         $this->tpl_data['user_site'] = $siteID;
--- a/modules/dashboard/php/dashboard.class.inc
+++ b/modules/dashboard/php/dashboard.class.inc
@@ -40,14 +40,15 @@ class Dashboard extends \NDB_Form
         $site   = $user->getSiteNames();
 
         $userID     = $user->getUsername();
-        $last_login = $user->getLastLogin($DB)->format('Y-m-d H:i:s');
+        $last_login = $user->getLastLogin($DB);
 
         $siteID = $user->getCenterIDs();
         $this->tpl_data['user_site'] = $siteID;
 
         // Welcome panel
         $this->tpl_data['username'] = $user->getFullname();
-        if ($last_login) {
+        if ($last_login !== null) {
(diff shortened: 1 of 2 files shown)


php/libraries/User.class.inc
```diff
@@ -532,5 +532,27 @@ class User extends UserPermissions
...
+
+
+    /**
+     * Return the last login of this user to the given database.
+     *
+     * @param \Database $DB The database connection to check
+     *
+     * @return string The date in YYYY-MM-DD format.
+     */
+    function getLastLogin(\Database $DB)
```
johnsaigle: I think this is going to generate a phan error. pselectOne has a `mixed` return type (array, string, or false).

It would be good to add a proper return type here. Bonus points if it's a DateTime object

driusan: I think changing it to a DateTime object would break the code which is currently using it, but I agree that that should be a future change

driusan: it was pretty easy to change here and just reformat in the one place that uses it, I'll deal with the fallout of that on #5896 when I eventually rebase it onto master.

modules/dashboard/php/dashboard.class.inc
```diff
@@ -40,15 +40,7 @@ class Dashboard extends \NDB_Form
...
-        $last_login = $DB->pselectOne(
-            "SELECT MAX(Login_timestamp)
-             FROM user_login_history
-             WHERE Login_timestamp < (SELECT MAX(Login_timestamp)
-                FROM user_login_history
-                WHERE userID=:UserID AND Success='Y')
-             AND userID=:UserID AND Success='Y'",
-            array('UserID' => $userID)
-        );
+        $last_login = $user->getLastLogin($DB)->format('Y-m-d H:i:s');
```
johnsaigle: Will this cause an error to the effect of "Can't use function 'format' on null"?

php/libraries/User.class.inc
```diff
@@ -532,5 +532,32 @@ class User extends UserPermissions
             (string) $this->userInfo['Password_hash']
         );
     }
+
```
johnsaigle: Looks good pending Travis. I have a suspicion that one of the static analysis tools won't like the two new lines here but maybe it doesn't matter.

modules/dashboard/php/dashboard.class.inc
driusan: Yes, good catch. I added a hack to handle null and ensure that it's in the same format as the DB was returning for this PR only if it's set, but I'll fix it properly in the modularized widgets PR because work on this one will end up getting lost in the end anyways

driusan:
One down, 4 to go..

PR #5900, opened 2020-01-07
[Dashboard/Statistics] Move dashboard ajax endpoints into the statistics module

driusan:
This moves the ajax calls from the dashboard module into the more appropriate
statistics module for the graph endpoints. It also converts some inexplicable
"post" requests into "get" requests but otherwise behaviour should remain
unchanged.

This was extracted from #5896 and is sent as a separate PR in order to simplify
that PR and make it easier to rebase data/ dashboard permission related PRs
which would otherwise conflict.

1. Test all the charts on the dashboard. They should load as previously and contain the same data.


maltheism:
@driusan I'm wondering if I can make a PR after this is merged for moving the php files into a directory named endpoints. I think it helps the structure somewhat.

driusan:
I don't have any plans to touch these endpoints after this is merged, so I don't foresee any conflicts and a PR that does that would need to be reviewed/merged on its own merits (I don't really see the point of doing it partially and for just the stats module since that's not a convention we use anywhere else and just about everything is an endpoint, but if other people like it I wouldn't block it.)

modules/statistics/php/charts.class.inc
```diff
@@ -0,0 +1,343 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

driusan:
This is also blocking #5889

modules/statistics/php/charts.class.inc
```diff
@@ -0,0 +1,343 @@
...
+     * Helper to generate labels for every month between startDate and endDate.
+     *
+     * @param string $startDate The start date for the labels in YYYY-MM-DD format.
+     * @param string $endDate   The end date for the labels in YYYY-MM-DD format.
+     *
+     * @return array
+     */
+    private function _createSiteLineChartLabels(string $startDate, string $endDate)
+    {
+        $startDateYear  = substr($startDate, 0, 4);
```
johnsaigle: I know you didn't write this code but do you think we could refactor this part?

My preference is to use DateTimes to do this but if you don't want to do that adding some class constants and removing the magic numbers would be helpful.

modules/statistics/php/charts.class.inc
```diff
@@ -0,0 +1,343 @@
...
+            ORDER BY YEAR(pf.Value), MONTH(pf.Value), s.CenterID",
+            array()
+        );
+
+        // Create the labels
+        $labels = [];
+        foreach ($data as $row) {
+            $labels[$row['datelabel']] = true;
+        }
+        $scanData['labels'] = array_keys($labels);
```
johnsaigle: I'm not really clear on the use of $labels here and why it needs to be set to true. It doesn't seem to be used elsewhere.

It looks like this part could be simplified to

```php
foreach ($data as $row) {
    $scanData['labels'][] = $row['datalabel'];
}
```

or something even simpler.


Not a big deal just found it hard to parse.

johnsaigle:
> I don't really see the point of doing it partially and for just the stats module since that's not a convention we use anywhere else

@driusan It seems to be an emerging convention in PRs by @maltheism and @xlecours. It might be a good meeting agenda item to discuss the folder structure of endpoints to make sure everyone is in the know.

modules/statistics/php/charts.class.inc
driusan: It doesn't need to be set to true, it just needs to be set to something so that it's an associative array in order to remove duplicate date labels from across sites. Your change would make it a numeric array and multiple sites with the same date label would appear multiple times.

johnsaigle: Fair enough. It's a small thing so we don't need to devote more brainpower to it

modules/statistics/php/charts.class.inc
driusan: Use DateTimes to do what? This is a helper to generate labels for every month between 2 dates, I don't see how DateTime objects would help/be used.

johnsaigle: This should do the same thing and is a lot easier to read imo.

```php
// Create example dates
$startDate = new DateTime(); // today
$endDate = $start->add(date_interval_create_from_date_string('6 months'));
$month = date_interval_create_from_date_string('1 month');
$betweenDate = clone $startDate;

while ($betweenDate < $endDate) {
    // DateTime::add modifies (increments) the $betweenDate object
    $labels[] = $betweenDate->add($month)->format('m-Y');
}
```

driusan: I had to make a few bug fixes to make sure the results were the same, but I rewrote it to use datetime objects.

modules/statistics/php/charts.class.inc
```diff
@@ -0,0 +1,349 @@
...
+    private function _createSiteLineChartLabels(\DateTime $startDate, \DateTime $endDate) : array
+    {
+        $month = date_interval_create_from_date_string('1 month');
+        // Since we're only concerned with months, act as if $startDate
+        // is always the first of the month so that the last month doesn't
+        // get truncated.
+        $betweenDate = new \DateTime($startDate->format('Y-m-01'));
+
+        while ($betweenDate <= $endDate) {
+            $labels[] = $betweenDate->format('n-Y');
```
johnsaigle:
```suggestion
            $labels[] = $betweenDate->format('m-Y');
```

modules/statistics/php/charts.class.inc
```diff
@@ -0,0 +1,349 @@
...
+            $betweenDate = $betweenDate->add($month);
```
johnsaigle:
```suggestion
            $betweenDate->add($month);
```

DateTime::add() modifies the object directly.

modules/statistics/php/charts.class.inc
driusan: This is wrong and produces different output. The current behaviour doesn't have a leading 0.

modules/statistics/php/charts.class.inc
driusan: I think this is less clear than the way I wrote it even if it's technically true.

modules/statistics/php/charts.class.inc
johnsaigle: Got it. I assumed it was a typo, my bad.

modules/statistics/php/charts.class.inc
johnsaigle: Yeah I see what you're saying. I think it could also mislead people about the behaviour of the function though if they're only seeing it in this context.
 Maybe we can add what I wrote above as a comment in the code?

modules/statistics/php/charts.class.inc
```diff
@@ -0,0 +1,358 @@
...
+                ),
+            );
+        }
+        return new \LORIS\Http\Response\JsonResponse($recruitmentData);
+    }
+
+    /**
+     * Helper to generate labels for every month between startDate and endDate.
+     *
+     * @param \DateTime $startDate The start date for the labels in
```
johnsaigle:
```suggestion
     * @param \DateTime $startDate The start date for the labels.
```

modules/statistics/php/charts.class.inc
```diff
@@ -0,0 +1,358 @@
...
+     *                             YYYY-MM-DD format.
```
johnsaigle:
```suggestion
```

modules/statistics/php/charts.class.inc
```diff
@@ -0,0 +1,358 @@
...
+     * @param \DateTime $endDate   The end date for the labels in
```
johnsaigle:
```suggestion
     * @param \DateTime $endDate   The end date for the labels.
```

modules/statistics/php/charts.class.inc
driusan: they're now DateTimeImmutable and no one can be misled.

PR #5904, opened 2020-01-09
[NDB_Config] Make field list explicit and remove unnecessary catch

driusan:
This makes the field list in NDB_Config->getProjectSettings explicit
so that it crashes hard if a patch is missed rather than continuing
on and providing unrelated warnings later. It removes the try/catch
for the same reason (and because I couldn't figure out any reason it
was there in the first place. It looks like it was added in a commit to appease
phan, but phan is passing without it.)

### DIFF
--- a/php/libraries/NDB_Config.class.inc
+++ b/php/libraries/NDB_Config.class.inc
@@ -440,16 +440,15 @@ class NDB_Config
      */
     function getProjectSettings(int $ProjectID): ?array
     {
-        try {
-            $info = \NDB_Factory::singleton()
-                ->database()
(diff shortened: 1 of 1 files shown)


driusan:
@johnsaigle the diff says you added the try/catch, can you shed any light on why it was there?

johnsaigle:
I looked at the git blame and the PR I made moved the line for formatting reasons. It seems to have been added in LORIS 16. https://github.com/aces/Loris/commit/c253d722e676a7a4d8265d69e6bfd0dbaf55acc8

PR #5905, opened 2020-01-09
[Dev] Update phpmd to the latest release

driusan:
This updates phpmd to the latest release. This is
primarily to fix a number of warnings that it outputs
on PHP 7.4 when running `make checkstatic`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "phpunit/phpunit" : "7.0.0",
         "facebook/webdriver" : "dev-master",
         "phan/phan": ">=2.3.0",
-        "phpmd/phpmd": "~2.7"
+        "phpmd/phpmd": "~2.8"
     },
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -19,7 +19,7 @@
         "phpunit/phpunit" : "7.0.0",
         "facebook/webdriver" : "dev-master",
         "phan/phan": ">=2.3.0",
-        "phpmd/phpmd": "~2.7"
+        "phpmd/phpmd": "~2.8"
```
johnsaigle:
```suggestion
        "phpmd/phpmd": "2.8.1"
```
Do you have any feelings about being more strict with this?

driusan: Why? It seems unnecessary, it would just make "composer update" less likely to get bug fixes. Exact versions if what the lock file is for.

PR #5907, opened 2020-01-09
[Tools] Add option to specify tables to dump for RB on command line

driusan:
This adds the ability to only dump certain tables by specifying them
as arguments on the command line to DB_dump_table_data.php.

### DIFF
--- a/tools/exporters/DB_dump_table_data.php
+++ b/tools/exporters/DB_dump_table_data.php
@@ -28,13 +28,19 @@ require_once __DIR__ . '/../generic_includes.php';
 $config = NDB_Config::singleton();
 $databaseInfo = $config->getSetting('database');
 
-// Get all tables in the database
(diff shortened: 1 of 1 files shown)


ridz1208:
I'm okai with that, but we might be running the risk of people accidentally forgetting to submit some changes that should submitted.

the opposite is true though, without it, we are running the risk of people submitting unnecessary changes... so pass!

kongtiaowang:
I tested it, If I gave a correct table name, it worked well.
If I gave a wrong table name, It showed that "Couldn't find table" --- But it will create a backup file with the wrong table name under "/raisinbread/RB_files".

driusan:
thanks for testing @kongtiaowang, I'll try and add a little more error checking for that

driusan:
@kongtiaowang I added some error checking to ensure tables exist, can you test again?

PR #5908, opened 2020-01-09
[Core] Candidate createNew infinite loop

xlecours:
## Brief summary of changes
There is a retry mechanism in Candidate::createNew "in the case of a race condition where another CandID with the same value has been inserted into the DB before this one."

An infinite loop occurs when the insert statement fails for an other reason.

This adds a limit to the number of insert attempts 

1. Set `userEDC` to false
2. Create a candidate with an invalid value for EDC (ex: an empty string)
3. A LorisException should be thrown after 3 attempts

### DIFF
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -283,6 +283,7 @@ class Candidate
         $generator = new \CandIDGenerator();
         // declare the variable to avoid static analysis warnings
         $invalidID = false;
+        $attemptcount = 0;
         do {
             // Reset on each iteration
             $invalidID = false;
@@ -296,6 +297,12 @@ class Candidate
                 // with the same value has been inserted into the DB before this
                 // one.
                 $invalidID = true;
+                $attemptcount++;
+                if ($attemptcount > 2) {
+                    throw new LorisException(
+                        $e->getMessage()
+                    );
+                }
             }
         } while ($invalidID);
 


php/libraries/Candidate.class.inc
```diff
@@ -296,6 +297,12 @@ class Candidate
                 // with the same value has been inserted into the DB before this
                 // one.
                 $invalidID = true;
+                $attemptcount++;
+                if ($attemptcount > 2) {
+                    throw new LorisException(
```
driusan: why not just re-throw $e?

johnsaigle:
> 1. Set userEDC to false
2. Create a candidate with an invalid value for EDC (ex: an empty string)

Shouldn't step 1 disable the input for step 2?

Also it sounds like this is an issue with input validation in the first place. If there's an invalid value then nothing should get inserted so there should be no race condition.

Also, @xlecours and @maltheism can you figure out the relationship between this PR, #5922, and #5758? It seems like they're all caused by the same issue of input validation but the solution is splintering off into many different PRs. 

I think of these 3 PRs only one should go through that properly fixes this issue (if my understanding is correct and they are all caused by the same thing).

maltheism:
@johnsaigle I agree this PR is similar (the infinite loop bug) but the issue causing it is slightly different here than in the PR I issued. I think my PR handles the empty string for DoB correctly and I went off comments from @driusan when I discovered it for empty strings. I'm not sure whats best for the loop in general.

johnsaigle:
Did this get fixed @xlecours ?

xlecours:
I don't think so.

johnsaigle:
Why did you close it?

php/libraries/Candidate.class.inc
```diff
@@ -296,6 +297,12 @@ class Candidate
...
+                        $e->getMessage()
+                    );
```
ridz1208, line 304:
```suggestion
                    throw $e
```

PR #5909, opened 2020-01-09
[dataquery] Update AJAX to fetch for GetDoc

maltheism:
## Brief summary of changes

I've been assigned to update the DQT codebase. I'm currently on the task of updating the $.ajax to be using fetch instead. I'm also moving the endpoint(s) code to an endpoints directory and similar to how the DataFramework is structured.

**Note:** This PR only removes the GetDoc.php use of $.ajax and to keep the review process simple.

1. Test the DQT module by visiting it.
2. Save a query as well.


modules/dataquery/php/module.class.inc
```diff
@@ -3,58 +3,74 @@
  * This serves as a hint to LORIS that this module is a real module.
  * It does nothing but implement the module class in the module's namespace.
  *
- * PHP Version 5
+ * PHP Version 7
```
driusan: I don't understand why this PR is changing the Module class. Endpoints shouldn't require any changes to the module class, they get auto-delegated to the class with the same name as the page name requested.

maltheism: @driusan I thought it was needed to have the handle ServerRequestInterface inside the module.class.inc otherwise the classes are not delegated to the correct classes in the php/endpoints directory. @xlecours

maltheism:
Blocked tag added, going to change class name to CreateDocuments

modules/dataquery/jsx/react.app.js
```diff
@@ -181,17 +181,19 @@ class DataQueryApp extends Component {
...
-            dataType: 'json'
-          }).then((value) => {
-          let queries = this.state.savedQueries;
-
-          queries[value._id] = value;
-          this.setState({savedQueries: queries});
-        });
+          fetch(
+            window.location.origin
+            + '/dataquery/Document?DocID='
```
driusan: by convention our URLs are usually lowercase, and I think it would be a lot cleaner if the URL was `/dataquery/documents/xxxx` and not `/dataquery/document?DocID=xxxx`

maltheism: @driusan I tried changing the url to `dataquery/documents/xxx` but then `$values = $request->getQueryParams();` doesn't seem to work for getting the DocID from the GET request. I did change it to lowercase and to be `documents` instead of `document` with latest commit. @xlecours

modules/dataquery/php/documents.class.inc
```diff
@@ -0,0 +1,88 @@
...
+     * @return ResponseInterface The outgoing PSR7 response
+     */
+    public function handle(ServerRequestInterface $request) : ResponseInterface
+    {
+        // Ensure GET or POST request.
+        switch ($request->getMethod()) {
+        case 'GET':
+            return $this->_handleGET($request);
+        default:
+            return new \LORIS\Http\Response\MethodNotAllowed(
```
johnsaigle:
```suggestion
            return new \LORIS\Http\Response\JSON\MethodNotAllowed(
```

modules/dataquery/php/documents.class.inc
```diff
@@ -0,0 +1,88 @@
...
+                $this->allowedMethods()
```
johnsaigle: It doesn't look like this method exists.

modules/dataquery/php/documents.class.inc
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * This function will return a json response.
+     *
+     * @param ServerRequestInterface $request The incoming PSR7 request
+     *
+     * @return ResponseInterface The outgoing PSR7 response
+     */
+    public function handle(ServerRequestInterface $request) : ResponseInterface
+    {
+        // Ensure GET or POST request.
```
johnsaigle:
```suggestion
        // Ensure GET request.
```

modules/dataquery/php/documents.class.inc
```diff
@@ -0,0 +1,88 @@
...
+     * Returns true if the user has permission to access
+     * the DQT module
+     *
+     * @param \User $user The user whose access is being checked
+     *
+     * @return bool true if user has permission
+     */
+    function _hasAccess(\User $user) : bool
+    {
+        return ($user->hasPermission('dataquery_view'));
```
johnsaigle:
```suggestion
        return $user->hasPermission('dataquery_view');
```

maltheism:
Thanks @johnsaigle for the review!

modules/dataquery/php/documents.class.inc
maltheism: I'll check this out tomorrow.

modules/dataquery/jsx/react.app.js
driusan: This isn't resolved. It should still be documents/xxx and not ?DocID=xxx, but after changing that the xxx isn't a query param, so it needs to be extracted from $request->getUri().

maltheism: @driusan I'm not really sure what you mean by documents/xxx. Do you mean just having the url like /documents/DocID=xxx. In any case let me know if the latest commit which is another way of adding params to a url isn't good either.

driusan: No, I mean a document named "foo" should be accessed at `/documents/foo`. It isn't a parameter, it's the document name.

maltheism: Thanks @driusan for the clarification, that makes sense. I wasn't thinking like that lol.

modules/dataquery/jsx/react.app.js
```diff
@@ -180,18 +180,22 @@ class DataQueryApp extends Component {
...
-            },
-            dataType: 'json'
-          }).then((value) => {
-          let queries = this.state.savedQueries;
-
-          queries[value._id] = value;
-          this.setState({savedQueries: queries});
-        });
+          fetch(
+            `${window.location.origin}/dataquery/documents/${document}`,
```
HenriRabalais: is there a reason to be using this instead of `loris.BaseURL`?

modules/dataquery/jsx/react.app.js
```diff
@@ -259,21 +263,31 @@ class DataQueryApp extends Component {
           queryIDs.User.push(id);
         }
       }
-      $.get(loris.BaseURL + '/AjaxHelper.php?Module=dataquery&script=GetDoc.php&DocID=' + id,
-        (value) => {
-          let queries = this.state.savedQueries;
+      fetch(
+        window.location.origin
```
HenriRabalais: same here

modules/dataquery/jsx/react.app.js
maltheism: @HenriRabalais the use of the dollar sign here isn't jQuery. See here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

modules/dataquery/jsx/react.app.js
maltheism: @HenriRabalais the use of the dollar sign here isn't jQuery. See here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

modules/dataquery/jsx/react.app.js
driusan: he's not asking about the dollar sign, he's asking why it isn't using "`loris.baseURL +  '/dataquery/etc`"

driusan: How would window.location.origin work if the instance isn't hosted on the root of the domain?

maltheism: @driusan @HenriRabalais  `console.log(window.location.origin);` when visiting a site with a subdomain such as docs.google.com returns `https://docs.google.com` on chrome, safari, and firefox.

maltheism:

PR #5766, opened 2019-11-25
Add PHPCS to tools folder

johnsaigle:
**Please review this script with `?w=1` to hide whitespace changes**

## Brief summary of changes

Adds PHPCS standards to the tools script.

Tools scripts are now also checked for syntax errors by `php -l`

Excludes the subfolders of tools for now since this PR will already be pretty big.

I added some PHPCS suppressions to some parts of the code that would've been difficult to refactor properly. I think they should eventually be removed but I didn't want to add more complexity to this PR.

PR #5913, opened 2020-01-10
[Tools] Modules helper script

driusan:
This adds a tool to help maintain the new modules table. It can either add new modules it finds in either project/modules or loris/modules or delete modules in the table which can't be instantiated based on the options provided on the command line.


PapillonMcGill:
@driusan is the `-n` flag common in LORIS? I though that the --confirm flag was to be used to actually do change from a sript? 
If it is to set a new way to do it, could we use `--dryrun` which would be more explict than `-n` albeit longer to type.

driusan:
`-n` is common in git and lots of unix tools. I can add a --dryrun alias.

tools/modules.php
```diff
@@ -0,0 +1,85 @@
...
+ * modules table that can not be instantiated with Module::factory from
+ * the modules table.
+ *
+ * If the -n flag is provided, it will not actually add/delete from the
+ * table, but only tell you what it would otherwise do.
+ *
+ * PHP Version 7
+ */
+
+require_once __DIR__ . "/../vendor/autoload.php";
```
ridz1208: this is already in generic includes

driusan: all of our other php scripts seem to include both

ridz1208: i've been trying to clean them up before placing them in the subdirectories. but yeah its redundant

driusan: it's removed

ridz1208:
@driusan a few comments, none critical but 1 and 3 highly recommended

1. this should go in the `tools/data_integrity/` directory
2. I kind of agree with @PapillonMcGill on the fact that our "default" in loris is to specify a flag to apply actions rather then a flag for dry-run. I won't however enforce it here because running the script without any arguments would not have any impact either way... so risk is low on accidentally running it in "confirm" mode.
3. A good practice would be to print the "man" when the script is run with -h or with the improper number of arguments. we tend to do that in other scripts.
4. even in `-n` mode the prints read `adding/deleting` instead of `will be added/will be deleted`

driusan:
4 is intentional, because that's standard behaviour for any software that has a `-n` flag. I'll work on 3, but for point 1 I'm not sure that "data_integrity" is the appropriate directory for it .. I looked at the directories that we had but didn't think it belonged in any of them.

ridz1208:
1. Okai we ll see if eventually we can group it into another directory.

For 3, you might find examples in the scripts already in directories

driusan:
@ridz1208 I added help

tools/modules.php
```diff
@@ -0,0 +1,113 @@
...
+ * have a valid \Module descriptor to the modules table.
+ *
+ * If the --remove parameter is provided, it will remove anything in the
+ * modules table that can not be instantiated with Module::factory from
+ * the modules table.
+ *
+ * If the -n flag is provided, it will not actually add/delete from the
+ * table, but only tell you what it would otherwise do.
+ *
+ * PHP Version 7
```
johnsaigle: Can you add the usual author, license etc here? 

It'll be needed soon anyway: #5766

tools/modules.php
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env php
+<?php
+/**
+ * The modules.php script a database and automatically adds any new modules from
```
johnsaigle: It might be good to rename the script to be a verb like `update_modules_table` or whatever.

tools/modules.php
driusan: done

tools/modules.php
```diff
@@ -0,0 +1,120 @@
...
+    usage();
+    fwrite(STDERR, "\nMust specify --add or --remove flag\n");
+    exit(1);
+
+}
+
+$currentModules = $DB->pselectCol(
+    "SELECT Name FROM modules",
+    []
+);
```
ridz1208: shouldn't this query exist somewhere like in the module class or something?

moduleController ???

(out of the scope of this PR )

driusan: the module class has a static getActiveModules, but this needs the query without the Active='Y' part of the the query. I don't think there's anywhere else in LORIS that we would ever want to get the inactive modules.

tools/modules.php
driusan: I went with "manage_modules.php" because if there's ever more to do than update the table to install a module the script should take care of that too.

tools/manage_modules.php
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env php
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

kongtiaowang:
I run "--add", it inserted the module name in the Database automatically.
But for "manage_modules.php --remove" It doesn't work for me.
I created a new module,
then I run the command with " --remove ". Nothing happens, the table name is still in the modules table. 
After I removed the new Module, I run " --remove" again, 
PHP Fatal error:  Uncaught LorisNoSuchModuleException: No such module: newmoulde in /var/www/loris/php/libraries/Module.class.inc:90
Stack trace:
#0 /var/www/loris/tools/manage_modules.php(55): Module::factory('newmoulde')
#1 {main}
  thrown in /var/www/loris/php/libraries/Module.class.inc on line 90

johnsaigle:
@kongtiaowang Typo? `newmoulde` --> `newmodule`?

driusan:
@kongtiaowang `--add` (and `--remove`) don't take module name arguments, they synchronize the table with what's on the filesystem. `--add` adds all valid modules that aren't in the table and `--remove` removes things that are in the table which aren't valid modules on the filesystem.

driusan:
@kongtiaowang if you're still having problems can you request changes so I don't accidentally merge this if it needs work?

kongtiaowang:
@driusan I didn't take module name as arguments. It can't remove a module.
Screen Shot 2020-01-16 at 11 14 10 AM

driusan:
@kongtiaowang sorry I misread your comment, thanks for the clarification.. I'll look into it.

driusan:
@kongtiaowang It looks like I was handling the case where there's no valid module descriptor in the directory but not the case where the directory doesn't exist (these two cases should probably be merged now that all modules need a module descriptor..). It should be fixed now.

kongtiaowang:
@driusan This PR is ready to go.

PR #5916, opened 2020-01-10
Phan mismatched return

driusan:
This makes a few stylistic changes and fixes some comments in order
to make it easier for Phan to reason about the code and as a result
fixes the PhanMismatchReturn error type.


driusan:
Blocked by #5915 (which is a slightly more substantial change that was detected by this rule.)

PR #5917, opened 2020-01-10
[Dashboard] Restricting users to see statistics only from sites they have access to.

racostas:
### Brief summary of changes
Address #5889, #5861 and #5847 after the redesign proposed in #5896 

#### Testing instructions

Now each user should be able to see the dashboard statistics only for the sites it belong/have access to.

* Resolves #5889,  #5861, #5847

### DIFF
--- a/modules/statistics/php/charts.class.inc
+++ b/modules/statistics/php/charts.class.inc
@@ -286,22 +286,27 @@ class Charts extends \NDB_Page
             $list_of_sites = $currentUser->getSiteList(true, false);
         }
 
-        $listOfSitesString = implode(",", array_keys($list_of_sites));
-
-        $recruitmentData      = array();
+        $params            = array();
+        $centerIDs         = array();
+        $list_of_sites_ids = array_keys($list_of_sites);
+        foreach ($list_of_sites_ids as $key => $siteID) {
+            $params[]            = ":id$key";
+            $centerIDs["id$key"] = $siteID;
+        }
         $recruitmentStartDate = $DB->pselectOne(
             "SELECT MIN(Date_registered)
              FROM candidate
-             WHERE RegistrationCenterID IN (" . $listOfSitesString . ")",
-            array()
+             WHERE RegistrationCenterID IN (" . implode(',', $params) . ")",
+            $centerIDs
         );
         $recruitmentEndDate   = $DB->pselectOne(
             "SELECT MAX(Date_registered)
              FROM candidate
-             WHERE RegistrationCenterID IN (" . $listOfSitesString . ")",
-            array()
+             WHERE RegistrationCenterID IN (" . implode(',', $params) . ")",
+            $centerIDs
         );
 
+        $recruitmentData = array();
         if ($recruitmentStartDate !== null
             && $recruitmentEndDate !== null
         ) {
--- a/modules/statistics/php/charts.class.inc
+++ b/modules/statistics/php/charts.class.inc
@@ -105,10 +105,16 @@ class Charts extends \NDB_Page
      */
     private function _handleSitePieData()
     {
-        $DB = \NDB_Factory::singleton()->database();
+        $DB          = \NDB_Factory::singleton()->database();
+        $currentUser = \NDB_Factory::singleton()->user();
 
         $recruitmentBySiteData = array();
-        $list_of_sites         = \Utility::getSiteList(true, false);
+
+        if ($currentUser->hasPermission('access_all_profiles')) {
+            $list_of_sites = \Utility::getSiteList(true, false);
+        } else {
+            $list_of_sites = $currentUser->getSiteList(true, false);
+        }
 
         foreach ($list_of_sites as $siteID => $siteName) {
             $totalRecruitment = $DB->pselectOne(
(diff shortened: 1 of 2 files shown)


PapillonMcGill:
Hello @racostas, well done but wouldn't it be better to use the new dataprovisioner than have the same function (getSiteList) in both Utility and User class? @driusan What do you think?

racostas:
Hello @PapillonMcGill, 

Anything that improves architecture is more than welcome. I'm not too much familiarized with LORIS architecture and best practices yet. I'm not sure to fully understand your proposal. If you could please delve a little more?

Thanks,
Rolando (CHU Sainte-Justine)

PapillonMcGill:
Hello, @racostas 

@ridz1208 put in place a dataframe work to access and filter data from the DB according to users' permision. In Candidate class, you could call the Utility/getSiteList() and then use the new dataprovisionner to filter them with users' permission. 

It should become the new standard soon.  @ridz1208 is to give a presentation in a Loris meeting.
@ridz1208 Do you want to comment?

ridz1208:
@racostas @PapillonMcGill 

My only comment is you should replace all instances of `@ridz1208` by `@driusan` in @PapillonMcGill 's comment above

modules/statistics/php/charts.class.inc
```diff
@@ -257,18 +277,36 @@ class Charts extends \NDB_Page
...
+        } else {
+            $list_of_sites = $currentUser->getSiteList(true, false);
+        }
 
-        $recruitmentData      = array();
+        $params            = array();
+        $centerIDs         = array();
+        $list_of_sites_ids = array_keys($list_of_sites);
+        foreach ($list_of_sites_ids as $key => $siteID) {
+            $params[]            = ":id$key";
```
PapillonMcGill:
```suggestion
            $params[]            = "$key";
```
This seam like a typo or bad copy/paste.
Same for next line.

racostas: Hello @PapillonMcGill ,

There is not typo or copy/paste error in the line 293. Please let me explain you what this code is doing: This line is doing a variable interpolation so each of the elements(**params**) in the `$params[]` array will take the shape '`:idkey`' for example: '`:id0`', '`:id1`', etc. As this will be passed to the `IN` clause in the `pselectOne` function (**line 305**) makes no sense to pass '0', '1', etc which will be the case if we go as you proposed changing `$params[] = ":id$key"` for: `$params[] = "$key"`.

I hope this helps you to understand. Please don't hesitate to ask me if you have any other doubt.

Best,
Rolando Acosta (CHU Sainte-Justine)

PapillonMcGill: @racostas I see!

racostas:
Hello @PapillonMcGill ,

I have a conflict of meetings for LORIS today and it's almost sure I could not attend. It would be very helpful if you could explain me a bit more how the new dataprovisionner you mention works and point me out to a class it's been using it. This way I would be able to implement it for this PR.

Regards,
Rolando Acosta (CHU Sainte-Justine)

PapillonMcGill:
You will need to extend new code that is inside #4918. So it might not be wise to do it now.  Maybe do it as currently and we can rebase later to use the provisioner. I was sure it was more advance than currently.

If you wish to use the provisioner now, you can check #4918 as an example. @driusan could give you more details.
How it will work, you retrieve every stats from the DB then filter it using user's sites before sending it to the browser. It require to extend the `DataFrameworkMenu` instead of the `NDB_Menu_Filter`.

Then you'll be able to create a class to represent each rows that extend `\LORIS\Data\DataInstance` and another class, the row provisioner that will extend `\LORIS\Data\Provisioners\DBRowProvisioner`.

php/libraries/User.class.inc
```diff
@@ -362,6 +362,45 @@ class User extends UserPermissions
         return $user_study_sites;
     }
 
+    /**
```
ridz1208: the function `getStudySites()` above should be deprecated, and should be made to use this function instead with the `$study_site`

php/libraries/User.class.inc
```diff
@@ -362,6 +362,45 @@ class User extends UserPermissions
...
+                    $user_sites[$val] = $site[$key]->getCenterName();
+                }
+            }
+        } else {
+            $site = array();
+            foreach ($site_arr as $key => $val) {
+                $site[$key]       = &Site::singleton($val);
+                $user_sites[$val] = $site[$key]->getCenterName();
+            }
+        }
```
ridz1208, line 394: too much redundant code here
```suggestion
        $site = array();
        foreach ($site_arr as $key => $val) {
            $site[$key] = \Site::singleton($val);
            if ($study_site && $site[$key]->isStudySite()) {
                    $user_sites[$val] = $site[$key]->getCenterName();
                    continue;
        }
        $user_sites[$val] = $site[$key]->getCenterName();
```

php/libraries/User.class.inc
```diff
@@ -362,6 +362,45 @@ class User extends UserPermissions
...
 
+    /**
+     * Returns a list with the user's sites
+     *
+     * @param boolean $study_site if true only return study sites
+     * @param boolean $DCC        Whether the DCC should be included or not
+     *
+     * @return array an associative array("center ID" => "site name")
+     */
+    function getSiteList(
```
ridz1208, line 373: As i suggested in the PR below I think the function name should be getSiteNameList() since this doesnt return sites but names of sites

https://github.com/aces/Loris/pull/5892/files#r406274087

php/libraries/User.class.inc
ridz1208: what you should do here is probably replace the content of the function with simply a call to `getSiteList(true,true);` and add an error_log warning stating that this function is deprecated and will be removed in a near future

racostas:
closed by #6878. Please refer to this new version.

PR #5918, opened 2020-01-13
[Phan] Fix PhanTypeMismatchDimFetch errors

driusan:
$_GET['foo'] is usually a string, so we need to ensure
that it's an array places that it's submitted/treated as
an array.


driusan:
Class vars were added accidentally because I seem to have staged them while working on setting 'allow_missing_properties' to false.. but as you said, they're harmless (or perhaps even beneficial) so I'll keep them so that we don't need to fix them later.

PR #5921, opened 2020-01-13
[Travis ] php7.2 to 7.3

kongtiaowang:
## Brief summary of changes
update PHP from 7.2 to 7.3

Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install:
 - make dev
 
 before_script:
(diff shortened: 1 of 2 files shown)


driusan:
Isn't this already done implicitly since it's replacing 7.2 with $TRAVIS_PHP_VERSION?

kongtiaowang:
@driusan It works well in Travis, but if I run the docker test locally, it is still PHP 7.2. it is necessary to update the  Dockerfile.test.php file.

PR #5922, opened 2020-01-13
[Candidate.class.inc] Fixes Empty String for DoB bug and makes DOB nullable

maltheism:
## Brief summary of changes

- Fixed a bug when an empty string is used for the DoB. Originally it would result in an infinite loop in php/libraries/Candidate.class.inc. Now we throw an error.
- birthday should not be required if using edc.

Note: Discovered when writing this PR: https://github.com/aces/Loris/pull/5758

1. Using the API to create a candidate. Supply an empty strong for DoB.


maltheism:
@driusan this PR fixes the bug I showed you a few weeks ago at your desk. The infinite loop bug when the DoB is an empty string. Should be fixed or LORIS instances are vulnerable to a malicious user exploiting it to degrade the server's performance.

php/libraries/Candidate.class.inc
```diff
@@ -200,7 +201,7 @@ class Candidate
      */
     static public function createNew(
         int $centerID,
-        string $dateOfBirth,
+        ?string $dateOfBirth,
```
johnsaigle: The error checking is good but I don't think the type of this param should change. I think DoB is always required? So this should be reverted and line 229 can be deleted also.

php/libraries/Candidate.class.inc
```diff
@@ -225,6 +226,15 @@ class Candidate
                 EDC_NOT_SPECIFIED
             );
         }
+        if ($dateOfBirth !== null
+            && (DateTime::createFromFormat('Y-m-d', $dateOfBirth) === false
+            || empty($dateOfBirth))
+        ) {
+            throw new Exception(
```
```suggestion
            throw new InvalidArgumentException(
```

php/libraries/Candidate.class.inc
maltheism: @johnsaigle DoB isn't always required.

johnsaigle: Why did you open this PR when the changes are already in the other one?

maltheism: @johnsaigle I thought it would be best to pull the code out and into a separate PR because people complain about PRs from me not being condensed. Fate doesn't seem to want me to win in any case. 🥺

johnsaigle: I think it's case-by-case. That other PR looks small and manageable but it's not passing Travis.

maltheism: @johnsaigle I'm not good at judging the case-by-case. I think this PR is good because it fixes a problem that should be fixed. The other PR changes something else but I ran into this problem when working on it and so I had it in it. So I thought today it would be good to pull it into a separate PR. I never have luck with modifying the test code and so I don't think that test code will be fixed anytime soon.

johnsaigle: > I'm not good at judging the case-by-case.

Fair enough. Maybe an explanation would be helpful?

The idea behind preferring PRs to be small and targeted to one issue is that it makes it easier to review. An ideal PR changes only a small number of things and does so clearly, with easy-to-follow testing instructions and documentation.

If you're being asked to make a PR smaller it's usually because someone is having a hard time reviewing it OR that it's doing many things that will make it more difficult to track changes going forward. 

> So I thought today it would be good to pull it into a separate PR. I never have luck with modifying the test code and so I don't think that test code will be fixed anytime soon

This seems like a different discussion. I don't think anyone is ignoring that PR because it's too complicated. It's one of the smaller PRs that are open. However, it's flagged as Needs Work and is not passing Travis. To me, that signals that it doesn't need my attention because the ball is in your court to finish the work and test it using our automated tools.

I think a better investment of time would be to debug the test code. If you're not comfortable doing so or get stuck, you should ask for help from someone (or use the "Help! I don't understand Travis!" flag). Otherwise, the PR can't progress.

maltheism:
```
This seems like a different discussion. I don't think anyone is ignoring that PR because it's too complicated. It's one of the smaller PRs that are open. However, it's flagged as Needs Work and is not passing Travis. To me, that signals that it doesn't need my attention because the ball is in your court to finish the work and test it using our automated tools.

I think a better investment of time would be to debug the test code. If you're not comfortable doing so or get stuck, you should ask for help from someone (or use the "Help! I don't understand Travis!" flag). Otherwise, the PR can't progress.
```

@johnsaigle can a new tag be made for differentiating `Needs work` from a `tests are just failing` tag. I don't like writing or modifying the test code at all because it never works when it should from my last attempts at it for PRs where I've tried. I've also been told to wait for new test code changes for react. So a lot of my PRs when tests don't go right just kept left aside.

On a side note I think this PR should be separate from the other PRs. More so than previous PRs where people have commented.

johnsaigle: What I'm trying to say is that it's helpful if you communicate your intentions to the team so that we can all work together on problems. 

If something is marked as Needs Work and isn't passing the testing suite then there's no reason for anyone to think that that pull request is in a state to be reviewed. 

It'll be easier to keep things from getting left aside if you use something like the "Help" tag or "Blocked" in case the PR needs new React testing changes.

maltheism:
@ridz1208 should this fix be added to 23.0-release?

php/libraries/Candidate.class.inc
```diff
@@ -184,7 +185,7 @@ class Candidate
      *
      * @param int         $centerID              centerID of the site to which the
      *                                           candidate will belong
-     * @param string      $dateOfBirth           date of birth written as YYYY-MM-DD
+     * @param string|null $dateOfBirth           date of birth written as YYYY-MM-DD
```
ridz1208, line 188: @maltheism when is the dob ever null ?

ridz1208: its a required field for participants

maltheism: @ridz1208 we had a loris discussion about this PR. dob can be null from what I remember was said at the loris meeting but I don't remember the exact example. I guess when the subject is expected to be born.

ridz1208: as far as I can tell in 23 DOB is still a required field so I don't see when this case occurs so this seems like it's allowing it where it shouldnt

maltheism: @ridz1208 I'll bring it up at next loris meeting.

ridz1208: looking at the other PR i think that's the usecase for making it nullable

`3. birthday should not be required if using edc.`

it's fine then especially if it was discussed in the LORIS meeting

ridz1208: @maltheism could you update the description to include the change so it doesn't seem accidental

maltheism: @ridz1208 the comment description?

ridz1208: no the description of the PR itself on top

maltheism:
@ridz1208 I think it would be best to have a separate PR for anything else that would fail if the DoB is null. I'm hopeful to get this PR merged as it fixes the specific issue discovered and possibly work on new PRs.

ridz1208:
@maltheism I'm okai with a separate PR just make sure to issue it after this one

ridz1208:
@driusan if ur okai with the proposed solution, you can review and merge this

driusan:
I'm okay with this if it passes Travis..

ridz1208:
@driusan I just fixed a conflict so unless PHPCS complains it should be fine. we'll wait and see

PR #5923, opened 2020-01-13
Add strict type to all files using sleevomat standard

### LINKED ISSUE #5342: [PHPCS] Add sniff for declaration of strict_types
https://github.com/slevomat/coding-standard/#slevomatcodingstandardtypehintsdeclarestricttypes-

This depends on adding the Slevomat coding standard as a developer dependency. This has been proposed already in #4064

PHPCS >= 3.0 is required so this isn't possible until #5349 is merged

driusan:
This adds strict_type to all files using phpcbf and the sleevomat coding standard. With any luck, it'll pass Travis.

Resolves #5342


johnsaigle:
@driusan It may be best to wait for #5396 to be merged before debugging this one in depth. It deals with a ton of type issues (and also adds the slevomat standard).

Otherwise I think that the process of getting this to pass Travis will be a duplication of a large part of that PR.

I think it would be a good idea to add the Blocked label here.

driusan:
@johnsaigle If this doesn't pass Travis (or has more than trivial failures), sure, but I thought this would be a simpler PR that changes far less than that one and resolves an open issue.

johnsaigle:
Agreed in theory but I think in practice that LORIS does a ton of type-juggling and Travis will choke. But we'll see what happens. If it's easy - great, let's merge it. No _a priori_ opposition to it.

PR #5927, opened 2020-01-14
[left bar navigation] shrink left bar width size

lingma:
## Brief summary of changes
The left side bar occupies unnecessarily takes too much space for the most cases, especially when some instruments has big table which will needs more space, even 10 pixels count. In fact, even 250px will not show all the item titles for some instruments.
It is better that the bar is extensible, but for now a fixed value is fine for me.

1. Check left side bar of the instrument under candidate profile, make sure that most items should be visible.

### DIFF
--- a/htdocs/css/simple-sidebar.css
+++ b/htdocs/css/simple-sidebar.css
@@ -3,15 +3,15 @@
 }
 
 .wrapper {
-    padding-left: 250px;
+    padding-left: 200px;
     transition: all 0.4s ease 0s;
 }
 
 #sidebar-wrapper {
-    margin-left: -250px;
+    margin-left: -200px;
     margin-top: 50px;
-    left: 250px;
-    width: 250px;
+    left: 200px;
+    width: 200px;
     top: 0;
     background: #E4EBF2;
     border: 1px solid #C3D5DB;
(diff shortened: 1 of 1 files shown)


johnsaigle:
Which sidebar? Can you provide more details in your testing instructions?

lingma:
before (current): 
image

after PR:
image

lingma:
> Which sidebar? Can you provide more details in your testing instructions?
> 
> Actually screenshots showing before and after would be best.

I just try not leave the left side bar take too much space while some instruments tries hard to have a little more space. Some even has no space left with small screen.

lingma:
> Works for me. I don't have a strong opinion either way

It is not a big issue for the most users, but some will have difficulty, I am trying to get an image, but always "504 Gateway time-out".

johnsaigle:
Is that related to this PR? The 504 Gateway Timeout might be an issue with the VPN, not anything in the code

lingma:
> Is that related to this PR? The 504 Gateway Timeout might be an issue with the VPN, not anything in the code

Nothing related with this PR, I was trying to say that I am not able to get you an example with lots of items on one line

johnsaigle:
Ah ok, thanks for explaining.

lingma:
image
Example of the space usage when the right side asks for a lot space.

PR #5911, opened 2020-01-09
[dataquery] Update AJAX to fetch for saveQuery

maltheism:
## Brief summary of changes

I'm updating the $.ajax to fetch with endpoints. This PR handles the saveQuery.
See #5909 

#### Testing instructions (if applicable)

1. Visit DQT
2. Make Query
3. Save Query

Blocked for #5909

PR #5929, opened 2020-01-15
[dataquery] Update AJAX to fetch for datadictionary requests

maltheism:
## Brief summary of changes

I'm updating the $.ajax to fetch with endpoints. This PR handles the datadictionary requests. They all get updated to the dataquery/php/view.class.inc endpoint.

**Note:** I had to use async await in this PR and required the @babel/plugin-transform-runtime @babel/runtime dependencies for react to be able to use async await. The reason is that part of the old code had synchronous ajax requests and fetch requires async await to mimic the past behaviour.  

1. Visit the DQT
2. Make query & run it.
3. Save query & load it.


maltheism:
Waiting on PR https://github.com/aces/Loris/pull/5911 to be merged first. That's the reason for the Blocked tag.

maltheism:

PR #5930, opened 2020-01-15
[dataquery] Update getAllSessions to new endpoint

maltheism:
## Brief summary of changes

The getAllSessions.php isn't being used in the dataquery tool. I moved it to the new endpoint regardless.

1. Visit the DQT
2. Make query & run it.
3. Save query & load it.


maltheism:
I added the Blocked tag for waiting on merge of PR https://github.com/aces/Loris/pull/5929

driusan:
If you're sure that it's not being used (I thought it was used in the case where there's no filters defined and it needs to get the data for everything?) why not just delete it?

maltheism:
@driusan no where in the frontend is making a request to getAllSessions. I thought to keep it because maybe the query is useful for something that never got implemented. Should I delete it?

driusan:
How does the frontend know what sessions to retrieve when the query has no filters defined?

maltheism:
@driusan all the sessions seems to be coming from the server.
The DataQueryApp component is supplied props 'AllSessions' in the form_dataquery.tpl found at modules/dataquery/templates/form_dataquery.tpl in LORIS. Inside the dataquery.class.inc there is a query that gets all the sessions on line 140.

maltheism:

PR #5932, opened 2020-01-15
[UI] Fix width of select elements that have no label

driusan:
Select elements that have no label are supposed to take the whole row,
but the logic in the if statement was incorrect and no label was resulting
in an empty 3 columns of the bootstrap grid.

This fixes the logic.

### DIFF
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -478,7 +478,7 @@ class SelectElement extends Component {
     // element will take up the whole row.
     let label = null;
     let inputClass = 'col-sm-12';
-    if (this.props.label || this.props.label == '') {
+    if (this.props.label && this.props.label != '') {
       label = (
         <label className="col-sm-3 control-label" htmlFor={this.props.label}>
           {this.props.label}


ridz1208:
@driusan can you add a before/after screenshot ?

driusan:
@ridz1208 This is in a react component. I don't know where it's currently used in LORIS.

I wanted to add the functionality for use in a new module and discovered that it was already supposed to be there (as described in the comment right above the code I changed), but not working. (The usage I was trying to use it for was adding a dropdown in a column in a table, but each row had a 3 bootstrap grid blank space being added for the label that wasn't there, because it's in the table header.)

maltheism:
This PR does fix the intended purpose of the if statement for when a label exists compared to an empty string. @driusan @ridz1208 The <SelectElement> component is being used for all forms in LORIS that needs a select element. The new_profile module uses it as well.

driusan:
@maltheism but do you know anywhere that uses it with an empty label to check before/after the bug fix for @ridz1208?

ridz1208:
@driusan maybe take a screenshot of the place you are trying to implement it ?

I dont care too much about it but it just makes it clearer

maltheism:
@driusan @ridz1208 I think every module in LORIS is using a label when using the SelectElement component.

PR #5933, opened 2020-01-15
[tools] Stole DQT fix from IBIS

ridz1208:
## Brief summary of changes
Somewhere in LORIS 17.0 branch maintenance the query on the `selected` flag in the MRI Couchdb importer was removed. this is a critical part of the script.

IBIS had fixed the problem, this is the fix copied from their code

all creds and comments to @nicolasbrossard

### DIFF
--- a/tools/CouchDB_MRI_Importer.php
+++ b/tools/CouchDB_MRI_Importer.php
@@ -96,23 +96,81 @@ class CouchDBMRIImporter
 
         foreach($s as $scan){
             $scantype=$scan['ScanType'];
-            $Query .= ", (SELECT f.File FROM files f LEFT JOIN mri_scan_type msc
-              ON (msc.ID= f.AcquisitionProtocolID)
-              WHERE f.SessionID=s.ID AND msc.Scan_type='$scantype' LIMIT 1)
-                    as `Selected_$scantype`, (SELECT fqc.QCStatus
-                    FROM files f 
-                    LEFT JOIN files_qcstatus fqc USING(FileID)
-                    LEFT JOIN mri_scan_type msc ON(msc.ID= f.AcquisitionProtocolID)
-              WHERE f.SessionID=s.ID AND msc.Scan_type='$scantype' LIMIT 1)
-                     as `$scantype"."_QCStatus`";
+
+            //-------------------------------------------------------------------------//
+            // Add to the query a subselect that will compute the selected file        //
+            // for the given session/modality. If more than one selected file exists,  //
+            // the subquery will return 'Multiple Selected files'. If no selected file //
+            // exists, the subselect will return 'No selected file'. Otherwise the     //
+            // file path of the selected file is returned by the subselect             //
+            //-------------------------------------------------------------------------//
+            $Query .= ', '
+                    . '('
+                    . ' CASE ('. $this->_getQueryForSelectedFiles('count(*)', $scantype) . ') '
+                    . '  WHEN 1 '
+                    . '    THEN (' . $this->_getQueryForSelectedFiles('f.File', $scantype) . ') '
+                    . '  WHEN 0 '
+                    . '    THEN "No selected file" '
+                    . '  ELSE "Multiple Selected files"'
+                    . ' END'
+                    . ") as Selected_$scantype";
+
+            //---------------------------------------------------------------------------//
+            // Add to the query a subselect that will compute the Qc status of the       //
+            // selected file for the given session/modality. If more than one selected   //
+            // file exists, the subquery will return 'Unknown: multiple Selected files'. //
+            // If no selected file exists, the subselect will return 'No selected file'. //
+            // Otherwise the subselect will return the Qc status of the selected file (or//
+            // 'No Qc on selected file' if the file has not been Qced)                   //
+            //---------------------------------------------------------------------------//
+            $Query  .= ', '
+                    . '('
+                    . ' CASE (' . $this->_getQueryForSelectedFiles('count(*)', $scantype) . ') '
+                    . '  WHEN 1 '
+                    . '    THEN ('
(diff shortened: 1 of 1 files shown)


ridz1208:
@driusan have you decided if you like or not this fix ?

driusan:
@ridz1208 it hasn't been reviewed by anyone.

ridz1208:
@nicolasbrossard can you review ?

tools/CouchDB_MRI_Importer.php
```diff
@@ -96,24 +96,82 @@ function _generateCandidatesQuery($ScanTypes)
...
-        $Query .= " FROM session s JOIN candidate c USING (CandID)
-            LEFT JOIN feedback_mri_comments fmric
-            ON (fmric.CommentTypeID=7 AND fmric.SessionID=s.ID)
-            WHERE c.Entity_type != 'Scanner' AND c.PSCID NOT LIKE '%9999'
-                  AND c.Active='Y' AND s.Active='Y' AND s.CenterID <> 1";
+
+        $Query .= " FROM session s"
+                . " JOIN candidate c USING (CandID)"
+                . " LEFT JOIN feedback_mri_comments fmric"
+                . " ON (fmric.CommentTypeID=7 AND fmric.SessionID=s.ID)"
```
nicolasbrossard: I'd define a constant here instead of using `7` (to give a clue as to what type of comment we're looking for).

tools/CouchDB_MRI_Importer.php
```diff
@@ -96,24 +96,82 @@ function _generateCandidatesQuery($ScanTypes)
...
+                . " WHERE c.Entity_type != 'Scanner' AND c.PSCID NOT LIKE '%9999'"
```
nicolasbrossard: I have no idea why we reject the pscids that are like `'%9999'`...You?

tools/CouchDB_MRI_Importer.php
```diff
@@ -96,24 +96,82 @@ function _generateCandidatesQuery($ScanTypes)
...
+                . " AND c.Active='Y' AND s.Active='Y' AND s.CenterID <> 1";
```
nicolasbrossard: Maybe a constant here also, instead of `1`, for the center ID.

tools/CouchDB_MRI_Importer.php
ridz1208: what type of comment are you looking for ?
 7 doesnt match many entries in RB

driusan: I don't know about RB, but in my data I have:

```
mysql> select * from feedback_mri_comment_types WHERE CommentTypeID=7\G
*************************** 1. row ***************************
     CommentTypeID: 7
       CommentName: Subject
       CommentType: visit
CommentStatusField: 
1 row in set (0.00 sec)
```

nicolasbrossard: Don't look in RB, look in LORIS' `0000-00-00-schema.sql`. 7 Matches a comment type that can be applied to all the images acquired during the visit (e.g subject super stressed out to be in the scanner, so movement can be seen in all images, etc...)

tools/CouchDB_MRI_Importer.php
```diff
@@ -96,24 +96,82 @@ function _generateCandidatesQuery($ScanTypes)
...
+            //-------------------------------------------------------------------------//
+            // Add to the query a subselect that will compute the selected file        //
+            // for the given session/modality. If more than one selected file exists,  //
+            // the subquery will return 'Multiple Selected files'. If no selected file //
+            // exists, the subselect will return 'No selected file'. Otherwise the     //
+            // file path of the selected file is returned by the subselect             //
+            //-------------------------------------------------------------------------//
+            $Query .= ', '
+                    . '('
+                    . ' CASE ('. $this->_getQueryForSelectedFiles('count(*)', $scantype) . ') '
```
cmadjar: Why give only a count here and not return the list of filenames? I get that they would not be downloadable anyway but I feel it would help the process to be able to see the filenames of the files listed in that cell when there is more than one file so you can then provide a list of the filenames you wish to get transferred to you?

tools/CouchDB_MRI_Importer.php
cmadjar: fascinating... In PREVENT-AD we had a candidate MTL9999 that needed to be excluded from the queries... Could this be some leftover of PREVENT-AD? LOL

tools/CouchDB_MRI_Importer.php
```diff
@@ -96,24 +96,82 @@ function _generateCandidatesQuery($ScanTypes)
...
+                    . '  WHEN 1 '
+                    . '    THEN (' . $this->_getQueryForSelectedFiles('f.File', $scantype) . ') '
+                    . '  WHEN 0 '
+                    . '    THEN "No selected file" '
+                    . '  ELSE "Multiple Selected files"'
```
cmadjar: Why only showing multiple here and not return the list of filenames? I get that they would not be downloadable anyway but I feel it would help the process to be able to see the filenames of the files listed in that cell when there is more than one file so you can then provide a list of the filenames you wish to get transferred to you?

ridz1208: @nicolasbrossard any answers ?

nicolasbrossard: @ridz @cmadjar I think displaying all the file names would take up too much space in the result table. I believe we decided to use 'Multiple Selected files' instead in an imaging meeting but I could be wrong...

ridz1208:
@driusan @cmadjar @nicolasbrossard 

I'm closing this PR because I believe more clean up is necessary for it to be useful and I'm not relaly the person to do that. I definitely think it should be done but I think someone in hte imaging team should take it on. 

issue #6054

PR #5937, opened 2020-01-16
[bug] pageclass object null warning

lingma:
## Brief summary of changes

Each page loading of timepoint list, instrument list and instrument will trigger a warning "PHP Notice: Trying to get property 'page' of non-object in UserPageDecorationMiddleware.php", which is not pleasant to see when you are working with them every day. I assume that the problem comes from the Loris logic to process first without passing by timepoint module, if not having gone through then fail back to timepoint/instrument. I don't want to modify the logic. As a result, I just add a check to avoid the PHP warning.

1. Load a timepoint list, instrument list and instrument page, then check web server log. The fix will avoid the warning.

This is the recreation of the PR #5888, which I have wrongly pushed some other commits.

### DIFF
--- a/src/Middleware/UserPageDecorationMiddleware.php
+++ b/src/Middleware/UserPageDecorationMiddleware.php
@@ -132,24 +132,24 @@ class UserPageDecorationMiddleware implements MiddlewareInterface
         // I don't think anyone uses this. It's not really supported
         $tpl_data['css'] = $this->Config->getSetting('css');
 
-        // Some page might call without pageclass
+        $page = null;
         if (!is_null($request->getAttribute("pageclass"))) {
             $tpl_data['subtest'] = $request->getAttribute("pageclass")->page;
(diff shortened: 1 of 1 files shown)


lingma:
> @lingma This needs to be rebased, it's including all sorts of commits that shouldn't be there.
@driusan 
I tried, not be able to find the exact cause that why the object is null, I assume that just check null is enough for now. If you look at the `PageDecorationMiddleware`, there is a `pageclass` object check, similarly I think a check in `UserPageDecorationMiddleware` is necessary and enough. 

I wrongly did the rebase, I don't know how to remove those commits. So I just recreated one PR.

johnsaigle:
I think this may be a duplicate of https://github.com/aces/Loris/pull/6013.

lingma:
> I think this may be a duplicate of #6013.

right, they are similar.

PR #5944, opened 2020-01-21
Rename project directory

### LINKED ISSUE #5931: Rename the "project/" directory (containing config.xml, overrides, etc.)
No

As discussed in the LORIS meeting, it would be good to disambiguate the word "project". Currently we have a Project class which represents the idea of a study project that is being managed by LORIS. At the same time we have the `project/` directory which contains configuration files and custom modules, instruments, etc.

We need to rename this folder and update the hard-coded paths in LORIS accordingly.

johnsaigle:
## Brief summary of changes
We discussed renaming this file to disambiguate our use of the word project. Currently, when we say "project" it can refer to a bundle of overrides contained in a directory OR to the concept of a Project which has relationships to the idea of a Subproject, Site, TimePoint, etc and is represented by the `Project.class.inc` file.

### Discussion
I prefer the new name "Custom" for the directory. I think it encompasses the idea that:
1. This folder is the place to put things particular to a LORIS instance/study, such as module overrides, instruments, and configuration settings
2. This folder belongs ONLY to this particular LORIS instance; it should not be shared or committed to the repository. It belongs to the project admins; we shouldn't add things to it.

We can decide on a name together. At that point, I'll update the various references to `project/config.xml` in the source code and documentation. Until then, I'll leave this PR in draft mode.

Rename your `project/` directory to `custom/`. LORIS should function as before without changes.

* Resolves #5931

### DIFF
--- a/docs/wiki/00 - SERVER INSTALL AND CONFIGURATION/02 - Website Configuration/Custom Directory.md
+++ b/docs/wiki/00 - SERVER INSTALL AND CONFIGURATION/02 - Website Configuration/Custom Directory.md
@@ -1,0 +1,38 @@
+# Custom Directory
+
+The `custom/` directory contains files that override the core functionality of
+LORIS. This can include custom modules and instruments that you can modify
+independently of the main source code of the LORIS repository.
+
+All files in this directory will be ignored by git and will not be committed
+to our repository if you push changes. The directory contains highly sensitive
+information (see below) and so these files should not be shared outside of your
+administrative team.
+
+## LORIS configuration file (config.xml)
+The file `custom/config.xml` will be created during the install process.
+It contains several important configuration settings for your LORIS instance,
+including credentials for the database. For this reason, this file must be kept
+private.
+
+## Module and Library Overrides
+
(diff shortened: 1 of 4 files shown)


docs/wiki/00 - SERVER INSTALL AND CONFIGURATION/02 - Website Configuration/Custom Directory.md
```diff
@@ -0,0 +1,38 @@
...
+
+By placing custom code in this directory instead of modifying the source code
+directly, you will be able to track and manage your custom code independent
+of updates to the source code.
+
+The same can be done with LORIS core libraries by copying code from
+`php/libraries/` to `custom/libraries/`.
+
+## Instruments
+TODO
```
johnsaigle: I don't know exactly how this works so I'd like help writing this part.

driusan:
I don't think "custom" is a good name and I don't think this should be done until #6022 when there's a centralized place to update if the name changes.

johnsaigle:
I'm going to close this because I don't think it will get resolved by the time I leave. The GitHub is still open.

PR #5945, opened 2020-01-21
[Candidate] Add getAge function

zaliqarosli:
## Brief summary of changes

Calculates the candidate's age today using php's DateTime obj diff() function in:
- years, months, days
- years
- months
- days

If you're not a huge fan of doing it this way, look at commit 2 which uses \Utility::calculateAge function (not as accurate for calculating number of days)

I have added changes to timepoint_list to show an example of how to use these functions and for testing. These can be removed before merging. Travis will fail until these changes to timepoint_list are removed.

1. Edit line 109 of timepoint_list.class.inc to use all the getAge..() functions.
2. Got to Access Profile, click on a candidate, view the Age in the table at the top of the candidate page. Test this for years, months, and days.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ changes in the following format: PR #1234***
 Quality Control, and Behavioural Quality Control. (PR #6041)
 - Addition of a new `account_request_date` in `users` table that will be used when
 requesting a new account and will be displayed in the User Accounts module (PR #6191)
+- Candidate's age can be retrieved from the Candidate class in days, months, or years (PR #5945)
 #### Bug Fixes
 - *Add item here*
 ### Modules
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,10 +46,6 @@ requesting a new account and will be displayed in the User Accounts module (PR #
 database (PR #5260)
 - New documentation for file permissions has been added to the README.md file. (PR #5323)
 - Dashboard study progression section performance improvement (PR #5887)
-- Very old instruments relying on QuickForm may have issues due to code changes (PR #4928)
-- Unix user permissions have been updated which may affect access to files. New
-documentation for file permissions has been added to the README.md file (PR #5323)
-- Candidate's age can be retrieved from the Candidate class in days, months, and years (PR #5945)
 
 #### Bug Fixes
 - Fix edge-case that gave a confusing error message when changing password (PR #5956)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,10 @@ requesting a new account and will be displayed in the User Accounts module (PR #
 database (PR #5260)
 - New documentation for file permissions has been added to the README.md file. (PR #5323)
 - Dashboard study progression section performance improvement (PR #5887)
+- Very old instruments relying on QuickForm may have issues due to code changes (PR #4928)
+- Unix user permissions have been updated which may affect access to files. New
+documentation for file permissions has been added to the README.md file (PR #5323)
+- Candidate's age can be retrieved from the Candidate class in days, months, and years (PR #5945)
 
 #### Bug Fixes
 - Fix edge-case that gave a confusing error message when changing password (PR #5956)
--- a/modules/timepoint_list/php/timepoint_list.class.inc
+++ b/modules/timepoint_list/php/timepoint_list.class.inc
@@ -79,10 +79,7 @@ class Timepoint_List extends \NDB_Menu
         $this->tpl_data['candID'] = $this->candID;
         $this->tpl_data['PSCID']  = $candidate->getPSCID();
         $listOfTimePoints         = $candidate->getListOfTimePoints();
-        $this->tpl_data['candidate'] = array_merge(
-            $candidate->getData(),
-            array('Age' => $candidate->getAgeInMonths())
-        );
+        $this->tpl_data['candidate'] = $candidate->getData();
 
         if (!empty($listOfTimePoints)) {
             $user     =& \User::singleton();
--- a/modules/timepoint_list/php/timepoint_list.class.inc
+++ b/modules/timepoint_list/php/timepoint_list.class.inc
@@ -79,7 +79,10 @@ class Timepoint_List extends \NDB_Menu
         $this->tpl_data['candID'] = $this->candID;
         $this->tpl_data['PSCID']  = $candidate->getPSCID();
         $listOfTimePoints         = $candidate->getListOfTimePoints();
-        $this->tpl_data['candidate'] = $candidate->getData();
+        $this->tpl_data['candidate'] = array_merge(
+            $candidate->getData(),
+            array('Age' => $candidate->getAgeInMonths())
+        );
 
         if (!empty($listOfTimePoints)) {
             $user     =& \User::singleton();
--- a/modules/timepoint_list/templates/menu_timepoint_list.tpl
+++ b/modules/timepoint_list/templates/menu_timepoint_list.tpl
@@ -6,6 +6,9 @@
     <th>
       DOB
     </th>
+    <th>
+      Age
+    </th>
     {if $candidate.EDC!=""}
       <th>
         EDC
@@ -29,6 +32,9 @@
   <tr>
     <td>
       {$candidate.DoB}
+    </td>
+    <td>
+      {$candidate.Age}
     </td>
     {if $candidate.EDC!=""}
       <td>
--- a/modules/timepoint_list/templates/menu_timepoint_list.tpl
+++ b/modules/timepoint_list/templates/menu_timepoint_list.tpl
@@ -6,9 +6,6 @@
     <th>
       DOB
     </th>
-    <th>
-      Age
-    </th>
     {if $candidate.EDC!=""}
       <th>
         EDC
@@ -32,9 +29,6 @@
   <tr>
     <td>
       {$candidate.DoB}
-    </td>
-    <td>
-      {$candidate.Age}
     </td>
     {if $candidate.EDC!=""}
       <td>
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -713,7 +713,7 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
      *
      * @return int The candidate's age today as years
      */
-    public function getAgeYears(): int
+    public function getAgeInYears(): int
     {
         return (int)$this->getAge()->format('%y');
     }
@@ -726,7 +726,7 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
     public function getAgeInMonths(): int
     {
         return (int)$this->getAge()->format('%m') +
-            12 * $this->getAgeYears();
+            12 * $this->getAgeInYears();
     }
 
     /**
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -689,17 +689,52 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
     }
 
     /**
-     * Returns the candidate's calculated age today
+     * Returns the candidate's calculated age today in years, months, days
      *
-     * @return int The candidate's age today
+     * @return array The candidate's age today in an array with values for keys 'year',
+     *               'mon', and 'day'
      */
-    public function getAge(): int
+    public function getAge(): array
     {
-        $ageArray = \Utility::calculateAge(
+        return \Utility::calculateAge(
             $this->candidateInfo["DoB"],
             (new DateTime())->format('Y-m-d')
         );
-        return $ageArray['year'];
+    }
+
+    /**
+     * Returns the candidate's calculated age today in int years
+     *
+     * @return int The candidate's age today as years
+     */
+    public function getAgeYears(): int
+    {
+        return $this->getAge()['year'];
+    }
+
+    /**
+     * Returns the candidate's calculated age today in total number of months
+     *
+     * @return int The candidate's age today in months
+     */
+    public function getAgeInMonths(): int
+    {
+        $age    = $this->getAge();
+        $months = $age['year']*12 + $age['mon'] + ($age['day']/30);
+
+        return (round($months*10) / 10.0);
+    }
+
+    /**
+     * Returns the candidate's calculated age today in total number of days
+     *
+     * @return int The candidate's age today in days
+     */
+    public function getAgeInDays(): int
+    {
+        $age    = $this->getAge();
+
+        return $age['year']*365 + $age['mon']*30 + $age['day'];
     }
 
     /**
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -693,7 +693,7 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
      *
      * @return DateInterval Between DoB and now
      */
-    public function getAgeDiff(): DateInterval
+    public function getAge(): DateInterval
     {
         $dob = new DateTime($this->candidateInfo["DoB"]);
 
@@ -707,7 +707,7 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
      */
     public function getAgeYears(): int
     {
-        return (int)$this->getAgeDiff()->format('%y');
+        return (int)$this->getAge()->format('%y');
     }
 
     /**
@@ -717,7 +717,7 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
      */
     public function getAgeInMonths(): int
     {
-        return (int)$this->getAgeDiff()->format('%m') +
+        return (int)$this->getAge()->format('%m') +
             12 * $this->getAgeYears();
     }
 
@@ -728,7 +728,7 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
      */
     public function getAgeInDays(): int
     {
-        return (int)$this->getAgeDiff()->days;
+        return (int)$this->getAge()->days;
     }
 
     /**
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -691,8 +691,8 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
     /**
      * Returns the candidate's calculated age today in years, months, days
      *
-     * @return array The candidate's age today in an array with values for keys 'year',
-     *               'mon', and 'day'
+     * @return array The candidate's age today in an array with values for
+     *               keys 'year', 'mon', and 'day'
      */
     public function getAge(): array
     {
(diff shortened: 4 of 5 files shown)


ridz1208:
@zaliqarosli I dont have a problem with it but I'm not sure why you are truncating the months and days. You are making it unusable for any autism study, they would only get 0 all the time.

johnsaigle:
Hmm yeah it may be good to change the name of the function to `getAgeInYears()`.

Maybe an even cleaner approach would be to move calculateAge from Utility to Candidate. Then you could add wrapper functions in the Candidate class that call calculateAge, e.g.

```php
public function getAgeInYears(): int {
    return $this->calculateAge(
         $this->candidateInfo['DoB'], 
         (new \DateTime()->format('Y-m-d')
    )['year'];
}
```

And do another one for getAgeInMonths and getAgeInDays. Then it would work for many types of studies.

zaliqarosli:
@johnsaigle @ridz1208 ya i am down to make separate functions for in years, in months, and in days

ridz1208:
Well... this is not technically giving you an age IN years. it's giving you the number of full years in the age.

ie a 3 month old baby should not be 0 but rather 0.25 years

johnsaigle:
Yeah it's int years not float years

zaliqarosli:
@ridz1208 i know what you mean, but it that case, the project can use getAgeInMonths() instead of getAgeInYears(). I don't think a Parkinson's project for example would find use in seeing a candidate's age as 85.80 years as opposed to just 85 years

zaliqarosli:
but now I'm thinking do we also want a function called getAgeInYears&Months????

thomasbeaudry:
Why not take an input variable that specifies years, months, or days instead of having 3 functions?

php/libraries/Candidate.class.inc
```diff
@@ -657,17 +657,52 @@ class Candidate
...
+
+    /**
+     * Returns the candidate's calculated age today in total number of months
+     *
+     * @return int The candidate's age today in months
+     */
+    public function getAgeInMonths(): int
+    {
+        $age    = $this->getAge();
+        $months = $age['year']*12 + $age['mon'] + ($age['day']/30);
```
johnsaigle: I think I wouldn't include the days portion for this one. I think rounding up or down could potentially be problematic especially since we're using 1 month = 30 days. The value for "days" doesn't actually have any precision in that case so it's misleading to include it as a significant figure in the calculation.

Either way, all of those caveats should be explained in this function comment as well as in `getAgeInDays()`.

php/libraries/Candidate.class.inc
```diff
@@ -657,17 +657,52 @@ class Candidate
...
         );
-        return $ageArray['year'];
+    }
+
+    /**
+     * Returns the candidate's calculated age today in int years
+     *
+     * @return int The candidate's age today as years
+     */
+    public function getAgeYears(): int
```
```suggestion
    public function getAgeInYears(): int
```

Just for consistency

php/libraries/Candidate.class.inc
```diff
@@ -657,17 +657,52 @@ class Candidate
...
+
+        return (round($months*10) / 10.0);
+    }
+
+    /**
+     * Returns the candidate's calculated age today in total number of days
+     *
+     * @return int The candidate's age today in days
+     */
+    public function getAgeInDays(): int
```
johnsaigle: Maybe this function shouldn't exist because we can't actually give any precision based on days.

zaliqarosli:
@thomasbeaudry because the calculations are all different and grouping them into one function that takes as input the different cases wouldn't necessarily make the code or functions any cleaner/succinct

php/libraries/Candidate.class.inc
zaliqarosli: @johnsaigle hold your horses, i have another commit coming, and then i'll update the PR description

php/libraries/Candidate.class.inc
zaliqarosli: this is done deliberately to mean that its the year or your age vs. your age in year i.e. 2.25 years as opposed to just 2 years

johnsaigle: Whether the word "In" is there or not doesn't make it clear that you're returning an integer year vs. a decimal year. 

The other functions have "In" in the name and also return integers.

zaliqarosli:
@johnsaigle @ridz1208 okay checkout the calculations using php's DateTime diff() functionality in my last commit. I made commit 2 incase commit 3 causes controversy xD

php/libraries/Candidate.class.inc
johnsaigle: I still think this is true even with DateTime. `calculateAge()` makes all months equal to 30 days so it's misleading to say we can provide precision at the level of days.

zaliqarosli: i don't understand your comment. we're not using calculateAge() anymore here. also, the the number of days that I get from using DateTime diff is different to when using the results from calculateAge() and averaging the number of days in a month to be 30.

how do you know that DateTime does it this way without precision?

zaliqarosli:
1. So regarding the issue in analysis, that has probably been brought up before hence the warning in the doc comments of the Utility::calculateAge() function itself. This is a caveat for behavioural data and is not a new issue. however, i'm no longer using this function, so why is this still a concern for this PR?

2. I don't really care to fight for this one but this is something that @ridz1208 and I agreed on in order to provide nuance

3. Yes, if I can learn how to make them

4. On the contrary, I think it is useful to return the year, month and day as an array so that an implementation of that function can choose to use all or only one of the values. For example, it allows a use to either use 'years months days' or just 'year months' if they so choose. I also think the function doc comments can help with any confusion. I'm also not sure about returning a DateTime when calling for an age, as the result wouldn't really make sense without already formatting it.

zaliqarosli:
@johnsaigle this is the document for formatting by ->days https://www.php.net/manual/en/class.dateinterval.php#dateinterval.props.days. it looks pretty legit to me. i'll do more research on it

php/libraries/Candidate.class.inc
zaliqarosli: @johnsaigle this is the document for formatting by ->days https://www.php.net/manual/en/class.dateinterval.php#dateinterval.props.days. it looks pretty legit to me. i'll do more research on it

johnsaigle:
1. It might be weird to have two different places in the code that give different results for somebody's age.

2. Agree to disagree. Won't block the PR over it.

3. Cool, let me know if you want help

4.  > it allows a use to either use 'years months days' or just 'year months' if they so choose

> the result wouldn't really make sense without already formatting it

Don't these sentences contradict each other? In the first you're saying that you want the users to format on their own and then you're saying you want to format it for them.

DateTime is unformatted which I think is ideal. If you return a DateTime then the calling code function could just do `$age->format('Y-m-d')` or `$age->format('Y-m)` to get 'years months days' or 'years months'. Also, what if they want seconds? (Yes, they probably won't for age, I know.) We don't have that option with a custom array. 

It would also allow us to move forward with using DateTimes everywhere for internal code instead of flip-flopping between arrays, strings, and DateTimes as we do now.

To me it seems DateTimes give us more flexibility and consistency and allow us to take advantage of types. 
With arrays, we have a fixed format, limited precision, inconsistency with other places in the code (except for Utility::calculateAge -- which we're deliberately trying to distinguish this code from because they behave differently), and make it so that a person has to read the function comment to know what's going on.


You're right about the "->days"/DateInterval part. I don't have a problem there.

php/libraries/Candidate.class.inc
johnsaigle: Sorry this is leftover from before when getAge was using calculateAge. This is fine now.

johnsaigle:
OK hold on I realize how I was being confusing/was confused.

I think getAge could return a `DateInterval` instead. DateTime doesn't make sense.

zaliqarosli:
@johnsaigle i don't see the contradiction. i think you're misunderstanding me. what the functions are formatting is not a DateTime object, its a DateInterval object, as that is what is returned by diff(). Returning a DateInterval object without formatting it doesn't seem useful to me. Users can format the age in whatever way they want. And I think providing the DateInterval data in a way that makes this possible can be useful. For example, having getAge() return a string "%y Years, %m Months, %d days" is not as useful as returning an array where its values %y, %m, %d can then be used.

You're talking about DateTime but that is not the result type that we're dealing with here. How would you yourself deal with the DateInterval returned by diff before you return it in the getAge() function?

zaliqarosli:
@johnsaigle getAge could just return a DateInterval but then we'd need more functions to then return meaningful data from the candidate class

johnsaigle:
> i think you're misunderstanding me.

I definitely was. Sorry. I think what I said still holds if getAge returns a DateInterval though.

>How would you yourself deal with the DateInterval returned by diff before you return it in the getAge() function?
> For example, having getAge() return a string "%y Years, %m Months, %d days" is not as useful as returning an array where its values %y, %m, %d can then be used.

DateInterval does that on its own though. You're creating an array around DateInterval that has fewer options.
I would use it like this:

```php
$ageInterval = $candidate->getAge();
// print time since birth.
echo $ageInterval->format("%y years, %m months, %d days, and %h hours");

> getAge could just return a DateInterval but then we'd need more functions to then return meaningful data from the candidate class

I think you'd just need to modify the ones you have here

```php
function getAgeInMonths(): int {
    $ageInterval = $this->getAge();
    return (int) $ageInternval->format("%y") * 12 + (int) $ageInterval->format("%m");
}

**Comment by ridz1208** [comment]
@zaliqarosli I like how this is looking

File: modules/timepoint_list/templates/menu_timepoint_list.tpl [hunk 4]
```diff
@@ -6,6 +6,9 @@
     <th>
       DOB
     </th>
+    <th>
+      Age
```
johnsaigle: Maybe this should be 'Age in Months' since that's what you're returning above?

johnsaigle:
As far as I can tell, the integration tests in `modules/timepoint_list/test/timepoint_listTest.php` are failing because that file creates a couple of Candidates and inserts them into the database. However they are not provided with a `DoB` value which is messing up the tests.

If you add DoBs for the candidates by modifying the arrays in this file, I think that should fix it. 

https://github.com/aces/Loris/blob/73ba88808030297da20674704b09f768b183a5c2/modules/timepoint_list/test/timepoint_listTest.php#L111-L149

modules/timepoint_list/templates/menu_timepoint_list.tpl
zaliqarosli: i only added this to show an implementation of the new feature. not sure if we want it to be merged. i can add "(months)" but what if people actually want in years?  or not show the age at all?

johnsaigle: Maybe a config option should be added then so that projects can decide?

zaliqarosli: then this PR becomes a whole more than just introducing these new functions. maybe a separate PR for implementation can be opened

johnsaigle: Works for me. I think this part of the code should be made consistent or deleted before merging though

zaliqarosli: yepp, i mentioned deleting before merging in the PR description

modules/timepoint_list/php/timepoint_list.class.inc
```diff
@@ -104,7 +104,10 @@ class Timepoint_List extends \NDB_Menu
         $this->tpl_data['candID'] = $this->candID;
         $this->tpl_data['PSCID']  = $candidate->getPSCID();
         $listOfTimePoints         = $candidate->getListOfTimePoints();
-        $this->tpl_data['candidate'] = $candidate->getData();
+        $this->tpl_data['candidate'] = array_merge(
+            $candidate->getData(),
+            array('Age' => $candidate->getAgeInMonths())
+        );
```
zaliqarosli: for testing purposes only. to be deleted before merging.

modules/timepoint_list/templates/menu_timepoint_list.tpl
```diff
@@ -6,6 +6,9 @@
...
+    </th>
```
zaliqarosli: for testing purposes only. to be deleted before merging.

modules/timepoint_list/templates/menu_timepoint_list.tpl
```diff
@@ -30,6 +33,9 @@
     <td>
       {$candidate.DoB}
     </td>
+    <td>
+      {$candidate.Age}
+    </td>
```
zaliqarosli: for testing purposes only. to be deleted before merging.

php/libraries/Candidate.class.inc
PapillonMcGill: I second @johnsaigle argument about the fact that the others 2 functions have `in` in the name. Should be all with or without for consistency.

test/unittests/CandidateTest.php
```diff
@@ -658,6 +658,46 @@ public function testGetNextVisitNoReturnsOneWhenNextVisitDoesNotExist()
         $this->_candidate->select($this->_candidateInfo['CandID']);
         $this->assertEquals(1, $this->_candidate->getNextVisitNo());
     }
+    /**
+     * Test getAgeYears returns age as int years
+     *
+     * @covers Candidate::getAgeYears()
+     * @return void
+     */
+    public function testGetAgeYearsReturnsIntYears()
```
zaliqarosli: @johnsaigle would you mind looking at these test functions? I'm not quite sure what I'm doing here. do you also have any insights as to how to write the test function for getAge()?

johnsaigle: The current unit tests look good!

I did a bit research on unit testing functions that refer to the current date "now" which is a little tricky because unit tests rely on a static value. For example we would want to write a test that, given a candidate born on e.g. January 1st, 2019, the `getAge()*` functions would return 1 year, 13 months, or 365 + 24 days depending on what date interval you care about. (I'm writing this on Feb. 25th 2020). But if we hardcode that then we can't test for 365 + 24 days because that will change each day...

The way to approach this is by changing the function getAge() to this:

```php
    public function getAge($referenceDate = new DateTime()): DateInterval
    {
        $dob = new DateTime($this->candidateInfo["DoB"]);

        return $dob->diff($referenceDate);
    }
```

This allows us to choose a moment from which to measure the candidate's age. It means adding an optional parameter to getAge() but we don't have to use it in the actual codebase.

Doing this means that when the unit test is run we can calculate it based on a fixed date and ensure that the calculation is correct. 

The unit test would look something like this:

```php
public function testGetAgeReturnsCorrectValues() {
    $referenceDate = new DateTime('2020-02-25');
    // For a candidate with birthday == '2019-01-01'
    assertsEquals($this->_candidate->getAgeInYears(), 1);
    assertsEquals($this->_candidate->getAgeInMonths(), 13);
    assertsEquals($this->_candidate->getAgeInDays(), 365 + 24);
}
```

johnsaigle: The current unit tests look good!

I did a bit research on unit testing functions that refer to the current date "now" which is a little tricky because unit tests rely on a static value. For example we would want to write a test that, given a candidate born on e.g. January 1st, 2019, the `getAge()*` functions would return 1 year, 13 months, or 365 + 24 days depending on what date interval you care about. (I'm writing this on Feb. 25th 2020). But if we hardcode that then we can't test for 365 + 24 days because that will change each day...

The way to approach this is by changing the function getAge() to this:

```php
    public function getAge($referenceDate = new DateTime()): DateInterval
    {
        $dob = new DateTime($this->candidateInfo["DoB"]);

        return $dob->diff($referenceDate);
    }
```

This allows us to choose a moment from which to measure the candidate's age. It means adding an optional parameter to getAge() but we don't have to use it in the actual codebase.

Doing this means that when the unit test is run we can calculate it based on a fixed date and ensure that the calculation is correct. 

The unit test would look something like this:

```php
public function testGetAgeReturnsCorrectValues() {
    $referenceDate = new DateTime('2020-02-25');
    // For a candidate with birthday == '2019-01-01'
    assertsEquals($this->_candidate->getAgeInYears(), 1);
    assertsEquals($this->_candidate->getAgeInMonths(), 13);
    assertsEquals($this->_candidate->getAgeInDays(), 365 + 24);
}
```

test/unittests/CandidateTest.php
```diff
@@ -658,6 +658,46 @@ public function testGetNextVisitNoReturnsOneWhenNextVisitDoesNotExist()
...
+     *
+     * @covers Candidate::getAgeInDays()
+     * @return void
+     */
+    public function testGetAgeInDaysReturnsDays()
+    {
+        $this->_setUpTestDoublesForSelectCandidate();
+        $this->_candidate->select($this->_candidateInfo['CandID']);
+
+        $this->assertEquals($this->_candidate->getAge()->days, $this->_candidate->getAgeInDays));
```
johnsaigle:
```suggestion
        $this->assertEquals($this->_candidate->getAge()->days, $this->_candidate->getAgeInDays());
```

Typo, I think

driusan:
Is this PR still necessary? I thought it was originally sent for the candidate profile, but the age in that is now done in javascript..

zaliqarosli:
@driusan hmm that's true. i still think it's useful to have available as candidate data. we could go even further and add it to the $candidate->candidateInfo array. there are a couple of projects now that prefer the display of age over DOB across loris

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ changes in the following format: PR #1234***
 - Very old instruments relying on QuickForm may have issues due to code changes (PR #4928)
 - Unix user permissions have been updated which may affect access to files. New
 documentation for file permissions has been added to the README.md file (PR #5323)
+- Candidate's age can be retrieved from the Candidate class in days, months, and years (PR #5945)
```
zaliqarosli: @driusan not sure about what kind of wording is necessary for this doc

ridz1208:
@zaliqarosli I fixed your conflicts

@driusan please re-review

php/libraries/Candidate.class.inc
driusan: I agree with the others, this is confusingly inconsistent with the other functions

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ changes in the following format: PR #1234***
 #### Updates and Improvements
 - Module-specific permissions added for Survey Accounts, Imaging Behavioural
 Quality Control, and Behavioural Quality Control. (PR #6041)
+- Candidate's age can be retrieved from the Candidate class in days, months, and years (PR #5945)
```
driusan, line 18:
```suggestion
- Candidate's age can be retrieved from the Candidate class in days, months, or years (PR #5945)
```

You only call one of them at a time.

driusan:
@zaliqarosli is this ready? It seems to be passing Travis

zaliqarosli:
@driusan i believe so!

PR #5946, opened 2020-01-21
[Candidate] Add get last visit subproject

zaliqarosli:
## Brief summary of changes

This PR adds a new function called `getLastVisitSubproject` in the candidate class. This new function replaces the function `getSubprojectIDUsingCandID` in Utility class.

`getSubprojectIDUsingCandID` is only currently being called in one place, the instrument_list class. The returned value there however affects nothing. It is passed as an argument to the function `getBatteryInstrumentList` that then does nothing with the parameter

1. Test instrument_list page on both master and this branch. Check that there are absolutely no changes.
2. Try to use getLastVisitSubproject and render the returned data somewhere on Loris and see the results

### DIFF
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -643,8 +643,35 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
 
         return $subprojList;
     }
+
     /**
-     * Returns first visit for a candidate based on Date_visit
+     * Returns the SubprojectID and subproject title of the candidate's last visit
+     * based on Date_visit
+     *
+     * @return array|null Array with values SubprojectID and title
+     *                    for the candidate's last visit or NULL if none exist
+     */
+    public function getSubprojectForMostRecentVisit(): ?array
+    {
+        $factory = NDB_Factory::singleton();
+        $db      = $factory->database();
+
+        $candID = $this->getCandID();
+        $query  = "SELECT SubprojectID, title FROM session
+          LEFT JOIN subproject USING (SubprojectID)
+          WHERE CandID=:cid AND Date_visit IS NOT NULL
+          ORDER BY Date_visit DESC LIMIT 1";
+        $where  = ['cid' => $candID];
+        $result = $db->pselect($query, $where);
+
+        if (empty($result)) {
+            return null;
+        }
+        return $result[0];
+    }
+
+    /**
+     * Returns first visit for a candidate based on VisitNo
      *
      * @return string The visit label for the candidate's first visit
      */
--- a/test/unittests/CandidateTest.php
+++ b/test/unittests/CandidateTest.php
@@ -605,12 +605,12 @@ class CandidateTest extends TestCase
      */
     public function testGetSubprojectForMostRecentVisitReturnsMostRecentVisitLabel()
     {
-        $subproject = array(
-            array(
+        $subproject = [
+            [
                 'SubprojectID' => 1,
-                'title'        => 'testSubproject' 
-            )
-        );
+                'title'        => 'testSubproject'
+            ]
+        ];
 
         $this->_setUpTestDoublesForSelectCandidate();
         $this->_candidate->select($this->_candidateInfo['CandID']);
@@ -619,17 +619,17 @@ class CandidateTest extends TestCase
             ->method('pselect')
             ->with(
                 $this->stringContains(
-                   "SELECT SubprojectID, title"
+                    "SELECT SubprojectID, title"
                 )
             )
             ->willReturn(
                 $subproject
             );
 
-        $expectedSubproject = array(
+        $expectedSubproject = [
             'SubprojectID' => 1,
             'title'        => 'testSubproject'
-        );
+        ];
 
         $this->assertEquals(
             $expectedSubproject,
@@ -638,27 +638,31 @@ class CandidateTest extends TestCase
     }
 
     /**
-     * Test getSubprojectForMostRecentVisit returns null if there is no visit with a Date_visit
+     * Test getSubprojectForMostRecentVisit returns null if there is
+     * no visit with a Date_visit
      *
      * @covers Candidate::getSubprojectForMostRecentVisit
      * @return void
      */
     public function testGetSubprojectForMostRecentVisitReturnsNull()
     {
-        $subproject = array();
+        $subproject = [];
         $this->_setUpTestDoublesForSelectCandidate();
         $this->_candidate->select($this->_candidateInfo['CandID']);
-        
+
         $this->_dbMock->expects($this->any())
             ->method('pselect')
             ->with(
                 $this->stringContains(
-                   "SELECT SubprojectID, title"
+                    "SELECT SubprojectID, title"
                 )
             )
             ->willReturn($subproject);
 
-        $this->assertEquals(null, $this->_candidate->getSubprojectForMostRecentVisit());
+        $this->assertEquals(
+            null,
+            $this->_candidate->getSubprojectForMostRecentVisit()
+        );
     }
 
     /**
--- a/test/unittests/CandidateTest.php
+++ b/test/unittests/CandidateTest.php
@@ -598,6 +598,74 @@ class CandidateTest extends TestCase
     }
 
     /**
+     * Test getSubprojectForMostRecentVisit returns most recent visit's label
+     *
+     * @covers Candidate::getSubprojectForMostRecentVisit
+     * @return void
+     */
+    public function testGetSubprojectForMostRecentVisitReturnsMostRecentVisitLabel()
+    {
+        $subproject = array(
+            array(
+                'SubprojectID' => 1,
+                'title'        => 'testSubproject' 
+            )
+        );
+
+        $this->_setUpTestDoublesForSelectCandidate();
+
+        $this->_dbMock->expects($this->any())
+            ->method('pselect')
+            ->with(
+                $this->stringContains(
+                   "SELECT SubprojectID, title 
+                    FROM subproject 
+                    WHERE SubprojectID = :subprj"
+                )
+            )
+            ->willReturn(
+                $subproject
+            );
+
+        $expectedSubproject = array(
+            'SubprojectID' => 1,
+            'title'        => 'testSubproject'
+        );
+
+        $this->_candidate->select($this->_candidateInfo['CandID']);
+        $this->assertEquals(
+            $expectedSubproject,
+            $this->_candidate->getSubprojectForMostRecentVisit()
+        );
+    }
+
+    /**
+     * Test getSubprojectForMostRecentVisit returns null if there is no visit with a Date_visit
+     *
+     * @covers Candidate::getSubprojectForMostRecentVisit
+     * @return void
+     */
+    public function testGetSubprojectForMostRecentVisitReturnsNull()
+    {
+        $subproject = array();
+        $this->_setUpTestDoublesForSelectCandidate();
+        
+        $this->_dbMock->expects($this->any())
+            ->method('pselect')
+            ->with(
+                $this->stringContains(
+                   "SELECT SubprojectID, title 
+                    FROM subproject 
+                    WHERE SubprojectID = :subprj"
+                )
+            )
+            ->willReturn($subproject);
+
+        $this->_candidate->select($this->_candidateInfo['CandID']);
+        $this->assertEquals(null, $this->_candidate->getSubprojectForMostRecentVisit());
+    }
+
+    /**
      * Test getFirstVisit returns first visit's label
      *
      * @covers Candidate::getFirstVisit
--- a/test/unittests/CandidateTest.php
+++ b/test/unittests/CandidateTest.php
@@ -613,14 +613,13 @@ class CandidateTest extends TestCase
         );
 
         $this->_setUpTestDoublesForSelectCandidate();
+        $this->_candidate->select($this->_candidateInfo['CandID']);
 
         $this->_dbMock->expects($this->any())
             ->method('pselect')
             ->with(
                 $this->stringContains(
-                   "SELECT SubprojectID, title 
-                    FROM subproject 
-                    WHERE SubprojectID = :subprj"
+                   "SELECT SubprojectID, title"
                 )
             )
             ->willReturn(
@@ -632,7 +631,6 @@ class CandidateTest extends TestCase
             'title'        => 'testSubproject'
         );
 
-        $this->_candidate->select($this->_candidateInfo['CandID']);
         $this->assertEquals(
             $expectedSubproject,
             $this->_candidate->getSubprojectForMostRecentVisit()
@@ -649,19 +647,17 @@ class CandidateTest extends TestCase
     {
         $subproject = array();
         $this->_setUpTestDoublesForSelectCandidate();
+        $this->_candidate->select($this->_candidateInfo['CandID']);
         
         $this->_dbMock->expects($this->any())
             ->method('pselect')
             ->with(
                 $this->stringContains(
-                   "SELECT SubprojectID, title 
-                    FROM subproject 
-                    WHERE SubprojectID = :subprj"
+                   "SELECT SubprojectID, title"
                 )
             )
             ->willReturn($subproject);
 
-        $this->_candidate->select($this->_candidateInfo['CandID']);
         $this->assertEquals(null, $this->_candidate->getSubprojectForMostRecentVisit());
     }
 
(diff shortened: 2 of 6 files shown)


php/libraries/Candidate.class.inc
```diff
@@ -613,8 +613,31 @@ class Candidate
...
     }
+
+    /**
+     * Returns the SubprojectID and subproject title of the candidate's last visit
+     * based on Date_visit
+     *
+     * @return array|null Array with values SubprojectID and title
+     *                    for the candidate's last visit or NULL if none exist
+     */
+    public function getLastVisitSubproject(): ?array
```
johnsaigle: This is a bit of a nit-pick but can you rename this to something like `getSubprojectForMostRecentVisit()`? I find the current name a bit unclear.

johnsaigle: I think a lengthy function name is the lesser of two evils compared with an ambiguous one

ridz1208:
Just to be 100% clear here, these are the limitations of that function:

- If the visit is created but not started, the date_visit would be empty and thus it would not be detected by this function. However a non-started visit can still be used to upload biospecimens, imaging and files.
- The lasted visit subproject is in no way more meaningful then the first of midway visit.
- The date_visit is not the only date for a session, when a visit gets approved, only the date_approval gets updated. The subproject of an approved visit might be as important as the subproject of the "lastest" one.

That being said, I don't care about this being merged if it helps the development of other modules.

zaliqarosli:
@ridz1208 what do you think about ordering by and using the VisitNo field instead of the Date_visit field? would that at least address the limitations 1 and 3 that you pointed out above?

ridz1208:
@zaliqarosli No i think the date_visit is more... "sturdy" then the auto incrementing visitNo

I dont really have a solution and thats why I'm not requesting changes or opposing the changes. just added the limitations on the PR for future records

PapillonMcGill:
I second @ridz1208 argument about date_visit. Auto-increment number can misled if not entered in chronological order.

php/libraries/Candidate.class.inc
zaliqarosli:
```suggestion
    public function getSubprojectForMostRecentVisit(): ?array
```

johnsaigle:
My feelings on this are the same as before. I think adding a unit test would be a good idea so that we can be confident that the function is working well.

> Code looks fine. I think it would be good to add a unit test for this function though. It seems fairly similar to the getFirstVisit() function so you can use the tests in test/unittests/CandidateTest.php as an example (testGetFirstVisitReturnsFirstVisitLabel() specifically, as well as the failure case)

php/libraries/Candidate.class.inc
```diff
@@ -606,8 +606,31 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
...
+        $db      = $factory->database();
+
+        $candID = $this->getCandID();
+        $query  = "SELECT SubprojectID, title FROM session
+          LEFT JOIN subproject USING (SubprojectID)
+          WHERE CandID=:cid AND Date_visit IS NOT NULL
+          ORDER BY Date_visit DESC LIMIT 1";
+        $where  = array('cid' => $candID);
+        $result = $db->pselect($query, $where);
+        return $result ?? null;
```
johnsaigle: I think Travis is having an issue with this part. `pselect()` returns the type array so I believe that if this query has no result, the function will return an empty array (instead of null). Your unit tests are doing assertions on `''`  which causes an error because the actual returned value is `array()`
```suggestion
        return $result;
```

test/unittests/CandidateTest.php
```diff
@@ -582,6 +582,44 @@ public function testGetValidSubprojectsReturnsEmptyArray(): void
...
+     * @return void
+     */
+    public function testGetSubprojectForMostRecentVisitReturnsEmptyString()
+    {
+        $this->_setUpTestDoublesForSelectCandidate();
+
+        $this->_dbMock->expects($this->any())
+            ->method('pselectOne')
+            ->with($this->stringContains("AND Date_visit IS NOT NULL"))
+            ->willReturn('');
```
johnsaigle:
```suggestion
            ->willReturn(array());
```

test/unittests/CandidateTest.php
```diff
@@ -582,6 +582,44 @@ public function testGetValidSubprojectsReturnsEmptyArray(): void
...
+        
+        $this->_candidate->select($this->_candidateInfo['CandID']);
+        $this->assertEquals('', $this->_candidate->getSubprojectForMostRecentVisit());
```
johnsaigle:
```suggestion
        $this->assertEquals(array(), $this->_candidate->getSubprojectForMostRecentVisit());
```

test/unittests/CandidateTest.php
```diff
@@ -582,6 +582,44 @@ public function testGetValidSubprojectsReturnsEmptyArray(): void
...
+        $this->assertEquals('V02', $this->_candidate->getSubprojectForMostRecentVisit());
+    }
+
+    /**
+     * Test getSubprojectForMostRecentVisit returns an empty string if there is no visit with a Date_visit
+     *
+     * @covers Candidate::getSubprojectForMostRecentVisit
+     * @return void
+     */
+    public function testGetSubprojectForMostRecentVisitReturnsEmptyString()
```
johnsaigle:
```suggestion
    public function testGetSubprojectForMostRecentVisitReturnsEmptyArray()
```

php/libraries/Candidate.class.inc
```diff
@@ -606,8 +606,31 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
...
+
     /**
-     * Returns first visit for a candidate based on Date_visit
+     * Returns the SubprojectID and subproject title of the candidate's last visit
+     * based on Date_visit
+     *
+     * @return array|null Array with values SubprojectID and title
+     *                    for the candidate's last visit or NULL if none exist
+     */
+    public function getSubprojectForMostRecentVisit(): ?array
```
```suggestion
    public function getSubprojectForMostRecentVisit(): array
```

php/libraries/Candidate.class.inc
```diff
@@ -606,8 +606,31 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
 
         return $subprojList;
     }
+
     /**
-     * Returns first visit for a candidate based on Date_visit
+     * Returns the SubprojectID and subproject title of the candidate's last visit
+     * based on Date_visit
+     *
+     * @return array|null Array with values SubprojectID and title
```
```suggestion
     * @return arrayArray with values SubprojectID and title
```

php/libraries/Candidate.class.inc
```diff
@@ -606,8 +606,31 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
...
+     *                    for the candidate's last visit or NULL if none exist
```
```suggestion
     *                    for the candidate's last visit or an empty array if none exist
```

zaliqarosli:
@johnsaigle thanks for pointing those out. i decided to maintain returning null as i find that more useful than an empty array as a return type, but tried to match the resulting types

test/unittests/CandidateTest.php
```diff
@@ -582,6 +582,61 @@ public function testGetValidSubprojectsReturnsEmptyArray(): void
...
+        );
+
+        $this->_setUpTestDoublesForSelectCandidate();
+
+        $this->_dbMock->expects($this->any())
+            ->method('pselect')
+            ->with($this->stringContains(
+                "SELECT SubprojectID, title FROM session
+                    LEFT JOIN subproject USING (SubprojectID)
+                    WHERE CandID=:cid AND Date_visit IS NOT NULL
```
johnsaigle, line 607: I think the unit tests might choke here on the prepared statements. `:cid` might get resolved to an actual value here making it so that the `stringContains()` check evalutes to false. It might work if you change this part to something more similar to what you had before, like just looking for `ORDER BY Date_visit DESC LIMIT 1` as this value cannot change.

zaliqarosli: i'm trying to figure out how to make the test spit out what i want. i'm not quite sure how the testing works. if you look at `testGetValidSubprojectsReturnsAListOfSubprojects()` on line 528, i'm trying to do something similar to match the returned array. do you have any idea why doing something similar doesn't work?

cmadjar:
@kongtiaowang can you help @zaliqarosli fix travis?

test/unittests/CandidateTest.php
```diff
@@ -602,7 +670,7 @@ public function testGetValidSubprojectsReturnsEmptyArray(): void
     public function testGetFirstVisitReturnsFirstVisitLabel()
     {
         $this->_setUpTestDoublesForSelectCandidate();
-
+
```
zaliqarosli, line 673:
```suggestion

```

test/unittests/CandidateTest.php
```diff
@@ -593,6 +593,74 @@ public function testGetValidSubprojectsReturnsEmptyArray(): void
...
+        
+        $this->_dbMock->expects($this->any())
+            ->method('pselect')
+            ->with(
+                $this->stringContains(
+                   "SELECT SubprojectID, title 
+                    FROM subproject 
+                    WHERE SubprojectID = :subprj"
+                )
+            )
```
kongtiaowang, line 657:
```suggestion
```

test/unittests/CandidateTest.php
```diff
@@ -593,6 +593,74 @@ public function testGetValidSubprojectsReturnsEmptyArray(): void
...
+
+        $this->_dbMock->expects($this->any())
+            ->method('pselect')
+            ->with(
+                $this->stringContains(
+                   "SELECT SubprojectID, title 
+                    FROM subproject 
+                    WHERE SubprojectID = :subprj"
+                )
+            )
```
kongtiaowang, line 621:
```suggestion
```

kongtiaowang:
I will reopen this PR to active the Travis.

kongtiaowang (dismissed):
LGTM

sruthymathew123 (dismissed):
Reviewed and tested as per the instructions given and all looks good.

php/libraries/Candidate.class.inc
```diff
@@ -638,8 +638,35 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
...
     }
+
+    /**
+     * Returns the SubprojectID and subproject title of the candidate's last visit
+     * based on Date_visit
+     *
+     * @return array|null Array with values SubprojectID and title
+     *                    for the candidate's last visit or NULL if none exist
+     */
+    public function getSubprojectForMostRecentVisit(): ?array
```
ridz1208, line 649: Remind me why you are creating this function but not using it anywhere ?? (it's completely removed from the instrument_list module)

zaliqarosli: i use it in QPN's candidate_profile page to determine which subproject the participant belongs to given their lastest visit. it replaces the Utility::getSubprojectIDUsingCandID function

ridz1208: That answers my question

driusan: then why doesn't the function live in the QPN's candidate_profile page if that's the only place it's used?

ridz1208: wouldnt it be used for the new LORIS candidate profile too ?
i thought it was based on QPN's

zaliqarosli:
@ridz1208 ready for re-review

php/libraries/Candidate.class.inc
driusan: no, it's not used by the new LORIS `candidate_profile`

test/unittests/CandidateTest.php
```diff
@@ -593,6 +593,74 @@ public function testGetValidSubprojectsReturnsEmptyArray(): void
...
+            ->willReturn(
+                $subproject
+            );
+
+        $expectedSubproject = array(
+            'SubprojectID' => 1,
+            'title'        => 'testSubproject'
+        );
+
+        $this->_candidate->select($this->_candidateInfo['CandID']);
```
kongtiaowang, line 631: Here , need a CandID type. not a array.

test/unittests/CandidateTest.php
```diff
@@ -593,6 +593,74 @@ public function testGetValidSubprojectsReturnsEmptyArray(): void
...
+            ->with(
+                $this->stringContains(
+                   "SELECT SubprojectID, title 
+                    FROM subproject 
+                    WHERE SubprojectID = :subprj"
+                )
+            )
+            ->willReturn($subproject);
+
+        $this->_candidate->select($this->_candidateInfo['CandID']);
```
kongtiaowang, line 660: Here , need a CandID type. not a array.

test/unittests/CandidateTest.php
kongtiaowang: the \n need to be removed in query.

test/unittests/CandidateTest.php
zaliqarosli: hi @kongtiaowang ,  could you add direct 'suggested changes' to the code? i'm not quite sure what I need to change here

test/unittests/CandidateTest.php
```diff
@@ -593,6 +593,74 @@ public function testGetValidSubprojectsReturnsEmptyArray(): void
...
+        $subproject = array();
+        $this->_setUpTestDoublesForSelectCandidate();
+        
+        $this->_dbMock->expects($this->any())
+            ->method('pselect')
+            ->with(
+                $this->stringContains(
+                   "SELECT SubprojectID, title 
+                    FROM subproject 
+                    WHERE SubprojectID = :subprj"
```
kongtiaowang, line 655:
```suggestion
                   "SELECT SubprojectID, title"
```

test/unittests/CandidateTest.php
kongtiaowang, line 660:
```suggestion
```

test/unittests/CandidateTest.php
```diff
@@ -593,6 +593,74 @@ public function testGetValidSubprojectsReturnsEmptyArray(): void
...
+     * Test getSubprojectForMostRecentVisit returns null if there is no visit with a Date_visit
+     *
+     * @covers Candidate::getSubprojectForMostRecentVisit
+     * @return void
+     */
+    public function testGetSubprojectForMostRecentVisitReturnsNull()
+    {
+        $subproject = array();
+        $this->_setUpTestDoublesForSelectCandidate();
+
```
kongtiaowang, line 648:
```suggestion
                $this->_candidate->select($this->_candidateInfo['CandID']);
```

test/unittests/CandidateTest.php
```diff
@@ -593,6 +593,74 @@ public function testGetValidSubprojectsReturnsEmptyArray(): void
...
+    {
+        $subproject = array(
+            array(
+                'SubprojectID' => 1,
+                'title'        => 'testSubproject' 
+            )
+        );
+
+        $this->_setUpTestDoublesForSelectCandidate();
+
```
kongtiaowang, line 612:
```suggestion
        $this->_candidate->select($this->_candidateInfo['CandID']);
```

test/unittests/CandidateTest.php
```diff
@@ -593,6 +593,74 @@ public function testGetValidSubprojectsReturnsEmptyArray(): void
...
+        $this->_dbMock->expects($this->any())
+            ->method('pselect')
+            ->with(
+                $this->stringContains(
+                   "SELECT SubprojectID, title 
+                    FROM subproject 
+                    WHERE SubprojectID = :subprj"
```
kongtiaowang, line 619:
```suggestion
                   "SELECT SubprojectID, title"
```

test/unittests/CandidateTest.php
kongtiaowang, line 631:
```suggestion
```

test/unittests/CandidateTest.php
zaliqarosli: hi @kongtiaowang can you explain what you mean by CandID type, not array?

kongtiaowang (approved):
These functions are only used in the QPN, I have no problem with it.

ridz1208:
@driusan it seems like this broke the main branch

ridz1208:
```
1) CandidateTest::testGetSubprojectForMostRecentVisitReturnsMostRecentVisitLabel
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    'SubprojectID' => 1
-    'title' => 'testSubproject'
+    'ID' => '97'
 )

/app/test/unittests/CandidateTest.php:636

2) CandidateTest::testGetSubprojectForMostRecentVisitReturnsNull
Array (...) does not match expected type "NULL".

/app/test/unittests/CandidateTest.php:664
```

PR #5947, opened 2020-01-21
label vs option vs attribute function misuses

lingma:
## Brief summary of changes

This is the recreation of the PR #5884 & #5886, at the same time this constitutes a rebase, which is necessary.

**label vs option vs attribute function misuses (PR #5884 part)**

Initially the third parameters of the form elements were set as the fourth parameters somehow in the program for `static` text Form type, this is clearly a bug. I found later that quite often the term `options` are mixed with the term `attributes`, or even `options` are passed to `label` which is what I stated early.

This will create confusion for programmer. Unfortunately since the same problem existed in the previous versions, migrations might become difficult once the patch has been applied.

I noticed that the  [wiki](https://github.com/aces/Loris/wiki/Instrument-Groups) manual suggests to use the format:
$group[] =& $this->form->createElement("static", null, null, "<label>");

The new format:

Some old instruments will stop displaying correctly once being migrated, the verification required besides some function signatures changes, including setup, score which is beyond this PR scope.

The instrument migration fix could be easy with a full text replacement, with verification. Some project has module overrides, everything could be easy in this case.

This PR has also a little doc modification and the removal of the variable `$closer`, the latter is not too much related with the main task.

**undefined name warning for label element (PR #5886 part)**

A instrument with
$this->addLabel('<label>') will always report PHP warning "Undefined index: name in LorisForm.class.inc on line 1242"

This happens because Loris form assumes that a name is mandatory, even for a label. I don't see the necessity here.

For the undefined name warning for label element issue:

1.  Create an instrument, or a unit test case, put a label $this->addLabel('<label>') anyway in an instrument, check whether the PHP warning is in the log.

### DIFF
--- a/php/libraries/LorisForm.class.inc
+++ b/php/libraries/LorisForm.class.inc
@@ -116,8 +116,7 @@ class LorisForm
      * @param string $name    The element name.
      * @param string $label   The label to attach to this element.
      * @param array  $attribs An array of other attributes that should
-     *                        get added. Currently only the "class"
-     *                        attribute gets added.
+     *                        get added, for example "class" attribute.
      *
      * @return &array A reference to the array that was added to $this->form
      */
@@ -137,8 +136,7 @@ class LorisForm
      * @param array  $options An array of the options (values) to add
      *                        to this select dropdown.
      * @param array  $attribs An array of other attributes that should
-     *                        get added. Currently only the "class"
-     *                        attribute gets added.
+     *                        get added, for example "class" attribute.
      *
      * @return &array
      */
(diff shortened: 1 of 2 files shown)


lingma:
VM down, there is some test case failed.

johnsaigle:
Travis is failing with the following error (although it's not showing up on this page for some reason):

1) TestInstrumentTestIntegrationTest::testCheckBoxElement
Failed asserting that '{"testCheckbox":"on","testText":null,"consent":null,"Data_entry_completion_status":"Complete"}' contains ""testCheckbox":"1"".
/app/test/test_instrument/test_instrumentTest.php:155

I think you'll need to either adapt your code or update the test.

lingma:
> I think you'll need to either adapt your code or update the test.

I know, thanks, my VM is still down. I could not make anything now.

johnsaigle:
No problem, it doesn't have to be now. My VM is down too.

lingma:
I wanted to keep `options` and add an `attribute` for `addCheckbox`, which proved not working. This time all should be OK since they represent almost the modification of the previous PRs of the last time, still waiting Travis to finish.

lingma:
This PR should be considered or examined because the initial implementation has some presumed error introduced. Early or later the error will be found out by the users. Of course, the risk to the existing instruments is real.

lingma:
>  like `createElement()` will affect LORIS.

Because `createElement()` is already in use in instruments, the position change means that users need to check affected instruments, not Loris itself.

lingma:
@driusan @ridz1208 I feel that this PR should be included in the 23.0. The LorisForm is weakly written, currently I am modifying LorisForm for another project, it is really painful to see this class.

PR #5949, opened 2020-01-22
[NDB_BVL_Instrument] Data entry completion progress function

zaliqarosli:
## Brief summary of changes

This PR adds a function that determines an instrument's completion progress based on how many of its required fields have been filled out. This new function evolved from the already existing `_determineDataEntryCompletionStatus` function.

This function is being used in the new candidate profile page as implemented by PR https://github.com/aces/Loris/pull/6121

1. Test https://github.com/aces/Loris/pull/6121 to see if the function determineDataEntryCompletionProgress() is doing what it's intending to do.
2. On that branch, visit `https://$LORISURL/candidate_profile/$candID` and click on one of the timepoint boxes in the 'Behavioural Data' card to expand the instrument list for that timepoint with corresponding completion progress bar
3. Select a fresh blank instrument (completion progress 0%) and fill it out page by page. After every page, check the instrument's completion progress display to see that it matches e.g. if 2/5 pages have been filled out and saved, the completion progress is 40%.
4. At the same time, test the control panel workflow outlined in https://github.com/aces/Loris/pull/6129. The workflow should remain the exact same.
5. Fill out the whole instrument and mark the Data Entry as 'Complete'. Test that you are able to do this even if you set every field to 'Not Answered'.

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1543,8 +1543,8 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             $statusField = $field . '_status';
             // field shouldn't be just empty() b/c 0 is a valid
             // value for some required fields
-            if ((($allData[$field] ?? null) === null)
-                || (($allData[$field] ?? "") === "")
+            if (((($allData[$field] ?? null) === null)
+                || (($allData[$field] ?? "") === ""))
                 && empty($allData[$statusField])
             ) {
                 $unanswered++;
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1523,6 +1523,40 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         return null;
     }
 
+    /**
+     * Determines what percentage of the data entry has been completed, based on the
+     * number of fields in $_requiredElements having data
+     *
+     * @return int the data entry completion percentage
+     */
+    function determineDataEntryCompletionProgress(): int
+    {
+        // don't bother checking anything if the required elements array is empty
+        if (empty($this->_requiredElements)) {
+            return 100;
+        }
+        $allData       = $this->loadInstanceData($this);
+        $unanswered    = 0;
+        $totalElements = count($this->_requiredElements);
+        foreach ($this->_requiredElements as $field) {
+            // consider status field if not_answered chosen as an answer
+            $statusField = $field . '_status';
+            // field shouldn't be just empty() b/c 0 is a valid
+            // value for some required fields
+            if ((is_null($allData[$field]) || $allData[$field] === "")
+                && empty($allData[$statusField])
+            ) {
+                $unanswered++;
+            }
+        }
+        if ($unanswered == 0) {
+            return 100;
+        } else if ($unanswered == $totalElements) {
+            return 0;
+        }
+
+        return (int)round((($totalElements-$unanswered)/$totalElements)*100);
+    }
 
     /**
      * Determines what the data entry status flag should be set to
@@ -1538,9 +1572,13 @@ abstract class NDB_BVL_Instrument extends NDB_Page
 
         $allData = $this->loadInstanceData($this);
         foreach ($this->_requiredElements as $field) {
-            // this shouldn't be just empty() b/c 0 is a valid
+            // consider status field if not_answered chosen as an answer
+            $statusField = $field . '_status';
+            // field shouldn't be just empty() b/c 0 is a valid
             // value for some required fields
-            if (is_null($allData[$field]) || $allData[$field] === "") {
+            if ((is_null($allData[$field]) || $allData[$field] === "")
+                && empty($allData[$statusField])
+            ) {
                 return 'Incomplete';
             }
         }
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1543,7 +1543,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             $statusField = $field . '_status';
             // field shouldn't be just empty() b/c 0 is a valid
             // value for some required fields
-            if ((is_null($allData[$field]) || $allData[$field] === "")
+            if ((($allData[$field] ?? null) === null) || (($allData[$field] ?? "") === "")
                 && empty($allData[$statusField])
             ) {
                 $unanswered++;
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1543,7 +1543,8 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             $statusField = $field . '_status';
             // field shouldn't be just empty() b/c 0 is a valid
             // value for some required fields
-            if ((($allData[$field] ?? null) === null) || (($allData[$field] ?? "") === "")
+            if ((($allData[$field] ?? null) === null)
+                || (($allData[$field] ?? "") === "")
                 && empty($allData[$statusField])
             ) {
                 $unanswered++;
--- a/php/libraries/NDB_BVL_Instrument_LINST.class.inc
+++ b/php/libraries/NDB_BVL_Instrument_LINST.class.inc
@@ -415,8 +415,8 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
                 $addElements = true;
             }
 
-            $parsingPage       = 'top';
-            $firstSelectOfPage = true;
+            $parsingPage      = 'top';
+            $firstFieldOfPage = true;
 
             $Group = array(
                 'Name'      => null,
@@ -438,6 +438,10 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
                     || $fieldname == "Window_Difference"
                     || $fieldname == "Examiner"
                 ) {
+                    if ($firstFieldOfPage) {
+                        $this->_requiredElements[] = $fieldname;
+                        $firstFieldOfPage          = false;
+                    }
                     continue;
                 }
                 switch($type) {
@@ -457,7 +461,7 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
                     } else {
                         $addElements = false;
                     }
-                    $firstSelectOfPage = true;
+                    $firstFieldOfPage = true;
                     break;
                 case 'table':
                     $this->testName = trim($pieces[1]);
@@ -531,6 +535,10 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
                             $this->addTextElement($pieces[1], $pieces[2]);
                         }
                     }
+                    if ($firstFieldOfPage) {
+                        $this->_requiredElements[] = $fieldname;
+                        $firstFieldOfPage          = false;
+                    }
                     $this->LinstQuestions[$pieces[1]] = array('type' => 'text');
                     break;
                 case 'textarea':
@@ -544,6 +552,10 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
                         } else {
                             $this->addTextAreaElement($pieces[1], $pieces[2]);
                         }
+                    }
+                    if ($firstFieldOfPage) {
+                        $this->_requiredElements[] = $fieldname;
+                        $firstFieldOfPage          = false;
                     }
                     $this->LinstQuestions[$pieces[1]] = array('type' => 'textarea');
                     break;
@@ -597,11 +609,19 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
                             );
                         }
                     }
+                    if ($firstFieldOfPage) {
+                        $this->_requiredElements[] = $fieldname;
+                        $firstFieldOfPage          = false;
+                    }
                     $this->LinstQuestions[$pieces[1]] = array('type' => 'date');
                     break;
                 case 'numeric':
                     if ($addElements) {
                         $this->addNumericElement($pieces[1], $pieces[2]);
+                    }
+                    if ($firstFieldOfPage) {
+                        $this->_requiredElements[] = $fieldname;
+                        $firstFieldOfPage          = false;
                     }
                     // FIXME: This variable is never populated.
                     $options = array();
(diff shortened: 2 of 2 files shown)


driusan:
I agree with @maltheism about the else statement.. I think it's looks a lot cleaner if functions always end with an unindented `return`.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1499,6 +1499,35 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         return null;
     }
 
+    /**
+     * Determines what the data entry completion progress percentage is
```
driusan: What is the point of this function being added?

zaliqarosli: for new candidate profile page (completion column) 
Screenshot 2020-01-28 12 52 05

driusan:
This is failing Travis

php/libraries/NDB_BVL_Instrument.class.inc
driusan: Can you update this comment to explain what the definition of "completion progress percentage" is in English so that it can be verified that that's what the code is doing? Also, there's a period missing at the end of the sentence.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1500,9 +1500,10 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     }
 
     /**
-     * Determines what the data entry completion progress percentage is
+     * Determines what pecentage of the data entry has been completed, based on the
+     * number of required fields having data
```
zaliqarosli: @driusan is this better?

driusan: How is a "required field" determined?

zaliqarosli: im not sure. i used the same "required fields" as used in the `_detemineDataEntryCompletionStatus()` function, which looks at $this->_requiredElements.

the function i'm proposing in this PR grabs the exact same data as in `_detemineDataEntryCompletionStatus()`. but instead of a boolean 'Complete' or 'Incomplete' , it returns the percentage of completeness. the results are the same, just with more detail.

i don't understand why this is sketchy if there's already a pretty important function that we're already using to determine data entry completeness that calculates via the same way

driusan: The `_determineDataEntryCompletionStatus` function is only to determine if all data entry is complete. It's only written to assume a binary "Yes" or "No" answer. It requires that there is at least one field in `_requiredElements` per page in order to force the triggering of rules (which prevents saving the whole page), but not all required elements are in `_requiredElements`.

zaliqarosli: hmmm so in that case, isn't it also a problem that data entry is being set as 'Complete' even though not all the required elements have been answered, and thus is not actually complete? i don't understand how `_determineDataEntryCompletionStatus` determines if all entry is complete if it only looks at the one field in `_requiredElements` and not all required elements in the instrument

driusan: No, it can't be set to 'Complete' because an invalid rule (XINRule, QuickForm rule, whatever) prevents the page from being saved if anything is invalid on the page.

driusan: I'm not saying there's anything wrong with using the field, I'm saying the comment needs to be clear about what it considers 'Complete Percentage' to avoid confusion in the future.

zaliqarosli: ahh i see, so the function checks any leftover required fields not taken care of by the other rules. i'll need to look into this more. i can tag this PR as 'Needs Work'. ideally, all required fields should be considered in this completion progress function

zaliqarosli: @driusan thanks for explaining this!

driusan: I don't think it necessarily needs work if it's clear, but in most cases it's probably giving you the percentage of pages complete, not the percentage of questions.

zaliqarosli: okayy that sounds good. as long as for linst instruments, it considers at least one field per page (which I think it does), then i'm happy to keep it this way

zaliqarosli: this is not working as expected for LINST instruments unfortunately. the test instrument i have returns 100% even if all fields are empty

zaliqarosli:
@driusan i made changes to LINST class to allow requiredElements to be populated for more than just select elements. please review my last commit!

driusan:
@zaliqarosli sorry I didn't see your comment for some reason..

I don't think the linst changes will work. The reason it was using the first select on the page is because the first field comprises of 2 HTML elements otherwise, the "not_answered" dropdown and the textbox (or whatever) for data entry if it has data entered. Have you verified that these changes don't break the required elements and claim that required elements are missing when "not answered" is chosen? This isn't an issue for select dropdowns because "not answered" is an answer in the same dropdown in that case.

zaliqarosli:
@driusan my last commit fixes the issue of having "not_answered" dropdown option

driusan:
@zaliqarosli I think you've likely fixed it for the percentage calculating, but I was more concerned about the changes affecting`_determineDataEntryCompletionStatus` and the places that it's used in LORIS.. can you confirm that the data entry administration flags aren't affected by this, and that saving still works and doesn't claim that you're missing required fields when there's a `_status` (for both the combination of `_status='not_answered'` and the `field='answered'`

zaliqarosli:
@driusan yep, the same issue was happening for `_determineDataEntryCompletionStatus`. if the first field of any subtest has "not_answered", the `Data_entry_completion_status` value is 'Incomplete'. I can fix this if I add the same `empty($allData[$statusField]` check in that function.

are there any other data entry administration flags besides the `Data_entry_completion_status` field?

saving still works though, no error is thrown if the first field in a subtest is "not_answered", i.e., you are able to save a subtest will all as not_answered.

driusan:
There's the Administration All/Partial/None flags in the instrument control panel. You can't click Administration "All" unless all the required fields are entered.

zaliqarosli:
@driusan on aces/master, i am able to click on "All" even though none of the fields, including Examiner, have data entered. did you maybe instead mean "You can't click Data Entry  "Complete" unless all the required fields are entered"? Even with "Complete" though, I'm finding that I'm able to click on it with no data entered if Administration is set to None or Partial. If Administration is set to "All", I can't click Data Entry "Complete". This is all on master. What's the intended workflow?

driusan:
@zaliqarosli The expected behaviour is that clicking "Administration: All" doesn't save unless all the `_requiredFields` are answered, I think.

zaliqarosli:
@driusan there appears to be a bug on master then. "Administration: All" does indeed save in an empty instrument with no data entered. i'll bring it up at the loris meeting to determine the full expected behaviour.

driusan:
@zaliqarosli can you check the latest release too?

zaliqarosli:
@driusan it exists on 22.0-release and 21.0-release

driusan:
Moving this to 23.0 because it's blocking a PR going there..

johnsaigle:
@zaliqarosli Happy to review if you want. Can you add testing instructions?

zaliqarosli:
@driusan i have just checked that this PR does not break the instrument flags workflow as described in the description of https://github.com/aces/Loris/pull/6129

zaliqarosli:
@johnsaigle that would be great! i just updated the testing instructions in the PR description. thanks :)

driusan:
@johnsaigle did this pass manual tests?

sruthymathew123:
@driusan @cmadjar  Got into some issues while switching from Loris 21.0 branch to 23.0 branch.
Will sort out soon and test it today.

sruthymathew123:
@cmadjar Hi Cecile, This PR was assigned to me as a Loris Monday Task. But I am having trouble testing this PR on my VM as I need to upgrade my VM which is on ubuntu 14.04 to support php 7.3. (Working on the upgrade with sysadmin)

But if it is needed to test on the sooner side, please reassign to someone else.

driusan:
@nicolasbrossard can you test this?

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1523,6 +1523,40 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+        }
+        $allData       = $this->loadInstanceData($this);
+        $unanswered    = 0;
+        $totalElements = count($this->_requiredElements);
+        foreach ($this->_requiredElements as $field) {
+            // consider status field if not_answered chosen as an answer
+            $statusField = $field . '_status';
+            // field shouldn't be just empty() b/c 0 is a valid
+            // value for some required fields
+            if ((is_null($allData[$field]) || $allData[$field] === "")
```
```suggestion
            if ((($allData[$field] ?? null) === null) || (($allData[$field] ?? "") === "")
```

When using this in the other PR, it generated index undefined errors for some fields which resulted in invalid JSON being returned. This fixes it.

sruthymathew123:
@zaliqarosli I cannot see the percentage completed on testing this PR along with Dave's PR.  
I personally feel like the instructions are confusing  to a person who is testing the PR and the UI is not matching with what Dave is having in his PR #6121 

Can you update some screenshots compatible with Dave's PR; then it will be useful to cross check.

zaliqarosli:
@driusan would you be able to add the actual percentage value on your UI? sorry @sruthymathew123 i wasn't aware that the UI had changed from how I had implemented it. This PR shouldn't worry about the UI though. This PR should just test that the function does what it is supposed to do, however way it is being displayed i.e. if half of the instrument is complete, the progress tracker should show half way. I will update the testing instructions to make it simpler

nicolasbrossard:
@driusan I see that Sruthy is testing this: do you still need me to test it also?

ridz1208:
@nicolasbrossard no, we just need 1 person on IBIS testing it

driusan:
@zaliqarosli I tried, but I couldn't find any way to make an HTML `<progress />` element display the text value or any way to put it in plain text around it that didn't look hideous.

I don't think that PR is the right way to test this anyways, since that would be testing the PR's implementation (which for all I know is broken from a rebase or not up to date), not the one here. This PR is blocking that one, not the other way around..

Here's a simple script which takes a commentID as an argument and calls the function with it that can be used for testing this PR instead:

```
<?php

require_once 'generic_includes.php';

if ($argc < 2) {
    die("Need commentID argument");
}
$commentID=$argv[1];
$testName = $DB->pselectOne("SELECT test_name FROM flag WHERE CommentID=:cid", [ 'cid' => $argv[1]]);
$instrument = \NDB_BVL_Instrument::factory(
                $testName,
                $commentID,
                '',
                false
            );
print "Completion is " . $instrument->determineDataEntryCompletionProgress();
```

zaliqarosli:
@sruthymathew123 could you test using the script dave gave above?

zaliqarosli:
@lingma for the issue outside of this PR regarding public/private declaration, please make a github issue about it because as you said, that is outside the scope of this PR.

about the `$_requiredElements` variable being empty or in fact populated in order to be useful, not all required elements need to be listed in the variable in order for the completion progress function to work. from what i understand, the variable is populated with at least 1 field per page of an instrument (usually the first field on that page) which is enough to calculate the completion of an instrument on a per page basis. 

calculating on a page by page basis is fine because if a required field has not been filled out, an error will pop up, disallowing you to save that page. a page is required to pass validation and be saved in order for the completion progress function to be called. so, in short, as long as the `$_requiredElements` array is populated with at least 1 field per page, this function works. if that one required field is not filled out, it's true that the page won't pass validation and will not save.

i'm not sure how the IBIS instruments are set up, but from looking at a few, it looks to me that you guys do define the requiredElements array.

does that make sense? let me know if there are still issues on your side

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1523,6 +1523,41 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+        $totalElements = count($this->_requiredElements);
+        foreach ($this->_requiredElements as $field) {
+            // consider status field if not_answered chosen as an answer
+            $statusField = $field . '_status';
+            // field shouldn't be just empty() b/c 0 is a valid
+            // value for some required fields
+            if ((($allData[$field] ?? null) === null)
+                || (($allData[$field] ?? "") === "")
+                && empty($allData[$statusField])
+            ) {
```
lingma: This condition looks wrong, maybe
`((($allData[$field] ?? null) === null
                || ($allData[$field] ?? "") === "")
                && empty($allData[$statusField]))`
? The current equals: `null || null && empty`, maybe you want `(null || null) && empty`

lingma:
> does that make sense? let me know if there are still issues on your side

The value returned is always an estimation instead of a real integer percentage, I feel that the percentage calculating tends to be confusing. Yes, the solution seems OK.

php/libraries/NDB_BVL_Instrument.class.inc
zaliqarosli:
```suggestion
            if (((($allData[$field] ?? null) === null)
                || (($allData[$field] ?? "") === ""))
                && empty($allData[$statusField])
            ) {
```

zaliqarosli:
> > does that make sense? let me know if there are still issues on your side
> 
> The value returned is always an estimation instead of a real integer percentage, I feel that the percentage calculating tends to be confusing. Yes, the solution seems OK.

why do you say its always an estimation? the function should return a real integer percentage. could you give me an example of what you're testing that makes the percentage calculating seem confusing?

lingma:
> @lingma why do you say its always an estimation? the function should return a real integer percentage. could you give me an example of what you're testing that makes the percentage calculating seem confusing?

The calculating itself is fine, just the solution is a little confusing to me. For example, I could have a form with all except "Examiner" to be required, which I did not input anything, except that the form has only "Examiner" filled, the solution could report 100% complete, which does not represent the reality that we expect at least something filled other than "Examiner". It is not practical to declare all fields as required and not possible because we don't know which part should be required.

As such, the solution is only an estimation, the calculation is to try to estimate. I think the documentation for this part should be clear.

zaliqarosli:
> > @lingma why do you say its always an estimation? the function should return a real integer percentage. could you give me an example of what you're testing that makes the percentage calculating seem confusing?
> 
> The calculating itself is fine, just the solution is a little confusing to me. For example, I could have a form with all but "Examiner" to be not required, then I could have a form with nothing except that the form has only "Examiner" filled. This solution could report 100% complete, which does not represent the reality that we expect at least something filled other than "Examiner". It is not practical to declare all fields as required and not possible because we don't know which part should be required.
> 
> As such, the solution is only an estimation, the calculation is to try to estimate. I think the documentation for this part should be clear.

@lingma yess, that's exactly how instrument completion is determined in LORIS. outside of this PR, if an instrument with 1 page has only 1 field completed i.e. the Examiner field, given there are no other required fields, the instrument would pass validation, be saved, and marked as complete. Hence, the solution is not an estimation, it is a calculation of page by page. An instrument can indeed be complete even if some fields are left empty. A clear case would be that a data entry person will want to see an instrument 100% complete when they intentionally choose to leave fields empty as part of their data entry (to show that this instrument has its data entry complete, and the person can move on to the next one, or submit this instrument to the DCC)

PR #5952, opened 2020-01-24
[React] New Card component

zaliqarosli:
## Brief summary of changes

very simple. essentially just uses the already existing Panel component but adds onclick functionality

### DIFF
--- a/jsx/Card.js
+++ b/jsx/Card.js
@@ -1,0 +1,62 @@
+/**
+ * This file contains React component for Card
+ *
+ * @author Zaliqa Rosli
+ * @version 1.0.0
+ *
+ */
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+
+import Panel from 'jsx/Panel';
+
+/**
+ * Card component - Wraps children in a customizable card
+ *
+ */
+class Card extends Component {
+  constructor(props) {
+    super(props);
+    this.handleClick = this.handleClick.bind(this);
+  }
+
+  handleClick(e) {
+    this.props.onClick(e);
+  }
+
+  render() {
+    const cursorStyle = this.props.onClick ? {
+      cursor: 'pointer',
+    } : null;
+    return (
+      <div style={{marginLeft: '5px', marginRight: '5px'}}>
+        <Panel
+          id={this.props.id}
+          title={this.props.title}
+          initCollapsed={this.props.initCollapsed}
+        >
+          <div
+            onClick={this.handleClick}
+            style={cursorStyle}
+          >
+            {this.props.children}
+          </div>
+        </Panel>
+      </div>
+    );
+  }
+}
+
+Card.propTypes = {
+  id: PropTypes.string.isRequired,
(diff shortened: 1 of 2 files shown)


driusan:
I'm a little confused by the 'onClick'.. isn't there already an onClick? It seems to just propagate it down.

zaliqarosli:
@driusan the Panel component has an onClick for expanding/collapsing the panel. the 
Card component's onClick activates when you click on the body of the card e.g.  If you click on the 'Imaging' Card, it can take you to the Imaging browser

jsx/Card.js
```diff
@@ -0,0 +1,62 @@
...
+  handleClick(e) {
+    this.props.onClick(e);
+  }
+
+  render() {
+    const cursorStyle = this.props.onClick ? {
+      cursor: 'pointer',
+    } : null;
+    return (
+      <div style={{marginLeft: '5px', marginRight: '5px'}}>
```
maltheism: Hey @zaliqarosli, I believe the best practice is to have the style object for marginLeft & marginRight defined outside the div as a constant.

HenriRabalais:
I'm a little confused by this as well. If the only change is adding an onClick to the Panel body, couldn't this be a new prop that can be passed to the Panel component?

zaliqarosli:
@HenriRabalais the Card component is essentially a wrapper for the Panel class with css on top of the onClick change so that they can be rendered next to each other nicely. I could have just used the Panel class and added a margin outside each one but I thought this was cleaner. I also thought that as we move forward with our UI/UX road map, there might be a need for a separate Card component to the Panel one. In my opinion, the two components provide different services so deserve their own class. It just so happens that the Panel class already had what I was looking for UI wise so I reused it! :)

driusan:
Failing Travis

jsx/Card.js
```diff
@@ -0,0 +1,66 @@
...
+            {this.props.children}
+          </div>
+        </Panel>
+      </div>
+    );
+  }
+}
+
+Card.propTypes = {
+  id: PropTypes.string.isRequired,
```
driusan: I'm trying to use this component and I don't see why `id` is required. It doesn't seem to be declared are required by the Panel that it's being passed to, so why add this requirement for cards?

zaliqarosli: Id is required for collapsing the intended panel. If all the cards on one page have panel component with id 'default-panel', trying to collapse any one card will always collapse the first card

driusan: I don't understand why id is required for collapsing the panel, the collapsed state is in the state of the panel. The fact that it's required and if it's not passed or distinct the wrong panel is collapsed is why I find it confusing that you need to pass an id when trying to use this card.

zaliqarosli: The code for collapsing the panel is in the panel component and it targets an element with the given id. If there are multiple elements with the same Id i.e. the default one, then the collapsing function can't select the correct element

driusan: Okay, I guess that's a separate issue that should be fixed outside of this PR.

PR #5953, opened 2020-01-24
[jsx/Panel] Make Panel title Bold

zaliqarosli:
## Brief summary of changes

This PR does the following:
- makes title of Panel **bold** (let me know what you think about this. i find making it bold lets it stand out from other text on the page i.e. for if there are multiple Panels rendered such as in the new candidate profile module that has multiple cards)
- add missing proptypes and default proptype values (cleanup)

1. run `npm run compile` , checkout Examiner module for now bold title 'Examiner' at the top of the Panel

### DIFF
--- a/jsx/Panel.js
+++ b/jsx/Panel.js
@@ -79,7 +79,7 @@ class Panel extends Component {
           id={this.props.id}
           className={this.panelClass}
           role="tabpanel"
-          style={{height: 'calc(100% - 3em)', overflowX: 'auto'}}
+          style={{height: 'calc(100% - 3em)'}}
         >
           <div className="panel-body"
                style={{...this.props.style, height: this.props.height}}>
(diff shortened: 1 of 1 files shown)


johnsaigle:
The bold looks nice! Wasn't able to verify the horizontal scrolling. I tried shrinking my browser horizontally but a scroll bar didn't appear for the Examiner Panel (apart from the table).

zaliqarosli:
@johnsaigle if you horizontally minimize your browser as much as you can and zoom in to 200%, the horizontal scrolling might activate for you

driusan:
can you include a screenshot?

zaliqarosli:
before:
Screenshot 2020-02-07 16 02 38Screenshot 2020-02-07 16 02 56

after:
Screenshot 2020-02-07 16 01 28Screenshot 2020-02-07 16 01 40

jsx/Panel.js
```diff
@@ -64,7 +64,7 @@ class Panel extends Component {
         data-target={'#' + this.props.id}
         style={{cursor: 'pointer', height: '3em'}}
       >
-        {this.props.title}
+        <b>{this.props.title}</b>
```
driusan, line 67: this should be done with CSS, not `<b>`. (There's already a style attribute on the div, you just need to add font-weight to it..)

PR #5954, opened 2020-01-24
[SQL] Add register_new_scanner Config setting for the imaging pipeline

cmadjar:
## Brief summary of changes

This creates a config setting for the imaging backend pipeline to remove hardcoded configuration of 'new_scanner' in the backend script. This should be a config option instead of some hardcoded obscure code in the backend. 

#### Testing instructions 

1. Test that the SQL queries works

#### Caveat for Existing Projects

Make sure to update your SQL database with the release patch and to select whether you want the imaging pipeline to create a new scanner when an uploaded DICOM dataset was acquired on a new scanner. 
Note: 'Yes' was the default behaviour of the imaging pipeline, which is the default set in the release patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,4 +15,9 @@ changes in the following format: PR #1234***
 - Menus are now maintained by modules and no longer in the SQL database (PR #5839)
 
 #### Modules 
-##### module1
+
+##### Configuration
+
+- Make sure to update your SQL database with the release patch and to select whether you want the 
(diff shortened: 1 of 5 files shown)


johnsaigle:
Since you tagged this as `Add To Release Notes`, can you update the CHANGELOG file?

cmadjar:
@johnsaigle @driusan oooooh, this is nice! I did not know that existed. Love it!

Anyway, change log updated :)

cmadjar:
In the end, no need for a config as discussed at the imaging meeting of January 31st

PR #5956, opened 2020-01-27
[User Accounts] Fix edge-case that gave a confusing error message

### LINKED ISSUE #5948: [user_accounts] new user
**Describe the bug**
User accounts add new user error message is confusing in some case. If you don't enter password, but click "Generate new password", without any email address, the system will say:  "Your password cannot be your email."

**To Reproduce**
1. Go to 'Admin/User Account'
2. Click on 'Add user'
3. enter a user, click "Generate new password", but without email address
4. See error on password field "Your password cannot be your email. "

The message "Your password cannot be your email. " is really confusing

- LORIS Version: [e.g. 21.0.0] 22.0

johnsaigle:
## Brief summary of changes

When both the password and email fields were empty, an error was displayed saying "Your password cannot match your email address"

While technically accurate, this isn't actually helpful for a user, especially since our checkbox option "Generate new password" requires leaving the password field blank.

1. Verify the confusing error message in the linked Issue.
2. On this branch, do the same thing. The password error won't appear.
3. Repeat 2. but uncheck 'Generate new password'. Ensure that the error message "Please specify password or click Generate new password" is displayed instead.

* Resolves #5948

### DIFF
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -1085,7 +1085,14 @@ class Edit_User extends \NDB_Form
         //==================================
         //        Password validation
         //==================================
-        if ($values['Email'] === $values['Password_hash']) {
+
+        // Make sure the user is not using their email address as their password.
+        // Do not show this error if the password is an empty string: in this
+        // case, that means the email is empty also. It's more appropriate to
+        // display only the error 'You must enter an email'.
+        if ($values['Email'] === $values['Password_hash']
+            && $values['Password_hash'] !== ''
+        ) {
             $errors['Password_Group'] = self::PASSWORD_ERROR_IS_EMAIL;
         }
         if (!is_null($this->identifier)) {


johnsaigle:
@lingma Would you mind reviewing?

lingma:
What about `(!empty($values['Password_hash']) && in_array($values['Email'], $values['Password_hash']))`?

johnsaigle:
Are you asking what happens if the Email equals the password? There should be an error condition for that. You can test and see.

lingma:
> Are you asking what happens if the Email equals the password? There should be an error condition for that. You can test and see.

No, I am wondering what if the password is `email`0

lingma:
The proposed code could solve this problem. But I did not test.

johnsaigle:
> No, I am wondering what if the password is email0

I don't understand what problem you are trying to address here.

lingma:
I am wondering what if the password is the same as the Email, only add a `0` or whatever, the program won't detect anything.

johnsaigle:
The `Password` class is self-validating and contains a reference to the [Zxcvbn](https://github.com/dropbox/zxcvbn) library which evaluates password-strength. We're using this library to check passwords instead of creating ad-hoc password rules for LORIS.

driusan:
I just had a look at the code, and the Zxcvbn only seems to check the complexity of the input in the Password class. It doesn't validate that the blacklisted things that the library doesn't know about (ie. the email address) aren't a substring as far as I can tell.

johnsaigle:
That's true; it's not a feature we discussed. The code has only ever checked if the Password === Email as far as I know. 

We can add that in another PR?

driusan:
Sure, I just meant that that's what @lingma seems to be saying that you're misinterpreting..

PR #5963, opened 2020-01-27
[Issue Tracker] Remove remaining references to LorisMenu table

### LINKED ISSUE #5938: [issue_tracker] error when editing issue
**Describe the bug**
The module issue_tracker is broken when editing an issue on master branch. 

See https://github.com/aces/Loris/pull/5394#issuecomment-575281470

**To Reproduce**
1. Go to issue_tracker
2. Click on an issue to edit.
3. See error

driusan:
Remove the remaining references to the LorisMenu table in the
issue tracker (in its ajax scripts) and update the code to
refer to the modules table instead.

Fixes #5938.

### DIFF
--- a/php/libraries/Module.class.inc
+++ b/php/libraries/Module.class.inc
@@ -59,6 +59,30 @@ abstract class Module extends \LORIS\Router\PrefixRouter
     }
 
     /**
+     * Retrieve all active module descriptors from the given database, indexed
+     * by their primary key in the database.
+     *
+     * @param \Database $db an open connection to a database containing a 'modules'
+     *                      table
+     *
+     * @return \Module[]
+     */
+    static public function getActiveModulesIndexed(\Database $db): array
+    {
+        $mnames = $db->pselect(
+            "SELECT ID, Name FROM modules WHERE Active='Y'",
+            []
+        );
+
+        $rv = [];
+        foreach ($mnames as $row) {
+            $rv[$row['ID']] = self::factory($row['Name']);
+        }
+        return $rv;
+    }
+
+
+    /**
      * Returns a Module instance for the module named $name. The returned
      * value may be a subtype of the base LORIS module class and override
      * it's method.
--- a/modules/issue_tracker/ajax/EditIssue.php
+++ b/modules/issue_tracker/ajax/EditIssue.php
@@ -435,13 +435,11 @@ function getComments($issueID)
     );
 
     //looping by reference so can edit in place
+    $modules = \Module::getActiveModulesIndexed($db);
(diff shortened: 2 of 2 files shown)


driusan:
@ridz1208 can you review since you were having troubles with this too?

driusan:
@ridz1208 see above

php/libraries/Module.class.inc
```diff
@@ -58,6 +58,30 @@ abstract class Module extends \LORIS\Router\PrefixRouter
...
+    {
+        $mnames = $db->pselect(
+            "SELECT ID, Name FROM modules WHERE Active='Y'",
+            []
+        );
+
+        $rv = [];
+        foreach ($mnames as $row) {
+            $rv[$row['ID']] = self::factory($row['Name']);
+        }
```
ridz1208: you can use `pselectWithIndexKey` function instead

php/libraries/Module.class.inc
```diff
@@ -58,6 +58,30 @@ abstract class Module extends \LORIS\Router\PrefixRouter
...
+    /**
+     * Retrieve all active module descriptors from the given database, indexed
+     * by their primary key in the database.
+     *
+     * @param \Database $db an open connection to a database containing a 'modules'
+     *                      table
+     *
+     * @return \Module[]
+     */
+    static public function getActiveModulesIndexed(\Database $db): array
```
ridz1208: I think this function belongs in a LORIS instance object more then this class...just saying

driusan: There is no LORIS instance object.

php/libraries/Module.class.inc
driusan: I didn't know that existed, but how would that work? This loop would still be required to convert the value to a `\Module` instance

php/libraries/Module.class.inc
ridz1208: create one

php/libraries/Module.class.inc
ridz1208: true, maybe we should add the possibility to give it a mapping function as well

php/libraries/Module.class.inc
driusan: That is not within the scope of this PR

php/libraries/Module.class.inc
driusan: That is not within the scope of this PR

driusan: #6021

php/libraries/Module.class.inc
driusan: #6022

PR #5964, opened 2020-01-27
[testing] filter test  - part-1

kongtiaowang:
## Brief summary of changes
Create a react filter test function

Travis

todo : part-2


kongtiaowang:
@driusan Could u review it?

htdocs/router.php
```diff
@@ -14,10 +14,10 @@
 $urlpath = ltrim($_SERVER['PHP_SELF'], "/");
 
 $request = $_SERVER['REQUEST_URI'];
-
 if ($request != '/'
     && (    file_exists(__DIR__ . $request)
-    || file_exists(__DIR__ . "/" . $urlpath))
+    || (file_exists(__DIR__ . "/" . $urlpath)
+    &&  strpos($urlpath, "acknowledgements") === false ))
```
driusan: Why is this change being done?

kongtiaowang: @driusan acknowledgments (Loris/**htdocs/acknowledgements**/acknowledgements.php ) is different from other modules, otherwise, this module will not work in PHP build-in server. It exists file under the acknowledgments folder.

driusan: then what is the (unchanged) line below that's already there for?

kongtiaowang: @driusan how about this change? The reason is that it shows "request /acknowledgements/ not found " before.

PR #5965, opened 2020-01-27
Create script to validate server configuration

### LINKED ISSUE #5021: [core] install script do not check if NodeJS is installed and a sufficient release
install script should check that NodeJS is properly installed and is of a sufficiently recent release to support the code.

If NodeJS is not installed, admin can still log-in but the only usable page are admin-config. All other page show a page with only the top banner.

johnsaigle:
## Brief summary of changes

@christinerogers recently presented some features of REDCap. Many of them seem beneficial to LORIS especially regarding set-up and installation.

One feature she recommended was creating a "Configuration Check". In REDCap, this does things like verifying minimum versions of dependencies, validating that directories exists and are writable, testing email, and so on.

This is an effort to implement this in LORIS using a PHP script in the `tools/` directory.

In the future I think this could be expanded into a module for front-end Administrators to check that their LORIS version is working properly.

Output:
Screen Shot 2020-02-07 at 14 29 59

##### Minimum versions of our requirements
- [x] PHP
- [x] Apache 
- [x] Composer 
- [x] NPM
- [x] MySQL/MariaDB

##### Web Paths
- [x] Do the directories exist and are they writable?

##### Connections to web-based dependencies
- ~~Google reCAPTCHA~~ This isn't working in LORIS at the moment but should be added in the future.
- [x] "Have I Been Pwned?", etc.

##### Outgoing email
- [x] Sends a test email

For now I am creating a script in `tools` to validate some settings in LORIS. Eventually I think this would be a nice module to include under `Admin`. But a tools script is a good start.

I'm also adding nice coloured output like I have in #5260 as well as logging via a "Helper class". This isn't necessary but I think it helps a lot and can be used by other tools.

Future work includes adding checks for PHP extensions.

### Testing instructions (if applicable)

Checkout the branch.
`cd` to the `tools/` directory and run the script. Make sure the output makes sense.

### Related issues

Resolves #5021 
Also resolves https://github.com/aces/Loris/issues/6811


tools/cli_helper.class.inc
```diff
@@ -0,0 +1,181 @@
+<?php declare(strict_types=1);
```
johnsaigle: This file has its own PR #5968. It's not necessary here but I find i helpful.

tools/configuration_check.php
```diff
@@ -0,0 +1,157 @@
...
+$result = $DB->pselect($query, []);
+
+foreach ($result as $setting) {
+    $name = $setting['Name'];
+    $path = $setting['Path'];
+
+    // If the path is an empty string, print a warning.
+    if ($path === "") {
+        $helper->printWarning(
+            "Setting `$name` is has no value associated with it. This may "
```
PapillonMcGill:
```suggestion
            "Setting `$name` has no value associated with it. This may "
```

tools/cli_helper.class.inc
PapillonMcGill: added specific comment in #5968

PapillonMcGill:
after proper rebase

driusan:
Blocked by #5968

PapillonMcGill:
why add a new file in smarty/? I taught that we are deprecating this tools

johnsaigle:
It's where our current email templates are stored so I added it there for consistency. We have been deprecating Smarty for front-end HTML but I don't know of an alternate standard for emails right now.

PapillonMcGill:
@johnsaigle Are you working on rebasing this PR? It is still in conflict at time of writing this.

johnsaigle:
I'll do that now. The rebase issue is just to do with the ordering of static testing so the PR can be reviewed now.

johnsaigle:
Also blocked by #6101

johnsaigle:
When #6101 is merged, the script should use the new admin email value for the email test feature. It should also print a warning if the value is empty as that means that users will not have a point of contact.

johnsaigle:
This script could also check whether a reCAPTCHA is properly configured and advise admins to enable it.

https://github.com/aces/Loris/pull/6344

johnsaigle:
I'll rebase this after the release

tools/configuration_check.php
```diff
@@ -0,0 +1,215 @@
...
+    if (!is_writeable($path)) {
+        $helper->printError(
+            "Setting `$name` has the value `$path` which is not writable "
+            . "by user `$username`."
+        );
+    }
+}
+
+if (! $config->settingEnabled('usePwnedPasswordsAPI')) {
+    printWarning(
```
kongtiaowang, line 145: $helper->printWarning

ridz1208:
@kongtiaowang can you rereview and test please

kongtiaowang:
php tools/configuration_check.php
PHP Fatal error:  Uncaught Error: Class 'OutputWrapper' not found in /var/www/loris/tools/configuration_check.php:23

driusan:
@kongtiaowang did you run `make`? (It's also possible that this needs to be rebased..)

kongtiaowang:
I did run "make" , but still can't work.

tools/configuration_check.php
```diff
@@ -0,0 +1,214 @@
+#!/usr/bin/env php
+<?php declare(strict_types=1);
+/**
```
kongtiaowang, line 3: missing 
require_once __DIR__ . "/generic_includes.php";

tools/configuration_check.php
```diff
@@ -0,0 +1,214 @@
...
+    shell_exec(
+        "composer --version | cut -d ' ' -f 3"
+    )
+);
+evaluateVersionRequirement('Composer', $versionString, MINIMUM_COMPOSER_VERSION);
+
+// Check "web paths". This is a data type found in the ConfigSettings table.
+// Settings with this Data Type are expected to exist on the file system and
+// should be writable by the current user.
+$username = $helper->getUsername();
```
kongtiaowang, line 98: Call to undefined method OutputWrapper::getUsername()

johnsaigle:
@driusan Any ideas on solving the errors Shen reported?

> PHP Warning:  require_once(generic_includes.php): failed to open stream: No such file or directory in /var/www/loris/php/libraries/OutputWrapper.class.inc on line 3
PHP Fatal error:  require_once(): Failed opening required 'generic_includes.php' (include_path='/data/loris/data/:.:/usr/share/php:/var/www/loris/tools/../project/libraries:/var/www/loris/tools/../php/libraries:') in /var/www/loris/php/libraries/OutputWrapper.class.inc on line 3

I'm not sure why these errors are appearing. There are no unusual permissions. I also tried adding `\LORIS\` in front of the OutputWrapper instantiation but it still didn't work.

Not sure how to solve it.

johnsaigle:
This PR also updates the OutputWrapper class in order to fix #6811

johnsaigle:
Update on above: the script runs fine as long as you're in the `tools/` directory

driusan:
@johnsaigle OutputWrapper should not be including anything at all, the file is a class, not a script. The requires are in the script itself..

johnsaigle:
Removing the require in the OutputWrapper solved the issue. 👍

driusan:
@kongtiaowang can you please re-review/test this now that your issue has been solved?

php/libraries/OutputWrapper.class.inc
```diff
@@ -39,15 +38,21 @@ class OutputWrapper
...
      * @return void
      */
     public function __construct(?SPLFileInfo $logDirectory = null)
     {
+        $config = \NDB_Factory::singleton()->config();
         $logDir = new SPLFileInfo(
-            $logDirectory ?? \NDB_Factory::singleton()->config()->getSetting('log')
+            $logDirectory ??
+            \Utility::pathJoin(
+                $config->getSetting('base'),
```
driusan, line 53: I don't think this is going to be correct if the log setting is an absolute path.

johnsaigle: Fortunately this is the only place in the codebase where the setting is being used

Screen Shot 2020-07-07 at 16 45 42

driusan: okay, but it's still going to be wrong if the log directory is set to an absolute path (ie. `/var/log/loris`)..

johnsaigle: So what's the solution?

The same can be said of the detect_conflicts script. The status quo right now is that, to the extent `log` is used, it's a partial path stored relative to `base` so this is fine.

It's possible to change it to instead change it to an absolute path: we can change the SQL to be of a `web_path` type which would cause the Config module to validate it when submitting. That would require updating `detect_conflicts` as well. It's probably not a bad idea for a separate pull request. 

But right now this is just using a config setting as it's expected to be used so I'm not sure what you want.

driusan: Why can't you just check if it starts with a `/` and not join if so? You said this is the only place it's being used..

johnsaigle: Other than the tools/detect_conflicts.php script, yes

johnsaigle: done

kongtiaowang:
@johnsaigle in this commit e27c8e0, It passed my test.
But the lastest commit e04e7b9, I got error [PHP Fatal error:  Uncaught ConfigurationException: Configuration setting for Log directory is not a directory. in /var/www/loris/php/libraries/OutputWrapper.class.inc:62
Stack trace:
#0 /var/www/loris/tools/configuration_check.php(23): OutputWrapper->__construct()]

johnsaigle:
@kongtiaowang What value do you have for your `log` setting?

kongtiaowang:
@johnsaigle tools/logs

php/libraries/OutputWrapper.class.inc
```diff
@@ -39,16 +38,26 @@ class OutputWrapper
...
     {
-        $logDir = new SPLFileInfo(
-            $logDirectory ?? \NDB_Factory::singleton()->config()->getSetting('log')
-        );
+        $config = \NDB_Factory::singleton()->config();
+        $logDir = $logDirectory ?? $config->getSetting('log');
+
+        // If the log directory is a partial path, try appending it to the
+        // LORIS base directory.
+        if (!strpos($logDir, '/') === 0) {
```
kongtiaowang, line 54: strpos($logDir, '/') !== 0
example: tools/logs

johnsaigle: Yes but the code within this if statement creates `base` + `log`, so in your exampe `/var/www/loris/tools/logs` which should work.

johnsaigle:
@kongtiaowang Does this path exist on your VM?

johnsaigle:
I just added a new commit that will print the path in the debug message. Can you try again?

php/libraries/OutputWrapper.class.inc
kongtiaowang: ex, !strpos($logDir, '/') will return !5, then !5 === 0 , so it will never go into the if condition.
 strpos($logDir, '/') !== 0 it will work.

PR #5966, opened 2020-01-28
[Statistics / Behavioural Statistics / Double Data Entry Statistics] Fix permission issue where users were allowed to see the double data entry breakdown per participat from sites they don't have access to.

racostas:
### Description of the issue

Users were allowed to see the double data entry breakdown per participant from sites they don't have access to.

### Brief summary of changes

The `_hasAccess` function was incorpoerate to this class (same functionality that the one in #5591) 

The function `_checkCriteria` was modified for only retrieve data from the sites(centers) the user have access to.

Note: In future developments the capacity of setting permissions per site per user could be desirable. A note in rapport was included in the codeset. In this case the proposed function `_hasAccess` should be updated accordingly .

A user with permission 'data_entry' should be now able to access the 'breakdown per participant' for the double data entry page only for the sites it have access to.

#### Links to related PRs

#5950 
#5591

### DIFF
--- a/modules/statistics/php/statistics_dd_site.class.inc
+++ b/modules/statistics/php/statistics_dd_site.class.inc
@@ -30,6 +30,64 @@ class Statistics_DD_Site extends statistics_site
     var $query_vars     = array();
 
     /**
+     * Checking user's permission
+     *
+     * @param \User $user The user whose access is being checked
+     *
+     * @return bool
+     */
+    function _hasAccess(\User $user) : bool
+    {
+        //TODO: Create a permission specific to statistics
+        $hasAccessToAllProfiles = $user->hasAllPermissions(
+            array(
+                'access_all_profiles',
+                'data_entry',
+            )
+        );
+
+        $hasCenterPermission = false;
+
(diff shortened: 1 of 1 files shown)


ridz1208:
@racostas I changed the base of this one to 23 as it is not exactly critical and it contains alot of code changes.

we will assign it to someone for testing accordingly

ridz1208:
@laemtl could you test and review this please ?

laemtl:
Since @55c15d0, 0000-00-02-Menus.sql has been replaced by 0000-00-02-Modules.sql. This branch throw a 500 error because the table LorisMenu doesn't exist. I fixed this by manually sourcing  0000-00-02-Menus.sql, but wonder if LorisMenu/LorisMenuPermissions should still be present by defualt in Loris DB?

christinerogers:
@laemtl what you're seeing is that it needs a serious rebase on the 23 branch.  old code is interfering with just pulling the change itself.  rebasing will clear that clutter.

ridz1208:
@laemtl

PR #5968, opened 2020-01-28
[Tools] Add command line helper class

johnsaigle:
## Brief summary of changes

In the PR for the raisinbread refresh, @ridz1208 and I thought it would be a good idea to create a "Command Line Helper" class that PHP tool scripts can optionally use to help with common tasks.

I'll keep this file simple to help with code review. We can iterate on the functions of the class as we go forward.

#### Testing Instructions

Right now the class provides helper functions to print colourful output as well as an option to log the output of those functions to a file in `tools/logs/` using the same naming format as the install script.

Copy the following code into a new PHP file called `test.php` in `tools/`.

```php
<?php

require_once 'generic_includes.php';

$helper = new OutputWrapper();
$helper->enableLogging($argv[0]);
$helper->printSuccess('The class is working!');
```

1. Verify that the message prints in green text
2. Check `tools/logs/` and make sure a new log file was created with the output.


tools/cli_helper.class.inc
```diff
@@ -0,0 +1,181 @@
...
+     * @return void
+     */
+    public function enableLogging(string $scriptName): void
+    {
+        // Create a new file in the format:
+        // `$scriptName-YEAR-MONTH-DAY_HOUR-MINUTE-SECOND.log`
+        $this->_loggingFileName = sprintf(
+            "%s-%s.log",
+            // Remove extension information
+            explode('.', $scriptName)[0],
```
PapillonMcGill: This pose a problem in the case of a script named like `this.wonderfull.script.php`
Could use pathinfo()

tools/cli_helper.class.inc
```diff
@@ -0,0 +1,181 @@
...
+     *
+     * @return void
+     */
+    function printBulletPoint(string $line): void
+    {
+        echo "\t* $line\n";
+    }
+
+    /**
+     * Prints a line of text with a newline before and after. Can also be
```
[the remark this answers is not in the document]
johnsaigle: I changed it to print only a line feed at the beginning and updated the comment.

The other print function are wrappers around this one and they handle colours and prefixes. They do log automatically since they pass through this function. So I think your idea is already present in the code.

johnsaigle:
@driusan Can you review and merge if satisfied? I'd like to use this class in other PRs that are modifying tools scripts.

driusan:
4. I also don't like the class name. It doesn't describe what it does in a meaningful way.

johnsaigle:
> I also don't like the class name. It doesn't describe what it does in a meaningful way.

I'm hoping to eventually extend it to do things like 
* verifying a script is passed the correct number of arguments
* containing an abstract function or a property that displays the usage string
* ...

Do you have an alternate suggestion for a name that is general enough so that we can extend the class?

johnsaigle:
Apart from the class name, I've made the other changes.

driusan:
`LogColourizer` or `OutputWrapper` or something.

I can't think of any time it would be a good idea to have a class that's so unfocused it deals with both colourizing output, logging data, and counting arguments all at the same time.

ridz1208:
@driusan 

> I can't think of any time it would be a good idea to have a class that's so unfocused it deals with both colourizing output, logging data, and counting arguments all at the same time.

you mean any of our current tools ? which are all doing it ad-hoc as well ?

johnsaigle:
> ..and files should either define classes or have side-effects (ie require), not both.

I don't think it does anything? It's only a collection of functions that can do things. It's separate from any script code

php/libraries/OutputWrapper.class.inc
```diff
@@ -0,0 +1,186 @@
+<?php declare(strict_types=1);
+require_once 'generic_includes.php';
```
johnsaigle:
```suggestion
```

driusan:
God classes are never a good idea. There's no reason you can't have separate classes for each of those concerns.

johnsaigle:
Isn't multiple inheritance not possible though? Each tool would need all of those concerns

php/libraries/OutputWrapper.class.inc
```diff
@@ -0,0 +1,185 @@
...
+     */
+    public function printLine(
+        string $line,
+        string $colour = '',
+        $handle = STDOUT
+    ) {
+        $output = $line;
+        // If a valid colour is supplied and the output is going to a TTY,
+        // colourize the output.
+        if (!empty($colour) && stream_isatty($handle)) {
```
driusan: You want to validate `$colour` regardless of if the `$handle` is a tty, you just don't want to modify the output if it's a TTY.

php/libraries/OutputWrapper.class.inc
```diff
@@ -0,0 +1,185 @@
...
+ * for LORIS.
+ *
+ * It contains helper functions to format output, etc.
+ *
+ * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ */
+class OutputWrapper
+{
+    // Colours associated with the bash shell.
+    // @see https://misc.flogisoft.com/bash/tip_colors_and_formatting
```
driusan: ANSI escape sequences and colours predate bash by a long shot, and the wikipedia page (https://en.wikipedia.org/wiki/ANSI_escape_code) is probably a better reference than someone's blog.

php/libraries/OutputWrapper.class.inc
johnsaigle: `$colour` should never be considered if the handle isn't a TTY though. We don't need it.

php/libraries/OutputWrapper.class.inc
johnsaigle:
```suggestion
    // Colours mapped to ANSI escape codes.
    // @see https://en.wikipedia.org/wiki/ANSI_escape_code#3/4_bit
```

php/libraries/OutputWrapper.class.inc
driusan: But the caller still passed an invalid value to the function and violated the function's precondition.

ridz1208: though call here but I think I'm slightly with @driusan 

It's just as confusing if you have a script that works fine for years when stream is not a TTY and then suddently you run it from terminal and it starts failing... I feel like it's a safer bet to have the colour validated immediately, most likely the developer would catch the error as soon as the code is changed for and invalid colour instead of only the first time it is run from a TTY.


The other option in that case would be to create a printColourfulLine and have the client decide which function to use... but i feel like we lose a lot of the streamlining with that so I'm not a big fan

ridz1208: @johnsaigle if you agree with the above I created a PR on ur repo to apply the changes and fix PHAN

https://github.com/johnsaigle/Loris/pull/18

johnsaigle: I merged your PR on my fork so this should be resolved

php/libraries/OutputWrapper.class.inc
```diff
@@ -0,0 +1,188 @@
...
+     *
+     * @param ?SPLFileInfo $logDirectory The directory that will contain log
+     *                                   files written by this script.
+     *
+     * @return void
+     */
+    public function __construct(SPLFileInfo $logDirectory = null)
+    {
+        $this->_logDirectory = new SPLFileInfo(
+            $logDirectory ?? __DIR__ . '/tools/'
```
driusan, line 44: This path is wrong. It'll resolve to "$LORIS/php/libraries/tools"

johnsaigle: Good catch, thanks.

php/libraries/OutputWrapper.class.inc
```diff
@@ -0,0 +1,188 @@
...
+                "Invalid argument for colour (`$colour`)"
+            );
+        }
+
+        // If a valid colour is supplied and the output is going to a TTY,
+        // colourize the output.
+        if (!empty($colour) && stream_isatty($handle)) {
+            // Format the output to colour the whole line.
+            $output = sprintf("\n\e[%sm$line\e[0m", self::COLOURS[$colour]);
+        }
```
driusan, line 94: this needs an else, otherwise it'll silently eat the line if not attached to a TTY.

php/libraries/OutputWrapper.class.inc
```diff
@@ -0,0 +1,188 @@
...
+            throw new \InvalidArgumentException(
+                "Invalid argument for colour (`$colour`)"
+            );
+        }
+
+        // If a valid colour is supplied and the output is going to a TTY,
+        // colourize the output.
+        if (!empty($colour) && stream_isatty($handle)) {
+            // Format the output to colour the whole line.
+            $output = sprintf("\n\e[%sm$line\e[0m", self::COLOURS[$colour]);
```
driusan, line 93: Why is colour interpolated but $line is directly inserted in the formatting string? $line should be using a formatting verb too so that it doesn't get interpolated by sprintf.

php/libraries/OutputWrapper.class.inc
johnsaigle: `$output` gets initialized above. The if statement modifies `$output` if `$handle` is a TTY. In every case `$output` is written to `$handle` on line 96 so nothing will get silently eaten.

johnsaigle:
@driusan this is ready for review

php/libraries/OutputWrapper.class.inc
driusan: my mistake.. I was misreading that as printf not sprintf.

php/libraries/OutputWrapper.class.inc
driusan: This isn't resolved

php/libraries/OutputWrapper.class.inc
```diff
@@ -0,0 +1,190 @@
...
+     *
+     * @param ?SPLFileInfo $logDirectory The directory that will contain log
+     *                                   files written by this script.
+     *
+     * @return void
+     */
+    public function __construct(SPLFileInfo $logDirectory = null)
+    {
+        $this->_logDirectory = new SPLFileInfo(
+            $logDirectory ?? \NDB_Factory::singleton()->config()->getSetting('log')
```
driusan, line 45: This should probably check that the resulting directory is writeable and is a directory.

johnsaigle: We should probably also modify the Log config setting in the SQL table to be a `web_path` since that will allow the Configuration module to validate this when changed on the front-end.

php/libraries/OutputWrapper.class.inc
johnsaigle: I made them before but it looks like they got reverted during rebasing

Issue #5940, opened 2020-01-17
[Module] lorisform_parser is no longer working in v22 master branch after recent commits


lingma:
**Describe the bug**
With new update to Module or some related modules, the lorisform_parser is no longer working.

**To Reproduce**
Go to tools/, `ls` any instrument to feed `lorisform_parser`, the system will report

PHP Fatal error:  Uncaught Error: Cannot instantiate abstract class Module in tools/lorisform_parser.php:57

**What did you expect to happen?**

There should be no problem like this, at least before Module or related program was modified recently.

**Server Environment (if known):**
- LORIS Version: [e.g. 22.0.0] 22.0.0
- MySQL/MariaDB Version: [e.g. MySQL 5.7]
MySQL 5.7

driusan:
the line mentioned needs to be updated to use `new NullModule()` instead of `new Module("", "")`

johnsaigle:
I think this is resolved by the linked PR

PR #5970, opened 2020-01-28
Module to NullModule change due to libraries modification

lingma:
## Brief summary of changes

With new update to Module or some related modules, the lorisform_parser is no longer working. See issue #5940, this PR is created according to the information provided by @driusan.

_To Reproduce_
Go to tools/, ls any instrument to feed lorisform_parser, the system will report

PHP Fatal error: Uncaught Error: Cannot instantiate abstract class Module in tools/lorisform_parser.php:57

1. Go to tools/, ls any instrument to feed lorisform_parser

Issue #5940

### DIFF
--- a/tools/lorisform_parser.php
+++ b/tools/lorisform_parser.php
@@ -54,7 +54,7 @@ foreach ($files AS $file) {
     echo "Requiring file...\n";
     include_once $file;
     echo "Instantiating new object...\n";
-    $obj =new $className(new Module("", ""), "", "", "", "");
+    $obj =new $className(new NullModule("", ""), "", "", "", "");
     echo "Initializing instrument object...\n";
     $obj->setup(null, null);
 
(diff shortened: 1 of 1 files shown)


tools/lorisform_parser.php
```diff
@@ -54,7 +54,7 @@
     echo "Requiring file...\n";
     include_once $file;
     echo "Instantiating new object...\n";
-    $obj =new $className(new Module("", ""), "", "", "", "");
+    $obj =new $className(new NullModule("", ""), "", "", "", "");
```
driusan:
```suggestion
    $obj =new $className(new NullModule(), "", "", "", "");
```

This shouldn't be passing Travis, it has the wrong number of arguments for the constructor for NullModule..

lingma: So there is no testcase for this or no rule? Strange.

lingma: > So there is no testcase for this or no rule? Strange.

@johnsaigle maybe you can check the rules? The tools directory is not verified in the rules. I checked, there will be a lot errors using the current rules.

johnsaigle: Yeah Phan doesn't check the tools directory. It would be good to enable this. We should probably get #5766 merged first

PR #5971, opened 2020-01-29
Add tools/assign_missing_instruments.php to PHPCS

johnsaigle:
This is one script from #5766

### DIFF
--- a/tools/assign_missing_instruments.php
+++ b/tools/assign_missing_instruments.php
@@ -140,7 +140,7 @@ if (isset($visit_label)) {
     foreach ($results as $result) {
         populateVisitLabel($result, $visit_label);
     }
-} elseif (isset($visit_labels)) {
+} else if (isset($visit_labels)) {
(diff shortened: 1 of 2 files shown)


tools/assign_missing_instruments.php
```diff
@@ -138,15 +137,15 @@ function populateVisitLabel($result, $visit_label)
     $where = array('vl' => $argv[1]);
 
     $results = $DB->pselect($query, $where);
-    foreach ($results AS $result) {
+    foreach ($results as $result) {
         populateVisitLabel($result, $visit_label);
     }
-} else if (isset($visit_labels)) {
+} elseif (isset($visit_labels)) {
```
driusan: stylistically I think this change is bad, what rule is requiring it?

johnsaigle: It's not required actually. Not sure why this change was made. I'll revert.

Issue #5962, opened 2020-01-27
Password rules are incorrect


thomasbeaudry:
On the Edit User's page, the password rules are outlined as:

    The password must be at least 8 characters long.
    The password cannot be your username or email address.

This is not in fact true, since the password that you choose has to also contain a special character or else you get the error message:   The password is not complex enough.  For example,  If you try the password:  **iamsosmrt**   you will see that it isn't accepted, but **iamsosmrt!**  is.

thomasbeaudry:
oops sorry @cmadjar, i didn't realize i was on LORIS-MRI repo

gdevenyi:
I reported issues with the password rules here:
https://github.com/aces/Loris/issues/3314

Which resulted in a PR here:
https://github.com/aces/Loris/pull/3962

Eventually changes were merged:
https://github.com/aces/Loris/pull/4411

But this issue indicates the password rules do not yet follow NIST standards?

gdevenyi:
Possibly waiting on this https://github.com/aces/Loris/issues/5761

johnsaigle:
The password rules aren't incorrect. The error message generated for that input is `The password is not complex enough.`, which is true. We use [Zxcvbn](https://github.com/dropbox/zxcvbn) to evaluate password strength according to NIST standards (see [this post](https://github.com/aces/Loris/issues/3314#issuecomment-534233065)). 

A special character isn't required; a sufficiently complex password is. The password `sufficientlengthandcomplexity` is a valid password in LORIS because its length alone gives it sufficient entropy so that it cannot be easily cracked by bruteforce password cracking software.

We have unit tests for passwords to confirm that weak passwords are rejected and that strong passwords are acceptable.

If you want to evaluate the strength of a given password I invite you to check out the tool https://lowe.github.io/tryzxcvbn/ which is an interactive demo of the library we use to evaluate password strength.

Also, there's a final line of text under the Password Rules you've quoted: 

> We suggest using a password manager to generate one for you.

thomasbeaudry:
I still think the rules could be written clearer.  For instance   iamsosmrt1  is rejected, but iamsosmrt!  passes the test (which is why i assumed it needed a special character).  Both passwords are the same length.  I have no desire in using a password manager.

johnsaigle:
@thomasbeaudry I just issued #5974. Do you mind reviewing to let me know if my changes are more clear?

driusan:
[...]

Issue #5761, opened 2019-11-25
Add front-end password strength indication for users changing passwords


johnsaigle:
We're using Zxcvbn to evaluate password strength on the back-end. https://github.com/bjeavons/zxcvbn-php

There is also a JS version that can be used to show a password strength measurement for users. This will help people to create better passwords if they aren't using a password manager.

maltheism:
I should add this to my react Login PR. Commenting here and so I don't forget.

PR #5828, opened 2019-12-04
[login] module upgraded for react

maltheism:
## Brief summary of changes

Upgraded module to use react. I had to fix some things in LORIS to get this all to work and there are multiple changes in some JSX files as well.

What was updated to react:
Login, password expired, request account, and reset password.

Summary:
Turns out the Content-Security policy was broken for google captcha and I made a fix to include the iframe that recaptcha uses. Major was also broken inside the php/libraries/Email.class.inc and I made some fixes. That had to do with the password reset.

Note:
This PR includes the bugfix PR #4386 and so this PR can function.

#### Testing instructions (if applicable)

1. Checkout PR.
2. Login
3. Logout
4. Request Account
5. Recover Password

#### Links to related tickets (GitHub, Redmine, ...)

* Redmine #16050
* Resolves #5828 
* Resolves #5214

PR #5974, opened 2020-01-29
Clarify documentation on password requirements

johnsaigle:
## Brief summary of changes

Adds a line to front-end explanation of password requirements to clarify what is expected from the users.

This is a small improvement but we need to add #5761 in order for this to be truly user-friendly.

* There was some confusion about the password requirements resulting in issue #5962. This is meant to address that confusion.

### DIFF
--- a/modules/login/templates/form_passwordexpiry.tpl
+++ b/modules/login/templates/form_passwordexpiry.tpl
@@ -7,8 +7,8 @@
       <p><b>Password Strength Rules</b></p>
       <ul>
         <li>The password must be at least 8 characters long.</li>
-        <li>The password and the user name must not be the same.</li>
         <li>The password and the email address must not be the same.</li>
+        <li>No special characters are required but your password must be sufficiently complex to be accepted.</li>
       </ul>
         <p><b>Please choose a unique password.</b></p>
         <p>We suggest using a password manager to generate one for you.</p>
--- a/modules/user_accounts/templates/form_edit_user.tpl
+++ b/modules/user_accounts/templates/form_edit_user.tpl
@@ -41,6 +41,7 @@
       <ul>
         <li>The password must be at least 8 characters long.</li>
         <li>The password cannot be your username or email address.</li>
+        <li>No special characters are required but your password must be sufficiently complex to be accepted.</li>
       </ul>
         <p>Please choose a unique password.</p>
         <p>We suggest using a password manager to generate one for you.</p>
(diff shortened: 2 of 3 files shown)


thomasbeaudry:
Well it's better but it could be clearer.  What does "sufficiently complex" mean?  Can you provide a link to the standards if a user wants more info?  Or if it's short enough can you summarize what is required?

johnsaigle:
There are no hard rules for standards which makes this tough to word.

We're using Zxcvbn to handle a few NIST requirements spelled out here https://github.com/aces/Loris/issues/3314

> Must compare passwords against a blacklist of known passwords when changing
(examples of blacklists given in is passwords from previous breach corpuses,
dictionary words, repetitive or sequential characters, context-specific words
such as the name of the service, username or derivatives thereof.)
>  If password appears in a blacklist, force a different password
> No complexity requirements other than the length

My interpretation of this is basically that we can require that passwords be of at least X characters in length (8, currently) but beyond that we cannot require any specific character set. 

Zxcvbn helps us handle the first point above by checking against those lists and substitutions.

I don't know how to word in a general way that one's password has to be just.... not "bad". The password `9YgR4%(#` is valid and so is `correct horse battery staple` but the only thing they have in common is that they have entropy/complexity.

modules/login/templates/form_passwordexpiry.tpl
```diff
@@ -7,8 +7,8 @@
       <p><b>Password Strength Rules</b></p>
       <ul>
         <li>The password must be at least 8 characters long.</li>
-        <li>The password and the user name must not be the same.</li>
         <li>The password and the email address must not be the same.</li>
+        <li>No special characters are required but your password must be sufficiently complex to be accepted.</li>
```
PapillonMcGill: Can you clarify what are the condition for "sufficiently complex" password?

Someone might think that `9876SA321` is enough

gdevenyi: Its not really possible to provide strict requirements as those are mostly meaningless compared to the entropy calculations provided by Zxcvbn.

When Zxcvbn is fully integrated on the frontend it will provide live feedback on on the password and show when it "passes" the minimum requirements.

johnsaigle: @PapillonMcGill see my above comment. There isn't really a way to explain it better. I agree we really need the front-end Zxcvbn module #5761 which is held up by #5828

PapillonMcGill: @johnsaigle maybe add 
"... sufficiently complex and hard to guest to be accepted."

johnsaigle: What is "hard to guess" though?

PapillonMcGill: I know, not easy. Ok, let go with it, better than current.

PR #5975, opened 2020-01-29
Move script cleanup_mri_tables_for_19-0_release to single_use/ directory

johnsaigle:


ridz1208:
@johnsaigle due to the change in location the `require_once` need changes in the path (you can make them absolute using __DIR__)

there are also paths in the script itself that will need changing

PR #5977, opened 2020-01-29
[Tools] Enable PHPCS for manage_modules.php

johnsaigle:


tools/manage_modules.php
```diff
@@ -49,52 +49,64 @@
...
-if(isset($flags['add'])) {
+if (isset($flags['add'])) {
     addDir(__DIR__ . "/../modules");
     addDir(__DIR__ . "/../project/modules");
 }
 
-function addDir(string $moduledir) {
+/**
+ * Adds a new entry to the `modules` table based on a path supplied in the
+ * $moduledir parameter.
```
driusan: This comment isn't right.

It adds any subdirectories under $moduledir that are valid modules to the modules table.

laemtl:
Replaced by #6787

PR #5984, opened 2020-01-29
[Tools] Enable PHPCS for fix_timepoint_date_problems.php

johnsaigle:


laemtl:
@driusan This is ready for review

ridz1208:
@laemtl just to be sure, when you are approving these are you also testing them ?

laemtl:
@rida Yes !

laemtl:
I tested the script and noticed a Fatal error:
PHP Fatal error:  Uncaught TypeError: Return value of TimePoint::getScreeningStatus() must be of the type string, null returned in /var/www/html/loris/php/libraries/TimePoint.class.inc:493

Should the return type of `getScreeningStatus` changed for ?string?

ridz1208:
@laemtl yes please change it

tools/fix_timepoint_date_problems.php
```diff
@@ -1,531 +1,673 @@
...
-$client->initialize($configFile);
-
-$db = \Database::singleton();
-
-/**
- * HELP SCREEN
- * display and stop processing if action=help
- */
-if (empty($argv[1]) || $argv[1] == 'help') {
-    fwrite(STDERR, "Usage: \n\n");
```
driusan, line 63: This is changing the destination of the error message. It should still be going to stderr if the usage was incorrect.

tools/fix_timepoint_date_problems.php
```diff
@@ -1,531 +1,673 @@
...
+$client->initialize($configFile);
+
+$db = \Database::singleton();
+
+/**
+ * HELP SCREEN
+ * display and stop processing if action=help
+ */
+if (empty($argv[1]) || $argv[1] == 'help') {
+    //phpcs:disable
```
driusan, line 71: 👎

johnsaigle: We either have to disable PHPCS here or modify the help content so that it's ahrder to read and maintain. Since it's only disabled for this heredoc I think it's a fine thing to ignore. This is a section of the code where PHPCS is not helpful.

laemtl:
Replaced by #6798

PR #5985, opened 2020-01-29
[Tools] Enable PHPCS for lorisform_parser.php

johnsaigle:


tools/lorisform_parser.php
```diff
@@ -54,7 +54,7 @@
     echo "Requiring file...\n";
     include_once $file;
     echo "Instantiating new object...\n";
-    $obj =new $className(new NullModule(), "", "", "", "");
+    $obj =new $className(new Module("", ""), "", "", "", "");
```
driusan: This is wrong and a regression.

PR #5987, opened 2020-01-29
[Tools] Enable PHPCS for resetpassword.php

johnsaigle:


johnsaigle:
@driusan This one is next

PR #5989, opened 2020-01-29
[Tools] Enable PHPCS for generate_project_statistics_csv.php

### LINKED ISSUE #5120: [tools] Update project statistics script
the project stats script `generate_project_statistics_csv.php`, created to get numbers from other projects  : 

- [ ] Print the $projectname in the csv output file : i.e. Move lines 23 and 24 down to 121, and prefix $projectname to $output_file
- [ ] Update email address to loris.info@mcin.ca (not exposed on the public listserv loris-dev@bic.mni.mcgill.ca )
- [ ] review for other updates since 2016...

from RM 8626

johnsaigle:
## Brief Summary

This PR update a script which is used to gather basic project information about LORIS, such as the number of installed instruments.

In its current state, the script is buggy and does not run properly for me. I also find the code to be somewhat messy, inefficient, and unclear. 

This PR keeps the same functionality but improves the operation of the script.

It also adds PHPCS checking to the script.

## Testing
Run the script in its current state. If it works for you, keep a copy of its output file.
Run the updated script. Compare the results and make sure the numbers match.

Resolves #5120


johnsaigle:
This script is blocked by #5968

tools/generate_project_statistics_csv.php
```diff
@@ -1,27 +1,17 @@
 #!/usr/bin/env php
-<?php
+<?php declare(strict_types=1);
 /**
- * @author Stella Lee
- * Date: 15-08-15
+ * Generates a CSV report of various statistics about this study.
  *
- * Project Statistics - Updating Counts From Each LORIS
```
christinerogers, line 7: I'd suggest keeping a line that documents what the script is for.

christinerogers:
@laemtl @AlexandraLivadas  Once we're back to working on `master` could you both test this script and suggest improvements?  Its purpose is to email back to us some basic metrics about the data in a LORIS instance. 

Thanks to @johnsaigle for this update.

tools/generate_project_statistics_csv.php
johnsaigle: If you look to the right I added a new one.

laemtl:
The previous script version doesn't work for me.
The output of the new version is cleaner and produces a csv with values that seems accurate.

johnsaigle:
>  Is the email not automatically sent for privacy reasons?

@laemtl We could send the email automatically. I don't think there's a problem with it. For this PR though I aimed to fix/keep the same functionality rather than add a new feature.

php/libraries/OutputWrapper.class.inc
```diff
@@ -89,6 +88,16 @@ class OutputWrapper
...
         $output = $line;
+
+        // Verify that if a colour is supplied, it is valid
+        if (!empty($colour)
+            && !array_key_exists(strtolower($colour), self::COLOURS)
+        ) {
+            throw new \InvalidArgumentException(
+                "Invalid argument for colour (`$colour`)"
+            );
+        }
```
driusan, line 99: This is duplicating the line below it.

tools/output_wrapper.class.inc
```diff
@@ -0,0 +1,186 @@
+<?php declare(strict_types=1);
```
driusan, line 1: This file looks like it's from a bad rebase and isn't supposed to be included..

laemtl:
Replaced by #6795

PR #5990, opened 2020-01-29
[Tools] Enable PHPCS for generate_tables_sql.php

johnsaigle:


laemtl:
@driusan this is ready for review

Issue #5969, opened 2020-01-28
[22.0.0 - new profile] Empty drop down with only 1 site affiliation


ridz1208:
**Describe the bug**
When a user entering a new candidate is only affiliated with 1 site, the site dropdown on the user profile page is empty

this bug was found by @thomasbeaudry

lingma:
This issue has already been addressed in PR #5892

ridz1208:
closing because fix is in the wrong place, new PR will be issued soon. the error is in the JS not hiding the site field

PR #5991, opened 2020-01-29
[New_profile] Fix 1 site affiliation empty dropdown

ridz1208:
## Brief summary of changes
fixes https://github.com/aces/Loris/issues/5969

An if block was preventing from loading any sites in the dropdown when user logged in is only affiliated with 1 site.


lingma:
I think the PR #5892 is better made than this PR.

ridz1208:
@lingma PR #5892 is going to the wrong branch and includes several unnecessary changes for a very simple bugfix. If it includes clean up then it is more than welcome for the next version. This PR is just fixing a bug in our latest release.

PR #5992, opened 2020-01-29
[Tools] Enable PHPCS for delete_candidate.php

johnsaigle:


tools/delete_candidate.php
```diff
@@ -122,27 +122,42 @@
...
+Usage: php delete_candidate.php delete_candidate CandID PSCID [confirm]
+Example: php delete_candidate.php delete_candidate 965327 dcc0007
+Example: php delete_candidate.php delete_candidate 965327 dcc0007 confirm
+Example: php delete_candidate.php delete_candidate 965327 dcc0007 tosql
+
+When the 'tosql' function is used, the SQL file exported will be located
+under the following path: 
+    loris_root/project/tables_sql/DELETE_candidate_CandID.sql
+USAGE;
+    die;
```
driusan: What happened to the `()`?

johnsaigle: Don't remember. `die` is the same as `exit` in PHP so passing brackets or an exit code is optional.

PR #5994, opened 2020-01-29
[Tools] Enable PHPCS for detect_conflicts.php

johnsaigle:


tools/detect_conflicts.php
```diff
@@ -105,22 +107,27 @@
...
 }
 print_r($opts);
-  
-if (($change && $change_all) || ($delete_ignored_conflicts && ($change || $change_all))) {
-    $die("Choose either Change (-c) or remove and re-insert all conflicts (-m) or remove ignored conflicts (-r)");
+
+if (($change && $change_all)
+    || ($delete_ignored_conflicts && ($change || $change_all))
+) {
+    $die(
```
driusan: I don't think the dollar sign is right?

johnsaigle: Very weird

laemtl:
Replaced by #6794

PR #5835, opened 2019-12-05
[genomic_browser] upgraded module for react

maltheism:
## Brief summary of changes

Upgraded the genomic_browser module to use FilterDataTable components and the DataFramework on the backend. A lot of the backend code isn't modified and using the same code from before. I did move the uploading file code to a new class and restructure the class files for replicating the directory structure of the conflict resolver DataFramework PR.

Note: The genomic_browser module is slow from observing all the data that's retrieved from the database. I don't believe this PR should fix that problem.

#### Testing instructions (if applicable)

1. Checkout PR
2. Test on Rasinbread.

Resolves #5720

PR #6003, opened 2020-01-29
[Travis] increase memory

kongtiaowang:
## Brief summary of changes
increase memory for Docker PHP container

Travis

### DIFF
--- a/Dockerfile.test.php7
+++ b/Dockerfile.test.php7
@@ -5,3 +5,6 @@ RUN apt-get update && \
 
 RUN docker-php-ext-install pdo_mysql && \
     docker-php-ext-install zip
+
+RUN echo "memory_limit = 512M" >> /usr/local/etc/php/conf.d/docker-php-memlimit.ini
+


johnsaigle:
What does this do? Is it going to speed up Travis? Were there memory issues?

driusan:
Yeah, why does Travis need more memory? If something in LORIS is eating memory we should investigate what and fix the root cause..

kongtiaowang:
@johnsaigle @driusan It only happens in Testing Docker system #5835.  
Loris works fine on VM.
<b>Fatal error</b>:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 4096 bytes) in <b>/app/src/Data/Table.php</b> on line <b>84</b><br />

johnsaigle:
Is there a specific test that is using all of that memory and causing that failure?

kongtiaowang:
https://github.com/aces/Loris/pull/5835 has been fixed.  I will close this PR

PR #6005, opened 2020-01-29
[bug] sessionID is never passed to feedback_mri_popup.php

lingma:
## Brief summary of changes
I noticed that since Loris 19.1, the sessionID was never passed to feedback_mri_popup, which is required in the program. As a result, quite a lot comments data missed sessionID field. The fix would address this problem for the future but won't touch the existing data.

If necessary, this patch could be applied to the old version with necessary adjustment.

1. Go to Image browser, click on "QC Comments" of any image file, check whether sessionID is passed or not.

### DIFF
--- a/modules/imaging_browser/jsx/ImagePanel.js
+++ b/modules/imaging_browser/jsx/ImagePanel.js
@@ -510,7 +510,7 @@ class ImageQCCommentsButton extends Component {
     e.preventDefault();
     window.open(
       this.props.BaseURL + '/feedback_mri_popup.php?fileID=' +
-      this.props.FileID,
+      this.props.FileID + '&sessionID=' + this.props.sessionID,
       'feedback_mri',
       'width=500,height=800,toolbar=no,location=no,' +
       'status=yes,scrollbars=yes,resizable=yes'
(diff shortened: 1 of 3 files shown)


lingma:
@ridz1208 @driusan @johnsaigle  similar as PR #5991, will you need to add this fix to 22.1 or old versions? I saw that the program is wrong since 19.1

driusan:
@cmadjar can you review this? I'm not really sure of the implications of this change.

lingma:
@cmadjar In the program feedback_mri_popup, the program does ask for sessionID. Moreover, IBIS has a requirement which needs sessionID, as a result I provide sessionID for all of them, I did not notice problem, could you clarify on that point?

lingma:
@nicolasbrossard What do you think? We use sessionID then don't pass it further? And if this is for a normal procedure, then get sessionID and reset it to null?

lingma:
@cmadjar In fact, if we provide the sessionID, then the sessionID will be inserted into the database together with FileID. Since FileID will take the precedence over sessionID, I think the impact is limited, unless there is other program that I have not noticed.

cmadjar:
@lingma I tested your branch and this is not the behaviour I noticed in the feedback_mri_comments table. The `SessionID` was not being populated in the feedback_mri_comments table when the comment was file related, which is the behaviour we want. 

From what I understand of your code, the only change associated to the SessionID is the URL of the feedback popup. Is there something I don't get?

@johnsaigle I removed the `bug` label as this is not a bug fix.

lingma:
> From what I understand of your code, the only change associated to the SessionID is the URL of the feedback popup. Is there something I don't get?

Right, I only added the SessionID in the URL, and I just rechecked the code, you are right. The SessionID is not set in the insert table part, there is a `if else` section

lingma:
This PR is not necessary for Loris branch, the modification is necessary for IBIS. I think maybe we just close this PR will be OK.

PR #6008, opened 2020-01-31
[new profile] fix for single site affiliation bug

### LINKED ISSUE #5969: [22.0.0 - new profile] Empty drop down with only 1 site affiliation

ridz1208:
## Brief summary of changes
Fix issue where empty dropdown shows up in the module when the logged in user is only affiliated to 1 site.

I'm not a big fan of hiding fields with a single option, but judging by the code, that was the expected behaviour. So this is the smallest possible change to fix the bug on 22

1. test items 5 and 6 of the testplan

* Resolves #5969


modules/new_profile/test/TestPlan.md
```diff
@@ -0,0 +1,27 @@
+# New Profile Test Plan:
+
+1. Access New Profile page and ensure that it renders.
+   [Automation Testing]
+2. Access with useEDC on/off and ensure that the field appears/disappears
+   [Manual Testing]
```
PapillonMcGill: @ridz1208 Not a blocker but could you ensure that all phrases either end with a dot (preferably) or do not end with a dot?

ridz1208: @PapillonMcGill ma bad. Fixed it

ridz1208:
@PapillonMcGill did you run make ?

the change is in the react file

PapillonMcGill:
yup!

PR #6011, opened 2020-02-03
[php/libraries/NDB_Menu_Filter.class.inc] filter undefined fix

maltheism:
## Brief summary of changes

Fixes this error I received today when visiting the user_accounts module.
`Notice: Undefined index: filter in /Users/alizee/Development/GitHub/McGill/Loris/php/libraries/NDB_Menu_Filter.class.inc on line 598`

1. Visit the user_accounts module on fresh rasinbread installation with only admin account.
2. Check error_log


php/libraries/NDB_Menu_Filter.class.inc
```diff
@@ -595,7 +595,7 @@ class NDB_Menu_Filter extends NDB_Menu
         $this->setTemplateVar('rowsPerPage', $rowsPerPage);
         $this->setTemplateVar('pageID', $pageID);
 
-        if (is_array($_GET['filter']['order'])) {
+        if (isset($_GET['filter']) && is_array($_GET['filter']['order'])) {
```
driusan: `$_GET['filter']` needs to be checked to be an array, not just set.

maltheism: @driusan yah both are checked. But currently I can get an error since it can be undefined.

driusan: Both aren't checked. `$_GET['filter']` is only checked to be set, `$_GET['filter']['order']` is checked to be an array. `$_GET['filter']` needs to be both set and an array for `$_GET['filter']['order']` to be valid regardless of whether or not `$_GET['filter']['order']` is an array.

driusan:
Closing since @maltheism approved #6047

PR #6016, opened 2020-02-03
Remove package-lock.json

driusan:
The package-lock file does not lock packages, but gets updated
every time `npm install` is run. As a result, the primary purpose
of including package-lock in the repository appears to be to cause
conflicts and arbitrary changes amongst different developer
environments.

This removes the file.


maltheism:
@driusan I'm in favour of removing it. Similar discussion to removing the file was in this PR https://github.com/aces/Loris/pull/5820. I think @johnsaigle wants to keep the package-lock.json because it's standard practice to have it. Since there is some security in regards to people that checkout the codebase should be getting the dependencies from where the developers last received them.

driusan:
He seems to be mistaken about what the file does, then. It gets overwritten and updated every time `npm install` is run.

driusan:
I was waiting for @johnsaigle to weigh in since he was the one who objected when @xlecours tried to do it.. as far as I know there's now no one in favour of keeping it.

PR #6017, opened 2020-02-03
[added function] Group error message cell positioning

lingma:
## Brief summary of changes

When working with Loris Form, one can only position the error message to row level or to element level if this is a standalone element. 

If the items is in a big table, Loris doesn't support cell level error message positioning, though the table rendering is supported and the row level error message could somehow be supported. Because of the fact that here we have two interpretations for the same id (element vs table row), the idea is not quite clear because one needs to find the table row id instead of element id to be able to show positioning error message. this could create some confusion for new developers.

The modification here will add this capacity, with the following formats:

`$errors[<table id>] = ["<cell id>" => "<error message>"]`

or `$errors[<table id>] = ["<cell id 1>" => "<error message 1>", "<cell id 2>" => "<error message 2>"]`

I suppose that this could be an added capacity for form processing.

1. Make a item like before, test error positioning capacity, in additional form validation instead of XIN rules.

2. Make a grouped item, with 1 and 2 items to verify the validation error message positioning capacity

This is the recreation of the PR #5882

### DIFF
--- a/php/libraries/LorisForm.class.inc
+++ b/php/libraries/LorisForm.class.inc
@@ -2200,17 +2200,33 @@ class LorisForm
      * Sets an error for the given element to be displayed to the
      * user
      *
-     * @param string $el    The name of the element which has an error
-     * @param string $error The error to display with the element.
+     * @param string       $el    The name of the element which has an error.
+     * @param string|array $error The error to display with the element.
+     *                            string: the error message to show;
+     *                            array: an array of element to error message,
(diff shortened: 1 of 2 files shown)


jesscall:
Looks good! 

Using the format specified in description; here is what it looks like:

Before:
Screen Shot 2020-08-04 at 11 11 33 AM


After:
Screen Shot 2020-08-04 at 11 18 35 AM
Screen Shot 2020-08-04 at 11 19 45 AM

PR #6019, opened 2020-02-04
Remove BaseURL config setting

### LINKED ISSUE #5395: Accessing loris via IP breaks interface if configured via hostname and vice-versa if configured via hostname
Main login page is okay but after logging in page rendering is broken in various ways:
image

Furthermore, my existing authentication done against one host connection doesn't translate to the other connection.

Seems like something hard-coded somewhere based on the backend configuration, other web-based systems I've used don't seem to have this sort of limitation.

### LINKED ISSUE #6007: [Installdb] hostname as a choice to be entered, the default as the visiting hostname

It would be nice that the installdb have a hostname field, default to be the current visiting hostname, editable to be any other hostname.

Add a text field, copy the current hostname as the default value. Make the field editable, then insert the value from this field instead of only the installer's hostname. This could give user a choice as to what the expected visiting hostname should be, instead of going into DB to modify.

**Describe alternatives you've considered**
You can also put the hostname in the config.xml file, so that it is much easy to modify the value for different purpose.

**Additional context**
A customer Sotirios Nikoloutsopoulos is creating Loris on dockered environment, which could be much easy if the field hostname is editable.
The raisinbread will be installed as `localhost`, which could be much better if we don't need to modify Config table each time after the raisinbread db loading.

driusan:
The \LORIS\BaseRouter was already calculating the BaseURL of the incoming request and adding it to the request object. Unfortunately, most places in LORIS are not using it. Most instances would need to be refactored to get access to it since they're often deeper in the code and using `\NDB_Factory->settings()->getBaseURL()` at places with no direct access to the PSR7 request instance. This updates the code to have the router to pass the calculated URL to the factory singleton instance which injects it into the Settings object.

There were a very straggling instances to the (even older) `$config->getSetting('www')['url']` style of getting the baseURL which had to be updated to use the factory.

Fixes #5395, #6007.

# Testing Instructions
1. Test LORIS. Everything should work as before.
2. Test LORIS via another hostname (ie. use the IP instead of the hostname or vice versa). Everything should still work.

### DIFF
--- a/SQL/New_patches/2020-02-05-NoHostConfig.sql
+++ b/SQL/New_patches/2020-02-05-NoHostConfig.sql
@@ -1,2 +1,0 @@
-DELETE FROM Config WHERE ConfigID IN (Select ID FROM ConfigSettings WHERE Name='url');
-DELETE FROM ConfigSettings WHERE Name='url';
--- a/SQL/New_patches/2020-02-05-NoHostConfig.sql
+++ b/SQL/New_patches/2020-02-05-NoHostConfig.sql
@@ -1,0 +1,2 @@
+DELETE FROM Config WHERE ConfigID IN (Select ID FROM ConfigSettings WHERE Name IN ('host', 'url'));
+DELETE FROM ConfigSettings WHERE Name IN ('host', 'url');
--- a/SQL/New_patches/2020-02-05-NoHostConfig.sql
+++ b/SQL/New_patches/2020-02-05-NoHostConfig.sql
@@ -1,2 +1,2 @@
-DELETE FROM Config WHERE ConfigID IN (Select ID FROM ConfigSettings WHERE Name IN ('host', 'url'));
-DELETE FROM ConfigSettings WHERE Name IN ('host', 'url');
+DELETE FROM Config WHERE ConfigID IN (Select ID FROM ConfigSettings WHERE Name='url');
+DELETE FROM ConfigSettings WHERE Name='url';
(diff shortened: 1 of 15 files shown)


driusan:
@ridz1208 Can you have a look at this? I think I found all references and fixed them, but you're good at finding things I overlooked.

gdevenyi:
Also maybe #5395?

driusan:
Yeah, thanks. It should fix that one too. I updated the description.

SQL/New_patches/2020-02-05-NoHostConfig.sql
```diff
@@ -0,0 +1,2 @@
+DELETE FROM Config WHERE ConfigID IN (Select ID FROM ConfigSettings WHERE Name='url');
```
ridz1208: i think this should be in cleanup not new patches

driusan: done

driusan:
@ridz1208 I updated all of those except for the one in LorisIntegrationTest, because that usage is different (it's where the tests are pointing selenium to) and can remain in the test config even if it's not used by LORIS

ridz1208:
@johnsaigle this might impact #5260

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -566,14 +566,17 @@ class Edit_User extends \NDB_Form
...
             }
         }
+
+        $factory = \NDB_Factory::singleton();
+        $baseURL = $factory->settings()->getBaseURL();
         // send the user an email
         if (!empty($send)) {
-            // create an instance of the config object
-            $config = \NDB_Config::singleton();
+            $config = $factory->singleton();
```
ridz1208: this doesnt look right

ridz1208:
```suggestion
            $config = $factory->config();
```

PR #6024, opened 2020-02-06
[SQL] Remove multipleSites config setting

driusan:
I could not find any references to this config setting in the LORIS
codebase. The only references are in SQL patches adding it. It's
also not a very useful config setting, since it would be more reliable
to just check whether the psc table has multiple entries or not.


ridz1208:
@driusan i remember Mouna saying that it wasn't used anywhere.

driusan:
@cmadjar can you confirm on the Loris-MRI side?

driusan:
@cmadjar

driusan:
@cmadjar now that the release is done would you have a couple minutes to look into this?

PR #6027, opened 2020-02-06
[Phan] Enable dead code detection and solve some occurrences

johnsaigle:
This enables Phan's ability to detect unused variables and dead code. As with default phan rules, some new rules have been suppressed and we will gradually enable these.

The following Phan rules are now checked and existing violations have been fixed:
* PhanUnusedProtectedMethodParameter
* PhanUnusedGlobalFunctionParameter
* PhanUnusedClosureParameter

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -27,11 +27,6 @@ return [
     //
     // When the module is refactored, this line should be deleted.
     "ignore_undeclared_variables_in_global_scope" => true,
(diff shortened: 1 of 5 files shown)


driusan:
Are you sure that deleting those lines enables them? I thought I tried explicitly setting them to true the other day and got a bunch of errors..

johnsaigle:
You're right I must have been distracted.

PR #6028, opened 2020-02-06
Add PhanPossiblyUndeclaredVariable rule to Phan checks

johnsaigle:
## Brief summary of changes

This rule checks for a case where a variable may not be declared, for example if it is declared only within an if-statement, but the code below it uses it anyway.

e.g.

```php

if ($a === $b) {
    $c = 10;
}

echo $a . $c;
```

The code assumes `$c` exists but this may not always be the case.

### DIFF
--- a/modules/examiner/php/editexaminer.class.inc
+++ b/modules/examiner/php/editexaminer.class.inc
@@ -112,18 +112,12 @@ class EditExaminer extends \NDB_Form
     {
         $DB = \Database::singleton();
 
+        // get the examinerID
+        $examinerID = $this->identifier;
         foreach ($values['certStatus'] as $testID => $certStatus) {
 
             $date_cert = $values['date_cert'][$testID];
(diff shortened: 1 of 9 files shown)


modules/examiner/php/editexaminer.class.inc
```diff
@@ -112,19 +112,13 @@ class EditExaminer extends \NDB_Form
...
 
+        // get the examinerID
+        $examinerID = $this->identifier;
         foreach ($values['certStatus'] as $testID => $certStatus) {
 
             $date_cert = $values['date_cert'][$testID];
             $comment   = trim($values['comment'][$testID]);
 
-            // get the examinerID - edit_examiner passes the ID through
-            // the identifier, training passes it as a value
```
johnsaigle: My impression is this comment is referring to two modules: `edit_examiner` and the deleted `training` module.

Since only `edit_examier` now exists, I declared `examinerID` above using the "Edit examiner" method.

PR #6029, opened 2020-02-06
Add PhanUndeclaredClassMethod to phan checks

johnsaigle:
This makes sure that classes really exist when we try to use them.

### DIFF
--- a/modules/instrument_list/php/instrument_list_controlpanel.class.inc
+++ b/modules/instrument_list/php/instrument_list_controlpanel.class.inc
@@ -117,9 +117,7 @@ class Instrument_List_ControlPanel extends \TimePoint
             && in_array($currentStage, array('Visit', 'Screening'))
             && $currentStatus != 'In Progress'
         ) {
-            $config    = \NDB_Config::singleton();
-            $study     = $config->getSetting('study');
-            $sessionID = $this->getData('SessionID');
+            $config = \NDB_Config::singleton();
 
             // Send to DCC - set Submitted to Y
             $this->setData(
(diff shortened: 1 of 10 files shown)


modules/instrument_list/php/instrument_list_controlpanel.class.inc
```diff
@@ -117,57 +117,7 @@ class Instrument_List_ControlPanel extends \TimePoint
             && in_array($currentStage, array('Visit', 'Screening'))
             && $currentStatus != 'In Progress'
         ) {
-
```
PapillonMcGill: Is suppression of the following lines (120-170) really part of this PR?

johnsaigle: Yes there was something about calling singleton on `TimePoint_Flag`....  a module that does not exist.

I think this section of the code is very old and somehow escaped cleanup.

php/libraries/NDB_Factory.class.inc
```diff
@@ -25,6 +25,11 @@ use \LORIS\StudyEntities\Candidate\CandID;
  * @author   Dave MacFarlane <david.macfarlane2@mcgill.ca>
  * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
  * @link     https://www.github.com/aces/Loris/
+ *
+ * Phan currently has bad support for PHPUnit's Mock objects which is
+ * creating false positive with the below rule.
+ *
+ * @phan-file-suppress PhanUndeclaredClassMethod
```
driusan: I think this is a rare reasonable use of suppressing the rule.. but at some point we should probably fix the code to not be so intertwined with PHPUnit.

PR #6031, opened 2020-02-07
[stats] fixed only_full_group_by error in query

ridz1208:
## Brief summary of changes
on MySQL 5.7 with SQL MODE `only_full_group_by` I get 500 errors on the dashboard for the query that was altered in this PR

1. open error_log
2. make sure your SQL mode is properly set to `only_full_group_by`
2. access the dashboard.

**NOTE** Whomever this affects should verify that the results on the query are unchanged from this PR

### DIFF
--- a/modules/statistics/php/charts.class.inc
+++ b/modules/statistics/php/charts.class.inc
@@ -182,8 +182,8 @@ class Charts extends \NDB_Page
             LEFT JOIN session s ON (s.ID=f.SessionID)
             JOIN parameter_type pt USING (ParameterTypeID)
             WHERE pt.Name='acquisition_date'
-            GROUP BY MONTH(pf.Value), YEAR(pf.Value), s.CenterID
-            ORDER BY YEAR(pf.Value), MONTH(pf.Value), s.CenterID",
+            GROUP BY datelabel, s.CenterID
+            ORDER BY datelabel, s.CenterID",
             array()
         );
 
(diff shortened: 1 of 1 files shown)


modules/statistics/php/charts.class.inc
```diff
@@ -182,8 +182,8 @@ class Charts extends \NDB_Page
             LEFT JOIN session s ON (s.ID=f.SessionID)
             JOIN parameter_type pt USING (ParameterTypeID)
             WHERE pt.Name='acquisition_date'
-            GROUP BY MONTH(pf.Value), YEAR(pf.Value), s.CenterID
-            ORDER BY YEAR(pf.Value), MONTH(pf.Value), s.CenterID",
+            GROUP BY datelabel, s.CenterID
```
driusan:
```
// [...]. Order matters to ensure that the
// labels are calculated in the correct order.
```

This changes the ordering since it goes from ordering by year, month (as integers) to ordering by month-year (as a string, so months are in the wrong order.)

ridz1208: The issue is that the value in the select is different then the one in the group by, it seems like MySQL is only happy if u group by the exact value in the select statement. Grouping by one then the other is different then grouping by both as a string but that's what it's selecting for

Any suggestions ?

ridz1208: I could select for them separately and have the code concat them...

driusan: can you add a group by on that without removing the others `GROUP BY MONTH(pf.Value), YEAR(pf.Value), s.CenterID, datelabel`?

driusan: the problem is selecting on things that aren't in the group by, not grouping by things that aren't in the select.

ridz1208: @driusan done

ridz1208:
@driusan please re review and merge if passes

PR #6034, opened 2020-02-07
[Instruments] Handle NotFound exception causing LORIS to crash

johnsaigle:
## Brief summary of changes

When a candidate has an entry for an instrument in `instrument_list` but the instrument is not installed, a crash occurs if the user clicks on this entry. The back-end displays the error `PHP Fatal error:  Uncaught NotFound: Instrument does not exist ...`

This PR adds error handling to return the message `{"error":"Instrument does not exist"}` instead of a crash.

Further work should be done to provide SWAL error handling for this module.

1. On a VM with raisinbread, go to `/instrument_list/?candID=300005&sessionID=5`.
2. Under "Behavioral Battery of Instruments" click Radiology review
3. If this instrument does not exist, LORIS will crash
4. On my branch you'll see instead a JSON response ``{"error":"Instrument does not exist"}``

### DIFF
--- a/modules/instruments/php/module.class.inc
+++ b/modules/instruments/php/module.class.inc
@@ -59,15 +59,21 @@ class Module extends \Module
         $params    = $request->getQueryParams();
         $commentID = $params['commentID'] ?? null;
         if (empty($commentID)) {
-            return (new \Laminas\Diactoros\Response())
-                ->withBody(new \LORIS\Http\StringStream("Missing CommentID"))
-                ->withStatus(400);
+            return new \LORIS\Http\Response\JSON\BadRequest(
+                "Missing CommentID"
+            );
         }
-        $instrument = \NDB_BVL_Instrument::factory(
-            $instrumentName,
-            $commentID,
-            $page
-        );
+        try {
+            $instrument = \NDB_BVL_Instrument::factory(
+                $instrumentName,
+                $commentID,
+                $page
+            );
+        } catch (\NotFound $e) {
+            return new \LORIS\Http\Response\JSON\NotFound(
+                $e->getMessage()
+            );
+        }
 
         $request = $request->withAttribute('pageclass', $instrument);
 
--- a/modules/instruments/php/module.class.inc
+++ b/modules/instruments/php/module.class.inc
@@ -69,15 +69,14 @@ class Module extends \Module
                 $commentID,
                 $page
             );
+            $request    = $request->withAttribute('pageclass', $instrument);
+
+            return $instrument->process($request, $instrument);
         } catch (\NotFound $e) {
             return new \LORIS\Http\Response\JSON\NotFound(
                 $e->getMessage()
             );
         }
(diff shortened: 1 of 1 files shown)


driusan:
I'm not convinced that we want to catch this. The server is misconfigured if the candidate is referencing instruments that don't exist and dying loudly makes sure it gets caught, silently returning an HTTP error code indicating failure won't ever get to the admin.

@ridz1208 @xlecours thoughts?

johnsaigle:
I think ideally it would report a decorated page that says "Please contact the admin" at the bottom

xlecours:
@driusan loud yes, but decorated like @johnsaigle said. A middleware to catch catchable 500 error would be nice

modules/instruments/php/module.class.inc
```diff
@@ -59,15 +59,21 @@ class Module extends \Module
...
+            return new \LORIS\Http\Response\JSON\BadRequest(
+                "Missing CommentID"
+            );
+        }
+        try {
+            $instrument = \NDB_BVL_Instrument::factory(
+                $instrumentName,
+                $commentID,
+                $page
+            );
```
driusan: Can you move the 2 lines

```
$request = $request->withAttribute('pageclass', $instrument);
return $instrument->process($request, $instrument);
```

into the try block so that the `$instrument` variable is unambiguously in scope?

johnsaigle: Done

modules/instruments/php/module.class.inc
```diff
@@ -59,19 +59,24 @@ class Module extends \Module
         $params    = $request->getQueryParams();
         $commentID = $params['commentID'] ?? null;
         if (empty($commentID)) {
-            return (new \Laminas\Diactoros\Response())
-                ->withBody(new \LORIS\Http\StringStream("Missing CommentID"))
-                ->withStatus(400);
+            return new \LORIS\Http\Response\JSON\BadRequest(
```
driusan: This is not an appropriate response here, the response is not JSON, it's an error message for the client.

modules/instruments/php/module.class.inc
```diff
@@ -59,19 +59,24 @@ class Module extends \Module
...
+                $instrumentName,
+                $commentID,
+                $page
+            );
+            $request    = $request->withAttribute('pageclass', $instrument);
 
-        return $instrument->process($request, $instrument);
+            return $instrument->process($request, $instrument);
+        } catch (\NotFound $e) {
+            return new \LORIS\Http\Response\JSON\NotFound(
```
driusan: JSON is also not appropriate here.

johnsaigle:
This can probably be closed if #6076 gets merged. It should catch this case.

driusan:
I don't think this should be closed. We should explicitly catch the exception at a higher level if possible.

ridz1208:
as per johns suggestion

PR #6035, opened 2020-02-07
[Utility] Add pathJoin() function

### LINKED ISSUE #5873: Create PHP equivalent to os.path.join (python) as Utility function
We should create a function to properly build and resolve paths in LORIS. 

The idea is to model our function off of e.g. os.path.join in python. See this [blog post](https://www.geeksforgeeks.org/python-os-path-join-method/) for intended behaviour.

This will also make it easier to verify that paths to directories end with a forward slash.

The function should be unit tested.

---

> As far as I know PHP doesn't have a built in equivalent. I think we should write one with appropriate unit tests, but I don't think this PR is the right place to do it. That's why I think these ones should just be fixed locally and then another issue/PR can tackle doing it "right" throughout LORIS

_Originally posted by @driusan in https://github.com/_render_node/MDI0OlB1bGxSZXF1ZXN0UmV2aWV3Q29tbWVudDM1OTAwNDE3Nw==/comments/review_comment_

johnsaigle:
## Brief summary of changes

Adds a utility function for intelligently joining paths. See linked ticket for more details.

* Resolves #5873

### DIFF
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -1027,28 +1027,17 @@ class Utility
     /**
      * A PHP implementation of python's os.path.join.
      *
-     * "Join one or more path components intelligently. The return value is
-     * the concatenation of path and any members of *paths with exactly one
-     * directory separator (os.sep) following each non-empty part except the
-     * last, meaning that the result will only end in a separator if the last
-     * part is empty. If a component is an absolute path, all previous
-     * components are thrown away and joining continues from the absolute
-     * path component."
+     * @param string ...$paths The path components to be joined.
      *
-     * @param string $path     The primary apth to which $parts will be joined.
-     * @param string ...$paths The path components to be joined to $path.
-     *
-     * @return string $path joined intelligently to $paths.
+     * @return string Path components joined without any doubled forward-slashes.
      */
-    public static function pathJoin(string $path, string ...$paths)
+    public static function pathJoin(string ...$paths)
     {
         // This array stores the components of the final path.
-        $result = [$path];
+        $result = [];
         foreach ($paths as $part) {
-            // If this part is an absolute path, discard $result.
-            if (mb_strpos($part, '/') === 0) {
-                $result = [];
-            }
+            // Remove all leading forward-slashes from $part.
+            $part = ltrim($part, '/');
             // Remove trailing forward-slashes from $part.
             while (mb_substr($part, -1) === '/') {
                 $part = chop($part, '/');
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -1047,7 +1047,7 @@ class Utility
             }
             $result[] = $part;
         }
-        
+
         // If the first part began with a front-slash, add it back
         return ($addSlash ? '/' : '') . implode('/', $result);
     }
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -1035,16 +1035,21 @@ class Utility
     {
         // This array stores the components of the final path.
         $result = [];
+
+        $addSlash = (strpos($paths[0], '/') === 0);
+
         foreach ($paths as $part) {
             // Remove all leading forward-slashes from $part.
             $part = ltrim($part, '/');
             // Remove trailing forward-slashes from $part.
             while (mb_substr($part, -1) === '/') {
-                $part = chop($part, '/');
+                $part = rtrim($part, '/');
             }
             $result[] = $part;
         }
-        return implode('/', $result);
+        
+        // If the first part began with a front-slash, add it back
+        return ($addSlash ? '/' : '') . implode('/', $result);
     }
 }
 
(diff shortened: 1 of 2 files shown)


php/libraries/Utility.class.inc
```diff
@@ -1008,5 +1008,39 @@ class Utility
...
+     *
+     * "Join one or more path components intelligently. The return value is
+     * the concatenation of path and any members of *paths with exactly one
+     * directory separator (os.sep) following each non-empty part except the
+     * last, meaning that the result will only end in a separator if the last
+     * part is empty. If a component is an absolute path, all previous
+     * components are thrown away and joining continues from the absolute
+     * path component."
+     *
+     * @param string $path     The primary apth to which $parts will be joined.
```
```suggestion
     * @param string $path     The primary path to which $parts will be joined.
```

test/unittests/UtilityTest.php
```diff
@@ -858,4 +858,37 @@ public function testValueIsPositiveIntegerReturnsTrue($int): void
     {
         $this->assertTrue(\Utility::valueIsPositiveInteger($int));
     }
+
+    /**
+     * Tests the pathJoin function. Test cases adapted from blog post on
+     * Python's os.path.join as this function is meant to give the same
+     * beahviour.
```
driusan: The behaviour of the edge cases you're using in this test don't make sense to me and are far less intuitive than Go's `filepath.Join`. Just because Python does something doesn't mean it's the right way to do it.

johnsaigle: OK, you originally asked for this in another PR which I quoted to make the issue. The link to the original PR seems to be broken so I'm not sure of the exact context.

Should I change this to behave like Go then?

php/libraries/Utility.class.inc
```diff
@@ -1008,5 +1008,28 @@ class Utility
...
         return rtrim($path, '/\\') . '/';
     }
+
+    /**
+     * A PHP implementation of python's os.path.join.
+     *
+     * @param string ...$paths The path components to be joined.
+     *
+     * @return string Path components joined without any doubled forward-slashes.
+     */
```
johnsaigle:
```suggestion
     * A PHP implementation of Go's filepath Join function.
     *
     * @param string ...$paths The path components to be joined.
     *
     * @return string Path components joined without any doubled forward-slashes.
     *
     * @see https://golang.org/pkg/path/filepath/#Join
     */
```

test/unittests/UtilityTest.php
johnsaigle:
```suggestion
     * Tests the pathJoin function. Test cases adapted from Go's documentation.
```

test/unittests/UtilityTest.php
```diff
@@ -858,4 +858,37 @@ public function testValueIsPositiveIntegerReturnsTrue($int): void
...
+     *
+     * @see https://www.geeksforgeeks.org/python-os-path-join-method/
```
johnsaigle:
```suggestion
     * @see https://golang.org/pkg/path/filepath/#Join
```

johnsaigle:
@driusan This function now behaves like a mole instead of a snake

php/libraries/Utility.class.inc
```diff
@@ -1008,5 +1008,29 @@ class Utility
...
+     *
+     * @return string Path components joined without any doubled forward-slashes.
+     *
+     * @see https://golang.org/pkg/path/filepath/#Join
+     */
+    public static function pathJoin(string ...$paths)
+    {
+        // This array stores the components of the final path.
+        $result = [];
+        foreach ($paths as $part) {
```
driusan, line 1025: This doesn't look like it'll properly handle paths with a leading "/". Can you add a test case for it?

php/libraries/Utility.class.inc
```diff
@@ -1008,5 +1008,29 @@ class Utility
...
+            // Remove all leading forward-slashes from $part.
+            $part = ltrim($part, '/');
+            // Remove trailing forward-slashes from $part.
+            while (mb_substr($part, -1) === '/') {
+                $part = chop($part, '/');
```
driusan, line 1030: The docs say chop is an alias for rtrim, you should probably just use rtrim.

php/libraries/Utility.class.inc
johnsaigle: Yes, that's a great idea

johnsaigle:
@driusan I incorporated your suggestions

php/libraries/Utility.class.inc
```diff
@@ -1035,5 +1035,33 @@ class Utility
     {
         return rtrim($path, '/\\') . '/';
     }
+
+    /**
+     * A PHP implementation of python's os.path.join.
```
driusan, line 1040: The comment here shouldn't reference other languages, it should just describe what the function does.

johnsaigle, line 1040:
```suggestion
     * Joins strings together to create a well-formed filepath.
```

PR #6036, opened 2020-02-07
[New Profile] Fix issue where candidate creation failed. Also, update phpstan + fix errors

### LINKED ISSUE #6059: [CORE] Can't generate a new candidate
**Describe the bug**
get `Uncaught DomainException: Values not configured properly for PSCIDGenerator. Please correct your configuration file.Length: `4`\nAlphabet: `0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ`\nMin: ``\nMax: `9999`\nLength: `CHA`\nGeneration: `sequential`\n in `

error reported on PSCID generation.

**To Reproduce**
1. Go to 'Candidate -> new profile'
2. Set the proper values for the new candidate
3. Click on `Create`
4. button `gray-out` and nothing append, error in the log

A new candidate to be created and page update with candidate's profile

- LORIS Version: [aces/master, last commit at time]
- Linux distribution and Version: [e.g. Ubuntu 16.04]
- MySQL/MariaDB Version: [e.g. MySQL 5.7]

**Additional context**
ran `checkout aces/master`, `make` and `php raisinbread_refresh.php`

config.xml file with proper value as indicated on docs/config/config.xml
```
   <!-- study variables -->
    <study>
        <PSCID>
            <!-- PSCID (Study Center ID) generation method possible options: sequential/random/user -->
            <generation>sequential</generation>
            <structure>
                <seq type="projectAbbrev"/>
                <seq type="alphanumeric" length="4" min='1' max='9999'/>  <!-- Ex: AAA1-->
            </structure>
        </PSCID>

        <ExternalID>
            <generation>random</generation>
            <structure>
                <seq type="static">BBB</seq>
                <seq type="numeric" length="4" min='1000' max='9999'/>
            </structure>
        </ExternalID>
```

johnsaigle:
## Brief summary of changes

Updates phpstan to the latest version.

SiteIDGenerator had to undergo refactoring to satisfy phpstan's strict type checks but should function the same as before.

By chance this also fixes #6059

### DIFF
--- a/php/libraries/CandIDGenerator.php
+++ b/php/libraries/CandIDGenerator.php
@@ -26,7 +26,7 @@ use \LORIS\StudyEntities\Candidate\CandID;
 class CandIDGenerator extends IdentifierGenerator
 {
     /**
-     * CandIDs should always be exactly 6 integers.
+     * CandIDs should always be exactly 6 digits.
      */
     private const LENGTH     = 6;
     private const MIN_CANDID = '100000';
--- a/php/libraries/CandIDGenerator.php
+++ b/php/libraries/CandIDGenerator.php
@@ -26,11 +26,11 @@ use \LORIS\StudyEntities\Candidate\CandID;
 class CandIDGenerator extends IdentifierGenerator
 {
     /**
-     * CandIDs should always be exactly 6 digits.
+     * CandIDs should always be exactly 6 integers.
      */
     private const LENGTH     = 6;
-    private const MIN_CANDID = 100000;
-    private const MAX_CANDID = 999999;
+    private const MIN_CANDID = '100000';
+    private const MAX_CANDID = '999999';
 
     /**
      * Creates a new CandIDGenerator by initializing properties based on class
@@ -42,7 +42,7 @@ class CandIDGenerator extends IdentifierGenerator
         $this->length           = self::LENGTH;
         $this->minValue         = self::MIN_CANDID;
         $this->maxValue         = self::MAX_CANDID;
-        $this->alphabet         = range(0, 9);
+        $this->alphabet         = range('0', '9');
     }
 
     /**
@@ -58,7 +58,7 @@ class CandIDGenerator extends IdentifierGenerator
             $this->checkIDRangeFull();
             $id = new CandID(
                 strval(
-                    random_int($this->minValue, $this->maxValue)
+                    random_int(intval($this->minValue), intval($this->maxValue))
                 )
             );
             // Check if the ID is in use. If so, the loop will continue and a new
--- a/php/libraries/IdentifierGenerator.php
+++ b/php/libraries/IdentifierGenerator.php
@@ -113,21 +113,6 @@ abstract class IdentifierGenerator
             break;
         }
 
-        if (!empty($this->siteAlias) && !empty($this->projectAlias)) {
-            throw new \InvalidArgumentException(
-                'Only one of siteAlias or projectAlias should have a value'
-            );
-        }
-        if (empty($this->siteAlias) && empty($this->projectAlias)) {
-            throw new \InvalidArgumentException(
-                'Either siteAlias or projectAlias must not be empty'
-            );
-        }
-        // Assign class property 'prefix' to whichever alias is not empty.
-        !empty($this->siteAlias) ?
-            $this->prefix   = $this->siteAlias
-            : $this->prefix = $this->projectAlias;
-
         return $this->prefix . $id;
     }
 
--- a/php/libraries/IdentifierGenerator.php
+++ b/php/libraries/IdentifierGenerator.php
@@ -37,7 +37,7 @@ abstract class IdentifierGenerator
     /**
      * An array of all the valid characters that can be used in the ID.
      *
-     * @var array<int,float|int> This is derived from the return type of PHP's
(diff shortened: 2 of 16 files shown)


johnsaigle:
```
modules/statistics/php/stats_behavioural.class.inc:157 PhanTypeMismatchDimFetch When fetching an array index from a value of type string, found an array index of type 'total', but expected the index to be of type int
modules/statistics/php/stats_behavioural.class.inc:223 PhanTypeMismatchDimFetch When fetching an array index from a value of type string, found an array index of type 'total', but expected the index to be of type int
```

Basically phan thinks that these lines in the stats module is being accessed with the wrong type of key. For some reason it thinks that there should be an `int` type key but it's actually a string.

The array it has an issue with is a **reference** to`&$tpl_data`, so I don't think there's any place where I can type hint this to make it more clear for Phan. 

I'm not sure how to fix this.

php/libraries/CandIDGenerator.php
```diff
@@ -26,11 +26,11 @@
 class CandIDGenerator extends IdentifierGenerator
 {
     /**
-     * CandIDs should always be exactly 6 digits.
+     * CandIDs should always be exactly 6 integers.
```
PapillonMcGill: I prefer `6 digits`. Or at least `6 digits integer value`

php/libraries/SiteIDGenerator.php
```diff
@@ -233,19 +221,19 @@ private function _getIDSetting(
                 );
             }
         }
-        // Min, max, and length values should be returned as integers or as
+        // The remaining valeues - min, max, and length - should be returned as
```
PapillonMcGill:
```suggestion
        // The remaining values - min, max, and length - should be returned as
```

php/libraries/IdentifierGenerator.php
```diff
@@ -28,13 +28,59 @@ abstract class IdentifierGenerator
...
+    /**
+     * How the ID will be generated. 'sequential' or 'random'.
+     *
+     * @var string
+     */
     protected $generationMethod;
+    /**
+     * An array of all the valid characters that can be used in the ID.
+     *
+     * @var array<int,float|int> This is derived from the return type of PHP's
```
PapillonMcGill: Are you sure `$alphabet` will either contain int or float values? I would have guest `char` would be more likely

johnsaigle: I'll look into it. It's possible that PHP stores this internally as ASCII because i remember phpstan giving an error about this.

johnsaigle:
Marking this with Priority High since without it new candidates can't be created. #6059

johnsaigle:
@PapillonMcGill I updated the function document. Can you update your review?

johnsaigle:
@driusan Can you review?

php/libraries/IdentifierGenerator.php
```diff
@@ -67,21 +113,6 @@ protected function createNewID(): string
...
-        if (empty($this->siteAlias) && empty($this->projectAlias)) {
-            throw new \InvalidArgumentException(
-                'Either siteAlias or projectAlias must not be empty'
-            );
-        }
-        // Assign class property 'prefix' to whichever alias is not empty.
-        !empty($this->siteAlias) ?
-            $this->prefix   = $this->siteAlias
-            : $this->prefix = $this->projectAlias;
-
```
driusan: Removing this checkig looks like a regression. Why is it deleted?

php/libraries/UserPermissions.class.inc
```diff
@@ -90,11 +90,7 @@ class UserPermissions
         // Fill the permissions array of this UserPermissions object with the
         // values extracted for this user from the database.
         if (is_array($results) and !empty($results)) {
-            foreach ($results AS $row) {
-                if (empty($row['userID'])) {
-                    $this->permissions[$row['code']] = false;
-                    continue;
-                }
```
driusan: This also appears to be a regression

php/libraries/IdentifierGenerator.php
johnsaigle: This check was actually causing a bug reported in the linked issued. 

This validation is now done in the child class SiteIDGenerator instead. It's more appropriate here because only PSCIDs and ExternalIDs need Site/Project aliases for their prefix. By contrast, CandIDGenerator extends IdentifierGenerator but doesn't use this functionality at all.

php/libraries/UserPermissions.class.inc
johnsaigle: iirc I removed this because it's redundant. `$row['userID']` can be 0 which would trigger the empty check. But in that case, Line 94 will always be false because `$this->userID` is taken from the ID of the `users` table which will never be 0. If line 94 is false, then the `$this->permissions` array will be correctly set to be false for that permission code.

driusan: This is intended to unset any permissions which were unchecked on the request, it has nothing to do with userID 0 (which as far as I know should never exist)

johnsaigle: That's what I'm saying, I think line 94 and 95 (in the new code) are logically equivalent to what I deleted. These two lines alone will unset the permissions.

johnsaigle: Just to be clear I meant 94 and 95 on the right-hand side, not in the deleted code.

PR #6039, opened 2020-02-10
Add strict type coding standard

### LINKED ISSUE #5342: [PHPCS] Add sniff for declaration of strict_types

johnsaigle:
## Brief summary of changes

Previous attempts to [force type hints](#5396 ) or [strict typing](#5923) en masse have failed because it's just too darn complicated to do it all at once.

This introduces a new Strict Type coding standard that we can selectively apply to files by adding to an array in `test/run-php-linter.sh`. Hopefully this will provide an approach to using stricter typing that is easier to author and review.

Using `vendor/bin/phpcbf --standard=test/StrictTypesCS.xml $file` will now allow PHPCBF to automatically provide 
* return types
* param types 
* strict_types =1 

based on PHPDocs. 

It will also throw errors if Traversable types do not have specific type annotations.

The Database.class.inc file has been modified as an example of the changes 

### Related

Resolves #5342

### DIFF
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -56,13 +56,13 @@ class Database
      * NOTE There is no return type declaration for this function as it returns
      * the type `Database` which is not supported by PHPs built in return types.
      *
-     * @param string  $database     the database to select
-     * @param string  $username     the username with which to log into
-     *                              the database server
-     * @param string  $password     the password that matches the username
-     * @param string  $host         the name of the database server
-     * @param boolean $trackChanges boolean determining if changes should be
-     *                              logged to the history table
+     * @param string $database     the database to select
+     * @param string $username     the username with which to log into
+     *                             the database server
+     * @param string $password     the password that matches the username
+     * @param string $host         the name of the database server
+     * @param bool   $trackChanges boolean determining if changes should be
+     *                             logged to the history table
      *
      * @return \Database
      * @throws \DatabaseException
@@ -73,7 +73,7 @@ class Database
         string $username     = '',
         string $password     = '',
         string $host         = '',
-        bool   $trackChanges = true
+        bool $trackChanges = true
     ): \Database {
 
         static $connections = array();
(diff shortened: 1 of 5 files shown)


php/libraries/Database.class.inc
```diff
@@ -839,18 +864,19 @@ class Database
...
-     * @param array  $params Values to use for binding to prepared statement
+     * @param string               $query  The SQL SELECT query to be run
+     * @param array<string,string> $params Values to use for binding to prepared
+     *                                     statement
      *
      * @throws DatabaseException if the query selected more than one column
      *
-     * @return array Associative array of form rowID=>value containing all values
-     *               for the only column of the select statement
+     * @return array<string,string> Associative Format: rowID=>value containing
```
johnsaigle:
```suggestion
     * @return array<string,string> Format: rowID=>value containing
```

johnsaigle:
Blocked by #5356. Otherwise, when I try to update the `composer.lock` file it inevitably tries to update phpstan even though I specify only the `slevomat/coding-standard` package. I think the phpstan file may be wrong on master.

driusan:
#5356 is an issue (which I think is already resolved and needs to be closed?)

johnsaigle:
You're right, I meant #6036. I'll close the issue.

test/StrictTypesCS.xml
```diff
@@ -0,0 +1,36 @@
...
+     <rule ref="SlevomatCodingStandard.TypeHints.ReturnTypeHint">
+         <exclude name="SlevomatCodingStandard.TypeHints.ReturnTypeHint.UselessAnnotation"/>
+         <properties>
+             <property name="enableObjectTypeHint" value="false"/>
+         </properties>
+     </rule>
+
+     <rule ref="SlevomatCodingStandard.TypeHints.LongTypeHints"/>
+
+     <rule ref="SlevomatCodingStandard.TypeHints.ReturnTypeHintSpacing"/>
```
driusan, line 31: What does this have to do with enforcing strict types?

test/StrictTypesCS.xml
```diff
@@ -0,0 +1,36 @@
...
+
+     <rule ref="SlevomatCodingStandard.TypeHints.NullableTypeForNullDefaultValue"/>
+
+     <rule ref="SlevomatCodingStandard.TypeHints.ParameterTypeHintSpacing"/>
```
driusan, line 35: Or this?

test/StrictTypesCS.xml
johnsaigle: It makes the spacing of type hints consistent. More of an aesthetic thing. Do you want me to remove it?

https://github.com/slevomat/coding-standard/blob/d99e1afb03cbbd42ab877c222105016adf07c4fb/README.md#slevomatcodingstandardtypehintsreturntypehintspacing-

driusan: If it's already passing and didn't require any changes I guess it's fine, but the PR title says it's for adding strict_type.

PR #6042, opened 2020-02-10
[Consent] Add ParentID to `consent` for consent grouping

zaliqarosli:
## Brief summary of changes

**PR 1 of 2**

This PR:
- adds ConsentGroupID to consent table (schema + patch)
- updates RB_consent.sql to have consent group IDs
- adds RB_consent_group.sql to have 2 groups
- update getConsentList() function in \Utility to query only ConsentGroupID as well
- fixes PHP notice of undefined index in formHandler.php

This PR keeps the Consent Status tab rendering the same. Only the back-end is affected in this PR. This feature is increasingly in demand from projects who would like to split up consents via e.g. consent forms. Front-end changes are in PR #6044 (to be rebased after this PR gets merged)

1. On aces/main, test Consent Status tab of candidate_parameters module. See that you can fill in consent for a candidate for study_consent, raisin_consent, and bread_consent. If you don't see the Consent Status tab, turn on the _'useConsent'_ config.
2. Git checkout zaliqarosli/2020-02-10-AddParentIDToConsent
3. Source SQL/New_patches/2020-02-10-AddConsentGrouping.sql
4. Repeat the first step and ensure that nothing on the Consent Status tab has changed and that you can fill in consent the same as before

### DIFF
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -2059,13 +2059,26 @@ CREATE TABLE `feedback_mri_comments` (
 -- Consent tables
 -- ********************************
 
+CREATE TABLE `consent_group` (
+  `ConsentGroupID` integer unsigned NOT NULL AUTO_INCREMENT,
+  `Name` varchar(255) NOT NULL,
+  `Label` varchar(255) NOT NULL,
+  CONSTRAINT `PK_consent_group` PRIMARY KEY (`ConsentGroupID`),
+  CONSTRAINT `UK_consent_group_Name` UNIQUE KEY `Name` (`Name`),
+  CONSTRAINT `UK_consent_group_Label` UNIQUE KEY `Label` (`Label`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+INSERT INTO `consent_group` (`ConsentGroupID`, `Name`, `Label`) VALUES ('1', 'main_consent', 'Main consent');
+
 CREATE TABLE `consent` (
   `ConsentID` integer unsigned NOT NULL AUTO_INCREMENT,
   `Name` varchar(255) NOT NULL,
   `Label` varchar(255) NOT NULL,
+  `ConsentGroupID` integer unsigned NOT NULL DEFAULT 1,
   CONSTRAINT `PK_consent` PRIMARY KEY (`ConsentID`),
   CONSTRAINT `UK_consent_Name` UNIQUE KEY `Name` (`Name`),
-  CONSTRAINT `UK_consent_Label` UNIQUE KEY `Label` (`Label`)
+  CONSTRAINT `UK_consent_Label` UNIQUE KEY `Label` (`Label`),
+  CONSTRAINT `FK_consent_ConsentGroupID` FOREIGN KEY (`ConsentGroupID`) REFERENCES `consent_group` (`ConsentGroupID`) ON DELETE RESTRICT ON UPDATE CASCADE
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
 CREATE TABLE `candidate_consent_rel` (
@@ -2224,4 +2237,3 @@ CREATE TABLE `publication_users_edit_perm_rel` (
   CONSTRAINT `FK_publication_users_edit_perm_rel_PublicationID` FOREIGN KEY (`PublicationID`) REFERENCES `publication` (`PublicationID`),
   CONSTRAINT `FK_publication_users_edit_perm_rel_UserID` FOREIGN KEY (`UserID`) REFERENCES `users` (`ID`)
 ) ENGINE=InnoDB DEFAULT CHARSET='utf8';
-
--- a/SQL/New_patches/2020-02-10-AddConsentGrouping.sql
+++ b/SQL/New_patches/2020-02-10-AddConsentGrouping.sql
@@ -1,0 +1,12 @@
+CREATE TABLE `consent_group` (
+  `ConsentGroupID` integer unsigned NOT NULL AUTO_INCREMENT,
+  `Name` varchar(255) NOT NULL,
+  `Label` varchar(255) NOT NULL,
+  CONSTRAINT `PK_consent_group` PRIMARY KEY (`ConsentGroupID`),
+  CONSTRAINT `UK_consent_group_Name` UNIQUE KEY `Name` (`Name`),
+  CONSTRAINT `UK_consent_group_Label` UNIQUE KEY `Label` (`Label`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+INSERT INTO `consent_group` (`ConsentGroupID`, `Name`, `Label`) VALUES ('1', 'main_consent', 'Main consent');
+
+ALTER TABLE `consent` ADD `ConsentGroupID` integer unsigned NOT NULL DEFAULT 1;
+ALTER TABLE `consent` ADD CONSTRAINT `FK_consent_ConsentGroupID` FOREIGN KEY (`ConsentGroupID`) REFERENCES `consent_group` (`ConsentGroupID`) ON DELETE RESTRICT ON UPDATE CASCADE;
--- a/modules/candidate_parameters/ajax/formHandler.php
+++ b/modules/candidate_parameters/ajax/formHandler.php
@@ -422,12 +422,15 @@ function editConsentStatusFields(\Database $db)
 
         // Process posted data
         // Empty strings and type null are not passed (null is passed as a string)
-        $status     = ($_POST[$consentName] !== 'null') ?
-                        $_POST[$consentName] : null;
-        $date       = ($_POST[$consentName . '_date'] !== 'null') ?
-                        $_POST[$consentName . '_date'] : null;
-        $withdrawal = ($_POST[$consentName . '_withdrawal'] !== 'null') ?
-                        $_POST[$consentName . '_withdrawal'] : null;
+        $status     = (isset($_POST[$consentName]) &&
+                          $_POST[$consentName] !== 'null') ?
+                          $_POST[$consentName] : null;
+        $date       = (isset($_POST[$consentName . '_date']) &&
+                          $_POST[$consentName . '_date'] !== 'null') ?
+                          $_POST[$consentName . '_date'] : null;
+        $withdrawal = (isset($_POST[$consentName . '_withdrawal']) &&
+                          $_POST[$consentName . '_withdrawal'] !== 'null') ?
+                          $_POST[$consentName . '_withdrawal'] : null;
 
         $updateStatus  = [
             'CandidateID'   => $candID,
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -100,15 +100,16 @@ class Utility
     /**
      * Returns list of consents in the database
      *
-     * @return array An associative array of consents, with their names and labels.
-     *               The keys of the arrays are the IDs of the consents
+     * @return array An associative array of consents, with their names,
+     *               labels, and group ID. The keys of the arrays are the
+     *               IDs of the consents.
      */
     static function getConsentList(): array
     {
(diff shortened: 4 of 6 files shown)


modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -425,11 +425,11 @@ function editConsentStatusFields($db, $user)
 
         // Process posted data
         // Empty strings and type null are not passed (null is passed as a string)
-        $status     = ($_POST[$consentName] !== 'null') ?
+        $status     = (isset($_POST[$consentName]) && $_POST[$consentName] !== 'null') ?
```
johnsaigle: You can use null coalesce on these lines here to make them simpler, e.g.

```php
$data =  $_POST[$consentName . '_date'] ?? null;
```

zaliqarosli: nope because the second condition is checking to see if the variable is a string 'null', not null

johnsaigle: Thanks, I missed that.

zaliqarosli:
@PapillonMcGill thanks for the approval. could you also test the patch and raisinbread data?

php/libraries/Utility.class.inc
```diff
@@ -108,7 +108,8 @@ class Utility
         $factory = NDB_Factory::singleton();
         $DB      = $factory->database();
 
-        $query = "SELECT ConsentID, Name, Label FROM consent";
+        $query = "SELECT ConsentID, Name, Label FROM consent
+                  WHERE ParentID IS NOT NULL";
```
driusan: Why is the `ParentID IS NOT NULL` constraint added? My understanding from the description is that the parent consent is still a type of consent..

zaliqarosli: in PR #6044, i added a getParentConsents()... a parent consent isn't classified as a consent because no status/date/withdrawal data should be entered in `candidate_consent_rel` table for those consents. they are parents only, not actual consents with consent data

zaliqarosli: ideally `getConsentList()` would be renamed to `getConsentFields()` but then we would have to change every where the function is being used

driusan: Now I'm really confused about the usage of "Parent Consent" in this PR. Is it a type of consent or is it just a way to group consents?

ridz1208: @zaliqarosli 

@driusan might be right here, it seems like the parent consent is not really a consent and thus should not live in the same table. it seems like what is being added is a consent category but that cant live in the same table.

To be in the consent table it needs to have a similar yes/no option to all it's children. that could be done technically where if you answer yes/no to the parent it would apply to all children as well... further design might be needed to ensure that if you later add a consent field to a parent it should get autofilled if the parent is set to yes... details which could be figured out later

zaliqarosli: I based this method of parenting/grouping on how the parent/child relationship works for ConfigSettings ....

driusan: `ConfigSettings` is done that way because of legacy from when all the settings were in `config.xml` and ported to the database. Since it's was an XML document, both the parent and child *are* valid config settings (getSetting() on the parent being an array of child values.). If it was designed from scratch it wouldn't be done that way..

zaliqarosli: hmm okay so are we thinking a new parent_consent table?

driusan: no, a ConsentGroup table.

zaliqarosli: i don't think i understand the difference. what makes it a "real parent"?

driusan: It's really a parent if both the parent and child are types of consent and they form a tree. It's not really a parent if only the children are types of consent and the parent is just to group them.

zaliqarosli: righhttt. so for ConfigSettings, are the parent configs real parent configs or no?

zaliqarosli: in some ways, the parent consents are consents themselves i.e. they are the whole ICF, instead of individual consent fields

ridz1208: if you define a consent as having a date of "signature" a potential date of withdrawal and a yes/no status, then in the current code the parent is not a "real parent", the parent is simply not a consent it's just a group.

driusan:
What's the status of this one now that the release is out?

zaliqarosli:
the status is that this was under discussion. it looks to me like we could go on one of two directions:

1. Make the 'parent consent' a real parent consent that has its status updated by the module, not by user input
```
            // If children status all yes, let parent status = 'Yes'
            // If children status all no, let parent status = 'No'
            // Otherwise, including null, let parent status = 'Varied'
```

2. Create another table called `consent_group` that provides grouping for the existing consent

@driusan which one do you prefer?

driusan:
I prefer 2, but I just flagged it as requested changes so it doesn't show up in the approved list until the discussion is resolved..

zaliqarosli:
@driusan can you and @ridz1208 direct me to how I should be implementing this so that we can resolve the discussion? i will implement whatever you decide

ridz1208:
@zaliqarosli I went through the entire thread right now. I think everything you need is basically there.

Summary:
1. The consent case is different from the Configurations case because configurations has a legacy requirement for that design.
2. "parent consents" are not actual consents, they are just named groups of consents, so in the database you should have a consent groups table separate from the consents table and a foreign key from the consents table to the consent groups table to classify consents.

SQL/New_patches/2020-02-10-AddConsentGrouping.sql
```diff
@@ -0,0 +1,13 @@
...
+  `ConsentGroupID` integer unsigned NOT NULL AUTO_INCREMENT,
+  `Name` varchar(255) NOT NULL,
+  `Label` varchar(255) NOT NULL,
+  CONSTRAINT `PK_consent_group` PRIMARY KEY (`ConsentGroupID`),
+  CONSTRAINT `UK_consent_group_Name` UNIQUE KEY `Name` (`Name`),
+  CONSTRAINT `UK_consent_group_Label` UNIQUE KEY `Label` (`Label`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+INSERT INTO `consent_group` (`ConsentGroupID`, `Name`, `Label`) VALUES ('1', 'main_consent', 'Main consent');
+
+ALTER TABLE `consent` ADD `ConsentGroupID` integer unsigned NOT NULL;
```
ridz1208, line 11: @zaliqarosli have you tested this out? not sure what the behaviour would be since you are specifying `NOT NULL` but not giving a default. logically this would fail.

ridz1208: I just tested it out, mariadb automatically makes the default 0 since no default was given. since you are adding a main consent category, I would suggest making it explicitly the default to avoid possible bugs

raisinbread/RB_files/RB_consent.sql
```diff
@@ -1,8 +1,8 @@
 SET FOREIGN_KEY_CHECKS=0;
 TRUNCATE TABLE `consent`;
 LOCK TABLES `consent` WRITE;
-INSERT INTO `consent` (`ConsentID`, `Name`, `Label`) VALUES (1,'study_consent','Consent to Study');
-INSERT INTO `consent` (`ConsentID`, `Name`, `Label`) VALUES (2,'raisin_consent','Consent to Raisin');
-INSERT INTO `consent` (`ConsentID`, `Name`, `Label`) VALUES (3,'bread_consent','Consent to Bread');
+INSERT INTO `consent` (`ConsentID`, `Name`, `Label`, `ConsentGroupID`) VALUES (1,'study_consent','Consent to Study', '1');
+INSERT INTO `consent` (`ConsentID`, `Name`, `Label`, `ConsentGroupID`) VALUES (2,'raisin_consent','Consent to Raisin', '1');
```
ridz1208, line 5:
```suggestion
INSERT INTO `consent` (`ConsentID`, `Name`, `Label`, `ConsentGroupID`) VALUES (2,'raisin_consent','Consent to Raisin', '2');
```

raisinbread/RB_files/RB_consent.sql
```diff
@@ -1,8 +1,8 @@
...
+INSERT INTO `consent` (`ConsentID`, `Name`, `Label`, `ConsentGroupID`) VALUES (3,'bread_consent','Consent to Bread', '1');
```
ridz1208, line 6:
```suggestion
INSERT INTO `consent` (`ConsentID`, `Name`, `Label`, `ConsentGroupID`) VALUES (3,'bread_consent','Consent to Bread', '2');
```

SQL/New_patches/2020-02-10-AddConsentGrouping.sql
ridz1208, line 11:
```suggestion
ALTER TABLE `consent` ADD `ConsentGroupID` integer unsigned NOT NULL DEFAULT 1;
```

SQL/0000-00-00-schema.sql
```diff
@@ -2059,13 +2059,26 @@ CREATE TABLE `feedback_mri_comments` (
...
+  CONSTRAINT `UK_consent_group_Label` UNIQUE KEY `Label` (`Label`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+INSERT INTO `consent_group` (`Name`, `Label`) VALUES ('main_consent', 'Main consent');
+
 CREATE TABLE `consent` (
   `ConsentID` integer unsigned NOT NULL AUTO_INCREMENT,
   `Name` varchar(255) NOT NULL,
   `Label` varchar(255) NOT NULL,
+  `ConsentGroupID` integer unsigned NOT NULL,
```
ridz1208, line 2077:
```suggestion
  `ConsentGroupID` integer unsigned NOT NULL DEFAULT 1,
```

SQL/New_patches/2020-02-10-AddConsentGrouping.sql
```diff
@@ -0,0 +1,13 @@
...
+UPDATE `consent` SET `ConsentGroupID`='1';
```
ridz1208, line 12:
```suggestion
```

SQL/New_patches/2020-02-10-AddConsentGrouping.sql
```diff
@@ -0,0 +1,13 @@
...
+ALTER TABLE `consent` ADD CONSTRAINT `FK_consent_ConsentGroupID` FOREIGN KEY (`ConsentGroupID`) REFERENCES `consent_group` (`ConsentGroupID`) ON DELETE RESTRICT ON UPDATE CASCADE;
```
ridz1208, line 13: Why are you cascading in this patch but in the schema above you added restrict instead ?

I personally prefer cascades for updates (delete can stay on restrict)

zaliqarosli:
@ridz1208 ready for re-review

SQL/0000-00-00-schema.sql
```diff
@@ -2059,13 +2059,26 @@ CREATE TABLE `feedback_mri_comments` (
...
+  CONSTRAINT `UK_consent_group_Label` UNIQUE KEY `Label` (`Label`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+INSERT INTO `consent_group` (`ConsentGroupID`, `Name`, `Label`) VALUES ('1', 'main_consent', 'Main consent');
+
 CREATE TABLE `consent` (
   `ConsentID` integer unsigned NOT NULL AUTO_INCREMENT,
   `Name` varchar(255) NOT NULL,
   `Label` varchar(255) NOT NULL,
+  `ConsentGroupID` integer unsigned NOT NULL DEFAULt 1,
```
driusan, line 2077:
```suggestion
  `ConsentGroupID` integer unsigned NOT NULL DEFAULT 1,
```

zaliqarosli: good catch!

PR #6045, opened 2020-02-11
Ignore compiled JS for all modules

johnsaigle:
## Brief summary of changes

Adds a line to the main `.gitignore` to ignore the folder containing compiled JS for all modules.

This can replace the individual `.gitignore` files in each of the modules and save us some code review time.


maltheism:
@johnsaigle we might still have js files in the /js directory of some modules that need to be converted to react or moved to the jsx directory.

driusan:
Per module .gitignore was intentionally done in #4426. This will also break the modules that have non-jsx javascript in the js directory.

PR #6046, opened 2020-02-11
Remove PHPCS requirement for boilerplate comments

johnsaigle:
## Brief summary of changes

PHPCS no longer fails code if it does not have:

* A class comment
* A PHP version
* Link tag
* Package tag
* Category tag

PHPCS still requires a file comment, author, and licence. I chose to keep file comments over class comments so that there will still be some documentation required for PHP code without classes, such as tools/ and ajax/ scripts.

### DIFF
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -9,7 +9,6 @@ define("DEBUG", false);
  * Throughout the class, all input arrays follow the same structure:
  * $array['column_name'] = 'value'.
  *
- * @author  Alex Zijdenbos <zijdenbos@example.com>
  * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
  */
 class Database
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -1,4 +1,6 @@
 <?php declare(strict_types=1);
+define("DEBUG", false);
+
 /**
  * Database abstraction layer presents a unified interface to database
  * connectivity.  Instantiate one object instance per database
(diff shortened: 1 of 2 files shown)


driusan:
Thanks! I wasn't able to find a way to do that when I looked into it.

I think this is backwards though, every class should have a comment. It's important to have documentation of what a class is for.

johnsaigle:
No problem! I've been getting deep with PHPCS these days.

I think we should enforce that tools/ and ajax/ scripts have some documentation though even if they don't have classes.

Also we have one class per file for the most part.

I think this is the best compromise unless you can think of a better solution.

driusan:
You can request comments in `tools` directories without PHPCS enforcing it and the `ajax` directory shouldn't exist or have any PRs add new files to it. I think it looks messy to have a comment before the namespace, then a namespace, then a class with no comment. It makes sense for the namespace to be the first thing (other than strict_types) and the comment (which should be describing the class) to go with the class.

johnsaigle:
I modified it to require class comments instead of file comments.

php/libraries/Database.class.inc
```diff
@@ -20,11 +9,8 @@ define("DEBUG", false);
  * Throughout the class, all input arrays follow the same structure:
  * $array['column_name'] = 'value'.
  *
- * @category Main
- * @package  Loris
- * @author   Alex Zijdenbos <zijdenbos@example.com>
- * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
- * @link     https://www.github.com/aces/Loris-Trunk/
+ * @author  Alex Zijdenbos <zijdenbos@example.com>
```
driusan: Is it possible to also not require the `@author` tag? That's also unnecessary (it's in git) and often wrong (people just copy the boilerplate from other files and don't update it)

johnsaigle: Done

PR #6049, opened 2020-02-11
Fix one of the comment in the EEG browser module

cmadjar:
## Brief summary of changes

This fixes one of the comment in the code for the electrophysiology browser that was spotted by @driusan.

### DIFF
--- a/modules/electrophysiology_browser/php/electrophysiologybrowserrow.class.inc
+++ b/modules/electrophysiology_browser/php/electrophysiologybrowserrow.class.inc
@@ -16,8 +16,8 @@
 namespace LORIS\electrophysiology_browser;
 
 /**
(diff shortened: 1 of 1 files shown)


modules/electrophysiology_browser/php/electrophysiologybrowserrow.class.inc
```diff
@@ -16,7 +16,8 @@
 namespace LORIS\electrophysiology_browser;
 
 /**
- * A DICOMArchiveRow represents a row in the DICOM Archive menu table.
+ * A DICOMArchiveRow represents a row in the electrophysiology browser
```
driusan: This one's still wrong

cmadjar: fixed

PR #6050, opened 2020-02-11
[Survey Accounts] Data Frameworkify the module

driusan:
This updates the survey accounts module to use the data framework. Functionality should remain identical to before, except that "Site" and "Project" permissions should now be enforced by the menu filter and users should not see rows from other project/site's sessions.

### DIFF
--- a/modules/survey_accounts/php/surveyaccountsprovisioner.class.inc
+++ b/modules/survey_accounts/php/surveyaccountsprovisioner.class.inc
@@ -1,0 +1,75 @@
+<?php declare(strict_types=1);
+/**
+ * This file implements a data provisioner to get all possible rows
+ * for the survey accounts menu page.
+ *
+ * PHP Version 7
+ *
+ * @category   Behavioural
+ * @package    Main
+ * @subpackage Core
+ * @author     Dave MacFarlane <dave.macfarlane@mcin.ca>
+ * @license    http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link       https://www.github.com/aces/Loris/
+ */
+
+namespace LORIS\survey_accounts;
+
+/**
+ * This class implements a data provisioner to get all possible rows
+ * for the survey accounts menu page.
+ *
+ * PHP Version 7
+ *
+ * @category   Behavioural
+ * @package    Main
+ * @subpackage Core
+ * @author     Dave MacFarlane <dave.macfarlane@mcin.ca>
+ * @license    http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link       https://www.github.com/aces/Loris/
+ */
+class SurveyAccountsProvisioner
+    extends \LORIS\Data\Provisioners\DBRowProvisioner
+{
+    /**
+     * Create a SurveyAccountsProvisioner, which gets rows for the
+     * survey accounts menu table.
+     */
+    function __construct()
+    {
+
+        parent::__construct(
+            "SELECT
+                c.PSCID AS PSCID,
+                s.Visit_label AS Visit,
+                p.Email as Email,
+                p.Test_name as SurveyName,
+                p.OneTimePassword as URL,
+                p.Status,
+                s.CenterID,
+                s.ProjectID
+             FROM participant_accounts p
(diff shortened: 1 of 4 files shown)


driusan:
Blocked by either #6047 or #6011.

modules/survey_accounts/php/surveyaccountsprovisioner.class.inc
```diff
@@ -0,0 +1,75 @@
+<?php declare(strict_types=1);
+/**
```
johnsaigle: You can delete these file comments now if you want as long as we merge #6046 first.

johnsaigle:
1) Survey_AccountsTestIntegrationTest::testSurveyAccountsSearchButton
Failed asserting that '0 rows displayed of 0. (Maximum rows per page: 20501001000500010000)' contains "1 rows".
/app/modules/survey_accounts/test/survey_accountsTest.php:304
/app/modules/survey_accounts/test/survey_accountsTest.php:273

driusan:
@kongtiaowang is there a good way to get the UserID to insert into the user_psc_rel and user_project_rel tables so that this test will still work as before?

driusan:
@kongtiaowang ping

driusan:
@kongtiaowang Can you please answer the above question?

kongtiaowang:
@driusan There are some minor issues. "missing void" Return type is not compatible.
1.  URL links
" PHP Fatal error:  Declaration of NDB_BVL_Instrument_No_Additional_Options::_addMetadataFields() must be compatible with NDB_BVL_Instrument::_addMetadataFields(): void in /var/www/loris/project/libraries/NDB_BVL_Instrument_No_Additional_Options.class.inc on line 0, referer: https://wangshen-dev.loris.ca/survey_accounts/addSurvey/"
2. add Survey
 "PHP Fatal error:  Declaration of NDB_BVL_Instrument_No_Additional_Options::addMonthYear($field, $label, $options = Array) must be compatible with NDB_BVL_Instrument::addMonthYear(string $field, string $label, array $options = Array): void in /var/www/loris/project/libraries/NDB_BVL_Instrument_No_Additional_Options.class.inc on line 0, referer: https://wangshen-dev.loris.ca/survey_accounts"
3. 
PHP Fatal error:  Class NDB_BVL_Instrument_No_Additional_Options contains 2 abstract methods and must therefore be declared abstract or implement the remaining methods (NDB_BVL_Instrument::getFullName, NDB_BVL_Instrument::getSubtestList) in /var/www/loris/project/libraries/NDB_BVL_Instrument_No_Additional_Options.class.inc on line 0

driusan:
@kongtiaowang even if those are changed there will still be the issue of the rows returned being 0 because of the project filter being added. How do we fix that?

driusan:
@kongtiaowang and where are those errors coming from? I don't see any NDB_BVL_Instrument_No_Additional_Options in LORIS.

kongtiaowang:
@driusan I found the issues when I test it manually. It can't add a new survey, so the result of the filter test result should be one, not zero.

kongtiaowang:
@driusan the testing data can't insert into the survey table. I will figure it out.

kongtiaowang:
@driusan The Survey_Accounts uses useSiteFilter and useProjectFilter, so the projectId and centerId of testing data do not match useSiteFilter and useProjectFilter.
to fix:
Screen Shot 2020-02-28 at 3 02 57 PM

kongtiaowang:
@driusan So sorry. Please forget it can't add a new survey issue. That is because of my instrument files aren't updated.

driusan:
@lingma I don't understand what you mean. It's mock data, it needs to match the table. The column is called RegistrationCenter/ProjectID in the candidate table and just center/project in the session table.

lingma:
> @lingma I don't understand what you mean. It's mock data, it needs to match the table. The column is called RegistrationCenter/ProjectID in the candidate table and just center/project in the session table.
select RegistrationCenterID as CenterID could be used as CenterID, so it is not technical problem, just what the real need

driusan:
The center/project is added for the filtering. Since there is a session in the query, the actual center of the session is more accurate/specific than the place where the candidate originally registered (which shouldn't grant them access to all of the candidate's data just because it was first.)

PR #6061, opened 2020-02-13
[Data Table] Provide default table cell when getFormattedCell is not defined

driusan:
When a getFormattedCell callback is not defined, the DataTable component
currently silently throws away all the cells in the table. Since it's not
a required prop, this is confusing.

This provides a default cell format when no cell formatter is defined.

### DIFF
--- a/jsx/DataTable.js
+++ b/jsx/DataTable.js
@@ -435,6 +435,8 @@ class DataTable extends Component {
                     celldata,
                     row
                 );
+            } else {
+                cell = <td>{celldata}</td>;
             }
             if (cell !== null) {
                 // Note: Can't currently pass a key, need to update columnFormatter


jsx/DataTable.js
```diff
@@ -435,6 +435,8 @@ class DataTable extends Component {
                     celldata,
                     row
                 );
+            } else {
+                cell = <td>{celldata}</td>;
             }
             if (cell !== null) {
```
maltheism: Line 441 will never be null.

driusan: it can be null if getFormattedCell is defined and returns null

maltheism: @driusan oops my mistake.

Issue #6060, opened 2020-02-13
[tools] raininbread_refresh try to use unix user for BD access


PapillonMcGill:
**Describe the bug**
running raisinbread_refresh script try to use local unix user to connect to BD instead of the user in config.xml
```
...
Checking connection via project configuration file...
Array
(
    [0] => ERROR 1045 (28000): Access denied for user 'lorisadmin'@'localhost' (using password: NO)
    [1] => mysql: [Warning] Using a password on the command line interface can be insecure.
)

[*] Dropping LORIS tables.
...
```
the script was able to complete.

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. run `php raisinbread_refresh.php`

**What did you expect to happen?**
script run flalessly

**Server Environment (if known):**
- LORIS Version: [aces/master, latest commit at date]
- Linux distribution and Version: [e.g. Ubuntu 16.04]
- MySQL/MariaDB Version: [e.g. MySQL 5.7]

**Additional context**
adminUser and adminPassword properly set in config.xml
Possible that a commit was lost in a rebase of https://github.com/aces/Loris/pull/5260

PapillonMcGill:
Message from exception displayed on the screen.  Script work around this. 

Not a bug

PapillonMcGill:
Redirection problem.

johnsaigle:
Yep I think it should be fixed by the above PR. It's trying to connect via a DB configuration file and the error message should be hidden if it doesn't exist.

PR #6062, opened 2020-02-13
[tools] fix redirection order

PapillonMcGill:
## Brief summary of changes
Change the ordering of output redirection.

1. in tools/  `php raisinbread_refresh.php`
Ensure that no error message appear when the script test database connexion before deleting tables

* Resolves # 6060

### DIFF
--- a/tools/raisinbread_refresh.php
+++ b/tools/raisinbread_refresh.php
@@ -134,7 +134,6 @@ if ($status != 0) {
         escapeshellarg($password)
     );
     exec($mysqlCommand . ' -e "show tables;" 1>/dev/null 2>&1', $output, $status);
-    print_r($output);
     if ($status != 0) {
         die(
             printWarning(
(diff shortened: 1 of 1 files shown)


tools/raisinbread_refresh.php
```diff
@@ -133,7 +133,7 @@
         escapeshellarg($host),
         escapeshellarg($password)
     );
-    exec($mysqlCommand . ' -e "show tables;" 2>&1 1>/dev/null', $output, $status);
+    exec($mysqlCommand . ' -e "show tables;" 1>/dev/null 2>&1', $output, $status);
     print_r($output);
```
johnsaigle: Can you delete this line? It shouldn't be here. Or it should at least be checked only in the if statement below.

PapillonMcGill: good catch, this remove the annoying "array" output.

johnsaigle:
Does this solve your problem in #6060? All it does is hide the output from `exec` but it looks like you had a connection problem in your bug report.

PapillonMcGill:
Yes, you can try it. It properly redirect STDERR to /dev/null. 
The connection problem is that I don't have a `.mysql` file in my VM, thus the failure.

PR #6068, opened 2020-02-13
[JSX] Resolve all NPM warnings

HenriRabalais:
I'm getting pretty tired of all these NPM warnings for missing jsdocs or lines that exceed 80 characters so I started resolving the warnings. Some of these jsdocs seem superfluous, but hey, that's what it's asking for.

Should I keep going down this path for all jsx files? Y/N?


maltheism:
@HenriRabalais I agree it would be really nice to witness the day we can compile without those warnings visible.

driusan:
Y.

HenriRabalais:
@johnsaigle In general I agree - but for constructor, lifecycle and render functions there really isn't much to say. I could have maybe a single sentence that is copied throughout the code, but because these are inherited react functions, there's not much to document. I'll see if inheritdoc works for these!

We don't really have any parent classes in our jsx code in the way that php does. Components are sort of used like building blocks that are nested within one another, but we don't really have functions that are redeclared/overwritten by children, they're just passed down by the parent and props and accessed that way.

In terms of the errors, it seems like now that I'm writing the docs I've forgotten about declaring params and returns!

maltheism:
@johnsaigle @HenriRabalais maybe only add a description when there is something complicated going on in the constructor?

johnsaigle:
It might be good to look into solutions for linting that are aware of React structures, if that's the issue we're having. 

Something like this?
https://www.npmjs.com/package/eslint-plugin-react

driusan:
There's nothing React specific about the failures, it's just functions missing the `@param` and `@return` annotations.

HenriRabalais:
@johnsaigle I've updated the comments. 

Let me know if this works for everyone, and I will keep going through the code with the changes!

jsx/Breadcrumbs.js
```diff
@@ -18,32 +6,36 @@ import PropTypes from 'prop-types';
...
 
+  /** {@inheritdoc} */
   componentWillUnmount() {
     if (typeof window !== 'undefined') {
       window.removeEventListener('resize', this.checkScreenSize);
     }
   }
 
+  /**
+   * User to check the current window size and sets the number of breadcrumbs
```
johnsaigle:
```suggestion
   * Checks the current window size and sets the number of breadcrumbs
```

jsx/DataTable.js
```diff
@@ -170,6 +195,12 @@ class DataTable extends Component {
     return filteredIndexes;
   }
 
+  /**
+   * Sort of the row indexes.
```
johnsaigle:
```suggestion
   * Returns the supplied array sorted by indexes.
```

PR #6070, opened 2020-02-14
Fix bad permission ordering in Raisinbread data

johnsaigle:
## Brief summary of changes

Two new modules were recently added to LORIS. Raisinbread was updated but the permission codes were inverted which could cause issues with test data.

Compare with https://github.com/aces/Loris/blob/4d9c4c456d0c7f7a1dc4363c3c785298a9604cbc/SQL/0000-00-01-Permission.sql#L106-L109

Also updates the CHANGELOG with a sentence about the new Electrophysiology Browser module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,8 @@ changes in the following format: PR #1234***
  This allows projects to modify their instrument battery without requiring backend access.
  (https://github.com/aces/Loris/pull/4221)
  
-##### module1
+##### Electrophysiology Browser
+ - New module created to manage electrophysiology data within LORIS. (PR #5230)
 
 
 #### Clean Up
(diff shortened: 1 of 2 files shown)


ridz1208:
Looks good, can you also make sure the admin user has permissions to all of those in user perm rel ?

PapillonMcGill:
Is this PR really necessary? `permID` number should not be a problem. Code should always check the proper `permissions.code`.  Even the 0000-00-01-Permission.sql file should be change to let auto-increment do it's work.

johnsaigle:
It might not be necessary but on the other hand it didn't seem intentional. Won't this cause a bug?

driusan:
if them being in a different order causes any bugs then the bug should be fixed. The ID should be completely irrelevant.

PR #6076, opened 2020-02-17
[Core] Generically handle uncaught exceptions (no more blank crashes)

### LINKED ISSUE #4368: "Start Next Stage" will crash when instrument does not exist
Error-handling should be improved so that an error message is displayed to the user instead of crashing.

Example stack trace (on `major` branch):
>PHP Fatal error:  Uncaught Exception: Instrument does not exist in /var/www/loris/php/libraries/NDB_BVL_Instrument.class.inc:208\nStack trace:\n#0 /var/www/loris/php/libraries/NDB_BVL_Battery.class.inc(195): NDB_BVL_Instrument::factory('tsi', '', '')\n#1 /var/www/loris/php/libraries/NDB_BVL_Battery.class.inc(156): NDB_BVL_Battery->addInstrument('tsi')\n#2 /var/www/loris/modules/next_stage/php/next_stage.class.inc(128): NDB_BVL_Battery->createBattery(2, 'Visit', 'M1', 2, true)\n#3 /var/www/loris/php/libraries/NDB_Form.class.inc(162): LORIS\\next_stage\\Next_Stage->_process(Array)\n#4 /var/www/loris/php/libraries/LorisForm.class.inc(1436): NDB_Form->_save(Array)\n#5 /var/www/loris/php/libraries/NDB_Form.class.inc(133): LorisForm->process(Array)\n#6 /var/www/loris/php/libraries/NDB_Form.class.inc(190): NDB_Form->save()\n#7 /var/www/loris/src/Middleware/UserPageDecorationMiddleware.php(173): NDB_Form->handle(Object(Zend\\Diactoros\\ServerRequest))\n#8 /var/www/loris/src/Middleware/PageDecorationMiddleware.php(49): LORIS\\Middleware\\UserPageDecor in /var/www/loris/php/libraries/NDB_BVL_Instrument.class.inc on line 208, referer: https://jsaigle-dev.loris.ca/next_stage/?candID=575139&sessionID=2153&identifier=2153

johnsaigle:
## Brief summary of changes

Modifies the middleware to handle uncaught exceptions in a generic way. **Instead of LORIS crashing and displaying a blank page, it should now display a decorated error template.**

I've also updated the error templates to encourage users to reach out to the project administrator or to report an issue using the issue tracker module.

##### Example of how this fixes #6034 (Uncaught \NotFound exception)

Screen Shot 2020-02-17 at 16 23 20

---

##### Example of how this fixes #4368 (Uncaught Internal Server Error)

Screen Shot 2020-02-17 at 18 00 28

---

1. Test either of the linked issues and make sure that the errors no longer cause crashes.
2. Go to a module that doesn't exist and confirm that LORIS doesn't crash.

#### Related

Fixes #4368

### DIFF
--- a/htdocs/index.php
+++ b/htdocs/index.php
@@ -30,7 +30,7 @@ $client->initialize();
 // any authentication middleware, because that's done dynamically
 // based on the module router, depending on if the module is public.
 $middlewarechain = (new \LORIS\Middleware\ContentLength())
-    ->withMiddleware(new \LORIS\Middleware\ResponseGenerator());
+    ->withMiddleware(new \LORIS\Middleware\ExceptionHandlingMiddleware());
 
 $serverrequest = \Laminas\Diactoros\ServerRequestFactory::fromGlobals();
 
--- a/htdocs/index.php
+++ b/htdocs/index.php
@@ -29,8 +29,9 @@ $client->initialize();
 // Middleware that happens on every request. This doesn't include
 // any authentication middleware, because that's done dynamically
 // based on the module router, depending on if the module is public.
-$middlewarechain = (new \LORIS\Middleware\ContentLength())
-    ->withMiddleware(new \LORIS\Middleware\ExceptionHandlingMiddleware());
+$middlewarechain = (new \LORIS\Middleware\ExceptionHandlingMiddleware())
+    ->withMiddleware(new \LORIS\Middleware\ContentLength())
+    ->withMiddleware(new \LORIS\Middleware\ResponseGenerator());
 
 $serverrequest = \Laminas\Diactoros\ServerRequestFactory::fromGlobals();
 
--- a/src/Http/Error.php
+++ b/src/Http/Error.php
@@ -50,23 +50,26 @@ class Error extends HtmlResponse
         $uri     = $request->getURI();
         $baseurl = $uri->getScheme() .'://'. $uri->getAuthority();
         // Admistrator email.
-        $contact = $factory->config()->getSetting('mail')['From'];
-        // Issue tracker data.
-        $issueURL  = $factory->config()->getSetting('issue_tracker_url');
-        $canReport = $factory->user()->hasAnyPermission(
-            [
-             'issue_tracker_reporter',
-             'issue_tracker_developer',
-            ]
-        );
-
+        $contact  = $factory->config()->getSetting('mail')['From'];
         $tpl_data = array(
-                     'message'         => $message,
-                     'baseurl'         => $baseurl,
-                     'contact'         => $contact,
-                     'issueTrackerURL' => $issueURL . '/issue/new',
-                     'canReport'       => $canReport,
+                     'message' => $message,
+                     'baseurl' => $baseurl,
+                     'contact' => $contact,
                     );
+        // Add issue tracker data if the error is encountered by an authenticated
+        // user with the correct permissions.
+        $user = $factory->user();
+        if (! $user instanceof \LORIS\AnonymousUser) {
+            $issueURL  = $factory->config()->getSetting('issue_tracker_url');
+            $canReport = $factory->user()->hasAnyPermission(
+                [
+                 'issue_tracker_reporter',
+                 'issue_tracker_developer',
+                ]
+            );
+            $tpl_data['issueTrackerURL'] = $issueURL . '/issue/new';
+            $tpl_data['canReport']       = $canReport;
+        }
 
         $template_file = (string) $status . '.tpl';
 
--- a/src/Http/Error.php
+++ b/src/Http/Error.php
@@ -50,7 +50,10 @@ class Error extends HtmlResponse
         $uri     = $request->getURI();
         $baseurl = $uri->getScheme() .'://'. $uri->getAuthority();
         // Admistrator email.
-        $contact  = $factory->config()->getSetting('mail')['From'];
+        $contact  = $request
+            ->getAttribute('loris')
+            ->getConfiguration()
+            ->getSetting('mail')['From'];
         $tpl_data = array(
                      'message' => $message,
                      'baseurl' => $baseurl,
@@ -58,9 +61,9 @@ class Error extends HtmlResponse
                     );
         // Add issue tracker data if the error is encountered by an authenticated
         // user with the correct permissions.
-        $user = $factory->user();
+        $user = $request->getAttribute('user');
         if (! $user instanceof \LORIS\AnonymousUser) {
-            $canReport = $factory->user()->hasAnyPermission(
+            $canReport = $user->hasAnyPermission(
                 [
                  'issue_tracker_reporter',
                  'issue_tracker_developer',
--- a/src/Http/Error.php
+++ b/src/Http/Error.php
@@ -60,14 +60,13 @@ class Error extends HtmlResponse
         // user with the correct permissions.
         $user = $factory->user();
         if (! $user instanceof \LORIS\AnonymousUser) {
-            $issueURL  = $factory->config()->getSetting('issue_tracker_url');
             $canReport = $factory->user()->hasAnyPermission(
                 [
                  'issue_tracker_reporter',
                  'issue_tracker_developer',
                 ]
             );
-            $tpl_data['issueTrackerURL'] = $issueURL . '/issue/new';
+            $tpl_data['issueTrackerURL'] = '/issue_tracker/issue/new';
             $tpl_data['canReport']       = $canReport;
         }
 
--- a/src/Http/Error.php
+++ b/src/Http/Error.php
@@ -49,30 +49,30 @@ class Error extends HtmlResponse
 
         $tpl_data = [];
 
-        $lorisInstance  = $request->getAttribute('loris');
-        $user  = $request->getAttribute('user') ?? new \LORIS\AnonymousUser();
+        $lorisInstance = $request->getAttribute('loris');
+        $user          = $request->getAttribute('user') ?? new \LORIS\AnonymousUser();
         
 
         // Add a link to the issue tracker as long as a LORIS Instance object
         // is present in the request.
-        if (! $user instanceof \LORIS\AnonymousUser 
+        if (! $user instanceof \LORIS\AnonymousUser
             && $lorisInstance !== null
         ) {
             // Add admistrator email.
-            $contact = $lorisInstance
+            $contact  = $lorisInstance
                 ->getConfiguration()
                 ->getSetting('mail')['From'];
             $tpl_data = array(
-                'message' => $message,
-                'baseurl' => $baseurl,
-                'contact' => $contact,
-            );
+                         'message' => $message,
+                         'baseurl' => $baseurl,
+                         'contact' => $contact,
+                        );
             // Add issue tracker data if the error is encountered by a user with
             // the correct permissions.
             $canReport = $user->hasAnyPermission(
                 [
-                    'issue_tracker_reporter',
-                    'issue_tracker_developer',
+                 'issue_tracker_reporter',
+                 'issue_tracker_developer',
                 ]
             );
             if ($canReport) {
--- a/src/Http/Error.php
+++ b/src/Http/Error.php
@@ -45,12 +45,27 @@ class Error extends HtmlResponse
         string $message = ''
     ) {
 
+        $factory = \NDB_Factory::singleton();
+
         $uri     = $request->getURI();
         $baseurl = $uri->getScheme() .'://'. $uri->getAuthority();
+        // Admistrator email.
+        $contact = $factory->config()->getSetting('mail')['From'];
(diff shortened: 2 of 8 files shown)


src/Http/Error.php
```diff
@@ -45,13 +45,30 @@ public function __construct(
...
+        // Admistrator email.
+        $contact  = $factory->config()->getSetting('mail')['From'];
         $tpl_data = array(
                      'message' => $message,
                      'baseurl' => $baseurl,
+                     'contact' => $contact,
                     );
+        // Add issue tracker data if the error is encountered by an authenticated
+        // user with the correct permissions.
+        $user = $factory->user();
```
xlecours:
```suggestion
        $user = $request->getAttribute('user');
```

johnsaigle: good catch, thanks!

johnsaigle:
@driusan Could you provide a bit more detail? I'd like to implement this but I don't really know my way around the architecture well enough to know where else this code should go.

driusan:
Middleware are things which implement a `process` function to modify a request before passing it to a handler or modifying a response that comes back from a handler. See PSR-15 for the interface.

``
/**
 * A ResponseGenerator is a type of MiddlewareChainer which terminates the
 * chain and forces a response to be returned unconditionally.
 *
```

You are attempting to do make it to something completely unrelated to the `ResponseGenerator`'s job. You're trying to modify the response by catching exceptions and displaying an error page. You should do that by creating a new middleware (like `ExceptionCatchingMiddleware` or something) in `src/Middleware` that is only concerned with catching exceptions.

**Comment by johnsaigle** [comment]
Thanks for explaining. I've added a new class containing the code I previously added to ResponseGenerator.

This still needs work because I need to figure out exactly where in the chain of middleware that this new one should go. I'm thinking the BaseRouter right now since I think we want this middleware to be as upstream as possible so we have a better chance of catching stray exceptions.

**Comment by johnsaigle** [comment]
@driusan This new approach is working for me locally so it's ready for your review.

File: src/Middleware/ExceptionHandlingMiddleware.php [hunk 2]
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Implements the MiddlewareChainer interface
+     *
+     * @param MiddlewareChainer $next The next middleware to chain. (Discarded)
+     *
+     * @return MiddlewareChainer The same MiddlewareChainer, unmodified.
+     */
+    public function withMiddleware(MiddlewareChainer $next)
```
driusan: This shouldn't be discarding next. It should be chaining with it. You can use the `MiddlewareChainerMixin` (see ContentLengthMiddleware)

src/Middleware/ExceptionHandlingMiddleware.php
```diff
@@ -0,0 +1,83 @@
...
+     *                                         valid requests to
+     *
+     * @return ResponseInterface the PSR15 response.
+     */
+    public function process(
+        ServerRequestInterface $request,
+        RequestHandlerInterface $handler
+    ) : ResponseInterface {
+        try {
+            return (new \LORIS\Middleware\ResponseGenerator())
```
driusan: This shouldn't be calling the ResponseGenerator, it should be chaining to the next middleware in the chain. See ContengLengthMiddleware for an example.

src/Middleware/ExceptionHandlingMiddleware.php
```diff
@@ -0,0 +1,83 @@
...
+                ->process(
+                    $request,
+                    $handler
+                );
+        } catch (\Exception $e) {
+            // Handle uncaught errors.
+            switch (get_class($e)) {
```
driusan: Why is this not 2 catch statements to catch different types of exceptions?

htdocs/index.php
```diff
@@ -30,7 +30,7 @@
 // any authentication middleware, because that's done dynamically
 // based on the module router, depending on if the module is public.
 $middlewarechain = (new \LORIS\Middleware\ContentLength())
-    ->withMiddleware(new \LORIS\Middleware\ResponseGenerator());
+    ->withMiddleware(new \LORIS\Middleware\ExceptionHandlingMiddleware());
```
driusan: The ResponseGenerator shouldn't be removed from the chain

src/Middleware/ExceptionHandlingMiddleware.php
johnsaigle: This seemed more concise

johnsaigle:
@driusan Thanks for the feedback. how's this?

htdocs/index.php
```diff
@@ -29,7 +29,8 @@
 // Middleware that happens on every request. This doesn't include
 // any authentication middleware, because that's done dynamically
 // based on the module router, depending on if the module is public.
-$middlewarechain = (new \LORIS\Middleware\ContentLength())
+$middlewarechain = (new \LORIS\Middleware\ExceptionHandlingMiddleware())
+    ->withMiddleware(new \LORIS\Middleware\ContentLength())
```
driusan: Are you sure these are the right order? I can't remember if the chain is a stack or a queue, but can you make sure this calculates the content-length of the error page, and not of the pre-modified data?

johnsaigle: > 1.4 Handling Exceptions
It is RECOMMENDED that any application using middleware includes a component that catches exceptions and converts them into responses. **This middleware SHOULD be the first component executed and wrap all further processing to ensure that a response is always generated.**

https://www.php-fig.org/psr/psr-15/

If that's any consolation

johnsaigle: Inverting the order doesn't seem to change the Content-Length value for the 500 response error page (see screenshot in OP). I made sure to refresh with Cmd + Shift + R to clear the cache

driusan: I don't see the network dev console/http response headers in the screenshot, how can I check the Content-Length value?

johnsaigle: I just meant that was the page I was testing, not that the content-length is there.

They match in manual testing for me. Maybe you can confirm locally to verify?

johnsaigle:
@ridz1208 @driusan I'd recommend that this PR goes into the release as it's a pretty simple change that adds huge utility by 1) preventing crashes in LORIS and 2) providing debug messages by handling uncaught exceptions. I expect that it should reduce our maintenance overhead when users encounter bugs.

driusan:
This emphatically does *not* prevent crashes in LORIS. It makes error pages prettier when things crash at the trade off of hiding the debugging information of uncaught exceptions. I don't think it merits an exception to the release freeze.

johnsaigle:
OK. Should I amend the pull request to also log the caught exceptions? That was it's only an improvement, not a trade-off.

src/Middleware/ExceptionHandlingMiddleware.php
```diff
@@ -0,0 +1,58 @@
...
+            error_log($e->getMessage() . $e->getTraceAsString());
+            $status = 404;
+        } catch (\Exception $e) {
+            error_log($e->getMessage() . $e->getTraceAsString());
+            $status = 500;
+        }
+
+        // Decorate the request.
+        return (new \LORIS\Middleware\PageDecorationMiddleware(
+            \NDB_Factory::singleton()->user()
```
driusan, line 47: This should be coming from the request object 'user' attribute which is populated by the BaseRouter.

src/Http/Error.php
```diff
@@ -45,13 +45,30 @@ public function __construct(
         string $message = ''
     ) {
 
+        $factory = \NDB_Factory::singleton();
```
driusan, line 48: No other response type is using globals, you can just pass the parameters you want in as function arguments.

johnsaigle:
@driusan I reworked this to use request attributes instead of globals. Can you take another look?

src/Http/Error.php
```diff
@@ -45,13 +45,33 @@ public function __construct(
...
```
driusan, line 48: This isn't being used now

src/Http/Error.php
```diff
@@ -45,13 +45,33 @@ public function __construct(
...
+
         $uri     = $request->getURI();
         $baseurl = $uri->getScheme() .'://'. $uri->getAuthority();
-
+        // Admistrator email.
+        $contact  = $request
+            ->getAttribute('loris')
+            ->getConfiguration()
+            ->getSetting('mail')['From'];
```
driusan, line 56: why don't you just pass in the administrator email and the user object directly?

src/Http/Error.php
```diff
@@ -45,13 +45,33 @@ public function __construct(
...
         $tpl_data = array(
                      'message' => $message,
                      'baseurl' => $baseurl,
+                     'contact' => $contact,
                     );
+        // Add issue tracker data if the error is encountered by an authenticated
+        // user with the correct permissions.
+        $user = $request->getAttribute('user');
+        if (! $user instanceof \LORIS\AnonymousUser) {
```
driusan, line 65: This if logic here is wrong. canReport should be based on hasAnyPermission regardless of the type of user (AnonymousUsers won't have the permission, and if LORIS is intentionally set up so that they do it shouldn't be bypassed.), and issueTrackerURL should be based on whether canReport is true.

src/Http/Error.php
johnsaigle: Wouldn't that require updating all of the calling code?

src/Http/Error.php
johnsaigle: OK I can remove the if statement.

I'm not really sure what you mean by "issueTrackerURL should be based on whether canReport is true."

src/Http/Error.php
johnsaigle, line 48:
```suggestion
```

src/Http/Error.php
driusan: I didn't realize $request was already being passed, I thought you added it for this and were changing it anyways.. nevermind, since it's already there you can use it directly

src/Http/Error.php
driusan: The other template variable conditionally being set in the if statement

```
  $tpl_data['issueTrackerURL'] = '/issue_tracker/issue/new';
```

The condition for that being set should be `if ($canReport === true)`, not `if (! $user instanceof \LORIS\AnonymousUser)`.. though it also doesn't matter if you just unconditionally set it, since the template won't use it if it doesn't need it.

src/Http/Error.php
```diff
@@ -44,14 +44,42 @@ public function __construct(
...
-                    );
+        $tpl_data = [];
+
+        $lorisInstance = $request->getAttribute('loris');
+        $user          = $request->getAttribute('user') ?? new \LORIS\AnonymousUser();
+        
+
+        // Add a link to the issue tracker as long as a LORIS Instance object
+        // is present in the request.
+        if (! $user instanceof \LORIS\AnonymousUser
```
driusan, line 58: Why is this check re-added?

johnsaigle: I removed it but it caused a bunch of errors, e.g. relating to calling `hasPermission()` on `null` farther down the stack. Removing it seems to break everything

src/Middleware/ExceptionHandlingMiddleware.php
driusan: This is still not done.

johnsaigle: I think this had similar issues to the Anonymous User call above. When I changed it there were cases where the 'user' attribute was null.

johnsaigle: @driusan Can you follow up on this point and the above point?

johnsaigle: @driusan bump

driusan: There should never be cases where it's null, it's added unconditionally from the BaseRouter

johnsaigle: I just tested again. It works as long as the Anonymous User check is there. Otherwise this error occurs when a user is not logged in:

Uncaught TypeError: Argument 1 passed to LORIS\\Middleware\\PageDecorationMiddleware::__construct() must be an instance of User, null given, called in /var/www/loris/src/Middleware/ExceptionHandlingMiddleware.php on line 47 and defined in /var/www/loris/src/Middleware/PageDecorationMiddleware.php:18

src/Http/Error.php
johnsaigle: Without this check, this error occurs for unathenticated users:

Uncaught TypeError: Argument 1 passed to LORIS\\Middleware\\PageDecorationMiddleware::__construct() must be an instance of User, null given, called in /var/www/loris/src/Middleware/ExceptionHandlingMiddleware.php on line 47 and defined in /var/www/loris/src/Middleware/PageDecorationMiddleware.php:18

PR #6078, opened 2020-02-18
[User Accounts] Separate "My Preferences" into its own module

### LINKED ISSUE #4440: "My Preferences" help content does not reflect its functionality
Since `my_preferences` and `edit_user` are really both the `user_accounts` module, they share help content documentation. The help content describes only the functionality of the edit_user page.

### LINKED ISSUE #6040: A user should always have access to My Preferences
**Describe the bug**
A "user management" permission is needed to edit one's preferences on LORIS since it is the same module as edit_user.

This is silly because it means a user can't update their own password, username, etc. easily.

**To Reproduce**
1. Create a user with no permissions
2. On the top right, above logout, click My Preferences
3. You will be forbidden

You should be able to access the page whenever.

- LORIS Version: `aces/master`

johnsaigle:
## Brief summary of changes

Creates a new module "My Preferences" containing code that was previously bundled with User Accounts functionality.

1. On this branch, run the patch and make sure My Preferences still works.

* Resolves #6040 
* Resolves #4440


modules/my_preferences/php/module.class.inc
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * {@inheritDoc}
+     *
+     * @param \User $user The user whose access is being checked.
+     *
+     * @return bool whether access is granted
+     */
+    public function hasAccess(\User $user) : bool
+    {
+        return parent::hasAccess($user);
```
PapillonMcGill: Shouldn't this always be `true`

driusan: No it shouldn't.. anonymous users don't have access to my preferences.

But there's no reason to override the method just to return the parent value either.

modules/my_preferences/README.md
```diff
@@ -0,0 +1,29 @@
+# My Preferences
+
+## Purpose
+
+This module allows a user to edit their basic profile information as well as
+customize their notification settings.
+
+## Intended Users
+
+Every user is able to modify their profile information.
```
PapillonMcGill:
```suggestion
Every authenticated user is able to modify their profile information.
```

johnsaigle:
@ridz1208 @driusan This should probably be tagged to go into the release as it fixes Critical Issue #6040

driusan:
I agree, I changed the branch on the GitHub frontend.

modules/my_preferences/README.md
```diff
@@ -0,0 +1,29 @@
...
+This module allows a user to edit their basic profile information as well as
+customize their notification settings.
+
+## Permissions
+
+A user is always able to access the My Preferences page.
+
+## Configurations
+
+Some settings related to email notifications affect this module.
```
driusan: What settings related to email notification affect this module?

modules/my_preferences/README.md
driusan: I agree with @PapillonMcGill

modules/my_preferences/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Interactions with LORIS
+
+- Modifications performed on the My Preferences page will be reflected on 
+the user account page when viewing the edited account.
```
driusan: I don't think that's really an interaction worth noting. Any data modified anywhere should be reflected other places that use the data..

driusan:
@kongtiaowang can you review all the test related changes in this PR to make sure they make sense?

modules/my_preferences/README.md
johnsaigle: Notification settings for different modules with upload/download. I'll modify this

modules/my_preferences/README.md
```diff
@@ -20,9 +20,5 @@ A user is always able to access the My Preferences page.
 
 ## Configurations
 
-Some settings related to email notifications affect this module.
-
-## Interactions with LORIS
-
-- Modifications performed on the My Preferences page will be reflected on 
-the user account page when viewing the edited account.
+This module allows a user to modify whether or not they are notified by events in other modules,
```
ridz1208: @johnsaigle I thought this section is for configurations that affect this module, not configurations that are affected by it.

not sure though

driusan: Yes.

johnsaigle: Thanks for clarifying. As far as I know there aren't any.

PR #6080, opened 2020-02-18
Send email to admin upon account request

### LINKED ISSUE #5943: Send email to administrator when a user requests an account
Login module, request account page.

A feature was requested in today's LORIS meeting to send an email to a coordinator when a user requests an account. Currently, when a user requests an account a coordinator receives a dashboard notification. However, they won't see this until they login to LORIS.

**Additional context**
Google reCAPTCHA is currently not working for LORIS. There is a risk of spam abuse without this protection. It would be ideal to fix this issue prior to or concurrently with this feature.

johnsaigle:
## Brief summary of changes

When a user requests an account, an email will be sent to the study admin.

---

Screen Shot 2020-02-18 at 16 50 38

---

1. Follow the new steps added to the test plan in this PR.

* Resolves #5943

### DIFF
--- a/modules/login/php/requestaccount.class.inc
+++ b/modules/login/php/requestaccount.class.inc
@@ -269,6 +269,35 @@ class RequestAccount extends \NDB_Form
             }
         }
 
+        // Verify reCAPTCHA
+        $err = array();
+        if (isset($values['g-recaptcha-response']) && isset($reCAPTCHAPrivate)) {
+            $recaptcha = new \ReCaptcha\ReCaptcha($reCAPTCHAPrivate);
+            $resp      = $recaptcha->verify(
+                $values['g-recaptcha-response'],
+                $_SERVER['REMOTE_ADDR']
+            );
+            if (!$resp->isSuccess()) {
+                $err['captcha'] = 'Please complete the reCaptcha!';
+            }
+        }
+        // Send an email to the administrator informing them that a user has
+        // requested an account.
+        if (empty($err) && empty($errors)) {
+            $config    = \NDB_Factory::singleton()->config();
+            $studyName = $config->getSetting('title');
+
+            \Email::send(
+                $config->getSetting('mail')['From'],
+                'request_account.tpl',
+                array(
+                    'Study'     => $studyName,
+                    'UserID'    => $from,
+                    'Real_name' => $fullname,
+                    'Email'     => $from,
+                ),
+            );
+        }
         // Show success message even if email already exists for security reasons
         $this->tpl_data['success'] = true;
     }
--- a/modules/login/php/requestaccount.class.inc
+++ b/modules/login/php/requestaccount.class.inc
@@ -281,7 +281,7 @@ class RequestAccount extends \NDB_Form
         //     - Associated with the Site
         //     - Associated with the Project
         //     - Have the User Management permission
-        $query = <<<SQL
+        $query  = <<<SQL
         SELECT DISTINCT u.Email,u.userID
         FROM users u
             JOIN user_psc_rel upr
--- a/modules/login/php/requestaccount.class.inc
+++ b/modules/login/php/requestaccount.class.inc
@@ -315,7 +315,7 @@ class RequestAccount extends \NDB_Form
                     'UserID'    => $from,
                     'Real_name' => $fullname,
                     'Email'     => $from,
-                ],
+                ]
             );
         }
     }
--- a/modules/login/php/requestaccount.class.inc
+++ b/modules/login/php/requestaccount.class.inc
@@ -245,7 +245,7 @@ class RequestAccount extends \NDB_Form
                     ["fn" => $fullname]
                 );
 
-                // If examiner not in table add him
+                // Add new account to the examiner tables
                 if (empty($examinerID)) {
                     $DB->insert(
                         'examiners',
@@ -271,6 +271,11 @@ class RequestAccount extends \NDB_Form
 
         // Show success message even if email already exists for security reasons
         $this->tpl_data['success'] = true;
+
+        // Don't send out emails if this is a sandbox VM.
+        if ($factory->config()->settingEnabled('sandbox')) {
+            return;
+        }
 
         // Send an email to all users meeting the following criteria:
         //     - Associated with the Site
--- a/modules/login/php/requestaccount.class.inc
+++ b/modules/login/php/requestaccount.class.inc
@@ -293,16 +293,14 @@ class RequestAccount extends \NDB_Form
         //     - Associated with the Site
         //     - Associated with the Project
         //     - Have the User Management permission
-        $query  = <<<SQL
-        SELECT DISTINCT u.Email,u.userID 
-        FROM users u 
-        JOIN user_psc_rel upr 
-        ON u.ID = upr.UserID 
-        JOIN user_project_rel uprojr 
-        ON u.ID = uprojr.UserID 
-        WHERE upr.CenterID=:center
-        AND uprojr.ProjectID=:project
-SQL;
+        $query  = "SELECT DISTINCT u.Email,u.userID
+             FROM users u
+                JOIN user_psc_rel upr
+                    (ON u.ID = upr.UserID)
+               JOIN user_project_rel uprojr
+                   (ON u.ID = uprojr.UserID)
+            WHERE upr.CenterID=:center
+                AND uprojr.ProjectID=:project"
         $result = $DB->pselect(
             $query,
             ['center' => $site, 'project' => $project]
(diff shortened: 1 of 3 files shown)


modules/login/php/requestaccount.class.inc
```diff
@@ -223,6 +223,23 @@ class RequestAccount extends \NDB_Form
                 $err['captcha'] = 'Please complete the reCaptcha!';
             }
         }
+        // Send an email to the administrator informing them that a user has
+        // requested an account.
+        if (empty($err)) {
+            $config    = \NDB_Factory::singleton()->config();
+            $studyName = $config->getSetting('title');
+
+            \Email::send(
```
driusan: Before doing this you should ensure that the config setting is a valid email. I think it's also risky to do this by default, including sandboxes and places that haven't set up their email.. especially since that's not the purpose of the 'From' config setting and the default is 'no-reply@example.com'.

I think we should add a new config setting which defaults to empty, and only send an email if it's explicitly enabled.

johnsaigle: Would this be two config settings then? e.g. `administratorEmail` and `emailAdminOnAccountRequest`?

driusan: wouldn't something like `accountRequestContactEmail` be enough? If it's not set, there's no one to try emailing..

johnsaigle: In #5965 I used the `From` setting in another context to test whether the server can send emails. Also in #6076 I did the same thing.

It's probably useful to have a separate email where an administrator can be reached if `From` is not suitable. I think 2 is better.

johnsaigle: Blocked by #6101

ridz1208:
@johnsaigle I don't like the idea of having an email sent to one particular "admin" thats not how most LORIS instances work. User management is generally done per-sire or per-project.

Instead I would send a email to ALL users which have the site  AND the project the request was made to AND also have the user management permission in their permissions. On a properly managed LORIS project, that would probably only be the coordinator and their assistant maybe.

johnsaigle:
Thanks @ridz1208 that's a good idea. I'll add that.

ridz1208:
@johnsaigle no longer blocked

johnsaigle:
@ridz1208 Just implemented it. Could you review?

modules/login/php/requestaccount.class.inc
```diff
@@ -269,7 +269,67 @@ class RequestAccount extends \NDB_Form
             }
         }
 
+        // Verify reCAPTCHA
+        $err = [];
+        if (isset($values['g-recaptcha-response']) && isset($reCAPTCHAPrivate)) {
```
ridz1208, line 274: @johnsaigle why are the recptcha changes here ? bad rebase?

modules/login/php/requestaccount.class.inc
```diff
@@ -269,7 +269,67 @@ class RequestAccount extends \NDB_Form
...
+                );
+            if ($hasPermission === false) {
+                continue;
+            }
+
+            // Send email to users with User Management permissions.
+            $config    = \NDB_Factory::singleton()->config();
+            $studyName = $config->getSetting('title');
+
+            \Email::send(
```
ridz1208, line 323: I wont block this PR over this but I think we should use the notifier class in the future for this

modules/login/php/requestaccount.class.inc
```diff
@@ -269,7 +269,67 @@ class RequestAccount extends \NDB_Form
...
+SQL;
+        $result = $DB->pselect(
+            $query,
+            ['center' => $site, 'project' => $project]
+        );
+        foreach ($result as $row) {
+            $hasPermission = \User::factory($row['userID'])
+                ->hasAnyPermission(
+                    ['user_accounts', 'user_accounts_multisite']
+                );
```
ridz1208, line 314: I find this is a weird combination of permission checking. you are doing SQL checks for site and project and User object checks for the module permissions. could you justify ? why not all SQL or all user class?

modules/login/php/requestaccount.class.inc
driusan: How much work is involved in changing it to use the Notifier class? That seems like a reasonable change..

ridz1208: not that much but it doesn't bring many benefits.. other I think than being able to subscribe or unsubscribe from email notifications

ridz1208: https://github.com/aces/Loris/wiki/Notification-system

modules/login/php/requestaccount.class.inc
```diff
@@ -269,7 +269,67 @@ class RequestAccount extends \NDB_Form
...
+        $query  = <<<SQL
+        SELECT DISTINCT u.Email,u.userID 
+        FROM users u 
+        JOIN user_psc_rel upr 
+        ON u.ID = upr.UserID 
+        JOIN user_project_rel uprojr 
+        ON u.ID = uprojr.UserID 
+        WHERE upr.CenterID=:center
+        AND uprojr.ProjectID=:project
+SQL;
```
driusan, line 305:
```suggestion
        $query  = "SELECT DISTINCT u.Email,u.userID
             FROM users u
                JOIN user_psc_rel upr
                    (ON u.ID = upr.UserID)
               JOIN user_project_rel uprojr
                   (ON u.ID = uprojr.UserID)
            WHERE upr.CenterID=:center
                AND uprojr.ProjectID=:project"
```

Formatting SQL isn't a good use of HEREDOCs and also the formatting is hard to read.

modules/login/php/requestaccount.class.inc
driusan: #6101 is merged

modules/login/php/requestaccount.class.inc
johnsaigle: That could be future work. You could make a ticket to refactor this to use the notification system

modules/login/php/requestaccount.class.inc
johnsaigle: I would do all SQL but I'm not skilled enough with it to make a big query that does all of it. I'm also working everything out before my last day so I'm using the quickest approach to getting all requested changes made + everything rebased. Feel free to refactor it or create a ticket to clean this up but it's not something I'm going to be able to prioritize

modules/login/php/requestaccount.class.inc
johnsaigle: Yep. I'll fix it

ridz1208:
@driusan 

@johnsaigle 's explanations make sense. your call

modules/login/php/requestaccount.class.inc
```diff
@@ -271,5 +271,46 @@ class RequestAccount extends \NDB_Form
...
+                );
+            if ($hasPermission === false) {
+                continue;
+            }
+
+            // Send email to users with User Management permissions.
+            $config    = \NDB_Factory::singleton()->config();
+            $studyName = $config->getSetting('title');
+
+            \Email::send(
```
driusan, line 304: This is still missing a sandbox check

johnsaigle:
@driusan I added the sandbox check.

I also reverted your MySQL changes -- the caused a syntax error on the back-end. I incorporated your indentation though to help with readability.

johnsaigle:
@driusan could you merge if satisfied?

modules/login/php/requestaccount.class.inc
```diff
@@ -271,5 +271,53 @@ class RequestAccount extends \NDB_Form
...
+        // Don't send out emails if this is a sandbox VM.
+        if ($factory->config()->settingEnabled('sandbox')) {
+            return;
+        }
+
+        // Send an email to all users meeting the following criteria:
+        //     - Associated with the Site
+        //     - Associated with the Project
+        //     - Have the User Management permission
+        $query  = <<<SQL
```
driusan, line 284: This still needs to be changed from a HEREDOC to normal PHP quotes by whoever picks this up, then it'll be ready to be merged.

modules/login/php/requestaccount.class.inc
cmadjar: unless I am misreading this, I believe the sandbox check is done line 276 (returns from the function and this code is not executed).

modules/login/php/requestaccount.class.inc
cmadjar: @ridz1208 is there an issue already created for that? If not, would you mind creating it since you know what you are talking about and I don't? Thank you! :)

modules/login/php/requestaccount.class.inc
cmadjar: nevermind, I guess this was resolved before John left. Will resolve this comment.

cmadjar:
@driusan ready for review and merge :)

PR #6082, opened 2020-02-19
[Issue Tracker] Add upgrade script for module column

### LINKED ISSUE #5959: [Issue Tracker] Write upgrade script for module column
The issue tracker has a column called "module". This was an integer which was referencing the LorisMenu table (without a foreign key). The table was dropped and the column became a foreign key to the modules table, which makes more sense for a column called "module"..

We need to:
1. Declare the new foreign key properly, now that the column is referencing something that makes sense.
2. Write a script which fixes the data before the release patch drops the LorisMenu table. This is tricky, since the issue tracker was previously calling something that has no relation to modules a "module" and we need a way to map the data. Options I can think of are:
 a. Null the data so that the "module" column for existing issues don't change randomly 
 b. Try and map the module getLongName() function to the previous LorisMenu label and warn/null the ones that can't be mapped.
 c. Try and parse the URL from the menu item to get the module name and null the ones that can't be mapped.

This needs to be done before the next release.

johnsaigle:
## Brief summary of changes

Issue Tracker has a `module` table which references IDs in the deleted LorisMenu table.

This PR adds a script to map the old LorisMenu IDs in this column to the corresponding IDs in the new `modules` table.

1. Your database must have both the `LorisMenu` table and the `modules` table.
- If you've already installed `modules` and dropped `LorisMenu`, download and run [this SQL file](https://raw.githubusercontent.com/aces/Loris/2f07cc6d75a7f78201748ec917e963062972fa8a/SQL/0000-00-02-Menus.sql)
- If you still have `LorisMenu` and do not have `modules`, download and run [this SQL file](https://raw.githubusercontent.com/driusan/Loris/ec0981150a29bed86eece3b3a3452d9f166cf72a/SQL/0000-00-02-Modules.sql).

2. Run my script. It should display a before and after comparison of the values as well as the SQL statements you need to run, e.g. on Raisinbread:

```SQL
Issue ID: 16. Module before (`LorisMenu`): `dataquery/`. Module after (`modules`): `dataquery`
Issue ID: 19. Module before (`LorisMenu`): `examiner/`. Module after (`modules`): `examiner`
Issue ID: 20. Module before (`LorisMenu`): `media/`. Module after (`modules`): `media`
Issue ID: 26. Module before (`LorisMenu`): `examiner/`. Module after (`modules`): `examiner`
Issue ID: 27. Module before (`LorisMenu`): `conflict_resolver/`. Module after (`modules`): `conflict_resolver`
UPDATE `issues` SET module=14 WHERE issueID=16;
UPDATE `issues` SET module=17 WHERE issueID=19;
UPDATE `issues` SET module=29 WHERE issueID=20;
UPDATE `issues` SET module=17 WHERE issueID=26;
UPDATE `issues` SET module=9 WHERE issueID=27;
ALTER TABLE `issues` ADD FOREIGN KEY (moduleID) REFERENCES (modules(ID));
```

* Resolves #5959


tools/single_use/22.0_To_23.0.php
```diff
@@ -0,0 +1,133 @@
+<?php declare(strict_types=1);
+/**
+ * This script is used to correctly establish a link betwen the `issues` table
+ * used by the Issue Tracker module and the new table `modules`.
+ * It outputs SQL statements necessary to upgrade this table from LORIS version
+ * 22 to version 23.
+ * It relies on the existence of the `LorisMenu` table and the `modules` table
+ * so it must be used in the following way:
+ *
+ * 1. A user must apply the patch file:
```
johnsaigle: @ridz1208 @driusan Please let me know if this is the upgrade process you had in mind. I'm happy to modify the script to make the upgrade process smooth.

johnsaigle:
@ridz1208 What's the best way to apply these changes to Raisinbread?

ridz1208:
source RB, run your tool, dump RB

johnsaigle:
I'm not sure where the foreign key should be added since `0000-00-00-schema.sql` creates `issues` but it needs a FK references to `0000-00-02-Modules.sql`. Any preference?

driusan:
This shouldn't be in the changelog, it should be in the release notes. This is a one time use script that no one cares about after their upgrade is finished.

tools/single_use/22.0_To_23.0.php
driusan: We'll need to change the patch name before the release and make sure it's not duplicated in the release patch, but yes, this seems reasonable.

ridz1208: adressed in my PR

ridz1208:
added blocked and needs changelog update.

blocked by https://github.com/aces/Loris/pull/6120 which contains a notes for existing projects section where this can be disclosed

SQL/0000-00-02-Modules.sql
```diff
@@ -47,3 +47,5 @@ INSERT INTO modules (Name, Active) VALUES ('survey_accounts', 'Y');
 INSERT INTO modules (Name, Active) VALUES ('timepoint_list', 'Y');
 INSERT INTO modules (Name, Active) VALUES ('user_accounts', 'Y');
 INSERT INTO modules (Name, Active) VALUES ('electrophysiology_browser', 'Y');
+
+ALTER TABLE issues ADD CONSTRAINT `fk_issues_7` FOREIGN KEY (`module`) REFERENCES `modules` (`ID`);
```
driusan: This doesn't belong in the Modules.sql, it belongs in wherever the 'issues' table is defined in the default schema.

tools/single_use/Convert_LorisMenuID_to_ModuleID.php
```diff
@@ -0,0 +1,163 @@
...
+unset($result);
+
+
+// Use mapping arrays to replace old `LorisMenu` ID values in issue_tracker
+// `module` column with new `modules` table ID values.
+$result = $DB->pselect('SELECT issueID,module FROM issues', array());
+$issueTrackerMapping = [];
+foreach ($result as $row) {
+    // Skip rows in this table that do not have a module associated with them.
+    if (is_null($row['module'])) {
```
driusan: just add `WHERE module IS NOT NULL` to the select statement.

SQL/0000-00-02-Modules.sql
ridz1208: cant add a foreign key to a table that doesnt exist

driusan: I see.. I still don't like it, but you have a point.

ridz1208: same here... it's the consequence of having multiple schema files

PR #6083, opened 2020-02-19
[webpack] Make module exports importable from LORIS

driusan:
Currently, there is no way to access a module's exported
components from plain javascript in LORIS. This refactors
the webpack config so that each module's exports can be
accessed in lorisjs.modulename.exportname. For instance,
if a LORIS module named "MyModule" defines as class as
`export class MyClass {}` it would be accessible as
`lorisjs.MyModule.MyClass`. Note that any exports done
via statements at the end of the file such as `export default MyClass;`
(separate from the named class definition) are accessed as
`lorisjs.MyModule.default`, not `lorisjs.MyModule.MyClass`. A
module may only have one default export, or it will be overridden
according to webpack's whims.

Everything in LORIS should work identically to before with the
exception that the exports are now available via plain javascript.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -2,96 +2,21 @@ const TerserPlugin = require('terser-webpack-plugin');
 const path = require('path');
 const fs = require('fs');
 
-const config = [{
-  entry: {
-    './htdocs/js/components/DynamicDataTable.js': './jsx/DynamicDataTable.js',
-    './htdocs/js/components/PaginationLinks.js': './jsx/PaginationLinks.js',
-    './htdocs/js/components/StaticDataTable.js': './jsx/StaticDataTable.js',
-    './htdocs/js/components/MultiSelectDropdown.js': './jsx/MultiSelectDropdown.js',
-    './htdocs/js/components/Breadcrumbs.js': './jsx/Breadcrumbs.js',
-    './htdocs/js/components/Form.js': './jsx/Form.js',
-    './htdocs/js/components/Markdown.js': './jsx/Markdown.js',
-    './modules/media/js/mediaIndex.js': './modules/media/jsx/mediaIndex.js',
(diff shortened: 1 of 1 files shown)


johnsaigle:
I appreciate you requesting my review but I don't know enough JS to comment on the work you've done here.

driusan:
@johnsaigle neither do I, I just requested review from the people who I thought had a slight hope of understanding webpack..

maltheism:
@driusan I might be confused on the PR's intentions. In Loris you can import components from other modules. Example: `import CommentList from '../../issue_tracker/jsx/CommentList';` from within the new_profile module.

driusan:
@maltheism You can not currently access exports from anywhere other than things compiled by webpack

maltheism:
Hey @driusan I went through a few modules logging the `lorisjs` variable.

Visiting the new_profile module:
```
> console.log(lorisjs);
[Log] {new_profile: {__esModule: true, Symbol(Symbol.toStringTag): "Module"}}
```

Visiting the behavioural_qc module:
```
> console.log(lorisjs);
behavioural_qc: {__esModule: true, default: Object, Symbol(Symbol.toStringTag): "Module"}
```

Visiting the dashboard module:
```
< undefined
> console.log(lorisjs);
< ReferenceError: Can't find variable: lorisjs
```

driusan:
@maltheism the dashboard doesn't seem to have a jsx directory and isn't in the `webpack.config.js` file. I don't think it exports anything.

ridz1208:
@driusan @maltheism was this tested with the webpack project extension ??

driusan:
I have no way of testing it with the webpack project extension since I don't know the format of any project webpack configs, but when I modified it I looked at the code and from my understanding of how it's supposed to work it should (the index/property being modified still exists)

ridz1208:
I will test it out and try to fix if necessary

maltheism:
@ridz1208 it was not and I don't have much knowledge on how to test project extensions. I do theorize the project extension shouldn't be broken from the changes in this PR.

PR #6084, opened 2020-02-19
[CandID] Fix validation of CandID object types

driusan:
The CandID object type was verifying that a CandID had 6 integers
in it. It did not validate that it only had 6 integers, and as a
result was considering everything that started with 6 integers valid
even if there was trailing invalid characters. This fixes the regular
expression so that CandID are 6 digit integers, and only 6 digit integers.


src/StudyEntities/Candidate/CandID.php
```diff
@@ -64,7 +64,7 @@ public function getType(): string
      */
     protected function validate(string $value): bool
     {
-        $pattern = sprintf("/[0-9]{%s}/", self::LENGTH);
+        $pattern = sprintf("/^[0-9]{%s}$/", self::LENGTH);
 
         return preg_match($pattern, $value) === 1 &&
             intval($value) >= self::MIN_VALUE;
```
johnsaigle: If I was writing this for the first time I'd probably do the following instead of using a regex:

```php
return mb_strlen($value) === self::LENGTH 
    && \Utility::valueIsPositiveInteger($value)
    && intval($value) >= self::MIN_VALUE;
```

I'm sure this fix is good though.

xlecours: But it 
    spread 
    over several
    lines.
Arguably, a regex is not easier to read.

driusan: It's already a regex, the expression is just wrong.

PR #6090, opened 2020-02-21
[JSX] Add CSSGrid component type

driusan:
This adds a CSSGrid component type. A CSSGrid aligns
a number of cards (passed as a prop) in a grid. The
component takes care of converting the values passed
to it into cards with an appropriate title, size
and positioning.

PR #6091, opened 2020-02-21
[JS] Fix eslint errors in Card and Panel classes

driusan:
This fixes the warnings from eslint in the Card and
Panel classes.

These had old warnings that were triggered while trying to validate that #6090 did not introduce any new errors.

### DIFF
--- a/jsx/Card.js
+++ b/jsx/Card.js
@@ -15,15 +15,30 @@ import Panel from 'jsx/Panel';
  *
  */
 class Card extends Component {
+  /**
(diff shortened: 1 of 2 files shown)


jsx/Card.js
```diff
@@ -15,15 +15,30 @@ import Panel from 'jsx/Panel';
  *
  */
 class Card extends Component {
+  /**
+   * Construct the React components
```
driusan:
```suggestion
   * Construct the React component
```

PR #6092, opened 2020-02-21
[candidate_profile] Add new candidate dashboard module to provide overview of candidates

driusan:
This adds the framework for a candidate dashboard, genericifying the work started by @zaliqarosli. It only adds the core framework so that modules can register widgets into the dashboard and new cards will be added to the appropriate modules in subsequent PRs.

This adds an `$options` argument to `\Module->getWidgets` for type-specific options. The candidate dashboard uses it to pass the `\Candidate` object whose data is being displayed.

Widgets registered to the dashboard consist of a React component which get wrapped in a Card component, and displayed in a CSS grid. The module passes the props of `BaseURL`, `Candidate`, `Visits`, and `VisitMap`, plus any props passed to the `\LORIS\candidate_profile\CandidateWidget` object returned by `getWidgets` to the React component. `BaseURL` is the LORIS base URL. `Candidate` is the candidate object returned by the API, `Visits` is an array of all visits for that candidate returned from the API, and `VisitMap` is a mapping of VisitLabel -> SessionID to be used for links (where appropriate) since the API doesn't expose this data. These default props are helpers to avoid redundant requests to the API, since nearly every card will need some combination of this data.

The only card provided by the module itself is a candidate info card, which summarizes the data returned by the API.

~Blocked by #6083~
~Blocked by #6090~
~Blocked by #6093~
~Blocked by #6098~


jsx/CSSGrid.js
```diff
@@ -0,0 +1,45 @@
...
+      style.gridColumnEnd = 'span ' + value.Width;
+    }
+    if (value.Height) {
+      style.gridRowEnd = 'span ' + value.Height;
+    }
+    if (value.Order) {
+      style.order = value.Order;
+    }
+
+    style.alignSelf = 'stretch';
```
maltheism: Hey @driusan, one sugge
```
const style = {
     ...(value.width ? {gridColumnEnd: `span $value.Width`} : {}),
     ...(value.Height ? {gridRowEnd: `span $value.Height`} : {}),
     ...(value.Order ? {order: value.Order} : {}),
     alignSelf: 'strech'
   };
```
You could have it like that instead of the if statements for adding to the style object. Thought it might be worth mentioning. You may not like.

driusan: This file is just pulled in from #6093, it should go there (and I find it order to much read with the ternary operator, I don't think it being const makes affects anything enough that we should sacrifice readability.)

maltheism: @driusan best to ignore my suggestion if it's harder to read.

johnsaigle:

driusan:
@johnsaigle test each of the blocking PRs individually then re-visit?

modules/candidate_profile/php/candidate_profile.class.inc
```diff
@@ -0,0 +1,151 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

modules/candidate_profile/php/candidate_profile.class.inc
```diff
@@ -0,0 +1,151 @@
...
+    {
+        $this->candidate = $request->getAttribute('candidate');
+        if ($this->candidate === null) {
+            // This should be impossible, because the Module router added the
+            // attribute before loading this page.
+            throw new \LorisException("No candidate provided");
+        }
+
+        $DB   = \NDB_Factory::singleton()->database();
+        $user = \User::singleton();
```
johnsaigle: Could you use the factory here instead?

modules/candidate_profile/php/module.class.inc
```diff
@@ -0,0 +1,88 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

modules/candidate_profile/php/module.class.inc
```diff
@@ -0,0 +1,88 @@
...
+            $candidate = \Candidate::singleton($candID);
+
+            $request = $request->withAttribute('candidate', $candidate);
+            $page    = $this->loadPage("candidate_profile");
+            return $page->process($request, $page);
+
+        } catch (\DomainException | \LORISException $e) {
+            // A LORISException means the \Candidate::singleton couldn't
+            // load the candidate (ie because the CandID doesn't exist
+            // in the database.) A \DomainException means the CandID format
```
johnsaigle: If the CandID format is invalid shouldn't that be a 400 instead of a 404?

modules/candidate_profile/php/module.class.inc
```diff
@@ -0,0 +1,88 @@
...
+     * {@inheritDoc}
+     *
+     * @param string $type    The type of widgets to get
+     * @param \User  $user    The user widgets are being retrieved for.
+     * @param array  $options A type dependent list of options to provide
+     *                        to the widget.
+     *
+     * @return \LORIS\GUI\Widget[]
+     */
+    public function getWidgets(string $type, \User $user, array $options) : array
```
johnsaigle: I don't see that param `$options` being used in this PR. Can you explain why you added it? I'm not really able to work out what it's supposed to do based on the PHPDoc.

modules/candidate_profile/php/module.class.inc
driusan: No, if the CandID format is not valid that means they accessed a url like '/candidate_profile/abc123'. The URL doesn't exist, it's not found.

modules/candidate_profile/php/module.class.inc
driusan: It's passed at https://github.com/aces/Loris/pull/6092/files#diff-6dadba9ad9ecd0d6c9ebb445ce787335R83. This doesn't have any widgets, but other modules can use it to customize the widget in some way for the candidate (ie adding props or selecting data from the database.)

webpack.config.js
```diff
@@ -83,7 +83,7 @@ function lorisModule(mname, entries) {
         output: {
             path: path.resolve(__dirname, 'modules') + '/' + mname + '/js/',
             filename: '[name].js',
-            library: ['lorisjs', mname],
+            library: ['lorisjs', mname, '[name]'],
```
driusan: This is required otherwise the last export always overwrites the whole module in the lorisjs object.

webpack.config.js
```diff
@@ -104,21 +104,20 @@ const config = [
...
+            'Markdown': './jsx/Markdown.js',
+            'CSSGrid': './jsx/CSSGrid.js',
         },
         output: {
-            path: __dirname + '/',
-            filename: '[name]',
+            path: path.resolve(__dirname, 'htdocs/js/components'),
+            filename: '[name].js',
+            library: '[name]',
+            libraryTarget: 'window',
```
driusan: this is so that we can access the CSSGrid component from plain javascript in the template. The files should all still be compiled to the same location.

zaliqarosli:
@driusan from where can I access the candidate_profile widget or module?

driusan:
@zaliqarosli the URL. `$LORIS/candidate_profile/$candID` .. it's not linked to from anywhere in the UI because it's not useful until widgets for all modalities are added. This just adds the framework for adding widgets so that new widgets themselves can be reviewed in separate reasonably-sized PRs.

zaliqarosli:
@driusan thanks. thats fine, i think it makes sense. in this PR's description above, you said "The only card provided by the module itself is a candidate info card, which summarizes the data returned by the API." however, when i go to e.g. `http://zrosli-dev.loris.ca/candidate_profile/300014/`, all I get is an empty page. is there supposed to be atleast one card there with the candidate info?

driusan:
@zaliqarosli yes.. there should be one. Did you run `make` or `make dev`? Are there any js console or apache errors?

driusan:
And do you have access to the candidate? (at least one project and site in common..?) (There should be an error page if you don't and not a blank page, but it's worth verifying.)

driusan:
@zaliqarosli I added `SQL/New_patches/2020-02-24-CandidateProfileModule.sql`.. I don't think it'll show up without it. You can either run that SQL patch or run `tools/manage_modules.php --add` to add it to the modules table so that the widget from the candidate_profile is enabled.

modules/candidate_profile/php/candidate_profile.class.inc
```diff
@@ -0,0 +1,151 @@
...
+            throw new \LorisException("No candidate provided");
+        }
+
+        $factory = \NDB_Factory::singleton();
+        $DB      = $factory->database();
+        $user    = $factory->user();
+
+        $modules = \Module::getActiveModules($DB);
+
+        $widgets = [];
```
zaliqarosli:
```suggestion
        $widgets = array();
```

standard?

modules/candidate_profile/php/module.class.inc
```diff
@@ -0,0 +1,88 @@
...
+        case 'candidate':
+            $factory = \NDB_Factory::singleton();
+            $baseurl = $factory->settings()->getBaseURL();
+
+            return [
+                new CandidateWidget(
+                    "Candidate Info",
+                    $baseurl . "/candidate_profile/js/CandidateInfo.js",
+                    "lorisjs.candidate_profile.CandidateInfo.CandidateInfo",
+                    [],
```
zaliqarosli:
```suggestion
                    array(),
```

driusan:
@zaliqarosli I can't figure out any way it would be possible since the JS code needs to be dynamically generate

modules/candidate_profile/jsx/CandidateInfo.js
```diff
@@ -0,0 +1,183 @@
...
+            {
+                label: 'Visits',
+                value: this.getVisitList(this.props.Visits),
+                flexBasis: '40%',
+            },
+        ];
+
+        const cardInfo = data.map((info, index) => {
+            const cardStyle = {
+                flexBasis: info.flexBasis || '20%',
```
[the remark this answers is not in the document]
driusan: I'm not sure I understand how changing this to 25% would fix things.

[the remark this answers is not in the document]
zaliqarosli: wouldn't it render the data in 4 columns instead of 5? so then at least in the screenshot example above, it would render 4,4,1 instead of 5,3,1

modules/bvl_feedback/php/module.class.inc
```diff
@@ -51,12 +51,14 @@ class Module extends \Module
     /**
      * {@inheritDoc}
      *
-     * @param string $type The type of widgets to get
-     * @param \User  $user The user widgets are being retrieved for.
+     * @param string $type    The type of widgets to get
```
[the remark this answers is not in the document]
driusan: I've reflected this in PR #6098.. can you review over there? This will be resolved in this PR once that gets merged and I rebase it.

modules/candidate_profile/jsx/CandidateInfo.js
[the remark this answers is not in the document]
driusan: I think we want it to be dynamic based on the size of the card, not a fixed grid (otherwise it should be grid and not the flexbox that we're using..) I've converted it to a font relative unit instead of a container percentage relative unit, can you see it you like this better at a variety of screen sizes?

driusan:
@zaliqarosli I think GitHub says all the changes from your review are incorporated.. can you re-review?

zaliqarosli:
@driusan getting an error:

```PHP Fatal error:  Uncaught Error: Call to a member function isAccessibleBy() on null in /var/www/loris/modules/candidate_profile/php/candidate_profile.class.inc:34\nStack trace:\n#0 /var/www/loris/php/libraries/Module.class.inc(351): LORIS\\candidate_profile\\Candidate_Profile->_hasAccess(Object(User))\n#1 /var/www/loris/modules/candidate_profile/php/module.class.inc(40): Module->handle(Object(Laminas\\Diactoros\\ServerRequest))\n#2 /var/www/loris/src/Middleware/ResponseGenerator.php(50): LORIS\\candidate_profile\\Module->handle(Object(Laminas\\Diactoros\\ServerRequest))\n#3 /var/www/loris/src/Middleware/AuthMiddleware.php(63): LORIS\\Middleware\\ResponseGenerator->process(Object(Laminas\\Diactoros\\ServerRequest), Object(LORIS\\candidate_profile\\Module))\n#4 /var/www/loris/src/Router/ModuleRouter.php(96): LORIS\\Middleware\\AuthMiddleware->process(Object(Laminas\\Diactoros\\ServerRequest), Object(LORIS\\candidate_profile\\Module))\n#5 /var/www/loris/src/Router/BaseRouter.php(105): LORIS\\Router\\ModuleRouter->handle(Object(Laminas\\Diactoros\\ServerReques in /var/www/loris/modules/candidate_profile/php/candidate_profile.class.inc on line 34
```

driusan:
@zaliqarosli odd.. that was added in #6093 and this doesn't delete it. I think there's also some PHPCSing to do after a rebase, I'll tag you when it's good to try again..

driusan:
@zaliqarosli I've fixed the PHPCS errors but I'm not able to reproduce your problem.. have you run `make dev`? Is it a valid candidate? What URL are you going to?

zaliqarosli:
@driusan i have run `make dev` and accessing `$LORIS/candidate_profile/$candID`. the breadcrumb `Candidate Dashboard $candID / $pscid` is rendering which shows that it's a valid candidate but I get the front-end message `An error occurred while loading the page.`. My apache log has the same error as the one i posted above. looking into `php/candidate_profile.class.inc`, `$this->candidate` is null but I'm not sure why

johnsaigle:
I'm getting a blank page with the same values as Zaliqa using Raisinbread (Candidate `300008`)

I checked out the branch, ran `make dev`, ran `tools/manage_modules.php --add`.

Here are the console errors:



Also, visiting `/candidate_profile/` with no CandID causes a blank crash with a 500 error.

>PHP Fatal error:  Uncaught Error: Call to a member function isAccessibleBy() on null in /var/www/loris/modules/candidate_profile/php/candidate_profile.class.inc:34

modules/candidate_profile/jsx/CandidateInfo.js
```diff
@@ -0,0 +1,183 @@
...
+                label: 'Visits',
+                value: this.getVisitList(this.props.Visits),
+                width: '12em',
+            },
+        ];
+
+        const cardInfo = data.map((info, index) => {
+            const cardStyle = {
+                width: info.width || '6em',
+                padding: '1em',
```
zaliqarosli: this padding value works better!

modules/candidate_profile/jsx/CandidateInfo.js
```diff
@@ -0,0 +1,183 @@
...
+                margin: 0,
```
zaliqarosli: could we add a left and right margin? info.value is really close to each other if the string is long 
Screenshot 2020-02-26 15 03 36

driusan: I'm not sure it'll help much if the string is even longer, but done..

driusan:
@johnsaigle the issues that rebasing introduced should be fixed and I've made hasAccess return false in that case since no one should at that URL there anyways.

johnsaigle:
Screen Shot 2020-02-26 at 16 30 16


This is loading now but can it be made wider? DoB will always be that length and it doesn't look very good to have it be split onto a second line

zaliqarosli:
i agree with @johnsaigle 's comment that the DoB display doesn't look very good. but besides, that this works!

driusan:
I added the CSS `whitespace: nowrap` attribute to the headers, is that better?

modules/candidate_profile/jsx/CandidateInfo.js
```diff
@@ -160,7 +160,7 @@ export class CandidateInfo extends Component {
             };
             return (
                 <div style={cardStyle} key={info.label}>
-                    <dt>{info.label}</dt>
+                    <dt style={{whiteSpace: 'nowrap'}}>{info.label}</dt>
```
zaliqarosli: would need it for the info.value too

driusan: no, we want info.Value to wrap

driusan: oh, I see what you mean in the screenshot. I'll add it to the DoB value tomorrow. We don't want that one to wrap.

johnsaigle: It's also a nice change for the headers to not wrap though

driusan:
@johnsaigle @zaliqarosli I added nowrap on the DoB date itself, but I can't reproduce it wrapping in the first place, so someone else will have to test it.

PR #6093, opened 2020-02-21
[Permissions] Add helper function to \Candidate class to validate accessibility

driusan:
Add AccessibleResource interface and implement for the Candidate class. This is used
to determine if a user can access a resource. Candidates can be accessed by a user
if they have a project and a center in common (either a RegistrationCenterID or a CenterID of a visit.)

### DIFF
--- a/modules/timepoint_list/php/timepoint_list.class.inc
+++ b/modules/timepoint_list/php/timepoint_list.class.inc
@@ -43,33 +43,8 @@ class Timepoint_List extends \NDB_Menu
         $candidate = \Candidate::singleton($this->candID);
 
         // check user permissions
-        if ($user->hasPermission('access_all_profiles')
-            || (in_array(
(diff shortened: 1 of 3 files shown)


johnsaigle:
I like this code, it's nice and clean. 

Can you clarify the role of the new class? Specifically I'm wondering if there are situations to prefer this over hasAccess?

driusan:
@johnsaigle I'm not sure I understand your question.. the function is primarily to be used inside of hasAccess (or similar) functions. The function is called "isAccessibleBy" because "hasAccess" is wrong. (ie. `$candidate->hasAccess($user)` implies it's determining whether the candidate has access to the user, not whether the user has access to the candidate.)

johnsaigle:
Timepoint list is not broken so I assume the code is working fine 👍

PR #6094, opened 2020-02-25
[ReadTheDocs] Add additional configuration instructions for enabling large file uploads

PapillonMcGill:
## Brief summary of changes
Add configuration setting in php.ini from user feedback

### DIFF
--- a/docs/wiki/00 - SERVER INSTALL AND CONFIGURATION/02 - Website Configuration/Handling Large File Uploads.md
+++ b/docs/wiki/00 - SERVER INSTALL AND CONFIGURATION/02 - Website Configuration/Handling Large File Uploads.md
@@ -7,8 +7,9 @@ Recommended sample values appropriate for compressed scans not exceeding 500MB i
 
 ```
 session.gc_maxlifetime = 10800  // After this number of seconds, stored data will be seen as 'garbage' and cleaned up by the garbage collection process.
-max_input_time = 10800          // Maximum amount of time each script may spend parsing request data (in seconds)
-max_execution_time = 10800      // Maximum execution time of each script (in seconds)
+max_input_time = 10800          // Maximum amount of time each script may spend parsing request data (in seconds).
+max_execution_time = 10800      // Maximum execution time of each script (in seconds).
 upload_max_filesize = 1024M     // Maximum allowed size for uploaded files.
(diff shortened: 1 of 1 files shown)


docs/wiki/00 - SERVER INSTALL AND CONFIGURATION/02 - Website Configuration/Handling Large File Uploads.md
```diff
@@ -7,8 +7,9 @@ Recommended sample values appropriate for compressed scans not exceeding 500MB i
 
 ```
 session.gc_maxlifetime = 10800  // After this number of seconds, stored data will be seen as 'garbage' and cleaned up by the garbage collection process.
-max_input_time = 10800          // Maximum amount of time each script may spend parsing request data (in seconds)
-max_execution_time = 10800      // Maximum execution time of each script (in seconds)
+max_input_time = 10800          // Maximum amount of time each script may spend parsing request data (in seconds).
+max_execution_time = 10800      // Maximum execution time of each script (in seconds).
 upload_max_filesize = 1024M     // Maximum allowed size for uploaded files.
 post_max_size = 1024M           // Maximum size of POST data that PHP will accept.
+memory_limit = 128M             // Maximum amount of memory a script can use.
```
johnsaigle:
```suggestion
memory_limit = 256M             // Maximum amount of memory a script can use.
```

128 seems to be the default value so I don't think this adds much. We should use a higher value to allow for larger file uploads

PapillonMcGill: @johnsaigle This is the size that solve problem for Alfredo. It is also the size of my VM and production server I manage.

johnsaigle: > I made the suggested changes in the link that Melanie sent but the error was still appearing. It turned out that the proper parameter to set was “memory_limit”. I set it like this:
memory_limit = **150M**
Now the imaging link is working.
Thank you for your help.
Regards,
Alfredo.

PapillonMcGill: @johnsaigle Sorry, I went from memory! No pun intended.

Issue #5941, opened 2020-01-17
[vendor/js/react] Outdated react & react-dom being used


maltheism:
**Describe the bug**
The browser is being sent either (development or production) version of react.js & react-dom.js from the server and found inside the htdocs/vendor/js/react directory. This happens inside the php/libraries/NDB_Page.class.inc on line 815-820.

The react dependency in package.json is a newer version than the one found in htdocs/vendor/js/react and the package.json react is basically being ignored.

I found out the problem today when trying to use a (newer) feature that our react in package.json supports. Updating the htdocs/vendor/js/react directory with the files from node_modules react & react-dom solved the issue.

I believe that we need to get rid of the htdocs/vendor/js/react directory because we should be able to just import react from js side and have the react & react-dom included in the exported module js files. Otherwise we need to keep updating the htdocs/vendor/js/react directory with newer files which isn't the standard practice. package.json is supposed to be enough with updating everything that is imported.

PR #6097, opened 2020-02-25
[htdocs/vendor/js/react] Automatic update library from node_modules

maltheism:
## Brief summary of changes

I've added the copy-webpack-plugin module for updating the `htdocs/vendor/js/react` libraries when package.json is updated.

The latest javascript files inside `node_modules` of the react library with the version being from `package.json` are copied & overwrite the existing files in `htdocs/vendor/js/react` that are served to the frontend.

Before this PR the react library from the package.json was not ever being served to the frontend and the old react library was instead.

1. Checkout PR
2. delete files inside htdocs/vendor/js/react
3. run make

* Resolves #  https://github.com/aces/Loris/issues/5941


package.json
```diff
@@ -20,16 +20,17 @@
     "@babel/plugin-proposal-object-rest-spread": "^7.6.2",
     "@babel/preset-env": "^7.6.3",
     "@babel/preset-react": "^7.6.3",
+    "alex": ">=8.0.1",
     "babel-eslint": "^10.0.1",
     "babel-loader": "^8.0.5",
+    "copy-webpack-plugin": "^5.1.1",
```
driusan: Why are the files modified instead of deleted in this PR if this is going to copy and overwrite them each time webpack is run?

package.json
```diff
@@ -20,16 +20,17 @@
     "@babel/plugin-proposal-object-rest-spread": "^7.6.2",
     "@babel/preset-env": "^7.6.3",
     "@babel/preset-react": "^7.6.3",
+    "alex": ">=8.0.1",
```
johnsaigle: Looks like a rebase issue. This dependency shouldn't be added in this PR

package.json
maltheism: @driusan that makes sense. They're modified here because current LORIS doesn't use the same react libraries served to the frontend in htdocs that node_modules has from package.json. I can remove them because when make will generate them.

maltheism:
@driusan latest commit adds react-dom as well.

driusan:
@maltheism conflicts

maltheism:
@driusan conflicts fixed.

PR #6098, opened 2020-02-25
[Widgets] Update signature of getWidgets to include options

driusan:
This factors out the signature change from PR #6092 to simplify
the review process.

The signature of getWidgets if changed to include a weakly typed
'$options' parameter, which can be used to specify type-dependent
options.


driusan:
@zaliqarosli requesting your review here since you commented on this part of the original PR.. if you don't care and you're fine with @johnsaigle's approval just let me know.

zaliqarosli:
@driusan i'm fine with it!

PR #6099, opened 2020-02-25
Remove empty "Upgrading 17.0 to 17.1" document

johnsaigle:
We don't support this version of LORIS anymore so this document doe snot need to come forward with us into the new ReadTheDocs wiki.

There's also no content in it


ridz1208:
@johnsaigle I dunno.... this is a risky deletion

johnsaigle:
Maybe we should push this to release 29.0.7 to be safe?

johnsaigle:
@ridz1208 pls rereview

driusan:
So in the end has anyone successfully updated from 17.0 to 17.1?

PR #6048, opened 2020-02-11
Update Creating a Module guide

johnsaigle:
Creates a guide for creating a modern LORIS module and adds it to Read The Docs.

Also moves the Help Text Style guide to the Developers section of the wiki.

Fixes #5103

PR #6105, opened 2020-02-25
[ReadTheDocs] Revise "Code Review Checklist" and add to wiki

johnsaigle:
## Brief summary of changes

Revises the "Code Review Checklist" from the GitHub wiki. I've changed the tone to be more of a 'code review guide' because I don't think anyone actually sits down and checks these things off.


docs/wiki/99 - Developers/Code Review Guide.md
```diff
@@ -0,0 +1,49 @@
...
+3. The pull request must have an 'Approved' review status.
+
+This document will help you to understand how to get your code to be Approved.
+
+## Checklist 
+
+In addition to what's listed in the CONTRIBUTING document, here are some things we are
+looking for when deciding whether to Approve your pull request.
+
+- [ ] 1.  Do database calls use prepared statements?
```
driusan: I think these should either have the checkbox or the numbering, not both.

docs/wiki/99 - Developers/Code Review Guide.md
```diff
@@ -0,0 +1,49 @@
...
+- [ ] 3.  Is CSS/Javascript/PHP/HTML code separated into their own files?
+- [ ] 4.  Have database changes been reflected in the schema, a new patch, and
+in Raisinbread?
+- [ ] 5.  Are the comments reasonable/understandable?
+- [ ] 6.  Are the variable names meaningful? In the proper scope?
+- [ ] 7.  Does the logic/structure make sense? Is it as efficient as possible?
+- [ ] 8.  Are inheritance/classes/etc used appropriately?
+- [ ] 9.  Is there an effort to limit code duplication?
+- [ ] 10. Are functions concise and clear?
+- [ ] 11. Are the commit messages meaningful, atomic, and rebased/squashed into a clean tree?
```
driusan: This isn't looked at in code review, it's squashed when it's merged. You can remove this point.

docs/wiki/99 - Developers/Code Review Guide.md
```diff
@@ -0,0 +1,49 @@
...
+- [ ] 12. Is the pull request appropriately named (starting with Module name or Core) and appropriately labelled?
+- [ ] 13. Is the comment on the pull request meaningful and informative, even to someone with limited knowledge of the code?
+- [ ] 14. Are there any other implications for existing projects that need to be documented or addressed? 
+- [ ] 15. Are changes made in a backward compatible manner (where possible)?
+- [ ] 16. Has relevant documentation been updated?
+- [ ] 17. Is the code written with thought given to how it can be maintained and extended? Is it written elegantly, without "hacks"?
```
driusan: - Does it use the `handle` function and not an `ajax` directory?
- Where possible, is new frontend code written using React rather than smarty templates?
- Are URLs structured as a well-designed API which follows the HTTP specification, rather than a single magic endpoint that changes behaviour via a multitude of query parameters?

johnsaigle: > Are URLs structured as a well-designed API which follows the HTTP specification, rather than a single magic endpoint that changes behaviour via a multitude of query parameters?

I agree but I feel like very few people are going to understand what this sentence actually means. We might need a more in-depth document that gets into the nitty-gritty of LORIS PHP guts.

PapillonMcGill: Once written, this document should be referenced here!

driusan: Whether they're going to understand the comment or not (I don't care if it's rewritten), it's the most common problem that prevents modules from getting merged and takes a long time to resolve, so it should be here in some form and people should be thinking about it before it gets to code review.

docs/wiki/99 - Developers/Code Review Guide.md
```diff
@@ -0,0 +1,50 @@
...
+- [ ] Is there an effort to limit code duplication?
+- [ ] Are functions concise and clear?
+- [ ] Is the pull request appropriately named (starting with Module name or Core) and appropriately labelled?
+- [ ] Is the comment on the pull request meaningful and informative, even to someone with limited knowledge of the code?
+- [ ] Are there any other implications for existing projects that need to be documented or addressed? 
+- [ ] Are changes made in a backward compatible manner (where possible)?
+- [ ] Has relevant documentation been updated?
+- [ ] Is the code written with thought given to how it can be maintained and extended? Is it written elegantly, without "hacks"?
+- [ ] Is it [PSR7](https://www.php-fig.org/psr/psr-7/) compliant?
+- [ ] Where possible, is new frontend code written using ReactJS rather than Smarty templates?
```
johnsaigle:
```suggestion
- [ ] Where possible, is new frontend code written using ReactJS rather than Smarty templates?
- [ ] Does new code follow the HTTP specification? Is it separated into specific endpoints?
```

docs/wiki/99 - Developers/Code Review Guide.md
johnsaigle: I added a brief note here. Maybe this point can be added in more detail to #6048 because this is an important point for people submitting new modules.

Issue #6066, opened 2020-02-13
[Document Repository] Module broken, does not load


johnsaigle:
**Describe the bug**
When loading the module, the table and filters load briefly then disappear, leaving a blank page. 

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to aces/master
2. Go to the Document Repository

**What did you expect to happen?**
The module should load.

**Browser Environment (please complete the following information):**
- OS: macOS Catalina 
- [x] Tested on Chrome and Firefox
- Version aces/master

**Additional context**
There is a lot of information in the browser console.log so a UI person should look here.

driusan:
PR merged

PR #6106, opened 2020-02-25
[document_repository] fix can't load 

kongtiaowang:
solve :[Document Repository] Module broken, does not load https://github.com/aces/Loris/issues/6066
 wrong variable name.

### DIFF
--- a/modules/document_repository/php/doctree.class.inc
+++ b/modules/document_repository/php/doctree.class.inc
@@ -166,8 +166,8 @@ class DocTree extends \NDB_Page
 
         $subcategories    = $db->pselect(
             "SELECT id,category_name,comments 
-             FROM document_repository_categories WHERE parent_id=:pid",
-            array('pid' => $pid)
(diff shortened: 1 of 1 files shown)


driusan:
Shouldn't there be a test to make sure that it didn't load before and that this fixes it?

johnsaigle:
It was a weird bug... the page did load but the contents got auto-hidden from the user. The doc repo integration tests are really basic so it's possible they are still contained in the HTML response body so Travis had no issue.

If it had tests for menu filters, buttons, etc then I imagine they would've failed.

kongtiaowang:
To test the react page, it is necessary to do "Ctrl + Shift + R" to reload the currently open page and override the locally cached version after running make dev.
@maltheism Could u manually test this PR?

kongtiaowang:
@johnsaigle when you load page "document_repository/doctree/0" if the subcategories are empty,  you may not on this branch.

johnsaigle:
I only tested `/document_repository/` and the page was still blank.

PR #2764, opened 2017-04-24
[Tools] Added tool to check for data consistency in flags and instrument tables

AnyhowStep:
The code does these checks:

+ Tables specified in `test_names` exist
+ Orphan flag rows
+ Orphan instrument rows
+ Duplicate instrument flags in a session

Run with: `php tools/instrument_flag_consistency_check.php`

Screenshots:

image
image

Issue #5096, opened 2019-08-23
[Tools] A data-integrity checker


ridz1208:
Untill all tables in the database are normalized, all foreign keys are created and all cascades properly defined, it would be useful to have a tool that simply crawls through a given LORIS database and reports any and all data inconsistencies such as:

- duplicated examiners (due to typos in names, double spacing in name, ...)
- non-jsondata instruments with no SQL tables associated
- jsondata instruments WITH a SQL table created
- orphan flag entries (dont exists in the instrument tables)
- orphan instrument entries (dont exist in the flag table)

This ticket can be expanded to cover every possible case of inconsistent data that would be caught in a flawless database setup.

johnsaigle:
Pasting @christinerogers comment from #5689 so that all of the desired DB checks can exist in one place

> quick notes:
~~(As john saigle mentioned ^^) #1053 is important~~ [_I do not remember saying this or know why it's relevant - JS_]
Date of birth should make sense, should not
check for duplicates - 1 participant with same instrument populated twice for 1 visit label (it happens)
DDE - over/under population

PR #6110, opened 2020-02-26
[Tools] Create tool to validate integrity of the database

### LINKED ISSUE #5689: Create data integrity check script to detect issues in the database
We should have a database integrity checker script and consistently add checks to it when we detect these kind of issues. I guarantee you EVERY project has orphaned rows, illogical and inconsistent data. a while back justinNG had started a script that covered several base cases, it was never merged and eventually closed. I think it's very valuable to start that up again. It speaks to the health of the project.

_Originally posted by @ridz1208 in https://github.com/aces/Loris/issues/5684#issuecomment-555620075_

Check #2764 for some preliminary work  on this script.

johnsaigle:
## Brief summary of changes

This builds on work that began in #2764. The aim is to create a script that does some basic validation on the state of the database, e.g. checking for duplicated or missing values.

Ideally this will be gradually remedied through the use of table normalization and foreign key constraints but in the mean time this can help us.

This script is not meant to be comprehensive but instead is something we can modify and iterate upon as needed.

1. Run the script
2. TODO

* Resolves #5689


johnsaigle:
Blocked by #5968

johnsaigle:
@ridz1208 Can you review since you were involved with the initial work of this script and created #5096 ?

tools/data_integrity_check.php
```diff
@@ -0,0 +1,129 @@
...
+    [$DB, 'tableExists']
+);
+$numberOfTestNames = count($instrument_arr);
+$numberOfTestNamesWithTables = count($instruments);
+if ($numberOfTestNames !== $numberOfTestNamesWithTables) {
+    $helper->printError(
+        'Found entries in the `test_names` table that do not have appear to '
+        . 'be installed: ' . join(', ', array_diff($instrument_arr, $instruments))
+    );
+}
```
ridz1208: this check is incorrect now, instruments can be JSONdata and not require a table

tools/data_integrity_check.php
```diff
@@ -0,0 +1,129 @@
...
+                SELECT
+                    COUNT(*)
+                FROM
+                    flag test
+                WHERE
+                    test.CommentID LIKE 'DDE_%' AND
+                    flag.SessionID = test.SessionID AND
+                    flag.Test_name = test.Test_name
+            ) > 1
+        )
```
ridz1208: Wouldnt a query grouping by session and test_name with a count do the same thing here ?

it would be significantly quicker i think

PapillonMcGill: could use `having count(*) > 1` with the `group by` clause

johnsaigle: Could one of you provide more details? SQL isn't my strong suit

PapillonMcGill: @johnsaigle Can do but would like to know which table is that "test" table in the cross-join? Cross-join should almost never be used except some specific case.

johnsaigle: I think in this statement `test` is an alias of the `flag` table. I adapted this query from the original PR (linked in description) which I did not write.

tools/data_integrity_check.php
ridz1208: @johnsaigle this check should still be removed. JSON instruments don't need tables. or it should be improved to check if the instrument has the JSONdata flag to true before reporting an error.

ridz1208:
@johnsaigle other than my comment, the rest looks nice.

@CamilleBeau @pierre-p-s can one of you run this on CCNA, I'm curious to see the results.

pierre-p-s:
I am getting this output when testing on CCNA:

[output.txt](https://github.com/aces/Loris/files/4782100/output.txt)

tools/data_integrity_check.php
johnsaigle: How can I determine whether an instrument has JSON data?

ridz1208:
@johnsaigle other than checking for JSONdata instruments, you might need tocheck if a table exists on the `$this->table` variable of the instrument. sometimes the test names are different than the SQL table name (see CCNA output
)

johnsaigle:
Is it possible to make this change in another PR? I'm not sure I'll have time to figure this otua nd debug especially since I didn't write this code in the first place. Otherwise somebody else will probably have to take this over.

ridz1208:
@johnsaigle yes its possible. just remove the checks in this PR entirely
keep the orphan checks after

johnsaigle:
@ridz1208 done, this is ready for review

ridz1208:
@johnsaigle travis

ridz1208:
@driusan in your hands as well. passed travis

PR #4229, opened 2019-01-03
New instruments loading module

driusan:
This adds a new instrument loading module (named "instruments") in order to handle the loading of instrument files which was done by NDB_Caller. This removes the last remaining place that `main.php` was used, which can now be deleted.

The NDB_Caller is still required for `survey.php`, until surveys are also converted into a proper module with its own routing.

### To test this change...

- [ ] Test all varieties of instruments (PHP, Linst, survey) and that they both load and save correctly.

PR #6116, opened 2020-02-26
[NDB_BVL_Instrument_LINST] Show subtests

zaliqarosli:
## Brief summary of changes

Linst instrument subtests are blank and not able to render its fields. This is because of legacy request variable left over from before PR https://github.com/aces/Loris/pull/4229 that introduced a new instruments loading module.

This PR replaces the call for a request variable that doesn't exist with the class' $this->page

*This might only be testable by a project developer who has access to a multi-page linst instrument. Raisinbread currently only has the bmi.linst, which does not have subtests. 

1. Go to any multi-page php instrument with subtests. There will be no errors. You are able to view everything fine.
2. Do the same with a mutli-page linst instrument with subtests. You will only be able to see the 'top' page, and not any of the other subtests.
3. On this PR branch, you will be able to see the linst subtests.


xlecours:
[bob.linst.txt](https://github.com/aces/Loris/files/4308689/bob.linst.txt)

PR #6118, opened 2020-02-27
[Core] Add LorisInstance class

### LINKED ISSUE #6022: Create object to represent a LORIS Instance
There is currently no concept of an installed LORIS instance for LORIS. Things that would logically belong in such a class tend to be scattered across the codebase in static functions or pseudo-globals (ie. the utility class, singletons, various factories) or be hardcoded with no canonical source of truth (ie. the `project/` directory.)

We should create a class to represent the installed LORIS instance, pass it around where appropriate, and use it.

driusan:
This adds a class to LORIS to represent an installed LORIS instance.

It is primarily a stub that can be expanded in the future.

Resolves #6022

### DIFF
--- a/src/LorisInstance.php
+++ b/src/LorisInstance.php
@@ -78,7 +78,7 @@ class LorisInstance
     /**
      * Return true if the LORISInstance has a module named
      * $name. To be installed it must be both available on
-     * the filesystem, and active in the modules table.
+     * the filesystem and active in the modules table.
      *
      * @return bool
      */
--- a/src/LorisInstance.php
+++ b/src/LorisInstance.php
@@ -1,0 +1,107 @@
+<?php declare(strict_types=1);
+
+namespace LORIS;
+
+/**
+ * A LorisInstance represents an installed instance of a LORIS
+ * project.
+ *
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ */
+class LorisInstance
+{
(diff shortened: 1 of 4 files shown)


src/LorisInstance.php
```diff
@@ -0,0 +1,107 @@
...
+                error_log($e->getMessage() . " " . $e->getTraceAsString());
+            }
+        }
+        return $modules;
+    }
+
+    /**
+     * Return true if the LORISInstance has a module named
+     * $name. To be installed it must be both available on
+     * the filesystem, and active in the modules table.
```
PapillonMcGill:
```suggestion
     * the filesystem and active in the modules table.
```

src/LorisInstance.php
```diff
@@ -0,0 +1,107 @@
...
+    protected $modulesDirs;
+    private $DB;
+
+    /**
+     * Construct a LORISInstance for the install connected to $db
+     * which uses modules from $moduleDirs.
+     *
+     * @param \Database $db         A database connection to this
+     *                              instance.
+     * @param string[]  $moduleDirs A list of directories that may
```
johnsaigle: It might be nice to use SPLFileInfo instead especially if the main purpose of this property is to do validation based on the filepaths.

driusan: That is not the main purpose. The main purpose is to have a list of directories that modules may be found in.

PR #6121, opened 2020-02-27
[candidate_profile] Add behavioural data widget

driusan:
This adds a "Behavioural Data" widget to the candidate_profile
page for anyone who has access to the instruments module. The
Behavioural Data consists of a list of (initially collapsed)
visits which on click expand to a list of instruments and the data entry
percentage which is complete. Links are provided to the instrument
for data entry, the visit, and the conflict resolver (if applicable).

### DIFF
--- a/modules/instruments/php/visitsummary.class.inc
+++ b/modules/instruments/php/visitsummary.class.inc
@@ -1,0 +1,136 @@
+<?php declare(strict_types=1);
+
+namespace LORIS\instruments;
+
+use \Psr\Http\Message\ServerRequestInterface;
+use \Psr\Http\Message\ResponseInterface;
+use \LORIS\StudyEntities\Candidate\CandID;
+
+/**
+ * The VisitSummary endpoint returns a summary of instrument
+ * data entry for a particular candidate's visit. This is primarily
+ * to be used by the instrument list widget.
+ *
+ * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ */
+class VisitSummary extends \NDB_Page
+{
+    /**
+     * Temp for testing.
+     *
+     * @param \User $user The user whose access is to be verified
+     *
+     * @return bool
+     */
+    function _hasAccess(\User $user) : bool
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @param ServerRequestInterface $request The incoming PSR7 request
+     *
+     * @return ResponseInterface
+     */
+    public function handle(ServerRequestInterface $request) : ResponseInterface
+    {
+        $user = $request->getAttribute("user");
+        if ($user === null && !($user instanceof \User)) {
+            return new \LORIS\Http\Response\JSON\InternalServerError(
+                "No valid user"
+            );
+        }
+        if (!($this->_hasAccess($user))) {
+            return new \LORIS\Http\Response\JSON\Forbidden();
+        }
+
+        $params = $request->getQueryParams();
+        if (!isset($params['CandID'])) {
+            return new \LORIS\Http\Response\JSON\BadRequest(
+                "Missing CandID query parameter"
+            );
+        }
+
+        if (!isset($params['VisitLabel'])) {
+            return new \LORIS\Http\Response\JSON\BadRequest(
+                "Missing VisitLabel query parameter"
+            );
+        }
+
+        switch ($request->getMethod()) {
+        case 'GET':
+            return $this->_handleGet(
+                new CandID($params['CandID']),
+                $params['VisitLabel'],
+            );
+        default:
+            return new \LORIS\Http\Response\JSON\MethodNotAllowed(
+                array('GET')
+            );
+        }
+    }
+
+    /**
+     * Helper to specifically handle PATCH HTTP methods to the endpoint.
+     *
+     * @param CandID $candid     The CandID for the visit.
+     * @param string $visitLabel The visit label string.
+     *
+     * @return ResponseInterface
+     */
+    private function _handleGET(
+        CandID $candid, string $visitLabel
+    ) : ResponseInterface {
+        $DB = \NDB_Factory::singleton()->database();
+
+        $bvl_result = $DB->pselect(
+            "SELECT
+                f.Test_name,
+                f.CommentID,
+                COUNT(cu.ConflictID) AS NumOfConflict
+            FROM session s
+                LEFT JOIN flag f ON (s.ID = f.SessionID)
+                LEFT JOIN conflicts_unresolved cu
+                    ON (cu.CommentId1=f.CommentID OR cu.CommentID2=f.CommentID)
+                LEFT JOIN candidate c USING (CandID)
+            WHERE c.Active='Y' AND s.Active='Y' AND 
+                c.CandID=:candid AND s.Visit_Label=:vl
+                AND f.CommentID NOT LIKE 'DDE%'
+            GROUP BY
+                s.Visit_label,
+                s.ID,
+                s.Date_visit,
+                f.Test_name,
+                f.CommentID,
+                s.Current_stage,
+                s.Screening,
+                s.Visit,
+                s.Approval
+            ORDER BY s.Visit_label, f.Test_name, f.CommentID",
+            array('candid' => $candid, 'vl' => $visitLabel)
+        );
+
+        $instrumentList = \Utility::getAllInstruments();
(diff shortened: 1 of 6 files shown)


johnsaigle:
@driusan Can you rebase this onto #5949? That way I can test this all at once.

driusan:
@johnsaigle I don't want them to get mixed up and have to keep rebasing this one every time the other changes, but you can merge them into a temporary branch for testing.

driusan:
Oh wait, sorry, I just realized the function is already there it's just outdated.. I'll rebase it.

johnsaigle:
> Fatal error: Cannot redeclare NDB_BVL_Instrument::determineDataEntryCompletionProgress() in /app/php/libraries/NDB_BVL_Instrument.class.inc on line 2816

modules/instruments/php/visitsummary.class.inc
```diff
@@ -0,0 +1,135 @@
...
+ * The VisitSummary endpoint returns a summary of instrument
+ * data entry for a particular candidate's visit. This is primarily
+ * to be used by the instrument list widget.
+ *
+ * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ */
+class VisitSummary extends \NDB_Page
+{
+    /**
+     * Temp for testing.
```
johnsaigle: This should be updated

modules/instruments/php/visitsummary.class.inc
```diff
@@ -0,0 +1,135 @@
...
+     * @return ResponseInterface
+     */
+    public function handle(ServerRequestInterface $request) : ResponseInterface
+    {
+        $user = $request->getAttribute("user");
+        if ($user === null && !($user instanceof \User)) {
+            return new \LORIS\Http\Response\JSON\InternalServerError(
+                "No valid user"
+            );
+        }
```
johnsaigle: I'm a little confused about this part. Doesn't the Middleware do something to detect whether the User is logged in or whether it's anonymous/Unauthorized? Why is this being checked here in particular when it isn't in other modules?

modules/instruments/php/visitsummary.class.inc
```diff
@@ -0,0 +1,135 @@
...
+
+        if (!isset($params['VisitLabel'])) {
+            return new \LORIS\Http\Response\JSON\BadRequest(
+                "Missing VisitLabel query parameter"
+            );
+        }
+
+        switch ($request->getMethod()) {
+        case 'GET':
+            return $this->_handleGet(
```
johnsaigle:
```suggestion
            return $this->_handleGET(
```

johnsaigle:
> list of (initially collapsed)

The list was not initially collapsed for me; the progress bar appears on load. e.g. for candidate_profile 300008 I saw a partially-completed progress bar for V1 on Raisinbread.

@zaliqarosli @driusan Based on my understanding of this PR and #5949 I'm not sure the progress bar is working correctly. Using the CandID above, these are my results on Raisinbread

Screen Shot 2020-03-11 at 11 48 38
Screen Shot 2020-03-11 at 11 48 33

Since there are 4 instruments here, one of which is complete, I think the expected result is for the progress bar to be (at least) 25% full. However it appears to be filled out maybe 1 or 2% -- unless the small sliver of yellow is meant to represent 0%.

On my VM and within Raisinbread, this could be caused by 2 different things:

1. The Fatal Error I posted in my above comment ("Cannot redeclare NDB_BVL_Instrument::determineDataEntryCompletionProgress()")

2. My VM does not have the Raisinbread instruments (bmi, aosi, ...) installed. I'll fix this now, but it would be good to have some error-checking and UI information appear if this occurs in production.

modules/instruments/php/visitsummary.class.inc
driusan: We only have middleware to add appropriate page decorations depending on whether someone is logged in. The parent does access checking, but the parent isn't called here because we only want to return JSON responses.

At any rate, this check is here because without it our static analysis tools (I think it was phan, but it might have been one of the other ones) can't deduce that `$user` is an instance of `\User`. It should always be added to the request by the router, that's why it's an internal server error if it's not.

driusan:
The screenshot that you posted *is* collapsed, and the percentage that that function calculates (when expanded) is the percentage of data entry for a particular instrument, not the percentage of instruments complete for a visit.

modules/instruments/php/visitsummary.class.inc
johnsaigle: Fair enough. It might be overkill just to fix phan but I believe if we were to extend `ServerRequestInterface` (e.g. `LORISRequest`) and add a typed `getUser()` function then we could address that issue.

johnsaigle:
Thanks for clarifying. I'm still unsure why it would display as 0% though if radiology review is marked as complete.

johnsaigle:
Ah OK. It would be helpful if you could provide bulleted testing instructions. It wasn't clear to me that I had to click to get a different view. 

The UI doesn't make that very clear also... it would be nice if on hover the row changed colour or something so that it's more clear that you can interact with it. Without that, it only appears you can interact with the link to `instrument_list`. Because of that I assumed 'expanding' meant the collapse arrow on the top-right of the box.

When I _do_ click on it, only "Loading..." appears and in the error log I get the "Cannot declare..." error that I posted about earlier.

driusan:
The function re-declaration was caused by rebasing and should be fixed now, and I've added hover effects.

zaliqarosli:
@driusan could you add in the PR description how to access this page to test on the front-end? right now, it's unclear where we should be going on LORIS to see the BVL data widget

zaliqarosli:
also, could you explain how you are using the `determineDataEntryCompletionProgress()` function from https://github.com/aces/Loris/pull/5949? it looks like your implementation and display of it is different to my example in my PR

driusan:
@zaliqarosli `https://$LORISURL/candidate_profile/$candID`. There's not yet a link on the front end, but modules enough have widgets now that it might be worth adding one from the candidate list..

I'm not sure what you mean by your second question. The UI had to be changed because it didn't scale to large datasets with many visits and/or instruments. The columns which had a row span in your initial implementation are the summary/header of the visit when it's collapsed and remain there when it's not, because they always spanned 100% of the rows. The visits are collapsed by default to save space. The percentage is listed in a `<progress />` HTML element instead of as text.  Other than that, it's the same table.

johnsaigle:
Thanks for the changes. The UI is much more clear now and the Fatal Error is fixed.

The UI behaves strangely when a battery does not have any registered instruments.

It would be good to copy the message from instrument_list or maybe even disable the expansion where there are no instruments instead of rendering only the headers.

Here are some examples for DCCID `300005`

Screen Shot 2020-03-11 at 16 28 23
Screen Shot 2020-03-11 at 16 28 29

johnsaigle:
Screen Shot 2020-03-11 at 16 36 16

It looks like the progress bar is working fine. mri_parameter_form was halfway full and when I completed the form it became complete. 

The UI looks different from the screenshot in @zaliqarosli's PR though. I'm not sure if that's an issue.

I'll approve after the empty case I posted about above is handled.

zaliqarosli:
@driusan @johnsaigle i'm fine with the UI change. my only issue is that the instrument list table looks a bit too much like older Loris prior to our UI work of summer 2018 (2019? i've forgotten). If the UI could look closer to the current DataTable component, that would be great!

besides that though, i think the timepoint/instrument grouping looks great. I would definitely add a bit more padding above and below each instrument in the list though, just to spread it out a bit more, especially for the last instrument in the list which is quite close to the timepoint box border. i also think that the 'Instruments' header for the instrument list is superfluous as it's clear it's a list of instruments from the first column header

driusan:
@zaliqarosli This is a default `<table>` with no special styling applied other than specifying text alignment in some cells. If there's a problem with it I think it's a bug in the LORIS CSS, not the module.

driusan:
@johnsaigle I fixed the message when there's no instruments and disabled expanding when the visit is not started. (I couldn't disable when there's no instruments in a started instrument because we don't know that until after the click retrieves the data..)

driusan:
@zaliqarosli I just added the classes that IMO should be unnecessary for now, can you re-review when you get the chance?

zaliqarosli:
@driusan center align 'Conflicts?' and then I think we're all good

Screenshot 2020-03-30 16 25 06

driusan:
@zaliqarosli I can't test because my screen isn't wide enough to have whitespace around that header, but it should be centered now..

PR #6122, opened 2020-03-01
[webpack] restore support for webpack-project

ridz1208:
## Brief summary of changes
Restore functionality broken in #6083 

Projects can define their own targets for webpack to compile. this functionality was already there, this is just a fix for it to be functional again.

projects should define their file as such
filename:`webpack-project.config.js`
```
const entry = {
    'candidate_parameters' : ['CandidateParameters'],
    'study_tracker' : ['index'],
    'user_accounts' : ['userAccountsIndex']
};

module.exports = entry;
```

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -72,16 +72,22 @@ const mod = {
  *
  * @return {object} - The webpack configuration
  */
-function lorisModule(mname, entries) {
+function lorisModule(mname, entries, override=false) {
     let entObj = {};
+    let base = './modules';
+
+    if (override) {
+        base = './project/modules';
+    }
+
     for (let i = 0; i < entries.length; i++) {
         entObj[entries[i]] =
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -203,8 +210,11 @@ const config = [
 
 // Support project overrides
 if (fs.existsSync('./project/webpack-project.config.js')) {
-  const projConfig = require('./project/webpack-project.config.js');
-  config[0].entry = Object.assign(config[0].entry, projConfig);
+    const projConfig = require('./project/webpack-project.config.js');
+
+    for (const [module, files] of Object.entries(projConfig)) {
```
driusan: Do you have an example `webpack-project.config.js` so I can try and figure out if this part makes sense?

ridz1208: its in the description
thats litterally the entire file

driusan: okay, looks good to me.

PR #5253, opened 2019-09-30
[WIKI] Install - Rewrite "Installing Loris in Depth" for new Wiki

johnsaigle:
This PR wewrites the wiki page ["Installing LORIS in-depth"](https://github.com/aces/Loris/wiki/Installing-Loris-in-Depth) and commits it to version control.

The wording is updated to reflect recent PRs updating the install process, such as the Ubuntu and CentOS readmes. (e.g. #5203)

The rewrite removes references to `ssh` and also makes clear where the install process diverges for production installs vs. development installs.

Resolves #5001

PR #6123, opened 2020-03-02
Change Travis distribution target from Ubuntu 16.04 to 18.04

johnsaigle:
## Brief summary of changes

16.04 is nearly 4 years old now. This PR changes Travis to test a more recent OS. It looks like it may also cut down on the start up time of the tests.

Modifications to the config file are made based on the official Travis update guide. https://docs.travis-ci.com/user/reference/bionic/

* #5253 should be updated with version information following this PR being merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ language: php
 php: 
   - 7.3
   - 7.4
+    
+# Testing Ubuntu 18.04
+dist: bionic
 
 matrix:
     fast_finish: true
@@ -15,6 +18,7 @@ env:
   - TEST_SUITE=integration
 
 services:
+- mysql
 - docker
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,3 @@
-sudo: required
-
 language: php
 
 php: 
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
 - sudo apt-get update
 - sudo apt-get install -y libapparmor1
 - pecl install -f ast-1.0.3
-- sudo apt-get install npm
+- sudo apt-get install nodejs-dev node-gyp libssl1.0-dev npm
 - make dev
 
 script: "npm run tests:$TEST_SUITE"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: required
+
 language: php
 
 php: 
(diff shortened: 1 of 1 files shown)


johnsaigle:
Making additional changes based on recommendations from Travis about our config:

Screen Shot 2020-03-02 at 11 25 58

.travis.yml
```diff
@@ -1,14 +1,16 @@
...
+os: linux
+dist: bionic
+
+jobs:
   include:
   - php: "7.3"
   - php: "7.4"
 
 services:
+- mysql
```
kongtiaowang: This is not necessary. Docker will create mysql service for testing.

johnsaigle: It has changed for 18.04. It explains in the link

kongtiaowang: "On the Ubuntu 18.04 based environment, to speed up boot time and improve performance we’ve disabled all services by default. Add any services that you want to start to your .travis.yml:  "
Integration testing only needs mysql service from docker system not from travis.

driusan: The link just says they've disabled all services by default (that I see).. I think @kongtiaowang's point is that we don't use the Travis mysql service, it gets launched from docker.

johnsaigle:
```suggestion
```

.travis.yml
```diff
@@ -23,7 +23,7 @@ install:
 - sudo apt-get update
 - sudo apt-get install -y libapparmor1
 - pecl install -f ast-1.0.3
-- sudo apt-get install npm
+- sudo apt-get install nodejs-dev node-gyp libssl1.0-dev npm
```
driusan, line 26: what are these new dependencies? Are they the same as documented elsewhere in our documentation?

johnsaigle: I think this is specific to the way Travis preloads things for Bionic. This was required to get node to stall properly (which in turn installs npm). Without this line, the build will fail

ridz1208: @kongtiaowang can you try installing LORIS on a new ubuntu 18.04 machine without these added dependencies and see if it works ?

John is saying these changes are needed for travis but we want to know if they are also needed for regular installs

driusan: @kongtiaowang

ridz1208: @kongtiaowang please answer comment above

cmadjar:
@kongtiaowang can you check if this is required outside of Travis?

kongtiaowang:
@cmadjar  This is not required outside of Travis. It is only using on Travis. "test-install.loris.ca" is already on ubuntu 18.04, it is not required.

PR #6124, opened 2020-03-02
Update phan to latest version

johnsaigle:


modules/api/php/views/project/instruments.class.inc
```diff
@@ -69,7 +69,9 @@ class Instruments
                 'Subgroup'               => $instrument->getSubgroupname(),
                 'DoubleDataEntryEnabled' => $instrument->isDDE(),
             );
-            $instruments[$shortname] = $item;
+            if (!is_null($shortname)) {
```
driusan: I don't see any reason this should be possible. This is going to cause silent failures down the line.

php/libraries/CouchDB.class.inc
```diff
@@ -610,9 +610,9 @@ class CouchDB
      *
      * @param array $doc A PHP array encoding the JSON document to be created
      *
-     * @return string the JSON string returned by the CouchDB server
+     * @return ?string the JSON string returned by the CouchDB server
      */
-    function postDoc(array $doc): string
+    function postDoc(array $doc): ?string
```
driusan: This shouldn't be nullable. The CouchDB response should always be valid JSON.

php/libraries/CouchDB.class.inc
```diff
@@ -758,9 +758,9 @@ class CouchDB
      * Sends a POST request to update all the documents that have been queued
      * since the BulkTransaction was started.
      *
-     * @return string The string that was returned by CouchDB
+     * @return ?string The string that was returned by CouchDB
      */
-    function commitBulkTransaction(): string
+    function commitBulkTransaction(): ?string
```
driusan: This shouldn't be nullable either.

php/libraries/CouchDB.class.inc
johnsaigle: This issue comes from `_postRelativeURL` which has a nullable return type because `json_decode` itself can return null in some cases. 

Should we throw an Exception here (and below) if `$json` is null?

modules/api/php/views/project/instruments.class.inc
johnsaigle: https://github.com/aces/Loris/blob/4155ea37030eecc913701b1aebd81e0798916dbb/modules/api/php/models/projectinstrumentsrow.class.inc#L50-L58

driusan: The shortname is used as a key in the return response. This silently removes instruments from the response. I don't know why the projectinstrumentsrow does that, an instrument should always have a shortname.

php/libraries/CouchDB.class.inc
driusan: the response can be: 1. empty (in which the empty string should be returned) or 2. valid json. If json_decode fails, it should be an exceptional circumstance. We can probably get that behaviour by passing the JSON_THROW_ON_ERROR option to json_decode.

modules/api/php/views/project/instruments.class.inc
johnsaigle: Should I modify projectsinstrumentsrow then?

driusan: @xlecours do you know why getShortname in projectinstrumentsrow is nullable?

xlecours: @driusan my guess is that the Name column is nullable in the project table:
```
`Name` VARCHAR(255) NULL,
```

johnsaigle:
@driusan Can you re review?

modules/instrument_manager/php/instrument_manager.class.inc
```diff
@@ -161,7 +161,7 @@ class Instrument_Manager extends \NDB_Menu_Filter
             // If no adminUser is configured, automatic installation is not
             // possible, so this is the last step.
             return new \LORIS\Http\Response\JSON\OK(
-                self::UPLOAD_NO_INSTALL
+                json_encode(self::UPLOAD_NO_INSTALL)
```
driusan, line 164: This is incorrect. The constructor takes an array, not a string.

```
    public function __construct(
        array $body = [],
        array $headers = []
    ) {
```

PR #6125, opened 2020-03-02
[candidate_profile] Add Candidate Parameters to Candidate Info card

driusan:
This adds a new type of widget ('candidateinfo') which can be used by modules or projects to add information to the "Candidate Info" card on the `candidate_profile` page. The new type of widget can be used by a module (or project module) to add information that can be summarized as a scalar to the quick summary.

It uses the new type to add information from the `candidate_parameters` module that make sense as an overview for the candidate. The Participant Status, Caveat, and any 'Candidate Parameters' category variables from the `parameter_candidate` table are now summarized in the card.

Other information from the candidate_parameters module is complex enough to merit its own card (ie. Consent often contains multiple entries to be summarized for different types of consent, and shouldn't be in the top-level summary.)

### DIFF
--- a/modules/candidate_parameters/php/module.class.inc
+++ b/modules/candidate_parameters/php/module.class.inc
@@ -1,6 +1,5 @@
 <?php
 namespace LORIS\candidate_parameters;
-use LORIS\candidate_profile\CandidateWidget;
 use LORIS\candidate_profile\CandidateInfo;
 
 /**
@@ -155,7 +154,7 @@ class Module extends \Module
      * @param \Database  $db        A LORIS database connection
      * @param \Candidate $candidate The candidate being accessed
      *
-     * @return ?CandidateInfo
+     * @return CandidateInfo[]
      */
     private function _getCandidateParamTerms(
         \Database $db,
--- a/modules/candidate_parameters/php/module.class.inc
+++ b/modules/candidate_parameters/php/module.class.inc
@@ -1,6 +1,7 @@
 <?php
 namespace LORIS\candidate_parameters;
 use LORIS\candidate_profile\CandidateWidget;
+use LORIS\candidate_profile\CandidateInfo;
 
 /**
  * {@inheritDoc}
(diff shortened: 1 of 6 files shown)


modules/candidate_profile/php/candidateinfo.class.inc
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @param string $term  The term being summarized for the candidate
+     * @param string $value The value of the term
+     */
+    public function __construct(
+        string $term,
+        string $value
+    ) {
+        $this->term  = $term;
+        $this->value = $value;
```
johnsaigle: These aren't declared as class properties. I'm not sure if PHP will do this implicitly but they should probably be added and scoped at the top of the class.

modules/candidate_parameters/php/module.class.inc
```diff
@@ -33,4 +18,131 @@ class Module extends \Module
...
+            $candidate = $options['candidate'];
+            if ($candidate === null) {
+                return [];
+            }
+
+            // Participant Status
+            $entries = [
+                new CandidateInfo(
+                    "Participant Status",
+                    $candidate->getParticipantStatusDescription($db)
```
johnsaigle:
```suggestion
                    $candidate->getParticipantStatusDescription($db) ?? 'N/A'
```
Or something similar. Participant Status is mandatory if the value is change in the front-end but there are cases where it can be blank (at least in Raisinbread).
Screen Shot 2020-03-03 at 15 23 27

modules/candidate_profile/php/candidateinfo.class.inc
driusan: it's valid PHP but you're right, they should be declared anyways, eventually 'allow_missing_properties" should be set to false in phan..

modules/candidate_parameters/php/module.class.inc
```diff
@@ -68,7 +69,116 @@ class Module extends \Module
...
                 )
             ];
+        case 'candidateinfo':
+            // We add 3 extra types of cards to the candidate info card on
+            // the candidate dashboard based on the default LORIS candidate_parameter
+            // module:
+            //
+            // 1. The Participant Status
+            // 2. The "Candidate Parameters" category variables which are displayed
+            //    in the Candidate Information tab of the Candidate Parameters page
```
PapillonMcGill:
```suggestion
            // 1. The Participant Status.
            // 2. The "Candidate Parameters" category variables which are displayed
            //    in the Candidate Information tab of the Candidate Parameters page.
```
for consistency with the rest of comments

driusan:
@johnsaigle what browser/version are you using at what resolution/window size? I'm not able to reproduce

driusan:
nevermind, I reproduced.. it's when there's no extra candidate param terms that the `<hr />` term goes wonky.

johnsaigle:
Screen Shot 2020-03-09 at 12 32 32


I think this card should be hidden if no media is associated with the candidate.

driusan:
@johnsaigle that screenshot seems to be about a different PR than this one?

johnsaigle:
Yep it should be on the media one

johnsaigle:
Screen Shot 2020-03-09 at 12 37 09

Looks much better.
A potential enhancement might be subheadings. The top section could say something like "Basic candidate information" and the bottom section could say "Candidate Parameters" or somerthing mroe clear to indicate that the divider represents a difference between "core" candidate info and project-specific extensions to candidate info. (At leas,t that's my understanding of the difference between e.g. 'access profile' and 'candidate parameters'

Right now it's not clear why they're separated

johnsaigle:
oh this is merged now

driusan:
Any module that want can insert a scalar summary term into the bottom, not just candidate parameters.

PR #6128, opened 2020-03-03
[Vendor] Upgrade React in htdocs and ensure it matches package.json

driusan:
This upgrades the React library that comes with LORIS for the release.

This allows us to use React hooks with the 23.0 release. React is
upgraded in place as a stop-gap until #6097 updates it automatically
for the next release.


driusan:
I don't think it's really possible for anyone to review this and it's passing travis, so I'm going to merge it on the assumption that it's better to be in once people start testing than to make a library change afterwards.

PR #6131, opened 2020-03-04
[candidate_profile] Add Unresolved Conflicts widget

driusan:
This adds an "Unresolved Conflicts" widget to the candidate dashboard if there are any unresolved conflicts. The conflicts are visualized as a pie chart broken down by instrument / visit. Clicking on any slice of the pie or the legend takes you to the conflict resolver with the filters preset* for the candidate/instrument/visit.

I'm not crazy about the way data is displayed, but a pie chart is the best way I could think of to display the data in a fixed amount of space when the number of instruments and visits may vary wildly between projects and candidates. I'm open to other ways to summarize the data.

* There's currently a bug in the conflict_resolver where the filter doesn't get applied until you change some other filter.


driusan:
The bug with the links is fixed by #6132.

johnsaigle:
Screen Shot 2020-03-09 at 11 27 45


Works for me. I agree this may not be the best way of presenting data - it's not clear what this one solid block means either. For now I think it's satisfactory if this gets explained in the module help text.

jesscall:
Tested this on CCNA and I have a few lil comments.

I know we've already acknowledged that a pie chart is not the best way to summarize the data, but maybe these images will help with prioritizing a different display. 

So if the candidate only has conflicts on one instrument, all is well in the world:

Screenshot from 2020-03-10 14-27-54

However, if the candidate has conflicts in many instruments (which is normally the case), it's really cluttered and you can't even see the chart...

Screenshot from 2020-03-10 14-30-07

Only way to see the pie chart is to zoom out... a lot.... as in 33%

Screenshot from 2020-03-10 14-32-31


Apart from the issues I have with how this is displayed, the links re-direct as expected. I did however notice that even though the filters are filled in properly in conflict resolver, the data table doesn't actually get filtered (see image). This obviously isn't in the scope of this PR, just thought it was good to mention it. Hopefully this is fixed by reactification of conflict resolver.

Screenshot from 2020-03-10 14-30-43

TL;DR

Things that work:
- In the case where candidate has no conflicts, the pane is not displayed. 
- Links redirect correctly.

Things that can be improved:
- Manner in which data is displayed.
- The breadcrumb brings you back to timepoint list, which I think should not be the case (?)

driusan:
thanks for the screenshots @jesscall. Do you have any suggestions for how you'd like the data visualized when there's that many conflict types? (How many is that?) We can have a cut off where if there's more than x rows returned by the query the visualization changes to a better mechanism if you have a suggestion.

(The links not working was fixed by #6132 but this branch hasn't been rebased since that was merged.)

(PS. you might want to do something about the fact that "candidate has conflicts in many instruments" is the normal case for you..)

driusan:
What about something like this: https://c3js.org/samples/chart_bar_stacked.html with the visits going across and the stacks being instrument breakdowns? Or only breaking it down by visit instead of visit/instrument?

jesscall:
I think what's taking up space is the legend with all the instruments listed. The stacked bar chart looks nice, but i feel like we might run into similar issues. Maybe it's a CSS thing where the chart should be anchored to the top of the card and if there are many instruments, the card just resizes

johnsaigle:
Is it possible to remove the legend? The pie chart on the dashboard seems to display the Site on hover, maybe that's enough?

driusan:
@jesscall I've redone the widget as a normal bar graph. The widget width is now variable, so that it spans more columns if there are a lot of instruments. Can you retest/review with CCNA data and let me know how it works?

jesscall:
@driusan It looks much better! Below are some screenshots.

Screen Shot 2020-03-24 at 2 17 01 PM


Screen Shot 2020-03-24 at 2 20 26 PM

Hovering over a visit also gives you a list of instruments with conflicts & screen is scrollable when there is overflow (like the case below):
Screen Shot 2020-03-24 at 2 20 51 PM

driusan:
@jesscall Can you do me a favour and try changing

```
            $width = 1;
            if (count($results) > 10) {
                $width = 2;
            }
```

to

```
            $width = 1;
            if (count($results) > 20) {
                $width = 3;
            } else if (count($results) > 10) {
                $width = 2;
            }
```

? I want to see how much that affects the second screenshot, it still looks a little crowded to me.

jesscall:
Befor change:

Screen Shot 2020-03-25 at 1 44 24 PM


After change:

Screen Shot 2020-03-25 at 1 43 53 PM

driusan:
@jesscall thanks.. that doesn't look like a very significant change, so I'll keep it as is.

PR #5256, opened 2019-10-01
[Media] Rename date_uploaded column to last_modified

johnsaigle:
## Brief summary of changes

The `Date Uploaded` column is not named accurately. It is a timestamp value that updates to the current timestamp upon edit. 

Given this behaviour, it should be called "Last Modified". This PR does that.

#### Testing instructions (if applicable)

1. Check out this pull request and run the patch
2. Run `make dev`
3. Go to the media module and verify that the column `Date Uploaded` now says `Last Modified`

#### Links to related tickets (GitHub, Redmine, ...)

* Resolves #5078

PR #6133, opened 2020-03-04
[candidate_profile] Add media widget

driusan:
This adds a widget to the candidate dashboard to display
any media associated with the candidate from the media
module. The visualization of how the data is displayed
is the same as the "Document Repository" widget on the
main LORIS dashboard.

### DIFF
--- a/modules/media/jsx/CandidateMediaWidget.js
+++ b/modules/media/jsx/CandidateMediaWidget.js
@@ -16,7 +16,7 @@ function CandidateMediaWidget(props) {
                     + '/media/ajax/FileDownload.php?File='
                     + encodeURIComponent(file.Filename)}>
                 <span className="pull-right text-muted small">
-                    Uploaded: {file.UploadDate}
+                    Last modified : {file.LastModified}
                 </span>
                 <br />
                 {file.Filename}
--- a/modules/media/jsx/CandidateMediaWidget.js
+++ b/modules/media/jsx/CandidateMediaWidget.js
@@ -1,0 +1,29 @@
+/**
+ * React component for a widget on the candidate dashboard
+ * displaying the media associated with the candidate.
+ *
+ * @param {array} props - The React props
+ *
+ * @return {object} - The component
+ */
+function CandidateMediaWidget(props) {
+    let files = [];
+    for (let i = 0; i < props.Files.length; i++) {
+        const file = props.Files[i];
+        files.push(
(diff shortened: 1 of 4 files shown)


modules/media/php/module.class.inc
```diff
@@ -58,4 +41,47 @@ class Module extends \Module
...
+            $factory   = \NDB_Factory::singleton();
+            $baseurl   = $factory->settings()->getBaseURL();
+            $candidate = $options['candidate'];
+            if ($candidate === null) {
+                return [];
+            }
+            $db = $factory->database();
+
+            $media = $db->pselect(
+                "SELECT file_name as Filename, date_uploaded as UploadDate
```
johnsaigle:
```suggestion
                "SELECT file_name as Filename, last_modified as LastModified
```
This was changed in https://github.com/aces/Loris/pull/5256

driusan: done in latest commit, the jsx had to be modified too.

johnsaigle:
Screen Shot 2020-03-09 at 12 32 32

When there is no media, the card should be hidden

driusan:
@johnsaigle done in latest commit

modules/media/jsx/CandidateMediaWidget.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+function CandidateMediaWidget(props) {
+    let files = [];
+    for (let i = 0; i < props.Files.length; i++) {
+        const file = props.Files[i];
+        files.push(
+            <a className="list-group-item" key={i}
+                href={props.BaseURL
+                    + '/media/ajax/FileDownload.php?File='
+                    + encodeURIComponent(file.Filename)}>
```
jesscall: you should get the config media path and append it here

jesscall: otherwise i am getting 404 error

driusan: Are you sure? This is the same way it's done in mediaIndex.js:

```
        const downloadURL = loris.BaseURL + '/media/ajax/FileDownload.php?File=' +
          encodeURIComponent(row['File Name']);
```

jesscall: you're right. might have something to do with how i upgraded. ill investigate and let you know if it's actually an issue with the PR

PR #6137, opened 2020-03-04
[media] File download should retrieve decoded path

### LINKED ISSUE #6136: [media] Downloading file with special characters in name leads to error: Failed - no file
**Describe the bug**
There is an error when downloading files from media module that contain special characters in its name (ex: '&').  Error message that appears says "Failed - No File". This is likely occurring because SQL changes such characters to their html encoding. 

**To Reproduce**
1. Go to media module.
2. If no file with special characters (&, >, <, etc) is available, upload one yourself.
3. Try downloading that file now.
4. See error - Failed, no file.

File should download normally.

- OS: Ubuntu
- Browser: Chrome

- LORIS Version: 21.0 and (i suspect) onward
- Linux distribution and Version: Ubuntu 16.04
- MySQL/MariaDB Version: MySQL 5.7

jesscall:
## Brief summary of changes
This PR is related to issue #6136. This PR aims to fix a bug that was experienced by a CCNA user when trying to download a file that had a '&' in the name. SQL encodes strings containing special chars and thus we need to use the decoded file names in the path when downloading files from media module. 

Issuing this fix to the 21.0-release branch as it is affecting our users and current instance on CCNA. 

1. Upload a file to media module that contains special chars (&, >, <, etc)
2. Try downloading the file and ensure that you do not receive an error. 

* Resolves #6136

### DIFF
--- a/modules/media/ajax/FileDownload.php
+++ b/modules/media/ajax/FileDownload.php
@@ -24,7 +24,7 @@ if (!$user->hasPermission('media_write')) {
 
 // Make sure that the user isn't trying to break out of the $path
 // by using a relative filename.
-$file     = basename($_GET['File']);
+$file     = html_entity_decode(basename($_GET['File']));
 $config   =& NDB_Config::singleton();
 $path     = $config->getSetting('mediaPath');
 $filePath = $path . $file;


driusan:
21 isn't supported anymore and this doesn't look like an issue that would prevent updating, can you send this to 22 or 23?

ridz1208:
@driusan what if it's needed for 21 ?

johnsaigle:
It would be nice to get this into 23 @ridz1208 @driusan

ridz1208:
@johnsaigle I can put it in 22. can you just review it please ?

PR #6139, opened 2020-03-04
[candidate_profile] Add widget for open candidate issues to the candidate_profile

driusan:
This adds a widget for any unclosed issues associated with a candidate to the
candidate dashboard. If there are no unresolved issues, the card is not displayed.

Issues are displayed as a simple list with links to the issue.

### DIFF
--- a/modules/issue_tracker/README.md
+++ b/modules/issue_tracker/README.md
@@ -20,4 +20,4 @@ Future development includes the ability to edit one’s own comments after submi
 
 ## Interactions with LORIS
 
-The issue tracker registers widgets on the `dashboard` and `candidate_profile` dashboard
+The issue tracker registers widgets on the `dashboard` and `candidate_profile` dashboard.
--- a/modules/issue_tracker/README.md
+++ b/modules/issue_tracker/README.md
@@ -17,3 +17,7 @@ Most of the permissions are controlled in `IssueForm.js`, dependent on values re
 
 ## Future Development
 Future development includes the ability to edit one’s own comments after submission, a pop-up or sliding view within other modules for easy issue reporting, further integration with the dashboard and with statistics, refinement of the form UI, a tabular sortable view of comments, the ability to click watching within the filter-form table and the inclusion of default users to notify in the config settings.
+
+## Interactions with LORIS
+
+The issue tracker registers widgets on the `dashboard` and `candidate_profile` dashboard
--- a/modules/issue_tracker/php/module.class.inc
+++ b/modules/issue_tracker/php/module.class.inc
@@ -110,8 +110,8 @@ class Module extends \Module
                     LEFT JOIN issues_comments ic ON (i.issueID=ic.issueID)
                  WHERE status != 'closed' AND CandID=:cid
                  GROUP BY i.issueID, i.Title",
-                     [ 'cid' => $candidate->getCandID() ],
-                 );
+                [ 'cid' => $candidate->getCandID() ],
+            );
             if (empty($issues)) {
                 return [];
             }
--- a/modules/issue_tracker/php/module.class.inc
+++ b/modules/issue_tracker/php/module.class.inc
@@ -112,6 +112,21 @@ class Module extends \Module
                  GROUP BY i.issueID, i.Title",
                      [ 'cid' => $candidate->getCandID() ],
                  );
+            if (empty($issues)) {
+                return [];
+            }
+            return [
+                new \LORIS\candidate_profile\CandidateWidget(
+                    "Open Issues",
+                    $baseURL . "/issue_tracker/js/CandidateIssuesWidget.js",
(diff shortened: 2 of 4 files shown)


modules/issue_tracker/php/module.class.inc
```diff
@@ -93,6 +93,41 @@ class Module extends \Module
                     "issue_tracker"
                 )
             ];
+        case 'candidate':
+            $factory = \NDB_Factory::singleton();
+            $DB      = $factory->database();
+            $baseURL = $factory->settings()->getBaseURL();
```
lingma: These 3 lines are repeated in `usertasks` and `candidate`, whether it is better to first check if the `$type` is in 'usertasks` and `candidate`, if yes, then get the `$factory`, etc. to avoid repeat of the same code?

modules/issue_tracker/README.md
```diff
@@ -17,3 +17,7 @@ Most of the permissions are controlled in `IssueForm.js`, dependent on values re
 
 ## Future Development
 Future development includes the ability to edit one’s own comments after submission, a pop-up or sliding view within other modules for easy issue reporting, further integration with the dashboard and with statistics, refinement of the form UI, a tabular sortable view of comments, the ability to click watching within the filter-form table and the inclusion of default users to notify in the config settings.
+
+## Interactions with LORIS
+
+The issue tracker registers widgets on the `dashboard` and `candidate_profile` dashboard
```
johnsaigle:
```suggestion
The issue tracker registers widgets on the `dashboard` and `candidate_profile` dashboard.
```

modules/issue_tracker/php/module.class.inc
driusan: done in the last commit

modules/issue_tracker/php/module.class.inc
```diff
@@ -74,11 +74,16 @@ class Module extends \Module
      */
     public function getWidgets(string $type, \User $user, array $options) : array
     {
+        if (!in_array($type, ['usertasks', 'candidate'], true)) {
+            return [];
```
johnsaigle: Would this be a Bad Request?

driusan: No, there are valid widget types not implemented by this module (`dashboard`, `candidateinfo`..).

This check is just to avoid calling the factory on things that aren't implemented by the switch statement after pulling is out of the individual `case`s.

johnsaigle: In the context of this module though, wouldn't `dashboard` be a Bad Request? Even if it's valid in other child classes.

driusan: No. You are allowed to visit the dashboard page even if you have the `candidate_profile` module installed.

johnsaigle: I'm not really sure what you have in mind - maybe I don't understand the callstack of how getWidgets works.

Anyway this is a minor point so I'll resolve this

driusan: `getWidgets` is called on all active modules with a `$type` for any module that wants to get all the widgets of that type. If the module doesn't know anything about that type, it doesn't mean it was a bad call, it means that it has no widgets to return. For instance, the dashboard will call `getWidgets('dashboard', ...)` on all modules. If the issue_tracker gets that call and doesn't have any dashboard widgets to return (either because it doesn't know about the dashboard or doesn't care) that doesn't make it a bad call or a bad request from a client, it makes the proper thing to do to return an empty array of no widgets.

lingma: I suppose that the module is only a API data provider, so Bad Request is not necessary, maybe we can return empty or 501 (Not Implemented) instead of "Bad Request"

driusan: There is nothing wrong with the request.

driusan:
@lingma can you re-review? I'm pretty sure all your changes are incorporated.

lingma:
Yes, I checked yesterday. I thought once you have done then it is OK I don't need to do anything. Currently this branch has conflicts

webpack.config.js
```diff
@@ -139,8 +139,9 @@ const config = [
         module: mod,
     },
     // Modules
-    lorisModule('media', ['CandidateMediaWidget', 'mediaIndex']),
     lorisModule('issue_tracker', ['issueTrackerIndex', 'index']),
+    lorisModule('media', ['CandidateMediaWidget', 'mediaIndex']),
+    lorisModule('issue_tracker', ['issueTrackerIndex', 'index', 'CandidateIssuesWidget']),
```
lingma: Why are there two 'issue_tracker's?

driusan: poorly resolved rebase conflict.. should be fixed.

PR #6140, opened 2020-03-04
Instruments variable incorrectly defined

ridz1208:
## Brief summary of changes
This was showing up as an error on my IDE, the variable is defined as private in the hchild class while the parent class refers to it.

change from `_select...` to `select...` because static checks enforce the ugly `_` before private variable rule

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -183,7 +183,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     public $preview = false;
 
     /**
-     * Array conaining all multiselct elements in an instrument.
+     * Array containing all multiselct elements in an instrument.
      */
     protected $selectMultipleElements;
 
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -183,7 +183,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     public $preview = false;
 
     /**
-     * Array containing all multiselct elements in an instrument.
+     * Array containing all multiselect elements in an instrument.
      */
     protected $selectMultipleElements;
 
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -183,6 +183,11 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     public $preview = false;
 
     /**
+     * Array used by LINST instruments
+     */
+    protected $selectMultipleElements;
+
+    /**
      * Factory generates a new instrument instance of type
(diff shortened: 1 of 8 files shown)


php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -182,6 +182,11 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     public $preview = false;
 
+    /**
+     * Array used by LINST instruments
```
driusan: Can you make the comment more descriptive? Used for what?

ridz1208:
@lingma as per the description yes it is mandatory by our static analyses tools

@driusan I can but do you know what it's used for ? Or do I dig deeper ?

lingma:
> @lingma as per the description yes it is mandatory by our static analyses tools

Then you need to modify several others files including at least 4 instruments, some wifi pages.

ridz1208:
@lingma please request changes and specify which classes need to be modified. Thanks

driusan:
It looks like it's a list of multi-select elements kept so that they can be converted to/from `{@}` delimited on save.

johnsaigle:
I agee with Ling that the other files should be updated too. At least the Raisinbread instruments.

ridz1208:
@lingma @johnsaigle does that cover everything ?

driusan:
@ridz1208 the comment update is still missing

lingma:
> @ridz1208 the comment update is still missing

Right, the `//` is missing. In fact `$this->selectMultipleElements = array();` or 
`//$this->selectMultipleElements = array();` should be the same.

driusan:
No, I mean the update to the comment that I requested be updated 3 days ago because it's not very descriptive still isn't very descriptive.

ridz1208:
@lingma i removed that comment on purpose, it's better to just declare the empty array

@driusan done

driusan:
@ridz1208 Makes sense to me, look better now.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -182,6 +182,11 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     public $preview = false;
 
+    /**
+     * Array conaining all multiselct elements in an instrument.
```
lingma: containing, not "conaining"

ridz1208:
```suggestion
     * Array containing all multiselct elements in an instrument.
```

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -182,6 +182,11 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     public $preview = false;
 
+    /**
+     * Array containing all multiselct elements in an instrument.
```
```suggestion
     * Array containing all multiselect elements in an instrument.
```

PR #6142, opened 2020-03-09
[SQL/imaging] Adding SQL code for the new HRRT PET insertion scripts

cmadjar:
## Brief summary of changes

This PR updates the SQL schema to include the new PET HRRT archive tables and their link to the already existing imaging tables.

1. Ensure the SQL patches and schema run properly

* This is linked to LORIS-MRI PR https://github.com/aces/Loris-MRI/pull/524  

* Github issue: https://github.com/aces/Loris-MRI/issues/480

### DIFF
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -419,6 +419,42 @@ CREATE TABLE `tarchive_files` (
   CONSTRAINT `tarchive_files_TarchiveSeriesID_fk` FOREIGN KEY (`TarchiveSeriesID`) REFERENCES `tarchive_series` (`TarchiveSeriesID`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
+-- ********************************
+-- HRRT PET archive tables
+-- ********************************
+CREATE TABLE `hrrt_archive` (
+  `HrrtArchiveID`     INT(11)          NOT NULL AUTO_INCREMENT,
+  `SessionID`         INT(10) unsigned          DEFAULT NULL,
+  `EcatFileCount`     INT(11)          NOT NULL DEFAULT '0',
+  `NonEcatFileCount`  INT(11)          NOT NULL DEFAULT '0',
+  `DateAcquired`      DATE                      DEFAULT NULL,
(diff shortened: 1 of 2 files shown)


SQL/0000-00-00-schema.sql
```diff
@@ -419,6 +419,42 @@ CREATE TABLE `tarchive_files` (
...
+-- HRRT PET archive tables
+-- ********************************
+CREATE TABLE `hrrt_archive` (
+  `HrrtArchiveID`     INT(11)          NOT NULL AUTO_INCREMENT,
+  `SessionID`         INT(10) unsigned          DEFAULT NULL,
+  `EcatFileCount`     INT(11)          NOT NULL DEFAULT '0',
+  `NonEcatFileCount`  INT(11)          NOT NULL DEFAULT '0',
+  `DateAcquired`      DATE                      DEFAULT NULL,
+  `DateArchived`      DATETIME                  DEFAULT NULL,
+  `PatientName`       VARCHAR(50)      NOT NULL DEFAULT '',
```
PapillonMcGill: Will this create a problem about ethics?

SQL/0000-00-00-schema.sql
```diff
@@ -419,6 +419,42 @@ CREATE TABLE `tarchive_files` (
...
+  `CenterName`        VARCHAR(50)      NOT NULL DEFAULT '',
```
PapillonMcGill: Could it be the CenterID of PSC table?

SQL/0000-00-00-schema.sql
cmadjar: This will have the exact same behaviour as the PatientName in the tarchive table meaning that the PatientName will be a code with PSCID_DCCID_VisitLabel and not an actual patient name.

SQL/0000-00-00-schema.sql
cmadjar: The CenterName field in here refers to the CenterName field read from the HRRT dataset and not to any of the center from the psc table. In the case of the MNI, it will be `BIC-MNI_MCGILL`. Same behaviour as the `tarchive` table.

PR #6143, opened 2020-03-09
[Docs] Add documentation for the getWidgets function

driusan:
This adds developer documentation for the existing types of widgets that
can be added to LORIS modules, and what their expectations are of
the `getWidgets` function.

### DIFF
--- a/docs/ModuleWidgets.md
+++ b/docs/ModuleWidgets.md
@@ -7,7 +7,7 @@ extend the capabilities of other modules within a new module.
 
 Widgets are can be used by implementing the `getWidgets` method
 on a `Module` module descriptor class. The first parameter of
-`getWidgets` is a type string, allowing modules to define their
+`getWidgets` is a string representing the widget type, allowing modules to define their
 own widget types in a loosely coupled manner. However since the
 interface is abstract enough to be flexible there must be an implied
 contract between what options and their returned values that are not
--- a/docs/ModuleWidgets.md
+++ b/docs/ModuleWidgets.md
@@ -1,0 +1,39 @@
+## Module Widgets
+
+Since LORIS 23.0, LORIS incorporates a concept of "widgets" which
+modules can use to dynamically display graphical widgets in other
+modules when installed and activated. This is a powerful way to
+extend the capabilities of other modules within a new module.
+
+Widgets are can be used by implementing the `getWidgets` method
+on a `Module` module descriptor class. The first parameter of
+`getWidgets` is a type string, allowing modules to define their
+own widget types in a loosely coupled manner. However since the
+interface is abstract enough to be flexible there must be an implied
+contract between what options and their returned values that are not
+enforced by PHP's typing system.
+
+Widget types currently implemented by LORIS and its default modules are
+
+# Dashboard
+
+- The 'dashboard' type displays a widget on the dashboard (default landing
+  page) of LORIS. No options are passed and the return of getWidgets is
+  expected to be an array of `\LORIS\dashboard\Widget` elements.
+- The 'usertasks' type widget gets added to the LORIS dashboard in the 
+  'My Tasks' panel to display outstanding user tasks. No options are passed
+  and the return value is expected to be an array of `\LORIS\dashboard\TaskWidget`
+  elements (usually created by `\LORIS\dashboard\TaskQueryWidget` which generates
+  the widget from an SQL query.
+
+# Candidate Profile
+
+- The 'candidate' type displays a widget on the `candidate_profile` page. The
+  options passed include a 'candidate' key which is a `\Candidate` object.
+  The return value is expected to be an array of `LORIS\candidate_profile\CandidateWidget`;
+  which each get displayed in their own card of the page.
+- The 'candidateinfo' type displays extra information in the 'Candidate Info'
+  card of the profile. The options passed are the same as for the 'candidate'.
+  The return value is expected to be an array of `LORIS\candidate_profile\CandidateInfo`
+  elements consisting of terms to render into the existing card and the value of
+  the term.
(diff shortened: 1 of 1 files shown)


docs/ModuleWidgets.md
```diff
@@ -0,0 +1,39 @@
+## Module Widgets
+
+Since LORIS 23.0, LORIS incorporates a concept of "widgets" which
```
johnsaigle:
```suggestion
LORIS incorporates a concept of "widgets" which
```

I don't think it's important to include when the feature was added.

docs/ModuleWidgets.md
```diff
@@ -0,0 +1,39 @@
...
+modules can use to dynamically display graphical widgets in other
+modules when installed and activated. This is a powerful way to
+extend the capabilities of other modules within a new module.
```
johnsaigle:
```suggestion
modules when installed and activated.
```

Not a big deal. I think it can be removed though. Reads more like a sales pitch than as providing useful info to the developer reading it

docs/ModuleWidgets.md
```diff
@@ -0,0 +1,39 @@
...
+
+Widgets are can be used by implementing the `getWidgets` method
+on a `Module` module descriptor class. The first parameter of
+`getWidgets` is a type string, allowing modules to define their
```
johnsaigle: I'm not sure I understand the phrase "type string"

docs/ModuleWidgets.md
```diff
@@ -0,0 +1,39 @@
...
+own widget types in a loosely coupled manner. However since the
+interface is abstract enough to be flexible there must be an implied
+contract between what options and their returned values that are not
+enforced by PHP's typing system.
```
johnsaigle: The wording of this section is a little confusing and I think the grammar may be off. 

From my understanding of this passage, I would conclude that types are not enforced and therefore the widget must be designed with caution because there is not a lot of validation being done. Is that right?

docs/ModuleWidgets.md
```diff
@@ -0,0 +1,39 @@
...
+
+Widget types currently implemented by LORIS and its default modules are
```
johnsaigle: Incomplete sentence? If this is meant to lead into the list it should probably end with a colon or should be removed.

It is worth considering whether the following info should be moved into the modules' respective READMEs instead of this document. As more widgets are added, I think it would be better and more realistic that devs update a module's README with widget information rather than come back to this document

driusan:
This has nothing to do with any wiki, it's new developer documentation.

docs/ModuleWidgets.md
```diff
@@ -0,0 +1,39 @@
...
+# Dashboard
+
+- The 'dashboard' type displays a widget on the dashboard (default landing
+  page) of LORIS. No options are passed and the return of getWidgets is
+  expected to be an array of `\LORIS\dashboard\Widget` elements.
+- The 'usertasks' type widget gets added to the LORIS dashboard in the 
+  'My Tasks' panel to display outstanding user tasks. No options are passed
+  and the return value is expected to be an array of `\LORIS\dashboard\TaskWidget`
+  elements (usually created by `\LORIS\dashboard\TaskQueryWidget` which generates
+  the widget from an SQL query.
```
johnsaigle:
```suggestion
  the widget from an SQL query).
```

docs/ModuleWidgets.md
driusan: The parameter named `$type` which is a string.

johnsaigle:
> This has nothing to do with any wiki, it's new developer documentation.

We have developer documentation is going into the wiki, such as the 'Making a module' guide, automated testing information, etc. etc.

docs/ModuleWidgets.md
driusan: There is an '$options' associative array parameter. The PHP type system won't enforce any of the types for key/values passed in that array. This is intentional, since the keys and their value types vary based on the widget type. The document is intended to document the types of widgets that can be added, what `$options` they come with, and what type of widget they expect to be returned in one centralized place that can be referenced when writing a new module.

I agree the wording is off, I was in a rush to write something before I forget what types of widgets we already have. I'm happy to accept suggestions for how to clarify.

docs/ModuleWidgets.md
driusan: The document is intended to be a centralized place to figure out what you can do using the widget system when writing a new module.

I added the missing colon.

johnsaigle: In that case maybe this should be added to #6048 instead? Widgets could be subsection fo that document.

docs/ModuleWidgets.md
johnsaigle:
```suggestion
`getWidgets` is a string representing the Widget type, allowing modules to define their
```

docs/ModuleWidgets.md
johnsaigle:
```suggestion
own widget types in a loosely coupled manner. The second parameter `$options` is 
an associative array. The key and value types vary based on the Widget and are therefore not
enforced by PHP. This allows flexibility when designing widgets. As a result, developers should be
careful to manually validate expected values and handle errors appropriately.
```

PapillonMcGill: @driusan What about:
"Widgets can be used by implementing the `getWidgets` method
on a `Module` module descriptor class. The first parameter of
`getWidgets` is a string that define the type of value(s) the module will pass to widget. Using this type, the widget will be able to format the output accordingly." ?

johnsaigle: made a suggestion below

docs/ModuleWidgets.md
driusan: I have nothing against incorporating it into a `Creating a Module` guide in the future, but I think this should be documented in this release, since this is the release where widgets were added.

docs/ModuleWidgets.md
driusan: I think it's important information

PR #6144, opened 2020-03-09
Add .gitignorefile and remove unnecessary transpiling of files

ridz1208:
## Brief summary of changes
After `make` command is run, some files were showing up in the untracked files category, we can ignore these JS files because they are compiled.

Also, some imported JSX files were being compiled unecessarily, I removed them from the webpack file

### DIFF
--- a/modules/electrophysiology_browser/.gitignore
+++ b/modules/electrophysiology_browser/.gitignore
@@ -1,2 +1,1 @@
-js/electrophysiologySessionView.js
-js/electrophysiologyBrowserIndex.js
+js/*
--- a/modules/electrophysiology_browser/.gitignore
+++ b/modules/electrophysiology_browser/.gitignore
@@ -1,0 +1,2 @@
+js/electrophysiologySessionView.js
+js/electrophysiologyBrowserIndex.js
--- a/modules/data_release/.gitignore
+++ b/modules/data_release/.gitignore
@@ -1,2 +1,1 @@
-user_uploads/*
-js/dataReleaseIndex.js
+js/*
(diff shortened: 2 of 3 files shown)


modules/electrophysiology_browser/.gitignore
```diff
@@ -0,0 +1,2 @@
+js/electrophysiologySessionView.js
```
driusan: don't we usually just include `js/*` in the module `.gitignore`? Are there other non-webpack generated files for the EEG browser?

ridz1208: I was debating that... there are none no, but if one gets added later it will be automatically ignored. if we selectively ignore them however we dont run the risk of automatically ignoring new ones

ridz1208: that beignsaid... I REALLY DONT CARE

driusan: We shouldn't be putting anything directly in the `js` directory, there's only some modules that have an explicit file list there because they have old files that were grandfathered in and can't use the wildcard.

maltheism: @driusan @ridz1208  we could move those old files to the jsx directory as a solution.

driusan: I don't think there are any old files in either of these modules.

maltheism: We should then just have `js/*` in the module like @driusan wrote.

ridz1208: Can anyone make the changes directly on the pr ? I'm in meetings for the afternoon

maltheism: @ridz1208 added changes.

ridz1208: @maltheism thanks

PR #6146, opened 2020-03-09
[Docs] Remove React.README.md

driusan:
This removes an unmaintained file describing how React was used in LORIS
in 2016. The changes described and best-practices described are now either
now standard parts of how LORIS is developed, documented elsewhere, or
opinionated pieces of documentation that LORIS doesn't require.

This document has not been touched in 3 years.

## Brief summary of changes


driusan:
Outdated documentation is worse than outdated code. We delete outdated code, why should outdated documentation stay around forever? Are you saying as soon as some document is added it needs to stick around forever and carry and infinite maintenance burden?

If that's the case, the burden for "does this document's value outweigh it's maintenance burden?" will need to be adjusted for any documentation added to LORIS.

johnsaigle:
Not every document needs to stay forever and I absolutely think some documentation is not worth it's burden which is why I've been removing big sections or whole documents from the deprecated wiki. 

I think it would be better to add a ticket about reviewing/updating the React documentation rather than deleting it. We should require that we document our coding standards because it helps have something for people to refer to when trying to contribute to the codebase. 

It's better to collectively amend a documented procedure for creating good code rather than rely solely on code review. Without public guidelines, some recommendations in code review can appear to come across as someone's personal preference or as personal criticism in the form of nitpicking. 

As a personal example, I don't really know what's going on with React - however if I got assigned to do a React task today, this document is the first place I'd go to. If we didn't have one, I'd have to stumble around asking people how we do React in LORIS... and when I was finished I would create this document. I think it's essential.

Could you point out specific sections that are outdated or redundant? I'm happy to do my share of the maintenance burden here but this is unfortunately a blind spot for me.

driusan:
There are no parts of this document that are not outdated or better documented elsewhere.

It consists of an Overview which doesn't say much of anything but tries to be witty, Coding Guidelines that no one follows or enforces, a file structure that's described elsewhere, a (likely outdated) description of how to install the tools that LORIS requires to be installed whether or not you read this document (and tells you to install them with `curl $URL | sudo`), outdated instructions on how to install the javascript dependencies (including an inexplicable note telling you to chown a directory), outdated instructions on how to update the webpack config if you add a new file. It then explains that there are 2 different 1-line commands which are aliases for 2 other 1-line commands should you choose to use them.

johnsaigle:
Could we compromise by merging this PR but at the same time creating a ticket to write better ReactJS documentation?

docs/React.README.md
```diff
@@ -1,146 +0,0 @@
...
-
-In many cases, React components are designed for a specific LORIS module. In such cases the `JSX` file should be placed in the
-`module/$DESIRED_MODULE/jsx/` and its compiled file in the `module/$DESIRED_MODULE/js/` directory.
-
-
->**Note**: When creating a new JSX file **always** use the `.js` extension or the continuous compilation will not work as expected. The file name of the `JSX` and its compiled file should have the same name.
-
----
-
-## Set up
```
ridz1208, line 37: @driusan i still use this section everytime I have to setup a new VM or instruct someone to install LORIS. If we can move this the some sort of FAQ somewhere else I'll approve the changes

driusan: Which part is missing? The CentOS part is already in `docs/wiki/00_SERVER_INSTALL_AND_CONFIGURATION/01_LORIS_Install/CentOS/README.md` and the Ubuntu part should be in the Ubuntu equivalent, but I haven't verified

driusan: (Also the instructions here are wrong and install an outdated version, at least compared to the RTD documentation version..)

ridz1208: you can merge it then I guess we'll revise the install instructions if people complain

driusan: If the install instructions are missing something we should be updating the install instructions anyways.. people (other than @ridz1208, apparently..) wouldn't think to look into `React.README.md` for LORIS install dependencies.

ridz1208: I only look in there cause Alex Ilea sent that file to me the first time lol.

However if we dont instruct users on how to install dependencies like git or php... maybe we dont need it for node and NPM either

PR #5739, opened 2019-11-22
[API] Add electrophysiology endpoints

cmadjar:
## Brief summary of changes

This adds endpoints to query electrophysiology datasets via the API.

#### Testing instructions (if applicable)

1. Follow the updated documentation of version v0.0.3-dev of the API

#### Links to related tickets (GitHub, Redmine, ...)

* #5235

PR #6148, opened 2020-03-09
[Core] CandID implements JsonSerializable

xlecours:
## Brief summary of changes

Since CandID are proper objects and no longer strings or integer, json_encode was returning those objects as `{}`

This add the JsonSerialize function to the class to specify how json_encode should treat thos objects. In this case, as a string which value is the candid.  

 
Query any API endpoint which returns a CandID (most do in the Meta fields)

This will also fix the new endpoints return value in https://github.com/aces/Loris/pull/5739

### DIFF
--- a/modules/api/php/views/candidate.class.inc
+++ b/modules/api/php/views/candidate.class.inc
@@ -44,7 +44,7 @@ class Candidate
     public function toArray(): array
     {
         $meta = array(
-            'CandID'  => (string) $this->_candidate->getCandID(),
+            'CandID'  => $this->_candidate->getCandID(),
             'Project' => $this->_candidate->getProjectTitle(),
             'PSCID'   => $this->_candidate->getPSCID(),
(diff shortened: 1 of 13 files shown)


cmadjar:
@xlecours @driusan Should this go to `23.0-release` branch since it fixes a bug?

xlecours:
@cmadjar I don't think so, the existing endpoints were casting the CandID as string.

driusan:
If I understand the CandID object correctly this will serialize it as `"123456"` and not `123456`. Shouldn't it be an int?

xlecours:
> If I understand the CandID object correctly this will serialize it as `"123456"` and not `123456`. Shouldn't it be an int?

No. The API documentation shows them as string when specified. A CandID is a numeric string. We never do arithmetical operation on CandID. There is non need to sort them as number instead of strings. The only operators used on them is the equality. Having them as integer would just be error prone when zeros are truncated `000001 != 1`.

PR #6151, opened 2020-03-09
[Docs] Move API docs into the API module

driusan:
This moves the documentation for the REST API into the
API module. It existed under docs previously because
it was written before the module existed.

Symlinks are used to keep the documentation rendered
under ReadTheDocs since it's useful to include a reference
for users looking to use the LORIS API.

If we're lucky, I didn't break anything.


johnsaigle:
> 1) MriViolationsTestIntegrationTest::testNotResolvedSaveButton

Transitory failure? We should look out for this during testing and maybe disable this test if it's not working

driusan:
It should definitely work with the symlinks. The question is if it would work without them..

christinerogers:
followup note - the wiki main link broke.  samir emailed me for an urgent fix on this today.

PR #6154, opened 2020-03-11
[candidate_profile] Add imaging browser widget

driusan:
This adds a widget to the candidate profile summarizing the QC
status of files inserted for the candidate.

It currently visualizes it using a C3 stacked bar graph.


driusan:
@cmadjar I initially tried that and it didn't work, the gray non-QCd files overwhelmed the pass/fail and it made the chart mostly unuseable (I think this was mostly because of the number of files that DTI modalities insert).

modules/imaging_browser/php/module.class.inc
```diff
@@ -102,7 +84,42 @@ class Module extends \Module
...
+
+            $scansummary = $DB->pselect(
+                "SELECT s.Visit_label,
+                    mst.Scan_type,
+                    COALESCE(fqc.QCStatus, '') as QC,
+                    COUNT(*) as nfiles
+                 FROM files f JOIN session s ON (s.ID=f.sessionID)
+                    JOIN mri_scan_type mst ON (f.AcquisitionProtocolID=mst.ID)
+                    LEFT JOIN files_qcstatus fqc ON (fqc.FileID=f.FileID)
+                WHERE CandID=:cid
```
cmadjar: should limit the query to f.OuputType='native' for now. All other output types will be very project dependant while 'native' refers to the raw data only

driusan: I've restricted it to 'native' and it seems to restrict it enough to re-add "Other"

modules/imaging_browser/jsx/CandidateScanQCSummaryWidget.js
```diff
@@ -0,0 +1,229 @@
...
+            },
+        });
+    });
+
+    return <div>
+        <div id='imagebreakdownchart' />
+            <ul>
+              <li>Red bar denotes number of failed QC scans.</li>
+              <li>Green bar denotes number of passed QC scans.</li>
+              <li>Grey bar denotes other QC statuses.</li>
```
cmadjar:
```suggestion
              <li>Grey bar denotes the number of non-QCed scans.</li>
```

driusan:
@cmadjar I wasn't able to reproduce your problem. If I click "Back" to get back to the profile  after changing the QC of a scan I need to refresh the page, but otherwise it's reflected after I refresh.

cmadjar:
@driusan I retested with the latest code and the issue seemed fixed somehow.

PR #6156, opened 2020-03-11
[new_profile] Auto-select only site option

zaliqarosli:
## Brief summary of changes

Resolves issue https://github.com/aces/Loris/issues/5380

(*make sure to run `make dev` when you switch branches*)
1. On main branch, on a user account affiliated with multiple sites, go to New Profile module and create a new candidate. You are required to choose a site on the form.
2. Repeat the above step but on a user account affiliated with only one site. Notice that the 'Site' field does not appear and on successful creation of the candidate, the candidate is created for the correct site.
3. On this PR branch, repeat the above steps. With a multiple-site user account, there should be no changes. With a single-site user account, notice that the 'Site' field appears and that the only site option is auto-selected. You should get no errors creating the new candidate.
4. Finally, test a couple of different select elements, both with multiple options and single options, in Loris to ensure that no unexpected behaviours appear. (select elements within module menu filters should not be affected because autoSelect is set to false there)

* Resolves https://github.com/aces/Loris/issues/5380


jsx/Form.js
```diff
@@ -438,7 +438,9 @@ class SelectElement extends Component {
     }
 
     // Add empty option
-    if (this.props.emptyOption) {
+    // If only one option exists, don't add empty option so that the only available
+    // option is selected by default
+    if (this.props.emptyOption && Object.keys(this.props.options).length != 1) {
```
driusan: I think this only makes sense if it's a required element. If it's not required, both empty and the 1 option are valid values.

zaliqarosli: see latest commit

zaliqarosli:
hmm im running into the issue of the form data object not populating with the default option if its automatically selected due to onUserInput never triggering

maltheism:
@zaliqarosli I got it to work by adding a few changes to the SelectElement component.
I gave the SelectElement a state:
```
this.state = {
      selected: null,
};
```
then made this change:
```
// Add empty option
// If only one option exists and element is required, don't add empty option so that the only available
// option is selected by default
if ((this.props.emptyOption && Object.keys(this.props.options).length != 1)
    || !this.props.required) {
        emptyOptionHTML = <option></option>;
} else if (this.state.selected === null) {
    const selected = this.props.options[
        Object.keys(this.props.options)[0]
    ];
    this.setState({selected: selected});
    this.props.onUserInput(this.props.name, selected, null, 'select');
}
```
I won't be offended if you go with something different because it isn't the most elegant solution.

zaliqarosli:
@maltheism hehe thank you ;) i was going the same way but was also unsure about the hackiness of it

HenriRabalais:
So I actually have an implementation of this in the Biobank, and one of the eventual issues I reached is that when there is only 1 value and the empty option is removed, although it appears that the only remaining option is selected, the `onUserInput` function is never triggered and so therefore the value is never passed to the parent component. I've see you've created a workaround in the `NewProfileIndex.js`, but I suggested doing this process in the selectElement itself so that it's taken care of across all modules.

What I did is I created a new prop called `autoSelect` and set the value to default to `true`. I then triggered the onUserInput function on component mount and update if the options array had a length value of 1.

Let me know what you think!

```
/**
 * Select Component
 * React wrapper for a simple or 'multiple' <select> element.
 */
class SelectElement extends Component {
  constructor(props) {
    super(props);
    this.handleChange = this.handleChange.bind(this);
  }

  componentDidMount() {
    const optionsArray = Object.keys(this.props.options);
    if (this.props.autoSelect && optionsArray.length === 1) {
      this.props.onUserInput(this.props.name, optionsArray[0]);
    }
  }

  componentDidUpdate(prevProps) {
    const options = Object.keys(this.props.options);
    const prevOptions = Object.keys(prevProps.options);
    if (options.length !== prevOptions.length ||
        !options.every((v, i) => v === prevOptions[i])) {
      if (this.props.autoSelect && options.length === 1) {
        this.props.onUserInput(this.props.name, options[0]);
      }
    }
  }
```

maltheism:
@HenriRabalais I had to comment that your solution is very cool!

zaliqarosli:
@HenriRabalais that solution looks good to me! the only thing that i would change though is to maybe make the default value false? i noticed that we wouldn't want to auto select the only option on menu filters. we would only really want to do it on forms where the field is required. we could either add the this.props.required check to the if statements here or make autoSelect an opt-in option

HenriRabalais:
@zaliqarosli Ah, yes that's a good point. It would be so nice if we could just default to `true`! I think opting in could work. Alternatively, since the filter fields are generated by the `Filter.js` class, the `autoSelect` prop could be set to `false` there for all the `SelectElement`s.

e.g.
```  
renderFilterFields() {
    return this.props.fields.reduce((result, field) => {
      const filter = field.filter;
      if (filter && filter.hide !== true) {
        let element;
        switch (filter.type) {
        case 'text':
          element = <TextboxElement key={filter.name}/>;
          break;
        case 'select':
          element = (
            <SelectElement
              key={filter.name}
              options={filter.options}
              sortByValue={filter.sortByValue}
              autoSelect={false}
            />
          );
```

zaliqarosli:
@HenriRabalais that's a good option! are all menu filters reactified for sure?

HenriRabalais:
@zaliqarosli Yeah exactly! If they're reactified they should be using the FilterableDatatable.js component.

zaliqarosli:
@HenriRabalais ready for re-review

ridz1208:
@HenriRabalais could you test it out as well ?

jsx/Form.js
```diff
@@ -469,6 +469,38 @@ class SelectElement extends Component {
     this.handleChange = this.handleChange.bind(this);
   }
 
+  /**
+   * @componentDidMount
```
driusan, line 473: Why do you have this `@` tag?

jsx/Form.js
```diff
@@ -469,6 +469,38 @@ class SelectElement extends Component {
...
+  componentDidMount() {
+    const optionsArray = Object.keys(this.props.options);
+    if (this.props.autoSelect && optionsArray.length === 1) {
+      this.props.onUserInput(this.props.name, optionsArray[0]);
+    }
+  }
+
+  /**
+   * @componentDidUpdate
+   * @param {object} prevProps - component props before component update
```
driusan, line 487: This documentation style is weird.. the description goes above the parameters in the rest of our documentation (also has the weird tag..)

zaliqarosli: yeahh this is a new requirement from eslinst and i just followed the syntax that already existed elsewhere in the file. i can modify the comment to be in whatever way, but i don't know what the proper way should be for our JS documentation. i don't think we have a standard yet?

driusan: you're right that we don't have a standard but the rest of the file goes:

```
Description

@param {type} argname - description
@return {type} 
```

from what I see? (Which is more consistent with our PHP documentation..)

driusan: The only place there's other `@something` is when it's `@constructor` on constructors instead of a description

zaliqarosli: truee i'll change it to suit that

zaliqarosli:
@HenriRabalais @driusan ready for re-review!

PR #6158, opened 2020-03-13
Fix data in fields where HTML characters were double escaped

ridz1208:
## Brief summary of changes
This fixes the issue where HTML special characters were escaped twice before being saved causing every save to re-escape the `&` of the previously escaped values.

The problematic characters are `&><"`

Example, if an instrument field contains a `>`:
after first save value becomes `&amp;gt;`
after second save `&amp;amp;amp;gt;`

Causing some values to look like this on projects running cronjobs nightly
`&amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;amp;gt;`

1. before checking out this branch, go to an instrument and save any of the characters listed above.
2. reload the page
3. [optional] save multiple times (make sure to reload the page before re-saving because values do not get refetched from the database automatically in instruments)
4. checkout this branch, and test out the script.
5. end result should show normal characters on the front end. Database values should still be escaped but without duplication

### DIFF
--- a/php/libraries/LorisForm.class.inc
+++ b/php/libraries/LorisForm.class.inc
@@ -693,10 +693,10 @@ class LorisForm
                 }
             }
         }
-        // Always sanitize user-controlled input
-        if (!is_array($newValue)) {
-            $newValue = htmlspecialchars($newValue);
-        }
+        //        // Always sanitize user-controlled input
+        //        if (!is_array($newValue)) {
+        //            $newValue = htmlspecialchars($newValue);
+        //        }
 
         return $newValue;
     }
--- a/tools/single_use/fix_double_escape.php
+++ b/tools/single_use/fix_double_escape.php
@@ -1,0 +1,165 @@
+<?php
+/**
+ * This tool scores any registered instrument that was built using the
+ * NDB_BVL_Instrument class and that has a working score() method.
+ * The command line arguments need to contain a valid test_name, a 'all' or 'one'
+ * option, a CandID and a SessionID.
+ *
+ * The 'all' option scores all existing records of an instrument if and only if
+ *  - They belong to an active timepoint
+ *  - The data-entry status of the timepoint is set to `complete`
+ *  - The administration of the timepoint is NOT `none`
+ *
+ * Note: This tool can also be used to debug the scoring algorithms for development.
+ *
+ * Limitation: This tool does not reset or nullify the score of an instrument
+ * which was previously scored but no longer meets the criteria above (i.e. if the
+ * administration was changed from `all` to `none`, the score will not be removed).
+ *
+ * @package behavioural
+ */
+
+require_once __DIR__."/../generic_includes.php";
+
+// LOGGING
+$dir = __DIR__ . "/../logs/";
+if (!is_dir($dir)) {
+    mkdir($dir);
+}
+$today   = getdate();
+$date    = strftime("%Y-%m-%d_%H:%M");
+$logPath = "$dir/fix_double_escaped_fields_$date.log";
+$logfp   = fopen($logPath, 'a');
+
+if (!$logfp) {
+    printError(
+        "No logs can be generated, path:$logPath ".
+        "does not exist or can not be written to.\n"
+    );
+}
+
+if (isset($argv[1]) && $argv[1] === 'help' || in_array('-h', $argv, true)) {
+    showHelp();
+}
+$confirm = false;
+if (isset($argv[1]) && $argv[1] === 'confirm') {
+    $confirm =true;
+}
+
+$instrumentNames = $DB->pselectCol("SELECT Test_name FROM test_names", array());
+$errorsDetected  = false;
+
+// get the list of CommentIDs for valid timepoints
+foreach($instrumentNames as $instrumentName) {
+    printOut("Checking $instrumentName");
+    try{
+        $instrument = \NDB_BVL_Instrument::factory($instrumentName);
+    } catch (Exception $e) {
+        printError(
+            "There was an error instantiating instrument $instrumentName.
+            This instrument will be skipped."
+        );
+        continue;
+    }
+    $instrumentCIDs = $DB->pselectCol(
+        "SELECT CommentID FROM flag WHERE Test_name=:tn",
+        array("tn" => $instrumentName)
+    );
+    foreach ($instrumentCIDs as $cid) {
+        $instrumentInstance = \NDB_BVL_Instrument::factory($instrumentName, $cid);
+
+        $instrumentData = \NDB_BVL_Instrument::loadInstanceData(
+            $instrumentInstance
+        );
+        $set            = array();
+        foreach ($instrumentData as $field=>$value){
+            // Each of the expressions below uniquely match each of the targeted
+            // characters indicated in the comment above the function.
+
+            // < : match any substring starting with `&`
+            // followed by 1 or more `amp;` and ending with `lt;`
+            $newValue = preg_replace('/&(amp;)+lt;/', '<', $value);
+            // > : match any substring starting with `&`
+            // followed by 1 or more `amp;` and ending with `gt;`
+            $newValue = preg_replace('/&(amp;)+gt;/', '>', $newValue);
+            // " : match any substring starting with `&`
+            // followed by 1 or more `amp;` and ending with `quot;`
+            $newValue = preg_replace('/&(amp;)+quot;/', '"', $newValue);
+            // & : match any substring starting with `&`
+            // followed by 2 or more `amp;` (because 1 is normal in the database
+            // since it is the escaped form of `&`) and
+            // NOT ending with `lt;` or `gt;` or `quot;` or `amp;`
+            // (the last one is to ensure we don't match subsequences from the
+            // case above).
+            $newValue = preg_replace('/&(amp;){2,}(?!(lt;|gt;|quot;|amp;))/', '&', $newValue);
+
+            if (!empty($value) && !empty($newValue) && $newValue !== $value) {
+                printOut(
+                    "CommentID: $cid - Value at $field will be modified. ".
+                    "\n\tCurrent Value: $value".
+                    "\n\tWill be replaced by: $newValue\n"
+                );
+
+                $set[$field]    = $newValue;
+                $errorsDetected = true;
+            }
+        }
+        if (!empty($set) && $confirm) {
+            $instrumentInstance->_save($set);
+        }
+    }
+}
+
+if (!$confirm && $errorsDetected) {
+    printOut("\nRun tool again with `confirm` argument to apply changes");
+} else {
+    printOut("End");
+}
+fclose($logfp);
+
+/*
+ * Prints to log file
+ */
+function logMessage($message)
+{
+    global $logfp;
+    if (!$logfp) {
+        //The log file could not be instantiated
+        //use print instead
+        print_r($message);
+    }
+    $now_string = strftime("%Y-%m-%d %H:%M:%S");
+    fwrite($logfp, "[$now_string] $message\n");
+
+}
+
+/*
+ * Prints to STDERR
+ */
+function printError($message)
+{
+    logMessage($message);
+    fwrite(STDERR, "$message \n");
+}
+
+/*
+ * Prints to STDOUT
+ */
+function printOut($message)
+{
+    logMessage($message);
+    print_r("$message\n");
+}
+
+function showHelp()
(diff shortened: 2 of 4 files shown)


ridz1208:
@driusan discussion required for : 
1. where to fix the issue ? I just commented out the first escaping in the lorisform class. the sesond escaping occurs in the `update()` function of the database class. but i think that one should stay there
2. which branch to submit this to.

ridz1208:
@sruthymathew123 @nicolasbrossard can you check if any of the IBIS data contains these characters ?

running the tool without any arguments should find such fields

php/libraries/LorisForm.class.inc
```diff
@@ -693,10 +693,10 @@ class LorisForm
                 }
             }
         }
-        // Always sanitize user-controlled input
-        if (!is_array($newValue)) {
-            $newValue = htmlspecialchars($newValue);
-        }
+        //        // Always sanitize user-controlled input
```
johnsaigle: I don't like this fix because it removes an important security measure. I understand though that the double-escaping bug has caused data loss which is a severe issue. 

I'll see if I can come up with an alternate fix this week. If I can't figure it out then I think we'll have to go ahead with this and find another way to address XSS issues.

driusan: What data loss? Double escaped entries can be restored with a script. It's very unlikely that a patient entered something like `&amp;` on a form.

ridz1208: due to the number of `&amp;` added several fields got truncated by mySQL

johnsaigle: @driusan This was discussed in a group chat on Slack and you can view the details there. `<`, `>`, and `&` seem to be common characters used by clinicians when entering instruments by hand (!) and so there is a need to transcribe them literally into forms.

Since they can't be black-listed, we need to do better at displaying these characters to the front-end.

johnsaigle: @ridz1208 The problem arose because of MySQL 5.1 though right? If CCNA can't/won't upgrade then this might be a project override rather than something that needs to go into core.

ridz1208: it's mostly the `"` that are used by clinicians.

and @johnsaigle not sure I see what can be an override. everyone still has the issue, it's just that the other people's data wont get truncated if they are on more recent versions of mysql, they will still have the double escaping issue. and given that this bug existed since LORIS 17, I think we should cover a wide range of versions in our fix

tools/single_use/fix_double_escape.php
```diff
@@ -0,0 +1,280 @@
+<?php
```
johnsaigle:
```suggestion
<?php declare(strict_types=1);
```

tools/single_use/fix_double_escape.php
```diff
@@ -0,0 +1,280 @@
...
+ *  - The data-entry status of the timepoint is set to `complete`
+ *  - The administration of the timepoint is NOT `none`
+ *
+ * Note: This tool can also be used to debug the scoring algorithms for development.
+ *
+ * Limitation: This tool does not reset or nullify the score of an instrument
+ * which was previously scored but no longer meets the criteria above (i.e. if the
+ * administration was changed from `all` to `none`, the score will not be removed).
+ *
+ * @package behavioural
```
johnsaigle:
```suggestion
```

Not needed

tools/single_use/fix_double_escape.php
```diff
@@ -0,0 +1,280 @@
...
+$report = false;
+if (isset($argv[1]) && $argv[1] === 'report') {
+    $report =true;
+}
+$repair = false;
+if (isset($argv[1]) && $argv[1] === 'repair') {
+    $repair =true;
+}
+$useHistory =false;
+if (in_array('use-history', $argv, true)) {
```
johnsaigle: Based on the logic for these if-statements you should probably throw an error in case too many arguments are passed, for example if someone tries to do both `report` and `repair`

tools/single_use/fix_double_escape.php
```diff
@@ -0,0 +1,280 @@
...
+
+// FIRST loop just reporting all potential problematic fields
+foreach($instrumentNames as $instrumentName) {
+    printOut("Checking $instrumentName");
+    try {
+        $instrument = \NDB_BVL_Instrument::factory($instrumentName);
+    } catch (Exception $e) {
+        printError(
+            "There was an error instantiating instrument $instrumentName.
+            This instrument will be skipped."
```
johnsaigle: It would be helpful to print the stack trace of the exception

tools/single_use/fix_double_escape.php
```diff
@@ -0,0 +1,280 @@
...
+                printError(
+                    "There was an error instantiating instrument $instrumentName for " .
+                    "$cid.This instrument will be skipped."
+                );
+                continue;
+            }
+            foreach ($fields as $field => $value) {
+
+                if ($useHistory && array_key_exists($field, $confirmedTruncations[$instrumentName][$cid])) {
+                    //TODO ADD CODE TO REPLACE THE VALUE BY THE DATA FROM THE HISTORY TABLE IF TRUNCATION IS CONFIRMED
```
johnsaigle: Is this still TODO?

tools/single_use/fix_double_escape.php
```diff
@@ -116,10 +119,70 @@
...
+	// Construct history query
+	$listCIDs = "'" . implode($CIDs, "','") . "'";
+	$listFields = "'" . implode($allFields, "','") . "'";
+	$listTables = "'" . implode($allTables, "','") . "'";
+	$query = "SELECT primaryVals as CommentID, col, tbl, new, changeDate
+		FROM history 
+		WHERE primaryCols='CommentID' AND
+		primaryVals IN ($listCIDs) AND
+		tbl IN ($listTables) AND
+		col IN ($listFields)";
```
ridz1208: not sure this query would work, lets say 
 - commentID XYZ has an issue with instrument A.
 - commentID WXY has an issue with instrument B.

Then your results will include 
 - CommentId XYZ instrument B 
 - Commentid WXY instruemnt A

Both of which might not be problematic. I think this query needs to be exact and exhaustive.

ridz1208: @jesscall

tools/single_use/fix_double_escape.php
```diff
@@ -116,10 +119,70 @@
...
+		col IN ($listFields)";	
+	
+	$result = $DB->pselect(
+		$query, 
+		[]
+	);
+
+	// Reverse the array obtained from history data so that it is in anti-chronological order
+	// Does not preserve keys, assuming that the keys are indexes & not relevant info
+	$historyData = array_reverse($result, false);
```
ridz1208: @jesscall there is no garantee that the array returned in in chronological order. you should instead use an order by in the query itself

tools/single_use/fix_double_escape.php
```diff
@@ -152,22 +216,28 @@
 
 
 
-    foreach ($escapedEntries as $instrumentName=>$cids) {
-        printOut("Opening $instrumentName");
+    foreach ($escapedEntries as $tableName=>$cids) {
```
ridz1208: @jesscall replacing instrument_name with tablename might be problematic since you need the actual instrument name to instanciate the instrument

jesscall: hmmm, i see... 
the reasoning for using table name in certain cases is very specific to how CCNA has set up Follow-Up instruments. 
I can make adjustments so that `tableName` is used specifically for constructing the queries to `information_schema` and `history`. However, I don't think the logic for doing so is apparent in the Core-LORIS use case, will adding a comment that indicates a project-specific choice be sufficient ?

tools/single_use/fix_double_escape.php
jesscall: Having the query be exact is very costly and I wasn't actually able to get results because the server kept disconnecting... 
This way, the query is not as costly and we do most of the heavy lifting in the script itself by comparing to the array of `confirmedTruncations` 

I know you mentioned it might be an issue with resourcing. Perhaps we can discuss offline and with IT and try to see if an exact query will work when more resources are placed on the history DB.

tools/single_use/fix_double_escape.php
ridz1208: I dont think there are project -specific stuff here. Every LORIS project might be using different instrument and table names. there is no reason here preventing you from associating each instruemnt to a table and use that mapping to instanciate them properly

jesscall:
still working on this. trying to see what i can do about the query to history, might potentially be able to rework the logic of the script entirely, depending if my query works

@ridz1208 i'll let you know when its ready for testing

jesscall:
@ridz1208 prod had 500 error because of the query that was being run... i think we should keep the script as is, even though the query is returning more data than needed. We can parse through it on PHP side and not put too much strain on the DB and server which has limited memory.

Can you do a prelim code review and let me know if there's anything else i should change or improve?

ridz1208:
@sruthymathew123 @cmadjar 

I have added to this PR a "lite" version of the tool designed to only report on issues. this script can do the checking using object instanciation or raw SQL queries. the raw SQL way obviously does not allow you to have a database table name different then the instrument name so it might omit some instruments in your database.

if you decide to use the SQL way to avoid object instanciation errors, you can use the `instrument_double_escape_report.php` script as follows

`php instrument_double_escape_report.php use-database`

let me know if it detects anything.

tools/single_use/instrument_double_escape_fix.php
```diff
@@ -0,0 +1,348 @@
...
+            $cid = $data['CommentID'];
+            $field = $data['col'];
+            $val = $data['new'];
+
+            // Check if in confirmed truncations list & length of field is less than max field length
+            // & no value has been assigned to untruncatedValues yet
+            // I.E. if this is the most recent value with character length < max
+            if (array_key_exists($field, $confirmedTruncations[$tbl][$cid]) &&
+                strlen($val) < $maxFieldLengths[$tbl][$field] &&
+                !isset($untruncatedValues[$tbl][$cid][$field])
```
jesscall, line 187:
```suggestion
            if (isset($val) &&
                isset($confirmedTruncations[$tbl][$cid][$field]) &&
                strlen($val) < $maxFieldLengths[$tbl][$field]['CHARACTER_MAXIMUM_LENGTH'] &&
                !isset($untruncatedValues[$tbl][$cid][$field])
```

tools/single_use/instrument_double_escape_fix.php
```diff
@@ -0,0 +1,348 @@
...
+                printError(
+                    "There was an error instantiating instrument $instrumentName for " .
+                    "$cid.This instrument will be skipped."
+                );
+                continue;
+            }
+            foreach ($fields as $field => $value) {
+                if ($useHistory &&
+                    array_key_exists($field, $confirmedTruncations[$tableName][$cid]) &&
+                    array_key_exists($field, $untruncatedValues[$tableName][$cid])
```
jesscall, line 241:
```suggestion
                    isset($confirmedTruncations[$tableName][$cid][$field]) &&
                    isset($untruncatedValues[$tableName][$cid][$field])
```

tools/single_use/instrument_double_escape_fix.php
```diff
@@ -0,0 +1,348 @@
...
+                        "CommentID: $cid - Value at $field will be modified. " .
+                        "\n\tCurrent Value: $value" .
+                        "\n\tWill be replaced by: $newValue\n"
+                    );
+                    $set[$field] = $newValue;
+                }
+            }
+            if (!empty($set) && $repair) {
+                $instrumentInstance->_save($set);
+            }
```
jesscall, line 282:
```suggestion
            }
            unset($set);
```

ridz1208:
Bugfix sent to 22 with reported script
bugfix also sent to 23

PRs below

https://github.com/aces/Loris/pull/6223
https://github.com/aces/Loris/pull/6477

PR #6164, opened 2020-03-17
[Database] Add default argument for pselect functions

johnsaigle:
## Brief summary of changes

`$params` argument for `pselect*` functions now defaults to the empty array. So instead of writing e.g.

```php
$result = pselectCol('SELECT Test_name from test_names', array());
```

we can write

```php
$result = pselectCol('SELECT Test_name from test_names');
```

### DIFF
--- a/php/libraries/Database.class.inc
+++ b/php/libraries/Database.class.inc
@@ -719,7 +719,7 @@ class Database
      *               row returned by the query, and each element is an associative
      *               array representing the row in the format ColumnName => Value
      */
-    function pselect(string $query, array $params): array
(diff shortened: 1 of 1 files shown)


johnsaigle:
> 1) ConflictResolverTestIntegrationTest::testFiltersForUnresolvedConflicts
UnexpectedJavascriptException: TypeError: document.querySelector(...) is null

Transitory (?) failure on PHP 7.3 tests. 7.4 is fine. Restarting Travis.

driusan:
Arguments should be explicit, I don't think this is a good idea.

PR #6165, opened 2020-03-17
[FilesUploadHandler] Remove special characters in file names

### LINKED ISSUE #6162: Forbid special characters in uploaded filenames
Yes, all modules with file uploading capabilties.

Special characters should not be allowed for filenames. This includes characters that do special things in UNIX, like `$`, `<`, `&`, etc. 

These should be validated on the back-end in `FilesUploadHandler` and issue a 400 response.

**Describe alternatives you've considered**
The alternative is to use either `unsafeInsert` for file names or to encode/decode special characters on insertion to the database. This is tricky to maintain and do in a general way especially since it relies on many other parts of the codebase not changing. 

**Additional context**
#6136 arose due to this problem.

johnsaigle:
## Brief summary of changes

Files uploaded using this class can now contain only alphanumeric characters as well as dashes and underscores.

This is to avoid problems with sanitizing and encoding/decoding filenames, such as #6136.

No but I don't think this is documented anywhere.

1. On this branch, upload a file to the instrument_manager module containing a forbidden character.
2. Ensure that your file is uploaded successfully but with the offending character removed.

* Resolves #6162

### DIFF
--- a/php/libraries/FilesUploadHandler.class.inc
+++ b/php/libraries/FilesUploadHandler.class.inc
@@ -177,9 +177,9 @@ class FilesUploadHandler implements RequestHandlerInterface
                 . preg_replace(
                     "/[^a-z0-9\_\-\.]/i",
                     '',
-                    $pathinfo[PATHINFO_BASENAME]
+                    $pathinfo['basename']
                 )
-                . $pathinfo[PATHINFO_EXTENSION];
+                . $pathinfo['extension'];
 
             /* If file exists, set response code to 'Conflict' unless the
              * calling code wants to overwrite the file.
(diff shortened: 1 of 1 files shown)


ridz1208:
@johnsaigle why do we want to block special characters exactly ? is there a know issue about this ?

johnsaigle:
Yeah you can look at the linked issue (and the issue tagged within that issue) for details on how this came up

ridz1208:
@johnsaigle The issues linked barely explain why this is needed other than saying "special characters should be forbidden"

As far I i could tell the issue might be caused AGAIN by the automated DB escaping of fields which I think we had decided to remove in a near future and instead escape special chars as upstream as possible when reloading the data

@driusan you think we should adopt this or just make sure to remove the escaping in this release ?

johnsaigle:
It's still not a bad idea to do on filenames but it's up to you two whether this goes forward or not

ridz1208:
This issue will be handled differently. the escaping of special characters is a high priority fix for the 24 release

ridz1208:
@johnsaigle thanks for bringing it up

Issue #6163, opened 2020-03-17
electrophysiology_browser SQL table not installed automatically


lingma:
new installation won't have access to `electrophysiology_browser` because the installer doesn't include the SQL of `electrophysiology_browser`

cmadjar:
What installer?

lingma:
> What installer?

The Loris installer. I checked, nothing mentioned, I will create a PR soon.

johnsaigle:
Resolved by linked PR

PR #6168, opened 2020-03-17
[Install] add ElectrophysiologyTables.sql to list of schemas sourced during install process

lingma:
This PR adds eletrophysiology related SQL in installer.

* Resolves issue #6163

### DIFF
--- a/php/installer/Installer.class.inc
+++ b/php/installer/Installer.class.inc
@@ -26,8 +26,15 @@ namespace LORIS\Installer;
  */
 class Installer
 {
-
     private $_lastErr;
+
+    private const SCHEMA_FILES = ['00-schema.sql',
+        '01-Permission.sql',
+        '02-Modules.sql',
+        '03-ConfigTables.sql',
+        '04-Help.sql',
+        '05-ElectrophysiologyTables.sql'
+    ];
 
     /**
      * Get the last error that occurred in this install process.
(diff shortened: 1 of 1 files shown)


php/installer/Installer.class.inc
```diff
@@ -96,11 +103,15 @@ class Installer
...
-        && file_exists(__DIR__ . "/../../SQL/0000-00-04-Help.sql");
+        $fileExist = true;
+        foreach ($this->_schemaFiles as $fileName) {
+            if ($fileExist) {
+                $fileExist = file_exists(__DIR__ . '/../../SQL/0000-00-'.$fileName);
+            } else {
+                return false;
+            }
+        }
+        return $fileExist;
```
johnsaigle:
```suggestion
        foreach ($this->_schemaFiles as $fileName) {
            if (!file_exists(__DIR__ . '/../../SQL/0000-00-'.$fileName)) {
                return false;
            }
        return true;
```

johnsaigle: This is a little cleaner

php/installer/Installer.class.inc
```diff
@@ -26,9 +26,16 @@ namespace LORIS\Installer;
...
-
     private $_lastErr;
 
+    private $_schemaFiles = ['00-schema.sql',
+        '01-Permission.sql',
+        '02-Modules.sql',
+        '03-ConfigTables.sql',
+        '04-Help.sql',
+        '05-ElectrophysiologyTables.sql'
+    ];
```
johnsaigle: This should probably be declared as a constant instead of a variable since it's not intended to change, but it's not a big deal

php/installer/Installer.class.inc
```diff
@@ -26,9 +26,16 @@ namespace LORIS\Installer;
  */
 class Installer
 {
-
     private $_lastErr;
 
+    private const _SCHEMA_FILES = ['00-schema.sql',
```
johnsaigle:
```suggestion
    private const SCHEMA_FILES = ['00-schema.sql',
```
underscores aren't needed for constants

php/installer/Installer.class.inc
```diff
@@ -26,9 +26,16 @@ namespace LORIS\Installer;
  */
 class Installer
 {
-
     private $_lastErr;
 
+    private const SCHEMA_FILES = ['00-schema.sql',
```
maltheism: Off-topic: I'm surprised the first array element syntax format passes phan. I would think `00-schema.sql` would need to be on a separate line.

lingma: It is a variable declaration, not command, so I think not strict. I don't like the format like
[
'00-schema.sql
]

PR #6169, opened 2020-03-17
[My Preferences] Change contents of email sent to user upon changing their account details

### LINKED ISSUE #6079: My Preferences module should not email your password to you
**Describe the bug**
Upon editing your password, my preferences will email it to you in plaintext.

**To Reproduce**
1. Change your password using My Preferences using a valid email.

The password should change but no email should be sent. We only want to email passwords to people on a one-time basis in specific scenarios, such as an admin changing a password for a user. These passwords are forced to reset upon login which mitigates the issue of sending passwords in cleartext. 

Changing one's own password is a routine operation and there's no reason to send an email.

johnsaigle:
## Brief summary of changes

My Preferences was sending a user's password in plaintext to their email whenever they changed it.

Edit User does this when a user has a password generated for them by the server. In this case it is acceptable because the password must be reset on login. My Preferences shares a lot of code with User Accounts so I believe it was included as a feature accidentally.

I've created a new email template notifying a user that their account details have changed.

This PR also does a bit of cleanup within the function.

1. Change your account details via the My Preferences module. You should receive an email. That email should not display your password.

* Resolves #6079

### DIFF
--- a/modules/my_preferences/php/my_preferences.class.inc
+++ b/modules/my_preferences/php/my_preferences.class.inc
@@ -195,18 +195,6 @@ class My_Preferences extends \NDB_Form
         $factory = \NDB_Factory::singleton();
         $config  = $factory->config();
 
-        // send the user an email
-        $msg_data          = [];
-        $msg_data['study'] = $config->getSetting('title');
-        $msg_data['url']   = $factory->settings()->getBaseURL();
-        $msg_data['realname'] = $values['Real_name'];
-        $msg_data['username'] = $user->getUsername();
-        $msg_data['password'] = $values['Password_hash'];
-
-        $template = (is_null($this->identifier))
-            ? 'new_user.tpl' : 'edit_user.tpl';
-        \Email::send($values['Email'], $template, $msg_data);
-
         $this->tpl_data['success'] = true;
     }
 
--- a/modules/my_preferences/php/my_preferences.class.inc
+++ b/modules/my_preferences/php/my_preferences.class.inc
@@ -110,13 +110,6 @@ class My_Preferences extends \NDB_Form
 
         $user = \User::factory($this->identifier);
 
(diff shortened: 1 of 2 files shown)


johnsaigle:
This is based on 23.0-release because only this branch contains the changes that made my_preferences its own module and basing it on master makes this PR look like it contains errors.

However, this PR should go to master after 23.0 is released.

smarty/templates/email/account_details_changed.tpl
```diff
@@ -0,0 +1,14 @@
+Subject: Your Account Details Have Changed
+
+{$realname},
+
+Your account details have changed for the LORIS study {$study}.
+
+If you made these changes then you can ignore this email.
+
+If you did not make these changes, please contact your project administrator
+immediately.
```
lingma: Hi {$realname},

Your account details have been changed for the LORIS study {$study}.

If you have made some changes then this email is a security notification.

If you did not make any changes, please contact your project administrator
immediately.

modules/my_preferences/php/my_preferences.class.inc
```diff
@@ -99,21 +99,15 @@ class My_Preferences extends \NDB_Form
...
         $values['Real_name'] = $values['First_name'] . ' ' . $values['Last_name'];
 
         $user = \User::factory($this->identifier);
 
-        // If editing a user and nothing was specified in the password text field
-        // remove the key from the value set, otherwise Password_hash
-        // will be set to '' by the system
-        if ($values['Password_hash'] == '') {
-            unset($values['Password_hash']);
-        }
```
lingma: If the password is not changed, the password won't be in the email. If changed, then yes.

johnsaigle: These lines were moved below and don't affect what's in the email. Whether the password appears is determined by the `$msg_data` array and the email template.

PR #6171, opened 2020-03-17
[Database] Add return type to pselectOne() and add new pselectOneInt()

johnsaigle:
## Brief summary of changes

Adds a return type of `string` to pselectOne() in order to assist us with using strict types throughout LORIS.

Also adds a new function `pselectOneInt()` as a wrapper of `pselectOne()`. This new function automatically validates the cell returned by pselectOne and converts it to an integer. It returns `null` when the query results in something that can't be reliably cast to an integer, such as the empty set.

This is to assist in cases where integer values are to be retrieved from the database, such as COUNT() results, IDs, etc. Rather than require the calling code to wrap results with `intval()` (which is problematic for some inputs), or `int()`, they can call this function to have validation be done automatically.

* Resolves #6164


johnsaigle:
I'm not sure why Travis is failing. I didn't modify that file. I see it has something to do with the `instrument_builder`. When I load the instrument builder module it doesn't work -- but there is also no error in the back-end or in the JavaScript console. @kongtiaowang do you have any ideas?

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1469,17 +1469,14 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
             [
                 'CID' => $this->getCommentID(),
             ]
         );
 
-        if (!empty($fieldValue)) {
-            return $fieldValue;
-        }
-        return -1;
+        return $fieldValue == 0 ? $fieldValue : -1;
```
kongtiaowang, line 1479: @johnsaigle in your way, it only allows return 0 or -1.
  It shuld be like this   [  return $fieldValue !== null ? $fieldValue : -1;   ]

johnsaigle: Thanks! I'll try that

johnsaigle: Hmm it looks like this function was always supposed to return an int in the function signature. If it returned null before there would've been an error. So I'm not sure this is the problem.

laemtl: @kongtiaowang @johnsaigle I slightly updated the return statement and PR is passing Travis now.

johnsaigle:
@kongtiaowang @driusan Please review

php/libraries/Database.class.inc
```diff
@@ -988,6 +987,34 @@ class Database
...
+     * the query returns a value that cannot be validated as an integer, the
+     * function returns null. The calling code is responsible for handling the
+     * null case appropriately.
+     *
+     * @throws \DomainException When the cell returned by the query cannot be
+     *                          validated as an integer.
+     */
+    function pselectOneInt(string $query, array $params): ?int
+    {
+        $result = filter_var(
```
ridz1208, line 1007: apparently `FILTER_VALIDATE_INT` does not interpret 0 as an int so a special case needs to be added for the 0 case

johnsaigle: > As of PHP 5.4.11, the numbers +0 and -0 validate as both integers as well as floats (using FILTER_VALIDATE_FLOAT and FILTER_VALIDATE_INT). Before PHP 5.4.11 they only validated as floats (using FILTER_VALIDATE_FLOAT). 

From PHP manual https://www.php.net/manual/en/filter.filters.validate.php

Where did you get this info from? did it fail manual tests?

ridz1208: https://www.w3schools.com/php/filter_validate_int.asp

johnsaigle: I ran their example code and it seems to work fine. What happens when you test it? it's possible that article was written before 5.4.11 -- I can't find a date on the page

php/libraries/Utility.class.inc
```diff
@@ -283,6 +283,118 @@ class Utility
         return self::getSubprojectList($projectID);
     }
 
+    /**
```
driusan, line 286: I don't think these functions are supposed to be added in this PR. It seems to be the result of a bad rebase since the functions were moved.. 

This PR should be re-based onto master, and hopefully this change will disappear. If not the functions need to be deleted from here. After rebasing, the functions need to be tested in their new location.

driusan:
Adding needs maintainer if someone wants to try to address my latest review, feel free..

php/libraries/Utility.class.inc
ridz1208: @driusan they were previously deleted, i redeleted them. lets see if travis is going to complain

PR #6175, opened 2020-03-18
Update phpstan

johnsaigle:


xlecours:
@johnsaigle Was it necessary?

johnsaigle:
It's not necessary for any specific reason but for developer tools I think it makes sense to try to use the more recent available versions. Since this won't affect production LORIS I don't think there's any harm.

driusan:
@johnsaigle this can't be merged because of conflicts, can you rebase?

johnsaigle:
@driusan done!

PR #6177, opened 2020-03-19
selenium update

kongtiaowang:
## Brief summary of changes
update selenium 2.53 to 3.81

Travis


kongtiaowang:
@maltheism The latest version of selenium is not compatible with our integration test system and is not only a matter of version number. This upgrade prevents Travis from creating two selenium iamges.

PR #6179, opened 2020-03-24
[Instrument] Add explicit scope on class properties

### LINKED ISSUE #6178: NDB_BVL_Instrument variable declarations not clear or wrong
In the declartion of NDB_BVL_Instrument, we have
   /**
     * Array of required fields to check
     *
     * @access private
     */
    var $_requiredElements = array();
@access private should be wrong, in fact $_requiredElements is public as declared, it could be protected. In both case, _ is not necessary. There are other variables with the similar issues in the same file.

The description should be more clear as to what this variable is to be used, what values are expected. The issue has been seen before, but noticed in the PR #5949, we need to address the issue once the PR has been merged to avoid merge conflicts.

johnsaigle:
## Brief summary of changes

Remove `var` from variables in favour of more explicit scope keywords.

I tried to maintain the current operation of the class rather than try to 'correct' variables; there are some properties that should likely be protected instead of public but this should be done in later refactoring.

* Resolves #6178

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -52,9 +52,9 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     /**
      * Database table containing data referenced by $this->commentID
      *
-     * @access private
+     * @access protected
      */
-    private $_table;
+    protected $table;
 
     /**
      * Cache location for the date of administration
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1071,7 +1071,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      * Gets the date of administration of an instrument instance.
      * Caches to reduce DB hits
      *
-     * @return mixed $_dateOfAdministration the date of administration from
+     * @return mixed $dateOfAdministration the date of administration from
      *         the DB, in the format "YYYY-MM-DD"
      */
     function getDateOfAdministration()
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1,29 +1,12 @@
 <?php
-/**
- * This file contains the base class for instruments in Loris
- *
- * PHP Version 7
- *
- * @category Main
- * @package  Behavioural
- * @author   Loris team <info-loris.mni@mcgill.ca>
- * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
- * @link     https://www.github.com/aces/Loris-Trunk/
- */
 use \Psr\Http\Message\ServerRequestInterface;
 use \Psr\Http\Message\ResponseInterface;
 use \Loris\StudyEntities\Candidate\CandID;
 
 /**
- * Base class for all NeuroDB behavioural instruments
+ * Base class for all LORIS behavioural instruments.
  *
- * Throws PEAR errors. Also requires PEAR HTML_Quickform.
- *
- * @category Main
- * @package  Behavioural
- * @author   Loris team <info-loris.mni@mcgill.ca>
- * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
- * @link     https://www.github.com/aces/Loris-Trunk/
+ * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
  */
 abstract class NDB_BVL_Instrument extends NDB_Page
 {
@@ -38,9 +21,10 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     protected $jsonData = false;
 
     /**
-     * HTML_Quickform object
+     * The form object for this instrument.
      *
-     * @access private
+     * @var    LorisForm
+     * @access public
      */
     public $form;
 
@@ -54,59 +38,59 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     /**
      * Test name (short name), equivalent to test_names.Test_name
      *
-     * @access private
+     * @access public
      */
-    var $testName;
+    public $testName;
 
     /**
      * Instrument instance CommentID
      *
-     * @access private
+     * @access public
      */
-    var $commentID;
+    public $commentID;
 
     /**
      * Database table containing data referenced by $this->commentID
      *
      * @access private
      */
-    var $table;
+    private $_table;
 
     /**
      * Cache location for the date of administration
      *
-     * @access private
+     * @access public
      */
-    var $dateOfAdministration;
+    public $dateOfAdministration;
 
     /**
      * Additional form defaults - should only be used for lorisform
      * static elements!
      *
-     * @access private
+     * @access public
      */
-    var $localDefaults = ['emptyCell' => '&nbsp;'];
+    public $localDefaults = ['emptyCell' => '&nbsp;'];
 
     /**
      * String to separate the group elements
      *
-     * @access private
+     * @access public
      */
-    var $_GUIDelimiter = "</td>\n<td>";
+    public $GUIDelimiter = "</td>\n<td>";
 
     /**
      * Commonly used level of indentation
      *
-     * @access private
+     * @access public
      */
-    var $indent = "&nbsp;&nbsp;&nbsp;&nbsp;";
+    public $indent = "&nbsp;&nbsp;&nbsp;&nbsp;";
 
     /**
      * Array of required fields to check
      *
-     * @access private
+     * @access protected
      */
-    var $_requiredElements = [];
+    protected $requiredElements = [];
 
     /**
      * Array of date fields to be processed into QuickForm dates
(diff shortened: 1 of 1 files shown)


php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1093,7 +1079,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      * Gets the date of administration of an instrument instance.
      * Caches to reduce DB hits
      *
-     * @return mixed $dateOfAdministration the date of administration from
+     * @return mixed $_dateOfAdministration the date of administration from
```
```suggestion
     * @return mixed $dateOfAdministration the date of administration from
```

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -54,59 +38,59 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+    protected $table;
 
     /**
      * Cache location for the date of administration
      *
-     * @access private
+     * @access public
      */
-    var $dateOfAdministration;
+    public $dateOfAdministration;
```
lingma, line 64: This one should be protected if not private.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -54,59 +38,59 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
 
     /**
      * Additional form defaults - should only be used for lorisform
      * static elements!
      *
-     * @access private
+     * @access public
      */
-    var $localDefaults = array('emptyCell' => '&nbsp;');
+    public $localDefaults = array('emptyCell' => '&nbsp;');
```
lingma, line 72: no need to be public

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -54,59 +38,59 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
 
     /**
      * String to separate the group elements
      *
-     * @access private
+     * @access public
      */
-    var $_GUIDelimiter = "</td>\n<td>";
+    public $GUIDelimiter = "</td>\n<td>";
```
lingma, line 79: same, no need to be public, could be protected

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -54,59 +38,59 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
 
     /**
      * Commonly used level of indentation
      *
-     * @access private
+     * @access public
      */
-    var $indent = "&nbsp;&nbsp;&nbsp;&nbsp;";
+    public $indent = "&nbsp;&nbsp;&nbsp;&nbsp;";
```
lingma, line 86: no need to be public

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -54,59 +38,59 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+    public $testName;
 
     /**
      * Instrument instance CommentID
      *
-     * @access private
+     * @access public
      */
-    var $commentID;
+    public $commentID;
```
lingma, line 50: Better be protected

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -54,59 +38,59 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     /**
      * Test name (short name), equivalent to test_names.Test_name
      *
-     * @access private
+     * @access public
      */
-    var $testName;
+    public $testName;
```
lingma, line 43: protected

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -38,9 +21,10 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     protected $jsonData = false;
 
     /**
-     * HTML_Quickform object
+     * The form object for this instrument.
      *
-     * @access private
+     * @var    LorisForm
```
lingma, line 26: protected might be better

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -116,18 +100,20 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
      */
-    var $dateTimeFields = array();
+    public $dateTimeFields = array();
 
     /**
      * Array of date fields that contain only month year fields in frontend with
      * a static value for the date
      */
-    var $monthYearFields = array();
+    public $monthYearFields = array();
```
lingma, line 109: I feel all these fields should be `protected`

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -141,29 +127,29 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
      */
-    var $displayAllFields = false;
+    public $displayAllFields = false;
 
-    var $WrapperTextElements           = array();
-    var $WrapperNumericElements        = array();
-    var $WrapperDateWithStatusElements = array();
+    public $WrapperTextElements           = array();
+    public $WrapperNumericElements        = array();
+    public $WrapperDateWithStatusElements = array();
```
lingma, line 152: protected

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -141,29 +127,29 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      * Whether the "Validity" field is shown as a flag for an instrument
      * or not.
      *
-     * @access private
+     * @access public
      */
-    var $ValidityEnabled = true;
+    public $ValidityEnabled = true;
```
lingma, line 132: protected

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -54,59 +38,59 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+    public $commentID;
 
     /**
      * Database table containing data referenced by $this->commentID
      *
-     * @access private
+     * @access protected
      */
-    var $table;
+    protected $table;
```
driusan, line 57: This is changing the access level. `var` is equivalent to `public` in PHP, even if there's a comment saying otherwise. Changing this will break scripts that access the table (ie. lorisform_parser)

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -54,59 +38,59 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+    public $indent = "&nbsp;&nbsp;&nbsp;&nbsp;";
 
     /**
      * Array of required fields to check
      *
-     * @access private
+     * @access protected
      */
-    var $_requiredElements = [];
+    protected $requiredElements = [];
```
driusan, line 93: This is changing the variable name which is going to break any instrument subclass that uses it.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -116,18 +100,20 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
-    var $monthYearFields = [];
+    public $monthYearFields = [];
     /**
      * Array of column names to be ignored by the double data entry conflict
      * detector.
+     *
+     * @access public
      */
-    var $_doubleDataEntryDiffIgnoreColumns
+    public $doubleDataEntryDiffIgnoreColumns
```
driusan, line 116: this is also changing the variable name which is going to break any subclass or script that uses it.

php/libraries/NDB_BVL_Instrument.class.inc
johnsaigle: PHPCS has a rule where only privately scoped variable can be prefixed with an underscore. We can change this to private or disable phpcs. Which do you want?

php/libraries/NDB_BVL_Instrument.class.inc
johnsaigle: PHPCS has a rule where only privately scoped variable can be prefixed with an underscore. We can change this to private or disable phpcs. Which do you want?

php/libraries/NDB_BVL_Instrument.class.inc
driusan: Keep it `var` until there's a better solution, either changing the name or changing the scope is going to break things and we shouldn't be disabling phpcs on anything.

ridz1208:
@driusan please re review

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2116,7 +2102,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
 
         // Loop over this instance data
         foreach ($thisData AS $key=>$value) {
-            if (!in_array($key, $this->_doubleDataEntryDiffIgnoreColumns)) {
+            if (!in_array($key, $this->doubleDataEntryDiffIgnoreColumns)) {
```
lingma, line 2105: doubleDataEntryDiffIgnoreColumns should be _doubleDataEntryDiffIgnoreColumns

driusan: thanks, I missed that on re-review. I just created a ticket.

PR #6180, opened 2020-03-24
[User] Add helper function to \User class to validate accessibility

johnsaigle:
## Brief summary of changes

Adds a function to determine whether a user can access another user.

See #6093

### DIFF
--- a/php/libraries/User.class.inc
+++ b/php/libraries/User.class.inc
@@ -572,11 +572,16 @@ class User extends UserPermissions implements \LORIS\StudyEntities\AccessibleRes
 
         $projMatch   = !empty(
             array_intersect(
-                $this->getProjectID(),
+                $this->getProjectIDs(),
                 $user->getProjectIDs()
             )
         );
-        $centerMatch = !empty(array_intersect($this->getCenterIDs(), $centers));
+        $centerMatch = !empty(
+            array_intersect(
+                $this->getCenterIDs(), 
+                $user->getCenterIDs()
+            )
+        );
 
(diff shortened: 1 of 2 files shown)


johnsaigle:
@jesscall This should be ready for review now

php/libraries/User.class.inc
```diff
@@ -557,5 +557,28 @@ class User extends UserPermissions
...
+     *
+     * @return bool
+     */
+    public function isAccessibleBy(\User $user): bool
+    {
+        $centers = $user->getCenterIDs();
+
+        $projMatch   = !empty(
+            array_intersect(
+                $this->getProjectID(),
```
driusan, line 575: This doesn't look like it's defined?

johnsaigle: Yeah you're right. I'll update this.

johnsaigle:
@driusan Should be fixed now

php/libraries/User.class.inc
```diff
@@ -557,5 +557,33 @@ class User extends UserPermissions
...
+     * Implements the AccessibleResource interface. A user can
+     * be accessed by another user if they have a project and a site in common.
+     *
+     * @param \User $user The "editing" user trying to access this user object.
+     *
+     * @return bool
+     */
+    public function isAccessibleBy(\User $user): bool
+    {
+        $centers = $user->getCenterIDs();
```
jesscall, line 571: this isn't being used anywhere

johnsaigle, line 571:
```suggestion
```

PR #6182, opened 2020-03-24
[Core] Add new FilesDownloadHandler class

johnsaigle:
## Brief summary of changes

Adds a class to be used for generic file downloading throughout LORIS. This serves as the counterpart to the generic file uploader from #4181.

1. Download a file from the document repository.

This is a remake of #3881

### DIFF
--- a/modules/document_repository/php/files.class.inc
+++ b/modules/document_repository/php/files.class.inc
@@ -1,6 +1,9 @@
 <?php
+namespace LORIS\document_repository;
+use \Psr\Http\Message\ServerRequestInterface;
+use \Psr\Http\Message\ResponseInterface;
+
 /**
- * It includes two functions.
  * Post method handles updating a file.
  * Delete method handles deleting a file.
  * Get method handles getting a file.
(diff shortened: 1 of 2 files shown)


xlecours:
@johnsaigle this is looking good. I see a lot of exception throwing but there is no try-catch. Is there a higher level try-catch in LORIS to handle any thrown exception?

johnsaigle:
@xlecours The exceptions in the constructor make sense to me because they indicate that LORIS is configured incorrectly. I'll change the empty filename exception to return a 400 response instead. 👍

johnsaigle:
@xlecours This is ready for review

php/libraries/FilesDownloadHandler.php
```diff
@@ -0,0 +1,114 @@
...
+                )
+            );
+        }
+        $this->downloadDirectory = $downloadDirectory;
+    }
+
+    /**
+     * Given an HTTP request, serve the file to the client.
+     *
+     * All files uploaded will get the same permissions.
```
xlecours, line 58: Unless I am missing something, I don't think this comment applies to this function. This is a downloader, not an uploader.

modules/document_repository/php/files.class.inc
```diff
@@ -1,37 +1,17 @@
...
 namespace LORIS\document_repository;
 use \Psr\Http\Message\ServerRequestInterface;
 use \Psr\Http\Message\ResponseInterface;
 
 /**
- * It includes two functions.
  * Post method handles updating a file.
  * Delete method handles deleting a file.
+ * Get method handles getting a file.
+ *            (Files/id) will return file's infomation
```
xlecours, line 10: I don't see this use case being handled. (Files/id)

johnsaigle:
@xlecours Would you mind testing?

PR #6188, opened 2020-03-26
[API] Remove -dev suffix from 0.0.3 version of API

driusan:
There has been a significant amount of changes and endpoints
added since 0.0.2. This removes the `-dev` suffix to officially
release 0.0.3 since it's for the most part what people are using
anyways.

This PR is mostly the result of the command `git grep '0\.0\.3\-dev' | cut -f 1 -d':' | uniq | xargs sed -i -e 's/0\.\0\.3\-dev/0.0.3/g'`


christinerogers:
added this to the Release notes since it didn't make the changelog yet.

PR #6189, opened 2020-03-26
[candidate_profile] Improve gridyness of grid

driusan:
This attempts to improve the layout of the CSS grid used by the `candidate_profile` page and after testing the page as a whole, rather than on an individual card basis.

- Height (in terms of number of grid rows spanned) of a card is now calculated dynamically, rather than needing to be heuristically guessed by the module registering the widget. 1 row = 100px. 
- Cards that have a width of "1" are always placed in the left column, until there are no more "large" (> 2) cards left, after which they start filling in underneath the other columns, so that column sizes the sizes are relatively balanced.

Combined, this improves the number of gaps when there are multiple cards added to the dashboard.

Before: 
Screenshot with big gaps

After:
More balanced screenshot

### DIFF
--- a/jsx/CSSGrid.js
+++ b/jsx/CSSGrid.js
@@ -1,4 +1,5 @@
 import Card from 'Card';
+import React, {useState, useEffect, useRef} from 'react';
 
 /**
  * Create a three column grid of cards using a CSS grid.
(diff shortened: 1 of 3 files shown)


maltheism:
I'll test this PR because of the word gridyness.

jsx/CSSGrid.js
```diff
@@ -16,38 +17,108 @@ import Card from 'Card';
...
-  let cards = props.Cards.map((value, idx) => {
-    let cardID = 'card' + idx;
-
-    let style = {};
-    if (value.Width) {
-      style.gridColumnEnd = 'span ' + value.Width;
-    }
-    if (value.Height) {
-      style.gridRowEnd = 'span ' + value.Height;
+    const cardsRef = useRef(null);
```
maltheism: The `cardsRef` doesn't seem to be used.

driusan: I'm not sure what you mean.. it's used a couple places according to ctrl-f?

driusan:
@maltheism what resolution was that SS taken at?

jsx/CSSGrid.js
maltheism: @driusan never mind.

maltheism:
@driusan 1920 x 1200. I'll test other resolutions.

driusan:
That certainly doesn't look 1200px wide to me, but I'll try and look into it..

zaliqarosli:
@driusan i'm not really able to test this because there are only 3 cards for me which results in no change on the front-end for this PR

jesscall:
@driusan I'll have a look at this tomorrow and be able to provide screenshots with CCNA data

maltheism:
@driusan I had made my browser window fairly slim in those images.

jesscall:
Here are some screenshots:
In some cases I zoomed out a little for purposes of capturing all cards in one image, because I'm not sure how to take a scrolling screenshot on Mac. However, card placement in zoomed out instances is the same as when at 100%. 

Screen Shot 2020-04-02 at 12 03 32 PM

Screen Shot 2020-04-02 at 12 04 24 PM

Screen Shot 2020-04-02 at 12 05 03 PM


Screen Shot 2020-04-02 at 12 05 45 PM


Let me know if you want other examples of specific cases.

driusan:
@jesscall so is that better/an approval or could it still use work (other than what @maltheism pointed out..)?

driusan:
@maltheism I think I fixed your issue.. I wasn't able to reproduce it with that card, but I was getting similar behaviour with other cards when I resized the window that should now overflow with a scrollbar properly after the last commit.

driusan:
@jesscall / @maltheism ping

maltheism:
@driusan the issue I saw is now fixed!

PR #6191, opened 2020-03-26
[users] Add 'request account date' to users

cmadjar:
## Brief summary of changes

This adds a new `account_request_date` field to the `users` table to register information about when a given user requested an account.

#### Testing instructions

1. source the SQL patch attached to the PR
2. test that requesting a new account (on the login page) inserts a new user with the date properly filled
3. test that the 'Account Request Date' is properly displayed in the menu filter page of user accounts
4. check that the 'Account Request Date' is properly displayed after the 'Active from' and 'Active to' fields in the edit user page of user accounts.

* I don't believe an issue was created for that but this came up with Open PREVENT-AD where they wanted to have information about when a user requested access

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ changes in the following format: PR #1234***
 #### Updates and Improvements
 - Module-specific permissions added for Survey Accounts, Imaging Behavioural
 Quality Control, and Behavioural Quality Control. (PR #6041)
+- Addition of a new `account_request_date` in `users` table that will be used when
+requesting a new account and will be displayed in the User Accounts module (PR #6191)
 #### Bug Fixes
 - *Add item here*
 ### Modules
@@ -44,8 +46,6 @@ database (PR #5260)
 - New documentation for file permissions has been added to the README.md file. (PR #5323)
 - Dashboard study progression section performance improvement (PR #5887)
 documentation for file permissions has been added to the README.md file (PR #5323)
-- Addition of a new `account_request_date` in `users` table that will be used when
-requesting a new account and will be displayed in the User Accounts module (PR #6191)
 
 #### Bug Fixes
 - Fix edge-case that gave a confusing error message when changing password (PR #5956)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,9 @@ Quality Control, and Behavioural Quality Control. (PR #6041)
 database (PR #5260)
 - New documentation for file permissions has been added to the README.md file. (PR #5323)
 - Dashboard study progression section performance improvement (PR #5887)
+documentation for file permissions has been added to the README.md file (PR #5323)
+- Addition of a new `account_request_date` in `users` table that will be used when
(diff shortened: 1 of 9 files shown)


cmadjar:
@ridz1208 @PapillonMcGill could you review this PR? I'll update the RB dataset after approval that the database changes are fine with you :). Thanks!

SQL/New_patches/2020-03-26_add_account_request_date_to_users_table.sql
```diff
@@ -0,0 +1,14 @@
+-- add a new column to the users table to store the date at which
+-- the user requested an account
+ALTER TABLE users
+  ADD COLUMN account_request_date DATE DEFAULT NULL;
+
+-- determine the account_request_date based on what is present in the history
+-- table for when the user was inserted the first time in the history table
+UPDATE users u, history h
```
cmadjar: @ridz1208 is the assumption that all past users requested an account and therefore the date could be derived from the history table correct?

This was the case for Open PREVENT-AD but it might not necessarily be the case for all....

cmadjar:
In the end, there is no need for raisin bread changes in this issue since the only user is admin and it does not have any request date associated to it. Therefore, I removed the RB labels.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ changes in the following format: PR #1234***
 - Very old instruments relying on QuickForm may have issues due to code changes (PR #4928)
 - Unix user permissions have been updated which may affect access to files. New
 documentation for file permissions has been added to the README.md file (PR #5323)
+- Addition of a new `account_request_date` in `users` table that will be used when
+requesting a new account and will be displayed in the User Accounts module (PR #6191)
```
driusan, line 20: This is in the wrong section of the CHANGELOG. It's not going into the 23.0 release..

driusan:
@cmadjar can you rebase this?

cmadjar:
Fascinating. I did not even remember sending that, lol 
@driusan rebased now. Thanks for tagging me!

driusan:
@cmadjar the changelog conflict wasn't resolved correctly.. it should be in the 24.0 part of the changelog, not the 23.

CHANGELOG.md
cmadjar: oooooooooh, oupsie... 

@driusan should be fixed in the last commit. Sorry about that...

CHANGELOG.md
```diff
@@ -43,6 +45,7 @@ Quality Control, and Behavioural Quality Control. (PR #6041)
 database (PR #5260)
 - New documentation for file permissions has been added to the README.md file. (PR #5323)
 - Dashboard study progression section performance improvement (PR #5887)
+documentation for file permissions has been added to the README.md file (PR #5323)
```
driusan, line 48: I don't think this line is supposed to be here.. it's a duplicate of 2 lines above it (but the formating is also messed u..)

cmadjar: good catch! I removed the line. Sorry about that...

PR #6193, opened 2020-03-29
Candidate GUIDs

ridz1208:
***NOTE:** I have made these changes based on LORIS 21 for some specific projects requiring that code. I only opened this PR to see if there is potential in it being in the core software eventually*

This PR adds the capability to generate Global Unique Identifiers (GUIDs) using candidate personally Identifying Information (PII) without ever submitting ths PII data to the server. The requirement for PII is project specific which allows projects within a LORIS instance to adopt or not the GUIDs. only the hash is submitted and stored in LORIS and never displayed on the front end. the GUID is generated using a specific set of data pre-established by the open science initiative and currently in use. Part of the code in this PR was written by other developers (externals and @johnsaigle) and consolidated here in this PR.

This also allows projects to setup and external server to store the PII which is discarded after the hash is generated. Communication between LORIS and the external server would be done using REST calls. This is optional, it is only used if all the necessary configurations are in place. This is also a requirement for sever live instances of LORIS.

here is the logical workflow I followed
1. Load module with default fields.
2. on selection of project, display GUID fields if project requires them
3. once data is entered, JS generates the hash and sends it to PHP
4. PHP takes the hash and checks if PII system is available
   1. If PII system set up, get token
5. Proceed to saving the data to LORIS with the hash included (on failure, exit and return error)
6. Return success message to JS
   1. if PII system available return token with response
7. JS: if token available submit PII info to PII system 
8. display success message

TODO if we want this for core.
1. improve on the https://github.com/aces/open_science_identity repo to make code more reusable and easier to integrate
2. clean-up JS code and make sure REACT better practices are followed
3. thorough security check to make sure no data is leaked anywhere
4. thorough hashing check to make sure the algorithms are functioning properly
5. add "unknown" options for PII fields
6. improve on configurations require for communication with PII systems (currently using a key for access)


modules/new_profile/js/tmp_hash.js
```diff
@@ -0,0 +1,85 @@
...
+    var id = new OpenScienceIdentity({
+      first_name: first_name,
+      middle_name: middle_name,
+      last_name: last_name,
+      city_of_birth: city_of_birth,
+      birth_day: date_of_birth,
+      gender: gender
+    });
+    try {
+      var guid = id.toSignature();
```
driusan: Unless I'm missing something this is just a hash of personal information, not a guid.

ridz1208: yes, we've just been calling it GUID until now

johnsaigle: Yeah GUID has been thrown around as an unfortunate misnomer. The idea is to create a unique hash based on a minimal set of PII. It has nothing to do with being a GUID in the sense that it's used in databasing.

modules/new_profile/jsx/NewProfileIndex.js
```diff
@@ -82,14 +84,46 @@ class NewProfileIndex extends React.Component {
...
+        formData.firstName
+        && formData.middleName
+        && formData.lastName
+        && formData.placeOfBirth
+      ) {
+        let id = new OpenScienceIdentity({
+          first_name: formData.firstName,
+          middle_name: formData.middleName,
+          last_name: formData.lastName,
+          city_of_birth: formData.placeOfBirth,
```
johnsaigle: This should always be "city of birth", not "place of birth". The precision is important.

SQL/New_patches/2020-04-01-GUID.sql
```diff
@@ -0,0 +1,16 @@
+ALTER Table Project ADD COLUMN GUIDRequired binary NOT NULL DEFAULT 0;
+ALTER Table candidate ADD COLUMN GUID VARCHAR(100) DEFAULT NULL;
+
+INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber)
+SELECT 'piiURL', 'Address of server hosting the personally identifying information', 1, 0, 'text', MAX(Parent), 'PII URL', MAX(OrderNumber)+1
```
johnsaigle: I don't think any server is supposed to have any kind of PII unless the design has changed radically since I was last involved. The point of this whole thing is that PII never gets stored (or even transferred over HTTP) to preserve privacy.

For this reason I'd avoid using PII in any comments or variable names because they're always going to seem dangerous.

modules/new_profile/jsx/NewProfileIndex.js
```diff
@@ -196,6 +265,41 @@ class NewProfileIndex extends React.Component {
...
+          <TextboxElement
+            name = "lastName"
+            label = "Last Name"
+            onUserInput = {this.setFormData}
+            value = {this.state.formData.lastName}
+            required = {true}
+          />
+          <TextboxElement
+            name = "placeOfBirth"
+            label = "Place of Birth (City, Municipality)"
```
johnsaigle: I'd probably just stick with city here. I don't know if "Municipality" helps clear things up.

modules/new_profile/php/new_profile.class.inc
```diff
@@ -84,7 +86,8 @@ class New_Profile extends \NDB_Form
             $values['edcDate'] ?? null,
             $values['sex'],
             $values['pscid'] ?? null,
-            isset($values['project']) ? intval($values['project']) : null
+            isset($values['project']) ? intval($values['project']) : null,
+            isset($values['GUID']) ? $values['GUID'] : null
```
johnsaigle: Null coalesce instead?

modules/new_profile/php/new_profile.class.inc
```diff
@@ -322,6 +336,74 @@ class New_Profile extends \NDB_Form
...
+    /**
+     * Get authentication token from PII system using key stored in DB
+     *
+     * @return string The authentication token
+     *
+     * @throws \LorisException
+     */
+    private function _getPIIToken() : string
+    {
+        $config = \NDB_Config::singleton();
```
johnsaigle: Factory

php/libraries/Project.class.inc
```diff
@@ -81,6 +83,7 @@ class Project
             $project->_projectId         = (int) $projectData['projectId'];
             $project->_projectName       = $projectData['name'];
             $project->_recruitmentTarget = (integer) $projectData['recTarget'];
+            $project->_guid = $projectData['GUIDRequired'] === 1 ? true : false;
```
johnsaigle:
```suggestion
            $project->_guid = $projectData['GUIDRequired'] === 1;
```

Logically equivalent.

php/libraries/Project.class.inc
```diff
@@ -104,7 +108,8 @@ class Project
      */
     static public function createNew(
         string $projectName,
-        ?int $recruitmentTarget = null
+        ?int $recruitmentTarget = null,
+        ?bool $guidRequired = false
```
johnsaigle: When would this be null?

php/libraries/Project.class.inc
```diff
@@ -186,6 +192,46 @@ class Project
...
+    {
+        $this->_guid = $guidRequired;
+    }
+
+    /**
+     * Get list of projects requiring the GUID for their candidates
+     *
+     * @return array Array of projects using the GUID system
+     *
+     * @throws DatabaseException
```
johnsaigle:
```suggestion
```

php/libraries/Project.class.inc
```diff
@@ -186,6 +192,46 @@ class Project
...
+     */
+    public static function getGUIDProjects()
```
johnsaigle:
```suggestion
    public static function getGUIDProjects(): ?array
```

This should match pselectCol() in recent versions

johnsaigle:
I'm confused about the token in general. Can you explain what it's meant to do? It appears in this code as both a "csrf token" as well as an API authentication token -- it's neither of those though. It's a hash of candidate input. I don't think it should do double duties as an authentication or "csrf" token as both of those use cases are very different.

SQL/New_patches/2020-04-01-GUID.sql
```diff
@@ -0,0 +1,16 @@
+ALTER Table Project ADD COLUMN GUIDRequired binary NOT NULL DEFAULT 0;
```
johnsaigle: I think you need to add these changes to RB. Travis is choking on it.

ridz1208: @john asmentioned above this PR was not meant to be merged but instead was meant as a proposal.

several things have to be done to make it meargeable including putting it on the correct branch

php/libraries/Project.class.inc
ridz1208: if it's not passed in the creatnew function call (ie so I dont need to go modify every createNew call in the codebase

modules/new_profile/jsx/NewProfileIndex.js
ridz1208: hmmm I copied this from openscience, have they been using it wrong ? what are the impacts.

we can take this offline for this one

modules/new_profile/jsx/NewProfileIndex.js
ridz1208: hmmm I copied this from openscience, have they been using it wrong ? what are the impacts.

we can take this offline for this one

modules/new_profile/js/tmp_hash.js
ridz1208: I can call it UID if it makes you guys feel better.

php/libraries/Project.class.inc
johnsaigle: I don't think you need to do that if you have a default parameter. Existing calls will just default to `false` in this case.

SQL/New_patches/2020-04-01-GUID.sql
johnsaigle: OK cool. In general I am in favour of adding this, but it should be done as a dependency as I mentioned in another comment.

The "token" stuff should probably be removed because I don't think it's done properly here. We can figure out how to connect to external servers in another way, probably using Guzzle or the API instead of the way it's implemented now.

ridz1208:
@johnsaigle I agree we should find a way to use the open_science_code without ovewriting it and that might require some changes on the repo itself. thats one of the things I would like to discuss here. what's the best way of doing that ?

as for the token I mistakenly committed the `tmp_hash` file. that was meant to be deleted. other then that file there should not be any reference to csfr. that was the old code.

The token is a key with LORIS uses to authenticate with the external system keeping track of PII. Realistically projects will want the PII stored in a different location and not completely disguarded... that's what has been done for at least 3 projects until now and if we can control the security on it I think is better then having it done ad-hoc by projects and independent contractors.

php/libraries/Project.class.inc
ridz1208: oh then nvm I'll remove it

johnsaigle:
I think the best way would be to add it via `package.json` like other JS dependencies we have. We might be able to just list `aces/open_science_identity` and go from there. I'm not super familiar with how this is done but there's no reason why we shouldn't be able to include it this way afaik.

What kind of server though? I don't think you can just POST `api_token` to some URL and expect that to work.
I'm not opposed to the feature. This is something we may need to do in a more generic way though. Like make a class or interface that does the network request stuff and child classes will define the format of specific headers etc. all using PSR7 stuff. I'm happy to help with the design and implementation if you want.

Another thing to note is the `file_get_contents()` often won't work cross-domain because it relies on a  `php.ini` flag to work and it's often disabled because it's a big security issue. We can use `Guzzle` to do requests instead. We also need to force TLS on these connections since they have private information.

ridz1208:
@johnsaigle your suggestions sound good. At this stage I'll wait for confirmation to know if we want this for loris or not.

If yes we can start that up right after the release

PR #6196, opened 2020-03-30
[Install] Prevent unbound variable error

### LINKED ISSUE #6185: [tools/install.sh] Error in install process on CentOS: "projectname: unbound variable"
**Describe the bug**
On CentOS i'm trying to run the install.sh script in tools and it fails because of the error:
```
[RuntimeException]                                                                     
  Source directory /var/www/Loris/vendor/facebook/webdriver has unpushed changes on the  
   current branch:                                                                       
  M	lib/WebDriverTargetLocator.php    
```

Also worth noting that the `composer update` command gives this output:
```
Package facebook/webdriver is abandoned, you should avoid using it. Use php-webdriver/webdriver instead.
Package zendframework/zend-diactoros is abandoned, you should avoid using it. Use laminas/laminas-diactoros instead.
Package phpunit/phpunit-mock-objects is abandoned, you should avoid using it. No replacement was suggested.

```

johnsaigle:
## Brief summary of changes

This PR fixes an issue where a variable was used in the install script without being declared. It caused a bug during CentOS installation.

1. On CentOS, run the install script.

* Resolves #6185

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -160,6 +160,7 @@ if [ -d logs ]; then
 fi
 
 echo ""
+export projectname=""
 
 if [[ " ${debian[*]} " == *" $os_distro "* ]]; then
 echo "Ubuntu distribution detected."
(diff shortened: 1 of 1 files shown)


johnsaigle:
@ridz1208 @driusan I marked this as critical because I think that currently install will fail on CentOS. Feel free to remove it.

johnsaigle:
@maltheism Can you test and review since you reported the error?

PR #6216, opened 2020-04-01
Update RB migration file and database

ridz1208:
## Brief summary of changes
- Updated the RB migration file to know which patches were applied on the data set
- ran missing patches on RB


raisinbread/RB_files/RB_Config.sql
```diff
@@ -19,7 +19,7 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (16,20,'false');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (17,21,'true');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (18,22,'true');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (19,23,'0');
-INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (20,24,'Modify this to your project\'s citation policy');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (20,24,"Modify this to your project\'s citation policy");
```
maltheism, line 22: Why is there a slash `project\'s`

ridz1208: because phpstorm hates me

raisinbread/RB_files/RB_Config.sql
```diff
@@ -34,6 +34,7 @@ INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (34,38,'/data/uploads/')
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (35,40,'main.css');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (36,41,'25');
 INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (37,44,'localhost');
+INSERT INTO `Config` (`ID`, `ConfigID`, `Value`) VALUES (38,45,'http://localhost');
```
maltheism: Should the `http://` be necessary? I'm not sure what the uses is for this row.

raisinbread/RB_files/RB_Config.sql
ridz1208: undid the change

raisinbread/RB_files/RB_Config.sql
ridz1208: yes
thats what was already there i just restored it because all tests failed when URL was removed

maltheism: @ridz1208 I mean the http part. What if the developer setups https instead. I would think just localhost would work? Unless the value is being used for something specific but I don't know.

ridz1208: afaik its being used for tests and any dev should change this value eitherway and if you try to use http but your SSL is setup it wont even load

you can ignore that for now

raisinbread/RB_files/RB_Config.sql
driusan: there should be a `\` in `project\'s` because `'` is the delimiter that the SQL standard uses for strings.

maltheism: @ridz1208 then it should be single not double quotes.

ridz1208: IT ISSSS

maltheism: @ridz1208 I see `"Modify this to your project\'s citation policy"` not `'Modify this to your project\'s citation policy'`

ridz1208: refresh

maltheism:
I can manually test after it passes travis.

Issue #6219, opened 2020-04-01
[instrument battery] imaging data button


lingma:
**Describe the bug**
See the image:
![image](https://user-images.githubusercontent.com/5295785/78161656-ccf48900-7413-11ea-850d-16b561f7cce1.png)

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. click on the candID 587630, V2
![image](https://user-images.githubusercontent.com/5295785/78161551-9f0f4480-7413-11ea-9ccc-e8f4b369bdd7.png)

2. Click on 'View Imaging data"'
3. See error 502

**What did you expect to happen?**
If error, say what happened.

**Server Environment (if known):**
- LORIS Version:23

lingma:
This behavior comes probably from the fact that some instruments are not in the list, this behaviour could probably be fixed in the imaging browser module

lingma:
I copied the following text from PR#6221:

The Imaging Browser is set to be anywhere at instrument list, at the same time if some instruments are not present, then the Imaging Browser module will report that fact as an exception, clearly the overall design logic is not correct somewhere.

lingma:
A new commit has been submitted, this will solve the link status 500 error.

johnsaigle:
This seems to be resolved based on the linked PRs

PR #6221, opened 2020-04-01
[imaging_browser] Imaging browser exception

lingma:
If some instruments are not present, the imaging browser will throw an exception, I feel that this behavour is not necessary.

Issue #6219

### DIFF
--- a/modules/imaging_browser/templates/imaging_session_controlpanel.tpl
+++ b/modules/imaging_browser/templates/imaging_session_controlpanel.tpl
@@ -30,6 +30,7 @@
        <input id="bbonly" type="button" class="btn btn-volume-viewer" accesskey="d" value="3D Only">
        <input id="bboverlay" type="button" class="btn btn-volume-viewer" accesskey="c" value="3D + Overlay">
 
+{if $subject.links|@count || $subject.tarchiveIDLoc|@count || $mantis}
     <h3>Links</h3>
     <ul>
         {foreach from=$subject.links item=link}
@@ -46,6 +47,7 @@
             <li><a target="mantis" href="{$issue_tracker_url}">Report a Bug (Mantis)</a></li>
         {/if}
     </ul>
+{/if}
 
     <h3>Visit Level QC</h3>
     <div class="visit-level-feedback">
@@ -57,9 +59,7 @@
                  </span>
         </a>
      </div>
-
-   <br>
-
+    <br>
     <div class='div-controlpanel-bottom'>
         <div class="form-group">
             <label>QC Status</label>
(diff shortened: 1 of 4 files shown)


johnsaigle:
Duplicate of #6034

lingma:
This is a little different from #6034, but they are related issues. Probably we need a combined PR to solve all those issues.

johnsaigle:
Yeah I'm sorry I made a mistake. I deleted by comment but forgot to undo the GitHub marking.

The PR #6076 should stop LORIS from crashing when exceptions are not caught.

lingma:
Imaging Browser is set to be anywhere, at the same time if some instruments are not present, then the module will report as an exception, clearly the overall design logic is not correct some where.

lingma:
Don't merge or test this PR, the solution is too simple and not good enough.

lingma:
Also changed:
Before this PR:
image

After:
image

nicolasbrossard:
@lingma With your PR, there is now an error message displayed in the left-hand side but all the other sections in the left-hand side have disappeared (Navigation, Volume Viewer, Visit Level Qc, in red on the screenshot). I think these should be displayed even though there are no linked instruments for that candidate/visit. You should also leave the links to the DICOM archive and the 'Download DICOM' button active in the 'Links' section (in yellow in the screenshot).

Screen Shot 2020-04-09 at 10 25 59 AM

modules/imaging_browser/templates/imaging_session_controlpanel.tpl
```diff
@@ -1,3 +1,7 @@
+{if $subject.error }
```
nicolasbrossard, line 1: You could still display parts of the panel here, even though `if $subject.error` is true

lingma: The left side link will only be displayed if there is anything to display. The QC part has been removed if the session is not configured to avoid to do quality control on a not configured session. Please check if this is the desired effect.

lingma:
I am confused, if the instrument is not configured, there is nothing to display. If there is a download link, then yes, the link will be displayed. Otherwise I display an error message to say that there is nothing which is the original message. is that the desired effect?

nicolasbrossard:
A session can have one or more instruments associated to it. I think your PR should only determine if the `MRI parameter form` link should be displayed or not, based on the presence (or absence) of that instrument in that session. All the other elements in the left-hand side panel (Navigation, Volume Viewer, Visit Level Qc) should still be displayed even though the MRI Parameter Form might be missing. @cmadjar What do you think?

lingma:
This should be good now. Please check. @nicolasbrossard 

If the instruments "MRI_parameter_form and Radiology_review" not found, then the system in principle will continue like nothing has happened. The initial exception throwing is removed.

The little difference in the template is: if the no link exists, then the block won't be displayed.

nicolasbrossard:
It now works whether there are instruments to link to or not. Approving! 👍

PR #6223, opened 2020-04-01
[Instruments] Bugfix multi-escaped special characters

ridz1208:
## Brief summary of changes
Removed duplicate call to `HTMLSpecialChars()` causing double escaping on any instrument field with one of the following characters `& < > "`

The removal is justified since these fields are being escaped directly in the database class
https://github.com/aces/Loris/blob/master/php/libraries/Database.class.inc#L538

### DIFF
--- a/php/libraries/LorisForm.class.inc
+++ b/php/libraries/LorisForm.class.inc
@@ -695,10 +695,6 @@ class LorisForm
                 }
             }
         }
-        // Always sanitize user-controlled input
-        if (!is_array($newValue)) {
-            $newValue = htmlspecialchars($newValue);
-        }
 
         return $newValue;
     }


ridz1208:
related to https://github.com/aces/Loris/pull/6158

maltheism:
@ridz1208 I'm not a security wizard but I think everywhere that _getFilteredValue used & returning some data. It would have to be verified that removing the sanitization doesn't void potential security.
So far I see `getFilteredValue` used only in LorisForm.class.inc on lines 578, 584, 604, 607, 615, 667, 689. I'm guessing all those functions that call it may return something that needs to be verified that removing the check doesn't add a vulnerability.

ridz1208:
@maltheism 
https://github.com/aces/Loris/pull/6158#pullrequestreview-376195755

johnsaigle:
Copying what I wrote in #6158 for clarity and easier reference:

>After doing some research in earlier PRs and some basic testing, I think we can go ahead with this fix:
The htmlspecialchars() line was added in an earlier phase of LORIS. Right now, the extensive use of strict typing prevents many injection points that this previously addressed. Additionally, the use of ReactJS provides sanitization of output by default on many data points. Finally, HTMLEscapeArray in the DB class ensures that almost all data going into the database is escaped.
Removing this line may slightly increase the risk of XSS attacks but I think the risk is small and outweighed by the potential for data loss that it creates. It was always a bit of a hack and I think it's one that we can probably do without now.

PR #6253, opened 2020-04-04
[UI] Fix navigation bar

### LINKED ISSUE #6235: [UI] on mobile devices once a menu tab is open it cannot be closed
**Describe the bug**
When one opens any of the menus, there is no way to re-close it to select another menu item.

Screenshot_20200402-130930_Chrome

haoweiqiu:
## Brief summary of changes
I think this should fix the problem with the navigation bar.
I deleted `.navbar-default .navbar-nav > li > a:focus` since this was the cause of the problem, also, I modified `background-color` in the `bootstrap.min.css` from grey to blue.

Haowei Qiu, GSOC Applicant
CC @christinerogers 

* Resolves #6235  (Reference the issue this fixes, if any.)


christinerogers:
Thanks @haoweiqiu  - Can you describe what the problem/fix was, in one line at the top of your PR description?  Would help to have that handy up top here.

Tagging @kongtiaowang to review it.

haoweiqiu:
@maltheism thanks for your comment! This is weird since I don't have this problem on my laptop. I will look into this now.

maltheism:
> I tried to work on modification for focus by changing the color from white to blue but this would cause problem with hover. I would need some time to find a solution for this problem.

@haoweiqiu changing the colour from white to blue would make like you said everything be blue when the browser action is triggered. I'm suggesting the css for focus be independently set to blue and for doing that the CSS of the existing code would need the `focus` moved out of the group of other actions (hover, active, ... ).

laemtl:
Replaced by #6235

Issue #6209, opened 2020-03-31
ReadTheDocs build failure on LORIS API symlinks


johnsaigle:
**Describe the bug**
ReadTheDocs is failing because it's unable to resolve the following symlinks:

docs/wiki/99\ -\ Developers/LORIS-REST-API-0.0.2.md
docs/wiki/99\ -\ Developers/LORIS-REST-API-0.0.3-dev.md

These were added in https://github.com/aces/Loris/pull/6151

**What did you expect to happen?**
ReadTheDocs and mkdocs should resolve the symlinks and render the documentation.

We may need to modify these to include relative path characters as RTD is served from `docs/` rather than the LORIS root. 

**Additional context**
ReadTheDocs build error:

```
INFO    -  Cleaning site directory 
INFO    -  Building documentation to directory: /home/docs/checkouts/readthedocs.org/user_builds/acesloris/checkouts/latest/_build/html 
INFO    -  The following pages exist in the docs directory, but are not included in the "nav" configuration:
  - React.README.md
  - SQLModelingStandard.md
  - deprecated_wiki/About-superuser.md
  - deprecated_wiki/CentOS-Imaging-installation-transcript.md
  - deprecated_wiki/Code-Customization.md
  - deprecated_wiki/Developer's-Instrument-Guide.md
  - deprecated_wiki/Getting-the-Release.md
  - deprecated_wiki/Guide-to-Loris-React-components.md
  - deprecated_wiki/How-to-Code-an-Instrument.md
  - deprecated_wiki/How-to-make-a-LORIS-module.md
  - deprecated_wiki/Installing-Loris-in-Brief.md
  - deprecated_wiki/Instrument-Groups.md
  - deprecated_wiki/Instrument-Insertion.md
  - deprecated_wiki/Instrument-Scoring.md
  - deprecated_wiki/Instrument-Scripts.md
  - deprecated_wiki/Instrument-Testing-and-Troubleshooting.md
  - deprecated_wiki/LORIS-Dictionary.md
  - deprecated_wiki/LORIS-Form.md
  - deprecated_wiki/LORIS-Module-Testing.md
  - deprecated_wiki/LORIS-Modules.md
  - deprecated_wiki/LORIS-scripts-in-the-tools--directory.md
  - deprecated_wiki/Notification-system.md
  - deprecated_wiki/Open-LORIS.md
  - deprecated_wiki/Other-Imaging-Scripts.md
[...]

PR #6192, opened 2020-03-27
minor fix on the link to the unit test guide

haoweiqiu:
The change is made to the markdown file. This fix a minor problem with the link to the unit test guide.

Haowei Qiu, Google Summer of Code Applicant
CC @christinerogers

PR #6255, opened 2020-04-04
[Documentation] fixing broken link

### LINKED ISSUE #6254: Broken Links in documentation
**Describe the bug**
The link for the unit testing guide in [testing guide](https://github.com/aces/Loris/blob/master/docs/wiki/99%20-%20Developers/Automated%20Testing.md) is broken

**To Reproduce**
1. Go to `CONTRIBUTING.md`
2. Click on testing guide link given in section **Some things to keep in mind**
3. See error

Unit testing guide was supposed to open

**Additional information:**
I am working on this issue

ridhishjain:
## Brief summary of changes

- [:+1: ] Have you updated related documentation?

* Resolves #6254

### DIFF
--- a/docs/wiki/99 - Developers/Automated Testing.md
+++ b/docs/wiki/99 - Developers/Automated Testing.md
@@ -91,7 +91,7 @@ We use Docker to help us simulate a server running LORIS. The installation and u
 
 LORIS uses the [PHPUnit library](https://phpunit.de/) for unit tests.
 
-Unit testing is covered in depth in our [Unit Test Guide](../../../test/UnitTestGuide.md).
+Unit testing is covered in depth in our [Unit Test Guide](https://github.com/aces/Loris/blob/master/test/UnitTestGuide.md).
 
 Unit test files can be found in the folder `test/unittests/`.
 


docs/wiki/99 - Developers/Automated Testing.md
```diff
@@ -91,7 +91,7 @@ We use Docker to help us simulate a server running LORIS. The installation and u
 
 LORIS uses the [PHPUnit library](https://phpunit.de/) for unit tests.
 
-Unit testing is covered in depth in our [Unit Test Guide](../../../test/UnitTestGuide.md).
+Unit testing is covered in depth in our [Unit Test Guide](https://github.com/aces/Loris/blob/master/test/UnitTestGuide.md).
```
driusan, line 94: This should remain a relative link, what was wrong with the old one?

johnsaigle: I think that ReadTheDocs may not be able to travel outside of its root directory. I think this is the related to the issue #6209. We may have to put all of our documentation under `docs/wiki/` in order to serve it from ReadTheDocs.

haoweiqiu: @johnsaigle Yes, ReadTheDocs is giving warning with all files outside the `docs/wiki` and I am not sure if using symlinks would pass ReadTheDocs since the log always tells me that the files could not be found for #6209, but I could move the documentation back to  `docs/wiki`.

ridhishjain: > This should remain a relative link, what was wrong with the old one?
It wasn't working, I also pointed it [here](https://neurostars.org/t/gsoc-2020-project-idea-24-loris-automated-testing/5763/7)

johnsaigle: I think the best approach here would be to move the Unit Test Guide to the folder `docs/wiki/99 - Developers/` and add a new entry in `mkdocs.yml`.

haoweiqiu: @johnsaigle I could do that now. Could I also move the `LORIS-REST-API-0.0.2.md ` back to `docs` folder or create a copy inside the `docs/wiki` ? I believe this could resolve my PR #6276 for Issue #6209.

johnsaigle: That sounds good to me @haoweiqiu. Before you do that though let's wait for @driusan to weigh in.

haoweiqiu: @johnsaigle Sure! However, I am not sure why we need to add new entry for the Unit Test Guide in `mkdocs.yml` since `Automated Testing.md` mentioned the Unit Test Guide already. Would you please explain a bit? Thanks!

johnsaigle: My thinking is that both documents, Unit Testing Guide and Automating Testing, should be in the developer section in the ReadTheDocs wiki. I don't think there's a good reason for both of these documents to be in different locations and I think the best location is in ReadTheDocs.

That way we can avoid the symlink/relative link issue and keep everything in the `docs/wiki/...` folder which should make things more simple.

Is that more clear? let me know if I understood you correctly.

driusan: We should not be moving things out of their modules just for read the docs, if anything we should just move the read the docs config up a level so that it can access anything in the repo.

haoweiqiu: @johnsaigle Thanks I understand now.

haoweiqiu: @driusan Thanks for your comment! I would continue to find solution for #6209.

laemtl:
Resolved in #6192

Issue #5355, opened 2019-10-22
Add to documentation that HTML should not be used in instrument fields


johnsaigle:
My guess is that this is caused by your instrument fields containing HTML which is not recommended.

removing HTML and rerunning the data dictionary builder should fix the issue

_Originally posted by @ridz1208 in https://github.com/aces/Loris/issues/5092#issuecomment-524439361_

Issue #5092, opened 2019-08-23
[DQT] Rendering issue with HTML UI elements of certain instruments


SantiagoTG:
DQT - Rendering issue with HTML UI elements of certain instruments (see image of frontend DQT display issues for example)

Not IBIS-specific.

Many more UI elements have been incorporated into instruments recently to help patients/parents filling them out directly, so the issue is now really glaring for certain instruments as it's almost every field broken on the front end.

![Bug_HTML_rendering_code-frontend_2018-11-7](https://user-images.githubusercontent.com/54332302/63618736-cd45bf80-c5ba-11e9-91f1-9ec0e97f7de4.JPG)

https://redmine.cbrain.mcgill.ca/issues/15607

ridz1208:
My guess is that this is caused by your instrument fields containing HTML which is not recommended.

removing HTML and rerunning the data dictionary builder should fix the issue

johnsaigle:
Closing because this is a "won't fix" issue. Instead, this should be documented as not being supported/recommended. I've opened a ticket for this.

PR #6267, opened 2020-04-06
Add to documentation that HTML should not be used in instrument fields

mohini-tripathi:
## Brief summary of changes
- DQT - Rendering issue was faced in https://github.com/aces/Loris/issues/5092, Related clarification is documented to avoid it in the future.
* Resolves #  
this fixes https://github.com/aces/Loris/issues/5355
#GSOC Aspirant
CC @christinerogers

### DIFF
--- a/modules/dataquery/README.md
+++ b/modules/dataquery/README.md
@@ -70,7 +70,8 @@ curl -H 'Content-Type: application/json' -X POST http://$YOURCOUCHDBADMIN:$YOURC
 
 ###### Import the LORIS instrument data
 
-This step is optional and not required if only the MRI portion of LORIS is used:
+This step is optional and not required if only the MRI portion of LORIS is used: \
+(It is NOT recommended to use HTML in instrument fields)
 
 `php CouchDB_Import_Instruments.php`
 


modules/dataquery/README.md
```diff
@@ -70,7 +70,8 @@ curl -H 'Content-Type: application/json' -X POST http://$YOURCOUCHDBADMIN:$YOURC
 
 ###### Import the LORIS instrument data
 
-This step is optional and not required if only the MRI portion of LORIS is used:
+This step is optional and not required if only the MRI portion of LORIS is used: \
```
PapillonMcGill, line 73: I would prefer to indicate when it is required. Other options might be invalidated in the future.
Similar change should also be applied to the MRI section below.
```suggestion
This step is optional and is required only if instruments are used: \
```

lingma: I believe the suggestion is not the original meaning, and contradictory.  @PapillonMcGill

christinerogers: @lingma you are right that it is not the original meaning, but @PapillonMcGill 's change is also more accurate and so better

christinerogers: however not relevant to this PR which I will close.

christinerogers:
Thanks @mohini-tripathi for this suggestion -- we'd rather document where users look up how to create an instrument, so I am closing this PR.  
Thanks for your initiative here.

PR #6292, opened 2020-04-06
[DICOM Archive] Change Sex filter from text to dropdown

### LINKED ISSUE #6286: "Dicom Archive" Sex Filter doesn't have Dropdown.
**Describe the bug**
In Dicom Archive page, the sex field doesn't contain any dropdown which will contain the values Male and Female it is just having normal text field in which a user can type anything 

For Example, You can see in the screenshot that in the displayed data the sex entered is O which is a duplicate value.

**To Reproduce**
1. Go to 'Imaging Tab'
2. Click on 'Dicom Archive'
3. In Selection filter 
4. See Sex Filter

There should be a dropdown in place of a text field. So that user can select Male or Female from the dropdown.

Browser Environment (please complete the following information):

OS: Ubuntu
Browser: Chrome

Server Environment (if known):
Note: We only support the most recent release of LORIS.
LORIS Version: [e.g. 23.0.0]

Dicom

johnsaigle:
## Brief summary of changes

Changes UI for the Sex filter to be a dropdown of possible values instead of a text field.

1. Go to DICOM Archive and filter by Sex

* Resolves #6286

### DIFF
--- a/modules/dicom_archive/jsx/dicom_archive.js
+++ b/modules/dicom_archive/jsx/dicom_archive.js
@@ -129,7 +129,7 @@ class DicomArchive extends Component {
       {label: 'Sex', show: true, filter: {
         name: 'sex',
         type: 'select',
-        options: options.sex.reduce((a, b)=> (a[b]=b, a), {}),
+        options: {M: 'M', F: 'F', O: 'O'},
       }},
       {label: 'Date of Birth', show: true, filter: {
         name: 'dateOfBirth',
--- a/modules/dicom_archive/jsx/dicom_archive.js
+++ b/modules/dicom_archive/jsx/dicom_archive.js
@@ -129,7 +129,7 @@ class DicomArchive extends Component {
       {label: 'Sex', show: true, filter: {
         name: 'sex',
         type: 'select',
-        options: options.sex,
+        options: options.sex.reduce((a, b)=> (a[b]=b, a), {}),
       }},
       {label: 'Date of Birth', show: true, filter: {
         name: 'dateOfBirth',
--- a/modules/dicom_archive/jsx/dicom_archive.js
+++ b/modules/dicom_archive/jsx/dicom_archive.js
@@ -128,7 +128,8 @@ class DicomArchive extends Component {
       }},
       {label: 'Sex', show: true, filter: {
         name: 'sex',
-        type: 'text',
+        type: 'select',
+        options: options.sex,
       }},
       {label: 'Date of Birth', show: true, filter: {
         name: 'dateOfBirth',
--- a/modules/dicom_archive/php/dicom_archive.class.inc
+++ b/modules/dicom_archive/php/dicom_archive.class.inc
@@ -75,7 +75,16 @@ class Dicom_Archive extends \DataFrameworkMenu
         $site_names   = array_values(\Utility::getSiteList());
         $site_options = array_combine($site_names, $site_names);
 
-        return [ 'sites' => $site_options ];
+        // Populate options for Sex in DICOM Archive.
+        // Note that these values differ from the ones used in Candidate
+        // as these values are extracted from imaging headers.
+        $sex_values  = ['F', 'M', 'O'];
+        $sex_options = array_combine($sex_values, $sex_values);
+
+        return [
+            'sites' => $site_options,
+            'sex'   => $sex_options,
+        ];
     }
 
     /**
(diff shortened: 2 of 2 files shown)


modules/dicom_archive/php/dicom_archive.class.inc
```diff
@@ -75,7 +75,16 @@ class Dicom_Archive extends \DataFrameworkMenu
         $site_names   = array_values(\Utility::getSiteList());
         $site_options = array_combine($site_names, $site_names);
 
-        return [ 'sites' => $site_options ];
+        // Populate options for Sex in DICOM Archive.
+        // Note that these values differ from the ones used in Candidate
+        // as these values are extracted from imaging headers.
+        $sex_values  = ['F', 'M', 'O'];
+        $sex_options = array_combine($sex_values, $sex_values);
```
driusan, line 82: There's no reason to change the PHP just to return a constant array which is created in a confusing way.

johnsaigle: It's a common idiom in LORIS at this point when creating values for a front-end dropdown.

> git grep 'array_combine'
modules/behavioural_qc/php/behavioural_qc.class.inc:        $visit_array   = array_combine($visit_array, $visit_array);
modules/dicom_archive/php/dicom_archive.class.inc:        $site_options = array_combine($site_names, $site_names);
modules/document_repository/php/document_repository.class.inc:        $instrumentsList = array_combine($instruments, $instruments);
modules/document_repository/php/files.class.inc:        $instrumentsList = array_combine($instruments, $instruments);
modules/electrophysiology_browser/php/electrophysiology_browser.class.inc:            $list_of_sites = array_combine($siteNames, $siteNames);
modules/imaging_uploader/ajax/read_log.php:    $files = array_combine($files, array_map("filemtime", $files));
modules/publication/ajax/FileUpload.php:        $currentCollabs = array_combine(
modules/publication/ajax/FileUpload.php:        $submittedCollaborators = array_combine(
modules/publication/ajax/getData.php:    $allVOIs['Behavioral'] = array_combine($bvlVOIs, $bvlVOIs);
modules/publication/ajax/getData.php:    $allVOIs['Imaging'] = array_combine($imgVOIs, $imgVOIs);
modules/publication/ajax/getData.php:    $kws = array_combine($kws, $kws);
modules/publication/ajax/getData.php:    $collabs = array_combine($collabs, $collabs);

It's used to the point where I created a generic function to do this in another PR but it didn't get merged (https://github.com/aces/Loris/pull/5328/files#diff-ebd65470c97636299b57db3ab5e1f5a8R984).

driusan: The value being returned here is constant and only being done to pass the hardcoded value `{ M: "M", F: "F", O: "O" }` to javascript. Just code the value in javascript.

maltheism: @johnsaigle the following might be helpful
```
const dropdownArray = ['a','b','c'];
const result = dropdownArray.reduce((a,b)=> (a[b]=b,a),{});
console.log(result); // {a: "a", b: "b", c: "c"}
```

driusan: @maltheism it is a hardcoded constant that will never change. There is no reason to use an array reduce. The value `{ M: "M", F: "F", O: "O" }` should just be hardcoded in the React.

maltheism:
Hi @driusan, I just pushed the changes to the PR and as requested in slack.

modules/dicom_archive/jsx/dicom_archive.js
```diff
@@ -128,7 +128,8 @@ class DicomArchive extends Component {
       }},
       {label: 'Sex', show: true, filter: {
         name: 'sex',
-        type: 'text',
+        type: 'select',
+        options: options.sex.reduce((a, b)=> (a[b]=b, a), {}),
```
driusan, line 132: This should not be coming from PHP. The values should just be coded in the React element.

johnsaigle: It feels more future-proof to do this on the front-end though. The values could change, why not do it in a functional way instead of hardcoding?

maltheism: @johnsaigle @driusan oh you want the array just made in the frontend. What if the database changes the M, F, O in the future to be Male, Female, Other. I think some of the backend should be sent over?

driusan: The dicom standard specifies that the valid values are M, F, and O. If there's a chance that they might change, it should remain a text field and this PR should just be closed.

maltheism: @driusan ah if the case is never going to change then just a frontend change is cleaner.

maltheism:
@driusan see latest changes

Issue #6323, opened 2020-04-07
[instrument list] 2 parameter mode doesn't work in the timepoint list


lingma:
**Describe the bug**
If you read the code, you can see that the program seems to support 2 parameter mode in timepoint list, but in fact this doesn't work. I am not sure whether this feature is necessary to check.

![image](https://user-images.githubusercontent.com/5295785/78621412-84443200-7850-11ea-89a4-dc466b939cc6.png)

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to 'timepoint list
2. put the session id at the end of the URL to access directly
3. See error

**What did you expect to happen?**
If the program says supported, then do it and add to the test cases, otherwise remove that part of the code.

**Server Environment (if known):**
- LORIS Version: 23.0

driusan:
Is this linked to from anywhere? If not maybe we should just remove the feature.

lingma:
This one is not working at least from 20.0, it is supposed to be to support .htaccess url rewrite, do we still need this feature? My fix could make it work, but if this feature is not necessary, we don't need further spend time to create test case

lingma:
> Is this linked to from anywhere? If not maybe we should just remove the feature.

I don't think this feature is used any where, otherwise someone should have reported as a bug

lingma:
I have removed this two parameter access from the router. I checked, found that the program was not correct to process 2 parameter mode since the version V20.1, while the V20.0 should be correct.

We can use the .htaccess to rewrite URL instead of processing by ourselves. I don't see the necessity and the test will be simple without this two parameter access mode support.

I did not close this issue for now for reference purpose.

johnsaigle:
Can this be closed now that https://github.com/aces/Loris/pull/6324 is merged?

Issue #6272, opened 2020-04-06
[create_timepoint] link problem on time point list


lingma:
**Describe the bug**
Create timepoint link on the time point list page has an extra "/"

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to 'candidate profile'
2. Click on any profile
3. Click on 'Create time point' button in the Actions zone
4. See the link itself, "v.loris.ca//create_timepoint..."

![image](https://user-images.githubusercontent.com/5295785/78569839-2c76de00-77f2-11ea-822f-e6b82f31beb7.png)

![image](https://user-images.githubusercontent.com/5295785/78569927-49131600-77f2-11ea-96a3-c6e40eb411b4.png)

**What did you expect to happen?**
An extra "/" is not expected here

**Server Environment (if known):**
- LORIS Version: 23.0

lingma:
The similar is the #6271

Issue #6271, opened 2020-04-06
[candidate_parameters] link problem on the timepoint list


lingma:
**Describe the bug**
Candidate Info link on the time point list page has an extra "/"

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to 'candidate profile'
2. Click on any profile
3. Click on 'Candidate Info' button in the Actions zone
4. See the link itself, "v.loris.ca//candidate_parameters/...."

![image](https://user-images.githubusercontent.com/5295785/78569338-76ab8f80-77f1-11ea-9d39-1913c23d6f7e.png)

![image](https://user-images.githubusercontent.com/5295785/78569627-e1f56180-77f1-11ea-9dbb-89835af2b0f1.png)

**What did you expect to happen?**
An extra "/" is not expected here

**Server Environment (if known):**
- LORIS Version: 23.0

lingma:
This is similar as the issue #6272, the origin is in the timepoint_list, don't need to be open now

PR #6324, opened 2020-04-07
Timepoint list baseurl problem fix

lingma:
## Brief summary of changes

This PR solves wrong URL problem in timepoint list at single parameter mode and the 404 error in the #6323. The PRs could be separated, but since the fix was applied to the same file instead of modifying instrument_list module, I just put the PRs together. The work were done in two steps.

Possible issue: if the value doesn't exist, the system will return an empty page with 500 error, #6325 which could be addressed in another PR.

Test directive:

Go to 'candidate profile'
Click on any profile

To test:
The issue #6271
Click on 'Create time point' button in the Actions zone
See the link itself, "v.loris.ca//create_timepoint..."

The correct one should not have double "/"

The issue #6272
Click on 'Candidate Info' button in the Actions zone
See the link itself, "v.loris.ca//candidate_parameters/...."
The correct one should not have double "/"

The issue #6323: (This issue should be considered useless, I remove this feature in the new commit)
When you click on any visit label to access the instrument list, you can see the candidate id, and the session id in the url, use the format `base url / candidate id / session id` to test the 2 parameter mode

Issue #6271, #6272 and #6323

### DIFF
--- a/src/Router/BaseRouter.php
+++ b/src/Router/BaseRouter.php
@@ -110,38 +110,22 @@ class BaseRouter extends PrefixRouter implements RequestHandlerInterface
         if (preg_match("/^([0-9]{6})$/", $components[0])) {
             // FIXME: This assumes the baseURL is under /
             $path    = $uri->getPath();
-            $baseurl = $uri->withPath("/")->withQuery("");
+            $baseurl = $uri->withPath("")->withQuery("");
 
             $factory = \NDB_Factory::singleton();
             $factory->setBaseURL($baseurl);
-            switch (count($components)) {
-                case 1:
-                    $request = $request
-                    ->withAttribute("baseurl", rtrim($baseurl->__toString(), '/'))
-                    ->withAttribute("CandID", $components[0]);
-                    $module  = \Module::factory("timepoint_list");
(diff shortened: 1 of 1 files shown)


johnsaigle:

lingma:
> Can you add testing instructions?

You can follow the issue report to do the test. By the way, I found that since version 20, the 2 parameter mode access is not possible. It is better to add some test cases to control this, I will do so when I have time.

johnsaigle:
You've linked 3 issues. Can you please copy or create clear testing instructions? When issuing a PR it's better to provide testing instructions in the description to help reviewers and to make sure different reviewers are following the exact same steps.

lingma:
Go to 'candidate profile'
Click on any profile

To test:
The issue #6271
Click on 'Create time point' button in the Actions zone
See the link itself, "v.loris.ca//create_timepoint..."

The issue #6272
Click on 'Candidate Info' button in the Actions zone
See the link itself, "v.loris.ca//candidate_parameters/...."

The issue #6323:
See the candidate id, and the session id, use the format
base url /candidate id / session id

format to test the 2 parameter mode

johnsaigle:
Thank you

PR #6330, opened 2020-04-07
[User Accounts] Disable password auto-completion on edit_user page

### LINKED ISSUE #6319: [User accounts] Usability: Proposal, not to force to change passwords every time a user info is edited.
User accounts

**Description of the issue**

It seems that every time a profile for a given user is updated (permission added/deleted, etc) the module user_accounts force the editing admin to change the password for a the edited user.

While I'm not sure if forcing this password change have some advantage from the security perspective, it does make the testing process a bit tedious when having to test different combinations of permissions and every time a new password should be filled before saving the configuration.

Although in real environments is most not likely to a user permission been changing to often I'm not sure yet if is convenient to set up a new password when this happen (assuming is a way to alert the end user something changed in its account) or to send an email to the user notifying it the configurations that were changed and which admin user did it and when.

johnsaigle:
## Brief summary of changes

The password field gets auto-completed on the edit_user page. This is a problem because the back-end expects this value to be blank unless you are changing a user's password.

What often happens is that when a user submits an unrelated account change and they get an error message saying that they must change their password.

This is symptomatic of general UI clunkiness in this module. Until it's reworked, disabling autocomplete seems like the best solution.

1. Visit `edit_user` and make sure the password fields are empty when you load the module.

I think this Resolves #6319

### DIFF
--- a/modules/my_preferences/templates/form_my_preferences.tpl
+++ b/modules/my_preferences/templates/form_my_preferences.tpl
@@ -1,5 +1,5 @@
 <br />
-<form method="post" name="edit_user" id="edit_user">
+<form method="post" name="my_preferences" id="my_preferences" autocomplete="off">
     <h3>Password Rules</h3>
       <ul>
         <li>The password must be at least 8 characters long.</li>
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -667,7 +667,7 @@ class Edit_User extends \NDB_Form
         $this->addGroup(
             $group,
             'Password_Group',
-            'Password',
+            'New Password',
             $this->_GUIDelimiter,
             false
         );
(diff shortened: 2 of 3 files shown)


ridz1208:
@johnsaigle nice !

I think this should be done for all the fields in the DQT as well
I might steal your fix

johnsaigle:
For sure, I think this will help. We should also do this on the my_preferences page.

modules/my_preferences/templates/form_my_preferences.tpl
```diff
@@ -1,5 +1,5 @@
 <br />
-<form method="post" name="edit_user" id="edit_user">
+<form method="post" name="my_preferences" id="my_preferences" autocomplete="off">
```
driusan, line 2: Why is an ID added to this element?

johnsaigle: It's not added, it's renamed from `edit_user` to `my_preferences` because those modules are now separate.

PR #6331, opened 2020-04-07
[Doc Repo] Update test plan

### LINKED ISSUE #6307: [Document Rpository] Notifications
Test Plan: 15

Cannot find: "Receive Document Repository Notifications". It is not in user account permissions nor on the module itself.

johnsaigle:
## Brief summary of changes

Updates the test plan to be more clear and more accurate.

This will be tested in further 23.0 testing rounds.

* Resolves #6307

### DIFF
--- a/modules/document_repository/test/TestPlan.md
+++ b/modules/document_repository/test/TestPlan.md
@@ -1,5 +1,7 @@
-1. User has access to document repository if they have the "View and upload files in Document Repository” permission or is the super user.
-   [Automation Testing]
+# Document Repository Test Plan
+
+1. User has access to document repository if they have the "View and upload files in Document Repository” or "Delete files in Document Repository" permission.
+   [Manual Testing]
 2. Create a category and a sub category.
    [Automation Testing]
 3. Check that the comments for a category are displayed properly as a tooltip.
@@ -28,20 +30,11 @@
    [Manual Testing]
 14. Edit a file in the repository: check that “Date Uploaded” date is updated.
     [Manual Testing]
-15. Check that the preference “Receive Document Repository Notifications” works.
-    [Manual Testing]
-16. Check that the performance is OK with large number of files.
-    [Manual Testing]
-17. Check that if you have the 'Receive document repository notification' activated you will receive an email each 
-    time one of the following event occurs:
+15. Visit the My Preferences module and enable notifications for the document_repository.
+    Make sure that you are notified for the follow changes:
        - Addition, deletion or modification of a file (by a user other than yourself)
        - Addition of a category (by a user other than yourself)
-    Also check that the www address contained in the notification email is correct.
     [Manual Testing]
-18. To upload large files, check that the following settings are in php.ini (default: /etc/php/{php_version}/apache2/php.ini):
-      session.gc_maxlifetime = 10800
-      max_input_time = 10800
-      max_execution_time = 10800
-      upload_max_filesize = 1024M
-      post_max_size = 1024M 
+16. Try uploading a large file, e.g. 1 GB. Ensure that an error message occurs
+    when the server has detected that the file is too large.
     [Manual Testing]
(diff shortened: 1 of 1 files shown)


modules/document_repository/test/TestPlan.md
```diff
@@ -1,5 +1,7 @@
-1. User has access to document repository if they have the "View and upload files in Document Repository” permission or is the super user.
-   [Automation Testing]
```
johnsaigle, line 2: This isn't in the integration tests. Only super user is checked, which is not a very useful test since super user will always have access.

modules/document_repository/test/TestPlan.md
```diff
@@ -1,5 +1,7 @@
...
+# Document Repository Test Plan
+
+1. User has access to document repository if they have the "View and upload files in Document Repository” or "Delete files in Document Repository" permission.
```
johnsaigle, line 3: Adds the delete permission which on its own should give access to the module.

modules/document_repository/test/TestPlan.md
```diff
@@ -28,20 +30,11 @@
    [Manual Testing]
 14. Edit a file in the repository: check that “Date Uploaded” date is updated.
     [Manual Testing]
-15. Check that the preference “Receive Document Repository Notifications” works.
```
johnsaigle, line 31: 15 and 17 are the same

modules/document_repository/test/TestPlan.md
```diff
@@ -28,20 +30,11 @@
...
-    [Manual Testing]
-16. Check that the performance is OK with large number of files.
```
johnsaigle, line 33: Removing this line because it's unusual; we don't ask for stress testing in other modules and with file uploads it's especially difficult/unrealistic to do this in manual tests.

modules/document_repository/test/TestPlan.md
```diff
@@ -28,20 +30,11 @@
...
-    [Manual Testing]
-17. Check that if you have the 'Receive document repository notification' activated you will receive an email each 
-    time one of the following event occurs:
+15. Visit the My Preferences module and enable notifications for the document_repository.
+    Make sure that you are notified for the follow changes:
        - Addition, deletion or modification of a file (by a user other than yourself)
        - Addition of a category (by a user other than yourself)
-    Also check that the www address contained in the notification email is correct.
     [Manual Testing]
-18. To upload large files, check that the following settings are in php.ini (default: /etc/php/{php_version}/apache2/php.ini):
```
johnsaigle, line 41: This information can be found at https://acesloris.readthedocs.io/en/latest/wiki/00%20-%20SERVER%20INSTALL%20AND%20CONFIGURATION/02%20-%20Website%20Configuration/Handling%20Large%20File%20Uploads/ instead.

modules/document_repository/test/TestPlan.md
pierre-p-s: Should it be tested automatically by a script?

modules/document_repository/test/TestPlan.md
pierre-p-s: I tested this and just having the "Delete files in Document Repository" does not seem to grant access to the module.

modules/document_repository/test/TestPlan.md
pierre-p-s: Should this still be tested but by an automated script?

modules/document_repository/test/TestPlan.md
johnsaigle: Thanks, I opened #6373 as a new issue.

modules/document_repository/test/TestPlan.md
johnsaigle: It sounds like a good idea but I don't think we have anything like that in our testing code.

modules/document_repository/test/TestPlan.md
```diff
@@ -28,20 +30,11 @@
...
-      session.gc_maxlifetime = 10800
-      max_input_time = 10800
-      max_execution_time = 10800
-      upload_max_filesize = 1024M
-      post_max_size = 1024M 
+16. Try uploading a large file, e.g. 1 GB. Ensure that an error message occurs
```
ridz1208, line 38: @johnsaigle too large is relative, this item should read `try uploading a file larger then the max upload size ...`

PR #6337, opened 2020-04-08
[candidate_profile] Add help text

### LINKED ISSUE #6206: Candidate Profile [Beta] does not have help text

driusan:
This adds some help text for the candidate_profile module. The help text is somewhat vague by necessity since it isn't context sensitive and the exact set of cards shown depends on the user's permissions and LORIS install's configurations. It only explains the visit link in the Candidate Info card (which is provided by the module itself and has minimal permissions, so should always be present.)

Fixes #6206

### DIFF
--- a/modules/candidate_profile/help/candidate_profile.md
+++ b/modules/candidate_profile/help/candidate_profile.md
@@ -1,0 +1,7 @@
+# Candidate Profile
+
+This module allows you to see a summary of a candidate's information
+across all modules. 
+
+Click on the visit link in the "Candidate Info" card in order to access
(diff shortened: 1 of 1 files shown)


johnsaigle:
>  the exact set of cards shown depends on the user's permissions and LORIS install's configurations. 

I think this sentence and a list of these permissions and configurations would make for good help text.

driusan:
The module has no knowledge of those permissions or configurations and they're meaningless to end users.

johnsaigle:
Sorry I read this too quickly and thought this was adding a README, not help text.

PR #6344, opened 2020-04-08
[Login] Enable optional reCAPTCHA on Request Account page

### LINKED ISSUE #6244: [login] Recaptcha not displaying
**Describe the bug**
The RECAPTCHA used to be available in a previous version when a user was requesting an account or even resetting a password.

I think the functionality should be restored

johnsaigle:
## Brief summary of changes

Adds a reCAPTCHA validation feature to the existing request account page. 

reCAPTCHA is only active when a private key has been configured on the server. Otherwise it will be ignored.

This will probably need to be reworked alongside #5828 in the future.

Screen Shot 2020-04-08 at 16 42 24

1. Enable reCAPTCHA by following the Google reCAPTCHA instructions for version 2.
2. Update the config settings in LORIS as described in the new wiki document
3. Try requesting an account with and without submitting a reCAPTCHA and make sure it seems to work well.

* Resolves #6244


modules/login/php/requestaccount.class.inc
```diff
@@ -33,6 +33,10 @@ class RequestAccount extends \NDB_Form
     function setup()
     {
         $this->tpl_data['page_title'] = 'Request Account';
+        $this->tpl_data['captcha_key'] = \NDB_Factory::singleton()
+             ->config()
+             ->getSetting('reCAPTCHAPublic');
+        $this->tpl_data['nonce'] = base64_encode(openssl_random_pseudo_bytes(16));
```
johnsaigle, line 39: This is required as a security measure to avoid making our Content Security Policy weaker. Detailed discussion here. https://github.com/google/recaptcha/issues/107

modules/login/php/requestaccount.class.inc
```diff
@@ -79,10 +83,36 @@ class RequestAccount extends \NDB_Form
...
+        // Only validate reCAPTCHA if a private key exists.
+        if (!empty($reCAPTCHAPrivate)) {
+            $success = false;
+            $resp    = (new \ReCaptcha\ReCaptcha($reCAPTCHAPrivate))
+                ->verify(
+                    $values['g-recaptcha-response'] ?? '',
+                    $_SERVER['REMOTE_ADDR']
+                );
+            $success = $resp->isSuccess();
+            // FIXME This leverages the error handling for the "site" text field.
```
johnsaigle, line 107: I couldn't figure out how to get an error code to show up under the reCAPTCHA part so this is a bit hack-y but is functioning. I'm open to suggestions

modules/login/php/requestaccount.class.inc
```diff
@@ -211,18 +242,6 @@ class RequestAccount extends \NDB_Form
             }
         }
 
-        // Verify reCAPTCHA
```
johnsaigle, line 214: This is moved to `_validate` instead of `_process`. It doesn't make sense to insert a user into the DB if they failed the reCAPTCHA.

modules/login/templates/form_requestaccount.tpl
```diff
@@ -61,15 +61,19 @@
...
         {if $captcha_key}
-            {* Google reCaptcha *}
-          <div class="form-group">
+{$form.captcha.html}
             <div class="g-recaptcha" data-sitekey="{$captcha_key}"></div>
-            <span id="helpBlock" class="help-block">
-              <b class="text-danger">{$error_message['captcha']}</b>
-            </span>
-          </div>
+            <script src="https://www.google.com/recaptcha/api.js?render={$captcha_key}&onload=onloadCallback nonce={$nonce}"></script>
```
johnsaigle, line 69: This should probably be refactored in a more generic way so that we can use reCAPTCHA in other places if we want to. That may be out of scope for now, plus I'm not really sure the best approach. Open for suggestions.

modules/login/test/Request_Account_test_plan.md
```diff
@@ -1,13 +1,11 @@
-#Request account feature - Test Plan 
-*Stored under User Accounts module test/ directory*
+# Request Account - Test Plan 
  
 ### Request Account form: 
 1. Verify that form returns error message for each field left blank
-2. Verify that form returns appropriate error message for each field with invalid value entered (e.g.invalid email)
+2. Verify that form returns appropriate error message for each field with invalid value entered (e.g. invalid email)
 3. Verify that verification code is enforced (if and only if re-captcha has been activated by the project)
-4. Verify that new verification code loads on every refresh of the page
```
johnsaigle, line 8: This refers to a very very old custom captcha that we deleted because it was insecure. https://github.com/aces/Loris/pull/3645

maltheism:
Hey @johnsaigle, I'll be happy to test this PR for you and I see it's set as work in progress. Let me know when I should manually test it.

johnsaigle:
You can manually test it now. Thanks!

It's in draft mode only because I have a couple of hacks in the code that could be done better. 

It's totally functional.

modules/login/php/requestaccount.class.inc
maltheism: @johnsaigle maybe `$resp` returns null if error? I'm wondering what happens when an error is thrown.

modules/login/php/requestaccount.class.inc
```diff
@@ -79,10 +83,36 @@ class RequestAccount extends \NDB_Form
...
         $errors = array();
+        // Verify reCAPTCHA
+        $err = array();
+        $reCAPTCHAPrivate = \NDB_Factory::singleton()
+            ->config()
+            ->getSetting('reCAPTCHAPrivate') ?? '';
+
+        // Only validate reCAPTCHA if a private key exists.
+        if (!empty($reCAPTCHAPrivate)) {
+            $success = false;
```
johnsaigle, line 100:
```suggestion
```

modules/login/php/requestaccount.class.inc
```diff
@@ -79,10 +83,36 @@ class RequestAccount extends \NDB_Form
...
+            $resp    = (new \ReCaptcha\ReCaptcha($reCAPTCHAPrivate))
+                ->verify(
+                    $values['g-recaptcha-response'] ?? '',
+                    $_SERVER['REMOTE_ADDR']
+                );
+            $success = $resp->isSuccess();
```
johnsaigle, line 106:
```suggestion
```

modules/login/php/requestaccount.class.inc
```diff
@@ -79,10 +83,36 @@ class RequestAccount extends \NDB_Form
...
+            // FIXME This leverages the error handling for the "site" text field.
+            // It should be done using its own error block.
+            if (!$success) {
```
johnsaigle, line 109:
```suggestion
            if (!$resp->isSuccess()) {
```

modules/login/templates/form_requestaccount.tpl
maltheism: I'm guessing a new react component will be made. 🙂

maltheism:
@johnsaigle probably good to update somewhere that we're using v3 of reCAPTCHA.

maltheism:
@johnsaigle frontend is giving me this output in console:
`Refused to frame 'https://www.google.com/' because it violates the following Content Security Policy directive: "default-src 'self'". Note that 'frame-src' was not explicitly set, so 'default-src' is used as a fallback.`
The reCAPTCHA window is greyed out because of it. I think the nonce has to be included in the content-security-policy header of the NDB_Client.class.inc? I'm unsure though.

johnsaigle:
@maltheism 

> Update the config settings in LORIS as described in the new wiki document


You need to change the CSP. I described how in the wiki document in this PR.

Also this is using v2, not v3. I couldn't get v3 to work. This is mentioned in that document as well.

maltheism:
@johnsaigle I'm not seeing the wiki readme. I was using a v3 key so I'll need to fix that as well.

modules/login/php/requestaccount.class.inc
johnsaigle: Oh sorry I didn't add the new file to Git. 😅

docs/wiki/99 - Developers/reCAPTCHA.md
```diff
@@ -0,0 +1,15 @@
...
+LORIS can be configured to require [Google reCAPTCHA](https://www.google.com/recaptcha/intro/v3.html) in order to limit spam. These following instructions have been tested for reCAPTCHA v2.
+
+After following Google's instructions for obtaining the your public and private keys, you can
+use them by updating your configuration settings in the `Config` table in the database:
+
+* `reCAPTCHAPrivate` and `reCAPTCHAPublic`. These should be set to the
+corresponding values from your Google reCAPTCHA admin account.
+This can also be configured via the Configuration module by modifying the values found in the "API Keys" heading.
+
+* `CSPAdditionalHeaders` should include "frame-src www.google.com;" in order
```
johnsaigle, line 12: @maltheism These are the instructions you need. Sorry for the confusion

maltheism: @johnsaigle maybe not for this PR but I'm a little confused why the way of going about this is by editing the database to include `frame-src www.google.com;`.  I almost think there should just be a check in the code to see if reCAPTCHAPRIVATE has been added to the config.xml or admin configuration and if so add the `frame-src www.google.com;` to the content security policy. I would think that would be best for having it be hands free for users.

johnsaigle: Yeah that would probably be a good enhancement.

maltheism:
Passed manual testing for me.

modules/login/php/requestaccount.class.inc
```diff
@@ -32,7 +32,13 @@ class RequestAccount extends \NDB_Form
      */
     function setup()
     {
-        $this->tpl_data['page_title'] = 'Request Account';
+        $this->tpl_data['page_title']  = 'Request Account';
+        $this->tpl_data['captcha_key'] = \NDB_Factory::singleton()
+            ->config()
+            ->getSetting('reCAPTCHAPublic');
+        $this->tpl_data['nonce']       = base64_encode(
+            openssl_random_pseudo_bytes(16)
```
driusan, line 40: I don't see why this is using an openssl extention instead of just using `random_bytes`

modules/login/php/requestaccount.class.inc
```diff
@@ -79,10 +85,33 @@ class RequestAccount extends \NDB_Form
...
+        if (!empty($reCAPTCHAPrivate)) {
+            $resp = (new \ReCaptcha\ReCaptcha($reCAPTCHAPrivate))
+                ->verify(
+                    $values['g-recaptcha-response'] ?? '',
+                    $_SERVER['REMOTE_ADDR']
+                );
+            // FIXME This leverages the error handling for the "site" text field.
+            // It should be done using its own error block.
+            if (!$resp->isSuccess()) {
+                print_r($resp->getErrorCodes());
```
driusan, line 109: This looks like a leftover debug statement

modules/login/templates/form_requestaccount.tpl
```diff
@@ -61,15 +61,19 @@
       <div class="form-group">
           {$form.radiologist.html}
       </div>
+      <div class="form-group">
+        <!-- Google reCaptcha. Don't display if a private key has not bee configured -->
```
driusan, line 65: This is an HTML comment, not a smarty comment (ie. it'll get sent to the client)

mkdocs.yml
```diff
@@ -32,4 +32,5 @@ nav:
         - 'SQL Dictionary': 'wiki/99 - Developers/SQL Dictionary.md'
         - 'Automated Testing': 'wiki/99 - Developers/Automated Testing.md'
         - 'Style Guide (for help text)': 'HelpStyleGuide.md'
+        - 'Enabling reCAPTCHA': 'wiki/99 - Developers/reCAPTCHA.md'
```
driusan, line 35: Why is this in the developers section? Shouldn't it be in the installation and configuration section? It's more for prod than devs..

modules/login/templates/form_requestaccount.tpl
```diff
@@ -61,15 +61,19 @@
       <div class="form-group">
           {$form.radiologist.html}
       </div>
+      <div class="form-group">
+        {* Google reCaptcha. Don't display if a private key has not bee configured *}
```
driusan, line 65:
```suggestion
        {* Google reCaptcha. Don't display if a private key has not been configured *}
```

mkdocs.yml
johnsaigle: Yeah good point

PR #6347, opened 2020-04-09
[LorisForm, NDB_Page] label vs option vs attribute function misuses

lingma:
## Brief summary of changes

This is the new version of the PR #5947 pushing to the version 23.0, which itself is the recreation of the PR #5884 & #5886, at the same time this constitutes a rebase, which is necessary. This PR contains some `var` to `public` variable declaration modifications at the beginnings of the two classes.

**label vs option vs attribute function misuses (PR #5884 part)**

Initially the third parameters of the form elements were set as the fourth parameters somehow in the program for `static` text Form type, this is clearly a bug. I found later that quite often the term `options` are mixed with the term `attributes`, or even `options` are passed to `label` which is what I stated early.

This will create confusion for programmer. Unfortunately since the same problem existed in the previous versions, migrations might become difficult once the patch has been applied.

I noticed that the  [wiki](https://github.com/aces/Loris/wiki/Instrument-Groups) manual suggests to use the format:
$group[] =& $this->form->createElement("static", null, null, "<label>");

The new format:

Some old instruments will stop displaying correctly once being migrated, the verification required besides some function signatures changes, including setup, score which is beyond this PR scope.

The instrument migration fix could be easy with a full text replacement, with verification. Some project has module overrides, everything could be easy in this case.

This PR has also a little doc modification and the removal of the variable `$closer`, the latter is not too much related with the main task.

**undefined name warning for label element (PR #5886 part)**

A instrument with
$this->addLabel('<label>') will always report PHP warning "Undefined index: name in LorisForm.class.inc on line 1242"

This happens because Loris form assumes that a name is mandatory, even for a label. I don't see the necessity here.

For the undefined name warning for label element issue:

1.  Create an instrument, or a unit test case, put a label $this->addLabel('<label>') anyway in an instrument, check whether the PHP warning is in the log.

### DIFF
--- a/php/libraries/LorisForm.class.inc
+++ b/php/libraries/LorisForm.class.inc
@@ -31,16 +31,16 @@
 class LorisForm
 {
 
-    var $form          = array();
-    var $defaultValues = array();
-    var $formRules     = array();
-    var $errors        = array();
-    var $filters       = array();
-    var $frozen        = false;
-    var $enctype       = '';
-    var $groupMapping  = array();
-    var $action        = '';
-    var $allowedRules  = array(
+    public $form          = array();
+    public $defaultValues = array();
+    public $formRules     = array();
+    public $errors        = array();
+    public $filters       = array();
+    public $frozen        = false;
+    public $enctype       = '';
+    public $groupMapping  = array();
+    public $action        = '';
+    public $allowedRules  = array(
         'compare',
         'requiredIf',
         'required',
@@ -49,6 +49,7 @@ class LorisForm
         'email',
         'maxlength',
     );
+
     public $validOperators;
     public $encrypt;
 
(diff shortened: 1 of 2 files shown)


lingma:
This PR contains a little more modifications and takes long time to rebase or test each time, please review.

driusan:
@lingma I don't understand the reason for this going to the 23 release. This seems to stylisticly be changing `var` to `public` and some parameters name, but I don't see anything else the PR is doing?

lingma:
This is not only the stylistic change, if you read the description then you can understand that this PR fixes some bug, at the same time, I did some name changes.

lingma:
Originally this PR tried to fix the problem of the "options" wrongly sent to the "label" and the "name" not found error.

lingma:
no problem, I will come back soon. This one contains many items for sure

Issue #6312, opened 2020-04-06
[Media Edit] The Language field is not been retrieved at editing time


racostas:
**Describe the bug**
The Language field applied when creating a new media file is not retrieved when trying to edit the same file.   

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Crate a new media and set a language in the language field. 
2. Save the media.
3. Edit the same media
4. The language field will appear as blank (not remembering it previous stage)

**What did you expect to happen?**
The field should show the same language applied until edited.

PR #6349, opened 2020-04-09
[media] use languageLabel derived from language_id to display

PapillonMcGill:
## Brief summary of changes

- go to media module, check if language of document is properly displayed in table
- edit a document with a language, current language should be displayed in the edit window.
- modify the language and check that change are properly reflected in table

https://github.com/aces/Loris/issues/6312

### DIFF
--- a/modules/media/jsx/editForm.js
+++ b/modules/media/jsx/editForm.js
@@ -40,6 +40,7 @@ class MediaEditForm extends Component {
           dateTaken: data.mediaData.dateTaken,
           comments: data.mediaData.comments,
           hideFile: data.mediaData.hideFile,
+          language: data.mediaData.language,
         };
 
         self.setState({
@@ -136,6 +137,7 @@ class MediaEditForm extends Component {
           />
           <SelectElement
             name='language'
+            id='language_id'
             label='Language'
             options={this.state.Data.language}
             onUserInput={this.setFormData}


modules/media/ajax/FileUpload.php
```diff
@@ -300,7 +300,7 @@ function getUploadFields()
             "comments, " .
             "file_name as fileName, " .
             "hide_file as hideFile, " .
-            "language_id as language," .
+            "language_id as language_id," .
```
ridz1208, line 303: you can remove the as since `X as X` is the same as just `X`

modules/media/jsx/editForm.js
```diff
@@ -40,6 +40,7 @@ class MediaEditForm extends Component {
           dateTaken: data.mediaData.dateTaken,
           comments: data.mediaData.comments,
           hideFile: data.mediaData.hideFile,
+          language_id: data.mediaData.language_id,
```
ridz1208, line 43: After reviewing all the changes, it seems like the only change needed here is simple adding this line. and if you undo all the renaming of variables, then this line becomes

```suggestion
          language: data.mediaData.language,
```

I just tested it out and it works. Is there any justification for all the other changes ?

PapillonMcGill: Yes, there was 2 defects in the module, all related to language. The fact that the label is in a separate table witrh an FK in the media table. All this was missed during reactification.

ridz1208: what's the other defect ? from what I could tell the only issue was that when loading the edit form, the language was not displayed.

PapillonMcGill: `language` is for the list of possible language and `language_id` the one that the documents is tag with.

ridz1208: I'm not sure I understand. After removing all changes and just changing this one line everything is functional for me.

I'll leave the decision to other reviewers.

modules/media/php/mediafileprovisioner.class.inc
```diff
@@ -38,22 +38,23 @@ class MediaFileProvisioner extends \LORIS\Data\Provisioners\DBRowProvisioner
...
+                m.uploaded_by as uploadedBy,
+                m.date_taken as dateTaken,
+                substring(m.comments, 1, 50) as comments,
+                m.last_modified as lastModified,
+                m.file_type as fileType,
+                s.CandID as candidateId,
+                s.ID as sessionId,
+                m.hide_file as hideFile,
+                m.id,
+                m.language_id as language
```
ridz1208, line 57: language label is already being queried on line 43.

why is this queried here ? and why use `language` when everywhere else changed to either `language_id` or `languageLabel` ?

PapillonMcGill: That line might be extra. I had trouble finding the exact place to correct for the edit window.

modules/media/jsx/editForm.js
PapillonMcGill: have you tried to edit an document meta-data without the other change? 2nd and 3rd bullet of how to test in the description.

ridz1208: yes, give it a try.

modules/media/php/mediafileprovisioner.class.inc
```diff
@@ -38,22 +38,22 @@ class MediaFileProvisioner extends \LORIS\Data\Provisioners\DBRowProvisioner
     {
         parent::__construct(
             "SELECT m.file_name as fileName,
-                    c.PSCID as pscid,
```
driusan, line 41: This shouldn't be making whitespace changes

PapillonMcGill: I know but ended up with missalign lines when I added a line here (that was later removed). I had to replace lots of tab with space.

modules/media/jsx/editForm.js
ridz1208: @PapillonMcGill have you had a chance to give it a try with only this line ?

modules/media/jsx/editForm.js
```diff
@@ -136,6 +137,7 @@ class MediaEditForm extends Component {
           />
           <SelectElement
             name='language'
+            id='language_id'
```
PapillonMcGill, line 140: @ridz1208 This line is required to have current value shown in the edit windows

modules/media/jsx/mediaIndex.js
```diff
@@ -159,7 +159,7 @@ class MediaIndex extends Component {
         options: options.visits,
       }},
       {label: 'Language', show: true, filter: {
-        name: 'language',
+        name: 'languageLabel',
```
PapillonMcGill, line 162: @ridz1208 removing this line, the current value is not shown in the edit form.

modules/media/jsx/editForm.js
PapillonMcGill: @ridz1208 You're right, minimal change required. Not sure what was the cause but with other change, the modal windows closed automatically when I click on the OK button after update during initial test. While retrying, the modal windows always stayed after clicking on the Ok button after update.

driusan:
@racostas can you confirm that this fixes the issue you reported?

racostas:
> @racostas can you confirm that this fixes the issue you reported?

It seems to be working fine. Tested locally.

The following is very minor:
Found that once the file is loaded (the user been in the tab `Upload`), if the user goes to the tab `Browse` the new entry doesn't appear until manually refreshed the page (F5). The concern is when not knowledgeable users are doing the data entry.

modules/media/jsx/editForm.js
ridz1208, line 140: @PapillonMcGill @driusan this still seems to be an unnecessary change (removing it does not impact functionality) but since the PR fixes the issue I won't block it over this

Issue #6245, opened 2020-04-02
[dashboard] Pending approval link error page 


ridz1208:
**Describe the bug**
When some user accounts are still pending approval, an entry in the `My Tasks` panel on the dashboard appears "X Pending account approval"

clicking on that link leads to a loris error page

<img width="1680" alt="Screen Shot 2020-04-02 at 5 36 29 PM" src="https://user-images.githubusercontent.com/15268287/78302343-95b3d400-7508-11ea-8fc8-be3e0df4cbb6.png">
<img width="516" alt="Screen Shot 2020-04-02 at 5 36 39 PM" src="https://user-images.githubusercontent.com/15268287/78302345-964c6a80-7508-11ea-9635-a01d086af307.png">

ridz1208:
Most probably related to #6233

johnsaigle:
@jesscall @ridz1208 Can this be closed?

Issue #6233, opened 2020-04-02
[Login] request account does not request an account


ridz1208:
**Describe the bug**
When requesting a new account, the account does not show in the list of the admin user accounts. 

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to login -> request account
2. Fill all the fields and submit
3. Login as admin
4. The newly requested account does not appear in the list of user-accounts

I have a suspicion that it might be related to the project permission. I.e. the account is there but not visible.

ridz1208:
If the suspected issue is correct. 

- Admin account should be able to view all users regardless of permissions
- request account should get a mandatory project field

johnsaigle:
Request account displays a Success message even when account creation fails. It's possible that there was another processing error that you can view in the error log.

johnsaigle:
I got this error when trying to load user_accounts after using request_account:

`PHP Fatal error:  Uncaught TypeError: explode() expects parameter 2 to be string, null given in /var/www/loris/modules/user_accounts/php/useraccountrowprovisioner.class.inc:48`

johnsaigle:
@ridz1208 is this resolved?

PR #6352, opened 2020-04-09
[login][request_account] Adding project field to request account

jesscall:
## Brief summary of changes
This PR adds `project` field to the request account form.

- [ ] Have you updated related documentation? **NOT NEEDED**.

1. Follow the test plan for request account and assure that the bugs reported during testing round have been amended.

#6233 
#6245

### DIFF
--- a/modules/login/php/requestaccount.class.inc
+++ b/modules/login/php/requestaccount.class.inc
@@ -173,21 +173,29 @@ class RequestAccount extends \NDB_Form
             // Get the ID of the new user and insert into user_psc_rel
             $user_id = $DB->getLastInsertId();
 
-            $DB->insert(
-                'user_psc_rel',
-                array(
-                    'UserID'   => $user_id,
-                    'CenterID' => $site,
-                )
-            );
+            // verify site exists
+            $siteList = \Utility::getSiteList();
+            if (!is_null($site) && array_key_exists($site, $siteList)) {
+                $DB->insert(
+                    'user_psc_rel',
+                    array(
+                        'UserID'   => $user_id,
+                        'CenterID' => $site,
+                    )
+                );
+            }
 
-            $DB->insert(
-                'user_project_rel',
-                array(
-                    'UserID'    => $user_id,
-                    'ProjectID' => $site,
-                )
-            );
+            // verify project exists
+            $projectList = \Utility::getProjectList();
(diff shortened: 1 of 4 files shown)


modules/login/php/requestaccount.class.inc
```diff
@@ -171,6 +181,14 @@ class RequestAccount extends \NDB_Form
                 )
             );
 
+            $DB->insert(
+                'user_project_rel',
+                array(
+                    'UserID'    => $user_id,
+                    'ProjectID' => $site,
```
ridz1208, line 188:
```suggestion
                    'ProjectID' => $project,
```

ridz1208:
@maltheism please test again... but better

maltheism:
@ridz1208 🥺 ..but the manual test resolved all the issues. 😭 I failed as a tester.

maltheism:
@ridz1208 I just manually tested this PR and the issues the PR resolves are solved. I'm still confused why the issues were resolved with the previous code mistake.

ridz1208:
@maltheism it was still putting a value in the table but instead of putting the project ID value
it was putting the site ID in the project column lol

modules/login/php/requestaccount.class.inc
```diff
@@ -171,6 +181,14 @@ class RequestAccount extends \NDB_Form
                 )
             );
 
+            $DB->insert(
+                'user_project_rel',
+                array(
+                    'UserID'    => $user_id,
+                    'ProjectID' => $project,
```
driusan, line 188: `$project` came from the value submitted by the user. It needs to be validated by the backend before being inserted and shouldn't be blindly trusted.

ridz1208: then same should be done with site

driusan: True, might as well do that at the same time.

jesscall: @driusan what kind of validation would you like to see here? The values for site and project are coming from drop-downs populated with values from the DB.

ridz1208: they are then passed to the insert query as parameters and inserted into a table with foreign keys...

driusan: 1. The value submitted is an integer.
2. The value integer submitted exists as a key in the appropriate (project/psc) tables.

Issue #6350, opened 2020-04-09
[DataFramework Filters] Filtering for userSiteMatch is bypassing 'across all sites' types of permissions


jesscall:
**Describe the bug**
Noticed this bug in the user_accounts module. Only rows of users that have a matching site are displayed despite having the `Across all sites create and edit users` permission. 
**[NOTE]** make sure you **do not** have `Across all sites access candidate profiles` permission

**To Reproduce**
Steps to reproduce the behaviour (attach screenshots if applicable):
1. Create a new user that is affiliated to a site that you **do not** have access to.
2. Go to User Accounts page and try to find that user in the list. You should be unable to view the user even when you have the permission `Across all sites create and edit users`.

**What did you expect to happen?**
Should be able to view all users regardless of their site affiliation when granted the `Across all sites create and edit users` permission.

PR #6354, opened 2020-04-09
[DataFramework Filters] Check Site Permissions Filter

jesscall:
## Brief summary of changes
This PR adds a new abstract function to the DataFrameworkMenu: `useSitePermissionsFilter`. This function is meant to signal to the base class that the page's provisioner can support the `HasAnyPermissionOrUserSiteMatch` filter. 

This is needed because currently there is no way to signal to the provisioner that the user has an "across all sites" type of permission for the data being retrieved, and should therefore not filter based on site match.

1. Go through the affected modules and make sure the data tables are filtered properly according to their related site access permissions. 
- [ ] candidate_list:  `access_all_profiles`
- [ ] dicom_archive: `dicom_archive_view_allsites`
- [ ] eeg_browser: `electrophysiology_browser_view_allsites`
- [ ] media: `access_all_profiles`
- [ ] user_accounts: `user_account_multisite`

Should remain unaffected, but should check functionality:
- [ ] datadict
- [ ] imaging_broswer
- [ ] module_manager
- [ ] survery_accounts

#6350

### DIFF
--- a/modules/survey_accounts/php/survey_accounts.class.inc
+++ b/modules/survey_accounts/php/survey_accounts.class.inc
@@ -47,6 +47,17 @@ class Survey_Accounts extends \DataFrameworkMenu
     }
 
     /**
+     * Tells the base class that this page's provisioner can support the
+     * HasAnyPermissionOrUserSiteMatch filter.
+     *
+     * @return ?array of site permissions or null
+     */
+    public function useSitePermissionsFilter() : ?array
+    {
+        return null;
+    }
+
+    /**
      * {@inheritDoc}
      *
      * @return bool
--- a/modules/survey_accounts/php/survey_accounts.class.inc
+++ b/modules/survey_accounts/php/survey_accounts.class.inc
@@ -52,7 +52,7 @@ class Survey_Accounts extends \DataFrameworkMenu
      *
      * @return ?array of site permissions or null
      */
-    public function useSitePermissionsFilter() : ?array
+    public function allSitePermissionNames() : ?array
     {
         return null;
     }
--- a/php/libraries/DataFrameworkMenu.class.inc
+++ b/php/libraries/DataFrameworkMenu.class.inc
@@ -67,6 +67,15 @@ abstract class DataFrameworkMenu extends NDB_Menu_Filter
     abstract public function useSiteFilter() : bool;
 
     /**
+     * Determines whether the DataProvisioner should filter out rows based on user's
+     * site viewing permissions.
+     *
+     * @return ?array of permissions to check or null
+     */
+    abstract public function useSitePermissionsFilter() : ?array;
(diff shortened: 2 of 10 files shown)


johnsaigle:
Travis is failing with this error:

/home/travis/build/aces/Loris/php/libraries/DataFrameworkMenu.class.inc:113	**Avoid unused local variables such as '$user'.**

php/libraries/DataFrameworkMenu.class.inc
```diff
@@ -101,19 +110,23 @@ abstract class DataFrameworkMenu extends NDB_Menu_Filter
...
 
-        $user = \NDB_Factory::singleton()->user();
-
         // UserSiteMatch throws an exception if used on a model without a getCenterID
         // or getCenterIDs function, so we only apply it on sub-classes that
         // specifically say they support it.
-        if ($this->useSiteFilter()
-            && $user->hasPermission("access_all_profiles") == false
-        ) {
+        if ($this->useSiteFilter()) {
```
driusan, line 116: I don't see what the difference between `useSiteFilter` and `useSitePermissionsFilter` is supposed to be. Why doesn't this just use something like a `allSitePermissionNames` function and update the if statement to check that, always using the `HasAnyPermissionOrUserSiteMatch` filter?

johnsaigle:
I think this maybe needs a rebase? Phan is having issues with the WebDriver library which was updated elsewhere.

modules/survey_accounts/php/survey_accounts.class.inc
```diff
@@ -46,6 +46,17 @@ class Survey_Accounts extends \DataFrameworkMenu
         return true;
     }
 
+    /**
+     * Tells the base class that this page's provisioner can support the
+     * HasAnyPermissionOrUserSiteMatch filter.
+     *
+     * @return ?array of site permissions or null
+     */
+    public function allSitePermissionNames() : ?array
```
johnsaigle, line 55: I think this change is causing Travis to fail.

php/libraries/DataFrameworkMenu.class.inc
```diff
@@ -101,16 +110,14 @@ abstract class DataFrameworkMenu extends NDB_Menu_Filter
...
 
-        $user = \NDB_Factory::singleton()->user();
-
         // UserSiteMatch throws an exception if used on a model without a getCenterID
         // or getCenterIDs function, so we only apply it on sub-classes that
         // specifically say they support it.
-        if ($this->useSiteFilter()
-            && $user->hasPermission("access_all_profiles") == false
-        ) {
+        if ($this->useSiteFilter() || $this->allSitePermissionNames) {
```
driusan, line 116: I think this is causing travis to fail.

PR #6357, opened 2020-04-09
[smarty template] add possible go back one page function

### LINKED ISSUE #6266: [instrument list] imaging browser link showing without permission
**Describe the bug**
A clear and concise description of what the bug is

**To Reproduce**
1. Go to 'instrument_list' of a candidate
2. Click on 'View Imaging data'
3. The system says "no permission" and a return button to the home page.

If I don't have permission, is it a good practice to not show the link? This is related with the issue #6219, the link is everywhere on the instrument lists, but some has no permission, some will report system crash error if there is no certain required instruments.

The return button should return to the last visiting point, not returning to the home page, the latter is the last resort when you don't know where to return

- LORIS Version: 23.0

image

image

lingma:
## Brief summary of changes
Add possibility to go back one page because if the permission control kicks in, we have to return to the home page if we don't know click on the browser back button

The #6266 is hard to do optimistically because imaging browser and instrument list are presented as two independent modules. The current solution could solve the similar problem in any module.

1. Click on any link that you don't have the permission, like a profile imaging data link, then the system will ask you to go back to the home page, you will have to start over again to navigate through the system.

Partially resolves #6266

### DIFF
--- a/smarty/templates/403.tpl
+++ b/smarty/templates/403.tpl
@@ -1,4 +1,4 @@
 <div class="container">
     <h3>{$message}</h3>
-    <div><a href="{$baseurl}">Go to main page</a></div>
+    <div><a href="{$baseurl}">Go to main page</a> | <a onclick="window.history.back();" href="#">Go back one page</a></div>
 </div>


maltheism:
Interesting @lingma. I haven't seen this javascript function used before. My only issue with it would be an edge case of somehow the user being redirected away from Loris. I don't know if that would happen. @driusan thoughts?

lingma:
> Interesting @lingma. 

You are right, this fix is not good enough

lingma:
To solve your concern, we will need to add a return page to every call so that we know where we were coming from, which might have a lot impact.

maltheism:
@lingma you could also use the `document.referrer` to see if it matches the domain of $baseurl. I'm not sure if @driusan will like the idea though.

lingma:
maybe we can assume that all the users should have the permission for the imaging browser. In this case, this PR will become useless.

lingma:
Even though this PR seems useless, I put the base url link as the first one to reduce the confusion to the users.

Issue #6259, opened 2020-04-05
[BVL Feedback] No BVL Feedback link at the instrument level


SantiagoTG:
There is no way to access the Behavioral Feedback when accessing an instrument. It is only available at the Candidate and Visit level. 

![Screen Shot 2020-04-04 at 8 07 23 PM](https://user-images.githubusercontent.com/54332302/78463871-2678e400-76b0-11ea-9bc5-80f460434d24.png)

johnsaigle:
I think this is resolved by the linked PR

SantiagoTG:
Reopening as issue is still present. BHV feedback link still missing at the instrument level.

driusan:
@SantiagoTG how did you test?

SantiagoTG:
Hi Dave, I tested the same way I did initially, by entering https://test-loris-dev.loris.ca/ and checking the BVL feedback links for a given candidate.

johnsaigle:
Is the Test VM up to date?

lingma:
I just checked, the new 23.0 should be OK. But not on the test VM

PR #6358, opened 2020-04-09
[BVL Feedback] Restore BVL Feedback link at the instrument level

zaliqarosli:
## Brief summary of changes

This PR passes the actual instrument/test name i.e. mri_parameter_form instead of the string "instruments" ($page->name is "instruments because it's the name of the module) to `UserPageDecorationMiddleware` so that `\NDB_BVL_Feedback::bvlFeedbackPossible()` can be called with the correct testname.

1. Go to any instrument page on 23.0-release branch. Notice that the BVL feedback icon does not appear on the navigation bar.
2. On this branch, notice that the icon is restored.

* Resolves https://github.com/aces/Loris/issues/6259

### DIFF
--- a/htdocs/css/simple-sidebar.css
+++ b/htdocs/css/simple-sidebar.css
@@ -119,7 +119,7 @@
 
     #sidebar-wrapper {
         left: 0;
-        color: white;
+        color: #064785;
     }
 
     .wrapper.active {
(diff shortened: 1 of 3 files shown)


zaliqarosli:
@driusan i narrowed the problem down to "instruments" string getting passed to `UserPageDecorationMiddleware` instead of the actual instrument testName which is what is needed for the BVL feedback panel/icon to show up. I'm not sure how far back `$page->name` needs to be changed to `$page->testName` and what else this change affects. At what point does the module name need to be passed, and the testname needs to be passed. I opted to change it at the deepest level to only get the bvl feedback to show up on instrument pages.

Issue #6285, opened 2020-04-06
[imaging_uploader] Uploading phantom scan unsuccessful


zaliqarosli:
**Describe the bug**
On trying to upload a phantom scan (no requirements for file name), the progress bar goes to 100%, but nothing happens (success swal doesn't pop up, page not redirected). In console log, we get a 500 error and a JS uncaught type error:

<img width="1190" alt="Screenshot 2020-04-06 13 32 52" src="https://user-images.githubusercontent.com/17415878/78588450-a6b45c00-780c-11ea-8c88-2b0acbc3aeb5.png">

Additionally, this issue happens: https://github.com/aces/Loris/issues/6288

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to 'Imaging Uploader' module
2. Click on 'Upload' tab
3. Fill in the form: 'Phantom Scans': Yes
4. Upload a file into 'File to Upload' field, with any name but with correct file types .tgz or tar.gz or .zip
5. Click 'Submit' 
6. See console log errors as in the screenshot above. No swal pops up where it should.

**What did you expect to happen?**
On clicking 'Submit', the upload should be successful, or if not, a swal with the corresponding error should pop up to notify the user.

ridz1208:
You can contact @cmadjar

PR #6361, opened 2020-04-10
[imaging_uploader] can't upload file with Phantom set 'yes'

kongtiaowang:
## Brief summary of changes
session ID must be null or integer, not ''.

Uploade a phantom scan 

https://github.com/aces/Loris/issues/6285

### DIFF
--- a/modules/imaging_uploader/php/imaging_uploader.class.inc
+++ b/modules/imaging_uploader/php/imaging_uploader.class.inc
@@ -541,7 +541,7 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
             'UploadedBy'               => $user_name,
             'UploadDate'               => $date,
             'UploadLocation'           => $this->uploaded_file_path,
-            'SessionID'                => $sessionid ?? '',
+            'SessionID'                => $sessionid,
             'PatientName'              => $pname,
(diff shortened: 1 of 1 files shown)


modules/imaging_uploader/php/imaging_uploader.class.inc
```diff
@@ -541,7 +541,7 @@ class Imaging_Uploader extends \NDB_Menu_Filter_Form
             'UploadedBy'               => $user_name,
             'UploadDate'               => $date,
             'UploadLocation'           => $this->uploaded_file_path,
-            'SessionID'                => $sessionid ?? '',
+            'SessionID'                => $sessionid ?? null,
```
johnsaigle, line 544: @cmadjar @nicolasbrossard Can you comment on whether this case should occur? Is it OK for this column to be null after uploading or should it always be attached to a session ID?

nicolasbrossard: @johnsaigle You can upload a scan that is not tied to *any* session. When the pipeline processes that scan, it will create a session for you if you have `createVisitLabel` set to `1` in your `prod` file. So yeah, `$sessionid` can be `null`.

PR #6362, opened 2020-04-10
[MRI Violations] Fixed bug in filter of the MRI violations module.

### LINKED ISSUE #6279: [MRI violations] Result set incorrect when user is associated to a single site
**Describe the bug**
I logged in as a user that only has the "Edit MRI protocol" permission, that is associated to one site (Montreal) and that does not have the `access_all_profiles` permission. When I access the MRI violations module, I see MRI violations that are associated to sites other than Montreal.

**To Reproduce**
Access the MRI violations module with the pre-conditions listed above and check that you can see all the MRI violations (you shouldn't).

nicolasbrossard:
## Brief summary of changes

This PR fixes the following bugs in the MRI violations module, both in the `Not Resolved` and `Resolved` tabs:

1. When you select a given site in the drop-down list, the results contain records for which the site entry is empty: these should not be part of the result set.
2. When you access the MRI violations page for the first time or hit the `Clear Filter` button, all the MRI violations are shown even if the user does not have the `access_all_profiles` permission.

1. Check that the filter's sites drop-down list offers choices that allow you to see/hide the violations that are not associated to any site.
2. Make sure that as a user that does not have the `access_all_profiles` permission you can never see the violations that is associated to a site you do not belong to (this happens when you first access the page and when you click on the `Clear Filter` button).

These tests should be performed on both the `Not Resolved` and `Resolved` tabs. 

3. On the dashboard page, check that the count of violated scans for a user that has access to the MRI violated scans module is correct. A test should be done for a user that has the `access_all_profile` permission and for a user that hasn't.

Note that it is assumed that violations not associated to any site can be seen by anybody, as long as that person has access to the MRI violations module.

* Resolves #6279

### DIFF
--- a/SQL/0000-00-01-Permission.sql
+++ b/SQL/0000-00-01-Permission.sql
@@ -109,7 +109,8 @@ INSERT INTO `permissions` VALUES
     (61,'battery_manager_edit','Add, activate, and deactivate entries in Test Battery',2),
     (62,'module_manager_view', 'Module Manager: access the module', 2),
     (63,'module_manager_edit', 'Module Manager: edit installed modules', 2),
-    (64,'candidate_dod_edit', 'Edit dates of death', 2);
+    (64,'candidate_dod_edit', 'Edit dates of death', 2)
+    (65,'violated_scans_view_ownsite','Violated Scans: View Violated Scans from own site','2');
 
 INSERT INTO `user_perm_rel` (userID, permID)
   SELECT u.ID, p.permID
--- a/SQL/0000-00-01-Permission.sql
+++ b/SQL/0000-00-01-Permission.sql
@@ -109,7 +109,7 @@ INSERT INTO `permissions` VALUES
     (61,'battery_manager_edit','Add, activate, and deactivate entries in Test Battery',2),
     (62,'module_manager_view', 'Module Manager: access the module', 2),
     (63,'module_manager_edit', 'Module Manager: edit installed modules', 2),
-    (64,'candidate_dod_edit', 'Edit dates of death', 2)
+    (64,'candidate_dod_edit', 'Edit dates of death', 2),
     (65,'violated_scans_view_ownsite','Violated Scans: View Violated Scans from own site','2');
 
 INSERT INTO `user_perm_rel` (userID, permID)
--- a/modules/mri_violations/php/module.class.inc
+++ b/modules/mri_violations/php/module.class.inc
@@ -28,17 +28,19 @@ class Module extends \Module
      * Gets the base query (FROM and WHERE clause) used to retrieve
      * all the MRI violations for a given user.
      *
-     * @return void
+     * @return string the computed query.
      */
     public static function getMriViolatedScanQuery()
     {
         // Find which sites the user has access to
-        $accessibleSites = \User::singleton()->hasPermission('access_all_profiles')
+        $user            = \User::singleton();
+        $accessibleSites = $user->hasPermission('violated_scans_view_allsites')
             ? \Utility::getSiteList() : \User::singleton()->getStudySites();
 
         // Build the component of the WHERE clause (for the query below)
         // that will be used to restrict the search to the data associated
         // to sites the user has access to
+        $siteClauseElements = [];
         foreach (array_keys($accessibleSites) as $siteID) {
             $siteClauseElements[] = \Database::singleton()->quote($siteID);
         }
@@ -171,7 +173,7 @@ class Module extends \Module
             $user->hasAnyPermission(
                 [
                     'violated_scans_view_allsites',
-                    'violated_scans_edit'
+                    'violated_scans_view_ownsite'
                 ]
             );
     }
@@ -212,23 +214,18 @@ class Module extends \Module
         case "usertasks":
             $factory = \NDB_Factory::singleton();
             $DB      = $factory->database();
-            $baseURL = $factory->settings()->getBaseURL();
+            $link    = $factory->settings()->getBaseURL() . '/' . $this->getName();
+            $number  = (int) $DB->pselectOne(
+                "SELECT count(*) " . $this->getMriViolatedScanQuery(),
+                []
+            );
+            $label   = $number >  1 ? 'Violated scans' : 'Violated scan';
 
             return [
-                new \LORIS\dashboard\TaskQueryWidget(
-                    $user,
-                    "Violated scan",
-                    $DB,
-                    "SELECT
-               count(*)". $this->getMriViolatedScanQuery(),
-                    "violated_scans_view_allsites",
-                    "c.RegistrationCenterID",
-                    $baseURL . "/" . $this->getName(),
-                    ""
-                )
+                new \LORIS\dashboard\TaskWidget($label, $number, $link, '')
             ];
-
         }
+
         return [];
     }
 }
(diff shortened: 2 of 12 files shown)


modules/mri_violations/php/mri_violations.class.inc
```diff
@@ -213,7 +213,24 @@ class Mri_Violations extends \NDB_Menu_Filter_Form
         //      then uses it as is, the path is correct.
         //   3. if it is not in the assembly dir, then it is in the trashbin,
         //      so we fix the path with that in mind.
-        $this->query    = " FROM (
+
+        // Find which sites the user has access to
+        $accessibleSites = \User::singleton()->hasPermission('access_all_profiles')
```
driusan, line 218: This doesn't seem like the right permission.. isn't there an imaging related all profiles permission?

driusan: @cmadjar maybe you know?

cmadjar: There are two permissions for imaging browser: imaging_browser_view_allsites and imaging_browser_phantom_allsites.

Not sure which one is best or if both should be used...

driusan: So what should we do? Just merge it with access_all_profiles?

driusan:
Adding Discussion Required because the code looks good and it's approved but a decision needs to be made on the right permission..

modules/mri_violations/php/mri_violations.class.inc
cmadjar: maybe we could use `dicom_archive_view_allsites`? Just realized that this is the one used by the DICOM archive module. @driusan what do you think? 

Makes more sense than access_all_profiles and less complex than the imaging browser permissions

driusan: I don't have any strong opinion. We could also create a new permission.

cmadjar:
@nicolasbrossard There are conflicts that need to be fixed. Could you take care of that? Thank you

nicolasbrossard:
@driusan We already have permission `violated_scans_view_allsites`, which, as the name suggests, should allow a user to view all violations. Maybe all we need is the new permission `violated_scan_own_site` to limit the violations that can be seen? Agreed?

driusan:
@nicolasbrossard okay

modules/mri_violations/php/module.class.inc
```diff
@@ -24,117 +24,140 @@ namespace LORIS\mri_violations;
...
+    /**
+     * Gets the base query (FROM and WHERE clause) used to retrieve
+     * all the MRI violations for a given user.
+     *
+     * @return void
+     */
+    public static function getMriViolatedScanQuery()
+    {
+        // Find which sites the user has access to
+        $accessibleSites = \User::singleton()->hasPermission('access_all_profiles')
```
cmadjar, line 36: I guess the permission needs to be changed to the mri_violation_all_site permission here. No?

modules/mri_violations/php/mri_violations.class.inc
```diff
@@ -314,19 +362,24 @@ class Mri_Violations extends \NDB_Menu_Filter_Form
         // Get sites
         if ($user->hasPermission('access_all_profiles')) {
```
cmadjar, line 363: same here, need to change the permission to be the violation_view all sites one

modules/mri_violations/php/resolved_violations.class.inc
```diff
@@ -194,6 +194,22 @@ class Resolved_Violations extends \NDB_Menu_Filter_Form
             ]
         );
 
+        // Find which sites the user has access to
+        $accessibleSites = \User::singleton()->hasPermission('access_all_profiles')
```
cmadjar, line 198: same comment about the permission

modules/mri_violations/php/resolved_violations.class.inc
```diff
@@ -378,18 +394,72 @@ class Resolved_Violations extends \NDB_Menu_Filter_Form
         // Get sites
         if ($user->hasPermission('access_all_profiles')) {
```
cmadjar, line 395: same comment about the permission

nicolasbrossard:
@cmadjar After my next commit, there will be two permissions related to the MRI violated scans module: `violated_scans_view_allsites` and `violated_scans_view_ownsite`. You need at least one of them to have access to the MRI violated scan module. With `violated_scans_view_ownsite`, you can only see the violations associated to the sites that you belong to (and those for which the site is unknown) whereas the other permission gives you access to everything. Although permission `violated_scans_edit` is still in the permissions table, it is not currently used: it will be though once we reactivate the edit MRI protocol table feature (I did not want to delete the permission only to add it again in the next LORIS version...). HTH

modules/mri_violations/php/module.class.inc
nicolasbrossard: Correct

modules/mri_violations/php/mri_violations.class.inc
nicolasbrossard: Correct

modules/mri_violations/php/resolved_violations.class.inc
nicolasbrossard: Correct

modules/mri_violations/php/resolved_violations.class.inc
nicolasbrossard: Correct

driusan:
@nicolasbrossard can you update the Permissions section of the README and the test plan too?

nicolasbrossard:
@driusan Will do right now.

SQL/0000-00-01-Permission.sql
```diff
@@ -109,7 +109,8 @@ INSERT INTO `permissions` VALUES
     (61,'battery_manager_edit','Add, activate, and deactivate entries in Test Battery',2),
     (62,'module_manager_view', 'Module Manager: access the module', 2),
     (63,'module_manager_edit', 'Module Manager: edit installed modules', 2),
-    (64,'candidate_dod_edit', 'Edit dates of death', 2);
+    (64,'candidate_dod_edit', 'Edit dates of death', 2)
```
kongtiaowang, line 112: missing a comma here.

nicolasbrossard: @kongtiaowang You are right. Thx!

kongtiaowang:
@nicolasbrossard  Loris/raisinbread/RB_files/RB_permissions.sql

modules/mri_violations/test/TestPlan.md
```diff
@@ -21,6 +21,15 @@
 6. Ensure all filters and sorting work in the `Resolved` tab.
 7. Click on the question mark on the right upper side of the windows and ensure
    that the help content about MRI Violation is showing up and is up-to-date.
+8. Ensure user has access to this page iff he/she has either permission
```
driusan, line 24:
```suggestion
8. Ensure user has access to this page if and only if he/she has either permission
```

People don't know words.

modules/mri_violations/test/TestPlan.md
```diff
@@ -21,6 +21,15 @@
...
+   `mri_violations_view_allsite` or `mri_violations_view_ownsite`.
+9. Check that if the user has permission `mri_violations_view_ownsite`, he/she
```
driusan, line 26:
```suggestion
9. Check that if the user only has the permission `mri_violations_view_ownsite`, he/she
```

They might have both, "all" should have priority.

Issue #6304, opened 2020-04-06
[Document Repository] Download issue


pierre-p-s:
Test plan: 6 

After uploading a file from a user account, I cannot download it from admin but admin can download it without problems. Same issue vice versa.

ridz1208:
>  I cannot download it from admin but admin can download it without problems

Not sure i understand this sentence

pierre-p-s:
> > I cannot download it from admin but admin can download it without problems
> 
> Not sure i understand this sentence

My issue is: if admin uploads a file, then admin can retrieve this file downloading it from Loris. However if another user tries to download it, it gives an error.

kongtiaowang:
This is a major bug in this module. Even admin user can only download his own files.

HenriRabalais:
I'm trying to test but getting this error. Do i have to manually make a director for each user?
```
[Thu Apr 30 14:42:36.747292 2020] [php7:warn] [pid 25091] [client 10.31.11.193:54259] PHP Warning:  mkdir(): No such file or directory in /var/www/loris/modules/document_repository/php/files.class.inc on line 370, referer: http://hrabalais-dev.loris.ca/document_repository/
[...]

PR #6363, opened 2020-04-10
[document_repository] fix download files issue

kongtiaowang:
## Brief summary of changes
Fix the wrong directory name.
Admin user and non-admin user can download file from others.

User A uploads a file, admin and user B can download it.
https://github.com/aces/Loris/issues/6304


modules/document_repository/jsx/docIndex.js
```diff
@@ -126,7 +126,7 @@ class DocIndex extends React.Component {
     let result = <td>{cell}</td>;
     switch (column) {
       case 'File Name':
-        let downloadURL = loris.BaseURL + '/document_repository/Files/' + encodeURIComponent(row['File Name']);
+        let downloadURL = loris.BaseURL + '/document_repository/Files/'+ row['Uploaded By'] +'/' + encodeURIComponent(row['File Name']);
```
driusan, line 129: I don't understand this.. why is uploaded by part of the URL?

kongtiaowang: I want to pass a username as a parameter by the front-end clicking.
If I get the username , I could locate the file which upload by users.

johnsaigle: I think a better way to do this would be to send the ID from the front-end and use it to query the database for the `Data_dir` value which contains the file location.

kongtiaowang: api /Files/ID has been used to get all the upload info of this file.

kongtiaowang: api /Files/filename will download the file. I will remove the Uploaded By from url.

modules/document_repository/php/files.class.inc
```diff
@@ -381,7 +387,7 @@ class Files extends \NDB_Page
                         'version'       => $version,
                         'File_name'     => $fileName,
                         'File_size'     => $fileSize,
-                        'Data_dir'      => $fileName,
+                        'Data_dir'      => $uploadPath,
```
johnsaigle, line 390: I'm not sure that this will work with existing files in the document repository. You'll have entries like

`image.jpg` 
and 
`/data/loris/data/document_repository/admin/image.jpg`

in the same column

kongtiaowang: fixed

modules/document_repository/php/files.class.inc
```diff
@@ -381,7 +388,7 @@ class Files extends \NDB_Page
...
```
ridz1208, line 391: @driusan @kongtiaowang we might have a serious problem on our hands here:

- Before LORIS 21.0 we were saving here the [FILEBASE](https://github.com/aces/Loris/blob/20.0-release/modules/document_repository/ajax/documentEditUpload.php#L93)  which looks like `$username/$filename`
- 21.0 and 22.0 we have been saving the FILENAME (as you can see in the before state of this PR) which is only `$filename`
- In this PR we are changing it to UPLOADPATH which is the absolute path !! `$configpath/*username/$filename`

All of these changes are done without taking into consideration what was already previously in the database which means all documents uploaded in previous versions would not be downloadable...

modules/document_repository/php/files.class.inc
```diff
@@ -108,10 +108,17 @@ class Files extends \NDB_Page
...
                 );
         case "GET":
-            $name   = \User::singleton()->getUsername();
             $record = urldecode(basename($request->getUri()->getPath()));
+
             if (!is_numeric($record)) {
-                $file = $path. "$name/$record";
+                $factory = \NDB_Factory::singleton();
+                $DB      = $factory->database();
+                $upload  = $DB->pselectOne(
```
ridz1208, line 116:
```suggestion
                $uploadedBy  = $DB->pselectOne(
```

modules/document_repository/php/files.class.inc
```diff
@@ -108,10 +108,17 @@ class Files extends \NDB_Page
...
+                    "SELECT uploaded_by FROM document_repository
+                    WHERE File_Name=:name",
+                    array('name' => $record)
+                );
+                $file    = $path. "$upload/$record";
```
ridz1208, line 121:
```suggestion
                $file    = $path. "$uploadedBy/$record";
```

modules/document_repository/php/files.class.inc
johnsaigle: @ridz1208 It doesn't really matter what's stored. In theory `Data_dir` can be deleted and we can construct the path using `$configpath/$username/$fileName`.

`Data_dir` has always seemed like a poor design choice to me and we should probably remove it.

ridz1208: if thats the case it should just be removed then. where is it used if it's not in rebuilding the path to the file ?

johnsaigle: I don't think it is used. The doc repo code has been rewritten at least twice since i've worked here and I think it needs a lot of cleanup in general. 

I agree it should be removed. I'll make a cleanup ticket.

ridz1208:
```suggestion
                        'Data_dir'      => $fileName,
```

Issue #6360, opened 2020-04-09
[module manager] Features doesn't meet the name


lingma:
**Describe the bug**
The Module manager doesn't do what a module manager should do, it is actually a menu manager as I can conclude.

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Disable any modules, the module will disappear from the menu,
2. Access that module with the original link, all works fine just as before

**What did you expect to happen?**
The function observed is not module manager, but menu manager. The functions should be enforced.

**Server Environment (if known):**
- LORIS Version: 23.0

lingma:
We can see what to do depending on the feedback on the PR #6365 or other propositions.

johnsaigle:
@driusan Can you comment on this since you created the module?

lingma:
The module manager itself looks OK for the general users as long as there is no link to other modules since the module does disappear from the menu, currently there is no way to stop those modules to be available even if the module is disabled. PR #6365 method could be a way but this might slow down the page rendering.

driusan:
This ticket doesn't make any sense. The module manager manages lists modules that are installed and let's you set them to active or inactive. It doesn't do any managing of menus at all and the only connection to menus is that LORIS doesn't display inactive modules in the menu.

johnsaigle:
That is my understanding also but I didn't have time to look at the code or try to reproduce.

lingma:
The menu calls the module manager's method to get the active module list, and the menu might be the only item to do so. Other modules are all available no matter the module is active or not. The only difference is that they disappear from the menu since this might be the only implementation.

johnsaigle:
> Other modules are all available no matter the module is active or not. The only difference is that they disappear from the menu

@driusan This sounds like it might be a bug

driusan:
Off the top of my head the menus only display items from active modules, the dashboard only displays widgets from active modules, and the candidate_profile only displays widgets from active modules. Both the candidate_profile and dashboard have subwidgets which only display information from active modules.. but even if the menus were the only place currently using it, the module still does nothing to manage menus.

If other modules have bugs, then file bug reports for those modules.

PR #6365, opened 2020-04-11
[instrument_list] control whether to show a imaging_browser link & active Module support

lingma:
## Brief summary of changes

Control whether showing an imaging_browser link depending on the user permission & the status of the `imaging_browser` module. The necessary method is added in the `Module`

1. Create a user without imaging_browser permission;
2. Login with the new user to verify whether the imaging_browser will be shown at the end of the instrument_list, which could be find if you click on a candidate profile -> timepoint -> visit;
3. The user without proper permission should not see the imaging_browser link;
4. Logout the new created user, login with one with the imaging_browser permission, check to make sure that the imaging_browser link is there;
5. Disable `imaging_browser` module, check again to see that the imaging_browser link is no longer displayed.

#6266, at the same time, the `checkActiveModules` could be used to solve part of the problem mentioned in the #6360, but the work itself is not done here.

### DIFF
--- a/modules/instrument_list/php/instrument_list.class.inc
+++ b/modules/instrument_list/php/instrument_list.class.inc
@@ -134,6 +134,31 @@ class Instrument_List extends \NDB_Menu_Filter
             // the candidate exists before getting here.
             throw new \LorisException("Timepoint list requires a timepoint.");
         }
+
+        // Check imaging_browser permission, used to control imaging_browser link
+        // on the instrument_list main page
+        $DB      =& \Database::singleton();
+        $modules = \Module::checkActiveModules($DB, ['imaging_browser']);
+
+        $perm = false;
+        if ($modules['imaging_browser']) {
+            $user = $request->getAttribute("user");
+            if ($this->candidate->getData('Entity_type') == 'Scanner') {
+                $perm = $user->hasPermission('imaging_browser_phantom_allsites')
+                || $user->hasCenterPermission(
+                    'imaging_browser_phantom_ownsite',
+                    $this->timePoint->getCenterID()
+                );
+            } elseif ($this->candidate->getData('Entity_type') == 'Human') {
+                $perm = $user->hasPermission('imaging_browser_view_allsites')
+                || $user->hasCenterPermission(
+                    'imaging_browser_view_site',
+                    $this->timePoint->getCenterID()
+                );
+            }
+        }
+        $this->tpl_data['imaging_browser_permission'] = $perm;
+
         return parent::process($request, $handler);
     }
 
--- a/php/libraries/Module.class.inc
+++ b/php/libraries/Module.class.inc
@@ -94,7 +94,7 @@ abstract class Module extends \LORIS\Router\PrefixRouter
         }
 
         if ($moduleCnt < count($modules)) {
-            // In rare occaision, the $modules might have more but never less.
+            // In some occasions, the $modules might contain more but never less.
             $keys = array_keys($moduleStatus);
             foreach ($modules as $key => $name) {
                 if (!in_array($name, $keys)) {
--- a/php/libraries/Module.class.inc
+++ b/php/libraries/Module.class.inc
@@ -63,6 +63,51 @@ abstract class Module extends \LORIS\Router\PrefixRouter
     }
 
     /**
+     * Checks whether one or more modules are active in the given database.
+     *
+     * @param \Database $db      an open connection to a database containing a
+     *                           'modules' table
+     * @param array     $modules the names of the modules to check
+     *
+     * @return array of module name => boolean (active or not)
+     */
+    static public function checkActiveModules(\Database $db, array $modules): array
+    {
+        $params = array();
+        $values = array();
+        foreach ($modules as $key => $name) {
+            $para          = "n$key";
+            $params[]      = ":$para";
+            $values[$para] = $name;
+        }
+        $list = $db->pselect(
+            "SELECT Name, Active FROM modules WHERE Name IN (" .
+            implode(', ', $params) . ")",
+            $values
+        );
+
+        $moduleStatus = [];
+        $moduleCnt    = 0;
+        foreach ($list as $item) {
+            $moduleStatus[$item['Name']] = $item['Active'] == 'Y';
+            $moduleCnt ++;
+        }
+
+        if ($moduleCnt < count($modules)) {
+            // In rare occaision, the $modules might have more but never less.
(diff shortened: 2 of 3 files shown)


php/libraries/Module.class.inc
```diff
@@ -62,6 +62,38 @@ abstract class Module extends \LORIS\Router\PrefixRouter
...
+            "SELECT Name, Active FROM modules WHERE Name IN (" .
+            implode(', ', $params) . ")",
+            $values
+        );
+
+        $checkedModules = [];
+        foreach ($list as $item) {
+            $checkedModules[$item['Name']] = ($item['Active'] == 'Y'? true : false);
+        }
+        return $checkedModules;
```
nicolasbrossard, line 94: What if `$modules` contains a module name that does not exist? Should the returned array`$checkedModules` contain a key for the invalid module name (with value `false`, obviously) or not? Example: what would `checkActiveModules($db, ['imaging_browser', 'xxxxxx'])` return?

modules/instrument_list/php/instrument_list.class.inc
```diff
@@ -345,11 +370,11 @@ class Instrument_List extends \NDB_Menu_Filter
 
         if (!is_null($stage)) {
             $query .= "
-            	LEFT JOIN test_battery b 
-       	        ON ((t.Test_name=b.Test_name OR b.Test_name IS NULL) 
-      	        AND (s.SubprojectID=b.SubprojectID OR b.SubprojectID IS NULL) 
-       	        AND (s.Visit_label=b.Visit_label OR b.Visit_label IS NULL) 
-       	        AND (s.CenterID=b.CenterID OR b.CenterID IS NULL)) ";
+                LEFT JOIN test_battery b
```
nicolasbrossard, line 373: Shouldn't we also add a clause to check whether `b.Active='Y'`?

php/libraries/Module.class.inc
lingma: I was assuming that the module name should be checked by the implementer to simplify the issue, I will check.

modules/instrument_list/php/instrument_list.class.inc
lingma: I don't think this is necessary, to not have more performance penalty.

php/libraries/Module.class.inc
```diff
@@ -62,6 +62,38 @@ abstract class Module extends \LORIS\Router\PrefixRouter
         return $modules;
     }
 
+    /**
+     * Check whether one or more module descriptors are active from the given
+     * database.
+     *
+     * @param \Database $db      an open connection to a database containing a
+     *                           'modules' table
+     * @param array     $modules a list of the interested module names to check
```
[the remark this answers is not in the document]
lingma: thanks

php/libraries/Module.class.inc
```diff
@@ -62,6 +62,38 @@ abstract class Module extends \LORIS\Router\PrefixRouter
         return $modules;
     }
 
+    /**
+     * Check whether one or more module descriptors are active from the given
```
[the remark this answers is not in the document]
lingma: ok

modules/instrument_list/php/instrument_list.class.inc
nicolasbrossard: If you do not add that check, inactive visits might be returned by this method, which is a bug. This check is definitely needed.

lingma: Oh, test_battery active or not might be another issue, nothing to do with module active or not. I can create another PR if necessary

nicolasbrossard: You are right, it does have to be part of this PR but since it is such a small change, you might as well incorporate it here. Your call. Thanks!

lingma: PR #6418 was submitted, I feel that Active = 'Y' check is not mandatory since it is in a left join, so I eventually did not add that check in this PR.

php/libraries/Module.class.inc
```diff
@@ -62,6 +62,51 @@ abstract class Module extends \LORIS\Router\PrefixRouter
...
+
+        $moduleStatus = [];
+        $moduleCnt    = 0;
+        foreach ($list as $item) {
+            $moduleStatus[$item['Name']] = $item['Active'] == 'Y';
+            $moduleCnt ++;
+        }
+
+        if ($moduleCnt < count($modules)) {
+            // In rare occaision, the $modules might have more but never less.
```
nicolasbrossard, line 97: occasions

php/libraries/Module.class.inc
```diff
@@ -62,6 +62,51 @@ abstract class Module extends \LORIS\Router\PrefixRouter
...
+    /**
+     * Checks whether one or more modules are active in the given database.
+     *
+     * @param \Database $db      an open connection to a database containing a
+     *                           'modules' table
+     * @param array     $modules the names of the modules to check
+     *
+     * @return array of module name => boolean (active or not)
+     */
+    static public function checkActiveModules(\Database $db, array $modules): array
```
driusan, line 74: This looks like it's basically duplicating the functionality of LorisInstance->hasModule from the master branch. Since it's not a critical bug fix, I think it would be better for this to go to master and use that functionality rather than duplicate the functionality in a static function that would be deprecated right from the start..

lingma: This function should be more efficient than `LorisInstance->hasModule` because it doesn't check all information. This is exactly the purpose of adding a new function. Otherwise I could have used the existing `getActiveModules` which is considered by me as not efficient enough to be useful for this PR's purpose.

driusan: What were the results when you benchmarked hasModule and getActiveModules compared to this function?

lingma: I don't have the number, `hasModule` and `getActiveModules` will loop around the whole modules, and check whether it exists or not, `checkActiveModules` only checks one or a few modules active or not in the database, this should not be in the same performance level.

lingma: I just manually did a comparison, out of two executions, my method with `checkActiveModules` will takes 0.000150442s, compared with 0.021488905s of `hasModule` and `getActiveModules`. @driusan

driusan: Assuming n=1 gives accurate statistics it's not worth the maintenance overhead of having 2 methods which do essentially the same thing in 2 different places in order to save 0.02s per page load of the imaging browser.

PR #6369, opened 2020-04-11
[Imaging Uploader] Incorrect values in upload form for phantom scans.

### LINKED ISSUE #6288: [imaging_uploader] Test plan #5 fails
**Describe the bug**
Test plan #5:
`Ensure that the 'Upload' tab has the CandID, PSCID, and Visit Label greyed out. These greyed out fields remain empty if the 'Phantom Scans' is set to Yes, and are properly filled based on the uploaded filename if 'Phantom Scans' is set to No.
[Manual Testing]`

The bug: The greyed out fields do not remain empty if the 'Phantom Scans' is set to Yes. On uploading a file, the greyed out fields are filled out according to the file name (this auto-filling should only happen if 'Phantom Scans' is No)

**To Reproduce**
1. Go to 'Imaging Uploader' module
2. Click on 'Upload' tab
3. Fill in the form: 'Phantom Scans': Yes
4. Upload a file into 'File to Upload' field, with any name but with correct file types .tgz or tar.gz or .zip
5. You will see that the CandID, PSCID, and Visit Label fields are auto-populated according to the file name (this shouldn't be the case).

See #5 of test plan

nicolasbrossard:
## Brief summary of changes

This PR corrects the failures or these test cases:

#### Test case 1:

1. Access the Imaging uploader module.
2. Select 'Yes' in the phantom scan drop-down.
3. Select a file with a name of the form `PSCID_CANDID_VISITLABEL.tar.gz`.

You'll that the CandID, PSCID and visit label field get populated even though they should be left empty.

#### Test case 2:

1. Access the Imaging uploader module.
2. Select 'No' in the phantom scan drop-down.
3. Select a file with a name of the form `PSCID_CANDID_VISITLABEL.tar.gz` (the CandID, PS and visit label fields will be populated, which is all right).
4. Select 'Yes' in the phantom scan drop-down.

You'll that the CandID, PSCID and visit label field stay populated even though they should be erased.

See above

* Resolves #6288

### DIFF
--- a/modules/imaging_uploader/jsx/UploadForm.js
+++ b/modules/imaging_uploader/jsx/UploadForm.js
@@ -51,13 +51,19 @@ class UploadForm extends Component {
         delete formData.candID;
         delete formData.pSCID;
         delete formData.visitLabel;
+      } else if (typeof formData.mriFile !== 'undefined') {
+        let patientName = formData.mriFile.name.replace(/\.[a-z]+\.?[a-z]+?$/i, '');
+        let ids = patientName.split('_', 3);
+        formData.candID = ids[1];
+        formData.pSCID = ids[0];
+        formData.visitLabel = ids[2];
       }
     }
 
     formData[field] = value;
 
     if (field === 'mriFile') {
-      if (value.name !== '') {
+      if (value.name !== '' && formData.IsPhantom === 'N') {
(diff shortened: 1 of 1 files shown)


jesscall:
looks good to me in both test cases!

PR #6371, opened 2020-04-13
[Statistics / Demographic] fix issue where users were able to see other projects in the drop-down menu.

### LINKED ISSUE #6239: [statistics] user has one project's permission still can see other projects in the drop-down menu
**Describe the bug**
This is a minor issue.
In Demographic, Behavioural Statistics
and Imaging Statistics tab, although the user has only one project's permission, the use still can see other projects in the project drop-down menu.
**To Reproduce**
1. assign a user with one project premisson.
2. go to statistics/Demographic Statistics
3. select project 
4. See all the projects in the dropdown.

racostas:
Fix for` Statistics --> Demographic `where a user with the permission of a single project was able to see other projects in the project drop-down menu.
## Brief summary of changes

- The project drop-down menu was filtered to only show the projects for the logged user. 
- The results were filtered as well for the case "All Projects" is selected only allowing "All Projects for the logged user"  

1. Assign a user with one project permission.
2. go to statistics/Demographic Statistics
3. select project
4. See only the allowed projects in the drop-down menu.
5. If the option "All Projects" is selected the results should reflect only "All Projects for the logged user" 

#### Note1:
To think if it will be good to add a new function to the `User class` for retrieving the projects for a given user the same way it exist for the user sites.

#### Note2:
This PR only solves the ` Statistics --> Demographic `. As the solution for` Behavioural Statistics`
and` Imaging Statistics` tab is very similar should be fixed once the present solution  approach approved. 

* Resolves #6239

### DIFF
--- a/modules/statistics/php/stats_demographic.class.inc
+++ b/modules/statistics/php/stats_demographic.class.inc
@@ -182,7 +182,7 @@ class Stats_Demographic extends \NDB_Form
         $projects[''] = 'All Projects';
 
         $userProjectsIDs = $user->getData('ProjectIDs');
-        foreach ($userProjectsIDs as $key => $ProjectID) {
+        foreach ($userProjectsIDs as $ProjectID) {
             $projects[$ProjectID] = $allProjectList[$ProjectID];
         }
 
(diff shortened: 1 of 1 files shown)


sruthymathew123:
Looks good to me; Projects which user is assigned to is visible in the dropdown list.

PR #6372, opened 2020-04-14
[MRI Violations] MRI protocol table is now read-only.

### LINKED ISSUE #6310: [MRI Violations] Modifications to mri_protocol table are not saved
**Describe the bug**
I access the MRI violations module as a user that has the permission to edit the MRI protocol. I make a valid edit of the MRI protocol table and the system asks if I am sure I want to proceed. I confirm that I want to and my modification seems to be saved but if I access another page (say, Access Profile) and go back to the MRI protocol edit page afterwards I see that it was not.

I think this happens no matter what column of the `mri_protocol` table you try to edit.

nicolasbrossard:
## Brief summary of changes

This PR makes the `mri_protocol` table read-only in the MRI violations module.

Access the MRI violations module and search for problems of type 'Could not identify scan type'. Verify that the `mri_protocol` table on the new page displayed is read-only.

* Resolves #6310


nicolasbrossard:
@johnsaigle Sure. Done. I also removed the 'Needs Documentation' label and added the 'Testing' label. Hope that's appropriate.

modules/mri_violations/templates/menu_mri_protocol_violations.tpl
```diff
@@ -107,11 +107,7 @@
                     {foreach from=$mpgroup item=protocol}
                         <tr>
                         {foreach from=$protocol key=k item=v}   
-                            {if $violated_scans_modifications && $k != 'Protocol Group'}
-                            <td id="row_{$protocol.ID}_td_{$k}" class='description' contenteditable = "true" nowrap {$rowBorder}>
-                            {else}
```
driusan, line 112: Can you delete the other code too?

1. The script in the ajax directory is now unnecessary
2. The js in `js/mri_protocol_violations.js` related to this feature

nicolasbrossard: Done.

christinerogers:
i don't think the documentation about this feature was update when it was removed

Issue #6328, opened 2020-04-07
[Issue Tracker] Dropdown options display order is alpha, not array-defined


christinerogers:
The `Priority` selection filter in the Issue Tracker presents dropdown options in alphabetical order, but should display them in the same order [defined in the php array](https://github.com/aces/Loris/blob/master/modules/issue_tracker/php/issue_tracker.class.inc#L122) - the order of priority.  See code snippet below

I suspect this is a known issue **_in multiple modules_** with an easy fix.

Is there already a way we should be forcing the order of these dropdowns?  
cc @johnsaigle @HenriRabalais 

Occurs with Chrome, Firefox, Safari... on Demo and on test-dev-loris(23 branch), tested on mac.

Order [defined in the php array](https://github.com/aces/Loris/blob/master/modules/issue_tracker/php/issue_tracker.class.inc#L122):
```php
        $priorities = array(
            'normal'    => 'Normal',
            'high'      => 'High',
            'urgent'    => 'Urgent',
            'immediate' => 'Immediate',
        );
```

Order as actually displayed : 
![issueTracker_PriorityOrder](https://user-images.githubusercontent.com/827619/78678347-cef89500-78b6-11ea-9abd-240c4da04278.png)

Related (sort of) to: #4602 - see @johnsaigle's [comment recognizing this sort-order behaviour/implication](https://github.com/aces/Loris/issues/4602#issuecomment-492743380) in 2019

johnsaigle:
As you said this should probably be case-by-case. Maybe there is a generic way to specify sorting for dropdown elements within React? @HenriRabalais @maltheism @zaliqarosli maybe one of you would know?

HenriRabalais:
IIRC, @ridz1208 had made a PR to enforce alphabetical ordering. I'm not sure if there's a way to override this at the moment.

johnsaigle:
Is there a way to extend/modify a component so that a sort ordering can be specified?

ridz1208:
@HenriRabalais yeah I had but I forgot how to use it. I do remember adding that though

HenriRabalais:
Looks like there might be - it's not commented in the code, but I think the props `sortByValue` determines if the drop down values are sorted alphabetically. The value defaults to `true`, so a value of `false` would have to be passed to the `SelectElement` for the prop `sortByValue`.

https://github.com/aces/Loris/blob/master/jsx/Form.js#L309

johnsaigle:
[...]

PR #6376, opened 2020-04-14
[Issue Tracker] Dropdown options display order

xlecours:
Fixes https://github.com/aces/Loris/issues/6328

### DIFF
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -538,7 +538,7 @@ SelectElement.defaultProps = {
   multiple: false,
   disabled: false,
   required: false,
-  sortByValue: true,
+  sortByValue: false,
   emptyOption: true,
   hasError: false,
   errorMessage: 'The field is required!',
--- a/modules/issue_tracker/php/issue_tracker.class.inc
+++ b/modules/issue_tracker/php/issue_tracker.class.inc
@@ -115,7 +115,10 @@ class Issue_Tracker extends \NDB_Menu_Filter_Form
             $assignees[$a_row['UserID']] = $a_row['Real_name'];
         }
 
-        $modules = \Module::getActiveModules($db);
+        $modules_names  = array_map(function(\Module $m): string {
+            return $m->getLongName();
+        },\Module::getActiveModules($db));
+        $modules = array_combine($modules_names, $modules_names);
 
         $priorities = array(
             'normal'    => 'Normal',
--- a/modules/issue_tracker/php/issue_tracker.class.inc
+++ b/modules/issue_tracker/php/issue_tracker.class.inc
@@ -115,10 +115,17 @@ class Issue_Tracker extends \NDB_Menu_Filter_Form
             $assignees[$a_row['UserID']] = $a_row['Real_name'];
         }
 
-        $modules_names  = array_map(function(\Module $m): string {
-            return $m->getLongName();
-        },\Module::getActiveModules($db));
-        $modules = array_combine($modules_names, $modules_names);
+        // Create an array where keys and values are the modules long name to be used
+        // for the module filter dropdown options.
+        $modules = array_reduce(
+            \Module::getActiveModules($db),
+            function ($carry, $module): array {
+                $name         = $module->getLongName();
+                $carry[$name] = $name;
+                return $carry;
+            },
+            []
+        );
 
         $priorities = array(
             'normal'    => 'Normal',


xlecours:
Merging this might bring more bugs to light though. 
The sorting had the side effect of making sure the options were an array. 
I had to fix the module dropdown filter in this module because of that.

modules/issue_tracker/php/issue_tracker.class.inc
```diff
@@ -115,7 +115,10 @@ class Issue_Tracker extends \NDB_Menu_Filter_Form
             $assignees[$a_row['UserID']] = $a_row['Real_name'];
         }
 
-        $modules = \Module::getActiveModules($db);
+        $modules_names  = array_map(function(\Module $m): string {
+            return $m->getLongName();
+        },\Module::getActiveModules($db));
+        $modules = array_combine($modules_names, $modules_names);
```
driusan, line 121: This line seems a little.. opaque. Why are you combining `modules_names` with itself?

xlecours: Because the datatable is filtering on the cell content which is the module lang name. Therefor, array combine creates an array where keys and values are the module long name for the dropdown options.

I could do all that in the array_map though.

Do you want me to change it?

driusan: I think at least add a comment explaining it.

driusan:
@xlecours I think the branch is broken, not your PR. See the webdriver PR thread..

xlecours:
ok, thanks @driusan 
I will look at it again when https://github.com/aces/Loris/pull/6392 is resolved.

driusan:
@xlecours have you rebased this since the 23.0-release branch was fixed?

xlecours:
@driusan I did earlier today but Travis do not trigger. I will try adding a change.

johnsaigle:
>There was 1 failure:
1) CandidateListTestIntegrationTest::testFilters
Failed asserting that '20 rows displayed of 210. (Maximum rows per page:\n
20\n
50\n
100\n
1000\n
5000\n
10000\n
)' contains "20 rows displayed of 225".


Not sure if this is related to the Form.js changes or unrelated.

driusan:
I don't see any changes in this PR that should change the number of rows

ridz1208:
@driusan my guess would be that the order of the options in the filter have changed and the test is using an absolute value in the filter (ie `option[1]`) to test the filter. since the options were reordered, we are effectively now testing for another selection then before

maltheism:
@johnsaigle @driusan @ridz1208 there isn't even a sex filter in the candidate_list module?

jsx/Form.js
```diff
@@ -538,7 +538,7 @@ SelectElement.defaultProps = {
   multiple: false,
   disabled: false,
   required: false,
-  sortByValue: true,
+  sortByValue: false,
```
driusan, line 541: I don't think this default should be changed. Sorting a sensible default. Changing this PR so it's just the issue tracker priority prop that's changed would likely fix Travis.

christinerogers: There's [a number of other places](https://github.com/aces/Loris/issues/6328#issuecomment-610599920) (at least) where the sort-order is having an unintended impact.  
Note -- given we limit this PR to the Issue tracker priority, tickets should be created to address these other sorting dropdowns individually.  And should be documented in our "how to build a module" docs

johnsaigle: I think ti would be a good idea to make those tickets. They would make for good Beginner/GSOC tickets.

kongtiaowang:
Travis is failing, because of the dropdown menu order. 
change [
//female
$this-> _testFilter(self::$sex, "20 rows displayed of 225", '1');
// male
 $this-> _testFilter(self::$sex, "20 rows displayed of 210", '2');
]  to [
//female
$this-> _testFilter(self::$sex, "20 rows displayed of 225", '**2**');
// male
 $this-> _testFilter(self::$sex, "20 rows displayed of 210", '**1**');
]

kongtiaowang:
Since the order is sorted by value, is it possible to fix it this way？
        $priorities = array(
            'normal'       => '1 - Normal',
            'high'           => '2 - High',
            'urgent'       => '3 - Urgent',
            'immediate' => '4-  Immediate',
        );

ridz1208:
@xlecours can you please restore the default to true and try to fix travis for this to pass ?

jsx/Form.js
HenriRabalais: I agree with Dave

HenriRabalais:
Issued new PR to resolve the issue #6478

ridz1208:
@xlecours closing this one in favour of new PR

Issue #6218, opened 2020-04-01
[Instrument List] Mysterious checkboxes require documentation


lingma:
**Describe the bug**
See the image attached: 
![image](https://user-images.githubusercontent.com/5295785/78158254-feb72100-740e-11ea-8d2a-6780d830b40c.png)

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to instrument list of any candidate
2. Click on "double data entry"
3. Look at the left top corner

**What did you expect to happen?**
There should be a text to explain what those buttons are

**Browser Environment (please complete the following information):**
- Chrome

**Server Environment (if known):**
- LORIS Version: 23-release

lingma:
It is better to have a help text which could popup when putting the mouse over there.

Another way to do is that the concept of the 2 checkbox has to be modified to be simple.

christinerogers:
Update: To be resolved in a new PR by @pierre-p-s 

in order to implement the solution discussed in #6225 by @driusan and @pierre-p-s


-- including [this update for swal2](https://github.com/aces/Loris/pull/6225#issuecomment-613476175) suggested by @haoweiqiu (Thanks Haowei!) :

> However, I think we are turning to use SweetAlert2 now. Here's my tweak on your code.
```
<script>
    function swalFunction() {
    Swal.fire({
      title: 'Please confirm deletion',
      text: 'The instrument data will be deleted',
      type: 'warning',
      showCancelButton: true,
      confirmButtonColor: '#3085d6',
      cancelButtonColor: '#d33',
      confirmButtonText: 'Yes, delete the data'
    }).then((result) => {
  if (result.value) {
    Swal.fire(
      'Deleted!',
      'The instrument data has been deleted.',
      'success'
    )
  }
});
  }
</script>
 ```

PR #6378, opened 2020-04-14
[Instrument] Delete Instrument Data

pierre-p-s:
## Brief summary of changes

- Removed the two unlabeled checkboxes
- Added a modal window for confirmation of data deletion
- Changed the condition in the NDB_BVL_InstrumentStatus_ControlPanel.class.inc file

#### Testing instructions 

Test the confirm button and the cancel button. Make sure the data is deleted (you need to reload the page)

#6218 

Thank you @maltheism  and @haoweiqiu  for the help.

### DIFF
--- a/htdocs/js/instrument_controlpanel_control.js
+++ b/htdocs/js/instrument_controlpanel_control.js
@@ -1,21 +1,3 @@
 $(function() {
     $('[data-toggle="tooltip"]').tooltip();
 });
-
-function swalFunction(event) {
-  swal({
-    title: 'Please confirm deletion',
-    text: 'The instrument data will be deleted',
-    type: 'warning',
-    showCancelButton: true,
-    confirmButtonColor: '#3085d6',
-    cancelButtonColor: '#d33',
-    confirmButtonText: 'Yes, delete the data'
-  }, function(result) {
-    if(result) {
-      // swal does not work here. The window instantly disappears
-      alert("\t\t\t Deleted! (You will need to reload the page again)");
-      document.getElementById('ConfirmDelete').submit();
-    }
-  });
-}
--- a/htdocs/js/instrument_controlpanel_control.js
+++ b/htdocs/js/instrument_controlpanel_control.js
@@ -1,3 +1,21 @@
 $(function() {
     $('[data-toggle="tooltip"]').tooltip();
 });
+
+function swalFunction(event) {
+  swal({
+    title: 'Please confirm deletion',
+    text: 'The instrument data will be deleted',
+    type: 'warning',
+    showCancelButton: true,
+    confirmButtonColor: '#3085d6',
+    cancelButtonColor: '#d33',
+    confirmButtonText: 'Yes, delete the data'
+  }, function(result) {
+    if(result) {
+      // swal does not work here. The window instantly disappears
+      alert("\t\t\t Deleted! (You will need to reload the page again)");
+      document.getElementById('ConfirmDelete').submit();
+    }
+  });
+}
--- a/htdocs/js/instrument_controlpanel_swaldeletedata.js
+++ b/htdocs/js/instrument_controlpanel_swaldeletedata.js
@@ -9,8 +9,6 @@ function swalFunction(event) {
     confirmButtonText: 'Yes, delete the data',
   }, function(result) {
     if (result) {
-      // swal does not work here. The window instantly disappears
-      alert('\t\t\t Deleted! (You will need to reload the page again)');
       document.getElementById('ConfirmDelete').submit();
     }
   });
--- a/htdocs/js/instrument_controlpanel_swaldeletedata.js
+++ b/htdocs/js/instrument_controlpanel_swaldeletedata.js
@@ -1,0 +1,19 @@
+function swalFunction(event) {
+  swal({
+    title: 'Please confirm deletion',
+    text: 'The instrument data will be deleted',
+    type: 'warning',
+    showCancelButton: true,
+    confirmButtonColor: '#3085d6',
+    cancelButtonColor: '#d33',
+    confirmButtonText: 'Yes, delete the data',
+  }, function(result) {
+    if (result) {
+      // swal does not work here. The window instantly disappears
+      alert('\t\t\t Deleted! (You will need to reload the page again)');
+      document.getElementById('ConfirmDelete').submit();
+    }
+  });
+}
+
+document.getElementById('ConfirmDelete').onclick = swalFunction;
--- a/php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
+++ b/php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
@@ -157,6 +157,7 @@ class NDB_BVL_InstrumentStatus_ControlPanel extends NDB_BVL_InstrumentStatus
             return false;
         }
         if (isset($_POST['ClearInstrument'])) {
+            header("Refresh: 0;");
             $user = User::singleton();
             if ($user->hasPermission('send_to_dcc')) {
                 $instrument = \NDB_BVL_Instrument::factory(
--- a/php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
+++ b/php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
@@ -158,7 +158,7 @@ class NDB_BVL_InstrumentStatus_ControlPanel extends NDB_BVL_InstrumentStatus
         }
 
         if (isset($_POST['ClearInstrument'])) {
-            if (isset($_POST['deleteconfirm']) && isset($_POST['deleteconfirm2']) ) {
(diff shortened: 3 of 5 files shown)


smarty/templates/instrumentstatus_controlpanel.tpl
```diff
@@ -1,67 +1,84 @@
 <script type="text/javascript" src="{$baseurl}/js/instrument_controlpanel_control.js"></script>
-
+<script type="text/javascript" src="{$baseurl}/htdocs/vendor/sweetalert/sweetalert.js"></script>
```
maltheism, line 2: This line for including the sweetalert.js may be unnecessary.

driusan, line 2: This should not be here, it should come from getJSDependencies of the place using it.

smarty/templates/instrumentstatus_controlpanel.tpl
```diff
@@ -1,67 +1,84 @@
...
+{literal}
+    <script type="text/javascript">
+        <!--
+        function swalFunction(event) {
```
driusan, line 6: This should be in a javascript file of the module, not inline.

johnsaigle:
Failing Travis due to JSLint errors so I'm marking this as `Needs Code Formatting`.

There are also "FirefoxProfile" errors which should be fixed on the most recent release branch. This `Needs Rebase` to pass.

pierre-p-s:
> Failing Travis due to JSLint errors so I'm marking this as `Needs Code Formatting`.
> 
> There are also "FirefoxProfile" errors which should be fixed on the most recent release branch. This `Needs Rebase` to pass.

Travis should now run without errors.
Also rebased the PR.

smarty/templates/instrumentstatus_controlpanel.tpl
pierre-p-s: Created a js file

smarty/templates/instrumentstatus_controlpanel.tpl
pierre-p-s: deleted the line

pierre-p-s: deleted the line (was giving an error)

htdocs/js/instrument_controlpanel_control.js
```diff
@@ -1,3 +1,21 @@
 $(function() {
     $('[data-toggle="tooltip"]').tooltip();
 });
+
+function swalFunction(event) {
```
maltheism, line 5: @pierre-p-s does the swalFunction need to be in both files:
`htdocs/js/instrument_controlpanel_control.js`
`htdocs/js/instrument_controlpanel_swaldeletedata.js `

pierre-p-s: No I forgot to remove it

pierre-p-s: removed

htdocs/js/instrument_controlpanel_swaldeletedata.js
```diff
@@ -0,0 +1,19 @@
...
+    text: 'The instrument data will be deleted',
+    type: 'warning',
+    showCancelButton: true,
+    confirmButtonColor: '#3085d6',
+    cancelButtonColor: '#d33',
+    confirmButtonText: 'Yes, delete the data',
+  }, function(result) {
+    if (result) {
+      // swal does not work here. The window instantly disappears
+      alert('\t\t\t Deleted! (You will need to reload the page again)');
```
maltheism, line 13: The alert should be removed.

pierre-p-s: yes but the data still appears. You need to reload another time for it to dissappear.

maltheism: @pierre-p-s I didn't notice that when testing.
Maybe have it as a note in the sweetalert dialogue for deleting. The alert function shouldn't be used. I'll also test again to see if there is something that can be done about the data.

PapillonMcGill: When I tried (Firefox), the page reloaded automatically with previous data. Manually reloading the page as ask in the modal window, cleared the old data.

maltheism: I'm getting that behaviour as well. I still don't think an alert function should be used in Loris the behaviour is kind of frowned upon in the design world.

pierre-p-s: Ill get rid of the alert and add the text to the swal modal window

maltheism: @prioux hold on.

Try moving the following code in Loris/src/Middleware/UserPageDecorationMiddleware.php

```
// This should be moved out of the middleware and into the modules that need it,
        // but is currently required for backwards compatibility.
        // This should also come after the above call to handle() in order for updated data
        // on the controlPanel to be properly displayed.
        if (method_exists($page, 'getControlPanel')) {
            $tpl_data['control_panel'] = $page->getControlPanel();
        }
```

To line 154 and the refresh won't be necessary.

johnsaigle: @maltheism You tagged the wrong pierre - you want @pierre-p-s  not Pierre Rioux

maltheism: @johnsaigle oops, @pierre-p-s look at the code fix above for the refresh problem.

pierre-p-s: @maltheism The fix you proposed did not work.

maltheism: @pierre-p-s did you remove the old code that was moved?

pierre-p-s: for me when I add it to the .class.inc file and delete it from Loris/src/Middleware/UserPageDecorationMiddleware.php the control panel disappears completely.

maltheism: @pierre-p-s can you message me on slack because I would like to replicate.

pierre-p-s: found the problem. Will be fixed tomorrow

pierre-p-s: Added the fix. Can be tested and reviewed again

src/Middleware/UserPageDecorationMiddleware.php
```diff
@@ -151,6 +151,14 @@ function ($a, $b) {
             );
         }
 
+        // This should be moved out of the middleware and into the modules that need it,
```
driusan, line 154: Why are you moving this? I don't see why this PR should touch any middleware

maltheism: Hi @driusan the new placement here, resolves the data not being refreshed and specifically when it's being deleted. Otherwise the browser had to be refreshed (unnecessary) for the data that was deleted to be displayed correctly as gone.

driusan: I don't think that should be within the scope of this PR and the comment leads me to believe that it's likely to introduce bugs in other modules or features. A simpler (and less risky) solution is to check if "ClearInstrument" it's a POST request with a 'ClearInstrument' element set and include a Refresh header in the response or a "205 Reset Content" response code (which might need to be handled properly in js)

pierre-p-s:
@driusan I used a refresh header. Please let me know if anything else needs to change.

php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
```diff
@@ -156,21 +156,17 @@ class NDB_BVL_InstrumentStatus_ControlPanel extends NDB_BVL_InstrumentStatus
...
-                        $this->testname,
-                        $this->_commentID,
-                        $this->subtest ?? ''
-                    );
-                    $instrument->clearInstrument();
-                    return false;
-                } else {
-                    return false;
-                }
+            header("Refresh: 0;");
```
driusan, line 160: This should not be using functions that create side effects on the output stream like `header`.

ridz1208:
@pierre-p-s please address review

php/libraries/NDB_BVL_InstrumentStatus_ControlPanel.class.inc
pierre-p-s: Where should the refresh statement be?

driusan: It should not be.

pierre-p-s: I deleted the reset header. This issue is outside of this PR's scope as you mentioned above.

PR #6384, opened 2020-04-15
[document_repository] fix no file under root folder 

kongtiaowang:
## Brief summary of changes
fix no file under root folder
load the page, it will show all the files and all categories in the table.
compare with https://demo.loris.ca/document_repository/

https://github.com/aces/Loris/issues/6405

### DIFF
--- a/modules/document_repository/jsx/docIndex.js
+++ b/modules/document_repository/jsx/docIndex.js
@@ -84,9 +84,10 @@ class DocIndex extends React.Component {
         let fillData = filterData.filter((data) => {
           return Object.values(nodesArray).includes(data[10]);
         });
+
         this.setState({
           global: false,
-          tableData: fillData,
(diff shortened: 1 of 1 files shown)


pierre-p-s:
This is what the test plan says: 
`10. Check that when performing a search without global filtering, the categories
   containing the resulting files (and only those) are shown in a list format. 
   [Manual Testing]`
Displaying all the files under the root folder does not solve it. From the test plan, if we search for 'test.pdf', since this pdf file is under the category 'pdf' then only this category should appear.

@johnsaigle is this an issue with the test plan or the code please?

kongtiaowang:
@pierre-p-s you are right, this PR fix another issue, not your issue.

kongtiaowang:
I merged this PR by accident. replace by https://github.com/aces/Loris/pull/6384

PR #5827, opened 2019-12-04
[Genomic Browser] Reformat README to be in the same style as other module specs

driusan:
This reformats the information in the genomic_browser README to follow the same style
as the other modules specs. Information for required sections that was not present in the existing README or that I couldn't derived from the code is listed as "Unknown". Edits from
people who are familiar with the module to fill in those sections is welcome.

PR #6390, opened 2020-04-15
[GenomicBrowser] Readme missing sections added 

### LINKED ISSUE #6389: [Genomic Browser]  README has incomplete sections
The README also has incomplete sections that need to be fleshed out since #5827 was merged

_Originally posted by @driusan in https://github.com/aces/Loris/issues/4753#issuecomment-571168909_

christinerogers:
Resolves #6389 

This PR adds content for the Scope and Intended users section, which have been empty since added in #5827

### DIFF
--- a/modules/genomic_browser/README.md
+++ b/modules/genomic_browser/README.md
@@ -2,7 +2,7 @@
 
 ## Purpose
 
-The Genomic Growser module is intended to allow users to visualize the
+The Genomic Browser module is intended to allow users to visualize the
 results of already analyzed genetic tests cross-linked with behavioural
 and imaging candidate data.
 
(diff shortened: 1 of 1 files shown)


christinerogers:
requesting @johnsaigle review  - see last item under Configuration-  the spaces in markdown paths seem to make these unstable for github linking.  See any solutions?

johnsaigle:
@christinerogers Yeah internal GitHub linking seems to conflict with ReadTheDocs linking. One needs spaces and the other needs URL-encoded spaces (`%20`) to work. We should probably just remove the spaces from the file names at some point. It will be a fairly quick clean up when we do do it.

christinerogers:
Thanks for clarifying @johnsaigle  -- i'll add a ticket.

christinerogers:
@johnsaigle Hmmm... if all headers within a document are handled similarly, this may get awkward.

Issue #6301, opened 2020-04-06
[Genomic Browser] Instructions in Test Plan do not match Module.


HenriRabalais:
**Describe the bug**
In the SNP tab of the Genomic browser, the columns of the datatable do not match the columns in Testing Plan for both "Summary Fields" and "All Fields".

Test Plan Reference: https://github.com/aces/Loris/blob/master/modules/genomic_browser/test/TestPlan.md#snp-tab

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to Test Plan for the Genomic Browser under section Filtering->SNP Tab
2. Go to SNP Tab of Genomic Browser
3. Compare columns listed in Test Plan and columns in the module.

johnsaigle:
Related to #4753

ridz1208:
@christinerogers will you be able to address this ?

christinerogers:
@ridz1208 this is addressed in #6391 - just needs the requested reviewers to look at it. They've been sent a reminder today.

Issue #4753, opened 2019-05-31
[Genomic Browser] Test Plan has incomplete sections


johnsaigle:
There is no information for `Tooltips` or `Data Upload`; both just say under Construction.

Someone with knowledge of these modules should flesh this out.

driusan:
The README also has incomplete sections that need to be fleshed out since #5827 was merged

ridz1208:
@christinerogers will you be able to fix this ?

christinerogers:
Xavier has the most knowledge of these but yes, I'll be able to take a stab.

ridz1208:
@christine please do and take a stab at the linked issue #6301

christinerogers:
@ridz1208 this and #6301 are addressed in #6391 - just needs the requested reviewers to look at it. They've been sent a reminder today.

Issue #6297, opened 2020-04-06
[Genomic Browser] Instructions unclear regarding requirements for file upload.


HenriRabalais:
**Describe the bug**
The requirements that need to be met for the file to uploaded in the Files Tab of the Genomic Broswer are unclear. For this reason, it is difficult to successfully upload a file.

Test Plan Reference: https://github.com/aces/Loris/blob/master/modules/genomic_browser/test/TestPlan.md#data-upload

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to Files Tab of Genomic Browser
2. Click on Upload File.

Front-end Instructions:
<img width="600" alt="Screen Shot 2020-04-06 at 14 45 30" src="https://user-images.githubusercontent.com/7550451/78593839-45dd5180-7815-11ea-887d-e8c3c0e875b6.png">

Test Plan Instructions:
<img width="975" alt="Screen Shot 2020-04-06 at 14 46 39" src="https://user-images.githubusercontent.com/7550451/78593937-6e654b80-7815-11ea-886e-040b04b6ee86.png">

ridz1208:
@christinerogers will you be able to address this ?

christinerogers:
@ridz1208 this is addressed in #6391 - just needs the requested reviewers to look at it.  They've been sent a reminder today.

PR #6391, opened 2020-04-15
[Genomic Browser] TestPlan tweak Format, Uploading, Help

christinerogers:
Refines some of the Genomic Browser Test plan 

Resolves: #4753 
Resolves: #6301 
and some of: #6297

#### Testing instructions 
* Does it read better for testing, @HenriRabalais ?


christinerogers:
@HenriRabalais @xlecours could you review this PR before the next round of testing for the 23 release? thanks!

modules/genomic_browser/test/TestPlan.md
```diff
@@ -286,22 +283,17 @@ The following columns should be presented
...
 
-> Under construction
-
-<a name="help_link">
-</a>
-
-## Help
-
-***
+Files may be uploaded under the Files tab (Genomic File Uploader subpage) IFF the Genomic Browser back-end is properly customized/configured and sample files of the appropriate format are available for upload testing.
```
maltheism, line 294: `IFF`

johnsaigle: Stands for "if and only if". I think it's probably better to spell it out rather than rely on this abbreviation. It's come up elsewhere that people don't know what it means.

maltheism: Ahh, I was confused by this last week as well. I'm not sure if "IFF" is popular and I just never have encountered it in my life up until now. Maybe it's alright? @christinerogers

johnsaigle: @maltheism I've come across it in academic contexts but not since undergrad... I added an issue to change this in our docs #6420

PR #6187, opened 2020-03-25
Rename webdriver package in composer.json

johnsaigle:
## Brief summary of changes

The `facebook/webdriver` package has been changed/renamed to `php-webdriver/webdriver` and as a result our current composer set up is causing errors and warnings in some cases.

This PR corrects the package name. It also uses a tagged release from that repo instead of relying on the master branch, which seems to cause issues on composer's side (see https://github.com/composer/composer/issues/7940)

- [x] Have you updated related documentation?

#### Testing instructions (if applicable)

1. Checkout this branch, try running `make dev`.

**If you get an error about PHP and curl**, make sure you have the CURL extension enabled for your command line version of PHP.
    - To fix this on my Ubuntu VM, I ran: `sudo apt install php7.3-curl` and modified `vim /etc/php/7.3/cli/php.ini` by searching for "curl" and uncommenting the line I found.

**If you get an error from composer about 'unpushed changes'**, try:
- make clean
- composer remove facebook/webdriver
- composer update php-webdriver/webdriver

#### Link(s) to related issue(s)

* Reported in #6185 (but this does not resolve it as there is another related error)
* Resolves #6220

PR #6392, opened 2020-04-16
[Travis] fix webdriver

kongtiaowang:
## Brief summary of changes

Fix this PR https://github.com/aces/Loris/pull/6187

Travis

all PRs after https://github.com/aces/Loris/pull/6187


lingma:
just say final try, then it will have to work

.phan/config.php
```diff
@@ -34,10 +34,17 @@
     "unused_variable_detection" => false,
     "suppress_issue_types" => [
         "PhanTypePossiblyInvalidDimOffset",
-        "PhanUndeclaredMethod",
         "PhanTypeMismatchArgument",
         "PhanTypeMismatchProperty",
         "PhanTypeArraySuspiciousNullable",
+        "PhanTypeMismatchReturn",
```
driusan, line 40: We shouldn't be adding to the suppressed types in phan. Every one of these is a bug that's being caught by static analysis. I sent #6404 to revert the change to webdriver. I think after I merge #6404 this PR should be updated to revert that change and then fix this without regressions to our phan rules.

kongtiaowang:
wrong ropo

PR #6395, opened 2020-04-16
[LorisForm] Form static element bug in group mode

lingma:
## Brief summary of changes

This is the recreation of the PR #5884 first commit [e0ffb13](https://github.com/aces/Loris/pull/5884/commits/e0ffb13fb558ad8cde4d5c200e6747ed0daa3f88)

The third parameter of the form elements in the group mode was set as the fourth parameter somehow for `static` text Form type, this is clearly a bug.

I noticed that the [wiki](https://github.com/aces/Loris/wiki/Instrument-Groups) manual suggests to use the format:
$group[] =& $this->form->createElement("static", null, null, "label");

I suggest to use the new format to be consistent with other types:

Some old instruments will stop displaying correctly once being migrated, some verification required, besides possibly some function signatures changes including setup, score, etc., which is beyond the scope of this PR.

The instrument migration fix could be easy with a full text replacement with further verification. Some project could use module overrides to handle.

### DIFF
--- a/php/libraries/LorisForm.class.inc
+++ b/php/libraries/LorisForm.class.inc
@@ -1960,10 +1960,7 @@ class LorisForm
         case 'submit':
             return $this->createSubmit($elname, $label, $attribs);
         case 'static':
-            /* Label seems to usually be the wrong attribute for static
-             * elements?
-             */
-            $el         = $this->createBase($elname, $options, $attribs);
+            $el         = $this->createBase($elname, $label, $attribs);
             $el['type'] = 'static';
             break;
         case 'advcheckbox':


lingma:
Note: The wiki page mentioned above needs to be updated once this PR accepted.

lingma:
I just noticed one line in the code: `Label seems to usually be the wrong attribute for static elements?` I am wondering whether the `options` is set to `label` in purpose? For what purpose?

lingma:
The change is to pass the `label` (third parameter) to the second parameter instead of the `options` which is usually the forth parameter. I believe that passing options to the position of label is unacceptable, unfortunately as the comment says, it seems someone did this on purpose, I don't know why.

driusan:
I'm not 100% convinced there wasn't a reason for this given that there was a comment describing it, but I don't know of one and it looks reasonable so I guess we'll find out..

ridz1208:
@driusan let's make sure that someone tests this in the next round. generally instruments are not rigurously tested and this might have an impact on all projects

Issue #6326, opened 2020-04-07
[Candidate list] Subproject header in candidate list sorts by ID instead of subproject


PapillonMcGill:
**Describe the bug**
clicking on the Subproject header in candidate list sort by ID instead of subproject 

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to 'Candidate -> Access profile'
2. Click on 'Subproject' header
3. Candidate are sorted by ID (first column) not Subproject

**What did you expect to happen?**
According to Test-plan should be sorted by Subproject
Not sure if test-plan is not up to date as I remember there was a discussion about the fact that some candidate are in multiple projects

**Additional context**
V23.0.0 testing

PR #6403, opened 2020-04-16
[Candidate list] Fixed Subproject header in candidate list sorting

maltheism:
## Brief summary of changes

Allowed sorting by array value for the DataTable.js

1. Checkout PR.
2. Sort by clicking on the table header of Subproject

https://github.com/aces/Loris/issues/6326

### DIFF
--- a/jsx/DataTable.js
+++ b/jsx/DataTable.js
@@ -195,6 +195,8 @@ class DataTable extends Component {
       } else if (isString) {
         // if string with text convert to lowercase
         val = val.toLowerCase();
+      } else if (Array.isArray(val)) {
+        val = val.join(', ');
       } else {
         val = undefined;
       }


maltheism:
@HenriRabalais it took me awhile to debug this issue.

The "value" for Subproject is actually an array from the server because there can be multiple Subprojects. The formatColumn does a cell.join for the Subproject array value and to make it a string for example displaying `"Fresh, Stale, Low Yeast"` from previously being `['Fresh', 'Stale', 'Low', 'Yeast']`. The value accessed in the DataTable.js (when sorting) is still an array `"Fresh, Stale, Low Yeast"` and the search function didn't handle Arrays it would default to the else statement where `val = undefined;` thus it would revert to sorting by the IDs.

HenriRabalais:
Neat! That makes sense. Do you think that will mess with formatColumn functions that are already doing this manually before sending it to the Datatable?

maltheism:
@HenriRabalais I don't think it will mess with it from manually testing it. The sort function in the DataTable.js wasn't using the value from formatColumn. See here: https://github.com/aces/Loris/blob/27a76e487d5b74cb1bd58c2221275ba5ef60834b/jsx/DataTable.js#L178 the variable `val` is being used in the following lines of code for `Value`:
```
if (this.props.RowNameMap) {
        index.push({RowIdx: idx, Value: val, Content: this.props.RowNameMap[idx]});
      } else {
        index.push({RowIdx: idx, Value: val, Content: idx + 1});
      }
```
then later when a.Value or b.Value is being compared. Previously a.Value or b.Value was undefined.

Issue #6405, opened 2020-04-17
[document_repository] no files under root folder


kongtiaowang:
Describe the bug:
loading the main page, there is no file under root folder.
It should like https://demo.loris.ca/document_repository/.

zaliqarosli:
this is non-issue, its the intended design.Files belonging to directories/sub directories should not also be showing at the top level  https://github.com/aces/CCNA/issues/4302

PR #6408, opened 2020-04-17
[document_repository] fix no file under root folder 

kongtiaowang:
fix no file under root folder

Testing instructions (if applicable)
load the page, it will show all the files and all categories in the table.
compare with https://demo.loris.ca/document_repository/

Link(s) to related issue(s)
#6405


zaliqarosli:
hey @kongtiaowang @driusan this was a design decision that was made with @3design that was implemented in https://github.com/aces/Loris/pull/5572. this PR here reverses that. no files should be under the root folder is they are actually in the subfolders - its not a bug

PR #6418, opened 2020-04-20
[instrument_list] Add battery active check

lingma:
## Brief summary of changes

This PR add a test_battery active or not check.

Request in the PR #6365 by @nicolasbrossard

### DIFF
--- a/modules/instrument_list/php/instrument_list.class.inc
+++ b/modules/instrument_list/php/instrument_list.class.inc
@@ -349,7 +349,8 @@ class Instrument_List extends \NDB_Menu_Filter
        	        ON ((t.Test_name=b.Test_name OR b.Test_name IS NULL) 
       	        AND (s.SubprojectID=b.SubprojectID OR b.SubprojectID IS NULL) 
        	        AND (s.Visit_label=b.Visit_label OR b.Visit_label IS NULL) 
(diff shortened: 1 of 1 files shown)


driusan:
Why are you trying to add this? The test battery shouldn't affect what's hows up in the instrument list. Instruments added to the flag table manually with the fix_timepoint_date_problems script should still appear in the list of that candidate's instruments.

lingma:
I don't see the necessity, neither would this PR cause problem so I add the check.

johnsaigle:
@ridz1208 This seems like an enhancement and potentially shouldn't be included in the release (since it's frozen)

nicolasbrossard:
@driusan Maybe I am missing something but why are inactive instruments part of the list of that candidate's instruments? If they are, what's the point of inactivating an instrument in the test_battery table then?

lingma:
I believe the test_battery Active status will only apply to the subsequent instruments, not previous ones.

driusan:
The Active flag in the test battery is used to select which standard instruments get added to the candidate's instrument when the instrument is started. Instruments that are added to that candidate outside of the battery are still instruments that were administered.

driusan:
This introduces a bug where some of some candidate's instruments would be inaccessible but doesn't fix anything.

Issue #6287, opened 2020-04-06
[Configuration Module] Usability: error could be more specific and the field with the error is not highlighted.


racostas:
**Is your feature request related to an existing module?**
Configuration Module.

**Describe the solution you'd like**
When having a duplicate entry the error shown is always "ERROR: Duplicate Value In The Instrument Form". The specific field with the issue  is not highlighted making a bit harder to the user to find were the error is.

**Describe alternatives you've considered**
To clarify the error message according to the field were is was issued and possible highlight it for facilitating the user fix.

**Additional context**
<img width="869" alt="ConfigMod-error_message" src="https://user-images.githubusercontent.com/37309344/78615493-ac776500-783f-11ea-9dcd-733c69c53ed9.png">

johnsaigle:
I issued a PR to improve the error message but this is more of a feature enhancement than a bugfix for the release. I'd suggest keeping this issue open but removing the 23 label @ridz1208

xlecours:
Hi @racostas. Since https://github.com/aces/Loris/pull/6419 is merged, is this issue fixed?

PR #6419, opened 2020-04-20
[Configuration] Improve error message

johnsaigle:
## Brief summary of changes

In the Config module, this error message is thrown whenever a duplicate value is submitted for a config ID (read the code comment for more details).

Previously this error message was targeted toward a specific part of the config module when it is in fact a more general error message. 

1. Reproduce the bug in #6287. You should get the new error message.

* This was brought up by #6287 but does not fully resolve it. The ticket requests error highlighting for the field generating the error which entails more significant changes to the module.

### DIFF
--- a/modules/configuration/ajax/process.php
+++ b/modules/configuration/ajax/process.php
@@ -89,7 +89,7 @@ foreach ($_POST as $key => $value) {
                 continue;
             }
             if (isDuplicate($ConfigSettingsID, $value)) {
-                displayError(400, "Duplicate Value In The Instrument Form");
+                displayError(400, "Duplicate value submitted");
                 exit;
             }
             // Get all the IDs in ConfigSettings with the web_path data type.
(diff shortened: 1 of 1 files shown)


modules/configuration/ajax/process.php
```diff
@@ -89,7 +89,7 @@
                 continue;
             }
             if (isDuplicate($ConfigSettingsID, $value)) {
-                displayError(400, "Duplicate Value In The Instrument Form");
+                displayError(400, "Duplicate value submitted");
```
maltheism, line 92: Would there be benefit by showing the `$value` in the dissplayError?

johnsaigle: Probably. I'll do this

driusan:
@racostas can you review/test since you submitted the issue this says it's resolving?

PR #6423, opened 2020-04-22
[instrument_list] control whether to show a imaging_browser link & active Module support

lingma:
## Brief summary of changes

Control whether showing an imaging_browser link depending on the user permission & the status of the `imaging_browser` module.

1. Create a user without imaging_browser permission;
2. Login with the new user to verify whether the imaging_browser will be shown at the end of the instrument_list, which could be find if you click on a candidate profile -> timepoint -> visit;
3. The user without proper permission should not see the imaging_browser link;
4. Logout the new created user, login with one with the imaging_browser permission, check to make sure that the imaging_browser link is there;
5. Disable `imaging_browser` module, check again to see that the imaging_browser link is no longer displayed.

### DIFF
--- a/modules/instrument_list/php/instrument_list.class.inc
+++ b/modules/instrument_list/php/instrument_list.class.inc
@@ -134,6 +134,29 @@ class Instrument_List extends \NDB_Menu_Filter
             // the candidate exists before getting here.
             throw new \LorisException("Timepoint list requires a timepoint.");
         }
+
+        // Check imaging_browser permission, used to control imaging_browser link
+        // on the instrument_list main page
+        $loris = $request->getAttribute("loris");
+        $perm  = false;
+        if (!is_null($loris) && $loris->hasModule('imaging_browser')) {
+            $user = $request->getAttribute("user");
+            if ($this->candidate->getData('Entity_type') == 'Scanner') {
+                $perm = $user->hasPermission('imaging_browser_phantom_allsites')
+                || $user->hasCenterPermission(
+                    'imaging_browser_phantom_ownsite',
+                    $this->timePoint->getCenterID()
+                );
+            } elseif ($this->candidate->getData('Entity_type') == 'Human') {
+                $perm = $user->hasPermission('imaging_browser_view_allsites')
+                || $user->hasCenterPermission(
+                    'imaging_browser_view_site',
+                    $this->timePoint->getCenterID()
+                );
+            }
+        }
+        $this->tpl_data['imaging_browser_permission'] = $perm;
+
         return parent::process($request, $handler);
     }
 
--- a/modules/instrument_list/templates/menu_instrument_list.tpl
+++ b/modules/instrument_list/templates/menu_instrument_list.tpl
@@ -217,9 +217,11 @@
      <tr><td nowrap="nowrap">The battery has no registered instruments</td></tr>
 {/section}
 </table>
+{if $imaging_browser_permission}
   <div class="col-xs-12 row">
   </div>
   <div class="col-xs-12 row">
     <button class="btn btn-primary" onclick="location.href='{$baseurl}/imaging_browser/viewSession/?sessionID={$sessionID}'">View Imaging data</button>
   </div>
+{/if}
 </div>


modules/instrument_list/php/instrument_list.class.inc
```diff
@@ -134,6 +134,26 @@ class Instrument_List extends \NDB_Menu_Filter
             // the candidate exists before getting here.
             throw new \LorisException("Timepoint list requires a timepoint.");
         }
+
+        // Check imaging_browser permission, used to control imaging_browser link
+        // on the instrument_list main page
+        $perm = false;
```
driusan, line 140: This seems to be missing the hasModule check that the other PR had. You should be able to get access to the LorisInstance class from `$request->getAttribute("loris")` if you need a reference to call it..

lingma: Oh, that hasModule check is done in the BaseRouter, so I don't need to do it here, right?

driusan: The BaseRouter checked if the `instrument_list` page being accessed exists, I thought you wanted to check if the `imaging_browser` exists and is active before adding a link to it?

lingma: Right, it is a different module.

lingma: @driusan Done, thanks for the remind.

modules/instrument_list/php/instrument_list.class.inc
```diff
@@ -134,6 +134,29 @@ class Instrument_List extends \NDB_Menu_Filter
             // the candidate exists before getting here.
             throw new \LorisException("Timepoint list requires a timepoint.");
         }
+
+        // Check imaging_browser permission, used to control imaging_browser link
+        // on the instrument_list main page
+        $loris = $request->getAttribute("loris");
+        $perm  = false;
+        if (!is_null($loris) && $loris->hasModule('imaging_browser')) {
```
maltheism, line 142: Hey @lingma, this is the first time I'm seeing the `hasModule` used and from the `$request->getAttribute("loris");` object and I'm wondering if it would be better to use the following code:
```
$user =& User::singleton();
$user->hasPermission('imaging_browser');
```
I'm also unsure if it's secure to do permission checks on the backend from frontend objects? Maybe @johnsaigle or @driusan can comment. My apologies if I'm mistaken. Otherwise the code looks good.

lingma: Hi @maltheism the `hasModule` from the `$request->getAttribute("loris");` object is to check whether the module in question is active or not and the module file exists or not, which is not the same as check if the user has permission or not. This kind of check is new and only exists in the master branch.

maltheism: thanks @lingma for the explanation. I'll manually test the branch and give the approval soon.

lingma: @maltheism This is almost the same code as the PR #6365 which has been approved by @nicolasbrossard, the method in the PR #6365 is faster than the method here but the previous added some complexity.

nicolasbrossard: I believe there should also be a check in there that the user has the appropriate project for the button to be present.

nicolasbrossard: @driusan @maltheism I think we should also check whether the user belongs to the appropriate project before displaying the button. In fact, the _entire page_ should not be displayed if the user does not have the appropriate project (which is not the case right now).

lingma: The project verification should be a standard function from MenuFilter. If the project verification doesn't work, then the filter configuration is not properly done. I believe the project verification is not part of the purpose of this PR.

nicolasbrossard: I am talking about the instrument list page, which is not a `MenuFilter` anyway. Even if it was, I think that the project-based filtering is not performed in order to find out whether a particular page can be displayed/accessed or not: the `hasAccess()` method should do that I think.

You are right when you say that this might be outside this PR's mandate. Just wanted to raise the issue.

lingma: `instrument_list` inherits `NDB_Menu_Filter`, project level filter is done by inheritance as I remember, I suggest you to report this bug as an issue.

lingma: I have difficulty to test as my old installation doesn't work well, I have not tested the issue you reported, I think you can report as an issue.

lingma: I wrongly assumed `Menu_Filter` as `DataFrameworkMenu`, the project is not filtered, you can modify instrument_list to inherit `DataFrameworkMenu`, then add a `useProjectFilter`, which might disable `instrument_list`'s project filter function in the user interface. As a result, I believe that not to filter is on purpose by system design, there is not any need to add project control unless you want to modify a lot more. It is up to you and @driusan

lingma: reported in #6928

christinerogers: Hi guys please see PR #6836 and (issue: #6829) -- it's a known issue with a solution committed a month ago.
it's awaiting input from Dave/core team.

lingma: They are not the same, #6836 (#6829) talks about user permission, @nicolasbrossard talked about project filter, though I believe it is not necessary to add project filter on individual instrument.

ridz1208:
@nicolasbrossard what is the status of this ?

nicolasbrossard:
@ridz1208 I reviewed this PR and made my comments but there seems to be a lot of discussion going back and forth. I guess @ridz1208 or @driusan should decide what has to be done.

ridz1208:
@cmadjar do you have an opinion on this? I don't see an inherent problem with it but I dont wanna test/review unless its a desired feature (and the permission logic makes sens)

cmadjar:
@ridz1208 I don't have any strong opinion on this. It definitely makes sense to hide the imaging button if a user has not the permission to view the images. From the user's perspective, it might be less infuriating not seeing the button than seeing a button that leads you to a permission denied page.

cmadjar (approved):
The changes make sense and look good to me. I did not test it though.

PR #6426, opened 2020-04-23
[DQT] 'Download Table as CSV' broken if StaticDataTable has non-empty data from mri_data field

maltheism:
## Brief summary of changes

Here is a bugfix PR for an issue with the CSV download in the DQT module. The StaticDataTable.js has `data` containing `react link elements` from the DQT module and will result in the an error being thrown. I'm still investigating if a better solution exists by modifying the DQT instead of cleaning the data in the StaticDataTable.js and before it's parsed to a worker for CSV download.

0. If you don't already have mri data in the DQT, run `php tools/CouchDB_Import_MRI.php`
1. Go to 'Data Query Tool' module
2. Click on 'Define Fields' tab, select 'mri_data' in 'Instrument:' field, input 'Selected_dwi25' in 'Search within instrument:' field.
3. Select the 'Selected_dwi25' div that shows up below
4. Go to 'View Data' tab, click on 'Run Query'.
5. Once the data table populates, scroll down to the bottom of the page, and click on 'Download Table as CSV' button.
6. Notice the CSV now downloads (before PR there would be an error)

6425


maltheism:
Hi @zaliqarosli I'm a little confused because the CSV (I downloaded from my PR) matches the expected.

See here:

zaliqarosli:
hmm that's strange. i'm testing it again and I get the same result as last time. I'm testing on Chrome, and the result I got was for 'cross-sectional' data. the download csv button doesn't even work for 'longitudinal' data

maltheism:
@zaliqarosli thanks for testing it again. I'll try again myself and see what happens from following the instructions of the bug. I don't believe I tested the cross-sectional or longitudinal data at all. I just clicked Download Table as CSV after following the instructions from the bug.

zaliqarosli:
@maltheism yeah, i didn't realize the results were different for cross-sectional/longitudinal, i only just noticed that now! maybe i should update the issue

maltheism:
Hi again @zaliqarosli, today the test produced what your screenshots show. I'm a little worried because last week my CSV files produced what the screenshots from me show. I wonder if a PR was merged into 23.0-release that broke something. 😬

zaliqarosli:
@maltheism hmm i'm not sure! let me know what you find. goodluck!

maltheism:
Turns out the Download CSVi is broken not just in the DQT but throughout LORIS.

ridz1208:
@maltheism could you specify what's broken throughout LORIS ? I tested the download on a couple modules and it was fine...

maltheism:
@ridz1208 never mind, I just fixed the issue here and something confused me last night.

maltheism:
Thanks @zaliqarosli, ready for re-testing.

I thought the null check wouldn't be necessary (after testing on the development version of react) but on production it throws that error.

maltheism:
Thanks @zaliqarosli, I didn't check for the longitudinal table but it should now be working as well.

PR #6414, opened 2020-04-20
[Instruments] adding a TestPlan.md file

PapillonMcGill:
## Brief summary of changes

Added a testplan.md file for the instrument module

#### Testing instructions (if applicable)

1.  Check that the instructions are
- accurate
- complete
- without syntax or grammar errors

#### Link(s) to related issue(s) 

* Resolves #  6293

PR #6428, opened 2020-04-27
[instruments] Check for data_entry permissions

ridz1208:
## Brief summary of changes
Adds check for `data_entry` permission before allowing access to instruments

related to #6414

### DIFF
--- a/modules/instruments/README.md
+++ b/modules/instruments/README.md
@@ -22,7 +22,7 @@ It also does not manage surveys in any way.
 
 ## Permissions
 
-None. Individual instruments may provide their own permissions.
+Accessing instruments requires the `data_entry` permission. Individual instruments may provide their own permissions.
 
 ## Configurations
 
--- a/modules/instruments/README.md
+++ b/modules/instruments/README.md
@@ -22,7 +22,8 @@ It also does not manage surveys in any way.
 
 ## Permissions
 
-Accessing instruments requires the `data_entry` permission. Individual instruments may provide their own permissions.
+Accessing instruments requires the `data_entry` permission. Individual 
+instruments may require their own additonal access permissions.
 
 ## Configurations
 
--- a/modules/instruments/README.md
+++ b/modules/instruments/README.md
@@ -23,7 +23,7 @@ It also does not manage surveys in any way.
 ## Permissions
 
 Accessing instruments requires the `data_entry` permission. Individual 
-instruments may require their own additonal access permissions.
+instruments may require their own additional access permissions.
 
 ## Configurations
 
--- a/modules/instruments/test/TestPlan.md
+++ b/modules/instruments/test/TestPlan.md
@@ -1,7 +1,8 @@
 # Instrument Test Plan
 
 1. Verify that in order to access the module the user meets one of the following criteria:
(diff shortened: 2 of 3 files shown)


driusan:
Blocked by #6292 6414

driusan:
@ridz1208 This is no longer blocked

modules/instruments/test/TestPlan.md
```diff
@@ -1,37 +1,37 @@
...
+     candidate_list module (Access Profile) 
      - go to the candidate_list
      - click on a PSCID to get to the timepoint_list module
      - click on a timepoint to get to the instrument_list module
      - click on an instrument to access the instrument itself.  
      _**Note:** All steps before clicking on the instrument itself are handled 
      by their own modules and necessary permissions to access these modules should be 
      part of their respective testplans._
-   - User is provided a direct link to the instrument.
-2. Select and instrument in the list of the session instruments.
```
ridz1208, line 13: @driusan item 2 was deleted because it's already one of the steps on item 1

ridz1208:
@driusan ready for review

ridz1208:
@CamilleBeau why don't you give this a test

modules/instruments/README.md
```diff
@@ -22,7 +22,7 @@ It also does not manage surveys in any way.
 
 ## Permissions
 
-None. Individual instruments may provide their own permissions.
+Accessing instruments requires the `data_entry` permission. Individual instruments may provide their own permissions.
```
driusan, line 25: Does "provide their own permissions" mean it overrides the data_entry or augments it? It looks like it's "permissions they require in addition to" from the code below but that should be clearly in the documentation as well.

ridz1208: I'm not sure about the interaction between this hasAccess call and the one in here https://github.com/aces/Loris/blob/826b0a3455eccad59839aa916ca78f4667bdbca5/php/libraries/NDB_BVL_Instrument.class.inc#L318

I would guess to get access to an instrument you would need both to return true ? so it would be augments it.

if you can confirm that I'll try to reformulate

driusan: It looks like the code you're changing is adding it to the hasAccess of the module class, so I'm pretty sure it augments it and the instrument classes themselves can't override it the way it's written.

But I'm not sure if that's your intention or not based on the documentation change's ambiguity.

ridz1208: my intention was to make data_entry ALWAYS required and whatever projects want to add is their call. so I think it works the way it is.

I'll try to change the wording a bit

driusan: Yeah, that's the way this code looks like it works. The module class is going to return a permission denied before it even tries to load the instrument class.

modules/instruments/README.md
```diff
@@ -22,7 +22,8 @@ It also does not manage surveys in any way.
 
 ## Permissions
 
-None. Individual instruments may provide their own permissions.
+Accessing instruments requires the `data_entry` permission. Individual 
+instruments may require their own additonal access permissions.
```
driusan, line 26:
```suggestion
instruments may require their own additional access permissions.
```

ridz1208:
@driusan changes done

PR #6429, opened 2020-04-27
[Candidate_list] integration test - complete skip test

kongtiaowang:
## Brief summary of changes
[Candidate_list] integration test - complete all skip tests, test react select dropdown menu by value not by order.
Travis

### DIFF
--- a/modules/candidate_list/test/candidate_listTest.php
+++ b/modules/candidate_list/test/candidate_listTest.php
@@ -2,7 +2,7 @@
 /**
  * Candidate_list automated integration tests
  *
- * PHP Version 5
+ * PHP Version 7
  *
  * @category Test
  * @package  Loris
@@ -11,6 +11,7 @@
  * @link     https://github.com/aces/Loris
  */
 use Facebook\WebDriver\WebDriverBy;
+use Facebook\WebDriver\WebDriverSelect;
 require_once __DIR__
     . "/../../../test/integrationtests/LorisIntegrationTestWithCandidate.class.inc";
 /**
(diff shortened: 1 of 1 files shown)


modules/candidate_list/test/candidate_listTest.php
```diff
@@ -244,22 +315,21 @@ function testDataEntryAndOpenProfile()
...
         $this->safeClick(WebDriverBy::cssSelector($btn));
-
         // input PSCID and DCCID
         $dccid = "#lorisworkspace > div > div:nth-child(1) > div >".
                  " div:nth-child(2)>form>div>div:nth-child(1)>div>div>input";
         $pscid = "#lorisworkspace > div > div:nth-child(1) > div >".
                  " div:nth-child(2)>form>div>div:nth-child(2)>div>div>input";
         // to do react input value
-        $this->webDriver->executeScript(
+        $this->webDriver->executescript(
```
lingma, line 324: I just don't understand why we are modifying `executeScript` to `executescript` in most occurrences. I don't see any advantage but I am OK to accept, I suggest revert this kind of modifications.

modules/candidate_list/test/candidate_listTest.php
```diff
@@ -244,22 +315,21 @@ function testDataEntryAndOpenProfile()
...
             "input = document.querySelector('$dccid');
                  lastValue = input.value;
                  input.value = '300001';
                  event = new Event('input', { bubbles: true });
                  input._valueTracker.setValue(lastValue);
                  input.dispatchEvent(event);"
         );
-        $this->webDriver->executeScript(
+        $this->webDriver->executescript(
```
lingma, line 332: The same as above

modules/candidate_list/test/candidate_listTest.php
kongtiaowang: @lingma "executeScript" and "executescript" are the same in test cases. All of the other test cases use the "executescript", only these two are uppercase. That is why I changed it.

lingma: I checked, currently not all is in uppercase, I don't see why we prefer a less advantaged solution.

PR #6444, opened 2020-04-27
[raisinbread] modification after the PR #6395

lingma:
## Brief summary of changes

Modification after the merge of the PR #6395

1. Check instrument medical_history page 2 display, especially the table header.

### DIFF
--- a/raisinbread/instruments/NDB_BVL_Instrument_medical_history.class.inc
+++ b/raisinbread/instruments/NDB_BVL_Instrument_medical_history.class.inc
@@ -280,27 +280,9 @@ class NDB_BVL_Instrument_medical_history extends NDB_BVL_Instrument
         );
 
         // Concussion table headers
-
-        $group[] = $this->form->createElement(
-            "static",
-            null,
-            null,
-            "Incident"
-        );
-
-        $group[] = $this->form->createElement(
-            "static",
-            null,
-            null,
-            "Hospitalized?"
-        );
-
-        $group[] = $this->form->createElement(
-            "static",
-            null,
-            null,
-            "Age"
-        );
+        $group[] = $this->createLabel('Incident');
+        $group[] = $this->createLabel('Hospitalized?');
+        $group[] = $this->createLabel('Age');
 
         $this->form->addGroup(
             $group,
--- a/php/libraries/NDB_Page.class.inc
+++ b/php/libraries/NDB_Page.class.inc
@@ -494,21 +494,14 @@ class NDB_Page implements RequestHandlerInterface
     /**
      * Create a QuickForm label element but does not add it to the form
      *
-     * FIXME: LorisForm::createElement does not use the label parameter when
-     *        the element type is "static". Also, the $labelString paramter here
-     *        has been placed as the fourth parameter which is actually the $options
-     *        array in the createElement function! - Alexandra Livadas
-     *
      * @param string $labelString The label to attach to the element
      *
      * @return array representing label element
      */
-    function createLabel(
-        $labelString
-    ) {
+    function createLabel($labelString)
+    {
         return $this->form->createElement(
             "static",
-            null,
             null,
             $labelString
         );
@@ -584,8 +577,6 @@ class NDB_Page implements RequestHandlerInterface
 
     /**
      * Creates an HTML checkbox but does not add it to the form.
-     *
-     * FIXME: $closer is never used or explained. - Alexandra Livadas
      *
      * @param string $field   The fieldname for this checkbox
      * @param string $label   The label to attach to this checkbox


driusan:
I think we should revert #6395. If we know this is going to break (at least some) instruments, and is going to add a lot of overhead to the upgrade. If there a strong justification for keeping it considering the work it's going to add other than that it looked like it was wrong?

lingma:
I have stated in the #6395 that there will be an impact, but if we don't do this, sooner or later there will be someone asking the same questions concerning their confusion.

driusan:
I sent #6451 to revert the change instead of this, I think it's safer at this point in the release cycle

driusan:
added blocked because I don't think this should go in until after any potential changes from  #6457

lingma:
Of course this should be after the #6457

ridz1208:
@driusan 

By modifying the create label function in this we might be mitigating a lot of issues in project instruments, I think most should be using the createLabel function.

but to be even less disruptive to projects. I would propose to add some logic to the `createElement` function's static case to dynamically use the options or the label depending on which one seems to contain the actual label string. I would also add an error_log statement to warn of the upcoming deprecation of the options-to-label conversion and recommend to upgrade to the create label function instead... That's if we insist on repairing this... bug

raisinbread/instruments/NDB_BVL_Instrument_medical_history.class.inc
```diff
@@ -284,21 +284,18 @@ class NDB_BVL_Instrument_medical_history extends NDB_BVL_Instrument
         $group[] = $this->form->createElement(
```
ridz1208, line 284: I think these calls should be converted to `createLabel()` instead.

lingma:
> create lab

`createLabel` function modification is optional, if #6457 is accepted, but it is preferable to use the modified `createLabel`. If you read #6457 then you can see that the modification here to `createLabel` doesn't cause any impact to the existing projects. Of course, prudence is better

raisinbread/instruments/NDB_BVL_Instrument_medical_history.class.inc
lingma: Yes, createLabel() is better, but I just don't want to touch too much the existing code.

lingma: There are lots of call without createLabel in the existing projects, and createLabel actually calls the createElement, so actually there is no big difference.

ridz1208: @lingma i understand your concern but if these projects were using createLabel in the first place, then they would not be affected by correcting the createElement code because we would have just fixed createLabel accordingly and projects would have never know about it.

I think it's fair to give projects a full release time to "amend" their instruments and we would then deprecate the old way in loris 24

lingma: If you read #6457, you can see that the existing projects will have ZERO impact with the PR.

lingma:
> That's what #6457 seems to be doing. so basically I would add the error_log warning to #6457, on top of what is done here and recommend the conversion to createLabel() instead

#6457 will make Loris support both `createElement('static', null, null, 'label')` and `createElement('static', null, 'label')`, the existing implementation of `createLabel` and the proposed `createLabel` will both work without any difference. The error_log warning is possible but not necessary in #6457 since there is no difference in execution, Choosing one format over the other is only a question of support to the old format, by preference I suggest the new format. Unless we plan to abandon the `createElement('static', null, null, 'label')`, I don't think the warning is necessary ant it will be cumbersome to the existing projects.

`createElement('static', null, null, 'label')` vs `createLabel` is just a question of preference, conversion to `createLabel` in some context will look better, but not in other contexts such as with subsequent `createElement('???'` other than `createElement('static'`

ridz1208:
@driusan this one is in your hands now

driusan:
@ridz1208 the labels on GitHub still say you requested changes, how is this in my hands? Were they resolved and the review just wasn't updated?

lingma:
I can change the calling method from `createElement('static'` to `createLabel`, as I explained, I did not want to modify too much the original code, and I don't feel the modification is necessary. But it is possible to modify. @driusan

ridz1208:
@driusan my requested changes are not necessary. I don't care enough to fight for them

lingma:
Just a sample raisinbread instrument, whatever the method for the same function, no big deal

driusan:
I agree with Rida that createLabel would be a superior solution for maintenance and our sample forms shouldn't be using $this->form because it might give people the mistaken impression that it's a good practice.. since you offered to do it 4 hours ago, would it be a quick change @lingma? I also don't feel that strongly about it and can merge it as is, but if you're offering to do it and have the time we might as well fix it properly.

lingma:
Done @driusan

lingma:
Wait

driusan:
@ridz1208 can you rereview? I think your suggested changes are done and it looks good to me, but you're more familiar with those helpers than I am to know if there's any subtle issues to look for...

ridz1208:
@driusan all good

lingma:
I noticed that the instrument uses `createText` instead of `createElement` while the later is common in other instruments I have seen. It is better to use `createLabel` in this context, @ridz1208 and @driusan have reason for the modification.

Issue #6314, opened 2020-04-06
[Imaging_uploader] Uploading file bigger than 1024MB doesn't throw error swal


zaliqarosli:
**Describe the bug**
Test plan item 7: 'A file that exceeds the max upload size' fails for both 'Phantom Scans' Yes and No, giving different errors

Phantom Scans No:
<img width="1194" alt="Screenshot 2020-04-06 15 49 52" src="https://user-images.githubusercontent.com/17415878/78599870-e20c5600-781f-11ea-8121-90379b149204.png">

Phantom Scans Yes:
<img width="1193" alt="Screenshot 2020-04-06 15 51 52" src="https://user-images.githubusercontent.com/17415878/78599892-eb95be00-781f-11ea-9f5e-b3950b695a5b.png">

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to 'Imaging Uploader' module
2. Click on 'Upload' tab
3. Fill in the form: 'Phantom Scans': Yes/No
4. Upload a file into 'File to Upload' field that is larger than 1024MB
5. Click 'Submit' 
6. See console log errors as in the screenshot above. No swal pops up where it should.

**What did you expect to happen?**
A swal with 'Please make sure files are not larger than 1024M' warning should pop up and progress bar should not start

johnsaigle:
I wasn't able to reproduce this.

Steps on my VM:
1. `git fetch aces`, `git checkout aces/23.0-release`, `make dev`
2. Created a file `MTL007_300007_V1.tar.gz` that is 100 MB.
3. My VM server upload limit is set to 2M(B).
4. Using both Phantom scans yes and no, the swal appeared saying the file was too big.

Where is the value `1024MB` coming from? That's a GB which is pretty large for a front-end uploader anyway. I didn't see this value in `modules/imaging_uploader/test/TestPlan.md`.

#### My gzipped tar file

<img width="374" alt="Screen Shot 2020-04-08 at 12 21 41" src="https://user-images.githubusercontent.com/4022790/78809131-71d31100-7994-11ea-878a-bb59499fc7fa.png">

#### Phantom scans yes

<img width="875" alt="Screen Shot 2020-04-08 at 12 22 14" src="https://user-images.githubusercontent.com/4022790/78809153-77305b80-7994-11ea-8275-c6354ff3bfe2.png">


#### Phantom scans no

[...]

Issue #5046, opened 2019-08-20
Do front-end validation of file sizes (and reject when too big)


johnsaigle:
When a user attempts to upload a file greater than a certain size, a sweet alert should be generated informing them that the file is too big and will be rejected.

johnsaigle:
List of file uploaders:

 Instrument Manager
 Media
 Document Repository
 Genomic Browser
 Imaging Uploader
 Data Release
 Instrument Builder
 Biobank (when merged)
Publications

An example of how to do this can be found in this PR: https://github.com/aces/Loris/pull/5224

PR #6447, opened 2020-04-27
[Imaging_uploader] Clear front-end console error logs

zaliqarosli:
## Brief summary of changes

This PR fixes the errors shown in https://github.com/aces/Loris/issues/6314. It also takes into account a 413 error, which should render the same swal error as 'mriFile' 400 error.

1. Go to 'Imaging Uploader' module
2. Click on 'Upload' tab
3. Fill in the form: 'Phantom Scans': Yes/No
4. Upload a file into 'File to Upload' field that is larger than the number stated in the 'Notes' field 'File cannot exceed ...'
5. Click 'Submit'
6. Swal will pop up with "Please make sure files are not larger than .."
7. See console log, which should only have the error `Failed to load resource: the server responded with a status of 400 (Bad Request)`
8. If you can, try to trigger a 413 (Request Entity Too Large) error. The swal should remain the same as above.

* Resolves https://github.com/aces/Loris/issues/6314

### DIFF
--- a/modules/imaging_uploader/jsx/UploadForm.js
+++ b/modules/imaging_uploader/jsx/UploadForm.js
@@ -273,7 +273,7 @@ class UploadForm extends Component {
         let messageToPrint = '';
         if (error.responseJSON.errors) {
           errorMessage = error.responseJSON.errors;
-        } else if (error.status == '413') {
+        } else if (error.status == 413) {
           errorMessage = {
             'mriFile': ['Please make sure files are not larger than ' + this.props.maxUploadSize],
           };
--- a/modules/imaging_uploader/jsx/UploadForm.js
+++ b/modules/imaging_uploader/jsx/UploadForm.js
@@ -63,7 +63,7 @@ class UploadForm extends Component {
     formData[field] = value;
 
     if (field === 'mriFile') {
-      if (value.name !== '' && formData.IsPhantom === 'N') {
+      if (value.name && value.name !== '' && formData.IsPhantom === 'N') {
         let patientName = value.name.replace(/\.[a-z]+\.?[a-z]+?$/i, '');
         let ids = patientName.split('_', 3);
         formData.candID = ids[1];
@@ -269,9 +269,11 @@ class UploadForm extends Component {
       // - Returns to Upload tab
       error: (error, textStatus, errorThrown) => {
         let errorMessage = Object.assign({}, this.state.errorMessage);
-        let hasError = this.state.hasError;
+        let hasError = Object.assign({}, this.state.hasError);
         let messageToPrint = '';
-        errorMessage = (error.responseJSON || {}).errors || 'Submission error!';
+        if (error.responseJSON.errors) {
+          errorMessage = error.responseJSON.errors;
+        }
         for (let i in errorMessage) {
           if (errorMessage.hasOwnProperty(i)) {
             errorMessage[i] = errorMessage[i].toString();
--- a/modules/imaging_uploader/jsx/UploadForm.js
+++ b/modules/imaging_uploader/jsx/UploadForm.js
@@ -275,7 +275,7 @@ class UploadForm extends Component {
           errorMessage = error.responseJSON.errors;
         } else if (error.status == '413') {
           errorMessage = {
-            'mriFile': ["Please make sure files are not larger than " + this.props.maxUploadSize]
+            'mriFile': ['Please make sure files are not larger than ' + this.props.maxUploadSize],
           };
         }
         for (let i in errorMessage) {
--- a/modules/imaging_uploader/jsx/UploadForm.js
+++ b/modules/imaging_uploader/jsx/UploadForm.js
@@ -63,7 +63,7 @@ class UploadForm extends Component {
     formData[field] = value;
 
     if (field === 'mriFile') {
-      if (value.name && value.name !== '' && formData.IsPhantom === 'N') {
+      if (value.name && formData.IsPhantom === 'N') {
         let patientName = value.name.replace(/\.[a-z]+\.?[a-z]+?$/i, '');
         let ids = patientName.split('_', 3);
         formData.candID = ids[1];
--- a/modules/imaging_uploader/jsx/UploadForm.js
+++ b/modules/imaging_uploader/jsx/UploadForm.js
@@ -273,6 +273,10 @@ class UploadForm extends Component {
         let messageToPrint = '';
         if (error.responseJSON.errors) {
           errorMessage = error.responseJSON.errors;
+        } else if (error.status == '413') {
+          errorMessage = {
+            'mriFile': ["Please make sure files are not larger than " + this.props.maxUploadSize]
+          };
         }
         for (let i in errorMessage) {
           if (errorMessage.hasOwnProperty(i)) {
--- a/modules/imaging_uploader/jsx/UploadForm.js
+++ b/modules/imaging_uploader/jsx/UploadForm.js
@@ -271,7 +271,7 @@ class UploadForm extends Component {
         let errorMessage = Object.assign({}, this.state.errorMessage);
         let hasError = Object.assign({}, this.state.hasError);
         let messageToPrint = '';
-        if (error.responseJSON.errors) {
+        if (error.responseJSON && error.responseJSON.errors) {
           errorMessage = error.responseJSON.errors;
         } else if (error.status == 413) {
           errorMessage = {


zaliqarosli:
@HenriRabalais yeah, currently the validation is happening on the php side. in order to validate before starting the upload, it would have to be done on the front-end. perhaps you can create a ticket for this if you want to keep track of the request!

HenriRabalais:
Already done by @johnsaigle! #5046

modules/imaging_uploader/jsx/UploadForm.js
```diff
@@ -63,7 +63,7 @@ class UploadForm extends Component {
     formData[field] = value;
 
     if (field === 'mriFile') {
-      if (value.name !== '' && formData.IsPhantom === 'N') {
+      if (value.name && value.name !== '' && formData.IsPhantom === 'N') {
```
maltheism, line 66: Not sure why this line changed.
The `value.name `check will return false if value.name is an empty string. So `value.name !== ''` would cover that check. Also if the submit button is pressed when "File to upload" has not been done, the browser will request the user to Select a file. The check on `value.name` will return false if it's an empty string. I'm also unsure if either the `value.name` or `value.name !== ''` are needed at all because of the browser check when the `File to Upload` field is required.

zaliqarosli: i added `value.name` to the condition to get rid of an undefined variable error that popped up in the error log. i can't seem to find the screen shot to show you though

maltheism: @zaliqarosli I haven't been able to replicate that error case. I sort of think the two checks are unnecessary but maybe they somehow can be thrown. I just thought to comment on it. It doesn't really matter.

nicolasbrossard:
@zaliqa I cannot reproduce the HTTP 413 that you get on the LORIS test VM. When we talked on slack, you told me you were able to reproduce this. The screenshot you included showed that you were getting an exception in the JS console log: `Uncaught type error: Cannot assign to read only property '0' of string 'Submission error!` (on line 271 of `UploadForm.js`). Since your PR does not modify that line, I believe you will still get the same uncaught error and the behaviour will not change. Am I missing something?

modules/imaging_uploader/jsx/UploadForm.js
HenriRabalais: By looking at the code, my best guess is that the error was thrown when `value.name` was `null` but then ` value.name.replace()` was called on an undefined variable. This would probably happen when `componentDidMount` is called since it called onFormChange like this: `this.onFormChange(this.state.form.IsPhantom.name, null);`

Not sure if the best solution is to add this if statement, or to address the fact it's passing `null` as the value.

But I agree with @maltheism, if there's the `if (value.name)` check then the `value.name !== ''` becomes redundant.

zaliqarosli: if i remove `value.name !== ''`, would either of you be willing to do another test in case that breaks something? @HenriRabalais @maltheism

zaliqarosli:
hi @nicolasbrossard i do change that line (now on line 274) https://github.com/aces/Loris/pull/6447/files#diff-dcebbc7ee7bb967d320adfac9a64c00fL274

modules/imaging_uploader/jsx/UploadForm.js
maltheism: @zaliqarosli I can test again.

zaliqarosli, line 66:
```suggestion
      if (value.name &&  && formData.IsPhantom === 'N') {
```

zaliqarosli, line 66:
```suggestion
      if (value.name && formData.IsPhantom === 'N') {
```

zaliqarosli:
@maltheism ready for testing again. thank you!

nicolasbrossard:
@zaliqarosli The first line of your error handling routine is still `let errorMessage = Object.assign({}, this.state.errorMessage);` and this line blows up according to the JS console screenshot you showed me on slack when you get a 413 error. So whatever code lies beyond that line will never be executed anyway. Correct?

zaliqarosli:
@nicolasbrossard no, the error `Uncaught type error: Cannot assign to read only property '0' of string 'Submission error!` is referring to the 'Submission error!' string you can see on line 274 which has been removed. line 271 doesn't throw an error

ridz1208:
@zaliqarosli and @nicolasbrossard can you re-add the pass manual testing when it's ready

maltheism:
Hi @zaliqarosli, my first round of testing was a bust because I'm encountering what @nicolasbrossard is commenting.

The browser console displays the following when console.log(error) inside the error function:
```
Object

abort: function(a)
always: function()

complete: function()

done: function()

error: function()

fail: function()

getAllResponseHeaders: function()

getResponseHeader: function(a)

overrideMimeType: function(a)

pipe: function()

progress: function()

promise: function(a)

readyState: 4

responseText: "<br />↵<font size='1'><table class='xdebug-error xe-warning' dir='ltr' border='1' cellspacing='0' cellpadding='1'>↵<tr><th align='left…"

setRequestHeader: function(a,b)

state: function()

status: 400

statusCode: function(a)

statusText: "Bad Request"

success: function()

then: function()
```

So the error.status happens to be 400.

zaliqarosli:
@maltheism yeah, thats fine, you should be getting a 400. i also get a 400 in my dev VM. the 413 case is only appearing on the test VM and we're not sure how to replicate it

nicolasbrossard:
@maltheism Actually Alizee, if you could try your test on the test VM and see if you get a 400 or a 413 error, that would help. I get a 400 while Zaliqa get a 413 (weird...)

maltheism:
@zaliqarosli I think to Could not upload file error from Max file size is supposed to be thrown by the php/libraries/File_Upload.class.inc and it's not handling it correctly.

The _saveFile($values) function gets called in the imaging_uploader/php/imaging_uploader.class.inc and the following code informs where the file is located for the File_Upload.class.inc.
```
 //Tell File_Upload what file handlers to use.
        $file->setFileHandler(
            "mriFile",
            $this
        );
```
Then after that code the `$file->processFiles();` function is called. Well inside php/libraries/File_Upload.class.inc you can see there is this code:
```
if (!$this->preProcessFile()) {
                $this->setError(
                    $field,
                    "Could not upload file. Max file size is "
                    . ini_get("upload_max_filesize") . "."
                );
                continue;
            }
```
Well the forgoing code doesn't have the `this->preProcessFile()` return false and so the setError never runs. I've modified the code in the `preProcessFile()` function to be:
```
function preProcessFile()
    {
        if (!$this->isCLImport) {
            $this->file = $this->form->getElement($this->fieldName);
            if (!method_exists($this->file, "isUploadedFile")) {
                return false;
            }
            if ($this->file->isUploadedFile()) {
                $this->fileInfo = $this->file->getValue();
                if ($this->fileInfo['error'] === 1) {
                    return false;
                }
                return true;
            } else {
                return false;
            }
        } else {
            $this->file     =$this->CLFiles[$this->fieldName];
            $this->fileInfo =$this->file['info'];
            return true;
        }
    }
```
So it will return false when the `$this->fileInfo['error'] === 1` check returns true. Although even with that modification the File_Upload.class.inc doesn't seem to stop because it continues.

zaliqarosli:
@maltheism what's the issue you're having right now on your set up? I don't feel like im on the same page

maltheism:
@zaliqarosli I'm going to test on my VM because maybe your change will satisfy it when the VM is somehow slightly different than my configuration.

I'm not sure why @nicolasbrossard is getting the error response code of 400 but I am because the File_Upload.class.inc is polluting the error_log in my apache and making my response throw the error function on the frontend instead of the success function. I've modified the File_Upload.class.inc to not pollute the error_log and that actually makes the frontend success function get called. So the whole thing is a mess to fix. It's also being done with ajax and when using the class Imaging_Uploader extends \NDB_Menu_Filter_Form. So I'm with little information on how to fix this problem for me besides doing a major rewrite. Anyway I'm going to test this on my VM.

maltheism:
I'm getting the correct status of 413 on my VM. @nicolasbrossard 
zal

nicolasbrossard:
@maltheism Ok. Can you also test on the test VM? Looks like I am the only one who gets a 400...

modules/imaging_uploader/jsx/UploadForm.js
```diff
@@ -269,9 +269,15 @@ class UploadForm extends Component {
       // - Returns to Upload tab
       error: (error, textStatus, errorThrown) => {
         let errorMessage = Object.assign({}, this.state.errorMessage);
-        let hasError = this.state.hasError;
+        let hasError = Object.assign({}, this.state.hasError);
         let messageToPrint = '';
-        errorMessage = (error.responseJSON || {}).errors || 'Submission error!';
+        if (error.responseJSON.errors) {
+          errorMessage = error.responseJSON.errors;
+        } else if (error.status == '413') {
```
zaliqarosli, line 276:
```suggestion
        } else if (error.status == 413) {
```

modules/imaging_uploader/jsx/UploadForm.js
```diff
@@ -269,9 +269,15 @@ class UploadForm extends Component {
       // - Returns to Upload tab
       error: (error, textStatus, errorThrown) => {
         let errorMessage = Object.assign({}, this.state.errorMessage);
-        let hasError = this.state.hasError;
+        let hasError = Object.assign({}, this.state.hasError);
         let messageToPrint = '';
-        errorMessage = (error.responseJSON || {}).errors || 'Submission error!';
+        if (error.responseJSON.errors) {
```
zaliqarosli, line 274:
```suggestion
        if (error.responseJSON && error.responseJSON.errors) {
```

maltheism:
Everything works with the latest commit! @zaliqarosli @nicolasbrossard 

zal4

zaliqarosli:
thank you @maltheism !!!

nicolasbrossard:
LGTM. This PR now handles HTTP 400 and 413 error codes. I believe that there are other errors (e.g. server goes down right before the file is uploaded) that are not handled properly by this PR. This will be the object of another fix/PR. Approving the PR and opening another issue.

PR #6451, opened 2020-04-28
Revert "LorisForm static element bug in group mode (#6395)"

driusan:
This reverts commit 28528f93cdb4c6733cd36f345f2bed5f604dfd05.

Because of how pervasive LorisForm is, PR #6395 (which is intended as a cleanup) is too likely to cause problems in project instruments, so should not be done on the 23.0-release branch.


driusan:
The "Restart" button no longer appears in Travis even while logged in, so I need to close and restart to fix a transient failure with..

driusan:
aand that only restarted the properly run build, not the one with the problem..

lingma:
No need to revert, I will provide with a better solution very soon.

lingma:
A new PR #6457 will fix the impact issue, which I ever considered not necessary to take into account because we have to check a lot method signatures in upgrading.

driusan:
Should be fixed by #6457 instead

PR #6452, opened 2020-04-28
[Login] Login control

lingma:
## Brief summary of changes

­login is accessible after user login, the current PR will redirect user to the dashboard if that happens.

1. Login then check /login link

@maltheism reported the behavior

### DIFF
--- a/modules/login/php/module.class.inc
+++ b/modules/login/php/module.class.inc
@@ -51,7 +51,8 @@ class Module extends \Module
     }
 
     /**
-     * Check whether a user is logined or not, if yes, redirect to the base URL
+     * Check whether a user is currently logged in or not. If logged in, redirect to
+     * the base URL.
      *
      * @param ServerRequestInterface $request The incoming PSR7 request
      *
@@ -59,7 +60,7 @@ class Module extends \Module
      */
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
-        $user = $request->getAttribute("user");
+        $user = $request->getAttribute('user');
         if (!($user instanceof \LORIS\AnonymousUser)) {
             return (new \LORIS\Http\Response())
                 ->withStatus(307)
--- a/modules/login/php/module.class.inc
+++ b/modules/login/php/module.class.inc
@@ -13,6 +13,8 @@
  * @link       https://www.github.com/aces/Loris/
  */
 namespace LORIS\login;
+use \Psr\Http\Message\ServerRequestInterface;
+use \Psr\Http\Message\ResponseInterface;
 
 /**
  * Class module implements the basic LORIS module functionality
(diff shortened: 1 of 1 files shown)


lingma:
OK, I will check, I don't know how to redirect in Loris, so I just show the dashboard to simplify the issue.

driusan:
In the handle function (in this case of the module function), include a "Location" header and appropriate response code in the PSR7 response returned.

lingma:
@driusan This PR is necessary for this release. Because currently if you visit /login, you will ALWAYS see a login page without end if you don't click somewhere else.

modules/login/php/module.class.inc
```diff
@@ -47,4 +49,22 @@ class Module extends \Module
...
+    /**
+     * Check whether a user is logined or not, if yes, redirect to the base URL
+     *
+     * @param ServerRequestInterface $request The incoming PSR7 request
+     *
+     * @return ResponseInterface the outgoing PSR7 response
+     */
+    public function handle(ServerRequestInterface $request) : ResponseInterface
+    {
+        $user = \User::singleton();
```
driusan, line 62: The base router includes a "user" attribute on the request which is a (slightly) better place to get the user than a singleton.

modules/login/php/module.class.inc
```diff
@@ -47,4 +49,22 @@ class Module extends \Module
...
+        if (!($user instanceof \LORIS\AnonymousUser)) {
+            return (new \LORIS\Http\Response())
+                ->withStatus(307)
+                ->withHeader("Location", '/');
```
driusan:
```suggestion
                ->withHeader("Location", \NDB_Factory::singleton()->settings()->getBaseURL());
```

This (untested) will work whether or not LORIS is hosted under the base of the domain.

modules/login/php/module.class.inc
```diff
@@ -47,4 +49,22 @@ class Module extends \Module
...
+     *
+     * @return ResponseInterface the outgoing PSR7 response
+     */
+    public function handle(ServerRequestInterface $request) : ResponseInterface
+    {
+        $user = \User::singleton();
+        if (!($user instanceof \LORIS\AnonymousUser)) {
+            return (new \LORIS\Http\Response())
+                ->withStatus(307)
+                ->withHeader('Location', '/');
```
driusan, line 66:
```suggestion
                ->withHeader('Location', \NDB_Factory::singleton()->settings()->getBaseURL());
```

This (untested) will work regardless of whether LORIS is hosted under the base of the domain name.

driusan:
@maltheism can you retest since it's been rewritten since you added the tag?

modules/login/php/module.class.inc
```diff
@@ -47,4 +49,25 @@ class Module extends \Module
     {
         return "Login";
     }
+
+    /**
+     * Check whether a user is logined or not, if yes, redirect to the base URL
```
driusan, line 54:
```suggestion
     * Check whether a user is logged in or not, if yes, redirect to the base URL
```

PR #6457, opened 2020-04-29
[LorisForm] keep old group mode text parameter mode

lingma:
## Brief summary of changes

The PR #6395 is radical, so the existing instruments might have table title text display problem, this PR will keep the old format and the new correct format available at the same time

1. Test instrument with table to check whether table title is displayed correctly.
For example: medical_history from raisonbread, check PR #6444

PR #6395, PR #6451

### DIFF
--- a/php/libraries/LorisForm.class.inc
+++ b/php/libraries/LorisForm.class.inc
@@ -1960,7 +1960,13 @@ class LorisForm
         case 'submit':
             return $this->createSubmit($elname, $label, $attribs);
         case 'static':
-            $el         = $this->createBase($elname, $label, $attribs);
+            /* the $options was passed to the 2nd parameter pre-v23.0 for some
+             reason, it is important not to impact the existing instruments. */
+            $el         = $this->createBase(
+                $elname,
+                is_string($options) ? $options : $label,
+                $attribs
+            );
             $el['type'] = 'static';
             break;
         case 'advcheckbox':


php/libraries/LorisForm.class.inc
```diff
@@ -1960,7 +1960,13 @@ class LorisForm
         case 'submit':
             return $this->createSubmit($elname, $label, $attribs);
         case 'static':
-            $el         = $this->createBase($elname, $label, $attribs);
+            /* the $options was passed to the 2nd parameter pre-v23.0 for some
+               reason, it is important not to impact the existing instruments. */
+            $el         = $this->createBase(
```
maltheism, line 1965: Should the @param string be modified to string|array? 
found here: https://github.com/aces/Loris/blob/74e49c445fb9f415c1796ef5ca6a7e4cf9bb48de/php/libraries/LorisForm.class.inc#L119 be modified to accept

maltheism: also @lingma I might be mistaken but I thought the label is a string and options is an array?

lingma: Oh, true, but the label is generally a string, an option here is only to accommodate the previous version compatibility. So I suppose we can just let it be for now. Otherwise the things will become ugly.

maltheism: @lingma the approach here might be necessary if errors are encountered from what @ridz1208 suggested. https://github.com/aces/Loris/pull/6395#issuecomment-620059344

lingma: I feel it is not necessary to add $label as string|array, because the `$label` passed to the createBase of line 69 is always a string, the `$options` could be passed to the label (second parameter), but only as string, and which is for back compatibility purpose.

maltheism: @lingma should the logic still be changed to `empty($options) ? $label : $options ` instead of `is_string($options) ? $options : $label`.

lingma: The currently function instruments have the forth parameter as string in this case, so this method could solve the back compatibility issue without causing problem, this is a special case treatment, I think we should not put `array|string` to createBase to accommodate the back compatibility, `string` is always passed in that case, using `options` as `label` is not our purpose, when we use it, the content is `string`

lingma: Oh, no, because we could pass `` as a table header to have a empty space.

maltheism: I'm just wondering why `is_string($options)` would ever return true. Since it's an array? Wouldn't the $label always get returned.

lingma: Please check PR #6444, that PR demonstrates the purpose of this PR

lingma: Oh, that is an unanswerable question that I am trying to fix with PR #6395, which is a little radical, so I have this PR to make backward compatibility.

lingma: I recall that I said there is already some rare problem there (passing array as label for some unknown reason) in some instruments code of v20 that I have observed, which is not functional since Loris version 20.0 in the current code before v22. I also observed that the Loris version 15.0 has the same problem, As a result I believe possibly that the code before v15.0 pass the 3rd parameter as label (as I intended to do), or the instruments have problem for other reasons.

lingma: Those instruments currently display `Array` instead of a text as label in the table header at least in the version 21.0, which is the same instrument code at 20.0 so I believe they should display as "Array" either in 20.0, this behaviour is either due to the concern I was trying to address or some other reasons.

lingma: I just found the code in question in a project: `createElement("static", null, $this->indent)`, the instruments display wrongly both in v20.0 and v21.0 and in production.

image

lingma:
`createLabel` correction is also added here.

Issue #6353, opened 2020-04-09
[user_accounts] error_log pollution when removing pending approval


maltheism:
**Describe the bug**

I receive the following errors in error_log when the admin is removing `pending approval` from a request account. 

`Notice: Undefined index: permID in /Users/alizee/Development/GitHub/McGill/Loris/modules/user_accounts/php/edit_user.class.inc on line 216`

`Warning: array_keys() expects parameter 1 to be array, null given in /Users/alizee/Development/GitHub/McGill/Loris/modules/user_accounts/php/edit_user.class.inc on line 216`

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Request account
2. Go to user_accounts change user from pending approval and check the box Generate new password and Send email to user.
3. See error

johnsaigle:
I'm not able to test this properly right now because request accounts is buggy. I will look into fixing this after #6233 is resolved.

cmadjar:
@johnsaigle #6233 has been resolved. Back to you :).

ridz1208:
@maltheism can you adress this please ?

PR #6460, opened 2020-04-29
[user_accounts] Resolve Notice: Undefined index: permID

maltheism:
## Brief summary of changes

Solves the `undefined index` error from polluting the error_log and when removing pending approval for a user in user_accounts module. 

1. Request account
2. Visit user_accounts module as admin
3. Remove pending approval for account.

#6353

### DIFF
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -213,7 +213,8 @@ class Edit_User extends \NDB_Form
         // 2) Certain old permissions that the editor is not permitted to alter
         // In addition, old permissions should be removed if the editor has
         // removed them and the editor has permissions to remove them.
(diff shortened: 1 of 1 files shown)


modules/user_accounts/php/edit_user.class.inc
```diff
@@ -213,7 +213,8 @@ class Edit_User extends \NDB_Form
         // 2) Certain old permissions that the editor is not permitted to alter
         // In addition, old permissions should be removed if the editor has
         // removed them and the editor has permissions to remove them.
-        $newPermissions    = array_keys($values['permID']) ?? array();
+        $newPermissions    = isset($values['permID'])
```
driusan, line 216: I think all you had to do was move the `??` inside the bracket, but okay..

PR #6461, opened 2020-04-29
[data_release] Preventing Duplication of File Names

HenriRabalais:
## Brief summary of changes
This PR Prevents the duplication of file_names in the data_release table. If there is an attempt to submit a filename that already exists in the table, the user is given the option of overwriting the previous entry.

#### Testing instructions
Part 1
1. Upload a file to data_release.
2. Upload file with same name but different content.
3. Make sure a swal appears asking whether you want overwrite the file.
4. Overwrite file and make sure of it by downloading the new resource.

Part 2
1. Upload file as user who is not super user.
2. Remove permission to that file for that user.
3. Have that user try to overwrite the file.
4. Receive "Forbidden" error message.
5. Repeat steps 1-4 with super user. They should be able to overwrite the file even if they don't have permission for it.

#### Link(s) to related issue(s)
* Resolves #6435

PR #6466, opened 2020-05-01
[Release Notes] Add Data Release Duplicate Specification

HenriRabalais:
NOTE: I've never made a change to release notes before and not sure if this is how it should be done. This addition was requested by @ridz1208 

## Brief summary of changes
If projects have duplicate file names in their data_release tables, this will cause errors when downloading files. Projects should remove duplicates before updating their version.

Related PR:  #6461


HenriRabalais:
Oops wasn't supposed to it like this!

PR #6477, opened 2020-05-06
Fix data in fields where HTML characters were double escaped on 22

ridz1208:
## Brief summary of changes
This PR fixes the issue where special characters entered in instrument forms were double escaped on saving but improperly decoded on instrument loading causing each save to prepend an `&amp;` escaping string to the already escaped characters. the problematic chars are `&`, `<`, `>` and `"`.

The escaping is done in the deleted lines as well as the insert and update functions of the database class making this first instance unnecessary.

A script has also been added to help projects identify if the bug has affected their data. Amending the data in the database will remain the responsibility of each project since it is highly specific to each project.

### DIFF
--- a/php/libraries/LorisForm.class.inc
+++ b/php/libraries/LorisForm.class.inc
@@ -693,10 +693,6 @@ class LorisForm
                 }
             }
         }
-        // Always sanitize user-controlled input
-        if (!is_array($newValue)) {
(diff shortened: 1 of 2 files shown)


ridz1208:
@driusan as agreed, I have added the bugfix and the reporter script
once merged i'd like to push it to 23 before the release

PR #6478, opened 2020-05-06
[issue_tracker] Fix Priority Ordering.

### LINKED ISSUE #6328: [Issue Tracker] Dropdown options display order is alpha, not array-defined

HenriRabalais:
## Brief summary of changes
This PR overrides the default alphabetical sorting of SelectElements for the list of priorities in the issue tracker module.

NOTE: This PR did make me think that maybe the index files for each module should really be passing components for the filter fields rather than a series of parameters from a JSON object. 

1. Go to Issue Tracker module
2. Make sure list of priorities in the menu filter and on the new Issue page are ordered: 
normal, high, urgent, immediate.

* Replaces #6376
* Resolves #6328

### DIFF
--- a/jsx/Filter.js
+++ b/jsx/Filter.js
@@ -66,7 +66,13 @@ class Filter extends Component {
           element = <TextboxElement key={filter.name}/>;
           break;
         case 'select':
-          element = <SelectElement key={filter.name} options={filter.options}/>;
+          element = (
+            <SelectElement
+              key={filter.name}
+              options={filter.options}
+              sortByValue={filter.sortByValue}
+            />
+          );
           break;
         case 'multiselect':
           element = <SelectElement key={filter.name} options={filter.options} multiple={true} emptyOption={false}/>;
--- a/modules/issue_tracker/jsx/IssueForm.js
+++ b/modules/issue_tracker/jsx/IssueForm.js
@@ -249,6 +249,7 @@ class IssueForm extends Component {
             required={false}
             disabled={!hasEditPermission}
             value={this.state.formData.priority}
+            sortByValue={false}
           />
           <SelectElement
             name='category'
--- a/modules/issue_tracker/jsx/issueTrackerIndex.js
+++ b/modules/issue_tracker/jsx/issueTrackerIndex.js
@@ -177,6 +177,7 @@ class IssueTrackerIndex extends Component {
       {label: 'Priority', show: true, filter: {
         name: 'priority',
         type: 'select',
+        sortByValue: false,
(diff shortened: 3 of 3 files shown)


ridz1208:
@maltheism please review and test.

christinerogers:
@driusan could you do a final review? I think this may be ready to merge

driusan:
> NOTE: This PR did make me think that maybe the index files for each module should really be passing components for the filter fields rather than a series of parameters from a JSON object.

Can you explain what you mean?

HenriRabalais:
@driusan In the jsx index file of each reactified module we are passing a `fields` object to the FilterableDatatable which uses the data from the `filter` key to generate all the filter fields. 
e.g.
```
      {label: 'Priority', show: true, filter: {
        name: 'priority',
        type: 'select',
        sortByValue: false,
        options: options.priorities,
        }},	        
       }},
```

I'm just finding that at this point, rather than dynamically generating the appropriate field type based on the attribute `type`, it may be better to simply declare the Field to be displayed right in the index file.

e.g.
```
      {label: 'Priority', show: true, filter: 
        <SelectElement
          name: 'priority'
          sortByValue: false,
          options: options.priorities,
        />,
       }},
```

driusan:
How would that work with filters that have dynamic values which are sent from PHP?

HenriRabalais:
I believe the values sent from PHP should be fine since they are still passed via the `options` prop.

PR #6480, opened 2020-05-06
[issue_tracker] Handle a CenterID of NULL in issues table.

### LINKED ISSUE #6470: [issue_tracker] If Issue has centerID of NULL, error is thrown.
**Describe the bug**
When navigating to the issue tracker module with a user who does not have permission to all centers, if one of the issues has a centerID of `NULL` in the database, an error thrown since the dataprovisioner's usersitematch filter expects getCenterID to return a non-null value.

**To Reproduce**
1. Set the centerID of one of the issues to NULL
2. Load Issue Tracker module with user who is not admin.

Error:
```
[Mon May 04 15:51:57.413901 2020] [php7:error] [pid 31367] [client 10.31.6.189:53417] PHP Fatal error:  Uncaught LorisException: getCenterID on resource returned null in /var/www/loris/src/Data/Filters/UserSiteMatch.php:62\nStack trace:\n#0 /var/www/loris/src/Data/ProvisionerInstance.php(130): LORIS\\Data\\Filters\\UserSiteMatch->filter(Object(User), Object(LORIS\\issue_tracker\\IssueRow))\n#1 [internal function]: LORIS\\Data\\ProvisionerInstance->LORIS\\Data\\{closure}(Object(LORIS\\issue_tracker\\IssueRow), 3, Object(class@anonymous))\n#2 [internal function]: CallbackFilterIterator->accept()\n#3 [internal function]: FilterIterator->rewind()\n#4 /var/www/loris/src/Data/Table.php(83): IteratorIterator->rewind()\n#5 /var/www/loris/src/Data/Table.php(98): LORIS\\Data\\Table->toArray(Object(User))\n#6 /var/www/loris/modules/issue_tracker/php/issue_tracker.class.inc(166): LORIS\\Data\\Table->toJSON(Object(User))\n#7 /var/www/loris/php/libraries/NDB_Menu_Filter.class.inc(698): LORIS\\issue_tracker\\Issue_Tracker->toJSON()\n#8 /var/www/loris/php/libraries/NDB_Page.class.inc(718): NDB_Menu_Filter->display()\n#9 /var/www/loris/php/libraries/NDB_M in /var/www/loris/src/Data/Filters/UserSiteMatch.php on line 62, referer: http://hrabalais-dev.loris.ca/issue_tracker/
```

The page to load and only issues that are associated with the user's center to be displayed.

HenriRabalais:
## Brief summary of changes
This PR handles values of NULL in the centerID column of the `issues` table by converting `NULL` values to an array of all centerIDs in the `psc` table in the `issuerowprovisioner` class.

1. Create an issue with no CenterID, or set centerID to null for an issue already in the `issues` database table.
2. Login as user who does not have access to all centers.
3. Make sure the Issue Tracker Module is working for that user.

* Resolves #6470

### DIFF
--- a/modules/issue_tracker/php/issuerowprovisioner.class.inc
+++ b/modules/issue_tracker/php/issuerowprovisioner.class.inc
@@ -30,16 +30,16 @@ namespace LORIS\issue_tracker;
  */
 class IssueRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisioner
 {
-    private $siteList;
+    private $_siteList;
 
     /**
      * Create a IssueRowProvisioner, which gets rows for the issues menu table.
      */
     function __construct()
     {
-        $this->siteList = array_keys(\Utility::getSiteList(false));
-        $user   = \User::singleton();
-        $userID = $user->getUsername();
+        $this->_siteList = array_keys(\Utility::getSiteList(false));
+        $user            = \User::singleton();
+        $userID          = $user->getUsername();
         //note that this needs to be in the same order as the headers array
         parent::__construct(
             "SELECT i.issueID,
@@ -99,7 +99,7 @@ class IssueRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisioner
             $module = &$modules[$mname];
         }
 
-        $cids = isset($row['centerId']) ? [$row['centerId']] : $this->siteList;
+        $cids = isset($row['centerId']) ? [$row['centerId']] : $this->_siteList;
         $row['lastUpdate'] = \Utility::toDateDisplayFormat($row['lastUpdate']);
         return new IssueRow($row, $cids, $module);
     }
--- a/modules/issue_tracker/php/issuerowprovisioner.class.inc
+++ b/modules/issue_tracker/php/issuerowprovisioner.class.inc
@@ -30,11 +30,14 @@ namespace LORIS\issue_tracker;
  */
 class IssueRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisioner
 {
+    private $siteList;
+
     /**
      * Create a IssueRowProvisioner, which gets rows for the issues menu table.
      */
     function __construct()
     {
+        $this->siteList = array_keys(\Utility::getSiteList(false));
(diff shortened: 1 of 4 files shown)


modules/issue_tracker/php/issuerowprovisioner.class.inc
```diff
@@ -96,8 +96,9 @@ class IssueRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisioner
             $module = &$modules[$mname];
         }
 
-        $cid = $row['centerId'];
+        $cids = isset($row['centerId']) ?
+            [$row['centerId']] : array_keys(\Utility::getSiteList(false));
```
ridz1208, line 100: Normally I would remind you that the proper way of declaring an array is `array($row['centerId'])` but i'll let it pass this time

ridz1208:
@christinerogers I think this could be manually test by one of the students

christinerogers:
Great thanks @ridz1208 - Laetitia @laemtl will pull and test.

modules/issue_tracker/php/issuerowprovisioner.class.inc
driusan, line 100: Can the getSiteList part of this be brought up to the constructor so that it's cached between rows? We shouldn't be re-querying the database to get the site list for every row of the table, that won't scale.

HenriRabalais: Sure! So move it up to the constructor and set it as a class variable?

driusan: Yeah, that way we query the database once per page load instead of once per row.

HenriRabalais:
@driusan done!

laemtl:
I tested and everything works fine! The only thing I noticed (and I don't know if it is really an issue) is that an Issue can't be created with a NULL CenterID. The field is required.

driusan:
@laemtl is that new with this branch or is that the existing behaviour? If it's the existing behaviour, can you open a separate ticket?

laemtl:
It is an existing behaviour. I create an issue!

PR #6483, opened 2020-05-07
[MRI Violations] Fixing a bug in dashboard widget registration and adding some tests

nicolasbrossard:
## Brief summary of changes

- The widget displayed on the `dashboard/My Task` panel for violated scans now specifies whether all known sites or only the user's sites were taken into account when computing the number of violations.
- Deletion from the dashboard module of the tests associated to the violated scans widget in the `My Task` panel.
- Addition of tests in the MRI violations module for the registration of the violated scans widget in the dashboard module.

Make sure that when the dashboard is displayed, there will be either label `Site: All` or `Site: All User Sites` besides the number of violated scans depending on whether the user has the `violated_scans_view_allsites` or `violated_scans_view_ownsite` permission respectively

None

### DIFF
--- a/modules/dashboard/php/taskwidget.class.inc
+++ b/modules/dashboard/php/taskwidget.class.inc
@@ -41,22 +41,25 @@ class TaskWidget implements \LORIS\GUI\Widget
     /**
      * Construct a TaskWidget with the given properties.
      *
-     * @param string $label    The label to describe the task.
-     * @param int    $number   The number of outstanding items for the task.
-     * @param string $link     A URL to go to when the user clicks on the task.
-     * @param string $cssclass A CSS class to add to the element. This is primarily
-     *                         used for testing.
+     * @param string $label     The label to describe the task.
+     * @param int    $number    The number of outstanding items for the task.
+     * @param string $link      A URL to go to when the user clicks on the task.
+     * @param string $cssclass  A CSS class to add to the element. This is primarily
+     *                          used for testing.
+     * @param string $siteLabel Site label to display besides the widget.
      */
     public function __construct(
         string $label,
         int $number,
         string $link,
-        string $cssclass
+        string $cssclass,
+        string $siteLabel=''
     ) {
-        $this->label    = $label;
-        $this->number   = $number;
-        $this->link     = $link;
-        $this->cssclass = $cssclass;
+        $this->label     = $label;
+        $this->number    = $number;
+        $this->link      = $link;
+        $this->cssclass  = $cssclass;
+        $this->sitelabel = $siteLabel;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


modules/dashboard/test/DashboardTest.php
```diff
@@ -510,7 +424,7 @@ public function testConflictResolver()
         $this->_testMytaskPanelAndLink(
             ".conflict_resolver",
             "574",
-            "- Conflict Resolver"
+            "- Conflict Resolzzzz"
```
kongtiaowang, line 427: ?

nicolasbrossard: Oooppss... :) Fixed.

ridz1208:
@kongtiaowang can you test and re-review please

kongtiaowang:
LGTM

nicolasbrossard:
@driusan There are MRI violations that are no associated to any site and you want to take these to be part of the count displaying by the widget. Creating a `TaskQueryWidget` would not accomplish this since the query it builds discards the violations with sites that are _not_ set.

modules/mri_violations/test/mri_violationsTest.php
```diff
@@ -581,7 +592,65 @@ function testResolvedSearchButton()
             "Site",
             "TESTinPSC"
         );
+    }
+
+    /**
+     * Verify that for a user with 'violated_scans_view_allsites' permission the
+     * number of MRI violated scans is reported in the My Task panel.
+     * Also ensure that when you click on this tas, the link takes you to the
```
driusan, line 600:
```suggestion
     * Also ensure that when you click on this tas, the link takes you to the
```

modules/mri_violations/test/mri_violationsTest.php
```diff
@@ -581,7 +592,65 @@ function testResolvedSearchButton()
...
+            "208",
+            "- MRI Violated Scans"
+        );
+        $this->resetPermissions();
+    }
 
+    /**
+     * Verify that for a user with 'violated_scans_view_ownsite' permission the
+     * number of MRI violated scans is reported in the My Task panel.
+     * Also ensure that when you click on this tas, the link takes you to the
```
driusan, line 627:
```suggestion
     * Also ensure that when you click on this task, the link takes you to the
```

modules/dashboard/php/taskwidget.class.inc
```diff
@@ -89,7 +109,7 @@ class TaskWidget implements \LORIS\GUI\Widget
      */
     public function siteLabel() : string
     {
-        return "";
+        return $this->sitelabel;
```
driusan, line 112: This is still going to change the behaviour  on any taskwidgets that don't call setSiteLabel since it changed from the empty string to null and I'm still not crazy about the setter.

All the other parameters are injected via the constructor, and you call setSiteLabel immediately after the constructor in the only place it's used. Why don't you pass a site label in the constructor with a default value of ""? I think that would take care of both problems without changing any behaviour. of widgets in any other modules that don't call setSiteLabel.

modules/mri_violations/test/mri_violationsTest.php
nicolasbrossard: Ok

modules/mri_violations/test/mri_violationsTest.php
nicolasbrossard: Ok

modules/dashboard/php/taskwidget.class.inc
nicolasbrossard: Changed in last commit.

ridz1208:
@driusan all urs

PR #6486, opened 2020-05-07
Password validate static

### LINKED ISSUE #6449: Make Password->validate() a static function
Right now the idiom for testing password strength is doing

```php
try {
    new Password($input);
}
```

laemtl:
## Brief summary of changes
I made Password::validate() static. Password::validate() can be used directly for password validation in place of the Password constructor.

Side note: trim_trailing_whitespace = true in .editorconfig, and some lines were altered on save.

* Resolves #6449

### DIFF
--- a/modules/login/php/passwordexpiry.class.inc
+++ b/modules/login/php/passwordexpiry.class.inc
@@ -90,7 +90,7 @@ class PasswordExpiry extends \NDB_Form
         $plaintext   = htmlspecialchars_decode($values['password']);
 
         try {
-            new \Password($plaintext);
+            \Password::validate($plaintext);
         } catch (\InvalidArgumentException $e) {
             $this->tpl_data['error_message'] = $e->getMessage();
             return array('password' => $e->getMessage());
--- a/modules/my_preferences/php/my_preferences.class.inc
+++ b/modules/my_preferences/php/my_preferences.class.inc
@@ -392,11 +392,11 @@ class My_Preferences extends \NDB_Form
         // if password is user-defined, and user wants to change password
         if (!empty($values['Password_hash'])) {
             try {
-                // The Password class is self-validating and will throw an
+                // Password::validate will throw an
                 // InvalidArgumentException if the value passed to it is
                 // bad or the input is too weak to be a good password.
                 $decoded = htmlspecialchars_decode($plaintext);
-                new \Password($decoded);
+                \Password::validate($decoded);
                 // New password must be different than current one
                 if (! \User::factory($this->identifier)->isPasswordDifferent(
                     $decoded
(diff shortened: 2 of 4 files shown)


driusan:
It shouldn't be static, it should be public. Static functions can't be overloaded by subclasses.

christinerogers:
@johnsaigle any objections to Dave's comment, since you made the original issue?

johnsaigle:
If it's not static then I think we would need to use the idiom `new Password()` as before because the object will need to be created in order to use the method. The issue was originally created to avoid this.

driusan:
Any code using the class would need to change regardless of whether it's `public` or `static public`. The only difference at the call site is the syntax used.

johnsaigle:
Right, I want the syntax change. I want us to be able to do

```php
Password::validate($input);
```

instead of
```php
new Password($input);
```

so it needs to be static.

driusan:
There are a lot of bad practices incorporated into:

```
try {
    new Password("input");
} catch {
 ...
}
```

Making the function static only solves one of them.

```
try {
    Password::validate($input);
} catch {
 ...
}
````

still has code smells of:
1. Using static methods on objects as namespaces for functions (instead of using namespaces and functions)
2. Using exceptions as control flow structures for conditions that occur under normal circumstances.

The code should be something akin to:

```
if (new Password($input))->validate()) {
    // do something 
}
```

(Sorry @lamtl I just realized this is a PR, not an issue and you've already made it static, but If we're going to fix it, we should fix it properly.)

johnsaigle:
@driusan The use of exceptions occurs because class is designed as a Value Object: there's no point creating a Password object that contains an invalid plaintext. 

The design is based on other Value Object classes in LORIS such as CandID and ValidatableIdentifier which also throw DomainExceptions in their constructors when given an invalid value. This code in these classes was written by Xavier and I think the use of value objects in scenarios like this are very beneficial.

Using a Domain Exception also allows us to pass precise error messages from within the password class. By changing `validate()` to a bool used by calling code would necessitate the use of something like 
* an `$errmsg` class property for Password, or
* creating a `getFailureResason()` method, or 
* hard-coding the error messages into the calling code everywhere the Password is used

All of these seem to be more clunky than the current code.

driusan:
> there's no point creating a Password object that contains an invalid plaintext.

That is clearly not true or this PR wouldn't exist. I can't think of any reason you'd ever need an object that contains a valid password. After validation is done you're done working with the password and don't need it anymore.

> Using a Domain Exception also allows us to pass precise error messages from within the  password class. By changing validate() to a bool used by calling code would necessitate the >use of something like
>
>    an $errmsg class property for Password, or
>    creating a getFailureResason() method, or
>    hard-coding the error messages into the calling code everywhere the Password is used

If you need to know the reason validation failed then  make validate return an error string.

johnsaigle:
@laemtl Sorry to derail your PR with this discussion.

@driusan Maybe this PR should be closed for now and we can work out a Password redesign in the original issue? I'm not really sure how this PR can proceed if you have objections to the underlying class.

christinerogers:
@johnsaigle Could you open a new Issue that `Needs Discussion` to cover the Password redesign? Laetitia could probably work on that with Dave, to get these resolved quickly.

PR #6498, opened 2020-05-07
[Raisinbread] Instructions and clarifications added to RaisinBread README

### LINKED ISSUE #6495: [Raisinbread] Add instructions about sourcing SQL to the Raisinbread README
The Raisinbread installation instructions in the [README.md](https://github.com/aces/Loris/blob/23.0-release/raisinbread/README.md) (23.0-release branch) do not specify that all tables must be dropped from a database before the raisinbread files can be sourced. It should also be clarified that there are potential foreign key constraints that can be mitigated.

- [ ] Clarify that all tables should be dropped from the DB pre-installation.

- [ ] Add a warning about foreign key constraints and alternative instructions for this case.

AlexandraLivadas:
The Raisinbread README was updated to include specific instructions for dropping all tables pre-installation. A note about foreign key constraints and sourcing the schema files directly on the MySQL command line was also added. 

* Resolves #6495

### DIFF
--- a/raisinbread/README.md
+++ b/raisinbread/README.md
@@ -29,8 +29,17 @@ that is not the case, replace all `mysql` commands below by the necessary values
 `mysql -u user -p -h host database_name`*
 
 ##### Sourcing SQL
-In order to be able to use the RaisinBread dataset, the LORIS SQL schema needs to be 
-sourced, followed by the different instrument schemas and finally the actual RB data. 
+If the database being used is already populated and contains any tables or data, the following
+command must be used in the main LORIS root directory:
+
+```bash
+cat raisinbread/instruments/instrument_sql/9999-99-99-drop_instrument_tables.sql \
+    SQL/9999-99-99-drop_tables.sql | mysql
+```
+***Note:** This command can also be used at any step to empty and delete all RaisinBread tables.*
+
+In order to be able to use the RaisinBread dataset, the LORIS SQL schema needs to be
+sourced, followed by the different instrument schemas and finally the actual RB data.
 The commands below assume that the current working directory is the main LORIS root
 directory.
 
@@ -49,13 +58,14 @@ cat SQL/0000-00-00-schema.sql \
     raisinbread/RB_files/*.sql | mysql
 ```
 
-Note: to empty and delete all RaisinBread tables, use the following command in the 
-main LORIS root directory
-```bash
-cat raisinbread/instruments/instrument_sql/9999-99-99-drop_instrument_tables.sql \
-    SQL/9999-99-99-drop_tables.sql | mysql
+***Important:** Ensure that the above commands (including those to drop tables) were all completed properly.* 
+If there is a foreign key constraint error, run the following command on the mysql command line to mitigate the error:
+
+```bash 
+SET FOREIGN_KEY_CHECKS=0;
 ```
 
+If the tables were not deleted or created properly, the above schemas can be sourced directly on the mysql command line.
 
 ##### Configuring
 In order to be able to load the LORIS front-end while using the RaisinBread dataset 
--- a/raisinbread/README.md
+++ b/raisinbread/README.md
@@ -30,13 +30,16 @@ that is not the case, replace all `mysql` commands below by the necessary values
 
 ##### Sourcing SQL
 If the database being used is already populated and contains any tables or data, the following
-command must be used in the main LORIS root directory:
+command must be used in the main LORIS root directory. Note that these commands will erase all the data
+in the database. Ensure that a backup is available if this data is important:
 
 ```bash
 cat raisinbread/instruments/instrument_sql/9999-99-99-drop_instrument_tables.sql \
     SQL/9999-99-99-drop_tables.sql | mysql
 ```
 ***Note:** This command can also be used at any step to empty and delete all RaisinBread tables.*
+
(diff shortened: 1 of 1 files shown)


raisinbread/README.md
```diff
@@ -49,13 +58,14 @@ cat SQL/0000-00-00-schema.sql \
     raisinbread/RB_files/*.sql | mysql
 ```
 
-Note: to empty and delete all RaisinBread tables, use the following command in the 
-main LORIS root directory
-```bash
-cat raisinbread/instruments/instrument_sql/9999-99-99-drop_instrument_tables.sql \
-    SQL/9999-99-99-drop_tables.sql | mysql
+***Important:** Ensure that the above commands (including those to drop tables) were all completed properly.*
```
ridz1208, line 61: Users tend to read as they apply commands so by the time they get to this note and if the drop commands did not complete successfully, they most likely already had an issue sourcing the default schemas. I would recommend adding a note like this one around line 40 asking the users to confirm that the drop statements got rid of all the tables in the database before moving on to the sourcing part.

raisinbread/README.md
```diff
@@ -29,8 +29,17 @@ that is not the case, replace all `mysql` commands below by the necessary values
 `mysql -u user -p -h host database_name`*
 
 ##### Sourcing SQL
-In order to be able to use the RaisinBread dataset, the LORIS SQL schema needs to be 
-sourced, followed by the different instrument schemas and finally the actual RB data. 
+If the database being used is already populated and contains any tables or data, the following
+command must be used in the main LORIS root directory:
```
ridz1208, line 33: I would recommend rephrasing this sentence to indicate very clearly that the commands below will effectively erase all data in the database and to make sure a backup is available if the existing data is immportant.

raisinbread/README.md
AlexandraLivadas: Sounds good!

raisinbread/README.md
AlexandraLivadas: I will add that information now. Thank you :)

raisinbread/README.md
```diff
@@ -49,13 +58,14 @@ cat SQL/0000-00-00-schema.sql \
...
-Note: to empty and delete all RaisinBread tables, use the following command in the 
-main LORIS root directory
-```bash
-cat raisinbread/instruments/instrument_sql/9999-99-99-drop_instrument_tables.sql \
-    SQL/9999-99-99-drop_tables.sql | mysql
+***Important:** Ensure that the above commands (including those to drop tables) were all completed properly.* 
+If there is a foreign key constraint error, run the following command on the mysql command line to mitigate the error:
+
+```bash 
+SET FOREIGN_KEY_CHECKS=0;
```
ridz1208, line 65: We generally discourage the use of the `SET FOREIGN_KEY_CHECKS=0;` as it is generally a sign that there is an issue somewhere else basically. I will not block this addition in this PR because the ultimate purpose is to erase the database however the following points are very important.

- `SET FOREIGN_KEY_CHECKS=0;` can only be used right before the droping of the tables and should immediately be undone with `SET FOREIGN_KEY_CHECKS=1;` after the dropping and before the sourcing of the schema !!
- the need for `SET FOREIGN_KEY_CHECKS=0;` signifies that there is an issue with our `9999-99-99 ...` drop patches which should be corrected instead of bypassed. which means depending on the reason the drops are failing with foreign key issues, the developer should instead resolve the issue and submit a fix to LORIS

AlexandraLivadas: I can remove this section and instead work on fixing the bug with the `9999-99-99 ... `patches. Is that a better approach?

ridz1208: yes for sure

driusan:
@ridz1208 is the Needs Work tag on this still valid?

ridz1208:
@driusan i don't think so

raisinbread/README.md
```diff
@@ -49,13 +61,7 @@ cat SQL/0000-00-00-schema.sql \
...
 ```
 
-Note: to empty and delete all RaisinBread tables, use the following command in the 
-main LORIS root directory
-```bash
-cat raisinbread/instruments/instrument_sql/9999-99-99-drop_instrument_tables.sql \
-    SQL/9999-99-99-drop_tables.sql | mysql
-```
-
+If the tables were not deleted or created properly, the above schemas can be sourced directly on the mysql command line.
```
driusan, line 64: I don't understand what this sentence is trying to explain or instruct the user to do.

ridz1208: I think it's in contrast to the `cat` command. 

the suggested command will stop running after the first failure while sourcing the script directly in the database will be more explicit and continues running the remainder of the commands after 1 has failed

driusan: I think that might be clearer if it was in the paragraph above the code which currently says:

"The commands below assume that the current working directory is the main LORIS root
directory."

Something like:

"The commands below assume that the current working directory is the main LORIS root
directory. If the tables are not deleted or created properly, the schemas files can be sourced directly on the mysql command line."

(I can't suggest changes because it's not a line changed by the commit..)

ridz1208: @AlexandraLivadas that makes sens to me, you can make the changes

AlexandraLivadas: Sounds good!

christinerogers:
John's right we have a [script for this](https://github.com/aces/Loris/issues/6495#issuecomment-626763538)

Issue #6473, opened 2020-05-05
[Readme] Update dependency impacting ubuntu dev install - minor tweak


christinerogers:
php7.3-curl specifically is required for the 23 release 
- [ ] change `curl` -> `php7.3-curl` in the main LORIS readme

laemtl:
Christine, is it possible that both curl and php7.3-curl are needed?

johnsaigle:
This package is only needed for development installs, not production, and it's required because of the change introduced here: https://github.com/aces/Loris/pull/6187.

Just mentioning it for the record. The linked PR looks good.

christinerogers:
thanks John.  
Is it still recommended that production instances stick to `curl`?  
Since this breaks the install for sandboxes, I think we should mention the package for development - which means we should update related documentation not yet covered by #6187.

christinerogers:
Updating issue title to be better: 
- this issue scope covers only the main Loris Readme on the 23 branch
- other install docs shoujld be added to the 23 branch and similarly update via # 6475

johnsaigle:
> Is it still recommended that production instances stick to curl?

`curl` is always needed AFAIK. `php7.3-curl` is needed for development environments.

ridz1208:
@johnsaigle so we just don't know why curl is required ?

johnsaigle:
No I don't know the reason. On the other hand I don't know if I could tell you why we need all of the other dependencies either. We could try to set up a prod environment without `curl` and wait to see if it breaks - otherwise I'm not sure how to test this.

ridz1208:
@johnsaigle I think I can justify some of them but curl... never knew why or if it's a new thing or always been like that. 

maybe we should run a test if it's a 23 thing. if it was there from before, we can actually see if all existing PRODs have curl

driusan:
It's likely that all existing prods have curl even if LORIS doesn't require it because it's such a common tool.

johnsaigle:
@ridz1208 It's been there for a very long time. I grabbed it from https://github.com/aces/Loris/wiki/Installing-Loris-in-Depth and I think that section was last edited by Justin Ng. It's possible that Guzzle uses it? But I'm not positive.

As Dave said it's a very common tool, similar to `software-properties-common` or `zip` which we also have as a requirement.

ridz1208:
@driusan @johnsaigle alright then... both it is

ridz1208:
changes done

PR #6512, opened 2020-05-08
[Changelog] Add note re php-curl dependency now required for dev install

### LINKED ISSUE #6509: [Changelog] Add note re php-curl dependency now required for dev install
Given `php7.3-curl` is now required for dev installs since #6187 --
This should be noted in the Changelog so that it makes it into the 23 Release Notes under the `Notes for Existing Projects` section

@laemtl -- we've just started the changelog for this release so let's see how the PR discussion goes with the team for this kind of change...

laemtl:
I added a note in Changelog under the Notes for Existing Projects section for the new  php7.3-curl dev install dependencies. I also fixed a redundancy in the file.

* Resolves #6509

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,18 +34,13 @@ database (PR #5260)
   modules. (Various PRs)
 
 ##### Issue Tracker
-- The issue_tracker module now has the feature of uploading attachments to new or existing issues.
+- The issue_tracker module now has the feature of uploading attachments to new or existing issues. (PR #5394)
 - All sites now appear in the dropdown for site, not only study sites. (PR #6135)
-
 
 #### Battery Manager
 - New module created to manage the entries in the `test_battery` table of the database.
 This allows projects to modify their instrument battery without requiring backend access.
  (PR #4221)
- 
-#### Issue Tracker
-- The `issue_tracker` module now has the feature of uploading attachments to new or existing
-issues. (PR #5394)
  
 #### Module Manager
 - New module created to manage the status of installed modules. (PR #6015)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -74,6 +68,7 @@ exception). It is recommended to run this tool for existing projects (PR #5270)
 
 ### Notes For Existing Projects
 - PHP should be upgraded to 7.3 to before upgrading LORIS.
+- For dev instances, php7.3-curl is now a required dependency.
```
ridz1208, line 71: @AlexandraLivadas why only for DEV ?

christinerogers: I think you mean Laetitia @laemtl 

@johnsaigle's PR header in #6187 suggests this is only an issue on Dev environments. 
John was also [asked for clarification on this exact point](https://github.com/aces/Loris/issues/6473#issuecomment-624948602) but not response so far... 

@ridz1208 Are you confirming that it applies to production too or just asking?

ridz1208: from what I can see it is indeed just a dev thing. I think it's still worth properly testing however

christinerogers: 👍 thanks @ridz1208 this might need shen's help with the install VMs

kongtiaowang: @ridz1208 @christinerogers This is correct, php-webdriver requires the php7.3-curl in the dev environment.

johnsaigle: @ridz1208 It's definitely only a dev requirement. https://github.com/aces/Loris/pull/6187 updates a package that was renamed. This package requires php7.X-curl. That package drives our integration tests. If you're not running integration tests, the package and its dependency are not required. It's only installed when someone runs `make dev`.

christinerogers: thanks for confirming @johnsaigle

christinerogers:
Looks like this is ready to merge for 23 @driusan 

Thanks @laemtl

PR #6514, opened 2020-05-08
[Documentation] DQT readme optional steps clarification

### LINKED ISSUE #6469: [Documentation] DQT readme - minor tweak suggested
an improvement suggested in closed PR thread #6267

> I would prefer to indicate when it is required. Other options might be invalidated in the future.
> Similar change should also be applied to the MRI section below.

```suggestion
This step is optional and is required only if instruments are used: \
```

_Originally posted by @PapillonMcGill in https://github.com/aces/Loris/pull/6267_

laemtl:
Small tweak to improve the phrasing in the DQT README.md file. 
Clarifies the steps which are optional and the scope when they are required.

* Resolves #6469

### DIFF
--- a/modules/dataquery/README.md
+++ b/modules/dataquery/README.md
@@ -70,13 +70,13 @@ curl -H 'Content-Type: application/json' -X POST http://$YOURCOUCHDBADMIN:$YOURC
 
 ###### Import the LORIS instrument data
 
-This step is optional and not required if only the MRI portion of LORIS is used:
+This step is optional and is required only if instruments are used: 
 
 `php CouchDB_Import_Instruments.php`
 
 ###### Import the LORIS MRI data
 
-This step is optional and not required if the MRI portion of LORIS isn't installed:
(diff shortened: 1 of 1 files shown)


modules/dataquery/README.md
```diff
@@ -70,13 +70,13 @@ curl -H 'Content-Type: application/json' -X POST http://$YOURCOUCHDBADMIN:$YOURC
...
 
-This step is optional and not required if only the MRI portion of LORIS is used:
+This step is optional and is required only if instruments are used: 
 
 `php CouchDB_Import_Instruments.php`
 
 ###### Import the LORIS MRI data
 
-This step is optional and not required if the MRI portion of LORIS isn't installed:
+This step is optional and is required only if instruments are used:
```
driusan, line 79: This change is not correct. Whether or not the MRI import part needs to be done has nothing to do with whether or not instruments are used.

laemtl: @driusan Ready for review

PR #6521, opened 2020-05-12
[IssueTracker] ActiveIssues tab displays Status filter: Closed

### LINKED ISSUE #6517: [IssueTracker] ActiveIssues tab displays Status filter: Closed
In the Issue Tracker `v23` : The `Active Issues` tab displays `Closed` in the Status filter element because the tab automatically filters for all issues *not Closed* (details below). 

The actual filtering _logic_ works well & as intended, and the data table filters appropriately for only active issues (*not* Closed). So this is a minor bug but still something users shouldn't be seeing. 

* Seems to be introduced during Reactification (#4112) 

Source: Active Issues tab `filterPresets` are defined: `{value: 'closed', opposite: true}}},` in [issueTrackerIndex.js Line 211](https://github.com/aces/Loris/blob/23.0-release/modules/issue_tracker/jsx/issueTrackerIndex.js#L211): 
```php
{label: 'Active Issues', filter: {status: {value: 'closed', opposite: true}}}, 
```

Can you advise on another way to select "all issues except Closed", @HenriRabalais ?

To reproduce: 
* Open the Issue Tracker
* Click `Active` tab. `Closed` appears printed in the Status filter, but the issues in the data table are clearly all except Closed issues (i.e. active)

Tested and reproduced on Firefox and Chrome, Loris 23 on test-loris-dev.loris.ca and sandbox

Originally reported in #6141

laemtl:
I updated the react component FilterableDataTable to add a parseFilter(filter) function.
These function loop over the filters, and copy them in a clone only if they do not have the opposite property. This is applied to the Filter component filters to prevent some field value like status to be updated when opposite is selected.

* Resolves #6517

### DIFF
--- a/jsx/FilterableDataTable.js
+++ b/jsx/FilterableDataTable.js
@@ -37,20 +37,6 @@ class FilterableDataTable extends Component {
   }
 
   /**
-   * Parse filter to remove filter properties when opposite is set to true
-   *
-   * @param {object} filter passed from FilterForm
-   * @return {object} filter
-   */
-  parseFilter(filter) {
-    const parsedFilter = {};
-    for (const [filterId, filterProps] of Object.entries(filter)) {
-      if (filterProps.opposite === undefined || !filterProps.opposite) parsedFilter[filterId] = filterProps;
-    }
-    return parsedFilter;
-  }
-
-  /**
    * Sets Filter to empty object
    */
   clearFilter() {
@@ -64,7 +50,7 @@ class FilterableDataTable extends Component {
         name={this.props.name + '_filter'}
         id={this.props.name + '_filter'}
         columns={this.props.columns}
-        filter={this.parseFilter(this.state.filter)}
+        filter={this.state.filter}
         fields={this.props.fields}
         updateFilter={this.updateFilter}
         clearFilter={this.clearFilter}
--- a/jsx/FilterableDataTable.js
+++ b/jsx/FilterableDataTable.js
@@ -37,6 +37,20 @@ class FilterableDataTable extends Component {
   }
 
   /**
+   * Parse filter to remove filter properties when opposite is set to true
+   *
+   * @param {object} filter passed from FilterForm
+   * @return {object} filter
+   */
+  parseFilter(filter) {
+    const parsedFilter = {};
+    for (const [filterId, filterProps] of Object.entries(filter)) {
+      if (filterProps.opposite === undefined || !filterProps.opposite) parsedFilter[filterId] = filterProps;
+    }
+    return parsedFilter;
+  }
+
+  /**
    * Sets Filter to empty object
    */
   clearFilter() {
@@ -50,7 +64,7 @@ class FilterableDataTable extends Component {
         name={this.props.name + '_filter'}
         id={this.props.name + '_filter'}
         columns={this.props.columns}
-        filter={this.state.filter}
+        filter={this.parseFilter(this.state.filter)}
         fields={this.props.fields}
         updateFilter={this.updateFilter}
         clearFilter={this.clearFilter}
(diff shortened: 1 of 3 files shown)


jsx/FilterableDataTable.js
```diff
@@ -36,6 +36,20 @@ class FilterableDataTable extends Component {
...
+  /**
+   * Parse filter to remove filter properties when opposite is set to true
+   *
+   * @param {object} filter passed from FilterForm
+   * @return {object} filter
+   */
+  parseFilter(filter) {
+    const parsedFilter = {};
+    for (const [filterId, filterProps] of Object.entries(filter)) {
+      if (filterProps.opposite === undefined || !filterProps.opposite) parsedFilter[filterId] = filterProps;
```
christinerogers, line 48: nice. 
* any other conditions/operators besides opposite that should be included? 
* If i'm reading this right, isn't there a case where the empty const `{}` would be returned?

laemtl: That's right, an empty object can be returned. I tested that case and it seemed to be fine.
There is no other conditions we can include that I noticed, but maybe Henri will have suggestions.

HenriRabalais: This seems like a good way of removing the value from displaying in the filter. If we want to push this through quickly, I think this works. 

However, my main concern with removing the value from the filter is that it obfuscates what's being filtered, and I generally don't think we should be hiding from the user what's being filtered. I'm also concerned about building more code infrastructure around the single-use `opposite` paradigm, which would need an overhaul when we properly implement excludes functionality. Preferably, a structure where users can choose to exclude certain values from their filter would be implemented, which would essentially mean removing this code in the long term.

In summary, I think this patch does what it's intended to do well, but I'm worried it moves the filtering in an alternative direction to what we should be moving towards to be handling this sort of functionality.

christinerogers: Would it be better just to change the element to a multi-select?  no obfuscation required.

laemtl: Sure, I can do that. The filter will still need to be modified for the complement field set.

HenriRabalais: Yes, making it a multi-select and have it auto-select all the values except for "close" could definitely work!

christinerogers: ok.  (inhale) let's do that.  i'll ping the team channel and otherwise @laemtl go for it.

laemtl:
Status is now a multiselect field.
What have been done:

issueTrackerIndex.js: filter can now take several comma separated values. 
```
const filterPresets = [
      {label: 'Active Issues', filter: {status: {value: 'assigned, new', opposite: true}}},
];
```
Filter.js has been modified to handle multiple selected options and the opposite attribute.
If filed.type is multiselect
* Element value is an array 
* If opposite is true, the selected options are the complement of the ones provided by the filter

laemtl:
I found an issue with the new multiselect filter. It works fine when updated with the tabs, but not if we select the filter manually. I'm working on this.

laemtl:
Problem solved. I'm migrating the multiple select feature to a new issue #6527

HenriRabalais:
I'm pretty sure you should be able to pass all the status as an array and then the `Filter` component should be able to handle it.

```
const filterPresets = [
      {label: 'Active Issues', filter: {status: {value: ['acknowledged', assigned', 'feedback', 'new', 'resolved'], opposite: true}}},
];
```

laemtl:
@HenriRabalais Thank you, it works well, with less modifications to the code base.

HenriRabalais:
No worries :) Glad it works!

HenriRabalais:
@laemtl I believe this should be closed in favour of the merged #6529

christinerogers:
agreed, closing.  thanks @HenriRabalais

PR #6522, opened 2020-05-12
[issue_tracker] Module filter has entry [object Object]

### LINKED ISSUE #6519: [issue_tracker] Module filter has entry [object Object]
In the Issue Tracker v 23.0, the module filter contains an entry labeled [object Object]

To reproduce:
* Open the Issue Tracker
* Click on the Module filter

laemtl:
I updated the issue_tracker class to avoid Object object to be populated in the Module dropdown.
I used getLongName() to retrieve the name of the module and populate the json.

* Resolves #6519

### DIFF
--- a/modules/issue_tracker/php/issue_tracker.class.inc
+++ b/modules/issue_tracker/php/issue_tracker.class.inc
@@ -115,12 +115,18 @@ class Issue_Tracker extends \NDB_Menu_Filter_Form
             $assignees[$a_row['UserID']] = $a_row['Real_name'];
         }
 
-        $mod_descriptors = \Module::getActiveModules($db);
-        $modules         = array();
-        foreach ($mod_descriptors as $mdesc) {
-            $name           = $mdesc->getLongName();
-            $modules[$name] = $name;
-        }
+        $modules = array_reduce(
+            \Module::getActiveModules($db),
+            function (
+                $result,
+                $module
+            ) {
+                $moduleName          = $module->getLongName();
+                $result[$moduleName] = $moduleName;
+                return $result;
+            },
+            []
+        );
 
         $priorities = array(
             'normal'    => 'Normal',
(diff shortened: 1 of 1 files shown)


driusan:
@lamtl needs PHPCS

```
FILE: .../travis/build/aces/Loris/modules/issue_tracker/php/issue_tracker.class.inc

--------------------------------------------------------------------------------

--------------------------------------------------------------------------------

 120 | ERROR | [x] Equals sign not aligned with surrounding assignments;

     |       |     expected 11 spaces but found 1 space

--------------------------------------------------------------------------------

--------------------------------------------------------------------------------
```

laemtl:
Sorry I forgot to check before my PR!

modules/issue_tracker/php/issue_tracker.class.inc
```diff
@@ -115,7 +115,12 @@ class Issue_Tracker extends \NDB_Menu_Filter_Form
             $assignees[$a_row['UserID']] = $a_row['Real_name'];
         }
 
-        $modules = \Module::getActiveModules($db);
+        $mod_descriptors = \Module::getActiveModules($db);
```
xlecours, line 118: I'll be a bit picky here but the variable name is not quite accurate. The returned value of `\Module::getActiveModules` is an array of `\Module` instances. It should probably be `$modules` but I understand that the intent is to use $modules to store the results of the next foreach loop. `$active_modules` would do it.

Suggestion: You could make 2 hits one stone by using array_reduce.
```
$modules = array_reduce(\Module::getActiveModules($db), function ($carry, $module) {
    ...
}, [])
```

ridz1208:
@h-karim can you please test it out and add the passed manual tests if everything is good

h-karim:
LGTM: Everything works well, filter shows all the modules.

christinerogers:
@driusan final review

Issue #6302, opened 2020-04-06
[Genomic Browser] Issues with the Profiles Tab Datatable.


HenriRabalais:
**Describe the bug**
A. Since there are no files uploaded in raisinbread for the Genomic Browser, clicking on a file value is not possible. Therefore, the "viewGenomicFile" file view cannot be tested as outlined by the test plan.

B. When there is a SNP for a candidate, the "Y" can be clicked to go to the SNP for that candidate, but the link is not obvious since it is presented as regular text. (PR SENT #6523)

C. There are no profiles with CNV or CPG data, so clicking on the link for those can't be tested.

Test Plan Reference: https://github.com/aces/Loris/blob/master/modules/genomic_browser/test/TestPlan.md#special-formated-columns

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to the Profiles Tab of the Genomic Browser
2. Look at the columns for File, SNP, CNV and CPG

![image](https://user-images.githubusercontent.com/7550451/78594803-f009a900-7816-11ea-9628-8b5b82a9e65d.png)

ridz1208:
@xlecours will you be able to address this ?

christinerogers:
My $0.02: 
a+c are RaisinBread/test-data issues, like #6298.  
b is a design choice, not a bug - could be improved upon but not sure it's release-critical.

xlecours:
@ridz1208 It would be low on my priorities list.

christinerogers:
I think this issue can be closed after #6523 is resolved/merged.  Not testable in this round at least.

christinerogers:
splitting off issue C for @AlexandraLivadas

driusan:
The referenced PR says it only partially resolves this.

PR #6523, opened 2020-05-12
[Genomic Browser] Make SNP,CNV,CPG 'Y' links obvious

h-karim:
**Summary**

The link for a candidate's SNP (as well as the CNV, CPG) from the Genomic browser's profile tab doesn't have any visual indication that it can be clicked on. I changed the tag from `<span>` to `<a>` around the cell entry to properly format it as a hyperlink. 

**Testing instructions**
Navigate to the genomic browser, observe the SNP, CNV and CPG columns (CNV and CPG columns do not have data for their links to be tested as per #6302) . Cells with a 'Y' should be colored blue and change color on hover with an underline:

image

**Link(s) to related issue(s)**

* Related to #6302 but only addresses point B.

### DIFF
--- a/modules/genomic_browser/jsx/profileColumnFormatter.js
+++ b/modules/genomic_browser/jsx/profileColumnFormatter.js
@@ -54,14 +54,14 @@ function formatColumn(column, cell, rowData, rowHeaders) {
       if (cell === 'Y') {
         reactElement = (
           <td>
-            <span
+            <a
               style={{cursor: 'pointer'}}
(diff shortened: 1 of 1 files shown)


h-karim:
Noted @christinerogers. Should I create a separate issue to address this behaviour?

christinerogers:
@xlecours - tagging you in case you'd like to review. 

@driusan a very simple cosmetic fix, it should be ready for your review/merge for `23.0-release`

PR #6525, opened 2020-05-12
[RaisinBread] Change participant accounts emails to different emails

### LINKED ISSUE #6515: [RaisinBread] Remove personal emails from raisinbread tables
**Description**
It seems as one of the maintainers' email (Leigh Evans) is placed as one of the candidates' emails in the raisinbread database:

**To Reproduce**
1. From home, navigate to 'Survey Accounts' by clicking on the 'Admin' tab on the top right
2. Observe the following:
image
One of the emails is not identical to the rest.

All of the emails to have been the same, or alternatively aren't real emails.

h-karim:
**Summary**
I changed the participants' emails to a generic ' person'[1..] [at] example ' for all the participants.

**Testing instructions**

With a fresh RB install:
- Navigate to 'Survey Accounts' under Admin tools
- Observe the emails column

* Resolves #6515


christinerogers:
@AlexandraLivadas can you review this one?

SQL/Cleanup_patches/2020-05-12-ChangeCandidateEmails.sql
```diff
@@ -0,0 +1,4 @@
+--Changes survey accounts emails to person[1...]@example.com
```
ridz1208, line 1: This patch is dangerous!!

I'm guessing this is what you used on the RB dataset to fix the emails. Now imagine we merge it and a real project runs this patch on their database. It will effectively erase all the real data and replace it with fake emails... Although we do want that for RB, we really don't for IBIS, CCNA, ...

So basically please remove this file

h-karim: For future reference, do changes to the RB database need a (cleanup) patch file for existing instances?

ridz1208: I would say it's the other way around.

In this case, we identified a flaw with RB and needed to fix it. this is not an issue for projects with active data. so the PR should only address the RB data.

In other cases, for example if we find a bug in the software, then a patch needs to be created for ALL projects (including live datasets) and we treat RB as one of them so we apply the patches on RB as well to keep our mock dataset up to date.

Quick example, let's say that the previous release of loris had an issue where it would automatically append a `+` sign to all user's first names. to fix the issue, we would fix the bug in the code and subsequently issue an SQL patch to "undo" the effects of this bug, so the patch would remove the appended `+` from all first names. In that case we would run that patch on RB to make sure our mock dataset also gets cleaned out of erroneous data

AlexandraLivadas:
This works for me!

christinerogers:
@h-karim note I'm removing the `Needs RB update` because this PR _is_ the needed update.

PR #6526, opened 2020-05-12
[issue_tracker] Update validation to allow NULL Site

### LINKED ISSUE #6518: [issue_tracker] Update validation to allow NULL Site
#6480 suggests this as a feature, which is broken

> Testing instructions (if applicable)
    Create an issue with no CenterID

In the Issue Tracker, when creating a new issue, the form can't be submitted with a NULL Site field.
Issues without a site assignment are viewable by users with restricted site permission, so we need to allow such issues to be created.

Updating the validation rules should solve the issue.

To reproduce:
 * Open the Issue Tracker
 * Click on New Issue
 * Submit the form with Site empty

Related to #6480

laemtl:
## Brief summary of changes
Since we want to allow new issues with NULL centerID to be created, the following changes have been made:

### Add/Edit issue
- a new Site option All Sites allows a user to save an issue with a NULL site
- The Comment History section displays **site to All sites** if an issue is edited/modified with the All Sites option.
- Alternatively, in the Comment History section, All sites can be replaced with a comma separated list of all CenterID (to discuss)

### Index
- An issue with a null centerID displays **All Sites** in the site column
- Issues can be filtered by sites: All Sites. 
- Alternatively, All sites can be replaced with a comma separated list of all CenterID (to discuss) in the datatable column, and sites can be made multi-selectable.

### Test plan
* Add modification to test instructions to handle the NULL CenterIDs
* Reintroduce Attachment testing instruction deleted by accident in #6636
 
### Linked Issues
* Resolves #6518

### DIFF
--- a/modules/issue_tracker/ajax/EditIssue.php
+++ b/modules/issue_tracker/ajax/EditIssue.php
@@ -583,6 +583,8 @@ function getIssueFields()
         $sites = $user->getStudySites();
     }
 
+    $sites[0] = "All Sites";
+
     //not yet ideal permissions
     $assignees = array();
     if ($user->hasPermission('access_all_profiles')) {
--- a/modules/issue_tracker/ajax/EditIssue.php
+++ b/modules/issue_tracker/ajax/EditIssue.php
@@ -101,11 +101,6 @@ function editIssue()
         $issueValues['candID'] = $validatedInput['candID'];
     }
 
-    // All Sites selected - Ignore value to store NULL in DB
-    if (isset($issueValues["centerID"]) && $issueValues["centerID"] === 'all') {
-        $issueValues["centerID"] = null;
-    }
-
     // Get changed values to save in history
     $historyValues = getChangedValues($issueValues, $issueID);
 
@@ -326,7 +321,10 @@ function getChangedValues($issueValues, $issueID)
     $changedValues = [];
     foreach ($issueValues as $key => $value) {
         // Only include fields that have changed
-        if ($issueValues[$key] != ($issueData[$key] ?? '') && !empty($value)) {
+        // centerID is allowed to be NULL
+        if ($issueValues[$key] != ($issueData[$key] ?? '')
+            && (!empty($value) || $key === 'centerID')
+        ) {
             $changedValues[$key] = $value;
         }
     }
@@ -349,9 +347,10 @@ function updateHistory($values, $issueID)
     $db   =& Database::singleton();
 
     foreach ($values as $key => $value) {
-        if (!empty($value)) {
+        // centerID is allowed to be NULL
+        if (!empty($value) || $key === 'centerID') {
             $changedValues = [
-                'newValue'     => $value,
+                'newValue'     => $value ?? '',
                 'fieldChanged' => $key,
                 'issueID'      => $issueID,
                 'addedBy'      => $user->getData('UserID'),
@@ -444,9 +443,8 @@ function getWatching($issueID)
  */
 function getSiteName($centerID): string
 {
-    if ($centerID == "all") {
-        $sites = Issue_Tracker::getSiteOptions(false, true);
-        return $sites["all"];
+    if ($centerID == null) {
+        return "All Sites";
     }
 
     $db =& Database::singleton();
@@ -598,10 +596,10 @@ function getIssueFields()
     $user  =& User::singleton();
     $sites = array();
 
-    //get field options
-    $sites = Issue_Tracker::getSiteOptions(false, true);
+    // get field options
+    $sites = Issue_Tracker::getSites(false, true);
 
-    //not yet ideal permissions
+    // not yet ideal permissions
     $assignees = array();
     if ($user->hasPermission('access_all_profiles')) {
         $assignee_expanded = $db->pselect(
--- a/modules/issue_tracker/ajax/EditIssue.php
+++ b/modules/issue_tracker/ajax/EditIssue.php
@@ -29,6 +29,7 @@
 require_once "Email.class.inc";
 
 use LORIS\issue_tracker\Provisioners\AttachmentProvisioner;
+use \LORIS\issue_tracker\Issue_Tracker;
 
 //TODO: or split it into two files... :P
 if ($_SERVER['REQUEST_METHOD'] === "GET") {
@@ -95,8 +96,14 @@ function editIssue()
     if (array_key_exists('sessionID', $validatedInput)) {
         $issueValues['sessionID'] = $validatedInput['sessionID'];
     }
+
     if (array_key_exists('candID', $validatedInput)) {
         $issueValues['candID'] = $validatedInput['candID'];
+    }
+
+    // All Sites selected - Ignore value to store NULL in DB
+    if (isset($issueValues["centerID"]) && $issueValues["centerID"] === 'all') {
+        $issueValues["centerID"] = null;
     }
 
     // Get changed values to save in history
@@ -429,6 +436,27 @@ function getWatching($issueID)
 }
 
 /**
+ * Returns the site name in the database corresponding to the centerID
+ *
+ * @param int $centerID - must match one from the psc table or 0
+ *
+ * @return string
+ */
+function getSiteName($centerID): string
+{
+    if ($centerID == "all") {
+        $sites = Issue_Tracker::getSiteOptions(false, true);
+        return $sites["all"];
+    }
+
+    $db =& Database::singleton();
+    return $db->pselectOne(
+        "SELECT Name FROM psc WHERE CenterID=:centerID",
+        array('centerID' => $centerID)
+    );
+}
+
+/**
  * Gets the changes to values, and the comments relevant to the given issue
  *
  * @param int $issueID the issueID
@@ -457,11 +485,7 @@ function getComments($issueID)
             $comment['newValue'] = $modules[$mid]->getLongName();
             continue;
         } else if ($comment['fieldChanged'] === 'centerID') {
-            $site = $db->pselectOne(
-                "SELECT Name FROM psc WHERE CenterID=:centerID",
-                array('centerID' => $comment['newValue'])
-            );
-            $comment['newValue']     = $site;
+            $comment['newValue']     = getSiteName($comment['newValue']);
             $comment['fieldChanged'] = 'site';
             continue;
         } else if ($comment['fieldChanged'] === 'candID') {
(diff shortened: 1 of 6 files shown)


driusan:
does this need a test plan update?

cmadjar:
@laemtl Could you update the test plan accordingly? Thank you!

laemtl:
Sure!

modules/issue_tracker/test/issue_tracker_test_plan.md
```diff
@@ -6,16 +6,18 @@ Issue Tracker Filter Form [Automation Testing]
...
-6. Check that all values are propagated and saved correctly.
-7. Add an attachment to the new issue and make sure that it is successfully uploaded.
-8. Check that watching logging is working - turn it off and on for your current user, and for other watchers on the issue
+3. Check that site can be null or populated by one of the dropdown values. 
+4. Should display message, and redirect after success. 
+5. Submit invalid and valid PSCID and visit label pairs. Error messages should respond accordingly. Not that you cannot submit PSCIDs from other sites unless you have access all profiles permission
+6. Submit just a visit label - this should give an error message.
+7. Check that all values are propagated and saved correctly.
+8. Add an attachment to the new issue and make sure that it is successfully uploaded.
+9. Check that watching logging is working - turn it off and on for your current user, and for other watchers on the issue
```
christinerogers, line 20:
```suggestion
5. Submit invalid and valid PSCID and visit label pairs. Error messages should display accordingly. 
6. A user should be able to submit a PSCID from other sites only if they have access_all_profiles permission.  
7.  Submit a PSCID and leave the Site blank _ .. this should/not work if... _
8.  Submit just a visit label - this should give an error message.
9. Check that all values are propagated and saved correctly.
10. Add an attachment to the new issue and make sure that it is successfully uploaded.
11. Check that watching logging is working - turn it off and on for your current user, and for other watchers on the issue.
```

christinerogers: Breaking down testing point 5.  
- [ ] To do : Point 7 - needs to be completed with what should be happening when a PSCID is entered with Null site.

laemtl: @christinerogers Are 7. and 3. redundant?

christinerogers: I see your point.  
3 could specify that PSCID is left blank.  
 7 tests the case where there is a PSCID, and Site doesn't match or is null.

modules/issue_tracker/test/issue_tracker_test_plan.md
```diff
@@ -6,16 +6,20 @@ Issue Tracker Filter Form [Automation Testing]
 5. Test that the watching checkbox works correctly (issues that your userID is watching in issues_watching table)
 6. Check that links to issues in table are correct.
 7. Check that table sorts and displays additional pages correctly 
+8. Check that a user who does not have access to all centers can see all issues with a site set to NULL.
```
driusan, line 9: this should say what permission it's referring to, since we have like 20 different access all site permissions..

modules/issue_tracker/test/issue_tracker_test_plan.md
```diff
@@ -6,16 +6,21 @@ Issue Tracker Filter Form [Automation Testing]
 5. Test that the watching checkbox works correctly (issues that your userID is watching in issues_watching table)
 6. Check that links to issues in table are correct.
 7. Check that table sorts and displays additional pages correctly 
+8. Check that a user who does not have access_all_profiles permission and limited sites access (see __user_psc_rel__ table for user/CenterID relationships) can see all issues with a site set to NULL.
```
christinerogers, line 9:
```suggestion
8. Check that a user who does not have access_all_profiles permission and belongs to only one site can see all issues with a site set to NULL.
```
@laemtl editing this because the User Accounts module is used to check and manage User-Site relationships via the front-end. (No need to check/update the `user_psc_rel` table manually in mysql)

christinerogers:
failing travis 😢 i know alex and shen are working on this

christinerogers:
> Perhaps to make it more intuitive the user, we should replace the empty option in the drop down to a string that says "All" to make it clear that by not selecting a site, they are in fact selecting all sites.

:+1:

This used to be the practice across all module filters before Reactification, or at least many filters.  @HenriRabalais Do you remember why we discontinued?

HenriRabalais:
@christinerogers iirc that was because it made the filter feel really busy with values that didn't need to be there. For filters, it's implied that if there's no value in the field, it's not filtering on that field (no value in filter field == no filtering for that field == all values for that field are presented). 

But for forms I don't think that same implication holds. It's expected that an empty field literally means that the value submitted will be NULL. If in this particular case for centerIds we are taking a NULL value in the back end to mean "All Sites", then I think the "All Sites" part is what should be communicated to the user on the front end. This way, the field would still be required, but they would have the option of selecting "All Sites".

christinerogers:
Thanks @HenriRabalais, good points - I wonder what the odds are we'll get a user confused because they entered an issue for "All Sites" but can't filter on that literal value.

driusan:
@HenriRabalais GitHub says you requested changes but it's not clear from the thread what changes you want her to make on this PR.. is there something outstanding?

Regarding the underlying issue, what if there was an `emptyValue=string` prop (or similar) so that forms could set the label they want for the empty value without affecting all of our filters?

driusan:
@HenriRabalais can you rereview/retest?

laemtl:
I noticed an issue if an issue is edited with All Sites. I am solving the issue and will submit a fix shortly.

modules/issue_tracker/php/issuerowprovisioner.class.inc
```diff
@@ -50,7 +50,7 @@ class IssueRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisioner
                     i.assignee,
                     i.status,
                     i.priority,
-                    i.centerID as centerId,
+                    COALESCE(i.centerID, 0) as centerId,
```
driusan, line 53: Why are you adding this coalesce and using the key "0" instead of handling null appropriately? This is making the code in getSiteOptions confusing. 0 is not a valid centerid.

laemtl: @driusan For the Site filter to work in the data table, I use 0 as a flag for All Sites. Replacing NULL by 0 is convenient as React can filter successfully.

HenriRabalais: Hmm, I think what may be better here (@driusan correct me if I'm wrong) is that if the value of `centerId` is NULL, then all centerIds currently in the psc table should be loaded into the issue row under the property `centerIds`. This would involve changing `int centerId` to `array centerIds` and loading only one value for most cases.

driusan: I don't have any preference on if it's done the way Henri suggests or by leaving it null and deferring the loading of sites to later, but using "0" as a magic number is confusing.. if anything needs to be magic, it should be `null` (or `''` in places that can't be null)..

christinerogers:
@h-karim can you help review this today if you have time?

christinerogers:
@driusan final re/review please - your changes were addressed

modules/issue_tracker/php/issuerowprovisioner.class.inc
```diff
@@ -99,7 +99,22 @@ class IssueRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisioner
...
+
+        if (isset($row['centerId'])) {
+            $row['centerId'] = [$row['centerId']];
+        } else {
+            $row['centerId'] = array_map(
+                "strval",
+                \User::singleton()->getCenterIDs()
+            );
+            array_push($row['centerId'], "all");
+        }
```
driusan, line 116: I don't understand what you're trying to do here.

laemtl: @driusan I added comments to clarify this section.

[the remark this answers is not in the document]
laemtl: That's something we can discuss. Currently, in the data table, an issue with a NULL centerId has the value 'All Sites' in the Site column. The option 'all' is added to be able to filter such Issues. 

I was hesitant to replace the label All Sites with a comma separated list of all centerIDs (and convert Sites to a multiselect field is also a good idea). From the comments I received I feel like this option will be less confusing.

modules/issue_tracker/php/issue_tracker.class.inc
```diff
@@ -67,6 +67,27 @@ class Issue_Tracker extends \NDB_Menu_Filter_Form
...
+        $user  = \User::singleton();
+        $sites = array();
+        if ($user->hasPermission('access_all_profiles')) {
+            // get the list of study sites - to be replaced by the Site object
+            $sites = \Utility::getSiteList();
+        } else {
+            // allow only to view own site data
+            $sites = $user->getStudySites();
+        }
+        $sites["all"] = "All Sites";
```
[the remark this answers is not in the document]
laemtl: In the comment history section, if Sites was updated to 'All Sites', this line allows to display the information properly.

Screenshot from 2020-06-12 14-58-00

modules/issue_tracker/ajax/EditIssue.php
```diff
@@ -102,6 +103,11 @@ function editIssue()
     // Get changed values to save in history
     $historyValues = getChangedValues($issueValues, $issueID);
 
+    // All Sites selected - Ignore value to store NULL in DB
+    if (isset($issueValues["centerID"]) && $issueValues["centerID"] == 0) {
+        $issueValues["centerID"] = null;
+    }
```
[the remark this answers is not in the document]
laemtl: This line should have been removed. Thanks!

modules/issue_tracker/php/issuerowprovisioner.class.inc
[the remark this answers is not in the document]
HenriRabalais: Yeah, I think my main concern is that storing the value `all` in an array that's supposed to hold centerIDs is a bit confusing to me. Personally, I think going the route of comma separated sites and a multiselect would be better. This also becomes extensible in case we move in the direction of having the issues be associated with a subset of sites (i.e. multiple sites but not ALL sites).

Nut then again I'm not sure if a user would prefer to know that the issue is for "All Sites" vs. "A list of all sites". I guess in that case in the format column table there could be a cross check between the centerIds array and a queried list of sites from the back end, and if they match, the table could display "All Sites" instead of the comma separated list of sites.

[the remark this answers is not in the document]
laemtl: Plenty of good suggestions here... Thanks @HenriRabalais for your feedback.

laemtl:
@HenriRabalais I updated the code following your suggestion.

modules/issue_tracker/jsx/issueTrackerIndex.js
```diff
@@ -95,7 +95,15 @@ class IssueTrackerIndex extends Component {
       };
       break;
     case 'Site':
-      result = <td>{this.state.data.fieldOptions.sites[cell]}</td>;
+      // if cell is an array containing all sites values
+      console.log(this.state.data.centerIDs);
+      console.log(cell);
```
HenriRabalais, line 100: looks like some forgotten console logs

ridz1208:
@driusan @HenriRabalais please re-review

PR #6531, opened 2020-05-13
[Loris] Horizontal scroll

### LINKED ISSUE #6530: [Loris] Horizontal scroll on webpages
All Loris pages have an horizontal scroll, which is not desirable.
The cause is the use of navbar-right outside a container/container-fluid element.

laemtl:
I added a container-fluid in navbar to fix the horizontal scrollbar issue in all pages of Loris.

Resolves #6530


christinerogers:
Bravo @laemtl 

@h-karim a good one to help test/review if you can...

PR #6541, opened 2020-05-14
[JSX - Filter] Enable Filter Presets

### LINKED ISSUE #6532: [Issue Tracker] Broken tab UI on page reload
In Issue Tracker v23, after a filter is selected, if the user reload the page, the filter has no more effect (all results are shown) and no tab is active anymore.

1. Go to Issue Tracker
2. Select a tab to filter the results (ex: Closed Issues)
3. Notice how the tab UI looks like: one tab is always active
4. Reload the page
5. No tab is active in the tab UI, and all issues are displayed.

Screenshot from 2020-05-13 16-16-17

### LINKED ISSUE #6533: [Issue Tracker] Filtered results and tab labels can disagree
In the Issue Tracker v23, it is possible to select the Closed Issues tab, then play with the filters to select issues with other status. The results will be updated under the Closed Issues tab, where active issues can now be listed.

1. Go to Issue Tracker
2. Select a tab to filter the results (ex: Closed Issues)
3. Change the status filter for New  
5. All new issues are displayed under the Closed Issues tab

### LINKED ISSUE #6700: [Issue Tracker] Incorrect behaviour of issue tracker dashboard widget
**Describe the bug**
I log in as a user that has access to the issue tracker module and that currently has issues assigned to him. On the dashboard, I click on the issue tracker widget in the My Task panel. I am redirected to the issue tracker module, but on the 'All Issues' tab.

**To Reproduce**
See above.

I was expecting to be redirected to the issue tracker module, tab 'My Issues' (even the url seems to indicate so).

HenriRabalais:
## Brief summary of changes

1. This PR introduces `Filter Presets` which auto-load filter parameters to the filter fields. They are presented as a selection drop down for easy access. 
2. The PR also ensures that these filter presets will update the URL query params so that when the page reloads, the same filter parameters are set. 
3. This PR also moves the `Clear Filter` button to be place next to the `Load Filter` dropdown for consistency across all modules.
4. This PR makes sure that the link between the Issues widget on the Dashboard and the Issue Track module sends the proper filter parameters to filter for the current User's issues.

In the near future, it would be great to add another button called `Save Filter` which allows the user to save the current filter parameters for future loading.

Screen Shot 2020-05-14 at 17 36 13

##### Filter Presets
1. Go to Issue Tracker.
2. Load a preset filter.
3. Ensure the Filter Fields update
4. Refresh page and make sure the same filter parameters are filled.

##### Widget Link
1. Go to the Dashboard.
2. Click on My Tasks > Your Assigned Issues
3. Make sure you are brought to the Issue Tracker Module with the Filter set to the current User's Issues.

Resolves #6533
Resolves #6532
Resolves #6700


christinerogers:
Cool idea, it does look great.  To keep consistency across LORIS modules, I think we'd rather keep a tabbed design, no? (this kind of menu appears so far only for DQT load queries i think) Or perhaps a change in the direction of the new Candidate module's tab/card GUI model might work.   cc Derek @3design

3design:
@christinerogers @HenriRabalais Thanks for the heads-up. I think this looks good all in all. 

Just to reiterate what this section is (so that I understand - please correct me if I'm wrong): This section is the filter, so anything inside the containing border should relate to filtering. The fields allow the user to define how to filter the results. The load filter/clear filter function will auto-fill/remove fields with preset values. 

Based on this, I think the load/clear is appropriate for what has already been designed. If we want to bring it more inline with the material design library that the new candidate module follows, this still works, though I would suggest a newer look to the entire card (which may be outside of the scope of this).  

I think the only additional suggestion I may have is somehow communicating to the user that the 'load filter' is loading a preset. Though it's wordy, maybe literally say 'load filter preset', or 'preset filters' or something.

3design:
As a follow up: There is a clear filter function that exists in other modules that set a precedent for the location and appearance of this function. I would suggest if moving forward that the load filter follow the same treatment as seen in the attached image

dqt

HenriRabalais:
@3design Thanks for your feedback!

I had originally had the dropdown labeled as `Load Filter Preset`, but I thought it might be a bit wordy so I reduced it. I'm open to keeping it longer if it makes more sense though.

I'm also open to a new look for the entire card. I know we've been talking about changing/moving the filter for a while, and I believe that would be a good thing to do. I've always found the the field lengths for each of the filter fields is unnecessarily long, and the whole thing feels like it's an inefficient use of space.

As for the follow-up, could you show me an example of the "clear filter function that exists in other modules"? The only thing I can find is the "Show Data" button on a few modules which is being deprecated as reactification occurs, since filters now respond dynamically to user input and don't need to be triggered. 

I'm personally a bit ambivalent to wether the buttons are on the top or the bottom, but in the current state, the "Clear Filters" button's placement is inconsistent throughout modules and it would be nice to tie it down to specific and consistent location. My argument for the top would be that it directly alters the fields bellow, and by placing them bellow, it feels like a reversal of information flow.

HenriRabalais:
@christinerogers I'm glad you like it!

I think the tabbed design makes a lot of sense when the information found in that tab is limitated to the parameters of the tab label. For example, if you click on the tab "Closed Issue", that tab should only show closed issues. Originally, when clicking on different tabs, a request to the server would be made to specifically query the data that matched that tab. However, with reactification, all data is now queried on module load and the tabs serve more as data filters.

With this, I think the tabs become a bit misleading since you are not limited to viewing "Closed Issues" on the "Closed Issues" Tab (which is the issue #6533 that @laemtl has reported). I mention some possible solutions on that issue, but I think moving in this direction to me seems the most conducive to customizability to user needs. We can always provide some default presets, but at the end of day if users can create, save and edit their own, they will be creating the most useful and inefficient presets for themselves. Under this design, I don't think tabs are the most appropriate choice because they aren't as flexible or extensible.

christinerogers:
I see what you're saying about tabs actually representing filter settings in some modules... good point @HenriRabalais

ridz1208:
@HenriRabalais i like the feature but that's not for 23 for sure, we are already on feature freeze. both issues it resolves are not critical at the moment. this should go to the master branch

3design:
@HenriRabalais Sorry for the delay. If you go to Access Profile and Imaging Browser, you can see the clear filter button. I think @zaliqarosli was the dev that restyled this last year, but then again this may be an old modification of the bootstrap btn class found in custom-css.css

HenriRabalais:
@3design Sorry, I think I had simply misread your original response! I had read "clear" as an adjective (e.g. obvious), rather than as a verb (e.g. remove). 

Yes, the "Clear Filter" button is in many modules. One of its current problems though is that its placement is very inconsistent within the filter card. I was hoping that by pairing it with a "Load Filter Preset" Button, we would be able to lock it to a specific part of the filter card, such as the top or bottom right. Although for aforementioned reasons, I do prefer the top right.

3design:
@HenriRabalais I agree with you. This is something that needs to be standardized across all LORIS modules sooner than later.

I think having them both stacked at the top also seems logical in that:

1. Title at top of card
2. Functions relating to the card top right
3. fields for the card in main body
4. submit button bottom right

This leads the user through a flow from top to bottom, left to right. Having a hanging clear bottom on the bottom right without a submit button may be misleading. Are there different search cards and filter cards in loris, or are they all filtered? In this question,  I mean search as a user entering in form fields and submitting that query for results, as opposed to filter where all results are shown and a filter changes things 'live'

ridz1208:
@henri please rebase this onto master branch

cmadjar:
@HenriRabalais Conflicts :)

driusan:
I like the way it's laid out in the original screenshot with the load/clear presets at the top better than the button which is a secret dropdown. How much work would be involved in making it consistent across modules if we went with your original design @HenriRabalais ?

HenriRabalais:
@driusan I think with this code change, the design will be consistent across all modules whose filters and datatables have been reactified using the `FilterableDatatable` component!

ridz1208:
@HenriRabalais rebased on master. development can move forward on this one. if discussion is still required please bring it up at the next lorismeeting

HenriRabalais:
@ridz1208 I think the PR is ready to be re-reviewed and tested!

christinerogers:
@h-karim would you have time to help test/review this, this week?

HenriRabalais:
@kongtiaowang I think this is having trouble with some automated testing

kongtiaowang:
reopen this PR to restart Travis

jsx/Filter.js
```diff
@@ -109,6 +101,38 @@ class Filter extends Component {
...
+          </li>
+        );
+      };
+    };
+
+    const filterActions = (
+      <ul className='nav nav-tabs navbar-right' style={{borderBottom: 'none'}}>
+        {filterPresets()}
+        <li>
+          <a role='button' onClick={this.props.clearFilter}>Clear Filter</a>
```
kongtiaowang, line 131: @HenriRabalais Could u add a name = 'reset' inside this link element. It will be convient to do automation test.

HenriRabalais:
@kongtiaowang done!

HenriRabalais:
@driusan could you take a look at this?

driusan:
@HenriRabalais it's going to the wrong branch

HenriRabalais:
@driusan fixed!

Issue #6539, opened 2020-05-14
Unable to configure user and admin on new install


pn2200:
**Describe the bug**
Cannot create non-administrator user or LORIS admin account on installdb.php page after a brand new installation.

**To Reproduce**
I followed the [installation guide](https://acesloris.readthedocs.io/en/latest/wiki/00%20-%20SERVER%20INSTALL%20AND%20CONFIGURATION/01%20-%20LORIS%20Install/Ubuntu/). I installed on Ubuntu 20.04. After the database setup step, the instructions are: 

> Click submit, and on the next screen that is presented, follow instructions to enter the username and password of your LORIS database user and front-end admin user.

When I do that, the page displays an "All entries must be provided" error message, even though all entries are provided. See screenshot:
<img width="629" alt="Screen Shot 2020-05-14 at 11 56 00 AM" src="https://user-images.githubusercontent.com/489253/81974849-fab81a80-95da-11ea-84ac-46776111b744.png">

There are no errors in the loris-error.log. There is no config.xml file in /var/www/loris/project.

**What did you expect to happen?**
I expected the users to be created, the config.xml file to be created, and I would expect to be able to go to the next configuration step.

**Browser Environment (please complete the following information):**
- OS: Ubuntu 20.04LTS
- Browser: Firefox 75.0
- Version: 22.0.1

**Server Environment (if known):**
- LORIS Version: 22.0.1
- Linux distribution and Version: Ubuntu 20.04LTS
- MySQL/MariaDB Version: MySQL 8.0

lingma:
You can create the user `rethinkuser` with your password and check the "Already created" to retry.

lingma:
There is a note about possible problem for MySQL 8.0 in the Readme.md: "If you use MySQL 8, please read this link..." Please refer to the Readme.md to check.

pn2200:
Unfortunately, checking the box "Already Created" and retrying does not change the result. The page still displays the error "All entries must be provided", the config.xml file is not created and there are no errors in the loris-error.log file.

Perhaps I am not understanding your suggestion. Are you suggesting that I manually create the user in mysql first?

[...]

PR #6511, opened 2020-05-08
[Documentation] Troubleshooting installation document created

AlexandraLivadas:
## Brief summary of changes

A general troubleshooting document is being created to supplement installation instructions for the 23.0-release. It will cover how to solve common errors faced when installing LORIS (on any OS)/

#### Link(s) to related issue(s)

* Resolves #5920 
* Resolves PR #6555 
* Related to PR #6510, as part of the new install documents for the 23 release

PR #6510, opened 2020-05-08
[InstallDocs] Install markdowns added to 23.0-release

AlexandraLivadas:
## Brief summary of changes

- Install instructions for CentOS and Ubuntu installation added to the `docs/wiki` for the 23.0-release. 

- php7.3 curl dependency added to Ubuntu installation instructions

- Markdown files under `docs/wiki` were changed (spaces replaced with underscores). All references to these files changed as well. 

#### Links to related issues

* Resolves #6475
* Resolves #6543
* Resolves #6415
* Related to #6492, #6181

PR #6555, opened 2020-05-20
[23.0 Documentation] Correction of PHP requirement with MySQL 8.0

lingma:
## Brief summary of changes

MySQL 8.0 with PHP 7.4 has no concern as with PHP 7.3

1. If you want to try, test with a MySQL 8.0 over PHP 7.4 vs 7.3
Or refer to the issue #6539

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ sudo service apache2 reload
     
 6. Go to http://localhost/installdb.php and follow the instructions to finalize LORIS installation, then restart apache.
 
-If you use MySQL 8, please read [this link](https://www.php.net/manual/en/mysqli.requirements.php) and also [this](https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password-compatible-connectors).
+If you use MySQL 8 and PHP 7.3 (instead of PHP 7.4), please read [this link](https://www.php.net/manual/en/mysqli.requirements.php) and also [this](https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password-compatible-connectors).
 
 7. Follow the [Setup Guide in the LORIS Wiki](https://github.com/aces/Loris/wiki/Setup) to complete your post-installation setup and configuration, and for more documentation.
 


lingma:
Something went wrong with Travis. Because this PR only touches one line in `Readme.md`, I don't think that it is necessary to verify Travis status.

lingma:
The Travis test has passed.

lingma:
The light is not green so let Travis do it again the same work

lingma:
Removed Travis fix since the fix is better done by @kongtiaowang in #6563

README.md
```diff
@@ -92,7 +92,7 @@ sudo service apache2 reload
     
 6. Go to http://localhost/installdb.php and follow the instructions to finalize LORIS installation, then restart apache.
 
-If you use MySQL 8, please read [this link](https://www.php.net/manual/en/mysqli.requirements.php) and also [this](https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password-compatible-connectors).
+If you use MySQL 8 and PHP 7.3 (instead of PHP 7.4), please read [this link](https://www.php.net/manual/en/mysqli.requirements.php) and also [this](https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password-compatible-connectors).
```
johnsaigle, line 95: @AlexandraLivadas I think this note is worth adding to #6511. If we do so, we can probably remove this line from the main README.

lingma: Yes, this line looks strange in the README.md, the note should have a better place, and the problem of MySQL 8.0 is much less a problem now as before vs PHP7.3. @johnsaigle

christinerogers:
Looks like we're agreed lingma @johnsaigle that @AlexandraLivadas can wrap this into her #6511 PR.    I'm assigning her to this PR.  
Alex, please add a reference to this PR somewhere in 6511, and we'll close this PR.

README.md
AlexandraLivadas: Should I add this to #6510, which is the Ubuntu install markdown? Or do you think it should also be added to the troubleshooting guide (#6511) @johnsaigle ?

johnsaigle: Yeah I think the troubleshooting guide is better. Since this isn't specific to Ubuntu as far as I can tell, it should be in the cross-platform document.

PR #5541, opened 2019-11-05
fix UI for statistics when on mobile mode

cmadjar:
## Brief summary of changes

Fixes minor UI issue in dropdown in <900px wide screens.

#### Links to related tickets (GitHub, Redmine, ...)

* #5460

PR #6602, opened 2020-05-25
[UI] DQT dropdown visibility issue

### LINKED ISSUE #6584: [DQT] Field - Some text barely visible
In DQT > Define field tab, if you click on a field, the text switches to white on a light background.
If you click on the instrument dropdown, the background switches to a dark color (Chrome only).

**To reproduce**
    Go to Reports > Data Query Tool
    Select the Define field tab
    Select an instrument
    Select a field

Screenshot from 2020-05-22 11-12-23

Screenshot from 2020-05-22 11-37-22

laemtl:
I updated a css rule to improve the visibility of some dropdown elements

* Resolves #6584

### DIFF
--- a/htdocs/bootstrap/css/custom-css.css
+++ b/htdocs/bootstrap/css/custom-css.css
@@ -808,7 +808,7 @@ a:hover {
 .list-group-item.active,
 .list-group-item.active:hover {
     z-index: 2;
-    color: white;
+    color: black;
     background-color: #E4EBF2;
     border-color: #E4EBF2;
 }


driusan:
Do you have before/after screenshots? @zaliqarosli can you review since you're the one who made the colour scheme?

zaliqarosli:
this issue was a side-effect of a PR meant to target UI inconsistency within the Statistics module https://github.com/aces/Loris/pull/5541. The change affected the DQT without us realizing it.

This solution is sufficient and doesn't seem to break anything else that uses classes  .list-group-item and .active

Issue #5214, opened 2019-09-10
[login] module rewrite with react


maltheism:
Redmine: https://redmine.cbrain.mcgill.ca/issues/16050

**Description:**
Make the login form use react instead of smarty.

#4395

PR #6611, opened 2020-05-26
[Login] Password reset password mismatch

### LINKED ISSUE #6606: [Login] Reset password - validation does not work
On the reset password screen, the password validation is not working. The confirm password box can be set to any value without preventing the first password to be saved in the database.

**To reproduce**
Change Password_expiry for a given user in the database for a date in the past
Login with that user
Type a new password in the password field, and either nothing/a different password in the confirm box.

laemtl:
In login/password-expiry/, I added a validation to display an error if the 2 passwords do not match.
This validation may have been removed [here](https://github.com/aces/Loris/commit/9c8a24bb395444a118c06b579cc4e44aeeceeb1d#diff-1dc7c9358a505a311690636f4e1bf51cL124)

**To test**
In the users table in the database, update the Password_expiry cell for a user you are testing to be a date in the past.
Login as that user. You should see an update password page.

* Related to #5417
* Resolves #6606

### DIFF
--- a/modules/login/php/passwordexpiry.class.inc
+++ b/modules/login/php/passwordexpiry.class.inc
@@ -88,6 +88,7 @@ class PasswordExpiry extends \NDB_Form
     {
         $this->_user = \User::factory($this->_username);
         $plaintext   = htmlspecialchars_decode($values['password']);
+        $confirm     = htmlspecialchars_decode($values['confirm']);
 
         try {
             new \Password($plaintext);
@@ -99,6 +100,13 @@ class PasswordExpiry extends \NDB_Form
         if (!$this->_user->isPasswordDifferent($plaintext)) {
             $this->tpl_data['error_message'] = 'You cannot keep the same password';
             return array('password' => 'You cannot keep the same password');
+        }
+
+        // Ensure that the password and confirm password fields match.
+        // TODO This validation should be done on the front-end too.
+        if ($plaintext !== $confirm) {
+            $this->tpl_data['error_message'] = 'The passwords do not match';
+            return array('password' => 'The passwords do not match');
         }
 
         return array();


laemtl:
The description of _validate says
`* _validate validates the data that was submitted by the user. Validate
 * for password-reset only checks that a username was submitted, not that
 * it's valid, in order to prevent brute force phishing attacks.`

Should we update that comment or perform the validation elsewhere? I also wonder if common password validations should be ideally centralized somewhere, for example in Password.

johnsaigle:
I think that comment only applies to password reset, which is a bit different (though related) to password expiry. That comment basically just means that the page will always say "An password reset was sent to your email" even if the user account doesn't exist. The server will validate on its own whether an account exists.

That functionality has to do with checking whether a User account exists rather than validating an input to be used as a Password. For that reason it shouldn't be in the Password class. 

> common password validations should be ideally centralized somewhere, for example in Password.

Exactly, the Password class does do this. However, it is designed to represent only the idea of a valid Password based on password strength. It doesn't check "state": the status of the database or anything to do with Users existing or not. This is done by User or the modules. 

For example, the Password class checks that an input is complex enough to be used as a Password, but it is the User class that would determine whether a user is trying to use their email as a Password, or if the Password has not been updated (i.e. a user is trying to use the password they already have as a "new" password). 

With this in mind, I do think there is value in having a reusable front-end component that does validation on the user input. For example:

* checking that a "confirm password" box matches what's entered for the password
* having a password strength estimator that matches the back-end code (#5761)
* as well as basic things like making sure the input is not empty.

#5214 has been issued to refactor the `login` module to use React but development and review has been stalled for a while. Until then that module is using Smarty templates that aren't super great for adding a component like that.

ridz1208:
@johnsaigle could you take it out for a test run ?

johnsaigle:
I'm still very occupied with UK biobank stuff but I'll see what I can do.

PR #6624, opened 2020-05-27
[survey_accounts] Error message when survey added on different project

### LINKED ISSUE #6582: [survey_accounts] User can create survey for Project they are not involved in
**Describe the bug**
A user account is able to create a survey for a candidate on a project that they are not affiliated with and should therefore not have access to. The survey is properly created (visible on Admin account survey list) and can be completed and submitted. The newly created survey does not appear on the list of surveys for the user who is not affiliated with the project (as expected) even though they created the survey. **This issue occurs on both Firefox and Chrome.**

**To Reproduce**
1. Create a user account on https://test-loris-dev.loris.ca/ with only one affiliated project, give the account "User management" permissions.
2. Go to this user account and view the Survey Accounts page
3. Click on "Add Survey" and create a survey for a candidate in a Project that the user account is not part of.
4. See that no error appears when you create the survey, but that the survey does not appear on the survey list for the user. Log into admin account and see that the new survey appears on the Survey Accounts list. 

Users without Project permissions should not be able to create a survey for this Project.

AlexandraLivadas:
## Brief summary of changes

If a user attempts to create a survey for a candidate on a project they are not involved in, an error message will appear and the survey will not be created.

#### Testing instructions
1. Create a user account who isn't affiliated with at least one project and with "User Management" permissions
2. Log-in as this user and go to the survey accounts page.
3. Try to add a survey using a candidate who is on a project that the user is not on. An error message should appear.
4. Check that the survey was not added to the list. 

#### Link to related issue

* Resolves #6582

### DIFF
--- a/modules/survey_accounts/php/addsurvey.class.inc
+++ b/modules/survey_accounts/php/addsurvey.class.inc
@@ -77,36 +77,6 @@ class AddSurvey extends \NDB_Form
               array('CandID' => $error);
         }
 
-        $projectID    = $db->pselectOne(
-            "SELECT RegistrationProjectID FROM candidate
-            WHERE PSCID=:v_PSCID
-            AND CandID=:v_CandID",
-            array(
-                'v_PSCID'  => $values['PSCID'],
-                'v_CandID' => $values['CandID'],
-            )
-        );
-        $user         = \User::singleton();
-        $errorProject = "You do not have access to this candidate's project";
-        $user_list_of_projects = $user->getProjectIDs();
-        if (count($user_list_of_projects) == 1) {
-            //if there is only one project, autoselect first project from array of 1
-            $userProjectID = array_pop($user_list_of_projects);
-            if ($projectID != $userProjectID) {
-                return array('Project' => $errorProject);
-            }
-        } else if (count($user_list_of_projects) > 1) {
-            $projectMatch = 0;
-            foreach ($user_list_of_projects as &$userProjectID) {
-                if ($projectID == $userProjectID) {
-                    $projectMatch = 1;
-                }
-            }
-            if ($projectMatch == 0) {
-                return array('Project' => $errorProject);
-            }
-        }
-
         $numSessions = $db->pselectOne(
             "SELECT COUNT(*) FROM session 
             WHERE CandID=:v_CandID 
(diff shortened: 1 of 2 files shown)


modules/survey_accounts/php/addsurvey.class.inc
```diff
@@ -150,6 +150,36 @@ class AddSurvey extends \NDB_Form
...
+                if ($projectID == $userProjectID) {
+                    $projectMatch = 1;
+                }
+            }
+            if ($projectMatch == 0) {
+                return array('Project' => $errorProject);
+            }
+        }
+
         return array();
```
johnsaigle, line 183: I think all of this code can be simplified to

```php
\NDB_Factory::singleton()->user()->hasProject($projectID)) ?
    return array('Project' => "You do not have access to this candidate's project") :
    return array();
```

AlexandraLivadas: Oh yes this would work. I wasn't aware of the hasProject method. Thanks!

christinerogers:
@laemtl Can you help us out by pulling this and testing on your sandbox?

modules/survey_accounts/php/addsurvey.class.inc
```diff
@@ -150,7 +150,19 @@ class AddSurvey extends \NDB_Form
             }
         }
 
-        return array();
+        $projectID = $db->pselectOne(
+            "SELECT RegistrationProjectID FROM candidate
```
driusan, line 154: This isn't the correct project ID to check, it's just where the candidate was originally created. The correct ProjectID is the session's ProjectID.

johnsaigle:
I agree that the permission error code should appear first.

modules/survey_accounts/php/addsurvey.class.inc
```diff
@@ -77,6 +77,19 @@ class AddSurvey extends \NDB_Form
...
         }
 
+        $projectID = $db->pselect(
+            "SELECT ProjectID FROM session
+            WHERE CandID=:v_CandID",
+            array(
+                'v_CandID' => $values['CandID']
+            )
+        );
+        if (!\NDB_Factory::singleton()->user()->hasProject(reset($projectID[0]))) {
```
driusan, line 87: This is just the first session of the candidate, not the one submitted from the addsurvey page. Isn't there a visit submitted to add the survey to?

johnsaigle, line 87: Can you explain your commit message "Special visit label case handled"? I'm not sure why the code was changed here.

driusan: Candidates aren't only related to one project, candidates can have sessions at different projects. That's why the field is called "RegistrationProjectID" and not "ProjectID".

AlexandraLivadas: Ah ok. In this case, I will add the visit label to the query, so that it checks for the project related specifically to this session. Thank you for the clarification!

AlexandraLivadas:
@johnsaigle @driusan This latest commit includes the 2 following changes: 
1. The projectID is queried using both the candidate ID and visit label for the session.
2. The project permission is checked before "Instrument Medical History already exists for given candidate for visit V1" as requested by @laemtl , but after checking that the candidate and session exist (since this was causing errors for some edge use-cases).

Issue #6631, opened 2020-05-27
[document_repo] Document Repo module does not describe or test widgets


HenriRabalais:
The Document Repo module registers widgets to the dashboard, but the README does not describe them and the test plan does not test them.

Related to #6467

Issue #6630, opened 2020-05-27
[bvl_feedback] Behaviour Feedback module does not describe or test widgets


HenriRabalais:
The bvl_feedback module registers widgets to the dashboard, but the README does not describe them and the test plan does not test them.

Related to #6467

Issue #6438, opened 2020-04-27
[issue_tracker] Widgets not described in test plan


driusan:
The issue tracker registers widgets on the dashboard and candidate_profile dashboard, but the test plan does not describe how to test the when the module is active.

ridz1208:
@driusan @cmadjar why wasnt this assigned to anyone ?

ridz1208:
@maltheism any progress on this one ?

Issue #6437, opened 2020-04-27
[instruments] Widgets are not described in README or test plan


driusan:
The instruments module registers widgets to the dashboard and candidate dashboard, but the test plan does not describe them and the README doesn't mention them as interactions.

PR #6632, opened 2020-05-27
[dashboard] Update Dashboard Testplan and Readme

### LINKED ISSUE #6467: [Dashboard] Update Readme post-refactor for 23 release
Module readme needs an update since it was refactored in jan 2020 #5896

Several issues linked below have been created to update the widget modules' testplans to add test cases for the their respective widgets.

After all issues are resolved and PRs merged, the dashboard readme and testplan should undergo a sanity check to make sure it is up to date with the new widget structures.

The below text could be handy -- This change was partially described by Dave's [comment quoted below](https://github.com/aces/Loris/pull/6464#discussion_r418694673) on #6464

> The dashboard test plan was written before the module used any widgets, and no longer houses the code being tested. There are many things you can do now which affect what shows up in the dashboard without the dashboard knowing anything about it.. write a new module for a project which has widgets, disable/enable an existing module in the module manager, override a (non-dashboard) module which is one of the ones that provides widgets.. it's also easier to have someone test the related functionality to a module in the dashboard when testing that module than have someone setup and test (effectively) every module in LORIS at the same time for the "dashboard" test.
> 
> The tickets came up when I made tickets for the `candidate_profile` widgets that are missing documentation since the `dashboard` and `candidate_profile` use the same `getWidgets()` function.

_Originally posted by @driusan in https://github.com/aces/Loris/pull/6464_

HenriRabalais:
## Brief summary of changes
This PR updates the Dashboard Testplan to reflect the fact that the widgets found on the Dashboard should be tested on the Modules from which they are provisioned. 

I have removed everything from the Testplan that I  believe to be under the purview of another module. I may have messed up and removed too many, so please correct me if I have!

I also added a small section to the README by @ridz1208's request which mentions the interaction with other modules through the widget. Please let me know if the text does not include enough information, or is redundant to any of the other README sections.

* Resolves #6467


modules/dashboard/test/TestPlan.md
```diff
@@ -1,56 +1,4 @@
 1. Log in. Note the time. Log out and log back in after 2 minutes. Check that welcome panel info is correct.
```
ridz1208, line 1: @HenriRabalais the deletions seem to make sens as they all relate to widgets now included in other modules.

However, I'm wondering if we should add a line item to test the "loading a widget" functionality... technically it's already being tested by the modules themselves...

@driusan thoughts ?

modules/dashboard/README.md
```diff
@@ -16,6 +16,10 @@ The Dashboard provides entrypoints to various workflows, but does not provide an
 
 The Study Progression chart views at bottom may or may not count `excluded` candidates, and this should be clearly verified and documented/presented for the use of your LORIS users. 
 
+## Interactions
```
ridz1208, line 19:
```suggestion
## Interactions with LORIS
```

modules/dashboard/README.md
```diff
@@ -16,6 +16,10 @@ The Dashboard provides entrypoints to various workflows, but does not provide an
...
+
+The dashboard calls upon other modules to supply data to the various "Widget" panels that can be viewed in the dashboard. Without access permission to these modules, a user will not be able to view the respective Widget on the Dashboard. 
+
```
ridz1208, line 22: please move the section to the bottom of the doc.

otherwise looks good

modules/dashboard/test/TestPlan.md
driusan: It think it would be a good idea to add tests for loading widget types/permissions. You can look at the `candidate_profile` test as an example.

driusan:
I believe this is Blocked by 
#6438
#6437
#6631
#6630

according to @HenriRabalais's comment in #6467. Please correct me if I'm wrong.

modules/dashboard/test/TestPlan.md
driusan: (I don't mind if that's not done in this PR since, as you say, the individual widget tests in the other PRs are implicitly testing that right now..)

HenriRabalais:
PR is no longer blocked and is pointed to the proper branch.

PR #6633, opened 2020-05-27
[Genomic Browser] Subproject Column label

### LINKED ISSUE #6295: [Genomic Browser] Subproject Column displayed as Integer instead of Label
**Describe the bug**
A. In the Profiles tab of the Genomic Browser, when downloading the Table as a CSV, the SubProject is displayed as an integer instead of a string label.

B. In the SNP tab of the Genomic Browser, the SubProject is displayed as an integer in the table instead of a string label

Test Plan Reference: https://github.com/aces/Loris/blob/master/modules/genomic_browser/test/TestPlan.md#data-download

**To Reproduce**
A.
1. Go to Profiles Tab of Genomic Browser
2. Click on Download as CSV.
3. Look at Subproject column of downloaded CSV.

B.
1. Go to SNP Tab of Genomic Browser
2. Look at Subproject column datatable.

A.
image

B.
image

laemtl:
In Genomic Browser > Profile and SNP screens, I added a parsing logic on the csv colums to allow special treatment of data (inspired by what is already implemented on the data table with [profileColumnFormatters](https://github.com/aces/Loris/blob/23.0-release/modules/genomic_browser/jsx/profileColumnFormatter.js)) and applied a Formatter on SNP > Subproject int he data table.

[Testing card](https://github.com/aces/Loris/projects/19#card-21296727)

* Resolves #6295

### DIFF
--- a/jsx/StaticDataTable.js
+++ b/jsx/StaticDataTable.js
@@ -35,7 +35,7 @@ class StaticDataTable extends Component {
     this.downloadCSV = this.downloadCSV.bind(this);
     this.countFilteredRows = this.countFilteredRows.bind(this);
     this.toCamelCase = this.toCamelCase.bind(this);
-    this.getSortedRows = this.getSortedRows.bind(this);//
+    this.getSortedRows = this.getSortedRows.bind(this);
     this.hasFilterKeyword = this.hasFilterKeyword.bind(this);
   }
 
@@ -131,9 +131,14 @@ class StaticDataTable extends Component {
     });
   }
 
-  downloadCSV(csvData) {
-    let csvworker = new Worker(loris.BaseURL + '/js/workers/savecsv.js');
+  downloadCSV() {
+    // Include only filtered data if filters were applied
+    let csvData = this.props.Data;
+    if (this.props.Filter && this.props.FilteredData.length > 0) {
+      csvData = this.props.FilteredData;
+    }
 
+    let csvworker = new Worker(loris.BaseURL + '/js/workers/savecsv.js');
     csvworker.addEventListener('message', function(e) {
       let dataURL;
       let dataDate;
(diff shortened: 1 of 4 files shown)


christinerogers:
@xlecours or @spell00 do you have time to review this one?

xlecours:
@laemtl Did you try to JOIN on the subproject table in the module query to send the subproject name instead of the subproject_id to the frontend?

laemtl:
Great suggestion @xlecours , I pushed a new version!

xlecours:
Did not test but the code looks good to me.

ridz1208:
@spell00 when something passes testing please add the "Passed Manual Test" Label

HenriRabalais:
There is already an implementation for resolving this issue by using the `getMappedCell()` function from the `FilterableDatatable` component, towards which we are trying to migrate all Filterable Datatable implementations. However, this makes use of the `FilterableDatatable` component, which the Genomic Browser is not currently using.

I haven't tested this yet, but if it works the code looks fine. But I'm also wondering if we should take the opportunity to transition the Genomic Browser over to the new `FilterableDatatable` component system since it already has a solution to this issue.

laemtl:
@HenriRabalais, is getMappedCell() solving specifically the csv formatting issue?

HenriRabalais:
@laemtl It was first designed to address that issue, yes. But it also resulted in addressing a larger underlying issue which was that `getFormattedCell()` was taking on the roll of both mapping and formatting data. This meant that the data was being filtered based on the mapped and formatted data, but only being downloaded as unmapped and unformatted. 

By disentangling those processes, the data can be a) filtered on unmapped data, b) both mapped and formatted to be displayed to the user and c) simply mapped when downloaded via csv.

HenriRabalais:
@laemtl I just realized that what I'm suggesting wouldn't be able to be properly implemented until #5835 is merged.

I'll take a closer look at the code and test the PR so that we can get this through as is, and hopefully get it changed over in the future.

jsx/StaticDataTable.js
```diff
@@ -432,7 +440,7 @@ class StaticDataTable extends Component {
 
         if (this.hasFilterKeyword(this.props.Headers[j], data)) {
           filterMatchCount++;
-          filteredData.push(this.props.Data[index[i].RowIdx]);
+          this.props.FilteredData.push(this.props.Data[index[i].RowIdx]);
```
HenriRabalais, line 443: Props should be read-only. Components should not be altering their own props.

laemtl: @HenriRabalais The filter works fine, and the query received the correct params (AND cohort.SubprojectID LIKE CONCAT('%', :v_cohortsubprojectid, '%')) without changes needed.

laemtl:
@HenriRabalais The filter works fine, and the query received the correct params (AND cohort.SubprojectID LIKE CONCAT('%', :v_cohortsubprojectid, '%')) without changes needed.

HenriRabalais:
@laemtl okay! sounds good! I'll test and then it should be good to go!

PR #6634, opened 2020-05-28
[imaging browser] Display different scan type columns

### LINKED ISSUE #6619: [imaging browser] Columns not appearing when adding some scan types
**Additional context**
This fails step 7 of the "main page" section of the test plan.

**Describe the bug**
When adding a scan type via the Config module, new columns don't appear on the front page of the imaging browser for the particular scan

**To Reproduce**
1. Go to Admin>Configuration>Imaging Modules
2. Next to "Imaging Browser Tabulated Scan Types" add a "flair" field
3. Click on submit
4. Navigate to the imaging browser front page
5. observe no new column appears
A new table column with the name "flair". 

Field values tested that reproduce this bug are (list is not exhaustive): 
- flair
- pd
- mrs
- dti
- -defaced (NOT "t1-defaced" and "t2-defaced")
- NA
- pve

- OS: Ubuntu 18.04 LTS
- Browser:
    firefox 76.0.1
    Chrome 83.0.4103.61

Tested using the loris testing VM.

h-karim:
## Brief summary of changes

- Currently, the function `Utilities::getScanTypeList` has an inner join `JOIN files f ON (f.AcquisitionProtocolID=mri.ID)` which excludes scan types that are NOT of the following: fMRI, t1, t2, dwi25, dwi65 and fieldmapDWI, from being returned through the function because the files.AcquisitionProtocolID only has the scan type IDs for the aforementioned scan types. This results in a lot (the majority) of scan types not appearing when attempting to add them through Configuration>Imaging Modules.
- I added an optional boolean paramater, `join_files` to `Utilities::getScanTypeList` that allows disabling its inner join, which results in all the scan types to be returned. 
- passing `false` to the function call in the imaging module class gets us all the scan types, allowing all the scan types to be displayed on the front page.

- `getScanTypeList` is only called twice throughout the code base, the proposed modifications do not change its default behaviour, so outside of the function call that is modified through this PR, there should not be any side effects observed with the one other call.

1. Navigate to  Configuration>Imaging Modules>
2. Add different scan type fields
3. Navigate to the imaging browser front page
4. observe the added fields 

* Resolves #6619

### DIFF
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -1011,7 +1011,7 @@ class Utility
     {
         $DB    = \Database::singleton();
         $query = "SELECT ID, Scan_type
-             FROM mri_scan_type mri\n";
+             FROM mri_scan_type mri ";
         if ($join_files) {
             $query .= "JOIN files f ON (f.AcquisitionProtocolID=mri.ID)";
         }
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -236,7 +236,7 @@ class Utility
         if (!is_null($projectID)) {
             $subprojects = $DB->pselect(
                 "SELECT * FROM subproject
-                JOIN project_subproject_rel USING (SubprojectID) 
+                JOIN project_subproject_rel USING (SubprojectID)
                 WHERE ProjectID=:pID",
                 array('pID' => $projectID)
             );
@@ -575,7 +575,7 @@ class Utility
 
         $instruments   = array();
         $instruments_q = $DB->pselect(
-            "SELECT Test_name,Full_name FROM test_names WHERE IsDirectEntry=true 
+            "SELECT Test_name,Full_name FROM test_names WHERE IsDirectEntry=true
              ORDER BY Full_name",
             array()
         );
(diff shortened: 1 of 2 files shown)


php/libraries/Utility.class.inc
```diff
@@ -1001,20 +1001,21 @@ class Utility
...
         $DB = \Database::singleton();
-
-        $scan_types_DB = $DB->pselect(
-            "SELECT ID, Scan_type 
-             FROM mri_scan_type mri
-                JOIN files f ON (f.AcquisitionProtocolID=mri.ID)",
-            array()
-        );
+        $query = "SELECT ID, Scan_type
+             FROM mri_scan_type mri\n";
```
christinerogers, line 1014: why the `\n` ?

php/libraries/Utility.class.inc
```diff
@@ -1001,20 +1001,21 @@ class Utility
...
+        if ($join_files) {
+            $query .= "JOIN files f ON (f.AcquisitionProtocolID=mri.ID)";
```
christinerogers, line 1016: Karim can you explain in your PR description what you're solving, why you're solving it like this, and what this query changes?   Like -- adding a parameter to join with the files table to return... (and _why_) 
This is important to state clearly upfront as there are potential side effects across the codebase where this function is called.

h-karim: Sure thing, I added more details to the PR description, let me know if there's anything more missing.

php/libraries/Utility.class.inc
h-karim: It's to preserve space within the query in case the join statement is concatenated. I will replace it with a white space character.

cmadjar:
@h-karim @christinerogers Sorry I did not see that before. 

The inner join is needed for the imaging browser and configuration I think. 

The default list of scan types we give in LORIS are rarely used by people and often left in the mri_scan_types table even though they are not being used by a project. This ends up in displaying too many scan types that are not necessarily relevant to a study. 

So the real list of scan types is determined when you join the mri_scan_type and the files tables => this will only list the scan types for which there are entries in the files table. This is the intended behaviour so I would tend to close this PR.

Unless there is argument as to why the JOIN should be removed that I don't understand?

h-karim:
@cmadjar @christinerogers 
>The default list of scan types we give in LORIS are rarely used by people and often left in the mri_scan_types table even though they are not being used by a project. This ends up in displaying too many scan types that are not necessarily relevant to a study.
This makes sense. The issue was proposed because the drop down menu in the config module displayed all of the scan types, and the test plan indicated that adding a scan type should add a column: 
screen shot of dropdown menu .

In this case, this issue should be closed and another one should be raised about the selection menu in the config module for the imaging browser.

PR #6639, opened 2020-05-28
[electrophysiology browser] Check if user is affiliated with site when opening individual scan sessions

### LINKED ISSUE #6557: [Electrophysiology_Browser] 500 error on view session when user only has view own site permission
**Context**
Fails step A2 of the module test plan.

**Describe the bug**
When the user is granted the _View own site Electrophysiology Browser pages_ only, navigating to an individual session produces a 500 http error.

**To Reproduce**
1. Create a user with the following permission: 
> View own site Electrophysiology Browser pages
2. Make sure user is assigned a site and project which has candidate data. For the test VM specifically user was assigned site: Ottawa and project: pumpernickel.
3. Navigate to the electrophysiology browser front page 
3. Click on either "raw" or "all types" link for one of the candidates
5. Observe blank page (firefox) or 500 http error (chrome).

For the candidate session data to load and display the relevant information when the user tries to view an individual session. 

- firefox 76.0.1 
- Chrome 83.0.4103.61

This was observed using the testing VM for loris

reproducing the bug on my local sandbox gives the following error log:

>PHP Fatal error:  Uncaught Error: Call to a member function getCenterID() on null in /var/www/loris/modules/electrophysiology_browser/php/sessions.class.inc:51\nStack trace:\n#0 /var/www/loris/php/libraries/Module.class.inc(351): LORIS\\electrophysiology_browser\\Sessions->_hasAccess(Object(User))\n#1 /var/www/loris/src/Middleware/ResponseGenerator.php(50): Module->handle(Object(Laminas\\Diactoros\\ServerRequest))\n#2 /var/www/loris/src/Middleware/AuthMiddleware.php(63): LORIS\\Middleware\\ResponseGenerator->process(Object(Laminas\\Diactoros\\ServerRequest), Object(LORIS\\electrophysiology_browser\\Module))\n#3 /var/www/loris/src/Router/ModuleRouter.php(75): LORIS\\Middleware\\AuthMiddleware->process(Object(Laminas\\Diactoros\\ServerRequest), Object(LORIS\\electrophysiology_browser\\Module))\n#4 /var/www/loris/src/Router/BaseRouter.php(105): LORIS\\Router\\ModuleRouter->handle(Object(Laminas\\Diactoros\\ServerRequest))\n#5 /var/www/loris/src/Middleware/ResponseGenerator.php(50): LORIS\\Router\\BaseRouter->handle(Object(Laminas\\Diactoros\\ServerRequest))\n#6 in /var/www/loris/modules/electrophysiology_browser/php/sessions.class.inc on line 51, referer: http://localhost:8080/electrophysiology_browser/

h-karim:
*Brief summary of changes*

The `_hasAccess` method is being called by `Module.class.inc` [here](https://github.com/aces/Loris/blob/23.0-release/php/libraries/Module.class.inc#L351). This is causing an error because because `$this.timepoint` is not initialised before the method call in the module class, which leads to calling `getCenterID` on null: https://github.com/aces/Loris/blob/a5304c762454fa8241ce0699af94b2da15f8879d/modules/electrophysiology_browser/php/sessions.class.inc#L51  

I added `hasStudySite` as part of the `_hasAccess` method to verify the user is affiliated with the study site of the scan session, and removed `$user->hasCenter($this->timepoint->getCenterID()`, to follow the same logic check as some of the other mdules. 

1. Assign your user "view own site EEG pages" only
2. Go to the EEG front page, click on one of the hyperlinks under the "Links" column
3. Observe session loading successfully

* Resolves #6557

### DIFF
--- a/modules/electrophysiology_browser/php/sessions.class.inc
+++ b/modules/electrophysiology_browser/php/sessions.class.inc
@@ -48,7 +48,7 @@ class Sessions extends \NDB_Page
     function _hasAccess(\User $user) : bool
     {
         return ($user->hasPermission('electrophysiology_browser_view_allsites')
-                || ($user->hasCenter($this->timepoint->getCenterID())
+                || ($user->hasStudySite() || $user->hasCenter($this->timepoint->getCenterID())
                 && $user->hasPermission('electrophysiology_browser_view_site')
             )
         );
--- a/modules/electrophysiology_browser/php/sessions.class.inc
+++ b/modules/electrophysiology_browser/php/sessions.class.inc
@@ -48,9 +48,10 @@ class Sessions extends \NDB_Page
     function _hasAccess(\User $user) : bool
     {
         return ($user->hasPermission('electrophysiology_browser_view_allsites')
-                || ($user->hasStudySite() || $user->hasCenter($this->timepoint->getCenterID())
+                || ($user->hasStudySite()
+                || $user->hasCenter($this->timepoint->getCenterID())
                 && $user->hasPermission('electrophysiology_browser_view_site')
-            )
+                    )
         );
     }
 
--- a/modules/electrophysiology_browser/php/sessions.class.inc
+++ b/modules/electrophysiology_browser/php/sessions.class.inc
@@ -19,7 +19,6 @@ use \Psr\Http\Message\ServerRequestInterface;
 use \Psr\Http\Message\ResponseInterface;
 use LORIS\electrophysiology_browser\Models\ElectrophysioFile;
 
-
 /**
  * Provides the PHP code for the menu filter for the electrophysiology browser
  *
@@ -49,10 +48,7 @@ class Sessions extends \NDB_Page
     {
         return ($user->hasPermission('electrophysiology_browser_view_allsites')
                 || ($user->hasStudySite()
-                || $user->hasCenter($this->timepoint->getCenterID())
-                && $user->hasPermission('electrophysiology_browser_view_site')
-                    )
-        );
+                && $user->hasPermission('electrophysiology_browser_view_site') ) );
     }
 
     /**


christinerogers:
also, is the PR title accurate?  maybe site instead of sessions?

h-karim:
> is it supposed to be used this way? Where else in the codebase is it used this way?

@christinerogers `hasStudySite` has been used similarly in the `_hasAccess` method in the brain browser:
https://github.com/aces/Loris/blob/a5304c762454fa8241ce0699af94b2da15f8879d/modules/brainbrowser/php/brainbrowser.class.inc#L34-L47

and the candidate list module: 
https://github.com/aces/Loris/blob/a5304c762454fa8241ce0699af94b2da15f8879d/modules/candidate_list/php/validateids.class.inc#L40-L46

modules/electrophysiology_browser/php/sessions.class.inc
```diff
@@ -48,9 +48,10 @@ class Sessions extends \NDB_Page
     function _hasAccess(\User $user) : bool
     {
         return ($user->hasPermission('electrophysiology_browser_view_allsites')
-                || ($user->hasCenter($this->timepoint->getCenterID())
+                || ($user->hasStudySite()
+                || $user->hasCenter($this->timepoint->getCenterID())
                 && $user->hasPermission('electrophysiology_browser_view_site')
-            )
+                    )
```
christinerogers, line 54: too much whitespace here i think

h-karim: I found it better indicates where the beginning parenthesis starts since they're vertically aligned. Will undo it.

christinerogers:
@johnsaigle  - How about spinning your Proposal off into a new issue for discussion?  
At a minimum this should be checked for consistency across modules.

johnsaigle:
I don't consider this a proposal, I think it's an issue with the permissions logic in this code.

h-karim:
@johnsaigle Considering the brain browser and candidate list modules (among possible others) have a similar line of logic for permissions:
https://github.com/aces/Loris/blob/a5304c762454fa8241ce0699af94b2da15f8879d/modules/brainbrowser/php/brainbrowser.class.inc#L34-L47
https://github.com/aces/Loris/blob/a5304c762454fa8241ce0699af94b2da15f8879d/modules/candidate_list/php/validateids.class.inc#L40-L46

Perhaps an umbrella issue/proposal should be created with the eventual goal to centralize all the permissions in one place? The way it's being handled now has the potential to create significant technical debt.

 
>I don't think that a user being able to access any study site has anything to do with whether they are viewing some particualr EEG session.

I think the logic was going on the premises that the module table is displaying only the site data which the user has access to, and that during user creation a user must be assigned a site, therefore the user would be able to access the individual sessions of the table data. 

The ideal way you quoted is what is currently implemented : https://github.com/aces/Loris/blob/23.0-release/modules/electrophysiology_browser/php/sessions.class.inc#L50-L54 , it's checking whether the user has the `electrophysiology_browser_view_allsites` permission or the alternative you proposed, which is why #6557 only comes up when the user _only_ has the `electrophysiology_browser_view_site` permission.

johnsaigle:
> The ideal way you quoted is what is currently implemented 
Yeah I took another look and I think you're right. The logical operators and the brackets are confusing to me in this module. The `candidate_list` example you quoted is much more clear to me but I think these are logically equivalent. 

Sorry for the tangent

laemtl:
Tested and works fine.

christinerogers:
> Sorry for the tangent

good. 

Karim's right -- yes, we are moving towards centralizing / coordinating common definitions of permissions, creating less risk of inconsistencies across modules and technical debt.  
So far this has happened via the new data framework -- which is not applying / designed to apply to module subpages.

christinerogers:
@AlexandraLivadas  or @johnsaigle could either of you re/review this permissions fix?  thanks

johnsaigle:
I agree with @ridz1208 that this should probably be done a bit differently. I know Dave commented on a similar issue on Alex was working on about a possible approach using the module's `process` method. That would be a better way to go then to deliberately include (or retain) broken code.

christinerogers:
@h-karim can you move forward with a solution to this today?  a new PR is fine if that seems best.  
(cc rope in @laemtl if you like)

h-karim:
@ridz1208 Echoing @AlexandraLivadas 's [comment](https://github.com/aces/Loris/pull/6640#discussion_r433250453) overriding the `process` method inside the Sessions class does not seem to solve the issue: 
```php

public function process(
        ServerRequestInterface $request,
        RequestHandlerInterface $handler ) : ResponseInterface {

        $this->timePoint = $request->getAttribute("timePoint");
        if ($this->timePoint === null) {
            // This shouldn't happen, the module level router verifies that
            // the candidate exists before getting here.
            throw new \LorisException("Timepoint list requires a timepoint.");
        }
        return parent::process($request, $handler);
    }
```

If the current PR is the wrong fix, there are some other modules (perhaps create a separate issue for them?) that implement this same approach that should be addressed. For example:
- Brainbrowser
https://github.com/aces/Loris/blob/d6c3fd70f44f2ae5d691f889887ab241755a7a96/modules/brainbrowser/php/brainbrowser.class.inc#L39-L49

- imagingbrowser: https://github.com/aces/Loris/blob/938c11c71953b4a0b906ce2e8bb38308f4878b16/modules/imaging_browser/php/imaging_browser.class.inc#L45-L55
- Candidate_list: https://github.com/aces/Loris/blob/938c11c71953b4a0b906ce2e8bb38308f4878b16/modules/candidate_list/php/candidate_list.class.inc#L33-L39

ridz1208:
@driusan can you look into these issues ?

driusan:
None of those pieces of code are being used to try and short-circuit out of other checks. They're being used to verify that if the user only has the "site" permission for the module, the user has a study site.

h-karim:
I'll remove `$user->hasCenter($this->timepoint->getCenterID())` from the check which will become :

```php
    {
        return ($user->hasPermission('electrophysiology_browser_view_allsites')
                || ($user->hasStudySite()
                    && $user->hasPermission('electrophysiology_browser_view_site')
            )
        );
    }
```
following the same logic as the other pieces of code.

christinerogers:
For LORIS Magic Monday @h-karim : See [Rida's comments for Alex in #6640](https://github.com/aces/Loris/pull/6640#issuecomment-650420925) and then follow her lead...

h-karim:
#6640 takes care of the linked issue. Closing this PR.

PR #6471, opened 2020-05-04
[issue_tracker] Prevent users from accessing issues to which they don't have the appropriate Center

HenriRabalais:
## Brief summary of changes
This PR fixes an issue which allowed users to view issues to which they didn't have the proper center.

#### Testing instructions (if applicable)
1. Go to issue_tracker module. 
2. Go to an issue url which exists but to which the user doesn't have access based center.
3. Ensure that permission to view is denied.

#### Link(s) to related issue(s)
#6282

PR #6640, opened 2020-05-28
[electrophysiology_browser] Session page: Add Project Permissions check

### LINKED ISSUE #6558: [EEG Browser] User can access individual sessions for projects they are not involed in
**Context**
Fails step A.2 of the test plan. Related to #6618

**Describe the bug**
A user can access individual sessions for a project that the user has not been assigned to if the user has a url from the 'raw' or 'all types' links or logically deduces the url.

**To Reproduce**

Steps to reproduce the behavior on the testing VM:
1. Using an admin user, create a user `foo` with the following permission:
>View all-sites Electrophysiology Browser pages
2. Assign the user project "pumpernickel" and site "Ottawa"
3. Navigate to the EEG Browser module front page using `foo`
4. Notice data for pumpernickel is displayed, copy either the "all types" or "raw" link from one of the rows to your clipboard, (e.g this link: https://test-loris-dev.loris.ca/electrophysiology_browser/sessions/167 )
5. Using an admin account, change  `foo`'s project assignment to any other project, make sure to exclude "pumpernickel"
6. Switch back to `foo`
7. Notice the module front page does not show any data (on testing vm)
8. paste the copied url
9. Notice `foo` is able to access the individual session, despite not being assigned the pumpernickel project, and is now able to click on "next" and "Previous" to access all the data for project pumpernickel.
For the user to have been denied access to the individual session when the user is not assigned the project covering the particular session

- OS: Ubuntu 18.04 LTS
- Browser chrome 83.0.4103.61 and firefox 76.0.1

This was done using the testing VM for loris

AlexandraLivadas:
## Brief summary of changes

When a user tries to access the individual session page for a project they are not affiliated with (by altering the URL), a "permission denied" message should appear.

1. Create a user and give them `View all-sites Electrophysiology Browser pages` permission.
2. Assign the user project "Pumpernickel" and site "Ottawa"
3. Navigate to the EEG browser page as that user.
4. Copy either the "all-types" or "raw" link from one of the rows.
5. As an admin, change the user to a different project (anything other than Pumpernickel)
6. Go back to the user account and to the EEG Browser page.
7. Note that there are no sessions listed (on the test VM)
8. Paste the copied URL.
9. There should be a "Permission denied" message and no data should be displayed.  

#### Links to related issues

* Resolves #6558
* Related to issue #6557 and associated PR #6639

### DIFF
--- a/modules/electrophysiology_browser/php/sessions.class.inc
+++ b/modules/electrophysiology_browser/php/sessions.class.inc
@@ -37,6 +37,7 @@ class Sessions extends \NDB_Page
     public $skipTemplate = true; // stops from looking for a smarty template
     protected $timepoint;
     protected $sessionID;
+    protected $candidate;
 
     /**
      * Determine whether the user has permission to view this page
@@ -47,29 +48,29 @@ class Sessions extends \NDB_Page
      */
     function _hasAccess(\User $user) : bool
     {
-        return ($user->hasPermission('electrophysiology_browser_view_allsites')
+        return (($user->hasPermission('electrophysiology_browser_view_allsites')
                 || ($user->hasCenter($this->timepoint->getCenterID())
                 && $user->hasPermission('electrophysiology_browser_view_site')
             )
-        );
+        ) && $user->hasProject($this->candidate->getProjectID()));
     }
 
     /**
-     * Handles a login request
+     * Load the required variables in order to check that the  user
+     * has access to the session.
      *
-     * @param ServerRequestInterface $request The incoming PSR7 request
+     * @param \User                  $user    The user to load the resources for
+     * @param ServerRequestInterface $request The PSR15 Request being handled
      *
-     * @return ResponseInterface The outgoing PSR7 response
+     * @return ResponseInterface
      */
-    public function handle(ServerRequestInterface $request) : ResponseInterface
-    {
-        $path       = $request->getUri()->getPath();
-        $parameters = $request->getQueryParams();
-        $user       = $request->getAttribute('user');
+    public function loadResources(
+        \User $user, ServerRequestInterface $request
+    ) : ResponseInterface {
 
+        $path    = $request->getUri()->getPath();
         $matches = [];
 
-        // check that the session ID is of type integer
         if (preg_match('#/sessions/(\d+)#', $path, $matches) !== 1) {
             return (new \LORIS\Middleware\PageDecorationMiddleware($user))
                 ->process(
@@ -82,6 +83,26 @@ class Sessions extends \NDB_Page
 
         $this->timepoint = \NDB_Factory::singleton()->timepoint($session_id);
         $this->sessionID = $session_id;
+
+        $candID          = $this->timepoint->getCandID();
+        $this->candidate = \Candidate::singleton($candID);
+
+        parent::loadResources($user, $request);
+        return (new \LORIS\Http\Response())
+            ->withBody(new \LORIS\Http\StringStream($this->display() ?? ""));
+    }
+
+    /**
+     * Handles a login request
+     *
+     * @param ServerRequestInterface $request The incoming PSR7 request
+     *
+     * @return ResponseInterface The outgoing PSR7 response
+     */
+    public function handle(ServerRequestInterface $request) : ResponseInterface
+    {
+        $parameters = $request->getQueryParams();
+        $user       = $request->getAttribute('user');
 
         if (!$this->_hasAccess($user)) {
             return (new \LORIS\Middleware\PageDecorationMiddleware($user))
--- a/modules/electrophysiology_browser/php/sessions.class.inc
+++ b/modules/electrophysiology_browser/php/sessions.class.inc
@@ -76,14 +76,9 @@ class Sessions extends \NDB_Page
             throw new \DomainException("Invalid session");
         }
         $session_id = intval($matches[1]);
-        try {
-            $this->timepoint = \NDB_Factory::singleton()->timepoint(
-                $session_id
-            );
-            $this->sessionID = $session_id;
-        } catch(\LorisException $e) {
-            throw $e;
-        }
+
+        $this->timepoint = \NDB_Factory::singleton()->timepoint($session_id);
+        $this->sessionID = $session_id;
 
         $candID          = $this->timepoint->getCandID();
         $this->candidate = \Candidate::singleton($candID);
--- a/modules/electrophysiology_browser/php/sessions.class.inc
+++ b/modules/electrophysiology_browser/php/sessions.class.inc
@@ -79,18 +79,9 @@ class Sessions extends \NDB_Page
         }
 
         $session_id = intval($matches[1]);
-        try {
-            $this->timepoint = \NDB_Factory::singleton()->timepoint(
-                $session_id
-            );
-            $this->sessionID = $session_id;
-        } catch(\LorisException $e) {
-            return (new \LORIS\Middleware\PageDecorationMiddleware($user))
-                ->process(
-                    $request,
-                    new \LORIS\Http\StringStream("Session not found")
-                )->withStatus(404);
-        }
+
+        $this->timepoint = \NDB_Factory::singleton()->timepoint($session_id);
+        $this->sessionID = $session_id;
 
         if (!$this->_hasAccess($user)) {
             return (new \LORIS\Middleware\PageDecorationMiddleware($user))
--- a/modules/electrophysiology_browser/php/sessions.class.inc
+++ b/modules/electrophysiology_browser/php/sessions.class.inc
@@ -77,8 +77,14 @@ class Sessions extends \NDB_Page
         }
         $session_id = intval($matches[1]);
 
-        $this->timepoint = \NDB_Factory::singleton()->timepoint($session_id);
-        $this->sessionID = $session_id;
+        try {
+            $this->timepoint = \NDB_Factory::singleton()->timepoint(
+                $session_id
+            );
+            $this->sessionID = $session_id;
+        } catch(\LorisException $e) {
+            throw new \NotFound("Session not found");
+        }
 
         $candID          = $this->timepoint->getCandID();
         $this->candidate = \Candidate::singleton($candID);
--- a/modules/electrophysiology_browser/php/sessions.class.inc
+++ b/modules/electrophysiology_browser/php/sessions.class.inc
@@ -60,7 +60,7 @@ class Sessions extends \NDB_Page
      * @param \User                  $user    The user to load the resources for
      * @param ServerRequestInterface $request The PSR15 Request being handled
      *
-     * @throws \DomainException If the session id is non-numerical
+     * @throws \NotFound If the session id is non-numerical
      * @throws \LorisException  If the session is not found
      *
      * @return void
@@ -73,7 +73,7 @@ class Sessions extends \NDB_Page
         $matches = [];
 
         if (preg_match('#/sessions/(\d+)#', $path, $matches) !== 1) {
-            throw new \DomainException("Invalid session");
+            throw new \NotFound("Invalid session");
         }
         $session_id = intval($matches[1]);
 
--- a/modules/electrophysiology_browser/php/sessions.class.inc
+++ b/modules/electrophysiology_browser/php/sessions.class.inc
@@ -50,9 +50,8 @@ class Sessions extends \NDB_Page
     {
         return (($user->hasPermission('electrophysiology_browser_view_allsites')
                 || ($user->hasCenter($this->timepoint->getCenterID())
-                && $user->hasPermission('electrophysiology_browser_view_site')
-            )
-        ) && $user->hasProject($this->candidate->getProjectID()));
+                && $user->hasPermission('electrophysiology_browser_view_site'))
+            ) && $user->hasProject($this->candidate->getProjectID()));
     }
 
     /**
--- a/modules/electrophysiology_browser/php/sessions.class.inc
+++ b/modules/electrophysiology_browser/php/sessions.class.inc
@@ -60,7 +60,8 @@ class Sessions extends \NDB_Page
      * @param \User                  $user    The user to load the resources for
      * @param ServerRequestInterface $request The PSR15 Request being handled
      *
-     * @throws \LorisException If the session is non-numerical or does not exist
+     * @throws \DomainException If the session id is non-numerical
+     * @throws \LorisException  If the session is not found
      *
      * @return void
      */
@@ -72,7 +73,7 @@ class Sessions extends \NDB_Page
         $matches = [];
 
         if (preg_match('#/sessions/(\d+)#', $path, $matches) !== 1) {
-            throw new \LorisException("Invalid session");
+            throw new \DomainException("Invalid session");
         }
         $session_id = intval($matches[1]);
         try {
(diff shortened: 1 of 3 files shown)


modules/electrophysiology_browser/php/sessions.class.inc
```diff
@@ -92,7 +92,11 @@ class Sessions extends \NDB_Page
                 )->withStatus(404);
         }
 
-        if (!$this->_hasAccess($user)) {
+        $candID    = $this->timepoint->getCandID();
+        $candidate = \Candidate::singleton($candID);
+        if (!$this->_hasAccess($user)
+            || !$user->hasProject($candidate->getProjectID())
```
ridz1208, line 98: @AlexandraLivadas I think it's better to put the project access check inside the hasAccess function

AlexandraLivadas: @ridz1208 This wouldn't work because the `timepoint` variable is not properly defined when the _hasAccess method is first called. This is also what caused the issue #6557. The _hasAccess method is called first by the Module class [here](https://github.com/aces/Loris/blob/a5304c762454fa8241ce0699af94b2da15f8879d/php/libraries/Module.class.inc#L351), and it is called before the `handle` method is called where the `timepoint` variable is properly defined. So, I couldn't include it in the method because it would give a null pointer error. Let me know if this explanation makes sense!

ridz1208: @AlexandraLivadas it makes sense. I think I've seen this usecase before

@HenriRabalais isnt that a similar problem to the one you had with issue tracker? do you remember what Dave suggested toadd to the class in order to instanciate the object before getting to the has access function ?

ridz1208: https://github.com/aces/Loris/pull/6471/files

this might be helpful. look at the comment of the `process` function.

AlexandraLivadas: I tried this approach and it did not work. I think it is because this class (Sessions) is of type `NDB_Page`, while the one you linked is of type `NDB_Form`. If you look at the [Module](https://github.com/aces/Loris/blob/23.0-release/php/libraries/Module.class.inc) class, the method `process` is never called for the actual page. So, when I created the `process` method and defined the timepoint variable in there and then moved the project permission line into the `_hasAccess` method, it gave me a null pointer error again.

johnsaigle: Shouldn't that mean that line 51 also fails because it accesses timepoint?

AlexandraLivadas: Yes @johnsaigle, this is addressed in issue #6557 and associated PR #6639. Line 51 checks for site affiliation permissions, which is why in the testing instructions I specify that in order to test this the user needs to have all-site access.

johnsaigle: OK thanks I didn't see the link previously.

AlexandraLivadas: I will add it in the PR description just in case!

driusan: If you need access to a class variable based on request data in the `hasAccess` function, just set it in the `process()` middleware before calling the parent. See modules/instrument_list/php/instrument_list.class.inc for an example.

ridz1208: @driusan I sugested that above. @AlexandraLivadas is saying it didnt work

driusan: All page types should have the `process` middleware called on it, there shouldn't be anything specific to forms in that regard. (Forms are special in that they have a `_process`, but that's different than and unrelated to the PSR15 middleware `process`)

AlexandraLivadas: @driusan I tried again following a very similar format to `modules/instrument_list/php/instrument_list.class.inc` and it still gives me the following error when I move the project permission to the _hasAccess method:
 ```
PHP Fatal error:  Uncaught Error: Call to a member function getProjectID() on null in /var/www/loris2020/modules/electrophysiology_browser/php/sessions.class.inc:56\nStack trace:\n#0 /var/www/loris2020/php/libraries/Module.class.inc(351): LORIS\\electrophysiology_browser\\Sessions->_hasAccess(Object(User))\n#1 /var/www/loris2020/src/Middleware/ResponseGenerator.php(50): Module->handle(Object(Laminas\\Diactoros\\ServerRequest))\n#2 /var/www/loris2020/src/Middleware/AuthMiddleware.php(63): LORIS\\Middleware\\ResponseGenerator->process(Object(Laminas\\Diactoros\\ServerRequest), Object(LORIS\\electrophysiology_browser\\Module))\n#3 /var/www/loris2020/src/Router/ModuleRouter.php(75): LORIS\\Middleware\\AuthMiddleware->process(Object(Laminas\\Diactoros\\ServerRequest), Object(LORIS\\electrophysiology_browser\\Module))\n#4 /var/www/loris2020/src/Router/BaseRouter.php(105): LORIS\\Router\\ModuleRouter->handle(Object(Laminas\\Diactoros\\ServerRequest))\n#5 /var/www/loris2020/src/Middleware/ResponseGenerator.php(50): LORIS\\Router\\BaseRouter->handle(Object(Laminas in /var/www/loris2020/modules/electrophysiology_browser/php/sessions.class.inc on line 56, referer: https://alivadas-dev.loris.ca/electrophysiology_browser
```

I included the if-statements from the instrument_list.class.inc both in the [_hasAccess](https://github.com/aces/Loris/blob/1f98c1596872adc965d1977f1a2cd23210451363/modules/instrument_list/php/instrument_list.class.inc#L85) and [process](https://github.com/aces/Loris/blob/1f98c1596872adc965d1977f1a2cd23210451363/modules/instrument_list/php/instrument_list.class.inc#L132) methods to check if the timepoint and candidate variables are null. The `\NotFound` error (from _hasAccess) appeared but the `\LorisException` error (from process) did not. So I think that `_hasAccess` is being called even before `process` is called, which is why there is still a null pointer exception.

**Comment by christinerogers** [comment]
pinging @ridz1208 for re-review, since @AlexandraLivadas should have addressed your changes already 

@driusan - should be ready for final review

File: modules/electrophysiology_browser/php/sessions.class.inc (hunk 1 above)
Comment (ridz1208) [reply modules/electrophysiology_browser/php/sessions.class.inc:98]: @AlexandraLivadas any news from you investigation ?

**Comment by AlexandraLivadas** [comment]
The changes requested by @ridz1208 and @driusan were attempted but did not work. A similar change was requested for PR #6639 and @h-karim also had trouble implementing it.

File: modules/electrophysiology_browser/php/sessions.class.inc (hunk 1 above)
Comment (AlexandraLivadas) [reply modules/electrophysiology_browser/php/sessions.class.inc:98]: @h-karim and I have been running into the same problem when trying to implement the `process` method (check out his PR #6639). After some digging, we both found that the `process` method is called **after** the `_hasAccess` method, so this fix would not work.

**Comment by ridz1208** [comment]
@AlexandraLivadas 

@driusan and I have discussed this issue at length yesterday.

**explanation :** the reason @HenriRabalais 's fix on #6471 worked is because he added a null check for `$this->issueID` in the `_hasAccess()` function causing it to return true when it is first checked. After which the resources were being instanciated and finally the `_hasAccess` was called again and only then proper permission checking was done.

We want to change this behaviour because it's not efficient and slightly hacky.

**What to do:** We decided that a short term stable fix would be the following. In the `php/libraries/Module.class.inc`, the `$page->_hasAccess()` function is called on line 351. We are proposing to add a function call on line 350 `$page->loadResources($user,$request)` this function would be defined in `NDB_Page` class and does absolutely nothing at that level, but in each module's pages you can override this function and instanciate all the resources you need (timepoint, issue, users,...)

So in this PR what we would expect to see is the addition of the `NDB_Page->loadResources($user,$request)` function, addition on line 350 of the `Module.class.inc` class of a call to the new `NDB_Page` function and, in `modules/electrophysiology_browser/php/sessions.class.inc` addition of an override for `loadResources()` function (the content of this function would instanciate the necessary timepoit to allow `_hasAccess()` to check for permissions)

File: php/libraries/NDB_Page.class.inc [hunk 2]
```diff
@@ -709,6 +709,22 @@ class NDB_Page implements RequestHandlerInterface
...
+     * resources to check the permissions of a user.
+     *
+     * @param User                   $user    The user to load the resources for
+     * @param ServerRequestInterface $request The PSR15 Request being handled
+     *
+     * @return ResponseInterface
+     */
+    public function loadResources(
+        \User $user, ServerRequestInterface $request
+    ) : ResponseInterface {
```
driusan, line 723: There is no reason for this to return anything (and also, the return value isn't used anywhere..)

AlexandraLivadas: I have it returning something because in the session class, I moved some of the code that was originally found in the `handle` method to `loadResources` and that includes returning exceptions if there is an invalid session or if the session is not found (lines 73-92). 

I can remove this part of the code and try to put it back in the `handle` method if this is preferrable.

AlexandraLivadas: @driusan In the latest commit, `loadResources` now returns void. I changed the code to throw LorisExceptions (instead of returning a ResponseInterface) if the session id is either non-numeric or does not exist. Let me know if this works better!

modules/electrophysiology_browser/php/sessions.class.inc
```diff
@@ -47,51 +47,61 @@ class Sessions extends \NDB_Page
...
         $matches = [];
 
-        // check that the session ID is of type integer
         if (preg_match('#/sessions/(\d+)#', $path, $matches) !== 1) {
-            return (new \LORIS\Middleware\PageDecorationMiddleware($user))
-                ->process(
-                    $request,
-                    new \LORIS\Http\StringStream("Invalid session")
-                )->withStatus(404);
+            throw new \LorisException("Invalid session");
```
driusan, line 75: sounds more like a DomainException

modules/electrophysiology_browser/php/sessions.class.inc
```diff
@@ -47,51 +47,61 @@ class Sessions extends \NDB_Page
...
                 $session_id
             );
             $this->sessionID = $session_id;
         } catch(\LorisException $e) {
-            return (new \LORIS\Middleware\PageDecorationMiddleware($user))
-                ->process(
-                    $request,
-                    new \LORIS\Http\StringStream("Session not found")
-                )->withStatus(404);
+            throw new \LorisException("Session not found");
```
driusan, line 84: There's no reason to catch an exception just to throw the same exception type. All you're doing is losing details from the original exception.

modules/electrophysiology_browser/php/sessions.class.inc
```diff
@@ -47,51 +47,62 @@ class Sessions extends \NDB_Page
...
                 $session_id
             );
             $this->sessionID = $session_id;
         } catch(\LorisException $e) {
-            return (new \LORIS\Middleware\PageDecorationMiddleware($user))
-                ->process(
-                    $request,
-                    new \LORIS\Http\StringStream("Session not found")
-                )->withStatus(404);
+            throw $e;
```
johnsaigle, line 85: I don't know if there's much point in catching an exception here just to throw it. I feel like this try-catch can be deleted -- or the error should be thrown with a relevant message.

modules/electrophysiology_browser/php/sessions.class.inc
```diff
@@ -47,11 +47,43 @@ class Sessions extends \NDB_Page
...
+     */
+    public function loadResources(
+        \User $user, ServerRequestInterface $request
+    ) : void {
+
+        $path    = $request->getUri()->getPath();
+        $matches = [];
+
+        if (preg_match('#/sessions/(\d+)#', $path, $matches) !== 1) {
+            throw new \DomainException("Invalid session");
```
driusan, line 76: This is going to cause a 500 error when the path is invalid, which is incorrect, it should still be a 404 error. It would be better to do something like:

Some solutions I can think of off the top of my head:
1. Set a class variable (something like "invalidSession = true") and return if that happens. Return "true" for hasAccess if invalid so that handle still gets called. Return the 404 from handle
2. change loadResources to return an error code instead of void, modify the module class to check it and return a 404 if the request was invalid (this assumes that the only reason an error would be returned is an invalid request, which is probably true.)

@ridz1208 any thoughts on which you prefer?

AlexandraLivadas: I can take a crack at both those options and let you know which works better or which is simpler. Thanks!

driusan: Option 3 is probably a better idea: change the exception type (again) to a NotFound exception so that the try/catch around the loadResources handles the exception to return a 404.

modules/electrophysiology_browser/php/sessions.class.inc
```diff
@@ -47,11 +47,43 @@ class Sessions extends \NDB_Page
...
+     * @return void
+     */
+    public function loadResources(
+        \User $user, ServerRequestInterface $request
+    ) : void {
+
+        $path    = $request->getUri()->getPath();
+        $matches = [];
+
+        if (preg_match('#/sessions/(\d+)#', $path, $matches) !== 1) {
```
AlexandraLivadas, line 75: @driusan This now correctly displays "404 Not Found" if the session information in the URL is not properly formatted (i.e a letter instead of a number)

modules/electrophysiology_browser/php/sessions.class.inc
```diff
@@ -47,11 +47,43 @@ class Sessions extends \NDB_Page
...
+            throw new \NotFound("Invalid session");
+        }
+        $session_id = intval($matches[1]);
+
+        $this->timepoint = \NDB_Factory::singleton()->timepoint($session_id);
```
AlexandraLivadas, line 80: @driusan This section returns a 500 Internal Server Error if the session/timepoint does not exist. Is this okay or should it return some other type of exception?

driusan: I think it should be converted to a 404 not found.

modules/electrophysiology_browser/php/sessions.class.inc
```diff
@@ -47,51 +47,63 @@ class Sessions extends \NDB_Page
...
     function _hasAccess(\User $user) : bool
     {
-        return ($user->hasPermission('electrophysiology_browser_view_allsites')
+        return (($user->hasPermission('electrophysiology_browser_view_allsites')
                 || ($user->hasCenter($this->timepoint->getCenterID())
-                && $user->hasPermission('electrophysiology_browser_view_site')
-            )
-        );
+                && $user->hasPermission('electrophysiology_browser_view_site'))
+            ) && $user->hasProject($this->candidate->getProjectID()));
```
driusan, line 53: This doesn't look right.. why is the site checking the timepoint's site but the project is checking the candidate's project? Shouldn't they both be the timepoint's (site/project)?

AlexandraLivadas: Yes, it should be checking for the timepoint's project. I assumed that the timepoint and candidate projects would match (since I'm getting the candidate related to that timepoint), but I will change the code to get the project ID directly from the timepoint variable instead.

AlexandraLivadas: This was resolved in the latest commit. I tested on both Chrome and Firefox and it still correctly stops a user from seeing pages for projects it's not affiliated with.

Issue #6567, opened 2020-05-21
[instrument_builder] Unable to remove Date range values once entered


zaliqarosli:
**Describe the bug**
When adding a 'Date' element, although the 'Start year' and 'End year' fields are not required (you can 'Add Row' if fields are left empty), you cannot fully delete values within the fields once the values are entered. The first digit is not deletable.

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to 'Instrument Builder' module and click on the 'Build' tab
2. Click on the 'Select One' dropdown and select 'Date' under 'Data entry'
3. Fill out required fields, leaving 'Start year' and 'End year' fields empty. Click 'Add Row'. The element should save in the 'Question Display (Front End)' table above.
4. Do this again but fill in the 'Start year' and 'End year' fields. Try to remove the values you have just entered in those fields. The first digit is forced to remain in the field.

**What did you expect to happen?**
Values entered in the 'Start year' and 'End year' fields should be fully removable if it is not required by default

PR #6642, opened 2020-05-28
[instrument_builder] Let mindate and maxdate options be removable

zaliqarosli:
## Brief summary of changes

This PR allows the Start year and End year values for Date elements be emptied

1. On branch 23.0-release, add a Date element and notice that once you enter a value for both Start year and End year, you are not able to make the fields empty
2. Save the instrument with one Date element without min/max values, and another Date element with min/max values.
3. On this PR's branch, add a Date element and notice that you can make it empty after inserting a value. Make sure you run `make dev` after changing branches.
4. Save the instrument with the same elements as step 2.
5. Compare the two saved linst files and see that they are identical. This PR should only affect the front-end instrument builder. Please report if otherwise.

* Resolves https://github.com/aces/Loris/issues/6567

### DIFF
--- a/modules/instrument_builder/jsx/react.questions.js
+++ b/modules/instrument_builder/jsx/react.questions.js
@@ -324,10 +324,10 @@ class DateOptions extends Component {
     // Keep track of the inputed years
   onChange(e) {
     let options = Instrument.clone(this.props.element.Options);
-    if (e.target.id === 'datemin' && e.target.value.length > 0) {
-      options.MinDate = e.target.value + '-01-01';
-    } else if (e.target.id === 'datemax' && e.target.value.length > 0) {
-      options.MaxDate = e.target.value + '-12-31';
+    if (e.target.id === 'datemin') {
+      options.MinDate = e.target.value;
+    } else if (e.target.id === 'datemax') {
+      options.MaxDate = e.target.value;
     } else if (e.target.id === 'dateFormat') {
       options.dateFormat = e.target.value;
     }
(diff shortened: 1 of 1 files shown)


modules/instrument_builder/jsx/react.questions.js
```diff
@@ -336,8 +336,8 @@ class DateOptions extends Component {
   // Render the HTML
   render() {
     // Truncate off the month and day from the date to only have the year.
-    let minYear = this.props.element.Options.MinDate.split('-')[0];
```
driusan, line 339: both the comment above and the variable names are wrong.. why is this change being done? It seems unrelated to the title and I'm not sure of the impact of it on existing instruments.

zaliqarosli: there's no impact on existing instruments i don't think. only within the instrument builder, the minyear and maxyear are being joined by 01-01 and 31-12 to make a full date which is unnecessary because linst only takes the year anyways. this line above that i'm removing then removes the month and day to only take the year. so i'm removing any addition/subtraction of day and month

driusan: what about when loading existing instruments?

zaliqarosli: hmmm something seems to be going wrong when you reload the linst and resave, i'll fix that now

zaliqarosli: @driusan that is a separate bug reported here https://github.com/aces/Loris/issues/6669. this PR makes no changes to the linst file output

zaliqarosli: > what about when loading existing instruments?

existing instruments only carry the year, so i don't understand why having the day and month within the instrument builder affects that

driusan:
@zaliqarosli Travis doesn't seem to be started on this, maybe it needs to be rebased?

driusan:
This passed Travis if you click the link, the icon just didn't update.

PR #6643, opened 2020-05-28
[Issue Tracker] Description and special characters decoding fix

### LINKED ISSUE #6614: [issue_tracker] Special characters in description
**Describe the bug**
Special characters appear as &amp;, etc... after saving (See screenshot below)

Screen Shot 2020-05-26 at 12 55 13 PM

laemtl:
Minor fix to decode an issue description special characters (&<>).
* Resolves #6614

### DIFF
--- a/modules/issue_tracker/ajax/EditIssue.php
+++ b/modules/issue_tracker/ajax/EditIssue.php
@@ -707,7 +707,12 @@ ORDER BY dateAdded LIMIT 1",
         $issueData['attachments']    = $attachments;
         $issueData['whoami']         = $username;
         $issueData['othersWatching'] = getWatching($issueID);
-        $issueData['desc']           = $desc[0]['issueComment'] ?? '';
+
+        // We need to unescape the string here:
+        // React is escaping the string in the template
+        // This fixes an issue with multiple escaping (#6643)
+        $comment           = $desc[0]['issueComment'];
+        $issueData['desc'] = html_entity_decode($comment) ?? '';
     }
     $issueData['comment'] = null;
 


ridz1208:
@laemtl were you able to identify where it was being encoded in the first place ? the PR @johnsaigle linked in the issue (#6223)  was the result of sequentially encoding the data during the process of saving.

Is it the same case here ? is the data being encoded several times before being saved and if so, can you identify where it is happening ?

laemtl:
@ridz1208 Yes, it is the same case here, encoding happens on save. My understanding of the problem is that the data is not encoding twice, but that the string is not decoded on load so the encoded characters are interpreted as plain text.

johnsaigle:
> whats the use of escaping the data if it's gonna get unescaped before being sent to the browser...

The best way to handle data safely is to:
* to validation on data as it comes in (reject bad data)
* "sanitize" the output as far downstream in the rendering as possible (i.e. Smarty/React)

We have a bunch of escaping/unescaping because:
* LORIS (still) passes around a bunch of raw HTML. (Mostly because of instruments but also some core stuff). This causes Smarty to break if we enable its autoescaping feature (https://github.com/aces/Loris/pull/2956)
* Modules are very ad-hoc and it makes it difficult to determine in one central place where output can be sanitized
* Not everything is React. React safely handles data in innerHTML by default which takes care of a lot of our security needs. ([Props are not sanitized](https://medium.com/javascript-security/avoiding-xss-in-react-is-still-hard-d2b5c7ad9412), however.)
* Old codebase, not enough development time, refactoring needs to be prioritized, code rot, etc etc etc

johnsaigle:
@ridz1208 See https://github.com/aces/Loris/issues/6648

ridz1208:
@johnsaigle what would be your recommendation for this release ? just unescape the HTML for now ?

johnsaigle:
@ridz1208 Yeah this is a very deep and longstanding problem. We won't be able to solve it for this release. Until we have the time and will to fix the root of this problem, we will have to keep unescaping things as bugs arise.

modules/issue_tracker/ajax/EditIssue.php
```diff
@@ -703,11 +703,12 @@ function getIssueFields()
             $issueData['watching'] = "Yes";
         }
         $username = $user->getUsername();
+        $comment  = $desc[0]['issueComment'];
         $issueData['commentHistory'] = getComments($issueID);
         $issueData['attachments']    = $attachments;
         $issueData['whoami']         = $username;
         $issueData['othersWatching'] = getWatching($issueID);
-        $issueData['desc']           = $desc[0]['issueComment'] ?? '';
+        $issueData['desc']           = html_entity_decode($comment) ?? '';
```
driusan, line 711: Can you add a comment (to the code) explaining why this should be safe from XSS attacks?

ridz1208: @driusan I approved the PR, can you just make sure the comment added is good for you ?

PR #6644, opened 2020-05-29
[Survey] Infinite redirect

### LINKED ISSUE #6587: [survey_accounts] Can't complete surveys in Firefox
**Describe the bug**
On a **Firefox** browser, once the last page of a survey is completed and the button "Save and Finalize" is pressed, the next page (to submit) does not load. A screenshot is included with the exact error message that appears and the Developer Console message. This error occurs both when the URL is selected from the Survey Accounts page and when the URL is selected from the e-mail sent to the user. The same error occurs for all types of instruments. The survey is (correctly) updated to "In Progress", but cannot be completed.
**Note:** The error message says that it could be caused due to disabling cookies. Cookies were enabled on the browser and the error still appeared. 
**Note:** When the URL is copied to a Chrome browser, the page loads correctly

Screen Shot 2020-05-22 at 12 12 45 PM

**To Reproduce (on a Firefox Browser)**
1. Create a user account on the testing VM with the "User management" permission.
2. Go to the Survey Accounts page for this user. 
3. Select the URL for a survey that has either "In Progress" or "Created" status
4. Fill out the survey until the option to "Save and Finalize" appears. Click on this button.
5. See error. 

The next page should show the completed survey, a "Survey Comments" section, and the button "Submit Data"

**Browser Environment:**
Firefox 76.0.1

laemtl:
A header location in survey is set with a variable which can be null. This has an unpredictable behaviour, and in Firefox, it seems to cause the reload of the page, where again, the header location is interpreted, causing an infinite redirection. I solved this with a check for Null before redirecting.

I also fixed some type issues (+1 on pageNum = finalpage).

* Resolves #6587

### DIFF
--- a/htdocs/survey.php
+++ b/htdocs/survey.php
@@ -82,7 +82,7 @@ class DirectDataEntryMainPage
             array('key' => $this->key)
         );
         $this->CommentID = $DB->pselectOne(
-            "SELECT CommentID FROM participant_accounts 
+            "SELECT CommentID FROM participant_accounts
             WHERE OneTimePassword=:key AND Status <> 'Complete'",
             array('key' => $this->key)
         );
@@ -94,8 +94,8 @@ class DirectDataEntryMainPage
         if (empty($this->TestName) && empty($this->CommentID)) {
             throw new Exception("Data has already been submitted.", 403);
         }
-        $pageNum = null;
 
+        $pageNum = null;
         if (!empty($_REQUEST['pageNum'])) {
             $pageNum = $_REQUEST['pageNum'];
         }
@@ -125,7 +125,7 @@ class DirectDataEntryMainPage
         $this->tpl_data  = array(
             'nextpage'    => $this->NextPageNum,
             'prevpage'    => $this->PrevPageNum,
-            'pageNum'     => $pageNum ? $pageNum + 1: 1,
+            'pageNum'     => $pageNum && is_numeric($pageNum) ? $pageNum + 1 : 1,
             'totalPages'  => $totalPages,
             'key'         => $this->key,
             'study_title' => $config->getSetting('title'),
@@ -146,6 +146,11 @@ class DirectDataEntryMainPage
         if ($currentPage === null) {
             return 1;
         }
+
+        if (!is_numeric($currentPage)) {
+            return -1;
+        }
+
         $nextPage = $currentPage+1;
         return intval(
             \Database::singleton()->pselectOne(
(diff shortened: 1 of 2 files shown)


AlexandraLivadas:
@laemtl I just checked for you, the build failed due to basic formatting issues. Here are the errors:
```
FILE: /home/travis/build/aces/Loris/htdocs/survey.php
--------------------------------------------------------------------------------
FOUND 1 ERROR AND 1 WARNING AFFECTING 2 LINES
--------------------------------------------------------------------------------
 364 | ERROR   | [x] Equals sign not aligned with surrounding assignments;
     |         |     expected 4 spaces but found 1 space
 368 | WARNING | [ ] Line exceeds 85 characters; contains 88 characters
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
Time: 43.55 secs; Memory: 22MB
```

laemtl:
@AlexandraLivadas Thank you! I still sometimes forgot to run the static checker. I will look for an automated solution pretty soon!

johnsaigle:
@laemtl If you run `git config core.hooksPath .githooks` in the LORIS directory, it will configure your project so that `make checkstatic` must complete successfully before you can git push. This will automate the process for you

laemtl:
@johnsaigle Thanks for this amazing tip!

htdocs/survey.php
```diff
@@ -94,14 +94,15 @@ function initialize()
...
-        $pageNum = null;
 
+        $pageNum = null;
         if (!empty($_REQUEST['pageNum'])) {
             $pageNum = $_REQUEST['pageNum'];
         }
 
         if ($pageNum === 'finalpage') {
             $this->Subtest = 'finalpage';
+            $pageNum       = -1;
```
driusan, line 105: I don't understand why you're making this change. It looks like it'll break the page numbering on the review page.

php/libraries/NDB_Caller.class.inc
```diff
@@ -141,7 +141,7 @@ class NDB_Caller
             );
 
             $this->instrument = $instrument;
-            header("Location: $redirectToOnSuccess");
+            $redirectToOnSuccess ? header("Location: $redirectToOnSuccess") : '';
```
driusan, line 144: Use if statements, don't abuse the ternary operator (in fact, it's best to avoid using it at all..)

htdocs/survey.php
laemtl: You're right, the change I did in [getNextPageNum()](https://github.com/aces/Loris/blob/a96e089e3d5f9f9b93341b98ea6cbb290acf8b55/htdocs/survey.php#L152) and [here](https://github.com/aces/Loris/blob/a96e089e3d5f9f9b93341b98ea6cbb290acf8b55/htdocs/survey.php#L129) makes that change not necessary anymore.

PR #6646, opened 2020-05-29
[document_repository] Permission to access module

### LINKED ISSUE #6373: [Doc Repo] Can't access the module with only the "Doc Repo delete" permission
These lines in the module need to be updated to also grant access to a user with the delete permission.

https://github.com/aces/Loris/blob/dde77f89cf3683955f0193265ed67aa02fa40b31/modules/document_repository/php/module.class.inc#L36-L40

I tested this and just having the "Delete files in Document Repository" does not seem to grant access to the module.

_Originally posted by @pierre-p-s in https://github.com/aces/Loris/pull/6331_

pierre-p-s:
## Brief summary of changes

- Changed the permission in module.class.inc to be able to access the module when you have only the permission to delete files
- When you only have permission to delete files, the tabs upload and category disappear (However, I was unable to forbid the user uploading a file or a category when they only have delete permission... Hiding the tabs should resolve it however I am not completely sure of it).

1. Check that if you only have delete permission, you can view the module.
2. Check that if you only have delete permission, you cannot access the upload and category tabs.

* Resolves #6373

### DIFF
--- a/modules/document_repository/jsx/docIndex.js
+++ b/modules/document_repository/jsx/docIndex.js
@@ -215,9 +215,46 @@ class DocIndex extends React.Component {
 
     let tabList = [
       {id: 'browse', label: 'Browse'},
-      {id: 'upload', label: 'Upload'},
-      {id: 'category', label: 'Category'},
     ];
+    let uploadDoc;
+    let uploadCategory;
+    if (loris.userHasPermission('document_repository_view')) {
+      tabList.push(
+        {
+          id: 'upload',
+          label: 'Upload',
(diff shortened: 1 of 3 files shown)


johnsaigle:
It looks like this is also changing some formatting in the index. Is this intentional?

pierre-p-s:
> It looks like this is also changing some formatting in the index. Is this intentional?

I was having issues with the permissions that were not stopping me from uploading a file so I hid the tabs for uploading files and creating category when you only have delete permission.

PR #5738, opened 2019-11-22
[ examiner ]  fix can't refresh page after Edit

kongtiaowang:
[ examiner ]  fix can't refresh page after Edit.
                      Header location doesn't work.
https://github.com/aces/Loris/issues/5735

PR #3359, opened 2017-12-20
Change certification validation to make date optional for inTraining and notCertified

PapillonMcGill:
This pull request check if new status is In Training or Not Certified before ensuring date check. Also change to insert/update to remove date value from query

See also: `https://redmine.cbrain.mcgill.ca/issues/13492`

PR #6647, opened 2020-05-30
[Examiner / Edit Examiner] fix various issues when establishing the Certification Status of the Examiner

### LINKED ISSUE #6605: [Examiner] Unexpected behavior of the Save and Reset buttons
**Describe the bug**
When adding a new _Examiner_, the _Certification Status_ for both instruments (_BMI_ and _Radiology Review_) need to be changed for a valid value (not _N/A_) or the action should fail when the button "_Save_" is clicked. Otherwise, an error is thrown and the _Change Log_ table should remain empty after clicking on "Reset", **but the invalid action updates the value of the _Change Log_ table, which should not have been modified**.

**To Reproduce**
1. Go to Clinial/Examiner
2. Add a new Examiner
3. Click on the new examiner's name to edit the new examiner
4. Click on the "Save" button, without filling anything. This will create an error, the page won't work with HTTP ERROR 500.
5. Go back to the previous page
6. Add a Certification Status for the instrument BMI (chose any status from the drop-down menu, except N/A)
7. Click on "Save": This will create an error, the page won't work with HTTP ERROR 500. (same error as in Step 4)
8. Go back to the previous page
9. Click on "Reset":  Notice the chosen Certification Status is now displayed in the "Change Log", which should not happen.

Step 4 and 7: There should be no HTTP ERROR 500. Instead, an error message should be displayed because the information for both instruments should be filled for to update to succeed.
Step 9: The filter should be reset with no values added to the Change Log. 

**Browser Environment:**
- OS: Ubuntu 16.04
- Browser: Chrome
- Version 23.0

racostas:
## Brief summary of changes
- Light modification of the function that handles the validation `function _validateEditExaminer($values)`. Cleaning an unused if condition.
- Light modification of the function that handles the process of the form `function _process($values)`
- Reestablishing old code accidentally replaced by https://github.com/aces/Loris/commit/058fa91d174e66f3e847f82a1bb7dc8408c3b112#diff-e429062f89aabd1a9ed7a99166cf2e85R195
- Properly function of buttons “Save” and “Reset”

1. Go to Clinical/Examiner.
2. Add a new Examiner.
3. Click on the new examiner's name to edit the new examiner.
4. Play around with the buttons "Save" and "Reset". The behaviour should be the intended one. The page should refresh after each action.
5. No error 500  should be shown when leaving one of the instruments as `certificated = N/A` for the first time. 
6. After changing from N/A to other estate for a given instrument will be not possible to go back to N/A and a message should be shown alerting this.

* Resolves #6605 
* Use the solution given in #5738 to handle the refresh part.

### DIFF
--- a/modules/examiner/php/editexaminer.class.inc
+++ b/modules/examiner/php/editexaminer.class.inc
@@ -265,10 +265,7 @@ class EditExaminer extends \NDB_Form
             }
         }
 
-        $baseurl = \NDB_Factory::singleton()->settings()->getBaseURL();
-        $url     = "{$baseurl}/examiner/editExaminer/?identifier="
-               . $examinerID;
-        header("Location: " . $url, true, 303);
+        header("Refresh:0");
     }
 
     /**
@@ -307,7 +304,7 @@ class EditExaminer extends \NDB_Form
             'maxYear'        => $config->getSetting('endYear'),
         );
         $statusOptions = array(
-            null            => 'N/A',
+            null            => '',
             'not_certified' => 'Not Certified',
             'in_training'   => 'In Training',
             'certified'     => 'Certified',
@@ -381,35 +378,26 @@ class EditExaminer extends \NDB_Form
      */
     function _validateEditExaminer($values)
     {
-        $DB     = \Database::singleton();
         $errors = array();
 
-        // check that there is both a status and a date (neither can be null)
+        $instruments = $this->getCertificationInstruments();
         foreach ($values['certStatus'] as $instrumentID => $certificationStatus) {
-            if (empty($certificationStatus)
-                || empty($values['date_cert'][$instrumentID])
-            ) {
-                if (($certificationStatus == "certified")
-                    && empty($values['date_cert'][$instrumentID])
-                ) {
-                    $errors['certStatus[' . $instrumentID .']'] = 'Both certification
-                        status and date must be filled out';
-                }
-            }
-        }
 
-        // check if previously recorded certification are all present
-        // (can not delete, only change status)
-        $rows = $DB->pselect(
-            "SELECT c.testID
-              FROM certification c
-              WHERE c.examinerID=:EID",
-            array('EID' => $this->identifier)
-        );
-        foreach ($rows as $row) {
-            if ($values['certStatus'][$row['testID']] == "") {
-                $errors['certStatus['.$row['testID'].']']
-                    = 'You can not delete a status';
+            // mandatory to choose an option for the Certification Status.
+            if (empty($certificationStatus)) {
+                $errors['certStatus[' . $instrumentID .']'] = 'Please choose
+                    a Certification Status for the Instrument: '
+                    . $instruments[$instrumentID];
+            }
+
+            // if the option chosen is "Certified"
+            // the Certification Day must be specified.
+            if (($certificationStatus == "certified")
+                && empty($values['date_cert'][$instrumentID])
+            ) {
+                    $errors['certStatus[' . $instrumentID .']'] = 'Please choose
+                    a Certification Date for the Instrument: '
+                    . $instruments[$instrumentID];
             }
         }
         return $errors;
--- a/modules/examiner/php/editexaminer.class.inc
+++ b/modules/examiner/php/editexaminer.class.inc
@@ -304,7 +304,7 @@ class EditExaminer extends \NDB_Form
             'maxYear'        => $config->getSetting('endYear'),
         );
         $statusOptions = array(
-            null            => '',
+            null            => 'N/A',
             'not_certified' => 'Not Certified',
             'in_training'   => 'In Training',
             'certified'     => 'Certified',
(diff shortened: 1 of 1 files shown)


modules/examiner/php/editexaminer.class.inc
```diff
@@ -265,10 +265,7 @@ class EditExaminer extends \NDB_Form
             }
         }
 
-        $baseurl = \NDB_Factory::singleton()->settings()->getBaseURL();
-        $url     = "{$baseurl}/examiner/editExaminer/?identifier="
-               . $examinerID;
-        header("Location: " . $url, true, 303);
+        header("Refresh:0");
```
driusan, line 268: Why is this changed? Neither usage of `header` is really good practice, but responding to a post request with a 303/location is still good practice..

racostas: Hi @driusan, I give here the same solution given in PR #5738 which seem to be merged into `aces/master` already.  The code in red (removed) was not allowing the save button to refresh once clicked.
image

driusan:
@ridz1208 can you review? You know the  examiner/certification design better than I do..

modules/examiner/php/editexaminer.class.inc
```diff
@@ -381,35 +378,26 @@ class EditExaminer extends \NDB_Form
...
-        $rows = $DB->pselect(
-            "SELECT c.testID
-              FROM certification c
-              WHERE c.examinerID=:EID",
-            array('EID' => $this->identifier)
-        );
-        foreach ($rows as $row) {
-            if ($values['certStatus'][$row['testID']] == "") {
-                $errors['certStatus['.$row['testID'].']']
-                    = 'You can not delete a status';
```
jesscall, line 412: Removing this allows users to change status from certified > anything else. Not sure if this is intended behaviour?

racostas: Hi @jesscall, 

What I understand from this removed code is that is not allowing to change from `ANY state different from BLANK(a sort of null) to BLANK.` Not precisely preventing to change from `CERTIFIED TO ANYTHING ELSE.`

If that the case, is already covered in the new lines 386 to 390:
```
            // mandatory to choose an option for the Certification Status.
	    if (empty($certificationStatus)) {
                $errors['certStatus[' . $instrumentID .']'] = 'Please choose
                a Certification Status for the Instrument: '
                . $instruments[$instrumentID];
```

ridz1208: @racostas as far as i understand it, these statements are not equivalent. Your new code will prevent saving the form with any of the certifications empty !! which means you would not be able to partially certify an examiner (that's not good)

the old code looked in the existing certification table and made sure that IF you were previously certified for a test, you cannot go back to `""` which means you cannot go from a certification to nothing, but you can still save with empty certifications if they were never set beforehand

racostas: Hi @ridz1208, 

Yes, this is totally the case. Completely agree. The previous code if I don't remember bad was preventing also that null values were saved even for first time in the database (maybe not intended). That could be not the expected behaviour and that's why I put a note on the description of this PR regarding the N/A values and null values.

I limited my fixing so far (waiting for comment or discussion on this other aspect) to grant just the same functionalities available before. The available certification status are (without taking N/A or null into account): Not Certified, In Training and Certified. N/A (or null, I'm more inclined to N/A) will be useful in a case we don't know the state, and could be valuable in my opinion.

I'm not completely sure if we want to prevent going from previously certificated to N/A in the case we decide to include N/A or null as an option. What if the "data entry" person assign one of the available status: Not Certified, In Training and Certified and after saving he/she realises made a mistake and assigned a status to the examiner but when corroborating realises didn't have information about it? In this case the user will not be able to rollback to N/A and will have to stick to any of the other three status (when not exactly the reality from the business perspective).

I'm very willing to implement the solution we decide fix best the user case.

ridz1208: @driusan do you know the reason why we never allowed unsetting the certification status ?

driusan: no, maybe @zaliqarosli does?

ridz1208: I think this predates Zaliqa by alot. I'll message Leigh

zaliqarosli: looks like that section was added in this PR https://github.com/aces/Loris/pull/3359/files. I commented on it saying that I liked the added check, but I don't remember the reason I was for it

ridz1208:
@racostas the errors described in the issue were caused by https://github.com/aces/Loris/commit/058fa91d174e66f3e847f82a1bb7dc8408c3b112#diff-e429062f89aabd1a9ed7a99166cf2e85R195

@johnsaigle is there a need for this newly added exception ?

If not, the exception should be removed and the old code restored without changing functionality

ridz1208:
@racostas 

@driusan said : "the right way to fix that phan error without changing behaviour should be to initialize them to null before the if statement so that they’re set even if the if doesn’t get triggered (or similar)"

driusan:
Re: 1. I think you also need to initialize the variables to `null` before the if statement to prevent the phan error that the previous commit was trying to solve.

ridz1208:
@racostas this is still not gonna work, see the changes in the PR that broke the module. an exception was added on lines 197+ that should be removed and the old `if` statement that was there before should be restored

racostas:
> @racostas this is still not gonna work, see the changes in the PR that broke the module. an exception was added on lines 197+ that should be removed and the old `if` statement that was there before should be restored

@ridz1208. I put the code in the exactly same state is was before my changes only leaving the solution given in #5738 and erasing an if condition it wasn't in use.

ridz1208:
@racostas I thin the changes made are fine. I just dont think they will fix the 500 errors from the linked issue.

driusan:
@racostas I think the code the way it was before your PR still has the 500 error because of the exception that @ridz1208 is talking about.

racostas:
> @racostas I think the code the way it was before your PR still has the 500 error because of the exception that @ridz1208 is talking about.

I'm manually testing and so far think no to get the 500 error. But no sure. Will do more extensive testing.

racostas:
> > @racostas I think the code the way it was before your PR still has the 500 error because of the exception that @ridz1208 is talking about.
> 
> I'm manually testing and so far think no to get the 500 error. But no sure. Will do more extensive testing.

OK. The 500 error seems to get in place the first time we assign N/A to a subject.

driusan:
@racostas are you going to be able to fix the exception today?

racostas:
> @racostas are you going to be able to fix the exception today?

@driusan. I think is ready in my VM now. Travis is taking a while to finish Today probably due to overload. Do you think it is save to stop it for this PR now and push new changes? May I do it by myself? I don't want to introduce more variables to the issue.

driusan:
If they're ready just push to the branch (make sure you run `make checkstatic` first to get the low hanging fruit..), Travis will queue up even if this one's not finished running.

racostas:
> If they're ready just push to the branch (make sure you run `make checkstatic` first to get the low hanging fruit..), Travis will queue up even if this one's not finished running.

It seem that post and commits landed in different order to GitHub but after done what suggested in this comment the latest changes should be on the commit afa59cd

ridz1208:
@CamilleBeau @jesscall can one of you test this one ASAP ?

jesscall:
Okay, tested and looks good in terms of not throwing 500 errors. 

**Functionality that was preserved:**
- Only need to enter date if status = `certified` 
- cannot delete old data
- comments only save when status is entered

**Functionality that has changed:** (unsure if this is intentional or not)
- You do not need to enter certification status' for each item in order to save the page. Previously you needed to have a status selected for both BMI and radiology review. Now you can save the form if only one is filled out.

ridz1208:
@jesscall yeah this is intended functionality. You should be able to certify examiners for as many instruments as you need without selecting all of them

so all good
@driusan urs to review

driusan:
@racostas can you update the PR description? This has changed enough that I'm not sure which parts are still accurate and which are outdated for the commit message.

racostas:
> 
> 
> @racostas can you update the PR description? This has changed enough that I'm not sure which parts are still accurate and which are outdated for the commit message.

Sounds good. Done.

PR #6649, opened 2020-06-01
[Statistics / Behavioural] fix query by project.

### LINKED ISSUE #6573: [Statistics / Behavioural Statistics] query by project not working.
**Describe the bug**
In `Reports-> Statistics -> Behavioural Statistics`, query by project seems not to by working.

**To Reproduce**
1. Go to `Reports-> Statistics -> Behavioural Statistics`
2. Choose a project on the project's comboBox.
3. Click on Submit Query
4. **Issue:** The results table is not updated except for "All projects" or "DCP" choices

The results table updated accordingly.

- OS: Windows
- Browser: Firefox
- Version: 76

racostas:
## Brief summary of changes

- fixes some bugs found in the code.
- adds permission restriction over the projects the user is able to query 

The query by project should be working now for `Reports->Statistics->Behavioural Statistics`

1. Go to `Reports->Statistics->Behavioural Statistics`
2. Select different projects on the `Projects ComboBox ` and click on `Submit Query`
3. The properly information should be retrieved.

#### Link to related issue

* Resolves #6573


racostas:
close and reopen to see if Travis is launch.

racostas:
close and reopen to see if Travis is launch.

modules/statistics/php/stats_behavioural.class.inc
```diff
@@ -157,7 +172,10 @@ class Stats_Behavioural extends \NDB_Form
                     AND s.Current_stage <> 'Recycling Bin'
                     AND f.CommentID NOT LIKE 'DDE%'
                     AND s.CenterID <> '1'
-	            AND s.CenterID IN (" . $sitesString . ")
+                    AND s.CenterID IN
+                        (" . implode(',', $paramCenters) . ")
+                    AND c.RegistrationProjectID IN
```
driusan, line 177: shouldn't this be `s.ProjectID`? (@ridz1208, maybe you have input?)

This also looks like it'll break with an SQL error if the `$paramCenters` or `$paramProjects` is empty, but the old code looked like that would happen (for sites at least) too

modules/statistics/php/stats_behavioural.class.inc
```diff
@@ -230,7 +248,10 @@ class Stats_Behavioural extends \NDB_Form
                     AND f.CommentID LIKE 'DDE%'
                     AND c.Active = 'Y'
                     AND s.CenterID <> '1'
-                    AND s.CenterID IN(" . $sitesString . ")
+                    AND s.CenterID IN 
+                        (" . implode(',', $paramCenters) . ")
+                    AND c.RegistrationProjectID IN
```
driusan, line 253: same comment as above

modules/statistics/php/stats_behavioural.class.inc
racostas: regarding to the use `s.ProjectID` or `c.RegistrationProjectID` my knowledge of this part of the business is not the best, but session (s) makes more sense to me right now so `s.ProjectID` should by the correct one.

regarding to the possibility of a SQL break, while not to much probable It is possible and this is an error in the code which it not taking this alternative into consideration. In this case we're talking about a user(logged user) that doesn't have any project assigned (and the analogous for the sites) [while mandatory during the user creation it could happen, or not... not sure.. but agree it is not a good code]: In this case I think the correct should be to show some kind of message alerting there is not data accessible to the user and the reason (not belonging to any project (or site). ?

modules/statistics/php/stats_behavioural.class.inc
```diff
@@ -239,13 +260,13 @@ class Stats_Behavioural extends \NDB_Form
 
         foreach ($result as $row) {
             $center = $row['CenterID'];
-            $vl     = intval($row['VLabel']);
+            $vl     = $row['VLabel'];
             $c      =& $this->tpl_data['dde']['C' . $center];
 
             if (isset($c[$vl]['total'])) {
-                $c[$vl]['total'] += $row['val'];
+                $c[$vl]['total'] += intval($row['val']);
```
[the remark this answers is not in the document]
racostas: I think we're mostly counting and I think that's why it makes sense to be an integer value but again my knowledge of the module is not the best.

modules/statistics/php/stats_behavioural.class.inc
racostas: @driusan, @ridz1208:

**IMPORTANT:**  Before this fixes the module was showing 500 error. So this could be critical to release. While `s.ProjectID `or `c.RegistrationProjectID` is a simple change, the SQL issue could be a bit more elaborated. Again this is not to probably to happen... but no sure if there is cases it could. What do you think we should do in any case?

racostas:
Until we take a decision on this one I'm sending #6724 which could be critical to 23.0-release. Aside of the projects restrictions the present was addressing there were other issues causing a 500 error that now are addressed by #6724. #6724 takes care only of the 500 error and leave aside the projects and sites restrictions which might  be not be critical to the release.

driusan:
What decision needs to be made? I think the query just needs to be fixed.

racostas:
> What decision needs to be made? I think the query just needs to be fixed.

What we'll show to the user in the case there are not project assigned to it. Before this PR we were showing all the projects and project's stats to him in a case like this (including in the comboBox for choosing it). Will show a comboBox blank? And a table blank? All the projects the way it was before the PR?
My suggestion  is the following:

> In this case I think the correct should be to show some kind of message alerting there is not data accessible to the user and the reason (not belonging to any project (or site). ?

driusan:
I think a message and reason is a good solution too.

racostas:
> I think a message and reason is a good solution too.

This are not minor changes at least with my understating of the module right now. That's why I issue the PR #6724 which address a 500 error not related with the projects or sites restriction but what seems to be a bug.

driusan:
@racostas how long do you think it would take to do the "non-minor" changes? Can this be done for, say, 23.0.2?

racostas:
> @racostas how long do you think it would take to do the "non-minor" changes? Can this be done for, say, 23.0.2?

It will be a pleasure. When is 23.0.2 intended to be released?

driusan:
After 23.0.1.. there's no time frame, I just mean is it possible to do in the short-medium time frame if it's not possible to in time for the next bug fix release.

ridz1208:
@racostas any news on this ?

driusan:
@racostas is this ready for review with the latest commit?

driusan:
@maltheism this was assigned to you in Cecile's email to review if it's ready for review.. looks like it's ready.

maltheism:
@driusan I'm unsure why the needs work tag still exists on this PR.
My manual testing worked for a new user with only permissions for site Montreal.

driusan:
@maltheism did you review it as well?

maltheism:
@driusan yes

PR #6663, opened 2020-06-03
[API] 2020-05-25 Add api testsuite

spell00:
Creation of a test suite for the API


johnsaigle:
This is failing Travis because of code formatting issues. If you run `make checkstatic` in the LORIS root directory you'll be able to see locally what the issues are.

modules/api/test/ApiTest.php
```diff
@@ -0,0 +1,392 @@
...
+    private $_headers;
+    private $_base_uri;
+
+    /**
+     * @var string[]
+     */
+
+    private function _login()
+    {
+        $this->_base_uri = 'https://test-loris-dev.loris.ca/api/v0.0.3/';
```
kongtiaowang, line 57: What if this VM is down or the host name changed.

modules/api/test/ApiTest.php
```diff
@@ -0,0 +1,392 @@
+<?php declare(strict_types=1);
+/**
```
johnsaigle, line 2: This part can be deleted. We used to have an automatic test rule to make file comments in PHP code mandatory but it's not anymore.

modules/api/test/ApiTest.php
```diff
@@ -0,0 +1,392 @@
...
+
+    /**
+     * @var string[]
+     */
+
+    private function _login()
+    {
+        $this->_base_uri = 'https://test-loris-dev.loris.ca/api/v0.0.3/';
+        $this->_client = new \GuzzleHttp\Client(['base_uri' => $this->_base_uri]);
+        $response = $this->_client->request('POST', 'https://test-loris-dev.loris.ca/api/v0.0.3/login', ['json' => ['username' => 'admin', 'password' => 'LORISitb100%']]);
```
johnsaigle, line 59:
```suggestion
        $response = $this->_client->request('POST', 'https://test-loris-dev.loris.ca/api/v0.0.3/login', ['json' => ['username' => 'admin', 'password' => 'LORISitb100%']]);
```

I don't think we want the password to this VM to be publicly exposed. As @kongtiaowang pointed out it's not necessarily stable. 

I'm trying to think whether there's an easy solution here. If there is a way to do this within the server Travis creates instead, that would be good. If not then it might be worth making a point in the LORIS meeting about how to approach API testing and what server to test on.

kongtiaowang: The easist way is that using integration test on Travis directly.
https://web:8000/api/v0.0.3/...

modules/api/test/ApiTest.php
```diff
@@ -0,0 +1,388 @@
...
+ * @package    Tests
+ * @subpackage Login
+ * @author     Simon Pelletier <simon.pelletier@mcin.ca>
+ * @license    http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link       https://www.github.com/aces/Loris/
+ */
+
+
+namespace LORIS\api\Test;
+ini_set('memory_limit', '1024M');
```
xlecours, line 17: This can be removed.
The testsuite will fail if something is wrong with memory limit and we will look at the issue globaly.

modules/api/test/ApiTest.php
```diff
@@ -0,0 +1,388 @@
+<?php declare(strict_types=1);
+/**
+ * This contains tests relevant to the API
+ *
+ * PHP Version 7
+ *
+ * @category   API
+ * @package    Tests
+ * @subpackage Login
```
xlecours, line 9: Change Login for integration maybe

modules/api/test/ApiTest.php
```diff
@@ -0,0 +1,388 @@
+<?php declare(strict_types=1);
+/**
+ * This contains tests relevant to the API
```
xlecours, line 3: Better description of the goal of the script. 
Something that explain what it does: `This script sends HTTP request to every enpoints of the api module and look at the response content, status code and headers where it applies.`

modules/api/test/ApiTest.php
```diff
@@ -0,0 +1,388 @@
...
+ * @author     Simon Pelletier <simon.pelletier@mcin.ca>
+ * @license    http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link       https://www.github.com/aces/Loris/
+ */
+
+
+namespace LORIS\api\Test;
+ini_set('memory_limit', '1024M');
+
+require './vendor/autoload.php';
```
xlecours, line 19: This is not a stand alone script so you don't need to do that. It will be called by phpunit which is calling test/bootstrap.php which as that line (the require ... autoloader.php)

What is missing though is :
`require_once __DIR__ . "/../../../test/integrationtests/LorisIntegrationTest.class.inc";`

Because you want this class to extend LorisIntegrationTest

modules/api/test/ApiTest.php
```diff
@@ -0,0 +1,388 @@
...
+
+use PHPUnit\Framework\TestCase;
```
xlecours, line 21: This can be removed if you require LorisIntegrationTest.class.inc

modules/api/test/ApiTest.php
```diff
@@ -0,0 +1,388 @@
...
+use Laminas\Diactoros\ServerRequest;
```
xlecours, line 22: You should be using Laminas\Diactoros\Request because we are on the client side.

modules/api/test/ApiTest.php
xlecours: ~@kongtiaowang What is the baseurl when the server is run in TravisCI?~
nerver mind. With LorisIntegrationTest.class.inc we just need to use`$this->url`

xlecours: @spell00 
`$this->_base_uri = $this->url . '/api/v0.0.3/' ` should work

modules/api/test/ApiTest.php
```diff
@@ -0,0 +1,388 @@
...
+
+    /**
+     * @var string[]
+     */
+
+    private function _login()
+    {
+        $this->_base_uri = 'https://test-loris-dev.loris.ca/api/v0.0.3/';
+        $this->_client = new \GuzzleHttp\Client(['base_uri' => $this->_base_uri]);
+        $response = $this->_client->request('POST', LORIS_URL, ['json' => ['username' => LORIS_USERNAME, 'password' => LORIS_PASSWORD]]);
```
xlecours, line 55: What is LORIS_URL?

xlecours: LORIS_USERNAME should be `UnitTester` and LORIS_PASSWORD should be `$this->validPassword`
Right @kongtiaowang ?

modules/api/test/ApiTest.php
```diff
@@ -0,0 +1,388 @@
...
+        $token = $response->getBody()->getContents();
+        $token = substr(json_decode((string)json_encode($token), true), 10, -2);
```
xlecours, line 57: `$response->getBody()->getContents()` will return a string. You could also do `(string) $response->getBody()`
doing both is "redondant". 
`json_decode($response->getBody()->getContents())->token ?? null` ou `json_decode($response->getBody()->getContents(),true)['token'] ?? null` should do it. (instread of using substr)

spell00:
This branch is to be deleted. The new branch is #6671.

christinerogers:
@spell00 important to mention -- Are @xlecours 's outstanding comments (above) covered in #6671 already, or will they be in your next commit ?

spell00:
Yes, all comments, including @xlecours 's comments above, are covered in #6671

johnsaigle:
Screen Shot 2020-06-08 at 11 08 52

@spell00 I think something may have gone wrong in the commit. 2000+ files seems like a lot

christinerogers:
fyi @johnsaigle you're seeing the commit necessary to rollback an unfortunate commit

PR #6593, opened 2020-05-22
add nii file type to the list of file types to show in the query

cmadjar:
## Brief summary of changes

This fixes projects that only inserts NIfTI files and not MINC files into the database. The query in the imagingbrowserrowprovisioner.class.inc file was limiting the query to `filettype='mnc'` only while we should also allow for `filetype='nii'`.

#### Testing instructions (if applicable)

1. Ensure the imaging browser filter menu still loads.

#### Link(s) to related issue(s)

no issues reported but a user of LORIS loading only BIDS dataset (aka NIfTI file types) could not see anything on the imaging browser menu filter page as there are no MINC files in their database, only NIfTI files.

PR #6668, opened 2020-06-03
[Imaging browser] Fix loading error for t1-defaced and t2-defaced scan types

### LINKED ISSUE #6616: [imaging browser] Adding some scan types renders the module unloadable
**Additional context**
This fails step 7 of the "main page" section of the test plan.
 
**Describe the bug**
When adding a tabulated scan type, selecting "t1-defaced" or "t2-defaced" and saving makes the module return an error when trying to load it 

**To Reproduce**
1. Navigate to Admin>Configuration>Imaging Modules
2. Click on the "Add field" button under "Imaging Tabulated Scan Types" 
3. Select "t1-defaced" or "t2-defaced" 
3. Click on the Submit button.
4. Navigate to Imaging>Imaging Browser
5. Observe loading error

To observe the front page loading properly with the relevant columns.

- OS: Ubuntu 18.04 LTS
- Browser:
    firefox 76.0.1
    Chrome 83.0.4103.61

This bug was observed on the Testing VM

h-karim:
Adding 't1-defaced' or 't2-defaced' scan types to the imaging browser was causing an SQL error due to the hyphen in the name.

#### Brief summary of changes

- ~I replaced all hyphen occurrences in the array containing the scan type names with underscores. This does not change how the scan type name appears on the frontend.~ 
- I used Database->quote() to quote scan type names, or surrounded them with backticks if they needed to be interpreted as identifiers rather than string literals.

1. Go to Admin>Configuration>Imaging types>
2. Add t1-defaced and/or t2-defaced scan types next to the tabulated scan types and save
3. Go to the imaging module front page
4. Observe the module loading properly, with t1-defaced/t2-defaced QC columns

* Resolves #6616

### DIFF
--- a/modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
+++ b/modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
@@ -112,7 +112,7 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
         $coalesce_desc = array();
 
         foreach ($scan_id_types as $id => $type) {
-            $pass[$id]          = $DB->quote($type);
+            $pass[$id]          = $DB->escape($type);
             $qc[$id]            = $DB->quote($type);
             $coalesce_desc[$id] = $DB->quote($pass[$id] . '.' . $qc[$id]);
             $case_desc[$id]     = "
@@ -155,8 +155,8 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
               JOIN files_qcstatus USING (FileID)
               WHERE files.AcquisitionProtocolID= $key
                 AND files_qcstatus.QCStatus IN (1, 2)
-              GROUP BY files.SessionID) `$pass[$key]`
-                ON (`$pass[$key]`.SessionID=f.SessionID
+              GROUP BY files.SessionID) $pass[$key]
+                ON ($pass[$key].SessionID=f.SessionID
             ) ";
         }
 
@@ -210,7 +210,7 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
               $left_joins
             WHERE
               s.Active = 'Y' AND
-              f.FileType='mnc'
+              f.FileType IN ('mnc', 'nii')
             GROUP BY s.ID
             ORDER BY c.PSCID, s.Visit_label
             ",
--- a/modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
+++ b/modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
@@ -33,10 +33,13 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
     /**
      * Create an ImagingBrowserRowProvisioner, which gets rows for
      * the imaging browser menu table.
+     *
+     * @throws \ConfigurationException
      */
     function __construct()
     {
         $config = \NDB_Config::singleton();
+        $DB     = \NDB_Factory::singleton()->database();
 
         // ===========================================================
         // Get the different scan types to be displayed in data table
@@ -92,7 +95,7 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
         }
 
         $NewDataSubquery = "
-            CASE 
+            CASE
                 COALESCE(Max(fqc.QCLastChangeTime), 'new')
                 WHEN 'new' THEN {$acqpif}
                 WHEN ''    THEN {$acqpif}
@@ -109,9 +112,9 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
         $coalesce_desc = array();
 
         foreach ($scan_id_types as $id => $type) {
-            $pass[$id]          = $type . 'pass';
-            $qc[$id]            = $type . 'QC';
-            $coalesce_desc[$id] = $pass[$id] . '.' . $qc[$id];
+            $pass[$id]          = $DB->quote($type);
+            $qc[$id]            = $DB->quote($type);
+            $coalesce_desc[$id] = $DB->quote($pass[$id] . '.' . $qc[$id]);
             $case_desc[$id]     = "
                 CASE
                     COALESCE($coalesce_desc[$id], '')
@@ -126,15 +129,15 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
                 WHEN 'Fail' THEN
                     IF(s.MRIQCPending='Y', 'Pending Fail', 'Fail')
                 WHEN 'Pass' THEN
-                    IF(s.MRIQCPending='Y', 'Pending Pass', 'Pass') 
+                    IF(s.MRIQCPending='Y', 'Pending Pass', 'Pass')
                 ELSE s.MRIQCStatus
-            END 
+            END
             ";
 
         $PendingNewquery = "
-            CASE 
+            CASE
                 WHEN s.MRIQCPending='Y' THEN 'P'
-                WHEN MAX(fqc.QCFirstChangeTime) IS NULL $newQueryCase THEN  'N'      
+                WHEN MAX(fqc.QCFirstChangeTime) IS NULL $newQueryCase THEN  'N'
             END";
 
         // ====================================================
@@ -146,14 +149,14 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
         foreach ($case_desc as $key => $value) {
             $left_joins .= "
             LEFT JOIN (
-              SELECT files.SessionID, 
-                MIN(files_qcstatus.QCStatus+0) as $qc[$key] 
-              FROM files 
-              JOIN files_qcstatus USING (FileID) 
-              WHERE files.AcquisitionProtocolID= $key 
-                AND files_qcstatus.QCStatus IN (1, 2) 
-              GROUP BY files.SessionID) $pass[$key] 
-                ON ($pass[$key].SessionID=f.SessionID
+              SELECT files.SessionID,
+                MIN(files_qcstatus.QCStatus+0) as $qc[$key]
+              FROM files
+              JOIN files_qcstatus USING (FileID)
+              WHERE files.AcquisitionProtocolID= $key
+                AND files_qcstatus.QCStatus IN (1, 2)
+              GROUP BY files.SessionID) `$pass[$key]`
+                ON (`$pass[$key]`.SessionID=f.SessionID
             ) ";
         }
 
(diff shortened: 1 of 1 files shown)


modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
```diff
@@ -43,7 +43,7 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
         // ===========================================================
 
         // Grep all the different scan types present in mri_scan_type
-        $all_scan_types = \Utility::getScanTypeList();
+        $all_scan_types = str_replace('-', '_', \Utility::getScanTypeList());
```
driusan, line 46: There's a Database->quote function for quoting field (or table) names which may have special characters.

h-karim: This was the initial strategy, however I found when using it indiscriminately on all scan types (eg "t1", which didn't have any issues before), it introduced an SQL syntax error for names without any special characters:
> Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''t1pass'

I'll have to implement it with a `preg_match` for special SQL characters.

h-karim: @driusan I just pushed a commit where I added Database->quote calls where needed. However where the names with special characters are used as parts of column names / identifiers, Database->quote won't work because they'll be interpreted as string literals. For those cases I had to use backticks \`  \`. Wouldn't the previous approach (replacing hyphens with underscores) be better since it introduces much less code change?

ridz1208: @h-karim The quoting is better practice since you can't always control what users input as fields in a table. the not-so-ideal issue is that we are using field values as column names in another table... that's not great but it's not a problem solvable in this PR.

as far as the quote not working, I think it's better if you terminate the string, add the quote and  re-concatenate the string instead. I made a suggestion

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
```diff
@@ -166,7 +171,7 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
         // $scan_types are set in the configuration module
         $modalities_subquery = '';
         foreach ($case_desc as $key => $value) {
-            $modalities_subquery = "$value as $scan_id_types[$key]_QC_Status";
+            $modalities_subquery = "$value as `$scan_id_types[$key]_QC_Status`";
```
ridz1208, line 174:
```suggestion
            $modalities_subquery = $value." as ".$instance->quote("$scan_id_types[$key]_QC_Status");
```

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
```diff
@@ -36,8 +36,13 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
      */
     function __construct()
     {
-        $config = \NDB_Config::singleton();
-
+        $config   = \NDB_Config::singleton();
+        $instance = null;
+        try {
+            $instance = \NDB_Factory::singleton()->database();
```
ridz1208, line 42: $instance is very non-specific here. your better off using `$DB` or `$db` instead

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
h-karim: You're right, on L174 this will work, it's my fault for missing it, however trying to use quote() on https://github.com/aces/Loris/blob/23.0-release/modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc#L155

Like this:

```php
GROUP BY files.SessionID) ". $instance->quote("$pass[$key] ") .
```
or 
```php
GROUP BY files.SessionID) ". $instance->quote($pass[$key] ) .
```
won't work, as it returns the following error:
>Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '**'t1pass '** ON (\`t1pass\`.SessionID=f.SessionID\n            ) \n            LEFT JOI' at line 76 

and on [L156](https://github.com/aces/Loris/blob/23.0-release/modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc#L156), we need to use backticks on ``" ON ( `$pass[$key]`.SessionID=f.SessionID`` otherwise `$pass[$key]` will be interpreted as a string literal instead of identifier, and the comparison will not work.

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
```diff
@@ -37,7 +37,12 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
     function __construct()
     {
         $config = \NDB_Config::singleton();
-
+        $DB     = null;
+        try {
+            $DB = \NDB_Factory::singleton()->database();
```
driusan, line 42: This try/catch should not be here. If an exception is thrown something is seriously wrong enough that it should be propagated.

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
```diff
@@ -146,14 +151,14 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
...
-                AND files_qcstatus.QCStatus IN (1, 2) 
-              GROUP BY files.SessionID) $pass[$key] 
-                ON ($pass[$key].SessionID=f.SessionID
+              SELECT files.SessionID,
+                MIN(files_qcstatus.QCStatus+0) as " . $DB->quote($qc[$key]) ." 
+              FROM files
+              JOIN files_qcstatus USING (FileID)
+              WHERE files.AcquisitionProtocolID= $key
+                AND files_qcstatus.QCStatus IN (1, 2)
+              GROUP BY files.SessionID) `$pass[$key]`
```
driusan, line 160: Why isn't this using $DB->quote like the other places? Simply adding "`" isn't robust. For instance, it will still error out if the value has a "`" in it.

h-karim: I pushed a commit to call quote() at the top of the file for a cleaner look (L115 for the pass array). This makes all the elements inside the array quoted, let me know if this works. 

We still need the backticks however, removing them makes `$pass[$key]` get interpreted as a literal instead of an identifier, and spits out an error:

>Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '''t2-defacedpass')'\n                ON (`'t2-defacedpass'`.SessionID=f.SessionID' at line 69

christinerogers:
@cmadjar just to check- is this a valid issue we should be trying hard to solve?  --> Enabling `t1-defaced` and `t2-defaced` as types to be displayed as valid QC Status columns in the ImgBrowser data table. 

Again @h-karim I'm not sure that the -defaced columns are the practical/desirable to show in this context.

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
```diff
@@ -146,14 +149,14 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
...
-              GROUP BY files.SessionID) $pass[$key] 
-                ON ($pass[$key].SessionID=f.SessionID
+              SELECT files.SessionID,
+                MIN(files_qcstatus.QCStatus+0) as $qc[$key]
+              FROM files
+              JOIN files_qcstatus USING (FileID)
+              WHERE files.AcquisitionProtocolID= $key
+                AND files_qcstatus.QCStatus IN (1, 2)
+              GROUP BY files.SessionID) `$pass[$key]`
+                ON (`$pass[$key]`.SessionID=f.SessionID
```
cmadjar, line 159: Couldn't it be re-written like that instead of using backtic?

```suggestion
              GROUP BY files.SessionID) " . $DB->quote($pass[$key]) 
              . "ON (" . $DB->quote($pass[$key]) . ".SessionID=f.SessionID
```

cmadjar:
In general, scan types could have `-` characters, regardless of defaced or not. Open science projects will only have `*-defaced` scan types for anatomicals since only defaced data can be present on the LORIS instance. 

It would be nice if it can get in the release. If not it could always goes into bug fix.

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
h-karim: when I tried this, it returned the following error:

>[Fri Jun 05 09:51:14.495832 2020] [php7:warn] [pid 24948] [client ::1:56090] PHP Warning:  PDOStatement::execute(): SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''t1'ON ('t1'.SessionID=f.SessionID\n            ) \n            LEFT JOIN (\n      ' at line 73 in /var/www/loris/src/Data/Provisioners/DBRowProvisioner.php

I think it's because the single quotes without around a word without any back tics wrapped gets it to be interpreted as a string literal instead of column/table name. 
Also, in the latest commit, the single quotes are used on L115: https://github.com/aces/Loris/pull/6668/files#diff-246d74737579a11922903fcadb7015ffL112-R117 , so the back tics would be wrapped around the table name resulting in `` `'t1pass'` `` for example which would escape any special characters as well as for it to be interpreted as a table name.

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
driusan: sorry to have sent you down the wrong path.. `$DB->quote` is the wrong function, it's for quoting strings. It should be `$DB->escape` to escape table or column names.

```
   /**
     * Escapes a string properly for mysql and appends and prepends backticks.
     * Any backticks you append or prepend will be escaped.
     *
     * @param string $tableName The column or table name that needs to be escaped
     *
     * @return string surrounded by backticks and with special characters escaped.
     */
```

(luckily, it should be a simple change to change the places you called quote on a table/column name to escape..)

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
driusan: `quote` was the wrong function (as per my other comment), it should be `escape` to escape table/column names.

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
```diff
@@ -189,24 +193,24 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
...
-              LEFT JOIN mri_scan_type modality ON 
+              LEFT JOIN mri_scan_type modality ON
                 (f.AcquisitionProtocolID=modality.ID)
-              $left_joins 
-            WHERE 
+              $left_joins
+            WHERE
               s.Active = 'Y' AND
-              f.FileType IN ('mnc', 'nii')
+              f.FileType='mnc'
```
driusan, line 213: This is a regression of #6593

h-karim:
Update on everything: Database->quote is used to escape special characters for the elements of the array `qc` and the array `coalesc_desc` because the elements in those arrays are used as strings. Database->escape is used for the elements of the array `pass` because those elements are used as identifiers.

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
h-karim: Sorry, I don't know how this slipped, I amended the latest commit to change that line to how it was.

PR #6719, opened 2020-06-10
[configuration] Add date_format as DataType in ConfigSettings

### LINKED ISSUE #6678: [candidate_parameters] Date format does not match dates specified in config
**Describe the bug**
Altering the _DOB format_ or the _DOD format_ fields in the configuration module do not change how the dates are displayed / stored on the front-end or back-end for this module. The TestPlan for this module specifies that it should : 
> 35. Check that the date stored matches the DOB Format specified in the configuration module.

> 40. Check that the date stored matches the DOD Format specified in the configuration module.

**To Reproduce**
1. From admin user, modify the values for _DOD format_ and _DOB format_ in the config module.
2. Navigate to a candidate's profile on the candidate parameters module and see that the dates remain the same.
3. Get the information in the back-end from the `candidate` table and see that the dates are stored in the same way.

The format of the dates to change in the `candidate_parameters` module.

jesscall:
## Brief summary of changes
Quick fix for candidate parameter tabs `Date of Birth` and `Date of Death`. Altering the DOB format or the DOD format fields in the configuration module to either `YM` or `Ym` should result in the dates in cand param being saved with `YYYY-MM-01` as the day of the month rather than whatever the user entered. 

**Edit:**
This PR now adds a new dataType to ConfigSettings: `date_format`. Fields such as DOB and DOD format will now have a dropdown in which users that select what format they prefer. This should also resolve the ambiguity in candidate parameters that is discussed in the thread below.

1. See test plan steps 35 & 40

* Resolves #6678

### DIFF
--- a/SQL/New_patches/2020_06_16_Add_Date_Format_to_ConfigSettings_DataType.sql
+++ b/SQL/New_patches/2020_06_16_Add_Date_Format_to_ConfigSettings_DataType.sql
@@ -1,0 +1,11 @@
+-- Add date_format option to DataType
+ALTER TABLE ConfigSettings
+	MODIFY COLUMN `DataType` enum('text','boolean','email','instrument','textarea','scan_type','date_format','lookup_center','path','web_path') DEFAULT NULL;
+
+-- Update DataType of dobFormat
+UPDATE ConfigSettings
+	SET DataType='date_format' WHERE Name='dobFormat';
+
+-- Update DataType of dodFormat
+UPDATE ConfigSettings
+	SET DataType='date_format' WHERE Name='dodFormat';
--- a/SQL/New_patches/2020_06_16_Add_Date_Format_to_ConfigSettings_DataType.sql
+++ b/SQL/New_patches/2020_06_16_Add_Date_Format_to_ConfigSettings_DataType.sql
@@ -9,3 +9,20 @@ UPDATE ConfigSettings
 -- Update DataType of dodFormat
 UPDATE ConfigSettings
 	SET DataType='date_format' WHERE Name='dodFormat';
+
+-- Convert old date casing combos to supported format
+UPDATE Config SET Value='Ymd' 
+	WHERE Value IN ('YMD', 'YMd', 'YmD', 'yMD', 'yMd', 'ymD', 'ymd') 
+	AND ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='dobFormat');
+
+UPDATE Config SET Value='Ymd' 
+        WHERE Value IN ('YMD', 'YMd', 'YmD', 'yMD', 'yMd', 'ymD', 'ymd') 
+        AND ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='dodFormat');
+
+UPDATE Config SET Value='Ym'
+	WHERE Value IN ('YM', 'ym', 'yM')
+	AND ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='dobFormat');
+
+UPDATE Config SET Value='Ym'
+        WHERE Value IN ('YM', 'ym', 'yM')
+        AND ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='dodFormat');
--- a/SQL/New_patches/2020_06_16_Add_Date_Format_to_ConfigSettings_DataType.sql
+++ b/SQL/New_patches/2020_06_16_Add_Date_Format_to_ConfigSettings_DataType.sql
@@ -12,17 +12,17 @@ UPDATE ConfigSettings
 
 -- Convert old date casing combos to supported format
 UPDATE Config SET Value='Ymd' 
-	WHERE Value IN ('YMD', 'YMd', 'YmD', 'yMD', 'yMd', 'ymD', 'ymd') 
+	WHERE LOWER(Value)='ymd'
 	AND ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='dobFormat');
 
 UPDATE Config SET Value='Ymd' 
-        WHERE Value IN ('YMD', 'YMd', 'YmD', 'yMD', 'yMd', 'ymD', 'ymd') 
+        WHERE LOWER(Value)='ymd'
         AND ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='dodFormat');
 
 UPDATE Config SET Value='Ym'
-	WHERE Value IN ('YM', 'ym', 'yM')
+	WHERE LOWER(Value)='ym'
 	AND ConfigID=(SELECT ID FROM ConfigSettings WHERE Name='dobFormat');
 
 UPDATE Config SET Value='Ym'
-        WHERE Value IN ('YM', 'ym', 'yM')
+        WHERE LOWER(Value)='ym'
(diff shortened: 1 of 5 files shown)


driusan:
@jesscall this is failing Travis

ridz1208:
@jesscall The way I understand it, the casing of the `YMD` letters actually dictates the formate of dates so I don't think it's a good idea to make all combinations of `Ym YM yM ym` do the same thing, just in case we decide to support more formats later on.

I think the proper fix to the reported issue would be to throw a configuration error in the config module if you choose a date different then the only 2 supported formats OR to change that configuration field to a drop down and progressively add formats to it as we move forward.

maltheism:
@ridz1208 @jesscall something like https://www.php.net/manual/en/datetime.format.php might be useful for converting to the correct format.

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -586,7 +586,7 @@ function editCandidateDOD(\Database $db, \User $user): void
     if (!empty($dod)) {
         $config    = \NDB_Config::singleton();
         $dodFormat = $config->getSetting('dodFormat');
-        if ($dodFormat === 'YM') {
+        if ($dobFormat === 'Ym') {
```
jesscall, line 589:
```suggestion
        if ($dodFormat === 'Ym') {
```

jesscall:
@CamilleBeau ready for review

**[Notes]**:
- The PR has gone in a different direction. It now adds a dropdown to configuration module where the user can select a format from a pre-determined list. 
- I've hard-coded this list as there are only 2 formats we support at the moment, however I wouldn't mind creating a table for this since it is possible the list grows over time. 
- The caveat for projects is that the date format entered in the text field may not match the dropdown options which will lead to some issues. On upgrade projects should double check and amend where applicable.

**[Testing]**: 
- Please test the configuration module DOB & DOD format fields. 
- Also test if the original bug is resolved.

SQL/New_patches/2020_06_16_Add_Date_Format_to_ConfigSettings_DataType.sql
```diff
@@ -0,0 +1,11 @@
+-- Add date_format option to DataType
```
ridz1208, line 1: I wouldnt mind seeing in this patch. conversion from any casing  combination into the supported format

ie

`UPDATE Config SET Value='Ymd' WHERE Value IN ('YMD','YMd','ymd',...) AND ConfigID=(SELECT ID ...);`

and a similar update for the different combinations of YM

CamilleBeau:
Error given when sourcing rb while in this branch:
image

johnsaigle:
There's also the config setting `dateDisplayFormat`. I think it should also be changed in the SQL for consistency along with any other date config settings.

Alternatively - I'm not sure if it's important for the DoD and DoB formats to be different. Could they both instead use `dateDisplayFormat`?

ridz1208:
@johnsaigle the way we save the data does impact the display. Users get confused when they can select a day but it defaults to 01 after saving. The display should clearly show that you are selecting a whole month at a time

maltheism:
@ridz1208 maybe the user's "format" of the date should be saved in the database and used for reverting the date in the database to the correct format for the user on the frontened.

johnsaigle:
@ridz1208 I think we're saying the same thing. We should save whatever the user enters but display it with either the day chopped off or with `-01`. Alternatively if the day is never needed then I think the issue is with the input fields. We shouldn't prompt for a day if it won't be used.

jesscall:
@johnsaigle the issue with saving the full date in the DB and then only displaying the day as `01` is that it makes anonymization / identification of a candidate a concern when it comes to data releases

maltheism:
@jesscall @johnsaigle I think having the actual day is important.
Why not just force the configuration module for DOB & DOD Format to always have "d" in the input field and an input error is displayed if otherwise. You could use regex.

ridz1208:
@maltheism those are ethics regulated requirements, some projects are not allowed to save the day of the date of birth because it's identifiable

maltheism:
@ridz1208 hmm if that's the case maybe there should be two more date columns in database.
Ex. actual_dod, actual_dob, ethics_od, ethics_dob

Someone that's born in the 1st of whatever month and dies on the 1st of whatever month. Wouldn't be happy if they were in the ethical study.

ridz1208:
@maltheism the idea is that we can't store the actual values in the database just like we can't store names and addresses

maltheism:
@ridz1208 the ethics_dod or ethics_dob could be scrambled and the actual_dob & actual_dod would be left null depending on if it's an ethical study.

SQL/New_patches/2020_06_16_Add_Date_Format_to_ConfigSettings_DataType.sql
jesscall: @ridz1208 implemented.

jesscall:
@CamilleBeau I suspect the data truncation is happening because of the note i left about existing projects having different values than the preset formats. Should be solved by running the last 4 queries in the patch submitted.

SQL/New_patches/2020_06_16_Add_Date_Format_to_ConfigSettings_DataType.sql
```diff
@@ -8,4 +8,21 @@ UPDATE ConfigSettings
 
 -- Update DataType of dodFormat
 UPDATE ConfigSettings
-	SET DataType='date_format' WHERE Name='dodFormat';
\ No newline at end of file
+	SET DataType='date_format' WHERE Name='dodFormat';
+
+-- Convert old date casing combos to supported format
+UPDATE Config SET Value='Ymd' 
+	WHERE Value IN ('YMD', 'YMd', 'YmD', 'yMD', 'yMd', 'ymD', 'ymd')
```
CamilleBeau, line 15: This might be a little bit cleaner
```suggestion
	WHERE LOWER(Value)='ymd' 
```

jesscall: @CamilleBeau beast. implemented.

ridz1208: @CamilleBeau DAYUM !

ridz1208:
@jesscall travis

CamilleBeau:
Getting `ERROR 1265 (01000) at line 2402: Data truncated for column 'DataType' at row 1` when sourcing raisinbread

modules/configuration/php/configuration.class.inc
```diff
@@ -61,6 +61,12 @@ class Configuration extends \NDB_Form
             $scan_types[$val] = $val;
         }
 
+        $date_format = [
+            ''    => '',
+            'Ymd' => 'Ymd',
```
driusan, line 66: Maybe it would be a good idea to show a more descriptive label for these on the frontend since it's a dropdown now?

Like `2015-12-29 (Ymd)` and `2015-12 (Ym)`?

(Date is arbitrary, just something that's unambiguous with the day of the month > 12.. I'm assuming m is numeric month, I haven't looked into it, but it would avoid that type of confusion for users to provide an example in the dropdown label..)

PR #6728, opened 2020-06-15
Fix order of SQL in release upgrade patch

### LINKED ISSUE #6726: 22.0_To_23.0_upgrade.sql attempts to insert into modules table before creating the table
**Describe the bug**
Upgrade script attempts to insert row into `modules` table before creating table. Attempted insert is at line 72, table creation is at line 308. 

**To Reproduce**
Upgrade mySQL database from v22 to v23 using SQL/Release_patches/22.0_To_23.0_upgrade.sql

driusan:
The modules table must be created before the battery_manager
attempts to insert into the module.

Resolves #6726


driusan:
This passed Travis when you click through, but the icon didn't update on the github frontend..

PR #6729, opened 2020-06-15
[Imaging Browser] Project Filter Permission

### LINKED ISSUE #6691: [Imaging Browser] project filter ComboBox shows all projects (including the ones the user doesn't have permissions to see)
**Describe the bug**
This seems to be a minor issue.

The `Project filter ComboBox` shows all projects (including the ones the user doesn't have permissions to see). In the other hand the results for querying projects where the user doesn't have permissions are not been shown in the result table which is a good thing. 

**To Reproduce**
1. Go to Imaging -> Imaging Browser
2. Click on the `Project ComboBox`
3. The user is able to see and choose projects it doesn't have permission to access.

In step number 3 the user should be able to see and choose only from the project it have access.

laemtl:
This PR modifies the way the project filter is populated. Now, a project is displayed in the filter if the current user have access to it.

* Resolves #6691
* Related to #6829

### DIFF
--- a/modules/imaging_browser/php/imaging_browser.class.inc
+++ b/modules/imaging_browser/php/imaging_browser.class.inc
@@ -112,12 +112,11 @@ class Imaging_Browser extends \DataFrameworkMenu
         }
 
         // get list of projects
-        $allProjectList   = \Utility::getProjectList();
-        $list_of_projects = array();
-        $userProjectsIDs  = $user->getData('ProjectIDs');
-        foreach ($userProjectsIDs as $ProjectID) {
-            $project = $allProjectList[$ProjectID];
-            $list_of_projects[$project] = $project;
+        $list_of_projects = [];
+        $userProjectsIDs  = $user->getProjectIDs();
+        foreach ($userProjectsIDs as $projectID) {
+            $projectName = \Project::getProjectFromID($projectID)->getName();
+            $list_of_projects[$projectName] = $projectName;
         }
 
         // get visit QC status options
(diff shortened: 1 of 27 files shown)


driusan:
This should be going to master

laemtl:
Really good point. Let me investigate.

christinerogers:
Also @laemtl - please check anything associated with the new DataFramework - which should be controlling Project access hereon out. 
e.g. See all Modules whose _menu filter_ pages (i.e. front page only, not subpages) were updated [according to this GitHub-project](https://github.com/aces/Loris/projects/7) and check how they dealt with it (or should have).

laemtl:
Related to #6829

modules/imaging_browser/php/imaging_browser.class.inc
```diff
@@ -112,10 +112,12 @@ class Imaging_Browser extends \DataFrameworkMenu
         }
 
         // get list of projects
+        // PROJECTS
+        $allProjectList   = \Utility::getProjectList();
         $list_of_projects = array();
-        $projectList      = \Utility::getProjectList();
-        foreach ($projectList as $key => $project) {
-            unset($projectList[$key]);
+        $userProjectsIDs  = $user->getData('ProjectIDs');
```
christinerogers, line 118: try $user->getProjectIDs() per dave's comment in #6831
Should be worth a new function in User to get the project list if this code needs to be recycled more than once across multiple modules

christinerogers: per my comment above -- Based on your [investigation across modules](https://github.com/aces/Loris/issues/6829#issuecomment-656357582) - 
Why not add a User class function to perform these 4-5 lines of code?

laemtl: I think the foreach is formatting an array for the select filter, this can change between modules, and the snippet is doing formatting fir the select filters. I don't think it makes sense in the User class.

laemtl:
Replaced by #6836

PR #6732, opened 2020-06-15
[Core] Require correct types for lorisform->addSelect

driusan:
Since the type is currently only enforced by PHPDocs and not
PHP, incorrect types can be passed by addSelect. This enforces
the types from the PHP side of things.

(Currently when a string is passed instead of an array in some
instruments, the array_merge call generates a PHP warning which
breaks endpoints which use the instrument in a JSON context. This
change makes the invalid instruments die with a type error instead,
forcing them to be fixed.)

### DIFF
--- a/php/libraries/NDB_Page.class.inc
+++ b/php/libraries/NDB_Page.class.inc
@@ -187,8 +187,12 @@ class NDB_Page implements RequestHandlerInterface
      *
      * @return void
      */
-    function addSelect(string $name, string $label, array $options, array $attribs=array())
-    {
+    function addSelect(
+        string $name,
+        string $label,
+        array $options,
+        array $attribs=array()
+    ) {
         $attribs = array_merge(
             array('class' => 'form-control input-sm'),
             $attribs
--- a/php/libraries/NDB_Page.class.inc
+++ b/php/libraries/NDB_Page.class.inc
@@ -187,7 +187,7 @@ class NDB_Page implements RequestHandlerInterface
      *
      * @return void
      */
-    function addSelect($name, $label, $options, $attribs=array())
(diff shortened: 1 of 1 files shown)


php/libraries/NDB_Page.class.inc
```diff
@@ -187,7 +187,7 @@ class NDB_Page implements RequestHandlerInterface
      *
      * @return void
      */
-    function addSelect($name, $label, $options, $attribs=array())
+    function addSelect(string $name, string $label, array $options, array $attribs=array())
```
johnsaigle, line 190: Looks good. Could you types to the arrays in the comment if known? e.g. changing line 183 to

      * @param array<string,string>  $options Options to pass to QuickForm for this

driusan: I don't know how it's being called, I just based the php types on the phpdoc and know that the places where `$attribs` was a string are wrong..

johnsaigle:
One of the code formatting issues comes from #6733

php/libraries/NDB_Page.class.inc
HenriRabalais: @johnsaigle does php allow type enforcement within arrays now?

johnsaigle: Not exactly. Some tools know how to analyze the PHPDocs and do additional validation if you specify that all entries need to be a specific type but something else gets passed in. It's not built into PHP itself.

driusan: @HenriRabalais No, he's referring to the doc comments. phan does analysis based on them in addition to the PHP types

PR #6733, opened 2020-06-15
Fix PHPCS error on main branch

johnsaigle:
There was a PHPCS issue in https://github.com/aces/Loris/pull/6180 but it was merged


driusan:
Travis doesn't seem to be starting at all on this one..

johnsaigle:
I noticed that one some other PRs also. I think it's a bit buggy today

PR #6735, opened 2020-06-15
[Core] Fix warnings about continue in a switch being equivalent to a break

driusan:
Fix warnings when running `make checkstatic`:

```
PHP Warning:  "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in tools/generate_tables_sql.php on line 74
PHP Warning:  "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in tools/generate_tables_sql.php on line 80
PHP Warning:  "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in tools/exporters/data_dictionary_builder.php on line 169
PHP Warning:  "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in tools/exporters/data_dictionary_builder.php on line 178
PHP Warning:  "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in tools/generate_tables_sql_and_testNames.php on line 55
PHP Warning:  "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in tools/generate_tables_sql_and_testNames.php on line 61
PHP Warning:  "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in tools/generate_tables_sql_and_testNames.php on line 67
```

All the instances were in a switch that was the last thing in a loop, so breaking out of the switch or continuing the loop were equivalent, but I tried to update the code to be whichever seemed to be what the statement was semantically trying to do.

### DIFF
--- a/tools/exporters/data_dictionary_builder.php
+++ b/tools/exporters/data_dictionary_builder.php
@@ -166,7 +166,7 @@ foreach ($instruments AS $instrument) {
 
                 // Skip lines that contains only label or notes where bit[1] is empty.
                 if (empty($bits[1])) {
-                    continue;
+                    continue 2;
                 }
 
                 print "\tInserting $table $bits[1]\n";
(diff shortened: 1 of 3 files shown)


johnsaigle:
Issue arises from #6733

PR #6416, opened 2020-04-20
[permissions] Renamed description of superuser

ridz1208:
## Brief summary of changes
Changed description of super user to a clearer one

#### Link(s) to related issue(s)

* Resolves #6355

PR #6738, opened 2020-06-16
Add name of pre-GitHub contributor

jpmckinney:
For whatever reason I was reminiscing about the NIH MRI study of normal brain development this morning, and wondering what happened to that software project I contributed to in 2004. It was wonderful to see the great progress that has been made! I also see you recently corrected one of my stranger contributions of "There can only be one" #6416.

Anyway, I don't know if there is a logic to the order or inclusion of past contributor names, but I put myself before-last, as I find going last carries an extra "oomph" that Samir deserves :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,4 +121,4 @@ LORIS is made by staff developers at the [McGill Centre for Integrative Neurosci
 
 Visit [the LORIS website](https://loris.ca) for the history of LORIS and our **Technical Papers**.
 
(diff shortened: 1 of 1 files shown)


christinerogers:
Hi James - Thanks for adding yourself!  
Yeah, we retired the Highlander, it had a good run and was a fun easter egg for more than a decade. :)

PR #6553, opened 2020-05-19
[Test] unittests for Database.class.inc

AlexandraLivadas:
## Brief summary of changes

Unit tests for the Database library under `php/libraries`. This PR will be updated as more unit test coverage is added. Comments were added to pre-existing and new unit tests.

setUp and tearDown functions were added as well.

As of 06/16: All **public** methods are covered except for `connect` and `singleton`. 

#### Testing instructions

1. Run `npm run tests:unit -- --filter Database` or check out the Travis output.

PR #6752, opened 2020-06-19
[Test] Fix failing unit test causing Travis fail

AlexandraLivadas:
## Brief summary of changes

This PR fixes the failing unit test that is causing Travis to fail on the master branch. This was caused by a change made in merged PR #6553.

### DIFF
--- a/test/unittests/Database_Test.php
+++ b/test/unittests/Database_Test.php
@@ -870,7 +870,9 @@ class Database_Test extends TestCase
         $stmt = $this->getMockBuilder('PDOStatement')->getMock();
 
         $stub->_PDO->expects($this->once())
-            ->method("prepare")->with($this->equalTo("SHOW TABLES"));
+            ->method("prepare")
+            ->with($this->equalTo("SHOW TABLES"))
+            ->willReturn(new PDOStatement());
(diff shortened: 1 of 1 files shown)


driusan:
Why did the linked PR pass Travis if it's the cause of the failure?

AlexandraLivadas:
> Why did the linked PR pass Travis if it's the cause of the failure?

@driusan I'm not sure. I'm trying to do some digging now, but I don't understand how that happened. I'm trying to see if any commits on master done after I made the branch could have affected it (changes in the code that weren't on my branch). I will keep you updated.

AlexandraLivadas:
@driusan I found the issue. 2 days before PR #6553 was merged, @johnsaigle made this [commit](https://github.com/aces/Loris/commit/4860cf1e6b2266a52add23f61a0e087be1bfe9d7#diff-c1d4495bb4192e908841940828b887a4). In this [line](https://github.com/aces/Loris/commit/4860cf1e6b2266a52add23f61a0e087be1bfe9d7#diff-c1d4495bb4192e908841940828b887a4R681), he added code that specified that the method **must** return a PDOStatement, which was not included before. This is what caused the unit test to fail.

PR #6761, opened 2020-06-22
[eslint] Modules - fix max-len

### LINKED ISSUE #6750: [eslint] Fix warnings

laemtl:
This PR:
* resolves all eslint max-len warnings
* raises max-len severity to error

### Step 1
```
rm -R nodes_modules/
npm install
```

### Step 2
```
make checkstatic
```
No warnings `max-len` should be printed.

### Step 3
```
git checkout 'one-modiifed file'
make checkstatic
```
An error `max-len` should be printed.

* resolves #6750 (partially)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -18,8 +18,7 @@
   "rules": {
     "extends": ["eslint:recommended", "plugin:react/recommended", "google"],
     "max-len": [2, 80, 2, {
-        "ignoreComments": true,
-        "ignoreStrings": true
+        "ignoreComments": true
     }],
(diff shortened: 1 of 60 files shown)


laemtl:
@maltheism Would you like to help me to review?

maltheism:
@laemtl looks good but you have to fix the conflict. 
My only thought is I'm unsure if the "80" characters length is best for javascript like it's the preference for php. I'll approve after conflict is fixed.

laemtl:
@maltheism Done!

PR #6764, opened 2020-06-23
[eslint] Fix prefer-rest-params warning

### LINKED ISSUE #6750: [eslint] Fix warnings

laemtl:
This PR:
* resolves all eslint prefer-rest-params warnings
* raises prefer-rest-params severity to error

### Step 1
```
rm -R nodes_modules/
npm install
```

### Step 2
```
make checkstatic
```
No warnings `Use the rest parameters instead of 'arguments'  prefer-rest-params` should be printed.

### Step 3
```
git checkout master htdocs/js/loris.js
make checkstatic
```
An error `Use the rest parameters instead of 'arguments'  prefer-rest-params` should be printed.

* resolves #6750 (partially)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -23,7 +23,7 @@
     }],
     "react/jsx-uses-react": "error",
     "react/jsx-uses-vars": "error",
-    "prefer-rest-params": "warn",
+    "prefer-rest-params": "error",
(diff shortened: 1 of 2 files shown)


laemtl:
npm run compile will not run es-lint on the modified file. To see the error make checkstatic will help.

PR #6770, opened 2020-06-26
[User Accounts] Superuser: add user permission fix

### LINKED ISSUE #6687: [User Accounts] User with superuser privilege cannot modify permissions for other users or own account
**Describe the bug**
Even when a user that is not `admin` has `superuser` permission, they are limited in their ability to modify user accounts.

**To Reproduce**
1. Using `admin`, create a new user `User1` that has only the permission `superuser`.
2. Login as `User1` and go to `user_accounts`
3. You will see that `User1` is not able to edit any of their permissions.
4. As `User1`, create a new user account.
5. `User1` will not be able to give any permission except for `superuser`.

Since superuser short-circuits every other permission, having this permission should allow a user to fully modify their own account, as well as others.

Right now it seems that if a user other than `admin` has `superuser` they still do not have full rights to modify user accounts.

- LORIS Version: 23.0-release branch

**Additional context**

Screen Shot 2020-06-08 at 15 50 14

---

Screen Shot 2020-06-08 at 15 52 07

laemtl:
When a user has superuser permission, they are limited in their ability to modify user accounts.
This PR modifies the permission class to assign all the permission to a user when superuser is selected.

* Resolves #6687


ridz1208:
@laemtl So this is my PERSONAL opinion about these changes because these changes would work to solve the issue but...

1. We currently exempt users with superuser permission from permission checks directly in the hasPermission call if we give them in parallell the setPermission changes we are doing here... it just seems like redundant work. This kind of redundancy that could make things confusing when it comes to debugging down the line.
2. it makes more sens to me to bypass permission checking for superusers rather than physically give them all permissions. i just find it more elegant and it gives me the hope that one day we could kill the superuser all together ?
3. I feel like the isue doesnt currently come from the core libraries. it seems like a user_accounts issue that is being solved too high up the infrastructure. I would do the fix directly in user accounts instead... simply check if the current user is a superuser and display all permissions... 
4. I dont like the changes to the query in this PR. I understand why it is necessary in your usecase but it seems like worst SQL to me to call it an improvement.

All in all, I have not looked at the user_accounts code in a while but I feel like there might be a better solution in there that would have less of a global effect over the rest of the codebase..

laemtl:
Thanks for this detailed review, your points makes a lot of sense. I don't really remember why I approached the problem this way but I agree that there is probably a better solution that limit the changes effects on the user accounts module.

laemtl:
@ridz1208 I followed your suggestion. Let me know what you think.

driusan:
@ridz1208 ?

ridz1208 (approved):

ridz1208:
These changes look fine by me. 

This is not a feature I would have pushed for myself but i see how it can be useful so I'm passing it

ridz1208:
@maltheism could you test this please ?

PR #6790, opened 2020-06-30
[Tools] Enable PHPCS for CouchDB_Import_MRI.php

laemtl:
Replaces #5999


spell00:
@laemtl I'm not sure how to test these changes, I ran `sudo bash test/run-php-linter.sh` and `phpcs tools/CouchDB_Import_MRI.php` without any error, is there anything else I should run?

PR #6792, opened 2020-06-30
[Tools] Enable PHPCS for CouchDB_MRI_Importer.php

laemtl:
Replaces #5997


nicolasbrossard (approved):
Code looks good. Pulled the branch on my MV and `Couch_MRI_Importer.php` passes `phpcs`. 👍

PR #6796, opened 2020-06-30
[Tools] Enable PHPCS for config_to_db.php

laemtl:
Replaces #5988

### DIFF
--- a/test/run-php-linter.sh
+++ b/test/run-php-linter.sh
@@ -19,6 +19,7 @@ find docs modules htdocs php src tools \
 declare -a tools_list=(
     'assign_missing_instruments.php'
     'configuration_check.php'
+    'config_to_db.php'
     'data_integrity_check.php'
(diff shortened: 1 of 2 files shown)


christinerogers:
This script can probably be deprecated/deleted. 
It was a one-time script for a 2015 release upgrade -- [see history](https://github.com/johnsaigle/Loris/commits/master/tools/config_to_db.php).
 
@johnsaigle, given your commit in Jan - do you recall any recent discussion? 
re value in keeping it for future re-use / possibility of maintaining it?

johnsaigle:
No I don't know the context behind it. We can put it in the `tools/deprecated` folder instead.

driusan:
This belongs in `tools/single_use`, not `tools/deprecated`. It was a script used for an upgrade.

laemtl:
Do we still want to run PHPCS for this script for each build then?

PR #6149, opened 2020-03-09
Add PHPCS to core integration tests

johnsaigle:
## Brief summary of changes

Adds PHPCS to files in `test/integrationtests/`

PR #6800, opened 2020-06-30
Add PHPCS to core integration tests 

laemtl:
Replaces #6149

### DIFF
--- a/test/run-php-linter.sh
+++ b/test/run-php-linter.sh
@@ -52,8 +52,8 @@ vendor/bin/phpcs --standard=test/LorisCS.xml --extensions=php,inc \
     php/ \
     htdocs/ \
     modules/ \
+    "test/integrationtests/" \
     "${tools_list[@]/#/tools/}" \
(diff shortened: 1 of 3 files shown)


test/run-php-linter.sh
```diff
@@ -52,8 +52,8 @@ vendor/bin/phpcs --standard=test/LorisCS.xml --extensions=php,inc \
     php/ \
     htdocs/ \
     modules/ \
+    "test/integrationtests/" \
```
driusan, line 55: The variable "test_list" above doesn't appear to be used anymore and should be deleted.

laemtl: @johnsaigle Any reason why PHPCS is not run on the test_list anymore?

driusan: You changed it to run on the entire directory that had the files in the list.

laemtl: Oh, missed that. You're right then.

PR #6801, opened 2020-06-30
[MRI Violations] Tabs Navigation Fix

### LINKED ISSUE #6689: [MRI Violations] Navigation issue on "/mri_protocol_violations/" page
**Describe the bug**
The page `whatever.loris.ca/mri_violations/mri_protocol_violations/` contains two tabs under the menu filter. 

The second tab appears as active on this page. 

Clicking the first tab takes the user to `whatever.loris.ca/mri_violations/` which removes both tabs.

**To Reproduce**
1. Go to `mri_violations/mri_protocol_violations/`
2. Click on the first tab (marked "Protocol Violations")
3. After the page loads, the tabs are gone

The tabs shouldn't disappear. A user should be able to alternate between them.

laemtl:
/mri_violations/mri_protocol_violations/ contains two tabs under the menu filter.
Clicking the first tab takes the user to /mri_violations/ which removes both tabs.

This PR adds the same tabs in /mri_violations to allow the users to alternate between them.

Screenshot from 2020-06-30 16-28-01

* Resolves #6689

### DIFF
--- a/modules/mri_violations/php/mri_protocol_check_violations.class.inc
+++ b/modules/mri_violations/php/mri_protocol_check_violations.class.inc
@@ -70,10 +70,10 @@ class Mri_Protocol_Check_Violations extends \NDB_Menu_Filter
             implode(',', $siteClauseElements)
         );
 
-        $this->query    = " FROM mri_violations_log l 
+        $this->query    = " FROM mri_violations_log l
             LEFT JOIN mri_scan_type m
-            ON m.ID=l.Scan_type 
-            LEFT JOIN mri_protocol_checks_group mpcg 
+            ON m.ID=l.Scan_type
+            LEFT JOIN mri_protocol_checks_group mpcg
             ON (mpcg.MriProtocolChecksGroupID = l.MriProtocolChecksGroupID)
             LEFT JOIN candidate c ON (l.CandID=c.CandID)
             LEFT JOIN session s ON (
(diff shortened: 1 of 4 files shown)


nicolasbrossard:
I think page 'MRI protocol Check Violations' should contain two tabs, 'Not resolved' and 'Resolved', just like the main module's page (MRI Violated Scans). These tabs should **only** show the violated scans for which the "problem" (see column with the same name on the main page) is 'Protocol Violation' and filter the display based on the resolution status of the scan. With your PR, clicking on tab 'Resolved Violations' actually takes you to the 'MRI Protocol Violations' page and basically displays all the violated scans for which the problem was set to 'Could not identify Scan Type'. I believe this was not the intended behaviour. @cmadjar What do you think?

cmadjar:
@laemtl Nicolas and I discussed about these code changes. Actually, the MRI Protocol Violation and MRI Protocol Check Violation pages are subpages of the MRI violation module.

The main page of the MRI violation module has unresolved and resolved tabs but the other subpage should not have those. Maybe we should find a way to make it clearer that you are on a different page than the landing page of the module to avoid any confusion but in any case, there should not be tabs unresolved for the subpages.

@nicolasbrossard  will modify the description of #6689 shortly.

Thank you both!

nicolasbrossard:
@cmadjar I just realized that since we plan on removing the two obsolete tabs, we should probably add a 'Resolution Status' column in the results table for both sub-pages so that you always know whether a given violation was resolved or not. Agreed?

cmadjar:
@nicolasbrossard Agreed!

PR #6805, opened 2020-07-06
[Tools] delete_candidate CanID Fix

### LINKED ISSUE #6757: delete_candidate.php should use CandID class
**Describe the Problem**
delete_candidate.php tool has not been updated to use CandID class and throws an error in newer versions of LORIS. 

**Suggested Action**
Add `use LORIS\StudyEntities\Candidate\CandID;` at the beginning of the file. Replace `$candID = $argv[2];` with `$candID = new CandID($argv[2]);`

laemtl:
Fix in delete_candidate.php tool to use the CandID class.

* Resolves #6757

### DIFF
--- a/tools/delete_candidate.php
+++ b/tools/delete_candidate.php
@@ -18,11 +18,15 @@
  * @license  Loris license
  * @link     https://www.github.com/aces/Loris-Trunk/
  */
+
+use LORIS\StudyEntities\Candidate\CandID;
+
 set_include_path(
     get_include_path().":".
     __DIR__."/../project/tools:".
     __DIR__."/../php/tools:"
 );
+
 require_once __DIR__ . "/../vendor/autoload.php";
 require_once "generic_includes.php";
 
@@ -39,7 +43,7 @@ require_once "generic_includes.php";
  * @package  Loris
  * @author   Various <example@example.com>
  * @license  Loris license
- * @link     https://www.github.com/aces/Loris-Trunk/
+ * @link     https://www.github.com/aces/Loris/
  */
 const MIN_NUMBER_OF_ARGS = 4;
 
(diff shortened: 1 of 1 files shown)


tools/delete_candidate.php
```diff
@@ -150,7 +154,7 @@ function showHelp()
  * (second-level relations) should have actions on delete specified in the
  * database schema
  *
- * @param string   $CandID     Identifying the candidate
+ * @param CanID    $CandID     Identifying the candidate
```
driusan, line 157:
```suggestion
 * @param CandID   $CandID     Identifying the candidate
```

tools/delete_candidate.php
```diff
@@ -18,11 +18,15 @@
  * @license  Loris license
  * @link     https://www.github.com/aces/Loris-Trunk/
  */
+
+use LORIS\StudyEntities\Candidate\CandID;
```
driusan, line 22: I'm pretty sure this needs to go below the autoload require.

tools/delete_candidate.php
```diff
@@ -18,11 +18,15 @@
  * @license  Loris license
  * @link     https://www.github.com/aces/Loris-Trunk/
```
christinerogers, line 19:
```suggestion
 * @link     https://www.github.com/aces/Loris/
```
cleanup

tools/delete_candidate.php
laemtl: @driusan For my personal knowledge, can you explain a bit why? I will take that into account fot future changes.

driusan: I might be mistaken, it might not try autoloading until the class is referenced with "new", but how would the class be loaded if the autoloader hasn't been setup yet?

tools/delete_candidate.php
driusan: @laemtl

laemtl:
Thanks @ridz1208

PR #6807, opened 2020-07-06
[candidate_profile] Show Consent Summary card iff useConsent Yes

### LINKED ISSUE #6680: [candidate_parameters] Consent card visible in candidate_profile
**Describe the bug**
The consent card appears in the Candidate Profile page even if `Use Consent` setting is set to no in configuration module, contrary to what is specified by the TestPlan. 

**To Reproduce**
1. From admin, set `Use Consent` to yes in configuration module
2. Navigate to Candidate Profile page from non-admin account and see the consent card there as it should be.
3. Now set `Use Consent` to no in the configuration module.
4. Refresh the candidate profile page (or navigate away and back again) and see that the consent card is still there. 

The consent card to only be visible if it is set to 'yes' in config module.

zaliqarosli:
## Brief summary of changes

Resolves https://github.com/aces/Loris/issues/6680

1. In Configurations module,  set 'Use consent' config option to Yes.
2. Navigate to Candidate Profile page from non-admin account and see the consent card there as it should be.
3. Now, set 'Use Consent' config to 'No'
4. Refresh the candidate profile page and see that the consent card is no longer there

* Resolves #6680

### DIFF
--- a/modules/candidate_parameters/php/module.class.inc
+++ b/modules/candidate_parameters/php/module.class.inc
@@ -42,6 +42,10 @@ class Module extends \Module
                 return [];
             }
 
+            if (!\NDB_Factory::singleton()->settings()->consentEnabled()) {
+                return [];
+            }
+
             // The candidate getConsents only returns the types of
             // consents that the candidate has saved, we want to
             // summarize all of the types configured in the database.
--- a/php/libraries/Settings.class.inc
+++ b/php/libraries/Settings.class.inc
@@ -164,5 +164,17 @@ class Settings
         return false;
     }
 
+    /**
+     * Determines whether the useConsent configuration setting is
+     * enabled
+     *
+     * @return bool True if the useConsent setting is enabled
+     */
+    function consentEnabled(): bool
+    {
+        return \NDB_Factory::singleton()->config()->settingEnabled(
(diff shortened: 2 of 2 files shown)


modules/candidate_parameters/php/module.class.inc
```diff
@@ -42,6 +42,10 @@ class Module extends \Module
                 return [];
             }
 
+            if (\NDB_Config::singleton()->getSetting('useConsent') != 'true') {
```
driusan, line 45: The settingEnabled function seems like it would be a little more robust.

zaliqarosli: okay, in which class would you want the consentEnabled() defined?

driusan: php/libraries/Settings.class.inc seems like the most logical place

zaliqarosli:
@driusan ready for re-review

driusan:
@zaliqarosli this is going to the wrong branch, it should be 23.0-release

zaliqarosli:
@driusan done

sruthymathew123:
Tested as per the testing instructions and passes the manual tests.

PR #6809, opened 2020-07-06
fix bug in #6179

lingma:
## Brief summary of changes

Fix bug in #6179


driusan:
thanks for fixing this @lingma !

PR #6810, opened 2020-07-06
[instrument builder] Min/max null default value

### LINKED ISSUE #6568: [instrument_builder] Numeric Min/Max fields auto-populated with "0"
**Describe the bug**
While a numeric element can be added without 'Min' and 'Max' values, the default value "0" for both 'Min' and 'Max' fields return 'Max value must be larger than min value' error on clicking 'Add Row'.

**To Reproduce**

1. Go to 'Instrument Builder' module and click on the 'Build' tab
2. Click on the 'Select One' dropdown and select 'Numeric' under 'Data entry'
3. Fill out 'Question Name' and 'Question Text' fields, leaving 'Min' and 'Max' fields (already populated with "0") untouched. Click 'Add Row'. This will trigger the error 'Max value must be larger than min value'.
4. Remove the zeros from the 'Min' and 'Max' fields, leaving them empty. Click 'Add Row'.
The element should save in the 'Question Display (Front End)' table above.

The 'Min' and 'Max' fields should be empty as the default

laemtl:
The default value "0" for both 'Min' and 'Max' fields return 'Max value must be larger than min value' error on clicking 'Add Row'.
The 'Min' and 'Max' fields are now empty as the default.

* Resolves #6568

### DIFF
--- a/modules/instrument_builder/jsx/react.questions.js
+++ b/modules/instrument_builder/jsx/react.questions.js
@@ -526,8 +526,8 @@ class ListElements extends Component {
         break;
       case 'numeric':
         newState.Options = {
-          MinValue: 0,
-          MaxValue: 0,
+          MinValue: null,
+          MaxValue: null,
         };
         break;
       default:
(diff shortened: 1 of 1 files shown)


laemtl:
Thanks @zaliqarosli, I will fix that!

PR #6812, opened 2020-07-06
[Core] DataInstance to implement JSONSerializable

### LINKED ISSUE #6739: Data Framework should use jsonSerializeable interface, not toJSON
The interfaces in the `\LORIS\Data` namespace that need to be serialized as JSON should be using the standard php jsonSerializeable interface, not a custom interface with a toJSON function. This would allow us to directly pass objects to json_encode and generally be more interoperable with PHP and potentially efficient.

laemtl:
The interfaces in the \LORIS\Data namespace that need to be serialized as JSON now use the standard php jsonSerializeable interface, replacing toJSON().

* Resolves #6739


laemtl:
@driusan Change done!

modules/electrophysiology_browser/php/models/electrophysiofile.class.inc
```diff
@@ -86,10 +86,10 @@ class ElectrophysioFile implements \LORIS\Data\DataInstance
...
      * Implements \LORIS\Data\DataInstance interface
      *
-     * @return string
+     * @return array which can be serialized by json_encode()
      */
-    public function toJSON(): string
+    public function jsonSerialize() : array
     {
-        return '';
+        return [];
```
driusan, line 93: This one doesn't look right. Why does it turn from the empty string which returns nothing to the caller to the empty array (which is an actual value)? I think the type signature is just wrong for this file and it shouldn't be returning an array.

PR #6815, opened 2020-07-07
[timepoint_list] Hiding Candidate Info and View Imaging Datasets buttons

### LINKED ISSUE #6673: [timepoint_list] Candidate Info and View Imaging Datasets buttons always appear
While accessing a same-site candidate as a user with no permissions, the  "Candidate Info" and  "View Imaging Datasets" links always appear. They do not require the permissions required in the test plan (though clicking on them gives a permission denied error, so the modules linked to are properly enforcing permissions.)

(ie "Action Buttons" step 2 and 3 of the test plan fail.)

CamilleBeau:
## Brief summary of changes
This PR hides the "Candidate Info" and "View Imaging Datasets" buttons along with the "Create time point" button when the necessary permissions are not present. This was done with the assumption that "View Imaging Datasets" should only be visible if the user has one of the four imaging browser permissions (please correct me if this assumption is wrong). Test plan documentation was edited accordingly. If the user has _none_ of the relevant permissions, the "Actions:" label is also hidden.

1. Access a same-site candidate as the user without any permissions using a link. There should be no "Actions:" label and none of the three buttons "Candidate Info", "View Imaging Datasets", or "Create time point".
2. Access a same-site candidate as the user with only the `data_entry` permission. There should be an "Actions:" label as well as the two buttons "Candidate Info" and "Create time point".
3. Access a same-site candidate as the user with any/all of the imaging browser permissions. There should be an "Actions:" label as well as the "View Imaging Datasets" button (and the other two buttons if the user still has `data_entry` permission).
4. Make sure that clicking on any of the buttons brings you to the right place.

* Resolves #6673


driusan:
@CamilleBeau I think this is supposed to go to the 23.0-release branch for a bugfix, the issue has the 23.0 tag on it..

jesscall:
Looks good!

With only data entry permission:
Screen Shot 2020-07-13 at 1 10 57 PM

With only imaging permission:
Screen Shot 2020-07-13 at 1 11 27 PM

With both:
Screen Shot 2020-07-13 at 1 11 48 PM

PR #6820, opened 2020-07-08
[Core: Middleware] Fix crash when a PSR Request's User attribute is null

johnsaigle:
## Brief summary of changes

The middleware added in #6076 will crash when the request's user object is equal to null.

I'm not sure why this case occurs, but given that it does, this adds a fallback to the Factory user singleton so that the application can continue to work.

1. On `aces/master`, login and go to My Preferences. It will crash.
2. Do the same on this branch. It should load.


driusan:
It should never be null. We should investigate why if that's happening.. when are you getting a null user attribute?

johnsaigle:
@driusan Did you read the description?

driusan:
my_preferences wasn't throwing an exception for me (until I explicitly added an unconditional throw new exception for testing..), but I sent #6832 to replace this.

PR #6823, opened 2020-07-09
Lower console.log from eslint error to warning

driusan:
This changes console.log back from an eslint error to an
eslint warning. With the current settings console.log can't
be used (even in development) because `make dev` fails with
an error about console.log. Making it impossible to use
console.log in development just encourages people to misuse
other console levels, which won't be flagged by eslint.


laemtl:
It's fine to have them lowered to warning if we manage to reduce significantly the number of warnings produced by es-lint.
It is convenient to have a flag in case someone forgets to clean code after development.

PR #6831, opened 2020-07-09
getProjectList() optional userID param

### LINKED ISSUE #6829: [Permissions] Update project filter based on  user access
The way the project filter is populated should be modified accross all modules, to display project in the filter only if the current user has access to it (See #6729 for an example).

laemtl:
Modifies getProjectList() to support an optional parameter int userID, 
which, if provided, filter projects the user has permission to.

* Resolves #6829


driusan:
I don't see why this should be done. It seems like a step backwards from just using `$user->getProjectIDs()`.

christinerogers:
Sounds like it would make sense to go back over all module to verify/update which function (user vs utility class) populates the visible front-end list of projects -- if it's utility, it should probably be updated to the user class in most cases.

PR #6832, opened 2020-07-13
[Exceptions] Fix null user on ExceptionHandlingMiddleware

driusan:
The ExceptionHandlingMiddleware uses the PageDecorationMiddleware,
and as such needs to be chained after all the prerequisites of the
PageDecorationMiddleware (ie. the lorisinstance and user) are added
to the request object.

This converts it from a MiddlewareChainer interface into a normal
middleware and explicitly adds it to the BaseRouter, instead of the
top level chain.

Replaces #6820.


jesscall:
@driusan how should this be tested?

PR #6847, opened 2020-07-16
 [new_profile/create_timepoint] code improvement

lingma:
## Brief summary of changes
This is the recreation of the #5892, which was wrongly operated under the master branch, then somehow closed, in a panic I thought it was merged. Sorry.

### DIFF
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -178,7 +178,8 @@ class Utility
         return $DB->pselectColWithIndexKey(
             "SELECT CenterID as ID, Name FROM psc WHERE CenterID in (" .
             implode(', ', $params) . ")",
-            $values
+            $values,
+            'ID'
         );
     }
 
--- a/php/libraries/Utility.class.inc
+++ b/php/libraries/Utility.class.inc
@@ -115,6 +115,7 @@ class Utility
 
         return $result;
     }
+
     /**
      * Returns a list of sites in the database
      *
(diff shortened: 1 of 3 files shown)


php/libraries/Utility.class.inc
```diff
@@ -154,6 +155,39 @@ class Utility
...
+        $sites = $DB->pselect(
+            "SELECT CenterID as ID, Name FROM psc WHERE CenterID in (" .
+            implode(', ', $params) . ")",
+            $values
+        );
+
+        $list = array();
+        foreach ($sites as $site) {
+            $list[$site['ID']] = $site['Name'];
+        }
```
ridz1208, line 187: this should use pselectColWithIndexKey

https://github.com/aces/Loris/blob/master/php/libraries/Database.class.inc#L907

lingma: I just tried to use `pselectColWithIndexKey`, but the installer doesn't work now, I have deleted my old DB because the links did not work. I don't see any difference between `pselect` and `pselectColWithIndexKey` for this particular case, so I just leave it for now.

lingma: You can modify if you wish, I just don't see any necessity for this modify. The master branch still doesn't work after I fixed the installer, I am not able to test

ridz1208: pselectColWIthIndexKey will return the array properly indexed in one step making lines 184 to 187 unnecessary.

I can make the changes it's fine and thanks for reporting the errors with the installer

php/libraries/Utility.class.inc
```diff
@@ -154,6 +155,39 @@ class Utility
...
+        $sites = $DB->pselect(
+            "SELECT CenterID as ID, Name FROM psc WHERE CenterID in (" .
+            implode(', ', $params) . ")",
+            $values
+        );
+
+        $list = [];
+        foreach ($sites as $site) {
+            $list[$site['ID']] = $site['Name'];
+        }
```
ridz1208, line 187:
```suggestion
        $list = $DB->pselectColWithIndexKey(
            "SELECT CenterID as ID, Name FROM psc WHERE CenterID in (" .
            implode(', ', $params) . ")",
            $values,
            "ID"
        );
```

lingma: OK, if you are sure, I am not able to test now. I will submit a commit

ridz1208:
@CamilleBeau can you please test the dropdowns in create timepoint and new_profile ?

php/libraries/Utility.class.inc
lingma: not manually tested though, because of a problem I just reported: 6850

CamilleBeau:
> @CamilleBeau can you please test the dropdowns in create timepoint and new_profile ?

Drop downs are good!

lingma:
Manually test passed.

PR #6853, opened 2020-07-20
[Data Tables] Add ability for data tables to stream data

driusan:
This is a proof of concept. It loads the data for the data dictionary
module from a binary stream rather than a JSON object and displays the
data in the table as it is loading. The stream consists of rows separated
by the byte 0x1f, and columns separated by 0x1e. The entire stream is
terminated with the byte 0x04.

FieldOptions are moved to a different endpoint rather than crammed
into the JSON since there is no longer any JSON to cram it into.

### DIFF
--- a/modules/datadict/jsx/dataDictIndex.js
+++ b/modules/datadict/jsx/dataDictIndex.js
@@ -25,9 +25,11 @@ class DataDictIndex extends Component {
     super(props);
 
     this.state = {
-      data: {},
+      data: [],
       error: false,
       isLoaded: false,
+      isLoading: false,
+      fieldOptions: {'sourceFrom': {'abc': 'def'}},
     };
 
     this.fetchData = this.fetchData.bind(this);
@@ -38,8 +40,18 @@ class DataDictIndex extends Component {
    * Called by React when the component has been rendered on the page.
    */
   componentDidMount() {
-    this.fetchData()
-      .then( () => this.setState({isLoaded: true}));
+      // Load the field options. This comes from a separate request than the
+      // table data stream. Once the fieldOptions are loaded, we set isLoaded
+      // to true so that the page is displayed with the data that's been
+      // retrieved.
+      fetch(this.props.fieldsURL, {credentials: 'same-origin'})
+          .then((resp) => resp.json())
+          .then((data) => this.setState({fieldOptions: data, isLoaded: true}))
+        .catch((error) => {
+            this.setState({error: true});
+            console.error(error);
+        });
+    this.fetchData();
   }
 
   /**
@@ -47,14 +59,72 @@ class DataDictIndex extends Component {
    *
    * @return {object}
    */
-  fetchData() {
-    return fetch(this.props.dataURL, {credentials: 'same-origin'})
-        .then((resp) => resp.json())
-        .then((data) => this.setState({data}))
-        .catch((error) => {
-            this.setState({error: true});
-            console.error(error);
-        });
+    fetchData() {
+        const processLines = async (data) => {
+            const utf8Decoder = new TextDecoder('utf-8');
+            let row = [];
+            let colStart = -1;
+            let rowStart = 0;
+            for (let i = 0; i < data.length; i++) {
+                switch (data[i]) {
+                    case 0x1e: // end of column
+                        const rowdata = data.slice(colStart+1, i);
+                        const encoded = utf8Decoder.decode(rowdata);
+                        colStart = i;
+                        row.push(encoded);
+                        continue;
+                    case 0x1f: // end of row
+                        this.state.data.push(row);
+                        rowStart = i+1;
+                        colStart = i;
+                        row = [];
+                        continue;
+                    case 0x04: // end of stream
+                        // Force re-render
+                        this.state.data.push(row);
+                        return {remainder: [], eos: true};
+                }
+            }
+            return {remainder: data.slice(rowStart), eos: false};
+        };
+
+        return (async () => {
+            const response = await fetch(
+                this.props.dataURL,
+                {credentials: 'same-origin'},
+            );
+            const reader = response.body.getReader();
+
+            let remainder = [];
+            let doneLoop = false;
+            while (!doneLoop) {
+                await reader.read().then(({done, value}) => {
+                        let combined;
+                        if (remainder.length == 0) {
+                            combined = value;
+                        } else {
+                            combined = new Uint8Array(
+                                value.length + remainder.length
+                            );
+                            for (let i = 0; i < remainder.length; i++) {
+                                combined[i] = remainder[i];
+                            }
+                            for (let i = 0; i < value.length; i++) {
+                                combined[i+remainder.length] = value[i];
+                            }
+                        }
+                        return processLines(combined);
+                    }).then(({remainder: rem, eos}) => {
+                        this.setState({isLoading: !eos, data: this.state.data});
+                        doneLoop = eos;
+                        remainder = rem;
+                    })
+                    .catch((err) => {
+                        console.error(err);
+                        doneLoop = true;
+                    });
+            };
+    })();
   }
 
   /**
@@ -107,7 +177,7 @@ class DataDictIndex extends Component {
       return <Loader/>;
     }
 
-    let options = this.state.data.fieldOptions;
+    const options = this.state.fieldOptions;
     let fields = [
         {
             label: 'Source From',
@@ -159,8 +229,9 @@ class DataDictIndex extends Component {
     return (
         <FilterableDataTable
            name="datadict"
-           data={this.state.data.Data}
+           data={this.state.data}
            fields={fields}
+           loading={this.state.isLoading}
            getFormattedCell={this.formatColumn}
         />
     );
@@ -174,7 +245,8 @@ DataDictIndex.propTypes = {
 window.addEventListener('load', () => {
   ReactDOM.render(
       <DataDictIndex
-        dataURL={`${loris.BaseURL}/datadict/?format=json`}
+        dataURL={`${loris.BaseURL}/datadict/?format=binary`}
+        fieldsURL={`${loris.BaseURL}/datadict/fields`}
       />,
       document.getElementById('lorisworkspace')
   );
--- a/modules/datadict/jsx/dataDictIndex.js
+++ b/modules/datadict/jsx/dataDictIndex.js
@@ -29,7 +29,7 @@ class DataDictIndex extends Component {
       error: false,
       isLoaded: false,
       isLoading: false,
-      fieldOptions: {'sourceFrom': {'abc': 'def'}},
+      fieldOptions: {'sourceFrom': {}},
     };
 
     this.fetchData = this.fetchData.bind(this);
@@ -74,6 +74,9 @@ class DataDictIndex extends Component {
                         row.push(encoded);
                         continue;
                     case 0x1f: // end of row
+                        const rowdata2 = data.slice(colStart+1, i);
+                        const encoded2 = utf8Decoder.decode(rowdata2);
+                        row.push(encoded2);
                         this.state.data.push(row);
                         rowStart = i+1;
                         colStart = i;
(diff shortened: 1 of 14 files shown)


driusan:
@HenriRabalais This proof of concept is for you.. you were looking for a way to do something similar for the biobank module, I believe..

HenriRabalais:
@driusan amazing! I'll try and implement it on my end and see how it goes :)

modules/datadict/test/datadictTest.php
```diff
@@ -106,6 +106,7 @@ function testDatadictDoespageLoad()
     function testPageUIs()
     {
         $this->safeGet($this->url . "/datadict/");
+        sleep(3);
```
kongtiaowang, line 109: line 38 - 42 change to 
           'Source From'        => '.col-xs-12:nth-child(2) .col-sm-3',
            'Name'               => '.col-xs-12:nth-child(3) .col-sm-3',
            'Source Field'       => '.col-xs-12:nth-child(4) .col-sm-3',
            'Description'        => '.col-xs-12:nth-child(5) .col-sm-3',
            'Description Status' => '.col-xs-12:nth-child(6) .col-sm-3',

kongtiaowang: in this case, the html tags has been changed. 
sleep function is not necessary.

driusan: I'll try that if it doesn't work after rebasing, but I don't see why the HTML tags would have changed. I didn't do anything that should have affected the filters.

driusan: @kongtiaowang I made your changes and rebased but it still seems to be failing.

kongtiaowang: I will look into it.

modules/datadict/test/datadictTest.php
```diff
@@ -35,11 +35,11 @@ class DatadictTestIntegrationTest extends LorisIntegrationTest
     private $_loadingUI
         =  [
             'Data Dictionary'    => '#bc2 > a:nth-child(2) > div',
-            'Source From'        => '#dynamictable > thead > tr > th:nth-child(2)',
-            'Name'               => '#dynamictable > thead > tr > th:nth-child(3)',
-            'Source Field'       => '#dynamictable > thead > tr > th:nth-child(4)',
-            'Description'        => '#dynamictable > thead > tr > th:nth-child(5)',
-            'Description Status' => '#dynamictable > thead > tr > th:nth-child(6)',
+            'Source From'        => '.col-xs-12:nth-child(2) .col-sm-3',
```
kongtiaowang, line 38: Sorry my bad, typo.
It should be
            'Source From'        => '.col-xs-12:nth-child(3) .col-sm-3',
            'Name'               => '.col-xs-12:nth-child(4) .col-sm-3',
            'Source Field'       => '.col-xs-12:nth-child(5) .col-sm-3',
            'Description'        => '.col-xs-12:nth-child(6) .col-sm-3',
            'Description Status' => '.col-xs-12:nth-child(7) .col-sm-3',

modules/datadict/test/datadictTest.php
```diff
@@ -35,11 +35,11 @@ class DatadictTestIntegrationTest extends LorisIntegrationTest
...
+            'Name'               => '.col-xs-12:nth-child(3) .col-sm-3',
+            'Source Field'       => '.col-xs-12:nth-child(4) .col-sm-3',
+            'Description'        => '.col-xs-12:nth-child(5) .col-sm-3',
+            'Description Status' => '.col-xs-12:nth-child(6) .col-sm-3',
```
kongtiaowang, line 42:
```suggestion
            'Source From'        => '.col-xs-12:nth-child(3) .col-sm-3',
            'Name'               => '.col-xs-12:nth-child(4) .col-sm-3',
            'Source Field'       => '.col-xs-12:nth-child(5) .col-sm-3',
            'Description'        => '.col-xs-12:nth-child(6) .col-sm-3',
            'Description Status' => '.col-xs-12:nth-child(7) .col-sm-3',
```
It is my bad, this one should work.

modules/datadict/jsx/dataDictIndex.js
```diff
@@ -38,23 +40,94 @@ class DataDictIndex extends Component {
...
    */
-  fetchData() {
-    return fetch(this.props.dataURL, {credentials: 'same-origin'})
-        .then((resp) => resp.json())
-        .then((data) => this.setState({data}))
-        .catch((error) => {
-            this.setState({error: true});
-            console.error(error);
-        });
+    fetchData() {
```
HenriRabalais, line 62: Is there a way to place this function somewhere where other modules could reuse it?

driusan: I don't know.. is there? We didn't for the previous fetchData that's often the same between modules, so I don't know of anywhere to do it for this one.

modules/datadict/jsx/dataDictIndex.js
```diff
@@ -159,8 +232,9 @@ class DataDictIndex extends Component {
     return (
         <FilterableDataTable
            name="datadict"
-           data={this.state.data.Data}
+           data={this.state.data}
            fields={fields}
+           loading={this.state.isLoading}
```
HenriRabalais, line 237: When I was working on making streaming work, I developed a `progress bar` that calculates the percent completion of the stream and displays it on the front end. Perhaps I can add this functionality on afterward so that the user knows how close all the data is to being loaded.

HenriRabalais:
When loading the data_dictionary I'm getting the following error in the console.log
Screen Shot 2020-07-28 at 12 00 16

and the following error in the error_log
```
[Tue Jul 28 12:00:27.251083 2020] [php7:error] [pid 22828] [client 10.31.8.194:64803] PHP Fatal error:  Uncaught Error: Call to undefined method LORIS\\datadict\\DataDictRow::toJSON() in /var/www/loris/src/Http/DataIteratorBinaryStream.php:190\nStack trace:\n#0 /var/www/loris/htdocs/index.php(76): LORIS\\Http\\DataIteratorBinaryStream->read(65536)\n#1 {main}\n  thrown in /var/www/loris/src/Http/DataIteratorBinaryStream.php on line 190, referer: http://hrabalais-dev.loris.ca/datadict/?
```

driusan:
Thanks for testing, I think I rebased it after #6812 and probably need to incorporate the changes from that.

modules/datadict/jsx/dataDictIndex.js
HenriRabalais: Not at the moment - but I guess we could start a generic Utility class for certain re-useable javascript functions? Not sure what best practices are in that regard though. @maltheism do you know of a good solution to not have to duplicate this in every module?

maltheism: @HenriRabalais I like the idea you have for a generic Utility include that would have functions such as the one @driusan wrote. Here is a blog post that shows three popular approaches: https://christhedeveloper.com/react-resuable-functions/
I like the second approach which is what I believe you're describing in your comment.

driusan:
@HenriRabalais the fix seemed pretty straight forward, it's working for me locally again.. can you test it again?

modules/datadict/jsx/dataDictIndex.js
driusan: You can add it for other uses, but I don't think that would work here because there's no way to know how much data there is until you get to the end. I initially had it start with a 64 bit int to keep track of the progress, but that requires going through the data twice to first calculate the number of rows and then to actually send it. (Doing it based on byte size would have a similar problem.)

HenriRabalais:
Tested and works well! 

It's honestly hard to tell the full effect of the streaming because the load is so instantaneous. I feel like it goes straight from "No result found" to a loaded datatable. It will be interesting to try this out on a larger dataset to be able to refine the user interface.

driusan:
What modules do you have a lot of data in? I have 20,000 rows in my data dictionary and it's noticeable.. but I can do another module if it would be more useful. data dict was arbitrary because I had it open and there was enough data that it took time to load for me.

modules/datadict/jsx/dataDictIndex.js
HenriRabalais: @maltheism thanks for the link! I personally have been using the 2nd approach as I'm finding it the simplest. I guess the difficulty would be more so in ensuring that the library file remains organized.

HenriRabalais:
I don't seem to have anything over 1000 rows in any of my modules. I can try and implement this in my version of the biobank which has 30k+ entries (which I would like to do asap) but that will take longer to interface. 

If someone else that has a module with more data wants to test it that may be easier. If not, I did test the PR with console logs and some state manipulation and I'm pretty confident it works as described.

modules/datadict/jsx/dataDictIndex.js
```diff
@@ -38,23 +40,94 @@ class DataDictIndex extends Component {
...
+                        const rowdata = data.slice(colStart+1, i);
+                        const encoded = utf8Decoder.decode(rowdata);
+                        colStart = i;
+                        row.push(encoded);
+                        continue;
+                    case 0x1f: // end of row
+                        const rowdata2 = data.slice(colStart+1, i);
+                        const encoded2 = utf8Decoder.decode(rowdata2);
+                        row.push(encoded2);
+                        this.state.data.push(row);
```
HenriRabalais, line 80: are you manipulating state directly here?

modules/datadict/jsx/dataDictIndex.js
```diff
@@ -38,23 +40,94 @@ class DataDictIndex extends Component {
...
+                        colStart = i;
+                        row = [];
+                        continue;
+                    case 0x04: // end of stream
+                        // Force re-render
+                        this.state.data.push(row);
+                        return {remainder: [], eos: true};
+                }
+            }
+            return {remainder: data.slice(rowStart), eos: false};
```
HenriRabalais, line 91: what does `eos` stand for?

ridz1208: end of string ?

ridz1208: EO is pretty standard for End OF ...

maltheism:
@driusan I think the genomic browser has a lot of data but is currently only showing a limited number of rows.

modules/datadict/jsx/dataDictIndex.js
maltheism: @HenriRabalais it would be awesome for organization. I like the idea!

modules/datadict/jsx/dataDictIndex.js
driusan: end of stream

modules/datadict/jsx/dataDictIndex.js
driusan: I'll have to look into it.. I think this was intentional because I don't care if it re-renders after each row (it's probably preferable not to for performance reasons, as long as it's re-rendering at the end of each chunk returned by the reader, which it does below.)

HenriRabalais: Would it be much work to alter this function so that it accepts a callback function to update state? that way we could extract this function from the component and it could be imported for use by other modules.

modules/datadict/jsx/dataDictIndex.js
driusan: So what's the consensus on what to do here? Where should I move the function?

maltheism: @driusan maybe it should be discussed at Loris meeting. The idea of having a new Utility include for this function and other useful functions we can import for frontend development.

HenriRabalais: @maltheism sounds good to me!

driusan: It's been moved to jslib, can you re-review?

modules/datadict/jsx/dataDictIndex.js
driusan: this is done in the refactored version

maltheism (approved):

Issue #6683, opened 2020-06-08
[battery_manager] Edit permission


pierre-p-s:
**Describe the bug**
The edit permission description on the front end is ambiguous in the permission list in user accounts. ‘Add, activate, and deactivate entries in Test Battery’: this description should be changed (confusing for user support) or test plan should be more precise. Still can’t access the module even with the 2 permissions 

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to user accounts
2. In the permission list, you should see the permission ‘Add, activate, and deactivate entries in Test Battery’. I had to go to my back end to make sure that was the right permission
3. Still can't access the module even with the 2 (view and edit) permission for this module

PR #6858, opened 2020-07-22
[battery_manager] Resolved Permission Issues

### LINKED ISSUE #6682: [battery_manager] View permission
**Describe the bug**
When giving only view permission to a user, an error occurs when they try to access the module.

**To Reproduce**
1. Create a user account 
2. Give it view permission for the battery manager module
3. Access the user account
4. The module should be clickable in tools but when you access it, an error should appear 

Note:
I couldn't test the module from a user account because I cannot access module at all except with admin

HenriRabalais:
## Brief summary of changes
This PR fixes the `getCenterId` function in `test.class.inc` so that it always passes a value to the UserSiteMatch filter by converting a centerId value of `null` from the database to an array of all sites.

1. Create a user and do not assign it battery_manager view or edit permission.
2. Ensure that battery_manager cannot be accessed.
3. Assign user battery_manager view permission.
4. Ensure that battery_manager entries at user's site can be view, but there is no ability to add, activate, deactivate or edit the entries.
5. Give user battery_manager edit permission.
6. Ensure that entries can be added, activated, deactivated and edited.

* Resolves #6682


driusan:
The two issues this PR says it resolved are tagged for the 23.0.x, this should go to the 23.0 branch, not main.

driusan:
Actually hold off on that, it involves SQL, so might need some discussion..

HenriRabalais:
@driusan The SQL is just permission label changes to clear up the confusing that was mentioned in #6683

SQL/New_patches/2020-07-22_RenameBatteryManagerPermissions
```diff
@@ -0,0 +1,2 @@
+UPDATE permissions SET description='Battery Manager: Edit Test Entries' WHERE code='battery_manager_edit';
```
ridz1208, line 1: file missing `.sql` extension

ridz1208:
@HenriRabalais you will need to separate the pure bugfix from the "clarification" we dont do patches on bugfix releases, specially non-urgent ones like this one

HenriRabalais:
@ridz1208 @driusan fixed.

pierre-p-s:
I am getting this error when testing:

Screen Shot 2020-08-04 at 10 39 45 AM

ridz1208:
@HenriRabalais the `loadResources()` function was added in https://github.com/aces/Loris/pull/6640 in order to be able to check for access before loading the target entity.

because it is being called on the testEndpoint class from the `php/libraries/module.class.inc` on line 350 the testendpoint class needs to either define the `loadResources()` function (see code changes in pr #6640) or needs to extend `NDB_Page` which has an empty implementation of `loadResources()` (see same PR)

HenriRabalais:
@ridz1208 but what does that have to do with this PR? Isn't that an issue that should be solved separately?

ridz1208:
@HenriRabalais not if u want your pr to pass testing. The base on which your pr is based has these changes which means your changes need to adapt as well

ridz1208:
@pierre-p-s can you give it another go

pierre-p-s:
@ridz1208 @HenriRabalais 
Testing worked

driusan:
Travis disagrees.

ridz1208:
@driusan I disagree with TRAVIS' disagreement

ridz1208:
@driusan finally urs

PR #6859, opened 2020-07-22
[instrument_builer] Blank line fix

ridz1208:
## Brief summary of changes
this is almost an exact replica of #5616 with the difference being that this would be backwards compatible with existing linst files havin `static{@}{@}<br />` encoding a blank line.

This functionality was removed 5 years ago in the following commit https://github.com/aces/Loris/commit/52b22d7fc66720e57ebc5f4d4564cdb288222409

### DIFF
--- a/modules/instrument_builder/js/instrument_builder.instrument.js
+++ b/modules/instrument_builder/js/instrument_builder.instrument.js
@@ -90,7 +90,7 @@ var Instrument = {
             for (element of elements[i].Elements) {
                 switch (element.Type) {
                     case "line":
-                        content += '<br>\n';
+                        content += 'static{@}{@}<br>\n';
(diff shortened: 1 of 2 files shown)


ridz1208:
@driusan after having read your comments on @maltheism's PR. I tweaked a couple parameters on her branch to accommodate for what you are asking for.

HOWEVER, I didn't push to her PR directly because I find this solution (although backwards compatible) much less elegant. I would rather go with @maltheism's solution and make the LINST instrument class treat both the old and new encoding as a blank line when loading instruments (not in the builder)

ridz1208:
@zaliqarosli can you give it another shot ?

ridz1208:
PARTTYYYYYYY

Issue #6669, opened 2020-06-04
[instrument_builder] Loading existing instrument corrupts Date element


zaliqarosli:
**Describe the bug**
When loading an existing instrument and resaving the linst into a new file, the Date element's status field is lost and "undefined" is concatenated to the end of the date element line

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to 'Instrument Builder module', Build tab and build an instrument with date elements
2. Click on 'Save' tab and save the instrument.
3. Read the linst file and see the output.
4. Refresh the instrument builder, and go to the 'Load' tab. Load the instrument linst file that you just created.
5. You can add extra fields to the build if you like, but without editing the date fields, resave the instrument.
6. Read the linst file and notice changes made to the otherwise untouched file.

**What did you expect to happen?**
Reloading an existing instrument without editing the date fields should not modify the date field lines in the linst file

ridz1208:
@zaliqarosli does the merged PR solve the issue. If yes please close this issue.

zaliqarosli:
no, this issue was discovered while testing the merged PR, the reason for which this issue was mentioned there. this issue still needs a fix

cmadjar:
@maltheism Can you fix this issue?

ridz1208:
@zaliqarosli I just looked into the issue and it doesn't even require a save and load. I created an instrument in the builder as seen in the image below. when saved (even first save), it generates the following LINST. As you can see even the second standard date in the LINST is undefined.

```
table{@}bla
title{@}BLA
date{@}Date_taken{@}Date of Administration{@}2006{@}2012
static{@}Candidate_Age{@}Candidate Age (Months)
static{@}Window_Difference{@}Window Difference (+/- Days)
select{@}Examiner{@}Examiner{@}NULL=>''
date{@}standard_date{@}STANDARD{@}{@}{@}
select{@}standard_date_status{@}{@}NULL=>''{-}'not_answered'=>'Not Answered'
date{@}standard2{@}STANDARD TWO{@}{@}{@}undefined
date{@}basic{@}BASIC{@}{@}{@}BasicDate
date{@}basic2{@}BASIC TWO{@}{@}{@}BasicDate
date{@}monthyear{@}MONTH YEAR{@}{@}{@}MonthYear
date{@}monthyear2{@}MONTH YEAR TWO{@}{@}{@}MonthYear
```

[...]

PR #6863, opened 2020-07-23
[instrument_builder] Loading existing instrument corrupts Date element

maltheism:
## Brief summary of changes

When loading an existing instrument and resaving the linst into a new file, the Date element's status field is lost and "undefined" is concatenated to the end of the date element line. This PR fixes the foregoing problem.

1. Go to 'Instrument Builder module', Build tab and build an instrument with date elements
2. Click on 'Save' tab and save the instrument.
3. Read the linst file and see the output.
4. Refresh the instrument builder, and go to the 'Load' tab. Load the instrument linst file that you just created.
5. You can add extra fields to the build if you like, but without editing the date fields, resave the instrument.
6. Read the linst file and notice no changes made to the otherwise untouched file.

* Resolves https://github.com/aces/Loris/issues/6669

### DIFF
--- a/modules/instrument_builder/js/instrument_builder.instrument.js
+++ b/modules/instrument_builder/js/instrument_builder.instrument.js
@@ -121,9 +121,18 @@ var Instrument = {
                         // Add dropdown and special naming when no date format is set
                         // (i.e when addDateElement() is used)
                         if (!element.Options.dateFormat) {
+                          if (elName === '' || !elName.includes('_date')) {
                             elName = elName + "_date";
-                            dropdown = "select{@}" + elName + "_status" +
+                          }
+                          console.log(element.Description);
+                          if (!element.Description) {
+                            element.Description = '';
+                          }
+                          dropdown = "select{@}" + elName + "_status" +
                             "{@}{@}NULL=>''{-}'not_answered'=>'Not Answered'\n";
+                        }
+                        if (!element.Options.dateFormat) {
+                          element.Options.dateFormat = 'Date';
                         }
 
                         content += 'date{@}';
--- a/modules/instrument_builder/js/instrument_builder.instrument.js
+++ b/modules/instrument_builder/js/instrument_builder.instrument.js
@@ -120,7 +120,7 @@ var Instrument = {
 
                         // Add dropdown and special naming when no date format is set
                         // (i.e when addDateElement() is used)
-                        if (element.Options.dateFormat === "") {
+                        if (!element.Options.dateFormat) {
                             elName = elName + "_date";
                             dropdown = "select{@}" + elName + "_status" +
                             "{@}{@}NULL=>''{-}'not_answered'=>'Not Answered'\n";
--- a/modules/instrument_builder/js/instrument_builder.instrument.js
+++ b/modules/instrument_builder/js/instrument_builder.instrument.js
@@ -124,14 +124,13 @@ var Instrument = {
                           if (elName === '' || !elName.includes('_date')) {
                             elName = elName + "_date";
                           }
-                          console.log(element.Description);
                           if (!element.Description) {
+                            // Prevents undefined being added to linst file.
                             element.Description = '';
                           }
                           dropdown = "select{@}" + elName + "_status" +
                             "{@}{@}NULL=>''{-}'not_answered'=>'Not Answered'\n";
-                        }
-                        if (!element.Options.dateFormat) {
+                          // Makes the date be the default 'Date' if undefined.
                           element.Options.dateFormat = 'Date';
                         }
 
--- a/modules/instrument_builder/js/instrument_builder.instrument.js
+++ b/modules/instrument_builder/js/instrument_builder.instrument.js
@@ -121,9 +121,9 @@ var Instrument = {
                         // Add dropdown and special naming when no date format is set
                         // (i.e when addDateElement() is used)
                         if (!element.Options.dateFormat) {
-                          if (!element.Description) {
-                            // Prevents undefined being added to linst file.
-                            element.Description = '';
+                          if (!elName.includes('_date')) {
+                            // prevents double _date if load & save linst file.
+                            elName = elName + "_date";
                           }
                           dropdown = "select{@}" + elName + "_status" +
                             "{@}{@}NULL=>''{-}'not_answered'=>'Not Answered'\n";
--- a/modules/instrument_builder/js/instrument_builder.instrument.js
+++ b/modules/instrument_builder/js/instrument_builder.instrument.js
@@ -121,9 +121,6 @@ var Instrument = {
                         // Add dropdown and special naming when no date format is set
                         // (i.e when addDateElement() is used)
                         if (!element.Options.dateFormat) {
-                          if (elName === '' || !elName.includes('_date')) {
-                            elName = elName + "_date";
-                          }
                           if (!element.Description) {
                             // Prevents undefined being added to linst file.
                             element.Description = '';


maltheism:
thanks @zaliqarosli, do you have the steps for what you did to replicate that error? I can try to fix.

zaliqarosli:
@maltheism yeah, i followed the steps in the issue and your PR description!

maltheism:
@ridz1208 @zaliqarosli latest commit solved a missing date format if user didn't change the default option for it. Also no longer get undefined because of that bug and which is what Zaliqa found as undefined.

maltheism:
@zaliqarosli Date should actually be there as it's the date's format. I'm unsure why min/max date is missing as I did see it when I did the steps. I'll check again.

zaliqarosli:
@maltheism oh i see! do you know what uses the 'Date' format? like is it being used somewhere? if in the SQL or something else

modules/instrument_builder/js/instrument_builder.instrument.js
```diff
@@ -120,11 +120,20 @@ var Instrument = {
 
                         // Add dropdown and special naming when no date format is set
                         // (i.e when addDateElement() is used)
-                        if (element.Options.dateFormat === "") {
+                        if (!element.Options.dateFormat) {
+                          if (elName === '' || !elName.includes('_date')) {
```
zaliqarosli, line 124: should elName ever be an empty string?  its also strange im seeing that the standard date option appends '_date' to the name you provide, but not for the other two options (basic and month)

maltheism: @zaliqarosli the element name is used as the column name in the database table for the instrument. So it cannot be empty is my belief and thus it shouldn't be.

maltheism:
@zaliqarosli I noticed the date format from the instrument builder code. The reason it was usually missing is because the select field for the date format might not be changed by the user and so the onChange wouldn't fire to set it. Although it should be set to know which date format is being used since it can be different depending on what the user picks or if just leaves as the standard which is Date.

modules/instrument_builder/js/instrument_builder.instrument.js
zaliqarosli: thats what i thought. so how come you're checking for if its empty?

maltheism:
@zaliqarosli I'm working on something but will hopefully have time soon to check everything out what you found.

modules/instrument_builder/js/instrument_builder.instrument.js
maltheism: @zaliqarosli might not be needed to check if empty. I thought for sanity reasons in case it ever happens to be. It's similar to me usually checking if something is undefined first.

maltheism:
Hi @zaliqarosli I just had the time to test again.

Here is output of test1.linst
```
table{@}test1
title{@}test1
date{@}Date_taken{@}Date of Administration{@}2006{@}2012
static{@}Candidate_Age{@}Candidate Age (Months)
static{@}Window_Difference{@}Window Difference (+/- Days)
select{@}Examiner{@}Examiner{@}NULL=>''
header{@}{@}Date Example 1
date{@}dateexamplequestion_date{@}Date Example 1{@}2020{@}2021{@}Date
select{@}dateexamplequestion_date_status{@}{@}NULL=>''{-}'not_answered'=>'Not Answered'

```

And here is output of test2.linst
```
table{@}test2
title{@}test2
date{@}Date_taken{@}Date of Administration{@}2006{@}2012
static{@}Candidate_Age{@}Candidate Age (Months)
static{@}Window_Difference{@}Window Difference (+/- Days)
select{@}Examiner{@}Examiner{@}NULL=>''
header{@}{@}Date Example 1
date{@}dateexamplequestion_date{@}Date Example 1{@}2020{@}2021{@}Date
select{@}dateexamplequestion_date_status{@}{@}NULL=>''{-}'not_answered'=>'Not Answered'

```

I just made an instrument with a header and date. I do see the min and max as the following:
`date{@}dateexamplequestion_date{@}Date Example 1{@}2020{@}2021{@}Date`
Is it supposed to be somewhere else? Also if not then what might my step be that's different than yours?

modules/instrument_builder/js/instrument_builder.instrument.js
zaliqarosli: i think it's okay to check if its empty, but i don't think the step after should be to just append '_date' and not provide a nonempty $elName

modules/instrument_builder/js/instrument_builder.instrument.js
```diff
@@ -120,11 +120,20 @@ var Instrument = {
...
                             elName = elName + "_date";
-                            dropdown = "select{@}" + elName + "_status" +
+                          }
+                          console.log(element.Description);
+                          if (!element.Description) {
+                            element.Description = '';
```
zaliqarosli, line 129: same here, i think the description shouldn't be empty?

modules/instrument_builder/js/instrument_builder.instrument.js
```diff
@@ -120,11 +120,20 @@ var Instrument = {
...
+                          }
+                          dropdown = "select{@}" + elName + "_status" +
                             "{@}{@}NULL=>''{-}'not_answered'=>'Not Answered'\n";
                         }
+                        if (!element.Options.dateFormat) {
+                          element.Options.dateFormat = 'Date';
+                        }
```
zaliqarosli, line 136: you could also probably move this logic over to the if statement above where you check for !element.Options.dateFormat already

modules/instrument_builder/js/instrument_builder.instrument.js
maltheism: @zaliqarosli I think that was to get rid of an undefined being added to the instrument file.

modules/instrument_builder/js/instrument_builder.instrument.js
maltheism: @zaliqarosli nice catch! I'm unsure why I did that.

modules/instrument_builder/js/instrument_builder.instrument.js
maltheism: @zaliqarosli yah I think the logic is messed up. I think the standard is to have {something}_date. I'm going to look why I added this again and maybe make sure to put a comment if there is a reason.

maltheism:
@zaliqarosli I cleaned it up some more.

modules/instrument_builder/js/instrument_builder.instrument.js
```diff
@@ -120,10 +120,15 @@ var Instrument = {
 
                         // Add dropdown and special naming when no date format is set
                         // (i.e when addDateElement() is used)
-                        if (element.Options.dateFormat === "") {
-                            elName = elName + "_date";
```
zaliqarosli, line 124: i'm not sure if we should be removing this line?

modules/instrument_builder/js/instrument_builder.instrument.js
```diff
@@ -120,10 +120,15 @@ var Instrument = {
...
-                            dropdown = "select{@}" + elName + "_status" +
+                        if (!element.Options.dateFormat) {
+                          if (!element.Description) {
+                            // Prevents undefined being added to linst file.
+                            element.Description = '';
+                          }
+                          dropdown = "select{@}" + elName + "_status" +
                             "{@}{@}NULL=>''{-}'not_answered'=>'Not Answered'\n";
+                          // Makes the date be the default 'Date' if undefined.
+                          element.Options.dateFormat = 'Date';
```
zaliqarosli, line 131: it looks to me like this line might be the only line to add as a change

modules/instrument_builder/js/instrument_builder.instrument.js
```diff
@@ -120,10 +120,15 @@ var Instrument = {
...
                         // Add dropdown and special naming when no date format is set
                         // (i.e when addDateElement() is used)
-                        if (element.Options.dateFormat === "") {
-                            elName = elName + "_date";
-                            dropdown = "select{@}" + elName + "_status" +
+                        if (!element.Options.dateFormat) {
+                          if (!element.Description) {
+                            // Prevents undefined being added to linst file.
+                            element.Description = '';
+                          }
```
zaliqarosli, line 127: similarly to making $elname='', i don't think the description should ever be empty or undefined because it's a required field

modules/instrument_builder/js/instrument_builder.instrument.js
maltheism: @zaliqarosli thanks! Unsure why I removed it. Sorry!

modules/instrument_builder/js/instrument_builder.instrument.js
maltheism: @zaliqarosli I'll test again I forgot to with last changes. I'm unsure why I thought this was needed and what you're writing makes sense.

modules/instrument_builder/js/instrument_builder.instrument.js
maltheism: Hi @zaliqarosli I did have to add this code back:
```
if (!elName.includes('_date')) {
   // prevents double _date if load & save linst file.
  elName = elName + "_date";
}
```
It turns out that I had it origionally because the load of a previous instrument file will add `_date` again to the elName and even if already existing to the name.

See test1.linst without the above code:
```
table{@}test1
title{@}test1
date{@}Date_taken{@}Date of Administration{@}2006{@}2012
static{@}Candidate_Age{@}Candidate Age (Months)
static{@}Window_Difference{@}Window Difference (+/- Days)
select{@}Examiner{@}Examiner{@}NULL=>''
header{@}{@}test1
date{@}test1date_date{@}test1date{@}2020{@}2021{@}Date
select{@}test1date_date_status{@}{@}NULL=>''{-}'not_answered'=>'Not Answered'

```
and test2.linst loaded from test1.linst:
```
table{@}test2
title{@}test2
date{@}Date_taken{@}Date of Administration{@}2006{@}2012
static{@}Candidate_Age{@}Candidate Age (Months)
static{@}Window_Difference{@}Window Difference (+/- Days)
select{@}Examiner{@}Examiner{@}NULL=>''
header{@}{@}test1
date{@}test1date_date_date{@}test1date{@}2020{@}2021{@}Date
select{@}test1date_date_date_status{@}{@}NULL=>''{-}'not_answered'=>'Not Answered'

```

maltheism:
@zaliqarosli fixed

zaliqarosli:
LINST1 --> LINST2 on MAIN branch:

Screen Shot 2020-08-12 at 5 20 30 PM

LINST1 --> LINST2 on this PR branch:

Screen Shot 2020-08-12 at 5 20 40 PM

here are my findings:
1. Main issue number 1 is that LINST1 on MAIN doesn't append 'Date' at the end of the line for Standard Date format.
2. Main issue number 2 is that LINST2 on MAIN doesn't append the format for any of the date fields. It saves it as 'undefined'. This leads me to think that on load, the instrument builder is not even reading that last piece of the line, hence the undefined. This needs to be fixed so that the instrument builder reads what's initially in LINST1.

in this PR currently:
1. 'Date' is being appended to the line for Standard Date formats. This fixes main issue number 1 above.
2. 'Date' is being appended to the line for all the date formats, replaced 'undefined' before. This leads me to believe that the instrument is still not reading the last piece of the line, and is instead just replacing 'undefined' with 'Date' for all. This bit still needs to be fixed.

maltheism:
Thanks @zaliqarosli! Interesting I'll try to correct.

zaliqarosli:
@maltheism i have discovered futher developments in this mystery. according to this if/else block (https://github.com/aces/Loris/blob/f95525a425bdf2138759e4c51cf3b8ec5c48d993/php/libraries/NDB_BVL_Instrument_LINST.class.inc#L585), the Standard Date format should have an empty format, whilst Month/Year needs 'MonthYear' and the Basic Date format can have any other string. As part of this PR, either we leave the format as black for the standard format, or you will have to change the lines i linked above so that it specifically checks for the existence of 'Date'

laemtl:
@maltheism Is this ready for testing?

maltheism:
@laemtl not ready yet. I have to add changes for what @zaliqarosli found.

zaliqarosli:
> @maltheism i have discovered futher developments in this mystery. according to this if/else block (
> 
> https://github.com/aces/Loris/blob/f95525a425bdf2138759e4c51cf3b8ec5c48d993/php/libraries/NDB_BVL_Instrument_LINST.class.inc#L585
> 
> ), the Standard Date format should have an empty format, whilst Month/Year needs 'MonthYear' and the Basic Date format can have any other string. As part of this PR, either we leave the format as blank for the standard format, or you will have to change the lines i linked above so that it specifically checks for the existence of 'Date'

hey @maltheism, ive made the change to check for either "" or "Date" in this PR https://github.com/aces/Loris/pull/6923 so you decide to use whichever one

maltheism:
Adding the Blocked tag and please see https://github.com/aces/Loris/pull/7147 for latest changes.
Keeping other changes separate from this PR to keep comment discussions/history separate and less confusing.

zaliqarosli:
hey @maltheism i don't think i understand what is the difference between this PR and the new PR #7147. does the new PR just replace this one or are we intending for both to go in?

maltheism:
@zaliqarosli this PR should be ignored in the meantime and other one has different changes with notice needing confirmation.

zaliqarosli:
so this one can be closed?

maltheism:
@zaliqarosli I left this PR open because I want to save previous discussion here. Expecially if new PR that needs clarification is incorrect.

zaliqarosli:
we will still be able to see the previous discussions even if its closed. Its confusing to have two PRs open that potentially do the same thing. only 1 should be open. we can also re-open one if need be. i'm just confused right now which PR to focus my attention on

maltheism:
Closing for @zaliqarosli and see the following:

> Adding the Blocked tag and please see #7147 for latest changes.
Keeping other changes separate from this PR to keep comment discussions/history separate and less confusing.

> @zaliqarosli this PR should be ignored in the meantime and other one has different changes with notice needing confirmation.

PR #6868, opened 2020-07-27
[Instrument] Make instrument data non-static

driusan:
NDB_BVL_Instrument has an.. unusual.. way of dealing with instrument
data. Rather than having a class method for getting the data, there's
a static method which takes an instrument as a parameter. The result
is that data can not be cached in the object and needs to be continuously
re-loaded from the database every time if it's needed multiple times in
the same request. (For instance, calling getFieldValue for different fields
in a script.)

This adds a normal getter and caches the results in the object, deprecating
the old static loadInstanceData and its inexplicable behaviour.

### DIFF
--- a/test/unittests/NDB_BVL_Instrument_Test.php
+++ b/test/unittests/NDB_BVL_Instrument_Test.php
@@ -1131,7 +1131,7 @@ class NDB_BVL_Instrument_Test extends TestCase
      * @covers NDB_BVL_Instrument::getInstanceData
      * @return void
      */
-    function getLoadInstanceData()
+    function testGetInstanceData()
     {
         $this->_setUpMockDB();
         $this->_setTableData();
--- a/test/unittests/NDB_BVL_Instrument_Test.php
+++ b/test/unittests/NDB_BVL_Instrument_Test.php
@@ -1126,18 +1126,18 @@ class NDB_BVL_Instrument_Test extends TestCase
     }
 
     /**
-     * Test that loadInstanceData returns data from the correct database
+     * Test that getInstanceData returns data from the correct database
      *
-     * @covers NDB_BVL_Instrument::loadInstanceData
+     * @covers NDB_BVL_Instrument::getInstanceData
      * @return void
      */
-    function testLoadInstanceData()
+    function getLoadInstanceData()
     {
         $this->_setUpMockDB();
         $this->_setTableData();
         $this->_instrument->commentID = 'commentID1';
         $this->_instrument->table     = 'flag';
-        $defaults = \NDB_BVL_Instrument::loadInstanceData($this->_instrument);
+        $defaults = $this->_instrument->getInstanceData();
         $defaults['Testdate'] = '2020-01-01 00:00:00';
         $this->assertEquals(
             $defaults,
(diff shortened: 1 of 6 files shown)


laemtl:
Should we update this reference too? https://github.com/aces/Loris/blob/main/raisinbread/instruments/NDB_BVL_Instrument_aosi.class.inc#L549

driusan:
Good catch, I missed that one.. updated it.

laemtl (approved):
Checked and caching system works.

laemtl:
One last thing! Phan is failing the CI tests:

test/unittests/NDB_BVL_Instrument_Test.php:1140 PhanDeprecatedFunction Call to deprecated function \NDB_BVL_Instrument::loadInstanceData() defined at php/libraries/NDB_BVL_Instrument.class.inc:2170 (Deprecated because: use $instrumentInstance->getInstanceData() instead)
test/unittests/NDB_BVL_Instrument_Test.php:1566 PhanDeprecatedFunction Call to deprecated function \NDB_BVL_Instrument::loadInstanceData() defined at php/libraries/NDB_BVL_Instrument.class.inc:2170 (Deprecated because: use $instrumentInstance->getInstanceData() instead)
test/unittests/NDB_BVL_Instrument_Test.php:1601 PhanDeprecatedFunction Call to deprecated function \NDB_BVL_Instrument::loadInstanceData() defined at php/libraries/NDB_BVL_Instrument.class.inc:2170 (Deprecated because: use $instrumentInstance->getInstanceData() instead)
test/unittests/NDB_BVL_Instrument_Test.php:1678 PhanDeprecatedFunction Call to deprecated function \NDB_BVL_Instrument::loadInstanceData() defined at php/libraries/NDB_BVL_Instrument.class.inc:2170 (Deprecated because: use $instrumentInstance->getInstanceData() instead)

laemtl (changes requested):
Phan errors need to be fixed (see above). NDB_BVL_Instrument_Test.php was added after the PR so a rebase is needed.

test/unittests/NDB_BVL_Instrument_Test.php
```diff
@@ -1126,18 +1126,18 @@ function testGetPSCID()
...
     /**
-     * Test that loadInstanceData returns data from the correct database
+     * Test that getInstanceData returns data from the correct database
      *
-     * @covers NDB_BVL_Instrument::loadInstanceData
+     * @covers NDB_BVL_Instrument::getInstanceData
      * @return void
      */
-    function testLoadInstanceData()
+    function getLoadInstanceData()
```
laemtl, line 1134: testGetInstanceData?

laemtl:
One test is faling.

1) Loris\Tests\NDB_BVL_Instrument_Test::testNullScores
Failed asserting that 'Test Examiner1' matches expected null.

/app/test/unittests/NDB_BVL_Instrument_Test.php:1602

laemtl (approved):

laemtl:
Impressive, the tests are running really fast with Github actions.

driusan:
I think the individual test suites take about the same amount of time, but they start a lot faster.

PR #6869, opened 2020-07-27
[Instrument] Add method for loading multiple instruments data at once

driusan:
There is currently no mechanism to load the data from multiple
instrument instances at the same time efficiently and work with them
in PHP. (ie. to work the the data from multiple commentIDs for a single
instrument type in batch in a script) other than directly querying SQL.
However, this has other drawbacks such as the fact that you only know
whether the instrument data should be loaded from the flag.Data column
or a separate table once the instrument is instantiated.

This adds a bulkLoadInstanceData function which will load the data from
multiple commentIDs at the same time in a single SQL query and return an
array of \NDB_BVL_Instruments with the data loaded.

In my non-scientific benchmarking where I loaded all the data from a single
instrument type on my sandbox multiple times, and printed the value of a single
field this (unsurprisingly) resulted in significant performance gains.

The naive way where factory is called multiple times took approximately
12 seconds:

```
foreach($commentIDs as $cid) {
    $inst = \NDB_BVL_Instrument::factory('adi_r_proband', $cid);
    print $inst->getFieldValue('informant_relation');
}
```

The slightly smarter way where we avoid the overhead of re-parsing the
instrument took approximately 1.2 seconds:

```
$inst = \NDB_BVL_Instrument::factory('adi_r_proband');
foreach($commentIDs as $cid) {
    // note: must hack NDB_BVL_Instrument to make instanceData
    // public for this to work.
    $inst->instanceData = null;
    $inst->commentID = $cid;
    print $inst->getFieldValue('informant_relation');
}
```

The new function took approximately 0.2s:

```
$inst = \NDB_BVL_Instrument::factory('adi_r_proband');
$data = $inst->bulkLoadInstanceData($commentIDs);
foreach($data as $i) {
    print $i->getFieldValue('informant_relation');
}
```

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2122,14 +2122,12 @@ abstract class NDB_BVL_Instrument extends NDB_Page
                 "SELECT Data FROM flag WHERE CommentID=:CID",
                 ['CID' => $this->getCommentID()]
             );
-
             $this->instanceData = json_decode($jsondata, true) ?? [];
         } else {
             $defaults = $db->pselectRow(
                 "SELECT * FROM $this->table WHERE CommentID=:CID",
                 ['CID' => $this->getCommentID()]
             );
-
             $this->instanceData = $defaults ?? [];
         }
         return $this->instanceData;
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2173,8 +2173,8 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      * Loads multiple instruments representing multiple commentIDs from the
      * database in bulk. This function must be called on an instrument instantiated
      * as \NDB_BVL_Instrument::factory('instrumentname') without a comment ID, and
-     * will return an array of copies of the instrument type with the data for the given
-     * commentIDs loaded.
+     * will return an array of copies of the instrument type with the data for
+     * the given commentIDs loaded.
      *
      * @param $commentIDs string[] A list of commentIDs to load in bulk.
      *
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2168,6 +2168,82 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     }
 
     /**
+     * Loads multiple instruments representing multiple commentIDs from the
+     * database in bulk. This function must be called on an instrument instantiated
+     * as \NDB_BVL_Instrument::factory('instrumentname') without a comment ID, and
+     * will an array of copies of the instrument type with the data for the given
+     * commentIDs loaded.
+     *
+     * @param $commentIDs string[] A list of commentIDs to load in bulk.
+     *
+     * @return \NDB_BVL_Instrument[]
+     */
+    function bulkLoadInstanceData(iterable $commentIDs): array
+    {
+        if ($this->commentID !== null && $this->commentID !== '') {
+            throw new \LogicException(
+                "Must bulk load from instrument loaded without commentID"
+            );
+        }
+        $db = \Database::singleton();
+
+        $prepBindings = [];
+        $prepValues   = [];
+
+        $i = 0;
+        foreach ($commentIDs as $commentID) {
+            $i++;
+
+            $prepBindings[]      = ":cid$i";
+            $prepValues["cid$i"] = $commentID;
+        }
+
+        if ($this->jsonData) {
+            // We filter out if Data is null for 2 reasons.
+            // 1. Otherwise getInstanceData will re-query the database since
+            //    $newinst->instanceData is null and commentID is set
+            // 2. The Data column being null in the flag table is equivalent
+            //    to there not being a row in the table in the non-JSON
+            //    case (where it will also not get returned because there's
+            //    no row in the table.)
+            $jsondata = $db->pselect(
+                "SELECT CommentID, Data FROM flag WHERE Data IS NOT NULL" .
+                " AND CommentID IN ("
+                . join(',', $prepBindings) . ')',
+                $prepValues,
+            );
+            return array_map(
+                function ($row) {
+                    $newinst = clone $this;
+
+                    $newinst->commentID    = $row['CommentID'];
+                    $newinst->instanceData = json_decode($row['Data'], true);
+
+                    return $newinst;
+                },
+                $jsondata
+            );
+        } else {
+            $defaults = $db->pselect(
+                "SELECT * FROM $this->table WHERE CommentID IN ("
+                . join(',', $prepBindings) . ')',
+                $prepValues,
+            );
+            return array_map(
+                function ($row) {
+                    $newinst = clone $this;
+
+                    $newinst->commentID    = $row['CommentID'];
+                    $newinst->instanceData = $row;
+
+                    return $newinst;
+                },
+                $defaults
+            );
+        }
+    }
+
+    /**
      * Calculates the candidate's age at the time of this instrument.
      * If Date_taken is passed, this function  will calculate the age as of
      * the value of that argument. This is used, for instance, in
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2209,8 +2209,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             //    case (where it will also not get returned because there's
             //    no row in the table.)
             $jsondata = $db->pselect(
-                "SELECT CommentID, Data FROM flag WHERE Data IS NOT NULL" .
-                " AND CommentID IN ("
+                "SELECT CommentID, Data FROM flag WHERE CommentID IN ("
                 . join(',', $prepBindings) . ')',
                 $prepValues,
             );
@@ -2219,7 +2218,10 @@ abstract class NDB_BVL_Instrument extends NDB_Page
                     $newinst = clone $this;
 
                     $newinst->commentID    = $row['CommentID'];
-                    $newinst->instanceData = json_decode($row['Data'], true);
+                    $newinst->instanceData = json_decode(
+                        $row['Data'],
+                        true,
+                    ) ?? [];
 
                     return $newinst;
                 },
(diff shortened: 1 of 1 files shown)


driusan:
Blocked by #6868

laemtl:
Waiting for #6868 to be merged and this PR rebased before testing, as some changes were made in #6868 which can possibly affects this PR.

driusan:
@laemtl #6868 is merged and this is rebased

laemtl:
There is some minot linting errors to fix:
FILE: ...nner/work/Loris/Loris/php/libraries/NDB_BVL_Instrument.class.inc
 2142 | ERROR | [x] Equals sign not aligned with surrounding
      |       |     assignments; expected 11 spaces but found 1 space
 2148 | ERROR | [x] Equals sign not aligned with surrounding
      |       |     assignments; expected 11 spaces but found 1 space

laemtl (approved):
Tested and work perfectly.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2190,6 +2190,82 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+            // 2. The Data column being null in the flag table is equivalent
+            //    to there not being a row in the table in the non-JSON
+            //    case (where it will also not get returned because there's
+            //    no row in the table.)
+            $jsondata = $db->pselect(
+                "SELECT CommentID, Data FROM flag WHERE Data IS NOT NULL" .
+                " AND CommentID IN ("
+                . join(',', $prepBindings) . ')',
+                $prepValues,
+            );
```
ridz1208, line 2237: Not sure if this makes a difference here at all, but it might later down the line i think.

given this approach you are getting different results:
1. between JSON instruments and SQL table instruments
2. between `bulkLoadInstanceData()` and `loadInstanceData()` 

because you are not querying any JSON data instrument which has no values saved, your results will actually look like the instrument never was populated in the flag table in the first place while for SQL table instruments you would actually get an entry in the array for the commentID and an associated array with the CommentID from the instrument table inside... making the returned result difficult to manage for any code calling this function because the return seems inconsistent.

Even loadInstanceData effectively returns an empty array for instruments with no Data (which is not exactly the same as SQL table instruments but still better than null).

ie. if commentID `123456MTL000999999` has no instrument data:
If it's a `jsonData=false` you get
```
[ 
    Instrument (CommentID:`123456MTL000999999`, data: [CommentID=>`123456MTL000999999`])
]
``` 
If its a `jsonData=true` you get
```
[]
```

I think there's value in matching the returns regardless of the type of data saving we decide to go for... or at least returning
```
[ 
    Instrument (CommentID:`123456MTL000999999`, data: [])
]
```

ridz1208 (changes requested):

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2190,6 +2190,82 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         }
     }
 
+    /**
+     * Loads multiple instruments representing multiple commentIDs from the
+     * database in bulk. This function must be called on an instrument instantiated
+     * as \NDB_BVL_Instrument::factory('instrumentname') without a comment ID, and
+     * will an array of copies of the instrument type with the data for the given
+     * commentIDs loaded.
```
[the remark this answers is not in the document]
driusan: I went with "return" instead of "create"

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2190,6 +2190,82 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+     *
+     * @param $commentIDs string[] A list of commentIDs to load in bulk.
+     *
+     * @return \NDB_BVL_Instrument[]
+     */
+    function bulkLoadInstanceData(iterable $commentIDs): array
+    {
+        if ($this->commentID !== null && $this->commentID !== '') {
```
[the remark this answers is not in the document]
driusan: does that have any benefit here? The if statement is already written and is more technically accurate.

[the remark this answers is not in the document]
ridz1208: not at all just curious

php/libraries/NDB_BVL_Instrument.class.inc
driusan: It's been a while since I wrote this code so I'll have to investigate. From what I remember (because of the comment..) the problem was that getInstanceData starts

```
   public function getInstanceData(): array
    {
        if ($this->instanceData !== null) {
            return $this->instanceData;
        }
```

So if this code sets instanceData to null (because the column is null in the database) it doesn't, in fact, bulk load the data, but still runs an SQL statement per row and defeats the purpose of using this function.. but if that was the only problem an empty array should suffice.

ridz1208: either empty array or an array with only the commentID as an element would do the trick I think and mimick what happens with SQL instruments i believe

driusan: I don't think that's true. With SQL table instruments the data is loaded with

```
SELECT * FROM $this->table WHERE CommentID
```

And there is no row in `$this->table` unless data entry has been done. (At least according to point 2 in the comment, if that's still accurate.)

ridz1208: I am 99.9 % sure the row in the instrument gets created when the battery is populated. along with the DDE_ row as well regardless if the instrument is DDE or not.

you give too much credit to our current instrument infrastructure

driusan: The flag table gets populated by NDB_BVL_Battery->addInstrument (along with the DDE row, regardless of if it's a DDE instrument...) but not the table. The instrument table either gets populated the first time the instrument is accessed in the frontend or the first time data is saved, I can't remember which. (The latter would make more sense but something tells me it's the former..)

laemtl: I think the instrument table is populated when we start the stage.

driusan: I dug a little more and addInstrument does insert into the table unconditionally, I just missed it the first time I was looking at it. It appears to have never been updated to use json instrument data. So from what I can tell (Ihaven't tested yet..) you can't currently include a json data instrument in a battery.

ridz1208: @laemtl is that true? have you tried start next stage on jsonData instruments ?

laemtl: Yes

driusan: How does it get past this line in addInstrument?

```
    // instantiate instrument object to get table name attribute
        $obj = NDB_BVL_Instrument::factory($testName, '', '');

        // insert into the test table
        $DB->insert($obj->table, ['CommentID' => $commentID]);
```

ridz1208: it doesnt #7155

ridz1208:
@driusan will you be fixing this one ?

driusan:
Eventually.

php/libraries/NDB_BVL_Instrument.class.inc
driusan: I think this thread is now resolved.

driusan:
I've rebased this and fixed the issue in the long thread.

I didn't retest this, whoever tests this should ensure that when bulkLoadInstanceData is called on a CommentID that has a null Data column for a JSON instrument, calling `getInstanceData` on the instance returned doesn't re-query the database.

driusan:
@laemtl I just saw that you tested/approved this in November.. can you re-test the case in my comment above since the behaviour has changed?

ridz1208 (approved):

laemtl:
@driusan I get an SQL error if $commentIDs is empty:

```
$commentIDs = [];
$inst = \NDB_BVL_Instrument::factory('bmi');
$data = $inst->bulkLoadInstanceData($commentIDs);
foreach($data as $i) {
    print($i->getFieldValue('informant_relation'));
}
```

[Tue Mar 30 18:14:22.877181 2021] [php7:notice] [pid 9760] [client ::1:39160] Could not execute SELECT * FROM bmi WHERE CommentID IN (). Stack trace#0 /var/www/html/loris-core/php/libraries/Database.class.inc(745): Database->execute()\n#1 /var/www/html/loris-core/php/libraries/NDB_BVL_Instrument.class.inc(2225): Database->pselect()\n#2 /var/www/html/loris-core/modules/dashboard/php/dashboard.class.inc(121): NDB_BVL_Instrument->bulkLoadInstanceData()\n#3 /var/www/html/loris-core/php/libraries/Module.class.inc(275): LORIS\\dashboard\\Dashboard->__construct()\n#4 /var/www/html/loris-core/php/libraries/Module.class.inc(342): Module->loadPage()\n#5 /var/www/html/loris-core/src/Middleware/ResponseGenerator.php(50): Module->handle()\n#6 /var/www/html/loris-core/src/Middleware/AuthMiddleware.php(63): LORIS\\Middleware\\ResponseGenerator->process()\n#7 /var/www/html/loris-core/src/Router/ModuleRouter.php(74): LORIS\\Middleware\\AuthMiddleware->process()\n#8 /var/www/html/loris-core/src/Middleware/ExceptionHandlingMiddleware.php(35): LORIS\\Router\\ModuleRouter->handle()\n#9 /var/www/html/loris-core/src/Router/BaseRouter.php(112): LORIS\\Middleware\\ExceptionHandlingMiddleware->process()\n#10 /var/www/html/loris-core/src/Middleware/ResponseGenerator.php(50): LORIS\\Router\\BaseRouter->handle()\n#11 /var/www/html/loris-core/src/Middleware/ContentLength.php(52): LORIS\\Middleware\\ResponseGenerator->process()\n#12 /var/www/html/loris-core/htdocs/index.php(48): LORIS\\Middleware\\ContentLength->process()\n#13 {main}, referer: https://localhost/
[Tue Mar 30 18:14:22.877319 2021] [php7:notice] [pid 9760] [client ::1:39160] You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ')' at line 1, referer: https://localhost/

laemtl:
@driusan Everything else works perfectly. Approving.

PR #6876, opened 2020-07-29
[Instrument] Rename Data_entry_completion_status  & move to its own column in flag

### LINKED ISSUE #6746: Mistaken field created in _nullStatus function of NDB_BVL_Instrument
**Describe the bug**
The function _nullStatus in NDB_BVL_Instrument class checks for fields ending with _status having set values, and sets their corresponding fields without _status to null. An array of data from flag often includes the field Data_entry_completion_status. When passing this array through _saveValues function to save, this results in a mistake where a new field Data_entry_completion is created and set to null.

A possible solution to this would be to move the Data_entry_completion_status field into its own column in the flag table.

CamilleBeau:
## Brief summary of changes
This PR moves Data_entry_completion_status field to its own column in flag following discussion and renames to Required_elements_completed. This is accomplished with the help of a patch and a single-use script (the patch must be run before the script).

1. Run patches `SQL/New_patches/2020-07-29_Required_elements_completed_flag_column.sql` and `SQL/Cleanup_patches/2020-12-08-remove_dataEntryCompletionStatus_conflict_resolver.sql`
2. Run `php tools/single_use/Set_Required_elements_completed_flag.php`. Check that the script successfully sets the new column to Y for instruments that already have Data_entry_completion_status=Complete and to N for instruments that had Data_entry_completion_status=Incomplete. Make sure that this holds for both JSON and non JSON instruments.
3. Run `php tools/single_use/Remove_Data_entry_completion_status_instr_column.php`
4. Check that the Data column of flag no longer contains any Data_entry_completion_status values.
5. Check in the backend that Data_entry_completion_status is only set to "Complete" when all required elements have been filled & saved
6. Check on the front end that the functionality for being able to set the instrument to complete remains the same (e.g. that you only have the option of setting Data_entry to "Complete" if Required_elements_completed='Y'

* Resolves #6746


php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1568,8 +1577,18 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     function getDataEntryCompletionStatus(): string
     {
-        $data = NDB_BVL_Instrument::loadInstanceData($this);
-        return $data["Data_entry_completion_status"];
+        $db = \Database::singleton();
+        if ($this->jsonData) {
+            // If JSON instrument, get Data_entry_completion_status
+            // from flag table column
+            return $db->pselectOne(
```
driusan, line 1584: Why does this behave differently for JSON instruments and PHP instruments? We should be getting the status from the same place in either case.

CamilleBeau: In the case of a json instrument, loadInstanceData loads the `Data` column from the flag table. Since we are migrating the Data_entry_completion_status field out of the Data column and into its own column, it will not be included in loadInstanceData.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ requesting a new account and will be displayed in the User Accounts module (PR #
 database (PR #5260)
 - New documentation for file permissions has been added to the README.md file. (PR #5323)
 - Dashboard study progression section performance improvement (PR #5887)
+- Data_entry_completion_status given its own column in flag (PR #6876)
```
driusan, line 50: This is in the wrong place, 23 is already released.

CamilleBeau:
@kongtiaowang @AlexandraLivadas Would you be able to help me figure out what's going on with travis here?

CamilleBeau:
@kongtiaowang I am a little bit confused about adding back the `Data_entry_completion_status` to the `testLoadInstanceData` and `setFakeTableData` functions in `NDB_BVL_Instrument_Test` since we are removing this column from instrument tables in this PR.

test/unittests/NDB_BVL_Instrument_Test.php
```diff
@@ -1146,11 +1146,11 @@ function testGetInstanceData()
                 'SessionID'      => '123',
                 'Test_name'      => 'Test_name1',
                 'CommentID'      => 'commentID1',
-                'Data_entry'     => null,
-                'Administration' => null,
-                'Validity'       => null,
-                'Exclusion'      => null,
-                'Flag_status'    => null,
+                'Data_entry'     => '',
+                'Administration' => '',
```
kongtiaowang, line 1150:
```suggestion
                'Data_entry'     => '',
                'Data_entry_completion_status' => 'Incomplete',
                'Administration' => '',
```

test/unittests/NDB_BVL_Instrument_Test.php
```diff
@@ -1146,11 +1146,11 @@ function testGetInstanceData()
...
+                'Validity'       => '',
+                'Exclusion'      => '',
+                'Flag_status'    => '',
```
kongtiaowang, line 1153:
```suggestion
                'Data_entry'     => null,
                'Data_entry_completion_status' => 'Incomplete',
                'Administration' => null,
                'Validity' => null,
                'Exclusion' => null,
                'Flag_status' => null,
```

kongtiaowang (commented):
In this case, 
        $this->_instrument->table     = 'flag';
        $defaults = $this->_instrument->getInstanceData();
It means that the instranceData from flag table.
you have inserted 'Data_entry_completion_status' into flag table.That is why the 'Data_entry_completion_status' is here.

CamilleBeau:
`Data_entry_completion_status` should not be present on line 1150 of `NDB_BVL_Instrument_Test` since this is testing getInstanceData, which takes from the instrument table data or the _Data column_ of flag. This PR removes `Data_entry_completion_status` from both of those places and puts it in its _own_ column in flag, so getInstanceData should not be returning it. 

However, removing the field from the array causes the test suite to fail. Following discussion with Shen, this may be because the unittest does not run the scripts in this PR. The field can therefore be removed from the test file in a future PR after all of the scripts from this PR are run and the deprecation of `Data_entry_completion_status` in instrument data is complete.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -818,15 +828,18 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         } else {
             $oldData = [];
         }
+
         $newData = array_merge($oldData ?? [], $values);
 
+        $flagDataToUpdate['Data'] = empty($newData) ? null : json_encode($newData);
```
ridz1208, line 834: @CamilleBeau in order to maintain previous behaviour, you should default to and empty array, not null; **unless** you noticed an issue with defaulting to `[]`. (you are doing the exact opposite here where you are going from empty to null)

Similarly, @driusan any reason you default to `[]` and not `null` in your code before this change ?

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -818,15 +828,18 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+
         // Save the JSON to the flag.Data column.
         //
         // json_encode ensures that this is safe. If we use the safe wrapper,
         // HTML encoding the quotation marks will make it invalid JSON.
         $db->unsafeUpdate(
```
ridz1208, line 840: @CamilleBeau there is no reason to update the `data_entry_completion_status` in an unsafe way. leave that for the instrument, make a different query for this field

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1574,8 +1587,11 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     function getDataEntryCompletionStatus(): string
     {
-        $data = $this->getInstanceData();
-        return $data["Data_entry_completion_status"];
+        $db = Database::singleton();
+        return $db->pselectOne(
+            "SELECT Data_entry_completion_status FROM flag WHERE CommentID=:CID",
+            ['CID' => $this->getCommentID()]
+        ) ?? '';
```
ridz1208, line 1594: I think this Should default to `Incomplete` or throw an error depending on if this value is always fille din or not

ridz1208 (changes requested):
1. you forgot the necessary changes to tools script (scripts that generate instrument tables mostly)

2. you are missing all changes in the conflict resolver tables where `Data_entry_completion_status` might appear as a conflict as well as the removal of that value from the `$_doubleDataEntryDiffIgnoreColumns` array of the instrument classes (this array exists in multiple places you should search for it)

3. It's a "convoluted" implementation of a change that should have made a bunch of stuff simpler, not more complex. You are removing the `Data_entry_completion_status` from within instrument data and yet choose to still use the instrument save function to save it when the saving functionality is already pretty-wrapped in a `_setDataEntryCompletionStatus` method... seems like the intuitive way is to modify that to simply store the value in `flag` rather then modify the save function to handle a field that is no longer in it's scope.

php/libraries/NDB_BVL_Instrument.class.inc
driusan: the coalesce to `[]` was because `array_merge` takes arrays as arguments, null is a type error.

modules/imaging_qc/php/imaging_qc.class.inc
```diff
@@ -151,6 +151,7 @@ class Imaging_QC extends \NDB_Menu_Filter
         JOIN session s ON (f.SessionID=s.ID)
         JOIN candidate c ON (s.CandID=c.CandID)
         LEFT JOIN mri_parameter_form m ON (m.CommentID=f.CommentID)
+        LEFT JOIN flag f2 ON (f2.CommentID=m.CommentID)
```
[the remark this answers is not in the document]
CamilleBeau: I don't remember. Fixing it.

CamilleBeau:
> 3\. It's a "convoluted" implementation of a change that should have made a bunch of stuff simpler, not more complex. You are removing the `Data_entry_completion_status` from within instrument data and yet choose to still use the instrument save function to save it when the saving functionality is already pretty-wrapped in a `_setDataEntryCompletionStatus` method... seems like the intuitive way is to modify that to simply store the value in `flag` rather then modify the save function to handle a field that is no longer in it's scope.

I addressed most of your comments in the latest commit. For this one, I put a direct update to flag in the `_setDataEntryCompletionStatus` method, and removed the branching for data_entry_completion_status in the _save function. Is this what you meant?

SQL/0000-00-00-schema.sql
```diff
@@ -275,6 +275,7 @@ CREATE TABLE `flag` (
   `Test_name` varchar(255) NOT NULL default '',
   `CommentID` varchar(255) NOT NULL default '',
   `Data_entry` enum('In Progress','Complete') default NULL,
+  `Data_entry_completion_status` enum('Incomplete','Complete') NOT NULL default 'Incomplete',
```
driusan, line 278: People have always gotten confused between the Data entry flag and the DECS status, maybe this would be a good opportunity to rename the column to something that more accurately describes what it's for and doesn't sound like a different user facing column?

ridz1208: @driusan agreed... suggestions ?

Cause as far as I see it... it seems like it's current name is exactly what it does

driusan: I don't know, the flag is used to determine if you can click the "Data Entry" button in the frontend, right? Maybe something like `Can_mark_data_entry_complete`? Though to be honest I'm not a fan of that name either.

ridz1208: this flag is used to determine if all required fields are completed.. that being the case enables/disables a bunch of control panel stuff

how about a boolean `required_elements_completed` Y/N?

CamilleBeau: I like `required_elements_completed`

driusan: LORIS Meeting Discussion: go forward with Rida's suggestion.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1558,45 +1558,59 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+                . "'Y' or 'N'<br>\n"
             );
         }
-
-        $this->_save(["Data_entry_completion_status" => $status]);
+        $DB  = Database::singleton();
+        $DB->update(
+            "flag",
+            ["Required_elements_completed" => $status],
+            $this->getCommentID()
```
CamilleBeau, line 1612:
```suggestion
            ['CommentID' => $this->getCommentID()]
```

CamilleBeau:
@ridz1208 After making all of the changes in this PR & running source_rb from the branch, I get the following error: 

image

Do you know what's causing this / what I should change?

ridz1208:
@CamilleBeau My guess would be that you have an invalid default value

test/unittests/NDB_BVL_Instrument_Test.php
```diff
@@ -1680,7 +1681,8 @@ function testClearInstrument()
         );
         $this->_DB->run("DROP TEMPORARY TABLE IF EXISTS conflicts_unresolved");
         $this->assertEquals(null, $data['Examiner']);
-        $this->assertEquals('Incomplete', $data['Data_entry_completion_status']);
+        $status = $this->_instrument->getRequiredElementsCompletedFlag();
+        $this->assertEquals('Incomplete', $status);
```
CamilleBeau, line 1685:
```suggestion
        $this->assertEquals('N', $status);
```

php/libraries/NDB_BVL_Instrument.class.inc
CamilleBeau: @ridz1208 Figured out why unsafeUpdate was being used

>  // json_encode ensures that this is safe. If we use the safe wrapper,
>  // HTML encoding the quotation marks will make it invalid JSON.

HTML double escape for the Data column in flag

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -823,7 +823,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         //
         // json_encode ensures that this is safe. If we use the safe wrapper,
         // HTML encoding the quotation marks will make it invalid JSON.
-        $db->unsafeUpdate(
+        $db->update(
```
ridz1208, line 826: In this case, it is unsafe to use `unsafeUpdate` but it's also problematic to use `update()`. 

with update(), the json daat saved will be `htmlspecialchar`ed and thus will no longer be a valid JSON format, when the `getinstancedata()` function attempts to run decode on it, the decoding will fail.

My proposal:
1. use unsafeupdate to insert the data in the database
2.  modify the loadInstance data to run `HTMLSpecialChars()` on the decoded data only as follows
   ```
        if ($this->jsonData) {
            $jsondata = $db->pselectOne(
                "SELECT Data FROM flag WHERE CommentID=:CID",
                ['CID' => $this->getCommentID()]
            );

            $this->instanceData = json_decode($jsondata, true) ?? [];
            $this->instanceData = array_map(
                'htmlspecialchars',
                json_decode($jsondata, true) ?? []
            );
        } else {
   ```


@driusan thoughts?

ridz1208: @driusan you got feedback on this ? security wise

CamilleBeau: @driusan Just wanted to check if this has made it's way to you yet!

cmadjar:
@CamilleBeau could you rebase?
@driusan once rebase, ready for you to review.

CamilleBeau:
This requires input from @driusan before completing

php/libraries/NDB_BVL_Instrument.class.inc
driusan: I'm having a bit of trouble wrapping my head around the proposal but I *think* it makes sense in that the behaviour should be similar to what we have now. My concerns are:

1. What about non-HTML places that use instanceData? (ie. API responses) Maybe we should do it right and just do the `htmlspecialchars` at the template level. Or would that break other things that depend on the current behaviour?
2. It's not really clear to me why we're changing this in this PR at all. What does this have to do with moving data out and into its own column?

ridz1208: @driusan 

> What about non-HTML places that use instanceData? (ie. API responses) Maybe we should do it right and just do the htmlspecialchars at the template level. Or would that break other things that depend on the current behaviour?

Good point but it's probably not the concern of this PR

>  It's not really clear to me why we're changing this in this PR at all. What does this have to do with moving data out and into its own column?

Also good point, my guess is we discovered the security flaw while testing this ? I don't see why it;s related.

ridz1208: @CamilleBeau i think you should just put it back to unsafe update and before we release we have to address the instrument encoding issue which is LONG overdue

tools/single_use/Set_Required_elements_completed_flag.php
```diff
@@ -0,0 +1,74 @@
...
+
+    if (isset($decs)) {
+        // change value from complete / incomplete to Y / N
+        if ($decs === 'Complete') {
+            $dataToUpdate['Required_elements_completed'] = 'Y';
+        } elseif ($decs === 'Incomplete') {
+            $dataToUpdate['Required_elements_completed'] = 'N';
+        }
+    }
+
```
ridz1208, line 61:
```suggestion
// $instrument->save() is not used here in order to explicitly REMOVE the 
// Data_entry_completion_status field from the JSON string saved in the Data 
// column in flag.
```

ridz1208 (changes requested):
a few minor stuff

raisinbread/RB_files/RB_consent.sql
```diff
@@ -1,8 +1,8 @@
 SET FOREIGN_KEY_CHECKS=0;
 TRUNCATE TABLE `consent`;
 LOCK TABLES `consent` WRITE;
-INSERT INTO `consent` (`ConsentID`, `Name`, `Label`, `ConsentGroupID`) VALUES (1,'study_consent','Consent to Study',1);
-INSERT INTO `consent` (`ConsentID`, `Name`, `Label`, `ConsentGroupID`) VALUES (2,'raisin_consent','Consent to Raisin',2);
-INSERT INTO `consent` (`ConsentID`, `Name`, `Label`, `ConsentGroupID`) VALUES (3,'bread_consent','Consent to Bread',2);
+INSERT INTO `consent` (`ConsentID`, `Name`, `Label`, `ConsentGroupID`) VALUES (1,'study_consent','Consent to Study', '1');
```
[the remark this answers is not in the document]
CamilleBeau: I don't remember but I don't think it was manually

CamilleBeau:
@kongtiaowang Do we still use test/RBData.sql ? I'm not sure if I should be changing that here.

kongtiaowang:
@CamilleBeau We don't use test/RBdata.sql anymore.

CamilleBeau:
> @CamilleBeau We don't use test/RBdata.sql anymore.

Thanks! This is ready to go then @ridz1208

raisinbread/RB_files/RB_consent.sql
```diff
@@ -1,8 +1,8 @@
...
+INSERT INTO `consent` (`ConsentID`, `Name`, `Label`, `ConsentGroupID`) VALUES (2,'raisin_consent','Consent to Raisin', '2');
+INSERT INTO `consent` (`ConsentID`, `Name`, `Label`, `ConsentGroupID`) VALUES (3,'bread_consent','Consent to Bread', '2');
```
ridz1208, line 6:
```suggestion
INSERT INTO `consent` (`ConsentID`, `Name`, `Label`, `ConsentGroupID`) VALUES (1,'study_consent','Consent to Study', 1);
INSERT INTO `consent` (`ConsentID`, `Name`, `Label`, `ConsentGroupID`) VALUES (2,'raisin_consent','Consent to Raisin', 2);
INSERT INTO `consent` (`ConsentID`, `Name`, `Label`, `ConsentGroupID`) VALUES (3,'bread_consent','Consent to Bread', 2);
```

ridz1208: manually undid the changes

tools/single_use/Set_Required_elements_completed_flag.php
```diff
@@ -0,0 +1,81 @@
...
+            $cmid,
+            ''
+        );
+    } catch (Exception $e) {
+        echo "Could not instantiate {$data['Test_name']} for {$cmid}.\n";
+        continue;
+    }
+
+    $dataArray = $instrument->getInstanceData();
+    $desc      = $data['Data_entry_completion_status'];
```
ridz1208, line 57: typo
```suggestion
    $decs      = $data['Data_entry_completion_status'];
```

ridz1208 (changes requested):
minor changes but otherwise it looks good !!

ridz1208 (approved):

ridz1208:
@laemtl @zaliqarosli can you guys give this a test please ?

cmadjar:
@laemtl @zaliqarosli could you please test this? Thank you :-)

zaliqarosli:
error:

```
lorisadmin@zrosli-dev:/var/www/loris$ php tools/single_use/Set_Required_elements_completed_flag.php 

*** Set Data_entry_completion_satus in flag table***

This script is to be run for one time use only to migrate and rename the 
Data_entry_completion_status field. 

If the instrument has a valid table in the database, Data_entry_completion_status 
will be taken for back-population from the instrument table. Otherwise, it will be 
taken from the data column of the flag table. An update on the flag table will set 
the Required_elements_completed column to the Data_entry_completion_status. 

After running this script, run Remove_Data_entry_completion_status_instr_column.php 
to generate a DROP COLUMN patch to remove the old column from instrument tables.

Migrating Data_entry_completion_status for 300001MTL0011211468524336.
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to NDB_BVL_Instrument::factory() must be an instance of LORIS\LORISInstance, string given, called in /var/www/loris/tools/single_use/Set_Required_elements_completed_flag.php on line 49 and defined in /var/www/loris/php/libraries/NDB_BVL_Instrument.class.inc:224
Stack trace:
#0 /var/www/loris/tools/single_use/Set_Required_elements_completed_flag.php(49): NDB_BVL_Instrument::factory()
#1 {main}
  thrown in /var/www/loris/php/libraries/NDB_BVL_Instrument.class.inc on line 224
```

CamilleBeau:
@zaliqarosli Thank you! Fixed.

zaliqarosli:
@CamilleBeau second round test:

running these queries:

`select CommentID, Data_entry_completion_status from $instrument where Data_entry_completion_status='Complete'; `
`select CommentID, Test_name, Required_elements_completed from flag where Required_elements_completed ='Y' and Test_name=$instrument;`

does not always give me the same number of results, when i should be expected the number of rows to match right?

example:

Screen Shot 2021-08-18 at 2 39 39 PM

there are some instruments however whose data seem to match (testing on raisinbread):

Screen Shot 2021-08-18 at 2 42 36 PM

CamilleBeau:
> does not always give me the same number of results, when i should be expecting the number of rows to match right?

@zaliqarosli  I think the raisinbread data has some inconsistencies / outdated info. I just looked into this on the main branch and, for example, the first commentID there (300319OTT319532261476556345) returns "Incomplete" from `_determineDataEntryCompletionStatus()`, but still has "Complete" in the database. In this case it is because one of the required fields, `examiner_location`, is null.

The script in this PR currently sets Required_elements_completed using `_determineRequiredElementsCompleted()` and `_setRequiredElementsCompleted()`, which has the effect of updating the flag instead of transferring it. Instead of this, do you think it's better if I just transfer it from the loadinstancedata values? Or is it good that this value is updated?

zaliqarosli:
@CamilleBeau i think it's good that the value is update if it has to be. but if there maybe a way to check inconsistencies and print them out to ask for confirmation? let me know if i can be clearer

driusan:
@zaliqarosli it's not quite clear from the above thread and it's showing up as approved.. is this ready to merge or are there still changes to do?

zaliqarosli (changes requested):
requesting changes:

> @CamilleBeau i think it's good that the value is updated if it has to be. but is there maybe a way to check inconsistencies and print them out to ask for confirmation before continuing? let me know if i can be clearer

tools/single_use/Set_Required_elements_completed_flag.php
```diff
@@ -0,0 +1,134 @@
+<?php
```
ridz1208, line 1: after SQL

PR #6879, opened 2020-08-03
[Test/Tools] Fix scripts issue with PR #6786

laemtl:
PR #6786 was merged with an issue which removes Database_test.php from the list of scripts PHPCS runs onto.
This PR fixes this issue.

### DIFF
--- a/test/run-php-linter.sh
+++ b/test/run-php-linter.sh
@@ -7,7 +7,7 @@ find docs modules htdocs php src tools \
     -name '*.class.inc' \
     -print0 -o -name '*.php' -print0 \
     |xargs -0 -n1 php -l \
-    >/dev/null  
+    >/dev/null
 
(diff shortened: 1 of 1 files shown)


test/run-php-linter.sh
```diff
@@ -49,8 +49,6 @@ declare -a tools_list=(
 
 # And on all PHP files in this array
 declare -a test_list=(
-    'integrationtests/LorisIntegrationTest.class.inc'
-    'integrationtests/LorisIntegrationTestWithCandidate.class.inc'
```
driusan, line 53: Are these supposed to be removed?

laemtl: Yes, "test/integrationtests/" folder is integrated L59.

laemtl: This was removed in PR #6800

PR #6882, opened 2020-08-05
[Statistics] fix behavioural tab without any project selection

ridz1208:
## Brief summary of changes
When no project is selected the behavioural tab should show results for all projects for all visits. instead it was just displaying an empty row see images.

before:
Screenshot from 2020-08-05 13-16-54

after:
Screenshot from 2020-08-05 13-14-21


ridz1208:
@racostas can test this ASAP please

racostas:
Hi @ridz1208 , changes are looking good to me. I think the intention of the first condition was to limit the visit labels shown only to those belonging to the project selected: I would said if we want to be strict, when "All Projects" are selected we should show the visit labels for only the "all projects the current user have access to" which seen not to be the case with this fix. Anyway I think this could be a future feature.

PR #6887, opened 2020-08-06
[meta] add compiled JS files to gitignore

ridz1208:
## Brief summary of changes
read the title

### DIFF
--- a/modules/instruments/.gitignore
+++ b/modules/instruments/.gitignore
@@ -1,0 +1,1 @@
+js/CandidateInstrumentList.js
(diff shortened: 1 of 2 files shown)


modules/instruments/.gitignore
```diff
@@ -0,0 +1 @@
+js/CandidateInstrumentList.js
```
driusan, line 1: I don't know about the imaging_browser one, but this one can just be `js/*`. It's a newish module and doesn't have any non-compiled js grandfathered in..

ridz1208: hahaha every PR we debate the same thing 😂 

I don't like jokers 🃏

PR #6898, opened 2020-08-12
[Raisinbread] Applied HRRT patch to RB

### LINKED ISSUE #6897: [Raisinbread] SQL patch for HRRT not applied to Raisinbread
The patch `2020-03-09-SQL_patch_for_HRRT_PET_images_insertion.sql` from PR #6142 was not applied to Raisinbread. It should be applied and added to the migration file. 

Related to PR #6142

AlexandraLivadas:
Applied the `2020-03-09-SQL_patch_for_HRRT_PET_images_insertion.sql` patch to Raisinbread and exported the changed RB files

#### Links to related issues

* Resolves #6897 
* Related to PR #6142


cmadjar:
@ridz1208 unfortunately, I don't think we have data that we can put in RB for those tables. The images were Samir's and I doubt he was scanned on the MNI HRRT scanner... Not sure either what other datasets could be used given ethics limitation...

ridz1208:
@AlexandraLivadas given the conflict in this case, it's easier to redo the work generally than to rebase since it's the entirety of the file that has a conflict.

What I usually do is the following
1. go to the `main` branch and source RB
2. checkout this PR
3. run the SQL patch on RB (which was sourced from main, not from this branch. so no conflicts)
4. run the exporter which will overwrite the RB_files file

PR #6899, opened 2020-08-12
[API] Added features for dicoms : PatientName and images: isPhantom

spell00:
## Brief summary of changes

`Patientname` is added to dicoms endpoint view and `IsPhantom` is added to images endpoint view

#### Testing instructions

1. Go to `<hostname>/api/v0.0.3/candidates/400162/V6/dicoms/`. Under the attribute `Tarname`, a new attribute should be there: `Patientname : "ROM162_400162_V6"`

2. Go to `<hostname>/api/v0.0.3/candidates/400162/V6/images/`. Each file should have 4 attributes, the last one being the new attribute, `IsPhantom: false`


xlecours:
@driusan do you mean that the spec should be adjusted in this PR? Or do you object to this change because it does not correspond to the API spec?

driusan:
0.0.3 was released, any changes need to go into a new version which incorporates the changes both in the code and the spec for that version

xlecours:
@driusan Additions do not compromise the specs for existing users. Managing a new version for every modification is a lot of overhead.

driusan:
New additions without a new version absolutely compromises the spec. It's an API intended. Values can't sometimes exist and sometimes not exist with the same version of the API and no way for the caller to know what it's going to get.

christinerogers:
@xlecours @driusan  Can we find a path forward here or is it better to mark this as Needs Discussion and we'll shelve it until the LORIS meeting? 
A new version of the spec could be done but up to you guys when/where we start it from.

spell00:
The feature added in this endpoint is not in the specs for v0.0.3, so the changes suggested in this PR need to be added to the API v0.0.4 initiated in PR #6944, which is in progress.

christinerogers:
@spell00 is this PR still valid and ready for re-review?  If so please re-request review from driusan (click the cycling arrows in the Reviewers list)

spell00:
@christinerogers It is still blocked by PR #6944

xlecours:
LGTM except it is failing travis.

spell00:
@driusan the modifications are now added to the specs of v0.0.4-dev

spell00:
@xlecours Your comment seem to indicate you approve the PR, but you did not approve it

spell00:
blocked by #7101

spell00:
@xlecours @driusan It is now passing Travis and ready for final review

modules/api/docs/LorisRESTAPI_v0.0.4-dev.md
```diff
@@ -553,6 +553,7 @@ the form:
         "OutputType" : "native",
         "Filename" : "abc.mnc",
         "AcquisitionType" : "t1w/t2w/etc",
+        "IsPhantom" : true/false,
```
driusan, line 556: I was going to say this isn't valid js so it's going to be rendered as an error since it's in a JS block, but then I decided to check if it's valid js in my JS console..

```
>> $x = true / false
Infinity
```
....

modules/api/php/views/visit/dicoms.class.inc
```diff
@@ -52,8 +52,9 @@ class Dicoms
     private static function _formatDicomTars(DicomTarDTO $dicom): array
     {
         return [
-            'Tarname'    => $dicom->getTarname(),
-            'SeriesInfo' => array_map(
+            'Tarname'     => $dicom->getTarname(),
+            'Patientname' => $dicom->getPatientname(),
```
driusan, line 56: This is changing the result for both 0.0.3 and 0.0.4-dev, it needs a version check.

modules/api/php/views/visit/images.class.inc
```diff
@@ -58,6 +58,7 @@ class Images
                     'OutputType'      => $image->getOutputType(),
                     'Filename'        => $image->getFilename(),
                     'AcquisitionType' => $image->getAcquisitionprotocol(),
+                    'IsPhantom'       => $image->isPhantom(),
```
driusan, line 61: Same

driusan (changes requested):

spell00:
@xlecours @driusan I did the corrections, this is ready for re-review

driusan (approved):
Looks fine to me, @xlecours does it still look good to you after the latest changes?

xlecours (approved):

xlecours:
> Looks fine to me, @xlecours does it still look good to you after the latest changes?

yes.
I guess this will be the way to go for new endpoint response format. 
1. switch on version
2. use a class which name is suffixed by the -dev version. 

Once -dev is removed, replace the file by the -dev content then delete -dev.

driusan:
I don't think we're committed to that just because it's done that way in 1 place. We can always update the coding style to something more robust later, but since it works for this and the author of the PR is leaving I think we should merge it.

PR #6900, opened 2020-08-12
[Raisinbread] Added CNV/CPG records for candidates with SNP data

### LINKED ISSUE #6302: [Genomic Browser] Issues with the Profiles Tab Datatable.

### LINKED ISSUE #6881: [RB/GenomicBrowser] : Add basic CNP/CPG records
This issue was reported as **c* in #6302 by henri during 23 testing. 
to do: 
* Add 3 records in raisin bread for each type of data for candidates that *already have* SNP data. 
i.e. write a simple mysql patch and I'll help refine it. Might involve multiple tables. 

The goal to just have a few records so that they display as having data.

AlexandraLivadas:
Added CNV and CPG records for candidates that already have SNP data. This is necessary so that the `CNV` and `Methylation` tabs in the Genomic Browser module can be tested.

**Note:** I can add more data if more is needed or requested.

#### Testing instructions
With a fresh install of RB or after running the patch:

1. Navigate to `Genomic Browser`.
2. See that the `CNV` and `Methylation` tabs now display and can filter data.

#### Links to related issues

* Resolves #6881 
* Partially resolves #6302


SQL/New_patches/2020-08-12_add_CNV_CPG_data_for_genomic_browser.sql
```diff
@@ -0,0 +1,25 @@
+--Add CNV data for some candidates
```
ridz1208, line 1: files in this directory get run by real projects with real data. this patch is adding mock data useful for testing.

This addition while very very helpful should not be in the patch directory. it's sufficient to make the changes to the RB files only.

if it makes it easier to understand:
 - anything that should be ran on all projects should also be ran on RB
 - however patches created FOR RB should never be ran on projects

AlexandraLivadas: Okay this makes sense. I'll remove the patch and leave the changes in the Raisinbread files.

christinerogers: Alex can you also add this note to the Cheat sheet?  this isn't isn't available to the team unless rida tells them otherwise.

christinerogers:
alex could you post a screenshot of what it looks like now?

AlexandraLivadas:
@christine The candidate's I used do have SNP data but I will check if they also have imaging data. I will add that content to the internal doc :)

driusan:
I can't tell from the thread.. is this ready to be merged or does it need a doc updated?

AlexandraLivadas:
@driusan It does not need a doc update. After Christine's comment, I checked that the candidate's have SNP data and they do, so I think this is all set.

PR #6671, opened 2020-06-04
[API] New branch for adding the testsuite for the API + addressed comments

spell00:
Creation of a test suite for the API.

PR #6516, opened 2020-05-11
[API] Adjust return values of API

HenriRabalais:
This PR replace #6401.

## Brief summary of changes
This PR removes `File` from being returned for project images. It also ensure that Project data is returned when querying visit information.

#### Testing instructions (if applicable)
1. Run `GET /projects/$ProjectName/images/` and ensure `File` key is not returned
2. Run `GET /candidates/$CandID/$VisitLabel` and ensure `Project` key returns project data.

#### Link(s) to related issue(s)
* Resolves #6321 (points 1 and 2)

PR #6905, opened 2020-08-12
[API] correct file name in /projects/{project}/recordings and /projects/{project}/images endpoints

### LINKED ISSUE #6904: [API] Incorrect or absent File name in endpoints `/projects/{project}/images` and `/projects/{project}/recordings`
**Describe the bug**
The File's name is incorrect in endpoint `/projects/{project}/images` and absent in `/projects/{project}/recordings`.

**To Reproduce**
1. Go to `<hostname>/api/v0.0.3/projects/pumpernickel/recordings`. 
recordings_correct_filename

`File` is actually the path. The first item of the list should have be the file's name.

2. Go to `<hostname>/api/v0.0.3/projects/pumpernickel/images`
images

There is no `File` attribute and there should be one, like for the recordings.

spell00:
## Brief summary of changes

1. The variable `_filename` is changed to `_filepath`
2. The attribute named `File` in the classes `ProjectImagesRowProvisioner` and `ProjectRecordingsRowProvisioner` is changed to FilePath
3. The attribute File is displayed for all images in the endpoint `/projects/{project}/images`
4. The attribute File is the actual file and not the path in the endpoint `/projects/{project}/recordings`

1. Go to `<hostname>/api/v0.0.3/projects/pumpernickel/recordings`. 
The attribute `File` of the first item of the list should read `sub-OTT167_ses-V1_task-faceO_eeg.edf`, like:

recordings_project_correct_filename

2. The attribute `File` should be present and read `demo_587630_V1_t1_001.mnc`, like:

image_correct_file

* Resolves #6904

### DIFF
--- a/modules/api/php/models/projectimagesrow.class.inc
+++ b/modules/api/php/models/projectimagesrow.class.inc
@@ -29,7 +29,7 @@ class ProjectImagesRow implements \LORIS\Data\DataInstance
     private $_visitdate;
     private $_centerid;
     private $_centername;
-    private $_path;
+    private $_filepath;
     private $_inserttime;
     private $_scantype;
     private $_qcstatus;
@@ -49,7 +49,7 @@ class ProjectImagesRow implements \LORIS\Data\DataInstance
         $this->_visitdate  = $row['Visit_date'] ?? null;
         $this->_centerid   = $row['CenterID'] ?? null;
         $this->_centername = $row['Site'] ?? null;
-        $this->_path       = $row['Path'] ?? null;
+        $this->_filepath   = $row['FilePath'] ?? null;
         $this->_inserttime = $row['InsertTime'] ?? null;
         $this->_scantype   = $row['ScanType'] ?? null;
         $this->_qcstatus   = $row['QC_status'] ?? null;
@@ -74,7 +74,7 @@ class ProjectImagesRow implements \LORIS\Data\DataInstance
             'QC_status'  => $this->_qcstatus,
             'Selected'   => $this->_selected,
         ];
-        $filename = basename($this->_path);
+        $filename = basename($this->_filepath);
         $candid   = $obj['Candidate'];
         $visit    = $obj['Visit'];
 
--- a/modules/api/php/models/projectimagesrow.class.inc
+++ b/modules/api/php/models/projectimagesrow.class.inc
@@ -29,7 +29,7 @@ class ProjectImagesRow implements \LORIS\Data\DataInstance
     private $_visitdate;
     private $_centerid;
     private $_centername;
-    private $_filename;
+    private $_path;
     private $_inserttime;
     private $_scantype;
     private $_qcstatus;
@@ -75,7 +75,7 @@ class ProjectImagesRow implements \LORIS\Data\DataInstance
             'Selected'   => $this->_selected,
             'Path'       => $this->_path,
         ];
-        $filename = basename($this->_filename);
+        $filename = basename($this->_path);
         $candid   = $obj['Candidate'];
         $visit    = $obj['Visit'];
 
--- a/modules/api/php/models/projectimagesrow.class.inc
+++ b/modules/api/php/models/projectimagesrow.class.inc
@@ -73,7 +73,6 @@ class ProjectImagesRow implements \LORIS\Data\DataInstance
             'ScanType'   => $this->_scantype,
             'QC_status'  => $this->_qcstatus,
             'Selected'   => $this->_selected,
-            'Path'       => $this->_path,
         ];
         $filename = basename($this->_path);
         $candid   = $obj['Candidate'];
--- a/modules/api/php/models/projectimagesrow.class.inc
+++ b/modules/api/php/models/projectimagesrow.class.inc
@@ -49,7 +49,7 @@ class ProjectImagesRow implements \LORIS\Data\DataInstance
         $this->_visitdate  = $row['Visit_date'] ?? null;
         $this->_centerid   = $row['CenterID'] ?? null;
         $this->_centername = $row['Site'] ?? null;
-        $this->_filename   = $row['File'] ?? null;
+        $this->_path       = $row['Path'] ?? null;
         $this->_inserttime = $row['InsertTime'] ?? null;
         $this->_scantype   = $row['ScanType'] ?? null;
         $this->_qcstatus   = $row['QC_status'] ?? null;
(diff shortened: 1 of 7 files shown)


spell00:
@christinerogers I did consider it, but I wasn't sure it would be useful. If it is, I can add it no problem

xlecours:
@spell00 Other than the changes in the code, this is only adding a `File` property to `/projects/{project}/images` right?

spell00:
@xlecours that's right

xlecours:
Do you mind if we wait for https://github.com/aces/Loris/pull/6671 to get merged so you could add test(s) for this endpoint?

spell00:
it would be better to wait for #6671 to be merged, adding new tests for this endpoint should be trivial after that

xlecours:
The API spec needs to be adjusted. Or do you want a new version @driusan ?

modules/api/php/models/projectimagesrow.class.inc
```diff
@@ -74,12 +74,13 @@ class ProjectImagesRow implements \LORIS\Data\DataInstance
...
             'Selected'   => $this->_selected,
         ];
-        $filename = basename($this->_filename);
+        $filename = basename($this->_filepath);
         $candid   = $obj['Candidate'];
         $visit    = $obj['Visit'];
 
         $obj['Link']       = "/candidates/$candid/$visit/images/$filename";
         $obj['InsertTime'] = date('c', (int) $obj['InsertTime']);
+        $obj['File']       = $filename;
```
driusan, line 83: This was specifically removed in #6516 and should not be added without a new version of the spec. (In fact, it shouldn't be added at all.. it's just leaking internal details and the file is available to API users through the link field)

driusan:
Any changes to the API need a new API version, but I'm not convinced that this change is a good idea even with a new API version.

modules/api/php/models/projectimagesrow.class.inc
spell00: In this case, adding the filename doesn't leak any additional details, the filename in https://demo.loris.ca/api/v0.0.3/projects/pumpernickel/images  is already available in the `Link` , so it is still leaking the file's name. I thought it might be redundant, but I tried to make it consistent with https://demo.loris.ca/api/v0.0.3/projects/pumpernickel/recordings , where the field `File` was actually the path.

spell00:
@driusan  Maybe this change is not a good idea, but some change should be done for 1) consistency and 2) The file's name in `/api/v0.0.3/projects/{project}/images`  is already available in the current version, even after #6516 in the `Link` field, so that should be removed too?

modules/api/php/models/projectimagesrow.class.inc
driusan: I don't mean it's leaking sensitive information, I mean it's leaking implementation details. The path on the filesystem is irrelevant for users of the REST API and unusable without access to the backend (in which case there's no reason to be using the REST API in the first place.) There is already a mechanism in the API to get the data for the file through the API -- the `Link` attribute, which doesn't require a leaky abstraction.

driusan:
No, the `Link` is relative to the API, whether or not you can parse the filesystem path out if it if you really wanted to doesn't hurt the data. If you mean the `Link` contains just the path and doesn't contain the correct link, then that should be fixed so that it agrees with the spec.

modules/api/php/models/projectimagesrow.class.inc
spell00: I know that the path is irrelevant to the user, that is actually what I am trying to change in  `modules/api/php/models/projectrecordingsrow.class.inc` . In this case, `File` doesn't mean the file's path, but the file's name. I can change it for `Filename` to be more precise. In the code, `File` keeps is referred to as the file's name (e.g. `$this->_filename   = $row['File'] ?? null;`), but it's actually the file path, as you can see in https://demo.loris.ca/api/v0.0.3/projects/pumpernickel/recordings (which is the same thing on my VM on the `aces/main` branch).

driusan: I know very little about the EEG part of the API, Cecile would need to provide feedback on that, but if you're changing it to something other than what the spec currently says you'd need to create a new version/update the spec for that new version.

From what I can see on the demo link that you provided, it looks like the "sub-OTT167/ses-V1/eeg/" part may be relevant for the BIDS standard and isn't in the Link. It might make sense to remove the `File` and add something like `BIDSName` with a value like `sub-OTT167/ses-V1/eeg/sub-OTT167_ses-V1_task-faceO_eeg.edf`, but @cmadjar would need to need to weigh in on that.

spell00:
In the files, I think `_filename` should be changed for `_filepath`. I don't think `filename` should mean the file's path, as it does before the changes requested. `/api/v0.0.3/projects/{project}/recordings` had the attribute `File`, which is the file's path. This is actually consistent with the spec, but why the spec for [/api/v0.0.3/projects/{project}/images](https://github.com/aces/Loris/blob/main/modules/api/docs/LorisRESTAPI_v0.0.3.md#211-single-project-images) doesn't have a `File` attribute, but [/api/v0.0.3/projects/{project}/recordings](https://github.com/aces/Loris/blob/main/modules/api/docs/LorisRESTAPI_v0.0.3.md#215-single-project-electrophysiology-recordings) does ?

spell00:
Ok I get it, so I will remove anything that adds anything that is not in the current specs for this PR. I will change the Issue accordingly. I did find an error in the specs for the electrophysiology endpoint though, in the [Single project electrophysiology recordings](https://github.com/aces/Loris/blob/main/modules/api/docs/LorisRESTAPI_v0.0.3.md#215-single-project-electrophysiology-recordings),  `Link : "/candidates/300167/V1/electrophysiology/sub-OTT167_ses-V1_task-faceO_eeg.edf"` should be `Link : "/candidates/300167/V1/recordings/sub-OTT167_ses-V1_task-faceO_eeg.edf"`. In that case, should I correct the spec without making a new version? If someone tries to follow the links given by that endpoint in the current version of the API, it's not going to work.

driusan:
yes, if the link is incorrect and points nowhere that's a bug that can be fixed.. but it'll have to wait for @cmadjar to come back from vacation to review it

spell00:
Last thing: in the code, wouldn't it be better if `_filename` was changed to `_filepath`, as it contains the file's path, and the variable `$filename` contains the file's name? I could close this PR and make 2 very small PRs for each Issue: 1 for the link pointing nowhere that can be solved when Cecile returns from vacation, and 1 for changing the variables names ?

christinerogers:
@spell00 if this should actually be 2 issues -- yes, please do 2 issues and 2 pr.  1 can wait for cecile, 1  can move forward next week given it aligns with dave's feedback above. 

@xlecours any thoughts?

Issue #6906, opened 2020-08-12
[Help Doc] No help document for the instruments module


AlexandraLivadas:
There is no help document for the instruments module. 

When you navigate to a data entry page for a specific timepoint's instrument and click on the `?` button, there is a `No help available for instruments` message:
<img width="1434" alt="Screen Shot 2020-08-12 at 6 00 57 PM" src="https://user-images.githubusercontent.com/35467361/90072843-1dcf9480-dcc6-11ea-8c14-955227318d39.png">

driusan:
The instruments module is just a module that delegates to the instrument, the help on it should be using the help from the help module. Is it working correctly when there's help content added in the help editor?

AlexandraLivadas:
@driusan My other PR #6907 adds data to the help editor for the BMI instrument. However, when I go to data entry for a BMI instrument even with data in the help editor, the same "No help available for instruments" message appears here.

PR #6195, opened 2020-03-30
sweetalert update

haoweiqiu:
## Brief summary of changes
This updates sweetalert to sweetalert2 in most javascript files. However, the only file that does not change is NDB_Page.class.inc, since there is some error in Travis. This Pull Request contains commits of my another Pull Request (#6192).

Haowei Qiu, GSOC Applicant
CC @christinerogers 

#### Link(s) to related issue(s)

* Partially Resolves #6114   (Reference the issue this fixes, if any.)

PR #6907, opened 2020-08-12
[Raisinbread] Fixing help_editor module to display RB instrument instructions

### LINKED ISSUE #5691: [Raisinbread] No data for help editor module?
It doesn't seem to be possible to edit any help files using the Raisinbread dataset. The Test VM, demo, and my sandbox contain no data.

How is this module being tested?

AlexandraLivadas:
**Edited** 
This PR aimed to add data to the `help` table in Raisinbread so that the data would generate in the Help Editor module. This will allow the `Help Editor` module to be manually tested.

It was found that there is a bug in this module that causes it not to save/display this information. This issue is resolved in this PR. 

Here is a screenshot of the `Help Editor` module with the new data:
Screen Shot 2020-08-12 at 6 11 24 PM

With a fresh RB install:
1. Navigate to `Help Editor` in the frontend and see that data loads for instructions for the BMI instrument.

#### Link to related issue

* Resolves #5691

### DIFF
--- a/modules/help_editor/ajax/help.php
+++ b/modules/help_editor/ajax/help.php
@@ -32,20 +32,28 @@ try {
     include_once "helpfile.class.inc";
 
     if (!empty($moduleName)) {
-        $helpID = \LORIS\help_editor\HelpFile::hashToID(
-            md5($subpageName ?? $moduleName)
-        );
-    }
+        try {
+            $helpID = \LORIS\help_editor\HelpFile::hashToID(
+                md5($subpageName ?? $moduleName)
+            );
+            $help_file       = \LORIS\help_editor\HelpFile::factory($helpID);
+            $data            = $help_file->toArray();
+            $data['content'] = trim($data['content']);
 
-    $help_file       = \LORIS\help_editor\HelpFile::factory($helpID);
-    $data            = $help_file->toArray();
-    $data['content'] = trim($data['content']);
-
-    if (empty($data['updated'])) {
-        $data['updated'] = "-";
-        // if document was never updated should display date created
-        if (!empty($data['created'])) {
-            $data['updated'] = $data['created'];
+            if (empty($data['updated'])) {
+                $data['updated'] = "-";
+                // if document was never updated should display date created
+                if (!empty($data['created'])) {
+                    $data['updated'] = $data['created'];
+                }
+            }
+        } catch (\NotFound $e) {
+            // Send data with empty strings so that the content can be edited
+            $data = [
+                'content' => '',
+                'topic'   => '',
+                'updated' => '-'
+            ];
         }
     }
     print json_encode($data);
--- a/modules/help_editor/ajax/help.php
+++ b/modules/help_editor/ajax/help.php
@@ -33,11 +33,11 @@ try {
 
     if (!empty($moduleName)) {
         try {
-            $helpID          = \LORIS\help_editor\HelpFile::hashToID(
+            $helpID    = \LORIS\help_editor\HelpFile::hashToID(
                 md5($subpageName ?? $moduleName)
             );
-            $help_file       = \LORIS\help_editor\HelpFile::factory($helpID);
-            $data            = $help_file->toArray();
+            $help_file = \LORIS\help_editor\HelpFile::factory($helpID);
+            $data      = $help_file->toArray();
 
         } catch (\NotFound $e) {
             // Send data with empty strings so that the content can be edited
--- a/modules/help_editor/ajax/help.php
+++ b/modules/help_editor/ajax/help.php
@@ -38,15 +38,7 @@ try {
             );
             $help_file       = \LORIS\help_editor\HelpFile::factory($helpID);
             $data            = $help_file->toArray();
-            $data['content'] = trim($data['content']);
 
-            if (empty($data['updated'])) {
-                $data['updated'] = "-";
-                // if document was never updated should display date created
-                if (!empty($data['created'])) {
-                    $data['updated'] = $data['created'];
-                }
-            }
         } catch (\NotFound $e) {
             // Send data with empty strings so that the content can be edited
             $data = [
@@ -54,6 +46,16 @@ try {
                 'topic'   => '',
                 'updated' => '-'
             ];
+        }
+        
+        $data['content'] = trim($data['content']);
+
+        if (empty($data['updated'])) {
+            $data['updated'] = "-";
+            // if document was never updated should display date created
+            if (!empty($data['created'])) {
+                $data['updated'] = $data['created'];
+            }
         }
     }
     print json_encode($data);
--- a/modules/help_editor/ajax/help.php
+++ b/modules/help_editor/ajax/help.php
@@ -43,7 +43,7 @@ try {
             $data = [
                 'content' => '',
                 'topic'   => '',
-                'updated' => '-'
+                'updated' => ''
             ];
         }
 
--- a/modules/help_editor/ajax/help.php
+++ b/modules/help_editor/ajax/help.php
@@ -33,7 +33,7 @@ try {
 
     if (!empty($moduleName)) {
         try {
-            $helpID = \LORIS\help_editor\HelpFile::hashToID(
+            $helpID          = \LORIS\help_editor\HelpFile::hashToID(
                 md5($subpageName ?? $moduleName)
             );
             $help_file       = \LORIS\help_editor\HelpFile::factory($helpID);
--- a/modules/help_editor/ajax/help.php
+++ b/modules/help_editor/ajax/help.php
@@ -47,14 +47,15 @@ try {
                 'updated' => '-'
             ];
         }
-        
+
         $data['content'] = trim($data['content']);
 
         if (empty($data['updated'])) {
-            $data['updated'] = "-";
             // if document was never updated should display date created
-            if (!empty($data['created'])) {
+            if (!empty($data['created']) && isset($data['created'])) {
                 $data['updated'] = $data['created'];
+            } else {
+                $data['updated'] = "-";
             }
         }
     }
--- a/modules/help_editor/help/help_editor.md
+++ b/modules/help_editor/help/help_editor.md
@@ -1,6 +1,6 @@
 # Help Editor
 
-This module displays existing help content for LORIS instrument pages
+This module displays existing help content for LORIS instrument pages.
 
 Use the *Selection Filter* section to search by Topic or Content keywords.
 
--- a/modules/help_editor/help/help_editor.md
+++ b/modules/help_editor/help/help_editor.md
@@ -1,7 +1,10 @@
 # Help Editor
 
-This module displays existing help content for LORIS modules and pages. 
+This module displays existing help content for LORIS instrument pages
 
 Use the *Selection Filter* section to search by Topic or Content keywords.
 
-Click on the blue link in any *Topic* or *Parent Topic* column to edit the content within. Make your edits within the Edit Help Content page, and click **Save**. 
+Click on the blue link in any *Topic* or *Parent Topic* column to edit the content within. Make your edits within the Edit Help Content page, and click **Save**.
+
+To add new help content for an instrument, navigate to the instrument's page from any candidate and any session. Click on the `?` icon in the navigation bar at the top
+ of the page. In the help pop-up box, click the `Edit` button. This will take you to the *Edit Help Content* page where you can save help content for that instrument.
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -277,6 +277,8 @@ abstract class NDB_BVL_Instrument extends NDB_Page
 
         // Sets up page variables such as $this->commentID and $this->form
         $obj->setup($commentID, $page);
+        // Set page name to testName
+        $obj->name = $instrument;
 
         if (!empty($commentID)) {
             $obj->setupCandidateInfoTables();
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -275,10 +275,10 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             $obj->displayAllFields = true;
         }
 
-        // Sets up page variables such as $this->commentID and $this->form
-        $obj->setup($commentID, $page);
         // Set page name to testName
         $obj->name = $instrument;
+        // Sets up page variables such as $this->commentID and $this->form
+        $obj->setup($commentID, $page);
 
         if (!empty($commentID)) {
             $obj->setupCandidateInfoTables();
--- a/raisinbread/RB_files/RB_help.sql
+++ b/raisinbread/RB_files/RB_help.sql
@@ -1,6 +1,6 @@
 SET FOREIGN_KEY_CHECKS=0;
 TRUNCATE TABLE `help`;
 LOCK TABLES `help` WRITE;
-INSERT INTO `help` (`helpID`, `parentID`, `hash`, `topic`, `content`, `created`, `updated`) VALUES (123,12,'810dc6911c825b55eff684098f2beb19','bmi','The BMI calculator instrument can calculate the BMI and BMI Classification of a visitor. The two required fields for this instrument are the Date of Administration of the instrument and the Examiner name. The height and weight of the visitor must then be entered in either standard or metric units. The instrument will not accept an entry where the height and weight are given in both units. After this information is given, the instrument will calculate the BMI and BMI Classification of the visitor.',NULL,NULL);
+INSERT INTO `help` (`helpID`, `parentID`, `hash`, `topic`, `content`, `created`, `updated`) VALUES (123,-1,'810dc6911c825b55eff684098f2beb19','bmi','The BMI calculator instrument can calculate the BMI and BMI Classification of a visitor. The two required fields for this instrument are the Date of Administration of the instrument and the Examiner name. The height and weight of the visitor must then be entered in either standard or metric units. The instrument will not accept an entry where the height and weight are given in both units. After this information is given, the instrument will calculate the BMI and BMI Classification of the visitor.','2020-09-01 01:03:09',NULL);
 UNLOCK TABLES;
 SET FOREIGN_KEY_CHECKS=1;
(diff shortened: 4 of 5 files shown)


christinerogers:
great, can you also comment in one line on whether the help editor seems to work in terms of editing help text?

AlexandraLivadas:
After testing with the data added, the Help Editor does not correctly save changes made in the frontend to the file contents. I made changes to the contents and hit `Save`. The changes displayed correctly but once I hit `Return to Help Editor` or navigated back to the main page myself, the changes were actually not saved.

Another issue is that the button `Return to Help Editor` does not actually return me to the main Help Editor page. Instead, I stayed on the editing content page and the changes I made were not saved.

raisinbread/RB_files/RB_help.sql
```diff
@@ -1,5 +1,6 @@
 SET FOREIGN_KEY_CHECKS=0;
 TRUNCATE TABLE `help`;
 LOCK TABLES `help` WRITE;
+INSERT INTO `help` (`helpID`, `parentID`, `hash`, `topic`, `content`, `created`, `updated`) VALUES (123,12,'hash','BMI Instructions','The BMI calculator instrument can calculate the BMI and BMI Classification of a visitor. The two required fields for this instrument are the Date of Administration of the instrument and the Examiner name. The height and weight of the visitor must then be entered in either standard or metric units. The instrument will not accept an entry where the height and weight are given in both units. After this information is given, the instrument will calculate the BMI and BMI Classification of the visitor.',NULL,NULL);
```
christinerogers, line 4: Alex, since this doesn't actually render for the BMI -- [ per your comment here ](https://github.com/aces/Loris/issues/6906#issuecomment-674946742)
I think it might not be because of the value in the `topic` column.  if you change it to `bmi` which i think would be the relevant test_name value for this instrument -- what happens? 

Could you also take another step here towards resolving it by checking the code and to see how the Help display/dropdown actually trying to locate relevant help text?  If it's not trying to match the `topic` column against a valid `test_name` value -- then what is it trying to look for?

AlexandraLivadas: Yes I can definitely look into this :)

AlexandraLivadas: @christinerogers @driusan  So after some digging:
The help file is handled by [helpHandler.js](https://github.com/aces/Loris/blob/main/htdocs/js/helpHandler.js). This then calls the [modules/help_editor/ajax/help.php](https://github.com/aces/Loris/blob/main/modules/help_editor/ajax/help.php) file with a GET request.
 
In THAT file, it first tries to render the help file by either using the module name (which is given by the request parameter `testName` or using the `subtest` parameter (in this case "bmi" or "aosi"). If it's not a module, it instead tries to load data from the `help` table using the request parameter `subtest` as the identifier. But when I went to an instrument data entry page and used the Developer Tool, I saw that when the GET request is sent to `help.php`, it only provides the 'testName` parameter in the request and doesn't provide a `subtest` value, so it never checks if there's a file in the help table and instead just looks for the markdown help file for the Instruments module. 

Let me know if this makes sense, I can explain it in more detail if necessary. I don't know how to fix this as of now. I can keep looking to see how to edit the GET request to provide the `subtest` parameter, but I haven't figured out where to do that yet.

ridz1208:
@zaliqarosli can you team up with @AlexandraLivadas to figure out what is happening in this module ?

our options would be to either fix it if it doesnt work as intended or remove the module entirely

zaliqarosli:
@ridz1208 yeah sure @AlexandraLivadas wanna slack me or get on a call to explain what's going on?

ridz1208:
@zaliqarosli @AlexandraLivadas have you guys firgured it out ? is it working for  bmi specifically now ?

ridz1208:
fixes #5691

zaliqarosli:
@ridz1208 yeah i fixed it - sent a PR to this PR branch. @AlexandraLivadas can you merge the changes?

modules/help_editor/js/help_editor_helper.js
```diff
@@ -1,5 +1,3 @@
-import swal from 'sweetalert2';
```
driusan, line 1: why is this downgrading from swal2 to swal1?

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -253,6 +253,8 @@ abstract class NDB_BVL_Instrument extends NDB_Page
 
         // Sets up page variables such as $this->commentID and $this->form
         $obj->setup($commentID, $page);
+        // Set page name to testName
+        $obj->name = $instrument;
```
driusan, line 257: I think this should go before the setup call in case the setup function intentionally set it to something.

modules/help_editor/js/help_editor_helper.js
AlexandraLivadas: When @zaliqarosli was making the changes, she said that this statement threw an error. However, when I added this line back on my local version, I did not get the same error. I'm not sure why this was causing issues for her but not me.

php/libraries/NDB_BVL_Instrument.class.inc
AlexandraLivadas: The setup function being used here only sets the `commentID` and `page` values, so it should have no effect on the `name` value. I can make this change (and test it) anyway though 👍

modules/help_editor/js/help_editor_helper.js
driusan: I don't think it should be downgraded.

zaliqarosli: undoing the changes on this file is still giving me the error:

Screen Shot 2020-09-21 at 1 16 02 PM

@laemtl also reported seeing this error in https://github.com/aces/Loris/issues/6940. (Laetitia, any chance you could give this PR a test, with and without these changes to this file?)

@driusan @HenriRabalais @maltheism any of you know how to properly fix this without downgrading to swal1?

laemtl: Yes sure !

maltheism: The reason swal2 isn't accessible is because the help_editor_helper.js is inside the `js` directory and not the `jsx` directory that webpack handles. I think she could move the file to the jsx directory while maybe needing to add it to webpack where the other file the help_editor module uses.

laemtl: @zaliqarosli You are right, there is a bug introduced by #6195. The migration to swal2 introduced problems.

maltheism: @laemtl see htdocs/vendor/sweetalert for the old sweetalert.js that gets accessed anywhere in Loris. I think if we placed the `node_modules/sweetalert/dist/sweetalert2.js` inside that directory or a new one. Then swal2 would be accessible without needing to import.

laemtl: I would like to fix this in a compatible way with my PR #7018. I did fix this for another script by reactifying it.

zaliqarosli: @driusan how would you prefer us to proceed?

zaliqarosli: i think this PR should just fix the bug introduced in https://github.com/aces/Loris/pull/6195/files#diff-319f29c33167f3c507dcb5516ec2578c and the move to sweetalert2 should be tackled in another PR

driusan: I'm not sure I understand why the file can't just be moved from the `js` directory to `jsx` and added to the webpack config so that webpack compiles it and it can use swal2?

laemtl: @driusan Eslint fails with /var/www/html/loris/modules/help_editor/jsx/help_editor_helper.js
33:13  error  A function with a name starting with an uppercase letter should only be used as a constructor  new-cap


```
markdownContent = document.createElement('div');
ReactDOM.render(
       RMarkdown({content: content}),
       markdownContent
);
```

zaliqarosli: wouldn't it be kind of against our repo standard that we're calling a non-react non-jsx file a jsx file and putting it under that directory? help_editor_helper.js isn't a component and doesn't need to be compiled. there's a way to call sweetalert2 with normal JS:

Screen Shot 2020-09-21 at 5 37 34 PM

but using require() also throws an error. surely just making the file a jsx file isn't the right move?

maltheism: @zaliqarosli nodejs, browserify, or webpack supports require() but not javascript on the browser. The browser shouldn't have access to node_modules on the server anyways.

laemtl: @zaliqarosli You are right, I migrated my fix to #7018.

christinerogers (dismissed):

christinerogers:
per Loris meeting discussion today :
* @laemtl will merge the solution from #7018 here 
* unassigning @AlexandraLivadas from this PR (thanks and great work!)
* @laemtl please request re-review from the usual suspects when ready

laemtl:
Updated. As a stated above,  I'm not sure the way I tackle it is the best. I kind of agree that it is weird to have this file under jsx/. I wonder if reactifying it should be preferable? Isn't the plan to completely remove the jQuery dependencies in the future?

zaliqarosli:
this works beautifully now! i agree with laetitia's comments above so i will leave the approving review for someone else. this PR passes testing though

modules/help_editor/jsx/help_editor_helper.js
```diff
@@ -0,0 +1,84 @@
+/* eslint new-cap: ["error", {capIsNewExceptions: ["RMarkdown"]}]*/
```
maltheism: @laemtl shouldn't have this line

modules/help_editor/ajax/help.php
```diff
@@ -32,20 +32,28 @@
...
-        if (!empty($data['created'])) {
-            $data['updated'] = $data['created'];
+            if (empty($data['updated'])) {
+                $data['updated'] = "-";
+                // if document was never updated should display date created
+                if (!empty($data['created'])) {
+                    $data['updated'] = $data['created'];
+                }
+            }
+        } catch (\NotFound $e) {
```
maltheism, line 50: I'm unsure having a giant try catch here is best compared to encompassing just the lines that throw the error and then handling it from there.

maltheism:
@laemtl I pushed a commit that removes the eslint ignore comment and added two isset() checks to not have the error_logs be populated. I hope you don't mind.

Also I made a comment about using try cache. I think it's fine what you did in this case but it's better to encompass only the lines that throw the error and for making it easier for future developers know where the try cache is actually able to fail. The code can keeps getting modified over time by more being added inside the try {} to make it confusing where it fails.

maltheism (commented):
The solution works nicely. I think eventually this ugly jquery code should be replaced by a component that swaps the help editor preview text with whatever is supplied. Since that's actually what's being done with it. None of the jquery is doing anything really fancy besides that.

jsx/Markdown.js
```diff
@@ -40,6 +40,6 @@ Markdown.propTypes = {
 let RMarkdown = React.createFactory(Markdown);
 
 window.Markdown = Markdown;
-window.RMarkdown = RMarkdown;
+window.rMarkdown = RMarkdown;
```
zaliqarosli, line 43: @maltheism i'm not sure if this is the PR to incorporate this change because it affects all of Loris, or atleast wherever Markdown is being used. i think changing the Markdown is out of scope. i would agree with laetitia's solution for right now to add the eslint exception comments

modules/help_editor/ajax/help.php
zaliqarosli: ill try make this change

jsx/Markdown.js
zaliqarosli: fair enough! if its limited to this module, i guess that's okay

maltheism: @zaliqarosli that change `window.rMarkdown = RMarkdown;` is just setting the global `window` variable to have "rMarkdown" (Ex. window.rMarkdown). That line removed could still exist in this PR if people are worried about it. Since we would then have window.RMarkdown and window.rMarkdown to use whenever. Although I'm somewhat confident my global search through the code didn't miss anything.

maltheism (approved):

modules/help_editor/jsx/help_editor_helper.js
laemtl: I agree. I was just not sure if renaming RMarkdown was desired as it's a convention accross a lot of modules.

jsx/Markdown.js
laemtl: Just to add my grain of salt here...  As this file will/should be rewritten soon, I tend to think that it's preferable to add the eslint ignore line and not modifying the core. Once this file will be reactified, the core change will not be necessary anymore, and this change break the naming policy consistency. I agree that it's preferable to avoid using eslint ignore for long term change though :)

maltheism: @laemtl looking at this code in question. I doubt that's anytime soon.

raisinbread/RB_files/RB_help.sql
```diff
@@ -1,5 +1,6 @@
 SET FOREIGN_KEY_CHECKS=0;
 TRUNCATE TABLE `help`;
 LOCK TABLES `help` WRITE;
+INSERT INTO `help` (`helpID`, `parentID`, `hash`, `topic`, `content`, `created`, `updated`) VALUES (123,-1,'810dc6911c825b55eff684098f2beb19','bmi','The BMI calculator instrument can calculate the BMI and BMI Classification of a visitor. The two required fields for this instrument are the Date of Administration of the instrument and the Examiner name. The height and weight of the visitor must then be entered in either standard or metric units. The instrument will not accept an entry where the height and weight are given in both units. After this information is given, the instrument will calculate the BMI and BMI Classification of the visitor.','2020-09-01 01:03:09',NULL);
```
christinerogers, line 4:
```suggestion
INSERT INTO `help` (`helpID`, `parentID`, `hash`, `topic`, `content`, `created`, `updated`) VALUES (123,-1,'810dc6911c825b55eff684098f2beb19','bmi','The BMI calculator instrument calculates the BMI and BMI Classification of an individual person. The two required fields for this instrument are the Date of Administration of the instrument and the Examiner name. Enter height and weight in either standard or metric units. After data is entered and the form is saved, the calculated BMI value and BMI Classification of the individual will be displayed.','2020-09-01 01:03:09',NULL);
```

christinerogers: just realized this text has been referring to the candidate as "visitor" -- see my fix/edits above

christinerogers (changes requested):
edit to fix the content of the help text blurb.  @laemtl can you able to find a way to accept this suggestion ?   (or Alex could probably find time)

christinerogers (approved):

laemtl:
This PR and PR#7025 are in conflicts. The last to be merged will have to update raisinbread/RB_files/RB_help.sql to remove the parendID column.

christinerogers:
great catch, thanks @laemtl

christinerogers:
just to check -- is this ready for @driusan  to review, @zaliqarosli @laemtl ?

zaliqarosli:
@christinerogers i believe so! i'll give it one last test

laemtl:
Yes, it's ready for review :)

zaliqarosli:
@driusan should this be closed? ^ i think github did it automatically

AlexandraLivadas:
@zaliqarosli  This PR has been rebased and is ready for testing :)

zaliqarosli:
is discussion still required here?

AlexandraLivadas:
@zaliqarosli  No, I don't think so. I think this was approved a while ago and just needed to be tested and reviewed again on the 23.0 branch

zaliqarosli (commented):
everything else looks good. just wondering about one leftover comment. i am happy to review this PR as is though

zaliqarosli (approved):

zaliqarosli:
could someone test this? i'm having trouble with my dev VM. @jstirling91 would you be interested?

modules/help_editor/ajax/help.php
```diff
@@ -32,20 +32,31 @@
...
-            md5($subpageName ?? $moduleName)
-        );
-    }
+        try {
+            $helpID    = \LORIS\help_editor\HelpFile::hashToID(
+                md5($subpageName ?? $moduleName)
+            );
+            $help_file = \LORIS\help_editor\HelpFile::factory($helpID);
+            $data      = $help_file->toArray();
+
```
zaliqarosli, line 41:
```suggestion
```

zaliqarosli (changes requested):
everything else looks good! will test now

zaliqarosli:
@AlexandraLivadas any chance you could edit the help_editor .md help file to say "This module displays existing help content for LORIS instrument pages" instead of what's there?

Screen Shot 2021-07-06 at 11 46 37 AM

ridz1208 (approved):

ridz1208:
@driusan this one is in your hands, it's going to 23 since it's a bugfix requested by IBIS

modules/help_editor/help/help_editor.md
```diff
@@ -1,7 +1,10 @@
 # Help Editor
 
-This module displays existing help content for LORIS modules and pages. 
+This module displays existing help content for LORIS instrument pages
```
zaliqarosli, line 3: full stop at the end, sorry for being picky!

zaliqarosli (changes requested):

zaliqarosli (approved):
LGTM!!

modules/help_editor/ajax/help.php
```diff
@@ -32,20 +32,30 @@
...
+                md5($subpageName ?? $moduleName)
+            );
+            $help_file = \LORIS\help_editor\HelpFile::factory($helpID);
+            $data      = $help_file->toArray();
+        } catch (\NotFound $e) {
+            // Send data with empty strings so that the content can be edited
+            $data = [
+                'content' => '',
+                'topic'   => '',
+                'updated' => '-'
```
zaliqarosli, line 46: @AlexandraLivadas sorry just noticed, this should be empty

ridz1208: @zaliqarosli any other comments??

```suggestion
                'updated' => ''
```

zaliqarosli: this should be the last one! i can re-approve once the above suggested change is comitted

ridz1208:
@driusan i think this one is yours as well

zaliqarosli (approved):
ill approve again since the last commit

PR #6916, opened 2020-08-18
[Tools] Enable PHPCS for exporters/data_dictionary_builder.php

laemtl:


laemtl:
@driusan Rebased and ready to be reviewed.

PR #6910, opened 2020-08-13
[Documentation] Adding Developers Guide to docs

AlexandraLivadas:
This document offers general guidance to developers (both inside and outside the LORIS team) on how to contribute to LORIS and the general workflow of LORIS code development. It is still in progress. 

As of the latest commit, this doc was added to ReadTheDocs. To see a preview of how it renders, go to this link: https://loris-fork.readthedocs.io/en/2020-08-13-developersguidemarkdown/docs/wiki/99_Developers/Developers_Guide.html

TODO:

- [x] Complete the Code Review section

- [x] Add more detail to the Accessing a Database section

- [x] Fix formatting and wording

PR #5352, opened 2019-10-22
Use the SessionID class in LORIS

johnsaigle:
## Brief summary of changes

This updates core libraries and related code to use the SessionID class.

Some TLC and optimization is also given to aging code in these files.

#### Links to related tickets (GitHub, Redmine, ...)

* Resolves #4369 
* Class first created (but not implemented) in #4333
* Assists with push for stronger typing #4210

PR #6932, opened 2020-08-25
[survey_account] fix can't add survey to candidate in main branch

kongtiaowang:
## Brief summary of changes

1. Fixing the SessionID type in addSurvey function.
According to @driusan 's commnet in https://github.com/aces/Loris/pull/5352#discussion_r450254444
"The semantics of this function shouldn't be changed to throw an exception. This isn't backwards compatible and is unrelated to using the SessionID class."
2. Fixing the return value of getSession.

Add a new survey for a candidate.


laemtl:
This may fix #6910 6913 as well.

PR #6936, opened 2020-08-26
New Data Dictionary framework and module

driusan:
This updates the LORIS Data Framework to include a concept of a data dictionary that a module may manage. It adds a new "dictionary" module to manage the module dictionary but leaves the `datadict` module in place because it's currently the only place to manage the `parameter_type` table used by the DQT.

The extension of the data framework to include dictionaries also includes many enhancements from the current "data dictionary" in LORIS which only includes a field name and description. Types, Scopes, and Cardinalities are added which more thoroughly describe the data and will eventually be able to be used to build richer data-type sensitive interfaces in places like the DQT. DictionaryItems implement the AccessibleResource interface to enable modules to give column-level granularity over data.

Modules provide categories of DictionaryItems, which allows them to group items in the frontend. This is primarily to make it possible to filter the instruments from the "instruments" module (in the current `data_dict` module, the `Source From` options are literally populated by a call to `\Utility::getAllInstruments()`) in a manageable way, but can be used by any module that wants to group the fields.

Instruments written in PHP must now implement a new `getDataDictionary()` abstract function to describe the fields provided by the instrument, but a `LorisFormDictionaryImpl` is provided which populates it in the same way that the `lorisform_parser` uses it to populate the `parameter_type` table. The `candidate_parameters` and `imaging_browser` also implement dictionaries in this PR.

NB: I intend to try to break this PR into smaller, easier to review chunks, but it may be helpful to reviewers to have a PR with the entire dictionary overhaul for context.

PR #6938, opened 2020-08-27
Add Scope, Cardinality, Type, and DictionaryItem classes

driusan:
This adds the classes representing the data scope, data type,
data cardinality, DictionaryItem (and dictionary Category) from
PR #6936 but does not use yet use them.

### DIFF
--- a/src/Data/Cardinality.php
+++ b/src/Data/Cardinality.php
@@ -1,0 +1,100 @@
+<?php
+namespace LORIS\Data;
+
+/**
+ * Cardinality represents the number of data points which
+ * apply to the scope of a data type.
+ *
+ * Since the Cardinality class represents an enumeration, the
+ * class is final.
+ *
+ * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ */
+final class Cardinality implements \JsonSerializable
+{
+    // Valid cardinality types for data to apply to.
+
+    /**
+     * A Unique Cardinality signifies that the data is unique
+     * across the scope. Examples of unique data are CandID
+     * for the candidate scope or VisitLabel for the Session
+     * scope.
+     */
+    const UNIQUE = 1;
+
+    /**
+     * A Single Cardinality signifies that each data point in
+     * the scope should have exactly one value. For instance,
+     * date of birth for a candidate in the candidate scope.
+     */
+    const SINGLE = 2;
+
+    /**
+     * An Optional Cardinality signifies that each data point
+     * in the scope may have zero or one value. For instance,
+     * the date of death for a candidate in the candidate scope.
+     */
+    const OPTIONAL = 3;
+
+    /**
+     * A Many Cardinality signifies that each data point will
+     * have zero or more values associated. For instance,
+     * the T1 scans acquired at a session.
+     */
+    const MANY = 4;
+
+    protected $cardinality;
+
+    /**
+     * Constructs a Scope object. $scope should be a class constant
+     * to construct the scope for, not an int literal.
+     *
+     * @param int $scope The scope
+     */
+    public function __construct(int $card)
+    {
+        switch ($card) {
+            case self::UNIQUE: // fallthrough
+            case self::SINGLE: // fallthrough
+            case self::OPTIONAL: // fallthrough
+            case self::MANY: // fallthrough
+                $this->cardinality = $card;
+                break;
+            default:
+                throw new \DomainException("Invalid cardinality");
+        }
+    }
+
+    /**
+     * Convert the enumeration from a memory-friendly integer to a
+     * human-readable string when used in a string context.
+     *
+     * @return string
+     */
+    public function __toString() : string
+    {
+        switch ($this->cardinality) {
+            case self::UNIQUE: // fallthrough
+                return "unique";
+            case self::SINGLE: // fallthrough
+                return "single";
+            case self::OPTIONAL: // fallthrough
+                return "optional";
+            case self::MANY: // fallthrough
+                return "many";
+            default:
+                return "invalid cardinality";
+        }
+    }
+
+    /**
+     * Implement the JsonSerializable interface by
+     * converting to a string
+     *
+     * @return string
+     */
+    public function jsonSerialize() : string
+    {
+        return $this->__toString();
+    }
+}
--- a/src/Data/Dictionary/Category.php
+++ b/src/Data/Dictionary/Category.php
@@ -1,0 +1,78 @@
+<?php
+declare(strict_types=1);
+namespace LORIS\Data\Dictionary;
+
+/**
+ * A \LORIS\Data\Dictionary\Category represents a grouping of
+ * DictionaryItems.
+ *
+ * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ */
+class Category
+{
+    protected $name;
+    protected $description;
+    protected $items = null;
+
+    /**
+     * Construct a dictionary Category
+     *
+     * @param string $name             The machine name of the category
+     * @param string $desc             The human readable description of
+     *                                 the category
+     * @param ?DictionaryItem[] $items An optional iterable of items which
+     *                                 the category contains.
+     */
+    public function __construct(string $name, string $desc, ?iterable $items = null)
+    {
+        $this->name        = $name;
+        $this->description = $desc;
+        $this->items       = $items;
+    }
+
+    /**
+     * Return the name of the Category
+     *
+     * @return string
+     */
+    public function getName() : string
+    {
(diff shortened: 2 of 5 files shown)


src/Data/Dictionary/DictionaryItem.php
```diff
@@ -0,0 +1,117 @@
...
+use \LORIS\Data\Type;
+use \LORIS\Data\Cardinality;
+
+/**
+ * A DictionaryItem represents a description of a type of data
+ * managed by LORIS.
+ *
+ * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ */
+class DictionaryItem implements \LORIS\StudyEntities\AccessibleResource
```
xlecours, line 15: abstract?

xlecours: Nervermind. I just looked at the "Main" PR and the intention is not for module to create nes classes of DictionnaryItem but to instanciate objects.

driusan: yes, but they can extend it to customize the accessibility if need be.

src/Data/Dictionary/Category.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public function getDescription() : string
+    {
+        return $this->description;
+    }
+
+    /**
+     * Return the items which belong to the Category
+     *
+     * @return ?DictionaryItem[]
```
maltheism, line 56: Why ?iterable when comment says @return ?DictionaryItem[]

src/Data/Scope.php
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Constructs a Scope object. $scope should be a class constant
+     * to construct the scope for, not an int literal.
+     *
+     * @param int $scope The scope
+     */
+    public function __construct(int $scope)
+    {
+        switch ($scope) {
+            case self::CANDIDATE: // fallthrough
```
maltheism, line 34: Shouldn't there be a `break;` for case self::CANDIDATE:

src/Data/Dictionary/Category.php
driusan: Because as far as I know there's no way to specify what type it's an iterable of in the php type hint.

src/Data/Scope.php
driusan: no, it intentionally falls through (hence the line comment)

maltheism: Oh I thought break is necessary to avoid passing through to the code under the next label

driusan: it does intentionally pass through to the code under the next label (which just sets the internal value to the variable $scope)

src/Data/Cardinality.php
```diff
@@ -0,0 +1,100 @@
...
+ * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ */
+final class Cardinality implements \JsonSerializable
+{
+    // Valid cardinality types for data to apply to.
+
+    /**
+     * A Unique Cardinality signifies that the data is unique
+     * across the scope. Examples of unique data are CandID
+     * for the candidate scope or VisitLabel for the Session
```
ridz1208, line 20: if you mean the SQL sessions table that's not entirely correct. if you mean the visits table (ie list of possible visits independent of any candidate) then it is correct but confusing here

src/Data/Cardinality.php
```diff
@@ -0,0 +1,100 @@
...
+     * An Optional Cardinality signifies that each data point
+     * in the scope may have zero or one value. For instance,
+     * the date of death for a candidate in the candidate scope.
+     */
+    const OPTIONAL = 3;
+
+    /**
+     * A Many Cardinality signifies that each data point will
+     * have zero or more values associated. For instance,
+     * the T1 scans acquired at a session.
```
ridz1208, line 42: feels like the example is a bit too specific, why not use something like `For instance, site affiliations for a user` for us non-imaging people

src/Data/Dictionary/DictionaryItem.php
```diff
@@ -0,0 +1,117 @@
...
+     *                           applies
+     * @param Type        $t     The data type of this dictionary item
+     * @param Cardinality $c     The data cardinality
+     */
+    public function __construct(
+        string $name,
+        string $desc,
+        Scope $scope,
+        Type $t,
+        Cardinality $c
```
ridz1208, line 37: why single letter variables, i'm not a big fan and in the long run they always tend to be a hindrance (same goes for shortcuts like `cand` for `candidate`...)

src/Data/Dictionary/DictionaryItem.php
```diff
@@ -0,0 +1,117 @@
...
+ * managed by LORIS.
+ *
+ * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ */
+class DictionaryItem implements \LORIS\StudyEntities\AccessibleResource
+{
+    protected $name;
+    protected $description;
+    protected $scope;
+    protected $typ;
```
ridz1208, line 20: `typ` ?? not `type` ?

src/Data/Scope.php
```diff
@@ -0,0 +1,67 @@
+<?php
+namespace LORIS\Data;
+
+/**
+ * A Scope is an enumeration class which represents the scope
+ * that a piece of data may apply to in LORIS.
+ *
+ * The Scope class is final because the list of enumeration types
+ * can not be dynamically extended without modifying all places
+ * that must deal with the enumeration options.
```
ridz1208, line 10: this seems limiting to projects wanting to define their own scopes... why not a database table defining the constants...

would avoid overriding/overloading specially since we dont offer any easy overriding structure for the `src` directory

ridz1208 (commented):
just a few comments on design decisions. nothing serious

src/Data/Cardinality.php
driusan: I don't mind changing the example but I think site/project affiliations is a bad example because there's no "user" scoped data, and the example might get confusing since site/project affiliations exist on both candidate and session scopes.

src/Data/Dictionary/DictionaryItem.php
driusan: `type` is a reserved keyword in php.

driusan: at least I thought it was and editor was highlighting it as one.. I'll look into it a little more.

src/Data/Scope.php
driusan: Scopes can't be extensible because everywhere in the code that deals with the data needs to understand how to deal with that scope. Let's say a module added a scope named "foobar" and someone accessed that a variable with that scope in the DQT. Questions like "how should this be displayed? How does it relate back to candidates? How does it relate to sessions? Does it need to show a list of sessions to select? How does it get displayed in the results table? How do I join it with candidate scoped data to show in a row?" need to be understood by the code to build the interface and display results.

driusan: (if you have a suggestion for how to make the comment more clear I'll update it.)

ridz1208: no your right... I'm just thinking if there is a way we can add project defined scopes in the same way. my immediate interest is the Biospecimen scope for CBIGR

ridz1208: it could come later though

src/Data/Cardinality.php
ridz1208: Edit: my comment is incorrect. you dont need to answer it

src/Data/Cardinality.php
ridz1208: how about using the candidate/session relationship then ? a candidate can have 0 to many sessions

src/Data/Dictionary/DictionaryItem.php
driusan: It's just the argument name in the constructor, the context only lasts long enough to assign it to a property right below.. the property name is the full word.

src/Data/Cardinality.php
driusan: That would be confusing too though, because there's the session scope and there's no variable named "Session", and session based data is usually described as 1:1 with a session scope, not 1:many with the candidate.

driusan (commented):
for some reason my replies to comments aren't being submitted unless I click submit a review..

src/Data/Scope.php
driusan: There are similar problems with things like imaging QC. "Pass/Fail" is really file scoped data but there is no file scope. For now I've defined them as "many" cardinality scoped to the session.

ridz1208 (approved):

PR #6939, opened 2020-08-27
Add data type classes

driusan:
This implements the known types of types from PR #6936.


driusan:
@zaliqarosli can you review these types for compatibility with the JSON instruments spec?

src/Data/Types/URI.php
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    /**
+     * URIs are stored in the SQL database as varchar columns.
+     *
+     * @return string
+     */
+    public function asSQLType() : string
+    {
+        return "varchar(255)";
```
xlecours, line 39: This can be tight. What about text?

src/Data/Types/Interval.php
```diff
@@ -0,0 +1,37 @@
+<?php
+namespace LORIS\Data\Types;
+
+/**
+ * An Interval data type represents a time interval or duration.
+ *
+ * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ */
+class Interval implements \LORIS\Data\Type
```
xlecours, line 9: Shouldn't this be an abstract class with a constructor that takes 2 `\LORIS\Data\Type` of the same class?
We would have a DateInterval, TimeInterval? 
I am not sure how or where this is meant to be used.

driusan: What's the difference between a date interval and a time interval?.. I think I originally called it "Duration" but after using it for candidate age an age being a duration didn't quite sit right with me.

src/Data/Types/URI.php
driusan: text seems excessive, what do we do for existing URLs in the database? I based this on the `File` column in the files table.

xlecours: We usually construct them using chunks from the database (e.g: basename, host, ...).
I tested our database class and it throws an exception when mysql emits a warning (data too long in this case.) But I am not sure if the mri_pipeline perl script does the same.

If we use text instead of varchar(255), it will cost 1 additional byte per row.  I don't think the cost is high enough to counterweight the flexibility is gives.

src/Data/Types/Interval.php
xlecours: I would expect the unit to be different. 
The constructor would take 2 dates, 2 time or 2 whatever.

zaliqarosli (dismissed):
maybe boolean and multi select are missing?

driusan:
Good point about boolean, but I don't see how `multiselect` is a type of data.

src/Data/Types/URI.php
driusan: Sorry for the delay responding.. my concern isn't that it's too heavy or light, it's that it doesn't match what our columns in the database currently are. Regardless, I changed it to "text" since this isn't actually used anywhere and the dictionary items that will end up using it will probably end up being constructed, as you say..

src/Data/Types/Interval.php
driusan: Why would it take 2 anything? This is the data type, not the instance of the data. The instance of the data can be whatever magnitude (day, time, whatever) that it wants.

xlecours: Maybe I misunderstand the meaning of "Interval". In my mind, it is the distance measured in "unit" between two "things".

src/Data/Types/DecimalType.php
```diff
@@ -0,0 +1,47 @@
...
+ * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ */
+class DecimalType implements \LORIS\Data\Type
+{
+    /**
+     * Construct a DecimalType
+     */
+    public function __construct()
+    {
+    }
```
ridz1208, line 17: why do some classes have empty constructors like this and others no constructors (like integer)

src/Data/Types/Interval.php
ridz1208: @driusan i guess i'm confused too, if that's the case why does enumeration take a list of options if its the datatype not the instance ?

it seems like the min/max for interval is equivalent to options for enumeration

(notice here im not comparing to the maxsize in `integer` because that seems to be a meta-value required to define the size of the int rather than the int itself)

src/Data/Types/Enumeration.php
```diff
@@ -0,0 +1,65 @@
...
+class Enumeration implements \LORIS\Data\Type
+{
+    protected $options = [];
+
+    /**
+     * Construct an Enumeration type
+     *
+     * @param string ...$values The options for this enumeration
+     */
+    public function __construct(string ...$values)
```
ridz1208, line 19: out of curiosity, why the splat operator instead of an array ?

src/Data/Types/DecimalType.php
driusan: Probably leftover from when I was first writing it, I'll remove the empty constructors.

src/Data/Types/Enumeration.php
driusan: I find in this case it just looks more aesthetically pleasing to do ie `new Enumeration('Male', 'Female', 'Other')`.

On a more technical level it also ensures that it's an array of strings, while `array` would be any arbitrary array with no type enforcement (for the key/values)

src/Data/Types/Interval.php
driusan: An Interval is the time between 2 points in time. For instance: the data type for candidate age or an instrument question asking "how long do you sleep each night?" There's no start/end point, there's just a duration of time. It can be days, minutes, seconds, or years, that's just a question of magnitude, not type.

I don't know what min/max you're talking about for intervals?

The constructors take the amount of parameters they need to describe the type. An enumeration data type needs to know what the options are because it's a part of the type. Two enumerations with different options aren't the same. A string needs to know the max size to know the SQL data type that it corresponds to. The integer should have the `maxsize` removed (at least for now), it looks like it's not used and I think it's just leftover from basing the file on the string type initially (but later it might be a good idea to add the ability to define constraints such as max or min size..)

driusan: To make things concrete, looking at #6936 Interval is the data type currently used for Candidate Age and Window Difference in instruments. How would you describe the data type of those fields?

driusan: Looking at https://en.wikipedia.org/wiki/ISO_8601#Durations and comparing the ISO8601 definition of "Time Interval" and "Duration" it might be better to go back to calling them "Durations", but it doesn't feel right calling a window difference or an candidate age a duration to me.. I'm open to any other suggestions.

src/Data/Types/Enumeration.php
ridz1208: I guess I have 3 concerns:
1. not sure what code is currently using it but in the event where we try to build the enum on the go, an array offers the advantage of being able to append and prepend until we are ready to create the enum. if each element needs to be specified manually it makes creating the enum progressively almost impossible.
2. it seems much more difficult to build the enum from the results of a query, especially if the enum is of a large list
3. why do we only want strings? are int enums not an option ?

driusan: For 1 and 2: I don't see why that's a problem. You can build an array as normal and then use the splat operator to expand the array such as here: https://github.com/aces/Loris/pull/6936/files#diff-88088f05f1a41366915499faede2a27ee316859e38c7990a86662680c1b89f4eR289

For 3: no. At least not from a MySQL perspective.

```
mysql> create temporary table foo ( id enum(2, 3) );
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '2, 3) )' at line 1
mysql> create temporary table foo ( id enum('2', '3') );
Query OK, 0 rows affected (0.01 sec)
```

from a conceptual perspective enums are neither ints nor strings. They're "one of a set of things" independent of the internal representation, but we always want the labels used to identify them to be strings (much like mysql does in the above example..)

src/Data/Types/Interval.php
ridz1208: I guess the confusion for me stemmed from the fact that I thought the "interval" was computed, in this function, from 2 days, 2 times, 2 whatever (thus the min/max values i mentioned previously). I also did not necessarily associate an interval with only points in time... and interval can be [a space between objects, units, points, or states](https://www.merriam-webster.com/dictionary/interval).

Thinking of it now the interval/duration you are talking about is basically an int/decimal with a unit. does it need it's own data type? isnt the candidate age and int of years, months, days... ? 

When i hear "interval" i automatically think of 2 values... not 1 value with a unit so it might just be a terminoology issue

driusan: Okay, I've switched it from Interval to Duration. I don't know if it makes sense to extend from a programmatic perspective (they don't really have much behaviour to inherit and if they do I'm not sure if they'll be more or less likely to be the same..) but I made it extend Decimal.

src/Data/Types/Enumeration.php
ridz1208: Yeah for 1 and 2 I edited my comment. I only realised it after.

3 makes sens as well, i guess I was thinking more of PHP arrays than SQL enums.

src/Data/Types/Interval.php
ridz1208: I think it makes more sens like that. and yeah you'r right there isn't much to inherit.

ridz1208:
@driusan to follow up on @zaliqarosli 's comment, how would you represent a "selectmultiple" as a datatype? The instinctive choice is using an enum but shouldn't you be able to see from the dictionary if this particular field can have multiple options or only one ? 

It seems like we can make our own rules for enums here. Java defines it's enums as `a special data type that enables for a variable to be a set of predefined constants. The variable must be equal to **one** of the values that have been predefined for it. ` we could define it as `one or more` for loris I guess ?

I realise the datatype itself only offers a static list of possible options, I'm just wondering if we need to add a flag on the field offering additional info such as is it a multiselect or a single select...

driusan:
The Cardinality defines how many values the field can have separately from the Type.

ridz1208:
> The Cardinality defines how many values the field can have separately from the Type.

so I guess the `selectmultiple` case in #6936 should be fixed to have a different cardinality than the `select`

see: https://github.com/aces/Loris/pull/6936/files#diff-1d18bbc27ba2cabe74bdf1a067307ce735f0f84fd29c735010eb122abc370c14L635

ridz1208 (approved):

driusan:
I think you're right, I added Needs Work and commented on that PR..

src/Data/Types/Interval.php
driusan: I took away the extends because the more I think about it the more I think it might be a limiting assumption. For instance, the example in the above ISO8601 duration:

```
For example, "P3Y6M4DT12H30M5S" represents a duration of "three years, six months, four days, twelve hours, thirty minutes, and five seconds". 
```

That *could* be represented as a whole number of seconds, but it's not a very natural way of doing it.

ridz1208:
@zaliqarosli your changes have been addressed, could you re-review or dismiss please

PR #6942, opened 2020-09-01
[API] Users get candidates from own projects only in endpoint /candidates that list all candidates

### LINKED ISSUE #6931: [API] Add UserProjectMatch to /candidates
**Describe the bug**
The list of candidates returned by the `/candidates` endpoint is not filtered for projects that the user has access to.

See candidate_list for example
https://github.com/aces/Loris/blob/main/modules/candidate_list/php/candidate_list.class.inc#L59-L62
https://github.com/aces/Loris/blob/main/php/libraries/DataFrameworkMenu.class.inc#L119-L123

spell00:
## Brief summary of changes
The list of candidates returned by the /candidates endpoint is filtered for projects that the user has access to. Only candidates from projects a user has access to is returned.

#### Testing instructions
1- Login your VM with admin
2- Make sure one of the users don't have access to all projects. 
3- Login as the user that don't have access to all projects.
4- Go to <hostname>/api/v0.0.3/candidates
5- Make sure only the projects the user has access to appear in the list displayed on the page

* Resolves #6931

### DIFF
--- a/modules/api/php/endpoints/candidates.class.inc
+++ b/modules/api/php/endpoints/candidates.class.inc
@@ -79,6 +79,17 @@ class Candidates extends Endpoint implements \LORIS\Middleware\ETagCalculator
     }
 
     /**
+     * Tells the base class that this page's provisioner can support
+     * the UserProjectMatch filter.
+     *
+     * @return bool always true
+     */
+    public function useProjectFilter() : bool
+    {
+        return true;
+    }
+
+    /**
      * Handles a request starts with /candidates
      *
      * @param ServerRequestInterface $request The incoming PSR7 request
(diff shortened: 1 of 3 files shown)


modules/api/php/endpoints/candidates.class.inc
```diff
@@ -154,6 +165,12 @@ class Candidates extends Endpoint implements \LORIS\Middleware\ETagCalculator
         $provisioner = (new \LORIS\api\Provisioners\CandidatesProvisioner())
             ->filter($filter);
 
+        if ($this->useProjectFilter()) {
```
xlecours, line 168: Why add a function that always return true?

spell00: I took this part from the example you suggested https://github.com/aces/Loris/blob/main/modules/candidate_list/php/candidate_list.class.inc#L59-L62. I thought maybe it was to be completed and it would become meaningful, but I don't see a case where filters would not be used, so I'm removing it.

PR #6780, opened 2020-06-29
[API] PUT and PATCH methods added to Candidates visit/instruments/flags

spell00:
## Brief summary of changes

The methods `handlePUT` and `handlePATCH` were added to the endpoint `/candidates/<candid>/<visit>/instruments/<instrument>`.

#### Testing the PUT handle method

1. Go to `https://<hostname>/api/v0.0.3/candidates/300002/V1/instruments/aosi/flags` in a browser.

2. The fields `Data_entry`, `Administration` and `Validity` field should be `NULL`
3. In a terminal, login by sending the POST request below:
`~$ curl https://<hostname>/api/v0.0.3/login -d '{"username": "<username>", "password": "<password>"}'`
4. The expected response is: {"token": "<a-really-long-string>"}. Store the token in a variable `token='<a-really-long-string>'`.

5. In the terminal, enter `curl -X PUT -H "Authorization: Bearer $token" https://<hostname>/api/v0.0.3/candidates/300002/V1/instruments/aosi/flags -d '{"Flags": {"Data_entry": "In Progress", "Administration": "All", "Validity": "Valid"}}'`

6. Go back to `https://<hostname>/api/v0.0.3/candidates/300002/V1/instruments/aosi/flags` in the same browser as in step 1. `Data_entry` should be `"In Progress"`, `Administration` should be `"all"` and `Validity` should be "Valid".

7. In the terminal, enter `curl -X PUT -H "Authorization: Bearer $token" https://<hostname>/api/v0.0.3/candidates/300002/V1/instruments/aosi -d '{"Flags": {"Data_entry": "In Progress", "Administration": null, "Validity": "Questionable"}}'`

8. The field `Administration` should be changed to `null` and `Validity` should be `"Questionable"`

#### Testing the PATCH handle method

1. Go to `https://<hostname>/api/v0.0.3/candidates/300002/V1/instruments/aosi/flags` in a browser.

2. The fields `Data_entry` field should be `NULL` , `Administration` should be `"Partial"` and `Validity` should be `"Questionable"`

3. In the terminal, enter `curl -X PATCH -H "Authorization: Bearer $token" https://<hostname>/api/v0.0.3/candidates/300002/V1/instruments/aosi/flags -d '{"Flags": {"Data_entry": null, "Administration": "All"}}'`

[...]

PR #6775, opened 2020-06-27
[API] Add Project endpoint "Dicoms"

spell00:
## Brief summary of changes
This Pull Request adds an endpoint at /projects/\<project\>/dicoms. It returns a complete list of DICOM files contained in a project (and other pertinent informations)

#### Testing instructions
Go to https://\<hostname\>/api/v0.0.3/projects/rye/dicoms in a browser. The page should return a JSON in a string with information on all DICOM contained in the project

#### Link(s) to related issue(s)

* Resolves #6774

PR #6161, opened 2020-03-16
[API] Adding dicom upload

xlecours:
Copy of https://github.com/aces/Loris/pull/5016

This adds 
`POST /candidates/<candid>/<visit>/dicoms` to upload a dicom.
`GET /candidates/<candid>/<visit>/dicoms/<filename>/processes` to get the server processes related to that dicom upload (mri_upload)
`POST /candidates/<candid>/<visit>/dicoms/<filename>/processes` to start a mri_upload server process on that dicom.
`GET /candidates/<candid>/<visit>/dicoms/<filename>/processes/<processid>` to get the state of a process

This work is based on https://github.com/aces/Loris/pull/4244 
To see this PR changes go to https://github.com/xlecours/Loris/compare/api-module-candidates-and-subs...xlecours:adding_dicom_upload

TODO:

- [ ] Add documentation for the new endpoints.
- [ ] Make the imaging_uploader use it.
- [x] Sync FilesUploadHandler.class.inc, UploadException.class.inc with John's PR https://github.com/aces/Loris/pull/4181

PR #5354, opened 2019-10-22
[API-docs] Adding static swagger-ui

xlecours:
## Brief summary of changes
This is adding a static version of swagger-ui to LORIS 
see: https://github.com/swagger-api/swagger-ui 

#### Testing instructions (if applicable)

go to `https://<devname>.loris.ca/api_docs`

You should see:
swagger-ui

To test with different specs, create a `static` directory in the configuration module and download [this file.](https://raw.githubusercontent.com/xlecours/swagger-schemas/configurations/configs.yml) Change the filename to schema.yml. There should be a new entry in the `Select a spec` dropdown.

PR #6944, opened 2020-09-01
Transition to API version 0.0.4

### LINKED ISSUE #6941: [API] New version of the API (v0.0.4)
This is a feature for the `api` module.

A new version of the API, v0.0.4, is required to make modifications to the API that are not described in the current specs of v0.0.3 .

spell00:
## Brief summary of changes
Transition to v0.0.4-dev to include new extensions of the API that are not in the current specs.
The changes include:
- The specs of v0.0.3 is copied into v0.0.4-dev
- The integration tests are changed to test v0.0.4-dev
- All current endpoints allow v0.0.4-dev of the API

* Resolves #6941


spell00:
The version new name needs to be discussed.

xlecours (changes requested):

spell00:
@xlecours Could you please have a look at this?

xlecours (changes requested):
Could you add something in CHANGELOG.md to mention that there is a new version of the api specification in the modules section. We will add items under it for each new or changed endpoints.

xlecours:
@driusan This PR set the table for additional endpoints and modifications of existing endpoints. I will evaluate if the following PR can be include in v0.0.4
https://github.com/aces/Loris/pull/6905
https://github.com/aces/Loris/pull/6899
https://github.com/aces/Loris/pull/6780
https://github.com/aces/Loris/pull/6775
https://github.com/aces/Loris/pull/6161
https://github.com/aces/Loris/pull/5354

Meanwhile, I think this PR is the first step. Would you prefer to have a feature branch instead?

CHANGELOG.md
```diff
@@ -120,4 +122,4 @@ be used by projects having custom modules not in LORIS. (PR #5913)
 - The tool `phpstan` has been added to our automated test suite. (PR #4928)
 - Config files for static analysis have been moved to the `test/` directory. (PR #5871)
 - Dashboard was refactored to turn panels into module widgets. (PR #5896)
-- Add CSSGrid component type (PR #6090)
\ No newline at end of file
+- Add CSSGrid component type (PR #6090)
```
spell00, line 125: I don't see why this is highlighted as a change, it looks like both lines are exactly the same

xlecours, line 125:
```suggestion
- Add CSSGrid component type (PR #6090)
```

xlecours: I don't know what changed but copy-pasting the line seems to propose a change.

spell00: it's impossible to commit the suggestion though, it says `this suggestion is invalid because no changes were made to the code`

xlecours (approved):

driusan (changes requested):
Since there's still other PRs to go into 0.0.4, please add the -dev suffix to the version number until it's ready. Otherwise we'll have multiple incompatible versions of "0.0.4".

Other than that it looks good to me.

xlecours:
> Since there's still other PRs to go into 0.0.4, please add the -dev suffix to the version number until it's ready. Otherwise we'll have multiple incompatible versions of "0.0.4".
> 
> Other than that it looks good to me.

@driusan it is not released.It is implicitly `-dev`. Otherwise, we would need to change all the `0.0.4` to `0.0.4-dev` then back to `0.0.4` before the release.

spell00:
@xlecours @driusan  So what should it be? Is `-dev` because there is not a new version of the API at each release? How is it decided when it is time to remove the `-dev` suffix and start a new version? It's an easy and fast update to do, I wonder if updating the version systematically on each release wouldn't be faster and more straightforward, as it would avoid having a discussion on every release about removing or adding `-dev`? 

Also, I thought it was implicitly `-dev` too, because it is `v0`. Wouldn't it make sense if all official new version of the API would update the first number (e.g. v0.0.4 -> v1.0.0) and all updates of the last number would be implicitely a `dev` version (e.g. v1.0.0  -> 1.0.1 -> 1.0.2)?

christinerogers:
@driusan i think the naming is now resolved per today's loris meeting. 
Could you please re-review? 
 this PR is holding up several other ones and will be key for release.

modules/api/docs/LorisRESTAPI_v0.0.4-dev.md
```diff
@@ -1,5 +1,4 @@
-# Loris API - v0.0.2
```
spell00, line 1: It looks like I modified a lot of things in the doc of v0.0.2, but it is actually modified from v0.0.3. I removed `LorisRESTAPI.md` and added `LorisRESTAPI_v0.0.4-dev.md`. 

I did not modifiy the original `LorisRESTAPI_v0.0.3.md` so Github must consider I replaced `v0.0.2` by `v0.0.4-dev`.

modules/api/php/endpoints/candidate/visit/visit.class.inc
```diff
@@ -75,8 +75,10 @@ class Visit extends Endpoint implements \LORIS\Middleware\ETagCalculator
      */
     protected function supportedVersions() : array
     {
-        // Removed 0.0.2 since session requires a project.
```
spell00, line 78: this comment is no longer relevant, since 0.0.2 is no longer supported anywhere

modules/api/docs/LorisRESTAPI_v0.0.4-dev.md
spell00: I solved this problem by renaming `LorisRESTAPI_v0.0.3.md` to `LorisRESTAPI.md`.

spell00:
@driusan @xlecours as per our discussion this morning, I have removed support for `v0.0.2` in all endpoints. I also removed the doc for `v0.0.2`.

I am not sure if @xlecours agreed with the use of the `-dev` suffixe, but I think we should proceed with this PR as it is (with `-dev`). If we were to decide to change the version to remove `-dev`, it is very fast to change it.

From the directory `modules/api/php/endpoints/`, I would just execute the command line:

`grep -rli 'v0.0.4-dev' * | xargs -i@ sed -i 's/v0.0.4-dev/v0.0.4/g' @`

cc @christinerogers

christinerogers:
@xlecours @driusan could we get some joint attention on this PR? It's blocking the rest of Simon's work from moving forward, and this needs to be resolved in time.

xlecours:
ok @driusan let's move forward.

PR #6947, opened 2020-09-02
[dataquery] Eslint fixes

### LINKED ISSUE #6750: [eslint] Fix warnings

laemtl:
It's the end of an amazing journey, thousands lines of warnings are finally gone for good.

Also edited webpack.config.js to get rid of:

```
WARNING in configuration
    The 'mode' option has not been set, webpack will fallback to 'production' for this value. Set 'mode' option to 'development' or 'production' to enable defaults for each environment.
    You can also set it to 'none' to disable any default behavior. Learn more: https://webpack.js.org/configuration/mode/

```
Resolves #6750

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -122,6 +122,7 @@ function lorisModule(mname, entries, override=false) {
 const config = [
   // Core components
   {
+    mode: 'production',
     entry: {
       DynamicDataTable: './jsx/DynamicDataTable.js',
       PaginationLinks: './jsx/PaginationLinks.js',
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -2,6 +2,8 @@ const TerserPlugin = require('terser-webpack-plugin');
 const CopyPlugin = require('copy-webpack-plugin');
 const path = require('path');
 const fs = require('fs');
+const xml2js = require('xml2js');
+const parser = new xml2js.Parser();
 
 const optimization = {
   minimizer: [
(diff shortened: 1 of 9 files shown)


driusan:
No longer blocked, probably needs to be rebased..

webpack.config.js
```diff
@@ -122,6 +122,7 @@ function lorisModule(mname, entries, override=false) {
 const config = [
   // Core components
   {
+    mode: 'production',
```
maltheism, line 125: I don't believe this line should be here.

maltheism:
@laemtl Load & Save are missing.

Before:


After:

maltheism:
@laemtl you're missing the code around here https://github.com/aces/Loris/blob/12e384ad9b532f9b0c79a928e5666a72c1d79312/modules/dataquery/jsx/react.app.js#L200

webpack.config.js
laemtl: Is there a better place?

maltheism: The mode could be either production or development. I don't believe it should be statically written as production. Loris currently has an option in the project/config.xml whether in development or production mode. It would be best if we could grab the switch logic from there and then set depending on it.

laemtl: Didn't know that. Thanks, will look into it.

laemtl:
@maltheism Thank you for this catch. I deleted the code by accident because eslint was complaining the var was unused. Fixed!

driusan:
```
/home/travis/build/aces/Loris/webpack.config.js
    5:1   error  Unexpected var, use let or const instead        no-var
    6:1   error  Unexpected var, use let or const instead        no-var
  127:41  error  Unexpected space before function parentheses    space-before-function-paren
  128:5   error  Expected space(s) after "if"                    keyword-spacing
  131:1   error  Line 131 exceeds the maximum line length of 80  max-len
```

laemtl:
@driusan Solved.

maltheism:
@laemtl `label` in react.tabs.js on line 658 is undefined.

laemtl:
@maltheism Forgot to delete this line too, thanks!. The label variable is not used.

laemtl:
The changes made in webpack.config.js may become unecessary if PR #7042 is merged.

maltheism (approved):

PR #6959, opened 2020-09-03
[Battery Manager] Testplan update

### LINKED ISSUE #6956: [Battery Manager]  Test plan update
- **Check that you can only enter numbers between 0 and 99999 in Minumum age (days) and Maximum age (days).** Value can be bigger than 99999.
- **Press the Deactivate button in the Change Status column on an entry in the data table.
Ensure that a warning message appears that asks you to confirm the action.** There is no confirmation.
- **Check that when you try to edit an entry without making changes to the form, you receive an error message.** There is no error.
- **Ensure the original entry was deactivated and the new entry has been added to the data table.** The original entry was modified, not copied and desactivated.
- **Check that you can only enter a site that exists.** Site is a dropdown so there is no point to test this.

laemtl:
* Resolves #6956

### DIFF
--- a/modules/battery_manager/test/TestPlan.md
+++ b/modules/battery_manager/test/TestPlan.md
@@ -20,7 +20,7 @@ Test Battery.
 ### Permissions
 
 **Testing with no permissions** [Automation Testing]
-  1. Access the module with a user without superuser, `battery_manager_view`, `battery_manager_edit permissions`.
+  1. Access the module with a user **not having** the `superuser`, `battery_manager_view` or `battery_manager_edit` permissions.
   2. The access to module should be denied.
 
 **Testing with view permission** [Automation Testing]
@@ -45,7 +45,7 @@ Test Battery.
 
 **Testing add functionality**
   1. Check that you cannot add an entry without filling out the required fields: `Instrument`, `Minimum age (days)`, `Maximum age (days)`, `Stage`.
-  2. Check that you can only enter positive numbers (including 0) in Minumum age (days) and Maximum age (days).
+  2. Check that you can only enter positive numbers (including 0) in Minimum age (days) and Maximum age (days).
   3. Check that you can only enter positive numbers (including 0) in Instrument order.
   4. Check that when you try to add an entry that has an active duplicate in the table (Active = 'Y'), you receive an error message.
   5. Try to add an entry that does not have a duplicate.
@@ -88,7 +88,7 @@ Test Battery.
 
 **Testing edit (activate/deactivate/add) functionality**
   1. Check that you cannot edit an entry without filling out the required fields: `Instrument`, `Minimum age (days)`, `Maximum age (days)`, `Stage`.
-  2. Check that you can only enter positive numbers (inclusing 0) in Minumum age (days) and Maximum age (days).
+  2. Check that you can only enter positive numbers (including 0) in Minimum age (days) and Maximum age (days).
   3. Check that you can only enter positive numbers (including 0) in Instrument order.
   4. Check that when the edited entry has the same values as another active entry in the Test Battery, you receive an error message.
   5. Try to edit an entry so that it has the same values as another deactivated entry in the Test Battery.
--- a/modules/battery_manager/test/TestPlan.md
+++ b/modules/battery_manager/test/TestPlan.md
@@ -20,17 +20,17 @@ Test Battery.
 ### Permissions
 
 **Testing with no permissions** [Automation Testing]
-  1. Access the module with a regular user (without superuser permissions).
-  2. By default, the access to module should be denied.
+  1. Access the module with a user without superuser, `battery_manager_view`, `battery_manager_edit permissions`.
+  2. The access to module should be denied.
 
 **Testing with view permission** [Automation Testing]
-  1. Add view permission to the aforementioned user.
-  2. Battery Manager module should be accessible and only present with **one** tab (Browse).
+  1. Add `battery_manager_view` permission to the aforementioned user.
+  2. Battery Manager module should be accessible in read-only mode.
   3. The **Change Status** column should not be in the data table.
   4. The **Edit Metadata** column should not be in the data table.
 
 **Testing with edit permission** [Automation Testing]
-  1. Add edit permission.
+  1. Add `battery_manager_edit permissions` permission.
   2. Battery Manager module should now have an action button labeled "New Test".
   3. The **Change Status** column should be in the data table.
   4. The **Edit Metadata** column should be in the data table.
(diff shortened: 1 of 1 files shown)


modules/battery_manager/test/TestPlan.md
```diff
@@ -67,7 +67,6 @@ Test Battery.
 
 **Testing Change Status column**
   1. Press the `Deactivate` button in the `Change Status` column on an entry in the data table.
-     - Ensure that a warning message appears that asks you to confirm the action.
```
ridz1208, line 70: @HenriRabalais isnt there supposed to be a warning ?

modules/battery_manager/test/TestPlan.md
```diff
@@ -89,15 +88,13 @@ Test Battery.
 
 **Testing edit (activate/deactivate/add) functionality**
   1. Check that you cannot edit an entry without filling out the required fields: `Instrument`, `Minimum age (days)`, `Maximum age (days)`, `Stage`.
-  2. Check that you can only enter a site that exists.
```
ridz1208, line 92: @laemtl just for fun. open the inspector and change the value of the dropdown manually to a site that does not exist, try submitting the form and see what happens.

Having a dropdown does not guarantee that only these values can be submitted, it only offers these values but a user can edit the html to whatever they want and still submit it.

modules/battery_manager/test/TestPlan.md
```diff
@@ -89,15 +88,13 @@ Test Battery.
...
-  3. Check that you can only enter numbers between 0 and 99999 in Minumum age (days) and Maximum age (days).
-  4. Check that you can only enter numbers between 0 and 127 in Instrument order.
```
ridz1208, line 94: not sure for what this is useful ? @HenriRabalais ?

modules/battery_manager/test/TestPlan.md
laemtl: So maybe clarifying the Test plan with this procedure will help?

modules/battery_manager/test/TestPlan.md
laemtl: It's not working anymore if it is.

HenriRabalais: Umm, I don't there is for the Deactivate/Activate button.

modules/battery_manager/test/TestPlan.md
HenriRabalais: I don't think I wrote this, so I'm not sure why these limitations are imposed.

modules/battery_manager/test/TestPlan.md
HenriRabalais: @ridz1208  I may be wrong here, but just because you change the html doesn't mean that you've triggered a state change to update the actual object that will be submitted. I'm sure there's a way to manipulate the state from the front end if you're clever, but if the center really doesn't exist, the backend will throw a foreign key error. So I'm not sure if this is an issue.

HenriRabalais: I just checked and if you force-submit a centerId that doesn't exist, at the database access stage it sets the centerId to `null` without throwing an error (atleast I can't find any). I think the select drop down is sufficient enough front-end validation, so if you want an error to be thrown in this case, I think it should be handled in the back end. This validation should, in my opinion, be applied to a center/centerId object that can be used throughout LORIS.

ridz1208: I think this test case was added exactly for that reason, making sure the backend validates objects before persisting them even if they make it past the front end validations (which can be manipulated by a malicious user)

modules/battery_manager/test/TestPlan.md
ridz1208: maybe but that wouldnt be the place for it. @laemtl you can remove this line as well

modules/battery_manager/test/TestPlan.md
ridz1208: I think a bad request error should be thrown in that specific case

modules/battery_manager/test/TestPlan.md
laemtl: To be clear, which line should be removed?

ridz1208: 4. Check that you can only enter numbers between 0 and 127 in Instrument order.

HenriRabalais (approved):
lgtm

driusan:
What is the state of this PR? I can't tell if the open threads are resolved or not..

laemtl:
@driusan Everything should have been addressed.

driusan:
@ridz1208 if that's true can you approve? (looks like Henri already did..)

modules/battery_manager/test/TestPlan.md
```diff
@@ -89,15 +88,13 @@ Test Battery.
...
-  5. Check that when you try to edit an entry without making changes to the form, you receive an error message.
-  6. Check that when the edited entry has the same values as another active entry in the Test Battery, you receive an error message.
-  7. Try to edit an entry so that it has the same values as another deactivated entry in the Test Battery.
+  2. Check that you can only enter positive numbers (inclusing 0) in Minumum age (days) and Maximum age (days).
```
ridz1208, line 91:
```suggestion
  2. Check that you can only enter positive numbers (including 0) in Minimum age (days) and Maximum age (days).
```

modules/battery_manager/test/TestPlan.md
```diff
@@ -45,8 +45,8 @@ Test Battery.
 
 **Testing add functionality**
   1. Check that you cannot add an entry without filling out the required fields: `Instrument`, `Minimum age (days)`, `Maximum age (days)`, `Stage`.
-  2. Check that you can only enter numbers between 0 and 99999 in Minumum age (days) and Maximum age (days).
-  3. Check that you can only enter numbers between 0 and 127 in Instrument order.
+  2. Check that you can only enter positive numbers (including 0) in Minumum age (days) and Maximum age (days).
```
ridz1208, line 48:
```suggestion
  2. Check that you can only enter positive numbers (including 0) in Minimum age (days) and Maximum age (days).
```

modules/battery_manager/test/TestPlan.md
```diff
@@ -20,17 +20,17 @@ Test Battery.
 ### Permissions
 
 **Testing with no permissions** [Automation Testing]
-  1. Access the module with a regular user (without superuser permissions).
-  2. By default, the access to module should be denied.
+  1. Access the module with a user without superuser, `battery_manager_view`, `battery_manager_edit permissions`.
```
ridz1208, line 23:
```suggestion
  1. Access the module with a user **not having** the `superuser`, `battery_manager_view` or `battery_manager_edit` permissions.
```

ridz1208 (changes requested):
Just a few typos.

@driusan its good to go after these are fixed

laemtl:
@driusan @ridz1208 Ready for review.

ridz1208 (approved):

PR #7027, opened 2020-09-10
[media] Update datatable upon file upload without having to refresh

### LINKED ISSUE #6949: [media] Update datatable upon file upload without having to refresh
Currently, when a new file is uploaded, the media module needs to be refreshed for the new entries to appear in the datatable. It would be preferably if these new entries automatically appeared in the datatable upon file upload.

maltheism:
## Brief summary of changes

Solves #6949 

1. Visit media module
2. Upload file
3. Go visit the table and see new data in table

* Resolves #6949

### DIFF
--- a/modules/media/jsx/mediaIndex.js
+++ b/modules/media/jsx/mediaIndex.js
@@ -62,19 +62,16 @@ class MediaIndex extends Component {
   }
 
   /**
-   * Refresh table data after successful file upload
-   * and display recent data.
+   * Refresh table data with new row after successful
+   * file upload and display recent data.
+   * @param {object} data - row to add to table
    */
-  refreshData() {
-    fetch(this.props.dataURL, {credentials: 'same-origin'})
-      .then((resp) => resp.json())
-      .then((data) => this.setState({
-        data: data.Data,
-      }))
-      .catch((error) => {
-        this.setState({error: true});
-        console.error(error);
-      });
+  refreshData(data) {
+    let tableData = this.state.data;
+    tableData.push(Object.values(data));
+    this.setState({
+      data: tableData,
+    });
   }
 
   /**
@@ -262,8 +259,6 @@ class MediaIndex extends Component {
         );
       }
     };
-
-    console.log(this.state.data);
 
     return (
       <Tabs tabs={tabs} defaultTab="browse" updateURL={true}>
--- a/modules/media/jsx/mediaIndex.js
+++ b/modules/media/jsx/mediaIndex.js
@@ -30,6 +30,7 @@ class MediaIndex extends Component {
     this.fetchData = this.fetchData.bind(this);
     this.formatColumn = this.formatColumn.bind(this);
     this.mapColumn = this.mapColumn.bind(this);
+    this.refreshData = this.refreshData.bind(this);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


modules/media/jsx/uploadForm.js
```diff
@@ -281,11 +281,12 @@ class MediaUploadForm extends Component {
     let formData = this.state.formData;
     let formObj = new FormData();
     for (let key in formData) {
-      if (formData[key] !== '') {
-        formObj.append(key, formData[key]);
+      if (formData.hasOwnProperty(key)) {
```
maltheism, line 284: Note this was missing so I added it. Not really related to the PR.

modules/media/jsx/mediaIndex.js
```diff
@@ -60,6 +61,22 @@ class MediaIndex extends Component {
...
+      .then((resp) => resp.json())
+      .then((data) => this.setState({
+        data: data.Data,
+      }))
+      .catch((error) => {
+        this.setState({error: true});
+        console.error(error);
+      });
+  }
+
```
HenriRabalais, line 79: Personally, this seems a bit redundant and I think `fetchData` could just be used again.

HenriRabalais:
In response to the note in the brief summary of changes:

I've found that the most effective way for modules with a small amount of data is to simply re-fetch all the data. It's not that costly, and is much simpler to implement.

I had played around with the idea of adding the record on the front end upon successful save, but I ran into some problems where sometimes before saving to the database, there would be some further alterations done to the record on the backend. Therefore, there would be a mismatch between the front-end record and then one saved to the database. In addition, the code became a bit convoluted and complicated to try and determine the project, site, etc.

The best solution I found is to actually have the post/put request submitted to the back-end return all the records that have been created or updated. That way, it's guaranteed that those records have been created/altered and represent exactly what's in the database. This also avoid a second query to fetch all data again, or the specific records necessary. This is a bit more difficult to implement, but I think is the most streamline and effective solution that I can think of.

maltheism:
@HenriRabalais in response to your comment. There is a big performance hit with the current implementation or the one you're suggesting if the data in the database is actually very large (many rows). That's why I thought just adding one row to the frontend table would be better if not ignoring that performance problem.

modules/media/jsx/mediaIndex.js
maltheism: @HenriRabalais reason I chose not to use fetchData is that the user filters would be reset (I assume) if any user input existed.

HenriRabalais:
@maltheism The 3rd implementation I'm suggesting would have no performance hit since you're just returning the record that was updated from the post/put directly back to the front-end, so there's no need for a new request.

maltheism:
@HenriRabalais that would still be a performance issue if many rows populate the table. Sending a large amount of data would result in a bandwidth problem depending on the user's internet speed. I'm unaware if an Loris install would potentially encounter this problem but its being experienced with the DQT. I thought to write about it because it's something to consider compared to just updating the table with one row of data being sent back from the server to the client.

maltheism:
@HenriRabalais I did this fix because the project & site weren't accessible all the time from the data on the frontend.

HenriRabalais (approved):
Looks good!

PR #7030, opened 2020-09-14
[DQT beta] fix console warnings

maltheism:
## Brief summary of changes

Fixes the warning in browser console:
```
Warning: Encountered two children with the same key, `demographics`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

also

```
react.development.js:1641 Warning: Failed prop type: The prop label is marked as required in RadioElement, but its value is undefined.
in RadioElement (created by ViewDataTabPane)
in ViewDataTabPane (created by DataQueryApp)
```

and lastly solves
```
When clicking Run Query:
react_devtools_backend.js:2273 Warning: Failed prop type: Invalid prop Data of type object supplied to StaticDataTable, expected array.
in StaticDataTable (created by ViewDataTabPane)
in ViewDataTabPane (created by DataQueryApp)
in div (created by StepperPanel)
in StepperPanel (created by DataQueryApp)
in div (created by DataQueryApp)
in div (created by DataQueryApp)
in DataQueryApp
in div (created by StepperPanel)
in StepperPanel (created by DataQueryApp)
in div (created by DataQueryApp)
in div (created by DataQueryApp)
in DataQueryApp
```

### DIFF
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -2233,7 +2233,7 @@ class RadioElement extends React.Component {
 }
 RadioElement.propTypes = {
   name: PropTypes.string.isRequired,
-  label: PropTypes.string.isRequired,
+  label: PropTypes.string,
   options: PropTypes.object.isRequired,
   disabled: PropTypes.bool,
   required: PropTypes.bool,
--- a/modules/dqt/jsx/react.app.js
+++ b/modules/dqt/jsx/react.app.js
@@ -47,7 +47,7 @@ class DataQueryApp extends Component {
         show: false,
       },
       ActiveTab: 'Info',
-      rowData: {},
+      rowData: [],
       filter: {
         type: 'group',
         activeOperator: 0,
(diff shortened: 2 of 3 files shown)


laemtl (approved):

jsx/Form.js
```diff
@@ -2233,7 +2233,7 @@ class RadioElement extends React.Component {
 }
 RadioElement.propTypes = {
   name: PropTypes.string.isRequired,
-  label: PropTypes.string.isRequired,
+  label: PropTypes.string,
```
driusan, line 2236: This looks like it is, in fact, required in the render method.

@HenriRabalais @zaliqarosli @maltheism ? (I'm not sure who wrote this particular component..)

maltheism: @driusan I was the one that made the RadioElement component and the label isn't required for functionality. Specifically, if the developer doesn't supply a "label" the radio options will be visible without a label. The redesigned DQT uses the RadioElement component without a label and for switching between Cross-sectional and Longitudinal. It looks ugly if a label is supplied.

driusan: But the render function unconditionally uses the label prop on line 2223

maltheism: @driusan nothing will be rendered on line 2223 if this.props.label is null

PR #7037, opened 2020-09-15
[media] hidden files do not display for superuser

### LINKED ISSUE #6950: [media] hidden files do not display for superuser
**Describe the bug**
When a media file is edited with the field `Hide File` set to `Yes`, upon module refresh the file entry in the datatable disappears even for a superuser. The testing plan stipulates that the file should not disappear for a superuser.

**To Reproduce**
1. User superuser account
2. Go to Media module
2. Click on 'Edit' for a media file in the table
3. Change value of `Hide File` to `Yes` and click `Update File`
4. Refresh page
5. The file will have disappeared from the datatable.

maltheism:
## Brief summary of changes

Solves the bug found in #6950

1. Hide file as superuser.
2. Refresh module.
3. Hidden file row should still be visible to superuser.
4. Hidden file row should be hidden from non-superusers.

* Resolves #6950

### DIFF
--- a/modules/media/php/media.class.inc
+++ b/modules/media/php/media.class.inc
@@ -155,8 +155,7 @@ class Media extends \DataFrameworkMenu
     public function getDataProvisionerWithFilters() : \LORIS\Data\Provisioner
     {
         $provisioner = parent::getDataProvisionerWithFilters();
-        $factory     = \NDB_Factory::singleton();
-        $user        = $factory->user();
+        $user        = \NDB_Factory::singleton()->user();
         if ($user->hasPermission('superuser',)) {
             // We don't filter out hidden files.
             return $provisioner;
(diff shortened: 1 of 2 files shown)


maltheism:
Hi @kongtiaowang, I'm assuming test that's failing shouldn't be. I'm unsure though. Latest code change in PR makes the correct behaviour for rows being displayed when file is hidden or not with superuser permission.

driusan:
Does the test user have the superuser permission? If not, I don't think it should affect the number of results for the test case..

maltheism:
@driusan that's why I'm confused. I don't believe the test case should change from the original code if the test user doesn't have superuser permission.

HenriRabalais (approved):
looks good to me!

will test when passing travis.

kongtiaowang:
@maltheism There is a test media file with hide_file attributes, it doesn't show in the table before.
After you fixing the filter, this file showes in the table. That is why test result turn 26 to 27. 
The testing user has superuser permisson in this test case. so you need to change the number (26 to 27) in MediaTest::testBrowseFilter.

PR #7041, opened 2020-09-16
[DQT] Performance fix for load query and fix vars exceed on save

maltheism:
## Brief summary of changes

Found by @pierre-p-s when lots of fields are saved. Performance part was found once we got passed that and attempted to load the saved query.

1. Visit DQT
2. Add many fields
3. Save and see if everything is okay.
4. Attempt to load query after browser refresh.


pierre-p-s:
It works. Solves the error: "PHP Warning:  Unknown: Input variables exceeded 1000."

modules/dqt/ajax/datadictionary.php
```diff
@@ -43,18 +43,23 @@
             "endkey"   => "[\"$category\", \"ZZZZZZZZ\"]",
         ]
     );
-} else if (isset($_REQUEST['key']) && $_REQUEST['key']) {
-    $key = explode('%2C', urlencode($_REQUEST['key']));
+} else if (isset($_REQUEST['keys']) && $_REQUEST['keys']) {
+    $keys    = json_decode($_REQUEST['keys']);
+    $results = [];
```
zaliqarosli, line 48: $results is already defined as empty array above
```suggestion

```

modules/dqt/ajax/datadictionary.php
```diff
@@ -43,18 +43,23 @@
...
+        $query = $cdb->queryView(
+            "DQG-2.0",
+            "datadictionary",
+            [
+                "reduce" => "false",
+                "key"    => "[\"$key[0]\",\"$key[1]\"]",
+            ]
+        );
+
+        array_push($results, $query[0]);
```
zaliqarosli, line 61: just wanna check that it should be $query[0] instead or just $query, which was how the results were being used in the deleted code on the left here for the case where there's only 1 key

maltheism: Hi @zaliqarosli, please manually test if you're concerned because that change is from the logic change for fixing the performance problem. No data is lost that I've aware about and the filter & fields will load for the DQT to function from the tests I did.

maltheism: Hi @zaliqarosli & @pierre-p-s, I went and took a few screen shots of before and after to help illustrate what I did in the past. Basically, I cleaned up the data being sent back to the frontend and also note that the frontend was modified to not require many requests but just one. So the $query[0] is part of the improvement.

Before:
before


After:
after

zaliqarosli (approved):

PR #7042, opened 2020-09-17
[eslint] Option to compile without  eslint

laemtl:
This PR add a new command: `npm run compile-production`, to compile jsx files with eslint. `npm run compile` now runs on dev mode and have eslint disabled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,8 +50,9 @@
     "tests:unit:debug": "DEBUG=true ./test/dockerized-unit-tests.sh",
     "tests:integration": "./test/dockerized-integration-tests.sh",
     "tests:integration:debug": "DEBUG=true ./test/dockerized-integration-tests.sh",
-    "compile": "webpack",
-    "watch": "webpack --watch"
+    "compile": "webpack --mode=development",
+    "compile-production": "webpack --mode=production",
+    "watch": "webpack --watch --mode=development"
   },
   "repository": {
     "type": "git",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -17,6 +17,7 @@ const optimization = {
     }),
   ],
 };
+
 const resolve = {
   alias: {
     util: path.resolve(__dirname, './htdocs/js/util'),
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -50,7 +50,8 @@
     "tests:unit:debug": "DEBUG=true ./test/dockerized-unit-tests.sh",
     "tests:integration": "./test/dockerized-integration-tests.sh",
     "tests:integration:debug": "DEBUG=true ./test/dockerized-integration-tests.sh",
-    "compile": "webpack",
+    "compile-noeslint": "webpack --mode=development",
```
maltheism, line 53: @laemtl can we make `compile` be the development mode and `compile-eslint` or `compile-production` be for production?

laemtl: Sure. Travis is running run-js-linter.sh I think, not the npm command, is that right?

maltheism: I'm unsure about Travis but Makefile uses `npm run compile`

maltheism: @kongtiaowang might know.

maltheism: @laemtl also docs/React.README.md mentions npm run compile

laemtl: I checked, Travis uses npm run lint:javascript. So do we need a compile-production? A combo of compile (with eslint off) and npm run lint:javascript can achieve the same result.

maltheism: I just would like to keep `npm run compile` for development since it's simpler to type. I think Makefile is used when setting up production VMs.

maltheism (approved):

driusan (changes requested):
eslint is a development tool. It doesn't make sense to use it on production but not development.

This should be adding something like `make quickdev` to run it without it without touching the current `make dev` target used by Travis and other dev environments.

maltheism:
Hi @laemtl, I thought initially when reviewing that the PR was just to have the mode of either production or development be set correctly for webpack. I agree with @driusan that we shouldn't disable eslint completely for development and have just a quick mode alternative to use if a developer wishes.

PR #7043, opened 2020-09-17
[Login] Password expiry - Password visibility button

### LINKED ISSUE #6961: [login] Password visibility
**Describe the bug**
When creating an account and receiving the generated password by email, after the first login we need to create a new password. We cannot toggle the visibility of the password like when we login. It would be a good feature to add.
Also happens when the user is asked to change password when it expires

**To Reproduce**
1. Request an account
2. On admin account, approve the account and send email with generated password
3. Login with new account and the generated password
4. You should be directed to a page asking for a new password

laemtl:
Add password visibility buttons on the password fields when a user is requested to change his password (account request and password expiry).

Resolves #6961

### DIFF
--- a/modules/login/templates/form_passwordexpiry.tpl
+++ b/modules/login/templates/form_passwordexpiry.tpl
@@ -1,3 +1,7 @@
+
+<script type="text/javascript" src="{$baseurl}/js/passwordVisibility.js"></script>
+<link rel="stylesheet" href="/css/password.css" type="text/css" />
+
 <div class="container">
   <div class="panel panel-default panel-center">
     <div class="panel-heading">
@@ -36,3 +40,9 @@
     </div>
   </div>
 </div>
+
+<script>
+  document.addEventListener('DOMContentLoaded', function(event) {
+    passwordVisibility();
+  })
+</script>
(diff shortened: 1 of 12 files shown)


maltheism (approved):

pierre-p-s (approved):

modules/login/templates/form_passwordexpiry.tpl
```diff
@@ -1,3 +1,7 @@
+
+<script type="text/javascript" src="{$baseurl}/js/passwordVisibility.js"></script>
+<link rel="stylesheet" href="/css/password.css" type="text/css" />
```
driusan, line 3: Is there a reason these can't be added to getJSDependencies and getCSSDependencies of the password expiry page's class? (And similarly, the script tag added to a `.js` file for the module..)

driusan (changes requested):

modules/login/templates/form_passwordexpiry.tpl
laemtl: @driusan Fixed!

christinerogers:
@driusan re-review please, comments seem to be addressed.

PR #7050, opened 2020-09-21
 [Parent Portal] from IBIS

### LINKED ISSUE #7366: Participant Portal
Different version of this already exist on different project but it would be nice to consolidate this feature into the loris core and have all participant-concerns displayed in one participant facing page.

This module can work with the survey module and the e-consent module (#7363) to display tasks the participant needs to do and tasks that are already done

any other participant data can go in there too

kongtiaowang:
## Brief summary of changes

add the parent portal form IBIS

Sep-17-2020 13-19-09

Add a new candidate, then create an instrument for this new user.
Go to survey account module, click the parentID link to load this module.


php/libraries/Candidate.class.inc
```diff
@@ -323,7 +325,23 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
         // Return the new CandID
         return $candID;
     }
-
+    /**
+     * Generate a ParentID for a candidate
```
zaliqarosli, line 329:
```suggestion
     * Generate an ID for a candidate with given type
```

php/libraries/Candidate.class.inc
```diff
@@ -323,7 +325,23 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
...
+     *
+     * @param string $IDtype  likes Parent
```
zaliqarosli, line 331:
```suggestion
     * @param string $IDtype  a string that describes the ID to be generated, and is prepended to 'ID' i.e. "Parent" to generate a ParentID
```

php/libraries/Candidate.class.inc
```diff
@@ -323,7 +325,23 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
...
+     *
+     * @return string $randid likes Parent23456
```
zaliqarosli, line 333:
```suggestion
     * @return string $randid the randomly generated ID with $IDtype prefix i.e. Parent23456
```

php/libraries/Candidate.class.inc
```diff
@@ -323,7 +325,23 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
...
+     * @throws DatabaseException
+     */
+    function _generateID(string $IDtype): string {
```
zaliqarosli, line 336: perhaps a ParentIDGenerator class can be created that extends SiteIDGenerator or IdentifierGenerator class. and then in the createNew() method above, you can call `$parentID = (new \ParentIDGenerator())->generate();`. this would be in line with how CandIDs, PSCIDs and ExternalIDs are currently being generated.

Here's an example of how I did it for QPN (I copied the PSCIDGenerator and ExternalIDGenerator classes): https://github.com/aces/QPN/blob/master/libraries/PatientIDGenerator.php

zaliqarosli (changes requested):
good so far! some minor changes, and one big one (how PatientIDs are generated). I also think that as part of this PR, we should hide the 'Email' column of surveyAccountsIndex but keep it filterable (because emails can potentially be identifying information)

kongtiaowang:
to restart Travis

php/libraries/Candidate.class.inc
```diff
@@ -9,6 +9,7 @@
 use LORIS\StudyEntities\Candidate\CandID;
 use LORIS\StudyEntities\Candidate\EntityType;
 use LORIS\StudyEntities\Candidate\Sex;
+use LORIS\StudyEntities\Candidate\ParentIDGenerator;
```
zaliqarosli, line 12: not sure if the ParentIDGenerator should be classed within StudyEntities. I think it makes sense for ParentID to be a candidate study entity, but not the generator itself

php/libraries/ParentIDGenerator.php
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * PHP Version 7
+ *
+ * @category Main
+ * @package  LORIS
+ * @author   Shen Wang <wangshen.mcin@gmail.com>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+namespace LORIS\StudyEntities\Candidate\ParentIDGenerator;
```
zaliqarosli, line 13: i guess instead of creating this namespace, this ParentIDGenerator class would `use \LORIS\StudyEntities\Candidate\ParentID`

php/libraries/ParentIDGenerator.php
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * PHP Version 7
+ *
+ * @category Main
+ * @package  LORIS
+ * @author   Shen Wang <wangshen.mcin@gmail.com>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+class ParentIDGenerator extends \CandIDGenerator
```
zaliqarosli, line 26: i don't think the ParentIDGenerator should be extending the CandIDGenerator and call its generate() method because that method checks for whether a certain randomly generated CandID already exists in the databases. The ParentID is separate to the CandID and should not be dependent on it at all

php/libraries/ParentIDGenerator.php
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Creates a new ParentID.
+     *
+     * @return string The new identifier.
+     */
+    public function generateParentID(): string
+    {
+        //  $generator = new \CandIDGenerator();
+        $candID = parent::generate();
+        return $this->prefix.(string)$candID;
```
zaliqarosli, line 52: i dont like this ... the ParentID shouldn't have anything to do with the CandID. are you essentially using the CandIDGenerator to utilize the fact that it creates a 6 digit number?

zaliqarosli: i think we should leave the pattern/sequence of the ParentID to be configurable by the project by supplying it in config.xml:

```
        <ParentID>
            <generation>random</generation>
            <structure>
                <seq type="static">PARENT</seq>
                <seq type="numeric" length="6" min='100000' max='999999'/>
            </structure>
        </ParentID>
```

zaliqarosli: doing this will allow you to have a very simple ParentIDGenerator that looks like this:

```
class ParentIDGenerator extends SiteIDGenerator
{
    /**
     * Generates ParentIDs.
     *
     * @param ?string $prefix The site prefix to prepend to the ID value.
     *
     * @return void
     */
    public function __construct(?string $prefix = null)
    {
        $this->kind = 'ParentID';
        parent::__construct($prefix);
    }
}
```

calling ParentIDGenerator->generate() would just call the SiteIDGenerator generate() method which utilizes the IdentifierGenerator abstract class's createNewID() method

zaliqarosli (changes requested):

php/libraries/Candidate.class.inc
zaliqarosli: i think you can create a LORIS\StudyEntities\Candidate\ParentID class that looks exactly like LORIS\StudyEntities\Candidate\CandID, extending the ValidatableIdentifier abstract class, and specififying `return 'ParentID'` within the getType() method

driusan:
I don't understand the reasoning behind reactifying the parent portal.. isn't the parent portal for subjects doing data entry on their own computer?  How do you know they haven't disabled javascript or installed some other plugin that will interfere with this? Plain HTML seems like it would be much safer.

zaliqarosli:
hey @driusan the parent portal isn't actually used for data entry, thats the new survey module. the parent portal is essentially just a table listing the surveys the parents can fill out, with links to the survey. i don't see an issue with this module being reactified, especially given that we're are reactifiying all of loris, i don't see why we would leave this module out. also, the new survey module that IBIS and QPN are using is reactified so the ship has kind of already sailed on that.

I just feel like we may be creating double to work if we review and test this PR as plain html, to only then reactify it in the future. when we can just accept the already reactified code now. what do you think?

kongtiaowang:
rebase

kongtiaowang:
This PR is out of date.

kongtiaowang:
replace by participant portal

Issue #7002, opened 2020-09-07
[Document Repository] Unable to edit an uploaded file 


jftunteng:
**Describe the bug**
Getting an error (screenshot below)  when clicking 'edit' for a file that has already been uploaded. 
![doc repo error ](https://user-images.githubusercontent.com/22081216/92416464-9beb5380-f12b-11ea-8ce9-6213273cbe3d.png)

driusan:
What does the apache error log say?

PR #7056, opened 2020-09-23
[Document Repository]  fix Unable to edit an uploaded file issue #7002

kongtiaowang:
## Brief summary of changes
fixing the file id method and handle put method to edit a file.
upload a file and edit it. 

Sep-23-2020 12-49-34

https://github.com/aces/Loris/issues/7002

### DIFF
--- a/modules/document_repository/php/files.class.inc
+++ b/modules/document_repository/php/files.class.inc
@@ -62,9 +62,7 @@ class Files extends \NDB_Page
         case "POST":
             return $this->uploadDocFile($request);
         case "PUT":
-            $tmp = $request->getBody();
-            $arr = (array)json_decode($tmp);
-            return $this->editDocFile($arr);
+            return $this->editDocFile($request);
         case "DELETE":
             $id = basename($request->getUri()->getPath());
             $this->deleteFile($id);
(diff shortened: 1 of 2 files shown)


laemtl (approved):

modules/document_repository/php/files.class.inc
```diff
@@ -62,17 +62,25 @@ class Files extends \NDB_Page
...
             $this->deleteFile($id);
             return (new \LORIS\Http\Response\JSON\OK(
                 ['message' => 'File deleted.']
             ));
         case "GET":
-            $filename   = urldecode(basename($request->getUri()->getPath()));
+            $filename = urldecode(basename($request->getUri()->getPath()));
+            if (is_numeric($filename)) {
+                return (new \LORIS\Http\Response())
+                    ->withHeader("Content-Type", "text/plain")
```
driusan, line 76: I don't understand what this is trying to do.. what does the filename being numeric have to do with anything? What if it's not a plain text file?

kongtiaowang: @driusan I hope the URL will get two kinds of filenames.  [ id(number only) and filename ]
ex. "/document_repository/files/4" will return all the information of Upload Fields for edit_document. 
ex. "/document_repository/files/test.pdf" will return the PDF file.

driusan: Sorry for the delay in responding, but I don't think that's robust. What if someone uploads a file named "4"?

It would make more sense to have 2 different endpoints like `/document_repository/metadata/id` and `/document_repository/files/name` to handle the 2 different types of data being retrieved (though I don't have any opinions on if the first one should be by name or id.)

driusan (changes requested):

modules/document_repository/php/files.class.inc
kongtiaowang: " /document_repository/Files/meta/id" will return meta data for each file.
" /document_repository/Files/filename" will return the file.

kongtiaowang: I added a new endpoint to handle the meta data.  
" /document_repository/Files/meta/id" will return meta data for each file.
" /document_repository/Files/filename" will return the file.

cmadjar:
@driusan review please :)

PR #7058, opened 2020-09-23
[Candidate] Allow null Sex for scanner candidates

### LINKED ISSUE #6990: [Candidate parameter] Sibling link does not work
**Describe the bug**
Make sure `Use family ID` is set to true in the configuration module. Access the profile of candidate DCC090 and click on `Candidate Info`. Select the family tab and click on the link displayed for the CandID of the first sibling (284119): you get the `This page is not working` page.  This might be an error with the way RB data is set up or a genuine bug. TBD.

CamilleBeau:
## Brief summary of changes

Currently, the default value for "Sex" when a scanner is created is NULL. This causes an error when the candidate profile is visited from the family tab (as described in issue #6990). This PR makes it so that a Sex value is not required when selecting a Candidate to accommodate these Scanner candidates.

1. Make sure that you can visit the candidate profile of a candidate with entity type `scanner`
2. Follow the procedure in the description of https://github.com/aces/loris/issues/6990 and make sure that you do not encounter the error.

* Resolves #6990

### DIFF
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -102,7 +102,7 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
         foreach ($row as $key=>$value) {
             switch ($key){
             case 'Sex':
-                $this->candidateInfo[$key] = new Sex($value);
+                $this->candidateInfo[$key] = empty($value) ? null : new Sex($value);
                 break;
             case 'EntityType':
                 $this->candidateInfo[$key] = new EntityType($value);


maltheism (approved):

driusan:
I don't think this is supposed to be the case. The options for Sex in the candidate table are "enum('Male','Female','Other')". Aren't scanners supposed to be 'Other'? (@cmadjar @nicolasbrossard ?)

cmadjar:
Right. I believe Sex = Other is (or should be) used for Scanners.

nicolasbrossard:
@driusan @cmadjar I think `Other` is misleading in the sense that it suggests that the entity is human and has a sex but it's neither `Male` nor `Female`. If `Other` is strictly reserved for scanners, then why not label it something like `N/A (scanner)`?

driusan:
My understanding is that the options map directly from the (required) DICOM PatientSex column where scanners are classified as "O"(ther).

nicolasbrossard:
@driusan Right. Maybe change the label (not the enum value itself) to 'Other (scanner)'? Just to emphasize that when you select 'Other' you are not going to get any humans.

CamilleBeau:
Are we decided on changing the "other" label for sex to "Other (scanner)" for scanners and continuing to not allow null sex?

driusan:
No, I don't think "Other (scanner)" makes sense. "Other" does not mean it's a scanner, it means it's not Male or Female. (which happens to include scanners.)

CamilleBeau:
> No, I don't think "Other (scanner)" makes sense. "Other" does not mean it's a scanner, it means it's not Male or Female. (which happens to include scanners.)

@driusan So is the action plan be to leave it as other, and change the sex of Scanners to "Other" instead of null so that it can still be mapped from DICOM PatientSex column?

driusan:
@CamilleBeau I don't understand the question. Is that already the behaviour that this PR is trying to change it from?

Sorry for the delay in responding.

CamilleBeau:
> @CamilleBeau I don't understand the question. Isn't that already the behaviour that this PR is trying to change it from?

No this PR is working off of scanners having "null" as sex. I'm just a bit unclear of where we want to go with this based on the conversation in this PR. Do we want scanners to have "null", "other", or a unique scanner sex?

driusan:
I think "Other", but maybe it's worth holding off until an imaging meeting.

CamilleBeau:
> I think "Other", but maybe it's worth holding off until an imaging meeting.

Ok that sounds good! I'll check in with this soon then.

CamilleBeau:
@driusan Has an imaging meeting happened yet to give more insight on this?

driusan:
As far as I know there hasn't been one yet.

cmadjar:
Will be discussed on Friday. @CamilleBeau let me know if you want to join regarding this bit.

PR #7070, opened 2020-09-28
[API] Adding files in the docker container for API integration tests 

spell00:
## Brief summary of changes

The integration test for the GET request to download an image had to be skipped because the docker container don't have the data in `/data-raisinbread`.

The data is not available during Travis checks, so empty files are added and, during Travis checks, are moved to where the data should be to run the integration tests that assert the files can be downloaded with the API.

* Resolves #6907


spell00:
The changes I propose in this PR don't work with Travis because the raisinbread data is not included. The only solution I could see would be to have the data for these endpoints stored somewhere where they can be downloaded. The instructions to download them could be encrypted as here: https://docs.travis-ci.com/user/encrypting-files/ and added in the git repo. It would then be decrypted when Travis is run, then the instructions it contains would be used to download the secured data.

docker-compose.yml
```diff
@@ -8,6 +8,7 @@ services:
         BASE_DIR: /app/
     volumes:
       - ./test/mysql-config:/etc/mysql/conf.d
+      - ./raisinbread/test/data-raisinbread/:/data-raisinbread
```
kongtiaowang, line 11: this line is not necessary.

docker-compose.yml
```diff
@@ -16,6 +17,7 @@ services:
     image: selenium/standalone-firefox-debug:3.8.1-aluminum
     volumes:
       - /dev/shm:/dev/shm
+      - ./raisinbread/test/data-raisinbread/:/data-raisinbread
```
kongtiaowang, line 20: This line is not necessary.

docker-compose.yml
```diff
@@ -53,6 +56,7 @@ services:
       dockerfile: Dockerfile.test.php7
     volumes:
       - ./:/app
+      - ./raisinbread/test/data-raisinbread/:/data-raisinbread
```
kongtiaowang, line 59: This line is not necessary.

kongtiaowang (changes requested):

test/unittests/BreadcrumbTrailTest.php
```diff
@@ -44,21 +44,35 @@ protected function setUp(): void
     /**
      * Test __toString() returns correct string
      * TODO: Add potential edge cases (such as white space)
+<<<<<<< 334ece634082b9b39992f74814a5fb21fa9ecf51
```
kongtiaowang, line 47: These files have confilcts, you need to fix them.

kongtiaowang (changes requested):

docker-compose.yml
spell00: I did try to remove it (and the others you marked as not necessary), but then it didn't work anymore. Unfortunately, I am unable to test this change right now. As I think you may have noticed, there is a problem with `composer` packages that makes Travis unable to run checks (ends up being errored). This is also causing errors when I try to run tests today...

I tried to add the ` composer/package-versions-deprecated` package as suggested and it did solve the composer packages problem, but it shows a lot of new valid formatting problems that were not detected before.

kongtiaowang: I will fix the composer issue. Then you can test it again.

spell00: I was able to run the tests and it looks like you were right, it works without the 3 lines you said were not necessary. I removed them.

docker-compose.yml
spell00: I removed it

docker-compose.yml
spell00: It is removed

test/unittests/BreadcrumbTrailTest.php
spell00: it is resolved.

christinerogers:
@kongtiaowang it looks like @spell00 has addressed your comments.  
Could you re-review when you have a moment friday?

kongtiaowang (approved):
LGTM , This PR will be rebase after fixing Travis.
https://github.com/aces/Loris/pull/7101

christinerogers:
@driusan requesting final review - this is our last stretch in which @spell00 can readily make changes.  thank you!

spell00:
blocked by #7101

spell00:
@driusan @xlecours This is now passing Travis checks and is ready for final review

driusan (approved):
Since they're empty I don't see why these files are added to raisinbread instead of just created by the test suite's setup if they don't exist, but I don't think it's a blocker so I'll merge it anyways.

PR #7081, opened 2020-10-02
[API] Changing Candidate Visit view to allow v0.0.2

### LINKED ISSUE #6058: [API] Can not access visit through v0.0.2 of API
When attempting to access the URL http://localhost:8000/api/v0.0.2/candidates/965327/V06 on my sandbox I get an error "Unsupported Version".

The URL  http://localhost:8000/api/v0.0.3-dev/candidates/965327/V06 works properly.

However, the documentation for v0.0.2 says that the endpoint should be available in version 2 of the API, so both the released and development version of the API should be equivalent.

spell00:
## Brief summary of changes

Change to the Candidate Visit endpoint to allow the methods for version 0.0.2 of the API, which did not have projects.

#### Testing instructions

1. Go to <HOST>/api/v0.0.3/candidates/300001/V1

2. Go to <HOST>/api/v0.0.2/candidates/300001/V1. The page should display the same information, except `v0.0.2` should not contain the field `Project`

#### Link to the related issue

* Resolves #6058

### DIFF
--- a/modules/api/php/endpoints/candidate/visit/visit.class.inc
+++ b/modules/api/php/endpoints/candidate/visit/visit.class.inc
@@ -75,8 +75,10 @@ class Visit extends Endpoint implements \LORIS\Middleware\ETagCalculator
      */
     protected function supportedVersions() : array
     {
-        // Removed 0.0.2 since session requires a project.
-        return ['v0.0.3'];
+        return [
+            'v0.0.2',
+            'v0.0.3'
+        ];
     }
 
     /**
@@ -117,6 +119,7 @@ class Visit extends Endpoint implements \LORIS\Middleware\ETagCalculator
         $subendpoint = array_shift($pathparts);
         switch($subendpoint) {
         case 'instruments':
+            $version = $request->getAttribute('LORIS-API-Version');
             $handler = new Instruments($this->_visit);
             break;
         case 'images':
--- a/modules/api/php/endpoints/candidate/visit/visit.class.inc
+++ b/modules/api/php/endpoints/candidate/visit/visit.class.inc
@@ -91,10 +91,11 @@ class Visit extends Endpoint implements \LORIS\Middleware\ETagCalculator
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
         $pathparts = $request->getAttribute('pathparts');
+        $version   = $request->getAttribute('LORIS-API-Version');
         if (count($pathparts) === 0) {
             switch ($request->getMethod()) {
             case 'GET':
-                return $this->_handleGET();
+                return $this->_handleGET($version);
 
             case 'PUT':
                 return $this->_handlePUT($request);
@@ -119,8 +120,7 @@ class Visit extends Endpoint implements \LORIS\Middleware\ETagCalculator
         $subendpoint = array_shift($pathparts);
         switch($subendpoint) {
         case 'instruments':
-            $version = $request->getAttribute('LORIS-API-Version');
-            $handler = new Instruments($this->_visit, $version);
+            $handler = new Instruments($this->_visit);
             break;
         case 'images':
             $handler = new Images($this->_visit);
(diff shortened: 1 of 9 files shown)


modules/api/php/endpoints/candidate/visit/visit.class.inc
```diff
@@ -89,10 +91,11 @@ class Visit extends Endpoint implements \LORIS\Middleware\ETagCalculator
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
         $pathparts = $request->getAttribute('pathparts');
+        $version   = $request->getAttribute('LORIS-API-Version');
```
xlecours, line 94: I would prefer to have it done inside handleGET so that both handleGET and handlePUT have a similar signature.

modules/api/php/endpoints/candidate/visit/visit.class.inc
```diff
@@ -89,10 +91,11 @@ class Visit extends Endpoint implements \LORIS\Middleware\ETagCalculator
...
         if (count($pathparts) === 0) {
             switch ($request->getMethod()) {
             case 'GET':
-                return $this->_handleGET();
+                return $this->_handleGET($version);
```
xlecours, line 98:
```suggestion
                return $this->_handleGET($request);
```

modules/api/php/endpoints/candidate/visit/visit.class.inc
```diff
@@ -147,16 +150,18 @@ class Visit extends Endpoint implements \LORIS\Middleware\ETagCalculator
...
-    private function _handleGET(): ResponseInterface
+    private function _handleGET(string $version): ResponseInterface
     {
         if ($this->_visit === null) {
             return new \LORIS\Http\Response\JSON\NotFound('Visit not found');
         }
 
         return new \LORIS\Http\Response\JsonResponse(
-            (new \LORIS\api\Views\Visit($this->_visit))
+            (new \LORIS\api\Views\Visit($this->_visit, $version))
```
xlecours, line 164: This seems to be a good use case for polymorphism. 
I think `\LORIS\api\Views\Visit_0_0_2` would do the trick.

xlecours (changes requested):
Here is some feedback.
Also, will handlePUT be able to handle v0.0.2 requests?

modules/api/php/endpoints/candidate/visit/visit.class.inc
spell00: I'm not sure of what you are suggesting, you mean we could create a new Class `\LORIS\api\Views\Visit` that would take `$version` as a parameter? If it is polymorphism, shouldn't it also be called `\LORIS\api\Views\Visit` instead of `\LORIS\api\Views\Visit_0_0_2`? Plus I think this solution (with the modifications of `modules/api/php/views/visit.class.inc` works, is there a reason why creating a new polymorphic class would be better than modifying the current `\LORIS\api\Views\Visit` class?

xlecours: > you mean we could create a new Class \LORIS\api\Views\Visit that would take $version as a parameter?  
 
No, I mean that depending on `$version` value, the JsonResponse would take a `\LORIS\api\Views\Visit` or a `\LORIS\api\Views\Visit_0_0_2` object as a parameter. 
With PHP, we can't have 2 classes with the same name in the same namespace. 
There are many ways we could solve that. Factory pattern seems overkill here so just having  `switch` on the version should be enough.

> is there a reason why creating a new polymorphic class would be better than modifying the current \LORIS\api\Views\Visit class?

Modifying the existing Visit class adds to the number of things that class needs to handle.

spell00: OK, I think I understand the advantage of doing it the way you are suggesting.

I thought it was unnecessary and redundant to create a whole new class that is almost identical except for a small difference. 

However, if I understand correctly, when the versions will start accumulating, adding parameters and exceptions to a single class to deal with all the different versions is likely to become more cumbersome and convoluted (in this case by adding the element `Project` when it's not v0.0.2).

xlecours: This is exacly what I am foreseeing.
Since this seems to be the first of probably many divergence in supported versions, maybe @driusan can add his 2 cents?

spell00:
@xlecours if this GET request was invalid in v0.0.2 because there was no `Project` field, would that also be true for every endpoint that required the field `Project`? For example, should I create a `modules/api/php/views/candidate_0_0_2.class.inc` and do exactly like for the Visits? The endpoint `/api/{version}/projects` allows `v0.0.2`, but shouldn't it be unavailable if there was no `Projects`?

xlecours:
> @xlecours if this GET request was invalid in v0.0.2 because there was no `Project` field, would that also be true for every endpoint that required the field `Project`? For example, should I create a `modules/api/php/views/candidate_0_0_2.class.inc` and do exactly like for the Visits? The endpoint `/api/{version}/projects` allows `v0.0.2`, but shouldn't it be unavailable if there was no `Projects`?
> 
> I already implemented the change, but maybe I didn't understand why the field `Project` had to be removed in `v0.0.2` in the Candidate Visits?

Let's ask @driusan 
I think v0.0.2 should not be supported anymore.

spell00:
I'm closing this PR, v0.0.2 will no longer be supported

PR #7086, opened 2020-10-07
Improved consistency of RB data for table mri_upload.

nicolasbrossard:
## Brief summary of changes

Given that:

1. There is currently only one user in raisin bread: `admin`.
2. The MRI incoming directory is set to `/data/incoming`.
3. When an MRI archive is successfully uploaded, it goes to `<MRIIncomingDir>/<user>` (where user is the ID of the user who uploaded the file).

I think it makes more sense to have all records in table `mri_upload` have an `UploadLocation` with a root directory set to `/data/incoming/admin`. This PR does just that.

Make sure that the RB data in table `mri_upload` loads correctly and that  all records in table `mri_upload` have an `UploadLocation` with a root directory set to `/data/incoming/admin`.


cmadjar (approved):
Did not test but the SQL changes make sense.

PR #7090, opened 2020-10-08
[API] New api candidates endpoint post tests

spell00:
## Brief summary of changes

Adds two new tests for POST requests to the \Candidates endpoint. 

1- A valid POST request to create a new candidate at a site where the user is not affiliated with. It expects a Forbidden HTTP error (403). 
2- An invalid POST request where the project is an empty string. It has to be an existing project, so it fails with NotFound error (400). 

* Resolves #


kongtiaowang (approved):
LGTM

driusan:
Might be a good idea to also check the backend instead of just trusting the response code, but this is still a step forward, so I'll merge it as is..

PR #7094, opened 2020-10-13
[DQT] fixes weird bug found by Pierre

maltheism:
## Brief summary of changes

Solves a weird bug found by Pierre

1. Delete first filter rule in DQT (redesigned version).
2. Add Group rule.
3. Test other buttons for rules.


modules/dqt/jsx/react.filterBuilder.js
```diff
@@ -592,6 +592,7 @@ class FilterGroup extends Component {
             <FilterGroup group={child}
                          items={this.props.items}
                          index={index}
+                         updateFilter={this.updateChild}
```
maltheism, line 595: This line fixes the bug. The other code change is just cleanup I saw when debugging.

driusan: What is "the bug"?

maltheism: @driusan The bug occurs when deleting the first rule of the filter, then making a Group rule and attempting to use the buttons (Add Rule, Add Group, Delete Group) of the newly created Group Rule.
Console output of error:
`TypeError: this.props.updateFilter is not a function. (In 'this.props.updateFilter(group)', 'this.props.updateFilter' is undefined)`

maltheism: @driusan also should note that @pierre-p-s found that current fix solves the buttons from not working but the query results strangely aren't correct. I believe the query returns the correct results for the built filter when the query is saved and loaded. So I'm adding needs work tag until I can debug that part.

pierre-p-s:
So when saving a query to check the filter tree I am getting this result:

Screen Shot 2020-10-14 at 3 35 50 PM

For this query: 

Screen Shot 2020-10-14 at 3 36 29 PM

Also the query does not work when you create it and run it but works when you save it, load it and then run it.

laemtl:
Does this still need work?

maltheism:
Hi @laemtl, yep I haven't had time to work on it.

cmadjar:
@maltheism when you come back, can you work on this? Thank you!

maltheism:
Hi @pierre-p-s, @laemtl, can you both test the change I made and see if solves the functionality while the query results are still accurate? The interface will update if the user does this order: delete all rules, Add Group, set the rule of the Group and you will notice what I mean by update. I think this is a good improvement for correcting the functionality even with the slight update to the user interface for that order of steps.

maltheism:
Hi @pierre-p-s, @laemtl, please test the latest change I made. I believe it solves the functionality of the user interface and the query results should be correct. Note: If the user deletes all rules, Adds Group, then when setting "the rule of the Group" the interface updates where it becomes the baseline Group rule and I'm doubtful I can really prevent that from happening. The important thing is that the query seems to be functioning correctly. I'm open to any suggestions as well.

cmadjar:
@pierre-p-s can you test and review?

pierre-p-s:
Im still getting an error when following the test instructions

Screen Shot 2021-07-19 at 2 36 44 PM

pierre-p-s:
Please disregard earlier comment I was testing on old dqt, I deleted the comment.

When following testing instructions, when I delete the first rule and then add a group, if I select an instrument in the filters, the group disapears and becomes a normal rule. Just checking that this is the intended behaviour.

Otherwise seems to be working.

maltheism:
@pierre-p-s yep see my earlier comment. Latest changes fix the logic/functionality but the UI does do that.

pierre-p-s (approved):

cmadjar:
@driusan ready for you

maltheism:
Hi @xlecours, I just tested this PR again and it seems to be doing what I remember after the code changes I added. What behaviour are you witnessing that makes it not fixed?

Btw, there's one situation that can look weird to the user but still has the correct logic of the filter. That's when user deletes first filter rule, then creates group, then creates rule for the group and they will become two rules grouped for the root. Similar if you do what's in Pierre's comment mentioned but neither cases break the logic of the filter being created. Note it doesn't happen if other rules exist. I also don't think it can be avoided with how the logic of the making of the filters works.

PR #7095, opened 2020-10-14
[Tools] Generate Candidate External IDs

laemtl:
New tool to fill external IDs for all candidates where a NULL value is found.

### DIFF
--- a/tools/generate_candidate_externalids.php
+++ b/tools/generate_candidate_externalids.php
@@ -5,6 +5,19 @@
  * where external ID is NULL.
  * Config useExternalID property must be set.
  *
+ * Take an optional argument to reset all the externalID to NULL
+ * before running the algorithm
+ *
+ * Ways to use this script:
+ * -- display a help screen
+ * generate_candidate_externalids.php help
+ *
+ * -- to fix missing external IDS on candidates
+ * generate_candidate_externalids.php
+ *
+ * -- to reset all candidates external IDs and regenerate them
+ * generate_candidate_externalids.php reset
+ *
  * PHP Version 7
  *
  * @category Main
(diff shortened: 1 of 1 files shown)


driusan:
Is there a reason this script is needed?

laemtl:
This tool was created for another project to populate missing candidates External IDs. Submitting in case it can be useful to add such tool here too.

tools/generate_candidate_externalids.php
```diff
@@ -0,0 +1,101 @@
...
+    "SELECT CandID, ExternalID, RegistrationCenterID as site,
+    RegistrationProjectID as project
+    FROM candidate",
+    []
+);
+
+$i = 0;
+// Update all candidates with an external ID
+foreach ($cands as $cand) {
+    if (!$reset && $cand['ExternalID'] != null) {
```
ridz1208, line 77: you are using a loose operator here `!=` instead of `!==` which means that this statement will return true for `$cand['ExternalID'] = false`, `$cand['ExternalID'] = 0`, `$cand['ExternalID'] = ""` or `$cand['ExternalID'] = NULL`. if this is intentional, I usually recommend using `empty()` to be more clear or making sure to add a specific comment justifying the `!=`.  If this is accidental, please use `is_null` instead


https://www.php.net/manual/en/types.comparisons.php

ridz1208 (changes requested):

tools/generate_candidate_externalids.php
laemtl: Thanks, it's a typo.

laemtl:
Thanks @ridz1208, comments addressed.

ridz1208 (approved):

PR #7097, opened 2020-10-19
New tests for of POST requests to the \Candidates endpoints

spell00:
## Brief summary of changes

New tests to assert that invalid requests get rejected as expected.


driusan:
What's the difference between this and #7090?

spell00:
I added a new test. It needs work though, the first part should not show modifications... I also have other tests I wanted to add, I should have labeled this PR as a draft from the start.

PR #7102, opened 2020-10-30
Session Current_stage - Not Started as default value

### LINKED ISSUE #6874: Missing NOT NULL constraint on session.Current_stage
[ TimePoint.class.inc::getCurrentStage()](https://github.com/aces/Loris/blob/main/php/libraries/TimePoint.class.inc#L622) is expected to return a string. As a result, when timePointInfo["Current_stage"] is NULL, a 500 error is thrown (Candidate detail page with a session having Current_stage = NULL).

 To avoid such scenario, session.Current_stage in the DB should be updated to have the NOT NULL constraint.

laemtl:
Update Current_stage column definition in session:
- Set default value of Current_stage to 'Not Started'
- Set Current_stage not nullable

Resolves #6874


cmadjar (approved):
LGTM

PR #7105, opened 2020-10-30
Update NodeJS version

### LINKED ISSUE #7067: [Install] Update Node 8 (deprecated) for Node 10 (12 recommended)
In the install guides, we should update the version of Node to adjust to the current release.
Node 8 is not supported anymore. The deprecation warning suggests to use v12 (recommended), or v10.

laemtl:
* Resolves #7067

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Deploy and log in with username *admin* and the password that's set up during de
  * MySQL >= 5.7 (or MariaDB >= 10.3)
  * PHP <b>7.3</b> or higher
  * [Composer](https://getcomposer.org/) <b>1.4</b> or higher
- * NodeJS <b>8.0</b> or higher
+ * NodeJS <b>10.0</b> or higher
  * NPM
  * make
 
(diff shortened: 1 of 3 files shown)


driusan:
This is failing Travis

driusan:
Also it doesn't seem to update the NodeJS version, just the documentation?

laemtl:
Yes, It does only update the documentation. What about something like this:
* NodeJS 8.0 or higher  (10.0 or higher recommended)

driusan:
The documentation aren't suggestions. They're under the heading minimum requirements in each of the documents.

driusan:
I just tried doing a `make dev` on the master branch and got a bunch of errors like:

```
npm install
npm WARN notsup Unsupported engine for copy-webpack-plugin@6.3.2: wanted: {"node":">= 10.13.0"} (current: {"node":"8.4.0","npm":"6.13.4"})
```
so the node version *does* need to be updated in the documentation now, but it looks like the minimum is 10.13, not 10.0, can you update this PR?

laemtl:
@driusan Done

driusan (approved):

PR #7110, opened 2020-11-02
[jsx/MultiSelectDropdown.js] Close multiSelectDropdown when user clicks outside the dropdown

maltheism:
## Brief summary of changes

When clicking outside the multiSelectDropdown the dropdown should close. This PR adds that functionality.

1. Visit DQT.
2. Define Fields
3. Click on multiselectdropdown "None selected"
4. Click outside dropdown and it should close.


laemtl (approved):

maltheism:
@driusan ready

jsx/MultiSelectDropdown.js
```diff
@@ -287,24 +297,38 @@ class SelectDropdown extends Component {
...
-          </div>
-          <div className="pull-right">
-            <span className="glyphicon glyphicon-menu-down"></span>
-          </div>
-        </button>
-        <ul className="dropdown-menu">
-          {options}
-        </ul>
-      </div>
+      <>
```
driusan, line 312: I don't see any reason to use a fragment here. It should just be contained in a real element.

driusan (changes requested):

jsx/MultiSelectDropdown.js
maltheism: @driusan the {overlay} on line 330 is a separate div than the div on line 313 and so the fragment is needed.

cmadjar:
@driusan ready for you

PR #7114, opened 2020-11-05
[DQT] Longitudinal bug when multiple spaces exist in visit label

maltheism:
## Brief summary of changes

Solves Longitudinal bug when multiple spaces exist in visit label.

Also solves text overflow from fields. Thanks @zaliqarosli for that fix.

1. Visit DQT
2. Make a query that has spaces in visit label for query.
3. Visit the Longitudinal after query is run.
4. Rows should be populated with the data.

### DIFF
--- a/modules/dqt/jsx/react.app.js
+++ b/modules/dqt/jsx/react.app.js
@@ -886,13 +886,21 @@ class DataQueryApp extends Component {
             if (RowHeaders.hasOwnProperty(colHeader)) {
               temp = Identifiers[identifier]
                 + ','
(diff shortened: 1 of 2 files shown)


maltheism:
@ridz1208 maybe you can review since you discovered this bug

driusan:
@maltheism can you rebase so that the php8 tests runs?

kongtiaowang (approved):
LGTM

driusan:
What's the "bug" that this is fixing? What was previously happening?

ridz1208:
Data doesnt load in the DQT results page when a visit label contains a space in the label.

PR #7115, opened 2020-11-05
[DQT] solves saved queries having wrong operator in list.

maltheism:
## Brief summary of changes

Solves saved queries having wrong operator in list.

1. Visit DQT
2. Look at saved queries
3. The operator should be correct


laemtl (changes requested):
Dismissed.

laemtl (approved):

laemtl:
Works great. I noticed the same issue in Define filters (operator is always AND). (#7287)

maltheism:
@laemtl oh I think that was fixed on cbigr but I forgot to make a PR here for it.

PR #7117, opened 2020-11-06
Add AccessibleResource interface to TimePoint and cleanup Candidate access logic

driusan:
The candidate access logic was treating sites and projects
inconsistently in terms of whether or not you needed access
to the Registration or any TimePoint's. This implements the
AccessibleResource on a TimePoint and uses it to clean up the
logic of the Candidate's isAccessibleBy to be closer to the
intended behaviour.

Time timepoint_list now filters out any timepoints that the
user who's logged in doesn't have access to and permission
is verified on accessing a timepoint from the API.

### DIFF
--- a/modules/timepoint_list/php/timepoint_list.class.inc
+++ b/modules/timepoint_list/php/timepoint_list.class.inc
@@ -98,7 +98,7 @@ class Timepoint_List extends \NDB_Menu
             foreach ($listOfTimePoints as $currentTimePoint) {
                 $sessionID = new \SessionID($currentTimePoint);
                 $timePoint = \TimePoint::singleton($sessionID);
-                if(!$timePoint->isAccessibleBy($user) {
+                if(!$timePoint->isAccessibleBy($user)) {
                     continue;
                 }
 
--- a/modules/timepoint_list/php/timepoint_list.class.inc
+++ b/modules/timepoint_list/php/timepoint_list.class.inc
@@ -174,7 +174,7 @@ class Timepoint_List extends \NDB_Menu
 
                 // get the outcome data
                 $outcomeStage
-                    = $this->_determinePreviousStage($currentTimePoint);
+                    = $this->_determinePreviousStage($timePoint);
 
                 $getStatusMethod = 'get'.$outcomeStage.'Status';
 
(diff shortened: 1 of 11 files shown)


modules/timepoint_list/php/timepoint_list.class.inc
```diff
@@ -98,6 +98,9 @@ class Timepoint_List extends \NDB_Menu
             foreach ($listOfTimePoints as $currentTimePoint) {
                 $sessionID = new \SessionID($currentTimePoint);
                 $timePoint = \TimePoint::singleton($sessionID);
+                if(!$timePoint->isAccessibleBy($user)) {
```
xlecours, line 101: shouldn't `$listOfTimePoints` be filtered before the `if` at line 85?

driusan: I didn't notice that.. the only thing that `if` statement seems to do is avoid the overhead of `user     = \User::singleton();` which needs to be done to call isAccessibleBy, but good point. I'll clean it up anyways.

driusan: Okay, I cleaned it up and made `getListOfTimePoints` return an array instead of ?array. (I don't see why we had 2 ways to represent the same thing in the first place, and it's possible that the if empty was there to avoid an error when it returns null..). The diff is a little heavier now, but it's easier to see the actual changes with `?w=1`

xlecours: hopefully,  no one was doing isnull or === null on the returned value of getListOfTimePoints

xlecours (dismissed):
LGTM

modules/timepoint_list/php/timepoint_list.class.inc
driusan: If they were they already had a bug if the function returned `[]` since the function doesn't say what the difference is between null and the empty array..

driusan:
This needs the tests to be updated because they're failing due to the behaviour of permissions having changed and the user not having access.

driusan:
@xlecours I finally got the tests to pass.. the API tests were making a lot of assumptions about everyone being able to access every site's visit. Can you rereview?

driusan:
@xlecours can you review this? I don't think it's changed other than the API tests since you approved it.

xlecours (approved):
It looks all right. 
I wish getListOfTimepoints would return a list of Timepoint but I understand the implication of changing that.

driusan:
That would be nice but that's a different refactoring..

Issue #7096, opened 2020-10-15
Install issues on 23.0.2


ridz1208:
1. installdb.php could not write `config.xml` file to the `project/` directory. (the install.sh should make that directory writeable by default for apache)
2. typo `administrater` -> `administrator`
3. when config.xml is generated, the database credentials for both mysql admin and mysql user are the mysql user credentials, even though both the admin and user creds are requested in the installer.

ridz1208:
these were encoutered when i was trying to install cbigr open VM, i have not taken the time to verify if these issues were tackled in the latest branch.

ridz1208:
solutions:
1. copied the output the script generated and created the config.xml myself. then proceeded to change the group ownership on the project directory to apache to allow apache to be able to write in the directory
2. just a typo
3. manually changed values in config.xml for admin user

christinerogers:
needs a priority look for the next 23 minor release. 
the key point is #1 as identified by Rida.  
* Can we do anything in the script to ensure that project/ is writable by apache? 
* Ideally the install.sh script will check at the end of its run whether project/ is writable by apache and if not, ask them to try setting this before they go to installdb.php.

kongtiaowang:
1. If the user has appropriate permissions, the installer will create a writable folder for apache. If not, the installer will notice the user to create the config file manually. 
2. typo- fixed
3. https://github.com/aces/Loris/pull/7118

christinerogers:
I don't think this is resolved yet. 
Updating with [Shen/Dave's comments from closed #7118](https://github.com/aces/Loris/pull/7118#issuecomment-772751356)
> as @driusan said, The install process explicitly says that it will not save the admin username/password. We can manually save the admin user in the config file. 

If I'm reading this right - This issue should either be resolved in the install scripts or a new documentation issue should be opened ?  -- to let users know they'll need to populate config.xml manually.

PR #7118, opened 2020-11-07
[Installer] can't save db admin user into confilg.xml

kongtiaowang:
## Brief summary of changes
save db admin user into confilg.xml.

https://github.com/aces/Loris/issues/7096

### DIFF
--- a/docs/config/config.xml
+++ b/docs/config/config.xml
@@ -20,8 +20,8 @@
         <username>%USERNAME%</username>
         <password>%PASSWORD%</password>
         <database>%DATABASE%</database>
-        <adminUser>%USERNAME%</adminUser>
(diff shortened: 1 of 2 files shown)


kongtiaowang:
restart git actions

driusan:
Shouldn't this be going to the 23.0-release branch if it's tagged 23.0-testing?

driusan:
The template `smarty/templates/install.tpl` needs to be changed if we save this value. The install process explicitly says that it will not save the admin username/password.

driusan (changes requested):
see previous comment

kongtiaowang:
as @driusan said, The install process explicitly says that it will not save the admin username/password. We can manually save the admin user in the config file. I will close this PR.

PR #7122, opened 2020-11-12
Add ProjectID class to LORIS

driusan:
This adds a ProjectID class to LORIS so that the typing
system can catch errors where the wrong type of integer
is passed.

I updated Project->getId and then updated the callsites
which were required to get `make checkstatic` to pass.
It ended up being a more significant change than expected,
but that's probably for the best.

### DIFF
--- a/modules/api/php/models/candidatesrow.class.inc
+++ b/modules/api/php/models/candidatesrow.class.inc
@@ -41,13 +41,15 @@ class CandidatesRow implements \LORIS\Data\DataInstance
     {
         $this->_candid      = $row['CandID'] ?? null;
         $this->_projectname = $row['ProjectName'] ?? null;
-        $this->_projectid   = $row['ProjectID'] ?? null;
-        $this->_pscid       = $row['PSCID'] ?? null;
-        $this->_sitename    = $row['SiteName'] ?? null;
-        $this->_edc         = $row['EDC'] ?? null;
-        $this->_dob         = $row['DoB'] ?? null;
-        $this->_sex         = $row['Sex'] ?? null;
-        $this->_centerid    = $row['CenterID'] ?? null;
+        if ($row['ProjectID'] !== null) {
+            $this->_projectid = new \ProjectID($row['ProjectID']);
+        }
+        $this->_pscid    = $row['PSCID'] ?? null;
+        $this->_sitename = $row['SiteName'] ?? null;
+        $this->_edc      = $row['EDC'] ?? null;
+        $this->_dob      = $row['DoB'] ?? null;
+        $this->_sex      = $row['Sex'] ?? null;
+        $this->_centerid = $row['CenterID'] ?? null;
     }
 
     /**
@@ -89,13 +91,13 @@ class CandidatesRow implements \LORIS\Data\DataInstance
      * Returns the ProjectID for this row, for filters such as
      * \LORIS\Data\Filters\UserProjectMatch to match against.
      *
-     * @return integer ProjectID
+     * @return \ProjectID
      */
-    public function getProjectID(): int
+    public function getProjectID(): \ProjectID
     {
         if ($this->_projectid === null) {
             throw new \Exception('ProjectID is null');
         }
-        return intval($this->_projectid);
+        return $this->_projectid;
     }
 }
--- a/modules/create_timepoint/php/timepoint.class.inc
+++ b/modules/create_timepoint/php/timepoint.class.inc
@@ -193,7 +193,7 @@ class Timepoint extends \NDB_Page implements ETagCalculator
             // if there is only one project, autoselect first project from array of 1
             $project = \Project::getProjectFromID(array_pop($user_list_of_projects));
         } else if (count($user_list_of_projects) > 1) {
-            $project_id = intval($values['project']);
+            $project_id = new ProjectID($values['project']);
             $project    = \Project::getProjectFromID($project_id);
         }
 
--- a/modules/create_timepoint/php/timepoint.class.inc
+++ b/modules/create_timepoint/php/timepoint.class.inc
@@ -130,7 +130,7 @@ class Timepoint extends \NDB_Page implements ETagCalculator
         $projectOptions        = [];
         foreach ($user_list_of_projects as $projectID) {
             $project = \Project::getProjectFromID($projectID);
-            $projectOptions[$projectID] = $project->getName();
+            $projectOptions["$projectID"] = $project->getName();
         }
         $values['project'] = $projectOptions;
 
--- a/modules/create_timepoint/php/timepoint.class.inc
+++ b/modules/create_timepoint/php/timepoint.class.inc
@@ -193,7 +193,7 @@ class Timepoint extends \NDB_Page implements ETagCalculator
             // if there is only one project, autoselect first project from array of 1
             $project = \Project::getProjectFromID(array_pop($user_list_of_projects));
         } else if (count($user_list_of_projects) > 1) {
-            $project_id = new ProjectID($values['project']);
+            $project_id = new \ProjectID($values['project']);
             $project    = \Project::getProjectFromID($project_id);
         }
 
--- a/modules/statistics/php/stats_demographic.class.inc
+++ b/modules/statistics/php/stats_demographic.class.inc
@@ -255,7 +255,9 @@ class Stats_Demographic extends \NDB_Form
         $subprojects     = \Utility::getSubprojectList();
         // Only search for subprojects if $demographicProject is a positive int.
         if (is_numeric($demographicProject) && intval($demographicProject) >= 0) {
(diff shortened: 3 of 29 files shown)


modules/create_timepoint/php/timepoint.class.inc
```diff
@@ -188,7 +188,7 @@ class Timepoint extends \NDB_Page implements ETagCalculator
             // if there is only one project, autoselect first project from array of 1
             $project = \Project::getProjectFromID(array_pop($user_list_of_projects));
         } else if (count($user_list_of_projects) > 1) {
-            $project_id = intval($values['psc']);
+            $project_id = new \ProjectID($values['project']);
```
driusan, line 191: note: this incorporates #7261 (which was found by phan on this PR..)

driusan:
@ridz1208 now that this is finally passing would you mind reviewing it sometime? As the description says I just changed Project->getId() to return a ProjectID class and this is the minimal changes that were needed to get tests to pass after that.. it's way bigger than expected, but each individual line should be pretty self-evident/easy to review.

driusan:
@ridz1208 now that this is rebased/passing against can you review it before it goes stale?

driusan:
@laemtl since @ridz1208 is M.I.A. would you mind reviewing this? GitHub suggests you as a reviewer at the top of the PR (though I'm not entirely sure why..)

laemtl:
@driusan Sure!

modules/api/php/models/candidatesrow.class.inc
```diff
@@ -41,13 +41,15 @@ class CandidatesRow implements \LORIS\Data\DataInstance
...
         $this->_candid      = $row['CandID'] ?? null;
         $this->_projectname = $row['ProjectName'] ?? null;
-        $this->_projectid   = $row['ProjectID'] ?? null;
-        $this->_pscid       = $row['PSCID'] ?? null;
-        $this->_sitename    = $row['SiteName'] ?? null;
-        $this->_edc         = $row['EDC'] ?? null;
-        $this->_dob         = $row['DoB'] ?? null;
-        $this->_sex         = $row['Sex'] ?? null;
-        $this->_centerid    = $row['CenterID'] ?? null;
+        if ($row['ProjectID'] !== null) {
```
laemtl, line 44:
```
PHP 7.4 adds typed properties which do not default to null by default like untyped properties, but instead default to a special "uninitialised" state which will cause an error if the property is read before it is written.
``` 
Will it be safer to add a null default value, like for the other properties?

driusan: The property being set is untyped, so it should already default to null.

php/libraries/User.class.inc
```diff
@@ -423,16 +423,15 @@ class User extends UserPermissions implements \LORIS\StudyEntities\AccessibleRes
...
      *
      * @return bool
      */
-    public function hasProject(int $projectID) : bool
+    public function hasProject(ProjectID $projectID) : bool
     {
         return in_array(
             $projectID,
-            $this->getProjectIDs(),
-            true
```
laemtl, line 435: Loose checking can return weird results.

php/libraries/TimePoint.class.inc
```diff
@@ -1212,7 +1218,7 @@ class TimePoint implements \LORIS\StudyEntities\AccessibleResource
     {
         $centers = $user->getCenterIDs();
 
-        $projMatch   = in_array($this->getProjectID(), $user->getProjectIDs(), true);
```
laemtl, line 1215: Loose checking can return weird results.

php/libraries/TimePoint.class.inc
```diff
@@ -148,6 +148,12 @@ class TimePoint implements \LORIS\StudyEntities\AccessibleResource
             );
             $title = $subprojectSettings['title'] ?? '';
             $this->_timePointInfo['SubprojectTitle'] = $title;
+
+            if ($this->_timePointInfo['ProjectID'] !== null) {
```
laemtl, line 152: I'm a bit confused by this part.

php/libraries/TimePoint.class.inc
driusan: The PHP documentation for `in_array` is incorrect (according to my testing..) From what I can tell when the third parameter is true what it actually does is verifies with a `===` check instead of `==`. This results in the `in_array` requiring it to be the exact same instance of a class when the array is an array of objects (rather than simply checking the type (as the documentation says). Removing the strict checking still enforces types (since it's an object the object won't be `==` to other types either.) The note in the documentation about third parameter being to "validate types" seems to only be for built in types that get compared loosely like string/int/etc.

php/libraries/User.class.inc
driusan: It's required for the tests to pass. (See reply to other comment..)

php/libraries/ProjectID.php
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @return string
+     */
+    public function getType(): string
+    {
+        return 'ProjectID';
+    }
+
+    /**
+     * Validate that the value of the ProjectID is a positive integer that does
```
laemtl, line 20: Comment is truncated.

php/libraries/TimePoint.class.inc
driusan: What about it? The SQL query returned a nullable string. This is to convert it to a ProjectID type if it wasn't null.

php/libraries/ProjectID.php
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * This does not check for uniqueness in the database or any other
```
laemtl, line 22: So a ProjectID can still store a non-valid ID at the moment?

php/libraries/Candidate.class.inc
```diff
@@ -888,7 +892,7 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource
     {
         $centers = $user->getCenterIDs();
 
-        $projMatch   = in_array($this->getProjectID(), $user->getProjectIDs(), true);
```
laemtl, line 891: Loose checking can return weird results.

php/libraries/ProjectID.php
driusan: Yes, that's true of all our ID classes. (This was based on SessionID which was based on CandID class). They validate the structure of the identifier but not the value. Database calls are expensive to be making as often as these classes are instantiated (especially if you're doing something like going through the results of `SELECT ProjectID, otherstuff from sometable` in order to implement the getProjectID function on a data model.)

php/libraries/ProjectID.php
driusan: Fixed

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -108,7 +108,7 @@ class Stats_Demographic extends \NDB_Form
         $tpl_data['TableHeader']   = $tableHeader;
         $tpl_data['Disclamer']     = $disclamer;
         $tpl_data['Subcategories'] = $subcats;
-        $tpl_var = \Utility::getSubprojectsForProject(intval($projectID));
+        $tpl_var = \Utility::getSubprojectsForProject(($projectID));
```
laemtl, line 111: Not a big issue but I think the extra parenthesis are not necessary here.

driusan: Fixed

modules/api/php/models/candidatesrow.class.inc
laemtl: You are right, I thought it was, but I checked too fast.

php/libraries/TimePoint.class.inc
laemtl: Makes sense. Thanks for the clarification.

php/libraries/TimePoint.class.inc
laemtl: Now that I see that $this->_timePointInfo and $row store the same value, it makes sense, but it may be a little clearer to replace $row by $this->_timePointInfo

driusan: I'm not sure I understand. The line is already `$this->_timePointInfo`?

php/libraries/ProjectID.php
laemtl: I understand and it's good to know.

php/libraries/TimePoint.class.inc
laemtl:
```
$this->_timePointInfo['ProjectID'] = new \ProjectID(
    $this->_timePointInfo['ProjectID']
);
```

driusan: I changed it in the last commit.

laemtl:
After a quick tour, I found 2 problems introduced by the ProjectID conversion. I pushed a fix and everything LGTM.

laemtl (approved):

PR #7125, opened 2020-11-17
[statistics] Fix subprojects on demographics page

ridz1208:
## Brief summary of changes
subprojects are not properly displaying on the demographics tab see before and after.

before:
screen_shot_2563-11-17_at_09 09 48

after:
Screenshot from 2020-11-17 10-06-56

fix for code broken in: https://github.com/aces/Loris/pull/4046


driusan:
The first column in the "fixed" still looks wrong? Is there supposed to be a row of "Undefined" with all the values 0/0?

zaliqarosli (changes requested):
it looks to me here that the intended functionality was always to only show the subproject columns once you've selected a Project in the dropdown. I don't mind this added info by subproject as default though. The 'undefined' columns should be removed though i think, as pointed out by dave aswell.

ridz1208:
@zaliqarosli the undefined column always existed since suproject is not a required field in the SQL table, so all candidates with no subproject would not be counted anywhere but in the undefined column. Last discussion I had with @driusan was to either rename the column to something more informative or completely remove it and ignore candidates with no subprojects

any opinions ?

zaliqarosli:
@ridz1208 no opinions! i'll review after the change

ridz1208:
@driusan your call then. 

I just realised as well that any candidate with no sessions will also end up in the undefined column since only sessions have subprojects.

driusan:
I think at least we should change the column header to something like "No defined Subproject" so it doesn't look like it's a javascript UI bug.

ridz1208:
@driusan done

PR #7126, opened 2020-11-17
[statistics] Fix remove hardcoded IDs

ridz1208:
## Brief summary of changes
The demographics page currently uses hardcoded IDs `1` and `5` for `Active` and `Inactive` respectively, however the SQL 0000-00-00 patch does not insert with fixed IDs.

Added a subquery to refer to the description field instead

### DIFF
--- a/modules/statistics/php/stats_demographic.class.inc
+++ b/modules/statistics/php/stats_demographic.class.inc
@@ -349,7 +349,7 @@ class Stats_Demographic extends \NDB_Form
             "SELECT s.subprojectid as rowid,
                           count(DISTINCT(c.PSCID)) as val
                       FROM candidate as c
-                      	  LEFT JOIN participant_status ps ON (ps.CandID=c.CandID)
+                          LEFT JOIN participant_status ps ON (ps.CandID=c.CandID)
                           LEFT JOIN participant_status_options pso 
                                     ON (pso.ID=ps.participant_status)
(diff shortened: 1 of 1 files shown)


modules/statistics/php/stats_demographic.class.inc
```diff
@@ -350,15 +350,17 @@ class Stats_Demographic extends \NDB_Form
                           count(DISTINCT(c.PSCID)) as val
                       FROM candidate as c
                       	  LEFT JOIN participant_status ps ON (ps.CandID=c.CandID)
+                          LEFT JOIN participant_status_options pso
```
laemtl, line 353: Indentation is slightly off.

modules/statistics/php/stats_demographic.class.inc
```diff
@@ -381,14 +383,16 @@ class Stats_Demographic extends \NDB_Form
...
```
laemtl, line 386: Indentation is slightly off.

laemtl (commented):
Indentation is slightly off.

modules/statistics/php/stats_demographic.class.inc
ridz1208: tried fixing it

modules/statistics/php/stats_demographic.class.inc
ridz1208: tried fixing it

laemtl (approved):

PR #7128, opened 2020-11-18
Add configuration to run CI with GitHub actions

driusan:
Given the changes to Travis pricing which are changing
at the end of the year and the long VM startup times
we've been experiencing lately, this adds a configuration
for GitHub actions to run our continuous integration
directly on GitHub.

The .travis.yml is kept around for now, but will likely
eventually be removed. PHP 7.4 is also removed from the
Travis configuration, because I discovered while setting
this up that there's a bug where both "PHP 7.3" and "PHP 7.4"
are running PHP 7.3 under docker with our current configuration.
(The bug is fixed in the GitHub actions configurations, but
given that we'll be moving away from Travis, isn't worth fixing
there.)

### DIFF
--- a/.github/workflows/loristest.yml
+++ b/.github/workflows/loristest.yml
@@ -1,0 +1,49 @@
+name: LORIS Test Suite
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+        matrix:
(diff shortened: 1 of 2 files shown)


driusan:
Blocked by #7101 .. GitHub Actions comes with composer 2.0, which breaks with our current dependencies.

PR #7145, opened 2020-11-26
[Create timepoint] Language dropdown index fix

laemtl:
Small change to fix a reindexing of the languages, causing a database insert error (key constraint failure on languageID).

## To test
Add languages to the database (language) at sparse indexes (ex: 2, 4).
You should be able to submit without a 500 error.

* Related to #7146 (v23.0 release)


driusan:
Replaced by #7145

laemtl:
@driusan Create timepoint was reactified and this PR fix the same issue as #7146, but for main and for the reactified version.

driusan:
I didn't notice that it's changing a different file name because of reactification so it won't get changed when the branch is pulled forward, this does need to be a separate PR afterall.

jesscall (approved):

PR #7146, opened 2020-11-26
[Create timepoint] Language dropdown index fix

laemtl:
Small change to fix a reindexing of the languages, causing a database insert error (key constraint failure on languageID).

## To test
Add languages to the database (language) at sparse indexes (ex: 2, 4).
You should be able to submit without a 500 error.

* Related to #7145 (main)

### DIFF
--- a/modules/create_timepoint/php/create_timepoint.class.inc
+++ b/modules/create_timepoint/php/create_timepoint.class.inc
@@ -205,7 +205,7 @@ class Create_Timepoint extends \NDB_Form
 
                 $languages = \Utility::getLanguageList();
                 if (count($languages) > 1) {
-                    array_unshift($languages, null);
+                    $languages = [null] + $languages;
                     $this->addSelect(
                         'languageID',
                         'Language',


modules/create_timepoint/php/create_timepoint.class.inc
```diff
@@ -205,7 +205,7 @@ class Create_Timepoint extends \NDB_Form
 
                 $languages = \Utility::getLanguageList();
                 if (count($languages) > 1) {
-                    array_unshift($languages, null);
+                    $languages = [null] + $languages;
```
maltheism, line 208: @laemtl I'm a little confused since I've never seen it done this way before. So I did a quick test in php with the following:

```
<?php
// Your code here!

$test =['a','b','c'];
print_r($test);

array_unshift($test, null);
print_r($test);

$test =['a','b','c'];

$test = [null] + $test;
print_r($test);

?>
```

The output:
```
Array
(
    [0] => a
    [1] => b
    [2] => c
)
Array
(
    [0] => 
    [1] => a
    [2] => b
    [3] => c
)
Array
(
    [0] => 
    [1] => b
    [2] => c
)
```

It looks like your way replaces the first element with null?

laemtl: Indexes start from 1 in Mysql, this case never occurs.

maltheism (approved):

modules/create_timepoint/php/create_timepoint.class.inc
ridz1208: @laemtl yes and no, a database admin could set the field to `0` it's a valid value. I think the solution is easy
```suggestion
                    $languages = [null=>null] + $languages;
```

I havent tried it but this should fix both of your issues

ridz1208:
@kongtiaowang any idea whats happening to travis ?

driusan:
It needs the equivalent of #7113 or #7101 since Travis upgraded to composer 2.

ridz1208:
@driusan can we put github actions on 23 and remove travis at this stage ?

driusan:
@ridz1208 probably, but that won't help in this case because GitHub actions also uses composer 2 by default.

ridz1208:
@laemtl this will need a rebase after https://github.com/aces/Loris/pull/7148 is merged

laemtl:
No problem.

laemtl:
@ridz1208 Rebased.

jesscall (approved):

PR #7148, opened 2020-11-27
[TRAVIS] fix composer issue for 23

ridz1208:
Fix composer travis issue on 23. replica of #7113


ridz1208:
@driusan can we merge this one ?

driusan:
There's probably a flag to say "the latest 1.x release", but since this is a temporary thing until we're off travis it's not worth the time to look into it..

PR #7149, opened 2020-11-27
Update documented version of composer in README for 23 release

driusan:
It does not work with "or higher", composer 2 only works on the
main branch.


ridz1208:
requires https://github.com/aces/Loris/pull/7148

driusan:
Rebased, it should pass now..

ridz1208 (approved):

PR #7151, opened 2020-11-27
[Datadict] filter issue when table name is different then test_name

ridz1208:
## Brief summary of changes
The Data Dictionary allows for filtering on the data from the `parameter_type` table. The `parameter_type` table is built by the `data_dictionary_builder` script which parses the LINST format of any instrument to generate the table entries. the LINST format of a PHP instrument is generated by the `lorisform_parser` script which writes the TABLE of the instrument into the LINST file it creates...

result: The table name is stored in the parameter_type SQL table, and the datadict module expects to filter on test name causing any filtering by instrument to return 0 results when the testname and the table name of an instrument differs.

1. create or modify an instrument to have a testname different then it's table name. (both parameters are usually defined in the instrument class' setup function)
2. run the lorisform_parser on your instrument and the data_dict_builder as instructed in the LORIS wiki
3. navigate to the datadict module and try to filter by the instrument you have chosen

Before these changes, the filter will always return 0 results. with these changes you will see all your fields for that instrument

* Fixes issues for both CBIGR and CCNA


ridz1208:
@jesscall give it a try on the CCNA instruments

ridz1208:
@kongtiaowang I'm not sure why this is failing travis, The module works fine on my VM with RB installed. My guess would be that the test suite might be trying to install another instrument with is throwing an error because the new code attemps to load any instrument to get their table names.

Can you please take a look ?

ridz1208:
@kongtiaowang still failing

modules/datadict/php/datadict.class.inc
```diff
@@ -58,8 +58,21 @@ class Datadict extends \DataFrameworkMenu
...
+        foreach ($instruments as $instrument => $name) {
+            // Since the testname does not always match the table name in the
+            // database we need to instantiate the object to get the table name
+            $iObj = \NDB_BVL_Instrument::factory(
+                $instrument,
+                '',
+                ''
+            );
+            $key  = $iObj->usesJSONData() ? $iObj->testName:$iObj->table;
+            $dictInstruments[$key] = $name;
```
jesscall, line 72: maybe wrap this in a try/catch 

I'm running into issues with the module not loading if a test exists in test_names without any actual instrument existing for it.

modules/datadict/php/datadict.class.inc
```diff
@@ -58,8 +58,21 @@ class Datadict extends \DataFrameworkMenu
      */
     public function getFieldOptions() : array
     {
+        $instruments     = \Utility::getAllInstruments();
+        $dictInstruments = array();
+        foreach ($instruments as $instrument => $name) {
```
jesscall, line 63:
```suggestion
        $instruments     = \Utility::getAllInstruments();
        $DB  =& \Database::singleton();
        $activeInstruments = array_filter(
            $allInstruments,
            [$DB, 'tableExists'],
            ARRAY_FILTER_USE_KEY
        );
        $dictInstruments = array();
        foreach ($activeInstruments as $instrument => $name) {
```

modules/datadict/php/datadict.class.inc
jesscall: see my suggestion above ^

jesscall (changes requested):
Aside from the suggestion; I also noticed that the dropdown filter is not populating properly.

Screen Shot 2020-12-01 at 2 53 55 PM

ridz1208:
@jesscall on RB the dropdown is populating fine. my guess would be that an error is being thrown on CCNA somewhere whengenerating the list that is preventing the list from being returned. could you pinpoint the source ?

modules/datadict/php/datadict.class.inc
ridz1208: @jesscall that would automatically eliminate all JSONdata instruments which do not have tables

ridz1208: and that would eliminate all instruments which have testnames different then the table names which is exactly to opposite of what I'm trying to do here

modules/datadict/php/datadict.class.inc
ridz1208: see my replies above

modules/datadict/php/datadict.class.inc
jesscall: Very true

ridz1208:
@kongtiaowang thank you for making this PR pass tests. I was just wondering if we can do this  bit differently.

I dont like having the test instrument part of raisinbread, I feel like raisinbread should be useable separately not only for loris testing. 

Is it possible to keep the raisinbread/instruments mounted (so you change to docker compose is good) but undo the move of the test instrument into raisinbread and instead just copy it into the directory either from composer or the testing suite ?? the idea here is to keep the test_instrument separated in the codebase, but right before the tests are run, copy it directly into project/instruments

jesscall:
Before:

Screen Shot 2021-01-05 at 2 41 04 PM

After:
Screen Shot 2021-01-05 at 2 42 51 PM


Notice that **previously** both test names appear in the list:
 [1] `Surgical History `
 [2] `Surgical History (Follow-Up)` --> where this one is empty with no results

After checking out this branch:
All data dictionary entries (for both [1] and [2]) appear under the `Surgical History (Follow-Up)` test name

ridz1208:
@jesscall this is a tricky problem. 

1. i'm gonna assume that if you select the (follow up) you at least get the results in the table (i.e. the proper key is being selected but its the incorrect name)

In which case the issue is that the loop I have that is fetching the instrument table can not determine in your case which of the instruments is the "primary" one... it just knows that 2 instruments are using the same table and decides at random which one to use for the name. the choice is not exactly at random I guess, it is dependent on which one is queried first from SQL whihc in turn can be random...

our options:

1. always go with the last test name (behaviour that is currently implemented)
2. always go with the first testname, do not override the test name as you go through the list (in your case this MIGHT display the non followup instrument)
3. display all test names sharing the same table ie `Surgical History/Surgical History (Follow-up)`


@driusan you might have an opinion here

driusan:
I don't understand.. why are there 2 instruments using the same table? What's the use case for that? It seems to be enough of an edge case that I don't think it's going to affect most people.

Regardless, for option 3 are you talking about in the table column or in the dropdown? (Either way, 3 sounds the most intuitive to me..)

ridz1208:
@driusan the usecase is when 2 instruments are very very similar and we dont wanna make a full copy of the instrument. the usecase for CCNA for example is that the instrument used at the first visit has 1 question less than the ones used at the followup visits. instead of duplicating the instrument and making maintence twice harder they extend the initial instrument and just add the 1 question for the followup visit

ridz1208:
@jesscall can you give this another try ?

modules/datadict/php/datadict.class.inc
```diff
@@ -58,8 +58,26 @@ class Datadict extends \DataFrameworkMenu
...
+            $iObj = \NDB_BVL_Instrument::factory(
+                $instrument,
+                '',
+                ''
+            );
+            $key  = $iObj->usesJSONData() ? $iObj->testName:$iObj->table;
+            // Append other testnames if multiple tests use the same table
+            $otherNames = '';
+            if (isset($dictInstruments[$key])) {
+                $otherNames = "/".$dictInstruments[$key];
```
jesscall, line 75:
```suggestion
                $otherNames = " / ".$dictInstruments[$key];
```

modules/datadict/php/datadict.class.inc
```diff
@@ -58,8 +58,26 @@ class Datadict extends \DataFrameworkMenu
...
+            }
+            $dictInstruments[$key] = $name . $otherNames;
```
jesscall, line 77: Small detail, but since `$name` refers to the last test name, can we reverse the order here for which name is displayed first ?

Screen Shot 2021-01-06 at 12 16 22 PM

ridz1208:
@jesscall done

jesscall (approved):

modules/datadict/php/datadict.class.inc
```diff
@@ -58,8 +58,26 @@ class Datadict extends \DataFrameworkMenu
      */
     public function getFieldOptions() : array
     {
+        $instruments     = \Utility::getAllInstruments();
+        $dictInstruments = array();
+        foreach ($instruments as $instrument => $name) {
+            // Since the testname does not always match the table name in the
+            // database we need to instantiate the object to get the table name
+            $iObj = \NDB_BVL_Instrument::factory(
```
driusan, line 66: This is going to cause the datadict to crash if you have an invalid instrument that throws an exception (ie. something invalid in test_names, maybe because you're switching between branches while developing an instrument or debugging an old/new version of LORIS..) . Can you put this in a try/catch block?

ridz1208: what should I put in the catch ?
1. throw exception ? (defeats the purpose here I guess lol)
2. skip ?
3. default to testname ?

driusan: Probably skip. I don't think defaulting to the test name is likely to result in any matches since the instrument is invalid, but either option 2 or 3. Maybe write to the error log too.

ridz1208: @driusan I made the changes you requested. I have not tested however can you make sure to review carefully ?

modules/datadict/php/datadict.class.inc
```diff
@@ -58,8 +58,36 @@ class Datadict extends \DataFrameworkMenu
...
+            } catch (\Exception $e) {
+                error_log(
+                    "There was a problem instantiating the instrument ".
+                    "'$instrument'. Make sure the instrument is valid and ".
+                    "functional in order for it to be displayed in the data ".
+                    "dictionary."
+                );
+                continue;
+            }
+            $key = $iObj->usesJSONData() ? $iObj->testName:$iObj->table;
```
driusan, line 81:
```suggestion
            // If the instrument saves data in the flag table, there is no table to
            // use, use the test name.
            $key = $iObj->usesJSONData() ? $iObj->testName:$iObj->table;
```

It took me a while to figure out what this was doing.. it's worth adding a comment.

driusan (approved):
looks good once the CI passes

ridz1208:
@driusan passed CI

PR #7153, opened 2020-11-30
[LINST] Support for different test_name and table name

ridz1208:
## Brief summary of changes
The `NDB_BVL_Instrument` class has supported instruments having test names different from the table names. The testname is generally used to load instruments in the front en while the table name is used to save/load data as well as the data dictionary.

These LINST tinstrument class extends the `NDB_BVL_Instrument` class but does not incorporate such functionality making it more restrictive than it's parent. This fix adds that capability while keeping the class 100% backwards compatible

This branch also adds the capacity to define if your LINST instrument should use the jsonData saving methid or the regular SQL saving.

All changes in this PR are 100% backwards compatible and futre deprecations will be done in the next major release.

### testing
- load this branch and test any existing LINST instrument. everything should work fine. (you can use the old version of the bmi for that before the changes)
- On this branch again, create (or modify) and instrument to use a table name different than the testname and add a line in the LINST file for the `test` variable (see code changes for BMI example). The testname should now be loaded from `test` and the table name separately loaded from `table`. toggle the jsondata value from false to true and make sure the instrument then saves as JSON data in the flag table

PS: I did not modify the behaviour of the instrument builder since I think we should instead make the instrument builder define (by default) `jsonData=true` instead of using and SQL table, in which case the table variable would be deprecated and the LINST class would need to be updated. This work should be done in the next major release

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -273,6 +273,11 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         if (!empty($commentID)) {
             $obj->setupCandidateInfoTables();
         }
+        if (file_exists($base."project/instruments/$instrument.meta")) {
+            $obj->loadInstrumentMetadata(
+                $base . "project/instruments/$instrument.meta"
+            );
+        }
         // Adds all of the form element and form rules to the page after
         // having instantiated the form above
         $obj->loadInstrumentFile(
@@ -2313,6 +2318,22 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             array('CID' => $this->commentID),
             array('nofetch' => true)
         );
+    }
+
+    /**
+     * Parses the metadata of a LINST instrument and sets the proper variables
+     * in the instrument class accordingly
+     *
+     * This is intended to be overridden by a subclass of NDB_BVL_Instrument
+     * to add support for other file formats.
+     *
+     * @param string $filename The filename to be loaded, or a base64 encoded
+     *                         string.
+     *
+     * @return void
+     */
+    function loadInstrumentMetadata(string $filename): void
+    {
     }
 
     /**
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2321,14 +2321,13 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     }
 
     /**
-     * Parses the metadata of a LINST instrument and sets the proper variables
+     * Loads the metadata of an instrument and sets the proper variables
      * in the instrument class accordingly
      *
      * This is intended to be overridden by a subclass of NDB_BVL_Instrument
      * to add support for other file formats.
      *
-     * @param string $filename The filename to be loaded, or a base64 encoded
-     *                         string.
+     * @param string $filename The filename to be loaded
      *
      * @return void
      */
--- a/php/libraries/NDB_BVL_Instrument_LINST.class.inc
+++ b/php/libraries/NDB_BVL_Instrument_LINST.class.inc
@@ -461,14 +461,13 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
                     }
                     $firstFieldOfPage = true;
                     break;
-                case 'test':
-                    $this->testName = trim($pieces[1]);
-                    break;
                 case 'table':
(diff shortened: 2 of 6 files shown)


ridz1208:
will be done properly and reopened

driusan:
@zaliqarosli can you review/test this?

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2315,6 +2320,22 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+     *
+     * This is intended to be overridden by a subclass of NDB_BVL_Instrument
+     * to add support for other file formats.
+     *
+     * @param string $filename The filename to be loaded, or a base64 encoded
+     *                         string.
+     *
+     * @return void
+     */
+    function loadInstrumentMetadata(string $filename): void
```
zaliqarosli, line 2335: is this maybe supposed to be an abstract function?

php/libraries/NDB_BVL_Instrument_LINST.class.inc
```diff
@@ -734,6 +738,44 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
...
+    {
+        $fp = fopen($filename, "r");
+        while (($line = fgets($fp, 4096)) !== false) {
+            $pieces = preg_split("/{@}/", $line);
+            $this->LinstLines[] = $pieces;
+
+            $type = $pieces[0];
+
+            switch ($type) {
+            case 'test':
```
zaliqarosli, line 763: should 'test' maybe be 'name' instead? i feel like test doesnt mean much without context here and i'm under the impression that 'test' is an older implementation that we're now replacing with 'instrument'? or even 'testname' would be better i think

raisinbread/instruments/bmi.linst
```diff
@@ -1,4 +1,3 @@
-table{@}bmi
```
zaliqarosli, line 1: could we maybe add a .meta file to this PR if we're removing this line from the raisinbread example?

zaliqarosli (commented):
i don't fully understand the purpose of this change. i'm not sure what case exists that requires the test name and the table name to be different. i feel like its an okay restriction to set. it might help my understanding to know what future deprecations (as mentioned in PR description) will be implemented as a follow up to this change.

php/libraries/NDB_BVL_Instrument.class.inc
ridz1208: I made it to be empty just to be consistent with the function below `loadInstrumentFile()`

In practice however, making this abstract would require for every project to implement this function for every instrument which extends this class... it would be too costly for projects

raisinbread/instruments/bmi.linst
ridz1208: I thought I had but it seems like i didnt submit it. ma bad

php/libraries/NDB_BVL_Instrument.class.inc
zaliqarosli: truee

ridz1208:
@zaliqarosli its done

zaliqarosli (approved):

php/libraries/NDB_BVL_Instrument_LINST.class.inc
```diff
@@ -734,6 +738,44 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
         }
     }
 
+    /**
+     * Parses the metadata of a LINST instrument and sets the proper variables
```
driusan, line 742: Why make this support LINST specific instead of putting it in the base class? I can't think of any reason you wouldn't want to allow other instruments to set this metadata and/or would want a different format for the meta file based on the type of instrument..

ridz1208: wait... I'm not sure if your saying to apply the same logic for PHP instruments?

its not the same logic, ie PHP instruments already set the table and testnames variables in their setup functions usually (or at least in all the instruments and templates we currently have).

However I just looked at the similar function `loadInstrumentRules` and its indeed not specific to Linst... so I guess i will just modify the documentation to not specifically mention LINST although i find it improbable that PHP instruments would use it

ridz1208: @driusan i did make the changes in the instrument class not this one since this is the LINST class the doc seems appropriate

PR #7155, opened 2020-11-30
[instrument] insert JSON intrument fix

laemtl:
Fixes minor warnings and the logic to populate an instrument with json_data === true.

### DIFF
--- a/php/libraries/NDB_BVL_Battery.class.inc
+++ b/php/libraries/NDB_BVL_Battery.class.inc
@@ -193,11 +193,19 @@ class NDB_BVL_Battery
         // generate a commentID
         $commentID = $this->_createCommentID($testName);
 
+        // generate the double data entry commentid
+        $ddeCommentID = 'DDE_'.$commentID;
+
         // instantiate instrument object to get table name attribute
         $obj = NDB_BVL_Instrument::factory($testName, '', '');
 
-        // insert into the test table
-        $DB->insert($obj->table, array('CommentID' => $commentID));
+        if ($obj->usesJSONdata() !== true) {
+            // insert into the test table
+            $DB->insert($obj->table, array('CommentID' => $commentID));
+
+            // insert the dde into the test table
+            $DB->insert($obj->table, array('CommentID' => $ddeCommentID));
+        }
 
         // insert into the flag table
         $DB->insert(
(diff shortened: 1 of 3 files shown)


php/libraries/NDB_BVL_Battery.class.inc
```diff
@@ -193,11 +193,19 @@ class NDB_BVL_Battery
...
 
+        // generate the double data entry commentid
+        $ddeCommentID = 'DDE_'.$commentID;
+
         // instantiate instrument object to get table name attribute
         $obj = NDB_BVL_Instrument::factory($testName, '', '');
 
-        // insert into the test table
-        $DB->insert($obj->table, array('CommentID' => $commentID));
+        if ($obj->jsonData !== true) {
```
driusan, line 202: JSON and DDE are not related. Why are you making DDE conditional on not saving the data as JSON?

driusan (changes requested):

php/libraries/NDB_BVL_Battery.class.inc
ridz1208: I think because the instrument table does not exist in a JSON data instrument, so you cant really insert anything into it even DDE rows

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -35,7 +35,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      * JSON format from the "Data" column of the flag table, or
      * from the traditional SQL tables.
      */
-    protected $jsonData = false;
+    public $jsonData = false;
```
ridz1208, line 38: why was this made public ? you can reach the value from outside using `usesJSONdata()` or a soemthing like that. I dont think this variable should be modifiable by external classes

ridz1208 (changes requested):

php/libraries/NDB_BVL_Instrument.class.inc
laemtl: @ridz1208 Fixed! I missed this accessor. I agree that it's safer to rely on it. Should it be the same for other properties like table, which is public?

ridz1208 (approved):

driusan (approved):

PR #7161, opened 2020-12-02
[Instrument.class.inc] Fix _nullScores() so that only score fields are saved

### LINKED ISSUE #7140: score_instruments.php is causing double escaping on each run
Running score_instruments.php is causing double escaping on each run

To Reproduce - 
Run `php tools/data_integrity/score_instruments.php all inprogress`. 

I specifically was running "all inprogress" but I imagine this happens for all scoring updates with this script. I verified this script was causing the issue by running the instrument_double_escape_fix.php script from @ridz1208 to remove all instances of double escaping and then running the score_instruments script. 

LORIS v23.0.2
php 7.3.11

jesscall:
## Brief summary of changes
Currently the `_nullScores()` function saves all fields in the table, leading to an issue with double escaping of special characters. In other to amend, this PR ensures that only score fields are being saved by the `_nullScores()` function. 

1. Input a special character in an instrument text field (ex: `<`, `>`, `"`, `&`)
2. Run score instrument script to ensure no double escaping occurs.

* Resolves #7140

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2086,7 +2086,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     function _nullScores(array $scoreCols): void
     {
-        $record = $this->loadInstanceData($this);
+        $record    = $this->loadInstanceData($this);
         $scoreCols = array_combine($scoreCols, $scoreCols);
         // order is important here because it keeps values of first array
         // unsetting all columns which are not scores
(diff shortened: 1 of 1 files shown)


ridz1208 (approved):

driusan:
Can you rebase? Haven't looked into why it's failing but if it's old enough that it was running Travis instead of GitHub Actions it's likely things have changed on the branch since then..

jesscall:
@driusan done

driusan:
phpcs is failing

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2086,7 +2086,11 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     function _nullScores(array $scoreCols): void
     {
-        $data = $this->loadInstanceData($this);
+        $record    = $this->loadInstanceData($this);
```
driusan, line 2089: Note: this will need to change to $this->getInstanceData() when merged back into main. (loadInstanceData is deprecated, but getInstanceData isn't on this branch..)

PR #7157, opened 2020-12-01
[composer] Dependencies updates required for Php 8.0.0

maltheism:
## Brief summary of changes

Tonight brew on my mac, upgraded Php to 8.0.0 from 7.4.x and LORIS.. well, composer & smarty became unhappy about it.

I went ahead by updating the dependencies "composer.json" and notably had success getting past the composer problems. Although, smarty seems to be an issue when I visit my LORIS. Smarty apparently requires isset() checks for variables now. Everything seems to be working after updating that foregoing nuisance throughout LORIS.

TODO Required smarty fixes for LORIS:
I think I got them all...

The smarty errors example (refreshing loris on main page) needing isset:
```
Warning: Undefined array key "bvl_feedback" in /Users/alizee/Development/GitHub/McGill/Loris/smarty/templates_c/6b3affa8fc897d3dd4f16fd36c5406c7e07e622e_0.file.main.tpl.php on line 148
Warning: Attempt to read property "value" on null in /Users/alizee/Development/GitHub/McGill/Loris/smarty/templates_c/6b3affa8fc897d3dd4f16fd36c5406c7e07e622e_0.file.main.tpl.php on line 148
Warning: Undefined array key "control_panel" in /Users/alizee/Development/GitHub/McGill/Loris/smarty/templates_c/6b3affa8fc897d3dd4f16fd36c5406c7e07e622e_0.file.main.tpl.php on line 157
Warning: Attempt to read property "value" on null in /Users/alizee/Development/GitHub/McGill/Loris/smarty/templates_c/6b3affa8fc897d3dd4f16fd36c5406c7e07e622e_0.file.main.tpl.php on line 157
```

The composer complaints before dependency updates:
```
Problem 1
    - Root composer.json requires bjeavons/zxcvbn-php ^0.4 -> satisfiable by bjeavons/zxcvbn-php[0.4.0].
    - bjeavons/zxcvbn-php 0.4.0 requires php ^5.6 || ^7.0 -> your php version (8.0.0) does not satisfy that requirement.
  Problem 2
    - Root composer.json requires phpunit/phpunit 7.0.0 -> satisfiable by phpunit/phpunit[7.0.0].
    - phpunit/phpunit 7.0.0 requires php ^7.1 -> your php version (8.0.0) does not satisfy that requirement.
  Problem 3
    - Root composer.json requires phan/phan 3.0.0 -> satisfiable by phan/phan[3.0.0].
[...]

PR #7166, opened 2020-12-03
[smarty] use default to prevent error_log pollution

maltheism:
## Brief summary of changes

Updating to use php 8 and updating composer.json dependencies results in smarty polluting error_log about undefined variables. $some_variable|default will mimic the existing behaviour and have the variable be set to an empty string if undefined. 

1. Checkout PR.
2. Visit Loris everywhere

### DIFF
--- a/modules/statistics/php/stats_behavioural.class.inc
+++ b/modules/statistics/php/stats_behavioural.class.inc
@@ -245,13 +245,10 @@ class Stats_Behavioural extends \NDB_Form
             $center = $row['CenterID'];
             $vl     = $row['VLabel'];
             $value  = intval($row['val']);
-            if (!isset($this->tpl_data['behaviour']['C' . $center])) {
-                $this->tpl_data['behaviour']['C' . $center] = [$vl => []];
-            }
+
+            $this->tpl_data['behaviour']['C' . $center][$vl]
+                = $this->tpl_data['behaviour']['C' . $center][$vl] ?? [];
             $c =& $this->tpl_data['behaviour']['C' . $center];
-            if (!isset($c[$vl])) {
-                $c[$vl] = [];
-            }
 
             if (array_key_exists('total', $c[$vl])) {
                 $c[$vl]['total'] += $value;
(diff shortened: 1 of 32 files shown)


driusan:
Why is this blocked?

maltheism:
@driusan not all fixes have been added yet. I used git cherry-pick but I'm missing some. I'll remove blocked tag when they're all added. I'm manually going through loris right now.

maltheism:
@driusan I believe it's ready. I might of missed something but I'm doubtful that happened.

src/Middleware/UserPageDecorationMiddleware.php
```diff
@@ -135,7 +135,7 @@ function ($a, $b) {
 
         $tpl_data['subtest'] = $request->getAttribute("pageclass")->page ?? null;
 
-        $page = $request->getAttribute("pageclass");
+        $page = $request->getAttribute("pageclass") ?? '';
```
maltheism, line 138: this made it so I could view other smarty problems. @driusan maybe I should undo it?

modules/statistics/php/stats_behavioural.class.inc
```diff
@@ -245,7 +245,13 @@ class Stats_Behavioural extends \NDB_Form
             $center = $row['CenterID'];
             $vl     = $row['VLabel'];
             $value  = intval($row['val']);
-            $c      =& $this->tpl_data['behaviour']['C' . $center];
+            if (!isset($this->tpl_data['behaviour']['C' . $center])) {
```
maltheism, line 248: @driusan the fixes in here maybe should be moved to a separate PR? Solves error_pollution and might not be the prettiest code. @laemtl you wouldn't happen to know a better way? I believe you were the one that cleaned up this file.

laemtl: What about 
$c =& $this->tpl_data['behaviour']['C' . $center][$vl] ?? [$vl => []];

driusan:
Shouldn't this update the smarty version in composer.json/lock too?

modules/statistics/php/stats_behavioural.class.inc
maltheism: @laemtl my IDE is giving me the warning: `Only variables can be passed by reference` message. Can you test that for me whenever you have time and check your error_log? The error log gets polluted when visiting the statistics module and clicking on the Behavioural Statistics tab.

laemtl: @maltheism This shoul works:
$this->tpl_data['behaviour']['C' . $center][$vl] = $this->tpl_data['behaviour']['C' . $center][$vl] ?? [];
$c =& $this->tpl_data['behaviour']['C' . $center];

maltheism:
@driusan ready

maltheism:
Added Priority High tag since https://github.com/aces/Loris/pull/7157 has it and this PR is needed to be merged for it.

PR #2006, opened 2016-07-19
Instrument_list query mod to left join on test battery

ridz1208:

PR #7167, opened 2020-12-03
[instrument_list/battery] Clean-up battery retrieval function

ridz1208:
## Brief summary of changes

1. Removed getBatteryInstrumentList function from the instrument_list module and made it query the data from the correct place which is the NDB_BVL_Battery class
2. Modified query in NDB_BVL_Battery to remove duplicate returns and avoid "jumpy" test_names depending on if an instrument_order is defined or not (see individual comments below for more details)
3. Removed unused/inaccurate parameters for Battery functions in Battery class. The parameters STAGE, SUBPROJECTID and VISIT LABEL should all be queried from the session itself (Which was already mandatory to define. see `_assertBatterySelected`) but are also either completely unsued in these functions or not necessary at all.

@jesscall @sruthymathew123 @zaliqarosli each of you should test this on at least a handful of timepoints since you all use different functionalities affecting the instrument list module
- QPN uses instruments at the "Not Started" stage which might be affected by these modifications
- IBIS has several test_batteries which diverge from the "defined batteries" which is what the now deleted function was added for in the first place (see #2006). Also can test compatibility with survey instruments
- CCNA uses instrument ordering as well as manual addition of individual instruments to sessions

make sure:
1. instrument ordering works when order is defined in the battery table
2. if ordering is not defined for some or all of the instruments, order should default to Full Name displayed
3. ALL instruments in the flag table for that session appear in the instrument_list module, even instruments which are not in the test_battery defined for that session (ie survey instruments and manually added instruments)
4. Control panel properly checks that all instruments are complete in order to allow pass/fail changes and completion of the session (this functionality depends on being able to query all instruments of the session)


php/libraries/NDB_BVL_Battery.class.inc
```diff
@@ -292,67 +285,49 @@ class NDB_BVL_Battery
...
-        if (!is_null($stage) && !is_null($SubprojectID)) {
-                $query .= " AND (
-                    b.SubprojectID IS NULL
-                    OR b.SubprojectID=:SubprojID
-                )";
-                $qparams['SubprojID'] = $SubprojectID;
-        }
-
-        if (Utility::columnsHasNull('test_subgroups', 'group_order')) {
+        $query = "SELECT DISTINCT f.Test_name,
```
ridz1208, line 298: This should be converted to a group BY to avoid start next stage duplication issues

ridz1208:
@jesscall @CamilleBeau @pierre-p-s I need someone from CCNA to test this one, you guys have the more complex batteries and use all sorts of ordering. 

read the description carefully, theres a bunch of clean up here and testing should be rigorous

zaliqarosli:
@ridz1208 does this fix a specific issue? I am wondering if this PR indeed fixes the error you get when you try to 'Start Visit Stage' on a session that already has a survey created i.e. a test battery generated, blocking the whole test battery from getting populated

php/libraries/NDB_BVL_Battery.class.inc
zaliqarosli: does this still need to be resolved?

ridz1208:
@zaliqarosli no this might fix instruments not displaying in the same order when you load instrument_list multiple times. the issue you are talking about was however fixed in a different PR

php/libraries/NDB_BVL_Battery.class.inc
ridz1208: yes but its unrelated to the survey_accounts bug if thats what you were thinking

php/libraries/NDB_BVL_Battery.class.inc
```diff
@@ -294,67 +287,49 @@ class NDB_BVL_Battery
...
+                      t.isDirectEntry,
+                      b.instr_order as instrument_order
+                  FROM flag f
+            	      JOIN test_names t ON (f.Test_name=t.Test_name)
+            	      JOIN test_subgroups ts ON (ts.ID = t.Sub_group)
+            	      LEFT JOIN session s ON (s.ID=f.SessionID)
+            	      LEFT JOIN test_battery b
+       	                ON ((t.Test_name=b.Test_name OR b.Test_name IS NULL)
+      	                AND (s.SubprojectID=b.SubprojectID OR b.SubprojectID IS NULL)
+       	                AND (s.Visit_label=b.Visit_label OR b.Visit_label IS NULL)
```
pierre-p-s, line 315: Why do you check for Null subproject IDs or visit labels? 
Shouldnt the instruments in the test battery always have those information?

php/libraries/NDB_BVL_Battery.class.inc
```diff
@@ -294,67 +287,49 @@ class NDB_BVL_Battery
...
-        //print_r($query);
+        $query  .= ", instrument_order, Full_name";
+        $qparams = ['SID' => $this->sessionID];
+
         // get the list of instruments
         $rows = $DB->pselect($query, $qparams);
 
         // return all the data selected
         return $rows;
-    } // end getBatteryVerbose()
```
pierre-p-s, line 357: There s the same thing line 284 for the getBattery function

pierre-p-s (commented):
Tested and the ordering seemed to be working. Pass fail function worked as intended.

cmadjar:
@ridz1208 can you rebase? Thanks!

php/libraries/NDB_BVL_Battery.class.inc
ridz1208: no, see https://github.com/aces/Loris/blob/main/modules/battery_manager/help/battery_manager.md

ridz1208:
@pierre-p-s please re-review/re-test

@zaliqarosli if you can find some time to test (see description for more)

pierre-p-s:
Manual test and review worked. 

However we found an issue in the patch 2020-02-10_NewModulePermissions.sql in SQL/New_patches with @ridz1208. Should I make an issue?

ridz1208:
@pierre-p-s let's investigate that together for another 10 minutes before doing that.

ridz1208:
@zaliqarosli waiting on your testing

zaliqarosli:
testing on QPN data:

- instruments defined at the "Not Started" stage populated in the test battery table on instrument_list page i.e. not affected by changes in this PR
- instrument ordering and default ordering by Full Name if instr_order field is NULL works as described in this PR
- survey instruments (not defined in test_battery) show up on instrument_list page

failed:
- I am able to set Session BVL QC status to 'Complete' even if instruments are 'In progress'. PR description suggests I shouldn't be able to do this. @ridz1208 is that right?

ridz1208:
@zaliqarosli testing ITEM 4 involves making sure you can only toggle the VISIT to `pass` or `fail` when all instruments are set to complete. Not `BVL QC status` nor `BVL QC type` if that's the only issue you found it means this passes tests

zaliqarosli (approved):
passes testing

PR #7169, opened 2020-12-04
[LINST] Implement getFullName() and getSubtestList() and fix numeric type

ridz1208:
## Brief summary of changes

LINST instrument class was using the legacy trait to bypass defining it's own instrument's full name and subtests, both values were already somewhere in the file and should now be implemented without using the database values

additional fix for numeric elements for which the min and max values were not passed on to the instrument

1. LINST instruments should work... more specifically LINST instruments with pages, pages should load properly

### DIFF
--- a/php/libraries/NDB_BVL_Instrument_LINST.class.inc
+++ b/php/libraries/NDB_BVL_Instrument_LINST.class.inc
@@ -23,18 +23,15 @@ namespace Loris\Behavioural;
  */
 class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
 {
-    /**
-     * FIXME: LINST instruments should parse the full name
-     * and the subtest names from the linst file, not select
-     * them from the database.
-     */
-    use \LegacyInstrumentTrait;
-
     public $InstrumentType = 'LINST';
 
     public $LinstQuestions = [];
 
     public $LinstLines = [];
+
+    protected $subtests = [];
+
+    protected $fullName;
 
     /**
      * Sets up the variables required for a LINST instrument to load
@@ -160,7 +157,9 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
             // dealing with _status==not_answered easier, because the rules for
             // enforcing range are independent of the rules for enforcing
             // required.
-            if ($this->LinstQuestions[$elname]['type'] === 'numeric') {
+            if (isset($this->LinstQuestions[$elname])
+                && $this->LinstQuestions[$elname]['type'] === 'numeric'
+            ) {
                 // Cast everything to a double so that < and > work as
                 // expected
                 $value = doubleval($elements[$elname]);
@@ -391,7 +390,6 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
     {
         if (file_exists($filename) || $base64 === true) {
             $this->InstrumentType = 'LINST';
-            $db = \Database::singleton();
 
             if (!isset($this->form)) {
                 $this->form = new \LorisForm();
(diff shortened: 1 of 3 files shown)


php/libraries/NDB_BVL_Instrument_LINST.class.inc
```diff
@@ -445,16 +441,13 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
...
-                        "SELECT Subtest_name
-                        FROM instrument_subtests
-                        WHERE Test_name=:testinst AND Description=:parsing",
-                        [
-                            'testinst' => $this->testName,
-                            'parsing'  => $parsingPage,
-                        ]
-                    );
+                    $pageDescription  = trim($pieces[2]);
+                    $pageName         = "page_$subtestCount";
```
driusan, line 445: subtestCount doesn't seem to ever be incremented.. I think this will make every page "page_1"

ridz1208: yeah working on another commit that also fixes a bunch of notices

ridz1208: done, tested and its working perfectly on my end
see BMI changes

php/libraries/NDB_BVL_Instrument_LINST.class.inc
```diff
@@ -503,15 +500,6 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
                     break;
 
                 case 'begintable':
-                    if ($addElements) {
```
driusan, line 506: Why is this removed?

ridz1208: that should have been done inanother PR butbasically it's because its an exact copy of `case 'endtable'`, I think it's a mistake anyways in the begin/end table logic. I will remove it from this PR and submit a PR specifically for it

driusan: I don't think it is a mistake.. I think it's because of something about the smarty template for instruments using a table, and starting a new table needing to close that one to start a new table, while ending a table needs to end the table and re-start a table for the smarty template, but I can't remember the details.

ridz1208: in that case their order of close first then open, or open then close should be reversed no ??

if not then you dont need to implement it twice in the switch which is what i did here... i just let it fall through

ridz1208: @driusan restored

driusan:
Conflicts in the changelog.

ridz1208:
@driusan rebased

raisinbread/instruments/bmi.linst
```diff
@@ -6,17 +6,20 @@ static{@}Window_Difference{@}Window Difference (+/- Days)
...
 static{@}{@}Note: Please enter measurements in Standard OR Metric units
 select{@}unit_classification{@}Choose unit of measurement{@}NULL=>''{-}'metric'=>'Metric'{-}'standard'=>'Standard'
+page{@}{@}Standard units
 static{@}{@}Standard units
 numeric{@}height_feet{@}Your height (feet) :
 select{@}height_feet_status{@}{@}NULL=>''{-}'not_answered'=>'Not Answered'
 numeric{@}height_inches{@}Your height (inches) :
 select{@}height_inches_status{@}{@}NULL=>''{-}'not_answered'=>'Not Answered'
-numeric{@}weight_lbs{@}Your weight (lbs) :
+numeric{@}weight_lbs{@}Your weight (lbs) :{@}50{@}500
```
driusan, line 15: Is this required or it just because you assume no one weighs outside of that range? If it's just to fix warnings we should fix it in the code, because numeric without any max/min is valid (like on line 13..)

ridz1208: it is valid, no warnings

I just try to keep diverse examples for different features in the RB dataset to catch failing feature when possible.

PR #7171, opened 2020-12-07
Update composer dependencies for PHP8

driusan:
This is a version of #7157 without the changes for PHPUnit which were merged separately.


driusan:
Not as easy as I was hoping, should to just take the dependencies now that phpunit is merged, should probably continue on with #7157

PR #7175, opened 2020-12-09
[GUI] Do not separate checkboxes from linebreaks

driusan:
I noticed while testing CCNA on the user_accounts page there
are often line breaks inserted between a checkbox and the
label at the end of the line when multiple checkboxes are used
in a group (because of the number of sites, the "Examiner At" spans
multiple lines while it doesn't in raisinbread/most projects.)

This ensures that there's no linebreaks between a checkbox
in LorisForm and the label that goes with it. I can't think
of any circumstances where that would be desirable behaviour.


driusan:
@cmadjar @ridz1208 This is just something annoying I noticed while doing the CCNA manual testing, but should we assign it? It's mostly noticeable on CCNA because the large number of sites causes line wraps.

maltheism:
Hi @driusan, https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nobr says `<nobr>` is obsolete  but it does currently support all browsers. I'm wondering if surrounding the label & checkbox with a <span style="white-space: nowrap;">  would be better?

php/libraries/LorisForm.class.inc
```diff
@@ -1584,9 +1584,15 @@ class LorisForm
         if (isset($el['disabled']) || $this->frozen) {
             $disabled = 'disabled';
         }
-        return "<input name=\"$el[name]\" type=\"checkbox\" $checked $value "
+        /// XXX: There seems to be a problem when using &nbsp; to separate the
+        //  checkbox from the label. Both Firefox and Chrome will still put a
+        //  linebreak between the space and the checkbox. Instead, we wrap it
+        //  in a <nobr> tag which avoids the label from being separated from the
+        //  label. We don't wrap the whole label in it because if it's a long label
```
ridz1208, line 1591:
```suggestion
        //  in a <nobr> tag which avoids the label from being separated from the
        //  checkbox. We don't wrap the whole label in it because if it's a long label
```

?

cmadjar:
@driusan can you work on it?

cmadjar:
@CamilleBeau could you please review/test? Thank you!

CamilleBeau (approved):

CamilleBeau:
looks good!

cmadjar:
@driusan ready for you I guess.

PR #7179, opened 2020-12-09
[instruments] Add instrument data dictionary to module

driusan:
This extracts the part of PR#6936 related to instruments'
data dictionary into a new PR for reviewability.

LINST instruments' dictionaries are creating while parsing the
instrument file.

PHP instruments may use the trait LorisFormDictionaryImpl
in order to implement the data dictionary using the same
mechanism that lorisform_parser.php uses to populate the
parameter_type table (or may hardcore the dictionary themselves
in order to avoid the overhead of parsing LorisForm.) The reason
this is done with a trait instead of being implemented directly
in the NDB_BVL_Instrument class is that there's no reason to assume
for new instruments that the dictionary is generated by the form.
It's just as reasonable to write a new instrument by doing the opposite:
starting with the dictionary and generating the form based on the
dictionary.

# Testing Instructions
1. Use the LorisFormDictionaryImpl in a variety of PHP instruments
2. Print the data dictionary from the getDataDictionary method and ensure that it looks correct
3. Do the same for LINST instruments with a variety of field types.


modules/instruments/php/module.class.inc
```diff
@@ -146,5 +146,37 @@ class Module extends \Module
...
+     * @param \LORIS\LorisInstance $loris The loris instance whose dictionary
+     *                                    should be retrieved
+     *
+     * @return \LORIS\Data\Dictionary\Category[]
+     */
+    public function getDataDictionary(\LORIS\LorisInstance $loris) : iterable
+    {
+        $DB = $loris->getDatabaseConnection();
+
+        $rows = $DB->pselectCol("SELECT Test_name FROM test_names", []);
```
ridz1208, line 163: if the instrument uses `getFullName()` the test name would not be in the test_names table no ?

ridz1208: nvm you would still need it for flag and battery and what not

driusan: I think the test_names is the closest we have to an "instruments" table to list the installed instruments.. as you say, flag joins against it and I don't think we support anything showing up on the instrument_list page unless it's in the test_names table. (It's also how the instrument manager gets a list of instruments..)

modules/instruments/php/module.class.inc
```diff
@@ -146,5 +146,37 @@ class Module extends \Module
...
+            try {
+                $inst   = \NDB_BVL_Instrument::factory($testname, "", "");
+                $cat    = new \LORIS\Data\Dictionary\Category(
+                    $testname,
+                    $inst->getFullName()
+                );
+                $fields = $inst->getDataDictionary();
+                $dict[] = $cat->withItems($fields);
+            } catch (\LorisException $e) {
+                error_log($e);
```
zaliqarosli, line 176: are we okay with error logging? i feel like we usually just throw the exception

php/libraries/NDB_BVL_Instrument_LINST.class.inc
```diff
@@ -461,6 +471,43 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
...
+                            new DictionaryItem(
+                                'Examiner',
+                                'Examiner',
+                                $scope,
+                                // Should this be an enum of examiners?
+                                new StringType(255),
+                                new Cardinality(Cardinality::SINGLE),
+                            ),
+                        ]
+                    );
```
zaliqarosli, line 510: would all this above be better placed after the call to  _addMetadataFields() on line 520  like how its organized for the other fields?

modules/instruments/php/module.class.inc
zaliqarosli, line 176: are we okay with error_logging? i feel like we usually just throw the exception

zaliqarosli (commented):
i think this all most looks good and makes sense. i don't have a clear understanding of how it should be implemented for php instrument but for LINST, its all good

zaliqarosli:
could we maybe add these implementations to the raisinbread PHP instruments to provide examples on how to use?

modules/instruments/php/module.class.inc
ridz1208, line 163: Should probably use getAllInstruments instead (and fix it to query the proper Full name)
https://github.com/aces/Loris/blob/main/php/libraries/Utility.class.inc#L459

modules/instruments/php/module.class.inc
driusan: the problem with letting the exception be thrown here is that if any single instrument is broken, it breaks the entire instruments module and DQT and data dictionary module.

driusan:
I'm not sure what you mean, it's already had to be updated in RB in order for the tests to pass https://github.com/aces/Loris/pull/7179/files#diff-a841cfae1260da57edf6c1a81993af0d08d52456fba1b971fa6696c5832895b3R13 ?

php/libraries/NDB_BVL_Instrument_LINST.class.inc
driusan: I'll look into it, I think there was originally a reason it had to be here but I can't remember what it was. (I think maybe the initial version depended on testName but that doesn't seem to be the case anymore?)

zaliqarosli:
@driusan step 1 of your testing instructions says

> Use the LorisFormDictionaryImpl in a variety of PHP instruments

What does this mean? How do you use it? maybe this can be answered by "using" the LorisFormDictionaryImpl in the php instrument under https://github.com/aces/Loris/tree/main/raisinbread/instruments.
By adding this as part of this PR, our raisinbread example instruments can utilize this new features whilst at the same time being available to be used to test how the functionality works for PHP instruments

you said that "it's already had to be updated in RB". so does this mean that If I grab lets say aosi and print its data dictionary by calling the getDataDictionary method, that it would work? i am thinking no because aosi doesn't (on this PR branch) have the getDataDictionary method defined anywhere right? or am i totally misunderstanding?

php/libraries/LorisFormDictionaryImpl.class.inc
```diff
@@ -0,0 +1,184 @@
...
+                    $items,
+                    $this->_getDataDictionaryGroup(
+                        $prefix,
+                        $cat,
+                        $element['elements'],
+                        $label,
+                    ),
+                );
+                continue;
+            }
```
ridz1208, line 76: Unless I'm mistaken this handles tables in instruments? 

It seems like here you break it down and create the dictionary item for each cell in the table. that's okai and I think it's the status quo...

HOWEVER I always wondered if we could tackle this differently and this might be the occasion to do it.
The label for a cell in the table should be the concatenation of the column header + row label. do you think this is the right place to do it ?

ridz1208 (commented):
Give that your implementation of the `getDataDictionary()` function was done in a trait, I assume you want that to be a temporary/fallback solution for existing instruments... Does that mean that the best practice is for PHP instruments to define their own dictionaries ?

if that is the case could you maybe update 2 RB instruments as examples? one with the trait (single line addition) and one properly ? ~like BMI if it could be done the right way would be helpful~ (Edit: BMI is LINST, another instrument maybe ?)

driusan:
@zaliqarosli Okay, I see what you mean now. I thought the RB instruments were being used by the test suite and would need the line to pass, but it looks like testtest is outside of the RB instruments.

I updated the PHP raisinbread instruments to use the trait. They can be used for testing, but it would be better to use it on a variety of real instruments to make sure the dictionaries are right/reasonable.

driusan:
@ridz1208 it's not a question of one being "right" and one being "wrong". There are 2 completely reasonable things to do if you were to start from scratch:

1. Build the form and generate the dictionary from the form
2. Define the dictionary and build the form from the dictionary.

It's abstract in order to not make any assumptions, and the trait is just to simplify the first case (which is how all of our instruments are currently done). I don't think we should force one way or the other.

ridz1208:
@driusan okai, if there is now right/wrong solution then I'm okai with the current implementation. I would have added the code in the instrument class but I see how this makes less assumptions and might be less confusing to users in the longrun... it does hhowever need to be added to the "how to create instruments" instructions

php/libraries/LorisFormDictionaryImpl.class.inc
driusan: This is based on/done to be the same way as the lorisform_parser script in order to ensure that the old and new dictionaries are the same.

Maybe it would be a good idea to leave it this way initially, and then re-visit once #6936 is merged? That would make it easier to see the outcome of any changes.

ridz1208: makes sense, note for the future

php/libraries/LorisFormDictionaryImpl.class.inc
```diff
@@ -0,0 +1,184 @@
...
+                $label = trim(preg_replace('/\s+/', ' ', $groupLabel));
+            }
+
+            if ($element['type'] === 'group') {
+                $items = array_merge(
+                    $items,
+                    $this->_getDataDictionaryGroup(
+                        $prefix,
+                        $cat,
+                        $element['elements'],
```
ridz1208, line 71:
```suggestion
                        $element['elements'] ?? [],
```

to fix the phan fail ?

modules/instruments/php/module.class.inc
driusan: It is using the full name: https://github.com/aces/Loris/pull/7179/files#diff-e52203453b6573a6a78d0ecbf197ba50a8a893a8d70bb452dac8788d8d19b9ddR171 .. the full name from that Utility function is unreliable. It would be incorrect for ie. LINST instruments, or any instrument not using LegacyInstrumentTrait. Do you still want me to change it to call the utlity function/use array_keys on it to get the test_name for the factory call? I can, but it doesn't seem worthwhile.

ridz1208: No i was thinking modifying the utility function to get the correct Full names instead

php/libraries/NDB_BVL_Instrument_LINST.class.inc
driusan: I've moved this code into the addMetadataFields itself so that it properly handles the conditionals.

modules/instruments/php/module.class.inc
driusan: I changed it to use the Utility function to avoid the database call, but I don't want to change the behaviour of the Utility function in this PR since it's (probably?) used other places in LORIS

zaliqarosli (approved):
this all looks good to me

ridz1208 (approved):

modules/instruments/php/module.class.inc
ridz1208: https://github.com/aces/Loris/pull/7398

PR #7180, opened 2020-12-09
Add AccessibleResourceFilter for filtering for data framework

driusan:
This adds a standard AccessibleResourceFilter which can be used
to filter out data for provisioners dealing with resource types
that use the AccessibleResource interface in LORIS. Any resource
which the user accessing the provisioner does not have access
to is filtered from the results.

### DIFF
--- a/src/Data/Filters/AccessibleResourceFilter.php
+++ b/src/Data/Filters/AccessibleResourceFilter.php
@@ -1,0 +1,30 @@
+<?php
+namespace LORIS\Data\Filters;
+
+/**
+ * The AccessibleResouceFilter is a data filter which enforces access permissions
+ * on entities which implement the \LORIS\StudyEntities\AccessibleResource interface.
+ *
+ * It filters out any rows which the user does not have access to.
+ *
+ * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ */
+class AccessibleResourceFilter implements \LORIS\Data\Filter
+{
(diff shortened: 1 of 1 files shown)


HenriRabalais (commented):
The code looks good to me. I guess my major questions would be:
1. is there a current use case for this that I can test? 
2. Is this a generic filter that is meant to replace the `hasAccess()` (I think that's what it's called) function?
3. If so, should we be making such a generic filter? My fear is that it becomes a catch all (unless that's the point) for data filters that should be more rigorously standardized.

HenriRabalais:
@driusan For 1 +2. Okay, that makes sense. I see now the implementation in the Candidate and User Libraries.

For point 3: It seems like the other filters implemented by the dataprovisioner have a specific filtering purpose in mind, whereas this one is very customizable. I'm just wondering if there's any concern of these functions being filled with checks that are replicable and could belong in their own filter. For example, both the Candidate and User libraries seem to do project and center matching within `isAccessibleBy`.

driusan:
for 3: This one is intended to specifically filter based on the criteria "do you have permission to access this resource?" The entity type itself is the only one who knows what criteria that should be based on.. for arbitrary DB rows returned by a DBRowProvisioner, I agree that it shouldn't be used arbitrarily, it's intended to be used for classes that do have specifically defined access criteria.

HenriRabalais (approved):

PR #7183, opened 2020-12-10
Instrument class and LINST class cleanup

ridz1208:
## Brief summary of changes
1. Deprecation of `begintable` and `endtable` in LINST instruments. these elements do not appear to be used anywhere and when tested they only seemed to corrupt the HTML without bringing any advantages. The instrument template automatically converts groups to tables (behaviour that should be changed in the future) so removing these elements will just bring LINST to par with PHP instruments for now

2. Added a bunch of PHP docs on variables not having them

3. Deletion of the `dateTimeFields` and `monthYearFields` class arrays. Projects still maintain off and on these arrays which are unused anywhere. These arrays are being populated but their usage died with the LorisForm replacing quickform. [Here is an example of where they were used](https://github.com/aces/Loris/blob/53dd32dcb657e6f7199a63be6acfc8307b6e8efe/php/libraries/NDB_BVL_Instrument.class.inc#L537)

4. Deletion of the unused `_getDatabaseDate()` function

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1129,21 +1129,6 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     abstract function getSubtestList(): array;
 
     /**
-     * Marks an element in the lorisform object as being required (for
-     * use with the requiredIf rule)
-     *
-     * @param string $elementName the element to add to the required list
-     *
-     * @return void
-     */
-    function setRequired(string $elementName): void
-    {
-        if (!in_array($elementName, $this->form->_required)) {
-            $this->form->_required[] = $elementName;
-        }
-    }
-
-    /**
      * This runs the XIN rules on all the elements on the current page to ensure
      * that no rules were violated.
      *
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -29,11 +29,41 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     public $form;
 
+    /**
+     * Variable defining the type of form in use
+     *
+     * Note: this seems to always be defined as 'XIN'
+     *
+     * @var string
+     */
     public $formType;
 
+    /**
+     * Array holding the date format
+     *
+     * @var array
+     */
     public $dateOptions;
+
+    /**
+     * Type of instrument. Can be 'normal' or 'DirectEntry'
+     *
+     * @var string
+     */
     public $DataEntryType;
+
+    /**
+     * Array holding XIN rules
+     *
+     * @var array
+     */
     public $XINRules;
+
+    /**
+     * Boolean enabling/disabling debugging statements for XIN rules
+     *
+     * @var boolean
+     */
     public $XINDebug;
 
     /**
@@ -102,21 +132,6 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     var $_requiredElements = [];
 
-    /**
-     * Array of date fields to be processed into QuickForm dates
-     * NOTE: When abstracting the _saveValues() method in the child <instrument>
-     * class, add only the dates from the current page to the array using the
-     * instruments switch($page) otherwise you'll end up overwriting (NULLing)
-     * all other dates in the table that appear on other pages (if they are
-     * listed in the array of date fields) - note by @dariovins
-     */
-    public $dateTimeFields = [];
-
-    /**
-     * Array of date fields that contain only month year fields in frontend with
-     * a static value for the date
-     */
-    public $monthYearFields = [];
     /**
      * Array of column names to be ignored by the double data entry conflict
      * detector.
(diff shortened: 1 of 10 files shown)


php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -28,11 +28,41 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     public $form;
 
+    /**
+     * Variable defining the type of form in use
+     *
+     * Note: this seems to always be defined as 'XIN'
```
ridz1208, line 34: @driusan should we set this to 'XIN' in here and avoid that every project sets it in every instrument ?

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -270,11 +285,9 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             $base . "project/instruments/$instrument.linst"
         );
         // Add rules only if they exist.
-        if (file_exists($base."project/instruments/$instrument.rules")) {
```
ridz1208, line 273: removed the `if file exists` because that check is done at the entry of the function but also has an `|| base64` check attached to it. (see instrument_LINST class)

php/libraries/NDB_BVL_Instrument.class.inc
driusan: That seems to be an assumption.. I think it's better to be explicit. Though to be honest, I'm not sure what this is used for. The only place I see it used in the code is:

```
        //XIN specific functionality
        if ($this->formType=="XIN") {
            $this->_nullStatus($values);
        }
```

driusan:
I don't think 1. is entirely true. By the tags ending the table and creating a new one, it allows the new table to have a different number of columns than the wrapping table since it's a new table.. but you're right that I don't think they're used by anyone since the instrument builder doesn't support them, so I think we can still remove them.

ridz1208:
@driusan I really don't mind fixing them but from all my attempts, the only real fix was to remove them... anything else resulted in emppty spaces, broken tables and broked DIVs... your call

driusan:
no, it's fine, go ahead and remove them. I'm pretty confident no one is using them.

cmadjar:
@ridz1208 can you please rebase?

@driusan once rebased can you review it?

ridz1208:
@cmadjar @driusan rebased

php/libraries/NDB_BVL_Instrument_LINST.class.inc
```diff
@@ -577,26 +576,15 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
...
-                            "</td></tr></table><table><tr><td>&nbsp;",
-                            ""
-                        );
-                    }
+                    error_log(
+                        "DEPRECATION MESSAGE: The use of `begintable` and ".
+                        "`endtable` elements of LINST instruments has is no longer ".
+                        "necessary. The $this->testName LINST instrument should ".
+                        "be modified to remove these elements as they will be ".
+                        "completely deprecated in a future version of LORIS"
```
driusan, line 586: I don't think this error message is good.

1. There's a typo
2. It's not that they're no longer necessary, it won't just magically work right if you delete the, it's that they're no longer supported..

driusan:
I don't think I understand points 3 and 4. Isn't the dateTimeFields used for converting dates to the proper format.

`getDatabaseDate` is likely used in a lot of project instruments. It might be a good idea to keep it around for a bit even if it's not called by the base code so that they don't break when upgrading.

php/libraries/NDB_BVL_Instrument_LINST.class.inc
ridz1208: it does work right if you delete them.

basically without them, everything is fine. with them everything is broken!

ridz1208:
> I don't think I understand points 3 and 4. Isn't the dateTimeFields used for converting dates to the proper format?

These variables are not called on anywhere in the code, they are defined but never used

> getDatabaseDate is likely used in a lot of project instruments. It might be a good idea to keep it around for a bit even if it's not called by the base code so that they don't break when upgrading.

When did we start catering to functions that projects use but are not used anywhere in LORIS?? That being said, I'll run a quick check and see the impact of deleting it on projects.

ridz1208:
@driusan for `_getDatabaseDate ` 
 - IBIS only has a single usage and overrides the function either way
 - CCNA, QPN, CBIGR no usages

driusan:
Okay, I thought you were deleting the references in this PR, but looking a little closer it looks like all the places you're doing that are just setting it if it's not set.

If IBIS doesn't have references to it anymore, I doubt newer projects do. I thought it used to be required to call it and expected there to be hundreds of references which we should give more time to deprecate/update. If there's only one, I'm fine with getting rid of it right off the bat.

php/libraries/NDB_BVL_Instrument_LINST.class.inc
driusan: How about something like:

```
`begintable` and `endtable` elements are deprecated.
$this->testName LINST instrument should be
 modified to remove these elements as they will be
completely removed in a future version of LORIS.
```

ridz1208:
@nicolasbrossard I searched the github repo for `_getDatabaseDate` could you do a quick search directly on the code to confirm ?

nicolasbrossard:
@ridz1208 
```
[ibis@nihace assembly]$ cd /var/www/loris/project/instruments/
[ibis@nihace instruments]$ fgrep -r _getDatabaseDate *
[ibis@nihace instruments]$ 
```
so yeah, all instances are gone.

PR #7186, opened 2020-12-10
Update phan to 3.2 and fix errors which were identified

driusan:
With this, we can finally run "composer install" with php8.

Note that phpmd still doesn't work with php8 (see phpmd/phpmd#853),
so our CI still can't be updated to test under php8 yet and `make
checkstatic` will fail, but `make` and `make dev` now work.

### DIFF
--- a/modules/candidate_parameters/ajax/formHandler.php
+++ b/modules/candidate_parameters/ajax/formHandler.php
@@ -281,7 +281,7 @@ function editFamilyInfoFields(\Database $db)
     $relationship     = isset($_POST[$relationshipType]) ?
         $_POST[$relationshipType] : null;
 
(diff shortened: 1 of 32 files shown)


modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -281,7 +281,7 @@ function editFamilyInfoFields(\Database $db)
     $relationship     = isset($_POST[$relationshipType]) ?
         $_POST[$relationshipType] : null;
 
-    while ($siblingCandID != null ) {
+    if ($siblingCandID != null ) {
```
driusan, line 284: This was an infinite loop, changing it to an `if` maintains the side-effect that would have happened inside the loop if it was ever entered.

PR #7199, opened 2020-12-14
[NDB_Factory] Remove testing mode

driusan:
The "Testing" mode in NDB_Factory complicates our code
and frequently confuses phan. It appears to not be used
by our testing suite, and the test suite often explicitly
sets it to not-test mode in order to get real connections,
and uses setDatabase/setConfig/etc to inject mocks rather
than using the factory's internal test mode. This removes
the mode in order to simplify our code.


kongtiaowang:
@driusan This commit should be help.
https://github.com/kongtiaowang/Loris/pull/300/commits/937eab907f6110776e69f212ac2344b94523b4b0

driusan:
@kongtiaowang Now that this is passing Travis can you review? It's blocking something else that I was trying to do that I couldn't get phan to pass because of the (previous) convoluted return types in different paths.

driusan:
@kongtiaowang can you please review this now that it's passing Travis. It's blocking other work.

driusan:
@kongtiaowang This is still preventing work on LORIS from going forward. Please review.

kongtiaowang (approved):
I have no problem with this PR. 
LGTM.

PR #7228, opened 2020-12-16
Add setup php step to GitHub Actions workflow

driusan:
This is the same as #7227 but fixes the build on the main branch instead of the release branch.


driusan:
Will be fixed by #7234 instead

PR #7231, opened 2020-12-16
[DICOM archive] Fix patient name and patient ID hidden logic of the menu filter

### LINKED ISSUE #7208: [23.0][DICOM archive] Patient name should not be shown in result table
Issue originally reported by @nicolasbrossard on CCNA Repo during candidate release testing of LORIS 23.0.2
aces/CCNA#4289
___
As user testadmin, I accessed the Configuration module and set both the patient ID and patient name regex to `JGH0174`. This should basically hide the patient ID and patient name for every candidate except this guy. I accessed the DICOM archive module and observed this in the result table:

- the patient ID and patient name column of the result table were visible for `JGH0174` 
- for all other candidates , the patient ID was set to `INVALID - HIDDEN` but the patient name was still visible (it should also have been set to `INVALID - HIDDEN`).

When accessing the details of a specific archive, the patient ID and patient name were correctly displayed or hidden so the bug seems to occur only in the result table.

cmadjar:
## Brief summary of changes

The logic of the HIDDEN patient name was incorrect in the DICOM archive menu. This fixes the logic.

* Resolves #7208

### DIFF
--- a/modules/dicom_archive/jsx/dicom_archive.js
+++ b/modules/dicom_archive/jsx/dicom_archive.js
@@ -157,6 +157,7 @@ class DicomArchive extends Component {
       {label: 'TarchiveID', show: false},
       {label: 'SessionID', show: false},
       {label: 'CenterID', show: false},
+      {label: 'IsPhantom', show: false},
     ];
 
     return (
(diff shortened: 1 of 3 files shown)


nicolasbrossard:
Code looks good. I tested your PR on my VM and it fixes the bug. 👍

nicolasbrossard (approved):

PR #7233, opened 2020-12-16
[23.0][document_repository] Fix for contents displayed at root

jesscall:
## Brief summary of changes
Related to aces/CCNA#4302

Root directory of the doc repo should only contain files that are located at root. All other files will be in subfolders.
Can see all files by filtering globally. 

This was a design decision made by Zal and Derek a while ago but was reverted in #6408.

Please see @zaliqarosli 's comment [here](https://github.com/aces/Loris/pull/6408#issuecomment-745620992)

### DIFF
--- a/modules/document_repository/jsx/docIndex.js
+++ b/modules/document_repository/jsx/docIndex.js
@@ -86,7 +86,7 @@ class DocIndex extends React.Component {
         });
         this.setState({
           global: false,
-          tableData: id == 0 ? filterData : fillData,
+          tableData: fillData,
           childrenNode: myJson['subcategories'],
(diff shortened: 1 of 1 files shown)


driusan:
@jesscall can you double check the test plan too to make sure it's accurate and we don't bounce back and forth on this?

jesscall:
Seems to be correct:

```
10. Check that when performing a search without global filtering, the resulting files which are under the current category are shown in a list format. [Manual Testing]
11. Check that when performing a search with global filtering, all files matching the filtered case (regardless of category) are shown in a list format. [Manual Testing]
```

zaliqarosli (approved):
thanks for making this PR!

PR #7236, opened 2020-12-16
[instrument_manager] Fix write permission

### LINKED ISSUE #7129: [23.0][instrument_manager] Write permissions not being used
**Describe the bug**
The `instrument_manager_write` permission is not really being used. If you do not have the permission, you can still see and access the upload tab. 

**To Reproduce**
1. Go to `Instrument Manager`
2. Remove `instrument_manager_write`, and refresh the module. Nothing changes.

For permissions to be more granular and affect the usage of the module.

CamilleBeau:
## Brief summary of changes
This PR makes it so that users with the instrument_manager_write permission are able to upload on instrument manager, and people without this permission do not have access to the upload tab. 

1. Access the instrument manager page with only the view permission. Make sure that you only see the "Browse" tab, and not the "Upload" tab. 
2. Attempt to go directly to the upload tab by adding /instrument_manager/?#upload to your base link. Make sure that the page loads with "You do not have access to this page"
3. Give yourself the instrument_manager_write permission.
4. Make sure that you now have access to the upload tab, either by reloading the page and clicking on the tab, or by navigating directly to the link.
5. Attempt to upload a file. Ensure that you are not given the "forbidden" error (you may get other errors though).

* Resolves #7129

### DIFF
--- a/modules/instrument_manager/jsx/instrumentManagerIndex.js
+++ b/modules/instrument_manager/jsx/instrumentManagerIndex.js
@@ -110,7 +110,7 @@ class InstrumentManagerIndex extends Component {
     }
 
     const feedback = () => {
-      if (!this.state.data.caninstall) {
+      if (!this.state.data.caninstall && this.props.hasPermission('instrument_manager_write')) {
         return (
           <div className='alert alert-warning'>
             Instrument installation is not possible given the current server
--- a/modules/instrument_manager/jsx/instrumentManagerIndex.js
+++ b/modules/instrument_manager/jsx/instrumentManagerIndex.js
@@ -103,8 +103,11 @@ class InstrumentManagerIndex extends Component {
 
     const tabs = [
       {id: 'browse', label: 'Browse'},
-      {id: 'upload', label: 'Upload'},
     ];
+
+    if (this.props.hasPermission('instrument_manager_write')) {
+      tabs.push({id: 'upload', label: 'Upload'});
+    }
 
     const feedback = () => {
       if (!this.state.data.caninstall) {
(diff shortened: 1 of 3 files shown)


driusan:
shouldn't this go to the 23 branch?

jesscall:
@CamilleBeau Dave is right, please rebase! :)

CamilleBeau:
My bad, rebased!

modules/instrument_manager/jsx/instrumentManagerIndex.js
```diff
@@ -103,11 +103,14 @@ class InstrumentManagerIndex extends Component {
...
-      {id: 'upload', label: 'Upload'},
     ];
 
+    if (this.props.hasPermission('instrument_manager_write')) {
+      tabs.push({id: 'upload', label: 'Upload'});
+    }
+
     const feedback = () => {
-      if (!this.state.data.caninstall) {
+      if (!this.state.data.caninstall && this.props.hasPermission('instrument_manager_write')) {
```
driusan, line 113: out of curiousity is there a reason to do it this way instead of incorporating it into the "caninstall"?

CamilleBeau: I didn't want to put it in the function `isQuatUserConfigured` which is used for `caninstall` in instrument_manager.class.inc because it seemed specific to whether the user is configured properly in the sql DB. But I could put it in that function if you think it's better, or add it to line 322 in instrument_manager.class.inc where caninstall is set.

driusan: I would have probably put it in toArray where caninstall is set to avoid needing to do permission checks in javascript, but it's fine the way it is since it looks like you need to do some js side checks anyways..

jesscall (approved):

PR #7237, opened 2020-12-16
[dashboard] Fix step 6 of TestPlan

CamilleBeau:
## Brief summary of changes
The dashboard TestPlan previously implied that there should be a separate issue tracker panel. This PR changes the wording of step 6 to refer to a row in the My Tasks panel rather than a separate panel. 

* Resolves https://github.com/aces/CCNA/issues/4275

### DIFF
--- a/modules/dashboard/test/TestPlan.md
+++ b/modules/dashboard/test/TestPlan.md
@@ -24,7 +24,7 @@
    Check that you are taken to that page (with the Selection Filter correctly set) when you
    click on the task.
    [Automate Test]
-6. Verify that the number of new issues in the My Tasks panel is accessible if and only if a user has the 'issue_tracker_reporter'
+6. Verify that the number of assigned issues in the My Tasks panel is accessible if and only if a user has the 'issue_tracker_reporter'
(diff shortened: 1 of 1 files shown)


ridz1208 (approved):

modules/dashboard/test/TestPlan.md
```diff
@@ -24,7 +24,7 @@
    Check that you are taken to that page (with the Selection Filter correctly set) when you
    click on the task.
    [Automate Test]
-6. Verify that the issue tracker panel is accessible if and only if a user has the 'issue_tracker_reporter'
+6. Verify that the number of assigned issues in the My Tasks panel is accessible if and only if a user has the 'issue_tracker_reporter'
```
driusan, line 27: I don't think "number of issues [..] is accessible" is clear.. is it that the number of issues shouldn't be displayed? The number should be displayed but not a clickable link? There should be a link but there should be a permission denied error when you click on it? Etc..

CamilleBeau: Ammended.

PR #7241, opened 2020-12-17
Fix PhanUndeclaredMethod in tests directory

driusan:
This fixes all of the "PhanUndeclaredMethod" class errors in the test directory. These are mostly caused by phan not knowing that mock objects have the methods of the class being mocked, and it needs to be specifically told so with a phan type assertion.

Calling methods that don't exist is a bad idea. We should unignore this error type, but first the errors need to be fixed. This PR starts with a large number of false positives caused by mock objects..

#### Testing instructions
1. Remove "PhanUndeclaredMethod" from `.phan/config.php`.
2. Verify that there are no errors from the test directory (there are other errors from our code)
3. Check that the tests still pass. (The behaviour of them shouldn't have changed.)

### DIFF
--- a/test/unittests/LorisForms_Test.php
+++ b/test/unittests/LorisForms_Test.php
@@ -157,7 +157,11 @@ class LorisForms_Test extends TestCase
         // first child of that is the rendered element
         $element = $html->documentElement->firstChild->firstChild;
 
+        // The DOMNode $element needs to be a DOMElement for
+        // hasAttribute to exist.
+        assert($element instanceof \DOMElement);
         $this->assertEquals($element->nodeName, "select");
+
         $this->assertTrue($element->hasAttribute("multiple"));
     }
 
@@ -404,12 +408,14 @@ class LorisForms_Test extends TestCase
      */
     function testAddElementSelect()
     {
-        $this->form = $this->getMockBuilder('LorisForm')
+        $form = $this->getMockBuilder('LorisForm')
             ->onlyMethods(['addSelect', 'addDate'])
             ->getMock();
-        $this->form->expects($this->once())
+        $form->expects($this->once())
             ->method('addSelect');
-        $this->form->addElement("select", "abc", "Hello");
+
+        '@phan-var \LorisForm $form';
+        $form->addElement("select", "abc", "Hello");
     }
 
     /**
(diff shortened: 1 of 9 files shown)


test/unittests/LorisForms_Test.php
```diff
@@ -403,12 +408,14 @@ function testAddPageBreak()
      */
     function testAddElementSelect()
     {
-        $this->form = $this->getMockBuilder('LorisForm')
+        $form = $this->getMockBuilder('LorisForm')
```
driusan, line 411: These need to be converted from a property to a simple variable for `@phan-var` to work on them. Adding the type hint isn't sufficient since phan picks up on the fact that it was assigned to a mock right before the call and uses that type instead of the phpdoc type, but you can't use `@phan-var` on a property..

test/unittests/NDB_BVL_FeedbackTest.php
```diff
@@ -61,7 +60,7 @@ protected function setUp(): void
      */
     public function testCreateFeedbackTypeWithInvalidName()
     {
-        $this->setExpectedException('LorisException');
+        $this->expectException('LorisException');
```
driusan, line 63: I don't know why this test was passing since setExpectedException was removed from phpunit, but phan picked it up here..

test/unittests/NDB_BVL_Instrument_Test.php
```diff
@@ -720,7 +722,7 @@ function testScoreElement()
         );
         $this->_instrument->addScoreColumn(
             "FieldName2",
-            null
+            "Field 2",
```
driusan, line 725: After adding the type to $this->_instrument phan picked up on the argument type mismatch. (neither argument to addScoreColumn in NDB_Page is nullable according to the phpdocs.)

test/unittests/NDB_BVL_Instrument_Test.php
```diff
@@ -1774,7 +1783,6 @@ function testToJsonParseSmartyDateTypeNotLorisForm()
             ],
             'NoResponse' => true
         ];
-        $this->_instrument->form = new \Candidate();
```
driusan, line 1777: After adding the type hint phan picked up on this too.. a candidate isn't a LorisForm. This is clearly wrong (and unused.)

driusan:
@kongtiaowang since this is just touching the test directory could you review?

test/unittests/NDB_Factory_Test.php
```diff
@@ -236,6 +239,12 @@ function testProject()
      */
     function testCandidate()
     {
+        $mockdb = $this->getMockBuilder("\Database")->getMock();
```
driusan, line 242: These tests were failing in the same way as #7199 so I stole the fix from there. I'm not sure why they're failing on this branch it seems to be unrelated, but was necessary for the tests..

laemtl (approved):

PR #7242, opened 2020-12-17
[Core] Download CSV issues

### LINKED ISSUE #7229: [Core] Download csv duplicate/non-matching rows on filtered content
**To Reproduce**
1. Go to the publications page
2. Filter using more than one criteria (ex: title and keyword value)
3. Download the CSV. The rows matching all criteria appear twice, the ones matching only some are not removed.

laemtl:
This PR solves 2 issues:
- when a row matches 2 filtering criteria, this row is included twice in the csv
- the rows matching the filtering criteria partially are included in the csv 

The bug was found on the publication module and probably affects all modules using the StaticDataTable.

## To test
Filter a data table with 2 criteria:
- check that the rows matching all criteria are not included twice in the csv
- check that the rows that match some but not all criteria are not included in the csv

* Resolves #7229

### DIFF
--- a/jsx/StaticDataTable.js
+++ b/jsx/StaticDataTable.js
@@ -432,7 +432,6 @@ class StaticDataTable extends Component {
 
         if (this.hasFilterKeyword(this.props.Headers[j], data)) {
           filterMatchCount++;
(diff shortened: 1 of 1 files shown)


maltheism (approved):
Passed manual testing

PR #7243, opened 2020-12-17
Remove legacy code to support QuickForm

driusan:
This removes some code that was in place to handle
differences between LorisForm and QuickForm. At this
point there's no excuse for instruments to not have
been migrated (and it's likely that at some point they
would have broken in a PHP upgrade anyways if they
weren't migrated to LorisForm yet.)

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2505,8 +2505,8 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             }
         } else if ($element['type'] === "date") {
             $element['options'] = [
-                'mindate'         => "1990-01-01",
-                'maxdate'         => "2000-12-31",
+                'mindate'         => $this->dateOptions['minYear'] . '-01-01',
+                'maxdate'         => $this->dateOptions['maxYear'] . '-12-31',
                 'RequireResponse' => false,
             ];
 
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -23,7 +23,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     /**
      * The form object for this instrument.
      *
-     * @var    LorisForm
+     * @var LorisForm
      */
     public $form;
 
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -24,7 +24,6 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      * The form object for this instrument.
      *
      * @var    LorisForm
-     * @access public
      */
     public $form;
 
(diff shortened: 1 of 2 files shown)


driusan:
@ridz1208 You feel like handling this? It's just some cleanup I noticed that's long overdue while cleaning up other stuff..

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2505,46 +2504,13 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
-                    $element['options'] = [
-                        'mindate' => $minYear . "-01-01",
-                        'maxdate' => $maxYear . "-12-31",
-                    ];
+            $element['options'] = [
+                'mindate'         => "1990-01-01",
+                'maxdate'         => "2000-12-31",
+                'RequireResponse' => false,
+            ];
+        }
```
ridz1208, line 2512: why are these values even there ?

driusan: I don't know, it's the body of the `if (get_class($this->form) === 'LorisForm') {}` part. I just removed the condition which changed the indentation.

ridz1208: yeah i saw, should we just use the dateoptions already in the file ?
https://github.com/aces/Loris/blob/80b1dfa92d01f5ab5e2ab9918016758f51dbd487/php/libraries/NDB_BVL_Instrument.class.inc#L854

ridz1208 (approved):

php/libraries/NDB_BVL_Instrument.class.inc
driusan: seems like a good idea. Want me to change it or just create an issue? It'll likely make the PR much larger since I assume some tests depend on it

ridz1208: hmm if it makes the PR alot larger forget it. it will be done eventually in a future check

driusan: after the tests pass with the current state I'll try it and see how much bigger it makes it. I could be mistaken, it's just an assumption.

driusan: Actually doing a `git grep` for those particular dates there seems to only be 2 tests that depend on it, I'll just fix it here

driusan: It's not as simple as it looks because the dateOptions isn't populated (or isn't populated yet..) when the tests are called. I'll just create an issue.

ridz1208: yeah its loaded in add metadata fields... we would have to move it to the constructor first.

what I'm more confused about is the fact that its not even checking if these options are already set, which means its overriding whatever is already working... ? how does that even work for projects ?

driusan: Have you tried the instrument form API endpoint on a project? I think this function was added to generate the JSON for that, and I doubt it's used anywhere else.

PR #7252, opened 2020-12-18
[instruments] Save value into UserID field

zaliqarosli:
## Brief summary of changes

Currently, nothing is ever saved into the UserID field of instruments. This PR fixes that by inserting into the field the user's Real_name.

Context: Project require to render the Name of the Data Entry personnel on the instrument page on 21.0-release.

1. On main branch, go to any instrument page. Complete an instrument and save. Check the Data column of flag table and the UserID column of the instrument table. You will see that the UserID data is empty / doesn't exist. There is a UserID column of flag that has data in it, however that is specific to the session, not the instrument data entry.

2. On this PR branch, repeat the above step and notice the UserID in the instrument table will now be populated with your user account username, as well as the UserID key in the JSON data of flag table Data column.

3. On this PR branch, run `php tools/single_use/SaveUserIDToInstrumentData.php` to back populate all the UserID fields with data queried from the history table. Run without the confirm flag to check that the correct data has been queried and printed. Run `php tools/single_use/SaveUserIDToInstrumentData.php confirm` to apply the changes.

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -767,6 +767,14 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             $this->_saveCandidateAge($values);
         }
 
+        // Save UserID, the data entry personnel, into flag Data and
+        // instrument table
+        $user = \NDB_Factory::singleton()->user();
+        if (! $user instanceof \LORIS\AnonymousUser) {
+            // Add UserID to $values
+            $values['UserID'] = $user->getUsername();
+        }
+
         //Convert select multiple elements into database storable values
         if (!empty($this->selectMultipleElements)) {
             foreach ($this->selectMultipleElements AS $elname) {
--- a/tools/single_use/SaveUserIDToInstrumentData.php
+++ b/tools/single_use/SaveUserIDToInstrumentData.php
@@ -47,7 +47,7 @@ $history = $DB->pselect(
               AND type='U'
             GROUP BY primaryVals
         ) h2 USING (primaryVals, changeDate)
-            GROUP BY CommentID",
+            GROUP BY CommentID, h1.userID",
     array()
 );
 // Loop and index results from history by testname
@@ -70,7 +70,10 @@ foreach(\Utility::getAllInstruments() as $testname => $fullName) {
     $table = null;
     if ($JSONData === false) {
         $table = $instrument->table;
-        if (!$DB->tableExists($table)) {
+        if (!$table) {
+            echo "The table name for instrument $testname cannot be found.\n";
+            continue;
+        } else if (!$DB->tableExists($table)) {
             echo "Table $table for instrument $testname does not exist in the Database.\n";
             continue;
         }
--- a/tools/single_use/SaveUserIDToInstrumentData.php
+++ b/tools/single_use/SaveUserIDToInstrumentData.php
@@ -1,0 +1,117 @@
+<?php
+/**
+ * This script is intended for a one-time use only to restore the value of the
+ * `UserID` column of instrument tables and the `UserID` key of the instrument
+ * JSON `Data` in the flag table.
+ * 
+ * This tool queries data from the `history` table and meaningfully imports
+ * data from its `userID` column back into the instrument and `flag` tables.
+ *
+ * Example use:
+ * $ php SaveUserIDToInstrumentData.php [confirm]
+
+ * PHP Version 7
+ *
+ * @category Tools
+ * @package  Loris
+ * @author   Zaliqa Rosli <zaliqa.rosli@mcin.ca>
+ * @licence  LORIS License
+ * @link     https://www.github.com/aces/Loris
+ */
+require_once __DIR__ . '/../generic_includes.php';
+
+$confirm = false;
+if (isset($argv[1]) && $argv[1] === 'confirm') {
+    $confirm = true;
+}
+
+echo "\n\nQuerying data...\n";
+$db      = \NDB_Factory::singleton()->database();
+// Query CommentIDs with DECS 'Complete', and
+// userID for latest changes made to the instrument data
+$history = $db->pselect(
+        "SELECT h1.tbl, h1.primaryVals AS CommentID, h1.userID
+            FROM history h1
+            INNER JOIN
+            (
+                SELECT primaryVals, MAX(changeDate) AS max_date
+                    FROM history
+                    WHERE primaryVals IN (
+                        SELECT primaryVals
+                        FROM history
+                        WHERE col = 'Data_entry_completion_status'
+                            AND primaryCols = 'CommentID'
+                            AND new = 'Complete'
+                    )
+                    GROUP BY primaryVals
+            ) h2
+                ON h1.primaryVals = h2.primaryVals
+                    AND h1.changeDate = h2.max_date
+                WHERE userID <> 'unknown'
+                    AND tbl <> 'flag'
+                    AND col <> 'CommentID'
+                GROUP BY h1.primaryVals, h1.changeDate, h1.userID",
+        array()
+);
+if (empty($history)) {
+    echo "\n\nThere is no data to import.\n";
+    die();
+}
+echo "\n\nThe following data can be imported into the instrument tables:\n";
+foreach ($history as $index => $row) {
+    $result = $index + 1;
+    echo "\n\nResult $result:\n";
+    print_r($row);
+
+    $table     = $row['tbl'];
+    $commentID = $row['CommentID'];
+    $userID    = $row['userID'];
+
+    // Extract old data from flag so we can update it with merge so that we
+    // don't overwrite it
+    $oldData = $db->pselectOne(
+        "SELECT Data FROM flag WHERE CommentID=:cid",
+        array('cid' => $commentID)
+    );
+    echo "\n\nOld Data: \n";
+    print_r($oldData);
+
+    // (The PDO driver seems to return null as "null" for JSON column types)
+    if (!empty($oldData) && $oldData !== "null") {
+        $oldData = json_decode($oldData, true);
+    } else {
+        $oldData = array();
+    }
+    $newData = array_merge(
+        $oldData ?? array(),
+        array('UserID' => $userID)
+    ); 
+    echo "\n\nNew Data: \n";
+    print_r(json_encode($newData));
+
+    if ($confirm) {
+        echo "\n\nImporting data into respective instrument tables and commentID flag entries...\n";
+        // Update instrument table
+        $db->update(
+            $table,
+            array('UserID' => $userID),
+            array('CommentID' => $commentID)
+        );
+
+        // Save the JSON to the flag.Data column.
+        //
+        // json_encode ensures that this is safe. If we use the safe wrapper,
+        // HTML encoding the quotation marks will make it invalid JSON.
+        $db->unsafeUpdate(
+            "flag",
+            array("Data" => json_encode($newData)),
+            array('CommentID' => $commentID)
+        );
+        echo "\n\nDone. $userID saved as UserID in $table and flag tables for CommentID $commentID.\n";
+    }
+}
+
+if (!$confirm) {
+    echo "\n\nRun this tool again with the argument 'confirm' to ".
+         "perform the changes.\n";
+}
--- a/tools/single_use/SaveUserIDToInstrumentData.php
+++ b/tools/single_use/SaveUserIDToInstrumentData.php
@@ -40,11 +40,11 @@ $result_count = 1;
 $history = $DB->pselect(
     "SELECT f.Test_name, f.CommentID, h1.userID
       FROM flag f
-      JOIN $hDB.history h1 on (f.CommentID=h1.primaryVals)
+      JOIN " . $DB->escape($hDB) . ".history h1 on (f.CommentID=h1.primaryVals)
       JOIN
         (
           SELECT primaryVals, MAX(changeDate) as changeDate
-            FROM $hDB.history
+            FROM " . $DB->escape($hDB) . ".history
             WHERE userID <> 'unknown'
               AND type='U'
               AND NOT (
--- a/tools/single_use/SaveUserIDToInstrumentData.php
+++ b/tools/single_use/SaveUserIDToInstrumentData.php
@@ -45,6 +45,9 @@ $history = $DB->pselect(
             FROM $hDB.history
             WHERE userID <> 'unknown'
               AND type='U'
+              AND NOT (
+                col='Data_entry' AND old IS NULL
+              )
             GROUP BY primaryVals
         ) h2 USING (primaryVals, changeDate)
             GROUP BY CommentID, h1.userID",
--- a/tools/single_use/SaveUserIDToInstrumentData.php
+++ b/tools/single_use/SaveUserIDToInstrumentData.php
@@ -26,88 +26,103 @@ if (isset($argv[1]) && $argv[1] === 'confirm') {
 }
 
 echo "\n\nQuerying data...\n";
-$db      = \NDB_Factory::singleton()->database();
-// Query CommentIDs with DECS 'Complete', and
-// userID for latest changes made to the instrument data
-$history = $db->pselect(
-        "SELECT h1.tbl, h1.primaryVals AS CommentID, h1.userID
-            FROM history h1
-            INNER JOIN
+
+$db = \NDB_Factory::singleton()->database();
+
+foreach(\Utility::getAllInstruments() as $table => $name) {
+    // Query entries in the history table for the latest change
+    // made to the Data column of the flag table, for every CommentID.
+    $history = $db->pselect(
+        "SELECT h1.primaryVals AS CommentID, h1.userID, JSON_MERGE_PATCH(
+            f.Data, CONCAT('{\"UserID\": \"', h1.userID, '\"}')
+         ) AS new_Data
+            FROM $table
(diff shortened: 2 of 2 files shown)


ridz1208:
but why???? it's tradition !!

driusan:
Why would the real name go in the UserID column?

zaliqarosli:
UserID column doesn't have a reference to the user table. it's set as just a varchar. Putting the full name and not the user table id will let the user's actual name to be rendered as is, without having to query the user table. i find this useful for linst instruments where you can just grab the value of the UserID directly. it's the same case with the UserID column of the flag table, except that the value there only contains the user's first name. in this case, its more information to have the full name

driusan:
The users table has a unique key named "UserID". The column is clearly supposed to be a reference to that and using a different column is confusing and will lead almost certainly lead to bugs down the line.

xlecours:
> UserID

The issue I am foreseeing would occur if a user's Full Name changes. We would need to update all the values in the flag table as well. But no one will remember that so we would need to guess who that was; hopefully the names will be similar.

zaliqarosli:
> > UserID
> 
> The issue I am foreseeing would occur if a user's Full Name changes. We would need to update all the values in the flag table as well. But no one will remember that so we would need to guess who that was; hopefully the names will be similar.

that is a good point!

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -744,6 +744,14 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             $this->_saveCandidateAge($values);
         }
 
+        // Save UserID, the data entry personnel, into flag Data and
+        // instrument table
+        $user = \NDB_Factory::singleton()->user();
+        if (! $user instanceof \LORIS\AnonymousUser) {
+            // Add User Full Name to $values
```
zaliqarosli, line 751:
```suggestion
            // Add UserID to $values
```

zaliqarosli: move to 23 branch

zaliqarosli:
hey @driusan @ridz1208 i made a script for myelineurogene to back populate the missing userIDs with data queried from the history table. is this something you guys think would be useful for loris core / other projects? if yes, i can add it to this PR. i wouldn't mind some help from loris core reviewing it either

driusan:
Yes, if we have the data and can restore it from the history table it seems like it would be a good idea to include the script

tools/single_use/SaveUserIDToInstrumentData.php
```diff
@@ -0,0 +1,117 @@
...
+$confirm = false;
+if (isset($argv[1]) && $argv[1] === 'confirm') {
+    $confirm = true;
+}
+
+echo "\n\nQuerying data...\n";
+$db      = \NDB_Factory::singleton()->database();
+// Query CommentIDs with DECS 'Complete', and
+// userID for latest changes made to the instrument data
+$history = $db->pselect(
```
zaliqarosli, line 32: @ridz1208 if you could take a look at this query, i would appreciate it!

ridz1208: @zaliqarosli i'm not sure I understand the need for the temporary table ? what is the role of `h2` here?

zaliqarosli: the h2 gives all the max changeDate values for each commentID, and then i use that max date to query the row in the table that has that changeDate

zaliqarosli: https://stackoverflow.com/a/12102216/10924787

zaliqarosli: i'm essentially trying to make sure that the userID i'm querying is from the last data entry done on the instrument

PapillonMcGill:
@ridz1208 @driusan Could you review? This is blocking fixes on Myelineurogene.
Thanks

zaliqarosli:
@PapillonMcGill its actually waiting on my to update, and then review by xavier

xlecours:
@zaliqarosli I did not tested it yet but I really like the way it looks. :)

zaliqarosli:
i made a couple of changes i noticed needed fixing @xlecours.

ive tested this on myelineurogene staging and it works as intended/expected. could we see if someone can test this on raisinbread data or maybe another project's data? @ridz1208 @driusan

ridz1208:
@pierre-p-s  please test asap

zaliqarosli:
@pierre-p-s let me know if you have questions or run into trouble

pierre-p-s:
On CCNA DB: The first part works but for the back population of the UserID field I am getting the error below:

Screen Shot 2021-01-27 at 11 04 33 PM
 
Any idea of what I am doing wrong please?

ridz1208:
@pierre-p-s can you provide your SQL version ?

pierre-p-s:
> @pierre-p-s can you provide your SQL version ?

`Server version: 10.1.46-MariaDB MariaDB Server`

ridz1208:
@zaliqarosli @xlecours any ideas ?

xlecours:
JSON_MERGE_PATCH was introduced in MariaDB 10.2.25, MariaDB 10.3.16 and MariaDB 10.4.5.

I am looking for the alternative

driusan:
I'm not sure what JSON_MERGE_PATCH is, but wouldn't an alternative be to do it in PHP? It would be slower but I don't think that would be a big concern for a one time script..

xlecours:
@zaliqarosli 

> I'm not sure what JSON_MERGE_PATCH is, but wouldn't an alternative be to do it in PHP? It would be slower but I don't think that would be a big concern for a one time script..

yes, it could. 

Sorry @zaliqarosli

tools/single_use/SaveUserIDToInstrumentData.php
```diff
@@ -0,0 +1,133 @@
...
+if (isset($argv[1]) && $argv[1] === 'confirm') {
+    $confirm = true;
+}
+
+echo "\n\nQuerying data...\n";
+
+$db = \NDB_Factory::singleton()->database();
+
+$result_count = 1;
+foreach(\Utility::getAllInstruments() as $table => $name) {
```
ridz1208, line 33: @zaliqarosli This line is incorrect. the `$table` is actually the `$testname` and the `$name` is actually the `$Full_name`

in order to get the `table` you need to instanciate the instrument and request the `table` variable !

A good example to follow is the fix_candidate_age tool in this case. Including the `tableExists()` and `columnExists()` checks in the script to make sure not to get an SQL error.
https://github.com/aces/Loris/blob/23.0-release/tools/data_integrity/fix_candidate_age.php#L46



You might also wanna add a check for JSONData instruments since they would not have a table in the database
https://github.com/aces/Loris/blob/23.0-release/tools/CouchDB_Import_Instruments.php#L139

ridz1208 (changes requested):

tools/single_use/SaveUserIDToInstrumentData.php
ridz1208: After Testing i also realized the same comment ID is being returned multiple times indicating that something might be wrong with the query.

tools/single_use/SaveUserIDToInstrumentData.php
```diff
@@ -0,0 +1,139 @@
...
+            $commentID = $row['CommentID'];
+            $userID = $row['userID'];
+
+            try {
+                $sessionInst = \NDB_BVL_Instrument::factory(
+                    $testname,
+                    $commentID,
+                    ''
+                );
+            } catch (Exception $e) {
```
ridz1208, line 111: is the only reason it would throw an exception that its not active ?

zaliqarosli: i guess so? i took this bit of code from fix_candidate_age.php

pierre-p-s:
The first part works correctly.

The script is taking a long time to run on the history table of CCNA

pierre-p-s:
Sorry it took a really long time to run the script.

It seems to be working for the most part, however I ran into an issue where the script stops because in a table on CCNA there is no UserID field. Otherwise I checked other instruments and everything seemed to be fine with the UserID entered correctly.

Screen Shot 2021-02-12 at 9 26 07 PM

ridz1208:
@zaliqarosli can you add the columnExists check ?

zaliqarosli:
@ridz1208 is there a way to check whether UserID key in Data column should exist? for instrument that don't use SQL tables

ridz1208:
@zaliqarosli you could by loading the instrument and parsing its form content but I would say thats an overkill. I think you can just add it to the data element by default.

ridz1208:
@pierre-p-s you can test again

pierre-p-s:
Tested again on CCNA. No visible PHP errors this time. 

However I noticed that on some candidates, the UserID in flag does not correspond to the one in the instrument table (I have 2 instances where UserID is lorisadmin in the instrument table but it is oveisiniki@gmail.com in the flag table).

Side note: on CCNA DB it takes around 14 hours to run the script. Maybe add a comment at the begining of the script for the projects to be aware of this.

ridz1208:
@pierre-p-s have you noticed any other users saved incorrectly ? or is it only sometimes lorisadmin ? have you made sure the entry in history for that instrument is not lorisadmin itself ?

pierre-p-s:
No, I have noticed this on 2 users. 

Wouldnt it save lorisadmin as well in the flag table if the UserID in the history table was lorisadmin?

zaliqarosli:
this is weird to me because the NDB_BVL_Instrument _save() function should be saving the same value to the instrument and flag table. could it be that the UserID value in flag was already there and the value wasn't overwritten? just thinking about possiblities

ridz1208:
@zaliqarosli... highly unlikely, there is no logic for that anywhere. its very weird

zaliqarosli:
@pierre-p-s is the UserID key in the Data object of flag table at the end of the JSON?

pierre-p-s:
> this is weird to me because the NDB_BVL_Instrument _save() function should be saving the same value to the instrument and flag table. could it be that the UserID value in flag was already there and the value wasn't overwritten? just thinking about possiblities

You are right we have an override on CCNA to save to the flag table.

tools/single_use/SaveUserIDToInstrumentData.php
```diff
@@ -0,0 +1,145 @@
+<?php
+/**
+ * This script is intended for a one-time use only to restore the value of the
```
zaliqarosli, line 3:
```suggestion
  * CAVEAT FOR PROJECTS: It can take up to a whole day to run this script. It took 14 hours to run the script on the CCNA DB during testing.
  *
  * This script is intended for a one-time use only to restore the value of the
```

ridz1208 (changes requested):
@driusan The fix here is functional and @zaliqarosli did her best to make the "recovery" script be compatible with other projects. I don't think we should delay this more. I think any project with overrides should make the effort to accomodate the script to their codebase/database

ridz1208 (approved):

tools/single_use/SaveUserIDToInstrumentData.php
```diff
@@ -0,0 +1,147 @@
...
+          SELECT primaryVals, MAX(changeDate) as changeDate
+            FROM $hDB.history
+            WHERE userID <> 'unknown'
+              AND type='U'
+              AND NOT (
+                col='Data_entry' AND old IS NULL
+              )
+            GROUP BY primaryVals
+        ) h2 USING (primaryVals, changeDate)
+            GROUP BY CommentID, h1.userID",
```
driusan, line 55: This query is fairly complex and I haven't looked at it closely enough yet to make sure it's correct, but the `$hDB` value should be quoted.

tools/single_use/SaveUserIDToInstrumentData.php
```diff
@@ -0,0 +1,147 @@
...
+    $confirm = true;
+}
+
+echo "\n\nQuerying data...\n\n";
+
+// Get history DB
+$DBInfo = $config->getSetting('database');
+$hDB = isset($DBInfo['historydb']) ? $DBInfo['historydb'] : $DBInfo['database'];
+
+$result_count = 1;
```
driusan, line 36: We're not allowed to have 0 problems?

I think this can be moved to before the foreach loop on ~line 94 and initialized to 0 if the increment is moved to the start of the loop body instead of the end..

ridz1208: result_count is only used for display on the front end, never to check how many problems there are... it doesnt make much of a difference

tools/single_use/SaveUserIDToInstrumentData.php
```diff
@@ -0,0 +1,147 @@
...
+        echo "\nImporting data into $testname...\n\n";
+
+        // Instantiate instrument at the session level in order to
+        // use _save() function
+        foreach ($idxHist[$testname] as $row) {
+            $commentID = $row['CommentID'];
+            $userID = $row['userID'];
+
+            try {
+                $sessionInst = \NDB_BVL_Instrument::factory(
```
driusan, line 114: This is likely why this script is taking so long. Can we try to use #6869? If that fixes the efficiency, we can either backport the fix to 23 so that it can be used here (slightly complicated/risky since I think it also depends on #6868), or retarget this fix to main for the next release (since it's been broken for so long anyways..)

tools/single_use/SaveUserIDToInstrumentData.php
driusan: no, it doesn't make much difference, it's just confusing when it's initialized at the top of the script to "1".

tools/single_use/SaveUserIDToInstrumentData.php
ridz1208: I'm pretty sure 90% of the wait time is the queries on history. I have ran that query separately and its infernal on large projects

driusan: I did some investigation on my sandbox (which has a much smaller history table), and if that's the case I think we can fix it by temporarily creating an index. I don't know if we would want to keep it or not since the table is frequently written and seldom read, but:

On my sandbox the first time I ran the query it took 45 seconds, the second time it took 40. So in the 40-45s range.

Running: `create index tmptest on history(primaryVals(255))` to add an index took 1.46s. After creating the index, the query took 3.3 seconds. That's down to about 5 seconds total for creating an index and selecting, from 40-45s without an index and assuming a similar proportion for CCNA that would be about 1.5-2h from 14 hours.

(Dropping the index after my tests was pretty much instantaneous..)

ridz1208: would we then recommend the index creation in the release notes ?

tools/single_use/SaveUserIDToInstrumentData.php
zaliqarosli: why does the $hDB value need to be quoted? just wondering because i dont get an error when the script runs the query as is

tools/single_use/SaveUserIDToInstrumentData.php
zaliqarosli: the result count doesn't really need to be there, but it helps to verify the numbers you are expecting when testing or running the script for real

tools/single_use/SaveUserIDToInstrumentData.php
ridz1208: in case there are SQL special characters i nthe DB name. its an issue we have had in the past with other projects

zaliqarosli: okayy, so by quote, do we mean sql backticks or actual quote

ridz1208: SQL backticks, there might even be a function for it in the database class... escape() i think

tools/single_use/SaveUserIDToInstrumentData.php
driusan: Can we test that my back of the envelope calculation is accurate first? If it is, that seems like a reasonable way to handle it since it's only a problem for large projects..

ridz1208: @driusan can you post the command you used to create the index here?

@pierre-p-s can you give it one last try with the index creation and tell us how long it takes for each operation ?

driusan: it's in my comment above.. `create index tmptest on history(primaryVals(255))`

pierre-p-s: I tried running the CREATE index command @driusan mentionned but after more than 4 hours it failed twice when the SQL server disconnected (ERROR 2013 (`HY000): Lost connection to MySQL server during query`)

pierre-p-s: related issue on redmine: https://redmine.cbrain.mcgill.ca/issues/18993

driusan: If it's taking 4 hours and then crashing that's more than my back of the envelope calculation had said for index creation, so we should probably just make a note in the release notes that it's slow for large, old projects since it's a 1 time use script.

driusan:
That repo isn't public and the link is a 404.

zaliqarosli:
@driusan 
Screen Shot 2021-05-20 at 3 11 40 PM

tools/single_use/SaveUserIDToInstrumentData.php
```diff
@@ -0,0 +1,147 @@
+<?php
```
ridz1208, line 1: before SQL

PR #7253, opened 2020-12-18
[Instrument] Date field required rule

laemtl:
By design a DateElement is required but is not displaying a red asterisk. 
This PR adds a rule to solve this.

Screenshot from 2020-12-18 16-21-37

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1914,6 +1914,12 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             "A Date, or Not Answered is required.",
             $name . "_date_group"
         );
+
+        $this->addRule(
+            $name . "_date_group",
+            $label . 'is required.',
+            'required'
+        );
     }
 
     /**


pierre-p-s:
The PR works however as you can see in the screenshot below, the second element is required as well but does not have an asterisk. Maybe it should be discussed during a meeting whether they should all have an asterisk or non of them should.

Screen Shot 2020-12-22 at 4 42 29 PM

laemtl:
@pierre-p-s As a user, I prefer to have asterisks. I don't have such issues on my side though, all the required selects have an asterisk if I add the required option with no XIN rule.

ridz1208:
@laemtl with your changes what happens if you want to make a date not required. At first sight in the code your date will always be required.

also for selects i'm assuming you are using the options array to pass in a required flag. does that not work for dates ?

laemtl:
@ridz I think this field is always required (you can't leave the select ANd the date field empty) that's the reason why I didn't use a required option. Let me know if I'm wrong.

The previous lines defined a XIN rules like so:
$this->XINRegisterRule(
    $name . "_date",
    array($name . "_date_status{@}=={@}"),
    "A Date, or Not Answered is required.",
    $name . "_date_group"
);

My change add a red asterisk to improve the UX.

ridz1208:
@pierre-p-s @laemtl can you schedule a meet between the 2 of you to figure out why and when asterisk show on the front end ? I dont think its normal (without overrides) that 2 projects get different results on the same version of loris.

laemtl:
Closing after discussion with @ridz1208 and @pierre-p-s . Red markers are not consistently used on other projects.

PR #7259, opened 2020-12-22
[instrument_list] Add project permissions

driusan:
This updates the instrument_list page to use the TimePoint's
isAccessibleBy function for access checking rather than manually
doing a site comparison for more consistency across LORIS. A result
of this is that project permissions for the timepoint are now enforced
on the instrument_list page.

The "or the same site of the candidate" is removed from the README
because:
1. That doesn't make sense if they have access to a different timepoint
   than the one being accessed.
2. The test plan doesn't agree and says you must have the timepoint's
   visit.

### DIFF
--- a/modules/instrument_list/README.md
+++ b/modules/instrument_list/README.md
@@ -21,11 +21,10 @@ entered.
 
 ## Permissions
 
-One of three conditions must be met to have access to the module:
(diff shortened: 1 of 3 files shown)


kongtiaowang (approved):
LGTM and pass test

PR #7260, opened 2020-12-23
[LINST] modify LINST class to only strip '_date' from standard dates

ridz1208:
## Brief summary of changes
These changes were made to reflect the https://github.com/aces/Loris/pull/6923 which was issued to the `main` branch

Additional changes not in #6923:
 - modified the "stripping" function to only remove `_date` if it is positioned at the end of the field name. For example if the user names their date field `my_date_of_release_date`, previously the function would strip the field name to `my`; now it would only strip the last `_date` to become `my_date_of_release`
- removed the check for `isset($dateformat)` since the code is already converting a null $dateformat to an empty string 2 lines above !
- reorganized the if statements to reduce over all complexity
- added an exception if the date format does not match any known formats

### DIFF
--- a/php/libraries/NDB_BVL_Instrument_LINST.class.inc
+++ b/php/libraries/NDB_BVL_Instrument_LINST.class.inc
@@ -604,7 +604,12 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
                                     0,
                                     -5
                                 );
-                            }
+                            } else {
+                                throw new \LorisException(
+                                    "Standard Date format field $pieces[1] must end"
+                                        . " with '_date'."
+                                );
+                            }                            
                             // Shows standard date
                             $this->addDateElement(
                                 $pieces[1],
@@ -954,5 +959,4 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
 
 
 }
-
 
--- a/php/libraries/NDB_BVL_Instrument_LINST.class.inc
+++ b/php/libraries/NDB_BVL_Instrument_LINST.class.inc
@@ -609,7 +609,7 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
                                     "Standard Date format field $pieces[1] must end"
                                         . " with '_date'."
                                 );
-                            }                            
+                            }
                             // Shows standard date
                             $this->addDateElement(
                                 $pieces[1],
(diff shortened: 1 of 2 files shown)


ridz1208:
@zaliqarosli could you review

php/libraries/NDB_BVL_Instrument_LINST.class.inc
```diff
@@ -582,42 +575,47 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
...
+                            // it is not duplicated by the addDateElement function
+                            if (substr($pieces[1], -5) === "_date") {
+                                $pieces[1] = substr(
                                     $pieces[1],
-                                    $pieces[2],
-                                    $dateOptions
+                                    0,
+                                    -5
                                 );
                             }
```
zaliqarosli, line 607:
```suggestion
                            } else {
                                throw new \LorisException(
                                    "Standard Date format field $pieces[1] must end"
                                        . " with '_date'."
                                );
                            }                            
```

zaliqarosli (changes requested):

php/libraries/NDB_BVL_Instrument_LINST.class.inc
ridz1208: Nice

zaliqarosli (approved):
👍

ridz1208:
@zaliqarosli do you have time to test it quick quick ?

zaliqarosli:
@ridz1208 ya on it!

php/libraries/NDB_BVL_Instrument_LINST.class.inc
```diff
@@ -582,42 +575,52 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
...
-                            if ($dateFormat === 'Date') {
-                                // Shows standard date
-                                $this->addDateElement(
-                                    $pieces[1],
-                                    $pieces[2],
-                                    $dateOptions
+                            } else {
+                                throw new \LorisException(
+                                    "Standard Date format field $pieces[1] must end"
+                                        . " with '_date'."
```
zaliqarosli, line 610:
```suggestion
                                    "Standard Date format field `$pieces[1]` in LINST file `$filename` must end"
                                        . " with '_date'."
```

zaliqarosli (changes requested):

php/libraries/NDB_BVL_Instrument_LINST.class.inc
ridz1208: its gonna make it fail phpcs again lol

zaliqarosli:
ill re-approve after you fix phpcs!

zaliqarosli (approved):

PR #7261, opened 2020-12-23
[create_timepoint] Use the project for the project, not the site

driusan:
Timepoint creation was incorrectly using the site dropdown as the
projectID instead of the project dropdown when the user had more
than 1 project. This corrects it to use the appropriate dropdown
for creation.


driusan:
This was introduced while Reactifying the module. It's only on the main branch right now and shouldn't affect real projects data..

ridz1208 (approved):

PR #7264, opened 2020-12-23
[survey_accouts] just phpcs

ridz1208:


driusan (approved):
Just waiting for the tests to finish before merging..

driusan:
It looks like the integration tests for survey accounts also need to be updated, but weren't run on 21 or 22..

PR #7265, opened 2020-12-23
This fixes the undeclared properties identifier by phan in one file

driusan:
There are many more that need to be fixed before the phan rule can
be enforced, but you eat an elephant one bite at a time.

1. Set `"allow_missing_properties" => false,` (is currently true) in .phan/config.php
2. Run phan and check for errors in the modified file. (There are still many from other files, but there shouldn't be any in this one.)


driusan:
@xlecours can you have a quick look? I'm just starting from the top of the errors and sending on PR per module, I think this one is yours and should be pretty straight forward..

xlecours (approved):

PR #7266, opened 2020-12-24
[Core] Date field js validation

laemtl:
For one of my projects, I need to validate a date field in react and display front-end error message. Select fields have an errorMessage and hasError props for such scenario, but those are not implemented on DateElement. This PR implement the same logic on DateElement.

### DIFF
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -1042,10 +1042,18 @@ class DateElement extends Component {
     let disabled = this.props.disabled ? 'disabled' : null;
     let required = this.props.required ? 'required' : null;
     let requiredHTML = null;
+    let errorMessage = null;
+    let elementClass = 'row form-group';
 
     // Add required asterix
     if (required) {
       requiredHTML = <span className="text-danger">*</span>;
+    }
+
+    // Add error message
+    if (this.props.hasError || (this.props.required && this.props.value === '')) {
+      errorMessage = <span>{this.props.errorMessage}</span>;
+      elementClass = elementClass + ' has-error';
     }
 
     // Check if props minYear and maxYear are valid values if supplied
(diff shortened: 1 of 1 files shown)


driusan:
Why is this going to 23?

laemtl:
I'm using v23 in the project I implemented this for.

jsx/Form.js
```diff
@@ -1104,19 +1113,23 @@ DateElement.propTypes = {
...
+  hasError: PropTypes.bool,
+  errorMessage: PropTypes.string,
   onUserInput: PropTypes.func,
 };
 
 DateElement.defaultProps = {
   name: '',
   label: '',
-  value: '',
+  value: undefined,
```
maltheism, line 1124: @laemtl should value be undefined? Line 1054 still checks `this.props.value === ''` if empty string. I'm wondering if null is more appropriate instead of undefined if empty string isn't preferred for some reason?

laemtl: If value is set to '', an error is detected on page load (initial value enter the error test case). value can be null instead, but other element (select for example) use undefined as their initial value. Any reason why undefined should be avoided?

maltheism: I've not witnessed undefined from an input tag before and so I was just curious. I get an empty string when the variable has not been set for <input> if I make one and try to console.log the value. I'm guessing date works differently.

maltheism (approved):

ridz1208:
@laemtl I have changed the base on github from 23 to main so this PR stops showing up when we query the 23 branch. The PR now shows a bunch of files changed/commits/conflicts. dont worry about these for now, as soon as we puch 23 into main (after the release) they should auto resolve and if not a minor rebase should do the trick.

driusan:
@laemtl can you rebase this? It's both showing the wrong changes since it was originally sent to a different branch and needs to have the tests run on php8.

cmadjar:
@laemtl could you rebase the PR please? Thank you!

laemtl:
@cmadjar Rebased.

driusan:
@laemtl the diff looks better now but it looks like the tests are failing since rebasing

cmadjar:
@driusan ready for you :)

PR #7267, opened 2021-01-04
[battery_manager] Fix phan undeclared property errors

driusan:
This declares the properties that are undeclared in the battery_manager module.

1. Set `"allow_missing_properties" => false,` (is currently true) in .phan/config.php
2. Run phan and check for errors in the modified file. (There are still many from other files, but there shouldn't be any in this one.)


driusan:
@HenriRabalais can you give this a quick review?

HenriRabalais (approved):
LG

PR #7270, opened 2021-01-04
Fix phan mismatched property errors

driusan:
This fixes all the errors identified by phan where a property
was being assigned to an incompatible type. Most of these were
caused by assigning null to a non-nullable property, which was
fixed by declaring the property nullable.

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -39,7 +39,6 @@ return [
         "PhanTypePossiblyInvalidDimOffset",
         "PhanUndeclaredMethod",
         "PhanTypeMismatchArgument",
-        "PhanTypeMismatchProperty",
         "PhanTypeArraySuspiciousNullable",
     ],
     "analyzed_file_extensions" => ["php", "inc"],
--- a/modules/api/php/endpoints/candidate/visit/visit.class.inc
+++ b/modules/api/php/endpoints/candidate/visit/visit.class.inc
@@ -38,7 +38,7 @@ class Visit extends Endpoint implements \LORIS\Middleware\ETagCalculator
     /**
      * The requested Visit
      *
-     * @var \Timepoint
+     * @var ?\Timepoint
      */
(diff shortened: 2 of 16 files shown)


driusan:
@nicolasbrossard Do you think you'd be able to look at this? The server_process_manager had the most significant changes and are where I'm most likely to have made a mistake..

driusan:
@laemtl since you were highlighting some type errors in another PR and @nicolasbrossard hasn't responded maybe you can look at this?

(Note that this only catches type errors for properties, function arguments is a different (much bigger..) phan error type..)

laemtl:
No pb!

nicolasbrossard:
@laemtl @driusan Ooopss...Sorry I didn't respond...I can take care of this since I wrote the SPM module.

nicolasbrossard (approved):
Build is successful and code looks good. Approving.

driusan:
Thanks @nicolasbrossard .. had to rebase because of a conflict, I'll merge after it passes again..

nicolasbrossard:
@driusan Will re-review once the build is finished.

driusan:
Looks like rebasing found new things.. I'll tag you when they're fixed.

driusan:
@nicolasbrossard it's passing now after the rebase.. the changes are basically 57e100b7c638c17f497d94888fdb35ad1a0c2e46 plus phpcs and a couple typos.

Phan was getting confused by some tests because it didn't realize that mocks should be considered *both* a mock type and the type of object being mocked. I had to explicitly tell it to treat the objects as the type being mocked instead of as a MockObject type after calling the mock related methods. In a couple cases it was easier to just not assign to a property and leave them as variables because the tests weren't doing anything that depended on it being a class property..

nicolasbrossard (approved):
Code still looks good after rebase and build successful. Approving.

PR #7271, opened 2021-01-04
Remove mandatory underscore from private properties

driusan:
This removes the rule from PHPCS enforcing a mandatory `_` on private class properties in the php/libraries directory. Having this rule enforced prevents us from converting some of our code from PHP4 phpdoc style `@access` comments to proper php private/protected/public keywords (without updating all the code referencing the variables, which is risky since `allow_missing_properties` is true in our phan config.) It's also one of the incompatibilities between our old PEAR based php/libraries and PSR2 based src coding standards which prevents us from using the same standard throughout LORIS.

### DIFF
--- a/php/libraries/UserPermissions.class.inc
+++ b/php/libraries/UserPermissions.class.inc
@@ -25,18 +25,16 @@ class UserPermissions
     /**
      * User's ID
      *
-     * @var    string
-     * @access private
+     * @var ?string
      */
-    var $userID;
+    private $userID;
 
     /**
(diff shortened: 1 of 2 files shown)


driusan:
@ridz1208 Maybe you can review? I didn't update most of the `@access` comments, just one to ensure it was working, but this lets someone update them when they touch a portion of the code that has them without introducing a bunch of other errors..

ridz1208 (approved):

ridz1208:
approve approve !

PR #7278, opened 2021-01-08
[Publication] Permission duplicate fix

### LINKED ISSUE #7006: [publication] 'LORIS Users with Edit Permission' in 'View Project' mode renders user twice
**Describe the bug**
Screen Shot 2020-09-07 at 5 23 13 PM

**To Reproduce**
1. Propose a project and add your own user to 'LORIS Users with Edit Permission' field. Submit the form.
2. (Refresh and) Click on 'Browse' tab. Click on the project you just proposed. You should be able to view the project and edit it. Add another user to the 'LORIS Users with Edit Permission' field. Click submit.
3. Refresh the page. And you should see your user name twice. Delete one of the chips for that user. Click submit.
4. Refresh the page again. You'll see that the chip you just deleted is still there.

User chip should only appear once

laemtl:
The publication_users_edit_perm_rel table does not have any PK. As a result, it allows duplicates and produces side effects as described in #7006.

### To apply the patch
Remove the duplicates
`php tools/single_use/remove_publication_users_edit_perm_rel_duplicates.php `

Update the DB schema
`cat SQL/New_patches/2020-01-07-publication_users_edit_perm_rel_pk.sql | mysql`

* Resolves #7006


kongtiaowang (dismissed):
LGTM

SQL/New_patches/2020-01-07-publication_users_edit_perm_rel_pk.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE temp SELECT DISTINCT * FROM publication_users_edit_perm_rel;
+DROP TABLE publication_users_edit_perm_rel;
+ALTER TABLE temp RENAME publication_users_edit_perm_rel;
```
driusan, line 3: I don't understand why you're going through this temp table/drop table/alter table dance. Why can't you just alter the table to add the primary key? It seems like that would be much safer.

driusan: in most upgrades we fix that type of problem with a one time use PHP script to cleanup/run at upgrade time before sourcing the patch. Doing it this way seems dangerous because the script isn't atomic if something goes wrong.

@ridz1208 what do you think?

laemtl: Good point, atomicity is an issue. Another alternative is to surround the queries with a transaction.

driusan: A transaction wouldn't help. All the statements in this patch are DDL and cause an implicit commit in MySQL.

ridz1208: @laemtl I agree with @driusan a PHP script in the single use directory is usually the solution we use.

In this case I would make a php script that would need to be run BEFORE the SQL patch (make sure to specify this in the changelog in the notes for developers and release notes and in the script itself and on this PR... everywhere) which basically removes duplications. should be pretty easy in PHP

ridz1208: @laemtl not to say that your solution wouldnt work though !!  this is just... less risky and the PHP script's changes would be tracked in the history table if the script is done properly

ridz1208 (dismissed):
see comments on patch

tools/single_use/remove_publication_users_edit_perm_rel_duplicates.php
```diff
@@ -0,0 +1,45 @@
...
+foreach ($result as $row) {
+    foreach ($row as $key => $value) {
+        echo $key . ':' . $value . '    ';
+    }
+
+    $query = "DELETE
+        FROM publication_users_edit_perm_rel
+        WHERE PublicationID = $row[PublicationID]
+        AND UserID = $row[UserID]
+        LIMIT 1";
```
jesscall, line 38: The issue with this `LIMIT 1` clause is that if there are multiple duplicates, the script does not delete them!

For example:
Before running script –
Screen Shot 2021-03-18 at 1 05 42 PM

After script was run –
Screen Shot 2021-03-18 at 1 07 23 PM

SQL Table (after script):
Screen Shot 2021-03-18 at 1 08 15 PM

jesscall (changes requested):

tools/single_use/remove_publication_users_edit_perm_rel_duplicates.php
ridz1208: good point

ridz1208:
@driusan any solution ideas for this one ?. It's starting to feel to me like a second/temporary table might be the easiest option

laemtl:
@jesscall, @driusan, @ridz1208 One solution might be to detect the number of duplicates with a count query and delete in a loop of COUNT -1 iteration.

laemtl:
@jesscall Your issue has been fixed.

ridz1208 (changes requested):
After a discussion with @driusan we have reached the following decisions:

1. For the PHP tool script, there are several ways of removing the errors, we think its best to identify duplicates, delete all duplicated entries and then re-add them uniquely in the same script. this will avoid complex/error prone queries and solve the traceability issue by logging everything automatically in the history table. (extra points for creating a log of the changes and print it out to the user)
2. this PR will be rebased onto main due to the number of changes
3. ~This PR does not appear to fix the code in order to prevent a user from entering duplicated entries in the first place, that part should probably be added as well.~ (INSERT IGNORE is used)

jesscall:
@ridz1208 
I think the fix is necessary for 23.0 release. Especially considering CCNA is starting to use the module more frequently.
What is the reasoning for rebasing on main?

ridz1208:
@jesscall 
- its non critical (duplicate users should not break anything)
- it contains an SQL patch (which is generally only on bugfix in case its critical)
- it has a PHP script adding to the complexity
- The issue itself in the code might or might not be solved by this PR (which could mean that it could do more harm than good)

I dont really see why CCNA wouldnt be able to use the module now, people should anyways not be added in duplication

jesscall:
@ridz1208 
I agree the issue itself should be fixed. 

> [...] people should anyways not be added in duplication

I just want to highlight where the issue is arising because your comment is a little ambiguous and I'm not sure it's very clearly stated in #7006 

Duplication is occurring on every submit where new users are given permission. If you print the POST request, it's submitting an array containing both the new userIDs AND the old ones previously submitted.

laemtl:
@jesscall As I mentioned to Rida, preventing insertion of duplicates in the DB with appropriate PKs is enough to fix the pb, as INSERT IGNORE is used.

driusan:
@jesscall the main problem with it going here is that we've never had an SQL patch required for a bug fix release. We concatenate the SQL patches for every minor release, but the workflow for bug fix releases tends to be just run `git pull` or download/overwrite the code. An SQL patch going to a bug fix release is very likely to be missed by projects that are doing a major upgrade.

Can you just run the SQL portion of the patch manually on your project? (It means that statement in the release upgrade will fail when the patches are concatenated, but shouldn't otherwise break anything as far as I understand)

jesscall:
@driusan yes that can be done

driusan:
Is there anything else in the PR? It's hard to tell in the diff now because the branch is updated in github but the branch hasn't been rebased.

ridz1208:
@driusan theres the script that removes the duplications

driusan:
Okay, but that can just be copied for anyone that needs to backport the fix to 23.

I still think we should proceed with putting it in `main` to avoid the risk of upgrade patches being missed.

laemtl:
@cmadjar @ridz1208 Rebased.

ridz1208 (approved):

ridz1208:
@jesscall have you ran this on the CCNA database? if so can you add the passed manual test flag please ?

driusan:
@jesscall when you do can you tag me so I remember to look at it

jesscall (approved):

jesscall:
Looks good!

@driusan you can review now

CHANGELOG.md
```diff
@@ -111,6 +111,7 @@ exception). It is recommended to run this tool for existing projects (PR #5270)
 
 ### Notes For Existing Projects
 - PHP should be upgraded to 7.3 to before upgrading LORIS.
+- The script tools/single_use/remove_publication_users_edit_perm_rel_duplicates.php should be run before upgrading LORIS.
```
driusan, line 114: This looks like it's in the wrong place. Shouldn't it be under 24 and not 23?

tools/single_use/remove_publication_users_edit_perm_rel_duplicates.php
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env php
+<?php declare(strict_types=1);
+error_reporting(E_ALL);
```
driusan, line 3: I don't see why this is here in this script. I don't think we override the error reporting level in any other script?

driusan (changes requested):

ridz1208:
@laemtl please fix

cmadjar:
@laemtl it looks like this is in your court. Thanks!

ridz1208:
@driusan all fixed up, running tests

tools/single_use/remove_publication_users_edit_perm_rel_duplicates.php
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env php
```
ridz1208, line 1: before SQL

Issue #4213, opened 2018-12-20
Replace jQuery $.ajax with the javascript fetch API


driusan:
The $.ajax (and related) functions are all that's forcing us to keep our dependency on jQuery. We should replace them with the more standard javascript [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) so that we can remove the unnecessary dependency.

driusan:
There are still instances that can be found with `git grep '\$\.ajax'` (it's probably better to just include the command than list them in case it gets outdated.)

driusan:
Adding priority high because we really shouldn't be including jQuery as a dependency anymore and this is probably the main blocker.

PR #7283, opened 2021-01-08
[Publication] jQuery usage cleaning

laemtl:
First attempt to clean progressively jQuery from the modules.
Replace $.ajax by fetch and $.ready by document.addEventListener('DOMContentLoaded')

Makes also use of arrow functions to avoid the use of bind(this)

* Will make a step toward #4213 resolution
* Testing requires the changes made in #7284

### DIFF
--- a/modules/publication/jsx/viewProject.js
+++ b/modules/publication/jsx/viewProject.js
@@ -78,7 +78,6 @@ class ViewProject extends React.Component {
    * Fetch data
    */
   fetchData() {
-    let self = this;
     fetch(this.props.DataURL, {
       method: 'GET',
     }).then((response) => {
@@ -121,7 +120,7 @@ class ViewProject extends React.Component {
             });
           }
 
-          self.setState({
+          this.setState({
             formData: formData,
             users: data.users,
             statusOpts: data.statusOpts,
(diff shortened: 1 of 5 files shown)


modules/publication/jsx/viewProject.js
```diff
@@ -87,60 +79,66 @@ class ViewProject extends React.Component {
    */
   fetchData() {
     let self = this;
```
maltheism, line 81: Hi @laemtl, I believe `let self = this` isn't necessary with your changes.

maltheism (changes requested):

modules/publication/jsx/viewProject.js
laemtl: Your totally right, thanks for catching this!

laemtl: @maltheism Fixed.

maltheism (approved):

driusan:
Seems to have conflicts..

driusan:
Skimmed it and it looks good (if long overdue!) but it has conflicts and can't be merged

laemtl:
@driusan Rebased.

PR #7091, opened 2020-10-09
[Publication] Fix Issues Reported

sruthymathew123:
## Brief summary of changes
Fixes Issues Reported in Publication Module.
I think these bugs are introduced after merging of this PR #6195 (Some of the code in publication module needs to be updated as well with the sweetalert 2 update)

* Resolves 
#7005 (File Deletion not working)
#7007 (After successfully proposing project, project doesn't appear in browse list)
#7006

PR #7284, opened 2021-01-09
[Publication] File delete not working (#6195 regression)

laemtl:
To complement #7091, an additional error introduced by #6195 is fixed by this PR, when a file previously uploaded is deleted.

## To test
- Create a publication with a file and the necessary permission to edit (LORIS Users with Edit Permission).
- Edit and delete the file.
- The file should be deleted without error


kongtiaowang (approved):
LGTM

PR #7063, opened 2020-09-24
[cand. profile/instr. list] Send to DCC/BVL Feedback sessionID fixes

laemtl:
This PR completes the work done on #7036 to solve some errors introduced by #5352 (candidate profile/instrument list)

error 500 if $_POST['sessionID'] is an empty string (react_get_bvl_threads.php)
console warning: sessionID is of type object (BVL_Feedback_Panel.class.inc)
one omitted conversion of $_POST['sessionID'] to new \SessionID($_POST['sessionID']) (bvl_panel_ajax.php)

Resolves #7139
Resolves #7017 
Resolves #6987

PR #7289, opened 2021-01-18
[libraries/BVL_Feedback_Panel] $sessionID must be type SessionID

maltheism:
## Brief summary of changes

A string was given and this change resolves it.

### DIFF
--- a/modules/instrument_list/php/instrument_list.class.inc
+++ b/modules/instrument_list/php/instrument_list.class.inc
@@ -288,7 +288,10 @@ class Instrument_List extends \NDB_Menu_Filter
      */
     function getFeedbackPanel($candID, $sessionID): string
     {
-        $feedbackPanel = new \BVL_Feedback_Panel($candID, $sessionID);
+        $feedbackPanel = new \BVL_Feedback_Panel(
+            $candID,
+            new \SessionID($sessionID)
+        );
         $html          =  $feedbackPanel->display();
         return $html;
     }
--- a/modules/instrument_list/php/instrument_list.class.inc
+++ b/modules/instrument_list/php/instrument_list.class.inc
@@ -1,4 +1,6 @@
 <?php
+declare(strict_types=1);
+
 /**
  * Instrument_List
  *
@@ -21,6 +23,7 @@ namespace LORIS\instrument_list;
 use \Psr\Http\Message\ServerRequestInterface;
 use \Psr\Http\Server\RequestHandlerInterface;
 use \Psr\Http\Message\ResponseInterface;
+use \Loris\StudyEntities\Candidate\CandID;
 
 /**
  * Instrument_List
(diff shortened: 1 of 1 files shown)


laemtl:
There is another [getFeedbackPanel here](https://github.com/aces/Loris/blob/5166ada28fbb01ac020f03c6eb38766f2ff7e32a/php/libraries/NDB_BVL_Instrument.class.inc#L2415). I wonder if the 2 signature should agree, it will be confusing otherwise.

If so, the sessionID type change should be done before calling getFeedbackPanel.

laemtl:
Related to #7063

laemtl:
Just noticed the [signature](https://github.com/aces/Loris/blob/4af1256b6072ec79e08870dc1f44e851b343600d/modules/instrument_list/php/instrument_list.class.inc#L293), sessionID should be of type SessionID, so this function is probably called somewhere with the wrong argument. Maybe also specifying the types in the function definition will help to clarify the confusion.

driusan:
@maltheism Can you rebase so that this runs with php8 and to ensure all the latest phan fixes still pass?

kongtiaowang (approved):
LGTM

modules/instrument_list/php/instrument_list.class.inc
```diff
@@ -296,7 +296,10 @@ class Instrument_List extends \NDB_Menu_Filter
      */
     function getFeedbackPanel($candID, $sessionID): string
```
laemtl, line 297: The signature (sessionID) disagrees with the current change.
This function is called somewhere with the wrong argument. Specifying the types in the function definition will help to clarify the confusion.

laemtl (changes requested):

laemtl:
@maltheism I submitted a commit with suggested changes. Let me know what you think.
Also, I investigated I couldn't find an call of getFeedbackPanel with $sessionID of type string.

driusan:
Our tests aren't amazing at finding mismatches because `PhanTypeMismatchArgument` is still ignored. After #7322 I have another PR in the pipeline that fixes the remaining `PhanTypeMismatchArgument` errors and unignores it.

In the meantime, maybe adding `declare(strict_types=1);` to the file with the signature change would help the tests find any culprits calling it with a string?

laemtl:
@driusan I added declare(strict_types=1) and it confirms my suspicion: The code breaks when we call 
`new \SessionID($sessionID)` with ` PHP Fatal error:  Uncaught TypeError: Argument 1 passed to ValidatableIdentifier::__construct() must be of the type string, object given`.

laemtl:
Since the strict _types flag, Phan is complaining with 
` modules/instrument_list/php/instrument_list.class.inc:255 PhanTypeMismatchArgumentProbablyReal Argument 1 ($select_inactive_threads) is $feedback_select_inactive of type ?'Y' but \NDB_BVL_Feedback::getMaxThreadStatus() takes bool (no real type) defined at php/libraries/NDB_BVL_Feedback.class.inc:380 (the inferred real argument type has nothing in common with the parameter's phpdoc type)`

driusan:
That sounds a lot like one of the errors I had to fix while fixing the phan errors.

laemtl:
@driusan I tried something and it fixes phan.

driusan:
The 'Y' => true thing looks familiar. I think that's what I had to do to fix it too..

laemtl (approved):

PR #7291, opened 2021-01-20
[Eslint] Warning/Error on jQuery usage

laemtl:
Add new rules to eslint to flag all the jQuery usage in the modules.
Severity is error for the cleaned modules, warn for the others.


maltheism (approved):
Hi @laemtl, it looks good! I assume we eventually remove the eslint-plugin-no-jquery dependency and all no-jquery rules after all jquery code including the library itself is removed from Loris.

PR #7294, opened 2021-01-21
[Core] Datastream iterator fix

laemtl:
Fix problems found on the datastream iterator
* the first element of the stream is always skipped
* an empty element is added at the end

To see this issue, go to /datadict/
* on the last page, notice the empty element
* compare the number of results with https://localhost/datadict/?format=json

### Origins of the problem
* The iterator call next before current, causing the first element to be skipped
* the row callback is called when the last line is detected, causing the last empty line issue.

### DIFF
--- a/src/Http/DataIteratorBinaryStream.php
+++ b/src/Http/DataIteratorBinaryStream.php
@@ -183,15 +183,17 @@ class DataIteratorBinaryStream implements StreamInterface
         if ($this->eof) {
             return "";
         }
-        $this->rowgen->next();
-        $row = $this->rowgen->current();
         if (!$this->rowgen->valid()) {
             $this->eof = true;
             return chr(0x04);
         }
+        $row = $this->rowgen->current();
+        $this->rowgen->next();
+
         $rowArray        = array_values(json_decode(json_encode($row), true));
         $rowVal          = join(chr(0x1e), $rowArray) . chr(0x1f);
         $this->position += strlen($rowVal);
+
         return $rowVal;
     }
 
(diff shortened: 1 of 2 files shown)


src/Http/DataIteratorBinaryStream.php
```diff
@@ -181,15 +181,17 @@ public function read($length)
         if ($this->eof) {
             return "";
         }
-        $this->rowgen->next();
-        $row = $this->rowgen->current();
         if (!$this->rowgen->valid()) {
             $this->eof = true;
             return chr(0x04);
         }
+        $row = $this->rowgen->current();
```
driusan, line 188: I'm not sure this is right according to the documentation. https://www.php.net/manual/en/iterator.valid.php

```
This method is called after Iterator::rewind() and Iterator::next() to check if the current position is valid. 
```

You're changing it to call valid before calling next.

driusan: I think you just need to explicitly call rewind in the constructor to fix this problem.

laemtl: @driusan I unfortunately does not seem to work. The [generator documentation](https://www.php.net/manual/en/language.generators.comparison.php) also mentions that _generators are forward-only, meaning that it should be compared to an iterator with a dummy rewind function defined_.

laemtl: Also, the documentation above is for Iterator. The [valid() method for Generators](https://www.php.net/manual/en/generator.valid.php) is used to detect if the end of the loop has been reached.

kongtiaowang (approved):
LGTM

driusan:
@laemtl can you rebase so that it runs with php8 before I merge?

laemtl:
@driusan Done

PR #7295, opened 2021-01-25
Fix all references to undeclared properties identified by Phan

driusan:
This toggles sets the `allow_undeclared_properties` flag to false and fixes all references to undeclared properties identified by Phan.

### DIFF
--- a/.github/workflows/loristest.yml
+++ b/.github/workflows/loristest.yml
@@ -14,7 +14,6 @@ jobs:
             - unit
             - integration
             php:
-            - 7.3
             - 7.4
 
     steps:
--- a/modules/genomic_browser/php/models/cnvdto.class.inc
+++ b/modules/genomic_browser/php/models/cnvdto.class.inc
@@ -239,14 +239,4 @@ class CnvDTO implements \LORIS\Data\DataInstance
             'Platform'          => $this->_Platform,
         ];
     }
-    /**
-     * Returns the CenterID for this row, for filters such as
-     * \LORIS\Data\Filters\UserSiteMatch to match again.
-     *
-     * @return string
-     */
-    public function getPSC(): string
-    {
-        return (string) $this->psc;
-    }
 }
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1105,22 +1105,6 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     abstract function getSubtestList(): array;
 
     /**
-     * Marks an element in the lorisform object as being required (for
-     * use with the requiredIf rule)
-     *
-     * @param string $elementName the element to add to the required list
-     *
-     * @return void
-     */
-    function setRequired(string $elementName): void
-    {
-        if (!in_array($elementName, $this->form->_required)) {
-            $this->form->_required[] = $elementName;
-        }
-    }
-
-
-    /**
      * Convert a lorisform date or timestamp into a database acceptable
      * (and storable) date or time
      *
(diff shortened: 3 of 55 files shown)


modules/genomic_browser/php/models/cnvdto.class.inc
```diff
@@ -239,14 +239,4 @@ class CnvDTO implements \LORIS\Data\DataInstance
             'Platform'          => $this->_Platform,
         ];
     }
-    /**
-     * Returns the CenterID for this row, for filters such as
-     * \LORIS\Data\Filters\UserSiteMatch to match again.
```
driusan, line 244: UserSiteMatch doesn't do anything with a function named `getPSC` and if it did, this is referencing a property that doesn't exist..

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1104,22 +1104,6 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
-     * Marks an element in the lorisform object as being required (for
-     * use with the requiredIf rule)
-     *
-     * @param string $elementName the element to add to the required list
-     *
-     * @return void
-     */
-    function setRequired(string $elementName): void
-    {
-        if (!in_array($elementName, $this->form->_required)) {
```
driusan, line 1117: LorisForm doesn't declare a _required, and makes no references to one. If this function is ever called by anyone it wouldn't do anything.

php/libraries/NDB_Menu_Filter.class.inc
```diff
@@ -544,23 +553,6 @@ class NDB_Menu_Filter extends NDB_Menu
             }
         }
 
-        if (isset($this->searchKeyword)
```
driusan, line 547: I did a `git grep searchKeyword` and no modules in LORIS are making use of this.

php/libraries/NDB_Menu_Filter.class.inc
```diff
@@ -855,10 +847,6 @@ class NDB_Menu_Filter extends NDB_Menu
         }
         $query .= $this->order_by;
 
-        if (!empty($this->limit)) {
```
driusan, line 858: I couldn't find any modules making use of this either.

.github/workflows/loristest.yml
```diff
@@ -14,7 +14,6 @@ jobs:
             - unit
             - integration
             php:
-            - 7.3
```
driusan, line 17: PHP 7.3 doesn't support typed properties.

driusan:
@kongtiaowang could you review this? I think the most significant number of changes are in the tests.. I'd rather this doesn't go stale because this setting being disabled makes it a lot more dangerous to do refactorings or rename things in classes..

php/libraries/NDB_Menu_Filter.class.inc
driusan: (if there were, it should cause phan to fail since this PR is enabling phan detecting undefined property accesses..)

src/Router/ModuleFileRouter.php
```diff
@@ -45,6 +45,8 @@ class ModuleFileRouter implements RequestHandlerInterface
     protected $subdir;
 
     protected $contenttype;
+
+    protected string $moduledir;
```
kongtiaowang, line 49: PHP Parse error:  syntax error, unexpected 'string' (T_STRING), expecting function (T_FUNCTION) or const (T_CONST) in /var/www/loris/src/Router/ModuleFileRouter.php on line 49

kongtiaowang (approved):
LGTM

PR #7137, opened 2020-11-23
[webpack] eslint-loader deprecated and replaced by eslint-webpack-plugin

maltheism:
## Brief summary of changes

The eslint-loader dependency has been deprecated and it's recommended to be replaced with eslint-webpack-plugin. I've added .eslintcache to the .gitignore rules.

The eslint-webpack-plugin package requires a newer version of eslint. The newer version of eslint has the .eslintrc.json use the extends array outside of rules. I'm unsure if the extends rules were ever being used before. Since, these errors are now being displayed:
```

 ERROR in 
    /Users/alizee/Development/GitHub/McGill/Loris/jsx/CSSGrid.js
      63:31  error  'Cards' is missing in props validation          react/prop-types
      63:37  error  'Cards.length' is missing in props validation   react/prop-types
      64:32  error  'Cards' is missing in props validation          react/prop-types
      65:20  error  'Cards' is missing in props validation          react/prop-types
      65:29  error  'Cards[].Order' is missing in props validation  react/prop-types
    
    /Users/alizee/Development/GitHub/McGill/Loris/jsx/Card.js
      52:20  error  'style' is missing in props validation     react/prop-types
      53:52  error  'style' is missing in props validation     react/prop-types
      55:20  error  'cardSize' is missing in props validation  react/prop-types
      56:40  error  'cardSize' is missing in props validation  react/prop-types
      65:34  error  'cardSize' is missing in props validation  react/prop-types
      71:25  error  'children' is missing in props validation  react/prop-types
    
    /Users/alizee/Development/GitHub/McGill/Loris/jsx/DynamicDataTable.js
       31:4   error  Unnecessary semicolon                              no-extra-semi
      141:28  error  'Filter' is missing in props validation            react/prop-types
      142:38  error  'getFormattedCell' is missing in props validation  react/prop-types
      143:34  error  'freezeColumn' is missing in props validation      react/prop-types
      144:28  error  'onSort' is missing in props validation            react/prop-types
    
    /Users/alizee/Development/GitHub/McGill/Loris/jsx/Form.js
        70:39  error  'children' is missing in props validation                                  react/prop-types
       106:30  error  'fileUpload' is missing in props validation                                react/prop-types
       197:39  error  'children' is missing in props validation                                  react/prop-types
       332:34  error  'sortByValue' is missing in props validation                               react/prop-types
       373:21  error  Do not access Object.prototype method 'hasOwnProperty' from target object  no-prototype-builtins
[...]

PR #7299, opened 2021-01-25
[eslint] fix 'Modernizr' is not defined  no-undef

maltheism:
## Brief summary of changes

Solves eslint errors for no-undef and useful for removing the off found in this PR: https://github.com/aces/Loris/pull/7137

Fixes 'Modernizr' is not defined  no-undef


laemtl:
@maltheism eslint is not passing :)

laemtl (dismissed):
@maltheism eslint is not passing :)

maltheism:
Hi @laemtl see latest commit and I'm open for any suggestions.

cmadjar:
@driusan could you review this?

modules/api_docs/jsx/swagger-ui_custom.js
```diff
@@ -3,6 +3,7 @@ import 'swagger-ui-dist/swagger-ui.css';
 
 window.addEventListener('load', () => {
   const ui = new SwaggerUIBundle({
+    // eslint-disable-next-line no-undef
```
driusan, line 6: why is this here?

maltheism: Hi @driusan the variable "schema_urls" is undefined and created in php. So the eslint-disable-next-line no-def is one solution. @laemtl do you know if there's a more elegant solution for this?

driusan: I still don't understand.. how is php defining things in the javascript file? Is it a global variable? If so, can it be explicitly defined as one somewhere?

maltheism: @driusan I meant the php line `$this->tpl_data['schema_urls'] = json_encode($api_specs);`. I think the global variable solution is best as well. I added the change in latest commit.

modules/media/jsx/editForm.js
```diff
@@ -198,13 +198,6 @@ class MediaEditForm extends Component {
     e.preventDefault();
 
     let xhr = new XMLHttpRequest();
-    xhr.upload.addEventListener('progress', (evt) => {
```
maltheism, line 201: I don't believe the following removed code did anything visual for the user. I had the same experience when testing the media upload of a file.

PR #7300, opened 2021-01-25
[eslint] missing in props validation   react/prop-types

maltheism:
## Brief summary of changes

Solves eslint errors for react/prop-types and useful for removing the off found in this PR: #7137

### DIFF
--- a/jsx/DynamicDataTable.js
+++ b/jsx/DynamicDataTable.js
@@ -154,7 +154,7 @@ DynamicDataTable.propTypes = {
   Filter: PropTypes.object,
   getFormattedCell: PropTypes.func,
   freezeColumn: PropTypes.string,
-  onSort: PropTypes.bool,
+  onSort: PropTypes.func,
 };
 
 DynamicDataTable.defaultProps = {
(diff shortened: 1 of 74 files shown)


driusan:
This doesn't seem to remove the "off"?

maltheism:
Hi @driusan, I had to do a rebase and now it should be there.

laemtl (dismissed):
@maltheism eslint is not passing :)

maltheism:
Hi @laemtl & @driusan, after rebasing I'm seeing 1194 problems and this is kind of like react's version of type checking. I think this PR should be broken into multiple PRs because it will take awhile to correct all of the problems with likely requiring rebasing overtime.

maltheism:
@driusan @laemtl this PR is ready for review. Some of the prop types might be incorrect but I believe that shouldn't break anything.

driusan:
I've rebased this and it looks like it's failing from new things that have crept in since it was sent.. I think they should be pretty easy to fix so I'll try and find time to fix this PR.

jsx/DynamicDataTable.js
```diff
@@ -151,6 +151,10 @@ class DynamicDataTable extends Component {
 DynamicDataTable.propTypes = {
   DataURL: PropTypes.string.isRequired,
   hiddenHeaders: PropTypes.array,
+  Filter: PropTypes.object,
+  getFormattedCell: PropTypes.func,
+  freezeColumn: PropTypes.string,
+  onSort: PropTypes.bool,
```
laemtl, line 157: It's a PropTypes.func

driusan: good catch, fixed it.

PR #7303, opened 2021-01-25
Add ability to get database connection info from environment

driusan:
This adds the ability for LORIS to get the database connection
information from environment variables prefixed by "LORIS_DB_".

This is primarily intended to make it easier to make it easier
to write a re-useable Docker image. The config.xml is normally
set up by the install.sh script with the connection info. However,
the values need to be known at the time that install.sh is run for
this to work. In the case of a Dockerfile, that needs to be known
when the image is being built. Allowing them to be overridden at runtime
allows the same docker image to be used to run multiple servers.

Note that this is only a first step. There is still a requirement to
mount a config.xml in the usual location when running a container in
order to get visit label and identifier format settings, but this
allows a container to be run using the default values, rather than
breaking entirely, if database environment variable settings are set.

There should be further work done to move the remaining settings into the
database so that a container can just point at a database and work.

### DIFF
--- a/php/libraries/NDB_Client.class.inc
+++ b/php/libraries/NDB_Client.class.inc
@@ -53,23 +53,9 @@ class NDB_Client
 
         $config = $factory->config($configFile);
 
-        // Load the database object with settings from config
-        // so that after this we can just call Database::singleton()
-        // with no parameters
-        $DBSettings = $config->getSetting("database");
-        if (!defined("UNIT_TESTING")) {
-            $DB = Database::singleton(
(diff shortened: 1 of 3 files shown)


kongtiaowang:
@driusan This is a good idea. 
I have created a docker installation service for Loris with config.xml successfully on Mac. 
There are a lot of specific situations that need to test. I will send my PR later. 
Users could build a Loris service with an external database or docker database.

kongtiaowang (approved):
LGTM

PR #7304, opened 2021-01-25
[phan] Upgrade phan to v4.x

driusan:
Update phan to the latest version.. there were only 2 errors that had to be fixed:

```
src/Middleware/ExceptionHandlingMiddleware.php:51 PhanPossiblyUndeclaredVariable Variable $status is possibly undeclared
test/integrationtests/LorisIntegrationTest.class.inc:593 PhanPossiblyUndeclaredVariable Variable $e is possibly undeclared
```

### DIFF
--- a/src/Middleware/ExceptionHandlingMiddleware.php
+++ b/src/Middleware/ExceptionHandlingMiddleware.php
@@ -31,6 +31,7 @@ class ExceptionHandlingMiddleware implements MiddlewareInterface
         RequestHandlerInterface $handler
     ) : ResponseInterface {
         try {
+            $status = 200;
             return $handler->handle($request);
         } catch (\NotFound $e) {
             error_log($e->getMessage() . $e->getTraceAsString());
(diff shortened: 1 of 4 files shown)


src/Middleware/ExceptionHandlingMiddleware.php
```diff
@@ -31,6 +31,7 @@ public function process(
         RequestHandlerInterface $handler
     ) : ResponseInterface {
         try {
+            $status = 200;
```
maltheism, line 34: @driusan I don't think this does anything since next line returns.

maltheism (approved):

src/Middleware/ExceptionHandlingMiddleware.php
driusan: It fixes the `src/Middleware/ExceptionHandlingMiddleware.php:51 PhanPossiblyUndeclaredVariable Variable $status is possibly undeclared` error

PR #7306, opened 2021-01-25
[issue_tracker] Only show inactive user if already selected in form

zaliqarosli:
## Brief summary of changes

Currently, the issue tracker shows inactive users. You are able to create a new issue and assign it to them.

This PR filters out inactive users so that they don't show up in the list of assignees in the 'New Issue' form. In the 'Edit Issue' form, the inactive user appears in the list of assignees but the option is disabled. this is so that inactive users that are already assigned show up but cannot be reselected.

I thought of this fix super quickly, so am open to suggestions of perhaps better ways of doing this! I know that there are multiple modules across loris that show inactive users in a select element only because they may have been previously selected and the value cannot be orphaned.

1. Set one of your users as Inactive using the user accounts module.
2. On 23.0-release branch, notice that the inactive user shows up as an option in the 'Create New Issue' and 'Edit Issue' forms.
3. On this PR branch, `make dev` and check both forms again. You will see that in the 'Create New Issue' form, the inactive user is not there.
4. Select an existing issue assigned to an inactive user. In the 'Edit Issue' form, you will see that inactive assignee is selected but the option is greyed out. You can update the issue i.e. Save the form, without changing assignees with no problems. If you change the assignee, you will see that you cannot reassign the issue back to the inactive user.
5. Select an existing issue assigned to an active user. You will see the inactive user in the options for assignee but you won't be able to select it.

### DIFF
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -532,6 +532,7 @@ class SelectElement extends Component {
     let disabled = this.props.disabled ? 'disabled' : null;
     let sortByValue = this.props.sortByValue;
     let options = this.props.options;
+    let disabledOptions = this.props.disabledOptions;
     let errorMessage = null;
     let emptyOptionHTML = null;
     let requiredHTML = null;
@@ -564,16 +565,28 @@ class SelectElement extends Component {
         }
       }
       optionList = Object.keys(newOptions).sort().map(function(option) {
+        let isDisabled = (newOptions[option] in disabledOptions);
         return (
-          <option value={newOptions[option]} key={newOptions[option]}>
+          <option
+            value={newOptions[option]}
+            key={newOptions[option]}
+            disabled={isDisabled}
+          >
             {option}
           </option>
         );
       });
     } else {
       optionList = Object.keys(options).map(function(option) {
+        let isDisabled = (option in disabledOptions);
         return (
-          <option value={option} key={option}>{options[option]}</option>
+          <option
+            value={option}
+            key={option}
+            disabled={isDisabled}
+          >
+            {options[option]}
+          </option>
         );
       });
     }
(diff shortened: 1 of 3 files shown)


zaliqarosli:
@HenriRabalais @maltheism @ridz1208 i am curious what you guys think

driusan:
I didn't realize you could disable individual options in a select.. I like the idea.

zaliqarosli:
@driusan i know right! i was also surprised finding this out. im not 100% happy with how i implemented it so i am curious if henri knows how to make this more adaptable for across loris

maltheism:
I think it's a great idea too! I like your implementation as well.

HenriRabalais:
I think this is cool! I can also see uses for it in the Biobank Module where protocols have been deprecated, so they shouldn't be selected but they still need to appear for specimens that already use them.

This looks like the most straight-foward implementation to me. I can't but feel that it would be nicer to still pass a single "options" object and have "disabled" be a flag. E.g.
```
options: {
  index1: {value: value1, disabled: false},
  index2: {value: value2, disabled: true},
  index3: {value: value3, disabled: false},
  etc.
}
```
However, this would necessitate a restructuring of the way we pass options to `SelectElements` and may present its own messiness.

I also think that it would be nice to have a standardized way of querying users, or subsets of those users (such as those that are inactive) so that it doesn't have to be rewritten, but that's a whole other issue unrelated to this pr lol.

Is there anything in particular about the implementation that you weren't happy with?

driusan:
This is going to the wrong branch

ridz1208:
this is a really cool feature, but a feature nonetheless. I changed the base branch to main

kongtiaowang (approved):
LGTM

PR #7307, opened 2021-01-27
Remove NDB_BVL_Instrument_instrument_preview.class.inc

driusan:
This class was at one point in time used for by the
instrument builder to render the instrument using the
same mechanisms as normal instruments, in order to give
a preview of the instrument form being built.

There is no longer any "Preview" button in the instrument_builder
module and the class only adds maintenance overhead.


php/libraries/NDB_Caller.class.inc
```diff
@@ -92,9 +92,7 @@ class NDB_Caller
         $phpfile   = $base . "project/instruments/"
             . "NDB_BVL_Instrument_$test_name.class.inc";
         $linstfile = $base."project/instruments/$test_name.linst";
-        if ($this->existsAndRequire($phpfile)
-            || $test_name=='instrument_preview'
```
driusan, line 96: Even if the instrument builder added the button back, this is wrong.. the body of the if statement goes through the code path for survey instruments, not normal instruments.

driusan:
@zaliqarosli can you review this? You probably know the current state of the instrument builder best and would be best suited to know if I missed something..

driusan:
@zaliqarosli ping

zaliqarosli (approved):
yeah i think this is fine. i wasn't around when the instrument preview button existed so I don't have the historical knowledge regarding this. but as far as i can see, this PR doesn't affect the instrument builder in any way.

i am getting a javascript error when running `make dev` or `make` (so I am not able to test the front-end) but this exists on the main branch so I believe is unrelated to this PR

PR #7308, opened 2021-01-27
Add my_preferences module to list of tests to run

driusan:
The my_preferences module is not currently in the list of modules for which the test suite is run. The module seems to have tests, they should be run.


driusan:
@kongtiaowang the tests are failing. Could you take over this PR to fix the tests in the my_preference module? They seem to have gotten outdated with some schema changes in the time that they weren't being run.. they also call an undefined "accessUser" function.

driusan:
Looks like they're passing.. thanks @kongtiaowang!

PR #7309, opened 2021-01-27
Fix all calls to undeclared methods

driusan:
This fixes all calls to undeclared methods caught by phan and removes the PhanUndeclaredMethod from the suppressed issue types.

The fixes are broadly classified into:
1. False positives caused by phan/phan#2628
2. False positives caused by phan/phan#2129
3. Mostly harmless positives caused by chaining withStatus on a PSR interface *after* a different call which lower the specificity from ResponseInterface to MessageInterface (because of which level of inheritance the method is defined on..)
4. Real positives

The first class of problems is fixed by putting the variable into a simple variable so that `@phan-var` annotations can be used on them to assert the type. These should eventually be removed when the above issue is solved.
The second class of problems by asserting the correct type before the calls
The third class is solved by re-ordering the method chaining
The fourth class is fixed, mostly by adding method_exists or instanceof checks where required.

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -36,7 +36,6 @@ return [
         "PhanUnusedVariable",
         "PhanUnusedPublicNoOverrideMethodParameter",
         "PhanTypePossiblyInvalidDimOffset",
-        "PhanUndeclaredMethod",
         "PhanTypeMismatchArgument",
     ],
     "analyzed_file_extensions" => ["php", "inc"],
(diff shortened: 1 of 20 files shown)


driusan:
Blocked by #7307 (I didn't fix the class which should just be removed) and #7308 (it's a true positive, the tests need to be fixed..)

driusan:
@kongtiaowang can you review?

kongtiaowang (approved):
LGTM

PR #7312, opened 2021-02-01
[phan] Fix errors of type PhanTypeArraySuspiciousNullable

driusan:
This fixes all errors from phan of type PhanTypeArraySuspiciousNullable. This catches situations where a variable or property may be null in some code paths, but the array is unconditionallly indexed as an array.


driusan:
@maltheism This seems like it would fix the type of warning that bothers you, so maybe you want to review once it passes the tests?

maltheism (approved):

PR #7313, opened 2021-02-02
git actions php8

kongtiaowang:
## Brief summary of changes

1.


.github/workflows/loristest.yml
```diff
@@ -37,7 +38,7 @@ jobs:
           ${{ runner.os }}-php-
 
     - name: Change PHP Version in Dockerfile
-      run: sed -i "s/7.3/${{ matrix.php }}/g" Dockerfile.test.php7
+      run: sed -i "s/7.4/${{ matrix.php }}/g" Dockerfile.test.php7
```
driusan, line 41: Why is this change needed? If the script is doing what it should then it shouldn't matter if it's "7.3", "7.4" or "apple".. this makes me worried that it's not properly testing 8.0.

php/libraries/User.class.inc
```diff
@@ -338,7 +338,7 @@ class User extends UserPermissions implements \LORIS\StudyEntities\AccessibleRes
     {
         $site_list = $this->userInfo['examiner'];
         unset($site_list['pending']);
-        natcasesort($site_list);
+        ksort($site_list);
```
driusan, line 341: Why is this needed? It's a different sorting order..

src/Middleware/UserPageDecorationMiddleware.php
```diff
@@ -1,4 +1,16 @@
 <?php
+/**
+ * This file handles the UserPageDecorationMiddleware for LORIS
+ *
+ * PHP Version 7
+ *
+ * @category LORIS
+ * @package  Main
+ * @author   Dave MacFarlane <driusan@bic.mni.mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
```
driusan, line 10: This doesn't have anything to do with PHP8 and looks like a step backwards in terms of our phpcs standards..

driusan (changes requested):

php/libraries/User.class.inc
kongtiaowang: Site_list is a multiple dimension array. The natcasesort function only works with multiple dimension array in php7.4, not php8.  The ksort will do the same thing with natcasesort. only sorts the first level key.

.github/workflows/loristest.yml
kongtiaowang: 7.4 is the default version of the Docker PHP file.
if will be replaced by matrix.php (7.4 or 8.0) .
Screen Shot 2021-02-05 at 10 58 09 AM
Screen Shot 2021-02-05 at 10 57 44 AM

php/libraries/User.class.inc
driusan: The sorting order is different. ksort sorts by key, while natcasesort sorts more human-like way for the front end (ie. the sorting goes "1", "2", .. "10" rather than "1, "10", "2" ..). But if that's not working with php8 and that's a trade-off we need to make, I guess having PHP8 tested is more important.

PR #7315, opened 2021-02-02
[Issue Tracker] jQuery cleanup

laemtl:


driusan:
@maltheism Can you review? Looks good to me but I just skimmed it to be sure it's just doing what it says..

maltheism:
Hi @laemtl @driusan, the PR seems to be creating the issue without jquery. I don't know why but the redirect is broken after the issue is created. That seems to be on main as well. Also the error message from the server is not being shown and once again that's on main as well.

maltheism (approved):
Hi @laemtl @driusan, the PR is creating the issue without jquery successfully so I'm approving. I did notice that the redirect is broken after the issue is created. That seems to be on the main branch as well. Lastly, the error message from the server is not being shown and once again that's on main as well.

laemtl:
@maltheism Thanks for your comment. I believe the same problems were fixed for Publication, after the swal migration. I can have a look.

PR #7316, opened 2021-02-02
[Candidate list] jQuery cleanup

laemtl:
### To test
- Login with a user that does not have access_all_profile permission
- Go to /candidate_list/
- Open a candidate profile using the Open profile action.

Note: A second fix (L138) prevents the validation to be fired twice (and a fetch network error on the second run).

### DIFF
--- a/modules/candidate_list/jsx/openProfileForm.js
+++ b/modules/candidate_list/jsx/openProfileForm.js
@@ -70,12 +70,22 @@ class OpenProfileForm extends Component {
     };
     this.setState(state);
 
-    $.get(loris.BaseURL + '/candidate_list/validateIDs',
-      {
-        CandID: state.CandID,
-        PSCID: state.PSCID,
-      },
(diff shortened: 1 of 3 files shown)


modules/candidate_list/jsx/openProfileForm.js
```diff
@@ -135,7 +149,6 @@ class OpenProfileForm extends Component {
         <ButtonElement
           name='Open Profile'
           label='Open Profile'
-          onUserInput={this.validateAndSubmit}
```
laemtl, line 138: This fires a second validation process (the first is fired by the form onSubmit event).
 A fetch network error occurs on the second run.

laemtl:
The failing integration test needs changes in #7319 to succeed.

driusan:
@maltheism since this seems to be passing if you have a few minutes can you have a look at it?

maltheism (approved):

PR #7322, opened 2021-02-04
PhanTypeMismatchArgument tests

driusan:
This fixes all the PhanTypeMismatchArguments in the tests directory. These are primarily caused by either phan not understanding that PHPUnit mocks should be compatible with the underlying type or incorrect phpdoc comment.

Mismatches outside of the test directory are not yet fixed.

1. Unignore PhanTypeMismatchArgument in `.phan/config.php` and ensure that there are no errors in the tests directory.

### DIFF
--- a/php/libraries/Visit.class.inc
+++ b/php/libraries/Visit.class.inc
@@ -25,13 +25,6 @@ namespace LORIS;
 class Visit
 {
     /**
-     * The visit ID
-     *
-     * @var ?string
-     */
-    protected $id;
-
-    /**
      * The visit name (formely known as visit_label)
      *
      * @var ?string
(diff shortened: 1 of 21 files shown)


driusan:
Blocked by #7309, which it's based on..

php/libraries/Visit.class.inc
```diff
@@ -24,13 +24,6 @@ namespace LORIS;
  */
 class Visit
 {
-    /**
-     * The visit ID
-     *
-     * @var ?string
-     */
-    protected $id;
```
driusan, line 32: This is a protected variable which has no getter and isn't used internally in the class.

kongtiaowang (approved):
LGTM. It needs to fix conflicts.

PR #7326, opened 2021-02-05
[Publications] Error and swal fix

laemtl:
Small changes following the comment made in #7315 
- Fix missing error message
- Fix missing swal callback change (#6195 regression) previously in #7091

### Replace #7091

### DIFF
--- a/modules/publication/jsx/uploadForm.js
+++ b/modules/publication/jsx/uploadForm.js
@@ -42,6 +42,9 @@ class PublicationUploadForm extends React.Component {
     }).then((response) => {
       if (!response.ok) {
         console.error(response.status);
+        this.setState({
+          loadError: 'An error occurred when loading the form!',
+        });
         return;
       }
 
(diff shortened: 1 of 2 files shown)


driusan:
@maltheism Maybe you can review?

maltheism:
Hi @laemtl, I'm not seeing the swal when there's an error in the publication module. Example if I'm trying to make a publication with the exact details as an existing one. I think the showPublicationError() function is to blame in the ajax/FileUpload.php

maltheism (changes requested):
Hi @laemtl, I'm not seeing the swal when there's an error in the publication module. Example if I'm trying to make a publication with the exact details as an existing one. I think the showPublicationError() function is to blame in the ajax/FileUpload.php

laemtl:
I just tried and saw the error popup: 
Screenshot from 2021-02-11 19-40-42

maltheism (approved):
Hi @laemtl, it works now for me. I'm unsure why it wasn't before. Maybe I didn't delete my cache, sorry!

PR #7327, opened 2021-02-05
[permissions] Renaming and organizing permissions phase 0

ridz1208:
## Brief summary of changes
In an attempt to organise and standardise the existing permissions I have created a couple categories in the permissions table to leverage the modules architecture and add the `action` column and generate the lable based on the combination of columns as a first phase.

Changes in this PR:
- addition of 2 new columns composing the permission label
- deletion of 4 unsused permissions
- renumbering of entries in permission patch
- swap of the permission and modules patch (01 and 02)
- RB permission changes
- logic for label generation for permissions

new permissions......
image


test/unittests/UserTest.php
```diff
@@ -151,27 +151,48 @@ class UserTest extends TestCase
...
         3 => ['permID' => 4,
             'code'        => "test_permission3",
             'description' => "description 3",
-            'categoryID'  => 4
+            'categoryID'  => 4,
+            'action'      => 'View/Create',
+            'moduleID'    => 5
         ]
     ];
+
```
ridz1208, line 184: @driusan @kongtiaowang 

I had to modify the test class to have the PR pass checks, the problem is, because I need toinstantiate a module to get its Long name, i resorted to use real modules (candidate_list and timepoint_list) which means I am king of breaking the rules of unit testing here because I am not only testing the `getPermissionsVerbose()` function but also the `getActiveModulesIndexed()` function of the module class. I know we should instead be using a mock module but I'm not sure how to do that. let me know if and how you want this PR changed

raisinbread/RB_files/RB_user_perm_rel.sql
```diff
@@ -7,16 +7,12 @@ INSERT INTO `user_perm_rel` (`userID`, `permID`) VALUES (1,3);
 INSERT INTO `user_perm_rel` (`userID`, `permID`) VALUES (1,4);
 INSERT INTO `user_perm_rel` (`userID`, `permID`) VALUES (1,5);
 INSERT INTO `user_perm_rel` (`userID`, `permID`) VALUES (1,6);
-INSERT INTO `user_perm_rel` (`userID`, `permID`) VALUES (1,7);
```
driusan, line 10: How did these get deleted from RB? There doesn't seem to be any `DELETE` patch in this PR to have been sourced..

ridz1208: @driusan yes theres a cleanup
SQL/Cleanup_patches/2021-01-31-deleting_permissions.sql

driusan: Oh weird, I don't know how I missed that with ctrl-F

laemtl:
@ridz1208 #6770 is kind of related. Do you mind taking a look into it? It's open to discussion and was following Issue #6687.

driusan (changes requested):
This is missing changes to the filenames in the php/installer directory

ridz1208:
@driusan changes made, tests running

ridz1208:
@driusan anything still needed for this one ? you want someone else to review first ?

driusan:
I just re-skimmed it, but it looks good to me.

PR #7330, opened 2021-02-08
[issue_tracker] Fix incorrect url to issues in notification email

zaliqarosli:
## Brief summary of changes

Currently, the notification email sent by the issue tracker module has the old url . This PR fixes this issue by removing the lines of code completely because the url is correctly set to the new url on lines 534-537

1.

### DIFF
--- a/modules/issue_tracker/ajax/EditIssue.php
+++ b/modules/issue_tracker/ajax/EditIssue.php
@@ -572,11 +572,6 @@ function emailUser($issueID, $changed_assignee)
         )
     );
 
-    $msg_data['url']         = $baseurl .
-        "/issue_tracker/issue/?issueID=" . $issueID;
-    $msg_data['issueID']     = $issueID;
-    $msg_data['currentUser'] = $user->getUsername();
-
     foreach ($issue_change_emails as $email) {
         $msg_data['firstname'] = $email['firstname'];
         Email::send($email['Email'], 'issue_change.tpl', $msg_data);


ridz1208:
@zaliqarosli I don't think we are sending stuff that far back unless its a critical security issue. I think in this case its about time for projects to move forward to 23

ridz1208 (dismissed):

ridz1208:
@CamilleBeau can you give this one a quick test today ?

CamilleBeau:
I tested this and did not get the expected emails. After setting a user with my email to "watching" on an issue, and then editing that issue by commenting or changing a field from the admin account, I received this email off of both this branch _and_ the 23.0-release branch: 
image

This was unexpected for the following reasons:
- This email was an "assigned" email instead of a "change" email, even thought there was only a comment being added, or a field other than "assigned" being changed.
- Off of the 23.0 branch, the link in the email did not contain "?issueID=n" as expected
- The link was not clickable. 

I tested this with RB data on my dev using both ssh tunnels and VPN.

zaliqarosli:
it's expected for the link to not be clickable because you are running the instance locally @CamilleBeau

ridz1208:
@zaliqarosli 

> it's expected for the link to not be clickable because you are running the instance locally @CamilleBeau

What do you mean ? shouldnt it still have the baseurl which is "localhost" in that case ?

zaliqarosli:
> @zaliqarosli
> 
> > it's expected for the link to not be clickable because you are running the instance locally @CamilleBeau
> 
> What do you mean ? shouldnt it still have the baseurl which is "localhost" in that case ?

@ridz1208 yeah it looks like maybe her baseurl isn't set up properly because its not showing up at all

zaliqarosli:
ooh @CamilleBeau could it be a site issue? are you affiliated with the site that the issue is affiliated with?

driusan:
There is no baseurl variable to set.. it was removed from LORIS a long time ago and should be derived from the incoming request now.

zaliqarosli:
@driusan in that case, then the base url got lost between 21 and 23 (see example where 21.0 is the bottom email)

Screen Shot 2021-02-11 at 3 00 16 PM

should this go in a different PR or added to this one?

ridz1208:
@driusan any idea on what might be happening ? seems to be instantiated properly...

driusan:
@ridz1208 I would guess it's because it's in the ajax directory and the factory's not returning it correctly.

driusan:
The title of the issue is "Fix incorrect url to issues in notification email".. if it's not being set I think it should be fixed in this one..

zaliqarosli:
@driusan do you have an idea of how to fix this?

driusan:
I think the EditIssue.php needs to be converted to a real LORIS endpoint and the issue would take care of itself. I don't know of any way to get the base url from a script in the ajax directory.

zaliqarosli:
@driusan @ridz1208 ooff that will take me weeks to figure out. is there someone we can assign this PR who knows how to quickly convert to a real LORIS endpoint? I haven't had to do that in over 2 years

pierre-p-s:
Hi @zaliqarosli. We ran into some issues when we moved this version of issue tracker to CCNA. I sent you a message on slack if you want to discuss it and I could also send you the files we have.

zaliqarosli:
unfortunately CCNA's fixes are on 21.0 where the baseurl is still working fine

zaliqarosli:
@ridz1208 do you have an update on what our next steps for this PR is?

ridz1208:
@zaliqarosli 

@HenriRabalais seems to believe that this will require a partial module revamp to get that functionality back and given that the upcoming bugfix release is already substantial and overdue we decided that we would push for this fix for the next bugfix release. so not 23.0.3 but 23.0.4 and we will take the time to run a full test on the module make sure everythign is functional before release

@HenriRabalais when can you start working on it ?

zaliqarosli:
@ridz1208 let me know if i need to change branches or anything for this PR

pierre-p-s:
Hi @zaliqarosli @ridz1208 @driusan 
We are having the same issue on CCNA since we upgraded to 23.0. Would anyone have time to explain why the getBaseURL function does not work anymore in the issue tracker module emails please?

driusan:
The baseURL is derived from the incoming request. Since the ajax/ directory bypasses LORIS, it does not have it set.

pierre-p-s:
> The baseURL is derived from the incoming request. Since the ajax/ directory bypasses LORIS, it does not have it set.

Thank you @driusan.
Do you think that hardcoding the base url is a safe solution until the refactoring of the module happens?

cmadjar:
@maltheism could you work on this? Ask @driusan or @ridz1208 what needs to be done for this one.

cmadjar:
@maltheism could you work on this one? Thanks! Ask @driusan what needs to be done for this one.

maltheism:
@cmadjar I've been working on this one and I've asked already.

PR #7334, opened 2021-02-12
Remove references to User->getData()

driusan:
The `getData` function is a code smell which we still have scattered throughout our code. It loses type information and generally makes it harder for static analysis tools (or humans) to figure out what our code is doing.

This updates most of the references to use the User class's getters for the appropriate field, which does not lose type information.

There are a couple exceptions which were not updated: 1. 2 places in the code use `getData()` to send the all the user's data to a template (`my_preferences` and `edit_user`). 2. Unit tests for the user class use it to get test the internal state. These situations need to be fixed/tested separately before the function could be completely removed.

### DIFF
--- a/src/Middleware/UserPageDecorationMiddleware.php
+++ b/src/Middleware/UserPageDecorationMiddleware.php
@@ -197,14 +197,14 @@ class UserPageDecorationMiddleware implements MiddlewareInterface
         }
 
         $oneIsStudySite   = in_array("1", $isStudySite);
-        $tpl_data['user'] = $this->user->getData();
+        $tpl_data['user'] = [];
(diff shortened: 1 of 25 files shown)


kongtiaowang (approved):
LGTM

src/Middleware/UserPageDecorationMiddleware.php
```diff
@@ -197,14 +197,14 @@ function ($a, $b) {
         }
 
         $oneIsStudySite   = in_array("1", $isStudySite);
-        $tpl_data['user'] = $this->user->getData();
+        $tpl_data['user'] = [];
```
ridz1208, line 200: @driusan this was a very bold move lol

PR #7335, opened 2021-02-12
Remove most references to Candidate->getData

driusan:
This is analagous to #7334 but for `Candidate->getData`. It removes calls to `getData` in the candidate class and replaces them with appropriate getters. There are a few places where the array for getData is passed to a smarty template and some unit tests which rely on it, and these are untouched as they need more thorough testing 

This should improve the ability of phan to track our type information while it analyses the code.

### DIFF
--- a/modules/create_timepoint/php/create_timepoint.class.inc
+++ b/modules/create_timepoint/php/create_timepoint.class.inc
@@ -48,11 +48,7 @@ class Create_Timepoint extends \NDB_Form
         // check user permissions
         return (
             $user->hasPermission('data_entry') &&
-            (in_array(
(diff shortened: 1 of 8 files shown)


kongtiaowang:
imaging_browser/viewSession doesn't work.
Screen Shot 2021-02-26 at 11 58 50 AM

driusan:
@kongtiaowang There seems to be a bug with getEntityType unrelated to this PR where it's using the wrong key.. I fixed it (and another bug where the EntityType class was *only* allowing invalid entity types), but I'm not sure why it would have started in this PR. It seems to be more related to PHP8.

kongtiaowang (approved):
LGTM

PR #7336, opened 2021-02-13
[Imaging Browser] Fix nullable type fatal error

laemtl:
Fix for
```
PHP Fatal error:  Uncaught TypeError: Typed property LORIS\\imaging_browser\\MRINavigation::$_FilteredSessionList must be array, null used in /var/www/html/loris-core/modules/imaging_browser/php/mrinavigation.class.inc:48
```

### DIFF
--- a/modules/imaging_browser/php/mrinavigation.class.inc
+++ b/modules/imaging_browser/php/mrinavigation.class.inc
@@ -46,12 +46,12 @@ class MRINavigation
     {
         $this->_sessionID = $sessionID;
 
-        $this->_FilteredSessionList = $_SESSION['State']
-            ->getProperty('mriSessionsListed');
-
-        if (!is_array($this->_FilteredSessionList)) {
-            $this->_FilteredSessionList = [];
+        $this->_FilteredSessionList = [];
+        if (is_array($_SESSION['State']->getProperty('mriSessionsListed'))) {
+            $this->_FilteredSessionList
+                = $_SESSION['State']->getProperty('mriSessionsListed');
         }
+
         $this->_currentListIndex = array_search(
             $this->_sessionID,
             $this->_FilteredSessionList


modules/imaging_browser/php/mrinavigation.class.inc
```diff
@@ -29,7 +29,7 @@ class MRINavigation
 {
     private array $_bits;
     private int $_currentListIndex;
-    private array $_FilteredSessionList;
+    private ?array $_FilteredSessionList;
```
driusan, line 32: How does it get set to null? In the constructor right below this it's initialized as:

```
        $this->_FilteredSessionList = $_SESSION['State']
            ->getProperty('mriSessionsListed');
        if (!is_array($this->_FilteredSessionList)) {
            $this->_FilteredSessionList = [];
        }
```

laemtl: Here:
```
$this->_FilteredSessionList = $_SESSION['State']
            ->getProperty('mriSessionsListed');
```

the value returned by `$_SESSION['State']->getProperty('mriSessionsListed')` is null. The if case is not evaluated because a fatal is triggered, as the type of $this->_FilteredSessionList does not accept null value.

That could be fixed with something like:


```
$this->_FilteredSessionList = $_SESSION['State']
    ->getProperty('mriSessionsListed') ?? [];
```

driusan: That makes sense. The check isn't done until after the assignment..

I think it would be better to not change the type, so that places dealing with the property don't need to worry about the distinction between null and empty array

modules/imaging_browser/php/mrinavigation.class.inc
```diff
@@ -46,11 +46,8 @@ class MRINavigation
     {
         $this->_sessionID           = $sessionID;
         $this->_FilteredSessionList = $_SESSION['State']
-            ->getProperty('mriSessionsListed');
+            ->getProperty('mriSessionsListed') ?? [];
 
-        if (!is_array($this->_FilteredSessionList)) {
```
driusan, line 51: I think this needs to stay. It's theoretically possible for the property to be a non-null non-array value, and in that case we still want it to be treated as the empty array.

laemtl, line 51: I removed this with the assumption that $_SESSION['State']->getProperty('mriSessionsListed') is either of type array or null.

laemtl: Update -- I will adjust to take your previous comment into account.

modules/imaging_browser/php/mrinavigation.class.inc
```diff
@@ -45,12 +45,11 @@ class MRINavigation
     function __construct($sessionID = null)
     {
         $this->_sessionID           = $sessionID;
-        $this->_FilteredSessionList = $_SESSION['State']
-            ->getProperty('mriSessionsListed');
+        $this->_FilteredSessionList
+            = is_array($_SESSION['State']->getProperty('mriSessionsListed'))
+            ? $_SESSION['State']->getProperty('mriSessionsListed')
+            : [];
```
driusan, line 51: I don't think this is a good use of the ternary operator. It makes the code too opaque. An if/else would be easier to read.

PR #7337, opened 2021-02-15
Add interface to replace method_exists('getCenterID') checks

driusan:
This adds 2 interfaces, SiteHaver and MultiSiteHaver to implement on classes that implement `getCenterID` and `getCenterIDs` functions respectively. This replaces the lossy method_exists checks for the 2 methods, which only verifies that a method with that name exists, and doesn't validate any type information such as arguments or return types. (In fact, it turns out there's a class in the examiners module which has an unrelated getCenterID($sitename) function, it's just luckily never passed to a pass which does a method_exists check).

This has a few benefits:
1. Static analysis tools are better able to analyze types
2. Some code can be simplified, since the types/nullability are explicit in the interface they don't need to be verified inside the method_exists checks
3. Some hacks that were necessary because of the method_exists checks (such as assigning arguments to a temporary variable and coercing the types for phan to recognize the methods exist inside of the method_exists blocks) aren't necessary inside of instanceof checks. 
3. If we change the signature (or decide other methods should be added) it can be updated in the interface to ensure we don't miss any classes that were implicitly implementing the getCenterID contract.

I'm not crazy about the interface names, but I couldn't think of anything better. The downside is that classes now need to explicitly implement the interface in their signature in order to have the data filters work on them, but I've updated all the ones that currently exist in LORIS.

### DIFF
--- a/php/libraries/User.class.inc
+++ b/php/libraries/User.class.inc
@@ -284,20 +284,6 @@ class User extends UserPermissions implements \LORIS\StudyEntities\AccessibleRes
         return explode(';', $this->userInfo['Sites']);
     }
 
-     /**
-      * Get the user's site's ID number
-      *
-      * @return array
-      */
-    function getCenterID()
-    {
-        throw new \LorisException(
-            "The function getCenterID
-                                   is deprecated and is replaced
-                                   with getCenterIDs"
-        );
-    }
-
     /**
      * Get the user's sites' ID numbers
      *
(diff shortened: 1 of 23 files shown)


php/libraries/User.class.inc
```diff
@@ -284,20 +286,6 @@ class User extends UserPermissions implements \LORIS\StudyEntities\AccessibleRes
         return explode(';', $this->userInfo['Sites']);
     }
 
-     /**
-      * Get the user's site's ID number
-      *
-      * @return array
-      */
-    function getCenterID()
```
driusan, line 292: It seemed silly to implement an interface just to throw an exception, so I removed this.

kongtiaowang (approved):
LGTM

PR #7339, opened 2021-02-16
[phan] Fix all PhanTypeMismatchArgument errors

driusan:
This fixes all PhanTypeMismatchArgument errors and unignores the error type from our configuration. The result should be that phan can now catch situations where methods are called with the wrong types.


php/libraries/NDB_Menu_Filter.class.inc
```diff
@@ -374,13 +345,6 @@ class NDB_Menu_Filter extends NDB_Menu
             $this->_setSearchKeyword($key);
 
         }
-
-        // 4. If the user clicked on a sort column, set the order.
-        //    It's likely a GET request even though it modifies the data,
-        //    so just use $_REQUEST.
-        if (isset($_REQUEST['filter']['order'])) {
```
driusan, line 381: This seems to be dead code, the filtering/order is done on the react side now. The types get confused because it's a confusing request, so I just removed it (and the test)

driusan:
@laemtl @kongtiaowang could one of you review/test this? I think this is the last of the high priority phan rules.. after this we won't have undefined properties, undefined methods, *or* methods called with the wrong argument type..

kongtiaowang (approved):
LGTM

PR #7344, opened 2021-02-22
[23.0][Instrument] Fix freeze on completed instruments

jesscall:
## Brief summary of changes
Bugfix for disabling fields when instrument is set to complete. 

1. Make sure that fields are disabled when setting an instrument to Data Entry = `Complete`


jesscall:
@driusan Can you help with this?

Currently the page needs to be refreshed for freeze to take effect.

driusan:
I can't test very well because the 23 branch doesn't work with php 8, but it seems to work for me when I redirect in the process function if the Data Entry status changes. It doesn't work in the handle function because of the order of operations where the `NDB_BVL_InstrumentStatus_ControlPanel->save` method gets called.

```
diff --git a/php/libraries/NDB_BVL_Instrument.class.inc b/php/libraries/NDB_BVL_Instrument.class.inc
index 44aa274a3..82d12a5c5 100644
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -11,7 +11,9 @@
  */
 use \Psr\Http\Message\ServerRequestInterface;
+use \Psr\Http\Server\RequestHandlerInterface;
+
 use \Loris\StudyEntities\Candidate\CandID;
 
 /**
@@ -51,6 +53,12 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     public $XINRules;
     public $XINDebug;
 
+    /**
+     * A flag determining whether the ControlPanel changed
+     * the Data_entry flag status on this page load.
+     */
+    private $DataEntryChanged;
+
     /**
      * Test name (short name), equivalent to test_names.Test_name
      *
@@ -2402,6 +2410,8 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             $config = NDB_Config::singleton();
             $paths  = $config->getSetting('paths');
 
+            $this->DataEntryChanged = false;
+
             // check if the file/class exists
             $BasePath = $paths['base'] . 'project/instruments/';
             $TestName = $this->testName;
@@ -2412,7 +2422,12 @@ abstract class NDB_BVL_Instrument extends NDB_Page
                 || file_exists($BasePath . "$TestName.linst")
             ) {
                 // save possible changes from the control panel...
+                $originalstatus = $controlPanel->getDataEntryStatus();
                 $controlPanel->save();
+
+                if ($controlPanel->getDataEntryStatus() !== $originalstatus) {
+                    $this->DataEntryChanged = true;
+                }
             }
 
             // display the control panel
@@ -2822,6 +2837,33 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         return $this->jsonData;
     }
 
+      public function process(
+        RequestHandlerInterface $handler
+    ) : ResponseInterface {
+        // The parent middleware needs to be called in order for
+        // the instrument control panel to saved, which triggers
+        // $this->DataEntryChanged being set appropriately.
+        $response = parent::process($request, $handler);
+
+        if ($this->DataEntryChanged) {
+            $sessionID = $request->getQueryParams()["sessionID"];
+            $candID    = $request->getQueryParams()["candID"];
+
+            $baseURL = \NDB_Factory::singleton()->settings()->getBaseURL();
+
+            $url = $baseURL . "/instruments/" .
+                urlencode($this->testName) .
+                '?commentID=' . urlencode($this->getCommentID()) .
+                '&candID=' . urlencode($candID) .
+                '&sessionID=' . urlencode($sessionID);
+                ->withStatus(303)
+                ->withHeader("Location", $url);
+        }
+        return $response;
+      }
+
     /**
      * A form extends the basic page handler to
      *  (1) call save() on a POST request
@@ -2839,14 +2881,14 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         }
 
         // Disable form if data entry is complete
-        $commentID = $request->getAttribute("pageclass")->commentID;
         $user      = $request->getAttribute("user");
         $sessionID = $request->getQueryParams()["sessionID"];
+        $candID    = $request->getQueryParams()["candID"];
         $timepoint = \Timepoint::singleton($sessionID);
 
         // create an instrument status object
         $status = new \NDB_BVL_InstrumentStatus;
-        $status->select($commentID);
+        $status->select($this->commentID);
 
         // freeze the form to prevent data entry
         if ($status->getDataEntryStatus() != 'In Progress'
@@ -2863,7 +2905,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             }
         }
 
-        return (new \Laminas\Diactoros\Response())
             ->withBody(new \LORIS\Http\StringStream($this->display() ?? ""));
     }
 }
```

kongtiaowang:
@jesscall to pass test, you need to add "_$this->DataEntryType = 'DirectEntry';_ " into the  setup function of "test/test_instrument/NDB_BVL_Instrument_testtest.class.inc"

kongtiaowang (approved):
LGTM

test/test_instrument/NDB_BVL_Instrument_testtest.class.inc
```diff
@@ -42,6 +42,9 @@ class NDB_BVL_Instrument_testtest extends NDB_BVL_Instrument
 
         // setup the form
         $this->_setupForm();
+
+	// set data entry type
+	$this->DataEntryType = 'DirectEntry';
```
ridz1208, line 47: something is wrong with the indentation here. could you fix it ? @jesscall

ridz1208 (changes requested):

driusan:
Why is testtest a DirectEntry instrument? AFAIK it's not used for testing surveys

test/test_instrument/NDB_BVL_Instrument_testtest.class.inc
jesscall: @ridz1208 done

kongtiaowang:
https://github.com/aces/Loris/pull/7369

jesscall:
Testing suite blocked by #7369

@driusan is it possible to have that merged soon?

.github/workflows/loristest.yml
```diff
@@ -24,6 +24,7 @@ jobs:
       uses: shivammathur/setup-php@v2
       with:
         php-version: ${{ matrix.php }}
+        extensions: php-ast
```
kongtiaowang, line 27: add "zip" as well

.github/workflows/loristest.yml
```diff
@@ -41,9 +41,6 @@ jobs:
     - name: Change PHP Version in Dockerfile
       run: sed -i "s/7.3/${{ matrix.php }}/g" Dockerfile.test.php7
 
-    - name: Install OS package dependencies
-      run: sudo apt-get install php-ast
-
```
ridz1208, line 46: I dont think these changes should be happening

test/test_instrument/NDB_BVL_Instrument_testtest.class.inc
ridz1208: These still show up as bad indent and I dont think they should be there at all

ridz1208 (changes requested):
there seems to be hanges to files that are unrelated to the PR. I think this might be the result of a bad rebase.

test/test_instrument/NDB_BVL_Instrument_testtest.class.inc
jesscall: not sure why it's not working, hold on

test/test_instrument/NDB_BVL_Instrument_testtest.class.inc
```diff
@@ -42,6 +42,9 @@ class NDB_BVL_Instrument_testtest extends NDB_BVL_Instrument
 
         // setup the form
         $this->_setupForm();
+
+        // set data entry type
+        $this->DataEntryType = 'DirectEntry';
```
driusan, line 47: Why is this set to DirectEntry? As far as I know this test instrument isn't used as a survey instrument

jesscall: I believe it is needed because the following logic was added in this PR and it affects testing.
@kongtiaowang can maybe provide more details

```
            if ($this->preview !== true && $this->DataEntryType !== 'DirectEntry') {
                $this->freeze();
            }
```

kongtiaowang: Hi @driusan ,the integration test of testtest instrument will test the direct input information. I added DirectEntry since @Jesscall have modifed the DataEntryType rule.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2835,8 +2889,32 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+            && !in_array(
+                $timepoint->getData("CenterID"),
+                $user->getData("CenterIDs")
+            ))
+        ) {
+
+            if ($this->preview !== true && $this->DataEntryType !== 'DirectEntry') {
+                $this->freeze();
+            }
+        }
```
ridz1208, line 2915: Why do we have nested IFs ? there is no alternate to the second if, it could just be combined with the first

test/test_instrument/NDB_BVL_Instrument_testtest.class.inc
ridz1208: You have the following 5 conditions that would make the instrument be frozen
1. Instrument is NOT in progress
2. the user does not have data_entry permission
3. the user does not have access to the timepoint at which the instrument is instantiated
4. the instrument is not in preview mode
5. the instrument is not a direct entry

Why choose to solve the problem with option 5 ? I would say options 1, 2 and 3 are your easy logical solutions !

ridz1208 (changes requested):

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2835,8 +2889,32 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+        $status = new \NDB_BVL_InstrumentStatus;
+        $status->select($this->commentID);
+
+        // freeze the form to prevent data entry
+        if ($status->getDataEntryStatus() != 'In Progress'
+            || !$user->hasPermission('data_entry')
+            || (isset($timepoint)
+            && !in_array(
+                $timepoint->getData("CenterID"),
+                $user->getData("CenterIDs")
```
ridz1208, line 2908: please add the third argument `true` to the in_array function to avoid unexpected behaviour

test/test_instrument/NDB_BVL_Instrument_testtest.class.inc
driusan: I think Rida means "make the instrument not be frozen". Assuming he does, I agree. "Direct entry" means survey mode and doesn't get frozen because when it's set, the participant is entering data directly (after the entry is complete they can no longer access the URL..)

If the test doing data entry on an instrument that was supposed to be frozen and is now broken because the freeze bug is fixed, the solution should be to fix the test and/or data, not fake being a survey instrument for testing non-survey functionality in order to work around the freeze call.

php/libraries/NDB_BVL_Instrument.class.inc
jesscall: just realized the `&&` on 2912 should likely be a `||` operator

test/test_instrument/NDB_BVL_Instrument_testtest.class.inc
jesscall: Thanks for the feedback. 
I agree with Dave, it's likely an issue with the test doing data entry on an instrument meant to be frozen and this fix is messing with that. In that case the test case should be updated. 

Pushing Rida's recent feedback and will monitor the test suite logs for indications of where the issue might be.

php/libraries/NDB_BVL_Instrument.class.inc
driusan: I think `$user->hasCenter($timepoint->getCenterID())` would be better so that it doesn't need to change after #7359 (and it's more explicit what it's trying to do..)

kongtiaowang:
hi @jesscall, when the user has data_entry permission or status is in_progress, the form always is frozen.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2835,8 +2889,38 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+            || !$user->hasPermission('data_entry')
+            || (isset($timepoint)
+            && !in_array(
+                $timepoint->getData("CenterID"),
+                $user->getData("CenterIDs"),
+                true
+            ))
+            || $this->preview !== true
+            || $this->DataEntryType !== 'DirectEntry'
+        ) {
```
ridz1208, line 2919:
```suggestion
        if ($status->getDataEntryStatus() != 'In Progress'
            || !$user->hasPermission('data_entry')
            || (isset($timepoint)
            && !in_array(
                $timepoint->getData("CenterID"),
                $user->getData("CenterIDs"),
                true
            ))
            && ($this->preview !== true
            || $this->DataEntryType !== 'DirectEntry')
        ) {
```

driusan: "Move the if into the same statement", they said.

"That'll make it simpler," they said.

kongtiaowang:
@jesscall please add "'Data_entry'=> 'In Progress', "  into test/test_instrument/test_instrumentTest.php at line 58, it will pass the integration test.

ridz1208 (approved):

ridz1208:
@jesscall this is in your court

PR #7369, opened 2021-03-02
[Travis] add php extension

kongtiaowang:
add missing zip ext


driusan:
what is the zip extension used for?

kongtiaowang:
@driusan We found a new issue in Travis last week.
  Problem 1
    - php-webdriver/webdriver is locked to version 1.9.0 and an update of this package was not requested.
    - php-webdriver/webdriver 1.9.0 requires ext-zip * -> it is missing from your system. Install or enable PHP's zip extension.

driusan (approved):

PR #7372, opened 2021-03-02
[candidate_list] Move Project column over to array definition

zaliqarosli:
## Brief summary of changes

This PR moves the Project field column over to where the fields array is defined. At the moment, it is being pushed to the fields array, which will create a misalignment of columns in the data table if a new column is added to the end of where the array is initialized. 

1. On the main branch, and this PR branch, test the front-end of the candidate_list module. Make sure that there are no changes to the data table, specifically, the Project column and its neighbours.

### DIFF
--- a/modules/candidate_list/jsx/candidateListIndex.js
+++ b/modules/candidate_list/jsx/candidateListIndex.js
@@ -322,19 +322,16 @@ class CandidateListIndex extends Component {
           },
         },
       },
-
-    ];
-     fields.push(
-        {
-          'label': 'Project',
-          'show': true,
-          'filter': {
-            name: 'project',
-            type: 'select',
-            options: options.project,
-          },
+      {
+        'label': 'Project',
+        'show': true,
+        'filter': {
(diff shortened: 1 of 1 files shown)


ridz1208 (approved):

driusan:
I don't have any problem with this change but it looks like it's just stylistic to me. I don't see how 

`x = [..]; x.push(y);` is semantically different than `x = [..y]`. There's nothing in between the initialization and the push in the diff.

ridz1208:
@driusan you need to be on a project to understand XD

driusan:
Should we just close this or is there an explanation for why it's not just a stylistic change?

ridz1208:
@driusan we should merge it regardless. why push something in an array when its not conditional ?!

zaliqarosli:
@driusan the reason is that projects wanting to add extra fields will have their data table column headers and values misaligned because 'Project' field is hardcoded as the last field in the table when it doesn't need to be the last field always

PR #7373, opened 2021-03-02
[Instruments] Conditional Field Display

laemtl:
### Usage

```
        // Question 1
        $sexOptions = array(
            ""             => "",
            "male"         => $this->getLocale("male"),
            "female"       => $this->getLocale("female"),
            "intersex"     => $this->getLocale("intersex"),
            "not-answered" => $this->getLocale("not-answered"),
            "other"        => $this->getLocale("other"),
        );

        $this->addSelect(
            "biological_sex",
            $this->getLocale("biological-sex"),
            $sexOptions,
            ['required' => true]
        );

        $this->addBasicText(
            "biological_sex_other",
            "$this->indent $this->indent" . $this->getLocale("specify")
        );
        
        $this->addDisplayRule(
            "biological_sex_other",
            [
                [
                    "name" => "biological_sex",
                    "operator" => "is",
                    "value" => "other",
                ]
            ]
        );
```


php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1418,6 +1421,27 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+     *
+     * @return void
+     */
+    function addDisplayRule(
+        string $field,
+        array $rules
+    ) {
+        $this->displayRules[] = [
+            'field' => $field,
+            'container' => '.form-group',
```
laemtl, line 1438: maybe better to make this configurable via parameters with default values.

jesscall:
Should the conditional field be nulled if it goes from showing to hidden ?

Ex:
Biological Sex = 'Other', Specify = ' Some explanation'

If sex is changed to any other option, the specify field disappears but the value is still in the db.

laemtl:
@jesscall That hasn't been done yet for simplicity but I agree that I should find a way to clean conditional data that are not part of the POST request.

jesscall:
@laemtl Another option could be to only hide the conditional field if it's empty && does not satisfy original condition.

Other than this comment however, the PR looks good and works as expected.

laemtl:
@jesscall Unfortunately hiding the field if empty will not work as the field, once empty, will not be submitted in the post array if hidden.

cmadjar:
@jesscall could you reply to @laemtl ?

jesscall:
@laemtl Sure, if it is too complex let's go with the simpler option of just setting the values to NULL in the POST request!

htdocs/vendor/js/mf-conditional-fields.js
```diff
@@ -0,0 +1,487 @@
+/**
```
maltheism, line 1: Hi @laemtl and @driusan, should this file be in htdocs/vendor/js or htdocs/js?

laemtl: I am confused by the difference between the two. Usually vendor/ stores libraries from external developers and js/ stores  code written internally.

driusan: External things should be in vendor but it's possible that there are some leftover things from before we created the directory directly in js.

ridz1208:
@laemtl if this is still functional. Please move it to `modules/instruments/js/` because the `vendor/` directory can be removed and rebuilt.

@driusan if the licensing is all in order, its not a bad easy addition

laemtl:
@ridz1208 Ready for your second review.

PR #7375, opened 2021-03-08
[API] Fix v0.0.3 of API

driusan:
The switches doing a version check for the API
were missing break statements, causing all versions
to fall through to the default and have the `$view`
variable immediately overwritten by the dev version,
for all versions of the API.

This adds a break so that the correct code is used.


driusan:
@xlecours could you review?

xlecours (approved):

PR #7376, opened 2021-03-08
[instrument_list] Fix side effect in conditional

driusan:
The check for `$nextStage == 'Recycling Bin'` in the
instrument_list_controlpanel was doing an assignment,
not an equality check.


driusan:
@ridz1208 probably worth fixing for v23? If not I can switch to main.

ridz1208 (approved):
:joy:

PR #7377, opened 2021-03-09
[eslint] Fix babel loader and eslint webpack plugin conflict

laemtl:
Since eslint loader has been replaced by  eslint webpack plugin (#7137) a conflict babel-loader prevent eslint to run.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,5 @@
 node_modules/*
+vendor/*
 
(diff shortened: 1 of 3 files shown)


maltheism (approved):
Reference https://github.com/webpack-contrib/eslint-webpack-plugin/issues/59
@driusan ready to merge.

PR #7379, opened 2021-03-09
[Bvl Feedback] jQuery cleanup

laemtl:


kongtiaowang:
"Comment added!" and "The new thread has been submitted!" should be the notice message.
before we have to modify the message to save a new comment. if not, it will save an empty comment.
This PR keeps the notice message as new content after clicking the send button. 
Screen Shot 2021-03-25 at 11 25 29 AM

cmadjar:
@jstirling91 could you please review and test?

laemtl:
@cmadjar @kongtiaowang Fixed (FYI this bug is already present in main and has nothing to do with this PR).

modules/bvl_feedback/ajax/new_bvl_feedback.php
```diff
@@ -19,14 +19,20 @@
 //Creating a new array to pass the set values into the DB.
 $newThreadValues = [];
 
+// With formData, null is passed as a string
+foreach ($_POST as $key => $val) {
+    if ($_POST[$key] === 'null') {
```
jstirling91, line 24: Seems like this would be a useful function in other places. Might be worth moving this function to the Utilities Class

jstirling91 (approved):

PR #7381, opened 2021-03-10
[core] fileElement multitype value fix

laemtl:
The value props of a fileElement can be of type string or Object, but the assignment to filename only covers the object case.
As a result, in media, if we edit a file, the fileElement does not display the file name.
This PR aims to fix this.


kongtiaowang (approved):
LGTM

PR #7383, opened 2021-03-10
[document_repository] Solves special characters issue

pierre-p-s:
## Brief summary of changes

Solves an issue when if using special characters (&, <, >, ") the file download fails and the name of the file appears incorrectly in the browse tab of the document repository

1. Upload a file with special charcacters &, <, >, " into the document repository.
2. Look for the file in the file tree, check that the file name appears correctly
3. Try downloading the file and opening it.

### DIFF
--- a/modules/document_repository/php/files.class.inc
+++ b/modules/document_repository/php/files.class.inc
@@ -116,7 +116,7 @@ class Files extends \NDB_Page
                 $uploadedBy = $DB->pselectOne(
                     "SELECT uploaded_by FROM document_repository
                     WHERE File_Name=:name",
-                    array('name' => htmlspecialchars($record))
+                    array('name' => $record)
                 );
                 $file       = $path. "$uploadedBy/$record";
                 return (new \LORIS\Http\Response())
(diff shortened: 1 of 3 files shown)


kongtiaowang:
I only tested the file name with ">" or "<".  It works.
I can't create a file name with '"' in my OS.

kongtiaowang (approved):
LGTM

pierre-p-s:
@ridz1208 If we could add this PR to the 23 bugfix release it would be great

pierre-p-s:
This PR still needs a script to change the file names in the sql table as well as in /data/document_repository_uploads

modules/document_repository/php/files.class.inc
```diff
@@ -348,6 +348,7 @@ class Files extends \NDB_Page
         $uploadedFile = $uploadedFiles['file'];
         $fileSize     = $uploadedFile->getSize();
         $fileName     = $uploadedFile->getClientFileName();
+        $fileName     = urldecode($fileName);
```
ridz1208, line 351:
```suggestion
        // urldecode() necessary to decode double quotes encoded automatically 
        // by chrome browsers to avoid XSS attacks
        $fileName     = urldecode($fileName);
```

pierre-p-s:
The script works on my VM but I cant seem to find why the test suite is failing.

pierre-p-s:
Thanks @kongtiaowang

kongtiaowang (approved):
LGTM

tools/single_use/Cleanup_Special_Chars_Document_Repository.php
```diff
@@ -0,0 +1,55 @@
+<?php
```
ridz1208, line 1: before SQL

PR #7386, opened 2021-03-11
[media] Special characters issue

pierre-p-s:
## Brief summary of changes

Solves an issue when if using special characters (&, <, >, ") the file download fails and the name of the file appears incorrectly in the browse tab of the media module.
The script removes the special characters from the sql data table media as well as from the file system.

1. Before checking out this branch, on 23.0-release, upload a file with special characters &, <, >, " into media.
2. Check that the file name appears incorrectly in the media table
3. Check out this branch.
4. Upload a different file with special characters &, <, >, " into media.
5. Look for the file in the browse tab, check that the file name appears correctly
6. Try downloading the file and opening it.
7. run (need sudo to be able to change the name of the files under /data/media_uploads/): sudo php /var/www/loris/tools/single_use/Cleanup_Special_Chars_Media.php
8. Check that the first file now appears correctly in the data table and that when downloading you are able to read the file correctly.

### DIFF
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -130,7 +130,7 @@ function uploadFile()
     checkDateTaken($dateTaken);
 
     $fileName  = preg_replace('/\s/', '_', $_FILES["file"]["name"]);
-    $fileName  = str_replace("%22", "\"", $fileName);
+    $fileName  = urldecode($fileName);
     $fileType  = $_FILES["file"]["type"];
     $extension = pathinfo($fileName)['extension'];
 
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -116,7 +116,7 @@ function uploadFile()
     // If required fields are not set, show an error
     if (empty($_FILES)) {
         showMediaError(
-            "File could not be uploaded successfully. 
+            "File could not be uploaded successfully.
             Please contact the administrator.",
             400
         );
@@ -130,6 +130,7 @@ function uploadFile()
     checkDateTaken($dateTaken);
 
     $fileName  = preg_replace('/\s/', '_', $_FILES["file"]["name"]);
+    $fileName  = str_replace("%22", "\"", $fileName);
     $fileType  = $_FILES["file"]["type"];
     $extension = pathinfo($fileName)['extension'];
 
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -179,7 +179,7 @@ function uploadFile()
     if (move_uploaded_file($_FILES["file"]["tmp_name"], $mediaPath . $fileName)) {
         try {
             // Insert or override db record if file_name already exists
-            $db->insertOnDuplicateUpdate('media', $query);
+            $db->unsafeInsertOnDuplicateUpdate('media', $query);
             $uploadNotifier->notify(array("file" => $fileName));
         } catch (DatabaseException $e) {
             showMediaError("Could not upload the file. Please try again!", 500);
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -129,7 +129,7 @@ function uploadFile()
 
     checkDateTaken($dateTaken);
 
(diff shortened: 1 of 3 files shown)


pierre-p-s:
The test suite seems to be failing when running make checkstatic however on my VM the command runs without errors.

pierre-p-s:
Blocked by #7390

pierre-p-s:
After discussion with Rida: Remove encoding when inserting into SQL table, line 186 of fileUpload use unsafe insert and fix issues in the database

pierre-p-s:
@ridz1208 Let me know if this is a better fix please.

modules/media/ajax/FileUpload.php
```diff
@@ -130,6 +130,7 @@ function uploadFile()
     checkDateTaken($dateTaken);
 
     $fileName  = preg_replace('/\s/', '_', $_FILES["file"]["name"]);
+    $fileName  = str_replace("%22", "\"", $fileName);
```
ridz1208, line 133: does this mean you are accepting quotes in the name of the file ? also its a bit clearer if you use `'"'` in the second argument

pierre-p-s: Well I wanted to include the double quotes because right now we can add a file with it in the name. However we need this check fo it to work. Otherwise I could also send an error through swal refusing the quotes, but in that case what else are we not allowing?

ridz1208: `%22` is URI encoding, your $filename is never coming from the URI... I'm not clear on why its being "decoded" (and BTW decoding should be done with `urldecode()` https://www.php.net/manual/en/function.urldecode.php not manually like this)

pierre-p-s: I changed it to urldecode but I cant find where the fileName is being encoded...

pierre-p-s:
This PR still needs a script to change the file names in the sql table as well as in /data/uploads

modules/media/ajax/FileUpload.php
```diff
@@ -130,6 +130,7 @@ function uploadFile()
     checkDateTaken($dateTaken);
 
     $fileName  = preg_replace('/\s/', '_', $_FILES["file"]["name"]);
+    $fileName  = urldecode($fileName);
```
ridz1208, line 133:
```suggestion
    // urldecode() necessary to decode double quotes encoded automatically 
    // by chrome browsers to avoid XSS attacks
    $fileName  = urldecode($fileName);
```

pierre-p-s: Ill add the comment when I add the script

jesscall (approved):

jesscall:
Looks good!

tools/single_use/Cleanup_Special_Chars_Media.php
```diff
@@ -0,0 +1,54 @@
...
+            [
+                "file_name" => $fileName
+            ],
+            [
+                "id" => $file['id']
+            ]
+        );
+
+        // update name in file system
+        shell_exec("mv " . escapeshellarg($media_path . $fileNameURLencoded) . " " . escapeshellarg($media_path . $fileName));
```
driusan, line 51: Why isn't this just using `rename`?

https://www.php.net/manual/en/function.rename.php

driusan (changes requested):

tools/single_use/Cleanup_Special_Chars_Media.php
driusan: I missed it while reviewing the document_repository PR, but can you change it in that script too?

pierre-p-s: I changed it to rename

pierre-p-s: changed the doc repo script in this PR: #7428

pierre-p-s: @driusan let me know if it needs more changes

tools/single_use/Cleanup_Special_Chars_Media.php
```diff
@@ -0,0 +1,54 @@
+<?php
```
ridz1208, line 1: before SQL

PR #7389, opened 2021-03-12
Fix all PhanUnusedVariable errors

driusan:
Fix all PhanUnusedVariable class errors.

### DIFF
--- a/test/unittests/CandidateTest.php
+++ b/test/unittests/CandidateTest.php
@@ -515,6 +515,7 @@ class CandidateTest extends TestCase
      */
     public function testGetListOfVisitLabels()
     {
+        $this->markTestSkipped("Test is incomplete");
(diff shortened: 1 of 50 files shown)


test/unittests/CandidateTest.php
```diff
@@ -515,6 +515,7 @@ public function testLastRecordChangeBy()
      */
     public function testGetListOfVisitLabels()
     {
+        $this->markTestSkipped("Test is incomplete");
```
driusan, line 518: I had to uncomment the assertEquals line because otherwise "$expected" is unused, but then the test fails because it seems to have never worked properly. (The select in getListOfVisitLabels was never mocked, there's just a line comment saying to mock it in the test..)

kongtiaowang (approved):
LGTM

PR #7398, opened 2021-03-18
[Utility] Deprecate old getInstrument functions and replace with new ones

ridz1208:
Instrument Full names are no longer queried from the database `test_names` unless the instrument implements the LegacyTrait. that means that new "properly" implemented instruments should derive their full names directly from the instrument class.

The deprecated functions in this PR wrongly assume that the full name comes from the database table. the functions were deprecated to allow a sow transition to the newly created functions. The following 2 reasons explain why a deprecation was better than simply modifying the existing functions:

1. The old functions were in the Utility class which is a hodge-podge of function with no home. These functions belong more to the Instrument class in our libraries.
2. the is an added limitation to the new system of getting the instrument's full name. The instrument needs to be valid and instantiable, if the instrument can not be instantiated it's name will not be in the list of instruments

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2778,4 +2778,118 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     abstract public function getDataDictionary() : iterable;
 
+    /**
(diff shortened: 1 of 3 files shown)


kongtiaowang (approved):
LGTM

PR #7400, opened 2021-03-18
[EEG] Sites/Project/subproject filters and permission

### LINKED ISSUE #6691: [Imaging Browser] project filter ComboBox shows all projects (including the ones the user doesn't have permissions to see)

### LINKED ISSUE #6829: [Permissions] Update project filter based on  user access

laemtl:
This PR:
1. modifies the way project/subproject/site filter are populated. Now those filters are displayed with values the current user has permission for.
3. add some integration tests

## To test
* Visit the module and check that all data entries belong to sites and projects the user has access to.
* Visit the module and check that the site/project/subproject menu filters contain entries the user has access to.

Part of #6829


cmadjar (approved):
@laemtl The changes work perfectly!

Should the test plan be updated? I see you updated the tests. Maybe it is not needed to update the test plan if it is automatically tested? I don't remember. @laemtl @driusan @ridz1208 Thoughts?

christinerogers:
@driusan could please review ?   This is important for some of the EEG work going on in the next few weeks. 
Thank you

christinerogers (approved):
LGTM

php/libraries/NDB_Page.class.inc
```diff
@@ -927,6 +927,110 @@ class NDB_Page implements RequestHandlerInterface
...
+     *                           should be numeric (ProjectID)
+     *                           or strings (project name)
+     *
+     * @return array - an associative array of the form (ID => name)
+     */
+    public function getProjectOptions(
+        bool $numeric_keys = true
+    ): array {
+        $projectsOptions = [];
+        $userProjectsIDs = \NDB_Factory::singleton()->user()->getProjectIDs();
```
driusan, line 944: This looks like it would make more sense as a parameter, but if it's a parameter I don't see why this function should be in the NDB_Page class at all. It would make more sense to have a `getProjects` method to the `User` class that returns an iterable of `\Project` instances. It could then be massaged into any format wanted by `array_map`/`array_values`/etc. A user's projects doesn't really have anything to do with the page.

php/libraries/NDB_Page.class.inc
```diff
@@ -927,6 +927,110 @@ class NDB_Page implements RequestHandlerInterface
...
+     * Get the list of subprojects for which the current user has permission
+     * To populate the module subprojects filter options
+     *
+     * @param bool $numeric_keys If the returned subprojects keys
+     *                           should be numeric (SubprojectID)
+     *                           or strings (subproject name)
+     *
+     * @return array - an associative array of the form (ID => name)
+     */
+    public function getSubprojectOptions(
```
driusan, line 970: Similarly to User projects, Subprojects don't seem to be related to pages. If the user had a `getProjects` method this would just be an `array_reduce` on `$user->getProjects()`

php/libraries/NDB_Page.class.inc
```diff
@@ -927,6 +927,110 @@ class NDB_Page implements RequestHandlerInterface
...
+     */
+    public function getSiteOptions(
+        bool $numeric_keys = true,
+        array $allsitesPermission = ['access_all_profiles'],
+        bool $study_site = true
+    ): array {
+        $user         = \NDB_Factory::singleton()->user();
+        $site_options = $user->getStudySites();
+
+        foreach ($allsitesPermission as $permission) {
```
driusan, line 1017: This one's a little more complicated since it incorporates the site parameter, but given that our menu filters should all be extending `DataFrameworkMenu` where there's a `abstract public function allSitePermissionNames() : ?array;` function I don't think that's a problem. This could still be moved there instead of the NDB_Page class.

php/libraries/NDB_Page.class.inc
```diff
@@ -927,6 +927,110 @@ class NDB_Page implements RequestHandlerInterface
...
+            if ($user->hasPermission($permission)) {
+                $site_options = \Utility::getSiteList($study_site);
+                break;
+            }
+        }
+
+        if (!$numeric_keys) {
+            foreach ($site_options as $id => $name) {
```
driusan, line 1025: This loop is duplicated in each of your three function. Does this loop do anything that `array_column` doesn't?

https://www.php.net/manual/en/function.array-column.php

If not, I don't see the need for it, and if so, maybe it should be a Utility function.

driusan (changes requested):

laemtl:
@driusan Ready for a second look.

php/libraries/DataFrameworkMenu.class.inc
```diff
@@ -183,5 +183,91 @@ abstract class DataFrameworkMenu extends NDB_Menu_Filter
...
+    /**
+     * Format the list of user's projects to display the module filter options
+     *
+     * @param bool $numericKeys If the returned projects keys
+     *                          should be numeric (ProjectID) or project name
+     *
+     * @return array - an associative array of the form (ID|mame => name)
+     */
+    public function getProjectOptions(
+        bool  $numericKeys = true
```
driusan, line 196: Why does this option exist? (Same with on the helper functions).. it doesn't seem to ever be true.

php/libraries/DataFrameworkMenu.class.inc
```diff
@@ -183,5 +183,91 @@ abstract class DataFrameworkMenu extends NDB_Menu_Filter
...
+    ): array {
+        $projects     = \NDB_Factory::singleton()->user()->getProjects();
```
driusan, line 198: If you need a user, I think it should be passed as a parameter.

php/libraries/DataFrameworkMenu.class.inc
```diff
@@ -183,5 +183,91 @@ abstract class DataFrameworkMenu extends NDB_Menu_Filter
...
+     *                          should be numeric (SubprojectID)
+     *                          or strings (subproject name)
+     *
+     * @return array - an associative array of the form (ID => name)
+     */
+    public function getSubprojectOptions(
+        bool $numericKeys = true
+    ): array {
+        $subprojectOptions = array_reduce(
+            \NDB_Factory::singleton()->user()->getProjects(),
```
driusan, line 228: Same, should be function parameter.

php/libraries/DataFrameworkMenu.class.inc
```diff
@@ -183,5 +183,91 @@ abstract class DataFrameworkMenu extends NDB_Menu_Filter
...
+     * To populate the module subprojects filter options
+     *
+     * @param bool $numericKeys If the returned subprojects keys
+     *                          should be numeric (SubprojectID)
+     *                          or strings (subproject name)
+     *
+     * @return array - an associative array of the form (ID => name)
+     */
+    public function getSubprojectOptions(
+        bool $numericKeys = true
```
driusan, line 225: Also don't see anywhere this is used..

php/libraries/DataFrameworkMenu.class.inc
```diff
@@ -183,5 +183,91 @@ abstract class DataFrameworkMenu extends NDB_Menu_Filter
...
+            fn($subprojects, $project) =>
+                $subprojects
+                + array_column($project->getSubprojects(), 'title', 'subprojectId'),
+            []
+        );
+
+        if (!$numericKeys) {
+            $subprojectNames = array_values($subprojectOptions);
+            return array_combine($subprojectNames, $subprojectNames);
+        }
```
driusan, line 238: in the above function you have
```
if (..) {
   return 1
} else {
   return 2
}
```

But in this one the style is:

```
if (...) {
    return 1
}
return 2
```

You should be consistent (Stylistically I prefer the second, but as long as it's consistent it doesn't really matter which..)

php/libraries/DataFrameworkMenu.class.inc
```diff
@@ -183,5 +183,91 @@ abstract class DataFrameworkMenu extends NDB_Menu_Filter
...
+     * @param bool $studySite   If the returned sites should be study sites only
+     *
+     * @return array An associative array of the form (ID|name => name)
+     */
+    public function getSiteOptions(
+        bool $numericKeys = true,
+        bool $studySite = true
+    ): array {
+        $user        = \NDB_Factory::singleton()->user();
+        $siteOptions = $user->getStudySites();
```
driusan, line 257: You can make this an else of the hasPermission check. There's no reason for the overhead if you're immediately overwriting it.

php/libraries/DataFrameworkMenu.class.inc
```diff
@@ -183,5 +183,91 @@ abstract class DataFrameworkMenu extends NDB_Menu_Filter
...
+                                numeric (CenterID) or strings (site name)
+     * @param bool $studySite   If the returned sites should be study sites only
+     *
+     * @return array An associative array of the form (ID|name => name)
+     */
+    public function getSiteOptions(
+        bool $numericKeys = true,
+        bool $studySite = true
+    ): array {
+        $user        = \NDB_Factory::singleton()->user();
```
driusan, line 256: parameter

php/libraries/User.class.inc
```diff
@@ -306,6 +306,19 @@ class User extends UserPermissions implements
...
+    /**
+     * Get the user's projects
+     *
+     * @return Project[]
+     */
+    function getProjects(): array
+    {
+        return array_map(
+            fn($projectID) => \Project::getProjectFromID($projectID),
+            $this->userInfo['ProjectIDs']
```
driusan, line 318: This should be `$this->getProjectIDs()`. It makes it easier for static analysis tools (and humans) to determine the type.

driusan (changes requested):
mostly minor comments

php/libraries/DataFrameworkMenu.class.inc
laemtl: Some module filters works with index keys (half uses indexes, half names):
- modules/mri_violations/php/mri_violations.class.inc
- modules/mri_violations/php/resolved_violations.class.inc
- modules/behavioural_qc/php/behavioural.class.inc
- modules/behavioural_qc/php/conflicts.class.inc
- modules/behavioural_qc/php/incomplete.class.inc
- modules/media/php/media.class.inc
- modules/new_profile/php/new_profile.class.inc
- modules/statistics/php/stats_behavioural.class.inc
- modules/statistics/php/stats_demographic.class.inc
- modules/statistics/php/stats_mri.class.inc
- modules/user_accounts/php/edit_user.class.inc

php/libraries/DataFrameworkMenu.class.inc
laemtl: Some module filters works with index keys (half uses indexes, half names):
- modules/mri_violations/php/mri_violations.class.inc
- modules/mri_violations/php/resolved_violations.class.inc
- modules/behavioural_qc/php/behavioural.class.inc
- modules/behavioural_qc/php/conflicts.class.inc
- modules/behavioural_qc/php/incomplete.class.inc
- modules/media/php/media.class.inc
- modules/new_profile/php/new_profile.class.inc
- modules/statistics/php/stats_behavioural.class.inc
- modules/statistics/php/stats_demographic.class.inc
- modules/statistics/php/stats_mri.class.inc
- modules/user_accounts/php/edit_user.class.inc

php/libraries/DataFrameworkMenu.class.inc
laemtl: I will make this change but I'm curious to know more about the reasoning behind this. It will help if I understand the motivations of using a parameters (and why user should be defined in each module getFieldsOptions instead).

driusan: Because it's a functional dependency of the method, so it should be explicit. A singleton is a hidden global variable. I think once you get up to getFieldOptions you don't have a choice but to use the singleton since you can't change the signature, but the dependencies should be as explicit as possible (which also makes it easier to inject values for testing.)

laemtl: Thanks for the clarifications.

laemtl:
@driusan All your comments have been addressed.

PR #7408, opened 2021-03-24
[Utility/Instrument] Replace simple calls to getAllInstruments() by getInstrumentNamesList()

ridz1208:
1. Replace calls to getAllInstruments() function (about to be deprecated) by getInstrumentNamesList as a followup to #7398.
2. added sorting to the Instrument class functions to reduce differences with old deprecated functions.

This PR is not exhaustive!! not all instances were replaced here, only the ones where instruments were queried only for their names. Other instances where instruments are queried to instantiate them and extract other data have not yet been included here.

### DIFF
--- a/modules/api/php/views/project.class.inc
+++ b/modules/api/php/views/project.class.inc
@@ -54,7 +54,9 @@ class Project
         return [
             'Meta'        => $meta,
             'Candidates'  => $this->_project->getCandidateIds(),
-            'Instruments' => array_keys(\Utility::getAllInstruments()),
+            'Instruments' => array_keys(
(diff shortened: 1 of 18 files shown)


ridz1208:
@kongtiaowang I would like to remove the tests for the old functions but I don't want to do that before having tests for the new functions. could you create these tests in another PR ? (see new functions in #7398 )

driusan:
The signature for the functions looks wrong since it doesn't say where to get the things from.. I probably should have caught this in #7398 but I didn't.. can you review #7409 before going too far down this rabbit hole since adding the explicit parameter might complicate things a little bit outside of web requests?

PR #7410, opened 2021-03-25
[Instrument] add sorting to getInstrumentNamesList()

ridz1208:
added sorting to the Instrument class functions to reduce differences with old deprecated functions.

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2830,6 +2830,9 @@ abstract class NDB_BVL_Instrument extends NDB_Page
             $instrumentNames[$testName] =$fullName;
         }
 
+        // sort array by instrument name for modules that rely on ordering
+        asort($instrumentNames, SORT_STRING | SORT_FLAG_CASE | SORT_NATURAL);
+
         return $instrumentNames;
     }
 
@@ -2860,6 +2863,9 @@ abstract class NDB_BVL_Instrument extends NDB_Page
                 $instrumentNames[$testName] = $fullName;
             }
         }
+
+        // sort array by instrument name for modules that rely on ordering
+        asort($instrumentNames, SORT_STRING | SORT_FLAG_CASE | SORT_NATURAL);
 
         return $instrumentNames;
     }
@@ -2894,6 +2900,9 @@ abstract class NDB_BVL_Instrument extends NDB_Page
                 $instrumentNames[$testName] = $fullName;
             }
         }
+
+        // sort array by instrument name for modules that rely on ordering
+        asort($instrumentNames, SORT_STRING | SORT_FLAG_CASE | SORT_NATURAL);
 
         return $instrumentNames;
     }


ridz1208:
@CamilleBeau could you test this PR please ? (shouldnt take more than 5 minutes).

Context, in PR #7398 I created 3 new getInstrument functions and deprected the old ones in the utility class, I also updated one of the scripts using the old functions to use the new ones (see https://github.com/aces/Loris/pull/7398/files#diff-a94bbfce132587b164266e65a79529d4b6a8752dd5c8fbc696961af0f6e6c007L22) I realized later that the old functions were also returning instruments in a sorted manner so In this PR i simply added the sorting to the new functions.

To test:
In the script that I modified fix_candidate_age.php, add the following:
```
print_r($instruments);
die();
```
On line 23 (right after the creation of the array) and try it out with 3 cases
1. using the old functions `Utility::getAllInstruments`
2. using the new functions without this PR (on branch main) `\NDB_BVL_Instrument::getInstrumentsList()`
3. using the new functions WITH this PR `\NDB_BVL_Instrument::getInstrumentsList()`

let me know what the ordering looks like

CamilleBeau:
Tried testing with fix_candidate_age.php with the statements you listed. 
Adding `\NDB_BVL_Instrument::getInstrumentsList()`, there was an error because a loris instance argument needs to be passed through getInstrumentsList. When I did it with the Loris instance argument, I got this error: 

`PHP Fatal error:  Class NDB_BVL_Instrument_aosi contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (NDB_BVL_Instrument::getDataDictionary) in /var/www/loris/project/instruments/NDB_BVL_Instrument_aosi.class.inc on line 2
`
Am I missing something in how I was doing it or is this a bug that needs to be fixed?

driusan:
It sounds like you need to use the `LorisFormDictionaryImpl` trait in your instrument.

ridz1208:
@CamilleBeau https://github.com/aces/Loris/blob/main/raisinbread/instruments/NDB_BVL_Instrument_aosi.class.inc#L5

The RB instruments have changed recently

CamilleBeau:
@ridz1208 So I'm testing now. Did you mean to test with \NDB_BVL_Instrument::getInstrument**Names**List() instead of \NDB_BVL_Instrument::getInstrumentsList()?

Testing with steps 1, 2, and 3 from your comment but with \NDB_BVL_Instrument::getInstrumentNamesList(), arrays are similar and have the same order (alphabetical), except that compared to `getAllInstruments`, `getInstrumentNamesList` adds a newline to the title of BMI Calculator. Here is a screenshot of the vardump / print_r to illustrate: 

image

Otherwise, if you did mean getInstrumentsList, the result is the same alphabetical order but a multi-dimensional array with details about the instrument & form in both main & this PR

ridz1208:
@CamilleBeau yes I didn mean getinstrumentNamesList()

and results look good. My guess is the new line is somewhere in the code or dataabse, dont think its a bug in the instrument class

CamilleBeau:
Ok cool! In that case, code & manual test look good.

CamilleBeau (approved):

PR #7411, opened 2021-03-25
[publication] Give publication creator edit permissions

### LINKED ISSUE #7401: [publication] Edit Permissions not assigned to creator of proposal
**Describe the bug**
When a publication proposal is submitted, the user creating the issue should be able to go back and edit the form. This is currently not the case. I believe the permission for the creator is never inserted into `publication_users_edit_perm_rel`

**To Reproduce**
1. Go to `Publication`
2. Create & submit a project proposal
3. Try accessing the created proposal and notice that you cannot edit the fields (other than status)

To be able to edit the form.

jesscall:
## Brief summary of changes
This PR ensures that the creator of a publication project is added to the `publication_users_edit_perm_rel` table and given edit access to their submission.

1. Before checking out this PR, go to `Publication`
2. Create & submit a project proposal
3. Try accessing the created proposal and notice that you cannot edit the fields (other than status)
4. Checkout this PR
5. Repeat steps 1-3 and make sure that this time you can edit the fields

* Resolves #7401

### DIFF
--- a/modules/publication/ajax/FileUpload.php
+++ b/modules/publication/ajax/FileUpload.php
@@ -100,6 +100,15 @@ function uploadPublication() : void
     $db->insert('publication', $fields);
     $pubID = $db->getLastInsertId();
 
+    // give creator permission to edit
+    $db->insert(
+        'publication_users_edit_perm_rel',
+        [
+            'PublicationID' => $pubID,
+            'UserID'        => $uid,
+        ]
+    );
+
     try {
         // process files
         processFiles($pubID);


modules/publication/ajax/FileUpload.php
```diff
@@ -100,6 +100,15 @@ function uploadPublication() : void
...
 
+    // give creator permission to edit
+    $db->insert(
+        'publication_users_edit_perm_rel',
+        [
+            'PublicationID' => $pubID,
+            'UserID'        => $uid,
+        ]
+    );
+
```
pierre-p-s, line 111: I think admin should always be added to users with edit permissions

```suggestion
    $db->insert(
        'publication_users_edit_perm_rel',
        [
            'PublicationID' => $pubID,
            'UserID'        => $uid,
        ]
    );


    $db->insert(
        'publication_users_edit_perm_rel',
        [
            'PublicationID' => $pubID,
            'UserID'        => 'admin',
        ]
    );
```

jesscall: `admin` is only a CCNA username. Other projects may have different names. For example I believe raisinbread uses `lorisadmin`. Also `admin` is not the `userID`.

Should I just add `userID => 1`? Is this good practice ? @ridz1208

ridz1208: correct, but also, we don't really want to automatically assign admin permissions in that way. If the goal is to give admin access to everything then this should be handled where you check for the permissions not where you assign the permissions (and even at that I don't think admin should be able to have access and edit everything).

ridz1208 (approved):

ridz1208:
@pierre-p-s can you confirm manual tests pass ?

pierre-p-s:
> @pierre-p-s can you confirm manual tests pass ?

Yes it passed manual test

pierre-p-s (approved):

PR #7412, opened 2021-03-29
[data_release] Refactor Module

driusan:
This refactors the data release module to get rid of the scripts in the `ajax` directory, and replace them with real LORIS endpoints that return JSON data.

The uploading and downloading is also refactored to use the `FilesUploadHandler` and `FilesDownloadHandler` classes instead of custom written PHP validation in the ajax scripts themselves.

## Testing Instructions

The frontend behaviour should be unchanged, so the normal module's test plan should (hopefully?) test all the behaviour..


kongtiaowang (approved):
LGTM

PR #7413, opened 2021-03-29
[JSX] Refactor and Update Pagination

HenriRabalais:
## Brief summary of changes
This PR refactors the Pagination component by changing it from a class component to a functional component that utilizes React Hooks. It also removes the bootstrap styling, replacing it with entirely custom dependency-free styling.

## Display
Screen Shot 2021-03-29 at 20 27 00
Screen Shot 2021-03-29 at 20 28 08

## Grander Vision
This is the first PR in a series that I am planning to propose over the coming months. This is with the main objective of refactoring much of the now outdated jsx code found in LORIS. In the process of refactoring, I would like to accomplish the following:
1. Convert React components from Class Components to Functional Components that use Hooks state management
- Functional components and Hooks are the lastest in React development, but have now been out for several years and have had time to be rigourously tested and seem to be very encouraged by the community. From my use of them, I find they tend to lead to simplified code that results in less lines and increased legibility.
- References: 
 -- Class vs. Functional: https://www.twilio.com/blog/react-choose-functional-components)
 -- React Hooks: https://reactjs.org/docs/hooks-intro.html
2. Reduce our reliance on Bootstrap Boilerplate
- Bootstrap was a good launching pad for LORIS to develop a cohesive design aesthetic. However, the bootstrap "look" is both outdated and also leads to stylesheet graveyards as we (rightfully) begin to overwrite the original bootstrap classes with our own custom css. However, I believe keeping the formatting inline with the React Components will increase our ability to quickly access and make changes to css when necessary since they will be found within the same file.
- This will also make it easier to have global "themes" that permeate through LORIS and will be easily changeable from a single location (e.g. dark vs light themes). I believe this to be a much more modern approach to styling LORIS which will result in less time wasted looking and sorting through style sheets and make customizability and quick changes easier.
3. Incorporate New Components
- In the creation of the Biobank module, I've added many components and made alterations to ones that currently exist in LORIS. In the process of refactoring the LORIS Components, I would like to both include my modification as well as incorporate the new components that I've written.
4. Create a Design Theory for LORIS
- It's no secret that LORIS has inconsistence when it comes to the UI throughout the platform. I am hoping that through this process, a series of standardized layouts components can be created, accompanied by a design "theory" for their use that will act as guidelines for how modules and pages within LORIS should be organized to maintain a consistent and coherent user experience.
- This will include specific sets of panels, buttons, fields, menus, modals, dropdowns etc. that have a unified design with the flexibility of the users needs in mind (e.g. using 'rem' instead of 'px'  and making the webpage mobile and tablet friendly).
- Reference for use of REM: https://www.24a11y.com/2019/pixels-vs-relative-units-in-css-why-its-still-a-big-deal/

### DIFF
--- a/jsx/PaginationLinks.js
+++ b/jsx/PaginationLinks.js
@@ -38,16 +38,14 @@ function PaginationLink({
   };
 
   return (
-    <>
-      <div
-        style={style}
-        onClick={onClick}
-        onMouseOver={hoverOn}
-        onMouseOut={hoverOff}
-      >
-         {children}
-      </div>
-    </>
+    <div
+      style={style}
+      onClick={onClick}
+      onMouseOver={hoverOn}
+      onMouseOut={hoverOff}
+    >
+       {children}
+    </div>
   );
 };
 
--- a/jsx/PaginationLinks.js
+++ b/jsx/PaginationLinks.js
@@ -1,148 +1,135 @@
-/* exported RPaginationLinks */
+import React, {useState} from 'react';
 
-import React, {Component} from 'react';
-import PropTypes from 'prop-types';
+const theme = {
+  primary: '#053665',
+  secondary: '#A6D3F5',
+  accent: '#E89A0C',
+  neutral: '#DDD',
+};
 
 /**
- * Pagination component
+ * Produces the pagination number links
+ *
+ * @param {jsx} children
+ * @param {bool} active
+ * @param {func} onClick
+ *
+ * @return {jsx}
  */
-class PaginationLinks extends Component {
-  /**
-   * @constructor
-   * @param {object} props - React Component properties
-   */
-  constructor(props) {
-    super(props);
+function PaginationLink({
+  children,
+  active,
+  onClick,
+}) {
+  const [hover, setHover] = useState(false);
+  const hoverOn = () => setHover(true);
+  const hoverOff = () => setHover(false);
+  const style = {
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    margin: '0.2rem',
+    width: '3rem',
+    height: '3rem',
+    background: onClick && hover && !active && theme.neutral,
+    border: active && '0.1rem solid '+theme.primary,
+    borderRadius: '50%',
+    cursor: onClick && 'pointer',
+  };
 
-    this.state = {
+  return (
+    <>
+      <div
+        style={style}
+        onClick={onClick}
+        onMouseOver={hoverOn}
+        onMouseOut={hoverOff}
+      >
+         {children}
+      </div>
+    </>
+  );
+};
 
-    };
-    this.changePage = this.changePage.bind(this);
+/**
+ * Produces the series of pagination links
+ *
+ * @param {int} rowsPerPage
+ * @param {int} total
+ * @param {int} active
+ * @param {func} onChangePage
+ *
+ * @return {jsx}
+ */
+function PaginationLinks({
+  rowsPerPage = 10,
+  total,
+  active = 1,
+  onChangePage,
+}) {
+  let pageLinks = [];
+  const lastPage = Math.ceil(total / rowsPerPage);
+  if (lastPage < active) {
+    onChangePage(1);
   }
 
-  /**
-   * Called by React when the component is updated.
-   *
-   * @param {object} prevProps - Previous React Component properties
-   */
-  componentDidUpdate(prevProps) {
-    if (this.props.Total < prevProps.Total) {
-      this.props.onChangePage(1);
+  let startPage = 1;
+  let endPage = lastPage;
+  if (lastPage > 5) {
+    if (active - 2 > 1) {
+      startPage = active - 2;
+    }
+    if (active + 2 < lastPage) {
+      endPage = active + 2;
     }
   }
 
-  /**
-   * Creates an onClick Event Handler
-   * execyting this.props.onChangePage(i)
-   *
-   * @param {int} i - Page index
-   * @return {function(event)} - onClick Event Handler
-   */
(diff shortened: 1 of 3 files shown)


jsx/PaginationLinks.js
```diff
@@ -1,148 +1,135 @@
...
+    width: '3rem',
+    height: '3rem',
+    background: onClick && hover && !active && theme.neutral,
+    border: active && '0.1rem solid '+theme.primary,
+    borderRadius: '50%',
+    cursor: onClick && 'pointer',
+  };
+
+  return (
+    <>
```
maltheism, line 41: @HenriRabalais I think you don't need the `<></>` since there's only one div.

jsx/PaginationLinks.js
```diff
@@ -1,148 +1,135 @@
...
-  /**
-   * @constructor
-   * @param {object} props - React Component properties
-   */
-  constructor(props) {
-    super(props);
-
-    this.state = {
+function PaginationLink({
+  children,
```
maltheism, line 20: @HenriRabalais do you prefer your syntax approach here, more than the following?
```
const PaginationLink = (props) => {
   // props.children
}
PaginationLink.propTypes = {
  children: PropTypes.node,
  active: PropTypes.bool,
  onClick: PropTypes.func,
};
```

jsx/PaginationLinks.js
```diff
@@ -1,148 +1,135 @@
...
+  active,
+  onClick,
+}) {
+  const [hover, setHover] = useState(false);
+  const hoverOn = () => setHover(true);
+  const hoverOff = () => setHover(false);
+  const style = {
```
maltheism, line 27: @HenriRabalais I haven't tested the following suggestion on Loris but we might benefit from moving towards css classes. The top of the file can have an import of the css file needed and then we could use classNames instead of passing style objects. Ex:
`import '../../htdocs/css/Pagination.css';`

jsx/PaginationLinks.js
HenriRabalais: Yeah, you're right. I believe I added it because I was checking some styling issues, and was hoping that would fix it. I'll remove it!

maltheism:
@HenriRabalais I really agree with your bullet point, "From my use of them, I find they tend to lead to simplified code that results in less lines and increased legibility" and I hope we can someday just have function components. They take a bit of getting used to, but they really do simplify the code and make things more elegant.

jsx/PaginationLinks.js
HenriRabalais: I don't necessarily have a preference at this point. I tend to use `const ComponentName` if it's a sub component that is used exclusively by the main component contained by the file, while keeping the word `function` for components that are exported. But that's not based on any sort of standard, so I'm not tied to it.

As for propTypes, I like the idea, but I don't find them really useful from a development/testing point of view. Maybe it's just me, but I've never had useful errors thrown when I pass the wrong type of prop. Not only that, but I find they tend to be forgotten at the bottom and can be forgotten about . So I'm leaning towards declaring them in the funciton comment at the top so that the prop names, types, and defaults can all be found together. Do you have a different experience/find declaring proptypes at the bottom useful?

maltheism: @HenriRabalais I think one of @laemtl's eslint changes has a webpack error be thrown if the propsTypes aren't defined. I'm unsure if it's enabled or disabled currently. I know with syntax it's a personal preference and I'm open to being flexible. I just was curious to read your thoughts.

jsx/PaginationLinks.js
maltheism: I forgot that you're a vim user. I don't want to propose anything where the navigation needed will slow anyone down. It's somewhat fast using phpstorm to jump to where the css is located. I've been using the css classes in another project associated with Loris and it has been much nicer by having cleaner code (removing styling from the code logic). That's why I thought about it when I saw your PR. Maybe it's best to leave the idea out of this PR and just bring the discussion to a Loris meeting?

HenriRabalais: Yeah, I think the principles I outlined still hold for me even if I wasn't using vim, but I could be convinced otherwise since there really is no consensus on this and it does boil down to our choice as a team.

I would be happy to discuss it at a loris meeting. Here's an article that I believe outlines our three options:
1. Vanilla CSS (what we currently use)
2. CSS Module (what you're proposing)
3. CSS in JS ( what I'm proposing)

https://jsramblings.com/understand-the-react-styling-paradigms/

maltheism: @HenriRabalais I'm okay with continuing to use inline styling. One article for possibilities of not using inline styling 'always' would be: https://www.sderosiaux.com/articles/2015/08/17/react-inline-styles-vs-css-stupid-benchmark/

HenriRabalais: Hmm...  yeah, this is definitely concerning when it comes to the datatable.

Screen Shot 2021-03-30 at 00 23 32

maltheism: @HenriRabalais also the memory impact for users with not much ram.
```
Inline styles take way more size in the DOM, are converted more slowly from VDOM (have probably a bigger impact on memory), and take more time to be handled by the browser.

But they have no impact on performance once it’s rendered.
```

laemtl:
@HenriRabalais Let's be crazy.. what about using typescript in your refactored component? I used extensive typing for the visualization component and it really improves code semantic and debugging.

ridz1208:
@HenriRabalais Please resubmit PR from a branch on your own repo. this branch will be deleted tomorrow so try to submit it before that.

ridz1208:
incorrect issuing branch, new PR made #7419

PR #7414, opened 2021-03-29
[Core] Update NDB_Page constructor to include LorisInstance object

driusan:
This updates the NDB_Page constructor to include a LorisInstance object. The LorisInstance object represents the properties of the LorisInstance itself (such as the database connection) and can replace many factory/singleton calls, but is seldom used in LORIS because it's difficult to get a copy of the object. (It needs to be indirectly accessed from a PSR7 request attribute). After moving it to the constructor, it can now be accessed directly from any NDB_Page class or subclass in a simple class property.

I was originally planning to clean up the NDB_Page constructor more by removing unnecessary parameters (why does every page get a CommentID parameter? Why is there both CommentID and identifier? Why does the class need to be told its own name? etc) but simply adding one argument and propagating it to all the places that needed to be updated was already a fairly large PR. This also makes some other cleanup possible, such as moving other statics like `\Module::factory` to `$loris->getModule()` but, once again, that would increase the size of this PR more than I'd like.

### DIFF
--- a/htdocs/survey.php
+++ b/htdocs/survey.php
@@ -54,6 +54,8 @@ class DirectDataEntryMainPage
     var $tpl_data;
     var $caller;
 
+    private $loris;
+
     /**
      * Initialize all of the class variables and things required from the
      * REQUEST.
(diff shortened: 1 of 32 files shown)


driusan:
Blocked by #7412

kongtiaowang (approved):
LGTM, and passing test

Issue #7421, opened 2021-04-09
[instrument] _nullScores function non-backwards compatible changes


ridz1208:
**Describe the bug**
Due to inconsistency across projects in how they code instruments, score element arrays hold the score column name in different ways. Some projects use array values

```
var $scoreElements = [
        "borb_real_score",
        "borb_unreal_score",
        "borb_combined_score",
        "borb_z_score",
    ];
```

Other projects array keys
```
var $scores = array(
                   "total_score_1_18"                                 => "",
                   "total_score_1_18_override"                        => "",
                   "number_of_markers"                                => "",
                   "number_of_markers_override"                       => "",
                   "q1_visual_tracking_score"                         => "",
                   "q2_disengagement_of_attention_score"              => "",
                   "q3_orients_to_name_score"                         => "",
                   "q5_anticipatory_response_score"                   => "",
                   "q6_imitation_of_actions_score"                    => "",
                  );
```

Or sometimes both. After changes in #7161 the latter is no longer supported thus causing some project's scoring logic to fail

ridz1208:
https://github.com/aces/Loris/pull/7424

PR #7424, opened 2021-04-12
[git actions] fix unit test 

kongtiaowang:
fix unit test

### DIFF
--- a/test/unittests/NDB_BVL_Instrument_Test.php
+++ b/test/unittests/NDB_BVL_Instrument_Test.php
@@ -1549,7 +1549,7 @@ class NDB_BVL_Instrument_Test extends TestCase
         $this->_setTableData();
         $this->_instrument->commentID = 'commentID1';
         $this->_instrument->table     = 'medical_history';
-        $this->_instrument->_nullScores(['Test Examiner1'=> 'Examiner']);
+        $this->_instrument->_nullScores(['Examiner']);
         $data = $this->_instrument->getInstanceData();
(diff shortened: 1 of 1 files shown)


ridz1208:
@driusan do not merge before we discuss

driusan:
Discuss what? Isn't this to fix the broken main branch?

test/unittests/NDB_BVL_Instrument_Test.php
```diff
@@ -1549,7 +1549,7 @@ function testNullScores()
         $this->_setTableData();
         $this->_instrument->commentID = 'commentID1';
         $this->_instrument->table     = 'medical_history';
-        $this->_instrument->_nullScores(['Examiner' => 'Test Examiner1']);
+        $this->_instrument->_nullScores(['Test Examiner1'=> 'Examiner']);
```
ridz1208, line 1552:
```suggestion
        $this->_instrument->_nullScores(['Examiner']);
```

ridz1208:
@kongtiaowang I made a little change on this PR (hopefully it will pass the tests) because the issue here was caused by another PR merged on 23 #7161 when that was merged and then I pushed the changes to main the test caught a valid error.

PR #7161 changed in a non-backwards compatible way the way the null scores function works and I made an issue a few days ago that we still havent solved #7421. Either way the change I made on this PR should work regardless of what happens with #7421

ridz1208:
@driusan its passing

PR #7425, opened 2021-04-13
[23.0][instrument] Fix _nullScores() inconsistencies

### LINKED ISSUE #7421: [instrument] _nullScores function non-backwards compatible changes

jesscall:
## Brief summary of changes
See description in #7421 
This PR fixes logic so that both scoring formats are supported.

1. Check which score array format your project / raisinbread is using
2. Run `php score_instrument.php all`
3. Try implementing the opposite score array format on any instrument, make sure scoring succeeds.

* Resolves #7421

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2108,10 +2108,6 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     function _nullScores(array $scoreCols): void
     {
         $record    = $this->loadInstanceData($this);
-        $scoreCols = array_combine($scoreCols, $scoreCols);
-        // order is important here because it keeps values of first array
-        // unsetting all columns which are not scores
-        $data = array_intersect_key($record, $scoreCols);
 
         // set the scoring cols to NULL
         foreach ($scoreCols as $key => $val) {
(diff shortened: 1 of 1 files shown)


driusan:
@ridz1208 do you have any suggestions for who should review/test this?

ridz1208:
@driusan the logic here makes sense to me... but then again the original change made sense to me too

so maybe you can confirm ?

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2107,11 +2107,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     function _nullScores(array $scoreCols): void
     {
-        $record    = $this->loadInstanceData($this);
-        $scoreCols = array_combine($scoreCols, $scoreCols);
-        // order is important here because it keeps values of first array
-        // unsetting all columns which are not scores
-        $data = array_intersect_key($record, $scoreCols);
+        $record = $this->loadInstanceData($this);
```
driusan, line 2110: The loadInstanceData seems to be the cause of this problem, and thinking through it, I'm not sure I understand why this needs to be here at all.

Can't we just null the scores in `$scoreCols` without adding the data from loadInstanceData?

jesscall: You're right, I'll make these changes. 
Thanks for your feedback !

driusan:
looks good to me.

@ridz1208 does it still look good to you?

ridz1208 (approved):

PR #7428, opened 2021-04-16
[document_repository] Fix special characters script

pierre-p-s:
## Brief summary of changes

- fixes script from this PR: #7383

* See comment from #7386


pierre-p-s:
@driusan this should solve your comment in the media PR

PR #7430, opened 2021-04-21
[Behavioural QC] Add subproject filter

CamilleBeau:
## Brief summary of changes
This PR adds a subproject drop-down in the filters for Behavioural QC. This change was made on CCNA but should be relevant / useful to other projects as well.

1. Make sure that module loads properly
2. Make sure that all drop downs in the filters are populated with the correct options
3. Test the filters and make sure that the results are updated accordingly 
4. Make sure you can clear the filters, and the results are reset

### DIFF
--- a/modules/behavioural_qc/php/views/conflicts.class.inc
+++ b/modules/behavioural_qc/php/views/conflicts.class.inc
@@ -43,7 +43,6 @@ class Conflicts
             'projects'    => $projects,
             'subprojects' => $subprojects,
             'sites'       => $sites,
-            'subprojects' => $subprojects,
         ];
     }
 
--- a/modules/behavioural_qc/php/views/conflicts.class.inc
+++ b/modules/behavioural_qc/php/views/conflicts.class.inc
@@ -31,10 +31,10 @@ class Conflicts
      * @param array $sites       The sites for form element
      */
     public function __construct(
-        $visits, 
-        $instruments, 
-        $projects, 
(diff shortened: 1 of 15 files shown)


modules/behavioural_qc/php/views/conflicts.class.inc
```diff
@@ -27,15 +27,23 @@ class Conflicts
...
+        $subprojects,
+        $sites
+    ) {
         $this->_formElement = [
             'instruments' => $instruments,
             'visits'      => $visits,
             'projects'    => $projects,
+            'subprojects' => $subprojects,
             'sites'       => $sites,
+            'subprojects' => $subprojects,
```
HenriRabalais, line 46: You seem to have added subprojects twice here.

HenriRabalais (approved):
One small typo but other than that LGTM!

HenriRabalais (changes requested):
Meant to hit request changes!

modules/behavioural_qc/php/views/conflicts.class.inc
CamilleBeau, line 46:
```suggestion
            'subprojects' => $subprojects,
            'sites'       => $sites,
```

CamilleBeau: Thanks!

HenriRabalais (approved):

PR #7432, opened 2021-04-23
[TimePoint] Add TimepointData DTO class

driusan:
Add a TimePointData class to act as a data transfer object for
data related to TimePoints.

 A TimePointData object is a DTO (data transfer object) to
 store a representation of data associated with a TimePoint
 object in LORIS.
 
The TimePointData may be partially loaded, in which case any
getter trying to access an unknown will throw an exception.
This allows modules to only load relevant data in a query,
and avoid the overhead of the TimePoint::singleton() constructor.

For instance, a query that selected CenterID and ProjectID on
the database could instantiate a TimePoint as:

```
$visit = new TimePoint(new TimePointData(null, $projectid, $centerid));
```

with data loaded from the query, and then call

`$visit->isAccessibleBy($user)` without having to load all the
over data that `TimePoint::select()` does. If an attempt is
made to access a property not in the DTO (session in this example),
an exception is thrown, rather than silently returning the incorrect
value. T

### DIFF
--- a/php/libraries/TimePoint.class.inc
+++ b/php/libraries/TimePoint.class.inc
@@ -76,7 +76,7 @@ class TimePoint implements \LORIS\StudyEntities\AccessibleResource,
     /**
      * A DTO for TimePoint related data
      */
-    protected TimePointData $data;
+    protected ?TimePointData $data;
 
     /**
      * Construct a TimePoint object seeded with the data
@@ -159,8 +159,12 @@ class TimePoint implements \LORIS\StudyEntities\AccessibleResource,
 
         $this->data = new TimePointData(
             $sessionID,
-            new ProjectID($row['ProjectID']),
-            intval($row['CenterID']),
+            isset($row['ProjectID'])
+                ? new ProjectID($row['ProjectID'])
+                : null,
+            isset($row['CenterID'])
+                ? intval($row['CenterID'])
+                : null,
         );
 
         // store user data in object property
--- a/php/libraries/TimePoint.class.inc
+++ b/php/libraries/TimePoint.class.inc
@@ -74,6 +74,19 @@ class TimePoint implements \LORIS\StudyEntities\AccessibleResource,
 
 
     /**
+     * A DTO for TimePoint related data
+     */
+    protected TimePointData $data;
+
+    /**
+     * Construct a TimePoint object seeded with the data
+     * from the given DTO.
+     */
+    public function __construct(?TimePointData $data=null) {
+        $this->data = $data;
+    }
+
+    /**
      * Returns an object representing this Timepoint, and ensures that only
      * one is ever created per Session.
      *
@@ -141,6 +154,13 @@ class TimePoint implements \LORIS\StudyEntities\AccessibleResource,
             ['SID' => $sessionID]
         );
 
+
+        $this->data = new TimePointData(
+            $sessionID,
+            new ProjectID($row['ProjectID']),
+            intval($row['CenterID']),
+        );
+
         // store user data in object property
         if (is_array($row) && count($row) > 0) {
             $this->_timePointInfo = $row;
@@ -149,12 +169,6 @@ class TimePoint implements \LORIS\StudyEntities\AccessibleResource,
             );
             $title = $subprojectSettings['title'] ?? '';
             $this->_timePointInfo['SubprojectTitle'] = $title;
-
-            if ($this->_timePointInfo['ProjectID'] !== null) {
-                $this->_timePointInfo['ProjectID'] = new \ProjectID(
-                    $this->_timePointInfo['ProjectID']
-                );
-            }
         } else {
             // return error when 0 rows to prevent creation of an empty object
             throw new LorisException(
@@ -421,7 +435,7 @@ class TimePoint implements \LORIS\StudyEntities\AccessibleResource,
      */
     function getSessionID(): SessionID
     {
-        return (new SessionID($this->_timePointInfo["SessionID"]));
+        return $this->data->getSessionID();
     }
 
     /**
@@ -441,7 +455,7 @@ class TimePoint implements \LORIS\StudyEntities\AccessibleResource,
      */
     function getCenterID(): \CenterID
     {
-        return new \CenterID($this->_timePointInfo['CenterID']);
+        return $this->data->getCenterID();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


driusan:
@ridz1208 @xlecours @HenriRabalais you guys are probably the most likely to have opinions on this.

(Note that I extracted this from work I needed from the new `candidates` module where it's used exactly as described as a theoretical use case in the PR description because otherwise I couldn't check access efficiently.)

driusan:
Note2: Doing this comprehensively would be a lot easier with php8's named parameters/constructor property promotion.

xlecours (changes requested):
Minor comment/suggestion

php/libraries/TimePointData.php
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Return the SessionID for this TimePoint, or throw
+     * an exception if unknown.
+     *
+     * @return SessionID
+     */
+    public function getSessionID() : SessionID
+    {
+        if ($this->sessionID === null) {
+            throw new \Exception("No SessionID loaded into data model");
```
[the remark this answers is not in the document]
driusan: We have some stupid middleware that converts NotFound into 404 errors and already have problems with coding errors being turned into "404 Not Found" errors. Runtime could work or I could look into if there's a better SPL exception type.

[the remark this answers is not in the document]
driusan: `LogicException`? https://www.php.net/manual/en/class.logicexception.php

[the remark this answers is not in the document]
xlecours: Though call :) Both would work. The goal is to be able to distinguish between cases. You call.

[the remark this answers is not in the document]
driusan: I think it's a logic error (you tried to access data you didn't load) moreso than a runtime error (theoretically a static analysis tool that could parse/understand SQL could catch that.).. also RuntimeExceptions leave a bad taste in my mouth from what I remember of Java decades ago, so I changed it to a LogicException. I'm still willing to change it to something else though if anyone feels strongly about it.

php/libraries/TimePointData.php
```diff
@@ -0,0 +1,106 @@
...
+     * Construct a TimePointData DTO for a TimePoint object
+     *
+     * @param ?SessionID $SessionID The SessionID for the TimePoint
+     * @param ?ProjectID $ProjectID The ProjectID for the TimePoint
+     * @param ?int       $CenterID  The CenterID for the TimePoint
+     */
+    public function __construct(
+        ?SessionID $SessionID,
+        ?ProjectID $ProjectID,
+        ?int $CenterID
```
HenriRabalais, line 58: If these values can't be "got" when they are null, should they be able to be set as null in the first place?

php/libraries/TimePointData.php
```diff
@@ -0,0 +1,106 @@
...
+ * with data loaded from the query, and then call
+ * `$visit->isAccessibleBy($user)` without having to load all the
+ * over data that TimePoint::singleton() does. If an attempt is
+ * made to access a property not in the DTO (session in this example),
+ * an exception is thrown, rather than silently returning the incorrect
+ * value.
+ *
+ * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ */
+class TimePointData
```
HenriRabalais, line 25: To me, in terms of nomenclature, this object makes more sense to me as the `TimePoint`, and the current "TimePoint" is more akin to a "TimePointController" or "TimePointHandler".

php/libraries/TimePointData.php
driusan: They might not be loaded from the database if your code doesn't care about them, even though in the database there's data. How would you handle that situation if they were non-nullable in the code?

php/libraries/TimePointData.php
driusan: I don't think we can change the name of the `TimePoint` class at this point, but it could eventually be split into further classes (it's just not my concern here because this is being driven by needing more efficiency for `isAccessibleBy` for data querying.)

php/libraries/TimePointData.php
HenriRabalais: If there's situation where they can acceptably be `null` in the object, I think instead of making them non-nullable, I would just return `null` in the "getter" function rather than throw an exception. To me at least, if they can be set as `null`, then retrieving them when they are `null` should be allowed.

HenriRabalais: Perhaps if there's a validator function that you can run the object through to make sure it is in a `valid` state, then that would throw an exception for null values that shouldn't be null after having fully processed the timepoint.

driusan: Where would that be used?

driusan: and I think it would need to be in the TimePointData class because the properties are protected.

php/libraries/TimePoint.class.inc
```diff
@@ -73,6 +73,22 @@ class TimePoint implements \LORIS\StudyEntities\AccessibleResource,
     ];
 
 
+    /**
+     * A DTO for TimePoint related data
+     */
+    protected ?TimePointData $data;
```
xlecours, line 79: I am not sure why this is kept as an instance variable instead of spread into 3 instance variables (sessionId, centerId and projectId)

driusan: @ridz1208 @HenriRabalais any thoughts on inlining this into a TimePoint constructor rather than using a separate DTO object?

driusan: It doesn't make a big different to me as long as we have a way to avoid the overhead of `singleton`/`select`.

ridz1208: anything is better than what we have

xlecours: i like the idea of wraping contructors params in a type/class but I think those objects should be volatiles 🐦

xlecours: I don't think there is a volatile keyword indeed. My point is that what we need from that object is the values, not the structure. If we keep a reference to the object, I feel that we increase the coupling between the classes. I think we should take what we need and throw it away.

HenriRabalais: I think a separate DTO object makes sense to me!

PR #7443, opened 2021-05-10
[Instruments] Fix Display of 0 scores with json data

### LINKED ISSUE #7441: [Instrument] 0 values saved in json data instrument appear as null on the front end
**Describe the bug**

For one of my instrument, when using a score field, when the total score is 0 it is saved as 0 in the database (verified by querying the score in mysql), but is not displayed on the front end.

Field definition:
```
        $this->addScoreColumn(
            'Total_Depression',
            'Total Depression score:'
        );
```

Database values:
```
| {"Date_taken":"2021-04-09","Examiner":"3","Q01Anxiety":"0","Q01Depression":"0","Q02Depression":"0","Q02Anxiety":"0","Q03Anxiety":"0","Q03Depression":"0","Q04Depression":"0","Q04Anxiety":"0","Q05Anxiety":"0","Q05Depression":"0","Q06Depression":"0","Q06Anxiety":"0","Q07Anxiety":"0","Q07Depression":"0","Candidate_Age":9,"Window_Difference":270,"UserID":"admin","Total_Depression":0,"Total_Anxiety":0} |
```

image

**To Reproduce**
1. Create a jsonData->true instrument
2. Add a score column
3. Add a score function that adds a 0 to the score column
4. Hit save and view the results on the front end (the score field show up as empty instead of 0).

Zero is a valid score and should be displayed as zero.

- Chrome

driusan:
When a JSON data instrument has a score column that's saved
as 0 (the int value, not the string), the LORIS front end
does not display it, and instead displays the field as a header.

This is caused by PHP's weak equality treating `"string" == 0` as
true. Forcing the template to use a strict equality comparison
fixes this without affecting the header columns that the comparison
in the template was intended to catch.

Fixes #7441.

### DIFF
--- a/smarty/templates/instrument_html.tpl
+++ b/smarty/templates/instrument_html.tpl
@@ -193,7 +193,7 @@
 				<div class="row form-group col-xs-12">
 					{$element.html}
 				</div>
-			{elseif $element.label eq $element.html}
+			{elseif $element.label === $element.html}
 				<label class="row form-group col-xs-12">
 					{$element.label}
 				</label>


GabrielPelletier:
I just tested this and it works fine.

ridz1208 (approved):

driusan:
Thanks for testing!

PR #7445, opened 2021-05-10
[Core] Allow parallel handling of requests from same session

driusan:
This adds a session_write_close() call to the end of NDB_Client,
after the last reference to $_SESSION has been made. This causes
PHP to release the lock it holds on the session file which prevents
further incoming requests from being handled. After this point in
NDB_Client all of our session related state should be read from
the PSR request, not the superglobal.

PHP by default holds a lock on the session until it is closed in case
there is a write to the superglobal. However, this prevents other
requests from the same session from reading the variable (which
is where our login state is stored) until the end of the request,
when the lock is released. Closing it explicitly releases the lock
and allows us to handle multiple requests from the same user in parallel
instead of in series.

### DIFF
--- a/modules/mri_violations/test/mri_violationsTest.php
+++ b/modules/mri_violations/test/mri_violationsTest.php
@@ -514,6 +514,9 @@ class MriViolationsTestIntegrationTest extends LorisIntegrationTest
      */
     function testNotResolvedSearchButton()
     {
+        $this->markTestSkipped(
+            'Skipping MRI Violation tests'
+        );
         $this->safeGet($this->url . "/mri_violations/");
         //testing search by PatientName
         $this->_searchTest(
@@ -550,6 +553,9 @@ class MriViolationsTestIntegrationTest extends LorisIntegrationTest
      */
     function testNotResolvedSaveButton()
     {
+        $this->markTestSkipped(
+            'Skipping MRI Violation tests'
+        );
         $this->safeGet($this->url . "/mri_violations/");
         $this->safeFindElement(
             WebDriverBy::Name("PatientName")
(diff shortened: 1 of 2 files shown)


xlecours:
failing. see https://github.com/aces/Loris/blob/main/php/libraries/NDB_Menu_Filter.class.inc#L304

xlecours:
I think I found write operations on $_SESSION['State'] that occur after session_write_close

1. https://github.com/aces/Loris/blob/c38f1c1ea4bef8112882e0bf48c81a1c13337fc3/php/libraries/NDB_Menu_Filter.class.inc#L324
2. https://github.com/aces/Loris/blob/c38f1c1ea4bef8112882e0bf48c81a1c13337fc3/php/libraries/NDB_Menu_Filter.class.inc#L332

driusan:
Thanks.. that looks like dead code, but I think it might actually be why the MRI Violations tests are failing, since MRI Violations hasn't been converted to React yet. Gonna have to think about if there's an easier way to do this than rewriting the module for this PR..

driusan:
I think since MRI violations is the last module in https://github.com/aces/Loris/projects/8 I might as well just do it and then delete that code in this PR.

ridz1208:
This is cool !

ridz1208 (approved):

ridz1208:
Approving. noticeable difference in request speed. currenlty working on CBIGR biobank

driusan:
Note: Talked to @cmadjar before skipping the tests and we agreed to rely on manual testing the mri violations module for now, because of the significant performance implications for other modules (biobank, eeg, etc)

PR #7446, opened 2021-05-10
[Core] Remove duplicated getActiveModules function

driusan:
There was a static getActiveModules function in the module class
as well as a getActiveModules function in the LorisInstance class.

This removes the duplicate and unnecessary function, cleaning up the
references to the non-static version now that NDB_Page has a LorisInstance
object passed in the constructor and all references should be able to use
the non-static version.

### DIFF
--- a/modules/dashboard/php/dashboard.class.inc
+++ b/modules/dashboard/php/dashboard.class.inc
@@ -79,7 +79,6 @@ class Dashboard extends \NDB_Form
         // calculating the body.
         $factory = \NDB_Factory::singleton();
         $user    = $factory->user();
-
         $modules = $loris->getActiveModules();
 
         $this->_smallwidgets = [];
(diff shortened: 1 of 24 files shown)


driusan:
@kongtiaowang @ridz1208 can one of you review this? This is blocking the DQT API because, in removing the function, it also updates all the places that needed it to have access to a `\LORIS\LorisInstance` object, which is necessary for the final querying API.

modules/dashboard/php/dashboard.class.inc
```diff
@@ -79,7 +79,6 @@ class Dashboard extends \NDB_Form
         // calculating the body.
         $factory = \NDB_Factory::singleton();
         $user    = $factory->user();
-
```
ridz1208, line 82: I think we should discuss this change

ridz1208 (approved):

PR #7447, opened 2021-05-13
[docker-install]

kongtiaowang:
test:
1. install docker system
2. run "make docker-install"
3. run "docker-compose up -d --build site mysql"  - run loris with docker database   (option)
    run "docker-compose up -d --build site mysql phpmyadmin"  - run loris, docker database and phpmyadmin (option)
4. goto installdb.php page and set up database.     
Screen Shot 2021-05-13 at 9 14 41 AM


jstirling91 (approved):

apache.dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM php:8.0.1-apache
+
+
```
driusan, line 3: This docker file doesn't look like it's copying the LORIS code into the docker image.

Makefile
```diff
@@ -45,3 +45,12 @@ check: checkstatic unittests
 
 testdata:
 	php tools/raisinbread_refresh.php
+
+docker-install:
+	if [ ! -d "smarty/templates_c" ]; then mkdir smarty/templates_c; fi
+	if [ ! -d "uploadedfiles" ]; then mkdir uploadedfiles; fi
+	chmod 770 smarty/templates_c
+	chmod 770 uploadedfiles
```
driusan, line 53: I don't see why building the docker image should change things in the non-docker code ase

driusan (changes requested):
see comments

apache.dockerfile
kongtiaowang: This docker file builds a PHP-apache environment to set up Loris.
Loris code bind to docker with volume.
   " volumes:
      - .:/var/www/html
"

Makefile
kongtiaowang: It is not necessary, it has been removed.

apache.dockerfile
driusan: This isn't resolved. We need a docker file that is a docker image of LORIS. It needs to include LORIS and be self-contained.

PR #7453, opened 2021-05-17
[FilterableDataTable] Fix issue with GROUPCONCAT and multiselect

pierre-p-s:
## Brief summary of changes

When using a multiselect on a group concat set of parameters, the filters gave only perfect matches. For example, if a candidate has 2 subproject AD and MCI, setting the filter to AD will not return the candidate although it should

Issue caught after release on CCNA of v23.0.3

Also thanks @HenriRabalais for the help

### DIFF
--- a/jsx/DataTable.js
+++ b/jsx/DataTable.js
@@ -317,7 +317,7 @@ class DataTable extends Component {
         searchKey = filterData[i].toLowerCase();
         searchString = data ? data.toString().toLowerCase() : '';
 
-        match = (searchString === searchKey);
+        match = (searchString.indexOf(searchKey) > -1);
         if (match) {
           result = true;
         }
--- a/jsx/DataTable.js
+++ b/jsx/DataTable.js
@@ -317,7 +317,8 @@ class DataTable extends Component {
         searchKey = filterData[i].toLowerCase();
         searchString = data ? data.toString().toLowerCase() : '';
 
-        match = (searchString.indexOf(searchKey) > -1);
+        let searchArray = searchString.split(',');
+        match = (searchArray.includes(searchKey));
         if (match) {
           result = true;
         }
--- a/jsx/Filter.js
+++ b/jsx/Filter.js
@@ -38,7 +38,7 @@ function Filter(props) {
     const {fields} = JSON.parse(JSON.stringify(props));
     const type = fields
       .find((field) => (field.filter||{}).name == name).filter.type;
-    const exactMatch = (!(type === 'text' || type === 'date'));
+    const exactMatch = (!(type === 'text' || type === 'date' || type === 'multiselect'));
 
     if (value === null || value === '' ||
         (value.constructor === Array && value.length === 0)) {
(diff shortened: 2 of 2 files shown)


jsx/DataTable.js
```diff
@@ -317,7 +317,7 @@ class DataTable extends Component {
         searchKey = filterData[i].toLowerCase();
         searchString = data ? data.toString().toLowerCase() : '';
 
-        match = (searchString === searchKey);
+        match = (searchString.indexOf(searchKey) > -1);
```
CamilleBeau, line 320: Doing this will give a more precise result, and change to Filter.js not needed:
```suggestion
        let searchArray = searchString.split(',');
        match = (searchArray.includes(searchKey));
```

jsx/Filter.js
```diff
@@ -38,7 +38,7 @@ function Filter(props) {
     const {fields} = JSON.parse(JSON.stringify(props));
     const type = fields
       .find((field) => (field.filter||{}).name == name).filter.type;
-    const exactMatch = (!(type === 'text' || type === 'date'));
+    const exactMatch = (!(type === 'text' || type === 'date' || type === 'multiselect'));
```
pierre-p-s, line 41:
```suggestion
    const exactMatch = (!(type === 'text' || type === 'date'));
```

CamilleBeau (approved):

jsx/DataTable.js
```diff
@@ -317,7 +317,8 @@ class DataTable extends Component {
         searchKey = filterData[i].toLowerCase();
         searchString = data ? data.toString().toLowerCase() : '';
 
-        match = (searchString === searchKey);
+        let searchArray = searchString.split(',');
```
driusan, line 320:
```suggestion
        const searchArray = searchString.split(',');
```

?

HenriRabalais (changes requested):
I think this is definitely functionality worth investigating since it seems like it has a big CCNA use case. I think the issue here though is that this PR changes the functionality of the multiselect filter from operating using a `equals` paradigm, to an `includes` paradigm. In my opinion, a multiselect should function as multiple selects. Meaning, exact matching between all the selected filter fields, and all the data fields. From what I understand, the proposed PR's changes this to an `includes` or "multisearch" paradigm, analogous to a search bar that has predetermined options to choose from, which still allow the data to pass the filter even if it is a partial match. 

Ideally, I would eventually like to see a drop down beside every filter that allows the user to select the filtering style as `equals` (e.g. =), `includes` (e.g. ⊂ or ≈), `excludes` (e.g. ! or ⊄), or any other relevant filtering types. In the meantime, I personally think this functionality could go into a filter type of it's own, so as not to replace the current multiselect functionality. 

I think a discussion of this at a Loris meeting could be beneficial, but I know @laemtl and @ridz1208 had opinions on this as well.

pierre-p-s:
After discussion during loris meeting, this fix should go through and we should look into making a toggle for the filter that Henri proposed in his above comment

driusan:
@HenriRabalais can you update your review

PR #7456, opened 2021-05-28
[Media] Display instrument name

laemtl:
The field instrument in media (Upload) displays the Test_name instead of the Full_name. This PR fixes the issue.
Per @ridz1208's comment, this PR also changes the way the full name is queried in the rest of the module to use the instrument getFullName()'s  method.

### DIFF
--- a/modules/media/ajax/FileUpload.php
+++ b/modules/media/ajax/FileUpload.php
@@ -248,12 +248,23 @@ function getUploadFields()
     $user   = \User::singleton();
     $config = \NDB_Config::singleton();
 
+    $lorisinstance = new \LORIS\LorisInstance(
+        $db,
+        $config,
+        [
+            __DIR__ . "/../../../project/modules",
+            __DIR__ . "/../../",
+        ],
+    );
+
     // Select only candidates that have had visit at user's sites
     $qparam       = [];
-    $sessionQuery = "SELECT c.PSCID, s.Visit_label, s.CenterID, f.Test_name
-                      FROM candidate c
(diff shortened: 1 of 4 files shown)


ridz1208 (changes requested):
That is not the correct full name. the full name must be queried from the instrument itself with `getFullName()` that column in the database is just legacy

laemtl:
@ridz1208 In this case, the logic applied here (https://github.com/aces/Loris/blame/main/modules/media/php/media.class.inc#L100) should be revisited too.

ridz1208:
@laemtl certaainly seems like it. Back then however the full name was in the DB. 👍

laemtl:
@ridz1208 Comment addressed.

laemtl:
@driusan Fixed.

ridz1208 (dismissed):

driusan:
That might be the case if the new LorisInstance instantiation is the only thing that needs to be moved out, but if the whole factory call does then ` $record["Test_name"]` is only available in the loop. You'd need to create a separate loop before the loop to create a dictionary of test_name => full name mappings for all instruments (O(instrument types) instead of O(sessions)) and use that map inside the loop.

driusan:
I took a stab at fixing it with https://github.com/laemtl/Loris/pull/38 .. can you see if it helps?

ridz1208:
@driusan rebased and fixed. Only `_LINST` file changes were added. it's ready for you to review

ridz1208 (approved):
Approving all changes OTHER THEN MY OWN

Issue #6973, opened 2020-09-04
[user_accounts] Examiners with same name removed from table


CamilleBeau:
**Describe the bug**
If a new user is set as an examiner who has the same first and last name as an examiner that already exists, the pre-existing examiner is removed from the `examiners` table, replaced by the new one.

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Create a new user with the exact same first / last name as a pre-existing examiner, and set this user as an examiner
2. look at the examiners table and notice that the first examiner is no longer there, but the newly created examiner is.

ridz1208:
@jesscall don't start working on this just yet, let's discuss it first.

CamilleBeau:
Resolved

PR #7462, opened 2021-06-07
[user_accounts] Use userID to update examiners table

CamilleBeau:
## Brief summary of changes
Currently when updating the examiners table in edit_user, it is assumed that a unique examiner exists based on their full name instead of their userID. This causes 2 problems:
- If an examiner with the same name as another examiner is edited, the older one will be replaced in the table and will no longer be considered an examiner. I.e. only one examiner can exist per full name.
- If a user's name is edited, then a new row will be created in the examiners table and there will be 2 rows for the same user.

This PR fixes this issue by assuming that examiners have a unique userID instead of assuming examiners have a unique full name.

1.Request an account and select "Examiner role"
2. check that a row in the examiners table exists for this newly requested user
3. From user_accounts, edit this user to change the name.
4. Check that the full name is updated in the examiners table for that user, and that only one row exists for that userID
5. Create an account for a user with the same name as an already-existing examiner
6. Make sure that a new row with a different userID is created in the examiners table. Both the older user and the newer user should have rows in the examiner table.

- #6973

### DIFF
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -327,9 +327,9 @@ class Edit_User extends \NDB_Form
             $examinerID = $DB->pselect(
                 "SELECT e.examinerID
              FROM examiners e
-             WHERE e.full_name=:fn",
+             WHERE e.userID=:uid",
                 array(
-                    "fn" => $values['Real_name'],
+                    "uid" => $uid,
                 )
             );
 
@@ -353,8 +353,8 @@ class Edit_User extends \NDB_Form
                 $examinerID = $DB->pselectOne(
                     "SELECT examinerID
                             FROM examiners
-                            WHERE full_name=:fullName",
-                    array('fullName' => $values['Real_name'])
+                            WHERE uid=:uid",
+                    array('uid' => $uid)
                 );
             } elseif (!empty($examinerID)
                 && ((!empty($ex_radiologist)
@@ -370,10 +370,10 @@ class Edit_User extends \NDB_Form
                     'examiners',
                     array(
                         'radiologist' => $ex_radiologist,
-                        'userID'      => $uid,
+                        'full_name'   => $values['Real_name']
                     ),
                     array(
-                        "full_name" => $values['Real_name'],
+                        'userID' => $uid
                     )
                 );
                 $examinerID = $examinerID[0]['examinerID'];


driusan:
@jesscall maybe you can review?

jesscall:
Looks good, just waiting for a cleanup script to be written

jesscall:
@CamilleBeau let me know when this is ready for re-rview

CamilleBeau:
@jesscall Ready for review!

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -323,13 +325,63 @@ class Edit_User extends \NDB_Form
...
+                        "CenterID" => $site,
+                    )
+                );
+            }
+            foreach ($us_curr_projects as $project) {
+                $DB->insert(
+                    'user_project_rel',
+                    array(
+                        "UserID"    => $uid,
+                        "ProjectID" => $project,
```
ridz1208, line 366: @CamilleBeau I'm not a fan of the examiner update section being split into 2 with this section (unrelated to examiners) now being in the middle. Can you instead move lines 303-334 to after 376 ?

CamilleBeau: The problem here is that there is some processing of the $values array that is done in those lines, which is used in the insert for the new user. However, we also need the insert for the new user to happen before examiner update because we need the userID to be generated beforehand. One option would be to hand-pick each of the values from $values that will be used in inserting the user instead of unsetting the values. That way all of the examiner related code could be after the insert. This might be bulky code though.

ridz1208: i c

racostas:
Step 5. "Create an account for a user with the same name as an already-existing examiner" isn't working for me, got 500 error. (I'm crating the account as Admin->add User).

CamilleBeau:
@racostas I couldn't replicate this error. I was able to request an account for a user who had the same name (mine for testing) as 2 other users in the DB. You were testing this on CCNA data right? The login module has an override on CCNA

racostas:
> 
> 
> @racostas I couldn't replicate this error. I was able to request an account for a user who had the same name (mine for testing) as 2 other users in the DB. You were testing this on CCNA data right? The login module has an override on CCNA

Strange. I would suggest somebody else to test. I tested in a fresh version of Loris 23.0.3 using RB not in CCNA but could have some missconfig. Steps from 1 to 4 worked fine nevertheless.

CamilleBeau:
> Strange. I would suggest somebody else to test. I tested in a fresh version of Loris 23.0.3 using RB not in CCNA but could have some missconfig. Steps from 1 to 4 worked fine nevertheless.

I was able to replicate the error Rolando found after all, and I have committed a fix. Ready for re-testing @racostas

racostas (approved):
LGTM now.

tools/single_use/Remove_duplicate_examiners.php
```diff
@@ -0,0 +1,85 @@
+<?php
```
ridz1208, line 1: before SQL

PR #7463, opened 2021-06-07
[API] Add support for project names with space

laemtl:
Remark: Project names are not designed to be unique in the DB. In case of duplicates, the Project factory fails [here](https://github.com/aces/Loris/blob/main/php/libraries/Project.class.inc#L64) (pselectRow used on a query that returns more than one row). (reported in #7464)

### DIFF
--- a/modules/api/php/endpoints/projects.class.inc
+++ b/modules/api/php/endpoints/projects.class.inc
@@ -101,7 +101,7 @@ class Projects extends Endpoint implements \LORIS\Middleware\ETagCalculator
 
         // Delegate to project specific endpoint.
         try {
-            $project_name = $pathparts[1] ?? '';
+            $project_name = urldecode($pathparts[1] ?? '');
             $project      = \NDB_Factory::singleton()
                 ->project($project_name);
         } catch (\NotFound $e) {


modules/api/php/endpoints/projects.class.inc
```diff
@@ -101,7 +101,9 @@ class Projects extends Endpoint implements \LORIS\Middleware\ETagCalculator
 
         // Delegate to project specific endpoint.
         try {
-            $project_name = $pathparts[1] ?? '';
+            $project_name = isset($pathparts[1]) ?
+                urldecode($pathparts[1]) :
+                '';
```
xlecours, line 106:
```suggestion
            $project_name = urldecode($pathparts[1] ?? '');
```

xlecours (approved):
I made a suggestion. Feel free to include it or not.  The null coalesce operator is just a shorter version. 
Other than that, all good.

xlecours:
And regarding the unicity of the project name, I think it is a good example of why the generic `ID` column makes a poor primary key in most situation. 
Thanks for creating https://github.com/aces/Loris/issues/7464

laemtl:
@xlecours  PHP Fatal error:  Uncaught TypeError: urldecode() expects parameter 1 to be string, null given 

> I made a suggestion. Feel free to include it or not. The null coalesce operator is just a shorter version.
> Other than that, all good.

Good suggestion, thanks!

PR #7467, opened 2021-06-08
[API] Project Subprojects

laemtl:
**API 0.0.4 dev** 
Adds the list of project's subprojects to the response of a /projects/$ProjectName GET request.

--------------

2.1 Single project

GET /projects/$ProjectName

Returns a 200 OK response if the project exists, and 404 Not Found if it does not (the same is true of any portion of the API under /projects/$ProjectName.)

The body of the request to /projects/$ProjectName will be an entity of the form:

{
    "Meta" : {
        "Project" : "ProjectName"
    },
    "Visits" : ["V1", "V2", ... ],
    "Instruments" : ["InstrumentName", "InstrumentName2", ...],
    "Candidates" : ["123543", "523234", ...],
    "Subprojects": ["SubprojectName1", "SubProhjectName2"]
}

-------------------------------------
To do:

- [ ] CHANGELOG.md
- [ ] modules/api/docs/LorisRESTAPI_v0.0.4-dev.md 
- [ ] test suite

### DIFF
--- a/modules/api/php/views/project_0_0_4_dev.class.inc
+++ b/modules/api/php/views/project_0_0_4_dev.class.inc
@@ -1,0 +1,46 @@
+<?php declare(strict_types=1);
+/**
+ * PHP Version 7
+ *
+ * @category ApiViews
+ * @package  Loris
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+
+namespace LORIS\api\Views;
+
+/**
+ * This class formats a Project object into arrays following the API
+ * specifications.
+ *
+ * @category ApiViews
+ * @package  Loris
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+
+class Project_0_0_4_Dev extends Project
+{
+    /**
+     * Produce an array representation of this project.
+     *
+     * @return array
+     */
+    public function toArray(): array
+    {
+        $meta = ['Project' => $this->_project->getName()];
+
+        return [
+            'Meta'        => $meta,
+            'Subprojects' => array_values(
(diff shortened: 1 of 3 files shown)


ridz1208:
@laemtl why does this not have a description ?

maltheism (approved):

driusan (changes requested):
seems to be missing the api documentation and schema definition updates.

christinerogers:
@xlecours if the API docs Changes Requested are covered by another PR already, could you please: 
* Link that PR here
* Update your review / dismiss the other change requests so that it's clear what changes still need to be made on this PR?

christinerogers:
Now blocked by #7747

modules/api/php/views/project_0_0_4_dev.class.inc
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $meta = ['Project' => $this->_project->getName()];
+
+        return [
+            'Meta'        => $meta,
+            'Subprojects' => array_values(
+                \Utility::getSubprojectList(
+                    $this->_project->getId()
+                )
+            ),
```
driusan, line 40: I think this key should be removed in order to decouple this from #7747 and unblock it. The rest of the results in this return seem uncontroversial and are all things that already exist in the API. Since it's still `-dev`, it's easier to add something to it if the other moves forward than remove it.

xlecours: The whole point of this PR is to add that key.

driusan: In that case I don't see why it's a separate PR. The whole point of the key seems to be to make the other one accessible.

xlecours: Because at first, the subporjects were at the root of the api (like sites and projects and candidates)

xlecours:
this will be integrated in https://github.com/aces/Loris/pull/7747

PR #7471, opened 2021-06-10
[publication] Lead investigator issue

pierre-p-s:
## Brief summary of changes
As it currently stands, when a publication is proposed (tab: propose a project) the user is asked, amongst other things, for a lead investigator name and email if these values are not already in the database, they get inserted into the collaborator table, if they do already exist they get linked to the new project.

The issue here is when 2 or more users share a name and/or an email address (project address like cbigr@mcin.ca for example) Since the SQL query currently uses and `OR` operator, the query returns the first name that matches or the first email that matches without checking the other value causing in some instances to select the wrong investigator and displaying the incorrect name or email in the menu filter.

Solves issue from CCNA: https://github.com/aces/CCNA/issues/4555

1. Create publication with your name as lead investigator and your email
2. Create another publication with someone else's name as lead investigator but with the same email
3. Check that the 2 publications lead investigators have saved properly

### DIFF
--- a/modules/publication/ajax/FileUpload.php
+++ b/modules/publication/ajax/FileUpload.php
@@ -67,7 +67,6 @@ function uploadPublication() : void
         'FROM publication_collaborator '.
         'WHERE Email = :e',
         array(
-            'n' => $leadInvest,
             'e' => $leadInvestEmail,
         )
     );
--- a/modules/publication/ajax/FileUpload.php
+++ b/modules/publication/ajax/FileUpload.php
@@ -65,7 +65,7 @@ function uploadPublication() : void
     $leadInvID = $db->pselectOne(
         'SELECT PublicationCollaboratorID '.
         'FROM publication_collaborator '.
-        'WHERE Name = :n OR Email = :e',
+        'WHERE Email = :e',
         array(
             'n' => $leadInvest,
             'e' => $leadInvestEmail,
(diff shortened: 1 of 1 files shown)


driusan:
This is failing the tests and can you please update the description to be more descriptive.

ridz1208:
@pierre-p-s tests are now passing but I have no idea why this PR is making that change. can you please add a description with a detail issue and fix ?

modules/publication/ajax/FileUpload.php
```diff
@@ -65,7 +65,7 @@ function uploadPublication() : void
     $leadInvID = $db->pselectOne(
         'SELECT PublicationCollaboratorID '.
         'FROM publication_collaborator '.
-        'WHERE Name = :n OR Email = :e',
+        'WHERE Email = :e',
         array(
             'n' => $leadInvest,
```
ridz1208, line 70: this should also be removed

ridz1208 (dismissed):

ridz1208:
@pierre-p-s I just tested on demo and I cannot reproduce the error.

modules/publication/ajax/FileUpload.php
```diff
@@ -65,7 +65,7 @@ function uploadPublication() : void
...
             'e' => $leadInvestEmail,
```
pierre-p-s, line 71:
```suggestion
        array(
            'e' => $leadInvestEmail,
```

ridz1208:
@pierre-p-s I updated the description to the best of my capacity

Now that I understand the issue a bit more and given the possibility or emails not being unique in this usecase, I think the collaborator table should have a combined primary key of name and email which should low significantly the risk of an overlap and an incorrect display

pierre-p-s:
@ridz1208 I changed the OR to an AND.

ridz1208 (approved):

PR #7473, opened 2021-06-10
[mri_violations] Convert module to React and update UI

### LINKED ISSUE #5714: [Docs] Help text missing for 2 MRI Violations subpages
For 2 subpages in MRI violated scans module: 
* MRI Protocol Violations  : http://demo.loris.ca/mri_violations/mri_protocol_violations
This subpage shows the contents of the `mri_protocol` table and makes it editable in the front end (and therefore highly dangerous) so -- yes, needs help text. 
as well as: 
* MRI Protocol Check Violations : http://demo.loris.ca/mri_violations/mri_protocol_check_violations

discovered in the hunt for #5709 : 

> While MRI Violations does have subpages, it seems like only one help file exists in the repo for this module. So I don't think it applies here unless there used to be files and they've been deleted somehow.

> Hey john, good point - looks like there was never Help for this subpage, only one file was committed in #3266  I'll make a new issue.

_Originally posted by @christinerogers in https://github.com/aces/Loris/issues/5709#issuecomment-556488283_

### LINKED ISSUE #6576: [MRI Violations] Standardize contents of Problem drop-down
**Describe the bug**
Both the 'Not resolved' and 'Resolved' tabs of the MRI violations module have a drop-down in the filter section where you can specify the type of problem you are looking for. I think both drop-downs should be populated similarly, but they are not: the one in the 'Not Resolved' tab is populated by listing all the possible "problems" found in the database, whereas the drop-down in the 'Resolved' tab always has three fixed entries. 

**To Reproduce**
Check the 'Problem' drop-downs contents in the 'Not Resolved' and 'Resolved' tab.

### LINKED ISSUE #6839: [MRI Violations] Improve look and feel
1. The 'MRI Protocol Violations' page should not contain any tab. It should only report the violations of type 'Could not identify scan type' that match the sub-page's filter criteria. The result table for this sub-page should have a new column 'Resolution Status' (non-editable) that indicates whether the violation was resolved or not.

2. The 'MRI Protocol Checks Violations' page should not contain any tab either. It should only report the violations of type 'Protocol Violation' that match the sub-page's filter criteria. The result table for this sub-page should have a new column 'Resolution Status' (non-editable) that indicates whether the violation was resolved or not.

This issue was created after we decided to close PR#6801

### LINKED ISSUE #7470: StaticDataTable Hide defaultColumn does not work
*When a StaticDataTable has the Hide prop set to include `{defaultColumn: true}` the header is hidden, but the numbered column in the table itself is still displayed, causing all the headers and rows to be misaligned.

driusan:
This converts the MRI Violations module to React. While doing so, is simplifies the subpages. They are no longer menu filters, but modal popup windows that display all details of errors for a scan when the link is clicked. All violations for the SeriesUID are shown to provide context. The maintenance should also be simplified by having converted most types of queries to DataProvisioners, which makes it easier to see the query and copy/paste into SQL. This should also result in more consistency with how CenterID/ProjectID matches are handled with the rest of LORIS.

Fixes #7470
Fixes #5714 (the pages no longer exist)
Fixes #6576 (there is only one page with filters now)
Fixes #6839 (the pages no longer exist, so don't have tabs..)


laemtl:
@driusan Failing tests

jsx/Modal.js
```diff
@@ -109,7 +109,7 @@ class Modal extends Component {
       margin: 'auto',
       padding: 0,
       border: '1px solid #888',
-      width: '700px',
+      width: width ? width : '700px',
```
laemtl, line 112: width: width || '700px',

driusan: done

cmadjar:
Discussed at the imaging meeting of June 11th, 2021:
- @nicolasbrossard , @jesscall and @cmadjar will test the new module
- @laemtl, @zaliqarosli will do a quick review of the React code

Thank you :)

cmadjar (changes requested):
@driusan Thank you for reactifying the module! :-)

I did an initial round of testing and found the following:
- The resolution status drop down gets truncated on smaller windows:
Screen Shot 2021-06-14 at 9 47 40 AM
- the data table behaviour when shrinking laterally the window behaves differently in the two different tabs. 
  - For the resolved tab, the table shows the left right navigation bars or column (not sure how they are named, lol)
 Screen Shot 2021-06-14 at 2 22 27 PM
  - For the "Not Resolved" tab, the data table gets out of the border
Screen Shot 2021-06-14 at 2 23 52 PM
- the filter called "Minc File" and column name should probably be renamed "Image File" (or at minimum, MINC should be spelled all upper case)
- for the time run filter, I notice you cannot filter for year/month: if you type in 2019-11, it does not restrict the list to all scans run in Nov 2019 for example. Do you know if this is feasible with the date picker? Not a huge deal but I could see this being handy :-)
- I find confusion the term Violations for 'Could not identify scan type' pop up
- I would tend to rearrange the content of the 'Could not identify scan type' pop up window as follows:
  - Title: use the MINC location instead of the series UID since this is the only information available in the main filter page.  
  - Move the table with the image protocol before the Study protocol table and label it: image protocol instead of Violations
  - add the SeriesUID and Echo Time to the
Screen Shot 2021-06-14 at 11 18 01 AM
- For the Protocol Violation pop up, I would modify the title to contain the image name instead of the seriesUID. It might also be good to have a raw like for the "Could not identify scan type" pop up with the CandID, PSCID, visit, seriesUID etc... information


Let me know if something isn't clear in my fuzzy explanations ;). Thank you!

jsx/StaticDataTable.js
```diff
@@ -91,7 +91,7 @@ class StaticDataTable extends Component {
    * @param {object} prevState - Previous React Component state
    */
   componentDidUpdate(prevProps, prevState) {
-    if (jQuery.fn.DynamicTable) {
+    if (jQuery.fn.DynamicTable && !this.props.NoDynamicTable) {
```
zaliqarosli, line 94: can we add this new prop to StaticDataTable.propTypes and StaticDataTable.defaultProps

modules/mri_violations/jsx/mriViolationsIndex.js
```diff
@@ -0,0 +1,152 @@
...
+          }
+        },
+        (error) => {
+           setIsError(true);
+        }
+      );
+  }, [props.name]);
+
+  if (isError) {
+    return <h3>An error occured while loading the page.</h3>;
```
zaliqarosli, line 97:
```suggestion
    return <h3>An error occurred while loading the page.</h3>;
```

jsx/Modal.js
```diff
@@ -62,7 +62,7 @@ class Modal extends Component {
    * @return {JSX} - React markup for the component
    */
   render() {
-    const {show, children, onSubmit, title} = this.props;
+    const {show, children, onSubmit, title, width} = this.props;
```
zaliqarosli, line 65: add width to proptype definition

driusan:
> The resolution status drop down gets truncated on smaller windows:

Does the non-reactified version work that way? I didn't do anything special with the column sizes, so I'm not sure what I should do about this.

>    the data table behaviour when shrinking laterally the window behaves differently in the two different tabs.
>        For the resolved tab, the table shows the left right navigation bars or column (not sure how they are named, lol)
>        For the "Not Resolved" tab, the data table gets out of the border

The DynamicTable jQuery plugin doesn't work with multiple tables on a page, which is why I had to disable it in the popup windows and it seems that the two tabs also have this problem since it's a single page react app now. I'm not sure what to do about this.
 
 > /   the filter called "Minc File" and column name should probably be renamed "Image File" (or at minimum, MINC should be spelled all upper case)

Okay, that should be a small change.

>    for the time run filter, I notice you cannot filter for year/month: if you type in 2019-11, it does not restrict the list to all scans run in Nov 2019 for example. Do you know if this is feasible with the date picker? Not a huge deal but I could see this being handy :-)

I think the components just use a `<input type="date">` element, I don't think we have any way to do partial dates.

>    I find confusion the term Violations for 'Could not identify scan type' pop up

Okay, I can change it. I think I just started by basing the popup on the other popup when I was writing it and didn't update the title.

>    I would tend to rearrange the content of the 'Could not identify scan type' pop up window as follows:
>        Title: use the MINC location instead of the series UID since this is the only information available in the main filter page.

Okay.

>        Move the table with the image protocol before the Study protocol table and label it: image protocol instead of Violations
>        add the SeriesUID and Echo Time to the

I can add SeriesUID but echo time is already in the protocol table?

>    For the Protocol Violation pop up, I would modify the title to contain the image name instead of the seriesUID. It might also be good to have a raw like for the "Could not identify scan type" pop up with the CandID, PSCID, visit, seriesUID etc... information

I don't think I understand this part.

cmadjar:
> The resolution status drop down gets truncated on smaller windows:

The non reactified version does not do that. I just checked on demo.

> I can add SeriesUID but echo time is already in the protocol table?

ooops. I guess I forgot to remote the echo time from my comment when I realized what you mention here...

> For the Protocol Violation pop up, I would modify the title to contain the image name instead of the seriesUID. It might also be good to have a raw like for the "Could not identify scan type" pop up with the CandID, PSCID, visit, seriesUID etc... information

Sorry, I meant, there could also be a section above the table with the following (like in the other pop up):
Screen Shot 2021-06-14 at 3 54 34 PM

Is that clearer?

ridz1208:
@driusan any movement on this ?

modules/mri_violations/jsx/mriViolationsIndex.js
```diff
@@ -0,0 +1,152 @@
...
+      </TabPane>
+    </Tabs></div>;
+}
+
+/**
+ * Load a data table and render it
+ *
+ * @param {object} props - React Component properties
+ *
+ * @return {ReactDOM}
```
laemtl, line 72: I think JSX is a more appropriate type here

modules/mri_violations/jsx/mriViolationsIndex.js
```diff
@@ -0,0 +1,152 @@
...
+        />;
+}
+
+/**
+ * A helper to generate a column mapper callback which maps fieldOptions
+ * to values.
+ *
+ * @param {object} fieldOptions - the dynamic field options
+ *
+ * @return {callable}
```
laemtl, line 118: Function

modules/mri_violations/jsx/protocolModal.js
```diff
@@ -0,0 +1,273 @@
...
+ *
+ * @return {ReactDOM}
+ */
+function ProtocolModal(props) {
+    switch (props.Type) {
+        case 'protocolviolation':
+          return <ProtocolViolationModal {...props} />;
+        case 'protocolcheck':
+          return <ProtocolCheckViolationModal {...props} />;
+        default: return <div />;
```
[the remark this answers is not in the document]
driusan: nulls always make me nervous but I think react handles it correctly so I'll change it..

modules/mri_violations/jsx/mriViolationsIndex.js
```diff
@@ -0,0 +1,152 @@
...
+        case 'Site':
+            if (fieldOptions.sites) {
+                return fieldOptions.sites[value];
+            }
+        }
+        return value;
+    };
+}
+
+window.addEventListener('load', () => {
```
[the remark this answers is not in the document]
driusan: We don't have a performance bottleneck here.

jsx/Modal.js
driusan: not sure when this was resolved but I just went to fix it and it's already there

driusan:
@cmadjar I think I addressed all your comments and its passing tests now.. can you re-review?

cmadjar (changes requested):
@driusan thank you for rebasing the branch and implemented the changes requested. Having used this module a lot lately, I appreciate that this module is so much more reactive!! :-) 


I noticed the following bugs when going through the testing plan:

- **in the "Resolved" tab:**
  - [ ] the problem column is not a clickable link that would open a pop with details about the violation. Works well in the "Not Resolved" tab.
  - [ ] when filter for SeriesDescription = 'ep2d_diff 25dir av1-updated_FA' (which corresponds to one of the file listed in the table), there is no results shown (data from RB) Looks like filter works for 'ep2d_diff ' but when I add the '2' after the space, the filter do not show anything. Not sure why. Hopefully this will help figure out what the bug is.

- **in the "Could not identify scan type" pop up:**
  - [ ] should add the following fields: `PhaseEncodingDirection`, `ImageType` and `EchoNumber` (those columns were added in 24.1, see #8150) - fields to be added to scan parameter list and list of protocols' table. 
  - [x] in the list of protocol table, slice thickness range is wrongly labelled `slice_thickness_max`

- **in the "Protocol Violation" pop up:**
  - all good :-)

- **Permissions:**
  - [ ] when user only belong to one site, the user can still see the full list of sites in the Site drop down menu filter (in Resolved and Not Resolved tabs)
  - [ ] when user only belong to one project, the user can still see the full list of projects in the Project drop down menu filter (in Resolved and Not Resolved tabs)

- **Help text**
  - [x] `MincFile` string in help text needs to be replaced by `Image File`

- **Dashboard**
  - [ ] in the dashboard, the number of violations listed do not match the number of rows in the MRI violation module (if user belongs to site Ottawa and project Pumpernickle, the dashboard displays 168 violations to resolve but there are 169 rows displayed in the violation module)

- **violation resolving feature**: I find it not super user friendly, maybe because I am used to the save button... I am starting to wonder whether we need to tabs for this module. Couldn't it be only one page with two columns: one with the resolution status displayed and one where the user can change the resolution status. Does not need to be on that PR but just wanted to see your thoughts on that workflow. Would also allow users to change the resolution status. Let's say a resolution status is "Emailed site", eventually this would need to be changed and there is no way to do that on the frontend. I can create an issue for that but I am curious about your thoughts on this first.

driusan:
> when filter for SeriesDescription = 'ep2d_diff 25dir av1-updated_FA' (which corresponds to one of the file listed in the table), there is no results shown (data from RB) Looks like filter works for 'ep2d_diff ' but when I add the '2' after the space, the filter do not show anything. Not sure why. Hopefully this will help figure out what the bug is.

I traced through this and the Series Description is actually `ep2d_diff  25dir av1-updated_FA` (note: 2 spaces between ep2d_diff and 25dir)

> Permissions:
>
> when user only belong to one site, the user can still see the full list of sites in the Site drop down menu filter (in Resolved and Not Resolved tabs)
>  when user only belong to one project, the user can still see the full list of projects in the Project drop down menu filter (in Resolved and Not Resolved tabs)

Isn't this usually the case in the rest of LORIS too?

driusan:
Created issue #8385 for the whitespace thing since it's related to the data table and not the MRI violations module.

driusan:
>Dashboard
>
> in the dashboard, the number of violations listed do not match the number of rows in the MRI violation module (if user belongs to site Ottawa and project Pumpernickle, the dashboard displays 168 violations to resolve but there are 169 rows displayed in the violation module)

One was only filtering for user center match and the other only for user project match. They should both do both now (and therefore match.. )

driusan:
@cmadjar Links should be clickable on resolved tab now. I think this is in your court again.

cmadjar:
@driusan see below for the weird display of the main table when the width of the browser gets smaller. Noticed that out of luck.
Screen Shot 2023-02-20 at 2 01 19 PM

Another little thing because I am a little neurotic... In the top table for the could not identify scan type, every header has the first letter of the words capitalized but in the Default MRI protocol group table, xspace, yspace, zspace, xstep, ystep, zstep, time range headers do not have the first letter of their words capitalized. Any chance it could be harmonized?

Thanks!

cmadjar:
@driusan I just realized an important feature is missing from the MRI violation module that used to be there. Sorry for the multiple messages... The image file link used to be clickable and that opened the image in brainbrowser. Any reason why this feature does not exist anymore?

driusan:
Looks like the link disappeared because the column was renamed from Minc File to Image File and the formatter wasn't updated.. should be fixed now. I also fixed the header capitalization.

I'm not sure what I can do about the overflowing table. That's usually handled by DynamicTable adding the overflow bars, but DynamicTable doesn't work if there's more than 1 table on the page. It had to be removed since there are multiple tables (one on each tab)

cmadjar (approved):
LGTM 👍 Thank you!!

PR #7479, opened 2021-06-14
[API] Visit patch request to start next stage

laemtl:


modules/api/php/endpoints/candidate/visit/visit.class.inc
```diff
@@ -171,10 +174,120 @@ class Visit extends Endpoint implements \LORIS\Middleware\ETagCalculator
...
+        }
+
+        if (!$nextStageDate) {
+            return new \LORIS\Http\Response\JSON\BadRequest(
+                'There is no stage date specified'
+            );
+        }
+
+        $dateSegments = explode("-", $nextStageDate);
+        if (!checkdate((int)$dateSegments[1], (int)$dateSegments[2], (int)$dateSegments[0])) {
```
xlecours: Using `$date = new \DateTime($nextStageDate);` would take care of many date formats as well as validate. It throws an \Exception so you would need to put it in a try..catch.

modules/api/php/endpoints/candidate/visit/visit.class.inc
```diff
@@ -171,10 +174,120 @@ class Visit extends Endpoint implements \LORIS\Middleware\ETagCalculator
...
+            return new \LORIS\Http\Response\JSON\BadRequest(
+                'The date specified in the POST data is invalid. date must be of the form YYYY-MM-DD'
+            );
+        }
+
+        $sessionID = \NDB_Factory::singleton()->database()->pselectOne(
```
xlecours, line 226: To avoid this static call, the `_handlePUT` code is using 
```
        $sessionid = array_search( 
               $meta['Visit'], 
               $this->_candidate->getListOfVisitLabels()
        );
```

But, in this case, the timepoint must exists so `$this->_visit` must not be null. You should still check if `$meta['Visit'] == $this->_visit->getVisitLabel()`

xlecours (changes requested):
I think we should find a solution that will allow to go from "Not started" to"In Progress" and also from "In Progress" to "Pass" or any other valid status. The get request returns a "Stages" with a property fro each stages. 
```
"Stages": {
    "Screening": {
      "Date": "string",
      "Status": "Pass"
    }
  }
```
I think that a PATCH request with a similar object in the body could handle more types of transitions. Also, the frontend ask if scans were acquired, maybe we should add that.

modules/api/php/endpoints/candidate/visit/visit.class.inc
laemtl: Thanks for the pointer. I refactored the validation section.

modules/api/php/endpoints/candidate/visit/visit.class.inc
laemtl: I rewrote this section before your comment in a slightly different way. Let me know if you agree with it.

laemtl:
@xlecours My intuition is that implementing a solution to switch from _In Progress_ to _Pass_ implies some checks that I'm not familiar with. It's probably the same for _Withdrawal_ and maybe _Failure_. This PR covers the functionality I need for a particular project, I may not have time to address your suggestion before a while.

driusan:
I agree with Xavier that it would be better to infer it from the same structure based on what stages are set in

```
    "Stages" : {
              "Screening" :  {
                  "Date" : "YYYY-MM-DD",
                  "Status" : "Pass|Failure|Withdrawal|In Progress"
              },
              "Visit" : {
                  "Date" : "YYYY-MM-DD",
                  "Status" : "Pass|Failure|Withdrawal|In Progress"
              },
              "Approval" : {
                  "Date" : "YYYY-MM-DD",
                  "Status" : "Pass|Failure|Withdrawal|In Progress"
              }
  ```

For the sake of starting a next stage, handling a `PATCH` request that sets the `Status` of `Visit` to `In Progress` is all that's really needed to handle starting a visit stage. The rest of the transitions can either return a "Not Implemented" or "Bad Request" result as appropriate.

**Comment by laemtl** [comment]
@driusan Leaving some cases Not Implemented is a really good idea to address Xavier's suggestion in a limited time frame. I will do that shortly.

**Comment by xlecours** [comment]
I have a bad feeling about using inheritance to solve the versioning problem. I understand that both endpoints are sharing code but, when we will move 0.0.4 from dev to prod, everything that is not in the subclass will need to be merge anyway. 
@laemtl @driusan ?

File: modules/api/php/endpoints/candidate/visit/visit.class.inc [hunk 3]
```diff
@@ -95,23 +108,19 @@ class Visit extends Endpoint implements \LORIS\Middleware\ETagCalculator
...
+            $method = $request->getMethod();
 
-            case 'PUT':
-                return $this->_handlePUT($request);
-
-            case 'OPTIONS':
+            if (in_array($method, $this->allowedMethods())) {
+                $handle = 'handle'.$method;
+                return $this->$handle($request);
+            } else if ($method === 'OPTIONS') {
```
xlecours, line 120: There is not `else` in the API. (with 1 exception that I never took the time to fix)

modules/api/php/endpoints/candidate/visit/visit_0_0_4_dev.class.inc
```diff
@@ -0,0 +1,155 @@
...
+        if (!$nextStageDate) {
+            return new \LORIS\Http\Response\JSON\BadRequest(
+                'There is no stage date specified'
+            );
+        }
+
+        $date_format = 'Y-m-d';
+        $date        = \DateTime::createFromFormat($date_format, $nextStageDate);
+
+        if (!($date && $date->format($date_format) === $nextStageDate)) {
```
xlecours, line 96: Is there a case when `$date->format($date_format) === $nextStageDate` would not be true?

modules/api/php/endpoints/candidate/visit/visit_0_0_4_dev.class.inc
```diff
@@ -0,0 +1,155 @@
...
+            return new \LORIS\Http\Response\JSON\BadRequest(
+                'The date specified is invalid. Date must be of the form YYYY-MM-DD'
+            );
+        }
+
+        if ($nextStageDate > date($date_format)) {
+            return new \LORIS\Http\Response\JSON\BadRequest(
+                'The date specified must be not later than today'
```
xlecours, line 104: Is this a new requirement?

modules/api/php/endpoints/candidate/visit/visit_0_0_4_dev.class.inc
```diff
@@ -0,0 +1,155 @@
...
+            );
+        }
+
+        $newStage = $this->_visit->getNextStage();
+        if (!$newStage) {
+            return new \LORIS\Http\Response\JSON\Conflict(
+                "Next stage is null"
+            );
+        }
+
```
xlecours, line 120: The only validations missing are useScanDone and allowPrenatalTimepoints config dependent ones. see: https://github.com/aces/Loris/blob/main/modules/next_stage/php/next_stage.class.inc#L256-L273

@driusan All this, and the code in the next_stage module should all be migrate to the timepoint class IMO. It would be convinient to call `$visit->startNextState(stage='Visit', startdate='2020-04-23', scandone=false);`

xlecours: It is not in the scope of this PR, i know :)

xlecours (changes requested):
This should also modify `modules/api/static/schema.yml`. I can push a contribution if you would like me to.

modules/api/php/endpoints/candidate/visit/visit_0_0_4_dev.class.inc
laemtl: Can we start a stage in the future?

modules/api/php/endpoints/candidate/visit/visit_0_0_4_dev.class.inc
laemtl: > This should also modify `modules/api/static/schema.yml`. I can push a contribution if you would like me to.

I am not modifying v0.0.3 so do we need to change anything?

laemtl:
> This should also modify `modules/api/static/schema.yml`. I can push a contribution if you would like me to.

I am not modifying v0.0.3 so do we need to change anything?

xlecours:
> > This should also modify `modules/api/static/schema.yml`. I can push a contribution if you would like me to.
> 
> I am not modifying v0.0.3 so do we need to change anything?

You are right. Let's add a new file then.

laemtl:
@driusan @xlecours Comments addressed.

modules/api/docs/LorisRESTAPI.md
```diff
@@ -416,7 +416,7 @@ The JSON object is of the form:
 }
 ```
 
-A PUT of the same format but with only the Meta fields will create the VisitLabel
+A PUT with only the Meta fields will create the VisitLabel
```
xlecours, line 419:
```suggestion
A PUT request with only the Meta fields will create the VisitLabel
```

modules/api/docs/LorisRESTAPI_v0.0.4-dev.md
```diff
@@ -416,10 +416,10 @@ The JSON object is of the form:
 }
 ```
 
-A PUT of the same format but with only the Meta fields will create the VisitLabel
+A PUT request with only the Meta fields will create the VisitLabel
 for this candidate, in an unstarted stage if the Visit label provided is valid.
 
-PATCH is not supported for Visit Labels.
+A PATCH request with the Meta fields and a date NextStageDate will start the next stage, if the stage is 'Not Started' and the CandID and Visit label provided are valid.
```
xlecours, line 422:
```suggestion
A PATCH request with the Meta fields and a Stages property that contains a Date and a Status == 'In Progress' will start the next stage, if the current stage is 'Not Started' and the CandID and Visit label provided are valid.
```

modules/api/php/endpoints/candidate/candidate.class.inc
```diff
@@ -112,10 +112,18 @@ class Candidate extends Endpoint implements \LORIS\Middleware\ETagCalculator
...
+            $handler = new Visit\Visit(
+                $this->_candidate,
+                $visit
+            );
+        } else {
+            $handler = new Visit\Visit_0_0_4_Dev(
+                $this->_candidate,
+                $visit
+            );
+        }
```
xlecours, line 126:
```suggestion
        $subendpoint = __NAMESPACE__ . '\\' . 'Visit\Visit';

        if ($version == 'v0.0.4-dev') {
            $subendpoint = __NAMESPACE__ . '\\' .'Visit\Visit_0_0_4_Dev';
        }

        $handler = new $subendpoint(
            $this->_candidate,
            $visit
        );
```

xlecours (changes requested):
A few suggestions.
I'll send a PR for the schema.yml file

laemtl:
@driusan @xlecours Ready for re-review.

modules/api/docs/LorisRESTAPI_v0.0.4-dev.md
```diff
@@ -416,10 +416,37 @@ The JSON object is of the form:
...
+        "Approval" : {
+            "Date" : "YYYY-MM-DD",
+            "Status" : "Pass|Failure|Withdrawal|In Progress"
+        }
+    }
+}
+```
+will update the stages date and status. 
+
+At this moment, we only support requests for the Visit stage with the In progress status,
```
driusan, line 448: It's not clear to me what this means for the other fields. Should they be left out of the request? Included with garbage data?

laemtl: I removed the other fields.

modules/api/static/schema-v0.0.4-dev.yml
```diff
@@ -0,0 +1,2493 @@
...
+          enum:
+            - Pass
+            - Failure
+            - Withdrawal
+            - In Progress
+    VisitPatchFields:
+      type: object
+      allOf:
+        - $ref: '#/components/schemas/VisitMetaFields'
+        - $ref: '#/components/schemas/VisitStages'
```
xlecours, line 1925:
```suggestion
    VisitPatchFields:
      type: object
      allOf:
        - $ref: '#/components/schemas/VisitMetaFields'
        - type: object
          properties:
            Stages:
              $ref: '#/components/schemas/InstrumentVisit'
```

xlecours (changes requested):

modules/api/php/endpoints/candidate/visit/visit_0_0_4_dev.class.inc
```diff
@@ -0,0 +1,519 @@
...
+                    $battery->createBattery(
+                        $request->getAttribute("loris"),
+                        $this->_visit->getSubprojectID(),
+                        $newStage,
+                        $this->_visit->getVisitLabel(),
+                        $this->_visit->getCenterID(),
+                        $firstVisit
+                    );
+
+                    return new \LORIS\Http\Response\JSON\OK();
```
xlecours, line 502: This should return 204 No content instead of an empty array

modules/api/static/schema-v0.0.4-dev.yml
```diff
@@ -0,0 +1,2496 @@
...
+          explode: false
+          schema:
+            type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/VisitPatchFields'
+      responses:
+        '200':
```
xlecours, line 364: This should be 204 No Content

xlecours (approved):
I have tested it. It works all right.  It's just that an empty array as a response is not ideal. I will approve this in it's current state but it should return and empty response with a 204 status code.

PR #7481, opened 2021-06-15
[DQT] performance modifications to data table for DQT only.

maltheism:
## Brief summary of changes

@ridz1208 encountered a few performance issues with the DQT "generating the table" after the query is run while testing CBIGR's dataset.

We found a lot of issues with how react would re-render the table multiple times after the table was rendered but the user clicked other options in the DQT, how some of the table "filtering" code was causing extra time spent when the DQT doesn't need filtering like other modules of LORIS, some of the regex was causing a lot of time wasted for generating the table and a few loops in the foregoing mentioned types of code was removed while making the performance better. It feels like a good idea to have a separate table of cleaned up code that's specifically for the DQT module and that's stripped of things not needed for the DQT while making the performance better but also with the goal of making it easier to maintain for the DQT with performance in mind.

1. Setup DQT with a complicated query.
2. run the query.
3. click options around the DQT (performance shouldn't be effected).
4. the table after running the query should be rendered fast.


driusan (dismissed):
Why is this adding another new table class? What is the difference between it and the old one?

driusan:
Yes, it should be modified by a prop, not copy/pasted into a different file.

maltheism:
Hi @driusan, I added prop for using the existing StaticDataTable.

maltheism:
@driusan it's ready for you

driusan:
@xlecours can you re-test? I looked it over with `?w=1` and it looked fine, but it should probably be re-tested. If you confirm it's still working I'll merge.

xlecours (approved):
The DQT still works. The table rendering is smooth. 
I did not look at the code neither checks if new bugs were nintroduced.

PR #7484, opened 2021-06-16
[Instrument] Add support for jsonData instrument in clearInstrument function

### LINKED ISSUE #7465: Clear Instrument fails with jsonData=true
When data is saved in an instrument with jsonData=true (ie it's saved in the flag table instead of an instrument table) `NDB_BVL_Instrument->clearInstrument()` throws an exception. ClearInstrument attempts to update the (non-existant) table.

In particular, this happens with the API handling a `PUT` request (since it tries to clear the old data before saving the new data).

The function should have an if statement to update the appropriate flag or instrument table based on the jsonData property.

ridz1208:
## Brief summary of changes

`clearInstrument` Function gives an error when the instrument has jsonData enabled. this fix adds support for clearInstrument by clear the data from the proper location

PS, make sure to ignore white spaces changes as most changes here are just indentation

1. use the API to save an instrument

* Resolves #7465

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2300,29 +2300,33 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         $config   = NDB_Config::singleton();
         $db       = Database::singleton();
         $dbconfig = $config->getSetting('database');
-        $columns  = $db->pselect(
-            "SELECT COLUMN_NAME FROM information_schema.columns
+        if ($this->jsonData) {
+            $db->update('flag', ['Data'=>null], ['CommentID' => $this->commentID]);
+        } else {
+            $columns = $db->pselect(
+                "SELECT COLUMN_NAME FROM information_schema.columns
             WHERE TABLE_NAME=:table AND TABLE_SCHEMA=:db",
-            array(
-                'table' => $this->table,
-                'db'    => $dbconfig['database'],
-            )
-        );
+                array(
+                    'table' => $this->table,
+                    'db'    => $dbconfig['database'],
+                )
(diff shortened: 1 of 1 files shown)


GabrielPelletier:
Tested this on MCNRR and it all looks good.

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2300,29 +2300,33 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         $config   = NDB_Config::singleton();
         $db       = Database::singleton();
         $dbconfig = $config->getSetting('database');
-        $columns  = $db->pselect(
-            "SELECT COLUMN_NAME FROM information_schema.columns
+        if ($this->jsonData) {
+            $db->update('flag', ['Data'=>null], ['CommentID' => $this->commentID]);
```
driusan, line 2304: Shouldn't this update be unconditional? The `Data` column in the flag table exists regardless of whether jsonData is true or not, and the update when saving data in `_save` isn't conditional.

ridz1208: You are correct. Ma bad

ridz1208:
@driusan changes done

GabrielPelletier:
Re-tested after the changes with MCNRR and it works alright.

driusan (approved):
I can settle for the downgrade from "good" to "alright" if it means it's ready to merge.

ridz1208:
@driusan I think we should merge before @GabrielPelletier takes it from "alright" to "its not bad"

GabrielPelletier:
I like a bit of variation; it makes for richer github Conversations, literature-wise

PR #7497, opened 2021-06-29
[Filter] Fix Menu Filter logic to recognise sex for exact match

### LINKED ISSUE #7492: Loris Menu Filter on Candidate Sex not working
**Describe the bug**
When adding a menu filter for the candidate sex, filtering on `male` is returning `female` as the SQL query is using `candidate.sex LIKE '%male%` instead of `candidate.sex = 'male'`. Seems the substring [if condition](https://github.com/aces/Loris/blob/main/php/libraries/NDB_Menu_Filter.class.inc#L412) was not properly updated to use the `sex` string length.

**To Reproduce**
Unsure of with core modules still use POST to preform menu filters. Most see to be reactified. 

Expected only `male` candidates to return. 

- ALL

**Server Environment (if known):**\\
- LORIS Version: 23.0.3\\

**Additional context**

ridz1208:
## Brief summary of changes
This PR fixes a missing change from #3897 where `gender` was changed to `sex` but the substring query was not altered accordingly.

Fixes #7492

### DIFF
--- a/php/libraries/NDB_Menu_Filter.class.inc
+++ b/php/libraries/NDB_Menu_Filter.class.inc
@@ -72,7 +72,7 @@ class NDB_Menu_Filter extends NDB_Menu
      * @var    array
      * @access private
      */
-    var $validFilters = array();
+    var $validFilters = [];
 
     /**
      * Items selected from the GUI to filter, where the underlying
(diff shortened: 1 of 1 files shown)


driusan:
why was this closed if it's a bug?

ridz1208:
@driusan I think I made another one. I [was](url) too lazy to fix the unnecessary changes
https://github.com/aces/Loris/pull/7498

PR #7502, opened 2021-06-29
[API] Subproject endpoint

laemtl:


maltheism (approved):
Great work!

modules/api/php/views/subproject.class.inc
```diff
@@ -0,0 +1,91 @@
...
+        $visits = [];
+
+        $subprojectID = array_search(
+            $this->_subproject,
+            \Utility::getSubprojectList()
+        );
+
+        $config =& \NDB_Config::singleton();
+        $visitLabelSettings = $config->getSetting('visitLabel');
+        foreach (\Utility::associativeToNumericArray($visitLabelSettings)
```
driusan, line 74: @ridz1208 is this the best way to get visits for a subproject?

driusan (changes requested):
This looks like it's missing the update to the v4 api documentation and the api schema file.

modules/api/php/endpoints/subprojects.class.inc
```diff
@@ -0,0 +1,149 @@
...
+     */
+    private function _handleGET() : ResponseInterface
+    {
+        if (!isset($this->_cache)) {
+            // The subprojects list should be related to the user in the request
+            // attributes.
+            $DB           = \NDB_Factory::singleton()->database();
+            $subprojects  = $DB->pselect("SELECT * FROM subproject", []);
+            $this->_cache = (new \LORIS\api\Views\Subprojects($subprojects))
+                ->toArray();
```
xlecours, line 130: This should use a dataprovisioner instead of querying the database directly. The dataprovisionner should return a Traversable of Subproject objects that would be passe to the view class. 
In the view class, getting the the visitlabels fro each subproject should be something like `$subproject->getVisitLabels();`

Using the dataprovisioner will allow us to add  UserProjectMatch to filter out the subprojects not available to the user.
Creating a Subproject "model" class will create a single source of truth (to answer Dave's comment about using the Utility class to to get visits for a subproject)

xlecours (changes requested):

PR #7503, opened 2021-07-02
Add Alias in parameter type table and map BIDS Alias to MINC headers present in parameter_type

cmadjar:
## Brief summary of changes

A new column is added to the parameter_type to add an Alias to the parameters. For imaging related fields that comes from the MINC header of images, this Alias will be used to store the BIDS term mapping for a given parameter. This will be needed for the dcm2bids LORIS insertion pipeline in order to map the MINC terms with the BIDS terms.

It should be noted that currently, some fields are present multiple times in the parameter_type table with different parameter name (this can be observed in the raisin bread dataset). An example would be the echo time. There are 3 different parameter types available for the echo time: `echo_time` (ID 130), `acquisition:echo_time` (ID 113) and `dicom_0x0018:el_0x0081` (ID 209). For the BIDS terms, only the `echo_time` has the BIDS term Alias `EchoTime` to avoid mapping `EchoTime` randomly to 3 different parameter type ID.

We might want to think whether we should not clean those duplicata at some point for the MINC files. To be discussed at the imaging call on Thursday July 8th.

**All descriptions added to the parameter fields in this PR came from [InterLex](https://scicrunch.org/scicrunch/interlex/dashboard).**

#### Testing instructions 

1. Run the patch and check that the Alias and Description have been updated for the fields present in this PR
2. Test that 0000-00-000-schema.sql is running properly
3. Reload the raisin bread dataset and check that the Alias and Description have been updated for the fields present in this PR. 

This is not linked to a particular issue but was discussed during a roadmap meeting. This feature will become handy for the dcm2bids LORIS insertion script.


cmadjar:
@driusan do you know what the deal is with this PR? Not sure why it fails or how to restart the jobs :S

cmadjar:
@driusan I am able to load the schema and the raisinbread data... I rebased the branch in case it was caused by something else? Let's see if the tests will complete this time. Fingers crossed!!

driusan:
It doesn't look like it's based properly? (118 files changed?)

cmadjar:
ooops. Just rebased it properly, I think.

cmadjar:
@driusan urgh, it does not fail on Mac but I decided to test it on my sandbox VM, just in case and found the following error:
```
BLOB/TEXT column 'SourceFrom' used in key specification without a key length
```
Is there a reason why `SourceFrom` was set to `TEXT` and not `VARCHAR`? Especially since this is referring to other table names that will unlikely be from than 64? (https://www.tutorialspoint.com/what-is-the-maximum-length-of-a-table-name-in-mysql#:~:text=The%20maximum%20length%20of%20a%20table%20name%20is%2064,according%20to%20MySQl%20version%208.0.)

I changed the column type to VARCHAR instead of TEXT so it can be used in conjunction to the `Name` column as a UNIQUE KEY.

cmadjar:
@driusan passing the tests finally!!

driusan:
It looks like it's `text` in the oldest version of the file I can find, so I'm not sure why.

https://github.com/aces/Loris/commit/cb4a17d4127879193af4ac2acc33949d61f4cfc7#diff-cc0a6e8ceb09212c1e1af4d1837e26bb6a6c559604c416ef7ee4e56f07c5baf4R902

cmadjar:
@driusan @ridz1208 do you think converting it to varchar would break anything? I don't see why it would honestly. 

@PapillonMcGill any chance you want to review this PR?

PapillonMcGill:
@cmadjar The only problem that can arise from converting `text`to `varchar(255)` is if you already have data that is more than 255 chars long. In that case, it will be shorten to the first 255 chars.

You can find the maximum size already recorded in an instance to LORIS with:
`SELECT MAX(CHAR_LENGTH(SourceFrom)) FROM parameter_type;
`

PapillonMcGill (approved):
Approved for the structure part. 
I haven't checked the actual data that is put in the table.

As for replacing the `text` with `varchar(255)`, it depend on the length of text required.

cmadjar:
> @cmadjar The only problem that can arise from converting `text`to `varchar(255)` is if you already have data that is more than 255 chars long. In that case, it will be shorten to the first 255 chars.
> 
> You can find the maximum size already recorded in an instance to LORIS with:
> `SELECT MAX(CHAR_LENGTH(SourceFrom)) FROM parameter_type;`

@PapillonMcGill since technically that column refers to table names, I believe converting it to `varchar(255)` should be fine since table names are usually shorter. Unless there is another use for that column I am not aware of @driusan @ridz1208?

PapillonMcGill:
In that case, you may even use `varchar(64)` as identifier are limited to 64 chars.

PR #7507, opened 2021-07-14
[Configuration] Fix saving of values when no previous value present

driusan:
When there is a value in ConfigSettings but not Config, the value does
not get saved to the database but silently discarded. This is caused by
the save logic assuming that the ID always exists to update. However,
other logic in the configuration saving deletes from the Config table
if the value is "". This means that once a config option is set to the
empty string it can never be saved again. (This can also happen for
new ConfigSettings.)

Since the name used for the text field in the HTML is the Config table ID
(which doesn't exist) and not the ConfigSetting ID to support ConfigSetting
options that allow multiple entries, the name of the ID is "0" in the form
and can not be converted to a ConfigSetting to change the statement from an
update to an insertOnDuplicateUpdate. AllowMultiple settings, on the other
hand use "add-ConfigSettingID" or "remove-ConfigSettingID" as their name.

This modifies the template to use add-ConfigSettingID when the ConfigID is
"0" in order to have it inserted if it does not exist, regardless of the
AllowMultiple setting for the ConfigSetting.

### DIFF
--- a/modules/configuration/templates/form_configuration.tpl
+++ b/modules/configuration/templates/form_configuration.tpl
@@ -74,20 +74,29 @@
     <div class="config-form-group" id="{$node['ID']}">
     {foreach from=$node['Value'] key=k item=v}
         {if $node['AllowMultiple'] == 1}<div class="input-group entry">{/if}
-        {if $node['DataType'] eq 'boolean'}
-            {call createRadio k=$k v=$v d=$node['Disabled']}
-        {elseif $node['DataType'] eq 'instrument'}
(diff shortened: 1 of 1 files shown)


driusan:
The actual problem is here: https://github.com/aces/Loris/blob/main/modules/configuration/ajax/process.php#L36-L68

But since the name of the HTML element is the Config ID, without changing the smarty template all the backend knows is that it's trying to update ID "0" and there's no way to map that to a ConfigID (which doesn't exist.)

jstirling91 (approved):

PR #7509, opened 2021-07-20
Add PSR3 Logger

driusan:
This adds a PSR3 logger implementation to LORIS. This  implements a single
logger type, ErrorLogLogger, which logs to the web server's standard error_log
based on the current log level verbosity configuration in the LORIS config
module.

Three types of logs are implemented in this PR, the verbosity: 1. Database logging
2. HTTP Request logging and 3. Exception logging

### DIFF
--- a/php/libraries/Module.class.inc
+++ b/php/libraries/Module.class.inc
@@ -366,7 +366,12 @@ abstract class Module extends \LORIS\Router\PrefixRouter
         $user  = $request->getAttribute("user") ?? new \LORIS\AnonymousUser();
         $loris = $request->getAttribute("loris");
 
-        $page = $this->loadPage($loris, $pagename);
+        try {
+            $page = $this->loadPage($loris, $pagename);
+        } catch(\NotFound $e) {
+            $this->logger->debug("Page $pagename not found");
+            return $resp->withStatus(404);
+        }
         // FIXME: Hack required for breadcrumbs. This should be removed,
         // but some tests depend on it.
         if ($this->getName() === $pagename) {
--- a/php/libraries/Module.class.inc
+++ b/php/libraries/Module.class.inc
@@ -169,7 +169,7 @@ abstract class Module extends \LORIS\Router\PrefixRouter
      */
     public function __construct(string $name, string $moduledir)
     {
-        $config   = \NDB_Config::singleton();
+        $config   = \NDB_Factory::singleton()->config();
         $loglevel = $config->getLogSettings()->getRequestLogLevel();
 
         $this->logger = new \LORIS\Log\ErrorLogLogger($loglevel);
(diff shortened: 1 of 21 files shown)


driusan:
@jstirling91

driusan:
Blocked by #7507

php/libraries/Module.class.inc
```diff
@@ -351,78 +365,31 @@ abstract class Module extends \LORIS\Router\PrefixRouter
...
                     new \LORIS\Http\Error(
                         $request,
-                        404,
-                        "File not found: " . $request->getURI()->__toString()
+                        403,
+                        "You do not have access to this page."
                     )
                 )
             );
-            /* The order of these catch statements matter and should go from
```
driusan, line 392: This was moved to the ExceptionHandlingMiddleware (where it should have been in the first place) so that the exception log handler is used.

driusan:
@kongtiaowang can you help debug this? The first time PHP 8 passed but not 7.4. I restarted GitHub actions a couple times and now it's failing with PHP8 too. I downgraded to 7.4 locally and the feature is working for me when I try it in the browser. From what I can tell the query selector `.col-sm-12 > .row .btn` isn't matching anything in the unit test, but this PR doesn't touch anything that should affect that.

kongtiaowang:
https://github.com/aces/Loris/pull/7510
This PR could fix the issue. The reloading page of "open profile" takes more time than before, but it is acceptable (1s).

driusan:
I'm not sure what would have changed in this PR that affects the speed but I'll give it a shot.

driusan:
Thanks @kongtiaowang, that seemed to fix it.

@jstirling91 this should be ready for you now that it's passing the tests.

kongtiaowang:
The last fix is also in the https://github.com/aces/Loris/pull/7510.

driusan:
@kongtiaowang It still seems to be failing because " 1) TestInstrumentTestIntegrationTest::testSelectOptionElement
Facebook\WebDriver\Exception\ElementClickInterceptedException: Element <select name="consent"> is not clickable at point (1158,625) because another element <div class="panel-body"> obscures it"

test/test_instrument/test_instrumentTest.php
```diff
@@ -203,7 +206,9 @@ function testSelectOptionElement()
 
         // select 'No' option and check it.
         $this->_landing();
-        $select  = $this->safeFindElement(WebDriverBy::Name("consent"));
+        $select  = $this->safeFindElement(
+            WebDriverBy::cssSelector("select[name='consent']")
+        );
```
kongtiaowang, line 211: please try add "sleep(1);" into line 213

driusan:
@jstirling91 This time it's passing for real.

jstirling91 (approved):
Looks good to me. I'm not able to test this though as my VM is still on Loris v21.0.0 since IBIS prod hasn't been upgraded to LORIS v23.0.0

PR #7512, opened 2021-07-26
[DQT] Check for scan types in MRI Importer script

driusan:
Add a check to verify that GetScanTypes returned some
valid, configured scan types with data in CouchDB import
script.

If the import script is run with no scan types, it currently
only imports the session QC status. The fact that the script
failed to import the important data isn't caught until after
a user tries to use the DQT.

This adds a check to error out if getScanTypes does not return
any scan types at import time.


driusan:
@cmadjar Any thoughts on this? It seems to be a pretty common problem that would be better to catch sooner rather than debug after the fact.

PR #7517, opened 2021-07-29
physiological_task_event columns types according to BIDS spec

laemtl:
According to the [BIDS specs](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/05-task-events.html), the


cmadjar (approved):
LGTM. I am not sure I understand why a PHP script would be required for project updates? Wouldn't the patch be enough to convert the columns? Unless you are talking about projects for which the data did not get inserted because of the type error?

laemtl:
@cmadjar The patch may fail if the table is not empty and the data conversion is not possible (not tested). 

> LGTM. I am not sure I understand why a PHP script would be required for project updates? Wouldn't the patch be enough to convert the columns? Unless you are talking about projects for which the data did not get inserted because of the type error?

@cmadjar After a quick check, existing data should be automatically converted without the need of any script.

cmadjar:
@laemtl thank you for the follow up!

@driusan ready for you :)

PR #7519, opened 2021-07-30
Convert physiological_parameter_file Value column to text

laemtl:
Fix the following error:

```
Executing query:
	INSERT INTO physiological_parameter_file (PhysiologicalFileID, ParameterTypeID, Value) VALUES (%s,%s,%s)
With arguments:
	(47, 1135, 'participants were asked to press one of two keys with either their left or right index finger (assignment counter-balanced across participants). Their key-press was based on how symmetric they regarded each image: pressing one or the other key depending whether they thought the image was ‘more’ or ‘less symmetric’ than average.')

MySQLdb._exceptions.DataError: (1406, "Data too long for column 'Value' at row 1")

```


cmadjar (approved):
LGTM. Thank you @laemtl for the fix!

PR #7535, opened 2021-08-16
[Conflict Resolver] Displaying user projects and sites for filters

laemtl:
This PR:
1. Migrates the filter helpers from \DataFrameworkMenu to \NDB_Menu_Filter and add an $allsitesPermission parameter in getSiteOptions, since the PHP classes do not extend DataFrameworkMenu.
2. Modifies the way project/subproject/site filters are populated. Now those filters are displayed with values the current user has permission for.
3. Fixes the data table to remove entries for Projects and Sites the user does not have access to.
4. Fixes the integration tests

## To test
* Visit the module and check that all data entries belong to sites and projects the user has access to.
* Visit the module and check that the site/project/subproject menu filters contain entries the user has access to.

Related to #6829
Related to #6729


laemtl:
@driusan Since the migration of getSiteOptions and getprojectOptions in DataFrameworkMenu per your suggestion, the methods are not accessible to some modules not extending it like this one. Since multiple inheritances is not possible in PHP, this needs to be discussed.

PR #7539, opened 2021-08-19
[Security] Fix SameSite cookie CSRF protection

driusan:
This fixes 2 issues related to the SameSite cookie CSRF protection:

1. The session variable that indicates to use it was set to "true".
   However, it should be a string indicating the value to use (either
   "strict" or "lax"). See: https://www.php.net/manual/en/session.configuration.php#ini.session.cookie-samesite
2. The use_strict_mode setting (See: https://www.php.net/manual/en/session.configuration.php#ini.session.use-strict-mode)
   in PHP is not enabled by default. This means that, even when logging out, the
   session_destroy/session_start procedure reuses the old cookie and does not set a new cookie with
   the proper samesite flag. All the PHP documentation I've seen says use_strict_mode should always
   be enabled, but defaults to disabled. This explicitly overrides the php.ini setting (default false)
   to set it at the beginning of the code, ensuring that NDB_Client properly generates a new session cookie.

### DIFF
--- a/htdocs/index.php
+++ b/htdocs/index.php
@@ -25,6 +25,7 @@ session_cache_limiter("");
 // PHP documentation says this should always be enabled for session security.
 // PHP documentation says this is disabled by default.
 // Explicitly enable it.
+// phpcs:ignore
 // See: https://www.php.net/manual/en/session.configuration.php#ini.session.use-strict-mode
 ini_set('session.use_strict_mode', '1');
 
--- a/htdocs/index.php
+++ b/htdocs/index.php
@@ -22,6 +22,12 @@ require_once __DIR__ . '/../vendor/autoload.php';
 // to be done before NDB_Client starts the PHP session.)
 session_cache_limiter("");
 
+// PHP documentation says this should always be enabled for session security.
+// PHP documentation says this is disabled by default.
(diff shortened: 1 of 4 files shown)


ridz1208 (approved):

php/libraries/NDB_Client.class.inc
```diff
@@ -132,8 +132,8 @@ class NDB_Client
             . $config_additions
         );
         // start php session
-        $sessionOptions = array('cookie_httponly' => true);
-        $sessionOptions['cookie_samesite'] = true;
+        $sessionOptions = ['cookie_httponly' => true];
```
driusan, line 135: The bracket style was changed here to avoid conflicts with main (main is using `[`) when pulling this forward

htdocs/index.php
```diff
@@ -22,6 +22,13 @@
 // to be done before NDB_Client starts the PHP session.)
 session_cache_limiter("");
 
+// PHP documentation says this should always be enabled for session security.
+// PHP documentation says this is disabled by default.
+// Explicitly enable it.
+// phpcs:ignore
```
driusan, line 28: I don't think there's any choice but to ignore the next line since the URL is so long

PR #7514, opened 2021-07-28
[Utility] Refactor getVisitInstruments function

jesscall:
## Brief summary of changes

Comment in Utility Class

>      * @note    Function comment written by Dave, not the author of this function.
>      *       The return value format seems weird. Should possibly be refactored.
>      * @note    Used in:
>      *      - behavioural_qc/php/NDB_Form_behavioural_qc.class.inc
>      *      - behavioural_qc/ajax/GetInstruments.php
>      * @cleanup

I should mention that the files mentioned in the comment no longer exist and as far as I can tell there are no other uses of this function in the code base.

#### Testing instructions (if applicable)

1. Use the utility somewhere and make sure the output is in the form of an array: `[$Test_name => $Full_name]`

PR #7560, opened 2021-08-31
[New Feature] Tracking and Configuration of Candidate's Diagnosis Evolution

jesscall:
## Brief summary of changes
This PR adds a way for project administrators to define and configure diagnosis trajectories for their project. The aim of this new feature is multifaceted; it allows study coordinators to 
1. Track the evolution of a patient's diagnosis over time
2. Determine (& query) the candidate's latest diagnosis is without having to manually compare the longitudinal data

**_Interest for this feature was expressed by multiple projects during the DCC meeting on July 7th._**
 ______

### What is included in this PR?
**1. A configuration page for diagnosis evolution** 

Screen Shot 2021-08-31 at 2 50 15 PM

**2. A tab in candidate parameters to view details on the evolution of the diagnosis.**

Screen Shot 2021-08-31 at 2 51 17 PM

**3. Logic to propagate the latest diagnosis data as instruments are updated.
4. A script that all projects should run when configuring Dx Trajectories and updating their instances.
5. Updates the demographic import to couchDB so the latest diagnosis can be queried.**

_____

### Info
A `Diagnosis Trajectory` is defined by a source field, or group of source fields, that are found within the test battery.
A patient's `Diagnosis Evolution` is defined by the progression of changing diagnoses over time. The evolution is determined by the order of configured `Diagnosis Trajectories`. 

There are two ways that the `Latest Diagnosis` of the candidate can be updated. First, the script in this PR can be run. Secondly, when ANY instrument part of a configured diagnosis track is saved, the latest diagnosis is determined and updated in the `candidate` table.  

EX:

Screen Shot 2021-08-31 at 3 29 58 PM

Screen Shot 2021-08-31 at 3 16 59 PM

1. `fetch jesscall && git checkout jesscall/2021_07_27_diagnosis_evolution`
2. `cat SQL/New_patches/2021-07-28_diagnosis_evolution.sql | mysql`
3. `make dev`
4. Follow instructions in configuration & caniddate parameters TestPlan.md files included in this PR
5. **After setting up the configuration for a diagnosis evolution please do the following:**
6. `php tools/data_integrity/update_candidate_latest_diagnosis.php`
7. Ensure that updating data in a configured instrument updates the latest diagnosis in the `candidate` table (if it's the latest) and the candidate parameters tab.

#### Link(s) to related PR(s)
* Blocked by #7514


zaliqarosli (changes requested):
this is really cool! setting it all up works well. i have a couple bugs, comments and a request.

some bugs:
- if I click on the trajectory tab I just created, then back to 'New Diagnosis Trajectory' tab, the form doesn't re-render as a new form, it still shows the trajectory I just created

Screen Shot 2021-09-30 at 2 44 22 PM

- reset button doesn't work 

request: is there a possibility to add a timestamp?

questions: is candidate_latest_diagnosis really just candidate_diagnosis_rel? i.e. do all the entries in the Diagnosis Evolution table on the front-end get stored in there?

SQL/New_patches/2021-07-28_diagnosis_evolution.sql
```diff
@@ -0,0 +1,25 @@
...
+  `sourceField` varchar(255) DEFAULT NULL,
+  `orderNumber` int(10) unsigned DEFAULT NULL,
+  PRIMARY KEY (`DxEvolutionID`),
+  CONSTRAINT `FK_DxEvolution_1` FOREIGN KEY (`ProjectID`) REFERENCES `Project` (`ProjectID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE TABLE `candidate_latest_diagnosis` (
+  `ID` int(10) unsigned NOT NULL AUTO_INCREMENT,
+  `CandID` int(6) NOT NULL,
+  `ProjectID` int(10) unsigned NOT NULL,
```
zaliqarosli, line 17: i don't think ProjectID needs to be in this table? it can be queried from diagnosis_evolution via the DxEvolutionID, not sure if i'm missing something though

zaliqarosli (changes requested):
1 more bug, the second field from the group of source fields is not getting saved in the LatestDiagnosis column
Screen Shot 2021-09-30 at 4 48 07 PM

modules/configuration/templates/form_configuration.tpl
```diff
@@ -13,13 +13,17 @@
 {/function}
 
 {function name=createInstrument}
+<div>
     <select class="form-control" name="{$k}" {if $d eq "Yes"}disabled{/if}>
-        {foreach from=$instruments key=name item=label}
+        {foreach from=$visits key=name item=label}
```
zaliqarosli, line 18: is this an intended change?

tools/data_integrity/update_candidate_latest_diagnosis.php
```diff
@@ -0,0 +1,126 @@
...
+        foreach ($projectTrajectories as $key => $data) {
+            // search if candidate has a matching visit
+            $sessionID = array_search(
+                $data['visitLabel'],
+                array_reverse($candidateVisits, true)
+            );
+            if ($sessionID) {
+                // Get projectID of this session
+                $timepoint = \TimePoint::singleton(new SessionID($sessionID));
+                $projectID = $timepoint->getProjectID();
```
zaliqarosli, line 66: is this $projectID the same as the $projectID defined in the foreach on line 56?

modules/candidate_parameters/jsx/DiagnosisEvolution.js
```diff
@@ -0,0 +1,172 @@
...
+
+    latestDiagnosis.map((entry) => {
+      const projectName = this.state.data.projects[entry.ProjectID];
+      const diagnosis = entry.LatestDiagnosis ?
+        Object.values(JSON.parse(entry.LatestDiagnosis)).join(', ') : '';
+
+      element.push(
+        <StaticElement
+          key={entry.DxEvolutionID}
+          label={projectName}
```
zaliqarosli, line 109: @jesscall how possible would it be here to render the description of the diagnosis vs the value. QPN has its diagnoses set up so that in the back-end, the options are 1,2,3,etc .. and those values are mapped to proper texts in the linst file i.e. PD, Essential Tremors, Undetermined etc. The proper text would be more useful for the front-end tab, but the value still useful to query from the DQT. I can't think of a way to do this...

jesscall:
Hey @zaliqarosli 
I'm working on reactifying this PR as well as adding a proper endpoint as per @driusan request in a private convo.

I'll try to get to your feedbacks afterwards and see what is still relevant.

As for the rendering of diagnosis for QPN, that's a tough one. I imagine they are dropdown options? That sort of information isn't captured by data dictionary / parameter type. I guess you could hard-code the mapping or use grep on the instrument linst file... But these are both pretty hacky solutions. Maybe @ridz1208  has a suggestion

modules/candidate_parameters/jsx/DiagnosisEvolution.js
jesscall: see below!

modules/configuration/templates/form_configuration.tpl
jesscall: nope! good catch :)

tools/data_integrity/update_candidate_latest_diagnosis.php
jesscall: yes! I'll update

jesscall:
@zaliqarosli 
I've updated the PR. It is now reactified, using a php endpoint rather than ajax and I've also addressed your comments :)
Feel free to re-review at your convenience

@driusan your changes have been implemented

zaliqarosli (commented):
Screen Shot 2021-10-05 at 6 50 24 PM

should the Source Field be the database field name without the instrument prepended to it? i believe that was how it was implemented before reactification

modules/configuration/jsx/DiagnosisEvolution.js
```diff
@@ -0,0 +1,386 @@
...
+                            <NumericElement
+                                name='orderNumber'
+                                min={1}
+                                max={100}
+                                label='Order Number'
+                                onUserInput={this.setFormData}
+                                value={trajectoryData.orderNumber}
+                                required={true}
+                            />
+                            <div>
```
zaliqarosli, line 182:
```suggestion
                            <div
                              style={{
                                display: 'flex',
                                justifyContent: 'center',
                              }}
                            >
```

this will align the Save and Reset buttons in a row

modules/candidate_parameters/ajax/getData.php
```diff
@@ -58,6 +58,9 @@
 case 'candidateDOD':
     echo json_encode(getDODFields());
     exit;
+case 'diagnosisEvolution':
```
zaliqarosli, line 61: is the ajax/getData also being replaced by a real php endpoint?

jesscall:
> Screen Shot 2021-10-05 at 6 50 24 PM
> 
> should the Source Field be the database field name without the instrument prepended to it? i believe that was how it was implemented before reactification

@zaliqarosli 

So in switching from templates to React, the datalist is implemented differently with our predefined react components. Before, you could search by the field name or the table name because both values were displayed in the select option. With the react component you can only have one of the two. So either:
1. **The field name alone**  - EX: `weight_lbs` (parameter_type -> SourceField)
2. **The compounded table name and field name**  - EX: `bmi_weight_lbs` (parameter_type -> Name)

I opted for (2) to make searching by instrument an option for the user. If you want this changed, let me know.

modules/configuration/jsx/DiagnosisEvolution.js
jesscall: done!

jesscall:
Refactored this PR to track the diagnosis evolution as a history rather than the current latest diagnosis. Also implemented @ridz1208 suggestion for showing the confirmation status of a diagnosis (whether the diagnosis comes from an approved visit or not). 

Screen Shot 2021-10-28 at 3 22 20 PM

___

Screen Shot 2021-10-28 at 3 17 45 PM

SQL/New_patches/2021-07-28_diagnosis_evolution.sql
```diff
@@ -0,0 +1,25 @@
...
+  `visitLabel` varchar(255) DEFAULT NULL,
+  `instrumentName` varchar(255) DEFAULT NULL,
+  `sourceField` varchar(255) DEFAULT NULL,
+  `orderNumber` int(10) unsigned DEFAULT NULL,
+  PRIMARY KEY (`DxEvolutionID`),
+  UNIQUE KEY `TrajectoryName` (`Name`),
+  CONSTRAINT `FK_DxEvolution_1` FOREIGN KEY (`ProjectID`) REFERENCES `Project` (`ProjectID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE TABLE `candidate_diagnosis_evolution` (
```
zaliqarosli, line 14:
```suggestion
CREATE TABLE `candidate_diagnosis_evolution_rel` (
```

maybe?

SQL/New_patches/2021-07-28_diagnosis_evolution.sql
```diff
@@ -0,0 +1,25 @@
...
+  `ID` int(10) unsigned NOT NULL AUTO_INCREMENT,
+  `CandID` int(6) NOT NULL,
+  `DxEvolutionID` int(10) unsigned NOT NULL,
+  `Diagnosis` text DEFAULT NULL,
+  `Confirmed` enum('Y', 'N') DEFAULT NULL,
+  `LastUpdate` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY (`ID`),
+  UNIQUE KEY `candidateDxEvolution` (`CandID`, `DxEvolutionID`),
```
zaliqarosli, line 22: and then i think you can make the Primary Keys CandID and DxEvolution, and remove the ID key. kind of like the candidate_consent_rel table if i understand correctly that only the CandID and DxEvolutionID combination needs to be unique

zaliqarosli (changes requested):

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2224,6 +2225,123 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+        foreach ($sourceFields as $k => $fieldName) {
+            // None of the diagnosis components should be empty
+            // If so, the latest diagnosis should not be updated.
+            if (is_null($instrumentData[$fieldName])) {
+                return;
+            }
+            $diagnosis[$fieldName] = $instrumentData[$fieldName];
+        }
+
+        $confirmed = $timepoint->getApprovalStatus() === 'Pass' ? 'Y' : 'N';
```
zaliqarosli, line 2327: i'm not too familiar with the timepoint approval status. what data is this approval status dependent on and how does it compare to the instrument status data_entry?

i am just thinking about the case where a timepoint has two different instruments capturing diagnosis information i.e. one instrument administered at the start, and the other at the end. would the first diagnosis not be confirmed until the second diagnosis in collected, and the whole timepoint approved?

SQL/New_patches/2021-07-28_diagnosis_evolution.sql
jesscall: I think `rel` would imply that the diagnosis value itself is stored elsewhere. This table links to the configured Dx tracjectory, but the Diagnosis is only present here. not sure if that makes sense.

SQL/New_patches/2021-07-28_diagnosis_evolution.sql
jesscall: dxEvolutionID refers to the configuration id, whereas the id of this table refers to the candidate's actual diagnosis for that configuration.

more like config / configSettings table

SQL/New_patches/2021-07-28_diagnosis_evolution.sql
jesscall: I made the changes you requested.

There will be one row in the table per candidate/ DxEvolutionID. The latest Dx is just the one with highest orderNumber (but all are stored in the table)

jesscall:
@driusan ready for testing and review!

SQL/New_patches/2021-07-28_diagnosis_evolution.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TABLE `diagnosis_evolution` (
+  `DxEvolutionID` int(10) unsigned NOT NULL AUTO_INCREMENT,
+  `Name` varchar(255) DEFAULT NULL,
+  `ProjectID` int(10) unsigned DEFAULT NULL,
```
zaliqarosli, line 4:
```suggestion
  `ProjectID` int(10) unsigned NOT NULL,
```

should this be not null because an evolution needs to belong to a project specific evolution?

SQL/New_patches/2021-07-28_diagnosis_evolution.sql
```diff
@@ -0,0 +1,24 @@
...
+  `visitLabel` varchar(255) DEFAULT NULL,
+  `instrumentName` varchar(255) DEFAULT NULL,
+  `sourceField` varchar(255) DEFAULT NULL,
+  `orderNumber` int(10) unsigned DEFAULT NULL,
+  PRIMARY KEY (`DxEvolutionID`),
```
zaliqarosli, line 9:
```suggestion
  CONSTRAINT `PK_diagnosis_evolution` PRIMARY KEY (`DxEvolutionID`),
```

SQL/New_patches/2021-07-28_diagnosis_evolution.sql
```diff
@@ -0,0 +1,24 @@
...
+  UNIQUE KEY `TrajectoryName` (`Name`),
```
zaliqarosli, line 10:
```suggestion
  CONSTRAINT `UK_diagnosis_evolution_Name` UNIQUE KEY `Name` (`Name`),
```

SQL/New_patches/2021-07-28_diagnosis_evolution.sql
```diff
@@ -0,0 +1,24 @@
...
+  CONSTRAINT `FK_DxEvolution_1` FOREIGN KEY (`ProjectID`) REFERENCES `Project` (`ProjectID`)
```
zaliqarosli, line 11:
```suggestion
  CONSTRAINT `FK_diagnosis_evolution_ProjectID` FOREIGN KEY (`ProjectID`) REFERENCES `Project` (`ProjectID`) ON DELETE RESTRICT ON UPDATE CASCADE
```

SQL/New_patches/2021-07-28_diagnosis_evolution.sql
```diff
@@ -0,0 +1,24 @@
...
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE TABLE `candidate_diagnosis_evolution_rel` (
+  `CandID` int(6) NOT NULL,
+  `DxEvolutionID` int(10) unsigned NOT NULL,
+  `Diagnosis` text DEFAULT NULL,
+  `Confirmed` enum('Y', 'N') DEFAULT NULL,
+  `LastUpdate` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```
zaliqarosli, line 19:
```suggestion
  `LastUpdate` datetime NOT NULL DEFAULT NOW() ON UPDATE NOW(),
```

SQL/New_patches/2021-07-28_diagnosis_evolution.sql
```diff
@@ -0,0 +1,24 @@
...
+  PRIMARY KEY (`CandID`, `DxEvolutionID`),
+  UNIQUE KEY `candidateDxEvolution` (`CandID`, `DxEvolutionID`),
```
zaliqarosli, line 21:
```suggestion

```

SQL/New_patches/2021-07-28_diagnosis_evolution.sql
```diff
@@ -0,0 +1,24 @@
...
+  CONSTRAINT `FK_candidateDxEvolution_1` FOREIGN KEY (`CandID`) REFERENCES `candidate` (`CandID`),
+  CONSTRAINT `FK_candidateDxEvolution_2` FOREIGN KEY (`DxEvolutionID`) REFERENCES `diagnosis_evolution` (`DxEvolutionID`)
```
zaliqarosli, line 23:
```suggestion
  CONSTRAINT `FK_candidate_diagnosis_evolution_rel_DxEvolutionID` FOREIGN KEY (`DxEvolutionID`) REFERENCES `diagnosis_evolution` (`DxEvolutionID`)
```

SQL/New_patches/2021-07-28_diagnosis_evolution.sql
```diff
@@ -0,0 +1,24 @@
...
+  `instrumentName` varchar(255) DEFAULT NULL,
+  `sourceField` varchar(255) DEFAULT NULL,
+  `orderNumber` int(10) unsigned DEFAULT NULL,
+  PRIMARY KEY (`DxEvolutionID`),
+  UNIQUE KEY `TrajectoryName` (`Name`),
+  CONSTRAINT `FK_DxEvolution_1` FOREIGN KEY (`ProjectID`) REFERENCES `Project` (`ProjectID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE TABLE `candidate_diagnosis_evolution_rel` (
+  `CandID` int(6) NOT NULL,
```
zaliqarosli, line 15: @ridz1208 do you care if this is called `CandID` vs `CandidateID`?

zaliqarosli (changes requested):
sorry i'm being very picky! a few things to abide by the [SQL modelling standard](https://github.com/aces/Loris/blob/main/docs/SQLModelingStandard.md)

SQL/0000-00-00-schema.sql
```diff
@@ -141,6 +141,19 @@ CREATE TABLE `caveat_options` (
...
+  `ProjectID` int(10) unsigned DEFAULT NULL,
+  `visitLabel` varchar(255) DEFAULT NULL,
+  `instrumentName` varchar(255) DEFAULT NULL,
+  `sourceField` varchar(255) DEFAULT NULL,
+  `orderNumber` int(10) unsigned DEFAULT NULL,
+  PRIMARY KEY (`DxEvolutionID`),
+  UNIQUE KEY `TrajectoryName` (`Name`),
+  CONSTRAINT `FK_DxEvolution_1` FOREIGN KEY (`ProjectID`) REFERENCES `Project` (`ProjectID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
```
jesscall, line 156:
```suggestion
CREATE TABLE `diagnosis_evolution` (
  `DxEvolutionID` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `Name` varchar(255) DEFAULT NULL,
  `ProjectID` int(10) unsigned NOT NULL,
  `visitLabel` varchar(255) DEFAULT NULL,
  `instrumentName` varchar(255) DEFAULT NULL,
  `sourceField` varchar(255) DEFAULT NULL,
  `orderNumber` int(10) unsigned DEFAULT NULL,
  CONSTRAINT `PK_diagnosis_evolution` PRIMARY KEY (`DxEvolutionID`),
  CONSTRAINT `UK_diagnosis_evolution_Name` UNIQUE KEY `Name` (`Name`),
  CONSTRAINT `FK_diagnosis_evolution_ProjectID` FOREIGN KEY (`ProjectID`) REFERENCES `Project` (`ProjectID`) ON DELETE RESTRICT ON UPDATE CASCADE
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

SQL/0000-00-00-schema.sql
```diff
@@ -178,6 +191,18 @@ CREATE TABLE `candidate` (
...
+  `CandID` int(6) NOT NULL,
+  `DxEvolutionID` int(10) unsigned NOT NULL,
+  `Diagnosis` text DEFAULT NULL,
+  `Confirmed` enum('Y', 'N') DEFAULT NULL,
+  `LastUpdate` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY (`CandID`, `DxEvolutionID`),
+  UNIQUE KEY `candidateDxEvolution` (`CandID`, `DxEvolutionID`),
+  CONSTRAINT `FK_candidateDxEvolution_1` FOREIGN KEY (`CandID`) REFERENCES `candidate` (`CandID`),
+  CONSTRAINT `FK_FK_candidateDxEvolution_1_2` FOREIGN KEY (`DxEvolutionID`) REFERENCES `diagnosis_evolution` (`DxEvolutionID`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
jesscall, line 204:
```suggestion
  `DxEvolutionID` int(10) unsigned NOT NULL,
  `Diagnosis` text DEFAULT NULL,
  `Confirmed` enum('Y', 'N') DEFAULT NULL,
  `LastUpdate` datetime NOT NULL DEFAULT NOW() ON UPDATE NOW(),
  CONSTRAINT `PK_candidate_diagnosis_evolution_rel` PRIMARY KEY (`CandidateID`, `DiagnosisEvolutionID`),
  CONSTRAINT `FK_candidate_diagnosis_evolution_rel_CandID` FOREIGN KEY (`CandID`) REFERENCES `candidate` (`CandID`) ON DELETE RESTRICT ON UPDATE RESTRICT,
  CONSTRAINT `FK_candidate_diagnosis_evolution_rel_DxEvolutionID` FOREIGN KEY (`DxEvolutionID`) REFERENCES `diagnosis_evolution` (`DxEvolutionID`) ON DELETE RESTRICT ON UPDATE CASCADE
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

jesscall:
@driusan @zaliqarosli @ridz1208 can someone help with phan ?

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2233,6 +2234,121 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+            WHERE DxEvolutionID=:dxEvID",
+            ['dxEvID' => $dxEvolutionID]
+        );
+
+        $this->instanceData = null;
+        $instrumentData     = $this->getInstanceData();
+        $sourceFields       = explode(",", $sourceFields);
+
+        $diagnosis = [];
+        foreach ($sourceFields as $k => $fieldName) {
```
zaliqarosli, line 2325:
```suggestion
        foreach ($sourceFields as $fieldName) {
```
php/libraries/NDB_BVL_Instrument.class.inc:2325 PhanUnusedVariable Unused definition of variable $k

php/libraries/Utility.class.inc
```diff
@@ -1020,4 +1020,41 @@ class Utility
...
+        $dxToUpdate = $DB->pselectRow(
+            "SELECT cde.CandID, cde.DxEvolutionID
+            FROM candidate_diagnosis_evolution_rel cde 
+            JOIN diagnosis_evolution de USING (DxEvolutionID)
+            JOIN session s ON (cde.CandID=s.CandID AND s.Visit_label=de.visitLabel)
+            WHERE s.ID=:sid",
+            ['sid' => $sessionID]
+        );
+
+        foreach ($dxToUpdate as $key => $row) {
```
zaliqarosli, line 1049:
```suggestion
        foreach ($dxToUpdate as $row) {
```
php/libraries/Utility.class.inc:1049 PhanUnusedVariable Unused definition of variable $key

zaliqarosli (changes requested):

SQL/New_patches/2021-07-28_diagnosis_evolution.sql
```diff
@@ -0,0 +1,23 @@
...
+  CONSTRAINT `FK_diagnosis_evolution_ProjectID` FOREIGN KEY (`ProjectID`) REFERENCES `Project` (`ProjectID`) ON DELETE RESTRICT ON UPDATE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE TABLE `candidate_diagnosis_evolution_rel` (
+  `CandID` int(6) NOT NULL,
+  `DxEvolutionID` int(10) unsigned NOT NULL,
+  `Diagnosis` text DEFAULT NULL,
+  `Confirmed` enum('Y', 'N') DEFAULT NULL,
+  `LastUpdate` datetime NOT NULL DEFAULT NOW() ON UPDATE NOW(),
+  CONSTRAINT `PK_candidate_diagnosis_evolution_rel` PRIMARY KEY (`CandidateID`, `DiagnosisEvolutionID`),
```
zaliqarosli, line 20:
```suggestion
  CONSTRAINT `PK_candidate_diagnosis_evolution_rel` PRIMARY KEY (`CandID`, `DxEvolutionID`),
```

SQL/0000-00-00-schema.sql
```diff
@@ -178,6 +190,17 @@ CREATE TABLE `candidate` (
   CONSTRAINT `FK_candidate_RegistrationProjectID` FOREIGN KEY (`RegistrationProjectID`) REFERENCES `Project` (`ProjectID`) ON UPDATE CASCADE
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
+CREATE TABLE `candidate_diagnosis_evolution_rel` (
+  `CandID` int(6) NOT NULL,
+  `DxEvolutionID` int(10) unsigned NOT NULL,
+  `Diagnosis` text DEFAULT NULL,
+  `Confirmed` enum('Y', 'N') DEFAULT NULL,
+  `LastUpdate` datetime NOT NULL DEFAULT NOW() ON UPDATE NOW(),
+  CONSTRAINT `PK_candidate_diagnosis_evolution_rel` PRIMARY KEY (`CandidateID`, `DiagnosisEvolutionID`),
```
zaliqarosli, line 199:
```suggestion
  CONSTRAINT `PK_candidate_diagnosis_evolution_rel` PRIMARY KEY (`CandID`, `DxEvolutionID`),
```

modules/configuration/jsx/DiagnosisEvolution.js
```diff
@@ -0,0 +1,386 @@
...
+                    Use this page to manage the configuration of the study's
+                    diagnosis trajectory.
+                </p>
+                <p>
+                    To configure study subprojects
+                    <a href="{$baseurl}/configuration/subproject/"> click here
+                    </a>.
+                    To configure study projects
+                    <a href="{$baseurl}/configuration/project/"> click here</a>.
+                </p>
```
zaliqarosli, line 242:
```suggestion

```
i dont think these links are needed on this page, they also don't work for me

zaliqarosli: yeahh i think it's fine if theyre not there, users can use the breadcrumbs to navigate to the previous page where these links can be found

modules/configuration/php/diagnosis.class.inc
```diff
@@ -0,0 +1,285 @@
...
+            );
+        }
+
+        // Validation: Order Number should be unique for specified project
+        $DB    = \NDB_Factory::singleton()->database();
+        $match = $DB->pselectOne(
+            "SELECT Name
+            FROM diagnosis_evolution
+            WHERE ProjectID=:projID
+            AND orderNumber=:ord",
```
zaliqarosli, line 193:
```suggestion
            AND orderNumber=:ord
            AND NOT FIND_IN_SET(DxEvolutionID, :dxEvoID)",
```

modules/configuration/php/diagnosis.class.inc
```diff
@@ -0,0 +1,285 @@
...
+            [
+                'projID' => $projectID,
+                'ord'    => $orderNumber
```
zaliqarosli, line 196:
```suggestion
                'ord'    => $orderNumber,
                'dxEvoID' => $dxEvolutionID
```

this validation returns conflict error for orderNumber if you're updating an existing evolution

zaliqarosli (changes requested):

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2233,6 +2234,121 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+        $instrumentData     = $this->getInstanceData();
+        $sourceFields       = explode(",", $sourceFields);
+
+        $diagnosis = [];
+        foreach ($sourceFields as $fieldName) {
+            // None of the diagnosis components should be empty
+            // If so, the latest diagnosis should not be updated.
+            if (is_null($instrumentData[$fieldName])) {
+                return;
+            }
```
zaliqarosli, line 2330: hey @jesscall just wondering what the original workflow/use case is. i'm having a bit of an issue right now for the COPN workflow where a set of let's say 3 diagnosis fields do end up possibly having at least 1 component empty e.g.
Screen Shot 2022-06-15 at 3 41 03 PM

if the third field is left empty because it doesn't need to be filled out or not_answered is chosen, the diagnosis data isn't updated. what are the issues you've come across with updating them anyways when some fields are missing?

zaliqarosli: also, just to add, i think the data_integrity script allows diagnosis data to be inserted/updated if not all fields are filled out

jesscall: I think this if statement can be removed or at least updated to check that at least 1 of the diagnosis fields has data

tools/data_integrity/update_candidate_latest_diagnosis.php
```diff
@@ -0,0 +1,129 @@
...
+                    $data['instrumentName'],
+                    $commentID
+                );
+                $instrumentData = $instrument->getInstanceData();
+
+
+                $diagnosis    = [];
+                $sourceFields = explode(",", $data['sourceField']);
+                foreach ($sourceFields as $k => $fieldName) {
+                    if (!isset($instrumentData[$fieldName])) {
```
zaliqarosli, line 95:
```suggestion
                    if (!isset($instrumentData[$fieldName]) || (isset($instrumentData[$fieldName] && is_null($instrumentData[$fieldName]))) {
```

zaliqarosli (changes requested):
@jesscall this should allow the diagnosis to be updated as long as at least 1 field is filled out :)

modules/candidate_parameters/jsx/DiagnosisEvolution.js
```diff
@@ -0,0 +1,223 @@
...
+          <StaticElement
+            label='PSCID'
+            text={this.state.data.pscid}
+          />
+          <StaticElement
+            label='DCCID'
+            text={this.state.data.candID}
+          />
+          {latestDiagnosis}
+          {latestConfirmedDiagnosis}
```
zaliqarosli, line 197:
```suggestion
          {latestConfirmedDiagnosis || latestDiagnosis}
```

at the moment, both confirmed and not confirmed show up. i assume that once the diagnosis is confirmed, that's the only one that should be shown?

Screen Shot 2022-06-16 at 12 54 06 PM

zaliqarosli (changes requested):

modules/candidate_parameters/jsx/DiagnosisEvolution.js
jesscall: Not necessarily, there could be a confirmed diagnosis from V1 and an unconfirmed diagnosis from V2. In this case, the V2 unconfirmed dx is actually the latest.

However the case shown in this screenshot could call for a wording change. The unconfirmed section could say to be cautious that it MAY be unconfirmed ? What do you think ?

zaliqarosli: hmm i see, that makes sense. i think if only 1 diagnosis trajectory exists though and its confirmed, then we should only be showing that its confirmed, since its a bit confusing if both confirmed and unconfirmed exist. could we do something like this?

Screen Shot 2022-06-16 at 6 16 12 PM
Screen Shot 2022-06-16 at 6 16 04 PM

zaliqarosli: ill request line changes in a separate comment for this to work

modules/candidate_parameters/jsx/DiagnosisEvolution.js
```diff
@@ -0,0 +1,223 @@
...
+  render() {
+    if (this.state.error) {
+        return <h3>An error occured while loading the page.</h3>;
+    }
+
+    if (!this.state.isLoaded) {
+        return <Loader/>;
+    }
+
+    const latestDiagnosis = this.state.data.latestProjectDiagnosis.length > 0 ?
```
zaliqarosli, line 156:
```suggestion

    const latestConfirmedProjectDiagnosisID = this.state.data.latestConfirmedProjectDiagnosis.length > 0 ?  this.state.data.latestConfirmedProjectDiagnosis[0].DxEvolutionID : null;
    const latestProjectDiagnosisID = this.state.data.latestProjectDiagnosis.length > 0 ?
    this.state.data.latestProjectDiagnosis[0].DxEvolutionID : null;
    
    const latestDiagnosis = (this.state.data.latestProjectDiagnosis.length > 0 &&
    latestConfirmedProjectDiagnosisID != latestProjectDiagnosisID) ?
```

modules/candidate_parameters/jsx/DiagnosisEvolution.js
jesscall: yes I'm okay with this change

modules/candidate_parameters/jsx/DiagnosisEvolution.js
jesscall: If I'm remembering correctly, the `latestConfirmedProjectDiagnosis` and `latestProjectDiagnosis` will have an entry for each of the candidate's projects. What happens in this case if there is more than one project (even if only 1 trajectory is defined for each project)? By explicitly accessing the first array position [0], you're only fixing the display for the first project. 

Not sure if my comment is clear.

zaliqarosli: yeah you're clear! good point, i will think about this :)

zaliqarosli: sent a PR to your branch fixing this! https://github.com/jesscall/Loris/pull/1

tools/data_integrity/update_candidate_latest_diagnosis.php
```diff
@@ -0,0 +1,129 @@
...
+                        continue;
+                    }
+                    $diagnosis[$fieldName] = $instrumentData[$fieldName];
+                }
+
+                if (!empty($diagnosis)) {
+                    $confirmed = \TimePoint::singleton(
+                        new SessionID($sessionID)
+                    )->getVisitStatus() === 'Pass' ?
```
zaliqarosli, line 104: similarly to above comment, should the diagnosis be confirmed on the visit status passing (as here), the approval status starting (as via control panel), or the approval status passing? on the jsx, a confirmed diagnosis is described as "one that belongs to an approved visit" but im not quite sure what this means.

modules/instrument_list/php/instrument_list_controlpanel.class.inc
```diff
@@ -142,10 +142,17 @@ class Instrument_List_ControlPanel extends \TimePoint
...
             if ($currentStatus == 'Pass') {
                 $nextStage = 'Approval';
+                $isPassed  = true;
             } else {
                 $nextStage = 'Recycling Bin';
+                $isPassed  = false;
             }
             $this->startStage($nextStage);
+            // update candidate_diagnosis_evolution_rel
+            \Utility::updateDiagnosisConfirmationStatus(
```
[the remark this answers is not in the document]
jesscall: We discussed this in a LORIS meeting, and Samir said Send to DCC was adequate for confirming diagnosis. The Approval stage being set to pass is something that's not currently part of the CCNA workflow

tools/data_integrity/update_candidate_latest_diagnosis.php
jesscall: Yes this was recently changed after discussion at LORIS meeting. Visit pass is sufficient.

zaliqarosli: how come visit pass isn't sufficient when using the control panel, where the timepoint needs to be in an approval stage i.e. when sent to DCC? i don't see why the workflow would be different between front-end and back-end data entry. whats the reasoning?

tools/data_integrity/update_candidate_latest_diagnosis.php
```diff
@@ -98,11 +98,16 @@
...
-                        'Y' : 'N'
-                    ;
+                    );
+                    // Check for Visit Status 'Pass' and
+                    // Approval Status at least 'In Progress'
+                    // to check that Send To DCC already occurred
+                    $confirmed = (
+                        $timepoint->getVisitStatus() === 'Pass'
+                        && $timepoint->getApprovalStatus() !== null
+                    ) ? 'Y' : 'N';
```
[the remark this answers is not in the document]
CamilleBeau: Currently on CCNA we are not using "Send to DCC", however it is our intention to use it soon, so I think this workflow works well!

tools/CouchDB_Import_Demographics.php
```diff
@@ -287,10 +287,39 @@ function _generateQuery()
                             $cField.DateWithdrawn";
             }
         }
+
+        // Latest Diagnosis by project
+        $projects = \Utility::getProjectList();
+        foreach ($projects as $projectID => $project) {
+            $projectName  = str_replace(' ', '_', $project);
+            $latestProjDx = "latestDiagnosis_$projectName";
+
```
CamilleBeau, line 296: To avoid error with projects that have a space in their name
```suggestion
            $projectAlias = \Project::getProjectFromID($projectID)->getAlias();
            $latestProjDx = "latestDiagnosis_$projectAlias";

```

tools/CouchDB_Import_Demographics.php
```diff
@@ -353,6 +382,18 @@ function _updateDataDict()
                 ];
             }
         }
+
+        // Update data dictionary for latest diagnosis by project
+        $projects = \Utility::getProjectList();
+        foreach ($projects as $projectID => $project) {
+            $projectName = str_replace(' ', '_', $project);
+            $fieldName   = "latestDiagnosis_" . $projectName;
```
CamilleBeau, line 390:
```suggestion
            $projectAlias = \Project::getProjectFromID($projectID)->getAlias();
            $fieldName   = "latestDiagnosis_" . $projectAlias;

File: modules/configuration/php/diagnosis.class.inc [hunk 36]
```diff
@@ -0,0 +1,288 @@
...
+        if (!($dxEvolutionID && $name && $visit
+            && $instrumentName && $sourceFields && $orderNumber)
+        ) {
+            return new \LORIS\Http\Response\JSON\BadRequest(
+                'Please fill out all fields!'
+            );
+        }
+
+        // Validation: Visit is part of Project's config
+        $projectVisits = \Utility::getVisitList(new \ProjectID($projectID));
```
CamilleBeau, line 161: This caused a bug on CCNA when trying to set diagnosis evolution if the visits haven't yet been started, so I did a select from test_battery instead of session
```suggestion
        $projectVisits = $DB->pselectCol(
            "SELECT DISTINCT Visit_label FROM test_battery tb
            JOIN project_subproject_rel psr ON (psr.SubprojectID=tb.SubprojectID)
            WHERE psr.ProjectID=:projID
            ORDER BY Visit_label",
            ['projID' => $projectID]
        );
	```

File: modules/configuration/php/diagnosis.class.inc [hunk 37]
```diff
@@ -0,0 +1,288 @@
...
+        if (!array_key_exists($visit, $projectVisits)) {
+            return new \LORIS\Http\Response\JSON\Conflict(
+                'Visit is not defined for selected Project.'
+            );
+        }
+
+        // Validation: Instrument is part of Visit's test battery
+        $visitInstruments = \Utility::getVisitInstruments($visit);
+        if (!array_key_exists($instrumentName, $visitInstruments)) {
```
CamilleBeau, line 170:
```suggestion
        $visitInstruments = $DB->pselectCol(
            "SELECT DISTINCT test_name FROM test_battery
            WHERE Visit_label=:visit
            ORDER BY test_name",
            ['visit' => $visit]
        );

        if (!in_array($instrumentName, $visitInstruments)) {
```

zaliqarosli:
TO-DO:

- [ ] implement Camille's comments
- [ ] change Trajectory Name 409 Conflict to case-insensitive

modules/configuration/php/configuration.class.inc
```diff
@@ -67,7 +67,7 @@ class Configuration extends \NDB_Form
             'Ym'  => 'Ym (EX: 2015-12)'
         ];
 
-        $instruments     = \Utility::getAllInstruments();
+        $instruments     = \Utility::getInstrumentNamesList();
```
zaliqarosli, line 70: what's this change necessary for? @CamilleBeau

CamilleBeau: https://github.com/aces/Loris/blob/c70e61511182c7ad8d6ff8d8c92949d03cb2a540/php/libraries/Utility.class.inc#L523

This function is being deprecated & replaced by getInstrumentNamesList

modules/configuration/php/diagnosis.class.inc
```diff
@@ -0,0 +1,337 @@
...
+     */
+    public function handle(ServerRequestInterface $request) : ResponseInterface
+    {
+        // Ensure GET or POST request.
+        switch ($request->getMethod()) {
+        case 'GET':
+            return $this->_handleGET();
+        case 'POST':
+            $values = $request->getParsedBody();
+            $action = isset($values['action']) ?? null;
```
CamilleBeau, line 61:
```suggestion
            $action = $values['action'] ?? null;
```

modules/configuration/test/TestPlan.md
```diff
@@ -72,3 +72,25 @@ project_cohort_rel table in the database.
   [Manual Test]
 5. Test the breadcrumb to see if you can navigate back to the main config page.
   [Automation Test]
+6. Go back to the project configuration page and click on the subproject link and 
+confirm that it brings you to the subproject configuration page.
```
zaliqarosli, line 76:
```suggestion
6. Go back to the project configuration page. Click on the cohort link at the top of the page and
confirm that it brings you to the cohort configuration page.
```

modules/configuration/test/TestPlan.md
```diff
@@ -72,3 +72,25 @@ project_cohort_rel table in the database.
...
+  [Manual Test]
+
+## Diagnosis Evolution 
+
+1. Click on the diagnosis evolution link at the top of the page and confirm that it brings you to
+the diagnosis trajectory configuration page.
+  [Manual Test]
+2. Click the 'New Diagnosis Trajectory' tab in the left column navigation. Try to add a new trajectory.
+  [Manual Test]
+3. Test that you cannot add a diagnosis trajectory for any source field or instrument that does not exist in the selected visit. Also test that you can add multiple source fields.
```
zaliqarosli, line 86: having an empty "Source Field" field (when you dont actually click on the "Add Field" button) gives a 409 "Source Field does not exist in instrument" error when the form should handle its own errors for required fields not filled out. happens for the trajectory name and order number fields as well

Screen Shot 2022-12-19 at 1 21 44 PM

modules/configuration/test/TestPlan.md
```diff
@@ -72,3 +72,25 @@ project_cohort_rel table in the database.
...
+  [Manual Test]
+4. Validate that you cannot create duplicate trajectory names.
+  [Manual Test]
+5. Validate that you cannot duplicate an order number within the same project.
+  [Manual Test]
+6. Test editing a currently existing trajectory. Test resetting the form as well with 
+the reset button.
+  [Manual Test]
```
zaliqarosli, line 94: Add step to test deletion

zaliqarosli (changes requested):
for the deletion, once you click 'Delete', the page immediately goes blank (which kind of looks like a bug even though it isnt) and then a 'Submission successful' confirmation pops up. because its a deletion, i think you could pop up a swal on clicking the delete button that asks the user to confirm the deletion, and then display a confirmation on the swal saying 'Deletion successful'

modules/candidate_parameters/test/TestPlan.md
```diff
@@ -69,51 +69,59 @@
...
+2. Change the date of death and try to save it. Check that it is updated on front-end and database.
+3. Check that entering a date of death before the date of birth gives a data entry error.
+4. Check that entering a date later than today gives a data entry error.
+5. Check that the date stored matches the _DOD Format_ specified in the configuration module.
+
+### Diagnosis Evolution Tab
+1. Confirm that all the fields in this panel correspond to what's stored in the candidate table.
+	* PSCID
+    * DCCID
+    * Latest Diagnosis
```
zaliqarosli, line 88:
```suggestion
    * PSCID
    * DCCID
    * Latest Diagnosis
```

zaliqarosli: for some reason, a project diagnosis evolution configuration is being applied to a candidate's visit in another project
Screen Shot 2022-12-19 at 1 50 03 PM
Screen Shot 2022-12-19 at 1 50 15 PM

zaliqarosli (changes requested):

CamilleBeau:
@zaliqarosli Changes made!

zaliqarosli (changes requested):
1. The error messages are a bit buggy still
Screenshot 2022-12-22 at 4 25 42 PM

2. The resetting is broken now, it resets to an empty form, instead of what is currently in database
3. Can you remind me how a diagnosis gets confirmed again? I'm not able to get one confirmed (this might be a bug?)

CamilleBeau:
@zaliqarosli I finally got back to this!

> 1. The error messages are a bit buggy still
> Screenshot 2022-12-22 at 4 25 42 PM

I couldn't reproduce this error, do you have any more details?

> 2. The resetting is broken now, it resets to an empty form, instead of what is currently in database
Fixed!

> 3. Can you remind me how a diagnosis gets confirmed again? I'm not able to get one confirmed (this might be a bug?)
The diagnosis is confirmed when the visit is set to "Send to DCC".

tools/CouchDB_Import_Demographics.php
```diff
@@ -277,10 +277,39 @@ function _generateQuery()
...
+                    WHERE ProjectID=$projectID
+                ) AS $latestProjDx ON ($latestProjDx.CandID=c.CandID)";
+            $groupBy       .= ", $latestProjDx.Diagnosis";
+        }
+
         $whereClause = " WHERE s.Active='Y' AND c.Active='Y' "
                        ."AND c.Entity_type != 'Scanner'";
 
         $concatQuery = $fieldsInQuery . $tablesToJoin . $whereClause . $groupBy;
+        print_r($concatQuery);
```
CamilleBeau, line 312: I think this is a debugging print statement

CamilleBeau (changes requested):
Just one small comment

tools/CouchDB_Import_Demographics.php
zaliqarosli, line 312:
```suggestion
```

zaliqarosli: @CamilleBeau resolved!

CamilleBeau (approved):

driusan:
I'm not sure who's maintaining this PR now but the tests on it are failing and there's a few eslint warnings that need to be fixed (the error ones are the reason it's failing)

```
/home/runner/work/Loris/Loris/modules/configuration/jsx/DiagnosisEvolution.js
  125:1   error    The type 'int' is undefined                                           jsdoc/no-undefined-types
  303:1   error    The type 'bool' is undefined                                          jsdoc/no-undefined-types
  399:13  warning  Unexpected console statement                                          no-console
  494:13  warning  Unexpected console statement                                          no-console
  504:1   warning  @param "id" does not match an existing function parameter             jsdoc/check-param-names
  548:1   warning  @param "pendingValKey" does not match an existing function parameter  jsdoc/check-param-names
```

zaliqarosli:
@driusan ya im working on it

driusan:
I am getting 2 errors and a blank page on the configuration -> Diagnosis Evolution page:

1.
```
Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot [react-dom.development.js:10:1247](http://localhost:8000/vendor/js/react/react-dom.development.js)
```

2.
```
Uncaught ReferenceError: FormElement is not defined
    renderDiagnosisForm DiagnosisEvolution.js:158
    render DiagnosisEvolution.js:291
    React 16
    componentDidMount DiagnosisEvolution.js:54
    promise callback*componentDidMount DiagnosisEvolution.js:54
    React 12
    <anonymous> DiagnosisEvolution.js:582
    EventListener.handleEvent* DiagnosisEvolution.js:581
    <anonymous> DiagnosisEvolution.js:6
    <anonymous> DiagnosisEvolution.js:6
[DiagnosisEvolution.js:158:20](webpack://lorisjs.configuration/modules/configuration/jsx/DiagnosisEvolution.js)
    renderDiagnosisForm DiagnosisEvolution.js:158
    render DiagnosisEvolution.js:291
    React 11
    performSyncWorkOnRoot self-hosted:1406
    React 5
    componentDidMount DiagnosisEvolution.js:54
    (Async: promise callback)
    componentDidMount DiagnosisEvolution.js:54
    React 7
    performSyncWorkOnRoot self-hosted:1406
    React 5
    <anonymous> DiagnosisEvolution.js:582
    (Async: EventListener.handleEvent)
    <anonymous> DiagnosisEvolution.js:581
    <anonymous> DiagnosisEvolution.js:6
    <anonymous> DiagnosisEvolution.js:6
```

The first seems to be because something is written using an old/deprecated style from react. The second is because import statements from `jsx/Form` are missing (I'm getting similar errors on the candidate_parameters page but I'll fix those on the other tabs and send a PR since I think that's broken on main.)

zaliqarosli:
@driusan fixed, and checks passed. could you have another look?
@CamilleBeau would you be able to do a final quick run through test? i had to make some extra changes after the rebase (see last 4 commits)

driusan (commented):
This is pretty cool, just some small comments.

I also find the "SearchableDropdown" on the configuration page confusing because it looks like a text field.

modules/candidate_parameters/jsx/DiagnosisEvolution.js
```diff
@@ -0,0 +1,246 @@
...
+      </div>
+      : null;
+    const latestConfirmedDiagnosis =
+      this.state.data.latestConfirmedProjectDiagnosis.length > 0 ?
+        <div className='col-md-10'>
+          <h3>Latest Confirmed Diagnosis</h3>
+          <p>This diagnosis is <strong style={{color: 'green'}}>
+            confirmed</strong>.
+            A confirmed diagnosis is one that belongs to a visit
+            that has been sent to DCC for approval.
```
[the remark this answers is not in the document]
zaliqarosli: that's the workflow that was agreed on during a loris meeting, and what was implemented by CCNA. sending a timepoint to DCC makes the diagnosis for that configured visit confirmed, before that the diagnosis isn't confirmed

modules/candidate_parameters/jsx/DiagnosisEvolution.js
```diff
@@ -0,0 +1,246 @@
...
+            text={this.state.data.candID}
+          />
+          {latestDiagnosis}
+          {latestConfirmedDiagnosis}
+          <h3>Diagnosis Evolution</h3>
+          <StaticDataTable
+            Headers={[
+              'Trajectory Name',
+              'Project',
+              'Configured Order',
```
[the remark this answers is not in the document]
zaliqarosli: no i don't think so. i think the order determines the order of the display in the evolution

SQL/0000-00-00-schema.sql
```diff
@@ -143,6 +143,19 @@ CREATE TABLE `caveat_options` (
   PRIMARY KEY (`ID`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
+CREATE TABLE `diagnosis_evolution` (
+  `DxEvolutionID` int(10) unsigned NOT NULL AUTO_INCREMENT,
+  `Name` varchar(255) DEFAULT NULL,
+  `ProjectID` int(10) unsigned NOT NULL,
+  `visitLabel` varchar(255) DEFAULT NULL,
```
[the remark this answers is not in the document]
zaliqarosli: in the end, the values come from session tables's visit_label column so opting to not add FK

SQL/0000-00-00-schema.sql
```diff
@@ -143,6 +143,19 @@ CREATE TABLE `caveat_options` (
...
+  `instrumentName` varchar(255) DEFAULT NULL,
```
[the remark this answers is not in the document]
zaliqarosli: add FK in last commit

modules/candidate_parameters/jsx/DiagnosisEvolution.js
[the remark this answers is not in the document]
zaliqarosli: i just checked and the order column is required, it doesn't determine the order of display and i think it's a good idea to display for clarity

zaliqarosli:
@driusan the SearchableDropdown being used is just the one that is in the jsx form elements file so its UI can be targeted in a separate if necessary. I added the FK for test name, ready for your review/testing again

PR #7571, opened 2021-09-18
[create_timepoint] Display only visit labels related to candidate project

AlexandraLivadas:
When creating a timepoint, the only visit labels displayed in the dropdown menu will be ones related to the candidate's project and the selected subproject using the `visit_project_subproject_rel` table. This feature is helpful for projects or instances of Loris with a long list of visit labels, so that only relevant visits are displayed, making the workflow easier for users. 

#### Testing instructions:
1. Add data to `project_subproject_rel` and `visit_project_subproject_rel` so that some project only has some of the available visit labels related to it.
2. Create a new timepoint for a candidate with the given project
3. See that it only displays relevant visit labels


ridz1208:
@AlexandraLivadas please rebase when possible

AlexandraLivadas:
@ridz1208 Rebased and ready for re-review!

modules/create_timepoint/php/timepoint.class.inc
```diff
@@ -90,9 +90,23 @@ class Timepoint extends \NDB_Page implements ETagCalculator
         );
 
         // Retrieve visit labels.
-        $visit_options = [];
+        $projectID = $candidate->getProjectID();
+        // Visits associated with the candidate's project
+        $visitsForProject = \Project::getVisitList($projectID);
```
ridz1208, line 95: why not use https://github.com/aces/Loris/blob/main/php/libraries/Utility.class.inc#L228 instead ?

(I agree the utility class function is less than ideal but an upcoming PR should be cleaning that up and the addition of another function in the project class at this stage might confuse things)

ridz1208 (changes requested):

ridz1208:
@AlexandraLivadas @driusan 

This was assigned to me so I figured I might as well fix the conflicts on it. While trying to make sense of the conflict I realized that the conflict is coming from this PR https://github.com/aces/Loris/pull/7825 which I now realized is conflicting with this one because it's doing exactly the same thing...

soooo can someone please confirm that it is indeed doing the same thing and in this case @AlexandraLivadas do you mind if we close your PR since the other one is merged (and sorry for only getting to this now)

AlexandraLivadas:
@ridz1208 I just confirmed that this is doing the exact same thing as the PR that you mentioned (listing visit labels according to both project and subproject). I am fine with closing this PR!

PR #7655, opened 2021-09-29
[webpack] Target directories for ESlint

ridz1208:
## Brief summary of changes
The reason for this is to prevent eslint from parsing and flagging every error in random JS files (including local directories) and only targetting LORIS files. 

For more context, when working on multiple projects, one way of mani[ulating project directories is to swap in and out different project directories. without this change, running make dev would fail every single time due to errors in JS files in old projects and unmaintained overrides.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -160,6 +160,7 @@ const config = [
     devtool: 'source-map',
     plugins: [
       new ESLintPlugin({
+        files: ['modules/', 'jsx/', 'htdocs/js'],
         cache: true,
       }),
       new CopyPlugin({


webpack.config.js
```diff
@@ -160,6 +160,7 @@ const config = [
     devtool: 'source-map',
     plugins: [
       new ESLintPlugin({
+        files: ['modules/', 'jsx/', 'htdocs/js'],
```
laemtl, line 163: files: ['modules/', 'jsx/', 'htdocs/js/', 'webpack.config.js'],

laemtl (changes requested):

webpack.config.js
laemtl: This change will implement the same set of files we are using for lining js.

laemtl (approved):

webpack.config.js
```diff
@@ -160,6 +160,7 @@ const config = [
     devtool: 'source-map',
     plugins: [
       new ESLintPlugin({
+        files: ['modules/', 'jsx/', 'htdocs/js/', 'webpack.config.js'],
```
maltheism, line 163: @ridz1208 you're forgetting about jslib

maltheism (changes requested):

Issue #7623, opened 2021-09-28
[User Accounts] Unable to create new user 500 Error returned


jstirling91:
**Describe the bug**
When creating a new user with all permissions, sites, and projects selected a 500 internal error is returned. 

**To Reproduce**
Steps to reproduce the behaviour (attach screenshots if applicable):
1. Login in as admin user
2. Go to User Accounts and select add user
3. Click  `Make user name match email address`
4. Click `Generate new password`
5. Enter valid First and Last Name
6. Enter and confirm valid email address
7. Select all Sites
8. Select all Projects
9. Select all Examiner At
10. Select Radiologist Yes Pending No
11. Select Pending Approval No
12. Select all Roles and Permissions

**What did you expect to happen?**
A new user to be created 

**Browser Environment (please complete the following information):**
- OS: Mac 11.5.2
- Browser Chrome
- Version 93.0.4577.82

**Server Environment (if known):**
- LORIS Version: 24.0.0
- Linux distribution and Version: on test vm
- MySQL/MariaDB Version: on test vm

**Additional context**
Example Payload:
```
NA_UserID: on
NA_Password: on
First_name: Jordan
Last_name: Stirling
Degree: 
Position_title: 
Institution: 
Department: 
Address: 
City: 
State: 
Country: 
Fax: 
Email: jordan.stirling@mcin.ca
SendEmail: on
__ConfirmEmail: jordan.stirling@mcin.ca
CenterIDs[]: 1
CenterIDs[]: 2
CenterIDs[]: 3
CenterIDs[]: 4
ProjectIDs[]: 1
ProjectIDs[]: 2
ProjectIDs[]: 3
ProjectIDs[]: 4
ex_1: on
ex_2: on
ex_3: on
ex_4: on
examiner_radiologist: Y
examiner_pending: N
Active: Y
active_from: 
active_to: 
Pending_approval: N
permID[1]: on
permID[2]: on
permID[3]: on
permID[4]: on
permID[5]: on
permID[6]: on
permID[8]: on
permID[9]: on
permID[10]: on
permID[11]: on
permID[12]: on
permID[13]: on
permID[17]: on
permID[18]: on
permID[19]: on
permID[22]: on
permID[23]: on
permID[24]: on
permID[25]: on
permID[28]: on
permID[29]: on
permID[31]: on
permID[32]: on
permID[33]: on
permID[34]: on
permID[35]: on
permID[36]: on
permID[37]: on
permID[38]: on
permID[39]: on
[...]

PR #7661, opened 2021-09-30
[User_accnout] can't create a new user

kongtiaowang:
## Brief summary of changes
Before the new user is not created, the Userid is '', so it can't insert into the "examiners" table.
remove the '' value from insert function.  

1. create a new user with an examiner and without an examiner.

[User Accounts] Unable to create new user 500 Error returned #7623


maltheism (approved):
Passed manual testing for me.

PR #7663, opened 2021-09-30
[Visits] Population tool for visit table

ridz1208:
## Brief summary of changes

The objective to remove the visits definitions from everywhere other than the `visit` table in the database can only be achieved by backpopulating the visit table from all other locations. In this case 3 sources are used:
 - config.xml
 - Visit_Windows table
 - session table

this tool queries the 3 sources above and inserts into the `visit` and the `visit_project_subproject_rel` the necessary values to make the transition to the table possible.


ridz1208:
@driusan part 1 as promised

tools/single_use/populate_visits.php
```diff
@@ -0,0 +1,167 @@
+<?php
```
ridz1208, line 1: after SQL

PR #7698, opened 2021-10-07
[issue_tracker] Fix incorrect url to issues in notification email

maltheism:
## Brief summary of changes

Currently, the notification email sent by the issue tracker module has the old url . This PR fixes this issue by refactoring the code from an ajax directory to issue_tracker/Edit.

1. Login to loris as admin account. Go to the Issue Tracker module and edit an issue. Select a user that has your email associated with the account in the 'Add others to watching?' field and click on 'Update Issue'. You will receive a 'Change to Issue' email notification. If you don't receive one, edit another field and 'Update Issue' again. Once you receive the email, click on the url in that email and notice that it ends with something like issue_tracker/issue/?issueID=78. This is the old way of accessing issues which will now just take you to the 'Add Issue' page. The url should be issue_tracker/issue/78.
2. On this PR branch, repeat the above step. The email you receive should have issue_tracker/issue/78 instead. Click on the link, and make sure it takes you to the correct issue.


CamilleBeau (changes requested):

CamilleBeau:
This PR looks great other than the one missing line change. I'm wondering though if the PR should be renamed because it seems that the main purpose of the PR is to refactor

modules/issue_tracker/php/edit.class.inc
```diff
@@ -0,0 +1,876 @@
...
+                        AND u.UserID <> :assignee",
+            [
+                'issueID'  => $issueID,
+                'uid'      => $user->getUsername(),
+                'assignee' => $changed_assignee,
+            ]
+        );
+
+        $msg_data['url']         = $baseurl .
+            "/issue_tracker/issue/?issueID=" . $issueID;
```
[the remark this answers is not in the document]
maltheism: Thanks @CamilleBeau

CamilleBeau (approved):

modules/issue_tracker/php/edit.class.inc
```diff
@@ -0,0 +1,876 @@
...
+     * Gets the changes to values, and the comments relevant to the given issue
+     *
+     * @param int $issueID the issueID
+     *
+     * @return array $commentHistory
+     */
+    function getComments(int $issueID): array
+    {
+        $factory = \NDB_Factory::singleton();
+        $db      = $factory->database();
```
driusan, line 298: this should come from the lorisinstance object

modules/issue_tracker/php/edit.class.inc
```diff
@@ -0,0 +1,876 @@
...
+     * Gets a list of those watching an issue.
+     *
+     * @param int $issueID the relevant issue
+     *
+     * @return array those who are watching
+     */
+    function getWatching(int $issueID): array
+    {
+        $factory = \NDB_Factory::singleton();
+        $db      = $factory->database();
```
driusan, line 350: this should come from the lorisinstance object

modules/issue_tracker/php/edit.class.inc
```diff
@@ -0,0 +1,876 @@
...
+    /**
+     * Returns the site name in the database corresponding to the centerID
+     *
+     * @param int $centerID - must match one from the psc table or 0
+     *
+     * @return ?string
+     */
+    function getSiteName(int $centerID): ?string
+    {
+        $db =& \Database::singleton();
```
driusan, line 373: I don't see why this should use the database instead of a Site object, since you have the $centerID to instantiate it.

modules/issue_tracker/php/edit.class.inc
```diff
@@ -0,0 +1,876 @@
...
+        $values = $request->getParsedBody();
+        if (!is_array($values)) {
+            return new \LORIS\Http\Response\JSON\BadRequest(
+                'Invalid request'
+            );
+        }
+
+        $factory = \NDB_Factory::singleton();
+        $db      = $factory->database();
+        $user    = $factory->user();
```
driusan, line 399: these should come from the request/lorisinstance object here too.

driusan (changes requested):
there are a lot of other places with the same problem, but all the NDB_Factory calls should be removed. The page's LorisInstance object should be used to get the database connection and the $request object should be used to get the user throughout.

CamilleBeau:
The code changes look good to me and all of the _GET / $factory references were updated, although @driusan should take another look because I am not as familiar with the current conventions. 

I manually tested again but found issue #7799 (unrelated to this PR) so could only test with admin / super user. Everything was working well!

CamilleBeau (approved):

modules/issue_tracker/php/edit.class.inc
```diff
@@ -0,0 +1,860 @@
...
+                    'DCC'      => $DCCID,
+                ]
+            );
+
+            $inactive_users_expanded = $db->pselect(
+                "SELECT DISTINCT u.Real_name, u.UserID FROM users u
+                 LEFT JOIN user_psc_rel upr ON (upr.UserID=u.ID)
+                 WHERE FIND_IN_SET(upr.CenterID,:CenterID) OR (upr.CenterID=:DCC)
+                 AND Active='N'",
+                []
```
driusan, line 100: This doesn't look like a valid query? There is nothing bound to the bind parameters.

modules/issue_tracker/php/edit.class.inc
```diff
@@ -0,0 +1,860 @@
...
+     * @param string $message error message to display
+     *
+     * @return void
+     */
+    function showError(string $message) : void
+    {
+        if (!isset($message)) {
+            $message = 'An unknown error occurred!';
+        }
+        header('HTTP/1.1 400 Bad Request');
```
driusan, line 814: These are side-effect inducing functions that aren't valid in an class that is supposed to be returning a PSR ResponseInterface.

driusan (dismissed):

modules/issue_tracker/php/edit.class.inc
maltheism: @driusan wow I'm something went wrong awhile back in the code because it's like that on main.

modules/issue_tracker/php/edit.class.inc
```diff
@@ -0,0 +1,859 @@
...
+                        )
+                ",
+                    [
+                        "center_id" => $result['centerID'],
+                        "psc_id"    => $result['PSCID'],
+                    ]
+                );
+            }
+            if (!$validCenter) {
+                error_log('PSCID and Center ID do not match a valid session!');
```
driusan, line 680: I don't see why this is going to the error log, if it needs to it should be using `$this->logger`

modules/issue_tracker/php/edit.class.inc
```diff
@@ -0,0 +1,859 @@
...
+                    'visitLabel' => $result['visit'],
+                ]
+            );
+
+            if (isset($session[0]['sessionID'])) {
+                $result['sessionID'] = $session[0]['sessionID'];
+                $result['candID']    = $session[0]['candID'];
+            } else {
+                error_log(
+                    'PSCID and Visit Label do not match a valid candidate session!'
```
driusan, line 703: I don't see why this is going to the error log, if it needs to it should be using $this->logger

modules/issue_tracker/php/edit.class.inc
```diff
@@ -0,0 +1,859 @@
...
+            if (!$user->hasPermission('access_all_profiles')) {
+                $params['CenterID'] = implode(',', $user->getCenterIDs());
+                $query .= " AND FIND_IN_SET(RegistrationCenterID,:CenterID)";
+            }
+
+            $candidate = $db->pSelectOne($query, $params);
+            if ($candidate) {
+                $result['candID'] = $candidate;
+            } else {
+                error_log('PSCID does not match a valid candidate!');
```
driusan, line 726: I don't see why this is going to the error log, if it needs to it should be using $this->logger

modules/issue_tracker/php/edit.class.inc
```diff
@@ -0,0 +1,859 @@
...
+            $issueValues['dateCreated'] = date('Y-m-d H:i:s');
+            $db->insert('issues', $issueValues);
+            $issueID = intval($db->getLastInsertId());
+        }
+
+        $this->updateHistory($historyValues, $issueID, $user);
+        $this->updateComments($values['comment'], $issueID, $user);
+
+        // Attachment for new issue.
+        if (isset($_FILES['file'])) {
```
driusan, line 460: There shouldn't be any references to superglobals. This should be using getUploadedFiles from the request: https://www.php-fig.org/psr/psr-7/ (can't directly hyperlink, but you can ctrl-f on the page for the behaviour of getUploadedFiles

driusan (dismissed):

modules/issue_tracker/php/uploadhelper.class.inc
```diff
@@ -156,10 +178,13 @@ class UploadHelper
             $DB->insert('issues_attachments', $values);
             // Move file from tmp to issue_tracker attachment directory.
             $path = $fileToUpload->full_path . '/' . $fileToUpload->file_hash;
-            move_uploaded_file($_FILES['file']['tmp_name'], $path);
+            move_uploaded_file($tmpFilePath, $path);
         } catch (\DatabaseException $e) {
             $this->endWithFailure();
-            error_log($e->getMessage());
+            $this->_logger->log(
```
driusan, line 184: I think this is the wrong logger and there's no obvious way to fix it. It's using the Request logger instead of the Exception one since you're passing the one from the page class.

Maybe leave this an `error_log` for now (or remove the try/catch if that's an option.)

driusan: actually you can just re-throw `$e` after the endWithFaliure call.

maltheism: @driusan updated

driusan:
I'm not convinced the logs are necessary and aren't going to be duplicated by the exception logger (and it would be cleaner to use `$this->logger->error` than `$this->logger->log(...)` but I'm not going to block over that

PR #7736, opened 2021-10-21
[mri_violations] Scrolling

### LINKED ISSUE #7203: [mri_violations] Scrolling
Issue originally reported by @pierre-p-s on CCNA Repo during candidate release testing of `LORIS 23.0.2`
aces/CCNA#4280
____
When using a filter, you cannot scroll through the table data (scroll to the right to see problem or resolution status). Same thing when using the arrow

maltheism:
## Brief summary of changes

The odd scrolling behaviour "bug" is from the jquery.dynamictable.js code.

Basically, the scrolling glitch happens when the static column "PatientName" is attempted to be always shown but the table width & browser width conflicts with all the other "widths" of the table columns.

This is a rare bug happening when the user's browser width including the table's columns widths are just right for the bug to happen. Furthermore, the module mri_violations hasn't been reactified and (I believe) only non-reactified tables attempt to show a static column like "PatientName" when scrolling throughout the table if the table width is too large.

My solution was to delete the code that attempts to always show the static column like "PatientName" and so the bug will never happen. The functionality that's rarely used anymore is lost but it solves a complicated bug.

See issue ticket for testing instructions.

* Resolves #7203

### DIFF
--- a/htdocs/js/jquery.dynamictable.js
+++ b/htdocs/js/jquery.dynamictable.js
@@ -221,9 +221,9 @@
         $(header).parent().find('.frozenHeader').is(':visible')
       ) {
         addFrozenHeaderColm(header);
-        return true;
+        return colmStatic;
       }
-      return false;
+      return colmStatic;
     }
     return colmStatic;
   };
@@ -302,7 +302,11 @@
           $(child1).addClass(id + 'FrozenColumn');
         });
         $(this).parent().scroll(function() {
-          colmStatic = freezeColm(id, colmStatic);
+          if (window.innerWidth <= document.getElementById(
+            'dynamictable'
+          ).offsetWidth) {
+            colmStatic = freezeColm(id, colmStatic);
+          }
         });
       }
       return this;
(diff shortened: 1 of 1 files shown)


driusan:
This looks like it's completely removing column freezing across all of LORIS, not fixing the mri violations?

maltheism:
@driusan column freezing doesn't seem to exist in Loris at all besides modules that haven't been reactified. Also see earlier comment I made about it.

maltheism:
@driusan @jesscall new fix that keeps the static column functionality but only uses it if the user's browser window is smaller than the table width.

driusan:
@jesscall can you re-test?

jesscall:
On page load, everything looks fine;

Screen Shot 2021-11-09 at 8 47 27 AM



When you scroll to the right side, the `PatientName` column is frozen (no glitching). However when scrolling back to the left, the column stays frozen and blocks the Project name.

Screen Shot 2021-11-09 at 8 47 10 AM


Not sure if this is an issue. The glitching however is resolved. I will wait to see what @driusan thinks before determining if this behaviour should pass.

driusan:
If it's an improvement over the current behaviour, I think it's fine to merge it.

jesscall:
@driusan yes it is an improvement.
In that case, looks good to me!

jesscall (approved):

PR #7739, opened 2021-10-26
[survey_accounts] Ensure Multiselect Value is Array

### LINKED ISSUE #7675: [survey_accounts] Can't Submit Medical History Survey
**Describe the bug**
After filling out the Medical History survey, clicking "Submit Data" causes a 500 error. 

**To Reproduce**
1. Go to the survey accounts module
2. Click a URL for a Medical History survey (the bug appears to happen regardless of whether the survey status is "In Progress" or "Created")
2. Fill out the survey 
3. On the last page, click the "Submit Data" button

I expected a page that had a message indicating the survey was submitted. 

**Browser Environment**
- Chrome 93.0.4577.82
- Firefox 92.0.1

**Additional context**
Bug reproducible on both firefox and chrome

jstirling91:
Fix issue where multiselect value is of type string instead of array when only one value is selected

## Brief summary of changes

- With the addition of support for PHP8, the function `in_array(string needle, array haystack)` expects the parameter `haystack` to be of type `array`. Unlink PHP7.4, PHP8 will throw an error instead of a warning. When mapping the multiselect values from the DB, we expected the special chars `{@}` to be present in the string and only exploded the string if it was. In the case when only one value is selected the chars `{@}` aren't present and so the string is never split into an array. This would then cause the `haystack` value to be of type `string` instead of `array`. This PR removes the check for the special chars `{@}` and just always calls the explode method no matter if the `{@}` is present or not. by doing this it will always force the string to be converted to an array.

1.

* Resolves #7675

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -566,9 +566,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         //Convert select multiple elements into a lorisform array
         if (!empty($this->selectMultipleElements)) {
             foreach ($this->selectMultipleElements AS $elname) {
-                if (isset($defaults[$elname])
-                    && stristr($defaults[$elname], "{@}")
-                ) {
+                if (isset($defaults[$elname])) {
                     $defaults[$elname] = explode("{@}", $defaults[$elname]);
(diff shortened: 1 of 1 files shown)


suzanne-lee:
LGTM, passes manual testing

PR #7754, opened 2021-10-28
[php/libraries/UserPermissions.class.inc] can't get longname

kongtiaowang:
## Brief summary of changes
if set one module as "No"  in the module_manager, user account module can't edit the user ( 500 error)
module[1] = y
module[2] = n
module[3] = y
when one user has module 2 permission, it will call module[2]->getLongName function. 

1. set one module as "No" in module_manager.
2. create a new user or edit a user without 500 error.

### DIFF
--- a/php/libraries/UserPermissions.class.inc
+++ b/php/libraries/UserPermissions.class.inc
@@ -317,7 +317,7 @@ class UserPermissions
         // Module Long Name: action description
         foreach ($results as $key=>$perm) {
             $newDesc = '';
-            if (!empty($perm['moduleID'])) {
+            if (!empty($perm['moduleID']) && !empty($modules[$perm['moduleID']])) {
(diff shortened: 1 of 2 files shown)


driusan:
why is the query including disabled modules? The line above the loop is `$modules = \Module::getActiveModulesIndexed($DB);` .. it seems like it would be better to update the query so that `$result` doesn't include disabled modules.

kongtiaowang:
@driusan The query of getActiveModulesIndexed is correct. The issue is that one user has the permissions for one module, and the module is not activated,  although getActiveModulesIndexed is correct , but  "$modules[$perm['moduleID']]" will be null.

driusan:
Yes, because the `$query` variable is incorrectly including inactive modules. Why not fix the query in the method being modified?

kongtiaowang:
@driusan yes, you are right. The query has been fixed. Could you review again?

PR #7756, opened 2021-10-29
[bvl_feedback] Only show icon if module enabled

### LINKED ISSUE #7521: Bvl Feedback visible/accessible when de-activated in Module Manager
The pencil icon that opens Bvl Feedback is still visible in the top menu bar (next to the  Help "?" icon), when the module has been set to Active:No in the Module Manager.  Clicking on the pencil icon will open Bvl Feedback.

To reproduce: set Active=No for bvl_feedback.  The pencil icon in the top menu bar is still Visible and clickable - opening Bvl feedback - when viewing the following: 
* Timepoint List
* Instrument list
* from any instrument form

driusan:
The bvl_feedback module can technically be disabled, but the pencil
icon for submitting feedback is added by middleware outside of the
module. Update the Middleware to not display bvl_feedback icon if
the module is disabled.

Resolves #7521


kongtiaowang (approved):
LGTM

PR #7758, opened 2021-10-29
[Module] Add return type hint to isPublicModule

### LINKED ISSUE #4282: Module function isPublicModule() should return a bool

driusan:
Add a return type hint to the isPublicModule method.

Resolves #4282

### DIFF
--- a/modules/api/php/module.class.inc
+++ b/modules/api/php/module.class.inc
@@ -35,7 +35,7 @@ class Module extends \Module
      *
      * @return bool
      */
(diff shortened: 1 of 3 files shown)


driusan:
@kongtiaowang can you review/test this?

Issue #7093, opened 2020-10-13
[candidate_list] Remove "Latest Visit Status" column?


driusan:
From what I can see in the code, the column "Latest Visit Status" on the candidate_list contains the result of `max(s.Current_stage) AS LatestVisitStatus` (grouped by CandID). This means that the column contains the highest (alphabetically) of the set `'Not Started', 'Screening', 'Visit', 'Approval', 'Subject', 'Recycling Bin'` (the valid options of Current_stage in the session table) that the candidate has for any session. ie. if a candidate has sessions with values of "Not Started", then "Visit", then a last visit of "Approval" the value of "Latest Visit Status" will be "Visit".

It's not clear how that is supposed to be meaningful or what relation it has to the "Latest Visit Status". It seems to be the result of someone not understanding how "max" works on enums when writing the query.

Given that it doesn't work and no one has noticed, it seems like the column isn't really used and should be removed to free up space in an already excessively large table.

zaliqarosli:
duplicate https://github.com/aces/Loris/issues/5115

driusan:
Resolved by #7759

PR #7759, opened 2021-10-29
[candidate_list] Remove meaningless Latest_Visit_Status column

### LINKED ISSUE #5115: [Candidate_list] Latest Visit Status value is ambiguous
In the candidate list query, `max(s.Current_stage) AS LatestVisitStatus` will return the the highest value amongst all session Current_stage for the candidate. The highest value is calculated alphabetically. 

(enum('Not Started','Screening','Visit','Approval','Subject','Recycling Bin'))

So as long as any session Current_stage is 'Visit', the  LatestVisitStatus will always be 'Visit'.

driusan:
The max alphabetical visit doesn't mean anything. (See resolved issues for explanation.)

Resolves #5115, #7093

### DIFF
--- a/modules/candidate_list/jsx/candidateListIndex.js
+++ b/modules/candidate_list/jsx/candidateListIndex.js
@@ -307,22 +307,6 @@ class CandidateListIndex extends Component {
         },
       },
       {
-        'label': 'Latest Visit Status',
-        'show': true,
-        'filter': {
-          name: 'latestVisitStatus',
-          type: 'select',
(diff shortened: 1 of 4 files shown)


driusan:
Not sure if this should target v24 or be kept in draft mode until after the release since we've already tested it with the meaningless behaviour.

driusan:
@xlecours since you filed #5115 can you review/test this?

maltheism (approved):
@driusan passed manual testing

PR #7761, opened 2021-10-29
[bvl_feedback] Remove redundant bvlFeedbackPossible method

### LINKED ISSUE #7665: [bvl_feedback] Module cannot be accessed on instrument page
The "pencil" icon to open the behavioural feedback module does not appear on the sidebar of an instrument form, but it does appear on a candidate's profile page or on the timepoint page. 

**To Reproduce**
1. Go to a candidate's profile (one with pre-existing visits)
2. See that the pencil icon is in the top right corner and that it opens bvl_feedback
3. Click on a timepoint and then select an instrument
4. See that when on an instrument page, there is no pencil icon or option to open bvl_feedback

The bvl_feedback module should be accessible from an instrument form page

driusan:
The middleware makes a call to NDB_BVL_Feedback::bvlFeedbackPossible
to determine whether the feedback icon should be displayed. This
function checks the test_name (based on the page passed by the middleware) and user permission to filter out who should not see the
icon. However, permission is already checked by the middleware, and
the test_name check breaks the loading of the icon for instruments
since the page passed by the middleware is the NDB_BVL_Instrument
(sub)class, not the "instruments" module.

The middleware checks for the existence of a "getFeedbackPanel"
method on the page (which is more robust, and exists on the instrument class),
so the test name check is unnecessary and only breaks instrument feedback.
The result is that the template variable can just be set to "true" if we're in
the if statement without loss of functionality, and the static method
(which is unused anywhere else) can be removed.

Resolves #7665

### DIFF
--- a/php/libraries/NDB_BVL_Feedback.class.inc
+++ b/php/libraries/NDB_BVL_Feedback.class.inc
@@ -1049,34 +1049,4 @@ class NDB_BVL_Feedback
         $record = $db->pselectOne($query, $param);
         return $record;
     }
-
-    /**
-     * This function is currently used to check whether or not to render
-     * the bvl_feedback_icon in main.tpl. It checks if the user has the
-     * bvl feedback permission first, if the user has the permissions it
-     * then checks if the user is on the timepoint_list, instrument_list
-     * or on any instrument page.
-     *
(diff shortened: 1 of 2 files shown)


AlexandraLivadas (approved):
Tested and it works great!

PR #7762, opened 2021-10-29
[next_stage] Updated test plan for clarity

### LINKED ISSUE #7641: [Next Stage] Missing instrument in timepoint using Age Min/Max in battery
**Describe the bug**
When starting a new stage, instrument is missing when the candidate falls age at visit falls between Min and Max age. See [test plan](https://github.com/aces/Loris/blob/main/modules/next_stage/test/TestPlan.md) step 6

**To Reproduce**
1. Create new test in Battery Manager setting the Min and Max age and subproject. Leave the visit label blank.
2. Create a new candidate
3. For new candidate, create a new timepoint, selecting the same subproject you used above.
4. Go to new timepoint and click Start Visit Stage.
5. Select a date which the candidates age will fall between the Min/Max age used above. Select yes to scan done
6. Click Start Visit and navigate to Visit page

The instrument used in step 1 above should show up in instrument list.

- OS: Mac
- Browser Chrome

- LORIS Version: 24.0.0

**Additional context**

AlexandraLivadas:
Updated Step 6 of the test plan to clarify that the instruments of the test_battery entry with a NULL visit label and that falls within the candidate's age should only be displayed for a timepoint if there does **not** exist other test_battery entries where the visit label is set to the visit of the timepoint. See [this section](https://github.com/aces/Loris/blob/main/modules/battery_manager/help/battery_manager.md#visitlabel) of the module help text for the behaviour expected. 

* Resolves #7641

### DIFF
--- a/modules/next_stage/test/TestPlan.md
+++ b/modules/next_stage/test/TestPlan.md
@@ -13,8 +13,10 @@
    [Manual Testing]
 6. Create a timepoint where the age (date entered minus DoB) falls within the
    AgeMin and AgeMax for a `test_battery`. The `test_battery` needs to have a 
-   matching `subprojectID` and a NULL `Visit_label`. Ensure that those 
-   instruments are inserted.
+   matching `subprojectID` and a NULL `Visit_label`. There should be no 
+   other `test_battery` entries with the `Visit_label` set to the visit label 
+   of the timepoint. Ensure that the instruments of the `test_battery` with the 
+   NULL `Visit_label` are inserted.
    [Manual Testing]
 7. Create a timepoint where there is a `test_battery` entry for the
    `Visit_label` and `subprojectID`, and start the stage. Ensure that those 


driusan:
@ridz1208 @jstirling91 can you review?

ridz1208:
I will differ to @jstirling91 since he's the one for whom this was not clear. @zaliqarosli your opinion is always welcome

zaliqarosli:
@AlexandraLivadas what happens if there is an entry in the test_battery for that instrument where the Visit_label field is set to the timepoint's visit label?

zaliqarosli (approved):

AlexandraLivadas:
@zaliqarosli So there are essentially 2 cases here:
1. There are no entries in the test_battery with the timepoint's visit label. If this is the case and the candidate falls within the age range of a test_battery entry (with a NULL visit label), then the instrument should appear
2. There are entries in the test_battery with the timepoint's visit label. In this case, these instruments are the ONLY ones that should appear, so any instrument with a test_battery entry with a null visit label should not appear.

In order to test this, I first removed any test_battery entry with a visit label set to for example V3. Then I created a test_battery entry for the BMI instrument and made sure the visit label was null. I made a candidate timepoint with visit label V3 that fell within the age range of the test_battery entry and made sure that the BMI instrument appeared. Then I made another test_battery entry with the visit label set to V3 for the MRI instrument. I made another candidate timepoint with visit label V3 and saw that now the MRI instrument appears but not the BMI, as expected.

Let me know if this makes sense!

PR #7763, opened 2021-10-29
[BVL_feedback] fix icon display issue

kongtiaowang:
## Brief summary of changes
if the bvl_feedback module is not active, then the icon should disappear.
set bvl_feedback module as not activated then check the "bvl" icon on the top menu.

https://github.com/aces/Loris/issues/7521


kongtiaowang:
replace by https://github.com/aces/Loris/issues/7521

Issue #7642, opened 2021-09-28
[User Accounts] Page loads with error when site's only permission set


jstirling91:
**Describe the bug**
When given just the permission `User Accounts: View/Create/Edit User Accounts - Own Sites`, when loading the page you get the error `An error occurred while loading the page.`.

**To Reproduce**
Steps to reproduce the behaviour (attach screenshots if applicable):
1. Create a new user and assign them the `User Accounts: View/Create/Edit User Accounts - Own Sites` permission
2. Login as new users and go to User Accounts page

**What did you expect to happen?**
The page should load with all users's at you given site.

**Browser Environment (please complete the following information):**
- OS: Mac
- Browser Chrome

**Server Environment (if known):**
_Note: We only support the most recent release of LORIS._
- LORIS Version: 24.0.0
- Linux distribution and Version: [e.g. Ubuntu 16.04, CentOS 7]
- MySQL/MariaDB Version: [e.g. MySQL 5.7, MariaDB 10.3]

**Additional context**

PR #7764, opened 2021-11-01
[user_account] Fix error when loading the page in the case the user have site's only permission set.

racostas:

1. Create a new user and assign them the `User Accounts: View/Create/Edit User Accounts - Own Sites` permission
2. Login as new users and go to User Accounts page
3. The page should load with all users at you given site.

* Resolves https://github.com/aces/Loris/issues/7642

### DIFF
--- a/src/Data/Filters/UserSiteMatch.php
+++ b/src/Data/Filters/UserSiteMatch.php
@@ -45,7 +45,7 @@ class UserSiteMatch implements \LORIS\Data\Filter
             // If the Resource belongs to multiple CenterIDs, the user can
             // access the data if the user is part of any of those centers.
             foreach ($resource->getCenterIDs() as $site) {
-                if ($user->hasCenter($site)) {
+                if ($user->hasCenter(new \CenterID($site))) {
                        return true;
                 }
             }


src/Data/Filters/UserSiteMatch.php
```diff
@@ -45,7 +45,8 @@ public function filter(\User $user, \Loris\Data\DataInstance $resource) : bool
             // If the Resource belongs to multiple CenterIDs, the user can
             // access the data if the user is part of any of those centers.
             foreach ($resource->getCenterIDs() as $site) {
-                if ($user->hasCenter($site)) {
+                $center = new \CenterID($site);
```
driusan, line 48: This isn't correct. It's inside of a `instanceof \LORIS\StdyEntities\MultiSiteHaver` check. MultiSiteHaver is defined as

```
interface MultiSiteHaver
{
    /**
     * Return true if the entity is accessible by the
     * user.
     *
     * @return \CenterID[]
     */
    public function getCenterIDs() : iterable;
}
```

`$site` in the foreach loop should already be a centerid

driusan (changes requested):

src/Data/Filters/UserSiteMatch.php
racostas: Hi @driusan The issue here seem to be variables types. `hasCenter()` is expecting and object of type `Center.` while **$resource->getCenterIDs()** is not returning it.

driusan: What is the `$resource` that is not returning it and using the filter? The docblock for the interface is explicit that it's supposed to be returning CenterIDs (the only reason the php return type isn't is because php doesn't directly support that annotation.)

racostas: According to the test done I would said `$resource` is not returning an object of type `CenterID`. When "manually" typecasted as in this PR it does take it.

racostas: This is what shown at the error log before the fix. 
image

driusan: What is the class type of `$resource`? The interface that it's known to be within the if statement is guaranteed to return CenterIDs according to the interface contract. If `getCenterIDs()` from something that is defined as a MultiSiteHaver is returning something other than `CenterID[]` it's a bug with that class, not places that use the interface.

racostas: It seems there is not a "easy" way of solving this. The main issue seems to be due the PHP typeless default design. It seems we cannot have an array of Objects (at least not a classic array).

This post explains a bit more (dated for a couple of years but still current it seems).
https://stackoverflow.com/questions/40693469/function-return-type-hinting-for-an-array-of-objects-in-php7/40693938

We could try to make ourselves the class "array ofCenterIDs". I think we don't' have something similar in LORIS so far (a custom class array) and not sure the possible implications: it seems PHP array functions like` in_array()` will not work (at least not in a classic way again).

I see this kind change more as an architectural one, not sure if convenient/possible before the next release in a couple of weeks. Open to suggestions we could also discuss it  a bit more during the LORIS meeting.

racostas:
Not sure how to solve this test errors. Manual test is fine.

src/Data/Filters/UserSiteMatch.php
driusan: php can not, but phan (and most tooling) can follow the docblock typing when it's properly defined in a comment such as:

```
     * @return \CenterID[]
```

This should not involve changing the architecture or any code outside of user accounts, user accounts just needs to return a CenterID when it says it's going to.

See #7814 and #7815 for similar fixes to the issue_tracker and battery_manager

racostas: I have been trying to implement the solution given in #7815, but somehow I'm getting the same (similar issues). I think I'm missing something. Sorry, seems will need help with this one.

racostas:
I have been trying to implement the solution given in #7815, but somehow I'm getting the same (similar issues). I think I'm missing something. Sorry, seems will need help with this one.

kongtiaowang:
phpcs issue 
------ ---------------------------------------------------- 
  Line   php/libraries/CenterIDs.php                         
 ------ ---------------------------------------------------- 
  43     Call to an undefined method Traversable::append().  
 ------ ----------------------------------------------------

driusan:
Replaced by #7850

PR #7765, opened 2021-11-01
[Tests] Replace webDriver->findElement with safeFindElement

driusan:
This replaces all remaining instances of $this->webDriver->findElement
with $this->safeFindElement. This is unlikely to completely fix the
random test failures, but should help with some that still have
problems.


driusan:
@kongtiaowang can you look at this? I don't think it should interfere with #7753, they're not touching the same thing. This one completely gets rid of any missed references to $this->webDriver->findElement.

kongtiaowang (approved):
LGTM

PR #7766, opened 2021-11-01
[test] Add safeTextContent method and use it

driusan:
One of my PRs randomly failed on mri_violations tests. The
problem appears to be that code is using javascript->execute
script to get the text content of some elements. However, if
an action was just triggered by javascript, the element may
not yet have appeared on the page. This

1. Extracts the wait code from safeFindElement to a waitForElement
   wrapper, to reduce boiler plate when a test needs to wait
   for an element
2. Adds a safeTextContent helper, which will wait for an element
   to appear before running executescript on it.
3. Updates various tests (ones that I found by either grepping the
   code for "sleep" or were randomly failing on other PRs) to use
   these methods.
4. Changes some execute scripts which just do a "click" to use
   the analogous safeClick method (which already exists).

This is not done throughout all tests yet, but hopefully fixes some
of the problematic ones (if it works).

### DIFF
--- a/modules/candidate_list/test/candidate_listTest.php
+++ b/modules/candidate_list/test/candidate_listTest.php
@@ -352,7 +352,6 @@ class CandidateListTestIntegrationTest extends LorisIntegrationTestWithCandidate
     {
         $this->safeGet($this->url . "/candidate_list/");
         $link = self::$pscidLink;
-        sleep(1);
         $this->safeClick(WebDriverBy::cssSelector($link));
         $bodyText = $this->safeFindElement(
             WebDriverBy::cssSelector("body")
--- a/modules/create_timepoint/test/create_timepointTest.php
+++ b/modules/create_timepoint/test/create_timepointTest.php
@@ -12,7 +12,6 @@
  * @link     https://github.com/aces/Loris
  */
 use Facebook\WebDriver\WebDriverBy;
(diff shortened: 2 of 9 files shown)


modules/candidate_list/test/candidate_listTest.php
```diff
@@ -352,7 +352,6 @@ function testPscidLink()
     {
         $this->safeGet($this->url . "/candidate_list/");
         $link = self::$pscidLink;
-        sleep(1);
```
driusan, line 355: This was already doing a safeClick and using safeFindElement, so shouldn't be necessary.

modules/create_timepoint/test/create_timepointTest.php
```diff
@@ -78,33 +77,6 @@ function testCreateTimepoint()
...
-    /**
-     * Create a timepoint with three parameters.
-     *
-     * @param string $canID      ID of candidate
-     * @param string $subproject text of subproject
-     * @param string $visitlabel text of visit label
-     *
-     * @return void
-     */
-    private function _createTimepoint($canID, $subproject, $visitlabel)
```
driusan, line 90: This was a private method which wasn't called anywhere in the code (found when grepping for sleep)

driusan:
The random failures are no more, this might be an improvement but seems to be unnecessary

PR #7767, opened 2021-11-02
[Test] Integration test random failure fix test

xlecours:
## Brief summary of changes

I have noticed that each random failure was showing the dashboard page instead of the one asked by safeGet. I figured that the login step was causing somewhat causing the issue. Looking at the login function, it calls click on a button which loads the dashboard page. But is safeGet is called too early than the old body is the login page instead of the dashboard page. When the login page is finally detached from the DOM, safeGet returns the dashboard page instead.


xlecours:
all passed once.
restarting

driusan:
Passed a second time. Restarted. Let's go for 3!

xlecours:
Passed trice

driusan (approved):

PR #7772, opened 2021-11-03
[candidate_parameters] Allow clearing of consent status

### LINKED ISSUE #4149: [ConsentStatus] Add explicit button/action to clear consent
- [ConsentStatus.jsx] Consent status select element emptyOption=false if registered consent=yes/no
- 'Clear' button that calls separate action allowing the saving of all empty values to database when needing to erase wrong data (currently, formHandlers.php validates submitted data so that empty values cannot  be submitted)
- This clear button would issue a swal requiring confirmation 'Proceed' or 'Dismiss'

Needed as follow-up to https://github.com/aces/Loris/pull/4034

CamilleBeau:
## Brief summary of changes
This PR allows users to set a consent status back to null. I chose not to do this with a "Clear" button because a clear button would mean you are clearing all or none of the consent statuses on the page, or otherwise we would need a clear button for each individual consent. Instead, the participant can set the status to be empty and update the consent as usual. The date and withdrawal date are cleared and disabled when the status is set back to empty. 

1. Check that you are successfully able to update a consent status from non-null to empty
2. Check that when you update the status to empty, it only updates empty values for that consent and not for any other consent that was already empty
3. Make sure that the consent status history is updated appropriately
3. Check that all of the validation for consent still persists:
- The status should be required if a date is entered
- a date should be required if the status is entered
- a withdrawal date should be required if the consent status is changed from yes to no only

* Resolves #4149

### DIFF
--- a/modules/candidate_parameters/ajax/formHandler.php
+++ b/modules/candidate_parameters/ajax/formHandler.php
@@ -450,6 +450,7 @@ function editConsentStatusFields(\Database $db)
         // Validate data
         $recordExists  = array_key_exists($consentID, $candidateConsent);
         $oldStatus     = $candidateConsent[$consentID]['Status'] ?? null;
+        $oldDate       = $candidateConsent[$consentID]['DateGiven'] ?? null;
         $oldWithdrawal = $candidateConsent[$consentID]['DateWithdrawn'] ?? null;
         $validated     = false;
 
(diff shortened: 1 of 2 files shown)


maltheism:
Hi @CamilleBeau the consent fields revert back to the previous values after clicking the "Update" button.
Example:
Screenshot 2022-10-17 at 12 15 19 PM

then selecting empty field:
Screenshot 2022-10-17 at 12 15 30 PM

then after clicking the Update button:
Screenshot 2022-10-17 at 12 15 42 PM

I might be mistaken but I don't think that's intended?

CamilleBeau:
@maltheism Rebased!

I tested this again on my VM and I was not able to replicate the problem you mentioned:

First values:
image

Clearing value & pressing submit:
image

It stayed clear after refreshing as well. Would you be able to re-test it now that I've rebased?

CamilleBeau:
Does anyone have insight on the test suite failures here?

cmadjar:
@kongtiaowang can you help @CamilleBeau resolves the test failures please? Thanks!

sruthymathew123 (approved):
Tested and looks good

PR #7776, opened 2021-11-04
[Instruments] Escaping issue sub-optimal - Redirect on save success to load defaults from database

### LINKED ISSUE #7489: [LorisForm] HTML Text Element does not display quotes

ridz1208:
## Brief summary of changes
This PR fixes the escaping issue that occurs when a text field in an instrument contains an HTML special character. The solution employed here is to redirect on successful save to reload data directly from the database instead of loading it using the _POST data.

Cons:
 - This solution is **sub-optimal** because the escaping issue will still occur when an error is detected on instrument save and the values MUST be reloaded from the _POST array where the reload can not occur to avoid losing the unsaved data.

Pros:
 - relatively minor changes
 - easy to test

Alternate to: #7777
Fixes #7489
Replaces #7490

4. check out this PR

Make sure to test other usecases as well
1. behaviour when the form contains an error
2. behaviour when there are multiple `"`,`<`,`>` or `&` in the string
3. behaviour when the instrument status sidebar is modified (is it affected by the reload)

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -2881,9 +2881,10 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
-        $req = $request->getParsedBody();
+        $req     = $request->getParsedBody();
+        $success = false;
         if ($request->getMethod() === "POST" && !isset($req['ClearInstrument'])) {
-            $this->save();
+            $success = $this->save();
         }
 
         // Disable form if data entry is complete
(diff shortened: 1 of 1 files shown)


ridz1208:
@driusan on second thought, I don't see why this would need a changelog update... the other fix does, not this one

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -2913,6 +2914,21 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+            $baseURL   = \NDB_Factory::singleton()->settings()->getBaseURL();
+            $pageURL   = !empty($this->page) ? urlencode($this->page) ."/" : "";
+            $url       = $baseURL . "/instruments/" .
+                urlencode($this->testName) . "/" .
+                $pageURL .
+                '?commentID=' . urlencode($this->getCommentID()) .
+                '&candID=' . urlencode($candID) .
+                '&sessionID=' . urlencode($sessionID);
+            return (new \LORIS\Http\Response())
+                ->withHeader("Location", $url);
```
driusan, line 2929: The response should also have a 303 See Other response code.

ridz1208: what other response code ?

driusan: No other one. The code's name is `See Other`.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303

ridz1208: line 2846 above in the same file has the exact same code, should we move this to a function ?

ridz1208:
@driusan urs again

php/libraries/NDB_BVL_Instrument.class.inc
driusan: 2846 already had the 303? I don't see why it would be moved to a function.. setting the response code is always going to be context specific

ridz1208: dont like code duplication

CamilleBeau:
Saved properly with all cases except when there was a validation error as you mentioned. Coupled with a validation error, the following string ended up partly displaying outside of the textbox:

image
image

ridz1208:
@CamilleBeau please add passed manual test

CamilleBeau (approved):

PR #7781, opened 2021-11-04
Access Profile (Beta)] Minimizing boxes & inlcusion of 'Action' Items on 'Candidate's dashboard'

### LINKED ISSUE #7536: [Access Profile (Beta)] Minimizing boxes & inlcusion of 'Action' Items on 'Candidate's dashboard'
When accessing a candidate via the 'Access Profile (Beta)' route, the following improvements can be implemented to the 'Candidate's dashboard':

- When trying to minimize to hide a given card or box, the content will be hidden, but the card/box will not minimize.

Screen Shot 2021-08-09 at 4 04 13 PM

- Missing the 'Actions' button for 'Create Timepoint', 'Candidate Info' and 'View Imaging Datasets'. The latter can be accessed via the images 'card', but the other two actions can't and the user has to navigate to the standard 'Candidate's Profile' to do this, which is impractical.

maltheism:
## Brief summary of changes

See issue: https://github.com/aces/Loris/issues/7536

1. Hide/Show panel in new access profile beta

* Resolves #7536

### DIFF
--- a/jsx/CSSGrid.js
+++ b/jsx/CSSGrid.js
@@ -106,10 +106,6 @@ function CSSGrid(props) {
         if (value.Order) {
             style.order = value.Order;
         }
-        console.log('style is ');
-        console.log(style);
-        console.log('cardSize is ');
-        console.log(pSize);
 
         style.alignSelf = 'stretch';
         return (
--- a/jsx/CSSGrid.js
+++ b/jsx/CSSGrid.js
@@ -106,11 +106,16 @@ function CSSGrid(props) {
         if (value.Order) {
             style.order = value.Order;
         }
+        console.log('style is ');
+        console.log(style);
+        console.log('cardSize is ');
+        console.log(pSize);
 
         style.alignSelf = 'stretch';
         return (
             <Card title={value.Title} id={cardID} key={cardID} style={style}
-                cardSize={pSize}>
+                cardSize={pSize}
+            >
             {value.Content}
             </Card>
         );
--- a/jsx/Card.js
+++ b/jsx/Card.js
@@ -52,9 +52,6 @@ class Card extends Component {
     if (this.props.style) {
         divStyling = {...divStyling, ...this.props.style};
     }
-    if (this.props.cardSize) {
-        divStyling.height = this.props.cardSize;
-    }
     return (
       <div style={divStyling}>
         <Panel
--- a/jsx/Panel.js
+++ b/jsx/Panel.js
@@ -66,7 +66,6 @@ class Panel extends Component {
       <div
         className="panel-heading"
         onClick={this.toggleCollapsed}
-        data-toggle="collapse"
         data-target={'#' + this.props.id}
         data-parent={this.props.parentId ?
           '#'+this.props.parentId :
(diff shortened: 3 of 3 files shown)


jstirling91:
Seems that although the boxes now collapse, there's is still a lot of white space between boxes. Any chance when a box collapses above, the one below moves right below it? 
Screen Shot 2021-11-11 at 10 54 09 AM

maltheism:
@jstirling91 I'm somewhat confused because I fixed that issue when I was working on the code. Could you try clearing your cache?

jstirling91:
@maltheism I tried that, in an incognito and in a different browser and still seeing the same issue.

maltheism:
Hi @jstirling91 it turns out the shrinking depends on the grid's row. If there's another element to the left or right that's visible then the row below will be effected.
Example
Screen Shot 2021-11-11 at 1 04 04 PM
and
Screen Shot 2021-11-11 at 1 07 00 PM

I'm unsure if I can solve this when each panel's size is dynamic with depending on user's resolution while viewing the module.

jstirling91:
@maltheism it should be possible to do this. Have a look at the main dashboard, as it works like there.

maltheism:
@jstirling91 you're right I think changing the layout of this module to be similar to the dashboard will work.

maltheism:
@ridz1208 I'm working on getting rid of smarty to make things easier for this PR.

driusan:
Making the candidate_profile similar to the dashboard would be a step backwards. The dashboard is a hardcoded 2 column layout generated from html. The candidate_profile is a CSS grid loaded from javascript.

Smarty has nothing to do with the layout in this PR. All smarty is used for in the candidate_profile is loading the javascript.

maltheism:
@driusan I mean I'm converting form_candidate_profile.tpl to react to make things easier going forward for solving the issue. Currently the CSSGrid prevents a panel (that's under a panel) from moving upwards if the panel above shrinks but has a panel next to it with some height taking up space in the grid. I'm thinking of generating the panels in a layout that's similar to the dashboard because the dashboard is divided into two columns and which would make it possible for panels to not have this issue. I can write code that will loop through the panels and adds them to column 1 or column 2.

driusan:
Do not change the architecture of the module. It's a dynamically generated grid of arbitrary size, not a hardcoded 2 column layout.

maltheism:
@driusan how about dynamically generated columns? Otherwise I think the rows in CSS grid prevent fixing the panels from going upwards if a panel above is minimized.

maltheism:
Discussion happened on slack and one proposal is to remove the collapsable functionality from the panels. The idea of the module is to display a grid of cards being widgets. Anyone object to this proposal?

driusan:
I don't see what purpose the panels being collapsable serves on a dashboard where users are going to scroll to whatever they're interested in and then probably click something to navigate away (or already have scrolled to what they're interested in seeing) (especially since I don't think the collapsing is documented anywhere user facing), I think removing the collapsability makes sense.

maltheism:
@driusan closing this PR in favour of https://github.com/aces/Loris/pull/7840

Issue #7746, opened 2021-10-28
[User Accounts] No error feedback when creating/modifying a user


racostas:
**Describe the bug**
The page shows the message "The form you submitted contains data entry errors", but don't specify which error is happening.

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to User Accounts
2. Select Add/Modify User 
3. File out the form with new username and other required information
4. Please refer to the attached image to see which specify options were used (note this is not and extensive list)
5. Save the form and see the general error message but not and specific one.

![Screenshot Testing 24 - User Accounts - Edit User](https://user-images.githubusercontent.com/37309344/139267395-d80bf687-8348-47f4-94c4-2ac90775bd87.png)

**What did you expect to happen?**
A clear indication of the error that is happening.

**Browser Environment (please complete the following information):**
- OS: Microsoft Windows 10. 
- Browser: Chrome, Version 95.0.4638.54 (Official Build) (64-bit)

**Server Environment (if known):**
_Note: We only support the most recent release of LORIS._
LORIS testing server.

PR #7274, opened 2021-01-04
[User accounts] Fix edit user layout

laemtl:
* Resolves #7187

PR #7789, opened 2021-11-05
[user account] fix can't show email errors

kongtiaowang:
## Brief summary of changes
fix the variable name
input an invalid email to create or edit a user, then the error shows on the page.
https://github.com/aces/Loris/issues/7746

Screen Shot 2021-11-05 at 11 44 09 AM
)

### DIFF
--- a/modules/user_accounts/templates/form_edit_user.tpl
+++ b/modules/user_accounts/templates/form_edit_user.tpl
@@ -1,5 +1,6 @@
 <br />
 <form method="post" name="edit_user" autocomplete="off">
+    {if $form.errors}
     <div class="alert alert-danger" role="alert">
         The form you submitted contains data entry errors
     </div>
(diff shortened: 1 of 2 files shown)


modules/user_accounts/templates/form_edit_user.tpl
```diff
@@ -266,9 +265,11 @@
     	</div>
     </div>
     {/if}
-
-    {if $form.errors.Email|default}
+    {if $form.errors.Email_Group|default}
```
ridz1208, line 268: I don't think this is the correct solution, you are just undoing work from #7274 I think what should be done instead is to replace email_group in the edit_user class by "email" in the validate function.

ridz1208 (changes requested):

ridz1208:
@kongtiaowang have you checked that other fields in that same form haven't changed names as well ?

Read the issue (#7746) carefully there are more details there

kongtiaowang:
@ridz, could you review again?

ridz1208 (approved):
Code looks good

ridz1208:
@racostas could you test please ?

racostas:
LGTM

... a related issue:

When a user request an account by itself (not created by the admin) the validations in the email field are not triggered. We should do this consistent.

image

image

ridz1208:
@racostas please make a separate issue for what you have found

PR #7451, opened 2021-05-14
[User Accounts] Permissions panel fix

AlexandraLivadas:
The permissions panel on the edit user page is not correctly displaying Permission descriptions for user's with superuser permission. This is because of a superuser permissions check which returned a list of permissions without a `label` parameter specified. 

View before:
Screen Shot 2021-05-14 at 2 49 21 PM

View after:
Screen Shot 2021-05-14 at 2 48 43 PM

#### Testing instructions (if applicable)
1. Go to edit a user as a superuser and check that the permissions are correctly displayed with their descriptions
2. Verify that the permissions are also displayed correctly for a user who is not a superuser

* Resolves #7437 
* Related to #6770

PR #7796, opened 2021-11-05
[user_accounts] Very strange organization of permissions on user_accounts page

### LINKED ISSUE #7791: [user_accounts] Very strange organization of permissions on user_accounts page
**Describe the bug**
While approving a new user, this is the list of permissions I get (see below)
Screen Shot 2021-11-05 at 1 32 26 PM

maltheism:
## Brief summary of changes

Alphabetically sort by labels of the permissions.

* Resolves #7791

Preview before:
Screen Shot 2021-11-05 at 3 33 30 PM

Preview after:
Screen Shot 2021-11-10 at 4 10 21 PM


ridz1208:
I still see the duplication of roles/permissions in blue. it should look more like this:

Screen Shot 2021-11-05 at 3 43 33 PM

maltheism:
@ridz1208 do you want the alphabetical arrangement or just removing duplicate roles?

ridz1208:
@maltheism alphabeticall arrangement of the perissions and removing duplication of categories (so basically alphabetical sorting IN each category)

maltheism:
Hi @ridz1208 the PR is ready again for your eyes.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -909,40 +909,68 @@ class Edit_User extends \NDB_Form
...
-        $this->addGroup($group, 'PermID_Group', 'Permissions', "");
-        unset($group);
+
+        $collection   = [];
+        $collection[] = $this->form->createElement(
+            'static',
+            '',
+            '',
+            '</div>'
+                        . "<h3 id=\"header_Permission\" "
```
ridz1208, line 947: These fields you are hardcoding here are actually dynamic values read from the permission_category table in the dtabase

See RB example https://github.com/aces/Loris/blob/main/raisinbread/RB_files/RB_permissions_category.sql

so that's not a valid change

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -909,40 +909,68 @@ class Edit_User extends \NDB_Form
...
-                    '',
-                    '</div>'
-                    . "<h3 id=\"header_$lastRole\" "
-                    . "class=\"perm_header button\" "
-                    . "style=\"text-align: center; margin-top: 5px;\">"
-                    .ucwords($row['type'])
-                    . '</h3>'
-                    . "<div id=\"perms_$lastRole\" style=\"margin-top: 5px;\">"
-                );
+        usort(
```
ridz1208, line 912: the sort could just as easily (if not easier be done on the query coming from SQL)

ridz1208 (dismissed):
I ended up looking into the issue myself and concluded that the only change necessary to fix this is to add `ORDER BY p.categoryID, m.Name, p.description` to the query on line 311 of `php/libraries/UserPermissions.class.inc`. that addition should bring it up to par with the query on line 307 and fix the display issue that only occurs when the users doing the editing is an admin user.

see https://github.com/aces/Loris/blob/main/php/libraries/UserPermissions.class.inc#L311

Issue caused by changes in #7451

maltheism:
@ridz1208 thanks for the amazing one line fix!

ridz1208 (approved):
Beautiful work. Very elegant solution if I may say

PR #7803, opened 2021-11-08
[brainbrowser] Remove getImageName script

driusan:
This removes the getImageName script from the brainbrowser module.

The script's only purpose is to return a file's filename, given the
FileID after image.php retrieves the file. However, image.php
already looks up the filename. image.php passes the FileID in an X-FileID
header, so this extends that to add an X-Filename header to pass the
name as well as the ID. This is cached in an object and used to update
the file template after the volumes are downloaded.

1. Test Brainbrowser in LORIS. Behaviour should be unchanged (in particular the filename displayed.)


driusan:
Replaced by #7824

PR #7806, opened 2021-11-08
[BrainBrowser] Fix duplicate download of image data

### LINKED ISSUE #7805: BrainBrowser downloads image files twice
**Describe the bug**
When loading a scan from the imaging browser in brainbrowser, the call to `image.php?fileID=xxx` is made twice. This results in twice as much data being sent than required, as the (potentially significant) file is downloaded twice.

**To Reproduce**
1. Load an image in brainbrowser
2. Open the network tools
3. Reload the page so that you can see the requests sent
4. Note that the same request is made for the same fileid multiple times.

Image files would be downloaded once and used once downloaded.

This is happening both on demo.loris.ca and on my dev sandbox environment, so it seems to have existed for a while.

driusan:
This fixes an issue where image data is downloaded twice by
BrainBrowser. The brainbrowser.loris.js file makes a request to
image.php to get metadata for the file from http headers, and then passes
the URL to brainbrowser which makes a separate data.

Replace the GET request with a HEAD request from the LORIS side so
that only the brainbrowser library downloads the data.

Resolves #7805

### DIFF
--- a/modules/brainbrowser/js/brainbrowser.loris.js
+++ b/modules/brainbrowser/js/brainbrowser.loris.js
@@ -754,7 +754,7 @@ $(function() {
         $.ajax({
           url: loris.BaseURL + "/brainbrowser/ajax/image.php",
           data: 'file_id=' + minc_id,
-          method: 'GET',
+          method: 'HEAD',
           success: function (response, status, jqXHR) {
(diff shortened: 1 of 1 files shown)


driusan:
Replaced by #7824

PR #7807, opened 2021-11-08
[survey_accounts] Fixed base URL not being set properly

### LINKED ISSUE #7668: [survey_accounts] Survey Email does not contain clickable link
**Describe the bug**
The email that is supposed to contain a link to a particular survey only contains the path, instead of a clickable link. 

survey_accounts Survey Email

**To Reproduce**
1. Go to the Survey Accounts module and click "Add Survey" button
2. Fill out the form and send email by clicking "Email Survey"
3. Go to your email to view 

I expected the email to contain a clickable link. 

**Additional context**
The path is for the correct survey, so pasting the path after "https://test-loris-dev.loris.ca" works.

AlexandraLivadas:
When `\NDB_Factory::singleton()->settings()->getBaseURL()` was called, an empty string was returned, so the emails sent to users when a survey is created were not displaying the correct URL. This bug fix should also solve similar errors from occurring in other modules that use the above line.

The issue occurs because the line `$DB = $factory->database()` is called in [NDB_Cient](https://github.com/aces/Loris/blob/6356a1ea5d01e719ca560695d31bcf569004f3e6/php/libraries/NDB_Client.class.inc#L59) before the baseURL for the factory object is set to the correct value in the BaseRouter using `setBaseURL`. The database() function constructs the factory's static Settings object with an empty baseURL, so the baseURL ends up being an empty string instead of the correct value. This bug is fixed by adding the line  `$factory->reset()` before setting the URL, which resets the Settings object and its variables.  

**For the survey account module**
1. Go to the Survey Accounts page and add a new survey
2. Include an email address and select "Email Survey"
3. Check that the email contains a **clickable** link that directs you to the correct survey 

**For the bug in general:**
To see that the line `\NDB_Factory::singleton()->settings()->getBaseURL()` works correctly throughout the codebase, you can add an error_log message anywhere it is used and see that it returns the base URL for your Loris instance. 

#### Link to related issue

* Resolves #7668

### DIFF
--- a/src/Router/BaseRouter.php
+++ b/src/Router/BaseRouter.php
@@ -118,6 +118,8 @@ class BaseRouter extends PrefixRouter implements RequestHandlerInterface
             $baseurl = $uri->withPath($baseurl)->withQuery("");
             $request = $request->withAttribute("baseurl", $baseurl->__toString());
 
+            // Reset factory object so that the base URL can be reset
+            $factory->reset();
             $factory->setBaseURL($baseurl);
 
             $module  = \Module::factory($modulename);
--- a/src/Router/BaseRouter.php
+++ b/src/Router/BaseRouter.php
@@ -118,8 +118,6 @@ class BaseRouter extends PrefixRouter implements RequestHandlerInterface
             $baseurl = $uri->withPath($baseurl)->withQuery("");
             $request = $request->withAttribute("baseurl", $baseurl->__toString());
 
-            // Reset factory object so that the base URL can be reset
-            $factory->reset();
             $factory->setBaseURL($baseurl);
 
             $module  = \Module::factory($modulename);
--- a/php/libraries/NDB_Factory.class.inc
+++ b/php/libraries/NDB_Factory.class.inc
@@ -313,7 +313,9 @@ class NDB_Factory
      */
     public function setBaseURL(string $baseURL)
     {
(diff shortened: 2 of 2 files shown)


src/Router/BaseRouter.php
```diff
@@ -118,6 +118,8 @@ public function handle(ServerRequestInterface $request) : ResponseInterface
             $baseurl = $uri->withPath($baseurl)->withQuery("");
             $request = $request->withAttribute("baseurl", $baseurl->__toString());
 
+            // Reset factory object so that the base URL can be reset
+            $factory->reset();
```
ridz1208, line 122: @AlexandraLivadas What's the value of the baseURL before the reset ??

AlexandraLivadas: The value is an empty string

ridz1208: @AlexandraLivadas `reset()` will clear important values along with the settings singleton that we need cleared at this point. this is too risky.

What I think would be better is to reset the settings variable directly in NDB_Factory::setBaseURL() function. so directly in that function, set the settings to null and next time it is instanciated it will have the correct value.

Please try that change and test it and if all is good I'll pass review

ridz1208 (changes requested):

ridz1208 (approved):
Code looks good

PR #7808, opened 2021-11-08
[user_accounts] User with only superuser permission selected can edit other user's permissions

### LINKED ISSUE #7657: [user_accounts] Missing error message when assigning permissions as superuser
**Describe the bug**
When a superuser, that only have the "highlander" permission, adds permissions to an other user, there is no error message even though nothing happens in the backend. 

**To Reproduce**
0. As a superuser that only have the "highlander" permission.
1. Go to '/user_accounts/edit_user/{any_user_id}'
2. Add permissions
3. Scroll down to click Save.
4. See no error message

A message saying "Unable to add permission X; you don't have that permission."

- On test vm, any browser.

AlexandraLivadas:
## Brief summary of changes
If a user has superuser permission but no other permissions individually selected, they were not able to edit the permissions of another user. This is fixed by adding an override in the `_process` function of the edit_user class which checks if a user is a superuser before removing any permissions that the superuser "does not have" from the list of edited permissions.

1. Create a new user and select the superuser permission but not any other permissions specifically
2. Logged in as that superuser, go to some other user that has some permissions already set.
3. Add/delete permissions for that user.
4. See that the permission list is correctly updated once the page is reloaded and in the backend.

* Resolves #7657

### DIFF
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -234,10 +234,12 @@ class Edit_User extends \NDB_Form
             }
         }
 
-        // Keep old permissions if the editor is not allowed to edit them.
+        // Keep old permissions if the editor is not allowed to edit them,
+        // unless editor is superuser
         foreach ($current_permissionids as $perm) {
             if (!in_array($perm, $newPermissions)
                 && !in_array($perm, $editorPermissions)
+                && !$editor->hasPermission('superuser')
             ) {
                 $permIDs[] = $perm;
             }
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -224,10 +224,14 @@ class Edit_User extends \NDB_Form
         if (!empty($values['permID'])) {
             // An editor can only grant permissions that they have already been
             // granted.
-            $permIDs = array_intersect(
-                $newPermissions,
-                $editorPermissions
-            );
+            if ($editor->hasPermission('superuser')) {
+                $permIDs = $newPermissions;
+            } else {
+                $permIDs = array_intersect(
+                    $newPermissions,
+                    $editorPermissions
+                );
+            }
         }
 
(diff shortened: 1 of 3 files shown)


driusan:
Linked issue is tagged as critical, PR should probably be too..

ridz1208:
@AlexandraLivadas tests are failing

And can you detail what's the required discussion ?

maltheism (approved):

driusan:
@AlexandraLivadas just noticed this is still failing

ridz1208:
@AlexandraLivadas given the few places where that function is used in the edit_user logic and the complexity of the edit_user form I think the best path is not to amend the getPermissionIDs directly (because we might still want to know the real permissions assigned to a "superuser" especially when a user is made superuser exceptionally or erroneously. With this PR is a superuser edits another superuser they will automatically associate them with ALL permissions without wanting to do it).


Instead I think you should just alter the logic of the setup function to allow for the permissions to be displayed even if not checked when the user is a superuser.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -224,10 +224,14 @@ class Edit_User extends \NDB_Form
         if (!empty($values['permID'])) {
             // An editor can only grant permissions that they have already been
             // granted.
```
ridz1208, line 226:
```suggestion
            // granted. Unless they are a superuser
```

ridz1208:
@AlexandraLivadas I think the description at the top here needs updating given the latest changes

ridz1208 (changes requested):
It seems to be working for adding a permission but not for removing a permission

ie. login as a superuser with no other permissions than the superuser one -> add a couple permissions to another user -> save and reload ->permissions are selected -> try to remove one of those permissions -> reload after save -> permission is still there

AlexandraLivadas:
@ridz1208 Ready for re-review

ridz1208 (approved):

PR #7816, opened 2021-11-10
[imaging_browser] Use API for thumbnail and download

driusan:
This updates the imaging browser to use the "thumbnail"
API which is specifically defined for that purpose. It
also updates the "Download Minc" button to use the API.


cmadjar (approved):
LGTM 👍

PR #7817, opened 2021-11-10
[Subproject] New class

ridz1208:
## Brief summary of changes
New implementation of the subproject class in LORIS. Currently subprojects have a role in the dataabse but are not represented in any way in the code. This PR adds a Subproject class and controller to handle the buisness logic side.

### Design
As you can imagine several decisions made here are pure design decisions and can be contested/discussed. here is a quick breakdown of the classes and their intended use.

`Subproject.class.inc`
The subproject object is intended to be as "dumb" as possible. Its is only meant to be a DTO. In the current implementation the subproject object is meant to be **almost** "valid". Perfect in the sense that object attributres that should not be null are not nullable*(small exception stated later), attributes must be of the correct type to be added to the object and getters and setters can do basic validation and throw exceptions. 

**_NOTE:_** The "**almost** valid" statement is because the only exception to the validity of the object is when it comes to the ID. The Identifier is not a nullable field in the database and yet is allowed to be null in the object (Note however that it can not be nullified nor altered once it is set and is only ever set by the constructor on creation of the object !) and the reason for that is to allow code to create an object and set all it's attributes properly BEFORE the object is created in the database. The advantage of this method is that we don't need to duplicate the subproject object's constructor signature in the `createSubproject()` function, instead we can simply build a subproject without ID and pass it to the create method

`SubprojectController.class.inc`
This class plays the role of the Controller and Data access object (interfacing with the database). This class contains ALL the business logic for subprojects including contextual validation of the subproject object (duplication, incorrect input, and what not) as well as all the database interfacing code (insert, create, update, and other entity association modifications). Finally this class contains "Meta" functions such as `getAllSubprojects()`, `loadVisits()`, ... and factory methods like `getSubprojectFromID(ID)`


php/libraries/Subproject.class.inc
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Gets the visit names for that subproject
+     *
+     * @return array An array of Visit object
+     */
+    public function getVisits(): array
+    {
+        $factory    = \NDB_Factory::singleton();
+        $visitsData = $factory->database()->pselect(
```
driusan, line 65: The constructor should probably have a `\LORIS\LorisInstance` in order to avoid singletons

php/libraries/Subproject.class.inc
```diff
@@ -0,0 +1,109 @@
...
+            ",
+            [
+                'v_subproject_id' => $this->_subproject_id,
+                'v_project_id'    => $this->_project_id
+            ]
+        );
+
+        return array_map(
+            function ($visit) {
+                return new \LORIS\Visit($visit['name']);
```
driusan, line 86: The `Visit` object takes a name and a label, should probably include both since we don't know which the caller needs

php/libraries/Subproject.class.inc
xlecours: @driusan In addition to the properties of a subproject? Should there be 2 classes then? One to "manage" subprojects and one that contains a subproject properties/data.

driusan: if the class depends on it it should be injected somewhere. An alternative would be to make it an argument to `getVisits()` since it's getVisits that depends on it.

ridz1208: I dont like passing non-DTO stuff in the constructor, mixes concerns

ridz1208: I'm going with the controller solution

controller/DAO + DTO

xlecours:
Why all the getters and setters? Are they ever used?

ridz1208:
@xlecours i use the getters off and on. But the setters are just there for good measure. I don't think ik using them anywhere but they can be used since the variables are protected. The setters would be used for example when updating the subproject from the configuration module.

php/libraries/SubprojectController.class.inc
```diff
@@ -0,0 +1,252 @@
...
+    public function getAllSubprojects(): array
+    {
+        $subprojectRows = $this->database->pselect(
+            "SELECT
+                SubprojectID,
+                title,
+                useEDC,
+                WindowDifference,
+                RecruitmentTarget
+            FROM subproject
```
xlecours, line 70: That select statement should be put in a class variable so it sould be used in `getAllSubprojects`, `getSubprojectByTitle` and `getSubprojectByID`.

php/libraries/SubprojectController.class.inc
```diff
@@ -0,0 +1,252 @@
...
+        foreach ($subprojectRows as $row) {
+            $subproject    = new \LORIS\Subproject(
+                intval($row['SubprojectID']),
+                $row['title'],
+                $row['useEDC'] === '1',
+                $row['WindowDifference'],
+                isset($row['RecruitmentTarget'])
+                    ? intval($row['RecruitmentTarget']) : null
+            );
+            $subprojects[] = $subproject;
```
xlecours, line 85: The instantiation can be done from the PDO using `PDO::FETCH_CLASS`. I think it would be an elegant way to reduce code duplication. The same validations and type conversions are done here and in `getSubprojectByTitle`.

What would need to be done though is to have a constructor without parameters and to do the type conversion inside it.

php/libraries/SubprojectController.class.inc
```diff
@@ -0,0 +1,252 @@
...
+                    useEDC,
+                    WindowDifference,
+                    RecruitmentTarget
+                FROM subproject
+                WHERE title = :title",
+                ['title' => $subprojectTitle]
+            );
+
+            if (empty($subprojectData)) {
+                throw new \LorisException(
```
xlecours, line 116: We have a NotFound exception

php/libraries/SubprojectController.class.inc
```diff
@@ -0,0 +1,252 @@
...
+              title
+            FROM
+              subproject
+            WHERE SubprojectID = :id
+            ",
+            ['id' => $subprojectID]
+        );
+
+        if (empty($subprojectTitle)) {
+            throw new \LorisException("No subproject with ID $subprojectID found.");
```
xlecours, line 157: We have a NotFound exception

php/libraries/SubprojectController.class.inc
```diff
@@ -0,0 +1,252 @@
...
+    public function createSubproject(\LORIS\Subproject $subproject)
+    : \LORIS\Subproject
+    {
+        if (!empty($subproject->getID())) {
+            throw new \LorisException(
+                "An ID can not be supplied when creating a new subproject, the ID
+                will be generated by the database."
+            );
+        }
+        $subprojectExists = $this->subprojectExists($subproject->getTitle());
```
xlecours, line 179: You don't need to do that check. There will be a DatabaseException on constraint violation.
Or, if you really want that check, you should also add it in the update function.

php/libraries/SubprojectController.class.inc
```diff
@@ -0,0 +1,252 @@
...
+        );
+
+        if (empty($subproject)) {
+            return false;
+        }
+        return true;
+    }
+
+    //TODO: loadProjects
+    //TODO: loadVisits(Project)
```
xlecours, line 250: I don't think a SubprojectController should produce Projects nor Visits. I think ProjectController and VisitController should produce them based on a given Subproject. Inversely, subprojectController should produce subprojects based on given visits and/or projects.

```
getSubprojectFromProject(\Project $p): \Traversable
{
    $subprojects =  $this->database->pselect(
        "SELECT 
            sp.*
            from project_subproject_rel psr
            LEFT JOIN subproject sp ON (psr.SubprojectID = sp.SubprojectID)
            WHERE psr.ProjectID = :v_projectid",
        ['v_projectid' => $p->getID()]
    );
    ... if PDO::FETCH_CLASS is used then ...
    yield from $subprojects;
}

getSubprojectFromVisit(\Visit $v): Subproject
{
   ...
}
```

php/libraries/SubprojectController.class.inc
```diff
@@ -0,0 +1,252 @@
...
+            [
+                'title'             => $subproject->getTitle(),
+                'useEDC'            => is_null($subproject->getUseEDC())
+                    ?: ($subproject->getUseEDC() ? 1 : 0),
+                'WindowDifference'  => $subproject->getWindowDifference(),
+                'RecruitmentTarget' => $subproject->getRecruitmentTarget(),
+            ],
+            ['title'  => $subproject->getTitle()]
+        );
+
```
xlecours, line 225: The `$_instances` variable should be updated as well.

php/libraries/SubprojectController.class.inc
```diff
@@ -0,0 +1,252 @@
...
+ * @link     https://www.github.com/aces/Loris/
+ */
+class SubprojectController
+{
+    /**
+     * Array holding existing instances
+     *
+     * @var array
+     */
+    private static $_instances = [];
```
xlecours, line 32: Using a static variable as a cache can provide economy but it comes with it share of bugs to detect and patch (ex: my comment in the updateSubproject).  I don't think it provides any economy here in most cases. And little economy in few cases. I can't see when multiple subprojects would be instantiated multiple times in the same HTTP request handling. And even if that happened, I am sure it won't be noticeable.

php/libraries/SubprojectController.class.inc
```diff
@@ -0,0 +1,252 @@
...
+        $this->database->update(
+            'subproject',
+            [
+                'title'             => $subproject->getTitle(),
+                'useEDC'            => is_null($subproject->getUseEDC())
+                    ?: ($subproject->getUseEDC() ? 1 : 0),
+                'WindowDifference'  => $subproject->getWindowDifference(),
+                'RecruitmentTarget' => $subproject->getRecruitmentTarget(),
+            ],
+            ['title'  => $subproject->getTitle()]
```
xlecours, line 223: What if the title is updated?

php/libraries/SubprojectController.class.inc
ridz1208: So you are proposing to kill the singleton logic ? just reinstanciate when needed ? (I have 0 oppositions, just asking)

php/libraries/SubprojectController.class.inc
ridz1208: I don't like that, but I could maybe make a class array (for selected fields) and a tablename (for FROM clause) variable and concatenate in the different queries... if that's okai with you I'll do it

php/libraries/SubprojectController.class.inc
xlecours: Technically, the subproject controller is not a singleton. It would need to have  a private constructor and a "getInstance" function to make sur there is only a single static subprojectcontroller in the session. But yes, I think we should stay away from using static variables unless it provides substantial benefits.

ridz1208: I meant a factory of subproject singletons. But I see your point and don't mind dropping that

php/libraries/SubprojectController.class.inc
ridz1208: Henri had implemented, `objectToSQL()` and `sqlToObject()` functions to do this works. you wanna go with that ?

php/libraries/SubprojectController.class.inc
ridz1208: I know but I thought not found was for files not found on system or something... not database entities (I actually had it as notfound and changed it)

php/libraries/SubprojectController.class.inc
ridz1208: I didnt check if its a constraint violation (is there unique on the database side ?)
and good point for update, will change it depending on first question

php/libraries/SubprojectController.class.inc
ridz1208: hah, yes! I'll use the ID

php/libraries/SubprojectController.class.inc
xlecours: I don't see why you need those values in distinct variables. 
What would be the difference between :
`$db->select('SELECT' . implode(',', $selectedfields) . " FROM " . $tablename);`
and 
`$db->select($sqlselect);`

php/libraries/SubprojectController.class.inc
ridz1208: probably gonna remove it

php/libraries/SubprojectController.class.inc
xlecours: I don't think we need any function to do it. What I am saying is that PDO can handle this.

php/libraries/SubprojectController.class.inc
ridz1208: aesthetic mostly!!

php/libraries/SubprojectController.class.inc
driusan: NotFound exceptions tend to get converted to 404 errors by middleware.

That should probably be the criteria you use to determine if it makes sense to use it.

php/libraries/SubprojectController.class.inc
xlecours: Humm, no. There is no contraint on the subproject table other than the surrogate key.
In that case, `getSubprojectByTitle` should return a Traversable of Subprojects. Or we should add a unique constraint.

php/libraries/SubprojectController.class.inc
xlecours: Basically, 
```
public function getAllSubprojects(): \Traversable
{
    $statement = $this->_db->prepare($selectstring);
    $statement->setFetchMode(\PDO::FETCH_CLASS, '\LORIS\Subproject');
    yield from $statement->execute();
}

public function getSubprojectByID(string $id): Subproject
{
    $statement = $this->_db->prepare($selectstring . 'WHERE SubprojectID = :v_id');
    $statement->setFetchMode(\PDO::FETCH_CLASS, '\LORIS\Subproject');
    $statement->execute(['v_id' => $id]);
    return $statement->fetch();
}
```

driusan: if that's going to become a common idiom I would think a $DB->pselectInto  and pselectOneInto would be a better idea than duplicate code everywhere.. I can send a PR, but what's the point of `yield from` when execute already returns a \Traversable?

ridz1208:
@driusan true but it allows you to control what variables you want to make immutable like ID
and define some basic validation if needed like throw exception when it's null and shouldnt be

php/libraries/SubprojectController.class.inc
ridz1208: i think constraint is the way to go and we remove the soft validation I think

php/libraries/SubprojectController.class.inc
ridz1208: @driusan @xlecours i'm not sure that answers my question. if you query for a candidate that doesnt exist... is that a 404?? i dont think so no ?

php/libraries/SubprojectController.class.inc
ridz1208: @xlecours how does PDO map what column maps to what variable ?

xlecours: @driusan you are right, `yield from` is redondant. And you could add `$DB->pselectInto` but you will have to alter the execute function.

xlecours: @ridz1208 It sets the object variable as the column names from your query. Then it call the constructor.

php/libraries/SubprojectController.class.inc
```diff
@@ -0,0 +1,252 @@
...
+    public function updateSubproject(\LORIS\Subproject $subproject)
+    : \LORIS\Subproject
+    {
+        // Update subproject table
+        $this->database->update(
+            'subproject',
+            [
+                'title'             => $subproject->getTitle(),
+                'useEDC'            => is_null($subproject->getUseEDC())
+                    ?: ($subproject->getUseEDC() ? 1 : 0),
```
ridz1208, line 219:
```suggestion
                    ? null : ($subproject->getUseEDC() ? 1 : 0),
```

php/libraries/SubprojectController.class.inc
ridz1208: completely agree

php/libraries/SubprojectController.class.inc
```diff
@@ -0,0 +1,252 @@
...
+                "SubprojectController - Database not connected"
+            );
+        }
+        $this->database = $database;
+    }
+
+    /**
+     * Retruns a list of configured Subprojects
+     *
+     * @return array The list of subprojects
```
ridz1208, line 59:
```suggestion
     * @return \Subproject[] The list of subprojects
```

php/libraries/SubprojectController.class.inc
ridz1208: I don't think I like that solution. 

1. I'm trying to dissociate the object from the DB as much as possible and keep the DAO/Controller more DB involved
2. It seems to instantiate the variables implicitly and I don't like that
3. some sub-entities like projects or visits will anyways be loaded by the controller and will have to be defined separately making a mix of implicit and explicit variables...

All in all it's kind of cool but I think I would go for it if there was a way to define the variables explicitly and typed and then map them somehow to the DB column names using PDO. not the current implementations

php/libraries/SubprojectController.class.inc
ridz1208: will change to notFound

php/libraries/SubprojectController.class.inc
ridz1208: will change to NotFound

php/libraries/SubprojectController.class.inc
ridz1208: @xlecours adding the constraint will send this PR to 25 directly, not adding it means we have to "expect" multiple subprojects with the same name but it could still be merged in 24...

I'm torn

ridz1208:
Will remove all getters and setters in favour of typed variables in HP 7.4

php/libraries/SubprojectController.class.inc
xlecours: I understand you like writing code. It was just a suggestion.

php/libraries/SubprojectController.class.inc
xlecours: If you do it wrong and it works, it will stay until eternity ends. I am feeling philosophical :)

ridz1208: no pressure... jeez!

php/libraries/SubprojectController.class.inc
```diff
@@ -0,0 +1,254 @@
...
+    protected $database;
+
+    /**
+     * The constructor
+     *
+     * @param \Database $database The database connection to use
+     */
+    public function __construct(\Database $database)
+    {
+        if (!$database->isConnected()) {
```
xlecours, line 41: do you need to do that? It's unlikely that tis is the first place an exception will rise if the database is not connected.

php/libraries/Subproject.class.inc
```diff
@@ -0,0 +1,120 @@
...
+        }
+        return $this->id;
+    }
+
+    /**
+     * Getter for associated projects
+     *
+     * @return array Projects associated with this subproject
+     */
+    public function getProjects() : array
```
xlecours, line 78: this can be removed now that the project class provides the same function.

php/libraries/Subproject.class.inc
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /**
+     * Setter for Projects associated to this subproject
+     *
+     * @param array $projects Array of project objects to add to the subproject
+     *
+     * @return void
+     */
+    public function setProjects(array $projects)
```
xlecours, line 97: user should look at the Project:getProjectsForSubproject now

php/libraries/SubprojectController.class.inc
```diff
@@ -0,0 +1,254 @@
...
+     * Query projects associated with the supplied subproject and load the
+     * projects into the subproject object. The querying and instanciation
+     * of projects is done by the Project class.
+     *
+     * @param \LORIS\Subproject $subproject The subproject object for which we want
+     *                                      the associated subprojects
+     *
+     * @return void
+     */
+    public function loadProjects(\LORIS\Subproject $subproject) : void
```
xlecours, line 247: user should look at the Project:getProjectsForSubproject instead

php/libraries/SubprojectController.class.inc
ridz1208: good point, not sure where I got that from, it might be a leftover from melanie's original code

php/libraries/Subproject.class.inc
ridz1208: to be discussed

php/libraries/Subproject.class.inc
ridz1208: to be discussed

ridz1208:
@driusan @xlecours are we sufficiently happy with this to get it merged ? (keep in mind there is SQL in this PR that is not exactly backwards compatible but if its good enough to merge and it's approved I can start basing other work on top of it)

xlecours (approved):

xlecours:
I don't see how or why `subprojectExists` would be used but... you arleady deleted alot :)

ridz1208:
Adding blocked until release is out

ridz1208:
blocking: https://github.com/aces/Loris/pull/7981
blocked by: https://github.com/aces/Loris/pull/7980

ridz1208:
@driusan @xlecours 

I rebased this and modified it a bit to remove some unnecessary code now with 8.1. I'm still not sure where to put the files (I also added a comment on Dave's CandidateData PR) but other than that it should be pretty much in line with other objects we have been creating.

ridz1208:
@driusan @xlecours 

I rebased this and modified it a bit to remove some unnecessary code now with 8.1. I'm still not sure where to put the files (I also added a comment on Dave's CandidateData PR) but other than that it should be pretty much in line with other objects we have been creating.

PS I think its failing tests beacause readonly is only for 8.1 ...

php/libraries/SubprojectData.class.inc
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @category Main
+ * @package  Loris
+ * @author   Xavier Lecours Boucher <xavier.lecours@mcin.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+class SubprojectData implements DataInstance
+{
+    public readonly int $id;
```
driusan, line 30: This is causing the tests to fail. readonly properties only exist in PHP 8.1+

ridz1208:
@driusan @xlecours ready again

PR #7819, opened 2021-11-10
[NDB_BVL_Instrument] JSON Data support fix DECS field

ridz1208:
## Brief summary of changes
When Data column is empty (before first save) this returns `NULL` which causes an issue since this method expects a string returned


ridz1208:
@driusan this is a fix going to 23

CamilleBeau (approved):

driusan (approved):

PR #7818, opened 2021-11-10
[NDB_BVL_Instrument] fix support for JSONData instruments

ridz1208:
## Brief summary of changes
Fix a couple instances where the Instrument class does not support JSONData instruments by either looking for a specific field in the not-yet-populated instrument data object or using the `$this->table` variable

Testing

1. choose a JSONData=true instrument (you can turn `false` into `true` in the `bmi.meta` file for an easy instruement)
2. set he `Data` column to `NULL` directly in the flag table in the database
3. try entering the "Date of Administration" and other required fields on the top page and save.

Without this PR the save will fail claiming the `Date_taken` field does not exist in the instrument. With this PR, issue should be solved

PR #7820, opened 2021-11-10
[RB] Fix BMI scoring code

ridz1208:
## Brief summary of changes
updating scoring code to latest standards

### DIFF
--- a/raisinbread/instruments/bmi.score
+++ b/raisinbread/instruments/bmi.score
@@ -1,32 +1,34 @@
 #!/usr/bin/php
 <?php
 /**
-* BMI Scoring
-*
-* PHP version 5
-*
-* @category Main
-* @package  BMI_Instrument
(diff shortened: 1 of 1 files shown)


ridz1208:
@zaliqarosli could you test this at the same time when testing #7818 they are related. The BMI instrument in 7818 tries to run the scoring based on the SQL values but if it is set to `jsondata=true` the SQL table would be empty.

All you need to do is make sure the data loads, calculates and save in the JSON string the proper scores

zaliqarosli (changes requested):
getting the console output error

```
An error occurred while running the scoring
                algorithm of instrument bmi.
```

once again, i'm not able to see anything useful in my apache log

ridz1208:
can you just make sure the .score file is available and executable on your VM ?

zaliqarosli (approved):
@ridz1208 you're right, i needed to make the file executable. the scoring ran well, and score fields saved to Data column of flag table

PR #7824, opened 2021-11-11
[brainbrowser] Fully remove ajax directory

driusan:
This updates brainbrowser in LORIS to get rid of the ajax directory.
In doing so, it also updates the architecture to be somewhat more
sane.

Previously, a request would be made from javascript to an ajax script
which takes a FileID and returns the raw file content with some metadata
in HTTP headers. This request would be made once per file to be displayed in BrainBrowser. The body of the request would then be discarded, and the
same endpoint would be passed to brainbrowser. Brainbrowser would download
the file content a second time, and after all files were re-downloaded
display the volume viewer.

In the new architecture, an imageinfo endpoint is added which returns
the metadata in JSON format. Included in the information in the response
is a URL which can be passed to brainbrowser to download the raw file
data. The LORIS API is used for imaging browser files, and a new endpoint
is added to the mri_violations module for files which can not be accessed
through the API.

The end result is that the ajax directory is no longer used by brainbrowser,
the API is used to download files, and  more access checking is
done on accessing mri violations files.


cmadjar (changes requested):
This does not work for me. I get the following error with an infinite loop:
Screen Shot 2022-01-18 at 12 02 25 PM

driusan:
@cmadjar I think I managed to reproduce and fix this problem, can you try again?

cmadjar (approved):

PR #7825, opened 2021-11-11
[create_timepoint] Code refactor to display proper select options based on previous selections

### LINKED ISSUE #7602: [create_timepoint] No error message to user if visit_label not set
**Describe the bug**
Step 4 of the TestPlan indicates: `Ensure that if the config.xml file does not contain a matching <visitLabel> entry, an error message should be displayed to the user.` 
For a subproject that does not have any visits set, no error message is displayed even though there are no visits available to select.

image

**To Reproduce**
1. Go to a candidate profile
2. Click on 'Create Timepoint'
3. Select a subproject that does not have any visitlabels set in the config.xml file
4. See that there is no error message

An error message to be displayed indicating that the subproject does not have any visits available.

maltheism:
## Brief summary of changes

This PR fixes a few long-standing issues exacerbated by the addition of user-multisites and user-multiprojects. Below is the list of changes:
 - Cleanup of code duplication and inefficiencies
 - Reorganization of the fields in a more logical order on the front end
 - conditional display of options based on previous selections
 - swal prompts when study configurations are missing preventing the creation of a timepoint

* Resolves #7602

### DIFF
--- a/modules/create_timepoint/jsx/createTimepointIndex.js
+++ b/modules/create_timepoint/jsx/createTimepointIndex.js
@@ -141,7 +141,6 @@ class CreateTimepoint extends React.Component {
             state.form.display.project = true;
           }
           // Populate the select options for visit.
-          console.log(data);
           if (data.visit) {
             // Store the (complete) visit selection information.
             state.storage.visit = data.visit;
--- a/modules/create_timepoint/jsx/createTimepointIndex.js
+++ b/modules/create_timepoint/jsx/createTimepointIndex.js
@@ -210,16 +210,7 @@ class CreateTimepoint extends React.Component {
    */
   handleVisitLabel() {
     const state = Object.assign({}, this.state);
-    if (Array.isArray(state.storage.visit[state.form.value.subproject])) {
-      // Display error message to user.
-      const errorMessage = `No visit labels defined for subproject: ${
-        this.state.form.options.visit[
-          this.state.form.value.subproject
-      ]}`;
-      state.messages = [errorMessage];
-      swal.fire(errorMessage, '', 'error');
-      state.form.options.visit = {};
-    } else if (state.storage.visit[
+    if (state.storage.visit[
         state.form.value.project
       ] !== undefined) {
       if (Array.isArray(state.storage.visit[
--- a/modules/create_timepoint/jsx/createTimepointIndex.js
+++ b/modules/create_timepoint/jsx/createTimepointIndex.js
@@ -178,12 +178,13 @@ class CreateTimepoint extends React.Component {
   handleSubproject() {
     const state = Object.assign({}, this.state);
     if (Array.isArray(state.storage.subproject[state.form.value.project])) {
-      // Display error message to user. show swal error.
-      state.messages = [
-        `No subprojects defined for project: ${this.state.form.options.project[
+      // Display error message to user.
+      const errorMessage = `No subprojects defined for project: ${
+        this.state.form.options.project[
           this.state.form.value.project
-        ]}`,
-      ];
+      ]}`;
+      state.messages = [errorMessage];
+      swal.fire(errorMessage, '', 'error');
       state.form.options.subproject = {};
       state.form.options.visit = {};
     } else {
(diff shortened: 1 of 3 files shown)


ridz1208:
@maltheism this is good, I'll add some stuff to this PR

maltheism:
@ridz1208 let me know if you need me to change anything cause of your PR logic.

ridz1208:
@maltheism I pushed the backend cleanup which include a couple logic changes

- The subproject list is no longer a single dimension list, it is not 2D and grouped by project. What needs to be fixed on the front end is the following. When a user selects a project, the subproject list should dynamically be updated based on the selected project. The array of subprojects should be of this format 
   ```
   [
        projectID_1 => [subprojectID_1 => title_1, subprojectID_2 => title_2],
        projectID_2 => [subprojectID_3 => title_3, subprojectID_2 => title_2]
   ]
   ```
   Also, the `subproject` key was changed to `subprojectGroups` to be more accurate.
- The visit list is similar, it was changed to a 3D array grouped by Project and then subproject. so based on the selection of those 2 variables, the visit list should be dynamically updated.
-  Lastly the ordering of the fields on the front end should also now be
    1. Site
    2. Project
    3. Subproject
    4. Visit

maltheism:
@ridz1208 which PR did you push those mentioned changes?

ridz1208:
@maltheism refresh

modules/create_timepoint/php/timepoint.class.inc
```diff
@@ -64,43 +64,39 @@ class Timepoint extends \NDB_Page implements ETagCalculator
...
+            }
         }
-        $values['visit'] = $visit_options;
-
-        // List sites
-        $values['psc'] = array_intersect(
-            \Utility::getSiteList(),
-            \User::singleton()->getSiteNames()
-        );
+        $values['visitGroups'] = $visitGroups;
```
ridz1208, line 99: @maltheism note key name change `visitGroups` for front end changes

modules/create_timepoint/php/timepoint.class.inc
```diff
@@ -64,43 +64,39 @@ class Timepoint extends \NDB_Page implements ETagCalculator
...
+        foreach ($projectOptions as $pid=>$pname) {
+            $subprojectGroups[$pid] = \Utility::getSubprojectList(
+                new \ProjectID($pid)
+            );
         }
-        $values['subproject'] = array_intersect_key(
-            $allSubprojects,
-            $candidate->getValidSubprojects()
-        );
+        $values['subprojectGroups'] = $subprojectGroups;
```
ridz1208, line 87: @maltheism note key name change `subprojectGroups` for frontend changes

php/libraries/Utility.class.inc
```diff
@@ -350,6 +350,41 @@ class Utility
         return $result;
     }
 
+    /**
```
ridz1208, line 353: @driusan I had to add this because we don't have a function that reurns visits associated to the project-subproject tuple (even though that is what defines visits) 

This also DOES NOT return proper visit objects because the visitclass does not properly define the getVisitFromName but instead defines getVisitsFromName which is incorrect and I am fixing in a futur PR.

ridz1208:
@maltheism I tried it out and the only problem I could find is when the visit list is empty there is no swal saying no visits associated with your project and subproject
Screen Shot 2021-12-03 at 3 08 46 PM

maltheism:
Hi @ridz1208, please review latest changes.

modules/create_timepoint/jsx/createTimepointIndex.js
```diff
@@ -222,10 +222,22 @@ class CreateTimepoint extends React.Component {
...
       ] !== undefined) {
-      state.form.options.visit = state.storage.visit[
-        state.form.value.project
-      ][state.form.value.subproject];
-      state.form.value.visit = null;
+      if (Array.isArray(state.storage.visit[
+        state.form.value.project][state.form.value.subproject])
+      ) {
+        const errorMessage = `No visit labels defined for subproject: ${
+          this.state.form.options.subproject[
```
ridz1208, line 220: the visit labels are dependent on a project-subproject combination. so the error message should display both project and subproject accordingly

maltheism: @ridz1208 do you mean add the project name to the error message?

maltheism:
Hi @ridz1208, I updated the error message for your suggestion.

ridz1208 (dismissed):
Tiny change and it's ready

ridz1208 (approved):

Issue #7190, opened 2020-12-14
[23.0][bvl_feedback / profile level] Not consistent behaviour of "bvl_feedback" at the "profile level" when the user has the permission "Behavioural QC" but not the "Across all sites access candidate profiles" one


jesscall:
Issue originally reported by @racostas on CCNA Repo during candidate release testing of `LORIS 23.0.2`
aces/CCNA#4269

___

Not consistent behaviour of  **bvl_feedback** at the **profile level** when the user **has the permission Behavioural QC** **but not the Across all sites access candidate profile**s one.

How to test:
(take a `user having the permission Behavioural QC` but `not the Across all sites access candidate profiles (neither the superadmin one)`)

1. Go to `MainMenu->Canditate-> Access Profile.`
2. Click on a candidate of choice.
3. Open the ` bvl_feedback` panel clicking on the notepad icon in the left part of the Main Menu bar (at the left of the help button sign) 
4. Create a new feedback on the `New profile level feedback tab` (middle one)
<img width="282" alt="image_bvl_feedback_profileLevelNotSuperUser_create" src="https://user-images.githubusercontent.com/37309344/102096079-4beaef80-3df2-11eb-9476-80a9907cb13f.png">

5. Notice in this same middle panel the message` (The new thread has been submitted!) ` but the upper `Feedback Threads `and bottom `Feedback Threads` tabs remains unchanged .  (if the same process is done with a `superuser` account or a `user having at the time Behavioural QC but and Across all sites access candidate profiles these two tabs are populated`. [Please note that the not shown info is nevertheless stored]
<img width="282" alt="image_bvl_feedback_profileLevelNotSuperUser_read" src="https://user-images.githubusercontent.com/37309344/102096479-c156c000-3df2-11eb-82bc-65d51cb9065b.png">

6. In the dashboard (clicking in the LORIS button at the top left) the feedback does shows on the `Behavioural Feedback Notifications` (the same not shown in step 5)

<img width="295" alt="image_bvl_feedback_profileLevelNotSuperUser_dashboard" src="https://user-images.githubusercontent.com/37309344/102096612-e9462380-3df2-11eb-9723-5f67cad33472.png">

7. If the notification referred in step 6 is clicked, and then again in the `bvl_feedback button` (the notepad icon) the feedback is not shown (step 5). 

**NOTE:** Please note that this issue is not observed when in the same conditions the user is providing the feedback at the `Visit or Instrument level`

jesscall:
[...]

PR #7826, opened 2021-11-12
[bvl_feedback] Fix Permissions for Feedback Summary & Thread List

suzanne-lee:
## Brief summary of changes

Currently, only users with the `access_all_profiles` permission can see Feedback Threads and Open Thread Summary at the profile level. Even if a user adds a feedback entry, they won't be able to see their own feedback thread unless they have the `access_all_profiles` permission.

This PR allows users to see threads and the summary that exists for candidates that they have access to (i.e., if a user is affiliated with MTL, they can now see the feedback threads & summary of MTL candidates).

1. Ensure that your user DOES NOT have the `access_all_profiles` permission and DOES have the `bvl_feedback` permission
2. Go to Candidate --> Access Profile
3. Click on any candidate 
4. Click the notepad icon in the top right-hand corner of the page (next to the question mark icon). This will open the feedback panel from the right side of the screen
5. Create a new feedback entry by typing in the large text field under "New profile level feedback", selecting a feedback type, then clicking the "Save data" button
6. The "Feedback Threads" card should populate with your newly created feedback entry, and the "Open Thread Summary" card should increment 
7. Log out and switch to a different user, and visit the same candidate to check if you can view the feedback thread 

* Resolves  https://github.com/aces/Loris/issues/7190

### DIFF
--- a/php/libraries/NDB_BVL_Feedback.class.inc
+++ b/php/libraries/NDB_BVL_Feedback.class.inc
@@ -475,25 +475,38 @@ class NDB_BVL_Feedback
             $query .= " LEFT JOIN flag as f ON (ft.CommentID = f.CommentID)";
         }
         $query .= " WHERE ft.Active ='Y'";
+
+        $query .= " AND Public = 'Y' AND Status <> 'closed'";
+        if (!empty($this->_feedbackObjectInfo['CandID'])) {
+            $query            .= " AND ft.CandID = :CaID";
+            $qparams['CaID']   = $this->_feedbackObjectInfo['CandID'];
+            $candidate         = Candidate::singleton(new CandID($qparams['CaID']));
+            $hasReadPermission = (
+                $hasReadPermission ||
+                $candidate->isAccessibleBy($user)
+            );
+        }
+        if (!empty($this->_feedbackObjectInfo['SessionID'])) {
+            $query            .= " AND ft.SessionID = :SID";
+            $qparams['SID']    = $this->_feedbackObjectInfo['SessionID'];
+            $timepoint         = Timepoint::singleton(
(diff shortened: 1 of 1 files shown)


php/libraries/NDB_BVL_Feedback.class.inc
```diff
@@ -479,25 +479,38 @@ class NDB_BVL_Feedback
...
+            );
         }
         if (!empty($this->_feedbackObjectInfo['CommentID'])) {
-            $query          .= " AND ft.SessionID = :CSID";
-            $qparams['CSID'] = $this->_feedbackCandidateProfileInfo['SessionID'];
+            $query           .= " AND ft.CommentID = :ComID";
+            $qparams['ComID'] = $this->_feedbackCandidateProfileInfo['CommentID'];
         }
+
+        if (!$hasReadPermission===true) {
```
jstirling91, line 509: This could be simplified to `!$hasReadPermission` or `$hasReadPermission` depending on what the expected statement is. Not overly clear in this logic.

suzanne-lee: i can fix this, but this piece of code was already in the file (i just moved it) and there's another place in the file that has the same condition as well

ridz1208:
@suzanne-lee your PR does indeed fix the reported issue, however as soon as you navigate into one of the candidate's visits, a similar issue occurs where the summary does not display the summary of open threads.

ridz1208:
@driusan this should probably go to 24.1, please confirm

driusan:
yes, looks like a bug fix.

laemtl:
@racostas Can you review after @ridz1208 rebases it?

racostas (approved):

PR #7831, opened 2021-11-17
[PHPCS] Forbid side-effect functions in src directory

driusan:
This uses PHPCS to forbid a number of deprecated or side-effect
inducing functions in the src/ directory of LORIS. They shouldn't
be used in general, but because of heavy usage in the ajax/ directories
can't be forbidden in the default ruleset right now.


test/SrcCS.xml
```diff
@@ -20,6 +20,19 @@
...
 
+    <rule ref="Generic.PHP.ForbiddenFunctions">
+        <properties>
+            <property name="forbiddenFunctions" type="array">
+                <element key="print" value="(PSR-3 logger)"/>
+                <element key="die" value="null" />
+                <element key="create_function" value="null"/>
+                <element key="http_response_code" value="(PSR-7)"/>
+                <element key="header" value="(PSR-7)"/>
+                <element key="error_log" value="(PSR-3 logger)"/>
```
ridz1208, line 31: this might be an issue for deprecated functions

driusan: What do you mean for deprecated functions? It's only the rule set for the `src` directory.

Though I had to remove it because LorisInstance (where there's not currently access to a logger) and the ErrorLogLogger (which needs to use it for obvious reasons) have references.

ridz1208: when we deprecate functions we usually add an error_log message. I guess there is a different way to do it with PSR3 ?

driusan: PSR3 is a logging standard, we have the ErrorLogLogger which prints to the error log and is customizable based on the config module. `$this->logger->warning("some message");` would be the preferred way in a context that has access to a PSR3 logger. (This might not be the case in places like the Utility class)

driusan:
@ridz1208 @xlecours could one of you review this?

xlecours (approved):
It compiles! ship it.
The list is a good start. Otherwise, we can always change it.

PR #7837, opened 2021-11-18
[Meta] Add PHP 8.1 and remove PHP 7.4 from GitHub actions

driusan:
PHP 8.1 is set to be released in a week. We should have our tests run with it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "squizlabs/php_codesniffer" : "^3.5",
         "phpunit/phpunit" : "9.4.4",
         "phan/phan": "^5.0",
-        "phpstan/phpstan": "0.12.58",
+        "phpstan/phpstan": "^1.4",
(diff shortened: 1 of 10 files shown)


driusan:
Seems to be failing because of a bug with phpcs and php 8.1 (it's failing on a rule that's excluded by our ruleset) but I don't have php 8.1 locally to test with yet and it's not reproducing on 8.0 for me. Will wait and test after the official release..

maltheism:
php 8.1 is released today

maltheism:
@driusan I'm unable to view the modules of Loris on php 8.1 after fixing some of the fatal errors in PR: https://github.com/aces/Loris/pull/7889

Here's a screenshot of Loris loading on php 8.1 but without modules displaying:
Screen Shot 2021-12-07 at 12 13 56 PM

driusan:
Sending a new PR, this was accidentally overridden with a force push.

PR #7842, opened 2021-11-22
[User Accounts] Users request are allowed to be created with "incorrect" email address via "user request" UI

### LINKED ISSUE #7827: [User Accounts] Users request are allowed to be created with "incorrect" email address via "user request" UI
**Describe the bug**
When a user request an account by itself (not created by the admin) the validations in the email field are not triggered. Thus users request are allowed to be created with "incorrect" email addresses. 

**To Reproduce**
1. Go to the main Loris page and request a new user.
2. Fill the email address field with and "incorrect" email address.
3. Fill the rest of the form. And send the request.
image
4. The request will be created.
5. Access the system now as a user with permissions to approve user request (admin, etc.)
6. Search for the new user request that was done in step # 2-3.
7. The request could be found with the respectively "incorrect" email address. 
image

**Related issue**
https://github.com/aces/Loris/pull/7789

In steep 2-3 an error would be shown to the user to notify the error. 
The request would not be created until a "correct" email address is provided.

- OS: Microsoft Windows 10.
- Browser: Chrome, Version 95.0.4638.54 (Official Build) (64-bit)

- LORIS testing server.

maltheism:
## Brief summary of changes

Created an EmailElement that uses type="email" (see [here for specs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email)). The user's browser will prevent the form being completed if the user has not entered a valid email when type="email".

1. Checkout PR.
2. Request account with invalid email

* Resolves #7827

### DIFF
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -1213,6 +1213,137 @@ TextboxElement.defaultProps = {
 };
 
 /**
+ * EmailElement Component
+ * React wrapper for a <input type="email"> element.
+ */
+class EmailElement extends Component {
+  /**
+   * @constructor
+   * @param {object} props - React Component properties
+   */
+  constructor(props) {
+    super(props);
+    this.handleChange = this.handleChange.bind(this);
+    this.handleBlur = this.handleBlur.bind(this);
+  }
+
+  /**
+   * Handle change
+   *
+   * @param {object} e - Event
+   */
+  handleChange(e) {
+    this.props.onUserInput(
+      this.props.name,
+      e.target.value,
+      e.target.id, 'textbox',
+    );
+  }
+
+  /**
+   * Handle blur
+   *
+   * @param {object} e - Event
+   */
+  handleBlur(e) {
+    this.props.onUserBlur(this.props.name, e.target.value);
+  }
+
+  /**
+   * Renders the React component.
+   *
+   * @return {JSX} - React markup for the component
+   */
+  render() {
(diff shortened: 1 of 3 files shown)


racostas (approved):
LGTM.

This don't have to be covered in this specific PR but during the testing other related issues we may want to address: 
We should provide a field for "confirm email", also it seems that if the user enter and existing email address nothing warned it about: The process seems to be "working fine" to the user requesting the account, but the admin will never see the request as it is "ignored". Will make an issue with more details.

ridz1208:
@racostas If the email already exists, we purposefully DO NOT display a warning because that would be a security flaw. It would inform a potential "malicious" user that this email exists in the database. So that's not an issue.

for the confirm email part, bring it up at the next Loris meeting

racostas:
> @racostas If the email already exists, we purposefully DO NOT display a warning because that would be a security flaw. It would inform a potential "malicious" user that this email exists in the database. So that's not an issue.
> 
> for the confirm email part, bring it up at the next Loris meeting

... makes some sense. Not sure what could be done if it's an authentic request... or if we should notify somehow the admin who may contact the "original registered email owner " user to check. I'm thinking in users "inactive" for a long period that could have forgot which email they used for registering, etc. In any case, this is more a long term feature. Agree.

AlexandraLivadas (changes requested):
This element should be validated in the backend as well, because validation in the frontend with React is not sufficient.

PR #7847, opened 2021-11-23
[NDB_BVL_Instrument] Add null value to examiner array

CamilleBeau:
## Brief summary of changes
This PR adds an empty value to the array of examiner names. This was mistakenly taken out in #7462, and is causing the first value to be selected by default in instruments.

1. Check that there is an empty option in the examiner drop down of instruments, and make sure that no value is selected by default.

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1085,7 +1085,9 @@ abstract class NDB_BVL_Instrument extends NDB_Page
                 'examinerID'
             );
         }
-        $examiners = [];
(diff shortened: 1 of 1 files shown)


ridz1208:
@CamilleBeau any reason or effect that the removed code was `array(''=>'')` and you are adding `[null=>'']` ?

CamilleBeau:
> @CamilleBeau any reason or effect that the removed code was `array(''=>'')` and you are adding `[null=>'']` ?

There is no reason. I just put `null => ''` because that's what we usually do in instruments on CCNA.

ridz1208 (approved):

PR #7849, opened 2021-11-23
Fix type of physiological electrode X, Y and Z fields to be nullable

cmadjar:
## Brief summary of changes

The values stored in the X, Y and Z columns of the `*electrodes.tsv` BIDS file can be set to `n/a` which cannot be inserted into the current type of the corresponding fields in the `physiological_electrode` table. This PR modifies the default schema and adds a patch to make those columns NULLABLE so the values of X, Y and Z can be set to `NULL` when `n/a` is provided in the BIDS `*electrodes.tsv` file.


cmadjar:
@driusan @ridz1208 I added the critical to release tag here as these SQL changes will be required for LORIS-MRI release 24.0.

@ridz1208 is that a big enough change to add to the CHANGELOG? Happy to add it. Just not sure whether it is relevant or not...

cmadjar:
Blocking LORIS-MRI PR: https://github.com/aces/Loris-MRI/pull/726

AlexandraLivadas (approved):
The patch and the new `CREATE TABLE`  command work as expected. Looks great to me!

cmadjar:
@driusan ready for you. Thank you!

PR #7850, opened 2021-11-23
[user_accounts] Fix Access when only own site permission is set

### LINKED ISSUE #7642: [User Accounts] Page loads with error when site's only permission set

driusan:
Fix access to the user accounts module when the user has only the
"own site" permission. The user_accounts module was returning
strings of CenterIDs instead of objects, this updates it to return
CenterID objects.

Resolves #7642
Replaces/Closes #7764

### DIFF
--- a/modules/user_accounts/php/useraccountrow.class.inc
+++ b/modules/user_accounts/php/useraccountrow.class.inc
@@ -20,9 +20,9 @@ class UserAccountRow implements
      /**
       * Create a new ModuleRow Instance.
       *
-      * @param array $row  The ModuleRow Instance
-      * @param array $cids The CenterIDs
-      * @param array $pids The ProjectIDs
+      * @param array        $row  The ModuleRow Instance
+      * @param \CenterID[]  $cids The CenterIDs
+      * @param \ProjectID[] $pids The ProjectIDs
       */
     public function __construct(array $row, array $cids, array $pids)
     {
@@ -35,7 +35,7 @@ class UserAccountRow implements
     /**
      * Returns the CenterIDs for this row.
      *
-     * @return array The CenterIDs
+     * @return \CenterID[] The CenterIDs
      */
     public function getCenterIDs() : array
     {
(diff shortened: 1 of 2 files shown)


driusan:
@racostas this is the replacement for #7764.. can you review/test?

racostas (commented):
The PR is doing what is intended to showing the users that have only same sites that the one logged in. 
It is failing some automatic test after fixing it should be fine.

Something I notice during the testing that not sure if is a "desired effect" is that every user with this permission (User Accounts: View/Create/Edit User Accounts - Own Sites) will be able to modify the permissions of the admin user as well.

image

I was effectively able to disable the admin account logged as test@test.test 

It's also seems that if one user have more than one site and get edited by one user having other sites (after saving) the "edited user" will be able to have "at most" the sites of the "editing user".

**(This is a bug)**

1. Sites of testrs2 before been edited
image

2. testrs2 been edited by test@test.test.
image

3. testrs2 after been edited.
image

driusan:
@racostas are these new bugs, or is it the existing behaviour from before this change?

racostas:
> @racostas are these new bugs, or is it the existing behaviour from before this change?

As mentioned in the first comment. This should not be introduced by this PR. This should be coexisting. Nevertheless don't  think nobody have been able to detect it as nobody have access before (This PR granted access for first time). 

If you agree, I will approve this PR after the automatics test pass. The issues should be addressed at some point nevertheless. @driusan

driusan:
Can you make an issue for it and tag it as critical to release/discussion required?

racostas (approved):

PR #7853, opened 2021-11-24
[datadict] Need to fetch data after editing Description field

### LINKED ISSUE #7645: [datadict] Need to fetch data after editing Description field
**Describe the bug**
If you edit the Description fields in the data table, the 'Description status' value does not change. 
If you try to filter for 'Description status' 'Modifed', the row that you just modified won't show up. if you cancel the filter, you will see that the changes you just made are no longer there. if you refresh, you will see the changes. this means, that a refetch of data needs to happen after every value change.

**To Reproduce**
1. Go to 'Data Dictionary', make sure you have edit permissions for this module
2. Click on one of the cells in the Description column. Edit the value. The row you just modified should now have a 'Description Status' of 'modified' which can be filtered for in the selection filter form. You will notice that it doesnt i.e. the status does not change and you cannot filter for it.
3. Filter for 'Description Status' = 'Modifed'. The row you just modified isn't included in the query. Click on 'Clear Filter' and find that same row in the data table. You will notice that your changes are gone.
4. Refresh the page, you will see your changes were indeed registered.

Changes made and registered should be reflected on the page immediately.

maltheism:
## Brief summary of changes

Made filter work with new "value" after description cell is modified by the user.

1. Checkout PR
2. visit datadict
3. modify description cell in table
4. search for modified value with filter

* Resolves #7645


suzanne-lee:
Works for the Description field. I think the Description Status field also needs to be updated immediately in order for it to appear in the search query when searching by "Modified".  🙂

cmadjar:
@jeffersoncasimir could you please review and test this PR? Thanks!

jeffersoncasimir:
Tested.

The 'Description' changes are no longer lost when filtering for 'Modified' and then clicking 'Clear Filter'.

However, the 'Description Status' is not being changed to 'Modified' as described in the expected behaviour in issue #7645.

jeffersoncasimir (dismissed):
Please see my comment above

jeffersoncasimir (approved):
Works as described.

I would suggest wrapping it in the condition that the 'Description' actually changed. Accidentally clicking a field and leaving it unchanged will also mark it as 'Modified'.

jeffersoncasimir (approved):
LGTM

PR #7857, opened 2021-11-29
[api] Modify YAML schema so that Swagger Codegen PHP Client works

zaliqarosli:
## Brief summary of changes

This PR:

Fixes:
- [x] remove '#/components/schemas/InstrumentData' because the Data object does not have any explicitly defined properties
- [x] fixes the schema where things are incorrectly defined
- [x] adds 'Other' Sex option now available on LORIS core

Changes:
- [x] modifies $InstrumentShortName key in Instrument API to 'Data' so that the key is static (i don't think there's a need for it to be dynamic depending on the instrument - the key is just the instrument instance data)

1.

### DIFF
--- a/modules/api/static/schema-v0.0.4-dev.yml
+++ b/modules/api/static/schema-v0.0.4-dev.yml
@@ -308,7 +308,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/VisitMeta'
+              $ref: '#/components/schemas/VisitMetaFields'
       responses:
         '201':
           description: The visit was created successfully
--- a/modules/api/static/schema-v0.0.4-dev.yml
+++ b/modules/api/static/schema-v0.0.4-dev.yml
@@ -308,7 +308,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/VisitMetaFields'
+              $ref: '#/components/schemas/VisitMeta'
       responses:
         '201':
           description: The visit was created successfully
@@ -925,7 +925,7 @@ paths:
         - Images
       summary: Get the qc values of an image
       parameters:
-        - name: candid
+        - name: id
           in: path
           description: The candidate identifier. Either PSCID or CandID.
           required: true
@@ -1288,7 +1288,7 @@ paths:
         - Dicoms
       summary: List all dicoms for that timepoint
       parameters:
-        - name: candid
+        - name: id
           in: path
           description: ID of the candidate. Either PSCID or CandID
           required: true
@@ -1852,7 +1852,9 @@ components:
       type: object
       properties:
         Candidates:
-          $ref: '#/components/schemas/Candidate'
+          type: array
+          items:
+            $ref: '#/components/schemas/Candidate'
     Candidate:
       type: object
       properties:
@@ -1873,6 +1875,7 @@ components:
           enum:
             - Female
             - Male
+            - Other
     NewCandidate:
       type: object
       properties:
@@ -1944,8 +1947,10 @@ components:
       properties:
         Meta:
           $ref: '#/components/schemas/InstrumentMeta'
-        $InstrumentShortName:
-          $ref: '#/components/schemas/InstrumentData'
+        Data:
+          type: object
+          additionalProperties:
+            type: string
     InstrumentMeta:
       type: object
       properties:
@@ -1957,10 +1962,6 @@ components:
           type: string
         DDE:
           type: boolean
-    InstrumentData:
-      type: object
-      additionalProperties:
-        type: string
     InstrumentFlags:
       type: object
       properties:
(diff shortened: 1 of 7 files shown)


modules/api/static/schema-v0.0.4-dev.yml
```diff
@@ -79,13 +79,20 @@ components:
...
             Stages:
               $ref: '#/components/schemas/InstrumentVisit'
-
     InstrumentVisit:
       type: object
       properties:
         Visit:
           $ref: '#/components/schemas/VisitStage'
-
+    Instrument:
```
zaliqarosli, line 87: @xlecours i don't fully understand how this file works to update the schema. is just redefining the component schema like this enough?

xlecours (approved):

ridz1208:
@zaliqarosli after speaking to @driusan the only remaining issues on this PR is that the instrument.class.inc endpoint is being modified which impacts all API versions.

@xlecours could you instruct @zaliqarosli how to deal with that? I think we were creating a copy of the class and tagging it with the api version... is that it? like `instrument-0.0.4.class.inc` or something ?

modules/api/static/schema-v0.0.4-dev.yml
```diff
@@ -2492,13 +2492,20 @@ components:
...
```
xlecours, line 2500: Instrument is already define at line 1942

modules/api/static/schema.yml
```diff
@@ -1823,7 +1823,9 @@ components:
       type: object
       properties:
         Candidates:
-          $ref: '#/components/schemas/Candidate'
+          type: array
```
xlecours, line 1826: Can you apply this to modules/api/static/schema-v0.0.4-dev.yml as well? please

xlecours (changes requested):

modules/api/static/schema.yml
zaliqarosli: hey @xlecours thanks for the review! i added other fixes here to the v0.0.4-dev schema as well

xlecours (approved):
@zaliqarosli While I was trying the schema-v0.0.4-dev.yml file in swagger editor, it complained about multiple entries in the components that are defined twice. Would you mind fixing these in this PR as well?  
 ex:  VisitStage, InstrumentVisit, ...
 
 I'll approve the PR in its current stage anyways.
 Thanks

modules/api/static/schema-v0.0.4-dev.yml
```diff
@@ -925,7 +925,7 @@ paths:
         - Images
       summary: Get the qc values of an image
       parameters:
-        - name: candid
+        - name: id
```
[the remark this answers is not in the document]
xlecours: as long as it's consistent with what is in the path (in the yml file) it's fine. But I would keep candid to follow what is there in the other endpoints definitions.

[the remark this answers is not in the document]
zaliqarosli: the rest of the definitions are id so i'll stick to that

zaliqarosli:
@xlecours i am getting these failed tests

```
There were 4 failures:

1) LorisApiInstrumentsTest::testPatchCandidatesCandidVisitInstrumentsInstrument
Failed asserting that 200 matches expected 204.

/app/raisinbread/test/api/LorisApiInstrumentsTest.php:120

2) LorisApiInstrumentsTest::testPutCandidatesCandidVisitInstrumentsInstrument
Failed asserting that 200 matches expected 204.

/app/raisinbread/test/api/LorisApiInstrumentsTest.php:147

3) LorisApiInstrumentsTest::testPatchCandidatesCandidVisitInstrumentsInstrumentDde
Failed asserting that 200 matches expected 204.

/app/raisinbread/test/api/LorisApiInstrumentsTest.php:370

4) LorisApiInstrumentsTest::testPutCandidatesCandidVisitInstrumentsInstrumentDde
Failed asserting that 200 matches expected 204.

/app/raisinbread/test/api/LorisApiInstrumentsTest.php:403
```

which I don't understand as the instrument endpoint is clearly returning 204 No Content, and I'm not sure where the 200 is coming from. do you have any ideas?

zaliqarosli:
@ridz1208 @driusan @xlecours can/should this go to the release branch?

xlecours:
@ridz1208 Do you mean 24.1 or the next release? 
Adding specification for `Other` Sex value is a documentation fix that could be applied to 24.1 but changing the variable property name to fixed `Data` is a new feature.

ridz1208:
I agree with @xlecours but isnt that the reason the API is versioned separately than LORIS? as long as its going to 0.0.4-dev it should be fine if it goes in 24.1 as far as I understand

driusan:
@ridz1208 yes but this PR is changing 0.0.3

driusan:
nm, just saw that it's only the "other" part going to 0.0.3.

PR #7898, opened 2021-12-07
Remove dataquery from the modules table in the default schema

cmadjar:
## Brief summary of changes

This removes the `dataquery` module from the `modules` database table. Note: there is already a [patch](https://github.com/aces/Loris/blob/main/SQL/New_patches/2021-09-07_dqt_drop_old_dataquery.sql) taking care of that but the default schema was not up to date.

Note: RB `modules` table does not contain `dataquery` so no need to update RB data here.


cmadjar:
@maltheism would you want to review this PR by any chance? Thanks!

maltheism (approved):

cmadjar:
thank you @maltheism :)

@driusan ready for you!

PR #7902, opened 2021-12-07
Update install order and add npm dependency to the list of dependencies

cmadjar:
## Brief summary of changes

While testing the install, I noticed the order in the installation instructions were off. Should first create the `lorisadmin` user and `/var/www/loris` and then download the source code to move it to `/var/www/loris`.

Also, NPM is missing in the prerequisites.


cmadjar:
@ridz1208 maybe you can take a look at this one and make sure the changes I made make sense to you. Thanks!

PR #7908, opened 2021-12-09
[Login] Request account with bad email UI seems to work

### LINKED ISSUE #7897: [Login] Request account with bad email UI seems to work
**Describe the bug**
When requesting an account using an invalid email address (ex. bad@email) the UI shows that the account was created. The response from the BE was a 409 Conflict instead of a 400 bad request.

**To Reproduce**
1. Go to login page
2. Click on request account 
3. Fill out all fields with the email field as `bad@email`
4. Submit form

An error to show that the email is not valid.

- OS: Mac
- Browser Chrome

Loris Test VM

**Additional context**

maltheism:
## Brief summary of changes

Updates the email component to use the pattern property for the email input. Also using the type property for the message to displayed to the user when email isn't matching the pattern property. See here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email

edit: I added frontend logic for if by any chance the php validation finds an error not caught by the frontend and so the user will get a notification that email is invalid.

1. Checkout PR
2. Attempt request account with bad email ex. "bad@email"
3. User should not be able to use bad email.

* Resolves #7897

### DIFF
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -1295,6 +1295,10 @@ class EmailElement extends Component {
         <div className={inputClass}>
           <input
             type="email"
+            pattern={'/^[a-zA-Z0-9.!#$%&\'*+\\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[' +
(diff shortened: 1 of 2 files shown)


cmadjar (approved):
Works great 👍

PR #7913, opened 2021-12-13
[user_accounts] Random "none" on the page when adding new user

### LINKED ISSUE #7912: [user_accounts] Random "none" on the page when adding new user
**Describe the bug**
The `None` displayed on the page (see screenshot) is normally adjacent to the `Account Request Date` label. for some reason on a new user the label does not show but the "none" does

Screen Shot 2021-12-10 at 4 33 17 PM

maltheism:
## Brief summary of changes

Updates the logic for `Account Request Date` to always be displayed. Unsure if we want to hide it for administrators. @ridz1208 ?

1. See issue.

* Resolves #7912

### DIFF
--- a/modules/user_accounts/templates/form_edit_user.tpl
+++ b/modules/user_accounts/templates/form_edit_user.tpl
@@ -397,6 +397,7 @@
                 {/if}
             </div>
       {/if}
+      {if $form.Active|default}
       <div class="row form-group form-inline">
            <label class="col-sm-2">
                {$form.Active.label|default}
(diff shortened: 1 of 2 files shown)


modules/user_accounts/templates/form_edit_user.tpl
```diff
@@ -397,6 +397,7 @@
                 {/if}
             </div>
       {/if}
+      {if $form.Active|default}
```
maltheism, line 400: This changes was made because I noticed while displaying an admin account that an extra invisible div taking up space was being made for "Active" while Active isn't meant to be displayed for admin accounts.

ridz1208 (approved):

PR #7914, opened 2021-12-13
[User Accounts] A user with the permission "View/Create/Edit User Accounts - Own Sites" could be unintentionally restrict the access of other users to the sites it doesn't have access to.

### LINKED ISSUE #7855: [User Accounts] A user with the permission "View/Create/Edit User Accounts - Own Sites" could be unintentionally restring the access to other users to the sites it don't have access to.
**Describe the bug**
A user with the permission `View/Create/Edit User Accounts - Own Sites` could be unintentionally restring the access to other users to the sites it don't have access to.

**To Reproduce**
1. Two users are need for reproduce this issue (this is an example with the current RB data):
- `test@test.test ` is a user that have the permission  `View/Create/Edit User Accounts - Own Sites` and only have access to the site Ottawa.
-  `testrs2@testrs2.com` is a user that have two or more sites (in this example: Ottawa and Montreal). At least one of the sites have to coincide with the site of the first user (but not all because the issue could remain hidden then) 
2. Logged as `test@test.test`  and `go to Admin->User Accounts.`
3. Notice that the user `testrs2@testrs2.com` have access to two sites.
image
3. Click to edit the user `testrs2@testrs2.com`
4. You will notice in the filed "Sites" the only site listed is the one both have in common **( This is an expected behavior)**
image
5. Modify anything in the user you're editing and save.
6. Go back to `Admin->User Accounts.`
7. You will notice the user `testrs2@testrs2.com` have lost access to it other sites. In this case Montreal.
image

The "edited user" to keep its other sites.

- OS: Microsoft Windows 10.
- Browser: Chrome, Version 95.0.4638.54 (Official Build) (64-bit)

- LORIS Version: 24.0 testing

maltheism:
## Brief summary of changes

Here's an attempt at solving the bug of when a user has permissions to edit users for "Own Sites" and where the editor can possibly remove sites from a user that has additional sites that the editor modifying does not.

See issue created for testing instructions if not clear from summary.

* Resolves #7855

### DIFF
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -298,7 +298,7 @@ class Edit_User extends \NDB_Form
             }
 
             $DB->delete('user_psc_rel', ["UserID" => $uid]);
-            foreach ($us_curr_sites as $site) {
+            foreach ($userNewCenterIDs as $CenterID) {
                 $DB->insert(
                     'user_psc_rel',
                     [
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -273,16 +273,18 @@ class Edit_User extends \NDB_Form
         $uid           = $user->getId();
         $us_curr_sites = $values['CenterIDs'];
         $userNewCenterIDs = array_map(
-            function($val) {return new \CenterID($val);},
+            function ($val) {
+                return new \CenterID($val);
+            },
             $values['CenterIDs']
         );
         if (!$this->isCreatingNewUser()) {
             // Prevent deletion of rules/sites editor may not have access to edit.
             if (!$editor->hasPermission('user_accounts_multisite')) {
                 // Collect CenterID(s) of user being modified.
-                $userCenterIDs       = $user->getCenterIDs();
+                $userCenterIDs = $user->getCenterIDs();
                 // Collect CenterID(s) of the editor modifying sites.
-                $editorCenterIDs       = $editor->getCenterIDs();
+                $editorCenterIDs = $editor->getCenterIDs();
                 // If the editor does not have access to all of the
                 // user's sites the difference between the two arrays
                 // will be the sites we additionally need to add to
@@ -290,14 +292,14 @@ class Edit_User extends \NDB_Form
                 $differenceOfCenterIDs = array_udiff(
                     $userCenterIDs,
                     $editorCenterIDs,
-                    function($userCenterID, $editorCenterID) {
+                    function ($userCenterID, $editorCenterID) {
                         return strcmp(
                             $userCenterID,
                             $editorCenterID
                         );
                     }
                 );
-                $userNewCenterIDs         = array_merge(
+                $userNewCenterIDs      = array_merge(
                     $userNewCenterIDs,
                     $differenceOfCenterIDs
                 );
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -288,8 +288,8 @@ class Edit_User extends \NDB_Form
                 // will be the sites we additionally need to add to
                 // the user being modified.
                 $differenceOfCenterIDs = array_diff(
-                    $userCenterIDsValues,
-                    $editorCenterIDsValues
+                    $userCenterIDs,
+                    $editorCenterIDs
                 );
                 $us_curr_sites         = array_merge(
                     $us_curr_sites,
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -273,6 +273,29 @@ class Edit_User extends \NDB_Form
         $uid           = $user->getId();
         $us_curr_sites = $values['CenterIDs'];
         if (!$this->isCreatingNewUser()) {
+            // Prevent deletion of rules current user may not have access to edit.
+            if (!$editor->hasPermission('user_accounts_multisite')) {
+                // Collect CenterID(s) of user being modified.
+                $userCenterIDs = array_values($DB->pselect('SELECT CenterID FROM user_psc_rel WHERE UserID=:uid', [
+                    'uid' => $uid
+                ]));
+                // Collect CenterID(s) of the editor modifying sites.
+                $editorCenterIDs = $editor->getCenterIDs();
+                $userCenterIDsValues   = [];
+                $editorCenterIDsValues = [];
+                foreach ($userCenterIDs as $userCenterID) {
+                    $userCenterIDsValues[] = $userCenterID['CenterID'];
+                }
+                foreach ($editorCenterIDs as $editorCenterID) {
+                    $editorCenterIDsValues[] = (string) $editorCenterID;
+                }
+                // If the editor does not have access to all of the user's sites
+                // the difference between the two arrays will be the sites we additionally
+                // need to add to the user.
+                $differenceOfCenterIDs = array_diff($userCenterIDsValues, $editorCenterIDsValues);
+                $us_curr_sites = array_merge($us_curr_sites, $differenceOfCenterIDs);
+            }
+
             $DB->delete('user_psc_rel', ["UserID" => $uid]);
             foreach ($us_curr_sites as $site) {
                 $DB->insert(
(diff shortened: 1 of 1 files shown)


modules/user_accounts/php/edit_user.class.inc
```diff
@@ -273,6 +273,41 @@ class Edit_User extends \NDB_Form
         $uid           = $user->getId();
         $us_curr_sites = $values['CenterIDs'];
```
ridz1208, line 274:
```suggestion
        $us_curr_sites = $values['CenterIDs'];
        $userNewCenterIDs = array_map(
            function($val) {return new \CenterID($val);},
            $values['CenterIDs']
        );
```

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -273,6 +273,41 @@ class Edit_User extends \NDB_Form
...
+            if (!$editor->hasPermission('user_accounts_multisite')) {
+                // Collect CenterID(s) of user being modified.
+                $userCenterIDs = array_values(
+                    $DB->pselect(
+                        'SELECT CenterID FROM user_psc_rel WHERE UserID=:uid',
+                        [
+                            'uid' => $uid
+                        ]
+                    )
+                );
```
ridz1208, line 286:
```suggestion
                $userCenterIDs       = $user->getCenterIDs();
```

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -273,6 +273,41 @@ class Edit_User extends \NDB_Form
...
+                // Collect CenterID(s) of the editor modifying sites.
+                $editorCenterIDs       = $editor->getCenterIDs();
+                $userCenterIDsValues   = [];
+                $editorCenterIDsValues = [];
+                foreach ($userCenterIDs as $userCenterID) {
+                    $userCenterIDsValues[] = $userCenterID['CenterID'];
+                }
+                foreach ($editorCenterIDs as $editorCenterID) {
+                    $editorCenterIDsValues[] = (string) $editorCenterID;
+                }
```
ridz1208, line 296:
```suggestion
```

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -273,6 +273,41 @@ class Edit_User extends \NDB_Form
...
+                // If the editor does not have access to all of the
+                // user's sites the difference between the two arrays
+                // will be the sites we additionally need to add to
+                // the user being modified.
+                $differenceOfCenterIDs = array_diff(
+                    $userCenterIDsValues,
+                    $editorCenterIDsValues
```
ridz1208, line 303:
```suggestion
                    $userCenterIDs,
                    $editorCenterIDs
```

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -273,6 +273,41 @@ class Edit_User extends \NDB_Form
...
+                );
+                $us_curr_sites         = array_merge(
+                    $us_curr_sites,
```
ridz1208, line 306:
```suggestion
                    $userNewCenterIDs,
```

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -273,6 +273,41 @@ class Edit_User extends \NDB_Form
...
+                }
+                // If the editor does not have access to all of the
+                // user's sites the difference between the two arrays
+                // will be the sites we additionally need to add to
+                // the user being modified.
+                $differenceOfCenterIDs = array_diff(
+                    $userCenterIDsValues,
+                    $editorCenterIDsValues
+                );
+                $us_curr_sites         = array_merge(
```
ridz1208, line 305:
```suggestion
                $userNewCenterIDs         = array_merge(
```

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -273,6 +273,41 @@ class Edit_User extends \NDB_Form
...
+                    $us_curr_sites,
+                    $differenceOfCenterIDs
+                );
+            }
+
             $DB->delete('user_psc_rel', ["UserID" => $uid]);
             foreach ($us_curr_sites as $site) {
```
ridz1208, line 312:
```suggestion
            foreach ($userNewCenterIDs as $CenterID) {
```

ridz1208 (changes requested):
@maltheism The logic seems proper, the code is however overly complex I think. I made some suggestions and went down the route of using the array of CenterIDs instead of converting it before the diff. I'm not entirely sure if it will work but try it out and let me know. This might also break some static checks and PHPcs that might need to be fixed

*Important* Github would not let me make a suggestion on line 317 change `$site` to `$CenterID->__toString()`

maltheism:
@ridz1208 array_diff doesn't seem to work for your changes.

ridz1208:
@driusan @xlecours an example of somewhere where storing a list of instances would have been handy (array_diff would catch the proper differences if the objects were actually the same istead of 2 objects with the same value)

ridz1208:
@maltheism its working fine for me !

xlecours:
> @driusan @xlecours an example of somewhere where storing a list of instances would have been handy (array_diff would catch the proper differences if the objects were actually the same instead of 2 objects with the same value)

What was not working with array_diff?  Our CenterID class has a __toString function and according to [ doc](https://www.php.net/manual/en/function.array-diff.php): 
```
Two elements are considered equal if and only if (string) $elem1 === (string) $elem2. That is, when the string representation is the same.
```

maltheism:
Hi @xlecours, I used array_udiff instead because array_diff wasn't supporting two arrays containing CenterID objects with string values. Previously @ridz1208 found the code overly complex when I was using array_diff after creating an array of strings from looping through the array containing CenterID objects. I think array_udiff is preferable in the end and I was somewhat confused yesterday that it wasn't working but it was.

ridz1208:
From @xlecours, in curious to try it out again with array_diff now.

@xlecours initially we thought that array_diff was not working because it was comparing the objects themselves instead of the string value of each object. But since we found that our testing was incorrect but we had already implemented array_udiff.

@maltheism can you revert to array_diff and give it a try? The documentation is not super clear, it says array_diff will use the string values for compare for "non-matching types" in our case we have matching types but they are objects so I'm curious to know if it would work

maltheism:
Hi @ridz1208 & @xlecours, turns out it works and I pushed the changes.

ridz1208 (approved):

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -272,14 +272,40 @@ class Edit_User extends \NDB_Form
...
+                // Collect CenterID(s) of user being modified.
+                $userCenterIDs = $user->getCenterIDs();
+                // Collect CenterID(s) of the editor modifying sites.
+                $editorCenterIDs = $editor->getCenterIDs();
+                // If the editor does not have access to all of the
+                // user's sites the difference between the two arrays
+                // will be the sites we additionally need to add to
+                // the user being modified.
+                $differenceOfCenterIDs = array_diff(
+                    $userCenterIDs,
```
xlecours, line 293: `$userCenterIDs` should only contain values belonging to the editor. It should be `array_intersect($userNewCenterIDs, $editorCenterIDs);`

xlecours (dismissed):

modules/user_accounts/php/edit_user.class.inc
ridz1208: @xlecours are you suggesting that we should modify the user class' `getCenterIDs()` function to only return the sites that the editor has access to? (that's what im understanding here)

xlecours: No.

I made my comment at the wrong place, mentioning the wrong variable... my mistake.

What I meant was:
Because `$userNewCenterIDs` is basically the values received in the POST request, it could be any values if a malicious user forge the request. So, to prevent that user from adding sites not belonging to the editor, we should make sure that the final set of sites to add (`$userNewCenterIDs`) is composed of A) the sites the editor can't modify that already belong to the editee (that is `$differenceOfCenterIDs`), and B) the sites that were POSTed that must belong to the editor (that is the modification I am asking).

Something like:
```
                $userNewCenterIDs      = array_merge(
                    array_intersect($userNewCenterIDs, $editorCenterIDs)
                    $differenceOfCenterIDs
                );
``` 

Sorry for the confusion.

CamilleBeau:
I was not able to successfully test this PR because of [this unrelated issue](https://github.com/aces/Loris/issues/7931). The code looks good but I will wait until Xavier's comment is addressed to re-review / test.

modules/user_accounts/php/edit_user.class.inc
maltheism: thanks @xlecours

maltheism:
@CamilleBeau it's ready i think

CamilleBeau:
Looks good!

CamilleBeau (approved):

PR #7927, opened 2021-12-17
[Battery] Fix error when multiple rows of the same instrument match

### LINKED ISSUE #7130: [23.0][battery_manager] Issue with test battery assignnment for duplicate tests
**Describe the bug**
If you add the same instrument to a timepoint (changing the min/max dates or another parameter, but keeping test and timepoint the same), you get duplicated entries in the instrument list. For example, in the screenshot below. Test was added twice, but appears 4 times in the instrument list. Duplicated commentIDs.

image

**To Reproduce**
1. Go to `Battery Manager`
2. Add a test.
3. Add the same test again to the same timepoint with a different max age.
4. Start a new visit for that timepoint and observe the error in the test battery.

No duplication.

**Additional notes**
Not even sure we want to be able to add the same test to the timepoint anyways regardless of the age min/max being different ? Should this simply be flagged as attempting to add the same instrument to the battery and thus not processed?

ridz1208:
## Brief summary of changes
2 Issues are fixed By the following changes, both related to the fact that when multiple entries exist in the test_battery that have different attributes (not duplicates) but all match the lookup at a specific timepoint the next_stage and instrument_list modules fail.

Issue 1: The next stage module calls on the battery class to populate the flag table when the visit is started. When multiple entries of the same instrument are returned from the lookup query, the battery class attempts to create both which triggers a duplicate error in MySQL

Issue 2: The instrument_list module uses the battery class' `getBatteryVerbose` function to list the instruments properly grouped and ordered. the issue occurs when multiple entries of the same instrument are found in the lookup results. This PR removes the duplication of the instrument in the front end when that happens

1. Choose a battery entry and duplicate it
2. change a few parameters which still allow the entry to overlap with another (change age range for example)
3. create a new timepoint with paramters that match the battery entry (subproject, visit_label,...) and start it
4. Without this PR, the next stage module will give a 500 error when populating the battery
5. navigate to the instrument_list of the new timepoint. notice that without these changes, the instrument is displayed twice on the page.

* Resolves #7130

### DIFF
--- a/php/libraries/NDB_BVL_Battery.class.inc
+++ b/php/libraries/NDB_BVL_Battery.class.inc
@@ -318,14 +318,14 @@ class NDB_BVL_Battery
         $this->_assertBatterySelected();
 
         // craft the select query
-        $query = "SELECT DISTINCT f.Test_name,
+        $query = "SELECT f.Test_name,
                       t.Full_name,
                       f.CommentID,
                       CONCAT('DDE_', f.CommentID) AS DDECommentID,
                       ts.Subgroup_name as Sub_group,
                       ts.group_order as Subgroup_order,
                       t.isDirectEntry,
-                      b.instr_order as instrument_order
+                      MAX(b.instr_order) as instrument_order
                   FROM flag f
             	      JOIN test_names t ON (f.Test_name=t.Test_name)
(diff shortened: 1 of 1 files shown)


jesscall:
looks good! 

Tested with the use case that was failing in #7130 and the error was not reproduced.

PR #7932, opened 2022-01-03
[webpack.config.js] update dependencies for copy-webpack-plugin

maltheism:
## Brief summary of changes

Updates package.json for latest version of copy-webpack-plugin.

1. make clean
2. make dev
3. navigate around Loris

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -198,7 +198,6 @@ const config = [
           {
             from: path.resolve(__dirname, 'node_modules/react/umd'),
             to: path.resolve(__dirname, 'htdocs/vendor/js/react'),
-            flatten: true,
             force: true,
             globOptions: {
               ignore: ['react.profiling.min.js'],
@@ -215,7 +214,6 @@ const config = [
           {
             from: path.resolve(__dirname, 'node_modules/react-dom/umd'),
             to: path.resolve(__dirname, 'htdocs/vendor/js/react'),
-            flatten: true,
             force: true,
             filter: async (path) => {
               const file = path.split('\\').pop().split('/').pop();
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "@babel/plugin-transform-runtime": "^7.10.5",
     "@babel/runtime": "^7.10.5",
     "@fortawesome/fontawesome-free": "^5.11.2",
-    "copy-webpack-plugin": "^6.2.1",
+    "copy-webpack-plugin": "^11.0.0",
     "jstat": "^1.9.5",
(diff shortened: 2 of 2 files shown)


webpack.config.js
```diff
@@ -125,11 +132,12 @@ function lorisModule(mname, entries, override=false) {
...
       'react-dom': 'ReactDOM',
     },
-    node: {
-      fs: 'empty',
-    },
     devtool: 'source-map',
-    plugins: [],
+    plugins: [
+      new webpack.DefinePlugin({
+        'process.env.NODE_ENV': JSON.stringify('development'),
```
maltheism, line 138: Solves the following error in the browser:
`Webpack: Bundle.js - Uncaught ReferenceError: process is not defined`
See https://webpack.js.org/migrate/5/ and https://stackoverflow.com/questions/41359504/webpack-bundle-js-uncaught-referenceerror-process-is-not-defined

webpack.config.js
```diff
@@ -1,21 +1,23 @@
...
   minimizer: [
-    new TerserPlugin({
-      cache: true,
-      parallel: true,
-      terserOptions: {
-        compress: false,
-        ecma: 6,
-        mangle: false,
-      },
-      sourceMap: true,
```
maltheism, line 17: I believe that `devtool: 'source-map'` handles the `sourceMap: true` from looking at the latest terser-webpack-plugin documentation.
```
[webpack-cli] Invalid options object. Terser Plugin has been initialized using an options object that does not match the API schema.
54
 - options has an unknown property 'sourceMap'. These properties are valid:
55
   object { test?, include?, exclude?, terserOptions?, extractComments?, parallel?, minify? }
```

webpack.config.js
```diff
@@ -1,21 +1,22 @@
...
 const TerserPlugin = require('terser-webpack-plugin');
 const CopyPlugin = require('copy-webpack-plugin');
+const webpack = require('webpack');
 const path = require('path');
 const fs = require('fs');
 
 const optimization = {
   minimizer: [
-    new TerserPlugin({
-      cache: true,
```
maltheism, line 10:
```
5.0.0 (2020-10-14) ⚠ BREAKING CHANGES
* webpack@4 no longer supported
* removed the cache option (respect the cache option by default from webpack)
* removed the cacheKeys option (respect the cache option by default from webpack)
* removed the sourceMap option (respect the devtool option by default from webpack)
Bug Fixes
* reduce deps
* performance improvement
```

maltheism:
I rebased. Hi @driusan this PR is ready again.

driusan (approved):

PR #7933, opened 2022-01-03
[user_accounts] Prevent password autofill in chrome when editing or creating an account

ridz1208:
## Brief summary of changes
I was getting tired of chrome autofilling the password field when editing a user in the user accounts module. I followed the instructions [here](https://stackoverflow.com/questions/12374442/chrome-ignores-autocomplete-off) to get rid of that and it seems to work

Issue:
Using chrome, if you have a password saved for user **X** on the loris instance you are using and you navigate to the user_accounts module and edit user **X**, Chrome decides to autofill the "password" field (I think simply because it's called password). If you are not mindful of the autofill and try to modify some permission of other parameter and hit save you get an error (see images)

Autofill on page load:
image

Error after saving without clearing the password field:
image

This PR should simply prevent chrome from being annoying

~#firstPRin2022~ #secondPRin2022 (@maltheism beat me by 38 seconds)

### DIFF
--- a/modules/user_accounts/templates/form_edit_user.tpl
+++ b/modules/user_accounts/templates/form_edit_user.tpl
@@ -71,7 +71,7 @@
             <span class="pwd-star password {if isset($form.Password_hash.required) && $form.Password_hash.required} required{/if}" style="color: red">*</span>
         </label>
         <div class="col-sm-3">
-            <input type="password" name="{$form.Password_hash.name}" />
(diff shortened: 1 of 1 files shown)


maltheism (dismissed):

maltheism (approved):
Everything looks good! Maybe this PR will be the first to be merged for 2022 @ridz1208

PR #2414, opened 2016-11-15
Ajax login

jstirling91:
As we move towards using more and more react, we are tending to use more ajax calls instead of page loads. This can become problematic when the users session has timed out and a 401 error is returned from the serve. Currently our pages don't handle this error accordingly.

To fix the issue, a generic error handler is introduced in this pull request to handle all 401 errors when using ajax. It works by having a pop out login for users to login without losing their current work the have been doing in the various react modules. 

This current implementation doesn't resend the failed ajax call, users will have to trigger the call again but atleast they won't loose all their work they did.

PR #7944, opened 2022-01-04
[DQT] Prevent Autofill in chrome

### LINKED ISSUE #7938: [DQT] Prevent Autofill in chrome
**Describe the bug**
Several fields in the DQT module trigger the chrome username/password autofill (see screenshots) the currently identified locations are

 1. in define filters -> select a category -> select a field -> select an operator -> AUTOFILL username
    Screen Shot 2022-01-04 at 10 27 56 AM
 2. in the query SAVING functionality, the query name field is autofille
    Screen Shot 2022-01-04 at 10 26 59 AM
d

maltheism:
## Brief summary of changes

Prevents autofill in chrome being incorrectly applied throughout LORIS like in the DQT module.

1. Checkout PR.
2. make clean && make dev
3. Visit the DQT and examine fields in chrome if passwords are being autofilled.

* Resolves #7938

### DIFF
--- a/smarty/templates/main.tpl
+++ b/smarty/templates/main.tpl
@@ -296,35 +296,5 @@
         {if $FormAction}
         </form>
         {/if}
-
-        <a id="login-modal-button" href="#" data-toggle="modal" data-target="#login-modal" style="display: none;">Login</a>
-
-        <div class="modal fade" id="login-modal" tabindex="-1" role="dialog" aria-labelledby="myModalLabel" aria-hidden="true" style="display: none;">
-            <div class="modal-dialog">
-                <div class="modal-content">
-                    <div class="modal-header">
-                        Login to Your Account
-                    </div>
-                    <div class="modal-body">
-                        <div class="row">
-                            <div class="col-xs-12">
-                                <font color="red" align="middle" id="login-modal-error" style="display: none;">
-                                    Incorrect username or password
-                                </font>
-                            </div>
-                            <div class="form-group col-xs-12">
-                                <input id="modal-username" name="username" class="form-control" type="text" value="" placeholder="User">
-                            </div>
-                            <div class="form-group col-xs-12">
-                                <input id="modal-password" name="password" class="form-control" type="password" placeholder="Password">
-                            </div>
-                            <div class="form-group col-xs-12">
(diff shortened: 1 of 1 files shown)


smarty/templates/main.tpl
```diff
@@ -296,35 +296,5 @@
         {if $FormAction}
         </form>
         {/if}
-
```
maltheism, line 299: Hi @ridz1208, I think this extra html that was being added to Loris is the problem.

ridz1208:
@maltheism at first sight, this seems like a very risky change (and the bug it's causing is not that critical) so it will require a much deeper dive than I expected. Also, if the cause is `main.tpl` how come free text fields in instruments and other modules don't trigger the autofill ?

maltheism:
@ridz1208 I don't believe it's a risky change because the login is handled by the login module that's been reactified. I've been reading about the google autofill problem and it seems that if Chrome finds a field that has name="password" then it might assume fields rendered before the password field could possibly be password fields.

ridz1208:
@jstirling91 the code removed was added by you once upon a time. Can you confirm what @maltheism is saying ? 

reference PR https://github.com/aces/Loris/pull/2414

ridz1208:
closing because main.tpl will be entirely removed in the next iteration

maltheism:
Reopening after slack discussion @ridz1208. I think the code in this PR is safe to be removed. It isn't used anywhere from what I can see looking in the codebase of Loris.

maltheism:
@ridz1208 needed

smarty/templates/main.tpl
driusan: A comment on #8309 refreshed my memory of how this is supposed to work and why it's not working. This "extra" HTML is used by .ajaxError to trigger a popup modal after a timeout where the user can log back in without losing data.

It's not working anymore because it was using `$(document).ajaxError` and we're using the fetch API instead of jQuery in most places now.

From what I can tell, there's no way to globally hook in to the fetch API to check for 403 errors like this was doing so it's probably safe to remove (https://stackoverflow.com/questions/43594266/fetch-equivalent-to-jquerys-document-ajaxerror) this HTML since we're not going back to jQuery's `$.ajax`.

Can you remove this JS code too which will now no longer work even for `$.ajax`? https://github.com/aces/Loris/blob/main/htdocs/js/helpHandler.js#L98-L123?

PR #7945, opened 2022-01-05
[DQT] Results header is stuck on initial query run when scrolling down and won't update when changing display modality (Cross sectional vs Longitudinal)

### LINKED ISSUE #7929: [DQT] Results header is stuck on initial query run when scrolling down and won't update when changing display modality (Cross sectional vs Longitudinal)
**Describe the bug**
For the Data Query Tool, when a query is run, the headers on the results table are consistent with the display modality selected, either cross sectional or longitudinal. However, if the display modality is change, the headers will change in the results table accordingly, but when scrolling down, the headers will revert back to those under which the query was initially run. This happens independent of initial modality. When scrolling down, the headers will thus not correspond to the values in the columns as these will be shifted.

**To Reproduce**
1. Go to the Data Query Tool and create a query. The run the query in either 'Cross-sectional' or 'Longitudinal' modality.
2. Click on the alternate display modality.
3. Scroll down the results table.
4. Headers will revert to those of the initial query results (see screenshots).

Screen Shot 2021-12-20 at 5 45 27 PM

Screen Shot 2021-12-20 at 5 45 41 PM

maltheism:
## Brief summary of changes

Fixed the issue where the component update wasn't refreshing the dynamictable's floating headers. The dynamictable's floating headers visible to the user when scrolling.

1. Checkout PR
2. make clean && make dev
3. visit the DQT module
4. run a query and see if the floating headers when scrolling update after changing back & forth from (Cross sectional vs Longitudinal).

* Resolves #7929

### DIFF
--- a/jsx/StaticDataTable.js
+++ b/jsx/StaticDataTable.js
@@ -72,7 +72,6 @@ class StaticDataTable extends Component {
         $('#dynamictable').find('tbody td:eq(0)').hide();
       }
     }
-
     if (!this.props.DisableFilter) {
       // Retrieve module preferences
       let modulePrefs = JSON.parse(localStorage.getItem('modulePrefs'));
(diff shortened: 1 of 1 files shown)


ridz1208 (changes requested):
Seems to be working, just needs tweaking

scrolling down the headers look fine but if the table many columns once you scroll the columns and headers no longer align
Screenshot from 2022-01-19 13-59-06
Screenshot from 2022-01-19 13-59-37

@maltheism if the changes to fix this are too extensive, we can maybe ignore that issue for now. let's just see if its easily fixable

ridz1208 (approved):

PR #7952, opened 2022-01-13
[Media] Content does not update after a new file is uploaded

### LINKED ISSUE #5055: [Media] Content does not update after a new file is uploaded
After uploading a file, going back to the browse tab does not refresh/reload the data with the uploaded file, and the page must be manually refreshed to load the new content. Similarly, when trying to upload the same content without refreshing the page it does not detect that the file has been already uploaded, and so does not prompt to confirm overwriting the file.

From redmine https://redmine.cbrain.mcgill.ca/issues/14559

maltheism:
## Brief summary of changes

Made the upload file process better.

1. Checkout PR
2. Upload file
3. Observe if user experience is okay.

* Resolves #5055

### DIFF
--- a/modules/media/jsx/uploadForm.js
+++ b/modules/media/jsx/uploadForm.js
@@ -318,7 +318,15 @@ class MediaUploadForm extends Component {
           formData: {}, // reset form data after successful file upload
           uploadProgress: -1,
         });
-        swal.fire('Upload Successful!', '', 'success');
+        swal.fire(
+          'Success!',
+          'Upload of media file completed.',
+          'success'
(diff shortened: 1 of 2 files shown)


cmadjar (approved):
👍

modules/media/jsx/uploadForm.js
```diff
@@ -348,7 +356,7 @@ class MediaUploadForm extends Component {
       });
       swal.fire(msg, '', 'error');
     }, false);
-
+    console.log(this.props.action);
```
driusan, line 359:
```suggestion
```

PR #7955, opened 2022-01-17
[PHAN] update to 5.0

kongtiaowang:
## Brief summary of changes

update Phan to 5.0 
fix phan errors 

pass git action

preparing to update the gitactions to PHP8.1


modules/datadict/php/datadict.class.inc
```diff
@@ -60,11 +60,20 @@ class Datadict extends \DataFrameworkMenu
     {
         $instruments     = \Utility::getAllInstruments();
         $dictInstruments = [];
+        $lorisinstance   = new \LORIS\LorisInstance(
```
driusan, line 63: There is no reason to instantiate a new object here, it should just use $this->lorisinstance

driusan (changes requested):

modules/datadict/php/datadict.class.inc
kongtiaowang: fixed

PR #7953, opened 2022-01-14
[behavioural_qc] Add site and project filters

ridz1208:
## Brief summary of changes

- [ ] Have you updated related documentation?

#### Testing instructions (if applicable)

1.

#### Link(s) to related issue(s)

* Resolves #  (Reference the issue this fixes, if any.)

PR #7956, opened 2022-01-18
[behavioural_qc] Add Site and Project filters access restrictions

CamilleBeau:
## Brief summary of changes

1. Test that non-super user only sees their own sites and projects in the filter drop-downs in each of the three tabs
2. Make sure that only rows that pertain to the user's sites / projects are populated in the table results

* Continuation of #7953


ridz1208:
@driusan I think this should be rebased to 24. no?

PR #7957, opened 2022-01-24
[DQT] Fix warnings in instrument import script

driusan:
This fixes warnings related to deprecated functions in the
CouchDB_Import_Instruments script.

### DIFF
--- a/tools/CouchDB_Import_Instruments.php
+++ b/tools/CouchDB_Import_Instruments.php
@@ -32,15 +32,19 @@ class CouchDBInstrumentImporter
                 // of using Database::singleton in case it's a mock.
     var $CouchDB; // reference to the CouchDB database handler
 
+    var $loris;
+
     /**
      * Create new instance.
      */
-    function __construct()
+    function __construct(\LORIS\LorisInstance $loris)
     {
-        $factory       = \NDB_Factory::singleton();
-        $config        = \NDB_Config::singleton();
+        $this->loris = $loris;
+        $config        = $this->loris->getConfiguration();
         $couchConfig   = $config->getSetting('CouchDB');
-        $this->SQLDB   = $factory->Database();
+        $this->SQLDB   = $this->loris->getDatabaseConnection();
+
+        $factory = \NDB_Factory::singleton();
         $this->CouchDB = $factory->couchDB(
             $couchConfig['dbName'],
             $couchConfig['hostname'],
@@ -207,7 +211,7 @@ class CouchDBInstrumentImporter
             $JSONData      = $instrumentObj->usesJSONData();
             $tableName     = "";
             if ($JSONData === false) {
-                $tableName = $instrumentObj->table;
+                $tableName = $instrumentObj->table ?? "";
             }
 
             $this->CouchDB->beginBulkTransaction();
@@ -257,6 +261,10 @@ class CouchDBInstrumentImporter
                 $success = $this->CouchDB->replaceDoc($CommentID, $doc);
                 print "$row[PSCID] $row[Visit_label] $instrument: $success\n";
 
+                if(!isset($results[$success])) {
+                    $results[$success] = 0;
+                }
+
                 $results[$success] += 1;
 
                 // Count every 200 //
@@ -278,7 +286,7 @@ class CouchDBInstrumentImporter
      */
     function getInstruments()
     {
-        return \Utility::getAllInstruments();
+        return \NDB_BVL_Instrument::getInstrumentNamesList($this->loris);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


tools/CouchDB_Import_Instruments.php
```diff
@@ -207,7 +211,7 @@ function updateCandidateDocs($Instruments)
             $JSONData      = $instrumentObj->usesJSONData();
             $tableName     = "";
             if ($JSONData === false) {
-                $tableName = $instrumentObj->table;
+                $tableName = $instrumentObj->table ?? "";
```
driusan, line 214: Fixes an error when a LINST file doesn't have a `.meta` file present. Without this "tableName" is null instead of the empty string and causes an error when used

tools/CouchDB_Import_Instruments.php
```diff
@@ -257,6 +261,10 @@ function updateCandidateDocs($Instruments)
                 $success = $this->CouchDB->replaceDoc($CommentID, $doc);
                 print "$row[PSCID] $row[Visit_label] $instrument: $success\n";
 
+                if(!isset($results[$success])) {
```
driusan, line 264: Fixes a warning the first time it's set.

tools/CouchDB_Import_Instruments.php
```diff
@@ -278,7 +286,7 @@ function updateCandidateDocs($Instruments)
      */
     function getInstruments()
     {
-        return \Utility::getAllInstruments();
+        return \NDB_BVL_Instrument::getInstrumentNamesList($this->loris);
```
driusan, line 289: getAllInstruments prints a big "deprecated" warning and tells you to use \NDB_BVL_Instrument::getInstrumentNamesList, so that's what this does.

tools/CouchDB_Import_Instruments.php
```diff
@@ -312,15 +320,15 @@ function createRunLog($results)
...
+        $tests = $this->getInstruments($this->loris);
         $this->updateDataDicts($tests);
         $results = $this->updateCandidateDocs($tests);
         $this->CreateRunLog($results);
     }
 }
     // Don't run if we're doing the unit tests, the unit test will call run..
 if (!class_exists('UnitTestCase')) {
-    $Runner = new CouchDBInstrumentImporter();
+    $Runner = new CouchDBInstrumentImporter($lorisInstance);
```
driusan, line 331: Uses the LorisInstance object defined in generic_includes

driusan:
@ridz1208 the bulk of this is fixes for warnings caused by things you deprecated/added (`.meta` file and `getAllInstruments` deprecation), can you have a look?

tools/CouchDB_Import_Instruments.php
```diff
@@ -278,7 +288,7 @@ function updateCandidateDocs($Instruments)
      */
     function getInstruments()
```
ridz1208, line 289: this function should take a lorisinstance as a parameter no?

ridz1208 (changes requested):

tools/CouchDB_Import_Instruments.php
driusan: why? It has access to the class variable..

ridz1208: line 325, you are passing it in arent you ?

tools/CouchDB_Import_Instruments.php
driusan: It looks like there's no `table` tag in the bmi linst file that comes with raisinbread.

https://github.com/aces/Loris/blob/main/raisinbread/instruments/bmi.linst

(This seems like it was always the case since the reason I encountered the bug was because I had copied bmi.linst into my project/instruments directory before the creation of the .meta file..)

ridz1208: no, it was only removed when the meta was added
https://github.com/aces/Loris/commit/544da8d62f9948cf29226d4540baccd4f669b925#diff-3047eea24c5234ad4857c3257d698a9efb3af98136856bc04dfbc5484f0b1c2c

tools/CouchDB_Import_Instruments.php
driusan: Oops.. just fixed line 325.

ridz1208 (approved):

tools/CouchDB_Import_Instruments.php
driusan: ok, that's weird.. I had a bmi.linst with no table tag but no meta file either on my sandbox. Should I just revert this part?

ridz1208: I don't think theres harm in it, but I just wanted to make sure its not a core issue...

PR #7963, opened 2022-01-26
Add S3 Support to LORIS

driusan:
This adds support for accessing files via S3 to LORIS. The AWS SDK is used so that `s3://` URLs can be accessed as normal files within PHP if the environment variables `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are set (which roughly correspond to AWS username and passport). One of the LORIS config settings `AWS_S3_Region` or `AWS_S3_Endpoint` must be set for the SDK to work (the latter for AWS hosted instances, and the latter for self-hosted.)

In this PR the LORIS API for imaging is updated (for both thumbnails and the mnc files). The imaging_browser and brainbrowser have already been updated to use the LORIS API, so should work transparently.

Other modules that deal with files (document_repository, media, issue_tracker, etc) will need to be tested separately. (The most likely cause of bugs is that a path config option is concatenated to the filename at some point, invalidating the url.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ changes in the following format: PR #1234***
 - Added ability for scripts to bulk load instrument data (PR #6869)
 - Multiple classes of errors flagged by phan are now fixed (various PRs)
 - A PSR3 compatible logging interface was added (PR #7509)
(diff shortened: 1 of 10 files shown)


cmadjar (approved):
Tested and works great!

PR #7976, opened 2022-02-03
PHPSTAN update

kongtiaowang:
## Brief summary of changes

update phpstan from 0.12 to 1.4 
Several files in the "excludePaths" will be fixed after this PR.

passing Travis

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "squizlabs/php_codesniffer" : "^3.5",
         "phpunit/phpunit" : "9.4.4",
         "phan/phan": "^5.0",
-        "phpstan/phpstan": "0.12.58",
+        "phpstan/phpstan": "^1.4",
         "slevomat/coding-standard": "^6.4",
         "php-webdriver/webdriver" : "dev-main"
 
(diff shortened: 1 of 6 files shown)


modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -576,18 +576,16 @@ function editCandidateDOD(\Database $db): void
         throw new \LorisException('Date not valid.');
     }
 
-    if (!empty($dod)) {
```
driusan, line 579: I don't see why this if statement is removed, it's changing the behaviour.

modules/user_accounts/ajax/rejectUser.php
```diff
@@ -53,7 +52,7 @@
         "This account does not exist"
     );
 }
-
+    /* @phpstan-ignore-next-line */
```
driusan, line 55: Why is this ignored?

modules/candidate_parameters/ajax/formHandler.php
```diff
@@ -105,7 +105,7 @@ function editCandInfoFields(\Database $db)
 
     $db->update('candidate', $updateValues, ['CandID' => $candID]);
 
-    foreach (array_keys($_POST ?? []) as $field) {
+    foreach (array_keys($_POST) as $field) {
```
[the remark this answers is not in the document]
kongtiaowang: $_POST default value is [], so the phpstan shows the error. 
Screen Shot 2022-02-26 at 9 26 16 PM

PR #7981, opened 2022-02-07
[Visits Architecture (3 of 4)] - Visit class & UI for new visits

ridz1208:
## Brief summary of changes

This PR Creates revamps the visit class infrastructure and leverages the new subproject class to consolidate the project - subproject - visit associations. This PR also creates a basic configuration module UI to add/remove/edit and associate visits to projects and subprojects

TODO:
- [ ] Copy Subproject class structure
- [ ] remove Utility functions as much as possible

1.


ridz1208:
@driusan @xlecours Ignore everything other than the VisitData and VisitController classes for now.

I went with a slightly different approach just to compare things here to the design I used for Cohorts. The main difference being, I added functions like `withProjects()` and `withCohorts()` that load the projects and cohorts into the VisitData object instead of just returning them separately as opposed to the `getProjectsForCohort()` function I had added in the cohort class. They are not yet immutable but I think its useless to do immutability before 8.1 because 8.1 makes it soo much simpler with the readonly attribute. 

I didnt update all the UI classes yet until I see what you think of that design first. do we go back to just the controller fetching a list of nested entities or is it worth loading those in the Data object and returning the entire thing?

PR #7987, opened 2022-02-10
Upgrade PHPStan

driusan:
This upgrade phpstan to the latest version. It's a simplified version of #7976 which only does the changes required for upgrading it.


driusan:
@kongtiaowang @cmadjar can you review so I can move forward with the php 8.1 PR that's blocked by this?

cmadjar (approved):
looks good but given I don't know much about phpstan, might be worth having an extra look at it from someone else

kongtiaowang (approved):
LGTM

PR #7989, opened 2022-02-10
Add PHP 8.1 Support

driusan:
This fixes the loading of the main LORIS pages with PHP 8.1.

Between 8.0 and 8.1 the behaviour of the PDO changed so that database selects are typed according to the column, while previously they were always a string. This restores the behaviour of 8.0 by always casting them with strval to avoid fatal type errors.

Some signatures for jsonSerialize also had to be modified to prevent warnings and phpstan had to be upgraded. The dashboard and a few random pages now load, but I haven't tested this extensively.

Blocked by #7987

### DIFF
--- a/.github/workflows/loristest.yml
+++ b/.github/workflows/loristest.yml
@@ -13,7 +13,7 @@ jobs:
             - static
             - unit
             - integration
-            php: ['7.4','8.0']
+            php: ['8.0','8.1']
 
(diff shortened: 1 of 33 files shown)


kongtiaowang (approved):
LGTM

kongtiaowang:
@driusan resolve conflicts please.

PR #7991, opened 2022-02-14
24.0 release patch fix NewModulePermissions.sql

kongtiaowang:
## Brief summary of changes
fix some users is missing from the users table. 

run the patch, will not fail.


ridz1208:
With the existen foreign keys, orphaned users entries should not exist. this is most likely a project data issue

Issue #7960, opened 2022-01-25
[create_timepoint] Test plan outdated


driusan:
The test plan for create_timepoint needs updating. It refers to validating/changing settings in `config.xml` that were removed in #7729 and are now in database tables.

driusan:
Fixed by #8000

PR #8000, opened 2022-02-19
[create_timepoint] Use visit name for session table insert

### LINKED ISSUE #7998: [Manual testing] - New timepoint
**Describe the bug**
When selecting a visit during the creation of a new timepoint, the proper visit name appears. however, you get to the list of visits, they appear as numbers (i.e. 1-250) instead of the visit name.

ridz1208:
## Brief summary of changes
Visit ID was accidentally being submitted at timepoint creation instead of visit name. This fixes that issue and corrects the documentation in the Utility class.

1. Create a new timepoint for any candidate and make sure the value inserted into the session table is the visitName, (Not vistID nor visit label, see `visit` table in the DB for distinction)

* Resolves #7998, #7960  (Reference the issue this fixes, if any.)

### DIFF
--- a/modules/create_timepoint/php/timepoint.class.inc
+++ b/modules/create_timepoint/php/timepoint.class.inc
@@ -3,6 +3,7 @@
 namespace LORIS\create_timepoint;
 
 use LORIS\StudyEntities\Candidate\CandID;
+use LORIS\VisitController;
 use \Psr\Http\Message\ServerRequestInterface;
 use \Psr\Http\Message\ResponseInterface;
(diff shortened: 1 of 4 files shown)


ridz1208:
@driusan the new function in the visitcontroller class was not really necessary but I made it to avoid a DB call in the timepoint class of create_timepoint module... that function will most likely be removed when the session table is normalized (visit labels replaced by visit IDs)

The upcoming visit cleanup should deal with all of these on branch 25

driusan (approved):
Created a visit, it's now using the visit label..

PR #8043, opened 2022-03-09
[data_release] Fix syntax error

### LINKED ISSUE #8023: [data_release] Cannot add a permission for "Unversioned" files
In the data release module there is the option to add Unversioned files. When a user with `data_release_edit_file_access` permission tries to add a permission for another user to access all files that are Unversioned, there is an Internal Access Error and the permission is not added.
 **Note** that a user can add/delete the Unversioned permission from the `Manage Permissions` page but not from the `Add Permissions` page.

**To Reproduce**
1. Create one user with all data release permissions (including edit file access) and another user with only the view/upload permissions
2. Go to the data release module as the user with edit_file_access permission
3. If there are no Unversioned files, upload some test files and do not specify a version
4. Click on the Add Permission button and select the other user and the "Unversioned" option from the Data Release Version dropdown (leave the File option blank)
5. See error
6. You can also check that if you click on Manage Permissions and add the Unversioned option from this page, everything works as expected 

Adding the permission should not return an error and when the other user opens the data release module, they should see all the Unversioned files

**Comments**
[This](https://github.com/aces/Loris/blob/main/modules/data_release/php/permissions.class.inc#L151) section of code might be the cause of this error

ridz1208:
## Brief summary of changes
Fixes a syntax error in the database query for add permission functionality

* Resolves #8023

### DIFF
--- a/modules/data_release/php/permissions.class.inc
+++ b/modules/data_release/php/permissions.class.inc
@@ -148,7 +148,8 @@ class Permissions extends \NDB_Page
             if ($version == 'Unversioned') {
                 $IDs = $DB->pselectCol(
                     "SELECT id FROM data_release WHERE
-                        version IS NULL OR version=''"[],
+                        version IS NULL OR version=''",
+                    []
                 );
             } else {
                 $IDs = $DB->pselectCol(


ridz1208:
@AlexandraLivadas can you please test that #8023 is fixed ?

AlexandraLivadas (approved):
LGTM and solves the issue! I think the integration tests just need to be run again to make sure they pass

Issue #8041, opened 2022-03-08
[survey_accounts] Survey email being sent for both "Create survey" and "Email survey"


CamilleBeau:
**Describe the bug**
When I click "Create survey" instead of "Email survey", a survey email is still sent to the participant.

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to survey accounts module and click "Add survey"
3. Fill out information to add a survey, add your own email as the participant email
5. Click "Create survey" instead of "Email survey"
6. See that you receive an email

**What did you expect to happen?**
No email to be sent when "Create survey" is selected.

driusan:
PR merged

PR #8053, opened 2022-03-10
[survey_accounts] Survey email being sent for both "Create survey" and "Email survey" - fix 

kongtiaowang:
## Brief summary of changes
add a "Create survey" check into the function.
https://github.com/aces/Loris/issues/8041

Describe the bug
When I click "Create survey" instead of "Email survey", a survey email is still sent to the participant.

To Reproduce

Go to survey accounts module and click "Add survey"
Fill out information to add a survey, add your own email as the participant email
Click "Create survey" instead of "Email survey"
See that you receive an email
What did you expect to happen?
No email to be sent when "Create survey" is selected.

### DIFF
--- a/modules/survey_accounts/php/addsurvey.class.inc
+++ b/modules/survey_accounts/php/addsurvey.class.inc
@@ -247,7 +247,7 @@ class AddSurvey extends \NDB_Form
         }
         $this->tpl_data['success'] = true;
 
-        if ($email) {
+        if ($email && ($_REQUEST['fire_away'] !== 'Create survey')) {
             $config   = \NDB_Config::singleton();
             $baseURL  = \NDB_Factory::singleton()->settings()->getBaseURL();
             $msg_data = [


modules/survey_accounts/php/addsurvey.class.inc
```diff
@@ -247,7 +247,7 @@ class AddSurvey extends \NDB_Form
         }
         $this->tpl_data['success'] = true;
 
-        if ($email) {
+        if ($email && ($_REQUEST['fire_away'] !== 'Create survey')) {
```
driusan, line 250: This shouldn't be using the superglobal, we should be using the PSR7 request object to get variables.

driusan (dismissed):

driusan:
@CamilleBeau since you reported the bug can verify that this fix works?

CamilleBeau:
Just tested and I am still receiving the email when hitting "Create survey" instead of "Send survey"

kongtiaowang:
It has been fixed, and test on https://wangshen-dev.loris.ca/

CamilleBeau:
Looks good!

CamilleBeau (approved):

Issue #8033, opened 2022-02-28
[media] 504 Gateway Timeout + Unexpected Token in JSON 


suzanne-lee:
**Describe the bug**

The media module does not load. 

When you navigate to the media module from the dashboard, the blue tail spin loader appears for on average 60 seconds before the "An error occurred while loading the page." message appears. The console shows a 504 Gateway Timeout message and a syntax err message that reads: "Syntax Error: Unexpected token < in JSON at position 0" 

<img width="1135" alt="Screen Shot 2022-02-25 at 4 11 38 PM" src="https://user-images.githubusercontent.com/29237731/156059712-ecf9926d-e40b-473e-9bae-400244476c36.png">

When you refresh the page, the following page occurs, with only the 504 error in the console: 

<img width="1077" alt="Screen Shot 2022-02-28 at 4 11 29 PM" src="https://user-images.githubusercontent.com/29237731/156059953-4546c7dc-025f-4689-93ce-dc4e66ad8533.png">

**Browser Environment (please complete the following information):**
- Chrome 98.0.4758.109
- Firefox 97.0.1

ridz1208:
@CamilleBeau is that one of the modules you had issues with ?

CamilleBeau:
> @CamilleBeau is that one of the modules you had issues with ?

Yes this is one of them!

xlecours:
@CamilleBeau @suzanne-lee 
I can't reproduce the behavior on test VM
Can it be a permission issue?

Usually, `SyntaxeError: Unexpected token '<' in JSON at position 0` means that HTML was received instead of JSON. Can you look at the value returned? It migth contain additional information about the error. Also, is there something in apache error_log?

ridz1208:
@xlecours I can help here. 

The reason that error is showing in the console s that the response to `https://ccna-dev.loris.ca/media/?format=json` is
```
<html>
<head><title>504 Gateway Time-out</title></head>
<body>
<center><h1>504 Gateway Time-out</h1></center>
<hr><center>nginx/1.20.1</center>
</body>
</html>
```

That's the HTML that is received instead of the JSON. Nothing in the apache error log or access log.

xlecours:
ok thanks @ridz1208

xlecours:
it's not the database, the query runs in in less than a second `31535 rows in set (0.36 sec)`.
Next thing to look at is what is done for each rows.

xlecours:
Oh well...
[...]

PR #8055, opened 2022-03-15
[media] Caching instrument names

xlecours:
## Brief summary of changes
This adds a static variable in the getInstance function of the provisioner to load each instrument only once instead of for each rows.

1. Access `/media`
2. Check if datatable loads. (There should be no timeout)

Resolves https://github.com/aces/Loris/issues/8033

### DIFF
--- a/modules/media/php/mediafileprovisioner.class.inc
+++ b/modules/media/php/mediafileprovisioner.class.inc
@@ -82,7 +82,7 @@ class MediaFileProvisioner extends \LORIS\Data\Provisioners\DBRowProvisioner
      */
     public function getInstance($row) : \LORIS\Data\DataInstance
     {
-        $testname = $row['testName'] ?? null;
+        $testname = $row['testName'] ?? '';
         unset($row['testName']);
 
         if (!isset($this->_instrumentnames[$testname])) {
--- a/modules/media/php/mediafileprovisioner.class.inc
+++ b/modules/media/php/mediafileprovisioner.class.inc
@@ -32,6 +32,8 @@ class MediaFileProvisioner extends \LORIS\Data\Provisioners\DBRowProvisioner
 {
     protected $lorisinstance;
 
+    private $_instrumentnames = [];
+
     /**
      * Create a MediaFileProvisioner, which gets files for the meida
      * menu table.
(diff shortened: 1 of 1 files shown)


xlecours:
@driusan I could have use an instance variable instead of a static one. Any thoughts?

driusan:
@CamilleBeau can you try this on the CCNA dev environment and see if it fixes the media module?

CamilleBeau:
No longer a timeout error, but getting a 500 now.

image

> [Tue Mar 15 13:12:42.396488 2022] [php7:error] [pid 22865] [client 192.168.122.1:51664] PHP Fatal error:  Uncaught TypeError: Argument 2 passed to NDB_BVL_Instrument::factory() must be of the type string, null given, called in /var/www/ccna/modules/media/php/mediafileprovisioner.class.inc on line 92 and defined in /var/www/ccna/project/libraries/NDB_BVL_Instrument.class.inc:223\nStack trace:\n#0 /var/www/ccna/modules/media/php/mediafileprovisioner.class.inc(92): NDB_BVL_Instrument::factory(Object(LORIS\\LorisInstance), NULL)\n#1 /var/www/ccna/src/Data/Provisioners/DBRowProvisioner.php(118): LORIS\\media\\MediaFileProvisioner->getInstance(Array)\n#2 [internal function]: class@anonymous->current()\n#3 /var/www/ccna/src/Data/Table.php(83): FilterIterator->rewind()\n#4 /var/www/ccna/src/Data/Table.php(98): LORIS\\Data\\Table->toArray(Object(User))\n#5 /var/www/ccna/php/libraries/DataFrameworkMenu.class.inc(148): LORIS\\Data\\Table->toJSON(Object(User))\n#6 /var/www/ccna/php/libraries/NDB_Menu_Filter.class.inc(648): DataFrameworkMenu->toJSON()\n#7 /var/www/ccna/php/libraries/NDB_Page.class.inc(753): NDB_Menu_Filter->display()\n#8 /var in /var/www/ccna/project/libraries/NDB_BVL_Instrument.class.inc on line 223, referer: https://ccna-dev.loris.ca/media/

xlecours:
@CamilleBeau this should be ready now. Sorry for the previous commits. I couldn't test them on my VM and I took a chance.
There is still a bug with the langage column. I'll fix it in an other PR.

xlecours:
Nevermind, there is no bug in the language column. 
I though it needed to join the session.laguageID column but saw that there was a langugae_id column in the media table.

CamilleBeau:
Still getting a timeout in the ccna-dev VM unfortunately. 
image

This time no 500 error, only 504.

xlecours:
Oh well. We gave it a try. Can we try with more RAM ? Otherwise, pagination or storing the instrument names in the database would be options.

CamilleBeau:
> Oh well. We gave it a try. Can we try with more RAM ? Otherwise, pagination or storing the instrument names in the database would be options.

Unfortunately still getting the timeout with 3.7G RAM, which matches production.

driusan:
Can you try it without the try/catch to see if there's an exception being thrown by one (or more) of the instruments?

`isset` will be false if the value is null and won't actually be cached.

CamilleBeau:
> Can you try it without the try/catch to see if there's an exception being thrown by one (or more) of the instruments?
> 
> `isset` will be false if the value is null and won't actually be cached.

Without the try / catch there is no timeout and we get this error instead 
image

xlecours:
Projects that have entries in `media.intrument` column that do not match an existing instrument on their filesystem won't be able to load the module. They will get a `404 Not Found` page.

CamilleBeau:
> @CamilleBeau can you review?

Looks good! Might be worth adding to the CHANGELOG in "Notes for existing projects" about the possible error in the case that instruments don't exist in the filesystem like we have on CCNA

PR #8070, opened 2022-03-22
[tools] Fix delete_candidate & delete_timepoint

### LINKED ISSUE #8059: [tools] delete_candidate failing
**Describe the bug**
The delete_candidate.php script is failing because it is using test_name as tableName. The script must instantiate an instrument and get tableName instead.

**To Reproduce**
1. Run delete_candidate for a candidate that has an instrument with a different table name / test name
2. See error

CamilleBeau:
## Brief summary of changes
This PR updated delete_timepoint and delete_candidate scripts to run smoothly. The following is changed:
- Adds "candidate_adverse_event" table to be deleted from in delete_candidate
- Fixes bug where DB was not called properly in delete_candidate
- Uses table name instead of test_name to delete from in delete_timepoint & checks for whether it is a json instrument.

1. Test scripts with a candidate / timepoint that has instruments with a different table name from test_name
2. Test delete_candidate with candidate that has entry in candidate_adverse_event

* Resolves #8059

### DIFF
--- a/tools/delete_timepoint.php
+++ b/tools/delete_timepoint.php
@@ -173,19 +173,27 @@ function deleteTimepoint($CandID, $sessionID, $confirm, $printToSQL, $DB, $outpu
         ['sid' => $sessionID]
     );
 
+    $loris = new \LORIS\LorisInstance(
+        \NDB_Factory::singleton()->database(),
+        \NDB_Factory::singleton()->config(),
+        [
+            "project/modules",
+            "modules",
+        ]
+    );
+
     // Print each instrument instance
     foreach ($instruments as $instrument) {
         try {
-            $result = $DB->pselect(
-                'SELECT CommentID FROM '
-                . $DB->escape($instrument['Test_name'])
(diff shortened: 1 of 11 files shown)


CamilleBeau:
@

> The table "candidate_adverse_event" doesn't exist in LORIS. I think that's a CCNA customization that shouldn't be in this PR.
> 
> There's also conflicts with changelog..

Fixed

tools/delete_timepoint.php
```diff
@@ -173,19 +173,27 @@ function deleteTimepoint($CandID, $sessionID, $confirm, $printToSQL, $DB, $outpu
         ['sid' => $sessionID]
     );
 
+    $loris = new \LORIS\LorisInstance(
```
driusan, line 176: This script already includes `generic_includes.php`, there's already a LorisInstance object instantiated there that should just be used

CamilleBeau: Done

PR #8078, opened 2022-04-04
[DQT] add delete feature

kongtiaowang:
## Brief summary of changes

Add delete feature for DQT
Add tools/dqt_delete_saved_query.php

1. Save a private or global query.
2. click delete button to delete or using dqt_delete_saved_query.php to delete the new saved query.

### DIFF
--- a/modules/dqt/ajax/DeleteDoc.php
+++ b/modules/dqt/ajax/DeleteDoc.php
@@ -1,0 +1,33 @@
+<?php
+/**
+ * Data Querying Module
+ *
+ * PHP Version 5
+ *
+ * @category Data_Querying_Module
+ * @package  Loris
+ * @author   Loris Team <loris-dev@bic.mni.mcgill.ca>
+ * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ * @link     https://www.github.com/aces/Loris/
+ */
+require_once __DIR__ . '/../../../vendor/autoload.php';
+$client = new NDB_Client();
+$client->makeCommandLine();
+$client->initialize(__DIR__ . "/../../../project/config.xml");
+header("Content-Type: application/json");
+$config      = \NDB_Config::singleton();
+$couchConfig = $config->getSetting('CouchDB');
+$cdb         = \NDB_Factory::singleton()->couchDB(
+    $couchConfig['dbName'],
+    $couchConfig['hostname'],
+    intval($couchConfig['port']),
+    $couchConfig['admin'],
+    $couchConfig['adminpass']
+);
+$docID       = urlencode($_REQUEST['DocID']);
+
+$results = $cdb->deleteDoc(
+    $docID
+);
+print json_encode($results);
+
--- a/modules/dqt/ajax/DeleteDoc.php
+++ b/modules/dqt/ajax/DeleteDoc.php
@@ -39,6 +39,6 @@ if ($doc_author == $user->getUsername()) {
 if ($user->hasPermission('superuser') || $is_author) {
     $results = $cdb->deleteDoc($docID);
     print json_encode($results);
-) else {
-   print json_encode("false");
+} else {
+    print json_encode("false");
 }
--- a/modules/dqt/ajax/DeleteDoc.php
+++ b/modules/dqt/ajax/DeleteDoc.php
@@ -10,10 +10,12 @@
  * @license  http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
  * @link     https://www.github.com/aces/Loris/
  */
-require_once __DIR__ . '/../../../vendor/autoload.php';
+// ########################## IBIS OVERRIDE START #######################
+require_once __DIR__ . '/../../../../vendor/autoload.php';
 $client = new NDB_Client();
 $client->makeCommandLine();
-$client->initialize(__DIR__ . "/../../../project/config.xml");
+$client->initialize(__DIR__ . "/../../../../project/config.xml");
+// ########################## IBIS OVERRIDE END #######################
 header("Content-Type: application/json");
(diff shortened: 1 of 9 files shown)


modules/dqt/ajax/DeleteDoc.php
```diff
@@ -0,0 +1,33 @@
+<?php
```
maltheism, line 1: Hi @kongtiaowang I think @driusan and @xlecours don't want additional ajax files anymore. Instead the request for delete can be done with the RequestHandlerInterface similar to the dqt_setup.class.inc used by the js handleProgressBarSetup function.

modules/dqt/ajax/saveQuery.php
```diff
@@ -69,7 +69,7 @@
                                     ': ' .
                                     $_REQUEST['QueryName'];
 }
-$fields = json_decode($_REQUEST['Fields']);
+$fields = json_decode(strval($_REQUEST['Fields']));
```
maltheism, line 72: Hi @kongtiaowang what's the reason for this change? Also @ridz1208 do you remember a discussion about whether it's good to take the string value here or not because I somewhat remember us discussing whether bad or good.

modules/dqt/jsx/react.savedqueries.js
```diff
@@ -70,7 +71,35 @@ const ManageSavedQueryFilters = (props) => {
 const ManageSavedQueryRow = (props) => {
   const [fieldsVisible, setFields] = useState(null);
   const [filtersVisible, setFilters] = useState(null);
-
+  /**
+   * @deleteclick
+   */
+function publicquerydelete() {
+           let id = props.Query['_id'];
```
maltheism, line 78: This can be a const instead of a let.

maltheism (changes requested):
Hi @kongtiaowang the feature works great!

modules/dqt/php/dqt_setup.class.inc
```diff
@@ -166,10 +166,11 @@ class Dqt_Setup extends \NDB_Form implements RequestHandlerInterface
 
         $usersavedNames   = array_map($IDMapCallback, $usersaved);
         $globalsavedNames = array_map($IDMapCallback, $globalsaved);
-
+        $user = \User::singleton();
```
xlecours, line 169: $user is already defined at line 88
```
$user = $request->getAttribute('user');
```

modules/dqt/ajax/saveQuery.php
kongtiaowang: "$_REQUEST['Fields']" returns a object ({}),  we need a ("{}") to put inside json decode. that is the reason. 
Json_decode only takes string value inside the function.  
json_decode(
    **string** $json,
    ?bool $associative = null,
    int $depth = 512,
    int $flags = 0
): [mixed](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.mixed)

modules/dqt/ajax/DeleteDoc.php
xlecours, line 1: This would allow any logged in user to delete any document in CouchDB regardless of permissions or document "type".

xlecours (changes requested):

modules/dqt/ajax/DeleteDoc.php
kongtiaowang: Good catch, I have added the permission check.

maltheism (changes requested):
Hi @kongtiaowang I reviewed again and the only thing I see left is the eslint-disable-line(s) that should probably be removed? Also this new delete feature should probably be added to the changelog file of Loris.

maltheism (approved):

kongtiaowang:
@maltheism Thanks

modules/dqt/ajax/DeleteDoc.php
```diff
@@ -0,0 +1,43 @@
...
+$doc_author  = str_replace("global:", '', $tmp_author[0]);
+if ($doc_author == $user->getUsername()) {
+    $is_author = true;
+}
+
+if ($user->hasPermission('superuser') || $is_author) {
+    $results = $cdb->deleteDoc($docID);
+    print json_encode($results);
+} else {
+    print json_encode("false");
```
xlecours, line 42: This will return a `HTTP/1.1 200 OK` with the string "false"
It should be a 4XX

modules/dqt/ajax/DeleteDoc.php
```diff
@@ -33,5 +33,5 @@
     $results = $cdb->deleteDoc($docID);
     print json_encode($results);
 } else {
-    print json_encode("false");
+    header("HTTP/1.1 404 Not Found");
```
xlecours, line 36: 403 Forbidden

modules/dqt/jsx/react.savedqueries.js
```diff
@@ -92,10 +92,15 @@ function publicquerydelete() {
...
               cache: 'no-cache',
               credentials: 'same-origin',
-              }).then((resp) => resp.json())
+              }).then((resp) => {
+                    if (resp.status == 200) {
+                           swal.fire('delete Successful!', '', 'success');
+                      } else {
+                           swal.fire('delete Not Successful!', '', 'error');
+                      }
+                })
```
xlecours, line 101: check indentation

xlecours (changes requested):

kongtiaowang:
Hi @maltheism Could you test it again? Thanks

maltheism:
HI @kongtiaowang I can still delete successfully.

kongtiaowang:
Thanks @xlecours @maltheism

racostas (approved):
LGTM.

PR #8074, opened 2022-03-28
Move DB credentials to environment

driusan:
This prevents the possibility of PHP printing a stack trace
with database credentials if an exception is thrown from the
database constructor, by moving the credentials from function
arguments to environment variables.

Since the function signature changed, all calls to \Database::singleton
were (finally) updated to use LorisInstance->getDatabaseConnection (if
available) or NDB_Factory->database (if no LorisInstance available in
the current context) in order to centralize the change to one place.

Resolves #8014

PR #8083, opened 2022-04-13
Remove more credentials from stack trace.

driusan:
#8074 removed username/password related arguments from the database singleton call, but not the connect call.

This removes them from select too, so there's nothing vulnerable in the stack trace.


driusan:
@kongtiaowang can you review/test this?

PR #8085, opened 2022-04-14
[Biobank] Add Biobank module from separate repo

driusan:
I decided to see what would be required to bring the biobank module written by @HenriRabalais up to speed with the latest version of LORIS. 

This takes the code from https://github.com/aces/biobank/ and incorporates the changes from https://github.com/HenriRabalais/Loris/tree/Biobank required to get it to build on the latest main branch. Theoretically, merging it into LORIS would ensure that it gets tested with upgrades and doesn't fall behind

I just copied the code from the repo(s) and made enough changes that `make dev` succeeds. My JSDocs (required by eslint) aren't very good since I don't know the module. 

TODO:
- [x] `make dev` should work and the module should load 
- [x] PHPCS / `make checkstatic` should pass
- [ ] README needs to be rewritten in the same format as other module specs
- [ ] JSDoc comments should be reviewed/rewritten by a developer who knows the module
- [ ] At least basic tests should be written
- [ ] There needs to be a test plan
- [ ] SQL needs to be incorporated into the schema

### DIFF
--- a/jsx/LoadingBar.js
+++ b/jsx/LoadingBar.js
@@ -1,5 +1,12 @@
 import React from 'react';
 
+/**
+ * React component to display a loading bar with progress
+ *
+ * @param {object} props - React props
+ *
+ * @return {JSX}
+ */
 function LoadingBar(props) {
   const {progress} = props;
 
--- a/modules/biobank/jsx/Shipment.js
+++ b/modules/biobank/jsx/Shipment.js
@@ -5,6 +5,8 @@ import {get, post} from './helpers.js';
  * React effect for creating a request to create a new
  * shipment
  *
+ * @param {object} initShipment - the initial value for the shipment
+ *
  * @return {Shipment}
  */
 export function UseShipment(initShipment = {}) {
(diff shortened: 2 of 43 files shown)


maltheism:
Hi @driusan maybe close/reopen your PR for the last checks to succeed? I can review the PR when passing.

driusan:
@maltheism I rebased it.. you can review it if you want (it's @HenriRabalais 's work, not mine, but I'll try and address anything you find if I have time.).. but what's mostly missing is the things that aren't checked off in the description that I don't know enough about the module to do.

maltheism:
Hi @driusan, I rebased onto main because I was getting an error when running the tools/raisonbread_refresh.php script. The error doesn't occur on main and rebasing got rid of the error. Everything seems to be okay when visiting the module.

laemtl:
@ridz1208 To talk to @hrabalais about this.

driusan:
I think @HenriRabalais is doing this separately and will eventually send a different PR for it.

PR #8089, opened 2022-05-03
Button to generate JWT token in the frontend

xlecours:
DRAFT


driusan:
This seems like a new feature. It should probably go to either 24.1 or main?

xlecours:
yes

PR #8091, opened 2022-05-12
[LINST/Instruments] Backporting fixes from 24 + group errors

ridz1208:
## Brief summary of changes
This PR cherry picks a few key bugfixes already merged into 24 back into the 23 branch for projects not looking to upgrade immediately and are using LINST

1. Print out (debugging) showing up on the front end of instruments: https://github.com/aces/Loris/commit/a6e1744bc642a0413d2e27bd10f832ee32f4da3e#diff-1d18bbc27ba2cabe74bdf1a067307ce735f0f84fd29c735010eb122abc370c14L843
2. Date elements are always required even when a rule is explicitly added to make them not required: https://github.com/aces/Loris/pull/7925/files#diff-1d18bbc27ba2cabe74bdf1a067307ce735f0f84fd29c735010eb122abc370c14R660
3. The `multiple` parameter should be passed as an array, not a string: https://github.com/aces/Loris/commit/137f1fb5fa19d8934a2d80c340477e1e1b9bee90#diff-1d18bbc27ba2cabe74bdf1a067307ce735f0f84fd29c735010eb122abc370c14R665

4. NEW: This PR also fixes the problem where table rows (rows using `begingroup` and `endgroup` elements) are not highlighted in red when an element within the group breaks XIN rules. the reason for that is the XIN validation in LINST assumes all elements are purely linear, no group logic.

### DIFF
--- a/php/libraries/NDB_BVL_Instrument_LINST.class.inc
+++ b/php/libraries/NDB_BVL_Instrument_LINST.class.inc
@@ -36,6 +36,9 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
 
     public $LinstLines = array();
 
+    // array of the format "field_name" => "group_name"
+    protected $GroupElements = [];
+
     /**
      * Sets up the variables required for a LINST instrument to load
      *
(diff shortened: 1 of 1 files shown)


ridz1208:
blocked by https://github.com/aces/Loris/pull/8092

PR #8103, opened 2022-06-06
Push v23.0.11 into 24.0 release

ridz1208:
DO NOT SQUASH


driusan:
Icon on GitHub doesn't look like it's updating, but clicking through on the commit shows all the tests passed..

PR #8116, opened 2022-06-14
[DQT] v23 - Fix dictionary builder and dictionary display issues

ridz1208:
## Brief summary of changes
This PR is 2 parts:

1. backporting fix for the display of the dictionary in the DQT module (#7098)
2. in the data_dictionary_builder script, reference the `tesname` value in parsed instruments (lorisform parser output format) instead of `table` while keeping it backwards compatible. This is needed because JSONdata instruments do not define a `table` variable and since [23.0.3] (https://github.com/aces/Loris/pull/7153/files#diff-962394c093e5275dcfa7cc6eca8927b2f8f592f42b7b95ec5921cb7fc0dc8453R88) the lorisform parser exports the test name in its output for this specific reason.

### DIFF
--- a/tools/exporters/data_dictionary_builder.php
+++ b/tools/exporters/data_dictionary_builder.php
@@ -73,14 +73,14 @@ $DB->run(
 );
 
 $DB->delete("parameter_type_category", array("Type" => "Instrument"));
-   
+
 print "Cleared data from BVL instruments\n";
 
 print "Reading instruments\n";
 //Read the ip_output.txt staging file.
 $fp = fopen(__DIR__."/../ip_output.txt", "r");
 if (!$fp) {
-    print "There was an issue opening the ip_output file. Ensure the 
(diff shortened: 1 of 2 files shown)


zaliqarosli (approved):
LGTM

tools/exporters/data_dictionary_builder.php
```diff
@@ -185,7 +195,7 @@
                     "Type"        => $bits[0],
                     "Description" => $bits[2],
                     "SourceField" => $bits[1],
-                    "SourceFrom"  => $table,
+                    "SourceFrom"  => $testname,
```
driusan, line 198: I don't think this part is accurate. SourceFrom is supposed to be the table in the quat tables that the data is sourced from.. but I also don't think it matters, because the quat tables don't exist/aren't used anymore and as far as I know this isn't used anywhere, so I'll just merge it..

PR #8119, opened 2022-06-14
fixes getDataDictionary() not support the 'file' type

shonibare:
This PR fixes getDataDictionary() that doesn't support the 'file' type (which is getting called indirectly while saveValues() calls getCandidateAge()).


ridz1208:
@driusan can this be sent to 23 please?

PR #2018, opened 2016-07-22
[User Accounts] Email encoding issue (Redmine10193)

stellarxo:
image

PR #8137, opened 2022-07-12
[user_accounts] Revert #2018 - Sanitization before validation of email falsifies validity

ridz1208:
## Brief summary of changes

The issue came up when registering a user using the User Accounts -> add user functionality. If you select the "Make user name match email address" checkbox and your email address contains HTML special characters, you end up with a 500 error.

The email address failing here is: `test@gmail.com>` This address does not pass `FILTER_VALIDATE_EMAIL` check but because of the sanitization step removed in this PR, the email actually being validated is `test@gmail.com` which is valid.

This PR also removes an old check for < > and " because the new check offers a clearer error message and covers a broader range of characters we don't want in emails (because of escaping issues)

### DIFF
--- a/modules/login/php/requestaccount.class.inc
+++ b/modules/login/php/requestaccount.class.inc
@@ -135,6 +135,12 @@ class RequestAccount extends \NDB_Form
             $errors['from'] = "Please provide a valid email!";
         } else if (!filter_var($_REQUEST['from'], FILTER_VALIDATE_EMAIL) ) {
             $errors['from'] = 'Please provide a valid email!';
+        } else if (preg_match('[<|>|"|&]', $_REQUEST['from'])) {
+            // Although some of these characters are legal in emails, due to the
+            // current HTML escaping method, it is better to reject email
+            // addresses containing them
+            $errors['from'] =  'Email address can not contain any the following' .
+                ' characters: <, >, & and "';
         }
 
         if (empty($values['site'])) {
--- a/modules/login/php/requestaccount.class.inc
+++ b/modules/login/php/requestaccount.class.inc
@@ -133,14 +133,14 @@ class RequestAccount extends \NDB_Form
         }
         if (empty($values['from'])) {
             $errors['from'] = "Please provide a valid email!";
-        } else if (!filter_var($_REQUEST['from'], FILTER_VALIDATE_EMAIL) ) {
-            $errors['from'] = 'Please provide a valid email!';
-        } else if (preg_match('[<|>|"|&]', $_REQUEST['from'])) {
+        } else if (preg_match('/(<|>|"|&)/', $_REQUEST['from'])) {
             // Although some of these characters are legal in emails, due to the
             // current HTML escaping method, it is better to reject email
             // addresses containing them
-            $errors['from'] =  'Email address can not contain any the following' .
-                ' characters: <, >, & and "';
+            $errors['from'] =  'Email address can not contain the following' .
+                ' characters: <,>,& and "';
+        } else if (!filter_var($_REQUEST['from'], FILTER_VALIDATE_EMAIL) ) {
+            $errors['from'] = 'Please provide a valid email!';
         }
 
         if (empty($values['site'])) {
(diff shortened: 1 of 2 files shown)


ridz1208:
@jstirling91 assigned you just to take a look see if it makes sense

ridz1208:
I think tis might stil fail if there is a `&` in the email.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -1287,8 +1287,6 @@ class Edit_User extends \NDB_Form
      */
     private function _getEmailError(\Database $DB, string $email): ?string
     {
-        // remove illegal characters
-        $email = filter_var($email, FILTER_SANITIZE_EMAIL);
```
driusan, line 1291: I'm not sure I fully understand the PR description. Is the problem that the email is being filtered, or is the problem that filter_var is being called twice? (Once here, and once in the if statement immediately below).

It's not clear to me from the PHP documentation what filter_var is supposed to remove when there's invalid characters. The documentation says "Returns the filtered data, or false if the filter fails." but playing around with it on the commandline the filter seems to "fail" instead of filter out invalid characters when I add a `>` instead of returning the string with the characters filtered out.

Maybe the PHP documentation is just unclear and it means "return the exact same string or false if it fails the filter", but it seems like it might be safer to keep this line and change the if statement to `if($email === false)`?

ridz1208: It's not called twice. Different parameter. One is filter, the other is validate

driusan: Ok, now I understand. I'll let someone else review it but this makes sense to me.

modules/login/php/requestaccount.class.inc
```diff
@@ -135,6 +135,13 @@ class RequestAccount extends \NDB_Form
             $errors['from'] = "Please provide a valid email!";
         } else if (!filter_var($_REQUEST['from'], FILTER_VALIDATE_EMAIL) ) {
             $errors['from'] = 'Please provide a valid email!';
+        } else if (preg_match('[<|>|"|&]', $_REQUEST['from'])) {
+            error_log("matched");
```
driusan, line 139: Leftover debugging statement? If it's intentional it should be using a PSR logger instead.

modules/user_accounts/php/edit_user.class.inc
```diff
@@ -1287,11 +1287,15 @@ class Edit_User extends \NDB_Form
...
         if (!filter_var($email, FILTER_VALIDATE_EMAIL)) {
             // If email not syntactically valid
-            return "Invalid email address";
+            return "Invalid email addrezz";
```
driusan, line 1292: More debugging? This should be reverted
```suggestion
            return "Invalid email address";
```

ridz1208: jeeezz

modules/login/php/requestaccount.class.inc
```diff
@@ -135,6 +135,12 @@ class RequestAccount extends \NDB_Form
             $errors['from'] = "Please provide a valid email!";
         } else if (!filter_var($_REQUEST['from'], FILTER_VALIDATE_EMAIL) ) {
             $errors['from'] = 'Please provide a valid email!';
+        } else if (preg_match('[<|>|"|&]', $_REQUEST['from'])) {
+            // Although some of these characters are legal in emails, due to the
+            // current HTML escaping method, it is better to reject email
+            // addresses containing them
+            $errors['from'] =  'Email address can not contain any the following' .
```
driusan, line 142: This is rarely getting triggered because of other validation. `<` is giving the error "You can't use tags in from" from line 165. `"` is giving  the error "You can't use quotes in from" from the same part of the code. `>` is failing with "Please provide a valid email" from line 137.

But there's a bigger problem with quotes in that everything after the `"` gets truncated. I think there's still a reflected XSS attack despite this validation because, while the error is getting triggered, smarty is not escaping it when re-putting the user input into the form with the values that were entered by the user.

driusan: This does catch "&" though.

driusan (approved):
I tested request accounts and this now work. I'm unable to test user_accounts until this is pushed to 24 because of schema differences between 23 and 24 making it impossible to login in my sandbox, but the code is the same so I don't see any reason it wouldn't work.

The code looks good to me too. We lose the checking for < and " in firstname/lastname but the data is escaped by the Database class on insert, so it shouldn't open any vulnerabilities and the comment deleted says it was just there for a double check.

PR #8139, opened 2022-07-14
[user_accounts] Prevent crash when one user has no site/project affiliations

ridz1208:
## Brief summary of changes
Although this usecase should never happen as it is prevented by rules in the edit_user logic, a couple bugs have let it happen in the pass and the entire module becomes inaccessible. This makes the module load and displays a message in the user row informing the editor that the user does not currently have any sites and projects.

The only way to test this would be to manually delete all site associations directly in the DB `user_psc_rel` OR all project associations `user_project_rel`

Note that the module crashes even if the user not having any sites or projects is not the editor nor the editee... any user can cause this issue

### DIFF
--- a/modules/user_accounts/jsx/userAccountsIndex.js
+++ b/modules/user_accounts/jsx/userAccountsIndex.js
@@ -77,15 +77,25 @@ class UserAccountsIndex extends Component {
             .join(', ')}
           </td>
         );
+        if (cell.length === 0) {
+          result = (
+            <td>This user has no site affiliations</td>
+          );
+        }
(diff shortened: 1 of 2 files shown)


ridz1208:
@driusan i was not expecting it but it seems to be what happens. I'm not as concerned about it. its better than a crash

driusan:
I'm not sure if it's what we want.. I think the users will become orphans with no one able to find them but as you say, it's better than a crash, so I'm approving..

driusan (approved):

PR #8148, opened 2022-07-28
[tools/CouchDB_Confirm_Integrity] Add case where VL doesn't exist in SQLDB

zaliqarosli:
## Brief summary of changes

1. Delete a session in SQL DB that has data associated with it i.e. instrument data
2. On main branch, run `php tools/CouchDB_Confirm_Integrity.php`. Go to the DQT, run a query with instrument data fields and notice that your deleted session data is still there.
3. On this PR branch, run `php tools/CouchDB_Confirm_Integrity.php` again. In the script output, you should see a print out that flags the deleted session. Go to the DQT, using the same query as you used above, notice that your deleted session data is no longer there.

### DIFF
--- a/tools/CouchDB_Confirm_Integrity.php
+++ b/tools/CouchDB_Confirm_Integrity.php
@@ -94,7 +94,12 @@ class CouchDBIntegrityChecker
                 ]
             );
 
-            if (!empty($sqlDB) && $sqlDB['cActive'] == 'N') {
+            if (empty($sqlDB)) {
+                print "PSCID $pscid VL $vl does not exist but $row[id] still exists.
+                Deleting Doc. \n";
+
+                $this->CouchDB->deleteDoc($row['id']);
+            } else if (!empty($sqlDB) && $sqlDB['cActive'] == 'N') {
                 print "PSCID $pscid is inactive but $row[id] still exists. 
                 Deleting Doc.\n";
 


tools/CouchDB_Confirm_Integrity.php
```diff
@@ -94,7 +94,12 @@ function run()
                 ]
             );
 
-            if (!empty($sqlDB) && $sqlDB['cActive'] == 'N') {
+            if (empty($sqlDB)) {
+                print "PSCID $pscid VL $vl does not exist but $row[id] still exists.
+                Deleting Doc. \n";
+
+                $this->CouchDB->deleteDoc($row['id']);
```
ridz1208, line 101: If you change the query on line 87 to 
```
SELECT c.*, s.Visit_label, s.Active, c.Active as cActive
                FROM candidate c
                LEFT JOIN session s USING (CandID)
                WHERE c.Active='Y' AND c.PSCID=:PID AND s.Visit_label=:VL
```

Then you can keep these lines and remove lines 102 to 106. you are basically doing a PHP check here that can be done in SQL. something similar might be feasable for session... 
just to remove useless code. no change in behaviour

driusan: I don't think that would work because it would break the check on line 102.

ridz1208: ... read my comment, I'm saying remove line 102 to 106

driusan: That seems like it would make the diagnostics/logging less explicit. Is there any benefit other than a couple fewer lines of code?

ridz1208: none, just code cleanliness

CamilleBeau:
Manual tests went well!

PR #8149, opened 2022-08-10
Add phase encoding direction and scan options to mri protocol and mri protocol violated scans

cmadjar:
## Brief summary of changes

This adds the columns `PhaseEncodingDirections` and `ScanOptions` to the `mri_protocol` and `mri_protocol_violated_scans` tables so that those headers could be used to discriminate between some MRI sequences.

1. Test the SQL patch
2. Test the default schema on an empty DB

* Did not create an issue. This was discussed during a roadmap call.


cmadjar:
@driusan eventually, we will need to modify the mri_violations module to display those new fields. I see that image_type is not in the mri_violations either so I was thinking of sending a different PR to fix this so that at least the MRI side of things work for the insertion.

Thoughts? If you agree, then I will create an issue and assign to me. Thanks!

cmadjar:
New PR based out of 24.1-release branch: https://github.com/aces/Loris/pull/8150

PR #8158, opened 2022-08-25
[imaging_uploader] HasApiSpecification-->true

xlecours:
This adds a schema file to describe the imaging_uploader module endpoints. 
The file can be used by the api_docs module.


cmadjar (approved):
LGTM 👍 Thanks!!

modules/imaging_uploader/static/schema.yml
```diff
@@ -0,0 +1,68 @@
+openapi: 3.0.1
```
driusan, line 1: @cmadjar I think you might be the best person to review if this schema is accurate?

PR #8159, opened 2022-08-25
[imaging browser] add download buttons for BIDS JSON file (and NIfTI bval, bvec files for DWI)

cmadjar:
## Brief summary of changes

Currently, the image download button says 'Download MINC' which is wrong when the file attached is a NIfTI file or another image file format. Instead, the "Download MINC" has been rephrased to a more generic option "Download Image".

For DWI NIfTI files, users should also be able to download the BVAL and BVEC files that come with NIfTI DWI files. This PR adds a download button for each of those files when they are linked to the image file in the `parameter_file` table.

Similarly, for BIDS datasets, there are JSON files that comes with the image file that should be made available for download so this PR adds a download button to the imaging browser if a JSON file is linked to the NIfTI image file in the `parameter_file` table.

#### Testing instructions

Load a BIDS dataset with DWI data and other modalities into LORIS and check that the relevant buttons show in the frontend.

- if don't want to run an entire pipeline, a quick hack would be to add a NIfTI file to the `files` table and add the following entries in the `parameter_file` table
  - one entry linked to the `parameter_type` `Name`  `check_bval_filename `
  - one entry linked to the `parameter_type` `Name`  `check_bvec_filename`
  - one entry linked to the `parameter_type` `Name`  `bids_json_file`

The button should then be seen on the imaging browser. Make sure the paths are relative to the `imagePath` Config.

* Resolves https://github.com/aces/HBCD/issues/226


cmadjar:
@driusan rebase on 24.1-release done!

nicolasbrossard (approved):

nicolasbrossard:
I was able to successfully download BVEC, BVAL and JSON files with the new buttons. Approving.

htdocs/mri/jiv/get_file.php
```diff
@@ -148,6 +148,17 @@
...
     break;
+case 'bvec':
+case 'bval':
+    $FullPath         = $imagePath . '/' . $File;
+    $MimeType         = 'text/plain';
+    $DownloadFilename = basename($File);
+    break;
+case 'json':
+    $FullPath         = $imagePath . '/' . $File;
+    $MimeType         = 'apllication/json';
```
driusan, line 159: I think this is a typo?

driusan (changes requested):

htdocs/mri/jiv/get_file.php
cmadjar: should be fixed in the last commit

PR #8160, opened 2022-08-26
[Instruments] Fix display and loading of time elements

ridz1208:
## Brief summary of changes

This PR fixes the display of time fields in instrument tables to be inline (hours at the left of minutes) instead of vertical (hours on top of minutes) AND fixes the loading of stored time values in the database when the time entry contains leading zeros i.e. `05:05:00`

More info on time data loading: When we were using SQL data saving, SQL seems to have been fixing our time fields by adding a leading 0 to the hours or minutes that were <10 so even when the code was sending `5:5:00`, SQL was converting it to `05:05:00`. Since we started using JSONdata instruments, this correction has not been happening and any time value saved without correction CAN NOT be reloaded when the instrument is reloaded thus magically erasing data.

### DIFF
--- a/smarty/templates/instrument_html.tpl
+++ b/smarty/templates/instrument_html.tpl
@@ -3,8 +3,8 @@
 	.table-instrument>tbody>tr>th{
 		color: black;
 	}
-	.table-instrument>tbody>tr>th, .table-instrument>tbody>tr>td  { 
-	     border-top: none; 
+	.table-instrument>tbody>tr>th, .table-instrument>tbody>tr>td  {
+	     border-top: none;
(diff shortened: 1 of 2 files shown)


paiva (approved):
I have reviewed these changes and tested it on MGSS's prod loris instance. The changes fixed my issue with the dates (hourmin not saving on some specific values like 00h00 or 12h06). Thanks @ridz1208  for fixing this issue

cmadjar:
@ridz1208 milestone should probably be changed for the PR, no? Also, should there be an entry in the CHANGELOG for this PR?

PR #8172, opened 2022-09-22
Move feedback MRI comments popup code out of htdocs into the imaging_browser module code

cmadjar:
## Brief summary of changes

This moves the feedback MRI comments popup code out of the `htdocs` directory to have it in the `imaging_browser` module code instead .

1. Checkout the code from this branch
2. Run `make` or `make dev`
3. Open an imaging session in the imaging browser module and click on the `QC Comments` button below a given image
4. Test that the information displayed is correct and that it saves new data when user has qc permission
5. In the imaging session page of the imaging browser, click on the `Visit Level Feedback` button on the left Navigation panel and make sure the information displayed is correct and that it saves new data when user has qc permission

Related to https://github.com/aces/HBCD/issues/225 (Step 1 of the issue)


modules/imaging_browser/php/feedback_mri_popup.class.inc
```diff
@@ -0,0 +1,309 @@
...
+
+    /**
+     *  Gets data for the feedback popup template
+     *
+     * @return array ($subjectData) with data for the template
+     */
+    function getData() : array
+    {
+        $DB       = $this->loris->getDatabaseConnection();
+        $user     = \User::singleton($_SESSION['State']->getUsername());
```
driusan, line 132: I don't understand the `$_SESSION['State']` part of this.. why does that need to be there?

driusan: Also I think you can get the user out of the $request object in your handle function and use it from there.

driusan: Actually it would be even better to store it in a class variable in `loadResources`, since it's a resource you need for handling the request. It's even passed as an argument there so you don't need to get it from the request object..

cmadjar: ok. Done in the next commit.

modules/imaging_browser/php/feedback_mri_popup.class.inc
```diff
@@ -0,0 +1,309 @@
...
+        // instantiate feedback mri object
+        $sid         = strval($this->session_id);
+        $feedbackObj = new \FeedbackMRI(
+            $this->file_id ?? null,
+            isset($this->session_id) ? new \SessionID($sid) : null,
+        );        // clear all predefined comments
+        $feedbackObj->clearAllComments();
+
+        // set selected predefined comments
+        if (isset($_POST['savecomments']['predefined'])) {
```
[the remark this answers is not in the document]
cmadjar: ok. Done!

zaliqarosli:
mostly good! here are a few screenshots and comments:

1. The breadcrumbs in the pop-up are a bit faulty. The "Imaging Browser" link in the breadcrumbs work, but the "Feedback Mri Popup" link gives you a 404

Screen Shot 2022-09-26 at 5 44 10 PM

Because the page is a pop-up, I would suggest removing the breadcrumbs entirely as I don't foresee users needing to navigate within that window.

2. Removing the QC permissions renders the data in a way that doesn't exactly match. I'll provide screenshots so you can make the judgement on what it's supposed to look like.

- multiple lines are condensed to one
Screen Shot 2022-09-26 at 5 54 04 PM
Screen Shot 2022-09-26 at 5 52 14 PM

- selected comments like 'Fair', 'Suspicious' should maybe not look like it's part of the header
Screen Shot 2022-09-26 at 5 54 17 PM
Screen Shot 2022-09-26 at 5 53 03 PM

zaliqarosli (changes requested):
previous comment

zaliqarosli (approved):

PR #8176, opened 2022-09-30
[core] fix invalid date error

### LINKED ISSUE #8127: Incorrect error message with invalid date format
**Describe the bug**
Since 614b8d82117b434b3ae49827cb1302c99155b2a8 the LORIS API error out with an (HTML) 500 page if the date format is not 'YYYY-MM-DD' saying "Date of Birth must be specified" in the error logs. This is incorrect if the date is specified in another format, rather than missing (and should be a JSON error response instead of the generic LORIS 500 error page).

**To Reproduce**
Post to the candidate API with a DoB in another date format

Either the date format is accepted, or an error message with the correct format is given to the API user. Instead, a 500 error is displayed to the user and an incorrect error message is displayed in the error logs.

### LINKED ISSUE #8144: Able to create a candidate born on 0000-00-00
Using Safari, an IBIS user was able to create a candidate born on 0000-00-00. Since Safari does not offer calendar widgets to enter dates, she had to rely on a text field to enter the DoB and she used the wrong date separator ('/' instead of '-'). The date was saved as '0000-00-00' afterwards since its format was invalid. For reference, IBIS is on LORIS 23 and uses MySQL 5.1.73.

regisoc:
## Context

Several behaviours related to date parsing. In some cases, when using `/new_profile` or `/api/v0.0.3/candidates` to add candidates, date parsing allowed date such as 0000-00-00 or with possible multiple separators when only months were provided (e.g. 'yyyy/mm'). 

## Brief summary of changes

This fix:

- valid date format to use is: 'yyyy-mm-dd'.
- returns 400 Bad Requests/JSON errors for invalid dates.

1. Log into to get a `token`: 
```bash
curl -v -k -X 'POST' \
  'https://<hostname>/api/v0.0.3/login' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "username": "<username>",
  "password": "<password>"
}'
```
2. Test for `/api/v0.0.3/candidates` endpoint:
```bash
curl -v -k -X POST 'https://<hostname>/api/v0.0.3/candidates' \
    -H "Authorization: Bearer <token>" \
    -d '{
        "Candidate": {
            "Project": "Rye",
            "Site": "Montreal",
            "DoB": "<a_date>",
            "Sex": "Female"
        }
    }'
```
3. Test for `/new_profile` endpoint:
```bash
curl -v -k -X POST \
  'https://<hostname>/new_profile' \
  -H "Authorization: Bearer <token>" \
  -F site='1' \
  -F project='3' \
  -F dobDate='<a_date>' \
  -F sex='female'
```

* Resolves #8127
* Resolves #8144


php/libraries/Candidate.class.inc
```diff
@@ -240,15 +246,32 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource,
...
+                DOB_INVALID
+            );
+        }
+        if ($dob > new \DateTime('now')) {
+            throw new \LorisException(
+                "Date of Birth cannot be set in the future",
+                DOB_FUTURE
+            );
+        }
+        if ($dob < new \DateTime('1900-01-01')) {
```
xlecours, line 269: This is arbitrary. Why is there a limit ?

php/libraries/Candidate.class.inc
```diff
@@ -1,4 +1,6 @@
-<?php declare(strict_types=1);
+<?php
+
+declare(strict_types=1);
```
xlecours, line 3: Why is this changed?

modules/new_profile/php/new_profile.class.inc
```diff
@@ -206,12 +206,14 @@ class New_Profile extends \NDB_Form
...
-            if (isset($values['dobDate'])
-                && (new \DateTime($values['dobDate'])) > new \DateTime('now')
-            ) {
+            $parsedDoB = \Utility::parseDate($values['dobDate']);
+            if ($parsedDoB > new \DateTime('now')) {
                 $errors['dobDate'] = 'Date of Birth cannot be set in the future.';
             }
-        } catch (\Exception $e) {
+            if ($parsedDoB < new \DateTime('1900-01-01')) {
+                $errors['dobDate'] = 'Date of Birth cannot be set before 1900.';
```
xlecours, line 214: Why?

modules/api/php/endpoints/candidates.class.inc
```diff
@@ -1,4 +1,6 @@
-<?php declare(strict_types=1);
+<?php
+
+declare(strict_types=1);
```
xlecours, line 3: Why is this changed? Is it a new convention?

xlecours (changes requested):
I don't understand the purpose of these changes. Is it to allow invalide dates to bypass the interface specifications? If we say that the backend expect a YYYY-MM-DD date then the clients needs to provide a YYYY-MM-DD date. 
As for the date range (YYYY-MM), in the end, it will need to be put in a SQL date data type column. the parseDate function only adds `-01` to the value; something that the clients should be able to do.  (And I believe our javascript client is doing that by truncating the returned value of the date picker.)

xlecours:
@regisoc This is unrelated to the API. The title should be changed.

xlecours:
@regisoc Should EDC be handled the same way?

ridz1208:
@regisoc @xlecours  this was automatically closed because the branch was issued from aces

make sure to only issue PRs from you fork of LORIS

you can still move the contents of this branch to your fork and re-issue the PR if necessary (you should still have the branch locally)

regisoc:
Yes will do it today

PR #8181, opened 2022-10-06
Add script which converts a REDCap data dictionary to LINST format

driusan:
This adds a new `tools/` script, redcap2linst, which takes a REDCap data dictionary CSV and outputs a LINST file for each instrument in it. It does not (currently) deal with rules or scoring, only builds the LINST files. It also doesn't deal with inserting into the Test_names or creating the `.meta` files, as that's currently done by the instrument_manager when the LINST file is uploaded to LORIS.

I was only able to implement the types in the data dictionary that I was given for HBCD, so it's possible there are some data types that are missing which need to be implemented (in which case the script will currently throw an exception.)

### DIFF
--- a/tools/redcap2linst.php
+++ b/tools/redcap2linst.php
@@ -96,6 +96,11 @@ function toLINST(
     case 'sql':
         // The "SQL" data type seems to just be for presentation? I hope?
         return "";
+    case'slider':
+        return "numeric{@}$redcapfieldname{@}$label{@}0{@}100";
+    case 'file':
+        // File upload - NOT SUPPORTED BY LINST
+        return "";
     case 'notes':
         // REDCap calls textareas notes
         return "textarea{@}$redcapfieldname{@}$label";
--- a/tools/redcap2linst.php
+++ b/tools/redcap2linst.php
@@ -1,0 +1,124 @@
+#!/usr/bin/env php
+<?php
+
+require_once __DIR__ . "/../vendor/autoload.php";
+require_once "generic_includes.php";
+
+$options = getopt("f:", 
+    [
+        "output-dir:",
+        "file:",
+    ]
+);
+
+$inputFile = $options['f'] ?? $options['file'] ?? '';
+$outputDir = $options['output-dir'] ?? '';
+
+if (empty($inputFile) || empty($outputDir)) {
+    showHelp();
+    exit(1);
+}
+if (!is_dir($outputDir)) {
+    fprintf(STDERR, "Output directory $outputDir doesn't exist.\n");
+    exit(1);
+}
+if (!is_writeable($outputDir)) {
+    fprintf(STDERR, "Output directory $outputDir is not writeable.\n");
+    exit(1);
+}
+
+$fp = fopen($inputFile, "r");
+if ($fp === false) {
+    fprintf(STDERR, "Could not open file $inputFile\n");
+    exit(1);
+}
+
+$headers = fgetcsv($fp);
+$instruments = [];
+while($row = fgetcsv($fp)) {
+    $inst = $row[1];
+    if (!isset($instruments[$inst])) {
+        $instruments[$inst] = [];
+    }
+    $linstFormat = toLINST($row[3], $row[0], $row[4], $row[5]);
+    if (!empty($linstFormat)) {
+        $instruments[$inst][] = $linstFormat;
+    }
+}
+fclose($fp);
+
+outputFiles($outputDir, $instruments);
+
+function toLINST(
+    string $redcaptype,
+    string $redcapfieldname,
+    string $redcaplabel,
+    string $redcapChoices,
+) : string {
+    $label = str_replace($redcaplabel, "\n", "<br /><br />");
+    switch ($redcaptype) {
+        case 'text':
+            // text maps directly to LORIS
+            return "text{@}$redcapfieldname{@}$label";
+        case 'descriptive':
+            // descriptive maps to label with no field.
+            return "static{@}{@}$label";
+        case 'radio':
+        case 'dropdown':
+            // Radio or dropdown maps to a select and the options are in the
+            // same format in the dictionary.
+            // select{@}fieldname{@}fieldlabel{@}NULL=>''{-}'a'=>'a'{-}'b'=>'b'{-}'c'=>'c'{-}'not_answered'=>'Not Answered'
+            return "select{@}$redcapfieldname{@}$label{@}" . optionsToLINST($redcapChoices);
+        case 'checkbox':
+            // checkboxes are the same format as radios but allow multiple options, so map to a
+            // selectmultiple instead of a select
+            return "selectmultiple{@}$redcapfieldname{@}$label{@}" . optionsToLINST($redcapChoices);
+        case 'yesno':
+            // Map yes/no fields to dropdowns with yes and no options.
+            return "select{@}$redcapfieldname{@}$label{@}NULL=>''{-}'yes'=>'Yes'{-}'no'=>'No'";
+        case 'calc':
+            // Calc maps to a score field. We create the DB field but don't do the score.
+            return "static{@}$redcapfieldname{@}$label";
+        case 'sql':
+            // The "SQL" data type seems to just be for presentation? I hope?
+            return "";
+        case 'notes':
+            // REDCap calls textareas notes
+            return "textarea{@}$redcapfieldname{@}$label";
+        default:
+            throw new \LorisException("Unhandled REDCap type $redcaptype");
+    }
+}
+
+function optionsToLINST(string $dictionary) {
+    $choices = explode(' | ', $dictionary);
+    $linstChoices = [];
+    foreach ($choices as $choice) {
+        $matches = [];
+        if (preg_match("/^(\s)*(\d+)(\s)*,(.*)$/", $choice, $matches) !== 1) {
+            throw new \DomainException("Could not parse radio option: $choice");
+
+        }
+        $backend = $matches[2] . '_' . preg_replace("/\s+/", "_", trim($matches[4]));
+        $linstFormat = "'$backend'=>'" . trim(strtolower($matches[4])) . '\'';
+        $linstChoices[] = $linstFormat;
+
+    }
+    return join('{-}', $linstChoices);
+}
+
+function showHelp() {
+    global $argv;
+    fprintf(STDERR, "Usage: $argv[0] --file=filename --output-dir=instrumentdirectory\n");
+}
+
+function outputFiles(string $outputDir, array $instruments) {
+    foreach ($instruments as $instname => $instrument) {
+        $fp = fopen("$outputDir/$instname.linst", "w");
+        foreach ($instrument as $field) {
+            fwrite($fp, "$field\n");
+        }
+        fclose($fp);
+
+    }
+}
--- a/tools/redcap2linst.php
+++ b/tools/redcap2linst.php
@@ -146,8 +146,11 @@ Zipcode
     case 'dropdown':
         // Radio or dropdown maps to a select and the options are in the
         // same format in the dictionary.
-        return "select{@}$redcapfieldname{@}$label{@}"
-            . optionsToLINST($redcapChoices);
+        $selectoptions = optionsToLINST($redcapChoices);
+        if (!empty($selectoptions)) {
+            $selectoptions = "NULL=>''{-}" . $selectoptions;
+        }
+        return "select{@}$redcapfieldname{@}$label{@}$selectoptions";
     case 'checkbox':
         // checkboxes are the same format as radios but allow multiple options,
         // so map to a selectmultiple instead of a select
--- a/tools/redcap2linst.php
+++ b/tools/redcap2linst.php
@@ -164,8 +164,10 @@ Zipcode
         // Calc maps to a score field. We create the DB field but don't do the score.
         return "static{@}$redcapfieldname{@}$label";
     case 'sql':
-        // The "SQL" data type seems to just be for presentation? I hope?
-        return "";
+        // The "SQL" data type is used for a dynamic display of enum options. Since
+        // we don't have access to the redcap database that the sql is selecting from,
+        // we treat it the same way as a score/calc field so that data can be imported.
+        return "static{@}$redcapfieldname{@}$label";
     case'slider':
         return "numeric{@}$redcapfieldname{@}$label{@}0{@}100";
     case 'file':
--- a/tools/redcap2linst.php
+++ b/tools/redcap2linst.php
@@ -210,14 +210,14 @@ function optionsToLINST(string $dictionary) : string
     $linstChoices = [];
     foreach ($choices as $choice) {
         $matches = [];
-        if (preg_match("/^(\s)*(\d+)(\s)*,(.*)$/", $choice, $matches) !== 1) {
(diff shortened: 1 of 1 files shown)


tools/redcap2linst.php
```diff
@@ -0,0 +1,166 @@
...
+ * @param string $redcapChoices   The choices columnfrom REDCap
+ *
+ * @return string
+ */
+function toLINST(
+    string $redcaptype,
+    string $redcapfieldname,
+    string $redcaplabel,
+    string $redcapChoices,
+) : string {
```
zaliqarosli, line 69:
```
PHP Parse error:  syntax error, unexpected ')', expecting variable (T_VARIABLE) in /var/www/loris/tools/redcap2linst.php on line 69
```

tools/redcap2linst.php
```diff
@@ -0,0 +1,166 @@
...
+        // Calc maps to a score field. We create the DB field but don't do the score.
+        return "static{@}$redcapfieldname{@}$label";
+    case 'sql':
+        // The "SQL" data type seems to just be for presentation? I hope?
+        return "";
+    case 'notes':
+        // REDCap calls textareas notes
+        return "textarea{@}$redcapfieldname{@}$label";
+    default:
+        throw new \LorisException("Unhandled REDCap type $redcaptype");
```
zaliqarosli, line 103:
```
PHP Fatal error:  Uncaught LorisException: Unhandled REDCap type file in /var/www/loris/tools/redcap2linst.php:103
Stack trace:
#0 /var/www/loris/tools/redcap2linst.php(44): toLINST()
#1 {main}
  thrown in /var/www/loris/tools/redcap2linst.php on line 103
```

could we print out the type and fieldname here?

tools/redcap2linst.php
```diff
@@ -0,0 +1,166 @@
...
+            . optionsToLINST($redcapChoices);
+    case 'checkbox':
+        // checkboxes are the same format as radios but allow multiple options,
+        // so map to a selectmultiple instead of a select
+        return "selectmultiple{@}$redcapfieldname{@}$label{@}"
+            . optionsToLINST($redcapChoices);
+    case 'yesno':
+        // Map yes/no fields to dropdowns with yes and no options.
+        return "select{@}$redcapfieldname{@}$label{@}"
+            . "NULL=>''{-}'yes'=>'Yes'{-}'no'=>'No'";
```
zaliqarosli, line 92: i believe 'yesno' in REDCap is actually mapped to 0 and 1 values in the back-end

tools/redcap2linst.php
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * @return string
+ */
+function toLINST(
+    string $redcaptype,
+    string $redcapfieldname,
+    string $redcaplabel,
+    string $redcapChoices,
+) : string {
+    $label = str_replace($redcaplabel, "\n", "<br /><br />");
```
zaliqarosli, line 70: arguments here might be in the wrong place?

tools/redcap2linst.php
```diff
@@ -0,0 +1,166 @@
...
+    case 'calc':
+        // Calc maps to a score field. We create the DB field but don't do the score.
+        return "static{@}$redcapfieldname{@}$label";
+    case 'sql':
+        // The "SQL" data type seems to just be for presentation? I hope?
+        return "";
+    case 'notes':
+        // REDCap calls textareas notes
+        return "textarea{@}$redcapfieldname{@}$label";
+    default:
```
zaliqarosli, line 102: could add 'section header' case and map it to linst "header" elements

wondering also if there's any desire to add pages as well? in my other PR, I used the section header in REDCap to split those sections into pages, and then add the header at the top of the page

zaliqarosli (changes requested):
inital review

tools/redcap2linst.php
zaliqarosli: just noticed the type is already printed out ... i have a 'file' type in my dictionary

zaliqarosli: also have 'slider' type

tools/redcap2linst.php
```diff
@@ -0,0 +1,166 @@
...
+    $linstChoices = [];
+    foreach ($choices as $choice) {
+        $matches = [];
+        if (preg_match("/^(\s)*(\d+)(\s)*,(.*)$/", $choice, $matches) !== 1) {
+            throw new \DomainException("Could not parse radio option: $choice");
+
+        }
+        $backend        = $matches[2] . '_'
+                . preg_replace("/\s+/", "_", trim($matches[4]));
+        $linstFormat    = "'$backend'=>'" . trim(strtolower($matches[4])) . '\'';
```
zaliqarosli, line 127: the way you've done is a sensible choice i think, e.g.:

```
select{@}method_questionnaire16{@}<br /><br />{@}'1_Online/En_ligne'=>'online/en ligne'{-}'2_Over_the_phone/Par_téléphone'=>'over the phone/par téléphone'{-}'3_In-person_during_a_study_visit/En_personne_lors_d'une_visite_d'étude'=>'in-person during a study visit/en personne lors d'une visite d'étude'{-}'4_Over_videoconference/Par_vidéconférence'=>'over videoconference/par vidéconférence'
```

however in COPN's case, they really wanted the option mappings to be this way i.e.:

```
select{@}method_questionnaire16{@}<br /><br />{@}'1'=>'online/en ligne'{-}'2'=>'over the phone/par téléphone'{-}'3'=>'in-person during a study visit/en personne lors d'une visite d'étude'{-}'4'=>'over videoconference/par vidéconférence'
```
with the option labels not made to all be lower-case

tools/redcap2linst.php
zaliqarosli: also, can "\n" just be replaced by one "<br />" instead of two?

tools/redcap2linst.php
driusan: What version of PHP? I think the trailing comma means at least 8.0 is required.

tools/redcap2linst.php
driusan: Are sliders always numeric in redcap? Do we need to parse the dictionary options column or anything? Is the min/max customizable? Maybe you can send me a line from a CSV that's an example of one.

Not sure about file. @ridz1208 do you know if we can do a file upload type in LINST? I don't see one in the code but I thought we had one..

tools/redcap2linst.php
driusan: I think you're right, can you swap the order to test and if it works do a request changes suggestion from the github frontend?

tools/redcap2linst.php
driusan: Do you have an example of a section header? Is there a type or are you talking about watching the column and keeping track of when it changes?

tools/redcap2linst.php
driusan: There's some decisions to make here that maybe we can discuss at the next LORIS meeting. My understanding is that in REDCap all values are integers (which makes it easier for them to do their scoring, I guess), so we need to decide if we want to keep that, or whether we want to map it to something more descriptive for querying.

tools/redcap2linst.php
zaliqarosli: im on 7.4 so this is fine!

tools/redcap2linst.php
zaliqarosli: i'm not sure whether sliders are all numeric! i would assume so but only one of my slider elements explicitly say "numeric" under the "Text Validation Type OR Show Slider Number" column.

```
qod_annoying | questionnaire_for_olfactory_disorders |   | slider | 30. Please use the scale below to rate how annoying the changes in your sense of smell are to you | Not annoying at all \| Extremely annoying |
```

zaliqarosli: the min/max isn't specified in the dictionary, i just know its 0 to 100 cz copn user told me so i would assume its fixed

tools/redcap2linst.php
zaliqarosli, line 70:
```suggestion
    $label = str_replace("\n", "<br /><br />", $redcaplabel);
```

zaliqarosli: gives

```
text{@}date19{@}Questionnaire completed: <br /><br /><br /><br />Questionnaire rempli:
select{@}method_questionnaire16{@}How is the questionnaire completed? <br /><br /><br /><br />Comment le questionnaire est-il rempli?{@}'1_Online/En_ligne'=>'online/en ligne'{-}'2_Over_the_phone/Par_téléphone'=>'over the phone/par téléphone'{-}'3_In-person_during_a_study_visit/En_personne_lors_d'une_visite_d'étude'=>'in-person during a study visit/en personne lors d'une visite d'étude'{-}'4_Over_videoconference/Par_vidéconférence'=>'over videoconference/par vidéconférence'
select{@}lose_thought{@}Do you lose your train of thought?{@}NULL=>''{-}'yes'=>'Yes'{-}'no'=>'No'
select{@}problems_decision{@}Are you having problems making decisions?{@}NULL=>''{-}'yes'=>'Yes'{-}'no'=>'No'
select{@}seeing_things{@}Are you seeing or hearing things that you know or are told are not there?{@}NULL=>''{-}'yes'=>'Yes'{-}'no'=>'No'
```

i think just 1 `<br />` is sufficient though

zaliqarosli (changes requested):

tools/redcap2linst.php
zaliqarosli: section header would be like this:

```
page{@}{@}Attention and Working Memory
header{@}{@}Attention and Working Memory
```

whenever there's an entry in "Section Header" column on redcap data dictionary, i add a page and a header element to translate that section, so yes watching the column and keeping track of when its not empty

tools/redcap2linst.php
```diff
@@ -0,0 +1,166 @@
...
+        // Map yes/no fields to dropdowns with yes and no options.
+        return "select{@}$redcapfieldname{@}$label{@}"
+            . "NULL=>''{-}'yes'=>'Yes'{-}'no'=>'No'";
+    case 'calc':
+        // Calc maps to a score field. We create the DB field but don't do the score.
+        return "static{@}$redcapfieldname{@}$label";
+    case 'sql':
+        // The "SQL" data type seems to just be for presentation? I hope?
+        return "";
+    case 'notes':
```
zaliqarosli, line 99:
```suggestion
    case'slider':
        return "numeric{@}$redcapfieldname{@}$label{@}0{@}100";
    case 'file':
        // File upload - NOT SUPPORTED BY LINST
        return "";
    case 'notes':
```

i just found these leftover on my dev VM so thought i'd add these as other types to add

tools/redcap2linst.php
```diff
@@ -167,6 +167,10 @@ function outputFiles(string $outputDir, array $instruments)
 {
     foreach ($instruments as $instname => $instrument) {
         $fp = fopen("$outputDir/$instname.linst", "w");
+        fwrite($fp, "table{@}$instname\n");
+        // FIXME: There should be a title{@} tag for it to be a valid linst
+        // that the instrument manager can recognize, but there doesn't seem
+        // to be any way to derive that from the redcap data dictionary csv?
```
zaliqarosli, line 173: @driusan when i ran into this issue, i took the table name, split it on the delimiter and changed every first letter to uppercase. would this work for you?

zaliqarosli:
to add date validation, look at column 'Text Validation Type OR Show Slider Number' (8th col) which can have values such as 'date_ymd'. The 9th and 10th columns, 'Text Validation Min' and 'Text Validation Max' respectively, can also have further validation info

xlecours:
this needs to add `{-@-}testname{@}<test_name>` to each linst files

driusan:
Why? The only reference I see to `testname` in NDB_BVL_Instrument_LINST comes from the meta file, not the LINST file.

driusan:
(also I don't think the test name is in the redcap data dictionary)

xlecours:
> Why? The only reference I see to `testname` in NDB_BVL_Instrument_LINST comes from the meta file, not the LINST file.

The issue is with [tools/exporters/data_dictionary_builder.php](https://github.com/aces/Loris/blob/main/tools/exporters/data_dictionary_builder.php#L87-L100) which loops through ip_output looking for testname

zaliqarosli:
@xlecours this is fixed somewhat by https://github.com/aces/Loris/pull/8434
@driusan that PR is assigned to you to sign off because the solution is a bit of a quick fix that @ridz1208 and I thought up. if it's all good with you, then we won't need to add `{-@-}testname{@}<test_name>` as xavier suggested

tools/redcap2linst.php
zaliqarosli: @driusan why is a new line replaced by two line breaks instead of just one?

tools/redcap2linst.php
zaliqarosli: @driusan can you confirm this is the case for HBCD? @xlecours would you know?

tools/redcap2linst.php
```diff
@@ -36,17 +38,37 @@
...
     if (!isset($instruments[$inst])) {
         $instruments[$inst] = [];
     }
-    $linstFormat = toLINST($row[3], $row[0], $row[4], $row[5]);
+    $fieldname = $row[0];
+    if ($trimName) {
+        $formname = $row[1];
+        if (strpos($fieldname, $formname) !== 0) {
+            $badMap++;
+            fwrite(STDERR, "Field $fieldname does not have form name $formname as a prefix\n");
```
zaliqarosli, line 63: @driusan i feel like this is very project specific. it's not a LORIS requirement to prefix instrument name to the field name, this won't work for maybe projects outside of HBCD i would guess

tools/redcap2linst.php
```diff
@@ -36,17 +38,37 @@
...
+        } else {
+            // debug
+            //$oldname = $fieldname;
+            $mapped++;
+            $fieldname = preg_replace("/^" . preg_quote($formname) . "(_*)/", "", $fieldname);
```
zaliqarosli, line 68: @driusan does this just remove all underscores?

tools/redcap2linst.php
driusan: it's specific to the `--trim-formname` option?

LORIS prefixes the instrument name to the field name to make it unique, while redcap has a requirement that fieldnames be unique across all instruments, so if it's not  trimmed it gets duplicated in LORIS, which is why the `--trim-formname` option exists. 

It does apply to HBCD but I would expect it to be the most common naming convention used to meet redcap's globally unique fieldname requirement and this code only applies if the option for it is explicitly passed.

tools/redcap2linst.php
driusan: It removes the form name from the start of the field name and removes the underscores that are left at the front after trimming (ie if it's the bmi and the field is bmi_weight_kg it trims `bmi_` instead of `bmi` so that the fieldname that's left isn't `_weight_kg`

tools/redcap2linst.php
zaliqarosli: i see, sorry i misunderstood, i thought it was doing the opposite, got it!

zaliqarosli:
@driusan this PR is failing tests, probably just needs a rebasing. do you have some time to take a look at the unresolved comments before I leave?

regisoc:
@driusan I think we can close this PR.  
The redcap2linst has been modified and integrated in the #9474 as a new tool.

regisoc:
@ridz1208 @driusan closing it.

PR #8201, opened 2022-10-27
Fix for tools/lorisform_parser

regisoc:
Following #7446, one of the changes involves `php/libraries/NullModule.class.inc` that now requires a `LorisInstance` type parameter.

This fix adds a missing parameter in the `lorisform_parser.php` script.


driusan (approved):
Will merge after the tests pass

maltheism (approved):

PR #8202, opened 2022-10-27
[JSX] Add InfoPanel component

driusan:
This adds an InfoPanel component to show a message to the user in a box that's styled to draw their attention. The "alert alert-info" bootstrap classes weren't working for me with our current stylesheet, so the InfoPanel wraps around what's needed to make it look reasonable and uses a FontAwesome icon to mark it as an information panel.

Usage:
```
<InfoPanel>
Message that should go in the panel, can include arbitrary other react or html if you want.
</InfoPanel>
```

It looks like this

InfoPanel example

### DIFF
--- a/jsx/InfoPanel.js
+++ b/jsx/InfoPanel.js
@@ -1,0 +1,33 @@
+/**
+ * Display a message in an information panel.
+ *
+ * @param {object} props - React props
+ *
+ * @return {JSX}
+ */
+function InfoPanel(props) {
+    return (
+        <div className="alert alert-info"
+           style={{
+             backgroundColor: '#BEDFFF',
+             color: 'black',
+             display: 'flex',
+             flexBasis: 'row',
+             flexWrap: 'nowrap',
+         }}>
+            <div style={{
+                alignSelf: 'center',
+                padding: '1em',
+            }}>
+                <i style={{fontSize: '2em', color: 'blue'}}
+                    className="fas fa-info-circle"></i>
+            </div>
+            <div style={{
+                paddingLeft: '1em',
+                alignSelf: 'center',
+            }}>{props.children}</div>
+         </div>
+   );
+}
+
+export default InfoPanel;


jsx/InfoPanel.js
```diff
@@ -0,0 +1,33 @@
...
+    return (
+        <div className="alert alert-info"
+           style={{
+             backgroundColor: '#BEDFFF',
+             color: 'black',
+             display: 'flex',
+             flexBasis: 'row',
+             flexWrap: 'nowrap',
+         }}>
+            <div style={{
```
maltheism, line 18: Hi @driusan it's better to use <span> tag instead of <div> tag here.

jsx/InfoPanel.js
```diff
@@ -0,0 +1,33 @@
...
+                alignSelf: 'center',
+                padding: '1em',
+            }}>
+                <i style={{fontSize: '2em', color: 'blue'}}
+                    className="fas fa-info-circle"></i>
+            </div>
+            <div style={{
```
maltheism, line 25: Hi @driusan it's better to use <span> tag instead of <div> tag here.

jsx/InfoPanel.js
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @param {object} props - React props
+ *
+ * @return {JSX}
+ */
+function InfoPanel(props) {
+    return (
+        <div className="alert alert-info"
+           style={{
+             backgroundColor: '#BEDFFF',
```
maltheism, line 12: Hi @driusan just something to keep in mind that inline styling has worse performance than using className. See https://reactjs.org/docs/faq-styling.html#are-inline-styles-bad

maltheism (changes requested):
See the two comments about changing the div tags to span tags.

jsx/InfoPanel.js
driusan: Why would `span` be better than `div`? They're block elements, not inline.

zaliqarosli: i'm not sure if this is an opinion that i have - i think div tags are fine to use!

maltheism (approved):

jsx/InfoPanel.js
driusan: That's interesting because it's the exact opposite of the advice that react gave when it was first released and they said to use inline styling so that components are self-contained.. but we're not hitting any performance bottlenecks yet.

PR #8207, opened 2022-11-01
[Candidate DTO] Add CandidateData DTO object

driusan:
This adds a CandidateData DTO object which mirrors the TimepointData which already exists. It is the Candidate equivalent of PR#7432.

A CandidateData object is a DTO (data transfer object) to
store a representation of data associated with a Candidate
object in LORIS.

The CandidateData may be partially loaded, in which case any
getter trying to access an unknown will throw an exception.
This allows modules to only load relevant data in a query,
and avoid the overhead of the Candidate::singleton() constructor.

### DIFF
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -35,8 +35,20 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource,
     \LORIS\StudyEntities\SiteHaver
 {
     var $candidateInfo;
-    var $listOfTimePoints;
 
+    protected $data;
+
+    /**
+     * Construct a new Candidate object optionally populated by
+     * the given DTO.
+     *
+     * @param ?CandidateData $data a Candidate DTO object to bypass
+     *                             the singleton constructor.
+     */
+    public function __construct(?CandidateData $data=null)
+    {
+        $this->data = $data;
+    }
     /**
      * This returns a single Candidate object for the given
      * candID, and will only ever instantiate one object in
@@ -99,9 +111,18 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource,
             );
         }
 
+        // We need to refer to these when instantiating the CandidateData
+        // object after Timepoints have been selected, but by then $row
+        // will have been overwritten, so put them in other variables..
+        $registrationProjectID = new \ProjectID($row['RegistrationProjectID']);
+        $registrationCenterID  = $row['RegistrationCenterID'];
+
         // store user data in object property
         foreach ($row as $key=>$value) {
             switch ($key) {
+            case 'RegistrationProjectID':
+            case 'RegistrationCenterID':
+                break;
             case 'Sex':
                 $this->candidateInfo[$key] = empty($value) ? null : new Sex($value);
                 break;
@@ -118,8 +139,6 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource,
                 $this->candidateInfo[$key] = $value;
             }
         }
-
-        $this->candidateInfo['ProjectTitle'] = $this->getProjectTitle();
 
         $headerSetting = $config->getSetting('HeaderTable');
         if (!empty($headerSetting)) {
@@ -176,16 +195,29 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource,
         }
 
         // get Time Point SessionIDs for the $candID
-        // select ALL of them - later you can filter out cancelled and inactive ones
-        $query = "SELECT s.ID FROM session as s
+        $query = "SELECT s.ID, s.ProjectID, s.CenterID FROM session as s
             WHERE s.CandID=:Candidate AND s.Active='Y'
             ORDER BY ID";
         $row   = $db->pselect($query, $CandArray);
 
         // store user data in object property
+        $timepoints = [];
         foreach ($row as $value) {
-            $this->listOfTimePoints[] = intval($value["ID"]);
+            $timepoints[] = new TimePoint(
+                new TimePointData(
+                    new SessionID($value["ID"]),
+                    new ProjectID($value["ProjectID"]),
+                    new CenterID($value["CenterID"]),
+                )
+            );
         }
+
+        $this->data = new CandidateData(
+            $registrationProjectID,
+            new \CenterID($registrationCenterID),
+            $timepoints,
+        );
+        $this->candidateInfo['ProjectTitle'] = $this->getProjectTitle();
     }
 
     /**
@@ -374,7 +406,12 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource,
      */
     function getListOfTimePoints(): array
     {
-        return $this->listOfTimePoints ?? [];
+        return array_map(
+            function ($timepoint) {
+                return intval($timepoint->getSessionID());
+            },
+            $this->data->getTimePoints(),
+        );
     }
 
     /**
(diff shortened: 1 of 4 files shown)


php/libraries/CandidateData.class.inc
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * Construct a CandidateData object
+     *
+     * @param ?\ProjectID  $project    The RegistrationProjectID
+     * @param ?\CenterID   $centerID   The RegistrationCenterID
+     * @param ?TimePoint[] $timepoints TimePoint objects for each visit of the
+     *                                 candidate
+     */
+    public function __construct(
+        ?\ProjectID $project,
```
ridz1208, line 40: why not `$projectID`

driusan: sure.. I started working on this before named parameters were a thing and just got around to trying to send a PR and wrap it up now.

`$projectID` or `$registrationProjectID`?

php/libraries/CandidateData.class.inc
```diff
@@ -0,0 +1,88 @@
...
+ * `$candidate->isAccessibleBy($user)` without having to load all the
+ * other data that Candidate::singleton() does. If an attempt is
+ * made to access a property not in the DTO, an exception is thrown,
+ * rather than silently returning the incorrect value.
+ *
+ * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ */
+class CandidateData
+{
+    protected $registrationProjectID;
```
ridz1208, line 27: I would type these variables, and make them public. You can then get rid of all you getters I believe because unless they are initialized, they will throw an error if accessed

driusan: No, that doesn't work.. then you need to pass all of them and there's no way to partially construct the object

ridz1208: you can set them one by one since they are public. if you want the constructor, you can set if statements that dont set the variable if its passed a null value and then set it manually ?

driusan: done

php/libraries/CandidateData.class.inc
```diff
@@ -0,0 +1,57 @@
...
+    ) {
+        if ($projectID !== null) {
+            $this->registrationProjectID = $projectID;
+        }
+        if ($registrationCenterID !== null) {
+            $this->registrationCenterID  = $registrationCenterID;
+        }
+        if ($this->timepoints !== null) {
+            $this->timepoints = $timepoints;
+        }
```
ridz1208, line 53: I think you should loop through the array mand make sure they are instanceOf timepoint class

driusan: ok, done

php/libraries/Candidate.class.inc
```diff
@@ -99,9 +111,18 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource,
             );
         }
 
+        // We need to refer to these when instantiating the CandidateData
+        // object after Timepoints have been selected, but by then $row
+        // will have been overwritten, so put them in other variables..
+        $registrationProjectID = new \ProjectID($row['RegistrationProjectID']);
+        $registrationCenterID  = $row['RegistrationCenterID'];
```
ridz1208, line 118:
```suggestion
        $registrationCenterID  = new \CenterID($row['RegistrationCenterID']);
```

php/libraries/Candidate.class.inc
```diff
@@ -176,16 +195,29 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource,
...
+                    new SessionID($value["ID"]),
+                    new ProjectID($value["ProjectID"]),
+                    new CenterID($value["CenterID"]),
+                )
+            );
         }
+
+        $this->data = new CandidateData(
+            $registrationProjectID,
+            new \CenterID($registrationCenterID),
```
ridz1208, line 217:
```suggestion
            $registrationCenterID,
```

php/libraries/Candidate.class.inc
```diff
@@ -374,7 +406,12 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource,
      */
     function getListOfTimePoints(): array
     {
-        return $this->listOfTimePoints ?? [];
+        return array_map(
+            function ($timepoint) {
+                return intval($timepoint->getSessionID());
+            },
+            $this->data->getTimePoints(),
```
ridz1208, line 413: I think we removed this function

php/libraries/Candidate.class.inc
```diff
@@ -387,8 +424,9 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource,
      */
     function getSessionID(int $visitNo): ?int
     {
-        if (isset($this->listOfTimePoints[$visitNo-1]) ) {
-            return intval($this->listOfTimePoints[$visitNo-1]);
+        $timepoints = $this->data->getTimePoints();
```
ridz1208, line 427: getTimePoints() was removed

php/libraries/Candidate.class.inc
```diff
@@ -905,20 +943,15 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource,
...
-        // have access to any sites. If we have access to a site,
-        // we need to be able to access the candidate to get to it.
-        $timepoints = $this->getListOfTimePoints();
-        foreach ($timepoints as $sessionID) {
-            $timepoint = \TimePoint::singleton(
-                new \SessionID(strval($sessionID))
-            );
-
+        // have access to any timepoints.
+        foreach ($this->data->getTimePoints() as $timepoint) {
```
ridz1208, line 947: getTimePoints() was removed

ridz1208 (commented):
Once the changes are done, I think the rest makes sense but I have not check the entirety of the candidate.class to make sure no other places need to be replaced by the DTO

ridz1208 (changes requested):
I think we may want to put the classes in the right place as well? src/studyentities? or src/data/models?

php/libraries/CandidateData.class.inc
```diff
@@ -0,0 +1,62 @@
...
+ * rather than silently returning the incorrect value.
+ *
+ * @license http://www.gnu.org/licenses/gpl-3.0.txt GPLv3
+ */
+class CandidateData
+{
+    public \ProjectID $registrationProjectID;
+    public \CenterID $registrationCenterID;
+
+    public array $timepoints;
```
ridz1208, line 30: I know this is not the endproduct but I think this object should at least have a candID. readonly probably

php/libraries/Candidate.class.inc
driusan: yeah phan caught it too now that tests are running again.

php/libraries/CandidateData.class.inc
driusan: I added a CandID but not readonly because that would up the minimum PHP version to 8.1.. also if that was readonly I think all of them should be.

ridz1208:
@driusan I'm fine with moving them later, but do we know where the correct location is? are we still pushing to put things in the `src/` directory?

driusan:
Things should go in the src directory if:

1. They're designed into proper, consistent namespaces 
2. They can't be overridden

I didn't know the src/Data/Models directory existed and don't think it belongs there because it has nothing to do with the the data framework that the src/Data namespace is managing.

StudyEntities could make sense, or we could also create a new namespace for DTOs and move them all there.

ridz1208 (approved):

PR #8209, opened 2022-11-02
[integration test]check react page load

kongtiaowang:
## Brief summary of changes
Make sure react pages load well.
It will find the page has error.
截屏2022-11-02 上午11 13 01
#### Testing instructions 
Automation test will pass gitactions test.


cmadjar:
@shonibare could you review/test please? Thanks!

cmadjar:
@kongtiaowang can you look into that? Thanks!

shonibare:
@cmadjar I have reviewed again and it works

xlecours (commented):
This will only work for as long as the error messages don't change.
We would need to add a test to find the exact error message for when they are legitimate. 
Instead, I think the test should check for a distinctive element in the DOM or count a number of elements (e.g.: input fields).
Bonus if the test error contains the bodytext so we can see what the html page was when the error occured.

PR #8215, opened 2022-11-04
QueryEngine interface

driusan:
This defines the QueryEngine interface that I've been using to build a new, CouchDB-free DQT as well as used to create a new candidates module. It's stable enough that I'm confident it will require little modification now.

```

/**
 * A QueryEngine is an entity which represents a set of data and
 * the ability to query against them.
 *
 * Queries are divided into 2 phases, filtering the data down to
 * a set of CandIDs, and retrieving the data for a known set of
 * CandID/SessionIDs.
 *
 * There is usually one query engine per module that deals with
 * candidate data.
 */
```

After this is merged, the QueryEngine()->getDataDictionary() for a module will replace the getDataDictionary() method that currently exists and I'll begin sending individual engines as they're completed and well tested. (I currently have instruments, candidate parameters, imaging, and bvl feedback that I've been using for the DQT but the latter two still need significant testing and refactoring.)

Other than the `QueryEngine` interface, the rest of the classes in this PR are classes that the interface depends on (ie are used by the interface.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ changes in the following format: PR #1234***
 ## LORIS 25.0 (Release Date: ????-??-??)
 ### Core
 #### Features
-- placeholder 
+- Added new interface intended to be used for querying module data from PHP (PR #8215) 
 
 #### Updates and Improvements
 - Rename subproject to Cohort (PR #7817)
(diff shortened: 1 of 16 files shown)


maltheism (commented):
@driusan the interface looks great. Should the CHANGELOG.md reference it?

driusan:
I don't have any strong opinions on if the new interface should be mentioned in the CHANGELOG before the rest of the new DQT architecture is done. @ridz1208 do you have an opinion?

ridz1208:
@driusan mention it in improvements, nothing to lose

driusan:
@maltheism @ridz1208 rebased and added to the changelog

maltheism (approved):

PR #8219, opened 2022-11-07
[dataquery] Define new swagger schema for data query API

driusan:
This defines the schema I've been using to design a QueryEngine (rather than CouchDB) based data query tool.

The meaning of the 3 types of queries in the /queries endpoints are:
1. `recent`: There are no more "saved queries". Instead, all recent queries are "saved" and can be reloaded. Users can star or name a query in order to indicate that they're important queries that shouldn't be purged and duplicate the functionality of the old saved queries. (Currently, no queries are ever purged, but if we do change that in the future, starred ones are the equivalent of saved ones.)
2. `shared`: Shared queries, similar to existing shared queries. These queries are public for everyone.
3. `topqueries`: Admin queries that were pinned to the top of the DQT. Similar to shared queries, but pinned by an admin who think they're important.

I believe the rest should be explained by the swagger schema descriptions

### DIFF
--- a/modules/dataquery/static/schema.yml
+++ b/modules/dataquery/static/schema.yml
@@ -80,36 +80,32 @@ paths:
           style: spaceDelimited
           schema:
             type: boolean
-          
-        - name: unshare
-          in: query
-          style: spaceDelimited
-          schema:
-            type: boolean
+          description: if true, the query will be shared. If false, it will be unshared
         - name: star
           in: query
           style: spaceDelimited
+          description: if true, the query will be shared. If false, it will be unshared
           schema:
             type: boolean
-        - name: unstar
+        - name: adminname
           in: query
-          style: spaceDelimited
-          schema:
-            type: boolean
-        - name: type
+          style: pipeDelimited
+          description: The admin name to pin the query as. If the empty string, will be unpinned.
+        - name: dashboardname
           in: query
-          style: spaceDelimited
-          schema:
-            type: string
-            enum: ['top', 'untop', 'dashboard']
+          style: pipeDelimited
+          description: The admin name to pin the query to the dashboard as. If the empty string, will be unpinned.
         - name: name
           in: query
           style: pipeDelimited
+          description: The name to set for the query for this user.
           schema:
             type: string
       responses:
-        '400':
-          description: Bad request. Most likely caused by one or more incompatible patches to be done at the same time.
+        '200':
+          description: The access was performed
+        '403':
+          description: The access was not performed because of permissions being denied.
   /queries/{QueryID}/run:
     parameters:
       - name: QueryID
@@ -119,7 +115,7 @@ paths:
         style: simple
         schema:
           type: integer  
-    get:
+    post:
       description: |- 
         Run the query QueryID and returns the results. 
         
@@ -255,6 +251,10 @@ components:
           type: integer
           description: A reference to an object in the queries property identified by QueryID
           example: 3
+        QueryRunID:
+          type: integer
+          description: A reference to the run number of this query
+          example: 4
     QueryField:
       type: object
       properties:
--- a/modules/dataquery/static/schema.yml
+++ b/modules/dataquery/static/schema.yml
@@ -123,7 +123,7 @@ paths:
       description: |- 
         Run the query QueryID and returns the results. 
         
-        This endpoint will result in a new query run being generated, which will be returned in the 'recent' queries of the user on the /queries endpoint.
+        This endpoint will result in a new query run being generated, which will be returned in the queries of the user on the /queries endpoint.
       responses:
         '200':
           description: The query was able to be successfully run
@@ -196,24 +196,35 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/QueryRun'
-        shared:
+        queries:
           type: array
           items:
             $ref: '#/components/schemas/Query'
-        topqueries:
-          type: array
-          items:
-            $ref: '#/components/schemas/Query'
-
     Query:
       type: object
       properties:
+        self:
+          type: string
+          description: |-
+             A URL that this query can be accessed at.
+
+             Accessing the query directly through the URL is sure to have the same fields
+             and criteria, but other details of the object returned may not be identical.
+
+             For instance, the starred and name properties may vary based on the user
+             accessing the URI.
+          example: "https://example.com/dataquery/queries/4"
         type:
           type: string
           enum: ['candidates']
           example: "candidates"
+        AdminName:
+          type: string
+          descripton: The name given by the admin for a pinned query
+          example: "Important Study Query Of Missing T1s"
         Name:
           type: string
+          description: The name given by the current user for this query
           example: "My Query"
         SharedBy:
           type: array
@@ -222,8 +233,13 @@ components:
             example: "admin"
         Starred:
           type: boolean
-        Shared:
+          description: The query has been starred by the user
+        Public:
           type: boolean
+          description: The query has been shared (made public to all users)
+        Pinned:
+          type: boolean
+          description: The query has been pinned by an administrator
         QueryID:
           type: integer
           example: 3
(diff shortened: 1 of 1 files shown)


modules/dataquery/static/schema.yml
```diff
@@ -0,0 +1,331 @@
...
+    This is the API for the parts of the Loris data query module which are intended to be used as an API.
+  contact:
+    name: LORIS Development Team
+    url: 'https://github.com/aces/Loris'
+  license:
+    name: 'GNU Public License, Version 3'
+    url: 'https://opensource.org/licenses/GPL-3.0'
+  version: "3.0"
+servers:
+  - url: /dataquery/
```
maltheism, line 14: @driusan should it be dqt?

driusan:
I'm planning on sending a PR with the php-side implementation of this to make it easier to test, but it's going to be blocked by #8260 and #8216 (or at least one other module implementing the interface..)

modules/dataquery/static/schema.yml
```diff
@@ -0,0 +1,331 @@
...
+        '500':
+          description: Not implemented
+          content:
+            application/json:
+              schema:
+                type: object
+            
+components:
+  schemas:
+    AllQueries:
```
xlecours, line 192: Not sure why there are 3 properties here.
They can be known by checking the `Stared` and `Shared` properties of each queries. I would add `Top` or `Pinned` for the third one.

modules/dataquery/static/schema.yml
```diff
@@ -0,0 +1,331 @@
...
+            $ref: '#/components/schemas/QueryField'
+        criteria:
+          $ref: '#/components/schemas/QueryCriteriaGroup'
+      required:
+        - type
+        - fields
+    QueryRun:
+      type: object
+      properties:
+        RunTime:
```
xlecours, line 242: Could this  be add to the Query component so that /queries returns a list of Query instead?

modules/dataquery/static/schema.yml
```diff
@@ -0,0 +1,331 @@
...
+          items:
+            $ref: '#/components/schemas/Query'
+        topqueries:
+          type: array
+          items:
+            $ref: '#/components/schemas/Query'
+
+    Query:
+      type: object
+      properties:
```
xlecours, line 210: I would add a `Link` or `Self` property which would be this query's url. When returned in /queries, it would help a client access it directly without building the url by itself. 
One of the REST principal suggests:

> Hypermedia as the engine of application state – The client should have only the initial URI of the application. The client application should dynamically drive all other resources and interactions with the use of hyperlinks.

xlecours (changes requested):

modules/dataquery/static/schema.yml
driusan: Ok

modules/dataquery/static/schema.yml
driusan: You are probably right. I'll have to look into the backend code to see if this is just because of how the code evolved or if there's some reason related to efficiency that I did this.

modules/dataquery/static/schema.yml
xlecours: So a query could be build then saved without ever being run. That would result in a null value for RunTime which would still hold meaning.

xlecours: So recent refers to RunTime oor Creation/Save time?

driusan: RunTime

modules/dataquery/static/schema.yml
```diff
@@ -0,0 +1,353 @@
...
+      responses:
+        '200':
+          description: The Query was successfully retrieved
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Query'
+        '403':
+          description: Permission denied
+    patch:
```
xlecours, line 76: Can I send: 
share=true&unshared=true
?

modules/dataquery/static/schema.yml
```diff
@@ -0,0 +1,353 @@
...
+  /queries/{QueryID}/run:
+    parameters:
+      - name: QueryID
+        in: path
+        description: the QueryID returned by posting to /queries
+        required: true
+        style: simple
+        schema:
+          type: integer  
+    get:
```
xlecours, line 122: I think post would make more sense. This is creating a queryRun so I don't think it is safe (as in rfc2616 safe)

xlecours: returning 201 Created with location header set to  `/queries/{QueryID}/run/{QueryRunID}` would be good as well.

xlecours:
> I didn't change the wording of the endpoint "queries" to say "all" because that might not scale if we get to a point where the recent queries displayed need to be limited.

I suggest you add `/queryRun` endpoint to get the run out of `/queries`. If there is too many run then implementing pagination would be an option then.

```
GET /queries
{
  "Queries": []
}

GET /queryRun
{
  QueryRun: []
}
```

modules/dataquery/static/schema.yml
driusan: how do you denote "Location header" in swagger?

driusan: actually that can't be done right now because that end point is not implemented so it would mean it's impossible to get results. Once it is implemented, the schema could be updated.

modules/dataquery/static/schema.yml
driusan: previously, yes, but you would get a bad request response. After latest update: no.

driusan:
I don't think a QueryRuns endpoint wouldn't solve anything related to performance. Checking the access is expensive because it needs to go through every field and module of the query's fields and criteria. Sending it over the wire is cheap. A separate endpoint (or pagination) would just add overhead to loading the list.

modules/dataquery/static/schema.yml
xlecours: why not use share=true|false and not use unshare?

modules/dataquery/static/schema.yml
xlecours: '201':
          description: The visit was created successfully
          headers:
            Content-Location:
              description: The URL of the new visit.
              style: simple
              explode: false
              schema:
                type: string

xlecours:
> I don't think a QueryRuns endpoint wouldn't solve anything related to performance.  
It would remove the burden of loading the runs along the queries when the client request the list of existing queries.

driusan:
Sending query runs isn't a burden. It's sending a small amount of data directly from the database over the wire. Checking if a query is accessible by a user *is* slow, however, so, "all queries accessible by the user" is much slower than "queries which there is some reason to send to the user". The performance hit comes from the access check for queries, not the inclusion of a query run list.

xlecours:
I don't know about the implementation.
Getting the queries and the queryRun must take A + B microseconds. Getting only the queries can't take longer than getting both.

driusan:
Getting all queries accessible by the user takes n*accesschecktime seconds. There is no reason to include queries other users have run in the queries list just because they happen to have access to everything used in the query. That increases it to n*accesschecktime*numusers (roughly) seconds. The user only needs a subset of queries. The ones they've run, the ones that have been shared because they're interesting, the ones that an admin pinned, etc, not all queries on the server. Including all accessible queries makes the response unuseably slow. Including the times that they've been run has negligible impact on the performance.

xlecours:
Ok I get this. But I don't see how including queryRun in the results of /queries would solve any of that.

driusan:
It doesn't. Leaving the description "Get a list of a recent, shared, and study (top) queries to display." instead of "All queries accessible to the user" solves that.

Thinking about it a little more, I see how pagination could be useful for run queries. I'll separate them out on Monday.

modules/dataquery/static/schema.yml
driusan: that is what's done in the latest update

driusan:
@xlecours the queries and runs are split into two different endpoints now.

xlecours (approved):
Good! 
The only thing I wasn'T sure about is  QueryRun having a link to the Query. At this point, it will be a suggestion, not a change request.
I have eager to see the implementation.

driusan:
Added QueryURI to the QueryRun object.

driusan:
@xlecours the implementation is already in #8268, just updated it to the latest changes..

driusan:
@maltheism you reviewed an earlier draft. Are you okay with this version of the schema?

maltheism (approved):

PR #8221, opened 2022-11-07
[LorisInstance] add getModule function

driusan:
LorisInstance already has functions to check if a module exists, get the module directories, etc.

This adds a getModule to replace \Module::factory() so that the code dealing with modules can be centralized in one place.

### DIFF
--- a/src/LorisInstance.php
+++ b/src/LorisInstance.php
@@ -76,7 +76,7 @@ class LorisInstance
         $modules = [];
         foreach ($mnames as $name) {
             try {
-                $mod       = \Module::factory($this, $name);
+                $mod       = $this->getModule($name);
                 $modules[] = $mod;
             } catch (\LorisModuleMissingException $e) {
                 error_log($e->getMessage() . " " . $e->getTraceAsString());
--- a/src/LorisInstance.php
+++ b/src/LorisInstance.php
@@ -76,7 +76,8 @@ class LorisInstance
         $modules = [];
         foreach ($mnames as $name) {
             try {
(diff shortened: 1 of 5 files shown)


src/LorisInstance.php
```diff
@@ -76,7 +76,8 @@ public function getActiveModules(): array
         $modules = [];
         foreach ($mnames as $name) {
             try {
-                $modules[] = \Module::factory($this, $name);
+                $mod       = \Module::factory($this, $name);
+                $modules[] = $mod;
```
ridz1208, line 80: shouldnt this use getModule() ??

ridz1208 (commented):
are you planning on replacing the existing instances of module::factory? otherwise it will never be replaced, we would just be adding another way of doing it

driusan:
Yes, I'm planning on removing completely removing \Module::factory calls in a different PR. I removed the ones that are in standard LORIS endpoints in the latest commit since they all had access to a LorisInstance object.

When the code is it's moved, it'll use $loris->getModuleDirs() instead of hardcoded paths. So I left the calls in scripts or ajax endpoints which will break since moduleDirs isn't properly set right now and they just use an empty [] directory stub.

I also left the ones that are just using `\Module::factory` for the side-effect of registering the autoloader. I think when the loading code is moved to getModule() I'll likely decouple the side-effect of the autoloader from the getting (ie. add something like a `$module->registerAutoloader()` function and it'll help identify the places that need to be updated.

ridz1208 (approved):

PR #8224, opened 2022-11-08
[npm] reproducible builds

laemtl:
Prevent package-lock.json to be automatically regenerated with `npm install` to have more stable builds.
- enable package-lock.json tracking (recommended practice)
- replace `npm install` with `npm ci`  to achieve fast and reproducible builds
- strict packages versions from package-lock.json are now used for install. To propagate a change made into package.json, run `npm install` and commit the new package-lock.json.
- any version disagrement between package.json and package-lock.json will trigger an error.


Makefile
```diff
@@ -11,7 +11,7 @@ phpdev:
 	composer install
 
 javascript:
-	npm install
+	npm ci
```
maltheism, line 14: Hi @laemtl, this change breaks the development build process of `make clean && make dev` and I think we should still have `npm install` for the development build process but using `npm ci` for both production and testing on github. I think it's beneficial to know when the codebase starts breaking from npm install then to encounter it much later.

docs/wiki/00_SERVER_INSTALL_AND_CONFIGURATION/01_LORIS_Install/Ubuntu/README.md
```diff
@@ -24,7 +24,7 @@ Additionally, the following package manager are required to build LORIS:
 
 * NodeJS 16.10.0 (or higher)
 
-* NPM
+* NPM 5.7.0 (or higher)
```
maltheism, line 27: Hi @laemtl, I think now could be a good time to update the recommended NodeJS version to the current LTS (long-term-support) version and that uses the npm version 8.19.2.

maltheism (changes requested):

Makefile
laemtl: Thanks for pointing this out, the package-lock files are not preserved when make clean is executed.

laemtl: As for having `npm install` for the development build process, I would prefer to have distinct processes for compiling/installing. make dev is intended to compile the code in a dev env, so the package tree shouldn't be modified and npm ci is a better option since it's faster. To install/update packages, npm install can still be run separately but in that case, it is expected that compilation can potentially fail.

maltheism: My understanding is that `npm ci` isn't always faster than `npm install` in the case of if node_modules not being present. Example `npm install` uses the existing node_modules similar to cache and while `npm ci` removes node_modules while generating it based on `package-lock.json`.

I agree the change going forward is a matter of personal taste. My development preference has always been to keep dependencies updated as frequently as possible and similar for the version of nodejs I use for development. Reason being that if the development build process breaks when frontend packages are updated then it's easier to debug what change in the package is causing the new error and then alert the dependency maintainer including looking at new issues on the dependencies repository for finding a solution sooner; which is typically easier than trying to find a solution much later whenever we get to updating the dependencies. The benefits of using the latest in my opinion is for improved bugfixes, stability and security. I also like to follow the latest nodejs & dependencies versions to observe the new increment of features being added. I also believe the downside of locking the development process to fixed packages on the frontend for development  is comparable to if we locked down Loris on the backend to php version 7 for developpment. The codebase typically become harder to update as time progresses while we're not updating dependencies or versions of php. That's why I think it makes sense to lock down production dependencies and our test suite on github.

@laemtl @driusan thoughts?

maltheism (approved):

driusan:
Can you rebase? I just want to be sure it still passes after I just merged the 24.1->main PR.

PR #8225, opened 2022-11-09
[SQL] Make release patch compatible with mariadb 10.3

ridz1208:
An error was reported where the query being modified returns an SQL error because of the missing `FROM` clause on mariaDB 10.3

This fix was taken from the MAriaDB official documentation to fix the issue https://mariadb.com/kb/en/dual/


cmadjar (approved):
LGTM 👍

PR #8226, opened 2022-11-09
Push 24.1.0 release into main

ridz1208:
DO NOT SQUASH


ridz1208:
so that means that we should never commit the composer.lock if we ran composer update with 8.1

PR #8230, opened 2022-11-15
[Webpack] Viz status check fix

laemtl:
To prevent a compilation error when protobuf is not installed, we check if the compiled protobuf file exists to turn 'ON' the visualization. Fix a bug that turns 'OFF' the viz when the EEG Browser is overridden in project/

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -59,6 +59,7 @@ if (!fs.existsSync(
 ) {
   mod.rules.push({
     test: /react-series-data-viewer\/src\/chunks/,
+    exclude: /project/,
     use: 'null-loader',
   });
 }


driusan:
The diff for this doesn't seem to match the description?

laemtl:
If the module is overridden, the file `/modules/electrophysiology_browser/jsx/react-series-data-viewer/src/protocol-buffers/chunk_pb.js` do not exist since it is located in `/project/modules/electrophysiology_browser/jsx/react-series-data-viewer/src/protocol-buffers/chunk_pb.js`. In that case we don't want the file inside project/... /react-series-data-viewer\/src\/chunks/ to be ignored during compilation.

jeffersoncasimir (approved):
I confirm that this works and is needed by EEGNet

laemtl:
This will also be needed for HBCD

PR #8232, opened 2022-11-17
[Tools] delete physiological files

laemtl:
- Script to delete all the files/db entries created by an eeg file ingestion.
  Usage: php delete_physiological_file.php [PID] confirm|tosql
  Split files not supported at the moment.

- Move deletion tools into data_integrity/data_deletion


regisoc (commented):
Seems it lacks a table for the physiological script: `physiological_annotation_archive`.

regisoc (approved):

PR #8234, opened 2022-11-22
Annotations

laemtl:
- includes the missing piece for the [EEGBrowser] annotation features to work 
- fix a glitch with clippath
- truncate the epoch timing values


christinerogers:
this partially does what Jess Callegaro's PR to Main did:  #7829 
HBCD needs this version, so we will have reconcile at some point, or close/re-implement the work on EEGNet.

PR #8244, opened 2022-11-30
npm postinstall script

laemtl:
Following a discussion with @regisoc,
this script defines a set of paths where a package.json with additional dependencies can potentially exists.
This script is executed with npm install (postinstall step).

This fixes several issues:
- allow automated installation of /project additional npm dependencies;
- allow to automatically detect if a path is inside modules and overridden in project, and amend the path with project/

Potential improvements: Migrating the [EEG Browser install step](https://github.com/aces/Loris/tree/main/modules/electrophysiology_browser#-installation-requirements-to-use-the-visualization-features) in this script with a config Enable_EEGBrowser_visualization set to yes/no.

### DIFF
--- a/npm-postinstall.js
+++ b/npm-postinstall.js
@@ -32,7 +32,7 @@ submodules.forEach((submodule) => {
 
   cp.spawnSync(
     'npm',
-    ['i', '--package-lock-only'],
+    ['ci'],
     {
       env: process.env,
       cwd: submodule,
--- a/npm-postinstall.js
+++ b/npm-postinstall.js
@@ -1,0 +1,43 @@
+const fs = require('fs');
+const path = require('path');
+const cp = require('child_process');
+
+const submodules = [
+  'project',
+  'modules/electrophysiology_browser/jsx/react-series-data-viewer',
+];
+
+submodules.forEach((submodule) => {
+  // Check if the submodule is in modules/, and has an override
+  submodulePath = submodule.split(path.sep);
+
+  if (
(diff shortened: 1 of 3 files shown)


npm-postinstall.js
```diff
@@ -0,0 +1,43 @@
+const fs = require('fs');
+const path = require('path');
+const cp = require('child_process');
+
+const submodules = [
+  'project',
```
regisoc, line 6: Extract that lsit into a config file

regisoc:
The postinstall script only involves the last commit right?

npm-postinstall.js
laemtl: @ridz1208 @driusan @regisoc  Do you have any suggestion of where that list of paths should go?

regisoc:
@CamilleBeau this might be interesting to you

regisoc:
@laemtl add in eslint?

laemtl:
> @laemtl add in eslint?

It's done!

npm-postinstall.js
driusan: No. I'm not even sure it needs to be, personally.. when do we have situations where it will need to be updated that isn't in the project directory and isn't a module committed to the repo?

laemtl: Agree, let's keep it there for now.

regisoc (approved):
Looks good to me. 
Screenshot_2022-12-06_18-44-35

PR #8251, opened 2022-12-05
[login] Make it possible to set default page and values via query params

driusan:
This makes it possible to link to the Request Account page by making it possible to set a default mode in the login module via the page=?? query param. It also makes it possible to set the firstname, lastname, or email address by passing them as query params.

This is useful for authentication workflows (such as OpenID Connect) that might want to direct the user to sign up for an account and pre-populate it with known values

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,3 +51,7 @@ check: checkstatic unittests
 
 testdata:
 	php tools/raisinbread_refresh.php
+
+login:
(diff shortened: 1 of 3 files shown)


driusan:
@maltheism I think you wrote the current JSX for the login page, would you mind looking at this?

maltheism (approved):

PR #8252, opened 2022-12-05
[Libraries] Update JWT library

driusan:
We're currently using 3.0, but the library is up to 6.3.1. This updates the LORIS code to use the latest version of our JWT library.


maltheism (dismissed):

maltheism:
Hi @driusan I manually tested login module and the user login was successful. I'm unsure why the test is failing.

driusan:
It's the JWTAuthenticateWithValidToken that's failing. I have a feeling something's upgraded in the library for security..

driusan:
@maltheism the syntax for decode changed slightly. Should be fixed now..

PR #8260, opened 2022-12-07
[Modules] Replace Module->getDataDictionary with Module QueryEngine dictionary.

driusan:
This adds a Module->getQueryEngine() to return the QueryEngine interface for a module. It adds a default NullQueryEngine that implements the interface for modules that have not defined their own interface.

The dictionary module is updated to use the new function call.

Since no modules define the interface yet, testing this probably requires merging it into a branch along with PR #8216 so that there's data to test against in the dictionary. (Assuming you have instruments installed/configured.)


laemtl:
@jeffersoncasimir Do you have time to review?

jeffersoncasimir:
> @jeffersoncasimir Do you have time to review?

I do not have any instruments set up. I think it may be best for someone else to review it ASAP since it is a blocking PR.

xlecours:
There are still a few places that makes or defines the old calls
https://github.com/aces/Loris/blob/main/modules/instruments/php/module.class.inc#L163
https://github.com/aces/Loris/blob/main/modules/dictionary/php/fields.class.inc#L88
... more

driusan:
https://github.com/aces/Loris/blob/main/modules/instruments/php/module.class.inc#L163 isn't calling it, it's defining the old function. It's removed in #8216 where the module is updated to provide a full query engine. Looks like you're right about https://github.com/aces/Loris/blob/main/modules/dictionary/php/fields.class.inc#L88. Not sure why phan didn't catch that. Just fixed it. 

What are the others? The old `getDataDictionary()` implementations will stick around (but not be called) for a reference while the modules are updated to provide the full interface in separate PRs.

driusan:
I deleted the old implementations in the latest commit. The updated instruments query engine will be re-added in PR#8216, the updated candidate_parameters is re-added in PR#8288. The imaging browser one is a simple enough dictionary that it's probably not needed for reference when the module is updated to provide a full query engine.

driusan:
@xlecours ping.. I just rebased this because the required tests changed, but it was passing before the rebase.

xlecours (approved):
Tested using https://github.com/aces/Loris/pull/8288

driusan:
Xavier approved, no take-backs!

PR #8263, opened 2022-12-08
[EEG Browser] Automate viz install process in npm postinstall script

laemtl:
Integrate the [EEG visualization postinstall step](https://github.com/aces/Loris/tree/main/modules/electrophysiology_browser#-installation-requirements-to-use-the-visualization-features) into npm-postinstall.js.
This simplifies the EEG vis setup by eliminating the need to edit the [package.json](modules/electrophysiology_browser/jsx/react-series-data-viewer/package.json:) file.

New steps to enable the EEG visualization: 
- set the new config (GUI/useEEGBrowserVisualizationComponents) to on/off.
- `make dev` or `npm install && npm run compile`
- Tada!

This new config will be used later on the LORIS-MRI side to decide if eeg chunks should be generated or not.

----
As part of this feature, overrides (./project/module) are now automatically compiled in place of the original module (./module) for optimization.

Thanks to @driusan for his suggestion to use a php tool to collect the DB config.


npm-postinstall.js
```diff
@@ -2,23 +2,39 @@ const fs = require('fs');
...
+
+/*
+ * ----------------------------------------
+ * Install npm packages in LORIS submodules
+ * ----------------------------------------
+ */
+const eegVisualization =
+  'modules/electrophysiology_browser/jsx/react-series-data-viewer';
+const submodules = [
+  'project',
```
jeffersoncasimir, line 32: I believe this array is not structured as intended. This is looking for the file 'project/package.json' on line 40.

jeffersoncasimir (changes requested):

npm-postinstall.js
laemtl: What is the issue?

jeffersoncasimir: After discussing on Slack I understand why this is needed.

No change is necessary. This works as intended.

jeffersoncasimir (approved):
Works as intended

SQL/New_patches/2022-12-05-AddVizConfig.sql
```diff
@@ -0,0 +1,27 @@
+-- Adds the option to toggle the EEG Browser visualization components (disabled by default).
```
driusan, line 1: You're missing the update to the schema for new LORIS installs in the 0000-00-xxx files. And raisinbread probably needs to be updated too?

driusan:
I cna't comment on it because there's no line to comment on, but why is there an empty `raisinbread/instruments/NDB_BVL_Instrument_testtest.class.inc` file added?

SQL/New_patches/2022-12-05-AddVizConfig.sql
laemtl: @driusan Done

laemtl:
> I cna't comment on it because there's no line to comment on, but why is there an empty `raisinbread/instruments/NDB_BVL_Instrument_testtest.class.inc` file added?

Oups, solved, thanks!

PR #8265, opened 2022-12-12
[CandidateData.class.inc] syntax error, unexpected ')'

maltheism:
## Brief summary of changes

Fixes PHP parse error:
````
PHP Parse error:  syntax error, unexpected ')', expecting variable (T_VARIABLE) in /Users/alizee/Development/GitHub/McGill/Loris/php/libraries/CandidateData.class.inc on line 48
````


driusan:
You are using an unsupported version of PHP for the main branch. The trailing comma is valid since PHP 8.0.

https://php.watch/versions/8.0/trailing-comma-parameter-use-list

maltheism:
Thanks @driusan I forgot to switch back after working on CCNA.

PR #8268, opened 2022-12-13
[DataQuery] New DQT backend implementation

driusan:
This is intended to make it easier to review/test #8219 or #8216 by implementing the swagger schema, so that it can be tested via the the api docs module.

It's blocked by #8221 and #8260. In order to have anything to query against, you probably also need #8216 (or another QueryEngine implementation for a module.)

Nonetheless, the implementation should be complete (other than the one endpoint described as aspirational in the swagger schema.) and can be reviewed. I may have broken something while trying to extract it from the branch where I was developing the frontend or while trying to appease phpcs/phan.


driusan:
Needs work, not in sync with the latest updates to #8219

xlecours:
I have sent the proposed data to POST /dataquery/queries and got a 200 {"QueryID": 1}
But then, using GET /dataquery/queries nothing shows

driusan:
I believe the sample data in the schema.yml assumes that the candidate_parameter's QueryEngine is implemented (Done in PR#8288, as of now unmerged). It should have returned a 400 error. It looks like it accepted it because of this FIXME needing to be implemented: https://github.com/aces/Loris/pull/8268/files#diff-d21396483bc372130220b07b323e521f63ff1aca56372eb09102c3b4c19e324aR65

modules/dataquery/php/queries.class.inc
```diff
@@ -0,0 +1,356 @@
...
+            case 'POST':
+                $requestjson = (string) $request->getBody();
+                $requestdata = json_decode($requestjson, true);
+                if ($requestdata === false) {
+                    return new \LORIS\Http\Response\JSON\BadRequest(
+                        'Invalid JSON data'
+                    );
+                }
+                // FIXME: Validate query before storing it. Must contain fields.
+                $queryID = $this->_storeQuery($requestjson);
```
xlecours, line 66: This should not succeed if the fields or criteria are invalid.

xlecours (changes requested):
This makes sense. It's hard to test until there is a frontend to it.

driusan:
TODO: After this PR is merged make an issue about instruments module needing to separate viewing data permission from data_entry. (Currently, the `instruments` module does has a `hasPermission('data_entry')` permission check. That means that you need data_entry to query the module data in the dataquery module.)

zaliqarosli (approved):
Looks good! the only issue I had is the one already added by dave as a to-do above ^^

driusan:
TODO: Update patch to insert into modules table and remove drop table statements.

modules/dataquery/php/queries.class.inc
driusan: Should be fixed now

driusan:
@zaliqarosli @xlecours re-factored code as per Xavier's request, can you guys re-review?

SQL/9999-99-99-drop_tables.sql
```diff
@@ -1,3 +1,9 @@
+DROP TABLE IF EXISTS dataquery_run_results;
+DROP TABLE IF EXISTS dataquery_run_queries;
+DROP TABLE IF EXISTS dataquery_shared_queries_rel;
+DROP TABLE IF EXISTS dataquery_pinned_queries_rel;
+DROP TABLE IF EXISTS dataquery_queries;
```
zaliqarosli, line 5: there are 7 tables in the schema file and only 5 here

driusan: So there are.. I had just copied the drop tables that were in the New_patches patch. Should be fixed now.

zaliqarosli (approved):

xlecours (approved):
Let's roll :)

PR #8277, opened 2022-12-15
[tools] Replace a constant date with current date

### LINKED ISSUE #8066: [tools] Hard-coded date in DB_date_zeros_removal.php Patch
**Describe the bug**
Tool runs successfully, but the Patch generated always has the same name with the date "2016-06-01"
`The output file is SQL/Archive/18.0/2016-06-01-update_zero_fields_statements.sql and includes foreign key checks disabling and re-enabling.
`

The output should have a name with the current date.

**To Reproduce**
1. Run tool as intended
2. Observe name of generated patch

shonibare:
fixes #8066


ridz1208:
@driusan this should probably go to 24.1, please confirm

driusan:
I don't think we need to. The script should probably be moved to single use though.

ridz1208:
@shonibare can you make sure to move this script to the single use directory and update all relative paths in the script ?

shonibare:
> @shonibare can you make sure to move this script to the single use directory and update all relative paths in the script ?
Updated

PR #7528, opened 2021-08-11
[instrument_builder] Add REDCap CSV --> .linst functionality

zaliqarosli:
## Brief summary of changes

- [x] Have you updated related documentation?

This PR adds the functionality to load onto the instrument builder a REDCap data dictionary for an instrument in .csv format. Note: you cannot load the entire REDCap data dictionary - you have to extract a single instrument's data dictionary for upload.

The loading functionality will render the instrument under the 'Build' tab as usual i.e. as for linst format, and allow to 'Save' the data dictionary as a .linst instrument. I have also added the capability of downloading rules if they exist in the .csv file.

_**Caveat:**_ This PR is dependent on a new XINValidate feature. I created a "CONTAINS" operator for the XINValidateMultiSelect() function of NDB_BVL_Instrument_LINST.class.inc (on QPN), to work with REDCap rules for multiselect. E.g.:

```
[site] = '3' <--- normal select element
[income_source(8)] = '1' <--- multi select i.e. this field is required if option 8 was selected (among other potential options selected)
```
I can either add the "CONTAINS" feature in this PR, in a separate PR, or take it out completely as it doesn't yet exist in LORIS.

**_Another note_**: I had to play with already messy code, some that were there but never used i.e. `modules/instrument_builder/js/instrument_builder.rules.js ` seems to be doing something but was never implemented on the module. I tried to utilize it for this PR, but maybe there is a better way to code the .rules download. I'm open to all feedback, I coded this all fairly quickly.

#### Testing instructions (if applicable)

1. Load a test .linst file as you would normally. Notice how the instrument is rendered on the 'Build' tab. Make some modifications to the instrument if you like. Save the instrument under the 'Save' tab. Look at the linst file and make sure it is what you expect.
2. Take a REDCap .csv data dictionary for 1 instrument. (If you don't have one, let me know and I can share one with you). Load the .csv file as you would a .linst file. The instrument should be rendered as you would expect in the 'Build' tab. Make sure there are no errors. You can 'Save' the instrument, giving you a .linst file and maybe a .rules file. Read the .linst  file and make sure it is what you expect, following the same format as the linst file generated from step 1.
3. Generate the SQL file for the instrument: `cat $instrument.linst  | php tools/generate_tables_sql_and_testNames.php`
Check the SQL statements were generated fine.
4. Run the SQL script and make sure you get no SQL errors.
5. Add the instrument to the test_battery table.
6. Run `php tools/assign_missing_instruments.php`
[...]

PR #8282, opened 2022-12-16
[instrument_builder] REDCap 2 LINST

zaliqarosli:
## Brief summary of changes

This PR replaces #7528 and mirrors what dave's PR (#8181) tools/redcap2linst.php tool does but just in js and on the browser, with the exception of 2 commits which don't translate to the front-end feature:

- Add --trim-formname option to script
- Add support for using REDCap API instead of CSV file

**Additional feature**: I have added XIN rules functionality for the branching and required fields columns as an extra if its desirable (as a PR to this PR branch) https://github.com/zaliqarosli/Loris/pull/45/

*Note: test on both Firefox and Chrome

1. Load a test .linst file as you would normally. Notice how the instrument is rendered on the 'Build' tab. Make some modifications to the instrument if you like. Save the instrument under the 'Save' tab. Look at the linst file and make sure it is what you expect.
2. Take a REDCap .csv data dictionary (If you don't have one, let me know and I can share one with you). Load the .csv file using the appropriate loader. All the instruments in the data dictionary should download in your browser as multiple .linst files in a zip folder. Make sure the number of instruments match. Read the .linst files and make sure it is what you expect, following the same format as the linst file generated from step 1.
3. Repeat step 1 for one of the downloaded linst files.
4. Generate the SQL file for the instrument: cat $instrument.linst | php tools/generate_tables_sql_and_testNames.php
Check the SQL statements were generated fine.
5. Run the SQL script and make sure you get no SQL errors.
6. Add the instrument to the test_battery table.
7. Run php tools/assign_missing_instruments.php
8. Test the instrument on the front end, and make sure it works like any normal linst instrument


modules/instrument_builder/jsx/redcap2linst.js
```diff
@@ -0,0 +1,92 @@
...
+    //   // remove the element once it has been clicked
+    //   // setTimeout(() => {
+    //   // }, 10);
+    // });
+    // add element to the document so that it can be clicked
+    // this is need to download in firefox
+    document.body.appendChild(element);
+    element.click();
+    document.body.removeChild(element);
+    URL.revokeObjectURL(url);
```
zaliqarosli, line 90: hey @maltheism @laemtl can i pick your brain on this race condition?

this bit of code intends to download in the browser some 80+ linst files. I'm running into the issue where revokeObjectURL isn't being called before the next createObjectURL in the loop [here](https://github.com/aces/Loris/pull/8282/files#diff-3db2a0bc8c5cd4dbd2a87129ddbbd17747ffbb8ceae93dd69b004f02ad6bcda3R162), resulting in the browser running out of memory and only 20+ files getting downloaded (i **think** this is what is happening). the most number of files ive been able to download with this code is 60, but most times only 20 or 10 get downloaded.

do you know how I can make sure that revokeObjectURL is called before the next object url is created? I've tried playing with setTimeout and Promises to no avail :(

maltheism: Hi @zaliqarosli `setTimeout` is an asynchronous function and I believe is what's causing your problem. It's also set in the code to 10 milliseconds and will be very fast.

ridz1208:
blocked by #8181

joshunrau (approved):

joshunrau (approved):

joshunrau:
Oops sorry I did not mean to approve I am not associated with this project so please disregard my review

package.json
```diff
@@ -11,6 +11,7 @@
         "@babel/runtime": "^7.10.5",
         "@fortawesome/fontawesome-free": "^5.11.2",
         "copy-webpack-plugin": "^11.0.0",
+        "csv-parse": "^5.3.3",
```
driusan, line 14: can this use papaparse instead since we already have it as a dependency?

driusan:
I tested this on the dictionary I had (36 instruments, ~1mb) and it worked, but the UX was kind of weird. After you select the file and click "Convert":
1. In Firefox is prompted with 36 different "Save" dialogs
2. In Vivaldi (Chromium-based) it only prompted for saving 1 instrument. No further ones were saved after the first, and the "Convert" button was disabled until reloading the page.

I'm not sure what the right behaviour would be to convert many instruments in a user friendly way. Two potential solutions I can think of are either:
1. Zip the LINST files into 1 file using whatever dependency the dqt is using, so that only one file is saved
2. After converting, rather than trying to download the files automatically, put them in a list of links that are individually clicked to download that particular instrument by the user

zaliqarosli:
@driusan resolved using papaparse and zipping files before download. ready for your review and testing again! thanks

regisoc:
@jeffersoncasimir I heard you built something CBIG, is this related? If so, do you plan on pushing it to Core?

jeffersoncasimir:
> @jeffersoncasimir I heard you built something CBIG, is this related? If so, do you plan on pushing it to Core?

@regisoc Yes it is related and with similar outcome. Here is an [outdated draft PR](https://github.com/aces/Loris/pull/9472). A few changes were since made on the CBIG side that are likely desired for this (generalized) version. If you would like this revived and on core, please bring it up in a weekly meeting so that it gets triaged accordingly.

regisoc:
@jeffersoncasimir TBH I was about to close it. I have no incentive for this to be revived, AFAIK no immediate use for this in HBCD/IBIS, and I do not know if this feature is required or needed more broadly. 
You would be the best suited to say if that is useful on Core. 
Sure I can add a point to the next LORIS meeting to discuss it. Maybe other projects might be interested. 
Closing it for now and will be re-opened or referenced later.

Issue #8171, opened 2022-09-22
[Bug] Incorrect display error message when restricting access to an instrument


paiva:
**Describe the bug**
An incorrect error message is displayed on the frontend when user does not have the permission to access an instrument. 

Error on the log file:

```
[ERROR] You do not have access to this page.#0 /var/www/loris/php/libraries/NDB_BVL_Instrument.class.inc(312): NDB_BVL_Instrument->_hasAccess()\n#1 /var/www/loris/modules/instruments/php/module.class.inc(81): NDB_BVL_Instrument::factory()\n#2 /var/www/loris/src/Middleware/ResponseGenerator.php(50): LORIS\\instruments\\Module->handle()\n#3 /var/www/loris/src/Middleware/AuthMiddleware.php(63): LORIS\\Middleware\\ResponseGenerator->process()\n#4 /var/www/loris/src/Router/ModuleRouter.php(74): LORIS\\Middleware\\AuthMiddleware->process()\n#5 /var/www/loris/src/Middleware/ExceptionHandlingMiddleware.php(55): LORIS\\Router\\ModuleRouter->handle()\n#6 /var/www/loris/src/Router/BaseRouter.php(126): LORIS\\Middleware\\ExceptionHandlingMiddleware->process()\n#7 /var/www/loris/src/Middleware/ResponseGenerator.php(50): LORIS\\Router\\BaseRouter->handle()\n#8 /var/www/loris/src/Middleware/ContentLength.php(52): LORIS\\Middleware\\ResponseGenerator->process()\n#9 /var/www/loris/htdocs/index.php(55): LORIS\\Middleware\\ContentLength->process()\n#10 {main}, referer: http://localhost:8080/instrument_list/?candID=965432&sessionID=8
```
 
**Screenshot**

![Screen Shot 2022-09-21 at 3 53 51 PM](https://user-images.githubusercontent.com/1534575/191637524-0014d768-8a34-4ce0-8bbc-e1f37e8c8848.png)

**What did you expect to happen?**
The message on the frontend should be "You do not have access to this page." as in the first part of the error log, but instead it says "Something went wrong." instead as shown in the screenshot 

**Browser Environment (please complete the following information):**
- Browser Firefox
- Version 104.0.2

**Server Environment (if known):**
_Note: We only support the most recent release of LORIS._
- LORIS Version: 24.0
- Linux distribution and Version: Ubuntu 18.04
[...]

PR #8284, opened 2022-12-19
[core] Fix incorrect message/error for instrument access

regisoc:
## Brief summary of changes

An incorrect error message was displayed on the frontend when a user did not have the permission to access an instrument.
The code was test looping on ALL permissions, but it failed when trying to do so on only ONE permission (the loop went through the instrument name and permission instead of the actual instruments objects). 
This PR fix the error message linked to permission listing. 

1. Add a new user
2. Give the user permissions through the `db` and `config.xml` file except one permission on one instrument (e.g. instrument "bmi", permission "data_entry")
3. Enter the permission selected in 2. in the config.xml file (`instrumentPermissions` section)
4. Go to Loris > Menu > Candidate > Access Profile
5. Choose one PSCID (candidate) then one associated visit (visit label)
6. In the "Behavioural Battery of Instruments", choose one instrument the user should not have access to according to the access rules
7. This should print the 403 page, with "You do not have access to this page." message 

* #8171

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -362,24 +362,18 @@ abstract class NDB_BVL_Instrument extends NDB_Page
         } else { //check if user has instrument's permissions
             //is the instrument listed at all in instrumentPermissions?
             $instrumentListed = false;
-            $instrumentList   = $instrumentPermissions["instrument"];
-            if (array_key_exists("Test_name", $instrumentList)) {
-                // if only one instrument is listed in config.xml file
-                if ($instrumentList["Test_name"] == $this->testName) {
+            // get an array of instruments
+            // if one instrument, get it into an array
+            $instrumentList   = Utility::asArray(
+                $instrumentPermissions["instrument"]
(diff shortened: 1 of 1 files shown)


ridz1208:
@regisoc If I'm understanding correctly the issue is that when only 1 instrument permission is set it is treated like a string instead of an array but when more are set you get an array. If that is the case I would prefer one of the following fixes instead of making 2 different usecases:

1.  fix the code generating the instrumentpermissions array and make sure it always returns an array even if there is a single one
2.  do an `is_array()` check and if it returns false, wrap it in an array and then default to going to the forloop.

Option 1 would be the better option but it MIGHT (should be verified) affect other configurations in which case it might be a risky change for a bugfix.

driusan:
The code generating the array is `$config->getSetting` which does have that behaviour, but can't be changed. The usual way to get that behaviour is to wrap it in a `Utility::asArray` call to ensure you're dealing with an array.

regisoc:
I changed for the wrapper `Utility::asArray`.

ridz1208 (approved):

PR #8285, opened 2022-12-19
[EEG Browser] Visualization performance optimization

laemtl:
Improve the performance of the visualization component.
Changes:
- decreases the memory usage (ChannelReducer/datasetReducer were both keeping a memoized copy of the chunks)
- update the chunks in batch to minimize the number of React rendering.
- remove the cursor props from SeriesRenderer to minimize the number of components to render when mouse is moved.


jeffersoncasimir (approved):
This works as expected and is a huge performance improvement. Great work!

PR #8287, opened 2022-12-19
[core] Remove \Module::factory, update remaining instances to $loris->getModule()

driusan:
This finishes what was started in PR#8221 and eliminates the \Module::factory method. All remaining instances are updated to use $loris->getModule instead, which takes the LorisInstance module paths into consideration.


driusan:
@ridz1208  ping.. I just rebased this because the required tests changed, but it was passing before the rebase.

ridz1208 (approved):

PR #8288, opened 2022-12-19
[candidate_parameters] Add candidate parameters query engine

driusan:
This adds a candidate parameters query engine to deal with candidate meta-data from the candidate_parameters module for the
new data query API.

It also includes a new src/Data/Query/SQLQueryEngine.php abstract class to make it easier to build query engines based on SQL statements for modules where that can be done efficiently with a single statement for any dictionary terms in the data dictionary.

Blocked by #8260


zaliqarosli (approved):
LGTM

Issue #8218, opened 2022-11-07
[Document Repository] Allow for multiple simultaneous file upload


SantiagoTG:
For the 'Document Repository', when uploading files users cannot select multiple files in a single upload. Having to upload one file at a time can become cumbersome and time-consuming, so it would be great to have the option to upload more than one file at a time.

PR #8289, opened 2022-12-19
[Document Repository] Upload multiple files at once

jeffersoncasimir:
### Brief summary of changes

- \<FileElement\> now supports the 'multiple' attribute.
- The document_repository module has been modified to support multiple uploaded files.
- Some visual feedback has been added (error reporting + uploading status).

### Testing instructions

- Ensure that another module using \<FileElement\> still works as intended.
- Try with different files, file types and folder permissions to ensure proper error reporting.

### Related issue: #8218

### DIFF
--- a/modules/document_repository/jsx/uploadForm.js
+++ b/modules/document_repository/jsx/uploadForm.js
@@ -197,7 +197,6 @@ class DocUploadForm extends Component {
         body: formObject,
       })
         .then((resp) => {
-          console.error(resp);
           if (resp.ok) {
             resp.json().then((data) => {
               if (data.error_count === 0) {
@@ -209,6 +208,7 @@ class DocUploadForm extends Component {
                     }
                 });
               } else {
+                console.error(resp);
                 swal.fire('Upload Incomplete', data.message, 'warning');
               }
             }).catch((error) => {
@@ -221,6 +221,7 @@ class DocUploadForm extends Component {
             });
           } else {
             resp.json().then((data) => {
+              console.error(resp);
               swal.fire('Could not upload files', data.error, 'error');
             }).catch((error) => {
               console.error(error);
--- a/modules/document_repository/jsx/uploadForm.js
+++ b/modules/document_repository/jsx/uploadForm.js
@@ -8,7 +8,7 @@ import Loader from 'Loader';
  * Media Upload Form
  *
  * Fetches data from Loris backend and displays a form allowing
- * to upload a document file
+ * to upload document files
  *
  * @author Shen Wang
  * @version 1.0.0
@@ -32,7 +32,7 @@ class DocUploadForm extends Component {
     };
 
     this.setFormData = this.setFormData.bind(this);
-    this.uploadFile = this.uploadFile.bind(this);
+    this.uploadFiles = this.uploadFiles.bind(this);
     this.fetchData = this.fetchData.bind(this);
   }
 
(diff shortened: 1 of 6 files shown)


christinerogers:
tagging Santiago for review of the feature - i think the issue originated from his request?  If it was another project/person requesting, feel free to pass this on

christinerogers:
resolves #8218

christinerogers:
this PR needs review @miladheshmati @SantiagoTG 

@laemtl this is still on the roadmap for 25?

laemtl:
@christinerogers Yes it is still on the roadmap.

christinerogers:
@miladheshmati any more success testing it this week?

miladheshmati:
> @miladheshmati any more success testing it this week?

There was challenges with my VM but thanks to @CamilleBeau, it is solved now and I was able to test it today

miladheshmati:
I was trying to upload two images together but it displays this message:

Screenshot 2023-03-25 at 9 26 34 AM

miladheshmati:
The other issue is that after the previous attempt to upload, the drop-down for the "Site" field seems to have issues. It cannot be changed.

Screenshot 2023-03-25 at 9 30 26 AM

miladheshmati:
Other than that, uploading multiple files works fine.

Screenshot 2023-03-25 at 9 47 28 AM

jeffersoncasimir:
> One small issue is that after the previous attempt to upload, the drop-down for the "Site" field seems to have issues. It cannot be changed.

Thanks for the review! The issue you are experiencing was not introduced by this PR and it is because the 'Site' element is not a <SelectElement> like the fields above and below, but a <SearchableDropdown> element. You will notice that you can also type in this field, to narrow the options, and when you delete the text in the field, all options are available again.

modules/document_repository/jsx/uploadForm.js
```diff
@@ -164,46 +169,80 @@ class DocUploadForm extends Component {
...
-            this.setState({formData: {}});
-            this.props.refreshPage();
+      fetch(this.props.action, {
+        method: 'POST',
+        cache: 'no-cache',
+        credentials: 'same-origin',
+        body: formObject,
+      })
+        .then((resp) => {
+          console.error(resp);
```
driusan, line 200: why is resp always unconditionally logged as an error?

modules/document_repository/php/files.class.inc
```diff
@@ -312,109 +312,136 @@ class Files extends \NDB_Page
...
+                    'Instrument'    => $instrument,
+                    'PSCID'         => $pscid,
+                    'visitLabel'    => $visit,
+                    'File_type'     => $fileType,
+                ]
+            );
 
-        $fileSize = $uploadedFile->getSize();
-        $fileType = pathinfo($filename, PATHINFO_EXTENSION);
+            $uploadNotifier     = new \NDB_Notifier(
```
driusan, line 423: This being in a loop means that the users will get one notification per file instead of one notification per upload action. Is that what we want? @ridz1208

modules/document_repository/jsx/uploadForm.js
jeffersoncasimir: I'm not sure why that was done but it was not introduced by this PR. It seems to be a forgotten debugging snippet.

modules/document_repository/php/files.class.inc
jeffersoncasimir: Very good observation. That was actually not a conscious intention on my part and can make the change if needed.

modules/document_repository/jsx/uploadForm.js
jeffersoncasimir: I agree. Done.

modules/document_repository/php/files.class.inc
jeffersoncasimir: The change has been made

PR #8290, opened 2022-12-20
[Instrument] Remove exception from NDB_BVL_Instrument::factory

driusan:
This removes the exception from NDB_BVL_Instrument::factory(), so that it's possible to call hasAccess on an instrument. Currently the hasAccess or Factory call mix concerns, making it difficult for a module to check if a user has access to an instrument.

### DIFF
--- a/modules/conflict_resolver/php/endpoints/unresolved.class.inc
+++ b/modules/conflict_resolver/php/endpoints/unresolved.class.inc
@@ -253,7 +253,7 @@ class Unresolved extends Endpoint implements ETagCalculator
             true
         );
         if ($instrument->_hasAccess($user) == false) {
-            return new \LORIS\Http\Response\JSON\NotFound(
+            return new \LORIS\Http\Response\JSON\Forbidden(
                 'Permission denied for ' . $conflict['TableName']
             );
         }
--- a/modules/conflict_resolver/php/endpoints/unresolved.class.inc
+++ b/modules/conflict_resolver/php/endpoints/unresolved.class.inc
@@ -252,6 +252,11 @@ class Unresolved extends Endpoint implements ETagCalculator
             '',
             true
         );
+        if ($instrument->_hasAccess($user) == false) {
+            return new \LORIS\Http\Response\JSON\NotFound(
+                'Permission denied for ' . $conflict['TableName']
+            );
+        }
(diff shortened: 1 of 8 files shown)


modules/conflict_resolver/php/endpoints/unresolved.class.inc
```diff
@@ -252,6 +252,11 @@ class Unresolved extends Endpoint implements ETagCalculator
             '',
             true
         );
+        if ($instrument->_hasAccess($user) == false) {
+            return new \LORIS\Http\Response\JSON\NotFound(
```
ridz1208, line 256: Not Found?

modules/instruments/php/visitsummary.class.inc
```diff
@@ -111,6 +113,9 @@ class VisitSummary extends \NDB_Page
                 '',
                 false
             );
+            if ($instrument->_hasAccess($user) == false) {
```
ridz1208, line 116: no `$this->logger->debug()` here ?

modules/conflict_resolver/php/endpoints/unresolved.class.inc
driusan: Fixed

ridz1208 (commented):
one last comment

modules/instruments/php/visitsummary.class.inc
driusan: Added

php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -380,9 +369,7 @@ abstract class NDB_BVL_Instrument extends NDB_Page
                     return true;
                 }
             }
-
-            //no user permissions match required instrument permissions
-            throw new Exception("You do not have access to this page.", 403);
+            return false;
```
ridz1208, line 372: any value in replacing this with `isAccessibleBy` here or out of scope ?

driusan: I don't think NDB_BVL_Instrument currently implements the interface, I think renaming the function to implement it would be out of scope here.

ridz1208 (approved):

PR #8294, opened 2022-12-20
[Core] Replace all $this->lorisinstance with $this->loris

### LINKED ISSUE #8049: Delete duplicate lorisinstance property
NDB_Page has a duplicate `lorisinstance` and `loris` object, both of which store `\LORIS\LorisInstance` objects. We should standardize the usages of `$this->lorisinstance` (which is set by some handle functions) to `$this->loris` (which is set by the page constructor) and delete the duplicate property.

driusan:
This removes the duplicate references to the LorisInstance object in NDB_Page and harmonizes them to all use $this->loris, which is set from the constructor.

Resolves #8049


driusan:
@ridz1208 this might interest you since you were asking about it today

maltheism (approved):

PR #8296, opened 2022-12-20
Fix TimePoint getScreeningStatus return type

### LINKED ISSUE #7282: getScreeningStatus() return type issue
getScreeningStatus() return type does not allow null value, but Screening, in the session table, can be null (a quick look at the session table can convince you). When it is, calling getScreeningStatus() triggers a fatal error:

```
php fix_timepoint_date_problems.php diagnose 587630
Candidate: 587630 
 Timepoint V1 ; SubProjectID: 1 ; Effective DOB: 2016-05-01 ; (SessionID): 1053 
PHP Fatal error:  Uncaught TypeError: Return value of TimePoint::getScreeningStatus() must be of the type string, null returned in /var/www/html/loris-core/php/libraries/TimePoint.class.inc:491
Stack trace:
#0 /var/www/html/loris-core/tools/fix_timepoint_date_problems.php(661): TimePoint->getScreeningStatus()
#1 /var/www/html/loris-core/tools/fix_timepoint_date_problems.php(286): diagnose()
#2 {main}
  thrown in /var/www/html/loris-core/php/libraries/TimePoint.class.inc on line 491
```

The return type should either allow null values, or the function should return an empty string instead of null values.

laemtl:
Updates the return type of getScreeningStatus() to allow null values.

See:
https://github.com/aces/Loris/blob/9860f7f4bba8d70f29936594b8372f45c71b6050/php/libraries/TimePoint.class.inc#L774

https://github.com/aces/Loris/blob/main/SQL/0000-00-00-schema.sql#L194

Fixes #7282

### DIFF
--- a/php/libraries/TimePoint.class.inc
+++ b/php/libraries/TimePoint.class.inc
@@ -500,9 +500,9 @@ class TimePoint implements \LORIS\StudyEntities\AccessibleResource,
     /**
      * Gets this timepoint's screening status.
      *
-     * @return string 'Pass', 'Failure', 'Withdrawal', or 'In Progress'
+     * @return ?string 'Pass', 'Failure', 'Withdrawal', or 'In Progress'
      */
-    function getScreeningStatus(): string
+    function getScreeningStatus(): ?string
     {
         return $this->_timePointInfo["Screening"];
     }


ridz1208 (approved):

laemtl:
@driusan Rebased

PR #8302, opened 2022-12-21
[instruments/instrument_manager] Migrate instrument permissions from config.xml to database

driusan:
This migrates instrument permissions from the config.xml to the database in a new `testnames_permissions_rel` table. The table can be managed from the frontend in the `instrument_manager` module by any user who has the `instrument_manager_write` permission. The script `tools/single_use/migrate_instrument_permissions.php` can be used by existing projects to migrate their permissions from the config.xml to the new table.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,8 +52,12 @@ check: checkstatic unittests
 testdata:
 	php tools/raisinbread_refresh.php
 
+instrument_manager:
+	target=instrument_manager npm run compile
+
 login:
 	target=login npm run compile
 
 mri_violations:
 	target=mri_violations npm run compile
+
--- a/docs/config/config.xml
+++ b/docs/config/config.xml
@@ -73,30 +73,4 @@
 
     </study>
     <!-- end of study definition -->
-
-    <!-- used by _hasAccess in NDB_BVL_Instrument to determine
-         what permissions should be required for each instrument
-         on a configurable basis. -->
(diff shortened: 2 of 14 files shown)


driusan:
Blocked by #8290

Makefile
```diff
@@ -55,3 +55,5 @@ testdata:
 login:
 	target=login npm run compile
 
+instrument_manager:
+	target=instrument_manager npm run compile
```
ridz1208, line 59: is this a leftover or are we adding this for every module now?

driusan: I've been adding it for modules when I work on them because it makes life easier but haven't systematically added every module.

docs/config/config.xml
```diff
@@ -70,37 +70,10 @@
...
             </CertificationInstruments>
         </Certification>
-
     </study>
     <!-- end of study definition -->
 
     <gui>
         <showDatabaseQueries>0</showDatabaseQueries>
     </gui>
-
```
ridz1208, line 80: should probably do the same to the test config ?

driusan: okay, once the PR that's blocking this (#8290) is in and I can rebase one this I'll do that.

driusan:
Note: this will need to be rebased/re-tested after #8422 in case of conflicts with the instrument manager

ridz1208 (approved):

PR #8188, opened 2022-10-18
[package.json] Upgrade react to version 18

maltheism:
## Brief summary of changes

Upgrades react to version 18 and upgrade to latest rendering of react. See the following:
> ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot

#### Testing instructions (if applicable)

1. checkout pr
2. make clean
3. make dev

PR #8309, opened 2023-01-04
[Help] Reactification

laemtl:
Reactification of the menu help widget and the help editor module.

To test: 
- visit a page with help content from a markdown file
- visit a page with help content from the db with context_help and edit
- Test that the help editor support markdown, display the changes in the help box and save the content in the db.


jsx/Help.js
```diff
@@ -0,0 +1,151 @@
...
+ */
+const Help = (props) => {
+  const [isActive, setIsActive] = useState(props.content ? true : false);
+  const [format, setFormat] = useState(null);
+  const [content, setContent] = useState(null);
+  const [isMounted, setIsMounted] = useState(true);
+
+  const getParams = {
+    testName: loris.TestName,
+    ...(loris.Subtest !== '' && {subtest: loris.Subtest}),
```
driusan, line 28: Maybe these should be props?

jsx/Help.js
```diff
@@ -0,0 +1,151 @@
...
+  };
+
+  useEffect(() => {
+    props.content && setContent(props.content);
+  }, [props.content]);
+
+  useEffect(() => {
+    !props.content && fetch(
+      loris.BaseURL
```
driusan, line 37: This one should be a prop.

jsx/Help.js
```diff
@@ -0,0 +1,151 @@
...
+          setFormat(data.format);
+
+          // Render Markdown in wrap div.
+          // If help content is from Markdown helpfile.
+          if (format === 'markdown') {
+            setContent(<RMarkdown content={data.content}/>);
+          } else {
+            // If help content is from DB.
+            setContent(
+              <>
```
driusan, line 58: I don't see why this is a fragment instead of encapsulated in a div

jsx/Help.js
```diff
@@ -0,0 +1,151 @@
...
+    });
+
+    return () => setIsMounted(false);
+  }, []);
+
+  let editBtn = null;
+  // If help content comes from DB `help` table and can
+  // be edited online.
+  if (
+    loris.userHasPermission('context_help')
```
driusan, line 79: should be a prop

jsx/Help.js
```diff
@@ -0,0 +1,151 @@
...
+    >Edit</button>;
+  }
+
+  /**
+   * Renders the React component.
+   *
+   * @return {JSX} - React markup for the component
+   */
+  return (
+    <>
```
driusan, line 109: Should be a div, not a fragment

jsx/Help.js
```diff
@@ -0,0 +1,151 @@
...
+          {content}
+        </div>
+      </div>
+    </>
+  );
+};
+
+document.addEventListener('DOMContentLoaded', () => {
+  const helpContainers = document.getElementsByClassName('help-container');
+  for (let i = 0; i < helpContainers.length; i++) {
```
driusan, line 143: I don't understand the purpose of this loop. Why are there multiple "help containers"? We only have one help icon.

modules/help_editor/jsx/helpEditorForm.js
```diff
@@ -0,0 +1,168 @@
...
+const HelpEditorForm = (props) => {
+  const [title, setTitle] = useState(props.title ?? '');
+  const [content, setContent] = useState(props.content ?? '');
+
+  const returnLabel = 'Return to ' + props.module;
+  const helpContainers = document.getElementsByClassName('help-container');
+
+  useEffect(() => {
+    Array.from(helpContainers).forEach((helpContainer) =>
+      unmountComponentAtNode(helpContainer)
```
driusan, line 27: When I googled for this it says it's being replaced by "root.unmount()". Seems like it's already deprecated.

modules/help_editor/jsx/helpEditorForm.js
```diff
@@ -0,0 +1,168 @@
...
+    );
+  }, []);
+
+  const helpPreview = [];
+  Array.from(helpContainers).forEach((helpContainer) => {
+    helpPreview.push(createPortal(
+      <Help content={
```
driusan, line 34: I don't see why this should be using fragments instead of encapsulated in a div either. Better yet, why not use props.children instead of a content property?

modules/help_editor/php/edit_help_content.class.inc
```diff
@@ -25,124 +27,133 @@ namespace LORIS\help_editor;
...
-        $this->addBasicTextArea(
-            'content',
-            'Content',
-            [
-                'cols' => 140,
-                'rows' => 30,
-            ]
+        return (new \LORIS\Http\Response())->withBody(
+            new \LORIS\Http\StringStream(
+                '<div id="app" ' .
```
driusan, line 140: Why is the help page returning a div with a series of data- attributes? It seems like it would be cleaner to return json.

driusan (changes requested):

htdocs/js/loris-scripts.js
```diff
@@ -0,0 +1,37 @@
+/* eslint new-cap: ["error", {capIsNewExceptions: ["DynamicTable", "FileUpload"]}]*/
+
+$(document).ready(function() {
```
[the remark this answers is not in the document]
laemtl: This snippet of code was in htdocs/js/helpHandler.js (delaeted). I migrated it in another file since it is not doing anything related to the help widget.

jsx/Help.js
laemtl: Because we don't need the extra div in the DOM. Is there any issue in using a fragment?

jsx/Help.js
laemtl: We do have 2. I agree this is not ideal but it is actually design this way. One for the desktop view, one for the mobile view. We only see one at a time.

modules/help_editor/jsx/helpEditorForm.js
laemtl: Yes, I saw that too.
`In React 18, unmountComponentAtNode was replaced by [root.unmount()](https://beta.reactjs.org/reference/react-dom/client/createRoot#root-unmount).`
We are using react 17 so I'm not sure if root.unmount() is already available.

driusan: Maybe it would be best to rebase on #8188 and use it

modules/help_editor/jsx/helpEditorForm.js
laemtl: That's a good suggestion , thanks!

modules/help_editor/jsx/helpEditorForm.js
laemtl: Yes, that's a good plan! I can also replace the other deprecated functions (ReactDOM.render, ...).

jsx/Help.js
driusan: It's a code smell that implies that the component isn't properly encapsulated.. and there's no cost to divs.

laemtl: I'm not sure I agree that using fragments makes the component not properly encapsulated. And there is a cost in using DIV when they are not required since it complexifies the DOM tree.
https://www.geeksforgeeks.org/why-are-fragments-better-than-container-divs/

driusan: Ignoring the assertions about performance with no benchmarks to back them up (and the fact we're talking about a single div here) in that blog post, the only thing left is "because of flexbox and grid" which is exactly wrong. If a component is encapsulated in a single tag, you can use it as an element in a flexbox or grid. If it's not, you're adding multiple elements to the parent which will affect the spacing and alignment of all the other elements in the parent as if you were adding multiple elements, not a single component.

modules/help_editor/ajax/help.php
```diff
@@ -12,9 +12,12 @@
...
  */
 
+$factory  = \NDB_Factory::singleton();
+$user     = $factory->user();
+$editable = $user->hasPermission('context_help');
+
 try {
-    $factory = \NDB_Factory::singleton();
-    $loris   = new \LORIS\LorisInstance(
+    $loris = new \LORIS\LorisInstance(
```
driusan, line 20: I don't think this is necessary anymore. The `$loris` global is now  defined in AjaxHelper.php.

modules/help_editor/php/edit_help_content.class.inc
```diff
@@ -25,124 +27,148 @@ namespace LORIS\help_editor;
...
-        $this->addBasicTextArea(
-            'content',
-            'Content',
-            [
-                'cols' => 140,
-                'rows' => 30,
-            ]
+        return (new \LORIS\Http\Response())->withBody(
+            new \LORIS\Http\StringStream(
+                "<div id='app'>
```
driusan, line 139: If HTML is needed this should be in a smarty template, not a string.

driusan (changes requested):
The LorisInstance already being defined comment isn't a blocker.. but the PHP shouldn't have hardcoded HTML strings.

modules/help_editor/php/edit_help_content.class.inc
driusan: There are no plans to remove it in the near future. It's needed to format email templates.

laemtl:
@driusan Ready for second review.

PR #8310, opened 2023-01-04
main.tpl jQuery cleanup

laemtl:
- remove jQuery instances from main.tpl
- bootstrap jQuery tooltip is replaced by a simple css tooltip (Site affiliations)
- ~~jQuery eslint rules are simplified and turned off for files that still uses jQuery to clean the console output.~~


driusan (changes requested):
The eslint rules shouldn't be changed. The code violating them should be fixed.

driusan:
(in other PRs, eventually.)

laemtl:
> The eslint rules shouldn't be changed. The code violating them should be fixed.

That's the plan, but jQuery rules produce a lot of warnings that prevent devs to notice and fix new warnings they introduce.

ridz1208:
@driusan & @laemtl to discuss removal of warnings

PR #7248, opened 2020-12-18
[survey_accounts] Remove all traces of emails from the database

ridz1208:
## Brief summary of changes

Removing all traces of stored emails in the database without affecting emailing functionality.

PR #8313, opened 2023-01-05
[survey_accounts] Clean up after Email col removed

zaliqarosli:
## Brief summary of changes

The `Email` column of the participant_accounts table was removed in https://github.com/aces/Loris/pull/7248.
This PR cleans up the remaining references to that column, and removes from the schema/raisinbread

1.


driusan:
Seems to be missing an `SQL/Cleanup_patches/` patch (I doubled checked and I don't see one already there either.)

zaliqarosli:
@driusan its been archived https://github.com/aces/Loris/blob/main/SQL/Archive/21.0/Cleanup/2020-12-18-Remove_survey_emails.sql

driusan:
@ridz1208 are we supposed to archive cleanup patches? I thought the point of them were that they're run at your discretion outside of the release cycle.

ridz1208:
@driusan I think we are archiving them so they dont grow into a eternal list, basically to know which patches were added for which release but they are optional yes

zaliqarosli:
@driusan @ridz1208 it is true though that the patch was never added to Cleanup_patches for the releases it was added to, the patch went straight to archive. would it need to be added there?
https://github.com/aces/Loris/tree/v21.0.7/SQL/Cleanup_patches and https://github.com/aces/Loris/tree/v22.0.4/SQL/Cleanup_patches

PR #8315, opened 2023-01-05
Fix skipped unit tests

laemtl:


test/config.xml
```diff
@@ -11,7 +11,7 @@
 <config>
     <!-- set to 1 if development environment -->
     <dev>
-        <sandbox>0</sandbox>
+        <sandbox>1</sandbox>
```
driusan, line 14: Not sure how I feel about this. We want the dangerous tests to run on a GitHub Actions run but we also want them to test as if it's a production environment..

ridz1208: @laemtl to undo this and replace with a SED command to replace the value dynamically within github actions

PR #8323, opened 2023-01-13
[EEG Browser] Fix Download broken links

laemtl:
Fix a few issues with EEGLAB set/fdt files:
- Download link broken
- incorrect placement, it should be logically after the .set file

Download `All files` fails when no annotations exists. 
Migrate the annotation files creation logic from `update` to `updateFiles`, since `update` handles DB updates only.


jeffersoncasimir (approved):
Works great.

PR #8325, opened 2023-01-18
[EEG Browser] Fix EEG montage scale

laemtl:


jeffersoncasimir:
Tested and works. Fixes the scaling of one of my datasets!

jeffersoncasimir (approved):
Tested and works. Fixes the scaling of one of my datasets!

PR #8334, opened 2023-01-23
add feedback to dropdown field when is empty

### LINKED ISSUE #7951: [media] No feedback provided when candidate has no sessions or instruments
**Describe the bug**
In the media module,s uploading functionality, if the user selectsa  PSCID for a candidate with no timepoints, the Visit Label dropdown is empty with no clear explanation or message. Ideally the empty field should read "No visits available for this candidate" or something along those lines, an overkill would be to pop a swal but it might be too intrusive in this case

Similarly when no instruments are available the dropdown should indicate so as well.

**To Reproduce**
1. TEST loris -> media -> updload
2. choose DCC435
3. see dropdowns for Visit Label and instrument

shonibare:
Fixes #7951

### DIFF
--- a/modules/media/jsx/uploadForm.js
+++ b/modules/media/jsx/uploadForm.js
@@ -109,8 +109,10 @@ class MediaUploadForm extends Component {
                         && this.state.formData.visitLabel ?
       this.state.Data.sessionData[this.state.formData.pscid]
         .instruments[this.state.formData.visitLabel] :
-      {};
-    return (
+          {};
+          const emptyVisitMsg = ['No visits available for this candidate'];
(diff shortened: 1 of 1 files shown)


ridz1208 (changes requested):
Did you test what happens when you try to submit the form if this is displayed? my guess is you are going to get all sorts of errors because you are creating an array with one custom option but its a fake option and the server will not recognize it. you are also accidentally bypassing the "required" flag on the field when the dropdown is empty by always adding at least one option

I think you should instead look into leveraging the "placeholder" value of the field which would not be submitted but only displayed when the list is empty

shonibare:
New PR #8402

PR #8339, opened 2023-01-24
[acknowledgements] Remove "Present" column from database and dynamically calculate

### LINKED ISSUE #6231: [acknowledgements] "Present" option in module unclear
There is a column "Present" in the acknowledgement table. It's a boolean. Neither the help text nor test plan describe what it is supposed to mean. It doesn't appear to be used for anything other than displaying "Present: Yes/No" in the table.

It works properly (in that it saves), but it's confusing. Can it just be removed?

driusan:
Since this was triggered by confusion over what "Present" means, this also updates the help text to add a brief description.

Resolves #6231


cmadjar (approved):
The changes look good.

PR #8344, opened 2023-01-30
[DQT] Array and link support in cell data

laemtl:
Screen Shot 2023-01-30 at 1 53 24 PM


xlecours (changes requested):
This works as expected, but only works when display is cross sectional.
 The getRowData function block is split in two depending on display type.

PR #8349, opened 2023-01-31
[configuration] Fix error messages & saving with null Alias

### LINKED ISSUE #7948: [configuration] New Project can be created without alias although alias is NOT NULL
**Describe the bug**
It is possible to create a new project from the frontend without giving it a valid Alias although the SQL field is set to NOT NULL. It seems like the code stores an empty string in the field to bypass the NOT NULL

Additionally, if the project is then edited to add an alias and saved (saving is successful) and subsequently edited to remove the alias (so go back to empty alias) submitting the form DOES NOT generate any errors, but also DOES NOT clear the alias. A simple reload of the page reveals that the alias is still there...

image

**To Reproduce**
1. Go to the Configuration module -> Projects
2. add a new project with no alias (First bug)
3. add an alias by editing the project
4. try to remove alias and submit (second bug)

In both cases an error should be returned forcing the user to enter an alias for the project

CamilleBeau:
## Brief summary of changes
This PR prevents the user from saving a Project configuration with a null Alias in a new project as well as existing projects. 

While testing, I also noticed that several of the error messages were broken, so those were fixed as well

1. Try testing a new project with the following mistakes and make sure that the form does not save, and a relevant error message is given in each case.
- No project name
- No Alias
- An Alias longer than 4 characters
2. Repeat above steps but in an existing Project

* Resolves #7948

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ be found for the image (PR #8155)
 - Add psr/log to composer (PR #8109)
 - Fixed broken DB calls in `assign_missing_instruments` and `instruments` (PR #8162)
 - Add support for PHP 8.1 (PR #7989)
+- Fix Project tab of Configuration module to give correct errors, and prevent saving without Alias (PR #8349)
 ### Modules
 #### API
 - Ability to use PSCID instead of the CandID in the candidates API (PR #8138)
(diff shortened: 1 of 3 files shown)


laemtl:
@CamilleBeau Tests are failing

CamilleBeau:
> @CamilleBeau Tests are failing

@laemtl Fixed!

driusan:
Yes, I think it should go to the release since it doesn't involve schema changes.

CamilleBeau:
@driusan @ridz1208 Rebased!

racostas (approved):
LGTM.

PR #8355, opened 2023-02-02
Create Release Reminders

ridz1208:
Historically we keep forgetting to do at least 1 task at every release. Instead of working on improving our memory I suggest actually checking boxes.

@driusan note how I did not use the word checklist in the title

### DIFF
--- a/release_checklist.md
+++ b/release_checklist.md
@@ -1,0 +1,5 @@
+- [ ] Create SQL Release patch into SQL release directory and move cleanup SQL patches
+- [ ] composer update ?
+- [ ] Update README and install instructions dependencies
+- [ ] Create release notes and upgrade instructions
+- [ ] 
--- a/release_checklist.md
+++ b/release_checklist.md
@@ -2,4 +2,7 @@
 - [ ] composer update ?
 - [ ] Update README and install instructions dependencies
 - [ ] Create release notes and upgrade instructions
-- [ ] 
+- [ ] is there any VERSION documentation that needs to be updated?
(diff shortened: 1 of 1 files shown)


release_checklist.md
```diff
@@ -0,0 +1,5 @@
+- [ ] Create SQL Release patch into SQL release directory and move cleanup SQL patches
+- [ ] composer update ?
+- [ ] Update README and install instructions dependencies
+- [ ] Create release notes and upgrade instructions
+- [ ]
```
cmadjar, line 5: no idea if those should be added to the reminders? Just dropping as I think about them
```suggestion
- [ ] is there any VERSION documentation that needs to be updated?
- [ ] tag all PRs going into the release with proper milestone and review the ones that have "add to release notes"
- [ ] for major and minor releases, create new branches out of the release for the bug fix to go into
- [ ] pull release into main and other appropriate branches if it is bug fix release
```

release_checklist.md
```diff
@@ -0,0 +1,8 @@
+- [ ] Create SQL Release patch into SQL release directory and move cleanup SQL patches
+- [ ] composer update ?
```
driusan, line 2: shouldn't this be done before release testing?

ridz1208: sure we can reorganize this list

driusan:
The description for Projects (classics) says "Projects (classic) 
Projects (classic) are the legacy version of Projects and are created at the repository level. You can create projects for single repository feature work, simple sprint planning, or **release checklists.**"

Maybe this would be good to turn into a project instead of a PR?

PR #8356, opened 2023-02-02
[NPM] Use npm ci in place of npm install to avoid deps conflicts

laemtl:


Makefile
```diff
@@ -21,7 +21,6 @@ clean:
 	rm -f VERSION
 	rm -rf vendor
 	rm -rf node_modules
-	rm package-lock.json
```
driusan, line 24: Comparing this to main this shouldn't be removed but `rm -rf modules/electrophysiology_browser/jsx/react-series-data-viewer/node_modules` should be.

driusan: sorry, misread the diff.. removing this line is correct but `rm -rf modules/electrophysiology_browser/jsx/react-series-data-viewer/node_modules` is missing.

laemtl: Done!

Issue #7876, opened 2021-12-06
[imaging_browser] The Download table as CSV is retrieving more information that's what's shown in the Web Browser.


racostas:
**Describe the bug**
The Download table as CSV is retrieving more information that's what's shown in the Web Browser.

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to `MainMenu->Imaging->Imaging Browser`
2. Click on the `Download table as CSV` button.
4. The excel file downloaded have more (different columns) than the ones shown in the web browser. The last column is shown incorrect info. 
![image](https://user-images.githubusercontent.com/37309344/144879569-34adb96c-14e3-4471-99ee-4e2137f2dd1c.png)

**What did you expect to happen?**
The excel file to reflect the same info as the web browser. 

**Browser Environment (please complete the following information):**
- Microsoft Windows 10
- Google Chrome Version 96.0.4664.45 (Official Build) (64-bit)

**Server Environment (if known):**
- LORIS 24 official testing VM.

laemtl:
Seing additional columns is expected if some are hidden in the data table. @cmadjar will look at the missing header issue.

cmadjar:
Resolved by #8363. Not sure why this did not auto-close.

PR #8363, opened 2023-02-06
[imaging browser] Fix Human dangling out of nowhere

cmadjar:
## Brief summary of changes

This resolves https://github.com/aces/Loris/issues/7876.

When downloading the imaging browser list data as CSV, there are dangling "Human"s with no header associated to it as can be seen below:
image

This PR corrects the issue so that the Human's have a proper header. It also resolves the Human being placed under T1 QC status for example when no scan types are selected to be displayed in the Config module under the Imaging Module tab. See screenshot below for the bug that was introduced in 24:
Screen Shot 2023-02-06 at 5 24 52 PM

As for the content of the CSV showing more columns than what is being displayed, this is a LORIS wide behaviour, not an actual bug.

### DIFF
--- a/modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
+++ b/modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
@@ -189,9 +189,9 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
               s.ID as sessionID,
               GROUP_CONCAT(DISTINCT modality.Scan_type) as sequenceType,
               $PendingNewquery as pending,
-              $modalities_subquery
               s.CenterID as CenterID,
               c.Entity_type as entityType,
+              $modalities_subquery
               s.ProjectID
             FROM psc AS p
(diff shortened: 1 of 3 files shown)


cmadjar:
@driusan @ridz1208 I sent this to 24.1-release as this corrects a bug that was introduced in LORIS 24. The one with "Human" under T1W QC. I noticed that on HBCD last week and the same week @thomasbeaudry updated his LORIS instance to 24 and noticed it as well. So this should probably go in the next bug fix.

modules/imaging_browser/php/imagingbrowserrowprovisioner.class.inc
```diff
@@ -172,6 +172,42 @@ class ImagingBrowserRowProvisioner extends \LORIS\Data\Provisioners\DBRowProvisi
         // =================================================
         // Final query
         // =================================================
+        error_log(print_r(
```
driusan, line 175: seems to be a leftover debug statement.

nicolasbrossard (approved):
Can't see anything wrong with the code. Pulled the branch on my VM and tested: works perfectly. Approving.  👍

PR #8364, opened 2023-02-08
[API] Minor schema fixes

ridz1208:
## Brief summary of changes

Added missing `ProjectID` tag and change descriptors to be less copy pasted

### DIFF
--- a/modules/api/php/models/candidatesrow.class.inc
+++ b/modules/api/php/models/candidatesrow.class.inc
@@ -61,13 +61,13 @@ class CandidatesRow implements \LORIS\Data\DataInstance,
     public function jsonSerialize() : array
     {
         $obj = [
-            'CandID'    => $this->_candid,
-            'Project'   => $this->_projectname,
-            'PSCID'     => $this->_pscid,
-            'Site'      => $this->_sitename,
-            'EDC'       => $this->_edc,
-            'DoB'       => $this->_dob,
-            'Sex'       => $this->_sex,
+            'CandID'  => $this->_candid,
+            'Project' => $this->_projectname,
+            'PSCID'   => $this->_pscid,
+            'Site'    => $this->_sitename,
+            'EDC'     => $this->_edc,
+            'DoB'     => $this->_dob,
+            'Sex'     => $this->_sex,
         ];
 
         return $obj;
--- a/modules/api/php/models/candidatesrow.class.inc
+++ b/modules/api/php/models/candidatesrow.class.inc
@@ -63,7 +63,6 @@ class CandidatesRow implements \LORIS\Data\DataInstance,
         $obj = [
             'CandID'    => $this->_candid,
             'Project'   => $this->_projectname,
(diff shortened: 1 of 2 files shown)


ridz1208:
@driusan I did not make the white space changes, must have been my IDE, let me know if you want me to revert them

modules/api/static/schema.yml
```diff
@@ -129,7 +129,7 @@ paths:
             type: string
       responses:
         '200':
-          description: An object containing the list of instrument summary descriptors
+          description: An object containing the list of projects
```
driusan, line 132: This doesn't make sense to me. why is `/projects/{project}/instruments` returning "a list of projects"?

ridz1208: ma bad, I was distracted

driusan: not sure if you fixed it and lost it in the rebase but it still seems to be wrong.

ridz1208: @driusan this never happened

modules/api/static/schema.yml
```diff
@@ -1831,6 +1831,8 @@ components:
           type: string
         Project:
           type: string
+        ProjectID:
```
driusan, line 1834: This isn't correct. There's supposed to be a project name (line 1832) not a projectid

https://github.com/aces/Loris/blob/main/modules/api/docs/LorisRESTAPI_v0.0.4-dev.md#user-content-30-candidate-api

driusan (changes requested):

modules/api/static/schema.yml
driusan: It looks like it was incorrectly added in https://github.com/aces/Loris/commit/7bd3ddde3079ea667140d2c8c1079d52b73f76fc and missed in review. It should not be in the result that's serialized, only the query for the filter to work. Ihe API isn't intended to expose internal IDs.

ridz1208: I agree but I dont think we can remove it now can we?

should be removed in 0.0.4 I guess

driusan: Why can't we remove it? It's not documented as being there in any API documentation.

ridz1208: I don't know... it was still being returned and someone might be using it? You are the one who usually avoid modifying the API outside of a new release. I'm okai with wtv

modules/api/php/models/candidatesrow.class.inc
```diff
@@ -63,7 +63,6 @@ class CandidatesRow implements \LORIS\Data\DataInstance,
         $obj = [
             'CandID'    => $this->_candid,
             'Project'   => $this->_projectname,
-            'ProjectID' => $this->_projectid,
```
ridz1208, line 66: only removed it from json serialize. I kept the class variable because its used for permission checks

ridz1208:
@driusan made the changes

driusan (approved):

Issue #8025, opened 2022-02-25
[data_release] Test plan and module README need updating


AlexandraLivadas:
Both the test plan and README of this module do not describe or handle the case of files with no version. The test plan should include steps to test adding Unversioned files, filtering by Unversioned files, adding/managing permissions for Unversioned files. The README should mention that files can be uploaded with no version and that when you add a version permission, the user will also be automatically granted permission to see any Unversioned files.

christinerogers:
I've never worked with this module.  the contributors or folks from a project that uses it could do this far more efficiently. let me know if you need help writing it up --

christinerogers:
@ridz1208 which project uses this module?  I don't mind helping someone co-write some of this but i've never used...

ridz1208:
@charliehenrib does CCNA still use this ?

charliehenrib:
@ridz1208 yes, COMPASS-ND does, but not CTU.

ridz1208:
@christinerogers you found your person @charliehenrib

christinerogers:
wonderful @ridz1208  - @charliehenrib  feel free to assign me for review (or co-assign)

PR #8375, opened 2023-02-15
[Data release] - Update Test Plan instruction

charliehenrib:
## Brief summary of changes

Adding testing step for unversioned files. 

#### Link(s) to related issue(s)
#8025

PR #8377, opened 2023-02-15
[Data Release] - Update README information

charliehenrib:
Brief summary of changes

Adding information about unversioned files

#8025

### DIFF
--- a/modules/data_release/README.md
+++ b/modules/data_release/README.md
@@ -21,7 +21,7 @@ be made available as the dataset grows.
 The module is used to manage, upload and download data releases of the study.
 - Different data releases can be differentiated using the version tag.
 - Data coordinators can upload data releases of the study via the front-end
-web interface.
+web interface with or without a version.
 - For each data release, data coordinators can manage permissions of any
   specific user.
 


christinerogers (approved):
this looks ok to me @charliehenrib

driusan (approved):
don't see why this is necessary but it doesn't harm anything either.

driusan (changes requested):
Just looked at the issue (#8025) and I don't think it addresses what it's trying to.

This seems to be trying to address the sentence "The README should mention that files can be uploaded with no version and that when you add a version permission, the user will also be automatically granted permission to see any Unversioned files." but this change doesn't explain the permissions, just changes the wording of a sentence in the scope.

ridz1208:
@driusan I think the other PR is handling the rest https://github.com/aces/Loris/pull/8375

driusan:
@ridz1208 #8375 is handling the test plan. The sentence I quoted from the issue is about the README (which is only touched by this PR, which is why I assume that's what it was trying to address.)

ridz1208:
ok so if I understand correctly, @charliehenrib can you verify and add wording in the ?permissions? section of the readme to say that any user can see unversioned files?

@driusan is the permission section the correct place?

driusan:
Permissions section makes sense to me.

charliehenrib:
@ridz1208 @driusan When I tested the section "when you add a version permission, the user will also be automatically granted permission to see any Unversioned files." on CCNA, it was not the case that unversioned files were visible. @AlexandraLivadas Do you have more information on when the unversioned files were automatically visible or specific projects where this happens?

ridz1208:
@driusan 

@charliehenrib 's comment makes more sense to me. I believe you need to specifically be given access to unversionned documents to actually see them. @charliehenrib maybe you can give it a test on demo to confirm that its not a CCNA override?

charliehenrib:
@ridz1208 I just tested the data release module files permissions and I am getting to the same conclusion. To test, I uploaded a versioned and unversioned file to the data release module using the admin account. I then granted permission to the versioned file for my testing user and only that file was available and not the unversioned files. I then added the permission to unversioned files for that test user and then the unversioned files were available. 

Let me know if you need me to test anything else.

ridz1208:
@driusan could you update your review in light of Charlie's last comment?

driusan:
@ridz1208 @charliehenrib I didn't make the issue in question so maybe @AlexandraLivadas would be better placed than me to answer if this addresses it.

ridz1208 (approved):

ridz1208:
@driusan lets merge this

ridz1208:
the end result of this PR does not add much more information and the changes done in other PRs is sufficient to fix the issue linked above. This PR is closed

PR #8378, opened 2023-02-15
[document_repository] Fix download path to correct user

zaliqarosli:
## Brief summary of changes

1. Checkout branch 24.1-release and run make dev or npm run compile
2. Upload a file as User 1 (i.e. your own). Download that file once uploaded and confirm that the download worked.
3. Login as User 2 (i.e. admin). Try to download the same file and confirm that the download fails.
4. Go to the dashboard and click on one of the items in the 'Document Repository Notifications' widget. Confirm that the download fails.
5. Checkout this PR branch. Download the same file as User 2 and confirm that the download worked
6. Go to the dashboard and click on one of the items in the 'Document Repository Notifications' widget. Confirm that the download worked.

### DIFF
--- a/modules/document_repository/php/files.class.inc
+++ b/modules/document_repository/php/files.class.inc
@@ -71,25 +71,26 @@ class Files extends \NDB_Page
                 ['message' => 'File deleted.']
             ));
         case "GET":
-            $filename = urldecode(basename($request->getUri()->getPath()));
             $route    = explode('/', $request->getUri()->getPath());
             if ($route[2] == 'meta') {
+                $id = urldecode(basename($request->getUri()->getPath()));
                 return (new \LORIS\Http\Response())
                     ->withHeader("Content-Type", "text/plain")
                     ->withBody(
                         new \LORIS\Http\StringStream(
-                            json_encode($this->getUploadDocFields($filename))
+                            json_encode($this->getUploadDocFields($id))
                         )
                     );
             }
 
-            // the file's user directory is the username of the file uploader,
-            // not the user currently requesting the download
-            $uploader = urldecode(basename(dirname($request->getUri()->getPath())));
+            $filename = explode(
+                'Files',
+                urldecode($request->getUri()->getPath())
+            )[1];
 
             $downloader = new \LORIS\FilesDownloadHandler(
                 new \SPLFileInfo(
-                    $downloadpath . $uploader
+                    $downloadpath
                 )
             );
             return $downloader->handle(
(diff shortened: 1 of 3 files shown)


ridz1208:
@zaliqarosli @charliehenrib I'm confused how this is not a much much bigger issue. Shouldn't this prevent pretty much anyone from downloading any file that they didnt upload themselves ??

ridz1208:
@CamilleBeau do you have a VM where you can pull this for charlie to test ?

CamilleBeau:
> @CamilleBeau do you have a VM where you can pull this for charlie to test ?

Yes I will pull this onto our dev VM

zaliqarosli:
@ridz1208 i think it depends on what projects have this bug cz it was fixed for a certain version, and then the code refactored that re-introduced it. i know that it was something I had to fix for COPN on 21.. but it had already been fixed for a later version

charliehenrib:
@CamilleBeau @ridz1208 I followed the testing steps above and it works exactly as described. Is there a passed manual testing label that I should add?

modules/document_repository/php/files.class.inc
```diff
@@ -82,9 +82,14 @@ class Files extends \NDB_Page
                         )
                     );
             }
+
```
driusan, line 85: This whole section of code looks questionable to me. Why are we parsing the username out of the url, parsing the filename out of the url, re-concatenating them, and then passing it to the `FilesDownloadHandler`? It looks to me like the handler should just be `$downloadpath` and the `$filename` should just be `$request->getUri()->getPath()` with the endpoint name prefix stripped out.

I suspect that at some point that this was hacked in because the subdirectories weren't working until fixed in #8392 but the FilesDownloadHandler should handle subdirectories properly now that @cmadjar fixed it.

zaliqarosli: im not sure, john saigle had written that bit of code at the same time he created the FilesDownloadHandler

zaliqarosli: i'm just fixing the username here

zaliqarosli: @driusan fixed it to what you suggested. can you re-review?

zaliqarosli:
@driusan i fixed it to what you suggested. can you re-review?

driusan:
Code looks good to me. @charliehenrib can you re-test it?

charliehenrib:
I retested and it works as it should :)

PR #8388, opened 2023-02-20
Remove unused ajax scripts

driusan:
This removes any ajax scripts that I couldn't find any reference to. `git grep -i setSortedRow`, `git grep -i read_log` and `git grep -i ProcessStateRequestHandler` all return no results.


driusan:
@cmadjar since these are all imaging related modules could you review when you get a chance? There's no rush but it would be good to get it merged before we start release testing for 25.

cmadjar (approved):
LGTM! Thanks @driusan

PR #8389, opened 2023-02-20
Cleanup endpoint hierarchy

driusan:
This makes sure that endpoints always have access to things that are always necessary (such as the LorisInstance object and PSR logger). NDB_Page is updated to extend Endpoint, since pages are types of endpoints.

### DIFF
--- a/modules/api/php/module.class.inc
+++ b/modules/api/php/module.class.inc
@@ -101,13 +101,13 @@ class Module extends \Module
 
         switch ($handlername) {
         case 'candidates':
-            $handler = new \LORIS\api\Endpoints\Candidates();
+            $handler = new \LORIS\api\Endpoints\Candidates($this->loris);
             break;
         case 'login':
-            $handler = new \LORIS\api\Endpoints\Login();
+            $handler = new \LORIS\api\Endpoints\Login($this->loris);
             break;
         case 'projects':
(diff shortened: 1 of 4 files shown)


driusan:
@xlecours When you get a chance can you review? It should still be lightweight but losing access to the logger or LorisInstance context makes it impossible to do things like log or call utility functions.

xlecours:
The loris instance need to be passed in the constructor because it's needed before `handle` is called?

driusan:
It needs to be passed in the constructor to make sure it always exists in any scope that it's used

xlecours:
Except for the dashboard module, nothing is happening between when the constructor of NDB_Page is called and the page `process`  function is called. There is the `loadResources` function but it takes the lorisinstance as a parameter.

I am hesitant in adding parameter to the constructor of all the endpoints but if you have strong opinion in favor of it then let it be.

driusan:
I think it's important for `$this->loris` to always be consistently available on every page/endpoint type or people will get confused when they try and access the LORISInstance context and it works in some contexts but not others and it will result in obscure bugs or losing the context which is required for something that requires it as an argument. We can probably remove the argument from loadResources() now that it's consistently passed in the constructor of page/module objects now.

xlecours (approved):

PR #8392, opened 2023-02-20
Fixes links to BrainBrowser from the MRI violations module

cmadjar:
## Brief summary of changes

The links to BrainBrowser from the MRI violations module were not working for two reasons:
- queried the wrong table name for the protocol violations (`mri_violations_log` should have been used instead of `mri_violation_log`)
- the `FileDownloadHandler` class was using the basename of the file instead of the relative path which resulted in BrainBrowser not finding the file on the file system.

* Those bugs were found when testing https://github.com/aces/Loris/pull/7473 but are present in 24.0.


driusan (approved):
I think the branch is broken but I'll merge it once it's fixed.

PR #8396, opened 2023-02-21
[dqt] Replace jQuery ajax calls with fetch in DQT

driusan:
This replaces all the jQuery ajax calls with fetch in the files modules/dqt/jsx/react.fieldselector.js and modules/dqt/jsx/react.filterBuilder.js.

N.B. this should be tested. I don't have CouchDB setup to test it but the behaviour should theoretically be the same.

### DIFF
--- a/modules/dqt/jsx/react.filterBuilder.js
+++ b/modules/dqt/jsx/react.filterBuilder.js
@@ -115,12 +115,16 @@ class FilterRule extends Component {
     let rule = this.props.rule;
     if (event.target.value) {
       rule.instrument = event.target.value;
-      $.get(loris.BaseURL
-        + '/dqt/ajax/datadictionary.php',
-        {category: rule.instrument}, (data) => {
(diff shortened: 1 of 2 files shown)


modules/dqt/jsx/react.filterBuilder.js
```diff
@@ -115,12 +115,16 @@ class FilterRule extends Component {
     let rule = this.props.rule;
     if (event.target.value) {
       rule.instrument = event.target.value;
-      $.get(loris.BaseURL
-        + '/dqt/ajax/datadictionary.php',
-        {category: rule.instrument}, (data) => {
+      fetch(
+        loris.BaseURL + + '/dqt/ajax/datadictionary.php?category='
```
xlecours, line 119: `+ + ` doesn't work

xlecours (changes requested):

xlecours (approved):

PR #8397, opened 2023-02-21
[Meta] Remove css config setting and rowsPerPage

driusan:
This removes the `css` config setting (which has the ability to point to a different file than main.css but in practice is never edited because everything in LORIS would break without main.css loaded) and rowsPerPage (which is no longer supported because it's not honoured by our React menu filters.)

### DIFF
--- a/php/libraries/NDB_Menu_Filter.class.inc
+++ b/php/libraries/NDB_Menu_Filter.class.inc
@@ -457,7 +457,7 @@ class NDB_Menu_Filter extends NDB_Menu
         }
 
         $config         =& NDB_Config::singleton();
-        $resultsPerPage = $config->getSetting('rowsPerPage');
+        $resultsPerPage = 25;
         $limit          = " LIMIT $resultsPerPage"
                           ." OFFSET " . (($pageNum-1)*$resultsPerPage);
         $result         = $DB->pselect($query . $limit, $qparams);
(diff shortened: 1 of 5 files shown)


php/libraries/NDB_Menu_Filter.class.inc
```diff
@@ -551,14 +550,11 @@ class NDB_Menu_Filter extends NDB_Menu
      */
     function _setDataTable()
     {
-        // create instance of config object
-        $config =& NDB_Config::singleton();
-        // configure the page pagination
-        $rowsPerPage = $config->getSetting('rowsPerPage');
+        $rowsPerPage = 25;
```
KLaFleur, line 553: Is this still being used? Is there a reason to leave $rowsPerPage in a hard coded state?

driusan: The class is still being used because DataFrameworkMenu extends it and it's possible that some projects have modules using it so it can't be completely removed yet, but all our menus have been re-done in react with the DataFrameworkMenu.

ridz1208: @driusan should we sxplicitely deprecate it? at least to see how many places it's used ?

driusan: maybe next release? I don't think there's any rush, it's going to take some work to find out how interwoven DataFrameworkMenu is to NDB_Menu_Filter before we can sever that tie and deprecate it. This PR is just removing the (now unused) rowsPerPage config.

PR #8399, opened 2023-02-21
Remove mri violation module tabs

cmadjar:
## Brief summary of changes

This simplifies the MRI violation module to only one table instead of having two tabs. It also allows users to edit the MRI violation resolution instead of having to ask a developer every time a resolution needs to be changed.

Screen Shot 2023-03-07 at 2 14 58 PM


cmadjar:
@nicolasbrossard could you please review/test this? Thank you!!

SantiagoTG:
**With regards to the setting the 'Resolution status':**

Currently there is no 'Save' button, which means that any selection done is the final selection done for a given image file once the page is closed or refreshed. This is not ideal for two reasons:

1.  It may not be cleat to the user what to do to set the 'Resolution status', leaving the user to figure it out which can lead to naively changing a resolution status thinking nothing will be changed and closing or refreshing the page, leading to erroneous entries.
2. If the user is scrolling around and inadvertently clicks on a dropdown and changes the setting, the moment you close the page or refresh it, you have no way of reverting the change. This is in opposition to being able to click 'Save', where you confirm or commit to any changes made, giving the user time to review and confirm these (it denotes intention to modify the entries). Thus, having the 'Save' button would be more intuitive and safe.

driusan:
Discussed on the LORIS meeting. Will create another issue about the feedback since this is the current behaviour.

PR #8408, opened 2023-02-23
[Core] Check for PSCID uniqueness when generation is user

xlecours:


driusan:
1. Should this go into a bug fix release? It looks like the check was there until #4241 when it looks like it was intentionally removed https://github.com/aces/Loris/pull/4241/files#diff-b1fa8904ac8ed924c216dfbbf3167d0c39cf50edb92a8d3f583256b56dfb8e01L255-L262 .. if it was intentional, we should find out why. If it wasn't, it should be a bug fix.
2. If it's going to main, should we include an ALTER TABLE patch to add a uniqueness constraint too?

xlecours:
PR #4241 seems to have the PSCID uniqueness check to the PSCID Generator and forgot it was also needed/done for user generated PSCIDs. I agree that it could be sent to the bug fix branch to seal the leak asap.
Yes for the alter table in main. But I am not sure if all the projects will be able to apply it.

xlecours:
This will be resent to 24.1

PR #8409, opened 2023-02-24
[EEG] EEG uploader module

laemtl:


laemtl:
@jeffersoncasimir, ready for testing

jeffersoncasimir (approved):
I successfully uploaded a recording to a candidate. Works as intended.

Review notes:
- `The uploaded file exceeds the upload_max_filesize directive in php.ini` was properly reported on the front-end but an error that came before, which was due to the `post_max_size` being too low, reported as a generic failure on the front-end but was adequately reported in the server log.

- After uploading successfully, the only way to upload again or to have an updated view of the browser is by refreshing the page.

- Did you mean to hard-code the notification e-mails [here](https://github.com/aces/Loris/pull/8409/files#diff-92c1d9737cca3e066e5b7c502f660305e440be1a9d7da9c8aeec9b5cc7bf528fR212-R218)?

jeffersoncasimir (changes requested):
Tested the new changes and the only issue remaining is that the new e-mail template was not pushed to the PR.

This actually does not trigger an error on the front-end (no success/fail modal, hangs forever after 100%, unobvious 500 response) nor the server log (nothing unusual reported). The file does still get uploaded.

I have also tested it with success by replacing the e-mail template reference with an existing one.

laemtl:
Oups, thought I did. Solved :)

modules/electrophysiology_uploader/php/upload.class.inc
```diff
@@ -0,0 +1,249 @@
...
+        $uploadedFile = $request->getUploadedFiles()['eegFile'] ?? null;
+        $values       = $request->getParsedBody();
+        $pscid        = $values['pscid'] ?? null;
+        $candID       = $values['candID'] ?? null;
+        $visitLabel   = $values['visit'] ?? null;
+
+        if (empty($request->getUploadedFiles()) && empty($values)) {
+            $upload_max_size = \Utility::getMaxUploadSize();
+            return new \LORIS\Http\Response\JSON\InternalServerError(
+                "The uploaded file size ($upload_max_size) exceeds " .
```
driusan, line 128: This error message is wrong. The variable contains the max upload size, not the size of the uploaded file.

modules/electrophysiology_uploader/php/upload.class.inc
```diff
@@ -0,0 +1,249 @@
...
+        $targetdir = new \SplFileInfo($config->getSetting('EEGUploadIncomingPath'));
+        try {
+            $uploader = (new \LORIS\FilesUploadHandler($targetdir));
+        } catch (\Exception $e) {
+            return new \LORIS\Http\Response\JSON\InternalServerError(
+                $e->getMessage()
+            );
+        }
+
+        // Check for existing files
```
driusan, line 179: Why is it checking for existing files? What is it checking for? What is it doing if it finds them? I'm confused about this part of the code because the FilesUploadHandler already checks if files exist (and either returns an error or overwrites depending on the withOverwrite setting in it.)

driusan (changes requested):
In addition to the couple comments, this seems to be a new module, in which case it's missing the standardized README that every module in LORIS is supposed to have.

modules/electrophysiology_uploader/jsx/UploadViewer.js
```diff
@@ -0,0 +1,94 @@
...
+      show: true,
+    },
+    {
+      label: 'Uploaded By',
+      show: true,
+    },
+  ];
+
+  return (
+    <>
```
driusan, line 80: There's no reason for this fragment to be here.

driusan (changes requested):
This is still missing a standardized README for the module.

modules/electrophysiology_uploader/README.md
```diff
@@ -0,0 +1,51 @@
...
+instructons for [Handling Large File Uploads](../../docs/wiki/00_SERVER_INSTALL_AND_CONFIGURATION/02_Website_Configuration/Handling_Large_File_Uploads.md).
+
+#### <a name="database_config_link"></a> Database Configurations
+
+EEGUploadIncomingPath - This setting determines where on the filesystem the 
+        uploader is to place the uploaded recordings.
+        **Note**: Uploaded recordings are erased from the 
+        `EEGUploadIncomingPath` following a successful archival and insertion 
+        through the LORIS-MRI pipeline. 
+
```
driusan, line 51: Are there any interactions with other modules / LORIS that need to be defined? EEG Browser presumably?

PR #8410, opened 2023-02-24
[Form.js] Add DateTimeElement wrapper

driusan:
This adds a DateTimeElement wrapper to our form classes for use cases where a datetime needs to be selected. It is a wrapper around the datetime-local input type.


jeffersoncasimir (approved):
Works great and as intended!

PR #8412, opened 2023-02-24
[Form.js] Add placeholder option for TextareaElement

driusan:
This adds a placeholder option for the TextareaElement, similar to TextboxElements.


jeffersoncasimir (approved):
Works as intended!

PR #8413, opened 2023-02-24
[Form.js] Add support for step to NumericElement

driusan:
This adds support for the "step" attribute to the NumericElement wrapper. The step is a string type (rather than an integer) so that step="any" is supported for decimals (as per the <input type="number" /> spec.)


jeffersoncasimir (approved):
Works as intended!

PR #8414, opened 2023-02-24
[Form.js] Add support for vertical Radio button groups

driusan:
This adds a vertical=true/false attribute to our <RadioElement> wrapper. The option toggles whether the radio options are displayed vertically or horizontally on the page.

### DIFF
--- a/jsx/Form.js
+++ b/jsx/Form.js
@@ -2549,7 +2549,7 @@ class RadioElement extends React.Component {
 
     const styleRow = {
       display: 'flex',
-      flexDirection: 'row',
+      flexDirection: this.props.vertical == false ? 'row' : 'column',
       flexWrap: 'wrap',
       width: '100%',
     };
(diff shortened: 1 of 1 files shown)


jsx/Form.js
```diff
@@ -2549,7 +2549,7 @@ class RadioElement extends React.Component {
 
     const styleRow = {
       display: 'flex',
-      flexDirection: 'row',
+      flexDirection: this.props.vertical == false ? 'row' : 'column',
```
jeffersoncasimir, line 2552:
```suggestion
      flexDirection: this.props.vertical ? 'column' : 'row',
```
This shouldn't cause any issues but the suggestion or '===' are more conventional in JS. Perhaps I am overlooking a detail which gives reasoning to the original decision.

The feature otherwise works great and as intended!

jeffersoncasimir (approved):
Works as intended!

PR #8416, opened 2023-02-28
Make the number of days since last login before making a user inactive a config

cmadjar:
## Brief summary of changes

This adds a new Config in the config module to allow projects to configure the number of days since a user last logged in before making their account inactive.

#### Testing instructions

1. Run the SQL patch `SQL/New_patches/2023-02-28_create_max_days_inactive_config_for_users.sql`
2. Create a new user if needed and login as that user
3. For that new user, update the last login to 3 years ago `UPDATE user_login_history SET Login_timestamp='2020-02-28 19:28:35' WHERE userID=<new user id>`
4. Try logging in again with that user. You should see the following:
Screen Shot 2023-02-28 at 2 32 22 PM


php/libraries/SinglePointLogin.class.inc
```diff
@@ -356,7 +361,8 @@ class SinglePointLogin
             }
 
             if ($row['Active'] == 'N'
-                || $this->disabledDueToInactivity($username, 365)
+                || ($maxDaysInactive
```
ridz1208, line 364: because of the not-so-predictable behaviour of intval(), I would instead do a more explicit check like `(!empty($maxDaysInactive) && is_numeric($maxDaysInactive))`

intval:
```
Return Values ¶
The integer value of value on success, or 0 on failure. Empty arrays return 0, non-empty arrays return 1.

The maximum value depends on the system. 32 bit systems have a maximum signed integer range of -2147483648 to 2147483647. So for example on such a system, intval('1000000000000') will return 2147483647. The maximum signed integer value for 64 bit systems is 9223372036854775807.

Strings will most likely return 0 although this depends on the leftmost characters of the string. The common rules of integer casting apply
```

php/libraries/SinglePointLogin.class.inc
```diff
@@ -253,6 +253,11 @@ class SinglePointLogin
 
         $this->_lastError = '';
 
+        $factory = NDB_Factory::singleton();
+        $config  = $factory->config();
+
+        $maxDaysInactive = intval($config->getSetting("UserMaximumDaysInactive"));
+
```
ridz1208, line 260: not sure if we want to throw an exception if this value is not set? like do we have HIPAA regulations where an account has to at some point expire?

cmadjar:
@ridz1208 Thanks for the quick review! Your comments should be addressed and I added a note to the roadmap call to discuss whether it is OK to have `maxDaysUserInactive` not set for LORIS.

driusan:
@cmadjar needs rebase, conflicts.

driusan:
@cmadjar seems like there's a problem with the tests on this PR, can you look into it when you get a chance?

cmadjar:
@driusan this passes the tests now.

PR #8426, opened 2023-03-07
[dictionary] Add visits for session scoped variables to API

driusan:
PR#6936 includes a /dictionary/module/(modulename) REST endpoint to act as an API which provides the dictionary for the QueryEngine of a module.

This updates it to include the valid visits for which the data point is valid for session scoped variables.

1. This requires a module with session based variables in its QueryEngine such as #8216 (instruments) or #8288 (candidate parameters). Access `$LORISURL/dictionary/module/$modulename` and look at the JSON returned. Session scoped variables should include a visits key with an array of the correct visits for that variable.

### DIFF
--- a/modules/dictionary/php/moduledict.class.inc
+++ b/modules/dictionary/php/moduledict.class.inc
@@ -90,6 +90,7 @@ class ModuleDict extends \NDB_Page
             }
             $fieldname = $row->getFieldName();
             $datatype  = $row->getDataType();
+            $scope     = $row->getScope();
             $organized[$cat][$fieldname] = [
                 'description' => $row->getDescription(),
                 'scope'       => $row->getScope(),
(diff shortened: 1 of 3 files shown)


modules/dictionary/php/moduledict.class.inc
```diff
@@ -90,6 +90,7 @@ class ModuleDict extends \NDB_Page
             }
             $fieldname = $row->getFieldName();
             $datatype  = $row->getDataType();
+            $scope     = $row->getScope();
             $organized[$cat][$fieldname] = [
                 'description' => $row->getDescription(),
                 'scope'       => $row->getScope(),
```
ridz1208, line 96:
```suggestion
                'scope'       => $scope,
```

ridz1208 (approved):

PR #8427, opened 2023-03-07
[Core] Remove Database::singleton function

driusan:
There are currently 3 methods of getting a reference to a database in LORIS, in order of preference:

1. LorisInstance->getDatabaseConnection()
2. NDB_Factory::singleton()->database();
3. Database::singleton()

The only reference to 3 left is in the implementation of option 2. This removes the Database::singleton by inlineing the reference into the factory and removes all remaining references to Database::singleton (mostly from documentation and a couple unit tests.)

### DIFF
--- a/docs/instruments/NDB_BVL_Instrument_UPLOADER_TEMPLATE.class.inc
+++ b/docs/instruments/NDB_BVL_Instrument_UPLOADER_TEMPLATE.class.inc
@@ -189,7 +189,7 @@ class NDB_BVL_Instrument_TEST_NAME extends NDB_BVL_Instrument
    function importFile(&$file, $args) {
         $fp=fopen($file->fileInfo['tmp_name'], "r");
 
-        $db=& $this->loris->getDatabaseConnection();
+        $db= $this->loris->getDatabaseConnection();
         ///Setting trackchanges to false because getting error messages
         $db->_trackChanges = false;
         ////////////////////////////////////////////////////////////////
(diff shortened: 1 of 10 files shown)


docs/instruments/NDB_BVL_Instrument_UPLOADER_TEMPLATE.class.inc
```diff
@@ -189,7 +189,7 @@ class NDB_BVL_Instrument_TEST_NAME extends NDB_BVL_Instrument
    function importFile(&$file, $args) {
         $fp=fopen($file->fileInfo['tmp_name'], "r");
 
-        $db=& Database::singleton();
+        $db=& $this->loris->getDatabaseConnection();
```
ridz1208, line 192: is the `=&` still necessary ?

php/libraries/Database.class.inc
```diff
@@ -65,70 +66,6 @@ class Database implements LoggerAwareInterface
         // NDB_Factory will set an appropriate ApacheLogger after the fact.
         $this->setLogger(new \Psr\Log\NullLogger);
     }
-    /**
```
ridz1208, line 68: no deprecation message for projects ?

ridz1208 (approved):
looks good other than the couple comments

driusan:
@kongtiaowang it seems like some unit tests can't connect to the database but the log has the credentials and it's consistently the same tests.. do you have any ideas?

driusan:
@ridz1208 getting the tests to pass took more significant changes than expected, can you re-review?

ridz1208 (approved):

PR #8428, opened 2023-03-07
[Build] Fix issues with apt dependencies

laemtl:


ridz1208:
@laemtl we might need the same for 24.1, my PRs are failing

Issue #7880, opened 2021-12-07
[Publication] Can't submit Project proposal


SantiagoTG:
**Describe the bug**
Afete several attempts, the 'Propose project' on the 'Propose a project' tab of the Publication module no longer appears to work. Previously I had attempted submitting several proposals with a 'Attempt to use pselectRow on a query that returns multiple rows' error being displayed as specified on https://github.com/aces/Loris/issues/7877. 

However, now I can no longer submit any proposal and no error message appears. I tried reloading the module and the same. Tried various combinations for the field entries and emails used, but none worked.

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to 'Publications module and try to propose a new project.'

ridz1208:
@SantiagoTG the linked issue was closed because you were not able to reproduce. is this one still valid ?

SantiagoTG:
I am now able to propose a new project on demo.loris.ca (assuming it is on 24.0). However, if I try attaching a file when proposing a new project, an error message will appear and the project will not be created. 

**Error message:**

<img width="554" alt="Screenshot 2023-01-17 at 11 16 29 AM" src="https://user-images.githubusercontent.com/54332302/212958364-9ef765f2-3a8f-46ac-a56a-ceb736a11c43.png">



Creating the project without the attached file, and then going into the newly created publication and adding a new attachment will display a success message, apparently successfully adding the file to the publication. 

<img width="975" alt="Screenshot 2023-01-17 at 11 52 30 AM" src="https://user-images.githubusercontent.com/54332302/212962041-764ba33b-41f0-41bc-88de-26332fe08cf3.png">


However, no file is actually attached and the entered data for the attachment is not available when opening the publication.

ridz1208:
@SantiagoTG the first part is my bad. I just fixed it please try again. the second part might also be fixed however the fact that it doesnt report an error is worrying. might still be an issue on it's own

SantiagoTG:
I am now able to propose a new project including an attachment, which is successfully reported and include the attachment when opening the publication entry. 

However, if on the other hand I create a new publication, and after creating the publication I add a new attachment, it reports the edit was successful, but no attachment will be added to the publication.

skarya22:
The last comment is still an issue -- the original issue is not an issue though

PR #8433, opened 2023-03-08
[media/publication] Remove FileDownload ajax scripts

driusan:
This removes the FileDownload.php ajax scripts from both the media and publication module.

A new `LORIS\Http\FilesPassthroughEndpoint` abstract class is added to handle shared code between files download endpoints so that the modules only need to define module specific things such as the directory or permissions required. 

The FilesPassthroughEndpoint should be updated to also handle POST/PUT/DELETE (and replace the corresponding ajax scripts) but this first pass only replaces the FileDownload.php scripts with a real endpoint that uses the FilesDownloadHandler.

Note: Uploading in the publication module seems to be broken since #7880 and RB doesn't have any uploads by default, so I wasn't able to test the download, but since it's the same code as the media module I'm pretty confident in it.


driusan:
@kongtiaowang This has been rebased and the tests fixed, can you have another look?

kongtiaowang (approved):
LGTM

PR #8437, opened 2023-03-10
Make getModule register autoloader for module

driusan:
PR#8287 refactored the code in a way which removed the side-effect of registering an autoloader when calling Module::factory. This was probably not a good idea. Calls to methods on a Module instance retrieved now crash when attempting to call a method on the module if the module uses a class internal to the module.

Since the caller doesn't know if the module is going to use internal namespace classes, this re-adds the implicit autoloader registration when getting a module to the LorisInstance object.

It also caches the module descriptor in the LorisInstance after it's retrieved in order prevent registering the same autoloader multiple times.

### DIFF
--- a/src/LorisInstance.php
+++ b/src/LorisInstance.php
@@ -117,6 +117,7 @@ class LorisInstance
         return false;
     }
 
+    private array $moduleInstances;
     /**
      * Get the \Module class for the module named $name,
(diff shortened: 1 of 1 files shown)


xlecours (approved):
This is straight forward. Do you want me to test it?

driusan:
It fixes it for me if you're happy with the amount of testing.

PR #8442, opened 2023-03-14
[mri_violations] Provide feedback on resolution status change

### LINKED ISSUE #8441: MRI violation drop down saved with no feedback
**Describe the bug**

Before the reactification of the MRI violation module, there used to be a save button that the user would click on to save the resolution status. Right now, the user has no feedback that the data will be saved from the dropdown selection.

Discussed either adding a save button or some kind of feedback to the user and updates to the Resolution Status column with the saved results.

driusan:
There is currently no feedback to the user when they change the resolution status of a violation that something has changed, which can be confusing.

This fades the table cell to green after the cell has been changed in the same way that the conflict resolver does, so that the user is aware that something happened.

Green is probably not the ideal colour, but is consistent with the conflict resolver.

Fixes #8441


cmadjar (approved):
LGTM 👍

PR #8446, opened 2023-03-15
[user_account] fix when one user edit other user's permission that requires password change.

kongtiaowang:
## Brief summary of changes

[user_account] fix when one user edit other user's permission that requires password change.
截屏2023-03-15 下午1 08 23

When one user edit other user's permission that no need to password change.

### DIFF
--- a/modules/user_accounts/php/edit_user.class.inc
+++ b/modules/user_accounts/php/edit_user.class.inc
@@ -1200,6 +1200,7 @@ class Edit_User extends \NDB_Form
         if (isset($values['NA_Password'])
             && $values['NA_Password'] == "on"
             && $values['SendEmail'] != "on"
+            && empty($this->identifier)
         ) {
             $errors['Email']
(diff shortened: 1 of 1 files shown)


modules/user_accounts/php/edit_user.class.inc
```diff
@@ -1200,6 +1200,7 @@ class Edit_User extends \NDB_Form
         if (isset($values['NA_Password'])
             && $values['NA_Password'] == "on"
             && $values['SendEmail'] != "on"
+            && empty($this->identifier)
```
ridz1208, line 1203: I dont think this is a good idea. even if the user has been created without a password (i.e. from request account) you have to validate that they are emailed when a password is generated

kongtiaowang: Good catch! this is not a bug, the user pass is null. I will close this PR.

PR #8451, opened 2023-03-16
[Issue tracker] change userID to full name

regisoc:
## Brief summary of changes

Change user IDs to user full names throughout the main pages.

1. Go to the issue tracker.
2. Check the `assignee` and `reporter` columns in the main table (there should be user full names).
3. Click on one issue to print the details.
4. All information should contain the user full names (i.e top of the page, table, and history section).

### DIFF
--- a/modules/issue_tracker/php/edit.class.inc
+++ b/modules/issue_tracker/php/edit.class.inc
@@ -104,11 +104,11 @@ class Edit extends \NDB_Page implements ETagCalculator
         }
 
         foreach ($assignee_expanded as $a_row) {
-            $assignees[$a_row['UserID']] = $a_row['Real_name'];
+            $assignees[$a_row['UserID']] = $this->formatUserInformation($a_row['UserID']);
         }
 
         foreach ($inactive_users_expanded as $u_row) {
-            $inactive_users[$u_row['UserID']] = $u_row['Real_name'];
+            $inactive_users[$u_row['UserID']] = $this->formatUserInformation($u_row['UserID']);
         }
 
         $otherWatchers = [];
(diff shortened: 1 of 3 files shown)


kongtiaowang:
@regisoc in this file issue_tracker/test/Issue_TrackerTest.php  add       [  **'assignee' => 'TestUser',** ]   after line 63, it will pass test.

racostas:
Hi @regisoc , let's add the changes to the CHANGELOG file for keeping track https://github.com/aces/Loris/blob/main/CHANGELOG.md

modules/issue_tracker/php/edit.class.inc
```diff
@@ -429,7 +452,7 @@ class Edit extends \NDB_Page implements ETagCalculator
 
         $issueID = intval($values['issueID']);
 
-        $issueValues['lastUpdatedBy'] = $user->getUsername();
+        $issueValues['lastUpdatedBy'] = $user->getUserName();
```
racostas, line 455: since it is assigned in line 483 I think this input in 455 is not longer necessary .

racostas (changes requested):
One single comment. Aside of that looks good.

modules/issue_tracker/php/edit.class.inc
regisoc: It will be because it is used in line 473 (update on existing issue case) and 477 (new issue case).
This needed to be the username (userID) for the `issues` table. The value is just changed for the full name after database insert/update (line 483).

racostas: I see, yes, it's a bit confusing the way the code is wrote but I think this precedes this PR.  For example **$historyValues** is assigned in 470 the used in 486, passing the DB updated in between without using the value :) 

The PR is fine !

racostas (approved):
LGTM.

Issue #8026, opened 2022-02-25
[examiner] Column for 'Instrument' in certification menu appears narrow


SantiagoTG:
**Describe the bug**
In the 'Certification' page for a given examiner, the 'Instrument' column is very narrow and bundles the display of the instrument name, making the display of the information not optimal. 

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to the examiner module
2. Select an examiner to access the 'Certification' page
3. Text in 'Instrument' column is bundled as column width is too narrow.

**What did you expect to happen?**
An adequate column width would improve display of the table.

**Browser Environment (please complete the following information):**
- Chrome

![Screen Shot 2022-02-25 at 2 39 00 PM](https://user-images.githubusercontent.com/54332302/155785715-09d9eaf2-a43c-4146-9a11-2b664a3f4b26.png)

miladheshmati:
This can be fixed by adding this class to `bootstrap.min.css` or `main.css` which are two CSS files used on that page.  ( I didn’t know where to add it and I couldn't find any of those CSS files in the repository).

```css
    @media (min-width: 992px) {
        form#edit_examiner .col-md-1 {
            width: 22%;
        }
    }
```

@ridz1208, I will be thankful if you can guide me with that.

ridz1208:
@miladheshmati that is funny because I just might be the WORST person to ask for help when it comes to CSS.

That being said. Here is where I would start.
 1. I would look first at what class is currently dictating the column width and where it is defined
 2. I would see if that same class is being used for other forms, if yes, then modifying it might have undesired side effects, if not then you can just edit that to get it to work like  you want it
 3. if you need to create custom more specific css code, the bestplace to put it is in the module itself under a `css` directory ([see example](https://github.com/aces/Loris/tree/v24.0.0/modules/dqt)) and make sure to add the necessary getCSSDependencies to import the file when the module is loaded ([example](https://github.com/aces/Loris/blob/v24.0.0/modules/dqt/php/dqt.class.inc#L92))


Anything beyond that, I might do more harm than good being involved with. @laemtl @zaliqarosli @CamilleBeau anyone of them might be more help than me

laemtl:
@miladheshmati  it's better to not modify `bootstrap.min.css ` directly, since it is not maintained/authored by our team. The file `htdocs/bootstrap/css/custom-css.css` is where LORIS custom css rules are defined. 
[...]

PR #8453, opened 2023-03-17
[examiner] Column for 'Instrument' in certification menu appears narrow

miladheshmati:
## Brief summary of changes

This PR is changing the CSS class used in the examiner page. The former class had a small percentage of the width causing the first field become narrow, by changing it to another class, the problem is solved.

1. Go to the examiner module
2. Select an examiner to access the 'Certification' page
3. Text in the 'Instrument' column should not be narrow anymore.

* Resolves #  #8026

### DIFF
--- a/modules/examiner/templates/form_editExaminer.tpl
+++ b/modules/examiner/templates/form_editExaminer.tpl
@@ -12,7 +12,7 @@
                     </div>
                     {/foreach}
                     <div class="row hidden-xs hidden-sm">
-                        <div class="col-md-1">
+                        <div class="col-md-3">
                             <label>Instrument</label>
                         </div>
                         <div class="col-md-2">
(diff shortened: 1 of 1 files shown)


racostas (changes requested):
Doesn't seem to be working for me.

This is before the change:
image

This is after the change:
image

miladheshmati:
@racostas, ready to be retested.

racostas (changes requested):
nice!
Just one minor thing. Let's try to give a bit of space between the textboxes and the buttons. 
image

racostas (changes requested):
This proposal will be introducing a **"line"** between the textboxes and the buttons. While I'm not strongly  in favor or against it does differs from the original design.  

What do you thing about it @miladheshmati ? Could you please also consult it with Rida and Dave, maybe Christine.  

Thanks. 

image

PR #8458, opened 2023-03-17
Fix PropsTypes for children

laemtl:
- Proptypes.element does not allow multiple children
- Proptypes.array/Proptypes.object is too vague

Proptypes.node is a compound type for any renderable object: nothing, string, single element, several elements, fragment, component.


driusan:
I don't think the new file modules/help_editor/js/helpEditorForm.js with +14000 lines is supposed to be in this?

PR #8466, opened 2023-03-20
[eslint] enable no-fallthrough rule

driusan:
Forgetting a "break" in a switch statement is a common developer error. For instance, most of the places caught by this rule which is currently disabled.


regisoc (approved):
Running `$ npm run lint:javascript` without listed breaks raise errors. 

Screenshots with a break removed in `jsx/batteryManagerIndex.js`: 
<details>
  <summary>Click me</summary>

eslint_break_error


</details>

LGTM

PR #8471, opened 2023-03-21
[configuration] Reactify module

driusan:
This rewritse the main page of the configuration module--which is the cause of the majority of our javascript warnings--with React. The configuration module was written with jQuery/smarty before we began using React.

The reactified version should work the same as the old with a few exceptions.
1. Our standard React element types are used instead of custom ones. This means, unfortunately, that the tooltip description is gone because our standard elements do not support it.
2. The tabs are fetched dynamically instead of all upfront.
3. swals are used to notify the user about setting update success/failure.

Other than that, the module should be unchanged from the end-user perspective.

### DIFF
--- a/modules/configuration/test/ConfigurationTest.php
+++ b/modules/configuration/test/ConfigurationTest.php
@@ -61,6 +61,8 @@ class ConfigurationTest extends LorisIntegrationTest
     public function testConfigurationPageLoads()
     {
         $this->safeGet($this->url . "/configuration/");
+        // wait for the React to render
+        $this->safeFindElement(WebDriverBy::cssSelector("#lorisworkspace div"));
         $bodyText = $this->safeFindElement(
             WebDriverBy::cssSelector("body")
         )->getText();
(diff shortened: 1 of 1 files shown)


driusan:
@kongtiaowang rebased

modules/configuration/test/ConfigurationTest.php
```diff
@@ -155,6 +157,8 @@ public function testAllLinks()
     {
 
         $this->safeGet($this->url . "/configuration/");
+        // wait for the React to render
```
kongtiaowang, line 160: Please replace this function in this file. using safeFindelement replace findElements 
    private function _linkTest($text)
    {
        $this->safeClick(WebDriverBy::linkText($text));
	$webActives =  $this->safeFindElement(WebDriverBy::cssSelector(".active"));
        $bodyText   = $webActives->getText();
        $this->assertStringContainsString($text, $bodyText);
    }

kongtiaowang: Regarding the testplan3 in the dashboard test, you can temporarily comment it out. I've identified a solution to address this issue, and I'll implement it later. https://github.com/kongtiaowang/Loris/pull/381/files#diff-2af5e853d357c98c23e22e3a7d855bc560485103cb82026fa408d7fdf6f3fa5d

PR #8242, opened 2022-11-25
[EEG] Multiple coord systems

regisoc:
## Brief summary of changes

Some BIDS files contain multiple coordinate systems. Currently, only one coordinate system is considered, others are ignored. This PR allows keeping multiple coordinate systems description (CapTrak, CTF, ...) per BIDS files.

Linked to:

- [Loris-MRI 885](https://github.com/aces/Loris-MRI/pull/885)

PR #8478, opened 2023-03-23
[EEG] Electrodes query and file download fix

laemtl:
PR #8242 modified the electrodes DB schemas and since then, the EEG Browser is broken. 
This fixes the issue by rewriting the query logic. 
Also fixes a few issues with the download buttons and the layout.

To test:
----------- 
- Make sure the EEG Browser loads
- Create a new annotation
- Check that the comment section of the new comment is displayed without any layout issues
- Make sure the annotations, electrodes and all files are downloadable.

### DIFF
--- a/modules/electrophysiology_browser/css/electrophysiology_browser.css
+++ b/modules/electrophysiology_browser/css/electrophysiology_browser.css
@@ -45,7 +45,7 @@ svg {
 .list-group-item {
   position: relative;
   display: flex;
-  flex-direction: row;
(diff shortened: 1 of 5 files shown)


laemtl:
> Hi thanks for this PR @laemtl I cannot interact with the Electrode download button right now.
> 
> image

This is expected when no electrode file is found. Can tou check if this is the case?

regisoc (approved):
LGTM

PR #8483, opened 2023-03-27
[Webpack] build/make fix

laemtl:
A change required in webpack to prevent generation of LICENSE.txt files was incorrectly added at the wrong place, breaking compilation for the production mode (npm run build or make). This fixes the issue.


ridz1208:
included in merge PR

PR #8485, opened 2023-03-28
[battery_manager] use visit controller to fetch visit list

### LINKED ISSUE #8459: [battery manager] Get list of visits from visits table not session
**Describe the bug**
The battery manager currently only has access to visit labels already used in at least one session. it should have access to visits that have never been used

ridz1208:
## Brief summary of changes
The battery manager currently uses the `Utility::getVisitList()` function to list the visits available. That utility function queries the session table for existing visitlabels used in at least one session. That is problematic when you are creating a new visit and new instruments and want to assign instruments to the new visit, it simply does not show up in the drop down.

This PR fixes that problem and removes the use of Utility::getInstrumentList to get an accurate list of full names for instruments and get rid of deprecation warnings 

* Resolves #8459

### DIFF
--- a/modules/battery_manager/php/testoptionsendpoint.class.inc
+++ b/modules/battery_manager/php/testoptionsendpoint.class.inc
@@ -63,7 +63,9 @@ class TestOptionsEndpoint extends \NDB_Page
             $this->loris->getDatabaseConnection()
         );
         return [
-            'instruments' => \Utility::getAllInstruments(),
+            'instruments' => \NDB_BVL_Instrument::getInstrumentNamesList(
+                $this->loris
+            ),
             'stages'      => $this->_getStageList(),
             'subprojects' => \Utility::getSubprojectList(null),
             'visits'      => $visitController->getAllVisits(),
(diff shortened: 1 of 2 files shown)


KLaFleur:
When I try this out, the dropdowns which should show the visits (V1,V2,V3, etc) just return the default js object serialization `[object Object]` a screenshot is attached below. It's possible this is due to a database issue on my end, but I'm not sure what that'd be.

Screen Shot 2023-04-27 at 11 27 56 PM

ridz1208:
@KLaFleur this is interesting. The list I was modifying is the one under the "add test" modal window. can you test if that one is working and if not I can take a look at whats happening

KLaFleur:
I was getting the same behavior on the ‘add test’ window as well — I can
check again to confirm that but I’m 99% sure that was the case.

ridz1208:
@KLaFleur could you test again please

KLaFleur:
Yes just tested (in both the general filter and in the 'new test' window,
the dropdown works and the list looks complete!

ridz1208:
@KLaFleur take a look at the code and try to spot any issues. @driusan will do a secondary review after. when you take a look at the code, if all is well make sure to "Approve" the PR

KLaFleur (approved):

Issue #7882, opened 2021-12-07
[Issue Tracker] Deleting attachment leaves no record of the attachment's existance


SantiagoTG:
**Is your feature request related to an existing module?**
For tickets with an attachment previously uploaded, if the user deletes the attachment, it would be useful to keep a record of the deleted file.

**Describe the solution you'd like**
When an attachment is deleted, a log on the 'Attachment History' should appear. Currently once a file is deleted, the user has no way of knowing if an attachment had been previously loaded or not.

ridz1208:
@GeorgeMurad  The history should reflect that an attachment existed at some point and was removed but not archive the file in any way nor reference the file name!

Text: `Attachment has been deleted` and make sure the timestamp is there

PR #8490, opened 2023-04-03
[Issue Tracker] deleting attachment leaves no record of the attachment history

GeorgeMurad:
- Adding a new feature to the `Issue Tracker` module for the `Attachment History`, when deleting the log on the 'Attachment History' should appear.
- Adding a new line called `Date of deletion` to the `Attachment History` when it is deleted. It will show the user when they deleted the uploaded file (Attachment).

#### Testing instructions (if applicable):
1. Create a new issue under the issue tracker.
2. Add attachment.
3. Upload attachment.
4. Delete the attachment when it appears.
5. The `Attachment History` should appear.

### Note: The Attachment History Section needs to be translated. 

Related To: https://github.com/aces/Loris/issues/7882

### DIFF
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -1701,7 +1701,7 @@ CREATE TABLE `issues_attachments` (
     `description` text DEFAULT NULL,
     `file_size` int(20) DEFAULT NULL,
     `mime_type` varchar(255) NOT NULL DEFAULT '',
-    `date_deleted` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    `date_deleted` timestamp NULL DEFAULT NULL,
     CONSTRAINT `fk_issues_attachments_issue` FOREIGN KEY (`issueID`) REFERENCES `issues` (`issueID`),
     PRIMARY KEY (`ID`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -1701,7 +1701,7 @@ CREATE TABLE `issues_attachments` (
     `description` text DEFAULT NULL,
     `file_size` int(20) DEFAULT NULL,
     `mime_type` varchar(255) NOT NULL DEFAULT '',
-    `date_deleted` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP;
+    `date_deleted` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
     CONSTRAINT `fk_issues_attachments_issue` FOREIGN KEY (`issueID`) REFERENCES `issues` (`issueID`),
     PRIMARY KEY (`ID`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -1697,7 +1697,6 @@ CREATE TABLE `issues_attachments` (
     `file_hash` varchar(64) NOT NULL,
     `date_added` timestamp NOT NULL DEFAULT current_timestamp(),
     `file_name` varchar(255) NOT NULL DEFAULT '',
-    `deleted` tinyint(1) NOT NULL DEFAULT 0,
     `user` varchar(255) NOT NULL DEFAULT '',
     `description` text DEFAULT NULL,
     `file_size` int(20) DEFAULT NULL,
--- a/SQL/0000-00-00-schema.sql
+++ b/SQL/0000-00-00-schema.sql
@@ -1702,6 +1702,7 @@ CREATE TABLE `issues_attachments` (
     `description` text DEFAULT NULL,
     `file_size` int(20) DEFAULT NULL,
     `mime_type` varchar(255) NOT NULL DEFAULT '',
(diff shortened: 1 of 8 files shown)


SQL/0000-00-00-schema.sql
```diff
@@ -1577,6 +1577,7 @@ CREATE TABLE `issues_attachments` (
     `description` text DEFAULT NULL,
     `file_size` int(20) DEFAULT NULL,
     `mime_type` varchar(255) NOT NULL DEFAULT '',
+    `date_deleted` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP;
```
laemtl, line 1580: The semi-colon here is syntactically incorrect.

modules/issue_tracker/jsx/attachments/attachmentsList.js
```diff
@@ -191,9 +191,48 @@ class AttachmentsList extends Component {
         const deleteData = JSON.stringify(item);
         // Hide "soft" deleted attachments
         if (parseInt(item.deleted) === 1) {
-          continue;
-        }
-        attachmentsRows.unshift(
+            attachmentsRows.unshift(
```
laemtl, line 194: Indention is incorrect.

modules/issue_tracker/php/attachment.class.inc
```diff
@@ -148,7 +148,7 @@ class Attachment extends \NDB_Page implements ETagCalculator
             $DB      = $factory->database();
             $DB->update(
                 'issues_attachments',
-                ['deleted' => 1],
+                ['deleted' => 1, 'date_deleted' => date('Y-m-d H:i:s')],
```
laemtl, line 151: I wonder is deleted is still necessary? Having date_deleted <> NULL is sufficient to tell if an attachment is deleted.

GeorgeMurad: @laemtl It is necessary otherwise will through an error

GeorgeMurad: @laemtl Yes we still need  the `date_deleted` column to display the `date of delation`.

GeorgeMurad: @laemtl Yes we do still need `deleted` column in the database table.

laemtl: This is used only once:

`if (parseInt(item.deleted) === 1) {`

which can become

`if (item.date_deleted) {`

laemtl:
@GeorgeMurad Following our roadmap meeting discussion today, can you automatically create a comment when a file is delated instead? That will avoid changing the SQL schema  which  is a better option.

ridz1208:
@driusan is this  a critical bugfix? its going to 24

CamilleBeau:
@GeorgeMurad Can you please resolve conflicts in this one?

SQL/New_patches/2023-04-03_Add_Column_to_issues_attachments_table.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE issues_attachments ADD COLUMN `date_deleted` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP;
```
CamilleBeau, line 1: Should this be "NOT NULL"? Since it's called "date_deleted" It seems this value should be null if the attachment has not been deleted.

modules/issue_tracker/php/attachment.class.inc
CamilleBeau: I would second Laetitia's thought here!

CamilleBeau (changes requested):

modules/issue_tracker/php/attachment.class.inc
driusan: I agree.

racostas (changes requested):
Only two small comments:

- I think we usually don't send the patch to `SQL/Release_patches` , I think to remember this will be done when creating the release script for updating LORIS. @driusan could tell more about this than myself.
-  When I was testing this PR in main I got a 500 error that was not caused by the PR but that maybe we should solve as part of it. The issue was a type mismatch for the following variables in the data model `attachmentdto.class.inc`:
    -  `$ID, $issueID and $file_size` have to be declared **int** in place of string.
    image

GeorgeMurad:
@racostas I agree with updating the variable types to INT. And how about the `SQL/Release_patches`?

racostas:
> @racostas I agree with updating the variable types to INT. And how about the `SQL/Release_patches`?

I think the Patch inside `SQL/Release_patches` is not necessary because this folder only contains "release patches" that are made from the ones that were already included in `SQL/New_patches`/ and `SQL/Cleanup_patches`, @driusan can you confirm this ?? I'm not 100% sure. Thanks.

driusan:
Yes. Patches go into New_patches. @ridz1208 or @kongtiaowang usually merge them into a release patch before the release.

GeorgeMurad:
OK, great. I will remove it from the release patches!

racostas:
@GeorgeMurad , some test failing: mostly indentation
image

GeorgeMurad:
@racostas I am not sure what is happening here. I fixed all the indentation automatically using phcbf.

racostas:
@GeorgeMurad, try to fix the conflicts. Maybe a rebase is need since it seem the latest activity in this PR (before this week) was a while ago.

GeorgeMurad:
Yes, you're right, but the idea is I don't see any conflicts.

GeorgeMurad:
image

PR #8573, opened 2023-04-05
Fix main.css broken link

laemtl:
Since PR #8397, the main.css. is broken since it includes the base_url twice:
https://github.com/aces/Loris/pull/8397/files#diff-d4b838452a61b616f3f7e797208ebad8d83d2f7ec9412183f691f2631cdadf64R42

https://github.com/aces/Loris/blob/main/smarty/templates/main.tpl#L5

This should fix the styling issues for the help tooltip and the user page.


zaliqarosli (approved):
fixed!

PR #8594, opened 2023-04-11
[RB] Clean deleted configs

laemtl:
#8397 deleted the css and rowsPerPage config settings. This PR clean RB with the deleted config entries.


laemtl:
@ridz1208 Done

PR #8595, opened 2023-04-11
[Imaging Browser] Fix React upgrade regression

laemtl:


driusan:
@laemtl this needs to go to 25

PR #8613, opened 2023-04-12
[issue_tracker] Fix warning

driusan:
40f4d594d3f6d21b8d843b432f013755c70b759b introduced a regression in eslint warnings

where the warning:

```
modules/issue_tracker/jsx/attachments/attachmentsList.js
  157:1  warning  Invalid JSDoc @param "event" type "Object"; prefer: "object"  jsdoc/check-types
```

was introduced.

Fix the warning and upgrade the error type from warning to error so warnings of this type don't get missed in the future, since there are no other instances.


CamilleBeau (approved):
Everything looks good. Sorry about that!

PR #8614, opened 2023-04-12
[data_release] Fix console warnings in data_release modules

driusan:
This fixes the console warnings (which were mostly proptype errors) in the data_release module.

### DIFF
--- a/modules/data_release/jsx/managePermissionsForm.js
+++ b/modules/data_release/jsx/managePermissionsForm.js
@@ -82,12 +82,10 @@ class ManagePermissionsForm extends Component {
         onClose={this.props.onClose}
         onSubmit={this.handleSubmit}
       >
-        <FormElement>
-          {Object.entries(data).map(([userId, user]) =>
-            <StaticElement
-              label={user.name}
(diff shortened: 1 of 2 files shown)


modules/data_release/jsx/managePermissionsForm.js
```diff
@@ -97,9 +95,14 @@ class ManagePermissionsForm extends Component {
...
                 </div>
-              )}
-            />
-          )}
+            );
+
+            return <StaticElement
+                      key={userId}
+                      label={user.name}
+                      text={<div>{versions}</div>}
```
laemtl, line 103: I think the div here is not necessary?

laemtl:
The PR looks good.

laemtl (approved):

modules/data_release/jsx/managePermissionsForm.js
driusan: I can't remember why but I'm pretty sure it was to fix a warning

laemtl: Oh, ok then :)

PR #8615, opened 2023-04-12
[Core] Fix React.createFactory warning

driusan:
This fixes the warning:

```
Warning: React.createFactory() is deprecated and will be removed in a future major release. Consider using JSX or use React.createElement() directly instead.
```

that appears in the javascript console of every page load of LORIS. The Help and Breadcrumbs are changed to use React.createElement (as the error suggests) while the rest of the createFactories in our core classes are unused (except for RDynamicDataTable, which had some references in mri_violations templates that have been converted to react and are no longer used.)

There are still some modules that use createFactory, but the warning will now only appear on those module pages instead of on every page in LORIS because of the usage in core JS files.


kongtiaowang (approved):
LGTM

Issue #8554, opened 2023-04-05
[DICOM Archive] archive not downloaded


regisoc:
**Describe the bug**
When clicking on any of `Archive Location` column link, no download is triggered. We are only redirected to a blank page.

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to menu > imaging > DICOM archive
2. Click on any link in the `Archive Location` column that should trigger a file download.

![Screenshot_2023-04-05_11-23-25](https://user-images.githubusercontent.com/25327259/230127926-cbc709fb-6228-4ceb-adfc-87d48cd9484b.png)

**What did you expect to happen?**
The file should be downloaded and its name should be prepended with the `Patient Name` field.

**Browser Environment (please complete the following information):**
- OS: 6.1.21-1-MANJARO
- Browser [e.g. chrome, firefox]: firefox
- Version [e.g. 22]: 111.0.1 (64-bit)

regisoc:
@kongtiaowang This is resolved today, I can download without issue. 
It should be after the modifications Cecile did to the test vm.

kongtiaowang:
@regisoc This is a RB_config data issue. I fixed this issue on the testing VM, I will send a PR to fix it.

driusan:
PR merged

PR #8618, opened 2023-04-12
[RB] add missing downloadpath 

kongtiaowang:
## Brief summary of changes

[RB] add missing downloadpath into RB files #8618

1. goto configration module
2. goto Paths
3. The download path appears here
截屏2023-04-12 上午10 44 55

[DICOM Archive] archive not downloaded #8554


SQL/Cleanup_patches/2023-04-12-add-missing-downloadpath.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent, Label, OrderNumber) SELECT 'DownloadPath', 'Where files are downloaded', 1, 0, 'text', ID, 'Downloads', 4 FROM ConfigSettings WHERE Name="paths";
```
driusan, line 1: This should already be there. There's no cleanup patch required. According to `git grep` it was added pre-14.10 and I don't see anywhere it was removed. It's also already in the default schema in `SQL/0000-00-03-ConfigTables.sql`.

SQL/Archive/Pre-14.10/2014-09-25-ConfigToDB.sql:INSERT INTO ConfigSettings (Name, Description, Visible, AllowMultiple, DataType, Parent) SELECT 'DownloadPath', 'Where files are downloaded', 1, 0, 'text', ID FROM ConfigSettings WHERE Name="paths";

SQL/Cleanup_patches/2023-04-12-add-missing-downloadpath.sql
```diff
@@ -0,0 +1,3 @@
...
+
+INSERT INTO Config (ConfigID, Value) SELECT ID, "/var/www/loris" FROM ConfigSettings WHERE Name="DownloadPath";
```
driusan, line 3: This doesn't seem like a very good default value to set it to.

driusan (changes requested):

kongtiaowang:
@driusan it has been rebased. it is good to go.

Issue #8630, opened 2023-04-13
[Mri Violation] can't load mri violation on local inner integration test - ReactDOM.createRoot(container) is not a function


kongtiaowang:
**Describe the bug**
can't load mri violation on local inner integration test - ReactDOM.createRoot(container) is not a function
**To Reproduce**
run intergration test local with node 12.

**What did you expect to happen?**
test function fail on step 8 9 10

**Browser Environment (please complete the following information):**
- OS: php 8.1 node 12
- Browser firefox 
- Version 25.0-release

driusan:
Fixed by #8651 (which replaced the above PR)

PR #8631, opened 2023-04-13
[Integration Test] can't load mri violation on local inner integration test - ReactDOM.createRoot(container) is not a function

kongtiaowang:
## Brief summary of changes
ReactDOM.createRoot(container) is not a function
git action
https://github.com/aces/Loris/issues/8630


driusan:
Fixed across LORIS by #8651 (which this conflicts with), this should no longer be needed.

PR #8633, opened 2023-04-14
[new_profile/sex] Remove strtolower() and force uppercase

ridz1208:
## Brief summary of changes

the keys in the new_profile module were lower case which forces the Sex library class to use a `strtolower()` function to validate the value and then submits the lowercase value in the SQL insert statement where SQL implicitly converts it to uppercase. This workflow is very risky as different versions of SQL or different databases may not recognise the lowercase and uppercase as the same word and treat it as a truncation. This is also simply bad practice and unnecessary here.

1. Make sure new profile still works properly and that the sex loads properly on other modules (i.e. candidate_list, timepoint list...)

### DIFF
--- a/modules/new_profile/php/new_profile.class.inc
+++ b/modules/new_profile/php/new_profile.class.inc
@@ -142,9 +142,9 @@ class New_Profile extends \NDB_Form
         $dobFormat = $config->getSetting('dobFormat');
         $edc       = $config->getSetting('useEDC');
         $sex       = [
-            'male'   => 'Male',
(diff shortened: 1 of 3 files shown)


ridz1208:
@AlexandraLivadas please make sure the case changes don't affect other modules and is not caught here.

AlexandraLivadas (approved):
new_profile is working (creates a new profile with the correct sex) and the sex is displayed on other modules properly (checked on candidate_list, timepoint_list, genomic_browser, etc)

PR #8641, opened 2023-04-18
Images cleanup

laemtl:
Migrates the documentation images to a remote server.


driusan:
1. I'm getting permission denied errors when trying to access any of those images.
2. Can we have a different server alias than amnesia.cbrain.mcgill.ca? Maybe images.loris.ca or something.

laemtl:
@driusan Fixed.

Issue #7681, opened 2021-10-04
[MRI Violated Scans] - Scan images do not upload when clicking on links on the MINC file column, leading to frozen Brain Browser page. 


SantiagoTG:
**Describe the bug**
When clicking on any of the links on the 'MINC file' column of the results table in the 'MRI Violated Scans' module, the link leads to the 'Imaging Browser' page for the selected scan. However, scan images are not uploaded and the page remains frozen with a 'Loading' message displayed. 

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. Go to 'MRI Violated Scans' module
2. Click on any of the links under the column 'MicFile' for any of the scans.
3. This will lead to a 'Imaging Browser' page for the specified scan, but the page will remain frozen with a message saying 'Loading'.

**What did you expect to happen?**
Imaging browser page would display scans.

**Browser Environment (please complete the following information):**
- Browser : Chrome

**Server Environment (if known):**
- LORIS Version: v24.0 test site

christinerogers:
Some of the links load for me, so I'll try to specify what Santiago found : e.g. 
https://test-loris-dev.loris.ca/mri_violations/ for **DCC292_676061_V1**

These MincFile links work from the data table, loading successfully in BrainBrowser: 
* assembly dwi minc:  https://test-loris-dev.loris.ca/brainbrowser/?minc_id=2l79
* Trashbin T2 weighted minc example: https://test-loris-dev.loris.ca/brainbrowser/?minc_id=1l220
* However This Trashbin (no type identified) **didn't load**: trashbin/TarLoad-14-24-W659q6/mtl004_300004_visit1_20110917_103014_12d1_mri.mnc which is : https://test-loris-dev.loris.ca/brainbrowser/?minc_id=3l44

Maybe only Trashbin mincs with no type ID'd won't load? maybe worth confirming if they're executable in the trashbin.

christinerogers:
Note: I think "upload" in this issue title and description should be changed to "load" 
-- if this is correct, @SantiagoTG could you give this comment a 👍  and I'll make the change?

SantiagoTG:
Hi Christine, I changed the title and description accordingly. (Load instead of upload). :)

PR #8643, opened 2023-04-19
Brainbrowser 404 fix

### LINKED ISSUE #8016: [Brainbrowser] Constantly loading when accessing image
**Describe the bug**
When navigating to view 3D model from imaging_browser's viewSession, a new window launches but the page never fully loads. See attached screen shot:
Screen Shot 2022-02-23 at 2 14 43 PM

**To Reproduce**
1. Create new user and assign all Imaging Browser permissions
2. Login as new user
3. Navigate to imaging browser
4. Select `all types` for a given candidate
5. For one of the images, click the checkbox in the top left corner of the image panel
6. Click `3D Only` on the left panel of the page

A new browser page to open and the scan to load.

- OS: Mac
- Browser [e.g. chrome, firefox] Chrome
- Version

- LORIS Version: 24

**Additional context**

laemtl:
- Add the brainbrowser js files to the list of files parsed by eslint
- Remove the bracket notation for the minc_id param: brainbrowser/?minc_id=[32,33] -> brainbrowser/?minc_id=32,33
- Display an error when the loader fails instead of LOADING...

Fixes #8016, #7681

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,6 +5,8 @@ babel.config.js
 
 # compiled js ignored since it is run on the jsx directory
 modules/*/js/*
+!modules/brainbrowser/js/brainbrowser.config.js
+!modules/brainbrowser/js/brainbrowser.loris.js
 modules/electrophysiology_browser/jsx/react-series-data-viewer/src/protocol-buffers/chunk_pb.js
 htdocs/js/components/*
 
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -199,7 +199,8 @@
         "modules/bvl_feedback/**",
         "modules/dataquery/**",
         "modules/dqt/**",
-        "modules/configuration/jsx/configuration_helper.js"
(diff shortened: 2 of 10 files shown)


driusan:
Why is the brackets removed? That seems like an arbitrary breaking change if there's not a reason for it.

laemtl:
> Why is the brackets removed? That seems like an arbitrary breaking change if there's not a reason for it.
The url params bracket notation is not valid and requires extra parsing. Since those extra brackets aren't necessary, I switched to the native comma separated notation that can do the same job. I carefully covered all the cases so I don't believe it will introduce any breaking changes.  I can also revert the changes if you find it safer.

laemtl:
> I don't think we should risk breaking existing links and the code to strip the `[` and `]` is already there. It's safer and keeping them makes the empty case cleaner.

Ok, I will revert the change then :)

laemtl:
@driusan Reverted

PR #8648, opened 2023-04-19
[candidate] Allow dob format Y-m

### LINKED ISSUE #8586: Unable to create candidate because the date field doesn't allow inclusion of day of the month
**Describe the bug**
unable to create candidate because the date field would not allow me to include the day of the month

**To Reproduce**
1. Go to new profile --> create new profile --> try to select DOB
2. Click on 'create'
3. See error (stopped because the date passed in is not complete

to be allowed to include a full dob

Browser Environment:

OS: [osx]
Browser: Chrome
Version [111.0.5563.110]

CamilleBeau:
## Brief summary of changes
When a dob Format of Ym (year and month without day) is set in the config module, candidates could not be created for 2 reasons. 
1. The wrong format (`!Y-m-d`) was being passed through the function `DateTime::createFromFormat` in the `createNew` function of the candidate class, which gave a wrong format error to the user even if they were passing through the expected format of Year-month
2. Even if this was fixed, the date was inserted into the candidate table as YYYY-MM, but this is not compatible with the SQL Type `date`

This PR fixes this issue by getting the date format from config, processing it to the right format and passing it through to `DateTime::createFromFormat`, as well as adding a '-01' to the end of the date if Ym format is selected in the configuration. This inserts the first of the month as the arbitrary date for a candidate.

Would someone be able to help me out as to whether this should be added to the changelog? I'm not sure if this is a bug introduced in 25 but I would be surprised if it existed in previous versions since it is so critical.

1. Attempt to create a candidate with both 'Ymd' and 'Ym' selected as dob format in the configuration module.
 
* Resolves #8586

### DIFF
--- a/php/libraries/Candidate.class.inc
+++ b/php/libraries/Candidate.class.inc
@@ -284,7 +284,7 @@ class Candidate implements \LORIS\StudyEntities\AccessibleResource,
         // Get expected format from config
         $dobFormat = $config->getSetting('dobFormat');
         $dobFormat = '!' . implode("-", str_split($dobFormat, 1));
-        $dob = DateTime::createFromFormat($dobFormat, $dateOfBirth);
+        $dob       = DateTime::createFromFormat($dobFormat, $dateOfBirth);
 
         if ($dob === false) {
             throw new InvalidArgumentException(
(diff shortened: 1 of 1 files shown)


racostas (approved):
It is working.
Have some components of hack.
Not sure what will be doing in the case we have to show at some point this DoB and we override the day to the first of each month.
I don't see any other "easy solution"
I'm approving it. 
I let the final judge to @driusan

driusan:
For display, we can update any places that display it to honour the config if they're missing.

But I'm not sure if it should be the first or middle (15th) of the month, I think different projects use different days. Added something to the LORIS agenda to discuss/decide.

Issue #8544, opened 2023-04-05
[Module_Manager] - Update test plan


charliehenrib:
**Describe the bug**
Update Module Manager testplan file with the permissions frontend name.

driusan:
Fixed by the above PR

PR #8679, opened 2023-04-26
[module_manager] - updating testplan

charliehenrib:
## Brief summary of changes

Updating test plan with Frontend name for the permissions


charliehenrib:
https://github.com/aces/Loris/issues/8544

PR #8682, opened 2023-04-27
[MRI Violations] UI improvements

### LINKED ISSUE #8670: [MRI violation] Make the resolution action more user friendly.
**Describe the bug**

Discussed at the imaging call. The current way the violations are resolved is not very user friendly. Would be best to have some feedback and an update of the resolution status column once the resolution is selected.

laemtl:
Fixes #8670


cmadjar (approved):
Works great! Thank you @laemtl :-)

PR #8689, opened 2023-05-02
[instruments] Fix LINST metadata fields not in data dictionary

zaliqarosli:
## Brief summary of changes

Currently, the metadata fields are only defined in the LINST instrument's data dictionary for the 'top' page, causing an error when trying to access those fields' values from other subtests. This is happening because the data dictionary definition is being called only when the fields' input elements are being added to the page in `_addMetadataFields` function. This PR changes to logic to follow the data dictionary definition for all other fields which get called whether or not the fields' input element is being added to the current page i.e. outside of `if ($addElements) { }` block

1.

### DIFF
--- a/php/libraries/NDB_BVL_Instrument_LINST.class.inc
+++ b/php/libraries/NDB_BVL_Instrument_LINST.class.inc
@@ -85,18 +85,6 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
         $this->dateOptions = $dateOptions;
 
         $this->addBasicDate('Date_taken', 'Date of Administration', $dateOptions);
-        $this->dictionary = array_merge(
-            $this->dictionary,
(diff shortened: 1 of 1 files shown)


zaliqarosli:
@driusan im only sending this PR to 25.0-release because I wanted to send it to 24.1-release but it depends on code only on 25. let me know if you want me to move this to main instead

zaliqarosli:
i have an override in C-OPN that calls getCandidateAge(), and therefore getFieldValue('Date_taken'), from a subtest page display function. because 'Date_taken' isn't loaded to the instrument's data dictionary outside of the 'top' page, i get a 500 error as getFieldValue checks the data dictionary first. can also discuss in loris meeting!
```
[ERROR] Invalid field Date_taken for instrument clinical_medications_questionnaire
```

PR #8694, opened 2023-05-10
[NDB_BVL_Feedback] Fixing 500 error on instrument list page

AlexandraLivadas:
A recent change (#7826) to the NDB_BVL_Feedback class causes the following 500 Error to appear when the instrument_list page is loaded. This is because references are made to `Timepoint::singleton` instead of `TimePoint::singleton`

Screenshot 2023-05-10 at 18 40 45

#### To replicate bug:

1. Checkout the branch aces/24.1-release, go to any candidate's page and click on a timepoint. When the instrument_list page loads, you should see an error both in the console and in your Apache log.

#### Testing:
1. Checkout this branch and go to any candidate's instrument list page. The error should no longer appear.


KLaFleur (approved):
This looks good and worked on my end!

PR #8713, opened 2023-05-29
Add an option to disable sortByValue on filter multiselects

laemtl:
- adds a sortByValue option that can be set to false to disable the sortByValue default behaviour on multiselects.
- the sortByValue option is already available for selects
- set the default value to true to keep the default behaviour of the existing filters.

### DIFF
--- a/jsx/Filter.js
+++ b/jsx/Filter.js
@@ -74,6 +74,7 @@ function Filter(props) {
             element = (
               <SelectElement
                 options={filter.options}
+                sortByValue={filter.sortByValue}
                 multiple={true}
                 emptyOption={false}
               />


xlecours (approved):

jsx/Filter.js
```diff
@@ -74,6 +74,7 @@ function Filter(props) {
             element = (
               <SelectElement
                 options={filter.options}
+                sortByValue={filter.sortByValue ?? true}
```
driusan, line 77: This should go into Filter.defaultProps instead of coalescing here for consistency

laemtl: I don't think that will work.
DefaultProps only work when the props is not set, but here it will be set to filter.sortByValue, which is undefined and therefore false by default.

laemtl: Wait I'm wrong I think it should work :)

laemtl: Fixed.

driusan: Did you forget to commit/push it? The coalesce is gone but I don't see the prop defined or the default in the proptypes.

xlecours: It's already there : https://github.com/aces/Loris/blob/main/jsx/Form.js#LL680C3-L680C21

PR #8714, opened 2023-05-29
[LINST] Add .meta option to skip XIN rules

driusan:
XIN rules default to requiring everything unless otherwise specified. This is not useful for all instruments (for instance, LINST files that are used to import data from other systems where we don't support their rule format) so this adds an option to disable the rules for an instrument in the .meta file.

1. Add a `norules{@}true` line to a LINST file's corresponding .meta. Rules should no longer be enforced.


xlecours (approved):

xlecours:
Simple enough.

PR #8759, opened 2023-06-08
[configuration] Use unsafeInsert/unsafeUpdate for saving values

### LINKED ISSUE #8748: [configuration module] - Study description issue
**Describe the bug**
A clear and concise description of what the bug is
When updating the study description in the configuration module the value saved is escaped. 
image

**To Reproduce**

- OS: [e.g. Plan9, HURD, Oberon, etc]

- LORIS Version: [e.g. 22.0.0]

**Additional context**

driusan:
The values get double-escaped when modified now if they contain HTML. Use unsafe variants of database calls so that the values to not get modified when re-saved.

Fixes #8748

### DIFF
--- a/modules/configuration/ajax/process.php
+++ b/modules/configuration/ajax/process.php
@@ -60,7 +60,7 @@ foreach ($_POST as $key => $value) {
                 }
             }
             // Update the config setting to the new value.
-            $DB->update(
+            $DB->unsafeUpdate(
                 'Config',
                 ['Value' => $value],
                 ['ID' => $key]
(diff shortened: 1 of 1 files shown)


modules/configuration/ajax/process.php
```diff
@@ -60,7 +60,7 @@
                 }
             }
             // Update the config setting to the new value.
-            $DB->update(
+            $DB->unsafeUpdate(
```
zaliqarosli, line 63: is this okay to change for all update/inserts on the config module, not knowing what other front-end values this might affect?

driusan: I looked on demo and only saw 3 textarea config settings:

1. The studyDescription used on the login page. This already uses DOMPurify to make sure the value is clean.
2. The projectDescription used by the dashboard welcome widget. An admin would be able update the description to include unsafe HTML there, but this is addressed by PR #8762. 
3. The citation_policy used by the acknowledgements module. This is used by the acknowledgements module inside of a React component, so it should be auto-escaped by React.

zaliqarosli (approved):

PR #8762, opened 2023-06-08
[dashboard] Load project description from ajax and run through DOMPurify

### LINKED ISSUE #8750: [dashboard] Project Description text should be called with DOM Purifier
**Describe the bug**
Related to https://github.com/aces/Loris/issues/8748, text in DB won't be sanitized in the DB, and so should be sanitized on the front-end

**To Reproduce**

- OS: [e.g. Plan9, HURD, Oberon, etc]

- LORIS Version: [e.g. 22.0.0]

**Additional context**

driusan:
This makes sure the Project Description on the dashboard runs through
DOMPurify. In order to do that, it was also necessary to move the description
from being loaded in a smarty template to being loaded from an AJAX call
(so that we can import the DOMPurify module.)

Fixes #8750


driusan:
@zaliqarosli do you think you could review/test this?

zaliqarosli:
is there more that i need to test other than the display on the dashboard not changing at all? @driusan

modules/configuration/ajax/process.php
```diff
@@ -108,7 +108,7 @@
                 }
             }
             // Add the new setting
-            $DB->insert(
+            $DB->unsafeInsert(
```
zaliqarosli, line 111: is this okay to change for all update/inserts on the config module, not knowing what other front-end values this might affect?

driusan: This is actually just because the PR is based on #8759 .. can you move your comment over there? I added the blocked tag to this PR.

driusan:
@zaliqarosli The display not changing, and any unsafe things (ie. javascript) getting stripped before being rendered.

modules/configuration/ajax/process.php
driusan: actually now that I think of it they're independent, so I removed the commit from this PR and I'm removing the blocked tag.. but can you still move your comment to the other PR?

php/libraries/NDB_Page.class.inc
```diff
@@ -743,20 +743,6 @@ class NDB_Page extends \LORIS\Http\Endpoint implements RequestHandlerInterface
...
-    /**
-     * This function can be overridden in a module's page to load the necessary
-     * resources to check the permissions of a user.
-     *
-     * @param User                   $user    The user to load the resources for
-     * @param ServerRequestInterface $request The PSR15 Request being handled
-     *
-     * @return void
-     */
-    public function loadResources(
```
zaliqarosli, line 755: should this change be in here?

driusan: `\Module::handle` calls `loadResources()` on all pages that are accessed through `https://loris/$modulename/$pagename`. NDB_Page extends Endpoint, so I moved the function to the base of the hierarchy. I could put an empty `loadResources()` in the `ProjectDescription` class instead, but I think it makes more sense to just move the definition up a level since it affects every Endpoint.

zaliqarosli: okayy that makes sense but im worried about this breaking other things that we won't pick up cz most of the testing has been done

zaliqarosli: i was thinking it could break if other classes extend Endpoint but don't have the loadResources function implemented but i see now that the function is not an abstract method so it shouldn't break anything

zaliqarosli: why does an empty loadResources need to be defined in the first place?

driusan: because it's called at https://github.com/aces/Loris/blob/main/php/libraries/Module.class.inc#L313

zaliqarosli (approved):

PR #8768, opened 2023-06-09
[data_release] Permission not added to unversioned files if version specified

### LINKED ISSUE #8741: [data_release] "Unversioned" permissions granted with every version
**Describe the bug**
When adding version permissions for a user, the "Unversioned" permission is also granted to that user unintentionally.

**To Reproduce**
1. Go to data_release
2. Make sure that there is at least one file uploaded without a version (version will show up as "Unversioned"), and at least one file uploaded under a specified version
3. Add permissions for a user to give them permission to the specified version
4. See that the user now has permission to both the specified version, and "Unversioned"

When adding permissions to one version for a user, that user should have permission to only that version, and not "Unversioned". "Unversioned" should only be granted to a user if it is added as a permission on its own.

charlottesce:
## Brief summary of changes

When adding version permissions for a user, permissions for unversioned files are no longer added 

1. Go to data_release
2. Make sure that there is at least one file uploaded without a version (version will show up as "Unversioned"), and at least one file uploaded under a specified version
3. Add permissions for a user to give them permission to the specified version
4. See that the user now only has permission to the specified version, and not "Unversioned" too

* Resolves #8741

### DIFF
--- a/modules/data_release/php/permissions.class.inc
+++ b/modules/data_release/php/permissions.class.inc
@@ -143,8 +143,7 @@ class Permissions extends \NDB_Page
             );
         } elseif (empty($releaseid) && !empty($version)) {
             // If there is no release id but there is a version, give
-            // access to every file with that version AND every file
-            // where there is no version.
+            // access to every file with that version.
             if ($version == 'Unversioned') {
                 $IDs = $DB->pselectCol(
                     "SELECT id FROM data_release WHERE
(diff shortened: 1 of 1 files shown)


driusan:
@CamilleBeau you reported the issue this is reported to fix, can you review/test?

CamilleBeau (approved):

CamilleBeau:
Works well & code looks good to me!

PR #8774, opened 2023-06-12
[api] Add cohorts to project info

driusan:
The API documentation says you can create a visit by sending a PUT request to the visit label including all of the visit meta data that the endpoint returns, but there's no way to get a list of cohorts which is part of the required data for creating a new visit.

This makes the following changes to v0.0.4-dev in order to support the creation of visits through the API without resorting to out-of-band information:
1. Adds cohorts to /projects/$projectname (similarly to how there's a visits key)
2. Renames the "Battery" key to "Cohort" in v0.0.4-dev (since it's no longer a SubprojectID.)

Ideally, cohorts would be more comprehensively supported by the API, but this is the minimal that seems to be required in order to create visits through the API and is consistent with how projects visits are handled.

### DIFF
--- a/modules/api/php/views/project.class.inc
+++ b/modules/api/php/views/project.class.inc
@@ -58,7 +58,11 @@ class Project
             'Instruments' => array_keys(\Utility::getAllInstruments()),
             'Visits'      => $this->_getVisits(),
         ];
-        if ($this->apiversion == 'v0.0.4-dev') {
+        switch ($this->apiversion) {
+        case 'v0.0.3':
+            return $obj;
+        default:
+            // >= v0.0.4
             $obj['Cohorts'] = array_map(
                 function ($cohort) {
                     return $cohort['title'];


modules/api/php/views/visit.class.inc
```diff
@@ -43,11 +43,12 @@ class Visit
      */
     public function toArray(): array
     {
+        $cohortKey = $this->apiversion == 'v0.0.4-dev' ? 'Cohort' : 'Battery';
         $meta = [
             'CandID'  => $this->_timepoint->getCandID(),
             'Visit'   => $this->_timepoint->getVisitLabel(),
             'Site'    => $this->_timepoint->getPSC(),
-            'Battery' => $this->_timepoint->getData('CohortTitle'),
+            $cohortKey => $this->_timepoint->getData('CohortTitle'),
```
xlecours, line 51: PHPCS won't be happy here. Indentation

driusan:
This should probably not go in without #8778. Might wait until after the release.

modules/api/php/views/project.class.inc
```diff
@@ -51,12 +52,21 @@ class Project
...
         $meta = ['Project' => $this->_project->getName()];
 
-        return [
+        $obj = [
             'Meta'        => $meta,
             'Candidates'  => $this->_project->getCandidateIds(),
             'Instruments' => array_keys(\Utility::getAllInstruments()),
             'Visits'      => $this->_getVisits(),
         ];
+        if ($this->apiversion == 'v0.0.4-dev') {
```
xlecours, line 61: Should we use this so w don't have to go over every version checks each time we have a new version?

```
if ($this->apiversion >= 'v0.0.4') {
```

driusan: I don't think that will work since 'v0.0.4' is a string, but I can make it a switch to make it easier to add new versions

ridz1208 (approved):
LGTM

modules/api/php/views/project.class.inc
xlecours: It would work but there is issues when/if we get to 2 digits parts
~$ php -r 'var_dump("v1.0.14" <= "v1.0.5");'
bool(true)

PR #8776, opened 2023-06-13
[api] Log filename in error message

driusan:
There is an error message in the web server logs when a file is in the database but not on the filesystem but it doesn't tell you where it was looking on the filesystem. This adds the full path to make it easier to determine where LORIS is trying to find the file when ie. setting up raisinbread.


laemtl:
Does this need a rebase?

KLaFleur (approved):

PR #8777, opened 2023-06-13
[instrument] norules meta instruction

regisoc:
## Brief summary of changes

Port #8714 to `24.1`.

1. take one linst instrument.
2. add a `.meta` file and/or add one line inside it: `norules{@}true`.
3. go to the frontend > candidate list > choose a visit and timepoint with the corresponding instrument.
4. try entering/modifying a form value with an empty answer (e.g. a blank value in a dropdown).
5. This should save the form.
6. Remove the `norules{@}true` from the `meta file`.
7. try saving the form from the frontend again, it should make an error.


driusan:
This is a new feature. Why would it go in 24.1?

regisoc:
Will be put as an override in HBCD.

PR #8778, opened 2023-06-13
[api] Test multiple versions

driusan:
This attempts to add tests for multiple versions of the API. The tests from v0.0.4-dev are copied to v0.0.3 classes and the version in the new classes changed to v0.0.3.

### DIFF
--- a/test/phpunit.xml
+++ b/test/phpunit.xml
@@ -2,6 +2,7 @@
 <phpunit
         bootstrap="bootstrap.php"
         verbose="true"
+        testdox="true"
         colors="true">
     <testsuites>
         <testsuite name='LorisUnitTests'>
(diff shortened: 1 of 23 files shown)


driusan:
I don't think this should have passed. Needs investigation.

driusan:
@kongtiaowang can you review this?

test/phpunit.xml
```diff
@@ -2,6 +2,7 @@
 <phpunit
         bootstrap="bootstrap.php"
         verbose="true"
+        testdox="true"
```
kongtiaowang, line 5: What is its purpose here? It was not found to be used. The rest of the code I think is fine, except for this little problem.

kongtiaowang (approved):
LGTM

PR #8779, opened 2023-06-13
[imaging_qc] Improve Error Message

### LINKED ISSUE #5325: Improve error message when mri_parameter_form doens't exist
From loris-dev mailing list

> I can't load the page for quality_control. I get the following error in the loris-error.log, but in short the problem is "Table 'LORIS.mri_parameter_form doesn' exist"
image

The error message displayed here should say something more helpful along the lines of "The MRI parameter form instrument must be installed in order to use this module."

skarya22:
## Brief summary of changes

- Changed the error message to be more specific

1. run `alter table mri_parameter_form rename to mri_parameter_form2;`
2. Try to open Imaging>Imaging Quality Control
3. See the new error message

* Resolves #5325 

New error:
image

**If there is an error caused by something else since mri_parameter_form exists:**
image

### DIFF
--- a/modules/imaging_qc/jsx/imagingQCIndex.js
+++ b/modules/imaging_qc/jsx/imagingQCIndex.js
@@ -96,7 +96,6 @@ class ImagingQCIndex extends Component {
     if (!this.state.isLoadedImg) {
       return <Loader/>;
     }
-
     if (Object.keys(this.state.ImgData).length > 0) {
       const ImgOptions = this.state.ImgData.fieldOptions;
 
--- a/modules/imaging_qc/jsx/imagingQCIndex.js
+++ b/modules/imaging_qc/jsx/imagingQCIndex.js
@@ -96,6 +96,7 @@ class ImagingQCIndex extends Component {
     if (!this.state.isLoadedImg) {
       return <Loader/>;
     }
+    console.log(this.state)
     if (Object.keys(this.state.ImgData).length > 0) {
       const ImgOptions = this.state.ImgData.fieldOptions;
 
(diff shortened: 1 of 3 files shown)


modules/imaging_qc/jsx/imagingQCIndex.js
```diff
@@ -217,7 +217,10 @@ class ImagingQCIndex extends Component {
     } else {
       return (
         <div>
-          <h3>An error occurred while loading the page.</h3>
+          <h3>
+            The MRI parameter form instrument must be installed
```
driusan, line 221: What if an error other than the MRI parameter_form not being setup occurs?

skarya22: Currently any error from the php script leads to this error being shown. I am unable to figure out how to make a connection between php file and the jsx code in-order to tell it what the exact error is. Any help on this would be really appreciated :)

driusan: You would need to catch the exception in PHP, and then return a JSON error response. The JSX could then display that error message with the "An error occurred" as a fallback.

modules/imaging_qc/php/imaging_qc.class.inc
```diff
@@ -386,5 +389,29 @@ class Imaging_QC extends \NDB_Menu_Filter
...
+     *
+     * @return ResponseInterface The PSR15 response for the page.
+     */
+    public function handle(ServerRequestInterface $request) : ResponseInterface
+    {
+        $db = $this->loris->getDatabaseConnection();
+        if ($db->tableExists('mri_parameter_form')) {
+            return parent::handle($request);
+        } else {
+            return (new \LORIS\Http\Response())
```
skarya22, line 408: @driusan I am getting this error in the tests because of this line, do you know how to fix it?
`PhanUndeclaredMethod Call to undeclared method \Psr\Http\Message\MessageInterface`
It works when running it locally :/

sruthymathew123:
@driusan @skarya22 Tested this PR and pass the manual tests.
Screen Shot 2023-08-10 at 11 14 59 AM

PR #8782, opened 2023-06-14
[battery_manager] Update Test Plan

### LINKED ISSUE #8739: [battery_manager] update test plan
**Describe the bug**

Test plan needs an update.

1.

```
Testing Change Status column
    Press the Deactivate button in the Change Status column on an entry in the data table.
        Ensure that when you press "Yes", a success message appears and the table data refreshes.
```

No `Yes` confirmation option is shown. When clicking the deactivate button, the change is immediate.

2.

```
Instrument, Stage, Cohort, Visit Label, Site, First Visit, Instrument Order, and Active (as dropdown fields with blank default option).
```

This line should not include `Instrument Order` (not a dropdown field).

3.

```
Try to edit an entry so that it has the same values as another deactivated entry in the Test Battery. 
```

We cannot create an entry with the same values, so I am not sure how to test this.

Change the test plan

skarya22:
## Brief summary of changes

- Update test plan for Battery Manager with the requested changes

* Resolves #8739

### DIFF
--- a/modules/battery_manager/test/TestPlan.md
+++ b/modules/battery_manager/test/TestPlan.md
@@ -68,7 +68,7 @@ Click the "New Test" button in the upper right corner of the data table.
 
 **Testing Change Status column**
   1. Press the `Deactivate` button in the `Change Status` column on an entry in the data table.
-     - Ensure that when you press "Yes", a success message appears and the table data refreshes.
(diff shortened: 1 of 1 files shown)


driusan:
@ridz1208 can you review?

PR #8783, opened 2023-06-14
[Tests] Remove duplicated login test

### LINKED ISSUE #8780: [api] login test run multiple times
In the raisinbread api tests located in `raisinbread/test/api`, there is a base LorisApiAuthenticatedTest class which implements testLoginSuccess. All the other API tests extend this class and inherit the method, resulting in the same test being run 9 times. (each taking ~10 seconds).

driusan:
Test the login once, instead of testing it once per class by moving it out of the base class.

Resolves #8780


kongtiaowang (approved):
LGTM

Issue #8756, opened 2023-06-08
[instrument_manager] Error got during uploading instrument.


racostas:
**Describe the bug**
When uplaoding and instrument throw the instrument_manager and error is got. 

**To Reproduce**
Steps to reproduce the behavior (attach screenshots if applicable):
1. MainMenu->Admin->Instrument Manager (having the permission `instrument_manager_write` granted)
2. Try to upload and instrument (ample .linst instrument file` test_all_fields.linst` in `test/instruments` suggested in the Test Plan)
3. See error
![image](https://github.com/aces/Loris/assets/37309344/6ff6e5af-acd8-4d4d-8aca-0f0f094a6b9f)

**What did you expect to happen?**
The file to be properly uploaded.

**Browser Environment (please complete the following information):**
- MsWindows
- Chrome Version 114.0.5735.110 (Official Build) (64-bit)

**Server Environment (if known):**
- Loris 25 candidate release.

driusan:
@zaliqarosli Since this was on copn-stg and it might be backend/configuration related I'm assigning it to you.

driusan:
This was caused by a misconfiguration as per comment on #8786

PR #8786, opened 2023-06-14
[instrument_manager/tools] Fix SQL generation to fix installation error

zaliqarosli:
## Brief summary of changes

Fixes https://github.com/aces/Loris/issues/8756

1. MainMenu->Admin->Instrument Manager (having the permission instrument_manager_write granted)
2. Try to upload and instrument (ample .linst instrument file test_all_fields.linst in test/instruments suggested in the Test Plan)
3. You should receive a success confirmation

https://github.com/aces/Loris/issues/8756

### DIFF
--- a/tools/generate_tables_sql_and_testNames.php
+++ b/tools/generate_tables_sql_and_testNames.php
@@ -99,7 +99,8 @@ foreach ($instruments as $instrument) {
     $fp = fopen($filename, "w");
     fwrite($fp, $output);
     $sql = "REPLACE INTO test_names (Test_name, Full_name, Sub_group) "
-         . "VALUES ('$tablename', '$title', 1);";
+         . "SELECT '$tablename', '$title', ID "
+         . "FROM test_subgroups WHERE Subgroup_name='Instruments';\n";
 
     fwrite($fp, $sql);
     foreach ($pages as $pageNo => $page) {


driusan:
@racostas Looks like you opened the issue. Can you review/test that this solves it?

tools/generate_tables_sql_and_testNames.php
```diff
@@ -99,7 +99,8 @@
     $fp = fopen($filename, "w");
     fwrite($fp, $output);
     $sql = "REPLACE INTO test_names (Test_name, Full_name, Sub_group) "
-         . "VALUES ('$tablename', '$title', 1);";
+         . "SELECT '$tablename', '$title', ID "
+         . "FROM test_subgroups WHERE Subgroup_name='Instruments';\n";
```
driusan, line 103: I don't think there's any guarantee that there's a subgroup named "Instruments" either. We don't have any guarantees for what the subgroups names are other than that there's at least 1 of them, which I think is why it was using a hardcoded 1. This might just be trading it breaking on some environments for breaking on others. I'm not sure what the solution is here.

Two options I can think of are:
1. Prompt for which subgroup it should go into in the instrument manager. (Probably a lot more work.)
2. Use `MIN(ID)` (or `MAX(ID)`) (minimal work and should work in all environments but will sometimes go into the wrong subgroup.. though that's also the case now.)

zaliqarosli: isn't in guaranteed if its part of our schema? https://github.com/aces/Loris/blob/189a43d15dfba0c6584f1264682a6baa0da9b874/SQL/0000-00-00-schema.sql#L249

if a DB veers from that, then they should have an override to ensure there's no breaking changes?

driusan: No, that's just the default data. They can organize it however they want. Many projects group them differently.

driusan: But "Instruments" is a database value here, not a part of the schema.

zaliqarosli: yeah the value is inserted as part of the schema (linked above)

racostas: Hi @driusan, the original issue reported by me was due a misconfiguration (mean the issue don't appear in a properly configured environment @zaliqarosli and myself tested.)

Regarding "Instruments" or "1"  agree is more a "feature" than a "bug fix" but since "Instruments" is inserted "by a default installation" I don't see it worse than "1".

zaliqarosli:
@driusan i still see this being a LORIS issue. the instrument manager calls the script `tools/generate_tables_sql_and_testNames.php` which will always throw an error if the ID '1' doesn't exist. the script currently hardcodes a value that isn't guaranteed in the DB, this isnt a misconfiguration

driusan:
But the change will make it throw an error if the database doesn't contain a row with the name "Instruments", which isn't guaranteed by the DB.

zaliqarosli:
ya so maybe this PR isn't the solution but the issue is still an issue that hasn't been resolved. in my opinion, the LORIS default install does guarantee a row with the name "Instruments", and if it doesn't then the LORIS requires an override for this script so that the instrument manager can function accordingly

zaliqarosli:
no it doesn't, an ID of 1 isn't guaranteed. some DBs have IDs starting with 2, as is with newer mariadbs, where IDs skip i.e.

```
mysql> select * from test_subgroups;
+----+---------------+-------------+
| ID | Subgroup_name | group_order |
+----+---------------+-------------+
|  2 | Instruments   |        NULL |
|  5 | Imaging       |        NULL |
|  7 | Genetics      |        NULL |
|  8 | Consent       |        NULL |
+----+---------------+-------------+
4 rows in set (0.01 sec)
```

driusan:
Made issue #8756

PR #8790, opened 2023-06-15
[bvl_feedback] Edit comment section

skarya22:
## Brief summary of changes
- Moved [CCNA overrides to bvl_feedback module](https://github.com/aces/CCNA/pull/6360) to LORIS Core
- Switched Add Comment button from Pencil to a Comment (Original was confusing)
- Set threads to be shown automatically for open threads as having them hidden made the UI confusing.
- Added an Edit and Delete button for comments of which the author is the user viewing them
- Flipped the order of comments around so that the newer comments show up below. Makes more sense when reading the comments
- Made New Comment TextArea section show up below the thread, as that is where the new comment will appear
- Changed panel width to work on mobile devices

#### Comment section
Behaviour:
- Defaults to closed
- Date is not shown
- Comments are ordered in a confusing way
image

#### Mobile device
(Everything is cut off)
image

### After:
Behaviour:
- Defaults to open
- Date is shown
- Comments are in an easier to read order
- Ability to delete and edit comments you wrote
image

#### Mobile device

image

### DIFF
--- a/modules/bvl_feedback/jsx/react.behavioural_feedback_panel.js
+++ b/modules/bvl_feedback/jsx/react.behavioural_feedback_panel.js
@@ -3,6 +3,8 @@
 import React, {Component} from 'react';
 import PropTypes from 'prop-types';
 
+import '../css/bvl_feedback_panel.css';
+
 /**
  * Slider panel component
  */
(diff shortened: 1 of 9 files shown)


modules/bvl_feedback/jsx/react.behavioural_feedback_panel.js
```diff
@@ -279,6 +282,42 @@ class FeedbackPanelRow extends Component {
...
+      loris.BaseURL +
+      '/bvl_feedback/ajax/delete_thread_comment_bvl_feedback.php',
+      {
+        method: 'POST',
+        body: formData,
+      }
+    ).then(() => {
+      this.loadServerState();
+    }).catch((error) => {
+      console.log(error);
```
driusan, line 317:
```suggestion
      console.error(error);
```

PR #8798, opened 2023-06-19
Eeg uploader status update

laemtl:
- update the electrophysiology_uploader status when archiving an upload.
- add a new status, 'Failed Extraction'.


regisoc:
@laemtl I guess the new status will be used in LORIS-MRI? I do not see any reference to it right now .

laemtl:
> @laemtl I guess the new status will be used in LORIS-MRI? I do not see any reference to it right now .

Here is the reference: https://github.com/aces/Loris-MRI/blob/24.1-release/python/extract_eeg_bids_archive.py#L240

regisoc (approved):
LGTM

PR #8802, opened 2023-06-20
[api/tests] Run API tests in separate job

driusan:
This removes the API tests from the integrationtests target and moves them into a separate job in GitHub Actions. The result is that we can see the status of tests against different API versions without trying to debug the dockerized integration tests.

Since the API tests do not depend on Selenium, these tests are run directly on the Github Actions image and bypass docker compose. A step is added at the end to print the error log, so that it's easier to see what went wrong.

Theoretically, this should also make the integration tests finish faster since they no longer run the API tests. (In practice, this is going to depend on the number of jobs GitHub runs in parallel.)


kongtiaowang (approved):
LGTM

Issue #8804, opened 2023-06-21
getInstanceData inconsistent between JSON and table-based instruments


driusan:
The getInstanceData method (and hence, the API instrument's endpoint for instrument data) has inconsistent behaviour based on if the instrument saves the data to JSON in the flag table or whether it saves it to a table for the instrument.

## Inconsistency 1
Table-based data does `SELECT * from $tablename WHERE CommentID=..`, resulting in an array of keys set to the value `NULL` for every column in the table.

JSON-based data does `SELECT Data FROM flag WHERE CommentID=...`. since Data is nullable, it coalesces to the empty array. The result is that empty($result) is true for JSON-based data but false for table-based data.

## Inconsistency 2
The CommentID is included in the table as a primary key. This does not get included in the flag instance data. (Resolved by #8805)

## Needs investigation
UserID, Examiner, and Testdate may or may not be inconsistent for the same reason as point 2.

driusan:
Fixed by the above 2 PRs

PR #8805, opened 2023-06-21
[instruments] Remove CommentID from getInstanceData.

driusan:
The CommentID is not part of the data, it's the foreign key used between the flag table and the instrument table. JSON-based instruments do not have it, and this ensures better consistency between the two so that issues such as #8796 and #8801 will not vary based on instrument type and will be caught sooner.

Resolves part of #8804 (Inconsistency #2)


driusan:
@xlecours assuming this passes the tests can you review it?

xlecours (approved):
Let's see if other tests will break.

PR #8816, opened 2023-06-26
[tests] Speed up tests by caching Raisinbread in Docker image

driusan:
This increases the speed of our API tests by removing the sourcing of raisinbread from the test job and having a separate job conditionally build a docker image containing raisinbread to push to ghcr.io. The tests then use that database Docker image instead of duplicating the work of sourcing the entire database with every run. The DB image is only re-built when one of the files that it depends on is modified.

(The docker-compose tests can not be sped up this way since it depends on GitHub Actions calculating the image to use, but the API tests do not use docker-compose.)

### DIFF
--- a/.github/workflows/loristest.yml
+++ b/.github/workflows/loristest.yml
@@ -1,21 +1,74 @@
 name: LORIS Test Suite
-
 on:
   - push
   - pull_request
-
 env:
-  EEG_VIS_ENABLED: 'true'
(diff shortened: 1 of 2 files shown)


ridz1208:
@driusan this is going to 25, should it be going to a more recent branch ?

ridz1208:
@driusan ultimately your call

PR #8819, opened 2023-06-27
[tests] Increase speed by removing duplication of work between jobs

driusan:
This increases the speed of our tests by using separate jobs on GitHub Actions for building and testing. Currently every job in our test matrix does a complete build of all javascript, PHP, and sourcing of the database (often twice inside of the docker compose image.) About 2 minutes of each test job is building LORIS and another 3-4 sourcing raisinbread while < 1 minute of each job is spent actually running the tests (except for the integration tests, which are slow.)

This fixes that for the API tests so that there is 1 job to build the javascript (one time, instead of once per test job), 1 job to build the PHP (once for each version of PHP, instead of once for each version of PHP times the number of tests in the test matrix) and has the test jobs depend on those building jobs and use their output instead of duplicating the work 16 times.

Mostly, this will be noticable in combination with #8816 (which does a similar optimization for the Raisinbread database) when building javascript (slow) and raisinbread (also slow) will happen in parallel.

This is only done for the API tests (which don't depend on docker compose, where much of the work is duplicated yet again) but after #8816 should be fairly easy to extend to the unit tests and static tests (which don't depend on Selenium.)

### DIFF
--- a/.github/workflows/loristest.yml
+++ b/.github/workflows/loristest.yml
@@ -4,12 +4,97 @@ on:
   - push
   - pull_request
 
-env:
-  EEG_VIS_ENABLED: 'true'
 
 jobs:
+  buildjs:
+    env:
+      EEG_VIS_ENABLED: 'true'
+    runs-on: ubuntu-latest
+    steps: 
+      - uses: actions/checkout@v2
+
+      - name: Install EEG package dependencies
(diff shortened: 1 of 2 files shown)


kongtiaowang (approved):
LGTM

PR #8824, opened 2023-07-04
[API] Using isAccessibleBy on candidate endpoint

xlecours:


ridz1208 (approved):
Approving although I realize now that it may be redundant (still not a bad idea though) because all subendpoints (visits) already have the isaccessible by and both hendleGET and handlePUT check for user sites and projects.

I think it's the ad-hoc locations of the site/project checks that is a bit confusing.

Either way the redundancy is welcome here

xlecours:
interesting. I'll change /candidate/$candid

xlecours:
@driusan passing

PR #8828, opened 2023-07-06
Fix for inactive module based issues

### LINKED ISSUE #8075: [issue_tracker] Cannot access issue tied to inactive module
Let's say you create an issue and associate it to a given module. Later on, you deactivate that module and try to access the issue: you'll get an HTTP 500 error. I don't think that's the appropriate behaviour but I am not sure what the appropriate behaviour should be:
- Display the issue as usual?
- Display the issue in read-only mode?
- Don't display the issue?
- Other?

Need specs on this. 

NB: This was found in IBIS (runs LORIS 23) but I believe it is present in LORIS 24 also.

sanjay-thiyagarajan:
## Brief summary of changes

- Update the comment for module to reflect on the unavailability of the module linked to the issue

* Resolves #8075


racostas (approved):
LGTM.

PR #8833, opened 2023-07-11
[Docs] Update references from LORIS 24 to LORIS 25 in documentation

driusan:
This updates the documentation that references LORIS 24 to instead reference LORIS 25 in preparation for the release.

I removed the module testing issue template that was used for the 24 release because we didn't use it for 25 and the release testing phase is over at any rate. Other than that, there were fewer references than I expected, but this is all I could find with `git grep 24` (and a lot of false positives such as ids in raisinbread data.)


.github/ISSUE_TEMPLATE/module_testing.md
```diff
@@ -1,35 +0,0 @@
----
```
ridz1208, line 1: why delete this? should we keep it for 26?

driusan: We forgot about it and didn't use/update it for 24.1 or 25, so it's out of date and has the wrong label and wasn't used for the last two releases. 

I figured it's easiest to just clean it up and if we want to use a template for 26 we can bring it back and update it then.

PR #8841, opened 2023-07-19
[issue_tracker] Do not display inactive users in issue form

charlottesce:
## Brief summary of changes
Inactive users are not displayed in dropdowns to assign / watch an issue

1. Go to Tools -> Issue Tracker
2. Make a new issue / edit an existing one -> ensure you don't see any inactive user in the "Assignee" nor "Add others to watching" dropdowns

#### Note

This is a CCNA override - https://github.com/aces/CCNA/pull/4100

### DIFF
--- a/modules/issue_tracker/php/edit.class.inc
+++ b/modules/issue_tracker/php/edit.class.inc
@@ -66,7 +66,7 @@ class Edit extends \NDB_Page implements ETagCalculator
         $inactive_users = [];
         if ($user->hasPermission('access_all_profiles')) {
             $assignee_expanded = $db->pselect(
-                "SELECT Real_name, UserID FROM users",
+                "SELECT Real_name, UserID FROM users WHERE Active='Y'",
                 []
             );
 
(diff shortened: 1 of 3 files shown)


modules/issue_tracker/php/edit.class.inc
```diff
@@ -66,7 +66,7 @@ class Edit extends \NDB_Page implements ETagCalculator
         $inactive_users = [];
         if ($user->hasPermission('access_all_profiles')) {
             $assignee_expanded = $db->pselect(
-                "SELECT Real_name, UserID FROM users",
+                "SELECT Real_name, UserID FROM users WHERE Active='Y'",
```
CamilleBeau, line 69: And so on and so forth for the other changes here
```suggestion
                "SELECT Real_name, UserID FROM users WHERE Active='Y' AND Pending_approval='N'",
```

CamilleBeau (changes requested):
This is working well! I am however wondering now if we should also not display users that have Pending_approval='Y'?

CamilleBeau (changes requested):
As per Dave's request, can we add the assignee / watchers that are currently assigned to the issue to the drop down even if they are inactive? Should be simple to do with a join on issues using issueID from $values, and an OR statement

charlottesce:
Two other changes have been added as bug fixes I noticed during testing: 
1. Added a getBreadcrumbs function for the edit class, otherwise we get the following error when clicking on the 'Issue' link of the path on the top left of the page, as it links to 'issue_tracker/issue' which doesn't exist.
image

2. The updateComments function of the edit class can now take a null value as input, otherwise if someone tries to update an issue without adding a comment, a PHP error is raised that updateComments expects a string while null is given.

CamilleBeau (approved):

CamilleBeau:
Working well!

PR #8842, opened 2023-07-19
[issue_tracker] Resolves special characters in titles

charlottesce:
## Brief summary of changes
Resolves how special characters were appearing encoded in the title in the filterable table and the title of the issue itself e.g. `&amp;, &quot;, etc.` instead of `&, ", etc.`

1. Go to Tools -> Issue Tracker
2. Create a new issue and put the special characters `&, <, >, "` in the title; fill out necessary fields -> click Submit 
a. In the main page, the title of the issue in the table should have the characters above as is, and not "`&amp;`", etc. 
3. Click to Edit the issue you just created
a. In the edit form, the title of the issue in the table should again have the characters above, and not "`&amp;`", etc. 

#### Note

This is a CCNA override - https://github.com/aces/CCNA/pull/4032, https://github.com/aces/CCNA/pull/4084

### DIFF
--- a/modules/issue_tracker/jsx/IssueForm.js
+++ b/modules/issue_tracker/jsx/IssueForm.js
@@ -55,6 +55,23 @@ class IssueForm extends Component {
   }
 
   /**
+   * Restores special characters in the text string
+   * @param {string} text
+   * @return {string}
+   */
+  htmlSpecialCharsDecode(text) {
+    if (text != null) {
+      return text
+        .replace(/&amp;/g, '&')
+        .replace(/&quot;/g, '"')
+        .replace(/&lt;/g, '<')
+        .replace(/&gt;/g, '>');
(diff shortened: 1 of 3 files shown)


regisoc (approved):
Tested. LGTM.

ridz1208 (changes requested):
no no no no no no no no no no no

ridz1208:
@charlottesce The current solution is only masking the fact that the data is encoded for security but never decoded. The real solution here is that we simply no longer need to encode the data and should instead be saving it with unsafeupdate and unsafeinsert instead.

see my comments on the lines

ridz1208 (approved):
bellissimo !

@driusan please double check the security risk

driusan:
do we need an unescaping script to fix existing data?

PR #8845, opened 2023-07-19
[conflict_resolver]  Add description field

charlottesce:
## Brief summary of changes

This adds a 'Description' field to the Conflict Resolver module (both Resolved and Unresolved tabs), which corresponds to the description of the 'Question' with the conflict. Instead of using parameter_type like in the CCNA code, it uses /dictionary/module/instruments to get the description of the fields. Since some fields don't have descriptions (e.g. some _status variables), the empty string is used as default. 

Changes have been made to the Instrument_LINST file such that numeric and date elements are added to the dictionary, even if they are not in the top page (inspired by the change in https://github.com/aces/Loris/pull/8869). 

1. Go to Clinical -> Conflict Resolver 
2. Ensure both tabs 'Unresolved' and 'Resolved' load correctly, and in both, ensure that:
a. You can see a 'Description' selection filter and the field in the table with conflicts 
b. You can input some value in the 'Description' selection filter (e.g. 'date') and that conflicts are filtered appropriately
3. Make sure the rest of the module works as expected according to the test plan 

#### Note

This is a CCNA override - https://github.com/aces/CCNA/pull/6138

### DIFF
--- a/modules/conflict_resolver/php/provisioners/resolvedprovisioner.class.inc
+++ b/modules/conflict_resolver/php/provisioners/resolvedprovisioner.class.inc
@@ -39,6 +39,7 @@ class ResolvedProvisioner extends \LORIS\Data\Provisioners\DBObjectProvisioner
               session.Visit_label as visitlabel,
               Project.Name as project,
               conflicts_resolved.FieldName as question,
+              parameter_type.Description as description,
               conflicts_resolved.OldValue1 as value1,
               conflicts_resolved.OldValue2 as value2,
               CASE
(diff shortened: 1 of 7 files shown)


xlecours (approved):
All good.
And it works just fine :)

modules/conflict_resolver/php/provisioners/resolvedprovisioner.class.inc
```diff
@@ -55,6 +56,9 @@ class ResolvedProvisioner extends \LORIS\Data\Provisioners\DBObjectProvisioner
               Project.ProjectID as projectid
              FROM
               conflicts_resolved
+             LEFT JOIN parameter_type
```
driusan, line 59: I don't think this is going to work with json data (I'm not sure what the conflict_resolver or parameter_type tables look like in that case, it's possible they lie about the the tablename but I hasn't looked into it) and it's not going to take overrides into account or work if the data dictionary builder hasn't been recently run.

I think what should happen instead of trying to factor all those cases into account in the conflict_resolver, is the frontend should make one call to `/dictionary/module/instruments` (since only instruments have conflicts) and get all the field metadata, and then look up the key directly instead of making changes to the backend.

xlecours (approved):
Splendide :)

PR #8854, opened 2023-08-07
[LorisForm] Read attribute

regisoc:
## Brief summary of changes

Currently, trying to put a `value` attribute in a form is not possible through the `attribs` parameter.

E.g. `$this->addBasicText('test_text', 'My test text', [], ['value' => 'test']);`  will not create the expected 

image
field but an empty field.

This is due to the value being filtered by `getValue` function https://github.com/aces/Loris/blob/f8f123db63f07f69ac36c1aacd431fbb04201a3f/php/libraries/LorisForm.class.inc#L570
The `value` attribute will not be read.

This PR adds the `attribute->value` case inside the `getValue` function so it can be executed during the form rendering process.

1. Create a form with a `value` attribute. (e.g. `$this->addBasicText('test_text', 'My test text', [], ['value' => 'test']);`).
2. Render the form and see the error (the field is not populated with the string `test`).
3. Switch to this PR branch and re-render the page.

### DIFF
--- a/php/libraries/LorisForm.class.inc
+++ b/php/libraries/LorisForm.class.inc
@@ -622,7 +622,7 @@ class LorisForm
         }
 
         // check if a 'value' attribute is set for this element
-        if (!empty($this->form[$name]['value'])) {
+        if (isset($this->form[$name]['value']) && !empty($this->form[$name]['value'])) {
             return $this->_getFilteredValue(
                 $name,
                 $this->form[$name]['value']
--- a/php/libraries/LorisForm.class.inc
+++ b/php/libraries/LorisForm.class.inc
@@ -622,7 +622,9 @@ class LorisForm
         }
 
         // check if a 'value' attribute is set for this element
-        if (isset($this->form[$name]['value']) && !empty($this->form[$name]['value'])) {
+        if (isset($this->form[$name]['value'])
+            && !empty($this->form[$name]['value'])
+        ) {
             return $this->_getFilteredValue(
                 $name,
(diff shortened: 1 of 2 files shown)


test/test_instrument/test_instrumentTest.php
```diff
@@ -165,9 +165,11 @@ function testTextElement()
     function testCheckBoxElement()
     {
         $this->_landing();
-        $this->safeFindElement(
+        $checkbox = $this->safeFindElement(
             WebDriverBy::Name("testCheckbox")
-        )->click();
+        );
+        $checkbox->click();
+        $this->assertTrue($checkbox->isSelected());
```
kongtiaowang, line 172: try to run "        $checkbox = $this->safeFindElement(
            WebDriverBy::Name("testCheckbox")
        );" again

kongtiaowang: Please remove line 172 to 175 to pass the test.

regisoc: Will do, but not sure it will pass this test. Lets see.

php/libraries/LorisForm.class.inc
```diff
@@ -621,6 +621,15 @@ class LorisForm
             }
         }
 
+        // check if a 'value' attribute is set for this element
+        if (!empty($this->form[$name]['value'])) {
```
kongtiaowang, line 625: try this
if (isset($this->form[$name]['value']) && !empty($this->form[$name]['value'])) {

PR #8855, opened 2023-08-08
[instruments] Bulk load SessionID and VisitLabel

driusan:
When calling bulkLoadInstanceData, currently only instrument data is bulk loaded from the database. However, calling getSessionID or getVisitLabel is almost always required in order to use the data. This bulk loads the session id and visit label from the same query when calling bulkLoadInstanceData in order to further optimize it.

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -1442,7 +1442,9 @@ abstract class NDB_BVL_Instrument extends NDB_Page
     function getVisitLabel(): string
     {
         if ($this->visitLabel === null) {
-            $this->visitLabel = \TimePoint::singleton($this->getSessionID())->getVisitLabel();
+            $this->visitLabel = \TimePoint::singleton(
+                $this->getSessionID()
+            )->getVisitLabel();
         }
         return $this->visitLabel;
     }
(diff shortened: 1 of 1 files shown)


php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -1409,37 +1409,44 @@ abstract class NDB_BVL_Instrument extends NDB_Page
...
+    private ?string $visitLabel = null;
     /**
      * Gets the VisitLabel for the timepoint for this instrument.
      *
      * @return string The visit label
      */
     function getVisitLabel(): string
     {
-        return \TimePoint::singleton($this->getSessionID())->getVisitLabel();
+        if ($this->visitLabel === null) {
```
laemtl, line 1444: I'm just curious, why is this done in a lazy way instead of being instantiated in the factory?

laemtl (approved):

php/libraries/NDB_BVL_Instrument.class.inc
driusan: I don't know, I didn't want to change the behaviour so I left if the way it was and just added caching.

laemtl: Makes sense :)

PR #8857, opened 2023-08-08
[Instruments] Handle empty bulkLoadInstanceData

driusan:
The logic of bulkLoadInstanceData constructs an SQL statement with the parameters bound to it by looping over the iterable passed to it. If the iterable is empty, the result will be SQL with the string `.. IN ()` which is not valid SQL. This returns an empty array in that case before attempting to run the generated SQL, since the result of bulk loading nothing should be nothing.


ridz1208 (approved):
havent seen my friend `$i` since Java

PR #8858, opened 2023-08-09
[LINST] fix survey and date issues

ridz1208:
@zaliqarosli some changes derived from the QPN overrides that I believe apply to LORIS

- add fields to XIN validation which are not parts of groups

### DIFF
--- a/php/libraries/NDB_BVL_Instrument_LINST.class.inc
+++ b/php/libraries/NDB_BVL_Instrument_LINST.class.inc
@@ -98,7 +98,9 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
             ]
         );
 
-        if (strrpos($this->testName ?? '', '_proband') === false) {
+        if (strrpos($this->testName ?? '', '_proband') === false
+            && !$this->DataEntryType == 'DirectEntry'
+        ) {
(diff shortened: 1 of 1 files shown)


laemtl:
Code looks good.

php/libraries/NDB_BVL_Instrument_LINST.class.inc
```diff
@@ -98,7 +98,9 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
             ]
         );
 
-        if (strrpos($this->testName ?? '', '_proband') === false) {
+        if (strrpos($this->testName ?? '', '_proband') === false
```
ridz1208, line 101: consider moving this up

ridz1208:
@CamilleBeau @charliehenrib @laemtl 

removed the change relating to age on direct instruments, I'll make a PR just for that

for the remaining changes I think what needs to be tested is to have a LINST instruments with 3 different types of dates `Date`, `BasicDate` and `MonthYear` (date types are defined as the 6th parameter of the date LINST element) also make sure to add some rules in the instrument.rules file and try it with and without this PR.

zaliqarosli (approved):

PR #8859, opened 2023-08-09
[survey_accounts] notices, warnings and filter options

ridz1208:
This PR was made to remove overrides on COPN and CBIGR

- removes notices
- removes deprecation warning
- add filter options to filter with limited set of options


kongtiaowang (approved):
LGTM

PR #8863, opened 2023-08-10
[LINST] skip metadatafields for surveys

ridz1208:
Skip over metadatafields and calculate ages in surveys

This is not a big change, one way or another the fields were either being skipped with an empty check, through a project override or through code in the module itself handling the submission. this is just an explicit check to make everything clearer

### DIFF
--- a/php/libraries/NDB_BVL_Instrument.class.inc
+++ b/php/libraries/NDB_BVL_Instrument.class.inc
@@ -779,7 +779,11 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     function _saveValues(array $values): void
     {
-        if (strrpos($this->testName, "_proband") === false) {
+        // Surveys do not ask for the metadata informatio nand thus date_taken is
+        // always null so there is no use in going into the savecandidateage function
+        if (strrpos($this->testName, "_proband") === false
+            && $this->DataEntryType === 'normal'
+        ) {
             $this->_saveCandidateAge($values);
         }
 
--- a/php/libraries/NDB_BVL_Instrument_LINST.class.inc
+++ b/php/libraries/NDB_BVL_Instrument_LINST.class.inc
@@ -556,7 +556,9 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
                             'instrument_title',
                             $pieces[1]
                         );
-                        $this->_addMetadataFields();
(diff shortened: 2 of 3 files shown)


php/libraries/NDB_BVL_Instrument.class.inc
```diff
@@ -779,7 +779,11 @@ abstract class NDB_BVL_Instrument extends NDB_Page
      */
     function _saveValues(array $values): void
     {
-        if (strrpos($this->testName, "_proband") === false) {
+        // Surveys do not ask for the metadata informatio nand thus date_taken is
+        // always null so there is no use in going into the savecandidateage function
+        if (strrpos($this->testName, "_proband") === false
+            && $this->DataEntryType === 'normal'
```
ridz1208, line 785: @driusan I'm a bit unsure if $this->DataEntryType is set at this stage. it seems to only be set in the `display()` method so.... I don't think it has any value here? Unless I'm missing something

zaliqarosli: @ridz1208 i think this check is good here, the display method gets called on loading an instrument, so DataEntryType is set. it's not necessarily required though cz the first line of _saveCandidateAge checks if $values['Date_taken'] is empty

php/libraries/NDB_BVL_Instrument_LINST.class.inc
```diff
@@ -556,7 +556,9 @@ class NDB_BVL_Instrument_LINST extends \NDB_BVL_Instrument
                             'instrument_title',
                             $pieces[1]
                         );
-                        $this->_addMetadataFields();
+                        if ($this->DataEntryType!=="DirectEntry") {
```
zaliqarosli, line 559: from what i can tell, the DataEntryType is never set for surveys (display method doesn't get called) so changing this check to `$this->DataEntryType === 'normal'` would fix this check

zaliqarosli (changes requested):

php/libraries/NDB_BVL_Instrument.class.inc
ridz1208: fair point

ridz1208:
closing because @zaliqarosli has a much better version

zaliqarosli:
@ridz1208 should the new PR still go to 24.1-release?

PR #8864, opened 2023-08-11
[IssueTracker] Add description field to issues table

### LINKED ISSUE #8353: [issue tracker] Can not edit description of already created issue.
In the issue tracker, you can not edit the description of an issue if it's already been created, only add new comments. This is odd. Most bug tracking software lets you edit the description.

We should let users edit descriptions on issues.

sanjay-thiyagarajan:
## Brief summary of changes

- Added "description" field to "issues" table
- Created script "update_issues_with_description.php" for copying all the existing descriptions from "issues_comments" table
- Updated raisinbread SQL file to reflect the change in field
- Added a new field to the issue creation / modification form in the UI.

* Resolves #8353


racostas:
> @racostas is this ready for your review?

Yes, I did a first pass with him "in person" before the end of the GSoC period. The comments were properly addressed but it's missing a second test/review. I had been a bit taken with CCNA upgrade since them. If after second review there is something to fix I will take care myself.

CamilleBeau (changes requested):
Missing the patch in "New_patches" directory to add the changes

Getting this error: 
image

modules/issue_tracker/php/edit.class.inc
```diff
@@ -424,6 +414,29 @@ class Edit extends \NDB_Page implements ETagCalculator
...
+            "SELECT description from issues WHERE issueID=:issueID",
+            ['issueID' => $issueID]
+        );
+
+        $description = "";
+        foreach ($desc as $d) {
+            $description = $d['description'] ?? '';
+        }
+        return $description;
+    }
```
CamilleBeau, line 438: I'm a bit confused about what's going on here, why is the description being re-set every time in a foreach loop? Wondering if something can be done to the database query to select the description instead?

tools/update_issues_with_description.php
```diff
@@ -0,0 +1,62 @@
...
+            WHERE ic.issueID = :issue_id
+            ORDER BY ic.dateAdded ASC
+            LIMIT 1",
+            ['issue_id' => $issueID]
+        );
+        $DB->update(
+            "issues",
+            ["description" => $description[0]],
+            ["issueID" => $issueID]
+        );
```
CamilleBeau, line 58: Should this be pselectOne instead of pselectCol?
```suggestion
        $description = $DB->pselectOne(
            "SELECT issueComment
            FROM issues_comments ic
            WHERE ic.issueID = :issue_id
            ORDER BY ic.dateAdded ASC
            LIMIT 1",
            ['issue_id' => $issueID]
        );
        $DB->update(
            "issues",
            ["description" => $description],
            ["issueID" => $issueID]
        );
```

SQL/New_patches/2024-02-06-issuetracker_AddsDescriptionToIssuesTable.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE `issues`
+    ADD `description` longtext DEFAULT NULL
+    AFTER `category`;
+
+ALTER TABLE `issues_history`
+    MODIFY `fieldChanged` enum('assignee','status','comment','sessionID','centerID','title','category','module','lastUpdatedBy','priority','candID', 'description') NOT NULL DEFAULT 'comment'
+    AFTER `dateAdded`;
```
CamilleBeau, line 7: I don't think we need the "AFTER" here since it's a column modification

CamilleBeau (changes requested):

CamilleBeau (changes requested):
Everything works well now. The only thing (but I don't think it's related to this PR) is that the "New issue" page does not load because there is a missing jsx/Form import for FileElement. This can be solved by adding this line in IssueForm.js
image

I added it to test. Even though it's not caused by this PR I think it would be good to fix here

racostas:
> Everything works well now. The only thing (but I don't think it's related to this PR) is that the "New issue" page does not load because there is a missing jsx/Form import for FileElement. This can be solved by adding this line in IssueForm.js image
> 
> I added it to test. Even though it's not caused by this PR I think it would be good to fix here

Yes, definitely something we can address in as part of this PR too. Thanks Camille.

racostas:
ready to be re-reviewed.

ay-bh:
@racostas I have tested the changes, and they work as expected without any issues. Looks good to me!

ay-bh (approved):

driusan:
@CamilleBeau GitHub still says you requested changes, can you re-review?

ay-bh (changes requested):
tools/update_issues_with_description.php gives this error. 
``` [DEBUG] SELECT ID FROM users WHERE UserID =:lorisadmin LIMIT 2
[DEBUG] SELECT issueID FROM issues_comments
[DEBUG] SELECT `description` FROM issues WHERE issueID=:14 LIMIT 2
[DEBUG] SELECT issueComment
            FROM issues_comments ic
            WHERE ic.issueID = :14
            ORDER BY ic.dateAdded ASC
            LIMIT 1 LIMIT 2
PHP Fatal error:  Uncaught PDOException: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'LIMIT 2' at line 5 in /var/www/loris/php/libraries/Database.class.inc:695
Stack trace:
#0 /var/www/loris/php/libraries/Database.class.inc(695): PDOStatement->execute()
#1 /var/www/loris/php/libraries/Database.class.inc(745): Database->execute()
#2 /var/www/loris/php/libraries/Database.class.inc(841): Database->pselect()
#3 /var/www/loris/php/libraries/Database.class.inc(981): Database->pselectRow()
#4 /var/www/loris/tools/update_issues_with_description.php(52): Database->pselectOne()
#5 {main}
  thrown in /var/www/loris/php/libraries/Database.class.inc on line 695

**Review by CamilleBeau — APPROVED** [review]

**Comment by CamilleBeau** [comment]
> @CamilleBeau GitHub still says you requested changes, can you re-review?

@driusan Done!

File: tools/update_issues_with_description.php [hunk 4]
```diff
@@ -0,0 +1,62 @@
...
+        "SELECT `description` FROM issues WHERE issueID=:issue_id",
+        ["issue_id" => $issueID]
+    );
+    if (!$existingDescription) {
+        $description = $DB->pselectOne(
+            "SELECT issueComment
+            FROM issues_comments ic
+            WHERE ic.issueID = :issue_id
+            ORDER BY ic.dateAdded ASC
+            LIMIT 1",
```
driusan, line 51: pselectOne already adds a limit. Removing the "limit 1" here should fix your error.

ay-bh: @driusan That gives another error, since the LIMIT is 2. Using pselect directly solves it. I am just doing some final tests, will commit after that. What are your thoughts?

driusan: is the query expected to return more than 1 row? pselectOne does a LIMIT 2 because that's sufficient to efficiently tell if it returned zero rows and should be null, returned multiple rows and should throw an exception (because the code expected one value but the query matched many rows), or returned exactly 1 rows and it should be returned.

If the query is expected to have multiple matches and the code handles them properly then pselect is fine. But you should not run pselect with a "LIMIT 1" because that will just mask the root cause of the error and will return a random one of the multiple rows that the query matched.

ay-bh: Yes without limit 1 it will return multiple rows.  The multiple matches will be handled by this `ORDER BY ic.dateAdded ASC LIMIT 1`. This will also ensure that there is no randomness. Shall I proceed with pselect?

driusan: if there is an order by and you're sure the first one by date is the value that you want, then that would make sense.

ay-bh (approved):

racostas (approved):
@driusan, I re-reviewed this one. All looks good to me. If all fine, let merge it. It's a good feature for the study tracker.

PR #8867, opened 2023-08-15
[instrument_list] Fix deprecation warning

driusan:
Fix the warnings:

```
<b>Deprecated</b>:  strtolower(): Passing null to parameter #1 ($string) of type string is deprecated in <b>modules/instrument_list/php/instrument_list_controlpanel.class.inc</b> on line <b>475</b><br />
<br />
<b>Deprecated</b>:  strtolower(): Passing null to parameter #1 ($string) of type string is deprecated in <b>modules/instrument_list/php/instrument_list_controlpanel.class.inc</b> on line <b>514</b><br />
```

in PHP 8.2 and above (and maybe below, I didn't check.)


kongtiaowang (approved):
LGTM

